PR #2057, opened 2016-02-04
Make usage of the performance.now method safer and more centralized

### LINKED ISSUE #2056: existence of window.performance does not imply performance.now method
Until [Android 4.4](http://caniuse.com/#search=performance), Android had a performance object but not the now object. I'm seeing reports of a wide range of niche browsers - Facebook, Chrome Mobile, Firefox - that have performance but not performance.now. Our feature checks for this need to be more robust.

tmcw:
Fixes #2056 

cc @ansis for the review

### DIFF
--- a/js/util/browser/browser.js
+++ b/js/util/browser/browser.js
@@ -2,6 +2,25 @@
 
 var Canvas = require('./canvas');
 
+/*
+ * Unlike js/util/browser.js, this code is written with the expectation
+ * of a browser environment with a global 'window' object
+ */
+
+/**
+ * Provides a function that outputs milliseconds: either performance.now()
+ * or a fallback to Date.now()
+ * @private
+ */
+module.exports.now = (function() {
+    if (window.performance &&
+        window.performance.now) {
+        return window.performance.now.bind(window.performance);
+    } else {
+        return Date.now.bind(Date);
+    }
+}());
+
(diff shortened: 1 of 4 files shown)


ansis:
look good to me

js/util/browser/browser.js
```diff
@@ -2,6 +2,27 @@
...
+var nowFn = (function() {
+    if (window.performance &&
+        window.performance.now) {
+        return window.performance.now.bind(window.performance);
+    } else {
+        return Date.now.bind(Date);
+    }
+}());
+
+module.exports.nowFn = nowFn;
```
lucaswoj: Can we call this method `now`? We don't need to use Hungarian notation here.

tmcw: We've been [calling its output now](https://github.com/mapbox/mapbox-gl-js/pull/2057/files#diff-e60957272552aeccdb45f65d82d523e1L47). Unless we rename the output, we shouldn't call this now.

lucaswoj: Something more descriptive, like `getEpochMilliseconds`, then?

jfirebaugh:
Please call it `now`. There is nothing wrong with `var now = browser.now()`.

tmcw:
Okay, renamed to `now`

Issue #1632, opened 2015-10-19
validate style API inputs


jfirebaugh:
The following methods should validate their inputs with the style spec validator:

* `addSource`
* `addLayer`
* `setLayoutProperty`
* `setPaintProperty`
* `setFilter`

This will require the validator to expose granular validation methods for specific types.

Will catch issues like #1631.

mourner:
+ validate `source` and `source-layer` references to make sure those are present (not sure if this can be validated on the spec side)

lucaswoj:
We can validate that a layer's `source` is defined but cannot validate the existence of a `source-layer` without downloading a tile (or maybe all the tiles). I don't think that behavior belongs in the style validator -- it would require a network connection, API keys, and potentially large downloads.

lucaswoj:
currently blocked on mapbox/mapbox-gl-style-spec#407

PR #2063, opened 2016-02-04
Add Map#getStyle method

### LINKED ISSUE #1982: Possibility to get current map style state
I'm probably using a non-supported way to pass styles between two map instances using map.style.stylesheet which works quite good. The obvious problem with this approach is that I don't get any sources or layers that were added programmatically to the map. Could the API be extended with a getStyle() method which returns the current *effective* style including sources and layers added programmatically? Essentially I want to be able to do map2.setStyle(map1.getStyle())...

lucaswoj:
This makes #1632 a little cleaner, so I hacked it together.

fixes #1982 

cc @ansis @jfirebaugh @tmcw @bhousel

### DIFF
--- a/debug/site.js
+++ b/debug/site.js
@@ -11,7 +11,7 @@ var map = new mapboxgl.Map({
 
 map.addControl(new mapboxgl.Navigation());
 
-map.on('style.load', function() {
+map.on('load', function() {
     map.addSource('geojson', {
         "type": "geojson",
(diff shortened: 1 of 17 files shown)


jfirebaugh:
:+1: This looks great.

lucaswoj:
Great! I just pushed a little tweak (storing a clone of the `VectorSource` options instead of the original reference. I'll merge on green.

PR #2064, opened 2016-02-05
fix tile debug rendering

### LINKED ISSUE #2034: map debug setting incorrect in 0.13.x
The `.debug` option on the map object only shows a single tile outlined in 0.13, which is a regression from 0.12.x, which showed all tile boundaries when the debug option was turned on.

2016-02-01 at 12 29 pm

ansis:
fix #2034

- [x] add render test https://github.com/mapbox/mapbox-gl-test-suite/issues/74

### DIFF
--- a/js/render/draw_debug.js
+++ b/js/render/draw_debug.js
@@ -2,40 +2,53 @@
 
 var textVertices = require('../lib/debugtext');
 var browser = require('../util/browser');
+var mat4 = require('gl-matrix').mat4;
+var EXTENT = require('../data/buffer').EXTENT;
 
 module.exports = drawDebug;
 
-function drawDebug(painter, coords) {
+function drawDebug(painter, source, coords) {
     if (painter.isOpaquePass) return;
     if (!painter.options.debug) return;
 
     for (var i = 0; i < coords.length; i++) {
-        drawDebugTile(painter, coords[i]);
+        drawDebugTile(painter, source, coords[i]);
     }
 }
 
-function drawDebugTile(painter, coord) {
+function drawDebugTile(painter, source, coord) {
     var gl = painter.gl;
 
+    gl.lineWidth(1 * browser.devicePixelRatio);
+
+    var posMatrix = painter.calculatePosMatrix(coord, source.maxzoom);
     var shader = painter.debugShader;
-    gl.switchShader(shader, painter.calculatePosMatrix(coord));
+    gl.switchShader(shader, posMatrix);
 
     // draw bounding rectangle
     gl.bindBuffer(gl.ARRAY_BUFFER, painter.debugBuffer);
     gl.vertexAttribPointer(shader.a_pos, painter.debugBuffer.itemSize, gl.SHORT, false, 0, 0);
     gl.uniform4f(shader.u_color, 1, 0, 0, 1);
-    gl.lineWidth(4);
     gl.drawArrays(gl.LINE_STRIP, 0, painter.debugBuffer.itemCount);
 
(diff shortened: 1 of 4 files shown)


ansis:
@lucaswoj can you review?

js/render/draw_debug.js
```diff
@@ -2,40 +2,53 @@
...
+    // Draw the halo with multiple 1px lines instead of one wider line because
+    // the gl spec doesn't guarantee support for lines with width > 1.
+    var tileSize = source.getTile(coord).tileSize;
+    var onePixel = EXTENT / (Math.pow(2, painter.transform.zoom - coord.z) * tileSize);
+    for (var x = -1; x <= 1; x += 2) {
+        for (var y = -1; y <= 1; y += 2) {
+            gl.setPosMatrix(mat4.translate([], posMatrix, [onePixel * x, onePixel * y, 0]));
+            gl.drawArrays(gl.LINES, 0, vertices.length / painter.debugTextBuffer.itemSize);
+        }
+    }
```
lucaswoj: :grimacing:

js/render/painter.js
```diff
@@ -152,7 +152,7 @@ Painter.prototype.setup = function() {
     gl.bufferData(
         gl.ARRAY_BUFFER,
         new Int16Array([
-            0, 0, EXTENT - 1, 0, EXTENT - 1, EXTENT - 1, 0, EXTENT - 1, 0, 0]),
+            0, 0, EXTENT, 0, EXTENT, EXTENT, 0, EXTENT, 0, 0]),
```
lucaswoj: Why did we have `- 1` in here before?

js/render/draw_debug.js
```diff
@@ -2,40 +2,53 @@
...
     gl.uniform4f(shader.u_color, 1, 1, 1, 1);
-    gl.drawArrays(gl.LINES, 0, vertices.length / painter.debugTextBuffer.itemSize);
-    gl.lineWidth(2 * browser.devicePixelRatio);
+
+    // Draw the halo with multiple 1px lines instead of one wider line because
+    // the gl spec doesn't guarantee support for lines with width > 1.
+    var tileSize = source.getTile(coord).tileSize;
+    var onePixel = EXTENT / (Math.pow(2, painter.transform.zoom - coord.z) * tileSize);
+    for (var x = -1; x <= 1; x += 2) {
+        for (var y = -1; y <= 1; y += 2) {
```
lucaswoj: I think this loop would be more intuitive written as 

```js
var translations = [[-1, -1], [-1, 1], [1, -1], [1, 1]];
for (var i = 0, i < translations.length, ...
```

js/render/draw_debug.js
ansis: yeah, it would be good to switch away from gl.LINES completely. Until then we can do this or not render halos. I'm not sure which is worse.

js/render/painter.js
ansis: I'm not sure. I think conceptually with the -1 the lines are always strictly within a tile's boundaries. But since gl.LINES are snapped to the nearest pixel this isn't actually useful. -native doesn't have this

With the -1 the tile boundaries sometimes overlap with neighbours (creating a 1px line) and sometimes doesn't overlap (creating a 2px line). This makes it a consistent 1px.

js/render/draw_debug.js
ansis: I switched to this and pushed a new version.

lucaswoj:
Looks good to me! :ship:

PR #2065, opened 2016-02-05
consider size change when placing symbols

### LINKED ISSUE #1196: Make collision detection take into account "text-offset" and size changes
In the Y direction, the symbol simply outruns its bounding box. Looking at `France`, going from [0,-10] to [0,0] to [0,10]:

offsety

In the X direction... it's bizarre. Looks like the box both scales and shifts. Going from [-10,0] to [0,0] to [10,0]:

offsetx

cc/ @ansis

ansis:
fix #1196

Symbols were sometimes not contained by their collision boxes. This fixes that.

Symbols can size change within a single zoom level. When size changes all the units that are relative to size, like text-offset, change.

Collision detection can't handle boxes that scale at different rates so we try to create a collision box that will cover the symbol at all sizes within that zoom level.

It used to just use the size at z+1 with the assumption that text always increases and that the anchor is always within the collision box.

This commit makes it use both the size at z and z+1 so that symbol is always within the collision box.
- [ ] add new render test (related to https://github.com/mapbox/mapbox-gl-test-suite/issues/74 )
- [ ] fix tests

### DIFF
--- a/js/data/symbol_bucket.js
+++ b/js/data/symbol_bucket.js
@@ -36,10 +36,16 @@ function SymbolBucket(options) {
     var zoomHistory = { lastIntegerZoom: Infinity, lastIntegerZoomTime: 0, lastZoom: 0 };
 
     this.adjustedTextMaxSize = this.layer.getLayoutValue('text-size', 18, zoomHistory);
-    this.adjustedTextSize = this.layer.getLayoutValue('text-size', this.zoom + 1, zoomHistory);
+    this.adjustedTextSizes = [
+        this.layer.getLayoutValue('text-size', this.zoom, zoomHistory),
+        this.layer.getLayoutValue('text-size', this.zoom + 1, zoomHistory)
+    ];
 
     this.adjustedIconMaxSize = this.layer.getLayoutValue('icon-size', 18, zoomHistory);
-    this.adjustedIconSize = this.layer.getLayoutValue('icon-size', this.zoom + 1, zoomHistory);
+    this.adjustedIconSizes = [
+        this.layer.getLayoutValue('icon-size', this.zoom, zoomHistory),
+        this.layer.getLayoutValue('icon-size', this.zoom + 1, zoomHistory)
+    ];
 }
 
 SymbolBucket.prototype = util.inherit(Bucket, {});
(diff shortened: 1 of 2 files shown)


js/data/symbol_bucket.js
```diff
@@ -36,10 +36,16 @@ function SymbolBucket(options) {
...
+        this.layer.getLayoutValue('text-size', this.zoom, zoomHistory),
+        this.layer.getLayoutValue('text-size', this.zoom + 1, zoomHistory)
+    ];
 
     this.adjustedIconMaxSize = this.layer.getLayoutValue('icon-size', 18, zoomHistory);
-    this.adjustedIconSize = this.layer.getLayoutValue('icon-size', this.zoom + 1, zoomHistory);
+    this.adjustedIconSizes = [
+        this.layer.getLayoutValue('icon-size', this.zoom, zoomHistory),
+        this.layer.getLayoutValue('icon-size', this.zoom + 1, zoomHistory)
+    ];
```
lucaswoj: This isn't an ideal general-case solution because zoom stops can be at fractional zooms. How would you feel about adding a `getMaxLayoutValue(name, minZoom, maxZoom, zoomHistory)` that does a little more function introspection?

ansis: Yes, if a value between z and z+1 is bigger or smaller than both of the two values at z and z + 1 then this solution isn't ideal. Having a `getMaxLayoutValue` and a `getMinLayoutValue` would be more correct but I'm not sure it would make a significant enough difference to be worth the extra complexity.

For this to make an actual difference in practice the difference between the maximum size between z and z + 1 would need to be significantly bigger than the maximum of the value at z and z + 1.

It might be worth it. What do you think? If it's worth it, do you want to take a shot at adding this?

lucaswoj: It is a simple algorithm. It really ought to go in the `mapbox-gl-function` repo but I'd look the other way. 

```
Math.max(at(minZoom), at(maxZoom), at(all stops between minzoom and maxzoom)
```

I have too much on my plate to tackle this in the next couple days. I think you should do it. Its 3 - 4 lines of code that'll 

 - be the fully correct solution to the problem
 - eliminate all known edge cases
 - expose an API that might be useful in the future
 - hedge against ever having to look at this particular code again

mourner:
@ansis any updates here?

lucaswoj:

PR #2073, opened 2016-02-05
Fire map#load if no initial style is set

### LINKED ISSUE #2042: Map#load never fires fire if a style isn't passed to the constructor
... even if a style is later set with `setStyle`

h/t @scothis

lucaswoj:
fixes #2042

cc @scothis @jfirebaugh @ansis @bhousel

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -736,7 +736,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     loaded: function() {
         if (this._styleDirty || this._sourcesDirty)
             return false;
-        if (this.style && !this.style.loaded())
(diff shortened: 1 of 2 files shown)


jfirebaugh:
:-1: 

I don't expect `load` to be fired if there is no style.

scothis:
Feels like you're attacking #2042 from the wrong side. I agree with https://github.com/mapbox/mapbox-gl-js/issues/2042#issuecomment-180598721

lucaswoj:
Great. I agree. Pushing a revision now.

lucaswoj:
@jfirebaugh @scothis How do you feel about the latest revision?

jfirebaugh:
:+1: :star2:

Issue #2030, opened 2016-01-30
Update to 0.13.1 breaks line-width


BrunoSalerno:
I updated to 0.13.1 version (from 0.12.3) and I get this error while loading:
`Error: layers[34].paint.line-width.stops[1][0]: array stops must appear in ascending order and have no duplicates`.
**I'm not using stops, just `"line-width":9`**.

See the console here: https://jsfiddle.net/brunosalerno/zn2abLok/ . It looks like a style that was working fine in the previous version now it's not.

Thanks

mourner:
@lucaswoj something probably regressed from the StyleLayer refactoring.

lucaswoj:
Thanks for providing a fiddle @BrunoSalerno! I think this style sheet is actually invalid. I see, for example,

```json
{"interactive":true,"layout":{"line-cap":"butt","line-join":"miter","visibility":"visible"},"metadata":{"mapbox:group":"1444850892815.3088"},"filter":["all",["in","class","major_rail","minor_rail"],["==","type","rail"]],"type":"line","source":"mapbox://mapbox.mapbox-streets-v6","id":"road-rail-tracks","paint":{"line-color":"rgba(161,156,156,1)","line-width":{"base":1.5,"stops":[[22,1],[22,1]]},"line-opacity":1,"line-dasharray":[1,1]},"source-layer":"road"},
```

... which does have some duplicate stops.

```
"line-width":{"base":1.5,"stops":[[22,1],[22,1]]}
```

This might be a strictening of our style validation, a bug in one of our basemaps, or caused by a `StyleLayer` refactor. I'll dig in further on Monday. 

Thank you for your patience!

BrunoSalerno:
@lucaswoj thanks! There were, indeed, duplicated stops in the style I was working on. When I removed them in the Mapbox studio it started to work. So It looks like the style validation is more strict now. 

The problem is that now I'm getting the following error when trying to render a geojson line layer:
 `Uncaught TypeError: Cannot read property 'type' of undefined`   (`draw_line.js:77`). 
That line is 
```js
posA = painter.lineAtlas.getDash(dasharray.from, layer.layout['line-cap'] === 'round');
```
The `paint` object I'm using width the geojson line layer is as follows:
```json
{"line-color": "#000",
 "line-width": 7,
 "line-opacity": 0.4}
```

¿Do you think that this is related to the new *strictness*?

BrunoSalerno:
[...]

PR #2075, opened 2016-02-05
Bump min style spec version to 8.4.1

scothis:


lucaswoj:
I say :shipit: 

For reference, the only difference between 8.4.1 and 8.4.0 is https://github.com/mapbox/mapbox-gl-style-spec/commit/2ced5102d3be7d4307aaa7bfe5974d821aaa3d86, a relaxation of function validation that was breaking many styles (e.g. https://github.com/mapbox/mapbox-gl-js/issues/2030)

Issue #2067, opened 2016-02-05
points produced by clipLine are not integers


ansis:
The `x` and `y` components of points created by `clipLine` are not integers. https://github.com/mapbox/mapbox-gl-js/blob/master/js/symbol/clip_line.js#L33

In -native they get truncated to integers: https://github.com/mapbox/mapbox-gl-native/blob/fa210516834feada45d69e218f9c1dce2d2ee6f6/src/mbgl/util/clip_lines.cpp#L24

in some other spots we round instead of truncating: https://github.com/mapbox/mapbox-gl-native/blob/fa210516834feada45d69e218f9c1dce2d2ee6f6/src/mbgl/text/get_anchors.cpp#L35

This should be consistent

Should we have separate float and integer point types in -js?
Should we truncate or round when converting from float to integer points?
Or should we just use floats for points in -native? 

@jfirebaugh what are the right answers?

mourner:
I think there should be `round` everywhere.

PR #2077, opened 2016-02-05
make sure icons fit before the ends of the line

### LINKED ISSUE #2062: oneway arrows going in the wrong direction
screen shot 2016-02-04 at 1 16 48 pm

I think this is happening near tile edges.

http://localhost:9966/debug/#16.1/38.90339/-77.03731

https://github.com/mapbox/mapbox-gl-js/pull/1980 might fix this.

ansis:
fixes #2062

This skips anchors if there is not enough room before the end of the line for the icon to fit.

Previously it allowed icons that extended past the end of a line as long as the icon's anchor fit on the line. This changes it so that the icon itself has to fit on the line. With these changes there are no icons on really short lines and no icons right at the end of lines.

The test suite changes (https://github.com/mapbox/mapbox-gl-test-suite/commit/6b6a03ebe8c106db1a85e88572293e523c71ba72) make it look like this is a huge change, but in actual maps it only drops a small number of icons that are right at the ends of lines or are on really short lines.

#2062 would also be fixed by https://github.com/mapbox/mapbox-gl-js/issues/2067 (which we should still do).

:eyes: @lucaswoj @jfirebaugh

### DIFF
--- a/js/symbol/get_anchors.js
+++ b/js/symbol/get_anchors.js
@@ -48,6 +48,12 @@ function getAnchors(line, spacing, maxAngle, shapedText, shapedIcon, glyphSize, 
 
 function resample(line, offset, spacing, angleWindowSize, maxAngle, labelLength, continuedLine, placeAtMiddle, tileExtent) {
 
+    var halfLabelLength = labelLength / 2;
+    var lineLength = 0;
+    for (var k = 0; k < line.length - 1; k++) {
+        lineLength += line[k].dist(line[k + 1]);
+    }
+
     var distance = 0,
         markedDistance = offset - spacing;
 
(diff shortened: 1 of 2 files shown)


js/symbol/get_anchors.js
```diff
@@ -68,7 +74,9 @@ function resample(line, offset, spacing, angleWindowSize, maxAngle, labelLength,
                 x = interpolate(a.x, b.x, t),
                 y = interpolate(a.y, b.y, t);
 
-            if (x >= 0 && x < tileExtent && y >= 0 && y < tileExtent) {
+            if (x >= 0 && x < tileExtent && y >= 0 && y < tileExtent &&
+                    markedDistance - halfLabelLength >= 0 &&
+                    markedDistance + halfLabelLength <= lineLength) {
```
lucaswoj: This statement could use some more documentation, either in the form of a `// comment` or factored out subexpressions

ansis: added comment f914a34

lucaswoj:
This looks good to me.

PR #2085, opened 2016-02-08
Disable handlers before enabling

### LINKED ISSUE #2069: Handlers should be aware of their enabled/disabled state.
Currently handlers like [DoubleClickZoom](https://github.com/mapbox/mapbox-gl-js/blob/2e326cecd534658f7d04220b3ab13adeac98af75/js/ui/handler/dblclick_zoom.js#L12) are unaware of their enabled/disabled state. It is also not possible as a consumer of mapbox-gl to know this state. This makes it impossible to know if you should renable a handler that you have disabled. I think the best place to check for this is in the handler's enable function. A consumer could fake this themselves, but that will lead to some hard to debug bugs when another application or plugin is also enabled and disabling handlers.

I'm currently handling this is gl-draw by disabling before I enable.

lucaswoj:
… in order to prevent duplicate event listeners from being bound

fixes #2069 

cc @mcwhittemore @bhousel @ansis

### DIFF
--- a/js/ui/handler/box_zoom.js
+++ b/js/ui/handler/box_zoom.js
@@ -17,6 +17,7 @@ function BoxZoom(map) {
 
 BoxZoom.prototype = {
     enable: function () {
+        this.disable();
         this._el.addEventListener('mousedown', this._onMouseDown, false);
     },
 
--- a/js/ui/handler/dblclick_zoom.js
+++ b/js/ui/handler/dblclick_zoom.js
@@ -9,6 +9,7 @@ function DoubleClickZoom(map) {
 
 DoubleClickZoom.prototype = {
     enable: function () {
+        this.disable();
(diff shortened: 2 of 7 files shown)


bhousel:
Calling `disable` is a good idea, but I got the impression that #2069 was more asking for these methods to be written as accessor/mutators, so that other code could get the current handler state.

```js
handler.enabled();
// returns true/false

handler.enabled(false);
// disables handler
```

mcwhittemore:
@bhousel getters might be worthwhile, but the problem I was running into was too many callbacks being bound by calling enable when the handler was all ready enabled. This should solve that problem.

lucaswoj:
Let's merge this as-is and create getters and setters if a use case emerges.

mo' state, mo' problems.

PR #2093, opened 2016-02-09
Don't share an options object between instances of Map

lucaswoj:
cc @ansis @jfirebaugh @bhousel 

If there were multiple instances of `Map` on a page, they were all sharing the same `options` object.


lucaswoj:
Ah. I jumped to conclusions. We were using `inherit`, which does clone `options`. I still think this refactor is more clear and more safe.

jfirebaugh:
I'd rather keep the existing code. See https://github.com/Leaflet/Leaflet/issues/2294.

lucaswoj:
:+1:

Issue #145, opened 2013-10-23
OpenGL state management system


kkaefer:
We should maintain all current state in the painter object so that we can avoid doing redundant calls (e.g. to bind buffers or switch shaders). Some of that is already done in the `switchShader` code, but we're still happily binding buffers repeatedly and setting other WebGL state that isn't strictly necessary.

lucaswoj:
My dream interface here is to do away with all GL getters / setters in favor of passing all desired state to the `draw*` call as an object. This is a clean abstraction that will allow us to run the smallest possible # of WebGL api calls, prevent bugs stemming from unexpected GL state, and make our code easier to understand.

jfirebaugh:
I've been meaning to have a closer look at the API of [glium](https://github.com/tomaka/glium) (Rust OpenGL bindings). It promises:

> Glium is stateless. There are no set_something() functions in the entire library, and everything is done by parameter passing. The same set of function calls will always produce the same results, which greatly reduces the number of potential problems.

lucaswoj:
As part of this refactoring, we should merge `Painter` and `gl_utils`

lucaswoj:
:thought_balloon: 

```js
gl2.drawElements({
    shader: getShader('circle'),
    mode: gl2.TRIANGLES,

    // Allow circles to be drawn across boundaries, so that
    // large circles are not clipped to tiles
    stencilTest: false,

    depthMask: false,
    depthSubrange: getDepthSubrange(0),

    count: group.getCount(),
    offset: group.getOffset(),
    elementBuffer: group.getElementBuffer(),
    vertexBuffer: group.getVertexBuffer(),

    attributes: {
        a_pos: group.getVertexBuffer().pos,
    },

    uniforms: {
        u_exmatrix: transform.exMatrix,
        u_posmatrix: translatePosMatrix(
            calculatePosMatrix(coord, source.maxzoom),
            tile,
            layer.paint['circle-translate'],
            layer.paint['circle-translate-anchor']
        ),
        u_color: util.premultiply(layer.paint['circle-color'], layer.paint['circle-opacity']),
        u_blur: Math.max(layer.paint['circle-blur'], 1 / browser.devicePixelRatio / layer.paint['circle-radius'])),
        u_size: layer.paint['circle-radius']
    }
});
```

lucaswoj:
It'd be cool to build this as an external library, maybe including our `Buffer` implementation too.

lucaswoj:
[...]

PR #2095, opened 2016-02-09
Fix incorrect rendering caused by drawing symbols across edges

### LINKED ISSUE #2092: Layout properties for icon behaves strange
```javascript
  let pinLayer = {
    'id':'pin',
    'type': 'symbol',
    'source':'location-pin',

    'layout': {
      'icon-image': 'map-pin'
      //'icon-allow-overlap': true,
      //'icon-ignore-placement': true
    },
    'interactive' : false,
    'minzoom': minZoom,
    'maxzoom': maxZoom
  };

map.addLayer(pinLayer);
```
On switching to the latest version 0.14.0, I have observed that the style breaks on ```icon-allow-overlap``` and ```icon-ignore-placement```.

On commenting those properties, the map looks like this (expected)
screen shot 2016-02-09 at 5 53 02 pm

And on adding those properties, the map renders like this
screen shot 2016-02-09 at 5 53 34 pm

P.S.: I am using the style 'bright-v8'. And this is also the case with ```text-allow-overlap``` and ```text-ignore-placement```

lucaswoj:
fixes #2092

cc @ansis @jfirebaugh @AbrahamArun

Ideally this would come with a rendering test but there isn't much precedent for rendering tests that exercise the interactions between layers. I think https://github.com/mapbox/mapbox-gl-test-suite/issues/26 is the right long term solution. 

This bug would have been prevented by https://github.com/mapbox/mapbox-gl-js/issues/145

### DIFF
--- a/js/render/draw_background.js
+++ b/js/render/draw_background.js
@@ -74,7 +74,6 @@ function drawBackground(painter, source, layer) {
         gl.drawArrays(gl.TRIANGLE_STRIP, 0, painter.tileExtentBuffer.itemCount);
     }
 
-    gl.enable(gl.STENCIL_TEST);
     gl.stencilMask(0x00);
     gl.stencilFunc(gl.EQUAL, 0x80, 0x80);
 }
--- a/js/render/draw_circle.js
+++ b/js/render/draw_circle.js
@@ -64,6 +64,4 @@ function drawCircles(painter, source, layer, coords) {
             gl.drawElements(gl.TRIANGLES, count, gl.UNSIGNED_SHORT, elementOffset);
         }
     }
-
-    gl.enable(gl.STENCIL_TEST);
 }
--- a/js/render/draw_collision_debug.js
+++ b/js/render/draw_collision_debug.js
@@ -38,6 +38,4 @@ function drawCollisionDebug(painter, source, layer, coords) {
         );
 
     }
-
-    gl.disable(gl.STENCIL_TEST);
 }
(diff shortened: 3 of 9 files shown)


ansis:
I think this is still buggy in some cases

Based on the [conditional disable](https://github.com/mapbox/mapbox-gl-js/blob/6f9ffa8bb6dfce0a354875b1cd58f31737462551/js/render/draw_symbol.js#L25) I think drawSymbols assumes the stencil test is enabled when it starts. With this change the stencil test is always disabled after it ends. If the next layer is `drawAcrossTileEdges === false` then it should be clipped but it won't be.

We should stop expecting that the stencil test has a certain state and set it to whatever we need it to be for that layer. It's way too statefull right now.

Or if we want to leave that for later, I think removing the `gl.disable(gl.STENCIL_TEST)` at the end of `drawCollisionDebug` would fix this.

lucaswoj:
Good catch. I will both remove the `gl.disable(gl.STENCIL_TEST)` at the end of `drawCollisionDebug` and explicitly call `gl.enable(gl.STENCIL_TEST)` at the beginning of `drawSymbols`

lucaswoj:
@ansis Does the latest push look ok? I reviewed all our `draw_*` methods and, as of this PR, they all explicitly set their desired `gl.STENCIL_TEST` state at the beginning and enable `gl.STENCIL_TEST` at the end for good measure.

ansis:
I think there are a couple more spots it should be enabled:
- drawFill should enable it
- drawLine should enable it
- I think this needs to enable it as well https://github.com/mapbox/mapbox-gl-js/blob/ce3195f50d937977f2b5d9464bcd959d0a0be0ae/js/render/painter.js#L190

If all calls are expected to set their desired value we can probably drop all the `enable`s at the ends of the draw functions.

ansis:
looks good to me!

lucaswoj:
Thanks for being thorough @ansis! :ship:

Issue #2100, opened 2016-02-10
featuresAt behaviour changed in 0.14.0


averas:
I noticed a bug in mapbox-gl-draw (https://github.com/mapbox/mapbox-gl-draw/issues/225) which is only present when using mapbox-gl-js 0.14.0 and not 0.13.1. It seems as if when running featuresAt on a layer with features that are close to each other (geographically) they start to interfere.

If you bring up http://codepen.io/anon/pen/wMRBgr and click on features you will see that it is working as expected. If you however change the mapbox-gl-js version from 0.13.1 to 0.14.0 you will see that when you click on a marker, another marker will be selected. You can even click far out on the background, and still markers will be selected, even though radius is set to 5.

mcwhittemore:
@averas we might be able to patch this fix in mapbox-gl-draw by sorting the features returned by closet to the click point.

averas:
@mcwhittemore Alright, but there must also be some kind of bug involved.. even with a featuresAt radius set to 5 in the example I posted you get hits much further away than that.

mcwhittemore:
@averas spent some time debugging this on the gl-draw side and I agree, this is best handled in gl-js. Looking for the commit that caused the regression now

mcwhittemore:
@ansis it looks like 1da1e5ab2bd10ea4d427d4f8dacd77e54f47e0b9 is the culprit here

ansis:
@mcwhittemore thanks for tracking it down

https://github.com/mapbox/mapbox-gl-js/pull/2104 should fix this

ansis:
fixed by https://github.com/mapbox/mapbox-gl-js/pull/2104

PR #2052, opened 2016-02-04
make featuresAt return symbols and use style properties

ansis:
This adds currently visible text and icons to featuresAt results.

Style properties are used to return features whose rendered representation matches the query.

query test pr: https://github.com/mapbox/mapbox-gl-test-suite/pull/77

fix #303
fix #316
fix #862
fix #902
fix #1822
fix #2100
fix #2103

This increases worker memory usage by around 10% (a couple of MB).
1/3 of the increase comes from retaining the CollisionTile.
2/3 of the increase comes from retaining VectorTileFeatures.

SymbolBucket merges lines to improve placement. This means that sometimes there is more than one VectorTileFeature responsible for that symbol. It currently returns the first VectorTileFeature and ignores the others that were merged into it. Should it do something else?

:eyes: @jfirebaugh

PR #2104, opened 2016-02-10
fix featuresAt with overscaled tiles

### LINKED ISSUE #2103: featuresAt doesn't return anything on overscaled / underscaled tiles
cc @ansis

ansis:
fix #2103 and https://github.com/mapbox/mapbox-gl-js/issues/2100#issuecomment-182446562

:eyes: @lucaswoj

### DIFF
--- a/test/js/data/feature_tree.test.js
+++ b/test/js/data/feature_tree.test.js
@@ -22,6 +22,8 @@ test('featuretree', function(t) {
     ft.insert(feature.bbox(), 'road', feature);
     ft.query({
         params: { },
+        scale: 1,
+        tileSize: 512,
         x: 0,
         y: 0
     }, function(err, features) {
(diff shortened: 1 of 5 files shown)


test/js/data/feature_tree.test.js
```diff
@@ -68,7 +72,8 @@ test('featuretree point query', function(t) {
 
     ft.query({
         source: "mapbox.mapbox-streets-v5",
-        scale: 724.0773439350247,
+        scale: 1.4142135624,
+        tileSize: 512,
```
lucaswoj: were these wrong before?

ansis: `1.4142135624 * 512 = 724.0773439350247`

I split the number into two numbers in https://github.com/mapbox/mapbox-gl-js/pull/2052 because I thought it would be clearer and closer to how we treat conversions elsewhere.

lucaswoj: I understand. This is much more clear :+1:

lucaswoj:
Can you add regression tests for the two bugs fixed? Otherwise this looks :ship:.

ansis:
Added two tests for `tileAt` to cover the two bugs.

PR #2105, opened 2016-02-11
Fixed bug where unsetting style properties could mutate the value of other style properties

lucaswoj:
cc @scothis @ansis @jfirebaugh

### DIFF
--- a/js/style/style_layer.js
+++ b/js/style/style_layer.js
@@ -164,18 +164,19 @@ StyleLayer.prototype = {
         for (var klass in this._paintDeclarations) {
             if (klass !== "" && !classes[klass]) continue;
             for (var name in this._paintDeclarations[klass]) {
-                applyDeclaration(this._paintDeclarations[klass][name]);
+                applyDeclaration(name, this._paintDeclarations[klass][name]);
             }
         }
 
         // Apply removed declarations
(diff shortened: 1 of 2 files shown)


js/style/style_layer.js
```diff
@@ -164,18 +164,19 @@ StyleLayer.prototype = {
...
         for (var i = 0; i < removedNames.length; i++) {
-            var spec = this._paintSpecifications[removedNames[i]];
-            applyDeclaration(new StyleDeclaration(spec, spec.default));
+            name = removedNames[i];
+            var spec = this._paintSpecifications[name];
+            applyDeclaration(name, new StyleDeclaration(spec, spec.default));
         }
 
-        function applyDeclaration(declaration) {
+        function applyDeclaration(name, declaration) {
```
lucaswoj: Not passing `name` as an argument to `applyDeclaration` was a mistake. We captured the outer scope's `name`, which wasn't always the desired value.

ansis: should we add `block-scoped-var` to the eslint config to catch this kind of bug?

lucaswoj: That's a good idea. I'll make it a separate PR.

ansis:
looks ready to me!

PR #2109, opened 2016-02-11
Fix the serialization of paint classes

### LINKED ISSUE #2107: Map#getStyle() gets confused about classes
If I have a layer with a class, such as "paint.night", <code>Map#getStyle()</code> will return a layer with a property called "paint.paint" instead of "paint.night".

Demo: http://codepen.io/anon/pen/JGwerB

lucaswoj:
fixes #2107

cc @ansis @averas


ansis:
looks good to me!

PR #2113, opened 2016-02-11
round points created during bucket creation

### LINKED ISSUE #2067: points produced by clipLine are not integers

ansis:
fix #2067

:eyes: @lucaswoj

### DIFF
--- a/js/data/line_bucket.js
+++ b/js/data/line_bucket.js
@@ -174,7 +174,7 @@ LineBucket.prototype.addLine = function(vertices, join, cap, miterLimit, roundLi
         if (isSharpCorner && i > 0) {
             var prevSegmentLength = currentVertex.dist(prevVertex);
             if (prevSegmentLength > 2 * sharpCornerOffset) {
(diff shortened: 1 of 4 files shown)


lucaswoj:
Looks good, :shipit:! Make sure to merge https://github.com/mapbox/mapbox-gl-test-suite/commit/8859b504ef241bca49d4c2f9a79ff4d0bfdf81a1 into `master` too.

PR #2114, opened 2016-02-12
reset line distance when it is near max representable distance

### LINKED ISSUE #2099: dashed line bug
screen shot 2016-02-09 at 3 46 34 pm

I'm guessing this is what happens when `distance` is > than the maximum distance we can store in the buffers.

ansis:
fix #2099 

When the distance of the current vertex is > half the maximum distance that can be represented, reset the distance. This fixes the original problem where a line with many segments is longer than the maximum distance.

We still need to make sure we can represent the distance of the longest possible segment so we need to make two more changes:
- improve how `linesofar` is packed into the buffer so that we have two more bits
- scale the distance down before adding it to the buffer to increase the maximum distance that can be represented

I don't like how `var distance` changed to `this.distance` but `addCurrentVertex` needs to be able to reset the distance and I'm not seeing a better way to do this.

:eyes: @lucaswoj

### DIFF
--- a/js/data/line_bucket.js
+++ b/js/data/line_bucket.js
@@ -53,17 +53,19 @@ LineBucket.prototype.shaders = {
         }, {
             name: 'data',
             components: 4,
-            type: Bucket.AttributeType.BYTE,
+            type: Bucket.AttributeType.UNSIGNED_BYTE,
             value: [
-                'Math.round(' + EXTRUDE_SCALE + ' * extrude.x)',
-                'Math.round(' + EXTRUDE_SCALE + ' * extrude.y)',
+                // add 128 to store an byte in an unsigned byte
+                'Math.round(' + EXTRUDE_SCALE + ' * extrude.x) + 128',
+                'Math.round(' + EXTRUDE_SCALE + ' * extrude.y) + 128',
 
-                // Encode the -1/0/1 direction value into .zw coordinates of a_data, which is normally covered
-                // by linesofar, so we need to merge them.
-                // The z component's first bit, as well as the sign bit is reserved for the direction,
-                // so we need to shift the linesofar.
-                '((dir < 0) ? -1 : 1) * ((dir ? 1 : 0) | ((linesofar << 1) & 0x7F))',
-                '(linesofar >> 6) & 0x7F'
+                // Encode the -1/0/1 direction value into the first two bits of .z of a_data.
+                // Combine it with the lower 6 bits of `linesofar` (shifted by 2 bites to make
+                // room for the direction value). The upper 8 bits of `linesofar` are placed in
+                // the `w` component. `linesofar` is scaled down by `LINE_DISTANCE_SCALE` so that
+                // we can store longer distances while sacrificing precision.
+                '((dir === 0 ? 0 : (dir < 0 ? -1 : 1)) + 1) | (((linesofar * ' + LINE_DISTANCE_SCALE + ') & 0x3F) << 2)',
+                '(linesofar * ' + LINE_DISTANCE_SCALE + ') >> 6'
             ]
         }]
     }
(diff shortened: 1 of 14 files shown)


js/data/line_bucket.js
```diff
@@ -63,7 +68,7 @@ LineBucket.prototype.shaders = {
                 // The z component's first bit, as well as the sign bit is reserved for the direction,
                 // so we need to shift the linesofar.
                 '((dir < 0) ? -1 : 1) * ((dir ? 1 : 0) | ((linesofar << 1) & 0x7F))',
-                '(linesofar >> 6) & 0x7F'
+                '((linesofar >> 6) & 0xFF) - 128'
```
lucaswoj: Is this complex bit packing really necessary? If so, can you add a comment to the source about why it is necessary and how it works.

js/data/line_bucket.js
```diff
@@ -27,6 +27,11 @@ var EXTRUDE_SCALE = 63;
 var COS_HALF_SHARP_CORNER = Math.cos(75 / 2 * (Math.PI / 180));
 var SHARP_CORNER_OFFSET = 15;
 
+// The maximum distance along a line that can be stored in the buffers.
+var MAX_LINE_DISTANCE = Math.pow(2, 14);
+var LINE_DISTANCE_SCALE = 1 / 2;
```
lucaswoj: Please add a comment clarifying the unit of this value (even if it is "extent units")

shaders/linepattern.vertex.glsl
```diff
@@ -8,6 +8,10 @@ precision mediump float;
 // #define scale 63.0
 #define scale 0.015873016
 
+// We scale the distance before adding it to the buffers so that we can store
+// long distances for long segments. Use this value to unscale the distance.
```
lucaswoj: This comment would be useful in `LineBucket` too

js/data/line_bucket.js
```diff
@@ -175,8 +181,8 @@ LineBucket.prototype.addLine = function(vertices, join, cap, miterLimit, roundLi
             var prevSegmentLength = currentVertex.dist(prevVertex);
             if (prevSegmentLength > 2 * sharpCornerOffset) {
                 var newPrevVertex = currentVertex.sub(currentVertex.sub(prevVertex)._mult(sharpCornerOffset / prevSegmentLength)._round());
-                distance += newPrevVertex.dist(prevVertex);
-                this.addCurrentVertex(newPrevVertex, flip, distance, prevNormal.mult(1), 0, 0, false);
+                this.distance += newPrevVertex.dist(prevVertex) * LINE_DISTANCE_SCALE;
```
lucaswoj: Do we have to scale the distance here? Or can that be done [here](https://github.com/mapbox/mapbox-gl-js/pull/2114/files#diff-7ffec697472833439424ba5814cb0c00R71)? My intent was for all buffer packing craziness to go there.

js/data/line_bucket.js
```diff
@@ -377,6 +383,14 @@ LineBucket.prototype.addCurrentVertex = function(currentVertex, flip, distance,
     }
     this.e1 = this.e2;
     this.e2 = this.e3;
+
+    // There is a maximum "distance along the line" that we can store in the buffers.
+    // When we get close to the distance, reset it to zero and add the vertex again with
+    // a distance of zero.
```
lucaswoj: Please clarify that this maximum distance is determined by the number of bits in the `linesofar` attribute

lucaswoj:
> I don't like how `var distance` changed to `this.distance` but `addCurrentVertex` needs to be able to reset the distance and I'm not seeing a better way to do this.

I can live with this for now.

(Though I acutely feel the pain of this class's overcomplicated control flow and too-long methods)

ansis:
I added comments, moved LINE_DISTANCE_SCALE and added a couple more simplifications: 3b5bc9e 75ccf64

I closed this pr and opened a new one against master in https://github.com/mapbox/mapbox-gl-js/pull/2159

PR #2115, opened 2016-02-12
Swipe example

tristen:
swipe-example

cc @jfirebaugh for a review

### DIFF
--- a/docs/_posts/examples/3400-01-21-mapbox-gl-compare.html
+++ b/docs/_posts/examples/3400-01-21-mapbox-gl-compare.html
@@ -1,0 +1,49 @@
+---
+layout: example
+category: example
+title: Swipe between maps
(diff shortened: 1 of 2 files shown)


musicformellons:
Cool! An additional feature could be to have a moving daylight map!

jfirebaugh:
:+1: This is great! Should we consider making this into a plugin?

tristen:
> Should we consider making this into a plugin?

Hmm yeah! could be a super simple plugin: `map.addControl(new mapboxgl.Compare(mapA, mapB));`

tristen:
https://github.com/mapbox/mapbox-gl-js/commit/495002269a1b0d675f425ada2bc5a5a433090f30

pratikyadav:
I was just searching for this yesterday. Thanks  @tristen  :)

PR #2117, opened 2016-02-12
default to highp values in vertex shaders

lucaswoj:


ansis:
If a uniform is used in both the vertex and fragment shader it needs to have the same precision in both. The precision for these uniforms should be set to mediump in the vertex shader. I think this could be causing the errors.


I *think* highp is guaranteed for vertex shaders. We should check the spec and drop the `replace(/ highp /g, ' mediump ')`.

ansis:
afd74e7 fixes the shader compilation errors

Since vertex shaders always have highp we don't need. 5f7b700e53f2b8b4049fc94e3923cd11558a7bec See 4.5.2 in https://www.khronos.org/files/opengles_shading_language.pdf

399ab75 lowers precision where I think it's safe

ansis:
I think this should be ready for merging. @lucaswoj want to review my changes?

@kkaefer 
- does switching to highp as the default vertex precision sound good?
- do these precision qualifiers look good? https://github.com/mapbox/mapbox-gl-js/commit/399ab75d4542fbed15cc5dade8d43be1fd7484a8

lucaswoj:
LGTM.

:ship: once @kkaefer has a chance to weigh in

kkaefer:
I don't have any objections, but I also don't know what all of the implications are; if we see any rendering artifacts, we'd probably have to look at the qualifiers on a case-by-case basis. I looked through 399ab75 and all of the qualifiers you added make sense.

lucaswoj:
Did a quick sanity check on my iPhone. This branch looks :+1:

PR #2128, opened 2016-02-16
make panning while zooming more linear

### LINKED ISSUE #2070: Shift zooming is swoopy
zoomz

I'm noticing if I shift zoom to a location that isn't centered, it will zoom to the center and then shift over to the spot. This feels very weird. I'm on the latest Chrome.

Is this intentional or maybe a regression? I don't recall it feeling like this before.

cc @ansis @bhousel

ansis:
fix #2070

easeTo used to interpolate locations based on projected values. As zoom changed the new screen distance of these projected values changed either speeding up or slowing down the perceived panning.

With this change, the new center moves across the screen from it's previous point location to it's new point location linearly.

This also implements `panTo`, `rotateTo`, and `zoomTo` with `easeTo` to remove a lot of duplication. If you want I can split that out into a separate pr.

:eyes: @mourner @lucaswoj @bhousel

### DIFF
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -203,13 +203,6 @@ Transform.prototype = {
         this.center = this.coordinateLocation(coordCenter._sub(translate));
     },
 
-    setZoomAround: function(zoom, center) {
-        var p;
-        if (center) p = this.locationPoint(center);
-        this.zoom = zoom;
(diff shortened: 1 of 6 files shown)


bhousel:
:+1: love this.. The code is much clearer this way..

jfirebaugh:
:+1:

PR #1932, opened 2016-01-14
Add data driven styling for circle-color and circle-radius

lucaswoj:
This branch adds data-driven styling for `circle-color`.

screen shot 2016-04-06 at 11 55 07 am

```js
map.addLayer({
    "id": "random-points",
    "type": "circle",
    "source": "geojson-random-points",
    "paint": {
        "circle-radius": 5,
        "circle-opacity": 0.5,
        "circle-color": {
            stops: [[0, 'red'], [100, 'violet']],
            property: 'mapbox'
        }
    }
});

map.addLayer({
    "id": "random-points-outline",
    "ref": "random-points",
    "paint": {
        "circle-radius": 10,
        "circle-color": {
            stops: [[0, 'white'], [100, 'black']],
            property: 'mapbox'
        }
    }
}, 'random-points');
```

# Follow Up Tasks

 - [x] merge `mapbox-gl-function` branch
 - [x] merge `mapbox-gl-style-spec` branch
 - [x] merge `mapbox-gl-test-suite` branch 
 - [ ] add data-driven styling for all paint properties (except for fill layers, blocked on #1606)
 - [ ] refactor `Bucket` class into smaller units

PR #941, opened 2015-01-17
I rewrote most of style.js and it is awesome

jfirebaugh:
Introduce a StyleLayer class for the in-memory representation of a style layer.

It handles constant resolution, paint property cascading, and paint property
calculation. The paint functions now accept a layer as one of the parameters
and read all layer properties from it. This simplifies the rendering flow quite
a bit.

The concept of a "bucket" is now something that only the workers need to know
about. The main thread does not create buckets, it merely stores the buffers
and elementGroups on the tile when they are received from the worker.

There are a bunch of failing render tests but I'm now satisfied that this
approach is solid so getting this out for review.

PR #2140, opened 2016-02-18
Refactor Bucket class

### LINKED ISSUE #1585: Split up buffers
Port https://github.com/mapbox/mapbox-gl-native/pull/2103

We're currently reusing buffers across buckets until they fill up to 65535 vertices. Instead, every bucket should have its own buffer rather than sharing them.

cc @lucaswoj @jfirebaugh

### LINKED ISSUE #2004: Refactor Bucket class
In order to do data-driven styling the "right" way, we should refactor the `Bucket` class with the following objectives

 - remove extraneous state from `Bucket` and `SymbolBucket`
 - remove `Bucket`'s ownership of features
 - split up buffers https://github.com/mapbox/mapbox-gl-js/issues/1585
 - inline `ElementGroups` logic into `Bucket` and (maybe) remove `Bucket`'s ownership of element groups
 - refactor `resetBuffers` and `addFeatures` methods into `createBuffers` and `eachVertex` methods
 - look for opportunities to simplify `Bucket` by factoring out functionality
 - combine `data` and `source` folders
 - move `Buffer#EXTENT` to `Bucket#EXTENT` or elsewhere
 - move methods with a `gl` parameter from `Buffer` to `Bucket`

ref #1932

lucaswoj:
ref #1932 
fixes #2004 
fixes #1585

 - [x] transfer `Bucket`s from the worker to the main thread
 - [x] inline `ElementGroups` into `Bucket`
 - [x] move `Buffer#EXTENT` to `Bucket#EXTENT` or elsewhere
 - [x] rename `builder`s to `buffer`s in `buffer.test.js`
 - [x] split up buffers #1585


lucaswoj:
:eyes: @mourner @ansis @jfirebaugh?

ansis:
We used to have Buckets in the main thread. Then https://github.com/mapbox/mapbox-gl-js/pull/941 moved way from that:
> The concept of a "bucket" is now something that only the workers need to know
about. The main thread does not create buckets, it merely stores the buffers
and elementGroups on the tile when they are received from the worker.

What are the reasons for switching back to having full Buckets in the main thread?

lucaswoj:
> What are the reasons for switching back to having full Buckets in the main thread?

The `Bucket` class does a couple of things, some of which are necessary on the main thread (knowing which attributes are disabled, knowing how to calculate attribute values) and some of which are not necessary on the main thread (knowing how to iterate over features to build buffers). 

We should factor the `Bucket` class into several separate classes to isolate these distinct pieces of functionality. I think we can afford to do that refactor later.

ansis:
> We should factor the Bucket class into several separate classes to isolate these distinct pieces of functionality. I think we can afford to do that refactor later.

Sounds ok to me

I reviewed the changes and it looks good but we should get a second set of :eyes: on this before merging.

jfirebaugh:
>  We should factor the Bucket class into several separate classes to isolate these distinct pieces of functionality. I think we can afford to do that refactor later.

I probably won't have a chance to do a deep review here, but this sounds fine to me.

lucaswoj:
@mourner Would you care to be @ansis' second set of eyes on this one?

ansis:
> Would you care to be @ansis' second set of eyes on this one?

That was mostly about the high level changes which @jfirebaugh commented on. More review is always goo though

- how does 7c412647abb0851a74a93f4d42712f2c7fa0b552 impact memory usage? Does it create a couple new 8KB buffers for each bucket even if the bucket has few or no features? 
- does 7c412647abb0851a74a93f4d42712f2c7fa0b552  impact tile parsing time? I'd guess no, but it could be good to check.

lucaswoj:
> how does 7c41264 impact memory usage? Does it create a couple new 8KB buffers for each bucket even if the bucket has few or no features?

Yes, it does create 8KB buffers for each bucket. My understanding is that we take exactly this approach in mapbox-gl-native. Are there any other optimizations there? 

> does 7c41264 impact tile parsing time? I'd guess no, but it could be good to check.

I just ran the buffer benchmark and found a ~40% slowdown. We should investigate and try to ameliorate before merging.

lucaswoj:
Creating and transferring more buffers is more expensive. By tweaking some parameters, I was able to get the buffer benchmark time to within 15% of master. 

Another potential optimization is to use three* instances of `window.ArrayBuffer` as storage for many instances of `mapboxgl.Buffer`. This would be a medium-sized lift that could go into another PR. 

* one vertex buffer, one element buffer, and one secondElement buffer

lucaswoj:
I should note that, after doing parameter tuning, I came to an unintuitive optimal value for `Buffer.CAPACITY_RESIZE_MULTIPLE` of **`5`**. Does this strike anyone as a red flag? @mourner @ansis @jfirebaugh?

ansis:
> Another potential optimization is to use three[1] instances of window.ArrayBuffer as storage for many instances of mapboxgl.Buffer.

Sounds good to me. In general, I'd prefer if optimizations to avoid regressions are done before merging into master so that master never regresses.

> I should note that, after doing parameter tuning, I came to an unintuitive optimal value for Buffer.CAPACITY_RESIZE_MULTIPLE of 5. Does this strike anyone as a red flag?

What did you optimize for? allocation performance or memory usage? That doesn't seem surprising if all that matters is the allocation performance.

lucaswoj:
> Sounds good to me. In general, I'd prefer if optimizations to avoid regressions are done before merging into master so that master never regresses.

I prefer that too, especially given the recent perf regressions. 

However, the requirement that PRs do not regress _any_ performance metric by _any_ amount is pretty stiff, especially in the context of these kinds of refactors, which are necessary to ship data-driven styling. 

I will continue working on optimizing this PR today. As always, I appreciate extra eyes and thoughts on places where we could improve perf here.

> What did you optimize for? allocation performance or memory usage? That doesn't seem surprising if all that matters is the allocation performance.

I am optimizing for the "buffer benchmark" the [source of which is available here](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/buffer/buffer_benchmark.js).

I reckon we can get optimal memory usage with this allocation strategy by "shrinkwrapping" the `ArrayBuffer`s after they are populated

```js
arrayBuffer.slice(0, this.length * this.itemSize)
```

lucaswoj:
@ansis I added an `ArrayBuffer` size optimization in https://github.com/mapbox/mapbox-gl-js/commit/80c2acbebb5cf5692fc9dfe385e996d003e671ae which should make memory usage lower than it was before this refactor and improve overall performance.

js/data/buffer.js
```diff
@@ -158,12 +173,18 @@ Buffer.prototype.validate = function(args) {
     assert(argIndex === args.length);
 };
 
+Buffer.prototype.shrinkwrap = function() {
```
lucaswoj: On second thought, we could call this `Buffer#trim`

ansis:
@lucaswoj the buffer optimizations look great. Definitely fast enough for now! thanks

It looks like transfers back to the main thread are still slower:

bucket-2004:
```
parsed tile transfer (ms): 10
parsed tile transfer (ms): 16
parsed tile transfer (ms): 16
parsed tile transfer (ms): 23
parsed tile transfer (ms): 16
parsed tile transfer (ms): 15
```

master:
```
parsed tile transfer (ms): 2
parsed tile transfer (ms): 2
parsed tile transfer (ms): 3
parsed tile transfer (ms): 3
parsed tile transfer (ms): 2
parsed tile transfer (ms): 2
```

those times were produced by https://gist.github.com/ansis/572fb5caf1baff2452f7
(click anywhere on the map to reparse all tiles)

It looks like this is caused by:
- sending a serialized layer with each bucket
https://github.com/mapbox/mapbox-gl-js/commit/561c9b4d7ce9f5d53a5c05ada1123f42bef2d62f#diff-c1d5cefb566acef357121481a5a8e27eR173
- having many small buffers

lucaswoj:
With 5cf4b50, transfer times in the [benchmark](https://gist.github.com/ansis/572fb5caf1baff2452f7) posted above are between 6 and 12 seconds

ansis:
@lucaswoj Looks ready to me!

PR #2142, opened 2016-02-18
remove source update throttling

### LINKED ISSUE #1881: No continuous rendering during zooming with an enabled ImageSource
Aside from the fact that performance seems to degrade quite fast when using larger images with the ImageSource, it also seems that rendering if affected even with quite small/normal images.

When zooming out with a 200kB PNG overlayed using ImageSource I even noticed that there is no continuous rendering at all taking place and a sudden update comes when I stop zooming:

zoom-out-imagesource

This even happens regardless of whether the image is in the viewport or not, i.e. I can zoom in/out somewhere else on the map, but as long as the ImageSource is added this occurs. Is this expected?

When testing with larger images I experience a quite substantial performance impact on zooming/panning. Are there some expensive operations going on during rendering here?

### LINKED ISSUE #2139: sources are never updated while zooming
Sources are never updated while zooming. The set of tiles displayed for the current viewport is not updated until the map stops zooming. This makes tile loading (both from the cache and new tiles) really slow.

The function set with `setTimeout(fn, 50)` is not called after 50 ms. It is only called after the map stops zooming. Is this something we're doing somewhere else? Is the browser giving `requestAnimationFrame` priority over timeout functions?

The timeout exists at all because updating sources can be slow enough that frames are dropped. We should take another look at optimizing source updates so that they never drop frames.

console.log delays with:

```diff
diff --git a/js/ui/map.js b/js/ui/map.js
index e3a214f..ad4f413 100644
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -799,10 +799,21 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
             this.style._recalculate(this.transform.zoom);
         }
 
+        if (this._sourcesDirty) {
+            if (!this._time) {
+                this._time = Date.now();
+            }
+        }
+
         if (this.style && this._sourcesDirty && !this._sourcesDirtyTimeout) {
+            if (this._time) {
+                console.log('delay', Date.now() - this._time);
+                this._time = null;
+            }
             this._sourcesDirty = false;
             this._sourcesDirtyTimeout = setTimeout(function() {
                 this._sourcesDirtyTimeout = null;
+                this._rerender();
             }.bind(this), 50);
             this.style._updateSources(this.transform);
         }
```

ansis:
fix #2139 and fix #1881

Recent chrome scheduler changes aggressively de-prioritize `setTimeout` in some cases: https://code.google.com/p/chromium/issues/detail?id=567800

The unexpected setTimeout behavior was preventing tiles from being updated during zooming. Instead of fixing this, we should remove the timeout completely:

- The throttling was added when source updating was slow. Throttling increased the framerate by reducing the number of frames that went over their budget. **Source update has been heavily optimized since then and only takes 2% of the render time**.

- Throttling is not a good solution. Instead of preventing frames from exceeding their time budget by optimizing or distributing the work over a number of frames, throttling just reduces the number of frames that exceed their time.

:eyes: @jfirebaugh @lucaswoj @mourner

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -782,11 +782,8 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
             this.style._recalculate(this.transform.zoom);
         }
 
-        if (this.style && this._sourcesDirty && !this._sourcesDirtyTimeout) {
+        if (this.style && this._sourcesDirty) {
             this._sourcesDirty = false;
-            this._sourcesDirtyTimeout = setTimeout(function() {
-                this._sourcesDirtyTimeout = null;
-            }.bind(this), 50);
             this.style._updateSources(this.transform);
         }
 
@@ -826,7 +823,6 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     remove: function() {
         if (this._hash) this._hash.remove();
         browser.cancelFrame(this._frameId);
-        clearTimeout(this._sourcesDirtyTimeout);
         this.setStyle(null);
         if (typeof window !== 'undefined') {
             window.removeEventListener('resize', this._onWindowResize, false);


mourner:
That Chrome setTimeout deoptimization is surprising, but I'm :+1: on removing throttle here.

mourner:
Apparently this was also one of the causes of slowness in mapbox-gl-draw! https://github.com/mapbox/mapbox-gl-draw/issues/89

lucaswoj:
Deleting lines of code and making gl js faster. I love this PR!

PR #2146, opened 2016-02-18
Measure example using turf.distance

tristen:
distance-example

cc @jfirebaugh for review

### DIFF
--- a/docs/_posts/examples/3400-01-22-measure.html
+++ b/docs/_posts/examples/3400-01-22-measure.html
@@ -1,0 +1,158 @@
+---
+layout: example
+category: example
+title: Measure distances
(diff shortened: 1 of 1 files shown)


docs/_posts/examples/3400-01-22-measure.html
```diff
@@ -0,0 +1,170 @@
...
+                var id = features[0].properties.id;
+                geojson.features = geojson.features.filter(function(point) {
+                    return point.properties.id !== id;
+                });
+            } else {
+                var point = {
+                    "type": "Feature",
+                    "geometry": {
+                        "type": "Point",
+                        "coordinates": [
```
jfirebaugh: `"coordinates": e.lngLat`

jfirebaugh:
This should use `turf.line-distance` on the `LineString` feature, instead of implementing the equivalent manually.

tristen:
Hah - I missed that. Updated!

PR #2147, opened 2016-02-18
Cleanup event listener on popup close button

gmaclennan:
The event listener on the popup close button is not removed when content is changed or the popup is removed from the map.

### DIFF
--- a/js/ui/popup.js
+++ b/js/ui/popup.js
@@ -59,6 +59,10 @@ Popup.prototype = util.inherit(Evented, /** @lends Popup.prototype */{
      * @returns {Popup} `this`
      */
     remove: function() {
+        if (this._closeButton) {
+            this._closeButton.removeEventListener('click', this._onClickClose);
+        }
+
         if (this._content && this._content.parentNode) {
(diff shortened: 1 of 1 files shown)


jfirebaugh:
It looks like this will cause the close button to stop working if the popup is removed and then readded to the map.

Can you elaborate on why you think this is necessary?

gmaclennan:
Every time `setHTML()` is called, `_createContent()` is called, and a new DOM node is created for the close button (if `options.closeButton === true`). The `click` event attached to the previous DOM node is not removed, so my thought was that this would lead to a memory leak, since the event is never removed. Reviewing the code again, I don't think it should be removed when `remove()` is called, since it would stop working in this case.

jfirebaugh:
In general, unbinding event handlers isn't necessary to prevent memory leaks in any modern browser -- as long as there are no other outstanding references to the DOM node, both it and the bound callback functions will be garbage collected.

gmaclennan:
Good to know! Can close for now, and I can save a bunch of worrying about unbinding events in my own code!

PR #2150, opened 2016-02-18
Add an issue template. Fixes #2149

### LINKED ISSUE #2149: Add issue template
This repo would totally benefit from an [issue template!](https://github.com/blog/2111-issue-and-pull-request-templates). We should make sure users always provide

* Mapbox GL JS version
* Style used, if one of the stock styles
* JSFiddle or similar, if they can
* Browser version
* OS windows/mac/linux

tmcw:
cc @jfirebaugh for the review


ISSUE_TEMPLATE
```diff
@@ -0,0 +1,17 @@
+### Context
+
+* [ ] **mapbox-gl-js version**: 
+* [ ] **browser version**: 
+* [ ] **operating system**:
```
lucaswoj: We don't need checkboxes for this list. The completeness of each item is determined by whether or not the field is "filled in".

jfirebaugh:
:+1:

ISSUE_TEMPLATE
lucaswoj: Most bugs we encounter are not platform specific. I don't mind asking for browser / OS versions in a follow-up question if I can't reproduce. It is onerous to make users look this up without a strong need on our end.

ISSUE_TEMPLATE
```diff
@@ -0,0 +1,17 @@
...
+
+**Expected behavior**:
+
+**Actual behavior**:
+
+**Steps to reproduce the behavior**:
+
+1.
+2.
+3.
```
lucaswoj: About 90% of the bugs I've seen in this repo are most clearly described by a functioning minimal demo (i.e. in jsbin). I propose we keep this section more open-ended, with helpful guidelines:

1ec5: I’d pair this with a carrot:

> To help us diagnose your problem more quickly, please do as many of the following as you can:

ISSUE_TEMPLATE
```diff
@@ -0,0 +1,17 @@
+### Context
+
+* [ ] **mapbox-gl-js version**:
```
lucaswoj: Instead of asking for the mapbox-gl-js version they're using, can we insist that they use the latest version?

tmcw: No, I don't think we can ask that of all users. For any deployable software, a lot of people will be using an out-of-date version and won't be able to immediately update.

mourner: But if they use an out-of-date version, what's the point of reporting an issue in the first place? If an issue is already fixed on the master version, then a report will be immediately closed, and if it's not fixed, it would be nice to ask the user to make sure instead of checking it ourselves.

ISSUE_TEMPLATE
tmcw: These are great suggestions for the `CONTRIBUTING.md` topic, but they don't make sense within the context of an issue template.

mourner: Yeah, but in practice, _no one_ ever reads `CONTRIBUTING.md` :(

Issue #2049, opened 2016-02-03
Optimize SymbolInstance memory usage


ansis:
It looks like around 2/3 of all the memory retained in workers is related to SymbolInstance.

Compare by adding `bucket.symbolInstances = null` here: https://github.com/mapbox/mapbox-gl-js/blob/e745190838edf673151bc5f75869a0b0bb9c6fe8/js/source/worker_tile.js#L174
- check if SymbolInstance is retaining anything it shouldn't
- maybe use a Buffer to store SymbolQuads. All we do with them is add them to buffers so we could store them in a buffer and copy later on.
- maybe create SymbolQuads lazily. This would involve retaining the geometry, shapedText, shapedIcon so it might actually be worse. This could speed up initial tile loads though.

jfirebaugh:
@ansis @ChrisLoer Is this still relevant?

ChrisLoer:
I'd have to check to see what's actually happening, but I think this might be both still relevant and easier to address. I think right now `SymbolBucket#serialize` is doing a copy of the SymbolInstances, since they're not in TypedArray format. And once layout is done, there's no longer any reason for us to hold onto the SymbolInstances at all in the background.

ChrisLoer:
I looked closer, and I think @jfirebaugh fixed this in https://github.com/mapbox/mapbox-gl-js/commit/46b8566c0a7c216e9cef0d28d613571fae63a5a6#diff-2a5c03a2605a818d6f2d14697ce30e60 by making it so that `WorkerTile` no longer holds onto symbol buckets after `parse` returns.

PR #2151, opened 2016-02-18
set tile cache size based on viewport size

### LINKED ISSUE #2137: consider a larger parsed tile cache size
The default parsed tile cache size is 20: https://github.com/mapbox/mapbox-gl-js/blob/6122e19a92f85c71579fc446271451cbe72a5dbb/js/source/source.js#L27

Can we increase this without using way too much memory?
Can we reduce the memory footprint of a parsed tile so that we can store more of them?

ansis:
fix #2137

This makes makes the map feel faster when zooming/panning around the same area.

Larger viewports use more tiles and need a larger cache.

Using more memory for larger viewports is ok because larger viewports are often found on:
- more powerful machines with more available memory
- pages where the map is more important

After this change a 1280x728 map will cache 60 tiles instead of the previously hardcoded 20 tiles. This increases max memory usage about 30MB which seems worth it.

The cache size is calculated by multiplying the number of tiles in a single viewport by 10. Why 10? I tried a bunch of numbers and 10 seemed to prevent most reloading with regular panning and zooming without using too much memory.

:eyes: @mourner @lucaswoj @jfirebaugh

### DIFF
--- a/js/source/geojson_source.js
+++ b/js/source/geojson_source.js
@@ -69,7 +69,6 @@ function GeoJSONSource(options) {
         minzoom: this.minzoom,
         maxzoom: this.maxzoom,
         reparseOverscaled: true,
-        cacheSize: 20,
         load: this._loadTile.bind(this),
         abort: this._abortTile.bind(this),
         unload: this._unloadTile.bind(this),
--- a/js/source/source.js
+++ b/js/source/source.js
@@ -24,7 +24,6 @@ exports._loadTileJSON = function(options) {
 
         this._pyramid = new TilePyramid({
             tileSize: this.tileSize,
-            cacheSize: 20,
             minzoom: this.minzoom,
             maxzoom: this.maxzoom,
             roundZoom: this.roundZoom,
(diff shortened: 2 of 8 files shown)


mourner:
10 times the viewport tiles may be a little high. E.g. my Macbook has 12 tiles so it would have 120 and a much higher memory hit... Maybe decrease it a little bit? From 10 to something like 5-6?

Also, should we keep this configurable instead of hardcoding? Or maybe make the multiplier configurable?

ansis:
> 10 times the viewport tiles may be a little high. E.g. my Macbook has 12 tiles so it would have 120 and a much higher memory hit... Maybe decrease it a little bit? From 10 to something like 5-6?

I double checked and my math for calculating the number of tiles in a viewport was pretty optimistic. There can actually be a lot more tiles depending on how they are offset at edges. The right formula would be: `Math.ceil(width / 512 + 1) * Math.ceil(height / 512 + 1)`. When multiplied by `5` this formula gives me `60` and would give you `100` I think. Does that sound reasonable?

Parsed tiles are often around 500KB each so this would be 30-50MB.

> Also, should we keep this configurable instead of hardcoding? Or maybe make the multiplier configurable?

I think we should optimize our memory usage (https://github.com/mapbox/mapbox-gl-js/issues/2049, etc) first before exposing options to reduce usage at the cost of experience. I think we still have a bunch of ways we can reduce memory usage without reducing cache size.

ansis:
@mourner I accidentally merged this when merging https://github.com/mapbox/mapbox-gl-js/pull/2158 which I had based on this. Does the updated size calculation look ok to you or does this need further changes? https://github.com/mapbox/mapbox-gl-js/commit/3353d3926d69aa86efe3228cd50800f1e808384c

lucaswoj:
I think 30-50MB is reasonable. We can tune this ourselves going forward. Let's avoid making it configurable until that comes up.

eleijonmarck:
Is there any plan to make this configurable?

ansis:
@eleijonmarck I don't think there are any plans to make it configurable

PR #2152, opened 2016-02-19
fix tile render order for layer-by-layer

ansis:
Sort tiles by ascending zoom level instead of descending.

Child tiles had lower priority than parent tiles so child tiles weren't visible if their parent was also loaded. **This delayed newly loaded tiles from rendering while zooming in.**

Before layer-by-layer rendering clipping masks would be clipped by previously drawn tiles. Now clipping masks are drawn over other clipping masks. This means that clipping masks need to be drawn in the opposite order form before.

It's not possible to add a render test for this right now because we can't set up the tile state to reproduce this.

:eyes: @lucaswoj @jfirebaugh

### DIFF
--- a/js/source/tile_pyramid.js
+++ b/js/source/tile_pyramid.js
@@ -409,5 +409,5 @@ TilePyramid.prototype = {
 };
 
 function compareKeyZoom(a, b) {
-    return (b % 32) - (a % 32);
+    return (a % 32) - (b % 32);
 }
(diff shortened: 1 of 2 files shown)


jfirebaugh:
What about newly loaded tiles while zooming out? Did that work before and still works with this change?

ansis:
> What about newly loaded tiles while zooming out? Did that work before and still works with this change?

It sort of worked before and completely works after this change. It was less visible for zooming out because it's rarer to have child tiles that need to be rendered on top of parents.

At render time we want to always draw children above parents. TilePyramid.update is responsible for choosing the right parents and children.

PR #2153, opened 2016-02-19
Updated fitBounds to properly center on map

SoSDylan:
Updated fitBounds to properly center on map instead of focusing the bounds on the top left corner of the map.


mourner:
This doesn't look right. Can you demonstrate why the current behavior of fitBounds is incorrect, e.g. with a JSFiddle?

SoSDylan:
For some reason the behaviour has changed since I used the fitBounds function yesterday.

The points would be aligned left but now they are working correctly.

https://jsfiddle.net/SoSDylan/wm6kv2gh/12/

PR #2158, opened 2016-02-19
look for loaded parent tiles in cache

### LINKED ISSUE #1918: look for parent and child tiles in cache
https://github.com/mapbox/mapbox-gl-js/blob/57aa3e89a8086c3f3b091defd03d68e9e50c6e24/js/source/tile_pyramid.js#L239-L241

If a tile isn't loaded, it tries to cover that tile's area with other loaded tiles. However, it only looks for parent and child tiles in the set of tiles used for the previous frame. This works fairly well for zooming, but panning could be improved by checking the cache too.

ansis:
When the required tiles aren't loaded yet it tries to find parent tiles that area loaded that a cover the area. It used to just look for parents in the set of previously rendered tiles. There are many cases where the parents aren't in that set but they are in the cache.

This reduces the amount of the viewport not covered by tiles. It makes tile loading look a lot faster.

It helps a lot when you:
- zoom in and the pan around
- zoom in quickly and then zoom out more slowly

fixes #1918

---

This only loads parents, not children. I tried also loading children but it looked weird sometimes. Try the 
[load-children-from-cache](https://github.com/mapbox/mapbox-gl-js/tree/load-children-from-cache) branch to see what it's like.

If children are loaded from the cache you sometimes see flashes of high resolution tiles while zooming in. Seeing the area change from too-low-resolution to too-high-resolution to just-right-resolution is more jarring than just too-low-resolution to just-right-resolution.

:eyes: @mourner

### DIFF
--- a/js/source/tile_pyramid.js
+++ b/js/source/tile_pyramid.js
@@ -195,7 +195,12 @@ TilePyramid.prototype = {
             var tile = this._tiles[coord.id];
             if (tile && tile.loaded) {
                 retain[coord.id] = true;
-                return tile;
+                return true;
+            }
+            if (this._cache.has(coord.id)) {
+                this.addTile(coord);
+                retain[coord.id] = true;
+                return true;
             }
         }
     },


mourner:
:+1: for this PR, will look at the children branch later, but I'd guess we can leave it — loading parent tiles is much more important for smooth user experience.

ansis:
merged into master

PR #2166, opened 2016-02-21
bump feature filter, fix bug where 'in' filter drops features

samanpwbb:
See https://github.com/mapbox/feature-filter/pull/14

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "brfs": "^1.4.0",
     "csscolorparser": "^1.0.2",
     "envify": "^3.4.0",
-    "feature-filter": "^2.0.0",
+    "feature-filter": "^2.0.1",
     "geojson-vt": "^2.1.0",
     "gl-matrix": "^2.3.1",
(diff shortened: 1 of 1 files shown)


lucaswoj:
:shipit: when ready

mourner:
BTW the build should already include the fix, since ^2.0.0 range in package.json includes 2.0.1.

samanpwbb:
Feel free to close this if you want, I was not thinking about `^` when I submitted this PR. Still might be nice to force the upgrade though since this is a bad bug.

lucaswoj:
Developers of mapbox-gl-js won't get the upgraded package unless they run `npm update` (which I personally don't do regularly). If it fixes a serious bug, bumping the minor version of a dependency is prudent.

jfirebaugh:
:+1: to updating package.json. But feature-filter v2.0.1 _was_ in v0.14.2 right?

mourner:
Yes, it was. Although we need to quickly release 0.14.3 anyway because of a critical bugfix #2168.

PR #2168, opened 2016-02-22
fix pyramid.findLoadedParent for raster rendering

### LINKED ISSUE #2161: cannot export Canvas to image using satellite-hybrid-v8
It appears that when mapbox://styles/mapbox/satellite-hybrid-v8 , map.getCanvas().toDataURL("image/png") fails silently.

This may be intentional due to copyright issues or similar... ?  If so, is it documented?  Also, failing silently may not be ideal.

### LINKED ISSUE #2167: 0.14.2 regression, getOpacities() `Uncaught TypeError: Cannot read property 'z' of undefined`
Getting `Uncaught TypeError: Cannot read property 'z' of undefined` when zooming on this example: https://www.mapbox.com/mapbox-gl-js/example/map-tiles/

Doesn't happen in 0.14.1.

My attempt at debugging led me to this commit that returns true instead of a tile.
https://github.com/mapbox/mapbox-gl-js/commit/e6cb15c3f18c73cfc7aea2a797d3a33eedb946cd

Going to dig some more but wanted to report my findings.

ansis:
Fixes a sloppy bug introduced by https://github.com/mapbox/mapbox-gl-js/commit/e6cb15c3f18c73cfc7aea2a797d3a33eedb946cd

I didn't notice draw_raster was using `findLoadedParent` and I didn't test with a raster style.

fix #2167
fix #2161

:eyes: @mourner @lucaswoj

### DIFF
--- a/test/js/source/tile_pyramid.test.js
+++ b/test/js/source/tile_pyramid.test.js
@@ -648,3 +648,60 @@ test('TilePyramid#orderedIDs (ascending order by zoom level)', function(t) {
     ]);
     t.end();
 });
+
+
+test('TilePyramid#findLoadedParent', function(t) {
+
+    t.test('adds from previously used tiles (pyramid._tiles)', function(t) {
+        var pyramid = createPyramid({});
+        var tr = new Transform();
+        tr.width = 512;
+        tr.height = 512;
+        pyramid.updateCacheSize(tr);
+
+        var tile = {
+            coord: new TileCoord(1, 0, 0),
+            loaded: true
+        };
+
+        pyramid._tiles[tile.coord.id] = tile;
+
+        var retain = {};
+        var expectedRetain = {};
+        expectedRetain[tile.coord.id] = true;
+
+        t.equal(pyramid.findLoadedParent(new TileCoord(2, 3, 3), 0, retain), undefined);
+        t.deepEqual(pyramid.findLoadedParent(new TileCoord(2, 0, 0), 0, retain), tile);
+        t.deepEqual(retain, expectedRetain);
+        t.end();
+    });
+
+
+    t.test('adds from cache', function(t) {
+        t.end();
+        var pyramid = createPyramid({});
+        var tr = new Transform();
+        tr.width = 512;
+        tr.height = 512;
+        pyramid.updateCacheSize(tr);
+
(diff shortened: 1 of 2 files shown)


test/js/source/tile_pyramid.test.js
```diff
@@ -648,3 +648,31 @@ test('TilePyramid#orderedIDs (ascending order by zoom level)', function(t) {
...
+    t.equal(pyramid.findLoadedParent(new TileCoord(2, 3, 3), 0, {}), undefined);
+
+    // add from cache
+    t.deepEqual(pyramid.findLoadedParent(new TileCoord(2, 0, 0), 0, {}), tile);
+
+    // cache is empty
+    t.equal(pyramid._cache.order.length, 0);
+
+    // add from retained tiles
+    t.deepEqual(pyramid.findLoadedParent(new TileCoord(2, 0, 0), 0, {}), tile);
```
lucaswoj: What is this testing? It looks the same as line 669 above.

test/js/source/tile_pyramid.test.js
```diff
@@ -648,3 +648,31 @@ test('TilePyramid#orderedIDs (ascending order by zoom level)', function(t) {
...
+
+    t.end();
+});
```
lucaswoj: Thanks for adding unit tests to this method! Can we break this up into a couple separate tests? It looks like you're checking a couple isolate-able bits of functionality.

test/js/source/tile_pyramid.test.js
ansis: `findLoadedParent` looks for load parent tiles in both the cache and `._tiles` (the previous set of tiles).
669 tests whether it works when the tile is in the cache
675 tests whether it works when the tile is in `._tiles`
669 moves the tile from the cache to ._tiles internally which is what 672 confirms.

I'll split it up as suggested by https://github.com/mapbox/mapbox-gl-js/pull/2168/files#r53679813

test/js/source/tile_pyramid.test.js
ansis: done in 1a5f9b0

lucaswoj:
Looks good now! :shipit: on green.

Issue #1992, opened 2016-01-25
Add test coverage metrics


lucaswoj:
... I don't know why we haven't had these before now.

PR #2174, opened 2016-02-23
Blazing fast layer update

### LINKED ISSUE #2082: Highlight features example has become slow
I recall https://www.mapbox.com/mapbox-gl-js/example/hover-styles/ being quite snappy. In Chrome and Firefox it is now very laggy and CPU intensive. It seems this and other examples involving filters spend a lot of CPU-time on <code>Actor.postMessage</code>.

mourner:
Makes `setLayoutProperty`, `setFilter` and `setLayerZoomRange` much faster by limiting the style->worker transfer to just the updated layers instead of the whole style.

Closes #2082 and makes this hover example https://jsfiddle.net/0Lk8awwp/1/ fast and smooth again: https://jsfiddle.net/Mourner/0Lk8awwp/3/

cc @lyzidiamond @lucaswoj @scothis

### DIFF
--- a/js/style/style_batch.js
+++ b/js/style/style_batch.js
@@ -107,8 +107,8 @@ styleBatch.prototype = {
         return this;
     },
 
-    setPaintProperty: function(layer, name, value, klass) {
-        this._style.getLayer(layer).setPaintProperty(name, value, klass);
+    setPaintProperty: function(layerId, name, value, klass) {
+        this._style.getLayer(layerId).setPaintProperty(name, value, klass);
         this._change = true;
 
         return this;
@@ -116,6 +116,8 @@ styleBatch.prototype = {
 
     setLayoutProperty: function(layerId, name, value) {
         var layer = this._style.getReferentLayer(layerId);
+        if (layer.getLayoutProperty(name) === value) return this;
+
         layer.setLayoutProperty(name, value);
 
         this._broadcastLayers[layerId] = true;
@@ -136,6 +138,7 @@ styleBatch.prototype = {
         }))) return this;
 
         var layer = this._style.getReferentLayer(layerId);
+        if (filtersEqual(layer.filter, filter)) return this;
         layer.filter = filter;
 
         this._broadcastLayers[layerId] = true;
@@ -149,6 +152,8 @@ styleBatch.prototype = {
 
     setLayerZoomRange: function(layerId, minzoom, maxzoom) {
         var layer = this._style.getReferentLayer(layerId);
+        if (layer.minzoom === minzoom && layer.maxzoom === maxzoom) return this;
+
         if (minzoom != null) {
             layer.minzoom = minzoom;
         }
@@ -224,5 +229,16 @@ styleBatch.prototype = {
         return this;
     }
 };
+
+function filtersEqual(a, b) {
+    if (Array.isArray(a)) {
+        if (!Array.isArray(b) || a.length !== b.length) return false;
+        for (var i = 0; i < a.length; i++) {
+            if (!filtersEqual(a[i], b[i])) return false;
+        }
+        return true;
+    }
+    return a === b;
+}
 
 module.exports = styleBatch;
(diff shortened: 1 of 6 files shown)


tmcw:
Awesome, could you do a jsfiddle that includes this patch so we can compare?

mourner:
@tmcw yes, was waiting for the build to complete.

mourner:
JSFiddle with the new build (inlined) that demonstrates the performance: https://jsfiddle.net/Mourner/0Lk8awwp/2/

mourner:
Added another optimization that affects the example — avoid triggering an update when you set filter, layout property or zoom range to values that are already set.

js/style/style_batch.js
```diff
@@ -14,7 +14,8 @@ function styleBatch(style, work) {
 
     batch._style = style;
     batch._groupLayers = false;
-    batch._broadcastLayers = false;
+    batch._broadcastAllLayers = false;
+    batch._broadcastLayers = {};
```
lucaswoj: It's a little tricky that `this._broadcastLayers` and `this._style._broadcastLayers` both exist and are of different types

js/style/style_batch.js
```diff
@@ -217,4 +230,15 @@ styleBatch.prototype = {
     }
 };
 
+function filtersEqual(a, b) {
```
lucaswoj: This looks like an `isDeepEqual` method that works only on arrays. Can we make this a general purpose `util` method instead?

test/js/style/style.test.js
```diff
@@ -162,6 +162,34 @@ test('Style#_broadcastLayers', function(t) {
...
+    style.on('error', function(error) { t.error(error); });
+
+    style.on('load', function() {
+        style.dispatcher.broadcast = function(key, value) {
+            t.equal(key, 'update layers');
+            t.deepEqual(value.map(function(layer) { return layer.id; }), ['second', 'third']);
+            t.end();
+        };
+
+        style._broadcastLayers(['second', 'third']);
```
lucaswoj: Rather than testing the private method `_broadcastLayers` we should apply some style operations (like `setLayoutProperty`) that  trigger `_broadcastLayers`. This will increase our test coverage #1992 and make this test more robust against private method refactors.

mourner: Agree with your points Lucas. My intent here was to introduce the minimum
amount of code changes to fix a severe regression and get it into a release
ASAP — most of the things you mentioned were there before the PR. Should we
merge now but follow up with another refactoring PR?

вторник, 23 февраля 2016 г. пользователь Lucas Wojciechowski написал:

> In test/js/style/style.test.js
> <https://github.com/mapbox/mapbox-gl-js/pull/2174#discussion_r53835329>:
>
> > +            {id: 'first', source: 'source', type: 'fill', 'source-layer': 'source-layer'},
> > +            {id: 'second', source: 'source', type: 'fill', 'source-layer': 'source-layer'},
> > +            {id: 'third', source: 'source', type: 'fill', 'source-layer': 'source-layer'}
> > +        ]
> > +    });
> > +
> > +    style.on('error', function(error) { t.error(error); });
> > +
> > +    style.on('load', function() {
> > +        style.dispatcher.broadcast = function(key, value) {
> > +            t.equal(key, 'update layers');
> > +            t.deepEqual(value.map(function(layer) { return layer.id; }), ['second', 'third']);
> > +            t.end();
> > +        };
> > +
> > +        style._broadcastLayers(['second', 'third']);
>
> Rather than testing the private method _broadcastLayers we should apply
> some style operations (like setLayoutProperty) that trigger
> _broadcastLayers. This will increase our test coverage #1992
> <https://github.com/mapbox/mapbox-gl-js/issues/1992> and make this test
> more robust against private method refactors.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mapbox/mapbox-gl-js/pull/2174/files#r53835329>.
>


-- 
Vladimir Agafonkin
http://agafonkin.com/en
+380 (93) 745 44 61

lucaswoj:
I disagree that this :fire: is so bad that you don't have time to

 - rename one property
 - move one function to another module
 - refactor a unit test to use public rather than private methods

If you have too much on your plate, I can take care of these tweaks.

lucaswoj:
This looks great! :shipit:

PR #1985, opened 2016-01-25
Handle 'Not Found' errors to allow 'loaded' event

laurenancona:
We've been running our own fork with this patch since `0.11.0` because the errors prevent `loaded` from returning `true`. Still results in noisy logs, but find it preferable to silent failure while debugging. 

Closes #1800.

Issue #1800, opened 2015-12-04
Suppress error messages for missing tiles


lucaswoj:
There is log spew about missing tiles, though it is valid for a tile endpoint to return a 404 to express "no data". We should suppress these error messages.

ref #1549

rbonillajr:
This error causes me other component failures

for examples this add more and more delete button

![image](https://cloud.githubusercontent.com/assets/8268226/11639851/4725cf92-9cfd-11e5-969e-331402475c92.png)

gertcuykens:
Or maybe add the ability to import stuff in the original tile set street-v6 in studio so you don't have to make a incomplete data set with no tiles in the first place?

Ps also note that you can not disable does 404's in chrome console, so worker debug messages end up in between does 404.

mourner:
If we supress all 404 errors, how do we handle situations with e.g. a typo in the url leading to all tiles being 404? That would go unnoticeable.

jfirebaugh:
404s will still be visible in the network page of developer tools (and for unexpected errors like bad URLs, visible as missing tiles in the map). This ticket is about suppressing the _extra_ logging to the JS console that is happening right now. I don't think that there's a way to programmatically distinguish "expected" and "unexpected" 404s, so I think we'll accept that "unexpected" 404s only get logged to the normal places.

jfirebaugh:
Also, the long-term fix here is to encourage using HTTP 204 "No Content" response in tile APIs to indicate lack of a tile at the requested coordinates, and reserving 404 for real unexpected errors. I've ticketed this internally for future versions of the Mapbox tile API.

lucaswoj:
We will address this issue through using HTTP 204 "No Content" codes and implementing #1775

jfirebaugh:
It's going to be awhile before we can rely on 204, and this comes up often with @mapbox/support. Let's add the suppression.

Scarysize:
What would be the correct way to suppress the error? Not calling the callback with the `err` object here (?): 

https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/raster_tile_source.js#L61

redbmk:
[...]

PR #2176, opened 2016-02-23
Make "Map#load" fire, even if "Not Found" errors were encountered

lucaswoj:
ref #1985 
ref #1800 

If a vector tile request gets a 404 response, the `Map#load` event doesn't fire. 

#1800 is about a feature request that is different than this bug but the two have been conflated.

cc @jfirebaugh @ansis @geografa @laurenancona

### DIFF
--- a/js/source/vector_tile_source.js
+++ b/js/source/vector_tile_source.js
@@ -83,6 +83,7 @@ VectorTileSource.prototype = util.inherit(Evented, {
             return;
 
         if (err) {
+            tile.errored = true;
             this.fire('tile.error', {tile: tile, error: err});
             return;
         }


ansis:
looks good to me

superphly:
So by setting tile.errored = true;, that allows the map#load to fire?

lucaswoj:
@superphly Correct. To understand why, take a look at the `TilePyramid#loaded` method https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/tile_pyramid.js#L49

laurenancona:
\o/



Thanks, @lucaswoj

PR #2181, opened 2016-02-24
Rename "Bucket#shaders" to "Bucket#shaderInterfaces"

### LINKED ISSUE #2172: name of `shader` property of buckets
I'm not sure `shader` is the right name for [shader property of buckets](https://github.com/mapbox/mapbox-gl-js/blob/1ad40725cf925c61db2a6a80d18ff50144e85067/js/data/symbol_bucket.js#L83-L95). The buffers aren't necessarily 1-to-1 with a shader. `icon` could be rendered with either the sdf shader or the bitmap icon shader. Lines currently have one `shader` type in the bucket but three different shaders.

@lucaswoj

lucaswoj:
ref #2172 
 
cc @ansis

### DIFF
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -16,10 +16,10 @@ module.exports = Bucket;
  */
 Bucket.create = function(options) {
     var Classes = {
-        fill: require('./fill_bucket'),
-        line: require('./line_bucket'),
-        circle: require('./circle_bucket'),
-        symbol: require('./symbol_bucket')
+        fill: require('./bucket/fill_bucket'),
+        line: require('./bucket/line_bucket'),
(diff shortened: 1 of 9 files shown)


ansis:
looks good to me


In https://github.com/mapbox/mapbox-gl-js/issues/2172#issuecomment-187433232 I mentioned that shader interfaces and buffer instances should probably not be 1-to-1. Should I open a separate issue for that?

lucaswoj:
@ansis I answered in https://github.com/mapbox/mapbox-gl-js/issues/2172#issuecomment-188424487

lucaswoj:
Merged this as-is, we can leave #2172 open until all open questions are resolved.

PR #2183, opened 2016-02-24
Add contextual "key"s to validateStyle calls

### LINKED ISSUE #2170: Add descriptive "key"s to style validation calls
When we call style validation methods, like we do [here](https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style_layer.js#L75), we should pass a `key` parameter that provides context for the error message (layer name, source name, property name, etc).

lucaswoj:
These keys are contextual clues for users to help trace the errors. Before this PR, a style validation error might appear as

```
Error: undefined: number expected, boolean found
```

after this PR, the same error would appear as

```
Error: layers.roads.paint.line-width: number expected, boolean found
```

fixes #2170 

cc @ansis


ansis:
:+1:

PR #2184, opened 2016-02-25
Update coordinates of image/video sources

### LINKED ISSUE #1880: Update coordinates for ImageSource (and possibly VideoSource)
In a use case I have I need to adjust the coordinates of an ImageSource, but the current API only permits providing the coordinates when adding the source. I've managed to get around this by removing and adding the source when the coordinates change, but as you can imagine this is quite inefficient and causes considerable flicker when many changes occur. I imagine this is related to both the fact that removing and adding the source comes with some overhead, but also that the ImageSource performs an AJAX operation to fetch the image every time..

Would it be a huge effort to extend the ImageSource API to allow for updating the coordinates?

averas:
This fixes #1880 by adding support for updating the coordinates on image- and video sources.

While doing this I noticed this part which I don't really understand how it's supposed to work:

```
    onAdd: function(map) {
        this.map = map;
        if (this.image) {
            this.setCoordinates();
        }
    }
```

Won't a call without arguments to setCoordinates (perviously createTile) cause an error on the subsequent coordinates.map() in that function?

### DIFF
--- a/js/source/image_source.js
+++ b/js/source/image_source.js
@@ -16,7 +16,6 @@ module.exports = ImageSource;
  * @param {Object} [options]
  * @param {string} options.url A string URL of an image file
  * @param {Array} options.coordinates Four geographical [lng, lat] coordinates in clockwise order defining the corners (starting with top left) of the image. Does not have to be a rectangle.
- * starting at the top left: tl, tr, br, bl
  * @example
  * var sourceObj = new mapboxgl.ImageSource({
  *    url: 'https://www.mapbox.com/images/foo.png',
--- a/js/source/image_source.js
+++ b/js/source/image_source.js
@@ -68,13 +68,13 @@ ImageSource.prototype = util.inherit(Evented, {
      */
     setCoordinates: function(coordinates) {
         this.coordinates = coordinates;
-        
+
         /**
          * Calculate which mercator tile is suitable for rendering the image in
          * and create a buffer with the corner coordinates. These coordinates
          * may be outside the tile, because raster tiles aren't clipped when rendering.
          */
-                
+
         var map = this.map;
         var cornerZ0Coords = coordinates.map(function(coord) {
             return map.transform.locationCoordinate(LngLat.convert(coord)).zoomTo(0);
@@ -104,9 +104,9 @@ ImageSource.prototype = util.inherit(Evented, {
         this.tile.boundsBuffer = gl.createBuffer();
         gl.bindBuffer(gl.ARRAY_BUFFER, this.tile.boundsBuffer);
         gl.bufferData(gl.ARRAY_BUFFER, array, gl.STATIC_DRAW);
-        
+
         this.fire('change');
-        
+
         return this;
     },
 
--- a/js/source/image_source.js
+++ b/js/source/image_source.js
@@ -47,8 +47,7 @@ function ImageSource(options) {
         this._loaded = true;
 
         if (this.map) {
-            this.createTile(options.coordinates);
-            this.fire('change');
+            this.setCoordinates(options.coordinates);
         }
     }.bind(this));
 }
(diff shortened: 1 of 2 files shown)


js/source/image_source.js
```diff
@@ -57,19 +56,27 @@ ImageSource.prototype = util.inherit(Evented, {
     onAdd: function(map) {
         this.map = map;
         if (this.image) {
-            this.createTile();
+            this.setCoordinates();
         }
```
lucaswoj: This method gets fired when the `ImageSource` is added to an instance of `Map`. There are two cases to consider:

 - the HTTP request for the image finishes _before_ the `ImageSource` gets added to the map
 - the HTTP request for the image returns _after_ the `ImageSource` gets added to the map

The code in this method deals with the first case.

I think the lack of an argument to `setCoordinates` is a mistake. The line should be

```js
this.setCoordinates(this.coordinates)
```

js/source/image_source.js
```diff
@@ -57,19 +56,27 @@ ImageSource.prototype = util.inherit(Evented, {
...
      */
-    createTile: function(cornerGeoCoords) {
+    setCoordinates: function(coordinates) {
+        this.coordinates = coordinates;
+
+        /**
+         * Calculate which mercator tile is suitable for rendering the image in
+         * and create a buffer with the corner coordinates. These coordinates
+         * may be outside the tile, because raster tiles aren't clipped when rendering.
+         */
```
lucaswoj: This should be a `//` style comment. `/**` style comments are reserved for semantic documentation blocks.

js/source/image_source.js
```diff
@@ -57,19 +56,27 @@ ImageSource.prototype = util.inherit(Evented, {
...
     },
 
     /**
-     * Calculate which mercator tile is suitable for rendering the image in
-     * and create a buffer with the corner coordinates. These coordinates
-     * may be outside the tile, because raster tiles aren't clipped when rendering.
-     * @private
+     * Update image coordinates and rerender map
+     *
+     * @param {Array} coordinates lng, lat coordinates in order clockwise
```
lucaswoj: You should be a little more specific in your type declaration and description of this parameter. What do these coordinates represent? How many coordinates are expected? 

@jfirebaugh @ansis If we're going to make this method public, should we change the argument signature? We could

 - accept an array of 4 corner coordinates
 - accept an array of 2 corner coordinates
 - accept 4 corner coordinates as separate args
 - accept 2 corner coordinates as seperate args
 - accept an instance of `LngLatBounds`
 - (some combination of the above)

js/source/video_source.js
```diff
@@ -82,18 +81,27 @@ VideoSource.prototype = util.inherit(Evented, /** @lends VideoSource.prototype *
         this.map = map;
         if (this.video) {
             this.video.play();
-            this.createTile();
+            this.setCoordinates();
```
lucaswoj: As above, I believe we should pass `this.coordinates` as an argument

lucaswoj:
Thank you for the PR @averas! Looking forward to the next round of revisions, feedback from some other Mapbox-ers, and then shipping this!

js/source/image_source.js
averas: Oh, I am completely with you on why the code exists, it was the intentional or unintentional lack of arguments that got me confused.

js/source/image_source.js
averas: np. will fix.

js/source/image_source.js
averas: Hard to argue against! I honestly only copied the argument descriptor from the constructor which constituted the existing public API so to speak. So one could argue that this should already have been present on the constructor.. but I will add it to both!

lucaswoj: Ah! My apologies, I didn't notice this was the same as the documentation for the constructor args. In that case, we don't need to change the signature, but we should improve the docs slightly while we're mucking around in here.

ansis:
Looks good to me!

Is there a better name than `coordinates` to use here? I realize the choice to use `coordinates` happened before this pull request but if we want to rename it now would be a good time.

`corners`, `geometry`?

lucaswoj:
I like `coordinates`. 

Any last words before we :ship: @jfirebaugh @mourner?

mourner:
:shipit:

jfirebaugh:
:+1:

lucaswoj:
Merged in e7b551ccc8c36cf2031684398b7ae7b28fd9fa0d :rocket: 

Thanks @averas!

PR #2190, opened 2016-02-25
Streamline our issue template

lucaswoj:
Our new issue template has helped our users write better bug reports and feature requests. 

This PR aims to streamline our issue template to increase the signal to noise ratio.

I'm comfortable leaving this PR open for a while and going through many revisions so that everyone can weigh in.

cc @ansis @jfirebaugh @tmcw @mourner @lyzidiamond @danswick

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -2,8 +2,9 @@
 
 To help us diagnose your problem quickly, please
 
- - include a link to a minimal demonstration the bug (we recommend using https://jsbin.com)
+ - include a link to a minimal demonstration the bug, we recommend using https://jsbin.com
  - ensure you can reproduce the bug using the latest release
+ - only post to report a bug or request a feature, direct all support requests to https://mapbox.com/contact
 
 -->
 
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -7,6 +7,12 @@ To help us diagnose your problem quickly, please
 
 -->
 
+### Steps to Trigger Behavior
+
+ 1.
+ 2.
+ 3.
+
 ### Expected Behavior
 
 ### Actual Behavior
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,10 +1,10 @@
 <!--
 
-To help us diagnose your problem quickly, please
+To help us diagnose your problem quickly, please:
 
- - include a link to a minimal demonstration the bug, we recommend using https://jsbin.com
+ - include a link to a minimal demonstration of the bug (we recommend using https://jsbin.com)
  - ensure you can reproduce the bug using the latest release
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -1,23 +1,12 @@
-### Context
+<!--
 
-* **mapbox-gl-js version**: 
+To help us diagnose your problem quickly, please
 
-### Report
+ - include a link to a minimal demonstration the bug (we recommend using https://jsbin.com)
```
lucaswoj: I recommended jsbin over jsfiddle. 

jsfiddle is opinionated about dividing up css, javascript, and html into separate inputs which is a pain when copy-pasting examples and esiting websites.

jfirebaugh:
I would like to keep an explicit request for mapbox-gl-js version number. That's what I find myself most frequently asking for. Otherwise I'm fine with this version.

lucaswoj:
@jfirebaugh Why we would ask for a version number, other than to ensure it is the latest version? What is our course of action if a user reports a bug that is not present in the latest version?

ISSUE_TEMPLATE.md
```diff
@@ -1,23 +1,12 @@
...
+ - ensure you can reproduce the bug using the latest release
 
-**Expected behavior**:
+-->
 
-**Actual behavior**:
+### Expected Behavior
 
-**Steps to reproduce the behavior**:
```
tmcw: Can we cut the numbered list but say like "Details to help us trigger the bug"

jfirebaugh:
> Why we would ask for a version number, other than to ensure it is the latest version?

Ensuring it is the latest version, or at least a specific version or later, is reason enough to ask in and of itself. We often get reports of symptoms that are consistent with a bug that was fixed in a previous version. If the reporter says they're using a version that had that bug, then we ask them to upgrade. If they say they're already using the latest version, then we need to investigate more deeply.

We're going to get reports from people not using the latest version (even if we request that they try it), and if we don't ask for the version, we're going to get reports from people using the latest, but not explicitly saying so. Asking for the version will let us distinguish those cases straight off the bat much more frequently.

mourner:
Another reason is that when someone reports a bug and hangs for a while (and meanwhile a few new versions are released), we will know for sure what version was affected when we get back to fixing it, rather than the vague "latest", whatever it was at the time.

lucaswoj:
Good points. I added a version field.

ISSUE_TEMPLATE.md
lucaswoj: Added this back in the latest release

ansis:
looks good to me!

lyzidiamond:
I'm gonna give this some grammar :eyes: real quick. Stand by.

lyzidiamond:
Made a few grammar edits. This is lovely.

mourner:
:+1:

PR #2200, opened 2016-02-28
Added fs as dependency

DenisCarriere:
Added `fs` as dependency as it's required by `render/shaders.js`.

https://github.com/mapbox/mapbox-gl-js/blob/master/js/render/shaders.js#L3

Error was caused when building with Webpack.

> ERROR in ./~/mapbox-gl/js/render/shaders.js
Module not found: Error: Cannot resolve module 'fs' in /home/denis/test/node_modules/mapbox-gl/js/render
 @ ./~/mapbox-gl/js/render/shaders.js 3:9-22

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "unassertify": "^2.0.0",
     "unitbezier": "^0.0.0",
     "vector-tile": "^1.2.0",
-    "webworkify": "^1.0.2"
+    "webworkify": "^1.0.2",
+    "fs": "0.0.2"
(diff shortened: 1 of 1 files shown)


DenisCarriere:
Brain fart... `fs` is a standard module, I'll investigate more my webpack build.

DenisCarriere:
It was in fact an error in my Webpack config.

Added this and it worked:
```js
node: {
  fs: "empty"
}
```

Issue #1834, opened 2015-12-14
Ensure SourceCache reuses tiles across the dateline


buchanae:
This started when I was animating a flight path across the Pacific, and setting the map center to follow the flight. The flight path was generated using arc.js. I would get a frame or two where the map would turn black (satellite style background), so I started investigating.

I realized that the blackout happened as the path crossed the line from positive longitude to negative longitude (anti-meridian), i.e. as the flight passed that line, the center of the map goes from 179 longitude to -180 longitude.

Transforming the longitude into a positive value (i.e. -185 --> 175) fixed the blackout, and I think it's a slight improvement to tile rendering performance.

I tried to demonstrate that here: http://jsfiddle.net/wodvrjj0/3/

The transformation function I used is:

```
if (lng < 0) {
  lng = 360 + (lng % -360);
}
```

During my debugging, I modified Transform#center to do this fix and things in Mapbox still seemed to work ok under light testing. I have no idea if that's the correct place for a fix, I don't know the code well enough to decide. TileCoord#cover is another possible place for a fix; again, I don't know.

lucaswoj:
I have identified the underlying cause of this issue. We aspire to use the same tiles when rendering the same region across the dateline but fail to do so in some cases. This test currently fails but should pass.

```js
    t.test('reuses wrapped tile (when first tile was w != 0)', function(t) {
        var load = 0;
        var add = 0;

        var pyramid = createPyramid({
            load: function(tile) { tile.loaded = true; load++; },
            add: function() { add++; }
        });

        var t1 = pyramid.addTile(new TileCoord(0, 0, 0, 1));
        var t2 = pyramid.addTile(new TileCoord(0, 0, 0, 0));

        t.equal(load, 1);
        t.equal(add, 2);
        t.equal(t1, t2);

        t.end();
    });
```

lucaswoj:
Hm. While it is a real bug, fixing the tile retention problem doesn't seem to fix the flickering. 

Commenting out the body of `TilePyramid#removeTile` does fix the flickering.

mourner:
@lucaswoj did you figure it out eventually?

lucaswoj:
@mourner I fixed something but the bug persists. @kelvinabrokwa is looking at this bug this week.

lucaswoj:
@kelvinabrokwa The "something" I fixed is in this branch https://github.com/mapbox/mapbox-gl-js/tree/pyramid-1834

mourner:
[...]

PR #2211, opened 2016-03-01
Fix raster tiles flicker & hang

### LINKED ISSUE #2201: Styles with raster tiles make the browser unresponsive
**mapbox-gl-js version**: 0.14.3

### Steps to Trigger Behavior

 1. load mapbox://styles/mapbox/satellite-v8
 2. zoom and pan like a maniac

### Expected Behavior

The map remains responsive.

### Actual Behavior

Chrome will lockup without an error on the console. The tab will consume 100% and can only be terminated by `kill -9` the process. I have not tested in other browsers.

### LINKED ISSUE #2208: Flickering of background layer when zooming
### Problem
Severe flickering of background raster when zooming in and out. Seems to be related to which layers are available higher up in the tilepyramid. Seems it's going higher up in the pyramid than necessary, exchanging high-res tiles for much lower-res tiles, even when high-res are available.

Looks like an optimization for smoother zoom, but with the tradeoff of showing low-res tiles in a flickering way.

### jsbin
Zoom in and out to see flickering:
http://jsbin.com/bogelujaqe/1/edit?html,output

### Screendump
overlay_flicker

### Versions
Mapbox-gl.js: `0.14.3` 
Chrome: `Version 48.0.2564.116 (64-bit)`
OSX: `10.9.5` 
GPU: `NVIDIA GeForce 8600M GT 512 MB` 

### Code
```javascript


    var systemapic_access_token = 'pk.8FhhB90ax6KkQmoK0AMePd0R6IlkxM4VAGewsXw8';

    var systemapic = {
        "version": 8,
        "sources": {
            "norkart-tiles" : {
                "type" : "raster",
                "tiles" : [
                    "https://proxy-txa.systemapic.com/v2/tiles/norkart/aerial/{z}/{x}/{y}.jpeg?access_token=" + systemapic_access_token,
                    "https://proxy-txb.systemapic.com/v2/tiles/norkart/aerial/{z}/{x}/{y}.jpeg?access_token=" + systemapic_access_token,
                    "https://proxy-txc.systemapic.com/v2/tiles/norkart/aerial/{z}/{x}/{y}.jpeg?access_token=" + systemapic_access_token,
                    "https://proxy-txd.systemapic.com/v2/tiles/norkart/aerial/{z}/{x}/{y}.jpeg?access_token=" + systemapic_access_token,
                ],
                "tileSize" : 256
            }
        },
        "layers": [
            {
                "id": "background",
                "type": "raster",
                "source": "norkart-tiles",
                "minzoom": 0,
                "maxzoom": 22
            },
           
        ]
    };

  
    var map = new mapboxgl.Map({
        container: 'map', // container id
        center : [ 20.426873995740664, 69.31845195098202],
        zoom : 13,
        style : systemapic
    });



```

lucaswoj:
fixes #2208
fixes #2201 

cc @jfirebaugh @ansis @mourner

### DIFF
--- a/js/render/draw_raster.js
+++ b/js/render/draw_raster.js
@@ -12,7 +12,7 @@ function drawRaster(painter, source, layer, coords) {
     // Change depth function to prevent double drawing in areas where tiles overlap.
     gl.depthFunc(gl.LESS);
 
-    for (var i = coords.length - 1; i >= 0; i--) {
+    for (var i = 0; i < coords.length; i++) {
         drawRasterTile(painter, source, layer, coords[i]);
     }
 
--- a/js/util/lru_cache.js
+++ b/js/util/lru_cache.js
@@ -44,12 +44,20 @@ LRUCache.prototype.reset = function() {
  * @private
  */
 LRUCache.prototype.add = function(key, data) {
-    this.data[key] = data;
-    this.order.push(key);
 
-    if (this.order.length > this.max) {
-        var removedData = this.get(this.order[0]);
-        if (removedData) this.onRemove(removedData);
+    if (this.has(key)) {
+        this.order.splice(this.order.indexOf(key), 1);
+        this.data[key] = data;
+        this.order.push(key);
+
+    } else {
+        this.data[key] = data;
+        this.order.push(key);
+
+        if (this.order.length > this.max) {
+            var removedData = this.get(this.order[0]);
+            if (removedData) this.onRemove(removedData);
+        }
     }
 
     return this;
--- a/test/js/util/lru_cache.test.js
+++ b/test/js/util/lru_cache.test.js
@@ -18,13 +18,27 @@ test('LRUCache', function(t) {
     t.end();
 });
 
+test('LRUCache - duplicate add', function(t) {
+    var cache = new LRUCache(10, function() {
+        t.fail();
+    });
+
+    cache.add('a', 'b');
+    cache.add('a', 'c');
+
+    t.deepEqual(cache.keys(), ['a']);
+    t.ok(cache.has('a'));
+    t.equal(cache.get('a'), 'c');
+    t.end();
+});
+
 test('LRUCache - overflow', function(t) {
     var cache = new LRUCache(1, function(removed) {
-        t.equal(removed, 'c');
+        t.equal(removed, 'b');
         t.end();
     });
     cache.add('a', 'b');
-    cache.add('a', 'c');
+    cache.add('c', 'd');
 });
 
 test('LRUCache#reset', function(t) {


lucaswoj:
6171a2fc23224870c1e6d943bc4db4abbe36b525 makes me think this might be the wrong solution.

ansis:
This looks like the right solution! I think this broke with https://github.com/mapbox/mapbox-gl-js/commit/4f99e94634b4e0fb09b0280a2f08164f029547d2

lucaswoj:
@ansis I amended this PR to fix both bugs. :shipit:?

ansis:
In what cases are tiles added to the cache while they are already in the cache? Is that a bug?

lucaswoj:
It happened rarely during very fast zooms, probably due to a race condition. The LRU cache shouldn't get stuck in an infinite loop if the same key is added twice. We can debug the race condition separately. I want to get this out in a release ASAP.

lucaswoj:
You're right. Reading the code, it is mysterious that this method could ever add the same tile twice.

https://github.com/mapbox/mapbox-gl-js/blob/e7b551ccc8c36cf2031684398b7ae7b28fd9fa0d/js/source/tile_pyramid.js#L359

lucaswoj:
I have a hunch that the duplicate tiles have the same root cause as https://github.com/mapbox/mapbox-gl-js/issues/1834. Our re-use of tiles across the dateline isn't perfect. Because the LRU cache uses wrapped coords but the tile pyramid doesn't, we might add duplicate entries for a wrapped coord.

ansis:
Yeah, your hunch sounds right. I'll take a quick look to see if I can find something that regressed recently otherwise we can merge this to fix the fire and then properly fix it later.

lucaswoj:
Did you mention that we were (unintentionally) not using the cache at all until a recent fix?

ansis:
> Did you mention that we were (unintentionally) not using the cache at all until a recent fix?

We were using it, but only for the tiles at the current zoom. https://github.com/mapbox/mapbox-gl-js/commit/e6cb15c3f18c73cfc7aea2a797d3a33eedb946cd makes it look for parents too, which could have introduced this bug, or it could have just made it easier to hit.

ansis:
Could this be it?
- a tile across the dateline is need for the current view, `addTile` is called for it.
- [this](https://github.com/mapbox/mapbox-gl-js/blob/e5e5f321362e841225f55b1d92af22497b1df1bc/js/source/tile_pyramid.js#L312) does not return a tile because the tile has never been added for "world 0"
- [it gets added to `this._tiles`](https://github.com/mapbox/mapbox-gl-js/blob/e5e5f321362e841225f55b1d92af22497b1df1bc/js/source/tile_pyramid.js#L334)
- you pan and now need the "world 0" version of that tile as well, call `addTile` for it
- [all the checks for existing tiles](https://github.com/mapbox/mapbox-gl-js/blob/e5e5f321362e841225f55b1d92af22497b1df1bc/js/source/tile_pyramid.js#L312-L325) fail and it gets loaded instead of sharing the first version
- you now have tiles in `this._tiles` that have distinct `Tile` objects but the same `tile.coord.wrapped().id`
- both will be added to the cache with this wrapped id when they are removed

Ideally c33ad33 isn't needed but it is a safe fix. Merging now, releasing and then properly fixing later sounds good!

ryanbaumann:
In version 0.15.0, the performance is improved.  

However, I still notice flickering on Satellite maps with Map Pitch > 0 while panning and rotating on Chrome w/ Nvidia GTX680.  Example below:

[Video Example](https://drive.google.com/file/d/0B_LDe4RHtu0eTDFjdEJtSDlWenc/view)

[JSFiddle](https://jsfiddle.net/rsbaumann/g3v05y6x/3/)

ansis:
@ryanbaumann thanks for the video and jsfiddle. It looks like a different bug so I've opened a new issue: https://github.com/mapbox/mapbox-gl-js/issues/2212

PR #2214, opened 2016-03-02
Rework color switcher example

tristen:
color-switcher

Removes `onClick` attribute from `<li>` elements and derives swatch elements with events by appending them in sequence. This also standardizes the container/ui style to match other examples.

cc @jfirebaugh


jfirebaugh:
:+1:

PR #2217, opened 2016-03-02
Improve "Filter markers by text input" example

lucaswoj:
We shouldn’t require users to know Maki icons’ names by sight (like
Mapbox employees do!)

Also brings a little more visual emphasis to the icons over the basemap. 

cc @tristen


tristen:
:+1: good call

PR #2220, opened 2016-03-04
Better text label style

tristen:
I should have checked the previous pr out locally. These labels shouldn't look like link attributes. This is a light commit that cleans up the labels under markers by a hair

screen shot 2016-03-03 at 11 46 34 pm


lucaswoj:
:ship: Thanks!

PR #2221, opened 2016-03-04
Only recalculate properties that potentially change

### LINKED ISSUE #2141: optimize style recalculation
It takes 5-15ms to recalculate the mapbox-streets style. That's way too long for something that runs every frame while zooming.

Style recalculation takes twice as long as the rest of the rendering work done by the cpu:
screen shot 2016-02-17 at 5 36 33 pm

We should have a benchmark for this.

mourner:
Closes #2141. With this commit, style.recalculate takes ~~3-4x~~ 7-8x less time per frame, bringing its performance back to normal.  Sample 10s zooming profile: 2028ms before, ~~507ms~~ 260ms after.  cc @lucaswoj @ansis

### DIFF
--- a/js/style/style_layer/line_style_layer.js
+++ b/js/style/style_layer/line_style_layer.js
@@ -12,22 +12,21 @@ module.exports = LineStyleLayer;
 LineStyleLayer.prototype = util.inherit(StyleLayer, {
 
     getPaintValue: function(name, zoom, zoomHistory) {
-        var output = StyleLayer.prototype.getPaintValue.call(this, name, zoom, zoomHistory);
+        var value = StyleLayer.prototype.getPaintValue.call(this, name, zoom, zoomHistory);
 
         // If the line is dashed, scale the dash lengths by the line
         // width at the previous round zoom level.
-        if (output && name === 'line-dasharray') {
-            var lineWidth = this.getPaintValue('line-width', Math.floor(zoom), Infinity);
-            output.fromScale *= lineWidth;
-            output.toScale *= lineWidth;
+        if (value && name === 'line-dasharray') {
+            var flooredZoom = Math.floor(zoom);
+            if (this._flooredZoom !== flooredZoom) {
+                this._flooredZoom = flooredZoom;
+                this._flooredLineWidth = this.getPaintValue('line-width', flooredZoom, Infinity);
+            }
+
+            value.fromScale *= this._flooredLineWidth;
+            value.toScale *= this._flooredLineWidth;
         }
 
-        return output;
-    },
-
-    recalculate: function(zoom, zoomHistory) {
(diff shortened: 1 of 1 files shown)


mourner:
The second commit brings it down to 340ms, which means 5-6x faster.

mourner:
The last two commits bring recalculation down to 260ms. This is ready for review now.

jfirebaugh:
Why are paint functions not recalculated?

mourner:
@jfirebaugh why do you think they're not recalculated?

lucaswoj:
I'm a big fan of this approach! Great work! :rocket:

js/style/style_layer/line_style_layer.js
```diff
@@ -11,18 +11,20 @@ module.exports = LineStyleLayer;
...
+                this._flooredZoom = flooredZoom;
+                this._flooredLineWidth = this.getPaintValue('line-width', flooredZoom, Infinity);
+            }
+
+            this.paint['line-dasharray'].fromScale *= this._flooredLineWidth;
+            this.paint['line-dasharray'].toScale *= this._flooredLineWidth;
         }
-
-        return output;
     }
```
lucaswoj: Why did this logic get moved from `getPaintValue` to `recalculate`? There are some architectural advantages to being able to calculate style properties independent of a `recalculate` call.

mourner: Recalculate is called much less often so it gives a little perf boost, but I can move this back if you wish.

mourner:
> Why are paint functions not recalculated?

@jfirebaugh the confusion probably stems from the fact that paint properties are all wrapped with transition objects, which are then used in `getPaintValue` to get actual values.

js/style/style_layer/line_style_layer.js
lucaswoj: I'll need it moved back eventually for data-driven `line-dasharray`s. If the perf impact isn't too extreme, lets move it back now.

mourner: Right, moved it back! Turned not to have much impact since additional `getPaintValue` only gets called when a round zoom is changed.

mourner:
Messed up rebase/push order so the PR shows 0 commits now, but just for reference, they were:

- https://github.com/mapbox/mapbox-gl-js/commit/6fa79fb23c17f499e708bbe6e5e931457c926a5f
- https://github.com/mapbox/mapbox-gl-js/commit/5107fc27e5bcaffba419604ef3dd9171f330c31b
- https://github.com/mapbox/mapbox-gl-js/commit/c3692021a732abe8ae6d742dd3ad600db311d7b7
- https://github.com/mapbox/mapbox-gl-js/commit/61ca1c8bcfa24d38439e97ef0192cdac9e508081

PR #2001, opened 2016-01-27
Allow GeoJSON to be passed as stringify JSON.

dcervelli:
Mitigation for #1504.

With a complex GeoJSON object (say 20k features with 100k coordinates) the `postMessage` call to the worker takes several seconds on the UI thread. By manually `JSON.stringify`ing the GeoJSON object and `JSON.parse`ing on the worker side we're seeing a 6x perf improvement.

Ideally we would figure out a way to use transferables or similar (there is still a small UI freeze in our usage), but this is a step in the right direction.

Issue #1504, opened 2015-09-18
Optimize the GeoJSON ⇢ worker transfer


jfirebaugh:
Of course, it's always best to load directly into the worker with an external URL, but many use cases with GeoJSON requiring generating / filtering / sorting GeoJSON on the main thread and then passing it to a GeoJSON layer. We could try using blob URLs to see if it helps.

On the main thread:

```
var blob = new Blob([JSON.stringify(geoJSON)], { type: 'application/json' });
var url = (window.URL || window.webkitURL).createObjectURL(blob);
```

On the worker thread, do a normal ajax request and then `revokeObjectURL`.

My guess is this won't really help because you still have to stringify the GeoJSON, but it's worth a shot.

abuchanan:
What about transferable objects?
https://developers.google.com/web/updates/2011/12/Transferable-Objects-Lightning-Fast

abuchanan:
Although the transferred object would be removed from the main process? Not sure how you'd deal with that unless you want to deep clone the data before transfer. Maybe that's just as slow as stringify.

jfirebaugh:
That's another thing to try, although again, it would require a stringify + parse, plus additional overhead in copying bytes from a string to a typed array and back.

abuchanan:
Ya. Was later thinking that maybe one way to ensure great performance would be to get your ad hoc code into the worker somehow.

mcwhittemore:
Per chat with @jfirebaugh we should perf out a few different ways to move data to Worker processes before we try to improve this perf. He recommended looking into:

- Compare object
- self-stringify
- blob trick
- Transferrable object

I'd like to dive into this, but I'm a few weeks out before I'd have the time to really solve this problem.

lucaswoj:
In #2001, @dcervelli provided a PR that accepts pre-stringified GeoJSON, citing a 6x perf boost. 

We decided to take a step back and think about the best long term solution to the problem before including anything in the next release. 

@mourner suggested two additional transfer techniques we should investigate 

 - stringified JSON as a transferrable byte array
 - geojson as a byte array

mourner:
@dcervelli @jfirebaugh @lucaswoj Just did a quick test of the stringified JSON approach. Performance on a 11MB GeoJSON in Chrome: 

```
stringify json: 244.846ms
convert to Uint16Array: 70.497ms
convert back to string: 558.237ms
parse json: 231.529ms
```

[...]

PR #2222, opened 2016-03-04
pass GeoJSON in stringified form

mourner:
This is a reapplication of #2001 without changing the API, and serves as the middle ground for improving `setData` performance as discussed in #1504.

cc @lucaswoj @jfirebaugh @dcervelli

### DIFF
--- a/js/source/geojson_source.js
+++ b/js/source/geojson_source.js
@@ -142,18 +142,21 @@ GeoJSONSource.prototype = util.inherit(Evented, /** @lends GeoJSONSource.prototy
 
     _updateData: function() {
         this._dirty = false;
-        var data = this._data;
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Given http://nolanlawson.com/2016/02/29/high-performance-web-worker-messages/ I wonder if we should just build `JSON.stringify` / `JSON.parse` into `Actor`.

mourner:
@jfirebaugh I thought about that, but it's tricky because of transferables — you don't want to stringify an array buffer.

jfirebaugh:
Yeah, it would have to be only if the `buffers` argument to `send` is undefined.

lucaswoj:
I'm :+1: on shipping this as-is and continuing to optimize per #1504 :shipit:

Issue #1002, opened 2015-02-17
Add a "layers" parameter to "Evented#on" 


jfirebaugh:
Map events should accept an optional second parameter which is the name of a layer or an array of layer names for which to delegate events. Internally, this would bind an event handler that uses `featuresAt` and dispatches the event only when it matches features from the given layer(s). The features themselves would be passed as a member of the event object. For example:

``` js
map.on('click', 'poi', function(e) {
    (new Popup())
        .setLatLng(e.features[0].latLng)
        .setHTML(e.features[0].properties.description)
        .addTo(map);
});
```

This should not require setting `interactive: true` for each layer separately; binding a delegated event should automatically enable interactivity for the specified layers.

mourner:
> This should not require setting interactive: true for each layer separately; binding a delegated event should automatically enable interactivity for the specified layers.

This may be tricky, because you need to reparse tiles when `interactive` for a layer changes.

lucaswoj:
We should do this as soon as #2224 lands!

arthurhammer:
Playing with v0.35 and pretty happy about this change. Any plans to support multiple layers to query with `on`?

anandthakker:
Seems like a doable enhancement to me -- @jfirebaugh did you encounter any subtleties in implementing this that would make this problematic?

mourner:
Looks like a one-line change to me since underlying `queryRenderedFeatures` already accepts an array of layer ids.

jfirebaugh:
Yeah, should be straightforward. @arthurhammer can you open a new issue so we don't lose track of the request?

arthurhammer:
Done in #4578. Thanks!

Issue #2223, opened 2016-03-04
Tile number can't display on satellite style


mayagao:
**mapbox-gl-js version**:

0.15.0
### Steps to Trigger Behavior

With satellite style, turn on tile boundaries
### Expected Behavior

Display tile numbers just like on other styles
### Actual Behavior

Tile numbers can't display correctly

![screen shot 2016-03-04 at 8 48 32 am](https://cloud.githubusercontent.com/assets/5186564/13533580/ed078388-e1e5-11e5-9c58-7cac4ac79c2e.png)

lucaswoj:
bug confirmed, here's a functioning [demo](http://jsbin.com/mitepazese/edit?html,output)

lucaswoj:
Some context

 - this bug exists as far back as f06c3e53d6278542ef296050b90540c6aa27a1b7 (first included in v0.14.1)
 - before f06c3e53d6278542ef296050b90540c6aa27a1b7, debug labels don't render on the satellite map at all

lucaswoj:
If I fix the bug causing debug labels to not render before f06c3e5, then I see that this bug seems to have been introduced in 250a50056a1a0ec63fcaf1ab608727c5048cc400

lucaswoj:
@ansis Any ideas on why this would break iff there is a raster layer present?

ansis:
@lucaswoj I think it's because raster tiles are 256 instead of 512? which makes [this hack]( https://github.com/mapbox/mapbox-gl-js/blob/61ca1c8bcfa24d38439e97ef0192cdac9e508081/js/render/draw_debug.js#L44) wrong and also makes everything smaller.

lucaswoj:
Just confirmed that this bug is still present in `master` 😞

PR #1979, opened 2016-01-23
Add "skipWorker" option to GeoJSON sources [DO NOT MERGE]

lucaswoj:
Many users, especially mapbox-gl-draw, have demonstrated a need for high responsiveness in updating GeoJSON sources. One potential solution is to do GeoJSON parsing / tiling in the main thread rather than the worker.

https://github.com/mapbox/mapbox-gl-draw/pull/185

# Next Steps

 - [ ] quantify performance change
 - [ ] quantify bundle size change
 - [ ] DRY up code
 - [x] fix `featuresIn` / `featuresAt`
 - [ ] add integration tests for GeoJSON sources (or ticket out separately)

PR #2224, opened 2016-03-05
queryRenderFeatures continued [not ready]

ansis:
This builds on https://github.com/mapbox/mapbox-gl-js/pull/2052.

This is not ready, but it could use some high level :eyes:. I probably should have gotten some :eyes: a bit earlier. Let me know if any of the high level ideas here concern you. I'll comment again later when this is more polished.

The raw vector tile pbf, FeatureTree and CollisionTile are now transferred to the main thread when a tile finishes loading. This lets use have a synchronous `queryRenderedFeatures`, it lets us eliminate `includeGeometry` without a performance hit, and it starts moving us towards more state-less workers.

### A synchronous queryRenderedFeatures
This adds a synchronous version of `queryRenderedFeatures`.

- it's faster
- it's more convenient to use, especially when you want to make multiple queries
- it's more consistent. It never gets delayed for 100s of ms while a worker is busy parsing a tile
- you don't have to worry about callbacks being called out of order
- things like https://github.com/mapbox/mapbox-gl-js/issues/1002 could be done lazily

There is still an async version. It transfers data back to workers.

Despite being parallelized, the async version is not faster than the sync version. It ends up parsing from the pbfs twice instead of once. It spends less time on the main thread though which could be useful for large, expensive queries. I think it could be worthwhile to keep both versions as `queryRenderedFeatures` and `queryRenderedFeaturesAsync`

### Getting rid of `includeGeometry`

The `includeGeometry` param for `queryRenderedFeatures` only exists because sending geometries from the worker back to the main thread was slow. When the pbf is on the main thread individual geometries can be deserialized **lazily**. This lets us drop `includeGeometry` without a performance penalty.

### indexing with Grid

I replaced rbush with a [`Grid`](https://github.com/mapbox/mapbox-gl-js/blob/query-rendered-features-continued/js/util/grid.js), a different kind of spatial index. `Grid` can be serialized to an ArrayBuffer to make it transferrable. The serialized version can be queried directly without deserializing individual elements to javascript objects. It is [faster than rbush for collision detection](https://github.com/mapbox/mapbox-gl-js/commit/0ec60ff96749276f32710ec43f26ccff03f50096) and comparable for [rendered feature querying](https://github.com/mapbox/mapbox-gl-js/commit/6e3f8c7a2a2388718afa5c2dc4d64395a6c82f5d). It's faster because insertions are faster. 

 Grid is a grid of cells. When a bbox is inserted it is added to all the cells it intersects. When a bbox is queried it looks in all cells that intersects with that bbox. That's it.

It isn't faster than rbush generally, but it is faster for what we use it for: a fixed extent, mostly evenly sized bboxes, and a high number of insertions. More importantly, it's transferable. 

### Storing arrays of "structs" in ArrayBuffers
There are a bunch of cases where we have giant arrays of fixed size objects. I added a new abstraction, [StructArray](https://github.com/mapbox/mapbox-gl-js/blob/query-rendered-features-continued/js/util/struct_array.js), that makes it easier to store these kinds of objects in ArrayBuffers. It's inspired by the new Buffer implementation. It:
- saves a lot of space. Switching collision boxes saved 500KB per tile.
- makes transferring or even copying these arrays a lot faster
- avoids creating a ton of garbage

see https://github.com/mapbox/mapbox-gl-js/commit/9a2e5fd4645cde999005796538da975ab5b5d217 's commit message for an example

### Reducing worker state
Both Grid and StructArrays are transferrable and don't need to be deserialized to individual objects before they can be used. This makes moving them between workers and the main thread really fast and reduces the amount of data that is "stuck" on workers. When `symbolInstances` is made completely transferable we could get rid of WorkerTiles that live after parsing has finished.

The work requests would transfer all the data that is needed and not rely on the worker having any unique data. I think this is also more similar to how we want work requests to work in -native. 

Since work won't be tied to a specific thread, it should be possible to distribute work more consistently and speed things up.

### Benchmarks

Compared to https://github.com/mapbox/mapbox-gl-js/pull/2052

```

point query
http://localhost:9966/debug/#15.37/38.9047/-77.0299
mapbox-streets-v6 (before all features were merged)
new sync                        1 ms
new sync with geometry          1 ms
new async                       5-6 ms
new async with geometry         5-6 ms
old async                       1-2 ms
old async with geometry         2 ms

point query
http://localhost:9966/debug/#15.37/38.9047/-77.0299
mapbox-streets-v7 (after many layers are merged into a single huge multi-geometry feature)
new sync                        2 ms
new sync with geometry          2 ms
new async                       5 ms
new async with geometry         5-6 ms
old async                       5 ms
old async with geometry         15-30 ms

full screen box query
http://localhost:9966/debug/#15.37/38.9047/-77.0299
mapbox-streets-v7 (after many layers are merged into a single huge multi-geometry feature)
new sync                        23 ms
new sync with geometry          70 ms
new async                       19-25 ms
new async with geometry         75 ms
old async                       45 ms
old async with geometry         770 ms


full screen bbox query
http://localhost:9966/debug/#11.52/38.9262/-77.0420
mapbox-streets-v7 (after many layers are merged into a single huge multi-geometry feature)
new sync                        115 ms
new sync with geometry          175 ms
new aysnc                       125 ms
new async with geometry         180 ms
old async                       265 ms
old async with geometry         560 ms
```

- The old and new sync queryRenderedFeatures are both really fast for point queries.
- The new queryRenderedFeatures is much faster for large box queries, especially if geometries are included.
- The new sync and async queryRenderedFeatures take a similar amount of time, but the async version spends less time on the main thread.

### next steps
Cleaning things up. Fixing tests. Fixing a couple remaining bugs.

Do all the high level changes sound ok?

@lucaswoj @jfirebaugh @mourner


mourner:
What's the memory impact of switching to grid index and storing it in both the main thread and the worker?

lucaswoj:
I'm very :+1: on the high-level changes in this PR.

Before shipping, I'd like for us to think about 

 - what it'd take to merge `StructArray` and `Buffer` ([LOC is proportional to # of bugs](http://www.techrepublic.com/blog/it-security/the-danger-of-complexity-more-code-more-bugs/))
 - how stateless workers can be used to write better integration tests for mapbox-gl-js (the tests can be another PR but lets think about the architecture that enables those tests now)
 - how we can codify the performance benchmarks used in this PR into something reusable

ansis:
> What's the memory impact of switching to grid index and storing it in both the main thread and the worker?

Good question. I hadn't benchmarked that yet.

It is generally only stored in one thread at a time. Async queries create a duplicate but that duplicate is discarded at the end of the query. We could maintain a duplicate version to reduce the cpu cost, but the cpu cost is pretty low and async queries should only be used for expensive queries.

For each of the benchmarks I cropped the window so that it loaded only one tile. I used mapbox-streets-v6 because v7 merges features into huge multigeometries. This branch implements subfeature indexing (a bbox for each ring, not each feature) but the old one doesn't so it isn't a fair comparison when using v7.

```
At z12 with many lines and few labels.
http://localhost:9966/debug/#12.89/38.9188/-77.0388

FeatureTree with rbush          1843 KB         343 KB for the tree, 1500 KB for the bbox arrays that are inserted
CollisionTile with rbush        20 KB
CollisionBoxes

FeatureTree with Grid           340 KB          240 KB for the Grid, 100 KB for the featureIndexArray
CollisionTile with Grid         225 KB

At z17 with less lines but more labels.
http://localhost:9966/debug/#17.09/38.90398/-77.03116

FeatureTree with rbush          420 KB          50 KB for the tree, 270 KB for the bbox arrays that are inserted
CollisionTile with rbush        200 KB

FeatureTree with Grid           117 KB          100 KB for the Grid, 17 KB for the featureIndexArray
CollisionTile with Grid         280 KB


z12
CollisionBoxes used with rbush  140 KB
CollisionBoxes used with Grid   26 KB

z17
CollisionBoxes used with rbush  1160 KB
CollisionBoxes used with grid   235 KB

rbush could be switched to use the StructArray collision boxes currently used
by Grid but it still might need regular bbox arrays to avoid a cpu performance hit.
```

The Grid in FeatureTree uses far less memory than rbush. 340 KB vs 1843 KB and 117 KB vs 420 KB.

The Grid in CollisionTile uses more memory than rbush. 225 KB vs 20 KB and 280 KB.
- this excludes the cost of the actual CollisionBoxes that are inserted into the tree, which is way higher in the rbush version. But the rbush version could possibly be rewritten to use the typed array collision boxes which could reduce that cost
- this is mostly the memory cost of creating a Int32 subarray for each cell (112 bytes per cell). I'll look into optimizing this for cases where no or few cells are actually used (like the geojson source that is part of the benchmark)

---

Summary:
- lower memory usage overall
- much lower usage for feature indexes
- possibly high and possibly lower usage for collision indexes: it's hard to separate from other optimizations

ansis:
>  - what it'd take to merge `StructArray` and `Buffer` ([LOC is proportional to # of bugs](http://www.techrepublic.com/blog/it-security/the-danger-of-complexity-more-code-more-bugs/))

Buffer does a whole bunch of extra calculations like scaling values and it doesn't provide an interface for reading values. I think it might make sense to wait until dds settles down before trying to merge the two.

>  - how stateless workers can be used to write better integration tests for mapbox-gl-js (the tests can be another PR but lets think about the architecture that enables those tests now)

How would stateless workers help us write better integration tests?

> - how we can codify the performance benchmarks used in this PR into something reusable 

Yes, I definitely want to write reusable benchmarks. Do we want to start using a framework for this or continue writing it ourselves (like with the buffer benchmark)?

ansis:
ceb13b8 reduces Grid memory usage from 225 KB to 85 KB for the z12 case and from 280 KB to 169 KB for the z17 case.

mourner:
@ansis fantastic! I'm all :+1: on the changes then. I'll review the code in more detail.

lucaswoj:
> Buffer does a whole bunch of extra calculations like scaling values 

That's not correct. Those "extra calculations" are done in `Bucket`. Buffer is designed to be a lightweight generic `StructArray` implementation. 

> and it doesn't provide an interface for reading values. 

It has a [`Buffer#get` method](https://github.com/mapbox/mapbox-gl-js/blob/master/js/data/buffer.js#L152) for reading values. The docs say that it is "only used for debugging" as an indication that it is dead code, not as an indication of any technical limitations. 

> I think it might make sense to wait until dds settles down before trying to merge the two.

The `Buffer` class will not change in any future dds work.

-----

> How would stateless workers help us write better integration tests?

Out of scope for this PR. I'm excited about the implications of a simpler interface for writing better integration tests through `skipWorker`-like functionality #1979.

`test-suite` is slow and limited in what it can test. https://github.com/mapbox/mapbox-gl-js/issues/2223 is a perfect example of a bug that we won't be able to write regression tests for and maybe would've been caught if we had better integration tests.

----- 

> Yes, I definitely want to write reusable benchmarks. Do we want to start using a framework for this or continue writing it ourselves (like with the buffer benchmark)?

I'm going to design a simple benchmark interface specification today. It'll be similar to how the buffer benchmark works.

ansis:
> That's not correct. Those "extra calculations" are done in Bucket. Buffer is designed to be a lightweight generic StructArray implementation.

Yeah, you're right. I confused the two. I'll take a closer look at this today.

> The Buffer class will not change in any future dds work.

Great!

> I'm going to design a simple benchmark interface specification today. It'll be similar to how the buffer benchmark works.

Is it worth writing our own? Should we just use something like https://github.com/bestiejs/benchmark.js/ ?

lucaswoj:
benchmark interface specification pushed to #2231 :smile:

ansis:
Rebased on master. Latest commit used to be 3fc75bd3127436715d0ec2e88d52aade617ea526.

ansis:
d40a059 merges the similar parts of Buffer and StructArray. After the changes, Buffer is a thing that turns a StructArray into a WebGL buffer. @lucaswoj can you review?

lucaswoj:
A convention I've been using, and have found fruitful:

 - object constructors take a single `options` argument 
 - `serialize` methods output an object compatible with the constructor `options` argument

ansis:
> A convention I've been using, and have found fruitful
> ...

switched in fe8dad80fb4ce20d357b01bdc899a230289e09f6

----

and then I squashed all those commits that fixed things from d40a059ef0f94b10a93f0578658213ac5a21229

mcwhittemore:
Any time frame here? This would solve a bunch of event handler problems for gl-draw.

ansis:
@mcwhittemore soon, hopefully

@lucaswoj I added migration instructions to changelog, a benchmark for point queries, a benchmark for full-screen box queries, and updated all examples. All that's left is review and fixing things based on that.

averas:
@ansis @lucaswoj @jfirebaugh Did you read my last post on hit hierarchies over at #2052? If my reasoning has any merit I was thinking that it's best to squeeze in as many breaking changes as possible since this is a quite significant API update anyway.

jfirebaugh:
@averas We did discuss the result format, and the consensus was that for simplicity's sake we want to keep a consistent, flat format between `queryRenderedFeatures` and `querySourceFeatures`.

averas:
@jfirebaugh :+1: Got it.

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
+## Unreleased
+
+#### Breaking Changes
+
+- `map.featuresAt` and `map.featuresIn` are removed. Use `map.queryRenderedFeatures` or `map.querySourceFeatures` instead. To migrate:
+    - replace `featuresAt` and `featuresIn` with `queryRenderedFeatures`
+    - `queryRenderedFeatures` is synchronous. Remove the callback and use the return value. An async version `queryRenderedFeaturesAsync` exists for really slow queries.
+    - rename the `layer` parameters to `layers` and make it an array of strings.
+    - remove the `radius` paramter. `queryRenderedFeatures` automatically uses style property values when doing queries.
```
lucaswoj: This doesn't make sense to me. How do style property values <--> `radius`?

mourner: This was the initial intent of #2052. Style properties like line width, translate, etc. all affect whether an object is under a cursor or not, even with the same coordinates. Previously we would use `radius` because we checked against original coordinates, not rendered representation, but it's no longer needed.

js/data/bucket.js
```diff
@@ -176,9 +187,9 @@ Bucket.prototype.destroy = function(gl) {
     }
 };
 
-Bucket.prototype.trimBuffers = function() {
-    for (var bufferName in this.buffers) {
-        this.buffers[bufferName].trim();
+Bucket.prototype.trimArrays = function() {
```
lucaswoj: Nit: Can we call this `trimStructArrays` for consistency with `createStructArrays`?

lucaswoj: I'd also be :+1: with `structArray` -> `array` everywhere.

js/data/feature_tree.js
```diff
@@ -1,73 +1,137 @@
...
+    }
+    this.coord = coord;
+    this.overscaling = overscaling;
     this.x = coord.x;
     this.y = coord.y;
     this.z = coord.z - Math.log(overscaling) / Math.LN2;
-    this.rtree = rbush(9);
-    this.toBeInserted = [];
     this.setCollisionTile(collisionTile);
 }
```
lucaswoj: The implementation I had in mind in https://github.com/mapbox/mapbox-gl-js/pull/2224#issuecomment-197043523 was more like:

```js
function FeatureTree(options) {
    this.overscaling = options.overscaling;
    this.rawTileData = options.rawTileData;
    this.numberToLayerIDs = options.numberToLayerIDs;
    this.coord = options.coord;
    
    this.featureIndexArray = new FeatureIndexArray(options.featureIndexArray);
    this.grid = new Grid(options.grid || {n: 15, extent: EXTENT, padding: 0});
    this.rtree = rbush(9);

    this.toBeInserted = [];

    this.x = this.coord.x
    this.y = this.coord.y
    this.z = this.coord.z - Math.log(overscaling) / Math.LN2;

    this.setCollisionTile(options.collisionTile);
}
```

Instead of thinking about the constructor of having "unserialized" mode and "normal" mode, think about it as accepting an `options` object flexible enough to hit both use cases.

js/data/feature_tree.js
```diff
@@ -1,73 +1,137 @@
...
-};
+FeatureTree.prototype.insert = function(feature, featureIndex, sourceLayerIndex, bucketIndex) {
+    var key = this.featureIndexArray.length;
+    this.featureIndexArray.emplaceBack(featureIndex, sourceLayerIndex, bucketIndex);
+    var geometry = loadGeometry(feature);
+
+    for (var r = 0; r < geometry.length; r++) {
+        var ring = geometry[r];
+
+        // TODO: skip holes when we start using vector tile spec 2.0
```
lucaswoj: I would like this to be a GitHub issue rather than a comment.

js/data/feature_tree.js
```diff
@@ -88,80 +152,181 @@ FeatureTree.prototype.query = function(args, styleLayersByID) {
...
+    matching.sort(topDown);
+    var match = this.featureIndexArray.get(0);
+    this.filterMatching(result, matching, match, queryGeometry, filter, params.layers, styleLayersByID, args.bearing, pixelsToTileUnits, returnGeoJSON);
+
+    var matchingSymbols = this.collisionTile.queryRenderedSymbols(minX, minY, maxX, maxY, args.scale);
+    var match2 = this.collisionTile.collisionBoxArray.get(0);
+    matchingSymbols.sort();
+    this.filterMatching(result, matchingSymbols, match2, queryGeometry, filter, params.layers, styleLayersByID, args.bearing, pixelsToTileUnits, returnGeoJSON);
+
+    return result;
```
lucaswoj: Can `result` be either a `StructArray` or vanilla object here? Is there any way we can normalize it before returning?

lucaswoj: Ah. I see. This is predicated on `returnGeoJSON`. Are there perf implications to doing that conversion outside of this method?

js/data/feature_tree.js
```diff
@@ -88,80 +152,181 @@ FeatureTree.prototype.query = function(args, styleLayersByID) {
...
+};
+
+function topDown(a, b) {
```
lucaswoj: `compareReverse`

js/data/feature_tree.js
```diff
@@ -88,80 +152,181 @@ FeatureTree.prototype.query = function(args, styleLayersByID) {
...
+}
 
-    var matching = this.rtree.search(bounds).concat(this.collisionTile.getFeaturesAt(symbolQueryBox, args.scale));
+function getLineWidth(paint) {
+    if (paint['line-gap-width'] > 0) {
+        return paint['line-gap-width'] + 2 * paint['line-width'];
+    } else {
+        return paint['line-width'];
+    }
+}
```
lucaswoj: Is there any way to keep this code closer to other style calculation code? I wouldn't be opposed to making it a method like `LineStyleLayer#getHitboxWidth`

js/data/feature_tree.js
```diff
@@ -88,80 +152,181 @@ FeatureTree.prototype.query = function(args, styleLayersByID) {
...
 
+FeatureTree.prototype.filterMatching = function(result, matching, match, queryGeometry, filter, filterLayerIDs, styleLayersByID, bearing, pixelsToTileUnits, returnGeoJSON) {
```
lucaswoj: 10 arguments on a method has the potential to cause some headaches. Can we turn this into an `options` object? 

(This isn't an option in hot code but I don't think this code is hot)

js/data/feature_tree.js
```diff
@@ -88,80 +152,181 @@ FeatureTree.prototype.query = function(args, styleLayersByID) {
...
+    var seen = {};
     for (var k = 0; k < matching.length; k++) {
-        var feature = matching[k].feature,
-            layerIDs = matching[k].layerIDs;
+        var index = matching[k];
 
-        if (!filter(feature)) continue;
+        if (seen[index]) continue;
+        seen[index] = true;
```
lucaswoj: Under what circumstances are there duplicates in `matching`? Can we avoid them by making `matching` an object?

js/util/struct_array.js
```diff
@@ -0,0 +1,334 @@
...
+}
+
+/**
+ * Make this Struct object point to a different instance in the same array.
+ * It can be cheaper to use .setIndex to re-use an existing Struct than to
+ * create a new one.
+ * @param {number} index The index of the struct in the StructArray;
+ * @private
+ */
+Struct.prototype._setIndex = function(index) {
```
lucaswoj: You use this method extensively in other classes. Can we call it something more self-documenting like `setIndexOffset` and make it public?

js/source/worker_tile.js
```diff
@@ -98,10 +104,15 @@ WorkerTile.prototype.parse = function(data, layers, actor, callback) {
...
         otherBuckets = [];
 
+    featureTree.numberToLayerIDs = [];
+
     for (var id in bucketsById) {
         bucket = bucketsById[id];
         if (bucket.features.length === 0) continue;
 
+        bucket.index = featureTree.numberToLayerIDs.length;
+        featureTree.numberToLayerIDs.push(bucket.layerIDs);
```
lucaswoj: Could we get rid of `numberToLayerIDs` by changing `buckets` to an object indexed by `bucket.id`? If so, that'd be a cleaner solution and relatively easy change.

js/data/feature_tree.js
```diff
@@ -88,80 +152,181 @@ FeatureTree.prototype.query = function(args, styleLayersByID) {
...
+    }
+
+    var translated = [];
+    for (var i = 0; i < queryGeometry.length; i++) {
+        translated.push(queryGeometry[i].sub(translate._mult(pixelsToTileUnits)));
+    }
+    return translated;
+}
+
+function matchLayers(filterLayerIDs, featureLayerIDs) {
```
lucaswoj: Can we replace this with a generic `util` method? Something like `util.intersection`

js/data/feature_tree.js
```diff
@@ -88,80 +152,181 @@ FeatureTree.prototype.query = function(args, styleLayersByID) {
...
             var layerID = layerIDs[l];
 
-            if (params.layerIds && params.layerIds.indexOf(layerID) < 0) {
+            if (filterLayerIDs && filterLayerIDs.indexOf(layerID) < 0) {
                 continue;
             }
 
-            styleLayer = styleLayersByID[layerID];
-            var geometry = loadGeometry(feature);
+            var styleLayer = styleLayersByID[layerID];
```
lucaswoj: I think we can call `styleLayersByID` just `styleLayers`

js/data/feature_tree.js
lucaswoj: I can hit this stuff in a follow-up PR. Don't worry about it.

js/data/feature_tree.js
lucaswoj: After :thought_balloon: a little, I accept the idea that all "hit testing" code lives in `FeatureTree`. This is good as-is.

js/ui/map.js
```diff
@@ -374,27 +374,60 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
-        if (callback === undefined) {
+    queryRenderedFeaturesAsync: function(pointOrBox, params, callback) {
+        if (!(pointOrBox instanceof Point || Array.isArray(pointOrBox))) {
             callback = params;
             params = pointOrBox;
+            pointOrBox = undefined;
+        }
+        var queryGeometry = this._makeQueryGeometry(pointOrBox);
+        this.style.queryRenderedFeaturesAsync(queryGeometry, params, this._classes, this.transform.zoom, this.transform.angle, callback);
+        return this;
```
lucaswoj: Can we move these transformations into `Style#queryRenderedFeaturesAsync`? I would like to keep all that logic together and make this a simple proxy.

js/ui/map.js
```diff
@@ -374,27 +374,60 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+     * @example
+     * var features = map.queryRenderedFeatures([[10, 20], [30, 50]], { layers: ['my-layer-name'] });
+     */
+    queryRenderedFeatures: function(pointOrBox, params) {
+        if (!(pointOrBox instanceof Point || Array.isArray(pointOrBox))) {
+            params = pointOrBox;
+            pointOrBox = undefined;
+        }
+        var queryGeometry = this._makeQueryGeometry(pointOrBox);
+        return this.style.queryRenderedFeatures(queryGeometry, params, this._classes, this.transform.zoom, this.transform.angle);
```
lucaswoj: :point_up:

js/util/string_number_mapping.js
```diff
@@ -0,0 +1,13 @@
...
+
+function StringNumberMapping(strings) {
+    this.stringToNumber = {};
+    this.numberToString = [];
+    for (var i = 0; i < strings.length; i++) {
+        var string = strings[i];
+        this.stringToNumber[string] = i;
+        this.numberToString[i] = string;
+    }
+}
```
lucaswoj: Why this module instead of `(5).toString()` and `parseInt("5", 10)`?

js/util/struct_array.js
```diff
@@ -0,0 +1,334 @@
...
+
+    function StructArrayType() {
+        StructArray.apply(this, arguments);
+    }
+
+    StructArrayType.serialize = serializeStructArrayType;
+
+    StructArrayType.prototype = Object.create(StructArray.prototype);
+    StructArrayType.prototype.StructType = StructType;
+    StructArrayType.prototype.BYTES_PER_ELEMENT = StructType.prototype.BYTE_SIZE;
```
lucaswoj: `BYTES_PER_ELEMENT` -> `bytesPerElement`

I'm starting to reconsider my use of `UPPER_UNDERSCORE_CASE` in this codebase. If we do use it, it should be for values that are really static and global, as if a `#define` directive.

js/util/struct_array.js
```diff
@@ -0,0 +1,334 @@
...
+            });
+        }
+
+        offset += typeSize * member.components;
+
+        return member;
+    });
+
+    StructType.prototype.alignment = options.alignment;
+    StructType.prototype.BYTE_SIZE = align(offset, Math.max(maxSize, options.alignment));
```
lucaswoj: `BYTE_SIZE` -> `size`

We already state that `Note: all "sizes" are measured in bytes`.

js/util/struct_array.js
```diff
@@ -0,0 +1,334 @@
...
+        BYTES_PER_ELEMENT: this.prototype.BYTES_PER_ELEMENT
+    };
+}
+
+
+function align(offset, size) {
+    return Math.ceil(offset / size) * size;
+}
+
+function sizeOf(type) {
```
lucaswoj: `sizeof` is a protected symbol in JS. I would like to get around that by making this identifier more specific, not by changing capitalization.

js/util/struct_array.js
```diff
@@ -0,0 +1,334 @@
...
+ * Return the Struct at the given location in the array.
+ * @private
+ * @param {number} index The index of the element.
+ */
+StructArray.prototype.get = function(index) {
+    return new this.StructType(this, index);
+};
+
+/**
+ * Resize the buffer to discard unused capacity.
```
lucaswoj: buffer -> struct array

js/util/vectortile_to_geojson.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+var VectorTileFeature = require('vector-tile').VectorTileFeature;
+
+module.exports = Feature;
```
lucaswoj: I feel weird about creating a GeoJSON object with a constructor. How would you feel about using `Object.create` to get the prototype perf goodness with an interface like `convertVectorFeatureToGeoJSON(vectorTileFeature, z, x, y)`?

CHANGELOG.md
lucaswoj: Ah! That makes sense. This line would've made more sense to me initially if it were written like

 > - remove the `radius` parameter. `queryRenderedFeatures` considers style properties (like `line-width`) with an effective radius of 0 when doing queries.

lucaswoj:
The fundamentals of this PR are `:rock:` solid. Most of my feedback are suggestions to clarify the codebase. I found `FeatureTree` particularly hard to grock. I'm not opposed to leaving some of those for future refactors. 

Great work! I'm amazed at how fast this all came together. :fireworks:

CHANGELOG.md
ansis: What about this?
- remove the `radius` parameter. `radius` was often used with `featuresAt` to account for style properties like `line-width` and `circle-radius`. `queryRenderedFeatures` automatically accounts for these style properties. If you need to query a larger area, make the first argument a box instead of a point.

lucaswoj: Much better! Thanks @ansis .

js/data/feature_tree.js
ansis: Ah, yeah, I only changed `StructArrayType` and `StructArray`. `FeatureTree` probably has this same problem.

js/data/feature_tree.js
ansis: https://github.com/mapbox/mapbox-gl-js/issues/2293

js/data/feature_tree.js
ansis: I prefer `topDown` because the intention is to sort the keys by the order their features are drawn, which just happens to be reverse/descending order. There's probably an even clearer name I'm not thinking of right now.

js/data/feature_tree.js
ansis: I think it could be good to eventually have each render type have it's own index and hit testing code

js/data/feature_tree.js
ansis: The new mapbox streets vector tiles have huge combined multi-geometries. The entire buildings layer is one giant multipolygon. Because of this, rings are indexed individually: https://github.com/mapbox/mapbox-gl-js/blob/158227693667b2e7d6cfa2a7e5822c7d5ad81d30/js/data/feature_tree.js#L65-L79

If multiple rings intersect the query box then there will be multiple duplicate keys.

We could make Grid remove duplicates by creating an object or checking some other way, but I'm not sure that is a better spot for it.

Since the keys are in sorted order we could also just do this:
Is this better?

```diff
diff --git a/js/data/feature_tree.js b/js/data/feature_tree.js
index 954037f..c1e17ab 100644
--- a/js/data/feature_tree.js
+++ b/js/data/feature_tree.js
@@ -178,12 +178,14 @@ function getLineWidth(paint) {
 }
 
 FeatureTree.prototype.filterMatching = function(result, matching, match, queryGeometry, filter, filterLayerIDs, styleLayersByID, bearing, pixelsToTileUnits, returnGeoJSON) {
+    var previousIndex;
-    var seen = {};
     for (var k = 0; k < matching.length; k++) {
         var index = matching[k];
 
-        if (seen[index]) continue;
-        seen[index] = true;
+        // don't check the same feature more than once
+        if (index === previousIndex) continue;
+        previousIndex = index;
 
         match._setIndex(index);
```

js/data/feature_tree.js
lucaswoj: Oops! I was trying to emoji :+1: your comment, not delete it!

js/util/struct_array.js
ansis: What if you could pass a `Struct` as an optional second argument to `StructArray.get`. If a struct is provided it reuses it otherwise it creates a new one.

```js
var point;
for (var i = 0; i < points.length; i++) {
    point = points.get(i, point);
}
```

I'd need to check if the condition in `.get` and optional argument would have any perf impact. If it does we could have a separate method for reuse like this (but with a better name):

```js
var point = points.get(0);
for (var i = 0; i < points.length; i++) {
    points.getReuse(i, point);
}
```

I like the idea of calling something on the array instead of the struct.

js/data/feature_tree.js
lucaswoj: Ah! I understand now. I was just asking for my own understanding. 

If you think that diff :point_up_2: will work in all cases (incl GeoJSON) then let's do it.

js/data/feature_tree.js
lucaswoj: My worry is that `topDown`, as a symbol in file scope, is lacking contextual clues about what it is. `topDownFeatureComparator` or `compareFeatures` are a potentially good (potentially overkill) names.

js/util/struct_array.js
lucaswoj: Your comment makes me think I'm misunderstanding this method. My understanding is that it changes the position at which new elements will be inserted to allow an `ArrayBuffer` to be shared between `StructArrays`. Is that right?

`getReuse` smells like premature optimization to me. I've found v8 to be faster declaring new things than mutating existing things in some cases, probably because it can inline them or stack-allocate them instead of heap-allocate them.

js/data/feature_tree.js
ansis: done fc5c8eda832528e036f3c4f47d05cab9073252ff

CHANGELOG.md
ansis: done dcbf5917157f722e19c0605c6c644abb6c2265b1

js/data/bucket.js
ansis: renamed everything to array: 367bc5aa911dc5bf95b50885dda71ec6d56ec2ec

js/util/struct_array.js
ansis: > Your comment makes me think I'm misunderstanding this method. My understanding is that it changes the position at which new elements will be inserted to allow an ArrayBuffer to be shared between StructArrays. Is that right?

Nope. It's an odd method and its not documented clearly though
It's an alternative to `array.get` that doesn't create a new Struct object.

> getReuse smells like premature optimization to me. I've found v8 to be faster declaring new things than mutating existing things in some cases, probably because it can inline them or stack-allocate them instead of heap-allocate them.

It was added after profiling showed that it was significantly faster in `CollisionTile`. I'll double check that it still is.

lucaswoj: I understand now. Renaming to 'getReuse' sounds logical.

js/util/string_number_mapping.js
ansis: It's used for converting strings to numbers and back to strings. Kindof like a runtime enum.

It's used for one thing:
- converting a feature's sourceLayer's name into a number
- storing that number in a StructArray
- and then turning that number back into a string so that it can get that vectortile layer.

js/util/struct_array.js
ansis: fix in 040ea50806a347221faf86691e3e33fd3acb7996

It was originally done to match https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/BYTES_PER_ELEMENT but that's not a great idea because StructArray doesn't perfectly entire TypedArray interface and because StructArray types are defined on-the-fly.

js/util/struct_array.js
ansis: c53751f1434ca9c7cf6a9a75d3603db2c6a854de

js/util/struct_array.js
ansis: I didn't realize it was a protected symbol, I just capitalized it as if it were two words. `sizeOfType`? `sizeofType?`

js/util/vectortile_to_geojson.js
ansis: I have no strong preference either way. I don't feel weird about using `new` because it creates a js object that lazily generates geojson from a vectortile object that it holds a reference to. If it created a plain json-like object I'd feel weird.

js/source/worker_tile.js
ansis: `bucket.id` is a string. We could index buckets by a number. It gets trickier for async queries where we need to transfer everything back to workers. It helps if featureTree is self-contained and doesn't need buckets. I made it slightly cleaner here: fa0689f02b1acb5a338142026e1b05b2b0392a93

js/util/struct_array.js
ansis: I re-profiled it. It's faster but the difference is not significant if you're looking at more than just StructArray stuff. I removed setIndex completely in  5dc0b38. Thanks for flagging this

js/ui/map.js
ansis: I'm not seeing a great way of doing this now because some of that logic relies on `map.transform`.

ansis:
@lucaswoj I think I've addressed all your comments. Let me know if you see anything else!

js/util/struct_array.js
lucaswoj: Whoops. I was thinking of `typeof`. This is fine as-written.

js/util/vectortile_to_geojson.js
lucaswoj: I have a strong preference towards getting rid of `new` and using a plain function call like `convertVectorFeatureToGeoJSONFeature`

js/util/string_number_mapping.js
lucaswoj: I see. My confusion stemmed from this module's name. Brainstorming some alternates now...

lucaswoj: I suggest `DictionaryCoder` https://en.wikipedia.org/wiki/Dictionary_coder

mourner: I would be more confused with `DictionaryCoder` if I saw it in the code without seeing this discussion. Can't think of a better name than `StringNumberMapping`, but it definitely needs a clarifying comment / jsdoc.

ansis: From [wikipedia](https://en.wikipedia.org/wiki/Dictionary_coder), a dictionary coder is a 

> class of lossless data compression algorithms

Is there a noun for an instance of an algorithm?

`DictionaryCoder`, `DictionaryCoding`, `DictionaryCode`?
Or are they all the same thing?

`DictionaryCoder` sounds good enough to me.

lucaswoj: I think `DictionaryCoder` is a valid noun to use in this case :+1:

ansis: done in 8b52b55

lucaswoj:
I hereby give this pull request my :shipit:, pending a triage of #2320.

samanpwbb:
:tada:

Issue #2007, opened 2016-01-27
Use autogenerated production build for FPS benchmark


lucaswoj:
It is annoying to change this every release.

We [already do use an autogenerated build for the buffer benchmark](https://github.com/mapbox/mapbox-gl-js/blob/master/package.json#L64). Ideally we can merge this into the regular `npm start` command.

lucaswoj:
Fixed in #2231

Issue #954, opened 2015-01-26
performance test harness


jfirebaugh:
Questions about measuring relative performance of different rendering approaches come up frequently:

* [rendering layer-by-layer vs tile-by-tile](https://github.com/mapbox/mapbox-gl-js/issues/193)
* [tesselation strategy](https://github.com/mapbox/mapbox-gl-js/issues/682)
* [fill-outline vs separate layers](https://github.com/mapbox/mapbox-gl-style-spec/issues/223)

How can we accurately and repeatably measure rendering performance, so that we can make data-driven decisions about different possible rendering approaches?

Desiderata:

* Use real-world data
* Use real-world styles
* In browser
* Multiple browsers (Chrome + FF at least)
* Multiple OS/graphics card?
* Something automatable, that could be run in CI and flag performance regressions, would be a huge plus

ljbade:
A few thoughts:
- Doing this in a CI will be difficult unless you use mesa but the performance will be different to real GPU.
- Important consideration is desktop vs mobile GPUs which have very different architecture
- Real world data/styles - shouldn't be too hard, random selection of Mapbox Streets locations, perhaps weighted by popularity/number of views
- Perhaps could leverage the Android cloud testing services to test performance on a large variety of devices/GPUs
- We should add IE to the browser list since a lot of people/businesses still use it as only browser (if it is to be supported)
- Will want to have offline/fast cache of any network data for tiles/styles to remove influence of network speed

ansis:
This would be very useful. Do you have any ideas on how it should be implemented?

lucaswoj:
Related #1336

lucaswoj:
Most of the actionable ideas in this ticket have been done or have been ticketed out separately.

Issue #1362, opened 2015-07-06
Set up continuous benchmarking system


lucaswoj:
I'm not sure how we would implement this, but it'd be great to see the performance effect of each commit. 

https://github.com/travis-ci/travis-ci/issues/352

jfirebaugh:
Related: #954.

lucaswoj:
Related #1336

lucaswoj:
http://karma-runner.github.io/0.13/index.html looks promising, although we'd need to ensure it was running on a machine with a real graphics card (e.g. an AWS `g2.2xlarge` instance or a physical computer left running in a Mapbox office :grimacing:)

lucaswoj:
After a little digging it appears Karma is just smoke and mirrors around Selenium. We should use Selenium directly.

lucaswoj:
https://github.com/tmcw/testing-webgl

lucaswoj:
The new benchmark interface #2231 lays the groundwork for getting these tests running in an automated way. The big challenge is getting a GL-enabled WebDriver box running in the cloud somewhere. We might have to roll our own.

mourner:
I'd suggest closing this issue. Benchmarks on CI will not be representative of real world performance, and will fluctuate randomly — there's no way to create a controlled CI benchmarking environment for GL JS unless we set up our own servers for that, and this is out of scope for the nearest term.

PR #2231, opened 2016-03-08
Create standard benchmark interface

lucaswoj:
This pull request defines a standard benchmark interface.

Having a consistent interface for benchmarks allows us to use the same test runner and, in the future, collect benchmark results in an automated way.

I haven't yet moved the `fps` benchmark over to this new interface. Once I get some :eyes: and :+1: on the spec I'll tackle that. 

cc @ansis @mourner @jfirebaugh


jfirebaugh:
Looks good, I especially like capturing motivation and quality criteria in the README.

Do you think browser-based benchmarks are the way to go? Would we ever want headless / pure node benchmarks?

lucaswoj:
> Do you think browser-based benchmarks are the way to go? Would we ever want headless / pure node benchmarks?

We should explore headless and pure node benchmarks (especially #1362). The bench interface was designed to accommodate use in headless and pure node environments. For this PR, sticking to browser environments makes the most sense because it is simple and most representative of real-world usage.

ansis:
It would be good to document how access tokens can be set and provided

Other than that, this looks good!

lucaswoj:
One remaining task: I just realized that the env variables won't work for the debug page because it isn't run through browserify. I think it's time to combine the debug server and bench server into one.

lucaswoj:
Fixes #2007

mourner:
Still needs a few minor ESLint fixes.

lucaswoj:
Oops. I could've sworn I had fixed and pushed those.

:eyes:?

lucaswoj:
Fixes #954

PR #2233, opened 2016-03-09
Fix calls to setFilter, setLayoutProperty, and setLayerZoomRange on ref children

### LINKED ISSUE #2228: data not being filtered by setFilter() calls in v0.15.0
**mapbox-gl-js version**: 0.15.0

### Steps to Trigger Behavior
example case [here](https://jsfiddle.net/gjpeyewd/1/)
 1. load point dataset into studio 
 2. set above dataset as a layer on Basic map style
 3. try and filter said data using `map.setFilter()` in a GL script
 4. you can see the correct behavior if you switch to v. 0.14.0 of mapbox-gl-js in the jsfiddle

### Expected Behavior
dataset is filtered following invocation of `map.setFilter()`
### Actual Behavior
data remains unchanged

lucaswoj:
fixes #2228

looks like this bug was introduced in https://github.com/mapbox/mapbox-gl-js/pull/2174

cc @ansis @mourner @jfirebaugh @mollymerp

### DIFF
--- a/js/style/style_batch.js
+++ b/js/style/style_batch.js
@@ -118,6 +118,8 @@ styleBatch.prototype = {
 
     setLayoutProperty: function(layerId, name, value) {
         var layer = this._style.getReferentLayer(layerId);
+        layerId = layer.id;
+
         if (layer.getLayoutProperty(name) === value) return this;
 
         layer.setLayoutProperty(name, value);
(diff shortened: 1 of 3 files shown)


jfirebaugh:
:+1: 

Thanks for tracking this down.

mollymerp:
:clap: thanks Lucas!

PR #2234, opened 2016-03-09
adds #setMaxBounds and #removeMaxBounds methods

### LINKED ISSUE #2175: Add a Map#setMaxBounds method
Hi !

Any solution to update maps' maxBounds after a flyTo() or fitBounds() and fix the map to this specific area ?
Or a setMaxBounds(LngLatBounds) function which we could fire after a moveend event as an improvement ?

Thank you !

mollymerp:
Resolves #2175 

to do: 
- add test for removeMaxBounds method
- resolve error thrown by documentation.js
- maybe get rid of private method call to transform._constrain within the ui/map.js

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -74,9 +74,7 @@ var Map = module.exports = function(options) {
     this.transform = new Transform(options.minZoom, options.maxZoom);
 
     if (options.maxBounds) {
-        var b = LngLatBounds.convert(options.maxBounds);
-        this.transform.lngRange = [b.getWest(), b.getEast()];
-        this.transform.latRange = [b.getSouth(), b.getNorth()];
+        this.setMaxBounds(options.maxBounds);
     }
 
     util.bindAll([
@@ -269,7 +267,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
-     * Get the map's geographical bounds
+     * Get the map's geographical bounds.
      *
      * @returns {LngLatBounds}
      */
@@ -284,6 +282,29 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
         }
 
         return bounds;
+    },
+
+    /**
+     * Set constraint on the map's geographical bounds. Pan or zoom operations that would result in
+     * displaying regions that fall outside of the bounds instead result in displaying the map at the
+     * closest point and/or zoom level of the requested operation that is within the max bounds.
+     *
+     * @param {LngLatBounds | Array<Array<number>> | null | undefined} lnglatbounds Desired max bounds of the map. If null or undefined, function removes any bounds constraints on the map.
+     * @returns {Map} `this`
+     */
+    setMaxBounds: function (lnglatbounds) {
+        if (lnglatbounds) {
+            var b = LngLatBounds.convert(lnglatbounds);
+            this.transform.lngRange = [b.getWest(), b.getEast()];
+            this.transform.latRange = [b.getSouth(), b.getNorth()];
+            this.transform._constrain();
+            this._update();
+        } else if (lnglatbounds === null || lnglatbounds === undefined) {
+            this.transform.lngRange = [];
+            this.transform.latRange = [];
+            this._update();
+        }
+        return this;
     },
 
     /**
--- a/test/js/ui/map.test.js
+++ b/test/js/ui/map.test.js
@@ -292,6 +292,43 @@ test('Map', function(t) {
         }
     });
 
+    t.test('#setMaxBounds', function (t) {
+        t.test('constrains map bounds', function (t) {
+            var map = createMap({zoom:0});
+            map.setMaxBounds([[-130.4297, 50.0642], [-61.52344, 24.20688]]);
+            t.deepEqual(toFixed(map.getBounds().toArray()), toFixed([
+                [-112.5000192114, 24.2068800000],
+                [-79.4531207886, 50.0642000000]]));
+            t.end();
+        });
+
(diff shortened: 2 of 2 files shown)


lucaswoj:
This is looking great! Two minor things:

We don't allow "merge" commits in mapbox-gl-js (like d887e4b). There is a [section in `CONTRIBUTING.md` that describes our workflow](https://github.com/mapbox/mapbox-gl-js/blob/master/CONTRIBUTING.md#if-you-have-commit-access).

What led you to add a `removeMaxBounds` method rather than accepting falsey values to `setMaxBounds`?

mollymerp:
Oooh thanks for the heads up re: commit history. I may some help figuring out how to comply :blush: 

I honestly didn't consider using falsey values to un-set the bounds, but that is an interesting idea. I think there is clarity in having a separate method, but if we document it well we can avoid adding an unnecessary method :+1:

mollymerp:
I rebased with master and squashed the commit history. Anything else you'd like me to include/change for this PR? @lucaswoj @jfirebaugh :eyeglasses:

js/ui/map.js
```diff
@@ -287,6 +287,27 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
+     * Set constraint on the map's geographical bounds.
```
jfirebaugh: Can this be more specific? Like, "Pan or zoom operations that would result in displaying regions that fall outside of the bounds instead _____."

js/ui/map.js
```diff
@@ -287,6 +287,27 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+     *
+     * @param {LngLatBounds | Array<Array<number>>} lnglatbounds optional Desired max bounds of the map. If ommitted, function removes any bounds constraints on the map.
```
jfirebaugh: Instead of marking the parameter as optional, let's specify the type as `LngLatBounds | Array<Array<number>> | undefined` and encourage calling it as `setMaxBounds(undefined)`. The implementation stays the same, but `undefined` is a more explicit way of saying "unset the max bounds" than omitting the parameter.

lucaswoj: FWIW, we use `null` to unset paint and layout properties https://github.com/mapbox/mapbox-gl-js/blob/bench-interface/js/style/style_layer.js#L116-L117

js/ui/map.js
mollymerp: Good point. Does this work? I'm having trouble with the wording here.

> Set constraint on the map's geographical bounds. Pan or zoom operations that would result in displaying regions that fall outside of the bounds instead result in displaying the map at the full extent of the defined latlngbounds.

jfirebaugh: I'm not exactly even sure of the behavior. We may never have tried to precisely define it before. Is that description based on trying it out?

mollymerp: well for zoom operations this is the behavior. for panTo operations it just gets as close as it can to the requested point...

js/ui/map.js
```diff
@@ -287,6 +287,27 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
+     * Set constraint on the map's geographical bounds. Pan or zoom operations that would result in displaying regions that fall outside of the bounds instead result in displaying the map at the closest point and/or zoom level of the requested operation that is within the max bounds.
+     *
+     * @param {LngLatBounds | Array<Array<number>> | null | undefined} lnglatbounds optional Desired max bounds of the map. If null or undefined, function removes any bounds constraints on the map.
+     * @returns {Map} `this`
+     */
```
mollymerp: This is what I came up with for the documentation language. I also added explicit checks for null || undefined when un-setting the maxBounds.

lucaswoj: I'm happy with this wording! Would you mind wrapping the lines at 100 characters?

lucaswoj:
Except for the line wrap issue above, I'm feeling good about this. Lets :ship: tomorrow morning unless any more feedbac comes up.

lucaswoj:
One more thing: Can you change the [constructor to call `setMaxBounds` instead of duplicating its functionality](https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/map.js#L76-L80)?

js/ui/map.js
jfirebaugh: :+1: And also removing the word "optional", to further clarify that the intent is to explicitly pass `null` or `undefined`.

test/js/ui/map.test.js
```diff
@@ -292,6 +292,43 @@ test('Map', function(t) {
...
+            t.deepEqual(toFixed(map.getBounds().toArray()), toFixed([
+                [-112.5000192114, 24.2068800000],
+                [-79.4531207886, 50.0642000000]]));
+            t.end();
+        });
+
+        t.test('when no argument is passed, map bounds constraints are removed', function (t) {
+            var map = createMap({zoom:0});
+            map.setMaxBounds([[-130.4297, 50.0642], [-61.52344, 24.20688]]);
+            t.deepEqual(toFixed(map.setMaxBounds().setZoom(0).getBounds().toArray()), toFixed([
```
jfirebaugh: `setMaxBounds(undefined)` or `setMaxBounds(null)` (whichever's your preference).

mollymerp:
thank you for the feedback + suggestions @jfirebaugh @lucaswoj! I kept the parameter description on one line because that's the format of previous function documentation comments – let me know if I should wrap that one too. Just pushed the language/test changes up.

Issue #2237, opened 2016-03-10
Disabling and enabling the box zoom interaction messes it up


averas:
Discovered this while working on #2173 

**mapbox-gl-js 0.15.0**:
### Steps to Trigger Behavior
1. Wait for the map to load <code>map.on("load", ...)</code>
2. Disable box zoom <code>map.boxZoom.disable();</code>
3. Enable box zoom <code>map.boxZoom.enable();</code>
4. Perform a box zoom

Demonstration at http://codepen.io/anon/pen/LNNqrP
### Expected Behavior

That it works as before disable/enable.
### Actual Behavior

Map gets stuck dragging right after the box zoom, as if the mouse button is down even though it is not.

This may be related to OSX and have a similar root cause of #1888

averas:
I believe the problem here is that the current implementation of the interaction handlers depend on initialization order. When the handlers are initially enabled in the order defined in <code>interaction.js</code> their respective event listeners are also registered in "correct" order. However, later on, when <code>disable()</code> and <code>enable()</code> are invoked, the required event listener order chain is broken.

This could be adressed by making <code>enable()</code> and <code>disable()</code> "smarter" in how they register/unregister their listeners, or probably better, the overall interaction model has to be refactored to make initialization order inconsequential.

cc: @lucaswoj @bhousel

averas:
... another possible solution is actually to let disabled interaction handlers keep their event listeners registered, but ignore them.

averas:
@lucaswoj @jfirebaugh Any opinions on my last suggested option of letting listeners persist for disabled handlers but making them noops? Feels a bit like a workaround of the real problem, but at the same time, implementing arbitrary handler initialization order may be complicated and create more inter-handler-dependencies.

@mourner Any good experiences from how leaflet does it here?

mourner:
@averas in Leaflet, I don't recall any situations where interactions would depend on the order they were enabled. We need to figure out the core issue here. Why is it order-dependent in the first place?

averas:
> Why is it order-dependent in the first place?

In this particular case (there may be more), there is a check in <code>drag_pan.js</code> such as:

```
   _ignoreEvent: function (e) {
...
        if (map.boxZoom && map.boxZoom.isActive()) return true;
...
}
```

[...]

PR #2238, opened 2016-03-10
Expose enabled and active states on interaction handlers

### LINKED ISSUE #2173: Expose interaction handler enabled/disabled state
There's an issue on mapbox-gl-js-draw asking for disabling boxZoom temporarily when selecting features (https://github.com/mapbox/mapbox-gl-draw/issues/213), however, as far as I can tell there is no supported way to determine whether a certain interaction handler is enabled or not.

Even if it would be possible to always disable boxZoom temporarily when selecting features, to determine whether to enable it again requires knowledge of whether boxZoom was intentionally disabled overall.

Would it make sense to expand the API to expose the state of interaction handlers?

averas:
Fixes #2173.

While working on this I found this anomaly: #2237, please let me know if you think that one should be tracked down and addressed as part of this PR.

### DIFF
--- a/js/ui/handler/box_zoom.js
+++ b/js/ui/handler/box_zoom.js
@@ -10,6 +10,8 @@ module.exports = BoxZoomHandler;
  * The `BoxZoomHandler` allows a user to zoom the map to fit a bounding box.
  * The bounding box is defined by holding `shift` while dragging the cursor.
  * @class BoxZoomHandler
+ * @property {boolean} enabled Whether the "box zoom" interaction is currently enabled
+ * @property {boolean} active Whether the "box zoom" interaction is currently active
  */
 function BoxZoomHandler(map) {
     this._map = map;
@@ -21,6 +23,9 @@ function BoxZoomHandler(map) {
 
 BoxZoomHandler.prototype = {
 
+    enabled: false,
+    active: false,
+
     /**
      * Enable the "box zoom" interaction.
      * @example
@@ -29,6 +34,7 @@ BoxZoomHandler.prototype = {
     enable: function () {
         this.disable();
         this._el.addEventListener('mousedown', this._onMouseDown, false);
+        this.enabled = true;
     },
 
     /**
@@ -38,6 +44,7 @@ BoxZoomHandler.prototype = {
      */
     disable: function () {
         this._el.removeEventListener('mousedown', this._onMouseDown);
+        this.enabled = false;
     },
 
     _onMouseDown: function (e) {
--- a/js/ui/handler/box_zoom.js
+++ b/js/ui/handler/box_zoom.js
@@ -10,8 +10,6 @@ module.exports = BoxZoomHandler;
  * The `BoxZoomHandler` allows a user to zoom the map to fit a bounding box.
  * The bounding box is defined by holding `shift` while dragging the cursor.
  * @class BoxZoomHandler
- * @property {boolean} enabled Whether the "box zoom" interaction is currently enabled
- * @property {boolean} active Whether the "box zoom" interaction is currently active
  */
 function BoxZoomHandler(map) {
     this._map = map;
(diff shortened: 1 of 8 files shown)


js/ui/handler/box_zoom.js
```diff
@@ -21,6 +23,9 @@ function BoxZoomHandler(map) {
 
 BoxZoomHandler.prototype = {
 
+    enabled: false,
+    active: false,
```
lucaswoj: Are you using `active` anywhere?

averas: Some of the handlers, such as  <code>BoxZoomHandler</code>, already had a public (no underscore), albeit non-documented <code>active</code> property on them, I just took the opportunity to make code and behaviour consistent with the new enabled property.

lucaswoj: Got it! Thanks :smile:

js/ui/handler/keyboard.js
```diff
@@ -20,6 +20,7 @@ var panDelta = 80,
  *  * `Shift+⇡`: increase pitch by 5 degrees
  *  * `Shift+⇣`: decrease pitch by 5 degrees
  * @class KeyboardHandler
+ * @property {boolean} enabled Whether the keyboard interaction is currently enabled
```
lucaswoj: > Whether ~~the~~ keyboard interaction is currently enabled

lucaswoj:
Looks great! Let's give @bhousel a chance to look this over and then we can merge it.

averas:
Btw. @lucaswoj. There was a discussion on #2085 related to this. Obviously now that the handler has knowledge of whether it in fact is enabled or not the fallback of always disabling a handler before enabling it could be removed by just making <code>enable()</code> and <code>disable()</code> noops if the desired state is already present.

Let me know if you prefer this refactoring in contrast to the current fallback behaviour.

bhousel:
A problem with the previous code, but `active` should probably be an accessor not a public property.  For example in the dragpan handler, somebody could start panning, and if code elsewhere sets that public `active=false`, the handler will never run the things in `_onUp()`.

It's necessary for some handlers to know which other handlers are active, but no code outside the handler should be allowed to modify that state.

averas:
@bhousel That's true, and obviously an inherent problem with mutable properties. 

With the same reasoning I take it you are for keeping the <code>disable()</code> fallback call in the <code>enable()</code> functions, rather than making them noops based on current state (as outlined in my previous comment). Otherwise you may run into the same problem with the enable property, i.e. the property being mutated directly bringing the handler into a degenerative state.

lucaswoj:
# Next Steps

 - [ ] prefix `_active` and `_enabled` with underscores to mark them as private properties
 - [ ] add and document public `isActive` and `isEnabled` accessor methods
 - [ ] add a brief documentation note about the difference between "active" and "enabled"

Anything else we should add to this list @bhousel?

averas:
@lucaswoj The reason I went down the path of public properties in favour of accessors was @mourner's decision in #2173. Of course we can change that in accordance with your outline. Your suggested changes will be breaking, but I highly doubt anybody is currently using the non-documented <code>active</code> property.

averas:
@lucaswoj np. I'll provide a new set of commits...

averas:
Changes made according to @lucaswoj outline.

:eyes: @lucaswoj @bhousel

lucaswoj:
Thank you @averas! I'm going to test all the interaction handlers by hand and then I feel good about :ship: this!

 - [ ] KeyboardHandler
 - [ ] DragPanHandler
 - [ ] TouchZoomRotateHandler
 - [ ] ScrollZoomHandler
 - [ ] DoubleClickZoomHandler
 - [ ] DragRotateHandler

averas:
@lucaswoj Great! I made some additional comments on #2237 on that issue.

lucaswoj:
Merged in 494e2bb52be71d69d3a8f12904d3186824fb99a3! Thanks again @averas! :ship: :shipit: :rocket:

Issue #2235, opened 2016-03-09
Add .setMaxZoom/.setMinZoom method


peterqliu:
Analogous to https://github.com/mapbox/mapbox-gl-js/issues/2175, except that it allows setting the zoom bounds after the map is already declared.

peterqliu:
An alternate format could be `map.setZoomBounds({min:0, max:20})`

lucaswoj:
This would be awesome! 

It think our preferred API would be
```js
map.setZoomBounds(0, 20)
```

peterqliu:
@lucaswoj super naive stub [here](https://github.com/mapbox/mapbox-gl-js/commit/d8579e685f26256395500e89e419da7adac70e5c). Would appreciate some scope/context around this

lucaswoj:
@peterqliu Looks like you're on the right track!

# Next steps

 - [ ] turn this stub into a pull request 
 - [ ] take `minZoom` and `maxZoom` as separate parameters rather than an array
 - [ ]  document the parameters in the docblock comment above the method.
 - [ ] add a way to unset the zoom bounds too
 - [ ] add some simple unit tests for this method

Let me know if I can help in any way.

peterqliu:
:ship: https://github.com/mapbox/mapbox-gl-js/pull/2243

PR #2243, opened 2016-03-10
[not ready] setMinZoom() and setMaxZoom()

peterqliu:
Solves https://github.com/mapbox/mapbox-gl-js/issues/2235.

Per chat with @lucaswoj, splitting `min` and `max` bounds into separate methods would relieve developers from tracking the state of both bounds, in order to change either one.

### To dos

- [ ] Unit tests

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -21,6 +21,8 @@ var LngLatBounds = require('../geo/lng_lat_bounds');
 var Point = require('point-geometry');
 var Attribution = require('./control/attribution');
 
+var defaultMinZoom = 0;
+var defaultMaxZoom = 20;
 /**
  * Options common to Map#addClass, Map#removeClass, and Map#setClasses, controlling
  * whether or not to smoothly transition property changes triggered by the class change.
@@ -145,8 +147,8 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
         bearing: 0,
         pitch: 0,
 
-        minZoom: 0,
-        maxZoom: 20,
+        minZoom: defaultMinZoom,
+        maxZoom: defaultMaxZoom,
 
         interactive: true,
 
@@ -305,8 +307,49 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
             this._update();
         }
         return this;
+
+    },
+    /**
+     * Set the map's minimum zoom level, and zooms map to that level if it is currently below it. If no parameter provided, unsets the current minimum zoom (sets it to 0)
+     * @param {zoom} any number between 0 and 20
+     * @returns {Map} `this`
+     */
+    setMinZoom: function(minZoom) {
+
+        minZoom = minZoom === null || minZoom === undefined ? defaultMinZoom : minZoom;
+
+        if (minZoom >= defaultMinZoom && minZoom <= this.options.maxZoom) {
+            this.transform.minZoom = minZoom;
+            this._update();
+
+            if (this.getZoom() < minZoom) this.setZoom(minZoom);
+
+            return this;
+        }
+
+        else throw new Error('minZoom must be between ' + defaultMinZoom + ' and the current maxZoom, inclusive');
     },
 
+    /**
+     * Set the map's maximum zoom level, and zooms map to that level if it is currently above it. If no parameter provided, unsets the current maximum zoom (sets it to 20)
+     * @param {zoom} any number between 0 and 20
+     * @returns {Map} `this`
+     */
+    setMaxZoom: function(maxZoom) {
+
+        maxZoom = maxZoom === null || maxZoom === undefined ? defaultMaxZoom : maxZoom;
+
+        if (maxZoom >= this.options.minZoom && maxZoom <= defaultMaxZoom) {
+            this.transform.maxZoom = maxZoom;
+            this._update();
+
+            if (this.getZoom() > maxZoom) this.setZoom(maxZoom);
+
+            return this;
+        }
+
+        else throw new Error('maxZoom must be between the current minZoom and ' + defaultMaxZoom + ', inclusive');
+    },
     /**
      * Get pixel coordinates (relative to map container) given a geographical location
      *
--- a/test/js/ui/map.test.js
+++ b/test/js/ui/map.test.js
@@ -326,7 +326,58 @@ test('Map', function(t) {
                 [bounds[1][0].toFixed(n), bounds[1][1].toFixed(n)]
             ];
         }
+    });
 
+    t.test('#setMinZoom', function(t) {
+        var map = createMap({zoom:5});
+        map.setMinZoom(3.5);
+        map.setZoom(1);
+        t.equal(map.getZoom(), 3.5);
+        t.end();
+    });
+
+    t.test('unset minZoom', function(t) {
+        var map = createMap({minZoom:5});
+        map.setMinZoom(null);
+        map.setZoom(1);
+        t.equal(map.getZoom(), 1);
+        t.end();
+    });
+
+    t.test('ignore minZooms over maxZoom', function(t) {
+        var map = createMap({zoom:2, maxZoom:5});
+        t.throws(function() {
+            map.setMinZoom(6);
+        });
+        map.setZoom(0);
+        t.equal(map.getZoom(), 0);
+        t.end();
+    });
+
+    t.test('#setMaxZoom', function (t) {
+        var map = createMap({zoom:0});
+        map.setMaxZoom(3.5);
+        map.setZoom(4);
+        t.equal(map.getZoom(), 3.5);
+        t.end();
+    });
+
+    t.test('unset maxZoom', function(t) {
+        var map = createMap({maxZoom:5});
+        map.setMaxZoom(null);
+        map.setZoom(6);
+        t.equal(map.getZoom(), 6);
+        t.end();
+    });
+
+    t.test('ignore maxZooms over minZoom', function(t) {
+        var map = createMap({minZoom:5});
+        t.throws(function() {
+            map.setMaxZoom(4);
+        });
+        map.setZoom(5);
+        t.equal(map.getZoom(), 5);
+        t.end();
     });
 
     t.test('#remove', function(t) {


test/js/ui/map.test.js
```diff
@@ -292,6 +292,48 @@ test('Map', function(t) {
         }
     });
 
+    t.test('#setMinZoom', function (t) {
+        var map = createMap({zoom:0});
+        map.setMinZoom(3.5);
+        t.deepEqual(map.options.minZoom, 3.5);
+        t.end();
```
lucaswoj: Our convention is to wrap this in its own `t.test` block

```js
t.test('set minZoom', function(t) {
    var map = createMap({zoom:0});
    map.setMinZoom(3.5);
    t.deepEqual(map.options.minZoom, 3.5);
    t.end();
});
````

File: test/js/ui/map.test.js [hunk 2]
```diff
@@ -292,6 +292,48 @@ test('Map', function(t) {
...
+            t.deepEqual(map.options.minZoom, 0);
+            t.end();
+        })
+    });
+
+    t.test('#setMaxZoom', function (t) {
+        var map = createMap({zoom:0});
+        map.setMaxZoom(3.5);
+        t.deepEqual(map.options.maxZoom, 3.5);
+        t.end();
```
lucaswoj: Same as :point_up_2:

test/js/ui/map.test.js
```diff
@@ -292,6 +292,48 @@ test('Map', function(t) {
...
+            t.end();
+        })
+
+        t.test('ignore maxZooms over minZoom', function(t){
+            var map = createMap({minZoom:5});
+            map.setMaxZoom(4)
+            t.deepEqual(map.options.minZoom, 20);
+            t.end();
+        })
+    });
```
lucaswoj: Best practice for unit tests is to refrain from using undocumented/private members, like `map.options`. (We don't always follow this best practice, unfortunately.) This gives us more flexibility to rejigger the `Map` class without breaking unit tests and makes our tests stronger as they are more representative of real-world usage.

Can you think of a way to rewrite these tests using only documented/public methods?

peterqliu: hmm, I can't find a way to access the `options` object in the documentation... does this call for a `.getOptions()` method?

js/ui/map.js
```diff
@@ -287,6 +287,48 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
+     * Set the map's minimum zoom level, and zooms map to that level if it is currently below it. If no parameter provided, unsets the current minimum zoom (sets it to 0)
+     * @param {zoom} any number between 0 and 20
+     * @returns {Map} `this`
+     */
+    setMinZoom: function(minZoom) {
+
+        var zoom = minZoom === null ? 0 minZoom;
```
tmcw: syntax error here

js/ui/map.js
```diff
@@ -287,6 +287,48 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+        if (zoom <= this.options.maxZoom) {
+            this.options.minZoom = zoom;
+            this._update();
+
+            if (this.getZoom() < zoom) this.setZoom(zoom);
+
+            return this;
+        }
+
+        else throw('minZoom must be less than or equal to the current maxZoom')
```
tmcw: throw Error instances, not strings

test/js/ui/map.test.js
lucaswoj: I was thinking of using `getZoom` / `setZoom` to test the actual behavior of `Map`

js/ui/map.js
```diff
@@ -287,6 +287,48 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+    },
+
+    /**
+     * Set the map's maximum zoom level, and zooms map to that level if it is currently above it. If no parameter provided, unsets the current maximum zoom (sets it to 20)
+     * @param {zoom} any number between 0 and 20
+     * @returns {Map} `this`
+     */
+    setMaxZoom: function(maxZoom) {
+
+        var zoom = maxZoom === null ? 20 : maxZoom;
```
tmcw: make a constant for 20 so that it doesn't lag if someone changes it as the default for the map.

js/ui/map.js
tmcw: Also if the parameter isn't supplied, it will be `undefined`, not null.

js/ui/map.js
mourner: Why is it limited at all? 20 is the default maxZoom but I don't see any reason to make a hard unchangeable limit of 20. I would expect people to want higher zooms in some cases (e.g. for indoor mapping) and we don't want to force them to patch the code and rebuild.

peterqliu: @mourner right now if we instantiate a map without `maxZoom`, `map.options.maxZoom` returns 20

mourner: Sorry, I should stop reviewing code at midnight.

peterqliu: relatedly: should we rule out negative zoom levels?

js/ui/map.js
```diff
@@ -287,6 +287,50 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
+     * Set the map's minimum zoom level, and zooms map to that level if it is currently below it. If no parameter provided, unsets the current minimum zoom (sets it to 0)
+     * @param {zoom} any number between 0 and 20
+     * @returns {Map} `this`
+     */
+    setMinZoom: function(minZoom) {
+
+        var zoom = minZoom === undefined ? 0 : minZoom;
```
lucaswoj: Apologies for being a little late to this discussion.

 In the similar `setMapBounds` function, we decided on accepting either `null` or `undefined` values for this argument ([see the conversation](https://github.com/mapbox/mapbox-gl-js/pull/2234#discussion_r55750632)). 

To unset the `minZoom`, we want users to write
```js
map.setMinZoom(null) // clear and self-documenting
```

We do NOT want users to write
```js
map.setMinZoom() // wat?
```

js/ui/map.js
```diff
@@ -287,6 +287,50 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
+     * Set the map's minimum zoom level, and zooms map to that level if it is currently below it. If no parameter provided, unsets the current minimum zoom (sets it to 0)
```
lucaswoj: Please wrap this line at 100 chars

js/ui/map.js
```diff
@@ -287,6 +287,50 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+            if (this.getZoom() < zoom) this.setZoom(zoom);
+
+            return this;
+        }
+
+        else throw new Error('minZoom must be less than or equal to the current maxZoom')
+    },
+
+    /**
+     * Set the map's maximum zoom level, and zooms map to that level if it is currently above it. If no parameter provided, unsets the current maximum zoom (sets it to 20)
```
lucaswoj: :point_up_2:

js/ui/map.js
```diff
@@ -287,6 +287,50 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+     * @param {zoom} any number between 0 and 20
+     * @returns {Map} `this`
+     */
+    setMaxZoom: function(maxZoom) {
+
+        const defaultMax = this.options.maxZoom;
+
+        var zoom = maxZoom === undefined ? defaultMax : maxZoom;
```
lucaswoj: :point_up_2:

test/js/ui/map.test.js
```diff
@@ -292,6 +292,53 @@ test('Map', function(t) {
...
 
+    t.test('#setMinZoom', function(t) {
+        var map = createMap({zoom:0});
+        map.setMinZoom(3.5);
+        t.deepEqual(map.options.minZoom, 3.5);
+        t.end();
+    });
+
+    t.test('unset minZoom', function(t){
+        var map = createMap({minZoom:5});
```
lucaswoj: Let's pass a default `zoom: 5`  and assert that `t.deepEqual(map.getZoom(), 0);` to be 100% sure that this test is working the way that we expect

test/js/ui/map.test.js
```diff
@@ -292,6 +292,53 @@ test('Map', function(t) {
         }
     });
 
+    t.test('#setMinZoom', function(t) {
+        var map = createMap({zoom:0});
+        map.setMinZoom(3.5);
+        t.deepEqual(map.options.minZoom, 3.5);
```
lucaswoj: `deepEqual` is for testing the equality of arrays and objects. Since you're just testing a number, you can use `t.equal` here

js/ui/map.js
tmcw: To unset the minZoom, we want users to write

```js
map.setMinZoom(null) // clear and self-documenting
```

We do NOT want users to write

```js
map.setMinZoom() // wat?
```

I don't think this is anywhere near as clear a distinction as you lay out, way less than wat vs self-documenting. What about -Infinity? Or NaN? undefined vs null is not night and day.

If you want something clear, `map.removeMinZoom()`

lucaswoj: Agreed -- it's a compromise between clear and simple. More importantly, though, its consistent with the rest of our API.

js/ui/map.js
```diff
@@ -287,6 +287,50 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+    },
+
+    /**
+     * Set the map's maximum zoom level, and zooms map to that level if it is currently above it. If no parameter provided, unsets the current maximum zoom (sets it to 20)
+     * @param {zoom} any number between 0 and 20
+     * @returns {Map} `this`
+     */
+    setMaxZoom: function(maxZoom) {
+
+        const defaultMax = this.options.maxZoom;
```
tmcw: * why const here?
* please call this defaultMaxZoom, not defaultMax.
* this.options.maxZoom will change if someone customizes it, so this code will not work correctly, did you test it?

PR #2246, opened 2016-03-11
Don't delete empty buckets

### LINKED ISSUE #2244: filterEmptyBuckets breaks reloadSymbolData
https://github.com/mapbox/mapbox-gl-js/blob/352eb64e7a232c52919186c407845af0b951190d/js/source/worker_tile.js#L191
https://github.com/mapbox/mapbox-gl-js/blob/352eb64e7a232c52919186c407845af0b951190d/js/source/tile.js#L97

`oldBucket` is undefined if the bucket used to be empty and was filtered out by `filterEmptyBuckets`. Trying to call destroy on undefined will error.

If a bucket is filtered out in the current set of results but was present in the last one, it won't get removed from `this.buckets` and we will render stale symbols.

cc @lucaswoj

lucaswoj:
fixes #2244 

cc @ansis 

Still working on getting a good test case together (see #2244)

### DIFF
--- a/js/source/tile.js
+++ b/js/source/tile.js
@@ -85,18 +85,20 @@ Tile.prototype = {
      * @returns {undefined}
      * @private
      */
-    // TODO rewrite this
     reloadSymbolData: function(data, painter) {
         if (this.isUnloaded) return;
 
-        var newBuckets = unserializeBuckets(data.buckets);
-        for (var id in newBuckets) {
-            var newBucket = newBuckets[id];
-            var oldBucket = this.buckets[id];
-
-            oldBucket.destroy(painter.gl);
-            this.buckets[id] = newBucket;
+        // Destroy and delete existing symbol buckets
+        for (var id in this.buckets) {
+            var bucket = this.buckets[id];
+            if (bucket.type === 'symbol') {
+                bucket.destroy(painter.gl);
+                delete this.buckets[id];
+            }
         }
+
+        // Add new symbol buckets
+        util.extend(this.buckets, unserializeBuckets(data.buckets));
     },
 
     /**
--- a/js/source/worker_tile.js
+++ b/js/source/worker_tile.js
@@ -188,10 +188,8 @@ WorkerTile.prototype.parse = function(data, layers, actor, callback) {
             tile.redoPlacementAfterDone = false;
         }
 
-        buckets = filterEmptyBuckets(buckets);
-
         callback(null, {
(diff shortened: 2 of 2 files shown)


ansis:
Is this ok performance-wise? Does it increase the transfer/deserialization time significantly?

lucaswoj:
Sounds like its time to write a bucket transfer / deserialization benchmark to answer that question! #2231 

Why do you think this would affect deserialization or transfer time? The only change in this PR is that we don't mutate the persistent `buckets` variable when filtering empty buckets before transfer.

ansis:
Sorry, I misread the changes. You're right, this shouldn't affect the deserialization time.

I don't think this fixes the problem though.



### Case A (bucket doesn't get deleted)

- There is a style with a single symbol layer: cities.
- WorkerTile loads a tile and creates a `cities` bucket with a couple visible labels.
- `loadVectorData` gets `{ buckets: [ { layer: { id: 'cities', ... }, ... }], ...} `
- The map rotates and placement is redone by WorkerTile.
- At the new rotation the `cities` layer has no visible labels so the bucket gets filtered out.
- `reloadSymbolData` gets `{ buckets: [ ], ... }`.
- `reloadSymbolData` does not delete the old `cities` bucket.

### Case B (bucket doesn't exist)

- There is a style with a single symbol layer: cities.
- WorkerTile loads a tile and creates a `cities` bucket.
- At the current rotation the `cities` layer has no visible labels so the bucket gets filtered out.
- `loadVectorData` gets `{ buckets: [ ], ... }`.
- The map rotates and placement is redone by WorkerTile.
- At the new rotation the `cities` layer has some visible labels.
- `reloadSymbolData` gets `{ buckets: [ { layer: { id: 'cities', ... }, ... }], ...} `
- [oldBucket](https://github.com/mapbox/mapbox-gl-js/blob/352eb64e7a232c52919186c407845af0b951190d/js/source/tile.js#L95) is undefined. `oldBucket.destroy()` is called.

lucaswoj:
Thanks for the clarification @ansis. Does 75d2869 look like it'll do the trick? If so, I'll add some test cases and we can :ship:.

ansis:
Yep, 75d2869 looks good!

lucaswoj:
Ah. We [don't have any unit tests for `reloadSymbolData`](https://coveralls.io/builds/5412932/source?filename=js%2Fsource%2Ftile.js#L89). All coverage of `Tile` is due to test-suite but that doesn't give us the flexibility to trigger this behavior.

PR #2248, opened 2016-03-11
Add coveralls.io support

### LINKED ISSUE #1992: Add test coverage metrics

lucaswoj:
fixes #1992 

Pleasantly surprised to see we have 86% statement coverage. It'll be nice to track this going forward. 

The only downside to this PR is that it doubles CI build time (to ~10 minutes :grimacing:). I think we should merge this and try to make our CI build time faster in other ways.

cc @tmcw @ansis @jfirebaugh @bhousel @mourner


mourner:
Cool! The only thing I don't like is how test running time went from 5:20 to 11:30. :(

lucaswoj:
Per some offline chats, the consensus is "this is valuable information and makes our CI process slower." I think we should 

 - merge this as-is
 - try to improve the test-suite execution time in other ways
 - be open to reverting this if the slow CI builds become annoying

PR #2278, opened 2016-03-17
Fire map events for touchstart, touchend, touchmove, touchcancel

### LINKED ISSUE #2195: Touch events don't trigger mapbox events
### Context

* **mapbox-gl-js version**:  v0.14.3
* **devices**:  Nexus 5, iPhone 6s, MBP chrome device emulator

### Report

**Expected behavior**: Tapping map fires 'click' event. Same for doubletap: 'doubleclick',  touchstart: 'mousedown', touchend: 'mouseup'

**Actual behavior**: No events fired

**Steps to reproduce the behavior**:

1. Instantiate gl map
2. Add 'click' event listener to map
3. Touch map

bhousel:
(closes #2195)

### DIFF
--- a/js/util/browser/dom.js
+++ b/js/util/browser/dom.js
@@ -56,5 +56,18 @@ exports.mousePos = function (el, e) {
     e = e.touches ? e.touches[0] : e;
     return new Point(
         e.clientX - rect.left - el.clientLeft,
-        e.clientY - rect.top - el.clientTop);
+        e.clientY - rect.top - el.clientTop
+    );
+};
+
+exports.touchPos = function (el, e) {
+    var rect = el.getBoundingClientRect(),
+        points = [];
+    for (var i = 0; i < e.touches.length; i++) {
+        points.push(new Point(
+            e.touches[i].clientX - rect.left - el.clientLeft,
+            e.touches[i].clientY - rect.top - el.clientTop
+        ));
+    }
+    return points;
 };
--- a/js/ui/interaction.js
+++ b/js/ui/interaction.js
@@ -11,7 +11,8 @@ var handlers = {
 };
 
 var DOM = require('../util/dom'),
-    util = require('../util/util');
+    util = require('../util/util'),
+    Point = require('point-geometry');
 
 module.exports = Interaction;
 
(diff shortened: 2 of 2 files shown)


js/util/browser/dom.js
```diff
@@ -56,5 +56,18 @@ exports.mousePos = function (el, e) {
...
+exports.touchPos = function (el, e) {
+    var rect = el.getBoundingClientRect(),
+        points = [];
+    for (var i = 0; i < e.touches.length; i++) {
+        points.push(new Point(
+            e.touches[i].clientX - rect.left - el.clientLeft,
+            e.touches[i].clientY - rect.top - el.clientTop
+        ));
+    }
+    return points;
```
lucaswoj: What do you think about inlining this method into `_fireTouchEvent`? When might we reuse this code?

js/util/browser/dom.js
```diff
@@ -56,5 +56,18 @@ exports.mousePos = function (el, e) {
...
+
+exports.touchPos = function (el, e) {
+    var rect = el.getBoundingClientRect(),
+        points = [];
+    for (var i = 0; i < e.touches.length; i++) {
+        points.push(new Point(
+            e.touches[i].clientX - rect.left - el.clientLeft,
+            e.touches[i].clientY - rect.top - el.clientTop
+        ));
+    }
```
lucaswoj: You could simplify this code using `Array#map`.

lucaswoj:
This is solid code @bhousel :bow:

I left some minor suggestions. Merge this PR at your discretion. :ship:

js/util/browser/dom.js
bhousel: Nope!   [`TouchList`](https://developer.mozilla.org/en-US/docs/Web/API/TouchList) is one of those special DOM collections (like `NodeList`, `ClassList`, etc) that doesn't include all of the useful functions from `Array.prototype` that one would expect them to include.

js/util/browser/dom.js
bhousel: I'm not sure, but maybe anywhere that we use `DOM.mousePos` (like the boxzoom or touchpanrotate code) we might change to possibly support dispatching an event with multitouch properties.

js/util/browser/dom.js
lucaswoj: `Array.prototype.map.call(e.touches, function(touch) { ... })` :smiling_imp: 

Either way is ok :+1:

js/util/browser/dom.js
lucaswoj: Ah! I didn't know that this was a foil to another method called `mousePos`. I'm a sucker for consistency.

lucaswoj:
Looks great! :shipit:

bhousel:
merged!

PR #2280, opened 2016-03-17
Cleanup debugging code

bhousel:
Included:
* Remove canvas atlas texture debug code (closes #2261)
* Rename "debug" -> "tileDebug"

PR #2284, opened 2016-03-17
Rename debug settings

lucaswoj:
@bhousel raised some good points about the vocabulary we use for debug settings in #2280 

If we're going to make a breaking change to this API, I want to make certain that we are changing to something we are keen on supporting _forever_. 

As with any public API, I want to choose a name that is

 - short
 - grounded in well-known domain terminology
 - self documenting (i.e. a smart user could figure out what it does without checking the docs)

This PR renames

 - `collisionDebug` --> `showCollisionBoxes`
 - `debug` / `tileDebug` --> `showTileBoundaries`

cc @bhousel @mourner @bhousel @ansis

### DIFF
--- a/bench/buffer_benchmark.js
+++ b/bench/buffer_benchmark.js
@@ -143,7 +143,7 @@ function runSample(stylesheet, getGlyphs, getIcons, getTile, callback) {
             overscaling: 1,
             angle: 0,
             pitch: 0,
-            collisionDebug: false,
+            showCollisionBoxes: false,
             source: 'composite',
             uid: url
         });
--- a/debug/index.html
+++ b/debug/index.html
@@ -39,8 +39,8 @@
 <body>
 <div id='map'></div>
 <div id='checkboxes'>
-  <input id='tile-debug-checkbox' name='tile-debug' type='checkbox'> <label for='tile-debug'>tile debug</label><br />
-  <input id='collision-debug-checkbox' name='collision-debug' type='checkbox'> <label for='collision-debug'>collision debug</label><br />
+  <input id='show-tile-boundaries-checkbox' name='show-tile-boundaries' type='checkbox'> <label for='show-tile-boundaries'>tile debug</label><br />
+  <input id='show-symbol-collision-boxes-checkbox' name='show-symbol-collision-boxes' type='checkbox'> <label for='show-symbol-collision-boxes'>collision debug</label><br />
   <input id='buffer-checkbox' name='buffer' type='checkbox'> <label for='buffer'>buffer stats</label>
 </div>
 
--- a/debug/site.js
+++ b/debug/site.js
@@ -82,12 +82,12 @@ map.on('click', function(e) {
         .addTo(map);
 });
 
-document.getElementById('tile-debug-checkbox').onclick = function() {
-    map.tileDebug = !!this.checked;
+document.getElementById('show-tile-boundaries-checkbox').onclick = function() {
+    map.showTileBoundaries = !!this.checked;
 };
 
-document.getElementById('collision-debug-checkbox').onclick = function() {
-    map.collisionDebug = !!this.checked;
+document.getElementById('show-symbol-collision-boxes-checkbox').onclick = function() {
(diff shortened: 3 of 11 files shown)


jfirebaugh:
:+1: 

"Placement" is the term we use in the style spec, so `showSymbolPlacementBoxes`?

lucaswoj:
> `symbol-placement`
> _Optional enum.  One of point, line. Defaults to point._
> Label **placement** relative to its geometry. line can only be used on LineStrings and Polygons.

> `symbol-avoid-edges`
> _Optional boolean.  Defaults to false._
> If true, the symbols will not cross tile edges to avoid mutual **collisions**. Recommended in layers that don’t have enough padding in the vector tile to prevent **collisions**, or if it is a point symbol layer placed after a line symbol layer.

I'm not sure "placement" is a synonym for "collision" in our spec. Placement seems to refer to the location of a symbol relative to its source geometry. Collision seems to refer to the location of a symbol relative to other symbols.

ansis:
> I'm not sure "placement" is a synonym for "collision" in our spec. Placement seems to refer to the location of a symbol relative to its source geometry. Collision seems to refer to the location of a symbol relative to other symbols.

I agree. 

The `-ignore-placement` property uses the word "placement" in confusing way but I think we should get rid of that property: https://github.com/mapbox/mapbox-gl-style-spec/issues/429

bhousel:
I'm fine with leaving them as is.  We don't need to have a discussion on what to name 3 debug flags.

To give us more flexibility to change these later, it might make sense to roll them into an options hash like

```js
debugOptions = { 
  showBoundingBoxes: true,
  showTileBoundaries: true,
  showBufferStats: true
}
```

And I would also support wrapping the ui into a control that can be added to any mapbox-gl-js map if useful.

lucaswoj:
> I'm fine with leaving them as is. We don't need to have a discussion on what to name 3 debug flags.

`tileDebug` and `collisionDebug` are public APIs to which we're considering making a breaking change. We need to have a brief discussion to ensure we're doing this right and there will never be another breaking change to this API. 

> To give us more flexibility to change these later, it might make sense to roll them into an options hash like
> 
> ```
> debugOptions = { 
>  showBoundingBoxes: true,
>  showTileBoundaries: true,
>  showBufferStats: true
>}
>```

In what sense does this give us more flexibility?

> And I would also support wrapping the ui into a control that can be added to any mapbox-gl-js map if useful.

This is a separate issue. I encourage you to open a ticket to discuss. :smile: My feeling is that these aren't used often enough to require a UI control, especially by users uncomfortable using the console.

bhousel:
> In what sense does this give us more flexibility?

Because then the API just becomes:

```js
map.setDebugOptions(opts)
map.getDebugOptions()
```

And we don't need to change it again or bother downstream users with a semver bump when we want to add things. 

from your original comment:
>If we're going to make a breaking change to this API, I want to make certain that we are changing to something we are keen on supporting forever.

lucaswoj:
It'd still be a breaking change if we changed an option name in `setDebugOptions`

bhousel:
> It'd still be a breaking change if we changed an option name in setDebugOptions

Really?  I don't think of semver as working that way..

jfirebaugh:
I don't feel strongly about the exact name, just that we should pick something that folks who do care are happy with. This is a name that's used by Mapbox Studio and possibly other library consumers, and I'd like to make those developers' lives easier with a single name change. So I defer to @lucaswoj and @ansis.

mourner:
@lucaswoj let's merge this. Maybe also rename `showSymbolCollisionBoxes` to `showCollisionBoxes`? Shorter and collision only happens for symbol layers anyway.

lucaswoj:
> Maybe also rename `showSymbolCollisionBoxes` to `showCollisionBoxes`

:+1: sounds good

Will merge on green.

1ec5:
Great change – these names happen to match the debug settings exposed by the Mapbox iOS SDK.

lucaswoj:
@1ec5 what a happy accident :joy:

PR #1606, opened 2015-10-08
Switch to tesselated fill rendering with earcut

mourner:
A fresh recut of #948. Had to squish into one commit because all previous branches were a big mess. cc @jfirebaugh @ansis @kkaefer 

fixes #2269
fixes #682

PR #2285, opened 2016-03-17
fix winding order of geojson before rendering

### LINKED ISSUE #2089: Fix winding order for GeoJSON polygons before rendering
In [mapbox-gl-draw](https://github.com/mapbox/mapbox-gl-draw/issues/128) we're seeing a problem where holes are rendered in polygons when we don't expect them to be. At first I thought that this was due to poor winding order management in gl-draw (which is true) but after talking to @sgillies about this a bit today I'm pretty sure mapbox-gl-js is implementing geojson holes incorrectly.

In the above image each new polygon is a its own feature. [According to the spec](https://github.com/geojson/draft-geojson/blob/master/middle.mkd#polygon) holes happen when a second linear ring is provided to the array of linear ring coordinate arrays.

I'm guessing I can fix this in gl-draw by following the `a linear ring SHOULD follow right-hand rule` command, but still this should be fixed so end users don't get caught up in trying to check if their linear rights follow the right hand rule.

ansis:
fixes #2089

:eyes: @mourner or @bhousel or @lucaswoj


lucaswoj:
LGTM, but lets wait for @mourner's :+1:

mourner:
:ship:, but let's remove it once #1606 lands.

PR #2286, opened 2016-03-17
Remove Map#options

lucaswoj:
An little internal refactor to get rid of `Map#options` by moving its properties to top-level class properties.

cc @peterqliu @ansis @jfirebaugh @bhousel

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -71,10 +71,15 @@ var defaultMaxZoom = 20;
 var Map = module.exports = function(options) {
 
     options = util.inherit(this.options, options);
-    this._container = options.container;
     this._interactive = options.interactive;
     this._failIfMajorPerformanceCaveat = options.failIfMajorPerformanceCaveat;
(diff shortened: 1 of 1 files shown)


mourner:
But why? This doesn't look clearer to me.

lucaswoj:
Two main reasons:

1. the `options` object has no semantic meaning after `Map` is constructed, it just serves as a vehicle for named arguments
2. the `options` object contains duplicate and unused state which can be confusing for developers working on the class

@peterqliu and I hit 2 when working on #2235 together. It felt important enough to justify the 5 minutes it took to write this PR.

js/ui/map.js
```diff
@@ -70,7 +70,11 @@ var defaultMaxZoom = 20;
  */
 var Map = module.exports = function(options) {
 
-    options = this.options = util.inherit(this.options, options);
+    options = util.inherit(this.options, options);
+    this._container = options.container;
```
mourner: This should handle the id case so that the meaning `this._container` doesn't change depending on whether `_setupContainer` was called or not.

PR #2287, opened 2016-03-18
don't check cache when looking up a parent raster for fading

ansis:
[this](https://github.com/mapbox/mapbox-gl-js/blob/e2b0560802605f6805b1b666716f5a8137075b21/js/render/draw_raster.js#L49) used to check the cache for parent tiles and call `addTile` if it found one, and add unneeded tiles to `pyramid._tiles`.

:eyes: @lucaswoj

### DIFF
--- a/js/source/tile_pyramid.js
+++ b/js/source/tile_pyramid.js
@@ -190,10 +190,11 @@ TilePyramid.prototype = {
      * @param {Coordinate} coord
      * @param {number} minCoveringZoom
      * @param {boolean} retain
+     * @param {boolean} checkCache
      * @returns {Tile} tile object
      * @private
      */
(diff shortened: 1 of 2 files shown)


lucaswoj:
OOof. I see why this is unnecessary but this further adds to the spaghetti-code-ness of `TilePyramid`.

lucaswoj:
This is :ship: as is, but I'm going to ticket out some follow-up refactoring of `TilePyramid`

lucaswoj:
@ansis Why is this marked as "do not merge"?

ansis:
I'd like to look at a better way of doing this before merging. This isn't critical performance-wise

lucaswoj:
Closing as stale.

PR #2307, opened 2016-03-21
bind textures to correct texture units

### LINKED ISSUE #2304: Raster hillshading + complex vector fills creates kaleidoscope effect.
**mapbox-gl-js version**: 0.15.0 (also tested on local master build current to to [b26c036](https://github.com/mapbox/mapbox-gl-js/commit/b26c0367dd5d57e3f855d59b4a0542fb7b970340) to see if [2236](https://github.com/mapbox/mapbox-gl-js/issues/2236) fixed this issue. 

**[Example](https://jsbin.com/vuqerolama/1/edit?html,output)**. I've added buttons to turn on and off the complex vectors so you can see how they interact with the hill shading.

### Steps to Trigger Behavior

 1. Make a map that uses both raster hillshading and complex vector fill patterns.
 2. Zoom in on an area covered by a complex vector pattern.
 3. Try a few times because this does not happen 100% of the time

### Expected Behavior
Map zooms and styles as you'd expect.
screen shot 2016-03-21 at 1 57 54 pm

### Actual Behavior
At larger zooms (> 12) the raster layer switches to a strange kaleidoscope effect and moves above all the vector layers. There are also some other less common artifacts like flashing or hill shading and ground cover disappearing in areas that are covered in a vector pattern after you zoom out again.

I've observed this on multiple browsers and hardware configurations.

screen shot 2016-03-21 at 1 55 18 pm

ansis:
fix #2304

Raster rendering sets the active texture unit to gl.TEXTURE1 which broke all later texture binds. This fixes that by setting the correct active texture before each bind.

:eyes: @bhousel or @lucaswoj

### DIFF
--- a/js/render/draw_background.js
+++ b/js/render/draw_background.js
@@ -26,7 +26,6 @@ function drawBackground(painter, source, layer) {
         // Draw texture fill
         shader = painter.patternShader;
         gl.switchShader(shader);
-        gl.uniform1i(shader.u_image, 0);
         gl.uniform2fv(shader.u_pattern_tl_a, imagePosA.tl);
         gl.uniform2fv(shader.u_pattern_br_a, imagePosA.br);
         gl.uniform2fv(shader.u_pattern_tl_b, imagePosB.tl);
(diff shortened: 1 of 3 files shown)


bhousel:
LGTM :+1:

Issue #1714, opened 2015-11-13
Figure out how to make disabled attributes performant in IE


lucaswoj:
`uniforms` allow us to specify global values to WebGL shaders. `attribute`s allow us to specify per-feature values. 

If all features have the same value (i.e. all circles are `red`), we can disable the attributes using `gl.disableVertexAttribArray` and set a global value as if it were a `uniform`.

We converted most data-driven styling related `uniform`s to `attribute`s in https://github.com/mapbox/mapbox-gl-js/pull/1257. Then, there were reports of big performance regressions in IE https://github.com/mapbox/mapbox-gl-js/issues/1336, so we reverted our changes in https://github.com/mapbox/mapbox-gl-js/pull/1415

One of the big remaining challenges for implementing data driven styling is figuring out how to make disabled attributes performant in IE. 

cc @jfirebaugh @tmcw @ansis

ansis:
> Aside from using array buffers to feed attribute data to the vertex shader, OpenGL also supports attributes which remain constant for all vertices used in a draw call. Direct3D 9 doesn’t have a similar concept, at least not explicitly.

> Constant attributes are implementated using separate (static) vertex buffers, and uses a stride of 0 to ensure that every vertex retrieves the same data.

https://chromium.googlesource.com/angle/angle/+/master/doc/BufferImplementation.md#Constant-attributes

I haven't read through the ANGLE source yet to see what they are doing but we might be able to improve performance by managing this buffer ourselves since we know the values earlier.

ansis:
IE doesn't use ANGLE but I think it has an internal translation to Direct3D so it probably has a similar problem.

We can't manually manage a buffer with a stride of 0 because in webgl a 0 stride means "use a stride that matches the type and size". I don't think we can make it read the same buffer value for each vertex from webgl.

If that's the case, I think the only way we can improve performance on our side is by generating a new shader for each unique combination of constant vs enabled attributes. There are many possible combinations but in practice there probably won't be that many different combinations in a single style. We'd use glsl preprocessor directives to implement the variations in the shaders. When the style is loaded or changed we'd calculate which shader variations are needed and then compile them by prepending some `#define`s to the shader source.

We could use this approach only for IE/Windows and use constant attributes everywhere else, but we should strongly consider using this approach everywhere so that we don't have different implementations and bugs on different platforms.

PR #2310, opened 2016-03-22
Create declarative "useProgram" interface

lucaswoj:
This pull request includes

 - refactoring to make shader source rewriting for https://github.com/mapbox/mapbox-gl-js/issues/1714 possible 
 - renaming "shaders" to "programs" where appropriate to stay true to the domain terminology
 - merging `gl_util` and `painter`

I recommend going through this commit-by-commit to compartmentalize the renaming noise.

cc @ansis @mourner @jfirebaugh


lucaswoj:
The high level intent here is to get rid of the notions of programs as singleton objects and move to a declarative `useProgram` interface, which can accept parameters in the future.

ansis:
looks good to me!

Issue #1550, opened 2015-09-24
Test interaction handlers, UI controls, and browser utilities


jfirebaugh:
Figure out how to test our event handling and interaction modules.

jfirebaugh:
Possible approaches:

* JSDom. Pro: likely less unpredictability / hard-to-debug timing issues than a browser environment. Con: likely lower fidelity event implementation.
* Reintroduce real-browser test suite using smokestack. Pro: higher fidelity. Con: https://twitter.com/tmcw/status/646345048142032896
* Selenium?

lucaswoj:
<img width="512" alt="screen_shot_2016-03-10_at_5 47 58_pm_1024" src="https://cloud.githubusercontent.com/assets/281306/13691252/35b37bae-e6ec-11e5-9e19-c0f32001d46f.png">

anandthakker:
> JSDom. Pro: likely less unpredictability / hard-to-debug timing issues than a browser environment. Con: likely lower fidelity event implementation.

At this point in time, I would vote for this particular tradeoff.  I think it would be the quickest way to get started and I'm betting a large majority of these tests would not require a super high fidelity event environment.  (E.g. #2972 )

lucaswoj:
+ browser utilities (i.e. `js/util/browser/*.js`) 😄

jfirebaugh:
We've been gradually filling these out.

PR #2312, opened 2016-03-22
fix behavior of arrow keys when focused on a map container

### LINKED ISSUE #2302: Up-down keybindings don't preventDefault, so the page still scrolls
**mapbox-gl-js version**:

### Steps to Trigger Behavior

 1. Click on a map to focus it
 2. Hit up and down arrow keys
 3. The page scrolls

### Expected Behavior

The page should not scroll: only the map should move

### Actual Behavior

The page scrolls.

mollymerp:
fixes #2302 
I still need to add tests, and I'm not sure how to test user interactions like this so I will be looking into that!

### DIFF
--- a/js/ui/handler/keyboard.js
+++ b/js/ui/handler/keyboard.js
@@ -86,6 +86,7 @@ KeyboardHandler.prototype = {
             if (e.shiftKey) {
                 map.easeTo({ bearing: map.getBearing() - rotateDelta }, eventData);
             } else {
+                e.preventDefault();
                 map.panBy([-panDelta, 0], eventData);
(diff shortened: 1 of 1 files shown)


lucaswoj:
> I'm not sure how to test user interactions like this so I will be looking into that!

Neither are we! See #1550.

lucaswoj:
You'll need to rebase this branch on `master` to fix CI.

tmcw:
Did this get signoff before merging?

lucaswoj:
@tmcw Yes, I gave a verbal :+1:

PR #2314, opened 2016-03-22
Switch from prova to tape

### LINKED ISSUE #2313: Replace prova with another test runner
Prova was unpublished from npm just now :skull: 

Unless this is a temporary glitch, we'll need to find a long-term replacement for ~~npm~~ Prova

 - https://www.npmjs.com/package/prova
 - https://github.com/azer/prova

lucaswoj:
cc @mourner @jfirebaugh @ansis 

fixes #2313


package.json
```diff
@@ -68,7 +69,7 @@
     "start": "node server.js",
     "start-docs": "npm run production && npm run docs && jekyll serve -w",
     "lint": "eslint js test bench && documentation --lint",
-    "test": "npm run lint && istanbul test --include-pid --report none --print none prova -- test/js/*/*.js -q",
+    "test": "npm run lint && istanbul test --include-pid --report none --print none tape -- test/js/*/*.js | tap-dot",
```
jfirebaugh: I've had consistently bad experiences with piping tap output to some other process. It invariably breaks crash logging, `console.log`, or something else.

Can we use tap and one of its built-in reporters instead?

lucaswoj:
Update: I'm working on a transition to `tap` but there is some niggly manual refactoring of that needs to happen first.

lucaswoj:
In the latest commit, I switched to tap. I'm pleasantly surprised by how modern and stable tap is. It identified a number of problems with our test suite that prova was silently ignoring.

jfirebaugh:
:+1: on code changes.

Any idea why coverage dropped over 10%? Maybe after the prova ⇢ tap switch it's counting only unit tests, or only render tests?

lucaswoj:
Istanbul isn't playing nicely with tap. Tap has a built-in test runner but it is way too magic for us to use. Still brainstorming a solution.

mourner:
> Istanbul isn't playing nicely with tap.

Maybe try `nyc`? There may be some interprocess stuff going on.

ci.sh
```diff
@@ -3,35 +3,40 @@
...
 ulimit -c unlimited -S
 echo 'ulimit -c: '`ulimit -c`
 echo '/proc/sys/kernel/core_pattern: '`cat /proc/sys/kernel/core_pattern`
 sysctl kernel.core_pattern
 
-npm run test-suite
-RESULT=$?
+# run render tests
+istanbul cover --dir .nyc_output --include-pid --report none --print none test/render.test.js &&
+istanbul cover --dir .nyc_output --include-pid --report none --print none test/query.test.js
```
mourner: can be replaced with nyc?

lucaswoj: `nyc` is a wrapper around `istanbul`

The `nyc` CLI doesn't support our use case of running multiple test suites and aggregating their results. We'd have to do more file-copying or folder-renaming gymnastics to make it work.

PR #2319, opened 2016-03-23
fix fill slivers between adjacent polygons

### LINKED ISSUE #2060: Slivers appear between adjacent non-opaque polygon-fills as pitch increases
Experienced on OS.X, Chrome.

mapbox-gl-js 0.12.2:

mapbox-gl-js 0.12.3-0.13.1:

Please note that this only happens when fill-opacity is set below 1.

### LINKED ISSUE #2079: remove * 10 lineWidth
It blames back to the big layer-by-layer commit by me. I don't remember why I added it, but I don't think it should be there. 

https://github.com/mapbox/mapbox-gl-js/blob/da6a3928814778649ac2461a812bc6ce6a88c6fb/js/render/draw_fill.js#L27

ansis:
fix #2060
fix #2079

Pixels in neighbouring polygons were being dropped by antialiasing outlines that had been written to the depth buffer. This fixes it by not writing antialiasing outlines to the depth buffer.

The `numSubLayers` change in da4dfd3657240752204e5eee766a64e18eb960ce made the bug harder to reproduce but it still existed.

:eyes: @bhousel or @lucaswoj 

cc @averas

### DIFF
--- a/js/render/draw_fill.js
+++ b/js/render/draw_fill.js
@@ -25,6 +25,7 @@ function draw(painter, source, layer, coords) {
     if (!painter.isOpaquePass && layer.paint['fill-antialias'] && !(layer.paint['fill-pattern'] && !strokeColor)) {
         gl.switchShader(painter.outlineShader);
         gl.lineWidth(2);
+        painter.depthMask(false);
 
         if (strokeColor) {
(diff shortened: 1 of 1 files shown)


lucaswoj:
Beautiful. :ship: when ready.

PR #2327, opened 2016-03-24
Make benchmarks navigable and add fps tests for setData

mcwhittemore:
@lucaswoj & @ansis here is a benchmark showing fps degrading as the zoom level approaches 22 when setting data on a geojson source.

Would love feedback on how to make this better judge performance. FPS doesn't capture what is rendered on the canvas that well for instance.


mcwhittemore:
Is this good to merge?

lucaswoj:
Not yet. I'll give this a review later today.

bench/benchmarks/geojson_rendering.js
```diff
@@ -0,0 +1,127 @@
...
+
+    map.addSource('geojson', {
+      data: featureCollection,
+      type: 'geojson'
+    });
+
+    map.addLayer({
+      'source': 'geojson',
+      'id': 'geojson-polygon-fill'
+      type: 'fill',
```
lucaswoj: There is a syntax error here (no `,` after `'id': 'geojson-polygon-fill'`).

My apologies, I broke the linter in https://github.com/mapbox/mapbox-gl-js/pull/2314 and fixed in #2347. Can you rebase on `master` and fix the lint errors?

bench/index.html
```diff
@@ -25,6 +30,13 @@
 </head>
 
 <body class="prose">
+    <div id="benchmarks">
+        <a href='/bench/buffer' class='button'>Buffer</a>
+        <a href='/bench/fps' class='button'>FPS</a>
+        <a href='/bench/query-box' class='button'>Query Box</a>
+        <a href='/bench/query-point' class='button'>Query Point</a>
+        <a href='/bench/geojson-rendering' class='button'>GeoJSON Rendering</a>
+    </div>
```
lucaswoj: I'm :-1: on a manually-maintained list of benchmarks. We should either generate this automatically or omit it altogether.

lucaswoj: Generating it from the [`benchmarks` object in `bench/index.js`](https://github.com/mapbox/mapbox-gl-js/pull/2327/files#diff-4c5f2be87a037cceed511291a012024aL16) would be good.

bench/benchmarks/geojson_rendering.js
```diff
@@ -0,0 +1,127 @@
...
+        [0, 0+idx],
+        [0-idx, 0]
+      ]],
+      type: 'Polygon'
+    }
+  });
+}
+
+function runTest(map, cb) {
+  var numFrames = 0;
```
lucaswoj: We should extract `measureFramerate` function from the fps benchmark into a reusable utility. This will ensure our measurement methodology is consistent (and simplify this codepath).

lucaswoj:
Do you think we could generalize our existing fps benchmark to capture this perf regression? (i.e. by measuring FPS over the set of tiles specified in `coords.js`)

Having fewer & robuster benchmarks will help us find future performance regressions.

mcwhittemore:
@lucaswoj - Looking into this more, what I had feels a bit past contrived. Limiting this PR to a simple restructuring of the benchmarks as I look into this some more.

mcwhittemore:
@mourner or @lucaswoj for the review

bench/benchmarks/fps.js
```diff
@@ -0,0 +1,100 @@
...
+            });
+
+            measureFramerate(DURATION_MILLISECONDS, function(err, fps) {
+                running = false;
+                cb(err, fps);
+            });
+        };
+
+        map.repaint = true;
+        measureFramerate(DURATION_MILLISECONDS, function(err, fps) {
```
lucaswoj: Why did you choose fps to measure `setData` performance? Can we measure tile creation time directly?

lucaswoj: Ah! I just saw where you wrote:

> This adds to setData tests to the fps bench as well per chat. One for a large feature collection, for a small feature collection.

So sorry for misspeaking. I didn't mean to imply that we should test `setData` within the FPS benchmark. These should be two separate benchmarks. 

 - the `fps` benchmark tests the fps across a variety of streets-v8 viewports (i.e. those in `coords.js`)
 - the `set-data` benchmark tests the time from a `GeoJSONSource#setData` call to the data being rendered on screen for different sized GeoJSONs

bench/benchmarks/fps.js
```diff
@@ -0,0 +1,100 @@
...
+                testSetData(largeFeatureCollection, function(err, fps) {
+                    if (err) return evented.fire('error', {error: err});
+                    evented.fire('log', {message: 'setData (large) - ' + formatNumber(fps)});
+                });
+            });
+        });
+    });
+
+    return evented;
+};
```
lucaswoj: Make sure you're adhering to the benchmark spec, defined in [bench/README.md](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/README.md#end). This was spec designed to allow us to automate benchmark runs in the near future.

 If the benchmark doesn't have a "single" end score because you're measuring several different things, you can break the benchmark up into multiple benchmarks.

package.json
```diff
@@ -68,6 +68,7 @@
   },
   "scripts": {
     "start": "node server.js",
+    "postinstall": "mkdir -p ./bench/data && curl -s https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_50m_land.geojson > ./bench/data/naturalearth-land.json",
```
lucaswoj: We should not download megabytes of fixtures for every `mapbox-gl` install. This data is only needed for developers.

bench/benchmarks/fps.js
mcwhittemore: I'm struggling to reconcile

> you can break the benchmark up into multiple benchmarks

with

> Having fewer & robuster benchmarks will help us find future performance regressions.

Are we ok with emitting `end` multiple times?

I think these two tests should be in the same benchmark as seeing them both at the same time will be helpful when making a changes. An average of their scores like `query-point` doesn't make much sense though.

bench/benchmarks/fps.js
mcwhittemore: Cool. Breaking these up. I put them together as seeing the normal fps was helpful, but I agree this limits the FPS benchmark from getter better.

bench/benchmarks/fps.js
lucaswoj: I agree that it would be helpful to see the results of multiple benchmarks on the same page at the same time. That is a presentation issue. We could achieve this modifying the benchmark runner to run multiple benchmarks.

Having simple benchmarks that measure exactly one thing in a statistically significant way is more important than having fewer benchmarks. 

I want the ability to run these benchmarks and compare their results in an automated way. Having a single score per benchmark makes those comparisons possible.

mourner: One benchmark may affect the other and their order can change results significantly. The more simple/isolated a benchmark is the better.

mcwhittemore:
@lucaswoj - updated the PR to break setData and FPS apart. Also changed setData to track the average time it takes to load all the tiles after a setData. This should be a better metric for setData perf than fps, though time to load tiles could go down and fps be negatively impacted.

I don't think we want to break this test into two files per: `Having fewer & robuster benchmarks will help us find future performance regressions` So I'm using two ends which feels dirty. To be honest, I think we should worry about the best way to do this kind of tracking when we're writing the automation benchmark code. Right now we're trying to find what is best for our own usage. We can confirm automation to that.

lucaswoj:
Having two scores in this benchmark is a non-starter for me. Breaking this up is very easy. You are measuring two separate things.

mcwhittemore:
@lucaswoj updated to be two tests.

Issue #2334, opened 2016-03-25
draggable point example is slow


ansis:
**mapbox-gl-js version**: v0.16.0

### Steps to Trigger Behavior

 1. Drag point in draggable point example http://localhost:4000/mapbox-gl-js/example/drag-a-point/

### Expected Behavior

It should be fast and smooth.

### Actual Behavior

It's slow and choppy.

bsudekum:
I actually can't even drag the marker...

ansis:
> I actually can't even drag the marker...

It's broken but fixed by https://github.com/mapbox/mapbox-gl-js/pull/2333

mourner:
As @mcwhittemore points out in #2327, its lagginess depends on how much you're zoomed in. Investigating.

mourner:
OK, spent some time investigating this and I found at least two issues that cause most of the lagginess:

1. The `mousedown` handler needs `e.originalEvent.preventDefault()`. I assume this prevents Chrome (and possibly other browsers) from triggering some native drag-handling code that conflicts with point dragging.
2. As of https://github.com/mapbox/mapbox-gl-js/commit/1f9029cce00b230ec37f9b7792916f03fc191231 by @jfirebaugh, `Map` `setPaintProperty` (which we use to change point color on hover) calls `batch`, which in turn always calls `style._cascade`, which is very expensive to do on every `mousemove`. We should update the code to only cascade on paint property change if we change pseudo-paint properties like `text-size` (see #1451).

The dragging becomes much smoother if you fix both (not one or the other). 

I'm also convinced that it will get much smoother if mousemove is throttled — should we do this internally for the `Map` `mousemove` event? It's a very common gotcha.

mourner:
Fixing the second point can also improve Studio responsiveness considerably (cc @samanpwbb), so we should prioritize this. It may be pretty tricky though, considering upcoming DDS changes, after which cascading may be required by paint properties other than `text-size` and `icon-size`. How can we tell if a paint property needs cascade or not, without hardcoding? cc @lucaswoj

jfirebaugh:
Even before 1f9029cce00b230ec37f9b7792916f03fc191231 `Map#setPaintProperty` called `_cascade`, and I don't think "call `_cascade` only for `text-size`" is the right condition. Many (most) other cases need `_cascade` as well -- basically, any time you change a property that should have a visual effect, you need to cascade. The only exception would be changing the property of a class that's not currently active.

[...]

PR #2337, opened 2016-03-25
Throttle mousemove internally

mourner:
Ref #2334. I’ve lost count of times where I gave advice to throttle `mousemove` whenever someone had performance issues with interactivity. I think `mousemove` should be throttled internally by default — it _always_ helps with performance, and I can’t think of any drawbacks/gotchas in this particular throttling case.

This, coupled with the preventDefault fix in the example, makes the point dragging example much smoother. After we fix `setPaintProperty` triggering `cascade`, it’ll become butter-smooth.

cc @lucaswoj @jfirebaugh @ansis @tristen

### DIFF
--- a/js/ui/interaction.js
+++ b/js/ui/interaction.js
@@ -25,6 +25,8 @@ function Interaction(map) {
     }
 
     util.bindHandlers(this);
+
+    this._onMouseMove = util.throttle(this._onMouseMove, 16);
 }
 
 Interaction.prototype = {


ansis:
Are there any cases where a user would want unthrottled events?

Does this solve the problems or just make them less severe? Could the work triggered by the events still cause backups?

Would it be better to throttle the actual work instead of the events that trigger the work? (should `setData` queue up new data if work is still in progress?)

If we throttle, we should throttle with `requestAnimationFrame` instead of `setTimeout` 16ms.
- frames might take more than 16ms
- chrome aggressively de-prioritizes `setTimeout` in some cases: https://github.com/mapbox/mapbox-gl-js/issues/2139 https://code.google.com/p/chromium/issues/detail?id=567800

mourner:
> Are there any cases where a user would want unthrottled events?

I can't think of any, but let's see if anyone can.

> Does this solve the problems or just make them less severe? Could the work triggered by the events still cause backups?

It's a part of the solution, which is "avoiding unnecessary work". It can't solve all the problems but it always improves performance in case expensive mousemove handlers.

> Would it be better to throttle the actual work instead of the events that trigger the work? (should setData queue up new data if work is still in progress?)

I don't think so, because throttling the actual work can have unexpected side effects and introduce subtle bugs. But specifically in case of mouse events, I can't think of a case where it would cause problems.

> If we throttle, we should throttle with requestAnimationFrame instead of setTimeout 16ms.

Agreed. Should we modify `util.throttle` to automatically use `reqAnimFrame` when `time` is 16ms or less, or should we do a separate util method in `util/browser/browser.js`?

lucaswoj:
This looks great! 

:+1: to a separate `frameThrottle` util method in `util/browser/browser.js`.

mcwhittemore:
Won't this break `stopPropagation`?

jfirebaugh:
Yes, and `preventDefault`.

I'm :-1: here. We've seen several instances where internal throttling in Leaflet caused issues and had to remove it. It just generally seems to be a bad idea to do it at the library level.

mourner:
But you're strongly against exposing util methods like `throttle` in public too. And we can't use third-party utility libraries in our examples. What should we do here then? I'd really like to make our examples fast.

Issue #1907, opened 2016-01-07
Avoiding selection of content in Chrome when shift-zooming


averas:
This may not be quite related to mapbox-gl-js, but I hope I'd get some pointers from all the clever people around here.

I have a page with a map with panels next to it, and when the focus of the page is on any of these panels while I perform a shift-bbox-zoom the content on these panels is selected. Obviously one ugly hack would be to start applying <code>user-select: none</code> to some of these areas, but there are actually some areas I do want to be selectable, just not as a result of shift-zoom.

This appears to be a problem in Chrome, but not Safari or Firefox (tested in OSX).

You can examine this behaviour on the codepen below. Click somewhere in the text on the panel and then perform a shift-zoom.

http://codepen.io/anon/pen/adWvOO

bhousel:
We should probably prevent default handling of the events we capture in `interaction.js` and the other handlers.  I can't think of any good reason _not_ to.

But as a workaround until we do that, you can add this to your codepen:

```javascript
map.on('mousedown', function(e) {
  e.originalEvent.preventDefault(true);
});
```

jfirebaugh:
The hard-won knowledge from d3 and Leaflet is that preventDefault shouldn't be used on mousedown because it prevents desired browser-native behaviors. For example, the default behavior of mousedown switches focus out of a focused input textarea. Likewise [preventDefault on mousedown disallows the capture of mouseup outside of an iframe](http://bl.ocks.org/mbostock/109620145f090cda1a7d).

Instead we'll need to [prevent text selection in other ways](https://github.com/Leaflet/Leaflet/blob/161172bbb92e60d2d7fc21173da7e58fc725dd2b/src/dom/DomUtil.js#L192-L217).

jfirebaugh:
In fact [gl-js already has such code](https://github.com/mapbox/mapbox-gl-js/blob/75aab6d39fa4d221db3c2b2e82df39f816287c5e/js/util/browser/dom.js#L22-L34).

averas:
@jfirebaugh Which only needs to be applied to shift-zooming as well?

bhousel:
> Which only needs to be applied to shift-zooming as well?

We have a `DOM.disableDrag();` call in `box_zoom`, but it seems to happen on mousemove, which is too late.  The initial shift-mousedown is what extends the text selection in the sidebar.

PR #2338, opened 2016-03-25
Prevent default on mousedown

mourner:
See #2334. Along with #2337, makes point dragging example much smoother.

### DIFF
--- a/js/ui/interaction.js
+++ b/js/ui/interaction.js
@@ -72,6 +72,9 @@ Interaction.prototype = {
         this._map.stop();
         this._startPos = DOM.mousePos(this._el, e);
         this._fireMouseEvent('mousedown', e);
+
+        // prevent default container dragging behavior
+        e.preventDefault();
     },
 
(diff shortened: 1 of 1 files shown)


mourner:
cc @bhousel @lucaswoj

bhousel:
eh @jfirebaugh was :-1: on this:  https://github.com/mapbox/mapbox-gl-js/issues/1907#issuecomment-171386026

mourner:
Ah, you're right. Let's try to find an alternative fix.

bhousel:
The slowness could be from `getBoundingClientRect` in `DOM.mousePos` triggering layout thrashing.  I know we previously considered this as WONTFIX, but I'll profile the handlers some more this week and, if that's what it is, find a way to cache the container dimensions.

mourner:
@bhousel `getBoundingClientRect` would show up noticeably in the profiler, so I believe this is something else. Need to try `DOM.disableDrag` too...

PR #2355, opened 2016-03-29
Automatically batch style updates

### LINKED ISSUE #2343: remove `map.batch` and automatically batch style changes
I think we should automatically batch all style changes that happen between frames. Are there any cases where we wouldn't want to batch style changes?

If changes are automatically batched we can remove `map.batch`.

mourner:
Closes #2343. Removes `map.batch` and implements automatic batching of style changes, applying them once per frame. The style keeps track of all mutations done through `Style` `setLayoutProperty`, `addLayer`, etc., and then all mutations are applied with the `style.update` method which is called in `map._render` and happens once per frame.

Remaining tasks:

- [x] ~~fix API accessor methods to immediately reflect style changes before they're batch-updated, as per~~ they do already https://github.com/mapbox/mapbox-gl-js/issues/2343#issuecomment-202458808
- [x] figure out why `sprites-pattern` render tests failed
- [x] renames for clarity as suggested in https://github.com/mapbox/mapbox-gl-js/pull/2355#discussion_r57767349
- [x] update examples, make sure they all work properly
- [x] :eyes: @jfirebaugh @lucaswoj @ansis @scothis and subsequent clean up


js/style/style.js
```diff
@@ -247,19 +251,95 @@ Style.prototype = util.inherit(Evented, {
...
+
     /**
-     * Apply multiple style mutations in a batch
-     * @param {function} work Function which accepts the StyleBatch interface
+     * Apply queued style mutations in a batch
      * @private
      */
-    batch: function(work) {
-        styleBatch(this, work);
+    update: function(classes, options) {
```
lucaswoj: The terminology here is starting to feel overwhelming: `mutate`, `update`, `change`, `cascade`, `recalculate`...

How do you feel about picking a consistent verb (i.e. `update`, `change`) and differentiating these properties and methods with direct objects?

 - `cascade` -> `updateTransitions`
 - `recalculate` -> `updateZoom`
 - `mutations` -> `updates`
 - `broadcastLayers` -> `updateWorkerLayers`
 - `reloadSource` -> `updateSource`
 - `change` event -> `update`

mourner: Yes, great point! I like all the suggested renames except the `change` event, which we use throughout the code for other classes, and it's generally more common to have events like this called `change` rather than `update` in JS APIs.

mourner: I'd also leave `reloadSource` -> `updateSource` rename for another PR because it's tricky, potentially involving a lot of changes that are out of scope for this PR (e.g. it calls `source.reload`, which in turn calls `pyramid.reload`, but `pyramid` also has `update` method which is different, so you also have to rethink pyramid API).

mourner: I've renamed `mutations`, `broadcastLayers` and did some other minor cleanups/renames, but leaving the rest for future PRs (especially `cascade` and `recalculate`) so that it doesn't interfere with current DDS PRs, otherwise we would have a lot of conflicts to resolve.

test/js/ui/map.test.js
```diff
@@ -103,7 +98,7 @@ test('Map', function(t) {
             style.fire('source.error');
             style.fire('source.change');
             style.fire('tile.add');
-            style.fire('tile.load');
+            // style.fire('tile.load');
```
mourner: This is commented out because I discovered that the event isn't actually fired (but that wasn't caught by the test because it fires two change events, the second one after the sprite is loaded). We should fix this test in a separate PR (not trivial because of the asyncness).

lucaswoj: Commenting out code is bad practice. 

Can we delete this line and do one of

 - accept that this event isn't fired 
 - open an issue about improving our even testing 
 - improve our event testing in this PR

js/style/style.js
lucaswoj: :+1: I openend https://github.com/mapbox/mapbox-gl-js/issues/2359

js/style/style.js
```diff
@@ -247,19 +251,92 @@ Style.prototype = util.inherit(Evented, {
...
+        } else if (this._updates.changed) {
+            this.fire('change');
+        }
+
+        this._resetMutations();
+
+        return this;
+    },
+
+    _resetMutations: function() {
```
lucaswoj: `_resetMutations` --> `_resetUpdates`

lucaswoj:
This is ready to :ship: once you address the two minor comments above. 

Great work! I love a negative line PR.

test/js/ui/map.test.js
mourner: :+1:, removing the comments, opened https://github.com/mapbox/mapbox-gl-js/issues/2360

PR #2363, opened 2016-03-30
Make setPaintProperty a noop if given the same value

### LINKED ISSUE #2362: setPaintProperty should be a noop if given the same value
Unlike all other style setters, `setPaintProperty` doesn't check whether it's given the same value that it already has, so it can do unnecessary work. Ref https://github.com/mapbox/mapbox-gl-js/issues/2334#issuecomment-202500663

mourner:
Closes #2362. :eyes: @lucaswoj

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -551,7 +551,7 @@ Style.prototype = util.inherit(Evented, {
         var layer = this.getReferentLayer(layerId);
         layerId = layer.id;
 
-        if (layer.getLayoutProperty(name) === value) return this;
+        if (util.deepEqual(layer.getLayoutProperty(name), value)) return this;
 
         layer.setLayoutProperty(name, value);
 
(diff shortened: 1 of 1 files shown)


js/style/style.js
```diff
@@ -580,7 +580,11 @@ Style.prototype = util.inherit(Evented, {
     setPaintProperty: function(layerId, name, value, klass) {
         this._checkLoaded();
 
-        this.getLayer(layerId).setPaintProperty(name, value, klass);
+        var layer = this.getLayer(layerId);
+
+        if (layer.getPaintProperty(name, klass) === value) return this;
```
lucaswoj: Do you think it's worth doing a `deepEqual` comparison instead? `value` can be an object (function) or an array.

mourner: Good point. It's a rare case but it's free to cover it.

PR #2365, opened 2016-03-30
Faster render tests

mourner:
These changes cut down the render tests time on Travis by about 3 minutes. cc @jfirebaugh @lucaswoj


jfirebaugh:
:+1:

PR #2369, opened 2016-03-30
Upgrade eslint and some other deps

mourner:
:eyes: @lucaswoj. 

The `es6` environment in `.eslintrc` had to be added because typed arrays are a part of ES6 spec and default environment is now ES5 which doesn’t include them.

update: [found an explanation](https://github.com/eslint/eslint/blob/master/docs/user-guide/migrating-to-2.0.0.md#built-in-global-variables).


lucaswoj:
LGTM :shipit:

PR #2371, opened 2016-03-31
Add assertions to DictionaryCoder

lucaswoj:
I hit a hard-to-debug error caused by `DictionaryCoder` silently returning `undefined` for keys not in its lookup table. 

Do these `assert` statements make sense, @ansis?


ansis:
Yep, they make sense

Issue #2080, opened 2016-02-06
antialias fills without gl.LINES


ansis:
The minimum `ALIASED_LINE_WIDTH_RANGE` required by the spec is `[1, 1]`. We need at least a width of 2 for antialiasing to work. [ANGLE](https://en.wikipedia.org/wiki/ANGLE_(software)) only [supports](https://bugs.chromium.org/p/angleproject/issues/detail?id=119#c17) a width of 1. Angle is used in Chrome and Firefox on windows, and by our headless testing.

We should use triangles.

moving away from gl.LINES should fix https://github.com/mapbox/mapbox-gl-js/issues/1047

It could also help without our headless testing antialiasing headaches.

ansis:
cc @kkaefer

lucaswoj:
Do you think this is the same phenomena as https://github.com/mapbox/mapbox-gl-js/pull/1606#issuecomment-179494096?

tmpsantos:
`ANGLE` is also used on Mapbox GL Native + Qt on Windows.

jfirebaugh:
The core profile for desktop OpenGL on macOS supports only `[1, 1]` as well: https://developer.apple.com/opengl/OpenGL-Capabilities-Tables.pdf.

jfirebaugh:
Weirdly, support for >1 seems to have _increased_ abrupt in recent months:

![image](https://user-images.githubusercontent.com/98601/27562503-a9ea5664-5a81-11e7-88fa-46ec8399a58f.png)

http://webglstats.com/webgl/parameter/ALIASED_LINE_WIDTH_RANGE

jfirebaugh:
@kkaefer pointed out a potential [technique for doing polygon antialiasing/stroking using barycentric coordinates](https://stackoverflow.com/questions/18035719/drawing-a-border-on-a-2d-polygon-with-a-fragment-shader), and I spent some time prototyping it in gl-js. It's a clever technique, but unfortunately it doesn't look like it will work for us. The technique originates as a way of drawing wireframes, where all three sides of each triangle are stroked. When you want to stroke only some of the sides, the issue arises that the edges that are not stroked end up "cutting off" pieces of the stroke on other edges:

![image](https://user-images.githubusercontent.com/98601/28282730-5178df54-6ae0-11e7-8c89-67dd9e703d37.png)

The skinnier the triangles, the worse this gets -- and earcut tends to generate a lot of very skinny triangles.

[...]

PR #2372, opened 2016-03-31
fix anti-aliasing for fill-patterns

mollymerp:
a quick & dirty first attempt at resolving the problem with anti-aliasing when there is a fill-pattern – this code doesn't throw errors but also doesn't solve the problem. :neutral_face:

### DIFF
--- a/js/render/draw_fill.js
+++ b/js/render/draw_fill.js
@@ -23,32 +23,48 @@ function draw(painter, source, layer, coords) {
     }
 
     // Draw stroke
-    if (!painter.isOpaquePass && layer.paint['fill-antialias'] && !(layer.paint['fill-pattern'] && !strokeColor)) {
-        var outlineProgram = painter.useProgram('outline');
-        painter.lineWidth(2);
-        painter.depthMask(false);
+    if (!painter.isOpaquePass && layer.paint['fill-antialias']) {
+        if (strokeColor || !layer.paint['fill-pattern']) {
+            var outlineProgram = painter.useProgram('outline');
+            painter.lineWidth(2);
+            painter.depthMask(false);
 
-        if (strokeColor) {
-            // If we defined a different color for the fill outline, we are
-            // going to ignore the bits in 0x07 and just care about the global
-            // clipping mask.
-            painter.setDepthSublayer(2);
+            if (strokeColor) {
+                // If we defined a different color for the fill outline, we are
+                // going to ignore the bits in 0x07 and just care about the global
+                // clipping mask.
+                painter.setDepthSublayer(2);
+            } else {
+                // Otherwise, we only want to drawFill the antialiased parts that are
+                // *outside* the current shape. This is important in case the fill
+                // or stroke color is translucent. If we wouldn't clip to outside
+                // the current shape, some pixels from the outline stroke overlapped
+                // the (non-antialiased) fill.
+                painter.setDepthSublayer(0);
+            }
+            gl.uniform2f(outlineProgram.u_world, gl.drawingBufferWidth, gl.drawingBufferHeight);
+            gl.uniform4fv(outlineProgram.u_color, strokeColor ? strokeColor : color);
 
+            for (var j = 0; j < coords.length; j++) {
+                drawStroke(painter, source, layer, coords[j]);
+            }
         } else {
+            var outlinePatternProgram = painter.useProgram('outlinepattern');
+            painter.lineWidth(2);
+            painter.depthMask(false);
             // Otherwise, we only want to drawFill the antialiased parts that are
             // *outside* the current shape. This is important in case the fill
             // or stroke color is translucent. If we wouldn't clip to outside
             // the current shape, some pixels from the outline stroke overlapped
             // the (non-antialiased) fill.
             painter.setDepthSublayer(0);
+            gl.uniform2f(outlinePatternProgram.u_world, gl.drawingBufferWidth, gl.drawingBufferHeight);
+
+            for (var k = 0; k < coords.length; k++) {
+                drawStroke(painter, source, layer, coords[k]);
+            }
         }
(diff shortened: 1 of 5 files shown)


mollymerp:
looks like the anti-aliasing is fixed, but I still need to refactor and clean the code up, as well as adjust sensitivity for the test-suite (because as you can see they're still passing with this fix) and run a few more test cases. 

_fill pattern test on master branch_
screen shot 2016-03-31 at 11 07 06 am

_fill pattern test on this branch_
screen shot 2016-03-31 at 11 07 27 am

shaders/outlinepattern.fragment.glsl
```diff
@@ -21,5 +21,13 @@ void main() {
     vec2 pos2 = mix(u_pattern_tl_b, u_pattern_br_b, imagecoord_b);
     vec4 color2 = texture2D(u_image, pos2);
 
-    gl_FragColor = mix(color1, color2, u_mix) * u_opacity;
+    // find distance to outline for alpha interpolation
+    float dist_a = length(v_pos_a - gl_FragCoord.xy);
```
ansis: In order to fade the pixel appropriately we need the distance of the pixel from the center of the line. `gl_FragCoord.xy` is the position of the pixel which we need to subtract from the position of the closest point on the line.

`v_pos_a` is the position of texture pixel we need. I think we need another `varying v_pos` that has the position of the nearest point on the line like in the outline shader.

shaders/outlinepattern.fragment.glsl
```diff
@@ -21,5 +21,13 @@ void main() {
...
+    float alpha_a = smoothstep(1.0, 0.0, dist_a);
+
+    float dist_b = length(v_pos_b - gl_FragCoord.xy);
+    float alpha_b = smoothstep(1.0, 0.0, dist_b);
```
ansis: The `alpha` value here should be based on how far the pixel is from the center of the line. This distance is the same for both patterns so I think we only need one `alpha`.

shaders/outlinepattern.fragment.glsl
```diff
@@ -21,5 +21,13 @@ void main() {
...
+    
+
+    gl_FragColor = mix(color1, color2, u_mix) * mix(mix(alpha_a, alpha_b, u_mix), u_opacity, 0.5);
```
ansis: `alpha` and `u_opacity` should be combined with multiplication instead of `mix`. If the pattern opacity is already 0.5 and the alpha based on distance from the line is 0.5 the combined opacity should be 0.25

mollymerp:
@ansis pointed out that I was using incorrect variables to calculate the alpha value before. I have updated the shaders accordingly. the new changes cause 2 of the tests to fail, so I will have to update the test-suite once we confirm the output is what we expect.
some updated screen shots:
_master branch_
screen shot 2016-03-31 at 2 34 00 pm

_this branch_
screen shot 2016-03-31 at 2 33 28 pm

ansis:
The rendering in 7daa739 looks right to me.

The render test rendering is a bit off but that is a completely separate issue. The headless renderer we use only supports a gl.lineWidth of 1: https://github.com/mapbox/mapbox-gl-js/issues/2080 It should be fine to update the render tests with the current rendering.

I guess the remaining things are:
- updating render tests
- reducing duplication in `draw_fill.js`

js/render/draw_fill.js
```diff
@@ -49,6 +50,39 @@ function draw(painter, source, layer, coords) {
         for (var j = 0; j < coords.length; j++) {
             drawStroke(painter, source, layer, coords[j]);
         }
+    } else if (layer.paint['fill-pattern']) {
```
mollymerp: is this the correct conditional to have here? I was thinking ` else if (layer.paint['fill-pattern'] && !strokeColor)` might make sense? do we even need another conditional at all or would simply `else` suffice? I'm not super familiar with the possible permutations of the features checked in L27.

ansis: `drawFill` is called twice for each layer each frame: once for the opaque pass and once for the translucent pass. The opaque pass is only used to draw completely opaque geometry. All antialiasing strokes are translucent so they need to be drawn when `painter.isOpaquePass` is false.

No antialiasing should be drawn if `layer.paint['fill-antialias']` is false.

```
if (!painter.isOpaquePass && layer.paint['fill-antialias']) {
    if (strokeColor || !layer.paint['fill-pattern']) {
        // draw outline with either strokeColor or the fill color
    } else {
        // drawn the new patterned outline
    }
}
```

mollymerp:
PR made in [mapbox-gl-test-suite](https://github.com/mapbox/mapbox-gl-test-suite/pull/83) to update render tests

js/render/draw_fill.js
```diff
@@ -221,3 +201,55 @@ function drawStroke(painter, source, layer, coord) {
         gl.drawElements(gl.LINES, count, gl.UNSIGNED_SHORT, elementOffset);
     }
 }
+
+
+function imageOffset(image, opacity, tile, coord, painter, program) {
```
ansis: can this be renamed to something with a verb like `usePattern` or `setPattern`?

ansis:
This looks great to me! the remaining things I see are
 - [renaming imageOffset](https://github.com/mapbox/mapbox-gl-js/pull/2372#discussion_r58257755)
 - updating `mapbox-gl-test-suite` in package.json to your updated version (which should fix the failing tests)

debug/site.js
```diff
@@ -1,7 +1,7 @@
 
 mapboxgl.accessToken = getAccessToken();
 
-var map = new mapboxgl.Map({
+ window.map = new mapboxgl.Map({
```
mourner: leading space

peterqliu:
:clap:

PR #2376, opened 2016-03-31
better symbol fading with texture lookups

### LINKED ISSUE #590: Fade flicker when zooming back and forth
If you zoom in and then zoom out (or vise versa) before labels finished fading in, they'll flicker in a weird way.  It's very noticeable when you start to pan & zoom around a lot, and it doesn't look like a tile switching issue. @ansis 

fade-flicker

ansis:
fix #590

I revived a [branch from August 2014](https://github.com/mapbox/mapbox-gl-js/issues/590#issuecomment-49822040). Remember this @mourner? This branch uses fragment texture lookups instead of vertex texture lookups to avoid concerns about vertex texture support.

- this is simpler than predicting opacity based on current zooming speed
- this is smoother: symbols don't flicker when changing zoom speed or when zooming in
  and out rapidly.

:eyes: @mourner @lucaswoj

### DIFF
--- a/js/render/draw_symbol.js
+++ b/js/render/draw_symbol.js
@@ -143,11 +143,9 @@ function drawSymbol(painter, layer, posMatrix, tile, bucket, elementGroups, pref
 
     gl.uniform1f(program.u_zoom, (painter.transform.zoom - zoomAdjust) * 10); // current zoom level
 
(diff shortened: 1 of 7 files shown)


mourner:
Fantastic! Love the shader simplification too. :ship:

PR #2379, opened 2016-04-01
Add filter validation to query*Features

### LINKED ISSUE #2349: Validate filters passed to queryRenderedFeatures
**mapbox-gl-js version**: v0.16.0

### Steps to Trigger Behavior
queryRenderedFeatures filter is not working as expected as it is returning more features than desired.

1.  See fiddle: https://jsfiddle.net/rraponi/3pznu2ym/

### Expected Behavior
The cursor should change only over the markers that are displayed via
the filter that should only return features with the property of "test" that equal "bar".

 map.on('mousemove', function(e) {
    var features = map.queryRenderedFeatures(e.point, {
      filter: '["==", "test", "bar"]'
    });
    map.getCanvas().style.cursor = features.length ? 'default' : '';
  });

### Actual Behavior
More features are being returned and the cursor change is triggered in unexpected places.

lucaswoj:
fixes #2349

cc @ansis

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -629,6 +629,15 @@ Style.prototype = util.inherit(Evented, {
     },
 
     queryRenderedFeatures: function(queryGeometry, params, zoom, bearing) {
+        if (params.filter) {
+            validateStyle.throwErrors(this, validateStyle.filter({
+                key: 'queryRenderedFeatures.filter',
+                value: params.filter,
+                style: this.serialize(),
+                styleSpec: styleSpec
+            }));
+        }
+
         var sourceResults = [];
         for (var id in this.sources) {
             var source = this.sources[id];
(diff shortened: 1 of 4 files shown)


js/style/style.js
```diff
@@ -633,6 +633,15 @@ Style.prototype = util.inherit(Evented, {
     },
 
     queryRenderedFeatures: function(queryGeometry, params, classes, zoom, bearing) {
+        if (params.filter) {
+            if (validateStyle.emitErrors(this, validateStyle.filter({
+                key: 'queryRenderedFeatures.filter',
+                value: params.filter,
+                style: this.serialize(),
+                styleSpec: styleSpec
+            }))) return this;
```
ansis: This is a different return type than is expected. Should this throw instead?

js/style/style.js
```diff
@@ -643,6 +652,22 @@ Style.prototype = util.inherit(Evented, {
...
     },
 
+    querySourceFeatures: function(sourceID, params) {
+        if (params.filter) {
+            if (validateStyle.emitErrors(this, validateStyle.filter({
+                key: 'querySourceFeatures.filter',
+                value: params.filter,
+                style: this.serialize(),
+                styleSpec: styleSpec
+            }))) return this;
```
ansis: same question as https://github.com/mapbox/mapbox-gl-js/pull/2379/files#r58271553

js/style/style.js
lucaswoj: Yes, the return value should be an empty array.

We don't throw anywhere else in `Style`, even when we can. I think there's value in consistency in error reporting.

lucaswoj: I take that back. There are some places we do throw in `Style`. This should be a throw.

ansis:
looks good to me!

PR #2380, opened 2016-04-01
Batch class changes

### LINKED ISSUE #1384: Include setClasses, addClass, etc. in batch()
Looks like the class mutation operations all prompt a style cascade.  It would be nice to expose this within a `batch` to avoid having to cascade twice if I'm both updating styles and changing a class.

Happy to submit a PR if yall are :+1:

mourner:
Closes #1384. :eyes: @lucaswoj @anandthakker

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -200,7 +200,9 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     addClass: function(klass, options) {
         if (this._classes[klass]) return;
         this._classes[klass] = true;
-        if (this.style) this.style._cascade(this._classes, options);
+        this._classOptions = options;
+        if (this.style) this.style.cascade();
+        this._update(true);
     },
 
     /**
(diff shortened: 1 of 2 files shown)


js/ui/map.js
```diff
@@ -200,7 +200,9 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     addClass: function(klass, options) {
         if (this._classes[klass]) return;
         this._classes[klass] = true;
-        if (this.style) this.style._cascade(this._classes, options);
+        this._classOptions = options;
+        if (this.style) this.style.cascade();
+        this._update(true);
```
lucaswoj: I have a preference for moving logic out of `Map` and into `Style`. I wouldn't be opposed to adding more parameters to `_update` for the purpose.

mourner: What logic exactly? I don't see anything here that can be moved. Classes belong to the Map object.

lucaswoj: > Classes belong to the `Map` object.

This feels wrong to me. But it is a separate issue that can be addressed separately.

lucaswoj:
Looks good 🚢

js/ui/map.js
mourner: @lucaswoj I refer to @jfirebaugh's judgement here — he moved it from Style to Map earlier as far as I remember.

PR #2382, opened 2016-04-02
query gl for uniform and attribute names

ansis:
Instead of manually listing all that attribute and uniform names just ask gl to give them to you. After this there is one less thing that can break when editing a shader.

I wish I had known this was possible years ago

:eyes: @lucaswoj


lucaswoj:
Looks great! Thanks for doing this. 🚢 🐑 :shipit:

mollymerp:
:joy: :clap: now that #2372 is merged, we need one for the `outlinepattern` shader as well – let me know if you want me to add it!

PR #2010, opened 2016-01-28
support tiles with extent != 4096

ansis:
This supports layers with extent != 4096 by converting all data to the maximum extent (8192) supported by our buffers. After the conversion, only the constant maximum extent is used.

fixes https://github.com/mapbox/mapbox-gl-js/issues/1093

If a layer has an extent > 8192 it will lose precision when it is converted down to 8192. It will render ok, but it will be less exact when it is very overscaled. In order to increase the maximum extent, we would need to:
- pack the line extrude direction bits in either the `extrude` or `linesofar` values instead of the position. The reduce precision in `extrude` or `linesofar` could cause issues. https://github.com/mapbox/mapbox-gl-js/blob/3280d6dd85ff9cbadf4b5f6afc4daa5b21abeb56/js/data/line_bucket.js#L34-L35
- use unsigned integers and translate everything by EXTENT / 2 before adding it to the buffers. And then using the posMatrix to translate it back in the shader.

I'm not sure whether either of those changes are worthwhile right now. 8192 seems high enough

:eyes: @jfirebaugh @lucaswoj @mourner

PR #2383, opened 2016-04-02
Error if geometry extents exceeds allowed internal extent

yhahn:
Follow up to https://github.com/mapbox/mapbox-gl-js/pull/2010, throw an error if geometry exceeds internal extent after scaling.

- [ ] @ansis :eyes: for review? Not sure I got the negative min extent right.
- [ ] native port (error or warning)

### DIFF
--- a/test/js/data/load_geometry.test.js
+++ b/test/js/data/load_geometry.test.js
@@ -1,0 +1,45 @@
+'use strict';
+
+var test = require('tap').test;
+var fs = require('fs');
+var path = require('path');
+var Protobuf = require('pbf');
+var VectorTile = require('vector-tile').VectorTile;
+var loadGeometry = require('../../../js/data/load_geometry.js');
+
+// Load a line feature from fixture tile.
+var vt = new VectorTile(new Protobuf(new Uint8Array(fs.readFileSync(path.join(__dirname, '/../../fixtures/mbsv5-6-18-23.vector.pbf')))));
+
+test('loadGeometry', function(t) {
+    var feature = vt.layers.road.feature(0);
+    var originalGeometry = feature.loadGeometry();
+    var scaledGeometry = loadGeometry(feature);
+    t.equal(scaledGeometry[0][0].x, originalGeometry[0][0].x * 2, 'scales x coords by 2x');
+    t.equal(scaledGeometry[0][0].y, originalGeometry[0][0].y * 2, 'scales y coords by 2x');
+    t.end();
+});
+
+test('loadGeometry extent error', function(t) {
+    var feature = vt.layers.road.feature(0);
+    feature.extent = 2048;
+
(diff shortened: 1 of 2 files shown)


ansis:
@yhahn I can take over and finish this up.

EXTENT_MIN can be as low as `EXTENT * -2`.

@lucaswoj how should we expose this error to users? I think throwing in a worker will cause problems right now.

- should this be error the entire tile? or should it just drop the geometry? or should include the geometry and warn?
- should it console.warn or console.error in the worker? should the map emit an error event?

lucaswoj:
>  I think throwing in a worker will cause problems right now.

What problems do you anticipate? 

If this has the potential to break existing applications, I (with the limited context I have) think we should `console.warn` in the worker and leave the behaviour otherwise unchanged.

ansis:
> I (with the limited context I have) think we should console.warn in the worker and leave the behaviour otherwise unchanged.

Sounds good to me.

> > I think throwing in a worker will cause problems right now.

> What problems do you anticipate?

The parsing callback would never return. From the perspective of the main thread the tile parsing would still be in progress. But if we just warn we don't have to worry about catching anything.

ansis:
@lucaswoj using a console warning makes this hard to test. Any ideas?

lucaswoj:
You could mock out `console.warn`, like we do in [`fill_bucket.test.js`](https://github.com/mapbox/mapbox-gl-js/blob/master/test/js/data/fill_bucket.test.js#L16-L40)

```
    var originalWarn = console.warn;
    console.warn = function(message) {
      t.ok(foo.match(/foo/))
      t.end();    
    };

    ...

    console.warn = originalWarn;
```

There are some potential gotchas but we can work around them if they arise.

test/js/data/load_geometry.test.js
```diff
@@ -0,0 +1,45 @@
...
+    var warn = console.warn;
+    console.warn = function(warning) {
+        if (warning.match(/Geometry exceeds allowed extent, reduce your vector tile buffer size/)) {
+            numWarnings++;
+        }
+    };
+
+    loadGeometry(feature);
+
+    t.equal(numWarnings, 877);
```
lucaswoj: How hard would it be to log only one warning per `Map` instance? I don't think our users need to be told 877 times 😬.

ansis: most browsers de-dupe warnings on their end. Do you think there is a significant advantage to doing it on our end as well? I should have gotten :eyes: on this revision before merging

lucaswoj: I can imagine situations where these wouldn't / couldn't be deduped by the browser. We should try to dedupe on our end, at least per worker.

yhahn:
:rice: :bow:

Issue #2359, opened 2016-03-29
Rename "cascade" and "recalculate"


lucaswoj:
These names could be more descriptive. 

ref https://github.com/mapbox/mapbox-gl-js/pull/2355#discussion_r57767349

lucaswoj:
This is done now

PR #2390, opened 2016-04-04
Refactor and optimize layout and paint updates

### LINKED ISSUE #2339: Optimize cascade
One of the causes of #2334 is expensive `style.cascade` calls when doing `Map` `setPaintProperty`, which in turn calls `map.batch`. `cascade` in batch was only needed to fix `text-size` and `icon-size` (#1451) — in all other cases it's unnecessary. In addition, `batch` calls `_update(true)` (causing full style recalculation) which is also unnecessary in most cases.

@ansis want to grab this? [Lines in question](https://github.com/mapbox/mapbox-gl-js/blob/7ca58c020bcb9f3eb3ab9eb8b7d704db9ee2023c/js/ui/map.js#L476-L477).

**mapbox-gl-js version**: 0.16.0

mourner:
Closes #2339 and partially addresses #2359. 

The `cascade` term is repurposed as `updateClasses` in `Style` and `updatePaintTransitions` / `updatePaintTransition` in `StyleLayer`. The style keeps track of what paint properties need an update and updates only those, avoiding unnecessary work.

Update: DRYed up style validation code along the way because it bothered me.

:eyes: @lucaswoj @jfirebaugh @ansis

### DIFF
--- a/js/style/style_layer.js
+++ b/js/style/style_layer.js
@@ -85,13 +85,8 @@ StyleLayer.prototype = util.inherit(Evented, {
         if (value == null) {
             delete this._layoutDeclarations[name];
         } else {
-            if (validateStyle.emitErrors(this, validateStyle.layoutProperty({
-                key: 'layers.' + this.id + '.layout.' + name,
-                layerType: this.type,
-                objectKey: name,
-                value: value,
-                styleSpec: styleSpec
-            }))) return;
+            var key = 'layers.' + this.id + '.layout.' + name;
+            if (this._handleErrors(validateStyle.layoutProperty, key, name, value)) return;
             this._layoutDeclarations[name] = new StyleDeclaration(this._layoutSpecifications[name], value);
         }
         this._updateLayoutValue(name);
@@ -125,13 +120,7 @@ StyleLayer.prototype = util.inherit(Evented, {
             if (value === null || value === undefined) {
                 delete this._paintTransitionOptions[klass || ''][name];
             } else {
-                if (validateStyle.emitErrors(this, validateStyle.paintProperty({
-                    key: validateStyleKey,
-                    layerType: this.type,
-                    objectKey: name,
-                    value: value,
-                    styleSpec: styleSpec
-                }))) return;
+                if (this._handleErrors(validateStyle.paintProperty, validateStyleKey, name, value)) return;
                 this._paintTransitionOptions[klass || ''][name] = value;
             }
         } else {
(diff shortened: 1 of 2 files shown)


js/style/style_layer.js
```diff
@@ -189,17 +190,15 @@ StyleLayer.prototype = util.inherit(Evented, {
...
             if (!(name in declarations)) // apply removed declarations
-                this.updatePaintTransition(name, null, options, globalOptions, animationLoop);
+                this.applyPaintDeclaration(name, null, options, globalOptions, animationLoop);
         }
-
-        this.updateConstantValues();
     },
 
-    updatePaintTransition: function (name, declaration, options, globalOptions, animationLoop) {
+    applyPaintDeclaration: function (name, declaration, options, globalOptions, animationLoop) {
```
lucaswoj: Why the change here? Are you using `apply` to refer to methods that actually mutate `Style` and `update` to refer to methods that queue mutations to be made later?

mourner: Both actually mutate `Style`. `updatePaintTransition` updates a paint transition given a paint property name and classes. `applyPaintDeclaration` does the same but given a _declaration object_ (which is derived from classes/name). 

Both `updatePaintTransition` and `updatePaintTransitions` use `applyPaintDeclaration` internally, but the latter isn't supposed to be called anywhere else, so maybe it should be underscored (`_applyPaintDeclaration`) for clarity.

mourner:
Would it be OK to rename `recalculate` as a part of this PR (e.g. to `updateZoomBasedValues`), or should I wait for #1932 to land?

js/style/style_layer.js
mourner: I underscored all internal methods so this should be clearer now.

lucaswoj: What's unclear to me is use of the word `apply`. Why `apply` here but `update` elsewhere?

mourner: Because it's applying paint declaration to update paint transition. I just thought it would be confusing to have two methods that are named the same — `updatePaintTransition` and `_updatePaintTransition` with different parameters. What names would you suggest for those methods?

mourner: Because it's applying paint declaration to update paint transition. I just thought it would be confusing to have two methods that are named the same — `updatePaintTransition` and `_updatePaintTransition` with different parameters. What names would you suggest for those methods?

lucaswoj: This is ok for now 👍

lucaswoj:
🚢

PR #2392, opened 2016-04-05
Simplify useProgram

mourner:
:eyes: @ansis @lucaswoj

### DIFF
--- a/js/render/painter/use_program.js
+++ b/js/render/painter/use_program.js
@@ -98,7 +98,8 @@ module.exports._createProgram = function(name) {
     return util.extend({
         program: program,
         definition: definition,
-        attributes: attributes
+        attributes: attributes,
+        numAttributes: numAttributes
     }, attributes, uniforms);
 };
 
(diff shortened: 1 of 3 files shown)


js/render/painter/use_program.js
```diff
@@ -119,27 +120,19 @@ module.exports.useProgram = function (nextProgramName, posMatrix, exMatrix) {
...
+        var numPrevAttributes = previousProgram ? previousProgram.numAttributes : 0;
+        var i;
 
-                gl.disableVertexAttribArray(previousAttributes[i]);
-            }
+        // Disable all attributes from the previous program that aren't used in
+        // the new program. Note: attribute indices are *not* program specific!
+        // WebGL breaks if you disable attribute 0. http://stackoverflow.com/questions/20305231
+        for (i = Math.max(1, numNextAttributes); i < numPrevAttributes; i++) {
+            gl.disableVertexAttribArray(i);
```
ansis: it might be good to re-add this assertion https://github.com/mapbox/mapbox-gl-js/blob/a6c406cd45411151dc2ec1151835ff2e7060b57f/js/render/painter/use_program.js#L130

mourner: The condition still holds true because of `Math.max(1, numNextAttributes)`, so the code will never disable attribute 0.

ansis: yep, right

ansis:
Looks good to me!

PR #2393, opened 2016-04-05
Remove unnecessary hasData check in draw_line.js

mourner:
:eyes: @ansis @lucaswoj 
The function was in a hot loop and was reported as unoptimized by the Chrome profiler.

### DIFF
--- a/js/render/draw_line.js
+++ b/js/render/draw_line.js
@@ -20,15 +20,6 @@ module.exports = function drawLine(painter, source, layer, coords) {
     painter.setDepthSublayer(0);
     painter.depthMask(false);
 
-    var i, bucket;
-
-    var hasData = false;
-    for (i = 0; i < coords.length; i++) {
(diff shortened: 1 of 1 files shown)


ansis:
Is the early return significant in terms of performance? If it is, should this be added to the other draw* functions?

mourner:
@ansis as far as I remember, it was added to fix some errors when trying to draw empty line layers. Probably no performance difference.

ansis:
It looks like the check was first added in https://github.com/mapbox/mapbox-gl-js/commit/002c69071275997886fd98aa4dbc6025eed36a86 I don't remember why I added this. Probably a premature optimization. We should probably just remove it completely.

mourner:
@ansis 👍 removed.

ansis:
looks good. Can you squash before merging?

mourner:
@ansis the green button now automatically squashes in this repo.

PR #2394, opened 2016-04-05
Cache posMatrix in coordinate

### LINKED ISSUE #1993: optimize pos matrix calculation
It takes 10% of the time on the main thread. It's currently calculating a new matrix for each layer in a tile, when it should only need to do that once per-extent.

mourner:
Closes #1993. Makes `calculatePosMatrix` take ~0.7% rather than 6-10% of the main thread. :eyes: @ansis

### DIFF
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -5,6 +5,8 @@ var LngLat = require('./lng_lat'),
     Coordinate = require('./coordinate'),
     wrap = require('../util/util').wrap,
     interp = require('../util/interpolate'),
+    TileCoord = require('../source/tile_coord'),
+    EXTENT = require('../data/bucket').EXTENT,
     glmatrix = require('gl-matrix');
 
 var vec4 = glmatrix.vec4,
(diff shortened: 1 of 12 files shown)


js/geo/coordinate.js
```diff
@@ -15,6 +15,7 @@ function Coordinate(column, row, zoom) {
     this.column = column;
     this.row = row;
     this.zoom = zoom;
+    this.posMatrix = null;
```
lucaswoj: I'm 👎 on caching this within `Coordinate`. The `posMatrix` is a function of coordinate, source, and transform. It would make more sense to cache within the `calculatePosMatrix` method itself.

mourner: The convenient thing about caching inside `Coordinate` is that coordinate objects are discarded after rendering, so we don't retain outdated calculations in memory. Caching in `calculatePosMatrix` would be much more complicated technically to make it right. I prefer a simple and practical solution rather than ideologically pure one when it's very beneficial.

ansis: The coords used in draw* are actually TileCoords `js/source/tile_coord.js`. I think this is meant to be part of TileCoord.

> It would make more sense to cache within the calculatePosMatrix method itself.

`calculatePosMatrix` should probably not be a `Painter` method. Where should it go?

The equivalent method in -native is `matrixFor` and it is part of `TransformState`. -native stores matrices on it's `Tile` object. In -js wrapped tiles have one shared `Tile` object. In -native there are two `Tile` objects with a shared `TileData`.

ansis: I think precalculating the matrix and storing it on TileCoord is fine. TileCoord is the closest thing to -native's `Tile` that we have in -js now.

mourner: > The coords used in draw* are actually TileCoords js/source/tile_coord.js. I think this is meant to be part of TileCoord.

Yep, my bad — moved it there.

lucaswoj: > calculatePosMatrix should probably not be a Painter method. Where should it go?

It could always be a standalone module.

> Caching in `calculatePosMatrix` would be much more complicated technically to make it right. 

What if we clear the cache every time `transform` changes?

ansis: I'm not seeing this as caching. It calculates the matrix exactly once for every tile every frame. The question is where the calculated matrix should live and how it should be accessed.

The newly implemented approach is storing it on `coord` and accessing it as `coord.posMatrix`.

An alternative would be have a `painter.calculatePosMatrices(coords)` and `painter.getPosMatrix(coord)`. It would calculate the matrices and index them by `coord.id`. This would be similar to how clipping id's are used.

The first approach is more similar to -native. I think in -native it's preferable to just access a member of the tile than having a map lookup.

PR #2404, opened 2016-04-06
Add highlight/query similar features example

tristen:
highlight-similar-features

cc @mourner @lucaswoj @jfirebaugh for review

### DIFF
--- a/docs/_posts/examples/3400-01-24-query-similar-features.html
+++ b/docs/_posts/examples/3400-01-24-query-similar-features.html
@@ -1,0 +1,128 @@
+---
+layout: example
+category: example
+title: Highlight features containing similar data
+description: 'Hover over counties to highlight counties that share the same name.'
+---
+<style>
+.map-overlay {
+    font: 12px/20px 'Helvetica Neue', Arial, Helvetica, sans-serif;
+    background-color: #fff;
+    box-shadow: 0 1px 2px rgba(0,0,0,0.10);
+    border-radius: 3px;
+    position: absolute;
+    width: 25%;
+    top: 10px;
+    left: 10px;
+    padding: 10px;
+    display: none;
+}
+</style>
+
+<div id='map'></div>
+<div id='map-overlay' class='map-overlay'></div>
+
+<script>
+var map = new mapboxgl.Map({
+    container: 'map',
+    style: 'mapbox://styles/mapbox/streets-v8',
+    center: [-98, 38.88],
+    minZoom: 2,
+    zoom: 3
+});
+
+var overlay = document.getElementById('map-overlay');
+
+// Create a popup, but don't add it to the map yet.
+var popup = new mapboxgl.Popup({
+    closeButton: false
+});
+
+map.on('load', function() {
+    var canvas = map.getCanvasContainer();
+
+    // Add the source to query. In this example we're using
+    // county polygons uploaded as vector tiles
+    map.addSource('counties', {
+        "type": "vector",
+        "url": "mapbox://mapbox.82pkq93d"
+    });
+
+    map.addLayer({
+        "id": "counties",
+        "type": "fill",
+        "source": "counties",
+        "source-layer": "original",
+        "paint": {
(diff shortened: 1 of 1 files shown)


docs/_posts/examples/3400-01-24-query-similar-features.html
```diff
@@ -0,0 +1,137 @@
...
+        "filter": ["in", "COUNTY", ""]
+    }, 'place-city-sm'); // Place polygon under these labels.
+
+    map.on('mousemove', function(e) {
+        var features = map.queryRenderedFeatures(e.point, {
+            layers: ['counties']
+        });
+
+        // Change the cursor style as a UI indicator.
+        map.getCanvas().style.cursor = (features.length) ? 'pointer' : '';
```
jfirebaugh: No parens around `features.length`.

docs/_posts/examples/3400-01-24-query-similar-features.html
```diff
@@ -0,0 +1,137 @@
...
+
+        // Remove things if no feature was found.
+        if (!features.length) {
+            popup.remove();
+            map.setFilter('counties-highlighted', ['in', 'COUNTY', '']);
+            renderOverlay([]);
```
jfirebaugh: Inline the "no features" case of `renderOverlay` here.

docs/_posts/examples/3400-01-24-query-similar-features.html
```diff
@@ -0,0 +1,137 @@
...
+
+        // Query the counties layer visible in the map. Use the filter
+        // param to only collect results that share the same county name.
+        var relatedFeatures = map.querySourceFeatures('counties', {
+            sourceLayer: 'original',
+            filter: ['in', 'COUNTY', feature.properties.COUNTY]
+        });
+
+        // Render found features in an overlay.
+        renderOverlay(relatedFeatures);
```
jfirebaugh: And the "yes features" case here.

tristen: Not sure I follow this comment. There's an inline comment above this function that says

> Render found features in an overlay.

tristen: I'm following now :) inline the code I created a named function for. It only happens in one place.

tristen:
:ship:'d https://github.com/mapbox/mapbox-gl-js/commit/5342aa93d4ea111cc910cf68769e3716a6f0aba1

nitins1:
How do I find the data stored in the source file (mapbox://mapbox.82pkq93d) so I can manipulate the contents differently. For example, select counties in certain states as well, but using "STATES" as the key in the filter argument of the setFilter method doesn't work. Any ideas?

tristen:
👋 @nitins1 sounds like you probably want [`querySourceFeatures`](https://www.mapbox.com/mapbox-gl-js/api/#Map#querySourceFeatures)?

lbud:
@nitins1 — @tristen is right in that you can use `querySourceFeatures` to dig into the data added to a GL map.

Alternatively, you can look for that dataset's [tileJSON](https://www.mapbox.com/api-documentation/#retrieve-tilejson-metadata) — in this case, at `https://api.mapbox.com/v4/mapbox.82pkq93d.json?secure&access_token=<your access token>`
which contains this field:
```
  "vector_layers": [
    {
      "description": "",
      "fields": {
        "COUNTY": "String",
        "FIPS": "Number",
        "median-income": "Number",
        "population": "Number"
      },
      "id": "original",
      "maxzoom": 22,
      "minzoom": 0,
      "source": "mapbox.82pkq93d",
      "source_name": "counties"
    }
  ],
```

nitins1:
@tristen @lbud Thank you both for your responses! So I understand that I can use `querySourceFeatures` to return an array of objects that match a given filter. But after looking into the dataset's tileJSON, it doesn't seem like this data source has any information about states or counties relating to a state. Is there a data source/another walkthrough example with state data?

Issue #965, opened 2015-01-29
Enable custom raster tiled sources like WMS


Sumbera:
Would be great to have WMS or custom raster-tile source url support through official API. {x},{y},{z}  is not used by classic  WMS - based services  where spherical Mercator coordinates are needed as BBOX. Currently WMS works only by hacking into source._loadTile.
Sample and more details also here http://bl.ocks.org/Sumbera/ba554bb5cc8dfe4d7866

for simple bbox support, {mleft}, {mtop}, {mbottom}, {mright} could be added into TileCoord.url (only  for raster-tiles ?) plus calc Spherical Mercator bbox. However there might come other bbox , quadkey etc requirements...so custom url hook  would be perfect.

StepanKuzmin:
+1

fxi:
+1

DenisCarriere:
+1

tmcw:
Anyone want to roll a PR for this? The problem I've had in the past developing against WMS servers is that there aren't any stable WMS servers on the internet - I've used the NOAA server, but it doesn't support CORS, so this qualification of espg:3857 support + it'll stay online + CORS eliminates all WMS servers I know of.

DenisCarriere:
What if you host your own GeoServer at Mapbox for WMS testing purposes. There's a few [GeoServer Dockerfiles](https://hub.docker.com/r/winsent/geoserver/) that can get you started really fast. 

As for schemas, you could primarily focus on the latest OGC schema of [WMS 1.3](http://www.opengeospatial.org/standards/wms).

DenisCarriere:
I've been able to load a WMS 1.1 using Leaflet without a problem.
```javascript
L.tileLayer.wms("http://localhost:8080/geoserver/ows", {
    layers: 'imagery:Brockville',
    format: 'image/png',
    transparent: true,
    maxZoom: 23,
    maxNativeZoom: 23
}).addTo(map)
```
It would be really nice to have this built in MapboxGL. :+1:

tmcw:
I'd really prefer not to host a server just to develop a feature - WMS servers are unstable and slow, and nobody around here wants to maintain one. Unless someone wants to roll a PR for this, I don't think it's going to happen - this is a feature that connects to a kind of server that doesn't really exist on the internet.

DenisCarriere:
I agree with you, those WMS servers are old technology, so much trouble to enable this feature. :(

Sumbera:
[...]

PR #2409, opened 2016-04-08
Cheapo WMS support

tmcw:
2016-04-07 at 10 59 pm

(these changes would be resubmitted against master if people think this is a decent idea, this is more of a proposal than a nearly-finished PR)

This adds a `{bbox-epsg-3857}` token to URL requests. This would let
us support 3857-supporting CORS-supporting WMS servers in GL JS
with very little change - it would be a minor addition to the style
spec and easy to port to Native.

It's broken in a minor way right now: images aren't correctly
y-positioned through each zoom level. I know this is fixable,
someone with more sleep or fewer cold symptoms could probably get it
working quickly: the conversion from X/Y coords to "pseudo-meters"
or whatever you call 3857 units is wrong in a way that's probably
an off-by-one or something.

Example usage:

```js

    map.addSource('wms-test', {
        "type": "raster",
        "tiles": [
            "http://nowcoast.noaa.gov/arcgis/services/nowcoast/analysis_meteohydro_sfc_qpe_time/MapServer/WmsServer?SERVICE=WMS&REQUEST=GetMap&VERSION=1.1.1&LAYERS=5&STYLES=&FORMAT=image%2Fpng&TRANSPARENT=true&HEIGHT=256&WIDTH=256&SRS=EPSG%3A3857&BBOX={bbox-epsg-3857}",
        ],
        "tileSize": 256
    });

    map.addLayer({
        "id": "wms-test-layer",
        "type": "raster",
        "source": "wms-test",
        "paint": {
        }
    });
```

Refs #965

### DIFF
--- a/js/source/tile_coord.js
+++ b/js/source/tile_coord.js
@@ -49,8 +49,20 @@ TileCoord.fromID = function(id) {
 
 // given a list of urls, choose a url template and return a tile URL
 TileCoord.prototype.url = function(urls, sourceMaxZoom) {
+    var halfWidth = 20037508.342789;
+    var halfHeight = 20037508.342789;
+    var tilesAcross = Math.pow(2, this.z);
+    var tileWidth = 2 * (halfWidth / tilesAcross);
+    var tileHeight = 2 * (halfHeight / tilesAcross);
+    var leftPosition = ((this.x / tilesAcross) * 2 * halfWidth) - halfWidth;
+    var topPosition = halfHeight - ((this.y / tilesAcross) * 2 * halfHeight);
+    var bbox = leftPosition + ',' +
+        topPosition + ',' +
+        (leftPosition + tileWidth) + ',' +
+        (topPosition + tileHeight);
     return urls[(this.x + this.y) % urls.length]
         .replace('{prefix}', (this.x % 16).toString(16) + (this.y % 16).toString(16))
+        .replace('{bbox-epsg-3857}', bbox)
         .replace('{z}', Math.min(this.z, sourceMaxZoom || this.z))
         .replace('{x}', this.x)
         .replace('{y}', this.y);


bhousel:
Does it matter that the bbox parameter is constructed differently [depending on which version of WMS we're talking to](http://docs.geoserver.org/2.8.x/en/user/services/wms/basics.html#axis-ordering)?

mourner:
I like the approach. We could drop support for older WMS versions since we're already making assumptions like CORS and EPSG3857. Let's fix the discrepancy in the calculations, and then maybe refactor so that the bbox code only executes if there's a bbox token present in the url. Also, I think we could do away with just `bbox`, and describe all specifics in the docs. Otherwise we might want to name it `bbox-epsg-3857-wms-1-3-or-greater`.

bhousel:
We can avoid worrying about how to construct the bbox by having different tokens for the bounding edges, like `BBOX={lon_min},{lat_min},{lon_max},{lat_max}`

halset:
@bhousel This is EPSG:3857 with units in (quasi)meters, not lat/lon degrees. And the (irritating) issue of BBOX parameter ordering between WMS version is for EPSG:4326, not EPSG:3857. So having a single {bbox-epsg-3857} works fine for me.

indus:
If you are willing to add more replacement tokens, it would be nice to also add ```{-y}``` for TMS tile scheme support:
``` JS
.replace('{-y}', Math.pow(2, this.z) - this.y - 1)
```
(would close #2480 with smaller impact)

tmcw:
Closing in favor of #2612

Sumbera:
@bhousel, @indus I like your proposal as it is  more in-line with the original proposal here #965. I assume If coordinates are swapped, it is easier (and more flexible) to swap tokens  than lookup  in the documentation for how   BBOX parameter have been defined:
{one-with-min-max}
{another-with-max-min}

not to speak about more general getUrl-like API for raster tile that would enable to call any URL with whatever parameter sequence.

PR #2413, opened 2016-04-08
Touch debug

bhousel:
This just adds a debug page for showing console logs, and capturing map event debugging information.

I added this because 
1. I want to be able to better debug map events on mobile devices. 
2. It helps to see what events are being fired / how often / what order / etc..

debug_console


mourner:
Love this! 🚢

PR #2414, opened 2016-04-08
Make LatLngBounds obey its argument order

### LINKED ISSUE #2415: Allow "fitBounds" to span the date line
**mapbox-gl-js version**:

### Steps to Trigger Behavior

 1. Use fitBounds with correct SW/NE argument ordering and expect that a dateline-crossing bounds is accepted

### Expected Behavior

The map is centered over the dateline over a certain polygon.

### Actual Behavior

The fitBounds constructor calls LatLngBounds convert, which claims to support a specific argument order but really doesn't differentiate if the arguments are flipped ( see https://github.com/mapbox/mapbox-gl-js/pull/2414 for a PR). Even when the arguments are treated as strict, the fitBounds algorithm also doesn't handle the case of a dateline-crossing extent.

tmcw:
fixes #2415

### DIFF
--- a/js/geo/lng_lat_bounds.js
+++ b/js/geo/lng_lat_bounds.js
@@ -26,11 +26,11 @@ function LngLatBounds(sw, ne) {
     if (!sw) {
         return;
     } else if (ne) {
-        this.extend(sw).extend(ne);
+        this.setSW(sw).setNE(ne);
     } else if (sw.length === 4) {
-        this.extend([sw[0], sw[1]]).extend([sw[2], sw[3]]);
+        this.setSW([sw[0], sw[1]]).setNE([sw[2], sw[3]]);
     } else {
-        this.extend(sw[0]).extend(sw[1]);
+        this.setSW(sw[0]).setNE(sw[1]);
     }
 }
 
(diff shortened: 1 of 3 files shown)


lucaswoj:
Could you provide a little more context for this PR? What do you mean by "obey its argument order"? Can you demonstrate the behavior with a unit test?

tmcw:
See the referenced issue #2415 for the details

js/geo/lng_lat_bounds.js
```diff
@@ -36,6 +36,42 @@ LngLatBounds.prototype = {
...
+    setNE: function(obj) {
+        var ne = this._ne, ne2;
+
+        if (obj instanceof LngLat) {
+            ne2 = obj;
+        } else {
+            return obj ? this.setNE(LngLat.convert(obj) || LngLatBounds.convert(obj)) : this;
+        }
+        this._ne = new LngLat(ne2.lng, ne2.lat);
+        return this;
```
lucaswoj: Could this code be rewritten as

```js
setNE: function(ne) {
    ne = LngLat.convert(ne) || LngLatBounds.convert(ne) || this;
    this._ne = new LngLat(ne.lng, ne.lat);
    return this;
}
```

lucaswoj:
Refactored and fixed failing unit tests. 

There are some unit test "expected value" updates that raise a red flag in my head.

@mourner @morganherlocker What do yout think about these changes? How painful will this breaking change be to our users?

mourner:
@lucaswoj not sure how many people used the reverse order for the bounds, but it's easy to upgrade at least.

lucaswoj:
@mourner Sounds good to me. Is this ready to 🚢?

PR #2421, opened 2016-04-11
make attributes depend on only one paint property

ansis:
Having `a_color` depend on both `circle-color` and `circle-opacity` makes it break when one is data-driven and the other is a function of zoom. 

The `a_color` attribute used to depend on both the `circle-color` and `circle-opacity` paint properties. This splits them up into two attributes/uniforms, each of which only depends on one paint property.

I think it will be simpler to have all attributes be based on a single style property. We will need to split up:
- a_color, which depends on -color and -opacity
- u_linewidth, which depends on line-width and line-width-gap

This replaces a per-attribute `isDisabled` function with paint property key. All data-driven attributes must now depend on only a single paint property.

:eyes: @lucaswoj

### DIFF
--- a/js/data/bucket/circle_bucket.js
+++ b/js/data/bucket/circle_bucket.js
@@ -39,7 +39,7 @@ CircleBucket.prototype.programInterfaces = {
             name: 'color',
             components: 4,
             type: 'Uint8',
-            value: 'this._premultiplyColor(layer.getPaintValue("circle-color", globalProperties, featureProperties), 1)',
+            value: 'this._premultiplyColor(layer.getPaintValue("circle-color", globalProperties, featureProperties))',
             multiplier: 255,
             paintProperty: 'circle-color'
         }, {
--- a/js/render/draw_background.js
+++ b/js/render/draw_background.js
@@ -10,7 +10,7 @@ module.exports = drawBackground;
 function drawBackground(painter, source, layer) {
     var gl = painter.gl;
     var transform = painter.transform;
-    var color = util.premultiply(layer.paint['background-color'], layer.paint['background-opacity']);
+    var color = util.premultiply(layer.paint['background-color']);
     var image = layer.paint['background-pattern'];
     var opacity = layer.paint['background-opacity'];
     var program;
@@ -42,6 +42,7 @@ function drawBackground(painter, source, layer) {
 
         program = painter.useProgram('fill');
         gl.uniform4fv(program.u_color, color);
+        gl.uniform1f(program.u_opacity, opacity);
     }
 
     gl.disable(gl.STENCIL_TEST);
--- a/js/render/draw_fill.js
+++ b/js/render/draw_fill.js
@@ -9,12 +9,13 @@ function draw(painter, source, layer, coords) {
     var gl = painter.gl;
     gl.enable(gl.STENCIL_TEST);
 
-    var color = util.premultiply(layer.paint['fill-color'], layer.paint['fill-opacity']);
+    var color = util.premultiply(layer.paint['fill-color']);
     var image = layer.paint['fill-pattern'];
-    var strokeColor = util.premultiply(layer.paint['fill-outline-color'], layer.paint['fill-opacity']);
+    var strokeColor = util.premultiply(layer.paint['fill-outline-color']);
+    var opacity = layer.paint['fill-opacity'];
 
     // Draw fill
-    if (image ? !painter.isOpaquePass : painter.isOpaquePass === (color[3] === 1)) {
+    if (image ? !painter.isOpaquePass : painter.isOpaquePass === (color[3] === 1 && opacity === 1)) {
         // Once we switch to earcut drawing we can pull most of the WebGL setup
         // outside of this coords loop.
         for (var i = 0; i < coords.length; i++) {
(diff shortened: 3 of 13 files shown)


lucaswoj:
Good thinking. This would simplify our implementation a fair bit. We could get rid of the `isLayerConstant` property too.

> All data-driven attributes must now depend on only a single paint property.

Can you think of any circumstances in which we might need attributes that are a function of multiple paint properties or other state in the future?

js/data/bucket.js
```diff
@@ -487,12 +487,7 @@ function createAttributes(bucket) {
...
-    if (attribute.isDisabled === undefined || attribute.isDisabled === false) {
-        return false;
-    } else if (attribute.isDisabled === true) {
-        return true;
-    } else {
-        return !!attribute.isDisabled.call(bucket, layer);
-    }
+function isAttributeDisabled(attribute, layer) {
+    return attribute.paintProperty !== undefined &&
+        layer.isPaintValueFeatureConstant(attribute.paintProperty);
```
lucaswoj: Can we inline this function now that it is so much simpler?

js/data/bucket/circle_bucket.js
```diff
@@ -40,19 +40,9 @@ CircleBucket.prototype.programInterfaces = {
             components: 4,
             type: 'Uint8',
             isLayerConstant: false,
-            value: (
-                'this._premultiplyColor(' +
-                    'layer.getPaintValue("circle-color", globalProperties, featureProperties),' +
-                    'layer.getPaintValue("circle-opacity", globalProperties, featureProperties)' +
-                ')'
-            ),
+            value: 'this._premultiplyColor(layer.getPaintValue("circle-color", globalProperties, featureProperties), 1)',
```
lucaswoj: Do we still need to call `_premultiplyColor` if the opacity is always  `1`?

ansis:
> We could get rid of the isLayerConstant property too.

Yeah, good point

> Can you think of any circumstances in which we might need attributes that are a function of multiple paint properties or other state in the future?

I can't think of any right now. Hopefully we can find ways to keep them all independent of each other

js/data/bucket/circle_bucket.js
ansis: Yes, to multiply the alpha component of the color into the rgb components. After we split up color and opacity attributes for all layer types I think we'll be able to drop the second parameter from `premultiplyColor`.

js/data/bucket.js
ansis: Yep, good idea

js/data/bucket/circle_bucket.js
lucaswoj: Can we do that multiplication within the shader now? It could be a small perf win and a small code complexity win.

js/data/bucket.js
ansis: done in b41f4ff

js/data/bucket/circle_bucket.js
ansis: I left the premultiplication outside the shader but I switch all shaders to separate color and opacity inputs in adb11ed and then removed the `additionalOpacity` parameter from `premultiply`.

ansis:
- [ ] :eyes: from @lucaswoj on the extra changes I added

js/data/bucket.js
```diff
@@ -223,7 +225,7 @@ Bucket.prototype.setAttribPointers = function(programName, gl, program, offset,
     var disabledAttributes = this.attributes[programName].disabled;
     for (var i = 0; i < disabledAttributes.length; i++) {
         attribute = disabledAttributes[i];
-        if (attribute.isLayerConstant === false && layer.id !== attribute.layerId) continue;
+        assert(attribute.paintProperty);
```
lucaswoj: We need to check `layer.id !== attribute.layerId` to handle the case where there are multiple layers using property functions (data-driven styling functions) sharing a `ref` parent.

lucaswoj: An attribute could be disabled for some layers but not others.

lucaswoj:
Do you have an objection to dropping `premultiply`? I think it'd be as easy as

```
gl_FragColor = v_color * (1.0 - t) * u_opacity * v_color.a
```

js/data/bucket.js
ansis: fixed in f469330d8aeb9d5db5bb7c11cac7b347b6cdbed2

it's safe to remove the `attribute.isLayerConstant === false` because this is always true for disabled attributes, right?

lucaswoj: > it's safe to remove the attribute.isLayerConstant === false because this is always true for disabled attributes, right?

Correct (as of the new `paintProperty` implementation) 👍

ansis:
> Do you have an objection to dropping premultiply? I think it'd be as easy as
>
> gl_FragColor = v_color * (1.0 - t) * u_opacity * v_color.a

Only the `rgb` components need to be multiplied by `a` so it would have to be:

```glsl
gl_FragColor = v_color * (1.0 - t) * u_opacity;
gl_FragColor.rgb *= v_color.a;
```

or

```glsl
v_color = a_color;
v_color.rgb *= a_color.a;
```

It's either one line in the shaders or one line in javascript. I prefer doing it in javascript where we can use a named function.

Also, in general it's good to move as much of the calculations as possible to outside the shader where they get calculated once per draw instead of once per vertex/fragment. This probably doesn't matter much in this case.

lucaswoj:
Sounds good. This is ready to 🚢

PR #2422, opened 2016-04-11
use glsl macros instead of custom templating

ansis:
Using macros and preprocessor directives is the glsl way of doing this. After this change the .glsl files are valid glsl.

I personally think it's clearer to have both variations explicitly written in .glsl files.

:eyes: @lucaswoj

### DIFF
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -328,26 +328,16 @@ Bucket.prototype.recalculateStyleLayers = function() {
     }
 };
 
-Bucket.prototype.getUseProgramTokens = function(programInterface, layer) {
-    var tokens = {};
-
+Bucket.prototype.getProgramMacros = function(programInterface, layer) {
+    var macros = [];
     var enabledAttributes = this.attributes[programInterface].enabled;
     for (var i = 0; i < enabledAttributes.length; i++) {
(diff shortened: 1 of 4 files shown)


lucaswoj:
I don't think that this:

```
#ifdef ATTRIBUTE_COLOR
attribute vec4 a_color;
#else
uniform vec4 a_color;
#endif

#ifdef ATTRIBUTE_SIZE
attribute float a_size;
#else
uniform float a_size;
#endif

#ifdef ATTRIBUTE_BLUR
attribute float a_blur;
#else
uniform float a_blur;
#endif
```

is more readable, more robust, or more idiomatic* GLSL than this:

```
{{colorType}} attribute vec4 a_color;
{{sizeType}} attribute vec4 a_size;
{{blurType}} attribute vec4 a_blur;
```

Are there any cases where we'd want to pass fundamentally different qualifiers to the attribute vs uniform versions? 

----

* which is to say that both are pretty weird 😬

jfirebaugh:
I prefer the macro approach because it doesn't introduce a language concept that's not part of GLSL itself.

lucaswoj:
Ok. GLSL macros it is. 🚢

PR #2423, opened 2016-04-12
data driven styling without code generation

ansis:
This removes code generation from `Bucket` so that it's closer to something we could implement in -native.

:eyes: @lucaswoj 

Instead of adding paint attributes in the main vertex loop they are added in a separate loop later. This has a couple advantages:
- The value of a paint attribute is the same for all vertices in a feature. The value is now calculated once per feature instead of once for each vertex which. This should help with performance for data driven lines, fills and symbols.
- Adding them in a separate loop let's us use regular functions for the `addVertex` methods.

We don't have any benchmarks for data driven attributes yet I haven't measured whether this makes data driven attributes slower or faster. Evaluating the style function only once per feature should help with performance.  I think [this](https://github.com/mapbox/mapbox-gl-js/compare/data-driven-styling...data-driven-styling-less-code-generation?expand=1#diff-c1d5cefb566acef357121481a5a8e27eR346) is a very hot loop that we might need to optimize more. If we can't get it fast enough without code generation we could add a StructArray method that fills in a single value for a range of elements.

The commit messages explain my thinking for each individual commit.

### DIFF
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -47,9 +47,6 @@ Bucket.EXTENT = 8192;
  * style spec layer type. Because `Bucket` is an abstract class,
  * instances should be created via the `Bucket.create` method.
  *
- * For performance reasons, `Bucket` creates its "add"s methods at
- * runtime using `new Function(...)`.
- *
  * @class Bucket
  * @private
  * @param options
@@ -150,7 +147,6 @@ Bucket.prototype.createArrays = function() {
         var programInterface = this.programInterfaces[programName];
 
         if (programInterface.vertexBuffer) {
-            var vertexAddMethodName = this.getAddMethodName(programName, 'vertex');
             var vertexBufferName = this.getBufferName(programName, 'vertex');
 
             var VertexArrayType = new StructArrayType({
@@ -160,8 +156,6 @@ Bucket.prototype.createArrays = function() {
 
             arrays[vertexBufferName] = new VertexArrayType();
             arrayTypes[vertexBufferName] = VertexArrayType.serialize();
-
-            this[vertexAddMethodName] = this[vertexAddMethodName] || createVertexAddMethod(this, programName);
         }
 
         if (programInterface.elementBuffer) {
@@ -358,45 +352,6 @@ Bucket.prototype.addPaintAttributes = function(interfaceName, globalProperties, 
         }
     }
 };
-
-var createVertexAddMethodCache = {};
-function createVertexAddMethod(bucket, interfaceName) {
-    var enabledAttributes = bucket.attributes[interfaceName].enabled;
-    var programInterface = bucket.programInterfaces[interfaceName];
-
-    var body = 'var layer;\n';
-
-    var pushArgs = [];
-
-    for (var i = 0; i < enabledAttributes.length; i++) {
-        var attribute = enabledAttributes[i];
-        if (attribute.paintProperty) continue;
-
-        var attributePushArgs = [];
-        attributePushArgs = attributePushArgs.concat(attribute.value);
-
-        var multipliedAttributePushArgs;
-        if (attribute.multiplier) {
-            multipliedAttributePushArgs = [];
-            for (var k = 0; k < attributePushArgs.length; k++) {
-                multipliedAttributePushArgs[k] = attributePushArgs[k] + '*' + attribute.multiplier;
-            }
-        } else {
-            multipliedAttributePushArgs = attributePushArgs;
-        }
-
-        pushArgs = pushArgs.concat(multipliedAttributePushArgs);
-    }
-
-    var bufferName = bucket.getBufferName(interfaceName, 'vertex');
-    body += 'return this.arrays.' + bufferName + '.emplaceBack(' + pushArgs.join(',') + ');';
-
-    if (!createVertexAddMethodCache[body]) {
-        createVertexAddMethodCache[body] = new Function(programInterface.attributeArgs, body);
-    }
-
-    return createVertexAddMethodCache[body];
-}
 
 function createElementAddMethod(buffer) {
     return function(one, two, three) {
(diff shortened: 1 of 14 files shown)


ansis:
I merged https://github.com/mapbox/mapbox-gl-js/pull/2424 into here so everything can be reviewed in one spot.

js/data/bucket.js
```diff
@@ -449,7 +367,7 @@ function createAttributes(bucket) {
                 if (!attribute.paintProperty && layer.id !== bucket.layer.id) continue;
                 if (attribute.paintProperty && layer.isPaintValueFeatureConstant(attribute.paintProperty)) {
                     interfaceAttributes.disabled.push(util.extend({}, attribute, {
-                        getValue: createGetAttributeValueMethod(bucket, interfaceName, attribute, j),
+                        getValue: attribute.getValue,
```
lucaswoj: We can remove this line entirely because we are already extending `attribute`

js/data/bucket.js
```diff
@@ -458,6 +376,7 @@ function createAttributes(bucket) {
                     }));
                 } else {
                     interfaceAttributes.enabled.push(util.extend({}, attribute, {
+                        getValue: attribute.getValue,
```
lucaswoj: We can remove this line entirely because we are extending `attribute`.

js/data/bucket.js
```diff
@@ -150,34 +147,42 @@ Bucket.prototype.createArrays = function() {
         var programInterface = this.programInterfaces[programName];
 
         if (programInterface.vertexBuffer) {
-            var vertexAddMethodName = this.getAddMethodName(programName, 'vertex');
             var vertexBufferName = this.getBufferName(programName, 'vertex');
 
             var VertexArrayType = new StructArrayType({
-                members: this.attributes[programName].enabled,
+                members: this.attributes[programName].coreAttributes,
```
lucaswoj: Could these be called `layoutAttributes`? Then we'd have `layoutAttributes` as a function of layout properties and `paintAttributes` as a function of paint properties.

js/data/bucket.js
```diff
@@ -278,14 +241,10 @@ Bucket.prototype.bindBuffers = function(programInterfaceName, gl, options) {
...
 
-/**
- * Get the name of the method used to add an item to a buffer.
- * @param {string} programName The name of the program that will use the buffer
- * @param {string} type One of "vertex", "element", or "secondElement"
- * @returns {string}
- */
-Bucket.prototype.getAddMethodName = function(programName, type) {
-    return 'add' + capitalize(programName) + capitalize(type);
+Bucket.prototype.bindPaintBuffer = function(gl, interfaceName, layerID, program, vertexStartIndex) {
```
lucaswoj: 👍 and we should rename the other method `bindLayoutBuffer` now.

js/data/bucket/line_bucket.js
```diff
@@ -50,38 +50,37 @@ function LineBucket() {
...
+            Math.round(EXTRUDE_SCALE * extrude.x) + 128,
+            Math.round(EXTRUDE_SCALE * extrude.y) + 128,
+            // Encode the -1/0/1 direction value into the first two bits of .z of a_data.
+            // Combine it with the lower 6 bits of `linesofar` (shifted by 2 bites to make
+            // room for the direction value). The upper 8 bits of `linesofar` are placed in
+            // the `w` component. `linesofar` is scaled down by `LINE_DISTANCE_SCALE` so that
+            // we can store longer distances while sacrificing precision.
+            ((dir === 0 ? 0 : (dir < 0 ? -1 : 1)) + 1) | (((linesofar * LINE_DISTANCE_SCALE) & 0x3F) << 2),
+            (linesofar * LINE_DISTANCE_SCALE) >> 6);
+};
```
lucaswoj: 😂

js/util/struct_array.js
```diff
@@ -308,6 +308,17 @@ StructArray.prototype._resize = function(n) {
...
 /**
+ * Resize the array.
+ * If `n` is greater than the current length then additional elements with undefined values are added.
+ * If `n` is less than the current length then the array will be reduced to the first `n` elements.
+ * @param {number} n The new size of the array.
+ */
+StructArray.prototype.resize = function(n) {
+    this.length = n;
+    if (this.length > this.capacity) this.reserve(n);
+};
```
lucaswoj: Can we combine `resize` and `reserve` into one method?

js/data/bucket.js
ansis: done in 9b1bfb01258501a19f7ffd5159e8b3f4b75daa28

js/data/bucket.js
ansis: done in 9b1bfb01258501a19f7ffd5159e8b3f4b75daa28

test/js/data/bucket.test.js
```diff
@@ -92,14 +100,19 @@ test('Bucket', function(t) {
         var bucket = create();
 
         bucket.features = [createFeature(17, 42)];
+        bucket.debug = true;
         bucket.populateBuffers();
+        bucket.debug = false;
```
lucaswoj: Is this leftover debugging code?

lucaswoj:
This looks super 🚀. Thank you so much @ansis!

js/data/bucket.js
ansis: good idea, done 7852535e6e80353f761671f9c4c5263a68ec99de. The paint/layout categorization is surviving dds better than I thought it would

js/data/bucket.js
ansis: 7852535e6e80353f761671f9c4c5263a68ec99de

js/util/struct_array.js
ansis: yep, that's cleaner. 3ad6529b649a94b6503161c975dd96f48f1c8f43

test/js/data/bucket.test.js
ansis: yep, removed c3cb877730f9ea993aa3f45a5531aae699c06836

PR #2436, opened 2016-04-13
Add Popup#setDOMContent method

### LINKED ISSUE #1671: Popup.setHTML() doesn't accept DOM node
I am migrating from MapBox.js to GL.js. I used to have custom Angular directives as markers' popup, showing their properties with a button to call a controller function.
Anyway, I noticed that Angular code is not recognized in mapboxgl.Popup.setHTML().
Is there a real reason, or it's just a bug?
Thank you

anandthakker:
Closes #1671 

👀 @jfirebaugh 

 - Used `setDOMContent` instead of `setElement` or `setContent` just be totally explicit.  Can change it to whatever.
 - The only test I added was a manual one because I didn't see any other unit tests for popups.
 - `setText` refactored to reuse code in `setDOMContent`
 - Is it worth having `setDOMContent` accept a NodeList, too?  If so, `setHTML` could also just delegate to this.

### DIFF
--- a/js/ui/popup.js
+++ b/js/ui/popup.js
@@ -111,18 +111,16 @@ Popup.prototype = util.inherit(Evented, /** @lends Popup.prototype */{
      * @returns {Popup} `this`
      */
     setHTML: function(html) {
-        this._createContent();
-
+        var frag = document.createDocumentFragment();
         var temp = document.createElement('body'), child;
         temp.innerHTML = html;
         while (true) {
             child = temp.firstChild;
             if (!child) break;
-            this._content.appendChild(child);
+            frag.appendChild(child);
         }
 
-        this._update();
-        return this;
+        return this.setDOMContent(frag);
     },
 
     /**
--- a/js/ui/popup.js
+++ b/js/ui/popup.js
@@ -102,7 +102,7 @@ Popup.prototype = util.inherit(Evented, /** @lends Popup.prototype */{
      * @returns {Popup} `this`
      */
     setText: function(text) {
-        return this.setContent(document.createTextNode(text));
+        return this.setDOMContent(document.createTextNode(text));
     },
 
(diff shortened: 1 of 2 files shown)


js/ui/popup.js
```diff
@@ -102,11 +102,7 @@ Popup.prototype = util.inherit(Evented, /** @lends Popup.prototype */{
      * @returns {Popup} `this`
      */
     setText: function(text) {
-        this._createContent();
-        this._content.appendChild(document.createTextNode(text));
-
-        this._update();
-        return this;
+        return this.setContent(document.createTextNode(text));
```
jfirebaugh: `setContent` ⇢ `setDOMContent`

anandthakker: Whoops. Thanks.

jfirebaugh:
Thanks!

> Is it worth having setDOMContent accept a NodeList, too? If so, setHTML could also just delegate to this.

I wonder if we can keep the current `setDOMContent` and use `DocumentFragment` for `setHTML`. Does this work?

```
     setHTML: function(html) {
         var fragment = document.createDocumentFragment();
         fragment.innerHTML = html;
         return this.setDOMContent(fragment);
     }
```

anandthakker:
Argh, unfortunately it looks like setting `innerHTML` on `DocumentFragment` doesn't work. (SO confirms: http://stackoverflow.com/a/9285046)

anandthakker:
@jfirebaugh ^ not as streamlined as we'd hoped, but f2e3271at least gets all three methods sharing the setup & update stuff.

anandthakker:
Lemme know if you want me to squash

jfirebaugh:
Thanks, merged as bbfe05be4f27d53f93883bf2efb2b6b32bf3a6ba.

PR #2437, opened 2016-04-13
Fixing the Evented documentation

danzel:
The Evented documentation is a bit broken
https://www.mapbox.com/mapbox-gl-js/api/#Evented

It starts with a Parameters field (not sure what that is for)
For `off`/`on`/`once` there are 3 parameters, but really these all just have 2 parameters.

Pretty certain this is because the parameter name in the docs doesn't line up with the actual parameter name, so I've renamed the parameters to match the docs

I can't run the docs because I'm on windows :( So someone will need to check this fixes the issue. Thanks!

### DIFF
--- a/js/util/evented.js
+++ b/js/util/evented.js
@@ -13,6 +13,7 @@ var Evented = {
      *
      * @param {string} type Event type
      * @param {Function} listener Function to be called when the event is fired
+     * @returns {Object} `this`
      */
     on: function(type, listener) {
         this._events = this._events || {};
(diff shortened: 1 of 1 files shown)


danzel:
Just noticed that some of the methods were documenting they returned `this`, added.

danzel:
Build failure doesn't look like my fault *shrug*

lucaswoj:
> Build failure doesn't look like my fault shrug

Nope. That's my fault 😬 . Fixed and restarting your build now.

lucaswoj:
This looks great. Thank you for the contribution @danzel!

PR #2450, opened 2016-04-14
Replace "style.load" with "load" events in examples

lucaswoj:
`style.load` isn’t part of our public API. Its use became widespread because of a bug with `load` a few versions back. We should always use `load` in our examples. 

I tested all the modified examples by hand and ensured they all still work as intended.

cc @tristen @bhousel @lyzidiamond @ansis

### DIFF
--- a/docs/_posts/examples/3400-01-01-toggle-layers.html
+++ b/docs/_posts/examples/3400-01-01-toggle-layers.html
@@ -59,7 +59,7 @@ var map = new mapboxgl.Map({
     center: [-71.97722138410576, -13.517379300798098]
 });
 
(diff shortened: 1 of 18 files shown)


docs/_posts/examples/3400-01-01-toggle-layers.html
```diff
@@ -68,6 +68,9 @@
         'id': 'museums',
         'type': 'circle',
         'source': 'museums',
+        'layout': {
+            'visibility': 'visible'
+        },
```
lucaswoj: This fixes an unrelated bug I stumbled across.

Issue #1804, opened 2015-12-05
Duplicate attribution


planemad:
We're having this issue on the [Chennai flood map](http://osm-in.github.io/flood-map/chennai.html#16.86/12.99241/80.25987) after adding the `mapbox.satellite` layer.

<img width="665" alt="screenshot 2015-12-05 08 45 39" src="https://cloud.githubusercontent.com/assets/126868/11605735/144a0132-9b2d-11e5-80a1-c6f38af1f5d6.png">

mourner:
I'm not sure we can fix this automatically — it grabs attribution strings from corresponding TileJSON of both sources, and it's not a 1:1 match. Perhaps we could detect whether one attribution is a substring of another one and then use the bigger one.

lucaswoj:
We're due for an overhaul of our attribution system #1485. Perhaps we could support a per-source attribution syntax like

```
attribution: ["© Mapbox", "© OpenStreetMap", "© Digital Globe"]
```
 
So that we can better merge and dedupe strings

mourner:
@lucaswoj but that would require changing the TileJSON spec and propagating that change through our whole stack, right? Substring matching may be good enough of a hack for now.

lucaswoj:
Right. Substring matching sounds like a better short-term solution.

jfirebaugh:
See also #1752.

lucaswoj:
This was partially addressed by https://github.com/mapbox/mapbox-gl-js/commit/d7dacd63937b0429df46b8279a75090727501efd and I have not heard any reqests for refinements since. 

Related to #1485

1ec5:
A more structured attribution syntax like the one described in https://github.com/mapbox/mapbox-gl-js/issues/1804#issuecomment-162717387 is being tracked in mapbox/tilejson-spec/issues/20.

ygoe:
I see this all the time here. Is there something I can do in my code?

romankaravia:
Looks like this is currently happening with Light, Dark, Streets, and Outdoors styles, also on https://mapbox.com, see e.g. [Light style preview](https://www.mapbox.com/maps/light#:~:text=Mapbox%20Streets%20data):

![image](https://github.com/user-attachments/assets/cd23cd6c-9c3c-4e55-9792-1fce0938f652)

[...]

PR #2453, opened 2016-04-15
Dedupe attributions that are substrings of others

anandthakker:
Refs #1804

### DIFF
--- a/js/ui/control/attribution.js
+++ b/js/ui/control/attribution.js
@@ -49,6 +49,15 @@ Attribution.prototype = util.inherit(Control, {
             }
         }
 
+        // remove any entries that are duplicates or strict substrings of another entry.
+        attributions.sort();
+        attributions = attributions.filter(function (attrib, i) {
+            for (var j = i + 1; j < attributions.length; j++) {
+                if (attributions[j].indexOf(attrib) >= 0) { return false; }
+            }
+            return true;
+        });
+
         this._container.innerHTML = attributions.join(' | ');
         this._editLink = this._container.getElementsByClassName('mapbox-improve-map')[0];
         this._updateEditLink();
--- a/js/ui/control/attribution.js
+++ b/js/ui/control/attribution.js
@@ -19,6 +19,29 @@ function Attribution(options) {
     util.setOptions(this, options);
 }
 
+Attribution.createAttributionString = function(sources) {
+    var attributions = [];
+
+    for (var id in sources) {
+        var source = sources[id];
+        if (source.attribution && attributions.indexOf(source.attribution) < 0) {
+            attributions.push(source.attribution);
+        }
+    }
+
+    // remove any entries that are substrings of another entry.
+    // first sort by length so that substrings come first
+    attributions.sort(function (a, b) { return a.length - b.length; });
+    attributions = attributions.filter(function (attrib, i) {
+        for (var j = i + 1; j < attributions.length; j++) {
+            if (attributions[j].indexOf(attrib) >= 0) { return false; }
+        }
+        return true;
+    });
+
+    return attributions.join(' | ');
+};
+
 Attribution.prototype = util.inherit(Control, {
     options: {
         position: 'bottom-right'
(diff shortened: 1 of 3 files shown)


anandthakker:
(This isn't a full resolution of 1804, just a short-term one that works for, e.g., the Satellite Streets style provided in Studio)

js/ui/control/attribution.js
```diff
@@ -49,6 +49,15 @@ Attribution.prototype = util.inherit(Control, {
             }
         }
 
+        // remove any entries that are duplicates or strict substrings of another entry.
+        attributions.sort();
```
lucaswoj: Why do you sort the attributions before filtering them?

test/manual/attribution.html
```diff
@@ -28,6 +28,29 @@
...
+        }
+    }));
+    map.style.sources.bar.attribution = 'Another Source';
+
+    map.addSource('baz', new mapboxgl.GeoJSONSource({
+        data: {
+            type: 'FeatureCollection', features: []
+        }
+    }));
+    map.style.sources.baz.attribution = 'Another Source';
```
lucaswoj: Is anything blocking us from creating automated unit tests for this code?

js/ui/control/attribution.js
anandthakker: @lucaswoj i figured sorting the list first means that, in the filter function for the `i-th` item, you can safely start the loop at `i+1`.  Without this, you have to check every item, and I think it gets a little weird to correctly filter out exact duplicates.

test/manual/attribution.html
anandthakker: Ah, I meant to mention this in the PR: I looked and didn't find any other unit tests for `Attribution`, so I held back on adding one in case there was a reason not to.  Happy to add a test, though.

js/ui/control/attribution.js
lucaswoj: Whether you sort or not, you're comparing every attribution string to every other attribution string. I don't think the sort is necessary.

anandthakker: > you're comparing every attribution string to every other attribution string

Agreed, but it's also directional, since we're looking for a substring match and want to keep the longer of the two if there's a match.

I agree the sort isn't necessary, but removing it does make the logic in the filter function a little more particular... I think it would then need to be something like:

```js
attribution = attribution.filter(function (a, i) {
  // check if `a` is a substring of another string in the list
  var isSubsumedByAnother = attribution.some(function (b, j) {
    // skip if we're checking the same element
    if (i === j) { return false }
    // if a === b, then only call it a match if a's index is higher; otherwise
    // we would filter out both elements when there's an exact match
    if (a === b) { return i > j }
    // check if a is a substring of b
    if (b.indexOf(a) >= 0) { return true }
  })
  return !isSubsumedByAnother
})
```

@lucaswoj I considered this, and ended up feeling that both approaches were equally bad... lemme know if you find this way clearer and I can change it

lucaswoj: > Agreed, but it's also directional, since we're looking for a substring match and want to keep the longer of the two if there's a match.

That's a good point. We should sort by _length_ not alphabetic ordering, then.

anandthakker:
@lucaswoj Updated:
 - Fixed sort to be by length
 - Pulled out the source -> final attribution string into a separate method on `Attribution` itself to make it testable without needing a `Map` instance
 - Added test for deduping logic

mourner:
I'm wondering why the builds fail here... The error is not very telling, something about query tests.

anandthakker:
@mourner I found that this happened on at least one other PR, and I wondered if it could be some kind of Circle config issue relating to this branch being on a fork rather than the main repo?

anandthakker:
@mourner Oh yeah, I remember now--here's the message that made me think that last time:

```
/home/ubuntu/mapbox-gl-js/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```

lucaswoj:
The build is failing because the branch exists in @anandthakker's repo, not the @mapbox repo. It appears that Coveralls doesn't work with pull requests originating in other repos. We should fail silently in this case. I'll cut a separate ticket.

lucaswoj:
I'm feeling 👍 on merging this. 

Let's wait a day to give @tmcw @jfirebaugh @ansis @planemad a chance to weigh in.

ansis:
looks good to me

Issue #2428, opened 2016-04-12
Support style functions using both feature properties and zoom


ansis:
#1932 implemented data driven styling for circles. Other types will be implemented soon.

Style functions can currently either be functions of zoom or functions of data. They can't be functions of both at the same time. This makes it hard to use data-driven styles with properties should often change with zoom like `circle-radius`, `line-width` and `text-size`. For example, it would be nice if the circles could get bigger as you zoom in: https://bl.ocks.org/lucaswoj/raw/4fa16fcf9663f67f6b7d4a7225dc3524/

Data + zoom driven functions would also let us merge a lot of mapbox streets layers into one:
- all road layers could be one layer, all tunnel layers could be another layer
- all the multi-size label layers like country-label-lg, country-label-md, country-label-small could be merged into one layer

Data + zoom driven functions would be useful but would also add complexity to the spec and studio.

**If** we add it, imagine it would look something like a spreadsheet. On one axis you would have zoom levels and on the other axis you would have data properties. Each zoom/data pair would have it's own value.

### For example:

```
property: population

        0       1000        1000000
      ------------------------------
z0    | 1       10          20
z10   | 4       15          30
z20   | 6       20          40
```
is the function for circle radius. You would evaluate it for `{ properties: { population: 500000 } }` at z6 like this:
- find the two closest steps, z0 and z10
- at each step, find the value for the feature:
    - `interp(10, 20, (500000 - 1000) / (1000000 - 1000)) ~= 15 at z0` and
    - `interp(25, 30, (500000 - 1000) / (1000000 - 1000)) ~= 22.5 at z10`
- interpolate those two values based on the zoom levels to get the final value
    - `interp(15, 22.5, (6 - 0) / (10 - 0) = 19.5 at z6`

### The GL implementation:
When creating the buffers, find the 4(?) zoom level stops closest to the tile's zoom level. For each feature, calculate the value for the feature at each of these stops and add it to the buffer.
At render time, find the two stops the current zoom level is between and calculate the `t` interpolation value. This value is the same for all features because all features have stops at the same zoom levels. Pass this value as a uniform to the shader. The shader reads the values at the different stops as attributes and uses the uniform to interpolate the values.

@lucaswoj

lucaswoj:
Would that implementation only allow function stops at integer zoom levels? Does that matter from a design perspective? cc @peterqliu

ansis:
[...]

PR #2454, opened 2016-04-15
implement zoom and data driven functions

ansis:
fixes https://github.com/mapbox/mapbox-gl-js/issues/2428

For zoom+data driven functions, the buffers contain the values for four stops. Each shader reads all four values and uses a uniform, like `u_radius_t` to interpolate between the stops. Attributes with one component, like `radius`, are packed into a vec4. Attributes with more than one component are passed as separate glsl attributes.

Since functions may have more than four zoom stops the buffers may not hold all the stops. It adds the four nearest stops to the buffers (two stops before the tile's zoom, two stops after). The uniform used for interpolation accounts for this.

Zoom+data driven functions currently look like this:
```js
"circle-radius": {
    stops: [
        [0, { stops: [[0,  2], [100, 10]], property: 'mapbox' }], 
        [6, { stops: [[0,  20], [100, 100]], property: 'mapbox' }]
    ]   
},  
"circle-color": {
    stops: [
        [0, { stops: [[0, 'red'], [100, 'violet']], property: 'mapbox' }], 
        [6, { stops: [[0, 'blue'], [100, 'green']], property: 'mapbox' }]
    ]   
} 
```

The current implementation does not require that all the data subfunctions have the same stops or the same 'property' value. We may want to add that restriction to make UIs for editing functions simpler. Should these be valid? https://gist.github.com/ansis/625ed5b14e046a09b79a4ddb78cf214f 

Does this function definition look ok, or does anyone have other ideas? I'll fix style validation after we settle on a format.

https://github.com/mapbox/mapbox-gl-function/tree/zoom-and-data implements the mapbox-gl-function part of this.

:eyes: @lucaswoj @jfirebaugh


ansis:
example: http://bl.ocks.org/ansis/raw/00c2af09ba26e3e56c055a9ad24876b4/#1.5/38.9/-77.0

ansis:
@scothis how does the proposed function syntax look from a Studio perspective? what [restrictions](https://gist.github.com/ansis/625ed5b14e046a09b79a4ddb78cf214f) would help make the UI simpler?

lucaswoj:
I would prefer this simpler but equally expressive syntax:

```js
{
  'circle-radius': {
    stops: [
      [{$zoom: 0, mapbox: 0}, 2],
      [{$zoom: 0, mapbox: 100}, 10],
      [{$zoom: 6, mapbox: 0}, 20],
      [{$zoom: 6, mapbox: 100}, 100],
  }
}
```

scothis:
I lean toward the syntax @lucaswoj proposes because it follows the current key-value pattern we have now for stops. The difference is that we have a compound key.

However, the syntax does lead me to believe that the property being data-driven can change. For example:

```
{
  'circle-radius': {
    stops: [
      [{$zoom: 0, landarea: 0}, 2],
      [{$zoom: 0, population: 100}, 10],
      [{$zoom: 6, landarea: 0}, 20],
      [{$zoom: 6, population: 100}, 100],
  }
}
```

I might propose yet another option that enforces a single property to derive values from:

```
{
  'circle-radius': {
    property: 'landarea',
    stops: [
      [{$zoom: 0, $value: 0}, 2],
      [{$zoom: 0, $value: 100}, 10],
      [{$zoom: 6, $value: 0}, 20],
      [{$zoom: 6, $value: 100}, 100],
  }
}
```

Either syntax is workable if we apply the right validation constraints. It will be better to strongly enforce validation constraints early on and slowly relax them as needed than the other way around.

lucaswoj:
I would like to eventually support functions with stops based on multiple properties like:

```
[{$zoom: 0, population: 100, landarea: 0}, 2]
```

I agree that we should enforce homogeneous properties programmatically

ansis:
The main constraint would be that you need to have a stop for every pair of zoom and value used in other stops. If you imagine it as a spreadsheet with zoom on one axis and data values on the other axis then every cell needs a style value. In pseudo code:
```
zoomsUsedInStops = [0, 6];
valuesUsedInStops = [0, 100];
foreach z in zoomsUsedInStops:
    foreach v in valuesUsedInStops:
        assert(stop for { zoom: z, value: v } exists)
```

There would be exactly `zoomsUsedInStops.length * valuesUsedInStops.length` stops.

---

> I might propose yet another option that enforces a single property to derive values from:
> ```js
{
  'circle-radius': {
    property: 'landarea',
    stops: [
      [{$zoom: 0, $value: 0}, 2],
      [{$zoom: 0, $value: 100}, 10],
      [{$zoom: 6, $value: 0}, 20],
      [{$zoom: 6, $value: 100}, 100],
  }
}
```

This looks good to me. Maybe without the `$`. And maybe with `data` instead of `value`.


I slightly prefer syntaxes with nested arrays because it's closer to how the implementation works. The data functions are evaluated for a property to create a custom zoom function for each feature. The stops can be sorted into nested arrays easily so it's not a big difference.

---

> I would like to eventually support functions with stops based on multiple properties like:
> `[{$zoom: 0, population: 100, landarea: 0}, 2]`

I'm imagining the zoom+data functions as spreadsheets and 3d, 4d, etc spreadsheets are scary. I think the number of stops would get unwieldy really fast if stops support multiple properties. There might be [simpler ways](https://gist.github.com/ansis/491f86de2a0985c10548eb4fd59992e2) can handle these use cases.

I think it makes sense to stick with 2d functions for now.

**Comment by ansis** [comment]
spec and function pull requests:
https://github.com/mapbox/mapbox-gl-function/pull/16
https://github.com/mapbox/mapbox-gl-style-spec/pull/443

**Comment by lucaswoj** [comment]
> I think it makes sense to stick with 2d functions for now.

Let's think carefully about this spec so that we never have to break backwards compatibility. This should be orthogonal to what we want to do now vs later.

> I'm imagining the zoom+data functions as spreadsheets and 3d, 4d, etc spreadsheets are scary. I think the number of stops would get unwieldy really fast if stops support multiple properties. There might be [simpler ways](https://gist.github.com/ansis/491f86de2a0985c10548eb4fd59992e2) can handle these use cases.

I'm 👎 on a solution that only allows interpolation in some cases and only allows arbitrary functions in other cases. We should go all-in on interpolation or all-in on arbitrary functions. 

@scothis and I talked on Friday about the implications of supporting arbitrary functions in studio. He was in favor of arbitrary functions built atop the existing filter syntax. We could expose our existing interpolation functions as primitives to these functions:

```js
"circle-radius": [
    'exponential',            // function name
    [[0, 10], [20, 100]]     // stops
]
```

Spitballing additional syntax here:

```js
"circle-radius": [
    'exponential', // function name
    [[0, 10], [20, 100]], // stops
    '$zoom', // input
    2 // base
]
```

```js
"circle-radius": [
    'exponential', // function name
    [[0, 10], [20, 100]], // stops
    ['*', '$zoom', 'population'], // input
]
```

```js
"circle-radius: ['*', '$zoom', 'population']
```

```js
"circle-color: ['if',
    [['>=', 'population', '10000'], 'red']],
    ['else', 'blue']],
]
```

To support this, we'd probably have to move to a model of dynamically writing more shader code.

scothis:
> @scothis and I talked on Friday about the implications of supporting arbitrary functions in studio. He was in favor of arbitrary functions built atop the existing filter syntax.

I added a caveat that we will need the ability to have and detect pre-built functions in Studio. Most users should not need to drop to the low level syntax in order to do basic interpolation.

lucaswoj:
> I added a caveat that we will need the ability to have and detect pre-built functions in Studio. Most users should not need to drop to the low level syntax in order to do basic interpolation.

To clarify what this means: 

 - we support abitrary functions in the spec
 - we provide point & click UIs to create and edit common function patterns
 - we provide a raw JSON editor in Studio to create and edit complex functions

ansis:
I'm strongly 👎 on arbitrary functions that mix zoom and data inputs.

The current zoom+data implementation is designed so that the data part of the function can be evaluated separately from the zoom part. This let's us evaluate the data part in -js when creating the buffers and the zoom part in the shaders on every frame. The current limitations allow for highly optimized implementation, both in terms of shader complexity and buffer size.

For example, to evaluate `"circle-radius": ['*', '$zoom', 'population']` in a shader you'd need to store `population` in the buffers. With what type? what precision? How do deal with strings? And you'd also need complex code generation that you can't do with just the preprocessor language.

> I'm 👎 on a solution that allows only interpolation in some cases and allows only arbitrary functions in other cases.

I'm against 3d, 4d, etc interpolation because I can't think of a case where n-dimensional interpolation is what you actually want. I also think that in order for the interpolation to be unambiguous you would need to provide an exponential number of stops, which seems really unwieldy.

My [suggestion](https://gist.github.com/ansis/491f86de2a0985c10548eb4fd59992e2) was to use an expression of data properties as the input to the scale, instead of just a direct data property. There would be no conditionals or comparisons. Just math operators that combine data values. I don't think we need to do this yet.


> We should go all-in on interpolation or all-in on arbitrary functions.

We've already [committed to scales/interpolation for data driven functions](https://github.com/mapbox/mapbox-gl-js/pull/1932#issuecomment-209095582). It makes sense zoom+data functions work the same way. This is what's currently implemented.

The current implementation covers a lot of use cases and fits in with the existing spec. Any large changes that would be breaking for these zoom+data functions would likely also be breaking for the existing separate zoom functions and data functions. I think we should go with something similar to what's implemented here. It might be helpful to voice about this today.

lucaswoj:
I agree with all of your points above ☝️ 

I'm not particularly tied to any particular syntax specification for this feature but I have a couple requirements that aren't met by the nested function syntax:

 - backwards-compatible extensibility for more powerful function features 
 - style JSON readability & maintainability
 - encapsulation of implementation details
 - conceptual simplicity

Let's voice on this later today and see if we can figure out a function syntax specification that meets all of our requirements.

ansis:
>  the nested function syntax

Sorry, I never mentioned that I switched the syntax to the one proposed by you and @scothis. It now looks like this: https://github.com/mapbox/mapbox-gl-js/blob/8abd68872b5bd232cc75e66310da27bc91644adf/debug/site.js#L53-L69

js/data/bucket.js
```diff
@@ -336,36 +335,97 @@ function capitalize(string) {
...
     var attributes = {};
     for (var interfaceName in bucket.programInterfaces) {
-        var interfaceAttributes = attributes[interfaceName] = { layoutAttributes: [], paintAttributes: {} };
-        var layerPaintAttributes = interfaceAttributes.paintAttributes;
+        var layerPaintAttributes = attributes[interfaceName] = {};
 
         for (var c = 0; c < bucket.childLayers.length; c++) {
             var childLayer = bucket.childLayers[c];
-            layerPaintAttributes[childLayer.id] = { enabled: [], disabled: [] };
+            layerPaintAttributes[childLayer.id] = { enabled: [], uniforms: [] };
```
lucaswoj: Would it make more sense to stick to either a) `enabled`/`disabled` or b) `attributes`/`uniforms`?

js/style/style_function.js
```diff
@@ -8,7 +8,9 @@ exports.interpolated = function(parameters) {
         return inner(globalProperties && globalProperties.zoom, featureProperties || {});
     };
     outer.isFeatureConstant = inner.isFeatureConstant;
-    outer.isGlobalConstant = inner.isGlobalConstant;
+    outer.isZoomConstant = inner.isZoomConstant;
+    outer.getStopZoomLevels = inner.getStopZoomLevels;
```
lucaswoj: What led to your design decision to create a `getStopZoomLevels` method on `mapbox-gl-function` rather than parse these out within `mapbox-gl-js`?

js/style/style_layer.js
```diff
@@ -169,6 +169,20 @@ StyleLayer.prototype = util.inherit(Evented, {
...
+    getPaintValueStopZoomLevels: function(name) {
+        var transition = this._paintTransitions[name];
+        if (transition) {
+            return transition.declaration.getStopZoomLevels();
+        } else {
+            return [];
+        }
+    },
+
+    getPaintInterpolationValue: function(name, zoom) {
```
lucaswoj: What is an "interpolation value"?

js/data/bucket.js
ansis: I'll switch this to `attributes` / `uniforms`. Zoom + data functions create uniforms

js/style/style_function.js
ansis: I didn't consider parsing them out in mapbox-gl-js. Putting the function as member of the real function isn't great. Having it in mapbox-gl-function means -js doesn't need to know about the syntax of the functions though. Do you have any ideas on how this could be cleaned up?

lucaswoj: [`parseColor` knows about the syntax of functions](https://github.com/mapbox/mapbox-gl-js/blob/api-introduction-2461/js/style/parse_color.js#L18-L21)

When I'm trying to decide what goes into `mapbox-gl-js` and what goes into `mapbox-gl-function`, I ask myself if this is functionality that another downstream consumer, like Studio, would need to evaluate the function. The `getStopZoomLevels` method feels weirdly granular and mapbox-gl-js-specific. 

If you want to leave this as-is, I respect that decision.

js/style/style_layer.js
ansis: 2.333 would mean "interpolate 1/3 of the way from stop 2 to stop 3". It is passed to https://github.com/mapbox/mapbox-gl-js/pull/2454/files#diff-a8436c131e601f997ad9bb209e260f18R12
Names are hard. Any clearer ideas?

js/style/style_function.js
ansis: > The getStopZoomLevels method feels weirdly granular and mapbox-gl-js-specific.

It is. I'll split this out

ansis: split in 0c96d73bca5fd8dec4b8ef7a75f5db29e90255d9

js/style/style_layer.js
ansis: Two possible names:
`calculateInterpolationT`
`calculateInterpolationAmount`

I'll also move this out of mapbox-gl-function.

lucaswoj: Ah! I see. I'm not picky about the name as long as you're consistent with the name of the shader parameter. I found `calculateInterpolationT` enlightening for that reason.

lucaswoj:
All the code in this PR looks great!

# Next Steps

 - [ ] merge  mapbox/mapbox-gl-function#16
 - [ ] merge mapbox/mapbox-gl-style-spec#443
 - [ ] release new `mapbox-gl-function` version
 - [ ] release new `mapbox-gl-style-spec` version
 - [ ] add `mapbox-gl-test-suite` tests
 - [ ] update `package.json` with new `mapbox-gl-function` and `mapbox-gl-style-spec` versions & new `mapbox-gl-test-suite` SHA

shaders/circle.vertex.glsl
```diff
@@ -22,10 +31,32 @@ varying vec2 v_extrude;
...
+vec4 evaluate_zoom_function_4(const vec4 value0, const vec4 value1, const vec4 value2, const vec4 value3, const float t) {
+    if (t < 1.0) {
+        return mix(value0, value1, t);
+    } else if (t < 2.0) {
+        return mix(value1, value2, t - 1.0);
+    } else {
+        return mix(value2, value3, t - 2.0);
+    }
+}
+
```
mourner: If we want to bring this to all other shaders, it will start growing very complex and repetitive... Perhaps we should consider a GLSL build system like https://github.com/stackgl/glslify

lucaswoj: We could alternately use a simpler approach than `glslify` ref https://github.com/mapbox/mapbox-gl-js/commit/50f2cf5ec4200de0a327fee70216aa9816bffd9b#commitcomment-16958568

ansis: We could make a `zoom_functions.glsl` and prepend it to all vertex shaders.

dpieri:
How do I start using this? Will this be implemented in Studio? Documentation on how to do this via style JSON?

lucaswoj:
@dpieri This feature will be implemented in Studio once it has been expanded to all paint properties and is usable across both GL JS and GL Native platforms. Think of this like a dev preview for now.

lucaswoj:
The syntax we eventually decided upon is:

```js
{
  'circle-radius': {
    property: 'landarea',
    stops: [
      [{zoom: 0, value: 0}, 2],
      [{zoom: 0, value: 100}, 10],
      [{zoom: 6, value: 0}, 20],
      [{zoom: 6, value: 100}, 100],
  }
}
```

PR #2455, opened 2016-04-15
Animate point along route example

tristen:
animate-marker

cc @jfirebaugh @lucaswoj for review

### DIFF
--- a/docs/_posts/examples/3400-01-15-animate-point-along-route.html
+++ b/docs/_posts/examples/3400-01-15-animate-point-along-route.html
@@ -1,0 +1,162 @@
+---
+layout: example
+category: example
+title: Animate a point along a route
+description: "Use <a target='_blank' href='http://turfjs.org/'>Turf</a> to smoothly animate a point along the distance of a line."
+---
+<style>
+.overlay {
+    position: absolute;
+    top: 10px;
+    left: 10px;
+}
+
+.overlay button {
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Is this concealing the great circle vs. rhumb line difference via strategic choice of endpoints, or did you find another way to resolve that?

tristen:
> Is this concealing the great circle vs. rhumb line difference via strategic choice of endpoints, or did you find another way to resolve that?

Ahh you got me. point placement was strategic to avoid that. I wanted to avoid getting to complex here by adding code that just draws a linear point from one to the other.

jfirebaugh:
I think it's important that the example work for reasonable real-world modifications. How about adapting [this example](https://www.mapbox.com/mapbox.js/example/v1.0.0/arcjs/) and drawing the line as a great circle arc?

peterqliu:
@tristen since you're using turf to calculate transitional position anyway for the animation, you can use those waypoints to draw the linepath itself. Calculate once, draw the line, animate markers to new positions on the line.

I believe that achieves the same thing as arc.js, but with simpler control of increment distance (km/miles vs fractions of the whole).

tristen:
Okie dokie updated to draw an arc first then animate a point along it:

animate-point-along-route-example

jfirebaugh:
👍

tristen:
https://github.com/mapbox/mapbox-gl-js/commit/ce48e3ef36d0ecec009a5ccf12f0b6b19723d8d7

PR #2460, opened 2016-04-18
Guard transition.declaration being undefined

scothis:
We've seen many cases where the declaration object is undefined for a
transitions in Studio. Adding a guard.

### DIFF
--- a/js/style/style_layer.js
+++ b/js/style/style_layer.js
@@ -259,7 +259,7 @@ StyleLayer.prototype = util.inherit(Evented, {
             declaration = new StyleDeclaration(spec, spec.default);
         }
 
-        if (oldTransition && oldTransition.declaration.json === declaration.json) return;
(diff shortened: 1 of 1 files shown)


scothis:
This is a deeper issue if `declaration` should never be `undefined`

scothis:
https://app.getsentry.com/mapboxcom/studio-production/issues/122904433/

scothis:
This appears to be related to unsetting (set to undefined) of paint properties.

lucaswoj:
`transition.declaration` should never be undefined. This appears to be a regression introduced [here](https://github.com/mapbox/mapbox-gl-js/commit/a9263717c771f0b332c757162a0e6cd8e0b73212). Instead of checking `value === null`, this line should check `value === null || value === undefined`.

lucaswoj:
--> #2464

scothis:
👍

PR #2466, opened 2016-04-19
Fix local ajax calls

srgpqt:
Loading json, images or sprites from the local system will fire XMLHttpRequest onload event with status 0.

XMLHttpRequest onload event is only fired when the request completed successfully, there is no need for the status check -- so the status check should be removed.

### DIFF
--- a/js/util/browser/ajax.js
+++ b/js/util/browser/ajax.js
@@ -8,7 +8,8 @@ exports.getJSON = function(url, callback) {
         callback(e);
     };
     xhr.onload = function() {
-        if (xhr.status >= 200 && xhr.status < 300 && xhr.response) {
+        // xhr.status does not need to be checked because onload implies success
+        if (xhr.response) {
             var data;
             try {
                 data = JSON.parse(xhr.response);
@@ -32,7 +33,8 @@ exports.getArrayBuffer = function(url, callback) {
         callback(e);
     };
     xhr.onload = function() {
-        if (xhr.status >= 200 && xhr.status < 300 && xhr.response) {
+        // xhr.status does not need to be checked because onload implies success
+        if (xhr.response) {
             callback(null, xhr.response);
         } else {
             callback(new Error(xhr.statusText));


lucaswoj:
Thanks for the contribution @srgpqt! 

My reading of the available documentation and standards indicates that the `onload` event is fired when the HTTP request completes successfully but doesn't necessarily indicate that status code is 2xx. 

Perhaps we could solve the underlying problem by allowing a status code of 0 if the protocol is `file://`. I am hesitant to go this route because there may be additional challenges to supporting the `file://` protocol and running a local HTTP server is very easy. 

Thoughts @ansis @jfirebaugh @mourner?

jfirebaugh:
`onload` does not imply `xhr.status >= 200 && xhr.status < 300`. It simply means that a response was received.

We don't support using mapbox-gl-js from `file://` URLs -- the 0 XHR status is only one of several XHR/CORS related quirks. Please use a local http server. Here are some options: http://sweetme.at/2014/04/29/simple-local-http-servers/

srgpqt:
mapbox-gl-js works fine for us once the status check is removed.

From the relevant mozilla documentation :
https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIXMLHttpRequestEventTarget

* onload    A JavaScript function object that gets invoked when the operation is successfully completed.
* onloadend    A JavaScript function object that gets invoked when the operation is completed for any reason; it will always follow a an abort, error, or load event.

So onload does seem to indicate success, wheras the old onreadystatechange event did fire regardless of success/failure condition.

srgpqt:
Also, we were using a local web server but it's no longer needed with this change.

Less junk to get the application running is a Good Thing™

jfirebaugh:
```
var xhr = new XMLHttpRequest();
xhr.onload = function() { console.log(xhr.status); }
xhr.open('GET', 'http://httpstat.us/400', true);
xhr.send();
```

This logs 400.

srgpqt:
Indeed. I guess the docs are wrong then.

Ah well, I was hoping to get rid of the need for WKWebView in our app but it seems that UIWebView performs significantly worse when using mapbox-gl.

srgpqt:
Another note : setting responseType to "json" allows us to check for success with the presence of the "response" property. This logs null :

var xhr = new XMLHttpRequest();
xhr.onload = function() { console.log(xhr.response); }
xhr.responseType = 'json';
xhr.open('GET', 'http://httpstat.us/400', true);
xhr.setRequestHeader('Accept', 'application/json');
xhr.send();

PR #2471, opened 2016-04-19
Add "renderstable" event

### LINKED ISSUE #1715: Add datastart and dataend events
There are events for when individual tiles load, but it's difficult to know when all the tiles for a source have finished loaded. Essentially, I'd like an API to know if tiles requests are in flight, or if the all the data I would expect to view for a given view is already loaded. Firing an event when the state transitions would also be helpful.

Knowing when a source is fully loaded is useful for smoothly transitioning between different overlay sources without a flash as tiles load.

As a quick sketch, I'd imaging something like:

```
map.on('source.loading', function (e) {
  e.status; // one of 'fetching', 'idle', 'errored'
});
```

lucaswoj:
_This pull request isn't quite ready. I am publishing now to get some 👀  on the proposal and implementation._

Mapbox GL JS downloads many resources (glyphs, sprites, tiles, and styles) over the network. Giving our developers the right primitives to understand the state of these resources and respond to changes in that state is a hard problem. 

To that end, this PR adds the `data` and `dataend` events:

 - **`data` event** a resource has been loaded, changed, or removed. 
 - **`dataend` event** all resources for the current viewport have been loaded. Fired at most once per frame.

This PR also renames `Map#loaded` to `isDataStable`, to disambiguate the concept of "data stable" (all resources for the current viewport have been loaded) from "loaded" (all resources for the FIRST viewport have been loaded).

We could do more to simplify the existing events infrastructure. Once this pull request is merged, we should consider deprecating some granular `*.add`, `*.remove`, `*.load`, and `*.change` events.

fixes #1715 

cc @ansis @jfirebaugh @mourner @tmcw @scothis

### DIFF
--- a/debug/events.html
+++ b/debug/events.html
@@ -189,7 +189,8 @@ var checkboxContainer = document.querySelector('#checkboxes');
  'zoomstart', 'zoom', 'zoomend',
  'rotatestart', 'rotate', 'rotateend',
  'pitchstart', 'pitch', 'pitchend',
- 'boxzoomstart', 'boxzoomend', 'boxzoomcancel'
+ 'boxzoomstart', 'boxzoomend', 'boxzoomcancel',
+ 'renderstable'
 ].forEach(function (type) {
     var name = 'show-' + type;
     var input = document.createElement('input');
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -793,18 +793,11 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
-     * Is this map fully loaded? If the style isn't loaded
-     * or it has a change to the sources or style that isn't
-     * propagated to its style, return false.
-     *
-     * @returns {boolean} whether the map is loaded
+     * @returns {boolean} True if all the resources needed to display the
+     *   current viewport have been loaded.
      */
-    loaded: function() {
-        if (this._styleDirty || this._sourcesDirty)
-            return false;
-        if (!this.style || !this.style.loaded())
-            return false;
-        return true;
+    isRenderStable: function() {
+        return (!this._styleDirty && !this.sourcesDirty && this.style && this.style.loaded());
     },
 
     /**
@@ -854,9 +847,18 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
 
         this.fire('render');
 
-        if (this.loaded() && !this._loaded) {
-            this._loaded = true;
-            this.fire('load');
+        if (this.isRenderStable()) {
+            if (!this._wasLoaded) {
+                this._wasLoaded = true;
+                this.fire('load');
+            }
+
+            if (!this._wasRenderStable) {
+                this._wasRenderStable = true;
+                this.fire('renderstable');
+            }
+        } else {
+            this._wasRenderStable = false;
         }
 
         this._frameId = null;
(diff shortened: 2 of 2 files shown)


mourner:
```
TypeError: this.sprite.isDataStable is not a function
```

lucaswoj:
@mourner Sorry about that! Fixed now.

jfirebaugh:
For the `data` event, I propose that it be fired exactly once for each load, modification, or removal of a distinct resource, and that the event object contain information about the resource -- at a minimum, the resource type and action. Since the proposal is to remove type- and action-specific events like `tile.load`, this would ensure that what you can currently do with the existing events, you can also do with the all-purpose `data` event.

lucaswoj:
fun state diagram 

screen shot 2016-04-20 at 2 28 56 pm

lucaswoj:
I have modified my approach, adding a single new event: `renderstable`. This event is fired whenever the value of `Map#isRenderStable()` changes to true.

`Map#isRenderStable()` (formerly `Map#loaded`) returns true iff every resource needed to render the current viewport has been loaded.

lucaswoj:
👀 @jfirebaugh @mourner @ansis

jfirebaugh:
Last time we discussed this, the consensus was that this event should be called `dataend`. I think `dataend` is a better name for the behavior you've implemented. I would expect `renderstable` not to be fired if there's an active gesture, easing animation, or style transition, and these changes don't implement that.

jfirebaugh:
Hmm, actually, I think I might be misremembering our conversation.

lucaswoj:
# Next Stems

 - [ ] unrename `isRenderStable`
 - [ ] rename `renderstable` to `renderend`
 - [ ] add `loadend` event

jfirebaugh:
Yeah, so I'm reversing my earlier proposal of having "render" in the event name somewhere. Two reasons:

* I think we should reserve names like `renderstable` or `renderend` for the case where the map isn't going to render any more frames until perturbed. Such an event should take into consideration animated easing, property transitions, and fading behaviors.
* On Friday there was talk about this event being fired "iff `queryRenderedFeatures` changes", which is what led me to suggest the `renderend` name. But the event implemented here isn't fired on small pans that don't load new data, so "iff `queryRenderedFeatures` changes" isn't actually its definition.

I'm not sure "iff `queryRenderedFeatures` changes" is the right definition anyway -- it could be pretty complex to implement that. The simpler event that's implemented here is "all data for the current viewport is loaded". I think this is a good, useful event. I'm going to go back to calling it `dataend` for now.

With `dataend`, the sidebar can still be updated on small pans that don't load new data, by combining event handlers:

```JS
map.on('dataend', renderSidebar);
map.on('moveend', function () {
    if (map.isFullyLoaded()) {
       renderSidebar();
    }
});
```

What else can cause `queryRenderedFeatures` to change besides data loading and the map moving? Changing style properties, for one. The user can handle that by calling `renderSidebar` after making relevant style property changes. Anything else?

jfirebaugh:
A thought on naming: conceptually, this could be considered a family of three events:

* An event that is fired once, for the lifetime of the map (currently named `load`).
* An event that is fired once per perturbation that causes new data to be loaded (new event).
* An event that is fired any number of times, whenever a new "chunk" of data has been loaded (currently various `*.add` / `*.load` / `*.change` events, proposed to be unified in the future).

We don't necessarily have to have symmetry in the three names, but it would be nice if that fell out naturally.

lucaswoj:
> * An event that is fired once, for the lifetime of the map (currently named `load`).
> * An event that is fired once per perturbation that causes new data to be loaded (new event).
> * An event that is fired any number of times, whenever a new "chunk" of data has been loaded (currently various `*.add` / `*.load` / `*.change` events, proposed to be unified in the future).

Do these events correlate with my proposal of `data`, `dataend` and `load` events?

lucaswoj:

GertjanVandenKeybus:
Any updates on this feature? I need to upload a map with HTTPS and I need to know when to actually start my data upload.

PR #2475, opened 2016-04-20
add a cpu frame duration benchmark

ansis:
Why have this when we already have a fps benchmark?
- if a particular example is bottlnecked by the gpu on particular hardware then cpu changes that are relevant on other devices/styles or that will be after [future gpu optimizations](https://github.com/mapbox/mapbox-gl-js/pull/1606) might not affect the results
- actual frames are often synchronized with a screen refresh rate. If every frame takes exactly 17ms to render it might have to spend 15ms waiting around doing nothing. FPS can be a staircase function of frame duration
- this measures the duration more directly so I think it should be more accurate

screen shot 2016-04-20 at 12 35 00 pm

:eyes: @lucaswoj

### DIFF
--- a/bench/benchmarks/frame_duration.js
+++ b/bench/benchmarks/frame_duration.js
@@ -1,0 +1,110 @@
+'use strict';
+
+var Evented = require('../../js/util/evented');
+var util = require('../../js/util/util');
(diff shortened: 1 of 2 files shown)


lucaswoj:
🚢! Nice work.

PR #2483, opened 2016-04-21
use VertexArrayObject extension when available

ansis:
fixes #1361

When the extension is available, use it to record and re-apply buffer binds and attrib pointer setting. When the extension isn't available this just sets the pointers and binds the buffers each time.

This also namespaces bucket buffers/arrays with objects instead of string concatenation.

I ran the frame-duration benchmark four times for master (3a95d00) and this branch:

```
vao:        15.3 ms
master:     17.9 ms
vao:        15.5 ms
master:     18.2 ms
vao:        14.2 ms
master:     18.5 ms
vao:        15.2 ms
master:     17.4 ms

vao average:    15 ms
master average: 18 ms
```

This branch takes **83%** of the time that master takes.

:eyes: @lucaswoj

PR #2488, opened 2016-04-22
eliminate element groups and optimize some things

ansis:
- replaces element groups with buffer groups. Each buffer group has it's own buffers.
- cleans up and optimizes some things in drawSymbol
- makes frame-duration benchmark more independent of gpu performance

```
master:         11.7 ms
this branch:    8.6 ms
master:         11.3 ms
this branch:    8.3 ms
master:         11.9 ms
this branch:    9.3 ms

master average:         11.63 ms
this branch average:    8.7 ms
```

This branch (which includes the [unmerged vao changes](https://github.com/mapbox/mapbox-gl-js/pull/2483)) takes **75%** of the time it takes master to run through the render loop.

:eyes: @lucaswoj

### DIFF
--- a/bench/benchmarks/frame_duration.js
+++ b/bench/benchmarks/frame_duration.js
@@ -6,6 +6,16 @@ var formatNumber = require('../lib/format_number');
 
 var DURATION_MILLISECONDS = 1 * 5000;
 
+// The goal of this benchmark is to measure the time it takes to run the cpu
+// part of rendering. While the gpu rendering happens asynchronously, sometimes
+// when the gpu falls behind the cpu commands synchronously wait for the gpu to catch up.
+// This ends up affecting the duration of the call on the cpu.
+//
+// Setting the devicePixelRatio to a small number makes the canvas very small.
+// This greatly reduces the amount of work the gpu needs to do and reduces the
+// impact the actual rendering has on this benchmark.
+window.devicePixelRatio = 1 / 16;
+
 var zooms = [4, 8, 11, 13, 15, 17];
 var results = [];
 
@@ -20,7 +30,7 @@ module.exports = function(options) {
         measureFrameTime(options, zooms[index], function(err_, result) {
             results[index] = result;
             evented.fire('log', {
-                message: formatNumber(result.sum / result.count) + ' ms per frame at zoom ' + zooms[index] + '. ' +
+                message: formatNumber(result.sum / result.count * 10) / 10 + ' ms per frame at zoom ' + zooms[index] + '. ' +
                     formatNumber(result.countAbove16 / result.count * 100) + '% of frames took longer than 16ms.'
             });
             callback();
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -75,34 +75,28 @@ function Bucket(options) {
 
     this.paintAttributes = createPaintAttributes(this);
 
-    if (options.elementGroups) {
-        this.elementGroups = options.elementGroups;
+    if (options.arrays) {
+        var childLayers = this.childLayers;
+        this.bufferGroups = util.mapObject(options.arrays, function(programArrayGroups, programName) {
+            return programArrayGroups.map(function(programArrayGroup) {
 
-        for (var programName in this.elementGroups) {
-            var elementGroups = this.elementGroups[programName];
-            if (elementGroups) {
-                for (var i = 0; i < elementGroups.length; i++) {
-                    var elementGroup = elementGroups[i];
-                    elementGroup.vaos = {};
-                    elementGroup.secondVaos = {};
-                    for (var l = 0; l < this.childLayers.length; l++) {
-                        var layerName = this.childLayers[l].id;
-                        elementGroup.vaos[layerName] = new VertexArrayObject();
-                        if (elementGroup.secondElementLength) {
-                            elementGroup.secondVaos[layerName] = new VertexArrayObject();
-                        }
-                    }
-                }
-            }
-        }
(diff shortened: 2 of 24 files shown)


bench/benchmarks/frame_duration.js
```diff
@@ -6,6 +6,8 @@ var formatNumber = require('../lib/format_number');
 
 var DURATION_MILLISECONDS = 1 * 5000;
 
+window.devicePixelRatio = 1 / 16;
```
lucaswoj: What is your intention behind this line? (Answer probably best delivered as a `// comment` in the source)

js/data/bucket.js
```diff
@@ -132,26 +127,32 @@ Bucket.prototype.populateBuffers = function() {
...
+        if (ElementArrayType) currentGroup.layout.element = new ElementArrayType();
+        if (SecondElementArrayType) currentGroup.layout.secondElement = new SecondElementArrayType();
+
+        for (var i = 0; i < this.childLayers.length; i++) {
+            var layerName = this.childLayers[i].id;
+            var PaintVertexArrayType = arrayTypes.paint[layerName];
+            currentGroup.paint[layerName] = new PaintVertexArrayType();
+        }
+
         groups.push(currentGroup);
```
lucaswoj: The terminology in here is a little confusing. We've "eliminated element groups" but still have an entity that we're referring to as a "group" but putting into a property called "array." Is there an easy way to shore up the terminology here?

js/data/bucket/symbol_bucket.js
```diff
@@ -339,23 +339,20 @@ SymbolBucket.prototype.placeFeatures = function(collisionTile, showCollisionBoxe
...
+
+    this.metadata = {
         sdfIcons: this.sdfIcons,
-        iconsNeedLinear: this.iconsNeedLinear
+        iconsNeedLinear: this.iconsNeedLinear,
+        adjustedTextSize: this.adjustedTextSize,
+        adjustedIconSize: this.adjustedIconSize,
+        // Transfer the name of the fonstack back to the main thread along with the buffers.
+        // The draw function needs to know which fonstack's glyph atlas to bind when rendering.
+        fontstack: layout['text-font'].join(',')
```
lucaswoj: Why introduce the concept of a metadata object rather than make these properties on `SymbolBucket`?

js/render/draw_symbol.js
```diff
@@ -33,165 +31,181 @@ function drawSymbols(painter, source, layer, coords) {
...
+function drawLayerSymbols(painter, source, layer, coords, isText,
+        translate,
+        translateAnchor,
+        rotationAlignment,
+        size,
+        haloWidth,
+        haloColor,
+        haloBlur,
+        opacity,
+        color) {
```
lucaswoj: The large number of arguments makes this code less readable. Would you consider using an `options` object here?

lucaswoj:
Just a couple syntactic-level comments. This looks great! Thank you for making our codebase more robust and faster.

js/data/bucket/symbol_bucket.js
ansis: These are the properties that need to be serialized. Putting them in an objects makes it unnecessary to serialize/deserialize them individually. This is good because the serialization code is in Bucket and these individual properties are specific to SymbolBucket

js/data/bucket.js
ansis: Would renaming `this.arrays` to `this.arrayGroups` fix this?

js/render/draw_symbol.js
ansis: I agree that the large number of parameters is awkward but I prefer this to an options object because:
- this lists what parameters the function accepts. With an options object you need either document it separately or look where the function is called.
- none of these parameters are optional. Putting them in an options object is more likely to hide typos since they'll just be `undefined` instead of jslint errors. On the other hand this approach could suffer from ordering bugs.


I prefer keeping this as is

bench/benchmarks/frame_duration.js
ansis: done f047ba1d61799ec43ddf07d6a8b432eecc00b689

js/data/bucket.js
ansis: done 2d6394d2065591763ca09924179e23fd126b5f49

ansis:
@lucaswoj want to review the fixes?

js/data/bucket.js
lucaswoj: Yes. Looks good!

js/render/draw_symbol.js
lucaswoj: 👍

js/data/bucket/symbol_bucket.js
lucaswoj: The difference between `this.metadata[key]` properties and `this[key]` properties is that `this.metadata[key]` properties are specific to a particular subclass of `Bucket`? That strikes me as unconventional software architecture. 

> This is good because the serialization code is in Bucket and these individual properties are specific to SymbolBucket

I would handle this by creating a `SymbolBucket#serialize` method that calls `Bucket.prototype.serialize`

ansis: > I would handle this by creating a SymbolBucket#serialize method that calls Bucket.prototype.serialize

done 7758c817057868be8fb925269eb6c73d0b0b3e29

lucaswoj:
Thanks @ansis! 🚢

ansis:
merged into master

PR #2489, opened 2016-04-22
add overdraw inspector

ansis:
The overdraw inspector shows how many time each fragment has been shaded. White fragments were shaded 8 or more times. Black fragments were shaded 0 times. Lower is better.

This only counts shaded fragments that were written to the color buffer. Fragments that are written only to the stencil buffer are not included.

`map.showOverdrawInspector = true`

screen shot 2016-04-22 at 4 15 09 pm
mapbox-streets landuse overdraws a lot:
screen shot 2016-04-22 at 4 15 31 pm
mapbox-streets terrain overdraws a lot:
screen shot 2016-04-22 at 4 16 15 pm

:eyes: @kkaefer @mourner @lucaswoj

### DIFF
--- a/js/render/painter.js
+++ b/js/render/painter.js
@@ -64,7 +64,6 @@ Painter.prototype.resize = function(width, height) {
 
 };
 
-
 Painter.prototype.setup = function() {
     var gl = this.gl;
 
@@ -225,6 +224,8 @@ Painter.prototype.render = function(style, options) {
     this.clearColor();
     this.clearDepth();
 
+    this.showOverdrawInspector(options.showOverdrawInspector);
+
     this.depthRange = (style._order.length + 2) * this.numSublayers * this.depthEpsilon;
 
     this.renderPass({isOpaquePass: true});
@@ -255,7 +256,9 @@ Painter.prototype.renderPass = function(options) {
         }
 
         if (isOpaquePass) {
-            this.gl.disable(this.gl.BLEND);
+            if (!this._showOverdrawInspector) {
+                this.gl.disable(this.gl.BLEND);
+            }
             this.isOpaquePass = true;
         } else {
             this.gl.enable(this.gl.BLEND);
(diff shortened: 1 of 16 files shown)


js/render/painter.js
```diff
@@ -374,3 +377,20 @@ Painter.prototype.setExMatrix = function(exMatrix) {
 Painter.prototype.lineWidth = function(width) {
     this.gl.lineWidth(util.clamp(width, this.lineWidthRange[0], this.lineWidthRange[1]));
 };
+
+Painter.prototype.overdrawInspector = function(enabled) {
```
lucaswoj: Can we make this method a verb instead of a noun? i.e. `setOverdrawInspector`

shaders/circle.fragment.glsl
```diff
@@ -10,4 +10,8 @@ varying lowp float v_antialiasblur;
 void main() {
     float t = smoothstep(1.0 - max(u_blur, v_antialiasblur), 1.0, length(v_extrude));
     gl_FragColor = v_color * (1.0 - t) * u_opacity;
+
+#ifdef OVERDRAW_INSPECTOR
+    gl_FragColor = vec4(1.0);
+#endif
```
lucaswoj: Is there any way to avoid setting `gl_FragColor` and then overwriting `gl_FragColor`?

 - changing the color and opacity passed in the `draw_*` methods
 - putting the other set in a `#ifndef` block

js/render/painter.js
ansis: I'm going to change this to `showOverdrawInspector` to match the external property.

shaders/circle.fragment.glsl
ansis: > changing the color and opacity passed in the draw_* methods

This isn't easily doable for programs that read from textures

> putting the other set in a #ifndef block

Yes, this could be done, but I'd prefer not to. Functionally, both are the same. Using #ifndef would make the overdraw code more intertwined with the rest of the shader and not have any benefit.

I'll leave these lines as they are

PR #2491, opened 2016-04-25
[wip] Improve documentation: examples, links, more

tmcw:


tmcw:
Okay: these are all moderate changes that don't require anything more about documentation.js and don't change the behavior of Mapbox GL JS. Going to merge once tests pass and then start thinking bigger-picture.

PR #2495, opened 2016-04-26
Allow filter to be mutated and set again

### LINKED ISSUE #2486: setFilter won't update map if you update the same array that you gave the map
```js
var filter = ['in', 'thing', 'value'];
map.setFilter('layer', filter);
filter.push('otherValue');
map.setFilter('layer', filter);
```

This breaks because the map doesn't copy the filter array and then runs deepEqual on its mutated version.

lucaswoj:
fixes #2486

cc @tmcw @ansis

### DIFF
--- a/js/style/style_declaration.js
+++ b/js/style/style_declaration.js
@@ -2,13 +2,14 @@
 
 var MapboxGLFunction = require('./style_function');
 var parseColor = require('./parse_color');
+var util = require('../util/util');
 
 module.exports = StyleDeclaration;
 
 function StyleDeclaration(reference, value) {
     this.type = reference.type;
     this.transitionable = reference.transition;
-    this.value = value;
+    this.value = util.clone(value);
     this.isFunction = !!value.stops;
 
     // immutable representation of value. used for comparison
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -501,7 +501,7 @@ Style.prototype = util.inherit(Evented, {
         if (this._handleErrors(validateStyle.filter, 'layers.' + layer.id + '.filter', filter)) return this;
 
         if (util.deepEqual(layer.filter, filter)) return this;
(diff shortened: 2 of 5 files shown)


ansis:
looks good to me. Are there any other objects or arrays we should be cloning when they are passed in from the style api?

tmcw:
Filters can be nested and this only slices the top-level - if you use an operator like 'all' and modify one of the cases, I think the same bug will still be present.

lucaswoj:
Right you are, @tmcw! Fixed in latest commit.

lucaswoj:
@ansis I'm not 100% sure that this fault could turn into a bug but we don't presently clone style functions passed to `addLayer` or `setLayoutProperty` or `setPaintProperty`. Fixed in the next commit.

js/style/style_declaration.js
```diff
@@ -8,7 +8,7 @@ module.exports = StyleDeclaration;
 function StyleDeclaration(reference, value) {
     this.type = reference.type;
     this.transitionable = reference.transition;
-    this.value = value;
+    this.value = util.clone(value);
```
mourner: forgot to require `util`

mourner:
I wonder if we should instead just drop the `deepEquals` check — this might or might not be a better performance tradeoff than doing full cloning (which may be expensive for big filters).

lucaswoj:
Eliminating the `deepEquals` and `clone` calls would be better for performance. I worry, however, about the unexpected behaviour that would result from mutating the filter after setting it.

```js
var filter = ['in', 'class', 'highway'];
map.setFilter('roads', filter);
map.style.update(...);
filter.push('residential');
```

After these calls, `map.getFilter('roads')` would return `['in', 'class', 'highway', 'residential']` but no `residential` roads would be visible on the map or returned from `queryRenderedFeatures`

PR #2496, opened 2016-04-26
Allow updatePaintTransition to be called with extraneous classes

### LINKED ISSUE #2477: Updating paint properties on layers without classes when a class is active throws exception
**mapbox-gl-js version**: 0.17.0

### Steps to Trigger Behavior

 1. Add some layers to the map that do not contain classes
 2. Add a class to the map
 3. Set the paint property of a layer without a class

### Expected Behavior
That the regular paint properties (without classes) are updated. This has worked fine up until 0.17.0 and I guess it's either a bug in the new automatic batching behaviour, or the API has changed on this part.

### Actual Behavior
An exception is thrown.

Demo: http://codepen.io/anon/pen/JGwerB

lucaswoj:
fixes #2477

cc @averas @ansis

### DIFF
--- a/js/style/style_layer.js
+++ b/js/style/style_layer.js
@@ -231,7 +231,10 @@ StyleLayer.prototype = util.inherit(Evented, {
     updatePaintTransition: function(name, classes, options, globalOptions, animationLoop) {
         var declaration = this._paintDeclarations[''][name];
         for (var i = 0; i < classes.length; i++) {
-            declaration = this._paintDeclarations[classes[i]][name] || declaration;
+            var classPaintDeclarations = this._paintDeclarations[classes[i]];
+            if (classPaintDeclarations && classPaintDeclarations[name]) {
+                declaration = classPaintDeclarations[name];
+            }
         }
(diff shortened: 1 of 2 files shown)


js/style/style_layer.js
```diff
@@ -231,7 +231,9 @@ StyleLayer.prototype = util.inherit(Evented, {
     updatePaintTransition: function(name, classes, options, globalOptions, animationLoop) {
         var declaration = this._paintDeclarations[''][name];
         for (var i = 0; i < classes.length; i++) {
-            declaration = this._paintDeclarations[classes[i]][name] || declaration;
+            if (this._paintDeclarations[classes[i]] && this._paintDeclarations[classes[i]][name]) {
+                declaration = this._paintDeclarations[classes[i]][name];
```
mourner: Now that `this._paintDeclarations[classes[i]]` repeats 3 times, let's put it in a variable.

lucaswoj: 👍 fixed in the latest commit

PR #2498, opened 2016-04-26
Intro updates

lyzidiamond:
Updates the intro:

- updates to language
- updated code sample and link

cc @tmcw @jfirebaugh @lucaswoj

### DIFF
--- a/docs/_layouts/docs.html
+++ b/docs/_layouts/docs.html
@@ -35,8 +35,26 @@ options: full
 <script src='{{site.tileApi}}/mapbox-gl-js/{{site.version}}/mapbox-gl.js'></script>
 <link href='{{site.tileApi}}/mapbox-gl-js/{{site.version}}/mapbox-gl.css' rel='stylesheet' />{% endhighlight html %}
     </div>
+    <div class='pad1x small quiet space-bottom1'>
+      Include the JavaScript and CSS files in the <code>&lt;head&gt;</code> of your HTML file.
+    </div>
+
+    <div class='fill-white space-bottom1'>
+{% capture getting_started_html %}
+<div id='map' style='width: 400px; height: 300px;'></div>
+<script>
+var map = new mapboxgl.Map({
+    container: 'map',
+    style: 'mapbox://styles/mapbox/streets-v8'
+});
+</script>
+{% endcapture %}
+
+      {% highlight html %}{{ getting_started_html | insert_token }}{% endhighlight %}
+    </div>
     <div class='pad1x small quiet'>
-    Include the Mapbox GL JS and CSS files in your HTML header and use <a href='#Map'><code>mapboxgl.Map()</code></a> to load your first map. <a class='truncate rcon next' href='{{site.baseurl}}/examples/'>View basic example</a>
+      Include this code in the <code>&lt;body&gt;</code> of your HTML file to initialize a map on the page.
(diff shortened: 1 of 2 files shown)


lyzidiamond:
I don't think this test fail is related to anything I did?

tmcw:
👍 lgtm

docs/_layouts/docs.html
```diff
@@ -18,39 +18,29 @@ <h1 class='space-bottom2'>
...
+    <div class='fill-white space-bottom1'>{% highlight html %}
+<script src='{{site.tileApi}}/mapbox-gl-js/{{site.version}}/mapbox-gl.js'></script>
+<link href='{{site.tileApi}}/mapbox-gl-js/{{site.version}}/mapbox-gl.css' rel='stylesheet' />{% endhighlight html %}
     </div>
-  </section>
+    <div class='pad1x small quiet'>
+    Include the Mapbox GL JS and CSS files in your HTML header and use <a href='#Map'><code>mapboxgl.Map()</code></a> to load your first map. <a class='truncate rcon next' href='{{site.baseurl}}/examples/'>View basic example</a>
+    </div>
+  </div>
+</div>
```
lucaswoj: My intent with this section was to provide all the code necessary to get an instance of GL JS running: JavaScript, CSS, and HTML. It is less useful without all three pieces because a reader will need to consult additional documentation in order to get an instance of GL JS running. We should provide all three pieces or else omit this section entirely.

lyzidiamond: Disagree. There are several reasons we shouldn't have full code here:

- it assumes that the only thing on the page is a map, when in many cases users are adding maps to their existing pages
- it's opinionated and makes it unclear which parts of the code are necessary and which parts are editorial
- it doesn't do a good job of setting expectations -- by including a link to examples, we are moving users through the flow they will be taking through most of their development
- the landing page for API docs should not be a quick start; it makes the purpose of the page muddled/unclear and creates confusion for users

jfirebaugh: > the landing page for API docs should not be a quick start

I think there's some confusion and conflicting requirements here because this page lives at https://www.mapbox.com/mapbox-gl-js/api/, but the content here is more appropriate for https://www.mapbox.com/mapbox-gl-js/.

https://www.mapbox.com/mapbox-gl-js/ currently redirects to https://www.mapbox.com/mapbox-gl-js/api/. We should unredirect it and have two distinct pages with clear responsibilities:

* https://www.mapbox.com/mapbox-gl-js/ is a developer-oriented landing page that briefly explains Mapbox GL JS and provides prominent links to guides, examples, API documentation, and the GitHub project
* https://www.mapbox.com/mapbox-gl-js/api/ is API documentation

lucaswoj: Correct. I envision this as a stopgap solution until we create a proper https://www.mapbox.com/mapbox-gl-js landing page.

lucaswoj: If you're both against including this content, we should 🚢.

lyzidiamond: Let's merge here and test it out with analytics. If we're not seeing a large number of people clicking the link in the next couple weeks, we can add the code back. Sound alright @lucaswoj @tmcw?

lucaswoj:
🚢!

PR #2503, opened 2016-04-27
split up uniforms to prepare for data driven styling

ansis:
Combine line-width and line-gap-width in shader so that the input variables are independent of each other.

Split up `u_pattern_scale` into a bunch of uniforms so that all the style property inputs are independent of the inputs based on the zoom and currently calculated interpolation value.

:eyes: @lucaswoj


lucaswoj:
Thank you @ansis! 🚢

PR #2512, opened 2016-04-28
Documentation fixes

tmcw:
* GeoJSON source options object is optional, but the object
  for ImageSource and VideoSource is not.
* ImageSource serialization was broken because we were saving a urls
  property, not a url property

cc @lucaswoj for the review

### DIFF
--- a/js/source/image_source.js
+++ b/js/source/image_source.js
@@ -14,7 +14,7 @@ module.exports = ImageSource;
 /**
  * Create an Image source instance given an options object
  * @class ImageSource
- * @param {Object} [options]
+ * @param {Object} options
  * @param {string} options.url A string URL of an image file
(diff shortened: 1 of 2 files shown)


js/source/image_source.js
```diff
@@ -31,7 +31,7 @@ module.exports = ImageSource;
  * map.removeSource('some id');  // remove
  */
 function ImageSource(options) {
-    this.urls = options.urls;
+    this.url = options.url;
```
lucaswoj: ⚠️  The documentation was wrong but this is still a breaking change. ⚠️ 

Can we avoid the headaches by writing `this.url = options.url || options.urls;`?

tmcw: I'm pretty sure this is just a bugfix - sure, we'd serialize urls instead of url, but you couldn't rely on that behavior at all because ajax.getImage would immediately fail because of that mistake.

lucaswoj: Ah! I misread this code. Thanks 👍

PR #2530, opened 2016-05-04
Improve keyboard handlers with held-down keys.

### LINKED ISSUE #1661: Make keyboard navigation friendlier with held-down keys
Right now pressing down a key to continuously pan the map results in staggering, weird motion. We should fix that.

tmcw:
Fixes #1661

Simplest possible solution. Works pretty well.

Before:

before

After:

after

cc @lucaswoj  for the review


tmcw:
Ping @lucaswoj - does this seem workable?

lucaswoj:
Thanks @mourner! This looks great @tmcw! Thank you.

PR #2542, opened 2016-05-07
[not ready] Fire a `geolocate` event

tristen:
Inspired by this [StackOverflow question](http://stackoverflow.com/questions/36984884/mapbox-gl-geolocate-not-showing-an-indicator-of-where-i-am). Fire a `geolocate` event on success of `navigator.geolocation.getCurrentPosition`:

```js
map.addControl(new mapboxgl.Geolocate());
map.on('geolocate', function(e) {
  // console.log(e.coords)
});
```

### TODOs

- [x] Document this addition.
- [ ] There aren't currently UI tests for the geolocate control. I'm guessing this is complicated to do as it requires a user prompt to enable.

cc @jfirebaugh @lucaswoj

### DIFF
--- a/js/ui/control/control.js
+++ b/js/ui/control/control.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var util = require('../../util/util');
+var Evented = require('../../util/evented');
 module.exports = Control;
 
 /**
@@ -47,3 +49,5 @@ Control.prototype = {
         return this;
     }
 };
+
+util.extend(Control.prototype, Evented);
--- a/js/ui/control/geolocate.js
+++ b/js/ui/control/geolocate.js
@@ -59,10 +59,13 @@ Geolocate.prototype = util.inherit(Control, {
             bearing: 0,
             pitch: 0
         });
+
+        this.fire('geolocate', position);
         this._finish();
     },
 
-    _error: function() {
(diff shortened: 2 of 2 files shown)


jfirebaugh:
Seems like a good idea. I wonder if this should be fired on the map, or on the control itself. `Control` doesn't currently inherit from `Evented` but we could change this.

Also, there should be an event for the error case as well.

tristen:
> I wonder if this should be fired on the map, or on the control itself. `Control` doesn't currently inherit from `Evented` but we could change this.

Yeah, that's probably a good idea. I'll can add it in!

> Also, there should be an event for the error case as well.

Okie dokie. And yeah, if its fired from the control itself I can just name this as `error`!

tristen:
Addressed ^ usage now looks like this:

```js
var geolocate = new mapboxgl.Geolocate();
map.addControl(geolocate);

geolocate.on('geolocate', function(e) {
    // console.log(e);
});

geolocate.on('error', function(e) {
    // console.log(e);
});
```

jfirebaugh:
👍 :shipit:

tristen:
https://github.com/mapbox/mapbox-gl-js/commit/58ad4fa59cb7849cf334ffb61953898ce8dc2c43

PR #2544, opened 2016-05-07
fix #2541, scale circles in pitch views

### LINKED ISSUE #2541: circles do not scale when map is pitched
**mapbox-gl-js version**: all versions

### Steps to Trigger Behavior

 Create a circle layer, set pitch to anything besides zero.

### Expected Behavior

Circles are scaled based on distance: closer circles should be larger.

### Actual Behavior

Circles are unscaled:
screen shot 2016-05-06 at 5 24 55 pm

---

It was also unexpected that circles wouldn't be oriented with the map when enabling pitch, but that is already tracked in https://github.com/mapbox/mapbox-gl-js/issues/2420

ansis:
fix #2541 

Based on the comments in the old code the old the behaviour was intentional but I think the behavior expected in #2541 is better.

screen shot 2016-05-07 at 6 21 49 pm

:eyes: @lucaswoj

### DIFF
--- a/js/render/draw_circle.js
+++ b/js/render/draw_circle.js
@@ -27,7 +27,9 @@ function drawCircles(painter, source, layer, coords) {
 
         var program = painter.useProgram('circle', bucket.getProgramMacros('circle', layer));
 
-        gl.uniform2fv(program.u_extrude_scale, painter.transform.pixelsToGLUnits);
+        gl.uniform2f(program.u_extrude_scale,
+                painter.transform.pixelsToGLUnits[0] * painter.transform.altitude,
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Sounds good to me. I guess both symbols and circles should have a style property that controls this. Something like  `"{icon,text,circle}-pitch-scaling": "{perspective,none}"`, with default `"perspective"`.

lucaswoj:
LGTM! 🚢

jfirebaugh:
Followups:

* https://github.com/mapbox/mapbox-gl-test-suite/issues/87
* https://github.com/mapbox/mapbox-gl-native/issues/5006

PR #2547, opened 2016-05-09
Allow the use of special characters in property field names

### LINKED ISSUE #2271: Allow the use of special characters in property field names
... except "{" and "}"

Should require a one line change to https://github.com/mapbox/mapbox-gl-js/blob/master/js/util/token.js#L14

ref https://github.com/mapbox/mapbox-gl-style-spec/issues/419
ref https://github.com/mapbox/mapbox-gl-native/issues/4336

lucaswoj:
Fixes #2271 

cc @jfirebaugh @musicformellons

### DIFF
--- a/js/util/token.js
+++ b/js/util/token.js
@@ -11,7 +11,7 @@ module.exports = resolveTokens;
  * @private
  */
 function resolveTokens(properties, text) {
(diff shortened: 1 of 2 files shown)


jfirebaugh:
👍

Issue #2535, opened 2016-05-05
Better communicate internal failures


scothis:
https://github.com/mapbox/mapbox-gl-js/issues/2533 and https://github.com/mapbox/mapbox-gl-js/issues/2534 are issues that are easy to trigger in Studio. Both of these issues cause mapbox-gl-js to appear to hang; it stops responding to user input and style mutations.

There is no good way for Studio it know the gl-js rendering has stopped as no errors are thrown that are catchable, and no events are emitted. If gl-js was able to communicate its state, Studio could destroy the map and initialize a new map, causing a visual blip, but maintaining general working order.

Each recent release of mapbox-gl-js has fixed similar bugs and introduced new issues. It seems that these types are error happen and will continue to happen. Of course bug free code is ideal, but focusing on handling and communicating errors would go a long way.

lucaswoj:
I synced up with @scothis this morning. A thrown error in the render loop was causing the render loop to stop running without any error events or catchable thrown errors.

PR #2548, opened 2016-05-09
Emit instead of throw errors from Map#_render

lucaswoj:
If an error within `Map#_render` gets _thrown_ (rather than _emitted_) it causes rendering to stop. Furthermore, such errors are uncatchable because they are inside  a `requestAnimationFrame` callback. 

While we should aspire to never throw any errors, the reality is that our software sometimes has bugs and giving users the ability to recover from those bugs is important.  

This commit adds a try / catch block around the body of `Map#_render`, catching thrown errors and turning them into emitted errors.

cc @scothis @jfirebaugh

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -136,6 +136,7 @@ var Map = module.exports = function(options) {
     if (options.style) this.setStyle(options.style);
     if (options.attributionControl) this.addControl(new Attribution(options.attributionControl));
 
+    this.on('error', this.onError);
     this.on('style.error', this.onError);
     this.on('source.error', this.onError);
     this.on('tile.error', this.onError);
@@ -796,7 +797,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
         if (this._frameId) {
             browser.cancelFrame(this._frameId);
         }
-        this.fire("webglcontextlost", {originalEvent: event});
+        this.fire('webglcontextlost', {originalEvent: event});
     },
 
     /**
@@ -811,7 +812,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
         this._setupPainter();
         this.resize();
         this._update();
-        this.fire("webglcontextrestored", {originalEvent: event});
+        this.fire('webglcontextrestored', {originalEvent: event});
     },
 
     /**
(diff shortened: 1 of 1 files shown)


mourner:
Try/catch blocks can affect performance significantly (v8 can't optimize functions with them), and are usually strongly discouraged in hot loops. How do benchmarks look?

lucaswoj:
The performance degradation due to try/catch is only within the `Map#_render` function, not its callees. The fps benchmark does not indicate any negative impact of this change.

scothis:
I wouldn't expect this try/catch to cause a perf issue as this method is invoked once per animation frame. But if it does, you could instead introduce an inline function that wraps calling this._render() in a try/catch. https://github.com/mapbox/mapbox-gl-js/blob/catch-render/js/ui/map.js#L939

scothis:
Addresses #2535

mourner:
I generally don't like the change too much — it hides symptoms of issues, making it harder for us to fix bugs. I'd prefer to to catch errors or emit rather than throw in the origins rather than wrapping the whole rendering frame in a try/catch.

lucaswoj:
> it hides symptoms of issues

How does this hide symptoms? The error is emitted as en event and printed to the console.

mourner:
> The error is emitted as an event and printed to the console.

Is it? I only see `onError` bound on other downstream errors like style, but maybe I'm missing something.

lucaswoj:
Ooh! Good catch. Amend the commit at https://github.com/mapbox/mapbox-gl-js/pull/2548/files#diff-274e90d38623e2e16a5e7d34414f33c7R139

js/ui/map.js
```diff
@@ -855,41 +856,46 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+            if (!this.animationLoop.stopped()) {
+                this._styleDirty = true;
+            }
+
+            if (this._sourcesDirty || this._repaint || !this.animationLoop.stopped()) {
+                this._rerender();
+            }
+
+        } catch (error) {
+            this.fire('error', error);
```
scothis: Should we rethrow the error? If an unhandled error previously stopped the render loop, catching and not rethrowing will continue the render loop in an unknown state.

lucaswoj: That's a good idea and hopefully addresses @mourner's concerns

lucaswoj: Hm. On further thought, I don't think it makes so much sense to have both an error event and a thrown error. The render loop will still end because the `this._rerender` call is in the `try` block.

PR #2550, opened 2016-05-09
Rebind VAOs when program changes, fixing "line-pattern" and "fill-pattern" rendering

### LINKED ISSUE #2533: Unable to add a fill-pattern to a layer

### Steps to Trigger Behavior

Tested in Chrome 50.

 1. create a layer with a fill-color and no fill-pattern
 2. use `map.setPaintProperty()` to set a fill-pattern

### Expected Behavior

The layer should render with the pattern and continue to be responsive to user input and style mutations.

### Actual Behavior

The pattern is set for the layer. An error is logged to the console and the map is no longer interactive.

    Uncaught AssertionError: trying to bind a VAO to a different shader

screen shot 2016-05-05 at 4 56 28 pm

cc @lucaswoj

lucaswoj:
35cd69fcd77f6fd1ef97e0b1f9b10ec0759a1957 fixes #2533
6bd6bfe85a85a46d462898e5af85d57531cd9d06 adds a reusable `warnOnce` utility method to dedupe warnings 

cc @jfirebaugh @ansis @mourner

### DIFF
--- a/js/render/vertex_array_object.js
+++ b/js/render/vertex_array_object.js
@@ -27,8 +27,15 @@ VertexArrayObject.prototype.bind = function(gl, program, vertexBuffer, elementBu
         util.warnOnce('Not using VertexArrayObject extension.');
     }
 
-    if (!this.boundProgram) {
+    var isFreshBindRequired = !(
+        this.boundProgram &&
+        this.boundProgram === program &&
+        this.boundVertexBuffer === vertexBuffer &&
+        this.boundVertexBuffer2 === vertexBuffer2 &&
+        this.boundElementBuffer === elementBuffer
+    );
 
+    if (isFreshBindRequired) {
         var numPrevAttributes = ext ? 0 : (gl.currentNumAttributes || 0);
         var numNextAttributes = program.numAttributes;
         var i;
@@ -66,13 +73,6 @@ VertexArrayObject.prototype.bind = function(gl, program, vertexBuffer, elementBu
             this.boundVertexBuffer2 = vertexBuffer2;
             this.boundElementBuffer = elementBuffer;
         }
-
-    } else {
-        // verify that bind was called with the same arguments
-        assert(this.boundProgram === program, 'trying to bind a VAO to a different shader');
-        assert(this.boundVertexBuffer === vertexBuffer, 'trying to bind a VAO to a different vertex buffer');
-        assert(this.boundVertexBuffer2 === vertexBuffer2, 'trying to bind a VAO to a different vertex buffer');
-        assert(this.boundElementBuffer === elementBuffer, 'trying to bind a VAO to a different element buffer');
     }
 };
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
WDYT about kicking off https://github.com/mapbox/mapbox-gl-test-suite/issues/76 with this change?

ansis:
I would rather have the style change reset/destroy the related VAOs instead of allowing on-the-fly. The assertions that were removed are important for catching incorrect use of our VAO abstraction.

35cd69fcd77f6fd1ef97e0b1f9b10ec0759a1957 doesn't handle disabling vertex attributes. Destroying and recreating related VAOs when the style changes should solve this.

lucaswoj:
The increased coupling between `Style` and `VertexArrayObject` introduced by that solution is undesirable. Where would you put the logic to destroy the VAOs when `line-pattern` changes? Within `setPaintProperty`? Attached to a `Style#change` event?

lucaswoj:
Or are you suggesting that we delete all VAOs on any style mutation?

lucaswoj:
@jfirebaugh I would like to tackle that issue sometime soon. I estimate it'll take a couple days of dev time to do it right across Native and JS. Can we ship this bugfix and queue up that issue for the near future?

jfirebaugh:
My concern is that if it's not done now, it's likely to fall off our radar as soon as the immediate fire is out, thus increasing the likelihood that we ship more regressions that could have been caught with this kind of tests.

ansis:
> The increased coupling between Style and VertexArrayObject introduced by that solution is undesirable. 

Yeah, hmm.

> Or are you suggesting that we delete all VAOs on any style mutation?

This sounds like a good thing to try

lucaswoj:
I'm not feeling great about any designs that delete all VAOs on any style mutation.

 - deleting VAOs on any style mutation carries unnecessarily performance cost 
 - VAOs belong to `Bucket` which is recreated only during layout changes, though VAO attributes may change during a paint property change (e.g. `fill-pattern`)
 - VAOs belong to `Bucket` which only has a reference to `StyleLayer` (not `Style`) precluding it from listening to `Style#`change events
 - moving all VAOs to `Painter` may be conceptually cleaner but requires a more complex implementation

I infer that @ansis is concerned about a scenario in which we experience a silent failure, accidently `freshBind`ing VAOs on every frame, slowing down performance without any other warnings. This is certainly undesirable but less undesirable than the other solutions proposed above. 

The `Bucket` abstraction is 🙅. I'm working on a number of initatives to make it 💁.

lucaswoj:
@ansis @jfirebaugh How do you feel about merging this as-is? I would like to get this fix into the next release alongside earcut.

jfirebaugh:
Is #2533 a recent regression? If it's not a recent regression, let's just continue to live with it until we can apply a proper fix with regression tests.

lucaswoj:
This regression was introduced in the most recent release as a result of https://github.com/mapbox/mapbox-gl-js/pull/2483.

jfirebaugh:
Ok, how about we revert that instead then?

lucaswoj:
Sounds good

Issue #2473, opened 2016-04-20
Browser/Platform specific rendering issues with 0.17.0


tobinbradley:
I'm seeing some odd rendering issues with 0.17.0, but it's highly platform and browser specific. I can see the problem on the map on the API example page, but here's a [jsbin of it](http://jsbin.com/tihoyihudu/edit?html,output).

With Chrome 50 on Linux (Arch kernel 4.4) using Nvidia proprietary drivers, and with Chrome 49 on two different Chromebooks (Chrome OS, different ARM-based GPU's), the map doesn't render labels and the lines are strange:

![Chrome](http://i.imgur.com/vhImnLp.png "Chrome/Linux and Chrome OS/0.17.0")

On Firefox 45 on Linux (Arch kernel 4.4.6) with Nvidia proprietary drivers, the map looks good, but none of the labels show up:

![Firefox](http://i.imgur.com/g8tNSGR.png "Firefox/Linux")

On Windows 10, Firefox and Chrome show the map as expected, as does Chrome on Android, as does Chrome and Firefox on a *different* Arch box with intel graphics. 0.16.0 shows the map as expected everywhere. No console errors show up on any browser/platform, and I'm not seeing any network requests that are failing. If I swap 0.17.0 for 0.16.0 in the jsbin everything looks right again on every platform.

So...weird. If it were just one of my Arch boxes I'd dismiss it as a personal problem, but it's happening on Chrome OS devices as well, and it's specific to 0.17.0. I logged in as a guest on Chrome OS it to make sure a wayward browser extension wasn't raining on my parade.

I imagine this may be too difficult to simulate to fix, I just wanted to report it in case others are running into the same problem.

skokenes:
Yea I am having issues as well with 0.17.0. I cannot add a geojson layer of circles to a map. I am running on a MacBook Pro with NVIDIA GeForce GT 650M, assuming NVIDIA drivers...

In Chrome 49, works fine:
![screen shot 2016-04-20 at 2 39 51 pm](https://cloud.githubusercontent.com/assets/5554373/14686237/d67da8aa-0705-11e6-86d6-7a6cef192ee0.png)

In Firefox 45, I get no circles and some errors in the console:
<img width="936" alt="screen shot 2016-04-20 at 2 40 07 pm" src="https://cloud.githubusercontent.com/assets/5554373/14686249/e1d4fb72-0705-11e6-8b3b-c83fc61cc67a.png">

The errors point to:
[...]

PR #2553, opened 2016-05-10
fix for #2473

tobinbradley:
Index number wasn't being returned on some platforms in a reliable
order, causing the attributes variable to contained mismatched key/value
pairs and rendering errors (see screenshots in issue). Switching from index to gl.getAttribLocation() fixes the problem.


lucaswoj:
This looks great! Thank you @tobinbradley.

lucaswoj:
cc @ansis @jfirebaugh @mourner

ansis:
awesome, thanks for tracking this down!

Issue #2537, opened 2016-05-05
Add a way to create a bounding box of a size around a point


lucaswoj:
To make the  `queryRenderedFeatures` hitbox for a feature larger, it is necessary to generate a bounding box around the cursor. We should add a utility method to make this easier.

## Proposal 1

```js
queryRenderedFeatures(LngLatBounds.createAroundPoint(event.point, 10, 10));
```

## Proposal 2

```js
queryRenderedFeatures(event.point, { width: 10, height: 10 }); // define a rectangle
```

## Proposal 3

```js
queryRenderedFeatures(event.point, { radius: 10 }); // define a circle
```

## Proposal 4 (@1ec5 🚀)

```js
queryRenderedFeatures(event.point.outset(10, 10)); // define a rectangle
```

cc @lyzidiamond @jfirebaugh @mourner

lyzidiamond:
cc @ebendennis, the user who inspired this question!

peterqliu:
>queryRenderedFeatures(event.point, { width: 10 }); // like radius, but square

Would this be:

- a square of width 10 (so closer to a `radius: 5` with the old featuresAt)? or
- a square of width 20 (closer to `radius:10`)? or
- a square where distance from cursor to each corner of the square is 10 (so width of `20/Math.pow(2,0.5)`)?

1ec5:
Besides the ambiguity @peterqliu points out, I think it’s important to be able to make the dx and dy different, which option 3 wouldn’t allow for.

lucaswoj:
@peterqliu @1ec5 Thanks for the feedback! I updated proposals above to be more clear.

peterqliu:
I think using `radius` to describe any rectangle would be confusing. Proposal 2 is the least confusing to me ❤️

jfirebaugh:
I think we should not add anything to the library. The code to do this is trivial: `[[p.x - 5, p.y - 5], [p.x + 5, p.y + 5]]`.

lucaswoj:
Let's point people towards the solution above for now ☝️.

peterqliu:
>Let's point people towards the solution above for now ☝️.

This is probably the best answer, but won't decrease the support load of people asking this in the future. Is there some proactive way to communicate this, or do we just keep responding when the questions come?

jfirebaugh:
We can make an example. Like https://www.mapbox.com/mapbox-gl-js/example/using-box-queryrenderedfeatures/ except on click.

[...]

PR #2554, opened 2016-05-10
Add queryRenderedFeatures with bounding box around cursor API docs example

lucaswoj:
cc @1ec5 @jfirebaugh @lyzidiamond @danswick 

ref https://github.com/mapbox/mapbox-gl-js/issues/2537 

screen shot 2016-05-10 at 10 38 42 am

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -406,18 +406,25 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
      *
      * @example
      * // Find all features at a point
-     * var features = map.queryRenderedFeatures([20, 35], { layers: ['my-layer-name'] });
+     * var features = map.queryRenderedFeatures(
+     *   [20, 35],
+     *   { layers: ['my-layer-name'] }
+     * );
      *
      * @example
      * // Find all features within a static bounding box
-     * var features = map.queryRenderedFeatures([[10, 20], [30, 50]], { layers: ['my-layer-name'] });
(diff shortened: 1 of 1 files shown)


js/ui/map.js
```diff
@@ -405,10 +405,20 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
      * @example
+     * // Find all features within a static bounding box
      * var features = map.queryRenderedFeatures([[10, 20], [30, 50]], { layers: ['my-layer-name'] });
+     *
+     * @example
+     * // Find all features within a bounding box around a point
+     * var width = 10;
+     * var features = map.queryRenderedFeatures([
+     *   [point.x - width / 2, point.y - width / 2],
+     *   [point.x + width / 2, point.y + width / 2]
```
1ec5: There should be separate `width` and `height` variables, even if their values happen to be the same.

lucaswoj: Fixed in latest commit

mollymerp:
There might be confusion around units here – when I think height/width in web dev I'm usually thinking about pixels (or `em` etc) but in the last example in this PR width and height are degrees right? Also, the `radius` param in queryRenderedFeatures was in pixel units. 

Maybe add a note in a comment about the units?

lucaswoj:
The units are pixels. This is explained in the documentation for the method (cut off in the screenshot above).

mollymerp:
😳 Oops! Never mind then -- looks good!

PR #2560, opened 2016-05-11
Add util.warnOnce, don't rely on console being defined

### LINKED ISSUE #2558: Don't rely on console from web workers
We use the console object from within web workers:

* https://github.com/mapbox/mapbox-gl-js/blob/af237e92f5f3ded2077de7228cbcf3bced891679/js/data/load_geometry.js
* https://github.com/mapbox/mapbox-gl-js/blob/d496f1e633f39ac4d1816ffa8906e135f2bd0d2a/js/render/line_atlas.js#L52-L55

But the existence of the console object within the web worker scope is not guaranteed: https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/console

and it is not present in certain browsers:

* Maxthon 4.4 doesn't include console
* Google Chrome 28 doesn't include console at all
* Google Chrome 29 includes it as WorkerConsole, not console.

lucaswoj:
cc @tmcw 

fixes #2558

### DIFF
--- a/js/util/util.js
+++ b/js/util/util.js
@@ -465,3 +465,12 @@ exports.arraysIntersect = function(a, b) {
     }
     return false;
 };
+
+var warnOnceHistory = {};
+exports.warnOnce = function(message) {
+    if (!warnOnceHistory[message]) {
+        // console isn't defined in some WebWorkers, see #2558
+        if (typeof console !== "undefined") console.warn(message);
+        warnOnceHistory[message] = true;
+    }
+};
--- a/js/data/bucket/fill_bucket.js
+++ b/js/data/bucket/fill_bucket.js
@@ -35,10 +35,9 @@ FillBucket.prototype.addFeature = function(feature) {
 };
 
 FillBucket.prototype.addFill = function(vertices) {
-    if (vertices.length < 3) {
-        //console.warn('a fill must have at least three vertices');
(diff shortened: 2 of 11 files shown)


ansis:
Should warnings from workers be logged in all browsers by logging them in the main thread? if not, should we remove the warnings?

js/util/util.js
```diff
@@ -465,3 +465,12 @@ exports.arraysIntersect = function(a, b) {
     }
     return false;
 };
+
+var warnOnceHistory = {};
```
ansis: This will log warnings once per page instead of once per map. Based on what we log right now this seems ok to me though.

lucaswoj:
> Should warnings from workers be logged in all browsers by logging them in the main thread? if not, should we remove the warnings?

Are you asking if we should attempt to transfer warnings from `WebWorker`s to the main thread so that they can be printed in all cases? 

The intersection between people who need to see these warnings and people who are running browsers that don't support `WorkerGlobalScope.console` is pretty small. My intent was to avoid building much infrastructure to support this rare case.

ansis:
> The intersection between people who need to see these warnings and people who are running browsers that don't support WorkerGlobalScope.console is pretty small. My intent was to avoid building much infrastructure to support this rare case.

👍

PR #2561, opened 2016-05-11
Allow transitioned values to be unset

### LINKED ISSUE #2534: Unable to remove a fill-pattern from a layer

### Steps to Trigger Behavior

In Chrome 50

 1. create a layer with a fill-pattern
 2. use `map.setPaintProperty(layerId, 'fill-pattern', undefined)` to clear the pattern

### Expected Behavior

The pattern should be removed from the fill and the map should continue to be interactive and responsive to style mutations.

### Actual Behavior

The pattern is not removed, an error is sent on the console.

    Uncaught TypeError: Cannot read property 'stops' of undefined

screen shot 2016-05-05 at 5 07 53 pm

cc @lucaswoj

lucaswoj:
Fixes #2534 

cc @tmcw @jfirebaugh @ansis

### DIFF
--- a/js/style/style_declaration.js
+++ b/js/style/style_declaration.js
@@ -10,7 +10,7 @@ function StyleDeclaration(reference, value) {
     this.type = reference.type;
     this.transitionable = reference.transition;
     this.value = util.clone(value);
-    this.isFunction = !!value.stops;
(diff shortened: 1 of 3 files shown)


tmcw:
👍 lgtm

PR #2480, opened 2016-04-21
support 'tms' tile scheme in raster source

indus:
- Issue: https://github.com/mapbox/mapbox-gl-js/issues/1616

- Specs: https://www.mapbox.com/mapbox-gl-style-spec/#sources
> Tiled sources (vector and raster) must specify their details in terms of the TileJSON specification

- Reference: https://github.com/mapbox/tilejson-spec/blob/master/2.1.0/schema.json#L28

Issue #1616, opened 2015-10-13
Support TMS?


jleppert:
Adding scheme: tms to the layer source does nothing. Is it possible to get this supported? Can someone point me to the direction in the code where tile requests are done?

Thanks,

Johnathan

bhousel:
This is supported - you can specify your tile source as `type: raster` and embed TileJSON properties directly in the style, like this:

```
    "sources": {
        "example_id": {
            "type": "raster",
            "tiles": [
                "http://a.example.com/{z}/{x}/{y}.png",
                "http://b.example.com/{z}/{x}/{y}.png"
            ],
            "tileSize": 256
        }
    },
```

see also [Style Spec - sources](https://www.mapbox.com/mapbox-gl-style-spec/#sources)

tmcw:
That doesn't mean that we support the TMS scheme, though, right? TMS is flipped from the OSM/XYZ scheme.

I'm not sure we _want_ to support this in core: TMS is more or less a mistake of history and rare as a result.

bhousel:
Sorry, read this too quickly, reopening.  I didn't realize it was a subset of TileJSON that we are missing support for..

jleppert:
Yeah, I understand it's not supported although the TileJSON specs seem to imply that it would work (see the schema definition).

rm3dbeckers:
I'm having the same problem: I can set "scheme": "tms", but it would still behave as if I had left it at "xyz".

tmcw:
This is a known issue: as discussed earlier, Mapbox GL JS does not support TMS URLs yet. PRs accepted, otherwise please expect this to not work yet, because it isn't implemented yet.

lucaswoj:
Resolved by #2565

PR #2565, opened 2016-05-12
support 'tms' tile scheme

indus:
Issue: #1616

Specs: https://www.mapbox.com/mapbox-gl-style-spec/#sources

> Tiled sources (vector and raster) must specify their details in terms of the TileJSON specification

Reference: https://github.com/mapbox/tilejson-spec/blob/master/2.1.0/schema.json#L28

(a new clean attempt for #2480)

### DIFF
--- a/js/source/tile_coord.js
+++ b/js/source/tile_coord.js
@@ -51,12 +51,12 @@ TileCoord.fromID = function(id) {
 };
 
 // given a list of urls, choose a url template and return a tile URL
-TileCoord.prototype.url = function(urls, sourceMaxZoom) {
+TileCoord.prototype.url = function(urls, sourceMaxZoom, scheme) {
     return urls[(this.x + this.y) % urls.length]
         .replace('{prefix}', (this.x % 16).toString(16) + (this.y % 16).toString(16))
         .replace('{z}', Math.min(this.z, sourceMaxZoom || this.z))
         .replace('{x}', this.x)
-        .replace('{y}', this.y);
+        .replace('{y}', scheme === 'tms' ? (Math.pow(2, this.z) - this.y - 1) : this.y);
 };
 
 // Return the coordinate of the parent tile
--- a/js/source/raster_tile_source.js
+++ b/js/source/raster_tile_source.js
@@ -9,7 +9,7 @@ var normalizeURL = require('../util/mapbox').normalizeTileURL;
 module.exports = RasterTileSource;
 
 function RasterTileSource(options) {
-    util.extend(this, util.pick(options, ['url', 'tileSize']));
+    util.extend(this, util.pick(options, ['url', 'scheme', 'tileSize']));
 
     Source._loadTileJSON.call(this, options);
 }
@@ -18,6 +18,7 @@ RasterTileSource.prototype = util.inherit(Evented, {
     minzoom: 0,
     maxzoom: 22,
     roundZoom: true,
+    scheme: 'xyz',
     tileSize: 512,
     _loaded: false,
 
@@ -51,7 +52,7 @@ RasterTileSource.prototype = util.inherit(Evented, {
     getTile: Source._getTile,
 
     _loadTile: function(tile) {
-        var url = normalizeURL(tile.coord.url(this.tiles), this.url, this.tileSize);
+        var url = normalizeURL(tile.coord.url(this.tiles, null, this.scheme), this.url, this.tileSize);
 
         tile.request = ajax.getImage(url, done.bind(this));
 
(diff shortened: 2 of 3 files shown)


mourner:
What do you think about supporting `{-y}` token instead of a dedicated `scheme` option? This would probably be better because usage of TMS scheme would be encompassed in the source url rather than requiring options.

indus:
@mourner I suggested that in #2409;
But I´m not sure if it is a better or just a quick&dirty solution; I think it is favorable to bring the API in line with the specs (https://github.com/mapbox/tilejson-spec/blob/master/2.1.0/README.md) - or the other way round ;-)

mourner:
Oh, you're totally right — I wasn't aware it's in the TileJSON specs. Looks good to me then!

mourner:
@indus completely forgot to ask — can you PR a simple unit test for this change?

indus:
@lucaswoj Thanks for doing the hard work!

bartvde:
please note that the style spec only mentions scheme for raster sources, not for vector sources. Is this an omission in the style spec?

https://www.mapbox.com/mapbox-gl-js/style-spec/#sources-raster
https://www.mapbox.com/mapbox-gl-js/style-spec/#sources-vector

indus:
Did you really came across Vector Tiles with TMS scheme?
I would guess it is just that there is only a need to support the canonical XYZ scheme:
https://github.com/mapbox/vector-tile-spec/blob/master/2.1/README.md#3-projection-and-bounds

But you are right it says in the specs:
> Vector Tiles MAY be used to represent data with any projection and tile extent scheme.

(of course it says nothing about the support for it ;-)

bartvde:
I saw it in this post for instance: https://gis.stackexchange.com/questions/238301/mapbox-vector-tiles-from-geoserver-2-11-in-mapbox-gl-js

systemed:
> Did you really came across Vector Tiles with TMS scheme?

TMS is the scheme used in mbtiles, so it's not that unthinkable...

eugeneYWang:
Thanks for supporting TMS as OGC config vector tiles format, and make it configurable in Mapbox-gl-js. Do you think about elaborating it in official documentation? 

A lot of developers are relying on the documentation. For example, since this support is not elaborated, the type library of Mapbox-gl-js in angular lacks TMS support in vector tiles.

mollymerp:
@eugeneYWang I just opened a PR adding this to the docs #6847

js/source/tile_coord.js
```diff
@@ -51,12 +51,12 @@ TileCoord.fromID = function(id) {
...
 
 // given a list of urls, choose a url template and return a tile URL
-TileCoord.prototype.url = function(urls, sourceMaxZoom) {
+TileCoord.prototype.url = function(urls, sourceMaxZoom, scheme) {
     return urls[(this.x + this.y) % urls.length]
         .replace('{prefix}', (this.x % 16).toString(16) + (this.y % 16).toString(16))
         .replace('{z}', Math.min(this.z, sourceMaxZoom || this.z))
         .replace('{x}', this.x)
-        .replace('{y}', this.y);
+        .replace('{y}', scheme === 'tms' ? (Math.pow(2, this.z) - this.y - 1) : this.y);
```
initialdmg: Hi, I'm using mbtiles, but i find the tool i used to translate google xyz to tms schema does not use `Math.pow(2, this.z) - this.y - 1`, but `Math.pow(2, this.z - 1) - this.y - 1`. I konw that tms's original is (-180, -90), and xyz's original is (-180, 90), so it is easy to understand `Math.pow(2, this.z - 1) - this.y - 1`. 

Can somebody tell me how does `Math.pow(2, this.z) - this.y - 1` works and which is right method?

edpop: You may notice, that `vertor_zoom = raster_zoom - 1`.

indus: Can you provide a link to the tool you are using?
With quick websearch I only found the conversation in use:
https://gist.github.com/tmcw/4954720
http://bl.ocks.org/lennepkade/b6fe9e4862668b2d19fe26f6c2d7cbef
https://alastaira.wordpress.com/2011/07/06/converting-tms-tile-coordinates-to-googlebingosm-tile-coordinates/

initialdmg: I've red the 3rd blog and learned that i misunderstood the map tile method. So we transfer the world to a square using Web Mercator system, and we can get the total tile number which can be described as (1 << zoom - 1).
Now I guess it is the coordinate system i'm using cause the problem. In fact, i transformed a map to mbtiles which is projected to `EPSG:4326` (WGS84) . But as we know mbtiles spec using tms and mercator projection.  That might need to be supported by a projection-customized mapbox-gl.
Thanks for your help! :)  :+1:

PR #2566, opened 2016-05-13
Clarification around omitting `pointOrBox`

tristen:
- Include example
- Note that the `param` object is passed as the first argument when omitted.

cc @lucaswoj @mourner

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -400,7 +400,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     /**
      * Query rendered features within a point or rectangle.
      *
-     * @param {Point|Array<number>|Array<Point>|Array<Array<number>>} [pointOrBox] Either [x, y] pixel coordinates of a point, or [[x1, y1], [x2, y2]] pixel coordinates of opposite corners of bounding rectangle. Optional: use entire viewport if omitted.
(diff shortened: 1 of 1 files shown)


js/ui/map.js
```diff
@@ -417,6 +417,9 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
      *
      * @example
      * var features = map.queryRenderedFeatures([[10, 20], [30, 50]], { layers: ['my-layer-name'] });
+     *
+     * @example
```
jfirebaugh: To further clarify this, we could caption the examples. Here is the (wonky) JSDoc syntax for that:

```js
/**
 * ...
 * @example <caption>Retrieve all rendered features from a single layer</caption>
 * var features = map.queryRenderedFeatures({ layers: ['my-layer-name'] });
 */

tristen:
@jfirebaugh added that caption tag. This is merged into the `master` branch yeah? not the docs one (`mb-pages`)?

jfirebaugh:
[No, `mb-pages`.](https://github.com/mapbox/mapbox-gl-js/blob/master/docs/README.md#branch-usage)

jfirebaugh:
Committed as 7ca1b6792fc68293b20a333876fdf621fa21b9f5.

PR #2570, opened 2016-05-14
Fix color-switcher example typo

mdamien:
The `layer` variable is not declared but automatically created via the `id=layer`.


tristen:
Serious star-alignment O_O thanks! Added in https://github.com/mapbox/mapbox-gl-js/commit/f5ac15f8baad5e770b45a852ea586f6567b9c5cc

Issue #2434, opened 2016-04-13
#queryRenderedFeatures does not give the underlying JSON completely


hsinister:
**v0.16.0**:

Hi! This is my first issue report, so I hope I get it right.

Basically I'm trying to use the function queryRenderedFeatures to get the underlying data in JSON format to present on a popup. This works for the higher level of the JSON data (I can show the data on the popup), but fails to get the more deeply nested data - a console.log suggests that the objects have been converted into a string instead.

Here's the link to the issue as demonstrated using the example on "popup on hover": [jsbin](https://jsbin.com/leroki/edit?html,console,output)

I've edited the data such that there is only a single marker and also added an array of objects called "list" under "properties" a la feature.properties.list. I've also changed the popup HTML to display the first element of what is supposed to be a list.

### Expected Behavior
I expect that on hover the popup should display the first object, {"id":1,"name":"A"}.

### Actual Behavior
Instead, it shows "[", which confirms that it returns a string instead of the object.

tmcw:
Looks like a valid bug to me! The queryRenderedFeatures results are definitely coming out JSON-encoded and entering as objects.

lucaswoj:
@mourner @ansis Do we support arrays and objects within feature properties? If not, _should_ we support arrays and objects within feature properties?

ansis:
> Do we support arrays and objects within feature properties? 

Not right now. Why:

The vector tile spec [doesn't support array or object values](https://github.com/mapbox/vector-tile-spec/blob/2242b503e14a98c79b423fe7d398c1a1f3597e91/2.1/vector_tile.proto#L19). After geojson is tiled, it gets converted to a vector tile pbf so that it can be transferred from the worker to the main thread efficiently for querying. Converting it to a vector tile might not be the best idea, but that's why it's happening.

> If not, should we support arrays and objects within feature properties?

I think so. I'm not 100% sure, but I think so

Related questions:
- for -native, will the [geometry.hpp](https://github.com/mapbox/geometry.hpp) types support arrays and objects within feature properties?
- should queries return the full geojson instead of the tiled geojson? this could be more useful, but it would be inconsistent with vector tile results.

mcwhittemore:
> should queries return the full geojson instead of the tiled geojson?

Is this even possible? To get the full geojson all the tiles of the feature would have to be loaded. We could do this for geojson sources, but its more important that the api is consistent across source types than reporting full geojson.

[...]

PR #2573, opened 2016-05-16
Improve queryRenderedFeatures docs

lucaswoj:
ref https://github.com/mapbox/mapbox-gl-js/issues/2434#issuecomment-219238487

cc @jfirebaugh @gmaclennan @ansis


gmaclennan:
Thanks! This would have saved some debugging time and trawling through issues.
You should probably mention that the geometry is also copied from the geoJSON too. It is, right?

lucaswoj:
screen shot 2016-05-16 at 11 14 17 am

gmaclennan:
Ah, thanks @lucaswoj - reading the diff I missed that. Makes sense to me.

lucaswoj:
closed in favor of #2574

PR #2574, opened 2016-05-16
Improve query*Features docs

### LINKED ISSUE #2429: Improve query*Features documentation
The following caveats for feature querying were included in [the documentation for the `featuresCallback` for `featuresAt` / `featuresIn`](https://github.com/mapbox/mapbox-gl-js/blob/v0.15.0/js/ui/map.js#L914-L919):

```
 * Note: because features come from vector tiles or GeoJSON data that is converted to vector tiles internally, the returned features will be:
 *
 * 1. Truncated at tile boundaries.
 * 2. Duplicated across tile boundaries.
 *
 * For example, suppose there is a highway running through your rectangle in a `featuresIn` query. `featuresIn` will only give you the parts of the highway feature that lie within the map tiles covering your rectangle, even if the road actually extends into other tiles. Also, the portion of the highway within each map tile will come back as a separate feature.
```

We should restore and adapt this text for the documentation for queryRenderedFeatures and querySourceFeatures.

**EDIT: The scope of this ticket has expanded. Please see https://github.com/mapbox/mapbox-gl-js/issues/2429#issuecomment-214899878**

jfirebaugh:
Fixes #2429
Subsumes #2573

### DIFF
--- a/docs/_theme/index.js
+++ b/docs/_theme/index.js
@@ -57,10 +57,10 @@ module.exports = function (comments, options, callback) {
     },
     md: function (ast, inline) {
       if (inline && ast && ast.children.length && ast.children[0].type === 'paragraph') {
-        return formatMarkdown({
+        ast = {
           type: 'root',
-          children: ast.children[0].children
-        }, namespaces);
+          children: ast.children[0].children.concat(ast.children.slice(1))
+        };
       }
       return formatMarkdown(ast, namespaces);
     },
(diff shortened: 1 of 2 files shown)


gmaclennan:
This helps me understand much better what is happening @jfirebaugh, thanks! Given this, it would be useful to also copy the `id` property on a feature object as I mention in https://github.com/mapbox/mapbox-gl-js/issues/2434#issuecomment-219507190

jfirebaugh:
image

PR #2580, opened 2016-05-17
Add webpack documentation. Fixes #1649

### LINKED ISSUE #1649: Add webpack support
I'm able to bundle gl-js using webpack, but I'm getting a runtime error (in glify) with Chrome 46 on OSX.  @tmcw mentioned this has been reported before and that a webpack alternative to glify may be needed.  Adding an issue for reference and noting workaround.

```
Uncaught TypeError: Cannot read property 'match' of undefined
```
at https://github.com/mapbox/glify/blob/master/index.js#L13

Simple example showing my work -- https://github.com/twelch/react-mapbox-gl-js/tree/webpack-test
One workaround is to shim with the pre-built gl-js library using the webpack script-loader module.
```
require("script!mapbox-gl/dist/mapbox-gl.js");
```
instead of
```
import mapboxgl from 'mapbox-gl';
```

tmcw:
cc @twelch

cc @lucaswoj for the review

Tail tasks:

* [ ] Add webpack instructions to website
* [x] Document es6 import semantics
* [ ] comms

Adds a 'usage with webpack' section to the README.md, as well as an example webpack configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,34 @@ var map = new mapboxgl.Map({
 });
 ```
 
+## Using Mapbox GL JS with [Webpack](https://webpack.github.io/)
+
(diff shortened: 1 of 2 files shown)


tmcw:
Consensus from the thread and twitter is that this works, and it works locally in testing.

lucaswoj:
Looks great! Thank you @tmcw! 🚢 

fixes #1649

Issue #929, opened 2015-01-12
Pluggable API for third-party geospatial formats


mourner:
Currently we have awesome GeoJSON support that can handle big amounts of data, but all the work happens on the worker side. This means that to support third-party formats like Geobuf and Shapefile well, we can't just convert to GeoJSON in the app code and pass it to the API, because:

1. The main thread is blocked while the data is converted.
2. Transfering big GeoJSON to the worker is also blocking and extremely expensive because essentially you transfer stringified JSON with memory copy and then parse back.

We need to find a way to plug into the worker to do all the processing there, without the need to compile the decoding code into the bundle.

One way we could approach this is utilizing `importScripts` — we could pass a JS file URL with the decoding code (maybe wrapped and bundled specially for Mapbox GL JS) to the worker, and the worker would in turn load it and use it to process a particular third-party source:

```js
var geobuf = new GeoJSONSource({url: 'mydata.geobuf', decoder: 'mapbox-gl-geobuf.js'});
```

cc @tmcw @jfirebaugh

tmcw:
:+1: Though I'd assume that we would generalize `GeoJSONSource` to be something like `SingleFileSource` in this case.

I do have mild concerns about using importScripts for this in the context of the browserify-bundle issues that have cropped up before, and I'm interested in [workerify](https://github.com/shama/workerify)-like approaches, but... :shrug: those things can happen in prototyping

anandthakker:
I'd be into taking a crack at this for a use case I'm currently trying to deal with (involving using TopoJSON).

Is does the approach outlined here still seem viable?  I.e., generalizing `GeoJSONSource` to `SingleFileSource` (`VectorFileSource`?) that optionally accepts a url to a js decoder that preprocesses whatever comes back from `source.url`.

Is CSP relevant here?

jfirebaugh:
I'm pretty :-1: on this. Yes, CSP is relevant. Since [`importScripts`](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts) can only load same-domain resources, this would have to be an XHR + eval, AKA an API to inject arbitrary code into your webpage, which is a pretty big security footgun. Plus global eval means you can't do anything like `var decode = require('mapbox-gl-geobuf.js'); process(decode(geobuf));`. Instead we'd have to invent a naming convention that the decoders have to follow, juggle global variables, ensure there are no collisions, etc.

tmcw:
[...]

Issue #2516, opened 2016-04-29
transferrable symbol buffer quest


mollymerp:
This is step 1 to make cross-source label collision detection and placement possible.
Prior discussion on this topic at https://github.com/mapbox/mapbox-gl-js/issues/1042 
Plan is for me to tackle this with support from @yhahn. Thanks x 💯 to @ansis for helping think this through.

## Goal:
Let information needed to do symbol placement be transferrable as an array buffer between workers + main thread.

To do this, we need to:
1. Because we can only add typed numbers to a `structArrayType`, we first have to convert [`symbolInstance.glyphQuads`](https://github.com/mapbox/mapbox-gl-js/blob/88b99453cfe45238d2986d626f34aaf5e3ad6e6d/js/data/bucket/symbol_bucket.js#L550) into a [`structArrayType`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/util/struct_array.js) in a manner similar to the [`collisionBoxArray`](https://github.com/mapbox/mapbox-gl-js/blob/1df146627092b58c8db64d45530c74cd12a3fa02/js/symbol/collision_box.js)
2. using the results from 1. convert the [`symbolInstances`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/data/bucket/symbol_bucket.js#L149) array into a [`structArrayType`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/util/struct_array.js)
3. instead of calling [`placeFeatures()`](https://github.com/mapbox/mapbox-gl-js/blob/88b99453cfe45238d2986d626f34aaf5e3ad6e6d/js/data/bucket/symbol_bucket.js#L350) in `symbol_bucket.js` and `worker_tile.js`, add this new `symbolInstancesStructArray` aka `symbolBuffer` to `workerTile.js` [here](https://github.com/mapbox/mapbox-gl-js/blob/88b99453cfe45238d2986d626f34aaf5e3ad6e6d/js/source/worker_tile.js#L207) to transfer it back to the main thread
4. make modifications to other files in [`js/source/`](https://github.com/mapbox/mapbox-gl-js/tree/master/js/source) to make sure the transferred data is available where needed
5. create a process to transfer the `symbol_buffer` data to a new worker where it could be processed together with `symbol_buffers` from other sources using `placeFeatures()` at that time. 

When that is all done, we plan on starting another quest! Thoughts on that from @yhahn:
### Cross-source placement quest
[...]

PR #2582, opened 2016-05-17
Proof-of-concept: custom source types & worker-side plugin code

anandthakker:
Proof of concept proposal to address the concepts raised in #929.  Changes include:

 - Register Source types dynamically -- enables easier extensibility for source types (eg, `supercluster`); also removes a couple closure-captured `require` references.
 - Adds an API whereby a Source type can submit worker-side plugin code via script URL. Straw man API for this is to have the loaded script run a `registerPlugin` function that the worker puts in scope, which adds additional (namespaced) targets on the worker that the Source can then invoke with `dispatcher.send`.
 - Also includes a special-case to the above for tile loading: a `loadTile` method that's responsible for yielding up a `VectorTile` object, but _not_ for knowing how to do `tile.parse(...)`.  Not at all sure that this would be the right way to draw the line, but it seemed like a good balance to me.
 - Refactors GeoJSONSource to make use of ^. This is only as a proof of concept to show that the worker-side plugin code can be shipped using an Object URL (using `blobify`, a variation of `webworkify`).  In reality, it's probably wasteful to actually send the geojson source's worker-side code this way since it's a built in type.


anandthakker:
@jfirebaugh @tmcw @mourner I know that there's no consensus on this in the ticket, but I decided to take a swing at it since I will likely need something like this in a fork for an upcoming project.  Let me know what you think.

anandthakker:
Oops, I guess my local `master` was a few commits behind when i branched this, but I don't think the rebase will be too bad.

anandthakker:
Rebased

tmcw:
Wow. I need to review the code but preliminary reaction: wow, this is cool!

anandthakker:
@tmcw 👍 to save you time on code review: I left some of the code a bit rough (and no tests) in this first cut; if a version of this looks viable, I'll do a polishing pass.

mourner:
On the first glance, this looks great. I especially like that the code responsible for GeoJSON worker parsing neatly lies in a separate file. Will review more thoroughly later.

anandthakker:
Oh, I meant to raise this here too: this would open the door for custom source types, and I think it would allow for them to be added programmatically (`addSourceType` and then `addSource` and then `addLayer`), but it woudln't quite be possible to do it declaratively in the style json, since that gets validated against the spec.  Doesn't really seem like this would ever belong int he spec, but I wonder if it would be worthwhile to relax that part of style validation?

anandthakker:
^ last commit refactors basic geojson and supercluster into separate sources, and tries to make the common boilerplate easy to reuse.  This sets things up to enable custom data types pretty easily, which was the initial issue raised in #929 -- quick and dirty example: https://github.com/anandthakker/mapbox-gl-topojson 

Note the workaround that this required for the validation that happens in `Style.addSource()`

js/source/geojson_source.js
```diff
@@ -41,28 +44,13 @@ module.exports = GeoJSONSource;
  * map.removeSource('some id');  // remove
  */
 function GeoJSONSource(options) {
-    options = options || {};
+    this.options = options = options || {};
```
lucaswoj: I'd prefer to not save the `options` object. Doing so

 - doesn't make much sense semantically 
 - creates duplicate data within the class
 - opens up the possibility of the `options`  object being modified from the outside 

Instead, you can write `this.type = options.type`

js/source/geojson_source_worker.js
```diff
@@ -0,0 +1,98 @@
...
+var rewind = require('geojson-rewind');
+var GeoJSONWrapper = require('./geojson_wrapper');
+var vtpbf = require('vt-pbf');
+var geojsonvt = require('geojson-vt');
+
+/*
+ * A base prototype to use for creating GeoJSON-based source workers.  This is
+ * basically an abstract class; subclasses should implement (at least) the
+ * indexData method.
+ */
```
lucaswoj: Do you think we could implement a robust plugin system using interfaces and composition rather than inheritance? What's the simplest interface that would support all use cases?

js/source/worker.js
```diff
@@ -21,13 +15,31 @@ module.exports = function(self) {
...
     this.self = self;
     this.actor = new Actor(self, this);
-    this.loading = {};
 
+    this.loading = {};
     this.loaded = {};
-    this.geoJSONIndexes = {};
+
+    this.plugins = {};
+    self.registerPlugin = function (name, plugin) {
```
lucaswoj: Is "plugin" the right word here? To me, plugin implies some really generic module, whereas these are specifically source types. Would `registerSourceType` be a good name?

js/source/cluster_source.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+var util = require('../util/util');
+var GeojsonSource = require('./geojson_source');
+var webworkify = require('webworkify');
+
+module.exports = ClusterSource;
+module.exports.worker = URL.createObjectURL(webworkify(require('./geojson_worker_supercluster'), {bare: true}));
+
+function ClusterSource (options) {
```
lucaswoj: It feels funny that the `GeoJSONSource` is named by the type of data it processes while `ClusterSource` is named by a transformation it performs on the data it processes. Can you think of any way to make the source naming more consistent? `ClusteredGeoJSONSource`?

js/util/browser/dispatcher.js
```diff
@@ -17,9 +18,15 @@ function Dispatcher(length, parent) {
...
-            this.actors[i].send(type, data);
+    broadcast: function(type, data, cb) {
+        if (typeof cb !== 'function') {
+            for (var i = 0; i < this.actors.length; i++) {
+                this.actors[i].send(type, data);
+            }
+        } else {
+            util.asyncAll(this.actors, function (actor, done) {
+                actor.send(type, data, done);
+            }, cb);
```
lucaswoj: Can we remove the branching in this method by providing a default value for `cb` of `function() {}`?

```js
cb = cb || function() {};
```

lucaswoj:
This is looking _really_ good @anandthakker! Most of my comments concern very superficial issues that should only take a few minutes to fix. 

The biggest outstanding task is to carefully think through our interface design for registering and creating plugins, looking for opportunities to simplify and clarify. I'll think on this over the next few days. Let me know if there's anything else I can help with!

js/source/geojson_source.js
anandthakker: @lucaswoj Heh, I also just noticed that the `this.options = options` is actually duplicated a couple lines below, too.

Couple possible notes on removing this:
1. I noticed that `VectorTileSource` does keep ahold of its options and uses them again in `serialize` -- that actually does seem to me like an okay reason to hang on to it, but I could also see an argument for doing the serialization more explicitly... what's your preference?
2. Currently, down in `_updateData`, `this.options` gets used to load up the params that are sent over to the worker `{geojson|cluster}.parse`, allowing for those two (and potential other) source types to reuse the dirty-checking logic and such.  But anyway there are better ways to solve this.

js/source/worker.js
anandthakker: 👍 agreed

js/source/cluster_source.js
anandthakker: Yah, agreed, ClusteredGeoJSONSource / geojson-clustered is better.

js/util/browser/dispatcher.js
anandthakker: 👍

js/source/geojson_source.js
lucaswoj: I prefer to not add additional uses of `this.options` in the sources, but don't demand that you fix existing ones :-)

anandthakker:
Thanks for the review @lucaswoj !

> Do you think we could implement a robust plugin system using interfaces and composition rather than inheritance? What's the simplest interface that would support all use cases?

Yeah, good question -- definitely feels like this merits some careful thought...

Collecting some of the use cases that have come to mind here:
 - GeoJSONSource-based custom sources, where you want to provide some subset of:
  - The `params => GeoJSON` (e.g. download & decode topojson, or some other format)
  - The `GeoJSON => index` (e.g., supercluster)
 - VectorTileSource-based custom sources, where you want to provide:
  - the `x,y,z => raw tile data` step of the process.  Main example is to implement something like pre-fetching and caching for better (and custom-tailored) offline experience.
  - some kind of modification of the parsed vector tile, before it goes into buckets/rendering (e.g., joining vector tiles with custom data)

lucaswoj:
Would this interface be workable? Ideally we could encapsulate all source-specific implementation in this interface and get rid of the main thread classes (`GeoJSONSource`, `VectorSource`, etc). 

```js
type WorkerSourceType = (error: Error, source: WorkerSource) => void;

type WorkerSource = (x: number, y: number, z: number, callback: (error: Error, source: WorkerTile) => void) => void

function addSourceType(name, (options: {}, callback: (error: Error, source: Source) => void) => void);
```

anandthakker:
@lucaswoj couple questions on that:

1) `WorkerSourceType`: this might be a silly question... why is this a function?

2)
>  Ideally we could encapsulate all source-specific implementation in this interface and get rid of the main thread classes (GeoJSONSource, VectorSource, etc).

Wouldn't a lot of the main-thread logic in `VectorSource` and `GeoJSONSource` still need to be around?  (e.g. assigning workers to tiles, choosing between `load tile` and `reload tile`, lifecycle stuff)  Are you thinking that can basically be consolidated into a common driver?

3) Particular `WorkerSource` implementations probably need to expose other worker-side methods beyond the main `(x,y,z)` to `WorkerTile`, e.g. some way for GeoJSON's `setData` to invoke parsing/indexing -- is that basically just an implementation detail at this level?

anandthakker:
Oh, another small thing: I think `addSourceType` will need an optional `workerSourceURL` parameter in addition to the name and factory function

anandthakker:
> 3) Particular WorkerSource implementations probably need to expose other worker-side methods beyond the main (x,y,z) => WorkerTile, e.g. some way for GeoJSON's setData to invoke parsing/indexing -- is that basically just an implementation detail at this level?

To follow up / clarify this issue: while many custom source types would be fine with just providing the `(x,y,z) => WorkerTile` WorkerSource implementation, I think many would need additional methods to be available on the worker, which could be invoked by the main thread source.  Examples:
 - When `GeoJSONSource:setData` is called with new url or literal data, it needs to send it over to the WorkerSource so that the worker can re-download, -parse and -index it.
 - I've been thinking about a custom source type that caches the geometry data, but lets me update the properties (by, say, joining on an id property or something).  I would need a target method on the WorkerSource that could accept the updated properties.

anandthakker:
@lucaswoj What's happening now in this branch would basically amount to the `WorkerSource` interface being an object w/ a single `loadTile` method (instead of just a function), so that it's extensible with other methods.  Those other methods are accessible via `send('SOURCE_TYPE_NAME.METHOD', params, cb)`.  Does this feel off to you?

```js
type WorkerSource = {
  loadTile: (x: number, y: number, z: number, callback: (error: Error, source: WorkerTile) => void) => void
}

function addSourceType(name, (options: {}[, workerSourceUrl: URL], callback: (error: Error, source: Source) => void) => void);
```

lucaswoj:
> 1) WorkerSourceType: this might be a silly question... why is this a function?

Looking back, my code wasn't quite right. Here's what I intended to write:

```js
type WorkerSource = (x: number, y: number, z: number, callback: (error: Error, source: WorkerTile) => void) => void

function addSourceType(name, (options: {}, callback: (error: Error, source: WorkerSource) => void) => void);
```

lucaswoj:
> @lucaswoj What's happening now in this branch would basically amount to the `WorkerSource` interface being an object w/ a single `loadTile` method (instead of just a function), so that it's extensible with other methods.  Those other methods are accessible via `send('SOURCE_TYPE_NAME.METHOD', params, cb)`.  Does this feel off to you?

> ```js
> type WorkerSource = {
  loadTile: (x: number, y: number, z: number, callback: (error: Error, source: WorkerTile) => void) => void
}

> function addSourceType(name, (options: {}[, workerSourceUrl: URL], callback: (error: Error, source: Source) => void) => void);
> ```

This design feels good to me!

lucaswoj:
I'm also realizing that some source types (namely, the existing `image` and `video` source types) do need to be on the main thread.

anandthakker:
> I'm also realizing that some source types (namely, the existing image and video source types) do need to be on the main thread.

Yah -- and also, the main thread logic for the existing `GeoJSONSource` and `VectorTileSource` are not consistent.  Together with the need for methods like `setData` that could be easily accessed by client code, it seems to me like it would make sense to have a `Source` interface for the main thread that every source type must implement.

Here's what I think this could look like... heavy critique welcome!

---
## Main Thread `Source`

```js
type Source = {
  getTile: (coord: TileCoord) => Tile,
  onAdd: (map) => void,
  loaded: () => boolean,
  update: (transform: Transform) => void,
  reload: () => void,
  getVisibleCoordinates: () => TileCoord[],
  serialize: () => String,
  ?redoPlacement: () => void
  [... might have missed some ...]
}
// also must fire certain events at certain times
```

As is the case now, most sources would just defer to the common implementations currently in `source.js`, which in turn defer a lot to `TilePyramid`.  Maybe that's a common enough use case that it would be worth documenting a `TilePyramidHandlers` interface that specifies the various hooks needed in that approach (`load`, `unload`, `abort`, `add`, etc.), and having a `function createPyramidBackedSource(pyramidHandlers: TilePyramidHandlers) => Source` helper to reduce some boilerplate.

## Optional worker thread `WorkerSource`
Additionally, many sources would _also_ want to implement a `WorkerSource` interface for the supporting code they need on the worker threads.  As suggested above, this is presented as an object interface so that, in addition to the required method, particular implementations can expose additional worker-side targets.
```js
type WorkerSource = {
  loadTile: (x: number, y: number, z: number, callback: (error: Error, tile: WorkerTile) => void) => void
}
```

## Registering a Source Type

Actually registering a source type would involve calling `addSourceType` on the main thread:
```js
function addSourceType(name, ?workerSourceModuleURL, sourceFactoryFn: (options: {}, callback: (error: Error, source: Source) => void) => void)
```

## Implementation 
Because of the practical details of using `importScripts` to pull in worker-side code, the optional `workerSourceModuleURL` would point to a script that should call `self.registerWorkerSourceType(name, source: WorkerSource)`.

Also, I think it would be very convenient to be able to (selectively) reuse some of the existing logic on the main thread or the worker.  To make that possible, `worker['load tile']` uses the `type` parameter to choose which `WorkerSource` to defer to.  Then, make built in source types like `vector` and `geojson` pluggable by (1) exposing `type` as a public, overwrite-able property, and (2) clearly documenting the interfaces for their particular `WorkerSource`, etc.  That way, to write a custom geojson source, I can do something like:
```js
function myCustomSourceFactory(options, cb) {
  var source = extend(Source.create('geojson'), { type: 'geojson-custom' });
  cb(null, source);
}
```
And then just provide that along with my own implementation of the `GeoJSONWorkerSource` interface.

anandthakker:
Occurs to me that, presently, `source` objects get `source.style` and `source.dispatcher`members assigned to them by the Style in `addSource`; would it be better to pass the `style` and `dispatcher` dependencies as arguments to the factory function instead?

anandthakker:
@lucaswoj would love to get your :eyes on 2f73972 -- it's a first cut at refactoring the Source interface.

Inverts `TilePyramid` and `Source` dependencies.  `TilePyramid` now essentially serves as a common wrapper class for dealing with sources.  Source types are now essentially a tuple `(name, create: (id, options, dispatcher, onChange, callback: (err, source: Source) => void) => void`.  And a `Source` is an object implementing:

```js
tileSize: number
minzoom: number
maxzoom: number
// not sure about these three
roundZoom: boolean
reparseOverscaled: boolean
isTileClipped: boolean
load: (tile: Tile, cb: (err, data) => void) => void
abort: (tile: Tile) => void
unload: (tile: Tile) => void
serialize: () => void
```

Refactor is only completed for VectorTileSource, GeoJSONSource, and ClusteredGeoJSONSource so far -- would like to have another set of eyes on it before propagating to the other source types (but I think it'll be relatively straightforward).

There are also a bunch of TODO's sprinkled through, some of which raise a detailed question or two.

jfirebaugh:
> Inverts `TilePyramid` and `Source` dependencies.

👍. @ansis, @mollymerp and I were talking about doing this too -- in order to support [cross-source label placement](https://github.com/mapbox/mapbox-gl-js/issues/2516), we may need there to be _one_ tile pyramid that tracks tiles for _multiple_ sources and manages transferring placement data for all source tiles covering a nominal 512px "meta tile" to a shared worker. (Normally a "meta tile" would have exactly one tile for each 512px vector source, but if for some reason a vector source used 256px or 1024px tiles, the pyramid would need to handle 4:1 or 1:4 ratios. And of course most raster sources are still 256px.)

anandthakker:
@jfirebaugh Interesting!  Do you think it makes sense to do that after, before, or within this PR?  If "within," I could certainly give it a try, but would probably need to chat w someone to get more understanding of what's happening with "meta tiles" and placement

jfirebaugh:
Probably after -- I just wanted to give some additional context and support for the idea of inverting the dependencies.

anandthakker:
Tracking some todo's and questions:
 - [ ] The `Tile` objects are used by various source types to maintain tile-specific state; ideally, this implicit contract between `TilePyramid` and `Source` should be made explicit in the Source interface.
 - [ ] Decide on an approach to 
 - [ ] `Source.is` is only used in one place: to support passing already-created sources to `addSource` (and prevent the attempt to validate against the spec in that case).  What's the best way to do this in the factory function approach, where it's a little weird to rely on `instanceof`?

anandthakker:
@lucaswoj I'm flying pretty blind w/ the Video source; b73f98b was my guesswork-based attempt, but I'm clearly missing something.

js/geo/transform.js
```diff
@@ -127,6 +127,53 @@ Transform.prototype = {
...
 
+    /**
+     * Return a zoom level that will cover all tiles the transform
+     * @param {Object} pyramidProperties
+     * @param {number} pyramidProperties.tileSize
+     * @param {boolean} pyramidProperties.roundZoom
+     * @returns {number} zoom level
+     * @private
+     */
+    coveringZoomLevel: function(pyramidProperties) {
```
lucaswoj: The fact that these properties were once part of `TilePyramid` isn't important going forward. Let's just use a generic name like `options` 😄

js/geo/transform.js
```diff
@@ -127,6 +127,53 @@ Transform.prototype = {
...
     },
 
+    /**
+     * Return a zoom level that will cover all tiles the transform
+     * @param {Object} pyramidProperties
+     * @param {number} pyramidProperties.tileSize
+     * @param {boolean} pyramidProperties.roundZoom
+     * @returns {number} zoom level
+     * @private
+     */
```
lucaswoj: 🚀 Thanks for the docs!

js/render/draw_background.js
```diff
@@ -60,7 +58,7 @@ function drawBackground(painter, source, layer) {
     // the depth and stencil buffers get into a bad state.
     // This can be refactored into a single draw call once earcut lands and
     // we don't have so much going on in the stencil buffer.
-    var coords = pyramid.coveringTiles(transform);
+    var coords = transform.coveringTiles({ tileSize: 512 });
     for (var c = 0; c < coords.length; c++) {
         var coord = coords[c];
         var tileSize = 512;
```
lucaswoj: Seems like we're using `512` as a magic number in a few places. Could you factor this out into a constant?

anandthakker:
Closing in favor of #2648

PR #2583, opened 2016-05-17
Refactor "Interaction" class, fix #2575

### LINKED ISSUE #2575: scrollZoom: false doesn't disable ScrollZoomHandler
when instantiating a Map object, setting `scrollZoom: false` doesn't actually disable scroll zooming. 
```js
var map = new mapboxgl.Map({
  center: [-73.9921, 40.74],
  zoom: 13,
 // this doesn't work 
  scrollZoom: false,
  container: "map",
  style: "mapbox://styles/mapbox/light-v9"
});
```
you have to call `map.scrollZoom.disable()` after instantiation to get the desired behavior

example here: http://jsbin.com/logeyosame/edit?html,output
the first release where this happens is  `0.16.0`
first bad commit is [f8dac7e](https://github.com/mapbox/mapbox-gl-js/commit/f8dac7e38601f5b5904185e82020f3cc1b0c944e)

lucaswoj:
- refactors the `Interaction` class into a simple `bindHandlers` function
 - adds regression tests for `bindHandlers`
 - fixes #2575 
 - completely removes the `map.options` property (as started in #2286) 

cc @mollymerp @jfirebaugh @tmcw @bhousel @mourner


js/ui/interaction.js
```diff
@@ -1,325 +0,0 @@
...
-            point: singular,
-            lngLats: touches.map(function(t) { return this._map.unproject(t); }, this),
-            points: touches,
-            originalEvent: e
-        });
-    }
-};
-
-
-/**
```
jfirebaugh: Did you mean to remove this documentation?

lucaswoj: I moved it to `Map`! Given that the events are fired on `Map`, that feels like the right place for these docs.

lucaswoj: Hmm... I intended to move it to `Map`

js/ui/map.js
```diff
@@ -117,11 +145,15 @@ var Map = module.exports = function(options) {
         window.addEventListener('resize', this._onWindowResize, false);
     }
 
-    this.interaction = new Interaction(this);
```
jfirebaugh: This property wasn't underscore-private, but it doesn't look like we've ever publicly documented it or used it in an example, so I'm fine with removing it.

js/ui/interaction.js
lucaswoj: Fixed in latest commit

jfirebaugh:
👍

js/ui/interaction.js
```diff
@@ -1,325 +0,0 @@
...
-        el.removeEventListener('mouseup', this._onMouseUp);
-        el.removeEventListener('mousemove', this._onMouseMove);
-        el.removeEventListener('touchstart', this._onTouchStart);
-        el.removeEventListener('touchend', this._onTouchEnd);
-        el.removeEventListener('touchmove', this._onTouchMove);
-        el.removeEventListener('touchcancel', this._onTouchCancel);
-        el.removeEventListener('click', this._onClick);
-        el.removeEventListener('dblclick', this._onDblClick);
-        el.removeEventListener('contextmenu', this._onContextMenu);
-    },
```
mourner: There was a possibility to disable all interaction altogether after having them enabled, is this not available anymore?

lucaswoj: Users can still disable all the handlers by name if necessary. 

```js
map.scrollZoom.disable();
map.boxZoom.disable();
map.dragRotate.disable();
map.dragPan.disable();
map.keyboard.disable();
map.doubleClickZoom.disable();
map.touchZoomRotate.disable();
```

This method was never publicly documented or used internally, so I axed it. Easy to re-add if needed.

mourner: OK, fine by me then.

PR #2585, opened 2016-05-17
Reduce shader boilerplate, refactor "Bucket"

lucaswoj:
## Refactoring

This pull request contains some vocabulary refactoring within bucket, aiming for conceptual correctness. For example,  the `paintAttributes` array contained much more data than just attributes.

## Pragma Statements

This pull request allows us to expand property function styling support while keeping our codebase DRY and readable. It adds `#pragma` statements to the shaders which generate definition and initialization boilerplate.

### Before

```glsl
precision highp float;

uniform mat4 u_matrix;
uniform vec2 u_extrude_scale;
uniform float u_devicepixelratio;

attribute vec2 a_pos;

#ifdef ATTRIBUTE_A_COLOR
attribute lowp vec4 a_color;
#elif defined ATTRIBUTE_ZOOM_FUNCTION_A_COLOR0
uniform lowp float u_color_t;
attribute lowp vec4 a_color0;
attribute lowp vec4 a_color1;
attribute lowp vec4 a_color2;
attribute lowp vec4 a_color3;
#else
uniform lowp vec4 a_color;
#endif

#ifdef ATTRIBUTE_A_RADIUS
attribute mediump float a_radius;
#elif defined ATTRIBUTE_ZOOM_FUNCTION_A_RADIUS
uniform lowp float u_radius_t;
attribute mediump vec4 a_radius;
#else
uniform mediump float a_radius;
#endif

varying vec2 v_extrude;
varying lowp vec4 v_color;
varying lowp float v_antialiasblur;

float evaluate_zoom_function_1(const vec4 values, const float t) {
    if (t < 1.0) {
        return mix(values[0], values[1], t);
    } else if (t < 2.0) {
        return mix(values[1], values[2], t - 1.0);
    } else {
        return mix(values[2], values[3], t - 2.0);
    }
}

vec4 evaluate_zoom_function_4(const vec4 value0, const vec4 value1, const vec4 value2, const vec4 value3, const float t) {
    if (t < 1.0) {
        return mix(value0, value1, t);
    } else if (t < 2.0) {
        return mix(value1, value2, t - 1.0);
    } else {
        return mix(value2, value3, t - 2.0);
    }
}

void main(void) {

#ifdef ATTRIBUTE_A_RADIUS
    mediump float radius = a_radius / 10.0;
#elif defined ATTRIBUTE_ZOOM_FUNCTION_A_RADIUS
    mediump float radius = evaluate_zoom_function_1(a_radius, u_radius_t) / 10.0;
#else
    mediump float radius = a_radius;
#endif

    // unencode the extrusion vector that we snuck into the a_pos vector
    v_extrude = vec2(mod(a_pos, 2.0) * 2.0 - 1.0);

    vec2 extrude = v_extrude * radius * u_extrude_scale;
    // multiply a_pos by 0.5, since we had it * 2 in order to sneak
    // in extrusion data
    gl_Position = u_matrix * vec4(floor(a_pos * 0.5), 0, 1);

    gl_Position.xy += extrude;

#ifdef ATTRIBUTE_A_COLOR
    v_color = a_color / 255.0;
#elif defined ATTRIBUTE_ZOOM_FUNCTION_A_COLOR0
    v_color = evaluate_zoom_function_4(a_color0, a_color1, a_color2, a_color3, u_color_t) / 255.0;
#else
    v_color = a_color;
#endif

    // This is a minimum blur distance that serves as a faux-antialiasing for
    // the circle. since blur is a ratio of the circle's size and the intent is
    // to keep the blur at roughly 1px, the two are inversely related.
    v_antialiasblur = 1.0 / u_devicepixelratio / radius;
}
```

## After

```glsl
precision highp float;

uniform mat4 u_matrix;
uniform vec2 u_extrude_scale;
uniform float u_devicepixelratio;

attribute vec2 a_pos;

#pragma mapbox: define(color)
#pragma mapbox: define(radius)

varying vec2 v_extrude;
varying lowp float v_antialiasblur;

void main(void) {

    #pragma mapbox: initialize(color)
    #pragma mapbox: initialize(radius)

    // unencode the extrusion vector that we snuck into the a_pos vector
    v_extrude = vec2(mod(a_pos, 2.0) * 2.0 - 1.0);

    vec2 extrude = v_extrude * radius * u_extrude_scale;
    // multiply a_pos by 0.5, since we had it * 2 in order to sneak
    // in extrusion data
    gl_Position = u_matrix * vec4(floor(a_pos * 0.5), 0, 1);

    gl_Position.xy += extrude;

    // This is a minimum blur distance that serves as a faux-antialiasing for
    // the circle. since blur is a ratio of the circle's size and the intent is
    // to keep the blur at roughly 1px, the two are inversely related.
    v_antialiasblur = 1.0 / u_devicepixelratio / radius;
}
```

cc @jfirebaugh @mourner @ansis

### DIFF
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -298,7 +298,7 @@ Bucket.prototype.getProgramDefines = function(programInterface, layer) {
     return defines;
 };
 
-Bucket.prototype.addPaintAttributes = function(interfaceName, globalProperties, featureProperties, startGroup, startIndex) {
+Bucket.prototype.populatePaintArrays = function(interfaceName, globalProperties, featureProperties, startGroup, startIndex) {
     for (var l = 0; l < this.childLayers.length; l++) {
         var layer = this.childLayers[l];
         var groups = this.arrayGroups[interfaceName];
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -287,14 +287,15 @@ Bucket.prototype.recalculateStyleLayers = function() {
     }
 };
 
-Bucket.prototype.getProgramMacros = function(programInterface, layer) {
-    var macros = [];
+// TODO make static
+Bucket.prototype.getProgramDefines = function(programInterface, layer) {
+    var defines = [];
     var attributes = this.paintAttributes[programInterface][layer.id].attributes;
     for (var i = 0; i < attributes.length; i++) {
         var attribute = attributes[i];
-        macros.push('ATTRIBUTE_' + (attribute.isFunction ? 'ZOOM_FUNCTION_' : '') + attribute.name.toUpperCase());
+        defines.push('ATTRIBUTE_' + (attribute.isFunction ? 'ZOOM_FUNCTION_' : '') + attribute.name.toUpperCase());
     }
-    return macros;
+    return defines;
 };
 
 Bucket.prototype.addPaintAttributes = function(interfaceName, globalProperties, featureProperties, startGroup, startIndex) {
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -359,24 +359,29 @@ function createPaintAttributes(bucket) {
                 var attributeInputName = attribute.name;
                 assert(attribute.name.slice(0, 2) === 'a_');
                 var attributeInnerName = attribute.name.slice(2);
-                var definePragma = 'define(' + attributeInnerName + ')';
-                var initializePragma = 'initialize(' + attributeInnerName + ')';
+                var definePragma = 'define ' + attributeInnerName;
+                var initializePragma = 'initialize ' + attributeInnerName;
                 var attributeVaryingDefinition;
 
                 paintAttributes.fragmentPragmas[initializePragma] = '';
+
+                // This token is replaced by the first argument to the pragma,
+                // which must be the attribute's precision ("lowp", "mediump",
+                // or "highp")
+                var attributePrecision = '$1';
 
                 if (layer.isPaintValueFeatureConstant(attribute.paintProperty)) {
                     paintAttributes.uniforms.push(attribute);
 
                     paintAttributes.fragmentPragmas[definePragma] = paintAttributes.vertexPragmas[definePragma] = [
                         'uniform',
-                        attribute.precision,
+                        attributePrecision,
                         attributeType,
                         attributeInputName
                     ].join(' ') + ';';
 
                     paintAttributes.fragmentPragmas[initializePragma] = paintAttributes.vertexPragmas[initializePragma] = [
-                        attribute.precision,
+                        attributePrecision,
                         attributeType,
                         attributeInnerName,
                         '=',
@@ -390,7 +395,7 @@ function createPaintAttributes(bucket) {
 
                     attributeVaryingDefinition = [
                         'varying',
-                        attribute.precision,
+                        attributePrecision,
                         attributeType,
                         attributeInnerName
                     ].join(' ') + ';\n';
@@ -398,7 +403,7 @@ function createPaintAttributes(bucket) {
                     var attributeAttributeDefinition = [
                         paintAttributes.fragmentPragmas[definePragma],
                         'attribute',
-                        attribute.precision,
+                        attributePrecision,
                         attributeType,
                         attributeInputName
                     ].join(' ') + ';\n';
@@ -433,7 +438,7 @@ function createPaintAttributes(bucket) {
 
                     attributeVaryingDefinition = [
                         'varying',
-                        attribute.precision,
+                        attributePrecision,
                         attributeType,
                         attributeInnerName
                     ].join(' ') + ';\n';
@@ -463,7 +468,7 @@ function createPaintAttributes(bucket) {
 
                         paintAttributes.vertexPragmas[definePragma] += [
                             'attribute',
-                            attribute.precision,
+                            attributePrecision,
                             'vec4',
                             attributeInputName
                         ].join(' ') + ';\n';
@@ -488,7 +493,7 @@ function createPaintAttributes(bucket) {
                             }));
                             paintAttributes.vertexPragmas[definePragma] += [
                                 'attribute',
-                                attribute.precision,
+                                attributePrecision,
                                 attributeType,
                                 attributeInputName + k
                             ].join(' ') + ';\n';
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -5,6 +5,7 @@ var Buffer = require('./buffer');
 var util = require('../util/util');
 var StructArrayType = require('../util/struct_array');
 var VertexArrayObject = require('../render/vertex_array_object');
+var assert = require('assert');
 
 module.exports = Bucket;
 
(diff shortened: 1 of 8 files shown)


js/render/painter/use_program.js
```diff
@@ -61,26 +61,45 @@ var definitions = {
...
+    '    if (t < 1.0) {' +
+    '        return mix(value0, value1, t);' +
+    '    } else if (t < 2.0) {' +
+    '        return mix(value1, value2, t - 1.0);' +
+    '    } else {' +
+    '        return mix(value2, value3, t - 2.0);' +
+    '    }' +
+    '}'
+);
+
```
ansis: Can this be moved to a separate `.glsl` file?

js/data/bucket.js
```diff
@@ -337,32 +328,96 @@ function createElementBufferType(components) {
...
+            attribute.multiplier = attribute.multiplier || 1;
+
+            for (var j = 0; j < bucket.styleLayers.length; j++) {
+                var styleLayer = bucket.styleLayers[j];
+                var layer = dataTypeLayers[styleLayer.id];
+
+                var attributeType = attribute.components === 1 ? 'float' : 'vec' + attribute.components;
+                var attributeInputName = attribute.name;
+                assert(attribute.name.slice(0, 2) === 'a_');
+                var attributeInnerName = attribute.name.slice(2);
```
ansis: Can this slice be removed, either by using the `a_` version in the shaders or by removing `a_` from all the attribute names?

js/data/bucket/circle_bucket.js
```diff
@@ -44,7 +44,8 @@ CircleBucket.prototype.programInterfaces = {
                 return util.premultiply(layer.getPaintValue("circle-color", globalProperties, featureProperties));
             },
             multiplier: 255,
-            paintProperty: 'circle-color'
+            paintProperty: 'circle-color',
+            precision: 'lowp'
```
ansis: Can the precision qualifiers (lowp, mediump, highp) definitions be moved into the shaders? Maybe something like `#pragma mapbox: define(color, lowp)`? They are only relevant in the context of glsl code.

js/data/bucket.js
```diff
@@ -122,17 +123,17 @@ Bucket.prototype.populateBuffers = function() {
  * `vertexLength` vertices. If not, append a new elementGroup. Should be called
  * by `populateBuffers` and its callees.
  * @private
- * @param {string} programName the name of the program associated with the buffer that will receive the vertices
+ * @param {string} dataLayerTypeName type of buffer that will receive the vertices
```
ansis: A bunch of things are already named *layer* and to me these aren't layers conceptually. I'm not sure I'd say that the icons and text within a single symbol layer are in different layers.  Maybe something like `dataBufferTypeName` or `dataInterfaceName`? this isn't *that* important.

ansis:
Looks great

At first I was a bit skeptical about using pragmas. They add an extra layer of abstraction and make you read a bunch of extra files to understand what's actually going on in the shaders. It does make it a lot cleaner and reduces a ton of ugly repetition though. I think it's worth it!

Besides the couple of things I commented on in the diff this looks ready

js/render/painter/use_program.js
lucaswoj: Yes! Great idea.

js/data/bucket.js
lucaswoj: All good points. I feel less good about these names in hindsight. I will revert as much renaming as possible within this PR.

jfirebaugh:
Should we build on [glslify](https://github.com/stackgl/glslify) explicitly for the shader rewriting functionality?

js/data/bucket.js
lucaswoj: We need to have two identifiers for each attribute:

 - one to pass the multiplied (i.e. floating point colors are multiplied by 255 to fit in a byte) value to the shader as an attribute or uniform, called `attributeInputName` above, prefixed by `a_`
 - one to hold the unmultiplied value and pass it to the vertex shader, called `attributeInnerName` above, not prefixed by `a_`

```js
        layoutAttributes: [{
            name: 'a_pos',
            components: 2,
            type: 'Int16'
        }],
        paintAttributes: [{
            name: 'a_color',
            components: 4,
            type: 'Uint8',
            getValue: function(layer, globalProperties, featureProperties) {
                return util.premultiply(layer.getPaintValue("circle-color", globalProperties, featureProperties));
            },
            multiplier: 255,
            paintProperty: 'circle-color',
            precision: 'lowp'
        }, {
            name: 'a_radius',
            components: 1,
            type: 'Uint16',
            isLayerConstant: false,
            getValue: function(layer, globalProperties, featureProperties) {
                return [layer.getPaintValue("circle-radius", globalProperties, featureProperties)];
            },
            multiplier: 10,
            paintProperty: 'circle-radius',
            precision: 'mediump'
        }]
```

I went down the road of declaring `CircleBucket`'s `paintAttributes` without the `a_` prefix and decided against it because

 - it is inconsistent with the `layoutAttributes` declarations
 - it requires more mutation of the `attribute` objects within `createPaintAttributes` (i.e. we need to append the `a_` to the name within the attribute object because that's the identifier of the actual shader attribute)

What is your concern with this `slice` operation?

js/data/bucket/circle_bucket.js
lucaswoj: Not presently. The pragmas are written like functions for future extensibility but are [replaced in the source per a lookup table](https://github.com/mapbox/mapbox-gl-js/blob/pragma/js/render/painter/use_program.js#L150-L151). To support this, we would need to make our pragma system more advanced or begin using glslify again.

js/data/bucket.js
lucaswoj: Reverted in the latest commit. Let's do this right in a standalone PR.

lucaswoj:
As best I can tell, glslify does not provide any functionality that affects this feature. [glslify's API](https://github.com/stackgl/glslify#module-api) allows us to import and export functions from other glsl files but not inject definitions at compile time. glsl also has a [system for source transforms](https://github.com/stackgl/glslify#source-transforms) but we don't seem to gain anything by using that vs replacing the pragmas ourselves.

js/data/bucket/circle_bucket.js
lucaswoj: I'll take a stab at making the pragma system more powerful

lucaswoj:
I managed to find a somewhat elegant design for parameterized pragmas, leveraging Javascript's `String#replace` API. 

As of this commit, a pragma defined as 
```glsl
attribute $1 float a_mapbox;
```

Will replace a pragma written as
```glsl
#pragma mapbox: define mapbox highp
```

With this code
```glsl
attribute highp float a_mapbox;
```

js/render/painter/use_program.js
```diff
@@ -61,26 +61,26 @@ var definitions = {
...
     var program = gl.createProgram();
     var definition = definitions[name];
 
-    var defines = '';
-    if (macros) {
-        for (var m = 0; m < macros.length; m++) {
-            defines += '#define ' + macros[m] + '\n';
-        }
+    var definesSource = '';
+    for (var j = 0; definesSource && j < definesSource.length; j++) {
```
jfirebaugh: `definesSource && j < definesSource.length` is always false, because `definesSource` is an empty string. Is this intended?

lucaswoj: The line should read `defines && j < defines.length`. Unfortunately, this codepath is only used by the overdraw inspector, which is not tested. I'll push a fix shortly.

PR #2588, opened 2016-05-18
Fix broken interactivity

mourner:
Interactivity was fully broken by #2583 due to a typo. @lucaswoj don’t forget to check the debug page before merging PRs like this. :)


lucaswoj:
Thanks for the fix!

Issue #2539, opened 2016-05-06
Resizing issue when working with ionic.


nicoabie:
Due to the fact that mapbox listens to resize event, it may fall into a bug when working with ionic and not being on the current view.

**mapbox-gl-js version**: 0.18.0

### Steps to Trigger Behavior

 1. Use mapbox with ionic
 2. Navigate to other view
 3. Open soft keyboard
 4. Navigate to map view

### Expected Behavior
The map has the same size

### Actual Behavior
The map gets resized to default values (400*300)

Having an option to avoid resizing may mitigate the issue

kristfal:
We've solved this by calling `resize()` on the map with debounced timer of 550ms whenever the viewport size changes. Depending on your app layout, you may just want to call resize whenever your map view appears. 

The timer is required to prevent rendering bugs that may occur if you call `resize()` while the keyboard is closing or opening.

mourner:
Yeah, just call resize after the container size stabilizes.

nicoabie:
@mourner if we fix it that way the user may see the map being resized.
Is it not feasible the possibility to add an option to not register the window resize event?

@kristfal thanks for the hack but I'd rather prevent resize event registration.

mourner:
@nicoabie oh, yeah, that may be an easy thing to add — want to make a PR? Leaflet has a `trackResize` option and we could follow.

nicoabie:
Sure thing @mourner, It would be great if this could land on 0.19.0. Thanks, I'll do it.

nicoabie:
@johnbrattaker yeap, look into the documentation. You have to set `options.trackResize=false`.

johnbrattaker:
thanks
yes this solves the problem, but it's no longer possible to enable landscape mode:)

Issue #1891, opened 2016-01-05
Building master branch fails on Windows during "gl@2.1.5 install: `prebuild --download`"


fnicollet:
Hello,

I just updated to master and ran npm install (I just removed the npm run-docs as it crashes on Windows, I already opened another issue about this, but it's ok, I don't need the docs)
Here is the stack after launching npm-install:
```
npm WARN engine pngjs@0.4.0: wanted: {"node":"0.8.x"} (current: {"node":"0.12.4","npm":"2.10.1"})
npm WARN deprecated pngjs2@2.0.0: pngjs2 has now taken over the original pngjs package on npm. Please change to use pngjs dependency, version 2+.
|
> gl@2.1.5 install D:\Workspaces\Windows8\mapbox-gl-js-master\mapbox-gl-js\node_modules\gl
> prebuild --download

prebuild WARN install Prebuilt binaries for node version v0.12.4 are not available
gyp: angle\build\standalone.gypi not found (cwd: D:\Workspaces\Windows8\mapbox-gl-js-master\mapbox-gl-js\node_modules\gl) while reading includes of angle\src\angle.gyp while loading dependencies of binding.gyp while trying to load binding.gyp
prebuild ERR! configure error
prebuild ERR! stack Error: `gyp` failed with exit code: 1
prebuild ERR! stack     at ChildProcess.onCpExit (D:\Workspaces\Windows8\mapbox-gl-js-master\mapbox-gl-js\node_modules\gl\node_modules\node-gyp\lib\configure.js:305:16)
prebuild ERR! stack     at ChildProcess.emit (events.js:110:17)
prebuild ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:1074:12)
prebuild ERR! not ok
prebuild ERR! build Error: `gyp` failed with exit code: 1
prebuild ERR! build     at ChildProcess.onCpExit (D:\Workspaces\Windows8\mapbox-gl-js-master\mapbox-gl-js\node_modules\gl\node_modules\node-gyp\lib\configure.js:305:16)
prebuild ERR! build     at ChildProcess.emit (events.js:110:17)
prebuild ERR! build     at Process.ChildProcess._handle.onexit (child_process.js:1074:12)
npm ERR! Windows_NT 6.3.9600
[...]

PR #2591, opened 2016-05-18
Track resize feature

nicoabie:
Pull request for issue #2539

### DIFF
--- a/js/util/browser.js
+++ b/js/util/browser.js
@@ -9,6 +9,13 @@
  * it's used for tests.
  */
 
+var window = {
+    addEventListener: function() {},
+    removeEventListener: function() {}
+};
+
+exports.window = window;
+
 exports.frame = function(fn) {
     return setImmediate(fn);
 };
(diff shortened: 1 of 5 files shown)


lucaswoj:
I'm happy to 🚢  this once you add a unit test! Thanks @nicoabie.

js/util/browser.js
```diff
@@ -9,6 +9,13 @@
  * it's used for tests.
  */
 
+var window = {
+    addEventListener: function() {},
+    removeEventListener: function() {}
+};
+
+exports.window = window;
+
```
mourner: also needs to be added to `browser/browser.js` so that it doesn't break the non-test build

nicoabie: Gotcha @mourner, I'll add an export for window as browser/browser.js assumes there is already a global 'window' object defined.
Thanks for the hint.

nicoabie:
@lucaswoj sorry for the many commits but I'm having issues with gl dependecy as described in #1891 so have to commit blindly and check for errors using the ci.

lucaswoj:
Thanks @nicoabie! Tweaked, squashed, and merged as cb814378a30c634dce942dcfbd31f15d8e538b2d!

PR #2592, opened 2016-05-18
Add style API testing to mapbox-gl-test-suite

lucaswoj:
... and fixes a regression for #2534

ref https://github.com/mapbox/mapbox-gl-test-suite/issues/76

cc @jfirebaugh @mourner

PR #2596, opened 2016-05-19
Rebind VAOs when program changes, fixing "line-pattern" and "fill-pattern" rendering

### LINKED ISSUE #2533: Unable to add a fill-pattern to a layer

lucaswoj:
continuation of #2550 

fixes #2533 

The more I think about this problem, the more certain I am that this is the right solution. 

@ansis envisioned our `VertexArrayObject` class as a thing that maintains a persistent VAO per bucket per layer. We didn't consider that sometimes the VAO must change without a bucket being reloaded (namely, for when calling `setPaintProperty` for `"line-pattern"` or `"fill-pattern"`) and thus we cannot maintain a persistent VAO. 

This PR re-envisions the `VertexArrayObject` class as a proxy for some GL calls which uses a VAO when possible. 

Alternate solutions involve too much coupling (passing the `style` to each `tile`, deleting VAOs whenever the style changes) or nasty performance regressions (reducing the gains made by earcut) (reverting this commit) or a high risk of memory leaks (storing VAOs on `Painter`, deleting VAOs whenever a tile is unloaded)

This solution _does_ branch on system capabilities. However both forks run on any modern machine, creating a low chance of uncaught regressions due to untested code.

It is a huge bummer that #2592 can't help us test this code. Fortunately, there is only 1 line of code not being tested:

```js
gl.extVertexArrayObject.bindVertexArrayOES(this.vao);
```

cc @ansis @jfirebaugh @mourner

### DIFF
--- a/js/render/vertex_array_object.js
+++ b/js/render/vertex_array_object.js
@@ -27,8 +27,15 @@ VertexArrayObject.prototype.bind = function(gl, program, vertexBuffer, elementBu
         util.warnOnce('Not using VertexArrayObject extension.');
     }
 
-    if (!this.boundProgram) {
+    var isFreshBindRequired = !(
+        this.boundProgram &&
+        this.boundProgram === program &&
+        this.boundVertexBuffer === vertexBuffer &&
+        this.boundVertexBuffer2 === vertexBuffer2 &&
+        this.boundElementBuffer === elementBuffer
+    );
 
+    if (isFreshBindRequired) {
         var numPrevAttributes = ext ? 0 : (gl.currentNumAttributes || 0);
         var numNextAttributes = program.numAttributes;
         var i;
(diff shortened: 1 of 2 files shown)


mourner:
I like this.

lucaswoj:
I added a regression test. This test is good because it exercises the case where the GL program changes due to a paint property change. If and when `headless-gl` gets support for the VAO extension, it will fully test the regression.

ansis:
Merging this for the currently release is fine

The case I'm worried about is that you accidentally use a single vao where you should be using two different vaos. I accidentally did this a couple of times in the process of implementing it. Without the assertions it would be hard to catch.

Could we create one `VertexArrayObject` object for each possible shader? We already kind of do this with `secondVaos`. Or just create an array of vaos and just use a different one for each shader?

replace

```
group.vaos[layerName] = new VertexArrayObject();
if (group.layout.element2) group.secondVaos[layerName] = new VertexArrayObject();
```

with

```
group.vaos[layerName] = [];
// 4 is the maximum number of vaos needed by a single layer
for (var i = 0; i < 4; i++) {
    group.vaos[layerName][i] = new VertexArrayObject();
}
```

I'd prefer this

lucaswoj:
@ansis What about having `VertexArrayObject` maintain a map of program names to VAO objects? (We'd probably want to rename VertexArrayObject somehow).

ansis:
> What about having VertexArrayObject maintain a map of program names to VAO objects?

Yeah, a map would be cleaner. I think it could be better to have the map outside of VertexArrayObject instead of within it though. I like the idea of creating them all upfront instead of on-demand because I think is how it would need to be done in -native.

PR #2599, opened 2016-05-21
Use shaders from mapbox-gl-shaders

jfirebaugh:
Our shaders were getting out of sync in ways that made it more and more difficult to port changes from gl-js to gl-native and vice versa. By moving them to a shared repository, and pinning a precise revision in each project's package.json, we can ensure that shader evolution follows a single path.

This doesn't necessarily mean that gl-js and gl-native have to use the same exact version at all times. The two can be pinned to different mapbox-gl-shaders git SHAs, and we can create branches in mapbox-gl-shaders as needed for experimentation. It _does_ mean that if you want to integrate shader changes to one project, you have to also accept any other changes that came before. This is a good thing: it enforces non-divergence over the long term.

Currently, mapbox-gl-shaders contains the shader code that was in mapbox-gl-js as of today. I will be following up with PRs that bring mapbox-gl-shaders in sync with and compatible with mapbox-gl-native.

cc @mapbox/gl


js/render/painter/use_program.js
```diff
@@ -8,56 +8,56 @@ var util = require('../../util/util');
 // readFileSync calls must be written out long-form for brfs.
 var definitions = {
```
jfirebaugh: In a followup PR, I'll move this object to mapbox-gl-shaders and make it the main export.

tmpsantos:
Great idea.

mourner:
This is a great idea long-term. Currently though it will add some overhead to development because we'll still be forced to maintain two separate branches (for GL JS and GL Native) because it'll be a while before Native can support DDS-related constructs, which will appear in many GL JS shaders soon.

bleege:
> it'll be a while before Native can support DDS-related constructs

Is this due to natural GL JS / GL Native code bases being out of sync right now or is this something that's different between WebGL and OpenGL?

mourner:
> Is this due to natural GL JS / GL Native code bases being out of sync right now or is this something that's different between WebGL and OpenGL?

@bleege it's because they're out of sync. Porting DDS-related changes from JS to Native is a major effort.

jfirebaugh:
We should design the shader portion of DDS so that DDS shaders can somehow be automatically "downgraded" to non-DDS shaders that native can use while it doesn't support DDS.

jfirebaugh:
I've created a merged copy of the shaders [here](https://github.com/mapbox/mapbox-gl-shaders/tree/merged/src) which makes concrete where they've diverged, using `#ifdef`s. I've done it in a way that preserves history on both sides, so it should provide a good base for slowly reducing the differences. You can git annotate and see history from both sides, for example.

lucaswoj:
https://github.com/mapbox/mapbox-gl-js/pull/2585 adds custom `#pragma`s to our shaders which conveniently abstracts away the dds-specific code. I would appreciate if this branch was based on that work, which I hope to merge today or tomorrow.

jfirebaugh:
I can update mapbox-gl-shaders when #2585 merges.

lucaswoj:
👍 Just merged. At first I thought we might need to handle the pragmas on the GL Native side, but they should work with this `ifdef` approach seamlessly. Great design!

jfirebaugh:
🚢 ?

lucaswoj:
This is great! I'm excited about this PR. I just pushed two small commits that

 - fix `useProgram`'s `defines` argument
 - use `useProgram`'s `defines` argument to inject the `MAPBOX_GL_JS` definition (the more this code is used and tested, the better!)
 - rename `vertexUtilSource` to `utilSource` (it looks like the `mapbox-gl-shaders` module doesn't have separate vertex and fragment util files) 

🚢 on 🍏 if you're happy with these changes

PR #2604, opened 2016-05-22
[docs]fix caret-right does not toggle down in Instance Members

jingsam:


lucaswoj:
cc @tmcw for the review

tmcw:
👍 great

PR #2610, opened 2016-05-23
Fix flapping test-suite tests

lucaswoj:
cc @jfirebaugh

### DIFF
--- a/test/suite_implementation.js
+++ b/test/suite_implementation.js
@@ -30,6 +30,9 @@ module.exports = function(style, options, _callback) {
         attributionControl: false
     });
 
+    // Configure the map to never stop the render loop
+    map.repaint = true;
+
     if (options.debug) map.showTileBoundaries = true;
     if (options.collisionDebug) map.showCollisionBoxes = true;
 
--- a/test/suite_implementation.js
+++ b/test/suite_implementation.js
@@ -28,7 +28,9 @@ module.exports = function(style, options, callback) {
     var gl = map.painter.gl;
 
     map.once('load', function() {
-        applyOperations(map, options.operations, function() {
+        applyOperations(map, options.operations, function(err) {
+            if (err) return callback(err);
+
             var w = options.width * browser.devicePixelRatio;
             var h = options.height * browser.devicePixelRatio;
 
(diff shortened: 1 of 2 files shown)


lucaswoj:
This PR changes the semantics of the `wait` command in the operations array in test-suite. The `wait` operation now waits for a fully-loaded render to occur. Because of this change, we can make test-suite fully backwards compatible in the case that there are no operations.

test/suite_implementation.js
```diff
@@ -65,29 +67,26 @@ module.exports = function(style, options, callback) {
...
+
+        } else {
             map[operation[0]].apply(map, operation.slice(1));
-        } catch (e) {
-            console.error(e.stack);
-            throw e;
+            applyOperations(map, operations.slice(1), callback);
         }
-        applyOperations(map, operations.slice(1), callback);
+    } catch (e) {
```
jfirebaugh: Wrapping `callback()` with a try/catch that also calls `callback` is dangerous -- if `callback` throws, it'll get called twice. What happens if we don't do any exception handling here?

lucaswoj: Good point. I will restructure to move `callback` out of the error handler.

> What happens if we don't do any exception handling here?

If we don't do any error handling and there is a synchronous error (there were several in my testing) some global exception handler (not the default one) gets triggered. This handler is awful for debugging because it prints the error message (i.e. "undefined is not a function") without any stracktrace.

jfirebaugh:
Confirmed that this fixes test suite hangs for me. 👍

test/suite_implementation.js
jfirebaugh: It gets caught by the [exception handler in map](https://github.com/mapbox/mapbox-gl-js/blob/cb814378a30c634dce942dcfbd31f15d8e538b2d/js/ui/map.js#L908-L908), emitted as an `error` event, and logged by the [default handler](https://github.com/mapbox/mapbox-gl-js/blob/cb814378a30c634dce942dcfbd31f15d8e538b2d/js/ui/map.js#L946). I guess this is a difference in how node and a browser handle `console.log(error)` -- the browser behavior of logging the stack trace being much more useful.

What about having the test-suite harness bind an `error` handler that does more complete logging? This would be helpful in general for any time the test-suite happens to cause an error in gl-js internals.

lucaswoj:
@jfirebaugh 🚢 on 🍏?

jfirebaugh:
🚢

lucaswoj:
I fixed the new problem by setting `map.repaint = true`, configuring the render loop to never stop running. Given how hard it is to debug a problem that is only reproducible on Circle CI, I am going to merge this to unblock our contributors and seek better solutions separately.

PR #2611, opened 2016-05-23
Fix overdraw inspector

lucaswoj:
per https://github.com/mapbox/mapbox-gl-js/pull/2585#discussion_r64285284

fixes the overdraw inspector, as regressed by https://github.com/mapbox/mapbox-gl-js/pull/2585

This is a tricky thing to test because the output of the overdraw inspector may change unexpectedly as our implementation changes. @jfirebaugh do you think it would be appropriate to create a test-suite case for this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "express": "^4.13.4",
     "gl": "^2.1.5",
     "istanbul": "^0.4.2",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#a7facaea4d1fefa4b713a3c5324513c69fb52d75",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#c971675cf98eb929bb4182a6732b34273151d8d2",
     "nyc": "^6.1.1",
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Hmm, good question. If I understand it correctly, the use of the overdraw inspector is primarily to check that the renderer doesn't unnecessary overdraw, so one thing that comes to mind is we could write some tests where we know that some known small amount of overdraw (perhaps none) is expected, and render in the overdraw mode for those tests.

lucaswoj:
I added an overdraw test case for a simple symbol layer. I can't imagine any improvements to performance that would affect this trivial case so the test should be sufficiently stable.

lucaswoj:
ref https://github.com/mapbox/mapbox-gl-test-suite/pull/98

lucaswoj:
I am getting a big visual diff when testing the overdraw inspector using CI. The big diff may stem from differences in the WebGL implementations, making it difficult to fix. 

screen shot 2016-05-23 at 4 38 32 pm

PR #2612, opened 2016-05-23
Cheapo WMS, Part II

bhousel:
Like #2409 but uses whoots-js library to calculate the bounding box with more correct math.
Also added a `wms.html` debug page.

cc @tmcw @lucaswoj @mourner

### DIFF
--- a/debug/wms.html
+++ b/debug/wms.html
@@ -1,0 +1,76 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Mapbox GL JS debug page</title>
+<meta charset='utf-8'>
+<meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+<link rel='stylesheet' href='../dist/mapbox-gl.css' />
+
+<style>
+body { margin: 0; padding: 0; }
+html, body, #map { height: 100%; }
+#checkboxes {
+    position: absolute;
+    border-radius: 4px;
+    background: rgba(255,255,255,0.85);
+    top: 10px;
+    left: 10px;
+    padding: 10px;
+}
+
+#checkboxes label {
+    padding-left: 5px;
+    font: 14px sans-serif;
+}
+</style>
+</head>
+
+<body>
+<div id='map'></div>
+<div id='checkboxes'>
+  <input id='show-tile-boundaries-checkbox' name='show-tile-boundaries' type='checkbox'> <label for='show-tile-boundaries'>tile debug</label><br />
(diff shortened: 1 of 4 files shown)


lucaswoj:
The CI failure should be fixed by #2610. Can you rebase and re-push? Thanks!

lucaswoj:
I'm 👍 on merging this pending:

 - [ ] consensus on what to call the token. There was some debate about this in https://github.com/mapbox/mapbox-gl-js/pull/2409#issuecomment-207493586 cc @jfirebaugh @tmcw @mourner 
 - [ ] a quick code review of `whoots-js`

lucaswoj:
What do you think about namespacing the bbox token behind a `wms` TileJSON scheme, like our `tms` scheme? #2565

bhousel:
> What do you think about namespacing the bbox token behind a wms TileJSON scheme, like our tms scheme?

I thought about that, but worried that's misleading because we are only supporting a very specific subset of wms servers.  We'd likely see people opening issues because they try to add `wms://something` as a source and it doesn't do what they expect.

lucaswoj:
My understanding is that `{bbox-epsg-3857}` only makes sense within the context of wms.

Putting the wms-specific tokens to a wms namespace will increase the self-documenting-ness of the feature and give us maneuvering room to improve wms support in the future.

Our likelihood of getting weird wms support requests is unrelated to the existence of a wms scheme.

lucaswoj:
cc @yhahn @tmcw @jfirebaugh: As the lead TileJSON contributors, what's your feeling on this?

tmcw:
I love this PR!

* I don't think we should add this to TileJSON. WMS already has its own metadata stuff and allowing this to be configured WMS-style is an unnecessary can of worms. Also, WMS is not quite a tiling 'scheme', since it isn't tiles, we're just requiring it as tiles.
* `{bbox-epsg-3857}` seems fine to me.

I see why we might want to scope this bigger, but don't feel like we need to or should. I think what we have here is pretty great.

lucaswoj:
Ok! Let's 🚢!

lucaswoj:
... as soon as I finish the hotfix release for #2657

bhousel:
Just pushed a few commits to bump to the v2 classless library, and added a test.
I'll merge later today when CircleCI goes green..

We should think about how we want to communicate this feature. Should I document it in style-spec?

bhousel:
merged as
09c0c96 Add test for `{bbox-epsg-3857}` token replacement <Bryan Housel>
e793682 Bump to WhooTS v2 <Bryan Housel>
2ebe717 Add `wms.html` debug page <Bryan Housel>
4e9a523 Use whoots-js to calculate EPSG:3857 bounding box <Bryan Housel>

1ec5:
This is being ported to the native libraries in mapbox/mapbox-gl-native#5628.

PR #2622, opened 2016-05-25
Drop smallest inner rings for earcut performance

yhahn:
- Adds unit tests for `classifyRings()`
- Adds sorting of inner rings for polygons by area (desc) and truncating to `EARCUT_MAX_RINGS` (currently set to 100)

### Next actions

- [x] coordinate benchmarks with @jakepruitt 
- [x] more IRL testing with various datasources to determine a good `EARCUT_MAX_RINGS` value
- [x] mapbox-gl-native PR (https://github.com/mapbox/mapbox-gl-native/issues/5145)

cc @jakepruitt @jfirebaugh

### DIFF
--- a/js/util/classify_rings.js
+++ b/js/util/classify_rings.js
@@ -35,7 +35,7 @@ module.exports = function classifyRings(rings, maxRings) {
     if (maxRings > 1) {
         for (var j = 0; j < polygons.length; j++) {
             if (polygons[j].length <= maxRings) continue;
-            quickselect(polygons[j], maxRings - 1, 1, polygon.length - 1, compareAreas);
+            quickselect(polygons[j], maxRings, 1, polygon.length - 1, compareAreas);
             polygons[j] = polygon.slice(0, maxRings);
         }
     }
--- a/js/util/classify_rings.js
+++ b/js/util/classify_rings.js
@@ -4,7 +4,7 @@ module.exports = classifyRings;
 
 // classifies an array of rings into polygons with outer rings and holes
 
-function classifyRings(rings) {
+function classifyRings(rings, maxRings) {
     var len = rings.length;
 
     if (len <= 1) return [rings];
@@ -16,6 +16,8 @@ function classifyRings(rings) {
     for (var i = 0; i < len; i++) {
         var area = signedArea(rings[i]);
         if (area === 0) continue;
+
+        rings[i].area = area;
 
         if (ccw === undefined) ccw = area < 0;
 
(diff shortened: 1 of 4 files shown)


mourner:
If you're feeling like doing something fancy, you can replace `sort` here with https://github.com/mourner/quickselect, which runs in `O(n)` rather than `O(n log n)`, and would also allow you to avoid some copies like `concat`. :)

According to my unscientific tests, I think max rings will end up being something closer to 800, but let's wait to see what @jakepruitt figures out with real benchmarks.

jfirebaugh:
LGTM. Will be interested to see integration with benchmarking from @jakepruitt.

Our plan is to move `classifyRings` into `loadGeometry` for the vector-tile-js 2.0 release (https://github.com/mapbox/vector-tile-js/pull/33). I think ring-dropping could move along with it, such that `loadGeometry` takes an optional `maxRings` parameter.

lucaswoj:
What went into the decision to simplify to a max # of rings rather than filter rings by area? I imagine the latter would result in more predictable visual artifacts and improved perf, even for small polygons.

mourner:
Crossposting my comment from chat, answering @lucaswoj's question:

The source issue is that Earcut is slow on polygons with tons of holes, regardless of area. If we filter by area, we can’t guarantee that we’ll filter enough if the threshold is low, and that we won't introduce unwanted visual change if the threshold is high. E.g. I had to bump the area threshold pretty high in that test above (800 in tile coords), but this may be too high for other datasets. 

So we’ll guard against the source issue with Earcut, and leave the noise filtering as responsibility of the dataset designer.

lucaswoj:
Once we bump `EARCUT_MAX_RINGS` to 500, this is ready to 🚢!

lucaswoj:
I'm going to take a quick look at benchmarks using https://github.com/mourner/quickselect instead of `Array#sort` and then 🚢 this

mourner:
@lucaswoj in theory, it should be about 5-7 times faster for a 500-hole polygon, but if the total sort is just a few millisecond, it probably doesn't matter much. Although worth pointing out that quickselect is already in the dep tree through supercluster->kdbush, so it won't add to the bundle size.

js/util/classify_rings.js
```diff
@@ -35,10 +36,9 @@ function classifyRings(rings, maxRings) {
     if (maxRings > 1) {
         len = polygons.length;
         for (var j = 0; j < len; j++) {
-            polygon = polygons[j];
-            truncated = [polygon.shift()];
-            polygon.sort(sortByArea);
-            polygons[j] = truncated.concat(polygon.slice(0, maxRings - 1));
+            if (polygons[j].length <= maxRings) continue;
+            quickselect(polygons[j], maxRings - 1, 1, polygon.length - 1, sortByArea);
+            polygons[j] = polygon.slice(0, maxRings);
```
mourner: It seems that the 2nd arg to `quickselect` should be just `maxRings` — `k` is position independent of `left`.

lucaswoj:
🚢 on 🍏 @mourner @yhahn @jakepruitt?

js/util/classify_rings.js
lucaswoj: Got it! I was a little unsure of the relationship between `k` and `maxRings` from the `quickselect` docs.

PR #2623, opened 2016-05-25
[WIP] Add earcut benchmark for bucket.addFeature method

jakepruitt:
Adds benchmark script for checking the earcut rendering speed of any arbitrary tile and layer.

cc/ @yhahn @springmeyer @jfirebaugh @lucaswoj @mourner


lucaswoj:
@jakepruitt What's your current 💭 on this? Is it ready to 🚢? If so, it'd fit in well in the new `bench/unit` folder.

jakepruitt:
Going to close this PR and plan on revisiting if we see earcut performance on specific regions as a recurring difficulty that we need to face. 👍

PR #2624, opened 2016-05-25
make symbol placement information transferrable between workers

mollymerp:
first step in the quest to make cross-source label placement a reality! re: #2516 

next actions: 
- [x] code review @lucaswoj @ansis 
- [x] ~~tests! I have no idea how to do this, so I will spend some time poking around the existing tests and try to crib from those to make sure the correct data is getting transferred.~~ this is basically a refactor, no new functionality to test
- [x] create a process to transfer the `symbol_buffer` data to a new worker where it could be processed together with `symbol_buffers` from other sources using `placeFeatures()` at that time.

also: 
- does it make sense to pass `-1` as the index for the symbolquads that do not exist in a tile?
```js
return this.symbolInstancesBuffer.emplaceBack(
     startGlyphIndex || -1,
     endGlyphIndex || -1,
     iconQuadIndex || -1,
     symbolInstance.x,
     symbolInstance.y,
     symbolInstance.index);
```
- similarly, I was getting symbol buckets that had neither glyphs or icons... why would that be happening? or am I missing something?

### DIFF
--- a/debug/index.html
+++ b/debug/index.html
@@ -71,6 +71,17 @@ map.on('load', function() {
     });
 
     map.addLayer({
+        id: "symbolTest",
+        type: "symbol",
+        source: "composite",
+        "source-layer": "poi_label",
+        layout: {
+            "icon-image":"fast-food-15",
+            "icon-allow-overlap":true
+        }
+    });
+
+    map.addLayer({
         "id": "route",
         "type": "line",
         "source": "geojson",
--- a/debug/index.html
+++ b/debug/index.html
@@ -71,17 +71,6 @@ map.on('load', function() {
     });
 
     map.addLayer({
-        id: "symbolTest",
-        type: "symbol",
-        source: "composite",
-        "source-layer": "poi_label",
-        layout: {
-            "icon-image":"fast-food-15",
-            "icon-allow-overlap":true
-        }
-    });
-
-    map.addLayer({
         "id": "route",
         "type": "line",
         "source": "geojson",
--- a/js/data/bucket/symbol_bucket.js
+++ b/js/data/bucket/symbol_bucket.js
@@ -29,6 +29,8 @@ function SymbolBucket(options) {
     this.showCollisionBoxes = options.showCollisionBoxes;
     this.overscaling = options.overscaling;
     this.collisionBoxArray = options.collisionBoxArray;
+    this.symbolQuadsArray = options.symbolQuadsArray;
+    this.symbolInstancesArray = options.symbolInstancesArray;
 
     this.sdfIcons = options.sdfIcons;
     this.iconsNeedLinear = options.iconsNeedLinear;
@@ -327,6 +329,11 @@ SymbolBucket.prototype.addFeature = function(lines, shapedText, shapedIcon, feat
                         iconBoxScale, iconPadding, iconAlongLine));
         }
     }
+
+    for (var k = 0; k < this.symbolInstances.length; k++) {
+        this.addSymbolInstance(this.symbolInstances[k]);
+    }
+
 };
 
 SymbolBucket.prototype.anchorIsTooClose = function(text, repeatDistance, anchor) {
@@ -558,3 +565,56 @@ function SymbolInstance(anchor, line, shapedText, shapedIcon, layout, addToBuffe
                 shapedIcon, iconBoxScale, iconPadding, iconAlongLine, true);
     }
 }
+
+SymbolBucket.prototype.addSymbolInstance = function(symbolInstance) {
+    var startGlyphIndex, endGlyphIndex, iconQuadIndex;
+    var quads = symbolInstance.glyphQuads ? symbolInstance.glyphQuads : [];
+    for (var i = 0; i < quads.length; i++) {
+        if (i === 0) {
+            startGlyphIndex = this.addSymbolQuad(quads[i]);
+        } else if (i === quads.length - 1) {
+            endGlyphIndex = this.addSymbolQuad(quads[i]);
+        } else {
+            this.addSymbolQuad(quads[i]);
+        }
+    }
+
+    if (symbolInstance.iconQuads && symbolInstance.iconQuads.length === 1) {
+        iconQuadIndex = this.addSymbolQuad(symbolInstance.iconQuads[0]);
+    }
+
+    return this.symbolInstancesArray.emplaceBack(
+        startGlyphIndex || -1,
+        endGlyphIndex || -1,
+        iconQuadIndex || -1,
+        symbolInstance.x,
+        symbolInstance.y,
+        symbolInstance.index);
+
+};
+
+SymbolBucket.prototype.addSymbolQuad = function(symbolQuad) {
+    return this.symbolQuadsArray.emplaceBack(
+        // anchorPoints
+        symbolQuad.anchorPoint.x,
+        symbolQuad.anchorPoint.y,
+        // corners
+        symbolQuad.tl.x,
+        symbolQuad.tl.y,
+        symbolQuad.tr.x,
+        symbolQuad.tr.y,
+        symbolQuad.bl.x,
+        symbolQuad.bl.y,
+        symbolQuad.br.x,
+        symbolQuad.br.y,
+        // texture
+        symbolQuad.tex.h,
+        symbolQuad.tex.w,
+        symbolQuad.tex.x,
+        symbolQuad.tex.y,
+        //angle
+        symbolQuad.angle,
+        // scales
+        symbolQuad.minScale,
+        symbolQuad.maxScale);
+};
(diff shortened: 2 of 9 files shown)


js/data/bucket/symbol_bucket.js
```diff
@@ -29,6 +29,8 @@ function SymbolBucket(options) {
     this.showCollisionBoxes = options.showCollisionBoxes;
     this.overscaling = options.overscaling;
     this.collisionBoxArray = options.collisionBoxArray;
+    this.symbolQuadsBuffer = options.symbolQuadsBuffer;
+    this.symbolInstancesBuffer = options.symbolInstancesBuffer;
```
lucaswoj: Can we call these `Array`s instead of `Buffer`s? We generally use the term `Buffer` exclusively in the context of a WebGL buffer (though I'd like to phase that out too)

js/data/bucket/symbol_bucket.js
```diff
@@ -327,6 +331,12 @@ SymbolBucket.prototype.addFeature = function(lines, shapedText, shapedIcon, feat
                         iconBoxScale, iconPadding, iconAlongLine));
         }
     }
+
+    var addSymbolInstance = this.addSymbolInstance.bind(this);
+    this.symbolInstances.forEach(function(instance) {
+        addSymbolInstance(instance);
+    });
```
lucaswoj: I've found that bound functions are significantly slower than unbound ones. Because this is hot code, I recommend using plain ol `for` loop and calling `this.addSymbolInstance `

js/symbol/symbol_instances.js
```diff
@@ -0,0 +1,37 @@
...
+        { type: 'Int16', name: 'glyphQuadsStart' },
+        { type: 'Int16', name: 'glyphQuadsEnd' },
+        { type: 'Int16', name: 'iconQuadIndex' },
+
+        // each symbolInstance is centered around the anchor point
+        { type: 'Int16', name: 'anchorPointX' },
+        { type: 'Int16', name: 'anchorPointY' },
+
+        // index
+        {type: 'Int8', name: 'index'}
```
lucaswoj: It might be smart to add a note here about how we are using signed types so that we can use `-1` to represent a null value.

js/symbol/symbol_instances.js
```diff
@@ -0,0 +1,37 @@
...
+    ]
+});
+
+util.extendAll(SymbolInstancesArray.prototype.StructType.prototype, {
+    get anchorPoint() {
+        return new Point(this.anchorPointX, this.anchorPointY);
+    }
+});
```
lucaswoj: Injecting methods and accessors into a generated prototype makes me a little nervous. For example, this makes it impossible to serialize a `StructArrayType` to JSON. So long as this accessor is easy to factor out, I would prefer to factor it out.

lucaswoj:
This is looking great! The code is tight and idiomatic. I made a couple superficial comments and am feeling 👍 about the direction here.

js/symbol/symbol_instances.js
mollymerp: yes-- it wasn't a very thoughtful decision to put it in there, I just saw that the method was being added to the `CollisionBoxArray StuctArrayType` so I put it in here as well. Happy to remove if its unnecessary.

lucaswoj: Ah! I missed that. If there's a precedent, no need to remove it. 😄

mollymerp:
getting some interesting rendering test failures 💭 

screen shot 2016-05-25 at 5 36 52 pm

lucaswoj:
Try re-installing `mapbox-gl-test-suite`. I've seen that failure when using v1 tiles with earcut. 

```
rm -rf node_modules/mapbox-gl-test-suite && npm install
```

mollymerp:
Ah yep, that was it @lucaswoj!

ansis:
Looks good

> does it make sense to pass -1 as the index for the symbolquads that do not exist in a tile?

It might be nice to not need a special value for the empty case. Sync the `startGlyphIndex` and `endGlyphIndex` represent a range, there are no elements if both values are the same. For example, if the start index is 52 and the end index is 52 then there are zero things between 52 and 52.

> similarly, I was getting symbol buckets that had neither glyphs or icons... why would that be happening? or am I missing something?

It could be that the tile doesn't have any features matching the bucket's filter.

---

A good next step would be to get rid of the original `symbolInstances` array completely and use the new struct array version in `placeFeatures`

mollymerp:
Update: 
As @ansis suggested, I'm refactoring to allow the total removal of the `symbolInstances` array to remove data duplication. 

[`placeFeatures`](https://github.com/mapbox/mapbox-gl-js/pull/2624/files#diff-621f09c210ab28c3dbf7bc1f0f03c6b2R355) is now being called with the `StructArray` implementation of `symbolInstances` (`symbolInstancesArray`). Labels are being placed but collision detection is not working, so I'm continuing to dig around to figure that out.

mollymerp:
Update: 
As @ansis suggested, I'm refactoring to allow the total removal of the `symbolInstances` array to remove data duplication. 

Seems like the sort function I implemented for the `StructArray` for placement of overlapping symbols isn't working. Also the render tests make me think that tile boundary clipping isn't working as expected.

mollymerp:
I fixed the sort issue, and all but 2 of the remaining render tests pass if I set all the variables in the structArray to `Float32` so I can either update the render tests to match this loss of precision, or keep the floats.

lucaswoj:
What is the downside to setting the anchor point offset variables to `Float32`?

mollymerp:
@lucaswoj I'm not sure what the downside is, but I would guess it has to do with memory usage and/or speed of transfer from worker --> main thread. I ran all the bench marking tests and didn't get any warnings.

The last remaining piece of this puzzle is figuring out why cross-tile collision detection is not working
screen shot 2016-06-03 at 11 15 49 am

mollymerp:
Mmmkk it has something to do with `CollisionTile.grid` or maybe just `CollisionTile` in general? It seems that this branch is creating too many `CollisionTile`s and the `grid` isn't being fully populated? This is what happens when I console  `collisionTile.grid` [here](https://github.com/mapbox/mapbox-gl-js/blob/b3f1bef1249ad620d36366b9289565ce4b56d9ed/js/source/worker_tile.js#L206) for the same debug url:
_master_ 
screen shot 2016-06-03 at 3 39 45 pm

_this branch_
screen shot 2016-06-03 at 3 39 33 pm

mollymerp:
Seems to be an issue with how the `CollisionTile.grid` is being populated. The fully populated `grid`s on this branch have fewer `keys` and `bboxes` than those on `master`. When I console `collisionTile.grid` in the [`done ()`](https://github.com/mapbox/mapbox-gl-js/blob/b3f1bef1249ad620d36366b9289565ce4b56d9ed/js/source/worker_tile.js#L199) function in `WorkerTile` I get very different resultant `grids`:

_master_
screen shot 2016-06-03 at 4 56 22 pm

_this branch_
screen shot 2016-06-03 at 4 56 28 pm

😭

debug/index.html
```diff
@@ -71,6 +71,17 @@
...
+        layout: {
+            "icon-image":"fast-food-15",
+            "icon-allow-overlap":true
+        }
+    });
+
+    map.addLayer({
         "id": "route",
         "type": "line",
         "source": "geojson",
```
lucaswoj: Please revert your changes to this file

js/data/bucket/symbol_bucket.js
```diff
@@ -365,26 +366,28 @@ SymbolBucket.prototype.placeFeatures = function(collisionTile, showCollisionBoxe
     var mayOverlap = layout['text-allow-overlap'] || layout['icon-allow-overlap'] ||
         layout['text-ignore-placement'] || layout['icon-ignore-placement'];
 
-    // Sort symbols by their y position on the canvas so that they lower symbols
+    // Sort symbols by their y position on the canvas so that the lower symbols
```
lucaswoj: Did you intend this change?

js/symbol/symbol_instances.js
```diff
@@ -0,0 +1,72 @@
...
+            cos = Math.cos(angle);
+
+        var sorted = array.sort(function(a, b) {
+            var aRotated = (sin * a.anchorPointX + cos * a.anchorPointY) | 0;
+            var bRotated = (sin * b.anchorPointX + cos * b.anchorPointY) | 0;
+            return (aRotated - bRotated) || (b.index - a.index);
+        });
+
+        return sorted;
+    }
```
lucaswoj: I'm nervous about calling this method  `sort` while its behaviour is so different from `Array#sort` (i.e. this doesn't sort in place and also does a slice operation).

I recommend refactoring this method into `StructArray#toArray(startIndex, endIndex)` and moving the sorting logic to the caller.

PR #2628, opened 2016-05-26
Fix uncaught error around getStyle before a style is loaded. Fixes #2619

### LINKED ISSUE #2619: Calling map.getStyle before stylesheet is loaded throws an unexpected error
https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/map.js#L557

calls

https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style.js#L585

and this.stylesheet is undefined, so an error is thrown

tmcw:
cc @lucaswoj for the review


lucaswoj:
Looks good! Thank you! :shipit:

PR #2630, opened 2016-05-27
Require node v4 in package.json engines

lucaswoj:
`npm install` recently began failing on Node v0.10 with this error message:

```
uglify-js failed on node_modules/mapbox-gl-shaders/index.js : Unterminated string constant (line: 60, col: 22, pos: 33204)
```

Is it time to start requiring Node v4? Or should we try to fix Node v0.10?

cc @jfirebaugh @tmcw @mourner @bhousel @mollymerp

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ Install the Xcode Command Line Tools Package
 xcode-select --install
 ```
 
-Install [node.js](https://nodejs.org/)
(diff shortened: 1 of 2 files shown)


jfirebaugh:
`"engines"` doesn't really do anything. I think we should just state in the README that node >= 4 is needed for browserify/webpack.

lucaswoj:
That's a good idea! Added to `CONTRIBUTING.md` alongside our other environment setup instructions

PR #2641, opened 2016-05-31
Update to Jekyll 3

tristen:
Capturing small changes here to update for [Jekyll 3](https://github.com/mapbox/publisher/issues/159)


jfirebaugh:
This is backwards compatible with Jekyll 2 right? So we can merge it immediately?

tristen:
I don't think so? I'm removing specifics about `rdiscount` and `pygments` as the highlighter in this commit.

mourner:
@jfirebaugh not backwards compatible mostly because layout variables can only be accessed through `layout.var` and not `page.var` now. It requires careful update of layout code in many places.

jfirebaugh:
@mourner There aren't any changes to layout code in this PR.

tristen:
@jfirebaugh there are in `_layout/default.html` which this pr will require from www.mapbox.com

jingsam:
Today in the morning my rss subscription was filled of posts of mapbox blog. It seems that you have switched to Jekyll 3 and rebuilt all of the posts. Cheers!

PR #2648, opened 2016-06-01
Refactor Source interface to allow/prepare for extensibility

anandthakker:
Replaces #2582

 - Inverts `TilePyramid` and `Source` dependencies.  `TilePyramid` now essentially serves as a common wrapper class for dealing with sources.

- Source types are now essentially:
```js
(
  name: string,
  createSource: (id, options, dispatcher, onChange, callback: (err, source: Source) => void) => void
  ?workerSourceURL: string
)
```

- And a `Source` is an object implementing:

```js
id: string // <- it's a little weird to rely on sources presenting this as a property
tileSize: number
minzoom: number
maxzoom: number
// not sure about these three properties
roundZoom: boolean
reparseOverscaled: boolean
isTileClipped: boolean

load: (tile: Tile, cb: (err, ?data) => void) => void
abort: (tile: Tile) => void
unload: (tile: Tile) => void
serialize: () => Object // plain JS
```

Various todo's and questions:
 - [ ] The `Tile` objects are used by various source types to maintain tile-specific state; ideally, this implicit contract between `TilePyramid` and `Source` should be made explicit in the Source interface.
 - [ ] Decide on an approach to `serialize()` (see note [here](https://github.com/anandthakker/mapbox-gl-js/blob/c3016aee9d46877ce9a2d386f3469c5dfcf3613c/js/source/tile_pyramid.js#L30-L38))
 - [ ] `Source.is` is only used in one place: to support passing already-created sources to `addSource` (and prevent the attempt to validate against the spec in that case).  What's the best way to do this in the factory function approach, where it's a little weird to rely on `instanceof`?
 - [ ] `Map` methods (`addSource`, `getSource`).  Related to this and the previous item: is it important to retain the ability to create sources _without_ adding them to the map?  If so, it might make the most sense to expose a `createSource` on the map obj, since creation now depends on a reference to `dispatcher`, which probably _shouldn't_ be exposed.
 - [ ] It's weird to rely on sources presenting an `id` property with the correct value.
 - [x] Reinstate query*Features
 - [ ] I'm not sure what the `data` yielded by `load` should look like.  Currently, the only thing I can see being needed by code outside of the particular source is `bucketStats`, which the pyramid uses to emit the `tile.stats` event.

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -665,9 +665,8 @@ Style.prototype = util.inherit(Evented, {
         var sourceResults = [];
         for (var id in this.sources) {
             var source = this.sources[id];
-            if (source.queryRenderedFeatures) {
-                sourceResults.push(source.queryRenderedFeatures(queryGeometry, params, zoom, bearing));
-            }
+            var results = Source._queryRenderedVectorFeatures(source, this._layers, queryGeometry, params, zoom, bearing);
+            sourceResults.push(results);
         }
         return this._flattenRenderedFeatures(sourceResults);
     },
@@ -677,7 +676,7 @@ Style.prototype = util.inherit(Evented, {
             this._handleErrors(validateStyle.filter, 'querySourceFeatures.filter', params.filter, true);
         }
         var source = this.getSource(sourceID);
-        return source && source.querySourceFeatures ? source.querySourceFeatures(params) : [];
+        return source ? Source._querySourceFeatures(source, params) : [];
     },
 
     addSourceType: function (name, sourceFactoryFn, workerSourceURL, callback) {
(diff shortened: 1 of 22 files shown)


js/style/style.js
```diff
@@ -647,7 +676,27 @@ Style.prototype = util.inherit(Evented, {
             this._handleErrors(validateStyle.filter, 'querySourceFeatures.filter', params.filter, true);
         }
         var source = this.getSource(sourceID);
-        return source && source.querySourceFeatures ? source.querySourceFeatures(params) : [];
+        return source ? Source._querySourceFeatures(source, params) : [];
+    },
+
+    addSourceType: function (name, sourceFactoryFn, workerSourceURL, callback) {
```
anandthakker: @lucaswoj this could also be `(name, { create: (id, options, dispatcher, onChange, callback) => void, ?workerSourceURL }, callback)`.  Any preference between the two signatures?

anandthakker:
➡️ #2667

PR #2654, opened 2016-06-02
Fix Webpack Example Config

awmichel:
It looks like the `fs.readFileSync` method calls have moved out of `mapbox-gl-js` and into `mapbox-gl-shaders` as of f36efdf06349f56724b3fa60203a378d29ffe4cd. In order for Webpack to build, this example needs to apply the brfs transform to `mapbox-gl-shaders/index.js` instead of `mapbox-gl/js/render/painter/use_program.js`.

### DIFF
--- a/webpack.config.example.js
+++ b/webpack.config.example.js
@@ -51,11 +51,11 @@ module.exports = {
             loader: 'json-loader'
         }, {
             test: /\.js$/,
-            include: path.resolve(__dirname, 'js/render/painter/use_program.js'),
+            include: path.resolve('node_modules/mapbox-gl-shaders/index.js'),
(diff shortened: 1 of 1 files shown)


tmcw:
Thanks Adam! Looks good to me, I'll merge if CircleCI agrees.

awmichel:
@tmcw Tests seem to be passing locally and on CircleCI, but something is failing with Coveralls. Is that related to this PR somehow?

lucaswoj:
@awmichel That's "normal" for commits from 3rd party repos. I will fix GL JS upstream.

lucaswoj:
fixes #2670

Issue #899, opened 2014-12-19
Create a global worker pool


jfirebaugh:
mapbox-gl-js should use `navigator.hardwareConcurrency - 1` workers no matter how many maps are on a page or how many times you call `map.setStyle()`.

anandthakker:
@jfirebaugh I'd like to be able to control the number of workers that mapbox-gl uses -- I'm envisioning using it in an app that also has some other heavy lifting to do, so that I might want to use a couple workers of my own.

As you point out, this is less of an issue if browsers implement web workers are green threads, but it could still be useful to be able to tune it.

anandthakker:
Another, perhaps much more important, reason that could matter: if there are multiple maps on the same page that use the same sources, there could be a lot of redundant parsing/processing of the same vector tile data.  A worker pool shared among `map` instances could really help to reduce that, but it would require some pretty unwieldy logic for both distinguishing and correlating source ids from different instances.

anandthakker:
@jfirebaugh @lucaswoj would you be open to a PR to address this?

mourner:
@anandthakker we're always open to PRs :)

anandthakker:
> we're always open to PRs :)

@mourner Hah! I guess what I really mean to ask is: this would definitely add a bit of complexity, and perhaps it has other drawbacks that I haven't thought of.  Given that, do you all feel that the benefits for this change would be worth the added complexity?

mourner:
@anandthakker depends on the amount of added complexity, which is hard to judge before I see a PR. :) But you could outline the approach you'd like to take first before writing any code.

anandthakker:
Here is what I am thinking:
 - In `Dispatcher`, replace `this.actors` and `this.currentActor` with a module-scoped `ActorPool` instance.
 - `ActorPool` looks like:
```js
{
   forEach: (fn: (Actor) => void) => void,
   getActor: (id) => Actor // with 'id' now a property of Actor instances, managed by the pool
   getCurrentActor: () => Actor
}
```
 - In the worker, use `params.url` instead of `params.source` as the key for `this.loading`, `this.loaded`, so that if two maps use the same source with different source id's, there's still a chance for parsed/cached tiles to be reused
 - Similar approach for `geoJSONIndexes`, but only for ones where the data is passed as a url (as opposed to geojson literal)

anandthakker:
Questions:
[...]

PR #2655, opened 2016-06-02
Add maxWorkers option

### LINKED ISSUE #2022: Add an option to set the number of spawned WebWorkers
I am working on the app where we need to be able to control / reduce the number of web workers that mapbox-gl uses. Looks like now it is always _navigator.hardwareConcurrency - 1_ ?  Is there a reason why the number of workers should not be exposed / overwritten via API?

anandthakker:
Also sets maxWorkers on the debug page to 3, on the theory that a pretty
large majority of users have 4 cpus.

Closes #2022

### DIFF
--- a/debug/index.html
+++ b/debug/index.html
@@ -58,7 +58,8 @@ var map = window.map = new mapboxgl.Map({
     zoom: 12.5,
     center: [-77.01866, 38.888],
     style: 'mapbox://styles/mapbox/streets-v8',
-    hash: true
+    hash: true,
+    workerCount: 3
 });
 
 map.addControl(new mapboxgl.Navigation());
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -19,9 +19,9 @@ var StyleFunction = require('./style_function');
 
 module.exports = Style;
 
-function Style(stylesheet, animationLoop) {
+function Style(stylesheet, animationLoop, workerCount) {
     this.animationLoop = animationLoop || new AnimationLoop();
-    this.dispatcher = new Dispatcher(Math.max(browser.hardwareConcurrency - 1, 1), this);
+    this.dispatcher = new Dispatcher(workerCount || 1, this);
     this.spriteAtlas = new SpriteAtlas(512, 512);
     this.lineAtlas = new LineAtlas(256, 512);
 
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -52,7 +52,8 @@ var defaultOptions = {
     failIfMajorPerformanceCaveat: false,
     preserveDrawingBuffer: false,
 
-    trackResize: true
+    trackResize: true,
+    workerCount: Math.max(browser.hardwareConcurrency - 1, 1)
 };
 
 /**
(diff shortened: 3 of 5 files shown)


anandthakker:
cc @lucaswoj 

I don't see an especially good way to add meaningful tests for this, since it's basically about passing a number through to `new Style()` and then to `new Dispatcher()`.  I've used something like `proxyquire` in other projects to mock CJS module dependencies, but not sure that that's something yall want to adpot

lucaswoj:
> I don't see an especially good way to add meaningful tests for this

Looks like you made [one test](https://github.com/mapbox/mapbox-gl-js/blob/master/test/js/style/style.test.js#L1125) start failing. That might be a good place to start looking. 😉

debug/index.html
```diff
@@ -58,7 +58,8 @@
     zoom: 12.5,
     center: [-77.01866, 38.888],
     style: 'mapbox://styles/mapbox/streets-v8',
-    hash: true
+    hash: true,
+    maxWorkers: 3
```
lucaswoj: Why **`max`**`Workers`? Isn't this literally the number of workers?

lucaswoj:
Happy to 🚢  this with a test case, and agreenment on [the parameter name](https://github.com/mapbox/mapbox-gl-js/pull/2655#discussion_r65629384).

debug/index.html
mourner: 👍 `numWorkers`

anandthakker:
> Looks like you made one test start failing. That might be a good place to start looking. 😉

Oh dear. 😊 

Thanks, will do.

debug/index.html
anandthakker: @lucaswoj @mourner I considered "num", but chose "max" -- this was the argument in my head:

`maxWorkers` is really what the user (me) wants to control, whereas `numWorkers` is more of an internal property.  The fact that they are the same right now is an implementation detail; I could imagine the implementation changing to, say, construct/destroy workers on demand (especially given the [possibility of worker pooling](https://github.com/mapbox/mapbox-gl-js/issues/899)).  

That said, I don't feel especially strongly about this, so just lemme know what yall think

anandthakker:
@lucaswoj Fixed the failing test and added another, but the build still fails because coveralls (I think) doesn't like my fork:
```
/home/ubuntu/mapbox-gl-js/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```

lucaswoj:
The coveralls error should have been (might have been?) fixed by https://github.com/mapbox/mapbox-gl-js/commit/1b25043da7b3bd1c2e431d2db7ff8ff25e9241ca. Have you rebased in the last day?

anandthakker:
> The coveralls error should have been (might have been?) fixed by 1b25043. Have you rebased in the last day?

I had not!  Just rebased and force pushed.

lucaswoj:
Rad! Looks like it is working now.

debug/index.html
lucaswoj: Right now, this property literally controls the number of spawned workers. If we change the implementation in the future, we can change the property name 😄

Let's go with `workerCount` for an abbreviation-free, grammatically correct, and [best-practice compliant name](http://stackoverflow.com/questions/3742650/naming-conventions-for-number-of-foos-variables).

anandthakker:
Sweet!

Did you see my [note](https://github.com/mapbox/mapbox-gl-js/pull/2655#discussion_r65649405) about the option name?  Easy to change, just lemme know.

anandthakker:
Amended the commit:
 - `s/maxWorkers/workerCount/`
 - Add jsdoc for `workerCount` option

I believe this should now be ready

anandthakker:
➡️ #2666

PR #2665, opened 2016-06-03
Change default hardwareConcurrency from 8 to 4

anandthakker:
Currently, the default value for `browser.hardwareConcurrency` is 8.  I'd suggest that this be changed to 4, because:
 - `navigator.hardwareConcurrency` is [not supported in IE and has poor mobile support](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorConcurrentHardware/hardwareConcurrency)
 - In a certain (rather large) sample of website visitors, among unique visitors whose browsers _did_ support `hardwareConcurrency`, 4 was _much_ more common than 8 or 2.

### DIFF
--- a/js/util/browser/browser.js
+++ b/js/util/browser/browser.js
@@ -76,7 +76,7 @@ exports.timed = function (fn, dur, ctx) {
  */
 exports.supported = require('mapbox-gl-js-supported');
 
-exports.hardwareConcurrency = navigator.hardwareConcurrency || 8;
+exports.hardwareConcurrency = navigator.hardwareConcurrency || 4;
 
 Object.defineProperty(exports, 'devicePixelRatio', {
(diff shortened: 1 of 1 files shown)


lucaswoj:
This is interesting. I didn't know we had a default value for `navigator.hardwareConcurrency`. I agree that `4` is likely a better default value. 

Let's give @ansis @mourner @jfirebaugh a days to weigh in and then 🚢.

jfirebaugh:
SGTM.

mourner:
I'm fine with 4. I think the original thinking behind 8 was that there's no big overhead for having 2 workers per core, but utilizing only half of available cores on the other hand would affect performance. So 8 seemed like a better trade-off. But if majority are 4, it's fine either way.

PR #2666, opened 2016-06-03
Add workerCount option

### LINKED ISSUE #2022: Add an option to set the number of spawned WebWorkers

anandthakker:
Replaces #2655

Also sets workerCount on the debug page to 3, on the theory that a pretty large majority of users have 4 cpus.

Closes #2022

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -52,7 +52,8 @@ var defaultOptions = {
     failIfMajorPerformanceCaveat: false,
     preserveDrawingBuffer: false,
 
-    trackResize: true
+    trackResize: true,
+    workerCount: Math.max(browser.hardwareConcurrency - 1, 1)
 };
 
 /**
@@ -97,6 +98,7 @@ var defaultOptions = {
  * @param {number} [options.zoom] The zoom level of the map's initial viewport.
  * @param {number} [options.bearing] The bearing (rotation) of the map's initial viewport measured in degrees counter-clockwise from north.
  * @param {number} [options.pitch] The pitch of the map's initial viewport measured in degrees.
+ * @param {number} [options.workerCount=navigator.hardwareConcurrency] The number of WebWorkers the map should use to process vector tile data.
  * @example
  * var map = new mapboxgl.Map({
  *   container: 'map',
@@ -109,10 +111,16 @@ var defaultOptions = {
 var Map = module.exports = function(options) {
 
     options = util.extend({}, defaultOptions, options);
+
+    if (options.workerCount < 1) {
+        throw new Error('workerCount must an integer greater than or equal to 1.');
+    }
+
     this._interactive = options.interactive;
     this._failIfMajorPerformanceCaveat = options.failIfMajorPerformanceCaveat;
     this._preserveDrawingBuffer = options.preserveDrawingBuffer;
     this._trackResize = options.trackResize;
+    this._workerCount = options.workerCount;
 
     if (typeof options.container === 'string') {
         this._container = document.getElementById(options.container);
@@ -586,7 +594,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
         } else if (style instanceof Style) {
             this.style = style;
         } else {
-            this.style = new Style(style, this.animationLoop);
+            this.style = new Style(style, this.animationLoop, this._workerCount);
         }
 
         this.style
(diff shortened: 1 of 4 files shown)


debug/index.html
```diff
@@ -58,7 +58,8 @@
     zoom: 12.5,
     center: [-77.01866, 38.888],
     style: 'mapbox://styles/mapbox/streets-v8',
-    hash: true
+    hash: true,
+    workerCount: 3
```
mourner: I don't think we need it set in the debug page — let's leave it to rely on `hardwareConcurrency`, which most of the browsers used by GL JS developers support. Or we could add but comment out.

anandthakker: @mourner My thinking was that, if the majority of users have 4 cores, it might actually be beneficial to have developers working under those conditions by default.

lucaswoj: I've been doing a little housecleaning on this debug page to keep it as simple as possible. My intent (we're not there yet) is for this page to be as simple as possible, making it an ideal place to build one-off testing rigs, and to put any persistent specialized setups on other pages (i.e. `video.html`)

tl;dr 👍 @mourner

anandthakker: 👍

anandthakker:
@lucaswoj @mourner amended the commit to remove the change to debug/index.html

js/ui/map.js
```diff
@@ -97,6 +98,7 @@ var defaultOptions = {
  * @param {number} [options.zoom] The zoom level of the map's initial viewport.
  * @param {number} [options.bearing] The bearing (rotation) of the map's initial viewport measured in degrees counter-clockwise from north.
  * @param {number} [options.pitch] The pitch of the map's initial viewport measured in degrees.
+ * @param {number} [options.workerCount=navigator.hardwareConcurrency] The number of WebWorkers the map should use to process vector tile data.
```
lucaswoj: `options.workerCount=navigator.hardwareConcurrency `**`- 1`**

js/ui/map.js
```diff
@@ -586,7 +594,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
         } else if (style instanceof Style) {
             this.style = style;
         } else {
-            this.style = new Style(style, this.animationLoop);
+            this.style = new Style(style, this.animationLoop, this._workerCount);
```
lucaswoj: I have a preference towards passing `options.workerCount` directly to `Style`. I'm trying to push `Map` in the direction of becoming a dumb [facade](https://en.wikipedia.org/wiki/Facade_pattern) class because it is currently a [god object](https://en.wikipedia.org/wiki/God_object).

test/js/ui/map.test.js
```diff
@@ -930,6 +931,21 @@ test('Map', function(t) {
         t.end();
     });
 
+    t.test('workerCount option', function(t) {
+        var map = createMap({});
+        // TODO: it's not great to check this private member here; better would
+        // be to mock Style and then just check that the correct workerCount
+        // param is being passed.
```
lucaswoj: We have a policy against checking `// TODO` comments into `master` because they tend to get forgotten. You should either resolve the todo or open a ticket.

test/js/ui/map.test.js
```diff
@@ -930,6 +931,21 @@ test('Map', function(t) {
...
+        t.equal(map._workerCount, browser.hardwareConcurrency - 1, 'workerCount defaults to hardwareConcurrency - 1');
+        map = createMap({ workerCount: 3 });
+        t.equal(map._workerCount, 3, 'workerCount option is used');
```
lucaswoj: How 'bout checking `map.style.dispatcher.actors.length` here? It also accesses private properties but is 🆗 for now.

lucaswoj:
Feel free to 🚢 once you've addressed my comments above!

js/ui/map.js
anandthakker: @lucaswoj As long as `Map` has a `setStyle` method, I think it will need to keep a copy like this of the `workerCount`, because it needs to be able to do `new Style(...)` not just in the map's constructor.  We _could_ add `workerCount` as a parameter to `setStyle`, but that seems a little odd to me.

test/js/ui/map.test.js
anandthakker: Ticketed here: https://github.com/mapbox/mapbox-gl-js/issues/2685

anandthakker:
@lucaswoj most comments addressed, but see my note [above](https://github.com/mapbox/mapbox-gl-js/pull/2666#discussion_r66053368)

js/ui/map.js
lucaswoj: Ah! I didn't consider the `setStyle` case. I agree that this `this._workerCount` is necessary for now 👍

lucaswoj:
Looks good! :shipit:

Issue #2506, opened 2016-04-28
Move to a newless API


mourner:
Most of the objects in the GL JS API are created using `new`. Leaflet had a huge success moving to a newless API design where instead of constructors, you expose factories:

``` js
// before
var map = new mapboxgl.Map();
new mapboxgl.Popup().addTo(map);

// after
var map = mapboxgl.map();
mapboxgl.popup().addTo(map);
```

This makes it easier for beginners, less prone to errors, and less verbose. It's also better to make global API changes like this sooner than later if we decide to do it.

The main difficulty of the switch would be to figure out how to support the change in the docs with documentation.js, with JSDoc not having a model of factory-based classes. Is there an easy solution?

cc @jfirebaugh @lucaswoj @tmcw

tmcw:
Depends on how you implement factory-based classes, but [we previously had support for the revealing module pattern](https://github.com/documentationjs/documentation/issues/379) and could again - it's just a tricky inference step. If you're doing the 'if I'm not an instance, make me an instance' trick, then I think things should 'just work' because you'll still be adding methods to a prototype?

jfirebaugh:
I think the proposal is to make factory methods for each class, rather than do the constructor trick. So the factory method is a static member of `mapboxgl` which `@returns {Class}`, and the constructor itself is `@private`. As long as the output order is sensible -- like the factory methods are either all at the top, or grouped with the instance methods of the corresponding class -- documentation.js should handle this fine.

odub:
Hey! Is there a way I could document a factory-style API with the current version of documentation.js? Maybe by sprinkling in some extra JSDoc tags?

jfirebaugh:
My views on this have changed. I think with the move toward ES6 classes throughout the JavaScript ecosystem, it makes the most sense just to embrace `new` as the one true way to create class instances. _Not_ supporting `new` would certainly be bucking a trend. Given that, does anyone still feel that including a newless API _in addition_ to a `new`ful API is important? (I don't.)

anandthakker:
> I think with the move toward ES6 classes throughout the JavaScript ecosystem, it makes the most sense just to embrace new as the one true way to create class instances. Not supporting new would certainly be bucking a trend.

Agreed, and no, I don't see a need to also have/support a newless API

mourner:
Agreed, let's close.

PR #2667, opened 2016-06-03
Refactor Source interface to allow/prepare for extensibility

anandthakker:
Replaces #2648 

 - Inverts `TilePyramid` and `Source` dependencies.  `TilePyramid` now essentially serves as a common wrapper class for dealing with sources.

 - Renames `TilePyramid` to `SourceCache`.  This is partly aspirational, and partly to make it more reasonable to have things called `source` that are actually TilePyramids 😁 

 - Eliminates the ability to create a `Source` independently of a `Map` instance - i.e., no more `new mapboxgl.GeoJSONSource()`; sources must be created with `map.addSource(id, json)`. 

 - Eliminates bucketStats, as it is no longer needed (so @lucaswoj assures me)

- Source _types_ are now classes with constructor signature `(id, options, dispatcher) => Source` and an optional `workerSourceURL` static property

- The `Source` provided by the factory function is an object implementing:

```js
id: string
tileSize: number
minzoom: number
maxzoom: number
roundZoom: boolean
reparseOverscaled: boolean
isTileClipped: boolean

loadTile: (tile: Tile, cb: (err, ?data) => void) => void
abortTile: (tile: Tile) => void
unloadTile: (tile: Tile) => void
serialize: () => Object // plain JS
?prepare: () => void

// @fires `load` to indicate source data has been loaded, so that it's okay to call `loadTile`
// @fires `change` to indicate source data has changed, so that any current caches should be flushed 
```

 - `WorkerSourceURL` points to a script, which, when executed in the worker thread, must call `self.registerWorkerSource(WorkerSource)`, where `WorkerSource` is a class that the worker will construct with `new WorkerSource(actor, styleLayers)`.

WorkerSource instances may implement a `loadTile` method; if they do, then this will be used by the worker when its `'load tile'` target is invoked with a `type: 'source-type'` parameter (e.g. `type: 'geojson'`).  Any other methods on `workerSource` instance can be targeted by the main-thread `Source` via `dispatcher.send('source-type.methodname', params, callback)`.

Questions:
 - [x] Does the above design have broad (enough) approval?
 - [x] The `Tile` objects are used by various source types to maintain tile-specific state; ideally, this implicit contract between `TilePyramid` and `Source` should be made explicit in the Source interface... but I think this is perhaps out of scope for this particular PR - perhaps this would be naturally addressed in #2703, or else could be ticketed separately?
 - [x] I'm not sure what the `data` yielded by `loadTile` should look like.  Currently, the only thing I can see being needed by code outside of the particular source is `bucketStats`, which the pyramid uses to emit the `tile.stats` event.  @lucaswoj can you weigh on in this?

Remaining Tasks:
 - [x] Video source - @lucaswoj if you can help me get this working, I'm assuming I can probably then manage Image source.
 - [x] Image source
 - [x] Settle Geojson Source design
   - [supercluster: separate source type, or option for geojson source?](https://github.com/mapbox/mapbox-gl-js/pull/2667#issuecomment-224285621)
   - Simplify the geojson source's WorkerSource so that it is (a) as clean/simple as possible, but (b) nicely supports 3rd party "geojson-like" source types
 - [x] Document the `Source` interface
 - [x] Document `WorkerSource`
 - [x] Document `addSourceType` -- includes documenting the `create` function and the `self.registerWorkerSource(...)` requirement on `WorkerSource` modules.
 - [x] Document "How to implement a source type" -- there's probably a bit more to say beyond the `Source`, `WorkerSource`, and `addSourceType`.  Perhaps an example of a 3rd-party source type would suffice. **update:** went with a link to https://github.com/developmentseed/mapbox-gl-topojson in the docs for `GeoJSONWorkerSource`.
 - [x] Update docs for ImageSource, VideoSource, GeoJSONSource to reflect that their constructors are no longer publicly available
 - [x] Fix any test regressions
 - [x] Add new unit tests as appropriate
 - [x] Cleanup - drop async loading of core source types, and do it in the "right" place.
 - [x] Cleanup - geojson is a core source type, and so I think we should hard-code require its worker source implementation directly in worker.js--in addition to avoiding the function (self) {} wrapper, it would be silly to webworkify it and waste time at runtime sending that code to all the worker threads
 - [x] Cleanup - move `loadTileJSON`, `queryRenderedFeatures`, and `querySourceFeatures` out of `source.js`
 - [x] Cleanup - fix [upstream bug in WebWorkify](https://github.com/substack/webworkify/issues/26)
 - [x] Cleanup - hunt & remove all `TODO`


js/style/style.js
```diff
@@ -66,12 +72,36 @@ function Style(stylesheet, animationLoop) {
         this.glyphSource = new GlyphSource(stylesheet.glyphs);
         this._resolve();
         this.fire('load');
+
+        if (validateStyle.emitErrors(this, validateStyle(stylesheet))) return;
```
jfirebaugh: Why did this move? Don't we want to continue to short circuit if there's a validation error?

anandthakker: Hm, yup - not quite sure when/how this happened, but it's a mistake. Thanks!

js/source/geojson_source.js
```diff
@@ -40,50 +54,31 @@ module.exports = GeoJSONSource;
...
-        add: this._addTile.bind(this),
-        remove: this._removeTile.bind(this),
-        redoPlacement: this._redoTilePlacement.bind(this)
-    });
+    this.type = options.type;
+
 }
 
-GeoJSONSource.prototype = util.inherit(Evented, /** @lends GeoJSONSource.prototype */{
+GeoJSONSource.prototype = /** @lends GeoJSONSource.prototype */{
```
jfirebaugh: The `@lends` is no longer necessary, since its now a direct assignment.

jfirebaugh:
Regarding `Source.is`, `addSource`, and `createSource`: I think we could take either of the following approaches:

* Remove the ability to create or add a source other than `addSource(id, json)`, where `json` is what you'd see in the style spec. (No more [`new mbgl.GeoJSONSource()`](https://www.mapbox.com/mapbox-gl-js/api/#GeoJSONSource), etc.)
* Not go the factory function route just yet. Do whatever additional configuration is needed (e.g. setting `dispatcher`) on the source only when `addSource` is called.

jfirebaugh:
> It's weird to rely on sources presenting an `id` property with the correct value.

Seems fine to me. The reason that `id` is a separate parameter in `addSource`, and not included in the object like it is with layers, is that the style spec uses an `id`-keyed object rather than an array. This is probably a design mistake and we should been consistent: use arrays, validate `id` uniqueness externally. Anyway, I think of `id` as being an intrinsic property for both layers and sources.

js/style/style.js
```diff
@@ -68,10 +76,32 @@ function Style(stylesheet, animationLoop) {
         this.fire('load');
     }.bind(this);
 
-    if (typeof stylesheet === 'string') {
-        ajax.getJSON(normalizeURL(stylesheet), loaded);
+    var sourceTypesLoaded = function(err) {
```
jfirebaugh: Can you explain this part of the code? It looks like its loading support for core source types asynchronously? That seems odd to me, both that it should be async, and that something like `addSourceType` should be the responsibility of `Style` and attached as a non-underscored method.

js/style/style.js
```diff
@@ -635,9 +665,8 @@ Style.prototype = util.inherit(Evented, {
         var sourceResults = [];
         for (var id in this.sources) {
             var source = this.sources[id];
-            if (source.queryRenderedFeatures) {
-                sourceResults.push(source.queryRenderedFeatures(queryGeometry, params, zoom, bearing));
-            }
+            var results = Source._queryRenderedVectorFeatures(source, this._layers, queryGeometry, params, zoom, bearing);
```
jfirebaugh: What happens here for raster/image/video sources?

js/style/style.js
```diff
@@ -318,13 +348,13 @@ Style.prototype = util.inherit(Evented, {
         if (this.sources[id] !== undefined) {
             throw new Error('There is already a source with this ID');
         }
-        if (!Source.is(source) && this._handleErrors(validateStyle.source, 'sources.' + id, source)) return this;
+        var builtIns = ['vector', 'raster', 'geojson', 'video', 'image'];
+        var shouldValidate = !Source.is(source) && builtIns.indexOf(source.type) >= 0;
+        if (shouldValidate && this._handleErrors(validateStyle.source, 'sources.' + id, source)) return this;
 
-        source = Source.create(source);
+        source = new TilePyramid(id, source, this.dispatcher);
```
jfirebaugh: Is there some way we can avoid having variables or members named `source` that are actually `TilePyramid`s? This seems pretty confusing.

js/source/clustered_geojson_worker_source.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+var supercluster = require('supercluster');
+var GeoJSONWorkerSource = require('./geojson_worker_base');
+
+var workerSource = Object.create(GeoJSONWorkerSource);
+workerSource.indexData = function (data, params, callback) {
```
lucaswoj: Would it be appropriate to call this method `clusterData`? I'm not completely sure what the word "index" means in this context.

js/source/geojson_source.js
```diff
@@ -1,13 +1,27 @@
...
 var EXTENT = require('../data/bucket').EXTENT;
 
-module.exports = GeoJSONSource;
+var webworkify = require('webworkify');
+
+module.exports.create = function (id, options, dispatcher, onChange, callback) {
+    var source = new GeoJSONSource(id, options, dispatcher, onChange);
+
+    //TODO: is it to aggressive to do this on source creation, as opposed to
+    //lazily at the first tile request?
```
lucaswoj: I don't think so. Be aggressive! Given the choice between using less bandwidth and showing users their data faster, I'd choose the latter.

js/source/geojson_source.js
```diff
@@ -95,82 +90,52 @@ GeoJSONSource.prototype = util.inherit(Evented, /** @lends GeoJSONSource.prototy
...
         var data = this._data;
         if (typeof data === 'string') {
             options.url = typeof window != 'undefined' ? urlResolve(window.location.href, data) : data;
         } else {
             options.data = JSON.stringify(data);
         }
-        this.workerID = this.dispatcher.send('parse geojson', options, function(err) {
+        this.workerID = this.dispatcher.send(this.type + '.parse', options, function(err) {
+            // TODO: not quite sure if we need this anymore
             this._loaded = true;
```
lucaswoj: I think we still need a `this._loaded` flag and `this.loaded` method (though not necessarily by those names) to implement #1715

js/source/geojson_source.js
```diff
@@ -95,82 +90,52 @@ GeoJSONSource.prototype = util.inherit(Evented, /** @lends GeoJSONSource.prototy
...
-                this.fire('change');
+                return callback(err);
             }
-
+            this._load(tile, callback);
         }.bind(this));
     },
 
-    _loadTile: function(tile) {
+    _load: function(tile, callback) {
```
lucaswoj: `GeoJSONSource#_load` and `GeoJSONSource#load` could be named more clearly for what they do. It seems like `loadTile` and `loadTileUncached` would be appropriate names. The important part is that we differentiate the notion of loading tiles from the notion of the source being loaded.

js/source/geojson_source.js
```diff
@@ -202,31 +166,24 @@ GeoJSONSource.prototype = util.inherit(Evented, /** @lends GeoJSONSource.prototy
...
-    _redoTilePlacement: function(tile) {
-        tile.redoPlacement(this);
+    serialize: function() {
+        return {
+            type: this.type,
+            data: this._data
+        };
     }
-});
+};
```
lucaswoj: I'm unclear on why some methods / properties are prefixed with an `_`. Isn't this whole class private?

js/source/geojson_worker_base.js
```diff
@@ -0,0 +1,98 @@
...
+var GeoJSONWrapper = require('./geojson_wrapper');
+var vtpbf = require('vt-pbf');
+var geojsonvt = require('geojson-vt');
+
+/*
+ * A base prototype to use for creating GeoJSON-based source workers.  This is
+ * basically an abstract class; subclasses should implement (at least) the
+ * indexData method.
+ */
+module.exports = Object.create({
```
lucaswoj: Why `Object.create` here?

js/source/clustered_geojson_worker_source.js
lucaswoj: I see that this method is used to invoke `geojsonvt` elsewhere. Would `createTileFromData` be a good name? I'm still unsure.

js/source/geojson_worker_source.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+var GeojsonWorkerSource = require('./geojson_worker_base');
+module.exports = function (self) {
+    self.registerWorkerSource('geojson', Object.create(GeojsonWorkerSource));
+};
```
lucaswoj: Would it be possible to rejigger things so that this file could read more like:

```js
module.exports = require('./geojson_worker_base');
```

js/source/geojson_worker_base.js
```diff
@@ -0,0 +1,98 @@
...
+            extent: params.extent,
+            maxZoom: params.maxZoom
+        };
+        try {
+            return callback(null, geojsonvt(data, geojsonVtOptions));
+        } catch (e) {
+            return callback(e);
+        }
+    }
+});
```
lucaswoj: I'm feeling a little [weird about the inheritance here](https://en.wikipedia.org/wiki/Composition_over_inheritance). Do you think there's a simpler design that uses smart defaults and/or composition and/or minor code duplication?

js/source/raster_tile_source.js
```diff
@@ -48,11 +43,9 @@ RasterTileSource.prototype = util.inherit(Evented, {
         };
     },
 
-    getVisibleCoordinates: Source._getVisibleCoordinates,
-    getTile: Source._getTile,
-
-    _loadTile: function(tile) {
+    load: function(tile, cb) {
```
lucaswoj: We usually write out the whole word `callback`. Avoid acronyms helps the codebase remain readable and approachable for everyone.

js/source/source.js
```diff
@@ -3,70 +3,32 @@
...
-        return;
-    }
-
-    var ids = this._pyramid.orderedIDs();
-    for (var i = 0; i < ids.length; i++) {
-        var tile = this._pyramid.getTile(ids[i]);
-        this._redoTilePlacement(tile);
+        browser.frame(loaded.bind(null, null, options));
     }
 };
```
lucaswoj: Do you think it'd make sense to put `loadTileJSON`, `queryRenderedFeatures`, and `querySourceFeatures` into standalone files? They don't gain much by being coupled together with `Source.create` and the source type directory.

js/source/source.js
```diff
@@ -160,34 +114,28 @@ exports._querySourceFeatures = function(params) {
...
+    // work as naturally for this.  Is Source.is() still needed?
+
+    return source && source.id;
+    // for (var type in sourceTypes) {
+    //     if (source instanceof sourceTypes[type]) {
+    //         return true;
+    //     }
+    // }
+    // return false;
+};
```
lucaswoj: We use `Source.is` to differentiate between these two cases:

```js
var source0 = new mapboxgl.GeoJSONSource(...);
map.addSource('foo', source0);

var source1 = { type: 'geojson' };
map.addSource('bar', source1);
```

I'm all for eliminating the method if possible. Exposing `GeoJSONSource` in our public API was a weird design choice but it'll be hard to go back on within the scope of this PR.

lucaswoj: If using a `new`-based API is going to be cleaner than using a factory-based API, I'm all for it.

js/source/tile_pyramid.js
```diff
@@ -16,24 +17,65 @@ module.exports = TilePyramid;
...
+
+    // TODO: this is an ugly consequence of inverting tilepyramid / source
+    // dependency and making source creation async: we use a dumb, generic
+    // version of serialize() until the source is loaded at which point we
+    // switch to deferring to the source's method.
+    // Options:
+    // 1. Make Source.create synchronous.  Drawback: would require adding `loaded()` or `onLoad`  to the Source interface.
+    // 2. Remove Source:serialize() entirely, and instead let TilePyramid:serialize() always jut return the options object it was created with.
+    // 3. Leave the below wart as-is.
+    // I prefer option 2 if it's workable.
```
lucaswoj: I don't think option 2 will work in the case that a user

 - creates a `GeoJSONSource`
 - calls `GeoJSONSource#setData`

Option 1 may be cleanest. 

> Exposing GeoJSONSource in our public API was a weird design choice but it'll be hard to go back on within the scope of this PR. (this again)

js/source/tile_pyramid.js
```diff
@@ -16,24 +17,65 @@ module.exports = TilePyramid;
...
+    this.serialize = function () { return options; };
+
+    var sourceLoaded = function (err, source) {
+        if (err) {
+            this.fire('error', {error: err});
+            return;
+        }
+
+        // TODO: remove this once we eliminate need for sources to have a
+        // reference to the Map instance.
```
lucaswoj: ❤️

js/source/tile_pyramid.js
```diff
@@ -16,24 +17,65 @@ module.exports = TilePyramid;
...
+        this._load = (source.load || noop).bind(source);
+        this._abort = (source.abort || noop).bind(source);
+        this._unload = (source.unload || noop).bind(source);
+        this.serialize = (source.serialize || noop).bind(source);
+        this.prepare = source.prepare && source.prepare.bind(source);
+
+        this.fire('load');
+    }.bind(this);
+
+    Source.create(id, options, dispatcher, onChange, sourceLoaded);
```
lucaswoj: It seems like there are a number of chicken-and-egg problems with `Source`s and `TilePyramids` right now. What do you think of this design:

`TilePyramid` (maybe renamed to `SourceCache`) is treated as a [wrapper / decorator](https://en.wikipedia.org/wiki/Decorator_pattern) to `Source`s


```js
var source = new GeoJSONSource(...);
source.loadTile(...); // does all kinds of HTTP fetching and tile generation
var sourceCache = new SourceCache(...);
sourceCache.loadTile(...); // immediately returns tile from the cache if possible, otherwise sends the request to the underlying source
```

I admit that the current `TilePyramid` doesn't perfectly map onto the proposed `SourceCache` class but I think we can make it work / ticket out some tail work.

js/source/tile_pyramid.js
lucaswoj: By the time we add an `onLoad` parameter, it might be smart to require sources to mixin `Evented`. It's slightly more work for 3rd parties, I admit, but it helps us keep our architecture clean and consistent.

js/style/style.js
lucaswoj: Would [this comment](https://github.com/mapbox/mapbox-gl-js/pull/2667/files/3d6cf0a5ce5d118c52d2fe4446daffad25dd79b2#r65789289) tie in here?

js/style/style.js
lucaswoj: This would be fixed if we decide to go ahead with https://github.com/mapbox/mapbox-gl-js/pull/2667/files/3d6cf0a5ce5d118c52d2fe4446daffad25dd79b2#r65790830

js/style/style.js
anandthakker: @jfirebaugh Yeah, once this refactor settles, I think this async weirdness can be eliminated.  It's here now only as a quick and dirty way to validate/show that `addSourceType` works, but ultimately there's no reason that core source types need to be added asynchronously through the `addSourceType` interface.

I do think `addSourceType` should:
 - (a) be a public, non-underscored method -- because its purpose is to allow third-party source types can be plugged in, which is the motivating reason for this PR; and
 - (b) be async -- because if a source type needs to provide a WorkerSource implementation, then the process of having all the workers load that implementation via `importScript` is async.

js/style/style.js
anandthakker: @jfirebaugh I _think_ it'll just return `[]`, but I was counting on getting that straightened out during the process of reinstating the many broken unit tests

js/source/geojson_source.js
anandthakker: @lucaswoj Yeah, this should be made consistent... 

> Isn't this whole class private?

Hmm -`minzoom`, `maxzoom`, `load`, `serialize`, etc. are implementing the common "Source" interface that we're trying to establish in this PR.  `setData` is a user-facing method.  It seems to me that these should be public... am I misunderstanding your question?

anandthakker: @lucaswoj Yeah, this should be made consistent... 

> Isn't this whole class private?

Hmm -`minzoom`, `maxzoom`, `load`, `serialize`, etc. are implementing the common "Source" interface that we're trying to establish in this PR.  `setData` is a user-facing method.  It seems to me that these should be public... am I misunderstanding your question?

js/source/geojson_worker_base.js
anandthakker: A leftover, not needed.

js/source/geojson_worker_source.js
anandthakker: @lucaswoj Yeah, I think this can/will get cleaned up as part of addressing https://github.com/mapbox/mapbox-gl-js/pull/2667#discussion_r65789437

js/source/source.js
anandthakker: > If using a new-based API is going to be cleaner than using a factory-based API, I'm all for it.

I don't really see the benefit of being able to construct a `Source` object free of the map instance, and it seems to me that that is the main thing that a new-based API affords.

js/source/tile_pyramid.js
anandthakker: Based on a conversation with @lucaswoj on Slack yesterday, I'd like to propose:
 - [ ] In the scope of this PR, we just rename `TilePyramid` to `SourceCache`, addressing @jfirebaugh 's [comment](https://github.com/mapbox/mapbox-gl-js/pull/2667#discussion_r65787211) and preparing for:
 - [ ] Move the further refactoring of the `TilePyramid` ( aka `SourceCache`) interface to a ticket / future PR

mourner:
I'm still not sure if we want to have `geojson-clustered` as a separate type for clustering, and generally have different type names for the same source with slightly different interpretation. I'm more comfortable with the current API because _the source is GeoJSON_ whether clustered or not — clustering is a matter of interpretation. Suppose we want to make a plugin for clustered TopoJSON — would we have a separate source named `topojson-clustered` and somehow reuse parts of both `geojson-clustered` and `topojson` sources?

anandthakker:
@mourner I definitely see what you're saying, but I'm not 100% sold on the idea that different source types should only map to the format of the original source data being used.  I agree that clustering is an interpretation of the data; but then would hex-bins and square-bins also just be interpretations that belong as options on GeoJSONSource?

js/source/source.js
```diff
@@ -116,6 +116,11 @@ var sourceTypes = {};
  */
 exports.create = function(id, source, dispatcher) {
     source = sourceTypes[source.type](id, source, dispatcher);
+
+    if (source.id !== id) {
+        throw new Error('Expected Source id to be ' + id + ' instead of ' + source.id);
+    }
```
anandthakker: Actually, I think it probably just makes more sense to do `source.id = id` here.

js/source/geojson_source.js
anandthakker: @lucaswoj looking at this again, I think in this case the difference between `_load` and `load` is that the latter might first send new data to the worker before loading the tile.  Just noting that here for reference -- the broader issue of `load` being a poorly descriptive method name still stands.

anandthakker:
@lucaswoj @jfirebaugh @mourner Updated the top of the ticket with more complete description of the overall design and capturing outstanding questions/tasks from the reviews thus far (so we should now be okay to rebase w/o losing key discussion points).

anandthakker:
Transcribing from a diff comment:

@jfirebaugh : [It] seems odd to me, both that it should be async, and that something like addSourceType should be the responsibility of Style and attached as a non-underscored method.

@anandthakker : I do think addSourceType should:

 - (a) be a public, non-underscored method -- because its purpose is to allow third-party source types can be plugged in, which is the motivating reason for this PR; and
 - (b) be async -- because if a source type needs to provide a WorkerSource implementation, then the process of having all the workers load that implementation via importScript is async.

lucaswoj:
The high level plan is looks very good! Thank you for taking the time to do that writeup. 

---

> - Eliminates the ability to create a `Source` independently of a `Map` instance - i.e., no more `new mapboxgl.GeoJSONSource()`; sources must be created with `map.addSource(id, json)`. 

I'm very much in favor of this change. 

We will need to accommodate the `GeoJSONSource#setData` use case. I propose we allow `Map#addSource` to atomicly overwrite an existing source. With that change, it might make sense to rename to `Map#setSource`.

---

>  - `WorkerSourceURL` points to a module exporting a function: `(self) => void` that should call `self.registerWorkerSource(workerSource: WorkerSource)`

Could we make `WorkerSourceURL` point to a module exporting a function `(callback: (workerSource: WorkerSource) => void) => void`? The use of `self` feels unnecessary. 

---

>  - The `Tile` objects are used by various source types to maintain tile-specific state; ideally, this implicit contract between `TilePyramid` and `Source` should be made explicit in the Source interface... but I think this is perhaps out of scope for this particular PR - perhaps this would be naturally addressed in #2703, or else could be ticketed separately?

My preference here is to factor out tile state into other places (i.e. caching metadata can live in a separate data structure within `SourceCache`). I will ticket this out separately.

---

>  - I'm not sure what the `data` yielded by `loadTile` should look like.  Currently, the only thing I can see being needed by code outside of the particular source is `bucketStats`, which the pyramid uses to emit the `tile.stats` event.  @lucaswoj can you weigh on in this?

Would it help to move `bucketStats` into the `Tile` object? 

If this becomes a hassle, I wouldn't be opposed to entirely removing `bucketStats` for the moment and figuring out how to reimplement it atop this architecture from a clean slate.

-----

> - Video source - @lucaswoj if you can help me get this working, I'm assuming I can probably then manage Image source.

I'll dig in ASAP. 

-----

> - Settle Geojson Source design
   - [supercluster: separate source type, or option for geojson source?](https://github.com/mapbox/mapbox-gl-js/pull/2667#issuecomment-224285621)

Would it be possible to keep a single GeoJSON source for now and consider breaking changes in another PR? I imagine this could be done with a little branching here and there.

anandthakker:
> We will need to accommodate the GeoJSONSource#setData use case. I propose we allow Map#addSource to atomicly overwrite an existing source. With that change, it might make sense to rename to Map#setSource.

This looks nice to me, but it would also be possible to postpone this for later, because even under the current proposal, it would be possible to do `map.getSource(id).setData(...)`.  (Not opposed to doing the setSource thing here, but I mention this just in case there's advantage to getting this PR shipped sooner to allow @mollymerp 's work to proceed.)

anandthakker:
> Could we make WorkerSourceURL point to a module exporting a function (callback: (workerSource: WorkerSource) => void) => void? The use of self feels unnecessary.

Actually, I miswrote -- it should have said, "WorkerSourceURL points to a **script, which, when executed in the worker thread, must** call self.registerWorkerSource(workerSource: WorkerSource)".  The issue here is that we're really shipping a _script_, not a CJS module, over to the worker.  The nicest way to package that up will likely be to use `webworkify`, which is where the need for a `function (self) { ... }` wrapper is coming from.  I agree that this wrapper adds an unfortunate layer of indirection.  Two points to address that:
1. `geojson` is a core source type, and so I think we should hard-code `require` its worker source implementation directly in `worker.js`--in addition to avoiding the `function (self) {}` wrapper, it would be silly to webworkify it and waste time at runtime sending that code to all the worker threads!
2. Low priority, but at some point, we could develop a variation to `webworkify` that would allow 3rd-party source type authors to export something nicer, and that would take care of packaging it into a script that correctly calls `registerWorkerSource`.  This may or may not be worth the extra maintenance, depending on demand.

anandthakker:
> Would it be possible to keep a single GeoJSON source for now and consider breaking changes in another PR? I imagine this could be done with a little branching here and there.

 👍 SGTM

lucaswoj:
> This looks nice to me, but it would also be possible to postpone this for later, because even under the current proposal, it would be possible to do `map.getSource(id).setData(...)`.

Ah! Forgot about that. This is a good plan.

lucaswoj:
> Actually, I miswrote -- it should have said, "WorkerSourceURL points to a script, which, when executed in the worker thread, must call self.registerWorkerSource(workerSource: WorkerSource)". 

Sounds good to me.

anandthakker:
> Would it help to move bucketStats into the Tile object?
> If this becomes a hassle, I wouldn't be opposed to entirely removing bucketStats for the moment and figuring out how to reimplement it atop this architecture from a clean slate.

Talked with @lucaswoj on Slack -- we're going to just remove bucketStats and (for now) make the callback to `Source#loadTile` just `(err) => void` (no data passed).  This is because, currently sources are responsible for mutating the `tile` object (i.e., sticking vector / raster data onto them); this statefulness / implicit contract between Source and the rest of the world will be refactored a bit later.

anandthakker:
@lucaswoj As of d6ede33, the GeoJSONSource + GeoJSONWorkerSource are ready for a review.  We're back to a single source with the `cluster:` option, and I think the worker-side is now (a) fairly simple while (b) still general enough to support 3rd parties creating alternate source types in a very straightforward way.

mourner:
@anandthakker nice! It seems that `clusterMaxZoom` and `clusterRadius` options should be brought back too (they're currently not there).

anandthakker:
> @anandthakker nice! It seems that clusterMaxZoom and clusterRadius options should be brought back too (they're currently not there).

Oops! Thanks for the catch @morganherlocker

anandthakker:
All preexisting tests are now passing (locally, anyway -- we'll see what Circle says).  Next up is to think through what new tests should be added here.

anandthakker:
Rebased onto master; previous is at fd15bc657bf8e9513b831179456ccdc2c32c4817

anandthakker:
@lucaswoj @mourner One thing that I noticed during that rebase: the change in https://github.com/mapbox/mapbox-gl-js/commit/dceac30bbf425a3a095f1d78c5f8ab72da7e2d8b has `queryRenderedFeatures` returning `[]` instead of `{}` if the source isn't loaded.  I'm unclear on why the function should return an array in that case, when [`mergeRenderedFeatureLayers`](https://github.com/mapbox/mapbox-gl-js/blob/dceac30bbf425a3a095f1d78c5f8ab72da7e2d8b/js/source/source.js#L118) still returns an object.

anandthakker:
@lucaswoj I just removed the `TODO`'s referring to the aspiration to remove references to `map` within `SourceCache` -- could I leave it to you to ticket that out?

anandthakker:
> Add new unit tests as appropriate

As of caef048 I _think_ this should be handled, in the sense that the only new surface area that I could think of was the lifecycle coordination between SourceCache and Source, which are covered [here](https://github.com/mapbox/mapbox-gl-js/blob/caef0488e64ed2c6fc9c5d062a18ab0adf5b50f7/test/js/source/source_cache.test.js#L196-L226).  A couple of old tests had to be moved/reworked ( b0e6d1c f89b922 1bd3875 ), and so those probably merit a careful look to make sure I didn't drop something important.

If that all sounds right, then I think the only major work that remains is:
 - (a) ImageSource & VideoSource (which I'm still stumped on; I might need a bit of a primer on what exactly the GL-related code there is doing to get farther)
 - (b) Finish documentation.  There may be a couple of decisions to make here about where & how best to put the documentation for, e.g. the `Source` interface, `addSourceType`, etc -- these are "public" in the sense that they're not _internal_, but their audience also isn't the same as that for, say, `GeoJSONSource#setData()`.  Is that a distinction you want to make? @lucaswoj @mourner @jfirebaugh @tmcw

anandthakker:
> the only new surface area that I could think of was the lifecycle coordination between SourceCache and Source

Err, I guess an additional one that should have been obvious is `addSourceType`.  Heh.

anandthakker:
So close!  Not sure why the video render test is failing -- `video.html` debug file looks fine now.

anandthakker:
Video render test fix upstream at https://github.com/mapbox/mapbox-gl-test-suite/pull/115

anandthakker:
Another rebase, previous was 05fd3b45c8e273acdd4793a4ccad738ec07d4c6e

anandthakker:
> Update docs for ImageSource, VideoSource, GeoJSONSource to reflect that their constructors are no longer publicly available

I'm not quite sure about the best approach here.  At first I thought we could simply drop the constructor docs, leaving only the instance methods (`setData`, `setCoordinates`, etc.), and basically referring to the style spec for the source-creation options that are currently being documented as params to the constructor.  However, the `cluster*` options for GeoJSONSource complicate this, since they're not covered by the style spec.

anandthakker:
^ thoughts/opinions on that documentation issue, @tmcw @jfirebaugh @mourner @lucaswoj ?

jfirebaugh:
We're intending to support the same set of clustering options in gl-native, so I don't see an issue with moving them to the style specification so long as the API documentation has a clear cross-reference.

anandthakker:
Just updated https://github.com/anandthakker/mapbox-gl-topojson to work with the present state of this branch.  I think this could serve as a minimal, straightforward example of how to implement a 3rd-party "geojson-like" source type (which leverages the machinery already present for GeoJSONSource).

IMO, in-depth docs/guides for creating a third-party source type from _scratch_ (i.e. _not_ leveraging GeoJSONSource but doing all the work of providing `VectorTile` objects) may be better of being ticketed out of the scope of this PR, to be done after a couple such source types have been created in-house or by early adopters.  (Won't be too long before that happens -- I'll be creating a dynamic-data-joining source very soon.)

Either way, a question: does the (relatively) straightforward topojson source type belong in the official docs/examples?  Or would it be better to just link to that repo from the `addSourceType` docs?  (If the latter, I can move it over to `mapbox` so that the docs aren't relying on a link external to the org.)

anandthakker:
> We're intending to support the same set of clustering options in gl-native, so I don't see an issue with moving them to the style specification so long as the API documentation has a clear cross-reference.

Turns out they're in the spec but not in the generated docs.  Upstream: https://github.com/mapbox/mapbox-gl-style-spec/issues/463

anandthakker:
Okay, all boxes checked.  @jfirebaugh @mourner I think this is ready for another review if/when you have a chance.

anandthakker:
Updated the top of the ticket to reflect most up to date API

debug/places.geojson
```diff
@@ -0,0 +1 @@
...ank":5,"name":"Is. Martin Vaz","comment":null,"name_alt":null,"lat_y":-20.559422,"long_x":-29.338439,"region":"Seven seas (open ocean)","subregion":"Southern Atlantic Ocean","featureclass":"island"},"geometry":{"type":"Point","coordinates":[-29.338429328510017,-20.559502862505923]}},{"type":"Feature","properties":{"scalerank":5,"name":"Rockall","comment":null,"name_alt":null,"lat_y":58.163524,"long_x":-12.408715,"region":"Seven seas (open ocean)","subregion":"North Atlantic Ocean","featureclass":"island"},"geometry":{"type":"Point","coordinates":[-12.408741828510017,58.16339752811908]}},{"type":"Feature","properties":{"scalerank":5,"name":"I. de Cozumel","comment":null,"name_alt":null,"lat_y":20.444687,"long_x":-86.880555,"region":"North America","subregion":null,"featureclass":"island"},"geometry":{"type":"Point","coordinates":[-86.88060462147877,20.444708563275327]}},{"type":"Feature","properties":{"scalerank":5,"name":"Bermuda Islands","comment":null,"name_alt":null,"lat_y":32.317339,"long_x":-64.742895,"region":"Seven seas (open ocean)","subregion":"North Atlantic Ocean","featureclass":"island"},"geometry":{"type":"Point","coordinates":[-64.74290930897877,32.31726715702533]}}]}
```
lucaswoj: Do you think it'd be possible to get rid of `random.geojson` and use `places.geojson` on all our debug pages?

js/source/geojson_source.js
```diff
@@ -1,28 +1,23 @@
...
-var TilePyramid = require('./tile_pyramid');
-var Source = require('./source');
+var util = require('../util/util');
 var urlResolve = require('resolve-url');
 var EXTENT = require('../data/bucket').EXTENT;
 
-module.exports = GeoJSONSource;
+module.exports.create = function (id, options, dispatcher) {
+    return new GeoJSONSource(id, options, dispatcher);
+};
```
lucaswoj: Why export a method that calls the constructor rather than export the constructor, as before?

js/source/geojson_source.js
```diff
@@ -1,28 +1,23 @@
...
- * Create a GeoJSON data source instance given an options object
- * @class GeoJSONSource
- * @param {Object} [options]
- * @param {Object|string} options.data A GeoJSON data object or URL to it. The latter is preferable in case of large GeoJSON files.
- * @param {number} [options.maxzoom=18] Maximum zoom to preserve detail at.
- * @param {number} [options.buffer] Tile buffer on each side in pixels.
- * @param {number} [options.tolerance] Simplification tolerance (higher means simpler) in pixels.
- * @param {number} [options.cluster] If the data is a collection of point features, setting this to true clusters the points by radius into groups.
- * @param {number} [options.clusterRadius=50] Radius of each cluster when clustering points, in pixels.
- * @param {number} [options.clusterMaxZoom] Max zoom to cluster points on. Defaults to one zoom less than `maxzoom` (so that last zoom features are not clustered).
```
lucaswoj: Is this documentation being moved somewhere?

debug/places.geojson
anandthakker: @lucaswoj yep.  I pulled `places.geojson` from `mapbox-gl-test-suite`, along with things like `0.png` and a font pbf file.  You okay with those assets being duplicated in this repo?

js/source/geojson_source.js
anandthakker: Discussed this above with @jfirebaugh -- see https://github.com/mapbox/mapbox-gl-js/pull/2667#issuecomment-227453069

js/source/geojson_source.js
```diff
@@ -37,56 +32,84 @@ module.exports = GeoJSONSource;
...
+        }
+    }, options.workerOptions);
+
+    this._sendDataToWorker(function done(err) {
+        if (err) {
+            this.fire('error', {error: err});
+            return;
+        }
+        this.fire('load');
+    }.bind(this));
```
lucaswoj: What does this do?

js/source/geojson_source.js
anandthakker: Ah, shoot -- this was here because I had seen https://github.com/mapbox/mapbox-gl-js/issues/2506 with (a) no real objections and (b) marked on the v1.0 milestone... but @lucaswoj you mentioned in Slack that you'd rather wait and make that change wholesale, and I just lost track of the task of reverting to the `new`-based API here.

js/source/geojson_source.js
anandthakker: `_sendDataToWorker` used to be called `_updateData` -- it's responsible for invoking WorkerSource's `geojson.loadData` target (previously `'parse geojson'`), which handles loading the geojson data and preparing to serve it up as tiles, using `geojson-vt` or `supercluster` as appropriate.

This happens here in the constructor so we can get initial value of `data` from the source declaration over to the worker before firing `load` or `error`.  It will also happen in `loadTile` after `setData` has been called on the source.

js/source/load-tilejson.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
```
lucaswoj: This file should be `load_tilejson.js`

debug/places.geojson
lucaswoj: I understand the need for duplication and prefer as little duplication as is practical

js/source/geojson_source.js
lucaswoj: > It will also happen in `loadTile` after `setData` has been called on the source.

Why is it called in `loadTile` rather than `setData`? I think that is the source of my confusion.

debug/places.geojson
anandthakker: 😆 

One way to avoid duping these assets would be to modify the debug server to include `node_modules/mapbox-gl-test-suite/` as an additional root directory to serve things from, and then update the urls to things like `data/places.geojson`, `image/0.png`, etc.

js/source/geojson_source.js
anandthakker: Good question.  In `master` it's being called in `update` (not `setData`), but now that you point it out, I don't see any reason not to put it in `setData`.  I think one implication of doing this would be that we would want to wait till the callback to fire `change` (which is currently fired synchronously in `setData`), but that seems fine.

documentation.yml
```diff
@@ -53,3 +53,13 @@ toc:
       controls or other extensions.
   - EventData
   - Evented
+  - name: Custom Sources
+    description: |
+      These interfaces and methods allow you to create and use third-party data
+      sources.  See also [Map#addSourceType](#Map#addSourceType)
+  - Source
+  - WorkerSource
+  - CreateSourceFunction
```
lucaswoj: What is `CreateSourceFunction`?

anandthakker: Ah -- it is documented [here](https://github.com/mapbox/mapbox-gl-js/pull/2667/files#diff-d5506f072d1241401a7e31ff642c1606R103), but misspelled, so it wouldn't have shown up in a text search.  Anyway, this was for the source factory function, which is going away per [your comment above](https://github.com/mapbox/mapbox-gl-js/pull/2667#discussion_r69032008)

js/source/geojson_source.js
anandthakker: Fixed in 0843f9c73eac5a76d0c20993b46505dbed2417bc

debug/places.geojson
anandthakker: ^ done in 73a35079c9537dcdc985683e5262ac15309173b3

anandthakker:
@lucaswoj reverted to `new`-based API, and addressed your other comments as well (I think).

anandthakker:
rebased, previous at 73a35079c9537dcdc985683e5262ac15309173b3

anandthakker:
@lucaswoj An additional change in the last commit, for your 👀

This change clarifies worker-side tile loading responsibilities, making them explicit in the WorkerSource interface and moving the vector tile loading/parsing logic into VectorTileWorkerSource.

This paves the way for WorkerSource implementations that want to do more than just providing GeoJSON or VectorTile objects (e.g., custom implementations of parsing data into Buckets, as in the case of a source that accepts new properties and joins them to existing geometries).

Relatedly, I think it may also help with the work proposed in https://github.com/mapbox/mapbox-gl-js/issues/2703, because:
 - If I understand it correctly, the "tile combinator and symbol placer" will require the Worker to maintain a store of loaded & parsed tiles.
 - While it's true that this is already happening (with `this.loaded` / `this.loading`), I think that using the same store for both the tile loading/reloading logic _and_ the "tile combinator and symbol placer" would inhibit the independence of the combinator/symbol placer as its own module.

anandthakker:
Ah, I also forgot to mention: I think pulling out the `VectorTileWorkerSource` logic like this will also help to make https://github.com/mapbox/mapbox-gl-js/issues/899 much more approachable

anandthakker:
Rebased onto master to pick up changes from https://github.com/mapbox/mapbox-gl-js/pull/2847

anandthakker:
Rebased to pick up #2880

lucaswoj:
# Next Steps

 - [ ] rebase onto master 
 - [ ] start debug server and do sanity check
 - [ ] run all benchmarks
 - [ ] do final code read-through
 - [ ] 🐑

lucaswoj:
I'm about to push a rebased branch. Previous HEAD was https://github.com/mapbox/mapbox-gl-js/pull/2667/commits/def9c24f9b069c9f50b8adf58c0c5118d2488dfb

lucaswoj:
# Benchmark Results

All looks good 👍 

## `buffer`

**source-refactor**: 885 ms
**master**: 890 ms

## `fps`

**source-refactor**:  60fps
**master**: 59fps

## `frame-duration`

**source-refactor**: 7.8 ms per frame. 0% of frames took longer than 16ms.
**master**: 8.2 ms per frame. 1% of frames took longer than 16ms.

## `geojson-setdata-small`

**source-refactor**: 26ms
**master**: 27ms

## `geojson-setdata-large`

**source-refactor**: 123ms
**master**:182ms

lucaswoj:
Code coverage of this branch is 89.3%. Master is 89%. 👍

anandthakker:
🎉

mourner:
@lucaswoj For some reason, the build fails in master since the merge: https://circleci.com/gh/mapbox/mapbox-gl-js/4209

lucaswoj:
Thanks for the heads-up @mourner. I'll try to resolve ASAP.

hdrdavies:
@mourner how can one now configure clustering as was possible before i.e.

```
        let sourceObj = {
            type: 'foo',
            cluster: true,
            clusterRadius: 50,
            clusterMaxZoom: 14,
            data
        };

        map.addSource('foo', sourceObj);
```

mourner:
@hdrdavies same as before, as you can see on the example https://www.mapbox.com/mapbox-gl-js/example/cluster/

hdrdavies:
@mourner thanks very much!! It seems I'm getting other strange constructor errors even though I'm not using ```new GeoJsonSource``` anymore... 

Do you think you guys will be adding support for the clustering of custom HTML markers made using ```new mapboxgl.Marker``` ? Maybe I'm missing something in the documentation. Apologies if so.

mourner:
@hdrdavies currently no plans to do this for the HTML markers — this would be a non-trivial effort. Perhaps someone will contribute a plugin.

lucaswoj:
Thanks @mourner. @hdrdavies, you might be able to write your own HTML clustering code using the same backend as GL JS https://github.com/mapbox/supercluster

PR #2668, opened 2016-06-04
text-pitch-alignment

yhahn:
Implements full support for the proposed `text-pitch-alignment` property.

+ | rotation=map | rotation=viewport
--- | --- | ---
**pitch=map** | d | a
**pitch=viewport** | c | b

### How it works

- Adds passing of individual glyph angles as a vertex shader attribute
- Pass map pitch, bearing, aspect_ratio (for adjusting pitched angle) to SDF shader
- The SDF vertex shader is now responsible for rotating quads

### Backlog

- [x] Unit tests for symbol bucket/quad changes
- [x] Render tests for unskewed perspective map labels
- [x] Benchmarking to see how much the bigger burden on the vertex shader affects performance
- [x] Style spec decisions -- trying to think of reasons we would not just make this the default for map-oriented line labels (https://github.com/mapbox/mapbox-gl-style-spec/issues/459)
- [x] GL native port https://github.com/mapbox/mapbox-gl-native/pull/5288

cc @lucaswoj @mollymerp @ansis


js/data/bucket/symbol_bucket.js
```diff
@@ -469,10 +469,13 @@ SymbolBucket.prototype.addSymbols = function(programName, quads, scale, keepUpri
         // Lower min zoom so that while fading out the label it can be shown outside of collision-free zoom levels
         if (minZoom === placementZoom) minZoom = 0;
 
-        var index = addVertex(vertexArray, anchorPoint.x, anchorPoint.y, tl.x, tl.y, tex.x, tex.y, minZoom, maxZoom, placementZoom);
-        addVertex(vertexArray, anchorPoint.x, anchorPoint.y, tr.x, tr.y, tex.x + tex.w, tex.y, minZoom, maxZoom, placementZoom);
-        addVertex(vertexArray, anchorPoint.x, anchorPoint.y, bl.x, bl.y, tex.x, tex.y + tex.h, minZoom, maxZoom, placementZoom);
-        addVertex(vertexArray, anchorPoint.x, anchorPoint.y, br.x, br.y, tex.x + tex.w, tex.y + tex.h, minZoom, maxZoom, placementZoom);
+        // Encode angle of line together with symbol.alongLine
+        var placementAngleLine = symbol.curved ? 0 : 1 + Math.round(((angle % Math.PI) / Math.PI) * 255);
```
mourner: `angleLine` name implies a line rather than angle, perhaps `lineAngle` would be better?

js/render/draw_symbol.js
```diff
@@ -113,24 +113,26 @@ function drawSymbol(painter, layer, posMatrix, tile, bucket, bufferGroups, isTex
...
-        gammaScale = 1 / Math.cos(tr._pitch);
-        extrudeScale = [s, s];
-    } else {
-        s = painter.transform.altitude * fontScale;
-        gammaScale = 1;
-        extrudeScale = [ tr.pixelsToGLUnits[0] * s, tr.pixelsToGLUnits[1] * s];
-    }
+    var s1 = painter.transform.altitude * fontScale;
+    var s2 = pixelsToTileUnits(tile, 1, painter.transform.zoom) * fontScale;
+    var gammaScale = 1; // 1 / Math.cos(tr._pitch);
```
mourner: What's up with this variable? Should it be removed now?

yhahn: Yeah, I need to vet this further -- it adjusts the SDF sharpness when pitching the map but the previous calculation was not giving good results.

lucaswoj:
I'm not sold on this PR from a map-design-asthetic standpoint. All the labels should be "billboards" or none of them should be.
 
screen shot 2016-06-06 at 3 48 34 pm

For reference, here is a screenshot that demonstrates how Apple Maps draws labels. 

___nh_ch___p_m__n_h__nh_2016-06-06_l__c_15 50 04_720

(some wordage and screenshots from @1ec5)

PaulZed:
@lucaswoj Can you justify why it's inherently better to be consistent? I'm not sure what the objection is if unskewing the text increases readability. Roads like beachmont are much easier to resolve when they're unskewed when compared to their skewed version. There also isn't really a clear way to unskew linearly independent glyphs using this strategy, so until we discover a better solution I agree with not attempting to unskew them.

dpieri:
I agree with @lucaswoj on this one. The mixture of skewed and not skewed labels really messes with perspective and makes the map look distorted. I think it adds to the cognitive load required to orient yourself in 3D when looking at the map.

yhahn:
Per @lucaswoj + @1ec5's feedback I took another pass at unskewing curved labels. Overall the new approach is much cleaner:

- No codepath difference between how straight/curved line labels are rendered (no more curved vs straight heuristic),
- Visual result is pretty good

But there is one major caveat -- the way quads are positioned for curved labels isn't forgiving to this approach beyond around 45 degrees of pitching (and our general strategy for placing glyphs along curves has some :confounded: conflicts with placement on a pitched map that I can capture in a separate ticket). Also worth noting that most other map apps/renderers don't allow extreme pitching even beyond 30 degrees or so.

That said I definitely think this is the approach to go with right now and we can circle back around to a larger refactor around curved labels if needed in the future.

**master** | **unskew v2**
--- | ---
unskew-master | unskew-v2

Churning through the rest of the backlog above and native port next before we land on the style spec question.

lucaswoj:
This is awesome @yhahn! Even with the caveats, I feel 😍 about merging this from map-design-asthetic standpoint. Let me know if I can help with the remaining technical backlog.

PaulZed:
That example is pretty compelling :). You can definitely see the difference in glyph rotation for curved roads but it's quite minor. My only concern is that without some sort of curvature heuristic there might be cases where the difference between non-linearly placed glyphs is more severe.

mourner:
@yhahn great work! One question though — placement/collision detection are not affected by unskewing here, so in some cases, labels can appear slightly colliding or protruding beyond the end of a line, right? Do we want to account for that, or is that a subject to follow-up PRs?

yhahn:
> so in some cases, labels can appear slightly colliding or protruding beyond the end of a line, right?

@mourner yes, especially if the pitch angle is very high. Designers can adjust for this pretty well right now as a stopgap.

I think this is something we can address in a further iteration/concepting -- OTOH I think to account for pitch when calculating collision it'll require doing redoplacement on map x/y movement in addition to just rotation.

yhahn:
OP updated as this PR has evolved. Will hop next to updating native and then getting a full set of the rotate/pitch permutations in place in the render tests.

yhahn:
@lucaswoj I think this ready for your review!

I've also reviewed the current benchmarks and manually tested the debug page without seeing any noticeable difference in user experience. Let me know if you think we should take on adding some more systematic benchmarks for this work specifically.

lucaswoj:
Amazing work @yhahn! :shipit:

Issue #1231, opened 2015-06-01
Polygon text labels not centered when using symbol-placement: point


sillfishes:
Not sure if this is expected behavior or not, but when using "symbol-placement":"point", text labels appear at the bottom-right corner of polygons.  I was expecting this to place the label at the centroid of the polygon.  Is this an incorrect assumption?

![screen shot 2015-05-29 at 9 32 04 am](https://cloud.githubusercontent.com/assets/6751140/7916880/a0c40fe8-0854-11e5-815d-92c4ca983586.png)

Added here on stack overflow as well:

http://stackoverflow.com/questions/30531006/centering-text-label-in-mapbox-gl-js

ansis:
Labeling polygons isn't supported yet. We should implement it though!

animaux:
Huge problem for me too.

tmcw:
Would a decent first pass be a centroid?

To be clear, a first implementation of this is going to have some quirks: given that polygons are split into tiles and we can't identify repeated features cross-tile, large polygons will be labeled more than once.

animaux:
> Would a decent first pass be a centroid?

Thanks, definitely worth to try and check out the result!

Are polygons *always* split into tiles if they cross it’s border or is there a threshold of points that need to be outside the current tile?

animaux:
@tmcw any news on this so far?

lucaswoj:
@animaux This feature is not on our roadmap yet. Would your use case be met by using [`turf-centroid`](https://github.com/Turfjs/turf-centroid) to calculate the centroids yourself?

animaux:
@lucaswoj Thanks, that’s a pity.

I guess turf.js would do the trick but also add some more unnecessary complexity in the whole setup.

lyzidiamond:
@animaux if you want an HTTP endpoint to hit instead of including an additional library, you can send a POST with polygon GeoJSON to https://centroids.herokuapp.com/centroids and it'll return GeoJSON with centroids, all properties intact. Code here: https://github.com/lyzidiamond/centroids.

animaux:
Thanks @lyzidiamond, will have a look at it!

ansis:
@peterqliu 

[...]

PR #2678, opened 2016-06-06
#1231 Use polygon centroid for 'point' symbol placements

blanchg:
For Issue #1231. This is a first pass that uses a slightly modified centroid function from turf-centroid.js

It works great for simple polygons, but it doesn't handle holes or complex polygons.

For holes, maybe the solution would be to look at using the formula for [geometric decomposition](https://en.wikipedia.org/wiki/Centroid#By_geometric_decomposition)

For complex polygons (e.g. river polygons ) a more complex straight polygon tree could be used but I haven't found a nicie clean fast solution for that yet.

This is also my first PR for mapbox-gl-js, so please let me know if I have missed something.


js/util/util.js
```diff
@@ -474,3 +475,27 @@ exports.warnOnce = function(message) {
...
+        return [points[0].x, points[0].y];
+    }
+    // This tries to prevent overflow/underflow of the float
+    var xC = points[0].x, yC = points[0].y;
+    var xSum = 0, ySum = 0, len = 0;
+    points.forEach(function(coord) {
+        xSum += coord.x - xC;
+        ySum += coord.y - yC;
+        len++;
+    });
```
mourner: Please replace with a simple `for` loop for performance and simplicity.

mourner: Also, I think we should use polygon centroid algorithm like this https://en.wikipedia.org/wiki/Centroid#Centroid_of_polygon, which will choose a better center.

blanchg: Excellent thanks for pointing it out.  There was an error in the function on wikipedia that has now been corrected.  I have the additional changes but it causes tests to fail if we keep this change as symbol-placment: 'point'.  

The side effect is that lineStrings will use the centroid which I believe is incorrect.  I did ask if a symbol-placment: 'centroid' should be used in #1231.  This would require changes to the style spec as well, I may need some help/pointers to understand how that is also updated/released.

dpieri:
@blanchg thanks for starting this! Lack of this feature has been a huge pain point for us.

In my opinion, [Pole of Inaccessibility](https://en.wikipedia.org/wiki/Pole_of_inaccessibility) is the ideal way to position labels on polygons. I can't speak to the performance implications of doing it in real time compared to a centroid. If they are negligible then I hope you would consider using this method instead.

Here's a Turf [module](https://github.com/Turfjs/turf-polygon-reduce) that implements it, and some more [background](https://github.com/Turfjs/turf/issues/240).

blanchg:
That looks like a great solution, however that turf module uses turf-buffer which uses jsts to do the actual bufferring, that would increase the size of mapbox significantly.  I will look to see by how much if we only include the polygon offset/buffer code from jsts.

blanchg:
@dpieri I found a small library that performs the Pole Of Inaccessibility and extended it to support holes: https://github.com/blanchg/LabelPoint, however it appears to interfere with a lot of polygon draws so that on higher levels the polygons stop rendering completely and the +/- zoom buttons never get displayed.  Is there a way to find out why this is occuring (is there some sort of timeout maybe?) without going step by step through the code?

dpieri:
I'm not sure, maybe @mourner could point you in the right direction.

blanchg:
Thanks, I found that the simplification can produce zero area polygons with multiple points that was causing issues.

blanchg:
This is ready now.  It continues to use symbol-placment point.

If the ring is closed e.g. first and last point are the same and has an area != 0 then it chooses the Pole of Inaccessibility otherwise it falls back to the first point as it was before.  This works then for points, lines and polygons that have been simplified down to a point/line.

It adds about 1-2kb to the output, fps benchmarks look very similar.

I am positive there are things I can do better so please let me know!  Thanks

js/util/counter_clockwise.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+// http://bryceboe.com/2006/10/23/line-segment-intersection-algorithm/
+module.exports = function (a, b, c) {
+    return (c.y - a.y) * (b.x - a.x) > (b.y - a.y) * (c.x - a.x);
+};
```
mourner: I don't think we need to cut out small functions like this into a separate file. Maybe move to `util.js`? Or maybe we should have `geom_util.js` for all functions like this.

js/util/util.js
```diff
@@ -454,3 +455,40 @@ exports.warnOnce = function(message) {
...
+        p2 = ring[j];
+        mul = (p2.x * p1.y - p1.x * p2.y);
+        area += mul / 2;
+        cX += (p2.x + p1.x) * mul;
+        cY += (p2.y + p1.y) * mul;
+    }
+    var f = 6 * area;
+
+    return [cX / f, cY / f];
+};
```
mourner: If we use pole of inaccessibility algorithm, this one is no longer used, so should be removed, right?

mollymerp:
@blanchg thank you for contributing! 
could you please:
1. rebase your PR to squash your commits and remove commits from other PRs? you can use `git rebase -i origin/master` to do this.
2. add unit tests and/or [render tests](https://github.com/mapbox/mapbox-gl-test-suite)

thanks! 

👓  @lucaswoj @mourner @mollymerp for review

blanchg:
Thanks for the reviews @mollymerp and @mourner !

I managed to squish the changes and remove the unrelated ones thanks for pointing that out.  I added some unit tests for polygon_poi and I have a render test but I am uncertain how to proceed to get the render test included.  

I have a feature branch on a fork, I assume I do a PR on the test suite that should only be pulled in once this PR has been?

blanchg:
I found that building polygons in the render test suite data have outer rings that don't match the tile specification (2.1).  They have both positive and negative areas as outer rings.  I also found util/classify_rings that handles both cases.  Will update to use that.

blanchg:
Updated to use util/classify_rings which now handles polygons and holes of any winding order.  One raster test will fail until it is updated.  I still need some information on how to get the test-suite aligned with this PR.

mollymerp:
> I still need some information on how to get the test-suite aligned with this PR.

You can see information on how to update the tests [here](https://github.com/mapbox/mapbox-gl-test-suite#writing-new-tests). Once your tests are satisfactory, make a PR in mapbox/mapbox-gl-test-suite repo and then you can update the sha for that commit in the package.json for this PR. 

I don't see the raster tests failing, I just see the `debug/tile` test failing. Any idea why that might be happening?

@mourner I think you should have final review here.

blanchg:
@mollymerp Thanks for the information.  I have that PR submitted now.

The debug/tile test fails because one of the labels is in a closed polygon so it is now rendered at the center of that polygon, every other test doesn't render labels on closed polygons.

I have added the line style and zoomed to the label that moved slightly in the debug/tile test.
image

blanchg:
This is now up to date with the head, squashed, unit tests, there is a render test added and updated the failing debug test.  PR in the test-suite and sha pointing to it in package.json.

I think finally it is ready to merge :smiley: :tada: :balloon:

mourner:
Fantastic! I'll do a final review tomorrow and merge if everything's good.

dpieri:
Any idea when this will make it into a point release?

mourner:
This got a bit sidetracked but hopefully we will merge this week, followed by a point release next week.

blanche:
i think you ment @blanchg ;)

blanchg:
@mourner Thanks for looking at this again!  Updated and squashed.

js/util/polygon_poi.js
```diff
@@ -0,0 +1,150 @@
...
+        if (yMax < p.y) yMax = p.y;
+    }
+
+    var allpoints = holes ? points.concat.apply(points, holes) : points;
+    var lp = poleScan(xMin, yMin, xMax, yMax, points, holes, allpoints);
+    if (precision > 0) {
+        var r = ((xMax - xMin) * (yMax - yMin));
+        var dx, dy;
+        while (r > precision) {
+            lp = poleScan(xMin, yMin, xMax, yMax, points, holes, allpoints);
```
mourner: If precision is > 0, `poleScan` (the expensive part of the algorithm) is always called twice with the same arguments — can we eliminate the redundant call?

js/data/bucket/symbol_bucket.js
```diff
@@ -332,6 +338,39 @@ SymbolBucket.prototype.addFeature = function(lines, shapedText, shapedIcon, feat
...
+        }
+    }
+
+    if (isPolygon) {
+        var holes = null;
+        if (polygon.length > 1) {
+            holes = polygon.slice(1);
+        }
+        var poi = poleOfInaccessibility(line, holes, 1);
+        anchors = [ new Anchor(poi.x, poi.y, 0) ];
```
mourner: Anchor coordinates should probably be rounded, like polygon coordinates.

js/util/polygon_poi.js
```diff
@@ -0,0 +1,150 @@
...
+    } else if (param > 1) {
+        xx = x2;
+        yy = y2;
+    } else {
+        xx = x1 + param * C;
+        yy = y1 + param * D;
+    }
+    var dx = x - xx;
+    var dy = y - yy;
+    // return Math.sqrt(dx * dx + dy * dy);
```
mourner: No need for this comment.

js/util/polygon_poi.js
```diff
@@ -0,0 +1,150 @@
...
+    var c = false;
+    for (var i = -1, l = points.length, j = l - 1; ++i < l; j = i) {
+        var pi = points[i];
+        var pj = points[j];
+        if (((pi.y <= y && y < pj.y) || (pj.y <= y && y < pi.y)) &&
+            (x < (pj.x - pi.x) * (y - pi.y) / (pj.y - pi.y) + pi.x))
+            c = !c;
+    }
+    return c;
+}
```
mourner: Many of the functions introduced here (as well as in `util.js`) are also present in https://github.com/mapbox/mapbox-gl-js/blob/62ac39cf7f1f55a9a82b44e67fb5d71271e97f8b/js/util/intersection_tests.js#L102-L164. Can we eliminate the duplication?

js/util/polygon_poi.js
```diff
@@ -0,0 +1,150 @@
...
+
+    var allpoints = holes ? points.concat.apply(points, holes) : points;
+    var lp = poleScan(xMin, yMin, xMax, yMax, points, holes, allpoints);
+    if (precision > 0) {
+        var r = ((xMax - xMin) * (yMax - yMin));
+        var dx, dy;
+        while (r > precision) {
+            lp = poleScan(xMin, yMin, xMax, yMax, points, holes, allpoints);
+            dx = (xMax - xMin) / 24;
+            dy = (yMax - yMin) / 24;
```
mourner: Let's move out the magic number `24` into a constant at the top of the file, e.g. `GRID_SIZE`.

js/util/util.js
```diff
@@ -454,3 +454,60 @@ exports.warnOnce = function(message) {
...
+    // If it is 2 points that are the same then it is a point
+    // If it is 3 points with start and end the same then it is a line
+    if (points.length < 4)
+        return false;
+
+    var p1 = points[0];
+    var p2 = points[points.length - 1];
+
+    if (Math.abs(p1.x - p2.x) > 0.001 ||
+        Math.abs(p1.y - p2.y) > 0.001) {
```
mourner: Why do we need `0.001` thresholds here if coordinates are expected to be integer?

js/util/polygon_poi.js
```diff
@@ -0,0 +1,150 @@
...
+    yMin = yMax = p.y;
+    for (var i = 1; i < points.length; i++) {
+        p = points[i];
+        if (xMin > p.x) xMin = p.x;
+        if (xMax < p.x) xMax = p.x;
+        if (yMin > p.y) yMin = p.y;
+        if (yMax < p.y) yMax = p.y;
+    }
+
+    var allpoints = holes ? points.concat.apply(points, holes) : points;
```
mourner: By reorganizing the code a bit, I think we can fully avoid doing `concat` (which copies all the data), gaining a bit of performance.

mourner:
Aside from the comments above, I'm also slightly worried about performance — the algorithm used is pretty expensive. E.g. on a sample 4.6k water polygon, it runs for 350ms. Maybe that's fine for an initial implementation, but we should definitely think about a potential follow-up with a faster algorithm.

blanchg:
@mourner Thanks for the review, I will address those points.

Can I get that water polygon from somewhere to see if I can improve the performance?  When it is tiled on the map the number of points doesn't seem to get too large though I admit to not having tested on a very complex coastline/waterway.

mourner:
@blanchg yeah, unfortunately sometimes polygons can be quite big, especially water, landcover and other natural sources. You can use some of the fixtures from https://github.com/mapbox/earcut/tree/master/test/fixtures — the one I tested is the first ring of `water-huge.json`.

mourner:
@blanchg after spending some time on this, I have designed an algorithm (inspired by this one) that is both faster and _guarantees_ global optimum within given precision (unlike the one from the POI paper). On a few test polygons I tried it on, it runs 6-15x faster. I'll push it to NPM as a module and then we can switch to it in this PR.

mourner:
Pushed my algorithm to https://github.com/mapbox/polylabel repo. It looks very similar to this one, so the PR could be adapted easily. 

The library I intend to publish handles polygons in GeoJSON format though, so instead of using it as an NPM module, it probably makes sense to bring the code here, replacing `[x, y]` format with the `point-geometry` used in GL JS.

blanchg:
That looks very promising! Did you want me to pull that into mapbox or leave it for you?  It might be a few days before I get a chance to work on it.

blanchg:
@mourner if this build passes then using your new method has worked and it is ready for another review.

js/util/ajax.js
```diff
@@ -41,6 +41,7 @@ exports.getJSON = function(url, callback) {
             cache[url] = data;
             callback(null, data);
         } else {
+            console.log("Error", url, response);
```
mourner: accidental console.log

mourner:
Awesome! Note that additional advantage of this algorithm is that shapes like rectangles get their center as the label point. 

Can you please run the [benchmarks](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/README.md#running-benchmarks) now to make sure we didn't regress anywhere?

js/data/bucket/symbol_bucket.js
```diff
@@ -333,6 +340,45 @@ SymbolBucket.prototype.addFeature = function(lines, shapedText, shapedIcon, feat
...
+        isPolygon = line === polygon[0];
+        if (!isPolygon) {
+            for (var j = 1; !isHole && j < polygon.length; j++) {
+                isHole = line === polygon[i];
+            }
+        }
+    }
+
+    if (isPolygon) {
+        var poi = poleOfInaccessibility(polygon, 1);
```
mourner: Do you think we need the `1` precision?  It could probably be bumped to something like 16 (which is just 2 pixels).

js/util/polygon_poi.js
```diff
@@ -0,0 +1,134 @@
+'use strict';
+var Queue = require('tinyqueue');
+var Point = require('point-geometry');
+var distToSegmentSquared = require('./intersection_tests').distToSegmentSquared;
+var isClosedPolygon = require('./util').isClosedPolygon;
+
+/**
+ * Finds an approximation of a polygon's Pole Of Inaccessibiliy https://en.wikipedia.org/wiki/Pole_of_inaccessibility
+ * Note that for perfect rectangles the pole found ISN'T in the center (though the center is a valid POI)
```
mourner: The note about perfect rectangles is no longer valid since we get centroid as the first best guess and only choose a different point if it's better in terms of distance. 

I think it's also worth linking to the original repo (https://github.com/mapbox/polylabel) in the comments.

blanchg:
@mourner Thanks very much for the review I have updated as needed.

# Benchmarks Master/HEAD

**buffer:** 1,435 ms
**fps:** 60 fps
**frame-duration:** 8.4 ms, 3% > 16ms
**query-point:** 1.12 ms
**query-box:** 72.96 ms
**geojson-setdata-small:** 14 ms
**geojson-setdata-large:** 206 ms

# Benchmarks this PR

**buffer:** 1,320 ms
**fps:** 59 fps
**frame-duration:** 8.7 ms, 3% > 16ms
**query-point:** 1.03 ms
**query-box:** 76.07 ms
**geojson-setdata-small:** 13 ms
**geojson-setdata-large:** 210 ms

js/data/bucket/symbol_bucket.js
```diff
@@ -333,6 +340,46 @@ SymbolBucket.prototype.addFeature = function(lines, shapedText, shapedIcon, feat
...
+            for (var j = 1; !isHole && j < polygon.length; j++) {
+                isHole = line === polygon[i];
+            }
+        }
+    }
+
+    if (isPolygon) {
+        // 16 here represents 2 pixels
+        var poi = poleOfInaccessibility(polygon, 16);
+        if (poi === null) {
```
mourner: I don't think this can ever be null. The algorithm always returns some point.

js/data/bucket/symbol_bucket.js
```diff
@@ -333,6 +340,46 @@ SymbolBucket.prototype.addFeature = function(lines, shapedText, shapedIcon, feat
...
+    var isHole = false;
+    for (var i = 0; !isPolygon && !isHole && i < polygons.length; i++) {
+        polygon = polygons[i];
+        isPolygon = line === polygon[0];
+        if (!isPolygon) {
+            for (var j = 1; !isHole && j < polygon.length; j++) {
+                isHole = line === polygon[i];
+            }
+        }
+    }
```
mourner: It looks like for each polygon ring, you go through all other rings doing equality comparisons, which is O(n^2). However it does not seem to be necessary — you could instead add symbol instances as you loop through the rings above. This would need some code reorganization above.

js/util/polygon_poi.js
```diff
@@ -0,0 +1,134 @@
...
+ * @param {bool} [debug=false] Print some statistics to the console during execution
+ *
+ * @returns {Point} Pole of Inaccessibiliy.
+ */
+module.exports = function (polygons, precision, debug) {
+
+    if (polygons.length === 0) {
+        return null;
+    }
+    if (!isClosedPolygon(polygons[0])) {
```
mourner: This check should probably live outside of the poi function.

js/util/polygon_poi.js
```diff
@@ -0,0 +1,134 @@
...
+ * Finds an approximation of a polygon's Pole Of Inaccessibiliy https://en.wikipedia.org/wiki/Pole_of_inaccessibility
+ * This is a copy of http://github.com/mapbox/polylabel adapted to use Points
+ *
+ * @param {Array<Array<Point>>} polygons first item in array is the outer ring polygon followed optionally by the set of holes, should be the first element of the result of util/classify_rings
+ * @param {number} [precision=1] Specified in input coordinate units. If 0 returns after first run, if > 0 repeatedly narrows the search space until the radius of the area searched for the best pole is less than precision
+ * @param {bool} [debug=false] Print some statistics to the console during execution
+ *
+ * @returns {Point} Pole of Inaccessibiliy.
+ */
+module.exports = function (polygons, precision, debug) {
```
mourner: `polygons` is a bit confusing here because it really means rings of a single polygon in this case. Perhaps `rings` would be better?

blanchg:
@mourner all excellent points, hopefully I addressed them.

# Benchmarks Master/HEAD

**buffer:** 1,435 ms
**fps:** 60 fps
**frame-duration:** 8.4 ms, 3% > 16ms
**query-point:** 1.12 ms
**query-box:** 72.96 ms
**geojson-setdata-small:** 14 ms
**geojson-setdata-large:** 206 ms
# Benchmarks PR

**buffer:** 1,350 ms
**fps:** 59 fps
**frame-duration:** 8.3 ms, 4% > 16ms
**query-point:** 1.00 ms
**query-box:** 72.39 ms
**geojson-setdata-small:** 15 ms
**geojson-setdata-large:** 204 ms

mourner:
This looks pretty good now, thanks for addressing the feedback! I'll let @lucaswoj take a quick final look just in case, and aim to merge soon.

js/data/bucket/symbol_bucket.js
```diff
@@ -14,6 +14,8 @@ var clipLine = require('../../symbol/clip_line');
 var util = require('../../util/util');
 var loadGeometry = require('../load_geometry');
 var CollisionFeature = require('../../symbol/collision_feature');
+var poleOfInaccessibility = require('../../util/polygon_poi');
```
lucaswoj: An (admittedly unwritten) convention in GL JS is that file names match the names of the classes / functions they export. Can we rename `polygon_poi` -> `find_pole_of_inaccessibility` or similar?

lucaswoj:
This looks good to me! I'm excited to 🚢 .

blanchg:
@mourner @mollymerp @lucaswoj Thank you all for your help, I think this is finally ready :D :+1:

mourner:
Set up a JSBin debug page to demonstrate how it works in practice, using the choropleth from examples as a base but adding a population label: http://output.jsbin.com/tajiwociha

Generally it's pretty cool, but there are a few issues that are worth noting, although they don't necessarily warrant an immediate fix, and not all are possible to fix (e.g. because of tiling or the the way algorithm works).

 - You have to use `"symbol-avoid-edges": true`, otherwise there will be collisions across tile boundaries because the neighboring points don't match (they represent different parts cut off from a polygon by tiling).
- Due to buffers around polygons clipped by tile boundaries, labels in such polygons are placed closer to the boundary than needed; the polygons should probably be clipped before running the algorithm for more correct results.
- Polygons split across several tiles will of course have several labels.
- Label positions sometimes "jump" between zoom levels, because the algorithm can pick a different point in a higher-precision/bigger coords version of the same polygon; this is exaggerated on this particular example because of many rectangle-like shapes.

@samanpwbb can you play with the sample and tell us what you think? Should we go ahead with the feature in its current state?

samanpwbb:
It's much better than what we have now, so I believe that it is worth shipping in it's current state.

> You have to use "symbol-avoid-edges": true

That's fine – i know this is a hard / impossible problem to fix. Mapnik was similar.

> Due to buffers around polygons clipped by tile boundaries, labels in such polygons are placed closer to the boundary than needed

This problem is noticeable – definitely worth following up on: 
screen shot 2016-08-18 at 10 35 24 am

> Polygons split across several tiles will of course have several labels.

That's fine, Mapnik had the same issue. vector tiles ¯\_(ツ)_/¯

> Label positions sometimes "jump" between zoom level

Not a big deal.

I'm also seeing an issue with labels on features I expect to be smaller parts of multipolygons. Is this just a data issue, or is mapbox gl unable to distinguish multipolygons?

screen shot 2016-08-18 at 10 29 06 am

mourner:
> I'm also seeing an issue with labels on features I expect to be smaller parts of multipolygons. Is this just a data issue, or is mapbox gl unable to distinguish multipolygons?

It currently just adds a label for every polygon in a multipolygon. We could alter the algororithm to filter out labels for polygon where anchor point has a distance to the polygon below a certain threshold, but this would probably require a new obscure style-spec property, or have a very good default. Another option is to always only add one label to a multipolygon, choosing an anchor in one of the polygons that fits the best.

mourner:
It's also worth noting that the performance cost on the worker side is pretty high — on this example but without a basemap (just leaving the population map alone), placing polygon anchors takes ~18% of the worker time. If I switch from 2px to 4px precision, it takes ~13%. It may be acceptable though, since this is an opt-in feature, and doesn't affect our styles — only custom visualizations where you explicitly want to use auto polygon anchors.

samanpwbb:
> Another option is to always only add one label to a multipolygon, choosing an anchor in one of the polygons that fits the best.

I can see cases where both would be desireable, but _most of the time_ a multipolygon should only get a single label. you'd almost never want to include more than one label on a choropleth data viz (and this could even result in a map that 'lies' – making it look like a place has double the value that it actually has). 

On the other hand, a cartographer may want to place a tree icon in the center of each piece of a park multipolygon. 

I vote to enforce 1 label per multipolygon, leaving open the option of adding a new property to give users control in the future.

mourner:
@samanpwbb makes sense! Regarding polygons that have labels placed close to boundaries, I referred to a different case:

image

Here, the label is not further up because it calculates anchor based on the polygon geometry with a buffer, including an "invisible" part on the bottom. 

And regarding the case below, — it's just a property of the algorithm — in this case a label in the center may have a few pixels smaller distance to the polygon, and considered worse than the one picked; to fix this, we would have to modify the algorithm to somehow prefer points closer to centroid even if they have smaller distance, but this is not trivial.

image

samanpwbb:
> We could alter the algororithm to filter out labels for polygon where anchor point has a distance to the polygon below a certain threshold

After thinking more about the problem and various edge cases, this is my favorite option for handling multi-poly labelling. It wouldn't add a new property and would ideally look good in most cases. I also recognize that this is the hard approach. Again, like I said in the original post, my take on this PR is that it's a major improvement over what we have now no matter what, and I think we should merge something imperfect sooner (and ticket follow up work) rather than wait until it's perfect.

mourner:
> It wouldn't add a new property and would ideally look good in most cases.

How do we pick the threshold then? E.g. something arbitrary like 2 "ems" assuming em = 24px?

samanpwbb:
> How do we pick the threshold then? E.g. something arbitrary like 2 "ems" assuming em = 24px?

Yep something like that sounds good. See how it feels. 

--- 

Let me bring up another idea. I don't know if it's at all technically feasible to incorporate into your label placement algorithm. Archipelago-type features are really common in the world. Louisiana is mostly a single mass, but has an archipelago on it's southern tip. But what about a Phillipines?

Is there a way to incorporate multipolygons into the label placement algorithm in such a way that the label still ends up getting placed inside a shape, so we are able to handle archipelagos well?

blanchg:
I found if you ignore polygons that have a cellSize < precision * 2 it seems to handle the US quite well and fixes the label showing up on the islands off Louisiana and Texas when you are zoomed out.  This could cause issues for archipelagos though as it won't show any labels until one of the islands is a few pixels big.  Seems to speed up the rendering too.

http://jsbin.com/lopilikohe/1

`
var cellSize = Math.min(width, height);

if (cellSize < precision * 10) {
    return null;
}
`

fedex1:
Here's an example that I'm using to test.

Before centroid change: https://city.tidalforce.org/nycschoolallocation_usm.html
See the "white label"  "District 16"
screen shot 2016-08-18 at 11 08 43 pm

With the centroid change (I cloned mapbox-gl-js and switched to the pull request branch and built)
screen shot 2016-08-18 at 11 07 45 pm

It is definitely an improvement.

A few questions:

- Is there supposed to be multiple labels for a small polygon?  this may be a result of my data so I apologize if this is a data issue.

- I noticed that the copyright style change
```

 map.style.sources['mapbox://cityregister.nyc_school_allocation_titlei'].attribution = [
                '&copy;BrooklynMarathon',
                '<a href="http://schools.nyc.gov/offices/d_chanc_oper/budget/dbor/allocationmemo/fy16_17/FY17_PDF/sam08_T2.xls">City of New York, Department of Education</a>'
            ].join(' | ');

```
no longer seems to work.

as it did in

```
<script src='https://api.tiles.mapbox.com/mapbox-gl-js/v0.19.1/mapbox-gl.js'></script>
<link href='https://api.mapbox.com/mapbox-gl-js/v0.19.1/mapbox-gl.css' rel='stylesheet' />


```

blanchg:
@fedex1 I don't think this PR has any changes that impact the copyright maybe the others can comment on this
Those multiple labels are likely due to the tiles if you turn on the tile debug lines it should highlight this.

:+1: for giving it a go and demonstrating feedback

fedex1:
@blanchg Yes that is correct.  Thanks for the reply!

with debug lines on. https://city.tidalforce.org/nycschoolallocation_usm_centroid.html

screen shot 2016-08-19 at 12 43 01 am

lucaswoj:
Rebased and squashed 👉 https://github.com/mapbox/mapbox-gl-js/pull/3038

anna-geo:
Could you please explain if it is possible now to show only one text label in polygon center even if it`s splited into several tiles? I would like to avoid situation like on the last image of @fedex1 where label "District 16" is repeated several times inside one district.

mourner:
@anna-geo currently not possible unfortunately, and a very hard issue to address.

lucaswoj:
@anna-geo I suggest that you do that processing during tile generation time and produce point geometries where the label should go.

anna-geo:
@mourner @lucaswoj Thank you for reply! Will be waiting for this feature.

PR #1347, opened 2015-06-29
combine color and opacity in shaders

ansis:
Fixes #1331 

@lucaswoj can you do a quick review?

Some render tests are now failing:
- background-opacity color
- fill-color opacity
- fill-opacity function
- line-pattern opacity

All of them look identical but have tiny differences in every pixel. Multiplying opacity by 255 in js and the shader probably have slightly different results. @kkaefer does that sound right, or do you think there's a small bug somewhere?

PR #2686, opened 2016-06-07
Remove util.premultiply

lucaswoj:
GL JS had the notion of "premultiplying" colors just before rendering them, transforming rgba arrays into the color format expected by GL (all components between 0 and 1, all components multiplied by the alpha value).

This PR removes the premultiply step, opting instead to format the colors in the format expected by GL end-to-end in order to

 - reduce the complexity of GL JS
 - get minor perf gains in regular rendering
 - get big perf gains in data-driven styling rendering

### Frame Duration Benchmark

**this branch** 9.9 ms per frame. 4% of frames took longer than 16ms.
**master** 10.3 ms per frame. 5% of frames took longer than 16ms.

cc @mollymerp @jfirebaugh @mourner

### DIFF
--- a/js/style/parse_color.js
+++ b/js/style/parse_color.js
@@ -1,46 +1,40 @@
 'use strict';
 
-var parseCSSColor = require('csscolorparser').parseCSSColor;
+var parseColorString = require('csscolorparser').parseCSSColor;
 var util = require('../util/util');
+var StyleFunction = require('./style_function');
 
-var colorCache = {};
+var cache = {};
 
-function parseColor(input) {
+module.exports = function parseColor(input) {
 
-    if (colorCache[input]) {
-        return colorCache[input];
+    if (StyleFunction.isFunctionDefinition(input)) {
 
-    // RGBA array
-    } else if (Array.isArray(input)) {
-        return input;
-
-    // GL function
-    } else if (input && input.stops) {
         return util.extend({}, input, {
-            stops: input.stops.map(parseFunctionStopColor)
+            stops: input.stops.map(function(stop) {
+                return [stop[0], parseColor(stop[1])];
+            })
         });
 
-    // Color string
     } else if (typeof input === 'string') {
-        var parsedColor = parseCSSColor(input);
-        if (!parsedColor) { throw new Error('Invalid color ' + input); }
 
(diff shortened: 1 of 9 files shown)


mollymerp:
LGTM – seems like this simplifies the code base and does not have any perf drawbacks. I ran the frame duration benchmarks locally and the results were almost identical: both had 5.8ms / frame and 0% taking longer than 16 ms. 

🚢

js/style/parse_color.js
```diff
@@ -1,46 +1,40 @@
...
-    } else if (Array.isArray(input)) {
-        return input;
-
-    // GL function
-    } else if (input && input.stops) {
         return util.extend({}, input, {
-            stops: input.stops.map(parseFunctionStopColor)
+            stops: input.stops.map(function(stop) {
+                return [stop[0], parseColor(stop[1])];
+            })
```
mourner: Any reason this was inlined? It's faster to move it outside rather than create an unnecessary closure on every call.

mourner:
@ansis wasn't the original reason to premultiply in the final stage something to do with interpolating between two semi-transparent colors + opacity? I have a feeling this breaks something in a subtle way not catched by the test suite...

js/style/parse_color.js
lucaswoj: Inlining this function improves readability. This code is not hot enough to warrant sacrificing readability for performance. For example, this code path only runs 33 times on our debug page which renders mapbox streets.

lucaswoj:
You might be thinking of the behaviour that was removed in https://github.com/mapbox/mapbox-gl-js/pull/1347. 

It's possible that there is an edge case that occurs in transitions between semi-transparent colors but everything looked 👍 in my testing.

js/style/parse_color.js
mourner: I don't see how it improves readability — the previous way was more readable for me.

mourner:
@lucaswoj yeah, that might be it — shouldn't be an issue anymore probably.

lucaswoj:
I'll leave this PR open for a few days to give @ansis a chance to 🎐 in.

mourner:
@lucaswoj note that you've already merged it :)

PR #2688, opened 2016-06-07
Avoid unnecessary steps in npm install 

lucaswoj:
Running `npm install` for GL JS takes 33s.

This PR moves some build steps from `package.json` into a new `Makefile`. In particular, it moves build steps that result in creating a file on disk. `make` has some built-in logic that will allow us to avoid generating these files if none of their dependencies have changed.

## Wins

 - **`npm install` time goes from 33s to 8s** (in the best case)
 - `package.json` becomes cleaner
 - we gain the ability to generate more complicated build artifacts in the future, should we decide to

cc @tmcw @mollymerp @mourner @jfirebaugh


jfirebaugh:
Use `npm update` instead, so prepublish doesn't get triggered? I'm skeptical of build system changes. [In iD, the exact opposite change has been proposed.](https://github.com/openstreetmap/iD/pull/3104)

lucaswoj:
`npm update` is not the same as `npm install` semantically. `npm update` also tends to be much slower than `npm install` (several minutes vs several seconds). 

Thanks for linking to the iD issue. I did not consider the implications of requiring `make` for Windows developers. 

Based on your feedback, I have adjusted my approach a little:

 - I have removed the `Makefile`
 - I only build `dist` artifacts when they are needed (during tests, releases, and when generating docs)
 - I only build docs when they are needed (during releases)

jfirebaugh:
That sounds good. Are you intending that `npm run build-dev` and `npm run build-min` will be run manually prior to `npm publish`, or that the published package will no longer contain the `dist` artifacts? I think the latter would be fine, just wanted to confirm.

lucaswoj:
I intend for the published package to no longer contain the `dist` artifacts.

mourner:
Looks good to me!

lucaswoj:
🚢ing and keeping an eye out for doc generation / deploy weirdness.

PR #2689, opened 2016-06-07
Fix for issue #2680, root cause map.options is no longer available

### LINKED ISSUE #2680: Fix exception thrown by "drag rotate" handler
per https://github.com/mapbox/mapbox-gl-js/commit/4dd6786282e7d3e9e7b2a7fd866c3637d9e88737#commitcomment-17747349

blanchg:
Now pass all the mapOptions to the constructor of each handler to fix #2680

### DIFF
--- a/js/ui/bind_handlers.js
+++ b/js/ui/bind_handlers.js
@@ -13,15 +13,15 @@ var handlers = {
     touchZoomRotate: require('./handler/touch_zoom_rotate')
 };
 
-module.exports = function bindHandlers(map, options) {
+module.exports = function bindHandlers(map, bindOptions, mapOptions) {
     var el = map.getCanvasContainer();
     var contextMenuEvent = null;
     var startPos = null;
     var tapped = null;
 
     for (var name in handlers) {
-        map[name] = new handlers[name](map);
-        if (options[name]) {
+        map[name] = new handlers[name](map, mapOptions);
+        if (bindOptions[name]) {
             map[name].enable();
         }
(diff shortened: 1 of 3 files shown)


js/ui/handler/drag_rotate.js
```diff
@@ -17,9 +17,10 @@ var inertiaLinearity = 0.25,
  * dragging the cursor while holding the right mouse button or the `ctrl` key.
  * @class DragRotateHandler
  */
-function DragRotateHandler(map) {
+function DragRotateHandler(map, options) {
     this._map = map;
     this._el = map.getCanvasContainer();
+    this._options = options;
```
mourner: Better to explicitly save `this._bearingSnap` here instead of the options object.

mourner:
Nice, looks good to me other than the minor comment above.

js/ui/bind_handlers.js
```diff
@@ -13,15 +13,15 @@ var handlers = {
     touchZoomRotate: require('./handler/touch_zoom_rotate')
 };
 
-module.exports = function bindHandlers(map, options) {
+module.exports = function bindHandlers(map, bindOptions, mapOptions) {
```
lucaswoj: Why the second `mapOptions` object? Could the `bearingSnap` property go on the original `options`/`bindOptions` object?

blanchg:
It could, but would require modification for every option that gets passed to every handler (only bearingSnap right now)

PR #2702, opened 2016-06-09
Preserve fresh=true query param

### LINKED ISSUE #2699: Support query parameters in "mapbox://" urls
To allow for caching on immutable settings but no caching on Studio side:

From @scothis: 
There are a couple different ways to approach
1. introduce a “fresh" mode into mapbox-gl-js where all resources are fetched with `fresh=true`
2. allow query params to be preserved from the `mapbox://` formed URL to the `http://` URL
3. provide a supported API where URLs can be transformed

cc @jakepruitt @emilymcafee

xrwang:
For caching headers, allow `fresh=true` to be preserved and passed onto the http:// URL. This applies to `mapbox://` only URLS. 

fixes #2699

cc @scothis @jakepruitt @emilymcafee


js/util/mapbox.js
```diff
@@ -2,78 +2,87 @@
 
 var config = require('./config');
 var browser = require('./browser');
+var url = require('url');
+var querystring = require('querystring');
 
-function normalizeURL(url, pathPrefix, accessToken) {
+function normalizeURL(parsedUrl, pathPrefix, accessToken) {
```
lucaswoj: Would this module be simpler if we parsed the url within `normalizeURL` instead of within its callers?

lucaswoj: I wouldn't mind parsing the same URL twice. This code isn't very hot.

test/js/util/mapbox.test.js
```diff
@@ -61,8 +63,8 @@ test("mapbox", function(t) {
     t.test('.normalizeGlyphsURL', function(t) {
         t.test('normalizes mapbox:// URLs', function(t) {
             t.equal(
-                mapbox.normalizeGlyphsURL('mapbox://fonts/boxmap/{fontstack}/{range}.pbf'),
-                'https://api.mapbox.com/fonts/v1/boxmap/{fontstack}/{range}.pbf?access_token=key'
+                mapbox.normalizeGlyphsURL('mapbox://fonts/boxmap/{fontstack}/{range}.pbf?fresh=true'),
+                'https://api.mapbox.com/fonts/v1/boxmap/{fontstack}/{range}.pbf?fresh=true&access_token=key'
             );
```
lucaswoj: Can we add new tests for the `?fresh=true` case and keep the old tests intact?

js/util/mapbox.js
scothis: I'm not sure it's helpful to modify `normalizeURL`, it's already able to handle URLs with query strings. We'd only need to make sure that when URLs are manipulated before being passed to `normalizeURL` propagate the query string.

lucaswoj:
fixes #2699

jfirebaugh:
If you do `map.setStyle("mapbox://styles/username/id?fresh=true")`, should the subsequent sprite, glyph, and tile requests that mapbox-gl-js makes also carry a `fresh=true` parameter?

scothis:
> If you do map.setStyle("mapbox://styles/username/id?fresh=true"), should the subsequent sprite, glyph, and tile requests that mapbox-gl-js makes also carry a fresh=true parameter?

The query string should be opaque to mapbox-gl-js. If the server handling the request thinks the tile, sprite or glyph URLs should contain the query param, it can add it before responding with the style.

xrwang:
tests passing, lmk how it looks! I'll squash down to one commit on merge.

js/util/mapbox.js
```diff
@@ -2,78 +2,102 @@
...
+    var formattedQuery = '';
+
+    if (Object.keys(urlObject.query).length !== 0) {
+        formattedQuery = '?' + querystring.stringify(urlObject.query);
+    }
+
+    var parsedURL = url.format(urlObject.protocol + '/' + urlObject.pathname + formattedQuery);
+
+    return normalizeURL(parsedURL, '/styles/v1/', accessToken);
 };
```
lucaswoj: If we're parsing the URL, I think we can make this code much simpler:

```js
module.exports.normalizeStyleURL = function(url, accessToken) {
    var parsedUrl = URL.parse(url, true);

    if (parsedUrl.protocol === 'mapbox:') {
        return normalizeURL(URL.format({
            protocol: 'mapbox',
            pathname: parsedURL.pathname,
            query: parsedURL.query 
        }, '/styles/v1/', accessToken)

    } else {
        return url;
    }
};
```

lucaswoj: Err... My code doesn't do anything. I'll take a closer look in the morning 🙂

js/util/mapbox.js
```diff
@@ -2,78 +2,102 @@
...
-    var draft = split[5] ? '/draft' : '';
-    return normalizeURL('mapbox://' + user + '/' + style + draft, '/styles/v1/', accessToken);
+    var urlObject = url.parse(styleURL, true);
+    var formattedQuery = '';
+
+    if (Object.keys(urlObject.query).length !== 0) {
+        formattedQuery = '?' + querystring.stringify(urlObject.query);
+    }
+
+    var parsedURL = url.format(urlObject.protocol + '/' + urlObject.pathname + formattedQuery);
```
lucaswoj: What's your intent using `url.format` here? As best I can tell, this method is supposed to take a single `Object` argument.

js/util/mapbox.js
```diff
@@ -2,78 +2,102 @@
 
 var config = require('./config');
 var browser = require('./browser');
+var url = require('url');
+var querystring = require('querystring');
```
lucaswoj: If we're going to include `url` and `querystring` in our browserify bundle, it would be wise to use them everywhere in this file instead of string manipulation primitives and regexes. Doing so would be a big :+1: for readability and robustness.

xrwang:
@lucaswoj :
- I got rid of `url.format` (remnant, you were right, it is for a urlObject).
- updated to use `url.parse` in as many places as possible. Currently not using `url.parse`  for `glyphsURL` and `normalizeURL`, as @scothis and I rediscovered the `{ }` special character parsing issue. ( { } turns into /%7B with url.parse)
LMK how it looks, thanks !

xrwang:
also noting here that adding in `url` and `querystring`, size diff on the bundle size:
master of dist/mapbox-gl.js 485kb
this branch, dist/mapbox-gl.js 499kb

after running `npm run build-min`

lucaswoj:
I just pushed a commit that refactors `mapbox.js` to increase code clarity and robustness by leaning on the functionality provided by the `url` module. Turning the code review back over to @xrwang and @scothis.

xrwang:
\o/ thanks @lucaswoj.  @scothis, I think this is ready to merge?

lucaswoj:
Hang on -- I'm seeing some rendering defects & error messages in the console when running this branch...

1ec5:
This appears to have been implemented independently for the native SDKs in mapbox/mapbox-gl-native#5554.

Issue #2616, opened 2016-05-24
GeoJSON source with null geometry


andrewharvey:
http://jsbin.com/busubabude/edit?html,output

as per that example, loading a GeoJSON source with geometry: null causes the error `TypeError: Cannot read property 'type' of null`.

As per the GeoJSON spec at http://geojson.org/geojson-spec.html#feature-objects "The value of the geometry member is a geometry object as defined above or a JSON null value."

I would expect mapbox-gl-js to silently ignore this feature and continue processing other features rather than throwing an error.

I've been trying to nail down the code causing this error, but I can't trace it back in Chrome dev tools.

tmcw:
Maybe we should create a geojson-spectrum testcase that contains all valid geojson forms, so we can identify this kind of issue faster.

emakarov:
+1 for this issue. I have also error with null geometry -
`TypeError: Cannot read property 'coordinates' of null`

tmcw:
geojson-vt has this issue already fixed, by @andrewharvey ! https://github.com/mapbox/geojson-vt/blob/master/src/convert.js#L27 

But distributions of mapbox-gl-js are apparently using an older version that doesn't include the early check.

andrewharvey:
@tmcw Yeah I fixed it in geojson-vt a while back, which makes it annoying as I spent hours trying to find the root cause in mapbox-gl-js but when the error is thrown the stacktrace in Chrome dev tools never let me identify the exact line where it was trying to read the property of null.

So moving forward it looks like you'll need to do a release of geojson-vt then update mapbox-gl-js to use the updated version?

/cc @mourner

mourner:
Should be fixed.

PR #2708, opened 2016-06-09
bump geojson-vt to 2.2.0

### LINKED ISSUE #2704: Multipolygons disappearing from map at certain zoom levels.
I have a TopoJSON map of europe which I'm loading into mapbox (0.19.1). I have an issue where one of my country layers (Spain) disappears at certain zoom levels. I think this could also be an earcut issue  (similar to https://github.com/mapbox/mapbox-gl-js/issues/2657) as the issue is not present in 0.18.0.

See the attached gif

choro

mourner:
Closes #2704.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "highlight.js": "9.3.0",
     "istanbul": "^0.4.2",
     "lodash": "^4.13.1",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#b7da488f1b8ff93ed11905a175a0abc4f3c8ee84",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#7f102a6d12a2ec8e692069e57d50a17d46eb63c8",
     "minifyify": "^7.0.1",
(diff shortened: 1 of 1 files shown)


mourner:
Needs updated GeoJSON fixtures in test-suite because of a different ring winding order. :(

lucaswoj:
I pinned master to 2.1.8 pending a fix to the GeoJSON test fixtures.

mourner:
Needs PRs to:

- [ ] mapbox-gl-test-suite (updated query tests for polygons)
- [ ] geojson-vt-cpp (port the change)
- [ ] mason (publish newer geojson-vt-cpp)
- [ ] mapbox-gl-native (newer geojson-vt-cpp version)

mourner:
@jfirebaugh suggested just rewinding input polygons in fixtures so that native has the same results without changing geojson-vt-cpp (which is not necessary), and maybe adding a non-matching winding order test that is disabled in native.

andrewharvey:
Also closes https://github.com/mapbox/mapbox-gl-js/issues/2616

PR #2709, opened 2016-06-09
Add mapbox-gl-supported plugin and update example 

danswick:
Adds [mapbox-gl-supported](https://github.com/mapbox/mapbox-gl-supported) to the plugins page and updates the existing GL supported example to use the plugin instead of Mapbox GL JS's built-in method. 

Still needs: 

- [ ] Comments and description updates for the `check-for-support` example 

cc @tristen


lucaswoj:
@danswick CI should pass if you rebase on `master`

Issue #2711, opened 2016-06-10
Fix error handling in "GlyphSource#getSimpleGlyphs"


lucaswoj:
This issue comes from a [`// TODO` comment in `glyph_source.js`](https://github.com/mapbox/mapbox-gl-js/commit/d6877af8cb462b52baa0e4ff2ecbaa6dd8592964#diff-6632eb587812a66dac71654cb214cb6dR56)

```
// TODO not be silent about errors
```

cc @ansis

jfirebaugh:
The `ajax.getArrayBuffer` call in GlyphSource is also ignoring errors.

Issue #2710, opened 2016-06-10
Skip holes in "feature index"


lucaswoj:
This issue comes from a [`// TODO` comment in `feature_index.js`](https://github.com/mapbox/mapbox-gl-js/commit/571a8e73e1bb1059989e497b80e611ee0b4edd90#diff-111b47da6c9354c2d061b299864108e8R56)

```
// TODO: skip holes when we start using vector tile spec 2.0
```

cc @ansis

lucaswoj:
Duplicate of https://github.com/mapbox/mapbox-gl-js/issues/2293

PR #2712, opened 2016-06-10
Disallow TODO comments in the code

lucaswoj:
It is bad practice to commit `// TODO` comments. These comments should either be addressed or turned into GitHub issues lest they be forgotten. 

This PR enables a lint rule that disallows `// TODO` and `// FIXME` comments and removes existing comments. Two existing comments were turned into issues (https://github.com/mapbox/mapbox-gl-js/issues/2711, https://github.com/mapbox/mapbox-gl-js/issues/2710) and the rest were dropped. 

cc @mourner @tmcw @jfirebaugh @mollymerp

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,15 +1,16 @@
 {
   "extends": "mourner",
   "rules": {
-    "object-curly-spacing": 0,
-    "key-spacing": 0,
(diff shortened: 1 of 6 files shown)


jfirebaugh:
👍

PR #2722, opened 2016-06-13
Windows testing

blanchg:
1. `#` doesn't terminate the windows command like it does on osx/nix, but `# && rem` does.  The side effect of this was the word test in the comments was passing the render/query tests to the standard test runner which was causing a timeout.
2. Added a `[.bat` that takes the bash `-f FILENAME` and performs the file exists test so that once the benchmark data has been downloaded all the npm scripts will run.
3. Updated the `CONTRIBUTING.md` to include all the instructions I followed to get mapbox development running under windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "lint": "eslint js test bench",
     "start-docs": "npm run build-min && npm run build-docs && jekyll serve -w",
     "start": "npm run download-bench-data && node server.js",
-    "test-suite": "node test/render.test.js && node test/query.test.js # update ci.sh if test invocation changes",
-    "test": "npm run lint && npm run build-min && tap --reporter dot test/js/*/*.js # update ci.sh if test invocation changes"
+    "test-suite": "node test/render.test.js && node test/query.test.js",
+    "test": "npm run lint && npm run build-min && tap --reporter dot test/js/*/*.js"
   }
 }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,6 +64,32 @@ cd mapbox-gl-js &&
 npm install
 ```
 
+### Windows
+
+Install [git](https://git-scm.com/), [node.js](https://nodejs.org/) (version 4 or greater)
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -91,7 +91,7 @@
     "lint": "eslint js test bench",
     "start-docs": "npm run build-min && npm run build-docs && jekyll serve -w",
     "start": "npm run download-bench-data && node server.js",
-    "test-suite": "node test/render.test.js && node test/query.test.js # update ci.sh if test invocation changes",
-    "test": "npm run lint && npm run build-min && tap --reporter dot test/js/*/*.js # update ci.sh if test invocation changes"
+    "test-suite": "node test/render.test.js && node test/query.test.js # && rem update ci.sh if test invocation changes",
+    "test": "npm run lint && npm run build-min && tap --reporter dot test/js/*/*.js # && rem update ci.sh if test invocation changes"
```
lucaswoj: I would prefer remove these comments rathern than complicate this file with cross-platform shell hacks.

[.bat
```diff
@@ -0,0 +1,16 @@
...
+if EXIST %2. (
+	echo exists %2
+	exit /b 0
+) ELSE (
+	echo doesn't exist %2
+	exit /b 1
+)
+goto end
+
+:end
```
lucaswoj: Likewise, I would prefer remove the automatic bench data downloading rather than complicate this file with cross-platform shell hacks.

blanchg:
:+1: have reverted the hacks and simplified the npm scripts to remove the comments
Moved downloading the bench data to server.js

blanchg:
I made an incorrect assumption about polygon winding order.  According to the tile spec outer rings are DEFINED as clockwise with positive area and inner rings are counter clockwise with a negative area.  However the raster test data has buildings with outer rings clockwise and counter clockwise.  I also found the classifyRings function that does what I need and handles both cases.  Please hold off merging, thanks.

lucaswoj:
Thanks @blanchg! I did made some minor edits for consistency and simplicity and pushed as https://github.com/mapbox/mapbox-gl-js/commit/f426fc077c1e74ed592b5bde160b1ece08cdbbcc

PR #2724, opened 2016-06-13
Update line-pattern shader code/tests

brunoabinader:
Updates:
- mapbox-gl-shaders: 09ee512cd59a8fb1a241c78833b7c8022bf4f263
- mapbox-gl-test-suite: e78f09dce98080a9a6fa436d11fdf6fc2f271d7a

Part of https://github.com/mapbox/mapbox-gl-shaders/issues/1.


package.json
```diff
@@ -58,7 +58,7 @@
     "istanbul": "^0.4.2",
     "lodash": "^4.13.1",
     "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#e78f09dce98080a9a6fa436d11fdf6fc2f271d7a",
-    "nyc": "^6.1.1",
+    "nyc": "6.4.0",
```
lucaswoj: Are you sure this was the root of the problem? I think the segfault is happening in `headless-gl`, not `nyc`.

mourner: This was about a different issue where one of nyc `6.5.0` subdeps was broken and erroring on Travis, not related to headless tests.

lucaswoj: 👍 Got it. Thanks for the clarification.

PR #2690, opened 2016-06-07
Optimize transform conversions

mourner:
Reuses matrix calculations more, making `locationPoint` and `pointLocation` 13-15x faster according to the benchmark in `bench/unit/bench-transform.js`. cc @lucaswoj

PR #2725, opened 2016-06-13
Add an Overlay class

### LINKED ISSUE #656: Add support for markers
I want to use this library in conjunction with the Mapbox service for a game I am currently developing. However, we need support for custom markers like the Google Maps JavaScript API provides. We need markers with custom images, bindable events and a similar simple API.

mourner:
Closes #656 by introducing a simple Overlay class for marker-like functionality.

- Initially I made Overlay a base class Popup inherits from, but this turned out to be not worth it — complicating the code a lot while they share almost nothing.
- Overlays lag behind a little on panning animations like keyboard pan and inertia, but I haven’t yet found a source of the issue. Panning manually is fine.
- I eliminated one of the bottlenecks with #2690, but performance is still bottlenecked by browsers setting CSS transforms on each element. ~200 markers work fine on my machine, but 500 have a noticeable lag on a full-screen map.

cc @lucaswoj @tristen @jfirebaugh

### DIFF
--- a/js/ui/overlay.js
+++ b/js/ui/overlay.js
@@ -1,0 +1,87 @@
+/* eslint-disable */
+'use strict';
+
+module.exports = Overlay;
+
+var DOM = require('../util/dom');
+var LngLat = require('../geo/lng_lat');
+
+/**
+ * Creates an overlay component
+ * @class Overlay
+ * @param {HTMLElement=} element DOM element to use as an overlay (creates a div element by default)
+ * @example
+ * var overlay = new mapboxgl.Overlay()
+ *   .setLngLat([30.5, 50.5])
+ *   .addTo(map);
+ */
+function Overlay(element) {
+    if (!element) {
+        element = DOM.create('div');
+    }
+    element.classList.add('mapboxgl-overlay');
+    this._el = element;
+    this._update = this._update.bind(this);
+}
+
+Overlay.prototype = /** @lends Overlay.prototype */{
+    /**
+     * Attaches the overlay to a map
+     * @param {Map} map
+     * @returns {Overlay} `this`
+     */
+    addTo: function(map) {
+        this.remove();
+        this._map = map;
+        map.getCanvasContainer().appendChild(this._el);
+        map.on('move', this._update);
+        this._update();
+        return this;
+    },
+
+    /**
+     * Removes the overlay from a map
+     * @example
+     * var overlay = new mapboxgl.Overlay().addTo(map);
+     * overlay.remove();
+     * @returns {Overlay} `this`
+     */
+    remove: function() {
+        if (this._map) {
+            this._map.off('move', this._update);
+            this._map = null;
+        }
+        var parent = this._el.parentNode;
+        if (parent) parent.removeChild(this._el);
+        return this;
+    },
+
+    /**
+     * Get the overlay's geographical location
+     * @returns {LngLat}
+     */
+    getLngLat: function() {
+        return this._lngLat;
+    },
+
+    /**
+     * Set the overlay's geographical position and move it.
+     * @param {LngLat} lnglat
+     * @returns {Popup} `this`
+     */
+    setLngLat: function(lnglat) {
+        this._lngLat = LngLat.convert(lnglat);
+        this._update();
+        return this;
+    },
+
+    getElement: function() {
+        return this._el;
+    },
+
+    _update: function() {
+        if (!this._map) return;
+        var pos = this._map.project(this._lngLat);
+        DOM.setTransform(this._el, 'translate(' + pos.x + 'px,' + pos.y + 'px)');
+    }
+};
--- a/debug/markers.html
+++ b/debug/markers.html
@@ -10,7 +10,7 @@
         body { margin: 0; padding: 0; }
         html, body, #map { height: 100%; }
 
-        .mapboxgl-overlay {
+        .mapboxgl-marker {
             width: 10px;
             height: 10px;
             background: red;
@@ -37,7 +37,7 @@ var map = window.map = new mapboxgl.Map({
     hash: true
 });
 
-function addRandomOverlay() {
+function addRandomMarker() {
     var bounds = map.getBounds();
     var s = bounds.getSouth();
     var n = bounds.getNorth();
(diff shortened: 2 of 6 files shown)


js/ui/overlay.js
```diff
@@ -0,0 +1,87 @@
...
+function Overlay(element) {
+    if (!element) {
+        element = DOM.create('div');
+    }
+    element.classList.add('mapboxgl-overlay');
+    this._el = element;
+    this._update = this._update.bind(this);
+}
+
+Overlay.prototype = /** @lends Overlay.prototype */{
```
jfirebaugh: `@lends` is not needed here.

lucaswoj:
"overlays" seem similar to "native annotations" in GL Native. Would it be accurate to call these "annotations" instead of "overlays"? Consistent terminology across platforms is desirable. cc @1ec5 

-----

> performance is still bottlenecked by browsers setting CSS transforms on each element.

Would performance improve if we put all overlays into a single container and apply the transforms to that container? 

-----

This is a good time to reflect on our API design and make breaking changes. I would love to see this API move in a declarative & newless direction. For example:

```js
var overlay = map.addOverlay({
  element: document.createElement(...),
  coordinate: [lng, lat]
});
```

lucaswoj:
In some cases I'm seeing the overlays move in an unexpected pattern (i.e. not synced with the map) when zooming in/out. https://bl.ocks.org/lucaswoj/ec3cf390dc483435764b02e64516a395

mourner:
@lucaswoj could it be because of my CSS changes that are not in the block?

mourner:
> Would performance improve if we put all overlays into a single container and apply the transforms to that container?

Technically not possible, since we need to keep overlays unscaled while they inherit any scale set on the parent.

Regarding the API — I prefer factories like `mapboxgl.overlay()` for the API, but in any case, I would break the API in a separate PR push that affects everything, and keep this PR adhere to existing API conventions so that it remains focused.

peterqliu:
@mourner do you know if CSS smartly skips style updates on elements known to be out of view? If not, can we do this, by bbox/rbush the subset of markers known to be within a viewport, and transform only those?

mourner:
@peterqliu depends on the browser probably, I wouldn't count on that. RBush will only help if most markers are outside the view though, and in that case performance is usually not a problem anyway. It would also make the API more complicated.

peterqliu:
>Would performance improve if we put all overlays into a single container and apply the transforms to that container?
>Technically not possible, since we need to keep overlays unscaled while they inherit any scale set on the parent.

Can we negate the scale change by applying a compensatory scale-down *per marker*, when scaling up the parent element? Yes, we'd be saving one transform (per-marker translation) by adding another (per-marker scaling), but the former requires a separate projection calculation per marker, while the second is a globally-identical scale factor.

lucaswoj:
☑️  We will consider breaking API changes in a follow-up PR (ideally before the next release)
☑️  I understand CSS transform situation and think this solution is good
💭  We're still waiting on a final naming decision (overlay vs annotation vs ?)

lucaswoj:
Per chat with @1ec5, it doesn't make sense to borrow GL Native terminology here. I would still like to consider a couple other naming choices:

 - `Overlay` - existing name
 - `Marker` - this name has pretty intense name recognition, though doesn't emphasize that this is only one of several ways to implement markers 
 - `DOMMarker` - this name benefits from name recognition and emphasizes that this is just one type of marker
 - ??

yhahn:
My only thought here is that without knowing better I would get my hopes up that `Overlay` would support lines and polygons as well :sweat_smile:.

The `Marker` and `DOMMarker` as names definitely manage expectations better for me.

mourner:
> Can we negate the scale change by applying a compensatory scale-down per marker, when scaling up the parent element? Yes, we'd be saving one transform (per-marker translation) by adding another (per-marker scaling), but the former requires a separate projection calculation per marker, while the second is a globally-identical scale factor.

It gets more complicated when you consider rotation and pinch. Also, projection now takes negligible time after #2690, so the two ways would be roughly equivalent performance-wise.

1ec5:
> "overlays" seem similar to "native annotations" in GL Native. Would it be accurate to call these "annotations" instead of "overlays"? Consistent terminology across platforms is desirable.

In the gl-native core code, “annotations” are point or shape geometries without any associated popup-related content. This is an implementation detail completely hidden from native SDK consumers.

In the native SDKs, “annotations” (iOS, macOS) or “markers” (Android) are model objects that combine point or shape geometries with popup-related content (title, subtitle, tooltip). Shape annotations are also (primarily?) known as “overlays”. The “annotation view” (iOS) or “marker view adapter” (Android) is the equivalent of a DOM element used to display the annotation.

Currently, there isn’t an independent API for popups as there is in GL JS. “Callouts” (iOS), “popovers” (macOS), or “InfoWindows” (Android) are part of the behavior of an annotation or marker. The iOS and macOS SDKs rely heavily on the delegate pattern to populate callouts or popovers. This may change in the future with mapbox/mapbox-gl-native#4392. Meanwhile, in the Android SDK, a marker is constructed using the builder pattern, including properties for InfoWindow content.

These SDK terms originate from MapKit (iOS, macOS) and the Google Maps SDK (Android). The gl-native core terminology is loosely based on the iOS SDK terminology. Given this history and fragmented terminology, I think GL JS should choose the most logical terms for its own situation rather than attempting to match the native libraries.

I agree with all the points in https://github.com/mapbox/mapbox-gl-js/pull/2725#issuecomment-225724356. `Marker`, `DOMMarker`, or `ElementMarker` would all be preferable to `Overlay`.

antonhalim:
Hi, @lucaswoj, Is this feature is ready on current mapbox-gl-js, I am trying to instantiate the overlay by `new mapboxgl.Overlay()` but I m getting an uncaught TypeError: mapboxgl.Overlay is not a constructor. Thanks!

1ec5:
Since this PR landed, the `Overlay` class was renamed to `ElementMarker` (#2804) then renamed again to simply `Marker` (#2810). In any case, `Marker` hasn’t made it into a release yet.

antonhalim:
@1ec5 Thanks for the quick response. I know I might sound demanding, but this has been a long wait feature. Is there a way to give it a try or you know when it will be released?

mollymerp:
@antonhalim -- while you wait for the next release, feel free to use a copy of the master branch of this repo in your projects to get access to the new `Marker` class.

PR #2734, opened 2016-06-14
Ensure attribute type in pragma matches bucket

### LINKED ISSUE #2715: Assert that shader attribute type is as expected
Add an assertion to compare the shader attribute type as defined in the shader and as defined in the bucket to avoid bugs like https://github.com/mapbox/mapbox-gl-shaders/pull/16

cc @yhahn

lucaswoj:
fixes #2715 

cc @brunoabinader @yhahn @jfirebaugh


brunoabinader:
Nice!

Issue #2618, opened 2016-05-24
Feature request: multi touch panning


JarnoLeConte:
When using a multi touch trackpad _on OSX_ I would let two finger slide gestures result in map panning instead of zooming. That gesture is very intuitive, and allows panning with inertia, which is easier to use than dragging the map, although both options should be available of course. Zooming could already be done by using the pinch gesture.

This is implemented by Apple as well in their default Maps application on OSX. It would be a great improvement in my opinion.

How to implement this? Could handlers be customised / overridden? Or should it be implemented in this library itself?

1ec5:
We have this behavior by default in the [Mapbox OS X SDK](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/osx) for Cocoa applications, following MapKit’s lead. On the Web, Safari has a nonstandard [`GestureEvent`](https://developer.apple.com/library/iad/documentation/UserExperience/Reference/GestureEventClassReference/index.html) API to handle pinch and rotation gestures; I think it’s also used for two-finger panning gestures too.

1ec5:
On second thought, multitouch gestures are currently supported by GL JS in MobileSafari, in [touch_zoom_rotate.js](https://github.com/mapbox/mapbox-gl-js/blob/d6c34c81f7b0d6e77f1b25c2c080a3c5afba94ab/js/ui/handler/touch_zoom_rotate.js). Perhaps it could be generalized for desktop browsers too.

JarnoLeConte:
Firstly I thought pinch is already implemented, but probably it just works because it's interpreted as scroll gesture and therefore zooms in/out. Now it makes sense why that gesture not working smoothly.

It would be great to generalise touch_zoom_rotate.js for desktop. I will take a look at it.

jfirebaugh:
It looks like it isn't possible to detect pinch and rotate touch gestures on desktop browsers yet.

https://hammerjs.github.io/browser-support/

1ec5:
Desktop Safari isn’t included in that table; as I mentioned above, it does support a proprietary `GestureEvent` API for these gestures. Granted, desktop Safari has a lower usage share than some of the browsers listed there.

JarnoLeConte:
What about the default scroll event to implement panning?

1ec5:
> It looks like it isn't possible to detect pinch and rotate touch gestures on desktop browsers yet.

[...]

Issue #2719, opened 2016-06-11
Enable panning with multiple touch points


danielmorgan:
I've made a modification to the DragPanHandler to enable a feature I wanted in my implementation: Being able to pan the map with more than one finger. This hacky way I've done it probably wouldn't be suitable for merging back in, as evidenced by the fact that there is an explicit check to prevent this behaviour in `DragPanHandler::_ignoreEvent` that I have disabled.

Just wondering if multi-finger panning is something that could become a supported option, and if so what considerations are there for a robust implementation? Or is this something that can even be done in the user's implementation?

```patch
diff --git a/js/ui/handler/drag_pan.js b/js/ui/handler/drag_pan.js
index f0951d4..56128e5 100644
--- a/js/ui/handler/drag_pan.js
+++ b/js/ui/handler/drag_pan.js
@@ -94,8 +94,12 @@ DragPanHandler.prototype = {
             this._fireEvent('movestart', e);
         }
 
-        var pos = DOM.mousePos(this._el, e),
-            map = this._map;
+        var pos = (function(el, e) {
+            if (e.touches) return DOM.touchPos(el, e)[0];
+            return DOM.mousePos(el, e);
+        })(this._el, e);
+
+        var map = this._map;
 
         map.stop();
         this._drainInertiaBuffer();
@@ -180,9 +184,7 @@ DragPanHandler.prototype = {
 
         if (map.boxZoom && map.boxZoom.isActive()) return true;
         if (map.dragRotate && map.dragRotate.isActive()) return true;
-        if (e.touches) {
-            return (e.touches.length > 1);
-        } else {
+        if (!e.touches) {
             if (e.ctrlKey) return true;
             var buttons = 1,  // left button
                 button = 0;   // left button
```

lucaswoj:
Thanks for working on this feature @danielmorgan! I am supportive of adding this feature. Cut a PR with this diff and we'll talk about implementation details 😄

lucaswoj:
Closing as a duplicate of #2618

rantiev:
@lucaswoj I hope #2618 won't be closed as a duplicate of this one )

PR #2735, opened 2016-06-14
Enable panning with multiple touch points

danielmorgan:
As proposed in #2719, it would be nice to have the ability to pan while touching with multiple fingers, this behaviour is standard on native mobile map apps and feels intuitive.

This implementation is very naive as I don't know my way around the mapbox-gl library. My specific implementation has lots of other options disabled, so I haven't considered how this will interfere with other interaction handlers that deal with rotation and zooming for example.

Can someone take a look and perhaps point out the considerations I'm missing? It would be nice if this could become a supported feature, maybe as an option on the DragPanHandler API.

fixes https://github.com/mapbox/mapbox-gl-js/issues/2618

### DIFF
--- a/js/ui/handler/drag_pan.js
+++ b/js/ui/handler/drag_pan.js
@@ -94,8 +94,14 @@ DragPanHandler.prototype = {
             this._fireEvent('movestart', e);
         }
 
-        var pos = DOM.mousePos(this._el, e),
-            map = this._map;
+        var pos;
+        if (e.touches) {
+            pos = DOM.touchPos(this.el, e)[0]; // first touch point, ignore others
+        } else {
+            pos = DOM.mousePos(this.el, e); // or use mouse position
+        }
+
+        var map = this._map;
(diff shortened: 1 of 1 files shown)


lucaswoj:
This looks great @danielmorgan! Thanks for your contribution! 

This can be merged as soon as we make this behaviour configurable via a constructor option. I've just pushed a PR https://github.com/mapbox/mapbox-gl-js/pull/2740 that simplifies this part of the codebase. Check out how `bearingSnap` is implemented and see if you can build something similar.

mourner:
I think a better fix would be to modify [TouchZoomRotate](https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/handler/touch_zoom_rotate.js) handler to pan the map according to fingers position while pinching. This is what I intended to do originally (and how maps usually work) but haven't got to that yet.

lucaswoj:

PR #2746, opened 2016-06-15
Update all non-documentation uses of streets-v8 to streets-v9

### LINKED ISSUE #2563: Update examples and docs to use v9 styles

lucaswoj:
:new: 

fixes #2563


lucaswoj:
Thoughts on merging this @jfirebaugh @mourner @mollymerp? 

Updating the benchmarks to `-v9` could be contentious. My thoughts were

 - the benchmarks are only comparable in relative terms anyway because they are machine specific
 - we don't want to be stuck benchmarking based on old data forever, no time like the present

jfirebaugh:
Do it!

Issue #2750, opened 2016-06-15
Running mapbox-gl-js without a WebGL context


scothis:
Browser test environments [often do not support WebGL](https://saucelabs.ideas.aha.io/ideas/SLIDEA-I-54). So far, we've made due with trying to mock the mapbox-gl-js APIs to blackhole methods, however, our needs are growing more advanced and we want to interact with gl-js as if it were a real live instance emitting real events and supporting real user interactions.

Because the runtime doesn't support WebGL, we'd like to have a mode in gl-js where everything is real except for the WebGL context.

Initial thinking is to introduce a `MOCK_GL` config property that is `false` by default. When set to `true`, the `getWebGLContext` and `supportsWebGLContext`methods in `./util/browser/canvas` are mocked for new Canvas instances. We'll need to find a good way to mock a GL context in the browser.

jfirebaugh:
Can you explain the motivation here? It's difficult to see why running mapbox-gl-js without WebGL would be useful.

lucaswoj:
Mocking dependencies is a common pattern in unit testing. Let's explore some non-invasive mocking techniques before adding functionality to GL JS core. 

In particular, https://github.com/thlorenz/proxyquire plus https://github.com/thlorenz/proxyquireify might be useful. @anandthakker is currently working on a PR which adds proxyquire to our testing system #2685 

I would be happy to make small changes to GL JS to make WebGL easier to mock out with `proxyquire`.

scothis:
The primary use case I have is the Studio integration test suite which uses real web browsers and selenium. Hosted selenium environments typically do not support WebGL, and there have been numerous issues reported against those that do try to support WebGL.

To date, we have mocked gl-js at a high level by intercepting calls to `new Map()`. This worked as Studio was pushing data to gl-js and not as concerned about receiving data back from gl-js. This means we are not using automated tests for behavior like map movement, but instead rely on manual testing.

This pattern became problematic for us with the introduction of gl-draw. To properly integration a gl-js map with gl-draw installed we need the ability to both send data into the map and receive data from the map. Rather than attempt to mock the gl-js and gl-draw apis, we'd rather mock the limitation (WebGL).

allthesignals:
[...]

PR #2753, opened 2016-06-15
Allow the GL context to be mocked

scothis:
Mocking the GL context is useful in order to integration test
applications that interact with the mapbox-gl-js APIs, but do not have
WebGL available. By mocking the GL context, gl-js will continue to
respond to method invocations and emit events normally. It will fetch
styles, sprites, glyphs and tiles normally.

To mock the GL context, set `mapboxgl.config.MOCK_GL = true` before
creating the map instance.

Issue: #2750

### DIFF
--- a/debug/mock.html
+++ b/debug/mock.html
@@ -1,0 +1,151 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Mapbox GL JS debug page</title>
+    <meta charset='utf-8'>
+	<meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+
+    <link rel='stylesheet' href='../dist/mapbox-gl.css' />
+    <style>
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+        #checkboxes {
+            position: absolute;
+            background: #fff;
+            top:0;
+            left:0;
+            padding:10px;
+        }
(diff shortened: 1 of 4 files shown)


scothis:
Confirmed to work with a real browser hosted on SauceLabs that does not support WebGL.

lucaswoj:
As I said in https://github.com/mapbox/mapbox-gl-js/issues/2750, I am open to this change only if we cannot find a way to mock GL using non-invasive techniques such as proxyquire (even if that requires a small refactor).

scothis:
I was able to get this working directly inside Studio.

```js
  mockGL() {
    const MockWebGLRenderingContext = require('webgl-mock/src/WebGLRenderingContext');
    const MapboxGLCanvas = require('mapbox-gl/js/util/canvas');
    const contexts = new WeakMap();

    MapboxGLCanvas.prototype.getWebGLContext = function() {
      var context = contexts.get(this);
      if (!context) {
        context = new MockWebGLRenderingContext(this.canvas);
        context.getExtension = function(extension) {
          // disable VAO
          if (extension === 'OES_vertex_array_object') {
              return undefined;
          }
          return MockWebGLRenderingContext.prototype.getExtension.apply(this, arguments);
        };
        contexts.set(this, context);
      }
      return context;
    };
  }
```

Issue #2749, opened 2016-06-15
Revise existing documentation for compliance with "Writing API Documentation" guidelines


lucaswoj:
With the awesome new documentation guidelines defined in https://github.com/mapbox/mapbox-gl-js/pull/2744, it is time we did a spring cleaning of our existing documentation to bring it up to our own standards. 😄 

cc @davidtheclark @tmcw @lyzidiamond

lyzidiamond:
Looks like this was done in #2815! Close here?

lucaswoj:
Yes! Thanks for the 🎏  @lyzidiamond.

Issue #1975, opened 2016-01-22
all user-initiated events should include originalEvent


tmcw:
In order to determine the source of a map event, we need to specify originalEvent on any event initiated by a user.

bhousel:
I agree.  Some work already started on this in #1969 

I would need to add these to `touch_zoom_rotate` handler, and I'm also considering adding a param to the `jumpTo`/`easeTo` methods to be able to pass the original event info along to those methods which fire their own events.  Thoughts?

When you say "all"... I'm assuming you don't mean things like when user uses the zoom in/out or rotate controls.  We wouldn't want to propagate those click events, nor unproject and pass along the lng/lat underneath where the buttons happen to be.

tmcw:
An alternative idea is to create a way for API-initiated events to _not_ fire events, or indicate their sourcing in the same way.

bhousel:
@tmcw can you review e585f88 and see if that would help.  Going forward I'd like for all handlers to pass the original event source on to the camera methods.  Alternatively if you call camera methods directly, you could pass your own sourcing data in that parameter, then watch for it in the receiving code.

PR #1969, opened 2016-01-22
Event handling improvements 

bhousel:
Related improvements involving the event handlers.

- [X] Add `zoomstart` / `zoom` / `zoomend` events #1958
- [X] Propagate eventData to all event receivers #1975

PR #2755, opened 2016-06-16
API doc revisions

davidtheclark:
Ref #2749

@lucaswoj I was thinking I could gradually build up the revisions in this PR, working a little bit each day, and ask questions as I go. Sound good?

### DIFF
--- a/docs/_plugins/insert_token.rb
+++ b/docs/_plugins/insert_token.rb
@@ -1,6 +1,6 @@
 module InsertToken
     def insert_token(input)
-        input.sub("<script>", "<script>\nmapboxgl.accessToken = '<your access token here>';")
+        input.sub("<script>", "<script>\n  mapboxgl.accessToken = '<your access token here>';")
     end
 end
 
--- a/docs/_plugins/insert_token.rb
+++ b/docs/_plugins/insert_token.rb
@@ -1,6 +1,6 @@
 module InsertToken
     def insert_token(input)
-        input.sub("<script>", "<script>\n  mapboxgl.accessToken = '<your access token here>';")
+        input.sub("<script>", "<script>\nmapboxgl.accessToken = '<your access token here>';")
     end
 end
 
--- a/js/geo/lng_lat_bounds.js
+++ b/js/geo/lng_lat_bounds.js
@@ -5,12 +5,18 @@ module.exports = LngLatBounds;
 var LngLat = require('./lng_lat');
 
 /**
- * Creates a bounding box from the given pair of points. If parameteres are omitted, a `null` bounding box is created.
+ * A `LngLatBounds` object represents a geographical bounding box,
+ * defined by its southwest and northeast points in longitude and latitude.
+ *
+ * If no arguments are provided to the constructor, a `null` bounding box is created.
+ *
+ * Note that any Mapbox GL method that accepts a `LngLatBounds` object as an argument or option
+ * can also accept an `Array` of two [`LngLatLike`](#LngLatLike) constructs and will perform an implicit conversion.
+ * This flexible type is documented as [`LngLatLikeBounds`](#LngLatLikeBounds).
  *
  * @class LngLatBounds
- * @classdesc A representation of rectangular box on the earth, defined by its southwest and northeast points in longitude and latitude.
- * @param {LngLat} sw southwest
- * @param {LngLat} ne northeast
+ * @param {LngLatLike} sw The southwest corner of the bounding box.
+ * @param {LngLatLike} ne The northeast corner of the bounding box.
  * @example
  * var sw = new mapboxgl.LngLat(-73.9876, 40.7661);
  * var ne = new mapboxgl.LngLat(-73.9397, 40.8002);
@@ -31,9 +37,9 @@ function LngLatBounds(sw, ne) {
 LngLatBounds.prototype = {
 
     /**
-     * Extend the bounds to include a given LngLat or LngLatBounds.
+     * Extends the bounding box to include an area represented by a `LngLat` or `LngLatBounds`.
      *
-     * @param {LngLat|LngLatBounds} obj object to extend to
+     * @param {LngLat|LngLatBounds} obj The area that the bounding box will extend to include.
      * @returns {LngLatBounds} `this`
      */
     extend: function(obj) {
@@ -70,8 +76,9 @@ LngLatBounds.prototype = {
     },
 
     /**
-     * Get the point equidistant from this box's corners
-     * @returns {LngLat} centerpoint
+     * Returns the geographical coordinate equidistant from the bounding box's corners.
+     *
+     * @returns {LngLat} The bounding box's center.
      * @example
      * var llb = new mapboxgl.LngLatBounds([-73.9876, 40.7661], [-73.9397, 40.8002]);
      * llb.getCenter(); // = LngLat {lng: -73.96365, lat: 40.78315}
@@ -81,57 +88,66 @@ LngLatBounds.prototype = {
     },
 
     /**
-     * Get southwest corner
-     * @returns {LngLat} southwest
+     * Returns the southwest corner of the bounding box.
+     *
+     * @returns {LngLat} The southwest corner of the bounding box.
      */
     getSouthWest: function() { return this._sw; },
 
     /**
-     * Get northeast corner
-     * @returns {LngLat} northeast
+    * Returns the northeast corner of the bounding box.
+    *
+    * @returns {LngLat} The northeast corner of the bounding box.
      */
     getNorthEast: function() { return this._ne; },
 
     /**
-     * Get northwest corner
-     * @returns {LngLat} northwest
+    * Returns the northwest corner of the bounding box.
+    *
+    * @returns {LngLat} The northwest corner of the bounding box.
      */
     getNorthWest: function() { return new LngLat(this.getWest(), this.getNorth()); },
 
     /**
-     * Get southeast corner
-     * @returns {LngLat} southeast
+    * Returns the southeast corner of the bounding box.
+    *
+    * @returns {LngLat} The southeast corner of the bounding box.
      */
     getSouthEast: function() { return new LngLat(this.getEast(), this.getSouth()); },
 
     /**
-     * Get west edge longitude
-     * @returns {number} west
+    * Returns the west edge of the bounding box.
+    *
+    * @returns {LngLat} The west edge of the bounding box.
      */
     getWest:  function() { return this._sw.lng; },
 
     /**
-     * Get south edge latitude
-     * @returns {number} south
+    * Returns the south edge of the bounding box.
+    *
+    * @returns {LngLat} The south edge of the bounding box.
      */
     getSouth: function() { return this._sw.lat; },
 
     /**
-     * Get east edge longitude
-     * @returns {number} east
+    * Returns the east edge of the bounding box.
+    *
+    * @returns {LngLat} The east edge of the bounding box.
      */
     getEast:  function() { return this._ne.lng; },
 
     /**
-     * Get north edge latitude
-     * @returns {number} north
+    * Returns the north edge of the bounding box.
+    *
+    * @returns {LngLat} The north edge of the bounding box.
      */
     getNorth: function() { return this._ne.lat; },
 
     /**
-     * Return a `LngLatBounds` as an array
+     * Returns the bounding box represented as an array.
      *
-     * @returns {array} [lng, lat]
+     * @returns {array} The bounding box represented as an array, consisting of the
+     *   southwest and northeast coordinates of the bounding represented as arrays of numbers.
      * @example
      * var llb = new mapboxgl.LngLatBounds([-73.9876, 40.7661], [-73.9397, 40.8002]);
      * llb.toArray(); // = [[-73.9876, 40.7661], [-73.9397, 40.8002]]
@@ -141,9 +157,10 @@ LngLatBounds.prototype = {
     },
 
     /**
-     * Return a `LngLatBounds` as a string
+     * Return the bounding box represented as a string.
      *
-     * @returns {string} "LngLatBounds(LngLat(lng, lat), LngLat(lng, lat))"
+     * @returns {string} The bounding box represents as a string of the format
+     *   `'LngLatBounds(LngLat(lng, lat), LngLat(lng, lat))'`.
      * @example
      * var llb = new mapboxgl.LngLatBounds([-73.9876, 40.7661], [-73.9397, 40.8002]);
      * llb.toString(); // = "LngLatBounds(LngLat(-73.9876, 40.7661), LngLat(-73.9397, 40.8002))"
@@ -154,13 +171,14 @@ LngLatBounds.prototype = {
 };
 
 /**
- * Convert an array to a `LngLatBounds` object, or return an existing
- * `LngLatBounds` object unchanged.
+ * Converts an array to a `LngLatBounds` object.
  *
- * Calls `LngLat#convert` internally to convert arrays as `LngLat` values.
+ * If a `LngLatBounds` object is passed in, the function returns it unchanged.
  *
- * @param {LngLatBounds|Array<number>|Array<Array<number>>} input input to convert to a LngLatBounds
- * @returns {LngLatBounds} LngLatBounds object or original input
+ * Internally, the function calls `LngLat#convert` to convert arrays to `LngLat` values.
+ *
+ * @param {LngLatBoundsLike} input An array of two coordinates to convert, or a `LngLatBounds` object to return.
+ * @returns {LngLatBounds} A new `LngLatBounds` object, if a conversion occurred, or the original `LngLatBounds` object.
  * @example
  * var arr = [[-73.9876, 40.7661], [-73.9397, 40.8002]];
  * var llb = mapboxgl.LngLatBounds.convert(arr);
--- a/js/geo/lng_lat_bounds.js
+++ b/js/geo/lng_lat_bounds.js
@@ -12,7 +12,7 @@ var LngLat = require('./lng_lat');
  *
  * Note that any Mapbox GL method that accepts a `LngLatBounds` object as an argument or option
  * can also accept an `Array` of two [`LngLatLike`](#LngLatLike) constructs and will perform an implicit conversion.
- * This flexible type is documented as [`LngLatLikeBounds`](#LngLatLikeBounds).
+ * This flexible type is documented as [`LngLatBoundsLike`](#LngLatBoundsLike).
  *
  * @class LngLatBounds
  * @param {LngLatLike} sw The southwest corner of the bounding box.
--- a/js/source/geojson_source.js
+++ b/js/source/geojson_source.js
@@ -10,16 +10,19 @@ var EXTENT = require('../data/bucket').EXTENT;
 module.exports = GeoJSONSource;
 
 /**
- * Create a GeoJSON data source instance given an options object
+ * A datasource containing GeoJSON.
+ *
  * @class GeoJSONSource
  * @param {Object} [options]
- * @param {Object|string} options.data A GeoJSON data object or URL to it. The latter is preferable in case of large GeoJSON files.
- * @param {number} [options.maxzoom=18] Maximum zoom to preserve detail at.
- * @param {number} [options.buffer] Tile buffer on each side in pixels.
- * @param {number} [options.tolerance] Simplification tolerance (higher means simpler) in pixels.
- * @param {number} [options.cluster] If the data is a collection of point features, setting this to true clusters the points by radius into groups.
- * @param {number} [options.clusterRadius=50] Radius of each cluster when clustering points, in pixels.
- * @param {number} [options.clusterMaxZoom] Max zoom to cluster points on. Defaults to one zoom less than `maxzoom` (so that last zoom features are not clustered).
+ * @param {Object|string} options.data A GeoJSON data object or a URL to one. The latter is preferable in the case of large GeoJSON objects.
+ * @param {number} [options.maxzoom=18] The maximum zoom level at which to preserve detail (1-20).
+ * @param {number} [options.buffer] The tile buffer, measured in pixels.
+ * @param {number} [options.tolerance] The simplification tolerance, measured in pixels (higher means simpler).
+ * @param {boolean} [options.cluster] If `true`, a collection of point features will be clustered into groups,
+ *   according to `options.clusterRadius`.
+ * @param {number} [options.clusterRadius=50] The radius of each cluster when clustering points, measured in pixels.
+ * @param {number} [options.clusterMaxZoom] The maximum zoom level to cluster points in. By default, this value is
+ *   one zoom level less than the map's `maxzoom`, so that at the highest zoom level features are not clustered.
 
  * @example
  * var sourceObj = new mapboxgl.GeoJSONSource({
@@ -86,9 +89,9 @@ GeoJSONSource.prototype = util.inherit(Evented, /** @lends GeoJSONSource.prototy
     isTileClipped: true,
 
     /**
-     * Update source geojson data and rerender map
+     * Sets the GeoJSON data and re-render the map.
      *
-     * @param {Object|string} data A GeoJSON data object or URL to it. The latter is preferable in case of large GeoJSON files.
+     * @param {Object|string} data A GeoJSON data object or a URL to one. The latter is preferable in the case of large GeoJSON files.
      * @returns {GeoJSONSource} this
      */
     setData: function(data) {
--- a/js/source/geojson_source.js
+++ b/js/source/geojson_source.js
@@ -16,8 +16,14 @@ module.exports = GeoJSONSource;
  * @param {Object} [options]
  * @param {Object|string} options.data A GeoJSON data object or a URL to one. The latter is preferable in the case of large GeoJSON objects.
  * @param {number} [options.maxzoom=18] The maximum zoom level at which to preserve detail (1-20).
- * @param {number} [options.buffer] The tile buffer, measured in pixels.
- * @param {number} [options.tolerance] The simplification tolerance, measured in pixels (higher means simpler).
+ * @param {number} [options.buffer=128] The tile buffer, measured in pixels. The buffer extends each
+ *   tile's data just past its visible edges, helping to ensure seamless rendering across tile boundaries.
+ *   The default value, 128, is a safe value for label layers, preventing text clipping at boundaries.
+ *   You can read more about buffers and clipping in the
+ *   [Mapbox Vector Tile Specification](https://www.mapbox.com/vector-tiles/specification/#clipping).
+ * @param {number} [options.tolerance=0.375] The simplification tolerance, measured in pixels.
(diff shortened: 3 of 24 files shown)


js/ui/map.js
```diff
@@ -57,47 +57,51 @@ var defaultOptions = {
...
- * @param {boolean} [options.touchZoomRotate=true] If `true`, enable the "pinch to rotate and zoom" interaction (see `TouchZoomRotateHandler`).
- * @param {boolean} [options.trackResize=true]  If `true`, automatically resize the map when the browser window resizes.
- * @param {LngLat} [options.center] The geographic coordinate on which the map's initial viewport is centered.
- * @param {number} [options.zoom] The zoom level of the map's initial viewport.
- * @param {number} [options.bearing] The bearing (rotation) of the map's initial viewport measured in degrees counter-clockwise from north.
- * @param {number} [options.pitch] The pitch of the map's initial viewport measured in degrees.
+ * @param {boolean} [options.preserveDrawingBuffer=false] If `true`, the map's canvas can be exported to a PNG using `map.getCanvas().toDataURL();`. This is `false` by default as a performance optimization.
+ * @param {LngLatBounds|Array<Array<number>>} [options.maxBounds] If set, the map is constrained to the given bounds. // TODO: Can LngLatBounds be like LngLat, standing in for Array<Array<string>>
```
davidtheclark: My understanding is that throughout the docs `@param {LngLat}` stands in for `@param {LngLat | Array<string>}`. Does it make sense to do the same for `LngLatBounds` and `Array<Array<number>>`?

js/ui/map.js
```diff
@@ -57,47 +57,51 @@ var defaultOptions = {
...
+ * @param {boolean} [options.keyboard=true] If `true`, keyboard shortcuts are enabled (see [`KeyboardHandler`](#KeyboardHandler)).
+ * @param {boolean} [options.doubleClickZoom=true] If `true`, the "double click to zoom" interaction is enabled (see [`DoubleClickZoomHandler`](#DoubleClickZoomHandler)).
+ * @param {boolean} [options.touchZoomRotate=true] If `true`, the "pinch to rotate and zoom" interaction is enabled (see [`TouchZoomRotateHandler`](#TouchZoomRotateHandler)).
+ * @param {boolean} [options.trackResize=true]  If `true`, the map will automatically resize when the browser window resizes.
+ * @param {LngLat} [options.center] The geographic coordinate on which the map's initial viewport is centered. TODO: What is default?
+ * @param {number} [options.zoom] The zoom level of the map's initial viewport. TODO: What is default?
+ * @param {number} [options.bearing] The bearing (rotation) of the map's initial viewport, measured in degrees counter-clockwise from north. TODO: What is default?
+ * @param {number} [options.pitch] The pitch of the map's initial viewport, measured in degrees. TODO: What is default?
```
davidtheclark: What are the definitive default values for center, zoom, bearing and pitch? Is it accurate to just say 0 for all of them?

js/ui/map.js
```diff
@@ -57,47 +57,51 @@ var defaultOptions = {
...
+ * object.
+ *
  * @class Map
  * @param {Object} options
- * @param {string|Element} options.container HTML element to initialize the map in (or element id as string)
- * @param {number} [options.minZoom=0] Minimum zoom of the map
- * @param {number} [options.maxZoom=20] Maximum zoom of the map
- * @param {Object|string} [options.style] Map style. This must be an an object conforming to the schema described in
- * the [style reference](https://mapbox.com/mapbox-gl-style-spec/), or a URL to a JSON style. To load a style from the
+ * @param {Element|string} options.container The HTML element to initialize the map in, or the element's string `id`.
```
lucaswoj: >  - `@param`, `@property`, and `@returns` descriptions should be capitalized and end with a period. They should begin as if completing a sentence beginning with "This is..." or "This..."

What do you think about revising the standard to make these start with a lowercase letter? They aren't complete sentences so formatting them as such is odd.

js/ui/map.js
lucaswoj: Yes! That makes sense!

Another option is to create a `LngLatLike` `@typedef`. This would allow us to keep the docs more concise and add new supported formats in the future. 

```
 * @typedef {(LngLat | Array<string>)} LngLatLike
```

js/ui/map.js
lucaswoj: - center: [0, 0]
 - zoom: 0
 - bearing: 0
 - pitch: 0

lucaswoj:
This is so so so great! Thank you for spearheading this project @davidtheclark.

js/ui/map.js
davidtheclark: I understand the reaction, but I think it's actually a pretty standard format for definitions and reference material. Many dictionary or encyclopedia, I think, will start most entries with fragments that begin with capital letters and end with periods. We could go the other way — always lowercase — but then we end up being inconsistent when a sentence happens to be complete, or we have a fragment followed by full sentences. What do you think?

davidtheclark: BTW totally fine if you want to go to lowercase, was just explaining the rationale.

lucaswoj: @tmcw, do you have any thoughts?

tmcw: Yeah, I think I agree with @davidtheclark's viewpoint here. Though they are sentence fragments, having them start with caps and end with periods works functionally in this setting.

lucaswoj: Great! I'm on board with this decision 🚀.

js/ui/map.js
davidtheclark: That is a great idea. I'll try to figure out how to make that work with the generated page.

js/ui/map.js
jfirebaugh: We should note though that the style can also specify these values. In descending order of precedence:

* Explicitly-specified option to `Map` constructor
* Specified in style
* "0" defaults

js/ui/map.js
```diff
@@ -57,47 +57,61 @@ var defaultOptions = {
...
- * option, and the map's style with `style` and other attributes of the map,
- * and in return Mapbox GL JS initializes the map on your page and returns
- * a map variable that lets you programmatically call methods on the map.
+ * A [`LngLat`](#LngLat) object or an array of two numbers representing longitude and latitude.
+ * @typedef {(LngLat | Array<number>)} LngLatLike
+ * /
+
+/**
+ * A [`LngLatBounds`](#LngLatBounds) object or an array of [`LngLatLike`](#LngLatLike) objects.
+ * @typedef {(LngLatBounds | Array<LngLatLike>)} LngLatBoundsLike
```
davidtheclark: Added these `@typedef`s, but not sure how to get them to render in a good place via documention.js. Seems like maybe they aren't being parsed correctly, either ... or my syntax is off. @tmcw Any insight here?

lucaswoj:
All the latest commits look great! 🏆

davidtheclark:
Cool. I will continue in that pattern, then.

davidtheclark:
I finished going through all the documentation, at last, and rebased the work into a single commit. I do have a few outstanding questions:

- Am I right in thinking that pitch is on a scale of 0-60?
- I become uncertain about which events provide EventData objects and which just provide the original DOM event? And are there EventData objects that do not contain an originalEvent, or is EventData only ever produced as the result of an interaction?

davidtheclark:
Anybody in @mapbox/copycop interested in helping fix up the Mapbox GL JS docs? All feedback welcome on the changes in this PR!

docs/_plugins/insert_token.rb
```diff
@@ -1,6 +1,6 @@
 module InsertToken
     def insert_token(input)
-        input.sub("<script>", "<script>\nmapboxgl.accessToken = '<your access token here>';")
+        input.sub("<script>", "<script>\n  mapboxgl.accessToken = '<your access token here>';")
```
lucaswoj: Why did you add 2 spaces here?

The formatting of our examples suggests this line should not be indented. For example, see https://www.mapbox.com/mapbox-gl-js/examples

davidtheclark: Ah ... the docs page is different (screenshot below). But instead of changing what I changed, we should probably just unindent the `map` line there so it matches the examples, right?

screen shot 2016-06-28 at 2 17 49 pm

js/geo/lng_lat_bounds.js
```diff
@@ -5,12 +5,18 @@ module.exports = LngLatBounds;
...
 /**
- * Creates a bounding box from the given pair of points. If parameteres are omitted, a `null` bounding box is created.
+ * A `LngLatBounds` object represents a geographical bounding box,
+ * defined by its southwest and northeast points in longitude and latitude.
+ *
+ * If no arguments are provided to the constructor, a `null` bounding box is created.
+ *
+ * Note that any Mapbox GL method that accepts a `LngLatBounds` object as an argument or option
+ * can also accept an `Array` of two [`LngLatLike`](#LngLatLike) constructs and will perform an implicit conversion.
+ * This flexible type is documented as [`LngLatLikeBounds`](#LngLatLikeBounds).
```
lucaswoj: `LngLatLikeBounds` --> `LngLatBoundsLike`

js/source/geojson_source.js
```diff
@@ -86,9 +89,9 @@ GeoJSONSource.prototype = util.inherit(Evented, /** @lends GeoJSONSource.prototy
     isTileClipped: true,
 
     /**
-     * Update source geojson data and rerender map
+     * Sets the GeoJSON data and re-render the map.
```
lucaswoj: `re-render` -> `re-renders`

js/source/geojson_source.js
```diff
@@ -10,16 +10,19 @@ var EXTENT = require('../data/bucket').EXTENT;
...
- * @param {number} [options.maxzoom=18] Maximum zoom to preserve detail at.
- * @param {number} [options.buffer] Tile buffer on each side in pixels.
- * @param {number} [options.tolerance] Simplification tolerance (higher means simpler) in pixels.
- * @param {number} [options.cluster] If the data is a collection of point features, setting this to true clusters the points by radius into groups.
- * @param {number} [options.clusterRadius=50] Radius of each cluster when clustering points, in pixels.
- * @param {number} [options.clusterMaxZoom] Max zoom to cluster points on. Defaults to one zoom less than `maxzoom` (so that last zoom features are not clustered).
+ * @param {Object|string} options.data A GeoJSON data object or a URL to one. The latter is preferable in the case of large GeoJSON objects.
+ * @param {number} [options.maxzoom=18] The maximum zoom level at which to preserve detail (1-20).
+ * @param {number} [options.buffer] The tile buffer, measured in pixels.
+ * @param {number} [options.tolerance] The simplification tolerance, measured in pixels (higher means simpler).
```
lucaswoj: Would it be possible to link to a more thorough explanation of this parameter? For example, the wikipedia page https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm

js/source/geojson_source.js
```diff
@@ -10,16 +10,19 @@ var EXTENT = require('../data/bucket').EXTENT;
...
- * @param {Object|string} options.data A GeoJSON data object or URL to it. The latter is preferable in case of large GeoJSON files.
- * @param {number} [options.maxzoom=18] Maximum zoom to preserve detail at.
- * @param {number} [options.buffer] Tile buffer on each side in pixels.
- * @param {number} [options.tolerance] Simplification tolerance (higher means simpler) in pixels.
- * @param {number} [options.cluster] If the data is a collection of point features, setting this to true clusters the points by radius into groups.
- * @param {number} [options.clusterRadius=50] Radius of each cluster when clustering points, in pixels.
- * @param {number} [options.clusterMaxZoom] Max zoom to cluster points on. Defaults to one zoom less than `maxzoom` (so that last zoom features are not clustered).
+ * @param {Object|string} options.data A GeoJSON data object or a URL to one. The latter is preferable in the case of large GeoJSON objects.
+ * @param {number} [options.maxzoom=18] The maximum zoom level at which to preserve detail (1-20).
+ * @param {number} [options.buffer] The tile buffer, measured in pixels.
```
lucaswoj: Would it be possible to link to a more thorough explanation of this parameter? This documentation is the best I could find https://www.mapbox.com/vector-tiles/specification/#clipping

cc @jakepruitt @mapsam

js/source/image_source.js
```diff
@@ -15,11 +15,15 @@ var VertexArrayObject = require('../render/vertex_array_object');
...
 /**
- * Create an Image source instance given an options object
+ * A data source containing an image.
+ *
  * @class ImageSource
  * @param {Object} options
- * @param {string} options.url A string URL of an image file
- * @param {Array} options.coordinates Four geographical [lng, lat] coordinates in clockwise order defining the corners (starting with top left) of the image. Does not have to be a rectangle.
+ * @param {string} options.url The URL of an image file.
+ * @param {Array} options.coordinates Four geographical coordinates,
```
lucaswoj: Would it be accurate to list the type signature as `Array<Array<number>>`?

js/source/image_source.js
```diff
@@ -64,9 +68,12 @@ ImageSource.prototype = util.inherit(Evented, /** @lends ImageSource.prototype *
     },
 
     /**
-     * Update image coordinates and rerender map
+     * Sets the image's coordinates and re-renders the map.
      *
-     * @param {Array} coordinates Four geographical [lng, lat] coordinates in clockwise order defining the corners (starting with top left) of the image. Does not have to be a rectangle.
+     * @param {Array} coordinates Four geographical coordinates,
```
lucaswoj: Would it be accurate to list the type signature as `Array<Array<number>>`?

js/source/video_source.js
```diff
@@ -73,9 +77,9 @@ VideoSource.prototype = util.inherit(Evented, /** @lends VideoSource.prototype *
     roundZoom: true,
 
     /**
-     * Return the HTML video element.
+     * Returns the HTML `video` element.
      *
-     * @returns {Object}
+     * @returns {Node} The HTML `video` element.
```
lucaswoj: I think documentation.js plans to leverage Flow's type checker. The type Flow would use here is `HTMLVideoElement`.

js/ui/handler/scroll_zoom.js
```diff
@@ -146,31 +151,31 @@ ScrollZoomHandler.prototype = {
 
 
 /**
- * Zoom start event. This event is emitted just before the map begins a transition from one
- * zoom level to another, either as a result of user interaction or the use of methods such as `Map#jumpTo`.
+ * Fired just before the map begins a transition from one zoom level to another,
+ * as the result of either user interaction or methods such as [`Map#jumpTo`](#Map#jumpTo).
```
lucaswoj: 💭 Because `Map#jumpTo` is instantaneous, I don't think it triggers a `zoomstart` event. `Map#flyTo` is a better example.

My apologies -- I know this is tangential to the point of this PR.

davidtheclark: Perfect! I think of clarifying all these things as the point of the PR. It got pretty big, as you can see, so it's inevitable that I made typos and missed things like this. Thanks for looking things through in depth!

js/ui/map.js
```diff
@@ -57,48 +57,64 @@ var defaultOptions = {
...
- * @param {number} [options.workerCount=navigator.hardwareConcurrency - 1] The number of WebWorkers the map should use to process vector tile data.
+ *
+ * @param {boolean} [options.hash=false] If `true`, the map's position (zoom, center latitude, center longitude, and bearing) will be synced with the hash fragment of the page's URL.
+ *   For example, `http://path/to/my/page.html#2.59/39.26/53.07/-24.1`.
+ * @param {boolean} [options.interactive=true] If `false`, no mouse, touch, or keyboard listeners will be attached to the map, so it will not respond to interaction.
+ * @param {number} [options.bearingSnap=7] The threshold, measured in degrees, that determines when the map's
+ *   bearing (rotation) will snap to north. For example, with a `bearingSnap` of 7, if the user has rotated
+ *   the map within 7 degrees of north, the map will automatically snap to north exactly.
+ * @param {Array<string>} [options.classes] Mapbox GL style class names with which to initialize the map.
+ *   Keep in mind that these are used for filtering a style layer's pain properties, so are *not* reflected
```
lucaswoj: "for ~~filtering~~ a style layer's ~~pain~~ properties" -> "for **grouping** a style layer's **paint** properties"

js/ui/map.js
```diff
@@ -212,9 +228,13 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
-     * Adds a style class to a map.
+     * Adds a Mapbox GL style class to the map.
      *
-     * @param {string} klass name of style class
+     * Keep in mind that these are used for filtering a style layer's pain properties, so are *not* reflected
```
lucaswoj: paint properties

js/ui/map.js
```diff
@@ -267,30 +287,31 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
     /**
-     * Detect the map's new width and height and resize it. Given
-     * the `container` of the map specified in the Map constructor,
-     * this reads the new width from the DOM: so this method is often
-     * called after the map's container is resized by another script
-     * or the map is shown after being initially hidden with CSS.
+     * Resizes the map according to the dimensions of its
+     * `container` element.
+     *
+     * This method is often called after the map's `container` is resized by another script,
```
lucaswoj: "is often" -> "must be"

js/ui/map.js
```diff
@@ -399,51 +428,56 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
-     * @param {Array} [params.filter] A [filter](https://www.mapbox.com/mapbox-gl-style-spec/#types-filter).
+     * @param {PointLike|Array<PointLike>} [region] - The geometry of the query region:
+     * either a single point or southwest and northeast points describing a bounding box.
+     * Omitting this parameter (i.e. calling [`Map#queryRenderedFeatures`](#Map#queryRenderedFeatures) with zero arguments,
+     * or with only a `parameters` argument) is equivalent to passing a bounding box encompassing the entire
+     * map viewport.
+     * @param {Object} [parameters]
+     * @param {Array<string>} [parameters.layers] An array of style layer IDs for the query to inspect.
+     *   Only features within these layers will be returned. If this parameter is undefined, all layers will be checked.
+     * @param {Array<string>} [parameters.filter] A [filter](https://www.mapbox.com/mapbox-gl-style-spec/#types-filter)
```
lucaswoj: Let's leave this type parameter as `Array`. `Array<string>` doesn't accurately capture filters like `['any', ['==', '@penny', true], ['==', '@linus', true]]`

js/ui/map.js
```diff
@@ -548,11 +585,11 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
-     * Replaces the map's style object with a new value. Unlike the `style`
-     * option in the Map constructor, this method only accepts an object
-     * of a new style, not a URL string.
+     * Replaces the map's Mapbox GL style object with a new value.
+     *
+     * Unlike the `style` option in the `Map` constructor, this method only accepts an object, not a URL string.
```
lucaswoj: Let's remove this line. `setStyle` does accept URL strings as the [normalization is handled within `Style`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style.js#L71-L75).

docs/_plugins/insert_token.rb
lucaswoj: > we should probably just unindent the map line there so it matches the examples, right?

Sounds good 👍

lucaswoj:
Done proofreading. I think we're ready to rebase & merge once the above feedback is taken into account. 

Thank you for tackling this project! Our docs are so much better off now.

js/source/geojson_source.js
mapsam: Hmmm, I wish http://mapnik.org/documentation/node-mapnik/3.5/#VectorTile described it better, but it doesn't. I have some things to update to the mapnik docs, and can include this.

davidtheclark:
I've made changes based on your feedback:

- Fixed typos you found.
- Altered indentation on the homepage’s example `<script>` tag.
- Added a little more explanation and an external link for GeoJSONSource’s `options.tolerance` and `options.buffer`. Also added defaults to these from what I saw in the code. *Are these explanations and defaults correct?* I am no expert yet ...
- Fixed the param types that you identified.
- Added `mouseout` event while rebasing.

I also found a few other little tweaks to perform as I read through some of it again.

I'm still uncertain about the event data, though. Specifically, those events that now have their object defined something like this: `@property {EventData} data The original event data, if the event was the result of user interaction.` Two questions:
- Can we just say this is `@property {EventData} data`, since the documentation for EventData covers the original event? 
- If on a simple example map I log the event data fired with `zoomend`, then I double-click to zoom, the object I see only has `type` (`"zoomend"`) and `target` (the `Map` instance) properties, not EventData's properties, and not original event data even though it originated from my double-click. What am I missing here?

lucaswoj:
> I'm still uncertain about the event data, though. Specifically, those events that now have their object defined something like this: `@property {EventData} data The original event data, if the event was the result of user interaction.` Two questions:
> - Can we just say this is `@property {EventData} data`, since the documentation for EventData covers the original event? 

Absolutely! I think this falls under the bullet point in the doc guidelines that says "Documentation descriptions must contain more information than what is obvious from the identifier and JSDoc metadata."

> - If on a simple example map I log the event data fired with `zoomend`, then I double-click to zoom, the object I see only has `type` (`"zoomend"`) and `target` (the `Map` instance) properties, not EventData's properties, and not original event data even though it originated from my double-click. What am I missing here?

This sounds like a bug to me.

lucaswoj:
While we're on the topic of `EventData`...

The names `Event` and `EventData` are not reflective of the types' semantics.

 - `Event` - any DOM event
 - `EventData` - any Mapbox GL JS event (some of which "wrap" DOM events)

Would renaming `EventData` to `MapboxEvent` make more sense?

davidtheclark:
> Would renaming EventData to MapboxEvent make more sense?

Yeah, that makes a lot of sense, I think.

So do *all* events documented here fire with a MapboxEvent? And lines like https://github.com/mapbox/mapbox-gl-js/pull/2755/files#diff-274e90d38623e2e16a5e7d34414f33c7R1325 are incorrect? (I tried this out, and it does look like with `map.on('mousedown', ..)` I get a MapboxEvent, not just the original DOM MouseEvent.)

lucaswoj:
It looks like some events have DOM `Event`s and some have `MapboxEvent`s. The intended distinction seems to be

 - events that proxy DOM Events (i.e. `mousedown`) have `Event`s
 - events that are unique to GL JS (i.e. `moveend`) have `MapboxEvent`s

lucaswoj:
Here's the code that fires `mousedown` https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/bind_handlers.js#L45-L49

bhousel:
> Would renaming EventData to MapboxEvent make more sense?

I don't think so.. `EventData` is just an unstructured bucket to store whatever the listener wants to watch for.   It's not an event.  It's not guaranteed to always contain the same stuff.  (e.g. the click events dispatch an eventData with a `LngLat` property, but boxzoom events dispatch an eventData with `LngLatBounds` property.)

davidtheclark:
@lucaswoj So I'm seeing in that code that the native `MouseEvent` from `mousedown` ends up being wrapped in an object matching the documented `EventData` API: https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/bind_handlers.js#L126. Seems, then, like we may never actually send the raw DOM `Event`?

> EventData is just an unstructured bucket to store whatever the listener wants to watch for

@bhousel Sounds, then, like the existing docs are inaccurate, right? They suggest that `EventData` is a class and we send out instances of it — when it fact it seems like it's more of an interface, with a few possible properties that aren't documented yet, like `LngLatBounds` for `boxzoom` or `LngLats` and `points` for touch events (https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/bind_handlers.js#L142)?

lucaswoj:
Sounds like we should replace all references to `EventData` with `Object`

bhousel:
> From what you're saying it sounds like the existing docs are inaccurate, right? They suggest that EventData is a class and we send out instances of it

Yes, I think that's right.  It's not so much a class, but more that the caller can pass an object with whatever tracing data they want to watch for, and our handlers will decorate that object with additional properties and dispatch on to any event listeners.  (see #1975, #1969 for context)

I don't know the best way to document it.  We could just call them all `Object` and then in each documentation block enumerate the extra properties that the decorated object might contain, but I thought that would get repetitive, so I rolled them all into a thing called `EventData`.

davidtheclark:
In d79732151b583dc2f7878f261e2c48415b1b44f7 I tried something to account for everything you've pointed out above. What do you think of that approach, @bhousel and @lucaswoj?

lucaswoj:
I'm learning so much about our event system! A huge thanks to @bhousel for all the help.

From what I understand, we can document our event system the most accurately with the following interfaces. Most event handlers could declare their signature as `Map*Event | Object` because they may return a `Map*Event` or an arbitrary user-provided object. (Declaring the signature as `Object` would be accurate too but providing a pointer to the relevant `Map*Event` is helpful to humans)

# `MapMouseEvent`

https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/bind_handlers.js#L126-L130 

```
/**
 * @typedef {Object} MapMouseEvent
 * @property {string} type The event type.
 * @property {Map} target The `Map` object that fired the event.
 * @property {MouseEvent} originalEvent
 * @property {Point} point The pixel coordinates of the mouse event target, relative to the map
 *   and measured from the top left corner.
 * @property {LngLat} lngLat The geographic location on the map of the mouse event target.
 */
```

# `MapTouchEvent`

https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/bind_handlers.js#L139-L145

```
/**
 * @typedef {Object} MapTouchEvent
 * @property {string} type The event type.
 * @property {Map} target The `Map` object that fired the event.
 * @property {TouchEvent} originalEvent
 * @property {Point} point The pixel coordinates of the center of the touch event points, relative to the map
 *   and measured from the top left corner.
 * @property {LngLat} lngLat The geographic location on the map of the center of the touch event points.
 */
```

# `MapBoxZoomEvent`

```
/**
 * @typedef {Object} MapTouchEvent
 * @property {LngLatBounds} boxZoomBounds The bounding box of the "box zoom" interaction.
 * @property {MapMouseEvent} originalEvent
 */
```

davidtheclark:
Great! I will update the PR soon.

davidtheclark:
Looks like `zoom` and `move` events typically are just `type`, `target`, and `originalEvent` (either `MouseEvent` or `TouchEvent`), unless they are caused by a double-click zoom, in which case there's no `originalEvent`.

lucaswoj:
I am so excited! :shipit:

lucaswoj:
Do you want to merge this into `mb-pages` instead of `master`?

davidtheclark:
Sure, I can rebase onto that and submit a new PR.

What do you think of the ambiguity about `zoom` and `move` events?

lucaswoj:
> Sure, I can rebase onto that and submit a new PR.

👍 Feel free to merge that one as soon as CI passes.

> What do you think of the ambiguity about zoom and move events?

I think the way you've handled it is good. We can make the implementation more consistent in a follow-up task.

lucaswoj:
I just made a PR to fix the inconsistency in the `DoubleClickZoomHandler`

davidtheclark:
Moved this to #2815.

PR #2760, opened 2016-06-17
Support querystring in non-mapbox sprite source

jingsam:


lucaswoj:
Thanks @jingsam! I'm happy to merge this with the addition of a unit test.

jingsam:
URL is more complex than we thought, here is the structure of a URL:
```
┌─────────────────────────────────────────────────────────────────────────────┐
│                                    href                                     │
├──────────┬┬───────────┬─────────────────┬───────────────────────────┬───────┤
│ protocol ││   auth    │      host       │           path            │ hash  │
│          ││           ├──────────┬──────┼──────────┬────────────────┤       │
│          ││           │ hostname │ port │ pathname │     search     │       │
│          ││           │          │      │          ├─┬──────────────┤       │
│          ││           │          │      │          │ │    query     │       │
"  http:   // user:pass @ host.com : 8080   /p/a/t/h  ?  query=string   #hash "
│          ││           │          │      │          │ │              │       │
└──────────┴┴───────────┴──────────┴──────┴──────────┴─┴──────────────┴───────┘
(all spaces in the "" line should be ignored -- they're purely for formatting)
```

We should be very careful when handling URLs. Using urlObject will be better than url string concating.

I checked the implementation in [mapbox.js](https://github.com/mapbox/mapbox-gl-js/blob/master/js/util/mapbox.js). There are still many url string concating. They may work for now. But many edge cases are not handling right.

So if you have plan to solve this problem, I will send a PR to refactor all url handling using urlObject, thanks.

lucaswoj:
Replaced by https://github.com/mapbox/mapbox-gl-js/pull/2772

PR #2771, opened 2016-06-20
Fix transitions for *-translate properties

### LINKED ISSUE #2762: Text translate is throwing an error
**mapbox-gl-js version**: 0.20.0

### Steps to Trigger Behavior

 1. updating text-translate value throws an error

### Expected Behavior

The text should move

### Actual Behavior

image

jfirebaugh:
Fixes #2762, see https://github.com/mapbox/mapbox-gl-js/issues/2762#issuecomment-227278278 for details.

Setting a 0-duration transition doesn't work for testing this bug, because it short circuits the code that was at fault. So in order for this to be testable in the test suite, we need to tackle https://github.com/mapbox/mapbox-gl-test-suite/issues/116. I'm going to look into that next.

### DIFF
--- a/js/style/style_declaration.js
+++ b/js/style/style_declaration.js
@@ -7,15 +7,13 @@ var util = require('../util/util');
 module.exports = StyleDeclaration;
 
 function StyleDeclaration(reference, value) {
-    this.type = reference.type;
-    this.transitionable = reference.transition;
     this.value = util.clone(value);
     this.isFunction = MapboxGLFunction.isFunctionDefinition(value);
 
     // immutable representation of value. used for comparison
     this.json = JSON.stringify(this.value);
 
-    var parsedValue = this.type === 'color' ? parseColor(this.value) : value;
+    var parsedValue = reference.type === 'color' ? parseColor(this.value) : value;
     this.calculate = MapboxGLFunction[reference.function || 'piecewise-constant'](parsedValue);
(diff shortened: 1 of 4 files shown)


jfirebaugh:
> So in order for this to be testable in the test suite, we need to tackle mapbox/mapbox-gl-test-suite#116. I'm going to look into that next.

This looks like it will require significant refactoring. There are a number of internals which use `Date.now()` directly that need to be converted to using a passed in parameter.

* https://github.com/mapbox/mapbox-gl-js/blob/062f52d836e8991f694c6742ab24ecfdc643151b/js/style/style_declaration.js#L53-L53
* https://github.com/mapbox/mapbox-gl-js/blob/8eaeebbcaeea012fe7876c06fb558a0a2ea99475/js/style/style_transition.js#L54-L54
* https://github.com/mapbox/mapbox-gl-js/blob/8eaeebbcaeea012fe7876c06fb558a0a2ea99475/js/style/style_layer.js#L311-L311
* https://github.com/mapbox/mapbox-gl-js/blob/1df146627092b58c8db64d45530c74cd12a3fa02/js/render/frame_history.js#L17-L17
* https://github.com/mapbox/mapbox-gl-js/blob/7840d9c2265c1d9be52a329d9a9c38784b75b3ca/js/style/style.js#L250-L257

jfirebaugh:
Cherry picked for v0.20.1 patch release.

PR #2772, opened 2016-06-21
normalizes non-mapbox:// scheme when query string exists

jingsam:


CONTRIBUTING.md
```diff
@@ -50,7 +50,7 @@ npm install
 Install [git](https://git-scm.com/), [node.js](https://nodejs.org/) (version 4 or greater), [GNU Make](http://www.gnu.org/software/make/), and libglew-dev
 ```bash
 sudo apt-get update &&
-sudo apt-get install build-essential git nodejs libglew-dev
+sudo apt-get install build-essential git nodejs libglew-dev libxi-dev
```
mikemorris: What's this for?

jingsam: libxi-dev is one of the dependencies. Wtihout it, `npm install` would fail

jingsam:
@lucaswoj

lucaswoj:
Fantastic! Thanks @jingsam.

jingsam:
I found that https://github.com/mapbox/mapbox-gl-js/blob/master/js/util/mapbox.js#L54 would be problematic when `url` has querystring.

As I pointed out previously, url is more complicated than we think. We should use urlObject than url string concatenating. But I found there are still many url string concatenating in `mapbox.js`.

So, I would like to refactor `mapbox.js`, any suggestions?

mikemorris:
We should take a look at the URL handling in mapbox-gl-native as well, I think there's a problem with non-mapbox:// sprite URLs, but I haven't been able to sufficiently investigate this yet.

Refs https://github.com/mapbox/mapbox-gl-native/issues/4922#issuecomment-225724657

PR #1466, opened 2016-06-21
WIP - support for non-mercator projections

anandthakker:
**Editing this comment because I stupidly made this PR by "converting" the original issue.  The post (and all comments) from the original issue are preserved below.**  ( --@anandthakker )

---

Steps in this quest:

- [x] move worldSize out of lngX, latY, xLng, and yLat methods upstream (e.g. to project/unproject)
- [x] make Transform (and Map) accept alternative implementations of these functions in constructor
- [x] generate a sample tileset in a different projection
- [x] load it up in GL JS and see if anything broke (**update:** spot checked using `debug/projection.html` and things seem to work...)
- [ ] more carefully assess impact throughout codebase, and guard with unit tests
- [ ] decide how to handle the `constrain` option (see https://github.com/mapbox/mapbox-gl-js/pull/1466#issuecomment-227871946 )
- [ ] similarly, decide on the correct behavior of `fitBounds`
- [ ] Include support in `geojson` sources (upstream: https://github.com/mapbox/geojson-vt/issues/57)

---
**original issue comment:**

Hey mapbox team! I first want to say I'm truly blown away by what you've built here. This is really, really impressive.

Anyhow, I'm curious to know if there is any interest in support non-mercator projections? I checked out the transformation logic and the projection appears to be hard-coded.

I'm sure there are optimizations to be had by assuming a single projection, but it would be really neat to be able to provide mapbox-gl with custom projection functions: not only would this enable other "common" map formats, but would open the door for some really creative visualizations (check out [this](https://www.jasondavies.com/maps/transition/) and [this](https://github.com/d3/d3-geo-projection)).

It would also make mapbox-gl an ideal candidate for non-geo-based maps, like a cross-section of a human body with organs, circulatory system, nervous system, etc.

Thanks again for all the cool tech, and for developing in the open!

Cheers,
Mike


anandthakker:
PSA: if you use the `hub` cli, don't use `hub pull-request -i #`, [because it's bad](https://github.com/github/hub/commit/4f70dd126f46dec14fc341c97c18efae417743c7)

anandthakker:
>  is it plausible that this could be limited to providing alternate implementations for lngX, latY, xLng, and yLat? Seems too good to be true...

Heh, this was indeed a bit simplistic: in web mercator, x is a function of lng only and y is a function of lat only, but this is _not_ true of other (most?) projections.  So, to make this work, we'll need to remove the independent methods `xLng`, `latY`, etc., and work directly off of `project` and `unproject`.  From a very initial look it seems like this may be possible

anandthakker:
The option to constrain the map to a lnglat bounding box also only makes sense in a projection like web mercator where lnglat rectangles are projected into rectangles

anandthakker:
I think it sorta works!  Check out `debug/projection.html` for a working example.

Updating the top of the ticket with a couple of additional tasks.

> The option to constrain the map to a lnglat bounding box also only makes sense in a projection like web mercator where lnglat rectangles are projected into rectangles

^ @mourner @tmcw any ideas on how best to handle this?

anandthakker:
> decide how to handle the constrain option (see https://github.com/mapbox/mapbox-gl-js/pull/1466#issuecomment-227871946 )
> similarly, decide on the correct behavior of fitBounds

Having mulled on this a bit, I think a fairly complete solution would be:
 - A custom projection may additionally provide a `projectedBoundingBox` function which takes a latlong polygon (maybe a rectangle, maybe something else) and returns a rectangle in the projected coordinate space that fully contains the projected version of the given polygon. 
 - If `projectedBoundingBox` is _not_ provided, then all gl-js methods that normally expect bounds in latlong coordinates simply fall back to interpreting the bounds as being in projected coordinates.

lucaswoj:
We've been thinking about changing the behaviour of `fitBounds` https://github.com/mapbox/mapbox-gl-js/issues/2112. It doesn't make much sense in web mercator either once pitching and rotating is introduced. 

> Having mulled on this a bit, I think a fairly complete solution would be:
> ...

This solution looks great! 👍

anandthakker:
@lucaswoj oh, shoot -- reading that ticket re: pitch/rotate made me realize that my proposal actually isn't complete after all: the `projectedBoundingBox` method I propose covers the "fitBounds" use case, but it doesn't cover the "maxBounds" use case, where, as I understand it, the intent is for the map view to be contained wholly within the given bounds (instead _covering_ the given bounds).  

So I guess we would need a `projectedInnerBox` method... but maybe it would be reasonable to proceed without this for now, especially since it sounds like in the long run there may be some revisions to these parts of the API.

mtirwin:
👀

lucaswoj:

anandthakker:
👍  I do plan to return to this and follow the plan outlined above, but it may be a couple weeks, so keeping closed until then SGTM

PR #2777, opened 2016-06-22
Fire mouseout events

alcoheca:
Allows developers to hide popups / clear highlighted areas which can clutter the screen after the user moves the mouse pointer out of context.

### DIFF
--- a/js/ui/bind_handlers.js
+++ b/js/ui/bind_handlers.js
@@ -26,6 +26,7 @@ module.exports = function bindHandlers(map, options) {
         }
     }
 
+    el.addEventListener('mouseout', onMouseOut, false);
     el.addEventListener('mousedown', onMouseDown, false);
     el.addEventListener('mouseup', onMouseUp, false);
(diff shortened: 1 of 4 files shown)


lucaswoj:
Great! Thanks @alcoheca!

js/ui/bind_handlers.js
```diff
@@ -37,6 +38,11 @@ module.exports = function bindHandlers(map, options) {
     el.addEventListener('dblclick', onDblClick, false);
     el.addEventListener('contextmenu', onContextMenu, false);
 
+    function onMouseOut(e) {
+        map.stop();
```
jfirebaugh: Let's remove this line. By default, animated zooms/pans should not stop if the cursor leaves the canvas.

alcoheca: sorry - I'll be honest i wasn't 100% about this line..

should I do anything?

alcoheca: see #2789

PR #2790, opened 2016-06-27
Browserify rejigger

### LINKED ISSUE #2767: minifyify should be a devDependency
`minifyify` appears to be needed only at build time and thus shouldn't be included in the runtime (production) dependencies. I'm guessing the same goes for `envify`, `unassertify` and `webworkify`.

lucaswoj:
fixes #2767 

cc @jfirebaugh @jscheid 

 - [x] move `minifyify` to `devDependencies`
 - [x]  move envifyify to `devDependencies`
 - [x] add `unassertify` to `package.json` browserify transforms, disable using `--ignore-transform=unassertify` during debug builds
 - [x] add build test that checks that assert statements are included in debug builds 
 - [x] speed up `npm test` by moving build tests into build scripts


server.js
```diff
@@ -8,6 +8,7 @@ var http = require('http');
 var path = require('path');
 
 app.get('/mapbox-gl.js', browserify('./js/mapbox-gl.js', {
+    ignoreTransform: ['unassertify'],
```
mourner: Would it be useful to leave assertions on the debug page?

lucaswoj: We are ignoring the unassertify transform, effectively leaving the assertions in the source code as-is. Double negatives are weird.

jfirebaugh:
👍 Awesome!

evocateur:
The improper `dependencies` fixed in this commit are causing errors during `npm shrinkwrap` for our project. When can we expect a version with this change to be published? Thanks!

lucaswoj:
@evocateur We hope to have a release out by the end of next week.

PR #2794, opened 2016-06-28
Fix edge case in Buffer#populatePaintArrays

anandthakker:
I believe this test expresses a potential bug / edge case in `populatePaintArrays` -- specifically, where an inner loop is [initialized to `startIndex`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/data/bucket.js#L340) not only for the first group, but also for subsequent groups.  If I'm understanding how this array group thing works, I'd expect something like:

```js
var startIndexForThisGroup = g === startGroup.index  ? startIndex : 0
for (var i =  startIndexForThisGroup; i < length; i++) 
```

### DIFF
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -42,11 +42,12 @@ Bucket.create = function(options) {
 Bucket.EXTENT = 8192;
 
 /**
- * The maximum size of the arrays in an "array group."
+ * The maximum size of a vertex array. This limit is imposed by WebGL's 16 bit
+ * addressing of vertex buffers.
  * @private
  * @readonly
  */
-Bucket.MAX_ARRAY_LENGTH = 65535;
+Bucket.MAX_VERTEX_ARRAY_LENGTH = Math.pow(2, 16) - 1;
 
 /**
  * The `Bucket` class is the single point of knowledge about turning vector
@@ -152,7 +153,7 @@ Bucket.prototype.prepareArrayGroup = function(programName, numVertices) {
     var groups = this.arrayGroups[programName];
     var currentGroup = groups.length && groups[groups.length - 1];
 
-    if (!currentGroup || currentGroup.layout.vertex.length + numVertices > Bucket.MAX_ARRAY_LENGTH) {
+    if (!currentGroup || currentGroup.layout.vertex.length + numVertices > Bucket.MAX_VERTEX_ARRAY_LENGTH) {
 
         var arrayTypes = this.arrayTypes[programName];
         var VertexArrayType = arrayTypes.layout.vertex;
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -42,6 +42,13 @@ Bucket.create = function(options) {
 Bucket.EXTENT = 8192;
 
 /**
+ * The maximum size of the arrays in an "array group."
+ * @private
+ * @readonly
+ */
+Bucket.MAX_ARRAY_LENGTH = 65535;
+
+/**
  * The `Bucket` class is the single point of knowledge about turning vector
  * tiles into WebGL buffers.
  *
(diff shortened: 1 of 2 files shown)


anandthakker:
> If I'm understanding how this array group thing works

^ that is a pretty big "if"

anandthakker:
Committing that proposed fix in case it actually is correct.

By the way, unfortunately that test takes quite a while, because it's checking > 65536 vertices.  I suppose it could be simplified down to checking just the first vertex of each group, though it seems like there may be some regression-proofing benefit to leaving it as is.

anandthakker:
Per conversation with @lucaswoj , f449709 updates the test to use two features with < 65536 vertices each, since this is a broader constraint in WebGL / gl-js.   I confirmed locally that the test still failed when I temporarily reverted the fix in cc779fc

test/js/data/fill_bucket.test.js
```diff
@@ -52,3 +52,74 @@ test('FillBucket', function(t) {
...
+                t.end();
+                return;
+            }
+        }
+    }
+
+    t.end();
+
+
+});
```
lucaswoj: > By the way, unfortunately that test takes quite a while, because it's checking > 65536 vertices. I suppose it could be simplified down to checking just the first vertex of each group, though it seems like there may be some regression-proofing benefit to leaving it as is.

This test takes 12 seconds to run on my machine 😬 

Would this be a good compromise?

 - test vertex 0
 - test vertex 65535 - 1
 - test vertex 65535

anandthakker: 👍 Yeah, agreed, we should really just stick to testing around the boundaries.    I'd add a fourth one:
 - test _last_ vertex (and I'll change the sizes so that this is very clearly distinct from vertex 65535, which we'll be testing because it's the first vertex of the new group)

anandthakker: @lucaswoj Alas -- it's actually the `addFeature` part during the setup that's costing most of the time :(

anandthakker:
@lucaswoj - improved the speed by temporarily reducing the max array size within the scope of that test.  Also realized while adding assertions for last-item-in-group-0 that my original test was partly broken: it assumed that a single polygon would be broken up across the vertex boundary, but in fact [polygons are always contained in a single group](https://github.com/mapbox/mapbox-gl-js/blob/master/js/data/bucket/fill_bucket.js#L82).

anandthakker:
Also, once again, I did double-check that removing the change in cc779fc causes the test (specifically, the 'first vertex of second group' assertion) to fail.  Should be good to go.

js/data/bucket.js
```diff
@@ -42,6 +42,13 @@ Bucket.create = function(options) {
 Bucket.EXTENT = 8192;
 
 /**
+ * The maximum size of the arrays in an "array group."
+ * @private
+ * @readonly
+ */
+Bucket.MAX_ARRAY_LENGTH = 65535;
```
lucaswoj: readability++

I amended the definition to include some extra context about this constant:
```
/**
 * The maximum size of a vertex array. This limit is imposed by WebGL's 16 bit addressing of vertex buffers. 
 * @private
 * @readonly
 */
Bucket.MAX_VERTEX_ARRAY_LENGTH = Math.pow(2, 16) - 1;
```

test/js/data/fill_bucket.test.js
```diff
@@ -16,7 +17,7 @@ var feature = vt.layers.water.feature(0);
 function createFeature(points) {
     return {
         loadGeometry: function() {
-            return [points];
+            return points[0] instanceof Point ? [points] : points;
```
lucaswoj: I would rather edit the existing test cases to include an outer array than introduce this logic into our test cases.

(Logic in test fixture generation leads to bugs in tests. Bugs in tests leads to hair-pulling.)

lucaswoj:
I just pushed a commit implementing my two suggestions. This is ready to :shipit: when CI passes.

PR #2804, opened 2016-06-29
Rename "Overlay" to "ElementMarker"

lucaswoj:
How do you feel about this rename @mourner? I’m ready to :ship: if you feel happy with it. 

cc @mourner @1ec5 @jfirebaugh

### DIFF
--- a/debug/markers.html
+++ b/debug/markers.html
@@ -47,7 +47,7 @@ function addRandomMarker() {
     var lng = w + (e - w) * Math.random();
     var lat = s + (n - s) * Math.random();
 
-    var marker = new mapboxgl.DOMMarker()
+    var marker = new mapboxgl.ElementMarker()
         .setLngLat([lng, lat])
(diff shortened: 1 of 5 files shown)


bhousel:
Can we please only use names like `DOMSomething` to refer to an actual `DOM` interface or object?   This is a well established naming convention.

1ec5:
If the idea is to emphasize that these markers can contain arbitrary HTML elements, then how about `MarkerElement`?

lucaswoj:
> Can we please only use names like `DOMSomething` to refer to an actual `DOM` interface or object?   This is a well established naming convention.

Good point. This wasn't on my radar.

> If the idea is to emphasize that these markers can contain arbitrary HTML elements, then how about `MarkerElement`?

Per Slack chat, I like `ElementMarker`. New commit coming.

lucaswoj:
🚢 on 🍏!

PR #2805, opened 2016-06-30
Add quadkey to url templates

mike-marcacci:
This adds the ability to fetch tiles from servers (like Bing) that use the quadkey format. Simply replaces `{quadkey}` in a URL template with the appropriate value. Test included.

### DIFF
--- a/js/source/tile_coord.js
+++ b/js/source/tile_coord.js
@@ -51,15 +51,26 @@ TileCoord.fromID = function(id) {
     return new TileCoord(z, x, y, w);
 };
 
+function getQuadkey(z, x, y) {
+    var quadkey = '', mask;
+    for (var i = z; i > 0; i--) {
+        mask = 1 << (i - 1);
+        quadkey += ((x & mask ? 1 : 0) + (y & mask ? 2 : 0));
+    }
+    return quadkey;
+}
+
 // given a list of urls, choose a url template and return a tile URL
 TileCoord.prototype.url = function(urls, sourceMaxZoom, scheme) {
     var bbox = WhooTS.getTileBBox(this.x, this.y, this.z);
+    var quadkey = getQuadkey(this.z, this.x, this.y);
 
     return urls[(this.x + this.y) % urls.length]
         .replace('{prefix}', (this.x % 16).toString(16) + (this.y % 16).toString(16))
         .replace('{z}', Math.min(this.z, sourceMaxZoom || this.z))
         .replace('{x}', this.x)
         .replace('{y}', scheme === 'tms' ? (Math.pow(2, this.z) - this.y - 1) : this.y)
+        .replace('{quadkey}', quadkey)
         .replace('{bbox-epsg-3857}', bbox);
 };
 
--- a/test/js/source/tile_coord.test.js
+++ b/test/js/source/tile_coord.test.js
@@ -48,6 +48,11 @@ test('TileCoord', function(t) {
             t.end();
         });
 
+        t.test('replaces {quadkey}', function(t) {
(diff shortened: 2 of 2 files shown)


lucaswoj:
This is great! Thank you for the contribution @mike-marcacci. 

Two things we should do before 🚢: 

 - [ ] Add some edge case tests for `getQuadkey` (for example `0,0,0`). Are there any existing test suites we can draw on?
 - [ ] Add a `quadkey.html` debug page for manual testing
 - [ ] Get a 👍 from @tmcw @bhousel and/or @jfirebaugh

mike-marcacci:
Ah yes, `0,0,0` is an interesting use case I haven't tested (Bing requires z >= 1) but definitely should for generic correctness. I'll explore this tonight.

As far as the `quadkey.html` debug file goes, would it be idiomatic to extend [access-token.js](https://github.com/mapbox/mapbox-gl-js/blob/master/debug/access-token.js) to look for a Bing API key?

lucaswoj:
> Ah yes, `0,0,0` is an interesting use case I haven't tested (Bing requires z >= 1) but definitely should for generic correctness. I'll explore this tonight.

Ah! My mistake! In that case, let's look at the z=1 case. My intent here is to exercise valid edge cases of `getQuadkey`

> As far as the `quadkey.html` debug file goes, would it be idiomatic to extend [access-token.js](https://github.com/mapbox/mapbox-gl-js/blob/master/debug/access-token.js) to look for a Bing API key?

I'd prefer to not complicate our system with Bing API keys out of a fear of TOS violations. If there's no free quadkey server, we can skip this step.

bhousel:
Ugh I was going to help by pointing you to the quadkey tests in iD, but we don't have any!  😢 

This project looks ok too and has good test coverage (maybe just use it?):
https://www.npmjs.com/package/quadkeytools

> I'd prefer to not complicate our system with Bing API keys out of a fear of TOS violations. If there's no free quadkey server, we can skip this step.

☝️ Agree... I think we can skip the debug page - a user would need a Bing key to make it work, and afaik there are no other providers that use it.

mike-marcacci:
That makes sense about the debug page; I could possibly get a key from a URL param if that sounds simpler...

I saw that NPM package, which looks really great, but felt like a bit overkill for the one function. Also, I was a bit confused by its use of `charCodeAt` and `fromCharCode` [in its implementation](https://bitbucket.org/steele/quadkeytools/src/5c5cd9b253223ec494db840ad50065f5a2915190/lib/index.js?at=master&fileviewer=file-view-default#index.js-205), but it does produce the same results [`getQuadkey(6, 29, 3) => 011123`](https://bitbucket.org/steele/quadkeytools/src/5c5cd9b253223ec494db840ad50065f5a2915190/test/quadkey.js?at=master&fileviewer=file-view-default#quadkey.js-56)


Also for reference:
- [Bing's explanation](https://msdn.microsoft.com/en-us/library/bb259689.aspx)
- [OpenStreetMap QuadTiles document](http://wiki.openstreetmap.org/wiki/QuadTiles)

lucaswoj:
I'm going to merge this as-is and then add some additional test cases in a follow-up PR. Thanks @mike-marcacci and @bhousel 🙇

lucaswoj:
Extra test cases in https://github.com/mapbox/mapbox-gl-js/commit/0877cac1d065a57c6ff2ce32f6b1201376ff8116 🎉

mike-marcacci:
Hi @lucaswoj - thanks so much for merging this! (Was off the grid this weekend for a change, so just catching up now.) And great call on the other tests!

1ec5:
mapbox/mapbox-gl-native#5628 ports these changes to the other Mapbox GL libraries.

Issue #2703, opened 2016-06-09
Implement cross-source symbol placement


mollymerp:
Prior 🎨 can be found at #2516 + #2624 + #1042 

Per chat in Slack today, we will be moving forward with a big 💪 refactor that will allow our cross-source symbol placement dreams to become a reality. 🌈

Here is a brain dump based on what I understand from @yhahn @jfirebaugh @ansis @lucaswoj:
Goal:
- change the `TilePyramid` - `Source` relationship so that instead of having many sources with their own `TilePyramid`s storing data on each `Tile` for that source, we have one `TilePyramid` and each `Tile` id would represent an array of `Tile` information for every source in the style.

Options:
- `TilePyramid` can dispatch a worker to `placeFeatures` after all tiles have been sent back to the main thread with their symbol placement info (from #2624)
  - this would allow current Worker behavior to remain the same (?) and we'd just have to add an additional Worker dispatch at the end for the final symbol placement (?)
- new `TilePyramid` would dispatch _all_ tiles under a given `id` in the `TilePyramid` to the same worker to compute placement information
  In chat, it seemed like the latter is preferred ( @jfirebaugh ?) If so, the work from #2624 was purely educational for me 😭

☝️  – please correct or edit as you see fit if I got something wrong

Questions that need answering (by me or anyone else):
- should the main thread handle symbol vs. non-symbol layers differently? Does it already?
- @jfirebaugh suggested a [quadtree](https://en.wikipedia.org/wiki/Quadtree) structure for the `TilePyramid`. What would that look like and what benefits does it offer?
- how will non-placement related `Worker` tasks be affected? @jfirebaugh says 👇
  > tile parsing + bucket creation could still be per-source-tile
- how much of this does the work @anandthakker is doing solve? How far away is that from shipping?

Future expansion of the `TilePyramid` refactor for rendering perf improvements (from slack mostly)
- combining tiles for rendering efficiency, for example 4x512 tiles into one mega tile (cc @jfirebaugh ) ?
- this refactor would mean that only one clipping mask is necessary. No need to redraw the clipping mask when drawing layers in different sources (-js) or use the more complicated and not completely sound approach -native does. (cc @ansis )

I plan on 💭 on the questions above to solidify my understanding of the task. Once we've settled on the approach we want to take going forward, I will put together a more detailed roadmap of the process and start a PR. Comments / thoughts welcome!!

anandthakker:
[...]

Issue #2432, opened 2016-04-12
Refactor Bucket and Buffer classes


lucaswoj:
- combine `Tile`, `Buffer`, and tile-specific parts of `Bucket` class into one class
- factor non-tile-specific parts of `Bucket` class into another class

jfirebaugh:
⇢ #2875

Issue #281, opened 2014-02-07
Support custom layer types powered by WebGL 


edenh:
@ansis and I have discussed the possibility of allowing people to write custom glsl shaders. While the learning curve is fairly steep, the possibilities are pretty mindblowing when considering integrated data vis and animated, interactive basemaps.

Checkout these shaders for inspiration:
https://www.shadertoy.com/view/MsXGRf
https://www.shadertoy.com/view/Msl3Rr
https://www.shadertoy.com/view/XdsGDB
http://glsl.heroku.com/e#13858.13
http://glsl.heroku.com/e#13880.2

I'm not really sure exactly how we would accomplish this, but hoping for discussion here.

cc @mourner @kkaefer @yhahn

jfirebaugh:
No current plans to do this.

enjalot:
I know this is closed but I'd love to see it for things like controlling saturation or other levels on a raster layer.
custom shaders on a layer could potentially enable arbitrary clipping via mask images which would be fun too.

lucaswoj:
This is worth leaving open, now that we "support" this in gl-native

mourner:
> things like controlling saturation or other levels on a raster layer
@enjalot can't you control saturation and other stuff with [raster paint properties](https://www.mapbox.com/mapbox-gl-style-spec/#paint-raster-opacity)?

lucaswoj:
As I said in #1742, this would be less scary if we had a GL state tracking abstraction #145

enjalot:
@mourner oops I didn't see the other levels besides opacity before. I still believe that this could lead to very interesting customizations! depth of field, blurs, animations and a lot of other shiny stuff :100:

lucaswoj:
Here's our plan for implementing this feature: 

- There will be a new style layer type called "custom"
- Developers can register a rendering callback for each custom layer
- When a custom layer is drawn, the callback is called and passed the WebGL context within in a declarative state-tracking wrapper a la https://github.com/mapbox/mapbox-gl-js/issues/145
- The callback may draw directly to the WebGL framebuffer (higher performance, must use declarative state-tracking wrapper) or copy content from another framebuffer (lower performance, no restrictions on rendering)
- Custom layers must use the GL state tracker correctly or else the map may not render correctly or the application may stop working

lucaswoj:
Based on the experience we've had with custom layer types on GL Native and custom sources in GL JS, I'm proposing a new specification:

[...]

PR #2812, opened 2016-07-01
Bucket enhancements to allow property-only update

anandthakker:
@lucaswoj I've been experimenting a bit with the "join new properties to existing geometries" problem.  As you've suggested when we've discussed this, a couple of key enhancements to `Bucket` will be needed:
- Track a mapping from feature index to vertex array index range -- this would allow a source to know which parts of the paint-property-related vertex arrays to update for which features.
- Accept optional end indexes in `populatePaintArrays` -- currently, this method only takes a starting index and assumes that it should update from that index to the end of the current arrays, which makes it only suitable for populating the arrays the first time.  Adding an end index parameter would make it straightforward to make updates to an already-populated array.

This PR includes draft versions of those changes, along with:
- A proof of concept `updateProperties` method, which I stuck onto `WorkerTile` just for demo/prototyping purposes (but which should in reality probably go into a custom source type that encapsulates the data-joining/updating)
- A pair of benchmarks -- `dds_parse_bucket` and `dds_update_bucket` -- to get a quick-and-dirty estimate of the perf gains from just updating versus a full parse.  Results are pretty good at time of posting this = approx 43% gain:

screen shot 2016-07-01 at 10 52 31 am

screen shot 2016-07-01 at 10 53 08 am


js/data/bucket.js
```diff
@@ -327,16 +345,21 @@ Bucket.prototype.recalculateStyleLayers = function() {
...
+        endGroupIndex = endGroupIndex || (groups.length - 1);
+
+        for (var g = startGroupIndex; g <= endGroupIndex; g++) {
             var group = groups[g];
             var length = group.layout.vertex.length;
             var vertexArray = group.paint[layer.id];
             vertexArray.resize(length);
 
+            var start = g === startGroupIndex ? startVertexIndex : 0;
+            var end = (endVertexIndex && g === endGroupIndex) ? endVertexIndex : length - 1;
```
anandthakker: With this method now accepting ending indexes, maybe it would be simpler _require_ the end index params and move the responsibility for choosing `group.layout.vertex.length - 1` out to the callers?

lucaswoj: That makes sense 👍

mtirwin:
👀

mourner:
Nice! What's up with the build failure — needs test updates as well?

anandthakker:
@mourner Yep, tests needed small tweak.  Just fixed in a556c9760cbcc87fc7fab87b13bfbe5b2f0a027c along with the change to populatePaintArrays [mentioned above](https://github.com/mapbox/mapbox-gl-js/pull/2812#discussion_r69318637)

anandthakker:
Ah, looks like some rendering tests are still broken.  Will check those out soon.

anandthakker:
@lucaswoj @mourner ^ the last few commits clean things up as follows:
 - Accept an "array range" object (`{startGroup, startVertex, endGroup, endVertex}`) rather than individual indexes in `populatePaintArrays`.
 - Rename `_featureIndexToArrayIndex` => `featureIndexToArrayRange`; `updateBuffers` => `updatePaintArrays`
 - Contain `_featureIndexToArrayRange` wholly within `bucket.js`, by moving the responsibility for updating it into `populatePaintArrays`, because that seems like the most natural common code path for various bucket types to add paint data for a single feature.
 - Fix tests, and add a simple test for `updatePaintArrays`

anandthakker:
Removing "WIP" from the title now -- let me know if you think the changes in `*bucket*.js` look okay, and if so I will revert the temporary changes to `worker_tile` to make it mergeable

js/source/worker_tile.js
```diff
@@ -232,6 +232,55 @@ WorkerTile.prototype.parse = function(data, layerFamilies, actor, rawTileData, c
...
+// pretty independent of the rest of the code here, except that it reuses
+// isBucketEmpty and serializeBucket.
+//
+// Params analogous to WorkerTile#parse, except that `data`, rather than being
+// a `VectorTile`, is an object like:
+// {
+//     source-layer-name: [{ /* feature 0 properties */}, {/* feature 1 properties */}, ...],
+//     another-layer-name: [...]
+// }
+// where feature 0, 1, 2 are the features, in order, in this vector tile.
```
mourner: Should this comment be reformatted into JSDoc like rest of the code, while moving the note to a ticket or GH comment? It looks a bit like a TODO.

anandthakker: Ah @mourner sorry I wasn't clear in the comment--my intention was actually to revert this change entirely, excluding it from the PR and keeping the changes scoped to the Bucket classes.  That's what I meant by "this might make more sense as part of a custom source type".

I guess a downside to that is that the `dds_update_bucket` benchmark would also have to go.

If yall think it's worth keeping this here, though, I'd be happy to clean it up (including the comment, as you said), and also to add some tests for it.

anandthakker:
In preparation for a rebase, preserving some inline comments:

@mourner commented about fixing/removing the TODO-like comment above `WorkerTile#updateProperties`, and I responded:

>  sorry I wasn't clear in the comment--my intention was actually to revert this change entirely, excluding it from the PR and keeping the changes scoped to the Bucket classes. That's what I meant by "this might make more sense as part of a custom source type".
>
> I guess a downside to that is that the dds_update_bucket benchmark would also have to go.
> 
> If yall think it's worth keeping this here, though, I'd be happy to clean it up (including the comment, as you said), and also to add some tests for it.

cc @lucaswoj for you to weigh in

anandthakker:
Rebased (onto **`arrays-and-buffers`** branch, **NOT `master`**, so this PR is downstream of #2837 ). Previously branch was at 1f04efcfc3a7d8a4809397e411fb7b7743677aa2

In other news, I guess rebasing _doesn't_ mean losing the inline diff comments -- how nice!

anandthakker:
I'm finding that there are some problems here relating to the arrays getting "transferred" from the worker back to the main thread after the initial load/parse:

Transferring apparently leaves the array "neutered" in the worker thread, so that it can't be updated.  Moreover, when the bucket is reconstructed on the main thread, the transferred/serialized `arrayGroups` is currently not being unserialized and reinstated -- rather, it's only used to create `bufferGroups`.

mourner:
> Transferring apparently leaves the array "neutered" in the worker thread

Yeah, that's the point of transferring — moving data fast between threads without copy, but at the expense of not being able to access it after it was moved.

anandthakker:
Yeah, makes sense. I'm thinking I'll play with either (a) transferring the
arraybuffers *back* to the worker for the property update task or (b) doing
the property update on the main thread.

I think either of these will require reconstructing the "arrayGroups"
object somehow -- for (a) I think I would have to replace the neutered
array buffers w the ones that I transferred back; for (b) I'd need to
create arrayGroups from the serialized arrays, similar to how
"arrayBuffers" is created, so that populatePaintArrays can be used.

anandthakker:
I went with option (a), implemented in 5cfc637, and it works!  Demo:

dynamic-data

Those points are coming from a vector tile source; `update()` is hardcoded to send in random values for the dds property that's controlling circle-radius.   (Code [here](https://github.com/anandthakker/mapbox-gl-js/commit/ae676860a08ad97505974612d739feddfe74522c) -- relies on both this PR and https://github.com/mapbox/mapbox-gl-js/pull/2667)

js/data/bucket.js
```diff
@@ -344,16 +399,34 @@ Bucket.prototype.recalculateStyleLayers = function() {
...
+ * @property {number} arrayRange.startGroup
+ * @property {number} arrayRange.startVertex
+ * @property {number} arrayRange.endGroup
+ * @property {number} arrayRange.endVertex
+ * @private
+ */
+
+Bucket.prototype.populatePaintArrays = function(interfaceName, globalProperties, featureProperties, arrayRange, featureIndex) {
+    if (typeof featureIndex !== 'undefined') {
+        this._featureIndexToArrayRange[featureIndex] = arrayRange;
```
anandthakker: It occurs to me that this map should actually be `this._featureIndexToArrayRange[interfaceName][featureIndex]`

js/data/bucket.js
```diff
@@ -139,6 +139,21 @@ Bucket.prototype.populateArrays = function() {
 };
 
 /**
+ * Update paint arrays with the given feature properties, leaving geometries
+ * as-is.
+ * @private
+ */
+Bucket.prototype.updatePaintArrays = function(interfaceName, propertiesList) {
```
anandthakker: Add a note that the propertiesList is expected to correspond to the source tile's features, _not_ to the bucket's filtered list of features.

anandthakker:
@lucaswoj Soon I will rebase to bring this up to date with the refactor in https://github.com/mapbox/mapbox-gl-js/pull/2865; before I do, would love to get your 👍 / 👎 on each of the following changes that are currently included here:

1.  Updates `populatePaintArrays` to (a) require both starting and ending indexes (in a combined `arrayRange` param), and (b) if a feature index is provided, update the bucket's feature-index-to-array-range mapping.
2. Add `Bucket#updatePaintArrays` which, given an array of feature properties, uses the feature-index-to-array-range map to correctly call `populatePaintArrays` for each of the new properties objects
3. Add `Bucket#setArrayGroupData` to allow a worker-side bucket to update its ArrayBuffer references after they've been transferred to the main thread (and thus neutered) and then transferred back. 
4. Add `WorkerTile#updateProperties`, supported by items 1-3 above.

Number 4 raises a broader question: originally, I only included it so that yall could check out the benchmarks and such, assuming that in the end we would remove it and leave that logic as the responsibility of the dynamic-property-updating custom source implementation.  But this does require that custom to reach pretty far into the GL JS's implementation details.  Is this okay?  Would it be better to perhaps include `WorkerTile#updateProperties` as part of the public api?  Or even to include the dynamic-updating source as a core source type?

lucaswoj:
> 1.  Updates `populatePaintArrays` to (a) require both starting and ending indexes (in a combined `arrayRange` param), and (b) if a feature index is provided, update the bucket's feature-index-to-array-range mapping.

Sounds good 👍 

 > 2\. Add `Bucket#updatePaintArrays` which, given an array of feature properties, uses the feature-index-to-array-range map to correctly call `populatePaintArrays` for each of the new properties objects

Sounds good 👍 

> 3\. Add `Bucket#setArrayGroupData` to allow a worker-side bucket to update its ArrayBuffer references after they've been transferred to the main thread (and thus neutered) and then transferred back. 

> ```
> /**
> + * Sets the underlying `arrayBuffer` instances for this bucket's array groups.
> + * When the arrayBuffers are transferred across threads with postMessage, the
> + * references to them on the sending thread become "neutered". This method
> + * allows for those arrayBuffers to be sent _back_ to the worker they came
> + * from.
> ```

This sounds potentially problematic. What happens if the main thread tries to render the bucket while its arrays are "missing"? 

Is it possible to create a fresh paint array with the new data without transferring anything from the main thread back to the worker? I can't think of any reason why we'd need to read from the original paint arrays or why we have to use the same `ArrayBuffer` instance for the new paint array.

 > 4\. Add `WorkerTile#updateProperties`, supported by items 1-3 above.

> Number 4 raises a broader question: originally, I only included it so that yall could check out the benchmarks and such, assuming that in the end we would remove it and leave that logic as the responsibility of the dynamic-property-updating custom source implementation.  But this does require that custom to reach pretty far into the GL JS's implementation details.  Is this okay?  Would it be better to perhaps include `WorkerTile#updateProperties` as part of the public api?  Or even to include the dynamic-updating source as a core source type?

I would prefer to refrain from making this a public API. You may add this code and use it for your vector tile property updater but we do not plan to provide support for or make any external promises about the stability of this API.

anandthakker:
Thanks @lucaswoj !

> This sounds potentially problematic. What happens if the main thread tries to render the bucket while its arrays are "missing"?

Doh!  I hadn't thought of this.  

> Is it possible to create a fresh paint array with the new data without transferring anything from the main thread back to the worker?

Yeah, it does seem like this should be possible.  Since we'd be talking about creating new paint arrays but _not_ the corresponding "layout" arrays, doing this might involve some tweaks in a couple places, but pretty minor, I think.

> I would prefer to refrain from making this a public API. You may add this code and use it for your vector tile property updater but we do not plan to provide support for or make any external promises about the stability of this API.

👍 this makes sense and sounds reasonable to me.

anandthakker:
Rebased  (previous: 5cfc63730d0160127841aa0fd0e60b0fff680b6f)

> Is it possible to create a fresh paint array with the new data without transferring anything from the main thread back to the worker? I can't think of any reason why we'd need to read from the original paint arrays or why we have to use the same ArrayBuffer instance for the new paint array.

Also updated to reflect this suggestion, removing `Bucket#setArrayGroupData` in favor of `Bucket#updatePaintBufferData`, which, on the main-thread side, allows for updating with newly-created paint arrays.

Next steps:
 - [ ] Add unit tests to cover new methods in Bucket, BufferGroup, and WorkerTile
 - [ ] Fix benchmarks

anandthakker:
> Add unit tests to cover new methods in Bucket, BufferGroup, and WorkerTile

The tests now in place for Bucket#updatePaintVertexArrays, Bucket#updatePaintVertexBuffers, and WorkerTile#updateProperties should have this covered.

anandthakker:
@lucaswoj - okay, I think this is now ready for a review.  Key changes described in the commit message for d085e67 (but note the method renaming that happened since then).

js/data/bucket.js
```diff
@@ -109,6 +109,35 @@ Bucket.prototype.populateArrays = function() {
...
+ * @private
+ */
+Bucket.prototype.updatePaintVertexArrays = function(programName, propertiesList) {
+    this.recalculateStyleLayers();
+
+    for (var i = 0; i < propertiesList.length; i++) {
+        var range = this._featureIndexToArrayRange[i];
+        if (!range) continue;
+        var numVertices = range.endVertex - range.startVertex + (range.endGroup - range.startGroup) * ArrayGroup.MAX_VERTEX_ARRAY_LENGTH;
+        this.prepareArrayGroup(programName, numVertices);
```
lucaswoj: `Bucket#prepareArrayGroups` is not yet designed to handle the case where `numVerticies > ArrayGroup.MAX_VERTEX_ARRAY_LENGTH`. We may need to fundamentally rethink that method to elegantly handle this case.

lucaswoj: It might be smart to refactor into some smaller, sharper tools like `Bucket.vertexCapacity` and `Bucket.resize()`

js/data/bucket.js
```diff
@@ -109,6 +109,35 @@ Bucket.prototype.populateArrays = function() {
...
+        // Maintain the ArrayGroup invariant that layoutVertexArray.length ===
+        // paintVertexArrays[*].length, necessary for proper operation of
+        // prepareArrayGroup
+        for (var g = range.startGroup; g <= range.endGroup; g++) {
+            var start = g === range.startGroup ? range.startVertex : 0;
+            var end = g === range.endGroup ? range.endVertex : 0;
+            for (var v = start; v <= end; v++) {
+                this.arrayGroups[programName][g].layoutVertexArray.emplaceBack(0, 0);
+            }
+        }
```
lucaswoj: I'm confused by this code. Why isn't this invariant being met? Can we meet this invariant by imposing some invariants on `propertiesList` instead?

js/data/bucket.js
```diff
@@ -109,6 +109,35 @@ Bucket.prototype.populateArrays = function() {
 };
 
 /**
+ * Update paint arrays with the given feature properties, leaving geometries
+ * as-is.
+ * @private
+ */
+Bucket.prototype.updatePaintVertexArrays = function(programName, propertiesList) {
```
lucaswoj: How can we be sure that the feature order in `propertiesList` matches the feature order in the existing `StructArray`s?

js/data/bucket/circle_bucket.js
```diff
@@ -116,5 +116,18 @@ CircleBucket.prototype.addFeature = function(feature) {
...
     }
 
-    this.populatePaintArrays('circle', globalProperties, feature.properties, startGroup, startIndex);
+    var groups = this.arrayGroups['circle'];
+    var range = {
+        startGroup: startGroup.index,
+        startVertex: startIndex,
+        endGroup: groups.length - 1,
+        endVertex: groups[groups.length - 1].layoutVertexArray.length - 1
+    };
```
lucaswoj: Do you think it would be simpler to combine the group index and the vertex index into a single number? 

 - group index is `Math.floor(index / ArrayGroup.MAX_VERTEX_ARRAY_LENGTH)`
 - vertex index is `index % ArrayGroup.MAX_VERTEX_ARRAY_LENGTH`

lucaswoj: This may allow us to simplify `_featureIndexToArrayRange` to a simple `Array<number>`

bench/benchmarks/dds_update_bucket.js
```diff
@@ -0,0 +1,241 @@
...
+function createLayerFamilies(layers) {
+    if (layers !== createLayerFamiliesCacheKey) {
+        var worker = new Worker({addEventListener: function() {} });
+        worker['set layers'](layers);
+
+        createLayerFamiliesCacheKey = layers;
+        createLayerFamiliesCacheValue = worker.layerFamilies;
+    }
+    return createLayerFamiliesCacheValue;
+}
```
lucaswoj: There's a large amount of redundant code between these two benchmarks and the buffer benchmark. Can you abstract that out?

js/data/bucket.js
anandthakker: Ah, I think I see what you mean: this wasn't an issue before, because `prepareArrayGroup` only ever needed to make enough room for one feature, whereas here, it would need to make enough room for all the features in the bucket.

👍 I'll shoot for the revision you proposed here.

anandthakker: Oh, hold on -- actually, this call to `prepareArrayGroup` is happening once per feature here, which I think makes it quite a bit less likely that we would exceed `ArrayGroup.MAX_VERTEX_ARRAY_LENGTH`.  (Consider, e.g., https://github.com/mapbox/mapbox-gl-js/blob/master/js/data/bucket/fill_bucket.js#L79 -- although it's true that that is on an individual polygon ring rather than the whole feature.)

js/data/bucket.js
anandthakker: @lucaswoj Without this code, the invariant in question would not be met here because this method, `updatePaintVertexArrays`, would then only be resizing and populating the _paint_ arrays, and not the _layout_ arrays.    (The layout arrays are otherwise only sized/populated during `addFeature`.)

The only reason this invariant even matters here, however, is because it is assumed by the `prepareArrayGroup` discussed above.  The right refactor there could eliminate the need to do this here.

js/data/bucket.js
anandthakker: I figured this could be left as the caller's responsibility.

The reason I think this is somewhat reasonable is that `propertiesList` is expected to correspond to the _pre-filtered_ list of features in this tile.  I thought that the ordering of the features in a given vector tile would be stable and safe to rely on.

js/data/bucket/circle_bucket.js
anandthakker: Good point!  That would be a nice simplification, and might even have a perf upside.

js/source/worker_tile.js
```diff
@@ -108,9 +108,9 @@ WorkerTile.prototype.parse = function(data, layerFamilies, actor, rawTileData, c
         }
     }
 
-    var buckets = [],
+    var buckets = this.buckets = [],
```
anandthakker: Agreed.  I think [this line](https://github.com/mapbox/mapbox-gl-js/blob/96c010c3c4b2e87afcc669275e32b1a458f42f56/js/source/worker_tile.js#L195) must certainly help with that, but we should also investigate whether the arraybuffers continue to consume memory after they're transferred/neutered.  If not, or if there's a straightforward way to drop them after we've made the transfer, then probably the memory impact here is minimal?

However, if not:
 - I _think_ we could hone this down to only storing the feature-index->array-range mapping for each bucket, although that would require us to recreate the Bucket objects in `updateProperties`, and I would imagine there would be a few wrinkles there.
 - Alternatively, we could _not_ hold onto `this.buckets` by default, but only do so if a flag is set in the params

js/data/bucket.js
lucaswoj: 👍  Just wanted to make sure you had thought about this

js/source/worker_tile.js
lucaswoj: > investigate whether the arraybuffers continue to consume memory after they're transferred/neutered

Ah! I didn't think about this. The memory impact should be minimal thanks to the way transferables work. #2432

anandthakker:
@lucaswoj check out https://github.com/mapbox/mapbox-gl-js/pull/2812/commits/b038472be9960e72909e1f7b1528f77587968944 -- when I first started that update, I had ambitions of eliminating `prepareArrayGroup` even in preexisting places, but now I think that would need to be part of a bigger/different refactor.  For the purposes of this PR, I've extracted the primitives `Bucker#vertexArrayCapacity`, `Bucket#vertexArrayLength`, and `Bucket#resizeVertexArrays`, and, in the process, switched to using single, larger indexes for the feature->vertex index mapping.

anandthakker:
Memory check (see https://github.com/mapbox/mapbox-gl-js/pull/2812#discussion_r73260850)
 - mod the buffer benchmark by (a) pushing each WorkerTile in an array referenced from `window`; (b) creating a dummy web worker and using `postMessage(parsedData, buffers)` to simulate transfer back to the main thread; (c) waiting for a click after preloading assets.
 - in Chrome timeline, force GC, start recording, and click the window to run the benchmark (loads 1500 tiles)
 - force GC again and stop recording and compare before/after heap size.

| test | heap delta (MB) |
|------|-----------------|
| (sanity check) retain `this.buckets`, DON'T clear out `bucket.features` | 100 |
| (this branch) retain `this.buckets`, DO clear out `bucket.features` | 70 |
| (master) DO NOT retain `this.buckets` | 30 |

anandthakker:
Update: ce77b06 unreferences the buckets' array groups after they've been serialized at the end of `WorkerTile#parse`.  This gets us down to a 60 MB heap delta in the test above.  Not immediately sure where the extra 30 MB (~20K per WorkerTile) is coming from.

anandthakker:
Okay--I'm fairly certain that the > 90% of the delta in WorkerTile size between this branch and `master` is accounted for by `_featureIndexToArrayRange`:

screen shot 2016-08-08 at 3 13 20 pm

The last column is the retained size, and you can see that FillBucket objects account for 20%, and in an individual FillBucket, the _featureIndexToArrayIndex accounts for almost all of it.

The actual numbers seem roughly plausible: in the tile shown, where `_featureIndexToArrayRange` weighs ~ 30KB, there are 3188 features: that's about 9.4 bytes per feature, which sounds reasonable for two integer indexes each + overhead.

@lucaswoj how do you feel about this amount of additional memory usage?  After the latest commit, we're at about 13-14KB per WorkerTile.  (If this seems problematic, then I would propose that we hide the feature index->array index tracking behind a parameter to `load tile`, so that a custom source can switch it on.)

lucaswoj:
> how do you feel about this amount of additional memory usage? After the latest commit, we're at about 13-14KB per WorkerTile.

That's not too bad! Thanks for looking into it.

lucaswoj:
I'm feeling good about the architecture of this code but two lingering high-level questions:

 - Is `_featureIndexToArrayRange` dead code unless we include the property-updating-vector-tile-source in core?
 - Is there some way we can push the creation of `_featureIndexToArrayRange` downstream into the custom source type?

js/data/array_group.js
```diff
@@ -50,6 +50,10 @@ ArrayGroup.prototype.isEmpty = function() {
     return this.layoutVertexArray.length === 0;
 };
 
+ArrayGroup.prototype.vertexArrayLength = function() {
```
lucaswoj: `vertexArrayLength` -> `getVertexArrayLength`

js/data/bucket.js
```diff
@@ -109,35 +109,115 @@ Bucket.prototype.populateArrays = function() {
...
+/**
+ * Returns the length of the vertex arrays for the given programName.
+ *
+ * Note: the vertex arrays actually consist of a sequence of ArrayGroups, each
+ * with a max length of 65535.  Since each group is not necessarily filled to
+ * capacity, we effectively have 'gaps', meaning that vertexArrayLength() is
+ * generally greater than the literal number of vertices represented.
+ * @private
+ */
+Bucket.prototype.vertexArrayLength = function(programName) {
```
lucaswoj: `vertexArrayLength` -> `getVertexArrayLength`

js/data/bucket.js
```diff
@@ -109,35 +109,115 @@ Bucket.prototype.populateArrays = function() {
...
+    if (!groups.length) return 0;
+    return (groups.length - 1) * ArrayGroup.MAX_VERTEX_ARRAY_LENGTH +
+        groups[groups.length - 1].vertexArrayLength();
+};
+
+/**
+ * Returns the total capacity of the vertex arrays for the given programName.
+ * @private
+ */
+Bucket.prototype.vertexArrayCapacity = function(programName) {
```
lucaswoj: `vertexArrayCapacity` -> `getVertexArrayCapacity`

bench/benchmarks/dds_update_bucket.js
```diff
@@ -0,0 +1,102 @@
...
+            angle: 0,
+            pitch: 0,
+            showCollisionBoxes: false,
+            source: 'composite',
+            uid: url
+        });
+    }
+    return workerTiles[url];
+}
+
```
lucaswoj: I'm unsure if the `dds_parse_bucket` and `dds_update_bucket` benchmarks [operate in a way that mimics real-world usage](http://github.com/mapbox/mapbox-gl-js/blob/f287983ecf88ee4df6ad0ee16144de02d93477ae/bench/README.md#L22) enough to be merged into `master` and run on every PR from now on. I'm really really happy that you were looking at these perf #s in testing will hang onto them for future reference no matter what!

lucaswoj: (As always, open to discussion on ☝️)

bench/benchmarks/worker_tile_memory.js
```diff
@@ -0,0 +1,83 @@
...
+        if (err) throw err;
+        var data = new VT.VectorTile(new Protobuf(new Uint8Array(response)));
+        workerTile.parse(data, assets.layerFamilies, actor, null, function(err, data, buffers) {
+            if (err) return callback(err);
+            w.postMessage(data, buffers.filter(Boolean));
+            callback();
+        });
+    });
+}
+
```
lucaswoj: I am interested in creating a memory usage benchmark #2947 but don't know if it's worth the hassle of having one that requires manual intervention, given that this will need to be run on every non-trivial PR from now on.

js/data/bucket/fill_bucket.js
```diff
@@ -60,14 +60,12 @@ FillBucket.prototype.addFeature = function(feature) {
...
-    var startGroup = this.prepareArrayGroup('fill', 0);
-    var startIndex = startGroup.layoutVertexArray.length;
-
+    var startVertex = this.vertexArrayLength('fill');
     for (var i = 0; i < polygons.length; i++) {
         this.addPolygon(polygons[i]);
     }
-
-    this.populatePaintArrays('fill', {zoom: this.zoom}, feature.properties, startGroup, startIndex);
+    var endVertex = this.vertexArrayLength('fill') - 1;
```
lucaswoj: Looks like we need to do `endVertex - 1` everywhere. Does this have to do with `Bucket#populatePaintArrays`'s `endIndex` being exclusive?

bench/benchmarks/dds_update_bucket.js
anandthakker: I definitely agree in principle with the "mimics real-world usage" criterion.  Couple thoughts in this particular case:

`dds_parse_bucket` differs from the `buffer` benchmark only by:
```js
>         stylesheet.layers.forEach(function (layer) {
>             if (layer.type === 'fill' && layer.paint) {
>                 layer.paint['fill-color'] = {
>                     property: 'level',
>                     stops: [[0, 'white'], [100, 'blue']]
>                 };
>             }
>         });
```

Both do basically the same thing: for "the set of coordinates of tiles used in a "flyTo" from the Mapbox SF office to the Mapbox DC office," run `WorkerTile#parse` on each one.

I think it's worth keeping `dds_parse_bucket`, because:
 - by virtue of having a dds property, the `Bucket#addFeature` calls that are a crucial part of `WorkerTile#parse` are forced to actually populate the paint vertex arrays; this happens in real-world usage, and so it's good to have it covered in a benchmark.
 - moreover, although dds is relatively new, I think it's a pretty good bet that it will become _very_ common, especially once it's supported in Studio; thus, arguably, this might ultimately be mimicing real-world usage better than the `buffer` benchmark does.

As for `dds_update_bucket`, it is again pretty similar to the `buffer` benchmark, except that it focuses on `WorkerTile#updateProperties`.  IMO, we should keep this one only if we're going to bring the dynamic-updating VT source into core; otherwise, it probably better belongs in that repo.  ( 💭 I wonder if there's any good way for a third-party repo to leverage the benchmark infrastructure from here...)

bench/benchmarks/worker_tile_memory.js
anandthakker: Agreed.  Meant to remove this and forgot!

anandthakker:
> Is _featureIndexToArrayRange dead code unless we include the property-updating-vector-tile-source in core?

At least in the short-term, yes.  It's worth noting, though, that if we start to dig further into caching tiles on a per-StyleLayer basis ( https://github.com/mapbox/mapbox-gl-js/issues/2951 https://github.com/mapbox/mapbox-gl-js/issues/2875  https://github.com/mapbox/mapbox-gl-js/issues/2703#issuecomment-238448237 ), I think this--and the `Bucket#updatePaintVertexArrays` that it enables--would quickly become useful in core.

> Is there some way we can push the creation of _featureIndexToArrayRange downstream into the custom source type?

This might be possible if the dynamic-updating custom source type essentially forks `WorkerTile`, and then in `WorkerTile#parse`, instead of calling `Bucket#populateArrays`, it could basically do the same work 'manually', saving before/after indexes as it goes:

```js
    bucket.createArrays();
    bucket._featureIndexToArrayRange = [] // <-- added
    bucket.recalculateStyleLayers();

    for (var i = 0; i < bucket.features.length; i++) {
        var index = this.features[i].index // <-- added
        bucket._featureIndexToArrayRange[index] = bucket.getVertexArrayLength(); // <-- added
        bucket.addFeature(this.features[i]);
        bucket._featureIndexToArrayRange[index] = bucket.getVertexArrayLength(); // <-- added
    }
    bucket.trimArrays();
```

The upside here is clear: not adding dead code into core that needs to be supported.  The downside is that forking the whole WorkerTile and hardcoding a variation on the deeply internal method `populateArrays` would make the dynamic-updating source type a quite bit harder to support, esp. with respect to keeping up with bugfixes and improvements that happen in core.

So the options are:
1. keep _featureIndexToArrayIndex as dead code in core to (a) support an external dynamic updating custom source, and (b) possibly enable tricksy array-updating logic in core.
2. keep _featureIndexToArrayIndex in core and also put the dynamic updating source on the roadmap for core.
3. push _featureIndexToArrayIndex downstream

FWIW, I think I would lean towards option 2, despite a general preference for always keeping any 'core' library as small as possible.  In my mind, I have always thought of the dynamic-updating source more of an enhancement to the VectorTile source than a separate source type... and maybe even as an analogous enhancement to GeoJSONSource, too?

lucaswoj:
I don't think we can accept anything other than option 3. 

Our goal is to create a framework that allows users to add custom source types without modifying GL JS core. We should view `_featureIndexToArrayRange` through that lens. How can we refactor to enable a feature like `_featureIndexToArrayRange` to be implemented by a custom source type? There will be more custom source types that will need similarly deep access into core functionality.  Such access must not regress performance or create dead code.

lucaswoj:
I am removing this from the Barcelona milestone so that @jfirebaugh may have an opportunity to review this before it merges.

anandthakker:
> How can we refactor to enable a feature like _featureIndexToArrayRange to be implemented by a custom source type? There will be more custom source types that will need similarly deep access into core functionality. 

I am not convinced that this is the case -- or at least, not that it will happen very often.  In fact, building a custom source that can join dynamic properties to vector tile geometries is already possible as of #2667 even without this PR: the [`loadVectorData`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/vector_tile_worker_source.js#L18) param in VectorTileWorkerSource allows a custom source author to provide their own `params => VectorTile` function, so all that's needed is an [alternate implementation of the vector-tile-js interface](https://github.com/developmentseed/cached-vector-tile) that allows pulling in external properties.

What's happening here in this PR is really about deep access into VectorTileWorkerSource to allow for a particular _performance optimization_--namely, regenerating paint vertex arrays without having to generate the layout arrays (and thus without having to re-parse geometries, re-triangulate, etc).  

> Our goal is to create a framework that allows users to add custom source types without modifying GL JS core. We should view _featureIndexToArrayRange through that lens. 

It seems to me that the question should be whether or not _featureIndexToArrayRange provides enough of a potential benefit to GL JS core to be worth including.  I would argue that it does, because the above mentioned performance enhancement would not only support "dynamic feature properties" -- it would also support changing _style_ properties.  It would be possible, for instance, apply the same optimization when there are changes in a style layer's paint properties but not in its layout properties or filter: the VT worker source could then skip rebuilding the layout arrays.

anandthakker:
@lucaswoj What I'm getting at but didn't quite state in my last comment is: if the case I made for keeping `_featureIndexToArrayRange` in core isn't persuasive and the final call is to keep it out, then I believe it would be acceptable to _not_ refactor to enable it to be replicated custom source types.  

The perf optimization that `_featureIndexToArrayRange` affords is quite "advanced"; I think that a custom source author who needs it could be reasonably expected to just copy `WorkerTile` and modify it to enable the necessary index tracking.  If we keep the refactors to `populatePaintArrays` that allow it to take both a starting and ending index, this would still be doable.

anandthakker:
Rebasing to keep up with master.  Previous head is https://github.com/mapbox/mapbox-gl-js/pull/2812/commits/927c8d978052946af9ab8de518b178a9f34c1c2d

# Branch

**load-multiple-maps:** 250 ms, loaded 6 maps.
**buffer:** 875 ms
**dds-parse-bucket:** 2,155 ms
**dds-update-bucket:** 1,203 ms
**fps:** 60 fps
**frame-duration:** 8.2 ms, 0% > 16ms
**query-point:** 0.77 ms
**query-box:** 52.55 ms
**geojson-setdata-large:** 102 ms


# Master

**load-multiple-maps:** 864 ms, loaded 6 maps.
**buffer:** 874 ms
**fps:** 60 fps
**frame-duration:** 8.1 ms, 0% > 16ms
**query-point:** 0.76 ms
**query-box:** 51.31 ms
**geojson-setdata-large:** 104 ms

AndyMoreland:
Hey folks,

I spoke with @ryanbaumann at my office a few days ago regarding some performance issues that I'm having with using mapbox-gl-js. He pointed me at this PR as a starting place for trying to improve hover / selection performance on maps with large numbers of geojson elements (40k is a reasonable target).

I've been doing some work based off of this PR, but it doesn't rebase cleanly against 0.24 anymore. I totally understand that this exact code probably won't be what gets merged into the mainline eventually, but is this a total fools errand? I'm primarily concerned about getting stuck permanently on my 0.22 fork if this gets abandoned.

Andy

lucaswoj:
I cannot speak to @anandthakker's plans to maintain this branch.

We have plans to enable functionality like the functionality implemented in this pull request through custom source and layer types (#281 https://github.com/mapbox/mapbox-gl-js/pull/2982 https://github.com/mapbox/mapbox-gl-js/pull/2982). I suspect the final form of this feature will involve simplification of GL JS's architecture and emphasize object composition as the preferred way to create custom sources.

If you are looking for a way to improve the performance of hovering and selection, I recommend looking at #2874, other `Map#setFilter` performance improvements, and `Map#queryRenderedFeatures` performance improvements.

anandthakker:
> I cannot speak to @anandthakker's plans to maintain this branch.

Based on the most recent discussion in the PR here, I don't expect that this branch is going to make it into master.   The work here was motivated by a project we're working on that requires joining changing property data with geometries provided by vector tiles.  As discussed [here](https://github.com/mapbox/mapbox-gl-js/pull/2812#issuecomment-242373415), this is actually already possible via the 'custom source' interface, although that interface is likely to change as part of the architecture simplifications that @lucaswoj mentioned, and so isn't yet publicly documented.  

@AndyMoreland I'll be open sourcing our approach to this, but it's going to take a few weeks due to constraints beyond my control.

AndyMoreland:
Thanks for the update, both of you. It sounds like I should hold off for a bit on this until you get the new APIs standardized.

ryanbaumann:
Linking to the `categorical data-driven style` method for performing property updates, which may be useful until vector tile property updates are merged into master.  

https://www.mapbox.com/mapbox-gl-js/example/data-join/

lucaswoj:
Closing this as "stale." We are working on a slightly different approach, allowing custom sources which adhere to a specific interface, here: https://github.com/mapbox/mapbox-gl-js/projects/2.

PR #2813, opened 2016-07-01
preserveDrawingBuffer documentation addition

tobinbradley:
Added note that Firefox will not print the map unless this is set to true (longstanding/unassigned [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=976173)). Not sure if this is the appropriate place for a browser specific note or not, but it might save someone some hair pulling. As this bug has been open and unassigned for 2.5 years I don't think the Mozilla folks are fixing it any time soon.


mourner:
Interesting! Is Firefox the only browser affected by this?

tobinbradley:
I haven't run across another browser with this issue. Chrome, Edge, and IE11 are all fine. It's definitely a Firefox thing (checked on Windows and Linux), and it's definitely a WebGL thing (none of the threejs.org examples print in Firefox either).

PR #2833, opened 2016-07-07
Move geolocate event docs from Control to Geolocate

### LINKED ISSUE #2832: Add API documentation for the "geolocate" event
ref https://github.com/mapbox/mapbox-gl-js/issues/2824

lucaswoj:
It turns out that documentation for these events existed but was on the `Control` class rather than the `Geolocate` class. 

fixes #2832


lucaswoj:
Will re-open against `mb-pages`

PR #2836, opened 2016-07-07
Fix Bucket#trimArrays

jfirebaugh:
The prior implementation was a no-op, because it skipped a level in the object hierarchy.

Fixing this produced a meaningful benchmark improvement. Buffer benchmark before: 997ms; after: 895ms.


jfirebaugh:
Rolled into #2837.

PR #2837, opened 2016-07-07
Bucket: Arrays and buffers

jfirebaugh:
Some refactoring of `Bucket` to reduce overhead, simplify object structure, improve naming conventions, and make the use of the terms "array" and "buffer" consistent (the latter always referring specifically to an instance of `Buffer`, not loosely as any kind of data collection).

cc @lucaswoj

### DIFF
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -92,23 +92,21 @@ function Bucket(options) {
             var paintArrayTypes = options.paintArrayTypes[programName];
             return programArrayGroups.map(function(arrayGroup) {
                 var group = {
-                    layout: {
-                        vertex: new Buffer(arrayGroup.vertexArray,
-                            programInterface.vertexArrayType.serialize(), Buffer.BufferType.VERTEX)
-                    },
-                    paint: util.mapObject(arrayGroup.paintArrays, function(array, name) {
+                    vertexBuffer: new Buffer(arrayGroup.vertexArray,
+                        programInterface.vertexArrayType.serialize(), Buffer.BufferType.VERTEX),
+                    paintBuffers: util.mapObject(arrayGroup.paintArrays, function(array, name) {
                         return new Buffer(array, paintArrayTypes[name], Buffer.BufferType.VERTEX);
                     }),
                     vaos: {}
                 };
 
                 if (arrayGroup.elementArray) {
-                    group.layout.element = new Buffer(arrayGroup.elementArray,
+                    group.elementBuffer = new Buffer(arrayGroup.elementArray,
                         programInterface.elementArrayType.serialize(), Buffer.BufferType.ELEMENT);
                 }
 
                 if (arrayGroup.elementArray2) {
-                    group.layout.element2 = new Buffer(arrayGroup.elementArray2,
+                    group.elementBuffer2 = new Buffer(arrayGroup.elementArray2,
                         programInterface.elementArrayType2.serialize(), Buffer.BufferType.ELEMENT);
                     group.secondVaos = {};
                 }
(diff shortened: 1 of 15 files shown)


js/data/bucket.js
```diff
@@ -128,19 +140,17 @@ Bucket.prototype.populateBuffers = function() {
...
  * An individual array group looks like:
  * {
  *     index: number,
- *     layout: {
- *         layout: VertexArrayType,
- *         ?element: ElementArrayType,
- *         ?element2: ElementArrayType
- *     },
- *     paint: { [layerName]: PaintVertexArrayType, ...  }
+ *     vertexArray: VertexArrayType,
```
lucaswoj: Why did you choose to call this `vertexArray` instead of `layoutArray`? 

Conceptually the `paintArrays` and `layoutArray` are both vertex arrays. The distinguishing characteristic is that one contains data from paint properties and the other data from layout properties.

js/data/bucket/circle_bucket.js
```diff
@@ -28,14 +28,13 @@ CircleBucket.prototype.addCircleVertex = function(vertexArray, x, y, extrudeX, e
 
 CircleBucket.prototype.programInterfaces = {
     circle: {
-        vertexBuffer: true,
-        elementBuffer: true,
-
-        layoutAttributes: [{
+        vertexArrayType: new Bucket.VertexArrayType([{
```
lucaswoj: beautiful 👌

js/data/bucket.js
jfirebaugh: Makes sense, I'll rename to `layoutArray`.

js/source/worker_tile.js
```diff
@@ -260,20 +258,8 @@ WorkerTile.prototype.redoPlacement = function(angle, pitch, showCollisionBoxes)
...
-            for (var layoutOrPaint in programArrayGroup) {
-                var arrays = programArrayGroup[layoutOrPaint];
-                for (var bufferName in arrays) {
-                    if (arrays[bufferName].length > 0) return true;
-                }
-            }
-        }
-    }
-    return false;
+function isBucketNonEmpty(bucket) {
```
lucaswoj: Keeping function names in the affirmative (`isBucketEmpty`) will keep our codebase more readable.

jfirebaugh: That's not possible in this case because of how the function is used (passed to `Array#filter`).

js/data/bucket.js
jfirebaugh: Or how about `layoutVertexArray`, `paintVertexArrays`, `layoutVertexBuffer`, `paintVertexBuffers`?

lucaswoj:
This is ready to :shipit: once you've addressed the actionable comments above ^

-----

I have a bigger-picture refactoring of `Bucket` in mind that may be worth pursuing before porting this 😐 architecture to native. 

The hierarchy of objects in our data system roughly looks like this:
> `Tile` -> `Bucket` -> program interface -> array group -> layout and paint arrays

We could simplify by collapsing the hierarchy to:
> `Tile` -> `TileLayer`

There would be a 1:1 correspondence between `TileLayer`s and calls to `drawElements`. All information that was previously stored in the object hierarchy would be stored as state within `TileLayer`. 

```
class TileLayer {
    programInterface: string;
    layoutArray:
    vertexArray:
    paintArrays:
    ...
}
```

We could also factor all `TileLayer` creation logic (i.e. `addFeature`) out into a new set of `TileLayerFactory` classes. This functionality is only used within the workers.

js/data/bucket.js
lucaswoj: That's good too 👍

js/source/worker_tile.js
lucaswoj: 👍 

(I didn't notice this method's implementation was literally `!bucket.isEmpty()` :joy:)

jfirebaugh:
Holding on merge here until post 0.21 release to avoid regression risk.

lucaswoj:
Ready to 🚢 ?

PR #2844, opened 2016-07-08
Add dataset support the GeoJSON sources

mcwhittemore:
As the dataset editor goes into beta, it would be nice to be able to load a dataset into mapbox-gl-js via the dataset url you get out of the editor.

image

One problem this has is that dataset require secret keys. Puts the secret key into the url passed to the GeoJSON source. It looks like mapbox-gl-js, has a strict no secret keys rule. If this is true, maybe we should wait until public key support for datasets is possible.

@lucaswoj @rclark @scothis

### DIFF
--- a/debug/dataset.html
+++ b/debug/dataset.html
@@ -69,8 +69,7 @@ map.on('load', function() {
 
     map.addSource('circles', {
         "type": "geojson",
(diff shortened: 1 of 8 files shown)


lucaswoj:

PR #2845, opened 2016-07-08
Fix error when adding and removing a layer within the same update cycle

### LINKED ISSUE #2787: Adding and then removing a layer as part of the same update gives a 'Cannot read property 'updatePaintTransitions' of undefined(…)' error
MapBox GL v0.19.0:

### Steps to Trigger Behavior

 1. Add and remove a layer as part of the same update

Demonstration: https://jsfiddle.net/e41m9uup/2/ (open the javascript console to view the error)

### Expected Behavior

Render the map as normal

### Actual Behavior

Throws a 'Cannot read property 'updatePaintTransitions' of undefined(…)' error.

My debugging so far:
   - In addLayer MapBox adds the layer to this._updates.paintProps (by calling updateClasses)
   - In removeLayer MapBox doesn't remove the layer from this._updates.paintProps (that only happens at the end of the update method)
   - _applyClasses happens before the end of the update method (before _updates.paintProps gets cleared), and expects every layer in this._updates.paintProps to be in this._layers, but the removed layers are no longer in this._updates.paintProps. So it throws a 'cannot read property of undefined'.

lucaswoj:
fixes #2787 

ref https://github.com/mapbox/mapbox-gl-test-suite/pull/122

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -441,6 +441,8 @@ Style.prototype = util.inherit(Evented, {
         layer.off('error', this._forwardLayerEvent);
 
         delete this._layers[id];
+        delete this._updates.layers[id];
+        delete this._updates.paintProps[id];
         this._order.splice(this._order.indexOf(id), 1);
 
         this._updates.allLayers = true;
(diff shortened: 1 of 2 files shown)


lucaswoj:
Does this look good to 🚢 @mollymerp @jfirebaugh?

jfirebaugh:
🚢

PR #2847, opened 2016-07-08
Un-ignore test-suite runtime-styling tests

### LINKED ISSUE #2797: Fix crashing runtime-styling test-suite tests
ref https://github.com/mapbox/mapbox-gl-js/pull/2793

 - filter-default-to-false
 - filter-default-to-true
 - filter-false-to-default
 - filter-false-to-true
 - filter-true-to-default
 - layer-add-raster
 - layout-property-function-to-default
 - layout-property-function-to-literal
 - layout-property-literal-to-default
 - layout-property-literal-to-function
 - paint-property-function-to-default
 - source-add-raster-inline
 - source-add-raster-url

lucaswoj:
fixes #2797 

cc @jfirebaugh 

ref https://github.com/mapbox/mapbox-gl-test-suite/pull/123

# Todo 

 - [x] figure out if it is ok to not destroy the gl context in test-suite
 - [x] fix `[Error: Invalid file signature]` log messages due to lack of JPEG support
 - [x] fix CI failures due to `wait` operations returning too early
 - [ ] fix CI segfaults

### DIFF
--- a/js/data/buffer.js
+++ b/js/data/buffer.js
@@ -83,7 +83,7 @@ Buffer.prototype.setVertexAttribPointers = function(gl, program) {
  * @param gl The WebGL context
  */
 Buffer.prototype.destroy = function(gl) {
-    if (this.buffer) {
+    if (gl && this.buffer) {
         gl.deleteBuffer(this.buffer);
     }
 };
--- a/js/data/buffer.js
+++ b/js/data/buffer.js
@@ -83,7 +83,7 @@ Buffer.prototype.setVertexAttribPointers = function(gl, program) {
  * @param gl The WebGL context
  */
 Buffer.prototype.destroy = function(gl) {
-    if (gl && this.buffer) {
+    if (this.buffer) {
         gl.deleteBuffer(this.buffer);
     }
 };
--- a/test/suite_implementation.js
+++ b/test/suite_implementation.js
@@ -69,7 +69,8 @@ module.exports = function(style, options, _callback) {
                 [];
 
             map.remove();
-            gl.destroy();
+            gl.getExtension('STACKGL_destroy_context').destroy();
+            delete map.painter.gl;
 
             callback(null, data, results.map(function (feature) {
                 feature = feature.toJSON();
--- a/js/render/vertex_array_object.js
+++ b/js/render/vertex_array_object.js
@@ -88,7 +88,7 @@ VertexArrayObject.prototype.unbind = function(gl) {
 };
 
 VertexArrayObject.prototype.destroy = function(gl) {
-    var ext = gl && gl.extVertexArrayObject;
+    var ext = gl.extVertexArrayObject;
     if (ext && this.vao) {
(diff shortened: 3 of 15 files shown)


lucaswoj:
Remaining test suite failures appear to be related to the implementation of `wait`. There are some cases in which `wait` will return despite some resources being pending. Proper resolution may require finishing #1715

lucaswoj:
This is ready to go on 🍏 

How does this look @mollymerp @jfirebaugh @mourner?

js/data/buffer.js
```diff
@@ -83,7 +83,7 @@ Buffer.prototype.setVertexAttribPointers = function(gl, program) {
  * @param gl The WebGL context
  */
 Buffer.prototype.destroy = function(gl) {
-    if (this.buffer) {
+    if (gl && this.buffer) {
```
jfirebaugh: When would `gl` be falsy here? Instead of adding a guard clause, can we eliminate that case?

test/suite_implementation.js
```diff
@@ -69,7 +69,8 @@ module.exports = function(style, options, _callback) {
                 [];
 
             map.remove();
-            gl.destroy();
+            gl.getExtension('STACKGL_destroy_context').destroy();
```
jfirebaugh: What's the difference between the before and after here, and what prompted the change?

js/data/buffer.js
lucaswoj: A falsy value for `gl` indicates that we have manually destroyed the GL context. Tracking this information prevents "invalid GL context" errors in the runtime styling tests when asynchronous operations are run after the context has been destroyed. 

I will investigate some alternate designs today.

test/suite_implementation.js
lucaswoj: My reading of [the headless-gl changelog](https://github.com/stackgl/headless-gl/blob/1ccb244b006f79637cb1573dbdfddf0a07cd4ca2/CHANGES.md#400) is that `gl.destroy()` is deprecated in favor of `gl.getExtension('STACKGL_destroy_context').destroy()`.

> What's the difference between the before and after here

No functional difference. `gl.getExtension('STACKGL_destroy_context').destroy()` appears to be the preferred API going forward.

> what prompted the change?

In working on this PR, I reread the `headless-gl` docs and changelog to see if I could find any incorrect usage that would explain our test failures or segfaults

js/data/buffer.js
lucaswoj: Implemented a (hopefully) cleaner solution in https://github.com/mapbox/mapbox-gl-js/pull/2847/commits/ff077540842941f5161d5443abf10fb76a35f5be

lucaswoj:
Updated and ready for 👀

jfirebaugh:
🚢

lucaswoj:
I am investigating some test-suite segfaults that only occur on CI.

lucaswoj:
The CI segfaults are annoying but should not block this release. I will create a separate issue to track their resolution.

PR #2850, opened 2016-07-11
Minor corrections to documentation

davidtheclark:
While working on a Flow type interface for Studio I noticed a few more little typos and oversights to fix.

### DIFF
--- a/js/ui/camera.js
+++ b/js/ui/camera.js
@@ -18,7 +18,7 @@ var Point = require('point-geometry');
  * @property {number} zoom The destination's zoom level.
  * @property {number} bearing The destination's bearing (rotation), measured in degrees counter-clockwise from north.
  * @property {number} pitch The destination's pitch (tilt), measured in degrees.
- * @property {LngLat} around If a `zoom` is specified, `around` determines the zoom center (defaults to the center of the map).
+ * @property {LngLatLike} around If a `zoom` is specified, `around` determines the zoom center (defaults to the center of the map).
  */
 
 /**
@@ -29,7 +29,7 @@ var Point = require('point-geometry');
  * @typedef {Object} AnimationOptions
  * @property {number} duration The animation's duration, measured in milliseconds.
  * @property {Function} easing The animation's easing function.
- * @property {Array<number>} offset `x` and `y` coordinates representing the animation's origin of movement relative to the map's center.
+ * @property {PointLike} offset `x` and `y` coordinates representing the animation's origin of movement relative to the map's center.
  * @property {boolean} animate If `false`, no animation will occur.
  */
 
@@ -47,7 +47,7 @@ util.extend(Camera.prototype, /** @lends Map.prototype */{
      * Sets the map's geographical centerpoint. Equivalent to `jumpTo({center: center})`.
      *
      * @param {LngLatLike} center The centerpoint to set.
-     * @param {EventData} [eventData] Data to propagate to any event listeners.
+     * @param {Object} [eventData] Data to propagate to any event listeners.
      * @fires movestart
      * @fires moveend
      * @returns {Map} `this`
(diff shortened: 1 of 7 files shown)


js/util/browser/browser.js
```diff
@@ -66,14 +66,6 @@ exports.timed = function (fn, dur, ctx) {
...
 };
 
-/**
- * Test if the current browser supports Mapbox GL JS
- * @param {Object} options
- * @param {boolean} [options.failIfMajorPerformanceCaveat=false] Return `false`
- *   if the performance of Mapbox GL JS would be dramatically worse than
- *   expected (i.e. a software renderer would be used)
- * @return {boolean}
- */
```
lucaswoj: Why did you delete this?

lucaswoj: Ah. Duplicate docs. 👍

js/ui/map.js
```diff
@@ -1236,6 +1236,11 @@ function removeNode(node) {
   * Returns a Boolean indicating whether the browser [supports Mapbox GL JS](https://www.mapbox.com/help/mapbox-browser-support/#mapbox-gl-js).
   *
   * @function supported
+  * @param {Object} options
+  * @param {boolean} [options.failIfMajorPerformanceCaveat=false] If `true`,
+  *   the function will return `false` if it is determined
+  *   that the performance of the created WebGL context would be too low.
```
lucaswoj: Can we use the copy from the deleted documentation instead? It is the product of additional rounds of revision. 

```diff
/**
  * Test if the current browser supports Mapbox GL JS
  * @param {Object} options
  * @param {boolean} [options.failIfMajorPerformanceCaveat=false] Return `false`
  *   if the performance of Mapbox GL JS would be dramatically worse than
  *   expected (i.e. a software renderer would be used)
  * @return {boolean}
  */
```

davidtheclark: Yes. I will replace that sentence in both places it appears.

js/ui/map.js
```diff
@@ -417,8 +417,8 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
      * If the map's current zoom level is higher than the new maximum,
      * the map will zoom to the new maximum.
      *
-     * @param {number} maxZoom The maximum zoom level to set (0-20).
-     *   If`null` or `undefined` is provided, the function removes the current maximum zoom (sets it to 20).
+     * @param {number | null | undefined} maxZoom The maximum zoom level to set (0-20).
```
lucaswoj: Why not `?number`?

davidtheclark: That would be fine, too. I just copied from above for consistency. I will change both.

js/ui/camera.js
```diff
@@ -513,7 +515,7 @@ util.extend(Camera.prototype, /** @lends Map.prototype */{
      * @param {number} [options.screenSpeed] The average speed of the animation measured in screenfuls
      *     per second, assuming a linear timing curve. If `options.speed` is specified, this option is ignored.
      * @param {Function} [options.easing] An easing function for the animated transition.
-     * @param {EventData} [eventData] Data to propagate to any event listeners.
+     * @param {Object} [eventData] Data to propagate to any event listeners.
```
lucaswoj: Can we be more specific with these type signatures?

davidtheclark: `EventData` had been rendered meaningless, right? I had thought from prior conversation that this data you could send in is an arbitrary object. Is that wrong?

js/ui/control/geolocate.js
```diff
@@ -94,6 +94,6 @@ Geolocate.prototype = util.inherit(Control, {
  * @event error
  * @memberof Geolocate
  * @instance
- * @property {EventData} data The returned PositionError object from the callback in [Geolocation.getCurrentPosition()](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition)
+ * @property {PositionError} data The returned [PositionError](https://developer.mozilla.org/en-US/docs/Web/API/PositionError) object from the callback in [Geolocation.getCurrentPosition()](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition).
```
davidtheclark: @lucaswoj For context: I added these links because the documentation generator does not understand those particular data types.

js/ui/camera.js
lucaswoj: You're right. This type signature is correct 👍

lucaswoj:
Looks good! 🚢

PR #2852, opened 2016-07-11
Don't silently drop error events

### LINKED ISSUE #2447: Make `Evented` print a warning message if an "error" event is lost
If an `error` event is fired but doesn't trigger any listeners, we should print a warning and the error message. This would

 - display errors when a development mistake causes errors to not propogate to `Map`
 - display errors fired while running unit tests
 - eliminate the need for `Map#onError`

ref #2446

lucaswoj:
fixes #2447

### DIFF
--- a/js/util/evented.js
+++ b/js/util/evented.js
@@ -89,7 +89,7 @@ var Evented = {
             // To ensure that no error events are dropped, print them to the
             // console if they have no listeners.
             if (util.endsWith(type, 'error')) {
-                console.error(data.error.message);
+                console.error((data && data.error) || data || 'Empty error event');
             }
             return this;
         }
--- a/js/util/evented.js
+++ b/js/util/evented.js
@@ -89,7 +89,7 @@ var Evented = {
             // Error events are important for debugging and should not be
             // silently dropped.
             if (util.endsWith(type, 'error')) {
-                (console.trace || console.error)('Dropped "error" event: ', data);
+                console.error('Dropped "error" event: ', data);
             }
             return this;
         }
--- a/js/util/evented.js
+++ b/js/util/evented.js
@@ -86,10 +86,10 @@ var Evented = {
      */
     fire: function(type, data) {
         if (!this.listens(type)) {
-            // Error events are important for debugging and should not be
-            // silently dropped.
+            // To ensure that no error events are dropped, print them to the
+            // console if they have no listeners.
             if (util.endsWith(type, 'error')) {
-                console.error('Dropped "error" event: ', data);
+                console.error(data.error.message);
             }
             return this;
         }
--- a/js/util/evented.js
+++ b/js/util/evented.js
@@ -85,7 +85,14 @@ var Evented = {
      * @returns {Object} `this`
      */
     fire: function(type, data) {
-        if (!this.listens(type)) return this;
+        if (!this.listens(type)) {
+            // Error events are important for debugging and should not be
+            // silently dropped.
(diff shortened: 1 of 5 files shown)


js/util/evented.js
```diff
@@ -78,7 +78,14 @@ var Evented = {
      * @returns {Object} `this`
      */
     fire: function(type, data) {
-        if (!this.listens(type)) return this;
+        if (!this.listens(type)) {
+            // Error events are important for debugging and should not be
+            // silently dropped.
+            if (util.endsWith(type, 'error')) {
+                (console.trace || console.error)('Dropped "error" event: ', data);
```
jfirebaugh: `(console.trace || console.error)(...)` doesn't work:

image

You have to use `call` or `apply`.

js/util/evented.js
```diff
@@ -78,7 +78,14 @@ var Evented = {
      * @returns {Object} `this`
      */
     fire: function(type, data) {
-        if (!this.listens(type)) return this;
+        if (!this.listens(type)) {
+            // Error events are important for debugging and should not be
+            // silently dropped.
+            if (util.endsWith(type, 'error')) {
```
mourner: Can we do something like `if (data && data.error instanceof Error)` instead of checking the string?

lucaswoj: What's your concern with checking the string @mourner? I believe there are cases where we fire legitimate error events for which `data && data.error instanceof Error` would be `false`.

lucaswoj: For example, some error events might include a style-spec [`ValidationError` object](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb -pages/lib/error/validation_error.js). Ideally this would be a subclass of `Error` but subclassing `Error` in Javascript requires some hacks.

lucaswoj:
Updated and ready for 👀

jfirebaugh:
Which events will go through this pathway that won't already go through [this one](https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/map.js#L1006-L1021)? Do we need both?

lucaswoj:
This pathway exists to catch

 - errors emitted during testing (no errors go through the `Map#onError` pathway)
 - errors which are mistakenly not sent through the `Map#onError` method

lucaswoj:
Another design option is to remove the existing `Map#onError` pathway in favor of the new pathway (with some adjustments). This could be promising. The only potential downside is a break in backwards compatibility.

lucaswoj:
> Another design option is to remove the existing `Map#onError` pathway in favor of the new pathway (with some adjustments). This could be promising. The only potential downside is a break in backwards compatibility. 

Just pushed a proof-of-concept to [`error-remove-onError-2447`](https://github.com/mapbox/mapbox-gl-js/compare/error-remove-onError-2447?expand=1)

jfirebaugh:
I like that approach -- "bind a handler to avoid the default behavior" is a nicer API than "bind your own handler, and also unbind `map.onError`".

Note that if `error` is an `Error` instance, `console.error(error)` (at least on Chrome) will do some nice formatted logging with a disclosure triangle for the stack trace, while `console.error(error.message)` will not.

js/util/evented.js
mourner: Yeah, makes sense. I'm not very concerned, just thought it would be slightly less hacky/simpler. Maybe just `if (data && data.error)` would do?

lucaswoj:
Updated and ready for 👀

Issue #2848, opened 2016-07-09
Add a parameter to Popup that allows users to add CSS classes 


infacq:
how do I append a class along with class `mapboxgl-popup`. Thanks

lucaswoj:
This functionality is not currently supported in GL JS. I have repurposed this ticket to track that functionality. I encourage you to submit a PR adding it! Here is the relevant line: http://github.com/mapbox/mapbox-gl-js/blob/66a567cfb7b566f672f7e77399c6607ad86701ce/js/ui/popup.js#L174-L175

jfirebaugh:
Instead of adding a class to the `mapboxgl-popup` div, supply your own HTML content (using `setHTML` or `setDOMContent` that contains a div with the desired class.

andrewharvey:
> Instead of adding a class to the mapboxgl-popup div, supply your own HTML content (using setHTML or setDOMContent that contains a div with the desired class.

I down voted this because it is a child of the popup container which includes the popup tip, hence with this approach it seems you can't control the style of the popup tip.

Adding CSS classes to the root popup container would address that.

cyprianos:
Here my "brute-force" solution for changing background of popup belonging to marker.

Double use of parentNode to access main element, append className and create appropriate styles for tip.  In the same manner u can play with border length to change tip length or override any styles u need.

`
var popup = new mapboxgl.Popup();
var content = document.createElement('div');

content.innerHTML = '<p>some html</p>';
popup.setLngLat([-118.705, 34.142[)
                .addTo(map)
                .setDOMContent(content);

content.parentNode.parentNode.className += ' mapboxgl-popup--blue';'

CSS:

`
.mapboxgl-popup.mapboxgl-popup--blue {
/*any styles u wish*/
}

.mapboxgl-popup--blue .mapboxgl-popup-content{
    background-color: rgb(66,111,175);
    padding:20px;
    color: white;
    max-width:none;
}
.mapboxgl-popup--blue.mapboxgl-popup-anchor-bottom .mapboxgl-popup-tip{
    border-top-color: rgb(66,111,175);
}
.mapboxgl-popup--blue.mapboxgl-popup-anchor-top .mapboxgl-popup-tip{
    border-bottom-color: rgb(66,111,175);
}
.mapboxgl-popup--blue.mapboxgl-popup-anchor-left .mapboxgl-popup-tip{
    border-right-color: rgb(66,111,175);
}
[...]

PR #2853, opened 2016-07-12
Allow users to add CSS classes

infacq:
please refer to https://github.com/mapbox/mapbox-gl-js/issues/2848#issuecomment-231807671

### DIFF
--- a/js/ui/popup.js
+++ b/js/ui/popup.js
@@ -18,6 +18,7 @@ var LngLat = require('../geo/lng_lat');
  * @param {string} options.anchor - One of "top", "bottom", "left", "right", "top-left",
  * "top-right", "bottom-left", or "bottom-right", describing where the popup's anchor
  * relative to the coordinate set via `setLngLat`.
+ * @param {string} options.css for custom css
  * @example
  * var tooltip = new mapboxgl.Popup()
  *   .setLngLat(e.lngLat)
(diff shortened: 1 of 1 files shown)


mourner:
There may be use cases when you want to target the popup instead of internal HTML. However I'm not sure we need a special UI for this. You can already do `popup._container.classList.add('my-class');`, and to make that cleaner, I would rather add a `popup.getContainer` method.

infacq:
@peterqliu create a child an be solve by using setHTML or setDOMContent but you definitely don't have control over anchor tip. To have that, I must have access to inject a class into `.mapboxgl-popup`

@mourner i never knew we can alter it by using plain `js`. But of course it's a bit hacky and not preferable

andrewharvey:
I wanted to try using https://daneden.github.io/animate.css/ to fade in popups and a public API change to Popup allowing custom CSS classes to be added to the popup, would be great, since you could just do `var popup = new mapboxgl.Popup({ classes: 'animated fadeIn' })`.

 * Using setHTML misses the anchor tip, so won't work.
 * Directly overriding the `.mapboxgl-popup` class isn't ideal as in the example above adding the `animated fadeIn` classes is better abstraction.
 * Relying on private `popup._container` I don't like, it's undocumented and subject to change. But a public popup.getContainer would be okay, but the classes option feels much cleaner, simpler and encourages adding your own classes rather than overriding GL JS css which I feel is much cleaner. It seems you need to do this once you add it to the map (otherwise the DOM element doesn't exist?), at first I thought that delay in then setting the classes would be an issue, but it seems to be fine for the `animated fadeIn` classes.

Is supporting a classes option ruled out? If so then as an alternative solution would you likely accept a PR adding the `getContainer` method?

jcarenza:
@mourner I'm using mapbox-gl@0.43.0, and it doesn't appear that `popup._container` exists as a property, or any other methods like `getContainer` or `getElement`...is there another workaround to this issue that you could share?

mourner:
@jcarenza for a workaround, you can use `popup._container` — it should exist as soon as you add it to the map. https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/popup.js#L256

jcarenza:
Perhaps my use case is a bit different...I'm using it in combination with a Marker:
```javascript
marker = new mapboxgl.Marker(el).setLngLat([lng, lat]).addTo(map)
marker._popup._container.classList.add('my-class')
```
This throws `Uncaught TypeError: Cannot read property 'classList' of undefined`

PR #2857, opened 2016-07-12
Fix webpack support

michaelhue:
Fixes a [known issue](https://github.com/borisirota/webworkify-webpack#caveats) with anonymous functions and workerify-webpack in webpack-based build environments. See [original issue](https://github.com/borisirota/webworkify-webpack/issues/10#issuecomment-231610587).

Actual error message:

> blob:http://localhost:3000/e44fa451-b7fd-4eb3-871c-e5f0a502f7dd:12
> Uncaught TypeError: Cannot read property 'call' of undefined

### DIFF
--- a/js/source/worker.js
+++ b/js/source/worker.js
@@ -14,7 +14,7 @@ var rewind = require('geojson-rewind');
 var GeoJSONWrapper = require('./geojson_wrapper');
 var vtpbf = require('vt-pbf');
 
-module.exports = function(self) {
+module.exports = function MapboxGlWorker(self) {
     return new Worker(self);
 };
 
--- a/webpack.config.example.js
+++ b/webpack.config.example.js
@@ -32,7 +32,8 @@ const path = require('path');
  * Mapbox GL JS uses the webworkify module by default, and that module
  * does things very specific to browserify's module loading system. In this
  * configuration, we alias webworkify to webworkify-webpack to add webpack
- * support.
+ * support. NOTE: there is a known caveat when using 'eval'-related devtool
+ * settings, make sure to double-check your configuration.
  */
 module.exports = {
     entry: './test.js',
@@ -59,5 +60,6 @@ module.exports = {
             loader: 'transform',
             query: 'brfs'
         }]
-    }
+    },
+    devtool: 'cheap-source-map' // don't use eval-related options
 }


js/source/worker.js
```diff
@@ -14,7 +14,7 @@ var rewind = require('geojson-rewind');
 var GeoJSONWrapper = require('./geojson_wrapper');
 var vtpbf = require('vt-pbf');
 
-module.exports = function(self) {
+module.exports = function MapboxGlWorker(self) {
```
jfirebaugh: If I understand https://github.com/borisirota/webworkify-webpack/issues/9 correctly, this shouldn't be necessary with webworkify-webpack 1.1.1, correct?

michaelhue: It should be fixed, correct. However, I'm on `webworkify-webpack@1.1.1` and `mapbox-gl@0.20.1` using the recommended webpack.config.js setup. This produced the error mentioned in the pull request. Adding a name to the function fixed the issue for me.

bensleveritt:
I'd like to add that I'm seeing this too, having followed the recommended webpack setup.

Also on `webworkify-webpack@1.1.1`

bensleveritt:
Despite `webworkify-webpack` [updating to 1.1.3](borisirota/webworkify-webpack#13), I'm still seeing this issue.

Should this be pursued with `webworkify-webpack`?

lucaswoj:
If naming the function is the only way to fix webpack then I think we should do it. There are no substantial cons.

 - [ ] add a source-level comment explaining why this function must be named
 - [ ] file an upstream bug against `workerify-webpack`
 - [ ] create an automated test of building GL JS with webpack
 - [ ] merge this PR

bensleveritt:
Dang, you beat me @austinhyde.

In addition to that I had to remove `eval` from `devtool` in my webpack config, as per [webworkify-webpack's caveats](https://github.com/borisirota/webworkify-webpack#caveats).

lucaswoj:
Great! Can we close this issue now @austinhyde @bensleveritt @michaelhue?

austinhyde:
I think that if @michaelhue can confirm webworkify-webpack v1.1.3 fixes his issue, then yes. Otherwise, this is a good stopgap with no drawbacks, until it can be addressed in webworkify-webpack.

austinhyde:
Hmm... just ran into this issue: https://github.com/borisirota/webworkify-webpack/pull/13#issuecomment-232769037

I'd hold off on closing/merging this for now.

michaelhue:
@austinhyde `1.1.3` works for me now, even with minified builds. Thank you! And thank you @lucaswoj for considering the issue. I think the added note about `devtool` compatibility in `webpack.config.example.js` could still be useful though.

lucaswoj:
Sounds like the upstream fix is underway. I prefer to wait for that!

lucaswoj:
> I think the added note about devtool compatibility in webpack.config.example.js could still be useful though.

Just re-read this. I agree.

js/source/worker.js
lucaswoj: Can we remove this line ☝️

webpack.config.example.js
```diff
@@ -59,5 +60,6 @@ module.exports = {
             loader: 'transform',
             query: 'brfs'
         }]
-    }
+    },
+    devtool: 'cheap-source-map' // don't use eval-related options
```
lucaswoj: Why `inline-source-map` over `source-map` or`inline-source-map`?

webpack.config.example.js
```diff
@@ -32,7 +32,8 @@ const path = require('path');
  * Mapbox GL JS uses the webworkify module by default, and that module
  * does things very specific to browserify's module loading system. In this
  * configuration, we alias webworkify to webworkify-webpack to add webpack
- * support.
+ * support. NOTE: there is a known caveat when using 'eval'-related devtool
+ * settings, make sure to double-check your configuration.
```
lucaswoj: Can you provide more context here? What is the caveat? What should developers check their configurations _for_?

austinhyde:
I'm actually going to recommend merging this. There's some bigger issues going on in webworkify-webpack that I'm trying to work through, and this is a harmless fix in the meantime.

webpack.config.example.js
michaelhue: Apologies for the confusion, I'm afraid English is not my first language. I was referring to this part of `webworkerify-webpack`'s README:

> While developing make sure not to set webpack's devtool option to be with some eval configuration because from 1.1.0 version it won't work

webpack.config.example.js
michaelhue: I found `cheap-source-map` to be a good tradeoff but of course any other non-eval setting will do.

js/source/worker.js
austinhyde: For the time being, webworkify-webpack needs a named function to operate correctly.

lucaswoj:
Replaced by #2887 (bringing the PR within the Mapbox organization to make final edits easier)

PR #2859, opened 2016-07-13
Don't run "npm update" on CI unless testing "master"

lucaswoj:
Regularly running `npm update` is good practice. We want to test and use all the latest dependencies for GL JS. 

One downside to running `npm update` is that it takes much longer (7 minutes) than `npm install` (31 seconds).

By only running `npm update` for builds of the `master` branch, we regularly test the latest dependencies, keep our cached `node_modules` directory up to date, ensure all released versions of GL JS use the latest dependencies, and shorten our PR build time by 7 minutes. 

Any downsides to this design that I'm missing?

cc @mourner @jfirebaugh @mollymerp @tmcw

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,7 +3,7 @@ machine:
     version: '4'
 dependencies:
   override:
-    - npm update
+    - if [ "$(git symbolic-ref --short -q HEAD)" == "master" ]; then npm update; else npm install; fi;
 test:
   override:
     - ./ci.sh


bhousel:
FWIW I'm a big fan of eliminating semver ranges completely in `package.json` and just letting [Greenkeeper.io](https://greenkeeper.io/) test against new versions of things and open pull requests automatically.  In that situation CI can always just do a clean checkout and `npm install`.

I've taken this approach in iD and it's really nice to not have to think about dependency management anymore.

mourner:
Yeah, it may be worth trying the fixed version + Greenkeeper approach. The only downside I see is potentally a lot of PR noise from Greenkeeper on minor updates.

lucaswoj:
We've had surprisingly little trouble using semver + `npm update`. I think there've been 2 failures in my time at Mapbox, both of which were easily debuggable and fixable.

lucaswoj:
Let's merge this and talk about switching to greenkeeper.io when npm versioning becomes a pain point.

PR #2869, opened 2016-07-14
Correct tile coverage calculation for overscaled tiles

### LINKED ISSUE #2868: Fix incorrect z coordinate on overzoomed tiles

When calling `querySourceFeatures` past the max zoom of the source layer, the tiles returned are not in the viewport. They have the `x` and `y` values from the max zoom of the source layer but the `z` value from the actual zoom level. My recreation relies on some private layers but you can also see the error in the tile boundary calculation (11/328/794 is correct, 11/164/397 has the correct x/y for zoom 10):

screen shot 2016-07-14 at 11 08 03 am

I believe the [`reparseOverscaled` ternary operator](https://github.com/mapbox/mapbox-gl-js/blob/de61bbfa21f070e7afbe4fd42e24e8e44e7e25f1/js/source/tile_pyramid.js#L151) is reversed in the `TilePyramid`; when swapping `z` and `actualZ` locally the problem was resolved. I can add a PR but not sure of a good test case

drewbo:
Close #2868

### DIFF
--- a/js/source/tile_pyramid.js
+++ b/js/source/tile_pyramid.js
@@ -148,7 +148,7 @@ TilePyramid.prototype = {
             tr.pointCoordinate(new Point(tr.width, 0))._zoomTo(z),
             tr.pointCoordinate(new Point(tr.width, tr.height))._zoomTo(z),
             tr.pointCoordinate(new Point(0, tr.height))._zoomTo(z)
-        ], this.reparseOverscaled ? actualZ : z).sort(function(a, b) {
+        ], this.reparseOverscaled ? z : actualZ).sort(function(a, b) {
             return centerPoint.dist(a) - centerPoint.dist(b);
         });
     },


lucaswoj:
Thanks for tracking this down @drewbo! 

Can you help me put together regression test for [`mapbox-gl-test-suite`](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/query-tests)? What would this test look like? Querying any bounding box on an overzoomed tile?

drewbo:
> Querying any bounding box on an overzoomed tile?

I think something like that makes sense.

Also 17 of these are failing: https://github.com/mapbox/mapbox-gl-js/blob/master/test/js/source/tile_pyramid.test.js (hence the circleci redness). I need to do a little more reading up since I'm not as familiar with what's going on here: were the original tests wrong or is my "solution" not appropriate?

lucaswoj:
My hunch is that the original tests are wrong.

lucaswoj:
The more I dig into this issue, the more unsure I become about the intent of the author.

We are using `TileCoord` in a way that isn't semantically valid: the `z` property relates to unscaled parent coordinate while the `x` and `y` properties relate to the scaled child coordinate). 

I cannot discern if this behavior is intentional or a mistake. Thoughts @jfirebaugh @ansis?

Issue #2851, opened 2016-07-11
Add flow types for all public APIs


davidtheclark:
I create a Flow definition file of this library for Mapbox Studio, and chatted a little with @lucaswoj about putting that definition in this repo.

I can think of some reasons that it would be a good idea to maintain the definitions in this repo:
- Closer to the source code
- Updates would always go out with new releases
- If we ever figure out a way to autogenerate the definition from the thorough JSDoc comments it can happen here

I can't think of any reason _not_ to put the definitions here. Can anybody else?

And if we do put them here, where should they go? I'm happy to open a PR.

lucaswoj:
The flow definitions can go in this repository's root directory. PR away!

davidtheclark:
Are all the properties exposed in https://github.com/mapbox/mapbox-gl-js/blob/master/js/mapbox-gl.js intended to be part of the public API?

If so, we should document `version`, `util`, and `config`, and add them all the Flow definition.

(I was thinking about this because Studio uses `mapboxgl.config` so needed it in the definition.)

lucaswoj:
- :+1: `version` I added this property and failed to document it. This can and should be documented.
 - 👎 `util` There has been a lot of debate about whether or not we should document this. The current consensus is that we should not. https://github.com/mapbox/mapbox-gl-js/issues/1408
 - ❓ `config` I'm not aware of any existing discussion about documenting this property. My opinion is that we should look into simplifying the architecture and then document it. A simpler architecture would be dropping the `config` module in favor of top level properties on `mapboxgl`.

lucaswoj:
We now have flow built in! Reframing this ticket to be about adding flow types to all public APIs.

PR #2870, opened 2016-07-14
Add flow def

davidtheclark:
Ref #2851.

Unfortunately, I'm not yet sure I'm doing this right ...

[The Flow docs](https://flowtype.org/docs/declarations.html#declaration-files) suggest that we should be able to put `mapbox-gl.js.flow` next to `mapbox-gl.js` and in there `export` an object; then Flow will automatically pick up the definition. I've tried to do that following [the example I see in Immutable.js](https://github.com/facebook/immutable-js/blob/master/dist/immutable.js.flow) (the only example I've yet been able to find), but when I then type-check in Studio it looks like the definition file is not in fact registered.

What I have in this PR right now works as long as you manually add `mapbox-gl.js.flow` to a `.flowconfig` file.

I'm going to try poking around a little more to see why the automatic definition registration doesn't seem to work ...

### DIFF
--- a/js/mapbox-gl.js.flow
+++ b/js/mapbox-gl.js.flow
@@ -24,7 +24,7 @@ type MapOptions = {
   doubleClickZoom?: boolean,
   touchZoomRotate?: boolean,
   trackResize?: boolean,
-  center?: Array<number>,
+  center?: LngLatLike,
   zoom?: number,
   bearing?: number,
   pitch?: number,
--- a/js/mapbox-gl.js.flow
+++ b/js/mapbox-gl.js.flow
@@ -1,0 +1,367 @@
+type Style = Object;
+type Filter = Array<any>;
+type Source = Object;
+type Layer = Object;
+
+type MapOptions = {
+  container: HTMLElement | string,
+  minZoom?: number,
+  maxZoom?: number,
+  style?: Style | string,
+  hash?: boolean,
+  interactive?: boolean,
+  bearingSnap?: number,
+  classes?: Array<string>,
+  attributionControl?: boolean,
+  failIfMajorPerformanceCaveat?: boolean,
+  preserveDrawingBuffer?: boolean,
+  maxBounds?: LngLatBoundsLike,
+  scrollZoom?: boolean,
+  boxZoom?: boolean,
+  dragRotate?: boolean,
+  dragPan?: boolean,
+  keyboard?: boolean,
+  doubleClickZoom?: boolean,
+  touchZoomRotate?: boolean,
+  trackResize?: boolean,
+  center?: Array<number>,
+  zoom?: number,
+  bearing?: number,
+  pitch?: number,
+  workerCount?: number
+}
+
+declare class Map mixins Evented {
+  constructor(options: MapOptions): void;
+  accessToken?: string;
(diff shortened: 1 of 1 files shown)


js/mapbox-gl.js.flow
```diff
@@ -0,0 +1,367 @@
...
+  maxBounds?: LngLatBoundsLike,
+  scrollZoom?: boolean,
+  boxZoom?: boolean,
+  dragRotate?: boolean,
+  dragPan?: boolean,
+  keyboard?: boolean,
+  doubleClickZoom?: boolean,
+  touchZoomRotate?: boolean,
+  trackResize?: boolean,
+  center?: Array<number>,
```
tmcw: In this and other places where we mention the `center` argument, it is technically a LngLatLike, so can either be a two-element array _or_ a LngLat object. Maybe LngLatLike should be a type alias?

davidtheclark: Good catch. Fixed.

lucaswoj:
As a fan of flow, I am excited about providing a first party flow interface for our users 👍 

As a maintainer of GL JS, I'm nervous about endorsing this interface file as-written. If we are going to 🚢 this to users, we should be very confident in the accuracy and completeness of this interface. To me, this means:

 1. automatically generating the flow interface file from the jsdoc comments
 2. verifying this flow interface using `flow`

@tmcw @jfirebaugh What is the status of the documentation.js + flow integration? I know this has been on your roadmap for a while.

tmcw:
> automatically generating the flow interface file from the jsdoc comments

This isn't really possible, via documentation.js or any other way. JSDoc and Flow cover different ground, and hitting both in the same code is a totally unsolved problem. documentation.js can convert Flow annotations _into_ JSDoc annotations, but not vice-versa. And only for Flow annotations, not Flow comment annotations ([issue](https://github.com/documentationjs/documentation/issues/306)).

So, if your goal is two-in-one, the only current solution is to use Flow in Mapbox GL JS.

lucaswoj:
Can you provide an example of a flow construct that has no analog in JSDoc?

tmcw:
[Generics?](https://flowtype.org/docs/functions.html#function-based-type-annotations) In any case, if you feel like contributing a JSDoc-to-Flow converter to documentation.js, go for it :)

lucaswoj:
We cannot merge this PR until we have some way to test this interface. We can't commit to supporting untested and unused code in this repo.

lucaswoj:
To be clear, I'm in favor of this feature and excited to merge it once we have figured out a way to test the interface.

lucaswoj:
`flow-typed` has established some prior art in writing tests for flow interfaces https://github.com/flowtype/flow-typed#writing-tests

vicapow:
Can we add the flowtype annotations within the source it self? An additional babel transform can be used to strip the types into comments for publishing. see: https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-flow-comments

lucaswoj:
@vicapow Looking into that right now 😄

mourner:
We can now add Flow definitions to the source directly after #3397 landed, so closing as outdated (but having this as a reference will be very useful for future flow-annotating PRs).

PR #2871, opened 2016-07-14
Document version property

davidtheclark:
Ref https://github.com/mapbox/mapbox-gl-js/issues/2851#issuecomment-232123646

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -1220,7 +1220,7 @@ function removeNode(node) {
 /**
  * Gets and sets the map's [access token](https://www.mapbox.com/help/define-access-token/).
  *
- * @var accessToken
+ * @var {string} accessToken
  * @example
  * mapboxgl.accessToken = myAccessToken;
  */
@@ -1228,7 +1228,7 @@ function removeNode(node) {
 /**
  * Gets the version number of Mapbox GL JS.
  *
- * @var version
+ * @var {string} version
  */
 
 /**
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -1226,7 +1226,8 @@ function removeNode(node) {
  */
 
 /**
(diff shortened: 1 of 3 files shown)


js/ui/map.js
```diff
@@ -1225,18 +1225,24 @@ function removeNode(node) {
...
-  *   be dramatically worse than expected (i.e. a software renderer would be used).
-  * @return {boolean}
-  * @example
-  * mapboxgl.supported() // = true
-  */
+/**
+ * Gets the version number of Mapbox GL JS.
+ *
+ * @var version
+ */
```
lucaswoj: Can we declare that `version` is a `string`?

lucaswoj: Can we provide any additional context here for users who don't immediately make the leap from `version number` to git tags, npm, GitHub releases, etc

davidtheclark: I do not know of a way to declare that a `var` is a string. If I add a type, e.g. `@var version {string}`, compilation for that item seems to fail. Do you know of another way this should be marked with JSDoc?

Regarding context, what do you think of this? ` To learn more about the differences between published versions of Mapbox GL JS, have a look at [the changelog](https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md).` Sufficient? I guess I'm not sure what else to say here ... open to ideas.

lucaswoj:
Looks like `@var` is intended to support a type parameter: http://usejsdoc.org/tags-member.html. It is good to capture the metadata even if this doesn't show up in our documentation theme as-written. 

I've pushed a few commits. 🚢 if this looks good to you!

davidtheclark:
Thanks, @lucaswoj, and sorry for my obtuseness --- I just put the type in the wrong place (；一_一). What you've done makes sense.

Issue #2874, opened 2016-07-14
Optimize "Map#setFilter" performance for "hover effect" use case


lucaswoj:
Creating performant hover effects in GL JS is hard.  One solution to this problem is making our primitives (`Map#queryRenderedFeatures`, `Map#setFilter`, `GeoJSONSource#setData`, ...) faster. Another solution is adding a primitive optimized for the hover usecase.

Suppose we allowed filtering features by id within the shaders. With a buffer size increase equivalent to that of a single data-driven styling property, we could have zero-overhead filtering on even the largest data sets. 

A single filter id could be passed to the shader as a `uniform`, with some common boilerplate encapsulated within `#pragma`.

We could use this optimized filtering pathway automatically when a filter in the form `['==', 'id', *]` is used or provide a new `filterId` style property. 

cc @jfirebaugh @mourner @peterqliu @tmcw @tristen

lucaswoj:
This is a specific proposal for #200

jfirebaugh:
What would the shader do based on the uniform value? Is the idea that it would implement style syntax similar to https://github.com/mapbox/mapbox-gl-js/issues/200#issuecomment-75163677?

lucaswoj:
The shader would not draw a feature unless the feature's id (i.e. `feature.properties.id`) matches the feature id filter (ie. `filter: ['==', 'id', 5]` or `featureIdFilter: 5`)

The shader implementation would look roughly like this:
```glsl
if (u_feature_id_filter === a_feature_id) {
    gl_FragColor = gl_FragColor = vec4(0.0);
}
```

This could be a stepping stone to the syntax proposed in #200 or a standalone feature. The style syntax in #200 is more ergonomic for our users and offers less room for a fast implementation.

jfirebaugh:
Got it. So at least at first it's purely an optimization for existing functionality -- `map.setFilter(layerID, ['==', 'id', n])` would be special-cased to re-render with the new uniform value, rather than reparsing tiles.

Looking at our [current hover example](https://www.mapbox.com/mapbox-gl-js/example/hover-styles/), it wouldn't be able to benefit from this optimization, because it uses the "name" property rather than id. It's also appears to be fast enough as currently implemented.

Have we seen a lot of other cases where this pathway would be used?

anandthakker:
[...]

PR #2877, opened 2016-07-15
Pre-filter sources for queryRenderFeatures

anandthakker:
This change aims to improve queryRenderedFeatures performance in cases where a `layers` parameter is present by only querying the sources referenced by the requested layers.

This can lead to significant savings in some cases because (a) the first time a newly-loaded source tile is queried, [it must parse the pbf data into a VectorTile](https://github.com/mapbox/mapbox-gl-js/blob/master/js/data/feature_index.js#L100-L103), and (b) tiles for some sources may be reloaded quite often if, for instance, `setFilter` or `setData` are being used for "hover"-like interactivity. (See also #2874)

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -637,15 +637,17 @@ Style.prototype = util.inherit(Evented, {
             this._handleErrors(validateStyle.filter, 'queryRenderedFeatures.filter', params.filter, true);
         }
 
-        var includedSources;
+        var includedSources = {};
         if (params.layers) {
-            var styleLayers = this._layers;
-            includedSources = params.layers.map(function (layerId) { return styleLayers[layerId].source; });
+            for (var i = 0; i < params.layers.length; i++) {
+                var layerId = params.layers[i];
+                includedSources[this._layers[layerId].source] = true;
+            }
         }
 
         var sourceResults = [];
         for (var id in this.sources) {
-            if (includedSources && includedSources.indexOf(id) < 0) continue;
+            if (params.layers && !includedSources[id]) continue;
             var source = this.sources[id];
             if (source.queryRenderedFeatures) {
                 sourceResults.push(source.queryRenderedFeatures(queryGeometry, params, zoom, bearing));
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -638,7 +638,16 @@ Style.prototype = util.inherit(Evented, {
         }
 
         var sourceResults = [];
-        for (var id in this.sources) {
+        var sourcesToQuery = {};
+        if (params.layers) {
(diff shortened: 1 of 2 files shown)


mourner:
Nice! Let's fix the build.

anandthakker:
^ last commit should fix the build and adds a test of the new behavior

anandthakker:
@mourner @lucaswoj this is ready for 👀  now.  

(Also, I apologize for creating extra noise over in the feature-id-shader ticket)

js/style/style.js
```diff
@@ -638,7 +638,16 @@ Style.prototype = util.inherit(Evented, {
...
+        var sourcesToQuery = {};
+        if (params.layers) {
+            for (var i = 0; i < params.layers.length; i++) {
+                var sourceId = this._layers[params.layers[i]].source;
+                sourcesToQuery[sourceId] = true;
+            }
+        } else {
+            sourcesToQuery = this.sources;
+        }
+        for (var id in sourcesToQuery) {
```
lucaswoj: Given that this isn't hot code, do you think we could get by with a simpler implementation like this 👇 ?

```js
for (var id in this.sources) {
    if (!params.layers || params.layers.some(function(layerId) { return this._layers[layerId].source === id; }) continue;
    ...
```

anandthakker: Sure, 👍 for filtering the sources within the loop like you propose.

I'm a little bit wary of using `array.some()` here rather than a plain for loop over `params.layers`, only because this method could get called with a pretty high frequency, e.g., in a `mousemove`. handler, where this might make a difference.  (But I admit this is based on intuition, not quantitative evidence)

lucaswoj: I don't think the impact of `Array.some` or equivalent logic will ever be a bottleneck next to `Source#queryRenderedFeatures`. If it becomes one, we will identify it in the profiler and look for optimizations.

anandthakker: Heh, fair enough 😄

anandthakker:
@lucaswoj ^ simplified the logic -- I did it this way because the "no-loop-func" linter rule prohibited the particular form you suggested

mourner:
Given that we're dealing with sources, the difference between discussed forms will be negligible because there are typically only a very few sources on a particular map. However, in other situations like this where there are more objects to deal with (like layers), the original form with a simple `for` loop and a hash is highly preferable because  it's `O(n)` rather than `O(n^2)`.

lucaswoj:
If we can't do the `O(n^2)` logic as a one-liner, we should revert to the `O(n)` logic. I pushed a commit that implements this. We can merge on 🍏

PR #2883, opened 2016-07-17
rename container variable, toggling function for clarity in show/hide layers example

skorasaurus:
This PR is for the [show/hide layers example](https://www.mapbox.com/mapbox-gl-js/example/toggle-layers/). 
The function name (addLayer) used in the example is also the name of a method in mapbox-gl-js and the same method is used in the the example. This is confusing to users not very familiar with javascript and myself as well for several minutes.

I also renamed the 'map' variable (in the container?) as 'myMap' because 'map' is already defined in the example, in the css, as an id. 

I've had that issue (of map being used twice) come up as a stumbling block a few times from new users at maptime events for leaflet and mapbox.js that I've conducted and it occasionally confuses me too.

### DIFF
--- a/docs/_posts/examples/3400-01-01-toggle-layers.html
+++ b/docs/_posts/examples/3400-01-01-toggle-layers.html
@@ -52,19 +52,19 @@ description: Create a custom layer switcher to display different datasets.
 <div id="map"></div>
 
 <script>
-var map = new mapboxgl.Map({
+var myMap = new mapboxgl.Map({
     container: 'map',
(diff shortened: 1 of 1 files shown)


lucaswoj:
Hi @skorasaurus! 

I agree that the `addLayer` function is confusing in here. I have (hopefully) made things easier to read by removing the function altogether in f43c892a303e5fc324cb9961d2dfb11c56f8cb3a.

The `map` identifier is a convention we use throughout our examples and documentation. The collision with the HTML id is annoying but benign.

PR #2885, opened 2016-07-18
add marker offset option

andrewharvey:
If you only have PNG assets which you need to appear on a map it looks like using Marker is the best way to do this, however it's handy to set an offset so that the Marker element can be better positioned relative to its LngLat.

Here is an example using this branch http://bl.ocks.org/andrewharvey/01006319700c5352deaad3b58ec53b8c

### DIFF
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -5,22 +5,28 @@ module.exports = Marker;
 
 var DOM = require('../util/dom');
 var LngLat = require('../geo/lng_lat');
+var Point = require('point-geometry');
 
 /**
  * Creates a marker component
  * @class Marker
  * @param {HTMLElement=} element DOM element to use as a marker (creates a div element by default)
+ * @param {Object=} options
+ * @param {PointLike=} options.offset The offset in pixels as a [`PointLike`](#PointLike) object to apply relative to the element's top left corner. Negatives indicate left and up.
  * @example
  * var marker = new mapboxgl.Marker()
  *   .setLngLat([30.5, 50.5])
  *   .addTo(map);
  */
-function Marker(element) {
+function Marker(element, options) {
     if (!element) {
         element = DOM.create('div');
     }
     element.classList.add('mapboxgl-marker');
     this._el = element;
+
+    this._offset = Point.convert(options && options.offset || [0, 0]);
+
     this._update = this._update.bind(this);
 }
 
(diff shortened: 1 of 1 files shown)


js/ui/marker.js
```diff
@@ -5,22 +5,31 @@ module.exports = Marker;
...
     if (!element) {
         element = DOM.create('div');
     }
     element.classList.add('mapboxgl-marker');
     this._el = element;
+
+    this._offset = new Point(
+            ('offset' in options && 'x' in options.offset) ? options.offset.x : 0,
+            ('offset' in options && 'y' in options.offset) ? options.offset.y : 0
+            );
```
mourner: A cleaner way that will cover more cases (including array-like point):

```js
this._offset = Point.convert(options && options.offset || [0, 0]);
```

js/ui/marker.js
```diff
@@ -5,22 +5,31 @@ module.exports = Marker;
...
 var DOM = require('../util/dom');
 var LngLat = require('../geo/lng_lat');
+var Point = require('point-geometry');
 
 /**
  * Creates a marker component
  * @class Marker
  * @param {HTMLElement=} element DOM element to use as a marker (creates a div element by default)
+ * @param {Object} options
+ * @param {Point} options.offset The offset in pixels to apply to this Marker relative to the element's top left corner.
```
mourner: A corresponding Leaflet term is `anchor`, which would be equivalent to `-offset`. Should we align the terminology here, or is `offset` better? Not sure.

lucaswoj: We should align the terminology with Leaflet unless there is a clear advantage to doing otherwise. My impression is that `offset` and `anchor` are equally intuitive concepts.

jfirebaugh: I think "offset" is preferable. Consider the style properties `icon-offset`, `text-offset`, versus `text-anchor` and [the `anchor` option of `Popup`](https://www.mapbox.com/mapbox-gl-js/api/#Popup).

lucaswoj: Sounds like there's a clear advantage to `offset` in consistency with existing terminology in GL JS 👍

andrewharvey:
I've changed the commit.

* Mark options & options.offset as optional in the jsdoc.
* Use PointLike type
* Use suggestion from @mourner Thanks!
* Use point-geometry .add method.
* Improved documentation regarding negative direction.

js/ui/marker.js
```diff
@@ -81,7 +87,7 @@ Marker.prototype = {
 
     _update: function() {
         if (!this._map) return;
-        var pos = this._map.project(this._lngLat);
+        var pos = this._map.project(this._lngLat).add(this._offset);
```
mourner: Minor remark: we can do `_add` instead of `add` here to avoid allocating a new point object.

andrewharvey: Thanks for the tip. I've updated the commit.

PR #2887, opened 2016-07-18
Fix webpack support

lucaswoj:
This PR replaces #2857

fixes https://github.com/mapbox/mapbox-gl-js/issues/2933

cc @bensleveritt @austinhyde @michaelhue @tmcw


webpack.config.example.js
```diff
@@ -61,5 +60,9 @@ module.exports = {
             query: 'brfs'
         }]
     },
-    devtool: 'cheap-source-map' // don't use eval-related options
+
+    // You may use any "devtool" except "eval" and "eval-source-map" due to
```
contra: this should be in the documentation, just wasted an hour trying to figure out why i couldn't get this working because this was missing!

infacq:
is there any way to import the CSS as well

PR #2895, opened 2016-07-19
Update custom marker example

tristen:
custom markers

cc @lucaswoj for review


lucaswoj:
>   112:20  error  Don't make functions within a loop  no-loop-func

We can safely disable this in our examples eslint config

lucaswoj:
🚢  on 🍵

PR #2897, opened 2016-07-19
Pin "vector-tile" dependency

lucaswoj:
... to fix failing test cases due to 1.2.1 -> 1.3.0 update

cc @jfirebaugh


jfirebaugh:
:+1

PR #2905, opened 2016-07-23
minor jsdoc fix marker.setLngLat returns Marker type

andrewharvey:


mollymerp:
thanks @andrewharvey !

PR #2915, opened 2016-07-27
deploy to CN regions

zmully:
I manually updated CN this morning with v0.21.0, but this should be deployed automatically to CN regions on each release.

/cc @lucaswoj


lucaswoj:
LGTM! Thanks @zmully

Issue #2291, opened 2016-03-18
Refactor "SourceCache"


lucaswoj:
The `SourceCache` class could use a big refactor (maybe even a ground-up rewrite) to
- simplify and fix bugs in our wrapped tile retention system #1834
- simplify our tile caching system to make it easier to reason about
- create a cleaner API (the class currently includes some overly-complex methods, surprising side effects, and too-long argument lists)
- improve testability 
- move all code that isn't related to tile caching into separate modules 

ref #2287

jfirebaugh:
I think we should wholly revert the relationship inversion between `Source` and `TilePyramid`/`SourceCache`. Judging by the resulting complex coupling, this seems to have been an architectural mistake.

anandthakker:
@jfirebaugh can you say more about this?  I agree that the current coupling is not ideal, but it's not clear to me that inverting the dependencies would really eliminate it.  Would another way to improve this might be to clarify and simplify the Source interface?

mourner:
Closing as too vague — we should create new issues for any specific things we can improve.

PR #2917, opened 2016-07-28
WIP - share workers across Map instance with a global worker pool

anandthakker:
Refs #899

Before:
without-pool

After:
with-pool


js/source/worker_tile.js
```diff
@@ -24,7 +24,7 @@ function WorkerTile(params) {
 }
 
 WorkerTile.prototype.parse = function(data, layerFamilies, actor, rawTileData, callback) {
-
+    console.log('PARSING TILE', this.source, this.coord.z + '/' + this.coord.x + '/' + this.coord.y, this.uid)
```
anandthakker: For demo purposes only ^

js/source/vector_tile_source.js
```diff
@@ -58,11 +58,11 @@ VectorTileSource.prototype = util.inherit(Evented, {
...
         };
 
-        if (tile.workerID) {
+        if (tile.state === 'loaded') {
             params.rawTileData = tile.rawTileData;
-            this.dispatcher.send('reload tile', params, done.bind(this), tile.workerID);
+            this.dispatcher.send('reload tile', params, done.bind(this), tile.uid);
         } else {
-            tile.workerID = this.dispatcher.send('load tile', params, done.bind(this));
+            this.dispatcher.send('load tile', params, done.bind(this), tile.uid);
```
anandthakker: Note the change to Dispatcher (using the new WorkerPool), wherein to send multiple requests to the same worker, dispatcher.send() accepts a 'key' (`tile.uid` here) and guarantees that two requests sent with the same key go to the same worker.

js/source/worker.js
```diff
@@ -35,14 +61,28 @@ function Worker(self) {
...
+        var shouldDeletePreviousStyle = util.values(this.styles)
+            .some(function (usedKey) { return usedKey === prevKey; });
+        if (shouldDeletePreviousStyle) {
+            delete this.layers[prevKey];
+            delete this.layerFamilies[prevKey];
+        }
+
+        if (this.layers[key]) return;
+
+        var styleLayers = this.layers[key] = {};
```
anandthakker: In this change, `this.layers` and `this.layerFamilies` gain a layer of indirection, becoming maps from a 'style key' to the style's set of layers and layer families.  This allows a Worker to track multiple map instances' styles -- but by keying them off of a 'style key' rather than a map instance id, we allow for two map instances with the same style to share work.

anandthakker:
@lucaswoj added some comments and notes in the diff -- I think this is ready for a first 👀  to validate / modify / reject the approach.

mollymerp:
wow! thank you @anandthakker! at first glance -- and those gifs!! -- this is really impressive.

anandthakker:
Thanks @mollymerp !

Let me know if y'all think this approach is viable (generally speaking), and if so, I'll take another pass to do some cleanup, benchmarking, and unit tests.

js/source/geojson_source.js
```diff
@@ -138,14 +138,19 @@ GeoJSONSource.prototype = util.inherit(Evented, /** @lends GeoJSONSource.prototy
             options.data = JSON.stringify(data);
         }
 
+        // We use a 'worker key' that's tied to the current geojson data for
+        // this source, so that when we ask the worker for tiles, the request
+        // goes to the same worker that parsed/prepared the geojson.
+        var workerKey = options.url || options.data;
```
lucaswoj: It looks like we reference this `workerKey` against a map. Will this work of `options.data` is an object?

```js
var object = {};

object[{foo: true}] = 'foo';
object[{bar: true}] = 'bar';

object[{foo: true}] // => 'bar';
object['[Object object]'] // => 'bar'
```

anandthakker: I think the line 138 above ensures that this won't happen -- but now that I'm looking at this again, I do think it's not especially clear; I'll make it clearer.

lucaswoj:
Before I delve into a line-by-line review here, may I ask a few questions about this PR? 
 
 - How is the global worker pool architected? 
 - Could we get a similar effect by deduplicating `SourceCache`s across maps in the main thread? (I recognize this may require some major changes to `SourceCache` #2291) 
 - If a user calls `GeoJSONSource#setData`, will all maps change?
 - If two maps use the same vector tile, how to you avoid double-transferring the `StructArray`s? 
 - How does this handle the case where two maps use the same vector tile but have different styles (which require different `StructArray` layouts)?

js/source/geojson_source.js
lucaswoj: Ah! I forgot that we stringified it. LGTM 👍

js/util/dispatcher.js
```diff
@@ -1,57 +1,57 @@
...
-    workerBus.importScripts = function () {};
-
-    this.worker = new Worker(workerBus);
-    this.actor = new Actor(parentBus, parent);
-
-    this.remove = function() {
-        parentListeners.splice(0, parentListeners.length);
-        workerListeners.splice(0, workerListeners.length);
-    };
+    this.id = util.uniqueId();
```
anandthakker: bug: this id needs to be unique across map instances, but `util.uniqueId()` isn't.

anandthakker:
> If two maps use the same vector tile, how to you avoid double-transferring the StructArrays?

@lucaswoj SHOOT, yes, I completely missed this, and I think it might be a dealbreaker for this approach.

([Here](https://github.com/mapbox/mapbox-gl-js/pull/2917/files#diff-eb0e519da9b1317b69f900e8f7ed0121R62), in trying to handle the race condition of one instance calling `load tile` while a load is already in flight, I mistakenly dropped the transferables from the callback args, so I never tripped the double-transfer)

Maybe handling this will be best done by scrapping this approach and starting with something like this suggestion:

> Could we get a similar effect by deduplicating SourceCaches across maps in the main thread? (I recognize this may require some major changes to SourceCache #2291)

Closing this, but I'll dump a summary of the approach I started here, in case there are any parts of it that might be useful salvage or discussion:


## WorkerPool

Each `Map` instance has a single `Dispatcher` instance.  Workers are provided to the dispatcher via `WorkerPool`, which is a per-script (and thus almost always per-page) singleton.

Like before, dispatchers access workers by way of `Actor`s, which are what the `WorkerPool` actually provides to it.  I.e.:

```
                      +-------+    +--------+    +-------+
                   +--+ ACTOR +----+ WORKER +----+ ACTOR +--+
                   |  +-------+    +--------+    +-------+  |
                   |                                        |
                   |  +-------+    +--------+    +-------+  |
+------------+     +--+ ACTOR +----+ WORKER +----+ ACTOR +--+     +------------+
|            |     |  +-------+    +--------+    +-------+  |     |            |
| DISPATCHER +-----+                                        +-----+ DISPATCHER |
|            |     |  +-------+    +--------+    +-------+  |     |            |
+------------+     +--+ ACTOR +----+ WORKER +----+ ACTOR +--+     +------------+
(main thread)      |  +-------+    +--------+    +-------+  |     (main thread)
                   |                                        |
                   |  +-------+    +--------+    +-------+  |
                   +--+ ACTOR +----+ WORKER +----+ ACTOR +--+
                      +-------+    +--------+    +-------+
```

(This is a logical diagram -- in reality, each `ACTOR` above is really two actors--one on the main thread and one on the worker thread.)

## Actor and Worker

`Actor` remains mostly the same; only change here is that it accepts a `parentId` identifying the `Dispatcher` (and thus `Map`) it belongs to, and that it [passes this `parentId` along as part of the messages it sends](https://github.com/mapbox/mapbox-gl-js/blob/e07e77fb4e9e4a8190153fb3f222ad7d2cdaae31/js/util/actor.js#L59).  (The parentId for the worker-side Actor is just `null`.)

Changes to `Worker` are more involved:
 - All target tasks (`'load tile'`, `'set style'`, etc.) now take a `mapId`
   parameter. (This is the `parentId` of the calling Actor; the nomenclature
   here stinks, I know.)
 - `set style` and `update style` now take care of keeping track of the style
   layers (and layer families) for multiple `Map`s.  The most straightforward
   way to do this would have been simply to maintain an object hash from
   `mapId` to the `layers` and `layerFamilies` data for that map.  However,
   this would fail to improve the redundant VT parsing that was the original
   goal, because VT parsing depends on this style information.
 - Therefore, there's extra layer of indirection:
   - (a) The worker tracks a mapping from `mapId` to a "style key"; a style key
     corresponds uniquely to a given map _style state_, but may be shared in
     common among multiple map instances.
   - (b) `layers` and `layerFamilies` are tracked by style key.
 - `VectorTileWorkerSource` now tracks loading/loaded tiles with `[(style
   key)+(source id)][uid]` rather than just `[source id][uid]`.


## Smaller Changes
 - https://github.com/mapbox/mapbox-gl-js/pull/2917/commits/2cc58b4691053541a5acb162b4447cc5abe44ce0
   this just moves the point of bifurcation between node and browser to a
   smaller `web_worker.js` wrapper, so that `dispatcher.js` can be a single
   file.
 - https://github.com/mapbox/mapbox-gl-js/pull/2917/files#r72672765 - instead
   of returning a `workerID` and then using that to retarget the same worker
   that loaded a tile, we use a scheme where the tile's coordinates determine
   its uid, and its uid determines which worker does the work.  This lets us
   ensure that two map instances attempt to use the same worker to to load a
   given tile.


## Other Questions

> If a user calls GeoJSONSource#setData, will all maps change?

Right now, I think that's right: GeoJSONSource is still TBD.

> How does this handle the case where two maps use the same vector tile but have different styles (which require different StructArray layouts)?

This, I think, is handled by the "style key" system above.

mollymerp:
Aw shucks 😢. Thanks for your work on this nonetheless!

PR #2922, opened 2016-07-29
upgrade geojson-vt

### LINKED ISSUE #2912: Render features outside the range of real longitudes
We're having a problem in `mapbox-gl-draw`: https://github.com/mapbox/mapbox-gl-draw/issues/458. If you spin the globe a time or two and then try to draw, you'll create coordinates outside the real range of longitudes, like this point:

```json
{
	"id": "9fff69e4acb4b4e9d0175eaa7c7255ca",
	"type": "Feature",
	"properties": {},
	"geometry": {
		"coordinates": [460.6260419749865, 59.16322869014476],
		"type": "Point"
	}
}
```

Then when we try to render the features, GL JS errors.

screen shot 2016-07-27 at 8 48 33 am

The stack trace isn't very helpful here, but I thought someone who works on this library regularly might know the reason for the error.

I've tried some experiments in Draw to normalize positions as you create them, so that they always fall within the real range; but those failed because of lines and polygons. Those features might actually want to stretch around the world. And other weird things can happen because of the attempt at normalization (esp. lines wrapping the opposite direction you wanted them to wrap). 

So the simplest solution seems to be to allow rendering of these positions. Do you think that is a possibility?

cc @mcwhittemore, @tmcw, @samanpwbb

mcwhittemore:
In version 2.3.0 of geojson-vt a fix was added to make sure any geojson between +/-540 lng would be rendered. Before, if there were no features between +/-180 an undefined error would be thrown.

As this is tested upstream, I'm not sure if we want tests here. I can add them if you want.

This closes #2912.

### DIFF
--- a/test/js/ui/map.test.js
+++ b/test/js/ui/map.test.js
@@ -883,26 +883,34 @@ test('Map', function(t) {
 
     t.test('error event', function (t) {
         t.test('logs errors to console when it has NO listeners', function (t) {
+            var map = createMap({ style: { version: 8, sources: {}, layers: [] } });
+
+            sinon.spy(map, 'fire');
             sinon.stub(console, 'error', function(error) {
-                console.error.restore();
-                t.equal(error.message, 'version: expected one of [8], 7 found');
-                t.end();
+                if (error.message === 'version: expected one of [8], 7 found') {
+                    t.notOk(map.fire.calledWith('error'));
+                    console.error.restore();
+                    map.fire.restore();
+                    t.end();
+                } else {
+                    console.log(error);
+                }
             });
 
-            createMap({ style: { version: 7, sources: {}, layers: [] } });
+            map.setStyle({ version: 7, sources: {}, layers: [] });
         });
 
         t.test('calls listeners', function (t) {
-            sinon.stub(console, 'error', function(event) {
-                t.fail(event.error);
-            });
-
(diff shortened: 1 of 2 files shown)


mcwhittemore:
Tests are passing locally.

test/js/ui/map.test.js
```diff
@@ -883,26 +883,34 @@ test('Map', function(t) {
...
-                console.error.restore();
-                t.equal(error.message, 'version: expected one of [8], 7 found');
-                t.end();
+                if (error.message === 'version: expected one of [8], 7 found') {
+                    t.notOk(map.fire.calledWith('error'));
+                    console.error.restore();
+                    map.fire.restore();
+                    t.end();
+                } else {
+                    console.log(error);
```
mourner: When would that console.log happen? Could we do an assertion instead? like `t.fail(error.message)` or something.

lucaswoj: Modifying the global `console.error` method is tricky. This test was failing due to errors from other tests' map instances being emitted asynchronously after their test had completed. In this case, we don't want this test to fail and we do want the error message printed to the console (either the error message contains useful information or it shouldn't have been emitted in the first place).

lucaswoj:
@mourner 🚢  when you're satisified with https://github.com/mapbox/mapbox-gl-js/pull/2922#discussion_r73224022

PR #2923, opened 2016-07-30
fix contour button

colleenmcginnis:
`var id` was stuck at the id of the last item in the `for` loop so both buttons were toggling the museum layer. Created a `var clickedLayer` to get id on click.


mollymerp:
Thanks @colleenmcginnis !! This looks good to me, you've just got a tiny linting error causing the build to fail. 

```bash
/home/ubuntu/mapbox-gl-js/docs/_posts/examples/3400-01-01-toggle-layers.html
  113:44  error  Missing semicolon  semi
```

colleenmcginnis:
Oh! I see it now. Thanks @mollymerp! Take a look - I think I fixed it.

lucaswoj:
Thank you @colleenmcginnis! We appreciate the bug fix.

PR #2930, opened 2016-08-02
Increase the maximum size of glyph atlas

xrwang:
Enlarging the glyph sheet allows CJK characters to render and not overflow. Works with https://github.com/mapbox/mapbox-gl-test-suite/pull/131 tests.

Performance of this PR on bench:
Buffer, 803 ms
FPS, 60 frames per second

cc @amyleew


bhousel:
Yay!  Been waiting a long time for this..

1. I think this might necessitate a corresponding change in https://github.com/mapbox/mapbox-gl-shaders
2. If you're going to do this, I'd raise the limit to 4096x4096.

debug/cjk.html
```diff
@@ -0,0 +1,65 @@
...
+    style: 'mapbox://styles/mapbox/streets-v9',
+    hash: true
+});
+
+map.addControl(new mapboxgl.Navigation());
+
+map.on('load', function () {
+    map.addSource("points", {
+        "type": "geojson",
+        "data": 'https://raw.githubusercontent.com/mapbox/mapbox-gl-test-suite/cjk-overflow/data/cjkPts.geojson'
```
lucaswoj: You can change this line to `mapbox-gl-test-suite/cjk-overflow/data/cjkPts.geojson`

js/data/bucket/symbol_bucket.js
```diff
@@ -62,11 +62,11 @@ var layoutVertexArrayType = new Bucket.VertexArrayType([{
 }, {
     name: 'a_data1',
     components: 4,
-    type: 'Uint8'
+    type: 'Uint16'
 }, {
     name: 'a_data2',
     components: 2,
-    type: 'Uint8'
+    type: 'Uint16'
```
lucaswoj: The size of this array has a huge effect on GL JS performance. 

There are a couple of members of `data1` that don't need their size increased and no members of `data2` that need their size increased. 

I encourage you to take a closer look at the [code that defines the array structure](https://github.com/mapbox/mapbox-gl-js/blob/upgrade-geojson-vt/js/data/bucket/symbol_bucket.js#L54-L55) and the [code that populates the array](https://github.com/mapbox/mapbox-gl-js/blob/upgrade-geojson-vt/js/data/bucket/symbol_bucket.js#L74-L90). Is there any way we can modify this code to make the array smaller? Make the code more readable? Think big 😄 .

js/symbol/glyph_atlas.js
```diff
@@ -115,9 +115,8 @@ GlyphAtlas.prototype.resize = function() {
     var origw = this.width,
         origh = this.height;
 
-    // For now, don't grow the atlas beyond 1024x1024 because of how
-    // texture coords pack into unsigned byte in symbol bucket.
-    if (origw > 512 || origh > 512) return;
+    // Grow atlas to 2048 for CJK characters
+    if (origw > 1024 || origh > 1024) return;
```
lucaswoj: This would be more readable if this magic number were pulled out into a constant and documented.

bhousel: It is perfectly readable with the original comment.

lucaswoj:
> 1. I think this might necessitate a corresponding change in https://github.com/mapbox/mapbox-gl-shaders

Fortunately, shaders are not coupled to the bit-width of these attributes. https://github.com/mapbox/mapbox-gl-shaders/blob/master/src/icon.vertex.glsl#L11-L12

lucaswoj:
> Performance of this PR on bench:
> Buffer, 803 ms
> FPS, 60 frames per second

These benchmarks are machine-specific and therefore relative numbers. Can you re-run these benchmarks against both this branch and `master`? (This'll be easier when we get around to https://github.com/mapbox/mapbox-gl-js/issues/2405)

bhousel:
> Fortunately, shaders are not coupled to the bit-width of these attributes. https://github.com/mapbox/mapbox-gl-shaders/blob/master/src/icon.vertex.glsl#L11-L12

Isn't the glyph shader
https://github.com/mapbox/mapbox-gl-shaders/blob/master/src/sdf.vertex.glsl

But either way, that is helpful that the bit width doesn't need to change in the shader code.

> The size of this array has a huge effect on GL JS performance.
> 
> There are a couple of members of data1 that don't need their size increased and no members of data2 that need their size increased.

If performance is a concern, could we dynamically adjust `layoutVertexArrayType` from `Uint8` to `Uint16` in realtime as the atlas increases in size?  We only need it to be `Uint16` for very large symbol and glyph atlases.

lucaswoj:
> If performance is a concern, could we dynamically adjust `layoutVertexArrayType` from `Uint8` to `Uint16` in realtime as the atlas increases in size?  We only need it to be `Uint16` for very large symbol and glyph atlases.

This is a great idea and is theoretically possible. Implementing it will require some bigger architectural changes. Let's keep this idea in our back pocket.

xrwang:
Thanks for all the detailed feedback! Updated this PR per comments. Bench results on master versus branch below. Please let me know if I've missed / misunderstood something : )

**master**
---
**buffer**: 795 ms

**fps**: 60 frames per second

**frame-duration**: 
5.4 ms per frame at zoom 4. 0% of frames took longer than 16ms.
5.2 ms per frame at zoom 8. 0% of frames took longer than 16ms.
7 ms per frame at zoom 11. 0% of frames took longer than 16ms.
8.6 ms per frame at zoom 13. 0% of frames took longer than 16ms.
8.3 ms per frame at zoom 15. 0% of frames took longer than 16ms.
5.6 ms per frame at zoom 17. 0% of frames took longer than 16ms.

avg: 6.7 ms per frame. 0% of frames took longer than 16ms.


**this pr**
---

**buffer**: 833 ms

**fps**: 60 frames per second

**frame-duration**:
5.5 ms per frame at zoom 4. 0% of frames took longer than 16ms.
5.2 ms per frame at zoom 8. 0% of frames took longer than 16ms.
7.1 ms per frame at zoom 11. 0% of frames took longer than 16ms.
9.3 ms per frame at zoom 13. 0% of frames took longer than 16ms.
8.1 ms per frame at zoom 15. 0% of frames took longer than 16ms.
5.2 ms per frame at zoom 17. 0% of frames took longer than 16ms.

avg: 6.7 ms per frame. 0% of frames took longer than 16ms.

lucaswoj:
This is looking great @xrwang! A 5% slowdown in buffer creation time (795 ms -> 833 ms) is a good starting point. Do you think there's anything we can do to make the slowdown even smaller? Did you verify that the slowdown is related to the increase in `StructArray` size?

xrwang:
@lucaswoj, yup, I can definitely confirm that the slow down is related to the increase in `StructArray` size. Though from benching several times, it might also be related to the glyph atlas size itself, ie 2048 vs 4096? Although technically that shouldn't affect it, only the `StructArray` size should ? Anyhow, happy to keep it at 2048 for now to veer on the cautious side. benched several times below, and this branch with uint16, 2048 seems very close to the master branch's buffer time. 

**benching sequence, results in order***
1. master, buffer: 774ms

2. this branch, with uint 16, 4096 glyph atlas, buffer: 830ms

3. this branch, with uint 8, 4096 glyph atlas, buffer: 777 ms

4. this branch, with uint16, 2048 glyph atlas, buffer:  766 ms *

5. master, buffer: 765 ms

6. this branch with uint16, 2048 glyph atlas, buffer: 792 ms *

7. master, buffer: 784 ms

lucaswoj:
This looks great! Thank you for doing these benchmarks @xrwang. 

>  Though from benching several times, it might also be related to the glyph atlas size itself, ie 2048 vs 4096? Although technically that shouldn't affect it, only the `StructArray` size should ?

It is possible and looks to be the case that the glyph atlas size has a large effect on performance too. If a width of 2048 is enough for real world usage with CJK fonts then I'm happy with this result  and almost ready to ship this PR.

# Next Steps

 - [ ] confirm that a glyph atlas size of 2048 is big enough for real-world usage with CJK fonts
 - [ ] look for low-hanging preformance fruit in `GlyphAtlas` (specifically the "realloc" algorithm)
 - [ ] merge https://github.com/mapbox/mapbox-gl-test-suite/pull/131 into `master`
 - [ ] update `mapbox-gl-test-suite` in `package.json` in this PR to include https://github.com/mapbox/mapbox-gl-test-suite/pull/131
 - [ ] merge this PR into master

js/data/bucket/symbol_bucket.js
```diff
@@ -61,11 +61,11 @@ var layoutVertexArrayType = new Bucket.VertexArrayType([{
     type: 'Int16'
 }, {
     name: 'a_data1',
```
lucaswoj: What do you think about renaming this to something more self-documenting like `a_texture_pos`?

xrwang: 👍 sounds good, it'll be clearer that way.

xrwang:
cool, updated per comments. The name change for a_texture_pos is over here: https://github.com/mapbox/mapbox-gl-shaders/pull/22

Tests over on https://github.com/mapbox/mapbox-gl-test-suite/pull/131 should be ready.

Benched with the resizing by 4 (instead of 2) and performance improved a lot!

**master**
806 ms

**this branch**
796 ms

**master**
814 ms

**this-branch**
822 ms

brunoabinader:
Created https://github.com/mapbox/mapbox-gl-native/issues/5873 to track these changes in GL native 👍

lucaswoj:
# Next Steps

 - [ ] update test-suite SHA in GL JS
 - [ ] rename `a_data2` attribute to `a_data`
 - [ ] update shaders SHA in GL JS
 - [ ] squash & rebase GL JS 
 - [ ] run GL JS benchmarks
 - [ ] do a manual test of GL JS on debug pages
 - [ ] bring GL Native PR in sync with `a_data2` shader changes
 - [ ] 🍏 
 - [ ] 🚢 

# Related PRs

 - https://github.com/mapbox/mapbox-gl-test-suite/pull/131
 - https://github.com/mapbox/mapbox-gl-shaders/pull/22
 - https://github.com/mapbox/mapbox-gl-native/pull/5908/files

lucaswoj:
Pushing squashed & based version. Previous HEAD was db7c0d8bd85bc1a4e8622e10c902664c248d3f5f

lucaswoj:
# Branch Benchmarks

**buffer:** 960 ms
**fps:** 60 fps
**frame-duration:** 7 ms, 0% > 16ms
**query-point:** 0.67 ms
**query-box:** 50.55 ms
**geojson-setdata-large:** 114 ms

# `master` Benchmarks

**buffer:** 922 ms
**fps:** 60 fps
**frame-duration:** 6.9 ms, 0% > 16ms
**query-point:** 0.71 ms
**query-box:** 50.73 ms
**geojson-setdata-large:** 111 ms

js/symbol/glyph_source.js
```diff
@@ -28,7 +28,7 @@ GlyphSource.prototype.getSimpleGlyphs = function(fontstack, glyphIDs, uid, callb
         this.stacks[fontstack] = {};
     }
     if (this.atlases[fontstack] === undefined) {
-        this.atlases[fontstack] = new GlyphAtlas(128, 128);
```
bhousel: please leave this as 128,128.. most latin alphabets fit in this

lucaswoj: @bhousel It's still 128! I moved this constant inside `GlyphAtlas` because there are some tricky dependencies between the default size, growth factor, and max size. Having them all in one place makes that clearer.

bhousel: oh sorry!! you're right! :)

Issue #2924, opened 2016-07-31
Fix bug preventing map interactions near two controls in the same corner


yunluyl:
The navigation controller layer is above the map, which stops the mouse event going to the map layer.

Here is the JSFiddle. You can try hovering your mouse at the bottom right corner of the map.

https://jsfiddle.net/6wrqz1ru/

mollymerp:
Hi @yunluyl,

Thank you for reporting your issue. I'm not sure there is a bug is in this case, because when I opened the jsfiddle I did not experience any unexpected behavior. Any DOM element above the GL canvas will usually prevent user interaction with the map. When you have navigation control in the bottom right, it gets put in the same `div` as the map attribution (`.mapboxgl-ctrl-bottom-right`). this causes the div to grow to fit both controls and covers more map area than you might expect. 

<img width="622" alt="screen shot 2016-07-31 at 10 18 47 pm" src="https://cloud.githubusercontent.com/assets/2425307/17284132/d4ee0752-576c-11e6-9e0b-4d1b9585e03a.png">

I suggest putting the navigation control in another corner to avoid this behavior.

<img width="473" alt="screen shot 2016-07-31 at 10 20 48 pm" src="https://cloud.githubusercontent.com/assets/2425307/17284153/12b8321a-576d-11e6-9be4-405fb9e6d6c4.png">

Thanks!
Molly

lucaswoj:
We should be able to support multiple controls in the same corner. This is a bug.

mourner:
This should be an easy task to tackle — the trick is to update CSS to have `pointer-events: none` on the corner elements but reset to `auto` on its descendants. Here are the [relevant lines from Leaflet](https://github.com/Leaflet/Leaflet/blob/master/dist/leaflet.css#L106-L117). A PR would be welcome!

yunluyl:
PR created

mollymerp:
This was fixed in #2932 thank you @yunluyl!

PR #2932, opened 2016-08-02
Fixed issue #2924

yunluyl:
Fix bug preventing map interactions near two controls in the same corner #2924

### DIFF
--- a/dist/mapbox-gl.css
+++ b/dist/mapbox-gl.css
@@ -21,13 +21,13 @@
 .mapboxgl-ctrl-top-left,
 .mapboxgl-ctrl-top-right,
 .mapboxgl-ctrl-bottom-left,
-.mapboxgl-ctrl-bottom-right  { position:absolute; }
+.mapboxgl-ctrl-bottom-right  { position:absolute; pointer-events:none; }
 .mapboxgl-ctrl-top-left      { top:0; left:0; }
 .mapboxgl-ctrl-top-right     { top:0; right:0; }
 .mapboxgl-ctrl-bottom-left   { bottom:0; left:0; }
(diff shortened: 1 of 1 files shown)


lucaswoj:
Sorry about the failure! That's an upstream problem that we're fixing in https://github.com/mapbox/mapbox-gl-js/pull/2922 right now.

dist/mapbox-gl.css
```diff
@@ -21,13 +21,13 @@
 .mapboxgl-ctrl-top-left,
 .mapboxgl-ctrl-top-right,
 .mapboxgl-ctrl-bottom-left,
-.mapboxgl-ctrl-bottom-right  { position:absolute; }
+.mapboxgl-ctrl-bottom-right  { position:absolute; pointer-events:none; z-index:2; }
```
lucaswoj: What prompted the `z-index` change?

yunluyl: I saw in leaflet they have z-index defined for the controllers.

.leaflet-control {
	position: relative;
	z-index: 800;
	pointer-events: visiblePainted; /* IE 9-10 doesn't have auto */
	pointer-events: auto;
	}
.leaflet-top,
.leaflet-bottom {
	position: absolute;
	z-index: 1000;
	pointer-events: none;
	}

Then I saw the popup tip has z-index 1. z-index of 2 will make sure the controllers are on top of the popup tip.

.mapboxgl-popup-tip {
    width: 0;
    height: 0;
    border: 10px solid transparent;
    z-index: 1;
}

lucaswoj:
Looks good! Thank you for the contribution @yunluyl!

Issue #2729, opened 2016-06-13
Enable property functions for line paint properties


lucaswoj:
## Properties
- `line-opacity`
- `line-color`
- `line-width`
- `line-gap-width`
- `line-offset`
- `line-blur`
- `line-dasharray`

sguignot:
Below you will find a simple use case, using line-color as a feature property reference, the same way text-field does (using `resolveTokens` function): `'line-color': '{colour}'`

```
export const addTransitLayerToMap = (map) => {
  const source = {
    type: 'vector',
    tiles: ["https://vector.mapzen.com/osm/transit/{z}/{x}/{y}.mvt?api_key=API_KEY"]
  };
  map.addSource(transitSourceId, source);

  const localTransitFilter = ['in', 'kind', 'subway', 'tram', 'railway'];
  map.addLayer({
    id: transitLayerId,
    source: transitSourceId,
    'source-layer': 'transit',
    type: 'line',
    paint: {
      'line-color': '{colour}',
      'line-width': {
        base: 1,
        stops: [[10, 1], [13, 2], [16, 4]]
      },
      'line-gap-width': 0,
    },
    filter: localTransitFilter,
  });
};
```

@lucaswoj  please let me know if this is on your roadmap, if not I could work on it I guess :)

lucaswoj:
@sguignot We do not plan to support that syntax for property functions. We prefer using a syntax based on Javascript objects to a syntax based on a custom minilanguage.

sguignot:
@lucaswoj ok, it makes sense! I will use it as soon as it is released :)

PR #2938, opened 2016-08-03
Add support for property functions in line-color

lucaswoj:
... and pave the way for all other line properties 

This PR 

 - adds support for `line-color` property functions
 - builds the infrastructure needed for making other line properties support property functions

~~The naive implementation of line-color property functions was too slow, prompting me to investigate ways to reduce the perf overhead of setting uniforms per tile. We need to set uniforms per tile as if each were a different GL program because different tiles may actually require different GL programs due to a pending asynchronous operation.~~

~~My solution was a `Painter#setUniforms(uniforms: Object)` method, something I've been thinking about writing for a while anyway as part of #145 (needed for https://github.com/mapbox/mapbox-gl-js/issues/281).~~

ref https://github.com/mapbox/mapbox-gl-js/issues/2729

cc @mollymerp @mourner


js/data/bucket/line_bucket.js
```diff
@@ -347,6 +359,12 @@ LineBucket.prototype.addLine = function(vertices, join, cap, miterLimit, roundLi
         startOfLine = false;
     }
 
+    this.populatePaintArrays(
+        'line', {zoom: this.zoom},
+        feature.properties,
```
mourner: If we only use `feature.properties`, should we pass `featureProperties` in the outer function directly?

js/render/draw_line.js
```diff
@@ -119,45 +36,122 @@ module.exports = function drawLine(painter, source, layer, coords) {
...
+        'u_fade': image.t,
+        'u_pattern_size_a': [
+            pixelsToTileUnits(tile, imagePosA.size[0] * image.fromScale, painter.transform.tileZoom),
+            imagePosB.size[1]
+        ],
+        'u_pattern_size_b': [
+            pixelsToTileUnits(tile, imagePosB.size[0] * image.toScale, painter.transform.tileZoom),
+            imagePosB.size[1]
+        ]
+    });
```
mourner: Since this is in a hot path, it might be worth trying to rewrite that as multiple `painter.setUniform(name, value)` statements to avoid the `for in` loop and temporary objects. I would even do 4 different methods (`setUniform2`, etc.) to avoid temporary arrays or the `arguments` keyword.

js/render/painter/bind_program.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    return nextProgram;
+};
+
+function applyPragmas(source, pragmas) {
+    return source.replace(/#pragma mapbox: ([\w]+) ([\w]+) ([\w]+) ([\w]+)/g, function(match, operation, precision, type, name) {
+        return pragmas[operation][name].replace(/{type}/g, type).replace(/{precision}/g, precision);
+    });
+}
```
mourner: This file looks like a duplicate of `use_program` that is not used anywhere. Did you intend to rename the file?

mourner:
This looks great overall. How are the benchmarks looking?

js/render/painter/bind_program.js
lucaswoj: Oops. This file shouldn't be included. Removed in the next commit.

js/data/bucket/line_bucket.js
lucaswoj: Refactored in the next commit.

lucaswoj:
The benchmarks are looking pretty good. I was seeing around 5% slowdown in `frame-duration`, everything else 0%. I think `setUniforms` is the source of the slowdown and there are opportunities to make it faster, as you pointed out above. https://github.com/mapbox/mapbox-gl-js/pull/2938#discussion_r73764804

mollymerp:
Daaang awesome job keeping perf 🏃 !

lucaswoj:
Perf on the latest commit is 👌 

# `property-function-line` Benchmarks

**frame-duration:** 7.2 ms, 0% > 16ms

# `master` Benchmarks

**frame-duration:** 8.5 ms, 11% > 16ms

lucaswoj:
I got this running as fast or faster than `master` but there's a lingering test-suite failure that I can't understand. Any ideas @mourner?

lucaswoj:
I found the problem with `line-dasharray long-segment` and fixing it regressed all the perf gains. Pperf hasn't improved at all since the initial implementation in cd784123d1ed4a6b24f3ea27fa44b2178d71d62d. I recommend we consider merging that as-written.

mourner:
> fixing it regressed all the perf gains

Interesting, what was the fix about? It's hard to tell because of the rebased commit.

lucaswoj:
Ah! My apologies. I've gotten into the habit of squashing bugfixes. I'll unsquash and push so you can see. In short, there was a mistake leading to early loop termination. The performance gains / losses make a lot of sense.

lucaswoj:
The fix for the test is in 62564e80083e94bd24d597d4ec55f7863669be1b. It's obvious how this led to artificial performance gains 😢 

I am going to remove all ineffective perf optimizations from this branch, leaving only the initial commit.

mourner:
Oh wow. It's fascinating that this was caught by just a single obscure test. :) It may be worth adding a "functional" render test that renders a relatively detailed map tile with all stuff at once to catch things like this.

I agree that we can merge this now — the changes look good, and we can follow-up with perf improvements later if they're possible.

lucaswoj:
> It may be worth adding a "functional" render test that renders a relatively detailed map tile with all stuff at once to catch things like this.

Yes! https://github.com/mapbox/mapbox-gl-test-suite/issues/26

sguignot:
@lucaswoj Thanks for this :)
Is there a way to use directly the color from a 'color' property of the feature?

Note: I made a few tests using categorical stops but it's a bit heavy when rendering railways from OSM since they use hundreds of different colors in some areas.

lucaswoj:
@sguignot the closest thing is a proposal for `identity` functions https://github.com/mapbox/mapbox-gl-function/issues/20. PRs welcome!

PR #2949, opened 2016-08-05
Use tile uid to associate tiles with workers

anandthakker:
Instead of returning a workerID and then using that to retarget the same
worker that loaded a tile, use a scheme where the tile's coordinates
determine its uid, and its uid determines which worker does the work.


anandthakker:
Ack, sorry -- didn't mean to open this yet.

Issue #2951, opened 2016-08-05
Share parsed tiles across maps


lucaswoj:
When one page contains many maps which use the similar styles to show the same geographic region we can improve performance by sharing parsed tiles between maps. 
#1. Create Globally Unique Tile Identifier

Tiles can be uniquely identified by a hash of their
- coordinate
- source's URL or data
- style layer's `id` (to accommodate the `paintVertexArray`s)
- style layer's `type`
- style layer's `layout` properties
- style layer's `paint` properties which are property-functions 

I'm not sure how exactly to handle ref families. The easiest solution is to have an entry in the tile cache per layer, not per ref family. Every entry within the same ref family would still point to the same `Tile` object. 

This could replace the existing `Tile#id` property 
#2. Create Shared Tile Cache

The simplest way to implement this is to replace the `SourceCache#_cache` instance property with a static `SourceCache.cache` property, keyed by the globally unique tile identifier ☝️. We will need to tweak the cache's retention policy to suit. 

We may need to implement this cache as a locking cache to prevent simultaneous requests for the same time from spawning multiple requests. 

cc @anandthakker

lucaswoj:
Related to #899

anandthakker:
👍 

@lucaswoj thinking about this in the context of custom sources, it probably makes sense to implement 1 in a way that allows the `Style` and the `Source` to share responsibility for generating a tile's unique id.

anandthakker:
Hmm -- @lucaswoj right now, for a given coord, the source cache holds one `Tile` per _source_--not per style layer.  It seems to me like the "tiles" you're describing above are really closer to `Bucket`s.

It _almost_ seems like instead of caching `Tile`s, we could cache individual `Bucket`s--keyed off of a hash like you described above--and then use them to construct `Tile`s on the fly... except that `Tile`s also have stuff like:

 - collisionBoxArray
 - collisionTile
 - symbolInstancesArray
 - symbolQuadsArray
 - featureIndex

I don't yet know enough about how these work to guess whether they would make such a strategy infeasible.  What do you think?

lucaswoj:
@anandthakker Ah. Good thinking. We'll need to think a little bigger in terms of the architectural changes needed to support this feature (#2432 would do the trick)

andrewharvey:
see also https://stackoverflow.com/questions/47891132/create-several-mapbox-gl-maps-that-show-different-markers-using-single-map-insta/

mourner:
[...]

PR #2948, opened 2016-08-05
Unify node/browser dispatcher code

anandthakker:
Instead of having a mock Dispatcher for node, pull out the relevant code
into a small wrapper, web_worker.js, which contains the node/browser
code split to a smaller scope.

PR #2952, opened 2016-08-06
Add a minimal global worker pool

### LINKED ISSUE #899: Create a global worker pool

anandthakker:
Closes #899 

@lucaswoj this is a much more focused / less ambitious change than #2917.  It adds a global worker pool, making only the minimal changes necessary to allow a Worker to serve multiple map instances correctly. 

It does _not_ attempt to deduplicate the work that is done for different map instances.  Nonetheless, there are benefits to sharing workers.  Immediate benefit: the overhead to create a worker is significant, because it essentially involves copying all of the worker code over from the main thread.  Future benefit: using shared workers paves the way for more ambitious performance inter-map-instance improvements like #2951 

## Specific Notes
 - The majority of the +505 lines come from new tests, benchmark, and debug page.
 - Essentially, the simple strategy here is to change the worker's state-keeping objects `Worker#layers`, and `Worker#layerFamilies` into objects keyed off of `mapId`.
 - The WorkerSource abstraction helps here, because it allowed for implementing per-map state management for things like geojson-vt indexes and cached WorkerTiles, but with pretty minimal code changes.  Before this, we instantiated a single `WorkerSource` instance for each source type.  Now,  when a WorkerSource is registered, we just hold on to the class, and instantiate a separate one for each map (or rather, for each `mapId`).  Doing it this way means that none of the code in GeoJSONWorkerSource or VectorTileWorkerSource had to be touched.
 - The `workerCount` Map option is removed in favor of `mapboxgl.WorkerPool.WORKER_COUNT`

## Checklist

 - [x] Add unit test to assert that `Dispatcher` correctly consumes `WorkerPool`
 - [x] Add unit tests for `WorkerPool`
 - [x] Add unit test for two `Actor`s using the same underlying `Worker`.
 - [x] **(n/a)** [document any changes or additions to public APIs](https://github.com/mapbox/mapbox-gl-js/blob/master/docs/README.md)
 - [x] [post scores for all benchmarks on your branch and `master`](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/README.md#running-benchmarks)
 - [x] get a PR review :+1: / :-1:


anandthakker:
Note that the first commit here is equivalent to #2948

anandthakker:
Rebasing, previous at https://github.com/mapbox/mapbox-gl-js/pull/2952/commits/d3d0787c835af4b10201bde0d0f7a020011b2da3

anandthakker:
FWIW: I'm using this branch in a project where we have multiple maps on a page, and even just having a shared pool--without actually deduping any of the work--significantly reduces the initial load time, because initializing each Worker is a pretty significant overhead.

anandthakker:
Next steps:
 - [ ] 👍 / 👎 from @lucaswoj 
 - [ ] Add unit test to assert that `Dispatcher` correctly consumes `WorkerPool`
 - [ ] Add unit tests for `WorkerPool`

anandthakker:
Updated top of ticket with rationale and the checklist from the new PR template

anandthakker:
Rebased, with previous head = 94dfd96caf347cc898cb4f2e74e2ef4e45b8795e

anandthakker:
Wouldn't expect these changes to affect the benchmarks much, but why not:

## This branch

**buffer:** 886 ms
**fps:** 60 fps
**frame-duration:** 6.9 ms, 0% > 16ms
**query-point:** 0.64 ms
**query-box:** 47.80 ms
**geojson-setdata-small:** 13 ms
**geojson-setdata-large:** 115 ms

## Master

Broke w/ console errors at the end of query-point, but here are the first
three:

**buffer:** 868 ms
**fps:** 58 fps
**frame-duration:** 6.8 ms, 0% > 16ms

anandthakker:
Added a benchmark to measure the time from `new Map` to the `load` event, with multiple maps on a page.

## master

**load-multiple-maps:** 7,024 ms, loaded 6 maps.

## This branch

**load-multiple-maps:** 4,778 ms, loaded 6 maps.

anandthakker:
In light of a global worker pool, I don't know that either of the following seem necessary/useful: (a) having different map instances use a different number of workers, or (b) being able to resize the worker pool.

Does it make sense to remove the `workerCount` option from `Map` (and `Style` and `Dispatcher`)?  Instead there could be an "advanced" API like `mapboxgl.workerPool.setWorkerCount()`, which would need to be called before any map instances were constructed.

mourner:
Makes sense. I would go for something like `mapboxgl.WORKER_COUNT` — it would only be changed for development/debugging purposes anyway.

anandthakker:
> I would go for something like mapboxgl.WORKER_COUNT — it would only be changed for development/debugging purposes anyway.

I think it could be set by a user who was tying to balance a GL JS map versus other cpu-intensive stuff in a web app.  (For instance: in a project we're working on, I was thinking of intentionally limiting the GL JS worker count to 3/4 of the available workers, so that 1/4 could be dedicated to keeping some other expensive data-munging off the main thread to keep the UI snappier)

anandthakker:
^ but, with that said, I don't actually have any objections to making it a constant like you proposed, @mourner

anandthakker:
Actually, just a note that making this a constant on the `mapboxgl` object causes some weirdness, because worker_pool then has to depend on the parent mapboxgl module.  Means having to be careful about a dependency cycle, and I think might also make unit tests a bit more awkward.  How about `mapboxgl.workerPool.WORKER_COUNT`?   @mourner

anandthakker:
Went with `mapboxgl.WorkerPool.WORKER_COUNT` - @mourner let me know what you think.

(note: also rebased to master)

anandthakker:
Updated the top of the ticket.  @lucaswoj while I'm very much in favor of https://github.com/mapbox/mapbox-gl-js/issues/3034, I would propose that the changes in this PR  provide some immediate benefits and incur little to no cost.  What do you think?

bench/benchmarks/map_load.js
```diff
@@ -0,0 +1,40 @@
...
+            done();
+        }
+    }
+
+    function done () {
+    }
+
+    return evented;
+};
+
```
lucaswoj: As written, this benchmark is highly dependent on the tester's network speed. Can you think of a way to remove that dependency? What are we testing in this benchmark that isn't captured by the `buffer` benchmark?

debug/shared_workers.html
```diff
@@ -0,0 +1,122 @@
...
+        (new mapboxgl.Popup())
+            .setLngLat(map.unproject(e.point))
+            .setHTML("<h1>Hello World!</h1>")
+            .addTo(map);
+    });
+}
+</script>
+
+</body>
+</html>
```
lucaswoj: Can we adapt the `mutliple.html` test page to this use case?

js/style/style.js
```diff
@@ -18,12 +18,13 @@ var QueryFeatures = require('../source/query_features');
 var SourceCache = require('../source/source_cache');
 var styleSpec = require('./style_spec');
 var StyleFunction = require('./style_function');
+var getWorkerPool = require('../global_worker_pool');
```
lucaswoj: Can we replace the global worker pool module with a `var workerPool = new WorkerPool()` in file scope here?

bench/benchmarks/map_load.js
anandthakker: Good point.  My intent in this benchmark was to measure the impact of sending all the worker code to each of the worker threads on the map load time.  We can eliminate any outside network use

debug/shared_workers.html
anandthakker: 👍

js/source/worker.js
```diff
@@ -130,13 +130,28 @@ util.extend(Worker.prototype, {
...
+
+    getWorkerSource: function(mapId, type) {
+        if (!this.workerSources[mapId])
+            this.workerSources[mapId] = {};
+        if (!this.workerSources[mapId][type]) {
+            // simple accessor object for passing to WorkerSources
+            var styleLayers = {
+                getLayers: function () { return this.layers[mapId]; }.bind(this),
+                getLayerFamilies: function () { return this.layerFamilies[mapId]; }.bind(this)
+            };
```
lucaswoj: Why do you use an accessor object instead of a simpler `{layers: [], layerFamilies: []}` object?

js/style/style.js
anandthakker: We can, but it makes [this code](https://github.com/mapbox/mapbox-gl-js/pull/2952/files#diff-bd78950980d7affe72dbe3366d8f03f2R9) problematic: it means that there's no chance for client code to set `mapboxgl.WorkerSource.WORKER_COUNT` before the global pool is constructed.  We could work around this by having the worker pool wait to look up `WORKER_COUNT` until it's actually creating workers during the first `acquire` call.

js/source/worker.js
anandthakker: Ah - I think this is a holdover: before this change, `this.layers` and `this.layerFamilies` were not container objects but rather direct references to the layers and layerFamiliers, respectively, and so the accessor functions were necessary because the reference `this.layers` would change when 'update layers' was called.  You're right: that's no longer needed.

js/source/worker.js
```diff
@@ -15,29 +15,28 @@ function Worker(self) {
...
+        this.workerSourceTypes[name] = WorkerSource;
     }.bind(this);
 }
 
 util.extend(Worker.prototype, {
-    'set layers': function(layers) {
-        this.layers = {};
-        var that = this;
+    'set layers': function(mapId, layers) {
+        var styleLayers = this.layers[mapId] = {};
```
lucaswoj: `layers` -> `layerIds`
`styleLayers` -> `layers`

js/source/worker.js
```diff
@@ -60,20 +59,21 @@ util.extend(Worker.prototype, {
...
+        this.layerFamilies[mapId] = createLayerFamilies(this.layers[mapId]);
     },
 
-    'update layers': function(layers) {
-        var that = this;
+    'update layers': function(mapId, layers) {
         var id;
         var layer;
 
+        var styleLayers = this.layers[mapId];
```
lucaswoj: `layers` -> `layerIds`
`styleLayers` -> `layers`

js/source/worker.js
```diff
@@ -130,13 +130,28 @@ util.extend(Worker.prototype, {
...
+    'load worker source': function(map, params, callback) {
         try {
             this.self.importScripts(params.url);
             callback();
         } catch (e) {
             callback(e);
         }
+    },
+
+    getWorkerSource: function(mapId, type) {
```
lucaswoj: Why are source keyed by `type`? What happens if there are two `vector` sources on the same map?

js/mapbox-gl.js
```diff
@@ -13,6 +13,8 @@ mapboxgl.Attribution = require('./ui/control/attribution');
 mapboxgl.Popup = require('./ui/popup');
 mapboxgl.Marker = require('./ui/marker');
 
+mapboxgl.WorkerPool = require('./util/worker_pool');
```
lucaswoj: I'm 👎  on adding this to our public interface, especially if only to expose a constant.

js/style/style.js
```diff
@@ -731,7 +732,7 @@ Style.prototype = util.inherit(Evented, {
 
     // Callbacks from web workers
 
-    'get sprite json': function(params, callback) {
+    'get sprite json': function(_, params, callback) {
```
lucaswoj: Why did you choose to pass `mapId` to the `style` callbacks? It doesn't seem to be used anywhere.

js/util/worker_pool.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+var assert = require('assert');
+var WebWorker = require('./web_worker');
+var browser = require('./browser');
+
+module.exports = WorkerPool;
+
+WorkerPool.WORKER_COUNT = Math.max(browser.hardwareConcurrency - 1, 1);
```
lucaswoj: `WorkerPool.WORKER_COUNT` -> `WorkerPool.workerCount`

I'd like to start removing `MACRO_CASE` identifiers from the codebase. Javascript doesn't have any macros. The property's scope is already implied by its context. 

> Underscores with upper case, as in `UPPER_CASE`, are commonly used for C preprocessor macros, hence known as `MACRO_CASE`, and for environment variables in Unix, such as `BASH_VERSION` in bash.

test/js/util/worker_pool.test.js
```diff
@@ -0,0 +1,58 @@
...
+        t.comment('terminates workers if no dispatchers are active');
+        pool.release('map-1');
+        t.equal(terminated, WorkerPool.WORKER_COUNT);
+        t.notOk(pool.workers);
+
+        t.end();
+    });
+
+    t.end();
+});
```
lucaswoj: Thanks for the tests ❤️

js/source/worker.js
anandthakker: The `layers` parameter here isn't just ids.  `layers` => `layerDefinitions`, `styleLayers` => `layers`?

lucaswoj:
> Went with `mapboxgl.WorkerPool.WORKER_COUNT` - @mourner let me know what you think.

Let's go with `mapboxgl.workerCount` to match `mapboxgl.accessToken`. We can avoid cyclic dependencies by lazily calling `require('../../mapbox-gl').workerCount` when the value is needed.

js/source/worker.js
anandthakker: They are keyed by (mapId, type) pairs.  They're keyed by `mapId` so that WorkerSource code wouldn't need to change (i.e., each WorkerSource instance is still tied to a specific map -- see bullet 3 in "Specific Notes").  They're keyed by type because each `WorkerSource` handles requests from multiple individual sources (note that this has been true since before the custom source refactor)

js/style/style.js
anandthakker: For these targets on the main thread, this is a dummy parameter.  The only reason it exists here is so that `Actor` can be symmetric -- that is, Actor can consistently call the target method on its `parent` with (parentId, data, callback), even though parentId is irrelevant if it's receiving message from the worker for the main thread

js/source/worker.js
lucaswoj: Ah! I misunderstood. `layers` => `layerDefinitions`, `styleLayers` => `layers` looks good.

js/source/worker.js
lucaswoj: > They're keyed by type because each `WorkerSource` handles requests from multiple individual sources (note that this has been true since before the custom source refactor)

Dang. I didn't know this. This is 👍  as written.

js/mapbox-gl.js
lucaswoj: See https://github.com/mapbox/mapbox-gl-js/pull/2952#issuecomment-241135731

js/style/style.js
lucaswoj: > We could work around this by having the worker pool wait to look up `WORKER_COUNT` until it's actually creating workers during the first `acquire` call.

That sounds like a good compromise to me. Bonus side effect: no workers are created if no maps are created.

js/style/style.js
lucaswoj: I see. I'm ok with this. Let's rename `_` => `mapId` to reduce the mystery for future readers.

bench/benchmarks/map_load.js
anandthakker: Updated the benchmark to use an empty style, so it should now not depend at all on network speed

js/style/style.js
anandthakker: 👍 sorry about the mystery :)

js/style/style.js
anandthakker: @lucaswoj I changed to the lazy-lookup approach to allow for using `mapboxgl.workerCount`.  

However, it occurs to me that there's a reason to keep the singleton exposed as a separate module: once this PR goes in, we could adjust https://github.com/mapbox/mapbox-gl-js/pull/2982 so that WorkerSources are registered directly in `Source#addType`.  The most straightforward way to do that, IMO, would be to allow `source.js` to pull in a reference to the global WorkerPool instance.

^ If the value of doing that isn't clear from that description, I could also just make this a file-scoped variable as you're suggesting now, and re-extract it as part of that PR.

anandthakker:
@lucaswoj addressed all your notes except the one about scoping the worker pool to the style.js file, where there may be a reason to keep it as-is: https://github.com/mapbox/mapbox-gl-js/pull/2952#discussion_r75588490

anandthakker:
Heads up -- rebased to (hopefully) make coveralls happy again

anandthakker:
Ugh, I think the reason for the coveralls failure is that, for some reason, `mapbox-gl.js` (and some of the un-tested modules it requires, like control/geolocate.js, control/navigation.js, etc.) are _not_ included in the list of source files in the [`master` build](https://coveralls.io/github/mapbox/mapbox-gl-js), but _are_ included in the list for [this branch](https://coveralls.io/builds/7550387).

This is strange, because `mapbox-gl.js` does have a test in master, so it should be getting included.  My guess is that the difference between the two branches has something to do with using `proxyquire` in this branch to mock `mapboxgl.workerCount` -- but, again, mapbox-gl *should* have already been in the list of files being checked for coverage.

anandthakker:
> This is strange, because mapbox-gl.js does have a test in master, so it should be getting included. My guess is that the difference between the two branches has something to do with using proxyquire in this branch to mock mapboxgl.workerCount -- but, again, mapbox-gl should have already been in the list of files being checked for coverage.

Ah - it's a simple problem.  The `npm test` script runs tap on `test/js/*/*.js`, but the mapbox-gl test is at `test/js/mapbox-gl.js`.   See https://github.com/mapbox/mapbox-gl-js/pull/3043

js/style/style.js
lucaswoj: Makes sense 👌 .

lucaswoj:
This is ready to merge on 🍏 . Just rebased to include #3043.

anandthakker:
Whoops, will also need #3052 to get a :green: here.  Once that's merged I'll rebase again and then (hopefully) merge

anandthakker:
Rebasing -- previous at https://github.com/mapbox/mapbox-gl-js/pull/2952/commits/84ce97bdd44c6bc72508f482f28668f05a30c0b8

anandthakker:
Success! Merging.

anandthakker:
Ack - just realized that I dropped the ball on [this comment](https://github.com/mapbox/mapbox-gl-js/pull/2952#discussion_r75548284) -- the diff changed while I was addressing something else, and so I forgot about it.

@lucaswoj if I don't get a chance to open a PR patching that today, I'll file an issue so it doesn't get lost

lucaswoj:
@anandthakker It's a minor change, don't stress over it.

PR #2958, opened 2016-08-09
Enable ci.sh to run on osx

springmeyer:
This prepares `ci.sh` to run on OS X. Per chat with @lucaswoj we are going to try using OS X machines on circleci. If this works this could provide a fallback if headless-gl stability on linux continues to be a problem.  This can be tested by:

 - merging this PR
 - going to https://circleci.com/gh/mapbox/mapbox-gl-js/edit#build-environment and flipping the build environment to OS X

@lucaswoj - let's sync up tomorrow morning before enabling.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -22,8 +22,10 @@ tap --reporter dot --coverage --no-coverage-report test/js/*/*.js test/build/web
 # allow writing core files for render tests
 ulimit -c unlimited -S
 echo 'ulimit -c: '`ulimit -c`
-echo '/proc/sys/kernel/core_pattern: '`cat /proc/sys/kernel/core_pattern`
-sysctl kernel.core_pattern
+if [[ $(uname -s) == 'Linux' ]]; then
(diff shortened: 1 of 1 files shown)


lucaswoj:
This looks great! I'll put something on your calendar for tomorrow morning. 

Is the idea that we'll remove the `if [[ $(uname -s) == 'Linux' ]]; then` blocks once we switch CircleCI over to OSX? Or should we keep those for debugging?

mourner:
Let's try it out!

mourner:
Hmm, it failed starting with the following error in the "configure build" section:

```
No shared schemes were found in your repo. Please ensure you have shared your
  scheme. See the iOS documention for more information on sharing schemes.
  Alternatively, you can explicitly set a scheme with the 'XCODE_SCHEME'
  environment variable.

Action failed: Configure the build
```

Will switch it back for the time being, let's investigate.

springmeyer:
> Hmm, it failed starting with the following error in the "configure build" section:

Fix for that now over at #2965

PR #2961, opened 2016-08-09
Ensure _rerender is called if _styleDirty is true

### LINKED ISSUE #2936: Removing a layer causes map.loaded() to return false until map is moved.
**mapbox-gl-js version**:0.21.0

**jsBin**: http://jsbin.com/fewegikejo/edit?html,console,output

### Steps to Trigger Behavior

 1. Check that map.loaded() returns true
 2. Remove any layer from the map
 3. Check that map.loaded() returns false a few times
 4. Pan or zoom the map
 5. See that map.loaded() now returns true

anandthakker:
Closes #2936

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -1035,7 +1035,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
                 this._styleDirty = true;
             }
 
-            if (this._sourcesDirty || this._repaint || !this.animationLoop.stopped()) {
+            if (this._sourcesDirty || this._repaint || this._styleDirty) {
                 this._rerender();
             }
 
--- a/test/js/ui/map.test.js
+++ b/test/js/ui/map.test.js
@@ -13,23 +13,23 @@ var fixed = require('../../testutil/fixed');
 var fixedNum = fixed.Num;
 var fixedLngLat = fixed.LngLat;
 
-test('Map', function(t) {
-    function createMap(options) {
-        return new Map(extend({
-            container: {
-                offsetWidth: 200,
-                offsetHeight: 200,
-                classList: {
-                    add: function() {},
-                    remove: function() {}
-                }
-            },
-            interactive: false,
-            attributionControl: false,
-            trackResize: true
-        }, options));
-    }
+function createMap(options) {
+    return new Map(extend({
+        container: {
+            offsetWidth: 200,
+            offsetHeight: 200,
+            classList: {
+                add: function() {},
+                remove: function() {}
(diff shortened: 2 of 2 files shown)


lucaswoj:
I'm having trouble understanding the `_styleDirty` flag well enough to 👍 / 👎 on this diff. @mourner do you have any insight?

mourner:
@anandthakker is the `styleDirty` flag true while there are no style changes to apply but stuff is transitioning/fading?

lucaswoj:
As a sanity check to make sure this isn't making the render loop do a bunch of extra work, I ran some benchmarks on this branch & master. The numbers look 👌 

# Branch Benchmarks

**buffer:** 897 ms
**fps:** 60 fps
**frame-duration:** 6.8 ms, 0% > 16ms
**query-point:** 0.75 ms
**query-box:** 44.97 ms
**geojson-setdata-large:** 103 ms

# Master Benchmarks

**buffer:** 868 ms
**fps:** 60 fps
**frame-duration:** 6.8 ms, 0% > 16ms
**query-point:** 0.77 ms
**query-box:** 47.55 ms
**geojson-setdata-large:** 100 ms

anandthakker:
@mourner I don't understand the transition stuff well enough to know the precise answer to your question.  But here is my detailed understanding of this bug: [this line in _render](https://github.com/mapbox/mapbox-gl-js/blob/1829357f5c1361b83a0077e0dd1971889d18fef3/js/ui/map.js#L1006) clears the flag, and the only places it gets set are [here, further down in the same method](https://github.com/mapbox/mapbox-gl-js/blob/1829357f5c1361b83a0077e0dd1971889d18fef3/js/ui/map.js#L1035)  and [here, in Map#_update](https://github.com/mapbox/mapbox-gl-js/blob/1829357f5c1361b83a0077e0dd1971889d18fef3/js/ui/map.js#L989).

My understanding is that the reason this bug was happening is that the `style.update` call [here](https://github.com/mapbox/mapbox-gl-js/blob/1829357f5c1361b83a0077e0dd1971889d18fef3/js/ui/map.js#L1007) was resulting in a `style.change` event, which in turn resulted in `map._update()` being called.  However, since this is happening synchronously, we are still in the middle of a _render call, so the `_rerender` bails out without scheduling another render; thus, _styleDirty doesn't get cleared until something else happens to trigger a _rerender.

lucaswoj:
As I see it, this PR fixes a bug, shows no negative performance impact, and is sensical as best any of us can understand. :shipit:

PR #2963, opened 2016-08-09
Fix tile reloading race condition

### LINKED ISSUE #2559: Fix DataCloneError in Firefox and IE11
On firefox we get:

> DataCloneError: The object could not be cloned.

This error is encountered reliably on Firefox 43-46, hitting 217 users and 1.1k times so far. Occurs within the Actor object.

mourner:
Attempt to fix #2559. After this change, I can’t seem to reproduce the original bug. 

Will try writing a test case for this but not sure if it’s easy to do.

cc @lucaswoj @anandthakker

### DIFF
--- a/test/js/source/vector_tile_source.test.js
+++ b/test/js/source/vector_tile_source.test.js
@@ -112,6 +112,34 @@ test('VectorTileSource', function(t) {
     testScheme('xyz', 'http://example.com/10/5/5.png');
     testScheme('tms', 'http://example.com/10/5/1018.png');
 
+    t.test('reloads a loading tile properly', function (t) {
+        var source = createSource({
+            tiles: ["http://example.com/{z}/{x}/{y}.png"]
+        });
+        var events = [];
+        source.dispatcher.send = function(type, params, cb) {
+            events.push(type);
+            setTimeout(cb, 0);
+            return 1;
+        };
+
+        source.on('load', function () {
+            var tile = {
+                coord: new TileCoord(10, 5, 5, 0),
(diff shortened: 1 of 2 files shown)


anandthakker:
👍 Fix LGTM

> Will try writing a test case for this but not sure if it’s easy to do.

Shouldn't be too hard by mocking the dispatcher?  (like in [GeoJSONSource test](https://github.com/mapbox/mapbox-gl-js/blob/master/test/js/source/geojson_source.test.js))

mourner:
@lucaswoj 👀, now a bit simpler and with a test.

test/js/source/vector_tile_source.test.js
```diff
@@ -112,6 +112,34 @@ test('VectorTileSource', function(t) {
...
+        source.on('load', function () {
+            var tile = {
+                coord: new TileCoord(10, 5, 5, 0),
+                state: 'loading',
+                loadVectorData: function () {
+                    this.state = 'loaded';
+                    events.push('tile loaded');
+                }
+            };
+            source.loadTile(tile, function () {});
```
anandthakker: Just to be extra explicit, how about asserting `t.same(tile.state, 'loading')` after the first `loadTile` call?  The `setTimeout` in dispatcher.send ensures this, but I think including the assertion would make the test code a better expression of what we're trying to cover here

mourner: Good point, added.

lucaswoj:
This looks great! Let's 🚢 when CI passes.

PR #2965, opened 2016-08-09
[circleci] adapt circle.yml to osx environment

springmeyer:
This PR modifies the circle.yml to be able to be run on the circle OS X environment. It builds off of #2958.

The overall motivation is to be able to run all tests without segfaults (that are common on linux due to uncontrolled factors: bugs in headless-gl or mesa?).

Key changes:

  - Installs node with nvm because circleci OS X machines do not support:

```
 machine:
 -  node:
 -    version: '4'
```

  - Provides new environment options: `XCODE_SCHEME` and `XCODE_WORKSPACE` to trigger no-op in circleci configuration setup.

Background: The circleci OS X support assumes you are developing iOS applications with an xcodeproject (https://circleci.com/docs/ios-builds-on-os-x/). So this project will fail by default when an xcode project is not found. This PR fixes this problem by telling circleci to not look for xcode files.

More notes:

   - Speed is slightly slower than the previous build that passed on master
     - Build was: https://circleci.com/gh/mapbox/mapbox-gl-js/4346
     - Total time: 13:19 master/linux, 15:14 PR/osx
     - Install deps from cache: 1:08 master/linux, 0:38 PR/osx
     - Save cache: 0:02 master/linux, 1:06 PR/osx
     - Run build/tests: 11:04 master/linux, 12:23 PR/osx

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 
-set -e
+source ./nvm/nvm.sh
+nvm use ${NODE_VERSION}
+
+set -eu
 set -o pipefail
 
 # add npm packages to $PATH
--- a/ci.sh
+++ b/ci.sh
@@ -9,11 +9,6 @@ set -o pipefail
 # add npm packages to $PATH
 export PATH=$(pwd)/node_modules/.bin:$PATH
 
-# disable spotlight to ensure we waste no CPU on needless file indexing
-if [[ $(uname -s) == 'Darwin' ]]; then
-    sudo mdutil -i off /
-fi
-
 # set up code coverage instrumentation
 rm -rf coverage .nyc_output
 
--- a/ci.sh
+++ b/ci.sh
@@ -9,6 +9,11 @@ set -o pipefail
 # add npm packages to $PATH
 export PATH=$(pwd)/node_modules/.bin:$PATH
 
+# disable spotlight to ensure we waste no CPU on needless file indexing
+if [[ $(uname -s) == 'Darwin' ]]; then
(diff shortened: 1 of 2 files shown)


lucaswoj:
Looks good to me. 🚢

ci.sh
```diff
@@ -1,8 +1,36 @@
...
+}
+
+
+install_node 4
+
+if [[ "$(git symbolic-ref --short -q HEAD)" == "master" ]]; then
+    npm update
+else
+    npm install
+fi
```
lucaswoj: Are there any consequences of moving this from the `dependencies` step to the `test` step?  Or is that just presentational?

lucaswoj: My reading of the CircleCI docs is that these sections are entirely presentational https://circleci.com/docs/configuration/

springmeyer: Good catch: there is a consequence: caching. It looks like the cache is restored before the `Dependencies` section and saved after and before the `test` section. So to benefit from the `node_modules` cache we'd need things to be installed in the `Dependencies`. Hrm...

lucaswoj: I’m open to using the pre-installed node v6.

That simplifies `ci.sh` and enables dependency caching.

If upstream image changes become a problem in the future we can look into solutions then.

lucaswoj:
This looks good. :ship: on green.

lucaswoj:
# Next Steps

 - [ ] merge this PR
 - [ ] un-ignore as many test-suite cases as possible
 - [ ] fix any PRs that began failing during the CI debugging process

PR #2973, opened 2016-08-10
Update the geolocate icon

tristen:
Before 

screen shot 2016-08-10 at 10 46 58 am

After

screen shot 2016-08-10 at 10 42 31 am

cc @lucaswoj


indus:
UTF8 (as the old one) should be smaller: https://css-tricks.com/probably-dont-base64-svg/

Issue #2647, opened 2016-06-01
Fix bug preventing "queryRenderedFeatures" from returning all features


cammanderson:
Hi Guys, 

Love your work. 

I think that we may have a bug to contribute. Something that is unexpected and unpredictable behaviour around the queryRenderedFeatures API call on map. We observe occasions where the map markers are returned, and then other occasions that they are not - despite them still being rendered and no changes to layers/sources - just moving/zooming the map.

Please see the following code pen showing the issue;
http://codepen.io/cammanderson/pen/bc7ceb108e340cb30fa524ff982e761d?editors=1010

**mapbox-gl-js version**: 0.19.0

### Steps to Trigger Behavior

 1. Open your browser console
 2. Position the map viewport over Barcelona map marker (so it is only visible)
 3. Zoom in and out over this marker while the marker is still in view
 4. Watch the features that are console.logged/returned by the calling queryRenderedFeatures method

### Expected Behavior

The console should always show the map marker for Barcelona while you zoom in and zoom out. queryRenderedFeatures should always return the markers that are rendered on screen.

### Actual Behavior

There are occasions where the marker will not be returned while zooming in and out. It is like at some stages the marker is no longer present, even though it is rendered on screen at that time.

This can sometimes happen on dragging the map around, like it shows up in the top-right quadrant of the screen, but then disappears from being returned when on the bottom right. This seems to change as you move around the map. 

Let me know if my example has errors, if so, it's been a long day sorry. But I observed this on our application, then was able to reproduce this using codepen (just trying to make sure it wasn't just our code implementation in our app).

Thanks guys, 
Cheers
Cam

lucaswoj:
I can confirm that `queryRenderedFeatures` is broken in the way you describe in this example.

x9xjdzz9eaeeghcrar9c:
I'm also affected by this bug. I took the "Highlight features within a bounding box" example, and changed it to show the bug. I find that it only happens with symbol layers and only when you change from the initial zoom level. So, with the attached program, hold down shift to drag a rectangle and you will see that only some of the points are selected (they are colored red when selected). But it only happens after you have zoomed in or out.

[map.zip](https://github.com/mapbox/mapbox-gl-js/files/418718/map.zip)

jmcazaux:
We are affected as well here.

Is there a workaround we can use until this is solved ?

stevelacy:
[...]

PR #2979, opened 2016-08-10
queryRenderedFeatures: default params to empty object instead of throwing error

### LINKED ISSUE #2969: Fix bug preventing "queryRenderedFeatures" from being called with no arguments
The GL JS API documentation correctly lists this method's signature as 

```js
queryRenderedFeatures([region], [parameters])
```

... however calling `map.queryRenderedFeatures()` produces an error unless a parameters object is passed `map.queryRenderedFeatures({})`. 

This bug was causing some of our benchmark scripts to crash and was fixed in https://github.com/mapbox/mapbox-gl-js/pull/2950 ([exact spot in the diff](https://github.com/mapbox/mapbox-gl-js/pull/2950/files#diff-63abd99e8332ce21c1323063948b72f0R39))

Relevant source [here](http://github.com/mapbox/mapbox-gl-js/blob/f287983ecf88ee4df6ad0ee16144de02d93477ae/js/ui/map.js#L516-L517)

mapsam:
#### What changed

Added a default empty object `{}` to `map.queryRenderedFeatures()` if nothing is passed through. Currently gl-js throws an error if the query is missing a second parameter. In order to pass linting tests, I had to rename `params` to `parameters` within the logic of `queryRenderedFeatures`.

fixes #2969

#### Tests

Added two new tests that confirm the object is generated if it isn't passed through, and a test that confirms all parameters are preserved in the output (happy to remove this if it's unnecessary).

#### Refs

#2647 

cc @lucaswoj @bhousel @mollymerp

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -644,7 +644,7 @@ Style.prototype = util.inherit(Evented, {
         }
 
         var includedSources = {};
-        if (params.layers) {
+        if (params && params.layers) {
             for (var i = 0; i < params.layers.length; i++) {
                 var layerId = params.layers[i];
                 includedSources[this._layers[layerId].source] = true;
--- a/test/js/ui/map.test.js
+++ b/test/js/ui/map.test.js
@@ -593,6 +593,26 @@ test('Map', function(t) {
                 map.queryRenderedFeatures(map.project(new LngLat(360, 0)), opts);
             });
 
+            t.test('if no options, defaults to empty object', function(t) {
+                map.style.queryRenderedFeatures = function (coords, o) {
+                    t.ok(coords);
+                    t.deepEqual(o, {}, 'options are set to a default object');
+                    t.end();
+                };
+
+                map.queryRenderedFeatures(map.project(new LngLat(0, 0)));
+            });
+
+            t.test('options carry through query', function(t) {
+                map.style.queryRenderedFeatures = function (coords, o) {
+                    t.ok(coords);
+                    t.deepEqual(o, {foo: 'bar'}, 'options are expected');
+                    t.end();
+                };
+
+                map.queryRenderedFeatures(map.project(new LngLat(0, 0)), {foo: 'bar'});
+            });
+
             t.end();
         });
     });
--- a/test/js/ui/map.test.js
+++ b/test/js/ui/map.test.js
@@ -564,6 +564,32 @@ test('Map', function(t) {
         map.on('style.load', function() {
             var opts = {};
 
+            t.test('if no parameters provided, use defaults', function(t) {
+                var results = map.queryRenderedFeatures();
+                t.deepEqual(results, [], 'expected results');
+                t.end();
+            });
+
+            t.test('if no options, defaults to empty object', function(t) {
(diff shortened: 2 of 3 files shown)


test/js/ui/map.test.js
```diff
@@ -593,6 +593,26 @@ test('Map', function(t) {
...
+                map.style.queryRenderedFeatures = function (coords, o) {
+                    t.ok(coords);
+                    t.deepEqual(o, {foo: 'bar'}, 'options are expected');
+                    t.end();
+                };
+
+                map.queryRenderedFeatures(map.project(new LngLat(0, 0)), {foo: 'bar'});
+            });
+
             t.end();
```
lucaswoj: The shared `map` across these test cases make me a little nervous but I don't see any immediate harm coming of it.

test/js/ui/map.test.js
```diff
@@ -593,6 +593,26 @@ test('Map', function(t) {
...
+
+            t.test('options carry through query', function(t) {
+                map.style.queryRenderedFeatures = function (coords, o) {
+                    t.ok(coords);
+                    t.deepEqual(o, {foo: 'bar'}, 'options are expected');
+                    t.end();
+                };
+
+                map.queryRenderedFeatures(map.project(new LngLat(0, 0)), {foo: 'bar'});
+            });
```
lucaswoj: In the original issue I was getting exceptions from calling `map.queryRenderedFeatures()` (no arguments). Can we add another test for the "no arguments" case?

test/js/ui/map.test.js
mapsam: @lucaswoj I grabbed this from the [two](https://github.com/mapbox/mapbox-gl-js/pull/2979/files/76381412754bc117f0c6d35f030c50c49694e0cb#diff-277bb5e9f409aee234cc524db362a55bR568) tests [above](https://github.com/mapbox/mapbox-gl-js/pull/2979/files/76381412754bc117f0c6d35f030c50c49694e0cb#diff-277bb5e9f409aee234cc524db362a55bR580) without thinking much about it. Happy to find an alternative!

lucaswoj: Totally understood! I appreciate you sticking to convention. 😄 Let's leave these tests-as is for now.

test/js/ui/map.test.js
mapsam: Oh! Totally.

mapsam: @lucaswoj added an extra default for `pointOrBox` as well (defaults to `undefined`, matching the syntax below).

js/style/style.js
```diff
@@ -644,7 +644,7 @@ Style.prototype = util.inherit(Evented, {
         }
 
         var includedSources = {};
-        if (params.layers) {
+        if (params && params.layers) {
```
mapsam: @lucaswoj In the test [here](https://github.com/mapbox/mapbox-gl-js/pull/2979/files#diff-277bb5e9f409aee234cc524db362a55bR567) I hit the error `Error: TypeError: Cannot read property 'layers' of undefined`, which led to a check to ensure `params` exists in the first place.

PR #2982, opened 2016-08-11
Revise addSourceType to be independent of Map instance

### LINKED ISSUE #3004: Accept custom source types before or during Map creation
Currently, the only way to add a custom source type in client code is `Ma#addSourceType`.  This ends up being awkward, because (together with the issue raised in #3003) it means that you cannot construct a map instance with a stylesheet JSON, but rather have to use `Map#addSource` and `Map#addLayer`. (If #3003 is fixed, then `Map#setStyle` would also be an option, but this is still less ideal than `new Map({ style: stylesheet, ... })`.

Proposed fixes (I am proposing both, although either one alone would suffice):
1. Accept a `sourceTypePlugins: { /* maps source type name to the custom source type */ }` in the `Map` constructor, and handle it before creating the style. (This may mean moving where the Dispatcher is created, but I think it's not too bad.)
2. Move `Map#addSourceType` to something like `mapboxgl.Source.addSourceType`. <del>(This would require a deeper change, wherein when a new Worker is created, we check whether any custom sources require registering a WorkerSource.)</del>  Because adding a custom source type may require communication with the workers, this raises a complication: (i) when a new map is created, it must check for existing source types that have workerSourceURLs and register those with its workers; (ii) if a source type is added _after_ some map instances are created, how do we tell all of the instances to register the new worker source?

anandthakker:
Closes #3004

**The problem:**  `addSourceType` does not quite make sense as a method on the Map (or Style) instance, because: a) Sources, being types, are really 'global'; b) practically speaking, there are problems having this as an instance method, because that approach fails to coordinate setting up the WorkerSources for other map instances.

**Changes:**
 - Replace `Map#addSourceType` and `Style#addSourceType` with the static `Source.addType` -- exported as `mapboxgl.addSourceType`.  Leveraging the global worker pool, this method can now directly take care of registering the WorkerSource for the added source type if necessary.

## Checklist

 - [x] [document any changes or additions to public APIs](https://github.com/mapbox/mapbox-gl-js/blob/master/docs/README.md) **(skipping this because we are keeping the API private until it's vetted)**
 - [ ] [post scores for all benchmarks on your branch and `master`](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/README.md#running-benchmarks)


anandthakker:
I'm going to revise this a bit so that it can address #3004

anandthakker:
@lucaswoj ^ revised this to be a more comprehensive solution.  Ready for your 👀

anandthakker:
(also edited the top of the ticket)

anandthakker:
Here is a working example of the API that this change affords: https://github.com/developmentseed/mapbox-gl-topojson/blob/gh-pages/site.js

js/ui/map.js
```diff
@@ -202,6 +203,11 @@ var Map = module.exports = function(options) {
     this.resize();
 
     if (options.classes) this.setClasses(options.classes);
+    if (options.sourceTypes) {
```
mourner: Should we name it `customSourceTypes` to reflect that we supply those additionally to the core ones?

mourner:
This generally looks good to me!

js/ui/map.js
anandthakker: Makes sense to me, but I also don't have a strong opinion on this.

anandthakker:
@mourner @lucaswoj added one more small change in the last commit.  I believe this is ready to merge.

js/mapbox-gl.js
```diff
@@ -14,6 +14,7 @@ mapboxgl.Popup = require('./ui/popup');
 mapboxgl.Marker = require('./ui/marker');
 
 mapboxgl.Style = require('./style/style');
+mapboxgl.Source = require('./source/source');
```
lucaswoj: Can we expose `mapboxgl.addSourceType` instead of the whole `Source` module? There are lots of bits & pieces in `Source` that we don't want to be public.

js/ui/map.js
```diff
@@ -202,6 +203,11 @@ var Map = module.exports = function(options) {
     this.resize();
 
     if (options.classes) this.setClasses(options.classes);
+    if (options.customSourceTypes) {
+        for (var sourceTypeName in options.customSourceTypes) {
+            Source.addType(sourceTypeName, options.customSourceTypes[sourceTypeName]);
+        }
+    }
```
lucaswoj: Sources can be added either by `mapboxgl.addSourceType` or the `customSourceTypes` constructor option. Can we settle on a single API?

js/source/source.js
```diff
@@ -30,23 +35,53 @@ exports.create = function(id, source, dispatcher) {
...
+Source.addType = function (name, SourceType) {
+    if (Source.getType(name)) {
+        throw new Error('A source type named ' + name + ' already exists.');
+    }
+
+    Source.setType(name, SourceType);
+
+    // an internal event, used to notify Style instances that there is a new
+    // custom source type.
+    Source.fire('_add', { name: name });
```
lucaswoj: This could be simplified a little bit now that we have a global worker pool. It could be simplified even more once we implement #3034 😄

js/ui/map.js
anandthakker: 👍 agreed - I removed the `customSourceTypes` option, as I think `mapboxgl.addSourceType` makes more sense.

anandthakker:
Rebased, and updated to leverage the global worker pool.

Two key notes:

1. Note the fix in https://github.com/mapbox/mapbox-gl-js/pull/2982/commits/630cfe2a2aca164be6a2848cee2b0be8df1cc140 -- this isn't really the topic of this PR (it's actually a followup to #2952 ), but the issue got in my way while I was here so I figured I'd include the fix.
2. There is an outstanding TODO [here](https://github.com/mapbox/mapbox-gl-js/pull/2982/files#diff-d5506f072d1241401a7e31ff642c1606R88)

anandthakker:
@lucaswoj this should be good to go.  If you have reservations about this PR or it needs more discussion, it might be worth pulling out 8a7807f into a separate one to merge sooner

anandthakker:
Rebased

lucaswoj:
I'm coming to see the wisdom behind GL Native's custom layer architecture. Instead of defining custom source _types_, we should accept custom source _objects_ which conform to a given interface specification. 

```js
map.addCustomSource('foo', new WhizzBangSource({...}))
```

The benefits of this approach include

 - no need to worry about interactions between the style spec validator and custom sources
 - the ability to pass functions to custom sources (not allowed by the style spec)
 - smaller API surface / less boilerplate required to create a custom source
 - compatibility with the planned custom layer interface #281 
 - compatibility with GL Native's custom layer interface 

One downside is that we'll need to find a way to register worker-side logic without the existence of source "types". Does this seem doable @anandthakker? Would it be more doable if we implemented #3034? 

Overall, does implementing this architecture sound doable within the scope of this PR?

anandthakker:
@lucaswoj This is interesting!  I actually don't think it would be too difficult to deal with registering worker-side code; we'd just need an API like `mapboxgl.registerWorkerSource( url )`, where the URL would be to a script pretty similar to that which is at present provided by `CustomSourceType.workerSourceURL`

However, one drawback to this no-types approach is that it basically seals the deal against specifying custom sources (and their dependent style layers) declaratively with style json like `new Map(style: url)`.  Personally, I think it's nice that the source type approach could enable this (after appropriately relaxing style validation).

lucaswoj:
> However, one drawback to this no-types approach is that it basically seals the deal against specifying custom sources (and their dependent style layers) declaratively with style json like `new Map(style: url)`.  Personally, I think it's nice that the source type approach could enable this (after appropriately relaxing style validation).  

It's a tradeoff for sure. We are giving up some ergonomics in exchange for a simplicity and the preservation of a tight validator. I prefer to err on the side of API simplicity (and uniformity with GL Native) and see how things shake out. in practice. Down the road we may revisit the "register source type" design or a design that allows embedding `Source`-interfaced objects in calls to `Map#setStyle`.

anandthakker:
> It's a tradeoff for sure. We are giving up some ergonomics in exchange for a simplicity and the preservation of a tight validator. I prefer to err on the side of API simplicity (and uniformity with GL Native) and see how things shake out.

Gotcha.

I think the change from addSourceType-based to addCustomSource-based API should be reasonably manageable, but definitely makes more sense to do after the `PooledWorker` refactor that you're working on @lucaswoj.

lucaswoj:
We are going to go a different direction for this functionality: https://github.com/mapbox/mapbox-gl-js/issues/3004

PR #2990, opened 2016-08-11
Ensure _rerender is called if _styleDirty is true

### LINKED ISSUE #2936: Removing a layer causes map.loaded() to return false until map is moved.

anandthakker:
Closes #2936

Second attempt - either the fix or the unit test were causing problems in the original attempt (#2961)

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -1051,7 +1051,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
                 this._styleDirty = true;
             }
 
-            if (this._sourcesDirty || this._repaint || !this.animationLoop.stopped()) {
+            if (this._sourcesDirty || this._repaint || this._styleDirty) {
                 this._rerender();
             }
 
--- a/test/js/ui/map.test.js
+++ b/test/js/ui/map.test.js
@@ -1004,6 +1004,68 @@ test('Map', function(t) {
         t.end();
     });
 
+    test('render stabilizes', function (t) {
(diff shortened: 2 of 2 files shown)


anandthakker:
@lucaswoj I'm still not totally sure about this unit test -- it's working consistently locally, but the previous one was too.  Maybe we can leave this PR open and kick the build a few times to see if it appears to work consistently on Circle?

anandthakker:
Update: throughout the evening I've restarted the CircleCI build four times, and it's passed each time.  Not a huge sample, but promising.

anandthakker:
Okay, I've hit "Rebuild" 7 times, for a total of 8 successful builds.  If the unit test was [failing 50% of the time before](https://github.com/mapbox/mapbox-gl-js/issues/2936#issuecomment-239306091), then I think this is statistically significant (p < 0.005 heh)

mourner:
I've checked the PR before and I think it's correct — the dirty flag should be always set to true while the animation loop is still running. Let's merge and see if it causes any more problems. The volatility may have been caused but our switch to OS X image on Circle.

Issue #2967, opened 2016-08-09
Make benchmark runner able to run tests against both "master" and current branch 


lucaswoj:
💭 We can publish a browserified build of `master`'s benchmarks to s3

PR #2992, opened 2016-08-11
Fix deploy

springmeyer:
The `./deploy.sh` regressed in #2965 because OS X machines do not have the ability to install `pip` packages without `sudo` by default. Ideally we don't want to use sudo at all: this PR fixes.

Before merging:

  - [ ] @lucaswoj tests tagging via this branch to ensure it works
  - [ ] We think about how to prevent regression again, ideas:
    - [ ] Add "dry run" functionality to  `./deploy.sh` and run that as part of the ci.sh tests each commit
    - [ ] Always set up and test `awscli` and not only when deploying: this would cost a bit of time but easily ensure the tooling is working before deploy.


lucaswoj:
> We think about how to prevent regression again, ideas:

Implementing #2967 the way I proposed in the ticket would have the desirable side effect of testing our deployment infrastructure on every commit to `master`.

That said, there's no real consequence to testing this "live" other than a bunch of scary 🔴 on CircleCI.

PR #2996, opened 2016-08-12
Make sure we only call "mbutil" on Darwin systems

lucaswoj:
This is the only macOS-ism in our CI scripts. 

cc @springmeyer


springmeyer:
👍  🚢

PR #3001, opened 2016-08-15
Upgrade test suite for TMS rendering tests

kkaefer:
TMS support was added in #2565, but there are no tests for rendering styles that specify a TMS style scheme. https://github.com/mapbox/mapbox-gl-test-suite/pull/134 implements this and we should use it in GL JS.


lucaswoj:
Looks great! Thanks @kkaefer. :shipit:

PR #3022, opened 2016-08-17
Fix regression in querySourceFeatures

anandthakker:
This fixes a regression due to an obsolete method name ('SourceCache#renderedIDs'), and adds a test that would have caught this regression.


anandthakker:
@lucaswoj 👀  - note that `querySourceFeatures` is broken in the current release (0.22), confirmed in this example page: https://www.mapbox.com/mapbox-gl-js/example/filter-features-within-map-view/

lucaswoj:
Looks good! 🚢

PR #3024, opened 2016-08-17
Simpler debug/bench build setup

lucaswoj:
This PR builds upon #3019 by 

 - moving both `debug` and `bench` into a new `server` directory which allows us to keep the existing debug URLs and have a closer mapping between URLs and directory structure 
 - eliminating the need to run many instances of `watchify` and keep browserify builds on disk using an [`enchilada`](https://github.com/defunctzombie/node-enchilada) server (much like `browserify-middleware` but free of the undesirable precompiling behaviour)

Any ideas worth keeping in here @mourner?


mourner:
> moving both debug and bench into a new server directory which allows us to keep the existing debug URLs and have a closer mapping between URLs and directory structure

I still don't see the benefit of the existing URLs, and it seems to me that `server` directory will look a bit confusing, especially to new contributors.

> eliminating the need to run many instances of watchify

Are there any specific issues with running 2 watchify instances in parallel? It works pretty well in practice.

> using an enchilada server

Enchilada is another dependency that's pretty much unmaintained. It may be better to use https://github.com/mattdesl/watchify-middleware, which is more current/actively developed and also powers Budo, but I still like the static server setup much more.

lucaswoj:
> I still don't see the benefit of the existing URLs

The existing URLs are simpler. The existing URLs are muscle memory to me and other GL JS developers. 

> Are there any specific issues with running 2 watchify instances in parallel? It works pretty well in practice.

Running 2 watchify instances in parallel gives me pause because of the potential for orphaned background processes, potential for compatibility issues (Windows), and the required complexity of our `package.json` scripts. 

> Enchilada is another dependency that's pretty much unmaintained. It may be better to use https://github.com/mattdesl/watchify-middleware, which is more current/actively developed and also powers Budo, but I still like the static server setup much more.

 - `watchify-middleware` was last updated 7 months ago and has no outstanding issues
 - `node-enchilada` was last updated 10 months ago and has no outstanding Issues

mourner:
> The existing URLs are simpler.

Shorter doesn't necessarily mean simpler. E.g. what I like about `/dist/mapbox-gl-dev.js` as oppose to `mapbox-gl.js` is that:

- It directly refers to an existing file in the file system, rather than being a "magic" URL. 
- It points to the same build that's used in other places like `build` scripts, that gets published to CDN etc.
- It's immediately apparent from the URL that this is a development build, not production build like you would have guessed from the url (because `mapbox-gl.js` is the production build in `dist`)
- The fact that the file is actually present on the file system makes it much easier to diagnose any potential bugs in the build process (e.g. a faulty transform or a regression in browserify or babel).
- In case of `index.js` in `bench`, it's referred to in the HTML and also present on the file system, but then turns out to return a different `index.js` that new contributors would assume, which is a bit confusing.

I like the idea of using absolute URLs as opposed to having `../` though — indeed makes them both simpler and shorter.

> The existing URLs are muscle memory to me and other GL JS developers.

It's not like you have to type those URLs over and over — e.g. you would typically just copy over an existing debug page when creating a new one, rather than writing it from scratch. I also think that URLs that map to actual physical locations are easier to remember for new contributors.

> Running 2 watchify instances in parallel gives me pause because of the potential for orphaned background processes, potential for compatibility issues (Windows), and the required complexity of our package.json scripts.

Good point about Windows compatibility. #3019 avoids orphaned processes with the Unix `& wait` construct in the scripts, e.g. making sure all processes are terminated when you hit `Ctrl-C`. However, [npm-run-all](https://github.com/mysticatea/npm-run-all) looks like a great remedy — it's cross-platform, very actively developed, makes sure all processes are terminated correctly when run in parallel, and also makes the NPM scripts look MUCH simpler:

```
# before
npm run build-token && npm run download-bench-data && npm run watch-dev & npm run watch-bench & npm run start-server & wait

# after; isn't this wonderful?
run-s build-token download-bench-data && run-p watch-dev watch-bench start-server
```

Also remember that we're comparing a few more lines of NPM scripts with a 30-line express-based custom server, and the latter seems much more complex to me.

> node-enchilada was last updated 10 months ago and has no outstanding Issues

Yeah, technically it's a 3 months difference between last commits, but if you look at the [commit history](https://github.com/defunctzombie/node-enchilada/commits/master), there have been no activity apart from a few dependency updates since 1.5 years ago (and the deps are now vastly outdated anyway, e.g. browserify is v4 vs v13 latest). And additionally no responses to PRs over the last year.

But much more importantly, the issue with both `enchilada` and `browserify-middleware` is that they don't do incremental updates. E.g. if you do one tiny change in one file, it will do a full rebuild, which means you will still have to wait half a minute to run an updated benchmark for example. On the other hand, watchify-based solutions will update the builds in 0.2-0.3s, or pretty much instantly.

PR #3026, opened 2016-08-18
Get rid of async dependency

mourner:
Just a minor change to make `async` dependency unnecessary. Also removes `through` dependency because it doesn’t appear to be used anywhere anymore.

Same amount of code + less dependencies = good!

cc @lucaswoj


lucaswoj:
Thank you!

PR #3029, opened 2016-08-18
Make generated filenames consistent & explicit

lucaswoj:
As of this PR

 - debug pages reference `/debug/access-token.js` by absolute path to fix broken`http://localhost:9966/debug` page
 - `debug/access-token.js` is browserified to `debug/access-token-generated.js`
 - `bench/index.js` is browserified to `bench/index-generated.js`
 - `package.json` scripts use long-form command line flags for readability 

# Todo

 - [x] ignore `*-generated.js` files in eslint

cc @mourner


mourner:
Great! The ignore part is already done — that's what `eslint --ignore-path .gitignore` flag does, since generated files are in gitignore.

lucaswoj:
Ah! Makes sense. I think I was seeing eslint errors from the files after switching branches. Thanks for the 🚢.

PR #3042, opened 2016-08-22
Add maxWidth option

amishas157:
Though the actual width of scale is not in control of developer, but there is a possibility of keeping the maxWidth of scale as flexible.

### DIFF
--- a/js/ui/control/scale.js
+++ b/js/ui/control/scale.js
@@ -12,8 +12,10 @@ module.exports = Scale;
  * @class Scale
  * @param {Object} [options]
  * @param {string} [options.position='bottom-left'] A string indicating the control's position on the map. Options are `'top-right'`, `'top-left'`, `'bottom-right'`, and `'bottom-left'`.
+ * @param {integer} [options.maxWidth='150'] A number indicating the maximum length of the scale control in pixels.
  * @example
  * map.addControl(new mapboxgl.Scale({position: 'top-left'})); // position is optional
+ * map.addControl(new mapboxgl.Scale({maxWidth: 80})); //maxWidth is optional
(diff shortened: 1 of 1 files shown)


js/ui/control/scale.js
```diff
@@ -12,8 +12,10 @@ module.exports = Scale;
  * @class Scale
  * @param {Object} [options]
  * @param {string} [options.position='bottom-left'] A string indicating the control's position on the map. Options are `'top-right'`, `'top-left'`, `'bottom-right'`, and `'bottom-left'`.
+ * @param {integer} [options.maxWidth='150'] A number indicating the maximum length of the scale control in pixels.
```
mourner: `integer` -> `number`; you could also remove `A number indicating ` and preserve the meaning fully

js/ui/control/scale.js
```diff
@@ -26,30 +28,35 @@ Scale.prototype = util.inherit(Control, {
...
+function updateScale(map, scale, options) {
     // A horizontal scale is imagined to be present at center of the map
-    // container with maximum length (Also the initial length) as 100px.
+    // container with maximum length (Default) as 100px.
     // Using spherical law of cosines approximation, the real distance is
     // found between the two coordinates.
+    var maxWidth = 100;
+    if (options && options.maxWidth) {
+        maxWidth = options.maxWidth;
+    }
```
mourner: You could write this shorter:

```js
var maxWidth = options && options.maxWidth || 100;
```

amishas157:
Thanks @mourner for the review.

lucaswoj:
Thanks @amishas157! ❤️

PR #3043, opened 2016-08-22
Include test for mapbox-gl module

anandthakker:
`test/js/mapbox-gl.js` is currently not being run by `npm test`, because it is targeting `test/js/*/*.js`.  This change adds that test file explicitly, and fixes a require path.

Note that the nature of this change is such that it will cause coveralls to report a test coverage regression.


lucaswoj:
Good catch! It looks like `test/js/ui/control/attribution.test.js` isn't being run either. Pushing another commit to leverage `tap`'s built recursive directory walking so that we can hit both `test/js/mapbox-gl.js` and `test/js/ui/control/attribution.test.js`.

Back to @anandthakker for 👍 and 🚢

anandthakker:
👍 LGTM, merging!

Issue #2931, opened 2016-08-02
Add example showing use of Marker#setPopup


mollymerp:
In reviewing @YunjieLi 's Marker demo, I discovered that a `Popup` created on a click event to another DOM event, without `closeOnClick` set to `false`, will close itself immediately after being created giving the appearance that the popup class didn't work. 

Minimal example [here](https://jsfiddle.net/madedwqv/2/)

We should stop propagation by default in `Popup()`

jfirebaugh:
Interesting example! Can you say more about what you mean by stopping propagation in `Popup`? It's not clear to me how that would resolve the situation here:

1. `click` event sent to `marker_el` listener, which creates and adds the `Popup`.
2. `click` event bubbles to the canvas container, triggering the map `click` event.
3. Map `click` event triggers `Popup:: _onClickClose`.

The possible solutions I see are:

1. Fix this specific example directly, by stopping propagation in the `marker_el` listener.
2. In `Popup::addTo`, delay binding `Popup:: _onClickClose` until the next tick.

I prefer the first solution, as in my experience when you mix asynchronicity and event bindings at the library level, it starts to become difficult to reason about the behavior. In Leaflet I recall seeing fixes of this sort cause regressions elsewhere.

mollymerp:
Ah yes, you're right @jfirebaugh stopping propagation on `Popup` doesn't make any sense. 

Leaflet implements its own `DomEvent` class that has a [`stopPropagation` method](https://github.com/Leaflet/Leaflet/blob/cd30e283cf3b20453f7bffe300e0a1459e5c2060/src/dom/DomEvent.js#L156-L168), and I agree that we should avoid trying to add this layer of abstraction / complexity to DOM interactions in GL-JS. 

I suppose a solution would be to add an example showing how you'd need to implement a click listener on `Marker` to add a `Popup` because I think the use case is fairly common.

jfirebaugh:
Good point, it is a very common use case. We could also consider a [`bindPopup`](http://leafletjs.com/reference.html#marker-bindpopup)-like API for `Marker`.

mollymerp:
Oh I like that idea! I'll work on a PR.

jfirebaugh:
`Marker#setPopup` landed; now we should make an example that uses it.

dasulit:
I added this example in this PR: https://github.com/mapbox/mapbox-gl-js/pull/3287

PR #3056, opened 2016-08-23
Add bindPopup API to the Marker class 

mollymerp:
This PR introduces methods for using `Popup`s in conjunction with `Marker`s and is not-so-loosely based on [Leaflet's API of the same name](http://leafletjs.com/reference.html#marker-bindpopup). 

Also includes the first DOM tests in the gl-js unit test suite using [`jsdom`](https://github.com/tmpvar/jsdom). I had to make some changes to `js/util/dom` in order for the tests to complete without errors, but there might be other workarounds using jsdom that I'm not aware of. 

Provides a fix for the problem of adding a `Popup`on a `click` event to a `Marker` with the current API -- the Popup is created and then immediately destroyed unless you have `Popup.closeOnClick: false` because the click event bubbles up to the map canvas. Related discussion at #2931 

## Checklist

 - [x] [document any changes or additions to public APIs](https://github.com/mapbox/mapbox-gl-js/blob/master/docs/README.md)
 - [ ] [post scores for all benchmarks on your branch and `master`](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/README.md#running-benchmarks)
 - [ ] [do a quick sanity check on the debug page](https://github.com/mapbox/mapbox-gl-js/blob/master/CONTRIBUTING.md#serving-the-debug-page)

### DIFF
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -5,7 +5,6 @@ module.exports = Marker;
 
 var DOM = require('../util/dom');
 var util = require('../util/util');
-var Evented = require('../util/evented');
 var LngLat = require('../geo/lng_lat');
 var Point = require('point-geometry');
 var Popup = require('./popup');
@@ -33,7 +32,7 @@ function Marker(element, options) {
     this._update = this._update.bind(this);
 }
 
-Marker.prototype = util.inherit(Evented, {
+Marker.prototype = {
     /**
      * Attaches the marker to a map
      * @param {Map} map
@@ -151,7 +150,7 @@ Marker.prototype = util.inherit(Evented, {
 
     _openPopup: function(e) {
         // prevent event from bubbling up to the map canvas
-        // e.stopPropagation();
+        e.stopPropagation();
         if (!this._popup || !this._map) return;
 
         if (!this._popup._map) {
@@ -174,4 +173,4 @@ Marker.prototype = util.inherit(Evented, {
         var pos = this._map.project(this._lngLat)._add(this._offset);
         DOM.setTransform(this._el, 'translate(' + pos.x + 'px,' + pos.y + 'px)');
     }
-});
+};
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -4,8 +4,11 @@
 module.exports = Marker;
 
 var DOM = require('../util/dom');
+var util = require('../util/util');
+var Evented = require('../util/evented');
 var LngLat = require('../geo/lng_lat');
 var Point = require('point-geometry');
+var Popup = require('./popup');
 
 /**
  * Creates a marker component
@@ -30,7 +33,7 @@ function Marker(element, options) {
     this._update = this._update.bind(this);
 }
 
-Marker.prototype = {
+Marker.prototype = util.inherit(Evented, {
     /**
      * Attaches the marker to a map
      * @param {Map} map
@@ -59,6 +62,7 @@ Marker.prototype = {
         }
         var parent = this._el.parentNode;
         if (parent) parent.removeChild(this._el);
+        if (this._popup) this._closePopup();
         return this;
     },
 
@@ -77,6 +81,7 @@ Marker.prototype = {
      */
     setLngLat: function(lnglat) {
         this._lngLat = LngLat.convert(lnglat);
+        if (this._popup) this._popup.setLngLat(this._lngLat);
         this._update();
         return this;
     },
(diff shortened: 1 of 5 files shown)


js/ui/marker.js
```diff
@@ -85,6 +89,98 @@ Marker.prototype = {
...
+     * @param {boolean} [options.closeOnClick=true] If `true`, the popup will closed when the
+     *   map is clicked.
+     * @param {string} options.anchor - A string indicating the popup's location relative to
+     *   the coordinate set via [Popup#setLngLat](#Popup#setLngLat).
+     *   Options are `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`,
+     * `'top-right'`, `'bottom-left'`, and `'bottom-right'`.
+     * @returns {Marker} `this`
+     */
+
+    bindPopup: function(content, options){
```
jfirebaugh: Let's make `Popup` the only type accepted by `bindPopup`, and remove `options`. This separates concerns:

* `Popup` is responsible for construction and configuration of the popup.
* `bindPopup` is responsible for linking it with a `Marker`. It doesn't need to pass through all functionality provided by `Popup`.

js/ui/marker.js
```diff
@@ -85,6 +89,98 @@ Marker.prototype = {
...
+            }
+        }
+    },
+
+    /**
+     * Returns the Popup instance that is bound to the Marker
+     * @returns {Popup} popup
+     */
+    getPopup: function(){
+        if (this._popup) {
```
jfirebaugh: No need to guard the return with `if (this._popup)` here.

js/ui/marker.js
```diff
@@ -85,6 +89,98 @@ Marker.prototype = {
...
+            return this._popup;
+        }
+    },
+
+    /**
+     * Unbinds a Popup from the Marker instance
+     * Returns the Popup instance that was bound to the Marker
+     * @returns {Popup} popup
+     */
+    unbindPopup: function(){
```
jfirebaugh: The accessor convention we've been following in GL JS is to provide reset/unbind semantics via setters that accept `null`/`undefined`. If we applied that pattern here, it would look like:

* `marker.setPopup(popup)` to bind a popup
* `marker.getPopup()` (getter as is)
* `marker.setPopup(null)` to unbind a popup

I think I prefer the consistency with other parts of GL JS to consistency with Leaflet. WDYT?

mollymerp: Makes sense to me! Good to know 👍

test/js/ui/marker.test.js
```diff
@@ -0,0 +1,105 @@
...
+        return;
+    }
+    global.document = jsdom.jsdom(html);
+    global.window = global.document.defaultView;
+    global.navigator = {
+        userAgent: 'JSDOM'
+    };
+    global.HTMLElement = global.window.HTMLElement;
+
+}
```
lucaswoj: Can you refactor this so that we don't need to use global variables? You should be able to do everything you need with access to the `window` object within test-scope.

mollymerp: I was having issues with the functions run by the tests failing because document + window weren't in scope. I refactored to use the [`jsdom-global`](https://github.com/rstacruz/jsdom-global) module which provides a `cleanup` function that removes the created variables from the global scope and can be called when the tests are complete. Is this acceptable? 

Otherwise we can take the approach that [`mapbox-gl-directions`](https://github.com/mapbox/mapbox-gl-directions/blob/master/package.json#L16) uses (browserify + smokestack + tap) but that would require some refactoring of the test suite.

mollymerp:
@lucaswoj copying from my response to your 'outdated' line comment:
I was having issues with the functions run by the tests failing because document + window weren't in scope. I refactored to use the [`jsdom-global`](https://github.com/rstacruz/jsdom-global) module which provides a `cleanup` function that removes the created variables from the global scope and can be called when the tests are complete. Is this acceptable? 

Otherwise we can take the approach that [`mapbox-gl-directions`](https://github.com/mapbox/mapbox-gl-directions/blob/master/package.json#L16) uses (browserify + smokestack + tap) but that would require some refactoring of the test suite.

lucaswoj:
I put together a PR that removes all uses of browser globals in #3063. I vote that we merge this as-is, ship it in the release today, and tweak the `Marker` tests within #3063.

js/ui/marker.js
```diff
@@ -75,19 +78,91 @@ Marker.prototype = {
...
+     * Binds a Popup to the Marker
+     * @param {Popup=} popup an instance of the `Popup` class. If undefined or null, any popup 
+     * set on this `Marker` instance is unset
+     * @returns {Marker} `this`
+     */
+
+    setPopup: function (popup) {
+        if (popup == null) {
+            this._closePopup();
+            delete this._popupHandlersAdded;
```
jfirebaugh: Instead of tracking state in `_popupHandlersAdded`, call`removeEventListener` when the popup is removed. This will prevent memory leaks and unexpected behavior if the `Marker` continues to be used without a popup. You'll need to save the `this`-bound version of `_openPopup`, say by using `util.bindAll(['_openPopup'], this)`, so that it can be passed to `removeEventListener`.

js/ui/marker.js
```diff
@@ -52,21 +54,22 @@ Marker.prototype = {
...
      */
-    remove: function() {
+    remove: function () {
         if (this._map) {
             this._map.off('move', this._update);
             this._map = null;
         }
         var parent = this._el.parentNode;
         if (parent) parent.removeChild(this._el);
+        if (this._popup) this._closePopup();
```
jfirebaugh: `this._closePopup` has an internal `if (this._popup)` guard -- either call it unguarded here, or add external guards everywhere it's called and remove the internal guard.

js/ui/marker.js
```diff
@@ -75,19 +78,91 @@ Marker.prototype = {
...
+    /**
+     * Opens or closes the bound popup, depending on the current state
+     * @returns {Marker} `this`
+     */
+    togglePopup: function () {
+        if (this._popup) {
+            if (this._popup._map) {
+                this._closePopup();
+            } else {
+                this._openPopup();
```
jfirebaugh: This will throw an error because `_openPopup` is expecting an event argument.

Issue #2828, opened 2016-07-06
Fix unexpected results for fill-opacity property function


jfirebaugh:
```
        "fill-opacity": {
          "property": "property",
          "stops": [
            [
              0,
              1
            ],
            [
              4,
              0
            ]
          ]
        }
```

For values `0`, `1`, and `2`, I'd expect the rendering to look like:

![image](https://cloud.githubusercontent.com/assets/98601/16636454/14595bf6-438d-11e6-87fd-5930263d72f3.png)

But instead it renders everything with opacity 0.

Possibly related to #2827.

lucaswoj:
Fixed by #2971

jfirebaugh:
https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/fill-opacity/property-function is still failing on gl-js `master`.

Issue #3044, opened 2016-08-22
Deprecate "StyleLayer#paint" and "StyleLayer#layout"


lucaswoj:
`StyleLayer#paint` and `StyleLayer#layout` do not accommodate for data-driven styling. Any code that uses these properties probably doesn't accommodate for data-driven styling either. We should deprecate these properties in favor of `StyleLayer#getPaintValue` and `StyleLayer#getLayoutValue`. This might be done in conjunction with #2739. 

cc @davidtheclark

davidtheclark:
@lucaswoj So this is the root of the problem I was seeing? — that `queryRenderedFeatures` seemed to always look for circles that are as large as they *might* be instead of as large as they actually are?

lucaswoj:
@davidtheclark Worse, actually. `queryRenderedFeatures` assumed all circles were the size specified in the first function stop 😬

jfirebaugh:
I think we need to do this before adding further data-driven properties -- it's hiding a lot of bugs.

PR #3061, opened 2016-08-24
Unignore fill-opacity property-function test

lucaswoj:
fixes https://github.com/mapbox/mapbox-gl-js/issues/2828

cc @jfirebaugh


lucaswoj:
Root cause is #3044

lucaswoj:
Release time 🚢

PR #3062, opened 2016-08-24
Add perspective option to custom markers

yeldarby:
This pull request adds the option to make markers scale with the map perspective. This comes into play when zooming or changing the bearing on a pitched map.

This also fixes z-indexing of markers (previously they were z-indexed haphazardly; this changes them to be z-indexed with the closest marker on top using translate3d).

Example usage:

```
var marker = new mapboxgl.Marker(null, {
        perspective: true
    }) .setLngLat(toSet).addTo(map);
```

marker-perspective


yeldarby:
In thinking about this on my drive home I no longer think setting the default zoom of the marker to be the zoom of the map at the time it is added is a good idea.

Tomorrow I plan to update the default zoom level of he marker to be user configurable instead.

Reasoning for this being: imagine you add a marker, the user zooms out then you add another marker. One will be big one will be small currently which is probably not desired in most instances. Current functionality would still be easy for the developer to opt into using map.getZoom() as the value for native zoom.

js/ui/marker.js
```diff
@@ -13,6 +13,7 @@ var Point = require('point-geometry');
  * @param {HTMLElement=} element DOM element to use as a marker (creates a div element by default)
  * @param {Object=} options
  * @param {PointLike=} options.offset The offset in pixels as a [`PointLike`](#PointLike) object to apply relative to the element's top left corner. Negatives indicate left and up.
+ * @param {boolean=} [options.perspective=false] Whether the marker should scale along with the map geometry.
```
1ec5: Perspective can mean many things. For example, in some cases it might be desirable to not only scale a marker but also tilt it to lie flat against the tilted map. (More ideas at mapbox/mapbox-gl-native#5245.)

In the meantime, merely scaling the marker would still be desirable functionality; how about a more descriptive term, like `scalesWithViewingDistance`?

1ec5: Or perhaps `pitchScale: "map"`, for consistency with the [`circle-pitch-scale: map`](https://www.mapbox.com/mapbox-gl-style-spec/#paint-circle-pitch-scale) style property?

yeldarby:
That's a good point. I will read through that other thread and update the nomenclature tomorrow to try to maintain consistency.

yeldarby:
Ok, I've gone ahead and split this out into the two different options.

`pitchScale` for scaling the marker as the map is pitched
and `zoomScale` for scaling the marker as the map is zoomed

`zoomScale` also has some other configurable options (how much to zoom the marker and what the native zoom level for the marker is).

It's likely that in the future we may want to add a maxZoom and minZoom value and possibly an option to pass a zoom function to determine scale by zoom as well in case you want custom values.

dist/mapbox-gl.css
```diff
@@ -5,6 +5,12 @@
     -webkit-tap-highlight-color: rgba(0,0,0,0);
 }
 
+.mapboxgl-canvas-container {
+    -webkit-transform-style: preserve-3d;
+    -moz-transform-style: preserve-3d;
+    transform-style: preserve-3d;
```
lucaswoj: Unfortunately [`preserve-3d` is not supported in one of our supported platforms: IE11](http://caniuse.com/#feat=transforms3d)

js/geo/transform.js
```diff
@@ -346,6 +356,19 @@ Transform.prototype = {
...
+     * @param {Coordinate} coord
+     * @returns {Point} screen point
+     * @private
+     */
+    coordinatePoint3d: function(coord) {
+        var scale = this.worldSize / this.zoomScale(coord.zoom);
+        var p = [coord.column * scale, coord.row * scale, 0, 1];
+        vec4.transformMat4(p, p, this.pixelMatrix);
+        return p;
+    },
```
lucaswoj: This code was copied from `Transform#coordinatePoint`. Can you refactor to eliminate the code duplication?

js/geo/transform.js
```diff
@@ -260,6 +260,16 @@ Transform.prototype = {
...
 
     /**
+     * Like `locationPoint` but returns a raw 4d vector containing info about the transform
+     * @param {LngLat} lnglat location
+     * @returns {Array} The 4d vector pertaining to this point's transform
+     * @private
+     */
+    locationPoint3d: function(lnglat) {
+        return this.coordinatePoint3d(this.locationCoordinate(lnglat));
+    },
```
lucaswoj: What do you think about extending `Point` such that it includes a 3rd dimension rather than adding new methods to create a 3d point?

docs/README.md
```diff
@@ -42,7 +42,7 @@ npm run start-docs
 You can view the documentation at
 
 ```bash
-open http://127.0.0.1:4000/mapbox-gl-js
+open http://127.0.0.1:4000/mapbox-gl-js/
```
lucaswoj: Please revert this change. The related bug was fixed in #3069

dist/mapbox-gl.css
yeldarby: This should fall back to the behavior currently in production in IE11. Using preserve-3d lets us set the z-indexing of markers with a single CSS transform.

The workaround for IE would probably be to manually set the z-Index of the elements instead of using a transform on the z axis (which would likely be slower but should get the job done).

What would be the best way to detect when the client is IE11? Usually I'd use Modernizr but it doesn't look like you guys are pulling that in currently.

js/geo/transform.js
yeldarby: Yep, can do.

js/geo/transform.js
yeldarby: I'm pretty sure using 3d points would be a major undertaking across the codebase.

Right now these points don't actually represent a the position of the point on the map; just the position of the point on the screen. Everything is being projected down to 2d. Trying to match the proper 3d to 2d mapping with CSS using an x-y-z point instead would be hard.

dist/mapbox-gl.css
lucaswoj: The testing and debugging overhead of a separate codepath for IE11 would likely be a blocker for us. Have you measured the performance of manually setting the z-index in all browsers?

js/geo/transform.js
lucaswoj: My impression is that that the only difference between `locationPoint3d` and `locationPoint` is an extra property object returned from `locationPoint3d`. Is that not the case?

js/geo/transform.js
```diff
@@ -345,27 +336,17 @@ Transform.prototype = {
...
-     * @returns {Point} screen point
-     * @private
-     */
-    coordinatePoint3d: function(coord) {
-        var scale = this.worldSize / this.zoomScale(coord.zoom);
-        var p = [coord.column * scale, coord.row * scale, 0, 1];
-        vec4.transformMat4(p, p, this.pixelMatrix);
-        return p;
+        if(shouldReturnRawVector) return p;
+        return new Point(p[0] / p[3], p[1] / p[3]);
```
lucaswoj: 👎  on changing the return type based on an argument. Two seperate methods are preferable. 

It looks like this code could be written most succinctly if `coordiantePoint` called `coordinatePoint3d`

```
coordinatePoint: function() {
    var p = this.coordinatePoint3d(coord);
    return new Point(p[0] / p[3], p[1] / p[3]);
}
```

dist/mapbox-gl.css
yeldarby: Looks like the performance impact is actually relatively minor. I did some tests using `debug/markers3d.html` using 50 markers to check the framerate with different z-Indexing strategies.

No markers (baseline):
baseline-no-markers

Current version in production (ignores zIndexing and scaling completely):
old-random-ordering

With translate3d
ordering-with-translate3d copy

With zIndex property
setting-with-zindex

js/geo/transform.js
yeldarby: Thanks; I've refactored it back so that coordinatePoint calls coordinatePoint3d

dist/mapbox-gl.css
yeldarby: I changed the functionality to set the zIndex by default so it will work as expected in IE11 with an option to disable if the speed bump is necessary.

lucaswoj: I'm glad to hear that performance of the `z-index` option is good! 🎉 

We do not feel comfortable including both the `z-index` and `transform-style` options. Having two options will roughly double the amount of testing, support, and debugging required by Mapbox & our contributors. Let's drop the option and stick with just `z-index` for now.

debug/markers3d.html
```diff
@@ -0,0 +1,93 @@
...
+    var frame = function(t) {
+        map.setBearing(t/100);
+        requestAnimationFrame(frame);
+    };
+    requestAnimationFrame(frame);
+});
+</script>
+
+</body>
+</html>
```
lucaswoj: It makes sense that you built a debug page to use while developing this feature. 

Because repository size, simplicity, and stability is important to us, we would prefer not to merge this page into `master`.

lucaswoj:
@mourner This is close-to-ready from an architectural standpoint. I am interested to hear your thoughts on the concept and implementation.

js/geo/transform.js
yeldarby: Whoops, missed this. There are actually 2 other properties on there that are coming from a transform over pixelMatrix via Transform#_calcMatrices.

Honestly, I'm not sure what the third one is but the fourth does relate to depth.

The problem is x and y are projected into 2d space; if you were to have a true 3d point your x and y would need to be non-projected values in addition to adding z otherwise z wouldn't make sense.

docs/_posts/examples/3400-01-24-custom-marker-with-perspective.html
```diff
@@ -0,0 +1,78 @@
...
+    var m = marker.getElement();
+    m.onclick = function(e) {
+        alert('Hello world');
+        e.stopPropagation();
+    };
+
+    m.style.backgroundColor = 'rgb(' + [Math.floor(Math.random() * 16) * 16, Math.floor(Math.random() * 16) * 16, Math.floor(Math.random() * 16) * 16].join(',') + ')';
+}
+
+for (var i = 0; i < points; i++) addAnotherMarker();
```
lucaswoj: I appreciate the aesthetics of this demo and its utility for debugging this PR. Our intent with the GL JS examples is to provide the simplest possible code for demonstrating a feature, making it easy for users to see how they can adapt the feature to their use case. In this case, the simplest possible code would probably look like instantiating a single `pitchScale`-ed and`zoomScale`-ed marker on a tilted map.

js/geo/transform.js
lucaswoj: I understand. These methods are 👍 for now.

docs/_posts/examples/3400-01-24-custom-marker-with-perspective.html
yeldarby: In that case, how would you feel about updating the existing marker example instead to pass these options?

js/ui/marker.js
```diff
@@ -13,6 +13,10 @@ var Point = require('point-geometry');
  * @param {HTMLElement=} element DOM element to use as a marker (creates a div element by default)
  * @param {Object=} options
  * @param {PointLike=} options.offset The offset in pixels as a [`PointLike`](#PointLike) object to apply relative to the element's top left corner. Negatives indicate left and up.
+ * @param {boolean=} [options.pitchScale=false] Whether the marker should scale smaller into the background or larger into the foreground when the map is pitched and rotated
+ * @param {boolean=} [options.zoomScale=false] Whether the marker should scale with the map as you zoom in and out
+ * @param {number=} [options.nativeZoom=map.getZoom()] The native zoom level of the marker. If the map is at this zoom the marker will not be scaled. If zoom is greater the marker will be enlarged. If zoom is less the marker will be shrunk.
+ * @param {number=} [options.zoomFactor=1.6] The multiple by which to scale the size of the marker between each zoom level
```
lucaswoj: With `nativeZoom` and `zoomFactor` it feels like we're almost re-inventing [style functions](https://www.mapbox.com/mapbox-gl-style-spec/#types-function). What do you think about re-using the function [syntax](https://www.mapbox.com/mapbox-gl-style-spec/#types-function) and [evaluation logic](https://github.com/mapbox/mapbox-gl-function) here?

```js
var marker = new mapboxgl.Marker(null, {
    pitchScale: true,
    zoomScale: {
        stops: [[15, 1], [16, 1.6]],
        base: 2
    },
    offset: [-25, -50]
});
```

lucaswoj:
This is great! We're so close to being able to 🚢. Let me know if you have any questions or need any assistance 🙇

js/ui/marker.js
yeldarby: Oh nice. Yeah that sounds like a good idea! I will have a stab at that

docs/_posts/examples/3400-01-24-custom-marker-with-perspective.html
lucaswoj: There are situations in which our ability to explain concepts to users would be aided by having both a bare bones marker example and a perspective-scaled marker example. Let's keep this as two examples for now.

js/ui/marker.js
yeldarby: So I'm trying to implement this and I think I'm close..  but how to I get the style functions to overflow past the provided values?

For instance, for your example:
```
var f = glfunc.interpolated({
    stops: [[15, 1], [16, 1.6]],
    base: 2
});
```

This works for the range of [15, 16] as expected:
```
f(15); // => 1
f(15.5); // => 1.248528137423857
f(16); // => 1.6
```
But anything that falls outside that range is bounded by the highest and lowest defined stops:
```
f(14); // => 1
f(17); // => 1.6
```

How do I get it to extrapolate out the correct values beyond the first and last stops without having to rewrite the logic in the constructor?

lucaswoj: > But anything that falls outside that range is bounded by the highest and lowest defined stops

This is the intended behaviour of GL functions and reflects how GL styles are crafted.

 > How do I get it to extrapolate out the correct values beyond the first and last stops without having to rewrite the logic in the constructor?

If you want extrapolation over a larger domain, you need to declare a larger domain 😄 

```js
var f = glfunc.interpolated({
    stops: [[1, 1], [20, 5]],
    base: 2
});
```

yeldarby: The problem, though, is that in in order to do that you have to calculate the desired value of the function at stops 1 and 20 anyway which defeats the purpose of delegating the calculations to the function in the first place.

Eg to get the marker to scale with the map 1:1 in the zoom range [1, 20] you'd have to use

```
stops: [
    [1, Math.pow(base, 1-map.getZoom())]
    [map.getZoom(), 1],
    [20, Math.pow(base, 20-map.getZoom())]
]
```

Whereas without using `mapbox-gl-function` you just need to set the scale to

```
Math.pow(base, currentZoom - nativeZoom)
```

lucaswoj: Strictly speaking, in order to get the marker to scale with the ma 1:1 in the zoom range `[1, 20]` all you'd need to do is
```
stops: [
    [1, 1/1024]
    [20, 1024]
]
```

I agree the existing function syntax specification does not lend itself to an elegant implementation of your use case. That's a problem with the existing specification, not a reason to not use the existing specification. 

I'd be happy to entertain proposals for extending the function syntax or helper functions to generate function stops with certain characteristics.

yeldarby: I don't think that works unless you want your native scale to be 10 (which you almost never would). For example, here's the marker.html debug page with those stops:

image

I don't really see a good way to add this without completely hacking apart `mapbox-gl-function`

What about just not giving the user an option, getting rid of zoomFactor and just making zoomScale be an integer that represents the native zoom level of the marker.

Eg 

```
var marker = new mapboxgl.Marker(null, {
    zoomScale: 12
});
```

Means "create a marker with a scale of 1 at zoom 12, keep the marker the same size relative to the map as you zoom".

lucaswoj: > `stops: [[1, 1/512], [20, 1024]]`

 > I don't think that works unless you want your native scale to be 10

With a little math, you can adjust your "native scale" to be whatever you choose. 

> I don't really see a good way to add this without completely hacking apart `mapbox-gl-function`

Let's hack it! 😄 This is a highly requested feature for `mapbox-gl-function`. I'm more than happy to help you with that PR.

Some syntax ideas: 

```
// use a value which represents 500m of "real world" distance (would require more work)
{
    type: 'unit',
    value: 500
    units: 'meters' 
}

// at zoom 5, use a value of 1. scale the value proportionally with the zoom
{
    type: 'zoom-interpolated',
    zoom: 5,
    value: 1       
}
```

yeldarby: That's what I mean; that "little math" is exactly what the function is supposed to be computing ;P If that logic needs to be duplicated in the userland code it may as well be in Marker.js instead.

`zoom-interpolated` seems like it'd probably work; it'd probably need a `base` parameter too as a scale factor. I'd say `interpolated` is bad nomenclature because it specifically means to "insert (something) between fixed points".

Maybe `exponential-overflow` instead:

```
{
    type: 'exponential-overflow',
    stops: [[12, 1]],
    base: 2
}
```

It would work the exact same as `exponential` but when outside the bounds of the stops passed it would calculate an inferred overflow value.

lucaswoj: Good idea! What do you think of making `overflow` an option on the existing `exponential` function type?

```js
{
    type: 'exponential',
    overflow: true,
    stops: [[12, 1]],
    base: 2
}
```

Internally, we've usually referred to this behaviour as "clamping"

```js
{
    type: 'exponential',
    clamp: false,
    stops: [[12, 1]],
    base: 2
}
```

yeldarby: That seems like a good solution to me. My vote would be for `overflow` but `clamp` sounds cromulent as well.

lucaswoj:

PR #3063, opened 2016-08-25
Eliminate use of browser globals, introduce jsdom in test environment

lucaswoj:
This PR proxies all access to browser globals through a new `js/util/window` module. In the browser this module exports `window` normally. In the testing environment `window` is provided by jsdom. 

jsdom is a lightweight solution for DOM testing compared to Selenium. I have used it on other projects at Mapbox and found it to be stable, complete, and only slightly quirky. 

With this change we will be able to begin writing tests for interaction handlers, UI controls and `js/util/browser`. #1550 

Related to #3056 

cc @mollymerp @jfirebaugh @mourner @mapsam

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,6 +14,7 @@
     "space-before-function-paren": "off",
   },
   "env": {
-    "es6": true
+    "es6": true,
+    "browser": false
   }
 }
--- a/bench/.eslintrc
+++ b/bench/.eslintrc
@@ -7,5 +7,8 @@
   "globals": {
     "React": false,
     "ReactDOM": false
+  },
+  "env": {
+    "browser": true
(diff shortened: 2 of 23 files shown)


jfirebaugh:
I'm excited to see us start to test our DOM-dependent code!

Do we want to try to isolate tests with a fresh JSDom `window` for each test case?

lucaswoj:
> Do we want to try to isolate tests with a fresh JSDom `window` for each test case?

Yes! My plan was to do this progressively: using `proxyquire` to overwrite the `js/util/window` module with a fresh instance of JSDom within the scope of tests that manipulate the DOM. I haven't investigated what it'd look like to do this globally within each test.

PR #3067, opened 2016-08-25
rawTileData cleanup

jfirebaugh:
In preparation for https://github.com/mapbox/mapbox-gl-native/issues/5701, I reviewed the GL JS implementation of tile reloading and noticed a few issues.

* Whenever a `VectorTile` is created, we were accidentally making an extra copy of the data.
* This was hiding a latent bug, where the worker thread was giving up ownership of raw data that it needed to retain.
* We were copying raw data _back_ to the worker on tile reload requests for no reason.
* The naming of tile and data related variables in `VectorTileWorkerSource` was vague and confusing.
* The contract of the semi-public `loadVectorData` extension point was underspecified.

This PR fixes those issues.

### `master`

**load-multiple-maps:** 150 ms, loaded 6 maps.
**buffer:** 854 ms
**fps:** 60 fps
**frame-duration:** 7.6 ms, 0% > 16ms
**query-point:** 0.63 ms
**query-box:** 51.76 ms
**geojson-setdata-large:** 97 ms

### Branch

**load-multiple-maps:** 151 ms, loaded 6 maps.
**buffer:** 863 ms
**fps:** 60 fps
**frame-duration:** 7.6 ms, 0% > 16ms
**query-point:** 0.65 ms
**query-box:** 49.49 ms
**geojson-setdata-large:** 104 ms

## Checklist

 - [x] [post scores for all benchmarks on your branch and `master`](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/README.md#running-benchmarks)

### DIFF
--- a/js/source/vector_tile_worker_source.js
+++ b/js/source/vector_tile_worker_source.js
@@ -3,6 +3,7 @@ var ajax = require('../util/ajax');
 var vt = require('vector-tile');
 var Protobuf = require('pbf');
 var WorkerTile = require('./worker_tile');
+var util = require('../util/util');
 
 module.exports = VectorTileWorkerSource;
 
@@ -58,7 +59,12 @@ VectorTileWorkerSource.prototype = {
             if (!data) return callback(null, null);
 
             tile.data = data.tile;
-            tile.parse(tile.data, this.styleLayers.getLayerFamilies(), this.actor, data.rawTileData, callback);
+            tile.parse(tile.data, this.styleLayers.getLayerFamilies(), this.actor, function (err, result, transferrables) {
+                if (err) return callback(err);
+                callback(null,
+                    util.extend({rawTileData: data.rawTileData}, result),
+                    transferrables.concat(data.rawTileData));
+            });
 
             this.loaded[source] = this.loaded[source] || {};
             this.loaded[source][uid] = tile;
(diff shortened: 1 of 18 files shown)


mourner:
Looks great to me overall! Are we sure the main thread isn't blocked when `rawTileData` is transferred with a copy? Or does this only happen in the other direction, when the main thread sends to the worker?

js/source/vector_tile_worker_source.js
```diff
@@ -48,20 +49,27 @@ VectorTileWorkerSource.prototype = {
...
 
-            tile.data = data.tile;
-            tile.parse(tile.data, this.styleLayers.getLayerFamilies(), this.actor, data.rawTileData, callback);
+            workerTile.vectorTile = vectorTile;
+            workerTile.parse(vectorTile, this.styleLayers.getLayerFamilies(), this.actor, function (err, result, transferrables) {
+                if (err) return callback(err);
+
+                // Not transferring rawTileData because the worker needs to retain its copy.
+                callback(null,
+                    util.extend({rawTileData: vectorTile.rawData}, result),
```
lucaswoj: Can we set `result.rawTileData` within `WorkerTile#parse`? It does have access to `vectorTile`.

jfirebaugh: Why (re-)add coupling with `WorkerTile`? This is an internal detail of worker to main thread messaging that can be handled entirely by `VectorTileWorkerSource`.

jfirebaugh: Or perhaps more to the point -- `WorkerTile#parse` is used for both loading and reloading, and part of the purpose of this PR is to stop sending `rawTileData` for reloading. So if `WorkerTile#parse` were to set it, it would also have to know if it was being used for loading or reloading.

lucaswoj: Makes sense. Thanks for the explanation.

jfirebaugh:
@mourner I didn't investigate that. Do you think it's a blocker to getting this merged?

Issue #2792, opened 2016-06-28
Add "Map#isMoving" method


lucaswoj:
It is useful to know whether the map is currently moving (i.e. because of a `flyTo` call). This method would sit alongside `Map#loaded` (to be renamed to `Map#isTilesLoaded`?) as a guard against expensive operations during times of rapid map change

``` js
map.on('render', function() {
    if (!map.isMoving() && map.isTilesLoaded()) {
        map.queryRenderedFeatures(...)
    }
});
```

cc @peterqliu

lucaswoj:
I think this is a solid alternate solution for #2471

peterqliu:
👍  I think this would be useful for expensive operations that need to happen regularly ("keep doing this, *unless* the map is moving").

However, it shouldn't replace #2471 altogether. For occasional operations, it's easier to reason about what's going on via event callbacks (moveend, dataend, renderstable, etc), rather than constantly checking the map state before executing something.

averas:
There is currently issues with Mapbox GL JS Draw in regards to adding/removing draw in a controlled fashion (https://github.com/mapbox/mapbox-gl-draw/issues/572). This is in part related to that adding Draw currently depends on <code>Map#loaded</code> returning true, which during long transitions can take a while. If draw is removed before it has been successfully enabled (i.e. loaded returns true) Draw enters a degenerate state which is hard to recover from.

Even though this is more a concern for the Draw project rather than Mapbox GL JS I wonder if <code>Map#isMoving</code> and <code>Map#isTilesLoaded</code> covers the potential state queries a client or plugin would want to make. I imagine that adding draw would be possible before these states are fulfilled, but after the map's style is initialised.

I guess my question boils down to - What is the earliest state the map can reach when it is reasonable to add/remove controls/sources/layers and is it possible to query for _this_ state?

lucaswoj:
> What is the earliest state the map can reach when it is reasonable to add/remove controls/sources/layers and is it possible to query for this state?

This state is when `Map#loaded` returns `true`, also marked by the `Map#load` event. If that doesn't work, I'd consider it a bug.

averas:
@lucaswoj Currently <code>Map#loaded</code> returns false during any kind of transition, meaning that a 10s flyTo transition will block a plugin/client (such as Draw) from enabling itself if it is specifically waiting for <code>Map#loaded</code> to return true (which it is).

[...]

Issue #2801, opened 2016-06-29
Unify camera API into a single method


lucaswoj:
As part of the v1 effort, we should consider refactoring the camera API
- the names of methods like `panBy`, `panTo`, `flyTo`, `jumpTo`, `easeTo` don't make their behaviour obvious (I need to reference the docs regularly)
- there is an opportunity to express the camera API in terms of fewer methods (maybe even a single method 😄)

1ec5:
👍 on unifying `panTo`, `flyTo`, `jumpTo`, and `easeTo` for sure: it’d be nice if `AnimationOptions` were the one-stop shop for specifying how you get to the destination camera, and the default were to have no animation at all.

On the other hand, `panBy` is quite handy. It isn’t always obvious how to pan by a relative pixel distance, especially when bearing and pitch are involved. Could it be generalized to accept a options object that could vary the zoom level or bearing instead?

lucaswoj:
Combining this ticket with #2820, "Support overlapping and/or queued Camera animations"

1ec5:
The equivalent gl-native ticket is mapbox/mapbox-gl-native#3625.

mapsam:
Brainstorming:

## moveTo

Everything within the current animation methods involve moving the map based on a set of options (i.e. bearing, speed, zoom, etc.). What if we were to integrate all of them into a single method, called `moveTo` that includes an additional option `animationType`, which specifies how to animate the motion.

Example usage:

```javascript
map.moveTo({
  animationType: 'ease',
  center: [0, 0],
  // more options
});
```

This would allow us to set the `moveTo` animation default (possibly no animation whatsoever, replacing `jumpTo`). Internally we can update all motion to `moveTo` and extend their options to include `animationType`, which would reduce the number of methods used internally.

mapsam:
An alternative to the above ^

## moveTo w/ animation object

The example above lumps movement and animation all into the same `options` object. We could make animations, currently `AnimationOptions` a subset of `options`, so you'd write out your `moveTo` function like this:

```javascript
map.moveTo({
  center: [0, 0],
  bearing: 0,
  pitch: 0,
  zoom: 10,
  animation: {
    type: 'ease', // 'fly', 'none', 'etc'
    duration: 1000,
    easing: function(f) { return f; } 
  }
});
```

I think we can remove the `animate: false` option, and just default to no animations. Internal usage of `moveTo` could set animation then.

1ec5:
[...]

PR #2259, opened 2016-03-11
Speed up "camera flyTo ascends" test

lucaswoj:
This PR makes our unit tests 73% faster, from 19 seconds to 5 seconds.

fixes #2252 

### Before

```
$  time prova test/js/*/*.js -q

    990 passed assertions.


real	0m18.970s
user	0m4.120s
sys     0m0.623s
```
### After

```
$  time prova test/js/*/*.js -q

    990 passed assertions.


real	0m5.051s
user	0m2.771s
sys     0m0.373s
```

cc @jfirebaugh @ansis @bhousel

PR #3068, opened 2016-08-25
map#isMoving

mapsam:
Adds a method `Map#isMoving` that checks for the states of `this.zooming`, `this.rotating`, `this.pitching`, `map.dragPan._active`, and `map.dragRotate._active`. If any of them is `true`, the map is considered "moving" and the method returns `true`.

mapbox-gl-jsismoving

refs: #2792 

TODO
- [ ] add tests, @lucaswoj I was thinking a test that looks something like this:

``` javascript
test('Map#isMoving returns expected results', function (t) {
    var map = createMap();
    var moving = false;
    var notMoving = false;

    function check () {
        if (map.isMoving()) moving = true;
        if (!map.isMoving()) notMoving = true;
    }

    map.on('load', function() {
        map.flyTo({center: [0, 0], zoom: 9});
    });

    map.on('render', check);

    map.on('moveend', function() {
        t.ok(moving, 'map was considered moving at one point');
        t.ok(notMoving, 'map was considered not moving at one point');
        t.end();
    });
});
```

But not having any luck with it so far.

cc @mollymerp @jfirebaugh

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -985,7 +985,7 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
-     * Returns a Boolean indicating whether the map is currently in a "moving" state. 
+     * Returns a Boolean indicating whether the map is currently in a "moving" state.
      * Returns `false` if zooming, rotating, pitching, and dragging are all inactive.
      *
      * @returns {boolean} A Boolean indicating whether the map is moving.
@@ -994,8 +994,8 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
         if (this.zooming ||
             this.rotating ||
             this.pitching ||
-            map.dragPan._active ||
-            map.dragRotate._active) return true;
+            this.dragPan._active ||
+            this.dragRotate._active) return true;
         return false;
     },
 
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -985,6 +985,24 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
+     * Returns a Boolean indicating whether the map currently in a "moving" state.
+     *
+     * Returns `false` if the style is not yet fully loaded,
+     * or if there has been a change to the sources or style that
+     * has not yet fully loaded.
+     *
+     * @returns {boolean} A Boolean indicating whether the map is fully loaded.
+     */
+    isMoving: function() {
+        if (this.zooming ||
+            this.rotating ||
+            this.pitching ||
+            map.dragPan._active ||
+            map.dragRotate._active) return true;
+        return false;
+    },
+
+    /**
      * Update this map's style and sources, and re-render the map.
      *
      * @param {boolean} updateStyle mark the map's style for reprocessing as
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -985,13 +985,10 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
     },
 
     /**
-     * Returns a Boolean indicating whether the map currently in a "moving" state.
+     * Returns a Boolean indicating whether the map is currently in a "moving" state. 
+     * Returns `false` if zooming, rotating, pitching, and dragging are all inactive.
      *
-     * Returns `false` if the style is not yet fully loaded,
-     * or if there has been a change to the sources or style that
-     * has not yet fully loaded.
-     *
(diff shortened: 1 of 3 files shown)


js/ui/map.js
```diff
@@ -985,6 +985,21 @@ util.extend(Map.prototype, /** @lends Map.prototype */{
...
+     * Returns `false` if zooming, rotating, pitching, and dragging are all inactive.
+     *
+     * @returns {boolean} A Boolean indicating whether the map is moving.
+     */
+    isMoving: function() {
+        if (this.zooming ||
+            this.rotating ||
+            this.pitching ||
+            this.dragPan._active ||
+            this.dragRotate._active) return true;
```
lucaswoj: Interesting that we need to mix the "interaction handler" layer of abstraction the "camera" layer of abstraction to compute `Map#isMoving`. Should `Map#isMoving` return `true` if the user starts a "drag rotate" interaction but doesn't actually rotate the map? Or if the user starts a "drag pan" interaction but doesn't pan the map?

lucaswoj:
Using `Map#flyTo` in tests is tricky! 

The default duration of `Map#flyTo` is rather long relative to our unit test execution time. In https://github.com/mapbox/mapbox-gl-js/pull/2259 we sped up our unit tests from 19 seconds to 5 seconds by shortening the duration of a `Map#flyTo` test 😬 

Perhaps the `render` event isn't getting fired as expected in our testing environment. Have you tried listening to `move` events instead?

js/ui/map.js
mapsam: Good questions! I'm not familiar enough with these interactions enough to know the differences. The only way I was able to capture "pan" and "rotate" actions was through those `dragPan` and `dragRotate` parameters, though. Are there other places to look for them?

> Should Map#isMoving return true if the user starts a "drag rotate" interaction but doesn't actually rotate the map? 

I'm not sure - any chance you could provide a case where this happens?

lucaswoj: > The only way I was able to capture "pan" and "rotate" actions was through those `dragPan` and `dragRotate` parameters, though. Are there other places to look for them?

We may need to create the place to look for this state by

 - creating `map.panning` property
 - ensuring that `map.rotating` and `map.rotating` works with the `dragRotate` handler

> I'm not sure - any chance you could provide a case where this happens?

Imagine I click and drag the map by 5 pixels, initiating a "drag pan" gesture, put a rock on my mouse button, and go for lunch. `Map#isMoving` will be `true` the whole time I'm gone but the map won't be moving. Is that the behaviour we want? I'm not sure. It deserves some discussion.

mapsam:
@lucaswoj just added `map.easing` as a property, which is enacted within `map._easeTo` and cancelled within `map._easeEnd`

Since `zooming`, `rotating`, and `pitching` are all within the ease functions, they're probably not 100% required any longer, but I haven't removed them just in case.

I also moved the `map.dragRotate` and `map.dragPan` checks to use their respective `isActive()` methods, rather than checking a private property. 

Next up, tests!

lucaswoj:
Let's define a "moving" map as a map whose center coordinate, bearing, zoom, or pitch is expected to be different in the next frame. Whether or not we "expect" these parameters to be different on the next frame during an interaction is an open question. 

The word "easing" feels overloaded:

 - the dictionary definition is to "move carefully, gradually, or gently"
 - it can refer to the `Map#easeTo` operation
 - it can refer to the internal `Camera#_ease` operation, used by both `Map#easeTo` and `Map#flyTo`

Let's think carefully about how we're using these words and how we can refactor to make their use more clear :-) https://github.com/mapbox/mapbox-gl-js/issues/2801

---- 

To answer your original question:

> wondering if instead of the `map.easing` property, can I just use `map.isEasing()`

Looks like this would work 👍

mapsam:
Sure thing @lucaswoj. Since we already have the `isEasing()` method there's no need to start adding more properties/language like `map.easing`. 

Since all of the potential "movements" have their own `isSomething()` methods except for `zooming` (i.e. there is no `isZooming()` method), it seems like we don't need to add anything else.

mapsam:
@lucaswoj I've added a test that checks if the map is moving before, during, and after a `flyTo` animation. Let me know what you think!

mapsam:
Closing this. I've got a better grasp on the camera movement, and think this can be better implemented once #3583 lands.

Issue #1511, opened 2015-09-22
Popular Mapbox.js example: show polygon area


lyzidiamond:
No. 1 most popular Mapbox.js example by analytics in the last year

https://www.mapbox.com/mapbox.js/example/v1.0.0/show-polygon-area/

colleenmcginnis:
New Show drawn polygon area example is live! see #3075 and https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-draw/

PR #3075, opened 2016-08-26
New example: draw and find area of a polygon

colleenmcginnis:
Adds a new example for drawing a polygon and calculating its area using Mapbox GL JS, mapbox-gl-draw, and turf.js per #1511. @mollymerp can you please take a look?


mollymerp:
@colleenmcginnis I lied to you about the perma link thing -- I had to change the title of the tile in order for it to link properly. 

I also added a little rounding operation to the area to limit the # of digits that print out after the decimal, but feel free to change back if you don't like.

Otherwise LGTM 🚢  on 💚

PR #3103, opened 2016-08-30
Fix exception thrown when requiring window in a Worker

lucaswoj:


jfirebaugh:
`self` is defined in a non-worker context as well, as a reference to `window`. Should we just use `self` everywhere?

lucaswoj:
Didn't know that. Good idea @jfirebaugh. Do you think we should keep the module name as-is?

PR #3104, opened 2016-08-30
Fix "Marker#togglePopup", refactor "Marker"

### LINKED ISSUE #3096: Marker.togglePopup() does not work
mapbox-gl-js version 0.23:

### Steps to Trigger Behavior

 1. Add a Marker and bind a Popup to it with setPopup();
 2. Call Marker.togglePopup();

### Expected Behavior

Popup should be shown.

### Actual Behavior

"Uncaught TypeError: Cannot read property 'stopPropagation' of undefined"

The reason is that Marker.togglePopup does not supply an event to Marker._openPopup() (https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/marker.js#L136) and Marker._openPopup does not check if the event exists (https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/marker.js#L141).

lucaswoj:
This PR fixes #3096 and addresses @jfirebaugh's feedback in https://github.com/mapbox/mapbox-gl-js/pull/3056

cc @jfirebaugh @mollymerp @apkoponen 

## Checklist

 - [x] get a PR review :+1: / :-1:

### DIFF
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -91,12 +91,14 @@ Marker.prototype = {
 
     /**
      * Binds a Popup to the Marker
-     * @param {Popup=} popup an instance of the `Popup` class. If undefined or null, any popup 
+     * @param {Popup=} popup an instance of the `Popup` class. If undefined or null, any popup
      * set on this `Marker` instance is unset
      * @returns {Marker} `this`
      */
 
     setPopup: function (popup) {
+        var this = that;
+
         if (popup == null) {
             this._closePopup();
             delete this._popupHandlersAdded;
@@ -110,7 +112,10 @@ Marker.prototype = {
         if (this._popup && this._lngLat) this._popup.setLngLat(this._lngLat);
 
         if (!this._popupHandlersAdded) {
-            this.getElement().addEventListener('click', this._openPopup.bind(this));
+            this.getElement().addEventListener('click', function(event) {
+                event.stopPropagation();
+                that._openPopup();
+            });
             this._popupHandlersAdded = true;
         }
         return this;
@@ -139,9 +144,6 @@ Marker.prototype = {
     },
 
     _openPopup: function (e) {
-        // prevent event from bubbling up to the map canvas
-        e.stopPropagation();
-
         if (!this._popup || !this._map) return;
 
         if (!this._popup._map) {
@@ -165,4 +167,3 @@ Marker.prototype = {
         DOM.setTransform(this._el, 'translate(' + pos.x + 'px,' + pos.y + 'px)');
     }
 };
-
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -21,15 +21,16 @@ var Popup = require('./popup');
  *   .addTo(map);
  */
 function Marker(element, options) {
-    if (!element) {
-        element = DOM.create('div');
-    }
-    element.classList.add('mapboxgl-marker');
-    this._el = element;
-
     this._offset = Point.convert(options && options.offset || [0, 0]);
 
     this._update = this._update.bind(this);
+    this._onClick = this._onClick.bind(this);
+
+    if (!element) element = DOM.create('div');
+    element.classList.add('mapboxgl-marker');
+    this._element = element;
+
+    this._popup = null;
 }
 
 Marker.prototype = {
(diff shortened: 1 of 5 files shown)


js/ui/marker.js
```diff
@@ -86,36 +88,37 @@ Marker.prototype = {
...
-     * @param {Popup=} popup an instance of the `Popup` class. If undefined or null, any popup 
+     * @param {Popup=} popup an instance of the `Popup` class. If undefined or null, any popup
      * set on this `Marker` instance is unset
      * @returns {Marker} `this`
      */
 
     setPopup: function (popup) {
-        if (popup == null) {
-            this._closePopup();
-            delete this._popupHandlersAdded;
```
lucaswoj: > Instead of tracking state in `_popupHandlersAdded`, call`removeEventListener` when the popup is removed. This will prevent memory leaks and unexpected behavior if the `Marker` continues to be used without a popup. You'll need to save the `this`-bound version of `_openPopup`, say by using `util.bindAll(['_openPopup'], this)`, so that it can be passed to `removeEventListener`. -- https://github.com/mapbox/mapbox-gl-js/pull/3056#discussion_r76332110

js/ui/marker.js
```diff
@@ -59,9 +61,9 @@ Marker.prototype = {
             this._map.off('move', this._update);
             this._map = null;
         }
-        var parent = this._el.parentNode;
-        if (parent) parent.removeChild(this._el);
-        if (this._popup) this._closePopup();
```
lucaswoj: > `this._closePopup` has an internal `if (this._popup)` guard -- either call it unguarded here, or add external guards everywhere it's called and remove the internal guard.

https://github.com/mapbox/mapbox-gl-js/pull/3056#discussion_r76332547

js/ui/marker.js
```diff
@@ -129,40 +132,16 @@ Marker.prototype = {
      * @returns {Marker} `this`
      */
     togglePopup: function () {
-        if (this._popup) {
-            if (this._popup._map) {
-                this._closePopup();
-            } else {
-                this._openPopup();
```
lucaswoj: > This will throw an error because `_openPopup` is expecting an event argument.

https://github.com/mapbox/mapbox-gl-js/pull/3056#discussion_r76332587

jfirebaugh:
This prevents all click events on a marker from being propagated as map `click` events, which is probably not what a user expects. If it really is the desired behavior, a better way to accomplish it is to add the marker element to the control container instead of preventing event propagation.

lucaswoj:
@jfirebaugh Makes sense. Done in the latest commit.

mollymerp:
Thank you @lucaswoj 😳

lucaswoj:
@mollymerp my 😳 for merging #3056 prematurely!

Ready to 🚢 @jfirebaugh @mollymerp?

jfirebaugh:
I really think it will be surprising for clicks on markers not to produce map `click` events. If they don't, then for instance the `closeOnClick` option of `Popup` will not work as expected for the common case of having multiple markers with popups where opening one popup via its marker should close any previously open popup.

lucaswoj:
@jfirebaugh I understand now. I misread https://github.com/mapbox/mapbox-gl-js/pull/3104#issuecomment-243619885. The latest commit ensures that this event is forwarded to `map` as expected.

js/ui/popup.js
```diff
@@ -60,7 +60,14 @@ Popup.prototype = util.inherit(Evented, /** @lends Popup.prototype */{
         this._map = map;
         this._map.on('move', this._update);
         if (this.options.closeOnClick) {
-            this._map.on('click', this._onClickClose);
+            // We attach the `Map#click` listener asynchronously so that it is
+            // possible to open a popup within a `click` event on a child of
+            // the map's container. If we attached the listener syncronously
+            // the popup would close when the click event propogated to `map`.
+            var that = this;
+            setTimeout(function() {
```
jfirebaugh: Is there another way? Ad hoc `setTimeout`s in library code are really dangerous. For example, this will throw if for some reason the popup is removed before the timeout triggers, because `that._map` will be `undefined` after `remove` deletes it.

lucaswoj: Good point. There are some other approaches we can take. I'll push another commit soon.

lucaswoj: The latest commit fixes this problem by attaching the click listener to the map and checking the event's target to determine whether or not the popup should toggle. I did not see any performance degradation.

jfirebaugh: Good solution! It will probably be a performance improvement for lots of markers because adding native event listeners tends to be slower than event delegation.

PR #3105, opened 2016-08-31
Allow drag rotate handler to control pitch

scothis:
Right-click and drag previously only controlled the bearing of the map.
Now it also controls the pitch along with bearing.

Dragging up and down controls the pitch. Dragging left and right controls
the bearing.

The pitch control is opt-out and can be disabled by passing `pitchWithRotate:
false` when initializing the Map. The experience can be demoed at
http://localhost:9966/debug/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ npm-debug.log
 *.sublime-*
 coverage
 .DS_Store
-.mason
 .nyc_output
 /debug/access-token-generated.js
 /bench/index-generated.js
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ npm-debug.log
 *.sublime-*
 coverage
 .DS_Store
+.mason
 .nyc_output
 /debug/access-token-generated.js
 /bench/index-generated.js
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -3,7 +3,7 @@
 var LngLat = require('./lng_lat'),
     Point = require('point-geometry'),
     Coordinate = require('./coordinate'),
-    wrap = require('../util/util').wrap,
+    util = require('../util/util'),
     interp = require('../util/interpolate'),
     TileCoord = require('../source/tile_coord'),
     EXTENT = require('../data/bucket').EXTENT,
@@ -72,7 +72,7 @@ Transform.prototype = {
         return -this.angle / Math.PI * 180;
     },
     set bearing(bearing) {
-        var b = -wrap(bearing, -180, 180) * Math.PI / 180;
+        var b = -util.wrap(bearing, -180, 180) * Math.PI / 180;
         if (this.angle === b) return;
         this._unmodified = false;
         this.angle = b;
@@ -87,7 +87,7 @@ Transform.prototype = {
         return this._pitch / Math.PI * 180;
     },
     set pitch(pitch) {
-        var p = Math.min(60, pitch) / 180 * Math.PI;
+        var p = util.clamp(pitch, 0, 60) / 180 * Math.PI;
         if (this._pitch === p) return;
         this._unmodified = false;
         this._pitch = p;
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,3 @@
-.mason
 _site
 bench
 coverage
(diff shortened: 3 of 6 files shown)


1ec5:
Cool, this is similar to the macOS SDK (which uses Option-drag instead of right-drag for the combined gesture).

samanpwbb:
This is a great idea, would love to see this as part of mapbox gl js.

debug/rotate.html
```diff
@@ -0,0 +1,129 @@
...
+initMap(true);
+
+document.getElementById('pitch-checkbox').onclick = function() {
+    initMap(!!this.checked);
+};
+
+</script>
+
+</body>
+</html>
```
lucaswoj: Can we set `pitchOnRotate` to true on `debug/index.html` and drop this debug page?

js/geo/transform.js
```diff
@@ -87,7 +87,7 @@ Transform.prototype = {
         return this._pitch / Math.PI * 180;
     },
     set pitch(pitch) {
-        var p = Math.min(60, pitch) / 180 * Math.PI;
+        var p = util.clamp(pitch, 0, 60) / 180 * Math.PI;
```
lucaswoj: Was this causing a bug before? What was the bug? Can we add a test for it?

debug/rotate.html
scothis: I created a new page because `pitchOnRotate` can't be mutated at runtime and requires creating a new Map. The other values can be mutated in place.

js/geo/transform.js
scothis: If the pitch goes negative, the Map stops rendering. I'll add a test.

debug/rotate.html
scothis: In place mutations are possible via a private API, I'll fold `debug/rotate.html` into `debug/index.html`

scothis:
@lucaswoj updated the PR:

- rebased on master, fixing merge conflict
- renamed `pitchOnRotate` to `pitchWithRotate` (open to other names)
- merged `debug/rotate.html` into `debug.index.html`
- added a test for clamping pitch in a transition

.gitignore
```diff
@@ -8,6 +8,7 @@ npm-debug.log
 *.sublime-*
 coverage
 .DS_Store
+.mason
```
lucaswoj: Huh. I see we have a similar line in `.npmignore`. Where did this come from? What causes a `.mason` directory to be created?

scothis: I'm not sure what actually creates these files, I assume it's the test suite. I added it to the `.gitignore` because they were on my disk, shouldn't be checked-in and already excluded by `.npmignore` from being published.

jfirebaugh: We don't have anything that should create this anymore; it should be removed from `.npmignore`. It's probably just a stale directory on your machine.

scothis: removed from .npmignore and .gitignore

lucaswoj:
Looks good! 🚢

mourner:
Trying it out, while the ability to change the pitch is great, the new rotation behavior feels very confusing and counter-intuitive to me. :(

jfirebaugh:
Specifically, it's rotations with the cursor in the top half of the map that feel weird to me. I think it would be fine if it acted as if there is a pin in the center of the map and horizontal motion was like applying a rotational force relative to the pin.

scothis:
The old handler measured the angular rotation around the center point of the map. This is what allowed it to work intuitively when dragging the top or bottom of the map. When factoring in a second dimension distinguishing a pitch from a rotate becomes difficult without introducing modifier keys, which are even less intuitive.

The pattern implemented is very common in 3D software. There's a bit of a learning curve, but it becomes natural after a short time.

I'm open to ideas to improve the handler.

1ec5:
In the macOS SDK (as in MapKit), rotations with the cursor in the top half of the map do anchor the rotation at the cursor. However, the cursor is automatically hidden while performing the gesture. That seems to reduce the weirdness. It’s also consistent with how the two-finger rotation gesture works on iOS and Android devices natively.

scothis:
@1ec5 looks like the point of clicking is always the center of rotation. Without being able to detach the center of rotation for the center of the map, my hunch is this will also feel weird. It's worth trying.

1ec5:
Filed #3611 to anchor rotation at the start of the drag gesture.

PR #3115, opened 2016-08-31
Remove 'bin' directory

lucaswoj:
Are these still used? They haven't been touched in years. This looks like code that could live elsewhere. 

cc @mourner @jfirebaugh


jfirebaugh:
👍

PR #3120, opened 2016-08-31
Ensure fills with transparent colors and patterns are rendered as expected

### LINKED ISSUE #3107: 0.23.0 breaks some fill layers that use fill-pattern
**mapbox-gl-js version**: 0.23.0

When using 0.22.1 and earlier, a fill layer with `fill-outline-color` and `fill-color` and `fill-pattern` specified will render the pattern, but not the color or outline (as is expected based on my reading of the style spec).

As of 0.23.0 the same layer will now be rendered with no pattern and no color.

To further complicate things, the issue described above does not surface if you use a hex color instead of hsla(). If you use a hex then the pattern renders as expected.

### Examples
0.22.1 + hsla(), pattern _does_ render: http://jsbin.com/xujowiqeno/1/edit?html,output

0.23.0 + hsla(), pattern _does not_ render: http://jsbin.com/zifoxutavo/1/edit?html,output

0.23.0 + hex(), pattern _does_ render: http://jsbin.com/mozefaxura/1/edit?html,output

### Expected Behavior
Pattern renders.

### Actual Behavior
Pattern does not render.

lucaswoj:
This PR fixes #3107 by making `FillStyleLayer#isHidden` aware of `fill-pattern`. 

## Launch Checklist

 - [x] write tests for all new functionality https://github.com/mapbox/mapbox-gl-test-suite/pull/140
 - [ ] get a PR review

### DIFF
--- a/js/style/style_layer.js
+++ b/js/style/style_layer.js
@@ -218,8 +218,6 @@ StyleLayer.prototype = util.inherit(Evented, {
         if (this.minzoom && zoom < this.minzoom) return true;
         if (this.maxzoom && zoom >= this.maxzoom) return true;
         if (this.layout['visibility'] === 'none') return true;
-        if (this.isPaintValueFeatureConstant(this.type + '-opacity') && this.paint[this.type + '-opacity'] === 0) return true;
-        if (this.isPaintValueFeatureConstant(this.type + '-color') && this.paint[this.type + '-color'] && this.paint[this.type + '-color'][3] === 0) return true;
 
         return false;
     },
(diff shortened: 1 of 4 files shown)


lucaswoj:
cc @dpieri @jfirebaugh @mourner @mollymerp

jfirebaugh:
LGTM, although I question whether `isHidden` should be doing any checking of any properties other than `visibility`.

lucaswoj:
> LGTM, although I question whether `isHidden` should be checking any properties other than `visibility`.

That's a fair point. I'd be happy to refactor to make it so. Should we go for it?

jfirebaugh:
If it doesn't regress tests or benchmarks, I say go for it.

lucaswoj:
The latest commit removes all opacity tests from the `StyleLayer#isHidden` method. This is unlikely to cause performance regressions in core styles because we don't often have 0 opacity layers in core styles. It's possible a 3rd party style may see a perf regression due to this change.

Issue #3119, opened 2016-08-31
Fix bug causing pitched camera to animate on screen coordinates rather than geographic coordinates


yeldarby:
**mapbox-gl-js version**: v0.23.0

I've been trying to pin down issue #3112 (I think I found the root of it; going to post a comment over there with my findings in a moment) but encountered another bug in the process.

### Steps to Trigger Behavior
 1. Set a map pitch to 60
 2. Use `panBy` to move the map in the vertical coordinate of the viewport (positive delta y)
 3. Use `panBy` in the opposite direction (negative delta y)

### Expected Behavior
The map should return to where it started like it does when the map is not pitched

### Actual Behavior
The map moves much more in one direction than the other and hops along

### Demo
Here's a JS Fiddle demonstrating the issue. Click the map to toggle between pitch of 0 (working as expected) and pitch of 60 (hopping):
http://jsfiddle.net/qq3pqub2/3/

### Explanation
`panBy` is defined to move by a certain number of pixels. When the map is pitched the number pixels in the viewport is not the same as the number of pixels in the map. When these get projected back and forth they are producing unexpected behavior.

### Proposed solution
Change `panBy` to pan by an amount of degrees of latitude/longitude or a distance rather than by number of pixels. An alternative would be to pretend the map is not pitched and pan by that amount. Changing the pitch of the map shouldn't affect how far it pans.

indus:
I think ```panBy``` using viewport coordinates is preferable over using latitude/longitude and that the behaviour is correct. When you use lat/lng you would face the trouble of a rotated map pretty soon (and rotation is even more common than tilting).
Why not use ´´´easeTo´´´ with subtracted coordinates?

yeldarby:
Yeah, I agree. In my PR I ended up goIng with option 2. It pans by map pixels rather than viewport pixels.

panBy should be a reversible function. eg this operation should leave the map in the same place it started:

```
panBy([x, y]);
panBy([-x, -y]);
```

lucaswoj:
I think this is a symptom of the same problem we're tracking in #3112

PR #3130, opened 2016-09-01
Rework Camera#easeTo to address problems with pitch

### LINKED ISSUE #3112: Ensure camera animations interpolate geographic coordinates rather than screen coordinates
**mapbox-gl-js version**: v0.23.0
### Steps to Trigger Behavior
1. Set a map pitch to 60
2. Pan the map to a location using a function that has easing eg `panTo`
3. Set a custom linear ease function as an option `easing: function(t) { return t; }`
### Expected Behavior

The map to pan to the location linearly like it does with no pitch 
### Actual Behavior

The map eases to the location instead
### Demo

Here's a JS Fiddle demonstrating the issue. Click the map to toggle between pitch of 0 (working as expected) and pitch of 60 (eases instead of panning linearly):
http://jsfiddle.net/qq3pqub2/2/

yeldarby:
## Description

Fixes #3112 and #3119 

Previously, when the map was pitched panning caused unexpected easing and relative adjustments. Pitching the map should not affect easing or the final location of `panBy`.

This is fixed by calculating the finish state of the map prior to the transition beginning and interpolating from start state to end state accordingly.

One catch: when zooming you don't want to zoom and position to interpolate together since zoom affects the distance between start and end on the screen (so there's an adjustment to take that into account).
## Launch Checklist
- [ ] post benchmark scores - not sure how to do this?
- [ ] get a PR review


debug/linear-ease.html
```diff
@@ -0,0 +1,138 @@
...
+    	map.setPitch(60);
+    } else {
+    	map.setPitch(0);
+    }
+  });
+});
+</script>
+<div id="centerpoint"><div id="dot"></div></div>
+</body>
+</html>
```
lucaswoj: It makes sense that you built a debug page to use while developing this feature. 

Because repository size, simplicity, and stability is important to us, we would prefer not to merge this file into `master`. Can you please remove this file from the PR?

js/ui/camera.js
```diff
@@ -408,13 +409,27 @@ util.extend(Camera.prototype, /** @lends Map.prototype */{
...
+            fromLocation,
+            endTransform;
+
+        // Setup a Transform representing the state of the map at the end of the transition
+        endTransform = new Transform();
+        endTransform.center = tr.center;
+        endTransform.resize(tr.width, tr.height);
+        endTransform.zoom = zoom;
+        endTransform.bearing = bearing;
+        endTransform.pitch = 0; // fixes #3119 by pretending the map is not pitched; use endTransform.pitch = pitch to revert to the old behavior
```
lucaswoj: Oh! I have another idea that'll require much less work:

 - factor `Trasform#pointLocation` into a standalone utility function with the interface 

```js
function pointLocation(
    cameraOptions: CameraOptions,
    mapSize: {width: number, height: number},
    point: Point
)
```

lucaswoj:

Issue #3034, opened 2016-08-18
Refactor "Worker", "Dispatcher", "Actor" into a general purpose background processing utility


lucaswoj:
If we can refactor `Worker`, `Dispatcher`, and `Actor` into a stateless, general purpose background processing utility (or find such an existing utility), GL JS's architecture will simplify nicely, especially around custom sources. We will remove a lot of unnecessary coupling and pave the way for new background processing features like #1042.

anandthakker:
I think "stateless" is going to be a tough challenge here, because: (a) efficient communication of large amounts of data to/from worker threads is limited to ArrayBuffers (is this true?); (b) Parsing ArrayBuffers (i.e. VT pbf data) is moderately expensive, so that it's desirable to cache the parsed representation for reuse, e.g. when reloading a tile.

@lucaswoj do you have a vision for how we might manage this?

lucaswoj:
My vision (originally @ansis' vision) is to use `StructArray` for all persistent data that needs to be used by the workers.

anandthakker:
Oh, cool.  That makes sense... and definitely seems more tractable now than it did at midnight!

lucaswoj:
Did a quick scan and couldn't find any existing `WebGL` libraries that might give us the functionality we need.

lucaswoj:
I'm not necessarily intent on this being stateless, so long as state is somehow associated with the caller's scope rather than managed globally within the worker. The thinner the abstraction the better.

```js
// test.js
var Worker2 = require('../util/worker2');
var worker = new Worker2(require('./test_worker'));

worker.emit('setBar', 'foo');
worker.once('setBarResponse', function() {
    worker.emit('getBar');
    worker.once('getBarResponse', function(bar) {
         // bar => 'foo'
    });
});

// test_worker.js
module.exports = function(bus) {
    var bar;
    bus.on('setBar', function(nextBar) {
         bar = nextBar;
         bus.emit('setBarResponse');
    });

    bus.on('getBar', function(getBar) {
        bus.emit('getBarResponse', bar);
    });
}
```

This is almost exactly the `WebWorker` API. They key difference is that we have a `m:n` ratio of `Worker2`s to `WebWorker`s, allowing us to create arbitrarily many `Worker2`s without  any perf consequences.

mb12:
@lucaswoj Can you please make sure that the following issue is also handled as part of Worker refactoring? It will have a huge  impact on  productivity.

[...]

PR #3133, opened 2016-09-01
Fix bug causing worker-initiated messages to be broadcast to every map instance

### LINKED ISSUE #3086: Fix bug causing garbled text rendering when multiple maps are rendered on the page
**0.23.0**

 I've upgraded from 0.18.0.

This is a larger GWT application that uses JsInterop, so it's difficult for me to provide code to run. Before I isolate and try to reproduce this error with plain JS, it would be good to know if this is a known issue.

Ultimately all I do is use several instance of Map on the same DOM, and attach/detach them frequently. When the first instance of Map renders, it looks fine. When I detach the first instance and create and attach a second (and probably other) instance(s) of a Map, this second instance has broken rendering of text labels.

See the attached screenshot.

The code is unchanged except for some necessary API adjustments with handling GeoJSON sources.
screen shot 2016-08-27 at 22 13 24

anandthakker:
Closes #3086

### DIFF
--- a/js/source/worker.js
+++ b/js/source/worker.js
@@ -148,7 +148,16 @@ util.extend(Worker.prototype, {
                 getLayers: function () { return this.layers[mapId]; }.bind(this),
                 getLayerFamilies: function () { return this.layerFamilies[mapId]; }.bind(this)
             };
-            this.workerSources[mapId][type] = new this.workerSourceTypes[type](this.actor, layers);
+
+            // use a wrapped actor so that we can attach a target mapId param
+            // to any messages invoked by the WorkerSource
+            var actor = {
+                send: function (type, data, callback, buffers) {
+                    this.actor.send(type, data, callback, buffers, mapId);
+                }.bind(this)
+            };
+
+            this.workerSources[mapId][type] = new this.workerSourceTypes[type](actor, layers);
         }
 
         return this.workerSources[mapId][type];
--- a/js/util/actor.js
+++ b/js/util/actor.js
@@ -28,6 +28,12 @@ Actor.prototype.receive = function(message) {
         id = data.id,
         callback;
 
+    // if we're on the main thread handling a message from the worker, make
+    // sure that we only handle messages directed at this actor's the map
+    // instance
+    if (data.targetMapId && this.mapId !== data.targetMapId)
+        return;
+
     if (data.type === '<response>') {
         callback = this.callbacks[data.id];
         delete this.callbacks[data.id];
@@ -55,10 +61,16 @@ Actor.prototype.receive = function(message) {
     }
 };
 
-Actor.prototype.send = function(type, data, callback, buffers) {
+Actor.prototype.send = function(type, data, callback, buffers, targetMapId) {
     var id = callback ? this.mapId + ':' + this.callbackID++ : null;
     if (callback) this.callbacks[id] = callback;
-    this.postMessage({ mapId: this.mapId, type: type, id: String(id), data: data }, buffers);
+    this.postMessage({
+        targetMapId: targetMapId,
+        mapId: this.mapId,
+        type: type,
+        id: String(id),
+        data: data
+    }, buffers);
 };
 
 /**
--- a/js/util/actor.js
+++ b/js/util/actor.js
@@ -23,14 +23,34 @@ function Actor(target, parent, mapId) {
     this.target.addEventListener('message', this.receive, false);
 }
 
+/**
+ * Sends a message from a main-thread map to a Worker or from a Worker back to
+ * a main-thread map instance.
+ *
+ * @param {string} type The name of the target method to invoke or '[source-type].name' for a method on a WorkerSource.
+ * @param {object} data
+ * @param {Function} [callback]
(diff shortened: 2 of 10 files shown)


anandthakker:
@jfirebaugh This should fix it.  Confirmed it with a quick check using jsfiddle example linked in the issue.

js/source/worker.js
```diff
@@ -148,7 +148,16 @@ util.extend(Worker.prototype, {
...
             };
-            this.workerSources[mapId][type] = new this.workerSourceTypes[type](this.actor, layers);
+
+            // use a wrapped actor so that we can attach a target mapId param
+            // to any messages invoked by the WorkerSource
+            var actor = {
+                send: function (type, data, callback, buffers) {
+                    this.actor.send(type, data, callback, buffers, mapId);
+                }.bind(this)
+            };
```
lucaswoj: It feels like we're fighting our own architecture with this fake `Actor`. Can you think of a way to refactor & simplify? `Worker` and `Actor` are so tightly coupled it might make sense to merge them.

js/util/actor.js
```diff
@@ -28,6 +28,12 @@ Actor.prototype.receive = function(message) {
         id = data.id,
         callback;
 
+    // if we're on the main thread handling a message from the worker, make
+    // sure that we only handle messages directed at this actor's the map
+    // instance
+    if (data.targetMapId && this.mapId !== data.targetMapId)
+        return;
```
lucaswoj: Do you think there's a way you can make this code more self-documenting such that the comment is unnecessary?

js/util/actor.js
```diff
@@ -55,10 +61,16 @@ Actor.prototype.receive = function(message) {
...
 };
 
-Actor.prototype.send = function(type, data, callback, buffers) {
+Actor.prototype.send = function(type, data, callback, buffers, targetMapId) {
     var id = callback ? this.mapId + ':' + this.callbackID++ : null;
     if (callback) this.callbacks[id] = callback;
-    this.postMessage({ mapId: this.mapId, type: type, id: String(id), data: data }, buffers);
+    this.postMessage({
+        targetMapId: targetMapId,
+        mapId: this.mapId,
```
lucaswoj: Why do we need to transfer both `mapId` and a `targetMapId`? Doesn't [this guard](https://github.com/mapbox/mapbox-gl-js/pull/3133/files#diff-07866509a19229697416fd9d0d2cae10R34) ensure they are equal?

js/source/worker.js
anandthakker: I think merging Worker and Actor is probably beyond the scope of this bug fix, but I certainly agree your broader point over in #3034.

I was trying to keep the scope of this change as limited as possible, but maybe one way to simplify this a bit would be to replace WorkerSource's (and WorkerTile#parse's) `actor: Actor` parameter with just a `send: (type, data, callback)=>void` parameter.

js/util/actor.js
anandthakker: There are really two different things here:
 - For messages initiated from the main thread, `mapId` identifies the _sending_ map instance, which is really for the purpose of enabling the worker to track state related to different instances.
 - For messages initiated from the worker, `targetMapId` identifies the _target_ map instance.

It's true that since these two properties are used in totally disjoint cases, we could simply use `mapId` for both.... but following along your point above re: making the code better self-documenting, I think it's a bit clearer not to overload the same property for two different needs (especially since they're similar, but subtly different)

js/util/actor.js
anandthakker: Yep. Will do

jfirebaugh:
Another issue with #2952 is that it can leak WebWorker `message` event bindings: the binding added in the `Actor` constructor is never removed. It looks like if the last `Map` is `remove`d than the workers will get shut down, but as long as there is at least one `Map`, then all the `Actors` ever created will stay live.

Can you add a fix for that to this PR as well?

jfirebaugh:
Another issue: the blob URLs that webworkify creates aren't revoked; if you repeatedly create and remove maps, so that the worker pool is drained and refilled, they leak.

anandthakker:
> Another issue with #2952 is that it can leak WebWorker message event bindings: the binding added in the Actor constructor is never removed. It looks like if the last Map is removed than the workers will get shut down, but as long as there is at least one Map, then all the Actors ever created will stay live.

I think this predates 2952, but yep I can add a fix for this.

anandthakker:
> I think this predates 2952

Ah, actually it probably _doesn't_ predate 2952, since it's probably a pretty safe assumption that, before the worker pool, the `message` listeners were being removed in `WebWorker#terminate`.   My mistake.

anandthakker:
@jfirebaugh included fixes to dispose of event listeners and revoke object URLs.

@lucaswoj also tried to clarify the code in `Actor` to avoid the need for an inline comment; but at the same time, I thought it did make sense to include standard jsdoc for `Actor#send`.

lucaswoj:
This is ready to 🚢. We need to fix this bug before the next release. I'm also keen on simplify our worker architecture 😄

Issue #3121, opened 2016-08-31
Remove remaining "browser" modules


lucaswoj:
Continuing the work started in #3106:
- [x] try to eliminate `canvas` module #3135
- [x] try to eliminate `ajax` module #3134 
- [ ] try to eliminate `web_worker` module
- [ ] eliminate all utilities that provide no abstraction over functionality provided by the DOM
- [ ] move window mock from `js` directory to `test` directory

jfirebaugh:
I think we've captured the bulk of the benefits here.

PR #3134, opened 2016-09-01
Remove fake "ajax" module

lucaswoj:
This PR removes the fake `js/util/ajax` module, instead leveraging Sinon's fake [`XMLHttpRequest`](http://sinonjs.org/docs/#server) functionality. Because of this change

 - `js/util/browser/ajax.js` is fully tested (and moved to `js/util/ajax.js`)
 - there are no 404 error messages during unit testing
 - there is no need to start / stop real HTTP servers during testing
 - we have the ability to create cheap mock HTTP servers using Sinon

cc @jfirebaugh @mollymerp 

ref https://github.com/mapbox/mapbox-gl-js/issues/3121

## Launch Checklist

 - [x] diagnose test-suite hangs 
 - [x] ensure unit test error messages are suppressed 
 - [ ] get a PR review


lucaswoj:
This is ready for 👀 @jfirebaugh

test/suite_implementation.js
```diff
@@ -12,6 +17,7 @@ module.exports = function(style, options, _callback) {
         }
     }
 
+    window.XMLHttpRequest = XMLHttpRequest;
```
jfirebaugh: Is there a reason the default jsdom XHR isn't suitable? Can we document it with a comment?

test/js/source/vector_tile_source.test.js
```diff
@@ -1,13 +1,9 @@
 'use strict';
 
 var test = require('tap').test;
-var st = require('st');
```
jfirebaugh: Should be able to remove this from devDependencies as well.

test/suite_implementation.js
jfirebaugh: Oh, I see, it's overwriting the sinon fake XHR. Can we use jsdom XHR by default and only substitute sinon for the unit tests?

lucaswoj: The built-in jsdom XHR doesn't work but I don't fully understand why. I'll dig in now.

lucaswoj:
I listed my 4 reasons "why" in the original PR. Stubing 2 of the 4 methods when we run `test-suite` is an improvement over never testing the `ajax` module 😄.

mourner:
@lucaswoj no, I was referring to `st` module removal. It's used in npm scripts for running a debug/bench server.

lucaswoj:
Oops! You're right. I didn't think of that. Restoring shortly.

mourner:
@lucaswoj do you know why the test suite runs ~4 minutes slower in this branch, or is this some kind of a random Circle deviation?

lucaswoj:
@mourner I haven't looked into it yet. There were a few layers of abstraction added into the networking stack. Perhaps we should mock out the entire AJAX module as before during the test suite run.

jfirebaugh:
Looking at recent successful builds, the length of this one doesn't seem to be too far outside the normal variance of test times.

Good to merge from my perspective, and a much-welcome cleanup.

mourner:
@jfirebaugh it looks quite far outside to me — consistently 16-17m for test.sh (after rerunning it a few times) where all other recent builds are 11-13m.

lucaswoj:
The tests themselves aren't taking longer. However there is some background work preventing the process from exiting once the tests finish.

lucaswoj:
JSDom's XHR library is causing the process to hang. 

The most effective use of our time is to stub out all methods on the `ajax` module when running the render tests. We may want to spend time debugging JSDom and submitting an upstream patch in the future.

lucaswoj:
Ready to 🚢?

jfirebaugh:
🚢 

I have some pending ajax.js changes that will as a side-effect make stubbing it out cleaner.

PR #3135, opened 2016-09-02
Replace canvas.js with a minimal shim for substituting headless-gl

jfirebaugh:
Refs #3121. I found this was necessary for the `Popup` tests I want to write.

## Launch Checklist

 - [ ] get a PR review


lucaswoj:
I was just working on the same thing and came up with another solution. Take a look at the [`fake-canvas`](https://github.com/mapbox/mapbox-gl-js/compare/fake-canvas) branch and let me know what you think.

My intent is to eventually capture all browser mocking in the `window` module.

jfirebaugh:
Ah, good idea. I'll merge that approach in here. I still want to merge canvas.js into `Map` -- it only existed as a seam for replacing DOM-dependent code, and at this point the division with the DOM manipulation that `Map` does is arbitrary.

jfirebaugh:
@lucaswoj 🚢 ?

lucaswoj:
Looks great! 🚢

PR #3142, opened 2016-09-04
Fully switch to Nyc instead of Istanbul

mourner:
Instead of an Istanbul/Nyc mix, switches to Nyc fully for test coverage. Reasons:

1. Ensures all coverage code is run with the same tool/version (instead of a mix of Tap Nyc subdependency + standalone Istanbul).
2. Istanbul NPM module is getting deprecated soon — it's been rewritten as a set of independent modules in https://github.com/istanbuljs that Nyc now uses under the hood.
3. We'll be able to get the full benefit once my [performance patch](https://github.com/istanbuljs/istanbul-lib-instrument/pull/22) lands into Nyc.

cc @lucaswoj @jfirebaugh

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -19,12 +19,8 @@ npm run lint
 npm run build-min
 npm run build-dev
 
-# run unit tests
-tap --reporter dot --coverage --no-coverage-report test/js test/build/webpack.test.js
-
-# run render tests
-istanbul cover --dir .nyc_output --include-pid --report none --print none test/render.test.js
-istanbul cover --dir .nyc_output --include-pid --report none --print none test/query.test.js
+# run unit, render & query tests with coverage
+npm run test-cov
 
 # send coverage report to coveralls
 nyc report --reporter=lcov
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "gl": "^4.0.1",
     "handlebars": "4.0.5",
     "highlight.js": "9.3.0",
-    "istanbul": "^0.4.2",
     "jsdom": "^9.4.2",
     "json-loader": "^0.5.4",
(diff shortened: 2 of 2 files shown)


jfirebaugh:
The build is green but it doesn't look like it ran properly. There are many errors of the form:

```
Transformation error for /Users/distiller/mapbox-gl-js/js/util/util.js ; return original code
Cannot find module 'istanbul-lib-instrument'
```

And at then end:

```
=============================== Coverage summary ===============================
Statements   : Unknown% ( 0/0 )
Branches     : Unknown% ( 0/0 )
Functions    : Unknown% ( 0/0 )
Lines        : Unknown% ( 0/0 )
================================================================================
[error] "2016-09-04T10:59:23.279Z"  'error from lcovParse: ' 'Failed to parse string'
[error] "2016-09-04T10:59:23.281Z"  'input: ' ''
[error] "2016-09-04T10:59:23.282Z"  'error from convertLcovToCoveralls'

/Users/distiller/mapbox-gl-js/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Failed to parse string
```

mourner:
@jfirebaugh ah yeah, the last commit here was built properly (see https://circleci.com/gh/mapbox/mapbox-gl-js/4856), but then I pushed some other temp commits that installed a broken forked version (which I force-removed afterwards) and now I probably need to flush the Circle cache.

mourner:
@jfirebaugh yep, just reran the tests without cache and it's fine again. Note that coverage decreased by 1% probably because of some changes to what nyc instruments. E.g. function declarations are no longer tracked as statements.

lucaswoj:
This is great! I love how much this simplifies our build system, CI scripts, and architecture. Thanks @mourner.

PR #3154, opened 2016-09-06
fix Popup docs so that the offset option is displayed, add offset option to example

mollymerp:
documentation wasn't being generated bc the type wasn't recognize (I think)

screen shot 2016-09-06 at 3 21 04 pm

cc @lucaswoj @jfirebaugh


lucaswoj:
Thanks @mollymerp!

PR #3155, opened 2016-09-06
Fix #2827 and #2828 (fill zoom-and-property-fns)

### LINKED ISSUE #2827: Fix unexpected results for zoom-and-property function
```
        "fill-color": {
          "property": "property",
          "stops": [
            [{"zoom": 0, "value": 0}, "rgba(255,0,0,1)"],
            [{"zoom": 0, "value": 1}, "rgba(0,255,0,1)"],
            [{"zoom": 0, "value": 2}, "rgba(0,0,255,1)"],
            [{"zoom": 1, "value": 0}, "rgba(255,0,0,0)"],
            [{"zoom": 1, "value": 1}, "rgba(0,255,0,0)"],
            [{"zoom": 1, "value": 2}, "rgba(0,0,255,0)"]
          ]
        }
```

At z=0.1, I'd expect these fills to be rendered with opacity 0.9. Instead they are rendered with 0 opacity (fully transparent).

### LINKED ISSUE #2828: Fix unexpected results for fill-opacity property function

lbud:
Fixes #2827, #2828. Found while adding fill-extrude properties this bug wherein `getPaintValueStopZoomLevels` is being called with an array of its name (i.e. `getPaintValueStopZoomLevels(["fill-color"])`), the nature of `arguments`, and therefore always returning undefined.

fixes #2827 
fixes #2828

cc @lucaswoj

### DIFF
--- a/js/style/style_layer/fill_style_layer.js
+++ b/js/style/style_layer/fill_style_layer.js
@@ -21,7 +21,7 @@ FillStyleLayer.prototype = util.inherit(StyleLayer, {
         if (name === 'fill-outline-color' && this.getPaintProperty('fill-outline-color') === undefined) {
             return StyleLayer.prototype.getPaintValueStopZoomLevels.call(this, 'fill-color');
         } else {
-            return StyleLayer.prototype.getPaintValueStopZoomLevels.call(this, arguments);
+            return StyleLayer.prototype.getPaintValueStopZoomLevels.call(this, name);
         }
     },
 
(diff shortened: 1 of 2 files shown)


lucaswoj:
Good find! Thank you @lbud. 🚢

Issue #2977, opened 2016-08-10
Support imperial units in scale control


lucaswoj:
The scale control should support both metric and imperial units. We can switch between them manually (i.e. with a constructor option to `Scale`) and/or automatically [based on the user's locale](http://stackoverflow.com/a/18684014/5942199).  

ref https://github.com/mapbox/mapbox-gl-js/pull/2940#issuecomment-238882361

cc @mourner @tristen @amishas157

PR #3160, opened 2016-09-08
Support Imperial unit

amishas157:
Added the support for Imperial unit in scale control, following from Leaflet's implementation.
Reference : https://github.com/mapbox/mapbox-gl-js/issues/2977

cc @lucaswoj @mourner

### DIFF
--- a/js/ui/control/scale.js
+++ b/js/ui/control/scale.js
@@ -13,9 +13,11 @@ module.exports = Scale;
  * @param {Object} [options]
  * @param {string} [options.position='bottom-left'] A string indicating the control's position on the map. Options are `'top-right'`, `'top-left'`, `'bottom-right'`, and `'bottom-left'`.
  * @param {number} [options.maxWidth='150'] The maximum length of the scale control in pixels.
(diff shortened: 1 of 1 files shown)


js/ui/control/scale.js
```diff
@@ -46,17 +48,33 @@ function updateScale(map, scale, options) {
...
+            var feet = getRoundNum(maxFeet);
+            ratio = feet / maxFeet;
+            scale.style.width = maxWidth * ratio + 'px';
+            scale.innerHTML = feet + ' ft';
+        }
+    } else {
+        var meters = getRoundNum(maxMeters);
+        ratio = meters / maxMeters;
+        scale.style.width = maxWidth * ratio + 'px';
+        scale.innerHTML = meters < 1000 ? meters + ' m' : (meters / 1000) + ' km';
```
mourner: There is some very similar code in three places here. Can we reduce the repetition (e.g. by cutting out some function)?

DannyDelott:
Can't wait for this to land, the scale looks great!

PR #3162, opened 2016-09-08
Make our CI tests fast again

### LINKED ISSUE #2366: Speed up render tests
Render tests take a big portion of our test time (~9 min), especially after we switched on test coverage with `istanbul`. We could try to significantly speed this up:

1. Run it in two stages — the first generates all actual images, the second runs pixelmatch-powered diffing on all generated images. This allows us to run `istanbul` only on the first stage, speeding up the second one significantly.
2. Run both image generation and diffing in parallel batches with separate node processes (`child_process.fork`), similar to how `tile-reduce` parallelizes work. Since tests are independent, we can get a huge boost this way.

cc @lucaswoj @jfirebaugh

mourner:
Closes #2366. A temporary change that gets our total CI build under 10m again (with test.sh taking under 6m, compared to 12-14m previously). We can either merge this now and enjoy smaller times, or wait a few weeks until the change propagates to `nyc@latest`. I think there's no harm temporarily switching to a fork — the only change compared to nyc master is a bumped `istanbul-lib-instrument` version.

cc @lucaswoj @jfirebaugh

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "memory-fs": "^0.3.0",
     "minifyify": "^7.0.1",
     "npm-run-all": "^3.0.0",
-    "nyc": "^8.1.0",
+    "nyc": "mourner/nyc#fast",
     "proxyquire": "^1.7.9",
     "remark": "4.2.2",
     "remark-html": "3.0.0",


jfirebaugh:
👍

lucaswoj:
:heart:

PR #3163, opened 2016-09-08
Bump Circle to Node v6 and NPM v3

mourner:
Fiddling further with build optimization, it turns out that if we switch to Node v6 (and NPM v3), it cuts down total Circle build running time to 6m20s (!!!), down from 9m40s in #3162. This comes from several things:

- faster cache restore because of flat node_modules (4x less files)
- faster `npm install` after restored cache
- faster cache save (4x less files)
- faster test execution because of newer V8 — 5m vs 6m

I think the benefits (cutting down another 3m20s) are worth the risk of switching, and those versions will [become Active LTS in 3 weeks](https://github.com/nodejs/LTS) anyway. Also, NPM 3 seems to be much more stable than the last time we had problems with it in Circle.

cc @jfirebaugh @lucaswoj

### DIFF
--- a/ci/dependencies.sh
+++ b/ci/dependencies.sh
@@ -8,9 +8,9 @@ fi
 source ./nvm/nvm.sh && nvm install ${NODE_VERSION}
 
 if [ "$CIRCLE_BRANCH" == "master" ]; then
-    source ./nvm/nvm.sh && nvm use ${NODE_VERSION} && npm update
+    source ./nvm/nvm.sh && nvm use ${NODE_VERSION} && npm update && npm ls
 else
-    source ./nvm/nvm.sh && nvm use ${NODE_VERSION} && npm install
+    source ./nvm/nvm.sh && nvm use ${NODE_VERSION} && npm install && npm ls
 fi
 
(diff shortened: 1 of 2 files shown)


jfirebaugh:
I'm pretty wary of npm 3 still. [This bug](https://github.com/npm/npm/issues/11766), which I hit all the time whenever I tried it, is still unfixed.

mourner:
@jfirebaugh one good illustration of NPM2 being a big problem for the current JS ecosystem is the `documentation` dependency. Due to a ton of babel-related dependencies it has, installing it with NPM2 adds 360MB disk usage in node_modules with ~80k files, while NPM3 or pnpm makes it 45MB with ~11k files. cc @tmcw 

What do you think of taking NPM3 for a spin but doing 2 consecutive `npm install`s (which reportedly fix the issue above)?

jfirebaugh:
Ok, let's give NPM 3 another shot.

mourner:
@jfirebaugh added `npm ls` (doesn't seem to affect times much, finished in 6:26). Leaving this for you to approve/merge.

Issue #3153, opened 2016-09-06
Improve startup performance of workers


jfirebaugh:
One of the main determinants of time-to-first-render (TTFR) is how fast the workers are able to boot and begin processing tile data. On my laptop, the earliest I see tile requests getting made by workers is at about the 2000ms mark. We should try to improve this.

Ideas:
- Reduce the amount of code included in the worker blob (https://github.com/substack/webworkify/issues/12 may be relevant). (My guess is this is the most important thing we can do to improve TTFR.)
- Reduce the overhead of creating the blob URLs for the worker
  - Create only one, not one per worker
  - Start loading them earlier (at JS load time, rather than the first time a `Style` is created)
- Create only one worker at boot, and additional workers when idle
  - In my tests, booting with only a single worker (vs 3) reduces TTFR by several hundred milliseconds
- Reduce the overhead of transferring layers to the worker
  - Manual JSON.stringify/parse?
  - Reload from style URL on worker?
  - Custom `StructArray` format?
- Reduce style validation overhead (#3149, #3151, #3152)
- Benchmark removing *.tiles.mapbox.com DNS sharding altogether (use api.mapbox.com for everything). This would avoid additional DNS and SSL startup costs.

mourner:
> Reduce the amount of code included in the worker blob

I found a way to reduce it to only the things it needs — see PR https://github.com/substack/webworkify/pull/30. It reduces the size from 1.12MB to 467KB, although I'm not sure whether it actually affects time-to-first-render that much — can you check @jfirebaugh?

> Create only one blob, not one per worker

It seems to be too cheap to bother — the whole workify process up to blob URL creation takes just a few milliseconds in my tests.

jfirebaugh:
Before:

![image](https://cloud.githubusercontent.com/assets/98601/18321446/dc015a92-74e2-11e6-9128-d2fcb54a2ac5.png)

After:

![image](https://cloud.githubusercontent.com/assets/98601/18321457/e67a28e6-74e2-11e6-9cf4-1a08260b5f04.png)

It definitely helps blob creation and worker boot time, although the overall effect on TTFR is only a few hundred milliseconds. I think much of the benefit is being lost due to poor parallelization, and we can recoup it with improved main thread scheduling (boot workers as early as possible, start style XHR as early as possible, reduce validation overhead).

[...]

PR #3172, opened 2016-09-12
Stringify layers for faster worker transfer

mourner:
Ref #3153, not the cleanest change but appears to reduce time-to-fist-render by ~150ms. 

Couldn’t run benchmarks yet because they seem to be broken at the moment (only run on master).

cc @jfirebaugh @lucaswoj

### DIFF
--- a/bench/benchmarks/buffer.js
+++ b/bench/benchmarks/buffer.js
@@ -186,7 +186,7 @@ var createLayerFamiliesCacheValue;
 function createLayerFamilies(layers) {
     if (layers !== createLayerFamiliesCacheKey) {
         var worker = new Worker({addEventListener: function() {} });
-        worker['set layers'](0, layers);
+        worker['set layers'](0, JSON.stringify(layers));
 
         createLayerFamiliesCacheKey = layers;
         createLayerFamiliesCacheValue = worker.layerFamilies[0];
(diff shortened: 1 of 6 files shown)


mourner:
# Benchmarks

## buffer

**master 1f8351d:** 916 ms
**stringify-layers 6ad9724:** 943 ms

## frame-duration

**master 1f8351d:** 10.6 ms, 15% > 16ms
**stringify-layers 6ad9724:** 10.5 ms, 15% > 16ms

lucaswoj:
LGTM 

 > appears to reduce time-to-first-render by ~150ms

How many ms is the time-to-first-render now? Does the map-load bench capture this #?

mourner:
@lucaswoj looking at the timeline, it's still around 2000ms, but I haven't captured it exactly yet, and the benchmarks don't capture this too — I thought about adding one, although this should be a separate PR.

anandthakker:
> Does the map-load bench capture this metric? Could we write a benchmark that does capture this metric?

@lucaswoj @mourner I think it probably makes sense to just change the map-load bench into a first-render, since the latter is what matters most?

mourner:
Benchmarks barely show a ~50-100ms difference, so it might not be worth the added hackiness.

anandthakker:
@mourner If this change speeds up the `set layers` and `update layers` actions, then it could also affect the performance of style mutation methods like `map.setFilter()`, etc., right?  For those cases, I think even a few ms difference can matter.

lucaswoj:
Closing as stale. We can come back to this at any time.

PR #3173, opened 2016-09-12
Fix benchmarks

mourner:
cc @lucaswoj


lucaswoj:
Thanks. I didn't catch this rename.

PR #3174, opened 2016-09-12
Use package-json-versionify to trim package.json in min bundle

etpinard:
This PR adds the [`package-json.versionify`](https://github.com/nolanlawson/package-json-versionify) transform to the `npm run build-min` tasks and add one `test/build/min.test.js` test case.

This change reduces the size of the resulting `dist/mapbox-gl.js` bundle by `5K` (and even more on `npm` consumer builds where the README.md data is pasted into the package.json).


lucaswoj:
Looks good to me! Thank you @etpinard 🙇

PR #3179, opened 2016-09-13
Add support for new line in text

### LINKED ISSUE #955: Support newlines in symbol text fields
If a text field in a vector tile contains a newline character, it is ignored for rendering. For most geodata this would be uncommon, but I think GL should handle this like Mapnik and render the appropriate line breaks.

screenshot from 2015-01-26 17 17 24

mourner:
Closes #955. I'm not sure this is the cleanest way to fix but it seems to work:

image

 - [x] check benchmarks
 - [x] get a PR review @ansis @lucaswoj 

cc @ajashton @danpat


mb12:
@mourner A less intrusive way to fix this would be to handle it the way text transform is implemented.  Basically strip off new lines here. 

https://github.com/mapbox/mapbox-gl-js/blob/master/js/symbol/resolve_text.js#L25

mourner:
@mb12 I don't see how you could implement it there, but maybe I'm missing something — want to do an alternative PR?

js/symbol/shaping.js
```diff
@@ -24,6 +24,8 @@ function Shaping(positionedGlyphs, text, top, bottom, left, right) {
     this.right = right;
 }
 
+var newLine = 0x0a;
```
1ec5: What about carriage returns? Can we be sure that Windows-style CRLF won’t occur?

mourner: `\r` should be ignored in theory so `\r\n` should work the same way, but I could add a test to make sure the results are the same.

ansis:
Looks good to me!

mdehollander:
@mourner Can you give an example how to use this? I try it on city names via the mapbox editor, but that doesn't seem to get recognized: `{name_zh} \n {name_en}`

lanewinfield:
@mourner Also having issues with this. Has this since been broken?

mourner:
Hmm, I'm not sure. @ChrisLoer @1ec5 were there any changes about how we treat new lines that could break this?

ajashton:
The feature works when editing style JSON directly, but is not supported by the text input in the Studio interface (`\n` is automatically converted to `\\n`). We have an internal ticket open about allowing `\n` via Studio.

nextstopsun:
I'm having issues with this. `\n` in feature properties coming from vector tiles is interpreted as text. No line breaks occur.

Maybe an option `text-newline-character` could be added to the spec to give flexibility in choosing appropriate line break symbol?

image

1ec5:
@nextstopsun, it sounds like you’re running into this issue even though `\n` is in the vector tiles (as opposed to being input in Studio). Can you open a separate issue so folks can see and triage it? Thanks!

PR #1060, opened 2015-03-05
Improve mousewheel performance when zooming

tristen:
- Normalize mousewheel deltaY value to match trackwheel
- Drop duration condition that just targets `trackzoom`
- Fixes #1058

### Before merge

Performs as expected in:
- [x]  Firefox, Opera, IE
- [x] Windows
- [x] Non-Apple trackpads
- [x] A variety of different physical hardware (mice)

Issue #1058, opened 2015-03-05
Trackpad behaves differently than mouse wheel


tristen:
### Trackpad

![trackpad](https://cloud.githubusercontent.com/assets/61150/6512100/5f08b70c-c33f-11e4-87c5-3b7d9341824f.gif)

Nice and smooth. No lag.

### Mousewheel

![mouse-wheel](https://cloud.githubusercontent.com/assets/61150/6512105/6320e1fc-c33f-11e4-8a43-80b2dfa5514c.gif)

It looks like events are firing prematurely as the hash corrdinates change before mouse wheel zoom is complete.

PR #3189, opened 2016-09-15
Fix smooth zoom with mouse wheel

### LINKED ISSUE #1730: Smooth zoom transition
`mapbox-gl-js` used to have smooth zoom transition on mousewheel, as seen here: https://www.mapbox.com/blog/mapbox-gl-js/

Has this been disabled or is there a way to enable it?

mourner:
Closes #1730. This turned out to be a pretty tricky issue. Here's how the regression happened:

1. As a part of #2128, @ansis rewrote `zoomTo` to use `easeTo` rather than a custom implementation; however, it accidentally removed [this line of code](https://github.com/mapbox/mapbox-gl-js/pull/2128/commits/9846a4f6d54bd09dfba5c655c56e11d4484458f9#diff-91ad654374133c7bf9558652da3838c1L141) that was there specifically to smooth out mousewheel zoom animation (see https://github.com/mapbox/mapbox-gl-js/commit/f9b61f043ffef4e6287cb27bbd3b97468d628ff5).
2. @tristen noticed that mousewheel zoom feels awkward, and submitted https://github.com/mapbox/mapbox-gl-js/pull/1060, which removed animation from mousewheel zooming altogether, which improved mousewheel experience a lot compared to previously regressed state but in expense of smooth zoom we had earlier.
3. As a result, [this `_updateEasing` code](https://github.com/mapbox/mapbox-gl-js/blob/2924b910d7a9a0a12e0d71bd0771f7529acfade8/js/ui/camera.js#L749-L776) was completely unused for 7 months and I only noticed this because it was highlighted red in test coverage. 😆 

cc @tristen @ansis @lucaswoj @nthtran — please test with your mouses to make sure it works well.

- [x] get a PR review
- [x] make sure it works OK across browsers

### DIFF
--- a/js/ui/camera.js
+++ b/js/ui/camera.js
@@ -435,6 +435,10 @@ util.extend(Camera.prototype, /** @lends Map.prototype */{
         this.rotating = (startBearing !== bearing);
         this.pitching = (pitch !== startPitch);
 
+        if (options.smoothEasing && options.duration !== 0) {
+            options.easing = this._smoothOutEasing(options.duration);
+        }
+
         if (!options.noMoveStart) {
             this.fire('movestart', eventData);
         }
@@ -746,11 +750,12 @@ util.extend(Camera.prototype, /** @lends Map.prototype */{
         return bearing;
     },
 
-    _updateEasing: function(duration, zoom, bezier) {
-        var easing;
+    // only used on mouse-wheel zoom to smooth out animation
+    _smoothOutEasing: function(duration) {
+        var easing = util.ease;
 
-        if (this.ease) {
-            var ease = this.ease,
+        if (this._prevEase) {
+            var ease = this._prevEase,
                 t = (Date.now() - ease.start) / ease.duration,
                 speed = ease.easing(t + 0.01) - ease.easing(t),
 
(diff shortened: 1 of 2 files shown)


mourner:
Oops, nope, disregard the explanation — actually @tristen's change happened long before the new easing regression, so the easing change probably regressed in some other way earlier.

mourner:
With this change, mouse wheel zooming feels pretty good to me — not at all like described in #1058. Which is puzzling — maybe #1058 happened because there was some other regression that got magically fixed by newer easing rewrites but we didn't notice because it was not in action.

tristen:
@mourner 👍 this feels more or less the same as the zoom behaviour on the trackpad.

js/ui/camera.js
```diff
@@ -746,11 +750,12 @@ util.extend(Camera.prototype, /** @lends Map.prototype */{
         return bearing;
     },
 
-    _updateEasing: function(duration, zoom, bezier) {
```
lucaswoj: Was `_updateEasing` entirely dead code before?

js/ui/camera.js
```diff
@@ -435,6 +435,10 @@ util.extend(Camera.prototype, /** @lends Map.prototype */{
         this.rotating = (startBearing !== bearing);
         this.pitching = (pitch !== startPitch);
 
+        if (options.smoothEasing && options.duration !== 0) {
+            options.easing = this._smoothOutEasing(options.duration);
+        }
```
lucaswoj: What would be the implications of enabling smooth easing in all cases?

mourner: I think it's unnecessary — it only makes sense when there is a fast sequence of easings with varying intensities (the only use case for now is mouse wheel zoom), and the algorithm is only suited for easing with the default bezier curve.

js/ui/camera.js
mourner: Yes, entirely dead code. I had to do `git bisect` combined with project search to find the commit that made it so.

PR #3203, opened 2016-09-16
[not ready] Better handling for undefined icon|text-rotation-alignment

boundsj:
This updates the symbol style layer so that it uses the correct values (viewport or map) when the auto value is set explicitly or by default for the properties that support it.

`text-pitch-alignment` continues to parrot text-rotation-alignment but now also does so if its value is explicitly set to auto.

Companion to mapbox/mapbox-gl-style-spec#493, mapbox/mapbox-gl-native#6253, and https://github.com/mapbox/mapbox-gl-test-suite/pull/144.

Part of the federation of PRs to address https://github.com/mapbox/mapbox-gl-native/issues/5683

## Launch Checklist

 - [x] get a PR review

cc @jfirebaugh @lbud

### DIFF
--- a/js/style/style_layer/symbol_style_layer.js
+++ b/js/style/style_layer/symbol_style_layer.js
@@ -10,22 +10,20 @@ function SymbolStyleLayer() {
 module.exports = SymbolStyleLayer;
 
 SymbolStyleLayer.prototype = util.inherit(StyleLayer, {
-
     getLayoutValue: function(name, globalProperties, featureProperties) {
-        if (name === 'text-rotation-alignment' &&
-                this.getLayoutValue('symbol-placement', globalProperties, featureProperties) === 'line' &&
-                !this.getLayoutProperty('text-rotation-alignment')) {
-            return 'map';
-        } else if (name === 'icon-rotation-alignment' &&
-                this.getLayoutValue('symbol-placement', globalProperties, featureProperties) === 'line' &&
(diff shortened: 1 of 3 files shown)


boundsj:
@lbud I think e9b7cfb should unblock https://github.com/mapbox/mapbox-gl-test-suite/pull/144

cc @lucaswoj @jfirebaugh

js/style/style_layer/symbol_style_layer.js
```diff
@@ -12,19 +12,28 @@ module.exports = SymbolStyleLayer;
 SymbolStyleLayer.prototype = util.inherit(StyleLayer, {
 
     getLayoutValue: function(name, globalProperties, featureProperties) {
```
jfirebaugh: Here's a shorter and clearer way to write this:

```js
    getLayoutValue: function(name, globalProperties, featureProperties) {
        var value = StyleLayer.prototype.getLayoutValue.apply(this, arguments);
        if (value !== 'auto') {
            return value;
        }

        switch (name) {
        case 'text-rotation-alignment':
        case 'icon-rotation-alignment':
            return this.getLayoutValue('symbol-placement', globalProperties, featureProperties) === 'line' ? 'map' : 'viewport';
        case 'text-pitch-alignment':
            return this.getLayoutValue('text-rotation-alignment', globalProperties, featureProperties);
        default:
            return value;
        }
    }
```

Issue #3210, opened 2016-09-18
Preserve `id` property on internal GeoJSON objects


mike-marcacci:
When a call to `queryRenderedFeatures` returns features from a GeoJSON source, the `properties` field of matching features is provided in the response, but their `id` field is not. Since this is the recommend place to store such an identifier (per the [GeoJSON spec](http://geojson.org/geojson-spec.html#feature-objects)) it would be great if this were also returned in the results!

mike-marcacci:
I went to implement assuming it would be quite trivial (and it was!)... but then I discovered that the PCB type set for `id` is `Int`, which isn't flexible enough given that the spec states no preference and it's quite common to use UUIDs, etc.

mike-marcacci:
It appears that the [vector tile spec](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features) also has no preference of `id` type, but that it's quite established as an integer in the implementations. I submitted [a quick PR to the spec](https://github.com/mapbox/vector-tile-spec/pull/87) to help clarify.

To get around this, I've whipped up PR #3211 which uses the index of the feature in the GeoJSON `FeatureCollection` as its id over the Protobuf layer, and maps it back in `queryRenderedFeatures`. This is probably NOT the cleanest way to handle it (the conditional logic in particular), but it provides a proof of concept.

jfirebaugh:
Hi Mike, thanks for the bug report and PR!

> the `properties` field of matching features is provided in the response, but their `id` field is not

I believe integer `id`s are expected to be preserved, as of https://github.com/mapbox/geojson-vt/pull/70 and https://github.com/mapbox/vt-pbf/pull/8 -- although it looks as if our package.json is compatible with, but does not strictly require, vt-pbf v2.1.0.

Preservation of non-integer `id`s is #2716.

mike-marcacci:
Awesome! Thanks for linking #2716 - hadn't seen that one, and I definitely like the idea of its geobuf solution!

mike-marcacci:
@jfirebaugh - also, just to be clear, this is easily fixed by adding `this.id = feature.id` [here](https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/geojson_wrapper.js#L30), but this will *break existing apps where the GeoJSON `id` is a string*, hence my workaround.

Alternatively, you could just add a check for type, and only set it if it's a number until #2716 lands.

Issue #2716, opened 2016-06-10
Return features with their original ids from query*Features


danvk:
Here's a jsbin demonstrating the issue:
https://jsbin.com/bafewapela/edit?html,console,output

The GeoJSON spec [states][1] that features may have a top-level id property:

> If a feature has a commonly used identifier, that identifier should be included as a member of the feature object with the name "id".

However, when I have a source with such a feature:

```javascript
var markers = {
    "type": "FeatureCollection",
    "features": [{
        "id": "my-feature-id",  // <--- feature has an ID
        "type": "Feature",
        "properties": {
            "marker-symbol": "theatre",
        },
        "geometry": {
            "type": "Point",
            "coordinates": [-77.038659, 38.931567]
        }
    }]
};
```

add it to the map and then get features back via `queryRenderedFeatures`, the `id` is gone. The `properties` and `geometry` properties are there, but `id` is not. It would be really helpful if this special property were preserved.

The workaround is to duplicate IDs into `properties.id` before adding features to the map, but this creates extra work and contradicts the GeoJSON spec.

[1]: http://geojson.org/geojson-spec.html#feature-objects

**mapbox-gl-js version**: 0.19.1

jfirebaugh:
Upstream PR which this depends on: https://github.com/mapbox/geojson-vt/pull/60.

mourner:
This was fixed on the geojson-vt side, but now ids [don't survive the transfer](https://github.com/mapbox/mapbox-gl-js/pull/2888#issuecomment-233482084) through `vt-pbf`. We'll probably want to switch to something like Geobuf.

lucaswoj:
@mourner I'm excited to (finally!) make geobuf part of GL JS!! 

Have you considered transferring the feature ids through a specially named property, like `$id`?

mourner:
@lucaswoj we could do this, but this feels kinda hacky and fragile to me (e.g. what if the data has both `id` and a property `$id`?). It might be good enough as a stopgap though.

lucaswoj:
> what if the data has both `id` and a property `$id`

Then our `$id` filtering will break anyway!

anandthakker:
@mourner see https://github.com/mapbox/vt-pbf/pull/8

mourner:
@anandthakker thanks! I was referring about vector-tile-spec limitation oh having ids as uints which John pointed out in the PR, so the original issue would still be present as string ids are not encoded.

anandthakker:
[...]

PR #3211, opened 2016-09-18
POC: bypass Protobuf to supply GeoJSON IDs

mike-marcacci:
See issue #3210. This is just a proof of concept, and likely NOT the cleanest way to implement this feature.

## Launch Checklist

 - [ ] post benchmark scores
 - [ ] get a PR review

### DIFF
--- a/js/source/geojson_source.js
+++ b/js/source/geojson_source.js
@@ -56,6 +56,7 @@ module.exports = GeoJSONSource;
  *     data: {
  *        "type": "FeatureCollection",
  *        "features": [{
+ *            "id": 65342,
  *            "type": "Feature",
  *            "properties": { "name": "Null Island" },
  *            "geometry": {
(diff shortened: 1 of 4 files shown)


jfirebaugh:
Thanks for the PR! As you surmise, this is likely not the approach we want to take for fixing this issue. (In fact, I think it's a performance bug that `_data` is even accessible on the `GeoJSONSource`: #3220.) Appreciate the contribution nonetheless. Can I interest you in prototyping the geobuf-based approach suggested in #2716? 😄

mike-marcacci:
@jfirebaugh absolutely - I'll give the geobuf approach a go this weekend! Is there a ticket anywhere specifically for the geobuf switch? There are a few questions that I'd like to ask before taking a stab at it, like:

Right now the protobuf format appears to be the bottleneck that limits `properties` to a depth of one (thus limiting layer filters). It appears that the current approach is to stringify nested objects/arrays, and I just want to make sure that it's not common for people to use filters based on this assumption.

jfirebaugh:
I don't think anyone is writing filters that depend on the stringification of nested properties. In any case, using geobuf for the worker ⇢ main thread transfer wouldn't affect that, because filters are evaluated on the worker.

We do get requests that `query*Features` preserve nested properties in GeoJSON features, which using geobuf for the worker ⇢ main thread transfer could solve.

Issue #684, opened 2014-08-11
Support rendering 3d extrusions


amyleew:
Please add building height styling options for building data  in Mapbox GL. 

Specifically [building data](https://www.mapbox.com/carto/api/2.3.0/#building
) in CartoCSS can be rendered in a 3d look using building-fill, building-fill-opacity, and building-height.

Can we add this functionality to Mapbox GL?

trallarn:
I know this issue is closed, but I can't find any doc on how to render extrusions. Has this feature really been added?

kkaefer:
No, this features hasn't been added yet. There's an experimental branch that has real buildings, which will eventually get merged though.

jonathanward:
What is the status of development on the 2.5d line extruding?

ansis:
implementing this:

### the first part

- merge the earcut branch https://github.com/mapbox/mapbox-gl-js/pull/1606
- roughly recreate [the really old demo branch](https://github.com/mapbox/mapbox-gl-js/tree/building) ([example](https://www.mapbox.com/bites/00135/#15.92/40.7212/-73.9928/-39.3)) on top of master. Manually rewriting everything with the old branch as a guide is probably better than trying to rebase.
- figure out how this should look in the style spec. Discuss and implement that
- add render tests
- port to -native

### followup

- fix antialiasing (draw faded 1px lines along all edges?)
- fix depth related bugs

lbud:
I've roughly recreated last year's demo branch at https://github.com/mapbox/mapbox-gl-js/tree/building-redux (thanks to the codespirit of @ansis 👻 the old branch ported pretty easily).

![image](https://cloud.githubusercontent.com/assets/3170312/15556773/a685e50a-229d-11e6-9505-42a0233bd2b8.png)

Shaders at https://github.com/mapbox/mapbox-gl-shaders/tree/js-buildings, a spec branch (filled in only by copying fill properties for validation purposes, not definitive spec suggestions) at https://github.com/mapbox/mapbox-gl-style-spec/tree/buildings. I've modified this repo's debug page/style in this branch for the sake of quick iteration, not to be merged.

Still lots of TODOs to clean up rough edges in this branch + will open a spec ticket to discuss what this should look like.

jonathanward:
Nice work with the development!

I wanted to get a sense of how far along the development progress is - how long do you expect until functionality analogous to [http://keks.com/dev/line3/ ](http://keks.com/dev/line3/ ) will be present.

[...]

PR #3223, opened 2016-09-19
Add fill-extrude-height and fill-extrude-base properties.

lbud:
This PR adds support for `fill-extrude-height` and `fill-extrude-base`.

I would love to start getting some 👀  on this branch.

Notably, this diverges from our existing patterns in the following ways: 
* Includes two different vertex interfaces in fill_bucket.js and diverges into fill VAOs or extrusion VAOs depending on the style
* Adds a draw file that does not correspond to a style spec type: `draw_extrusion` handles complete drawing of extruded fills; this diversion happens in painter.js

## To do:
- [x] ~~spherical interpolation~~
- [x] update package.json shas once all corresponding PRs are finalized.
- [ ] work with @lucaswoj to figure out how to benchmark this feature

Closes https://github.com/mapbox/mapbox-gl-js/issues/684.

Companion PRs:
https://github.com/mapbox/mapbox-gl-style-spec/pull/495
https://github.com/mapbox/mapbox-gl-test-suite/pull/145
https://github.com/mapbox/mapbox-gl-shaders/pull/25

@jfirebaugh @lucaswoj @mourner @mollymerp 

## Launch Checklist

 - [ ] post benchmark scores


dagjomar:
Is this for real? 😮  Can you post some screenshots of this?

js/render/draw_extrusion.js
```diff
@@ -0,0 +1,292 @@
...
+    // Create a new texture to which to render the extrusion layer. This approach
+    // allows us to adjust opacity on a per-layer basis (eliminating the interior
+    // walls per-feature opacity problem)
+    var texture = new ExtrusionTexture(gl, painter, layer);
+    texture.bindFramebuffer();
+
+    gl.clearStencil(0x80);
+    gl.stencilMask(0xFF);
+    gl.clear(gl.COLOR_BUFFER_BIT | gl.DEPTH_BUFFER_BIT);
+    gl.stencilMask(0x00);
```
lucaswoj: Does this overwrite our tile clipping masks?

js/render/painter.js
```diff
@@ -105,6 +106,15 @@ Painter.prototype.setup = function() {
 };
 
 /*
+ * Set the lighting properties (used for extruded geometries).
+ */
+Painter.prototype.setLight = function(lightOptions) {
+    for (var key in lightOptions) {
+        this.light[key] = lightOptions[key];
+    }
+};
```
lucaswoj: I see `Painter` as a collection of GL utilities with no knowledge of map primitives and `Style` as the owner of state about the map style. Why did you choose to put this `light` state on `Painter` rather than `Style`?

js/render/painter.js
```diff
@@ -306,18 +326,28 @@ Painter.prototype.translatePosMatrix = function(matrix, tile, translate, anchor)
...
-
-Painter.prototype.getTexture = function(size) {
-    var textures = this.reusableTextures[size];
-    return textures && textures.length > 0 ? textures.pop() : null;
+Painter.prototype.getTexture = function(width, height) {
+    var widthTextures = this.reusableTextures[width];
+    if (widthTextures) {
+        var textures = widthTextures[height || width];
+        return textures && textures.length > 0 ? textures.pop() : null;
+    }
```
lucaswoj: Did you verify that the performance benefits of reusing textures are still worthwhile with these added constraints? Given that we never deallocate these textures, are you concerned about resource leaks?

js/style/style_layer/extrusion_style_layer.js
```diff
@@ -0,0 +1,11 @@
...
+
+var util = require('../../util/util');
+var StyleLayer = require('../style_layer');
+
+function ExtrusionStyleLayer() {
+    StyleLayer.apply(this, arguments);
+}
+
+module.exports = ExtrusionStyleLayer;
+ExtrusionStyleLayer.prototype = util.inherit(StyleLayer, {});
```
lucaswoj: Alongside `CircleStyleLayer` for `circles`, `FillStyleLayer` for `fills`, etc, this class name would suggest the existence of an `extrusion` layer type. Could we change to `FillExtrusionStyleLayer`?

js/ui/light.js
```diff
@@ -0,0 +1,256 @@
...
+        if (lightOpts.anchor) this._setLightAnchor(lightOpts.anchor);
+        if (lightOpts.color) this._setLightColor({rawColor: lightOpts.color});
+        if (lightOpts.direction) this._setLightDirection(lightOpts.direction);
+        if (typeof lightOpts.intensity !== 'undefined') this._setLightIntensity(lightOpts.intensity);
+
+        this._update();
+
+        return this;
+    }
+});
```
lucaswoj: This class could be simplified by taking advantage of the existing `StyleDeclaration` and `StyleTransition` classes.

js/ui/light.js
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Changes any combination of light values, with an animated transition
+     * between old and new values (with the exception of `anchor`, which does
+     * not animate). The map will retain its current values for any
+     * details not specified in `options`.
+     *
+     * @param {LightOptions} options Options describing the target light properties of the transition.
+     * @param {Object} [eventData] Data to propagate to any event listeners.
+     * @fires lightstart
+     * @fires lightend
```
lucaswoj: I like the idea of an events that  mark the beginning and end of a style property transition. I'm not sure we need such an event specifically for `light` properties, however. What do yout think about a follow-up PR that adds `transitionstart` and `transitionend` events?

js/ui/light.js
```diff
@@ -0,0 +1,256 @@
...
+
+        this.fire('lightend', eventData);
+    },
+
+    _setLightAnchor: function(anchor) {
+        if (anchor === 'map' || anchor === 'viewport') {
+            this.painter.setLight({
+                anchor: anchor
+            });
+        } else throw new Error('light.anchor must be one of: `map`, `viewport`');
```
lucaswoj: All validation logic that can be done in the style spec should be done in the style spec.

js/ui/light.js
```diff
@@ -0,0 +1,256 @@
...
+        return this;
+    },
+
+    _validateDirection: function(direction) {
+        if (!Array.isArray(direction) || direction.length !== 3 ||
+            !direction.every(function(i) { return !isNaN(+i); }))
+            throw new Error('light.direction must be an array of three numbers');
+
+        if (direction[1] < 0 || direction[1] >= 360) util.warnOnce('Azimuthal angle will be wrapped to [0, 360)');
+        if (direction[2] < 0 || direction[2] >= 180) util.warnOnce('Polar angle will be clamped to [0, 180)');
```
lucaswoj: Hard failures or smart fallbacks are preferable to warnings.

js/ui/light.js
```diff
@@ -0,0 +1,256 @@
...
+        azimuthal *= Math.PI / 180;
+        polar *= Math.PI / 180;
+
+        // spherical to cartesian (x, y, z)
+        return [
+            r * Math.cos(azimuthal) * Math.sin(polar),
+            r * Math.sin(azimuthal) * Math.sin(polar),
+            r * Math.cos(polar)
+        ];
+    },
```
lucaswoj: This looks like a good candidate for a standalone util

js/ui/light.js
lbud: Because of the `setLight` runtime API I think we should validate here, no?

lucaswoj: `light` is associated with the style. 

We keep all [style validation code separate](https://github.com/mapbox/mapbox-gl-style-spec/tree/mb-pages/lib/validate) from GL JS so that we can have a standalone style validation utility. This is used, for example, by the style upload API. 

[Here is an example of where we use the style spec's validation code to validate layers](http://github.com/mapbox/mapbox-gl-js/blob/ece53f12f4bd72aaa555813a5363a1fd0c9d19f6/js/style/style.js#L497-L498)

js/style/style_layer/extrusion_style_layer.js
lbud: Oops — this is actually unused, forgot to delete when we nixed the Extrusion layer type. Removing.

js/ui/light.js
lbud: Follow-up PR sounds good 👍

js/render/painter.js
lbud: changed

js/ui/light.js
lbud: done

js/ui/light.js
lbud: gone

js/ui/light.js
lbud: done

js/render/draw_extrusion.js
lbud: Here's what I've figured out:
* `gl.clear` only affects the currently bound framebuffer and is needed here
* I can delete several of these lines (basically all but the `gl.clear`) and get visually the same results. My best guess is that this was in here because the original implementation (from which I copied these) rendered to the map as a normal layer, so it needed to fiddle with the stencil mask (so as to clear the entire screen). Now that we're rendering in a new framebuffer, it doesn't need to redraw a mask that's not actually there — so, deleting all these lines except the `gl.clear`, which clears its last frame draw. (Fwiw/for future reference: based on my testing I actually don't think it was affecting the global stencil mask anyway because of the different framebuffer.)
* testing for the clipping masks made me realize I forgot to flip the depthMask back off after this draw, which was causing fills to z-fight if above an extrusion layer — fixed 😅

js/render/painter.js
lbud: fixed — we're now saving one texture at a time that can be reused for every subsequent frame until the map is resized, at which point we delete that texture and create a new one for the new size

lbud:
Per https://github.com/mapbox/mapbox-gl-style-spec/issues/240#issuecomment-252126033 I've disabled `fill-outline-color` for extruded fills in anticipation of that paint property being deprecated for fills entirely in the near future. 
CI doesn't agree with my render tests on extrusion patterns, but once those are fixed this is ready to 🚗 💨

lbud:
@dagjomar it is for real :) 

pan
dds

js/render/draw_extrusion.js
lbud: > testing for the clipping masks made me realize I forgot to flip the depthMask back off after this draw, which was causing fills to z-fight if above an extrusion layer — fixed 😅

nope I was totally wrong, this broke other fills and also is https://github.com/mapbox/mapbox-gl-js/issues/3320

lbud: ☝️ fixed

js/render/painter.js
lucaswoj: Perfect 😄

lucaswoj:
And enjoy your vacation!

PR #3232, opened 2016-09-20
Enforce LngLatLike param in LngLat.convert

lbud:
This PR enforces `LngLatLike`-ness of the `LngLatLike` argument to `LngLat.convert`. 

Previously `LngLat.convert` passed through non-`LngLatLike` (as class instance or array) arguments unchanged. In updating tests I caught once instance where its arguments become `{lng: <lng>, lat: <lat>}` but _not_ a `LngLat` class instance, so I added that conditional to `LngLat.convert`. If its argument is none of those three types it now throws an error describing the nature of a `LngLatLike` argument. 

This method is used in the internals of methods like `setCenter`, `panTo`, etc etc and should stop error reports of the ever-cryptic ["failed to invert matrix"](https://github.com/mapbox/mapbox-gl-js/issues?utf8=%E2%9C%93&q=is%3Aissue%20failed%20to%20invert%20matrix).

@lucaswoj — there's not some special reason I'm missing why `LatLng.convert` may need to pass through non-LngLatLike arguments, is there? (I removed the one test that verified that worked: https://github.com/mapbox/mapbox-gl-js/commit/c3820aed73fde3737be767de13dfa342731e5845#diff-c6936a6d2842db2080af49f8c5dac23aL27)

## Launch Checklist

### DIFF
--- a/js/geo/lng_lat.js
+++ b/js/geo/lng_lat.js
@@ -82,7 +82,7 @@ LngLat.prototype.toString = function () {
 LngLat.convert = function (input) {
     if (input instanceof LngLat) {
         return input;
-    } else if (typeof input !== 'undefined' && input.hasOwnProperty('lng') && input.hasOwnProperty('lat')) {
+    } else if (input && input.hasOwnProperty('lng') && input.hasOwnProperty('lat')) {
         return new LngLat(input.lng, input.lat);
     } else if (Array.isArray(input) && input.length === 2) {
         return new LngLat(input[0], input[1]);
--- a/js/geo/lng_lat.js
+++ b/js/geo/lng_lat.js
@@ -82,9 +82,11 @@ LngLat.prototype.toString = function () {
 LngLat.convert = function (input) {
     if (input instanceof LngLat) {
         return input;
-    }
-    if (Array.isArray(input)) {
+    } else if (input && input.lng && input.lat) {
+        return input;
+    } else if (Array.isArray(input) && input.length === 2) {
         return new LngLat(input[0], input[1]);
+    } else {
+        throw new Error("`LngLatLike` argument must be specified as a LngLat instance, an object {lng: <lng>, lat: <lat>}, or an array of [<lng>, <lat>]");
     }
-    return input;
 };
--- a/js/geo/lng_lat.js
+++ b/js/geo/lng_lat.js
@@ -82,8 +82,8 @@ LngLat.prototype.toString = function () {
 LngLat.convert = function (input) {
     if (input instanceof LngLat) {
         return input;
-    } else if (input && input.lng && input.lat) {
-        return input;
+    } else if (typeof input !== 'undefined' && input.hasOwnProperty('lng') && input.hasOwnProperty('lat')) {
+        return new LngLat(input.lng, input.lat);
     } else if (Array.isArray(input) && input.length === 2) {
         return new LngLat(input[0], input[1]);
     } else {
--- a/js/geo/lng_lat_bounds.js
+++ b/js/geo/lng_lat_bounds.js
@@ -81,9 +81,9 @@ LngLatBounds.prototype = {
 
         } else {
             if (Array.isArray(obj)) {
-                if (obj.every(function(i) { return Array.isArray(i); })) {
+                if (obj.every(Array.isArray)) {
                     return this.extend(LngLatBounds.convert(obj));
-                } else if (obj.every(function(i) { return !isNaN(i); })) {
+                } else {
                     return this.extend(LngLat.convert(obj));
                 }
             }
--- a/js/geo/lng_lat_bounds.js
+++ b/js/geo/lng_lat_bounds.js
@@ -80,7 +80,14 @@ LngLatBounds.prototype = {
             if (!sw2 || !ne2) return this;
 
         } else {
-            return obj ? this.extend(LngLat.convert(obj) || LngLatBounds.convert(obj)) : this;
+            if (Array.isArray(obj)) {
+                if (obj.every(function(i) { return Array.isArray(i); })) {
+                    return this.extend(LngLatBounds.convert(obj));
+                } else if (obj.every(function(i) { return !isNaN(i); })) {
+                    return this.extend(LngLat.convert(obj));
+                }
+            }
+            return this;
         }
 
         if (!sw && !ne) {
--- a/test/js/geo/lng_lat.test.js
+++ b/test/js/geo/lng_lat.test.js
@@ -24,7 +24,9 @@ test('LngLat', function(t) {
     t.test('#convert', function(t) {
         t.ok(LngLat.convert([0, 10]) instanceof LngLat, 'convert creates a LngLat instance');
         t.ok(LngLat.convert(new LngLat(0, 0)) instanceof LngLat, 'convert creates a LngLat instance');
-        t.equal(LngLat.convert('othervalue'), 'othervalue', 'passes through other values');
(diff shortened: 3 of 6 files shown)


js/geo/lng_lat.js
```diff
@@ -82,9 +82,11 @@ LngLat.prototype.toString = function () {
 LngLat.convert = function (input) {
     if (input instanceof LngLat) {
         return input;
-    }
-    if (Array.isArray(input)) {
+    } else if (input && input.lng && input.lat) {
```
tmcw: This check is dodgy: `0, 0` is a valid point but this will fail since `0` is falsy

js/geo/lng_lat.js
```diff
@@ -82,9 +82,11 @@ LngLat.prototype.toString = function () {
...
+        return input;
```
mourner: If given `{lng, lat}` object, it should return a `LngLat` instance rather than the object.

js/geo/lng_lat.js
```diff
@@ -82,8 +82,8 @@ LngLat.prototype.toString = function () {
 LngLat.convert = function (input) {
     if (input instanceof LngLat) {
         return input;
-    } else if (input && input.lng && input.lat) {
-        return input;
+    } else if (typeof input !== 'undefined' && input.hasOwnProperty('lng') && input.hasOwnProperty('lat')) {
```
mourner: `{}` is truthy, so no need for the first typeof. Also, I think you can safely use `'lng' in input` instead of `hasOwnProperty` because it doesn't matter in this case if it's own or inherited from a prototype.

js/geo/lng_lat_bounds.js
```diff
@@ -80,7 +80,14 @@ LngLatBounds.prototype = {
             if (!sw2 || !ne2) return this;
 
         } else {
-            return obj ? this.extend(LngLat.convert(obj) || LngLatBounds.convert(obj)) : this;
+            if (Array.isArray(obj)) {
+                if (obj.every(function(i) { return Array.isArray(i); })) {
+                    return this.extend(LngLatBounds.convert(obj));
+                } else if (obj.every(function(i) { return !isNaN(i); })) {
+                    return this.extend(LngLat.convert(obj));
+                }
```
mourner: Let's avoid closures here — they're not necessary and can be a perf hit.

lbud:
👍  thanks for the review everybody — have addressed in https://github.com/mapbox/mapbox-gl-js/pull/3232/commits/a702272a728bb1715551c42d850816dc8e1dbb91 and https://github.com/mapbox/mapbox-gl-js/pull/3232/commits/c1e423cc55ce3c7ef28f983eb1193ec56a66a225.

Good catch @lucaswoj — I fixed and went through everywhere we use `LngLat.convert` to ensure this doesn't break anything:

| line reference | code | comment |
| --- | --- | --- |
| [./geo/lng_lat.js:79:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/geo/lng_lat.js#L79) | `* var ll = mapboxgl.LngLat.convert(arr);` | documentation |
| [./geo/lng_lat.js:82:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/geo/lng_lat.js#L82) | `LngLat.convert = function (input) {` | function signature |
| [./geo/lng_lat_bounds.js:46:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/geo/lng_lat_bounds.js#L46) | `       this._ne = LngLat.convert(ne);` | depends on strict LngLatLike-ness |
| [./geo/lng_lat_bounds.js:57:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/geo/lng_lat_bounds.js#L57) | `       this._sw = LngLat.convert(sw);` | depends on strict LngLatLike-ness |
| [./geo/lng_lat_bounds.js:87:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/geo/lng_lat_bounds.js#L87) | `                   return this.extend(LngLat.convert(obj));` | depends on LngLatLike-ness (enforces array with numeric members) |
| [./geo/lng_lat_bounds.js:207:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/geo/lng_lat_bounds.js#L207) | `* Internally, the function calls `LngLat#convert` to convert arrays to `LngLat` values.` | documentation |
| [./geo/transform.js:282:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/geo/transform.js#L282) | `           ll = LngLat.convert(lnglat);` | depends on strict LngLatLike-ness |
| [./source/image_source.js:98:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/source/image_source.js#L98) | `           return map.transform.locationCoordinate(LngLat.convert(coord)).zoomTo(0);` | depends on strict LngLatLike-ness |
| [./source/video_source.js:122:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/source/video_source.js#L122) | `           return map.transform.locationCoordinate(LngLat.convert(coord)).zoomTo(0);` | depends on strict LngLatLike-ness |
| [./ui/camera.js:347:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/ui/camera.js#L347) | `           tr.center = LngLat.convert(options.center);` | depends on strict LngLatLike-ness |
| [./ui/camera.js:420:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/ui/camera.js#L420) | `           toLngLat = LngLat.convert(options.center);` | depends on strict LngLatLike-ness |
| [./ui/camera.js:423:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/ui/camera.js#L423) | `           toLngLat = LngLat.convert(options.around);` | depends on strict LngLatLike-ness |
| [./ui/camera.js:572:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/ui/camera.js#L572) | `       var center = 'center' in options ? LngLat.convert(options.center) : this.getCenter();` | depends on strict LngLatLike-ness / is only called if LngLatLike obj is defined |
| [./ui/map.js:429:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/ui/map.js#L429) | `       return this.transform.locationPoint(LngLat.convert(lnglat));` | depends on strict LngLatLike-ness |
| [./ui/marker.js:89:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/ui/marker.js#L89) | `       this._lngLat = LngLat.convert(lnglat);` | depends on strict LngLatLike-ness |
| [./ui/popup.js:141:](https://github.com/mapbox/mapbox-gl-js/blob/a702272a728bb1715551c42d850816dc8e1dbb91/js/ui/popup.js#L141) | `       this._lngLat = LngLat.convert(lnglat);` | depends on strict LngLatLike-ness |

PR #3244, opened 2016-09-22
Add event propagation system to Evented

lucaswoj:
Adding a new `Map` event that needs to be fired from another class requires a lot of GL JS boilerplate. Before tackling https://github.com/mapbox/mapbox-gl-js/issues/1715, I built an event propogation system that...

 - adds a `Evented#setEventedParent` method which propagates events 
 - uses `Evented#setEventedParent` to eliminate repetitive binding / unbinding of `_forward*Event` methods 
 - uses `Evented#setEventedParent` allow event names to remain constant from `Evented#fire` to `Map` listeners, making it easier to add new events and follow the path of existing events
 - improves test coverage of `Evented`
 - eliminates the ability to call `Evented#off` with less than two arguments, removing all listeners (this could break GL JS core)
 - eliminates lines of code

I'm still not 💯 on the name `Evented#setEventedParent`. Any other ideas?

cc @jfirebaugh @mourner @mollymerp @lbud @mapsam 

# Launch Checklist

 - [x] post benchmark scores

# Benchmarks

## map-load

**master 594248c:** 132 ms
**datastart 102807b:** 114 ms

## style-load

**master 594248c:** 113 ms
**datastart 102807b:** 110 ms

## buffer

**master 594248c:** 984 ms
**datastart 102807b:** 1,092 ms

## fps

**master 594248c:** 60 fps
**datastart 102807b:** 60 fps

## frame-duration

**master 594248c:** 8.1 ms, 1% > 16ms
**datastart 102807b:** 8 ms, 4% > 16ms

## query-point

**master 594248c:** 1.15 ms
**datastart 102807b:** 1.23 ms

## query-box

**master 594248c:** 92.81 ms
**datastart 102807b:** 92.15 ms

## geojson-setdata-small

**master 594248c:** 15 ms
**datastart 102807b:** 13 ms

## geojson-setdata-large

**master 594248c:** 131 ms
**datastart 102807b:** 133 ms

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -62,7 +62,7 @@ function Style(stylesheet, animationLoop, options) {
 
         if (stylesheet.sprite) {
             this.sprite = new ImageSprite(stylesheet.sprite);
-            this.sprite.forwardEvents(this);
+            this.sprite.setEventedParent(this);
         }
 
         this.glyphSource = new GlyphSource(stylesheet.glyphs);
@@ -141,7 +141,7 @@ Style.prototype = util.inherit(Evented, {
             if (layerJSON.ref) continue;
             layer = StyleLayer.create(layerJSON);
             this._layers[layer.id] = layer;
-            layer.forwardEvents(this, {layer: {id: layer.id}});
+            layer.setEventedParent(this, {layer: {id: layer.id}});
         }
 
         // resolve all layers WITH a ref
@@ -151,7 +151,7 @@ Style.prototype = util.inherit(Evented, {
             var refLayer = this.getLayer(layerJSON.ref);
             layer = StyleLayer.create(layerJSON, refLayer);
             this._layers[layer.id] = layer;
-            layer.forwardEvents(this, {layer: {id: layer.id}});
+            layer.setEventedParent(this, {layer: {id: layer.id}});
         }
 
         this._groupLayers();
@@ -335,7 +335,7 @@ Style.prototype = util.inherit(Evented, {
         source = new SourceCache(id, source, this.dispatcher);
         this.sources[id] = source;
         source.style = this;
-        source.forwardEvents(this, {source: source});
+        source.setEventedParent(this, {source: source});
 
         this._updates.events.push(['source.add', {source: source}]);
         this._updates.changed = true;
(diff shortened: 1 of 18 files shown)


jfirebaugh:
This looks great. We can apply these ideas to https://github.com/mapbox/mapbox-gl-native/issues/6383 as well.

> I'm still not 💯 on the name `Evented#setEventedParent`. Any other ideas?

I think it's fine, but if you are looking for something more descriptive, I'd suggest `setParentForEventBubbling`.

js/style/style.js
```diff
@@ -67,12 +62,12 @@ function Style(stylesheet, animationLoop, options) {
 
         if (stylesheet.sprite) {
             this.sprite = new ImageSprite(stylesheet.sprite);
-            this.sprite.on('load', this.fire.bind(this, 'change'));
```
jfirebaugh: Is there some other mechanism now for triggering a repaint when the sprite finishes loading?

lucaswoj: Yes. `SpriteImage` fires a `style.change` event which is propagated to `Style` then `Map`.  The same flow occurs without this `on` -> `fire` interaction within `Style.`

js/style/style.js
```diff
@@ -420,7 +399,7 @@ Style.prototype = util.inherit(Evented, {
         }
         this._validateLayer(layer);
 
-        layer.on('error', this._forwardLayerEvent);
+        layer.setEventedParent(this, {layer: {id: layer.id}});
```
mourner: Is it necessary to have nested `{layer: {id: ...}` data for some compatibility reasons, or could we just have `{layerId: ...}`?

js/ui/map.js
```diff
@@ -166,8 +157,8 @@ var Map = module.exports = function(options) {
     this._setupContainer();
     this._setupPainter();
 
-    this.on('move', this._update.bind(this, false));
-    this.on('zoom', this._update.bind(this, true));
+    this.on('move', this._update);
+    this.on('zoom', this._update);
```
mourner: Did you intentionally remove `bind`s here? They set the `updateStyle` argument of `_update` (e.g. so that `move` doesn't trigger zoom recalc as far as I remember and `zoom` does). Without `bind`, `updateStyle` will always be truthy because `move` and `zoom` pass event data to listeners.

js/style/style.js
lucaswoj: As this is a public event, I suspect that changing the structure would impact our users.

The intent behind this structure is to give us the flexibility to make `StyeLayer`s part of the public API. Think of `{id: ...}` as the public interface of `StyleLayer`. 😅 

I am working on adding custom layer types now, which bring these classes further into the public API.

lucaswoj:
Glad to hear this is in line with what we've done in other projects! Excited to 🚢 and keep digging on #1715

Issue #2741, opened 2016-06-14
Refactor to simplify Map


lucaswoj:
[`Map`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/map.js) is 1342 lines of code and exposes about a hundred methods, properties, and events. We should simplify `Map`'s interface and implementation by **moving as much logic downstream as possible** and maybe **moving some interface components downstream**.
# moving as much logic downstream as possible

Having many methods, properties, and events on `Map` would be more manageable if `Map` did not have any logic or state of its own. 

For example, we should endevour to rewrite `setMaxBounds`

``` js
    setMaxBounds: function (lnglatbounds) {
        if (lnglatbounds) {
            var b = LngLatBounds.convert(lnglatbounds);
            this.transform.lngRange = [b.getWest(), b.getEast()];
            this.transform.latRange = [b.getSouth(), b.getNorth()];
            this.transform._constrain();
            this._update();
        } else if (lnglatbounds === null || lnglatbounds === undefined) {
            this.transform.lngRange = [];
            this.transform.latRange = [];
            this._update();
        }
        return this;

    },
```

such that it reads a little more like

``` js
    setMaxBounds: function (lnglatbounds) {
        transform.setMaxBounds(lnglatbounds);
        return this;
    },
```
# (maybe) moving some interface components downstream

i.e. change the public interface from `map.setPaintProperty(...)` to `map.style.setPaintProperty(...)` (or `map.style.paint['circle-color'].set(...)`)

jfirebaugh:
> i.e. change the public interface from `map.setPaintProperty(...)` to `map.style.setPaintProperty(...)` (or `map.style.paint['circle-color'].set(...)`)

The core mbgl native equivalent is `map.getLayer(id).setCircleColor(...)` (modulo some downcasting from the base `Layer` type to `CircleLayer`).

The difficulty of pushing the setters down to below the map level is how do you trigger the subsequent style recalculations? When I was first implementing `map.setPaintProperty`, I considered several alternatives:

* Require downstream code to manually trigger style recalculation. For native I'm going to see if the "downstream code" in the iOS SDK or Android SDK bindings can do this. I don't think we want this implementation detail to leak to actual user code.
* Have every mutable object have a reference back to the containing `Map`, so that the setter can trigger the recalculation.
* Implement the observer pattern and have the map register for notifications from each mutable object. If individual layers or properties have a `set` method, this becomes cumbersome.

[...]

PR #3255, opened 2016-09-26
Add "data" event

lucaswoj:
This PR consolidates various undocumented data lifecycle events into the new `data` and `dataloading` events. 

 - [x] replace `source.change` with `data`
 - [x] augment `source.load` with `data`
 - [x] replace `style.change` with `data` 
 - [x] augment `style.load` with `data`
 - [x] replace `tile.add` with `dataloading`
 - [x] replace `tile.load` with `data`
 - [x] replace `tile.remove`with `data`
 - [x] replace `source.add` with `data`
 - [x] replace `source.remove` with `data`
 - [x] replace `layer.add` with `data`
 - [x] replace `layer.remove` with `data`
 - [x] document `data` event
 - [x] document `dataloading` event
 - [x] merge `geoJSON`, `tileJSON`, `image`, and `video` types into a `source` type
 - [x] merge `sprite` into the `style` type

Having an event per resource type per possible mutation produced a complex taxonomy of events. Understanding, documenting, and using these events was difficult. Most users did not require the level of granularity exposed by this taxonomy. Some use cases required listening to many different events. 

The unified `data` event is easy to understand, document, and use. Users who require more granularity may inspect the `MapDataEvent` object. 

> A `MapDataEvent` object is attached to the [`Map#data`](#Map.event:data) event.
> The `Map#data` event is fired when any map data loads or changes. The types
> of map data are:
> 
> - `'geoJSON'`: [GeoJSON](http://geojson.org/) data associated with a `geojson` source.
> - `'tileJSON'`: [TileJSON](https://github.com/mapbox/tilejson-spec) metadata associated with a `vector` or `raster` source.
> - `'style'`: The [style](https://www.mapbox.com/mapbox-gl-style-spec/) used by the map
> - `'sprite'`: The icons and patterns used by the style
> - `'image'`: A `image` source's image and coordinates
> - `'video'`: A `video` source's video and coordinates
> - `'tile'`: A vector or raster tile

cc @jfirebaugh @lbud @mollymerp @mapsam @mourner 

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/bench/lib/set_data_perf.js
+++ b/bench/lib/set_data_perf.js
@@ -7,7 +7,9 @@ module.exports = function(source, numCalls, geojson, cb) {
     var startTime = null;
     var times = [];
 
-    source.on('tile.load', function tileCounter() {
+    source.on('data', function tileCounter(event) {
+        if (event.dataType !== 'tile') return;
+
         tileCount++;
         if (tileCount === NUM_TILES) {
             tileCount = 0;
@@ -20,7 +22,7 @@ module.exports = function(source, numCalls, geojson, cb) {
                 var avgTileTime = times.reduce(function (v, t) {
                     return v + t;
                 }, 0) / times.length;
-                source.off('tile.load', tileCounter);
+                source.off('data', tileCounter);
                 cb(null, avgTileTime);
             }
         }
--- a/bench/lib/set_data_perf.js
+++ b/bench/lib/set_data_perf.js
@@ -1,33 +1,29 @@
 'use strict';
 
-var NUM_TILES = 6;
-
-module.exports = function(source, numCalls, geojson, cb) {
-    var tileCount = 0;
+module.exports = function(sourceCache, data, callback) {
+    var sampleCount = 50;
     var startTime = null;
-    var times = [];
+    var samples = [];
 
-    source.on('data', function tileCounter(event) {
-        if (event.dataType !== 'tile') return;
-
-        tileCount++;
-        if (tileCount === NUM_TILES) {
-            tileCount = 0;
-            times.push(performance.now() - startTime);
-
-            if (times.length < numCalls) {
+    sourceCache.on('data', function onData() {
+        if (sourceCache.loaded()) {
+            samples.push(performance.now() - startTime);
+            sourceCache.off('data', onData);
+            if (samples.length < sampleCount) {
                 startTime = performance.now();
-                source.setData(geojson);
+                sourceCache.clearTiles();
+                sourceCache.on('data', onData);
+                sourceCache.getSource().setData(data);
             } else {
-                var avgTileTime = times.reduce(function (v, t) {
-                    return v + t;
-                }, 0) / times.length;
-                source.off('data', tileCounter);
-                cb(null, avgTileTime);
+                callback(null, average(samples));
             }
         }
     });
 
     startTime = performance.now();
-    source.setData(geojson);
+    sourceCache.getSource().setData(data);
 };
+
+function average(array) {
+    return array.reduce(function (sum, value) { return sum + value; }, 0) / array.length;
+}
--- a/js/source/image_source.js
+++ b/js/source/image_source.js
@@ -106,7 +106,7 @@ ImageSource.prototype = util.inherit(Evented, /** @lends ImageSource.prototype *
                 Math.round((zoomedCoord.row - centerCoord.row) * EXTENT));
         });
 
-        this.fire('source.change');
+        this.fire('data', {dataType: 'image'});
         return this;
     },
 
--- a/js/source/image_source.js
+++ b/js/source/image_source.js
@@ -53,7 +53,8 @@ function ImageSource(id, options, dispatcher) {
 
         this.image = image;
         this._loaded = true;
-        this.fire('source.load');
+        this.fire('data', {type: 'image'});
+        this.fire('sourceload');
 
         if (this.map) {
             this.setCoordinates(options.coordinates);
--- a/js/source/image_source.js
+++ b/js/source/image_source.js
@@ -54,7 +54,7 @@ function ImageSource(id, options, dispatcher) {
         this.image = image;
         this._loaded = true;
         this.fire('data', {type: 'image'});
-        this.fire('sourceload');
+        this.fire('source.load');
 
         if (this.map) {
             this.setCoordinates(options.coordinates);
--- a/js/source/image_source.js
+++ b/js/source/image_source.js
@@ -53,7 +53,7 @@ function ImageSource(id, options, dispatcher) {
 
         this.image = image;
         this._loaded = true;
-        this.fire('data', {type: 'image'});
+        this.fire('data', {dataType: 'source'});
         this.fire('source.load');
 
         if (this.map) {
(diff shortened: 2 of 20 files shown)


lucaswoj:
# Benchmarks

## map-load

**master 24f2135:** 119 ms
**datastart 1ef4213:** 105 ms

## style-load

**master 24f2135:** 109 ms
**datastart 1ef4213:** 113 ms

## buffer

**master 24f2135:** 961 ms
**datastart 1ef4213:** 941 ms

## fps

**master 24f2135:** 60 fps
**datastart 1ef4213:** 60 fps

## frame-duration

**master 24f2135:** 6.9 ms, 0% > 16ms
**datastart 1ef4213:** 7.2 ms, 1% > 16ms

## query-point

**master 24f2135:** 1.04 ms
**datastart 1ef4213:** 1.12 ms

## query-box

**master 24f2135:** 85.34 ms
**datastart 1ef4213:** 85.61 ms

## geojson-setdata-small

**master 24f2135:** 13 ms
**datastart 1ef4213:** 14 ms

## geojson-setdata-large

**master 24f2135:** 125 ms
**datastart 1ef4213:** 121 ms

jfirebaugh:
Can you post a full rundown of the new taxonomy -- what are the valid combinations of event types and `dataType` values, and when are they emitted?

lucaswoj:
Below is the type specification for the `data` and `dataloading` events:

```js
interface DataEvent {
    type: 'dataloading' | 'data';
    dataType: (
        'geoJSON' |
        'tileJSON' |
        'style' |
        'sprite' |
        'image' |
        'video' |
        'tile'
    );
}
```

The `Map#data` event is fired when any map data loads or changes.

The `Map#dataloading` event is fired when any map data begins loading or changing asynchronously, to be followed by a `data` event when the operation is complete.

jfirebaugh:
But there's also `source.load` and `style.load` still?

jfirebaugh:
What's the rationale for having separate `geoJSON`, `tileJSON`, `image`, and `video` values for `dataType`, rather than a single `source` value? It seems like a `source` value would simplify the taxonomy and keep it consistent with the domain objects used in rest of the API and in the style specification.

js/source/image_source.js
```diff
@@ -53,6 +53,7 @@ function ImageSource(id, options, dispatcher) {
 
         this.image = image;
         this._loaded = true;
+        this.fire('data', {type: 'image'});
```
jfirebaugh: `type` ⇢ `dataType`

js/source/raster_tile_source.js
```diff
@@ -17,6 +17,7 @@ function RasterTileSource(id, options, dispatcher) {
             return this.fire('error', err);
         }
         util.extend(this, tileJSON);
+        this.fire('data', {type: 'tileJSON'});
```
jfirebaugh: `type`  ⇢ `dataType`

lucaswoj:
> But there's also `source.load` and `style.load` still?

The plan is to have `data` subsume those events in a follow-up PR. See the plan in https://github.com/mapbox/mapbox-gl-js/issues/1715#issuecomment-249293404

> What's the rationale for having separate `geoJSON`, `tileJSON`, `image`, and `video` values for `dataType`, rather than a single `source` value? It seems like a `source` value would simplify the taxonomy and keep it consistent with the domain objects used in rest of the API and in the style specification.

The rationale was to have a `dataType` per specific network resource type. Merging `geoJSON`, `tileJSON`, `image`, and `video` into `source` is a good idea. I'll push a commit shortly.

bench/lib/set_data_perf.js
```diff
@@ -7,7 +7,9 @@ module.exports = function(source, numCalls, geojson, cb) {
     var startTime = null;
     var times = [];
 
-    source.on('tile.load', function tileCounter() {
+    source.on('data', function tileCounter(event) {
+        if (event.dataType !== 'tile') return;
```
jfirebaugh: Does this event binding need to distinguish between "tile loading" and "tile unloading"?

js/style/style.js
```diff
@@ -22,8 +22,9 @@ var getWorkerPool = require('../global_worker_pool');
 
 module.exports = Style;
 
-function Style(stylesheet, animationLoop, options) {
-    this.animationLoop = animationLoop || new AnimationLoop();
+function Style(stylesheet, map, options) {
+    this.map = map;
```
jfirebaugh: I've been trying to reduce or eliminate the dependency of other class on `Map`. The dependency here is required by `Source#on{Add,Remove}(map)` -- we should work towards eliminating those two methods.

js/ui/map.js
```diff
@@ -197,23 +197,11 @@ var Map = module.exports = function(options) {
         this.style.update(this._classes, {transition: false});
     });
 
-    this.on('style.change', function() {
+    this.on('data', function() {
         this._update(true);
```
jfirebaugh: Not every `data` event should trigger `_update(true)` -- that triggers updating and recalculating, which is necessary only in certain cases.

bench/lib/set_data_perf.js
lucaswoj: I think so. I will refactor to make this benchmark utility more robust.

js/style/style.js
lucaswoj: Removing the dependency was my first instinct as well. Unfortunately it wasn't straightforward to do so and will require a standalone PR. 

Even though the dependency is undesirable, it is a real dependency and should be passed explicitly and simply. As our code is architected, `Source`s need to know about the state of `Camera` (which is a mixin on `Map`). I see #2741 as the proper long-term fix.

js/ui/map.js
lucaswoj: Good catch. Fortunately it looks like filtering on `event.dataType === 'style'` is equivalent to listening to the `style.change` event.

js/source/raster_tile_source.js
lucaswoj: Fixed

js/source/image_source.js
lucaswoj: Fixed

lucaswoj:
All feedback addressed, ready for a second pass of  👀 @jfirebaugh

js/ui/control/attribution.js
```diff
@@ -54,9 +54,7 @@ Attribution.prototype = util.inherit(Control, {
             container = this._container = DOM.create('div', className, map.getContainer());
 
         this._update();
-        map.on('source.load', this._update.bind(this));
-        map.on('source.change', this._update.bind(this));
-        map.on('source.remove', this._update.bind(this));
+        map.on('data', this._update.bind(this));
```
jfirebaugh: Add a `dataType === 'source'` guard in `_update` so that it doesn't mutate the DOM more frequently than it did before.

lucaswoj: Done.

lucaswoj:
Thank you for the review @jfirebaugh 🙇

strech345:
is there a 'unload' function?
i need the event which tile is unload.

lucaswoj:
@strech345 Not right now. Defining what it means for a tile to become "unloaded" isn't easy. Is a tile unloaded

 - When it is no longer at the desired zoom level for the current viewport but still rendered?
 - When it is entirely outside of the current viewport?
 - When it is removed from the cache? How big is the cache? Which cache?
 - When it is reloaded?

strech345:
@lucaswoj it the same question like for the loading. So i think it should be a equivalent to the dataloading

PR #3257, opened 2016-09-26
Update documentation, add instructions for Typescript definitions.

filoxo:
The Typescript definitions were contributed to DefinitelyTyped (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/11386), and were then pushed to `@types` npm namespace. 

This PR documents how to install the typings using `typings` as well as `@types` on npm for mapbox-gl-js.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,16 @@ import mapboxgl from 'mapbox-gl/dist/mapbox-gl.js';
 
 These projects are written and maintained by the GL JS community. Feel free to open a PR add your own projects to this list. We :heart: third party projects!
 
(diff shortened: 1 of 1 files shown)


lucaswoj:
https://github.com/mapbox/mapbox-gl-js/commit/335a0208e3293d910cf558dac0a9d129b50ab995 retains the URL switch to DefinatelyTyped and removes the installation instructions. I would not like this documentation to be a source of truth about how to use the TypeScript tooling.

filoxo:
Fair enough 😄

PR #3259, opened 2016-09-27
Serialize 'tiles' in raster tile source

AndyMoreland:
It seems like the raster tile source `serialize` method omits serializing the `tiles` field. I do not understand why this would be desirable, so my guess is it's a bug. If this is not intentional, this PR addresses the problem.

This popped up when I was trying to write save-and-restore functionality for debugging my application and I was losing my raster tiles each time.


lucaswoj:
This looks like an omission on our part. Thanks @AndyMoreland 🙇

PR #3260, opened 2016-09-27
pass tile sourceMaxZoom value to tile url

RipRyness:
This fixes an issue when using a maxzoom level above the tile source's available maxzoom level (over-zooming).  Currently when over-zooming a (raster) tile source all tiles at the over zoom levels disappear. As well as requests are made to the tile source server that results in 404 errors

### DIFF
--- a/js/source/raster_tile_source.js
+++ b/js/source/raster_tile_source.js
@@ -42,7 +42,7 @@ RasterTileSource.prototype = util.inherit(Evented, {
     },
 
     loadTile: function(tile, callback) {
-        var url = normalizeURL(tile.coord.url(this.tiles, tile.sourceMaxZoom, this.scheme), this.url, this.tileSize);
+        var url = normalizeURL(tile.coord.url(this.tiles, this.maxzoom, this.scheme), this.url, this.tileSize);
 
         tile.request = ajax.getImage(url, done.bind(this));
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
`VectorTileSource` [passes `this.maxzoom`](https://github.com/mapbox/mapbox-gl-js/blob/335a0208e3293d910cf558dac0a9d129b50ab995/js/source/vector_tile_source.js#L49-L49) at this point; `RasterTileSource` should probably follow.

We have [a test case that should have caught this](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/zoomed-raster/overzoom) and should investigate why it didn't.

RipRyness:
That makes sense.  I made the change and confirmed that it fixes the issue.

lucaswoj:
Thanks for moving forward on this @RipRyness. Before we can merge the PR, we'll need to investigate why the [`zoomed-raster/overzoom`](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/zoomed-raster/overzoom) test didn't catch this bug and/or create a test case that does catch this bug. Let me know if you need any help with that step.

RipRyness:
It seems that this is no longer an issue and possibly never was as I'm not able to reproduce it on any version back to 0.22.

PR #3263, opened 2016-09-27
Fix NPEs in raster tile source when destroying map with tiles loading

AndyMoreland:
Hi,

I've observed two different NPEs triggered from the `RasterTileSource` when I `map.remove` a map that is still loading.

The first npe occurs because `gl.createTexture()` can silently return `null` if the webgl context has been closed.

I am not sure why the second NPE occurs, but my change fixed it.

Neither of my fixes are great and I am sure that they can be obviated by someone with a better understanding of the teardown process for mapbox-gl-js, but they did solve my immediate problem so I figured I'd flag the issues.

Andy

### DIFF
--- a/js/source/raster_tile_source.js
+++ b/js/source/raster_tile_source.js
@@ -66,6 +66,9 @@ RasterTileSource.prototype = util.inherit(Evented, {
                 gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, img);
             } else {
                 tile.texture = gl.createTexture();
+                if (!tile.texture) {
+                    return callback(null);
+                }
                 gl.bindTexture(gl.TEXTURE_2D, tile.texture);
                 gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR_MIPMAP_NEAREST);
                 gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
@@ -78,7 +81,9 @@ RasterTileSource.prototype = util.inherit(Evented, {
             gl.generateMipmap(gl.TEXTURE_2D);
 
             tile.timeAdded = new Date().getTime();
-            this.map.animationLoop.set(this.style.rasterFadeDuration);
+            if (this.map.style != null) {
+                this.map.animationLoop.set(this.map.style.rasterFadeDuration);
+            }
 
             tile.source = this;
             tile.loaded = true;


AndyMoreland:
I looked at other calls to `gl.createTexture` and I suspect that there may be more NPEs lurking that I don't hit in my usage, so it might be worth auditing those.

js/source/raster_tile_source.js
```diff
@@ -66,6 +66,9 @@ RasterTileSource.prototype = util.inherit(Evented, {
                 gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, img);
             } else {
                 tile.texture = gl.createTexture();
+                if (!tile.texture) {
+                    return callback(null);
+                }
```
lucaswoj: > The first npe occurs because gl.createTexture() can silently return null if the webgl context has been closed.

I would have thought we were aborting all outstanding tile requests during a `Map#remove` operation but that does not seem to be the case. I would go about fixing this bug by adding `source.clearTiles()` to `Style#removeSource`. 

Without a demonstration of the bug, it is impossible for me to know for sure.

js/source/raster_tile_source.js
```diff
@@ -78,7 +81,9 @@ RasterTileSource.prototype = util.inherit(Evented, {
             gl.generateMipmap(gl.TEXTURE_2D);
 
             tile.timeAdded = new Date().getTime();
-            this.map.animationLoop.set(this.style.rasterFadeDuration);
+            if (this.map.style != null) {
+                this.map.animationLoop.set(this.map.style.rasterFadeDuration);
+            }
```
lucaswoj: I suspect the fix proposed above 👆 will fix this case too.

Without a demonstration of the bug, it is impossible for me to know for sure.

lucaswoj:
Thank you @AndyMoreland! Would you mind putting together a quick [jsbin](http://jsbin.com/) (or similar) that demonstrates the exception?

These null pointer exceptions appear to be a symptom of another bug: we do not abort outstanding tile requests during the `Map#remove` operation.

AndyMoreland:
I will try to do that tonight.

AndyMoreland:
Actually, this repros it for me: http://jsbin.com/zexefagebu/edit?html,js,output

AndyMoreland:
It's timing dependent of course (since if your raster tiles finish loading in <1 ms you'll not get an exception), but if I have my js console open and hit "run with js" on that page it breaks in the appropriate way for me.

lucaswoj:

I think this bug has since been fixed 👍

AndyMoreland:
Yes, it was fixed by the change you suggested. Apologies -- I should have
closed this.


Andy

Issue #3253, opened 2016-09-26
Custom SVG in Control elements


recound:
I'd like to be able to change SVG for the zoom in, zoom out, and compass UI controls.

mourner:
You can change the SVG by overriding this CSS code: https://github.com/mapbox/mapbox-gl-js/blob/master/dist/mapbox-gl.css#L73-L92

recound:
Thanks for cooperation. I suppose, the solution assumes to fork my changes in separate repo?

mourner:
No, you can just override `background-image` with external CSS.

PR #3268, opened 2016-09-28
Replace div in Navigation control with span

### LINKED ISSUE #3089: Replace <div> in compass control with an inline element
**mapbox-gl-js version**:

### Steps to Trigger Behavior

 1. Load mapbox

### Expected Behavior
The button should have an image as a child

### Actual Behavior
The button has a div as a child

image

colleenmcginnis:
Fixes #3089 

Changes the icons in the Navigation control that are currently inside `div` elements to be ~~`img`~~ `span` elements.

### DIFF
--- a/dist/mapbox-gl.css
+++ b/dist/mapbox-gl.css
@@ -65,7 +65,7 @@
     background-color: rgba(0,0,0,0.05);
 }
 .mapboxgl-ctrl-icon,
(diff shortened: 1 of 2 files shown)


jfirebaugh:
I think we should keep using `background-image` so that our recommendation in #3253 still applies. We can switch `div` ⇢ `span` instead.

PR #3270, opened 2016-09-29
ensure no subpixel transforms on marker elements

mollymerp:
## Launch Checklist

While subpixel transforms turned out not to be the issue on popups (#3249 ), they are an issue with markers. Here I just round the `Point` pos for the marker (like we do with [`Popup`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/popup.js#L231) to prevent the blurriness it causes. 

_subpixel transform_
screen shot 2016-09-28 at 5 16 27 pm

_rounded pixel transform_
screen shot 2016-09-28 at 5 17 03 pm

cc @jfirebaugh @lucaswoj 

 - [ ] post benchmark scores

### DIFF
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -43,9 +43,9 @@ Marker.prototype = {
         this.remove();
         this._map = map;
         map.getCanvasContainer().appendChild(this._element);
-        map.on('move', this._update);
-        map.on('moveend', this._update);
-        this._update();
+        map.on('move', this._update.bind(this));
+        map.on('moveend', this._update.bind(this, {roundPos: true}));
+        this._update({roundPos: true});
 
         // If we attached the `click` listener to the marker element, the popup
         // would close once the event propogated to `map` due to the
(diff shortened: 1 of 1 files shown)


mourner:
As far as I remember, I didn't add rounding because markers noticeably wiggled when zooming. Is this the case now? Maybe we should only round the position on moveend.

mollymerp:
@mourner -- thanks for flagging. I didn't remember that detail about the shaky zooming. I modified so it only rounds the transform on moveend.

js/ui/marker.js
```diff
@@ -148,9 +149,9 @@ Marker.prototype = {
         else popup.addTo(this._map);
     },
 
-    _update: function () {
+    _update: function (e) {
         if (!this._map) return;
-        var pos = this._map.project(this._lngLat)._add(this._offset);
+        var pos =  (e && e.type === "move") ? this._map.project(this._lngLat)._add(this._offset) : this._map.project(this._lngLat)._add(this._offset).round();
```
lucaswoj: This line could be a little more DRY

```
var pos = this._map.project(this._lngLat)._add(this._offset);
if (options.roundPos) pos = pos.round();
```

js/ui/marker.js
```diff
@@ -44,6 +44,7 @@ Marker.prototype = {
         this._map = map;
         map.getCanvasContainer().appendChild(this._element);
         map.on('move', this._update);
+        map.on('moveend', this._update);
         this._update();
```
lucaswoj: We should round the coords upon marker initialization too. 

Perhaps we could do this by adding a `roundPos` option to `Marker#_update`
```js
map.on('move', this._update.bind(this));
map.on('moveend', this._update.bind(this, {roundPos: true}));
this._update({roundPos: true});
```

mollymerp: As the PR stands, it does round coords on initialization (only keeps un-rounded coords on `move`). We can add this option in the interest of more explicit/understandable code 👍

lucaswoj:
Looks good! 🚢 on 🍏

Issue #3277, opened 2016-09-30
Attribution control does not render in a basic map


ajithranka:
In the latest version of GL, the attribution control does not render until at least one source is added to the map. Setting `attributionControl` to `true` in the map options has no effect.

**demo**: http://jsbin.com/cikobewuma/edit?js,output
**mapbox-gl-js version**: 0.25.0

mourner:
In fact, all controls are not displayed. Investigating.

mourner:
The bug happens because `data` listener in the attribution control fires before `source.load` listener in `SourceCache` (which copies the attribution string from underlying `Source`). It means that `map.sources.<source>.attribution` is empty at the time while `map.sources.<source>._source.attribution` isn't.

@lucaswoj we can either do `getSource` in `Attribution.createAttributionString`, or fiddle with the order of events. What approach do you prefer?

lucaswoj:
Closed in #3278

Issue #2957, opened 2016-08-09
Write unit test guidelines


lucaswoj:
Taking inspriation from @springmeyer's guidelines for Mapnik, ["A perfect testcase"](https://github.com/mapnik/mapnik/wiki/A-perfect-testcase), we should write guidelines for writing GL JS unit tests. 

cc @mapsam @springmeyer @mollymerp

PR #3278, opened 2016-09-30
Fix attribution regression

mourner:
This is the simplest fix for https://github.com/mapbox/mapbox-gl-js/issues/3277. 

However, to avoid things like this in future, we need to rewrite our tests to use less primitive mocks — this is the reason attribution disappearance was not caught by existing attribution tests, and this is a more challenging task.

cc @lucaswoj @mollymerp @ajithranka

### DIFF
--- a/js/render/draw_background.js
+++ b/js/render/draw_background.js
@@ -7,7 +7,7 @@ var tileSize = 512;
 
 module.exports = drawBackground;
 
-function drawBackground(painter, source, layer) {
+function drawBackground(painter, sourceCache, layer) {
     var gl = painter.gl;
     var transform = painter.transform;
     var color = layer.paint['background-color'];
--- a/js/render/draw_circle.js
+++ b/js/render/draw_circle.js
@@ -4,7 +4,7 @@ var browser = require('../util/browser');
 
 module.exports = drawCircles;
 
-function drawCircles(painter, source, layer, coords) {
+function drawCircles(painter, sourceCache, layer, coords) {
     if (painter.isOpaquePass) return;
 
     var gl = painter.gl;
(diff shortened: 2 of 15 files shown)


lucaswoj:
Thanks for pulling together this fix @mourner! 

I can see a couple ways to interpret the root cause of this bug:

 - `Style#sources` would be more appropriately named `Style#sourceCaches`
 - `SourceCache` shouldn't concern itself with proxying `Source` properties, they should be read from the source directly (more state, more problems)
 - we shouldn't have both `source.loaded` and `data` events (I plan to combine these in the near future and I'm trying to keep the scope of each PR small) 

> However, to avoid things like this in future, we need to rewrite our tests to use less primitive mocks — this is the reason attribution disappearance was not caught by existing attribution tests, and this is a more challenging task.

I agree. Tests like [this one](http://github.com/mapbox/mapbox-gl-js/blob/0240e7482a7582fd35354c762545bb11461ab5ad/test/js/ui/control/attribution.test.js#L34-L43) should be rewritten to use real domain objects. ref #2957

lucaswoj:
I'm going to invest a little time in addressing the root causes I identified above and improving the unit tests ☝️ 

# Todo

 - [ ] Improve attribution unit tests 
 - [ ] Rename `Style#sources` to `Style#sourceCaches`
 - [ ] Refactor `Attribution` to update attribution on `data` event rather than `source.load` event

lucaswoj:
Just pushed up some refactorings that should prevent similar bugs in the future. In particular, I was looking to address confusion about what values are `Source`s and which are `SourceCache`s. 

cc @mourner @jfirebaugh

PR #3281, opened 2016-09-30
Fix text fields with EOL at the end

### LINKED ISSUE #3254: Mapbox Studio Custom Tiles not working
**mapbox-gl-js version 0.24.0**:

I am creating a webmap using MapboxGL.js and just updated from 0.21.0 to 0.24.0 for the scale bar functionality. I am using custom vector tiles that I generated using Mapbox studio and rendering information client side with geojson. However, after updating to the most recent release I am now getting the following error:
>Uncaught TypeError: Cannot read property 'advance' of undefined

Some additional testing found that Mapbox styles such as Outdoor do not throw any errors. 
My tiles end up looking like so: 
image

From the same studio account using Mapbox's Outdoor Style (same access token):
image

### Steps to Trigger Behavior

 1. Using Customized Studio Tiles
 2. Draw Map using Custom Tiles via MapboxGL JS
 3. AddLayers from geojson over 

### Expected Behavior
Customized Vector Tiles from Studio would be fetched and drawn before adding additional geojson layers client side.
### Actual Behavior
Vector Tiles are unable to be rendered

mourner:
Closes #3254. cc @lucaswoj

### DIFF
--- a/test/js/symbol/shaping.test.js
+++ b/test/js/symbol/shaping.test.js
@@ -58,8 +58,9 @@ test('shaping', function(t) {
     t.equal(false, shaped);
 
     // https://github.com/mapbox/mapbox-gl-js/issues/3254
-    shaped = shaping.shapeText('foo bar\n', glyphs, 15 * oneEm, oneEm, 0.5, 0.5, 0.5, 0 * oneEm, [0, 0]);
-    t.ok(shaped);
+    shaped = shaping.shapeText('   foo bar\n', glyphs, 15 * oneEm, oneEm, 0.5, 0.5, 0.5, 0 * oneEm, [0, 0]);
+    var shaped2 = shaping.shapeText('foo bar', glyphs, 15 * oneEm, oneEm, 0.5, 0.5, 0.5, 0 * oneEm, [0, 0]);
+    t.same(shaped.positionedGlyphs, shaped2.positionedGlyphs);
 
     t.end();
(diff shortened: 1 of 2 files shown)


test/js/symbol/shaping.test.js
```diff
@@ -57,5 +57,9 @@ test('shaping', function(t) {
     shaped = shaping.shapeText(String.fromCharCode(0), glyphs, 15 * oneEm, oneEm, 0.5, 0.5, 0.5, 0 * oneEm, [0, 0]);
     t.equal(false, shaped);
 
+    // https://github.com/mapbox/mapbox-gl-js/issues/3254
+    shaped = shaping.shapeText('foo bar\n', glyphs, 15 * oneEm, oneEm, 0.5, 0.5, 0.5, 0 * oneEm, [0, 0]);
+    t.ok(shaped);
```
lucaswoj: What does this assertion test? Have you considered a writing a `mapbox-gl-test-suite` rendering test for this case?

mourner: Previously, this line of code would throw. :) Updated the test so that it's more self-documenting. I don't think test-suite test is necessary for this case because `\n` at the end should be equivalent to shaping without it.

lucaswoj: The new test is great! Clever idea. :shipit:

PR #3283, opened 2016-10-03
fix marker update with no options

runetvilum:
## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -152,7 +152,7 @@ Marker.prototype = {
     _update: function (options) {
         if (!this._map) return;
         var pos = this._map.project(this._lngLat)._add(this._offset);
-        if (options.roundPos) pos = pos.round();
+        if (options && options.roundPos) pos = pos.round();
         DOM.setTransform(this._element, 'translate(' + pos.x + 'px,' + pos.y + 'px)');
     }
 };


jfirebaugh:
cc @mollymerp -- looks like this fixes `Marker#setLatLng` when the marker is already on the map. Also, the case where `_update` is called via `map.on('move', this._update.bind(this))` works by accident, because `options` is actually an event object without a `roundPos` property. Let's use this as an opportunity to get full code coverage on marker.js.

mollymerp:
Thanks for the fix @runetvilum !

@jfirebaugh yes, I will work on some tests to get coverage up 😳 . 

My original approach used the `event` argument to determine whether to round (`633b986`) :
```js
var pos =  (e && e.type === "move") ? this._map.project(this._lngLat)._add(this._offset) : this._map.project(this._lngLat)._add(this._offset).round();
```

I'm thinking it might make sense to go back to that approach (DRY version) instead of adding an options parameter. 
```js
var pos = this._map.project(this._lngLat)._add(this._offset);
if (!e || e.type==="moveend") pos = pos.round();
```

cc @lucaswoj

mollymerp:
Addressing this in #3294

mollymerp:
Thanks again for contributing @runetvilum -- we decided to change the approach we're using to round the position coordinates and have merged a fix in #3294.

PR #3294, opened 2016-10-04
fix bug causing error when Marker#setLngLat is called on a marker that's already on the map

mollymerp:
## Launch Checklist

 - [ ] post benchmark scores

also adds a test that would have caught the bug.

I decided to go back to the approach where we don't add in an extra function argument to the `_update` function. IMO this reduces unnecessary complexity bc the `Evented` object that gets passed gives us all the information we need to produce the desired functionality. This code is less explicit than with `roundedPos`, but I added a detailed comment to compensate for this. 

Thanks @runetvilum for catching this bug!  ref #3283

cc @jfirebaugh @lucaswoj

### DIFF
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -43,9 +43,9 @@ Marker.prototype = {
         this.remove();
         this._map = map;
         map.getCanvasContainer().appendChild(this._element);
-        map.on('move', this._update.bind(this));
-        map.on('moveend', this._update.bind(this, {roundPos: true}));
-        this._update({roundPos: true});
+        map.on('move', this._update);
+        map.on('moveend', this._update);
+        this._update();
 
         // If we attached the `click` listener to the marker element, the popup
         // would close once the event propogated to `map` due to the
@@ -66,6 +66,7 @@ Marker.prototype = {
         if (this._map) {
             this._map.off('click', this._onMapClick);
             this._map.off('move', this._update);
+            this._map.on('moveend', this._update);
             this._map = null;
         }
         DOM.remove(this._element);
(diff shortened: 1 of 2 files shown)


js/ui/marker.js
```diff
@@ -66,6 +66,7 @@ Marker.prototype = {
         if (this._map) {
             this._map.off('click', this._onMapClick);
             this._map.off('move', this._update);
+            this._map.on('moveend', this._update);
```
jfirebaugh: `on` ⇢ `off`

test/js/ui/marker.test.js
```diff
@@ -29,6 +29,14 @@ test('Marker', function (t) {
         t.end();
     });
 
+    t.test('marker\'s lngLat can be changed', function (t) {
+        var map = createMap();
+        var marker = new Marker(window.document.createElement('div')).setLngLat([-77.01866, 38.888]).addTo(map);
+        t.ok(marker.setLngLat([-76,39]) instanceof Marker, 'marker.setLngLat() returns Marker instance');
+        t.ok(marker._lngLat.lng === -76 &&  marker._lngLat.lat === 39, 'marker\'s position can be updated');
```
jfirebaugh: Use `marker.getLngLat()` rather than a private property.

ajfickas:
@mollymerp When will this fix be officially released?

I've got an app we've been migrating to Mapbox GL JS which we're planning to release hopefully this week and this fix looks like it will address a few show-stopper bugs we're seeing. Specifically, `marker.setLngLat` throwing an error when called during a map move event (to allow dragging a marker on the map), and markers "jiggling" when panning the map.

(Didn't see a release schedule out there, so this was the best place I found to ask. Let me know if there a better place to get an answer!)

mollymerp:
@ajfickas sorry for the inconvenience! 

We track our releases at www.github.com/mapbox/mapbox-gl-js/milestones and our next one is scheduled for this Wednesday. There is a chance we won't get it out on Wednesday if there are outstanding release blockers, but we are also eager to get another release out ASAP.

ajfickas:
Thanks for the quick response, that answers my question! I'll stay tuned!

Issue #3247, opened 2016-09-23
Document "resize" event


lucaswoj:
Per #2232 (comment) we need to document or remove the "resize" event

http://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/map.js#L334-L335

colleenmcginnis:
fixed in #3295

Issue #3246, opened 2016-09-23
Document "remove" event 


lucaswoj:
Per https://github.com/mapbox/mapbox-gl-js/issues/2232#issuecomment-249243876 we need to document or remove the "remove" event 

http://github.com/mapbox/mapbox-gl-js/blob/24f2135f058372bd7906e3c09148c92c8ffebef0/js/ui/map.js#L1088

jfirebaugh:
Added in 9def6df1f98222c068813dfcd56f8b50fa553bce. Is it used anywhere? I couldn't find any uses.

jfirebaugh:
Oh, [here](https://github.com/mapbox/mapbox-gl-js/blob/9def6df1f98222c068813dfcd56f8b50fa553bce/bench/lib/create_map.js#L23-L23).

colleenmcginnis:
fixed in #3295

PR #3295, opened 2016-10-04
Document remove and resize events

colleenmcginnis:
Documents remove and resize events in the API docs to address #3246 and #3247 

cc @mollymerp @lucaswoj

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -1419,7 +1419,7 @@ function removeNode(node) {
   */
 
  /**
- * Fired immediately after the map container has been removed with [`Map#remove`](#Map#remove).
+ * Fired immediately after the map has been removed with [`Map#remove`](#Map#remove).
  *
  * @event remove
  * @memberof Map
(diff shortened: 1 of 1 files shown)


js/ui/map.js
```diff
@@ -1417,3 +1417,19 @@ function removeNode(node) {
   * @property {string} type The event type.
   * @property {string} dataType The type of data that has changed. One of `'source'`, `'style'`, or `'tile'`.
   */
+
+ /**
+ * Fired immediately after the map container has been removed with [`Map#remove`](#Map#remove).
```
lucaswoj: `Map#remove` doesn't remove the `container` (the HTML element that contains the map canvas & controls). I suggest changing this to

 >  Fired immediately after the map ~~container~~ has been removed with [`Map#remove`](#Map#remove)

js/ui/map.js
```diff
@@ -1417,3 +1417,19 @@ function removeNode(node) {
...
+ *
+ * @event remove
+ * @memberof Map
+ * @instance
+ */
+
+  /**
+ * Fired immediately after the map container has been resized.
```
lucaswoj: Probably best to remove "container" for similar reasons as above.

mollymerp:
Thanks @colleenmcginnis! 🎈

Issue #3076, opened 2016-08-26
Combine "mapbox-gl-plugins" repo and "docs/_posts/plugins"


lucaswoj:
We should combine these two sources of truth about GL JS plugins, simplifying the plugin publishing workflow and reducing the potential for error. 

As a bonus, we may be able to automatically publish plugins to `s3` from `publisher` when the `mb-pages` branch is deployed.

 - ["mapbox-gl-plugins" repo](https://github.com/mapbox/mapbox-gl-plugins)
 - ["docs/_posts/plugins"](https://github.com/mapbox/mapbox-gl-js/tree/master/docs/_posts/plugins)

mourner:
This was done in #3300.

PR #3300, opened 2016-10-05
Create single source of truth for plugin versions

mollymerp:
Now, whenever a new version of a plugin in https://github.com/mapbox/mapbox-gl-plugins is released, we will only need to update the version in `/docs/_data/plugins.json` to change the URL version on both the plugins page and the examples page. 

cc @ghoshkaj @mcwhittemore @tristen @lucaswoj 

## Launch Checklist

 - [ ] post benchmark scores


tristen:
Awesome stuff! 

- Need to also include https://github.com/mapbox/mapbox-gl-js/blob/master/docs/_posts/examples/3400-01-16-point-from-geocoder-result.html 
- It's only 50% related but it would be nice to hit up https://github.com/mapbox/mapbox-gl-js/issues/3076 sooonn

lucaswoj:
@mollymerp has graciously agreed to expand the scope of this PR to include https://github.com/mapbox/mapbox-gl-js/issues/3076 🙌

mollymerp:
Updated to cover #3076 as well -- once this 🚢 we should be able to delete that repo.

package.json
```diff
@@ -104,6 +104,8 @@
     "start-docs": "npm run build-min && npm run build-docs && jekyll serve --watch",
     "lint": "eslint  --ignore-path .gitignore js test bench docs/_posts/examples/*.html",
     "open-changed-examples": "git diff --name-only mb-pages HEAD -- docs/_posts/examples/*.html | awk '{print \"http://127.0.0.1:4000/mapbox-gl-js/example/\" substr($0,33,length($0)-37)}' | xargs open",
+    "plugin-test":"tap plugins/test",
```
lucaswoj: We should keep all our tests within `test`. How about `test/plugins`?

lucaswoj: When will these tests get run? If they're quick, we could run them whenever `npm test` is called.

docs/_posts/plugins/0100-01-01-mapbox-gl-compare.html
```diff
@@ -8,6 +8,8 @@
   - isc
 code: https://github.com/mapbox/mapbox-gl-compare
 license: ISC
+suffix: compare
+css: true
+js: true
```
lucaswoj: We may need to go back to the explicit markup if we hever have a 3rd party plugin that:

 - requires extra boilerplate
 - is hosted in a different way 
 - uses a different naming convention

lucaswoj: We might. And we might want to document plugins that are hosted other places.

mourner:
Is it absolutely necessary to include all plugin versions with their assets in the gl-js repository? It doesn't feel quite right to me... Could we use CDN plugin links instead?

lucaswoj:
It is not necessary for us to include all plugin versions & assets in the gl-js repository. Doing so was the simplest way to merge the GL JS and GL Plugins repositories. I'd be excited about follow-up issues / PRs that move to a more elegant system.

PR #3302, opened 2016-10-06
Fix a bug preventing map.loaded() from returning true with raster tiles.

vicapow:
## Launch Checklist

* Uses a transparent png when there's no content from raster tiles.
* Removes unused `return img;` because `getArrayBuffer` does nothing with the returned value from it's callback.
* properly sets the tile state when the request is aborted or errors.

### DIFF
--- a/js/source/raster_tile_source.js
+++ b/js/source/raster_tile_source.js
@@ -54,10 +54,13 @@ class RasterTileSource extends Evented {
         function done(err, img) {
             delete tile.request;
 
-            if (tile.aborted)
-                return;
+            if (tile.aborted) {
+                this.state = 'unloaded';
+                return callback(null);
+            }
 
             if (err) {
+                this.state = 'errored';
                 return callback(err);
             }
 
--- a/js/util/ajax.js
+++ b/js/util/ajax.js
@@ -34,6 +34,9 @@ exports.getArrayBuffer = function(url, callback) {
         callback(e);
     };
     xhr.onload = function() {
+        if (xhr.response.byteLength === 0 && xhr.status === 200) {
+            return callback(new Error('http status 200 returned without content.'));
+        }
         if (xhr.status >= 200 && xhr.status < 300 && xhr.response) {
             callback(null, xhr.response);
         } else {
@@ -50,6 +53,8 @@ function sameOrigin(url) {
     return a.protocol === window.document.location.protocol && a.host === window.document.location.host;
 }
 
+const transparentPngUrl = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAAC0lEQVQYV2NgAAIAAAUAAarVyFEAAAAASUVORK5CYII=';
+
 exports.getImage = function(url, callback) {
     return exports.getArrayBuffer(url, (err, imgData) => {
         if (err) return callback(err);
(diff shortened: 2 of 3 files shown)


js/util/ajax.js
```diff
@@ -59,7 +61,11 @@ exports.getImage = function(url, callback) {
             (window.URL || window.webkitURL).revokeObjectURL(img.src);
         };
         var blob = new window.Blob([new Uint8Array(imgData)], { type: 'image/png' });
-        img.src = (window.URL || window.webkitURL).createObjectURL(blob);
+        if (imgData.byteLength) {
+            img.src = (window.URL || window.webkitURL).createObjectURL(blob);
+        } else {
+            img.src = transparentPngUrl;
```
vicapow: If we don't do this, the tile becomes black and webGL warnings are logged.

js/source/raster_tile_source.js
```diff
@@ -51,10 +51,13 @@ RasterTileSource.prototype = util.inherit(Evented, {
         function done(err, img) {
             delete tile.request;
 
-            if (tile.aborted)
-                return;
+            if (tile.aborted) {
+                this.state = 'unloaded';
+                return callback(null);
+            }
```
lucaswoj: Is this fixing another bug? If so, thank you! Out of curiosity: what were the symptoms of this bug? I'm guessing something along the lines of "`map.loaded()` doesn't return true if there are any tiles were aborted."

js/util/ajax.js
```diff
@@ -59,7 +61,11 @@ exports.getImage = function(url, callback) {
             (window.URL || window.webkitURL).revokeObjectURL(img.src);
         };
         var blob = new window.Blob([new Uint8Array(imgData)], { type: 'image/png' });
-        img.src = (window.URL || window.webkitURL).createObjectURL(blob);
+        if (imgData.byteLength) {
```
lucaswoj: Why did you decide to branch on `imgData.byteLength` rather than "status is 204 No Content"?

js/source/raster_tile_source.js
vicapow: Yes, technically, it's a seperate bug that manifested itself in the same way (preventing loaded() from being true.)

>  I'm guessing something along the lines of "map.loaded() doesn't return true if there are any tiles were aborted."

Yes, exactly.

vicapow: > I'm interested in working with you to fix one of those two blockers

Happy to help with some guidance on how best to write tests for this.

js/util/ajax.js
lucaswoj: According to the w3c spec, a 204 response may not include any body

> ## 10.2.5 204 No Content
> ...
> 
> The 204 response MUST NOT include a message-body, and thus is always terminated by the first empty line after the header fields.

My intuition is that we should keep the behaviour you have implemented in response to a 204 header and fail loudly if any other type of response contains no data.

lucaswoj: > 10.2.1 200 OK

> The request has succeeded. The information returned with the response is dependent on the method used in the request, for example:
> 
> GET an entity corresponding to the requested resource is sent in the response;
> ...

My interpretation is that a 200 response must include a body. Per the spec, a trivial transparent PNG would be acceptable but the lack of a body would be unacceptable. In practice, a 200 response with no body is more likely a bug than intended behaviour.

js/util/ajax.js
```diff
@@ -34,6 +34,9 @@ exports.getArrayBuffer = function(url, callback) {
         callback(e);
     };
     xhr.onload = function() {
+        if (xhr.response.byteLength === 0 && xhr.status === 200) {
```
vicapow: specifically, this bit

js/util/ajax.js
lucaswoj: LGTM! Now we need to figure out a way to test this code path.

vicapow: @lucaswoj any ideas how best to do that?

lucaswoj: It is unfortunate that we don't have any unit tests for this module yet 😬. I suggest that we create a `test/js/util/ajax.test.js` file and add at least a test for this case. Sinon's [`window. useFakeXMLHttpRequest`](http://sinonjs.org/docs/#server) should make it relatively easy to simulate this case.

lucaswoj:
This is ready to merge once there is a unit test per https://github.com/mapbox/mapbox-gl-js/pull/3302/files#r83751352 😄

mourner:
And also needs another rebase after #3408.

vicapow:
@lucaswoj okay! I've rebased from master and added tests for `ajax.getArrayBuffer()`

PR #3306, opened 2016-10-06
Fire "dataloading" events from all resources that fire "data" events

lucaswoj:
This PR completes our support for "dataloading" events, ensuring they are fired from all resources that fire "data" events.

## Launch Checklist

### DIFF
--- a/js/util/evented.js
+++ b/js/util/evented.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var util = require('./util');
-var window = require('./window');
 
 /**
  * Methods mixed in to other classes for event capabilities.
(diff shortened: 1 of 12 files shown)


lucaswoj:
cc @jfirebaugh @mollymerp @lbud for review

js/util/evented.js
```diff
@@ -95,6 +96,18 @@ var Evented = {
...
     /**
+     * Asyncronously fires an event of the specified type. The event will be
+     * fired within a "setTimeout(..., 0)" callback.
+     *
+     * @param {string} type The type of event to fire.
+     * @param {Object} [data] Data to be passed to any listeners.
+     * @returns {Object} `this`
+     */
+    asyncFire: function(type, data) {
+        window.setTimeout(this.fire.bind(this, type, data), 0);
```
lucaswoj: @jfirebaugh How do you feel about this use of `setTimeout`? There are some situations where we cannot fire an event synchronously (namely in a constructor).

jfirebaugh: Well, it's probably technically a race condition -- I don't think the DOM specs guarantee that the `setTimeout(0)` will trigger before the AJAX request onload (which is plausible if, say, the request is immediately satisfied by a cache).

jfirebaugh: Passing the evented parent as a parameter to each of these constructors, calling `setEventedParent`, then firing a sync `dataloading`?

lucaswoj: I like that design 👍 I'll explore first thing tomorrow.

lucaswoj: Done. This turned out to be quite clean.

lucaswoj:
Ready for another round of 👁

PR #3309, opened 2016-10-06
Remove source.load

lucaswoj:
As a sister PR to https://github.com/mapbox/mapbox-gl-js/pull/3308, this PR removes the `Source#source.load` event in favor of the `Source#data` event.

Why? We expose a smaller API surface. `source.load` is redundant with the `data` event. We wish the `Map#load` event had been this way all along. There's little precedent for `.` namespaced events therefore we should try to keep them out of our public API. 

ref #1715 

## Launch Checklist

### DIFF
--- a/js/source/geojson_source.js
+++ b/js/source/geojson_source.js
@@ -88,8 +88,7 @@ function GeoJSONSource(id, options, dispatcher) {
             this.fire('error', {error: err});
             return;
         }
-        this.fire('data', {dataType: 'source'});
-        this.fire('source.load');
+        this.fire('data', {dataType: 'source', isFirst: true});
     }.bind(this));
 }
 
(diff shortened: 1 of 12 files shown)


jfirebaugh:
Well, it's replacing `source.load` with `data`+`isFirst: true`. I'm not really convinced that's a smaller or better API. What if we find there's another variety of `data` event? Do we add another boolean flag? As a general API design rule of thumb, if you find yourself adding a boolean flag, you probably should be adding an extensible enumeration instead -- which is what event types are.

lucaswoj:
After slogging through this implementation, I agree with your point @jfirebaugh. I'll look at this with fresh eyes tomorrow and see if there's anything worth keeping in this PR.

lucaswoj:
Planning to go a different direction per @jfirebaugh's comments. New plan to be proposed in the original ticket, #1715

PR #3313, opened 2016-10-06
[game controls example] Allow simultaneous turning and moving

mdamien:
Refactor the example to:

 - Allow turning while moving
 - Don't wait on repeated keys for continuous movement
 - Allow WASD controls
 - do not hijack the focus, let the user click before being able to interact

Some notes on the implementation:
 
 - if this example was to demonstrate a simple usage of `panBy`, I understand that this PR should not be accepted as-is
 - actually, using `easeTo` isn't that smooth since we should be keeping track of the position we should be in and always trying to transition to that instead of using the current position as the last goal (making the speed very irregular right now)
 - using `requestAnimationFrame` can be seen as a lot more heavier than that example should be
 - it's gonna `easeTo` even if no key is pressed

Random note on the `mapbox-gl` api: if I use `map.easeTo` without `center` but with an `offset` options, it won't complain and I will be left wondering why it doesn't move. I think this problem is a bit everywhere in the API but I can make a PR to fix that too.

### DIFF
--- a/docs/_posts/examples/3400-01-26-game-controls.html
+++ b/docs/_posts/examples/3400-01-26-game-controls.html
@@ -42,10 +42,10 @@ tags:
         interactive: false
     });
 
-    // pixels the map pans when the up or down arrow is clicked
-    var deltaDistance = 100;
-    // degrees the map rotates when the left or right arrow is clicked
-    var deltaDegrees = 25;
+    // pixels the map pans each second a key is pressed
+    var DELTA_DISTANCE = 2000;
+    // degrees the map rotates each second a key is pressed
+    var DELTA_DEGREES = 250;
 
     var instructions = document.getElementById('instructions');
     map.on('click', function() {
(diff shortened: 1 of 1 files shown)


mollymerp:
Hi @mdamien and thank you for contributing! 

Our examples are written with the goal of demonstrating a tightly scoped feature with an emphasis on concise, readable code. I think this PR introduces some great features, but also includes added complexity that I think is out of scope for the examples page.

My recommendation would be to build a [plugin](https://www.mapbox.com/mapbox-gl-js/plugins/) with a robust implementation of game-like map controls and we can feature it with other [third-party projects](https://github.com/mapbox/mapbox-gl-js#third-party-projects).
Thanks again! 
cc @lucaswoj @samanpwbb

mdamien:
ah ah, exactly what I was thinking, the culprit is this tweet: https://twitter.com/fallsemo/status/784019877707255808 :smile:

PR #3319, opened 2016-10-07
Fix wrong reported coordinates on Android Chrome

### LINKED ISSUE #3304: TouchEnd reported coordinates (e.lngLat) wrong on Android Chrome?
Android: 5.0.1
Chrome: 53.0.2785.124

MapboxGL JS : "version":"0.24.0"

### Steps to Trigger Behavior

 1. simple map, add listeners to touchstart and touchend with console.log(e.lngLat) on both
 2. tap the map on a mobile phone and watch the coordinates with chrome remote debugging, or use an alert instead of console
 

### Expected Behavior
The two events should have almost equal coordinates
### Actual Behavior
The two events have different coordinates (the touchend is usually shifted up and left), example:
touchstart coords: LngLat(9.435743459690002, 45.642242729483655)
touchend coords: LngLat(9.425303808457613, 45.646926857999205)
I noticed this trying to implement a drag behaviour for markers - they would always end away from where I shifted them after release of the finger.
Using desktop events on desktop chrome works perfectly, using touch events on mobile does not (unless I'm doing something spectacularly wrong).

FDN73:
Closes #3304.

 - [x] Changed the util/dom.js/ touchPos function to consider changedTouches on touchend event as touches is empty on such event
 - [x] Tested only on android tablet (from the docs the fix shoud work on ios too)

### DIFF
--- a/js/util/dom.js
+++ b/js/util/dom.js
@@ -65,7 +65,7 @@ exports.mousePos = function (el, e) {
 exports.touchPos = function (el, e) {
     var rect = el.getBoundingClientRect(),
         points = [];
-    var touches = e.type === 'touchend'?e.changedTouches:e.touches;
+    var touches = (e.type === 'touchend')? e.changedTouches : e.touches;
     for (var i = 0; i < touches.length; i++) {
         points.push(new Point(
             touches[i].clientX - rect.left - el.clientLeft,
--- a/js/util/dom.js
+++ b/js/util/dom.js
@@ -65,10 +65,11 @@ exports.mousePos = function (el, e) {
 exports.touchPos = function (el, e) {
     var rect = el.getBoundingClientRect(),
         points = [];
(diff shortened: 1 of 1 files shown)


mourner:
Great! Can you please fix the formatting (spacing in the ternary operator) so that the build passes?

FDN73:
Hello,
first time I contribute to a project, sorry for being lame :) I've fixed the line (needed two tries) and I see that the ci build now passes, should I do anything? Repeat the pull request?

mourner:
@FDN73 no, all looks good now!

mourner:
Leaflet [relies on changedTouches](https://github.com/Leaflet/Leaflet/blob/fd438cb2774b9cd404c996af62a03774dddd8791/src/map/handler/Map.Tap.js#L78) in `touchend` event too, so I think this is a correct fix. Merging when all checks pass.

PR #3332, opened 2016-10-10
Simplify draw_line.js

mourner:
- Removed lots of repetition in `draw_line.js`
- Moved calculation of line antialiasing matrix and stretch coefficient to transform so that it’s not repeated

👀 @lucaswoj @ansis

### DIFF
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -468,5 +468,16 @@ Transform.prototype = {
         m = mat4.invert(new Float64Array(16), this.pixelMatrix);
         if (!m) throw new Error("failed to invert matrix");
         this.pixelMatrixInverse = m;
+
+        // line antialiasing matrix
+        m = mat2.create();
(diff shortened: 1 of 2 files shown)


mourner:
## frame-duration benchmarks

**master cd6c761:** 10.6 ms, 15% > 16ms
**simplify-draw-line 3381447:** 10.2 ms, 14% > 16ms

jfirebaugh:
Native will soon look quite a bit different: [preview](https://github.com/mapbox/mapbox-gl-native/blob/3ef3084425b59ee1ea2bc26787c7e2d6363e8a62/src/mbgl/renderer/painter_line.cpp).

mourner:
> Can you port the lineAntialiasingMatrix and lineStretch changes to -native?

I can look into this, but I guess this would have much less impact in native.

jfirebaugh:
I'll open an issue in native to hoist these calculations.

Issue #3136, opened 2016-09-02
Popup dynamic anchor only uses bottom, left and right


apkoponen:
mapbox-gl-js version 0.23.0:

### Steps to Trigger Behavior

 1. Go to https://www.mapbox.com/mapbox-gl-js/example/popup-on-click/
 2. Click open a Popup
 3. Move the map so that the popup goes close to the canvas corners.

### Expected Behavior

When the popup would be partly hidden the anchor should be changed accordingly. For example, when the top of the map canvas overlaps with the Popup the popup should change the anchor to top and move below the marker. This works in 0.22.1 ([JSBin](http://jsbin.com/bohihanava/edit?html,css,js,output))

### Actual Behavior

The anchor is not changed properly. Instead the popup stays anchored to bottom and goes out of the canvas (is partly hidden).

PR #3333, opened 2016-10-10
Check offset when determining if should anchor to top

jisaacks:
## Launch Checklist

When auto determining the anchor for a popup, you were not checking the offset. I was using a negative bottom offset to place the popup _above_ the marker when anchored to the bottom. This means that when you checked `if (pos.y < height) {` to see if the popup would be off the viewport, your code would assume it is not, then the offset would put it off the screen. By adding the offset to the position when comparing, it prevents the offset from pushing the popup off the viewport.

The other conditions might benefit from checking the offset  as well (not positive) but this is the only one that affected me.

 - [ ] post benchmark scores

### DIFF
--- a/js/ui/popup.js
+++ b/js/ui/popup.js
@@ -234,7 +234,7 @@ Popup.prototype = util.inherit(Evented, /** @lends Popup.prototype */{
             var width = this._container.offsetWidth,
                 height = this._container.offsetHeight;
 
-            if (pos.y < height) {
+            if (pos.y + offset.bottom.y < height) {
                 anchor = ['top'];
             } else if (pos.y > this._map.transform.height - height) {
                 anchor = ['bottom'];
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Attempting to take the offset into account previously caused a regression in auto-anchoring: #3136. Can you confirm that that isn't the case here?

jisaacks:
@jfirebaugh when testing locally I do not see that regression. The anchoring updates as it should. 

Looking at the old code it was using `offset.bottom[1]` but the offset gets normalized to points (not arrays/ point likes) so `offset.bottom.y` is what I am using here. That may have been what the issue was before?

jfirebaugh:
Yes, I think you're right about `[1]` rather than `.y` being the issue before.

Can you add a [test case](https://github.com/mapbox/mapbox-gl-js/blob/master/test/js/ui/popup.test.js) for this? Something like `Popup auto-anchors at top if its bottom offset would push it off-screen`.

Thanks!

jisaacks:
@jfirebaugh Sure, I'll try to get that done in the next day or so.

jisaacks:
@jfirebaugh I added the requested test.

jfirebaugh:
Thank you!

PR #3335, opened 2016-10-10
Dedupe setPattern in fill/extrude code

mourner:
- reuse `setPattern` code between `draw_fill` and `draw_extrude` (it was duplicated before)
- a few minor cleanups in `draw_fill`

👀  @lucaswoj @lbud

### DIFF
--- a/js/render/draw_extrusion.js
+++ b/js/render/draw_extrusion.js
@@ -3,10 +3,10 @@
 var mat3 = require('gl-matrix').mat3;
 var mat4 = require('gl-matrix').mat4;
 var vec3 = require('gl-matrix').vec3;
-var pixelsToTileUnits = require('../source/pixels_to_tile_units');
 var Buffer = require('../data/buffer');
 var VertexArrayObject = require('./vertex_array_object');
 var StructArrayType = require('../util/struct_array');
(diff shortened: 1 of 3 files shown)


js/render/draw_extrusion.js
```diff
@@ -163,7 +163,7 @@ function drawExtrusion(painter, source, layer, coord) {
     );
 
     if (image) {
-        setPattern(image, tile, coord, painter, program);
+        setPattern(image, tile, coord, painter, program, true);
```
lucaswoj: Can we tweak the argument signature to avoid a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) here? Perhaps 

```js
setPattern(image, tile, coord, painter, program, {includeHeightFactor: true})
```

lucaswoj: What makes this is a "very hot path"?

mourner: Anything that may be executed dozens of times per frame.

lucaswoj: SGTM. 🚢

lbud:
🙏 thx @mourner!

PR #3341, opened 2016-10-11
Update mapbox-gl-compare plugin to v0.0.2

tristen:
cc @mollymerp for review


mollymerp:
Looks good, but please update the [`latest` field](https://github.com/mapbox/mapbox-gl-js/blob/f91523a0d9b85771377cd5a5ebff699f114278a5/docs/_data/plugins.json#L112) as well :)

PR #3340, opened 2016-10-11
3D Extrusion Example - Indoor Floor Map

ryanbaumann:
## Launch Checklist

 - [x] briefly describe the changes in this PR
 - [x] Code review @lucaswoj @mollymerp 
 - [x] Copy and style review 

Adds an example for using fill-extrude-height to create an indoor map.

Potential improvements:

* I stored all data for the example as geoJSON.  Convert to Mapbox Tileset for example succinctness.
* Design review of colors, style of map.
* Add room labels to highlight the "indoor map" use case.

PR #3349, opened 2016-10-11
Road profile extrusion example

ryanbaumann:
## 3D road profile example

Example of extruding nearby polygons from GPS data to create 3D road profiles.  Applicable in fitness, GIS, and resource industries such as construction and mining.  In this example, we extrude polygons to a height = altitude and a color = gradient.  

## Questions

1. For a fitness application, having the map in the background is great.  For a mining applicaiton, I'd want to add a custom GeoTiff background with site-specific context.
2. I want to communicate this as more than an example for fitness consumers.  Ideally agriculture, forestry, and construction/mining customers can work with this example as a step up from "doing GIS in a CAD tool" only.

## Launch Checklist

- [ ] Code review @mollymerp @lucaswoj 
- [ ] Style review @peterqliu @ajashton


docs/_posts/examples/3400-01-28-road-profile-extrusion.html
```diff
@@ -0,0 +1,55 @@
...
+});
+
+map.on('load', function() {
+
+    // source GeoJSON data for visual is at
+    // https://gist.github.com/ryanbaumann/f196a84c4c5728d860db0fc7a19d0c9b
+    map.addSource("roadProfile", {
+        'type': 'vector',
+        'url': 'mapbox://examples.ciu640p5o0ifh2pjwzwd6ga98-0h1qy'
+    });
```
lucaswoj: I worry that the `mapbox://examples...` URL is too much of a black box in trying to explain how this effect is implemented. Would it be possible to simplify the data and embed it as GeoJSON?

ryanbaumann: @lucaswoj yes.  I could cut the data down to a small subset with 30-40 polygons.  However the visual suffers then, but the proof of concept as an example still works great.

@peterqliu suggested one option, to put the data in a gist and link from the example.  For this example I added a comment above the source to this gist https://gist.github.com/ryanbaumann/f196a84c4c5728d860db0fc7a19d0c9b which explains the data.

ryanbaumann: @lucaswoj 
* I updated the data to source from geojson url hosted on the /assets folder of the doc.
* I added comments to the DDS functions
* I linked to a gist explaining how to get data in this format for creating the example.

peterqliu:
I don't think we should offer this example, for three reasons:

1) **it is largely redundant of https://github.com/mapbox/mapbox-gl-js/pull/3340**, using the same methods. The key insight here is "applying the same tech to Use Case X", which is better suited to a blog post than an example.

2) **it likely raises more questions than it answers.** @ryanbaumann if you're cribbing this from the [bikeshare example](https://www.mapbox.com/bites/00094/gl/), the data prep is considerable before we even get to the extrusion step:

- get a geojson line
- get turf
- use turf.along to recursively divide line into segments
- take each segment, and buffer it a certain distance into polygons
- pack those polygons into a geojson, and set that geojson as a source

3) **The extruded polygon method is a stopgap measure, for when we eventually support extruded lines.** Polygons here are jagged at the top, and overlap each other in three-dimensional space.

ryanbaumann:
@peterqliu agreed, it doesn't show anything technically novel over example #3340, and it would take some time to replicate the data format from 0 to the example.

Shelving.

mollymerp:
@ryanbaumann this would still be a great for a tweet or a blog post -- even though the implementation of extruded lines will change it is a nice visual.

ryanbaumann:
@mollymerp sounds like a plan - I'll move to  a bite for labs & blog discussion

Issue #3350, opened 2016-10-12
Eliminate Source#on{Add,Remove}(map)


jfirebaugh:
The main use of `on{Add,Remove}(map)` in source subclasses is to store a reference to the map for later use. This introduces both statefulness (certain source operations work only after `onAdd` has been called) and tight coupling between a source and the map.

Instead, sources should be passed the things to which they need access as parameters to the relevant methods.

Ref https://github.com/mapbox/mapbox-gl-js/pull/3255#discussion_r80765491

mourner:
> Instead, sources should be passed the things to which they need access as parameters to the relevant methods.

It's not always possible directly — e.g. `map.getSource('image').setCoordinates(coords)`, where `setCoordinates` depends on map transform to recalculate coords. We'll have to pass a map reference to methods like `loadTile`.

jfirebaugh:
In that case, `setCoordinates` should be a simple setter, deferring the calculations that require a transform until a later time. Ideally `loadTile` should not take a map reference but rather a reference to the transform or whatever other state is needed, so that there isn't a circular dependency between map and sources.

lucaswoj:
Thank you for opening! This will go hand-in-hand with https://github.com/mapbox/mapbox-gl-js/issues/3186.

Issue #3186, opened 2016-09-14
Refactor source interface to better accommodate non-tiled sources


lucaswoj:
We currently shoehorn all source types into the `VectorTileSource` interface by abusing `Tile` objects. For example, [`ImageSource` populates a single `Tile` with a number of ad-hoc properties](http://github.com/mapbox/mapbox-gl-js/blob/34fde741d80370b304163fd71f2ef230f78348ee/js/source/image_source.js#L118-L119). This architecture breaks and causes bugs like https://github.com/mapbox/mapbox-gl-js/issues/3010. 

There are several different types of sources in GL JS and each should have it's own interface:

```js
interface VectorTileSource {
    getVectorTile(coord: TileCoord, callback: (error: Error, tile: VectorTile) => void);
}

interface RasterTileSource {
    getRasterTile(coord: TileCoord, callback: (error: Error, tile: ImageData) => void);
}

interface ImageSource {
    getImage(callback: (error: Error, image: ImageData) => void);
}

interface VideoSource {
    getVideo(callback: (error: Error, video: () => ImageData) => void);
}
```

Inspired by https://github.com/mapbox/mapbox-gl-js/issues/3010#issuecomment-244201644

Related to #2982 #281 #3010

lucaswoj:
I just updated the interface proposals above to address the concerns of #3123 and the needs of https://github.com/mapbox/mapbox-gl-js/issues/3051 🚀

mourner:
@lucaswoj why does `getVectorTile` accept `url` but `getRasterTile` accept `coord`? I guess both should be either `url` or `coord`.

lucaswoj:
@mourner Typo. Fixed above.

1ec5:
This is also being tracked for the iOS and macOS SDKs in mapbox/mapbox-gl-native#6861.

anandthakker:
This refactor should also accommodate non-tiled _vector_ sources. 
 Related issues:
 - https://github.com/mapbox/mapbox-gl-js/issues/2275
 - https://github.com/mapbox/mapbox-gl-js/issues/4954#issuecomment-313738448 / https://github.com/mapbox/mapbox-gl-js/issues/3431#issuecomment-255817951

PR #3353, opened 2016-10-12
DRY up VideoSource code

mourner:
While looking into #3350, I noticed that most of the `VideoSource` code is a copy-paste from `ImageSource`. This PR removes a lot of duplication by making `VideoSource` inherit from `ImageSource`.

We might later consider some kind of composition instead of inheritance, or inheriting both from a shared base class (e.g. `GeoreferencedSource`), but I'd leave that for later when we address #3186.

👀 @lucaswoj @ansis

### DIFF
--- a/js/source/video_source.js
+++ b/js/source/video_source.js
@@ -1,15 +1,8 @@
 'use strict';
 
 var util = require('../util/util');
-var TileCoord = require('./tile_coord');
-var LngLat = require('../geo/lng_lat');
-var Point = require('point-geometry');
-var Evented = require('../util/evented');
 var ajax = require('../util/ajax');
-var EXTENT = require('../data/bucket').EXTENT;
-var RasterBoundsArray = require('../render/draw_raster').RasterBoundsArray;
-var Buffer = require('../data/buffer');
-var VertexArrayObject = require('../render/vertex_array_object');
+var ImageSource = require('./image_source');
 
 module.exports = VideoSource;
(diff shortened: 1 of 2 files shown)


mourner:
Forgot to mention that this PR also fixes `ImageSource` constantly recreating the GL texture on every frame — previously `_prepared` was never set to `true` for some reason (while it was in `VideoSource`).

js/source/video_source.js
```diff
@@ -107,92 +97,18 @@ VideoSource.prototype = util.inherit(Evented, /** @lends VideoSource.prototype *
     /**
      * Sets the video's coordinates and re-renders the map.
      *
+     * @method setCoordinates
      * @param {Array<Array<number>>} coordinates Four geographical coordinates,
      *   represented as arrays of longitude and latitude numbers, which define the corners of the video.
      *   The coordinates start at the top left corner of the video and proceed in clockwise order.
      *   They do not have to represent a rectangle.
      * @returns {VideoSource} this
      */
```
lucaswoj: Could we eliminate the need for this inherited comment by rewording it? `s/video/source` or `s/video/data` or `s/video/layer`?

mourner: Maybe, but what to do with `@returns {VideoSource}`? cc @jfirebaugh

lucaswoj: Oooh. Good point.

Issue #3299, opened 2016-10-05
Improve the map link does not go to the right location


planemad:
**mapbox-gl-js version**: latest

### Steps to Trigger Behavior

1. Goto [a GL style preview page](https://api.mapbox.com/styles/v1/rasagy/citw07bpz00682iolbbi27kjm.html?title=true&access_token=pk.eyJ1IjoicmFzYWd5IiwiYSI6Ilk2ay1VMEEifQ.CJfn_CxDfwyN5xaVfx4jaw#11.72/12.6671/80.0374).
2. Click `improve the map`

### Expected Behavior
https://www.mapbox.com/map-feedback/ page opens at the same map location

### Actual Behavior
The feedback page opens up at a zoomed out view of the world making it difficult to submit a report at the desired location. 

The link works correctly on [Mapbox JS](https://api.tiles.mapbox.com/v4/mapbox.streets/page.html?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpamV0Zms2ZzAwMGh0Zm01MDR3YmV2cHgifQ.m9a-_eKcMy9A6glxEtuj3Q#12/31.1173/33.8506)

mollymerp:
fixed in #3356

PR #3356, opened 2016-10-12
Fix 'Improve this map' link to include map location

mollymerp:
ref #3299 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/js/ui/control/attribution_control.js
+++ b/js/ui/control/attribution_control.js
@@ -68,10 +68,10 @@ AttributionControl.prototype = util.inherit(Control, {
             return true;
         });
         this._container.innerHTML = attributions.join(' | ');
-        this._editLink = this._container.querySelector('.mapbox-improve-map')
     },
 
     _updateEditLink: function() {
+        if (!this._editLink) this._editLink = this._container.querySelector('.mapbox-improve-map');
         if (this._editLink) {
             var center = this._map.getCenter();
             this._editLink.href = 'https://www.mapbox.com/map-feedback/#/' +
--- a/js/ui/control/attribution_control.js
+++ b/js/ui/control/attribution_control.js
@@ -68,6 +68,8 @@ AttributionControl.prototype = util.inherit(Control, {
             return true;
         });
         this._container.innerHTML = attributions.join(' | ');
+        // remove old DOM node from _editLink
+        this._editLink = null;
     },
 
     _updateEditLink: function() {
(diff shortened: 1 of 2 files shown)


mollymerp:
note that this should be merged in after #3355

js/ui/control/attribution.js
```diff
@@ -67,8 +67,14 @@ AttributionControl.prototype = util.inherit(Control, {
...
         });
-
         this._container.innerHTML = attributions.join(' | ');
+
+        var attributionNodes = this._container.childNodes;
+        for (var i = 0; i < attributionNodes.length; i++) {
+            if (attributionNodes[i].classList && attributionNodes[i].classList.contains('mapbox-improve-map')) {
+                this._editLink = attributionNodes[i];
+            }
+        }
```
lucaswoj: Why did you choose to put this logic within `_updateAttributions` rather than `_updateEditLink`? Did you consider using [`Element#getElementsByClassName`](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName) rather than a manual search for `mapbox-improve-map`?

mollymerp: If I put the logic in `_updateEditLink` the node is not attached the document and so changes to the node's `href` property do not update on the live attribution node. I had `document.getElementsByClassName` and that was failing the tests, but I didn't realize I could do `this._container.getElementsByClassName` 😮  I'll change that now.

lucaswoj: > If I put the logic in `_updateEditLink` the node is not attached the document and so changes to the node's `href` property do not update on the live attribution node. 

In what case is the element not attached to the document?

js/ui/control/attribution_control.js
```diff
@@ -67,11 +67,11 @@ AttributionControl.prototype = util.inherit(Control, {
             }
             return true;
         });
-
         this._container.innerHTML = attributions.join(' | ');
     },
 
     _updateEditLink: function() {
+        if (!this._editLink) this._editLink = this._container.querySelector('.mapbox-improve-map');
```
lucaswoj: When, if ever, does the "edit link" element get replaced in the DOM? Will we update `this._editLink` when the element is replaced?

mollymerp: This is another reason why I had the logic in `_updateAttributions` before. DOM updates happen if new sources are added. Do you think it's preferable to re-assign `this._editLink` on every `mapmove` / `_updateEditLink` call or put assignment into `_updateAttributions`?

lucaswoj: > reset `this._editLink` to null when the DOM is updated

Any of the approaches above would work. I think I would opt for this one.

PR #3361, opened 2016-10-13
Layout property proposal: numeric-precision

mourner:
An alternative solution for https://github.com/mapbox/mapbox-gl-js/pull/3358. Introduces `numeric-precision` layout property that affects how numbers are interpolated in text fields. If set, 
numbers in resolved text fields of a layer will have the specified precision (e.g. `"numeric-precision": 2` will make `123.45678` appear s `123.46`).

This is just a proof of concept; a proper implementation would require corresponding style-spec changes, including settling on a default value there.

cc @danpat @jfirebaugh

### DIFF
--- a/js/symbol/resolve_text.js
+++ b/js/symbol/resolve_text.js
@@ -13,16 +13,18 @@ module.exports = resolveText;
  */
 function resolveText(features, layoutProperties, codepoints) {
     var textFeatures = [];
+    var textField = layoutProperties['text-field'];
+    var transform = layoutProperties['text-transform'];
+    var numericMultiplier = Math.pow(10, layoutProperties['numeric-precision']);
 
     for (var i = 0, fl = features.length; i < fl; i++) {
-        var text = resolveTokens(features[i].properties, layoutProperties['text-field']);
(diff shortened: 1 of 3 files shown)


lucaswoj:
I'm not sure that the layer's layout properties is the best place for this bit of metadata. Users would be unable to use different numeric precisions for different `{tokens}` within `text-field` (and maybe even `{icon-image}`). Whether or not this is a common use case, it brings up a few questions about our long term API design direction.

I'm excited about exploring string operations in the spirit of https://github.com/mapbox/mapbox-gl-function/issues/28

jfirebaugh:
Agree 100% with @lucaswoj.

mourner:
I agree that arbitrary expression functions are much more powerful and may be better suited for this kind of task, although it may take us a very long time to get there. 

I'm also interested to see how would such an arbitrary expression look for rounding a number. Would we introduce special keywords like `round`? What if the property's name is `round`?

lucaswoj:
> ... we create a syntax that is:
> - a valid subset of an existing programming language (Javascript?)

PR #3375, opened 2016-10-14
Use sinon's sandbox API

jfirebaugh:
Create a sandbox in a global `beforeEach` callback, and restore it in a global `afterEach` callback. Inject the methods into the tap `Test` object so that they are accessible via `t.spy`, `t.stub`, etc., and use those exclusively.

This ensures that a stubbed or spied global method (e.g. `util.warnOnce`) does not linger beyond a single test case.

### DIFF
--- a/test/node_modules/mapbox-gl-js-test.js
+++ b/test/node_modules/mapbox-gl-js-test.js
@@ -1,0 +1,17 @@
+'use strict';
+
+var tap = module.exports = require('tap');
+var sinon = require('sinon');
+
+tap.beforeEach(function (done) {
+    this.sandbox = sinon.sandbox.create({
+        injectInto: this,
+        properties: ['spy', 'stub', 'mock']
+    });
+    done();
+});
+
+tap.afterEach(function (done) {
+    this.sandbox.restore();
+    done();
+});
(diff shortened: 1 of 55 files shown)


test/node_modules/mapbox-gl-js-test.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
```
lucaswoj: Why did you put this in `test/node_modules` rather than `test/testutil`?

test/node_modules/mapbox-gl-js-test.js
```diff
@@ -0,0 +1,17 @@
...
+    this.sandbox = sinon.sandbox.create({
+        injectInto: this,
+        properties: ['spy', 'stub', 'mock']
+    });
+    done();
+});
+
+tap.afterEach(function (done) {
+    this.sandbox.restore();
+    done();
```
lucaswoj: We may want to call [`window.restore()`](http://github.com/mapbox/mapbox-gl-js/blob/4dc27c3810cb8f09d0de9578ad2e6b0e8eeb499f/js/util/window.js#L8-L9) in here too (unless that causes a perf problem)

test/node_modules/mapbox-gl-js-test.js
jfirebaugh: So it can be required as `require('mapbox-gl-js-test')`, without a relative path which requires updating if the test file is moved up or down the directory hierarchy.

I'd like to do something similar for library requires, replacing `require('../../../js/foo/bar')` with `require('js/foo/bar')` via [app-module-path](https://www.npmjs.com/package/app-module-path).

test/node_modules/mapbox-gl-js-test.js
jfirebaugh: Without `window.restore()`:

```
real	0m40.833s
user	0m39.433s
sys	0m3.908s
```

With:

```
real	1m6.238s
user	1m2.931s
sys	0m6.347s
```

jfirebaugh: For completeness, prior to this PR:

```
real	0m40.549s
user	0m38.895s
sys	0m3.853s
```

PR #3382, opened 2016-10-14
Move other test utilities to mapbox-gl-js-test "module"

jfirebaugh:
Followup to https://github.com/mapbox/mapbox-gl-js/pull/3375#discussion_r83475060.


lucaswoj:
I've never seen the checked-in `node_modules` pattern used here. [`app-module-path`](https://www.npmjs.com/package/app-module-path) feels more idiomatic to me. Would you be open using that design instead? It would allow all modules to require `test/testutil/blah` regardless of their location in the file tree.

jfirebaugh:
Checked-in `node_modules` is the officially-sanctioned way to accomplish this ([1](https://gist.github.com/branneman/8048520#gistcomment-972519), [2](https://gist.github.com/branneman/8048520#gistcomment-972522), [3](https://gist.github.com/branneman/8048520#gistcomment-1880931)).

I'd prefer to figure out a way to make `require('mapbox-gl-js/foo/bar')` work from test files using this technique, rather than app-module-path, which has the downside of relying on undocumented node internals. So far it seems to require a symlink though, which has its own issues.

lucaswoj:
> Checked-in node_modules is the officially-sanctioned way to accomplish this (1, 2, 3).

Can't argue with that!

PR #3383, opened 2016-10-14
Add RFC_TEMPLATE.md

lucaswoj:
This template adds a `RFC_TEMPLATE.md` file. Below is a use of the template to justify its existence.

I was inspired by Yarn's RFC process https://github.com/yarnpkg/rfcs/blob/master/0000-template.md

cc @jfirebaugh @mourner @mollymerp @1ec5 @tmcw @mapsam 

----

## Motivation

The GL ecosystem is large and interconnected. Carefully considering and discussing design decisions is a crucial step for any big change or new feature.

There is no formal process for proposing a big change or new feature. Many times new features are "proposed" with a pull request, which leaves little room for big picture discussion. 

## Design Alternatives

 - require a "RFC" for all changes
 - provide a "RFC" template and suggest it be used for new features & big changes
 - disallow all changes except those "on the roadmap"
 - do nothing

## Design

We should provide a RFC template and suggest it be used for big changes and new features. This will foster discussion and careful design when needed without creating unnecessary bureaucracy when unneeded. 

### Implementation

We will add a `RFC_TEMPLATE.md` to `mapbox-gl-js`. If things go well, we can expand it to all `mapbox-gl*` repos.

We will fill out the template when looking to clarify our thinking and invite discussion on a big change or feature. We will suggest 3rd party users fill out the template when proposing a feature or big change.

 > Great idea @user! Let's start thinking concretely about the design and invite discussion by filling out a [RFC template](#). 

RFCs can live as comments on issues. 

### Concepts

RFCs are a common and self-explanatory concept.

### DIFF
--- a/RFC_TEMPLATE.md
+++ b/RFC_TEMPLATE.md
@@ -9,14 +9,14 @@ Focus on the problem and save design ideas for the next section.
 
 <!--
 How could we accommodate the use cases above?
-What are the pros and cons of each design?
-What are the pros and cons of doing nothing?
+Is "do nothing" an option?
 -->
 
 ## Design
 
 <!--
 Which design should we implement? Why?
+What are the pros and cons of this design?
 -->
 
 ### Implementation
--- a/RFC_TEMPLATE.md
+++ b/RFC_TEMPLATE.md
@@ -1,0 +1,37 @@
+## Motivation
+
+<!--
+What use cases are we trying to accommodate?
+Focus on the problem and save design ideas for the next section.
+-->
+
+## Design Alternatives
+
+<!--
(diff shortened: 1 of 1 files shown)


mourner:
👍, however, changes that are big enough to warrant an RFC submission are likely to impact the whole ecosystem rather then just the JS implementation (otherwise they'd be fine as a simple PR/issue discussion), so I think we should try this in `mapbox-gl` repo, and update the template to cover both implementations at once (if something is platform-specific, you could always leave it out).

lucaswoj:
Sounds good @mourner. I'll move the template to `mapbox-gl` now.

Issue #3368, opened 2016-10-14
Use (some) ES6 features


jfirebaugh:
Let's try allowing the following ES6 features:

* `let`/`const`
* arrow functions
* template strings
* `for...of`
* object literal extensions

This is a subset of ES6 that's implemented in node 4.x+ and available without runtime flags, meaning we can run the code in node without any transpiling, avoiding https://github.com/mapbox/mapbox-gl-js/issues/855#issuecomment-70935993.

For the browser build, we'd transpile with babel, under a configuration that allows only these features. (ES6 browser support is still lacking in browsers supported by GL JS, in particular IE 11 and iOS 9's Mobile Safari.)

mourner:
Node v6 [enters LTS phase in 4 days](https://github.com/nodejs/LTS), and we already use it in Circle CI, so we could already target ES6 features in v6 instead of v4.

For the browser build, I'd love to explore [Buble](https://buble.surge.sh/guide/) instead of Babel, which in my opinion aligns much better with our goal of transpiling a productive, safe subset of ES6 (without heavy stuff like generators and iterators), and producing clean lightweight code without huge slowdowns in compilation times.

@lucaswoj mentioned this as a good opportunity to also add Flow annotations, which I'd love to do to. But to keep the source runnable under Node for tests, we'd have to experiment with something like [flow-node](https://flowtype.org/docs/running.html) and I'm not sure how well it plays with our testing pipeline (e.g. whether `nyc` still runs properly with it, etc.).

mourner:
Useful ES6 features that are available in Node v6 but not v4:

- default function parameters
- rest parameters
- spread operator (`...`)
- destructuring assignment

I'd also love to use native `class` syntax (available in v4+).

jfirebaugh:
@mourner I'm confused by https://buble.surge.sh/guide/#unsupported-features and https://buble.surge.sh/guide/#dangerous-transforms. Does this mean buble does or does not support `for...of`?

mourner:
> Bublé offers 'dangerous transforms', which generate highly efficient ES5 code but with important caveats.

@jfirebaugh it supports `for...of`, but with a caveat that it can't be used with ES6 iterables (`Map`, `Set`, `[Symbol.iterator]` etc.) and generators, because it transpiles it to a simple `for` loop without complicated wrappers. It's fine to use in our case.

tmcw:
[...]

PR #3388, opened 2016-10-14
Bublé / ES6

jfirebaugh:
Refs #3368

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,8 +11,10 @@
     "no-var": "error",
     "no-warning-comments": "error",
     "object-curly-spacing": "off",
+    "prefer-template": "error",
     "quotes": "off",
     "space-before-function-paren": "off",
+    "template-curly-spacing": "error"
   },
   "env": {
     "es6": true,
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,6 +11,7 @@
     "no-var": "error",
     "no-warning-comments": "error",
     "object-curly-spacing": "off",
+    "prefer-arrow-callback": "error",
     "prefer-template": "error",
     "quotes": "off",
     "space-before-function-paren": "off",
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,6 +12,7 @@
     "no-warning-comments": "error",
     "object-curly-spacing": "off",
     "prefer-arrow-callback": "error",
+    "prefer-const": "error",
     "prefer-template": "error",
     "quotes": "off",
     "space-before-function-paren": "off",
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,6 +8,7 @@
     "key-spacing": "off",
     "no-eq-null": "off",
     "no-new": "off",
+    "no-var": "error",
     "no-warning-comments": "error",
     "object-curly-spacing": "off",
     "quotes": "off",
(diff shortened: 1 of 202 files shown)


mourner:
This looks great, I'm excited! Two things to keep in mind:

- Uglifyify currently seems not to work properly with Bubleify, with a pending fix here https://github.com/hughsk/uglifyify/pull/48 (not sure why it's still not merged)
- Bubleify currently uses Buble 0.12, but there's 0.14 already, so it needs to be updated (in particular, destructuring assignment landed after 0.12) https://github.com/garthenweb/bubleify/issues/7

jfirebaugh:
@mourner We're using `minifyify` rather than `uglifyify` though. I'll check well source maps are working on this PR (they seem to be broken about 90% of the time for me regardless of toolchain).

Besides destructuring assignment, which would be nice but not essential, are there important fixes from 0.12 ⇢ 0.14?

jfirebaugh:
My test case for source maps is as follows:

1. `rm -rf node_modules && yarn install && npm run build-min && jekyll serve`
2. Open http://127.0.0.1:4000/mapbox-gl-js/examples/
3. Open developer tools to "Sources" tab
4. ⌘-O, open `vector_tile_source.js`
5. Set a breakpoint on [the first line of `loadTile`](https://github.com/mapbox/mapbox-gl-js/blob/8fb8a9af2fe184bafc5bd249f84f6f7e032fe6bb/js/source/vector_tile_source.js#L52)
6. Refresh the page

Expected result is that the debugger stops at the breakpoint, in `vector_tile_source.js` (not in concatenated or minified source code), with correct stack and local variable names.

Current status on `master` is that at step 5, setting the breakpoint fails: when I click the sidebar at that line, the breakpoint instead gets set at [this line](https://github.com/mapbox/mapbox-gl-js/blob/8fb8a9af2fe184bafc5bd249f84f6f7e032fe6bb/js/source/vector_tile_source.js#L77). So, source maps for the production build are already broken. (This has pretty consistently been my experience across a variety of libraries and build toolchains -- source maps for minified code just don't really seem to work.)

For the `build-dev` version and http://localhost:9966/debug/, this test case does work as expected.

No change to this status quo on the `bublé` branch.

jfirebaugh:
# Benchmarks

## map-load

**master 57af5e3:** 288 ms
**bublé 3542376:** 110 ms

## style-load

**master 57af5e3:** 143 ms
**bublé 3542376:** 135 ms

## buffer

**master 57af5e3:** 1,072 ms
**bublé 3542376:** 1,040 ms

## fps

**master 57af5e3:** 60 fps
**bublé 3542376:** 60 fps

## frame-duration

**master 57af5e3:** 7.7 ms, 0% > 16ms
**bublé 3542376:** 7.3 ms, 0% > 16ms

## query-point

**master 57af5e3:** 1.01 ms
**bublé 3542376:** 1.16 ms

## query-box

**master 57af5e3:** 90.88 ms
**bublé 3542376:** 89.12 ms

## geojson-setdata-small

**master 57af5e3:** 9 ms
**bublé 3542376:** 7 ms

## geojson-setdata-large

**master 57af5e3:** 94 ms
**bublé 3542376:** 99 ms

PR #3397, opened 2016-10-18
Add flow typechecking infrastructure

lucaswoj:
This PR adds flow typechecking to `js/util/util` and the corresponding unit tests. 

The time it takes to run unit tests is unaffected by this change.

cc @tmcw @scothis @jfirebaugh @mourner @mollymerp 
## Launch Checklist


jfirebaugh:
This makes tests not runnable / debuggable in (pure) node. That's a pretty important property for my workflow, and I think I'm not alone.

Could we use flow comments instead?

mourner:
@jfirebaugh I didn't try `flow-node` yet, but I could see myself using it if the benefits outweigh the cons. Did you encounter any particular issues with running `flow-node` instead of `node` in your debugging workflow that we should be aware of?

tmcw:
Flow comments seem chill to me: been using them in simple-statistics for months and they work very well. There's some convenience cost of 5 chars per annotation, but the simplicity of not needing a special environment outweights that, imho - especially since you can move code in and out of the project without a transpilation step.

lucaswoj:
Using `require('flow-remove-types/register')` [in `mapbox-gl-js-test`](https://github.com/mapbox/mapbox-gl-js/pull/3397/files#diff-fb27ffa591985321c485d6a320f20ab0R3) has allowed me to continue using "vanilla" node so far. 

`tap` doesn't play nicely with `flow-node` out of the box.

I would prefer rigging up something with `flow-jsdoc` to using flow comments.

jfirebaugh:
On `npm test`, I'm getting:

```
> mapbox-gl@0.26.0 test-flow /Users/john/Development/mapbox-gl-js
> flow .

sh: flow: command not found
```

You've introduced object spread syntax, and I'm on node 4.5, so if I bypass the above failure, I get:

```
~/Development/mapbox-gl-js $ node test/js/util/util.test.js 
TAP version 13
not ok 1 - SyntaxError: Unexpected token ...
  ---
  at:
    line: 53
    column: 16
    file: vm.js
    function: exports.runInThisContext
  stack: |
    exports.runInThisContext (vm.js:53:16)
    Module._compile (node_modules/flow-remove-types/register.js:12:17)
    Object.<anonymous> (test/js/util/util.test.js:6:14)
  test: TAP
  message: 'SyntaxError: Unexpected token ...'
  ...
1..1
# failed 1 of 1 tests
# time=123.317ms
```

If I run with the `--harmony` flag I get a regular old test failure (`util.test.js:142:11`).

Debugging seems to work correctly without any special hoops, so that's 👍 . We'll need `require('flow-remove-types/register')` added to `render.test.js` as well.

package.json
```diff
@@ -53,7 +56,9 @@
     "envify": "^3.4.0",
     "eslint": "^3.3.1",
     "eslint-config-mourner": "^2.0.0",
+    "eslint-plugin-flowtype": "^2.20.0",
     "eslint-plugin-html": "^1.5.1",
+    "flow": "^0.2.3",
```
tmcw: use `flow-bin`, rather than flow here.

lucaswoj: Done in 6a870d5

package.json
```diff
@@ -50,8 +54,10 @@
     "envify": "^3.4.0",
     "eslint": "^3.3.1",
     "eslint-config-mourner": "^2.0.0",
+    "eslint-plugin-flowtype": "^2.20.0",
     "eslint-plugin-html": "^1.5.1",
     "github-slugger": "^1.1.1",
+    "flow-bin": "^0.2.3",
```
davidtheclark: 👋 I'm just wondering why this is such an old version of Flow? Right now they're up to 0.32 now. Each time we upgrade Flow in Studio it's a little bit of an ordeal since lots of new errors show up.

lucaswoj: Ah! Thanks @davidtheclark. I stupidly changed the package name without updating the version.

lucaswoj: Fixed in 1d3b413

lucaswoj:
Can I get a ✅ on this PR?

mourner:
Does `documentationjs` handle type annotations when rendering docs?

tmcw:
Yep, and very well! Documentation.js is designed to infer JSDoc data from Flow annotations, so you can even omit `@param` tags in favor of Flow annotations in some cases.

PR #3402, opened 2016-10-18
Add support for rendering CJK glyphs top-to-bottom

### LINKED ISSUE #1246: Orient CJK glyphs vertically along vertically oriented lines
For consistency with mapbox/mapbox-gl-native#1682, CJK labels running along north-south roads (or roads that go from top to bottom due to map rotation) should orient glyphs vertically.

/cc @ansis @nickidlugash @mikemorris

lucaswoj:
fixes #1246
see also https://github.com/mapbox/mapbox-gl-native/issues/1682

## Requirements

 - GL JS must render CJK glyphs top-to-bottom along vertical lines, as is conventional in cartographic design

vertical-label-rotation

(this PR does not add support for top-to-bottom point labels or mixed orientation glyphs within a single label)

## Specifications

 - we will use naïve "balanced" breaking 
 - we will enable / disable top-to-bottom labels based on language detection
 - we will use character ranges for language detection ([data](https://en.wiktionary.org/wiki/Module:scripts/data))
 - if a single label has mixed CJK/non-CJK glyphs, we will determine the appropriate glyph orientation.

## Launch Checklist

 - [x] remove all dead code
 - [x] define all jargon relevant to this project 
 - [x] do a code review (@lucaswoj) 
 - [ ] refactor code to use "top-to-bottom" terminology rather than "vertical" terminology
 - [ ] fix glyph centerline alignment for top-to-bottom labels
 - [ ] fix bug causing "top-to-bottom" labels to disappear at certain bearings ([code](http://github.com/mapbox/mapbox-gl-js/blob/c1e0dac21f53122a4f5aeb4049d37e5953800b86/js/data/bucket/symbol_bucket.js#L568-L572))
 - [ ] fix failing unit tests 
 - [ ] support automatically enabling / disabling top-to-bottom labels based on script detection
 - [ ] write test-suite tests for top-to-bottom labels
 - [ ] post benchmark scores
 - [ ] manually rebase onto master (copy-paste style 🍝 )


js/data/bucket/symbol_bucket.js
```diff
@@ -218,16 +218,28 @@ SymbolBucket.prototype.populateArrays = function(collisionTile, stacks, icons) {
...
 
+        var verticalOrientation = false;
+
         if (textFeatures[k]) {
+
             shapedText = shapeText(textFeatures[k], stacks[fontstack], maxWidth,
-                    lineHeight, horizontalAlign, verticalAlign, justify, spacing, textOffset);
+                    lineHeight, horizontalAlign, verticalAlign, justify, spacing, textOffset, oneEm, verticalOrientation);
+
+            if (layout['text-rotation-alignment'] === 'map' && layout['symbol-placement'] === 'line') {
```
1ec5: To avoid verticalizing non-CJK* text, match `textFeatures[k]` against this regular expression: `/[ᄀ-ᇿ가-힣ㄱ-ㆎ一-鿌㐀-䶵　-〿𠀀-𬺯！-￮ぁ-ゟꀀ-꓆᠀-ᢪ]/`.

<sub>* For the purposes of this PR, “CJK” is Hangul, Hanzi, Hiragana, Mongolian, and Yi scripts. However, note that Hangul and Mongolian are separated by spaces and thus should retain the Latin-style line breaking algorithm.</sub>

js/symbol/shaping.js
```diff
@@ -66,19 +64,123 @@ var invisible = {
...
     0x20:   true, // space
-    0x26:   true, // ampersand
-    0x2b:   true, // plus sign
-    0x2d:   true, // hyphen-minus
-    0x2f:   true, // solidus
-    0xad:   true, // soft hyphen
-    0xb7:   true, // middle dot
-    0x200b: true, // zero-width space
-    0x2010: true, // hyphen
-    0x2013: true  // en dash
```
lucaswoj: We need to re-add these ☝️ to the `breakable` lookup

js/data/bucket/symbol_bucket.js
1ec5: (You’ll need to make the regular expression a little more lenient to allow numerals and punctuation.)

1ec5:
> support automatically enabling / disabling vertical labels based on language detection

FYI, we don’t need language detection, only script detection. (As far as I can tell, there isn’t a major script that was traditionally written vertically for one language but never vertically for another.) So looking at Unicode codepoints is sufficient for this task.

lucaswoj:
Some notes as I try to grock the symbol placement code in GL JS

# Glossary

## Classes & Interfaces

### `Anchor`

The geographic coordiante and, optionally, line segment that determines the position of a symbol.

```
interface Anchor {
    x: number;
    y: number
    angle: number;
    segment: ?number; // ?
}
```

### `Glyph`

The bitmap and dimensions of a particular character in a particular font.

```
interface Glyph {
    id: number;
    bitmap: any; // ?
    width: number;
    height: number;
    left: number;
    top: number;
    advance: number; // glyph-specific right-edge padding
}
```

### `CollisionBox`

A rectangular area of the map that is covered by a source feature. Each source feature may have multiple `CollisionBox`es.

```
interface CollisionBox {
    anchorPointX: number;
    anchorPointY: number;
    x1: number;
    y1: number;
    x2: number;
    y2: number;
    maxScale: number;
    featureIndex: number;
    sourceLayerIndex: number;
    bucketIndex: number;
    bbox0: number;
    bbox1: number;
    bbox2: number;
    bbox3: number;
    placementScale: number;
}
```

### `CollisionFeature`

The set of all collision boxes for a source feature.

```js
interface CollisionFeature {
    boxStartIndex: number;
    boxEndIndex: number;
    boxes: any; // ?
}
```

### `CollisionTile`

The set of all `CollisionFeature`s for a map tile

```js
interface CollisionTile {
    grid: GridIndex;
    ignoredGrid: GridIndex;
    angle: number;
    pitch: number;
    rotationMatrix: [number, number, number, number];
    reverseRotationMatrix: [number, number, number, number];
    yStretch: number;
    collisionBoxArray: StructArray<CollisionBox>;
    tempCollisionBox: CollisionBox;
    edges: [CollisionBox, CollisionBox, CollisionBox, CollisionBox];
    minScale: number;
}
```

### `SymbolQuad` in `SymbolQuadsArray`

```js
interface SymbolQuad {
    anchorPointX: number;
    anchorPointY: number;
    tlX: number;
    tlY: number;
    trX: number;
    trY: number;
    blX: number;
    blY: number;
    brX: number;
    brY: number;
    texH: number;
    texW: number;
    texX: number;
    texY: number;
    anchorAngle: number;
    glyphAngle: number;
    maxScale: number;
    minScale: number;
}
```

### `SymbolQuad` otherwise

An icon or glpyh, its coordinates, and its size for rendering

```js
interface SymbolQuad {
    anchorPoint: Point; // ?
    tl: Point;
    tr: Point;
    bl: Point;
    br: Point;
    tex: Object; // ?
    anchorAngle: number;
    glyphAngle: number;
    minScale: number;
    maxScale: number;
}
```

### `Shaping`

A collection of positioned glyphs and their position on screen. Contains multiple orientations of each glyph. The best orientation for the current map bearing is chosen at render time.

```js
interface Shaping {
    positionedGlyphs: Array<PositionedGlyph>;
    text: string;
    top: number;
    bottom: number;
    left: number;
    right: number;
}
```

### `PositionedGlyph`

```js
interface PositionedGlyph {
    codePoint: number;
    x: number;
    y: number;
    glyph: Glyph;
}
```

### `PositionedIcon`

```js
interface PositionedIcon {
    image: any; // ?
    top: numbr;
    bottom: numbr;
    left: numbr;
    right: numbr;
}
```

### `SymbolBucket`

The collision tile, symbol quads, and symbol instances for a map tile

interface SymbolBucket {
    :grimacing:
}

### `SymbolInstance`

```js
interface SymbolInstance {
    textBoxStartIndex: number;
    textBoxEndIndex: number;
    iconBoxStartIndex: number;
    iconBoxEndIndex: number;
    glyphQuadStartIndex: number;
    glyphQuadEndIndex: number;
    iconQuadStartIndex: number;
    iconQuadEndIndex: number;
    anchorPointX: number;
    anchorPointY: number;
    index: number;
}
```

## Misc Terminology

- **box scale** zoom-specific scaling factor used to convert between glyph units and geometry units
- **CJK**: acronym for "Chinese, Japanese, and Korean", the three languages that require top-to-bottom orientation
- **leading**: distance between the baselines of subsequent lines of text (see also `advance`)
- **orientation**: the direction in which a text is rendered: top-to-bottom (CJK-only) or left-to-right
- **shaped icon**: a `Shaping` containing an icon
- **shaped text**: a `Shaping` containing glyphs
- **text feature**: the text string associated with a particular feature

1ec5:
> CJK: acronym for "Chinese, Japanese, and Korean", the three languages that require top-to-bottom orientation

Crossposted from https://github.com/mapbox/mapbox-gl-js/pull/3402#discussion_r83940620: For the purpose of this PR, “CJK” is Hangul, Hanzi/Hanja/Kanji, Hiragana, Katakana, Mongolian, and Yi scripts. However, note that Hangul and Mongolian words are delimited by spaces and thus should retain the Latin-style line breaking algorithm. If vertical, space-delimited Hangul and Mongolian is a problem, they can be horizontal for now.

friedbunny:
Per chat w/@1ec5, for Japanese it would probably be reasonable to exclude names that include romaji (roman characters) from verticalization. It can be done, but horizontal seems to be generally preferred (especially if a name only contains romaji).

We’ll have to contend with [fullwidth variants](https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms), as well.

1ec5:
> GL JS must render CJK glyphs top-to-bottom when appropriate

https://github.com/mapbox/mapbox-gl-js/pull/3402#discussion_r83940620 addresses horizontal/vertical switching based on scripts. 

Even in the context of a Chinese-only map, laying out all labels vertically only makes sense for an archaic-looking style. (Think yellowed background and calligraphic fonts.) A modern-looking style would typically lay out point-placed labels horizontally but fall back to a vertical layout to avoid collision. Meanwhile, line-placed labels would be laid our horizontally or vertically based on the angle of the road, in an attempt to avoid rotating glyphs beyond 45°. (See mapbox/mapbox-gl-native#1682 for further discussion.)

If there’s a need to land this feature before handling those nuances, I recommend placing vertical layout behind a style specification property such as `writing-mode: traditional`. Per-character line breaking (i.e., `word-break: break-all`), as described in mapbox/mapbox-gl-native#1223, remains the highest priority for general-purpose Chinese text, beyond vertical text fallback.

nickidlugash:
> Even in the context of a Chinese-only map, laying out all labels vertically only makes sense for an archaic-looking style. (Think yellowed background and calligraphic fonts.) A modern-looking style would typically lay out point-placed labels horizontally but fall back to a vertical layout to avoid collision. 

We are not implementing vertical labels for point placement. 

> Meanwhile, line-placed labels would be laid our horizontally or vertically based on the angle of the road, in an attempt to avoid rotating glyphs beyond 45°. 

Yes, this is what this PR does.

js/data/bucket/symbol_bucket.js
nickidlugash: @1ec5 thanks, will look at this tomorrow.

js/data/bucket/symbol_bucket.js
```diff
@@ -349,13 +338,6 @@ SymbolBucket.prototype.addFeature = function(lines, shapedText, shapedTextVertic
         for (var j = 0, len = anchors.length; j < len; j++) {
             var anchor = anchors[j];
 
-            /*if (shapedText && isLine) {
-                if (this.anchorIsTooClose(shapedText.text, textRepeatDistance, anchor)) {
-                    continue;
-                 // ^^^ this check is where all the vertical labels are being skipped    
-                }
-            }*/
```
lucaswoj: Thanks for the heads up! I'll restore this in a sec.

js/data/bucket/symbol_bucket.js
lucaswoj: @1ec5 Can we use plain ol' inequalities rather than a regex?

1ec5: Sure, whichever is easier to maintain and more performant. It probably doesn’t make a big difference either way. (Character classes can contain `\uxxxx` character references instead of Unicode literals, if that’s your concern.)

mourner: Note that since we switched to Buble, we can now use `u` `RegExp` flag that allows using any unicode characters in regexps (they get transpiled to `\u....`).

1ec5: But Babel doesn’t change the fact that JavaScript strings (and therefore regular expressions) must represent characters above U+FFFF as surrogate pairs.

lucaswoj:
Rebased and debugged 👉 #3438

PR #3403, opened 2016-10-18
Add a zoom to linestring bounds example

tristen:
Demonstrates chaining [extend](https://www.mapbox.com/mapbox-gl-js/api/#LngLatBounds#extend) to [LngLatBounds](https://www.mapbox.com/mapbox-gl-js/api/#LngLatBounds) as a cheap alternative to using Turf to determine a features bounds

fit-to-bounds

### DIFF
--- a/docs/_posts/examples/3400-01-28-zoomto-linestring.html
+++ b/docs/_posts/examples/3400-01-28-zoomto-linestring.html
@@ -1,0 +1,107 @@
+---
+layout: example
+category: example
+title: Fit to the bounds of a LineString
+description: "Get the bounds of a LineString by passing it's first coordinates to <a href='https://www.mapbox.com/mapbox-gl-js/api/#LngLatBounds'>mapboxgl.LngLatBounds</a> and chaining <a href='https://www.mapbox.com/mapbox-gl-js/api/#LngLatBounds#extend'>extend</a> to include the last coordinates."
+tags:
+  - user-interaction
+---
+<style>
+.btn-control {
+    font:bold 12px/20px 'Helvetica Neue', Arial, Helvetica, sans-serif;
+    background-color: #3386c0;
+    color: #fff;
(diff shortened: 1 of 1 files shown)


docs/_posts/examples/3400-01-28-zoomto-linestring.html
```diff
@@ -0,0 +1,101 @@
...
+            "line-width": 5
+        }
+    });
+
+    document.getElementById('zoomto').addEventListener('click', function() {
+
+        // Geographic coordinates of the linestring
+        var coords = geojson.features[0].geometry.coordinates;
+        var a = coords[0] // First coordinates in the array
+        var b = coords[coords.length - 1] // Last coordinates in the array
```
lucaswoj: This works for the line in this example but may not work in all cases. For example, a line representing a closed shape, like a lake or a racetrack, would not work.

tristen: That's true yeah. Feel free to close this if you think this example leads to confusion rather than a simple solution.

tmcw: I tend to do something like

```js
var bonds = coordinates.reduce(function (bounds, coord) {
  return bounds.extend(coord);
}, new mapboxgl.LatLngBounds(coordinates[0], coordinates[0]));
```

But regardless I fear releasing this example and having people expect it to work with more than one feature or any other kind of geometry.

tristen: Wrapping each coordinates in extend is a good idea! I think this is a good example to demonstrate

> But regardless I fear releasing this example and having people expect it to work with more than one feature or any other kind of geometry.

I'm calling out `linestring` in the example though? I could make it more clear in the comments how to make this work for other geometries?

PR #3408, opened 2016-10-19
Use ES6 class syntax

### LINKED ISSUE #3400: Use native ES6 classes
We should continue #3368 by converting to native ES6 class syntax. There are few things to figure out though:

- Is everyone OK with converting `Evented` from a mixin to a class? Given that it is inherited as a prototype everywhere it's used, this make sense, and we'll be able to use the native `class Foo extends Evented` syntax. It's a breaking change, although not sure if any downstream code depends on this.
- There are a few instances where we have some methods of a class defined in a separate file — specifically, `camera.js` (a part of `Map`) and `use_program.js` (a part of `Painter`). We can leave it as is, but after we convert to ES6 syntax, it would be much cleaner to get rid of prototype mixins like this. For the camera case, perhaps we should make it a composition with `Camera` as a separate class (similar to `Transform`). For the relatively small `use_program`, I would just move it back to `Painter`.

@jfirebaugh @lucaswoj

mourner:
Closes #3400. Some gotchas:

- ES6 class prototypes are not enumerable (e.g. you can't do `util.extend(obj, Foo.prototype)` or use `for prop in obj` to iterate over methods), so all places where we depend on this need to be fixed.
- In particular, I had to remove the magic `util.bindHandlers` in favor of explicit binding with `util.bindAll`.
- `Evented` is now a class rather than a mixin.
- I added `super` calls to constructors of all classes that have a parent class (e.g. `class Map extends Evented`). This is enforced by `eslint:recommended`, and I think this is a good practice, so I decided to fix this instead of turning off the rules.
- ES6 classes don't support prototype properties, and generally this is considered an antipattern. So e.g. I removed `_enabled` prototype property in handlers and replaced `return this._enabled` with `return !!this._enabled`. Another consequence is `util.setOptions` won't work (it depends on `options` prototype object) so we're ditching it.
- `module.exports` can't come before `class ...` — using class before it is defined is a JS error (unlike using function).
- If you have a circular dependency between a ES6 class and a ES6 class that extends it, you have to require child class after `module.exports` in the main class, otherwise the child class will try to extend `undefined` and fail.
- ES6 class syntax doesn't allow method names with spaces (for worker callbacks like `get anchors`), but I don't see any reason not to rename those to `camelCase`.
- I'm removing `@class Foo` jsdoc comments because `documentation` infers that type of thing now.

I suggest reviewing this PR commit by commit, and merging it without squash too, since all commits are self-contained, and the total diff is too big (5k+).

Checklist:

- [x] ~~Make sure the generated docs are still correct.~~ Current docs are broken in master, so I suggest fixing as a follow-up PR.
- [x] Run benchmarks.

Progress:

- [x] data/array_group.js
- [x] data/bucket.js
- [x] data/bucket/circle_bucket.js
- [x] data/bucket/fill_bucket.js
- [x] data/bucket/fill_extrusion_bucket.js
- [x] data/bucket/line_bucket.js
- [x] data/bucket/symbol_bucket.js
- [x] data/buffer.js
- [x] data/buffer_group.js
- [x] data/feature_index.js
- [x] geo/coordinate.js
- [x] geo/lng_lat.js
- [x] geo/lng_lat_bounds.js
- [x] geo/transform.js
- [x] render/frame_history.js
- [x] render/line_atlas.js
- [x] render/painter.js
- [x] render/vertex_array_object.js
- [x] source/geojson_source.js
- [x] source/geojson_worker_source.js
- [x] source/geojson_wrapper.js
- [x] source/image_source.js
- [x] source/raster_tile_source.js
- [x] source/source_cache.js
- [x] source/tile.js
- [x] source/tile_coord.js
- [x] source/vector_tile_source.js
- [x] source/vector_tile_worker_source.js
- [x] source/video_source.js
- [x] source/worker.js
- [x] source/worker_tile.js
- [x] style/animation_loop.js
- [x] style/image_sprite.js
- [x] style/light.js
- [x] style/style.js
- [x] style/style_layer.js
- [x] style/style_layer/background_style_layer.js
- [x] style/style_layer/circle_style_layer.js
- [x] style/style_layer/fill_style_layer.js
- [x] style/style_layer/line_style_layer.js
- [x] style/style_layer/raster_style_layer.js
- [x] style/style_layer/symbol_style_layer.js
- [x] style/style_layer_index.js
- [x] style/style_transition.js
- [x] symbol/anchor.js
- [x] symbol/collision_feature.js
- [x] symbol/collision_tile.js
- [x] symbol/glyph_atlas.js
- [x] symbol/glyph_source.js
- [x] symbol/sprite_atlas.js
- [ ] ui/camera.js
- [x] ui/control/attribution_control.js
- [x] ui/control/control.js
- [x] ui/control/geolocate_control.js
- [x] ui/control/navigation_control.js
- [x] ui/control/scale_control.js
- [x] ui/handler/box_zoom.js
- [x] ui/handler/dblclick_zoom.js
- [x] ui/handler/drag_pan.js
- [x] ui/handler/drag_rotate.js
- [x] ui/handler/keyboard.js
- [x] ui/handler/scroll_zoom.js
- [x] ui/handler/touch_zoom_rotate.js
- [x] ui/hash.js
- [x] ui/map.js
- [x] ui/marker.js
- [x] ui/popup.js
- [x] util/actor.js
- [x] util/dictionary_coder.js
- [x] util/dispatcher.js
- [x] util/evented.js
- [x] util/lru_cache.js
- [ ] util/struct_array.js
- [x] util/vectortile_to_geojson.js
- [x] util/worker_pool.js

сс @jfirebaugh @lucaswoj @mollymerp


mourner:
Updating the top comment with new gotchas as I go. There are quite a few, but all are reasonable.

mourner:
Benchmarks (freezed on geojson-setdata-large master, so slightly incomplete):

**map-load**
master 15960ff: 1,345 ms
es6-classes d261612: 184 ms

**style-load**
master 15960ff: 196 ms
es6-classes d261612: 188 ms

**buffer**
master 15960ff: 1,193 ms
es6-classes d261612: 1,212 ms

**fps**
master 15960ff: 60 fps
es6-classes d261612: 59 fps

**frame-duration**
master 15960ff: 10.2 ms, 15% > 16ms
es6-classes d261612: 10.3 ms, 15% > 16ms

**query-point**
master 15960ff: 1.07 ms
es6-classes d261612: 1.13 ms

**query-box**
master 15960ff: 103.44 ms
es6-classes d261612: 106.99 ms

**geojson-setdata-small**
master 15960ff: 11 ms
es6-classes d261612: 8 ms

mourner:
This is ready for review. I'm leaving out `struct_array.js` (will require a non-trivial rewrite because it's currently too complex) and also `camera.js` mixin (better as a follow-up PR, with discussion on how to compose it better). The docs are not building, but they're broken in master too, so I'll leave fixing docs and making sure they're correct to a follow-up PR too.

jfirebaugh:
None of the gotchas seem particularly concerning to me.

This will definitely cause merge conflicts. The one PR I'd call out specifically is #3402, because it's on a deadline.

mourner:
@jfirebaugh yeah, I specifically reviewed that one to make sure that it'll be relatively easy to rebase — I can help with that after this PR is merged.

jfirebaugh:
Great. I'm 👍  here, but will let @lucaswoj do the review approval and final say on when to time the merge.

Issue #3386, opened 2016-10-14
Unable apply runtime styles for extrusion properties


scothis:
**mapbox-gl-js version**: 0.26.0
### Steps to Trigger Behavior
1. load a style with a fill layer that has no extrusions properties set
2. `map.setPaintProperty(fillLayerId, 'fill-extrude-height', 1);`

example http://codepen.io/anon/pen/yaqRpK
### Expected Behavior

The style should render the new value and turn the layer from 2d fills into 3d extrusions.
### Actual Behavior

The layer disappears, sometimes there is an error on the console right away, sometimes you have to move the map before you see a console error.

<img width="530" alt="screen shot 2016-10-14 at 6 30 40 pm" src="https://cloud.githubusercontent.com/assets/302992/19404502/5aec9e7c-923c-11e6-8f73-1f34aed8f1a7.png">

lbud:
There's also a little bug in here that affects `setPaintProperty('fill-extrude-*')` when a layer is already extruded; I pushed a fix in https://github.com/mapbox/mapbox-gl-js/pull/3414.

### problem

The bigger problem you're seeing @scothis is that [in `bucket.js`](https://github.com/mapbox/mapbox-gl-js/blob/85097c3ec03abcf27da78c37ac3a2a1bb651c20c/js/data/bucket.js#L80-L83), we switch bucket types based on whether we infer that a layer _may_ be extruded — so, if a fill layer's `fill-extrude-height` property is either a function or is a value greater than 0, we use a `fillextrusion` bucket, which has `5n` vertices (where `n` = number of vertices of the source polygon); otherwise we use a `fill` bucket with `n` vertices. In painter we do such a switch as well.

When you see the polygons disappear, this is because the parsed fill buckets now have a style that sends them down the `draw_extrusion` render path; when you move the map and it tries to reparse buckets based on this modified style it gets even more confused 😞 

### solution

We would never want to quintuple the number of vertices for all fill layers, so **I think the answer here may be** to add a new `layout_fill` boolean (something like `can-extrude`) that indicates more explicitly to the worker which bucket type to create for a layer. Fill layers created with `can-extrude=true` and `fill-extrude-height=0` would then be less performant than regular fill layers. We will also likely have to do some trickery to ensure that flat extruded layers are rendered like fills (unaffected by lighting) to keep them consistent with other flat fill layers.

### hack 

** however **

[...]

PR #3414, opened 2016-10-19
Minor fix for 3386 (provide empty object for lookup)

lbud:
This is only a small part of the fix for #3386 (more details on that ticket) — `getPaintValue` does an object lookup on `globalProperties`, so without providing an object for lookup this was sometimes throwing.

### DIFF
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -282,7 +282,7 @@ Bucket.create = function(options) {
     let type = options.layer.type;
     if (type === 'fill' && (!options.layer.isPaintValueFeatureConstant('fill-extrude-height') ||
         !options.layer.isPaintValueZoomConstant('fill-extrude-height') ||
-        options.layer.getPaintValue('fill-extrude-height', {}) !== 0)) {
+        options.layer.getPaintValue('fill-extrude-height', {zoom: this.zoom}) !== 0)) {
         type = 'fillextrusion';
     }
     return new subclasses[type](options);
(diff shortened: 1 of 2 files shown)


lbud:
@lucaswoj can you 🍏  this? 🙏

js/data/bucket.js
```diff
@@ -79,7 +79,7 @@ class Bucket {
         let type = options.layer.type;
         if (type === 'fill' && (!options.layer.isPaintValueFeatureConstant('fill-extrude-height') ||
             !options.layer.isPaintValueZoomConstant('fill-extrude-height') ||
-            options.layer.getPaintValue('fill-extrude-height') !== 0)) {
+            options.layer.getPaintValue('fill-extrude-height', {}) !== 0)) {
```
lucaswoj: Would we head off similar bugs if we added this default value further upstream? (such as within `StyleLayer#getPaintValue`)

lucaswoj: Alternately, perhaps we _should_ be passing a `globalProperties` object with a zoom here and 👇

lbud:
Ok, spun around in circles a bit here before figuring out there's a related bug that affects all properties (#3425) but here I switched from the empty objects to providing zoom like in other globalProperties. @lucaswoj 🍏?

js/data/bucket.js
```diff
@@ -282,7 +282,7 @@ Bucket.create = function(options) {
     let type = options.layer.type;
     if (type === 'fill' && (!options.layer.isPaintValueFeatureConstant('fill-extrude-height') ||
         !options.layer.isPaintValueZoomConstant('fill-extrude-height') ||
-        options.layer.getPaintValue('fill-extrude-height') !== 0)) {
+        options.layer.getPaintValue('fill-extrude-height', {zoom: this.zoom}) !== 0)) {
```
jfirebaugh: This doesn't look right -- `this` here will be `Bucket` the class/constructor function, not a `Bucket` instance.

lbud: Oops, good catch @jfirebaugh. Following up in https://github.com/mapbox/mapbox-gl-js/commit/79dd26a2cebfb9a6b0543aa2bdf8672cd179339a

Issue #3237, opened 2016-09-21
Refactor benchmark suite to collect statistically significant results


lucaswoj:
> Note that map-load benchmarks pretty much always has the branch time much lower than the master one, making this particular benchmark not very useful. Maybe it's because V8 warms up on the first run.

https://github.com/mapbox/mapbox-gl-js/pull/3224#issuecomment-248438078

cc @mourner @jfirebaugh

jfirebaugh:
Can we use [Benchmark.js](https://benchmarkjs.com/)? It handles repeating the benchmarked code until it's able to collect a statistically significant result.

lucaswoj:
@jfirebaugh Sounds like a good fit! I'll look into it.

lucaswoj:
After toying with this a little, I think using `benchmark.js` specifically is more trouble than its worth. `benchmark.js` does not work well with asynchronous  setup / teardown or non-time-based tests. We should be able to implement our own stat sig metrics.

PR #3420, opened 2016-10-20
Add support for ideographic text breaking

xrwang:
## Launch Checklist

Splitting out from https://github.com/mapbox/mapbox-gl-js/pull/3402 for clearer tracking of two separate issues. 

This PR adds support for:
- CJK line breaking
## Requirements
- GL JS must line break for point labels in CJK even if there are no spaces in the label
## Specifications
- we will use naïve "balanced" breaking 
- we will enable / disable naive breaking based on language detection
- we will use character ranges for language detection ([data](https://en.wiktionary.org/wiki/Module:scripts/data))
## Launch Checklist
- [x] fix failing unit tests 
- [x] write tests
- [x] remove whitespace & debug page changes from diff where appropriate
- [x] post benchmark scores
- [ ] merge https://github.com/mapbox/mapbox-gl-test-suite/pull/153

### DIFF
--- a/debug/chinese.html
+++ b/debug/chinese.html
@@ -50,8 +50,8 @@ map.on('load', function () {
         "type": "symbol",
         "source": "points",
         "layout": {
+            "icon-image": "{icon}-15",
             "text-field": "{name}",
-            "text-max-width": 2,
             "text-font": ["Open Sans Semibold", "Arial Unicode MS Bold"],
             "text-offset": [0, 0.6],
             "text-anchor": "top",
--- a/debug/chinese.html
+++ b/debug/chinese.html
@@ -50,8 +50,8 @@ map.on('load', function () {
         "type": "symbol",
         "source": "points",
         "layout": {
-            "icon-image": "{icon}-15",
             "text-field": "{name}",
+            "text-max-width": 2,
             "text-font": ["Open Sans Semibold", "Arial Unicode MS Bold"],
             "text-offset": [0, 0.6],
             "text-anchor": "top",
--- a/js/symbol/shaping.js
+++ b/js/symbol/shaping.js
@@ -118,8 +118,6 @@ function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, vertic
     const positionedGlyphs = shaping.positionedGlyphs;
 
     if (maxWidth) {
-        const wordLength = positionedGlyphs.length;
-
         if (allowsIdeographicBreaking) {
             const lastPositionedGlyph = positionedGlyphs[positionedGlyphs.length - 1];
             const estimatedLineCount = Math.max(1, Math.ceil(lastPositionedGlyph.x / maxWidth));
@@ -159,9 +157,7 @@ function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, vertic
                 line++;
             }
 
-            if (allowsIdeographicBreaking) {
-                lastSafeBreak = i
-            } else if (breakable[positionedGlyph.codePoint]) {
+            if (allowsIdeographicBreaking || breakable[positionedGlyph.codePoint]) {
                 lastSafeBreak = i;
             }
         }
--- a/js/symbol/shaping.js
+++ b/js/symbol/shaping.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const scriptDetection = require('../util/script_detection');
+
 module.exports = {
     shapeText: shapeText,
     shapeIcon: shapeIcon
@@ -54,7 +56,7 @@ function shapeText(text, glyphs, maxWidth, lineHeight, horizontalAlign, vertical
 
     if (!positionedGlyphs.length) return false;
 
-    linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, verticalAlign, justify, translate);
+    linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, verticalAlign, justify, translate, scriptDetection.allowsBalancedBreaking(text));
 
     return shaping;
 }
@@ -105,7 +107,7 @@ const breakable = {
 
 invisible[newLine] = breakable[newLine] = true;
 
-function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, verticalAlign, justify, translate) {
+function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, verticalAlign, justify, translate, allowsBalancedBreaking) {
     let lastSafeBreak = null;
     let lengthBeforeCurrentLine = 0;
     let lineStartIndex = 0;
@@ -150,14 +152,7 @@ function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, vertic
                 line++;
             }
 
-            if (positionedGlyph.codePoint > 19968) {
-                if (breakable[positionedGlyph.codePoint]) {
-                    lastSafeBreak = i - 1;
-                }
-                if (!(breakable[positionedGlyph.codePoint])) {
-                    lastSafeBreak = Math.round(wordLength / 3);
-                }
-            }   else if (breakable[positionedGlyph.codePoint]) {
+            if (allowsBalancedBreaking || breakable[positionedGlyph.codePoint]) {
                 lastSafeBreak = i;
             }
         }
--- a/js/symbol/shaping.js
+++ b/js/symbol/shaping.js
@@ -56,7 +56,7 @@ function shapeText(text, glyphs, maxWidth, lineHeight, horizontalAlign, vertical
 
     if (!positionedGlyphs.length) return false;
 
-    linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, verticalAlign, justify, translate, scriptDetection.allowsIdeographicBreakingBreaking(text));
+    linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, verticalAlign, justify, translate, scriptDetection.allowsIdeographicBreaking(text));
 
     return shaping;
 }
@@ -107,7 +107,7 @@ const breakable = {
 
 invisible[newLine] = breakable[newLine] = true;
 
-function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, verticalAlign, justify, translate, allowsIdeographicBreakingBreaking) {
+function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, verticalAlign, justify, translate, allowsIdeographicBreaking) {
     let lastSafeBreak = null;
     let lengthBeforeCurrentLine = 0;
     let lineStartIndex = 0;
@@ -119,6 +119,13 @@ function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, vertic
 
     if (maxWidth) {
         const wordLength = positionedGlyphs.length;
+
+        if (allowsIdeographicBreaking) {
+            const lastPositionedGlyph = positionedGlyphs[positionedGlyphs.length - 1];
+            const estimatedLineCount = Math.max(1, Math.ceil(lastPositionedGlyph.x / maxWidth));
+            maxWidth = lastPositionedGlyph.x / estimatedLineCount;
+        }
+
         for (let i = 0; i < positionedGlyphs.length; i++) {
             const positionedGlyph = positionedGlyphs[i];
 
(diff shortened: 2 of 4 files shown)


js/symbol/shaping.js
```diff
@@ -123,7 +150,14 @@ function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, vertic
                 line++;
             }
 
-            if (breakable[positionedGlyph.codePoint]) {
+            if (positionedGlyph.codePoint > 19968) {
```
mourner: Magic number alert. We should add a comment about this and/or make it a constant.

js/symbol/shaping.js
```diff
@@ -123,7 +150,14 @@ function linewrap(shaping, glyphs, lineHeight, maxWidth, horizontalAlign, vertic
...
+                if (breakable[positionedGlyph.codePoint]) {
+                    lastSafeBreak = i - 1;
+                }
+                if (!(breakable[positionedGlyph.codePoint])) {
+                    lastSafeBreak = Math.round(wordLength / 3);
```
mourner: Here too. Why is it divided by 3?

debug/chinese.html
```diff
@@ -37,7 +37,6 @@
     hash: true
 });
 
-map.addControl(new mapboxgl.Navigation());
```
mourner: Not intentional?

js/symbol/shaping.js
```diff
@@ -65,16 +65,42 @@ const invisible = {
...
 
 const breakable = {
-    0x20:   true, // space
-    0x26:   true, // ampersand
-    0x2b:   true, // plus sign
-    0x2d:   true, // hyphen-minus
-    0x2f:   true, // solidus
-    0xad:   true, // soft hyphen
-    0xb7:   true, // middle dot
+    0x0020: true, // space
```
lucaswoj: Fortunately we never need to iterate over this object. We just do lookups into it. That should be the fastest solution 👍

js/symbol/shaping.js
lucaswoj: I reckon we will replace this with a modified version of the regex in #3438

1ec5:
See https://github.com/mapbox/mapbox-gl-js/pull/3438/files#diff-0612e4d64682542fe1be64751bc484a3 for a work-in-progress regular expression that covers all the traditionally top-to-bottom scripts. This regular expression largely overlaps with the scripts that also need “balanced” ideographic breaking, but Hangul (`'[ᄀ-ᇿ]'`, `'[가-힣]'`, `'[ㄱ-ㆎ]'`) and Mongolian (`[᠀-ᢪ]`) need to be removed.

js/symbol/shaping.js
```diff
@@ -65,21 +67,47 @@ const invisible = {
...
+    0xfe59: true, // small left parenthesis
+    0xfe5b: true, // small left curly bracket
+    0xff04: true, // fullwidth dollar sign
+    0xff08: true, // fullwidth left parenthesis
+    0xff0e: true, // fullwidth full stop
+    0xff3b: true, // fullwidth left square bracket
+    0xff5b: true, // fullwidth left curly bracket
+    0xff5e: true, // fullwidth tilde
+    0xffe1: true, // fullwidth pound sign
+    0xffe5: true  // fullwidth yen sign
```
lucaswoj: This lookup table isn't strictly necessary anymore, though it seems like an improvement

js/util/script_detection.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const ideographicBreakingRegExp = new RegExp([
```
1ec5: If #3438 lands first, `verticalRegExp` and `ideographicBreakingRegExp` will have some overlap. To wit:

* Hangul and Mongolian are vertical but don’t get ideographic breaking.
* Halfwidth forms and Yi are horizontal but do get ideographic breaking.

js/symbol/shaping.js
1ec5: I think the changes to this table should be reverted. Now it’s possible for Latin text to break on either side of an dollar sign, for instance:

```
Ye Old $
99 Store
```

Furthermore, in ideographic scripts, characters like U+3016 left white lenticular bracket (〖) visually combine a space and a punctuation mark, so it can only break on one side (in this case, on the left but not the right).

As things stand in this PR, the CJK punctuation in this table has no effect on ideographic breaking, since breaking can occur on either side of any character. Indeed, we will eventually want to avoid orphaning CJK punctuation, but I don’t think this PR accomplishes that yet.

xrwang: @1ec5, 👌 makes sense, will revert.

js/util/script_detection.js
lucaswoj: Would it make sense to break this regex up into a set of shared strings? For example:

```js
const hangulRegExpString = '...'
const halfwidthFormsRegExpString = '...'

const ideographicBreakingRegExp = new RegExp([
    hangulRegExpString,
    halfwidthFormsRegExpString
].join('|'));
```

I'd be happy to do so. Would you mind defining out which character ranges belong to which sets?

js/symbol/shaping.js
1ec5: What's the rationale for breaking at quotation marks or apostrophes? This would result in labels like "Linens 'n Things" turning into:

```
Linens '
n Things
```

lucaswoj: Good point. This is my mistake. Reverting to the original set now.

lucaswoj:
@1ec5 I think I've addressed all your feedback and my todos. Anything else you'd like to see before this 🚢s? 

@jfirebaugh @mourner @ansis Would you like to 👀 this PR?

package.json
```diff
@@ -56,7 +56,7 @@
     "highlight.js": "9.3.0",
     "jsdom": "^9.4.2",
     "lodash.template": "^4.4.0",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#28c76c64e8cfcee8764c6c0f6d4fcc2d15a8d1e1",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#77b281c9e6225471505f7daefa76806a8fbf22e2",
```
jfirebaugh: Need a rebase and merge of the mapbox-gl-test-suite branch.

lucaswoj:
benchmark | master e03a7b1 | cjk-break2 7462ab1
--- | --- | ---
**map-load** | 171 ms  | 179 ms 
**style-load** | 197 ms  | 147 ms 
**buffer** | 1,068 ms  | 1,200 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 6.4 ms, 1% > 16ms  | 6.8 ms, 1% > 16ms 
**query-point** | 1.25 ms  | 1.16 ms 
**query-box** | 82.52 ms  | 89.32 ms 
**geojson-setdata-small** | 10 ms  | 7 ms 
**geojson-setdata-large** | 316 ms  | 346 ms

ansis:
How does this handle cases where the text contains both ideographic glyphs and non-ideographic ones? `allowsIdeographicBreaking` looks like it returns true in these cases. Does this mean that linebreaks can occur anywhere in these cases?

1ec5:
> How does this handle cases where the text contains both ideographic glyphs and non-ideographic ones? `allowsIdeographicBreaking` looks like it returns true in these cases. Does this mean that linebreaks can occur anywhere in these cases?

Yes, and that goes for #3438 as well. That’s why I originally proposed a negation in https://github.com/mapbox/mapbox-gl-js/pull/3402#discussion_r83940620 that would only affect labels that consisted entirely of ideographic characters.

lucaswoj:
> Yes, and that goes for #3438 as well. That’s why I originally proposed a negation in #3402 (comment) that would only affect labels that consisted entirely of ideographic characters.

Ah. I see. I did not understand the intent behind that code. I will adjust this PR to accommodate.

js/util/script_detection.js
```diff
@@ -0,0 +1,18 @@
...
+    '\uD840[\uDC00-\uFFFF]|[\uD841-\uD872]|\uD873[\u0000-\uDEAF]', // '[𠀀-𬺯]'
+    '[！-￮]',
+    '[ぁ-ゟ]',
+    '[゠-ヿ]',
+    '[ㇰ-ㇿ]',
+    '[ꀀ-꓆]'
+].join('|')})+$`);
+
+module.exports.allowsIdeographicBreaking = function(input) {
+    return input.search(ideographicBreakingRegExp) !== -1;
```
1ec5: Now that a match for `ideographicBreakingRegExp` has `^` and `$` anchors, spanning the entire input string, this should be a call to `match()`. But I wonder if it’s even necessary to match the entire string. Searching for `[^一-鿌…]` instead would be more performant for strings that contain no CJK. The tricky part is the surrogate pairs, but it should definitely be possible to negate that part of the regular expression too.

1ec5:
Concretely, the problem is bilingual labels or those that include romanizations, like [this POI](http://www.openstreetmap.org/node/2666815562), named `曾大屋新村 Tsang Tai Uk New Village`. We wouldn’t want it become:

```
曾大屋新村
Tsang
Tai Uk
New Vi
llage
```

On the other hand, there are legitimate reasons for a CJK label to contain non-CJK characters, like (hypothetically) a subway station named `施氏食狮史（ 1A口）`.

What this PR calls “ideographic breaking” is a combination of two features: breaking on any character, and balancing lines when breaking. The two naturally go hand in hand for purely CJK labels, but bilingual labels force us to distinguish between them.

If we’re conservative and err on the side of applying word breaking throughout, then the worst that could happen is a label that collides other labels out. If we’re too aggressive and err on the side of applying CJK breaking throughout, then we risk breaking in the middle of non-CJK words, which looks amateurish.

Ideally, in the presence of non-CJK characters, we’d continue to break on any CJK character but go back to word breaking for non-CJK words. @lucaswoj has pushed an implementation that seems to do this quite well.

mourner:
> buffer	1,068 ms	  1,200 ms

Is this due to flaky benchmark, or a real regression? I wouldn't expect a perf regression on an latin-character area which we use in the bench. We can still merge but should follow-up if it's a real regression.

lucaswoj:
We might expect some slowdown in Latin areas because we need to check each label for ideographic characters. In the case of Latin-only labels, e1eee1b improves this from a `O(number of characters)` to `O(number of labels)` check.

Here are a couple of benchmark runs. We sorely need #3237. 

benchmark | master baa2b8d | cjk-break2 0f3ed22
--- | --- | ---
**buffer** | 889 ms  | 901 ms 


benchmark | master baa2b8d | cjk-break2 0f3ed22
--- | --- | ---
**buffer** | 901 ms  | 878 ms 

benchmark | master baa2b8d | cjk-break2 0f3ed22
--- | --- | ---
**buffer** | 879 ms  | 893 ms

lucaswoj:
@1ec5 are you ready to bless this with a ✅?

mourner:
> We might expect some slowdown in Latin areas because we need to check each label for ideographic characters.

@lucaswoj The bench results look great now! O(number of labels) is very fast. If we want to improve the check further, maybe we could bail out early on latin characters (e.g. `if (char <= ...) return false;` on the first line).

js/util/script_detection.js
```diff
@@ -0,0 +1,45 @@
...
+    if (char >= 0x3400 && char <= 0x4DB5) return true;
+
+    // eslint-disable-next-line no-irregular-whitespace
+    // "　" to "〿"
+    if (char >= 0x3000 && char <= 0x303F) return true;
+
+    // "𠀀" to "𬺯"
+    if (char === 0xD840 && nextChar >= 0xDC00) return true;
+    if (char >= 0xD841 && char <= 0xD872) return true;
+    if (char === 0xD873 && nextChar <= 0xDEAF) return true;
```
lucaswoj: I've realized that there are a few bugs in our handling of surrogate pairs. Because we don't actually have the ability to render surrogate pairs as glyphs, I can't make a test case. I'm thinking we should remove this range altogether.

1ec5:
A C++ port is ready in mapbox/mapbox-gl-native#6828. In that PR, I took another look at the exact character ranges we treat as ideographic. I decided to add some additional Unicode code blocks from [this Wikipedia table](https://en.wikipedia.org/wiki/Template:CJK_ideographs_in_Unicode) – just the BMP blocks, since we don’t support the SIP for glyphs yet (mapbox/mapbox-gl#29). Additionally, I expanded each range to cover the entire block, not just the currently assigned code points, to future-proof the code a bit. (Future versions of Unicode may assign more characters within these blocks and also add new blocks. However, new CJK blocks would likely fall within the SIP.)

Feel free to synchronize this PR with that one.

lucaswoj:
@1ec5 Done in https://github.com/mapbox/mapbox-gl-js/commit/b2cd6caa5c2a8b6811934e8c7d2993b638e65d9e. I'm going to rebase this branch and then 🚢.

Issue #3357, opened 2016-10-12
Refactor to clearly differentiate between "interaction handler" events and "camera mutation" events


lucaswoj:
Some interaction handlers fire events that are named as if they originate the `Camera` rather than the interaction handler. The separation between these two types of events should be explicit. 

 - `DragPanInteraction` fires `movestart`
 - `DragRotateInteraction` fires `rotate`, `move`, `rotatestart`, `movestart`, `rotateend`, `moveend`

Related to #3068 #2792

mourner:
Can you clarify a bit? The two items above seem unrelated to the issue title — which events would you consider "interaction handler" and which "camera mutation"?

lucaswoj:
I would consider all the above events to represent "camera mutations" not "interaction handler" events. Those events should be fired by `Camera` exclusively. 

 - `BoxZoomInteraction` fires `boxzoomstart` and `boxzoomend` events. :+1:
 - `Camera` fires `movestart` and `moveend` 👍 
 - `DragPanInteraction` also fires `movestart` and `moveend` :no_good_man: It should fire events called `dragpanstart` and `dragpanend`

mourner:
Agreed, thanks for clearing this up.

pke:
Is this being worked on? Having a list update while pinch zooming is in progress is really slow, despite using the `moveend` event.

mollymerp:
@pke adding this to my list of eventing chores here #4253 😅

Issue #3416, opened 2016-10-19
Fix broken JSDoc


mourner:
`npm run start-docs` currently errors out because of some mismatched parameter names in jsdoc comments.

PR #3421, opened 2016-10-20
Convert Camera to ES6 class syntax and inherit Map from it

mourner:
I was trying to make `Camera` a an object used from `Map`, as discussed in #3408, but it got really messy, with either lots of code duplication or splitting similar responsibilities between `Map` and `Camera`, and it also required rewriting tons of camera tests. 

So I'm proposing the make `Camera` the base class of `Map` instead. This way it looks much cleaner, is a clear improvement over the current state (mixing in `Camera.prototype`), and doesn't require as many changes. We can follow up on any improvements in subsequent PRs (specifically, balancing out responsibilities between `Camera`, `Transform` and `Map`, and camera event refactoring #3357).

In addition to the change above, I made some minor cleanups in `Camera` code and also moved all camera events there.

👀 @lucaswoj @jfirebaugh


mourner:
@jfirebaugh per discussion in chat, rebased the PR to only contain the Camera ES6 class changes so that the diff is easy to read. Will follow-up with other changes in a different PR. Note that I added `@memberof Map` to all `Camera` methods so that it still shows up under `Map` while being inherited (didn't test it though since #3416 needs to be fixed first).

PR #3438, opened 2016-10-21
Add support for rendering CJK in a vertical writing mode along line-placed features

### LINKED ISSUE #1246: Orient CJK glyphs vertically along vertically oriented lines

lucaswoj:
replaces https://github.com/mapbox/mapbox-gl-js/pull/3402 (HUGE ❤️ to @nickidlugash & @xrwang)
fixes #1246
see also https://github.com/mapbox/mapbox-gl-native/issues/1682

actual

expected

cc @nickidlugash @xrwang @ansis @1ec5 @willwhite @jakepruitt @jfirebaugh @mourner 
## Requirements
- GL JS must render CJK text in a vertical writing mode along vertical lines, as is conventional in cartographic design

vertical-label-rotation
## Specifications
- we will enable / disable the vertical writing mode by script detection
- we will use character ranges for script detection ([data](https://en.wiktionary.org/wiki/Module:scripts/data))
- if a single label has mixed scripts, we will use vertical writing mode

## Open Questions

- How can we test our script detection code?
- How can we make our script detection code more readable?
- Can we improve upon our character centerline alignment?
- Can we add proper vertical advance to our glyph metadata in the near future? 
- Can we pack the symbol structs more efficiently? 

## Launch Checklist

- [x] align glyphs by their centerline in vertical writing mode
- [x] fix bug causing vertical writing mode labels to disappear at certain bearings
- [x] fix failing unit tests 
- [x] rebase onto master (copy-paste style 🍝 )
- [x] support enabling / disabling vertical writing mode based on script detection
- [x] write test-suite tests for writing modes
- [x] refactor to unify script detection logic
- [x] update script detection logic per comments in PR
- [x] rotate non-CJK glyphs embedded in vertical writing mode labels https://github.com/mapbox/mapbox-gl-js/issues/3506
 - [x] rotate punctuation in labels in vertical writing mode https://github.com/mapbox/mapbox-gl-js/issues/3505
 - [x] fix rendering of `_` characters in vertical writing mode 
 - [x] investigate off-center line labels
 - [x] investigate rendering bugs when changing bearing #3582 
 - [x] add more fullwidth punctuation per https://github.com/mapbox/mapbox-gl-js/pull/3438#discussion_r87310327
 - [ ] add solidus test per https://github.com/mapbox/mapbox-gl-test-suite/pull/152#issuecomment-259519290
 - [ ] post benchmark scores


js/data/bucket/symbol_bucket.js
```diff
@@ -257,12 +258,24 @@ class SymbolBucket extends Bucket {
...
-            let shapedText;
+
+            let shapedTextOrientations;
             if (feature.text) {
-                shapedText = shapeText(feature.text, stacks[fontstack], maxWidth,
-                        lineHeight, horizontalAlign, verticalAlign, justify, spacing, textOffset);
+                const requiresVerticalWritingMode = (
+                    // eslint-disable-next-line
+                    feature.text.match(/[\u1100-\u11FF\uAC00-\uD7A3\u3131-\u318E\u4E00-\u9FCC\u3400-\u4DB5\u3000-\u303F\uFF01-\uFFEE\u3041-\u309F\u30A0-\u30FF\u31F0-\u31FF\uA000-\uA4C6\u1800-\u18AA]/) ||
+                    feature.text.match(/(\uD840\uDC00-\uFFFF)|(\uD841-\uD872)|(\uD873\u0000-\uDEAF)/)
```
lucaswoj: @1ec5 Any ideas on how we can construct an exhaustive test suite for this regex?

js/symbol/shaping.js
```diff
@@ -1,41 +1,48 @@
 'use strict';
 
+const WritingMode = {
+    horizantal: 1,
```
1ec5: s/horizantal/horizontal/

js/data/bucket/symbol_bucket.js
```diff
@@ -257,12 +258,23 @@ class SymbolBucket extends Bucket {
...
         for (const feature of this.features) {
-            let shapedText;
+
+            let shapedTextOrientations;
             if (feature.text) {
-                shapedText = shapeText(feature.text, stacks[fontstack], maxWidth,
-                        lineHeight, horizontalAlign, verticalAlign, justify, spacing, textOffset);
+                const requiresVerticalWritingMode = (
+                    feature.text.match(/[\u1100-\u11FF\uAC00-\uD7A3\u3131-\u318E\u4E00-\u9FCC\u3400-\u4DB5\u3000-\u303F\uFF01-\uFFEE\u3041-\u309F\u30A0-\u30FF\u31F0-\u31FF\uA000-\uA4C6\u1800-\u18AA]/) ||
+                    feature.text.match(/(\uD840[\uDC00-\uFFFF])|[\uD841-\uD872]|(\uD873[\u0000-\uDEAF])/)
```
1ec5: For possibly better performance, combine these two regular expressions with `|`. Also, use `search()` instead of `match()`, since you don’t really care about all the individual CJK characters. Finally, please indicate the unescaped characters (perhaps as a table mapping Unicode codepoints to literals) in a comment so we don’t have to look up each individual character when making changes.

mourner: Also, I would make them constants outside of the function scope/loops, because constructing a regex takes a bit of time.

js/symbol/shaping.js
lucaswoj: Done

js/util/script_detection.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const isVerticalWritingModeRegExp = new RegExp([
+    '[ᄀ-ᇿ]',
+    '[가-힣]',
+    '[ㄱ-ㆎ]',
+    '[一-鿌]',
+    '[㐀-䶵]',
+    // eslint-disable-next-line no-irregular-whitespace
+    '[　-〿]',
```
1ec5: This matches everything from U+0020 (space) to U+303F, including the entire ASCII alphabet.

js/util/script_detection.js
```diff
@@ -0,0 +1,22 @@
...
+const verticalRegExp = new RegExp([
+    '[ᄀ-ᇿ]',
+    '[가-힣]',
+    '[ㄱ-ㆎ]',
+    '[一-鿌]',
+    '[㐀-䶵]',
+    // eslint-disable-next-line no-irregular-whitespace
+    '[　-〿]',
+    '\uD840[\uDC00-\uFFFF]|[\uD841-\uD872]|\uD873[\u0000-\uDEAF]', // '[𠀀-𬺯]'
+    '[！-￮]',
```
1ec5: Per https://github.com/mapbox/mapbox-gl-js/pull/3402#issuecomment-254670626, we should restrict this to fullwidth forms only (`[！-｠￠-￦]`) since halfwidth forms would look quite odd unless doubled up.

/cc @friedbunny

js/util/script_detection.js
lucaswoj: @1ec5 Are you sure? I'm seeing an ideographic space here. Would you prefer the `\u` version in this case?

1ec5: Ah, you’re right. I was seeing a quirk in Gecko’s text selection behavior that causes the [Character Identifier](https://dbaron.org/mozilla/char-identifier/) extension to conflate various spaces. The escaped sequence would be less confusing in this case.

js/util/script_detection.js
```diff
@@ -0,0 +1,22 @@
...
+    '[ぁ-ゟ]',
+    '[゠-ヿ]',
+    '[ㇰ-ㇿ]',
+    '[ꀀ-꓆]',
```
1ec5: I was mistaken in including Yi script here. According to [this Wikipedia article](https://en.wikipedia.org/wiki/Yi_script), the Classical Yi syllabary was written vertically, but the Modern Yi syllabary has always been written horizontally. So we should remove this line.

js/data/bucket/symbol_bucket.js
1ec5: @mourner, per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions:

> Regular expression literals provide compilation of the regular expression when the script is loaded. When the regular expression will remain constant, use this for better performance.

So even if we were to keep this regular expression literal here, it would only be compiled once.

1ec5:
I’m not sure how prevalent punctuation is in road names, but various Chinese and Japanese dashes and brackets are [rotated](http://www.unicode.org/charts/PDF/UFE30.pdf) 90° or 180° when laid out vertically. Would it be possible to replace the following characters for vertical text?

Horizontal | Vertical
----|----
。|︒
—|︱
–|︲
_|︳
（|︵
）|︶
｛|︷
｝|︸
〔|︹
〕|︺
〘|︹ (?)
〙|︺ (?)
【|︻
】|︼
《|︽
》|︾
〈|︿
〉|﹀
「|﹁
」|﹂
『|﹃
』|﹄
［|﹇
］|﹈
“|﹁
”|﹂
‘|﹃
’|﹄
…|⋮

js/util/script_detection.js
```diff
@@ -0,0 +1,22 @@
...
+    '[᠀-ᢪ]'
```
1ec5: We should also remove Mongolian from this regular expression. Unlike the other scripts, Mongolian must be written vertically (never horizontally) and its glyphs are always rotated when set vertically. So apart from the fact that we don’t support complex text shaping yet (mapbox/mapbox-gl#4), Mongolian is already laid out correctly along lines from 3:00 to 6:00 and from 9:00 to 12:00. With this PR as is, Mongolian would only be laid out correctly from 3:00 to 4:30 and from 9:00 to 10:30.

If we want to address Mongolian, I think we should do that as tail work. But for that matter, I’m having a difficult time finding any area in OpenStreetMap or the Mapbox Streets source where roads are labeled in Mongolian. So best to leave Mongolian alone.

xrwang: @1ec5 makes sense. Only Inner Mongolia uses the traditional (vertical) Mongolian script, and all of those city names in Inner Mongolia are transliterated into Chinese characters. Mongolia the country uses Cyrillic (consistently LTR and horizontal).

kkaefer:
wow

js/data/bucket/symbol_bucket.js
1ec5: And of course I only now remember that `match` doesn’t look for all matches unless you give the regex a `g` flag.

lucaswoj:
benchmark | master 8127e17 | vertical 893d46c
--- | --- | ---
**map-load** | 221 ms  | 135 ms 
**style-load** | 151 ms  | 159 ms 
**buffer** | 1,326 ms  | 1,122 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 4.4 ms, 0% > 16ms  | 4.9 ms, 0% > 16ms 
**query-point** | 1.07 ms  | 1.09 ms 
**query-box** | 65.20 ms  | 68.88 ms 
**geojson-setdata-small** | 10 ms  | 8 ms 
**geojson-setdata-large** | 308 ms  | 261 ms

lucaswoj:
This is ready for 👀

js/util/is_char_in_unicode_block.js
```diff
@@ -0,0 +1,268 @@
...
+    // 'Cherokee': (char) => char >= 0x13A0 && char <= 0x13FF,
+    // 'Unified Canadian Aboriginal Syllabics': (char) => char >= 0x1400 && char <= 0x167F,
+    // 'Ogham': (char) => char >= 0x1680 && char <= 0x169F,
+    // 'Runic': (char) => char >= 0x16A0 && char <= 0x16FF,
+    // 'Tagalog': (char) => char >= 0x1700 && char <= 0x171F,
+    // 'Hanunoo': (char) => char >= 0x1720 && char <= 0x173F,
+    // 'Buhid': (char) => char >= 0x1740 && char <= 0x175F,
+    // 'Tagbanwa': (char) => char >= 0x1760 && char <= 0x177F,
+    // 'Khmer': (char) => char >= 0x1780 && char <= 0x17FF,
+    'Mongolian': (char) => char >= 0x1800 && char <= 0x18AF,
```
1ec5: Per https://github.com/mapbox/mapbox-gl-js/pull/3438#discussion_r84585841, we don’t want to rotate Mongolian vertically or give it ideographic breaking.

js/util/is_char_in_unicode_block.js
```diff
@@ -0,0 +1,268 @@
...
+    // 'Supplemental Punctuation': (char) => char >= 0x2E00 && char <= 0x2E7F,
+    'CJK Radicals Supplement': (char) => char >= 0x2E80 && char <= 0x2EFF,
+    'Kangxi Radicals': (char) => char >= 0x2F00 && char <= 0x2FDF,
+    'Ideographic Description Characters': (char) => char >= 0x2FF0 && char <= 0x2FFF,
+    'CJK Symbols and Punctuation': (char) => char >= 0x3000 && char <= 0x303F,
+    'Hiragana': (char) => char >= 0x3040 && char <= 0x309F,
+    'Katakana': (char) => char >= 0x30A0 && char <= 0x30FF,
+    // 'Bopomofo': (char) => char >= 0x3100 && char <= 0x312F,
+    'Hangul Compatibility Jamo': (char) => char >= 0x3130 && char <= 0x318F,
+    // 'Kanbun': (char) => char >= 0x3190 && char <= 0x319F,
```
1ec5: Hm, I guess it wouldn’t hurt to make this vertical too.

js/util/is_char_in_unicode_block.js
```diff
@@ -0,0 +1,268 @@
...
+    // 'Cham': (char) => char >= 0xAA00 && char <= 0xAA5F,
+    // 'Myanmar Extended-A': (char) => char >= 0xAA60 && char <= 0xAA7F,
+    // 'Tai Viet': (char) => char >= 0xAA80 && char <= 0xAADF,
+    // 'Meetei Mayek Extensions': (char) => char >= 0xAAE0 && char <= 0xAAFF,
+    // 'Ethiopic Extended-A': (char) => char >= 0xAB00 && char <= 0xAB2F,
+    // 'Latin Extended-E': (char) => char >= 0xAB30 && char <= 0xAB6F,
+    // 'Cherokee Supplement': (char) => char >= 0xAB70 && char <= 0xABBF,
+    // 'Meetei Mayek': (char) => char >= 0xABC0 && char <= 0xABFF,
+    'Hangul Syllables': (char) => char >= 0xAC00 && char <= 0xD7AF,
+    // 'Hangul Jamo Extended-B': (char) => char >= 0xD7B0 && char <= 0xD7FF,
```
1ec5: Might as well make this vertical too.

js/util/is_char_in_unicode_block.js
```diff
@@ -0,0 +1,268 @@
...
+    // 'Modifier Tone Letters': (char) => char >= 0xA700 && char <= 0xA71F,
+    // 'Latin Extended-D': (char) => char >= 0xA720 && char <= 0xA7FF,
+    // 'Syloti Nagri': (char) => char >= 0xA800 && char <= 0xA82F,
+    // 'Common Indic Number Forms': (char) => char >= 0xA830 && char <= 0xA83F,
+    // 'Phags-pa': (char) => char >= 0xA840 && char <= 0xA87F,
+    // 'Saurashtra': (char) => char >= 0xA880 && char <= 0xA8DF,
+    // 'Devanagari Extended': (char) => char >= 0xA8E0 && char <= 0xA8FF,
+    // 'Kayah Li': (char) => char >= 0xA900 && char <= 0xA92F,
+    // 'Rejang': (char) => char >= 0xA930 && char <= 0xA95F,
+    // 'Hangul Jamo Extended-A': (char) => char >= 0xA960 && char <= 0xA97F,
```
1ec5: Might as well make this vertical too.

js/util/is_char_in_unicode_block.js
```diff
@@ -0,0 +1,268 @@
...
+    // 'Hangul Jamo Extended-B': (char) => char >= 0xD7B0 && char <= 0xD7FF,
+    // 'High Surrogates': (char) => char >= 0xD800 && char <= 0xDB7F,
+    // 'High Private Use Surrogates': (char) => char >= 0xDB80 && char <= 0xDBFF,
+    // 'Low Surrogates': (char) => char >= 0xDC00 && char <= 0xDFFF,
+    // 'Private Use Area': (char) => char >= 0xE000 && char <= 0xF8FF,
+    'CJK Compatibility Ideographs': (char) => char >= 0xF900 && char <= 0xFAFF,
+    // 'Alphabetic Presentation Forms': (char) => char >= 0xFB00 && char <= 0xFB4F,
+    // 'Arabic Presentation Forms-A': (char) => char >= 0xFB50 && char <= 0xFDFF,
+    // 'Variation Selectors': (char) => char >= 0xFE00 && char <= 0xFE0F,
+    // 'Vertical Forms': (char) => char >= 0xFE10 && char <= 0xFE1F,
```
1ec5: Make this block vertical and ideographic. This block will also come in handy once we implement https://github.com/mapbox/mapbox-gl-js/pull/3438#issuecomment-255553240.

js/util/is_char_in_unicode_block.js
```diff
@@ -0,0 +1,268 @@
...
+    // 'Ethiopic Extended': (char) => char >= 0x2D80 && char <= 0x2DDF,
+    // 'Cyrillic Extended-A': (char) => char >= 0x2DE0 && char <= 0x2DFF,
+    // 'Supplemental Punctuation': (char) => char >= 0x2E00 && char <= 0x2E7F,
+    'CJK Radicals Supplement': (char) => char >= 0x2E80 && char <= 0x2EFF,
+    'Kangxi Radicals': (char) => char >= 0x2F00 && char <= 0x2FDF,
+    'Ideographic Description Characters': (char) => char >= 0x2FF0 && char <= 0x2FFF,
+    'CJK Symbols and Punctuation': (char) => char >= 0x3000 && char <= 0x303F,
+    'Hiragana': (char) => char >= 0x3040 && char <= 0x309F,
+    'Katakana': (char) => char >= 0x30A0 && char <= 0x30FF,
+    // 'Bopomofo': (char) => char >= 0x3100 && char <= 0x312F,
```
1ec5: This block doesn’t appear in OSM, but it could appear in custom data. It should be vertical and get ideographic breaking.

js/util/is_char_in_unicode_block.js
```diff
@@ -0,0 +1,268 @@
...
+    'Hangul Compatibility Jamo': (char) => char >= 0x3130 && char <= 0x318F,
+    // 'Kanbun': (char) => char >= 0x3190 && char <= 0x319F,
+    // 'Bopomofo Extended': (char) => char >= 0x31A0 && char <= 0x31BF,
```
1ec5: This block doesn’t appear in OSM, but it could appear in custom data. It should be vertical and get ideographic breaking.

lucaswoj:
All requested changes implemented in 85abbdc 🚀

js/util/script_detection.js
```diff
@@ -1,54 +1,89 @@
...
-    // CJK Unified Ideographs Extension A: “㐀” to past “䶵”
-    if (char >= 0x3400 && char <= 0x4DBF) return true;
+    return false;
+};
 
-    // CJK Unified Ideographs: “一” to past “鿕”
-    if (char >= 0x4E00 && char <= 0x9FFF) return true;
+function charAllowsVerticalWritingMode(char) {
+    // Return early for characters outside all ideographic ranges.
+    if (char < 0x2E80) return false;
```
1ec5: This will exclude the Hangul Jamo block. The cutoff should instead be 0x1100.

lucaswoj: :bow: fixed in 97ef972

1ec5:
> Would it be possible to replace the following characters for vertical text?

→ #3505

lucaswoj:
Thanks for the 🍏 @1ec5! I'm going to hold off merging until @nickidlugash gets a chance to look at this branch from a carto/desgin perspective.

nickidlugash:
@lucaswoj @1ec5 I'm still in the middle of reviewing this! Just wanted to note one thing now though:

Vertical labels seem off-centered towards the left of the line. Will it be possible to have them be centered? I haven't had a chance to dig into the code to see where the issue is.

screen shot 2016-11-01 at 7 32 25 pm

*Side note relevant to that screenshot: "/" character collides with all characters above it. This is a common character in one of our relevant data sources. I've seen conflicting information as to whether this is the correct character to use for Chinese in general and specifically for vertical orientation – solidus (U+002F), versus fullwidth solidus (U+FF0F). Either way, I think perhaps without a true vertical advance value this is not getting enough space. The design of this character is one that signficantly [overshoots](https://en.wikipedia.org/wiki/Overshoot_(typography)) the cap height of the font.*

1ec5:
> "/" character collides with all characters above it

This is a good argument for making #3505 a release blocker.

lucaswoj:
> Vertical labels seem off-centered towards the left of the line.

I can't think of a reason why this problem would only affect vertical labels. Would you mind double checking that this behaviour only affects vertical labels? If so, we should endeavour to fix.  

----

> "/" character collides with all characters above it

> This is a good argument for making #3505 a release blocker.

I'll make #3505 a release blocker. It should be easy to implement & help mitigate the problem.

lucaswoj:
I'm rather stuck on these remaining punctuation glitches. Any comments on the correctness / incorrectness on what's rendered or solutions @1ec5 @xrwang @nickidlugash @ansis?

screen shot 2016-11-09 at 10 39 17 am

Any bugs are most likely in this [code within `glyphs.js`](http://github.com/mapbox/mapbox-gl-js/blob/32413f311b2581d94698682f3c7537b068703ff7/js/symbol/quads.js#L180-L185) (this algo doesn't make much intuitive sense, this code has evolved based on emperical success) or this [code in `shaping.js`](http://github.com/mapbox/mapbox-gl-js/blob/32413f311b2581d94698682f3c7537b068703ff7/js/symbol/shaping.js#L65-L68).

1ec5:
> This should be an underscore

I think the underscore would be exceedingly rare in CJK line-placed labels.

> Should this be more to the right?

According to [this Wikipedia article](https://en.wikipedia.org/wiki/Chinese_punctuation#Punctuation_marks), the current horizontally centered behavior of `︒` is correct for Traditional Chinese, but ideally it would be offset to the right for Simplified Chinese.

Does `…` get centered vertically only when adjacent to CJK, or always? It shouldn’t be vertically centered for Latin text.

js/util/verticalize_punctuation.js
```diff
@@ -0,0 +1,60 @@
...
+  '｢': '﹁',
+  '｣': '﹂',
+  '［': '﹇',
+  '］': '﹈',
+  '“': '﹁',
+  '”': '﹂',
+  '‘': '﹃',
+  '’': '﹄',
+  '.': '・'
+};
```
1ec5: I think we should add all the remaining [fullwidth punctuation](https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms) so that characters like `？` and `！` don’t get rotated 90° when they’re sitting next to CJK. @nickidlugash, can you think of any exceptions?

nickidlugash: No, I can't think of any. They should probably all not get rotated. Are we using the fullwidth numerals and latin characters for single characters sitting next to CJK?

js/util/verticalize_punctuation.js
```diff
@@ -0,0 +1,59 @@
...
+    '』': '﹄',
+    '｢': '﹁',
+    '｣': '﹂',
+    '［': '﹇',
+    '］': '﹈',
+    '“': '﹁',
+    '”': '﹂',
+    '‘': '﹃',
+    '’': '﹄',
+    '.': '・'
```
lucaswoj: This comment was lost in the rebase cc @1ec5 @nickidlugash 

screen shot 2016-11-09 at 4 50 04 pm

lucaswoj:
> Vertical labels seem off-centered towards the left of the line. Will it be possible to have them be centered? I haven't had a chance to dig into the code to see where the issue is.

As best I can tell, this behaviour has not changed between `master` and this branch. It may have been fixed between when this was reported and now.

js/util/verticalize_punctuation.js
lucaswoj: @1ec5 I don't have a good sense of which punctuation should go into this table and which should not. Can you provide a table or algorithm? It may be easiest to just add a commit to the PR with your intended changes.

@nickidlugash We do not use fullwidth numerals and characters yet. Detecting when it is / isn't ok to use fullwidth characters is tricky. This can be done in a follow-up PR.

lucaswoj:
This PR is rebased, squashed, and 🍏. I hope we can merge this very soon! Maintaining a large & long-running feature branch is a huge amount of work.

js/util/verticalize_punctuation.js
1ec5: → https://github.com/mapbox/mapbox-gl-js/pull/3438#pullrequestreview-7928143

1ec5:
I’ve added all the punctuation marks that belong in this PR. In addition to the punctuation marks in #3505, I mapped various ASCII characters to [their fullwidth equivalents](https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms#In_Unicode). However, I mapped the ASCII brackets and their fullwidth equivalents to their vertical analogues in the [CJK Compatibility Forms](https://en.wikipedia.org/wiki/CJK_Compatibility_Forms), and I left out `"` and `'` because their vertical equivalents `＂` and `＇` are probably only appropriate for horizontal text.

Currently, we replace a punctuation character with its fullwidth analogue if both adjacent characters are compatible with vertical writing. This heuristic is conservative enough that we could implement _tate-chu-yoko_ (see #3506) simply by mapping all the ASCII numbers and letters to their fullwidth analogues in `lookup` (or by adding `0xfee0` to the character code of any ASCII number or letter). However, I think ultimately we’ll want to split `lookup` into three lookup tables (`leftBinding`, `leftOrRightBinding`, and `rightBinding`) and consider only the side(s) to which the character is bound.

Since both those improvements would require changes to the test suite, I’ll take that as a sign that we should go ahead and merge and follow up with more heuristics in a separate PR.

1ec5:
> Currently, we replace a punctuation character with its fullwidth analogue if both adjacent characters are compatible with vertical writing. This heuristic is conservative enough that we could implement _tate-chu-yoko_ (see #3506) simply by mapping all the ASCII numbers and letters to their fullwidth analogues in `lookup` (or by adding `0xfee0` to the character code of any ASCII number or letter).

→ #3588

> However, I think ultimately we’ll want to split `lookup` into three lookup tables (`leftBinding`, `leftOrRightBinding`, and `rightBinding`) and consider only the side(s) to which the character is bound.

→ #3587

PR #3440, opened 2016-10-21
Update to documentation 4.0.0-beta11

### LINKED ISSUE #3416: Fix broken JSDoc

jfirebaugh:
Fixes #3416.

Issues that still need resolution:

* [x] `Camera` methods are documented as being static members of `Map`:
 image
* [x] Various classes are documented to extend themselves:
 image
* [ ] Where it says "Extends Control", "Control" doesn't link correctly because the anchor has a lowercase "c".


mourner:
Awesome, thanks for taking care of this!

PR #3452, opened 2016-10-24
Add instructions to issue template requiring issue titles be a verb phrase

lucaswoj:
Apologies for jumping the gun by committing 3e7e422b498c9d01b566a0dac24eeff57cb89742 directly to `master`.

Requiring that all issue titles be a verb phrase means that

 - there is more standardization of issue title formatting
 - there is a higher bar for summarizing the body of an issue in its title
 - there is more likelihood that an issue has a clear "success criterion" defining when it can be closed
 - there is less likelihood of issues being opened that do not request a new feature or report a bug
 - there is more likelihood of an issue having a clear [next action](https://zenhabits.net/why-whats-the-next-action-is-the-most-important-question/)

----

@jfirebaugh raised a concern about this new standard in Slack:

 > In my opinion, this approach adds redundancy to label titles, and in some cases leads people to presuppose particular solutions or approaches, where what's needed is a good problem statement rather than a particular remedy.

I think that this is a valid but mostly orthogonal concern. An issue title may presuppose a particular solution whether it is a verb phrase or not. There are valid cases to open Issues that are specifications only, requirements only, and that blend the two. 

 ----

cc @jfirebaugh @mourner @tmcw @1ec5

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -2,6 +2,7 @@
 Hello! Thanks for contributing. 
 
 If you are reporting a bug, please:
+ - Make the issue an actionable verb phrase such as "Fix bug causing exception to thrown when rotating the map" (rather than "Map rotation is broken" or "Map rotation bug")
  - Include a link to a minimal demonstration of the bug. We recommend using https://jsbin.com
  - Ensure you can reproduce the bug using the latest release.
  - Check the console for relevant errors and warnings
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -2,7 +2,7 @@
 Hello! Thanks for contributing. 
 
 If you are reporting a bug, please:
- - Make the issue an actionable verb phrase such as "Fix bug causing exception to thrown when rotating the map" (rather than "Map rotation is broken" or "Map rotation bug")
+ - Make the issue title a succinct but specific description of the problem you encountered. Bad: "Map rotation is broken". Good: "map.setBearing(...) throws a TypeError for negative values"
  - Include a link to a minimal demonstration of the bug. We recommend using https://jsbin.com
  - Ensure you can reproduce the bug using the latest release.
  - Check the console for relevant errors and warnings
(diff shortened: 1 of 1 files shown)


tmcw:
I don't think we should set down a rule for issue titling. The main value of issue creation guides is "more information", and the rest is icing on the cake - since the main problem that issue creation guides solve is that typical issues don't provide nearly enough information. Retitling a reasonably complete issue is easy enough.

jfirebaugh:
I'd be 👍  for guidelines like:

> For _bug reports_, the issue title should be a succinct but _specific_ description of the problem you encountered. Bad: "Map rotation is broken". Good: "map.setBearing(...) throws a TypeError for negative values"
> For _feature requests_, the issue title should be a succinct but _specific_ description of your use case or the desired functionality. Bad: "More clustering options". Good: "Compute and expose aggregated values when clustering GeoJSON".

Making the key phrase "specific" rather than "actionable verb phrase"

* Reinforces the importance of specific information (agree with @tmcw here)
* Doesn't require every bug title to start with the filler words "Fix bug causing..."
* Doesn't require feature requests to come with a proposed implementation, which can [obscure the underlying use case](http://xyproblem.info/) and lead to assumptions that the original proposal is the correct one, or what we're likely to implement.

bhousel:
My 2¢ : If you want people to help with an open source project, your focus should be the relentless removal of friction.  Guidelines like these are unfriendly and drive people away / keep people out.

Sorry, but I feel strongly about this.  Work on things that matter.

tmcw:
That's an abrasive overstatement. The guidelines we already have in place haven't kept people out or driven them away: they've made tickets more productive, eliminating unnecessary back-and-forth. The guidelines we have in code, like tests and eslint, also add friction, but are worthwhile.

This new guideline might not be worth the added friction, but all guidelines add friction and many of them are worthwhile enough to merit it. This is not black & white, and is not categorically a "thing that doesn't matter."

1ec5:
As written, this PR would lead to bugs like “Fix broken map rotation” or “Fix map rotation bug”, leaving no one the wiser. If it’s important enough that a majority of issue titles begin with the word “Fix”, then the onus should be on the project maintainers to retitle issues.

Personally, I prefer that an bug report title “describe the unexpected behavior” (those words exactly) while a PR title describe the fix for that problem. If describing a bug doesn’t make a bug report easy enough to find, the project maintainer can append a technical hint. For example:

> Page freezes after easing to South Pole
> Page freezes after easing to South Pole — listen for onClimateChange events

lucaswoj:
:+1: on @jfirebaugh's standards, I'll update the PR now

Issue #2204, opened 2016-03-01
Disable drag rotate handler during touch events


DenisCarriere:
When holding down (touch) using a Mobile Device or even using Google Chrome inspect, the map seems to be very glitchy afterwards.

**mapbox-gl-js v0.14.3**:
### Steps to Trigger Behavior
1. Hold down map using Mobile device

The only way to prevent it from being glitchy was to disable `dragRotate`.

``` js
map.dragRotate.disable()
```

But that removes the really cool right click rotate functionality.

I tested this bug using a basic MapboxGL example.

![ycxfo5qykx](https://cloud.githubusercontent.com/assets/550895/13415877/86fe44b6-df2c-11e5-8350-dbc461b8b5dd.gif)

lucaswoj:
Thanks for the report @DenisCarriere! Could you tell me more about the glitch-iness? What is the exactly map doing? It is hard to tell from the posted gif. 

cc @bhousel

DenisCarriere:
Yea I know the GIF wasn't very clear, well just test it out using Google
chrome inspect and select mobile view. Hold down touching the map, after
that every time you click on the map it rotates the map at a random
location.

This bug is also found on Android devices viewing a Mapbox GL JS map.

DenisCarriere:
**Changes to Issue**:  The error is still the same, however it was the `map.dragRotate` that was causing the error and not `touchZoomRotate`.
cc @bhousel

DenisCarriere:
My quick fix was to implement a Mobile Detect library and turn off `map.dragRotate`, this doesn't change any of the functionality on the Browser and on mobile devices you have the Pinch gesture to rotate.

```javascript
import MobileDetect from 'mobile-detect'

const md = new MobileDetect(window.navigator.userAgent)

// Disable (Mobile)
if (md.mobile()) {
  map.dragRotate.disable()
}
```

lucaswoj:
Thanks for the debugging @DenisCarriere! @bhousel do you think we should automatically disable `dragRotate` for mobile devices? Are there any other follow-ups?

bhousel:
> @bhousel do you think we should automatically disable dragRotate for mobile devices? Are there any other follow-ups?

Probably a better idea would be to watch for `touchstart`/`touchend` and disable the handler if we detect a touch event happening.  The touch events fire before the mouse events.

Besides, some devices like new Windows Surface and Chromebooks support both mouse and touch, so the problem isn't just limited to mobile.

[...]

PR #3457, opened 2016-10-25
Disable drag rotate if touch detected

ChrisBellew:
A long press touch is interpreted by the browser as a right click,
which causes the drag rotate to start a rotation. Subsequent touches
perform the rotation, which causes the map to erratically rotate
around. If a touch start event is detected, just disable the feature.

As recommended by @bhousel in https://github.com/mapbox/mapbox-gl-js/issues/2204

### DIFF
--- a/js/ui/handler/drag_rotate.js
+++ b/js/ui/handler/drag_rotate.js
@@ -61,6 +61,7 @@ class DragRotateHandler {
     enable() {
         if (this.isEnabled()) return;
         this._el.addEventListener('mousedown', this._onDown);
+        this._el.addEventListener('touchstart', this._onTouchStart);
         this._enabled = true;
     }
 
@@ -73,7 +74,12 @@ class DragRotateHandler {
     disable() {
         if (!this.isEnabled()) return;
         this._el.removeEventListener('mousedown', this._onDown);
+        this._el.removeEventListener('touchstart', this._onTouchStart);
         this._enabled = false;
+    }
+
+    _onTouchStart() {
+        this.disable();
     }
 
     _onDown(e) {


js/ui/handler/drag_rotate.js
```diff
@@ -73,9 +74,14 @@ class DragRotateHandler {
     disable() {
         if (!this.isEnabled()) return;
         this._el.removeEventListener('mousedown', this._onDown);
+        this._el.removeEventListener('touchstart', this._onTouchStart);
         this._enabled = false;
     }
 
+    _onTouchStart() {
+        this.disable();
```
lucaswoj: Do we ever need to re-enable drag rotate after a touchstart event?

lucaswoj:
## Launch Checklist

ChrisBellew:
My opinion is no. There is no meaningful way to interact with it in its
current form, especially since it causes such erratic behavior after it is
activated. Tapping feels like selection/panning/movement, pinching like
zooming and long press for more information/options. Can't think of another
way besides having a separate menu?

On 31 Oct. 2016 9:12 pm, "Lucas Wojciechowski" <notifications@github.com>
wrote:

*@lucaswoj* commented on this pull request.
------------------------------

In js/ui/handler/drag_rotate.js
<https://github.com/mapbox/mapbox-gl-js/pull/3457#pullrequestreview-6432166>
:

>          this._enabled = false;
     }

+    _onTouchStart() {
+        this.disable();

Do we ever need to re-enable drag rotate after a touchstart event?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mapbox/mapbox-gl-js/pull/3457#pullrequestreview-6432166>,
or mute the thread

lucaswoj:
I see. We're using the existence of the `touchstart` event as an indication that we're on a touchscreen device and therefore disable any handlers that are mouse-specific. That makes sense to me 👍.

lucaswoj:
Some devices may have both touch events AND mouse events (i.e. the Chromebook Pixel). We can support those devices seamlessly if we enable the `DragRotateHandler` when a `touchend` event is encountered.

lucaswoj:
This may necessitate using a mechanism other than `enable` / `disable` so that we don't override user settings for this interaction handler. A private `_isTouchInteraction` property should work.

js/ui/handler/drag_rotate.js
mb12: Yes, on devices that have both touch screen and mouse (e.g Windows Surface Book and almost all modern Windows 10 laptops), drag rotate should be enabled when mouse interaction starts.

This can be tested on  desktop chrome using the following steps:

1. Use toggle device toolbar in desktop Chrome and switch to "Nexus 6P". Long press.
2. In the same debugging session, tap on toggle device toolbar to switch back to regular desktop chrome. Dragpan rotate should work (at least after first mouse click instead of being permanently disabled).

lucaswoj:
# Launch Checklist 

 - [ ] re-enable `DragRotateHandler` after a `touchend` event
 - [ ] ensure user calls to `DragRotateHandler#enable` / `#disable` do not interfere with this PR's behaviour

jfirebaugh:
Hey @ChrisBellew -- I'm in the process of cleaning out our PR queue and it looks like there hasn't been any activity on this one recently. If you're still interested in coming back to it, please let us know and we'll be happy to help you get it to completion. Until then, I'm going to close it out. Thanks again for your contributions!

PR #3458, opened 2016-10-25
Iterate array with for rather than for..in

anandthakker:
Reason: I just ran into a case where an ill-advised polyfill on a page I
don't fully control caused this loop to blow up, because the polyfill
attached enumerable properties to Array.prototype. (Ugh.)


anandthakker:
@mourner ^ switched to for..of

anandthakker:
(also, just to be paranoid, confirmed that buble transpiles it to plain for loop using index/length)

Issue #3447, opened 2016-10-24
symbol-features-in pitched-screen query test is failing


jfirebaugh:
https://github.com/mapbox/mapbox-gl-test-suite/tree/master/query-tests/symbol-features-in/pitched-screen

cc @brunoabinader

brunoabinader:
It turns out the generated expectations from -native are wrong. I've updated the [test suite](https://github.com/mapbox/mapbox-gl-test-suite/commit/2817eedfe76f38b63bb77cb58c3b691e47169a92) (https://github.com/mapbox/mapbox-gl-js/commit/baa2b8dc092d5c5fb35d954fbe83f4970ef72ac7) to reflect this.

Related -native issue: https://github.com/mapbox/mapbox-gl-native/issues/6817

PR #3459, opened 2016-10-25
Fix queryRenderedFeatures for symbol features

brunoabinader:
Use the query geometry as polygon for intersecting with symbol features collision box. Fixes situations where e.g. unseen features were returned or vice-versa.

This PR is a port from GL native (see https://github.com/mapbox/mapbox-gl-native/pull/6773). 

Query test `symbol-features-in/pitched-screen` (#3447) should be passing now, but because the feature order differs from GL native, the test fails. We should probably sort features by latitude/longitude prior to JSON comparison.

Benchmark reports are within standard deviation:
screen shot 2016-10-25 at 4 51 33 pm

/cc @ansis @jfirebaugh @mourner

### DIFF
--- a/js/symbol/collision_tile.js
+++ b/js/symbol/collision_tile.js
@@ -4,6 +4,8 @@ const Point = require('point-geometry');
 const EXTENT = require('../data/extent');
 const Grid = require('grid-index');
 
+const intersectionTests = require('../util/intersection_tests');
+
 /**
  * A collision tile used to prevent symbols from overlapping. It keep tracks of
  * where previous symbols have been placed and is used to check if a new
@@ -25,7 +27,7 @@ class CollisionTile {
             this.ignoredGrid = new Grid(EXTENT, 12, 0);
         }
 
-        this.minScale = 0.25;
+        this.minScale = 0.5;
         this.maxScale = 2;
 
         this.angle = angle;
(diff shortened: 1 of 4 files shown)


brunoabinader:
Possible fix for #2647.

js/symbol/collision_tile.js
```diff
@@ -173,56 +175,89 @@ class CollisionTile {
...
 
+        // Predicate for ruling out already seen features.
+        function seenFeature(sourceLayerFeatures, blocking) {
             const sourceLayer = blocking.sourceLayerIndex;
             const featureIndex = blocking.featureIndex;
             if (sourceLayerFeatures[sourceLayer] === undefined) {
                 sourceLayerFeatures[sourceLayer] = {};
             }
+            return sourceLayerFeatures[sourceLayer][featureIndex] === true;
+        }
```
mourner: Since this function does not need any variables from the outer scope (it gets everything it needs from arguments), move it out to the top level scope. This way you avoid the cost of setting up a closure on every call of `queryRenderedSymbols`.

js/symbol/collision_tile.js
```diff
@@ -173,56 +175,89 @@ class CollisionTile {
...
-                    result.push(blockingBoxKeys[i]);
-                }
-            }
+        // Account for the rounding done when updating symbol shader variables.
+        const roundedScale = Math.pow(2, Math.ceil(Math.log(scale) / Math.LN2 * 10) / 10);
+
+        // Check if feature is rendered (collision free) at current scale.
+        function visibleAtScale(blocking, roundedScale) {
+            return roundedScale >= blocking.placementScale && roundedScale <= blocking.maxScale;
+        }
```
mourner: Same for this function.

js/symbol/collision_tile.js
```diff
@@ -173,56 +175,89 @@ class CollisionTile {
...
+            const x2 = anchor.x + blocking.x2 / scale;
+            const y2 = anchor.y + blocking.y2 / scale * yStretch;
+            const bbox = [
+                new Point(x1, y1),
+                new Point(x2, y1),
+                new Point(x2, y2),
+                new Point(x1, y2)
+            ];
+            return intersectionTests.polygonIntersectsPolygon(rotatedQuery, bbox);
+        }
```
mourner: And this one too, if you pass `bbox` as an argument.

js/symbol/intersection_tests.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
```
jfirebaugh: Can these be consolidated with [`util/intersection_tests.js`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/util/intersection_tests.js)?

brunoabinader: Oh my - totally missed this - yes we can certainly reuse it 👍

brunoabinader:
Thank you @mourner and @jfirebaugh for the review - I've addressed the requested changes.

brunoabinader:
@mourner it turns out those functions made more sense in GL native context because we were using them as predicates for `boost::geometry::index::query`. I guess for GL JS we can just get rid of the extra overhead and drop its contents directly into the main function.

PR #3461, opened 2016-10-25
Switch from physical to logical segmentation of buffers

### LINKED ISSUE #207: Merge multiple vertex or elements buffers
Currently, we create separate vertex/elements buffers when the vertex buffer reaches an item count of 65536. However, we could just continue in the same buffer and just reset the vertexAttribPointer to start indexing the buffer at a later position. That way, we'd have fewer buffer switches and fewer buffer transfer operations overall.

### LINKED ISSUE #2038: Switch collision boxes to use index rendering
Collision debug rendering uses `drawArrays` while everything else uses `drawElements`. This leads to a couple weird exceptions:
https://github.com/mapbox/mapbox-gl-js/blob/5296c94469b161296736a0216268c6d6d6c2b9b2/js/data/symbol_bucket.js#L501
https://github.com/mapbox/mapbox-gl-js/blob/5296c94469b161296736a0216268c6d6d6c2b9b2/js/render/draw_collision_debug.js#L18

The only reason to switch is consistency

jfirebaugh:
Fixes #207. 

And while here, switch to indexed rendering for collision boxes. Fixes #2038.

Physical segmentation: each vertex buffer holds a maximum of 2¹⁶-1 vertices. When a layer requires more than that, create a new vertex buffer and index buffer.

Logical segmentation: no limit to the size of a vertex buffer. When reaching 2¹⁶-1 vertices, record the vertex and primitive offsets, then reset the indexing, starting a new "segment". Use the offsets when rendering each segment.

Logical segmentation has the following advantages:

* It uses a smaller number of larger buffer objects. In theory this is more efficient for WebGL/OpenGL. However in this case the performance difference is likely negligible. We already use VAOs, which largely eliminate the cost of switching between buffers, which leaves the cost of creating them. And buffers larger than 2¹⁶ vertices are relatively rare.
* It's what gl-native does. Having the two codebases work the same way is an advantage, as we frequently need to port changes back and forth.
* It simplifies `populatePaintArrays`. With physical segmentation, `populatePaintArrays` needs to track both a buffer index and an index into the first buffer, and might need to populate multiple buffer segments. With logical segmentation, it just fills the single paint vertex buffer until it's the same length as the layout vertex buffer.

benchmark | master e03a7b1 | logical-buffer-segmentation 59f39b7
--- | --- | ---
**map-load** | 150 ms  | 104 ms 
**style-load** | 100 ms  | 104 ms 
**buffer** | 1,030 ms  | 993 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 7.3 ms, 1% > 16ms  | 7.3 ms, 1% > 16ms 
**query-point** | 1.05 ms  | 1.16 ms 
**query-box** | 84.95 ms  | 88.68 ms 
**geojson-setdata-small** | 8 ms  | 10 ms 
**geojson-setdata-large** | 117 ms  | 99 ms 

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/js/data/bucket/fill_bucket.js
+++ b/js/data/bucket/fill_bucket.js
@@ -6,6 +6,7 @@ const ElementArrayType = require('../element_array_type');
 const loadGeometry = require('../load_geometry');
 const earcut = require('earcut');
 const classifyRings = require('../../util/classify_rings');
+const assert = require('assert');
 const EARCUT_MAX_RINGS = 500;
 
 const fillInterfaces = {
@@ -15,7 +16,7 @@ const fillInterfaces = {
             components: 2,
             type: 'Int16'
         }]),
-        elementArrayType: new ElementArrayType(1),
+        elementArrayType: new ElementArrayType(3),
         elementArrayType2: new ElementArrayType(2),
 
         paintAttributes: [{
@@ -55,8 +56,7 @@ class FillBucket extends Bucket {
     }
 
     addFeature(feature) {
-        const startGroup = this.prepareArrayGroup('fill', 0);
-        const startIndex = startGroup.layoutVertexArray.length;
+        const arrays = this.arrays.fill;
 
         for (const polygon of classifyRings(loadGeometry(feature), EARCUT_MAX_RINGS)) {
             let numVertices = 0;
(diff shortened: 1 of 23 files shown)


js/data/bucket/fill_bucket.js
```diff
@@ -15,7 +16,7 @@ const fillInterfaces = {
             components: 2,
             type: 'Int16'
         }]),
-        elementArrayType: new ElementArrayType(1),
+        elementArrayType: new ElementArrayType(3),
```
lucaswoj: `new ElementArrayType(3)` looks correct to me. I wonder why this was `1` before.

jfirebaugh: It worked with `1` before because `FillBucket#addFeature` [emplaced one index at a time](https://github.com/mapbox/mapbox-gl-js/pull/3461/files#diff-3d556b8c468e9f1285f261f235775d71L96). Now it emplaces triangles, for consistency with `FillExtrudeBucket` (and native).

PR #3462, opened 2016-10-25
Fix link in map event docs

DannyDelott:


jfirebaugh:
Thanks! Applied as 685c41f10cc9740cfd2a4dc238d76a1232d9f08a.

PR #3465, opened 2016-10-26
Simplify Bucket

jfirebaugh:
Some relatively straightforward API simplifications for `Bucket`:

* Combine `getTransferables` and `serialize` into a single method.
* Eliminate explicit array trimming -- this [happens automatically](https://github.com/mapbox/mapbox-gl-js/blob/b9d95095c63a8e0193b6ee391ab308e75e1ecea3/js/util/struct_array.js#L86) as part of serialization.

benchmark | master b9d9509 | simplify-bucket b75bbf4
--- | --- | ---
**map-load** | 164 ms  | 96 ms 
**style-load** | 106 ms  | 102 ms 
**buffer** | 1,058 ms  | 1,080 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 7 ms, 0% > 16ms  | 7.1 ms, 0% > 16ms 
**query-point** | 1.06 ms  | 1.41 ms 
**query-box** | 68.11 ms  | 72.10 ms 
**geojson-setdata-small** | 9 ms  | 8 ms 
**geojson-setdata-large** | 304 ms  | 286 ms 

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -92,17 +92,11 @@ class Bucket {
         return true;
     }
 
-    getTransferables(transferables) {
-        for (const programName in this.arrays) {
-            this.arrays[programName].getTransferables(transferables);
-        }
-    }
-
-    serialize() {
+    serialize(transferables) {
         return {
             zoom: this.zoom,
             layerIds: this.layers.map((l) => l.id),
-            arrays: util.mapObject(this.arrays, (a) => a.serialize())
+            arrays: util.mapObject(this.arrays, (a) => a.serialize(transferables))
         };
     }
 
--- a/js/data/bucket.js
+++ b/js/data/bucket.js
@@ -58,8 +58,6 @@ class Bucket {
                 options.featureIndex.insert(feature, this.index);
             }
         }
-
-        this.trimArrays();
     }
 
     createArrays() {
@@ -74,12 +72,6 @@ class Bucket {
     destroy() {
         for (const programName in this.bufferGroups) {
             this.bufferGroups[programName].destroy();
-        }
-    }
-
-    trimArrays() {
-        for (const programName in this.arrays) {
-            this.arrays[programName].trim();
         }
     }
 
--- a/js/source/worker_tile.js
+++ b/js/source/worker_tile.js
@@ -94,7 +94,7 @@ class WorkerTile {
                 });
 
                 bucket.populate(features, options);
-                featureIndex.bucketLayerIDs[bucket.index] = family.map(getLayerId);
+                featureIndex.bucketLayerIDs[bucket.index] = family.map((l) => l.id);
             }
         }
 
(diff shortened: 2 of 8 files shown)


js/data/bucket.js
```diff
@@ -58,8 +58,6 @@ class Bucket {
                 options.featureIndex.insert(feature, this.index);
```
mb12: @jfirebaugh Can you please clarify how this works correctly for symbols? Should featureIndex only have geometries that are actually rendered on the map? Would this (or not) add all point geometries as long as they match the filter?

js/source/worker_tile.js
```diff
@@ -94,7 +94,7 @@ class WorkerTile {
                 });
 
                 bucket.populate(features, options);
-                featureIndex.bucketLayerIDs[bucket.index] = family.map(getLayerId);
+                featureIndex.bucketLayerIDs[bucket.index] = family.map((l) => l.id);
```
mourner: This is inside a double-loop and the closure here doesn't need any of the outer scope — I'm not sure V8 optimizes cases like this well enough, so I'd keep it a top-level function.

js/source/worker_tile.js
```diff
@@ -106,23 +106,17 @@ class WorkerTile {
...
-                collisionTile: collisionTile_.data,
-                collisionBoxArray: collisionBoxArray,
-                symbolInstancesArray: symbolInstancesArray,
-                symbolQuadsArray: symbolQuadsArray
-            }, getTransferables(nonEmptyBuckets)
-                .concat(featureIndex_.transferables)
-                .concat(collisionTile_.transferables));
+                buckets: util.values(buckets)
+                    .filter((b) => !b.isEmpty())
+                    .map((b) => b.serialize(transferables)),
```
mourner: Same here.

js/data/bucket.js
jfirebaugh: This code isn't used for symbols. `SymbolBucket` has [it's own definition of `populate`](https://github.com/mapbox/mapbox-gl-js/blob/b9d95095c63a8e0193b6ee391ab308e75e1ecea3/js/data/bucket/symbol_bucket.js#L138).

js/source/worker_tile.js
jfirebaugh: The benchmarks don't show any meaningful performance difference.

### Written as is

benchmark | master b9d9509 | simplify-bucket b75bbf4
--- | --- | ---
**buffer** | 919 ms  | 914 ms 

### Variant 1

```
function serializeBuckets(buckets, transferables) {
    return buckets
        .filter((b) => !b.isEmpty())
        .map((b) => b.serialize(transferables));
}
```

benchmark | master b9d9509 | simplify-bucket b75bbf4
--- | --- | ---
**buffer** | 906 ms  | 911 ms 


### Variant 2

```
function bucketNonEmpty(bucket) {
    return !bucket.isEmpty();
}

function serializeBuckets(buckets, transferables) {
    return buckets
        .filter(bucketNonEmpty)
        .map((b) => b.serialize(transferables));
}
```

benchmark | master b9d9509 | simplify-bucket b75bbf4
--- | --- | ---
**buffer** | 893 ms  | 900 ms 

You can see that all the alternatives are within the range of the control (master b9d9509) -- ~890 - 920ms.

Variant 1 does avoid repeating the `filter`/`map` pipeline, so I'm inclined to go with it.

jfirebaugh: Here's replacing `(l) => l.id` with top level `getLayerID` (also no measurable difference):

benchmark | master b9d9509 | simplify-bucket b75bbf4
--- | --- | ---
**buffer** | 887 ms  | 908 ms

js/data/bucket.js
mb12: @jfirebaugh Is it possible to point out where the insertion into feature index happens for symbols in SymbolBucket and/or how do geometries for symbol features make into the index used for queryRenderedFeatures? 

The reason I am asking you this is because in an older version of GL JS, this insertion used to happen in WorkerTile and it was explicitly disabled for symbols. The only way to work around it and get symbols returned by queryRenderedFeatures was to comment out this condition.

jfirebaugh: Feature querying for symbols does not use `FeatureIndex`. Instead it uses the `CollisionTile` generated during placement. If you're seeing issues with symbol querying, please file a separate issue.

js/source/worker_tile.js
```diff
@@ -106,23 +106,15 @@ class WorkerTile {
...
-                symbolQuadsArray: symbolQuadsArray
-            }, getTransferables(nonEmptyBuckets)
-                .concat(featureIndex_.transferables)
-                .concat(collisionTile_.transferables));
+                buckets: serializeBuckets(util.values(buckets), transferables),
+                featureIndex: featureIndex.serialize(transferables),
+                collisionTile: collisionTile.serialize(transferables),
+                collisionBoxArray: this.collisionBoxArray.serialize(),
+                symbolInstancesArray: this.symbolInstancesArray.serialize(),
+                symbolQuadsArray: this.symbolQuadsArray.serialize()
```
mourner: Quick question, why are the latter three not put into transferables? Are they already there indirectly through the symbol buckets?

jfirebaugh: I believe it's an oversight from #2624.

PR #3468, opened 2016-10-26
Fix setPaintProperty to reparse buckets transitioning between flat and extruded fills

lbud:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/3386 by taking advantage of the worker roundtrip in `setPaintProperty` to transition between flat and extruded fill buckets.

cc @lucaswoj @jfirebaugh

### DIFF
--- a/js/render/painter.js
+++ b/js/render/painter.js
@@ -272,10 +272,7 @@ class Painter {
         this.id = layer.id;
 
         let type = layer.type;
-        if (type === 'fill' &&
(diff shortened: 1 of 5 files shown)


jfirebaugh:
Can you create a test-suite regression test for this as well, so we can ensure that the bug doesn't repeat itself in gl-native?

lbud:
Added a test + rebased. @jfirebaugh 🚦?

PR #3476, opened 2016-10-27
Always build dist files before publishing

### LINKED ISSUE #3475: NPM registry yields prior string version of latest mapbox-gl dist
**mapbox-gl-js version:** `0.26.0` (??? see below ???)

Latest build under [https://registry.npmjs.org/mapbox-gl/-/mapbox-gl-0.26.0.tgz](https://registry.npmjs.org/mapbox-gl/-/mapbox-gl-0.26.0.tgz) at `dist/mapbox-gl.js` reports the incorrect version.

### Expected Behavior
exports string version: `0.26.0`

### Actual Behavior
exports string version: `0.25.1`

screen shot 2016-10-27 at 10 30 05 am

mourner:
Closes #3475. @jfirebaugh

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
     "test-unit": "tap --reporter dot --no-coverage test/js",
     "test-render": "node test/render.test.js",
     "test-query": "node test/query.test.js",
-    "test-cov": "nyc --reporter=text-summary --cache run-s test-unit test-render test-query"
+    "test-cov": "nyc --reporter=text-summary --cache run-s test-unit test-render test-query",
+    "prepublish": "run-s build-dev build-min"
   }
 }


jfirebaugh:
`prepublish` is deprecated -- https://github.com/npm/npm/releases/tag/v4.0.0

mourner:
@jfirebaugh it's only deprecated in NPM4+. Since the default is currently NPM3, we can't use the new `prepublishOnly` script.

mourner:
Yeah, let's remove after we switch to publishing releases on Circle.

jfirebaugh:
I always have skepticism for build system changes. In this case, see #2688.

Does this not regress `npm install` build time again?

mourner:
@jfirebaugh fair point, perhaps we should also use something like https://github.com/iarna/in-publish?

PR #3478, opened 2016-10-28
Encapsulate SymbolBucket's unique requirements

jfirebaugh:
`SymbolBucket` is in several ways the "odd bucket out":

* It's the only bucket type that works with multiple program interfaces -- one for glyphs, one for icons, and one for collision boxes.
* It's the only bucket type whose instances are reused to build buffers multiple times.

Previously, the `Bucket` base class was generalized such that it could support multiple program interfaces for any `Bucket` subtype, and re-create buffers at any time. However, this added complexity to all subtypes other than Symbol.

Now, `SymbolBucket` encapsulates its differences:

* It uses `ArrayGroup`/`BufferGroup` directly, creating one for each of its three interfaces. In turn, the other bucket types, their corresponding layer rendering functions, and the `Bucket` base class, can all assume a single program interface.
* It recreates these arrays when needed, while the `Bucket` base class creates them once, in the constructor.

### DIFF
--- a/js/data/array_group.js
+++ b/js/data/array_group.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const util = require('../util/util');
+const ProgramConfiguration = require('./program_configuration');
 
 class Segment {
     constructor(vertexOffset, primitiveOffset) {
@@ -31,7 +32,9 @@ class Segment {
  * @private
  */
 class ArrayGroup {
-    constructor(programInterface, programConfigurations) {
+    constructor(programInterface, layers, zoom) {
+        this.globalProperties = {zoom};
+
         const LayoutVertexArrayType = programInterface.layoutVertexArrayType;
         this.layoutVertexArray = new LayoutVertexArrayType();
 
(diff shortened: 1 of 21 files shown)


jfirebaugh:
This is crashing the buffer benchmarks for some reason.

jfirebaugh:
I made a 🐛 . Fixed.

benchmark | master 965a2f3 | symbol-bucket f74e58c
--- | --- | ---
**map-load** | 144 ms  | 92 ms 
**style-load** | 118 ms  | 117 ms 
**buffer** | 1,053 ms  | 1,078 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 7 ms, 0% > 16ms  | 6.9 ms, 0% > 16ms 
**query-point** | 1.10 ms  | 1.14 ms 
**query-box** | 66.95 ms  | 67.50 ms 
**geojson-setdata-small** | 9 ms  | 8 ms 
**geojson-setdata-large** | 275 ms  | 278 ms

Issue #3342, opened 2016-10-11
Address rendering performance regressions after DDS


mourner:
**mapbox-gl-js version**: v0.23+

Digging into rendering performance, I discovered that #2938 regressed it more severely than was demonstrated by the benchmarks, which averages frame duration with too many "easy" zooms. 

With a simple isolated benchmark that renders a 1440x900 map of Mapbox Streets on z13 (the most detailed one) in DC, CPU duration of a frame is almost 2 times bigger in v0.23 compared to v0.22.

- v0.22: https://jsfiddle.net/Mourner/om3hv07o/2/, median frame duration: 9.33
- v0.23: https://jsfiddle.net/Mourner/2bzyLLp5/4/, median frame duration: 17.72

This leads to a huge increase in dropped frames when panning and zooming, especially for full-screen maps of cities. The effect should also be much more pronounced on weaker machines (I'm testing on my MBP).

Other DDS PRs regress performance too, but to a lesser extent because CPU performance is line-dominated according to the Chrome profiler.

I think that we need to prioritize addressing this technical debt now, before it grows out of control with more DDS properties support.

There are several things that seem to be contributing to the regression:

- `bucket.setUniforms` does a full recalculation of DDS-enabled properties (`getPaintValue`), even though they don't differ from values in `layer.paint` (that are recalculated on every frame too). Replacing `uniform.getValue(...)` with `layer.paint[uniform.paintProperty]` generally works, with a few glitches to fix (fallback values like `fill-outline-color` falling back to `fill-color` are not calculated during the main style recalculation phase).
- The big block of uniform sets and `useProgram` in `draw_line` that come before `painter.enableTileClippingMask(coord);` now execute individually for each tile instead of once for a layer, although in non-DDS case the program options are the same (and maybe in case of DDS as well — I'm not sure). We need to refactor this code so that it is back to being executed once per-layer.

@lucaswoj I'll need your help to tackle this as I don't fully understand all the quirks of DDS code/vision yet, but I'm sure we can go back to v0.22 performance for Mapbox styles while keeping the DDS features.
cc @jfirebaugh @mollymerp

jfirebaugh:
On the first point, it sounds like we need to bite off #3044 and refactor so that we have a single canonical representation of paint values that can be updated once per frame (or less if possible) and then shared by `bucket.setUniforms` and others.

mourner:
[...]

PR #3485, opened 2016-10-28
Optimize line drawing by avoiding per-tile work

mourner:
This is a slightly messy attempt at winning back some of the performance that regressed with #3342 back, until we solve the challenge of ensuring all tiles are always rendered with the same style.

The idea is to simply check whether the program changed after `useProgram` call, and not do a big portion of the drawing work if it didn't.

On my [3342 benchmark](http://jsfiddle.net/Mourner/2bzyLLp5/4/), results are as follows:

- v0.19: 8.3ms
- v0.25: 17.2ms
- master: 15.3ms (looks like some of your recent changes saved us some ms @jfirebaugh!)
- this PR: ~~12.2ms~~ 11.7ms

👀 @lucaswoj @jfirebaugh

### DIFF
--- a/js/render/draw_line.js
+++ b/js/render/draw_line.js
@@ -24,72 +24,84 @@ module.exports = function drawLine(painter, sourceCache, layer, coords) {
     // don't draw zero-width lines
     if (layer.paint['line-width'] <= 0) return;
 
+    const programId =
+        layer.paint['line-dasharray'] ? 'lineSDF' :
+        layer.paint['line-pattern'] ? 'linePattern' : 'line';
+
     for (let k = 0; k < coords.length; k++) {
-        drawLineTile(painter, sourceCache, layer, coords[k]);
+        const tile = sourceCache.getTile(coords[k]);
+        const bucket = tile.getBucket(layer);
+        if (!bucket) continue;
+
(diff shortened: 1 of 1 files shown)


jfirebaugh:
I'd prefer to follow the optimization that native uses: each program object records the value which was last set for each uniform, and avoids calling `glUniform*` when the value isn't changing. This optimization benefits every layer type as well as the `k === 0` case here (you can reuse the uniform values set by a previous layer, if they're not changing).

mourner:
@jfirebaugh yes, but that sounds like a more involved refactor (#145). Meanwhile this PR is an easy short-term stopgap that we can revert at any time.

mourner:
@jfirebaugh added another commit that removes redundant state changes for pattern/dash-related line drawing code by tracking if the tile zoom changed between tiles. Also cleaned it up a bit. This brings the benchmark result a little further down, from 12.2 to 11.7.

mourner:
`frame-duration` benchmark result isn't as significant because of the averaging I mentioned in #3342, but still noticeable:

benchmark | master | optimize-line-draw
--- | --- | ---
**frame-duration** | 9.6 ms, 5% > 16ms  | 8.9 ms, 1% > 16ms

PR #3488, opened 2016-10-28
Update extrusion examples to use separate fill-extrusion type

lbud:
Update extrusion-based examples (once https://github.com/mapbox/mapbox-gl-js/pull/3487 is merged). Refs https://github.com/mapbox/mapbox-gl-style-spec/issues/554.

### DIFF
--- a/docs/_posts/examples/3400-01-27-3d-extrusion-floorplan.html
+++ b/docs/_posts/examples/3400-01-27-3d-extrusion-floorplan.html
@@ -1,9 +1,9 @@
---- 
+---
 layout: example
 category: example
 title: Extrude polygons for 3D indoor mapping.
-description: Create a 3D indoor map with the fill-extrude-height paint property.
-tags: 
+description: Create a 3D indoor map with the fill-extrusion type.
+tags:
   - layers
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Can you retarget this at `master`, so that it goes live simultaneously with 0.27.0 being released?

lbud:
@jfirebaugh I'm not sure how the relationship between `master` and `mb-pages` works — I targeted here based on [release documentation I was going through the other day](https://github.com/mapbox/gl-internal/wiki/Javascript-release-process#merge-mb-pages-into-master). Should I still retarget at `master`?

jfirebaugh:
There's more explanation [here](https://github.com/mapbox/mapbox-gl-js/blob/master/docs/README.md#committing-and-publishing-documentation) -- this case is admittedly borderline, since we're making a breaking change to an existing property. But I think it's better overall if we don't have a period of time where the published example won't work with the current released version, which would happen if we pushed to `mb-pages` before releasing 0.27.0.

lbud:
@jfirebaugh makes sense, thanks 🙏

PR #3489, opened 2016-10-30
Program configuration optimizations

### LINKED ISSUE #3342: Address rendering performance regressions after DDS

mourner:
Along with #3485, this PR brings the rendering performance with the Streets z13 benchmark to pre-DDS levels and closes #3342! 🎉🎉🎉 [Benchmark](http://jsfiddle.net/Mourner/2bzyLLp5/4/) results:

- v0.22: 8.8ms
- v0.25: 17.2ms
- master before #3485: 15.3ms
- master: 11.7ms
- after this PR: 8.3ms

Changes:

- Avoid repeatedly restringifying program configurations for caching.
- Avoid repeatedly recreating the same fill program for clipping masks and background.

Note that the failing build is due to changes in ESLint 3.9.0; fixed up the warnings in the third commit.

👀 @jfirebaugh @lucaswoj @ansis

### DIFF
--- a/js/data/program_configuration.js
+++ b/js/data/program_configuration.js
@@ -106,6 +106,8 @@ class ProgramConfiguration {
             }
         }
 
+        self.cacheKey = JSON.stringify([self.vertexPragmas, self.fragmentPragmas]);
+
         return self;
     }
 
@@ -126,6 +128,8 @@ class ProgramConfiguration {
             pragmas.define[uniform.name.slice(2)] = `uniform ${type} ${uniform.name};\n`;
             pragmas.initialize[uniform.name.slice(2)] = `${type} ${uniform.name.slice(2)} = ${uniform.name};\n`;
         }
+
+        self.cacheKey = JSON.stringify(pragmas);
 
         return self;
     }
@@ -151,15 +155,6 @@ class ProgramConfiguration {
 
     paintVertexArrayType() {
         return new VertexArrayType(this.attributes);
-    }
-
-    programCacheKey(name, showOverdraw) {
-        return JSON.stringify({
-            name: name,
-            vertexPragmas: this.vertexPragmas,
-            fragmentPragmas: this.fragmentPragmas,
-            overdraw: showOverdraw
-        });
     }
 
     createProgram(name, showOverdraw, gl) {
--- a/js/render/draw_background.js
+++ b/js/render/draw_background.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const pixelsToTileUnits = require('../source/pixels_to_tile_units');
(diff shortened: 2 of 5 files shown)


mourner:
@jfirebaugh oops, nope, the two render tests for switching paint from default to property function are failing. Investigating.

mourner:
OK fixed the bug and rebased, merging on green.

mourner:
For the record, `frame-duration` benchmark results:

**before both PRs**: 9.6 ms, 5% > 16ms
**after**: 7.4 ms, 0% > 16ms

mourner:
Note that all these optimizations became much more obvious after the ProgramConfiguration #3439 refactor. It made understanding how programs are created and used **much** easier.

PR #3492, opened 2016-10-31
use aria-label on control buttons

andrewharvey:
## Launch Checklist

As far as I can tell it doesn't seem like the navigation controls (zoom, rotate, geolocate) are accessible. As per https://www.w3.org/TR/wai-aria/states_and_properties#aria-label and general advice at http://getbootstrap.com/components/#glyphicons-how-to-use I believe by adding the `aria-label` attribute to these buttons screen readers etc will be able to provide users more information about what these buttons do.

Not sure this is really applicable here?

No public API changes.

 - [x] post benchmark scores

Shouldn't affect performance.

Manually tested.


lucaswoj:
Thanks @andrewharvey

PR #3493, opened 2016-10-31
Only run prepublish script on actual npm publish

mourner:
Makes `prepublish` script run only when executed in context of `npm publish`, addressing previous concerns with #3476. We'll get rid of it after setting up publishing from Circle. @jfirebaugh @lucaswoj

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "github-slugger": "^1.1.1",
     "gl": "^4.0.1",
     "highlight.js": "9.3.0",
+    "in-publish": "^2.0.0",
     "jsdom": "^9.4.2",
     "lodash.template": "^4.4.0",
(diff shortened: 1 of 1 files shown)


andrewharvey:
cc @jfirebaugh @mourner @lucaswoj 

I'm using `yarn add https://github.com/mapbox/mapbox-gl-js.git#master` in my project to use the master branch, however this PR applied the `in-published` check in `prepublish`. That means that when mapbox-gl get's installed automatically when I yarn or npm install my project it doesn't build the mapbox-gl dist files, but the dist files are needed for both the stylesheet and the main defined in the package.json, so the package doesn't work.

Does this still make sense now that the dist files are used as the entry point (done in #4423)?

Issue #1835, opened 2015-12-14
Document GL JS plugin architecture


tmcw:
Right now we have a bunch of competing approaches for how mapbox-gl-js plugins express their dependency on GL-JS.
- peerDependencies
- expecting a global to exist or window.mapboxgl
- dependency with a require

Some plugins require-into mapbox-gl for utilities and thus actually rely on gl js internals.

Core questions
- when should dependencies get a reference to gl-js?
- how do we write a system that works equally well for cdn dependencies and browserify and webpack?

cc @jfirebaugh @tristen @mourner

tristen:
cc @mcwhittemore as he's looking at making `mapboxgl.Control` it's own module.

mcwhittemore:
> how do we write a system that works equally well for cdn dependencies and browserify and webpack?

Thats a hard one. The first idea that comes to mind is having mapbox-gl pass in a ref to itself via the `addControl` function. This would let us say that all plugins shouldn't require mapbox-gl. This introduces some problems with confirming that the plugin works with the current version of mapbox-gl.

While I like the idea above, what I've been doing is moving to peerDependencies as this doesn't take a change to mapbox-gl and lowers the bundle size by removing extra refs to mapbox-gl.

mourner:
I'd say let's do this:

1. Specify Mapbox GL as `peerDependencies`.
2. Rely `window.mapboxgl` to exist, and warn about including after `mapboxgl` if it's not available. Not a perfect solution but it worked OK for Leaflet plugins and will work for both Browserify systems and CDN.

@mcwhittemore what's the reason for making `mapboxgl.Control` a module? Also note that not all plugins are supposed to be controls. Some are new sources, some are additional methods, etc.

mcwhittemore:
@mourner The idea with making `mapbox.Control` a module is that it could be required in with little bloat. That said, we went with the peerDependencies route over the making `mapbox.Control` a module when cleaning up `mapbox-gl-draw` and I think it worked well.

jfirebaugh:
> Some plugins require-into mapbox-gl for utilities and thus actually rely on gl js internals.

They shouldn't do this. `require('mapbox-gl')` is the public interface.

> how do we write a system that works equally well for cdn dependencies and browserify and webpack?

Ideally, only CDN dependencies rely on a `mapboxgl` global, and browserify/webpack don't. I'm not aware of any turnkey solutions to this. It seems like it requires an approach like:

[...]

PR #3497, opened 2016-10-31
Make it obvious how to create Control plugins that do not reference mapboxgl

tmcw:
## Launch Checklist

 - [ ] post benchmark scores

This implements part of #1835

### CHANGES

- Controls no longer take a `position` option. Instead, you provide a `position` argument to the `addControl` map method.
- Controls _must_ implement `onAdd` and `onRemove`.
- mapboxgl.Control: deleted
- Controls need to implement addTo -> controls need to implement onAdd
- attribution & scale controls are no longer evented

## Bugs fixed

- https://github.com/mapbox/mapbox-gl-js/issues/3561 no longer calls AttributionControl with `true` as an options argument
- https://github.com/mapbox/mapbox-gl-js/issues/3562 detaches event listeners from removed controls

cc @lucaswoj for the review


js/ui/control/attribution_control.js
```diff
@@ -12,7 +12,7 @@ const DOM = require('../../util/dom');
  * var map = new mapboxgl.Map({attributionControl: false})
  *     .addControl(new mapboxgl.AttributionControl({position: 'top-left'}));
  */
-class AttributionControl extends Control {
+class AttributionControl extends Evented {
```
mourner: `AttributionControl` doesn't need to be `Evented`.

js/ui/map.js
```diff
@@ -212,14 +212,37 @@ class Map extends Camera {
...
+     * Adds a [`Control`](#Control) to the map, calling `control.onAdd(this)`.
      *
      * @param {Control} control The [`Control`](#Control) to add.
      * @returns {Map} `this`
      * @see [Display map navigation controls](https://www.mapbox.com/mapbox-gl-js/example/navigation/)
      */
     addControl(control) {
-        control.addTo(this);
+        control._map = this;
+        const container = control._container = control.onAdd(this);
```
lucaswoj: We give the control a reference to the map via both the `_map` property and the `onAdd` method. Is there a reason to give this reference twice? Could we standardize on one?

js/ui/map.js
```diff
@@ -212,14 +212,37 @@ class Map extends Camera {
...
+        if (control._position) {
+            const corner = this._controlCorners[control._position];
+            container.className += ' mapboxgl-ctrl';
+            if (control._position.indexOf('bottom') !== -1) {
```
lucaswoj: If we're going to be accessing the `_position`, `_map`, and `_container` properties within `Map`, we should remove the `_underscore` to indicate that they are in the public API.

js/ui/map.js
```diff
@@ -212,14 +212,37 @@ class Map extends Camera {
...
+                corner.insertBefore(container, corner.firstChild);
+            } else {
+                corner.appendChild(container);
+            }
+        }
+        return this;
```
lucaswoj: Where / how do you plan to document the `Control` API?

tmcw:
cc @lucaswoj for re-review.

mourner:
@tmcw the build still fails.

mourner:
```js
test/js/ui/map.test.js Map TypeError: Cannot use 'in' operator to search for 'nodeType' in undefined:
     Error: TypeError: Cannot use 'in' operator to search for 'nodeType' in undefined
      at Map.addControl (js/ui/map.js:78:488)
```

tmcw:
Okay, fixed that test bug.

tmcw:
ready for review @lucaswoj @mourner

tmcw:
👍 switched from arrow functions to bindAll

js/ui/map.js
```diff
@@ -212,14 +212,32 @@ class Map extends Camera {
...
+     * Adds a [`Control`](#Control) to the map, calling `control.onAdd(this)`.
      *
      * @param {Control} control The [`Control`](#Control) to add.
+     * @param {string} corner corner of the map to which the control will be added
      * @returns {Map} `this`
      * @see [Display map navigation controls](https://www.mapbox.com/mapbox-gl-js/example/navigation/)
      */
-    addControl(control) {
-        control.addTo(this);
+    addControl(control, corner = 'top-right') {
```
mourner: Our default Node version for GL JS is still v4, so we can't use default parameters (which are v6+ I think).

js/ui/map.js
```diff
@@ -1214,6 +1232,89 @@ function removeNode(node) {
...
+/**
+ * A [`LngLat`](#LngLat) object or an array of two numbers representing longitude and latitude.
+ *
+ * @typedef {(LngLat | Array<number>)} LngLatLike
+ * @example
+ * var v1 = new mapboxgl.LngLat(-122.420679, 37.772537);
+ * var v2 = [-122.420679, 37.772537];
+ */
+
+/**
```
mourner: Looks like this got duplicated accidentally (same typedef below)

mourner:
One last thing I'm thinking about — what do you think about readding the ability for the control to have a default position by adding an optional `getDefaultPosition()` method or `defaultPosition` property (optional) to the interface? This will make it slightly easier for people to use existing controls as they did before, without them having to update all `addControl` calls.

tmcw:
👍 now implements getDefaultPosition, uses it for attributioncontrol and scalecontrol, and documents it in IControl

PR #3498, opened 2016-10-31
Simplify & optimize style calculation

mourner:
A major cleanup of the style recalculation code. 50 less lines of code, much less confusion and also slightly improves performance. All commits are self-contained so review commit by commit.

One noteable architectural change is that `globalProperties` properties like `zoomHistory`, `time` and `duration` that were specific to transition implementation are moved out and live where they belong. So now `globalProperties` are always just `{zoom}`, although I didn't replace it with just `zoom` yet because @lucaswoj anticipates other global properties in future.

benchmark | master 8127e17 | simplify-calculate 55faf5f
--- | --- | ---
**frame-duration** | 6.9 ms, 0% > 16ms  | 6.4 ms, 0% > 16ms 

👀 @lucaswoj @jfirebaugh


mourner:
Had to rebase to resolve a minor conflict with Flow PR, but made sure the tests pass on each commit so that they can be merged without squashing.

test/js/style/style_declaration.test.js
```diff
@@ -27,44 +27,12 @@ test('StyleDeclaration', (t) => {
...
-            unset.calculate({
-                zoom: 1,
-                zoomHistory: { lastIntegerZoomTime: 0, lastIntegerZoom: 0 },
-                duration: 300
-            }),
-            undefined
-        );
-
-        t.end();
-    });
```
lucaswoj: Why did you delete these tests?

lucaswoj: Ah. I see. 👇

mourner: I moved it into a new `style_transition.test.js`

js/style/style_transition.js
```diff
@@ -40,45 +38,59 @@ class StyleTransition {
...
+        if (time >= this.endTime)
+            return value;
 
-        return value;
+        const oldValue = this.oldTransition.calculate(globalProperties, featureProperties, this.startTime);
+        const t = util.easeCubicInOut((time - this.startTime - this.delay) / this.duration);
+        return this.interp(oldValue, value, t);
+    }
+
+    _calculateTargetValue(globalProperties, featureProperties) {
```
mourner: This code earlier lived mostly in `style_declaration` `wrapTransitionedCalculate` [here](https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style_declaration.js#L58-L96). What this method does is calculate the value the property is _transitioning to_, as opposed to `calculate` which returns the current value that is transitioning from the old value to the target one with time — please see `calculate` above.

Issue #3385, opened 2016-10-14
Resolve units of measure in the Z direction


peterqliu:
As we flesh out our offerings in 3D space, it becomes necessary to render multiple layers in the same 3D space: labels atop extrusions, lofted lines alongside extrusions, etc. For `stuff to line up` properly, we need all layers to speak the same (or easily convertible) measurement language(s).

Currently, there are two things that "poke up" off the map:

- Extrusions. Relevant properties are `fill-extrusion-height` and `fill-extrusion-base`, which are specified in a **nameless, but real-world unit** that scale automatically with zoom.

- Point symbols, mediated by `text/symbol-translate`. This is specified in **pixels**, which are independent of zoom.

What should be our paradigm, moving forward?

lucaswoj:
We should measure extrusions in pixels. 

Everything else in GL is measured in pixels or ems. 

The need for a pixel -> meter conversion is a cross-cutting concern that suggests a separate feature implemented at the function level.

lbud:
Extrusions in pixels is a pretty weird concept to me because you'll actually never see the number of pixels specified, since we don't pitch the map to 90º. While presumably for any other properties measured in pixels you see variable sizes based on pitch, you would see the specified value for an unpitched (default) map, whereas in the case of extrusions on an unpitched map, while centered on an object you would actually see no vertical lines, and for features around it you'll see variable-length vertical lines depending on how far they are from the center of a viewport… 🤔

ansis:
> Extrusions in pixels is a pretty weird concept to me

I agree, pixels seem weird. I think some real-world units (meters) makes more sense as a default here. 

> The need for a pixel -> meter conversion is a cross-cutting concern (i.e. is needed on lots of other properties in lots of other situations). That suggests a separate feature implemented at the function level.

I agree but I think we can still make meter the default here.

peterqliu:
Real-world units are convenient when we think about buildings, but introduce some weird problems:

- **It binds gl-js to Web Mercator**, in order to make the meters-to-pixels conversion. This is less an issue of "let's support all projections!", than "gl-js has stayed agnostic to specific implementations, and this is a step in the opposite direction."

- **m-to-px conversion is dependent on latitude.** Should we calculate latitude separately for every vertex of the polygon, or just the centroid, or some other way? The former would lead to pitched roofs (trivial on the building scale, significant on the country scale), while the latter would make for crummy approximations at large scales.

[...]

PR #3509, opened 2016-11-01
Resolve Z units to meters 

lbud:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/3385.

I branched off the fill extrusion separate type branch and would like to merge this back into there. Lint tests are passing because of the new ESLint formatting — @mourner fixed this on master. I'd like to merge this (not passing) into the fill-extrusion-type branch and then just rebase once there.

My question is this: [this line](https://github.com/mapbox/mapbox-gl-js/blob/0f01a4ee45bcf79fdd716d53e00fe02e7dfbf0b5/js/geo/transform.js#L454) is affected when the map bearing is non-zero. Then here I'm introducing a z scaling in the same matrix. Effectively this means there is a single test in map.test.js failing [here](https://github.com/mapbox/mapbox-gl-js/blob/0f01a4ee45bcf79fdd716d53e00fe02e7dfbf0b5/test/js/ui/map.test.js#L372-L378), as 
```
      Error: should be equivalent
      + expected - actual

       [
         [
           "-49.7184455522"
      -    "0.0000000000"
      +    "-0.0000000000"
         ]
         [
           "49.7184455522"
      -    "0.0000000000"
      +    "-0.0000000000"
         ]
       ]
```
— before `toFixed` that number is something like -2<sup>-13</sup>, so it introduces a miniscule change. That's the only test that fails, and all non-extrusion render tests pass; is it okay to just update this test to parse/round a bit differently to make this test pass? Or should the matrix scaling happen elsewhere/will there be unintended consequences of scaling here (after the `rotateZ`)?

@ansis @jfirebaugh @lucaswoj

### DIFF
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -453,6 +453,10 @@ class Transform {
         mat4.rotateX(m, m, this._pitch);
         mat4.rotateZ(m, m, this.angle);
         mat4.translate(m, m, [-this.x, -this.y, 0]);
+        mat4.scale(m, m, [1, 1,
+            // scale vertically to meters per pixel (inverse of ground resolution)
+            (Math.pow(2, this.zoom) * 512) / (2 * Math.PI * 6378137 * Math.abs(Math.cos(this.center.lat * (Math.PI / 180)))),
+        1]);
 
         this.projMatrix = m;
 
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -456,7 +456,7 @@ class Transform {
         mat4.scale(m, m, [1, 1,
             // scale vertically to meters per pixel (inverse of ground resolution)
             (Math.pow(2, this.zoom) * 512) / (2 * Math.PI * 6378137 * Math.abs(Math.cos(this.center.lat * (Math.PI / 180)))),
-        1]);
+            1]);
 
         this.projMatrix = m;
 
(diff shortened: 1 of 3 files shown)


js/geo/transform.js
```diff
@@ -453,6 +453,10 @@ class Transform {
         mat4.rotateX(m, m, this._pitch);
         mat4.rotateZ(m, m, this.angle);
         mat4.translate(m, m, [-this.x, -this.y, 0]);
+        mat4.scale(m, m, [1, 1,
+            // scale vertically to meters per pixel (inverse of ground resolution)
+            (Math.pow(2, this.zoom) * 512) / (2 * Math.PI * 6378137 * Math.abs(Math.cos(this.center.lat * (Math.PI / 180)))),
```
ansis: Can you document or split this up a bit more? maybe replace `2 * Math.PI * 6378137` with a `circumferenceOfEarth` constant

ansis:
This looks good. Have you checked this visually? If you make a cube from real-world distances does it get rendered as a cube?

> is it okay to just update this test to parse/round a bit differently to make this test pass?

yeah, I think so

lbud:
@ansis yep
image
perfect cube in NYC
image
perfectly square building face at the equator

PR #3510, opened 2016-11-01
Simplify drawBackground & optimize patterns

mourner:
- Break `setPattern` into two phases (one is done once per layer and the other once per tile).
- Reuse `setPattern` in `drawBackground`, removing lots of duplicated code.

👀 @ansis @lucaswoj

### DIFF
--- a/js/render/draw_extrusion.js
+++ b/js/render/draw_extrusion.js
@@ -159,7 +159,8 @@ function drawExtrusion(painter, source, layer, coord) {
     programConfiguration.setUniforms(gl, program, layer, {zoom: painter.transform.zoom});
 
     if (image) {
-        setPattern(image, tile, coord, painter, program, true);
+        setPattern(image, painter, program);
+        setPattern.setTile(tile, painter, program, true);
     }
 
     setMatrix(program, painter, coord, tile, layer);
--- a/js/render/set_pattern.js
+++ b/js/render/set_pattern.js
@@ -4,7 +4,7 @@ const pixelsToTileUnits = require('../source/pixels_to_tile_units');
 
 module.exports = setPattern;
 
-function setPattern(image, tile, coord, painter, program, includeHeightFactor) {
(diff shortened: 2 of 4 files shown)


js/render/draw_extrusion.js
```diff
@@ -159,7 +159,8 @@ function drawExtrusion(painter, source, layer, coord) {
     programConfiguration.setUniforms(gl, program, layer, {zoom: painter.transform.zoom});
 
     if (image) {
-        setPattern(image, tile, coord, painter, program, true);
+        setPattern(image, painter, program);
+        setPattern.setTile(tile, painter, program, true);
```
ansis: This is a tiny detail, but can `set_pattern.js` export both functions as part of an object instead of making `setTile` a property on `setPattern`?

js/render/set_pattern.js
```diff
@@ -17,25 +17,31 @@ function setPattern(image, tile, coord, painter, program, includeHeightFactor) {
...
     const pixelY = tileSizeAtNearestZoom * tile.coord.y;
+
     // split the pixel coord into two pairs of 16 bit numbers. The glsl spec only guarantees 16 bits of precision.
     gl.uniform2f(program.u_pixel_coord_upper, pixelX >> 16, pixelY >> 16);
     gl.uniform2f(program.u_pixel_coord_lower, pixelX & 0xFFFF, pixelY & 0xFFFF);
 
-    if (includeHeightFactor) {
-        gl.uniform1f(program.u_height_factor, -Math.pow(2, painter.transform.tileZoom) / tileSizeAtNearestZoom >> 3);
+    if (isExtruded) {
+        gl.uniform1f(program.u_height_factor, -numTiles / tile.tileSize / 8);
```
ansis: Would this be cleaner as part of `drawExtrusion`? it's only called from one spot and only shares `numTiles = Math.pow(2, tile.coord.z);` with the rest of the function.

PR #3514, opened 2016-11-02
Simplify sprite image loading

mourner:
👀 @lucaswoj

### DIFF
--- a/js/util/ajax.js
+++ b/js/util/ajax.js
@@ -59,15 +59,7 @@ exports.getImage = function(url, callback) {
     return exports.getArrayBuffer(url, (err, imgData) => {
         if (err) return callback(err);
         const img = new window.Image();
-        img.onload = () => {
-            const canvas = window.document.createElement('canvas');
-            const context = canvas.getContext('2d');
-            canvas.width = img.width;
-            canvas.height = img.height;
-            context.drawImage(img, 0, 0);
-            img.data = context.getImageData(0, 0, img.width, img.height).data;
-            callback(null, img);
-        };
+        img.onload = () => callback(null, img);
         if (!sameOrigin(url)) {
             img.crossOrigin = "Anonymous";
         }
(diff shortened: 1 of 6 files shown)


js/util/ajax.js
```diff
@@ -59,24 +59,19 @@ exports.getImage = function(url, callback) {
...
-        }
-        img.getData = function() {
+        img.onload = () => {
             const canvas = window.document.createElement('canvas');
             const context = canvas.getContext('2d');
             canvas.width = img.width;
             canvas.height = img.height;
             context.drawImage(img, 0, 0);
-            return context.getImageData(0, 0, img.width, img.height).data;
+            img.data = context.getImageData(0, 0, img.width, img.height).data;
```
lucaswoj: `img` is an instance of [`HTMLImageElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/Image) which does not have a `data` property. It might be better to return a vanilla object with `width`, `height`, `data`, and `complete` properties.

js/util/ajax.js
```diff
@@ -59,24 +59,19 @@ exports.getImage = function(url, callback) {
...
+            callback(null, img);
         };
+        if (!sameOrigin(url)) {
+            img.crossOrigin = "Anonymous";
```
lucaswoj: Are you sure `Anonymous` should be capitalized? https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes

mourner: On the other hand, this example https://developer.mozilla.org/ru/docs/Web/HTML/CORS_enabled_image uses capitalized version. Video code below (unchanged) used capitalized version too.

js/util/ajax.js
mourner: Not does it have `getData()` method it was augmented with earlier. There are some places in the code that expect `getImage` to return the image object rather than underlying data, such as `ImageSource`. On the other hand, moving canvas-based getData code to `ImageSprite` will make stubbing in Node more difficult because you have to stub its methods rather than `ajax.getImage`. So I suggest leaving this as is for now.

mourner: On a second thought, we can make this much cleaner by introducing a separate `browser.getImageData` utility function and stubbing it out.

mourner:
@lucaswoj added another commit that makes handling/stubbing of image data much cleaner.

PR #3518, opened 2016-11-02
Calculate matrices after constraining

### LINKED ISSUE #3474: fitBounds randomly throw failed to invert matrix
Encountered on versions 0.21.0 && 0.25.1

Sorry about the "**randomly** issue", hard to reproduce, and I believe it's a size / timing issue. Which is also why I got fed-up with trying to debug it and just have hanging breakpoints waiting for next occurrences.

I'm sharing my finding hoping someone can help pin-point the root issue.

[`Transform.latY(lat, worldSize)`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/geo/transform.js#L224) Sometimes receive a `lat` of `-90` and then return `Infinity`. (Is that expected???)

``` javascript
const y = 180 / Math.PI * Math.log(Math.tan(Math.PI / 4 + lat * Math.PI / 360));
```

This results is then used (through `this.y`) to calculate the matrices:
https://github.com/mapbox/mapbox-gl-js/blob/master/js/geo/transform.js#L455

Resulting in `NaN` which end up raising the `failed to invert matrix` https://github.com/mapbox/mapbox-gl-js/blob/master/js/geo/transform.js#L467

I don't yet understand why `latY()` randomly receives `lat: -90`, but this all start with `fitBounds` with options: `linear: true`, a padding and offset.

mourner:
Fixes #3474 by putting of calculating matrices until the view has been constrained to valid bounds. cc @nrako

### DIFF
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -112,8 +112,8 @@ class Transform {
         this.scale = this.zoomScale(z);
         this.tileZoom = Math.floor(z);
         this.zoomFraction = z - this.tileZoom;
-        this._calcMatrices();
         this._constrain();
+        this._calcMatrices();
     }
 
     get center() { return this._center; }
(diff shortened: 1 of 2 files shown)


nrako:
@mourner 👍 thanks!

PR #3520, opened 2016-11-02
Switch to pure uglify for a smaller build

mourner:
Updates `build-min` script to use UglifyJS directly rather than through `minifyify` plugin. Reasons:

- Uglifying the full build makes it ~7% smaller (130kb to 121kb gzipped, 507kb to 460kb original).
- Minifyify is [not actively maintained](https://github.com/ben-ng/minifyify#warning-unmaintained).

Source maps still seem to work, at least when running profiler, similarly to how it worked before (mostly OK but with some files not properly fetched). Although there are more mangled names in the display.

I'm still not fully sure we should merge this but decided to show the results of my experiment in case you have any thoughts.

👀 @jfirebaugh @lucaswoj


package.json
```diff
@@ -98,7 +98,7 @@
   "scripts": {
     "build-dev": "browserify js/mapbox-gl.js --debug --ignore-transform unassertify --standalone mapboxgl > dist/mapbox-gl-dev.js && tap --no-coverage test/build/dev.test.js",
     "watch-dev": "watchify js/mapbox-gl.js --debug --ignore-transform unassertify --standalone mapboxgl --outfile dist/mapbox-gl-dev.js --verbose",
-    "build-min": "browserify js/mapbox-gl.js --debug --plugin [minifyify --map mapbox-gl.js.map --output dist/mapbox-gl.js.map] --standalone mapboxgl > dist/mapbox-gl.js && tap --no-coverage test/build/min.test.js",
+    "build-min": "browserify js/mapbox-gl.js --debug --standalone mapboxgl | exorcist dist/mapbox-gl.js.map > dist/mapbox-gl.js && cd dist && uglifyjs --compress warnings=false --mangle --screw-ie8 --in-source-map mapbox-gl.js.map --source-map mapbox-gl.js.map mapbox-gl.js -o mapbox-gl.js && tap --no-coverage ../test/build/min.test.js",
```
lucaswoj: `--screw-ie8` 😂

lucaswoj:
> I'm still not fully sure we should merge this but decided to show the results of my experiment in case you have any thoughts.

What are you unsure about? Source maps on the production build are just as unusable as before. If anything, it seems that switching to a more actively maintained project would increase our chances of making the source maps work in the future.

mourner:
> If you are currently using uglifyify and realized that your sourcemap is behaving strangely, you're not alone. Minifyify builds its own sourcemap instead of depending on uglify-js's insourcemap option, and has proven to be more reliable in practice.

@lucaswoj one concern is that Minifyify readme claims that UglifyJS source map generation is less reliable that Minifyify, although I'm not sure to what extent this is true. Another one is that names are more mangled in the profiler, although I'm not sure whether it's due to the aforementioned source map "unreliability" or just more effective minification. If you think that's all OK, I'm up to merging.

@jfirebaugh would like to hear your opinion too.

jfirebaugh:
Source maps are supposed to include information mapping mangled names back to the original, so it's somewhat concerning if they come out more mangled than before. But I've basically given up ever being able to debug uglified code, so ¯\\_(ツ)_/¯.

mourner:
I'll investigate the source maps difference to see if there's any change to improve it before merging.

stevewillard:
@mourner what ever happened with this PR? The results seem promising. Maybe I can take a shot at switching over to pure Uglify.

jfirebaugh:
Closing since this is stale and probably addressed by the switch to rollup anyway.

PR #3521, opened 2016-11-03
Optimize drawing solid backgrounds

### LINKED ISSUE #2773: Use glClear to draw solid background fills if they are the bottommost layer
This optimization is implemented in GL Native and is missing in GL JS.

Relevant GL native commits:
- [use glClear to draw solid background fills](https://github.com/mapbox/mapbox-gl-native/commits/d400fe0c91b66e066f50dbf3615db68af8895a10)
- [[core] Don't use glClear optimization for layers that aren't the bottommost](https://github.com/mapbox/mapbox-gl-native/commits/4a803f1a78d6d9adc603129c369cb75592c408b7)

:eyes: @mourner @lucaswoj @kkaefer

### LINKED ISSUE #2925: Wrong rendering results with background layer with opacity as overlay
If you have something like this in your style:

```
  "layers": [
    {   
      "id": "raster-none",
      "type": "raster",
      "source": "satellite",
      "paint": {
        "raster-fade-duration": 0
      }   
    },  
    {   
      "id": "background_overlay",
      "type": "background",
      "paint": {
        "background-opacity": 0.5,
        "background-color": "green"
      }   
    }   
  ]
```

You would expect this:
expected

But you get this:
actual

I wrote a test case for it here:
https://github.com/mapbox/mapbox-gl-test-suite/issues/130

mourner:
Closes #2773, closes #2925. There seems to be no FPS difference for me, but it might help on devices with weaker GPUs (e.g. phones).

👀 @ansis @brunoabinader

### DIFF
--- a/js/render/draw_background.js
+++ b/js/render/draw_background.js
@@ -6,13 +6,15 @@ module.exports = drawBackground;
 
 function drawBackground(painter, sourceCache, layer) {
     const gl = painter.gl;
-    const transform = painter.transform;
-    const tileSize = transform.tileSize;
     const color = layer.paint['background-color'];
     const image = layer.paint['background-pattern'];
     const opacity = layer.paint['background-opacity'];
 
     if (painter.isOpaquePass !== (!image && color[3] === 1)) return;
+
+    // if the background layer is bottommost and not patterned,
+    // we render it with gl.clearColor earlier
+    if (!image && painter.currentLayer === 0) return;
 
     gl.disable(gl.STENCIL_TEST);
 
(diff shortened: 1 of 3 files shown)


js/render/draw_background.js
```diff
@@ -6,14 +6,16 @@ module.exports = drawBackground;
 
 function drawBackground(painter, sourceCache, layer) {
     const gl = painter.gl;
-    const transform = painter.transform;
-    const tileSize = transform.tileSize;
     const color = layer.paint['background-color'];
     const image = layer.paint['background-pattern'];
     const opacity = layer.paint['background-opacity'];
 
     if (painter.isOpaquePass !== (!image && color[3] === 1)) return;
```
ansis: This is not related to your changes, but it looks like this doesn't consider `opacity` when it should. If it's convenient it would nice to squeeze that in here

mourner: Whoa, you're right! Fixing this also fixes https://github.com/mapbox/mapbox-gl-js/issues/2925 🎉

PR #3527, opened 2016-11-03
Simplify data-driven paint attributes implementation

### LINKED ISSUE #3525: Drive paint attribute binding logic off the style spec
I'd like to see [declarations like this](https://github.com/mapbox/mapbox-gl-js/blob/master/js/data/bucket/circle_bucket.js#L17-L56) replaced with:

```
paintProperties: ['circle-color', 'circle-radius', 'circle-blur', 'circle-opacity']
```

And ideally, even the list of data-driven properties would itself be derived from the style spec.

The logic for how to bind a paint property to an attribute or uniform can be driven by style specification properties, and dependent not on the individual property but on its type. All `color` properties are bound the same way. Numeric properties with a spec-defined minimum and maximum can derive an integer size and `multiplier` from that range. Numeric properties where the minimum or maximum value is not defined should probably just use a float attribute, to avoid issues like #3464.

mourner:
A work in progress, closes #3525. cc @jfirebaugh @lucaswoj 

The tests are failing because of `bucket.test.js`, which is mocked in a very complex way so I can't figure out how to fix a few of its failing assertions yet.

### DIFF
--- a/js/data/program_configuration.js
+++ b/js/data/program_configuration.js
@@ -42,6 +42,7 @@ class ProgramConfiguration {
             const multiplier = attribute.multiplier;
 
             const vert = self.getVertexPragmas(name);
+            const frag = self.getFragmentPragmas(name);
 
             if (layer.isPaintValueFeatureConstant(attribute.paintProperty)) {
                 self.uniforms.push(attribute);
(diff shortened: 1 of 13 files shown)


js/style/style_layer/fill_extrusion_style_layer.js
```diff
@@ -4,6 +4,15 @@ const StyleLayer = require('../style_layer');
 const FillExtrusionBucket = require('../../data/bucket/fill_extrusion_bucket');
 
 class FillExtrusionStyleLayer extends StyleLayer {
+
+    getPaintValue(name, globalProperties, featureProperties) {
+        const value = super.getPaintValue(name, globalProperties, featureProperties);
+        if (name === 'fill-extrusion-color') {
+            value[3] = 1;
```
jfirebaugh: Could we do this in the vertex shader instead?

jfirebaugh:
After #3523 I believe we have a regular naming convention for the attribute/uniform name -- it's the name of the paint property with the layer type prefix removed.

I haven't looked at the `bucket.test.js` failures, but basically the one thing that it mocks is the feature objects. The other collaborator objects in the test -- style layer and properties, bucket subclass, and program interface -- are the real thing.

js/style/style_layer/fill_extrusion_style_layer.js
mourner: Yes, thought about that too. I just placed it here temporarily.

mourner:
Yes, noticed about attribute/uniform name convention too — I figured this was the reason you asked for #3523. :)

I'm still unsure about deriving type/multipliers — we can do it for colors, but for things like circle-blur and fill-extrusion-height, converting them to float doubles the memory requirement (16 -> 32 bits), and picking the right `maximum` in the spec with the intent to make it automagically derived as `Uint16` feels a bit too magic and fragile to me. Thoughts?

mourner:
@lucaswoj @jfirebaugh @lbud @ansis this PR is ready for review. It didn't accomplish everything that was planned in #3525 fully, but instead led me to significantly improve the clarity of data-driven-properties code (and -100 lines net!). 

Performance also slightly improved — from 6.1ms to 5.5ms on the Streets z13 benchmark. Buffer bench results are unchanged. All commits are self-contained and pass tests.

I suggest merging this now and following up with the points below in other PRs:

- Move extrude color alpha ignore logic to the shaders.
- Derive type and multiplier from spec min/max.

js/style/style_layer/fill_extrusion_style_layer.js
mourner: Looking at the vertex shader, it seems that `fill-extrusion-color` alpha value is not used there at all. So we could simply remove this code. @lbud does this sound right?

js/data/program_configuration.js
```diff
@@ -22,227 +20,204 @@ const assert = require('assert');
...
-        const fragmentInit = self.fragmentPragmas.initialize;
-        const fragmentDefine = self.fragmentPragmas.define;
-        const vertexInit = self.vertexPragmas.initialize;
-        const vertexDefine = self.vertexPragmas.define;
+            if (layer.isPaintValueFeatureConstant(attribute.property)) {
+                self.addZoomDrivenAttribute(name, attribute);
+            } else if (layer.isPaintValueZoomConstant(attribute.property)) {
+                self.addDataDrivenAttribute(name, attribute);
+            } else {
+                self.addDataAndZoomDrivenAttribute(name, attribute, layer, zoom);
```
lucaswoj: In the codebase we use the terminology `property functions`, `zoom functions`, and `zoom and property functions`."data-driven" is a broader designation that may include unrelated features in the future.

lucaswoj: We could call these methods `addZoomAttribute`, `addPropertyAttribute`, and `addZoomAndPropertyAttribute`

lucaswoj:
This looks great! I have couple thoughts on terminology in `ProgramConfiguration`. I feel 👍 on merging this once those and https://github.com/mapbox/mapbox-gl-js/pull/3527#pullrequestreview-7027530 are cleared up.

jfirebaugh:
It's definitely easier to understand `ProgramConfiguration` now. 👍 

In gl-native, the implementation is going to center entirely around paint properties. We use code generation to create a unique type for each paint property, which can provide to C++ the data that's in the style specification. Having unique types allows the use of [C++ structural metaprogramming techniques](https://github.com/mapbox/cpp/blob/master/C%2B%2B%20Structural%20Metaprogramming.md) to define generic operations over tuples of paint properties. There will be an equivalent of `ProgramConfiguration`, but it's likely to use a more functional programming style, where things like the list of uniforms and pragmas are generated as needed from a tuple type representing the DDS-enabled paint properties.

In the initial implementation, I'm not going to use multipliers. `fill-extrusion-height` will use a `float` attribute. `*-color` and `*-opacity` will use [normalized](https://www.opengl.org/wiki/Normalized_Integer) `uint8`s.

I'm also going to use `glVertexAttrib*` rather than dynamically generated shaders as a first cut.

js/style/style_layer/fill_extrusion_style_layer.js
lbud: The reason for this is that if you provide an rgba color in the style, we premultiply it in StyleDeclaration, but we don't render using the alpha channel, so providing a non-opaque rgba value leads to unexpected color results. We decided the best of several not-great approaches here would be to deliberately ignore the alpha channel before premultiplying.

PR #3531, opened 2016-11-03
add geolocation example

incanus:
What it says on the box. 

/cc @tristen @YunjieLi

### DIFF
--- a/docs/_posts/examples/3400-01-05-locate-user.html
+++ b/docs/_posts/examples/3400-01-05-locate-user.html
@@ -1,0 +1,20 @@
+---
+layout: example
+category: example
+title: Locate the user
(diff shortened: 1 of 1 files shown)


tristen:
Just a thought that this example could demonstrate a bit more of it's API. There isn't example showing events tied to this control. i.e

```js
geolocate.on('geolocate', function(e) {
    // console.log(e);
});

geolocate.on('error', function(e) {
    // console.log(e);
});
```

How about setting a point on result [like this](https://www.mapbox.com/mapbox-gl-js/example/point-from-geocoder-result/) and window alerting if an error is thrown? (errors happen frequently with `NavigatorGeolocation`).

Issue #3540, opened 2016-11-04
Map.querySourceFeatures throws a TypeError when parameters option not included


andrewharvey:
**mapbox-gl-js version**: 0.26.0

http://jsbin.com/hoqobapaha/edit?html,console,output

### Steps to Trigger Behavior

 1. `map.querySourceFeatures('source');` omitting the optional `parameters`

### Expected Behavior

No error. The documentation seems to indicate `parameters` is optional.

### Actual Behavior

`tile.js:169 Uncaught TypeError: Cannot read property 'filter' of undefined` at https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/tile.js#L177

I've created a PR at https://github.com/mapbox/mapbox-gl-js/pull/3542

mapsam:
Closed by #3542

PR #3542, opened 2016-11-04
Fix error where querySourceFeatures has parameters omitted. #3540

andrewharvey:
## Launch Checklist

See #3540 for background on what this fixes.

There are a couple of places where this could be fixed, so I'm not sure where is best higher or lower in the call stack. I've opted to place it at the bottom of the stack.

The test is quite verbose because if you don't add any features to the source and if this source isn't included in a source the code path which I ran into this error isn't triggered.

If you remove the fix the test fails if you include this fix the test passes.

I've included a second commit which makes `parameters` optional although I'm not sure if this is correct as it's only optional for GeoJSON sources as per current documentation.

I needed to rename `params` as otherwise it would show up in the generated docs separately.

 - [ ] post benchmark scores

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -575,7 +575,7 @@ class Map extends Camera {
      * representing features within the specified vector tile or GeoJSON source that satisfy the query parameters.
      *
      * @param {string} sourceID The ID of the vector tile or GeoJSON source to query.
-     * @param {Object} parameters
+     * @param {Object} [parameters]
      * @param {string} [parameters.sourceLayer] The name of the vector tile layer to query. *For vector tile
      *   sources, this parameter is required.* For GeoJSON sources, it is ignored.
      * @param {Array} [parameters.filter] A [filter](https://www.mapbox.com/mapbox-gl-style-spec/#types-filter)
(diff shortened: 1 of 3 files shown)


mourner:
Can you please fix the linting errors?

andrewharvey:
> Can you please fix the linting errors?

Sorry I ran the unit tests but failed to run the full test suite.

Should be fixed now, I just overwrote the commits so not sure if ci automatically reruns or not...

js/ui/map.js
```diff
@@ -575,7 +575,7 @@ class Map extends Camera {
      * representing features within the specified vector tile or GeoJSON source that satisfy the query parameters.
      *
      * @param {string} sourceID The ID of the vector tile or GeoJSON source to query.
-     * @param {Object} parameters
+     * @param {Object} [parameters]
```
lucaswoj: Got it. This is more of a feature addition than a bug fix. We never supported or claimed to support omitting `parameters`. 👍

test/js/ui/map.test.js
```diff
@@ -646,6 +647,50 @@ test('Map', (t) => {
         t.end();
     });
 
+    t.test('#querySourceFeatures', (t) => {
```
lucaswoj: `querySourceFeatures` tests are currently split between [`tile.test.js`](http://github.com/mapbox/mapbox-gl-js/blob/0088929dcc934d1f41d5aec5c5bfe9bdd165f89f/test/js/source/tile.test.js#L38-L39) and [`style.test.js`](http://github.com/mapbox/mapbox-gl-js/blob/0088929dcc934d1f41d5aec5c5bfe9bdd165f89f/test/js/style/style.test.js#L1179-L1180). Could you move this test to one of those two places?

js/ui/map.js
andrewharvey: I suppose so it is :wink: 

Given `parameters.sourceLayer` and `parameters.filter` were optional I incorrectly assumed I could omit `parameters` entirely but I guess the JS doc says you still need to provide an empty object.

lucaswoj:
Looks great! Thank you so much!

PR #3549, opened 2016-11-05
Fix color buffer conflicts between background rendering and extrusions

lbud:
@mourner — in adding a regression test for the issue of color buffer conflict addressed in https://github.com/mapbox/mapbox-gl-js/commit/5b9ba0212df3810c4ea7d1be02c53de101f5e02e, I realized this wasn't the full fix — clearing the color buffer to (0,0,0,0) in draw_fill_extrusion causes the background not to render. I created the expected image in the regression test (https://github.com/mapbox/mapbox-gl-test-suite/pull/160) by locally reverting changes from https://github.com/mapbox/mapbox-gl-js/commit/2fc3651139d18f4e68e6d7b5d43169ff4c63b568.

This PR updates the test-suite SHA to reflect the regression test; right now it will fail. I haven't figured out yet if there's a way to maintain the optimizations from https://github.com/mapbox/mapbox-gl-js/commit/2fc3651139d18f4e68e6d7b5d43169ff4c63b568 while still effectively clearing the color buffer when needed in drawing extrusions, but this needs further investigation as right now it is broken 🚑  planning to dig in more this weekend.

### DIFF
--- a/js/render/draw_background.js
+++ b/js/render/draw_background.js
@@ -6,16 +6,14 @@ module.exports = drawBackground;
 
 function drawBackground(painter, sourceCache, layer) {
     const gl = painter.gl;
+    const transform = painter.transform;
+    const tileSize = transform.tileSize;
     const color = layer.paint['background-color'];
     const image = layer.paint['background-pattern'];
     const opacity = layer.paint['background-opacity'];
 
     const isOpaque = !image && color[3] === 1 && opacity === 1;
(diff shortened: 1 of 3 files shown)


mourner:
cc @ansis

mourner:
I think it's best to just revert that commit and try again later if needed. The optimization is small and didn't show any noticeable FPS gains.

Issue #3467, opened 2016-10-26
Create example of multiple types of geometries coming from a GeoJSON layer


tmcw:
This question came up in support:

> How do I show multiple types of data - like lines _and_ points - from the same GeoJSON source?

We don't have an example that directly demonstrates this yet, and we should, since one part is rather simple - just reference the same example - and another is tricky - you might want to avoid adding lines around polygons and adding markers to vertices of non-Point geometries.

Creating this example would be a great starter task for someone from @mapbox/support who hasn't made an example yet!

captainbarbosa:
Completed in #3554!

PR #3554, opened 2016-11-07
Add new multiple geometry example

captainbarbosa:
This example explains how multiple geometries can be added as two separate layers from the same GeoJSON source.

Addresses #3467, re-do of #3553 to target `mb-pages` instead of `master`

cc @tmcw for review


mollymerp:
Thank you @captainbarbosa !

PR #3570, opened 2016-11-08
Fix js crash while trying to justify empty lines

CruelSilence:
There is no need to justify empty lines.
If we have a text that contains multiple empty lines in the middle, JS crashes in `justifyLine` function. 
The `start` glyph index is bigger than the `end` index in such a case.


mourner:
Great! Could you also add a test for that?

CruelSilence:
No problem. Just did =)

PR #3571, opened 2016-11-08
Use "navigator.permissions" to detect geolocation support

### LINKED ISSUE #3568: browser.supportsGeolocation does not detect geolocation disabled because of http urls
**mapbox-gl-js version**:

### Steps to Trigger Behavior

 1. Add the geolocation control on a http-origin page
 2. Click the geocoder control
 3. Geocoding fails because of [deprecation of http geocoding](https://sites.google.com/a/chromium.org/dev/Home/chromium-security/deprecating-powerful-features-on-insecure-origins)

### Expected Behavior

The control should be missing, like it is when the browser does not offer geolocation support at all.

### Actual Behavior

The control is missing but breaks on click

tmcw:
Due to new [security restrictions](https://sites.google.com/a/chromium.org/dev/Home/chromium-security/deprecating-powerful-features-on-insecure-origins), the presence of `navigator.geolocation` is no longer a complete indication of geolocation support: the `navigator.permissions` API allows one to query for access to the API. This PR moves geolocation support checking to the GeolocationControl in order to support this asynchronous API.

Testing plan:

1. `npm run start-debug`
2. edit /etc/hosts to give your localhost another creative name, since the Chrome HTTPS rule whitelists localhost
3. confirm that the control is not visible on the alternative name, but is visible on localhost

Implementation notes:

Since the permissions API is async and is not instant, this approach caches the result. I have not found any cases where the result will change.

## Launch Checklist

Fixes #3568 - ~~waiting on #3497 because this needs to move the geolocation detection into the geocodercontrol itself.~~ Actionable - workingonit.

 - [ ] post benchmark scores

### DIFF
--- a/js/ui/control/geolocate_control.js
+++ b/js/ui/control/geolocate_control.js
@@ -11,34 +11,22 @@ const className = 'mapboxgl-ctrl';
 let supportsGeolocation;
 
 function checkGeolocationSupport(callback) {
-
     if (supportsGeolocation !== undefined) {
-        if (supportsGeolocation === true) {
-            callback();
-        }
-        return;
-    }
+        callback(supportsGeolocation);
 
-    // Test for the case where a browser disables Geolocation because of an
-    // insecure origin
-    // https://sites.google.com/a/chromium.org/dev/Home/chromium-security/deprecating-powerful-features-on-insecure-origins
-    //
-    // navigator.permissions has incomplete browser support
-    // http://caniuse.com/#feat=permissions-api
-    if (window.navigator.permissions !== undefined) {
-        window.navigator.permissions
-            .query({ name:'geolocation' })
-            .then((p) => {
-                supportsGeolocation = p.state !== 'denied';
-                if (supportsGeolocation === true) {
-                    callback();
-                }
-            });
+    } else if (window.navigator.permissions !== undefined) {
+        // navigator.permissions has incomplete browser support
+        // http://caniuse.com/#feat=permissions-api
+        // Test for the case where a browser disables Geolocation because of an
+        // insecure origin
+        window.navigator.permissions.query({ name: 'geolocation' }).then((p) => {
+            supportsGeolocation = p.state !== 'denied';
+            callback(supportsGeolocation);
+        });
+
     } else {
         supportsGeolocation = !!window.navigator.geolocation;
-        if (supportsGeolocation === true) {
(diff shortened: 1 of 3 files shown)


mourner:
> we should move the geolocation detection to the control
itself

Yep! This will allow us to consistently disable geolocation if it's not supported, no matter at what time the geolocation control is added.

tmcw:
This is now ready for review @mourner @lucaswoj

js/ui/control/geolocate_control.js
```diff
@@ -1,18 +1,57 @@
...
+                    callback();
+                }
+            });
+    } else {
+        supportsGeolocation = !!window.navigator.geolocation;
+        if (supportsGeolocation === true) {
+            callback();
+        }
+    }
+}
```
mourner: I think it would be slightly cleaner to call the callback in any outcome, and move responsibility for checking `supportsGeolocation` value to `_setupUI`:

```js
function checkGeolocationSupport(callback) {
    if (supportsGeolocation !== undefined) {
        callback();
        
    } else if (window.navigator.permissions !== undefined) {
        window.navigator.permissions.query({name: 'geolocation'}).then((p) => {
            supportsGeolocation = p.state !== 'denied';
            callback();
        });
    } else {
        supportsGeolocation = !!window.navigator.geolocation;
        callback();
    }
}
```

js/ui/control/geolocate_control.js
```diff
@@ -1,18 +1,57 @@
...
 /**
  * A `GeolocateControl` control provides a button that uses the browser's geolocation
  * API to locate the user on the map.
  *
+ * Not all browsers support geolocation,
+ * and some users may disable the feature. Geolocation support for modern
+ * browsers including Chrome requires sites to be served over HTTPS. If
+ * geolocation support is not available, the GeolocateControl will not
+ * be visible.
+ *
```
mourner: BTW, should we limit the line width to 70 chars in comments like this? IMHO increasing the limit would help readability.

tmcw: Not sure how you're set up, but in vim-land I never use wrapping so cutting lines like this is more readable.

mourner: I don't use wrapping too, it just that 70 seems too short, given that we use up to 120 for the code.

PR #3573, opened 2016-11-09
Upgrade documentation deps

mourner:
👀 @tmcw


package.json
```diff
@@ -70,8 +70,8 @@
     "sinon": "^1.15.4",
     "st": "^1.2.0",
     "tap": "^5.7.0",
-    "vinyl": "1.1.1",
-    "vinyl-fs": "2.4.3",
+    "vinyl": "2.0.0",
```
tmcw: You might want to test this change - vinyl v2 introduces some big changes, like changing the type/key of file paths.

mourner: I ran `npm run start-docs` and didn't seem to notice anything broken.

Issue #3365, opened 2016-10-13
redoPlacement is broken for GeoJSON sources


jfirebaugh:
The ['redo placement' message](https://github.com/mapbox/mapbox-gl-js/blob/31523a851362e3d3feb8b33d78dc8f92d99d8977/js/source/tile.js#L131-L137) does not include a `type` key, so when the worker [receives](https://github.com/mapbox/mapbox-gl-js/blob/691f3fee4df842275dbae3c1704c2b9ca272e147/js/source/worker.js#L126-L129) the message, it defaults to `'vector'` and calls `redoPlacement` on the wrong `WorkerSource` object.

@lucaswoj @anandthakker Advice on fixing this? `Tile#redoPlacement` is passed a `SourceCache` (the parameter name is misleading) and I don't immediately see a way to get the source type from that.

anandthakker:
@jfirebaugh oof.  At one point I'd thought to require each Source to expose a `type` property, but nixed it because I thought it was an unnecessary complication.  Perhaps I should have kept it.

Given that the custom source API will undergo [heavy development/revision](https://github.com/mapbox/mapbox-gl-js/projects/2) (especially w/r/t workers), I'd propose tabling the general question for now.  For the immediate-term fix, note that GeoJSONSources [do expose `type`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/geojson_source.js#L101), so `Tile#redoPlacement` could get it by digging into the SourceCache's inner Source.

jfirebaugh:
Actually it looks like `Tile#redoPlacement` is normally called with a `Source` as the argument, not a `SourceCache`. [This](https://github.com/mapbox/mapbox-gl-js/blob/d60575a2ffe145c294d2e2da044939e3033979db/js/source/source_cache.js#L516-L516) is the exception; I'll try changing it to pass `this._source`.

PR #3582, opened 2016-11-09
Text placement is not redone during rotation

lucaswoj:
https://github.com/mapbox/mapbox-gl-js/pull/3490 marks `SymbolBucket#collisionBoxArray`, `SymbolBucket#symbolInstancesArray`, `SymbolBucket#symbolQuadsArray` as transferables. These `StructArray`s need to be persisted in the workerside `SymbolBucket` for layer use by `SymbolBucket#redoPlacement`.

This manifested as a bug in which text would render upside down after the map was rotated. 

expected

ref https://github.com/mapbox/mapbox-gl-test-suite/pull/163

cc @jfirebaugh @mourner 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/js/source/worker_tile.js
+++ b/js/source/worker_tile.js
@@ -116,9 +116,9 @@ class WorkerTile {
                 buckets: serializeBuckets(util.values(buckets), transferables),
                 featureIndex: featureIndex.serialize(transferables),
                 collisionTile: collisionTile.serialize(transferables),
-                collisionBoxArray: this.collisionBoxArray.serialize(transferables),
-                symbolInstancesArray: this.symbolInstancesArray.serialize(transferables),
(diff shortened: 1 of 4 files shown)


jfirebaugh:
Previously: #3365. Does the [regression test for that issue](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/regressions/mapbox-gl-js%233365) catch this one too, providing the additional assertions are added?

lucaswoj:
You can never blame yourself for breaking something that wasn't tested. No worries.

I renamed `StructArray#trim` to `StructArray#_trim` because it isn't called externally anywhere. It used to be called explicitly by `Bucket` subclasses.

lucaswoj:
@jfirebaugh By eyeballing, I think this new regression test would've caught #3365 with or without the new assertions.

Issue #3576, opened 2016-11-09
Add map.moveLayer method for changing layer z-order


lucaswoj:
This is a small piece of #1989 which will address one of the most common use cases. 

If a layer is removed and an identical layer is added immediately we should re-use the layer resources to avoid unnecessary computation and loading time.

mourner:
Per discussion in chat, instead of making remove/addLayer a noop (which needs to cover many edge cases and will still have to do some unnecessary work), we'll add a `map.moveLayer(id, before)` method that will update the layer's order (with an instant change if it's a non-symbol layer, and redoing placement if symbol).

anandthakker:
Is this done as of #3584 ?

jfirebaugh:
Yep.

PR #3584, opened 2016-11-10
Improve batch updates

mourner:
A work in progress on improving batch style updates, including fixing #3576.

- [x] more efficient `addLayer` and `removeLayer` methods
- [x] add `moveLayer` method
- [x] clean up source reloading logic
- [ ] (optional) update `moveLayer` only do `redoPlacement` instead of reparsing tiles for symbol layers

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -360,16 +360,14 @@ class Style extends Evented {
      * @returns {Style} `this`
      * @private
      */
-    addLayer(layer, before, options) {
+    addLayer(layerObject, before, options) {
         this._checkLoaded();
 
-        if (!(layer instanceof StyleLayer)) {
-            // this layer is not in the style.layers array, so we pass an impossible array index
-            if (this._validate(validateStyle.layer,
-                    `layers.${layer.id}`, layer, {arrayIndex: -1}, options)) return this;
+        // this layer is not in the style.layers array, so we pass an impossible array index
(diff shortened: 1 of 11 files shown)


mourner:
@lucaswoj this is ready for review. I initially wanted to optimize reordering symbol layers further by making them only do `redoPlacement` rather than reparsing the tiles, but this turned out to be a more involved refactor so I'm leaving that for another PR.

In addition to `moveLayer`, this PR makes `addLayer` and `removeLayer` more efficient.

js/style/style.js
```diff
@@ -385,68 +342,95 @@ class Style extends Evented {
...
+        if (!layer) throw new Error(`Layer not found: ${id}`);
+
+        const index = this._order.indexOf(id);
+        this._order.splice(index, 1);
+
+        const newIndex = before ? this._order.indexOf(before) : this._order.length;
+        this._order.splice(newIndex, 0, id);
+
+        if (layer.type === 'symbol') {
+            this._changed = true;
```
jfirebaugh: Should `this._changed` also get set if a non-symbol layer is moved?

mourner: OK, fixed with https://github.com/mapbox/mapbox-gl-js/pull/3584/commits/6ced8625e1e2e1b2b55d6b385dd5a24b06843962. I also realized that it didn't reset the update state in this case.

PR #3589, opened 2016-11-10
Fix flaky flyTo minZoom test

### LINKED ISSUE #3502: Flaky flyTo test
`camera #flyTo peaks at the specified zoom level` test is flaky and sometimes fails on Circle. Need a more reliable test. cc @1ec5 

```
 1) test/js/ui/camera.test.js camera #flyTo peaks at the specified zoom level expect truthy value:
     Error: expect truthy value
      at Camera.camera.on (test/js/ui/camera.test.js:991:19)
      at Camera.fire (js/util/evented.js:34:317)
      at Camera.<anonymous> (js/ui/camera.js:322:1233)
      at Camera._finishEase (js/ui/camera.js:329:132)
      at Camera.<anonymous> (js/ui/camera.js:327:591)
      at Immediate.tick (js/util/browser.js:7:1758)
```

mourner:
OK I think I figured out what this test was about. Closes #3502. cc @blanchg

### DIFF
--- a/test/js/ui/camera.test.js
+++ b/test/js/ui/camera.test.js
@@ -979,21 +979,21 @@ test('camera', (t) => {
         t.test('peaks at the specified zoom level', (t) => {
             const camera = createCamera();
             camera.setZoom(20);
-            let minZoom = Infinity;
+
+            const minZoom = 1;
 
             camera.on('zoom', () => {
-                if (camera.getZoom() < minZoom) {
-                    minZoom = camera.getZoom();
+                const zoom = camera.getZoom();
+                if (zoom < 1) {
+                    t.fail(`${zoom} should be >= ${minZoom} during flyTo`);
                 }
             });
 
             camera.on('moveend', () => {
-                t.ok(fixedNum(minZoom, 2) < 1.1);
-                t.ok(fixedNum(minZoom, 2) >= 1);
                 t.end();
             });
 
-            camera.flyTo({ center: [1, 0], zoom: 20, minZoom: 1 });
+            camera.flyTo({ center: [1, 0], zoom: 20, minZoom });
         });
 
         t.end();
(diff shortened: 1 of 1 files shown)


test/js/ui/camera.test.js
```diff
@@ -979,21 +979,21 @@ test('camera', (t) => {
...
             camera.setZoom(20);
-            let minZoom = Infinity;
+
+            const minZoom = 1;
 
             camera.on('zoom', () => {
-                if (camera.getZoom() < minZoom) {
-                    minZoom = camera.getZoom();
+                const zoom = camera.getZoom();
+                if (zoom < 1) {
```
mapsam: I like this, but we still need a check to ensure zoom has changed, which is integral to the `fly` animation. How about we increase the leniency of the `t.ok(fixedNum(minZoom, 2) < 1.1)` to something like `t.ok(fixedNum(minZoom, 2) < 2);`? The zoom event fires regularly enough that this won't be as flaky.

mapsam: Just confirmed that this test passes if we use the `easeTo` method as well, so adding this check back is important.

test/js/ui/camera.test.js
```diff
@@ -979,21 +979,21 @@ test('camera', (t) => {
...
             });
 
             camera.on('moveend', () => {
-                t.ok(fixedNum(minZoom, 2) < 1.1);
-                t.ok(fixedNum(minZoom, 2) >= 1);
                 t.end();
             });
 
-            camera.flyTo({ center: [1, 0], zoom: 20, minZoom: 1 });
+            camera.flyTo({ center: [1, 0], zoom: 20, minZoom });
```
mapsam: Does just adding `minZoom` equate to `minZoom: 1`?

mapsam: Cool!

test/js/ui/camera.test.js
mapsam: Updated, @mourner!

PR #3590, opened 2016-11-10
"data" event revamp

### LINKED ISSUE #3577: "isLoaded" property on "sourcedata" events
The 'sourceLoaded" property would equal the value of `SourceCache#loaded()` at the time that the data event is fired. `SourceCache#loaded()` is not in the public API and is useful information for consumers of `data` events.

### LINKED ISSUE #3578: "tiledata", "sourcedata", "styledata" events
Users can consume the `data` event if they want a callback whenever the map's data changes. Users who are listening specifically for `data[type=tile]` events are forced to jump through some ergonomically 🙅 hoops. 

```
// 🙅
map.on('data', function(event) {
    if (event.type === 'tile') {
        ...
    }
});

// ✅ 
map.on('tiledata', function(event) {
    ...
});
```

lucaswoj:
This PR makes a few enhancements to the `data` event to improve developers' understanding of the state of their sources. 

 - adds `styledata`, `sourcedata`, `styledataloading`, and `sourcedataloading` events to remove the need for this common boilerplate: `map.on('data', function(event) { if (event.dataType === …` (fixes #3577)
 - adds `isSourceLoaded` and `source` properties to `MapDataEvent` (fixes #3578)

cc @mourner @jfirebaugh @mapsam @mollymerp @lbud 

## Launch Checklist

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -1604,14 +1604,25 @@ function removeNode(node) {
  */
 
 /**
- * Fired when one of the map's sources loads or changes. See
- * [`MapDataEvent`](#MapDataEvent) for more information.
+ * Fired when one of the map's sources loads or changes. This event is not fired
+ * if a tile belonging to a source loads or changes (that is handled by
+ * `tiledata`). See [`MapDataEvent`](#MapDataEvent) for more information.
  *
  * @event sourcedata
  * @memberof Map
  * @instance
  * @property {MapDataEvent} data
  */
+
+ /**
+  * Fired when one of the map's sources' tiles loads or changes. See
+  * [`MapDataEvent`](#MapDataEvent) for more information.
+  *
+  * @event tiledata
+  * @memberof Map
+  * @instance
+  * @property {MapDataEvent} data
+  */
 
 /**
  * Fired when any map data (style, source, tile, etc) begins loading or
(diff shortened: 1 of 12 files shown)


jfirebaugh:
>  merges the `tile` `dataType`s into the `source` `dataType` because not all sources are tiled, we don’t expose tiles as domain objects to our users anywhere else

I don't think I agree with this portion of the PR. We expose tiles:

* In the `tiles` property of sources, and related style properties such as `minzoom`, `maxzoom`, GeoJSON `buffer`, and `symbol-avoid-edges`.
* In the structure of results from `query*Features`.
* Via the vector tile specification, which is integral to Mapbox GL.

This PR adds separate `styledata` and `sourcedata` events because the distinction between different types of data proved to be important. I can easily imagine cases where that distinction is just as important for tiles -- loading custom overlay data for the loaded tiles, collecting tile analytics, "drawing a shading over tiles that are "loading" from my vector tile source" (#3328).

lucaswoj:
Good points @jfirebaugh. I'll go back in and restore the distinction between tile data and source data.

lucaswoj:
cc @jfirebaugh 

Updated to restore the `tile` events, making this a non-breaking change 😄

lucaswoj:
ping @mourner @jfirebaugh @mapsam @mollymerp @lbud for a final review of this PR

js/ui/map.js
```diff
@@ -1550,26 +1556,92 @@ function removeNode(node) {
...
+ * @instance
+ * @property {MapDataEvent} data
+ */
+
+/**
+ * Fired when one of the map's sources' tiles begins loading or changing
+ * asyncronously. All `sourcedataloading` events are followed by a `sourcedata`
+ * or `error` event. See [`MapDataEvent`](#MapDataEvent) for more information.
+ *
+ * @event sourcedataloading
```
lbud: @lucaswoj is this event + description supposed to be `tiledataloading`?

lucaswoj: Yes! Fixing now.

lucaswoj: Fixed

lucaswoj:
I just refactored simplify the PR and eliminate some redundancies (namely that `setEventedParent` overwrites the `source` property)

lucaswoj:
ping @mourner @jfirebaugh @mapsam @mollymerp @lbud for a final review of this PR

kronick:
@lucaswoj I just [documented a workaround](https://github.com/mapbox/labs/wiki/Animation-Techniques#moving-geometry-with-sourcesetdata) for similar issues when using `.setData()` on a `geojson` source to create animations. Looks like this PR only covers events on the initial data load-- would it be a good idea to also revamp events on `setData()` calls?

lucaswoj:
@kronick The new events and event properties added by this PR should apply in the case of a `GeoJSONSource#setData` operation. What makes it look like these only work on the initial data load?

lucaswoj:
Thank you @mourner!

Issue #3559, opened 2016-11-07
Create a project roadmap


lucaswoj:
I recently stumbled across the [GitHub Platform Roadmap](https://developer.github.com/early-access/platform-roadmap/). 

In few words, this page communicates the GitHub platform team's priorities. 

I imagine it is effective at managing external users' expectations and focusing internal engineers' attention.

We should build something similar for GL JS / the GL ecosystem.

mapsam:
I agree! I see this lives on GitHub's developer pages. Could the GL JS equivalent live on mapbox.com/mapbox-gl-js?

I'd be happy to take this on but first would like to think about how it is kept up to date (programmatically or something manual?)

lucaswoj:
A linear roadmap may be a more useful system for us than release milestones. (Though we'll still release every 2 weeks)

lucaswoj:
Here are a couple rough requirements for whatever system we build:

 - store all roadmap state within Github issues / projects / pull requests
 - give us a framework to determine what the top `n` highest priority tasks are (perhaps sort first by priority label then second by issue creation date?)
 - don't show projects so far into the future that they are likely to be bumped out by higher priority projects before we get around to the 
 - inform external users what we're working on next
 - help GL JS engineers figure out what they should work on next

mapsam:
> linear roadmap may be a more useful system for us than release milestones

Yeah, I think starting simple is the way to go. No need to overcomplicate it right away.

I think we could add a new page to the mapbox.com docs at `mapbox.com/mapbox-gl-js/roadmap` (or an equivalent). This will give us an easy spot to point people to, plus we can link to it easily from within the docs.

A brainstorm of ways to organize roadmap content:

1. By **issue labels**: we can create new labels, `roadmap priority 1`/`roadmap priority 2`/etc to query with the GitHub search API
1. Use **github projects** exclusively for the roadmap, anything created as a project is on our roadmap. Some negatives here: there's no way to label projects, so we wouldn't be able to show priority in anything other than date
1. Create a **single "roadmap" project**, which has columns based on priority. This could remove the need for adding issue labels, as well as even a need for mapbox.com integration.

lucaswoj:
cc @isiyu, who is interested in building something similar for another project. If we end up writing any code for this that could be shared, there is interest on other teams at Mapbox.

mapsam:
Adding a new page to the docs, here's a quick example:

[...]

Issue #3580, opened 2016-11-09
"canvas" source type


lucaswoj:
Creating a custom layer API that allows developers to draw directly into the map's canvas is a huge project.

A simpler solution (and perhaps a solution more in line with our users' needs) is to create a `canvas` source type which copies the contents of an external `<canvas>` element onto the map. 

The implementation of this source type would be almost identical to the implementation of the `video` source type, which copies the contents of a `<video>` element onto the map.

We'll need to include an example of using this source type along with the implementation.

lucaswoj:
I have some time to play with this today

lucaswoj:
Quick note: performance ought to be good when using this with a 2D canvas context but not with a 3D (WebGL canvas context)

lbud:
In progress in https://github.com/mapbox/mapbox-gl-js/tree/canvas_type (using 2D canvas context).
Example: http://bl.ocks.org/lbud/a1f6e07528cc62317309c0da5f00f3a2

mourner:
@lbud wow, that looks great!

mourner:
> Quick note: performance ought to be good when using this with a 2D canvas context but not with a 3D (WebGL canvas context)

BTW, I just realized that while `readPixels` would hit hard, accepting a WebGL framebuffer texture instead should enable using WebGL here without the performance penalty. What do you think @lbud  @ansis @lucaswoj?

ansis:
>  accepting a WebGL framebuffer texture instead should enable using WebGL here without the performance penalty

This would be the performant way to do it. We'd have to expose our gl context and make sure we unset and reset all state before and after they render.

lbud:
> We'd have to expose our gl context and make sure we unset and reset all state before and after they render.

I'm not sure it's worth going down this road, at least at this point. I think the intent of the canvas source type proposal was to quickly enable users familiar with common HTML drawing techniques like HTML canvas to attach their graphics to the map, and IMO it may make most sense here to at least start with a 2D canvas context iteration. If people are really dying to attach their own WebGL framebuffers to the map, we can consider going down that road as well, but particularly at a time when we've also been talking more and more about how much we'd like to revisit custom shaders https://github.com/mapbox/mapbox-gl-js/issues/281 — AND before we nail down more robust maintenance of our GL state https://github.com/mapbox/mapbox-gl-js/issues/145 — it seems to me we should leave the 3D webgl framebuffer transference on the shelf.

core44:
[...]

PR #3594, opened 2016-11-11
Public roadmap

mapsam:
This is a first step towards a beautiful, helpful, public roadmap #3559 

screen shot 2016-11-11 at 11 12 56 am

I've gone through a _very rough_ first draft of the roadmap, but my perspective is limited and it needs input from everyone else! I'll plan on manually updating this during weekly scrum. If anyone has extra features they think would be helpful, let me know and I can add them.

cc @lucaswoj @mollymerp @lbud @mourner @jfirebaugh @ansis

### DIFF
--- a/docs/_data/roadmap.yml
+++ b/docs/_data/roadmap.yml
@@ -1,0 +1,36 @@
+- 
+  term: "Active development"
+  items: 
+    -
+      name: Unify Camera API into a single method
+      issues: ["3583", "2801"]
+    -
+      name: Improving event system and schema
+      issues: ["1408", "3577", "3578"]
+    - 
+      name: Vertical text rendering improvements
+      issues: ["3438", "3470", "3402"]
+    -
+      name: Plugin architecture documentation updates
+      issues: ["1835"]
+-
+  term: "Upcoming"
+  items:
+    - 
+      name: Use Yarn for dependency installations on Circle CI instead of NPM
+      issues: ["3364", "3551"]
+    -
+      name: Improve performance around setting styles with <code>setStyle()</code>
+      issues: ["1370", "1989"]
+    -
+      name: Fix GeoJSON memory leak
+      issues: ["3220", "3595"]
+-
+  term: "Long-term"
+  items:
+    -
+      name: API refactors and breaking chnages
+      issues: ["2803", "2727", "2506", "2112", "1408"]
+    -
+      name: Custom layer types and shaders
+      issues: ["281"]
--- a/docs/_data/roadmap.yml
+++ b/docs/_data/roadmap.yml
@@ -22,12 +22,12 @@
     -
       name: Custom source type API
       project: 2
--
-  term: "Long-term"
-  items:
     -
       name: Version 1.0 and API stability
       issues: ["2803", "2727", "2506", "2112", "1408"]
     -
-      name: Custom shaders
-      issues: ["281"]
+      name: Raster source "loading" pattern and "errored" pattern
+      issues: ["3280"]
+    -
+      name: Benchmarking improvements
+      issues: ["3237", "3495", "3031", "2253", "3248"]
--- a/docs/_data/roadmap.yml
+++ b/docs/_data/roadmap.yml
@@ -5,32 +5,29 @@
       name: Unify Camera API into a single method
       issues: ["3583", "2801"]
     -
-      name: Improving event system and schema
-      issues: ["1408", "3577", "3578"]
+      name: Augmenting "data" events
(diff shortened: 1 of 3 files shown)


mapsam:
Good idea @mourner! That will keep the workflow similar to the `plugins.yml`.

mollymerp:
thank you for taking initiative on this @mapsam ! great work!

mapsam:
@mourner added a `roadmap.yml` file and simplified the layout.

lucaswoj:
I'm stoked about this 🙌 

Here are my notes about what the _contents_ of the roadmap should include: 

# Roadmap

We don’t have a roadmap yet so you get some discretion in populating it. Filling in the roadmap will have to be an iterative process as different stakeholders speak up. 

## What should go on the roadmap

* Everything in “Project ⛈” 
* Everything sitting in an open PR is top priority right now
* Everything labeled “high priority” is pretty high priority
* Everything labeled “breaking change” is pretty high priority (we wanna v1 as soon as we can)

## What should not go on the roadmap

* small bug fixes
* small features (~1 days of work)
* refactoring projects

mapsam:
Per discussion with @lucaswoj I've updated the content and added the ability to link an item to a project on Github. This is ready to go whenever we're comfortable!

image

mapsam:
@lucaswoj thanks for posting those notes.

> [do not include] refactoring projects

I'm a little hesitant to completely cut these out, especially if we know they are coming. It might be helpful for people to see these changes coming on the horizon and prepare.

lucaswoj:
Fair point re: refactoring. We should handle this on a case-by-case basis.

docs/_data/roadmap.yml
```diff
@@ -0,0 +1,33 @@
...
+      name: Custom source type API
+      project: 2
+-
+  term: "Long-term"
+  items:
+    -
+      name: Version 1.0 and API stability
+      issues: ["2803", "2727", "2506", "2112", "1408"]
+    -
+      name: Custom shaders
```
lucaswoj: Let's bump this off the roadmap while we see how #3580 pans out

docs/_data/roadmap.yml
```diff
@@ -0,0 +1,33 @@
...
+      name: Improve performance around setting styles with <code>setStyle()</code>
+      issues: ["1370", "1989"]
+    -
+      name: Fix GeoJSON memory leak
+      issues: ["3220", "3595"]
+    -
+      name: Custom source type API
+      project: 2
+-
+  term: "Long-term"
```
lucaswoj: I'm unsure how to determine 

 - if a particular issue should be in `upcoming` or `long term`
 - when a particular issue should be moved from `long term` to `upcoming`

Do you have thoughts on how we can make these ☝️ determinations?  Do you think it'd be easier to manage if we had only two pools of issues: `active development` and `upcoming`?

docs/_data/roadmap.yml
```diff
@@ -0,0 +1,33 @@
+-
```
lucaswoj: We could put some procedural notes at the top of this file in a comment. Some points we should hit:

 - This roadmap is updated every week during the Tuesday scrum
 - This roadmap should _not_ include...
  * small bug fixes
  * small features (~1 day of work)
 - This roadmap should not have more than N items in the "Upcoming" section at any time (10?)

docs/_data/roadmap.yml
mapsam: I think it's a great idea to just have two columns to start with. We can add a third if it feels necessary during Tuesday scrum.

docs/_data/roadmap.yml
mapsam: 👍

mapsam: I removed "custom shaders" ... do we want #3580 added @lucaswoj?

docs/_data/roadmap.yml
```diff
@@ -0,0 +1,33 @@
...
+      project: 2
+-
+  term: "Long-term"
+  items:
+    -
+      name: Version 1.0 and API stability
+      issues: ["2803", "2727", "2506", "2112", "1408"]
+    -
+      name: Custom shaders
+      issues: ["281"]
```
mapsam: Added!

mapsam:
Thanks for the comments @lucaswoj - I've updated the page to be two columns, added some additional procedural text, and updated the roadmap items. Screenshot:

screen shot 2016-11-15 at 8 21 04 am

mapsam:
Wahoo!

Issue #3595, opened 2016-11-11
GeoJSONWorkerSource leaks memory: it never deletes old WorkerTiles from removed sources/layers


Que3216:
**mapbox-gl-js version**: v0.26.0

### Steps to Trigger Behavior

 1. Open this jsfiddle: http://jsfiddle.net/h93hgmko/1/. The fiddle repeatedly adds and removes a layer and source, each time with a different id.
 2. Take repeated heap snapshots of one of the workers, and you'll see the retained memory usage gradually increase to an unbounded size

### Expected Behavior

Since the amount of data on the map never increases, the retained memory usage of the worker should hit a ceiling.

### Actual Behavior

Retained memory usage of the worker increases unboundedly:

![2a5fd2e3-fbb5-4362-9d40-ed61be855236](https://cloud.githubusercontent.com/assets/793481/20227727/7af46db8-a802-11e6-8fa7-23beee47fd8f.png)

If you add a breakpoint in `geojson_worker_source.js` you can see that the worker is retaining the old WorkerTiles for sources/layers which have since been deleted:

![image](https://cloud.githubusercontent.com/assets/793481/20227760/a4b2e274-a802-11e6-9d3c-64a8b216fa17.png)

It looks like the old data never gets deleted -- the 8 workers collectively have references to all of the data from each previous iteration.

mourner:
Thanks for the report! Related issues: #2607 #3220. This is definitely a priority for us — going to address those shortly.

Que3216:
Thanks for prioritising this! Do you have an estimate of when this may be fixed?

Que3216:
I tried repro-ing this on the master branch of MGL. The behaviour is slightly different, however it still leaks memory (the retained memory of a worker goes up over time). It now correctly erases old WorkerTiles in the loaded hash, however it still leaks GeoJSONVT's in the _geoJSONIndexes:

On MGL 0.26.0:

![2a1a3531-e5af-4102-b3e6-f5dbef67d2cc](https://cloud.githubusercontent.com/assets/793481/20232722/3885a3a8-a81f-11e6-97e0-da322bd79989.png)

On master:

![e0184c88-c7b8-4d5f-9814-9e95f9c5d1d0](https://cloud.githubusercontent.com/assets/793481/20232730/3e4a4b4a-a81f-11e6-8c3e-cdef00e733c2.png)

(good)

[...]

PR #3602, opened 2016-11-11
Remove source from GeoJsonWorkerSource._geoJSONIndexes when the source is removed from the map, to fix a memory leak

Que3216:
https://github.com/mapbox/mapbox-gl-js/issues/3595

### DIFF
--- a/js/source/geojson_source.js
+++ b/js/source/geojson_source.js
@@ -206,7 +206,7 @@ class GeoJSONSource extends Evented {
         this.dispatcher.send('removeTile', { uid: tile.uid, type: this.type, source: this.id }, () => {}, tile.workerID);
     }
 
-    onRemove(map) {
+    onRemove() {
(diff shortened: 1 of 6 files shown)


mourner:
This is great! Can you fix the build and possibly add a test that exercises the new code (removing a GeoJSON source)?

Que3216:
Yep I'll add this now

Que3216:
@mourner I've added a test, let me know what you think.

anandthakker:
> I don't see anything cleaning up this.workerSources[mapId] when a map is removed. Is this another leak?

@jfirebaugh yeah, this is an open leak, although my guess is that this is less urgent than the one addressed in this change, since (at least in my experience) destroying a map is much less common than removing a particular source.  Also, after this fix, the actual size of a lingering WorkerSource is minimal.  (Not that we shouldn't fix it, just pointing out that it probably isn't severe enough to be blocking.)

Issue #50, opened 2013-09-05
Wrap world


kkaefer:
If the user zooms out enough, we should start wrapping the world and then confine the internal transform coordinates (and the url hash) to -180...180 (it'll look the same if we wrap)

PR #3608, opened 2016-11-14
Refine vertical punctuation logic based on Unicode standard

1ec5:
The list of Unicode character blocks is now based on [the official Unicode 9.0 character database](http://www.unicode.org/Public/9.0.0/ucd/Blocks.txt), whereas the previous table was based on an older Unicode standard.

The logic distinguishing upright, rotated, and neutral characters has been refined based on [Unicode Technical Report #50](http://www.unicode.org/reports/tr50/proposed.html) (proposed revision 16, which corresponds with Unicode 9.0).

Here are some example effects of this PR:

* In a vertical “地圖©地圖盒子”, “©” is displayed upright instead of rotated.
* In a vertical “地……圖”, “…” is displayed as an upright “︙” instead of a rotated “…”, so it is centered horizontally and no longer runs into the following CJK character.
* In a vertical “地圖！”, “！” is displayed as an upright “︕”.

To keep things simple, the Letterlike Symbols and Number Forms blocks as a whole are considered neutrally oriented. We can refine that portion further if this simplification turns out to be problematic.

Depends on mapbox/mapbox-gl-test-suite#169 (although technically no test suite change was required because the differences were so minor).

/cc @lucaswoj @nickidlugash


js/util/is_char_in_unicode_block.js
```diff
@@ -64,6 +65,7 @@ module.exports = {
     // 'Batak': (char) => char >= 0x1BC0 && char <= 0x1BFF,
     // 'Lepcha': (char) => char >= 0x1C00 && char <= 0x1C4F,
     // 'Ol Chiki': (char) => char >= 0x1C50 && char <= 0x1C7F,
+    // 'Cyrillic Extended-C': (char) => char >= 0x1C80 && char <= 0x1C8F,
```
lucaswoj: Any idea how I might've missed this and other blocks the first time around?

js/util/script_detection.js
```diff
@@ -71,24 +102,181 @@ exports.charAllowsVerticalWritingMode = function(char, includeNeutrals) {
...
+    //if (isChar['Ornamental Dingbats'](char)) return true;
+    //if (isChar['Transport and Map Symbols'](char)) return true;
+    //if (isChar['Alchemical Symbols'](char)) return true;
+    //if (isChar['Geometric Shapes Extended'](char)) return true;
+    //if (isChar['Supplemental Symbols and Pictographs'](char)) return true;
+    //if (isChar['CJK Unified Ideographs Extension B'](char)) return true;
+    //if (isChar['CJK Unified Ideographs Extension C'](char)) return true;
+    //if (isChar['CJK Unified Ideographs Extension D'](char)) return true;
+    //if (isChar['CJK Unified Ideographs Extension E'](char)) return true;
+    //if (isChar['CJK Compatibility Ideographs Supplement'](char)) return true;
```
lucaswoj: I'd prefer if this code block were copied into https://github.com/mapbox/mapbox-gl/issues/29 than left commented out in the codebase.

js/util/is_char_in_unicode_block.js
lucaswoj: 👍

js/util/script_detection.js
1ec5: → https://github.com/mapbox/mapbox-gl/issues/29#issuecomment-260467775

Issue #3613, opened 2016-11-14
Document Control interface


scothis:
The Control object had significant changes introduced in 0.27, however, the interface is not documented. There are several links to the missing Control docs https://www.mapbox.com/mapbox-gl-js/api/#Control

tmcw:
Are you aware of the [IControl documentation](https://www.mapbox.com/mapbox-gl-js/api/#IControl)?

scothis:
Thanks @tmcw. We should updated the links in [Map#addControl](https://www.mapbox.com/mapbox-gl-js/api/#Map#addControl) and [Map#removeControl](https://www.mapbox.com/mapbox-gl-js/api/#Map#removeControl)

PR #3615, opened 2016-11-14
Update doc types from Control to IControl

scothis:
The `Control` type was renamed to `IControl` in v0.27.0, this PR updates references to the correct type.

Issue: #3613


scothis:
@jfirebaugh targeting `mb-pages` now

PR #3632, opened 2016-11-16
Fix geocoder example

tristen:
Updates the "[Set a point after Geocoder result](https://www.mapbox.com/mapbox-gl-js/example/point-from-geocoder-result/)" example to account for v2 changes.

cc @lucaswoj for review.


tmcw:
Thanks @tristen - approving since Lucas is busy today.

PR #3637, opened 2016-11-16
Fix redo placement for cached tiles

### LINKED ISSUE #3623: Tiles coming from memory cache don't have placement redone
[This conditional](https://github.com/mapbox/mapbox-gl-js/blob/f1d1bf464d589632fdad3b0c1fe10f30d822fbd7/js/source/source_cache.js#L386) is always false -- `SourceCache` has no `_redoPlacement` method -- so tiles being drawn from the cache don't have their placement correctly redone, and may render with incorrect labels if the map has been rotated since they went into the cache.

This needs a fix (easy) and a regression test (harder).

jfirebaugh:
Fixes #3623.

### DIFF
--- a/js/render/frame_history.js
+++ b/js/render/frame_history.js
@@ -8,14 +8,11 @@ class FrameHistory {
         this.opacities = new Uint8ClampedArray(256);
         this.array = new Uint8Array(this.opacities.buffer);
 
-        this.fadeDuration = 300;
         this.previousZoom = 0;
         this.firstFrame = true;
     }
 
-    record(zoom) {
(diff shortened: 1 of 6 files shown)


mourner:
The new test-suite test failed.

anandthakker:
Oh shoot -- @jfirebaugh just realized, I think the -test-suite update here is targeting a commit that will become hard to find in the tree because it's on a branch that will be squahsed/merged into master.

jfirebaugh:
Yes, I'll merge in test-suite and update the SHA before merging here.

PR #3643, opened 2016-11-17
Remove paint class support

anandthakker:
Prerequisites:
 - [x] Merge upstream PR #3621 and then rebase here
 - [x] `-test-suite`: merge https://github.com/mapbox/mapbox-gl-test-suite/pull/178 (Ignore paint class tests in gl-js) 
 - [x] `-style-spec`: merge https://github.com/mapbox/mapbox-gl-style-spec/pull/576 (update docs and spec)
 - [x] Fix bug wherein unsetting and resetting fill-outline-color causes TypeError to be thrown https://github.com/mapbox/mapbox-gl-js/issues/3657
 - [ ] Add setStyle examples to the docs #3660 

**UPDATE:** Removed https://github.com/mapbox/mapbox-gl-js/pull/3662#issuecomment-273650393 from the prerequisites, in light of `setSprite` being (a) orthogonal to paint classes, and (b) a pretty complicated / significant lift.

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -186,7 +186,7 @@ class Style extends Evented {
         this._loaded = true;
         this.stylesheet = json;
 
-        this.updateClasses();
+        this.updatePaintProperties();
 
         for (const id in json.sources) {
             this.addSource(id, json.sources[id], {validate: false});
@@ -274,10 +274,9 @@ class Style extends Evented {
         return ids.map((id) => this._layers[id].serialize());
     }
 
-    _applyClasses(classes?: Array<string>, options: ?{}) {
+    _applyPaintPropertyUpdates(options: ?{}) {
         if (!this._loaded) return;
 
-        classes = classes || [];
         options = options || {transition: true};
         const transition = this.stylesheet.transition || {};
 
@@ -288,10 +287,10 @@ class Style extends Evented {
             const props = this._updatedPaintProps[id];
 
             if (this._updatedAllPaintProps || props.all) {
-                layer.updatePaintTransitions(classes, options, transition, this.animationLoop, this.zoomHistory);
+                layer.updatePaintTransitions(options, transition, this.animationLoop, this.zoomHistory);
             } else {
                 for (const paintName in props) {
-                    this._layers[id].updatePaintTransition(paintName, classes, options, transition, this.animationLoop, this.zoomHistory);
+                    this._layers[id].updatePaintTransition(paintName, options, transition, this.animationLoop, this.zoomHistory);
                 }
             }
         }
@@ -360,7 +359,7 @@ class Style extends Evented {
     /**
      * Apply queued style updates in a batch
      */
-    update(classes: Array<string>, options: ?{}) {
+    update(options: ?{}) {
         if (!this._changed) return;
 
         const updatedIds = Object.keys(this._updatedLayers);
@@ -379,7 +378,7 @@ class Style extends Evented {
             }
         }
 
-        this._applyClasses(classes, options);
+        this._applyPaintPropertyUpdates(options);
         this._resetUpdates();
 
         this.fire('data', {dataType: 'style'});
@@ -602,7 +601,7 @@ class Style extends Evented {
             this._updatedSymbolOrder = true;
         }
 
-        this.updateClasses(id);
+        this.updatePaintProperties(id);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


anandthakker:
Hmm... what is the right thing to do in `-style-spec` alongside this change?

Issue: `paint.*` is referenced in the [spec definition](https://github.com/mapbox/mapbox-gl-style-spec/blob/0079101a7336e5df88e76b550eb23dd90a002985/reference/v8.json#L338-L341) and the [docs](https://github.com/mapbox/mapbox-gl-style-spec/blob/0079101a7336e5df88e76b550eb23dd90a002985/docs/_generate/index.html#L631-L633)

Options:
1. Remove from spec & docs, bump to v9
2. Leave in spec--with deprecation notice--and update docs to point to [declass](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/declass.js) for usage of `paint.*`.
3. Something else?

anandthakker:
I guess now that I've written that, I'm leaning towards option 2

anandthakker:
Also TODO: remove `classes` tests from the test-suite.  Updating top of ticket to reflect this and the -style-spec change.

anandthakker:
Rebased off of master instead of being branched off an open PR

anandthakker:
Rebased onto master to fixup minor merge conflict

anandthakker:
@jfirebaugh @lucaswoj updated top of ticket w/ the prerequisites we discussed.

averas:
I use paint classes quite extensively and just stumbled across this. Do you think it's justified to mention in the documentation that paint class support will be deprecated so that we who use it start migrating away from it and new users stay away from it?

lucaswoj:
Absolutely @averas. I'll do that ASAP.

anandthakker:
Deprecation flagged here: https://github.com/mapbox/mapbox-gl-js/pull/4038

lucaswoj:
I support considering adding `Map#setStyle` examples to the docs (#3660) to be a prerequisite for merging this PR.

Doing so will give us more confidence that our proposed alternative is sound, give our users more resources to make the transition, and give support more materials to reference.

averas:
I am currently migrating away from paint classes in my own applications so I thought I'd share the challenges that I've hit with the new smart <code>setStyle()</code> (which by the way has tons of other benefits):

#### Keeping GeoJSON sources in sync when the user switches style
I would say that it's fairly common that you have "base layers" that constitute the backdrop of your map, and of top of that you have some "custom layers" containing data that you visualise on your map. Sometimes different backdrops, such as with higher contrast, or with a satellite raster fits better, and a common use case is allowing the user to flip the backdrop while retaining the custom layers.

I used to fulfil this use case quite straightforwardly using paint classes. Any mutations to my "custom layers" would only require that I updated a single set of sources which allowed me to easily implement tons of filtering etc.

With smart <code>setStyle()</code> this is not as straight forward, as you need to manually keep your sources in sync, either by copying the sources from the style you're about to replace, to the style you're replacing it with, or by maintaining the "single-source-of-truth" outside of the style sources.

Except that copying sources is tedious I think it may become a performance issue as well. It isn't currently so for me, but if you copy a 70-80MB GeoJSON back and fourth between styles it may become one.

#### It's not just sources, sometimes you have custom layers as well
Some of the layers I have in my applications have their paint properties set dynamically based on user interaction (for example the color stops of a fill). This requires me to copy such layers much like with the sources described above. When I had paint classes I simply let such multi-style-layers have paint properties defined without classes (so they were active regardless of class).

anandthakker:
Thanks for this helpful feedback, @averas !

> Keeping GeoJSON sources in sync when the user switches style

> Except that copying sources is tedious I think it may become a performance issue as well. It isn't currently so for me, but if you copy a 70-80MB GeoJSON back and fourth between styles it may become one.

Interesting point -- I agree that this is an issue that isn't well addressed by the current `setStyle` API.  I'll file a separate ticket.

anandthakker:
Followed up on the GeoJSON source syncing issue here: https://github.com/mapbox/mapbox-gl-js/issues/4006#issuecomment-275811568

indus:
The deprecation of paint classes hit me hard and for me the new setStyle is not a sufficient replacement when you deal with toggleable layers and realtime data at the same time.

anandthakker:
Sorry to hear that, @indus.  Our intention was for "smart setStyle" to be sufficient (thus our waiting until it supported changes to `sprite:`, for example);  if there are specific use cases where it's not, please do open a new issue!

Issue #3649, opened 2016-11-17
Make use of moveLayer in `Style#setState`


anandthakker:
Follow-up to #3621 -- the initial implementation of `Style#setState` does not make use of the recently-added `Style#moveLayer` method.

I think the ideal way to do this is probably to introduce a new `moveLayer` operation in the -style-spec diff algorithm.

anandthakker:
Dupe of #4142

PR #3655, opened 2016-11-18
Clear source tiles for removed/re-added layer

### LINKED ISSUE #3633: Removing a symbol layer and immediately readding it as a circle layer causes TypeError in draw_circle
**mapbox-gl-js version**: 0.27

### Steps to Trigger Behavior

 1. Create a map with a symbol layer with id 'a'
 2. `map.removeLayer('a'); map.addLayer({ id: 'a', type: 'circle', source: (same as the symbol layer) })`

### Expected Behavior
There are no errors

### Actual Behavior
```
TypeError: Cannot read property 'circle' of undefined
    at Object.drawCircles [as circle] (/Users/anand/c/mapbox/mapbox-gl-js/js/render/draw_circle.js:27:44)
    at Painter.renderLayer (/Users/anand/c/mapbox/mapbox-gl-js/js/render/painter.js:271:25)
    at Painter.renderPass (/Users/anand/c/mapbox/mapbox-gl-js/js/render/painter.js:254:18)
    at Painter.render (/Users/anand/c/mapbox/mapbox-gl-js/js/render/painter.js:213:14)
    at Map._render (/Users/anand/c/mapbox/mapbox-gl-js/js/ui/map.js:1141:22)
    at Immediate.immediate._onImmediate (timers.js:585:18)
    at tryOnImmediate (timers.js:543:15)
    at processImmediate [as _immediateCallback] (timers.js:523:5)
```

I think the problem is that, since this is happening in a single batch, the it's being treated by the source cache as an update--specifically, the tiles' states are set to `reloading`, so that they are still considered renderable.  In the render pass that happens _after_ the layer's been re-added but _before_ the tile reloading is complete, we end up trying to render them as a circle-layer tiles, but the buffers aren't set up for that yet.

anandthakker:
Fixes #3633 by distinguishing the cases where the source tiles need to
be _reloaded_ from those where they need to be _cleared_

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const assert = require('assert');
 const Evented = require('../util/evented');
 const StyleLayer = require('./style_layer');
 const ImageSprite = require('./image_sprite');
@@ -252,7 +253,13 @@ class Style extends Evented {
             this._updateWorkerLayers(updatedIds, removedIds);
         }
         for (const id in this._updatedSources) {
-            this._reloadSource(id);
+            const action = this._updatedSources[id];
+            assert(action === 'reload' || action === 'clear');
+            if (action === 'reload') {
+                this._reloadSource(id);
+            } else if (action === 'clear') {
+                this._clearSource(id);
+            }
         }
 
         this._applyClasses(classes, options);
@@ -362,7 +369,16 @@ class Style extends Evented {
 
         this._layers[id] = layer;
 
-        delete this._removedLayers[id];
+        if (this._removedLayers[id]) {
+            // If, in the current batch, we have already removed this layer
+            // and we are now re-adding it, then we need to clear (rather
+            // than just reload) the underyling source's tiles.
+            // Otherwise, tiles marked 'reloading' will have buffers that are
+            // set up for the _previous_ version of this layer, confusing
+            // https://github.com/mapbox/mapbox-gl-js/issues/3633
+            delete this._removedLayers[id];
+            this._updatedSources[layer.source] = 'clear';
+        }
         this._updateLayer(layer);
 
         if (layer.type === 'symbol') {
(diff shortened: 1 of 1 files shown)


anandthakker:
Needs https://github.com/mapbox/mapbox-gl-test-suite/pull/179 to be merged and then the test-suite dependency updated here.

#monorepo

js/style/style.js
```diff
@@ -362,7 +369,16 @@ class Style extends Evented {
...
-        delete this._removedLayers[id];
+        if (this._removedLayers[id]) {
+            // If, in the current batch, we have already removed this layer
+            // and we are now re-adding it, then we need to clear (rather
+            // than just reload) the underyling source's tiles.
+            // Otherwise, tiles marked 'reloading' will have buffers that are
+            // set up for the _previous_ version of this layer, confusing
+            // https://github.com/mapbox/mapbox-gl-js/issues/3633
+            delete this._removedLayers[id];
+            this._updatedSources[layer.source] = 'clear';
```
anandthakker: I think is going to introduce a flash in cases where `remove/add` was being used to re-order the layers.

jfirebaugh:
cc @mourner

anandthakker:
^ also re: the 'flash' thing -- I think `moveLayer`, especially coupled with #3649 , should make this a nonissue.

js/style/style.js
lucaswoj: better than introducing corrupt buffers!

lucaswoj: Will these tiles still be marked as `reloading` within `SourceCache`? [As it is defined now, `reloading` means that there is an asynchronous update underway and the tiles are still renderable](https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/source_cache.js#L139-L142).

Is there a need for another state that means there is an asynchronous update underway and the tiles are NOT renderable? Or is there a need to redefine `reloading`?

Issue #2059, opened 2016-02-04
Add Map#{add,remove}Image


scothis:
In Mapbox Studio, keeping the sprite sheet in sync between the Studio UI, mapbox.com and the map is tricky and has been a consistent source of subtile bugs. Since we have the sprite json and png within Studio, we'd like a way to manually manage the sprite sheet in mapbox-gl-js.

`map.setSprite(spriteJson, springPNG)`

The json property is the metadata object, The png property could either be a data URI or a Blob.

This method should be supported in the style_batch API as well.

This API proposal would overload the API proposal in #2058, it has two args instead of one.

jfirebaugh:
I don't think we should support this; it would make it impossible to implement a 100% fidelity [`getStyle()`](https://github.com/mapbox/mapbox-gl-js/issues/1982) method.

scothis:
Would you be ok with leaving this functionality undocumented, or made as a private API? It would indeed break compatibility with `getStyle()`.

lucaswoj:
We could have both if we allow embedded base64 encoded sprites in the stylesheet https://github.com/mapbox/mapbox-gl-style-spec/issues/220

iH8:
Endusers myself included are really having trouble setting/creating spritesheets. But i gather from this answer that there's a way to do so in Studio? I'de really like it if somebody would take a look at this question: http://stackoverflow.com/questions/35565968/how-can-i-add-markers-to-the-mapbox-light-map-style That's not the first question regarding sprites and will certainly not the last. I hope there's a good solution so i can help these people. Sorry if this is not the place, feel free to delete this comment. Thanks in advance.

dcervelli:
We have a hacky workaround for this (#2277) that uses stringified JSON in the ImageSprite base to essentially implement a `map.setStyle(spriteJson, spritePng)` call. We create the png programmatically and send it over as a base64 data:/image/png URL. It serves our needs for now but ideally we'd being used a supported path. Is there any consensus around mapbox/mapbox-gl-style-spec#220 or another way of doing this?

AndyMoreland:
[...]

PR #3662, opened 2016-11-18
Add style#setSprite method

### LINKED ISSUE #2058: Add `map.setSprite(spriteUrl)`method

anandthakker:
⚠️ *note the base branch* of this PR is `smart-set-style` (https://github.com/mapbox/mapbox-gl-js/pull/3621), not `master` ⚠️ 

Closes #2058 

This turned out to be simpler than I thought (unless I'm missing something).  Render tests, including the new set-style runtime styling tests, still pass, and the light/dark setStyle debug page works, now with changing `sprite` URIs.

### DIFF
--- a/debug/dark-v9.js
+++ b/debug/dark-v9.js
@@ -85,7 +85,7 @@ window.darkv9 = {
       "type": "vector"
     }
   },
-  "sprite": "mapbox://sprites/mapbox/light-v9",
+  "sprite": "mapbox://sprites/mapbox/dark-v9",
   "glyphs": "mapbox://fonts/mapbox/{fontstack}/{range}.pbf",
   "layers": [
     {
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -34,9 +34,9 @@ const supportedDiffOperations = util.pick(diff.operations, [
     'removeSource',
     'setLayerZoomRange',
     'setLight',
-    'setTransition'
+    'setTransition',
+    'setSprite'
     // 'setGlyphs',
-    // 'setSprite',
 ]);
 
 const ignoredDiffOperations = util.pick(diff.operations, [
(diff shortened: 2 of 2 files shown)


anandthakker:
👋  @lucaswoj @jfirebaugh just flagging that this is a proposed amendment to #3621, which we talked about merging tomorrow.

jfirebaugh:
I'm seeing some visual artifacting when there are icons on screen while the transition occurs. I suspect there's a race condition where, prior to the tiles finishing their reload, buffers that contain texture coordinates relative to the prior sprite are drawn with the newer sprite.

anandthakker:
@jfirebaugh what do you think is the ideal behavior here?  Should we essentially track/detect when a tile's got stale spritesheet data, and bail out of rendering the offending layer?  Or do we want to try something more ambitious, like cross-fading between the old and new sprites?

jfirebaugh:
It's a hard problem, and I don't see any easy solution. The sprite texture is essentially a shared resource that safe to delete only when there are no buffers containing texture coordinates that refer to it. It's similar to why [updating a single image in the sprite is hard](https://github.com/mapbox/mapbox-gl-native/issues/7618).

jfirebaugh:
I think the solution here has to be something like: treat the texture itself as a persistent resource -- never delete it. Instead, reference count individual images within it, like we do for glyphs. We'll need to add a generation identifier to the user-specified image identifier, to allow `setSprite` to overwrite previous images, while still retaining the previous image until tiles that use it have been re-parsed.

anandthakker:
Closing here, since the issue it targets ( #2058 ) is closed in favor of #2059

PR #3667, opened 2016-11-21
Draw CDN 0.16.0

mcwhittemore:
There have been a few changes to MapboxDraw including changing its global name to MapboxDraw to match MapboxDirections and MapboxGeocoder.

@mollymerp - can I run plugin deploy before I merge this? I'm trying to find a less hacky way to test the demo than rewriting it to point at a local copy.


mollymerp:
Yep, I'd run the deploy script and then test to make sure the example is working!

mcwhittemore:
Cool. Plugins deployed and the example works. Merging.

PR #3672, opened 2016-11-22
Add support for circle-stroke-* properties

### LINKED ISSUE #1713: "circle-stroke-*" style properties
Currently, we can only style geoJSON points using the circle layer type but that doesn't give us the flexibility to style circles with a border. We can get close with two circle layers with different radii but the effect isn't ideal when they overlay.

screen shot 2015-11-12 at 4 02 51 pm

lucaswoj:
actual 5 14 06 pm

This PR adds support for `circle-stroke-*` properties:

 - `circle-stroke-color`
 - `circle-stroke-width`
 - `circle-stroke-opacity`

All of these support data-driven styling out of the box. 

fixes #1713 
ref https://github.com/mapbox/mapbox-gl-test-suite/pull/182
ref https://github.com/mapbox/mapbox-gl-style-spec/pull/579
ref https://github.com/mapbox/mapbox-gl-shaders/pull/39

cc @jfirebaugh @mourner @lbud 

## Launch Checklist

 - [ ] post benchmark scores


1ec5:
Don’t forget to run `make style-code && make style-code-android && make style-code-darwin` to update the various generated files.

mollymerp:
yesss so excited for this!

lucaswoj:
This is ready for 👁  @jfirebaugh @mourner @lbud

jfirebaugh:
Is there a defined z-order for circles within a layer? It didn't matter before because there was no way to visually distinguish the order. But now there is.

lucaswoj:
> Is there a defined z-order for circles within a layer? It didn't matter before because there was no way to visually distinguish the order. But now there is.

We render circles, like other non-symbol layers, in the order specified by the source. `geojson` and `vector` sources provide an ordered set of features.

Because ordering is more important in symbol layers, we sort them by "y coordinate" then "x coordinate." This would be possible to do for circles layers. 

It is possible to visually distinguish the order of patterned fills but I'm not aware of any ordering guarantees for that layer type.

lucaswoj:
Updated, squashed, and ready to 🚢  on 🍵 .

Issue #3663, opened 2016-11-18
SymbolBucket.prototype.isEmpty throws TypeErrors on startup


kristfal:
**mapbox-gl-js version**: Master

SymbolBucket randomly (roughly 1 out of 10 times) throws errors on startup. It is probably a race condition as I've only seen it happen on slower devices (iPhone 5s) which _might_ be triggered by a map resize during map initialization. Everything seems to be working fine when the map is running. 

Error:

```
TypeError: undefined is not an object (evaluating 'this.arrays.icon')
```

Offending code:

```
SymbolBucket.prototype.isEmpty = function isEmpty () {
    return this.arrays.icon.isEmpty() &&
        this.arrays.glyph.isEmpty() &&
        this.arrays.collisionBox.isEmpty();
};
```

The safari inspector is kind enough to not provide stack traces for this error 😢 , I'll keep looking for a proper stack trace.

### Expected Behavior

Errors should not be thrown.

### Actual Behavior

Errors are thrown.

elliotleelewis:
I get this error a lot since updating I believe to v0.27.0, happens in quick succession too. For instance, I'll get 10 or 15 errors exactly like that at once. I get this error on all browsers on my 2015 MacBook Pro so don't think its to do with device speed, at least for me.

ggho:
I am encountering a similar issue updating from v25.0.1 to v0.27.0 or v0.28.0.

In my case, this error got thrown out after Map#setFilter was called and it does not happen if I don't call Map#setFilter. 

For instance, I have a mousemove event listener which handle hover effect on a symbol layer (which would call Map#setFilter when a symbol is being hovered). After a few calls of Map#setFilter (approximately 4 to 5 calls), this error would be thrown consecutively a couple of times (from 4 - 1x times randomly). 

Error captured in Chrome:
`blob:http://localhost:9001/fcaf0a79-234f-4812-bc7d-237aabb2f995:3309 
Uncaught TypeError: Cannot read property 'icon' of undefined
isEmpty @ blob:http://localhost:9001/fcaf0a79-234f-4812-bc7d-237aabb2f995:3309
(anonymous function) @ blob:http://localhost:9001/fcaf0a79-234f-4812-bc7d-237aabb2f995:4872
serializeBuckets @ blob:http://localhost:9001/fcaf0a79-234f-4812-bc7d-237aabb2f995:4871
[...]

PR #3681, opened 2016-11-22
Fix race condition causing SymbolBucket isEmpty errors

lbud:
~~Fixes https://github.com/mapbox/mapbox-gl-js/issues/3663 which was introduced after recent optimizations.~~

~~Unfortunately since was caused by a race condition I don't think we could create a useful regression test…~~

_edited to summarize the below discussion + better reflect what this PR does:_

Fixes https://github.com/mapbox/mapbox-gl-js/issues/3663:
* There was a race condition wherein two calls to `VectorTileWorkerSource#reloadTile` would interleave calls to `parse` for the same tile, which would cause its SymbolBucket members to reset themselves while assumed to be complete for the purpose of transference. #3663 then spewed errors because in serializing buckets, some symbol buckets didn't have an `arrays` member with its array groups.
  * To fix, `VectorTileWorkerSource#reloadTile` now checks to see if its WorkerTile is still parsing (using the existing `WorkerTile.status`), if so, defers its next parse until it receives a response from the first: [vector_tile_worker_source.js#L89-L102](https://github.com/mapbox/mapbox-gl-js/blob/e3c849aa8e02749bf259657a5eafcdec9b376d8f/js/source/vector_tile_worker_source.js#L89-L102)
* This was masking a second race condition, which is also fixed in this PR: calls to `WorkerTile.redoPlacement` got deferred if a previous `parse` call to the same tile wasn't finished (with `WorkerTile.redoPlacementAfterDone`); and then were called in `WorkerTile#parse`'s `done` function, after receiving glyph/icon assets and re-parsing symbol buckets but _before_ serializing buckets to return via the `parse` callback. This meant the subset of buckets that were re-placed in `redoPlacement` — all symbol buckets — were already serialized and tripping assertion errors [here](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/util/struct_array.js#L88)
  * _However_, the only times we would ever interleave `WorkerTile#parse` and `WorkerTile#redoPlacement` would be when `WorkerTile#parse` is waiting on glyph/icon assets from the main thread. `WorkerTile#parse` only places symbols _after_ this, meaning we'd never need to re-call a deferred `redoPlacement` after this as it would return the same result. To fix, I removed the `redoPlacementAfterDone` concept altogether here; those interleaved `redoPlacement` calls set `this.pitch` and then return.

### DIFF
--- a/js/source/worker_tile.js
+++ b/js/source/worker_tile.js
@@ -140,14 +140,10 @@ class WorkerTile {
             deps++;
             if (deps === 2) {
                 for (const bucket of this.symbolBuckets) {
-                    // Layers are shared and may have been used by a WorkerTile with a different zoom.
-                    for (const layer of bucket.layers) {
-                        layer.recalculate(this.zoom);
-                    }
-
                     bucket.prepare(stacks, icons);
-                    bucket.place(collisionTile, this.showCollisionBoxes);
+                    placeSymbols(bucket, collisionTile, this.showCollisionBoxes, this.zoom);
                 }
+
                 done();
             }
         };
@@ -180,12 +176,7 @@ class WorkerTile {
         const collisionTile = new CollisionTile(angle, pitch, this.collisionBoxArray);
 
         for (const bucket of this.symbolBuckets) {
-            // Layers are shared and may have been used by a WorkerTile with a different zoom.
-            for (const layer of bucket.layers) {
-                layer.recalculate(this.zoom);
-            }
-
-            bucket.place(collisionTile, showCollisionBoxes);
+            placeSymbols(bucket, collisionTile, showCollisionBoxes, this.zoom);
         }
 
         const transferables = [];
@@ -204,5 +195,15 @@ function serializeBuckets(buckets, transferables) {
         .filter((b) => !b.isEmpty())
         .map((b) => b.serialize(transferables));
 }
+
+function placeSymbols(bucket, collisionTile, showCollisionBoxes, zoom) {
+    // Layers are shared and may have been used by a WorkerTile with a different zoom.
+    for (const layer of bucket.layers) {
+        layer.recalculate(zoom);
+    }
+
+    bucket.place(collisionTile, showCollisionBoxes);
+}
+
 
 module.exports = WorkerTile;
--- a/js/source/worker_tile.js
+++ b/js/source/worker_tile.js
@@ -107,22 +107,19 @@ class WorkerTile {
             this.status = 'done';
 
             const transferables = [];
-            serializeBuckets(util.values(buckets), transferables, (buckets) => {
-                callback(null, {
-                    buckets: buckets,
-                    featureIndex: featureIndex.serialize(transferables),
-                    collisionTile: collisionTile.serialize(transferables),
-                    collisionBoxArray: this.collisionBoxArray.serialize(),
-                    symbolInstancesArray: this.symbolInstancesArray.serialize(),
-                    symbolQuadsArray: this.symbolQuadsArray.serialize()
-                }, transferables);
+            callback(null, {
+                buckets: serializeBuckets(util.values(buckets), transferables),
+                featureIndex: featureIndex.serialize(transferables),
+                collisionTile: collisionTile.serialize(transferables),
+                collisionBoxArray: this.collisionBoxArray.serialize(),
+                symbolInstancesArray: this.symbolInstancesArray.serialize(),
+                symbolQuadsArray: this.symbolQuadsArray.serialize()
+            }, transferables);
 
-                if (this.redoPlacementAfterDone) {
-                    this.redoPlacement(this.angle, this.pitch, null, () => {});
-                    this.redoPlacementAfterDone = false;
-                }
-            });
-
+            if (this.redoPlacementAfterDone) {
+                this.redoPlacement(this.angle, this.pitch, null, () => {});
+                this.redoPlacementAfterDone = false;
+            }
         };
 
         // Symbol buckets must be placed in reverse order.
@@ -178,11 +175,11 @@ class WorkerTile {
         }
     }
 
-    redoPlacement(angle, pitch, showCollisionBoxes, callback) {
+    redoPlacement(angle, pitch, showCollisionBoxes) {
         if (this.status !== 'done') {
             this.redoPlacementAfterDone = true;
             this.angle = angle;
-            callback({});
+            return {};
         }
 
         const collisionTile = new CollisionTile(angle, pitch, this.collisionBoxArray);
@@ -197,23 +194,20 @@ class WorkerTile {
         }
 
         const transferables = [];
-
-        serializeBuckets(this.symbolBuckets, transferables, (buckets) => {
-            callback({
-                result: {
-                    buckets: buckets,
-                    collisionTile: collisionTile.serialize(transferables)
-                },
-                transferables: transferables
-            });
-        });
+        return {
+            result: {
+                buckets: serializeBuckets(this.symbolBuckets, transferables),
+                collisionTile: collisionTile.serialize(transferables)
+            },
+            transferables: transferables
+        };
     }
 }
 
-function serializeBuckets(buckets, transferables, callback) {
-    callback(buckets
+function serializeBuckets(buckets, transferables) {
+    return buckets
         .filter((b) => !b.isEmpty())
-        .map((b) => b.serialize(transferables)));
+        .map((b) => b.serialize(transferables));
 }
 
 module.exports = WorkerTile;
--- a/js/source/worker_tile.js
+++ b/js/source/worker_tile.js
@@ -34,10 +34,9 @@ class WorkerTile {
         this.collisionBoxArray = new CollisionBoxArray();
         this.symbolInstancesArray = new SymbolInstancesArray();
         this.symbolQuadsArray = new SymbolQuadsArray();
-        const collisionTile = new CollisionTile(this.angle, this.pitch, this.collisionBoxArray);
         const sourceLayerCoder = new DictionaryCoder(Object.keys(data.layers).sort());
 
-        const featureIndex = new FeatureIndex(this.coord, this.overscaling, collisionTile, data.layers);
+        const featureIndex = new FeatureIndex(this.coord, this.overscaling);
         featureIndex.bucketLayerIDs = {};
 
         const buckets = {};
@@ -103,6 +102,8 @@ class WorkerTile {
             }
         }
 
+        const collisionTile = new CollisionTile(this.angle, this.pitch, this.collisionBoxArray);
+
         const done = () => {
             this.status = 'done';
 
@@ -115,11 +116,6 @@ class WorkerTile {
                 symbolInstancesArray: this.symbolInstancesArray.serialize(),
                 symbolQuadsArray: this.symbolQuadsArray.serialize()
             }, transferables);
-
-            if (this.redoPlacementAfterDone) {
-                this.redoPlacement(this.angle, this.pitch, null, () => {});
-                this.redoPlacementAfterDone = false;
-            }
         };
 
         // Symbol buckets must be placed in reverse order.
@@ -177,7 +173,6 @@ class WorkerTile {
 
     redoPlacement(angle, pitch, showCollisionBoxes) {
         if (this.status !== 'done') {
-            this.redoPlacementAfterDone = true;
             this.angle = angle;
             return {};
         }
--- a/js/source/worker_tile.js
+++ b/js/source/worker_tile.js
@@ -102,9 +102,10 @@ class WorkerTile {
             }
         }
 
-        const collisionTile = new CollisionTile(this.angle, this.pitch, this.collisionBoxArray);
 
         const done = () => {
+            const collisionTile = this.placeSymbols(this.showCollisionBoxes);
+
             this.status = 'done';
 
             const transferables = [];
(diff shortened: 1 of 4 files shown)


jfirebaugh:
Can you describe what the race was?

lbud:
@jfirebaugh sure — here's what I gather:

* When a `WorkerTile` is responsible for collecting icons and/or glyphs, it waits on callbacks from the main thread to indicate that it has the glyph stacks/sprites it needs [(here)](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/source/worker_tile.js#L159-L175)
* Only once it has received two done messages (either from real asset dependency callbacks or, if the tile doesn't require glyphs/icons, simple passthrough function calls), it prepares a SymbolBucket [(here)](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/source/worker_tile.js#L145-L156), 
* which is when ArrayGroups are created [(here)](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/data/bucket/symbol_bucket.js#L218-L219) based on the symbol interfaces [(here)](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/data/bucket/symbol_bucket.js#L214). 
* We don't wait on `bucket.prepare` callbacks before finishing the loop and calling the worker's final callback [(here)](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/source/worker_tile.js#L106-L123) to transfer data back to the main thread, so sometimes when serializing the buckets when checking for `isEmpty` the symbol bucket's `arrays` object was not yet created, throwing an error when searching for its individual ArrayGroup members [(here)](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/data/bucket/symbol_bucket.js#L184-L188). 

(In all other layer cases, ArrayGroups are created in the bucket's constructor unless already present [(here)](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/data/bucket.js#L35))

I guess the alternate fix here might be something like: create another nested function inside `gotDependencies` that waits on a `bucket.prepare` callback (passed through to `createArrays`) to be called `n=symbolBuckets.length` times before calling `done` to transfer data back to the main thread. I'm not familiar enough with the threads' responsibilities to be sure which is "right?"

(Also in looking into this I realized that `SymbolBucket.place` [_also_ calls createArrays](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/data/bucket/symbol_bucket.js#L450), which seems like we're unnecessarily redefining `this.arrays`. Should this be removed?)

jfirebaugh:
What do you mean by `bucket.prepare` callback?

The behavior as I understand it is:

* If there are no symbol buckets, then `done` is called immediately, without waiting on any dependencies. In this situation, the `SymbolBucket.isEmpty` error shouldn't be possible, because there aren't any symbol buckets.
* If there are symbol buckets, then we wait for all dependencies. When they are complete, then `bucket.prepare` and `bucket.place` are called for all symbol buckets before calling `done`. This should result in all symbol buckets having an `arrays` member.

So I'm still not clear on how we can wind up with symbol buckets without `arrays`. I might be missing something and it is an expected state some other way, or the bug might be elsewhere.

> (Also in looking into this I realized that SymbolBucket.place also calls createArrays, which seems like we're unnecessarily redefining this.arrays. Should this be removed?)

`SymbolBucket#place` is called on it's own in `WorkerTile#redoPlacement`, so the `createArrays` there is necessary. I think it might be safe to remove from `SymbolBucket#prepare` though.

anandthakker:
@lbud @jfirebaugh I hit a similar race/exception a few weeks ago during the WaPo project, but I had thought it was because of our hacky custom source, and so didn't report it here.  Possibly unrelated to what's happening here, but I'll describe here in case it's helpful.

Essentially, the issue I had was inconsistent state in the glyph/icon callbacks.  I _think_ was happening for us looked something like:

- Main: loadTile()
- Worker: Worker#loadTile => WorkerTile#parse => getGlyphs(..., cb1) ...
  - Main: request glyphs asset...
  - Main: loadTile() // same tile requested before first one completes
- Worker: Worker#reloadTile => WorkerTile#parse (**replaces this.symbolBuckets**) => getGlyphs(..., cb2) ...  
- Main: ... glyph asset response
  - Worker: => cb1 [prepares the buckets in the _new_ `this.symbolBuckets`](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/source/worker_tile.js#L146), but the final `done()` callback uses the original symbol buckets [from its closure](https://github.com/mapbox/mapbox-gl-js/blob/95b440dee7103b4081762a36706091034f8115d6/js/source/worker_tile.js#L116).

jfirebaugh:
@anandthakker Doesn't https://github.com/mapbox/mapbox-gl-js/blob/85097c3ec03abcf27da78c37ac3a2a1bb651c20c/js/source/vector_tile_source.js#L71-L72 prevent `reloadTile` being called for a tile that hasn't completed loading yet?

anandthakker:
@jfirebaugh Ah.  Yeah, it does... maybe it was two `reloadTile`s rather than a `loadTile`+`reloadTile`.  Even when it's reloading, there's still the noop `get{Glyphs,Icons}` requests that @lbud referred to, right?

jfirebaugh:
Yeah, looks like two `reloadTile`s could produce a race where a worker tile is interleaving multiple calls to `parse` for the same tile. If that's what's causing this issue, I think the right fix is to prevent it from happening (maybe introduce a `reloading` state for `Tile`?), rather than adding a guard to `SymbolBucket#isEmpty`.

lbud:
@jfirebaugh @anandthakker thanks for the help — I was able to demonstrate that condition where it's reloading a tile before a previous `parse` on the same tile is done. Will fix now.

lbud:
😞 thought I fixed this in https://github.com/mapbox/mapbox-gl-js/pull/3681/commits/07c9d5a674ac327bc31f10c73522f2a5e78b18c5, but upon further interaction with the map (rotating/pitching, so I suspect `redoPlacement`) I've been using to debug came upon some assertion errors from serialization that I have to assume I just introduced — still digging.

lbud:
The redoPlacement race I mentioned is unrelated to the above fix (so presumably should be replicable outside of this branch — will verify) and occurs when `workerTile.parse` and `workerTile.redoPlacement` are called in very quick succession, as follows (apologies for the verbose description, mostly for my own understanding):

* `VectorTileWorkerSource` calls `workerTile.parse` ([vector_tile_worker_source.js#L92](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/vector_tile_worker_source.js#L92))
  * The `parse` execution immediately sets `workerTile.status = "parsing"` ([worker_tile.js#L31](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/worker_tile.js#L31))
* `VectorTileWorkerSource` calls `workerTile.redoPlacement` (almost immediately after the `parse` call) ([vector_tile_worker_source.js#L180](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/vector_tile_worker_source.js#L180))
  * Since `this.status === "parsing"`, we set `this.redoPlacementAfterDone = true;`, return an empty object and defer redoing placement until the `parse` is done ([worker_tile.js#L179-L183](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/worker_tile.js#L179-L183))
* `workerTile.parse` 
  * calls `SymbolBucket.place` per symbol bucket ([worker_tile.js#L153](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/worker_tile.js#L153)), which resets its StructArrays with `SymbolBucket.createArrays` ([symbol_bucket.js#L450](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/data/bucket/symbol_bucket.js#L450))
  * sets `this.status = done` ([worker_tile.js#L107](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/worker_tile.js#L107))
  * re-calls `this.redoPlacement` now that it's done re-parsing ([worker_tile.js#L110](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/worker_tile.js#L110))
  * **starts in on** serializing all of its buckets to return via its callback ([worker_tile.js#L116](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/worker_tile.js#L116))

meanwhile, in the `redoPlacement` that's just been called again:
* calls `SymbolBucket.place` per symbol bucket ([worker_tile.js#L193](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/worker_tile.js#L193), which resets its StructArrays with `SymbolBucket.createArrays` ([symbol_bucket.js#L450](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/data/bucket/symbol_bucket.js#L450))
* **starts in on** serializing all of its buckets to return ([worker_tile.js#L199](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/source/worker_tile.js#L199))

At this point,
* `workerTile.symbolBuckets` are mid-serialization/transfer from being called in `parse`; these are now partially transferred
* `workerTile.symbolBuckets` are also starting a new serialization/transfer process from being called in `redoPlacement`, because we didn't wait to re-call the batch of `SymbolBucket.createArrays` until after the first transfer was complete
* we then trip [this assert](https://github.com/mapbox/mapbox-gl-js/blob/07c9d5a674ac327bc31f10c73522f2a5e78b18c5/js/util/struct_array.js#L88) because we're trying to newly transfer a set of partially transferred arrays

(_Presumably_ (though I haven't confirmed yet) resetting symbol buckets' arrays while `parse` was still serializing them would also cause bugs from that standpoint (serializing empty arrays where there should be symbol content))

lbud:
Confirmed that the above AssertionError can be nearly replicated in master — the `isEmpty` bug was masking it before, but with a bandaid like https://github.com/mapbox/mapbox-gl-js/pull/3681/commits/4cacc0e339e929127caaa517f7103b0e3c20f862 that has nothing to do with tile reparsing, this happens as well.

bbodien:
Hi guys, are you anywhere near getting this merged and included in a release? I'm experiencing the problems described in #3663 (I'm using `setFilter()`), so I can't upgrade past v0.26.0, and would love to benefit from the performance improvements in v0.27.0

lbud:
@bbodien yes, we are actively working on it and will include the fix in the next release.

lbud:
I think this is ready — @jfirebaugh could I bother you for a review? 🙏  OP updated to reflect issues/fixes.

jfirebaugh:
> `redoPlacement` races ahead and resets each SymbolBucket's arrays (with `SymbolBucket#place`), which can still be mid-serialization.

Can you explain this in more detail? Bucket serialization is a synchronous operation, so I'm not clear how it can interleave with `SymbolBucket#place`. (Like the main JS thread, each worker is individually "single threaded" -- it can't race with itself unless there's an explicit asynchronous operation involved, like waiting for a message to be posted.)

lbud:
@jfirebaugh per chat, I was mistaken in my diagnosis of the problem (based on the single worker thread being synchronous) but had inadvertently fixed it by moving `redoPlacement` after buckets are serialized. I rolled back the callbacks I had unnecessarily added in https://github.com/mapbox/mapbox-gl-js/pull/3681/commits/1959cce29384b77d204a64bb576b213e955f7973 and updated OP.

lbud:
@jfirebaugh capturing from voice: I 
* removed the `redoPlacementAfterDone` callback altogether (as it was both unnecessary and wouldn't have returned deferred data as intended anyway), 
* moved `WorkerTile#parse`'s `collisionTile` to be more localized to its use (as it is unnecessary in this `FeatureIndex`), and 
* factored out a bit of symbol placement code to be shared between `parse` and `redoPlacement`.

jfirebaugh:
Almost there -- the last thing I see is that the `new CollisionTile(this.angle, ...)` needs to happen later in the process (while `done` is executing), so that it picks up the most recent `this.angle` value set by `redoPlacement`. One way to do this would be to create it inside, and return it from, `placeSymbols`, and then move the call to `placeSymbols` into `done`.

lbud:
@jfirebaugh 👍 addressed in https://github.com/mapbox/mapbox-gl-js/pull/3681/commits/dbd42593d0cb60ac7084574f1d250c47c332d255 (also, looks like we were not updating `this.pitch` in `redoPlacement` for the sake of re-parsing).

js/source/worker_tile.js
```diff
@@ -144,14 +141,9 @@ class WorkerTile {
             deps++;
             if (deps === 2) {
                 for (const bucket of this.symbolBuckets) {
-                    // Layers are shared and may have been used by a WorkerTile with a different zoom.
```
jfirebaugh: This needs to happen immediately prior to `prepare` as well, since `prepare` uses layer properties. (Making this less fragile is tracked in @3479.)

Issue #3307, opened 2016-10-06
Modify Map#fitBounds to delegate to Map#setCenter when only one point is provided as bounds.


bbodien:
**mapbox-gl-js version**: 0.25.1

**Example code**: http://codepen.io/bbodien/pen/ORQVkd?editors=0110
### Steps to Trigger Behavior
1. Create a `LngLatBounds` object and `extend()` it with a single LngLat point.
2. Call `fitBounds()` on the map using the `LngLatBounds` object created in `1`
3. If you uncomment JS line 49 in the Codepen example which expands the bounding box by 0.00001 in Longitude and Lattitude, giving the bounding box some dimensions, the error is not thrown.
### Expected Behavior

Despite the fact that the bounding box is just a single point and has no dimensions (SW and NE coordinates are identical), I'd expect the map to still fit around that point. 

To explain my use case, I'm receiving geojson data for locations on a trip. These could be Lines or Points. If the geojson contains Points, the number of Points is 1..n. I'm using `fitBounds()` to handle fitting my map to the data area regardless of what data I have, and I'm experiencing this error when I have 1 Point.
### Actual Behavior

An error is thrown after calling `fitBounds()`: `Error: Invalid LngLat object: (NaN, NaN)`.

mollymerp:
Calling [`new mapboxgl.LngLatBounds()`](https://www.mapbox.com/mapbox-gl-js/api/#LngLatBounds) without arguments should fire an error, bc `sw` and `ne` are required arguments. 

For your use case I would recommend using `map.setCenter()` when you only have one point, and only initializing the `LngLatBounds` when you have two or more points.

mollymerp:
Sorry, I misread our docs! There is also a typo in the docs where the parameters are not listed as optional. 

Yes, we have stated that `new mapboxgl.LngLatBounds()` should work without arguments. We should modify `Map#fitBounds` to delegate to `Map#setCenter` for unitary `LngLatBounds` objects. 

Thanks for reporting @bbodien !!

stepankuzmin:
It seems to be working with the master branch. At least it wouldn't reproduce with 519004151fbb724439fe8ee15470d1e3a6ce89a6

lucaswoj:
Yes indeed! This was fixed by #3683

PR #3683, opened 2016-11-22
improve maxBounds handling

### LINKED ISSUE #3677: Map#setMaxZoom caps maxZoom at 20
It is possible to initialize a map using options that specify an arbitrary maxZoom (Transform's default is 22).  However, Map#setMaxZoom will refuse to later set any value above 20.

setMaxZoom with this behavior was added by #2235 and #2243, but I can't find anywhere in the issue history or commit logs where it is explicitly said that this was intentional or necessary.

I have some high resolution rasters whose tiles reach their native resolution at zooms 22-24.

**mapbox-gl-js version**: 0.28.0

### Steps to Trigger Behavior

http://jsbin.com/sozesekomi/edit?html,js,output

1. Initialize map with maxZoom greater than 20.  This works and map can zoom way in.
2. Call setMaxZoom with another value greater than 20

### Expected Behavior

Max zoom is adjusted to given value, and current zoom is clamped as necessary.

### Actual Behavior

Error thrown that maxZoom must be between minZoom and 20.

mollymerp:
remove upper limit of 20 for Map#setMaxZoom, fix map.fitBounds(bounds) for a single point
fixes #3677 #3307 
## Launch Checklist

 - [ ] post benchmark scores

cc @lucaswoj @mourner

### DIFF
--- a/js/ui/camera.js
+++ b/js/ui/camera.js
@@ -296,7 +296,8 @@ class Camera extends Evented {
      * Pans and zooms the map to contain its visible area within the specified geographical bounds.
      *
      * @memberof Map#
-     * @param {LngLatBoundsLike} bounds The bounds to fit the visible area into.
+     * @param {LngLatBoundsLike} bounds The bounds to fit the visible area into. If bounds contains
+     *     only one point, the map center transitions to that point at zoom level options.maxZoom or {@link Map#getMaxZoom}.
      * @param {Object} [options]
      * @param {boolean} [options.linear=false] If `true`, the map transitions using
      *     {@link Map#easeTo}. If `false`, the map transitions using {@link Map#flyTo}. See
@@ -316,7 +317,7 @@ class Camera extends Evented {
         options = util.extend({
             padding: 0,
             offset: [0, 0],
-            maxZoom: 22
+            maxZoom: this.getMaxZoom()
         }, options);
 
         bounds = LngLatBounds.convert(bounds);
@@ -332,7 +333,6 @@ class Camera extends Evented {
         options.center = tr.unproject(nw.add(se).div(2));
         options.zoom = Math.min(tr.scaleZoom(tr.scale * Math.min(scaleX, scaleY)), options.maxZoom);
         options.bearing = 0;
-        console.log(options);
 
         return options.linear ?
             this.easeTo(options, eventData) :
--- a/js/ui/camera.js
+++ b/js/ui/camera.js
@@ -297,7 +297,8 @@ class Camera extends Evented {
      *
      * @memberof Map#
      * @param {LngLatBoundsLike} bounds The bounds to fit the visible area into. If bounds contains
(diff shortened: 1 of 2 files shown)


js/ui/camera.js
```diff
@@ -296,7 +296,8 @@ class Camera extends Evented {
      * Pans and zooms the map to contain its visible area within the specified geographical bounds.
      *
      * @memberof Map#
-     * @param {LngLatBoundsLike} bounds The bounds to fit the visible area into.
+     * @param {LngLatBoundsLike} bounds The bounds to fit the visible area into. If bounds contains
+     *     only one point, the map center transitions to that point at zoom level options.maxZoom or {@link Map#getMaxZoom}.
```
lucaswoj: Neat! I didn't know that `Map#fitBounds` did this. Should we expand this parameter's type signature to `LngLatBoundsLike | LatLngLike`?

js/ui/camera.js
```diff
@@ -329,6 +329,7 @@ class Camera extends Evented {
             scaleX = (tr.width - options.padding * 2 - Math.abs(offset.x) * 2) / size.x,
             scaleY = (tr.height - options.padding * 2 - Math.abs(offset.y) * 2) / size.y;
 
+
```
lucaswoj: Would you mind cleaning up the two whitespace changes in this file? ☝️

js/ui/camera.js
mollymerp: It used to error on `LngLatBounds` where ne + sw were the same point, this PR fixes that because that case is technically a valid instance of [`LngLatBounds`](https://www.mapbox.com/mapbox-gl-js/api/#LngLatBounds). I could make the docs language more clear to that effect. I don't think we should accept `LngLatLike` because I don't think its a recommended use-case for `fitBounds` 😕

js/ui/camera.js
```diff
@@ -296,7 +296,9 @@ class Camera extends Evented {
      * Pans and zooms the map to contain its visible area within the specified geographical bounds.
      *
      * @memberof Map#
-     * @param {LngLatBoundsLike} bounds The bounds to fit the visible area into.
+     * @param {LngLatBoundsLike} bounds The bounds to fit the visible area into. If bounds contains
+     *     only one point (i.e. bounds.ne and bounds.sw are the same location, the map center
+     *     transitions to that point at zoom level options.maxZoom or {@link Map#getMaxZoom}.
```
lucaswoj: I understand now. `bounds.ne === bounds.sw` is an edge case that demonstrates what happens when the zoom that would best fit the bounds in the viewport is greater than `Map#getMaxZoom()`.

Is there a way we can phrase these docs to make the more general pattern evident?

Here's a rough sketch:

 > Center these bounds in the viewport and use the highest supported zoom level (up to and including `Map#getMaxZoom()`) that fits them in the viewport.

Issue #2492, opened 2016-04-26
Performance of style mutations not good enough for hover interactivity


kristfal:
### Issue

Style mutations are expensive, and can be quite slow on mobile devices. When applying _new_ style mutations _faster_ than the worker is able to process an ongoing mutation, the _new_ mutation is queued and applied after the previous one has completed. 

This happens even when the same style `property` in the same `layer` is changed multiple times such as `setFilter('foo', [==, 'bar', 5])` followed by `setFilter('foo', [==, 'bar', 3])` followed by  `setFilter('foo', [==, 'bar', 11])` and so on...

If this process is repeated too quickly for an ongoing period in a performance constrained environment (mobile devices), browsers start freezing and crashing due to memory issues.

**Proposal:** New style mutations should be debounced, and only the latest change should be applied for a given layer and property.

**mapbox-gl-js version**: `earcut final` rebase 20th of April

### Steps to Trigger Behavior

1. Add layer `foo` to the map,
2. Run the map in a performance constrained environment (We're experiencing this on a range of devices around the performance level of iPhone5, Moto G2, Samsung S3)
3. Repeat style mutations quickly, for instance ` setFilter('foo', [==, 'bar', 5])` followed by `setFilter('foo', [==, 'bar', 3])` followed by  `setFilter('foo', [==, 'bar', 11])`..

### Expected Behavior

The map should apply the latest style mutation for a given mutation type and layer, and skip any previous queued mutations for the given layer and type to prevent the buildup of a "mutation backlog".

### Actual Behavior

The map queues and applies all mutations, quickly degrading the performance of the map and if mutations are continued, the browser crashes.

![skjermbilde 2016-04-26 11 55 03](https://cloud.githubusercontent.com/assets/5638107/14814269/e8a6bf4a-0ba5-11e6-9837-1efdafda180a.png)

jfirebaugh:
This is related to automatic batching. cc @mourner

mourner:
Hmm, this is actually pretty surprising since the current code is batching style mutations to happen once per frame — you can see how that works in [style.js](https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style.js).

Can you set up a minimal JSFiddle test case that reproduces the slowdown? That would make it easier to see what's going on in your case.

lucaswoj:
@kristfal, are you asking for mutations to be deduplicated such that if 10 calls to `setFilter` are made during one batch, only the final `setFilter` call is executed?

[...]

Issue #2443, opened 2016-04-14
Improve "setFilter" performance on big GeoJSON datasets


Pipoupi:
Hi, 
I've been facing some performance issue since 0.16 and I can't upgrade because of that.
Please refer to my fiddles : 
[Map with symbols on mapbox gl js 0.15](https://jsfiddle.net/e02fx38p/1/)
[Map with symbols on mapbox gl js 0.17](https://jsfiddle.net/6u8rpc1y/3/)

Loading is longer, as popup are on mousemove. But the biggest difference is on filter function. Performance is increased if I change symbol with circle (even if it's not my goal but it's for testing) but still 0.15 is better than 0.17.
 [Map with circle on mapbox gl js 0.15](https://jsfiddle.net/b3cuwf88/1/)
[Map with circle on mapbox gl js 0.17](https://jsfiddle.net/zy43d9u3/5/)

_Loading is very long because of my huge json_
_I need to stack filters, that's why I'm using an array with a lot of empty filters_

mourner:
Profiled this example and the regression seems to be caused by the conversion to `pbf` due to the `queryRenderedFeatures` changes. cc @ansis Let's investigate further.

mourner:
To be fair, this is 34MB of point data, quite an extreme case. While I'd like to investigate performance here, have you looked at speeding up the thing e.g. with point clustering?

Pipoupi:
I used to cluster my map with openlayers before I knew Mapbox & your awesome perf :) but my product has to replace marine traffic app (with cluster view) which my client had a subscription for, and it's quite more impressive for them to see all of these vessels,it's like a show case when they have visitors.. They're like "Hey, see all these vessels in that area? Some are mine".
I'm not criticizing your new version, but it's so frustrating to look at really good new function like queryrenderedfeature and querysourcefeature and can't be able to use them ! 
Anyway, I just can't go back to a clustered map

mourner:
OK, I'll see what we can do here. We probably won't be able to fully get rid of the performance hit here, but we might be able to significantly reduce it.

andrewharvey:
I don't think this is unique to GeoJSON sources, I've noticed Mapbox Tileset sources have the same poor performance for setFilter. Do you need a minimal example of this too?

mourner:
Now that we have Feature State API, this is not a pressing issue, so I'm going to close for now.

Issue #200, opened 2013-11-25
Support "hover" and "active" styles for features


kkaefer:
When building the geometry buffer, we should store the position where we encode every individual feature so that later on, we can render individual features differently from the rest of the herd, e.g. for highlighting a feature.

jfirebaugh:
I've been thinking about this in terms of pseudo-classes. Imagine you can define a style layer like:

```
{
  "id": "pois",
  "type": "symbol",
  "paint": {
    ...
  },
  "paint:hover": {
    ...
  },
  "paint:active": {
    ...
  }
}
```

[Insert handwaving about how pseudo-class state for a given feature would be determined (mouseover for `:hover`, click for `:active`).]

mourner:
@jfirebaugh I like it. This separates styling interactive features and actual behavior (which would be defined in JS), while also being less cumbersome than defining classes.

tristen:
Very interested in this feature. With what @jfirebaugh sketched above would it be possible to programmatically disable pseudo class state? I'm thinking about draw tools with modes like "edit" where you don't want to trigger a hover or active state of an existing feature unless you are in "select".

jfirebaugh:
Stacking classes with pseudo classes would allow that level of control:

```
"paint.edit-mode:hover": {
  ...
}
```

maletor:
Is something like this still planned?

lucaswoj:
@maletor Still planned. No timeline yet (sadly).

maletor:
Ya. Right now I'm using interpolation for icon-image and resetting setData
with the updated geoJSON. Must be done on every mouseend event.

Is there a better way?

knutole:
:+1:

dpieri:
+1 would love to see this implemented. Also a :selected pseudo class would be great. The code I've written to emulate this behavior is very fragile and ugly.

elliotleelewis:
Please implement this! Would be so useful for my work project! :)

1ec5:
“Hover” and “active” are concepts that don’t exist on mobile platforms (except when assistive technologies are enabled). It doesn’t seem appropriate to introduce platform-specific concepts into a file format that up to now has been largely platform-agnostic (see also #5545).

andrewharvey:
I think active works for mobile too, it's just a feature that the user has selected.

Agree that hover is inherently linked to desktop where you have a cursor.

kkaefer:
[...]

PR #3701, opened 2016-11-25
Replace setFilter highlighting with new source highlighting

mattesCZ:
Update "query similar features" example

Much faster way to highlight features proposed in #2492 discussion


mattesCZ:
The highlighting is not perfect now. Features should be unioned before highlighting. Is it fine to improve the example with dependency on [turf-union](https://github.com/Turfjs/turf-union)?

jfirebaugh:
Thanks for the suggestions, but we'd like to keep this example simple, without external dependencies. We'll look at issues such as #200 and #2443 for further performance improvements.

Issue #1775, opened 2015-12-01
Support bounds in TileJSON


steinbachr:
the combination of documentation <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources">here</a> which states, _"Tiled sources (vector and raster) must specify their details in terms of the TileJSON specification"_ and the TileJSON documentation <a href="https://github.com/mapbox/tilejson-spec/blob/master/2.1.0/README.md">here</a> leads me think it's possible to add a raster source that only adds tiles within certain bounds. However, when I try to do so, the entire map is tiled, rather than just the portion of map contained within `bounds`. For instance, this code tiles the entire map without respect to `bounds` whereas I would expect it to tile only within the bounds given (`[-10, -10, 10, 10]`):

```
"raster-layer": {
      "type": "raster",
      "tiles": [
           "https://some-tile-source/{z}/{x}_{y}.jpg"
       ],
      "bounds": [-10, -10, 10, 10],
      "tileSize": 256,
      "minzoom": 0,
      "maxzoom": 5
}
```

jfirebaugh:
You're right, TileJSON `bounds` is something that should be supported, but isn't currently.

steinbachr:
ah okay, thanks for the clarification @jfirebaugh

lucaswoj:
ref #1404

Scarysize:
We are using something like this to avoid requests to tiles which are located outside of a given bounding box:

https://gist.github.com/Scarysize/abcf22c514d64125111a5b130cf66fb3

PR #3703, opened 2016-11-28
Implement TileJSON bounds property

jlaxson:
This is rough code to implement the bounds property in TileJSON to prevent loading/rendering tile data outside an extent.  This functionality is useful for eliminating useless network roundtrips (that either result in 404 or an empty response) and eliminating lots and lots of error messages in the console in such a situation.  Addresses #1775 and probably others.

If someone can confirm this is something you'd accept and 👍 the approach, I'll clean it up for inclusion.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/js/source/raster_tile_source.js
+++ b/js/source/raster_tile_source.js
@@ -5,6 +5,7 @@ const ajax = require('../util/ajax');
 const Evented = require('../util/evented');
 const loadTileJSON = require('./load_tilejson');
 const normalizeURL = require('../util/mapbox').normalizeTileURL;
+const TileBounds = require('./tile_bounds');
 
 class RasterTileSource extends Evented {
 
@@ -20,6 +21,8 @@ class RasterTileSource extends Evented {
         this.tileSize = 512;
         this._loaded = false;
         util.extend(this, util.pick(options, ['url', 'scheme', 'tileSize']));
+
+        this.setBounds(options.bounds);
 
         this.setEventedParent(eventedParent);
         this.fire('dataloading', {dataType: 'source'});
@@ -37,13 +40,25 @@ class RasterTileSource extends Evented {
         this.map = map;
     }
 
+    setBounds(bounds) {
+      this.bounds = bounds;
+      if (bounds) {
+        this.tileBounds = new TileBounds(bounds, this.minzoom, this.maxzoom);
+      }
+    }
+
     serialize() {
         return {
             type: 'raster',
             url: this.url,
             tileSize: this.tileSize,
-            tiles: this.tiles
+            tiles: this.tiles,
+            bounds: this.bounds,
         };
+    }
+
+    hasTile(coord) {
+        return !this.tileBounds || this.tileBounds.contains(coord);
     }
 
     loadTile(tile, callback) {
--- a/js/source/tile_bounds.js
+++ b/js/source/tile_bounds.js
@@ -1,0 +1,40 @@
+const LngLatBounds = require('../geo/lng_lat_bounds');
+const clamp = require('../util/util').clamp;
+
+class TileBounds {
+  constructor(bounds, minzoom, maxzoom) {
+    this.bounds = LngLatBounds.convert(bounds);
+    this.minzoom = minzoom || 0;
+    this.maxzoom = maxzoom || 24;
+    this.cache = {};
+    this.updateCache();
+  }
+
+  contains(coord) {
+    const level = this.cache[coord.z];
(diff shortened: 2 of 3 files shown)


js/source/tile_bounds.js
```diff
@@ -0,0 +1,40 @@
...
+    let hit = coord.x >= level[0][0] && coord.x < level[1][0] && coord.y >= level[0][1] && coord.y < level[1][1];
+    return hit;
+  }
+
+  updateCache() {
+    for (let i = this.minzoom; i <= this.maxzoom; i++) {
+      this.cache[i] = [
+        [Math.floor(this.lngX(this.bounds.getWest(), i)), Math.floor(this.latY(this.bounds.getNorth(), i))],
+        [Math.ceil(this.lngX(this.bounds.getEast(), i)), Math.ceil(this.latY(this.bounds.getSouth(), i))]
+      ]
```
lucaswoj: What's the performance impact of not caching these values and doing the calculations from scratch every time `hasTile` is called? If we could reduce the amount of state we hold & simplify this logic without a measurable perf impact, that'd be worth looking into.

js/source/raster_tile_source.js
```diff
@@ -37,15 +40,27 @@ class RasterTileSource extends Evented {
...
+    }
+
     serialize() {
         return {
             type: 'raster',
             url: this.url,
             tileSize: this.tileSize,
-            tiles: this.tiles
+            tiles: this.tiles,
+            bounds: this.bounds,
```
lucaswoj: `serialize` is expected to return a plain object (aka JSON serializable object, POJO). `this.bounds` is an instance of `TileBounds`. We'll need to think of a way to serialize `this.bounds` into a plain object, probably by writing a `TileBounds#serialize` method.

js/source/tile_bounds.js
```diff
@@ -0,0 +1,40 @@
...
+
+  latY(lat, zoom) {
+    let f = clamp(Math.sin(Math.PI / 180 * lat), -0.9999, 0.9999);
+    let scale = Math.pow(2, zoom) / (2 * Math.PI);
+    return Math.pow(2, zoom-1) + 0.5 * Math.log((1 + f) / (1 - f)) * -scale;
+    return y;
+  }
+}
+
+module.exports = TileBounds;
```
lucaswoj: There are some linter errors in this file. You can see these errors on the [CI build](https://circleci.com/gh/mapbox/mapbox-gl-js/6131?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). You can run all the GL JS tests, including the linter, locally per [these instructions](https://github.com/mapbox/mapbox-gl-js/blob/master/test/README.md).

js/source/raster_tile_source.js
jlaxson: This is why I kept both `this.bounds` as unmodified from what the client provides as well as storing the initialized object as `this.tileBounds`.  The naming at the moment leaves something to be desired for sure.  It seems preferable to me that TileBounds not need to worry about how to round-trip whatever source format the tile source uses?

jlaxson:
@lucaswoj how do you feel about the approach of delegating hasTile from SourceCache to the Source object?  If good, I'll extend that to the vector tile source as well.

js/source/raster_tile_source.js
lucaswoj: You're right, I was tripped up by the naming of `bounds` vs `tileBounds`. Curious if there's a quick fix that would make this clearer, perhaps `serializedBounds`?

js/source/tile_bounds.js
lucaswoj: Definitely on the right track. Spiff away! 👍  You may want to check if there's a plugin for your editor that supports the [`.editorconfig`](http://editorconfig.org/) spec.

lucaswoj:
> @lucaswoj how do you feel about the approach of delegating hasTile from SourceCache to the Source object? If good, I'll extend that to the vector tile source as well.

That sounds like the right thing to do. `SourceCache` should be as dumb as possible. 😄

tgecho:
Any chance of movement on this? I'd be happy to help tidy up the lint issues, but the other changes will take a bit more for me to grok.

lrlinde:
Any chance this will be merge soon? Really looking forward to get this fixed! :)

stackTom:
I'm also in need of this feature.

mollymerp:
wrapping this up over at #4556 – thanks again @jlaxson!

PR #3705, opened 2016-11-28
More flow

vicapow:
Adding type annotations for a few more files. Most are straightforward except `js/util/struct_array.js` which got a little gnarly.


js/util/is_char_in_unicode_block.js
```diff
@@ -1,280 +1,281 @@
...
+    // 'Supplemental Symbols and Pictographs': (char: number) => char >= 0x1F900 && char <= 0x1F9FF,
+    // 'CJK Unified Ideographs Extension B': (char: number) => char >= 0x20000 && char <= 0x2A6DF,
+    // 'CJK Unified Ideographs Extension C': (char: number) => char >= 0x2A700 && char <= 0x2B73F,
+    // 'CJK Unified Ideographs Extension D': (char: number) => char >= 0x2B740 && char <= 0x2B81F,
+    // 'CJK Unified Ideographs Extension E': (char: number) => char >= 0x2B820 && char <= 0x2CEAF,
+    // 'CJK Compatibility Ideographs Supplement': (char: number) => char >= 0x2F800 && char <= 0x2FA1F,
+    // 'Tags': (char: number) => char >= 0xE0000 && char <= 0xE007F,
+    // 'Variation Selectors Supplement': (char: number) => char >= 0xE0100 && char <= 0xE01EF,
+    // 'Supplementary Private Use Area-A': (char: number) => char >= 0xF0000 && char <= 0xFFFFF,
+    // 'Supplementary Private Use Area-B': (char: number) => char >= 0x100000 && char <= 0x10FFFF,
```
mourner: Could we instead define the exported object type (`type UnicodeBlockLookup = {[string]: (number) => number}` etc.) and avoid setting types on every line?

js/util/lru_cache.js
```diff
@@ -82,7 +87,7 @@ class LRUCache {
      * @returns {Array<string>} an array of keys in this cache.
```
mourner: In theory, if we upgrade to the latest `documentation` beta, we can strip out all return types for flow-annotated methods since they'll be inferred.

js/util/mapbox.js
```diff
@@ -80,18 +106,21 @@ exports.normalizeTileURL = function(tileURL, sourceURL, tileSize) {
...
 }
 
 const urlRe = /^(\w+):\/\/([^/?]+)(\/[^?]+)?\??(.+)?/;
 
-function parseUrl(url) {
+function parseUrl(url: string): ?UrlObject {
     const parts = url.match(urlRe);
+    if (!parts) {
+        return null;
+    }
```
mourner: Can we avoid all the repetitive null checks and ambiguity here by just throwing an error?

js/util/struct_array.js
```diff
@@ -144,7 +183,25 @@ class StructArray {
...
+                this.int16 = new viewTypes.Int16(this.arrayBuffer);
+            } else if (type === 'Uint16') {
+                this.uint16 = new viewTypes.Uint16(this.arrayBuffer);
+            } else if (type === 'Int32') {
+                this.int32 = new viewTypes.Int32(this.arrayBuffer);
+            } else if (type === 'Uint32') {
+                this.uint32 = new viewTypes.Uint32(this.arrayBuffer);
+            } else if (type === 'Float32') {
+                this.float32 = new viewTypes.Float32(this.arrayBuffer);
+            } else if (type === 'float64') {
```
mourner: Should be capitalized.

js/util/struct_array.js
```diff
@@ -144,7 +183,25 @@ class StructArray {
...
+                this.float64 = new viewTypes.Float64(this.arrayBuffer);
+            }
```
mourner: Is there no way to make this work dynamically in Flow?

js/util/struct_array.js
```diff
@@ -204,15 +261,20 @@ const structArrayTypeCache = {};
  *
  * @private
  */
-function createStructArrayType(options) {
+
+function createStructArrayType(options: {|
+  members: Array<{type: ViewType, name: string, components?:? number}>,
```
mourner: Does there need to be two questionmarks, or is it enough to mark `components` optional?

js/util/config.js
```diff
@@ -1,6 +1,16 @@
 'use strict';
+// @flow
 
-module.exports = {
+type Config = {|
```
lucaswoj: What's the function of the `|`s? I don't see this syntax used in the [Flow Object Types documentation](https://flowtype.org/docs/objects.html#_).

js/util/lru_cache.js
lucaswoj: Yes. I think even the current version of `documentation` infers types.

js/util/struct_array.js
```diff
@@ -18,6 +19,10 @@ const viewTypes = {
     'Float64': Float64Array
 };
 
+/* eslint-disable no-undef */
+type ViewType = $Keys<typeof viewTypes>;
```
lucaswoj: This is not part of the flow public API and we should avoid using it.

js/util/struct_array.js
lucaswoj: I think it's acceptable, even idiomatic, to disable flow when performing operations that are easy in untyped Javascript but awkward with flow. Feel free to coerce to `any` or use a [`suppress_comment`](https://flowtype.org/docs/advanced-configuration.html).

js/util/struct_array.js
```diff
@@ -204,15 +261,20 @@ const structArrayTypeCache = {};
...
+|}) {
 
     const key = JSON.stringify(options);
 
     if (structArrayTypeCache[key]) {
         return structArrayTypeCache[key];
     }
 
-    if (options.alignment === undefined) options.alignment = 1;
+    const alignment = (options.alignment === undefined || options.alignment === null) ?
```
lucaswoj: It would be equivalent and more concise to write `options.alignment == null` instead of `(options.alignment === undefined || options.alignment === null)`. We have explicitly allowed this one use of the `==` operator in the GL JS codebase.

js/util/struct_array.js
mourner: It is a part of public API. See https://flowtype.org/docs/typeof.html and https://flowtype.org/docs/utility-types.html.

lucaswoj: I stand corrected! Thanks @mourner.

js/util/mapbox.js
vicapow: That's sort of a matter of preference but by returning `null` you can ensure callers have to handle the `null` case since there's no notion of `throws` in JavaScript. I'll remove for now.

js/util/struct_array.js
vicapow: good catch

js/util/struct_array.js
vicapow: How about I revert it to what it was originally since that should work with `?: alignment`

js/util/config.js
vicapow: It's the somewhat new "exact" object syntax. see: https://flowtype.org/docs/objects.html#exact-object-types

lucaswoj: Ah! Sorry I missed that. Thanks @vicapow.

PR #3719, opened 2016-11-30
Use a more helpful error message when Map options.container doesn't exist in the document

andrewharvey:
## Launch Checklist

Currently if create a new Map but pass in an element ID to `options.container` which doesn't actually exist in the document, the console will log the error:

    Uncaught TypeError: Cannot read property 'classList' of null

I suggest we use a more helpful error message

    Uncaught Error: No element found in the document with ID '${options.container}' as specified as the container in the map constructor.

Although the test I added just checks that an Error is thrown not the exact message, but an Error was already thrown before, so the test I added actually passes without this change.

None.

 - [ ] post benchmark scores

Shouldn't affect benchmarks.

Manually tested.

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -139,6 +139,7 @@ class Map extends Camera {
 
         if (typeof options.container === 'string') {
             this._container = window.document.getElementById(options.container);
+            if (!this._container) throw new Error(`Container '${options.container}' not found.`);
         } else {
             this._container = options.container;
         }
--- a/test/js/ui/map.test.js
+++ b/test/js/ui/map.test.js
@@ -56,6 +56,11 @@ test('Map', (t) => {
         t.ok(map.keyboard.isEnabled());
         t.ok(map.scrollZoom.isEnabled());
         t.ok(map.touchZoomRotate.isEnabled());
+        t.throws(() => {
+            new Map({
+                container: 'anElementIdWhichDoesNotExistInTheDocument'
+            });
+        }, new Error("Container 'anElementIdWhichDoesNotExistInTheDocument' not found"), 'throws on invalid map container id');
         t.end();
     });
 


test/js/ui/map.test.js
```diff
@@ -56,6 +56,11 @@ test('Map', (t) => {
         t.ok(map.keyboard.isEnabled());
         t.ok(map.scrollZoom.isEnabled());
         t.ok(map.touchZoomRotate.isEnabled());
+        t.throws(() => {
+            new Map({
+                container: 'anElementIdWhichDoesNotExistInTheDocument'
+            });
+        }, Error, 'throws on invalid map container id');
```
lucaswoj: > Although the test I added just checks that an Error is thrown not the exact message, but an Error was already thrown before, so the test I added actually passes without this change.

You can change `Error` to a regex or string that matches the desired error message. This will prevent other errors (i.e. `Map is not defined`) from causing this test to pass when an undesired error is thrown. 

See the tap assert docs for `t.throws` and `t.match` http://www.node-tap.org/asserts/

js/ui/map.js
```diff
@@ -139,6 +139,7 @@ class Map extends Camera {
 
         if (typeof options.container === 'string') {
             this._container = window.document.getElementById(options.container);
+            if (!this._container) throw new Error(`No element found in the document with ID '${options.container}' as specified as the container in the Map constructor.`);
```
lucaswoj: Do you think it's possible to shorten this error message without making it less intuitive for our users?

test/js/ui/map.test.js
andrewharvey: I had problems with getting the object with a regex approach to work. I tried `{Error:/^Container 'anElementIdWhichDoesNotExistInTheDocument' not found$/}`, so in the end I simply used `new Error("Container 'anElementIdWhichDoesNotExistInTheDocument' not found")` which worked.

I checked that the test failed without the change to js/ui/map.js and then passed with the change in js/ui/map.js.

PR #1955, opened 2016-01-20
Fix rendering of tiles with extent != 4096

lucaswoj:
fixes #1952 

regression caused by #1886, as predicted

PR #3721, opened 2016-11-30
Refactor transform

mourner:
Various clean up and simplification in transform code. As a side-effect, improves performance of `locationPoint` and related methods:

```
# before
locationPoint x 3,028,354 ops/sec ±0.82% (85 runs sampled)
pointLocation x 4,866,038 ops/sec ±1.07% (84 runs sampled)

# after
locationPoint x 4,814,249 ops/sec ±0.72% (89 runs sampled)
pointLocation x 4,535,500 ops/sec ±0.93% (90 runs sampled)
```

👀 @lucaswoj

### DIFF
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -153,18 +153,16 @@ class Transform {
         if (z < options.minzoom) return [];
         if (z > options.maxzoom) z = options.maxzoom;
 
-        const tr = this,
-            tileCenter = tr.locationCoordinate(tr.center)._zoomTo(z),
-            centerPoint = new Point(tileCenter.column - 0.5, tileCenter.row - 0.5);
-
-        return TileCoord.cover(z, [
-            tr.pointCoordinate(new Point(0, 0))._zoomTo(z),
-            tr.pointCoordinate(new Point(tr.width, 0))._zoomTo(z),
-            tr.pointCoordinate(new Point(tr.width, tr.height))._zoomTo(z),
-            tr.pointCoordinate(new Point(0, tr.height))._zoomTo(z)
-        ], options.reparseOverscaled ? actualZ : z).sort((a, b) => {
-            return centerPoint.dist(a) - centerPoint.dist(b);
-        });
+        const centerCoord = this.pointCoordinate(this.centerPoint, z);
+        const centerPoint = new Point(centerCoord.column - 0.5, centerCoord.row - 0.5);
+        const cornerCoords = [
+            this.pointCoordinate(new Point(0, 0), z),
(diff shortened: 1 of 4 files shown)


mourner:
This is just the first step — the next one will be fixing #2868 by refactoring the overscaling code.

js/geo/transform.js
```diff
@@ -186,16 +179,16 @@ class Transform {
     zoomScale(zoom) { return Math.pow(2, zoom); }
     scaleZoom(scale) { return Math.log(scale) / Math.LN2; }
 
-    project(lnglat, worldSize) {
```
lucaswoj: Are you sure the code in this PR works with tiles of different sizes?

lucaswoj:
LGTM other than the concern about de-parameterizing `worldSize` ☝️

js/geo/transform.js
mourner: These parameters are not related to supporting different tile sizes. I think I originally introduced it when implementing `flyTo` (which was the only occurrence where the `worldSize` argument was used outside of Transform), but I found a way to get rid of it. 

The only transform method that cares about source tile sizes is `coveringZoomLevel`, which I think should be moved elsewhere. Moreover,`Transform` `tileSize` doesn't really have the meaning it implies by its name — what it really does is establish an arbitrary relationship between zoom value and the pixel width of the world, so that e.g. it's 512px when we set zoom level to 0. I'd like to refactor this to make things clearer.

lucaswoj: Ah! Sorry about that. I think I am confusing "world size" and "tile extent" https://github.com/mapbox/mapbox-gl-js/pull/1955 in my memory. 

---

> These parameters are not related to supporting different tile sizes.

It does appear that `worldSize` has _some_ relation to `tileSize`.

```js
get worldSize() {
    return this.tileSize * this.scale;
}
```

I trust your judgement here and will ✅  the PR.

mourner: > It does appear that worldSize has some relation to tileSize.

This `tileSize` (which I mentioned in my comment above) is a global constant for `mapbox-gl` and does not need parametrization.

anandthakker:
@mourner you mentioned [here](https://github.com/mapbox/mapbox-gl-js/pull/3721#issuecomment-263976227) something about refactoring the overscaling code (and thus resolving this issue).  Do you have a plan/approach in mind for that?

mourner:
@anandthakker I attempted it a few weeks ago but got lost and would need to start again. The plan is to never treat the `z` coordinate of a tile object differently depending on context, and pass a separate zoom value (from the current view) for evaluating style and overscaling tiles.

PR #3733, opened 2016-12-02
Fallback of original jsdom HTML Canvas getContext

cammanderson:
Hi Guys, 

Submitting this PR. 

Motivation is that the current prototype affects HTMLCanvasElement to only supports 'webgl' context. If your current jsdom HTMLCanvasElement supports additional context (such as '2d' using node-canvas) the new PR will still support the context that is supported by the original window prototype. 

Scenario; If you try to create a mapboxgl.Map instance (in nodejs) with a mapbox style that has 2D context requirements, you will get an exception "context.drawImage is not a function". The jsdom supplied HTMLCanvasElement though could support other canvas context, but as this method forces context to only support webgl you will get TypeErrors thrown from inside Mapbox GL JS.

The PR works by falling back to the support of the current HTMLCanvasElement for any other context requests.

Thanks
Cam

### DIFF
--- a/js/util/window.js
+++ b/js/util/window.js
@@ -30,12 +30,19 @@ function restore() {
     window.HTMLElement.prototype.clientLeft = 0;
     window.HTMLElement.prototype.clientTop = 0;
 
-    window.HTMLCanvasElement.prototype.getContext = function(type, attributes) {
-        if (!this._webGLContext) {
+    window.HTMLCanvasElement.prototype.getContext = (function () {
+      let original = window.HTMLCanvasElement.prototype.getContext;
+      return function (type, attributes) {
+        if (type === 'webgl') {
+          if (!this._webGLContext) {
             this._webGLContext = gl(this.width, this.height, attributes);
+          }
+          return this._webGLContext;
         }
-        return this._webGLContext;
-    };
+        original = original.bind(this);
+        return original(type, attributes);
+      };
+    }());
 
     window.useFakeXMLHttpRequest = function() {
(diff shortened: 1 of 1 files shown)


js/util/window.js
```diff
@@ -30,12 +30,19 @@ function restore() {
...
+        let original = window.HTMLCanvasElement.prototype.getContext;
+        return function (type, attributes) {
+            if (type === 'webgl') {
+                if (!this._webGLContext) {
+                    this._webGLContext = gl(this.width, this.height, attributes);
+                }
+                return this._webGLContext;
+            }
+            original = original.bind(this);
+            return original(type, attributes);
```
mourner: Instead of doing bind and then calling the function, you could just do `original.call(this, type, attributes);`.

js/util/window.js
```diff
@@ -30,12 +30,19 @@ function restore() {
...
     window.HTMLElement.prototype.clientTop = 0;
 
-    window.HTMLCanvasElement.prototype.getContext = function(type, attributes) {
-        if (!this._webGLContext) {
-            this._webGLContext = gl(this.width, this.height, attributes);
-        }
-        return this._webGLContext;
-    };
+    window.HTMLCanvasElement.prototype.getContext = (function () {
+        let original = window.HTMLCanvasElement.prototype.getContext;
```
mourner: I don't think you need an self-invoking function here. You can just put the reference to original method outside of the scope and call it `originalGetContext`.

cammanderson:
Thanks for the feedback @mourner. I'll adjust now.

PR #3734, opened 2016-12-02
Use a custom gl-matrix build

mourner:
I made a ES6 fork of gl-matrix to be able to make customized builds here: https://github.com/mapbox/gl-matrix

The `mapbox` branch [configuration](https://github.com/mapbox/gl-matrix/blob/mapbox/src/mapbox-build.js) includes just the stuff needed by GL JS, and I published this version as `@mapbox/gl-matrix@0.0.1`. Resulting build sizes:

File | Original | New | Original (gz) | New (gz)
--- | --- | --- | --- | ---
mapbox-gl.js | 520.73KB | 471.02KB | 133.38KB | 122.21KB
mapbox-gl-dev.js | 3.69MB | 3.27MB | 785.75KB | 719.95KB

cc @jfirebaugh @lucaswoj @tmcw

### DIFF
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -7,7 +7,7 @@ const LngLat = require('./lng_lat'),
     interp = require('../util/interpolate'),
     TileCoord = require('../source/tile_coord'),
     EXTENT = require('../data/extent'),
-    glmatrix = require('gl-matrix');
+    glmatrix = require('@mapbox/gl-matrix');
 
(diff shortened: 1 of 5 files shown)


tmcw:
Almost 10% savings for the gzipped and uglified build = definitely worth it, imho.

mourner:
Also, 10% off the non-gzipped build probably means 10% better time-to-first-render, because workers take less to initialize with less code.

PR #3738, opened 2016-12-05
add Geolocation PositionOptions as option to Geolocate Control (#3473)

andrewharvey:
## Launch Checklist

 - [x] briefly describe the changes in this PR

Implementation for https://github.com/mapbox/mapbox-gl-js/issues/3473, for example allowing developers to [enableHighAccuracy](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions/enableHighAccuracy).

 - [ ] write tests for all new functionality

There are currently no tests for the Geolocation control, and I'm not sure how I would write tests without creating a mock Geolocation API.

 - [x] document any changes to public APIs
 - [ ] post benchmark scores
 - [x] manually test the debug page

Issue #3008, opened 2016-08-15
Geolocate question


uhuru-be:
When I add the Geolocate control to my map and click on it, the map is centered and zoomed in to my current location. 

I could than add a marker to the map (although I expected a marker to be shown on my current position)
When I then walk (or ride) I would like the marker to follow me on the map... Is this possible ? or does I have to implement it myself ?

peterqliu:
@uhuru-be you'd implement this yourself. The easiest way I imagine for this, is to put a static HTML div atop the map to represent the marker. Then, you periodically poll your current location and move the map to that position.

achalnexus:
I have only Geolocate autocomplete GL-Mapbox  use. Load particular country wise data load. What i to do.?

lucaswoj:
I can't see us adding a marker at the user's position. Most desktop web browsers do not have GPS and cannot determine their position accurately enough for this feature. You're welcome to add this feature yourself using `Marker`s or `symbol`s.

andrewharvey:
See also #4479

PR #3739, opened 2016-12-05
Geolocate watch position

andrewharvey:
Not sure if this is a feature which is wanted or not. I saw https://github.com/mapbox/mapbox-gl-js/issues/3008 which seems to rule out adding the a dot on the map for the users location, but people still use GL JS for mobile and native support to track a location is useful.

Added on option to the Geolocate control to watch the position (ie. reposition the map as the user moves) using Geolocation watchPosition.

 - [ ] post benchmark scores
Tested in Chrome and Firefox on Desktop and Mobile. (ps in Chrome Dev Tools, ... -> More Tools -> Sensors comes in handy)

I based this off https://github.com/mapbox/mapbox-gl-js/pull/3738

TODO
 - [x] use a blue for the geolocate toggle which stands out more


lucaswoj:
> Not sure if this is a feature which is wanted or not. I saw #3008 which seems to rule out adding the a dot on the map for the users location, but people still use GL JS for mobile and native support to track a location is useful.

Apologies. My language in #3008 reflects a position a little stronger than what I currently feel. While implementing this feature wasn't a priority for the GL team, we're happy to accept a PR that implements it.

lucaswoj:
For other reviewers' reference, this is what the "watching" and "not watching" icons look like. 

screen shot 2016-12-05 at 12 02 53 pm
screen shot 2016-12-05 at 12 02 20 pm

js/ui/control/geolocate_control.js
```diff
@@ -104,13 +112,30 @@ class GeolocateControl extends Evented {
...
-            this._onSuccess, this._onError, geoOptions);
-
-        // This timeout ensures that we still call finish() even if
-        // the user declines to share their location in Firefox
-        this._timeoutId = setTimeout(this._finish, 10000 /* 10sec */);
+        const positionOptions = util.extend(defaultGeoPositionOptions, this.options && this.options.positionOptions || {});
+
+        // toggle watching the device location
+        if (this.options.watchPosition) {
+            if (this._watching) {
```
lucaswoj: Would it be possible to eliminate the `_watching` flag and use `_geolocationWatchID` as the single source of truth about which state we're in?

andrewharvey: Good point. I've just changed it. The only caveat was that 0 seems to be a valid id (spec doesn't rule it out and Firefox Desktop seems to start at 0) so I need to explicitly check `!== undefined`.

andrewharvey:
I've just added [`aria-pressed`](https://www.w3.org/TR/wai-aria/states_and_properties#aria-pressed) support.

> I think the current design looks good. Any thoughts @tristen?

I think it works well too, it was more that when I was testing on mobile with the screen brightness turned down at night it was harder to distinguish between the two states.

The icon is duplicated in the stylesheet, I originally wanted to have just the black icon and use a [CSS filter like this](http://codepen.io/noahblon/post/coloring-svgs-in-css-background-images#css-filters-2) to change the color on toggle, but IE support for filter isn't present.

tristen:
The blue is nice ... it's a shame we can't just change the fill color and I can see how it would be harder to distinguish between the two states.

I think we should represent this as a disabled state by dropping the opacity of the icon instead? That would save us from adding an encoded string in the stylesheet and make state more evident on dimly lit screens.

andrewharvey:
> I think we should represent this as a disabled state by dropping the opacity of the icon instead? That would save us from adding an encoded string in the stylesheet and make state more evident on dimly lit screens.

So there are two modes for this control. Is this what you're suggesting @tristen ?
1. `watchPosition == false`
on page load
selection_426

user clicks the button which takes the map to their location, the button has no state.

2. `watchPosition == true`
on page load
selection_425
default state is off, user then toggles the button on to track location so state changes to on and becomes
selection_426

I like the blue, it indicates that the control is actively watching the position. I think the one with opacity makes it look like the control is disabled and can't be clicked.

What about a tad lighter blue?

tristen:
@andrewharvey I didn't fully read the proposed change. Right: An opacity change would make this look disabled which we shouldn't do. No strong opinions here - I think a tad lighter blue makes sense.

---

This might be out of the realm of this PR but we should add these icon strings as url encoded svg [like the others](https://github.com/mapbox/mapbox-gl-js/blob/master/dist/mapbox-gl.css#L79).

andrewharvey:
>@andrewharvey I didn't fully read the proposed change. Right: An opacity change would make this look disabled which we shouldn't do. No strong opinions here - I think a tad lighter blue makes sense.

@tristen No problem. I've changed the blue to `#00f`. Not sure if that's too bright now though.

selection_427

> This might be out of the realm of this PR but we should add these icon strings as url encoded svg like the others.

I agree, I've changed it over so they are all consistent and limits this premature optimization (at least you can change fill color in place).

PR #3740, opened 2016-12-05
Fix broken map on pitch=0 on some systems

### LINKED ISSUE #2270: Map does not render unless pitch is set and map is "tilted"
Mapbox does not render tiles unless map pitch is > 0 on certain Windows 7 systems. This may be related to https://github.com/mapbox/mapbox-gl-js/issues/930 but in 930 the map was rendering. In this case, the map does not render anything at all until a pitch is set.

In the screenshots below, the rendering visible on the map is custom GL we've written and render on a separate canvas external to mapbox.

The same behavior also occurs across all mapbox-gl examples at https://www.mapbox.com/mapbox-gl-js/examples/

Map with no pitch set.
blankmap

Map with pitch set to 60.
maptilted

System Specs Screenshot (Processor)
systemspecs1

System Specs Screenshot (Graphics Card)
systemspecs2

**mapbox-gl-js version**: 0.15.0
### Steps to Trigger Behavior
1. Test any current mapbox example. All examples are blank.
2. map.setPitch(45)
3. Tiles render correctly.
4. map.setPitch(0)
5. Map no longer renders.
### Expected Behavior

Map renders normally.
### Actual Behavior

Map does not render unless tilted/pitched.

mourner:
Fixes #2270 by making sure a floating point error does not accumulate into a broken projection matrix.

When pitch = 0, a `Float32Array` perspective matrix would end up with `1.5000001192092896, 1.5` as the last values, while `Float64Array` produces correct `1.5, 1.5`. Empirically, I found that the first value never exceeds the second one with proper precision. When it does (which happens even with `Float64Array` on some systems), the map disappears. So while I don't fully grasp the math behind this, this hack seems to fix the pitch=0 bug.

cc @ansis @jfirebaugh @lucaswoj

### DIFF
--- a/js/geo/transform.js
+++ b/js/geo/transform.js
@@ -408,6 +408,9 @@ class Transform {
         mat4.perspective(m, 2 * Math.atan((this.height / 2) / this.altitude), this.width / this.height, 0.1, farZ);
         mat4.translate(m, m, [0, 0, -this.altitude]);
 
+        // a hack around https://github.com/mapbox/mapbox-gl-js/issues/2270
+        m[14] = Math.min(m[14], m[15]);
+
         // After the rotateX, z values are in pixel units. Convert them to
         // altitude units. 1 altitude unit = the screen height.
         mat4.scale(m, m, [1, -1, 1 / this.height]);


js/geo/transform.js
```diff
@@ -408,6 +408,9 @@ class Transform {
         mat4.perspective(m, 2 * Math.atan((this.height / 2) / this.altitude), this.width / this.height, 0.1, farZ);
         mat4.translate(m, m, [0, 0, -this.altitude]);
 
+        // a hack around https://github.com/mapbox/mapbox-gl-js/issues/2270
+        m[14] = Math.min(m[14], m[15]);
```
lucaswoj: The rewards of this not-100%-understood fix outweigh the severity of the bug & risks of the fix in my mind.

Out of curiosity, is there any intuitive meaning of elements 14 and 15 in the transformation matrix? Can we anticipate any cases where elements 14 and 15 might end up with very different values? cc @ansis

mourner: There seem to be no intuitive meaning since the perspective projection math is [a bit involved](http://www.songho.ca/opengl/gl_projectionmatrix.html), but at this point of position matrix calculation, `m[15]` always equals the altitude value (`1.5`), and `m[14]` represents translation along Z axis (which always seems to be `<=1.5`).

Issue #3735, opened 2016-12-02
SourceCache race condition can cause sources to fail to bubble events up to `map`


lbud:
There is a race condition wherein sources that quickly fire events upon construction fail to bubble events to their omniparent (`map`) because sources are created in SourceCache [(source_cache.js#L30)](https://github.com/mapbox/mapbox-gl-js/blob/6b0bb7edccef0f9e95c066b01e808759d4356ee2/js/source/source_cache.js#L30), but the SourceCache parent is created before calling `setEventedParent`[(style.js#L376-L382)](https://github.com/mapbox/mapbox-gl-js/blob/6b0bb7edccef0f9e95c066b01e808759d4356ee2/js/style/style.js#L376-L382) to add it to the `Map` -> `Style` -> `SourceCache` ownership chain. Generally in our existing source types this isn't a problem, as they usually load data over a network connection and don't race to fire events as quickly (so when they do they already have a `map` grandparent), but I noticed this problem while working on the canvas source type #3580 which loads its data from the same page.

PR #3761, opened 2016-12-07
Wait to start loading source assets until `onAdd` is called

lbud:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/3735 by taking advantage of the `onAdd` method to load assets/fire events. Previously we were doing those in the source constructors, and they were constructed before establishing ownership (map -> style -> source_cache -> source), meaning events fired almost immediately wouldn't bubble up to the map object — so for example, vector tile sources (loaded over network connection) would reliably trigger `map.on('source.load', fn)` listeners, but canvas sources which loaded very quickly (from the same page) would not.

This isn't a perfect solution, but since sources will soon undergo a pretty big overhaul… 🔨 

cc @lucaswoj

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -109,7 +109,6 @@ class Style extends Evented {
         } else {
             browser.frame(stylesheetLoaded.bind(this, null, stylesheet));
         }
-
         this.on('source.load', (event) => {
             const source = this.sourceCaches[event.sourceId].getSource();
             if (source && source.vectorLayerIds) {
@@ -374,14 +373,14 @@ class Style extends Evented {
         if (shouldValidate && this._validate(validateStyle.source, `sources.${id}`, source, null, options)) return;
 
         const sourceCache = this.sourceCaches[id] = new SourceCache(id, source, this.dispatcher);
-        sourceCache.style = this;
         sourceCache.setEventedParent(this, () => ({
             isSourceLoaded: sourceCache.loaded(),
             source: sourceCache.serialize(),
             sourceId: id
         }));
-
+        sourceCache.style = this;
         sourceCache.onAdd(this.map);
+
         this._changed = true;
     }
 
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -94,8 +94,7 @@ class Style extends Evented {
             }
 
             if (stylesheet.sprite) {
-                this.sprite = new ImageSprite(stylesheet.sprite);
-                this.sprite.setEventedParent(this);
+                this.sprite = new ImageSprite(stylesheet.sprite, this);
             }
 
             this.glyphSource = new GlyphSource(stylesheet.glyphs);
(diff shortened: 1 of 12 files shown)


js/style/style.js
```diff
@@ -374,14 +373,14 @@ class Style extends Evented {
...
             isSourceLoaded: sourceCache.loaded(),
             source: sourceCache.serialize(),
             sourceId: id
         }));
-
+        sourceCache.style = this;
         sourceCache.onAdd(this.map);
+
         this._changed = true;
     }
```
lucaswoj: Are the changes to this file significant?

lbud:
@lucaswoj the way it has been working:
* `Style` creates instances of `SourceCache`s, _then_ sets their parent: [style.js#L376-L382](https://github.com/mapbox/mapbox-gl-js/blob/2d10449743a61456ed3585ad97a101ad6252f6e8/js/style/style.js#L376-L382)
  * it does this because its `eventedParentData` is dependent on the Source having already been created (`sourceCache.serialize()` and `sourceCache.loaded()` both just forward to `sourceCache._source` methods)
  * this is where the race occurs: when sources fire events in their constructors before we've set up SourceCache's parent. 
    * Specifically I was seeing this with the canvas source branch, since we load so quickly (from the same page). 
    * However, at a glance I can see that other events like [`VectorTileSource#dataloading`](https://github.com/mapbox/mapbox-gl-js/blob/2d10449743a61456ed3585ad97a101ad6252f6e8/js/source/vector_tile_source.js#L31) wouldn't always bubble up to `Map` for the same reason. Was just able to reproduce this with dynamically adding a vector tile set (it fires [`source.load`](https://github.com/mapbox/mapbox-gl-js/blob/2d10449743a61456ed3585ad97a101ad6252f6e8/js/source/vector_tile_source.js#L40) because it's slow enough, but never [`dataloading`](https://github.com/mapbox/mapbox-gl-js/blob/2d10449743a61456ed3585ad97a101ad6252f6e8/js/source/vector_tile_source.js#L31))
* `SourceCache` creates instances of `Source`s, passing itself as a constructor so a source can be responsible for calling `setEventedParent`
  * We were unnecessarily setting this twice: in individual source constructors and also in `Source.create` https://github.com/mapbox/mapbox-gl-js/pull/3761/files#diff-d5506f072d1241401a7e31ff642c1606L24

Another minor, related change I made in this PR is to pass `Style` in the `ImageSprite` constructor, which looked like it could feasibly suffer the same race condition: https://github.com/mapbox/mapbox-gl-js/pull/3761/files#diff-6369a555a491e3e9a58fd29c3fd5a371L97

lucaswoj:
@lbud Are you ready to merge this PR? I'd hate for a rebase conflict to emerge 😅

lbud:
Oops yes @lucaswoj — I paused on this with all my naïve optimism about #3186 😬 
🚢

PR #2027, opened 2016-01-30
add all points to circle bucket not just outer ring

ansis:
CircleBucket used to only add points from the outer ring. It now adds all of them just like -native.

test-suite: https://github.com/mapbox/mapbox-gl-test-suite/tree/multi-circle

:eyes: @lucaswoj

PR #3763, opened 2016-12-08
Fix symbol rendering for multipoints

lucaswoj:
Via a support request, @mapsam and I discovered that we do not render any symbols for multipoint geometries. This is inconsistent with our behaviour for circles in which we do render a circle at each point in the multipoint geometry (see https://github.com/mapbox/mapbox-gl-js/pull/2027). 

This PR tries to differentiate multipoint geometries from polygons and perform the correct label placement in each case. 

cc @mourner @flippmoke @mapsam @jfirebaugh @colleenmcginnis 

## Open Questions

 - Is there a better way to differentiate between a multipoint and a polygon within GL JS?
 - Should we have a different `symbol-placement` modes for `point`, `line`, and `polygon` / `centroid`?
 - Are there other cases in which we don't handle multipoints properly?

## Launch Checklist

### DIFF
--- a/js/data/bucket/symbol_bucket.js
+++ b/js/data/bucket/symbol_bucket.js
@@ -337,15 +337,29 @@ class SymbolBucket {
             mayOverlap = layout['text-allow-overlap'] || layout['icon-allow-overlap'] ||
                 layout['text-ignore-placement'] || layout['icon-ignore-placement'],
             symbolPlacement = layout['symbol-placement'],
-            isLine = symbolPlacement === 'line',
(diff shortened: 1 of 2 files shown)


lucaswoj:
Updated with new tests & use of the vector tile `type` field 😎

lucaswoj:
It's unfortunate that we place circles & symbols differently for lines, polygons, and multi-things. If pole-of-inaccessability symbol placement were a `symbol-placement` mode, we could have more consistent rendering.

PR #3532, opened 2016-11-03
Ensure images always fade in completely, fix broken "raster-fade-duration" property

lucaswoj:
As explained in #3526 , Mapbox GL JS does not properly implement `raster-fade-duration`. This PR fully removes support for "raster-fade-duration" and thereby ensures that images always fade in completely.

cc @jfirebaugh @mollymerp @mourner @mapsam 

## Launch Checklist

 - [x] briefly describe the changes in this PR
 - [x] ~write tests for all new functionality~ we do not have any infrastructure to test animations
 - [x] document any changes to public APIs
 - [x] manually test the debug page

PR #3764, opened 2016-12-08
Ensure animation loop stops after raster tiles fade in

### LINKED ISSUE #3757: Image layer seems to trigger render loop and consume CPU without interaction
mapbox-gl-js version: 0.28.0

For some reason it seems like simply having an image layer enabled triggers the render loop, even when no interaction with the map is carried out.

### Steps to Trigger Behavior

 1. Go to http://codepen.io/anon/pen/gLzQqw
 2. Perform a CPU profile in Chrome and notice that it's almost completely idle
 3. Click the "Add image" button in the top left corner
 4. Wait for short while and then carry out another CPU profile

You will notice something like:

screen shot 2016-12-07 at 16 15 44

### Expected Behavior
That the map consumes almost no CPU resources when not interacted with.

### Actual Behavior
The behavior described above.

lucaswoj:
PR https://github.com/mapbox/mapbox-gl-js/pull/3532 introduced a bug wherein the render loop never stops when rendering raster tiles. 

Raster tiles interact with the animation loop to allow them to fade in per `raster-fade-duration`. 

This PR improves our handling of raster tile fading to prevent the animation loop from running continuously. 

fixes #3757

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/js/render/draw_raster.js
+++ b/js/render/draw_raster.js
@@ -36,7 +36,7 @@ function drawRasterTile(painter, sourceCache, layer, coord) {
     const tile = sourceCache.getTile(coord);
     const posMatrix = painter.transform.calculatePosMatrix(coord, sourceCache.getSource().maxzoom);
 
-    tile.setAnimationLoop(painter.style.animationLoop, layer.paint['raster-fade-duration']);
+    tile.registerFadeDuration(painter.style.animationLoop, layer.paint['raster-fade-duration']);
 
     const program = painter.useProgram('raster');
     gl.uniformMatrix4fv(program.u_matrix, false, posMatrix);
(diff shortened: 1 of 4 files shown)


lucaswoj:
Found the problem causing the unit tests to fail @mourner. Glad to see the unit tests are working 😅

AndyMoreland:
Hey folks,

This PR looks like it addresses the high cpu usage we've been seeing lately on our map. This is causing pretty significant pain for us because our users tend to have several complicated maps running at once. 

Do y'all have any notion of when you might release an 0.29.0? Happy to cherry-pick this onto the private fork that I deploy off of, but if you're planning on releasing soon I'll hold off since it's the holidays.

Totally understand if you don't want to discuss release schedules publicly but I figured I'd try asking :-).

Andy

mollymerp:
Hi @AndyMoreland,
Glad this solves some issues for you! We don't have a finite schedule for the next release but we should have one out before the holidays.

Issue #3745, opened 2016-12-06
Canvas style integration questions


ibgreen:
Hi, as discussed, we've integrated the canvas styles in your example (on the `canvas_type` branch) with [`react-map-gl`](https://github.com/uber/react-map-gl/tree/canvas.demo) to see if we can use this feature to improve integration with deck.gl.

We can get a canvas to render as a "mapbox layer" in react-map-gl however we have some issues/questions:
1) We don't know when to render. This is probably not critical as we can keep rendering using our own dirty tracking but it would be cool with a callback.
  - it might be a little more efficient as we'd not draw unless mapbox draws.
  - it might also allow us to create nicer code for this on our end.
2) More importantly, this feature is limited without picking integration
  - presumably a callback with x,y coords?
  - If not we'll have to intercept events and do uncoordinated picking
3) Handling perspective seems to involve a bunch of tricky math (as usual). Assume we'll need to  calculate a canvas size using a perspective enable fitBounds operation with both long/lats and pixel sizes set to cover the entire screen.

Q: The code we are using right now is more or less straight from your example. Are there any docs on the branch we should be looking for?
```
      canvas: {
        type: 'canvas',
        canvas: 'canvas',
        coordinates: [
          [-81.490, 46.437],
          [-72.582, 46.437],
          [-72.582, 38.907],
          [-81.490, 38.907]
        ],
        dimensions: [0, 0, 400, 400]
      }
```

@pessimistress  @abmai @vicapow @apercu @gnavvy

lucaswoj:
Thanks for the interest @ibgreen. As you can tell, this branch is still under development so we don't have any documentation or good answers for all of those questions yet. We really appreciate the early feedback. It will help us shape our implementation. 

----

> 1\. We don't know when to render. This is probably not critical as we can keep rendering using our own dirty tracking but it would be cool with a callback.

We are considering a `beforerender`/`willrender` event to compliment our existing `render` event (which fires _after_ the map renders). Would that work for your use?

> 2\. More importantly, this feature is limited without picking integration

Are you asking about custom layer types in conjunction with`queryRenderedFeatures`? This isn't something we've considered yet. There's no technical reason we can't support `queryRenderedFeatures` on custom layer types.

 > 3. Handling perspective seems to involve a bunch of tricky math (as usual). Assume we'll need to calculate a canvas size using a perspective enable fitBounds operation with both long/lats and pixel sizes set to cover the entire screen.

[...]

Issue #1404, opened 2015-07-31
featuresAt should include raster layers


samanpwbb:
It would be useful if raster layers were included in featuresAt. Especially for raster layers with sources with limited bounds. 

I see here that this is a noop: https://github.com/mapbox/mapbox-gl-js/blob/2a2aec44a39e11e73bdf663258bd6d52b83775f5/js/source/raster_tile_source.js#L98

scothis:
Looks to already be working.

<img width="237" alt="screen shot 2015-07-31 at 1 40 01 pm" src="https://cloud.githubusercontent.com/assets/302992/9016927/b63072b8-3789-11e5-9027-114c26dafd99.png">

samanpwbb:
@scothis this is just hacked in to show up across the whole map for now: https://github.com/mapbox/app/blob/mb-pages/src/styles_components/workspace/direct_select.js#L70

Ideally the raster layer only appears in the popup if it exists where user clicked.

lbud:
At this time we don't return raster sources or features from `map.query*Features` at all. We should do this.

This will be dependent on https://github.com/mapbox/mapbox-gl-js/issues/3186: querying searches for features in a tile, but since we shoehorn features into a tile, querying a point on a visible raster layer may search for points past a tile (the source cache's tile)'s "extent."

The equivalent of the noop mentioned in https://github.com/mapbox/mapbox-gl-js/issues/1404#issue-98309860 is now [query_features.js#L12](https://github.com/mapbox/mapbox-gl-js/blob/2d10449743a61456ed3585ad97a101ad6252f6e8/js/source/query_features.js#L12): raster layers don't have feature indices. `querySourceFeatures` also [only queries geojson and vector data](https://github.com/mapbox/mapbox-gl-js/blob/2d10449743a61456ed3585ad97a101ad6252f6e8/js/source/tile.js#L177).

This was brought up in https://github.com/mapbox/mapbox-gl-js/issues/3745. Indeed for something like a canvas-based raster layer it would be helpful to return an [x,y] coordinate mapped back to the original raster data.

mollymerp:
This could be particularly useful for raster layers like the terrain rgb where users could query the pixel values and calculate elevation at a point based on the raster data 💭

jucor:
Any luck on this? This would very useful.

jfirebaugh:
#5916 notes that raster layers with a canvas source should be included with this feature.

shawnd:
[...]

PR #3765, opened 2016-12-08
Canvas type

lbud:
Closes https://github.com/mapbox/mapbox-gl-js/issues/3580.

This adds a `canvas` source type, which can be used in `raster` layers like existing `image` and `video` sources.

To be resolved:
* We have no solid plan for parity with GL Native for this feature. It's also not something that will be very useful in api-styles, because we're loading same-page-based canvases, unlike other sources which often live in the ☁️. These things are weird and we should make sure we're comfortable with this approach.
* [x] add tests
* [ ] add example

Related feature but not a blocker: raster querying https://github.com/mapbox/mapbox-gl-js/issues/1404 (came up in #3745)

## Launch Checklist

 - [x] document any changes to public APIs (generated)


lucaswoj:
> Do we _want_ to support `webgl` contexts (less straightforward, not necessarily the use case we're targeting, and likely to be obseleted by custom layer types)?

Based on what I've heard about perf, we don't need to go out of our way to support WebGL canvases. If we get some support for free from `texImage2D`, that's a nice bonus.

jliebrand:
Curious: will this feature allow me to draw to a 2d canvas, and have it show up on the map (correctly zooming/panning/etc)?

Eg will this allow me animate something on the chart, unrelated to geojson and webgl?

lucaswoj:
> Curious: will this feature allow me to draw to a 2d canvas, and have it show up on the map (correctly zooming/panning/etc)?

Yes!

> Eg will this allow me animate something on the chart, unrelated to geojson and webgl?

Yup 😄

jliebrand:
Awesome! Can't wait!

package.json
```diff
@@ -45,6 +45,7 @@
     "babel-eslint": "^7.0.0",
     "benchmark": "~2.1.0",
     "browserify": "^13.0.0",
+    "canvas": "^1.6.2",
```
lbud: NOTE: this is a relatively nontrivial dep for something just used in one test…discuss

lucaswoj: Can you create a fake Canvas API using Sinon?

lbud:
> * As currently implemented, this can support reading from either 2D or webgl canvas contexts. 2D is ideal and more performant and, in fact, gl.texImage2D / gl.texSubImage2D can take an HTML5 canvas element as an argument, meaning we don't even have to manually copy pixels. (When using a webgl context however this causes very incorrect rendering/artifacts.) Do we want to support webgl contexts (less straightforward, not necessarily the use case we're targeting, and likely to be obseleted by custom layer types)?

Alright, I simplified this a lot by using `texImage2D`/`texSubImage2D` no matter the context/not even doing context detection. This also allows us to take out the `dimensions` property since we don't manually `readPixels`. It performs fine on my simple canvas debug pages but I haven't made anything super flashy to stress test it with yet… We _could_ prohibit `webgl` contexts, or we could just _caveat emptor_ it. 🤔 

Once I pull together a good example and tighten up the docs this should be good to 👀 & 🚢 

> * We have no solid plan for parity with GL Native for this feature. It's also not something that will be very useful in api-styles, because we're loading same-page-based canvases, unlike other sources which often live in the ☁️. These things are weird and we should make sure we're comfortable with this approach.

🤔

lbud:
Rebased; I think this is good to go. I actually think this is fine to merge without an example, at least for now. I do have an example I've been playing around with, but it depends on loading 60-some images…this can be a follow-up task 🤔🌩

Secondly, tests were [failing by hanging/timing out at the ignored line-width property function test](https://circleci.com/gh/mapbox/mapbox-gl-js/6603) because the spec `canvas_type` branch is rebased off of current `mb-pages`, which says line-width property functions are not supported (they were rolled back). The suite implementation still renders even ignored tests, and just doesn't count them against the fail count. But in this case, it errors and hangs on validating the line-width property function test. It's not the cleanest, but I [added a commit](https://github.com/mapbox/mapbox-gl-js/pull/3765/commits/51bfa6a021bb043137fab5a8457d497a9dfdf809) to this PR that creates a style separately of the map for each render test and then adds it to the map, so that we can pass in a `validate` boolean based on whether a test is ignored. 👍/:-1:? (This test has been hanging locally for me for weeks 😕)

js/source/canvas_source.js
```diff
@@ -0,0 +1,129 @@
...
+            this.height = this.canvas.height;
+            this.resize = true;
+        }
+        if (this._hasInvalidDimensions()) {
+            return this.fire('error', new Error('Canvas dimensions cannot be less than or equal to zero.'));
+        }
+
+        if (!this.tile) return; // not enough data for current position
+        this._prepareImage(this.map.painter.gl, this.canvas, this.resize);
+        this.resize = false;
```
lucaswoj: Maybe convert this to a local var?

js/source/canvas_source.js
```diff
@@ -0,0 +1,129 @@
...
+            this.width = this.canvas.width;
+            this.resize = true;
+        }
+        if (this.canvas.height !== this.height) {
+            this.height = this.canvas.height;
+            this.resize = true;
+        }
+        if (this._hasInvalidDimensions()) {
+            return this.fire('error', new Error('Canvas dimensions cannot be less than or equal to zero.'));
+        }
```
lucaswoj: Maybe make this a no-op instead of a failure

test/suite_implementation.js
```diff
@@ -30,13 +31,15 @@ module.exports = function(style, options, _callback) {
...
-        style: style,
         classes: options.classes,
         interactive: false,
         attributionControl: false,
         preserveDrawingBuffer: true
     });
 
+    const _style = new Style(style, map, { validate: !options.ignored });
+    map.setStyle(_style);
+
```
lucaswoj: We can use `skipped` flags to get around validating these tests without adding new code

lbud:
> There's a regression test that also uses a data-driven line width. If you skip that one, everything should be 🍏 and you're clear to :shipit:!

aha :female_detective: I saw line-width was still failing and didn't get a chance to look into it yet. Alright, will do!

averas:
This is awesome! One use case I feel I haven't found an optimal solution for has been to effectively show small thumbnail images in Mapbox GL JS, to create maps like this:

screen shot 2015-06-03 at 20 53 08

While it is possible to come up with something quite close to that with the marker plugin I found it hard to make a seamless experience that switches between symbols (dots/icons) and thumbnails based on zoom, and the markers are not data-driven in a similar manner as with other sources which makes it tedious to add and remove them based on map interaction.

Could the <code>canvas</code> source type be used to create a data driven thumbnail map like the one above?

lucaswoj:
@averas 

It would be possible to create a `canvas` source which renders and places all of the thumbnails. I anticipate the difficulty of implementing this (you'll need to work with the canvas API and explicitly do some projections) may outweigh the benefits. 

Have you tried using implementing this using the `image` source?

averas:
@lucaswoj 

> Have you tried using implementing this using the image source?

No..? The <code> image </code> source can only reference a single image, and once created can't be mutated (one has to replace the source to update the image). I imagine it would quite ineffective to push/pop potentially hundreds of sources/layers when panning/zooming a map, but I may be wrong.

I think the best Mapbox GL JS examples of this are:
https://www.mapbox.com/blog/custom-markers-mapboxgl/
https://www.mapbox.com/mapbox-gl-js/example/custom-marker-icons/

.. but again, both of these use makers and are not data driven by sources (i.e. you have to micro manage the markers with discrete code hooked in to map events).

yangdonglai:
marked and keep attention

bennlich:
> (you'll need to work with the canvas API and explicitly do some projections)

@lucaswoj I'm working with a canvas source right now, wondering this exact thing. I how do I project from lat/lng to canvas coordinates?

PR #3778, opened 2016-12-10
Tweak tile debug to more closely match native

jfirebaugh:
Refs https://github.com/mapbox/mapbox-gl-native/issues/3841


jfirebaugh:
Can't get this test case to pass in CI everywhere due to `GL_LINE` rendering differences; abandoning.

Issue #2525, opened 2016-05-04
Is there a way to specify point/circle radius in meters?


invisibleroads:
Thanks for making Mapbox GL. We are considering migrating from Mapbox JS to Mapbox GL.

Is there a way to specify point/circle radius in meters? Mapbox JS differentiated between [Circle](https://www.mapbox.com/mapbox.js/api/v2.4.0/l-circle/) radius in meters (vary by zoom) and [CircleMarker](https://www.mapbox.com/mapbox.js/api/v2.4.0/l-circlemarker/) radius in pixels (does not vary by zoom).

lucaswoj:
Hi @invisibleroads!

There is not presently a turnkey way to specify circle radius in meters. With some precomputation, it should be possible to use functions to specify circle radius in meters, however this is left as an exercise to the reader.

I could imagine us creating a new GL Function `unit` operator along the lines of

```js
['units', 'size', 'meters']
```

invisibleroads:
Thanks for the response, @lucaswoj. 

I want to emphasize that it is a little more involved, since in the original Mapbox JS, a circle with a radius specified in meters would change size depending on the level of zoom.

princeofnaxos:
+1

jplante:
Required by us too

mblomdahl:
@invisibleroads @jplante @princeofnaxos We had the same issue.. So I created a library to solve it (inspired by the Google Maps circle). Feel free contributing PRs if it's missing something.  Should be sort-of production stable with this release; mblomdahl/mapbox-gl-circle#48

Issue #3770, opened 2016-12-08
document world wrap behavior for DOM elements


lyzidiamond:
While the map wraps around infinitely when panning in either direction, popups only show up on the "main" world. Unfortunately, it's hard to know which world is the main one. I would like to add [this code](https://github.com/mapbox/mapbox-gl-js/issues/2071#issuecomment-245990952) to the doc for `Marker` and `Popup` as well as in the examples that show interactions with markers and popups.

cc @jfirebaugh @lbud @lucaswoj @mollymerp

jfirebaugh:
I'd like to fix #2071 instead, which would make the library behave as most people expect it to behave, and likely remove the need for documentation altogether.

lyzidiamond:
@jfirebaugh what's the timeline on this? We should have something in place at least temporarily for the users who run into this issue -- this could easily be a saved reply, though.

jfirebaugh:
There's not a set timeline. We'd need someone to jump in and do the work. My guess is it's an hour to a day of work.

lucaswoj:
@lyzidiamond How commonly do you think users and developers will expect to be able to interact with multiple wrapped worlds at once at z0? With #2071, Asia on the left of the screenshot below, will not have any popups or markers. 

<img width="1680" alt="screen shot 2016-12-09 at 10 27 02 am" src="https://cloud.githubusercontent.com/assets/281306/21060354/c40b1eb6-bdfb-11e6-875f-3f06127b6faa.png">

jjb867:
The snippet on #2071 works but is less than desirable imho since the map needs to instantly re-render and therefore a flash is created.

Would be nice to have an option to limit the map to 1 world instance or to show a given popup on all instances of the world.

andrewharvey:
I feel it's okay for `Markers` to only appear once (so if you're zoomed out it doesn't duplicate DOM elements).

But even with #2071 implemented, I feel `Marker` will still need to dynamically move between world 0 and -1/1. If your map center is still in world 0, but parts of world -1 are visible, I think the `Marker` should render on world -1 so it's visible on the map.

andrewharvey:
In https://github.com/andrewharvey/mapbox-gl-js/tree/draw-markers-in-other-worlds I've changed the behaviour for Markers and Popups to render in world -1 and world 1 when best.

Together with #2071 I think it address this issue (short of creating duplicate markers and popups which might lead to too many headaches for little gain).

[...]

Issue #2071, opened 2016-02-05
By default, snap back to "w=0" world after panning operation (i.e. "worldCopyJump")


drewbo:
Because certain methods like `flyTo` send the user on to world copies, it would be great if there were an option to duplicate data/overlays to the other world copies as well.

1ec5:
Only `flyTo` goes to a world copy at the moment (#1853). It probably should go back to the original world on the last frame, though.

drewbo:
I think the only issue with that would be that it precludes setting `maxBounds` to prevent people from scrolling off the data as well (it interrupts the `flyTo` animation)

jfirebaugh:
I think `worldCopyJump` behavior should be the only behavior (no option), and `flyTo` should wrap on the first frame that crosses the antimeridian.

lucaswoj:
Duplicate of #2879

jfirebaugh:
I preferred the old implementation proposal, "Create Leaflet-like `worldCopyJump` option". That seems much more tractable than duplicating arbitrary DOM elements, which is what "Add option to copy markers and popups when panning across dateline" would require.

lucaswoj:
Ah! I misunderstood the behaviour of `worldCopyJump`. Changing the title now.

lucaswoj:
This feature is very easy to implement downstream. 

```js
map.on('moveend', function() {
  map.setCenter(map.getCenter().wrap());
});
```

cc @geografa

jjb867:
Any ideas with this?  I have been striuggling to accomplish the same behavior as described in this thread.

jfirebaugh:
@jjb867 Did you try the code snippet from https://github.com/mapbox/mapbox-gl-js/issues/2071#issuecomment-245990952?

jjb867:
Sorry but I'm not sure what you are referring to specifically.  I have a 'zoomend' event that I'm using to trigger a popup after flying to a location.  The basic idea is displaying fake transactions around the world in a popup.  The popup is displayed properly unless I'm displaying one in Hong Kong for example, it appears in Hong Kong, but in one world to the right of where I need it.

I tried 

`map.setCenter(map.getCenter().wrap());`

inside my zoomend function which does work but it creates a flash and a jarring user experience.

Is this what you are referring to?

jfirebaugh:
Yes. Can you create a JSFiddle that shows the issue you're talking about with that approach?

jjb867:
[...]

PR #3781, opened 2016-12-12
add positionMarker option to Geolocate Control and improve watchPosition with an active lock and background mode

andrewharvey:
# Summary
As of GL JS 0.28.0, the Geolocate Control has worked as a button that when clicked the map camera will change to the device's location from using the HTML5 Geolocation API. There is no state, users press the button and the camera updates. This works well for desktop where the device is generally fixed in location.

In this PR I'm proposing an optional secondary mode which opens up more mobile focused use cases where the device's location can change. This is a continuation of #3739, which in hindsight I requested prematurely and should have waited until I had a more complete design which I'm putting forward here.

In this PR the Geolocate Control has the option `watchPosition` defaulting to `false` (traditional desktop mode), which when set to `true` will allow the map camera to update when the device's location changes and optionally show a marker pin pointing the device's location. (#3739 suffers a usability issue where the user lacks freedom to pan the map as as soon as the device's location updates it overrides their pan operation.)

# Design considerations:
* Allows for [Neilsen usability heuristic](https://en.wikipedia.org/wiki/Heuristic_evaluation#Nielsen) number 3: User control and freedom by allowing users to change the camera by panning/zooming and stay in that alternate view even when their device's location is changing.

### Color
* `grey` represents the control is dissabled and you can't interact with it
* `black` represents the control is off/inactive (not watching for location updates)
* `blue` represents the control is on/active (is watching for location updates)
* `red` represents the control is on/active, but there was an error getting the device's location

### Animation
* `spinning` represents the control is working, such as waiting for the Geolocation API to get a GPS signal or get a response from a location service.

### Icon
* `locked` (dot in centre) represents that the control is locked onto a location. If the devices location changes then the map's camera will be updated to follow. If `showMarker` is enabled, then the location marker will be updated.
* `not locked` (so dot in centre) the user has manually changed the map's camera so if the device's location changes then the map's camera will not be updated to follow. If `showMarker` is enabled, then the location marker will be updated.

## State Diagram
geolocate-control-state-diagram

## Launch Checklist

  - [ ] post benchmark scores
 - [x] manually test the debug page -> I've done a fair bit of manual testing with various configurations of the control and user interactions
 - [x] clean up debug console.log code. It was handy for development and testing but should be okay to be removed. I can't find much other debug logging in GL JS so I assume it's best left out?
 - [x] ~~~once #3773 is merged, I'll need to apply the same fixes in new code here~~~
 - [x] This is independent of #3737 but I think I'll need to manually rebase either this PR or #3737 depending which is merged first.
 - [x] implement a feature so that if there is a Geolocation API error when watching the position, the marker fades to indicate that it may not be the current position.
- [x] fix/decide on a solution for if non-user camera changes (eg. by the application via the api) should transition to background state
- [x] fixes a bug in #3739 where watchPosition didn't stop watching when the control was removed from the map
 - [x] should the state changes to the control fire public events like the `geolocate` and `error` events?
- [ ] fitBoundsOptions vs AnimationOptions and future profing for the new camera api -> https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r96941691
- [ ] Injecting new Layers & Sources into the Style -> https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r96941336

## Future

- [ ] I'd like to have a dim circle around the location marker indicating accuracy. [mapbox-gl-native implements this](https://github.com/mapbox/mapbox-gl-native/blob/7cf10ee524b39d3fbcd0414ed392c4deac8677b2/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java#L258). Because `circle-radius` is specified in pixels we need to update the `circle-radius` every time the position's latitude changes or the camera's zoom changes. An alternative is to vectorise a circle as a polygon. Ultimately it might be better to wait for #2525 to specify the `circle-radius` in `meters` and use a common code path to update the size in pixels for the latitude and zoom.
- [ ] when the reported position contains orientation, use an arrow as a marker rather than a dot
- [ ] add an extra lock orientation state when the device supports it


planemad:
@andrewharvey amazing design documentation!

> locked (dot in centre) represents that the control is locked onto a location.

An additional state to consider adding is an orientation lock. This is a feature I use often while navigating in maps.me to get my bearings right.

lucaswoj:
Amazing @andrewharvey! Let me know if I can help.

Due to an upstream dependency change you'll need to rebase on master to get unit tests passing again.

andrewharvey:
> An additional state to consider adding is an orientation lock. This is a feature I use often while navigating in maps.me to get my bearings right.

geolocate-control-state-diagram-orientation

I'm thinking that could work something like this.

You could unlock those states of the control with `orientMap: true`. When an `orientation` is provided from the Geolocation API, an additional state is provided allowing users to have the map rotation updated based on their orientation.

I see this as working in tandem with another option `markerOrientation` which would add to the location marker a circle segment to show the device orientation (which could be present for any state of the control where the marker is shown).

Although I'm tempted to leave these out of this PR to get it in a review/mergeable state before adding new features.

planemad:
Beautiful @andrewharvey, been waiting for this in GL JS for quite a while.

andrewharvey:
I'm still working on improvements, cleaning up the code, more manual testing and more unit tests, but I have:

* integrated a mock geolocation library to add unit tests

* addressed this point:
> - [x] fix/decide on a solution for if non-user camera changes (eg. by the application via the api) should transition to background state

I've decided that any action by the user which changes the camera should change to background state AND any map api call which changes the camera should also change to background state. The reasoning is that if you change the camera using either method it's because you want that area to be shown and it seems counter that if the camera is immediately brought back to the device location when using watchPosition. I've used the eventData option to skip this for the internal jumpTo camera change.

* added a new `background` and `active_lock` event- fired when the control changes into background state and active lock respectively. I haven't added events for all states, I think these are the most useful. It should be fairly straight forward to add them later if a good use case or demand comes up.

lucaswoj:
Looks good! Thanks @andrewharvey. Let us know when you're done working on this PR and we'll get to work merging it 😄

andrewharvey:
@lucaswoj I've deleted the debugging console.log's. This PR ready for review now.

My preference is for #3737 to be merged first so I can rebase the changes into here.

js/ui/control/geolocate_control.js
```diff
@@ -40,14 +47,28 @@ function checkGeolocationSupport(callback) {
...
  * @param {Object} [options]
- * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
- * @param {Object} [options.watchPosition=false] If `true` the map will reposition each time the position of the device changes and the control becomes a toggle.
+ * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
+ * @param {Object} [options.fitBoundsOptions={maxZoom: 18}] A [`fitBounds`](#Map#fitBounds) options object to use when the map is panned and zoomed to the device location. The default is to use a `maxZoom` of 18 to limit how far the map will zoom in for very accurate locations.
+ * @param {Object} [options.watchPosition=false] If `true` the Geolocate Control becomes a toggle button and when active the map will receive updates to the device location as it changes.
+ * @param {Object} [options.showMarker=true] By default a marker will be added to the map with the device's location. Set to `false` to disable.
```
lucaswoj: The word "Marker" already has a [different specific meaning within GL JS](https://www.mapbox.com/mapbox-gl-js/api/#Marker). For consistency with GL Native, how about calling this option `showUserLocation` and the preceding option `trackUserLocation`?

js/ui/control/geolocate_control.js
```diff
@@ -40,14 +47,28 @@ function checkGeolocationSupport(callback) {
...
+ * @param {Object} [options.markerShadowPaintProperties={ 'circle-radius': 14, 'circle-color': '#000', 'circle-opacity': 0.5, 'circle-blur': 0.4, 'circle-translate': [2, 2], 'circle-translate-anchor': 'viewport' }] A [Circle Layer Paint Properties](https://www.mapbox.com/mapbox-gl-style-spec/#paint_circle) object to customize the device location marker, used as a "shadow" layer. The default is a blurred semi-transparent black shadow.
+ * @param {Object} [options.markerStalePaintProperties={'circle-color': '#a6d5e5', 'circle-opacity': 0.5, 'circle-stroke-opacity': 0.8}] A [Circle Layer Paint Properties](https://www.mapbox.com/mapbox-gl-style-spec/#paint_circle) object applied to the base markerPaintProperties to customize the device location marker in a stale state. The marker is stale when there was a Geolocation error so the previous reported location is used, which may no longer be current. The default is a faded blue dot with a white stroke.
```
lucaswoj: Do you think there are scenarios where injecting a new source & layers into the map's style could have unintended side effects? Are there any ways we could mark the user's location without needing to mutate the style?

js/ui/control/geolocate_control.js
```diff
@@ -40,14 +47,28 @@ function checkGeolocationSupport(callback) {
...
+ *
  * @implements {IControl}
  * @param {Object} [options]
- * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
- * @param {Object} [options.watchPosition=false] If `true` the map will reposition each time the position of the device changes and the control becomes a toggle.
+ * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
+ * @param {Object} [options.fitBoundsOptions={maxZoom: 18}] A [`fitBounds`](#Map#fitBounds) options object to use when the map is panned and zoomed to the device location. The default is to use a `maxZoom` of 18 to limit how far the map will zoom in for very accurate locations.
```
lucaswoj: The relationship between the `GeolocationControl` and `fitBounds` might not be intuitive to a non-expert user. Additionally, we may remove the `fitBounds` method in the near future #2801. Would you be open to refactoring to expose seperate [`animationOptions`](https://www.mapbox.com/mapbox-gl-js/api/#AnimationOptions) and `maxZoom` properties?

js/ui/control/geolocate_control.js
```diff
@@ -139,26 +448,65 @@ class GeolocateControl extends Evented {
...
+ * @event ready
+ * @memberof GeolocateControl
+ * @instance
+ *
+ */
+
+/**
+ * Fired when the Geolocate Control changes to the active_lock state, which happens either when we first obtain a successful Geolocation API position for the device (a geolocate event will follow), or the user clicks the geolocate button when in the background state which uses the last known position to recenter the map and enter active_lock state (no geolocate event will follow unless the device's location changes).
+ *
+ * @event active_lock
```
lucaswoj: We use `camelCase` for all event names

package.json
```diff
@@ -68,8 +68,10 @@
     "jsonlint": "^1.6.2",
     "lodash.template": "^4.4.0",
     "minifyify": "^7.0.1",
+    "mock-geolocation": "^1.0.11",
     "npm-run-all": "^3.0.0",
     "nyc": "^8.3.0",
+    "object.map": "^0.2.0",
```
lucaswoj: We're working really hard to pare down our external dependencies. We have an unmaintainable number of them, contributing to very slow installs, complexity, and occasional upstream bugs. Would you mind rewriting this functionality in yourself in our `util` module or doing the object map using built-in primitives? I apologize -- I know this is a bummer.

js/ui/control/geolocate_control.js
andrewharvey: The fitBoundsOptions is optional so non-expert users can ignore it but it's inclusion lets users override how the camera moves to the users location which is desirable.

I'm excited by #2801, but I thought it was still a proposal and I don't want to unnecessarily make it a dependency before being able to merge this. My intention has always been to get this merged, then when #2801 happens I'll refactor. That's my preference.

js/ui/control/geolocate_control.js
andrewharvey: :+1:

js/ui/control/geolocate_control.js
andrewharvey: Certainly this isn't the cleanest approach.

1. Style/Layer IDs could conflict, I've tried to reduce the likelihood of prefixing with a _ and using longer more specific names. I believe gl-draw has the same issue.

2. The layers get inserted as top layers but if the user adds more layers to the map later, but default the user layers will go on top of the geolocation marker layer.

3. If the application does a setStyle and replaces the style it would mess this control up.

I think the solution to these issues is to use a private style separate to the public one such that name conflicts won't occur, it can be set to always go on top of the public style, and it would be unaffected by the public setStyle method. Unfortunately that's a deeper change and I would be looking to you for advice on what approach makes sense.

package.json
andrewharvey: No problem, I've replaced it.

js/ui/control/geolocate_control.js
lucaswoj: The concept of the [`AnimationOptions` interface](https://www.mapbox.com/mapbox-gl-js/api/#AnimationOptions) exists outside of #2801. Using the name `animationOptions` now creates a naming convention that focuses on behaviour rather than implementation and resilient to change when #2801 lands (in progress at https://github.com/mapbox/mapbox-gl-js/pull/3583)

js/ui/control/geolocate_control.js
lucaswoj: GL Native's implementation works well. In that implementation I believe we draw the user location in the GL context separate from our rendering of the stylesheet. We also have the option of using an HTML overlay.

js/ui/control/geolocate_control.js
andrewharvey: Looking through the new camera api (at least based on the current API docs in #3583) the new API call would be:

map.fitBounds(bounds, cameraOptions, animationOptions)

It would make sense to have the GeolocateControl have and pass through `options.cameraOptions` and `options.animationOptions`. (and [have cameraOptions support a maxZoom](https://github.com/mapbox/mapbox-gl-js/pull/3583#issuecomment-274367505))

However at the moment `fitBounds` currently supports a few extra options which the `AnimationOptions` interface doesn't. `linear`, `padding`, `maxZoom`. So would you suggest I have an `options.animationOptions` and `options.linear` and `options.padding` and `options.maxZoom`? And then once the new camera api lands change the API to bring `linear` into AnimationOptions as `type` and `padding`,`maxZoom` into `options.cameraOptions`?

Either way there will be a public API change, but the new camera API already does that, so I think it's simpler to leave as is at the moment and then refactor for the new camera api.

js/ui/control/geolocate_control.js
andrewharvey: Okay.

Could you point me to the drawCircle module? [I can't find it](https://github.com/mapbox/mapbox-gl-js/search?q=drawCircle&type=Code&utf8=%E2%9C%93).

Obviously the current approach is the simplest to implement, but I understand the concerns. In light of this, I think the best alternative is for this control to create a [new style layer](https://github.com/mapbox/mapbox-gl-js/issues/281#issuecomment-236645859) "userLocation" to create something like

selection_444

Hopefully then then it would have paint properties to customize the design. What do you think? If we went this way it would make sense to share that same new style layer with native.

andrewharvey: I think the usefulness of having this in GL JS, outweighs some of the immediate downsides to the current approach. mapbox-gl-draw suffers the same issues, it injects Layers and Sources into the style just like this PR.

I'm certainly leaning towards a new UserLocation Layer in the style spec as a feature enhancement, this lets the designer control the UserLocation dot together with the map style, shared across platforms.

andrewharvey:
Just tracking the outstanding items to resolve for this PR:

- [ ] fitBoundsOptions vs AnimationOptions and future profing for the new camera api -> https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r96941691
- [ ] Injecting new Layers & Sources into the Style -> https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r96941336

js/ui/control/geolocate_control.js
lucaswoj: > shared across platforms.

Unfortunately not. GL Native has hardcoded user-location drawing code.

lucaswoj: Between the importance of conceptual parity with GL native, the potential for the user to overwrite the user location layer, and the potential for the user to be surprised by the inclusion of the user location layer when inspecting the style, I feel strongly that we should draw the user dot by means other than layer injection.

1ec5: On iOS, the user location annotation view was originally implemented as a custom view atop the map view independent of anything else. The analogy in GL JS would’ve been an HTML element atop the map.

At the time, we had the option of implementing it as a point annotation, which would take advantage of a gl-native feature that unconditionally injects a layer into every style on load. However, we opted for a custom view because it was much more straightforward (and more efficient) to implement the pulsing halo effect implicitly (declaratively) with Core Animation than to drive the animation with a timer.

When we finally implemented view-backed annotations (analogous to `Marker`s in GL JS) as views tied to annotations with transparent spacer images, we migrated the user location annotation view to be an annotation view, taking advantage of both the platform infrastructure around animation and the mbgl infrastructure around annotations.

[This document](https://www.mapbox.com/ios-sdk/api/3.4.1/adding-points-to-a-map.html) outlines the pros and cons of each of the three methods for displaying a point geometry on the map.

andrewharvey: I tried to prototype how this would look using an HTML Marker as the user location dot. However at the moment it suffers from the following due to #3770 :thinking:

* Marker's only show up on world 0 which can lead to the userLocation dot not appearing when in another world. I believe #2071 would resolve this issue.


* This also means on zoomed out maps you'd expect the userLocation dot to show up multiple times in each visible world but the way Marker's are implemented currently they can only appear in a single world.

Although by default the map will zoom into to show the userLocation, the userLocation dot still shows up when zooming out.

andrewharvey: Okay, I've got my prototype of the HTML Marker implementation at https://github.com/andrewharvey/mapbox-gl-js/tree/geolocate-track-show-user-location-html-marker

Haven't updated any tests though, would appreciate some feedback on if this approach is likely to be acceptable or not before I spend time on tests.

Thanks to @tristen for the userLocation Marker CSS from https://github.com/osmlab/census/blob/gh-pages/css/site.css

It breaks down when flyTo crosses the anti-meridian putting the view in world -1. I think the only solution is to implement https://github.com/mapbox/mapbox-gl-js/issues/2071#issuecomment-180595073

I've noticed the Marker wobbles a few pixels when zooming in.

andrewharvey: The Directions plugin also injects the style, which suffers the same issues. Would it be better if the Geolocate control was a plugin too?

lucaswoj: @andrewharvey Yes, I'd be much more comfortable with style injection if this were a plugin 😄

andrewharvey:
Cutting from a thread above at https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r98823742 the HTML Marker approach breaks down when flyTo crosses the anti-meridian putting the view in world -1. I think the only solution is to implement https://github.com/mapbox/mapbox-gl-js/issues/2071#issuecomment-180595073.

@lucaswoj Would you ship an HTML Marker version of this PR even with that major bug? Or is this blocked by it?

lucaswoj:
@andrewharvey This PR is blocked until we can figure out a way to draw the user location dot without any major caveats. It needs to work across wrapped worlds, persist across style changes,  and not modify the user's style. I would be comfortable with any of these caveats in a plugin but core requires a higher bar.

andrewharvey:
I'm closing this PR since, the branch based on the HTML Marker over at #4479 is going to be the best way to get this landed in core.

Issue #1513, opened 2015-09-22
Popular Mapbox.js example: draw & animate a line on a map


lyzidiamond:
No. 3 most popular Mapbox.js example by analytics in the past year

https://www.mapbox.com/mapbox.js/example/v1.0.0/dynamically-drawing-a-line/

YunjieLi:
Happy to work on this example

Curious why [this commit](https://github.com/mapbox/mapbox-gl-js/commit/423b2e1e8d3a794ef6f3179f51cf865c882f3f71) did not make its way to an example but may start from there.

cc @geografa @tristen

tristen:
@YunjieLi Oh! I forgot about this and it's unfinished. Go for it!

geografa:
@YunjieLi nice! From support questions, we get two types of requests for this kind of example - animating the line and making the line disappear after a specific distance or time similar to the [Courier example](https://www.mapbox.com/bites/00115/).

YunjieLi:
Thanks @geografa ! 
It's easy to stop after a specific time. But it will need `turf` for a certain distance or along a specific route. I'll give that a shot and see how it works.

I'll definitely add a play/pause button too.

mcurtiswilliams:
Hi,
Just wondering if there's an update on this, as I'm looking into this for a project I'm working on.
Thanks!

YunjieLi:
@mcurtiswilliams this just went alive!

Issue #3692, opened 2016-11-23
Automatically batch the work done by GeoJSON#setData()


lucaswoj:
If is relatively common for users to animate geometries on the map using `GeoJSONSource#setData`

 It can be hard to do so performantly and without causing the queue of data update operations to become seriously clogged.

We could make this use of GL JS easier via a new`Map#requestSetDataFrame(sourceId: string, callback: Function)` method (analogous to `requestAnimationFrame`). 

```js
    function requestSetDataFrame(map, sourceId, callback) {

      var countdown = 2;

      function dataCallback() {
          if (map.style.sourceCaches[sourceId].loaded()) {
              if (--countdown === 0) callback(performance.now());
          } else {
              map.once('data', dataCallback);
          }
      }

      dataCallback();

      requestAnimationFrame(function(now) {
          if (--countdown === 0) callback(now);
      });

   }
```

lucaswoj:
cc @kronick

kronick:
This sounds great and addresses the issue I ran into pretty quickly once I started trying out some relatively complicated animations.

Just thinking about the API for this, ideally I would be able to do:

    function animate() {
        // update myData
        source.setData(myData, function() {
            // "Data is updated" callback
            requestAnimationFrame(animate);
        });
    }

Would it be possible to integrate the `requestDataUpdateFrame()` code like this? Otherwise, given @lucaswoj's code above, I picture:
   
        function animate() {
             // update myData

             map.requestDataUpdateFrame("sourcename", function() {
                 requestAnimationFrame(animate);
             });
        
             source.setData(myData);
        }

Which is awkward 1) because for correctness you should call `requestDataUpdateFrame()` _before_ `setData()` (to avoid a race condition that I think is possible if you call it the other way around) and 2) I can't think of a use case where you would want to call `requestDataUpdateFrame()` without also calling `setData()`

lucaswoj:
> Would it be possible to integrate the `requestDataUpdateFrame()` code like this? Otherwise, given @lucaswoj's code above, I picture:

This would be possible syntactically if `GeoJSONSource#setData` accepted a callback. That is a separate discussion (one which would be valuable to have!)

Conceptually, this nesting would not result in the optimal animation loop. In the worst case, the time between calls to `GeoJSONSource#setData` would be `frame time + data update time` whereas we want the worst case to be `max(frame time, data update time)`. 

[...]

PR #3788, opened 2016-12-13
[WIP] Add example: animate a line

YunjieLi:
Per #1513 I created a new example of animating a line.

I rewrote the [Mpabox.js example](https://www.mapbox.com/mapbox.js/example/v1.0.0/dynamically-drawing-a-line/) using GL JS and requestAnimationFrame so it's smoother. I also add the play and stop buttons which is not in the [point animation example](https://www.mapbox.com/mapbox-gl-js/example/animate-point-along-line/).

[Here is a demo](https://bl.ocks.org/YunjieLi/5919254f2e8a517a2c7350da4598bccb).

Still working on:
[ ] Slight flickering when I stop and resume animation
[ ] Bug in leaving the tab inactive and come back
[ ] Pan the map to follow animation: speed control?

cc @geografa @mollymerp


docs/_posts/examples/3400-01-31-animate-a-line.html
```diff
@@ -0,0 +1,128 @@
+--- layout: example category: example title: Animate a line description: Animate a line by updating a GeoJSON source on each frame. tags: - layers - sources ---
```
mollymerp: this section needs to be formatted like the other examples (not all on one line, etc)

docs/_posts/examples/3400-01-31-animate-a-line.html
```diff
@@ -0,0 +1,128 @@
...
+#play.pause::after {
+  content: 'Play';
+}
+</style>
+
+<div id='map'></div>
+<button id='play' class='pause'></button>
+
+<script>
+mapboxgl.accessToken = 'pk.eyJ1IjoieXVuamllbGkiLCJhIjoiY2lxdmV5MG5rMDAxNmZta3FlNGhyMmpicSJ9.CTEQgAyZGROcpJouZuzJyA';
```
mollymerp: you don't need an access token here. a token is automatically added by jekyll when it compiles the examples

docs/_posts/examples/3400-01-31-animate-a-line.html
```diff
@@ -0,0 +1,128 @@
...
+    'source': 'line-animation',
+    'layout': {
+      'line-cap': 'round',
+      'line-join': 'round'
+    },
+    'paint': {
+      'line-color': '#3cc',
+      'line-width': 5,
+      'line-opacity': .4
+    }
```
mollymerp: This is hard to see on the map -- I would make it a brighter color or more opaque and/or change the basemap.

docs/_posts/examples/3400-01-31-animate-a-line.html
```diff
@@ -0,0 +1,128 @@
...
+    // append new coordinates to the lineStrin
+    // then update the map
+    geojson.features[0].geometry.coordinates.push([x, y]);
+    map.getSource('line-animation').setData(geojson);
+
+    // Fly the map to follow the animation
+    // map.panBy([60, 0]);
+
+    // Request the next frame of the animation.
+    animation = requestAnimationFrame(animateLine);
```
lucaswoj: There's some interesting discussion about using using `requestAnimationFrame` with `GeoJSONSource#setData` in https://github.com/mapbox/mapbox-gl-js/issues/3692

docs/_posts/examples/3400-01-31-animate-a-line.html
```diff
@@ -0,0 +1,128 @@
...
+    // to cancel the animation:
+    if (this.classList.contains('pause')) {
+      cancelAnimationFrame(animation);
+      // to play the animation
+    } else {
+      // resume previous progress if progress != 0 
+      startTime = performance.now() - progress;
+      animation = requestAnimationFrame(animateLine);
+    }
+  });
```
lucaswoj: Would you be open to removing the "play" button in order to reduce the # of lines of code in this example?

docs/_posts/examples/3400-01-31-animate-a-line.html
```diff
@@ -0,0 +1,128 @@
...
+        }]
+      };
+    }
+
+    progress = time - startTime;
+    // speedFactor controls the animation speed.
+    var x = progress / speedFactor;
+    // draw a sine wave with some math.
+    var y = Math.sin(x * Math.PI / 90) * 40;
+    // append new coordinates to the lineStrin
```
lucaswoj: You're missing a "g" on "line string"

docs/_posts/examples/3400-01-31-animate-a-line.html
```diff
@@ -0,0 +1,128 @@
...
+            "type": "LineString",
+            "coordinates": [
+              [0, 0]
+            ]
+          }
+        }]
+      };
+    }
+
+    progress = time - startTime;
```
kronick: This is the source of the inactive tab bug-- when on another tab, requestAnimationFrame does not get called, so your progress variable makes a big jump when you return and you don't add any points in between. Consider something like this to limit the maximum timestep between frames:

    var startTime;
    var lastTime;
    var progress = 0;

    function animateLine(time) {
      var dT = time - lastTime;
      if(dT > 30) dT = 30; // Don't jump too far between animation frames, even if the browser tab is inactive
      progress += dT;
      lastTime = time;
   
      // The rest of what you have should work the same

    }

docs/_posts/examples/3400-01-31-animate-a-line.html
YunjieLi: Good catch! I forgot to remove it from the gist

docs/_posts/examples/3400-01-31-animate-a-line.html
YunjieLi: I added it according to @geografa's [comment](https://github.com/mapbox/mapbox-gl-js/issues/1513#issuecomment-266136120) that user wants more control over the animation. But I do think I can simplify into just a 'play/replay' button + no looping for animations. Do you think that would make it clean enough?

docs/_posts/examples/3400-01-31-animate-a-line.html
YunjieLi: Thanks @kronick @tmcw! I'll research into both suggestions a little more. I think `visibilitychange` sounds more straightforward but am not sure if there are other things that can interrupt `requestAnimationFrame` besides inactive tabs.

docs/_posts/examples/3400-01-31-animate-a-line.html
```diff
@@ -0,0 +1,122 @@
...
+    };
+    // play animation
+    animateLine();
+  });
+
+  function animateLine(timestamp) {
+
+    // requestAnimationFrame pauses when the tab is inactive
+    // reset startTime and progress when switched back
+    document.addEventListener('visibilitychange', function(){
```
tmcw: this call shouldn't be inside of `animateLine`: `animateLine` will be called many times but `document.addEventListener` should be called once.

YunjieLi: Thanks @tmcw !  I just pushed some new edits if you want to take a look again?

YunjieLi:
@mollymerp which branch should I rebase this to?

mollymerp:
@YunjieLi you can rebase against `mb-pages` 👍

YunjieLi:
Also - currently the map shows a straight line between the starting point and x,y by the time the window is loaded, versus not starting animation until the map is ready(?). Will look into it next.
image

YunjieLi:
Somehow I couldn't get this running locally anymore, so moving to https://github.com/mapbox/mapbox-gl-js/pull/4780!

PR #3790, opened 2016-12-13
clean up matrix creation

ansis:
This cleans up a bunch of matrix and antialiasing code.

- removes the confusing `altitude` and replaces it with `fov`
- fixes text blurriness at non-default FOVs
- removes the hack added in #3740 and adds a different fix
- makes pitched line antialiasing clearer and removes `lineStretch` and `lineAntialiasingMatrix`

related prs:
https://github.com/mapbox/mapbox-gl-shaders/pull/41
https://github.com/mapbox/mapbox-gl-test-suite/pull/193

benchmark | master 32fcbfe | matrix-cleanup d2f5d09
--- | --- | ---
**map-load** | 207 ms  | 152 ms 
**style-load** | 95 ms  | 94 ms 
**buffer** | 1,027 ms  | 1,001 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 4.5 ms, 0% > 16ms  | 4.6 ms, 0% > 16ms 
**query-point** | 1.00 ms  | 1.14 ms 
**query-box** | 85.19 ms  | 79.48 ms 
**geojson-setdata-small** | 7 ms  | 5 ms 
**geojson-setdata-large** | 299 ms  | 286 ms 

## Launch Checklist

 - [x] briefly describe the changes in this PR
 - [x] write tests for all new functionality
 - [x] document any changes to public APIs
 - [x] post benchmark scores
 - [x] manually test the debug page

TODO:
- [x] debug the render test that's failing on ci but not locally
- [x] open issue about porting to -native

@lucaswoj @mourner

PR #3797, opened 2016-12-14
Clean up gamma scale

lucaswoj:
Continued from https://github.com/mapbox/mapbox-gl-shaders/pull/41

> First commit fixes text blurriness at non-default FOVs and removes a magic constant.
>
> Second commit changes how the line antialiasing adjustment is calculated. It now projects the extrusion and compares it's projected pitched pixel length with the length it would have in a non-pitched view. It drops `u_extra` and `u_antialiasing_matrix`.
>
> @kkaefer @lucaswoj 

cc @ansis

### DIFF
--- a/shaders/line.vertex.glsl
+++ b/shaders/line.vertex.glsl
@@ -17,9 +17,8 @@ attribute vec4 a_data;
 
 uniform mat4 u_matrix;
 uniform mediump float u_ratio;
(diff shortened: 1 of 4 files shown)


ansis:
merged as part of https://github.com/mapbox/mapbox-gl-js/pull/3790

PR #3815, opened 2016-12-16
Use polyfill for Math.sign, which is unsupported in IE11. Fix #3814.

### LINKED ISSUE #3814: Math.sign not supported in IE11
A usage of `Math.sign` was _just_ introduced in https://github.com/mapbox/mapbox-gl-js/commit/7e254fa03fd5081ff964d9b9c9280f9660cc3aa5#diff-b3089558a25dee1e74f59aa1829490e4R115, and unrelatedly I turned on a Windows computer for the first time in years to debug some Windows bugs and found out `Math.sign` is not supported in IE11, so this breaks. [(Here is a very complicated table that says as much.)](https://kangax.github.io/compat-table/es6/#test-Math.sign) We should replace it with boring-er Javascript until IE catches up 😞

lbud:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/3814: `Math.sign` is not supported in IE11.

### DIFF
--- a/js/symbol/shaping.js
+++ b/js/symbol/shaping.js
@@ -2,7 +2,6 @@
 
 const scriptDetection = require('../util/script_detection');
 const verticalizePunctuation = require('../util/verticalize_punctuation');
-const util = require('../util/util');
 
 
 const WritingMode = {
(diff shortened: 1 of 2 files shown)


lbud:
:doh: thanks @jfirebaugh — should have been doing less blind replacing and more reading. Fixed — 🍏 ?

PR #3822, opened 2016-12-17
Fix for Windows browsers capturing not current but previous right clicks

### LINKED ISSUE #3803: contextmenu event not fired at first click in windows browsers
**mapbox-gl-js version**: v0.21.0

I'm unable to catch "contextmenu" or rightclick event in windows browsers (tested chrome, edge, IE, FF). It is needed to click twice, and even after second click, lng lat coordinations are from the old event.
I have tried it in this simple example:
https://jsfiddle.net/6ctfssy9/1/

### Expected Behavior
Fire event at first right click

### Actual Behavior
Fires at second (that second click can be also left click)

lbud:
Fixes #3803.

Browsers on OSX fire `contextmenu` events on `mousedown`, whereas Windows browsers fire it on `mouseup`. The way we've been capturing `contextmenu` events is to preserve them on `contextmenu` and fire them on `mouseup` in order to produce consistent behavior between Mac and Windows. However, Windows browsers consistently fire `mouseup` _before_ `contextmenu`, resulting in #3803.

This fixes #3803 by preserving a `mousedown` boolean so that we can determine when to fire `contextmenu` events on `contextmenu` (which we want to do on Windows).

### DIFF
--- a/js/ui/bind_handlers.js
+++ b/js/ui/bind_handlers.js
@@ -17,7 +17,6 @@ module.exports = function bindHandlers(map, options) {
     const el = map.getCanvasContainer();
     let contextMenuEvent = null;
     let mouseDown = false;
-    let mapRotating = false;
     let startPos = null;
     let tapped = null;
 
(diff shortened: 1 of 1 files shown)


mollymerp:
💯🕵👏

PR #3830, opened 2016-12-19
Remove unused variables and imports

diegomura:
Both `Popup` and `util` were imported but never used in this class.
Also, a `that` variable was being created on `setPopup` without any apparent reason.

### DIFF
--- a/js/ui/marker.js
+++ b/js/ui/marker.js
@@ -2,10 +2,8 @@
 'use strict';
 
 var DOM = require('../util/dom');
-var util = require('../util/util');
 var LngLat = require('../geo/lng_lat');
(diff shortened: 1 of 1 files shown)


diegomura:
Thanks!

mourner:
I'm puzzled why this wasn't caught by current ESLint setup. @mollymerp can you investigate?

jfirebaugh:
The file has `/* eslint-disable */` at the top for some reason.

mourner:
@jfirebaugh ah, I probably added it during development and then forgot to remove.

mollymerp:
interesting -- i'll fix the eslint errors on this file and then make a separate PR.

PR #3832, opened 2016-12-19
Unbind events after dragging ends

diegomura:
Mouse callbacks are being created every time the `mousedown` event is triggered and never destroyed.

I think `once` method can be used to bind the `mouseup` event (after all, you want to call it once per drag), and unbind the `mousemove` normally when the point was dropped.


lucaswoj:
Cherry-picked to `mb-pages` as https://github.com/mapbox/mapbox-gl-js/commit/004804dd19fa7f89aa21d02414a6abc0e15c850c

diegomura:
Thanks!

Issue #2893, opened 2016-07-19
Issue an error when map canvas is larger than allowed by gl.MAX_RENDERBUFFER_SIZE


lucaswoj:
Once the map canvas is larger than `gl.MAX_RENDERBUFFER_SIZE / 2`, scaling behaviour starts to occur. I do not think we should invest significant resources into accommodating this case. 

See https://github.com/mapbox/mapbox-gl-js/issues/2673 and https://github.com/mapbox/mapbox-gl-js/issues/2863

stepankuzmin:
Hi @lucaswoj! I'd like to help with some issues at my spare time. Is it correct, that the Error need to be thrown when `canvas.width` or `canvas.height` is larger than `gl.MAX_RENDERBUFFER_SIZE / 2` [like this](https://github.com/stepankuzmin/mapbox-gl-js/commit/c14d390b704aa0d48f3169396f048a060ae3da37)?

lucaswoj:
Thank you thank you thank you @stepankuzmin. 🙇 

Your commit is on the right track. I'll leave my feedback as comments on that commit.

stepankuzmin:
I've made a [PR](https://github.com/mapbox/mapbox-gl-js/pull/3841).

1ec5:
This change was reverted in #3986 because it caused #3915 in Firefox. Reopening.

kristfal:
Just wanted to add that this change broke iPhone (6,6s,7) Plus and iPads with Retina Display support when the map was initialized as a full screen map. 

iOS have a MAX_RENDERBUFFER_SIZE of 4096, so please make sure that that it works with resolutions up and including 2732px.

stepankuzmin:
We still need to issue an error or not?

lucaswoj:
@stepankuzmin I'm going to add a warning message instead of an error. There are some systems in which we're firing an error and unnecessarily breaking functional maps.

andrewharvey:
@lucaswoj I can no longer replicate the two original issues #2673 and #2863 when using 0.29.0. If others can confirm this, then I don't think we need a warning for `gl.MAX_RENDERBUFFER_SIZE / 2`.

PR #3841, opened 2016-12-20
Issue an error when map canvas is larger than allowed by gl.MAX_RENDERBUFFER_SIZE #2893

stepankuzmin:
Issue an error when map canvas is larger than allowed by gl.MAX_RENDERBUFFER_SIZE #2893

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -1051,6 +1051,15 @@ class Map extends Camera {
             return;
         }
 
+        const MAX_RENDERBUFFER_SIZE = gl.MAX_RENDERBUFFER_SIZE / 2;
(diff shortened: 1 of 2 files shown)


andrewharvey:
Should the error message mention the halving factor? If I create a map canvas 5000x5000 and gl.MAX_RENDERBUFFER_SIZE is 8000 then seeing the error message "Map canvas is larger than allowed by gl.MAX_RENDERBUFFER_SIZE" would be a bit confusing for me as 5000 isn't larger than 8000.

stepankuzmin:
Hi @andrewharvey! I've updated error message with mentioning canvas size and the `gl.MAX_RENDERBUFFER_SIZE` halving factor.

PR #3853, opened 2016-12-22
Optimize layer update

mourner:
This change makes the worker avoid recreating all layer instances when updating just a few layers (e.g. with `setFilter`). 

This fixes performance for cases where the updated source doesn't contain much data, but the update is bogged down by all other layers from the basemap — e.g. as on our [official hover example](https://www.mapbox.com/mapbox-gl-js/example/hover-styles/). Ref #2874 (comments). 

cc @jfirebaugh @lucaswoj

### DIFF
--- a/js/style/style_layer_index.js
+++ b/js/style/style_layer_index.js
@@ -6,29 +6,29 @@ const featureFilter = require('feature-filter');
 const groupByLayout = require('mapbox-gl-style-spec/lib/group_by_layout');
 
 class StyleLayerIndex {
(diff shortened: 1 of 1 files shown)


zuo1188:
it may cause new bugs for applications using the 'setStyle' method. it refresh sprite any more

mourner:
@zuo1188 if you're experiencing a specific issue, please submit it as a new issue to GitHub (with clear steps to reproduce and a minimal test case).

Issue #3857, opened 2016-12-23
In-place source addition in addLayer


mourner:
Let's implement @lucaswoj's suggestion in https://github.com/mapbox/mapbox-gl-style-spec/issues/162#issuecomment-268997733. Marking it as a starter task.

lucaswoj:
Done in #3861 🎉

PR #3861, opened 2016-12-25
add in-place source addition support in addLayer #3857

stepankuzmin:
Hi there, @mourner and @lucaswoj! Is this a correct approach to #3857?

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -427,7 +427,7 @@ class Style extends Evented {
 
         if (typeof layerObject.source === 'object') {
             this.addSource(id, layerObject.source);
-            layerObject.source = id;
+            layerObject = util.extend(layerObject, { source: id });
         }
 
         // this layer is not in the style.layers array, so we pass an impossible array index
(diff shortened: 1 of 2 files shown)


mollymerp:
Thanks for your contribution @stepankuzmin!! This looks good to me, but I will let @mourner @lucaswoj give it some 👀 before ✅

stepankuzmin:
Another approach is to support `source` as object in `layer`. Would it be more suitable solution?

lucaswoj:
@mourner Just added a commit which clones rather than mutates `layerObject`. Planning to 🚢 on pass.

stepankuzmin:
Great! Thanks!

lucaswoj:
TY @stepankuzmin 😄

lucaswoj:
I've identified a some follow-ups which I'll ticket out separately:

 - add support for this syntax in the validator
 - port this feature to GL native
 - migrate GL JS examples to use this feature when it reduces boilerplate
 - write documentation about this feature
 - write code to deduplicate identical sources internally

Issue #2597, opened 2016-05-20
Display an error when calling `setFilter` on a non-existent layer ID 


jfirebaugh:
Currently, calling `setFilter` with an ID of a layer that does not exist produces "Uncaught TypeError: Cannot read property 'ref' of undefined". Instead, it should report an error using the same mechanism as when other runtime style API constraints are violated.

Issue #3208, opened 2016-09-16
Make error handling for queryRenderedFeatures consistent


jfirebaugh:
If `queryRenderedFeatures` is passed an invalid filter, it throws. If it's passed an invalid layer ID (one that doesn't exist in the style), it emits an `error` event.

Is there a reason these two errors are reported differently or can we use a consistent error reporting mechanism?

cc @mollymerp @lucaswoj

lucaswoj:
We have decided to always emit, not throw, errors.

lucaswoj:
We should entirely remove the "throws" branch in this code http://github.com/mapbox/mapbox-gl-js/blob/b3bffe06564dcd031feeba715bd80de36e94f990/js/style/style.js#L689-L690

PR #3864, opened 2016-12-26
Throw an error when calling layer operating functions with a non-existent layer ID #2597

stepankuzmin:
Hi @jfirebaugh! Is this a correct approach to #2597?


jfirebaugh:
Thanks for the contribution! The correct approach is to emit an `error` event: #3208, #3879.

stepankuzmin:
Thanks for the reply, @jfirebaugh! I've updated PR with emitting errors instead of throwing.

Issue #3868, opened 2016-12-28
Unexpected behaviour when calling "Map#addLayer" with invalid "before" argument


iskandari:
I am using the Mapbox GL [example](https://www.mapbox.com/mapbox-gl-js/example/setstyle/) change a map's style.

When the style is switched to `dark-v9`, `basic-v9`, or `streets-v9`, existing layers reload with the `map.on('style.load', function () {}`  This is expected. 

However, this is not the case with `satellite-v9` - when the satellite style is toggled, the existing layers load behind the style.  Please see my example [code](http://iskandarblue.github.io/mapbox/code/test_parcels.html)

This is unexpected behavior.  Why does this happen and Is there a workaround?

averas:
The problem is that you are providing a <code>before</code> parameter to addLayer referencing a layer (road-primary) that does exist in some of the styles but _not_ in satellite.

One might assume that this would cause an exception, or at the very least fall back to the default behaviour of adding the layer at the end of the layer stack, however, the current implementation:

```javascript
   const index = before ? this._order.indexOf(before) : this._order.length;
   this._order.splice(index, 0, id);
```

.. will actually place the the layer second to last when there is no hit on the before parameter. In your case that means that your layers are added before the last layer in the satellite style, which is the raster layer. Simply removing the "road-primary" parameter will make your code work: http://codepen.io/anon/pen/mONPJm

Although you did not use the API as documented I would argue that the current behaviour is not logical and can be improved (thrown exception or fallback to last as mentioned above).

stevage:
FWIW (and I realise this may not be helpful feedback), I'm not enjoying this change. I'm often running into situations where an `addLayer` is failing because some other layer isn't present. This happens:

- when I'm temporarily switching my base style out for something that loads faster (now my overlay layers don't get added at all)
- when I'm trying out different base styles
- when for some reason my layers didn't load in the order I anticipated
- when my layers, their names, and ordering are all in a bit of flux

Basically, this change is slowing me down and feels pedantic. I would prefer the "before" param to be taken as a hint ("stick the new layer before layer X if you can, otherwise stick it at the end"). That is, the "...or..." option suggested by the OP here:

>Either a hard fail, or add the layer to the end of the stack with a warning that the specified before doesn't exist.

PR #3869, opened 2016-12-29
Fallback for add/moveLayer when provided before-layer does not exist

averas:
In response to discussion in #3868.

When providing a <code>before</code> parameter to <code>addLayer</code> and <code>moveLayer</code> referencing a layer id that does not exist the current behaviour is to add the layer _second to last_ (since the indexOf lookup returns -1):

```javascript
   const index = before ? this._order.indexOf(before) : this._order.length;
   this._order.splice(index, 0, id);
```

This is not intuitive and this PR will instead make a fallback to adding the layer at the end (i.e. same as without the <code>before</code> parameter provided).

Another option could be to throw an exception in such a case.

### DIFF
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -434,7 +434,8 @@ class Style extends Evented {
 
         layer.setEventedParent(this, {layer: {id: id}});
 
-        const index = before ? this._order.indexOf(before) : this._order.length;
+        const beforeIndex = before ? this._order.indexOf(before) : -1;
+        const index = beforeIndex > -1 ? beforeIndex : this._order.length;
         this._order.splice(index, 0, id);
 
         this._layers[id] = layer;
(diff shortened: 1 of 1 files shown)


mollymerp:
Thanks for your contribution @averas! I think the behavior you're suggesting makes sense, but adding a console warning (`util.warnOnce('warning message')` should work) explaining the behavior would be worthwhile. Could you please also add a unit test that confirms this functionality?

jfirebaugh:
The correct approach is to emit an `error` event: #3208, #3879.

averas:
@jfirebaugh Emit error only, or with fallback as well?

averas:
To be clear, which alternative should I pursue?

1. Emit error and continue with current behaviour, i.e. add layer second to last (I really don't think so) 
2. Emit error (with explanation) and fallback to adding the layer at the end
3. Emit error and bail (do nothing)

jfirebaugh:
I'd say "Emit error and bail" -- that's consistent with what happens if a validation error occurs elsewhere.

mourner:
@averas any updates on this PR?

jfirebaugh:
Hey @averas -- I'm in the process of cleaning out our PR queue and it looks like there hasn't been any activity on this one recently. If you're still interested in coming back to it, please let us know and we'll be happy to help you get it to completion. Until then, I'm going to close it out. Thanks again for your contributions!

Issue #3785, opened 2016-12-12
Option to not wrap the world


Que3216:
We have a world map, in a panoramic container, that starts out fully zoomed out, but the wrapping looks pretty silly (can get 3+ copies of the world): http://jsfiddle.net/3wjybdf5/1/. Would it be possible to add a noWrap option? How hard would it be to add a noWrap option?

Solutions which don't work:
   - Using mapBounds (https://github.com/mapbox/mapbox-gl-js/issues/1488). We want to show the full world, and the map container is panoramic, not square.
   - Shrinking the map container, or changing the shape of the container. As you zoom in we still want the map to fill the the bounds of the panoramic container.

lucaswoj:
This wouldn't be too difficult to add @Que3216. I encourage you to submit a PR 😄

Que3216:
I've opened a PR: https://github.com/mapbox/mapbox-gl-js/pull/3885

Issue #2507, opened 2016-04-28
Add an option to use wrapped coordinates in URL hash


samanpwbb:
**mapbox-gl-js version**:  0.17.0, but present in much older versions as well
### Steps to Trigger Behavior
1. Navigate to a page like https://api.mapbox.com/styles/v1/saman/cimhtyumg001bqxnokqvveyx2.html?title=true&access_token=pk.eyJ1Ijoic2FtYW4iLCJhIjoiS1ptdnd0VSJ9.19qza-F_vXkgpnh80oZJww
2. Pan the map for a while
3. See that the URL changed to look like: https://api.mapbox.com/styles/v1/saman/cimhtyumg001bqxnokqvveyx2.html?title=true&access_token=pk.eyJ1Ijoic2FtYW4iLCJhIjoiS1ptdnd0VSJ9.19qza-F_vXkgpnh80oZJww#1.88/-27.1/-1887.5
### Expected Behavior

Lng value resets to 0 in hash when you pass the edge of the world 
### Actual Behavior

Lng value keeps increasing

ryanbaumann:
+1 on this bug. Longitude does not reset to zero when panning/scrolling across 180 degrees (middle of Pacific Ocean).

Scope:  Impacts map.getBounds() function for sure, important for geospatial queries made from the current map bounds.

lucaswoj:
@samanpwbb @ryanbaumann Can you explain why this behaviour is undesirable to you? Might there be cases where you _do_ want to differentiate between "wrapped" copies of the map? Are you able to normalize these coordinates internally?

samanpwbb:
> Might there be cases where you do want to differentiate between "wrapped" copies of the map? 

Not in any of my use cases, but maybe someone wants this (I am reminded of https://www.youtube.com/watch?v=kpk2tdsPh0A).

Are you able to normalize these coordinates internally?

Yes, this would be possible, but difficult to manage in more complex systems where you're storing and sharing map position coordinates across multiple applications.

Could we default to wrapped coordinates but allow user to pass in an option to accumulate coordinates if it's important for certain types of users?

lucaswoj:
> Yes, this would be possible, but difficult to manage in more complex systems where you're storing and sharing map position coordinates across multiple applications.

What can we do to make this less difficult? Expose some sort of `normalizeCoordinates` utility method? I would rather make it easy to normalize coordinates than complicate GL JS with more configuration options.

[...]

PR #3885, opened 2017-01-04
Added a 'wrapAroundWorld' option to the mapboxgl.Map constructor

Que3216:
If `wrapAroundWorld` is set to `false`, then only one copy of the world will be rendered, when zoomed out. This solves issue https://github.com/mapbox/mapbox-gl-js/issues/3785.

### DIFF
--- a/js/source/source_cache.js
+++ b/js/source/source_cache.js
@@ -317,7 +317,7 @@ class SourceCache extends Evented {
                 maxzoom: this._source.maxzoom,
                 roundZoom: this._source.roundZoom,
                 reparseOverscaled: this._source.reparseOverscaled,
-                wrapAroundWorld: this.style.wrapAroundWorld
+                wrapAroundWorld: this.style ? this.style.wrapAroundWorld : true
             });
         }
 
--- a/js/source/source_cache.js
+++ b/js/source/source_cache.js
@@ -316,7 +316,8 @@ class SourceCache extends Evented {
                 minzoom: this._source.minzoom,
                 maxzoom: this._source.maxzoom,
                 roundZoom: this._source.roundZoom,
-                reparseOverscaled: this._source.reparseOverscaled
+                reparseOverscaled: this._source.reparseOverscaled,
+                wrapAroundWorld: this.style.wrapAroundWorld
             });
         }
 
--- a/js/source/source_cache.js
+++ b/js/source/source_cache.js
@@ -316,8 +316,7 @@ class SourceCache extends Evented {
                 minzoom: this._source.minzoom,
                 maxzoom: this._source.maxzoom,
                 roundZoom: this._source.roundZoom,
-                reparseOverscaled: this._source.reparseOverscaled,
-                wrapAroundWorld: this.style ? this.style.wrapAroundWorld : true
+                reparseOverscaled: this._source.reparseOverscaled
             });
         }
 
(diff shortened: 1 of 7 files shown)


Que3216:
This PR changes the public API. Let me know if the API change looks good, and I'll update https://github.com/mapbox/mapbox-gl-style-spec.

js/source/source_cache.js
```diff
@@ -316,7 +316,8 @@ class SourceCache extends Evented {
                 minzoom: this._source.minzoom,
                 maxzoom: this._source.maxzoom,
                 roundZoom: this._source.roundZoom,
-                reparseOverscaled: this._source.reparseOverscaled
+                reparseOverscaled: this._source.reparseOverscaled,
+                wrapAroundWorld: this.style ? this.style.wrapAroundWorld : true
```
jfirebaugh: Can `wrapAroundWorld` be a property on `transform` (the `update` method parameter), so that `Style` doesn't have to be involved?

Que3216: Yep! That's much nicer. I'll update it.

Que3216: @jfirebaugh I've updated it -- let me know if this looks good now.

Que3216:
@jfirebaugh Is this good to merge now?

jfirebaugh:
The implementation looks good. Before merging I want to brainstorm a few other options for naming in hopes that we can find something more precise and self-descriptive than `wrapAroundWorld`, which could be confused with options such as described in #2507 and #2071.

* In Leaflet, the equivalent option is the [`TileLayer` `noWrap` option](http://leafletjs.com/reference.html#tilelayer-nowrap).
* Keeping it a global option: `renderWorldCopies` or `renderWrappedWorlds`.

lucaswoj:
Looks good! I can see two big picture items before we 🚢 this. 

# Todo before merging

 - [ ] decide on a name ☝️
 - [ ] add a render test

Que3216:
I think it makes more sense as a global option. I like `renderWrappedWorlds`. @lucaswoj / @jfirebaugh - are you happy to go with `renderWrappedWorlds`?

jfirebaugh:
I think `renderWorldCopies` makes more sense than `renderWrappedWorlds`. `LngLat#wrap()` establishes that a "wrapped" longitude is one that is guaranteed to be in the range (-180, 180). This implies there's only one wrapped world, and it's the one that this option _preserves_, as opposed to the "world copies" that it disables.

If you rename to `renderWorldCopies` I'm ready to approve. I'm okay with not having a render test for this, since it's not a style property and there is a unit test.

Que3216:
@jfirebaugh Ok cool! That makes sense. I've renamed it to renderWorldCopies.

cammanderson:
Hi Guys, 

With this new render option, is it possible to expose an API to toggle this behaviour opposed to re-instantiating the map? Or is it necessary? I can see some problems where the lnglat of the current camera our out of bounds due to world wrapping.

Just looking at adding this function to our react library [react-map-gl-alt](https://github.com/AlpacaTravel/react-map-gl-alt) but a goal of the project is to be able to configure the map through external exposed props - which means that the developer would have the option to toggle.

Thanks guys!
Cameron

huaruiwu:
Tried using `flyTo` with `renderWorldCopies: false` and map will try to fly to the closest copy coordinate even if that world is not being rendered.

mollymerp:
hmm @huaruiwu that sounds like a bug -- do you want to open a fresh ticket to report?

huaruiwu:
@mollymerp Sure! https://github.com/mapbox/mapbox-gl-js/issues/4449

PR #1682, opened 2015-11-09
Support padding positions in fitBounds

tristen:
- ref #995 

cc @jfirebaugh

PR #3890, opened 2017-01-04
Allow user to specify n, s, e, w padding options for fitBounds

### LINKED ISSUE #1339: Allow each edge to have different padding when fitting to bounds
[`Camera.prototype.fitBounds()`](https://github.com/mapbox/mapbox-gl-js/blob/37145b8b9bb0de022811ad46f3e89a2714896b18/js/ui/camera.js#L345) accepts `padding` and `offset` options for adding slop area (in pixels) around the fitted bounds. In mapbox/mapbox-gl-native#1783, I’m instead allowing each side of the bounds to have a different padding amount. It would be great to have this in gl-js as well.

Independent edge padding is necessary on iOS, because it’s customary to include about 32 points of top padding to make room for a translucent top bar overlay. In gl-js, it isn’t possible to express this behavior with a combination of `padding` and `offset`: you don’t want any padding to be added in the horizontal direction.

mollymerp:
## Launch Checklist
Continuing on @tristen's PR from #1682.
In order to have separate padding pixels on each side of the map bounds. Right now its set up as `[n,e,s,w]` but I'm not sure this makes the most sense. Should we be consistent with the `offset` option and take 2 `Points` (one for each corner of the bounds)?

~~Testing the debug page is showing me I'm slightly off on the adjustments right now so I have to figure out what I did wrong in adjusting the camera function.~~ flipped x, y 😭 

cc @lucaswoj 
closes #1339 

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -304,22 +304,32 @@ class Camera extends Evented {
      *     {@link Map#easeTo}. If `false`, the map transitions using {@link Map#flyTo}. See
      *     {@link Map#flyTo} for information about the options specific to that animated transition.
      * @param {Function} [options.easing] An easing function for the animated transition.
-     * @param {number} [options.padding=0] The amount of padding, in pixels, to allow around the specified bounds.
+     * @param {number|Array} options.padding how much padding there is around the given bounds on each side in pixels. Accepts a number for all padding edges or an array [n, e, s, w]
      * @param {PointLike} [options.offset=[0, 0]] The center of the given bounds relative to the map's center, measured in pixels.
      * @param {number} [options.maxZoom] The maximum zoom level to allow when the map view transitions to the specified bounds.
      * @param {Object} [eventData] Data to propagate to any event listeners.
      * @fires movestart
      * @fires moveend
      * @returns {Map} `this`
+	 * @example
+     * var bbox = [[-79, 43], [-73, 45]];
+     * map.fitBounds(bbox, {
+     *   padding: [10, 40, 10, 40]
+     * });
      * @see [Fit a map to a bounding box](https://www.mapbox.com/mapbox-gl-js/example/fitbounds/)
      */
     fitBounds(bounds, options, eventData) {
 
         options = util.extend({
-            padding: 0,
+            padding: [0, 0, 0, 0],
             offset: [0, 0],
             maxZoom: this.getMaxZoom()
         }, options);
+
+        if (typeof options.padding === 'number') {
+            var p = options.padding;
+            options.padding = [p, p, p, p];
+        }
 
         bounds = LngLatBounds.convert(bounds);
 
@@ -328,8 +338,8 @@ class Camera extends Evented {
             nw = tr.project(bounds.getNorthWest()),
             se = tr.project(bounds.getSouthEast()),
             size = se.sub(nw),
-            scaleX = (tr.width - options.padding * 2 - Math.abs(offset.x) * 2) / size.x,
-            scaleY = (tr.height - options.padding * 2 - Math.abs(offset.y) * 2) / size.y;
+            scaleX = (tr.width - (options.padding[1] + options.padding[3]) * 2 - Math.abs(offset.x) * 2) / size.x,
+            scaleY = (tr.height - (options.padding[0] + options.padding[2]) * 2 - Math.abs(offset.y) * 2) / size.y;
 
         options.center = tr.unproject(nw.add(se).div(2));
         options.zoom = Math.min(tr.scaleZoom(tr.scale * Math.min(scaleX, scaleY)), options.maxZoom);
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -311,7 +311,8 @@ class Camera extends Evented {
      * @memberof Map#
      * @param {LngLatBoundsLike} bounds Center these bounds in the viewport and use the highest
      *      zoom level up to and including `Map#getMaxZoom()` that fits them in the viewport.
-     * @param {AnimationOptions | CameraOptions | PaddingOptions} [options]
+     * @param {AnimationOptions | CameraOptions } [options]
+     * @param {number | PaddingOptions} [options.padding] The amount of padding in pixels to add to the given bounds.
      * @param {boolean} [options.linear=false] If `true`, the map transitions using
      *     {@link Map#easeTo}. If `false`, the map transitions using {@link Map#flyTo}. See
      *     {@link Map#flyTo} for information about the options specific to that animated transition.
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -34,6 +34,17 @@ const Evented = require('../util/evented');
  * @property {boolean} animate If `false`, no animation will occur.
  */
 
+/**
+ * Options for setting padding on a call to {@link Map#fitBounds}. All properties of this object must be
+ * non-negative integers.
+ *
+ * @typedef {Object} PaddingOptions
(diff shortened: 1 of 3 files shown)


mollymerp:
ready for your 👀 @lucaswoj

js/ui/camera.js
```diff
@@ -303,32 +303,58 @@ class Camera extends Evented {
      *     {@link Map#easeTo}. If `false`, the map transitions using {@link Map#flyTo}. See
      *     {@link Map#flyTo} for information about the options specific to that animated transition.
      * @param {Function} [options.easing] An easing function for the animated transition.
-     * @param {number} [options.padding=0] The amount of padding, in pixels, to allow around the specified bounds.
+     * @param {number|Object} [options.padding] Sets padding around the given bounds on each side in pixels. Accepts a number for all padding edges or an object with padding for each compass direction (north, south, east, and west)
```
lucaswoj: We can provide a more specific type definition by creating a `PaddingOptions` [`@typdef`](http://usejsdoc.org/tags-typedef.html) for `{north: Number; east: Number; south: Number; west: Number;}`.

lucaswoj: What are your thoughts about providing geographic padding options (`north`, `east`, ...) vs screen padding options (`top`, `left`, ...)?

mollymerp: ok! will investigate

mollymerp: I think screen padding options make more sense but would be more complicated to implement... 😬

js/ui/camera.js
```diff
@@ -303,32 +303,58 @@ class Camera extends Evented {
...
+     *   padding: {north: 10, east:25, south: 15, west: 5}
+     * });
      * @see [Fit a map to a bounding box](https://www.mapbox.com/mapbox-gl-js/example/fitbounds/)
      */
     fitBounds(bounds, options, eventData) {
 
         options = util.extend({
-            padding: 0,
+            padding: {
+                north: 0,
```
mohsen1: I think it makes more sense to name these `top`, `right`, `bottom` and `left`. It's more consistent with Leaflet and Google Maps APIs.

js/ui/camera.js
lucaswoj: I just did a little experimentation with this PR and discovered that `Map#fitBounds` always resets the bearing to 0 when it is called. 🤔 

We may be able to rename these properties, ship this PR as-is, and open a separate issue about preserving bearing through a `Map#fitBounds`.

js/ui/camera.js
jaapster: I agree. If fitBounds is called on a (potentially) rotated map n, s, e and w don't make sense as you would want the padding to be applied to the viewport, not the feature.

mollymerp:
I updated the approach to one @lucaswoj and I discussed a couple weeks ago where we separate the padding out into two parts: the part that does not alter the center of the map (i.e. equal padding amts on top/bottom and left/right) and the part that does (any left over padding from the first part) and the latter part gets added to the offset object which is handled in subsequent calls to `flyTo` or `easeTo`. 

So if you pass a padding object like `{padding: {top: 10, bottom: 20, right: 50, left: 75}}` this will get broken into `{lateralPadding: 50, verticalPadding: 10, offset: [25, -10]}`

This was a compromise to get accurate results without completely refactoring the camera 😅  -- the same results are achievable with the current API, but hopefully this option will be more clear to users going forward.

mollymerp:
@lucaswoj I also added a typedef for the `PaddingOptions`, but I kind of think this makes our docs less clear (i.e. users have to click through to find out how to add padding) and since padding is only used in this one function we don't end up saving a ton of space. What do you think?

screen shot 2017-01-31 at 7 09 31 pm

lucaswoj:
@mollymerp The typedef is the most semantically correct. I hope we'll have an oppurtunity to refactor when we hit #2801. I also hope we'll add better support for `{inline: declarations}` soon.

src/ui/camera.js
```diff
@@ -295,41 +306,90 @@ class Camera extends Evented {
...
+        if (!util.deepEqual(Object.keys(options.padding).sort((a, b)=> {
+            if (a < b) return -1;
+            if (a > b) return 1;
+            return 0;
+        }), ["bottom", "left", "right", "top"])) {
+            util.warnOnce("options.padding must be a positive number, or an Object with keys 'bottom', 'left', 'right', 'top'");
+            console.log(Object.keys(options.padding).sort((a, b)=> {
+                if (a < b) return -1;
+                if (a > b) return 1;
+                return 0;
```
lucaswoj: accidental `console.log`?

lucaswoj: As best I can tell, these lines are equivalent to `.sort((a, b)=> a - b)`

mollymerp:
`offset` is an `AnimationOptions` property (should really be a `Camera` option I think), while `padding` is only an option on `Map#fitBounds`. If we deprecate one of them it should be `padding` I think. 

Given that this is one of the most requested features, maybe we should merge and this in advance of a larger refactor of the Camera API + documentation?

src/ui/camera.js
mollymerp: 🙈

src/ui/camera.js
```diff
@@ -295,41 +306,90 @@ class Camera extends Evented {
 
     /**
      * Pans and zooms the map to contain its visible area within the specified geographical bounds.
+     * This function will also reset the map's bearing to 0 if bearing is nonzero.
      *
      * @memberof Map#
      * @param {LngLatBoundsLike} bounds Center these bounds in the viewport and use the highest
      *      zoom level up to and including `Map#getMaxZoom()` that fits them in the viewport.
-     * @param {Object} [options]
+     * @param {AnimationOptions | CameraOptions | PaddingOptions} [options]
```
anandthakker: The way I read this, it makes me think that `top`, `bottom`, etc. are properties that can go _directly_ on `options`.  Since these should actually be sub-properties of `options.padding`, should this be `@param {number|PaddingOptions} [options.padding=0] The amount of padding to allow around the given bounds.` (or something less stilted-sounding heh)

test/unit/ui/camera.test.js
```diff
@@ -1136,5 +1136,38 @@ test('camera', (t) => {
         t.end();
     });
 
+    t.test('#fitBounds', (t) => {
+        t.test('no padding passed', (t) => {
+            const camera = createCamera();
+            const bb = [[-133, 16], [-68, 50]];
+
+            camera.fitBounds(bb, {duration:0});
+            t.deepEqual(fixedLngLat(camera.getCenter()), { lng: -100.5, lat: 34.717077774 }, 'pans to coordinates based on fitBounds');
```
anandthakker: Small preemptive suggestion: should we round the latitude before comparing it?

mollymerp: `fixedLngLat` does round to a default precision of 9 decimal places – that is rather precise, though, so it might be a good idea to reduce precision a bit

anandthakker: Oh whoops 🙈 totally didn't see that!

src/ui/camera.js
mollymerp: Actually, I don't think they're equivalent... according to [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) the approach you suggest can be used for comparing numeric values, but not for strings.

lucaswoj:
Per chat with @mollymerp today, we're going to 🚢 this API and look to make simplifying + breaking changes in conjunction with #2801

Issue #3888, opened 2017-01-04
queryRenderedFeatures(<bbox>) does not work as expected


jczaplew:
**mapbox-gl-js version**: 0.29.0

### Steps to Trigger Behavior

 1. Get the bounds of the current map
 2. Use those coordinates to pass an array of `PointLike` features to `queryRenderedFeatures`
 3. Zero or incorrect results are given

### Expected Behavior
Per [the documentation](https://www.mapbox.com/mapbox-gl-js/api/#Map#queryRenderedFeatures), passing the bounds of the current map view to `queryRenderedFeatures` should produce the same result as passing no geometry.

For example:

````
var mapBounds = map.getBounds()
var features = map.queryRenderedFeatures([ [mapBounds._sw.lng, mapBounds._sw.lat], [mapBounds._ne.lng, mapBounds._ne.lat] ], {layers:['airport']});
````

should be the same as 

````
var features = map.queryRenderedFeatures({layers:['airport']});
````

Demo can be found here - http://jsbin.com/dajukatomu/edit?html,output

### Actual Behavior
Different or incorrect results are given.

**Comment by mollymerp** [comment]
Hi @jczaplew,
Thanks for submitting this issue. The `Map#queryRenderedFeatures` function accepts an array of [`PointLike`](https://www.mapbox.com/mapbox-gl-js/api/#PointLike)s describing the SW and NE _screen coordinates_ of the area you want to query, not an array of [`LngLatLike`](https://www.mapbox.com/mapbox-gl-js/api/#LngLatLike)s which is what you're passing in your example. I've adjusted your example and I believe now it works as expected: http://jsbin.com/sobabamotu/1/edit?html,output

```js
var features = map.queryRenderedFeatures([map.project(mapBounds.getSouthWest()), map.project(mapBounds.getNorthEast())], {layers:['airport']});
```
you could also use
```js
var mapCanvas = map.getCanvas();
var features = map.queryRenderedFeatures([[0,mapCanvas.scrollHeight],[mapCanvas.scrollWidth, 0]], {layers:['airport']});
```

jczaplew:
Hi @mollymerp - thanks for the clarification! I submitted [a pull request](https://github.com/mapbox/mapbox-gl-js/pull/3897) making this clarification in the documentation.

pl3xus:
Oh my.... could this please also find its way into the android javadoc? I was debugging for hours. Thank you.

ShafSpecs:
[...]

PR #3897, opened 2017-01-05
Clarify that Point and PointLike types are screen coordinates in pixels

jczaplew:
This pull request simply clarifies the data type and unit of the `Point` and `PointLike` types in the documentation. As is, it is unclear that `x` and `y` refer to _screen_ coordinates and not _geographic_ coordinates.

This is a follow up to issue #3888


jfirebaugh:
Thanks!

PR #3933, opened 2017-01-10
add Mapbox wordmark as LogoControl

### LINKED ISSUE #1825: Add Mapbox wordmark control
This watermark should be present on Mapbox maps like [Mapbox.js does](https://github.com/mapbox/mapbox.js/blob/mb-pages/src/mapbox_logo.js).

screen shot 2015-12-10 at 2 33 19 pm

For reference here's the logo

``` svg
<svg xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg" version="1.1" width="60" height="20" viewBox="0 0 60 20">
  <path d="M1.4 1C0.6 1.1 0 1.8 0 2.5l0 11.9C0 15.2 0.7 15.9 1.5 15.9l1.8 0c0.8 0 1.5-0.7 1.5-1.5l0-4.3 0.7 1.2c0.5 0.9 2 0.9 2.6 0l0.7-1.2 0 4.3c0 0.8 0.7 1.5 1.5 1.5l1.8 0c0.8 0 1.5-0.7 1.5-1.5l0-0.2c1 1.1 2.5 1.8 4.1 1.8l4.1 0 0 1.5c0 0.8 0.7 1.5 1.5 1.5l1.8 0c0.8 0 1.5-0.7 1.5-1.5l0-1.5c1.7 0 3.1-0.6 4.1-1.7l0 0.2c0 0.4 0.2 0.8 0.4 1.1 0.3 0.3 0.7 0.4 1.1 0.4l3.3 0c2 0 3.9-1.1 4.9-2.7 1.1 1.6 2.8 2.7 4.9 2.7 1 0 2-0.3 2.8-0.7 0.3 0.4 0.7 0.6 1.2 0.6l2.1 0c0.5 0 1-0.2 1.3-0.6l0.9-1.3 0.9 1.3c0.3 0.4 0.7 0.6 1.2 0.6l2.2 0c1.1 0 1.8-1.4 1.2-2.3l-2.6-3.6 2.4-3.4c0.6-0.9-0.1-2.4-1.2-2.4l-2.2 0c-0.5 0-0.9 0.2-1.2 0.6l-0.7 1-0.7-1c-0.3-0.4-0.7-0.6-1.2-0.6l-2.2 0c-0.5 0-1 0.3-1.2 0.7-0.9-0.5-1.8-0.8-2.9-0.8-2 0-3.8 1.1-4.9 2.7-1-1.6-2.9-2.7-4.9-2.7l0-1.5c0-0.4-0.2-0.8-0.4-1.1-0.3-0.3-0.7-0.4-1.1-0.4l-1.7 0c-0.7 0-1.5 0.8-1.5 1.5l0 3.2c-1.1-1.1-2.5-1.7-4.1-1.7l-3.3 0c-0.8 0-1.5 0.7-1.5 1.5l0 0.3c-1-1.1-2.5-1.8-4.1-1.8-1.6 0-3.1 0.7-4.1 1.8l0-3.3c0-0.8-0.7-1.5-1.5-1.5l-1.4 0C9.6 1 8.4 1.4 7.8 2.4l-1 1.7-1-1.7c-0.6-1.1-1.9-1.4-2.8-1.4l-1.4 0c0 0-0.1 0-0.1 0zM17.6 8.7c0.5 0 0.9 0.4 0.9 1.3l0 1.3-0.9 0c-0.5 0-0.9-0.4-0.9-1.3 0-0.9 0.4-1.3 0.9-1.3zm8.8 0c0.8 0 1.2 0.7 1.2 1.3 0 0.9-0.6 1.3-1.2 1.3zm8.7 0c0.7 0 1.4 0.6 1.4 1.3 0 0.8-0.5 1.3-1.4 1.3zm10 0c0.6 0 1.1 0.5 1.1 1.3 0 0.8-0.5 1.3-1.1 1.3-0.6 0-1.1-0.5-1.1-1.3 0-0.8 0.5-1.3 1.1-1.3z" fill="rgba(0,0,0,0.25)" />
  <path d="m1.5 2.5 0 12 1.8 0 0-9.9 3.5 5.9 3.5-5.9 0 9.9 1.8 0 0-12-1.4 0C9.9 2.5 9.3 2.7 9 3.2l-2.3 3.9-2.3-3.9c-0.3-0.5-0.9-0.7-1.6-0.7zm30.6 0 0 12 3.3 0c2.4 0 4.4-1.9 4.4-4.5 0-2.6-1.9-4.5-4.4-4.5-0.5 0-0.9 0-1.5 0l0-3zm-14.5 3c-2.4 0-4.2 1.9-4.2 4.5 0 2.6 1.8 4.5 4.2 4.5l4.2 0 0-4.5c0-2.6-1.7-4.5-4.2-4.5zm27.6 0c-2.4 0-4.4 2-4.4 4.5 0 2.5 2 4.5 4.4 4.5 2.4 0 4.3-2 4.3-4.5 0-2.5-1.9-4.5-4.3-4.5zm-22 0 0 12 1.8 0 0-3c0.6 0 1 0 1.6 0 2.4 0 4.3-1.9 4.3-4.5 0-2.6-1.9-4.5-4.3-4.5zm26.2 0 3 4.4-3.2 4.6 2.1 0 2.2-3.1 2.1 3.1 2.2 0-3.2-4.6 3-4.4-2.2 0-1.9 2.8-2-2.8zm-31.7 1.7c1.4 0 2.4 1.2 2.4 2.8l0 2.8-2.4 0c-1.4 0-2.4-1.2-2.4-2.8 0-1.6 1-2.8 2.4-2.8zm27.6 0c1.4 0 2.6 1.2 2.6 2.8 0 1.5-1.1 2.8-2.6 2.8-1.4 0-2.6-1.2-2.6-2.8 0-1.5 1.2-2.8 2.6-2.8zm-20.2 0 1.6 0c1.6 0 2.6 1.3 2.6 2.8 0 1.5-1.1 2.8-2.6 2.8l-1.6 0zm9 0 1.5 0c1.5 0 2.6 1.3 2.6 2.8 0 1.5-1 2.8-2.6 2.8l-1.5 0z" fill="#fff" />
</svg>
```

mollymerp:
## Launch Checklist

closes #1825 

 - [ ] post benchmark scores


js/ui/control/logo_control.js
```diff
@@ -0,0 +1,34 @@
...
+ * A `LogoControl` is a control that adds the Mapbox watermark
+ * to the map as required by the [terms of service](https://www.mapbox.com/tos/) for Mapbox
+ * vector tiles and core styles.
+ *
+ * @implements {IControl}
+**/
+
+class LogoControl {
+    constructor() {
+    }
```
mourner: you can omit empty constructors

js/ui/control/logo_control.js
```diff
@@ -0,0 +1,34 @@
...
+
+    onAdd(map) {
+        this._map = map;
```
mourner: Do we need to save the map reference if it's not used by the control?

js/ui/map.js
```diff
@@ -200,6 +203,7 @@ class Map extends Camera {
         if (options.style) this.setStyle(options.style);
 
         if (options.attributionControl) this.addControl(new AttributionControl());
+        if (options.logoControl) this.addControl(new LogoControl(), 'bottom-left');
```
mourner: I think it's better to code the default position in the control code by adding a `getDefaultPosition` method.

js/ui/control/logo_control.js
mollymerp: I guess we don't! I was following the `IControl` docs, but you're right it isn't necessary in this case.

mollymerp:
@jfirebaugh do you mean we should punt on this until we implement something at the TileJSON level? 

I'm looking in to what implementing a flag like that looks like, but some enterprise customers will still need to be able to remove the logo even if they're using our vector tiles depending on the terms of their Enterprise agreement.

mollymerp:
Per chat in slack and convos w @sean-heff @ErinQuinn I think we should add the logo as on by default, and actually not document how to remove it. This would be the same behavior that we have currently implemented in the iOS SDK (per @1ec5) and be the most effective way to prevent users from unwillingly violate our TOS. 

we have lines of communication open with all users that have approved white labeling (customer success) and so if they want info on how to remove the logo, we can provide it.

mourner:
@mollymerp it seems to be implemented at the TileJSON level (with `mapbox_logo` property) — you would need to listen to `map` `source.load` event and check whether `e.source.mapbox_logo` is true.

mollymerp:
@mourner I think the `mapbox_logo` flag is only enabled for mapbox.js tilesets because I don't see it coming through in the tileJSON in mapbox-gl-js. Even if it were, that flag automatically disables the logo for any enterprise accounts, which is not ideal because only a fraction of enterprise accounts actually have the rights to remove the logo from their maps. 

Both the Android and iOS SDKs enable the Mapbox logo automatically for all users and I think it makes most sense to follow this pattern.

andrewharvey:
@mollymerp Not required but might be a good idea to add `aria-label="Mapbox logo"` to that element for better accessibility support.

mollymerp:
@andrewharvey thank you so much for catching that! I'll have to get better about making a habit of including these.

andrewharvey:
While technically not a breaking change, it will act as a breaking change for those who've followed the current advice and manually added the logo (as they'll end up with two overlapping logos, potentially slightly offset from each other). So a notice about this in the CHANGELOG when released will be welcome.

lucaswoj:
We are going to collaborate with the teams that maintain Mapbox's APIs to implement a `mapbox_logo`-like property before moving forward with this PR.

mollymerp:
I think we should merge now. This would make Mapbox GL JS have parity with our mobile SDKs and prevent further accidental violations of our TOS. We can always go back later when the API response contains appropriate flags to indicate permissible white-labeling and configure the control that way. 

I opened a ticket requesting this feature in the meantime https://github.com/mapbox/api-core/issues/5594

mollymerp:
Ok I've updated this PR to rely on the `mapbox_logo` property returned in the TileJSON object. I'm also in the process of updating how that property is set by the APIs but this PR works independently of those changes. 

cc @lucaswoj @jfirebaugh

js/ui/map.js
```diff
@@ -100,7 +100,7 @@ const defaultOptions = {
  *   Keep in mind that these classes are used for controlling a style layer's paint properties, so are *not* reflected
  *   in an HTML element's `class` attribute. To learn more about Mapbox style classes, read about
  *   [Layers](https://www.mapbox.com/mapbox-gl-style-spec/#layers) in the style specification.
- * @param {boolean} [options.attributionControl=true] If `true`, an [AttributionControl](#AttributionControl) will be added to the map.
```
andrewharvey: Did you really mean to replace `attributionControl` with `logoPosition`, or did you mean to replace `logoControl` with `logoPosition`?

Given further down you've dropped out `options.logoControl` with `options.logoPosition` but retained `options.attributionControl`.

mollymerp: yipes good catch @andrewharvey -- this was not intentional!

js/ui/map.js
```diff
@@ -47,6 +48,7 @@ const defaultOptions = {
     hash: false,
 
     attributionControl: true,
+    logoControl: true,
```
andrewharvey: I think you missed updating this to `logoPosition: 'bottom-left'`.

js/ui/control/logo_control.js
```diff
@@ -0,0 +1,65 @@
...
+ * to the map as required by the [terms of service](https://www.mapbox.com/tos/) for Mapbox
+ * vector tiles and core styles.
+ *
+ * @implements {IControl}
+**/
+
+class LogoControl {
+
+    constructor() {
+        util.bindAll(['_logoRequired', '_updateLogo'], this);
```
jfirebaugh: Nit: `_logoRequired` doesn't need to be bound (it's not used as an event handler).

js/ui/control/logo_control.js
```diff
@@ -0,0 +1,65 @@
...
+        this._map = map;
+        this._container = DOM.create('div', 'mapboxgl-ctrl');
+
+        this._map.on('data', this._updateLogo);
+        this._updateLogo();
+        return this._container;
+    }
+
+    onRemove() {
+        this._container.parentNode.removeChild(this._container);
```
jfirebaugh: Also needs `this._map.off('data', this._updateLogo)`.

PR #3941, opened 2017-01-10
add "Map#isMoving" method #2792

stepankuzmin:
Add `Map#isMoving` method according to #2792

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -134,7 +134,6 @@ class Map extends Camera {
         const transform = new Transform(options.minZoom, options.maxZoom, options.renderWorldCopies);
         super(transform, options);
 
-        this._isMoving = false;
         this._interactive = options.interactive;
         this._failIfMajorPerformanceCaveat = options.failIfMajorPerformanceCaveat;
         this._preserveDrawingBuffer = options.preserveDrawingBuffer;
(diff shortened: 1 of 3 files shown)


js/ui/map.js
```diff
@@ -169,9 +170,13 @@ class Map extends Camera {
 
         this.on('move', this._update.bind(this, false));
         this.on('zoom', this._update.bind(this, true));
+        this.on('movestart', () => {
+            this._isMoving = true;
+        });
         this.on('moveend', () => {
             this.animationLoop.set(300); // text fading
             this._rerender();
+            this._isMoving = false;
```
lucaswoj: There are already a few places we store this state in `Camera`. I'd prefer to avoid adding a new one, especially on `Map`. We are trying to move towards making `Map` more simple: https://github.com/mapbox/mapbox-gl-js/issues/2741

 - `Camera#_finishFn` would be a drop-in replacement for `Map#_isMoving`
 - `Camera#panning`, `Camera#pitching`, and `Camera#zooming` are in the spirit of `Map#_isMoving`, should we add `Camera#panning` or `Camera#moving`?)

stepankuzmin: So we should move `Map#isMoving` to `Camera#isMoving`? And add `Camera#moving` similarly to `Camera#rotating`, `Camera#pitching`, and `Camera#zooming`?

Issue #1947, opened 2016-01-19
Request resources again when they expire per their HTTP headers


lucaswoj:
port https://github.com/mapbox/mapbox-gl-native/pull/2617 to JS

ref #1946

PR #3944, opened 2017-01-10
Re-request expired resources

lbud:
Closes https://github.com/mapbox/mapbox-gl-js/issues/1947.

This PR modifies `ajax#getArrayBuffer` to request `Cache-Control` and `Expires` headers. `SourceCache` maintains timers for tiles with set expiry dates for both active and cached tiles:
* when an active tile's timer expires, it re-requests the tile
* when a cached tile's timer expires, it is removed from the cache

## Launch Checklist

### DIFF
--- a/js/source/source_cache.js
+++ b/js/source/source_cache.js
@@ -487,19 +487,6 @@ class SourceCache extends Evented {
     }
 
     /**
-     * Remove tiles that h
-     * @private
-     */
-    clearExpiredTiles() {
-        for (const id in this._tiles)
-            // check if the header is before now
-            // remove it if it is
-            this.removeTile(id);
-        this._cache.reset();
-    }
-
-
-    /**
      * Search through our current tiles and attempt to find the tiles that
      * cover the given bounds.
      * @param {Array<Coordinate>} queryGeometry coordinates of the corners of bounding rectangle
--- a/js/source/source_cache.js
+++ b/js/source/source_cache.js
@@ -162,8 +162,7 @@ class SourceCache extends Evented {
 
         tile.sourceCache = this;
         tile.timeAdded = new Date().getTime();
-
-        this._setTimers(id, tile);
+        this._setTileReloadTimer(id, tile);
         this._source.fire('data', {tile: tile, coord: tile.coord, dataType: 'tile'});
 
         // HACK this is necessary to fix https://github.com/mapbox/mapbox-gl-js/issues/2986
@@ -414,7 +413,7 @@ class SourceCache extends Evented {
                 if (this._cacheTimers[wrapped.id]) {
                     clearTimeout(this._cacheTimers[wrapped.id]);
                     this._cacheTimers[wrapped.id] = undefined;
-                    this._setTimers(wrapped.id, tile);
+                    this._setTileReloadTimer(wrapped.id, tile);
                 }
             }
         }
@@ -433,7 +432,7 @@ class SourceCache extends Evented {
         return tile;
     }
 
-    _setTimers(id, tile) {
+    _setTileReloadTimer(id, tile) {
         const tileExpires = tile.getExpiry();
         if (tileExpires) {
             this._timers[id] = setTimeout(() => {
@@ -443,7 +442,7 @@ class SourceCache extends Evented {
         }
     }
 
-    _setCacheTimers(id, tile) {
+    _setCacheInvalidationTimer(id, tile) {
         const tileExpires = tile.getExpiry();
         if (tileExpires) {
             this._cacheTimers[id] = setTimeout(() => {
@@ -478,7 +477,7 @@ class SourceCache extends Evented {
         if (tile.hasData()) {
             const wrappedId = tile.coord.wrapped().id;
             this._cache.add(wrappedId, tile);
-            this._setCacheTimers(wrappedId, tile);
+            this._setCacheInvalidationTimer(wrappedId, tile);
         } else {
             tile.aborted = true;
             this.abortTile(tile);
--- a/js/source/source_cache.js
+++ b/js/source/source_cache.js
@@ -459,6 +459,19 @@ class SourceCache extends Evented {
     }
 
     /**
+     * Remove tiles that h
+     * @private
+     */
+    clearExpiredTiles() {
+        for (const id in this._tiles)
+            // check if the header is before now
+            // remove it if it is
+            this.removeTile(id);
+        this._cache.reset();
+    }
+
+
+    /**
      * Search through our current tiles and attempt to find the tiles that
      * cover the given bounds.
      * @param {Array<Coordinate>} queryGeometry coordinates of the corners of bounding rectangle
--- a/js/source/source_cache.js
+++ b/js/source/source_cache.js
@@ -263,15 +263,7 @@ class SourceCache extends Evented {
             }
             if (this._cache.has(coord.id)) {
                 retain[coord.id] = true;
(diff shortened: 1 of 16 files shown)


lbud:
ready for 👀 🚦

js/source/source_cache.js
```diff
@@ -413,6 +433,26 @@ class SourceCache extends Evented {
...
+            }, tileExpires - new Date().getTime());
+        }
+    }
+
+    _setCacheTimers(id, tile) {
+        const tileExpires = tile.getExpiry();
+        if (tileExpires) {
+            this._cacheTimers[id] = setTimeout(() => {
+                this._cache.remove(id);
+                delete this._timers[id];
```
anandthakker: should this be `this._cacheTimers[id]`?

js/util/lru_cache.js
```diff
@@ -111,6 +111,23 @@ class LRUCache<T> {
...
+     * Remove a key/value combination from the cache.
+     *
+     * @param {string} key the key for the pair to delete
+     * @returns {LRUCache} this cache
+     * @private
+     */
+    remove(key: string) {
+        if (!this.has(key)) { return this; }
+
+        this.onRemove(this.data[key]);
```
anandthakker: In `add()`, it looks like `this.onRemove()` gets called after the item is deleted from `this.data`, whereas here, it's getting called before; being inconsistent _could_ cause problem at some point.

Also, a question: am I missing something, or does `cache.get()` actually also remove the item?  If that's so, is `remove()` redundant?  (Although, admittedly, I normally would _not_ expect a method called `get()` to have a side effect...)

js/source/source_cache.js
```diff
@@ -255,7 +262,15 @@ class SourceCache extends Evented {
...
             if (this._cache.has(coord.id)) {
                 retain[coord.id] = true;
-                return this._cache.get(coord.id);
+                const cachedTile = this._cache.get(coord.id);
+
+                if (this._cacheTimers[coord.id]) {
+                    clearTimeout(this._cacheTimers[coord.id]);
+                    delete this._cacheTimers[coord.id];
+                    this._setTimers(coord.id, cachedTile);
+                }
```
anandthakker: It looks to me like we always do `addTile()` with nonempty return value from this method -- since that's the case, and since we have equivalent logic for removing cacheTimer / adding timer already in addTile, is it possible that we could avoid redundancy by only doing it in addTile?

js/source/source_cache.js
```diff
@@ -255,7 +262,15 @@ class SourceCache extends Evented {
             }
             if (this._cache.has(coord.id)) {
                 retain[coord.id] = true;
-                return this._cache.get(coord.id);
+                const cachedTile = this._cache.get(coord.id);
+
+                if (this._cacheTimers[coord.id]) {
+                    clearTimeout(this._cacheTimers[coord.id]);
+                    delete this._cacheTimers[coord.id];
```
lbud: Oo @anandthakker thank you — did not know this, googled a little hole for myself to understand further: https://github.com/google/google-api-nodejs-client/issues/375

js/source/source_cache.js
lbud: it should, yes :bowing_woman:

js/util/lru_cache.js
lbud: `cache.get` removes an item from the cache and returns it, so there is some repeated code here, and we _could_ use `cache.get` to delete tiles from the cache, but then it's returning expiring tiles to nowhere and would be less intuitive in code (when cached tile's timer expires, retrieve it from the cache … vs when cached tile's timer expires, remove it from the cache). 😕

anandthakker: Yeah, fair enough.  I'd actually be okay with "returning expired tiles to nowhere" (i.e. the calling code need not actually store the return value), BUT because it's called `get` I agree it would look pretty weird:

```js
// oh, we need to remove the cache entry now, so
cache.get(id)
// cool, moving on
```

I sort of wonder if we should just _rename_ `LRUCache#get` to `LRUCache#remove` -- that way, whether the calling code actually uses the return value or not, it is very clear that you're removing the entry from the cache.

anandthakker:
@lbud changes LGTM!

All I have left are a couple nits on naming/clarity that I should have thought of earlier:
 - LRUCache get vs remove see comment [here](https://github.com/mapbox/mapbox-gl-js/pull/3944#discussion_r97642010)
 - In SourceCache, what do you think about renaming `_setTimers` to something like `_setTileReloadTimer` and `_setCacheTimers` to `_setCacheRemovalTimer`?
 - I'm less sure of this than the above, but: would it make sense to move the `this._cacheTimers[id] = undefined` into `_setTileReloadTimer()` (and similar for `this._timers = undefined`), so that the calling code could be cleaner (and so that each tile having <= 1 timer going is enforced by those functions)?

^ this is all motivated by the notion that caching logic in general and SourceCache in particular tend to be hard to read/reason about and easy to mess up later...

lbud:
> LRUCache get vs remove see comment here

I'd be _okay_ with renaming it, although I instinctively think it's fine to keep both: adding `remove` doesn't add much code complexity, and they exist to serve different purposes. I'm going to leave it for now and if we feel strongly we can revise in the future.

> I'm less sure of this than the above, but: would it make sense to move the this._cacheTimers[id] = undefined into _setTileReloadTimer() (and similar for this._timers = undefined), so that the calling code could be cleaner (and so that each tile having <= 1 timer going is enforced by those functions)?

I started doing this, but hesitated because of [this part](https://github.com/mapbox/mapbox-gl-js/blob/9657e59bccdfd991aefb513e6cb8d71acbdfb6c2/js/source/source_cache.js#L467-L468) of `removeTile` — to be honest I'm not quite sure in which cases a tile would land here and return early, but I'm not comfortable reordering things in this fn at the moment (which combining timer resetting with cacheTimer setting would necessitate). 🤔

anandthakker:
> I started doing this, but hesitated because of this part of removeTile — to be honest I'm not quite sure in which cases a tile would land here and return early, but I'm not comfortable reordering things in this fn at the moment (which combining timer resetting with cacheTimer setting would necessitate)

Ah, yeah I see what you mean.  👍

cammanderson:
Very cool stuff - excited about this. We have a use case for this right now, great to see it is in action!

karlguillotte:
Since I upgraded to the version that include that feature, I started getting bunch of errors _Unable to get property 'state' of undefined or null reference_. 

It is coming from https://github.com/mapbox/mapbox-gl-js/blob/b9e10b939c6a3fe5d7ecac209f751b4871970ede/src/source/source_cache.js#L149. 

I am sorry, I can not point you out to any gist that shows the errors. I am still investigating. Wondering if someone has that same issue?

Thanks for MapBox GL.

anandthakker:
@karlguillotte thank you for reporting this.  Could you please open a new issue describing this bug?

I know you said that you're still investigating -- if you're able to provide a minimal example demonstrating the problem, that would greatly improve our chances of to diagnosing and solving it.

AndyMoreland:
I have this bug also. Working on tracking down exactttllyy what's going on.

AndyMoreland:
The root cause appears to be this line: https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/source_cache.js#L438

The issue occurs when `tile.getExpiry() - (new Date()).getTime()` is larger than a 32-bit signed int. In my case the delta is `2592000000` and the timeout in executes immediately which results in an infinite loop of reloading tiles and the errors that @karlguillotte reported if I pan.

PR #3962, opened 2017-01-12
Fix garbled text on zoom

### LINKED ISSUE #3669: Labels become momentarily garbled while zooming
7325884c-c5a4-11e5-83e0-d4323c606577

Regression of #2012 and #3086

We should use the test-suite operations API to create a regression test.

mourner:
Fixes #3669. Sorry for regressing this. The bug happened when different tiles of the same symbol layer had different fontstacks, which can sometimes happen during zoom when only a part of tiles is parsed before rendering.

I have no idea how to add a regression test for this though because it only happens intermittently and depends on a lot of factors like network speed, etc. So I suggest merging this without a test for now.

`frame-duration` benchmark shows no difference.

### DIFF
--- a/js/render/draw_symbol.js
+++ b/js/render/draw_symbol.js
@@ -88,7 +88,7 @@ function drawLayerSymbols(painter, sourceCache, layer, coords, isText, translate
         gl.disable(gl.DEPTH_TEST);
     }
 
-    let program;
+    let program, prevFontstack;
 
     for (const coord of coords) {
         const tile = sourceCache.getTile(coord);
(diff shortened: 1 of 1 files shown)


lucaswoj:
> I have no idea how to add a regression test for this though because it only happens intermittently and depends on a lot of factors like network speed, etc. So I suggest merging this without a test for now.

There have been so many untestable regressions lately. 😞 It might be time to invest in some test suite infrastructure for controlling network requests & the clock.

mourner:
@lucaswoj alternatively, we could add a semi-private API for explicitly loading a set of specified tiles and bypassing the source-cache pyramid logic; this way we would also be able to more easily test overscaled tiles, etc.

PR #3973, opened 2017-01-13
Added vector tile data-join example

ryanbaumann:
## Launch Checklist

- [ ] Final code review

## Vector tile data join

Use categorical data-driven styles to join client data to vector tile geometry.

cc @kronick


docs/_posts/examples/3400-01-31-data-join.html
```diff
@@ -0,0 +1,73 @@
...
+    map.addSource('states', {
+        type: 'vector',
+        url: 'mapbox://' + mapid
+    });
+
+    // First value is the default, used where the is no data
+    var stops = [["0", "rgba(0,0,0,0)"]];
+
+    // Calculate color based on the data value
+    Object.keys(data).forEach(function(key) {
```
mourner: It would be simpler to do `for (let key in data)`

docs/_posts/examples/3400-01-31-data-join.html
```diff
@@ -0,0 +1,73 @@
...
+        let val = parseInt(data[key])
```
mourner: Values are already numbers so no need for `parseInt`.

docs/_posts/examples/3400-01-31-data-join.html
```diff
@@ -0,0 +1,73 @@
...
+        let green = Math.round(val / 20 * 100);
```
mourner: Should be capped by 255 to make sure it'll be a valid color.

docs/_posts/examples/3400-01-31-data-join.html
```diff
@@ -0,0 +1,73 @@
...
+        let color = "rgba(" + 0 + ", " + green + ", " + 0 + ", 1)";
+        stops.push([String(key), color]);
```
mourner: Object keys are always strings.

docs/_posts/examples/3400-01-31-data-join.html
```diff
@@ -0,0 +1,73 @@
...
+    "35": 4,
+    "40": 18,
+    "45": 11,
+    "50": 3
+}
+
+// Get the vector data to join
+var mapid = 'rsbaumann.8h761oct' 
+var layername = 'states'       
+var vt_match_prop = 'STATE_ID'
```
mourner: All three should be camelCase.

mourner: Also lacks semicolons.

docs/_posts/examples/3400-01-31-data-join.html
mourner: We should either use `let`/`const` everywhere, or `var` everywhere — let's not mix the styles. I'd suggest `var` to keep things in line with other examples.

ryanbaumann:
Updated based on feedback above.  cc/ @kronick @mourner @lucaswoj 

- [x] style errors
- [x] use mapbox public vector tileset
- [x] improved data generation code

join client data to vector tiles mapbox

docs/_posts/examples/3400-01-31-data-join.html
ryanbaumann: fixed

docs/_posts/examples/3400-01-31-data-join.html
```diff
@@ -0,0 +1,70 @@
...
+var map = new mapboxgl.Map({
+    container: 'map',
+    style: 'mapbox://styles/mapbox/light-v9',
+    center: [-99.9, 41.5],
+    zoom: 1
+});
+
+// Generate client data to join to vector tiles
+// Key is the property to join, value is the measure to style on
+// EX: {"01" : 27, "02" : 89}
```
lucaswoj: Do you have any real data sets we could use with this example? 

It may improve the readability of the example to have more concrete terminology (ie. "Calculate color based on the GDP" vs "Calculate color based on the data value") and remove the data generation step.

test/integration/render-tests/regressions/mapbox-gl-js#3910/style.json
```diff
@@ -0,0 +1,54 @@
...
+      "paint": {
+        "circle-color": {
+          "type": "identity",
+          "property": "blue"
+        },
+        "circle-translate": [5, 0]
+      }
+    }
+  ]
+}
```
lucaswoj: Was the inclusion of this regression test intentional?

docs/_posts/examples/3400-01-31-data-join.html
```diff
@@ -0,0 +1,70 @@
+--- 
+layout: example 
+category: example 
+title: Join client data to vector tiles 
+description: Use data-driven styles to join client data to vector tiles
```
lucaswoj: Can you think of a way to frame this title & description that captures the problem being solved as well as the proposed solution?

docs/_posts/examples/3400-01-31-data-join.html
kronick: @ryanbaumann I agree with @lucaswoj here that a real dataset would help add context to this example. Population for each state might be a good dataset: https://en.wikipedia.org/wiki/2010_United_States_Census#State_rankings . Can you think of a way to drop those numbers in without making the example look too cluttered?

ryanbaumann: Yes, using 2009 USA unemployment.

docs/_posts/examples/3400-01-31-data-join.html
ryanbaumann: Great question, trying this in the latest commit:

```
title: Join data to vector tiles
description: Style vector tiles based on client data properties using data-driven styles
```

test/integration/render-tests/regressions/mapbox-gl-js#3910/style.json
ryanbaumann: Not intentional, good catch.  Reverted that commit.

ryanbaumann:
Updated title, used real data properties from US Census Unemployment, improved comment and code clarify, removed extraneous commit.  Ready for 👀 .

 cc @kronick @lucaswoj 

join client data to vector tiles mapbox

lucaswoj:
Just pushed up some formatting & copy tweaks. This is good to go when tests pass 😄

Issue #3951, opened 2017-01-11
Tile textures are retained indefinitely when raster sources are removed


Scarysize:
**mapbox-gl-js version**: v0.31.0 (also found the issue with 0.26.0)

### Steps to Trigger Behavior
jsbin: https://jsbin.com/kojobocuka/edit?js,output

 1. Open jsbin. Take a heap snapshot with DevTools. Check allocated `WebGLTexture` object count (should be around 8, if the map wasn't moved/zoomed)
 2. Hit the "GO!" button, this will add and remove raster sources & layers with random ids and varying tile urls every 500ms a 180 times.
 3. Take another heap snapshopt. Check the texture object count again, it's now at about 3000.
 4. (The GO! button hit again to re-run the test and further increase the object count)

### Expected Behavior

Textures aren't retained indefinitely for removed sources.

### Actual Behavior

Textures are retained for removed sources; as far as I can see they won't be deleted. For long running applications with multiple raster sources & layers being added/removed over time this will lead to increasing memory consumption.

mourner:
Thanks for the report! This issue actually looks like a good task for a contribution. Want to work on a PR fixing this? I'll be happy to review and help you land this.

Scarysize:
Yes, I could work on this.

What would be the behaviour? Delete all Textures as soon as a source is removed?

What is the intention of keeping those textures around anyway? Is it to re-use them, when tiles are re-entering the viewport?

mourner:
See the original issue: https://github.com/mapbox/mapbox-gl-js/issues/591. The reason is to avoid expensive gl.createTexture/deleteTexture calls. Let's find out why the pool grows out of control and fix it so that it's stable.

Scarysize:
Okay, I think I'm understanding it now. `WebGLTexture` instances with certain sizes are cached, we are not interested in the textures contents.

So textures are saved in an array indexed by their `size`:

```js
//...
this.reusableTextures[texture.size] = [texture];
```

If a texture of this size is already available, the texture is just pushed into that array:

```js
textures.push(texture);
```

The retained textures are accessed again through `painter.getTileTexture(width, height)`; this is called e.g. by a `RasterTileSource` before creating a new texture from a loaded tile image:

```js
tile.texture = this.map.painter.getTileTexture(img.width);
```

Now `getTileTexture` tries to look for a reusable texture, but does the indexing wrong (?):

```js
const widthTextures = this.reusableTextures[width];
if (widthTextures) {
[...]

PR #3974, opened 2017-01-13
fix(painter): correctly access retained tile textures

Scarysize:
## Launch Checklist

As explained here https://github.com/mapbox/mapbox-gl-js/issues/3951#issuecomment-272171042 the tile textures retained by the `painter` are indexed by their size only (1 dimension), but are accessed using `width` AND `height` (2 dimension) as indices. A lot of textures will be retained, but never accessed again, leading to a build-up of `WebGLTexture` objects in memory.
This also results in `RasterTileSources` doing a lot of expensive `gl.createTexture` calls, because they can't find reusable ones.

I though about a doing an extra nesting level in `this.reusableTextures`. So it will look like this:

```js
{
  tiles: {},
  viewport: {}
}
```

This would make it easier to differentiate the types of textures being saved. Any thoughts?

### DIFF
--- a/js/render/painter.js
+++ b/js/render/painter.js
@@ -319,12 +319,9 @@ class Painter {
         this.reusableTextures.viewport.texture = texture;
     }
 
-    getTileTexture(width, height) {
-        const widthTextures = this.reusableTextures[width];
-        if (widthTextures) {
-            const textures = widthTextures[height || width];
-            return textures && textures.length > 0 ? textures.pop() : null;
-        }
+    getTileTexture(size) {
+        const textures = this.reusableTextures[size];
+        return textures && textures.length > 0 ? textures.pop() : null;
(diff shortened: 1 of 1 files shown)


Scarysize:
@mourner Should I implement the nested property in this PR or submit a separate one?

mourner:
@Scarysize it's a small change, you can just append this PR.

mourner:
One last thing, why is it `reusableTextures.viewport.texture` instead of just `reusableTextures.viewport`? I think the additional level of nesting here is redundant, so we could remove it in this PR too. cc @lbud

Scarysize:
My thoughts exactly. If @lbud has no particular reason for the nesting, I will remove it.

lbud:
🚢 👍

Issue #1824, opened 2015-12-10
Create an example for fullscreen mode


lucaswoj:
mapbox.js has [such an example](https://www.mapbox.com/mapbox.js/example/v1.0.0/leaflet-fullscreen)

ref https://github.com/mapbox/mapbox-gl-js/issues/1821#issuecomment-163470008

cc @jingsam

josiekre:
+1

dshakya:
Is the team working on implementing Fullscreen plugin for MapboxGL? Missing this feature big time.

lucaswoj:
@dshakya While there is no pre-built plugin, you should be able to implement this feature downstream:

```js
map.getCanvas().webkitRequestFullScreen(); // there are a few other vendor prefixes to try
map.resize();
```

colleenmcginnis:
@lucaswoj I was thinking of working on this example. Do you think we should build a control or just show users how to implement what you suggested above?

dshakya:
Would be lovely if you could build a control like the one for Leaflet

dshakya:
Hi @colleenmcginnis, would you know if anyone is working on this feature? Or any directions on how would I go about implementing it?

karlguillotte:
@lucaswoj I created that fullscreen control. Would you be interested in a PR?

andrewharvey:
@karlguillotte Not sure if you saw [this one by @colleenmcginnis](https://github.com/mapbox/mapbox-gl-js/commit/4c5aff4f3319c15886ed72d370176c634b822913)  to compare?

karlguillotte:
Thanks @andrewharvey ! I have not seen that one yet.

shotor:
I tried this, but my controls disappear in fullscreen mode.

Doesn't work on JsFiddle, so here's a gist: https://gist.github.com/shotor/316d2dc8b07fb69b6b73e4589b07b6ce

Has anyone been able to get the controls to display in full screen?

anx-ckreuzberger:
Any chance the fullscreen stuff from [this commit](https://github.com/mapbox/mapbox-gl-js/commit/4c5aff4f3319c15886ed72d370176c634b822913) can make it into the master? Is there a PR or something?
Pinging @colleenmcginnis on this too

mollymerp:
this has shipped in #3977 !! will be available in the next release!

PR #3977, opened 2017-01-14
[not ready] create fullscreen example and control

colleenmcginnis:
Creates a fullscreen control and an example using the fullscreen control to address #1824. 

Hoping @mollymerp and @lucaswoj can help with some things that still need to be resolved:  
- [x] Do the control and the example need to be in two different PRs (`master` and `mb-pages`)? 
- [x] The current control works in example, but not in the debug page. @mollymerp looked at this earlier - do you remember what the issue was? 
- [x] In addition to changing the icon when the button is clicked, it also needs to change when the ESC key is pressed. Not sure what the best way is to do that as a part of the control. 
- [x] There are several errors in the `js/ui/control/fullscreen.js` file that I'm not sure how to interpret or resolve.


mollymerp:
awesome work @colleenmcginnis !! 
- no need to separate the `master` and `mb-pages` PRs bc we wouldn't want the example to show up before the control is released. `master` is merged into `mb-pages` on every release
- I do remember the thing with the debug page. I'll take a look into it on Monday
- I think you'll need to add a `case` in https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/handler/keyboard.js to handle the `esc` key functionality 
- I'll go through and see if I can help w the errors you're seeing.

js/ui/control/fullscreen.js
```diff
@@ -0,0 +1,59 @@
...
+    options: {
+        position: 'top-left'
+    },
+
+    onAdd: function(map) {
+        var className = 'mapboxgl-ctrl';
+        var container = this._container = DOM.create('div', className + '-group', map.getContainer());
+        var button = this._fullscreenButton = DOM.create('button', (className + '-icon ' + className + '-fullscreen'), this._container);
+        button.id = 'fullscreen-button';
+        this._fullscreenButton.type = 'button';
```
andrewharvey: @colleenmcginnis Could you include an aria-label attribute on the button with value of say "Fullscreen" for accessibility?

Similar to how it's done for the navigation buttons https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/control/navigation_control.js#L95

js/ui/control/fullscreen.js
```diff
@@ -0,0 +1,59 @@
...
+    },
+
+    _onClickFullscreen: function() {
+        var mapContainer = map.getContainer();
+        var button = document.getElementById('fullscreen-button');
+        var className = 'mapboxgl-ctrl';
+        if (window.innerHeight === screen.height) {
+            button.classList.remove(className + '-shrink');
+            button.classList.add(className + '-fullscreen');
+            if (document.exitFullscreen) {
```
mollymerp: I think you may need  to `require` the `window` util and use `window.document.exitFullscreen` etc to get past the linting errors.

dist/mapbox-gl.css
```diff
@@ -267,3 +274,11 @@
         display:none;
     }
 }
+.pseudo-fullscreen {
```
jfirebaugh: This doesn't seem to be used (and we shouldn't need to support "pseudo-fullscreen" mode, as all browsers supported by gl-js should support the HTML5 fullscreen API).

js/ui/control/fullscreen.js
```diff
@@ -0,0 +1,59 @@
...
+        return container;
+    },
+
+    isFullscreen: function () {
+        return this._isFullscreen || false;
+    },
+
+    _onClickFullscreen: function() {
+        var mapContainer = map.getContainer();
+        var button = document.getElementById('fullscreen-button');
```
jfirebaugh: Use `this._fullscreenButton` rather than looking up the element by ID.

js/ui/control/fullscreen.js
```diff
@@ -0,0 +1,59 @@
...
+Fullscreen.prototype = util.inherit(Control, {
+    options: {
+        position: 'top-left'
+    },
+
+    onAdd: function(map) {
+        var className = 'mapboxgl-ctrl';
+        var container = this._container = DOM.create('div', className + '-group', map.getContainer());
+        var button = this._fullscreenButton = DOM.create('button', (className + '-icon ' + className + '-fullscreen'), this._container);
+        button.id = 'fullscreen-button';
```
jfirebaugh: Avoid assigning an ID -- IDs are required to be unique, and there may be more than one map with fullscreen control on a page.

debug/index.html
```diff
@@ -26,6 +32,96 @@
     hash: true
 });
 
+map.addControl(new mapboxgl.NavigationControl());
+map.addControl(new mapboxgl.GeolocateControl());
+map.addControl(new mapboxgl.ScaleControl());
+map.addControl(new mapboxgl.FullscreenControl());
```
mollymerp: can you revert the changes on this file so it matches the `debug/index.html` file on master? you can add a new debug page for controls instead if you want too.

dist/mapbox-gl.css
```diff
@@ -85,7 +85,14 @@
 .mapboxgl-ctrl-icon.mapboxgl-ctrl-geolocate.watching  {
     background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg%20viewBox%3D%270%200%2020%2020%27%20xmlns%3D%27http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%27%3E%0D%0A%20%20%3Cpath%20style%3D%27fill%3A%2300f%3B%27%20d%3D%27M10%204C9%204%209%205%209%205L9%205.1A5%205%200%200%200%205.1%209L5%209C5%209%204%209%204%2010%204%2011%205%2011%205%2011L5.1%2011A5%205%200%200%200%209%2014.9L9%2015C9%2015%209%2016%2010%2016%2011%2016%2011%2015%2011%2015L11%2014.9A5%205%200%200%200%2014.9%2011L15%2011C15%2011%2016%2011%2016%2010%2016%209%2015%209%2015%209L14.9%209A5%205%200%200%200%2011%205.1L11%205C11%205%2011%204%2010%204zM10%206.5A3.5%203.5%200%200%201%2013.5%2010%203.5%203.5%200%200%201%2010%2013.5%203.5%203.5%200%200%201%206.5%2010%203.5%203.5%200%200%201%2010%206.5zM10%208.3A1.8%201.8%200%200%200%208.3%2010%201.8%201.8%200%200%200%2010%2011.8%201.8%201.8%200%200%200%2011.8%2010%201.8%201.8%200%200%200%2010%208.3z%27%20%2F%3E%0D%0A%3C%2Fsvg%3E");
 }
-
+.mapboxgl-ctrl-icon.mapboxgl-ctrl-fullscreen  {
+    padding: 5px;
```
mollymerp: I don't think you need the padding here because its included [here](https://github.com/mapbox/mapbox-gl-js/blob/9d28a22178d2d8be37d971cc4ab81df1f0febbf0/dist/mapbox-gl.css#L73-L75)

dist/mapbox-gl.css
```diff
@@ -85,7 +85,14 @@
...
+}
+.mapboxgl-ctrl-icon.mapboxgl-ctrl-shrink  {
+    padding: 5px;
```
mollymerp: same as L89

js/ui/control/fullscreen_control.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+const DOM = require('../../util/dom');
+const util = require('../../util/util');
+const window = require('../../util/window');
+
+class FullscreenControl {
```
mollymerp: can you add documentation markup here? here is an example from the `NavigationControl`

```
/**
 * A `NavigationControl` control contains zoom buttons and a compass.
 *
 * @implements {IControl}
 * @example
 * var nav = new mapboxgl.NavigationControl();
 * map.addControl(nav, 'top-left');
 * @see [Display map navigation controls](https://www.mapbox.com/mapbox-gl-js/example/navigation/)
 * @see [Add a third party vector tile source](https://www.mapbox.com/mapbox-gl-js/example/third-party/)
 */
```

package.json
```diff
@@ -117,7 +117,7 @@
     "build-docs": "documentation build --github --format html --config documentation.yml --theme ./docs/_theme --output docs/api/",
     "build": "npm run build-docs # invoked by publisher when publishing docs on the mb-pages branch",
     "start-docs": "npm run build-min && npm run build-docs && jekyll serve --watch",
-    "lint": "eslint --ignore-path .gitignore js test bench docs/_posts/examples/*.html debug/*.html",
+    "lint": "eslint --fix --ignore-path .gitignore js test bench docs/_posts/examples/*.html debug/*.html",
```
mollymerp: please revert this as well 😉

lucaswoj: I recently learned that you can do this with npm:

```
npm run lint -- --fix
```

.. and it'll pass the `--fix` flag down to the `eslint` command 😄

js/ui/control/fullscreen_control.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+class FullscreenControl {
+
+    constructor() {
+        this._fullscreen = false;
+        util.bindAll([
+            '_isFullscreen',
+            '_onClickFullscreen',
+            '_onKeyDown',
```
mollymerp: You can get rid of `_onKeyDown` here -- I think you deleted that function from a previous iteration.

mollymerp: Also I don't think you need to bind `_isFullscreen` bc its not used as an event listener.

js/ui/control/fullscreen_control.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    onAdd(map) {
+        const className = 'mapboxgl-ctrl';
+        const container = this._container = DOM.create('div', `${className} mapboxgl-ctrl-group`);
+        const button = this._fullscreenButton = DOM.create('button', (`${className}-icon ${className}-fullscreen`), this._container);
+        button.setAttribute("aria-label", "Toggle fullscreen");
+        this._fullscreenButton.addEventListener('click', this._onClickFullscreen);
+        this._mapContainer = map.getContainer();
+        window.document.addEventListener('webkitfullscreenchange', this._changeIcon);
```
jfirebaugh: `webkitfullscreenchange` is specific to WebKit -- you'll need to do feature detection to determine the appropriate event name for other browsers. [Here's how the Leaflet equivalent does it.](https://github.com/Leaflet/Leaflet.fullscreen/blob/0c599cf4f369ee50658dc6a77b6082e239fd4f9c/dist/Leaflet.fullscreen.js#L125-L135)

js/ui/control/fullscreen_control.js
```diff
@@ -0,0 +1,72 @@
...
+        return container;
+    }
+
+    onRemove() {
+        this._container.parentNode.removeChild(this._container);
```
jfirebaugh: `onRemove` should unbind the fullscreenchange event handler, using `window.document.removeEventListener`.

js/ui/control/fullscreen_control.js
```diff
@@ -0,0 +1,72 @@
...
+        this._map = null;
+    }
+
+    _isFullscreen() {
+        return this._fullscreen;
+    }
+
+    _changeIcon() {
```
jfirebaugh: `_changeIcon`, as the event handler bound to fullscreenchange event, needs to check if the event corresponds to the appropriate page element. There might be more than one map on the page, or other elements that go fullscreen. [Here's the equivalent code in the Leaflet plugin.](https://github.com/Leaflet/Leaflet.fullscreen/blob/0c599cf4f369ee50658dc6a77b6082e239fd4f9c/dist/Leaflet.fullscreen.js#L99-L111)

js/ui/control/fullscreen_control.js
```diff
@@ -0,0 +1,72 @@
...
+        this._fullscreen = !this._fullscreen;
+        const className = 'mapboxgl-ctrl';
+        this._fullscreenButton.classList.toggle(`${className}-shrink`);
+        this._fullscreenButton.classList.toggle(`${className}-fullscreen`);
+    }
+
+    _onClickFullscreen() {
+        if (this._isFullscreen()) {
```
jfirebaugh: There doesn't seem to be a case in this function for IE11 (`msRequestFullscreen` / `msExitFullscreen`).

colleenmcginnis:
@jfirebaugh - @mollymerp and I worked on addressing your comments the other day. Can you take another look when you get the chance? Thanks!

src/ui/control/fullscreen_control.js
```diff
@@ -0,0 +1,88 @@
...
+        this._fullscreenButton.addEventListener('click', this._onClickFullscreen);
+        this._mapContainer = map.getContainer();
+        window.document.addEventListener(this._fullscreenchange, this._changeIcon);
+        return container;
+    }
+
+    onRemove() {
+        this._container.parentNode.removeChild(this._container);
+        this._map = null;
+        window.document.removeEventListener(this._fullscreenchange);
```
jfirebaugh: A second argument is needed here -- `removeEventListener(this._fullscreenchange, this._changeIcon)`.

PR #3985, opened 2017-01-16
Fix race condition in Tile redoPlacement

### LINKED ISSUE #3700: Uncaught TypeError: Cannot read property 'length' of null (CollisionTile in Tile.reloadSymbolData)
**mapbox-gl-js version**: 0.28
Not reproduced with 0.26 ; Maybe present in 0.27 too.

### Steps to Trigger The Error

 1. Setup a geojson layer with a lot of features: for example, every building polygons from OSM in a city like Paris, France.
 2. Zoom in and out randomly inside the city
 3. Sometimes the following error occurs (details and stack below): `Uncaught TypeError: Cannot read property 'length' of null`

### Error
```
Uncaught TypeError: Cannot read property 'length' of null
    at new CollisionTile
    at Tile.reloadSymbolData
    at Tile.done
    at Actor.receive
```

```
var CollisionTile = function CollisionTile(angle, pitch, collisionBoxArray) {
    if (typeof angle === 'object') {
        var serialized = angle;
        collisionBoxArray = pitch;
        angle = serialized.angle;
        pitch = serialized.pitch;
        this.grid = new Grid(serialized.grid);
        this.ignoredGrid = new Grid(serialized.ignoredGrid);
    } else {
        this.grid = new Grid(EXTENT, 12, 6);
        this.ignoredGrid = new Grid(EXTENT, 12, 0);
    }

    this.minScale = 0.5;
    this.maxScale = 2;

    this.angle = angle;
    this.pitch = pitch;

    var sin = Math.sin(angle),
        cos = Math.cos(angle);
    this.rotationMatrix = [cos, -sin, sin, cos];
    this.reverseRotationMatrix = [cos, sin, -sin, cos];

    // Stretch boxes in y direction to account for the map tilt.
    this.yStretch = 1 / Math.cos(pitch / 180 * Math.PI);

    // The amount the map is squished depends on the y position.
    // Sort of account for this by making all boxes a bit bigger.
    this.yStretch = Math.pow(this.yStretch, 1.3);

    this.collisionBoxArray = collisionBoxArray;
    if (collisionBoxArray.length === 0) { // <===== !!!! CRASH HERE !!!!
        // the first collisionBoxArray is passed to a CollisionTile
```

stack_0

stack_1

stack_2

mourner:
Closes #3700. @sguignot please confirm that this fixes the issue for you.

The race condition happened when `redoPlacement` was called for an [empty GeoJSON tile](https://github.com/mapbox/mapbox-gl-js/blob/ac70fa7aaa1fe4d8505b4ca783047abf36a4cebd/js/source/tile.js#L69-L70). In this case tile `state` is `loaded`, but all the data variables are `null`, causing `redoPlacement` to proceed sending the request to redo placement when it's not necessary, and breaking in the callback. Added a basic tests to make sure this doesn't regress.

## Launch Checklist

 - ~~document any changes to public APIs~~
 - ~~post benchmark scores~~

### DIFF
--- a/js/source/tile.js
+++ b/js/source/tile.js
@@ -129,7 +129,7 @@ class Tile {
     }
 
     redoPlacement(source) {
-        if ((source.type !== 'vector' && source.type !== 'geojson')) {
+        if (source.type !== 'vector' && source.type !== 'geojson') {
             return;
         }
         if (this.state !== 'loaded') {
--- a/js/source/tile.js
+++ b/js/source/tile.js
@@ -129,12 +129,14 @@ class Tile {
     }
 
     redoPlacement(source) {
-        if (source.type !== 'vector' && source.type !== 'geojson') {
+        if ((source.type !== 'vector' && source.type !== 'geojson')) {
             return;
         }
-
-        if (this.state !== 'loaded' || this.state === 'reloading') {
+        if (this.state !== 'loaded') {
             this.redoWhenDone = true;
+            return;
+        }
+        if (!this.collisionTile) { // empty tile
             return;
         }
 
@@ -147,9 +149,7 @@ class Tile {
             angle: source.map.transform.angle,
             pitch: source.map.transform.pitch,
             showCollisionBoxes: source.map.showCollisionBoxes
-        }, done.bind(this), this.workerID);
-
-        function done(_, data) {
+        }, (_, data) => {
             this.reloadSymbolData(data, source.map.style);
             source.fire('data', {tile: this, coord: this.coord, dataType: 'tile'});
 
@@ -157,11 +157,12 @@ class Tile {
             if (source.map) source.map.painter.tileExtentVAO.vao = null;
 
             this.state = 'loaded';
+
             if (this.redoWhenDone) {
-                this.redoPlacement(source);
                 this.redoWhenDone = false;
+                this.redoPlacement(source);
             }
-        }
+        }, this.workerID);
     }
 
     getBucket(layer) {
--- a/test/js/source/tile.test.js
+++ b/test/js/source/tile.test.js
@@ -116,6 +116,14 @@ test('querySourceFeatures', (t) => {
     t.end();
 });
 
+test('Tile#redoPlacement on an empty tile', (t) => {
+    const tile = new Tile(new TileCoord(1, 1, 1));
+    tile.loadVectorData(null, createPainter());
+
+    t.doesNotThrow(() => tile.redoPlacement({type: 'vector'}));
+    t.end();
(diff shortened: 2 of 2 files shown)


js/source/tile.js
```diff
@@ -129,11 +129,11 @@ class Tile {
     }
 
     redoPlacement(source) {
-        if (source.type !== 'vector' && source.type !== 'geojson') {
+        if ((source.type !== 'vector' && source.type !== 'geojson') ||
+            !this.hasData() || !this.collisionTile) {
```
jfirebaugh: Does this affect the behavior if `redoPlacement` is called before the tile is initially loaded? For example, if the map is rotated while tiles are loading, it may be necessary to redo placement immediately after the load completes.

mourner: @jfirebaugh I thought about that, but the problem is that this logic doesn't work currently anyway, because `redoWhenDone` is only processed in the `done` callback which can only happen after a successful `redoPlacement` request (after the tile got `loaded`). Maybe we should rewrite the code to take this case into account in a follow-up PR.

jfirebaugh: `redoWhenDone` is also used when the tile is loaded, [here](https://github.com/mapbox/mapbox-gl-js/blob/fb173257e734cffa04f6636bc03b3798bdcd8a4f/js/source/vector_tile_source.js#L91-L94).

mourner: Ah, I missed that somehow. Will fix, thanks for spotting this.

sguignot:
@mourner yes it fixes the issue #3700, thanks!

js/source/tile.js
mourner: Fixed now. Ready for review.

jfirebaugh: Can you explain how this avoids the issue? It looks like checking `!this.collisionTile` will still prevent reaching the `this.redoWhenDone = true;` case when `redoPlacement` is called while the tile is initially loading.

mourner: Looks like I'm having a bad day. Amended the commit, need to add a test for this.

mourner: @jfirebaugh added a few more basic tests for this.

js/source/tile.js
```diff
@@ -129,14 +129,16 @@ class Tile {
     }
 
     redoPlacement(source) {
-        if (source.type !== 'vector' && source.type !== 'geojson') {
+        if ((source.type !== 'vector' && source.type !== 'geojson')) {
```
jfirebaugh: Nit: unnecessary extra parentheses.

PR #3997, opened 2017-01-17
Use mapbox-namespaced unitbezier

tmcw:
The published namespaced version, 0.0.0, has no code changes,
so there should not be any functional ramifications of this switch.


mourner:
Can you also push in the repo? It currently has the old name in `package.json`.

PR #4005, opened 2017-01-18
Require namespaced @mapbox/unitbezier package

anandthakker:
Fresh `npm install` places unitbezier in `node_modules/@mapbox/unitbezier`, so I believe we need to require it w/ the namespace.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,6 +1,9 @@
 [ignore]
 
-.*/node_modules/.*
+.*/node_modules/jsonlint/.*
+.*/node_modules/jsonlint-lines-primitives/.*
+.*/node_modules/unflowify/.*
+.*/node_modules/mapbox-gl-style-spec/test/fixture/.*
 
 [include]
 
--- a/js/util/util.js
+++ b/js/util/util.js
@@ -1,7 +1,7 @@
 'use strict';
 // @flow
 
-const UnitBezier = require('unitbezier');
+const UnitBezier = require('@mapbox/unitbezier');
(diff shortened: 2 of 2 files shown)


jfirebaugh:
So did #3997 pass on CI because the non-namespaced module was in the cache?

anandthakker:
> because the non-namespaced module was in the cache?

@jfirebaugh yeah, that's what I figured must have happened.  We could clear the cache and re-run to confirm?

mourner:
Why does the build fail? cc @tmcw

mourner:
I have reverted the change in master. Let's figure out the failures before we merge again.

anandthakker:
Hm yeah, I'm really not sure why the build failed.  The logs for ci/test.sh say:

> @mapbox/unitbezier. Required module not found

But, looking above at the dependency step, it looks like `@mapbox/unitbezier` was installed.  😕

anandthakker:
The command that's failing is `npm run test-flow`.  An observation: the file that fails, `util/util.js` happens to be the only one where we have both (a) `// @flow` checking turned on and (b) a namespaced NPM module required.

cc @tmcw - does this ring any bells for you?

anandthakker:
@mourner passing now - 👀 ?

mourner:
@anandthakker since I reverted the `package.json` change in master earlire, it has to be added again now.

anandthakker:
Oh, shoot - thanks @mourner , I'll add it again

Issue #3809, opened 2016-12-15
Severe line-offset artifacts when line-round-limit does not equal 1.05


aaronlidman:
![offset-yum](https://cloud.githubusercontent.com/assets/58878/21209070/1bedc03c-c240-11e6-9bc2-6a2c0e7076e5.gif)

**mapbox-gl-js version 0.28.0**

### Steps to Trigger Behavior

 1. set line-offset to > 0
 2. set line-round-limit to 0

### Expected Behavior

Nicely rendered lines

### Actual Behavior

Very artifacty lines.

---

Combined with https://github.com/mapbox/mapbox-gl-native/issues/5258 means I can't render offset lines consistently in JS and native with the same style.

Similar to https://github.com/mapbox/mapbox-gl-js/issues/2316, but applies with any line-join.

cc @lucaswoj @jfirebaugh @pveugen

lucaswoj:
One potential fix: https://github.com/mapbox/mapbox-gl-style-spec/issues/286

1ec5:
We ran into a similar issue on iOS when we changed the default line-round-limit in gl-native to 1, so we reverted the default to 1.05: mapbox/mapbox-gl-native#5258.

lbud:
I dug into this a bit last week (also, this and https://github.com/mapbox/mapbox-gl-js/issues/2316 are more or less the same issue — same underlying problem, just present in different joins).

The crux of the issue is that miter joins are able to offset correctly because they contain only a single (well, two-way) join vertex, and its normal is the correct normal vector for that line vertex (that is, its normal is the sum of the normals of the lines that meet there). Our method of line offsetting is more or less just shifting it along its normal vertex. Here's a miter-joined line, offset, where you can see all line vertices are just shifted along their normals:

![image](https://cloud.githubusercontent.com/assets/3170312/22439349/1f637d88-e6e4-11e6-8469-0d74196b0f75.png)

Now, a bevel join: The visual lines we draw are created with GL vertices with position attributes — that's the position of the center of the line — and then (simplified) data about the direction of their normal vertex (i.e. which way is perpendicular) and how wide to draw the line (how far out along the normal vertex do we visually draw). In bevel joins, we still draw the inner vertex at the true join normal (sum of the normals of both lines that meet there) but we draw two outer vertices, and calculate different normals to create a bevel. The problem is then when we offset all vertices in the direction of their normal vector, the bevel vertices go flying off in the wrong direction:

[...]

Issue #2316, opened 2016-03-23
Sloppy line rendering at high line-offsets


peterqliu:
**mapbox-gl-js version**: v0.15.0

### Steps to Trigger Behavior

Check out https://jsfiddle.net/nx2nh1vk/. It is identical to the [geojson line example](https://www.mapbox.com/mapbox-gl-js/example/geojson-line/), save with a `line-offset:20` property added.

### Expected Behavior

I expect lines to render as cleanly as if there were no offset.

### Actual Behavior

Line pinches, tapers, intersects itself, breaks, and is otherwise distended.

mourner:
Looks like this only happens with round linejoins. If you remove this property, it looks good.

mourner:
And bevel linejoins too. Screenshots:

Round:
![image](https://cloud.githubusercontent.com/assets/25395/13984364/5f74eb28-f0fe-11e5-8f4c-6b66befaf180.png)

Bevel:
![image](https://cloud.githubusercontent.com/assets/25395/13984369/6d88e1ce-f0fe-11e5-8a7f-167d0212fe1e.png)

Miter (default):
![image](https://cloud.githubusercontent.com/assets/25395/13984375/792972f0-f0fe-11e5-87d9-fcec2b96e7b8.png)

lucaswoj:
This is a known (but not ticketed) issue that affects mapbox-gl-js and mapbox-gl-native.

> [This patch implements rudimentary support for offsetting lines.
>
> It doesn't properly offset line joins that are very sharp, and generally only works okay-ish for miter joins (bevel or round joins start to look weird at a certain distance).](https://github.com/mapbox/mapbox-gl-native/pull/3092#issuecomment-159637466) -- @kkaefer

PR #4008, opened 2017-01-18
Fix line join artifacts with 180º or very sharp angles

lbud:
This fixes two separate (but in appearance seemingly related) 🐛s with line joins:
* Miter line joins with 180º angles were evaluated to have `(NaN, NaN)` normal join vectors (e.g. a (1, 0) unit vector and a (-1, 0) unit vector add to (0, 0), whose unit vector is (NaN, NaN)). Because they weren't calculating extrusion vectors correctly these resulted in artifacts like
image
after:
image

* Miter and bevel line joins with _very sharp_ angles and miter joins are evaluated via the ["`flipbevel`"](https://github.com/mapbox/mapbox-gl-js/blob/line-artifacts/js/data/bucket/line_bucket.js#L235-L248) path: since the angle is so acute, the miter would be far too big, so we bevel it off instead. However, the `flipbevel` path had a normal vector flipped, so the join's vertices were added in the opposite order expected for indexing and they were creating artifacts such as
image
(Note: this explanation may not visually make sense, because with very sharp angles we add an extra vertex near the join (something about [dashed lines](); if you remove that vertex and ⏬ opacity you can see the incorrect triangle indexing:)
image
after:
image

This also includes some comment spellchecking 😬 

I've added two lines to all of the line-join tests in the render suite: one that turns 180º, and one that turns very acutely. These now produce the following results (previously these tests only had the left three lines):

| test | `master` | `line-artifacts` |
| ---- | ---- | ---- |
| bevel | bevel-master | bevel-branch | 
| bevel-transparent | bevel-tr-master | bevel-tr-branch | 
| default | default-master | default-branch | 
| miter | miter-master | miter-branch | 
| miter-transparent | miter-tr-master | miter-tr-branch | 
| round | round-master | round-branch | 
| round-transparent | round-tr-master | round-tr-branch | 

(Turns out this bug wasn't actually ticketed 😬)

I'd like to get into https://github.com/mapbox/mapbox-gl-js/issues/3809 / https://github.com/mapbox/mapbox-gl-js/issues/2316 next (these are more or less the same issue, just applied to different join types), after I wrap up some other outstanding PRs — I dug just enough to know that those are pretty unrelated to ☝️.

## Launch Checklist


lbud:
Currently checking to see whether these affect gl-native too, since I've modified render tests here.

lbud:
This does indeed need a gl-native parity PR: ticketed at https://github.com/mapbox/mapbox-gl-native/issues/7778 and added ignores to line-join render tests.

aaronlidman:
🙇

lbud:
@aaronlidman sadly though this isn't the same as the line-offset issue you were seeing ☹️ going to hit that after I clear a few outstanding gl-js PRs off my plate.

PR #4010, opened 2017-01-18
Relax clearing of tiles on layer removed/readded

### LINKED ISSUE #3895: Layers flash on map when one is removed and reapplied in 0.29.0
In the latest version of mapbox-gl-js, all layers that share a tile source will flash if any of them are removed, then reapplied with the same id.

This is problematic in my application, where there is a large vector tile source for census blocks with many layers consuming it simultaneously. We need to be able to add and remove the layers if the user changes what they want to see, without flashing all of the layers the wish to keep.

**mapbox-gl-js version**:
0.29.0

### Steps to Trigger Behavior

jsbin example: https://jsbin.com/muyayuzeva/edit?js,output

### Expected Behavior
The `maine_layer` style layer should be removed and reapplied under the hood, but cause no flash of layers.

Also, the `new_york_layer` style layer should be unaffected by the removing/adding of the `maine_layer` style layer.

See same jsbin example w/ mapbox-gl@0.28.0 for expected behavior - https://jsbin.com/fabotihala/edit?html,js,output

### Actual Behavior
Both the `new_york_layer` and `maine_layer` flash on across the map.

anandthakker:
Fixes #3895

Quoting from the issue discussion:

> Potential fix is to relax the conditions under which we clear the underlying source tiles (rather than just reloading them). Right now, we do this anytime a layer of a given name is removed and re-added. I think we can instead do it when a layer is removed and re-added with a different type

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/debug/3895.html
+++ b/debug/3895.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Mapbox GL JS debug page</title>
+    <meta charset='utf-8'>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+    <link rel='stylesheet' href='/dist/mapbox-gl.css' />
+    <style>
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+
+<body>
+<div id='map'></div>
+
+<script src='/dist/mapbox-gl-dev.js'></script>
+<script src='/debug/access_token_generated.js'></script>
+<script>
+
+var map = window.map = new mapboxgl.Map({
+    container: 'map',
+    center: [-68.13734351262877, 45.137451890638886],
+    zoom: 5,
+    style: 'mapbox://styles/mapbox/streets-v10',
+    hash: true
+});
+
+map.on('load', function () {
+    const layer = {
+        id: 'test',
+        source: 'composite',
+        'source-layer': 'water',
+        type: 'fill',
+        paint: {
+            'fill-color': '#00ff00'
+        }
+    };
+
+    map.addLayer(layer);
+
+    setInterval(function() {
+        const test = map.getLayer('test');
+        if (test) {
+            map.removeLayer('test');
+            map.addLayer(layer);
+        }
+    }, 2000);
+});
+</script>
+</body>
+</html>
--- a/js/style/style.js
+++ b/js/style/style.js
@@ -455,13 +455,15 @@ class Style extends Evented {
 
         if (this._removedLayers[id] && layer.source) {
             // If, in the current batch, we have already removed this layer
-            // and we are now re-adding it, then we need to clear (rather
-            // than just reload) the underyling source's tiles.
-            // Otherwise, tiles marked 'reloading' will have buffers that are
-            // set up for the _previous_ version of this layer, confusing
+            // and we are now re-adding it with a different `type`, then we
+            // need to clear (rather than just reload) the underyling source's
+            // tiles.  Otherwise, tiles marked 'reloading' will have buckets /
+            // buffers that are set up for the _previous_ version of this
+            // layer, causing, e.g.:
             // https://github.com/mapbox/mapbox-gl-js/issues/3633
+            const removed = this._removedLayers[id];
             delete this._removedLayers[id];
-            this._updatedSources[layer.source] = 'clear';
+            this._updatedSources[layer.source] = removed.type !== layer.type ? 'clear' : 'reload';
         }
         this._updateLayer(layer);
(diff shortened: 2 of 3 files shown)


anandthakker:
I think this should be ready for 👀

debug/3895.html
```diff
@@ -0,0 +1,52 @@
...
+        const test = map.getLayer('test');
+        if (test) {
+            map.removeLayer('test');
+            map.addLayer(layer);
+        }
+    }, 2000);
+});
+</script>
+</body>
+</html>
```
lucaswoj: Would it be possible to test for this regression using an integration test rather than this debug page? Chances are that this debug page will not be checked very often 😞

js/style/style.js
```diff
@@ -455,13 +455,15 @@ class Style extends Evented {
...
+            // and we are now re-adding it with a different `type`, then we
+            // need to clear (rather than just reload) the underyling source's
+            // tiles.  Otherwise, tiles marked 'reloading' will have buckets /
+            // buffers that are set up for the _previous_ version of this
+            // layer, causing, e.g.:
             // https://github.com/mapbox/mapbox-gl-js/issues/3633
+            const removed = this._removedLayers[id];
             delete this._removedLayers[id];
-            this._updatedSources[layer.source] = 'clear';
+            this._updatedSources[layer.source] = removed.type !== layer.type ? 'clear' : 'reload';
```
lucaswoj: How will this implementation behave in the following edge cases?

1. I remove a layer, add a new layer with the same type and different property functions.
2. I remove a layer, add a new layer with a different type, remove that layer, add a new layer with the original type.

anandthakker: > 1. I remove a layer, add a new layer with the same type and different property functions

My (probably incomplete) understanding of this is that, for a given style property:
 - If we're going from a constant value to DDS and we attempt to render a not-yet-reloaded tile, then, since the `Style` says DDS, the painter will look for a corresponding paint vertex buffer.  Not finding it, it'll fall back to the default color. (from conversation w/ @lucaswoj the other day.)
 - If we're going from DDS to a constant value and we attempt to render a not-yet-reloaded tile, then the painter will pull the value directly from the `Style`, so it doesn't matter that the tile isn't yet reloaded.
 - If we're just changing a DDS function, then when we attempt to render a not-yet-reloaded tile, we'll simply get the old DDS values from the previous dds buffer.

> 2. I remove a layer, add a new layer with a different type, remove that layer, add a new layer with the original type.

Ugh, good point -- although I think this actually points to a more general issue, which is that we do not handle adding and removing a layer within the same 'batch'

debug/3895.html
anandthakker: Yeah, I agree that manual tests are of marginal use (at best), but I don't see a clear & reliable way to test this in an integration test, because the symptom in question here is a (very) ephemeral 'flash'.  🤔

js/style/style.js
anandthakker: > although I think this actually points to a more general issue, which is that we do not handle adding and removing a layer within the same 'batch'

Ah, nevermind -- we handle this more general thing fine.

@lucaswoj the case you raise -- "remove a layer, add a new layer with a different type, remove that layer, add a new layer with the original type" -- would, indeed, cause a flash as in the original issue here.  However, I'd be inclined to punt on this, since it's a pretty extreme edge case that I do not expect would actually appear in the wild, and even when it does occur, the symptom is not particularly severe.

debug/3895.html
anandthakker: @lucaswoj realized that I _could_ add unit tests for the correct `clearTiles()` vs `reload()` behavior that was underlying the flashing issue that we're trying to correct here -- added in c39fe08

js/style/style.js
lucaswoj: I'm not suggesting that we address these cases now. I'm just trying to better understand the scope of this PR. 😄

lucaswoj:
benchmark | master b85e493 | fix-layer-flash c39fe08
--- | --- | ---
**map-load** | 117 ms  | 117 ms 
**style-load** | 79 ms  | 74 ms 
**buffer** | 1,062 ms  | 1,063 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 3.9 ms, 0% > 16ms  | 4 ms, 0% > 16ms 
**query-point** | 0.85 ms  | 0.91 ms 
**query-box** | 64.96 ms  | 66.52 ms 
**geojson-setdata-small** | 8 ms  | 5 ms 
**geojson-setdata-large** | 122 ms  | 165 ms

anandthakker:
@lucaswoj 👌 to merge?

PR #4020, opened 2017-01-19
Hack transition from undefined fill-outline-color

### LINKED ISSUE #3657: Fix bug wherein unsetting and resetting fill-outline-color causes TypeError to be thrown

Using `setPaintProperty(..., 'fill-outline-color', undefined)` ( to 'unset' this paint property), and then setting it again to a value causes an exception to be thrown.

Example:

```js
var style = {
  "version": 8,
  "name": "Mapbox Dark",
  "sources": {
    "composite": {
      "url": "mapbox://mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v7",
      "type": "vector"
    }
  },
  "sprite": "mapbox://sprites/mapbox/dark-v9",
  "glyphs": "mapbox://fonts/mapbox/{fontstack}/{range}.pbf",
  "layers": [
    {
      "id":"background",
      "type":"background",
      "layout":{},
      "paint":{"background-color":"hsl(55, 1%, 20%)"}
    },
    {
      "id": "barrier_line-land-polygon",
      "type": "fill",
      "source": "composite",
      "source-layer": "barrier_line",
      "filter": [
        "all",
        [
          "==",
          "$type",
          "Polygon"
        ],
        [
          "==",
          "class",
          "land"
        ]
      ],
      "layout": {},
      "paint": {
        "fill-color": "hsl(55, 1%, 20%)",
        "fill-outline-color": "hsl(55, 1%, 20%)"
      }
    }
  ],
  "created": 0,
  "modified": 0,
  "owner": "mapbox",
  "id": "dark-v9",
  "draft": false
}
    container: 'map',
    zoom: 2,
    center: [-77.01866, 38.888],
    style: style,
    hash: style
});

var set = false
setInterval(function () {
    var value = set ? 'hsl(55, 1%, 20%)' : undefined;
    console.log(set ? 'set' : 'unset')
    map.setPaintProperty('background', 'background-color', set ? 'white' : 'black')
    map.setPaintProperty('barrier_line-land-polygon', 'fill-outline-color', value)
    set = !set
}, 1000);
```

Error:

```
Uncaught TypeError: Cannot read property '0' of undefined
    at StyleTransition.interpolate.color [as interp] (http://localhost:9966/dist/mapbox-gl-dev.js:18233:25)
    at StyleTransition.calculate (http://localhost:9966/dist/mapbox-gl-dev.js:10245:17)
    at FillStyleLayer.getPaintValue (http://localhost:9966/dist/mapbox-gl-dev.js:9699:31)
    at FillStyleLayer.getPaintValue (http://localhost:9966/dist/mapbox-gl-dev.js:9978:55)
    at FillStyleLayer.recalculate (http://localhost:9966/dist/mapbox-gl-dev.js:9795:46)
    at Style._recalculate (http://localhost:9966/dist/mapbox-gl-dev.js:8921:19)
    at Map._render (http://localhost:9966/dist/mapbox-gl-dev.js:16353:24)
```

anandthakker:
Closes #3657

Using "option 1" described [here](https://github.com/mapbox/mapbox-gl-js/issues/3657#issuecomment-261636490)

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/js/style/style_layer/fill_style_layer.js
+++ b/js/style/style_layer/fill_style_layer.js
@@ -6,11 +6,23 @@ const FillBucket = require('../../data/bucket/fill_bucket');
 class FillStyleLayer extends StyleLayer {
 
     getPaintValue(name, globalProperties, featureProperties) {
-        if (name === 'fill-outline-color' && this.getPaintProperty('fill-outline-color') === undefined) {
-            return super.getPaintValue('fill-color', globalProperties, featureProperties);
-        } else {
-            return super.getPaintValue(name, globalProperties, featureProperties);
+        if (name === 'fill-outline-color') {
+            const target = this.getPaintProperty('fill-outline-color');
+            if (target === undefined) {
+                return super.getPaintValue('fill-color', globalProperties, featureProperties);
+            }
+
+            // https://github.com/mapbox/mapbox-gl-js/issues/3657
+            const transition = this._paintTransitions['fill-outline-color'];
+            if (!transition.instant()) {
+                const source = transition.oldTransition._calculateTargetValue(globalProperties, featureProperties);
+                if (!source) {
+                    return super.getPaintValue('fill-color', globalProperties, featureProperties);
+                }
+            }
         }
+
+        return super.getPaintValue(name, globalProperties, featureProperties);
     }
 
     getPaintValueStopZoomLevels(name) {
--- a/js/style/style_layer/fill_style_layer.js
+++ b/js/style/style_layer/fill_style_layer.js
@@ -8,8 +8,12 @@ class FillStyleLayer extends StyleLayer {
 
     getPaintValue(name, globalProperties, featureProperties) {
         if (name === 'fill-outline-color') {
-            // Special-case handling when transitioning either to or from an
-            // undefined fill-outline-color values
+            // Special-case handling of undefined fill-outline-color values
+            if (this.getPaintProperty('fill-outline-color') === undefined) {
+                return super.getPaintValue('fill-color', globalProperties, featureProperties);
+            }
+
+            // Handle transitions from fill-outline-color: undefined
             let transition = this._paintTransitions['fill-outline-color'];
             while (transition) {
                 const declaredValue = (
--- a/js/style/style_layer/fill_style_layer.js
+++ b/js/style/style_layer/fill_style_layer.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const assert = require('assert');
 const StyleLayer = require('../style_layer');
 const FillBucket = require('../../data/bucket/fill_bucket');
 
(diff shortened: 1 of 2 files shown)


js/style/style_layer/fill_style_layer.js
```diff
@@ -6,11 +6,23 @@ const FillBucket = require('../../data/bucket/fill_bucket');
...
+        if (name === 'fill-outline-color') {
+            const target = this.getPaintProperty('fill-outline-color');
+            if (target === undefined) {
+                return super.getPaintValue('fill-color', globalProperties, featureProperties);
+            }
+
+            // https://github.com/mapbox/mapbox-gl-js/issues/3657
+            const transition = this._paintTransitions['fill-outline-color'];
+            if (!transition.instant()) {
+                const source = transition.oldTransition._calculateTargetValue(globalProperties, featureProperties);
```
anandthakker: This is an unfortunate use of private method `_calculateTargetValue`, but my feeling is that it's (slightly) less bad than the only other non-invasive alternative that I could see, which would be to hack [interpolate.color](https://github.com/mapbox/mapbox-gl-js/blob/master/js/util/interpolate.js#L27-L32) so that it can handle undefined values for `from`.  I prefer the hack here because it's more localized, and will be easier to excise when we get rid of fill-outline-color.

anandthakker:
benchmark | master 2793b42 | 3657 193e124
--- | --- | ---
**map-load** | 101 ms  | 139 ms 
**style-load** | 91 ms  | 87 ms 
**buffer** | 943 ms  | 954 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 4.7 ms, 0% > 16ms  | 4.6 ms, 0% > 16ms 
**query-point** | 0.67 ms  | 0.79 ms 
**query-box** | 59.62 ms  | 62.67 ms 
**geojson-setdata-small** | 6 ms  | 5 ms 
**geojson-setdata-large** | 113 ms  | 172 ms

anandthakker:
@mourner @lucaswoj ready for 👀

anandthakker:
Transcribing from chat

> @anandthakker I agree with you that using `_calculateTargetValue` feels non-optimal
> Do you think it’d be possible to fix the bug by changing the construction of `fill-outline-color`‘s `StyleTransition` object?
> Another alternative:
> tidy up the `StyleTransition` API and get rid of the underscore

We agreed to go with the second option, but after looking into this (and also, alas, a detour 👇🐰🕳  investigating the first option), I'm now 👎 on removing the underscore.  The method we'd have to remove the underscore on, `_calculateTargetValue`, really is an internal implementation detail of StyleTransition, and I don't think there's any natural/legit part of it to extract into a public method.   Doing so feels contrived, which IMO is no better than actually using the underscored method in the first place.

All that said, in 20ea68f83153261d84b48d37ed27ea6d18805f4a i did revise to a slightly cleaner approach by relying on the public--though sorta deep--object hierarchy rather than the private method

```
StyleTransition --> StyleDeclaration --> value
  +- oldTransition: StyleTransition // forms a chain of ongoing transitions
  +- declaration: StyleDeclaration

StyleDeclaration
  +- value: ... // the declared value, which might be undefined
```

cc @lucaswoj @mourner

PR #4021, opened 2017-01-19
Note CSS in npm installation instructions

davidtheclark:
I saw that there was no mention of the required CSS file outside of the "Using Mapbox GL JS with a `<script>` tag" section. I thought it might be a good idea to include it in each installation section, for readers who are skipping to the section that concerns them.

(I deleted the Launch Checklist since this is just a README change.)

PR #4023, opened 2017-01-20
Note CSS in Browserify installation instructions

davidtheclark:
Follow-up to #4021, noting CSS in the website's API docs. We end up with this (a new sentence at the end):

screen shot 2017-01-19 at 5 08 52 pm

I also added a slash in the `docs/README.md` instruction, because without it the command landed me on a 404.

### DIFF
--- a/docs/_theme/index.hbs
+++ b/docs/_theme/index.hbs
@@ -172,7 +172,7 @@ var map = new mapboxgl.Map({
 {% endhighlight %}
 </div>
 
-<p class="space-top2">Also, add to your site the Mapbox GL JS CSS from <code>mapbox-gl/dist/mapbox-gl.css</code> or (via CDN) <code>{{site.tileApi}}/mapbox-gl-js/v{{site.data.package.version}}/mapbox-gl.css</code>.</p>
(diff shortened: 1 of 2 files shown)


davidtheclark:
Note that I also included a mention of the CDN CSS file here, but did not in #4021. That's because I wasn't sure if you wanted to add another couple of places to the README that need to change with each version — whereas here the version number is autopopulated.

docs/_theme/index.hbs
```diff
@@ -171,6 +171,9 @@ var map = new mapboxgl.Map({
 });
 {% endhighlight %}
 </div>
+
+<p class="space-top2">Also, add to your site the Mapbox GL JS CSS from <code>mapbox-gl/dist/mapbox-gl.css</code> or (via CDN) <code>{{site.tileApi}}/mapbox-gl-js/v{{site.data.package.version}}/mapbox-gl.css</code>.</p>
```
anandthakker: As in https://github.com/mapbox/mapbox-gl-js/pull/4021#discussion_r96993197, I think it couldn't hurt to include `node_modules/...` in the path.

davidtheclark:
@anandthakker Good idea. I added `node_modules` to the path.

lucaswoj:
Made a minor copy edit & updated the language in `README.md` per chat with @davidtheclark.

PR #4027, opened 2017-01-20
Followup to #4008 (180º + sharp line-join artifacts)

lbud:
Addresses comments from #4008 and https://github.com/mapbox/mapbox-gl-native/pull/7786/

### DIFF
--- a/js/data/bucket/line_bucket.js
+++ b/js/data/bucket/line_bucket.js
@@ -159,16 +159,18 @@ class LineBucket extends Bucket {
 
             // Determine the normal of the join extrusion. It is the angle bisector
             // of the segments between the previous line and the next line.
-            let joinNormal = prevNormal.add(nextNormal)._unit();
+            let joinNormal = prevNormal.add(nextNormal);
 
-            if (isNaN(joinNormal.x) && isNaN(joinNormal.y) && join === 'miter') {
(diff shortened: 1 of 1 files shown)


js/data/bucket/line_bucket.js
```diff
@@ -159,16 +159,18 @@ class LineBucket extends Bucket {
 
             // Determine the normal of the join extrusion. It is the angle bisector
             // of the segments between the previous line and the next line.
-            let joinNormal = prevNormal.add(nextNormal)._unit();
+            let joinNormal = prevNormal.add(nextNormal);
 
-            if (isNaN(joinNormal.x) && isNaN(joinNormal.y) && join === 'miter') {
+            if (joinNormal.x === 0 && joinNormal.y === 0 && join === 'miter') {
```
kkaefer: this shouldn't include `join === 'miter'`; the division by zero happens in all cases. For non-miter joins, the `else` branch is still executed, leading to NaN values.

kkaefer: Alternate fix in https://github.com/mapbox/mapbox-gl-js/pull/4028

lbud: Thanks @kkaefer — didn't give this the time I should have yesterday :bowing_woman:

Issue #3691, opened 2016-11-23
"Map#is*Loaded" methods


lucaswoj:
There is a need for a public API that determines whether or not a source is loaded in throttling calls to `GeoJSONSource#setData`. I often find myself using `map.style.sourceCache[sourceId].loaded()` for this purpose. It would be nice to provide a public API instead: `Map#isSourceLoaded(sourceId: string): boolean`

-----

 - add `Map#isSourceLoaded(sourceId)`
 - add `Map#isStyleLoaded()`
 - add `Map#isTilesLoaded()` / `Map#isTileLoaded(coord)` / ...
 - rename `Map#loaded` -> `Map#isLoaded` / `Map#isEverythingLoaded` / ...

sarahkleins:
Hey @lucaswoj! Going to start working on a few gl-js starter tasks next week. This one seems like a good place to start :)

lucaswoj:
So happy to hear you'll be working on this @sarahkleins. Looking forward to working with you to 🚢 this feature! Let me know if there's anything I can do to help.

lucaswoj:
Expanding the scope of this ticket to include

 - adding `Map#isSourceLoaded(sourceId)`
 - adding `Map#isStyleLoaded()`
 - adding `Map#isTilesLoaded()` / `Map#isTileLoaded(coord)` / ...
 - renaming `Map#loaded` -> `Map#isLoaded` / `Map#isEverythingLoaded` / ...

mollymerp:
This will tie in nicely w the refactoring of our events system #3970 😂

PR #4033, opened 2017-01-21
add Map#isSourceLoaded #3691

stepankuzmin:
Add `Map#isSourceLoaded` according to #3691

### DIFF
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -740,12 +740,9 @@ class Map extends Camera {
      * @returns {boolean} A Boolean indicating whether the source is loaded.
      */
     isSourceLoaded(id) {
-        const source = this.style.sourceCaches[id];
+        const source = this.style && this.style.sourceCaches[id];
         if (source === undefined) {
-            this.fire('error', {
-                error: new Error(`There is no source with ID '${id}'`)
-            });
-            return;
+            throw new Error(`There is no source with ID '${id}'`);
         }
         return source.loaded();
     }
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -742,7 +742,10 @@ class Map extends Camera {
     isSourceLoaded(id) {
         const source = this.style && this.style.sourceCaches[id];
         if (source === undefined) {
-            throw new Error(`There is no source with ID '${id}'`);
+            this.fire('error', {
+                error: new Error(`There is no source with ID '${id}'`)
+            });
+            return;
         }
         return source.loaded();
     }
--- a/js/ui/map.js
+++ b/js/ui/map.js
@@ -734,6 +734,23 @@ class Map extends Camera {
     }
 
     /**
+     * Returns a Boolean indicating whether the source is loaded.
+     *
+     * @param {string} id The ID of the source to be checked.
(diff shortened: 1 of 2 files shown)


js/ui/map.js
```diff
@@ -734,6 +734,23 @@ class Map extends Camera {
     }
 
     /**
+     * Returns a Boolean indicating whether the source is loaded.
+     *
+     * @param {string} id The ID of the source to be checked.
+     * @returns {boolean} A Boolean indicating whether the source is loaded.
+     */
+    isSourceLoaded(id) {
+        const source = this.style.sourceCaches[id];
```
anandthakker: Note that it's possible for `Map` to be constructed without a style, which would cause this line to blow up.  Something like `const source = this.style && this.style.sourceCaches[id];` would work.

js/ui/map.js
```diff
@@ -734,6 +734,23 @@ class Map extends Camera {
...
+        if (source === undefined) {
+            this.fire('error', {
+                error: new Error(`There is no source with ID '${id}'`)
+            });
+            return;
+        }
```
anandthakker: While it's true that the contributing guidelines [say to use error events](https://github.com/mapbox/mapbox-gl-js/blob/81d6715f8cf13c87ab930257a86440fbd0f3a91f/CONTRIBUTING.md#L98), I think `throw`ing an error might be more appropriate here, because we know immediately(synchronously) that there's an error. 

When, as in this case, synchronous error reporting is possible, I think it's preferable, as it's generally easier to diagnose/fix.  (E.g.: because of the `return;`, a user of this method might not even notice the error, since something like `if (map.isSourceLoaded(id)) { ... }` would still work without blowing up.)

js/ui/map.js
```diff
@@ -734,6 +734,20 @@ class Map extends Camera {
...
     /**
+     * Returns a Boolean indicating whether the source is loaded.
+     *
+     * @param {string} id The ID of the source to be checked.
+     * @returns {boolean} A Boolean indicating whether the source is loaded.
+     */
+    isSourceLoaded(id) {
+        const source = this.style && this.style.sourceCaches[id];
+        if (source === undefined) {
+            throw new Error(`There is no source with ID '${id}'`);
```
lucaswoj: Per https://github.com/mapbox/mapbox-gl-js/issues/3208, we have a convention that we emit an `error` event rather than `throw` an error.

anandthakker: @lucaswoj ah, shoot -- this was a mistake on my part in [this comment](https://github.com/mapbox/mapbox-gl-js/pull/4033#discussion_r97206120).  I'd seen the guideline in CONTRIBUTING but didn't take it as strict/universal.  My bad.

anandthakker:
@stepankuzmin Oof, my apologies re: https://github.com/mapbox/mapbox-gl-js/pull/4033#pullrequestreview-17987456 -- you had it right the first time 😞

stepankuzmin:
No problem, @anandthakker. I've fixed that :)

anandthakker:
Thanks @stepankuzmin !

PR #4038, opened 2017-01-23
Add deprecation notice for paint classes

anandthakker:


js/ui/map.js
```diff
@@ -256,6 +256,8 @@ class Map extends Camera {
      * in an HTML element's `class` attribute. To learn more about Mapbox style classes, read about
      * [Layers](https://www.mapbox.com/mapbox-gl-style-spec/#layers) in the style specification.
      *
+     * **Note:** Class-specific paint properties are deprecated and will be removed in an upcoming release of Mapbox GL JS.
+     *
```
lucaswoj: What do you think about adding a `util.warnOnce` to the body of these methods?

anandthakker:
> This phrasing makes it seem as though we are removing a subset of the "class" functionality. Would you be open to some slightly stronger wording like:

> Note: Style classes are deprecated and will be removed in an upcoming release of Mapbox GL JS.

Yeah, nice edit.  I stole the language from the style-spec docs, so we may want to change it there as well -- I could wait on that till after you're done merging it into this repo, though?

lucaswoj:
@anandthakker We can wait until the two repos merge or make the change upstream now. Whichever makes the most sense for your workflow.

anandthakker:
@lucaswoj changes made.  Let's do the style-spec edit in a different PR

Issue #4054, opened 2017-01-26
setRTLTextPlugin broken by change to ajax.GetArrayBuffer interface


ChrisLoer:
We need to update it to use the new interface, until then `setRTLTextPlugin` will fail.

lucaswoj:
Fixed!

lbud:
😞 rats, sorry about this @ChrisLoer!

PR #4055, opened 2017-01-26
Fix regression #4054: use new ajax.getArrayBuffer interface for rtl-text

ChrisLoer:
There's still no automated testing covering the asynchronous load (because we don't want automated tests dependent on any network interaction). Do we have mocks that would be useful here?

### DIFF
--- a/js/source/rtl_text_plugin.js
+++ b/js/source/rtl_text_plugin.js
@@ -28,7 +28,7 @@ module.exports.setRTLTextPlugin = function(pluginURL, callback) {
             callback(err);
         } else {
             pluginBlobURL =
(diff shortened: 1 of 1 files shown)


lucaswoj:
> Do we have mocks that would be useful here?

First thing that comes to mind: we could use Sinon's "fake" `XMLHttpRequest` mock as we do in some unit tests http://sinonjs.org/docs/#server

anandthakker:
> we could use Sinon's "fake" XMLHttpRequest mock

👍 to that -- we technically _could_ use `proxyquire` to mock the `js/util/ajax` module, but that doesn't guard against ajax's interface changing, as in this case.

PR #4066, opened 2017-01-30
Start architecture documentation

anandthakker:
Occurred to me as I was reading through / learning how the rendering subsystem works in more detail that capturing the notes might be a good way to start an architecture doc.  (This is something I've long wanted to add, as a way to make contributing to GL JS more accessible.)

### DIFF
--- a/src/data/bucket/circle_bucket.js
+++ b/src/data/bucket/circle_bucket.js
@@ -81,4 +81,6 @@ class CircleBucket extends Bucket {
     }
 }
 
+CircleBucket.programInterface = circleInterface;
+
 module.exports = CircleBucket;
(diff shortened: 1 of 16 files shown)


mollymerp:
yessss thank you so much for taking this on!!

andrewharvey:
Thank you, this would be a big help!

anandthakker:
I had a few questions while reading through `SymbolBucket`, and added them as TODO's here so as to remember to get & document answers to them before merging.  @lucaswoj @jfirebaugh @mourner could one of you take a 👀 ?

src/data/bucket/circle_bucket.js
```diff
@@ -81,4 +81,6 @@ class CircleBucket extends Bucket {
     }
 }
 
+CircleBucket.programInterface = circleInterface;
```
lucaswoj: I'm confused -- is this related to the architectural documentation or something else?

lucaswoj: Ah. This is for the `describe_shaders` utility 👍

Issue #4088, opened 2017-02-01
Remove "fill-outline-color"


lucaswoj:
_From @jfirebaugh on January 26, 2015 19:30_

As discussed in #199 and #223.

`fill-outline-color` is currently used in the bright and outdoors styles. @nickidlugash, @peterqliu can you comment on the importance of the property in those styles and in general? Would it be ok to remove for v7 without an immediate replacement via a "polygon" type or "multiple types" feature?

_Copied from original issue: mapbox/mapbox-gl-style-spec#240_

lucaswoj:
_From @peterqliu on January 26, 2015 19:35_

:+1: causes no breaking changes for me. 

Looking forward to polygons

lucaswoj:
_From @nickidlugash on January 26, 2015 20:18_

@jfirebaugh For these styles (and styles based on Mapbox Streets in general), the two features that depend the most on `fill-outline-color` are water and buildings. For these features, having an outline option is pretty important. 

> Would it be ok to remove for v7 without an immediate replacement via a "polygon" type or "multiple types" feature?

How long do we think it'll take to implement either a replacement (or optimize the performance of drawing lines enough that creating a second GL layer for the outlines won't kill performance)? If this is something that we're definitely doing soon, we can remove outlines for v7. I will want to manually update the v7 styles though, so I can optimize the way they look sans outline.

lucaswoj:
_From @jfirebaugh on October 7, 2016 0:37_

We really should do this. `fill-outline-color` adds complex special cases to the renderer for very little value.

lucaswoj:
_From @nickidlugash on October 7, 2016 14:38_

@jfirebaugh we use `fill-outline-color` in our private styles and public styles for buildings and road polygons. If we decide to remove it, we'll need to coordinate a styles update with the spec update.

lucaswoj:
_From @mourner on November 7, 2016 22:59_

I was initially worried about performance implications of this, but it looks like the only massive layer we use it for is buildings, which only appear on z15+, which is very easy to render compared to heavier zooms like z11-13. So I'm 👍 on this.

1ec5:
[...]

PR #4175, opened 2017-02-02
Add "default" property for functions

### LINKED ISSUE #4018: Expose "parseColor" method on style spec package
Users who want to use the `function` method on the style spec package sometimes need to parse colors too (see https://www.mapbox.com/mapbox-gl-js/example/data-join/). We could make their workflow easier and more consistent by exposing our implementation of `parseColor` on the style spec package. 

cc @kronick @ryanbaumann

### LINKED ISSUE #4124: Specify a default for property functions when property value is undefined
_From @lbud on August 16, 2016 20:30_

In the mapbox-gl-function [`feature-defaults` branch](https://github.com/mapbox/mapbox-gl-function/tree/feature-defaults) I'm adding `reference` as an argument so that when an input to a property function is undefined, it defaults to the spec default. However, it seems to me there should also be a way to specify a user-defined default for this behavior. Does it seem reasonable to add an optional key in a property function for `default`, like

```
extrusion-height: {
    stops: [[0, 0], [1000, 3000]],
    type: "exponential",
    property: "levels",
    default: 20
}
```

(so any extrusion without a `levels` property will have a height of 20, rather than the spec-defined default of 10)?

cc @lucaswoj @mollymerp @mourner 

_Copied from original issue: mapbox/mapbox-gl-style-spec#480_

### LINKED ISSUE #4144: Specify what happens when evaluating a categorical function input that's not in the domain
_From @jfirebaugh on November 28, 2016 19:28_

E.g. what does `{"type": "categorical", "property": "p", "stops": [["foo", 0]]}` evaluate to for the property value "bar"?

_Copied from original issue: mapbox/mapbox-gl-style-spec#582_

### LINKED ISSUE #4146: Specify what happens with exponential or interval property function type mismatches
_From @jfirebaugh on November 29, 2016 18:19_

For example, what is the result of evaluating the following function against a feature whose `p` property value is not numeric?

```
{
  "property": "p",
  "stops": [
    [0, 0],
    [10, 100]
  ]
}
```

cc @lucaswoj 

_Copied from original issue: mapbox/mapbox-gl-style-spec#588_

### LINKED ISSUE #4150: Specify what happens when an identity function evaluates to the wrong type
_From @jfirebaugh on November 30, 2016 1:29_

```
    {
      "id": "fill",
      "type": "fill",
      "source": "source",
      "source-layer": "layer",
      "paint": {
        "fill-opacity": {
          "property": "p",
          "type": "identity"
        }
      }
    }
```

What happens if there's a feature whose "p" property value isn't a number?

_Copied from original issue: mapbox/mapbox-gl-style-spec#602_

jfirebaugh:
It is used in the following circumstances:

* In categorical functions, when the feature value does not match any of the stop domain values.
* In property and zoom-and-property functions, when a feature does not contain a value for the specified property.
* In identity functions, when the feature value is not valid for the style property (for example, if the function is being used for a `circle-color` property but the feature property value is not a string or not a valid color).
* In interval or exponential property and zoom-and-property functions, when the feature value is not numeric.

If no default is provided, the style property's default is used in these circumstances.

Implementing this required breaking changes to the function API in mapbox-gl-style-spec. The export is now a function which accepts the function definition and the spec JSON for a style property. It handles color parsing and validation of feature values internally.

Fixes #4124
Fixes #4144
Fixes #4146
Fixes #4150
Fixes #4018

## Launch Checklist

 - [x] post benchmark scores

benchmark | master fac6358 | function-default a90eaeb
--- | --- | ---
**map-load** | 223 ms  | 99 ms 
**style-load** | 77 ms  | 77 ms 
**buffer** | 1,026 ms  | 1,036 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 3.2 ms, 0% > 16ms  | 3.1 ms, 0% > 16ms 
**query-point** | 0.87 ms  | 0.89 ms 
**query-box** | 64.55 ms  | 66.92 ms 
**geojson-setdata-small** | 7 ms  | 6 ms 
**geojson-setdata-large** | 148 ms  | 179 ms


js/style-spec/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## master
+
+* BREAKING CHANGE: the API for the `function` module has changed. The `interpolated` and `piecewise-constant` exports
+were replaced with a single unnamed export, a function which accepts an object conforming to the style spec "function"
+definition, and an object defining a style spec property. It handles color parsing and validation of feature values
+internally.
+* Functions now support a "default" property.
+* `parseColor` was promoted from gl-js.
```
lucaswoj: cc @ryanbaumann @kronick

js/style-spec/function/index.js
```diff
@@ -20,15 +28,38 @@ function createFunction(parameters, defaultType) {
         const zoomAndFeatureDependent = parameters.stops && typeof parameters.stops[0][0] === 'object';
         const featureDependent = zoomAndFeatureDependent || parameters.property !== undefined;
         const zoomDependent = zoomAndFeatureDependent || !featureDependent;
-        const type = parameters.type || defaultType || 'exponential';
+        const type = parameters.type || (propertySpec.function === 'interpolated' ? 'exponential' : 'interval');
```
lucaswoj: Have you thought about renaming the `propertySpec.function` "enum" values now that this is part of our public API?

js/style-spec/function/index.js
```diff
@@ -86,34 +108,30 @@ function createFunction(parameters, defaultType) {
...
             for (const z in featureFunctions) {
-                featureFunctionStops.push([featureFunctions[z].zoom, createFunction(featureFunctions[z])]);
+                featureFunctionStops.push([featureFunctions[z].zoom, createFunction(featureFunctions[z], propertySpec)]);
             }
             fun = function(zoom, feature) {
                 return outputFunction(evaluateExponentialFunction({
                     stops: featureFunctionStops,
                     base: parameters.base
-                }, zoom)(zoom, feature));
+                }, propertySpec, zoom)(zoom, feature));
```
lucaswoj: Why do you need to pass `propertySpec` to `evaluateExponentialFunction` both as an argument and embedded within `featureFunctionStops`?

js/style-spec/function/index.js
```diff
@@ -200,6 +223,10 @@ function interpolate(input, base, inputLower, inputUpper, outputLower, outputUpp
         return function() {
             const evaluatedLower = outputLower.apply(undefined, arguments);
             const evaluatedUpper = outputUpper.apply(undefined, arguments);
+            // Special case for fill-outline-color, which has no spec default.
+            if (evaluatedLower === undefined || evaluatedUpper === undefined) {
+                return undefined;
+            }
```
lucaswoj: ref #4088

package.json
```diff
@@ -16,8 +16,6 @@
     "@mapbox/unitbezier": "^0.0.0",
     "brfs": "^1.4.0",
     "bubleify": "^0.5.1",
-    "csscolorparser": "^1.0.2",
-    "csscolorparser": "~1.0.2",
```
lucaswoj: We don't currently install the `dependencies` from `js/style-spec/package.json` so this may still be needed

jfirebaugh: Isn't that what [this](https://github.com/mapbox/mapbox-gl-js/blob/fac63584971bd9c1fcf076046e3c87e38509b5bc/package.json#L141) does?

lucaswoj: Ah. Yes. The postinstall hook. 🚢

js/style-spec/function/index.js
lucaswoj: Ah. I misread the code. This is 👍

js/style-spec/function/index.js
jfirebaugh: Filed https://github.com/mapbox/mapbox-gl-js/issues/4194 for what I think we should do there.

js/style-spec/function/index.js
jfirebaugh: Where is it embedded within `featureFunctionStops`?

jfirebaugh: Oop, didn't see your followup comment. 👍

js/data/bucket/symbol_bucket.js
```diff
@@ -118,11 +118,12 @@ class SymbolBucket {
...
 
     populate(features, options) {
-        const layout = this.layers[0].layout;
+        const layer = this.layers[0];
+        const layout = layer.layout;
         const textFont = layout['text-font'];
         const iconImage = layout['icon-image'];
 
-        const hasText = layout['text-field'] && textFont;
+        const hasText = textFont && (!layer.isLayoutValueFeatureConstant('text-field') || layout['text-field']);
```
jfirebaugh: FYI @anandthakker.

anandthakker: > If the text-field property is data-driven, the value of layout['text-field'] is not defined.

@jfirebaugh ah, whoops - I made the bad assumption that `layout` contained the style JSON.

Issue #2730, opened 2016-06-13
Enable property functions for symbol paint properties


lucaswoj:
## Properties
- `icon-opacity` ✅ 
- `icon-color` ✅
- `icon-halo-color` ✅
- `icon-halo-width` ✅
- `icon-halo-blur` ✅
- `icon-translate`
- `icon-translate-anchor`
- `text-opacity` ✅
- `text-color` ✅
- `text-halo-color` ✅
- `text-halo-width` ✅
- `text-halo-blur` ✅
- `text-translate`
- `text-translate-anchor`

indus:
- ```icon-rotate```?

please don´t miss it; any traffic application or stuff like field-data (e.g. wind directions) would benefit.

lucaswoj:
@indus This ticket is about _paint_ properties. I just opened https://github.com/mapbox/mapbox-gl-js/issues/2736 for `icon-rotate` which is a _layout_ property.

usefulthink:
Is there any plan to implement this anytime soon? Or at least an idea as to what needs to happen in order to implement this?

I have the resources to help with an implementation (my team needs some of those features for a project), so it would be great if we could contribute to your project instead of rolling our own solution..

ivelander:
Any update on this enhancement? Our team is specifically looking for a solution for data-driven icon coloring. This seems to be the best route, given the unclear future and lack of documentation for SDF icons.

lucaswoj:
When there is an update to share, it will be done so on this ticket. I regret that other priorities stand in the way of our ability to work on this particular issue. 

> Our team is specifically looking for a solution for data-driven icon coloring. This seems to be the best route, given the unclear future and lack of documentation for SDF icons.

There are no plans to support `icon-color` for non-SDF icons. This has nothing to do with data-driven styling. I recommend seeking other solutions.

ivelander:
@lucaswoj Is there a technical barrier to supporting a data-driven `icon-color` for non-SDF icons, or is it just not in the plans? I've spent a few hours seeing if there might be a path to replicate what circle_bucket/draw_circle do to allow this and have had luck getting the shader to apply a single color across all icons, but no luck with per-feature coloring. Specifically I am uncertain where a call to `populatePaintArrays` would have the desired affect in `symbol_bucket`.

[...]

Issue #596, opened 2014-07-22
draw text and halo in one call


ansis:
Text halos are currently drawn in a separate call. Since text is usually drawn with a halo we should add halo support to the shader so that we can do it in one.

anandthakker:
Looked into this again in working on #4186 -- @jfirebaugh proposed `gl_FragDepth` as a promising way to handle the overlap issue, but, alas, the browser support for it is very limited (65% desktop, < 5% mobile)

PR #4186, opened 2017-02-02
Enable property functions for {text,icon}-{color,halo-color}

anandthakker:
Adds property function support for:
 - {text,icon}-color
 - {text,icon}-halo-color

Refs #2730 

## Checklist

- [x] post benchmark scores


Scarysize:
Nice! We implemented this earlier this week in our fork (for sdf icons).

Just a thought going through the Symbol Bucket code: Would it make sense to split up text (glyph) and icon rendering completely? The bucket code contains a lot of `if(isText/isIcon)` statements, bucket methods receive a gazillion parameters because they handle both types and even `draw_symbols` needs to differentiate between the two types.

anandthakker:
This is working for the simple `text-color` case, but not yet working for `icon-color` for some reason (in the added render test, both icons are coming up red) -- troubleshooting now.

Also, the existing approach to `halo` rendering causes a complication here.  We currently re-use the same program to render the text/icon halo, and then the text/icon itself.  Introducing property functions for paint props complicates this, because now we need a way to switch which property--e.g. `text-color` or `text-halo-color`--is being used for `color`.  I'm really not sure where/how this ought to happen.  @jfirebaugh @lucaswoj thoughts?

anandthakker:
@Scarysize good question!  I'm 💯 with you on those crazy long parameter lists.  I definitely think this part of the codebase could use a solid refactor, and splitting up text/icon might be a good route to take.... I'm personally not very familiar w/ the history of the symbol stuff, though, so there might be lurking subtleties that I'm not aware of.

lucaswoj:
> I'm really not sure where/how this ought to happen. 

The simplest design I can think of is adding a new parameter to `VertexArrayObject#bind` and `Buffer#setVertexAttribPointers`. 

This method is responsible for setting up the connections between shader attributes and struct array members. 

By default there is a 1:1 mapping of shader attributes to struct array members. We could solve this problem by making that relationship a little looser, allowing callers of `Buffer#setVertexAttribPointers` to define some kind of aliasing. 

```js
segment.vaos[layer.id].bind(gl, program, buffers.layoutVertexBuffer, buffers.elementBuffer, null, segment.vertexOffset, {
    attributeAliases: {
        'color': 'halo-color'
    }
});
```

Happy to voice on this if you have any follow-up questions. 😄

jfirebaugh:
Other options:

* Bind both colors, add a uniform boolean that selects between them and an `if` statement in the shader.
* Create separate shaders for halo and fill. Factor the common parts into reusable GLSL functions.
* Figure out how to draw halo and fill in a single draw call.

jfirebaugh:
> Figure out how to draw halo and fill in a single draw call.

Previously: https://github.com/mapbox/mapbox-gl-js/issues/596. @lbud, how conclusive was the outcome there? Is there no way to overcome the overlap issue?

anandthakker:
Update: ruled out consolidating the halo and fill in a single call (see #596 ).  Because @jfirebaugh indicated in chat that the [attributeAliases](https://github.com/mapbox/mapbox-gl-js/pull/4186#issuecomment-277046761) approach would likely be difficult to carry over to `-native`, going to try binding both sets of attributes, along with a uniform switch so the shader can decide.

anandthakker:
benchmark | master fac6358 | dds-symbol-paint 7872482
--- | --- | ---
**map-load** | 215 ms  | 135 ms 
**style-load** | 115 ms  | 109 ms 
**buffer** | 1,163 ms  | 1,187 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 5.1 ms, 0% > 16ms  | 5.3 ms, 0% > 16ms 
**query-point** | 0.97 ms  | 1.05 ms 
**query-box** | 70.54 ms  | 64.17 ms 
**geojson-setdata-small** | 6 ms  | 5 ms 
**geojson-setdata-large** | 145 ms  | 174 ms

js/data/bucket/symbol_bucket.js
```diff
@@ -358,7 +366,7 @@ class SymbolBucket {
             // the buffers for both tiles and clipped to tile boundaries at draw time.
             const addToBuffers = inside || mayOverlap;
             this.addSymbolInstance(anchor, line, shapedTextOrientations, shapedIcon, this.layers[0],
-                addToBuffers, this.symbolInstancesArray.length, this.collisionBoxArray, feature.index, feature.sourceLayerIndex, this.index,
```
anandthakker: Per conversation w @mollymerp, I replaced the (unused) `index` field with a `featureIndex`, which provides a mapping back to from the symbol instance to the feature it came from.  This allows us to retrieve the feature's properties down in `place()`.

This will likely need a bit of reworking when we return to the cross-source symbol placement quest, but since we don't know exactly what that'll look like, I think we should cross that bridge when we come to it.

anandthakker:
@jfirebaugh @lucaswoj k -- this is still a subset of the symbol paint properties (list is here https://github.com/mapbox/mapbox-gl-js/issues/2730 ), but it'd be good to go ahead and get some 👀 on it, esp. since this is my first foray into this area of the codebase.

js/data/bucket/symbol_bucket.js
lucaswoj: Any context on how the `index` field became unused?

shaders/symbol_sdf.fragment.glsl
```diff
@@ -1,21 +1,32 @@
...
+        color_to_use = halo_color;
+    }
+
     lowp float dist = texture2D(u_texture, v_tex).a;
     lowp float fade_alpha = texture2D(u_fadetexture, v_fade_tex).a;
     lowp float gamma = u_gamma * v_gamma_scale;
     lowp float alpha = smoothstep(u_buffer - gamma, u_buffer + gamma, dist) * fade_alpha;
 
-    gl_FragColor = u_color * (alpha * u_opacity);
+    gl_FragColor = color_to_use * (alpha * u_opacity);
```
lucaswoj: Instead of `color`, `halo_color`, and `color_to_use` how about `fill_color`, `halo_color`, and `color`?

js/data/program_configuration.js
```diff
@@ -210,7 +210,8 @@ function getPaintAttributeValue(attribute, layer, globalProperties, featurePrope
 }
 
 function normalizePaintAttribute(attribute, layer) {
-    const name = attribute.property.replace(`${layer.type}-`, '').replace(/-/g, '_');
+    const layerTypePrefix = layer.type === 'symbol' ? /text-|icon-/ : `${layer.type}-`;
+    const name = attribute.property.replace(layerTypePrefix, '').replace(/-/g, '_');
```
lucaswoj: Can you think of any cases where this magic behaviour might bite us later? I'd be interested to see if there was a way to specify this behaviour explicitly within `SymbolBucket` instead.

js/data/bucket/symbol_bucket.js
anandthakker: @lucaswoj I believe it never was used, but rather was there on the hypothesis that it might be needed during cross-source placement.  @mollymerp does that sound right to you?

js/data/program_configuration.js
anandthakker: > Can you think of any cases where this magic behaviour might bite us later?

@lucaswoj not offhand, but I'm 💯 with you in being uneasy / less than satisfied with this magic.

>  I'd be interested to see if there was a way to specify this behaviour explicitly within SymbolBucket instead.

👍 I'm thinking we can allow the bucket to explicitly define `attribute.name`, and then have `normalizePaintAttribute` just use that if it's there, instead of computing the attribute name from the paint property name.

shaders/symbol_sdf.fragment.glsl
anandthakker: Yeah.  I'd assumed I needed to stick with the attribute-name <=> paint-property-name-sans-layer-type-prefix mapping, but doing https://github.com/mapbox/mapbox-gl-js/pull/4186#pullrequestreview-19925092 would free us up to improve the naming here.

anandthakker:
@lucaswoj addressed your points re: bad magic and naming by explicitly setting the attribute names in `paintAttributes`.

When `attribute.name` isn't set explicitly, I'm personally okay with [keeping the pre-existing, slightly-less-magical convenience](https://github.com/mapbox/mapbox-gl-js/blob/52653a5626adc62911edcfc5ac0831f25fd69055/js/data/program_configuration.js#L215-L221) where we construct attribute name based on the property name.  It seems harmless, particularly since there's now a way to override it.

js/data/bucket/symbol_bucket.js
```diff
@@ -184,6 +192,12 @@ class SymbolBucket {
             // It's better to place labels on one long line than on many short segments.
             this.features = mergeLines(this.features);
         }
+
+        // set up this mapping to allow for constant-time lookups in place()
+        this.featuresByIndex = {};
```
jfirebaugh: Rather than create a separate index, can we attach a reference to the feature properties to the `symbolInstance`?

anandthakker: @jfirebaugh the symbolInstance is being stored (at `populate()` time) in a StructArray (`this.symbolInstancesArray`), so we can't attach a reference.

Granted, (a) we're not yet using `symbolInstancesArray` on a different thread, and (b) once we do (for cross-source placement), we'll have to find a different way to handle the paint attributes, since presumably whichever thread gets the array won't have the property data from the source.

jfirebaugh: Let's adjust/replace `symbolInstancesArray` based on the requirements here for data-driven properties. We can revisit its use in cross-source symbol placement when we're ready to work on that feature again.

anandthakker: Sure; if we're doing that I'd say we should go with 'replace' over 'adjust': adjusting it to these requirements would mean changing it from a StructArray, thus making it non-transferrable and no longer suitable for the cross-source placement project...  if that's going to be the case, I don't see much reason to hold on to this state (`symbolInstancesArray`, `symbolQuadsArray`) at the WorkerTile level.

I'm good to go ahead with this, but I just want to flag that dropping symbolInstancesArray will make the changes in this PR a notch more invasive, removing some stuff that we may end up having to partially redo later when we tackle cross-source placement -- but it's probably true that even if we left it in place, making it work with data-driven styling would anyway have caused significant changes to what's there now.

anandthakker: This is done in https://github.com/mapbox/mapbox-gl-js/pull/4217.   Tomorrow AM I'll rebase on that, which will allow us to just place a direct reference to `featureProperties` onto each `symbolInstance` as you suggested.

anandthakker:
@lucaswoj @jfirebaugh rebased onto the refactor in https://github.com/mapbox/mapbox-gl-js/pull/4217, so that one should get merged first, but after that's in I think this is ready.  Compare this PR's changes only at https://github.com/mapbox/mapbox-gl-js/compare/refactor-symbol-instances...dds-symbol-paint

Issue #3947, opened 2017-01-11
Remove plugin build artifacts from this repository


lucaswoj:
Including the plugin build artifacts in this repository isn't consistent with our general policies around checking in build artifacts or very convenient for plugin authors. We should experiment with other workflows such as

 - requiring plugins be published to https://unpkg.com/
 - requiring plugins handle their own publishing to s3
 - creating a script that downloads, builds, and publishes plugins

lucaswoj:
Done in #4305

PR #4191, opened 2017-02-02
Bump directions plugin

mollymerp:


mollymerp:
Includes 2 versions due to a patch bump that happened before this was merged. ready for 👀 @lucaswoj @tristen

tristen:
Not sure we need to include `v3.1.0` here if it's a broken version?

mollymerp:
Yeah, happy to remove too. Wasn't sure if it was important to keep as a record or not, but guessing that #3947 will remove all of them anyways so that probably makes sense.

mollymerp:
Deployed and manually tested the example page 👍

PR #4198, opened 2017-02-03
Prohibit identity zoom functions

### LINKED ISSUE #4151: Are "identity" zoom functions allowed?
_From @jfirebaugh on November 30, 2016 2:7_

```
    {
      "id": "fill",
      "type": "fill",
      "source": "source",
      "source-layer": "layer",
      "paint": {
        "fill-opacity": {
          "type": "identity"
        }
      }
    }
```

Is this allowed?

_Copied from original issue: mapbox/mapbox-gl-style-spec#603_

jfirebaugh:
Our function taxonomy benefits from being pared down where possible. Identity zoom functions don't seem useful, and are not implemented in native.

Fixes #4151.

### DIFF
--- a/js/style-spec/validate/validate_function.js
+++ b/js/style-spec/validate/validate_function.js
@@ -35,6 +35,10 @@ module.exports = function validateFunction(options) {
         }
     });
 
+    if (functionType === 'identity' && isZoomFunction) {
+        errors.push(new ValidationError(options.key, options.value, 'missing required property "property"'));
+    }
+
     if (functionType !== 'identity' && !options.value.stops) {
         errors.push(new ValidationError(options.key, options.value, 'missing required property "stops"'));
     }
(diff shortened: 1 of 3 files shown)


jplante:
This is a feature we would like.  We think it's useful.

For example - ALA Leaflet's Circle and CircleMarker - where the radius is based on either distance or pixels - we could use this for setting different radius for a Circle based on data identity attributes

jfirebaugh:
> different radius for a Circle based on data identity attributes

That sounds like a regular property-based identity function (which is supported), not an identity zoom function. An identity zoom function would be a function whose output is always the map's current zoom level. (If that's really what you want, you can create an exponential zoom function with base 1 and stops `[[0, 0], [maxZoom, maxZoom]]`, where `maxZoom` is some arbitrary value >= the maximum zoom your map allows.)

PR #4200, opened 2017-02-03
Add Map#setControlPosition method

mollymerp:
## Launch Checklist

 - [ ] post benchmark scores

This adds a method that allows a user to change the position of a control. I have started saving the controls added to the map so that users will be able to modify controls that are automatically added to the map that they do not have a reference to.

### DIFF
--- a/src/ui/control/logo_control.js
+++ b/src/ui/control/logo_control.js
@@ -20,7 +20,6 @@ class LogoControl {
     onAdd(map) {
         this._map = map;
         this._container = DOM.create('div', 'mapboxgl-ctrl');
-
         this._map.on('data', this._updateLogo);
         this._updateLogo();
         return this._container;
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -254,7 +254,7 @@ class Map extends Camera {
     removeControl(control) {
         control.onRemove(this);
         if (this._controls[control.constructor.name]) {
-            delete this._controls[control.constructor.name]
+            // delete this._controls[control.constructor.name];
         }
         return this;
     }
@@ -274,7 +274,7 @@ class Map extends Camera {
      */
 
     setControlPosition(control, position) {
-        const _control = typeof controlType === "string" ? this._controls[control] : control;
+        const _control = typeof control === "string" ? this._controls[control] : control;
         this.removeControl(_control);
         this.addControl(_control, position);
         return this;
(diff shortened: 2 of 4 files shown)


js/ui/map.js
```diff
@@ -249,6 +253,30 @@ class Map extends Camera {
      */
     removeControl(control) {
         control.onRemove(this);
+        if (this._controls[control.constructor.name]) {
+            // delete this._controls[control.constructor.name];
```
lucaswoj: Why the `//`?

js/ui/map.js
```diff
@@ -238,6 +240,8 @@ class Map extends Camera {
         } else {
             positionContainer.appendChild(controlElement);
         }
+        this._controls[control.constructor.name] = control;
```
lucaswoj: `Function#name` has many caveats (see the [MDN compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name)). It will be safer to use a more explicit naming system or eliminate the need for a naming system.

js/ui/map.js
```diff
@@ -249,6 +253,30 @@ class Map extends Camera {
...
+     * @returns {Map} `this`
+     * @example
+     * map.setControlPosition('LogoControl', 'top-right');
+     *
+     */
+
+    setControlPosition(control, position) {
+        const _control = typeof control === "string" ? this._controls[control] : control;
+        this.removeControl(_control);
+        this.addControl(_control, position);
```
lucaswoj: I infer that the raison d'être of this method is to allow users to change the position of the wordmark. Why not offer a constructor option to place the wordmark? 

```js
new mapboxgl.Map({logoPosition: 'top-right'})

// vs

var map = new mapboxgl.Map()
map.setControlPosition('LogoControl', 'top-right');
```

js/ui/map.js
mollymerp: ooh this was left by mistake after a debugging sesh

js/ui/map.js
mollymerp: We do have a constructor option for the wordmark, but there's no way currently to change the position of the control at runtime. Our solution to that for other controls is adding the control manually and saving a reference to it so you can add/remove to change the position when you want. This need comes up single page apps that change layout based on user interaction / window-size. If we don't want to support this though that's fine.

lucaswoj: Got it. Sounds good 👍

mollymerp:
going to close this because it hasn't been requested at all and seems unnecessary. will reopen if there are future user requests.

PR #4204, opened 2017-02-03
Restore standalone package.json for integration tests

jfirebaugh:
And use a `file:` devDependency in the top-level package.json. This allows gl-native to also use a `file:` devDependency to obtain the dependencies for integration tests without installing the full set of mapbox-gl-js dependencies.


jfirebaugh:
Native side, for reference: https://github.com/mapbox/mapbox-gl-native/commit/3c3845048c2b9bcbbad1640b08bdf69c334c6fd4

PR #4205, opened 2017-02-03
Added strict type checking to categorical style functions

### LINKED ISSUE #4174: Categorical functions became non-strict
_From @jfirebaugh on January 26, 2017 23:58_

Prior to https://github.com/mapbox/mapbox-gl-function/pull/32, categorical functions were strict about types: a stop such as `["1", 1]` would not match a numeric property value `1`, and vice versa. But now that stop domain values are used as object keys, everything is coerced to a string, and there's no distinction between strings and numbers.

For comparison, filters are [strict about types](https://github.com/mapbox/mapbox-gl-style-spec/issues/194), and functions should probably be consistent with that.

cc @kronick 

_Copied from original issue: mapbox/mapbox-gl-style-spec#669_

kronick:
Fixes #4174 by storing `typeof` from the first categorical stop. Return the default value if passed an input value of a different type.

cc @jfirebaugh for review

### DIFF
--- a/js/style-spec/function/index.js
+++ b/js/style-spec/function/index.js
@@ -48,6 +48,7 @@ function createFunction(parameters, propertySpec) {
 
         let innerFun;
         let hashedStops;
+        let categoricalKeyType;
         if (type === 'exponential') {
             innerFun = evaluateExponentialFunction;
         } else if (type === 'interval') {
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Thanks @kronick!

PR #4217, opened 2017-02-06
Refactor: remove SymbolInstancesArray, SymbolQuadsArray

anandthakker:
We aren't using SymbolInstancesArray (and SymbolQuadsArray) now, and will need a different approach when we revisit cross-source placement, so this PR removes them, in favor of storing the necessary state down on each `SymbolBucket` instance (rather than at the `WorkerTile` level).

 - [x] post benchmark scores


anandthakker:
benchmark | master 7b2b879 | refactor-symbol-instances 2d7b354
--- | --- | ---
**map-load** | 169 ms  | 118 ms 
**style-load** | 97 ms  | 97 ms 
**buffer** | 1,036 ms  | 1,010 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 4.7 ms, 0% > 16ms  | 4.7 ms, 0% > 16ms 
**query-point** | 0.82 ms  | 0.90 ms 
**query-box** | 60.38 ms  | 62.08 ms 
**geojson-setdata-large** | 154 ms  | 143 ms

package.json
```diff
@@ -69,6 +69,7 @@
     "remark-html": "^5.0.1",
     "request": "^2.79.0",
     "sinon": "^1.15.4",
+    "st": "^1.2.0",
```
jfirebaugh: It's [here](https://github.com/mapbox/mapbox-gl-js/blob/af078d0ebd717dc8af5ad47dfdec468d94ab3183/test/integration/package.json#L14-L14). Is that dependency not working for you?

anandthakker: Ah.  Yeah, it wasn't working for me -- e.g. the `start-debug` script in the root package.json failed to find `st`.

anandthakker:
@lucaswoj @jfirebaugh @mollymerp This is mostly done, as discussed here: https://github.com/mapbox/mapbox-gl-js/pull/4186#discussion_r99660763.

There's one render test mysteriously failing (because a road turned 180 degrees from waht's expected).  Chasing that down now, but I think this is probably okay to 👀  nonetheless.

package.json
jfirebaugh: Oh right, `start-server` uses the st CLI. 👍

mourner:
This is awesome!

anandthakker:
@lucaswoj rebased onto master - this should be good to go 🤞

PR #4241, opened 2017-02-09
Unexport util module

### LINKED ISSUE #1408: Unexport or document mapboxgl.util
Nothing there is public.

jfirebaugh:
Fixes #1408.


lucaswoj:
@mapbox/labs -- heads up that we're removing the public `util` module in new versions of GL JS. You'll need to use native language features or an external utility library.

PR #4243, opened 2017-02-09
Respect property functions in queryRenderedFeatures

### LINKED ISSUE #3604: "Map#queryRenderedFeatures" fails on features whose size is determined by a property function
queryRenderedFeatures fails to return results for circles that use a data-driven `circle-radius` property, due to the use of `paint['circle-radius']` [here](https://github.com/mapbox/mapbox-gl-js/blob/401089b324e45646efc91acd126c7ecb815507e2/js/data/feature_index.js#L122-L122).

This is a special case of #3044. Possibly a hard one to fix. @ansis did we ever discuss how to solve this?

anandthakker:
Approach:
 - Collect `paintPropertyStatistics` for data-driven paint properties while populating paint vertex arrays on the Worker side. (Specifically, store the `max` value for each such property.)
 - In `FeatureIndex#query`, use these max values when calculating how much additional padding to use when querying the grid index.  As features are indexed by their _source_ geometries (rather than their rendered geometries), this ensures that our grid query yields all potential hits for  the user's query.
 - Then, when filtering candidates from 👆 to those actually matching the user's query, use `layer.getPaintValue()` to obtain the actual paint property values for each feature.

Closes #3604

**Launch Checklist**

 - [x] post benchmark scores

**Benchmark**

benchmark | master 3c52ddb | 3604 429686e
--- | --- | ---
**map-load** | 163 ms  | 115 ms 
**style-load** | 96 ms  | 98 ms 
**buffer** | 1,051 ms  | 1,044 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 5 ms, 0% > 16ms  | 4.9 ms, 0% > 16ms 
**query-point** | 0.88 ms  | 1.05 ms 
**query-box** | 57.66 ms  | 82.00 ms 
**geojson-setdata-large** | 158 ms  | 184 ms


anandthakker:
The core strategy is working here, but note that both query benchmarks took a hit.  Gonna do some profiling to see if we can recover a bit.

anandthakker:
Ran benchmarks again after 2352e12:

benchmark | master 80aa693 | 3604 2352e12
--- | --- | ---
**map-load** | 154 ms  | 98 ms 
**style-load** | 96 ms  | 96 ms 
**buffer** | 1,104 ms  | 1,109 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 4.4 ms, 0% > 16ms  | 4.4 ms, 0% > 16ms 
**query-point** | 0.88 ms  | 1.20 ms 
**query-box** | 69.58 ms  | 77.37 ms 
**geojson-setdata-small** | 7 ms  | 5 ms 
**geojson-setdata-large** | 171 ms  | 184 ms 


Given the change from the previous benchmark run (query-point: +19% => +36%, query-box: +41% => +10%), it's hard to say what the true perf hit is here.  Based on profiling, I think we can attribute about 10% increase in queryRenderedFeatures time to the overhead of using `getPaintValue()` versus just accessing `layer.paint[]` -- i.e., 10% is basically the cost of correctness here.  My gut feeling is that the rest is sample noise and that we probably won't do much better.

cc: @jfirebaugh @mourner

src/data/bucket/symbol_bucket.js
```diff
@@ -205,6 +205,10 @@ class SymbolBucket {
             this.arrays.collisionBox.isEmpty();
     }
 
+    getPaintPropertyStatistics() {
+        return util.extend({}, this.arrays.icon.paintPropertyStatistics, this.arrays.glyph.paintPropertyStatistics);
```
jfirebaugh: Both of these are keyed by layer ID, so doesn't this overwrite icon statistics with glyph statistics?

test/integration/query-tests/circle-radius/zoom-and-property-function/style.json
```diff
@@ -0,0 +1,71 @@
...
+  "metadata": {
+    "test": {
+      "width": 64,
+      "height": 64,
+      "queryGeometry": [
+        32,
+        16
+      ],
+      "ignored": {
+        "native": "https://github.com/mapbox/mapbox-gl-native/issues/6518"
```
jfirebaugh: https://github.com/mapbox/mapbox-gl-native/issues/6518 is "Add support for identity function type" and is closed. Can you open a new parity issue and update this URL (and in the other test too)?

src/data/array_group.js
```diff
@@ -44,6 +44,7 @@ class ArrayGroup {
         if (ElementArrayType2) this.elementArray2 = new ElementArrayType2();
 
         this.layerData = {};
+        this.paintPropertyStatistics = {};
```
anandthakker: @jfirebaugh yeah, sure -- I actually started down that route, and then bailed because doing so would bifurcate the contents of `ArrayGroup#layerData` and the corresponding `BufferGroup#layerData`, and thought that might be (slightly) worse.   I don't have a strong opinion on this though, so deferring to you.

src/data/bucket/symbol_bucket.js
anandthakker: Whoops -- good catch, thanks.

anandthakker:
@jfirebaugh back to you

src/data/program_configuration.js
```diff
@@ -160,7 +160,21 @@ class ProgramConfiguration {
...
+    // Since this object is accessed frequently during populatePaintArray, it
+    // is helpful to initialize it ahead of time to avoid recalculating
+    // 'hidden class' optimizations to take effect
+    createPaintPropertyStatistics() {
+        const paintPropertyStatistics = {};
+        for (const attribute of this.attributes) {
+            if (attribute.dimensions !== 1) continue;
+            paintPropertyStatistics[attribute.property] = {
+                max: -Infinity
+            };
```
mourner: Do we anticipate adding other statistics in the near term?

PR #4244, opened 2017-02-09
Unexport and undocument LogoControl

jfirebaugh:
Other than the logoPosition map option, there's no public API -- it's shown/hidden automatically.

### DIFF
--- a/documentation.yml
+++ b/documentation.yml
@@ -18,7 +18,6 @@ toc:
   - GeolocateControl
   - AttributionControl
   - ScaleControl
-  - LogoControl
   - name: Handlers
     description: |
       Handlers add different kinds of interactivity to the map -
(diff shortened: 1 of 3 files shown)


mollymerp:
It's not yet completely shown/hidden automatically because there are still remaining API changes that are blocked. Some users still need to add the logo manually, but if we want to make the Control private and just have them use the HTML snippet to add it as has been the practice up til now that's fine too. 

@jfirebaugh I think there's also a line in `documentation.yml` for `LogoControl` that should be removed if we do decide to make it completely private

jfirebaugh:
> Some users still need to add the logo manually

I assumed this wasn't possible/needed because all instances of `LogoControl` are going to be either shown or hidden based on what `_updateLogo()` returns. So if a user added one manually in addition to the automatic one, there would be either none visible still, or two visible. Do we need to handle this some other way?

PR #4248, opened 2017-02-10
Fix #4246: RTL text plugin caused leak on Style#_remove()

### LINKED ISSUE #4246: RTL text plugin causes memory leaks
Every time a style is created, it pushes a callback onto `pluginAvailableCallbacks` which is never removed. The callback binds the style object, and the style references the map, so this leaks the entire shebang.

cc @ChrisLoer 

Refs #3264.

ChrisLoer:
The unit test verifies that the "deregister" call happens, but because actually exercising the plugin-loading code depends on the network, I fell back to manual testing.

I tested by creating two maps, verifying there were two callbacks registered, removing one map and seeing that the second callback was removed, then calling setRTLTextPlugin to verify the remaining callback was successfully called and then discarded.

@jfirebaugh @mourner

### DIFF
--- a/src/source/rtl_text_plugin.js
+++ b/src/source/rtl_text_plugin.js
@@ -1,29 +1,22 @@
 'use strict';
 
 const ajax = require('../util/ajax');
+const Evented = require('../util/evented');
 const window = require('../util/window');
 
-const pluginAvailableCallbacks = [];
 let pluginRequested = false;
 let pluginBlobURL = null;
 
+module.exports.evented = new Evented();
+
 module.exports.registerForPluginAvailability = function(callback) {
     if (pluginBlobURL) {
-        callback(pluginBlobURL);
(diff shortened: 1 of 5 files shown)


lucaswoj:
This is starting to feel like an implementation of [`Evented`](https://www.mapbox.com/mapbox-gl-js/api/#Evented). Would it be possible to use `Evented` to manage plugin available callbacks? That code is well tested and believed to be memory-leak free.

ChrisLoer:
@lucaswoj I switched to using`Evented`, but had to modify the implementation `Evented.once()` so that `off()` works on one-time listeners as well as listeners added with `on()`.

I also modified `Evented.listens()` so that it wouldn't return true when listeners had been added and then removed. The rest of this PR doesn't depend on that change, but at least based on my reading of the function documentation, it's the expected behavior.

PR #4254, opened 2017-02-10
MapEventData#isSourceLoaded fix

### LINKED ISSUE #3958: MapDataEvent#isSourceLoaded always returns false
Could this be because the `data` event (which triggers the `sourcedata` event) is fired immediately _before_ the `source.load` event changes [`SourceCache._sourceLoaded` to true](https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/source_cache.js#L30) ?

I thought that was the case, but a quick gut check to switch the function calls didn't change the result. But still it might need to be addressed after we figure out 👇 .

I think the root cause is that [`isSourceLoaded`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style.js#L378) is set right when `addSource` is called via [`setEventedParent`](https://github.com/mapbox/mapbox-gl-js/blob/25950e634b97db8e8c4e6c465b14e386a534a193/js/util/evented.js#L120) so `sourceCache.loaded()` is invoked only once, not every time the event fires like I believe was the intended behavior ??

mollymerp:
Working on fixing evaluation of `MapEventData#isSourceLoaded`

Fixes #3958 

Questions:
We only fire a `data` event with `dataType: source` when the TileJSON is loaded, but `isSourceLoaded: true` only occurs after all the tiles in for the source in the current viewport are loaded, so for `dataType:source` `EventData#isSourceLoaded` is still never `true`. Should we fire another `data` event of type `source` when all tiles are loaded `isSourceLoaded` flips to `true` ?

cc @lucaswoj @anandthakker @lbud 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/util/evented.js
+++ b/src/util/evented.js
@@ -71,8 +71,12 @@ class Evented {
      */
     fire(type, data) {
         if (this.listens(type)) {
-
             data = util.extend({}, data, {type: type, target: this});
+
+            // determine if source is loaded when event is fired
+            if (typeof data.isSourceLoaded === "function") {
+                data.isSourceLoaded = data.isSourceLoaded();
+            };
 
             // make sure adding or removing listeners inside other listeners won't cause an infinite loop
             const listeners = this._listeners && this._listeners[type] ? this._listeners[type].slice() : [];
--- a/src/util/evented.js
+++ b/src/util/evented.js
@@ -73,11 +73,6 @@ class Evented {
         if (this.listens(type)) {
             data = util.extend({}, data, {type: type, target: this});
 
-            // determine if source is loaded when event is fired
-            if (typeof data.isSourceLoaded === "function") {
-                data.isSourceLoaded = data.isSourceLoaded();
-            };
-
             // make sure adding or removing listeners inside other listeners won't cause an infinite loop
             const listeners = this._listeners && this._listeners[type] ? this._listeners[type].slice() : [];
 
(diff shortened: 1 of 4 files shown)


src/util/evented.js
```diff
@@ -71,9 +71,13 @@ class Evented {
      */
     fire(type, data) {
         if (this.listens(type)) {
-
             data = util.extend({}, data, {type: type, target: this});
 
+            // determine if source is loaded when event is fired
+            if (typeof data.isSourceLoaded === "function") {
+                data.isSourceLoaded = data.isSourceLoaded();
+            };
```
anandthakker: An alternative to this approach might be to change the semantics of `Evented#setEventedParent()` so that, if the evented parent `data` parameter is a function, it is evaluated during `fire()` rather than during `setEventedParent()`.

On the surface, that seems preferable to me because it avoids the hardcoding here... but it would require a review of `setEventedParent` to make sure we aren't counting on `data` being evaluated immediately anywhere else.

mollymerp:
to keep the scope of this PR under control, I suggest we ship this fix as-in and I will handle a larger rejiggering of the source `data` event in a separate PR.

src/util/evented.js
```diff
@@ -82,7 +81,7 @@ class Evented {
             }
 
             if (this._eventedParent) {
-                this._eventedParent.fire(type, util.extend({}, data, this._eventedParentData));
+                this._eventedParent.fire(type, util.extend({}, data, typeof this._eventedParentData === 'function' ? this._eventedParentData() : this._eventedParentData));
```
lucaswoj: Could you add a test that ensures the `eventedParentData` function is called per emit and passes on the latest & greatest data?

lucaswoj: Looks good otherwise 😄

mollymerp: Great -- added a test!

PR #4266, opened 2017-02-13
Clarify easing function requirements and clean up references to Anima…

### LINKED ISSUE #4182: Document easing function requirements
https://www.mapbox.com/mapbox-gl-js/api/#AnimationOptions says:

>easing (Function) : The animation's easing function.

But it doesn't say what arguments will be passed, nor what it should return.

I took a punt that it might be f(t, b, c, d) (from [here](http://gizma.com/easing/)) but that didn't seem to be right.

stevage:
…tionOptions.

Fixes #4182.

### DIFF
--- a/js/ui/camera.js
+++ b/js/ui/camera.js
@@ -29,7 +29,8 @@ const Evented = require('../util/evented');
  *
  * @typedef {Object} AnimationOptions
  * @property {number} duration The animation's duration, measured in milliseconds.
- * @property {Function} easing The animation's easing function.
+ * @property {Function} easing A function taking a time in the range 0..1 and returning a number where 0 is 
+ *                             the initial state and 1 is the final state.
  * @property {PointLike} offset `x` and `y` coordinates representing the animation's origin of movement relative to the map's center.
  * @property {boolean} animate If `false`, no animation will occur.
  */
@@ -302,7 +303,8 @@ class Camera extends Evented {
      * @param {Object} [options]
      * @param {boolean} [options.linear=false] If `true`, the map transitions using
      *     {@link Map#easeTo}. If `false`, the map transitions using {@link Map#flyTo}. See
-     *     {@link Map#flyTo} for information about the options specific to that animated transition.
+     *     those functions and @{link AnimationOptions} for information about options available.
+     * @param {Function} [options.easing] An easing function for the animated transition. See [AnimationOptions](#AnimationOptions).
      * @param {Function} [options.easing] An easing function for the animated transition.
      * @param {number} [options.padding=0] The amount of padding, in pixels, to allow around the specified bounds.
      * @param {PointLike} [options.offset=[0, 0]] The center of the given bounds relative to the map's center, measured in pixels.
@@ -411,7 +413,8 @@ class Camera extends Evented {
      * details not specified in `options`.
      *
      * @memberof Map#
-     * @param {CameraOptions|AnimationOptions} options Options describing the destination and animation of the transition.
+     * @param {Object} options Options describing the destination and animation of the transition.
+    *            Accepts [CameraOptions](#CameraOptions) and [AnimationOptions](#AnimationOptions).
      * @param {Object} [eventData] Data to propagate to any event listeners.
      * @fires movestart
      * @fires zoomstart
@@ -553,7 +556,6 @@ class Camera extends Evented {
      *     It does not correspond to a fixed physical distance, but varies by zoom level.
      * @param {number} [options.screenSpeed] The average speed of the animation measured in screenfuls
      *     per second, assuming a linear timing curve. If `options.speed` is specified, this option is ignored.
-     * @param {Function} [options.easing] An easing function for the animated transition.
      * @param {Object} [eventData] Data to propagate to any event listeners.
      * @fires movestart
      * @fires zoomstart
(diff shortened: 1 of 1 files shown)


js/ui/camera.js
```diff
@@ -411,7 +413,8 @@ class Camera extends Evented {
      * details not specified in `options`.
      *
      * @memberof Map#
-     * @param {CameraOptions|AnimationOptions} options Options describing the destination and animation of the transition.
+     * @param {Object} options Options describing the destination and animation of the transition.
+    *            Accepts [CameraOptions](#CameraOptions) and [AnimationOptions](#AnimationOptions).
```
anandthakker: Should we stick with `{CameraOptions|AnimationOptions}` as the type for this parameter?

js/ui/camera.js
```diff
@@ -302,7 +303,8 @@ class Camera extends Evented {
      * @param {Object} [options]
      * @param {boolean} [options.linear=false] If `true`, the map transitions using
      *     {@link Map#easeTo}. If `false`, the map transitions using {@link Map#flyTo}. See
-     *     {@link Map#flyTo} for information about the options specific to that animated transition.
+     *     those functions and @{link AnimationOptions} for information about options available.
+     * @param {Function} [options.easing] An easing function for the animated transition. See [AnimationOptions](#AnimationOptions).
      * @param {Function} [options.easing] An easing function for the animated transition.
```
anandthakker: `options.easing` is duplicated here.  In fact, for consistency, would it make sense to remove `options.easing` (and `options.offset`) in favor of setting the type of `[options]` to `{AnimationOptions}`?  I realize that, at present, this means more clicks for readers of the docs, but my preference would be to solve that problem somehow in the documentation generation step.

anandthakker:
Thanks for this, @stevage!

js/ui/camera.js
anandthakker: cc @lucaswoj @mollymerp - I remember a brief discussion about this, but I can't find the comment/thread.

js/ui/camera.js
```diff
@@ -29,7 +29,8 @@ const Evented = require('../util/evented');
  *
  * @typedef {Object} AnimationOptions
  * @property {number} duration The animation's duration, measured in milliseconds.
- * @property {Function} easing The animation's easing function.
+ * @property {Function} easing A function taking a time in the range 0..1 and returning a number where 0 is 
+ *                             the initial state and 1 is the final state.
```
lucaswoj: What makes you think that the function is expected to be nondecreasing? (I'm ok with using math class language!)

js/ui/camera.js
```diff
@@ -302,7 +303,8 @@ class Camera extends Evented {
      * @param {Object} [options]
      * @param {boolean} [options.linear=false] If `true`, the map transitions using
      *     {@link Map#easeTo}. If `false`, the map transitions using {@link Map#flyTo}. See
-     *     {@link Map#flyTo} for information about the options specific to that animated transition.
+     *     those functions and @{link AnimationOptions} for information about options available.
```
lucaswoj: I'm 👎 on this edit. The `@see` tag already suggests the user may see those methods for more info. We don't need to state this explicitly.

js/ui/camera.js
lucaswoj: > in favor of setting the type of `[options]` to `{AnimationOptions}`?

Good idea. If we go this route, we may also want to:

 - rename/alias `options.linear` to `options.animate`
 - add `options.duration` (this will just be passed to `Map#flyTo`)

js/ui/camera.js
anandthakker: Eh, I think it was just an overly-cautious red flag instinct -- it's an assumption I could imagine making if I were implementing animations w/ a custom easing function -- but now that I look at `camera.js`, I don't see any obvious cases where we assume it.

stevage: I didn't test, but does the return value have to be within 0..1? Easing functions in systems like D3 generally don't.

js/ui/camera.js
stevage: Ok. To me it wasn't obvious that you can pass additional options and they will be respected by easeTo or flyTo respectively.

js/ui/camera.js
stevage: Whoops, messed up my git. I'll wait till you decide. (If you can add commits, feel free.)

js/ui/camera.js
stevage: That seemed slightly misleading, because the object doesn't have to be *either* a CameraOptions *or* an AnimationOptions, but can share attributes from both.

lucaswoj: In Flow, our type definition standard, the `|` operator (the "union" operator) indicates that an object may "share attributes from both: https://flowtype.org/docs/union-intersection-types.html#_

lucaswoj: You can think of it more like the bitwise `|` than the logical `||`

js/ui/camera.js
lucaswoj: I suggest we keep this PR simple and make the switch to `AnimationOptions` in a follow up. Would you mind ticketing this?

anandthakker: ✅ Ticketed at https://github.com/mapbox/mapbox-gl-js/issues/4286

js/ui/camera.js
stevage: Thanks. Should have known you'd be following a standard :)

(I actually found that Flow doc page a bit ambiguous, but the page on [disjoint unions](https://flowtype.org/docs/disjoint-unions.html#_) makes clear that unions are non-disjoint by default)

PR #4267, opened 2017-02-14
Revamp "Quickstart" instructions in API docs

lucaswoj:
This PR makes it easier to work with our Quickstart instructions, removes a step from the release process, removes some duplicated/out-of-sync documentation, and prepares us for some upcoming changes to mapbox.com.

Also:

 - adds `copyable` block plugin to API docs
 - fixes `site.js` path when running API docs locally
 - removes stay `circle.yml` file
 - removes `INSTALL.md` in favor of ‘Quickstart” instructions

screen shot 2017-02-13 at 4 15 51 pm
screen shot 2017-02-13 at 4 15 59 pm


docs/api/assets/site.js
```diff
@@ -53,8 +53,8 @@ window.addEventListener('hashchange', function() {
```
jfirebaugh: It looks like the directory containing this file is copied from `_theme/assets` when docs are built. Let's remove it from source control and gitignore the whole directory.

lucaswoj: Good catch. Done & tested. Will merge when CI passes.

Issue #4211, opened 2017-02-05
Adding a symbol layer with a Polygon source leaves map in unpredictable state as of 0.29.0


dpieri:
**Version:** 0.29.0+

### Steps to Trigger Behavior

 1. Use any version >= 0.29.0 (I've checked them all).
 2. Add a FeatureCollection with some polygons as a GeoJSON source.
 3. Add a symbol layer that uses this source.

### Expected Behavior
Map works like normal and looks like this (note the anchors):
<img width="719" alt="screen shot 2017-02-05 at 2 14 53 pm" src="https://cloud.githubusercontent.com/assets/539310/22630359/979efa8e-ebad-11e6-9c4c-aba6bf177e68.png">

### Actual Behavior
Map looks like this:
<img width="717" alt="screen shot 2017-02-05 at 2 14 37 pm" src="https://cloud.githubusercontent.com/assets/539310/22630360/9cf79342-ebad-11e6-96eb-7df4121ab120.png">

Map is in a strange state. It does not render the symbols, but it also seems to stop rending other features properly. It is less pronounced in the example I've included. In my codebase the map stops loading new vector layers as you move the map around (raster layers update just fine). In both cases, if you leave the browser window open long enough it will crash.

If you switch to version 0.28.0 these problems are not present.

#### Examples:
Broken in 0.29.0: https://jsbin.com/viyosubosa/1/edit?html,output
Working in 0.28.0: https://jsbin.com/koloyojiwe/edit?html,output

**Note:** If you take the 0.29.0 example and switch the layer filter from `Polygon` to `LineString`, it works fine.

AndyMoreland:
I am also effected by this bug.

PR #4270, opened 2017-02-14
fixed infinite loop when initial centroid returns NaN due to bad geometry

craiig:
Hi all,
I had a hang issue when drawing symbols from some gnarly multipolygons from a local geojson and I traced it to bestCell.d being set to NaN. I'm not totally sure this is the best overall fix, but this works for me. 

The polygon ring input to this function that was causing the issue was: [[4221, 3907], [4207,3909], [4221, 3907]]. A line! 😭  For what it's worth, my source geometry doesn't seem to include a line but it seems like this was interacting with tiles to produce a line.

## Launch Checklist

### DIFF
--- a/src/util/find_pole_of_inaccessibility.js
+++ b/src/util/find_pole_of_inaccessibility.js
@@ -52,7 +52,7 @@ module.exports = function (polygonRings, precision, debug) {
         const cell = cellQueue.pop();
 
         // update the best cell if we found a better one
-        if (cell.d > bestCell.d) {
+        if (cell.d > bestCell.d || !bestCell.d) { 
             bestCell = cell;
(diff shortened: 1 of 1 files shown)


craiig:
Just FYI I've found more issues in this function when testing with different geometry. I'll submit another PR with the full fixes soon

anandthakker:
cc @mourner

mourner:
There's a fix for this issue in the polylabel repo here https://github.com/mapbox/polylabel/commit/a8b579222712d8b2c6f61748e226be412a011b36, although this fix seems good too.

anandthakker:
Thanks @craiig !

lucaswoj:
Does this fix #4211?

craiig:
It's possible (when combined with the other pull request I submitted earlier today), as #4211 seems to describe exactly the problems I encountered

lucaswoj:
Great news @craiig!

Issue #4284, opened 2017-02-16
If npm is used instead of yarn, warn the contributor rather than silently installing possibly broken dependencies


jfirebaugh:
### Steps to Trigger Behavior

 1. Run `npm install` in a mapbox-gl-js repository checkout.
 
### Expected Behavior

An error such as the following:

> It looks like you're using npm to install mapbox-gl-js dependencies for development. mapbox-gl-js uses yarn to manage dependencies, not npm. Please see the README for further details: `<link>`.

### Actual Behavior

npm installs versions of dependencies that don't match the known good versions in `yarn.lock`. They might not work at all, such as in [this example](https://github.com/mapbox/mapbox-gl-native/pull/8078#issuecomment-280204666), leading to confusion and frustration for the potential contributor.

jfirebaugh:
Snippet from Mapbox Studio that does this:

    "preinstall": "ps -fp $PPID | grep -q yarn || (echo '\\033[0;33mNPM not supported, please use Yarn.\\033[0m'; exit 1) && yarn cache clean",

mollymerp:
closed by #4287

PR #4287, opened 2017-02-16
update docs to replace npm with yarn

### LINKED ISSUE #4283: Update CONTRIBUTING.md: npm ⇢ yarn
CONTRIBUTING.md needs to use `yarn` commands rather than `npm` commands. It should also include a sentence or two saying that yarn is required and common ways to obtain it.

mollymerp:
Closes #4283 #4284

cc @jfirebaugh

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,6 @@ xcode-select --install
 
 Install [node.js](https://nodejs.org/) version 4 or greater
 ```bash
-brew update
 brew install node
 ```
 Install [yarn](https://yarnpkg.com/en/)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,12 @@ xcode-select --install
 
 Install [node.js](https://nodejs.org/) version 4 or greater
 ```bash
+brew update
 brew install node
+```
+Install [yarn](https://yarnpkg.com/en/)
+```bash
+brew install yarn
 ```
 
 Clone the repository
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -126,6 +126,7 @@
     "test-flow": "flow .",
     "test-cov": "nyc --require=flow-remove-types/register --reporter=text-summary --reporter=lcov --cache run-s test-unit test-render test-query",
     "prepublish": "in-publish && run-s build-dev build-min || not-in-publish",
+    "preinstall": "ps -fp $PPID | grep -q yarn || (echo '\\033[0;33mNPM not supported, please use Yarn.\\033[0m'; exit 1) && yarn cache clean",
```
jfirebaugh: I suspect this command will fail on Windows, since it doesn't have `ps`, `grep`, etc.

CONTRIBUTING.md
```diff
@@ -11,8 +11,13 @@ xcode-select --install
 
 Install [node.js](https://nodejs.org/) version 4 or greater
 ```bash
+brew update
```
lucaswoj: This is not necessary in newer releases of brew https://github.com/Homebrew/brew/pull/679 😄

mourner:
I think this is a bit too much — in #4284, the concern is mostly about installing dependencies _for development_, while this PR will throw an error for all downstream NPM users, which we should not do IMO. @jfirebaugh thoughts? I would suggest reverting or updating with a milder solution before we release.

jfirebaugh:
Oh, yeah, I didn't think of that. Yeah, we need to detect whether the `preinstall` is being run as part of an `npm install` in mapbox-gl-js, versus when mapbox-gl-js is being installed as a dependency in a downstream project.

mollymerp:
Ah good catch @mourner - would replacing the thrown error with a warning be a good stopgap?

mourner:
@mollymerp much better, but it would be great to avoid the warning for downstream users as it can cause confusion. Maybe lets revert and follow-up with a full solution?

mollymerp:
I think adding this condition to the error will fix the problem of the error throwing when `mapbox-gl` is installed as a dependency: 
`process.env.npm_package_name === \"mapbox-gl\"`

If this sounds reasonable to you I'll make a PR with changing the `preinstall` script to:

```json
"preinstall": "node -e 'if(!/yarn\\.js$/.test(process.env.npm_execpath) && \
process.env.npm_package_name === \"mapbox-gl\")throw \
new Error(\"NPM not supported, please use Yarn to install dependencies ($ yarn install).\")'"
```

mourner:
If it works, sure — let's do it!

mollymerp:
Haha I tested it and it doesn't work - still errors when mapbox-gl is installed as a downstream dependency. I can't find any info on the internet about how to detect this so I will just revert for now 😞

mollymerp:
fixed in #4319

PR #4295, opened 2017-02-17
Removed line in fitBounds that was setting bearing to 0

### LINKED ISSUE #4221: Box zoom resets map bearing
**mapbox-gl-js version**: 0.32.1

### Steps to Trigger Behavior

 1. Drag with the right mouse button to set an arbitrary map bearing and tilt
 2. Perform a box-zoom by shift-dragging the mouse

### Expected Behavior

Map bearing ("rotation") and tilt should be kept unchanged.

### Actual Behavior

Map bearing is reset to zero (north up) while map tilt is kept unchanged. Rotating the map on a box zoom just feels.... like weird UX.

yappel:
Fixes #4221 
Inside the method fitBounds of the Camera class the bearing was specifically reset to 0 every time it was called, causing the "problem" as described in #4221 . However the resetting of the bearing was also documented in the methods comments, so it could be intended behavior. Either this PR fixes the issue or the issue #4221 should be closed.

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -306,7 +306,6 @@ class Camera extends Evented {
 
     /**
      * Pans and zooms the map to contain its visible area within the specified geographical bounds.
-     * This function will also reset the map's bearing to 0 if bearing is nonzero.
      *
      * @memberof Map#
      * @param {LngLatBoundsLike} bounds Center these bounds in the viewport and use the highest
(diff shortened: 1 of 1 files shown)


jfirebaugh:
I'm pretty sure this is intentional, and without it `fitBounds` isn't guaranteed to actually fit the bounds. The underlying issue, which I think has been discussed in prior issues, is that the `fitBounds` parameter is an axis-aligned bounding box, which isn't ideal for a rotatable map.

yappel:
I see. This issue is a lot bigger than I thought and it seems there are already issues and PR's present to redesign this part of the code.

PR #4314, opened 2017-02-22
Clarify how event data "propagates".

stevage:


js/ui/camera.js
```diff
@@ -556,7 +556,7 @@ class Camera extends Evented {
      *     It does not correspond to a fixed physical distance, but varies by zoom level.
      * @param {number} [options.screenSpeed] The average speed of the animation measured in screenfuls
      *     per second, assuming a linear timing curve. If `options.speed` is specified, this option is ignored.
-     * @param {Object} [eventData] Data to propagate to any event listeners.
+     * @param {Object} [eventData] Data to propagate to any event listeners. Properties on the object will be copied onto the event argument.
```
lucaswoj: These two sentences are each trying to convey the same bit of information. Do you think we could rewrite to convey it clearly and succinctly in one?

`Additional [event data]{@link Evented#fire} to be included with all events triggered by this method`

stevage: Well, I tried a few versions. The first sentence is the what and why, the second one is the explicit how. In your version (as at present), it's not clear *how* the data is included - my assumption was it would be a second object passed: f(eventData, extraData).

Maybe:

`Additional properties to be added to [event data]{@link Evented#fire} passed to events triggered by this method` ?

jfirebaugh: > Additional properties to be added to the event objects of events triggered by this method.

lucaswoj:
Rad! Thanks!

PR #4320, opened 2017-02-22
Rework Camera#easeTo and Transform#pointLocation

### LINKED ISSUE #4178: easeTo goes in the wrong direction
**mapbox-gl-js version**:0.32.1

### Steps to Trigger Behavior

 1. View https://bl.ocks.org/stevage/c80f9b241866642abdb2e6da8231cde6
 2. Wait for the second camera transition

### Expected Behavior
Camera should ease south

### Actual Behavior
Camera eases north, then miraculously reappears from the south.

It works properly with flyTo instead of easeTo.

joswinter:
The changes in this pull request fix #4178 and #3112.

bug description easing

Previously when using the Camera#easeTo function the following can happen where the camera only eases to the north. The path of the camera is illustrated in the left image where the camera moves across both poles when easing between two coordinates _A_ and _B_, this is unwanted behavior. The easeTo should ease in the direction of the shortest path to the endpoint, The correct behaviour is illustrated in the right image. The incorrect behaviour can be observed in the following gist: https://bl.ocks.org/joswinter/f2e9bc8413ed00875ac06eca35d7b83d

These bugs were already fixed by @yeldarby in pull request #3130 which hasn't been merged. I continued working on the code in that pull request and reworked the structure of Camera#easeTo and Transform#pointLocation as commented by @lucaswoj. The Transform#pointLocation now has an optional viewport argument which is used to calculate the pointLocation for the given point. I also changed the Camera#easeTo implementation so that all tests in camera.test.js pass.


src/geo/transform.js
```diff
@@ -239,12 +239,36 @@ class Transform {
...
+                center: this.center,
+                zoom: this.zoom,
+                bearing: this.bearing,
+                pitch: this.pitch
+            };
+
+            if ('center' in v) this.center = v.center;
+            if ('zoom' in v) this.zoom = v.zoom;
+            if ('bearing' in v) this.bearing = v.bearing;
+            if ('pitch' in v) this.pitch = v.pitch;
```
lucaswoj: This is a step in the right direction. I'm still uncomfortable with this stateful & implicit setting & resetting of parameters on `Transform`. Passing arguments this way is surprising and creates potential for bugs & unintended side effects. `coordinateLocation` and `coordinatePoint` worked with this viewport object directly rather than reading from `Transform`.

src/ui/camera.js
```diff
@@ -546,7 +575,9 @@ class Camera extends Evented {
                 tr.pitch = interpolate(startPitch, pitch, k);
             }
 
-            tr.setLocationAtPoint(toLngLat, fromPoint.add(toPoint.sub(fromPoint)._mult(k)));
+            var lng = interpolate(startCenter.lng, toLocation.lng, k2);
+            var lat = interpolate(startCenter.lat, toLocation.lat, k2);
+            tr.center = LngLat.convert([lng, lat]);
```
lucaswoj: Is it possible to write test for this new functionality? Regression tests for https://github.com/mapbox/mapbox-gl-js/issues/3112 & https://github.com/mapbox/mapbox-gl-js/issues/4178?

src/geo/transform.js
joswinter: I get what you mean by that, I looked into it. I could add an extra argument to the `coordinateLocation` and `pointCoordinate` methods but the reason why I used this implicit setting of parameters is to make sure that the pixelMatrix is calculated for the viewport. What do you suggest that I should do concerning the `_calcMatrices` function, should I rewrite it to allow a viewport parameter? `_calcMatrices` uses the following fields:

- `this.x`
- `this.y`
- `this._fov`
- `this._pitch`
- `this.width`
- `this.height`
- `this.worldSize`

If I'm going to do it with a viewport parameter I'm also a bit afraid to introduce a lot of duplicate code because every parameter has a seperate check and a lot of other `Transform` fields are dependent on the ones used in the viewport such as `this.worldSize` which is dependent on `this.scale` which is dependent on `this.zoom`.

lucaswoj: Ideally, we would work together to find a way to add a viewport parameter to `_calcMatrices`, reduce the amount of input needed by `_calcMatricies` where possible, and refactor to eliminate any duplicate code. Does that sound doable based on your experience in this part of the codebase?

joswinter:
I created a regression test for the `Camera#easeTo` function. It seems to be working but it is not 100% exact. In the in the [test](https://github.com/mapbox/mapbox-gl-js/pull/4320/commits/aaede12921d539387273b1b03f7472cc30e8195d), I check whether the camera is easing in the right direction in the case of this [example](https://bl.ocks.org/joswinter/f2e9bc8413ed00875ac06eca35d7b83d).

I have added a `_calcMatricesViewport` and a `_constrainViewport` method which helps in calculating the `inversePixelMatrix` for a viewport @lucaswoj. However, there are two tests that keep failing:

- the `zooms with specified offset` test.
- the `zooms with specified offset relative to viewport on a rotated camera` test.

The functionality should be the same as it was in commit [aaede](https://github.com/mapbox/mapbox-gl-js/pull/4320/commits/aaede12921d539387273b1b03f7472cc30e8195d), do the fields in `camera.js` are influenced by something else which differ in passing the viewport or in setting all transform fields manually from the viewport?

joswinter:
@lucaswoj I got the tests working now. I want to start refactoring now but I still have some questions. The first two things that should happen is that:

- `_constrain` and `_constrainViewport` need to be merged. How should I merge these two because at the moment one of them returns something and the other sets the field of the class directly?
- `_calcMatrices` and `_calcMatricesViewport` need to be merged. The same holds here, but I have some concerns with the `projMatrix` field that is being set in the middle of the `_calcMatrices` function.

Do you have any suggestion how I should merge these functions? For most of the functions in the `Transform` class I already added a `viewport` parameter and a line at the start with `if (viewport === undefined) viewport = this.getViewport();`, do you think that is the correct way to go?

src/geo/transform.js
```diff
@@ -263,17 +287,24 @@ class Transform {
...
+        const zoomedCoord = coord.zoomTo(viewport.zoom);
+        const worldSize = this.tileSize * this.zoomScale(viewport.zoom);
+        const x = (zoomedCoord.column * this.tileSize) * 360 / worldSize - 180;
+        const y2 = 180 - (zoomedCoord.row * this.tileSize) * 360 / worldSize;
+        const y = 360 / Math.PI * Math.atan(Math.exp(y2 * Math.PI / 180)) - 90;
         return new LngLat(
-            this.xLng(zoomedCoord.column * this.tileSize),
-            this.yLat(zoomedCoord.row * this.tileSize));
+            x,
+            y);
```
lucaswoj: Minor nit: can we make this all one line?

src/geo/transform.js
```diff
@@ -283,8 +314,10 @@ class Transform {
         const coord0 = [p.x, p.y, 0, 1];
         const coord1 = [p.x, p.y, 1, 1];
 
-        vec4.transformMat4(coord0, coord0, this.pixelMatrixInverse);
-        vec4.transformMat4(coord1, coord1, this.pixelMatrixInverse);
+        const pixelMatrixInverse = this._calcMatrices(viewport);
```
lucaswoj: This operation is expensive enough that we will need to cache `pixelMatrixInverse` for at least the global viewport.

src/geo/transform.js
```diff
@@ -334,82 +367,117 @@ class Transform {
...
+            sy = maxY - minY < this.size.y ? this.size.y / (maxY - minY) : 0;
         }
 
         if (this.lngRange) {
-            minX = this.lngX(this.lngRange[0]);
-            maxX = this.lngX(this.lngRange[1]);
-            sx = maxX - minX < size.x ? size.x / (maxX - minX) : 0;
+            minX = (180 + this.lngRange[0]) * worldSize / 360;
+            maxX = (180 + this.lngRange[1]) * worldSize / 360;
+            sx = maxX - minX < this.size.x ? this.size.x / (maxX - minX) : 0;
```
lucaswoj: Can you explain what changed & why in this algorithm?

mourner:
Happy to see a PR that fixes the issues, but surprised that it requires so many complicated changes, many of them introducing new expensive calculations. I'll need to spend some time on the problem to see if there's possibly a more elegant approach we're not yet seeing.

lucaswoj:
@mourner Should we delay merging this PR until you've had a chance to prototype a more elegant implementation?

mourner:
@lucaswoj yes, let's delay for a few days! We'll have a hard time maintaining the code merged as is.

mourner:
I'm moving with an alternative PR, submitting soon. Basically the root problem here is not `pointLocation` implementation, is that we use it for `easeTo` interpolations in the first place. Moving the approach closer to how `flyTo` works fixes the problem with simpler code changes, so I'll close the PR for now, but it'll be there in case we want to revisit making `pointLocation` more flexible. Thanks for working on this!

PR #4324, opened 2017-02-23
Display an error message for impossible values of minZoom and maxZoom

### LINKED ISSUE #2264: Display an error message for impossible values of minZoom, zoom, and maxZoom
https://jsfiddle.net/7d365j4s/

In this example,
- `zoom` is greater than `maxZoom`
- `minZoom` is greater than `maxZoom`

**mapbox-gl-js version**: v0.15.0
### Steps to Trigger Behavior

Click on link.
### Expected Behavior

Either error messages of contradictory parameters (preferred), and/or refuse to load the map.
### Actual Behavior

No errors, and map does not zoom.

joswinter:
Fixes #2264

When creating a map where holds that `minZoom > maxZoom` the map is loaded and the user cannot zoom. The changes in this pull request fire an error when the above case holds when initiating a Transform or when setting the minZoom or maxZoom. See the following example: https://jsfiddle.net/7d365j4s/

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -133,7 +133,7 @@ class Map extends Camera {
     constructor(options) {
         options = util.extend({}, defaultOptions, options);
 
-        if (options.minZoom && options.maxZoom && options.minZoom > options.maxZoom) {
+        if (options.minZoom != null && options.maxZoom != null && options.minZoom > options.maxZoom) {
             throw new Error(`maxZoom must be greater than minZoom`);
         }
 
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -133,6 +133,10 @@ class Map extends Camera {
     constructor(options) {
         options = util.extend({}, defaultOptions, options);
 
+        if (options.minZoom && options.maxZoom && options.minZoom > options.maxZoom) {
+            throw new Error(`maxZoom must be greater than minZoom`);
+        }
+
         const transform = new Transform(options.minZoom, options.maxZoom, options.renderWorldCopies);
         super(transform, options);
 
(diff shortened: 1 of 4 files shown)


joswinter:
Thanks for the review @jfirebaugh!

I investigated why the validation wasn't working in the case of issue #2264. The validation only works when directly using the `Map#set{Min,Max}Zoom` functions however when constructing the following map `createMap({minZoom:10, maxZoom:5})`, no exception will be thrown. This happens because in `Map#constructor` the maxZoom and minZoom values are set by passing it to `Transform#constructor` which doesn't check whether the minZoom and maxZoom values are valid. I added a simple check to `Map#constructor` which checks if the minZoom and maxZoom values are valid, if these values are invalid then a `Error('maxZoom must be greater than minZoom')` is thrown.

src/ui/map.js
```diff
@@ -133,6 +133,10 @@ class Map extends Camera {
     constructor(options) {
         options = util.extend({}, defaultOptions, options);
 
+        if (options.minZoom && options.maxZoom && options.minZoom > options.maxZoom) {
```
lucaswoj: There's one case that's slipping through the cracks here:

`maxZoom: 0, minZoom: 1`

This could be fixed by changing the test to `options.minZoom != null && options.maxZoom != null`

test/unit/ui/map.test.js
```diff
@@ -628,6 +628,13 @@ test('Map', (t) => {
         t.end();
     });
 
+    t.test('throw on maxZoom smaller than minZoom at init', (t) => {
+        t.throws(() => {
+            createMap({minZoom:10, maxZoom:5});
+        });
```
lucaswoj: We should test the error message mentions `maxZoom` and `minZoom`. Otherwise, we might pass because of an error line `createMap is undefined`.

joswinter:
Thanks for reviewing @lucaswoj! I implemented the requested changes.

PR #4329, opened 2017-02-25
Layer event delegation

### LINKED ISSUE #1002: Add a "layers" parameter to "Evented#on"

jfirebaugh:
Fixes #1002. I found that in order to really simplify most examples, we need special handling for `mouseenter`/`mouseleave` in addition to basic event delegation. But once we have that, a lot of common cases are simplified.

## Launch Checklist

 - [x] should the special behavior apply to `mouseenter`/`mouseleave` or `mouseover`/`mouseout`?
   - both
 - [x] document the behavior for enter/leave (or over/out)
 - [x] leave/out event should also be triggered when cursor leaves the map element

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -515,6 +515,136 @@ class Map extends Camera {
     }
 
     /**
+     * Adds a listener for events of a specified type.
+     *
+     * @method
+     * @name on
+     * @memberof Map
+     * @instance
+     * @param {string} type The event type to add a listen for.
+     * @param {Function} listener The function to be called when the event is fired.
+     *   The listener function is called with the data object passed to `fire`,
+     *   extended with `target` and `type` properties.
+     * @returns {Map} `this`
+     */
+
+    /**
+     * Adds a listener for events of a specified type occurring on features in a specified style layer.
+     *
+     * @param {string} type The event type to listen for; one of `'mousedown'`, `'mouseup'`, `'click'`, `'dblclick'`,
+     * `'mousemove'`, `'mouseenter'`, `'mouseleave'`, `'mouseover'`, `'mouseout'`, `'contextmenu'`, `'touchstart'`,
+     * `'touchend'`, or `'touchcancel'`. `mouseenter` and `mouseover` events are triggered when the cursor enters
+     * a visible portion of the specified layer from outside that layer or outside the map canvas. `mouseleave`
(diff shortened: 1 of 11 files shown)


src/ui/map.js
```diff
@@ -503,6 +503,91 @@ class Map extends Camera {
...
+     * @param {string} type The event type to listen for.
+     * @param {string} [layer] If provided, only events whose location is within a visible
+     * feature will trigger the listener. The event will have a `features` property containing
+     * an array of the matching features.
+     * @param {Function} listener The function to be called when the event is fired.
+     * @returns {Map} `this`
+     */
+    on(type, layer, listener) {
+        if (listener === undefined) {
+            return Camera.prototype.on.call(this, type, layer);
```
mourner: `return super.on(type, layer)`?

src/ui/map.js
```diff
@@ -503,6 +503,91 @@ class Map extends Camera {
...
+     * Removes a listener previously added with `Map#on`.
+     *
+     * @param {string} type The event type previously used to install the listener.
+     * @param {string} [layer] The layer previously used to install the listener, if any.
+     * @param {Function} listener The function previously installed as a listener.
+     * @returns {Map} `this`
+     */
+    off(type, layer, listener) {
+        if (listener === undefined) {
+            return Camera.prototype.on.call(this, type, layer);
```
mourner: `super` here too.

mourner:
> should the special behavior apply to mouseenter/mouseleave or mouseover/mouseout?

I think we could do both and alias one for the other, since there should be no difference in context of a map (in DOM context, there's only a difference when considering child elements, but we have a different model).

mourner:
Any updates on this PR?

jfirebaugh:
Finally was able to carve out time to finish this off. 🏁

PR #4333, opened 2017-02-27
Set default maxZoom to 22

jingsam:
As I know, all mapbox public sources such as satellite and streets are 0~22 zoom level. And maps in Mapbox studio can be zoomed into 22 max level. However, the default maxZoom level for mapbox-gl-js is 20. This cause inconsistency when a developer zoomed his maps in Mapbox studio to 22 and get stuck at 20 in his application. The shared URL provided by  mapbox studio also stuck at 20. I know developers can set default maxZoom of mapbox-gl-js to 22 explicitly, but it needs some effort for them to figure out the solution.

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -23,7 +23,7 @@ const LogoControl = require('./control/logo_control');
 const isSupported = require('mapbox-gl-supported');
 
 const defaultMinZoom = 0;
-const defaultMaxZoom = 20;
+const defaultMaxZoom = 22;
 const defaultOptions = {
     center: [0, 0],
     zoom: 0,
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Thanks @jingsam. Can you fix the test failure?

jingsam:
* failed line-width very-overscaled

I don't know why this test failed, please help me check it out

jfirebaugh:
I think the problem is [this line](https://github.com/mapbox/mapbox-gl-js/blob/878008895104b67861ce73f65809f7f6f0ed7263/test/integration/render-tests/line-width/very-overscaled/style.json#L12-L12), which requests a zoom of 21. Prior to your change, the default max zoom would affect the result, producing an effective zoom level of 20. Now it's actually rendering at 21. Try changing the line to `"zoom": 20`.

jingsam:
OK for now

jfirebaugh:
Thanks!

1ec5:
mapbox/mapbox-gl-native#8318 tracks doing the same in the native SDKs.

PR #4342, opened 2017-02-28
Fix #4337 eventData not passed during flyTo

### LINKED ISSUE #4337: eventData not passed during flyTo
**v0.32.1**:

### Steps to Trigger Behavior

 1. register event listener for "movestart"
 2. issue a flyTo to the current center passing arbitrary eventData

### Expected Behavior
eventData received by the "movestart" handler

### Actual Behavior
eventData not received by the "movestart" handler
This is because:
`

        // When u₀ = u₁, the optimal path doesn’t require both ascent and descent.
        if (Math.abs(u1) < 0.000001) {
            // Perform a more or less instantaneous transition if the path is too short.
            if (Math.abs(w0 - w1) < 0.000001) return this.easeTo(options);
`
around this [LOC](https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/camera.js#L739) when i type this.

The use case is simple:
- the user puts the map in the mode where it follows the user's position as the user moves.
- the user un-puts the map from that mode by dragging the map or looking something up elsewhere.

therefore:
When the map in the 'follow' mode moves in response to the user's position it should remain in the 'follow' mode. if the map moves for any other reason, it should exit the 'follow' mode - writing this as sometimes people give me interesting workarounds.

remster:
Fix to https://github.com/mapbox/mapbox-gl-js/issues/4337 and the corresponding test.

## Launch Checklist

 - [-] post benchmark scores

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -736,7 +736,7 @@ class Camera extends Evented {
         // When u₀ = u₁, the optimal path doesn’t require both ascent and descent.
         if (Math.abs(u1) < 0.000001) {
             // Perform a more or less instantaneous transition if the path is too short.
-            if (Math.abs(w0 - w1) < 0.000001) return this.easeTo(options);
+            if (Math.abs(w0 - w1) < 0.000001) return this.easeTo(options, eventData);
 
             const k = w1 < w0 ? -1 : 1;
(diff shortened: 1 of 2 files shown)


lucaswoj:
Tested this locally because the s3 outage is affecting CircleCI.

PR #4347, opened 2017-03-01
fire sourcedata when source metadata and tiles are loaded

mollymerp:
This PR does a couple things in order to make the `sourcedata` event more useful for users:
- for vector tile sources, we currently fire `sourcedata` when the tilejson loads, but that is not really useful for users, so this PR will fire the event when the source + all tiles in the current viewport are loaded
- we used to call `SourceCache#reload` on the `sourcedata` event, but that was causing all vector tiles to be loaded twice on initial load of the source, and because this is only meant to reload geojson sources after `setData` and image/video/canvas sources after `setCoordinates` I created a new event (😬) to handle this called `source.update`

### DIFF
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -108,11 +108,9 @@ class GeoJSONSource extends Evented {
                 this.fire('error', {error: err});
                 return;
             }
-            // `update` is included here to prevent a race condition where `Style#_updateSources` is called
-            // before the TileJSON arrives. this makes sure the tiles needed are loaded once TileJSON arrives
-            // ref: https://github.com/mapbox/mapbox-gl-js/pull/4347#discussion_r104418088
+            // although GeoJSON sources contain no metadata, we fire this event to let the SourceCache
+            // know its ok to start requesting tiles.
             this.fire('data', {dataType: 'source', sourceDataType: 'metadata'});
-            this.fire('data', {dataType: 'source', sourceDataType: 'update'});
         });
     }
 
@@ -134,7 +132,7 @@ class GeoJSONSource extends Evented {
             if (err) {
                 return this.fire('error', { error: err });
             }
-            this.fire('data', {dataType: 'source', sourceDataType: 'update'});
+            this.fire('data', {dataType: 'source', sourceDataType: 'content'});
         });
 
         return this;
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -108,8 +108,8 @@ class GeoJSONSource extends Evented {
                 this.fire('error', {error: err});
                 return;
             }
-            this.fire('data', {dataType: 'source'});
             this.fire('source.load');
+            this.fire('data', {dataType: 'source'});
         });
     }
 
@@ -132,7 +132,7 @@ class GeoJSONSource extends Evented {
             if (err) {
                 return this.fire('error', { error: err });
             }
-            this.fire('data', {dataType: 'source'});
+            this.fire('source.update');
         });
 
         return this;
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -126,13 +126,13 @@ class GeoJSONSource extends Evented {
      */
     setData(data) {
         this._data = data;
-
         this.fire('dataloading', {dataType: 'source'});
         this._updateWorkerData((err) => {
             if (err) {
                 return this.fire('error', { error: err });
             }
             this.fire('source.update');
+            this.fire('data', {dataType: 'source'});
         });
 
         return this;
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -108,7 +108,7 @@ class GeoJSONSource extends Evented {
                 this.fire('error', {error: err});
                 return;
             }
-            this.fire('data', {dataType: 'source', metadata: true});
+            this.fire('data', {dataType: 'source', sourceDataType: 'metadata' });
         });
     }
 
@@ -130,7 +130,7 @@ class GeoJSONSource extends Evented {
             if (err) {
                 return this.fire('error', { error: err });
             }
-            this.fire('data', {dataType: 'source', update: true});
+            this.fire('data', {dataType: 'source', sourceDataType: 'update'});
         });
 
         return this;
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -108,8 +108,7 @@ class GeoJSONSource extends Evented {
                 this.fire('error', {error: err});
                 return;
             }
-            this.fire('source.load');
-            this.fire('data', {dataType: 'source'});
+            this.fire('data', {dataType: 'source', metadata: true});
         });
     }
 
@@ -131,8 +130,7 @@ class GeoJSONSource extends Evented {
             if (err) {
                 return this.fire('error', { error: err });
             }
-            this.fire('source.update');
-            this.fire('data', {dataType: 'source'});
+            this.fire('data', {dataType: 'source', update: true});
         });
 
         return this;
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -108,7 +108,7 @@ class GeoJSONSource extends Evented {
                 this.fire('error', {error: err});
                 return;
             }
-            this.fire('data', {dataType: 'source', sourceDataType: 'metadata' });
+            this.fire('data', {dataType: 'source', sourceDataType: 'metadata'});
         });
     }
 
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -108,7 +108,11 @@ class GeoJSONSource extends Evented {
                 this.fire('error', {error: err});
                 return;
             }
+            // `update` is included here to prevent a race condition where `Style#_updateSources` is called
+            // before the TileJSON arrives. this makes sure the tiles needed are loaded once TileJSON arrives
+            // ref: https://github.com/mapbox/mapbox-gl-js/pull/4347#discussion_r104418088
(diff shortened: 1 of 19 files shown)


lucaswoj:
> - for vector tile sources, we currently fire `sourcedata` when the tilejson loads, but that is not really useful for users, so this PR will fire the event when the source + all tiles in the current viewport are loaded

I would like to advocate for `sourcedata` to be fired whenever a source changes or loads new data (not just when it finishes loading) on the basis that

 - it stays true to the heritage of the  [`data` event](https://nodejs.org/dist/latest-v6.x/docs/api/stream.html#stream_event_data), indicating that some data has arrived
 - it would allow us to simplify the API by removing the `tiledata` event
 - it creates a clear path towards custom sources which are always updating (for example due to animations or real time data) or untiled
 - it avoids stepping on the toes of `dataend` (perhaps we need a `sourcedataend` event)

> - we used to call `SourceCache#reload` on the `sourcedata` event, but that was causing all vector tiles to be loaded twice on initial load of the source, and because this is only meant to reload geojson sources after `setData` and image/video/canvas sources after `setCoordinates` I created a new event (😬) to handle this called `source.update` 

Can we eliminate the need for this new event by calling `SourceCache#reload` directly within `setData` and `setCoordinates`?

anandthakker:
> Can we eliminate the need for this new event by calling SourceCache#reload directly within setData and setCoordinates?

@lucaswoj as of now, sources are not 'aware of' / provided a reference to their containing SourceCache.  We could either refactor to add this dependency now, or temporarily keep the (private) `source.update` event and wait for the upcoming SourceCache refactor.

src/source/image_source.js
```diff
@@ -115,7 +114,8 @@ class ImageSource extends Evented {
         centerCoord.column = Math.round(centerCoord.column);
         centerCoord.row = Math.round(centerCoord.row);
 
-        this.minzoom = this.maxzoom = centerCoord.zoom;
+        // TODO this is making tests fail... what is it for?
+        // this.minzoom = this.maxzoom = centerCoord.zoom;
```
mollymerp: @lbud would love to know what this line is for. for some reason it's behaving in unexpected ways with the new eventing system and making the canvas source tests fail 😭

src/source/source_cache.js
```diff
@@ -475,6 +477,7 @@ class SourceCache extends Evented {
             clearTimeout(this._timers[id]);
             this._timers[id] = undefined;
         }
+        // TODO should we keep this??
         this._source.fire('data', { tile: tile, coord: tile.coord, dataType: 'tile' });
```
mollymerp: @lucaswoj @anandthakker do you think we need to fire an event when a tile is removed now that `tiledata` is ⚰️? I feel like this isn't useful, but let me know if you disagree.

mollymerp:
In its current state this PR:
- removes `source.load` and replaces it with a `sourcedata` with a `metadata` property on the `MapEventData`
- removes `tiledata` and replaces it with a `sourcedata` event with tile information in the `MapDataEvent`
- adds an `update` property on the event data that triggers a reload of the tiles for use with mutable sources (eg `GeoJSONSource#setData` and `ImageSource#setCoordinates`) in order to prevent double loading of sources when they're initially added to the map

TODO:
- figure out whats going on with the canvas source + tests
- figure out what to do with the remaining vestiges of the `tiledata` event (on `removeTile` and `redoPlacement`)

cc @lucaswoj @anandthakker @lbud

src/source/image_source.js
anandthakker: @lbud might be able to say more, but my recollection from a while back is that this had to do with 'tricking' the source cache into always requesting the 'right' tile from the ImageSource.

src/source/source_cache.js
anandthakker: I agree that this event isn't useful -- if we're not using it anywhere internally, then I'm all for removing it.

What _might_ be useful someday in the future would be a way for users to listen for changes to the list of currently renderable/rendered tile coordinates -- but this doesn't really provide that, and anyway that's a super non-urgent nice-to-have.

src/source/source_cache.js
```diff
@@ -27,23 +27,26 @@ class SourceCache extends Evented {
...
-            if (this._sourceLoaded && event.dataType === 'source') {
+        this.on('data', function(e) {
+            // this._sourceLoaded signifies that the TileJSON is loaded if applicable
+            // if the source type does not come with a TileJSON, the flag signifies the
+            // source data has loaded (i.e geojson has been tiled on the worker and is ready)
+            if (e.dataType === 'source' && e.metadata) this._sourceLoaded = true;
+
+            // for sources with mutable data, this event fires when the underlying data
+            // to a source is changed. (i.e. GeoJSONSource#setData and ImageSource#serCoordinates)
+            if (this._sourceLoaded && e.dataType==="source" && e.update) {
```
anandthakker: This is great ❤️ 

Notes/questions:

1) We should update the internal documentation [here](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/source.js#L53-L54) to reflect these changes to the Source interface contract.

2) Is `e.metadata` too generic a name?  I'm not even sure if I like this idea yet, but what about something like `e.sourceDataType == 'metadata'` and `e.sourceDataType == 'update'`?  (Basically having `sourceDataType` be analogous to `dataType`, but specific to source data events.)

3) Is there still a possibility of a race condition here for vector sources?  We need `SourceCache#update()` to be called after the TileJSON has loaded.  This currently happens either here (for events with `e.update`) or via `Style#_updateSources()`.   I'm imagining this problematic timeline:
- map.addSource('mysource', ...) (plus map.addLayer() using `mysource`) => requests TileJSON, marks `map._sourceDirty`.
- Next render frame fires; map._sourcesDirty being true leads to `style._updateSources()`... but since the TileJSON is still in flight, we don't get a `SourceCache#update()` for `mysource`.
- TileJSON lands => VectorTileSource fires `data` with `e.metadata` => `mysource`'s SourceCache marks itself `_sourceLoaded`.
- But now, the actual tiles for `mysource` are not loaded until something _else_ causes `_sourcesDirty` flag to be set.

src/source/image_source.js
lbud: Yeah, I think @anandthakker is right — because of https://github.com/mapbox/mapbox-gl-js/issues/3186 (👺) we have to trick the source cache into thinking there's only one appropriate tile to ever request for image + image-inheriting sources.

src/source/source.js
```diff
@@ -50,7 +50,7 @@ exports.setType = function (name, type) {
  * @param {string} options.type The source type, matching the value of `name` used in {@link Style#addSourceType}.
  * @param {Dispatcher} dispatcher A {@link Dispatcher} instance, which can be used to send messages to the workers.
  *
- * @fires load to indicate source data has been loaded, so that it's okay to call `loadTile`
```
anandthakker: Small addition: "@fires data **with `sourceDataType: 'metadata'`** to indicate..." -- since that's what source cache now requires of the event fired by `Source`.

src/source/geojson_source.js
```diff
@@ -108,8 +108,11 @@ class GeoJSONSource extends Evented {
                 this.fire('error', {error: err});
                 return;
             }
-            this.fire('data', {dataType: 'source'});
-            this.fire('source.load');
+            // `update` is included here to prevent a race condition where `Style#_updateSources` is called
+            // before the TileJSON arrives. this makes sure the tiles needed are loaded once TileJSON arrives
+            // ref: https://github.com/mapbox/mapbox-gl-js/pull/4347#discussion_r104418088
+            this.fire('data', {dataType: 'source', sourceDataType: 'metadata'});
```
anandthakker: In this case (GeoJSONSource), the term `'metadata'` is _slightly_ odd, since there isn't really any metadata for geojson sources.  As such, it might be more helpful to replace the comment about `update` in favor of a comment reminding us that we're firing `metadata` to notify the source cache that it's 👍 to ask us for tiles.   

(I.e., reading this source, I think it's `metadata` that might be confusing; reading the vector/raster sources, it's `update` that might be confusing)

test/unit/source/vector_tile_source.test.js
```diff
@@ -52,21 +54,25 @@ test('VectorTileSource', (t) => {
...
     });
 
-    t.test('fires "data" event', (t) => {
+    t.test('fires event with metadata property', (t) => {
         window.server.respondWith('/source.json', JSON.stringify(require('../../fixtures/source')));
         const source = createSource({ url: "/source.json" });
-        source.on('data', t.end);
+        source.on('data', (e)=>{
+            if (e.sourceDataType === 'metadata') t.end();
+        });
```
anandthakker: Should this test be checking for the `'update'` (or `'content'`) event, rather than `'metadata'`?

src/source/geojson_source.js
mollymerp: oh good catch! this should be in `vector_tile_source` and not here.

mollymerp:
benchmark | master 5fa31ef | source-event 94c35d6
--- | --- | ---
**map-load** | 450 ms  | 86 ms 
**style-load** | 120 ms  | 116 ms 
**buffer** | 900 ms  | 899 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 4.2 ms, 0% > 16ms  | 4.3 ms, 0% > 16ms 
**query-point** | 0.71 ms  | 0.74 ms 
**query-box** | 57.15 ms  | 56.11 ms 
**geojson-setdata-small** | 5 ms  | 1 ms 
**geojson-setdata-large** | 92 ms  | 93 ms 
 
🕐 benchmarks for posterity 🕙

PR #4350, opened 2017-03-01
Fix map.isMoving for drag-pan/rotate

mourner:
Just something I noticed and fixed while hacking on custom GL layers. Continuation of #3941. Another way to do this would be to call `dragPan/Rotate.isActive()` in `map.isMoving()` implementation, but I think this way is clearer.

Note that we currently don't cover interaction handlers with unit tests so no changes on this front.

cc @stepankuzmin


stepankuzmin:
Great!

Issue #4345, opened 2017-03-01
Ensure tiles with a HTTP "Expires" header in the past don't cause an infinite loop


lucaswoj:
Tiles which have a HTTP `Expires` header in the past cause an infinite loop of requests. 

Our use of the `Expires` header as a trigger for a tile reload may be considered a violation of the HTTP spec:

> An expiration time cannot be used to force a user agent to refresh its display or reload a resource; its semantics apply only to caching mechanisms -- https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.2.1

GL Native addressed this with https://github.com/mapbox/mapbox-gl-native/pull/4019

Issue #4252, opened 2017-02-10
Tiles flicker in Microsoft Edge


bcartmell:
In Microsoft Edge (tested on Windows 10 64-bit), map tiles flicker as if continuously being loaded/unloaded.  I observed this issue while working on my current project (proprietary).  However, while debugging, I noticed that the example at https://www.mapbox.com/mapbox-gl-js/example/satellite-map/ also demonstrates this behavior.  Copied example into JSbin (http://jsbin.com/dabulemife/edit?html,output), but I cannot reproduce the issue there.

**mapbox-gl-js version**: 0.32.1

### Steps to Trigger Behavior

 1.  load https://www.mapbox.com/mapbox-gl-js/example/satellite-map/
 2.  May need to pan and/or zoom
 3.  Observe tiles flickering

### Expected Behavior
  Tiles load and display cleanly

### Actual Behavior
  Tiles flicker as if continuously loading and unloading

jaapster:
I see the same thing happening every 60 secs. in other browsers in an app that we're building. Turns out "Cache Max-Age" was set at 60 seconds.

bcartmell:
Thanks for your insight Jaapster.  This seems to be fixed now.  I loaded the app in Edge again to check the headers that I get.  max-age is 43200 and s-maxage is 604800 and I can no longer reproduce the problem in my app or in the example map.

bcartmell:
Issue reappeared today.  cache-control valuse have not changed.

rgtaylor1989:
Having the same issue when cutting a raster on and off. In this example I have a button that turns the "aerials" off and on. The photo below is from edge. Those white tiles constantly flash and if you look at the network traffic to user, it is constantly trying to get the same raster tiles over and over again. Eventually it will make Edge unresponsive if you leave open network tab open because of the traffic it is creating.
 
![sample1](https://cloud.githubusercontent.com/assets/7094059/23236537/99ad8138-f927-11e6-87cf-8a801dcea7f0.png)

When reloading the page from this flashing state, the canvas never attempts to grab the vector tiles (according to the network tab under the console) and the webgl never populates even though map javascript loads. This might not be related to this issue though.

![sample2](https://cloud.githubusercontent.com/assets/7094059/23236919/257665ee-f929-11e6-95dc-35978eb73cd6.png)

bcartmell:
I've verified that this issue only occurs with cached tiles, disabling cache completely resolves it.

[...]

PR #4351, opened 2017-03-02
Handle for cases where tile returned from server is already

lbud:
Parity with https://github.com/mapbox/mapbox-gl-native/pull/4019 ; 
fixes https://github.com/mapbox/mapbox-gl-js/issues/4345 and https://github.com/mapbox/mapbox-gl-js/issues/4252.

This PR does the following:
* implements exponential backoff for retrying tile requests that are received with expiration times already in the past 
* disables crossfading on raster tiles that are refreshed from expiration refresh mechanisms (should fix flickering described in #4252)
* fixes bug wherein a tile was setting its expiration data before it had a `timeAdded` property, thus setting expiration incorrectly
* adds a `noRefreshOnExpiration` map option that disables refreshing-upon-tile-expiration mechanism altogether

cc @lucaswoj @anandthakker @jfirebaugh

### DIFF
--- a/src/render/draw_raster.js
+++ b/src/render/draw_raster.js
@@ -122,7 +122,13 @@ function getFadeValues(tile, parentTile, layer, transform) {
         // if no parent or parent is older, fade in; if parent is younger, fade out
         const fadeIn = !parentTile || Math.abs(parentTile.coord.z - idealZ) > Math.abs(tile.coord.z - idealZ);
 
-        const childOpacity = (fadeIn && tile.refreshedUponExpiration) ? 1 : util.clamp(fadeIn ? sinceTile : 1 - sinceParent, 0, 1);
+        const childOpacity = tile.refreshedUponExpiration ? 1 : util.clamp(fadeIn ? sinceTile : 1 - sinceParent, 0, 1);
+
+        // we don't crossfade tiles that were just refreshed upon expiring:
+        // once they're old enough to pass the crossfading threshold
+        // (fadeDuration), unset the `refreshedUponExpiration` flag so we don't
+        // incorrectly fail to crossfade them when zooming
+        if (tile.refreshedUponExpiration && sinceTile >= 1) tile.refreshedUponExpiration = false;
 
         if (parentTile) {
             return {
--- a/src/render/draw_raster.js
+++ b/src/render/draw_raster.js
@@ -122,7 +122,7 @@ function getFadeValues(tile, parentTile, layer, transform) {
         // if no parent or parent is older, fade in; if parent is younger, fade out
         const fadeIn = !parentTile || Math.abs(parentTile.coord.z - idealZ) > Math.abs(tile.coord.z - idealZ);
 
-        const childOpacity = util.clamp(fadeIn ? sinceTile : 1 - sinceParent, 0, 1);
+        const childOpacity = (fadeIn && tile.refreshedUponExpiration) ? 1 : util.clamp(fadeIn ? sinceTile : 1 - sinceParent, 0, 1);
 
         if (parentTile) {
             return {
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -10,6 +10,8 @@ const featureFilter = require('../style-spec/feature_filter');
 const CollisionTile = require('../symbol/collision_tile');
 const CollisionBoxArray = require('../symbol/collision_box');
 
+const CLOCK_SKEW_RETRY_TIMEOUT = 30000;
+
 /**
  * A tile object is the combination of a Coordinate, which defines
  * its place, as well as a unique ID and data tracking for its content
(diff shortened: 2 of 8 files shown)


src/source/tile.js
```diff
@@ -29,7 +31,13 @@ class Tile {
         this.sourceMaxZoom = sourceMaxZoom;
         this.buckets = {};
         this.expires = null;
-        this.cacheControl = null;
+        this.expired = false;
```
anandthakker: On the theory that all time-based logic is horrible to edit/debug, could we rename `this.expires` to something like `this.expirationTime` and `this.expired` to `this.isExpired`?

src/source/tile.js
```diff
@@ -29,7 +31,13 @@ class Tile {
...
+
+        // Counts the number of times a response was already expired when
+        // received. We're using this to add a delay when making a new request
+        // so we don't have to keep retrying immediately in case of a server
+        // serving expired tiles.
+        this.expiredRequests = 0;
```
anandthakker: Similar to above... maybe rename to `expiredRequestCount`?

src/source/tile.js
```diff
@@ -195,18 +203,61 @@ class Tile {
...
+            const now = Date.now();
+
+            if (this.expires > now) {
+                this.expired = false;
+            } else if (!prior) {
+                this.expired = true;
+            } else if (this.expires < prior) {
+                // Expiring date is going backwards:
+                // fall back to exponential backoff
+                this.expired = true;
```
anandthakker: 😱

src/source/tile.js
```diff
@@ -195,18 +203,61 @@ class Tile {
...
+                } else {
+                    // Assume that either the client or the server clock is wrong and
+                    // try to interpolate a valid expiration date (from the client POV)
+                    // observing a minimum timeout.
+                    this.expires = now + Math.max(delta, CLOCK_SKEW_RETRY_TIMEOUT);
+
+                }
+            }
+
+            if (this.expired) {
```
anandthakker: Am I reading it right that  `this.expired` only used within this method?  If so, I think it'd be better as a local var.

test/unit/source/tile.test.js
```diff
@@ -191,12 +192,33 @@ test('expiring tiles', (t) => {
...
+        tile.timeAdded = Date.now();
+
+        tile.setExpiryData({
+            cacheControl: 'max-age=10'
+        });
+
+        const expiryTimeout = tile.getExpiryTimeout();
+        t.ok(expiryTimeout >= 8000 && expiryTimeout <= 10000, 'expiry timeout as expected when fresh');
+
+        tile.expiredRequests = 1;
```
anandthakker: I think it would be slightly more reliable if we used the public `tile.setExpiryData()` api here instead of directly setting `tile.expiredRequests`... is it feasible to do this (together with either `setTimeout` to get past the expiration time or else maybe by using Sinon to fake the current time)?

src/ui/map.js
```diff
@@ -54,7 +54,9 @@ const defaultOptions = {
 
     trackResize: true,
 
-    renderWorldCopies: true
+    renderWorldCopies: true,
+
+    noRefreshOnExpiration: false
```
anandthakker: I'm in favor of avoiding negatives in variable names when possible, as they can make conditional logic harder to read. (I may well have stole this idea/opinion from someone here, but not actually sure.) . Proposed alternative: `refreshExpiredTiles`, with a default value of `true`?

src/render/draw_raster.js
```diff
@@ -122,7 +122,7 @@ function getFadeValues(tile, parentTile, layer, transform) {
         // if no parent or parent is older, fade in; if parent is younger, fade out
         const fadeIn = !parentTile || Math.abs(parentTile.coord.z - idealZ) > Math.abs(tile.coord.z - idealZ);
 
-        const childOpacity = util.clamp(fadeIn ? sinceTile : 1 - sinceParent, 0, 1);
+        const childOpacity = (fadeIn && tile.refreshedUponExpiration) ? 1 : util.clamp(fadeIn ? sinceTile : 1 - sinceParent, 0, 1);
```
anandthakker: Is the reason it's okay to snap to 1 in the first branch here because, if we're fading _in_ and this is a refreshed tile, then (a) sinceTile will no longer reflect the time since "this" tile was first loaded, and (b) we're probably done crossfading from parent to child anyway?

This crossfading logic feels scary enough to edit that it might be nice to stick a comment in to this effect (or with the correct explanation if my guess is wrong)

src/ui/map.js
lbud: 👍

src/render/draw_raster.js
lbud: @anandthakker good question. TBH I can't tell you exactly how the tile fading logic works, _but_ you'd only want to bypass fading logic if a tile is fading in (because it fades in newly loaded tiles). That…may not be a good enough explanation. 🤔

lbud:
@anandthakker I updated the draw_raster part so that we unset the `refreshedUponExpiration` flag once a tile has crossed the time threshold for crossfading — feels a lot safer 👍

PR #4379, opened 2017-03-07
Don't clamp crossfade in the event of fadeout

lbud:
@anandthakker caught this case yesterday — if a tile is fading out and was just refreshed (i.e. zoom maneuver _right_ after refresh that would cause it to be fading out within 300ms) we won't want to clamp to 1.

### DIFF
--- a/src/render/draw_raster.js
+++ b/src/render/draw_raster.js
@@ -122,7 +122,7 @@ function getFadeValues(tile, parentTile, layer, transform) {
         // if no parent or parent is older, fade in; if parent is younger, fade out
         const fadeIn = !parentTile || Math.abs(parentTile.coord.z - idealZ) > Math.abs(tile.coord.z - idealZ);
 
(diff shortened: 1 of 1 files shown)


src/render/draw_raster.js
```diff
@@ -122,7 +122,7 @@ function getFadeValues(tile, parentTile, layer, transform) {
         // if no parent or parent is older, fade in; if parent is younger, fade out
         const fadeIn = !parentTile || Math.abs(parentTile.coord.z - idealZ) > Math.abs(tile.coord.z - idealZ);
 
-        const childOpacity = tile.refreshedUponExpiration ? 1 : util.clamp(fadeIn ? sinceTile : 1 - sinceParent, 0, 1);
```
anandthakker: @lbud I was gonna say we should switch this from `!!fadeIn` to `Boolean(fadeIn)`, but realizing now that `fadeIn` is already safely a boolean, based on the line above, so we can just change `!!fadeIn` to `fadeIn`.

Approving, nitpick notwithstanding.

PR #4390, opened 2017-03-09
Ensure DragRotate stops when the window looses focus. Fixes #3389

### LINKED ISSUE #3389: Map remains in change camera mode (control key) after changing windows
**mapbox-gl-js version**: `v0.26.0`
- mac
- chrome 53.0.2785.143
### Steps to Trigger Behavior
1. Use the control key change the camera on the map
2. Quickly change windows with command + tab
3. Go back to chrome
4. Map remains in change camera mode even though the control key has been released
### Expected Behavior

The map would not be remain in change camera/pitch mode since the control key is not pressed.
### Actual Behavior

The map remains in change camera/pitch mode even though the control key is not pressed.

andrewharvey:
## Launch Checklist

The depends on #4389. Without it you might not observe this PR fixing the issue.

Ensure DragRotate stops when the window looses focus. Fixes #3389

I couldn't find any existing unit tests for DragRotate.

 - [ ] post benchmark scores

Manually tested on Linux with Chrome and Firefox (with #4389 applied).

### DIFF
--- a/src/ui/handler/drag_rotate.js
+++ b/src/ui/handler/drag_rotate.js
@@ -82,7 +82,7 @@ class DragRotateHandler {
 
         window.document.addEventListener('mousemove', this._onMove);
         window.document.addEventListener('mouseup', this._onUp);
-        /* Deactivate DragRotate when the window looses focus. Otherwise if a mouseup occurs when the window isn't in focus DragRotate will still be active even though the mouse is no longer pressed. */
+        /* Deactivate DragRotate when the window looses focus. Otherwise if a mouseup occurs when the window isn't in focus, DragRotate will still be active even though the mouse is no longer pressed. */
         window.addEventListener('blur', this._onUp);
 
         this._active = false;
--- a/src/ui/handler/drag_rotate.js
+++ b/src/ui/handler/drag_rotate.js
@@ -29,8 +29,7 @@ class DragRotateHandler {
         util.bindAll([
             '_onDown',
             '_onMove',
-            '_onUp',
-            '_onBlur'
+            '_onUp'
         ], this);
 
     }
@@ -83,7 +82,8 @@ class DragRotateHandler {
 
         window.document.addEventListener('mousemove', this._onMove);
         window.document.addEventListener('mouseup', this._onUp);
-        window.addEventListener('blur', this._onBlur);
+        /* Deactivate DragRotate when the window looses focus. Otherwise if a mouseup occurs when the window isn't in focus DragRotate will still be active even though the mouse is no longer pressed. */
+        window.addEventListener('blur', this._onUp);
 
         this._active = false;
         this._inertia = [[Date.now(), this._map.getBearing()]];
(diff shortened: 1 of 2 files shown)


src/ui/handler/drag_rotate.js
```diff
@@ -91,6 +93,14 @@ class DragRotateHandler {
         e.preventDefault();
     }
 
+    /**
+     * Deactivates DragRotate when the window looses focus. Otherwise the mouseup event is never fired and when the
+     * window comes back in focus DragRotate is still active even though the mouse isn't down.
+     */
+    _onBlur(e) {
+        this._onUp(e);
+    }
```
mourner: Since there's nothing else going on, can we just pass `this._onUp` instead of adding an identical method?

andrewharvey: Thanks for the tip. I've changed it. Note that you'll need to apply #4389 before this one.

andrewharvey:
This is incomplete, I need to apply the same logic to the `DragPanHandler`!

andrewharvey:
> This is incomplete, I need to apply the same logic to the DragPanHandler!

I've updated it to apply the same logic to DragPan to fix the same behaviour when dragging.

- [ ] BoxZoom has the same issue, but I think it's less of a problem as it leaves you in a state where you're finishing off the box zoom you started before changing tabs and it's clear that you need to click to finish it. However I'm not fussed either way if it remains this way or we apply the same logic to cancel the box zoom when the window looses focus. Thoughts?

- [ ] Same for TouchZoomRotate however I'm not sure if it would be an issue there as loosing focus while you're still interacting with the map is less common. If it does happen it might mean that if you went to DragPan it would still be stuck in TouchZoomRotate. Do you think I should just go ahead and apply the same logic to de-activate on window blur anyway? I don't have a desktop touch device to more easily test some of these scenarios though.

lucaswoj:
> BoxZoom has the same issue, but I think it's less of a problem as it leaves you in a state where you're finishing off the box zoom you started before changing tabs and it's clear that you need to click to finish it. However I'm not fussed either way if it remains this way or we apply the same logic to cancel the box zoom when the window looses focus. Thoughts?

> Same for TouchZoomRotate however I'm not sure if it would be an issue there as loosing focus while you're still interacting with the map is less common. If it does happen it might mean that if you went to DragPan it would still be stuck in TouchZoomRotate. Do you think I should just go ahead and apply the same logic to de-activate on window blur anyway? I don't have a desktop touch device to more easily test some of these scenarios though.

I see no harm in following up on these edge cases if/when they are reported. What's here is a huge improvement over what we have now! 🚢

lucaswoj:
I'm not able to repro the bug in master right now. Can you confirm that the bug is still present in `master` and verify that the repro steps in the bug report https://github.com/mapbox/mapbox-gl-js/pull/4389 still apply?

> 1. Use the control key change the camera on the map
> 2. Quickly change windows with command + tab
> 3. Go back to chrome
> 4. Map remains in change camera mode even though the control key has been released

andrewharvey:
> I see no harm in following up on these edge cases if/when they are reported. What's here is a huge improvement over what we have now!

Agreed.

> I'm not able to repro the bug in master right now. Can you confirm that the bug is still present in master and verify that the repro steps in the bug report #4389 still apply?

@lucaswoj I presume you mean #3389 (changing windows). I can't test that as it was reported on OSX. I presumed it was the same as changing tabs in Chrome which I can test, and this PR addresses, but maybe not.

lucaswoj:
I can sometimes repro changing tabs 👍 LGTM!

PR #4391, opened 2017-03-09
feature: export the correct js file in npm

### LINKED ISSUE #3551: Make the distribution build the main NPM entry point
Currently [our readme](https://github.com/mapbox/mapbox-gl-js#using-mapbox-gl-js-with-browserify) suggests `require('mapbox-gl')` for Browserify-bundled projects and `require('mapbox-gl/dist/mapbox-gl.js')` for all the rest. 

But is there a real need to support building Mapbox GL JS from within a Browserify-bundled project? Would it be better overall to just change the main entry point in the `package.json` to `dist/mapbox-gl-dev.js` and make the workflow the same (`require('mapbox-gl')`) for all module systems?

This would also enable us to safely use a Yarn lockfile in the repo (#3364 ) for producing the same build and not worry about any downstream modules, and also move all `dependencies` into `devDependencies` for a lighter downstream install.

The only cons I can think of:

- Possible code duplication in downstream Browserify-bundled modules that also use some `mapbox-gl` dependencies like `geojson-vt` or `earcut`. Very rare and not a big deal.
- Not being able to require a specific branch/hash of `mapbox-gl` in a Browserify-bundled downstream module. Not a big deal as well.

Thoughts @jfirebaugh @lucaswoj @tmcw?

BernhardRode:
When installing mapbox-gl via npm and importing the library like:

import * as mapboxgl from 'mapbox-gl';

The npm resolver is looking for tha main attribute in package.json to further resolve the correct JavaScript file for import.

fixes #3551


lucaswoj:
Our docs suggest using `import mapboxgl from 'mapbox-gl/dist/mapbox-gl';` to import this module into Webpack.

That said, I’m leaning towards :+1: on this. It’d make `require(‘mapbox-gl’)` work in more cases. The only downside I can think of is if someone wants to `npm link` and see their changes instantly. That ought to be a pretty rare use case for GL JS since the monorepo. 

Any thoughts @jfirebaugh? 

@BernhardRode You'll need to fix the CI build & ensure npm publish still works.

jfirebaugh:
@mourner suggested this in #3551 and I'm in favor. Let's do it.

lucaswoj:
CI fix:

```diff
diff --git a/package.json b/package.json
index c5b97fc..88338ee 100644
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "build": "npm run build-docs # invoked by publisher when publishing docs on the mb-pages branch",
     "start-docs": "npm run build-min && npm run build-docs && jekyll serve --watch",
     "lint": "eslint --ignore-path .gitignore src test bench docs/_posts/examples/*.html debug/*.html",
-    "lint-docs": "documentation lint",
+    "lint-docs": "documentation lint src/index.js",
     "open-changed-examples": "git diff --name-only mb-pages HEAD -- docs/_posts/examples/*.html | awk '{print \"http://127.0.0.1:4000/mapbox-gl-js/example/\" substr($0,33,length($0)-37)}' | xargs open",
     "test": "run-s lint test-unit test-flow",
     "test-suite": "run-s test-render test-query",
```

PR #4392, opened 2017-03-09
Add regression test for https://github.com/mapbox/mapbox-gl-native/issues/8303

jfirebaugh:
Passing on both js and native. 👌


jfirebaugh:
Let's discuss that separately. It's something we would want to document in the README and likely apply to other tests.

PR #4394, opened 2017-03-09
Restore ESLint checks

### LINKED ISSUE #4387: Restore ESLint check
While reviewing #4386, I suddenly realized that ESLint isn't checking our JS code since `js` folder was renamed to `src`, but the `lint` NPM script still references `js`. We should fix the command and any issues that may have slipped. @lucaswoj @mollymerp

mourner:
Closes #4387.


mourner:
Ah bummer.

PR #4404, opened 2017-03-10
Add "Map#addImage" / "Map#removeImage"

### LINKED ISSUE #2059: Add Map#{add,remove}Image

lucaswoj:
This PR adds the `Map#addImage` and `Map#removeImage` methods which give users the ability to modify their sprite at runtime.

Fixes #2059 

## Launch Checklist

 - [x] upgrade to latest `shelf-pack` version
- [x] make `SpriteAtlas` fire `error` events
 - [x] make `SpriteAtlas` fire `data` events 
 - [x] add `Map#removeImage`
- [x] convince [existing integration tests to pass](https://github.com/mapbox/mapbox-gl-js/blob/878008895104b67861ce73f65809f7f6f0ed7263/test/integration/render-tests/runtime-styling/image-add/style.json)
 - [x] add integration test for `Map#removeImage`
 - [x] add integration test which uses `Map#addImage` to create a repeating pattern
 - [x] document `Map#addImage` and `Map#removeImage`
  - [x] post benchmark scores
- [ ] create a rad example / demo


lucaswoj:
benchmark | master 787b3a2 | addimage e069dd0
--- | --- | ---
**map-load** | 115 ms  | 120 ms 
**style-load** | 101 ms  | 99 ms 
**buffer** | 1,011 ms  | 1,000 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 4.2 ms, 0% > 16ms  | 4.1 ms, 0% > 16ms 
**query-point** | 1.21 ms  | 0.88 ms 
**query-box** | 66.14 ms  | 63.76 ms 
**geojson-setdata-small** | 7 ms  | 3 ms 
**geojson-setdata-large** | 89 ms  | 162 ms

src/symbol/sprite_atlas.js
```diff
@@ -14,9 +16,11 @@ class AtlasImage {
     }
 }
 
-class SpriteAtlas {
+class SpriteAtlas extends Evented {
```
anandthakker: This isn't a comment about the changes in this PR, but just noting while I'm here: having never worked in this code before, I found it confusing, at first, to make sense of how/why `SpriteAtlas` works. E.g., "what's with the `this.copy()`?", "what's a `rect`?", "why does `AtlasImage` not have a reference to the underlying data?"

Reading more widely through the codebase, I think I've satisfied my questions, but this might be a nice time to add a brief, high-level description of what `SpriteAtlas` is/does.  OTOH ticketing that separately is also totally fair 😄

src/ui/map.js
```diff
@@ -818,6 +818,30 @@ class Map extends Camera {
...
+     * Add an image to the style. This image can be used in `icon-image`,
+     * `background-pattern`, `fill-patern`, and `line-pattern`.
+     *
+     * @param {string} name The name of the image.
+     * @param {HTMLImageElement|ArrayBufferView} image The image as an `HTMLImageElement` or `ArrayBufferView` (using the format of [`ImageData#data`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/data))
+     * @param {Object} [options] Only required if passing an `ArrayBufferView`
+     * @param {number} options.width The pixel width of the `ArrayBufferView` image
+     * @param {number} options.height The pixel height of the `ArrayBufferView` image
+     * @param {Object} options.pixelRatio The ratio of pixels in the `ArrayBufferView` image to physical pixels on the screen
+     */
```
anandthakker: Since this could [result in an error](https://github.com/mapbox/mapbox-gl-js/pull/4404/files#diff-69b64c0a443828ff53dfe302a3dffb5bR84) when there isn't enough space for a new image, should we document that possibility?

src/symbol/sprite_atlas.js
lucaswoj: All fair points! 

I lost about a week trying to refactor, simplify, and document our whole style image subsystem and eventually decided that the minimal patch to add this feature was worth executing. 

I'll push another commit with some targeted renames & comments to better explain `SpriteAtlas` until we get around to a bigger cleanup.

src/ui/map.js
```diff
@@ -818,6 +818,30 @@ class Map extends Camera {
     }
 
     /**
+     * Add an image to the style. This image can be used in `icon-image`,
+     * `background-pattern`, `fill-patern`, and `line-pattern`.
```
jfirebaugh: `fill-patern` ⇢ `fill-pattern`

src/ui/map.js
```diff
@@ -818,6 +818,30 @@ class Map extends Camera {
...
+     *
+     * @param {string} name The name of the image.
+     * @param {HTMLImageElement|ArrayBufferView} image The image as an `HTMLImageElement` or `ArrayBufferView` (using the format of [`ImageData#data`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/data))
+     * @param {Object} [options] Only required if passing an `ArrayBufferView`
+     * @param {number} options.width The pixel width of the `ArrayBufferView` image
+     * @param {number} options.height The pixel height of the `ArrayBufferView` image
+     * @param {Object} options.pixelRatio The ratio of pixels in the `ArrayBufferView` image to physical pixels on the screen
```
jfirebaugh: `{Object}` ⇢ `{number}`. Also, should this be marked as optional? It looks like the implementation will default to screen pixelRatio if it's not provided.

src/symbol/sprite_atlas.js
```diff
@@ -50,6 +54,68 @@ class SpriteAtlas {
...
+
+        this.fire('data', {dataType: 'style'});
+    }
+
+    removeImage(name) {
+        const image = this.images[name];
+        delete this.images[name];
+
+        if (!image) {
+            this.fire('error', {error: new Error('No image with this name exists.')});
```
lucaswoj: Yes! 😰

src/symbol/sprite_atlas.js
```diff
@@ -50,6 +54,68 @@ class SpriteAtlas {
...
+            this.fire('error', {error: new Error('Image provided in an invalid format. Supported formats are HTMLImageElement, ImageData, and ArrayBufferView.')});
+        }
+
+        if (this.images[name]) {
+            this.fire('error', {error: new Error('An image with this name already exists.')});
+        }
+
+        const rect = this.allocateImage(width, height);
+        if (!rect) {
+            this.fire('error', {error: new Error('There is not enough space to add this image.')});
```
lucaswoj: Yes! 😰

src/ui/map.js
lucaswoj: 👍

src/ui/map.js
lucaswoj: 👍

src/symbol/sprite_atlas.js
```diff
@@ -50,6 +54,68 @@ class SpriteAtlas {
...
+        const rect = image.rect;
+        const pixelRatio = image.pixelRatio;
+
+        this.copy(
+            new Uint32Array(image.width * image.height * pixelRatio * pixelRatio),
+            image.width * pixelRatio,
+            rect,
+            {pixelRatio: pixelRatio, x: 0, y: 0, width: image.width * pixelRatio, height: image.width * pixelRatio},
+            false
+        );
```
lucaswoj: I added this code out of fear of edge cases with semitransparent images & linear interpolation. Testing suggests these concerns are unwarranted. Removing it for now.

lucaswoj:
I think this is ready to go. How does it look @anandthakker?

src/symbol/sprite_atlas.js
```diff
@@ -3,24 +3,20 @@
...
+class SpriteAtlas extends Evented {
 
     constructor(width, height) {
+        super();
+
         this.width = width;
         this.height = height;
 
-        this.atlas = new ShelfPack(width, height);
+        this.shelfPack = new ShelfPack(width, height);
```
mourner: I'm not sure about this rename. `ShelfPack` is more of an algorithm name than a name that describes an instance. `atlas` is better IMO, and consistent with the class name (`SpriteAtlas`) since the latter is a wrapper around a generic atlas for image sprites.

anandthakker:
LGTM @lucaswoj

src/symbol/sprite_atlas.js
lucaswoj: I see your point and agree that the semantics of the name are questionable. My intent was to reduce the size of the vocabulary and amount of indirection in this class.

 - `SpriteAtlas` has a `shelfPack`
 - `SpriteAtlas#shelfPack` is an instance of `ShelfPack`
 - `SpriteAtlas#shelfPack` has `bins`

-----

 - `SpriteAtlas` has a `atlas`
 - `SpriteAtlas#atlas` is an instance of `ShelfPack`
 - `SpriteAtlas#atlas` has `bins`

lucaswoj: > a wrapper around a generic atlas for image sprites.

If `SpriteAtlas#atlas` were a generic texture atlas, I'd be keen on calling it such. As it stands, this member only provides the "shelf pack" algorithm and not the rest of the texture atlas functionality.

lucaswoj:
Merging and working on an example / demo as a follow-up 😄

pwilczynski:
@lucaswoj - thank you so much, this looks awesome, and is so exciting for us. No more flickering maps for us!

ChrisLoer:
I'm way late to this party, but how do we handle stale `icons` in `SymbolBucket` after a call to `Map#removeImage`? It seems like if you remove an image that's currently displayed on a tile, add another one (thus overwriting the old one's position in the `SpriteAtlas`), and then call `redoPlacement`, you'll get symbols referring to a garbage location in the `SpriteAtlas`. It looks like the expected behavior is if you call `Map#removeImage` while there are tiles displaying the image, the image should just disappear from the tiles the next time layout happens.

PR #4418, opened 2017-03-14
FIX hasInvalidDimension call with 1 argument where 0 are expected

lkrombeen:
**Problem**:
The function _hasInvalidDimensions takes 0 parameters but is called with 1 in function load.
Can cause an issue if the method ever changes (also see #4408).

**Problem Location**:
*src/source/canvas_source.js*
Class: *CanvasSource*
Function: *load*

**Fix**:
Removed the unwanted, unused parameter from the call.

**Test**:
All tests succeed. However, the load function which made the erroneous call has not been tested.
I do not know how to properly stub this (if possible). 
Advised to have a different manual check to rule out "Works on my machine" :)


lucaswoj:
Thank you for catching this & clearly explaining your work @lkrombeen 😄

PR #4423, opened 2017-03-14
Export the "dist" file in npm

### LINKED ISSUE #3551: Make the distribution build the main NPM entry point

lucaswoj:
Brought into this repository from https://github.com/mapbox/mapbox-gl-js/pull/4391

> When installing mapbox-gl via npm and importing the library like:
>
> import * as mapboxgl from 'mapbox-gl';
> 
> The npm resolver is looking for tha main attribute in package.json to further resolve the correct JavaScript file for import.
>
> fixes #3551

### DIFF
--- a/docs/_includes/quickstart-bundler.html
+++ b/docs/_includes/quickstart-bundler.html
@@ -11,8 +11,8 @@ npm install --save mapbox-gl
 <p>Include the following code in the <code>&lt;body&gt;</code> of your HTML file.</p>
 <div class='js-replace-token'>
 {% copyable %}{% highlight javascript %}
-import mapboxgl from 'mapbox-gl/dist/mapbox-gl';
-// or "const mapboxgl = require('mapbox-gl/dist/mapbox-gl');"
+import mapboxgl from 'mapbox-gl';
+// or "const mapboxgl = require('mapbox-gl');"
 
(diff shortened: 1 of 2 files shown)


lucaswoj:
cc @BernhardRode 😄

PR #4430, opened 2017-03-15
Use highp for shader variable "gamma" to avoid text aliasing

anandthakker:
Cherry picked from `release-ios-v3.5.0-android-v5.0.0` commit 28e9725bd5d18cc623435292a408df6fbefa1da9


anandthakker:
That was a poorly chosen title / commit message on the release branch; retitled it here

Issue #1064, opened 2015-03-06
Use anisotropic filtering for raster tiles in perspective view


ansis:
We should use the `EXT_texture_filter_anisotropic` extension if it is available. Anisotropic filtering will make [textures clearer when viewed at an angle](http://media.tojicode.com/webgl-samples/anisotropy.html). The filtering level used should depend on the current pitch.

lucaswoj:
@nickidlugash I wonder if this could help us render pitched labels more clearly 🤔

nickidlugash:
@lucaswoj this looks interesting. Do we use mipmaps for labels? I don't really know how our textures work.

andrewharvey:
This certainly makes pitched raster maps look much nicer!

![selection_490](https://cloud.githubusercontent.com/assets/117278/24040809/d753ec5e-0b5e-11e7-8687-3c1d9c4079c6.png)
![selection_489](https://cloud.githubusercontent.com/assets/117278/24040811/d9987926-0b5e-11e7-840e-3482f0f36a7f.png)

Working on this in #4443

mollymerp:
🚢 in #4443 - thank you @andrewharvey !

PR #4443, opened 2017-03-17
use anisotropic filtering for raster tiles #1064

andrewharvey:
## Launch Checklist

Implements anisotropic filtering for raster tiles #1064
Rendering test fixtures may need to be updated, apart from that I don't think it's something that we can/need unit tests for.
N/A
 - [x] post benchmark scores
I ran a modified fps benchmark with satellite-streets at pitch 60 and bearing 45. Got around the same 60FPS on both.

Before/After
selection_489
selection_490

- [ ] test it falls back on unsupported devices
Not sure how to do this.
- [x] investigate if it makes any difference that this is enabled even when the map isn't pitched
My fps benchmark didn't seem to indicate any performance issues, so I think it's fine as is.
- [ ] "The filtering level used should depend on the current pitch."
Do we need to do this, or can we just use the max level?

### DIFF
--- a/src/source/raster_tile_source.js
+++ b/src/source/raster_tile_source.js
@@ -90,6 +90,11 @@ class RasterTileSource extends Evented {
                 gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
                 gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
                 gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+
+                if (this.map.painter.extTextureFilterAnisotropic) {
+                    gl.texParameterf(gl.TEXTURE_2D, this.map.painter.extTextureFilterAnisotropic.TEXTURE_MAX_ANISOTROPY_EXT, this.map.painter.extTextureFilterAnisotropicMax);
+                }
+
                 gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, img);
                 tile.texture.size = img.width;
             }
--- a/src/render/painter.js
+++ b/src/render/painter.js
@@ -113,6 +113,15 @@ class Painter {
         rasterBoundsArray.emplaceBack(EXTENT, EXTENT, 32767, 32767);
         this.rasterBoundsBuffer = Buffer.fromStructArray(rasterBoundsArray, Buffer.BufferType.VERTEX);
         this.rasterBoundsVAO = new VertexArrayObject();
+
+        this.extTextureFilterAnisotropic = (
+            gl.getExtension('EXT_texture_filter_anisotropic') ||
+            gl.getExtension('MOZ_EXT_texture_filter_anisotropic') ||
+            gl.getExtension('WEBKIT_EXT_texture_filter_anisotropic')
+        );
+        if (this.extTextureFilterAnisotropic) {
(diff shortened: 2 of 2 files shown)


src/source/raster_tile_source.js
```diff
@@ -90,6 +90,17 @@ class RasterTileSource extends Evented {
...
+                const ext = (
+                    gl.getExtension('EXT_texture_filter_anisotropic') ||
+                    gl.getExtension('MOZ_EXT_texture_filter_anisotropic') ||
+                    gl.getExtension('WEBKIT_EXT_texture_filter_anisotropic')
+                );
+                if (ext) {
+                    const max = gl.getParameter(ext.MAX_TEXTURE_MAX_ANISOTROPY_EXT);
+                    gl.texParameterf(gl.TEXTURE_2D, ext.TEXTURE_MAX_ANISOTROPY_EXT, max);
+                }
+
```
lucaswoj: Are any of these values cachable for the lifetime of the page? Reducing the # of per-frame calls to WebGL APIs is important for performance. You may want to wrap that caching behaviour in the `painter` module.

andrewharvey: Yeah, good point!

andrewharvey: Done.

mourner:
Awesome! Do you know if it has a noticeable impact on performance? E.g. should we just have it enabled all the time, or add an option to turn it off?

src/source/raster_tile_source.js
lucaswoj: 👌

andrewharvey:
> Awesome! Do you know if it has a noticeable impact on performance? E.g. should we just have it enabled all the time, or add an option to turn it off?

I ran a modified fps benchmark with satellite-streets at pitch 60 and bearing 45. Got around the same 60FPS on both.

I also did manual testing, rotating and pitching the map with the mouse and couldn't notice any performance degradation. So I think it's simpler to leave it on.

mourner:
@andrewharvey fps benchmark is not very useful once it reaches 60fps — I think we should remove it. You might get more out of the `frame-duration` benchmark.

andrewharvey:
`streets` style no impact as expected

benchmark | master c189bbf | 1064-anisotropic-filtering 8f139a8
--- | --- | ---
**frame-duration** | 6.1 ms, 0% > 16ms  | 6.2 ms, 0% > 16ms 

---

`satellite` style with bearing: 0, pitch: 0 (default top down)

benchmark | master c189bbf | 1064-anisotropic-filtering 8f139a8
--- | --- | ---
**frame-duration** | 6.1 ms, 0% > 16ms  | 1.3 ms, 0% > 16ms 

---
`satellite-streets` style with bearing: 0, pitch: 0 (default top down)

benchmark | master c189bbf | 1064-anisotropic-filtering 8f139a8
--- | --- | ---
**frame-duration** | 6.1 ms, 0% > 16ms  | 5.1 ms, 0% > 16ms 


---

`satellite` style with bearing: 45, pitch: 60 (most extreme view)

benchmark | master c189bbf | 1064-anisotropic-filtering 8f139a8
--- | --- | ---
**frame-duration** | 6.2 ms, 0% > 16ms  | 3.1 ms, 0% > 16ms 

---

`satellite-streets` style with bearing: 45, pitch: 60 (most extreme view)

benchmark | master c189bbf | 1064-anisotropic-filtering 8f139a8
--- | --- | ---
**frame-duration** | 5.9 ms, 0% > 16ms  | 9.9 ms, 5% > 16ms

mourner:
Huh, so it's actually faster with anisotropic filtering most of the time? 🎉

andrewharvey:
> Huh, so it's actually faster with anisotropic filtering most of the time?

Yeah, unless there is an issue with the benchmark. I ran a second time to be sure and results were the same.

PR #4446, opened 2017-03-17
Prevent 404 HTTP code from throwing an exception

### LINKED ISSUE #1800: Suppress error messages for missing tiles

lucaswoj:
This PR prevents throwing an exception when a tile request yields a 404 HTTP status code. It also minorly improves our test coverage of the `ajax.js` util module. 

Fixes #1800

## Launch Checklist

### DIFF
--- a/debug/index.html
+++ b/debug/index.html
@@ -18,29 +18,12 @@
 <script src='/debug/access_token_generated.js'></script>
 <script>
 
-var tileset = 'lucaswoj.57aw96oc';
-var map = new mapboxgl.Map({
-    container: 'map', // container id
-    style: {
-        "version": 8,
-        "sources": {
-            "raster-tiles": {
-                "type": "raster",
-                "url": "mapbox://" + tileset,
-                "tileSize": 256
-            }
-        },
-        "layers": [{
-            "id": "simple-tiles",
-            "type": "raster",
-            "source": "raster-tiles",
-            "minzoom": 0,
-            "maxzoom": 22
-        }]
-    },
-    center: [-120.8, 39.1], // starting position
-    zoom: 11 // starting zoom
-
+var map = window.map = new mapboxgl.Map({
+    container: 'map',
+    zoom: 12.5,
+    center: [-77.01866, 38.888],
+    style: 'mapbox://styles/mapbox/streets-v10',
+    hash: true
 });
 
 </script>
--- a/debug/index.html
+++ b/debug/index.html
@@ -18,12 +18,29 @@
 <script src='/debug/access_token_generated.js'></script>
 <script>
 
-var map = window.map = new mapboxgl.Map({
-    container: 'map',
-    zoom: 12.5,
-    center: [-77.01866, 38.888],
-    style: 'mapbox://styles/mapbox/streets-v10',
-    hash: true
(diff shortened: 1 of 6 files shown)


debug/index.html
```diff
@@ -18,12 +18,29 @@
...
+            "id": "simple-tiles",
+            "type": "raster",
+            "source": "raster-tiles",
+            "minzoom": 0,
+            "maxzoom": 22
+        }]
+    },
+    center: [-120.8, 39.1], // starting position
+    zoom: 11 // starting zoom
+
```
mollymerp: do you want to add this permanently? I'd prefer to keep debug page simple but open to arguments 😊

src/source/source_cache.js
```diff
@@ -165,7 +165,7 @@ class SourceCache extends Evented {
     _tileLoaded(tile, id, previousState, err) {
         if (err) {
             tile.state = 'errored';
-            this._source.fire('error', {tile: tile, error: err});
+            if (err.status !== 404) this._source.fire('error', {tile: tile, error: err});
```
mollymerp: are there any other cases in which a tile error would 404 for a legitimate reason besides no data in the tile (and thus no tile) ? Should we check the message matches 'Error: not found' ?

src/source/vector_tile_source.js
```diff
@@ -87,7 +87,8 @@ class VectorTileSource extends Evented {
                 return;
 
             if (err) {
-                return callback(err);
+                if (err.status !== 404) callback(err);
```
mollymerp: but do we still need to call the callback (w/o `err`) if the tile is not found?

debug/index.html
lucaswoj: Oopsie. I meant to overwrite this before pushing.

src/source/vector_tile_source.js
lucaswoj: Yup. The latest version of this file didn't get pushed either. Fixed now.

src/source/source_cache.js
lucaswoj: The 404 status code strongly implies the "Not found" status text.

 > [At the HTTP level, a 404 response code is followed by a human-readable "reason phrase". The HTTP specification suggests the phrase "Not Found" and many web servers by default issue an HTML page that includes both the 404 code and the "Not Found" phrase.](https://en.wikipedia.org/wiki/HTTP_404)

I cannot imagine any cases in which a custom status message on a 404 response would necessitate different error handling. If we discover one in the future, we can accommodate it  then. 😄

src/util/ajax.js
```diff
@@ -2,6 +2,13 @@
 
 const window = require('./window');
 
+class AJAXError extends Error {
+    constructor(message, status) {
+        super(message);
+        this.status = status;
+    }
+}
```
anandthakker: 👍

JuanIrache:
Hi. I think there's a problem with completely preventing the expception. If you are waiting for the "idle" event and the 404 error prints on the browser, the idle event does not fire, so my program keeps hanging. Having the exception would allow my program to move on.

Is there a workaround to this?

PR #4451, opened 2017-03-20
jump back to world zero when crossing the antimeridian in flyTo and other camera operations. Fixes #2071

### LINKED ISSUE #2071: By default, snap back to "w=0" world after panning operation (i.e. "worldCopyJump")

andrewharvey:
## Launch Checklist

Some background in #2071. When `flyTo` decides to cross the anti-meridian from world 0, it will end up in either world -1 or 1. Because `Marker` DOM elements are only drawn in one place and only on world 0, they don't show up at the moment.

This PR aims to ensure that when flyTo is about to step across the anti-meridian into world -1 or 1, it teleports to ensure it lands back in world 0, and so `Marker` elements appear.

It also ensure's that other camera operations like `jumpTo`, `easeTo` don't leave the map in a non world 0 state.

I feel this should be transparent to the public API, it's a private behaviour.

 - [ ] post benchmark scores

I wrote a new debug page to test this, but haven't committed it.

- [ ] the world sometimes flashes white when you cross the anti-meridian, I'm not sure how to fix this.

- [x] only jump back to world 0 when renderWorldCopies is true

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -230,7 +230,10 @@ class Transform {
 
     setLocationAtPoint(lnglat, point) {
         const translate = this.pointCoordinate(point)._sub(this.pointCoordinate(this.centerPoint));
-        this.center = this.coordinateLocation(this.locationCoordinate(lnglat)._sub(translate)).wrap();
+        this.center = this.coordinateLocation(this.locationCoordinate(lnglat)._sub(translate));
+        if (this._renderWorldCopies) {
+            this.center = this.center.wrap();
+        }
     }
 
     /**
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -230,7 +230,7 @@ class Transform {
 
     setLocationAtPoint(lnglat, point) {
         const translate = this.pointCoordinate(point)._sub(this.pointCoordinate(this.centerPoint));
-        this.center = this.coordinateLocation(this.locationCoordinate(lnglat)._sub(translate));
+        this.center = this.coordinateLocation(this.locationCoordinate(lnglat)._sub(translate)).wrap();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


andrewharvey:
Sorry for the number of PR's here, after deciding to tackle the wider issue of #2071 per https://github.com/mapbox/mapbox-gl-js/issues/2071#issuecomment-287675896. This might be better applied higher up in geo/transform.

src/geo/lng_lat.js
```diff
@@ -42,7 +42,11 @@ class LngLat {
      * wrapped.lng; // = -73.9749
      */
     wrap() {
-        return new LngLat(wrap(this.lng, -180, 180), this.lat);
+        if (this.lng <= -180 || this.lng > 180) {
```
jfirebaugh: Any particular reason to change this behavior? We'd want to at least update the API documentation above, which says that `wrap` always returns a new object.

andrewharvey: Since my changes meant that LngLat.wrap is called a lot more (every camera change due to an interaction), I wanted to ensure that creating the new LngLat object wouldn't impact performance.

Although on further testing, map interaction seems to have no degradation without this check. So I'm not sure whether to leave it in and update the API docs or just remove this change, and leave it as is.

Thoughts?

jfirebaugh: If there's no performance impact, let's leave it as is.

andrewharvey: Agreed, fixed, and squashed commits.

andrewharvey:
> I didn't observe this in my testing. If you're still seeing it, can you open an issue with steps to replicate?

I only noticed it occasionally, what happens is the map is there and after the jump it redraws part of the map (a vector tile) so the map flashes the background colour while this happens.

I agree that we should merge this and follow up with a new issue to address that.

PR #4452, opened 2017-03-20
move markers to the best world for the current view

andrewharvey:
## Launch Checklist

DOM Elements like Markers and Popups are only drawn in world 0. However when you're close to the anti-meridian, DOM Elements which should appear might not, and in some cases it makes sense to draw the DOM elements in world -1 and 1.

This aims to address #3770. However until #2071 is fixed, this PR only addresses issues when the view is in world 0.

public LngLat.wrapToWorld(center) added JS doc.

 - [ ] post benchmark scores
I did some manual testing of this, with both a marker and popup around long 170 and at long -170, and it works well for world -1, 0 and 1, so long as the camera always wraps to world 0 this is enough.

When zoomed out the marker/popup will jump which might be a poor UX, but I think it's okay and probably better than the alternative of it disappearing off the view.

### DIFF
--- a/src/geo/lng_lat.js
+++ b/src/geo/lng_lat.js
@@ -68,6 +68,36 @@ class LngLat {
     toString() {
         return `LngLat(${this.lng}, ${this.lat})`;
     }
+
+    /**
+     * Returns a new `LngLat` object snapped to the best world to draw it provided a map `center` `LngLat`.
+     *
+     * When the map is close to the anti-meridian showing a point on world -1 or 1 is a better
+     * choice. The heuristic used is to minimize the distance from the map center to the point.
+     *
+     * Only works where the `LngLat` is wrapped with `LngLat.wrap()` and `center` is within the main world map.
+     *
+     * @param {LngLat} center Map center within the main world.
+     * @return {LngLat} The `LngLat` object in the best world to draw it for the provided map `center`.
+     * @example
+     * var ll = new mapboxgl.LngLat(170, 0);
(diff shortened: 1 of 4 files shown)


src/util/util.js
```diff
@@ -4,6 +4,7 @@
 const UnitBezier = require('@mapbox/unitbezier');
 const Coordinate = require('../geo/coordinate');
 const Point = require('point-geometry');
+const LngLat = require('../geo/lng_lat');
```
andrewharvey: This creates a circular dependency between util/util.js and geo/lng_lat.js (which uses util.wrap). Which seems to be causing an issue. Investigating...

andrewharvey: I've removed the circular dependency by moving the `bestWorld` function out of util/util.

mourner:
Alternatively, instead of a separate util file, we could add a `LngLat#snapToWorld(center)` method in `lnglat.js`.

lucaswoj:
Tagging @mourner for review on this one as he's the most familiar with this part of the code & similar code in Leaflet.

andrewharvey:
> Alternatively, instead of a separate util file, we could add a LngLat#snapToWorld(center) method in lnglat.js.

@mourner Agreed, it feels much cleaner and better style this way. I've made the change and squashed commits.

andrewharvey:
- [ ] I need to check `renderWorldCopies` and only enable this behaviour when it is true.

andrewharvey:
Testing #4452 #4451 #4459 all working together :tada: 

out

out2

PR #4455, opened 2017-03-20
Enable property functions for {text,icon}-size

### LINKED ISSUE #4228: Enable property functions for "text-size" and "icon-size"

anandthakker:
Closes #4228  

TODO:
- [x] Thoroughly document approach (which will require documenting some existing symbol placement logic as well)
- [x] Enable icon-size
- [x] <del>Fix up collision box debug shader/drawing</del> (Doesn't look like these are affected)
- Render tests for:
    - [x] icon-size/property-function (sdf icon)
    - [x] icon-size/property-function (non-sdf icon)
    - [x] text-size/composite-function (plain)
    - [x] text-size/composite-function (symbol-placement: line)
    - [x] <del>text-size/composite-function (symbol-placement: line, overzoomed tile - target edge case where shader's evaluated function value != layout-time function value)</del> (no longer relevant: sending placement-level size to the shader directly via an attribute)
- [x] Port to native https://github.com/mapbox/mapbox-gl-native/pull/8593
- [x] Fix interpolation value to respect `base` property (see https://github.com/mapbox/mapbox-gl-js/pull/4455#issuecomment-290589065)
- [x] post benchmark scores

benchmark | master a73a85b | dds-text-size 326e42c
--- | --- | ---
**map-load** | 233 ms  | 134 ms 
**style-load** | 117 ms  | 98 ms 
**buffer** | 1,065 ms  | 1,062 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 4.9 ms, 0% > 16ms  | 5 ms, 0% > 16ms 
**query-point** | 1.16 ms  | 1.18 ms 
**query-box** | 68.65 ms  | 70.20 ms 
**geojson-setdata-small** | 5 ms  | 2 ms 
**geojson-setdata-large** | 126 ms  | 166 ms


anandthakker:
Capturing notes on this issue from the TODO above:

> text-size/composite-function (symbol-placement: line, overzoomed tile - target edge case where shader's evaluated function value != layout-time function value)

The concern here is that we might run into a problem due to the shaders having to evaluate composite functions using a limited number of stops, because:
 - We evaluate text-size at _layout_ time (with `{zoom: tile.z + 1}`) using the complete style function; among other things, this determines the valid zoom range for each of the several instances of a glyph produced when `symbol-placement: line`.
 - But then, in the shader, we calculate an adjusted version of those zoom ranges.  To do so, we recover the value of text-size at the layout zoom level by [evaluating the style function in the shader](https://github.com/mapbox/mapbox-gl-js/pull/4455/commits/8f8e4b337c44d5fdd80835609972686a3ad0efff#diff-8d48f9dfab8ef594816590f56a9353b8R64), which means we're using a limited number of stops (4 in gl js, 2 in native)

Cases to consider:
(`u_zoom` is the current (rendered) zoom level; "layout zoom" is `tile.z + 1`)

## Normal case (`tile.z <= u_zoom < tile.z + 1`)
This case should be fine.  If the function has zoom stops above `u_zoom`, `ProgramConfiguration` [always includes them](https://github.com/mapbox/mapbox-gl-js/blob/master/src/data/program_configuration.js#L99-L103), and so the shader should get the "true" value of the style function at `tile.z + 1`.

## Underzoomed tile (`u_zoom < tile.z`)
Test case:
```js
[
  [{zoom: 0, value: 0}, 1],
  [{zoom: 1, value: 0}, 2],
  [{zoom: 2, value: 0}, 4],
  [{zoom: 3, value: 0}, 8],
  [{zoom: 4, value: 0}, 16]
]
```

If we render a z4 vector tile at zoom 1.8, then the shader would only have stop values up to `[{zoom: 3, value: 0}, 8]`.  Thus, interpolating linearly, it would incorrectly calculate that the size used at layout time was 12.  

## Overzoomed tile (`u_zoom > tile.z + 1`)

Similar to underzoomed.

src/shaders/symbol_sdf.vertex.glsl
```diff
@@ -43,8 +50,23 @@ void main() {
...
+    // In order to accommodate placing labels around corners in
+    // symbol-placement: line, each glyph in a label could have multiple layout
+    // "instances", only one of which should be shown at a given zoom level.
+    // The min/max zoom assigned to each instance is based on the font size at
+    // the vector tile's zoom level, which might be different than at the
+    // currently rendered zoom level if text-size is zoom-dependent.
+    // Thus, we compensate for this difference by calculating an adjustment
+    // based on the scale of rendered text size relative to layout text size.
+    mediump float adjustedSize = u_is_size_composite
+        ?  evaluate_zoom_function_1(a_size, u_adjusted_size_t) / 10.0
```
anandthakker: It is unfortunate that this breaks the abstraction provided by the #pragma system, and it's also going to need reworking to be usable in native.

anandthakker:
Okay, here's an draft based on treating text/icon-size as a "pseudo" paint property; the actual DDS part is in the last three commits (the previous ones are miscellaneous cleanup, docs, and a couple bug fixes).

A key difficulty is that, when text/icon-size is a composite function, we need per-vertex values for text-size at both the _rendered_ zoom and _layout_ zoom (the latter being `adjustedSize` described https://github.com/mapbox/mapbox-gl-js/issues/4228#issuecomment-286264725).  In this draft, my approach is to use the four stop-values that are already being provided to the shader for the (pseudo-)paint property, along with an extra interpolation t-value that represents the layout zoom level.  This allows us to only use one vertex attribute for text/icon-size, so that if we also pack together two of the existing layout attributes together we'd remain at 8--the max for native.  However, it's got some problems: https://github.com/mapbox/mapbox-gl-js/pull/4455#issuecomment-288227094 and https://github.com/mapbox/mapbox-gl-js/pull/4455#pullrequestreview-28246490.

I'm going to try an alternate approach next: instead of trying to make a pseudo paint property, I'll try doing it with a new layout attribute -- i.e. the second bullet here https://github.com/mapbox/mapbox-gl-js/issues/4228#issuecomment-286807767

src/render/draw_symbol.js
```diff
@@ -130,7 +129,10 @@ function setSymbolDrawState(program, painter, isText, isSDF, rotateWithMap, pitc
         gl.uniform2f(program.u_texsize, glyphAtlas.width / 4, glyphAtlas.height / 4);
     } else {
         const mapMoving = painter.options.rotating || painter.options.zooming;
-        const iconScaled = size !== 1 || browser.devicePixelRatio !== painter.spriteAtlas.pixelRatio || iconsNeedLinear;
+        const iconSizeScaled = !layer.isLayoutValueFeatureConstant('text-size') ||
+            !layer.isLayoutValueZoomConstant('text-size') ||
+            layer.getLayoutValue('text-size', { zoom: tr.zoom }, {}) !== 1;
```
anandthakker: These should be `icon-size`

anandthakker:
Replaced the problematic approach described in https://github.com/mapbox/mapbox-gl-js/pull/4455#issuecomment-288766100 with using a new layout attribute 👌

src/data/bucket/symbol_bucket.js
```diff
@@ -61,17 +60,18 @@ const symbolInterfaces = {
...
-        ]),
+        ],
         elementArrayType: createElementArrayType(2)
     }
 };
 
-function addVertex(array, x, y, ox, oy, tx, ty, minzoom, maxzoom, labelminzoom, labelangle) {
-    array.emplaceBack(
+function addVertex(array, x, y, ox, oy, tx, ty, sizeData, minzoom, maxzoom, labelminzoom, labelangle) {
+    array.emplaceBack.apply(array,
```
jfirebaugh: I think this is a pretty hot path, so this change should receive some thorough benchmarking.

src/render/draw_symbol.js
```diff
@@ -140,36 +145,63 @@ function setSymbolDrawState(program, painter, isText, isSDF, rotateWithMap, pitc
...
+            tr.pixelsToGLUnits[0] * s,
+            tr.pixelsToGLUnits[1] * s);
     }
 
     if (isSDF) {
         const haloWidthProperty = `${isText ? 'text' : 'icon'}-halo-width`;
         const hasHalo = !layer.isPaintValueFeatureConstant(haloWidthProperty) || layer.paint[haloWidthProperty];
-        const gammaScale = fontScale * (pitchWithMap ? Math.cos(tr._pitch) : 1) * tr.cameraToCenterDistance;
-        gl.uniform1f(program.u_font_scale, fontScale);
+        const gammaScale = (isText ? 1 / 24 : 1) * (pitchWithMap ? Math.cos(tr._pitch) : 1) * tr.cameraToCenterDistance;
```
jfirebaugh: Why does `fontScale` reduce to `1` elsewhere in this function but not here?

src/shaders/encode_attribute.js
```diff
@@ -0,0 +1,18 @@
...
+const util = require('../util/util');
+
+/**
+ * Packs two numbers, interpreted as 8-bit unsigned integers, into a single
+ * float.  Unpack them in the shader using the `unpack_float()` function,
+ * defined in _prelude.vertex.glsl
+ *
+ * @private
+ */
+exports.packUint8ToFloat = function pack(a, b) {
```
jfirebaugh: Is this used somewhere? I can only find it in the tests.

src/style-spec/function/index.js
```diff
@@ -98,21 +98,24 @@ function createFunction(parameters, propertySpec) {
 
         if (zoomAndFeatureDependent) {
             const featureFunctions = {};
-            const featureFunctionStops = [];
+            const zoomStops = [];
```
jfirebaugh: Can you split this into a separate fix/PR, with a test?

src/shaders/symbol_icon.vertex.glsl
```diff
@@ -3,11 +3,20 @@ attribute vec4 a_pos_offset;
 attribute vec2 a_texture_pos;
 attribute vec4 a_data;
 
+// icon-size data (see symbol_sdf.vertex.glsl for more)
+attribute vec3 a_size;
```
jfirebaugh: Are we combining two attributes somewhere else or does this put us over the eight attribute limit?

src/shaders/encode_attribute.js
anandthakker: Whoops. Leftover from previous attempt.

anandthakker: Oh wait, no -- I think we'll need this, it's just still TBD: we still need to pack `a_data` together with `a_texture_pos` so that there's no net increase in number of attributes, and this helper will be used for that.

src/shaders/symbol_icon.vertex.glsl
anandthakker: Yep we still need to combine https://github.com/mapbox/mapbox-gl-js/pull/4455#discussion_r108415748

src/render/draw_symbol.js
anandthakker: Good question.  I think I made this change before I'd introduced `u_is_text` and the [fontScale calculation](https://github.com/mapbox/mapbox-gl-js/pull/4455/files#diff-3b132f79f9e50d40cccc486b7eb73fa4R28) into the shader.  Now that those are in place, the `(isText ? 1 / 24 : 1)` can be removed here in favor of multiplying `u_gamma_scale` by `fontScale` instead of by `v_size`.

src/style-spec/function/index.js
anandthakker: ==> https://github.com/mapbox/mapbox-gl-js/pull/4509

src/data/bucket/symbol_bucket.js
anandthakker: Yeah, buffer benchmark shows a 5% increase

benchmark | master 7c4b9a1 | dds-text-size ae60c28
--- | --- | ---
**buffer** | 906 ms  | 952 ms

anandthakker: Further profiling confirms that `apply` is a bad idea here.

Timing `emplaceBack()` (`[min, 1st quartile, median, 3rd quartile, max]`):

 - master (i.e. w/o the size data added in this PR): `[ 0.001, 0.004, 0.005, 0.006, 0.778 ]`
 - using apply: `[ 0.002, 0.008, 0.01, 0.011, 12.945 ]`

Changed to avoid using apply by always passing three (possibly undefined) args for the size data; timing for emplaceBack with that approach is basically identical to master.

src/shaders/encode_attribute.js
anandthakker: It's now used (see 2d3b36a)

src/data/bucket/symbol_bucket.js
```diff
@@ -261,28 +301,40 @@ class SymbolBucket {
...
+
+        if (
+            layer.isLayoutValueFeatureConstant('icon-size') &&
+            !layer.isLayoutValueZoomConstant('icon-size')
+        ) {
+            this.adjustedIconSize = layer.getLayoutValue('icon-size', {zoom: this.zoom + 1});
+        }
+        if (!layer.isLayoutValueZoomConstant('icon-size')) {
+            this.iconSizeCoveringZoomStops = this.layers[0].getLayoutValueCoveringZoomStops('icon-size', this.zoom, this.zoom + 1);
+        }
```
anandthakker: Hmm.  This is overcomplicated.  We don't really need to serialize `adjusted{Text,Icon}Size`, or `{text,icon}SizeCoveringZoomStops` on the bucket -- drawSymbols() can (re)compute these at render time, since they only depend on the tile's zoom level.

src/data/bucket/symbol_bucket.js
```diff
@@ -61,32 +60,40 @@ const symbolInterfaces = {
...
+        Math.round(ox * 64),
+        Math.round(oy * 64),
+
+        // a_data
+        tx / 4, // x coordinate of symbol on glyph atlas texture
+        ty / 4, // y coordinate of symbol on glyph atlas texture
+        packUint8ToFloat(
             (labelminzoom || 0) * 10, // labelminzoom
-            labelangle, // labelangle
+            labelangle // labelangle
```
ChrisLoer: Packing labelangle into 8 bits is really tight (less than 1 degree precision). Let's try to find some spare bits somewhere else.

src/data/bucket/symbol_bucket.js
```diff
@@ -661,9 +736,25 @@ class SymbolBucket {
...
+     * {text,icon} size -- i.e. text/icon-size at `z: tile.zoom + 1`. If the
+     * size is zoom-dependent, then the zoom range is adjusted at render time
+     * to account for the difference.
+     *
+     * @private
+     */
     addSymbolInstance(anchor, line, shapedTextOrientations, shapedIcon, layer, addToBuffers, collisionBoxArray, featureIndex, sourceLayerIndex, bucketIndex,
-        textBoxScale, textPadding, textAlongLine,
-        iconBoxScale, iconPadding, iconAlongLine, globalProperties, featureProperties) {
+        layoutTextSize, textBoxScale, textPadding, textAlongLine,
```
ChrisLoer: Are `layoutTextSize` and `layoutIconSize` vestigial from some earlier version of this code? I don't see them being used.

src/shaders/_prelude.vertex.glsl
```diff
@@ -35,17 +35,25 @@ vec4 evaluate_zoom_function_4(const vec4 value0, const vec4 value1, const vec4 v
     }
 }
 
+// Unpack a pair of values that have been packed into a single float.
+// The packed values are assumed to be 8-bit unsigned integers, and are
+// packed like so:
+// packedValue = floor(input[0]) * 256 + input[1],
+vec2 unpack_float(const float packedValue) {
+    float v0 = floor(packedValue / 256.0);
```
ChrisLoer: I can't actually see a problem with the math/precision handling here, but for doing this type of math I would try to do the float -> int conversion as soon as possible (where we have to worry about weird precision stuff), and then do the rest of the math with ints instead of floats. Reasoning about things like `.99999 + 1 - 1 == 0.9999899999999999` is tricky.

I'm assuming glsl 100 reads UNSIGNED_SHORT vertex attributes as 32-bit floats... which I think should be able to handle up to 24 bits of integers without precision loss. I can't actually find where the behavior is specified. But I think we'd see more breakage if it were using half-floats.

src/shaders/symbol_sdf.vertex.glsl
```diff
@@ -43,8 +58,35 @@ void main() {
...
+    } else {
+        v_size = u_size;
+        adjustedSize = u_size;
+    }
+
+    float fontScale = u_is_text ? v_size / 24.0 : v_size;
+
+    mediump float zoomAdjust = log2(v_size / adjustedSize);
+    mediump float adjustedZoom = (u_zoom - zoomAdjust) * 10.0;
+    // result: z = 0 if a_minzoom <= adjustedZoom < a_maxzoom, and 1 otherwise
```
anandthakker: Yeah, I _think_ that's right.  I wasn't 100% sure of the clip space part, though, and so didn't include it in the comment in case that was incorrect and thus misleading.  But if that's your understanding, too, I think that's confirmation enough...

src/data/bucket/symbol_bucket.js
anandthakker: Yep, good catch, thanks!

src/shaders/_prelude.vertex.glsl
jfirebaugh: It's specified in [OpenGL® ES Shading Language](https://www.khronos.org/files/opengles_shading_language.pdf) §4.5.2 (p33). For `highp`, 16 bits of precision, integer range (-2¹⁶, 2¹⁶).

anandthakker:
@ChrisLoer I _think_ I've addressed the review points.  Still don't want to merge till -native port is done (finally making some progress there this evening), but could you check and see if this looks ready to you?

ChrisLoer:
@anandthakker It's looking good to me, and I've been using the branch locally for work on https://github.com/mapbox/mapbox-gl-js/tree/cloer_text_pitch_scale without any problems. I'm still hoping to get some time to dive into how the zoom stop interpolation works (it was the sort of thing that I read with the mindset "this is complicated... I'm sure it's fine"). No 👶  🐒  today so I probably can't get to that until tomorrow.

ChrisLoer:
@anandthakker These changes don't handle exponential interpolation consistently with other DDS styles. Here's the debug view I used to play with `text-size` vs `circle-radius`:

https://gist.github.com/ChrisLoer/a102c40ae5ad45b60c2d1b1d4bb738fc

With the base set to something high, you can see how the circles will change size rapidly as you approach the stop, while the text size doesn't. Also the 'interval' property function type doesn't work. 

I don't really have a great solution in mind, or haven't wrapped my head around what the solution would be. If we can't make changes to support these modes, we should at least disable them and document accordingly?

anandthakker:
Ah, dang.  Nice catch @ChrisLoer.  I suspect this is probably because I'm calculating the interpolation t-value [naively](https://github.com/mapbox/mapbox-gl-js/blob/af344c979f9f0de94ef162081e17985ebd55dda1/src/render/draw_symbol.js#L165) here, whereas for other DDS properties we use [StyleDeclaration#calculateInterpolationT](https://github.com/mapbox/mapbox-gl-js/blob/af344c979f9f0de94ef162081e17985ebd55dda1/src/style/style_declaration.js#L55-L66).

We may be able to do better here... but it's going to be complicated by the fact that for text/icon-size, when we're rendering a tile at `zoom = z`, we're basically dropping any stops that fall between `z` and `z+1`.

anandthakker:
@ChrisLoer updated to address the naive interpolation that you raised in https://github.com/mapbox/mapbox-gl-js/pull/4455#issuecomment-290589065.  We're still only using two stops (the best two that 'cover' the range `[tileZoom, tileZoom + 1]`) instead of the four that paint properties get, so it will be possible to have inconsistencies at certain zooms between a paint property and text/icon-size using fractional stop values... but I think this subtle discrepancy is acceptable for now.

ChrisLoer:
@anandthakker The approach looks good to me, although I haven't had the time to read through the changes in detail yet. What about zoom functions of the 'interval' type? I still don't see them being handled, but I may not have looked closely enough. I assume you can't have a 'categorical' zoom function since zoom doesn't come in discrete categories.

P.S. ❤️  the demystifying comments!

anandthakker:
> What about zoom functions of the 'interval' type?

Oh shoot, good point.  I missed this, because for _composite_ functions, we always treat the zoom component of the function as exponential (which is a separate question / potential bug?); but for camera functions, we should respect the 'interval' type.  Fixed in 2218a21

anandthakker:
@ChrisLoer @jfirebaugh I think this is ready for a final review.

ChrisLoer:
@anandthakker I'm working through the render tests this morning. 'interval' changes look good.

Could you explain your thinking on how these changes interact with collision detection? It looks like the idea is that since collision tiles are built on the `layoutTextSize` based on `tileZoom+1`, they should be on the conservative side. The collision logic assumes all labels on the tile shrink linearly (err, I guess exponentially but the point is "simply") as zoom increases -- or another way to put it is it assumes they stay the same screen size as you zoom in and out.

It's easy enough to make a text-size zoom function that violates the assumption in either direction (causing lots of collision or overly sparse placement). For the `text-pitch-scale` changes I've been trying to address this by adjusting the `a_labelminzoom` parameter that gets sent to the fragment shader based on the text size adjustments I'm doing in the vertex shader. However, that only works because I can make the simplifying assumption (approximation) that all labels changing scale the same way with respect to tilt. With DDS, there's so similar approximation you can make.

I can see the argument that this is just something the style designer has to be careful with, but it'd be nice to nail down what the constraints are for correct placement. "Don't make a label bigger as someone zooms out" may be pretty obvious -- and it might not even be that surprising if doing that leads to collisions.

ChrisLoer:
Playing around with this I noticed a kind of annoying problem even if you're on the "conservative" side of collision detection. Say you start at zoom `5.99`, with the collision box very close to correct. As you zoom out, the label gets smaller (faster than the collision box). At zoom `5.5`, the collision box gets triggered, and the label disappears. You keep zooming out, and then when you reach `4.99`, the collision box gets recalculated with the new smaller size, and the label pops back in. It seems to me this behavior could cause a lot of label flicker on crossing integer zoom levels, even for very "reasonable" zoom functions (my example is exaggerated, but with a less exaggerated zoom function, the collision box flicker will just happen over a very small zoom range, which could actually look more jarring).

PR #4457, opened 2017-03-20
Fix documentation type for Map.logoPosition

captainbarbosa:
Changes documentation type for `Map.logoPosition` from `{boolean}` to `{string}`

screen shot 2017-03-20 at 3 16 05 pm


lucaswoj:
Thank you!

Issue #4449, opened 2017-03-18
flyTo does not check the value of transform.renderWorldCopies


huaruiwu:
**mapbox-gl-js version**: v0.34.0

### Steps to Trigger Behavior

 1. initialize a map and set `option.renderWorldCopies = false`
 2. try to do a `flyTo` where `Math.abs(tr.center.lng) + Math.abs(center.lng) > 180`

### Expected Behavior
 `flyTo` should stay in the noncopied world where `-180 < lat < 180`.
### Actual Behavior
 `flyTo` will to to bring you to a worldCopy which is not rendered.

Something like this will fix the problem.
https://github.com/huaruiwu/mapbox-gl-js/commit/4baab4a23528098fe9563d80483b5367dd6b3aea

andrewharvey:
@huaruiwu I encourage you to submit this is a PR. Your fix looks good to me!

andrewharvey:
Fixed in 9941dde7e8f769f01917f3fc06b80a4d46c20422

PR #4459, opened 2017-03-20
make flyTo check if it can fly to worldCopies

huaruiwu:
## Launch Checklist

Issue is described here: https://github.com/mapbox/mapbox-gl-js/issues/4449.
For some reason I just get a `Not Found` when I try the debug page.

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -667,7 +667,7 @@ class Camera extends Evented {
 
         // If a path crossing the antimeridian would be shorter, extend the final coordinate so that
         // interpolating between the two endpoints will cross it.
-        if (Math.abs(tr.center.lng) + Math.abs(center.lng) > 180) {
+        if (tr._renderWorldCopies && Math.abs(tr.center.lng) + Math.abs(center.lng) > 180) {
             if (tr.center.lng > 0 && center.lng < 0) {
                 center.lng += 360;
             } else if (tr.center.lng < 0 && center.lng > 0) {
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -667,7 +667,7 @@ class Camera extends Evented {
 
         // If a path crossing the antimeridian would be shorter, extend the final coordinate so that
         // interpolating between the two endpoints will cross it.
-        if (tr._renderWorldCopies && Math.abs(tr.center.lng) + Math.abs(center.lng) > 180) {
+        if (tr.renderWorldCopies && Math.abs(tr.center.lng) + Math.abs(center.lng) > 180) {
             if (tr.center.lng > 0 && center.lng < 0) {
                 center.lng += 360;
             } else if (tr.center.lng < 0 && center.lng > 0) {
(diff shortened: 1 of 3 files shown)


andrewharvey:
@mollymerp With the changes I've proposed in #4451 calling `flyTo` to another world, it should wrap back to world 0. I'll do a quick manual test to confirm this.

andrewharvey:
>For some reason I just get a Not Found when I try the debug page.

@huaruiwu when `yarn start` starts the server the main debug page is at `http://localhost:9966/debug/debug.html`.

mollymerp:
@andrewharvey great thank you! #4451 takes care of my earlier comment. 

@huaruiwu sorry I missed your comment about the debug page - the default is `localhost:9966/debug/` and you can access the various other pages in the `debug/` folder by adding those filenames to the url as @andrewharvey suggested

huaruiwu:
@mollymerp @andrewharvey Thanks! It does say `localhost:9966/debug/` in the documentation but I just didn't see it.

andrewharvey:
> @mollymerp With the changes I've proposed in #4451 calling flyTo to another world, it should wrap back to world 0. I'll do a quick manual test to confirm this.

@mollymerp @huaruiwu I made some changes to #4451 to deal with a false `renderWorldCopies`, It will currently exit the main world `renderWorldCopies` is false but you call `flyTo` with coordinates in another world.

I think this circles back to your original question @mollymerp. What should the expected behaviour be here? I don't think it blocks this PR since it's a separate issue. I'm leaning towards leaving it as is, flyingTo to the white space. API users can always wrap their input to flyTo to prevent this. Leaving as is means than a box zoom outside the main world zooms in to that white area which I think is best.

test/unit/ui/camera.test.js
```diff
@@ -1011,6 +1011,46 @@ test('camera', (t) => {
             camera.flyTo({ center: [170, 0], duration: 10 });
         });
 
+        t.test('does not pan eastward across the antimeridian if no world copies', (t) => {
+            const camera = createCamera();
+            camera.transform._renderWorldCopies = false;
```
jfirebaugh: Please adjust the test so that it sets this option via the public API (an argument to the `Transform` constructor) rather than a private property.

src/ui/camera.js
```diff
@@ -667,7 +667,7 @@ class Camera extends Evented {
 
         // If a path crossing the antimeridian would be shorter, extend the final coordinate so that
         // interpolating between the two endpoints will cross it.
-        if (Math.abs(tr.center.lng) + Math.abs(center.lng) > 180) {
+        if (tr._renderWorldCopies && Math.abs(tr.center.lng) + Math.abs(center.lng) > 180) {
```
jfirebaugh: Please create a public `renderWorldCopies` getter on `Transform` rather than using an underscored (private) property.

huaruiwu:
@jfirebaugh Thanks for the review! updated

huaruiwu:
@jfirebaugh thanks for the merge

PR #3694, opened 2016-11-23
Add "placeholder-loading-*" properties

lucaswoj:
**[Play with the demo!](http://bl.ocks.org/lucaswoj/raw/66b9311793be3fd39b944779270f21e7/#16/40.6892/-74.0447)**

This PR allows developers to show a pattern or color in lieu of a still-loading raster tile. 

 - `placeholder-loading-color`
 - `placeholder-loading-pattern`
 - `placeholder-loading-opacity`

This feature is useful when loading raster tiles that use transparency to represent "no data." In the image below 👇  is there really no storm over SF? Or has the tile not loaded? Or has the tile errored? 

screen shot 2016-11-23 at 3 20 25 pm

In some industries, there is a regulatory requirement that a pattern be shown to disambiuate between "no data", "data loading", and "data errored"

cc @lbud @jfirebaugh @mourner 

fixes #3280 

## Launch Checklist

 - [x] create a new style layer type
 - [x] briefly describe the changes in this PR
 - [ ] write tests for all new functionality
 - [ ] document any changes to public APIs
 - [ ] post benchmark scores
 - [x] manually test the debug page

PR #4462, opened 2017-03-21
fix browserify + webpack

mollymerp:
Potential fix for the ongoing webpack saga 😭
@anandthakker @lucaswoj 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/docs/_includes/quickstart-bundler.html
+++ b/docs/_includes/quickstart-bundler.html
@@ -21,3 +21,18 @@ const map = new mapboxgl.Map({
 });
 {% endhighlight %}{% endcopyable %}
 <div>
+
+<p>If you're using <a target="_blank" href="https://webpack.js.org/">webpack</a>, add the following <a target="_blank" href="https://webpack.js.org/configuration/resolve/#resolve-alias">resolve.alias</a> to your config file</p>
+<div class='js-replace-token'>
+{% copyable %}{% highlight javascript %}
+{
+    resolve: {
+        alias: {
+          "mapbox-gl": "./node_modules/mapbox-gl/dist/mapbox-gl.js"
+          // or "mapbox-gl":path.resolve(__dirname, "./node_modules/mapbox-gl/dist/mapbox-gl.js")
+        }
+    }
+}
+{% endhighlight %}{% endcopyable %}
+<div>
+
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "mapbox-gl",
   "description": "A WebGL interactive maps library",
   "version": "0.34.0",
-  "main": "dist/mapbox-gl.js",
+  "main": "src/index.js",
   "license": "BSD-3-Clause",
   "repository": {
     "type": "git",


mollymerp:
manually tested that this works with webpack using https://gist.github.com/mollymerp/369e1287a473966e29fe92569495687e and editing the `main` field of `node_modules/mapbox-gl/package.json` to be `src/index.js`

anandthakker:
@mollymerp oooof thanks for figuring this out 😕 🏅 🌟 

I wonder if we should add browserify and webpack build tests.  Browserify would be easy to do without adding a new dependency.  Webpack's a heavy dep to add just for a unit test... maybe we could do that one only on CI?

anandthakker:
cc @mourner @jfirebaugh as well -- note that this reverts the change @mourner proposed in https://github.com/mapbox/mapbox-gl-js/issues/3551

lucaswoj:
Noting that at one point we had a browserpack unit test: https://github.com/mapbox/mapbox-gl-js/pull/3694

lucaswoj:
Nothing here that we did have a webpack test at one point: https://github.com/mapbox/mapbox-gl-js/commit/f6bcd061162bf01352169439d3e2fa0577179bb0#diff-cf4d6d7603881cba61aea62bcbf7e77d

lucaswoj:
Did you uncover the root cause of the build failures? I would much prefer a solution that doesn't require special steps for either webpack of browserify.

anandthakker:
@lucaswoj I'm not 100% on this, but I think the problem is that a browserified bundle is not, in fact, a valid CommonJS module.  In particular, my hypothesis is:
 - the browserified bundle will _itself_ have calls to `require`, but they are meant to refer to a private `require` function supplied by browserify in the bundle prelude.
 - when you try to _re_browserify that bundle (or rather, when you try to browserify a source file that requires it), those private `require` calls get treated as if they were "real" CommonJS `require`s

mollymerp:
I believe the root cause is that because the bundle we create in the /dist folder is `standalone` UMD module (what browserify creates), and so it wont work out of the box for webpack as @anandthakker said. I think in order for this to work in webpack there will be special instructions that need to go in the config file. Either what i've proposed above, or this https://webpack.js.org/concepts/output/#output-librarytarget set to `umd`

anandthakker:
> a browserified bundle is not, in fact, a valid CommonJS module

^ this might not be accurate, actually -- I think as Molly said it's a UMD module, which should work in a CommonJS environment (Node); but it _doesn't_ work when re-browserified because of [the `require` issue](https://github.com/mapbox/mapbox-gl-js/pull/4462#issuecomment-288485055)

anandthakker:
Just found https://github.com/calvinmetcalf/derequire -- it's possible that including this transform when building `dist/mapbox-gl.js` might allow browserify to work, in which case we could leave `main` set to `dist/mapbox-gl.js`.

jfirebaugh:
To me this looks like a bug in browserify's static analysis. A very simple example that demonstrates what's going on is:

```
module.exports = function(require) {
    require('nonesuch');
};
```

If you attempt to browserify this, you get:

```
Error: Cannot find module 'nonesuch' from '/Users/john/Development/mapbox-gl-js'
    at /Users/john/Development/mapbox-gl-js/node_modules/browser-resolve/node_modules/resolve/lib/async.js:46:17
    at process (/Users/john/Development/mapbox-gl-js/node_modules/browser-resolve/node_modules/resolve/lib/async.js:173:43)
    at ondir (/Users/john/Development/mapbox-gl-js/node_modules/browser-resolve/node_modules/resolve/lib/async.js:188:17)
    at load (/Users/john/Development/mapbox-gl-js/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (/Users/john/Development/mapbox-gl-js/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at /Users/john/Development/mapbox-gl-js/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:82:15)
```

Browserify shouldn't assume that any local variable that happens to be named `require` refers to the global.

jfirebaugh:
Relevant upstream tickets:

* https://github.com/substack/node-browserify/issues/374
* https://github.com/substack/node-browserify/pull/1151
* https://github.com/substack/node-detective/issues/22

anandthakker:
@jfirebaugh yah, I agree, but I don't think we should count on it getting
fixed anytime soon, alas

jfirebaugh:
Agreed. In the meantime, derequire (or perhaps [derequire-transform](https://github.com/calvinmetcalf/derequire-transform)) looks like the most commonly used workaround.

anandthakker:
I'm enjoying this comment in the README of `derequire-transform`

> It will only help in the case of a library that does something unwise like defining a local require function (*cough* ember *cough*)

since that's exactly what browserify does.

jfirebaugh:
The derequire approach works! #4486

PR #4478, opened 2017-03-22
Add `Map#loadImage` method and `Map#addImage` examples

lucaswoj:
This PR:

 - exposes `ajax#getImage` as `Map#loadImage` for use with `Map#addImage`
 - adds an example using `Map#addImage` with an external URL
 - adds an example using `Map#addImage` with a procedurally generated image

## Launch Checklist

 - [ ] write tests for all new functionality 😬 

### DIFF
--- a/docs/_posts/examples/.eslintrc
+++ b/docs/_posts/examples/.eslintrc
@@ -5,7 +5,8 @@
     "MapboxGeocoder": true,
     "MapboxDirections": true,
     "turf": true,
-    "d3": true
+    "d3": true,
+    "Uint8Array": true
   },
   "rules": {
     "strict": "off",
--- a/docs/_posts/examples/.eslintrc
+++ b/docs/_posts/examples/.eslintrc
@@ -5,8 +5,7 @@
     "MapboxGeocoder": true,
     "MapboxDirections": true,
     "turf": true,
-    "d3": true,
-    "Uint8Array": true
+    "d3": true
   },
   "rules": {
     "strict": "off",
@@ -17,7 +16,6 @@
     "prefer-template": "off"
   },
   "env": {
-    "es6": false,
     "browser": true
   }
 }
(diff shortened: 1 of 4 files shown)


docs/_posts/examples/.eslintrc
```diff
@@ -5,7 +5,8 @@
     "MapboxGeocoder": true,
     "MapboxDirections": true,
     "turf": true,
-    "d3": true
+    "d3": true,
+    "Uint8Array": true
```
mourner: Do you remember why we have `es6` env set to `false` a few lines below? As far as I remember, turning this back to `true` would allow not specifying common globals like this.

lucaswoj: As best I remember we were just to avoid using es6 features with limited browser support. This is less of a concern now. I have enabled the "es6" flag and removed this global in the latest commit.

lucaswoj:
@mollymerp There's a bunch of context on the naming decisions here: https://github.com/mapbox/mapbox-gl-style-spec/issues/220#issuecomment-261128332. 

There's a case to be made that these examples should be titled "Add an image to the map" but I think the current names will best direct those looking for this functionality.

docs/_posts/examples/.eslintrc
jfirebaugh: We need to continue to write examples without using ES6 features such as `let`/`const` or arrow functions (not supported by iOS 9) and `for...of` (not supported by IE 11). I think it's useful to keep this configuration option to catch accidental uses of these types of features. I suggest we revert the latest commit.

lucaswoj: Right -- the examples don't get `bubleify`ied. I'll revert the last commit and leave a comment about this in the eslint file.

mourner: @jfirebaugh as far as I remember, the `env` switches are only for predefined globals like `Uint8Array` — they're not used for catching keywords like `let`.

jfirebaugh: Yeah, you're right. And `"parserOptions": { "ecmaVersion": 5 }` doesn't work either. 😞

Issue #2900, opened 2016-07-21
Make markers anchored at their center by default


jensstalder:
https://www.mapbox.com/mapbox-gl-js/example/custom-marker-icons/

The images are anchored on the top left!!!!????

That makes them unusable.

If they were centered it would make 100000% more sense.

tmcw:
Hello!

Custom marker icons can be anchored any number of ways by using margins. This example doesn't do any modification of the anchor, and you're right - they would make more sense if they were anchored at the center. I'll make a note to fix the example.

Thanks for the issue - in the future, please spare the extra zeros, exclamation marks, and question marks: we can understand the intent of the issues without extra hyperbole.

- Tom

jensstalder:
^^ cool

So your saying something linke that?

```
el.style.marginLeft = -(imagewidth/2);
```

tmcw:
@jensstalder yes, exactly. Setting negative margins will shift the anchor point of the marker.

jensstalder:
ok cool.

mourner:
Also note that there will be an `offset` option in the next release: https://github.com/mapbox/mapbox-gl-js/pull/2885

1ec5:
I think this is an issue with the library, not the example. I think the default should be at the center, for consistency with symbols. The proper anchor of course depends on the image content (since a pushpin could be drawn to face in any direction), so #2885 is a welcome addition. But the most common marker images would either have a natural anchor at the center or the bottom-center, and I think it’s important for the various APIs for implementing a marker to have consistent behavior.

andrewharvey:
> I think this is an issue with the library, not the example. I think the default should be at the center, for consistency with symbols. The proper anchor of course depends on the image content (since a pushpin could be drawn to face in any direction), so #2885 is a welcome addition. But the most common marker images would either have a natural anchor at the center or the bottom-center, not the top-left, and I think it’s important for the various APIs for implementing a marker to have consistent behavior.

Spot on @1ec5. I agree.

One case for defaulting to the center is the userLocation marker proposed in #4479. In that PR if someone wants to change the marker dot style, eg. make it bigger or smaller, it's quite easy to override the default CSS, however the offset is hardcoded at https://github.com/mapbox/mapbox-gl-js/pull/4479/files#diff-8c498cf163efa60ad84199c531ab7c5aR283.

I understand that a workaround is to use a margin property in the marker css, but it just doesn't feel as simple or clean as defaulting to the center.

PR #4479, opened 2017-03-23
Add support to show and track user location in the GeolocateControl

### LINKED ISSUE #4354: Add show user location feature to GeolocateControl
Does `mapbox-gl-js` has a similar api that offers the my location/user location feature available on the native APIs?

```
#android
mapView.setMyLocationEnabled( true )
```

```
#ios
mapView.showsUserLocation = true
```

location circle

andrewharvey:
Fixes #4354
This PR builds upon #3781, please see that PR for background and design details.

This PR/branch uses a GL JS `Marker` (now that #4452 and #4452 are merged) to draw the user location dot rather than injecting the public style sheet. Rebased against master. Commits squashed.

From the original PR we did discuss some issues with the `fitBoundsOptions` option at https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r96941691. There wasn't a resolution, my thoughts are to leave as is and change once the new Camera API lands (since there will be breaking changes with it already).

## Breaking Changes

* The option `watchPosition` has been replaced with `trackUserLocation`.
* The camera operation has changed from `jumpTo` (not animated) to
  `fitBounds` (animated). An effect of this is the map pitch is no longer
  reset, although the bearing is still reset to 0.
* The accuracy of the geolocation provided by the device is used to set
  the view (previously it was fixed at zoom level 17). The `maxZoom` can
  be controlled via the new `fitBoundsOptions` option.

## New Features

* New option `showUserLocation` to draw a "dot" as a `Marker` on the map at
  the user's location.
* An active lock and background state are introduced with
  `trackUserLocation`. When in active lock the camera will update to follow
  the user location, however if the camera is changed by the API or UI
  then the control will enter the background state where it won't update
  the camera to follow the user location.
* New option `fitBoundsOptions` to control the camera operation.
* New `trackuserlocationstart` and `trackuserlocationend` events.
* New `LngLat.toBounds` method to extend a point location by a given
  radius to a `LngLatBounds` object.

## Launch Checklist
 - N/A post benchmark scores


src/ui/control/geolocate_control.js
```diff
@@ -40,27 +44,43 @@ function checkGeolocationSupport(callback) {
...
+ *
  * @implements {IControl}
  * @param {Object} [options]
- * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
- * @param {Object} [options.watchPosition=false] If `true` the map will reposition each time the position of the device changes and the control becomes a toggle.
+ * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
+ * @param {Object} [options.fitBoundsOptions={maxZoom: 18}] A [`fitBounds`](#Map#fitBounds) options object to use when the map is panned and zoomed to the user's location. The default is to use a `maxZoom` of 18 to limit how far the map will zoom in for very accurate locations.
```
lucaswoj: As we discussed in https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r96941691, I'm interested in exploring some other names for this parameter that emphasize its relationship to "user location" and downplay its relationship to the soon-to-be-deprecated `Map#fitBounds` method (#2801).

src/ui/control/geolocate_control.js
```diff
@@ -139,26 +373,65 @@ class GeolocateControl extends Evented {
...
+ * @event activeLock
+ * @memberof GeolocateControl
+ * @instance
+ *
+ */
+
+/**
+ * Fired when the Geolocate Control changes to the background state, which happens when a user changes the camera during an active position lock. This only applies when trackUserLocation is true. In the background state, the dot on the map will update with location updates but the camera will not.
+ *
+ * @event background
```
lucaswoj: What do you think about renaming these events to:

 - `activeLock` -> `trackuserlocationstart`
 - `background` -> `trackuserlocationend`

These names are consistent with existing event names (`start` and `end` suffixes) and more self-documenting (`activeLock` isn't a term of art AFIK). The all lowercase naming convention for events is unfortunate but consistent in our codebase so far.

andrewharvey: I'm okay with that. Your reasoning makes sense.

src/ui/control/geolocate_control.js
andrewharvey: From https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r97237660.

> Looking through the new camera api (at least based on the current API docs in #3583) the new API call would be:
> 
> map.fitBounds(bounds, cameraOptions, animationOptions)
> 
> It would make sense to have the GeolocateControl have and pass through options.cameraOptions and options.animationOptions. (and have cameraOptions support a maxZoom)
>
> However at the moment fitBounds currently supports a few extra options which the AnimationOptions interface doesn't. linear, padding, maxZoom. So would you suggest I have an options.animationOptions and options.linear and options.padding and options.maxZoom? And then once the new camera api lands change the API to bring linear into AnimationOptions as type and padding,maxZoom into options.cameraOptions?
>
> Either way there will be a public API change, but the new camera API already does that, so I think it's simpler to leave as is at the moment and then refactor for the new camera api.

I'm not sure the best way forward.

andrewharvey:
- [ ] move test file from legacy to new test/unit directory
- [ ] update tests based on using HTMLMarker
- [ ] rename `activeLock` and `background` events
- [ ] resolve fitBoundsOptions issue

andrewharvey:
From some testing, I've found that using
```
fitBoundsOptions: {
   linear: true
   duration: 0
}
```
seems to work better than the default easing. I feel it could be made the default but I'm not sure, anyone else have an opinion form a UX perspective?

andrewharvey:
Is there anything else I can do to help move this PR forward?

src/ui/control/geolocate_control.js
mollymerp: I think something `userLocationViewOptions` would work with what Lucas was suggesting (at the risk of extra verbosity 💭 )

mollymerp:
@andrewharvey sorry for the delay here. @lucaswoj is hiking from Mexico to Canada right now so it looks like this slipped by other reviewers. I will plan to give it a proper review on Monday.

src/ui/control/geolocate_control.js
andrewharvey: :confused: not sure exactly what that would look

andrewharvey:
@mollymerp rebased against master and updated to pass lint-css tests.

andrewharvey:
rebased against master, squashed commits and updated commit message based on recent changes to event names.

pathmapper:
@andrewharvey thanks a lot for your work on this! 

It works fine - one thing I noticed while playing around with it: 
With the latest release of Firefox for Android (53.0.2) there are all glyphs missing on the map (streets-v10). When using Chrome for Android or Firefox for Desktop the glyphs are there, also when using v0.37.0 of mapbox-gl.js with Firefox for Android (53.0.2).
 
I don't know whether this issue is related to your commit. It could be also that I was doing something wrong building mapbox-gl.js or it is a bug in Firefox or whatever - just wanted to let you know, maybe you are able to reproduce it.

andrewharvey:
I don't know whether this issue is related to your commit. It could be also that I was doing something wrong building mapbox-gl.js or it is a bug in Firefox or whatever - just wanted to let you know, maybe you are able to reproduce it.

@pathmapper It sounds unrelated to this PR. Do you notice it in https://www.mapbox.com/mapbox-gl-js/examples ? 

Not sure if it's https://github.com/mapbox/mapbox-gl-js/issues/4607 but I would try upgrading to master and see if that fixes it.

pathmapper:
@andrewharvey you're right, it is unrelated to this PR. Thanks for pointing me to the right direction.

> Do you notice it in https://www.mapbox.com/mapbox-gl-js/examples ?

Yes

> I would try upgrading to master and see if that fixes it.

It happens only with master, with latest release v0.37.0 everthing is fine.
Created an issue: https://github.com/mapbox/mapbox-gl-js/issues/4747

src/ui/control/geolocate_control.js
```diff
@@ -40,27 +44,42 @@ function checkGeolocationSupport(callback) {
...
  * }));
  */
 class GeolocateControl extends Evented {
 
     constructor(options) {
         super();
         this.options = options || {};
+
+        // apply default for options.showUserLocation
+        this.options.showUserLocation = (this.options && 'showUserLocation' in this.options) ? this.options.showUserLocation : true;
```
mollymerp: could you use `util.extend` (see https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/map.js#L138 for example) to set the defaults?

src/ui/control/geolocate_control.js
```diff
@@ -107,29 +241,118 @@ class GeolocateControl extends Evented {
...
+            this._watchState = 'OFF';
+        }
+
+        // when showUserLocation is enabled, keep the Geolocate button disabled until the device location marker is setup on the map
+        if (this.options.showUserLocation) {
+            if (this.options.trackUserLocation) this._watchState = 'INITILIZE';
+
+            this._dotElement = DOM.create('div', 'mapboxgl-user-location-dot');
+
+            this._userLocationDotMarker = new Marker(this._dotElement, { offset: [-13, -13] });
```
mollymerp: can you add a comment explaining the `offset` constants?

src/ui/control/geolocate_control.js
```diff
@@ -40,27 +44,42 @@ function checkGeolocationSupport(callback) {
...
+ *
  * @implements {IControl}
  * @param {Object} [options]
- * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
- * @param {Object} [options.watchPosition=false] If `true` the map will reposition each time the position of the device changes and the control becomes a toggle.
+ * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
+ * @param {Object} [options.fitBoundsOptions={maxZoom: 18}] A [`fitBounds`](#Map#fitBounds) options object to use when the map is panned and zoomed to the user's location. The default is to use a `maxZoom` of 18 to limit how far the map will zoom in for very accurate locations.
+ * @param {Object} [options.trackUserLocation=false] If `true` the Geolocate Control becomes a toggle button and when active the map will receive updates to the user's location as it changes.
```
mollymerp: is `trackUserLocation` this just renaming `watchPosition`? I think we should avoid a breaking change if that's the case.

andrewharvey: Yes, it was suggested over at https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r96940756 to provide consistency with GL native.

I find `showUserLocation` and `trackUserLocation` go hand in hand. `watchPosition` fits in better with the HTML Geolocation API terminology, but I think `trackUserLocation` fits in better with what this option does in GL JS.

`watchPositon` was only recently introduced (well recently before the original PR for this) so I feel real world usage mightn't be that big.

src/ui/control/geolocate_control.js
andrewharvey: Will do.

On that note, I'm in favour of https://github.com/mapbox/mapbox-gl-js/issues/2900 which would eliminate offset here (and also make it much easier for people to override CSS styles to change the dot size for their app.

src/ui/control/geolocate_control.js
```diff
@@ -4,8 +4,12 @@ const Evented = require('../../util/evented');
 const DOM = require('../../util/dom');
 const window = require('../../util/window');
 const util = require('../../util/util');
+const assert = require('assert');
+const LngLat = require('../../geo/lng_lat');
+const Marker = require('../marker');
 
 const defaultGeoPositionOptions = { enableHighAccuracy: false, timeout: 6000 /* 6sec */ };
+const defaultFitBoundsOptions = { maxZoom: 18 };
```
andrewharvey: Actually I settled upon this after real world testing using it to navigate walking around, but I agree z15 would be a better default for desktop maps, so I'll change this.

The best value is going to depend a lot on the application/use case, device type (mobile or desktop), even if the user is moving and how fast (on a plane, or walking down the street would warrant different zooms) so being configurable is good.

src/geo/lng_lat.js
```diff
@@ -73,6 +73,24 @@ class LngLat {
...
+     * @returns {LngLatBounds} A new `LngLatBounds` object representing the coordinates extended by the `radius`.
+     * @example
+     * var ll = new mapboxgl.LngLat(-73.9749, 40.7736);
+     * ll.toBounds(100).toArray(); // = [[-73.97501862141328, 40.77351016847229], [-73.97478137858673, 40.77368983152771]]
+     */
+    toBounds(radius: number) {
+        const latAccuracy = 360 * radius / 40075017,
+            lngAccuracy = latAccuracy / Math.cos((Math.PI / 180) * this.lat);
+
+        const LngLatBounds = require('./lng_lat_bounds');
```
andrewharvey: Yeah explained at https://github.com/mapbox/mapbox-gl-js/pull/3737#issuecomment-266634879 it was due to circular dependency issues which arose when it was outside the function :wink:

dist/mapbox-gl.css
```diff
@@ -303,6 +343,65 @@ a.mapboxgl-ctrl-logo {
     will-change: transform;
 }
 
+.mapboxgl-user-location-dot {
+    background-color: #1DA1F2;
+    width: 20px;
```
andrewharvey: ok

dist/mapbox-gl.css
```diff
@@ -303,6 +343,65 @@ a.mapboxgl-ctrl-logo {
...
+    height: 20px;
+    border-radius: 50%;
+    box-shadow: 0 0 2px rgba(0,0,0,0.25);
+    border: 3px solid #fff;
```
andrewharvey: ok

dist/mapbox-gl.css
```diff
@@ -303,6 +343,65 @@ a.mapboxgl-ctrl-logo {
...
+    animation: pulse 2s;
+
+    -webkit-animation-iteration-count: infinite;
+    -moz-animation-iteration-count: infinite;
+    -ms-animation-iteration-count: infinite;
+    animation-iteration-count: infinite;
+}
+@-webkit-keyframes pulse {
+    0%   {   -webkit-box-shadow: 0 0 0 0 rgba(29, 161, 242, 0.8); }
+    70%  {   -webkit-box-shadow: 0 0 0 30px rgba(29, 161, 242, 0); }
```
andrewharvey: ok

src/geo/lng_lat.js
```diff
@@ -73,6 +73,24 @@ class LngLat {
...
+     * Returns a `LngLatBounds` from the coordinates extended by a given `radius`.
+     *
+     * @param {number} radius Distance in meters from the coordinates to extend the bounds.
+     * @returns {LngLatBounds} A new `LngLatBounds` object representing the coordinates extended by the `radius`.
+     * @example
+     * var ll = new mapboxgl.LngLat(-73.9749, 40.7736);
+     * ll.toBounds(100).toArray(); // = [[-73.97501862141328, 40.77351016847229], [-73.97478137858673, 40.77368983152771]]
+     */
+    toBounds(radius: number) {
+        const latAccuracy = 360 * radius / 40075017,
```
andrewharvey: ok

dist/mapbox-gl.css
andrewharvey: It feels a tad tiny at 12px, to the point where it almost dissapears in the map at a glance, and compared to the size of other UI elements (like zoom buttons).

Noting at https://github.com/mapbox/mapbox-gl-js/pull/3781#discussion_r98823742 that the design came from @tristen .

src/ui/control/geolocate_control.js
andrewharvey: What's the reasoning behind keeping jumpTo? Is it related to the camera animation difference between fitBounds and jumpTo?

This was introduced in https://github.com/mapbox/mapbox-gl-js/pull/3737 which I merged into this PR. The reason for fitBounds is to take into account the Geolocation API accuracy which I feel is a useful feature, animation of the camera is a seperate issue but until the camera api refactor lands it is muddled into each api.

For example if the Geolocation accuracy is only city level, the map should only zoom to the size of that city.

src/geo/lng_lat.js
mollymerp: 👌 sorry I should have reviewed the comments in the old PR!

src/ui/control/geolocate_control.js
andrewharvey: I've done this, but also made a few other changes in other parts to simplify how default options are applied.

src/geo/lng_lat.js
andrewharvey: no worries

src/ui/control/geolocate_control.js
mollymerp: Yes, I agree that #2900 is something we should implement.

dist/mapbox-gl.css
mollymerp: Ah, I see. Maybe we could compromise at 16px? 😸

andrewharvey:
>* The option watchPosition has been replaced with trackUserLocation.

Comments inline above.

>* The camera operation has changed from jumpTo (not animated) to fitBounds (animated). An effect of this is the map pitch is no longer reset, although the bearing is still reset to 0 (a limitation of fitBounds).

The change from not animated to animated was necessary to get the extra options fitBounds provides, namely zoom to bounds (geolocation accuracy) and maxZoom. The fact that it's swapped from not animated to animated would only be fixed by the camera api changes to unbundle animation from camera parameters.

>* The accuracy of the geolocation provided by the device is used to set the view (previously it was fixed at zoom level 17). The maxZoom can be controled via the new fitBoundsOptions option.
>I'm not sure any of these breaking changes are 100% necessary to add the functionality you're looking for here. Am I'm missing something? I think we can add the maxZoom option while remaining backwards compatible.

Because it now uses geolocation accuracy which affects maxZoom, it can't have both this functionality and always use a fixed zoom (as it did previously) hence the breaking change.

dist/mapbox-gl.css
andrewharvey: I think that works, at that size it's pretty much the same size as the icon in the control button which I think is good.

mollymerp:
Thanks for the updates. I think I was just wondering if the added complexity of using geolocation accuracy at all is really necessary. I'm not sure it would be clear in this state to an end user that the bounds of the map view represent the accuracy of their geolocation, so I don't know if this is a value-add for the user.

andrewharvey:
> Thanks for the updates. I think I was just wondering if the added complexity of using geolocation accuracy at all is really necessary. I'm not sure it would be clear in this state to an end user that the bounds of the map view represent the accuracy of their geolocation, so I don't know if this is a value-add for the user.

It was more to prevent the map being on a street scale when the accuracy of the geocode is only city level, and vice versa allowing a street scale zoom when the accuracy of the geocode is within the meters.

mollymerp:
ok – I would still like to remove the connection to `fitBounds` with this method before merging. Can you move `maxZoom` to its own property in `options`? It would also be great if you could document all the different `_watchState` values and their meanings in some way.

src/ui/control/geolocate_control.js
```diff
@@ -40,27 +53,39 @@ function checkGeolocationSupport(callback) {
  * geolocation support is not available, the GeolocateControl will not
  * be visible.
  *
+ * The zoom level applied will depend on the accuracy of the geolocation provided by the device.
+ *
+ * The GeolocateControl has two modes. If `trackUserLocation` is `false` (default) the control acts as a button, which when pressed will set the map's camera to target the user location. If the user moves, the map won't update. This is most suited for the desktop. If `trackUserLocation` is `true` the control acts as a toggle button that when active the user's location is actively monitored for changes. In this mode there is a concept of an active lock and background. In active lock the map's camera will automatically update as the users's location changes until the user manually changes the camera (such as panning or zooming). When this happens the control is in background so the user's location dot still updates but the camera doesn't.
```
anandthakker: I think the explanation of `trackUserLocation: true` mode is still a bit hard to follow.  Suggestion to clarify:

---
In this mode, the GeolocateControl has three states:
 - _active_  - the map's camera automatically updates as the user's location changes, keeping the location dot in the center.
 - _passive_ - the user's location dot automatically updates, but the map's camera does not.
 - _disabled_.

When a user first clicks the control, it becomes `active`.  When the user manually pans or zooms the map, the geolocate control goes from `active` to `passive` mode.  Clicking the control returns the control to the `active` state. _(Is this true?)_

src/ui/control/geolocate_control.js
```diff
@@ -72,24 +97,139 @@ class GeolocateControl extends Evented {
...
+            case 'BACKGROUND_ERROR':
+                this._watchState = 'BACKGROUND';
+                this._geolocateButton.classList.remove('mapboxgl-ctrl-geolocate-waiting');
+                this._geolocateButton.classList.remove('mapboxgl-ctrl-geolocate-background-error');
+                this._geolocateButton.classList.add('mapboxgl-ctrl-geolocate-background');
+                break;
+            default:
+                assert(false, `Unexpected watchState ${this._watchState}`);
+            }
+            this._geolocateButton.classList.remove('mapboxgl-ctrl-geolocate-waiting');
```
anandthakker: Nit: I think this line makes the equivalent lines in the `case`s above redundant.

src/ui/control/geolocate_control.js
```diff
@@ -107,58 +247,177 @@ class GeolocateControl extends Evented {
...
+
         this._geolocateButton.addEventListener('click',
             this._onClickGeolocate.bind(this));
+
+        // when the camera is changed (and it's not as a result of the Geolocation Control) change
+        // the watch mode to background watch, so that the marker is updated but not the camera.
+        if (this.options.trackUserLocation) {
+            this._map.on('movestart', (event) => {
+                if (!event.geolocateSource) {
+                    if (this._watchState === 'ACTIVE_LOCK') {
```
anandthakker: Nit: change the nested `if`s to `if(!event.geolocateSource && this._watchState == 'ACTIVE_LOCK
'`)

src/ui/control/geolocate_control.js
```diff
@@ -72,24 +97,139 @@ class GeolocateControl extends Evented {
...
+        if (!this.options.trackUserLocation || this._watchState === 'ACTIVE_LOCK') {
+            this._updateCamera(position);
+        }
+
+        if (this.options.showUserLocation) {
+            this._dotElement.classList.remove('mapboxgl-user-location-dot-stale');
+        }
+
+        if (this._watchState === 'ACTIVE_LOCK') {
+            this.fire('trackuserlocationstart');
```
anandthakker: I may be misunderstanding something here, but it looks to me like this will get fired each time the tracked location is updated, rather than only when we _start_ tracking the location.

test/unit/ui/control/geolocate.test.js
```diff
@@ -0,0 +1,299 @@
...
+            duration: 0
+        }
+    });
+    map.addControl(geolocate);
+
+    const click = new window.Event('click');
+
+    let moveendCount = 0;
+    map.once('moveend', () => {
+        // moveend was being called a second time, this ensures that we don't run the tests a second time
```
anandthakker: The callback was being called a second time despite the use of `map.once()`?

andrewharvey: yes that is what I observed. I couldn't work out why that was happening so I just worked around it with this count.

andrewharvey:
> ok – I would still like to remove the connection to fitBounds with this method before merging. Can you move maxZoom to its own property in options?

Are you saying you don't want the feature where it takes into account the geolocation accuracy at all? or you're happy with that but just don't want to use fitBounds?

I think it's important to let the user set all [fitBounds options](https://www.mapbox.com/mapbox-gl-js/api/#map#fitbounds) so they can tweak how the animation works, or remove it (which might even be the better default)?

> It would also be great if you could document all the different _watchState values and their meanings in some way.

back in the [original PR](https://github.com/mapbox/mapbox-gl-js/pull/3781#issue-194914626) I had a state diagram:

3140861e-c0a0-11e6-8298-3bcb5850cda2

however a few things have changed since then. so my question is does it make sense to improve the diagram and use that as the (internal) documentation or it needs to be documented in the js file directly?

mollymerp:
>Are you saying you don't want the feature where it takes into account the geolocation accuracy at all? or you're happy with that but just don't want to use fitBounds?

@andrewharvey the second one 😄. The only options exclusive to fitbounds (i.e. not represented in `CameraOptions` or `AnimationOptions` ) are `padding` and `maxZoom`, so I don't think that's too many to add to the root `options` object for the `GeolocateControl` 

The diagram is awesome! I think it's great to have here on the PR for reference in the future, but I was thinking more just a short comment in-line explaining the difference between `WAITING ACTIVE`/`ACTIVE LOCK`/`BACKGROUND` and the difference between the two error types.

🙏 thanks for your patience on this! so close!!

andrewharvey:
> @andrewharvey the second one ￼. The only options exclusive to fitbounds (i.e. not represented in CameraOptions or AnimationOptions ) are padding and maxZoom, so I don't think that's too many to add to the root options object for the GeolocateControl

@mollymerp Sorry I'm not convinced :confused:. It feels messier than simply accepting a single `fitBoundsOption` parameter, which in the future would simply be replaced with `cameraOptions` when the new camera api lands. I'm not sure what you gain by having them as root options?

The fitBounds options at https://www.mapbox.com/mapbox-gl-js/api/#map#fitbounds I understand will accept all the options of AnimationOptions, CameraOptions + those described in fitBounds, so at the moment you could pass in:

```
fitBoundsOptions: {
   maxZoom: 19,
   linear: true,
   animate: false
}
```

Are you suggesting instead of this the GeolocateControl has options `cameraMaxZoom, cameraLinear, cameraPadding, cameraOptions (CameraOptions|AnimationOptions)`, which are then all merged together anyway to pass to `fitBounds`?

> The diagram is awesome! I think it's great to have here on the PR for reference in the future, but I was thinking more just a short comment in-line explaining the difference between WAITING ACTIVE/ACTIVE LOCK/BACKGROUND and the difference between the two error types.

Agreed, it's much easier to understand with the diagram, but it's outdated now as the states and class names have changed. I've tried to add in some internal docs on the states, plus I've removed the INIT state which wasn't doing anything (it was more important before when the user location dot was rendered as a Layer and we needed to wait for the map to load to add the Layer but now it's using Marker it's not doing anything).

michael8113:
Hi guys. Where may see the working example of geolocation work with the location point display? Thank you

andrewharvey:
> Hi guys. Where may see the working example of geolocation work with the location point display? Thank you

You should be able to do

    yarn install
    yarn run start-debug

and check out the main debug example http://localhost:9966/debug

michael8113:
Ок, If I just use external https://api.mapbox.com/mapbox-gl-js/v0.38.0/mapbox-gl.js, can not I just plug it in without having to install it?(

mollymerp:
Thanks for your patience + contribution @andrewharvey! 🎉

mollymerp:
@michael8113 this will go out in the next release `v0.39.0`

1updesign:
hey @mollymerp - any idea when v0.39.0 is due? or perhaps the best way to get this feature working on a stable version in the meantime? thanks!

mollymerp:
@1updesign We'll be doing a release this week!

nrako:
@andrewharvey Thanks for this, this is great with a nice UX!

I'm just wondering if there's any reason for the control to not fire an event on any `_watchState` change?

As a simple example, I'd like to be able to know when the control is `OFF` or **not** (`BACKGROUND`, `ACTIVE_LOCK`). I feel having a `statechange` event would be more useful than `trackuserlocationstart` and `trackuserlocationend`

andrewharvey:
> I'm just wondering if there's any reason for the control to not fire an event on any _watchState change?
> As a simple example, I'd like to be able to know when the control is OFF or not (BACKGROUND, ACTIVE_LOCK). I feel having a statechange event would be more useful than trackuserlocationstart and trackuserlocationend

@nrako I think you're right, a statechange event might be better. It would need to pass a value to indicate the new (and possible previous state) but I don't see that as an issue.

Do you think you need all the error states or just active, passive, disabled per https://www.mapbox.com/mapbox-gl-js/api/#geolocatecontrol.event:geolocate?

Could you open a new issue for this?

nrako:
@andrewharvey – Thanks, I've created the issue https://github.com/mapbox/mapbox-gl-js/issues/5136

I don't believe the previous state would need to be passed.
Maybe passing some states aren't useful and would just clutter the API. I leave this discussion for the issue => #5136 

On the bonus side, I believe this could help refactor some similar logics which are spread in different handlers. I put some clue in the implementation part.

If the API and implementation design are agreed upon. I don't mind pushing a PR if my time allows it.

Issue #4370, opened 2017-03-06
Distribution CSS is not minified


jasonbarry:
The distribution CSS is unminified: 
https://api.mapbox.com/mapbox-gl-js/v0.32.1/mapbox-gl.css

While the JS at the same location (w/ same file naming convention) is minified: 
https://api.mapbox.com/mapbox-gl-js/v0.32.1/mapbox-gl.js

jasonbarry:
It would be nice to have a build process that takes the source CSS and minifies it to dist.

jfirebaugh:
I don't think we're likely to implement this. The savings potential is very small when taking gzip content encoding into account; see https://github.com/mapbox/mapbox-gl-js/pull/4493#issuecomment-289170896

PeterDaveHello:
I think the effort isn't so huge, not acceptable?

PR #4493, opened 2017-03-24
Add build-css script to minify dist CSS

jasonbarry:
This PR addresses issue #4370 by using [cssnano](http://cssnano.co) to minify development CSS into production CSS. 

Summary of changes: 
1. `cssnano-cli` added to devDependencies
2. mapbox-gl.css was moved from `/dist` to `/src`
3. a `build-css` script was added to `package.json` that minifies and outputs the CSS in `/src` to `/dist`
4. the `prepublish` script was modified to include the new `build-css` script

The minified CSS output went from 24,595 bytes to 21,763 bytes (11.5% reduction in file size).


jfirebaugh:
Hi @jasonbarry, thanks for the contribution!

The majority of the time, this CSS will be served by a CDN or other static asset server that applies gzip content encoding to the body. The benefits of pre-compressing CSS are much reduced when taking into account gzip content encoding. In my measurements, it goes from 7257 bytes to 7094 bytes. I don't think a 163 byte improvement is enough to justify the additional complexity in the build pipeline or the loss in readability if you're looking at the dist CSS asset.

For these reasons, I'm going to decline this pull request, but I nevertheless appreciate the contribution, and thanks for using Mapbox GL.

jasonbarry:
No prob @jfirebaugh. Thanks for the polite explanation!

Issue #4107, opened 2017-02-01
Prevent step and interpolate expressions from having duplicate stops


lucaswoj:
_From @scothis on February 5, 2016 22:51_

Restore https://github.com/mapbox/mapbox-gl-style-spec/pull/397 with a migration for existing ramps similar to https://github.com/mapbox/mapbox-gl-style-spec/pull/411 for v9.

This behavior was merge in v8 and removed because it was backwards incompatible. The idea is sound and should be part of v9

_Copied from original issue: mapbox/mapbox-gl-style-spec#414_

jfirebaugh:
We did this for expressions; for functions, we're more likely to just drop support for them entirely than make any other breaking change to them.

jfirebaugh:
We didn't actually do this for expressions, but we should.

PR #4498, opened 2017-03-27
Validate duplicate stops in stylesheets.

vicapow:
Otherwise, duplicate pairs for a stop in a  style can cause [binarySearchForIndex()](https://github.com/mapbox/mapbox-gl-js/blob/b9e10b939c6a3fe5d7ecac209f751b4871970ede/src/style-spec/function/index.js#L211) to enter an infinite loop. This didn't immediately cause issues because of an earlier bug in [mapbox-gl-function](https://github.com/mapbox/mapbox-gl-function/commit/4f829622413f336080d3c710244c251421c0ec30) that was introduced between 0.32.1 and 0.33 as part of [this rebase with the mapbox-gl-function library](https://github.com/mapbox/mapbox-gl-js/commit/d92eb6eb397b9f05ddfb5ab52b1704cace14cc36#diff-b9cfc7f2cdf78a7f4b91a753d10865a2).

### DIFF
--- a/src/style-spec/validate/validate_function.js
+++ b/src/style-spec/validate/validate_function.js
@@ -78,8 +78,22 @@ module.exports = function validateFunction(options) {
             arrayElementValidator: validateFunctionStop
         }));
 
-        if (getType(value) === 'array' && value.length === 0) {
-            errors.push(new ValidationError(options.key, value, 'array must have at least one stop'));
+        if (getType(value) === 'array') {
+            if (value.length === 0) {
+                errors.push(new ValidationError(options.key, value, 'array must have at least one stop'));
+            } else {
+                const seenStops = Object.create(null);
+                value.every(stop => {
+                    if (getType(stop) !== 'array' || getType(stop[0]) !== 'number') {
+                        return false;
+                    }
+                    if (seenStops[stop[0]]) {
+                        errors.push(new ValidationError(options.key, value, 'stops must not have duplicate pairs.'));
+                    }
+                    seenStops[stop[0]] = stop[1];
+                    return true;
+                });
+            }
         }
 
         return errors;
--- a/test/unit/style-spec/fixture/functions.input.json
+++ b/test/unit/style-spec/fixture/functions.input.json
@@ -839,6 +839,20 @@
           "default": "invalid"
         }
       }
+    },
+    {
+      "id": "invalid duplicate stop pairs",
+      "type": "symbol",
+      "source": "source",
+      "source-layer": "layer",
+      "layout": {
+        "text-size": {
+          "stops": [
+            [0, 1],
+            [0, 1]
+          ]
+        }
+      }
     }
   ]
 }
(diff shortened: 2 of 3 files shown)


src/style-spec/validate/validate_function.js
```diff
@@ -78,8 +78,22 @@ module.exports = function validateFunction(options) {
...
+                value.every(stop => {
+                    if (getType(stop) !== 'array' || getType(stop[0]) !== 'number') {
+                        return false;
+                    }
+                    if (seenStops[stop[0]]) {
+                        errors.push(new ValidationError(options.key, value, 'stops must not have duplicate pairs.'));
+                    }
+                    seenStops[stop[0]] = stop[1];
+                    return true;
+                });
```
mourner: Since we don't actually use the returned value of `every`, it makes more sense to use a `for ... of` loop here.

mourner:
Also, what's the earlier bug you mentioned and should it be reported as well?

jfirebaugh:
See discussion on #4107 and https://github.com/mapbox/mapbox-gl-style-spec/pull/411: we tried adding this validation, but reverted it when it became clear that styles with duplicate stops were somewhat widespread and it would be difficult to migrate them without incrementing the style spec version number. Do we want to revisit that decision?

mourner:
On a second thought, we should probably fix the infinite loop problem in the binary search implementation rather then avoiding the situation with validation then. cc @kronick

vicapow:
> Also, what's the earlier bug you mentioned and should it be reported as well?

Prior to https://github.com/mapbox/mapbox-gl-function/pull/35, duplicate stops wouldn't trigger the infinite loop in my styles. It landed in mapbox-gl-js as part of [this merge](https://github.com/mapbox/mapbox-gl-js/commit/d92eb6eb397b9f05ddfb5ab52b1704cace14cc36). Specifically:

```diff
-    "mapbox-gl-function": "mapbox/mapbox-gl-function#41c6724e2bbd7bd1eb5991451bbf118b7d02b525",
+    "mapbox-gl-function": "mapbox/mapbox-gl-function#4f829622413f336080d3c710244c251421c0ec30",
```

Just to confirm, should [binarySearchForIndex()](https://github.com/mapbox/mapbox-gl-js/blob/b9e10b939c6a3fe5d7ecac209f751b4871970ede/src/style-spec/function/index.js#L204) be updated to only pick the first stop when there's duplicates?

lucaswoj:
> See discussion on #4107 and mapbox/mapbox-gl-style-spec#411: we tried adding this validation, but reverted it when it became clear that styles with duplicate stops were somewhat widespread and it would be difficult to migrate them without incrementing the style spec version number. Do we want to revisit that decision?

This seems like a change that'd need to wait for a semver major version bump. There are several other validation strict-ening changes we'd probably want to make at once. (i.e. disallowing arbitrary properties on the style root object)

vicapow:
I agree. It should fix this problem in the short term if we change `binarySearchForIndex` to not enter an infinite loop when there are duplicate keys.

A newer version of the spec can enforce the prevention of duplicate keys.

PR #4502, opened 2017-03-27
Add integration tests for Mapbox Streets core style

### LINKED ISSUE #3909: Add render tests for Mapbox core styles
Migrated from https://github.com/mapbox/mapbox-gl-test-suite/issues/26

> I've seen the test suite pass when a visual inspection of Mapbox Streets reveals many errors. This is especially true for bugs that involve global state.
>
> It doesn't seem to be in the spirit of the existing tests here, but we should thinking about adding some integration tests to the test suite -- tests that render full Mapbox Streets tiles -- to give us confidence that changes to GL don't break real world use cases.

lucaswoj:
This PR adds an integration test that renders the `mapbox/streets-v10` style. This test should help us catch bugs that only show up in large complex styles.

expected
expected

Fixes #3909

## Open Questions

 - Can we afford to have network requests within an integration test? If not, what's the best way to pull in the Mapbox Streets style? 
 - How many coordinates should we test? 
 - What other styles should we test? (Satellite didn't work out of the box. Something to do with image encoding.)
 - How likely are we to get a false signal due to a upstream style change? 

## Launch Checklist


jfirebaugh:
Ideally we'd pull the style via a module dependency, so that it doesn't require network requests and changes only when we update yarn.lock. Maybe it would be sufficient to test the Bright style from mapbox-gl-styles instead of Mapbox Streets.

anandthakker:
This is a great idea!

I wonder if it would make the image larger, i.e. 512x512.  With other integration tests, keeping the image small/simple helps the tests run faster and (most importantly) makes them easier to inspect for correctness.  For this set of tests, I think seeing more would make it easier to inspect, and casting a wider net would cover more potential cases that arise from combinations of many layers, which is part of the purpose.

lucaswoj:
> Maybe it would be sufficient to test the Bright style from mapbox-gl-styles instead of Mapbox Streets.

This is possible and presents some different tradeoffs:

 - We'd need to check in all vector tiles needed for test
 - We'd need to modify the style at runtime to make it use offline tiles
 - We would be testing a significantly simpler style that's unlikely to make use of the latest and greatest renderer features

I'll put my 💭  on these problems and see what I can come up with. The first 2 points could be addressed by a network mocking utility, which we've been talking about for a while.

jfirebaugh:
This particular effort has stalled, but @brunoabinader has been working on this separately, see e.g. #5055.

PR #4503, opened 2017-03-27
Prevent infinite loop in binarySearchForIndex when duplicate stops are present

vicapow:
Between 0.32.1 and 0.33.0 the `binarySearchForIndex` function [fixed a bug](https://github.com/mapbox/mapbox-gl-function/pull/35) where the index wasn't properly being set. However, when a style has multiple of the same stops, for example: `[[9, 10], [17, 11], [17, 11], [18, 13]]`, the `binarySearchForIndex` function will loop forever eating resources and will prevent  `map.loaded()` from ever returning `true`.


vicapow:
for context, see: https://github.com/mapbox/mapbox-gl-js/pull/4498

mourner:
Great fix!

PR #4518, opened 2017-03-29
Clamp negative heights to zero in extrusion shaders

lbud:
These changes clamp negative height and base values to zero; part of https://github.com/mapbox/mapbox-gl-native/pull/8431.


lucaswoj:
No pit mines @peterqliu ⚒

PR #4523, opened 2017-03-30
Fix dead "zoom and property function" links.

ChrisLoer:
I got a little lost trying to find these docs because the 'types-function-zoom-property' anchor didn't exist.

/cc @anandthakker @lucaswoj


mourner:
Should we target `mb-pages` instead so that this is effective immediately?

PR #4524, opened 2017-03-30
Fix dead "zoom and property function" links.

ChrisLoer:
Porting fix straight to `mb-pages`

/cc @mourner


mourner:
Thanks! BTW you could force-push this to the previous branch to avoid creating a new PR.

ChrisLoer:
Ah, of course -- I was thinking I should be able to do that somehow but it didn't occur to me to use a force push.

PR #4528, opened 2017-04-01
fix boxZoom bug

### LINKED ISSUE #2237: Disabling and enabling the box zoom interaction messes it up

mollymerp:
fix #2237 where disabling then re-enabling the boxZoomHandler would result in a bug where the drag handler would be erroneously active after a boxZoom triggered `mouseup` event

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/handler/drag_pan.js
+++ b/src/ui/handler/drag_pan.js
@@ -190,6 +190,10 @@ class DragPanHandler {
     _ignoreEvent(e) {
         const map = this._map;
 
+        // if boxZoom is disabled and then re-enabled, the drag_pan event listener for mousedown will fire before
+        // the boxZoom event listener, so map.boxZoom.isActive() will return false resulting in an erroneous
+        // drag on mousemove once the boxZoom is released, so we need this additional check.
+        if (map.boxZoom && e.shiftKey && e.button === 0) return true;
         if (map.boxZoom && map.boxZoom.isActive()) return true;
         if (map.dragRotate && map.dragRotate.isActive()) return true;
         if (e.touches) {
--- a/src/ui/handler/drag_pan.js
+++ b/src/ui/handler/drag_pan.js
@@ -190,10 +190,6 @@ class DragPanHandler {
     _ignoreEvent(e) {
         const map = this._map;
 
-        // if boxZoom is disabled and then re-enabled, the drag_pan event listener for mousedown will fire before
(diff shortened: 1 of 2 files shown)


src/ui/handler/drag_pan.js
```diff
@@ -190,6 +190,10 @@ class DragPanHandler {
     _ignoreEvent(e) {
         const map = this._map;
 
+        // if boxZoom is disabled and then re-enabled, the drag_pan event listener for mousedown will fire before
+        // the boxZoom event listener, so map.boxZoom.isActive() will return false resulting in an erroneous
+        // drag on mousemove once the boxZoom is released, so we need this additional check.
+        if (map.boxZoom && e.shiftKey && e.button === 0) return true;
```
anandthakker: Should this check also include `max.boxZoom.isActive()`?

mollymerp: the reason for this check is because `map.boxZoom.isActive()` returns `false` if BoxZoomHandler is disabled / re-enabled, so, no 🙃

mollymerp: I think I need to re-word this comment to make that more clear. its due to the firing order of event listeners attached to the `keydown` event.

anandthakker: Oh dear 🤦‍♂️.  I swear I really did read that comment , but then got turned upside-down thinking about event logic...

What made me ask was thinking about something like:
 - mousdown: start dragpan
 - press (and hold) shift
 - mouseup: i think the expectation would be to stop panning, but this check would  prevent that from happening (if I'm reading it right?)

mollymerp: you're right this is no bueno! thanks for catching. I'll rethink how to handle this 💭

mollymerp:
Yep! Thanks @mourner

PR #4530, opened 2017-04-02
Android SDK v5.0.0, iOS SDK v3.5.0, macOS SDK v0.4.0

1ec5:
This PR updates the style specification JSON and documentation to reflect the fact that the native SDKs now support data-driven styling as of Android SDK v5.0.0, iOS SDK v3.5.0, and macOS SDK v0.4.0.

The Android SDK v5.0.0 and iOS SDK v3.5.0 changes were previously pushed to the mb-pages branch in #4469. I’ve cherry-picked that change here so that the copy of v8.json in master can remain the single source of truth. This is important because the [mapbox-gl-style-spec](https://www.npmjs.com/package/mapbox-gl-style-spec) package would be obtained from master, not mb-pages. Going forward, changes to the SDK support tables regarding the native SDKs should be merged to master first, then cherry-picked to mb-pages (so that the updates don’t have to wait until to the next GL JS release).

/cc @mapbox/android @mapbox/ios

### DIFF
--- a/docs/style-spec/_generate/index.html
+++ b/docs/style-spec/_generate/index.html
@@ -889,50 +889,50 @@ navigation:
             <tr>
               <td><code>property</code></td>
               <td class='center'>&gt;= 0.18.0</td>
-              <td class='center'>Not yet supported</td>
-              <td class='center'>Not yet supported</td>
+              <td class='center'>&gt;= 5.0.0</td>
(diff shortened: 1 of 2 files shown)


1ec5:
#4536 cherry-picks this change into mb-pages.

PR #4540, opened 2017-04-04
Fix easeTo interpolation on pitched maps

### LINKED ISSUE #4178: easeTo goes in the wrong direction

mourner:
A PR in place of #4320. Should close #4178 and #3112.

Changes `easeTo` implementation to have more robust interpolation, fixing weird behavior on pitched maps, making it more consistent with `flyTo`, and makes the definition of `offset` less ambiguous.

Haven't updated expected values in the camera tests, but otherwise this is ready for review.

cc @joswinter

## Launch Checklist

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -501,7 +501,6 @@ class Camera extends Evented {
         }
 
         const tr = this.transform,
-            offset = Point.convert(options.offset),
             startZoom = this.getZoom(),
             startBearing = this.getBearing(),
             startPitch = this.getPitch(),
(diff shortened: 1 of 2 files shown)


mourner:
This turned out to be more complicated than I thought. We can't simply interpolate linearly through projected center because e.g. as you zoom in to a non-centered point, the scale increases exponentially while map center interpolates linearly, leading to a curved animation.

mourner:
OK, I think I finally got this right, with proper offset handling, and the tests pass. The easing while zooming interpolation formula is still probably not 100% scientifically correct, but it works well in practice, and me and @anandthakker couldn't come up with anything better.

I'd like to improve the easing code further as there is still some logic duplication going on between `easeTo` and `flyTo`, but I'd rather do that in a follow-up PR.

mourner:
@1ec5 should this be ported to native? It changes the way `easeTo` animations are calculated quite substantially.

1ec5:
The native implementation diverged from the GL JS implementation shortly after I ported it, but we probably do have to revisit the native implementation in light of mapbox/mapbox-gl-native#5292.

Issue #4376, opened 2017-03-07
Map pitch not accounted for in text orientation


nickidlugash:
From @ajashton https://github.com/mapbox/mapbox-gl-js/issues/3057#issuecomment-278766953:

> Noting that one specific aspect of this is that perspective is not accounted for in text orientation. If you look at [this JSFiddle](https://jsfiddle.net/ajashton/25peL7y2/1/) First, 2nd, & 3rd Streets should be rotated 180° to be closer to "upright", but if you change the pitch to 0 the same orientation is correct.

nickidlugash:
@ChrisLoer We haven't yet discussed this issue, but this may (or may not) be something worth thinking about in conjunction with other pitched label rendering refactors. 

Right now if text is kept upright, we [calculate which orientation of glyphs we should *not* add to symbol arrays](https://github.com/mapbox/mapbox-gl-js/blob/master/src/data/bucket/symbol_bucket.js#L576-L578) based on the angle of the glyph, but this doesn't account for how the final angle may change based on pitch.

I'm not sure how vertical text is affected – this issue is not as evident in vertical-enabled line labels, but that may just be because there are four possible orientations so the flip between orientations is more subtle and an "incorrect" orientation is not as suboptimal.

ChrisLoer:
Hmm, interesting -- yeah, this is something we could tackle at the same time.

This is another case where doing the fully correct solution at high speed requires sending a lot more information down to the GPU. Consider these two views of 15 St NW:

_Panned to lower left_
![screenshot 2017-03-27 12 15 43](https://cloud.githubusercontent.com/assets/375121/24373734/24b4e36a-12e7-11e7-90de-f5ee553f9267.png)

_Panned to lower right_
![screenshot 2017-03-27 12 15 31](https://cloud.githubusercontent.com/assets/375121/24373735/24bbc360-12e7-11e7-829f-f94b2b00a70f.png)

Both views came from a map at the same pitch/angle/zoom: the only difference was panning left and right.

We could calculate the right glyph to show in the SDF vertex shader, but we'd have to send down two copies of each glyph (one "upside down" copy). Alternatively, we can try to do this calculation at placement time and re-trigger placement whenever there's a pan operation. It would take a little longer for labels to flip to the right orientation, and it would add CPU intensity.

lucaswoj:
[...]

Issue #4165, opened 2017-02-01
Rename circle-pitch-scale to circle-pitch-scaling


lucaswoj:
_From @1ec5 on January 5, 2017 10:35_

`circle-pitch-scale` should be renamed `circle-scale-alignment`.

https://github.com/mapbox/mapbox-gl-style-spec/issues/459#issuecomment-224957055 proposed `{circle,icon,text}-pitch-scale` properties as complements to the `{circle,icon,text}-{rotation,pitch}-alignment` properties. As discussed in mapbox/mapbox-gl-native#6098, the `circle-pitch-scale` name incorrectly implies a numeric value. It’s also inconsistent with the rest of the properties:

Property |  | a feature’s | according to the map or viewport’s | when the map has
----|----|----|----|----
`circle-rotation-alignment` | **aligns** | **rotation** | **rotation** | rotation
`circle-pitch-alignment` | **aligns** | **pitch** | **pitch** | pitch
`circle-pitch-scale` | aligns | **scale** | **scale** | **pitch**

Renaming `circle-pitch-scale` to `circle-scale-alignment` would make the property’s type more obvious and better harmonize the name with the others:

Property | does what | a feature’s | according to the map or viewport’s | when
----|----|----|----|----
`circle-scale-alignment` | **aligns** | **scale** | **scale** | pitch

/cc @jfirebaugh @lucaswoj

_Copied from original issue: mapbox/mapbox-gl-style-spec#645_

lucaswoj:
_From @1ec5 on January 5, 2017 10:37_

An alternative to `*-scale-alignment` would be `*-size-alignment`, which would be consistent with `icon-size` and `text-size`.

jfirebaugh:
I don't agree with this proposal.

* `circle-rotation-alignment` -- there is no such property. Circles have radial symmetry and no rotation properties.
* `circle-scale-alignment` doesn't contain the word "pitch". The property specifies the scaling behavior of circles when the map pitches; it should have both "pitch" and "scale" in its name.

1ec5:
> `circle-rotation-alignment` -- there is no such property.

Sorry, I picked a poor example there. Here’s a revised table:

Property |  | a feature’s | according to the map or viewport’s | when the map is
----|----|----|----|----
`icon-rotation-alignment` | **aligns** | **rotation** | **rotation** | rotated
`icon-pitch-alignment` | **aligns** | **pitch** | **pitch** | tilted
[...]

Issue #4120, opened 2017-02-01
Complete set of *-rotation-alignment, *-pitch-alignment, and *-pitch-scale properties


lucaswoj:
![rotation-pitch-alignment](https://cloud.githubusercontent.com/assets/83384/15935723/4a942190-2e36-11e6-94ab-eba36b62e1b7.png)

Full set of properties:
- [x] `icon-rotation-alignment`
- [x] `icon-pitch-alignment` -- #4869 / https://github.com/mapbox/mapbox-gl-native/issues/9345
- [ ] ~~`icon-pitch-scale`~~ Removed in favor of hardwired 50% scaling factor
- [x] `text-rotation-alignment`
- [x] `text-pitch-alignment`
- [ ] ~~`text-pitch-scale`~~ Removed in favor of hardwired 50% scaling factor
- [x] `circle-pitch-alignment` -- #2420 / #4871 / https://github.com/mapbox/mapbox-gl-native/issues/9349
- [x] `circle-pitch-scale` -- https://github.com/mapbox/mapbox-gl-js/pull/2821 / https://github.com/mapbox/mapbox-gl-native/pull/5576

-----

_From @lucaswoj on June 9, 2016 15:28_

The style spec has the concept of `rotation-alignment` which specifies how features are placed with respect to map rotation. 
- `rotation-alignment: viewport` indicates that features should be rotated with respect to the viewport, drawn such that the top of the screen is "up". 
- `rotation-alignment: map` indicates that features should be rotated with respect to the map, drawn such that north is "up"

I propose that we add the concept of `pitch-alignment` which specifies how features are placed with respect to map pitch.
- `pitch-alignment: viewport` indicates that features should be pitched with respect to the viewport, drawn on plane of the viewport. 
- `pitch-alignment: map` indicates that features should be pitched with respect to the map, drawn on plane of the map.

cc @yhahn @mourner @tmcw @ansis @jfirebaugh @lbud 

_Copied from original issue: mapbox/mapbox-gl-style-spec#459_

lucaswoj:
_From @yhahn on June 9, 2016 15:40_

![rotation-pitch-alignment](https://cloud.githubusercontent.com/assets/83384/15935723/4a942190-2e36-11e6-94ab-eba36b62e1b7.png)

:+1: this makes total sense to me. It's worth noting that atm our renderers don't support this fully but it's something we can work toward.

### master gl js/native

| rotation | pitch | text support | icon support |
[...]

Issue #2321, opened 2016-03-23
fix queryRenderedFeatures box across dateline


ansis:
**mapbox-gl-js version**: 8b52b55251b4d63bc6713d76c8314c1f1007f04a

### Steps to Trigger Behavior

 1. query a full screen box at zoom 0

### Expected Behavior

Returns all features.

### Actual Behavior

Returns a subset of features based on the wrapped coordinates of the box corners.

ansis:
If you are zoomed out and there are multiple copies of a polygon, should queryRenderedFeatures return a feature for every instance visible on the screen? Or should it return only one feature?
<img width="943" alt="screen shot 2016-03-23 at 1 56 34 pm" src="https://cloud.githubusercontent.com/assets/1421652/14000605/1fc555fc-f0ff-11e5-974a-3f6c433cfc07.png">

mcwhittemore:
@ansis I'd expect it to only return one. It is returning the features not the renderings. If you update one of them the other will move.

ansis:
@mcwhittemore great. It'll return one

This is fixed by https://github.com/mapbox/mapbox-gl-js/commit/c418867f2ff419921dc8619a61b9b09a4d7efd95

mcwhittemore:
@ansis :+1: Should we close this?

PR #4547, opened 2017-04-05
Attenuate label size scaling with distance

ChrisLoer:
The motivation for these changes is that we want to improve our rendering support for the "navigation" use case, in which:

- Tilting the map is important, because in turn by turn directions you want to be able to see ahead
- Legibility becomes more important relative to information density or aesthetics, because users of the map need to be able to read labels with a quick glance.

This PR improves support for this case by:

- Changing label scaling behavior to halve the effect of perspective on label size. This preserves the "3d" feel of pitched maps while preventing near labels from becoming comically large and preventing distant labels from becoming impossibly small.
- ~~Giving style designers the ability to control how much labels scale down with distance from the camera. This addresses two of the remaining items in #4120 (`text-pitch-scale` and `icon-pitch-scale`).~~

Changes we make to placement logic to support collision detection over a wider range of zooms may make it easier to address #4376.

- [x] ~~Control how icons and text scale based on distance from camera using `text-pitch-scale` and `icon-pitch-scale`~~ Halve the effect of perspective on label scaling for both icons and text.
- [x] Approximate collision boxes based on distance from camera
- [x] ~~Make collision boxes conservative over the entire range of `text-pitch-scale` values in use~~ No longer necessary with single hardwired value
- [x] Re-run collision detection based on tile "effective zoom" during pan operations. Stop doing collision detection over such a wide zoom range (this is a performance requirement)
- [x] Prevent line labels with `pitch-alignment: viewport` from showing in the distant field at high pitches (the approximations used to draw the labels don't hold up well)
- [x] Performance benchmarking
- [x] Figure out how to handle going over 8 vertex attribute limit (the plan is to selectively bind 'fill_color' and 'halo_color' on gl-native in order to save one attribute)
- [x] ~~Rename `text-pitch-scale` based on https://github.com/mapbox/mapbox-gl-js/issues/4165~~
- [x] Add/update render tests
- [ ] [Port to gl-native](https://github.com/mapbox/mapbox-gl-native/pull/9009)
- [x] Implement collision debug boxes for new logic
- [x] ~~Decide on collision debug colors (since these show up in Studio, we want them to look decent... but also ideally be legible by colorblind folk)~~ Keep original colors.
- [x] Calculate collision tiles separately for wrapped copies of the same data tile


ChrisLoer:
@nickidlugash @anandthakker helped me draft up some math that could help us avoid uneven spacing on highly curved labels while still doing all the glyph position calculations in the shader:

https://gist.github.com/ChrisLoer/fb8c22845b6c3216afff0f19dc53f6d6

The math depends on adding an(other) approximation that will have the biggest effect on labels that are close to horizontal. I attached some screenshots to a comment on the gist to show the expected difference -- doesn't look like a big deal to me, but could use your input.

ChrisLoer:
@mourner @kkaefer I'm going to increasingly elaborate lengths to do complicated glyph positioning in the vertex shader. Tangram [looks like it does glyph positioning in the CPU at render time](https://github.com/tangrams/tangram-es/blob/6bde54513304a9e915c51decf941be10dfdc2e09/core/src/tangram.cpp#L383) and seems to get acceptable performance. Doing positioning that way also allows them to send smaller vertex arrays to the GPU, because they don't have to send copies of glyphs for each possible line segment they could be on.

Do either of you have any advice on how to think about the performance trade-offs of the two architectures here? I don't have strong enough intuitions to guess at the differences very well, and since the architectural approaches are so different, it's not easy to just find out through experiment...

nickidlugash:
@ChrisLoer Awesome to see the improvements and to do's on this PR!

> The math depends on adding an(other) approximation that will have the biggest effect on labels that are close to horizontal. I attached some screenshots to a comment on the gist to show the expected difference -- doesn't look like a big deal to me, but could use your input.

It's hard to say without being able to interact with it and experiencing the rate of change, but the `sin^2(angle)` version actually looks more correct to me...it seems slightly too condensed, but the "correct" `sin(angle)` version actually looks noticeably too spaced out. I noticed this earlier while playing around with this current branch – perfectly horizontal and vertical labels look correctly spaced, but all angles in between seem to be getting too much expansion. The more exaggerated easing for legibility expansion with the `sin^2(angle)` version seems to be improving this 🤔  Is it possible that it's compensating for an inaccuracy somewhere else in the legibility expansion calculation?

ChrisLoer:
> Is it possible that it's compensating for an inaccuracy somewhere else in the legibility expansion calculation?

Yes, the same thought occurred to me! It's definitely possible considering all the different approximations that are involved, although right now if that's actually what's happening it's just dumb luck. I don't have a theoretical explanation -- but with enough time staring at this, it should be possible to construct one. 😅

ChrisLoer:
BTW, if you want to play with the behavior of `sin^2(angle)` while looking at straight line labels, it's easy to swap into the vertex shader. Just replace `legibility_expansion += abs(sin(projected_label_angle)) * (incidence_stretch - 1.0)` with `legibility_expansion += pow(sin(projected_label_angle), 2.0) * (incidence_stretch - 1.0)`.

nickidlugash:
@ChrisLoer Another unrelated issue I noticed: the angle of the rendered glyphs for vertical labels seems off, the further away they are from vertical center: 

tilted_vertical_glyphs

They look like they are on tilted planes. Is this related to how the incident angle does not take into account `gl_Position.x`?

(This is an issue too in the current released implementation – it's just not as obvious with the current foreshortening behavior).

ChrisLoer:
I think that's our brain applying perspective to the labels based on what's around them, and interpreting the input as "tilted plane":

screenshot 2017-04-13 16 01 02

If we zoom in on the lower left corner, look at Wilshire Blvd, then use photoshop to rotate the street back to vertical, we can see the glyphs are lined up parallel to the road and "flat".

wilshire perspective

I think the effect is kind of cool because it makes the map feel a little more three-dimensional. Other mapping products<sup>™️</sup> seem to do the same thing.

ChrisLoer:
I got an initial implementation of the fancy trigonometry for label expansion... kind of working. It looks like it's doing a better job of keeping spacing even on curved labels, but there are two big problems:

* There's lots of glyph flicker/duplication around segment junctions. This might be caused by the placement-time calculations working with higher precision angles than the part of the calculation done in the shader (the positioning calculations only have to disagree by a tiny amount between two segments to cause artifacts). It looks bad enough it might just be a matter of me messing up the math.
* This approach ties the `minZoom` and `maxZoom` for a glyph to the "legibility expansion" calculated for that glyph... which breaks the `labelMinScale` trick we use at https://github.com/mapbox/mapbox-gl-js/blob/master/src/symbol/quads.js#L218 to make all the glyphs disappear if one of the glyphs has to disappear. To fix this, we'd need to add even more vertex attributes. 😞 

Overall, it's feeling like too complicated a solution, but I'll keep poking at it a bit longer to see if I can come up with anything compelling.

Here's a shot showing the algorithm actually doing its thing. This is the kind of label that does poorly with the simple algorithm: the center of the label is at about 45 degrees, while the left side is near vertical and the right side is near horizontal. As a result, the simple algorithm (on the left) squishes the left side and stretches out the right side. The fancy algorithm appears to do a reasonable job of spacing across the length of the label.

vistagrande improvements

kkaefer:
> Do either of you have any advice on how to think about the performance trade-offs of the two architectures here? I don't have strong enough intuitions to guess at the differences very well, and since the architectural approaches are so different, it's not easy to just find out through experiment...

We have never tried this out, but as a data point: We moved label placement on rotation from pre-calculcated to on-demand (CPU) and haven't run into any performance issues.

ChrisLoer:
> We moved label placement on rotation from pre-calculated to on-demand (CPU) and haven't run into any performance issues.

Are you talking about how we re-do collision tiles on rotation? Although we do it on demand with the CPU, we also do it asynchronously and coalesce multiple frames into one calculation. That works for collision detection which is just on/off, but for glyph placement to be smooth we would need to synchronously calculate positions on every frame.

ChrisLoer:
Per discussion with @nickidlugash, we think the navigation use case can benefit from the `text-pitch-scale` changes alone, so I'm shelving the curved label legibility work for now. This will leave us in a bit of a weird place: `pitch-alignment: viewport` curved labels are _already_ broken for labels in the distance, but the breakage will be exacerbated if combined with `text-pitch-scale < 1.0`.

kkaefer:
> That works for collision detection which is just on/off, but for glyph placement to be smooth we would need to synchronously calculate positions on every frame.

Yeah, but we do it asynchronously because the placement is the slow part, not the uploading part. If we can calculate the placement fast enough, a synchronous upload might be possible as well, but, as all things GL, we'll have to try them out to know how they perform.

ChrisLoer:
>  This will leave us in a bit of a weird place: pitch-alignment: viewport curved labels are already broken for labels in the distance, but the breakage will be exacerbated if combined with text-pitch-scale < 1.0.

Here's an example of how broken a curved label can get when it's in the far distance (with `text-pitch-scale: 0`):

screenshot 2017-04-26 11 35 27

Maybe a hardwired `max-distance` property for these labels could work as a stopgap.

ChrisLoer:
I implemented the logic that re-calculates collision tiles as a tile's distance from the camera changes, so that we can avoid the performance cost of calculating collision tiles over giant zoom ranges.

Playing with the changes, one concern I have is that re-doing the collision logic frequently makes labels feel jumpy: a relatively small change in position can cause a cascade of collision changes. Throttling how often we re-do placement (e.g. only after the map has panned at least half a tile) might reduce this effect a little, but a lot of it is just the result of the way we're adjusting collision boxes based on perspective at render time.

collision instability

ChrisLoer:
I implemented a `max-camera-distance` property that allows style designers selective control over which labels show in the distance. The unit for the property is "multiples of the distance from the camera to the center of the map"... which is kind of tricky to wrap your head around, but gives clearly defined behavior across different zooms and pitches.

I imagine that if we were to make a navigation style using just the functionality from this PR, we would use the `max-camera-distance` to hide `symbol-placement: line` labels whenever they're very far from the camera (maybe give them a `max-camera-distance` around 1.3-1.5..)

mb12:
@ChrisLoer Is it possible to explain the distance whose multiple is taken in "max-camera-distance" via a diagram? (Whiteboard photo etc should be good enough).

ChrisLoer:
I'm not great at hand-drawing 3d stuff, but hopefully this diagram captures the gist of it:

camera-to-center

ChrisLoer:
I moved to only re-doing collision detection every 300 ms while the map is panning. I think it helps with the feeling that labels are flickering all over the place, but overall label placement still doesn't feel super smooth or predictable.

There's a subtle problem with the approach I'm using. On a flat map, we calculate collisions for a tile between a range of `minScale: 0.5` and `maxScale: 2`. Now, if we have a distant tilted tile with `text-pitch-scaling: 0`, we might have labels be 4 times bigger, so we calculate collisions with the range `minScale: 0.125` and `maxScale: .5`.

Now consider a city label and a a highway shield placed near each other. The city label's minimum placement scale is `1.0`. The highway shield has lower precedence, so it can only show when the map has higher scale: say it shows at scale `1.25`. The problem occurs when our `maxScale` goes below `1.0` -- all of a sudden the city label is removed from the collision grid (to constrain the cost of collision detection), but since the highway shield is small and takes a lot of scaling o collide with whatever knocked the city label out, it now gets a much lower minimum placement scale (say `.25`).

The result of all this is that as you pan a titled map tile into the distance, you could see:

* First: highway shield + city label
* Second: city label
* Third: highway shield

In terms of avoiding collisions and getting good label density, this is fine. But it breaks the desideratum that once a label appears it should stick around as scale increases -- and when it happens across a whole bunch of small labels and highway shields, it contributes to that "flickery" feel.

Note that we already have the same problem when we cross integer zoom levels, it's just much less noticeable.

ChrisLoer:
Some problems to tack onto the list:

- Labels with low pitch scaling that are also near tile boundaries can collide in the distance. My guess is that this is tolerable, and I don't think we have a very good solution short of merging vector tiles before rendering.
- `queryRenderedFeatures` will have some additional inaccuracies because the "shape" of collision boxes isn't fully determined until render time. While it would be possible to mimic the render-time calculations when the query happens, I'm hoping we can get away with ignoring them. Even without the pitch-scaling changes, the collisions boxes aren't always shaped correctly on pitched maps, and we seem to be able to live with it.

ChrisLoer:
`293cd08` fixes a bug that was contributing a lot of unnecessary label flicker (I was setting draw uniforms in the wrong place, so that they could be unpredictably shared between tiles).

While debugging that issue, I noticed another design issue that we might want to think about. Because the collision detection is based off the "least scaled" item in a tile, label density can change a lot between a tile that has an item with low scaling vs a tile without any items that have low scaling.

Look at the tiles that have Highway 2 (Santa Monica Blvd) in them, vs. the more distant tiles that don't have any highway shields.

screenshot 2017-05-01 11 56 20

Because my test style gives these highway shields low scaling (i.e. bigger size relative to the map when they're in the distance), the tiles that include the highway shields get pretty empty (following the conservative algorithm). However, the tiles above without any shields get denser label placement even though they're further away.

This could just be something style designers need to be aware of, or we could change the code to share the same conservative collision scaling factor across all visible tiles (this would trade lower overall label density in exchange for more uniformity of label density).

ChrisLoer:
If you follow the `San Clemente` label collision box closely in this animation, you can see this PR working as intended, but also see some of the drawbacks:

* The shape of the collision box jumps around as it moves into the distance. This is because changes to the aspect ratio can only be calculated at placement time, which aren't happening on every frame.
* The collision box is drawn on four different tiles (although the label only does collision detection on its "own" tile, it also shows up as a potential collision source for labels on the other tiles). The aspect ratio of the collision box is not entirely consistent between the four tiles. This is because at placement time we approximate the aspect ratio based on the distance of the *center* of the tile from the camera. This approximation can get worse if tiles are really big relative to the viewport.

collision box dance

We actually have all the information necessary to make aspect ratios consistent across tiles (given the tile position, we could project the position of every single label at placement time). I don't know how big the performance impact would be, but it's a lot of matrix math that would be nice to leave to the GPU if we can get away with it.

ChrisLoer:
Playing around with the map generally feels performant to me at this point. Here's the specific performance benchmarking strategy I used:

* Build minified `cloer_text_pitch_scale`
* Load http://localhost:9966/debug#10.46/33.9951/-118.3343/-36/60
* Turn on profiling for Main
* Change zoom to 14
* Collect data
* Reload page at zoom 10
* Turn on profiling for first web worker (the point of the weird zoom changes is just to let me set up profiling for web workers in a more reproducible manner)
* Change zoom to 14
* Collect data
* Repeat with `mapbox-gl.js` downloaded from https://api.mapbox.com/mapbox-gl-js/v0.37.0/mapbox-gl.js

Looking through the profiles, I don't see any really big deviations. Overall program time is within 10%. At least in these test runs, the total time spent in `CollisionTile::placeCollisionFeature` does seem to have increased somewhat as might be expected (in two runs on the branch I saw total time of 20-40 ms, vs 15-17ms on `v0.37.0`). I have the profiles saved locally in case we want to look at any other parameters.

As expected, the maps actually look somewhat different (even though I'm not using a style with `text-pitch-scaling` enabled). This is because the more accurate "yStretch" calculations in `CollisionTile` allow for somewhat denser near label placement and somewhat sparser placement in the distance.

nickidlugash:
@ChrisLoer To recap what we discussed as a tentative proposal last friday:

### `-pitch-scaling`
 
Implement as style spec properties. Default value is `auto`, based on `symbol-placement`, `-rotation-alignment`, or other properties.
 
TODO: figure out what this auto default should be. e.g. `symbol-placement: point` --> `-pitch-scaling: 0.5` / `symbol-placement: line` --> `-pitch-scaling: 1`

### `max-camera-distance`

Hard coded implementation, no style spec change. Possible variables:

- pitch: no `max-camera-distance` below a certain pitch value
- `symbol-placement`: line placement gets `max-camera-distance` but point placement doesn't?
- `-pitch-scaling`

TODO: figure out variables and values

---

After playing around with variables and values this week, here's my thoughts so far:

### `-pitch-scaling`

I actually now think we should first try implementing a hard coded value for this, rather than exposing it as style spec properties, because:

1. I think this is the more conservative approach – improve the default rendering behavior as much as possible, and only if users or our team can identify clear use cases where we need more flexibility, we can then add style spec properties. If we do add properties later, we can make our hard coded value the default.

2. I think a value of `0.5` actually works pretty well across the board – for both symbol placements, both text and icons, at all pitches:

pitch-scaling-0-5
pitch-scaling-0-5-lines

The only situation where I'm a bit hesitant about this value is when
`layer.layout['symbol-placement'] === 'line' && rotateWithMap && layer.layout['icon-image']`

Because there is a possibility that a user may want icons styled that way to very closely match the pitched geometry. We could set the default in those situations to be `1.0` instead of `0.5`, although the predictability of having `0.5` across the board might be more overall beneficial. The only practical example we have in our core styles of this type of styling is oneway arrows along roads, but these icons are so small that `0.5` actually works better in this case (`1.0` makes them pretty illegible in a lot of cases). Icons are also generally problematic at `1.0` because scaling up the svgs so much in the lower half of the viewport makes them blurry.

### `max-camera-distance`

If we take the conservative approach that the sole purpose for this behavior is to hide potentially ugly line labels until we can merge in [better pitched label legibility](https://github.com/mapbox/mapbox-gl-js/pull/4682) (as opposed to using this behavior to reduce density generally at further out distances), then I think it might make sense to limit `max-camera-distance` to only situations where ugly line labels may occur. I sketched out something like this in `draw_symbol`:
 
```
if (tr.pitch > 30 && layer.layout['symbol-placement'] === 'line' 
        && layer.layout['text-rotation-alignment'] === 'map' 
        && layer.layout['text-pitch-alignment'] === 'viewport' 
        && layer.layout['text-field']) {
        gl.uniform1f(program.u_max_camera_distance, ((tr.pitch - 30) / (60 - 30) * (1.0 - 1.75) + 1.75));
    } else {
        gl.uniform1f(program.u_max_camera_distance, 10);
    }
```
And then tried to combine this with some logic in the shader to only apply `max-camera-distance` to very foreshortened labels:

```
if (foreshortening < 0.7) {
        gl_Position.z +=
            step(u_max_camera_distance * u_camera_to_center_distance, camera_to_anchor_distance) * gl_Position.w;
        }
```
^ but that didn't work because that caused only very foreshortened glyphs to be hidden, not the whole label. If there's a way to target whole labels, I think something like this could make sense (although maybe there isn't, otherwise we could just use the foreshortening value by itself to hide certain labels?) If there isn't a way to do this, then I think mapping pitches 30–60 to `max-camera-distance` values of about 1.75–1.5 could maybe work.

I haven't yet tested any of these values on styles other than Mapbox Streets, but I can do that tomorrow.

ChrisLoer:
@nickidlugash If we really can get away with a single hard-wired pitch-scaling factor, that is of course very easy to implement. As we discussed briefly before, if we get all the way down to a single pitch-scaling factor, we could also throw away all the "minimum/maximum pitch scaling" logic -- I doubt it would make a measurable difference in performance, but at least we could delete some code.

I like where you're going with making `max-camera-distance` hardwired and making it only apply to viewport-pitch-aligned line labels. Whereas I'm still a little uncomfortable with the backwards-compatibility implications of hardwiring `text-pitch-scaling`, I'm pretty convinced this should be hardwired. I think we should be able to find a single value there that works pretty well -- no need to change it based on pitch, because the distance to the furthest point of the screen is already scaling down based on pitch. It looks like you were thinking of scaling between 1 at pitch 60 (in which case only the bottom half of the screen would have line labels) and 1.75 at pitch 30 (which would be way above the top of the viewport for any reasonable viewport size). What happens if we just use a value of 1.5 across all pitch ranges? I've been playing with "Broad Branch Rd NW" at various pitches (`http://localhost:9966/debug/debug.html#14.31/38.9301/-77.0437/-8.1/55`) to get a feel for which parameters work. My viewport is ~800px tall and a value of 1.5 starts filtering things at the very top of the viewport around 45 degrees of pitch.

Within the vertex shaders, we don't have a way of knowing the "most foreshortened" glyph in the label because we don't have information on any of the other glyphs. If our concern is just preventing the "most broken" labels from showing, foreshortening probably isn't our primary concern. Highly foreshortened straight labels are illegible, but don't look "broken". The bigger problem is highly _curved_ labels (where the segments stop lining up with each other very well in the distance). We could try to pass a "curvature" parameter into the vertex shader to help with the filtering, but I'm hoping we can settle on just one constant.

ChrisLoer:
I hard-wired ".5" for `-pitch-scaling`, hardwired "1.5" for the `max-camera-distance` for viewport-pitch-aligned line labels, and removed all the code that supported variable values.

The 50 commits show some history of me stumbling towards a solution here -- before pushing to `master` I want to collapse them into a smaller/more intelligible set, but I thought I'd leave them up for now to make the evolution a little more traceable.

src/geo/transform.js
```diff
@@ -380,7 +380,12 @@ class Transform {
      * @param {TileCoord} tileCoord
      * @param {number} maxZoom maximum source zoom to account for overscaling
      */
+
     calculatePosMatrix(tileCoord: TileCoord, maxZoom: number) {
+        return new Float32Array(this._calculatePosMatrix(tileCoord, maxZoom));
+    }
+
+    _calculatePosMatrix(tileCoord: TileCoord, maxZoom: number) {
```
mourner: What's happening here? It seems like the no-underscore version essentially copies the matrix twice (note the ` new Float32Array` on line 399).

ChrisLoer: 😳 That's vestigial from earlier attempts to figure out the right way to calculate tile distance, when I thought a separate helper function was necessary. I just removed the helper, and `cameraToTileDistance` now calls `calculatePosMatrix` directly.

nickidlugash:
@ChrisLoer I think the hardwired values look great! 👍 

Regarding any increased label instability due to the approximations of collision box size/shape, I think that at pitch <=45, it's fairly minimal (there is noticeable instability, but there is also instability with the current behavior). I think this is something that would be great to improve, but think that the improvements with label scaling is great enough that its worth it to just release this as soon as possible, and then potentially make updates in a follow up PR. 

I haven't looked closely at how the collision boxes are calculated, but one thing I did notice is that it seems like all horizontal symbols (regardless of position in the viewport) all have collision boxes that seem taller than necessary. I'm not sure if that's just a necessary byproduct of how they are approximated, or whether that could be adjusted. Here's a screenshot with `-padding: 0` for all symbols (as opposed to the default, `-padding :2`:

screen shot 2017-05-12 at 4 00 22 pm

ChrisLoer:
> there is noticeable instability, but there is also instability with the current behavior

I think we're on the same page here, but just to be clear: with the current behavior on `master`, panning a pitched map won't affect collision logic at all, so there's no "instability" in that specific case. If you're just panning a pitched map (without rotation) before and after this change, you're going to see a lot more label flickering after the change. On the other hand, on top of distant labels being too small to read on `master`, they also end up colliding with each other because the "yStretch" approximation breaks down in the distance. I agree that at <= 45 degrees it's not too dramatic.

> all have collision boxes that seem taller than necessary

I may be missing which label you're talking about in the screenshot, but my guessplanation for what you're seeing is primarily two things:

* Even with `-padding: 0`, the height of a label's collision box isn't calculated based on the top of the highest glyph -- it's based on a multiple of the `text-line-height` property. So if the line height is greater than the height of the biggest glyph (it probably is) you will end up with a small amount of "whitespace" above and below the top and bottom glyphs of a label.
* The "yStretch" correction just makes the whole collision box bigger, so the "whitespace" between the top of the glyphs and the top of the collision box will grow proportionally with yStretch (aka at higher pitch there will be more unnecessary padding).

Do you think it's a problem? To me it seems like a pretty modest bit of conservatism. If it is a problem, it shouldn't be too hard to calculate collision box heights based on glyph heights (width is already calculated based on glyph width).

nickidlugash:
> If you're just panning a pitched map (without rotation) before and after this change, you're going to see a lot more label flickering after the change. 

@ChrisLoer Ah, yeah, sorry about my vague comment earlier – I think I was not referring to panning a pitched map, but rather the movement of pitching a map itself. When you _pitch_ a map, there is some flickering in both the current and new behavior. 

Regarding flickering on panning with the changed behavior, this is mainly related to panning in the up/down direction, right? (I had assumed there would be some issues in the right/left direction due to the approximation of collision box shape/size, but I'm not actually seeing much of this). In the up/down direction, there is some noticeable flickering, but I feel like it's not really "unstable", in the sense that if a label disappears as you pan upwards, it doesn't typically reappear again as you continue to pan upwards. If our intention is to have a lower label density on the top part of the viewport for pitched maps (which I think is ideal), then I'm not sure how it would even be possible to _not_ have labels disappear at some point while panning upwards? As we've discussed before, a lot of the issue may simply be that we can't apply a transition when these labels appear/disappear.

>The "yStretch" correction just makes the whole collision box bigger, so the "whitespace" between the top of the glyphs and the top of the collision box will grow proportionally with yStretch (aka at higher pitch there will be more unnecessary padding).

I think I'm mainly concerned about this aspect (whereas I think the increased height due to `text-line-height` values is largely necessary to fully encompass all characters, including descenders on letters like "g"). Since there is a disproportionately large padding on the bottom of collision boxes, I'm worried that the collisions seem less intuitive – i.e. it would be confusing to users as to why a label disappeared when it didn't seem to be colliding with another label. It's probably not a big issue, though. If there's a quick fix now that would be great, if not, maybe this is something we can revisit?

ChrisLoer:
> If there's a quick fix now that would be great, if not, maybe this is something we can revisit?

I can't think of a quick fix -- the stretching is done across all the labels, and the "label height to whitespace" ratio is going to be different between labels. I'd vote for re-visiting this later if it turns out to be a problem.

nickidlugash:
>  I'd vote for re-visiting this later if it turns out to be a problem.

Sounds good to me! 👍

ChrisLoer:
Working on the `gl-native` port made me realize that this PR is also broken for wrapped tile IDs. I need to change `SourceCache` not to use the same tile for two different wrap levels (the collision calculations can't be right for both of them).

src/shaders/symbol_icon.vertex.glsl
```diff
@@ -1,22 +1,25 @@
...
 uniform bool u_is_size_feature_constant;
-uniform mediump float u_size_t; // used to interpolate between zoom stops when size is a composite function
-uniform mediump float u_size; // used when size is both zoom and feature constant
-uniform mediump float u_layout_size; // used when size is feature constant
+uniform highp float u_size_t; // used to interpolate between zoom stops when size is a composite function
+uniform highp float u_size; // used when size is both zoom and feature constant
+uniform highp float u_layout_size; // used when size is feature constant
+uniform highp float u_camera_to_center_distance;
+uniform highp float u_pitch;
+uniform highp float u_collision_y_stretch;
```
ansis: Is highp necessary in all these cases?

src/symbol/collision_feature.js
```diff
@@ -72,42 +72,60 @@ class CollisionFeature {
     _addLineCollisionBoxes(collisionBoxArray, line, anchor, segment, labelLength, boxSize, featureIndex, sourceLayerIndex, bucketIndex) {
         const step = boxSize / 2;
         const nBoxes = Math.floor(labelLength / step);
+        // We calculate line collision boxes out to 150% of what would normally be our
+        // max size, to allow collision detection to work on labels that expand as
+        // they move into the distance
+        const nPitchPaddingBoxes = Math.floor(nBoxes / 4);
```
ansis: Is this where the extra collision calculation cost comes from?

src/ui/map.js
```diff
@@ -57,7 +57,9 @@ const defaultOptions = {
 
     renderWorldCopies: true,
 
-    refreshExpiredTiles: true
+    refreshExpiredTiles: true,
+
+    redoPlacementOnPan: true
```
ansis: Does this need to be an external option?

src/geo/transform.js
```diff
@@ -394,6 +394,17 @@ class Transform {
...
+    /**
+     * Calculate the distance from the center of a tile to the camera
+     * @param {Tile} tile
+     */
+    cameraToTileDistance(tile: Object) {
+        const posMatrix = this.calculatePosMatrix(tile.coord, tile.sourceMaxZoom);
+        const tileCenter = [tile.tileSize / 2, tile.tileSize / 2, 0, 1];
+        vec4.transformMat4(tileCenter, tileCenter, posMatrix);
+        return tileCenter[3];
+    }
```
ChrisLoer: I hesitate to say "pixels" because that's not quite right, but it's distance in view-space coordinates, which is pixel-derived. It's the same units as "cameraToCenterDistance", which is defined as:

`0.5 / Math.tan(this._fov / 2) * this.height` (and `height` is in pixels)

src/render/draw_symbol.js
```diff
@@ -192,6 +194,15 @@ function setSymbolDrawState(program, painter, layer, tileZoom, isText, isSDF, ro
...
         gl.uniform1f(program.u_size, sizeData.layoutSize);
     }
+    gl.uniform1f(program.u_camera_to_center_distance, tr.cameraToCenterDistance);
+    if (layer.layout['symbol-placement'] === 'line' &&
+        layer.layout['text-rotation-alignment'] === 'map' &&
+        layer.layout['text-pitch-alignment'] === 'viewport' &&
+        layer.layout['text-field']) {
+        gl.uniform1f(program.u_max_camera_distance, 1.5);
+    } else {
+        gl.uniform1f(program.u_max_camera_distance, 10);
```
ChrisLoer: This is a ratio, so technically unit-less. But yeah, it's a ratio of "camera to label distance" over "camera to center distance".

The "1.5" value came from experimentation -- basically just looking for a value that would clip out really ugly labels but leave most in place. "10" is effectively infinite (well, at least with our limit of 60 degrees of pitch).

Passing in the uniform as a float is left over from an earlier iteration of the PR where this value was controlled per-property in the style. Now that it's hardwired, it probably makes sense to push the value into the shader and just make the uniform into a boolean (something like `u_is_pitched_line_label`).

src/shaders/symbol_icon.vertex.glsl
ChrisLoer: I don't think it's necessary (although I did most of my testing with highp precision, so maybe lowering the precision requirement would expose some problems). I ended up standardizing on highp because I didn't have any better criteria for choosing precision and having a jumble of lowp/mediump/highp just seemed to make the shader code a little bit more confusing. @jfirebaugh pointed me at #2096 as a guideline.

src/symbol/collision_feature.js
ChrisLoer: Yeah, this one adds collision detection cost even in the un-pitched state because these extra boxes get returned in the `CollisionTile` grid query.

The other place where collision detection gets more expensive is in `CollisionTile::placeCollisionFeature`, where we expand the size of our grid query based on the "perspective ratio". That change, however, only really affects tiles that are pitched/distant.

src/ui/map.js
ChrisLoer: This is just to keep `map.test.js` working -- the style test stubs can't handle `redoPlacement` calls. I'll take another look at updating the stubs.

src/geo/transform.js
ansis: Thanks, can you add a this explanation as a comment?

src/render/draw_symbol.js
ansis: Commentifying this as well would be good

src/shaders/symbol_icon.vertex.glsl
ansis: 👍

src/symbol/collision_feature.js
ansis: I was thinking it might be good to only have the extra boxes for pitched maps, but considering we're planning on taking a deep look at collision stuff soon it's probably not worthwhile to add that now. This looks good as is

ChrisLoer: Yeah, although it makes collision detection slightly more responsive when you pitch the map, calculating these ahead of time probably isn't too important. The actual boxes are created at tile parsing time, so they need to be built regardless of pitch. However, at placement time, we could avoid inserting them into the collision grid in `CollisionTile::insertCollisionFeature`

src/ui/map.js
ChrisLoer: I fixed this by back-porting the 300ms throttling changes from `gl-native` (this expands the scope of the change by applying the throttling to symbol re-placement in response to pitch/rotate, as well as pan).

The new code fixes a real bug in the old implementation: the re-placement timeout can't outlive the map anymore.

src/symbol/collision_feature.js
ChrisLoer: I pushed a commit that skips adding these boxes to the CollisionTile on unpitched maps, so at least in that case there shouldn't be much of a performance impact. There is probably a more optimal version in which we calculate the number of padding boxes based on the tile distance, but I'm not worrying about that for now.

ChrisLoer:
@ansis Thanks for the review, I feel a lot better about this with your eyes on it!

I'm working on a gl-js fix for the wrapped tiles issue that I didn't discover until I started working on the gl-native port. I'm working through edge case tests in the query geometry stuff now (the `featureIndex` for two wrapped copies of the same tile can be different now). Once I have tests passing and can get some review on those changes, I think it'll be fine to merge -- and hopefully follow up with the native PR not too long after.

src/source/image_source.js
```diff
@@ -140,40 +141,43 @@ class ImageSource extends Evented {
...
-            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
-            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
-            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, image);
-        } else if (resize) {
-            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, image);
-        } else if (image instanceof window.HTMLVideoElement || image instanceof window.ImageData || image instanceof window.HTMLCanvasElement) {
-            gl.bindTexture(gl.TEXTURE_2D, this.tile.texture);
-            gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, image);
+        for (const w in this.tiles) {
+            const tile = this.tiles[w];
```
asheemmamoowala: This is creating multiple textures in the wrapped coordinates. Is it possible to create just one texture and reuse it in all the wrapped tiles ?

ChrisLoer: OK, I took a stab at sharing the texture between tiles -- it passes tests, but I could definitely use a review to see if the approach makes sense. All the image/texture binding stuff just looks like a magic incantation to me.

src/source/image_source.js
```diff
@@ -140,40 +141,43 @@ class ImageSource extends Evented {
...
 
-        this.tile.boundsBuffer = Buffer.fromStructArray(array, Buffer.BufferType.VERTEX);
-        this.tile.boundsVAO = new VertexArrayObject();
+        tile.boundsBuffer = Buffer.fromStructArray(array, Buffer.BufferType.VERTEX);
+        tile.boundsVAO = new VertexArrayObject();
     }
 
     prepare() {
-        if (!this.tile || !this.image) return;
+        if (this.tiles.length === 0 || !this.image) return;
```
anandthakker: `this.tiles` is an object, not an array, so I think you need `Object.keys(this.tiles).length === 0`

src/source/query_features.js
```diff
@@ -49,21 +51,38 @@ function sortTilesIn(a, b) {
...
                 resultFeatures = result[layerID] = tileFeatures;
+                tileFeatures.forEach((tileFeature) => {
+                    wrappedIDFeatures[tileFeature.id] = true;
+                });
             } else {
-                for (let f = 0; f < tileFeatures.length; f++) {
-                    resultFeatures.push(tileFeatures[f]);
+                for (const tileFeature of tileFeatures) {
+                    if (!tileFeature.id || !wrappedIDFeatures[tileFeature.id]) {
+                        wrappedIDFeatures[tileFeature.id] = true;
```
anandthakker: If I'm reading this right, it looks like in the case where features do not have an `id` and there are two tiles w/ same wrapped id, the `wrappedIDFeatures` object will look like `{ 'undefined': true }` after the first of those two tiles' results are collected.  Then, if the results from the second tile have features that _aren't_ duplicates, they'd be skipped.

src/source/image_source.js
ChrisLoer: Oops, good catch.

src/source/video_source.js
```diff
@@ -107,7 +107,7 @@ class VideoSource extends ImageSource {
     // setCoordinates inherited from ImageSource
 
     prepare() {
-        if (!this.tile || this.video.readyState < 2) return; // not enough data for current position
+        if (this.tiles.length === 0 || this.video.readyState < 2) return; // not enough data for current position
```
anandthakker: See comment in ImageSource

src/source/query_features.js
ChrisLoer: If the features don't have IDs, they'll get duplicated (because of the `!tileFeature.id` check). If the feature has an ID, it should get included once. Say the new feature has ID "1" and `wrappedIDFeatures` looks like `{ 'undefined': true }`. `wrappedIDFeatures[1]` will still be `undefined`. Am I missing something?

Do you have any opinion on whether it might make sense to just change the behavior here to "duplicate on wrapped tiles"? The current behavior is publicly documented (https://www.mapbox.com/mapbox-gl-js/api/#map#queryrenderedfeatures) so I'm thinking the answer is "no", but on the other hand users of the API can already see the same feature returned multiple times across tile boundaries.

ChrisLoer:
Benchmarks look reasonable to me, although I'm not sure what the expected variability is:

benchmark | master 010f425 | cloer_text_pitch_scale bdb217a
--- | --- | ---
**map-load** | 216 ms  | 102 ms 
**style-load** | 70 ms  | 42 ms 
**buffer** | 977 ms  | 1,106 ms 
**fps** | 59 fps  | 60 fps 
**frame-duration** | 4.2 ms, 0% > 16ms  | 3.6 ms, 0% > 16ms 
**query-point** | 1.92 ms  | 2.09 ms 
**query-box** | 76.89 ms  | 83.43 ms 
**geojson-setdata-small** | 5 ms  | 2 ms 
**geojson-setdata-large** | 70 ms  | 99 ms

src/source/query_features.js
ChrisLoer: @jfirebaugh Pointed me at #2321: we definitely intend to avoid duplicates. I pushed changes to get rid of the `!tileFeature.id` case by using the grid key from `feature_index` as a unique identifier for de-duplication.

src/source/video_source.js
```diff
@@ -107,7 +107,7 @@ class VideoSource extends ImageSource {
     // setCoordinates inherited from ImageSource
 
     prepare() {
-        if (!this.tile || this.video.readyState < 2) return; // not enough data for current position
+        if (Object.keys(this.tiles).length === 0 || this.video.readyState < 2) return; // not enough data for current position
```
lbud: The `CanvasSource` class also inherits from `ImageSource` and will need this change in its `prepare` method as well.

ChrisLoer: I updated `CanvasSource`, but I didn't see an equivalent of `debug/image.html` for testing it -- do we have any working examples?

ChrisLoer: I pushed a simple `debug/canvas.html`. Looks like it's rendering fine.

lbud: Awesome thanks @ChrisLoer!

PR #4549, opened 2017-04-06
Fix expired tiles refreshing

edpop:
Hello.
There is true and false mixed up.
When I initialized my map with 'refreshExpiredTiles: false', it started to refresh.


edpop:
Sorry, I was too hasty. Now I checked all code for this feature.

lbud:
🙇‍♀️ wonderful, thanks @edpop!

PR #4553, opened 2017-04-06
Fix bug calculating image source tile coordinates

### LINKED ISSUE #4550: Image source fails to render when coordinate area is too large

### Steps to Trigger Behavior

In the example below, change `longitude` to 91 or `latitude` to 67.

https://jsbin.com/daladaqaze/edit?html,output
```js
  var longitude = 90;
  var latitude = 60;
  var imageStyle = {
    "version": 8,
    "sources": {
        "image": {
            "type": "image",
            "url": "https://i.imgur.com/G06coPX.png",
            "coordinates": [
                [-longitude, latitude],
                [longitude, latitude],
                [longitude, -latitude],
                [-longitude, -latitude]
            ]
        },
        "geojson": {
          "type": "geojson",
          "data": {
            "type": "LineString",
            "coordinates": [
              [-180, -85],
              [-180, 85]
            ]
          }
        }
    },
    "layers": [{
        "id": "background",
        "type": "background",
        "paint": {
            "background-color": "rgb(4,7,14)"
        }
    },
    {
        "id": "dateline",
        "type": "line",
        "source": "geojson",
        "paint": { "line-color": "red" }
    },
    {
        "id": "image",
        "type": "raster",
        "source": "image"
    }]
  };

  var map = new mapboxgl.Map({
      container: 'map',
      style: imageStyle,
      hash: false
  });
```

### Expected Behavior

Image should appear

### Actual Behavior

Image does not appear

anandthakker:
Closes #4550

Using `Math.round` on the `centerCoord` value here causes a coordinate
like `{zoom: 2, column: 1.5, row: 1.5}` to become
`{zoom: 2, column: 2, row: 2}`, which is out of bounds for a z2 tile;
the correct tile for containing those coordinates is `{zoom: 2, column: 1, row: 1}`

### DIFF
--- a/src/source/image_source.js
+++ b/src/source/image_source.js
@@ -106,16 +106,28 @@ class ImageSource extends Evented {
         // may be outside the tile, because raster tiles aren't clipped when rendering.
 
         const map = this.map;
+
+        // transform the geo coordinates into (zoom 0) tile space coordinates
         const cornerZ0Coords = coordinates.map((coord) => {
             return map.transform.locationCoordinate(LngLat.convert(coord)).zoomTo(0);
         });
 
+        // Compute the coordinates of the tile we'll use to hold this image's
+        // render data
         const centerCoord = this.centerCoord = util.getCoordinatesCenter(cornerZ0Coords);
-        centerCoord.column = Math.round(centerCoord.column);
-        centerCoord.row = Math.round(centerCoord.row);
-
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Can you add a regression test for this?

anandthakker:
@jfirebaugh oops I must have failed to `git add` it (and then proceeded to lose the local copy 🙄 ).  Added in #4554

PR #4556, opened 2017-04-07
enforce TileJSON bounds 

### LINKED ISSUE #1775: Support bounds in TileJSON

mollymerp:
wraps up #3703 -- thank you @jlaxson for your contribution! 
closes #1775

cc @vincentsarago @anandthakker 

## Launch Checklist

 - [ ] post benchmark scores


src/source/tile_bounds.js
```diff
@@ -0,0 +1,33 @@
...
+        this.bounds = LngLatBounds.convert(bounds);
+        this.minzoom = minzoom || 0;
+        this.maxzoom = maxzoom || 24;
+    }
+
+    contains(coord) {
+        const level = [
+        [Math.floor(this.lngX(this.bounds.getWest(), coord.z)), Math.floor(this.latY(this.bounds.getNorth(), coord.z))],
+        [Math.ceil(this.lngX(this.bounds.getEast(), coord.z)), Math.ceil(this.latY(this.bounds.getSouth(), coord.z))]
+        ];
```
anandthakker: nit: for readability, instead of an array, could this be four individual `const`s `minX`, `minY`, etc.?

test/unit/source/vector_tile_source.test.js
```diff
@@ -178,5 +178,18 @@ test('VectorTileSource', (t) => {
...
+            maxzoom: 22,
+            attribution: "Mapbox",
+            tiles: ["http://example.com/{z}/{x}/{y}.png"],
+        });
+        source.setBounds([[-47, -7], [-45, -5]]);
+        t.false(source.hasTile({z: 8, x:96, y: 132}), 'returns false for tiles outside bounds');
+        t.true(source.hasTile({z: 8, x:95, y: 132}), 'returns true for tiles inside bounds');
+        t.end();
+    });
+
```
anandthakker: Let's also include tests:
 - for the case where `bounds` is included directly on the TileJSON (i.e. where the source is created with `{url: ...}`
 - equivalent ones for raster source
 - in `source_cache.test.js`, confirming that SourceCache uses `hasTile()` to filter which tiles it loads.

jlaxson:
Wow.  This is awesome, thanks @mollymerp.  Can't believe I never got back to this after Thanksgiving (!).

mourner:
@mollymerp looks like we now have some code duplication between the new `tile_bounds.js` and `transform` (which also has projection code). Can we get rid of the first and reuse the transform methods?

anandthakker:
@mourner we'd need to extract the common calculation into `util`, because they aren't 100% identical (the 'world size' differs), and also because using transform directly would introduce an unnecessary dependency of `TileBounds` on `Transform`.

anandthakker:
I'm also not convinced that avoiding _one_ duplication is critical...

mourner:
@anandthakker I mean, not directly instantiating `Transform` in `TileBounds`, but rather using `map.transform` methods (e.g. `locationCoordinate`) directly in `RasterTileSource` to avoid introducing `TileBounds` class altogether. This is not critical, but code simplification is always welcome.

anandthakker:
> rather using map.transform methods (e.g. locationCoordinate) directly in RasterTileSource

Hm, along the lines of https://github.com/mapbox/mapbox-gl-js/issues/3350, I would be hesitant to increase any `Source`'s reliance on the parent `Map` object.

underbluewaters:
Any reason this bounds parameter isn't in the [tile spec Sources documentation](https://www.mapbox.com/mapbox-gl-js/style-spec/#sources)? I was wondering if just this feature was supported but I had to dig through the tickets here.

kkaefer:
@underbluewaters good call. The source class is a combination of [TileJSON](https://github.com/mapbox/tilejson-spec/tree/master/2.2.0) (which does have this property), and a few custom properties. Ticketed in https://github.com/mapbox/mapbox-gl-js/issues/5124

PR #4542, opened 2017-04-07
Reduce number of varyings for line_sdf shader

kkaefer:
The `line_sdf` shader currently uses a lot of varyings—too many for some OpenGL implementations. They fail to link the program with a message of `Too many varyings`. [OpenGL ES 2.0](https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glGet.xml) has a minimum value of 8 for `GL_MAX_VARYING_VECTORS`, while this shader uses 10 or 11.

I believe we could easily merge some of the individual `float` varyings into a vector.

/cc @anandthakker

### DIFF
--- a/src/data/program_configuration.js
+++ b/src/data/program_configuration.js
@@ -68,7 +68,9 @@ class ProgramConfiguration {
         const pragmas = this.getPragmas(name);
 
         pragmas.define.push(`uniform {precision} {type} ${inputName};`);
+        pragmas.define_in.push(`uniform {precision} {type} ${inputName};`);
         pragmas.initialize.push(`{precision} {type} ${name} = ${inputName};`);
+        pragmas.initialize_in.push(`{precision} {type} ${name} = ${inputName};`);
 
         this.cacheKey += `/u_${name}`;
     }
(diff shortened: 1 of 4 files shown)


jfirebaugh:
Rather than introducing additional syntax, could the `#pragma` expander compare the set of pragmas in the vertex shader to the set in the fragment shader, and generate varyings only for attributes present in both?

jfirebaugh:
Even after doing that though, the circle and symbol_sdf shaders still use 9 varyings.

jfirebaugh:
Fixed in native with https://github.com/mapbox/mapbox-gl-native/pull/9251.

jfirebaugh:
https://github.com/mapbox/mapbox-gl-js/pull/4828 gets circle and symbol_sdf below 8 varyings. Remaining task in gl-js is:

> compare the set of pragmas in the vertex shader to the set in the fragment shader, and generate varyings only for attributes present in both

jfirebaugh:
I have work in progress to implement the above as part of bringing the gl-js DDS implementation closer to native.

PR #4563, opened 2017-04-10
Fix to add .json extension required in webpack environments

orangemug:
Webpack requires `.json` extension to be present when requiring json files. So `require('../reference/v7');` will currently fail, this PR fixes that. All other environments currently support the  `require('../reference/v7.json');` syntax.

### DIFF
--- a/src/style-spec/migrate/v7.js
+++ b/src/style-spec/migrate/v7.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const ref = require('../reference/v7');
+const ref = require('../reference/v7.json');
 
 function eachLayer(layer, callback) {
     for (const k in layer.layers) {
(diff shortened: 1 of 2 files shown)


mourner:
Thanks! I'll merge if the CI build passes.

orangemug:
Thanks @mourner can you release to https://www.npmjs.com/package/@mapbox/mapbox-gl-style-spec also?

orangemug:
Hi @anandthakker I've noticed you released v9 of https://www.npmjs.com/package/@mapbox/mapbox-gl-style-spec some things look a little odd

 - https://www.npmjs.com/package/@mapbox/mapbox-gl-style-spec points to https://github.com/mapbox/mapbox-gl-style-spec which is no longer a repo
 - https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/package.json is still on v8 as is the changelog in that directory.

I assume I'm looking in the wrong place, in which case this PR I raised might also be invalid. Any clarification would be great.

orangemug:
Ignore the previous comment, it's up to date in the repo now

Issue #1912, opened 2016-01-08
separate Tile's style zoom level from it's data zoom level


ansis:
The current implementation of overscaled tile reparsing abuses the `z` coordinate to store the zoom level that needs to be used for style parsing. The real `z` is recovered using sourceMaxZoom. It's really hard to follow how this works or if it works.

A TileCoord should have two separate z's. One for the style and one for the actual vectortile.

lucaswoj:
ref #1909

lucaswoj:
Closing in favor of #2868 which highlights user-facing behaviour and doesn't specify an implementation.

PR #4567, opened 2017-04-11
fix bug causing overzoomed tiles to disappear

### LINKED ISSUE #4564: Layer Not Shown Past Tileset Max Zoom
I have a vector tileset with a max zoom level of 11. When I zoom at level 12 or higher the layer for that tileset is no longer shown. This worked in version 0.34, but is broken in version 0.35.

Example: https://jsfiddle.net/x9xjdzz9/e7h55nvt/

mollymerp:
Fixes #4564

For overzoomed tiles, `TileCoord#cover` returns TileCoords with [incorrect `z`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/tile_coord.js#L161), so they dont appear to be within `TileJSON.bounds` for vector sources with that property, and thus (as of #4556 ) the SourceCache doesn't consider them part of `visibleCoords` (they get filtered out [here](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/source_cache.js#L341)) and so the source disappears if it's over its `maxzoom`

For instance in the case of the [test case](https://jsfiddle.net/x9xjdzz9/e7h55nvt/) provided in #4564, the source has a `maxzoom` of 11, and the tiles within its bounds are [`11/583/782`, `11/584/782`, `11/583/783`, `11/584/784`] but when you zoom to ZL 12, `transform.coveringTiles` returns [`12/583/782`, `12/584/782`, `12/583/783`, `12/584/784`] which are not the correct tiles, and I guess doesn't cause problems besides this? Or maybe it's like this for a reason?
 
This PR doesn't fix that bc I'm scared to change `TileCoord` without understanding what's going on there. If the behavior of `TileCoord` is indeed a bug I can modify this PR to fix that. This PR is just a workaround to prevent over-zoomed vector sources with `bounds` from disappearing. 

@anandthakker @jfirebaugh 

## Launch Checklist

 - [ ] post benchmark scores


mourner:
That's a known gotcha we've been meaning to fix — I attempted at some point but got stuck because it affects a lot of code throughout Mapbox GL. There should be an open ticket about this, trying to find it...

As for the fix, I think we should pass `source.maxzoom` to `hasTile` method and then just do `Math.min(maxzoom, coord.zoom)` inside — it's handled like this in other places currently. It's confusing, but will fix the issue while retaining the bounds filtering logic on overscaled zooms.

mourner:
@mollymerp found the relevant tickets: #1912 and #2868

mollymerp:
Thanks for the background @mourner - i've update the PR and will add a regression test next

PR #4581, opened 2017-04-12
Guard against non-numeric id in GeoJSON features

### LINKED ISSUE #4494: Fix bug causing queryRenderedFeatures to crash on polygon features that have an `id` field.
**mapbox-gl-js version**: v0.34.0

### Steps to Trigger Behavior

 1. Add a GeoJSON polygon source with non-integer (string) ids
 2. Call `map.queryRenderedFeatures()` on `mousemove`

### Expected Behavior

`map.queryRenderedFeatures()` should return an array of features under the mouse pointer

### Actual Behavior

`Uncaught Error: Unimplemented type: 7`

I have created a JSFiddle of the failing code: https://jsfiddle.net/p78o8exf/6/

This code works fine with v0.32.1 but since v0.33 it is broken. This is valid GeoJSON (checked with `@mapbox/geojsonhint`).

anandthakker:
Closes #4494

Since we're using the vector tile binary format even for geojson-vt tiles, make
sure we conform to the MVT requirement that `feature.id` is an integer


jfirebaugh:
#4494 reports that the issue didn't occur in v0.32.1. Were you able to confirm that and determine why?

anandthakker:
@jfirebaugh yep - v0.32.1 is the last release before https://github.com/mapbox/mapbox-gl-js/commit/cf031a6d072bda98b51c57021dfd767eb598ab4c, which is when the geojson-vt wrapper started copying `feature.id`.

src/source/geojson_wrapper.js
```diff
@@ -17,8 +17,8 @@ class FeatureWrapper {
             this.rawGeometry = feature.geometry;
         }
         this.properties = feature.tags;
-        if ('id' in feature) {
-            this.id = feature.id;
+        if ('id' in feature && !isNaN(feature.id)) {
```
jfirebaugh: Can you add a comment here explaining the original issue and the reason for `!isNaN`/`parseInt`?

PR #4584, opened 2017-04-12
Add stylelint CLI and Add mapbox class name rule in stylelintrc

asantos3026:
## Launch Checklist

- added stylelint as a dev dependency 
- added npm script lint-css 
- added rule in stylelintrc to make sure all selectors begin with mapbox
- tested npm script in CLI to make sure it works
- amended npm warning by adding || true to the script 


asantos3026:
@anandthakker - the Circle CI tests are failing but not for my PR issue.

debug/access_token.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 mapboxgl.accessToken = getAccessToken();
+require('dotenv').config()
```
mollymerp: what was the reason for adding this change?

.stylelintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "rules": {
+    "selector-class-pattern": "mapbox[a-z-]+",
+    "color-hex-case": "lower",
```
mollymerp: I'm not sure we need the color rules. we rarely change `dist/mapbox-gl.css` and we don't deal with much color in there.

package.json
```diff
@@ -18,6 +18,7 @@
     "@mapbox/whoots-js": "^3.0.0",
     "brfs": "^1.4.0",
     "bubleify": "^0.7.0",
+    "dotenv": "^4.0.0",
```
mollymerp: I dont think we need this dependency

package.json
```diff
@@ -45,6 +46,7 @@
     "clipboard": "^1.5.12",
     "concat-stream": "^1.6.0",
     "coveralls": "^2.11.8",
+    "csscolorparser": "^1.0.3",
```
mollymerp: I think we can do without this dependency too

package.json
```diff
@@ -115,8 +118,9 @@
     "start-docs": "npm run build-min && npm run build-docs && jekyll serve --watch",
     "lint": "eslint --cache --ignore-path .gitignore src test bench docs/_posts/examples/*.html debug/*.html",
     "lint-docs": "documentation lint src/index.js",
+    "lint-css": "stylelint 'docs/*.css' || true",
```
mollymerp: I think you should be pointing the linter at `dist/mapbox-gl.css` and not the `docs/` css files. We only enforce the `mapboxgl-` prefex for css classes that ship with the library distribution. The CSS in the `docs/` folder is only for the mapbox documentation website - www.mapbox.com/mapbox-gl-js/api 
That change might also remove the need for the `|| true` statement, although I'm not 100% clear on what that is doing in this case.

src/.env
```diff
@@ -0,0 +1 @@
+MAPBOX_ACCESS_TOKEN=sk.eyJ1IjoiYXNhbnRvczMwMjYiLCJhIjoiY2oxY293bGk4MDA2NDMzcGRxNGdyc3lpbSJ9.uxcZ_o19abopWFm2vwstmg
```
mollymerp: We'd like to avoid committing personal access tokens to the GitHub repo. Developers will need to use their own token with the library.

mollymerp:
Ah just saw your discussion w @anandthakker on #4584 👍  on adding `stylelint` but would like to avoid the other two dependencies in this PR. Thanks again! 

closes #4584

debug/access_token.js
asantos3026: to access the mapbox API with my keys in a .env file, come to think of it I probably do not need to do that because I am running the server in the terminal

.stylelintrc
asantos3026: Took out the color rules

package.json
asantos3026: took out the dotenv dependency

package.json
asantos3026: took out this dependency

mollymerp:
In order to get CI to pass based on I think you'll need to add a line ignoring `stylelint` in the [`.flowconfig`](https://github.com/mapbox/mapbox-gl-js/blob/master/.flowconfig) file based on https://github.com/stylelint/stylelint/issues/2322#issuecomment-280815935

package.json
```diff
@@ -115,14 +116,15 @@
     "start-docs": "npm run build-min && npm run build-docs && jekyll serve --watch",
     "lint": "eslint --cache --ignore-path .gitignore src test bench docs/_posts/examples/*.html debug/*.html",
     "lint-docs": "documentation lint src/index.js",
+    "lint-css": "stylelint 'dist/mapbox-gl.css' || true",
```
mollymerp: can we remove the `|| true` statements from here and the `test-flow` command – I'm afraid they will prevent the tests from failing if there are errors.

asantos3026:
@mollymerp Hey Molly! I made all of the following changes:

- removed true statement from test-flow and stylelint scripts
- added stylelint to be ignored in the .flowconfig file

Let me know if there is anything else I need to do to clean up my PR 👍

mollymerp:
great! thank you so much @asantos3026!!

PR #4595, opened 2017-04-17
gl-style-composite throws error on conflicting source layer names

andrewharvey:
## Launch Checklist

* Make gl-style-composite executable in line with the other tools.

* A style without source compositing can have layers referencing different source-layers with the same name but from different sources. Source compositing these styles is undefined as it could create a name conflict.

Studio asks users what they want to do in this case:
selection_559

This PR causes the gl-style-composite utility to throw an error if a name conflict would occur.

I'm not sure who else is using the utility and how, but I feel throwing an error and pushing it back to the user to resolve the conflicts first is probably the better way to go...

 - [ ] post benchmark scores


test/unit/style-spec/composite.test.js
```diff
@@ -78,3 +78,34 @@ test('incorrect url match', (t) => {
...
+                "source-layer": "sourcelayer",
+                "source": "mapbox-a"
+            }, {
+                "id": "b",
+                "type": "line",
+                "source-layer": "sourcelayer",
+                "source": "mapbox-b"
+            }]
+        });
+    }, 'Conflicting source layer names', 'throws error on conflicting source layer names');
```
jfirebaugh: `'Conflicting source layer names'` ⇢ `/Conflicting source layer names/`

andrewharvey: fixed

Issue #4578, opened 2017-04-12
Accept multiple layers for Map#on


arthurhammer:
#1002 (pull #4329) added a `layer` parameter to `Map#on` to handle mouse-related events for a specific layer. Currently, `on` accepts a single layer. It would be great if passing in an array of layers was supported. I believe this would make handling mouse-events even easier.

mourner:
Thanks for creating the issue! Want to go slightly further and submit a pull request for this? Should be a quick change.

stepankuzmin:
Hi there! If you don't mind, I've created PR for this issue https://github.com/mapbox/mapbox-gl-js/pull/4329.

I'm not sure about comparing `delegatedListener.layers` with `layers` param in `Map#off` (see https://github.com/mapbox/mapbox-gl-js/pull/4596/files#diff-c6d02b0c73aa7597f4224ec35497cfe2R639). Should we preserve layers order in `Map#off`?

jfirebaugh:
> I'm not sure about comparing `delegatedListener.layers` with `layers` param in `Map#off` (see https://github.com/mapbox/mapbox-gl-js/pull/4596/files#diff-c6d02b0c73aa7597f4224ec35497cfe2R639). Should we preserve layers order in `Map#off`?

This is a great question. Specifically, how should the following cases behave:

* `on(['a', 'b'])` followed by `off(['b', 'a'])`
* `on(['a', 'b'])` followed by `off(['a'])`
* `on(['a', 'b'])` followed by `on(['a'])` followed by `off(['a'])`
* `on(['a', 'b'])` followed by `on(['a'])` followed by `off(['a', 'b'])`
* `on(['a'])` followed by `on(['a', 'b'])` followed by `off(['a', 'b'])`

The ambiguity and additional complexity of these cases makes me think we probably shouldn't implement this. If someone needs it, they can add/remove listeners in a loop.

jfirebaugh:
Let's keep the API simple.

stepankuzmin:
This actually makes sense. Thanks for review!

PR #4596, opened 2017-04-17
Accept multiple layers for Map#on #4578

stepankuzmin:
Add multiple layers support in `Map#on` according to #4578


jfirebaugh:
Thanks for the contribution @stepankuzmin. However in this case I think it's better to keep the API simple; see https://github.com/mapbox/mapbox-gl-js/issues/4578#issuecomment-294530831.

PR #4603, opened 2017-04-18
enable query tests for mapbox-gl-native#8007

ivovandongen:
Enables query tests for https://github.com/mapbox/mapbox-gl-native/issues/8007

I've changed to geometry in line-offset/property-function to account for wagyu clipping. We need something more robust in the future, but that shouldn't block this pr.


test/integration/query-tests/regressions/mapbox-gl-js#4494/style.json
```diff
@@ -2,6 +2,9 @@
   "version": 8,
   "metadata": {
     "test": {
+      "ignored": {
```
jfirebaugh: See https://github.com/mapbox/mapbox-gl-js/pull/4604 for an alternative solution for this one.

ivovandongen:
@jfirebaugh @kkaefer I've disabled the test for https://github.com/mapbox/mapbox-gl-native/pull/8760 for now. We can re-enable once that is good to go.

Issue #3949, opened 2017-01-11
Flash when switching from a DDS paint property to a constant


stevage:
**mapbox-gl-js version**: 0.28.0 and 0.30.0

### Steps to Trigger Behavior

 1. https://city-of-melbourne.github.io/anything-map/#84bf-dihi
 2. Hover over a point, then click 'VENUE_RECN'
 3. Click the Close X button

### Expected Behavior
1. Circles the same size
2. Circles vary in size
3. Circles the same size

### Actual Behavior
In step 3:

```
Uncaught TypeError: this.functionInterpolationT is not a function
    at StyleDeclaration.calculateInterpolationT (style_declaration.js:58)
    at t.i.getPaintInterpolationT (style_layer.js:162)
    at ProgramConfiguration.setUniforms (program_configuration.js:193)
    at Object.drawCircles [as circle] (draw_circle.js:30)
    at Painter.renderLayer (painter.js:271)
    at Painter.renderPass (painter.js:254)
    at Painter.render (painter.js:213)
    at e._render (map.js:1128)
```

Sorry, this is far from a minimal repro case, but what is happening: after initially creating the style, in step 2 I call `map.setPaintProperty('points','circle-radius', p)` where p is:

```
{
  "property": "VENUE_RECN",
  "stops": [
    [
      {
        "zoom": 10,
        "value": 5
      },
      1
    ],
    [
      {
        "zoom": 10,
        "value": 1627
      },
      3
    ],
    [
      {
        "zoom": 17,
        "value": 5
      },
      3
    ],
    [
      {
        "zoom": 17,
        "value": 1627
      },
      10
    ]
  ]
}
```

In step 3, I call it with this parameter:

```
{
  "stops": [
    [
      10,
      2
    ],
    [
      17,
      5
    ]
  ]
}
```

Before I go too far creating a minimal repro case, just wanted to check it wasn't already known.

lucaswoj:
Thanks for the bug report! These symptoms don't look familiar to me.

jfirebaugh:
@stevage I don't see a "Close X" button on the page after following those steps. Would you mind creating a minimal demo?

stevage:
Sorry, I removed the close button in that case because of this bug :) Will create a minimal repro.

jfirebaugh:
Closing here for now; feel free to reopen if this continues to be an issue and you have a chance to create a minimal repro.

aparlato:
[...]

PR #4611, opened 2017-04-19
Fix error when switching from DDS to a constant paint value

mourner:
Makes #3949 less severe (a flash of incorrect value instead of a JS error). cc @aparlato @stevage


mollymerp:
@mourner does it make sense to add a test case here?

PR #4616, opened 2017-04-19
Adjust test to work around another bug

jfirebaugh:
Apparently gl-js has a bug where a symbol exactly on a tile boundary will get overdrawn. I'll file a separate issue for this.


jfirebaugh:
https://github.com/mapbox/mapbox-gl-js/issues/4617

PR #4624, opened 2017-04-23
Add flowtype definitions for geo/lng_lat.js

vicapow:


src/geo/lng_lat.js
```diff
@@ -98,30 +101,33 @@ class LngLat {
...
+        }
+        if (!Array.isArray(input) && typeof input === 'object' && input !== null) {
+            return new LngLat(Number(input.lng), Number(input.lat));
+        }
         throw new Error("`LngLatLike` argument must be specified as a LngLat instance, an object {lng: <lng>, lat: <lat>}, or an array of [<lng>, <lat>]");
     }
-};
+}
+
+/*:: export type LngLatLike = LngLat | {lng: number, lat: number} | [number, number]; */
```
mourner: the current setup doesn't handle non-comment version?

vicapow: @mourner unfortunately, no. eslint throws a parse error that's not ignorable because `sourceType` is set to "script" in [.eslintrc](https://github.com/mapbox/mapbox-gl-js/blob/master/.eslintrc). On top of this, Flow doesn't support requiring types using CommonJS. There may be a way to configure eslint to ignore type imports, but I haven't found it. An alternative is to migrate all the types to comments. Otherwise, I could see it being annoying if only some types are commented and others aren't.

PR #4625, opened 2017-04-24
Add flowtype definitions for geo/transform

vicapow:


src/geo/transform.js
```diff
@@ -365,19 +399,25 @@ class Transform {
 
         this._constraining = true;
 
-        let minY, maxY, minX, maxX, sy, sx, x2, y2;
+        let minY = -90;
+        let maxY = 90;
+        let minX = -180;
+        let maxX = 180;
```
mourner: Isn't this changing the current behavior? E.g. could break current hacks for non-geographic maps.

src/geo/transform.js
```diff
@@ -146,7 +171,7 @@ class Transform {
      *
      * @private
      */
-    getVisibleWrappedCoordinates(tileCoord) {
+    getVisibleWrappedCoordinates(tileCoord: {x: number, y: number, z: number}) {
```
mourner: Is there an advantage to being explicit like this instead of using the `TileCoord` type?

src/geo/transform.js
```diff
@@ -19,7 +20,31 @@ const vec4 = glmatrix.vec4,
...
+    _fov: number;
+    _pitch: number;
+    _zoom: number;
+    _unmodified: boolean;
+    _renderWorldCopies: boolean;
+    _minZoom: number;
+    _maxZoom: number;
+    _center: LngLat;
+    _constraining: boolean;
+    constructor(minZoom?: ?number, maxZoom?: ?number, renderWorldCopies?: boolean) {
```
mourner: What's the meaning of two questionmarks, on both ends of the colon?

src/geo/transform.js
```diff
@@ -199,12 +233,12 @@ class Transform {
...
-    get unmodified() { return this._unmodified; }
+    get unmodified(): boolean { return this._unmodified; }
 
-    zoomScale(zoom) { return Math.pow(2, zoom); }
-    scaleZoom(scale) { return Math.log(scale) / Math.LN2; }
+    zoomScale(zoom: number) { return Math.pow(2, zoom); }
+    scaleZoom(scale: number) { return Math.log(scale) / Math.LN2; }
 
-    project(lnglat) {
+    project(lnglat: {lng: number, lat: number}) {
```
mourner: Same — any advantage over just using `LngLat`?

src/geo/transform.js
vicapow: `?:` in front means the property can be absent (not just `undefined`). In the case of arguments to a function, it might not really make sense, though.
````js
({foo: undefined}: {foo?: string}); // no error
({}: {foo?: string}); // no error
({}: {foo: string| void}); // will error with property not found.
({}: {foo: string| typeof undefined}); // will error...
({}: {foo: ?string}); // will error...
````

File: src/geo/transform.js (hunk 2 above)
Comment (vicapow) [inline src/geo/transform.js]: It depends. In this case, I think it's OK since we don't actually require any methods or other properties of the TileCoord class. If we ever do, flow will complain and the type can be refined.

File: src/geo/transform.js (hunk 4 above)
Comment (vicapow) [inline src/geo/transform.js]: ditto

Comment (vicapow) [inline src/geo/transform.js]: Is also saves us two extra type imports

File: src/geo/transform.js (hunk 1 above)
Comment (vicapow) [inline src/geo/transform.js]: Flow was complaining about likes like 449 since they could still be uninitialized. It seems a little unexpected different projects would use different min/max values for longitude and latitude.

File: src/geo/transform.js (hunk 3 above)
Comment (vicapow) [inline src/geo/transform.js]: updated this to remove the extra `?`

File: src/geo/transform.js (hunk 2 above)
Comment (vicapow) [inline src/geo/transform.js]: updated this to use `TileCoord`

File: src/geo/transform.js (hunk 4 above)
Comment (vicapow) [inline src/geo/transform.js]: updated this to use `LngLat`

File: src/geo/transform.js (hunk 1 above)
Comment (vicapow) [inline src/geo/transform.js]: still wasn't sure what to do about this one. Could either add `// $FlowIgnoreMe` or use these values if minY is still undefined within each `if` block.

File: src/geo/transform.js [hunk 5]
```diff
@@ -19,7 +20,31 @@ const vec4 = glmatrix.vec4,
...
+    width: number;
+    height: number;
+    angle: number;
+    rotationMatrix: mixed;
+    zoomFraction: number;
+    pixelsToGLUnits: [number, number];
+    cameraToCenterDistance: number;
+    projMatrix: mixed;
+    pixelMatrix: mixed;
+    pixelMatrixInverse: mixed;
```
mourner: Why are matrices `mixed`? Does `?number[]` work here?

vicapow: Mainly because I wasn't sure of their shape at the time. I'll update

vicapow: I changed these to be `Float64Array`.

vicapow:
Hey @mourner can I get a merge?

PR #4632, opened 2017-04-24
Add flowtype definitions for mapbox-gl style.

vicapow:


vicapow:
To follow up, these style definitions aren't used anywhere yet, but I wanted to seperate them out into a seperate commit. I guess I could instead submit the full changes of this being used in at least one file first?

jfirebaugh:
I think we definitely want these, but it would probably be best to generate them from the style spec JSON so we don't have two sources of truth.

BTW, thanks for all the PRs adding flow types!

vicapow:
> I think we definitely want these, but it would probably be best to generate them from the style spec JSON so we don't have two sources of truth.
Could you point me to the style spec JSON definition?
> BTW, thanks for all the PRs adding flow types!
You're welcome! Glad to help

jfirebaugh:
https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/reference/v8.json

jfirebaugh:
Started on this #4958. I'm going to pull in your `TileJSON` definition though.

PR #4634, opened 2017-04-24
Fix an issue with rendering lines with duplicate points

### LINKED ISSUE #4579: Polyline with duplicate coordinate at the end not rendered correctly
See this [example](https://jsbin.com/lidamukopu/edit?html,js,output).

A polygon is not always rendered correctly when the last two coordinates are identical.
In the example, when you remove the last coordinate the polygon is rendered correctly.

mourner:
Fixes #4579.

 - [x] add a render test


kkaefer:
This potentially needs a fix in Native as well.

PR #4635, opened 2017-04-24
Increase duration of a flaky camera test

mourner:
A temporary measure to hopefully make a flaky test (failing 10-15% of the time) more stable, until we come up with a more reliable way to test this.

### DIFF
--- a/test/unit/ui/camera.test.js
+++ b/test/unit/ui/camera.test.js
@@ -1044,7 +1044,7 @@ test('camera', (t) => {
                 t.end();
             });
 
-            camera.flyTo({ center: [-170, 0], duration: 10 });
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Isn't the subsequent "westward" test just as likely to flake?

mourner:
@jfirebaugh I thought so too, but I checked all the recent failed builds and the only failed test was this one. I guess somehow the first test prepares the run of the second, maybe because of the non-changing location of animation.

jfirebaugh:
Still flaking FYI: https://circleci.com/gh/mapbox/mapbox-gl-js/7721

mourner:
Yeah, I noticed :(

PR #4673, opened 2017-05-03
stop using standard path module

pirxpilot:
it was only used to find shaders (glsl) files, but it's not really needed since
(1) we are using forward slashes anyway
(2) it's ignored by brfs plugin

removing it stops browserify from attaching browser compliant implementation of `path` to the resulting bundle

BEFORE:

    -rw-rw-r-- 1 user user 510K May  3 10:27 dist/mapbox-gl.js

AFTER:

    -rw-rw-r-- 1 user user 508K May  3 10:29 dist/mapbox-gl.js

### DIFF
--- a/src/render/shaders.js
+++ b/src/render/shaders.js
@@ -1,76 +1,77 @@
 'use strict';
 
 const fs = require('fs');
-const path = require('path');
 
 // readFileSync calls must be written out long-form for brfs.
+/* eslint-disable prefer-template, no-path-concat */
+
 module.exports = {
     prelude: {
-        fragmentSource: fs.readFileSync(path.join(__dirname, '../shaders/_prelude.fragment.glsl'), 'utf8'),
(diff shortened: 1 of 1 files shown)


pirxpilot:
> Can you disable no-path-concat in the shaders.js file instead of globally so that we could keep the check for other code (e.g. tests)?

Done.

> Can you make sure tests and the build pass correctly on Windows? That was the main motivation behind using path.

I cannot (no Windows machine) but...
- Windows FS and API are totally OK with forward slashes as path separators - Windows shell gets upset, but we are not using Windows shell here
- we were already using forward slashes in this file - `path.join(__dirname, '/../shaders/_prelude.fragment.glsl')` would have to be written as `path.join(__dirname, '..', 'shaders', '_prelude.fragment.glsl')` to ensure backslashes on Windows

That said - I totally get if you'd rather wait for someone with direct Windows experience to double check it.

PR #4681, opened 2017-05-04
Fixed popup misalignment with map during pan easing

bicubic:
Popup DOM transforms are deferred until `painter.render()` except when
Popup is being instantiated.

Popup offset integer rounding is suppressed during easing animations to
maintain sub-pixel alignment. The offset is rounded off during the final
tick of an easing animation to ensure DOM is pixel perfect on
completion.

DOM.setTransform takes new optional `immediate` parameter. If false, the
transform will be deferred until next canvas render. This change is
backwards compatible.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -300,7 +300,7 @@ class Popup extends Evented {
 
         //defer transform updates except when the DOM is instantiated
         //this ensures transform updates happen simultaneously with canvas flip
-        DOM.setTransform(this._container, transform, instantiating === true);
+        DOM.setTransform(this._container, transform, instantiating === true, this._map);
     }
 
     _onClickClose() {
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -300,7 +300,11 @@ class Popup extends Evented {
 
         //defer transform updates except when the DOM is instantiated
         //this ensures transform updates happen simultaneously with canvas flip
-        DOM.setTransform(this._container, transform, instantiating === true, this._map);
+        if (instantiating === true) {
+            DOM.setTransform(this._container, transform);
+        } else {
+            this._map.deferTransform(this._container, transform);
+        }
     }
 
     _onClickClose() {
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -75,7 +75,7 @@ class Popup extends Evented {
         if (this.options.closeOnClick) {
             this._map.on('click', this._onClickClose);
         }
-        this._update();
+        this._update(null, true);
         return this;
     }
 
@@ -229,7 +229,7 @@ class Popup extends Evented {
         }
     }
 
-    _update() {
+    _update(ev, instantiating) {
         if (!this._map || !this._lngLat || !this._content) { return; }
 
         if (!this._container) {
@@ -272,7 +272,13 @@ class Popup extends Evented {
             }
         }
 
-        const offsetedPos = this._pos.add(offset[anchor]).round();
+        let offsetedPos = this._pos.add(offset[anchor])
+
+        //snap to interger offset if this is last tick of an easing 
+        //animation, or if this event didn't come from easing loop
+        if(ev && (ev.finalEasingTick === true || ev.easing !== true)){
+            offsetedPos = offsetedPos.round();
+        }
 
         const anchorTranslate = {
             'top': 'translate(-50%,0)',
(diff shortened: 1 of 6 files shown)


src/ui/popup.js
```diff
@@ -272,7 +272,12 @@ class Popup extends Evented {
             }
         }
 
-        const offsetedPos = this._pos.add(offset[anchor]).round();
+        let offsetedPos = this._pos.add(offset[anchor]);
+        //snap to interger offset if this is last tick of an easing
+        //animation, or if this event didn't come from easing loop
+        if (ev && (ev.finalEasingTick === true || ev.easing !== true)) {
```
jfirebaugh: [`Marker` has similar logic](https://github.com/mapbox/mapbox-gl-js/blob/213d18a501fd505a846448599ca57d56a9469967/src/ui/marker.js#L153-L153), but the condition is based on checking the event type (`move` vs `moveend`). Will that work here too, instead of adding new properties to the event object?

jfirebaugh:
Thanks for tracking down the cause here and submitting a PR @bicubic! I have one question about the rounding part, and I think @mourner wants to look at the transform part.

src/ui/popup.js
bicubic: Thanks @jfirebaugh . It should work, I will look at folding that in after @mourner posts his feedback.

bicubic: @jfirebaugh @mourner do you guys have any further input or are you happy to merge once I look at switching to `moveend`

bicubic:
Alright, @mourner as per your feedback, transform management state has been moved to `Map`. 
`util.pushTransform` and `DOM.setTransform` now take a new optional map argument. As before, both functions are backwards compatible and using the old call signature will work the same as before. 

@jfirebaugh I have looked at your suggestion, unfortunately `moveend` is never received by `popup._update()` in this scenario and I don't have a good enough view of your codebase to understand why Marker receives that event. My guess is it's catering to a different use case and doesn't deal with pan easing (comments seem to refer to zooming there). I'm inclined to leave my workaround in place unless you feel strongly against it and are able to help me augment pan easing to fire `moveend` on the last tick.

bicubic:
Alright, @mourner as per your feedback, transform management state has been moved to `Map`. 
I have removed the previous util functions and reverted DOM.setTransform to the way it was in develop. Instead, callers must explicitly call the new `Map.deferTransform` method, e.g.:
```
if (instantiating === true) {
    DOM.setTransform(this._container, transform); //exactly the same as before, applied immediately
} else {
    this._map.deferTransform(this._container, transform); //coalesced, deferred until next render tick
}
```

The reason DOM.setTransform must continue to exist is because the transform must be set immediately when the overlay dom is created, deferring it causes the element to be out of position for one frame.

@jfirebaugh I have looked at your suggestion, unfortunately `moveend` is never received by `popup._update()` in this scenario and I don't have a good enough view of your codebase to understand why Marker receives that event. My guess is it's catering to a different use case and doesn't deal with pan easing (comments seem to refer to zooming there). I'm inclined to leave my workaround in place unless you feel strongly against it and are able to help me augment pan easing to fire `moveend` on the last tick.

jfirebaugh:
Thanks @bicubic, however this still doesn't feel like the right approach to me. The following issues give me pause:

* Why doesn't `moveend` work? I can't think of any reason why it wouldn't. This needs investigation.
* I'm not happy with introducing additional state to the map in the form of "deferred transitions". That seems like an approach that could cause race conditions or other unexpected behavior. Are there any other approaches you can think of?

bicubic:
Thanks @jfirebaugh 

> Why doesn't moveend work? 
I'll leave that up to you guys to figure out. I don't have the time to trace your eventing architecture nor do I know the intents of the original author. 

>  That seems like an approach that could cause race conditions or other unexpected behavior. Are there any other approaches you can think of?

Could look at having an event fired by the map immediately after painter.render, an event that's exclusively used to signal markers and popups to execute CSS state chaneges. However, I can't imagine a scenario where that would have any objective advantages over having the map store transition state. That solution would  be logically equivalent to the current one, but with state ownership being held by the popups instead of map. We can look into that if you can show a concrete case where the current approach would exhibit race conditions or unexpected behaviour.

mourner:
Superceded by #5743

Issue #3057, opened 2016-08-24
Improve readability of text-pitch-alignment: "viewport" text


mb12:
**mapbox-gl-js version**:Master
### Steps to Trigger Behavior
1. Tilt map using Shift + up-arrow key
2. Watch text along line.
3. Its extremely illegible. (See attached screenshot).

![illegible_text_screenshot](https://cloud.githubusercontent.com/assets/8890269/17917837/b723c2ce-6974-11e6-8ff4-f9c21994ffb6.png)

1ec5:
Setting `text-pitch-alignment` to `viewport` (#2668) untilts the labels, going a long way towards making the labels toward the back legible. The labels are still scaled by viewing distance, which would be addressed by a to-be-implemented `text-pitch-scale` property: https://github.com/mapbox/mapbox-gl-style-spec/issues/459#issuecomment-224957055.

The default Streets style doesn’t set `text-pitch-alignment` to `viewport`, in part because this functionality hadn’t been implemented by the time the current Streets style was published. However, you can use this property in your own custom styles.

mb12:
I am still seeing very high illegibility in text esp when placed on lines going vertically(screenshot below). 

![pitch_alignment_viewport](https://cloud.githubusercontent.com/assets/8890269/17935604/ce3860f8-69d0-11e6-9467-05d71c95c4ab.png)

1ec5:
That's related to the observations in https://github.com/mapbox/mapbox-gl-js/pull/2668#issuecomment-224113564 and the caveats in https://github.com/mapbox/mapbox-gl-js/pull/2668#issuecomment-224664159. Definitely agree there's room for improvement.

/cc @lucaswoj @yhahn

lucaswoj:
I agree. The readability of `text-pitch-alignment: "viewport"` is poor. We should look into improving it. 

Overall, our plans to address pitched text readability are captured by the `text-pitch-alignment` and `text-scale-alignment` style properties and this ticket.

jfirebaugh:
@ansis Was this fixed by #3790?

ansis:
@jfirebaugh I don't think so

ajashton:
[...]

PR #4682, opened 2017-05-04
WIP: Pitched label legibility

ChrisLoer:
Splitting changes to improve legibility of curved labels on tilted maps (#3057) out of https://github.com/mapbox/mapbox-gl-js/pull/4547.

The motivation for these changes is that we want to improve our rendering support for the "navigation" use case, in which:

- Tilting the map is important, because in turn by turn directions you want to be able to see ahead
- Legibility becomes more important relative to information density or aesthetics, because users of the map need to be able to read labels with a quick glance.

- [x] For curved labels with `pitch-alignment: viewport`: replace "foreshortening" behavior with "legibility expansion" that spaces vertical/tilted glyphs further apart in tile space so that they're still legible in projected space.
- [ ] Make "legibility expansion" use improved approximation math from https://gist.github.com/ChrisLoer/fb8c22845b6c3216afff0f19dc53f6d6
- [ ] Make collision box approximation take account of "legibility expansion"
- [ ] Pack attributes into as little space as possible
- [ ] Explore `max-curvature` property (or default) to prevent labels in the distance from ending up with silly geometry (e.g. "horseshoe labels")
- [ ] Verify `text-max-angle` logic (see `check_max_angle.js`) isn't broken


ChrisLoer:
From https://github.com/mapbox/mapbox-gl-js/pull/4547#issuecomment-294092835:

I got an initial implementation of the fancy trigonometry for label expansion... kind of working. It looks like it's doing a better job of keeping spacing even on curved labels, but there are two big problems:

* There's *lots* of glyph flicker/duplication around segment junctions. This might be caused by the placement-time calculations working with higher precision angles than the part of the calculation done in the shader (the positioning calculations only have to disagree by a tiny amount between two segments to cause artifacts). It looks bad enough it might just be a matter of me messing up the math.
* This approach ties the `minZoom` and `maxZoom` for a glyph to the "legibility expansion" calculated for that glyph... which breaks the `labelMinScale` trick we use at https://github.com/mapbox/mapbox-gl-js/blob/master/src/symbol/quads.js#L218 to make all the glyphs disappear if one of the glyphs has to disappear. To fix this, we'd need to add even more vertex attributes. 😞 

Overall, it's feeling like too complicated a solution, but I'll keep poking at it a bit longer to see if I can come up with anything compelling.

Here's a shot showing the algorithm actually doing its thing. This is the kind of label that does poorly with the simple algorithm: the center of the label is at about 45 degrees, while the left side is near vertical and the right side is near horizontal. As a result, the simple algorithm (on the left) squishes the left side and stretches out the right side. The fancy algorithm appears to do a reasonable job of spacing across the length of the label.

vistagrande improvements

ChrisLoer:
This work was superseded #4781.

Issue #4381, opened 2017-03-08
Use stylelint to enfore css prefixes


anandthakker:
To avoid regressions on fixes like https://github.com/mapbox/mapbox-gl-js/pull/4357, we should use https://github.com/stylelint/stylelint to enforce that every CSS class we define is prefixed with `mapbox`.

asantos3026:
@anandthakker I'm looking in to this issue. Is there a preference on how stylelint should be used? Here are the options for stylelint. 

```
Decide how you want to use stylelint:
via the stylelint CLI
via a plugin for your build tool (gulp, webpack etc)
via a plugin for your text editor (atom, sublime text etc)
via the stylelint Node API
via the stylelint PostCSS plugin
```

anandthakker:
@asantos3026 thanks for looking into it!

I think using the CLI is probably our best bet.  Using the stylelint CLI, we could add a `"lint-css"` script to [package.json](https://github.com/mapbox/mapbox-gl-js/blob/master/package.json#L116-L117), and then include that script in the [list of scripts run by `"test"`](https://github.com/mapbox/mapbox-gl-js/blob/master/package.json#L119)

asantos3026:
I was just reading this blog and I came to the same conclusion 👍  

`Usage Link
Stylelint can be used in many ways: in the command line, in a build tool (such as Gulp, Grunt or Webpack), in a code editor or as a Git pre-commit hook for staged changes in Git repository. I will focus here on two ways.
COMMAND LINE LINK
Using the command line is useful when you want to lint a project that doesn’t have stylelint or you want to use stylelint in an npm script.
Install stylelint globally:
npm install stylelint -g
Then, it will be available everywhere in your terminal:
stylelint "styles/**/*.css"
This command will lint all CSS files in the styles directory and any of its subdirectories.
`

asantos3026:
I'm new to developing and contributing to open source as part of my project this week :) I love Mapbox so wanted to contribute. Apologies for my novice question but for the style linter, is the declaration output in the css file supposed to be:

```
.mapbox-(class-name) {
  /* css code */
}
```

anandthakker:
@asantos3026 questions to clarify an issue are very welcome!

Yes, I think what you've got there looks right to me: every CSS class in our output file should be in the form `.mapbox-(blah-blah-blah)`

asantos3026:
Thanks @anandthakker. I added style lint and the naming convention seems to be working. How would I test?

[...]

Issue #685, opened 2014-08-11
Zoom levels 1 less in GL than raster?


ashleyclough:
It appears that the zoom levels reported in GL are 1 less than the same data source rendered as raster tiles. See the attached images of the same area as GL and in the Mapbox project view:

![gl zoom 12](https://cloud.githubusercontent.com/assets/912721/3878222/70bd0882-216d-11e4-8419-3d9abc3d5a07.png)
![raster zoom 13](https://cloud.githubusercontent.com/assets/912721/3878221/70b7b65c-216d-11e4-98d8-c574a75c8ae6.png)

rouen-sk:
Just found about this issue and I am really surprised. This is not a "minor documentation issue" - you practically decided that Mapbox GL **is not** slippy map, and **it is not compatible** with other standardized slippy map things - for example any server side code using standard zoom-to-scale-denominator (13 = 1:100000). You cant use one common datasource-level filtering based on zoom for both raster and vector tiles, you need different styling rules, etc. What a huge standard-breaker... This should be like first thing mentioned in the documentation - in big bold font.

timiyay:
Hmmmm, is there any chance this explains why I get incorrect results from `mapbox/geo-viewport`? https://github.com/mapbox/geo-viewport/issues/3

I use Mapbox's Static Image API, and I need to compute the bounding box of those images. Mapbox Support advised me that `geo-viewport` was written for this very purpose.

If I request a static image at MapboxGL zoom 17, and then ask `geo-viewport` to compute this bounding box, it gives the bbox for zoom 16 instead. I have to +1 to get the right answer.

I'm trying to learn enough about the underlying `mapbox/node-spherical-mercator` library to make a fix, but it sounds like it's mostly a misunderstanding on my part.

I'm honestly not fussed about the zoom matching with raster-based zooms. I just need a way to compute the bbox of a Mapbox Static Map API image.

lucaswoj:
I am closing this issue because the GL team feels 👍  about this decision and does not plan to revisit.

timiyay:
For posterity, the answer to my comment was:

Yes, the 512x512px tile size was the cause of my issue in https://github.com/mapbox/geo-viewport/issues/3

nrenner:
Documenting this in [Mapbox GL JS fundamentals](https://www.mapbox.com/help/mapbox-gl-js-fundamentals/) would be helpful.

PR #4685, opened 2017-05-05
fix update edit link, add query parameters for map feedback

### LINKED ISSUE #3558: Including a username in map-feedback attribution link
Similar to [this](https://github.com/mapbox/mapbox.js/issues/1187), It would be great if we could include a username in to the map-feedback links. This can be parsed out of the accessToken included with the map to understand the user better, irrespective of the map style they use. And hence will help us give proper support to them.

cc:// @tmcw

mollymerp:
This PR does two things:
- fix the edit link that we didn't know was broken as a result of #4381 – the HTML for the map feedback link comes from the TileJSON which mapbox-gl-js does not control, so when I changed the `mapboxgl` prefix for the `queryselector` class that modifies the edit link it stopped getting updated, and the test didn't catch it bc we provide the TileJSON html inside the test. 
- adds query parameters for #3558 

an alternative to this approach would be editing the tilejson in place to abide by our `mapboxgl-` prefix requirement.

closes #3558 

cc @tristen 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/control/attribution_control.js
+++ b/src/ui/control/attribution_control.js
@@ -68,13 +68,22 @@ class AttributionControl {
 
     _updateEditLink() {
         if (!this._editLink) this._editLink = this._container.querySelector('.mapbox-improve-map');
+        const params = [
+            {key: "pitch", value: this._map.getPitch()},
+            {key: "bearing", value: this._map.getBearing()},
+            {key: "owner", value: this.styleOwner},
+            {key: "id", value: this.styleId},
+            {key: "access_token", value: config.ACCESS_TOKEN}
+        ];
+
         if (this._editLink) {
             const center = this._map.getCenter();
-            const styleParams = (this.styleOwner && this.styleId) ? config.ACCESS_TOKEN ? `?owner=${this.styleOwner}&id=${this.styleId}&access_token=${config.ACCESS_TOKEN}` : `?owner=${this.styleOwner}&id=${this.styleId}` : '';
-            this._editLink.href = `https://www.mapbox.com/feedback/${styleParams}#/${
+            const paramString = params.reduce((acc, next, i) => acc += next.value !== undefined ? `${next.key}=${next.value}${i < params.length - 1 ? '&' : ''}` : '', `?`);
+            this._editLink.href = `https://www.mapbox.com/feedback/${paramString}#/${
                     Math.round(center.lng * 1000) / 1000}/${Math.round(center.lat * 1000) / 1000}/${Math.round(this._map.getZoom())}`;
         }
     }
+
 
     _updateData(e) {
         if (e && e.sourceDataType === 'metadata') {
--- a/src/ui/control/attribution_control.js
+++ b/src/ui/control/attribution_control.js
@@ -81,7 +81,7 @@ class AttributionControl {
                 }
                 return acc;
             }, `?`);
-            this._editLink.href = `https://www.mapbox.com/feedback/${paramString}${this._map.hash ? this._map.hash.getHashString() : ''}`;
+            this._editLink.href = `https://www.mapbox.com/feedback/${paramString}${this._map.hash ? this._map.hash.getHashString(true) : ''}`;
 
         }
     }
--- a/src/ui/control/attribution_control.js
+++ b/src/ui/control/attribution_control.js
@@ -69,11 +69,9 @@ class AttributionControl {
         if (!this._editLink) this._editLink = this._container.querySelector('.mapbox-improve-map');
         if (this._editLink) {
             const center = this._map.getCenter();
-            this._editLink.href = `https://www.mapbox.com/map-feedback/#/${
+            const styleParams = (this.styleOwner && this.styleId) ? `?owner=${this.styleOwner}&id=${this.styleId}` : '';
+            this._editLink.href = `https://www.mapbox.com/feedback/${styleParams}#/${
                     Math.round(center.lng * 1000) / 1000}/${Math.round(center.lat * 1000) / 1000}/${Math.round(this._map.getZoom())}`;
-            if (this.styleOwner && this.styleId) {
-                this._editLink.href += `?owner=${this.styleOwner}&id=${this.styleId}`;
-            }
         }
     }
 
--- a/src/ui/control/attribution_control.js
+++ b/src/ui/control/attribution_control.js
@@ -81,7 +81,7 @@ class AttributionControl {
                 }
                 return acc;
             }, `?`);
-            this._editLink.href = `https://www.mapbox.com/feedback/${paramString}${this._map.hash ? this._map.hash.getHashString(true) : ''}`;
+            this._editLink.href = `https://www.mapbox.com/feedback/${paramString}${this._map._hash ? this._map._hash.getHashString(true) : ''}`;
 
         }
     }
--- a/src/ui/control/attribution_control.js
+++ b/src/ui/control/attribution_control.js
@@ -69,18 +69,20 @@ class AttributionControl {
     _updateEditLink() {
         if (!this._editLink) this._editLink = this._container.querySelector('.mapbox-improve-map');
         const params = [
-            {key: "pitch", value: this._map.getPitch()},
-            {key: "bearing", value: this._map.getBearing()},
             {key: "owner", value: this.styleOwner},
             {key: "id", value: this.styleId},
             {key: "access_token", value: config.ACCESS_TOKEN}
         ];
 
         if (this._editLink) {
-            const center = this._map.getCenter();
-            const paramString = params.reduce((acc, next, i) => acc += next.value !== undefined ? `${next.key}=${next.value}${i < params.length - 1 ? '&' : ''}` : '', `?`);
-            this._editLink.href = `https://www.mapbox.com/feedback/${paramString}#/${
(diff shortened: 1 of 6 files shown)


mollymerp:
@tristen can you let me know if these query params (`owner` and `id`) are what you're looking for?
Scott confirmed that just the style id is not adequate to find the style associated bc ids are not globally unique, but owner+id pairs are.

tristen:
Ahh thanks for the clarification on capturing `owner`. What you are capturing looks great to me! 🎉

test/unit/ui/control/attribution.test.js
```diff
@@ -105,12 +107,12 @@ test('AttributionControl has the correct edit map link', (t) => {
     map.addControl(attribution);
 
     map.on('load', () => {
-        map.addSource('1', {type: 'vector', attribution: '<a class="mapboxgl-improve-map" href="https://www.mapbox.com/map-feedback/" target="_blank">Improve this map</a>'});
+        map.addSource('1', {type: 'vector', attribution: '<a class="mapbox-improve-map" href="https://www.mapbox.com/map-feedback/" target="_blank">Improve this map</a>'});
         map.on('data', (e) => {
             if (e.dataType === 'source' && e.sourceDataType === 'metadata') {
-                t.equal(attribution._editLink.href, 'https://www.mapbox.com/map-feedback/#/0/0/1', 'edit link contains map location data');
+                t.equal(attribution._editLink.href, 'https://www.mapbox.com/map-feedback/#/0/0/0?owner=mapbox&id=streets-v10', 'edit link contains map location data');
```
tristen: I think we can just omit these params if `owner` or `id` isn't set rather than passing defaults.

test/unit/ui/control/attribution.test.js
```diff
@@ -105,12 +107,12 @@ test('AttributionControl has the correct edit map link', (t) => {
...
                 map.setZoom(2);
-                t.equal(attribution._editLink.href, 'https://www.mapbox.com/map-feedback/#/0/0/3', 'edit link updates on mapmove');
+                t.equal(attribution._editLink.href, 'https://www.mapbox.com/map-feedback/#/0/0/2?owner=mapbox&id=streets-v10', 'edit link updates on mapmove');
```
tristen: One thing that needs to change! hash should follow params. Structure should look like this instead:

```
https://www.mapbox.com/feedback/?owner=mapbox&id=streets-v10#/0/0/2
```

tristen:
@mollymerp one last thing too: The path should change from `map-feedback` to `feedback`.

test/unit/ui/control/attribution.test.js
mollymerp: these aren't defaults – I set them up at L17-18 😛

tristen: omg I'm reading through the test 😭 😭 😭 😭

mollymerp:
@tristen I made the changes you requested. let us know when you need the release published by.

src/ui/control/attribution_control.js
```diff
@@ -66,11 +66,12 @@ class AttributionControl {
     }
 
     _updateEditLink() {
-        if (!this._editLink) this._editLink = this._container.querySelector('.mapboxgl-improve-map');
+        if (!this._editLink) this._editLink = this._container.querySelector('.mapbox-improve-map');
```
anandthakker: So, if I'm understanding this right, the change from `.mapboxgl-improve-map` to `.mapbox-improve-map` is in order to match against the CSS class used in our the TileJSON provided by Mapbox's vector tile sources?

src/ui/control/attribution_control.js
```diff
@@ -66,11 +66,12 @@ class AttributionControl {
...
         if (this._editLink) {
             const center = this._map.getCenter();
-            this._editLink.href = `https://www.mapbox.com/map-feedback/#/${
-                    center.lng}/${center.lat}/${Math.round(this._map.getZoom() + 1)}`;
+            const styleParams = (this.styleOwner && this.styleId) ? `?owner=${this.styleOwner}&id=${this.styleId}` : '';
+            this._editLink.href = `https://www.mapbox.com/feedback/${styleParams}#/${
+                    Math.round(center.lng * 1000) / 1000}/${Math.round(center.lat * 1000) / 1000}/${Math.round(this._map.getZoom())}`;
```
anandthakker: Why are we dropping the +1 from this._map.getZoom() + 1? Was it incorrect in the first place?

mollymerp: @anandthakker GL and raster zoom levels are off by one 😊

src/ui/control/attribution_control.js
mollymerp: correct @anandthakker - I will ticket out changing that css class at the API level separately

src/ui/control/attribution_control.js
mollymerp: https://github.com/mapbox/mapbox-gl-js/issues/685

src/ui/control/attribution_control.js
```diff
@@ -66,11 +66,12 @@ class AttributionControl {
...
 
     _updateEditLink() {
-        if (!this._editLink) this._editLink = this._container.querySelector('.mapboxgl-improve-map');
+        if (!this._editLink) this._editLink = this._container.querySelector('.mapbox-improve-map');
         if (this._editLink) {
             const center = this._map.getCenter();
-            this._editLink.href = `https://www.mapbox.com/map-feedback/#/${
-                    center.lng}/${center.lat}/${Math.round(this._map.getZoom() + 1)}`;
+            const styleParams = (this.styleOwner && this.styleId) ? `?owner=${this.styleOwner}&id=${this.styleId}` : '';
+            this._editLink.href = `https://www.mapbox.com/feedback/${styleParams}#/${
```
1ec5: If this map is using a custom style designed in Studio (as opposed to a default style), should the feedback tool display that style in its map? If so, we'll also need to pass in an access token or somehow make it so that the feedback tool can access any style.

/cc @tristen

tristen: @1ec5 yikes I think you brought up something pretty critical here. I assumed a token could be generated with the scope to include any custom style but that doesn't appear to be the case. It looks like an access token would need to be additionally passed for this to work.

@mollymerp `client` is encoded in the token so it might make sense to drop that parameter for `token` instead?

src/ui/control/attribution_control.js
1ec5: Is it still necessary to round the zoom level?

mollymerp:
@tristen Added access_token param!

1ec5:
Ported to iOS and macOS in mapbox/mapbox-gl-native#9078.

tristen:
@mollymerp I don't want to hold this PR much longer but we can we also capture bearing + pitch at the end of the hash?

mollymerp:
@tristen added it. note that the previous order I was using in the hash was different and also had an extra `/` compared to what we use as the existing map hash option in gl-js. (`#/lat/lng/zoom` vs `#zoom/lat/lng`) 😬 I imagine this was due to how the original map-feedback page was implemented. 

with the most recent change the new feedback page and mapbox gl hashes should remain consistent. let me know if this is 👌

mollymerp:
I made some decent changes here so @anandthakker if you want to take another 👀  that would probably be a good idea 😊

src/ui/control/attribution_control.js
```diff
@@ -69,7 +70,7 @@ class AttributionControl {
         if (!this._editLink) this._editLink = this._container.querySelector('.mapbox-improve-map');
         if (this._editLink) {
             const center = this._map.getCenter();
-            const styleParams = (this.styleOwner && this.styleId) ? `?owner=${this.styleOwner}&id=${this.styleId}` : '';
+            const styleParams = (this.styleOwner && this.styleId) ? config.ACCESS_TOKEN ? `?owner=${this.styleOwner}&id=${this.styleId}&access_token=${config.ACCESS_TOKEN}` : `?owner=${this.styleOwner}&id=${this.styleId}` : '';
```
anandthakker: `config.ACCESS_TOKEN` wouldn't pick up an access token that a user set using `mapboxgl.accessToken = ...` right?

mollymerp: @anandthakker I think it should https://github.com/mapbox/mapbox-gl-js/blob/78fc9c415a2d5bc71d9d10d8bdcf90c98387846a/src/index.js#L50

anandthakker: 🙈

src/ui/map.js
```diff
@@ -192,9 +192,9 @@ class Map extends Camera {
 
         bindHandlers(this, options);
 
-        this._hash = options.hash && (new Hash()).addTo(this);
+        this.hash = options.hash && (new Hash()).addTo(this);
```
anandthakker: IMO we should keep the `_` to signal that this is a private member rather than a public API.

mollymerp: oooh good point – for some reason I was thinking the convention was no `_` if its called from w/in another module but that is not correct 😊

mollymerp:
We good to pull the trigger on this @tristen @anandthakker ?

tristen:
@mollymerp Looks good to merge from my end!

PR #4686, opened 2017-05-08
clarify icon-size units, fix #4661

### LINKED ISSUE #4661: Document "icon-offset" units
`icon-offset` units are strange https://github.com/mapbox/mapbox-gl-js/issues/4078 - they are `px * icon-size`. This is not documented, and could potentially lead to somebody (me) spending lots of time figuring out what is going wrong.

Until https://github.com/mapbox/mapbox-gl-js/issues/4078 is done, it would be good to document current behaviour.

ansis:
@gmaclennan thanks for opening #4661. Does this clarify the units enough?

"units": "factor of the original icon size"
"doc": "Scales the original size of the icon by the provided factor. The new pixel size of the image will be `original pixel size * 'image-size'`. 1 is the original size, 3 triples the size of the image."

## Launch Checklist

 - [N/A] write tests for all new functionality
 - [N/A] post benchmark scores

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -811,7 +811,8 @@
       "function": "interpolated",
       "zoom-function": true,
       "property-function": true,
-      "doc": "Scale factor for icon. 1 is original size, 3 triples the size.",
+      "units": "factor of the original icon size",
+      "doc": "Scales the original size of the icon by the provided factor. The new pixel size of the image will be the original pixel size multiplied by `icon-size`. 1 is the original size; 3 triples the size of the image.",
       "requires": [
         "icon-image"
       ],


gmaclennan:
Looks good to me!

src/style-spec/reference/v8.json
```diff
@@ -811,7 +811,8 @@
       "function": "interpolated",
       "zoom-function": true,
       "property-function": true,
-      "doc": "Scale factor for icon. 1 is original size, 3 triples the size.",
+      "units": "factor of the original icon size",
+      "doc": "Scales the original size of the icon by the provided factor. The new pixel size of the image will be `original pixel size * 'image-size'`. 1 is the original size, 3 triples the size of the image.",
```
jfirebaugh: Tweak the second sentence so that only `image-size` is in backticks. This ensures that the macOS/iOS API documentation generator will replace "image-size" with the SDK-appropriate property. cc @1ec5

ansis: Tweaked: 		efc6feb

1ec5: In fe46990085f39f72bd55d168d5c21b778818474b, I replaced the `*` with “multiplied by”. The iOS/macOS SDK’s codegen script hard-wraps the documentation. I wasn’t sure whether the `*` could end up at the beginning of a line, where jazzy would interpret it as a bulleted list. Spelling it out should help the text flow in any case.

PR #4688, opened 2017-05-08
Ensure symbol shader attrib bound at 0 is always valid and used

### LINKED ISSUE #4607: Safari 10 / OS X (Retina) sometimes doesn't render symbol layers
**mapbox-gl-js version**:
0.35.1
### Steps to Trigger Behavior

 1. Open mapbox gl js on safari with init level zoom < 2
 2. Zoom to lower level.
 3. Reload 1 more time name of all countries and streets don't show on map (issue just happen on safari)

### Expected Behavior
All name of streets and coutries will show for all init level zoom on safari.
### Actual Behavior
All name of streets and coutries didn't show for all init level zoom < 2 on safari.

lbud:
Fixes #4607.
@jfirebaugh found the culprit [in Firefox](https://github.com/mapbox/mapbox-gl-js/issues/4607#issuecomment-299601225): binding an attrib to position 0 without using it "forces the browser to do expensive emulation work when running on desktop OpenGL platforms, for example on Mac." In this case, `a_size` was frequently being bound to position 0 and was not always used. This PR manually binds `a_data` (which is always used — either this or `a_pos_offset` would work) to position 0, and then lets the program dynamically bind the rest.

### DIFF
--- a/src/render/painter.js
+++ b/src/render/painter.js
@@ -393,27 +393,24 @@ class Painter {
         assert(gl.getShaderParameter(vertexShader, gl.COMPILE_STATUS), gl.getShaderInfoLog(vertexShader));
         gl.attachShader(program, vertexShader);
 
-        const result = {program};
 
         // For the symbol program, manually ensure the attrib bound to position 0 is always used (either a_data or a_pos_offset would work here).
         // This is needed to fix https://github.com/mapbox/mapbox-gl-js/issues/4607 — otherwise a_size can be bound first, causing rendering to fail.
         // All remaining attribs will be bound dynamically below.
         if (name === 'symbolSDF') {
             gl.bindAttribLocation(program, 0, 'a_data');
-            result['a_data'] = 0;
         }
 
         gl.linkProgram(program);
         assert(gl.getProgramParameter(program, gl.LINK_STATUS), gl.getProgramInfoLog(program));
 
         const numAttributes = gl.getProgramParameter(program, gl.ACTIVE_ATTRIBUTES);
+        const result = {program, numAttributes};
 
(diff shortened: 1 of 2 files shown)


lbud:
@jfirebaugh I haven't dug too far into it yet, but it appears it's not entering that branch because the `OES_vertex_array_object` (`gl.extVertexArrayObject` here) is present. Moreover, I'm not 100% clear on how the shader binding works but it appears that attrib positions are already determined by this point, having been dynamically/automatically assigned in painter.js.

It also appears that in Chrome on `master` when we rely on webgl to dynamically assign attribute locations (I think this happens in `gl.linkProgram`…?), `a_data` already happens to be assigned to position 0, whereas on Safari `a_size` is assigned to position 0, so I'm not sure that we'd even be able to reliably create a test for this, since we can't control browser environments on CI 🤕 

It'd probably be a good idea to eventually take the same approach as native with explicit bindings.

kkaefer:
> I think this happens in gl.linkProgram…?

Yeah, if you don't define explicit bindings, it auto-assigns them when linking.

> I'm not sure that we'd even be able to reliably create a test for this, since we can't control browser environments on CI

This is pretty hard to test. It only happens for me every 5-10 reloads of the map. When the labels are vanishing, the map is also really slow. Since this is a race condition we're trying to remove, I don't think there's a good way to test this. Even if we had a test, we'd only catch the error in a very limited number of test runs.

Longer term, it might make more sense to force explicit bindings, similar to how GL native does it.

dagjomar:
Anybody have time to review this? Waiting for the specific bugfix so we can upgrade the version :)

PR #4693, opened 2017-05-09
Clarify which properties can use exponential and interval functions in docs

samanpwbb:
Small fix to docs, @jfirebaugh for review. I went with "requires a property that can be interpolated" instead of "requires an interpolatable property" because "interpolated" is a keyword from the spec JSON that people might be searching for on the page.

### DIFF
--- a/docs/style-spec/_generate/index.html
+++ b/docs/style-spec/_generate/index.html
@@ -849,9 +849,16 @@ navigation:
                 <dt><var>identity</var></dt>
                 <dd>functions return their input as their output.</dd>
                 <dt><var>exponential</var></dt>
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Ah shoot, I misunderstood your question and misread the docs. "The domain must be numeric" is actually correct. Note: _domain_, not range. Both interval and exponential functions require numeric domain values -- they are either zoom functions, or property or zoom-and-property functions with numeric property values.

jfirebaugh:
We should say something about the range/output type though.

For exponential:

> The domain (input value) must be numeric, and the style property must support interpolation. Style properties that support interpolation are marked with ◢, and `exponential` is the default function type for these properties.

For interval:

> The domain (input value) must be numeric. Any property may use interval functions. For properties marked with 📶, the "interval" symbol, this is the default function type.

Issue #1489, opened 2015-09-16
Support 3D terrain meshes


trallarn:
Thank you guys for a great job so far. Do you have any plans on adding rendering support of 3D terrain meshes like google maps earth and cesiumjs are doing?

Example:
http://betaserver.icgc.cat/cesium/Girona3D.html

dpieri:
I'm curious about the technical feasibility of this. Having true 3D ability would be huge for us and we are willing to contribute developer resources towards making it happen. Is that a pipe dream, or could this feature be implemented in a reasonable timeframe?

mourner:
We are definitely going to take on the challenge of implementing 3D terrain for GL JS eventually. It wasn't a priority so far though compared to other pushes such as data-driven styling. 

It's certainly a complicated task that can take several months of effort to get right, both on the client and the server side. @ansis do you have any ballpark thoughts on the effort/challenges involved here?

ideamotor:
FWIW, we're looking at solutions for this right now and this would likely bring us over to Mapbox. Particularly with the Google Earth API deprecated, I suspect we're not alone in this.

Chris

dagatsoin:
A good source of inspiration could be www.eegeo.com

Since Mapbox GL uses OpenGL it seams faisable, does not? But I see two different case in 3D render:
- based on aerial scan data like the exemple above
- based on actual mapbox data driven styling.

The most ins the scope is the second option since all render pipeline could be reused.

**Terrain elevation**
A solution could be implement a runtime tessellation process based on elevation terrain frequency. Depending on the render pipeline, it could be:

- implemented on a base shader which from are inherited all the existing vertex shaders.
- implemented in a pass after the vertex shaders by collapsing the stack in a vertex buffer, then tesselate the new mesh.

**Display 3D meshes**
In my opinion (without really knowing the render pipeline) it could be more trivial here. Houses, cars, characters could be classical .fbx with a texture loading as needed. Also, I don't know if Z-buffer is used in actual implementation.
However I am curious to know what could be the more efficient since entering the 3D world could lead to infinite wish lists until MaboxGL will be like Unity3D or Unreal Engine... What should be the MapboxGL scope for 3D? Mesh animation, particules, collisions?
Based on that does it make sense to develop:
- the next version of MGL in full 3D?
- a plugin for Three.js, SceneKit, Unity3D, etc. ?

mourner:
[...]

PR #4701, opened 2017-05-11
[not ready] client side hillshading with terrain-rgb raster tiles

mollymerp:
still very very rough here. will update this ticket soon w the to-do list before this is ready for review. 

🗻
 

## Launch Checklist

 - [ ] post benchmark scores


andrewharvey:
Would it make more sense to call the layer type `hillshade`? I think we should try to use terminology which fits in with future 3D terrain support. I've proposed some ideas about terminology over at https://github.com/mapbox/mapbox-gl-js/issues/1489#issuecomment-287611953. Over there I've proposed using `terrain` as the source type for terrain-rgb tiles, and `elevation` as the layer type for 3D surface.

mollymerp:
Here's a rough list of what needs to happen before this can be merged into master:
- [ ] move mip map level creation + resampling to a worker thread
- [ ] fix tile seams with better design for backfilling tile borders
- [ ] get blending to work with other layers
- [ ] clean up shader debug code
- [ ] reduce shader attribute counts for compatibility with native
- [ ] implement light to behave consistently with extrusions
- [ ] finalize layer+source properties in the style spec #4711

kkaefer:
We could move the preparation step for terrain textures to the beginning of a frame to improve rendering performance.

Right now, the workflow of a frame is like this:
* (main framebuffer is bound)
* Clear main framebuffer (✅ first operation is clear -> no restore)
* Render some layers
* Render terrain layer
  * For non-preprocessed tiles
    * **Bind terrain framebuffer**
    * Clear terrain framebuffer (✅ first operation is clear -> no restore)
    * Draw prepare step
    * **Bind main framebuffer**
  * Draw preprocessed terrain tiles
* Draw other layers (:warning: FBO restore)

When binding a framebuffer and rendering operations on it, the GPU has to restore the contents of the framebuffer. The driver can optimize that when we do a clear as the first drawing operation (since that discards the contents of the framebuffer). Therefore, drawing to the main framebuffer, then binding a *different* framebuffer, and rebinding the main framebuffer triggers a framebuffer load on the first draw call.

Instead, we could do all operations that require binding a separate framebuffer (computing the preprocessed textures) as the first thing when rendering a frame (i.e. before calling clear on the main framebuffer) and have the following order:

* (main framebuffer is bound)
* Preprocess all non-preprocessed terrain tiles (if any)
  * **Bind terrain framebuffer**
  * Clear terrain framebuffer (✅ first operation is clear -> no restore)
  * Draw prepare step
  * **Bind main framebuffer**
* Clear main framebuffer (✅ first operation is clear -> no restore)
* Render some layers
* Render terrain layer
  * Draw preprocessed terrain tiles
* Draw other layers

kkaefer:
We'd likely need something like https://github.com/mapbox/mapbox-gl-native/issues/9415 to prevent flickering from overlapping parent/child tiles.

src/source/image_source.js
```diff
@@ -142,12 +142,11 @@ class ImageSource extends Evented {
...
         this.tiles[tile.coord.w] = tile;
-        const maxInt16 = 32767;
         const array = new RasterBoundsArray();
         array.emplaceBack(this._tileCoords[0].x, this._tileCoords[0].y, 0, 0);
-        array.emplaceBack(this._tileCoords[1].x, this._tileCoords[1].y, maxInt16, 0);
-        array.emplaceBack(this._tileCoords[3].x, this._tileCoords[3].y, 0, maxInt16);
-        array.emplaceBack(this._tileCoords[2].x, this._tileCoords[2].y, maxInt16, maxInt16);
+        array.emplaceBack(this._tileCoords[1].x, this._tileCoords[1].y, EXTENT, 0);
+        array.emplaceBack(this._tileCoords[3].x, this._tileCoords[3].y, 0, EXTENT);
+        array.emplaceBack(this._tileCoords[2].x, this._tileCoords[2].y, EXTENT, EXTENT);
```
mollymerp: yep -- these changes are in #4770 and https://github.com/mapbox/mapbox-gl-native/pull/9153, but they're blocked until text-pitch changes are merged into native

mollymerp:
this work is now tracked at #5286

Issue #4079, opened 2017-02-01
Allow developers to specify a default/fallback value for a token when a feature property is undefined


lucaswoj:
_From @kkaefer on July 16, 2014 0:16_

For certain tokens, a default string might be useful, e.g. when pulling the icon name from a feature, there may be features with that property not set. In those cases, it'd be cool to specify a default value for a token, for example:

`"{maki:generic}-12"` would generate "cafe-12" when the feature has the property `maki` set to `"cafe"`, and "generic-12" when that property isn't set.

_Copied from original issue: mapbox/mapbox-gl-style-spec#104_

lucaswoj:
_From @ansis on July 16, 2014 0:47_

Maybe the default values should be in a separate object so that the string parsing doesn't become more complicated.

lucaswoj:
_From @yhahn on July 16, 2014 11:53_

cc @ajashton trying to remember all the things we've run into with this

lucaswoj:
_From @ajashton on July 16, 2014 17:40_

The cases where I've most wanted fallbacks are not usually to a static string but to an alternate column or multiple alternate columns.
- Label translations without duplicated data in the vector tiles: `"text-field": "{name_en:name}"`
- POI icons with of optional custom rail network icons in a single layer: `"icon-image": "{network:maki}-12"`
- Maybe also make string fallbacks possible: `"{network:maki:'generic'}-12"`

lucaswoj:
_From @kkaefer on July 16, 2014 17:52_

Maybe we can do recursive replacements, like `{network:{maki}}`?

lucaswoj:
_From @kkaefer on July 17, 2014 19:0_

@ajashton how do you like that suggestion? `"{network:maki:'generic'}-12"` could be `"{network:{maki:generic}}-12"`

lucaswoj:
_From @ajashton on July 17, 2014 19:6_

Seems good to me

lucaswoj:
_From @ajashton on November 25, 2015 21:52_

Re-reading this discussion it sounds like we're mostly thinking about fallbacks for situations where the token field is null/unset in the data. But another important situation when working with sprites is being able to fall back when the token field contains an unexpected value.

Using the `{maki:'generic'}-12` example, could it be possible to fall back to `'generic'` if the `maki` value is `'cafe'` but the current sprite does not contain an icon for cafe? This would also be extremely useful.

lucaswoj:
[...]

Issue #4077, opened 2017-02-01
Allow arbitrary expressions in functions and filters


lucaswoj:
_From @nickidlugash on June 16, 2014 20:1_

Will it be possible to support `+`, `-`, `*`, and `/`?

_Copied from original issue: mapbox/mapbox-gl-style-spec#47_

lucaswoj:
_From @tmcw on June 16, 2014 20:21_

Could you go into a little more detail about how these would be represented? Like `"{{a}} + {{b}}"`?

Fwiw, expressions like this _are_ probably super useful, but I'd be interested in whether the approach we've taken for filters could be applied rather than doing infix like Carto or string like Mapnik - like we would have something like

``` js
{
  "+": ["{{a}}", 2]
}
```

Instead of

``` js
"{{a}} + 2"
```

lucaswoj:
_From @jfirebaugh on January 22, 2015 1:10_

Can we enumerate some uses cases for this feature? I'm particularly interested in how much it overlaps with #226 (Stop functions based on arbitrary properties).

lucaswoj:
_From @nickidlugash on January 22, 2015 22:40_

hrm. In theory, there are some use cases for this not covered by stop functions based on arbitrary properties. I might not be fully understanding the stop functions, but they seem to mainly be for assigning different values of the same style property for different values of the same field in a data layer. With math operators, you could manipulate values across different properties and different GL layers. 

A use case that I think would overlap with a use case for https://github.com/mapbox/mapbox-gl-style-spec/issues/226 is using math operators in conjunction with tokens, e.g. for building heights. 

The two main non-overlapping use cases I had in mind for this were for creating systems of line widths and font sizes across different GL layers. These could potentially be really useful for creating template styles that are fast to customize. However, these use cases require the ability to use constant values in functions, which you currently can't do in the editor.

lucaswoj:
_From @danpat on February 18, 2016 19:1_

I have a specific use case.  I'm generating vector tiles from OSRM that contain speed values on roads, in km/h.  It would be nice to show localized values (`mph` vs `km/h`) by doing some simple math in the style.

I'm ambivalent about the syntax, but I've found this capability in Mapnik hugely useful.

Other basic math functions, like `round()` would also be helpful in my case.

lucaswoj:
_From @timrobertson100 on March 12, 2016 15:49_

A use case:

[...]

Issue #4704, opened 2017-05-11
the future of labelling


ansis:
@ChrisLoer, @kkaefer and I talked about making large changes to text rendering and label placement.

## Problems

current problems:
- labels aren't legible in pitched views (https://github.com/mapbox/mapbox-gl-js/pull/4682)
- labels have collisions across sources (#2703)
- labels have collisions across tiles in pitched view
- labels aren't faded across tiles or on rotation
- collision prevention code is hard to expand to account for new features like [`text-pitch-scaling`](https://github.com/mapbox/mapbox-gl-js/pull/4547)

future problems:
- collision prevention for labels offset in the z direction (a label on top of a building)
- rendering and collision prevention for labels following a 3D path (a label for a road on 3D terrain)

---

Labelling has two subproblems:
- **placement**: figuring out where you want to put labels and avoiding collisions
- **rendering**: actually drawing glyphs where you want them

# Rendering

## Up until now
We're still using the [sliding-glyph-for-each-segment-that-gets-toggled-on-and-off](https://camo.githubusercontent.com/5e1f9d174172207ae5a840dfae2ffec8b420661a/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f313432313635322f313535343730382f62373964623831342d346534622d313165332d393864612d3438646233336438353463652e676966) approach ([original devlog](https://github.com/mapbox/hey/issues/1118)). This approach really only works if labels are always rendered on the map plane. [`text-pitch-alignment`](https://github.com/mapbox/mapbox-gl-js/pull/2668) was added to make labels more legible in pitched views. @ChrisLoer's [recent work](https://github.com/mapbox/mapbox-gl-js/pull/4682) improves on this. Both of these `text-pitch-alignment` implementations are workarounds that work fairly well but have some problems because of the limitations of the original implementation.

## Using a different approach

I think we all agreed that this approach is getting hard to adapt to our needs and is getting harder to understand. We should try something completely different. If we have access to the full line geometry when doing the vertex transform it gets a lot easier and more accurate. The rough algorithm:

1. Project the line into the plane you want to put the text on (map plane, viewport plane)
2. Iterate along the line to find a point the correct distance along the line.
3. Offset the corner of the glyph from this point on the line.
4. Project the resulting position into GL coords.

We need to figure out a way to do this fast enough that we can do it on every frame.

### CPU version

[...]

Issue #4558, opened 2017-04-07
Collision detection doesn't fully account for label zoom functions


ChrisLoer:
When we do collision detection for a tile, we calculate the size of labels at zoom level `tileZoom + 1`, and then assume that the size of the labels will scale directly with the zoom level in the range between `tileZoom` and `tileZoom + 1`. This allows us to calculate a single "minimum zoom" for each label and pass that value to the shader, so that the shader can know whether or not to show a particular glyph/symbol.

With a `text-size` or `icon-size` zoom function, this can break down in a couple ways:

* If a zoom function makes the label smaller as the zoom increases, the label can be considered a "collision" prematurely. Within a single zoom level, this isn't so bad -- we're just being more conservative with label placement. However, it becomes a problem when you reach the next higher zoom level, because when the collision boxes get recalculated at the new zoom level, the label will start showing again. With a very reasonable zoom function that decreases text-size as the zoom level increases, this could cause lots of label flicker on crossing integer zoom levels.
* Zoom functions that make the label bigger as the zoom increases can lead to label collisions. Also, depending on how quickly the label increased size, we might need a "maximum zoom" for the label instead of a "minimum zoom". I don't see why you'd want to do this for a simple zoom function, but it might inadvertently result from the calculations for a composite zoom-and-property function.
* A zoom function that's not monotonically increasing or decreasing with zoom is fundamentally incompatible with the whole "minimum zoom" approach.

I think the first case could be fixed with some complicated but not necessarily prohibitively expensive logic in `CollisionTile#getPlacementScale`. Right now, to determine the "min zoom" for a new label, we compare it to each already placed label it may touch, and against each new label we solve to find the scale at which their sides touch. It's a very easy solution, because the sizes are all linear in scale. In the new logic, we'll have to do the algebra to solve for where two different zoom (and property) functions intersect.

The [proposed changes to handling label size on pitched maps](https://github.com/mapbox/mapbox-gl-js/issues/4120) make this even a little more complicated, because labels will change size dynamically based on their position at render-time. The current proposal for handling that case is to treat distant pitched tiles as if they're at a much lower zoom level -- but that only works because the "minimum zoom" functionality *doesn't* account for label size zoom functions.

[...]

PR #4715, opened 2017-05-16
[for discussion] Draft proposal - arbitrary expressions for style functions

anandthakker:
Starting work here on a draft/strawman implementation for adding arbitrary expressions to style functions.  This first attempt attempts to synthesize the fruits of a few long-running discussions on the topic, in particular, https://github.com/mapbox/mapbox-gl-function/issues/28, https://github.com/mapbox/mapbox-gl-js/issues/4077, and https://github.com/mapbox/mapbox-gl-js/issues/4079.

## Important use cases that the final design should cover:
- unit conversion
- number formatting
  - decimal precision
  - thousands separator
- conditionals (for the token defaults use case)
- concatenation (for the token defaults use case)
- customizable interpolation
- arithmetic on feature properties within functions
- localization

## Expression-based style functions

Proposed plan:
 - deprecate zoom+property functions
 - deprecate property functions
 - keep stop-based zoom functions
 - allow data-driven expressions (a) as style property values, and (b) as output values for zoom function stops.  (I.e., (a) replaces "property functions", and (b) replaces zoom+property functions.)

See `docs/style-spec/expressions.md` for the spec details.

See `debug/expressions.html` for a couple example usages.

### Things we should design into the initial implementation
- [x] a way to map exact values to subexpressions, replacing type: categorical **currently proposed as `match`**
- [x] a way to match a value to a corresponding interval, replacing type: interval **currently proposed as a combination of `switch` and `between`**
- [x] a way to easily interpolate values linearly
- [x] number to string with specified number of decimal places (`1.2345` to `"1.23"`) **possible with `round`**

### Things we could add support for later
- a way to easily interpolate values exponentially, replacing type: exponential
- number to localized string (`21000` to `"21,000"` or `"21 000"` depending on locale)
- methods for converting between distance units (metres, feet, kilometres, miles)
- some way of localizing text by choosing the right name property
- time formatting/localization? temperature? money?

### Questions/issues
 - [ ] Should we cover image fallbacks as part of expression evaluation, or separately?

## Additional ideas
- [ ] Using something like http://jsep.from.so/ ( h/t @1ec5 ), it should be quite straightforward to parse expression text into the AST structure outlined above.
   - This may be useful both for the studio interface (if that's the route @mapbox/studio wants to go) and as a convenience alternative in the GL JS API (i.e. allowing something like `circle-radius: { type: "expression", expression: "(1 + properties.foo) * 2^map.zoom)" }` when specifying styles via the runtime api.)

cc @mapbox/gl

### DIFF
--- a/debug/expressions.html
+++ b/debug/expressions.html
@@ -8,14 +8,33 @@
     <style>
         body { margin: 0; padding: 0; }
         html, body, #map { height: 100%; }
+        #map, #chart { width: 49%; display: inline-block; }
+
+        path { 
+            stroke: steelblue;
+            stroke-width: 2;
+            fill: none;
+        }
+
+        .axis path,
+        .axis line {
+            fill: none;
+            stroke: grey;
+            stroke-width: 1;
+            shape-rendering: crispEdges;
+        }
     </style>
 </head>
 
 <body>
 <div id='map'></div>
+<div id='chart'></div>
 
 <script src='/dist/mapbox-gl-dev.js'></script>
 <script src='/debug/access_token_generated.js'></script>
+
+<script src="http://d3js.org/d3.v3.min.js"></script>
+
 <script>
 
 var style = {
@@ -48,7 +67,8 @@ var style = {
                             [ 'number_data', 'x' ],
                             [ 'number_data', 'y' ]
                         ]
-                    ]
+                    ],
+                    zoomInterpolationBase: 1
                 },
                 'circle-opacity': {
                     type: 'expression',
@@ -77,6 +97,87 @@ var map = window.map = new mapboxgl.Map({
     hash: true
 });
 
+// Set the dimensions of the canvas / graph
+var margin = {top: 30, right: 20, bottom: 30, left: 50},
+    width = 600 - margin.left - margin.right,
+    height = 270 - margin.top - margin.bottom;
+
+// Set the ranges
+var x = d3.scale.linear().range([0, width]);
+var y = d3.scale.linear().range([height, 0]);
+
+// Define the axes
+var xAxis = d3.svg.axis().scale(x)
+    .orient("bottom").ticks(5);
+
+var yAxis = d3.svg.axis().scale(y)
+    .orient("left").ticks(5);
+
+// Define the line
+var valueline = d3.svg.line()
+    .x(function(d) { return x(d.x); })
+    .y(function(d) { return y(d.y); });
+    
+// Adds the svg canvas
+var svg = d3.select("#chart")
+    .append("svg")
+        .attr("width", width + margin.left + margin.right)
+        .attr("height", height + margin.top + margin.bottom)
+    .append("g")
+        .attr("transform", 
+              "translate(" + margin.left + "," + margin.top + ")");
+
+map.on('load', function () {
+    const styleFunction = map.getPaintProperty('circle', 'circle-radius')
+    var compiledExpression = mapboxgl.createFunction(styleFunction, {});
+
+    const data = [];
+    for (let z = 0; z < 5; z += 0.01) {
+        let output
+        if (!compiledExpression.isFeatureConstant && !compiledExpression.isZoomConstant) {
+            const zLow = Math.floor(z);
+            const zHigh = Math.ceil(z);
+            const interp = mapboxgl.createFunction({
+                type: 'exponential',
+                stops: [ [ zLow, 0 ], [ zHigh, 1 ] ],
+                base: compiledExpression.zoomInterpolationBase
+            }, { type: 'number' })
+            const t = interp(z);
+            const outLow = compiledExpression(zLow, {x: 5, y: 5});
+            const outHigh = compiledExpression(zHigh, {x: 5, y: 5});
+            output = outLow + (outHigh - outLow) * t;
+        } else {
+            output = compiledExpression(z, {x: 5, y: 5});
+        }
+        data.push({x: z, y: output});
+    }
+
+    initChart(data);
+})
+
+// Get the data
+function initChart(data) {
+    // Scale the range of the data
+    x.domain(d3.extent(data, function(d) { return d.x; }));
+    y.domain([0, d3.max(data, function(d) { return d.y; })]);
+
+    // Add the valueline path.
+    svg.append("path")
+        .attr("class", "line")
+        .attr("d", valueline(data));
+
+    // Add the X Axis
+    svg.append("g")
+        .attr("class", "x axis")
+        .attr("transform", "translate(0," + height + ")")
+        .call(xAxis);
+
+    // Add the Y Axis
+    svg.append("g")
+        .attr("class", "y axis")
+        .call(yAxis);
+
+};
 </script>
 </body>
 </html>
--- a/debug/expressions.html
+++ b/debug/expressions.html
@@ -1,0 +1,86 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Mapbox GL JS debug page</title>
+    <meta charset='utf-8'>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+    <link rel='stylesheet' href='/dist/mapbox-gl.css' />
+    <style>
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+
+<body>
+<div id='map'></div>
+
+<script src='/dist/mapbox-gl-dev.js'></script>
+<script src='/debug/access_token_generated.js'></script>
+<script>
+
+var style = {
+    sprite: "mapbox://sprites/mapbox/streets-v10",
+    glyphs: "mapbox://fonts/mapbox/{fontstack}/{range}.pbf",
+    sources: {
+        "geojson": {
+            "type": "geojson",
+            "data": {
+                "type": "FeatureCollection",
+                "features": [{"type":"Feature","properties":{"x":-10,"y":-10},"geometry":{"type":"Point","coordinates":[-10,-10]}},{"type":"Feature","properties":{"x":-10,"y":-6},"geometry":{"type":"Point","coordinates":[-10,-6]}},{"type":"Feature","properties":{"x":-10,"y":-2},"geometry":{"type":"Point","coordinates":[-10,-2]}},{"type":"Feature","properties":{"x":-10,"y":2},"geometry":{"type":"Point","coordinates":[-10,2]}},{"type":"Feature","properties":{"x":-10,"y":6},"geometry":{"type":"Point","coordinates":[-10,6]}},{"type":"Feature","properties":{"x":-10,"y":10},"geometry":{"type":"Point","coordinates":[-10,10]}},{"type":"Feature","properties":{"x":-6,"y":-10},"geometry":{"type":"Point","coordinates":[-6,-10]}},{"type":"Feature","properties":{"x":-6,"y":-6},"geometry":{"type":"Point","coordinates":[-6,-6]}},{"type":"Feature","properties":{"x":-6,"y":-2},"geometry":{"type":"Point","coordinates":[-6,-2]}},{"type":"Feature","properties":{"x":-6,"y":2},"geometry":{"type":"Point","coordinates":[-6,2]}},{"type":"Feature","properties":{"x":-6,"y":6},"geometry":{"type":"Point","coordinates":[-6,6]}},{"type":"Feature","properties":{"x":-6,"y":10},"geometry":{"type":"Point","coordinates":[-6,10]}},{"type":"Feature","properties":{"x":-2,"y":-10},"geometry":{"type":"Point","coordinates":[-2,-10]}},{"type":"Feature","properties":{"x":-2,"y":-6},"geometry":{"type":"Point","coordinates":[-2,-6]}},{"type":"Feature","properties":{"x":-2,"y":-2},"geometry":{"type":"Point","coordinates":[-2,-2]}},{"type":"Feature","properties":{"x":-2,"y":2},"geometry":{"type":"Point","coordinates":[-2,2]}},{"type":"Feature","properties":{"x":-2,"y":6},"geometry":{"type":"Point","coordinates":[-2,6]}},{"type":"Feature","properties":{"x":-2,"y":10},"geometry":{"type":"Point","coordinates":[-2,10]}},{"type":"Feature","properties":{"x":2,"y":-10},"geometry":{"type":"Point","coordinates":[2,-10]}},{"type":"Feature","properties":{"x":2,"y":-6},"geometry":{"type":"Point","coordinates":[2,-6]}},{"type":"Feature","properties":{"x":2,"y":-2},"geometry":{"type":"Point","coordinates":[2,-2]}},{"type":"Feature","properties":{"x":2,"y":2},"geometry":{"type":"Point","coordinates":[2,2]}},{"type":"Feature","properties":{"x":2,"y":6},"geometry":{"type":"Point","coordinates":[2,6]}},{"type":"Feature","properties":{"x":2,"y":10},"geometry":{"type":"Point","coordinates":[2,10]}},{"type":"Feature","properties":{"x":6,"y":-10},"geometry":{"type":"Point","coordinates":[6,-10]}},{"type":"Feature","properties":{"x":6,"y":-6},"geometry":{"type":"Point","coordinates":[6,-6]}},{"type":"Feature","properties":{"x":6,"y":-2},"geometry":{"type":"Point","coordinates":[6,-2]}},{"type":"Feature","properties":{"x":6,"y":2},"geometry":{"type":"Point","coordinates":[6,2]}},{"type":"Feature","properties":{"x":6,"y":6},"geometry":{"type":"Point","coordinates":[6,6]}},{"type":"Feature","properties":{"x":6,"y":10},"geometry":{"type":"Point","coordinates":[6,10]}},{"type":"Feature","properties":{"x":10,"y":-10},"geometry":{"type":"Point","coordinates":[10,-10]}},{"type":"Feature","properties":{"x":10,"y":-6},"geometry":{"type":"Point","coordinates":[10,-6]}},{"type":"Feature","properties":{"x":10,"y":-2},"geometry":{"type":"Point","coordinates":[10,-2]}},{"type":"Feature","properties":{"x":10,"y":2},"geometry":{"type":"Point","coordinates":[10,2]}},{"type":"Feature","properties":{"x":10,"y":6},"geometry":{"type":"Point","coordinates":[10,6]}},{"type":"Feature","properties":{"x":10,"y":10},"geometry":{"type":"Point","coordinates":[10,10]}}]
+            }
+        }
+    },
+    layers: [
+        {
+            id: 'circle',
+            type: 'circle',
+            source: 'geojson',
+            paint: {
+                'circle-radius': {
+                    type: 'expression',
+                    expression: [
+                        '*',
+                        [ '^', 2, {ref: 'map', key: 'zoom'} ],
+                        0.125,
+                        [
+                            '+',
+                            20,
+                            {ref: 'feature', key: 'x'},
+                            {ref: 'feature', key: 'y'}
+                        ]
+                    ]
+                },
+                'circle-opacity': {
+                    type: 'expression',
+                    expression: [ '/', [ '+', 10, {ref: 'feature', key: 'x'} ], 30 ]
+                },
+                'circle-color': {
+                    type: 'expression',
+                    expression: [
+                        'concat',
+                        'rgb(',
+                            [ '+', 128, [ '*', 10, {ref: 'feature', key: 'x'} ] ],
+                            ',',
+                            [ '+', 128, [ '*', 10, {ref: 'feature', key: 'y'} ] ],
+                            ',',
+                            128,
+                        ')'
+                    ]
+                }
+            }
+        }
+    ],
+    version: 8
+}
+
+var map = window.map = new mapboxgl.Map({
+    container: 'map',
+    zoom: 0,
+    center: [0, 0],
+    style: style,
+    hash: true
+});
+
+</script>
+</body>
+</html>
--- a/debug/expressions.html
+++ b/debug/expressions.html
@@ -40,31 +40,27 @@ var style = {
                     type: 'expression',
                     expression: [
                         '*',
-                        [ '^', 2, {ref: 'map', key: 'zoom'} ],
+                        [ '^', 2, ['zoom'] ],
                         0.125,
                         [
                             '+',
                             20,
-                            {ref: 'feature', key: 'x'},
-                            {ref: 'feature', key: 'y'}
+                            [ 'number_data', 'x' ],
+                            [ 'number_data', 'y' ]
                         ]
                     ]
                 },
                 'circle-opacity': {
                     type: 'expression',
-                    expression: [ '/', [ '+', 10, {ref: 'feature', key: 'x'} ], 30 ]
+                    expression: [ '/', [ '+', 10, ['number_data', 'x'] ], 30 ]
                 },
                 'circle-color': {
                     type: 'expression',
                     expression: [
-                        'concat',
-                        'rgb(',
-                            [ '+', 128, [ '*', 10, {ref: 'feature', key: 'x'} ] ],
-                            ',',
-                            [ '+', 128, [ '*', 10, {ref: 'feature', key: 'y'} ] ],
-                            ',',
-                            128,
-                        ')'
+                        'rgb',
+                            [ '+', 128, [ '*', 10, ['number_data', 'x'] ] ],
+                            [ '+', 128, [ '*', 10, ['number_data', 'y'] ] ],
+                            128
                     ]
                 }
             }
--- a/docs/style-spec/expressions.md
+++ b/docs/style-spec/expressions.md
@@ -9,17 +9,20 @@
 - JSON string / number / boolean literal
 
 **Property lookup:**
-- Feature property: `[ "data", key_expr ]`
-- Map property:
+- Feature property:
+  - `[ "number_data", key_expr ]` reads `feature.properties[key_expr]`, coercing it to a number if necessary.
+  - `[ "string_data", key_expr ]` reads `feature.properties[key_expr]`, coercing it to a string if necessary.
+  - `[ "boolean_data", key_expr ]` reads `feature.properties[key_expr]`, coercing it to a boolean if necessary, with `0`, `''`, `null`, and missing properties mapping to `false`, and all other values mapping to `true`.
+  - `[ "has", key_expr ]` returns `true` if the property is present, false otherwise.
+  - `[ "typeof", key_expr ]` yields the data type of `feature.properties[key_expr]`: one of `'string'`, `'number'`, `'boolean'`, `'object'`, `'array'`, or, in the case that the property is not present, `'none'`.
+
+**Map property:**
   - `[ "zoom" ]` (Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.)
-  - `[ "pitch" ]`
-  - etc.
 
 **Decision:**
 - `["if", boolean_expr, expr_if_true, expr_if_false]` 
 
-**Boolean:**
-- `[ "has", key_expr ]`
+**Comparison and boolean operations:**
 - `[ "==", expr1, expr2]` (similar for `!=`)
 - `[ ">", lhs_expr, rhs_expr ]` (similar for <, >=, <=)
 - `[ "between", value_expr, lower_bound_expr, upper_bound_expr ]`
--- a/docs/style-spec/expressions.md
+++ b/docs/style-spec/expressions.md
@@ -1,0 +1,40 @@
+# Style Function Expressions
+
+**Constants:**
+- LN2
+- PI
+- e
+
+**Literal:**
+- JSON string / number / boolean literal
+
+**Property lookup:**
+- Feature property:  `{ "ref": "feature", "key": propertyName, "default": defaultValue}`
+- Map property:  `{"ref": "map",  "key": "zoom"|"pitch"|… }`
+  - Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.
+
+**Decision:**
+- `["if", boolean_expr, expr_if_true, expr_if_false]` 
+  - Alternative: `[` `"``if``"``, [ filter_expr, expr_if_true ], [ filter_expr, expr_if_true], …, expr_otherwise]` (where the first matching filter wins)
+
+**Boolean:**
+- has, !has
+- ==, !=
+- >, <, >=, <=, between(?),
+- in, !in
+- all, any, none
+
+**String:**
+- `["concat", expr1, expr2, …]`
+- `["transform", string_expr, "uppercase"|"lowecase"]`
+
+**Numeric:**
+- +, -, \*, /, %, ^ (e.g. `["+", expr1, expr2, expr3, …]`, `["-", expr1, expr2 ]`, etc.)
+- log10, ln, log2
+- sin, cos, tan, asin, acos, atan
+- ceil, floor, round, abs
+- min, max
+
+**Color:**
+- rgb, hsl, hcl, lab, hex, (others?)
+
--- a/docs/style-spec/expressions.md
+++ b/docs/style-spec/expressions.md
@@ -1,32 +1,34 @@
 # Style Function Expressions
 
 **Constants:**
-- LN2
-- PI
-- e
+- `[ "ln2" ]`
+- `[ "pi" ]`
+- `[ "e" ]`
 
 **Literal:**
 - JSON string / number / boolean literal
 
 **Property lookup:**
-- Feature property:  `{ "ref": "feature", "key": propertyName, "default": defaultValue}`
-- Map property:  `{"ref": "map",  "key": "zoom"|"pitch"|… }`
-  - Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.
+- Feature property: `[ "data", key_expr ]`
+- Map property:
+  - `[ "zoom" ]` (Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.)
+  - `[ "pitch" ]`
+  - etc.
 
 **Decision:**
 - `["if", boolean_expr, expr_if_true, expr_if_false]` 
-  - Alternative: `[` `"``if``"``, [ filter_expr, expr_if_true ], [ filter_expr, expr_if_true], …, expr_otherwise]` (where the first matching filter wins)
 
 **Boolean:**
-- has, !has
-- ==, !=
-- >, <, >=, <=, between(?),
-- in, !in
-- all, any, none
+- `[ "has", key_expr ]`
+- `[ "==", expr1, expr2]`
+- `[ ">", lhs_expr, rhs_expr ]` (similar for <, >=, <=)
+- `[ "between", value_expr, lower_bound_expr, upper_bound_expr ]`
+- `[ "in", value_expr, item1_expr, item2_expr, ... ]`
+- `[ "all", boolean_expr1, boolean_expr2, ... ]` (similar for any, none)
 
 **String:**
 - `["concat", expr1, expr2, …]`
-- `["transform", string_expr, "uppercase"|"lowecase"]`
+- `["upcase", string_expr]`, `["downcase", string_expr]`
 
 **Numeric:**
 - +, -, \*, /, %, ^ (e.g. `["+", expr1, expr2, expr3, …]`, `["-", expr1, expr2 ]`, etc.)
--- a/docs/style-spec/expressions.md
+++ b/docs/style-spec/expressions.md
@@ -16,6 +16,7 @@
   - `[ "has", key_expr ]` returns `true` if the property is present, false otherwise.
   - `[ "typeof", key_expr ]` yields the data type of `feature.properties[key_expr]`: one of `'string'`, `'number'`, `'boolean'`, `'object'`, `'array'`, or, in the case that the property is not present, `'none'`.
 - `[ "geometry_type" ]` returns the value of `feature.geometry.type`.
+- `[ "string_id" ]`, `[ "number_id" ]` returns the value of `feature.id`.
 
 **Map property:**
   - `[ "zoom" ]` (Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.)
--- a/docs/style-spec/expressions.md
+++ b/docs/style-spec/expressions.md
@@ -25,8 +25,6 @@
 **Comparison and boolean operations:**
 - `[ "==", expr1, expr2]` (similar for `!=`)
 - `[ ">", lhs_expr, rhs_expr ]` (similar for <, >=, <=)
-- `[ "between", value_expr, lower_bound_expr, upper_bound_expr ]`
-- `[ "in", value_expr, item1_expr, item2_expr, ... ]`
 - `[ "&&", boolean_expr1, boolean_expr2, ... ]` (similar for `||`)
 - `[ "!", boolean_expr]`
 
--- a/docs/style-spec/expressions.md
+++ b/docs/style-spec/expressions.md
@@ -15,6 +15,7 @@
   - `[ "boolean_data", key_expr ]` reads `feature.properties[key_expr]`, coercing it to a boolean if necessary, with `0`, `''`, `null`, and missing properties mapping to `false`, and all other values mapping to `true`.
   - `[ "has", key_expr ]` returns `true` if the property is present, false otherwise.
   - `[ "typeof", key_expr ]` yields the data type of `feature.properties[key_expr]`: one of `'string'`, `'number'`, `'boolean'`, `'object'`, `'array'`, or, in the case that the property is not present, `'none'`.
+- `[ "geometry_type" ]` returns the value of `feature.geometry.type`.
 
 **Map property:**
   - `[ "zoom" ]` (Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.)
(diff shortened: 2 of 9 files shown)


ansis:
Great!

> When an expression-based style function is both zoom-and-feature-dependent, we must use an approximation

I don't think we should support map properties (zoom, pitch, rotation) in expressions. I think the approximations would be rough enough that they'll produce confusing behavior. I'm not sure the rough edges and added complexity are worth it here.

I think that allowing expressions as the output values within all types of functions should provide enough flexibility for combining zoom + expressions.

kkaefer:
We should also add *e* to the constant list. Having some sort of call syntax for complex function calls like for https://github.com/mapbox/mapbox-gl-js/issues/4136, lowercasing/uppercasing, and formatting might be useful as well.

kkaefer:
It looks like this draft is synthesizing JavaScript functions from the stylesheet input. While this works for a mock implementation, I think we should go with an actual interpreter since it'll allow us to validate input much better, and we'll have to do it this way anyway in Native. Additionally, creating Function objects in JS will make it harder to pass through CSP (https://github.com/mapbox/mapbox-gl-js/issues/559).

anandthakker:
> lowercasing/uppercasing, and formatting might be useful as well

@kkaefer for "formatting", are you talking about printf-like formatting? or something else?

kkaefer:
Yeah, mostly number formatting (thousands separator, fractional separator, rounding/precision). String formatting with placeholders might be useful too.

anandthakker:
> I don't think we should support map properties (zoom, pitch, rotation) in expressions. I think the approximations would be rough enough that they'll produce confusing behavior. I'm not sure the rough edges and added complexity are worth it here.

I'm not convinced that the approximations we're already using for stop-based composite functions are really any _less_ confusing.  Sure, GL JS manages to send four stops into the shader, but in native, we only send two.  See also https://github.com/mapbox/mapbox-gl-native/issues/8460.

That said, I'm ambivalent between the current proposal and something like:

> I think that allowing expressions as the output values within all types of functions should provide enough flexibility for combining zoom + expressions.

Although I do think that the possibility of zoom-and-property stops _and_ expressions would be pretty weird and confusing.

anandthakker:
> I think we should go with an actual interpreter since it'll allow us to validate input much better, and we'll have to do it this way anyway in Native. Additionally, creating Function objects in JS will make it harder to pass through CSP (#559).

Capturing from chat:
Code generation is already being used for filters and `StructArray` because it's significantly faster than the alternative.  @mourner attempted fast non-eval feature-filter but it was still 5 times slower https://github.com/mapbox/feature-filter/pull/11

jfirebaugh:
Nice! I'm bullish on the approximation-based approach for zoom interpolation. (I think we should make the logic consistent between -js and -native though.)

I don't think we should embed the filter grammar directly; it's too awkward mixing it with `{ref}` syntax.

How much typing do we want? For validation purposes and easier implementation in a statically-typed language, it seems like static typing (as much as possible) is good. We of course can't type check feature-dependent references, but we can for example eschew string concatenation to construct color values in favor of built-in methods to construct them from RGBA components.

anandthakker:
> How much typing do we want? For validation purposes and easier implementation in a statically-typed language, it seems like static typing (as much as possible) is good. We of course can't type check feature-dependent references...

👍 on typing as much as possible.  

> ..., but we can for example eschew string concatenation to construct color values in favor of built-in methods to construct them from RGBA components

Agreed -- the string concatenation for colors is both fragile and cumbersome.  Added color functions to the list (still need to review csscolorparser and our color spaces functionality to get a comprehensive list -- will do)

anandthakker:
> I don't think we should embed the filter grammar directly; it's too awkward mixing it with {ref} syntax.

Yeah; I think we should just make expression-function versions of each of the filter operations.

This will be clearer, and also open the door for extending feature filters to allow expression-based filters (though of course we'll still have to keep the existing ones, too, until we're ready for a v9 of the style spec)

ChrisLoer:
This looks awesome! 

On the "what zoom levels should we use" question, we need to keep #4558 in mind. Supporting multiple zoom stops within the `[tileZoom, tileZoom + 1]` range would make it even more complicated to calculate the single "zoom at which collision occurs" number that we need to pass to the shaders for each label. Doing collision detection at render time would make that problem easier, but even with "future of labeling" (#4704) changes I don't think we have a firm plan to do collision detection synchronously.

anandthakker:
Good point @ChrisLoer  --  in light of this and in the interest of making js and native consistent, I updated the proposal to just make it `tile.zoom`, `tile.zoom + 1`.  I think this is also much easier to document/explain ("Expressions that refer to the map zoom level will only be evaluated at integer zooms. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.")

anandthakker:
Moved the expression AST details into `expressions.md` so that we can comment directly on it & track changes.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,40 @@
+# Style Function Expressions
+
+**Constants:**
+- LN2
```
jfirebaugh: What's the syntax for constants? Suggestion: `["ln2"]`, `["pi"]`, etc.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,40 @@
...
+- PI
+- e
+
+**Literal:**
+- JSON string / number / boolean literal
+
+**Property lookup:**
+- Feature property:  `{ "ref": "feature", "key": propertyName, "default": defaultValue}`
```
jfirebaugh: Syntax alternatives:

* `["data", key, default]`, where _key_ and _default_ are both expressions
* `["data", key]`, where _key_ is an expression, and we also provide a `["has", key]` expression, with which arbitrary default logic can be implemented.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,40 @@
...
+- Map property:  `{"ref": "map",  "key": "zoom"|"pitch"|… }`
```
jfirebaugh: Syntax alternative: `["zoom"]`, `["pitch"]`, etc.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,40 @@
...
+  - Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.
+
+**Decision:**
+- `["if", boolean_expr, expr_if_true, expr_if_false]` 
+  - Alternative: `[` `"``if``"``, [ filter_expr, expr_if_true ], [ filter_expr, expr_if_true], …, expr_otherwise]` (where the first matching filter wins)
+
+**Boolean:**
+- has, !has
```
jfirebaugh: Should also have `["not", expression]` in addition to, or instead of `!` variants of individual operators.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,40 @@
...
+- ==, !=
+- >, <, >=, <=, between(?),
+- in, !in
+- all, any, none
+
+**String:**
+- `["concat", expr1, expr2, …]`
+- `["transform", string_expr, "uppercase"|"lowecase"]`
```
jfirebaugh: Syntax alternative: `["upcase", expression]`, `["downcase", expression]`.

mb12:
@anandthakker  Is it reasonable to expect that any new operators, functions, expressions  will work seamlessly with filters as well?  The language used to select "list of features to layout"  should at least be as expressive as language used to describe style functions (with limitations).

docs/style-spec/expressions.md
anandthakker: I'm coming around to your original suggestion of just relying on the conditional operator to provide default behavior.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,40 @@
...
+- JSON string / number / boolean literal
+
+**Property lookup:**
+- Feature property:  `{ "ref": "feature", "key": propertyName, "default": defaultValue}`
+- Map property:  `{"ref": "map",  "key": "zoom"|"pitch"|… }`
+  - Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.
+
+**Decision:**
+- `["if", boolean_expr, expr_if_true, expr_if_false]` 
+  - Alternative: `[` `"``if``"``, [ filter_expr, expr_if_true ], [ filter_expr, expr_if_true], …, expr_otherwise]` (where the first matching filter wins)
```
anandthakker: As with conditionals, I'm leaning back towards the first proposal here (the one that's equivalent to ternary operator).  Compactness in the JSON representation doesn't seem particularly important, since no one will actually be authoring expressions via AST (except for us when we write unit tests 😄 )

debug/expressions.html
```diff
@@ -0,0 +1,86 @@
...
+                        '*',
+                        [ '^', 2, {ref: 'map', key: 'zoom'} ],
+                        0.125,
+                        [
+                            '+',
+                            20,
+                            {ref: 'feature', key: 'x'},
+                            {ref: 'feature', key: 'y'}
+                        ]
+                    ]
```
anandthakker: That's so nice!  What do `map.zoom`, `feature.x`, etc. end up looking like in the resulting NSExpression object?

debug/expressions.html
```diff
@@ -0,0 +1,86 @@
...
+                },
+                'circle-opacity': {
+                    type: 'expression',
+                    expression: [ '/', [ '+', 10, {ref: 'feature', key: 'x'} ], 30 ]
+                },
+                'circle-color': {
+                    type: 'expression',
+                    expression: [
+                        'concat',
```
anandthakker: 👍 yep, agreed that having color functions `rgb`, etc. as opposed to using string concatenation. (Ref: https://github.com/mapbox/mapbox-gl-js/pull/4715#issuecomment-302150023)

anandthakker:
> Is it reasonable to expect that any new operators, functions, expressions will work seamlessly with filters as well? 

Yes. We don't want to make a breaking change to the existing filter syntax, so my current thinking is that we'll extend the `filter` spec to allow both "legacy" filters and the new boolean expression functions.  Not sure yet whether we'll need to add any extra wrapping to avoid ambiguity between the two, but we'll cross that bridge when we come to it

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,40 @@
+# Style Function Expressions
+
+**Constants:**
+- LN2
+- PI
+- e
```
asheemmamoowala: I think using `$` to denote constants will help make the style-spec and our expression grammar more consistent. I could go either way on uppercase/lowercase.The uppercase makes it much easier to parse expressions when looking at a style spec as opposed to working with an SDK (not sure how important that use case is)

Wouldn't it be easier to read and debug if we used `"$ln2"`, `"$pi"` ?

jfirebaugh: The issue with introducing a mini-syntax inside of string literals is that you then also have to include an escaping syntax, so that the literal string "$pi", or other literals starting with a dollar sign, remain expressible. If you pick backslash as the escape character, then that has to be compounded with the escaping syntax of JSON itself (`"\\$pi"`). This is confusing accidental complexity. Better to keep the grammar simple and use s-expressions as the _only_ syntactical element. Expose constants as _constant functions_: `["pi"]`.

debug/expressions.html
1ec5: The initialized NSExpression consists of the following object tree:

* NSExpression representing the function `multiply:by:`, the operand `_NSPredicateUtilities`, and the following NSExpressions as arguments:
  * NSExpression representing the function `multiply:by:`, the operand `_NSPredicateUtilities`, and the following NSExpressions as arguments:
    * NSExpression representing the function `raise:toPower:`, the operand `_NSPredicateUtilities`, and the following NSExpressions as arguments:
      * NSExpression representing the key path `map.zoom`
      * NSExpression representing the constant value 2
    * NSExpression representing the constant value 0.125
* etc.

NSExpression comes with a built-in way to evaluate itself given a context dictionary. However, imperative filters aren’t on the roadmap – https://github.com/mapbox/mapbox-gl-native/issues/7860#issuecomment-275257972 – so the iOS/macOS SDK instead translates the AST provided by NSExpression into the AST expected by mbgl.

docs/style-spec/expressions.md
1ec5: I agree that we need to avoid a mini-language in the JSON format – that’s what derailed #4079 for so long.

My original comment above refers to an approach we could take for NSExpression conversion, since that DSL already makes a distinction between key paths, variables, and strings. For example, the iOS SDK has the following [guidance](https://www.mapbox.com/ios-sdk/api/3.5.2/Classes/MGLVectorStyleLayer.html#/c:objc%28cs%29MGLVectorStyleLayer%28py%29predicate) for using the special `$id` and `$type` feature properties in a predicate:

> It is possible to create expressions that contain special characters in the predicate format syntax. This includes the `$` in the `$id` and `$type` special style attributes and also `hyphen-minus` and `tag:subtag`. However, you must use `%K` in the format string to represent these variables: `@"%K == 'LineString'", @"$type"`.

In that example, the developer would want `$type` treated as a key path. If they put `$type` directly in the format string, NSPredicate interprets it as a variable instead. For these mathematical constants, however, the developer would expect it to be interpreted as a variable, making an inline `$pi` desirable.

1ec5: > Better to keep the grammar simple and use s-expressions as the only syntactical element.

By the way, one could make the same argument for treating bare strings in the JSON as constant values always and requiring the use of a `{ref: …}` object to represent a property reference. But that would either move the expression syntax away from the filter syntax or necessitate a backwards-incompatible change to the filter syntax.

NSExpression format strings make a distinction between `keyPath` and `'string'` or `"string"`, which makes it possible to represent the style JSON `["==" "property", "string"]` as either `property = 'string'` or `'string' = property` in an NSExpression format string.

anandthakker: > By the way, one could make the same argument for treating bare strings in the JSON as constant values always and requiring the use of a {ref: …} object to represent a property reference. But that would either move the expression syntax away from the filter syntax or necessitate a backwards-incompatible change to the filter syntax.

See https://github.com/mapbox/mapbox-gl-js/pull/4715#issuecomment-302153124

jfirebaugh: > By the way, one could make the same argument for treating bare strings in the JSON as constant values always and requiring the use of a `{ref: …}` object to represent a property reference. But that would either move the expression syntax away from the filter syntax or necessitate a backwards-incompatible change to the filter syntax.

Yes, that's exactly what @anandthakker and I are agreeing we should do (except I suggested `["data", …]` rather than `{ref: …}`).

In this proposed grammar, the first element of a JSON array is always a JSON string constant naming a style spec function. The zero or more other elements are expressions. One type of expression is a string constant expression, which is represented by a JSON string constant. In no case does a string constant alone denote "the value of the property with this name". This is indeed a break from the filter syntax.

This does mean that there can't be a direct correspondence between JSON array literals and style array literals. For example, an expression whose result is a font stack array will need to be something like `["array", "Open Sans Regular", "Arial Unicode MS Regular"]`.

docs/style-spec/expressions.md
anandthakker: This would be equivalent to `["none", expression]`, right?

jfirebaugh: Yes, although `!` is usually considered the more primitive operator, `["none", e₀ ... eⁿ]` being expressed as `["!",

jfirebaugh: Yes, although ! is usually considered the more primitive operator, `["none", e⁰ ... eⁿ]` being expressed as `["!", ["any", e⁰ ... eⁿ]]`.

anandthakker: Good point... I keep getting stuck thinking in terms of existing filter syntax, but there's no real need to match that.

docs/style-spec/expressions.md
```diff
@@ -1,32 +1,34 @@
...
 - JSON string / number / boolean literal
 
 **Property lookup:**
-- Feature property:  `{ "ref": "feature", "key": propertyName, "default": defaultValue}`
-- Map property:  `{"ref": "map",  "key": "zoom"|"pitch"|… }`
-  - Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.
+- Feature property: `[ "data", key_expr ]`
+- Map property:
+  - `[ "zoom" ]` (Note: expressions that refer to the map zoom level are only evaluated at integer zoomslevels. When the map is at non-integer zoom levels, the expression's value will be approximated using linear or exponential interpolation.)
+  - `[ "pitch" ]`
```
1ec5: Why would feature property lookups require a `data` function while `zoom` and `pitch` be usable independently of a `map` function, similar to constants? Is it because the set of map properties is finite and well-defined, in contrast to the set of potential feature properties?

docs/style-spec/expressions.md
```diff
@@ -1,32 +1,34 @@
...
-  - Alternative: `[` `"``if``"``, [ filter_expr, expr_if_true ], [ filter_expr, expr_if_true], …, expr_otherwise]` (where the first matching filter wins)
 
 **Boolean:**
-- has, !has
-- ==, !=
-- >, <, >=, <=, between(?),
-- in, !in
-- all, any, none
+- `[ "has", key_expr ]`
+- `[ "==", expr1, expr2]`
```
1ec5: How would the style author specify that a comparison between two strings should be performed case- or diacritic-insensitively (mapbox/mapbox-gl-js#4136)? Or would we require `downcase` and a `strip-diacritics` function to be applied explicitly to either operand?

docs/style-spec/expressions.md
anandthakker: > Is it because the set of map properties is finite and well-defined, in contrast to the set of potential feature properties?

^ yeah, that's my thinking.  there's no need for it to be `[ "map", "zoom" ]` when we can simply reserve `[ "zoom" ]` as a 0-ary function.

docs/style-spec/expressions.md
anandthakker: > Or would we require downcase and a strip-diacritics function to be applied explicitly to either operand?

This approach sounds good to me...

src/style-spec/function/expression.js
```diff
@@ -0,0 +1,304 @@
...
+        output: Type.Number,
+        math: true
+    },
+    'abs': {
+        input: [Type.Number],
+        output: Type.Number,
+        math: true
+    },
+    'min': {
+        input: [Type.Array[Type.Number]],
```
anandthakker: Currently, `Type.Array[t]` really just means 'n arguments of type t'; given that we'll probably end up needing to use that for actual array values (like `["array", e0, e1, e2...]`), this should probably be changed to something like `Type.Nargs[t]`.

docs/style-spec/expressions.md
1ec5: > Or would we require downcase and a strip-diacritics function to be applied explicitly to either operand?

I’m not so sure that would be robust enough for internationalization purposes: https://github.com/mapbox/mapbox-gl-js/issues/4136#issuecomment-302550706.

Relatedly, should `==` take multiple arguments in order to say, for instance, `a == b == c`? (If not, that would open the door to using the second argument for folding flags.) If `==` shouldn’t take multiple arguments, should arithmetic operators like `*` take multiple arguments?

samanpwbb:
How will arbitrary functions relate to existing function syntax? I can see how most functionality in current function syntax could be replicated with arbitrary functions. I can also see ways that the two types of functions could be combined. I strongly recommend we strive for a solution that promotes a single way to do things (for example, _either_ remove camera expressions from arbitrary functions and allow arbitrary functions as return values in zoom functions, _or_ keep camera expressions in arbitrary functions and have a real plan in place to deprecate zoom functions).

debug/expressions.html
davidtheclark: The CSS 4 color-modifying function draft might serve as inspiration here: https://drafts.csswg.org/css-color/#modifying-colors

anandthakker:
> keep camera expressions in arbitrary functions and have a real plan in place to deprecate zoom functions).

This is my strong preference: we'll still have to support stop-based functions for a while yet in GL implementations, but once we get arbitrary expressions in place, I propose that we formally mark stop functions as deprecated and exclusively focus on arbitrary expressions in docs/examples, Studio, and any other tools.

debug/expressions.html
anandthakker: Woah.  That's pretty awesome, although including "color adjuster" functions would probably require taking on a dependency like https://github.com/gka/chroma.js/.

For the scope of this PR, I'm inclined to stick to color _constructors_, and to ticket/discuss separately the possibility of adding more sophisticated color operations.

ansis:
Overall the direction of this stuff sounds great!

I have some concerns about supporting map properties in expressions.

### implementation
How will the approximate expressions that use pitch or rotation be implemented? What if they use pitch, zoom *and* rotation?

### interpolation
The plan for implementing expressions that use zoom is to evaluate them at several zoom levels and interpolate them. Will they be interpolated linearly? exponentially? it seems like for either `[‘*’, [‘zoom’], 4]` or `[‘^’, 2, [‘zoom’]` the interpolation within a zoom level won’t be what you expect.

How would you write an expression for a line-width that is:
- 2px at zoom 10
- 12px at zoom 18
- increases at some kind of exponentialish curve in between?

### approximation

The approximation used by the implementation would have to be part of the style spec, right? At zoom 4.5 would `[‘concat’, ‘zoom ’, [zoom]]` evaluate to *zoom 4*?

In previous style function discussions it seemed important to support changes to functions at fractional zoom levels. Is this still important to support? Would `[if [‘>’, [‘zoom’], 4.5], 10, 0]` result in a value that smoothly increases between z4 and z5 or the expected sharp jump?

### UI

I'm not caught up on the plans for dds in studio, but:

Expressions are more flexible, and with this flexibility comes the increase in variation and dimensionality that makes it harder to make UI's for.

A goal with the property+zoom functions was to make the form regular. They are basically a table. On one axis you have zoom on the other axis you have properties. I'm thinking a free form expression is going to a lot hard to build a ui for?

> I strongly recommend we strive for a spec that promotes a single way to do things (for example, either remove camera expressions from arbitrary functions and allow arbitrary functions as return values in zoom functions, or keep camera expressions in arbitrary functions and have a real plan in place to deprecate zoom functions).

I agree. I'm thinking we should leave exist functions responsible for zoom allow expressions wherever we allow property values.

nickidlugash:
Super excited for this! A few random questions (some of which I mentioned in the arbitrary expressions Studio meeting just now):

Conditionals: 

Will/can we include evaluating the existence of sprite values? This is in reference to discussions about "image stacks": https://github.com/mapbox/mapbox-gl-js/issues/4149. 

In the same vain, would we consider including evaluation of unicode characters as a conditional, and replacing fontstacks with a subset of arbitrary expression options (on the front end)?  

Formatting:

> lowercasing/uppercasing, and formatting might be useful as well

> Yeah, mostly number formatting (thousands separator, fractional separator, rounding/precision). String formatting with placeholders might be useful too.

Would we then deprecate the `text-transform` property in v9? And in general, would we avoid any future style spec properties for text formatting that change which glyphs are used to display a value, opting instead to include this as part of arbitrary expressions? It does seem that the ability to specify formatting for substrings is more valuable than spec properties that can only apply formatting for the whole value.

anandthakker:
## Map properties

> How will the approximate expressions that use pitch or rotation be implemented? What if they use pitch, zoom and rotation?

Good point... I kinda just threw pitch/rotation in there without a plan.  Perhaps supporting map properties other than `zoom` is not feasible -- let's just remove them for now and revisit in the future.

## Zoom Interpolation / Approximation

Here's my proposal.  At the top level, an expression-based style function is defined using `{type: "expression", expression: [...], zoomInterpolationBase: (number)`.  It's evaluated as follows:
1. If the expression is constant, zoom-dependent only, or feature-dependent only, then it is evaluated exactly; no approximation is needed, with the exception that layout properties that are zoom-dependent are evaluated at layout time, and thus use the tile's zoom level rather than the rendered zoom level.
2. Otherwise (i.e. if the expression is _both_ zoom- and feature-dependent), paint property expressions are evaluated for each feature at `tile.zoom` and `tile.zoom + 1`.  Then, for each frame, we use the current zoom level and the value of `zoomInterpolationBase` to compute the appropriate interpolation factor.

> Will they be interpolated linearly? exponentially?

Per 2, this would be configurable.

> The approximation used by the implementation would have to be part of the style spec, right? At zoom 4.5 would [‘concat’, ‘zoom ’, [zoom]] evaluate to zoom 4?

> In previous style function discussions it seemed important to support changes to functions at fractional zoom levels. Is this still important to support? Would [if [‘>’, [‘zoom’], 4.5], 10, 0] result in a value that smoothly increases between z4 and z5 or the expected sharp jump?

Since `['concat', 'zoom', [zoom]]` is only zoom-dependent, when map.zoom == 4.5, it would evaluate to _zoom 4_ for a layout property or _zoom 4.5_ for a paint property.  However `['concat', ['data', 'x'], ['zoom']`, being zoom-and-feature-dependent would evaluate to _(value of x) 4_ for both layout or paint properties.  Similar answer for the conditional example: we'd get the expected sharp jump for zoom-only expression, but an (inaccurate) smooth increase for a zoom-and-property expression.

These limitations on zoom-and-property functions aren't ideal, but they actually already exist for stop-based functions, because the existing implementation already uses (roughly) the strategy outlined in 1 & 2 above.

## UI

Just met w/ Studio team on this - @samanpwbb is going to cut a ticket for discussion, I believe.

1ec5:
> would we consider including evaluation of unicode characters as a conditional, and replacing fontstacks with a subset of arbitrary expression options (on the front end)?

This would require the introduction of a `font-has-glyph` function or somesuch. Can you describe the use case for a dedicated function of that sort that wouldn’t be addressed by a font stack?

Along the same lines, #4149 requests “image stacks” by analogy with font stacks. If we were to “nest” that capability within the expression language, then we’d have to introduce an `image-exists` function.

For the most common use case where the font or image name is specified literally, it would be far more straightforward to set `text-field` and `icon-image` to arrays of literal strings rather than to expressions containing conditionals containing existence checks. On the other hand, I’m not sure how feasible it would be to implement #4149 but then allow expressions within image names for the use cases that require them.

nickidlugash:
#### Font stacks

> Can you describe the use case for a dedicated function of that sort that wouldn’t be addressed by a font stack?

I was thinking more about user-defined ranges rather than just a `font-has-glyph` type function, so that you could define a categorical rather than hierarchical font selection for situations like this:

- Font 1 covers ranges A and B
- Font 2 covers ranges A and B
- I want to use Font 1 for A, and Font 2 for B

The example of this that I've encountered is trying to combine the fonts Noto Sans CJK and DIN in a Chinese language map: both fonts cover [Basic Latin](http://www.unicode.org/charts/PDF/U0000.pdf). I wanted numerals to be assigned Noto and the rest of Basic Latin to be assigned DIN.

---

#### Zoom approximation

> These limitations on zoom-and-property functions aren't ideal, but they actually already exist for stop-based functions, because the existing implementation already uses (roughly) the strategy outlined in 1 & 2 above.

I mentioned this briefly in the Studio meeting, but just wanted to mention again that I think the existing implementation is still just beginning to get used in the wild. We do currently use some fractional zoom assignments in our core styles that we wouldn't be able to port over to arbitrary expressions (or DDS). I don't think any of our core style use cases are critical though (they're mostly used to transition the opacity for certain layers). If it's helpful, carto team can compile a more specific list of our use cases for fractional zoom assignments (for zoom-and-property functions).

ansis:
> hey if you have time in the next day or so, interested to hear what you think about https://github.com/mapbox/mapbox-gl-js/pull/4715#issuecomment-302806891 — does it address your concerns, or do they linger?

### UI

I still have my concerns about what a UI for expressions would look like. Expressions are programming. They are powerful, but aren't that UI-friendly. It might make sense to make that tradeoff so that we convert units and combine properties, but I think it make sense to keep that power as contained as possible.

If the problems expressions are trying to solve were approached from Studio's perspective, what would that look like? What if we thought about the UI before (or while) we think about the spec?

I think restricting expressions to data properties only and sticking with a more rigid format for the zoom part of functions might be an ok tradeoff.

> This is my strong preference: we'll still have to support stop-based functions for a while yet in GL implementations, but once we get arbitrary expressions in place, I propose that we formally mark stop functions as deprecated and exclusively focus on arbitrary expressions in docs/examples, Studio, and any other tools.

This is what the `line-width` for `road-motorway` looks like:

screen shot 2017-05-24 at 11 02 11 am

```json
{
  "line-width": {
    "stops": [
      [5, 0.75],
      [18, 32]
    ],
    "base": 1.5
}

```

What would this look like as an expression? What would the UI look like for this?

### Other

> Since ['concat', 'zoom', [zoom]] is only zoom-dependent, when map.zoom == 4.5, it would evaluate to zoom 4 for a layout property or zoom 4.5 for a paint property. However ['concat', ['data', 'x'], ['zoom'], being zoom-and-feature-dependent would evaluate to (value of x) 4 for both layout or paint properties. Similar answer for the conditional example: we'd get the expected sharp jump for zoom-only expression, but an (inaccurate) smooth increase for a zoom-and-property expression.

I think we should try to avoid inconsistencies like this, where adding a reference to data affects how the zoom part of the expression is evaluated.

> These limitations on zoom-and-property functions aren't ideal, but they actually already exist for stop-based functions, because the existing implementation already uses (roughly) the strategy outlined in 1 & 2 above.

This can be fixed, I think. It should be possible to evaluate them exactly.

samanpwbb:
I'm planning to write up a more thorough issue on the UI implications of arbitrary functions, but I wanted to respond real quick here with a few points. I absolutely think we need to think about Studio's UI and this spec simultaneously:

- Stop-based zoom functions aren't necessarily ideal. In many cases a straightforward expression like `1 + ({zoom} x 1.5)` does the job. Arbitrary functions open up the possibility to iterate on the ideal zoom adjustment UI (it's more work but might result in a better UI).
- Stop-based functions are really useful for categorical data (for example, road class). Sounds like deeply nested conditional statements can do the same thing, but the syntax is really gross! Might be fine if we always have a layer of abstraction on top in Studio.
- It's okay for Mapbox Studio to be complex as long as there's a reasonable gain in power. We want to make a creative tool that enables people to experiment and unlock new possibilities.
- It would be possible to have two tiers in Studio's UI: one tier of easy-to-use widgets on top of arbitrary functions that write very complex functions but have very simple interfaces, and then another tier that lets users write expressions by hand with a string / token -based UI. We could manage which widgets are active with layer metadata or even with pattern matching.
- The amount of effort in Studio to handle arbitrary expressions is high. I can't give a solid estimate, but likely twice as much work as implementing a user-friendly zoom-and-stop based UI.

anandthakker:
> I think we should try to avoid inconsistencies like this, where adding a reference to data affects how the zoom part of the expression is evaluated.

Yep, I agree.  But again, I think this is orthogonal to the design decisions we're discussing here, since it's already a problem now. (I.e., switching from `stops: [[0, 1], [18, 10]]` to `stops: [[{zoom: 0, value: 0, 1], [zoom: 18, value: 0, 10]]` would incur the same inconsistency.)

ansis:
> I'm planning to write up a more thorough issue on the UI implications of arbitrary functions, but I wanted to respond real quick here with a few points.

Thanks @samanpwbb, that'll be really helpful

> Yep, I agree. But again, I think this is orthogonal to the design decisions we're discussing here, since it's already a problem now. (I.e., switching from stops: [[0, 1], [18, 10]] to stops: [[{zoom: 0, value: 0, 1], [zoom: 18, value: 0, 10]] would incur the same inconsistency.)

You're right that this is a problem currently. I think it's relevant here because the current approximation is a result of an incomplete implementation rather than an unavoidable consequence of the spec. The current implementation could be fixed to be consistent but it's not clear that this could be done for zoom+data expressions.

## Concrete examples

It's been helpful in past spec discussions to keep ideas grounded with specific, real examples. The `line-width` for `road-motorway` in mapbox streets:
```json
{
    "stops": [
        [5, 0.75],
        [18, 32]
    ],
    "base": 1.5
}
```

A rough translation to an expression:
```json 
["max",
    0.75,
    ["min",
        32,
        ["+", ["*", 0.0037, ["^", 2, ["-", ["zoom"], 5]], 0.75]]]
```
A linear version of the expression ("base": 1):
```json
["max",
    0.75,
    ["min",
        32,
        ["+", ["*", 2.38, ["zoom"]], -11.15]
```

In which representation is it easier to understand what resulting values are? In which is it easier to understand what needs to change if you want the line to be 24px wide at z18 instead of 32px?

There might be some examples that are easier with zoom-expressions than stops. Do you have any in mind?

## Strengths of stop functions
I'm sure I'm slightly biased towards them because I'm familiar with them, but here's a couple advantages they have:

- You can think in terms of "I want a line to be X px wide at Y zoom level" and input that. You don't have manually calculate how to turn that into an equation. (math theory: In the linear case, stops are conceptually closer to the [two point form](https://en.wikipedia.org/wiki/Linear_equation#Two-point_form) of an equation and expressions demand the [slope-intercept form](https://en.wikipedia.org/wiki/Linear_equation#Slope.E2.80.93intercept_form), which I think is harder to think in when styling.)
- Stop functions handle continuity for you. In some cases, the way you want a value to grow isn't easily representable by a single exponential or linear function. Currently this is done by using [> 2 stops](https://cloud.githubusercontent.com/assets/1421652/26419062/70e8d5c0-408c-11e7-9d6d-ae3315a9353e.png). To avoid sharp jumps, we need to have continuity at the points where it switches between subfunctions. To do this with expressions you'd need to manually construct subfunctions that have the same value at the points where they switch, which is not trivial in this context. (math theory: [continous piecewise functions](https://en.wikipedia.org/wiki/Piecewise#Continuity))
- Stop functions make constructing exponential functions relatively easy. With expressions you need to understand what an exponential function is, how to construct one from the (zoom, value) pairs you have in your mind, and how to tweak them.

The function we implement will have some text representation and some ui representation, but underlying them will be some mathematicalish abstraction. We should expose the abstract that is closest to how people think about styles and that makes editing and adjusting styles easiest.

Expressions could be expanded to make ^ easier, but I think this would turn into just adding a "stops" function to the expression language: `["stops", ["zoom"], [[5, 0.75], [18, 32]]`. This doesn't really help with the *have one way to do things* thing.

anandthakker:
> I think it's relevant here because the current approximation is a result of an incomplete implementation rather than an unavoidable consequence of the spec.  The current implementation could be fixed to be consistent but it's not clear that this could be done for zoom+data expressions.

Good point.  Is it plausible that we could fix the current approach in the foreseeable future?

> The function we implement will have some text representation and some ui representation, but underlying them will be some mathematicalish abstraction.  We should expose the abstract that is closest to how people think about styles and that makes editing and adjusting styles easiest.

This is a good principle for designing an API or UI, but I'm not sure it applies here.  Do we consider manual readability/editability of style JSON a primary use case?  In my experience as a user of GL JS, I always produced style JSON in one of two ways: 1) Studio; 2) programatically.

Based on that, while the strengths of stop functions listed above are great points, I also think they're more relevant as considerations for how we design DDS features in Studio and how we design programmatic APIs for constructing expression functions in JS and the various SDKs.  For the spec itself, I'm in favor of choosing the abstraction that's most expressive, even if it's not as directly/manually usable.

On the other hand, if we want to consider direct editing of style JSON as a primary use case, then I would actually want to go back and reconsider the notion of allowing text representations (via an expression mini-language) rather than an AST.

> There might be some examples that are easier with zoom-expressions than stops. Do you have any in mind?

One example: changing the shape of the curve for different zoom ranges.  E.g. constant from z0-z10, scale physically from z10-z15, scale linearly after that: `zoom < 10 ? 1 : zoom < 15 ? 2^(zoom - 10) : 32 * (zoom - 15)`.

ansis:
> Is it plausible that we could fix the current approach in the foreseeable future?

The shader only needs two steps as input. If we dynamically pick the right stops on the cpu and bind the corresponding buffer offset to the attribute it should work. There are still some challenges: adding a variable number of stops to the buffer (or limiting the number in the spec), and choosing the right offset at draw time.

> > The function we implement will have some text representation and some ui representation, but underlying them will be some mathematicalish abstraction. We should expose the abstract that is closest to how people think about styles and that makes editing and adjusting styles easiest.

> This is a good principle for designing an API or UI, but I'm not sure it applies here. Do we consider manual readability/editability of style JSON a primary use case? In my experience as a user of GL JS, I always produced style JSON in one of two ways: 1) Studio; 2) programatically.

No, manual readability/editability is not primary use case, but I was trying to communicate something different:

`["+", ["*", b, ["^", 2, ["-", ["zoom"], a]], c]]]` and `2^(zoom - a) * b + c`
have the same *abstraction* but a different *representation*. The current zoom stops functions and studio ui have the same *abstraction* but a different *representation*. (I'm not sure if I'm choosing the right words here)

Comparing the the current stops json to s-expressions might not have been the most fair comparison because s-expressions are a bit annoying to read, but when I compare `2^(zoom - 5) * 0.0037 + 0.75` with `"stops": [[5, 0.75], [18, 32]]` I still find the stops abstraction easier to work with.

I think we to make studio and the spec have the same *abstraction* because:
- converting between them is not always lossless
- ideally studio can nicely handle the full spec, not just the subset that studio's abstraction produces
- what's the advantage of not doing this? would a programmatic api want a different abstraction than studio?

ansis:
*capturing a discussion from chat*

It seems like the current consensus is to:
- deprecate zoom+property functions
- deprecate property functions
- keep zoom functions
- allow the values within zoom functions to be expressions that use data properties
- let non-function values be expressions that use data properties
- maybe add `["exponential, ...]` and `["match", ...` and `["interval", ...]` or something to make common use cases covered by property functions easier

ansis:
A bunch of stuff that might be worth supporting:

### scales

- a way to map exact values to subexpressions, replacing `type: categorical`
- a way to match a value to a corresponding interval, replacing `type: interval`
- a way to easily interpolate values exponentially, replacing `type: exponential`
- a way to easily interpolate values linearly 

external prior art: https://github.com/d3/d3-scale

### formatting, localization
- number to string with specified number of decimal places (`1.2345` to `"1.23"`)
- number to localized string (`21000` to `"21,000"` or `"21 000"` depending on locale)
- methods for converting between distance units (metres, feet, kilometres, miles)
- some way of localizing text by choosing the right name property
    - maybe `["localize", lambda]` where *lambda* is a user-supplied function that returns text or undefined for a specific language code. `["localize", ...]` would use the map's/browser's/device's language(s) to get the right text or fallback text
- lambdas, to support ^
- time formatting/localization? temperature? money?

anandthakker:
@ansis interesting re: localization.  Rather than introducing lambdas, would it suffice to add one or more device-driven constants (`['locale']` or something?), which could be used in `if`s or switches?

anandthakker:
- Updated top of ticket to reflect the most recent consensus, and w/ a few open items.
 - Updated `expressions.md` with a rough draft of copy for the style-spec docs reflecting same

1ec5:
There was previously a proposal in mapbox/mapbox-gl-native#7860 to allow the use of lambdas as layout or paint property values. Lambdas (blocks in Objective-C parlance, closures in Swift parlance) would be very attractive to developers and technically feasible on the native side, but there’s a risk that developers would tend to introduce concurrency bugs with such an API. On iOS, [blocks are used quite commonly with NSExpression](https://developer.apple.com/reference/foundation/nsexpression/1407823-expressionforblock?language=objc), but normally in the context of database querying, where a developer might anticipate having to be careful about thread safety.

ansis:
> @ansis interesting re: localization. Rather than introducing lambdas, would it suffice to add one or more device-driven constants (['locale'] or something?), which could be used in ifs or switches?

My thinking is that the [fallback logic](https://cloud.githubusercontent.com/assets/1092694/26008946/cd5261e0-3714-11e7-9edc-962bced4fa55.png) can be pretty complex and isn't something you want people to have to provide themselves every time they add text. I don't know much about localization so this definitely needs further research. I know a bunch of other people at mapbox have thought a lot about this stuff.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Decision:**
+- `["if", boolean_expr, expr_if_true, expr_if_false]` 
+- `["switch", default_result_expr, [bool_expr1, result_expr1], [bool_expr2, result_expr2], ...]`
+- `["match", input_expr, default_result_expr, [test_expr1, result_expr1], [test_expr2, result_expr2]]`
+
+**Comparison and boolean operations:**
+- `[ "==", expr1, expr2]` (similar for `!=`)
+- `[ ">", lhs_expr, rhs_expr ]` (similar for <, >=, <=)
+- `[ "between", "[]"|"(]"|"[)"|"()", test_expr, left_boundary_expr, right_boundary_expr ]` - tests whether `test_expr` is "between" the left and right boundaries, with the first argument dictating whether the each boundary is exclusive or inclusive.
```
anandthakker: Is `[)` / `(]` / etc. argument just too weird?  AFAICT there's no clear choice that's more common/expected than all the others...

anandthakker: Yeah, I've gone back and forth on this.   I had removed it, but then reinstated it while thinking about how to express `interval` type property functions that stays a bit closer to the intent than breaking it all the way down into basic comparisons would.  But I'm coming around to thinking we just include an `interval` expression type, in which case maybe we just drop `between`.

1ec5:
For number formatting (mapbox/mapbox-gl-js#4119), I think we should consider a function that takes a locale identifier as an argument rather than one that takes a format. (The locale identifier can come from a source or default to the user’s locale.) This appears to be feasible using existing platform APIs ([JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat), [Objective-C](https://developer.apple.com/reference/foundation/nsnumberformatter/1410565-format?language=objc), [Java](https://docs.oracle.com/javase/7/docs/api/java/text/NumberFormat.html)).

If on the other hand we allow the developer to specify the format explicitly, it would be more difficult to correctly format Arabic–Indic or Hindu numerals, or to group digits by [lakhs](https://en.wikipedia.org/wiki/Lakh) and [crores](https://en.wikipedia.org/wiki/Crore). There is [a Unicode standard](http://www.unicode.org/reports/tr35/tr35-numbers.html#Number_Format_Patterns) for number format patterns, but I think composing these patterns would add an unnecessary burden on style authors.

anandthakker:
> For number formatting (mapbox/mapbox-gl-js#4119), I think we should consider a function that takes a locale identifier as an argument rather than one that takes a format

I agree, but I think we should handle this in a follow-up to the initial expressions push.

docs/style-spec/expressions.md
jfirebaugh: Syntax proposal for intervals:

["intervals", _x_, _y<sub>1</sub>_, _n<sub>1</sub>_, ..., _y<sub>m</sub>_, _n<sub>m</sub>_, _y<sub>m+1</sub>_]

In this formulation, the input _x_ expression is followed by _m+1_ possible output _y_ expressions interspersed with _m_ breakpoints. In order to permit efficient evaluation, the breakpoints _n<sub>1</sub>_ < ... < _n<sub>m</sub>_ are required to be numeric literals in strictly ascending order; expressions are not allowed. The result is the first _y<sub>m_ such that _n<sub>m_ >= _x_, or _y<sub>m+1</sub>_ if _x_ is greater than all _n_'s.

Example:

```
"icon-image": {
    "type": "expression",
    "value": ["intervals", ["data", "population"],
        "small-dot", 10000, "medium-dot", 100000, "large-dot"]
}
```

anandthakker:
Continued => https://github.com/mapbox/mapbox-gl-js/pull/4754

samanpwbb:
Would it be possible to see example expressions for each of the `Important use cases that the final design should cover`? Even if they're simple?

PR #4717, opened 2017-05-17
Single-pass halo text

d1manson:
With this change, halo text is rendered in the same shader pass as the text itself.

The motivation for this was to improve the appearance of overlapping text as shown below:

screen shot 2017-05-17 at 10 55 23

Presumably it also has a somewhat beneficial impact on performance.

## Launch Checklist
 - [x] document any changes to public APIs   _...none in this case_
 - [ ] post benchmark scores

### DIFF
--- a/src/render/draw_symbol.js
+++ b/src/render/draw_symbol.js
@@ -214,13 +214,7 @@ function drawTileSymbols(program, programConfiguration, painter, layer, tile, bu
         const hasHalo = !layer.isPaintValueFeatureConstant(haloWidthProperty) || layer.paint[haloWidthProperty];
         const gammaScale = (pitchWithMap ? Math.cos(tr._pitch) : 1) * tr.cameraToCenterDistance;
         gl.uniform1f(program.u_gamma_scale, gammaScale);
-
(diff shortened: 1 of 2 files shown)


d1manson:
I have just noticed a potential issue, which is that now the halo from one character can overlap with the non-halo part of the previous character (within the same piece of text).  The `text-letter-spacing` style value can be increased as a hack workaround, but it's not really what I wanted.

kkaefer:
refs https://github.com/mapbox/mapbox-gl-js/issues/596

ChrisLoer:
If we handle the other considerations that are blocking this, there's one more thing to worry about.. we're actually planning to _rely_ on halos being done in a separate draw call on gl-native: https://github.com/mapbox/mapbox-gl-native/pull/9009. The reason is that device compatibility limitations require us to have 8 or fewer vertex attributes (and using two draw calls allows us to selectively use _only_ `fill_color` or `halo_color`). Even though that requirement doesn't directly affect gl-js, we want to keep the shader implementations in sync between the two code bases.

PR #4726, opened 2017-05-19
Sanitize flyto's zoom parameter to clamp between map's minzoom and maxzoom

### LINKED ISSUE #4724: Map.flyTo() allows zooming past maxZoom
Bug report. The `flyTo()` functionality doesn't perform a sanity check on the requested zoom, so that a user can zoom past the map's `maxZoom`. 

The final zoom level is the map's `maxZoom`, but the map center jumps to an arbitrary location.

Minimal reproducible case:

```html
<div id='map'></div>
<button id='fly'>Fly!</button> 
<script>
var map = new mapboxgl.Map({
    container: 'map',
    center: [0, 0],
    zoom: 0
});
  
fly.addEventListener('click', ()=>{ map.flyTo({
  center: [10.39170933059438, 63.4309997568285],
  zoom: 22.5   // 22.5 is greater than the default maxZoom of 20
}) });
</script>
```

I've put that code in a [jsbin over here](http://jsbin.com/kapejuwive/edit?html,console,output). That jsbin uses v0.37.0.

Expected behaviour: final state of the map's camera would be zoom = map's maxzoom, center ≃ [10.39, 63.43].

Actual behaviour: final state of the map's camera would be zoom = map's maxzoom, center ≃ [89.25, 38.99] (more than 10000km away from expected center).

IvanSanchez:
Blind-ish attempt to fix #4724

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -689,7 +689,7 @@ class Camera extends Evented {
             startBearing = this.getBearing(),
             startPitch = this.getPitch();
 
-        const zoom = 'zoom' in options ?  +options.zoom : startZoom;
+        const zoom = 'zoom' in options ? util.clamp(+options.zoom, tr.minZoom, tr.maxZoom) : startZoom;
         const bearing = 'bearing' in options ? this._normalizeBearing(options.bearing, startBearing) : startBearing;
         const pitch = 'pitch' in options ? +options.pitch : startPitch;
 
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Unit tests added. Without the fix, the new tests fail like so:

>       Error: should be equal to within 1e-10
>       + expected - actual
> 
>       -201326412.0000001
>       +12

Please note that this depends on https://github.com/mapbox/mapbox-gl-js/pull/4731 for the tests to actually be useful

jfirebaugh:
Thank you!

Issue #4208, opened 2017-02-03
Disable letter-spacing for Arabic text


ChrisLoer:
Styles that increase letter-spacing don't play nice with languages where the glyphs are meant to be in contact with each other (Arabic now, Hindi and others in the future). For now this is a problem that can always be addressed with a style modification, but it's something we shouldn't make style editors think about.

Not sure if we would need to surface this change in Studio in any explicit way. Once we fix https://github.com/mapbox/studio/issues/8396, the preview should show designers what's happening. It still might be a surprise if increased letter spacing didn't do anything to Arabic text -- not that the designer would *expect* it to be spaced, but just that they wouldn't think about the difference ahead of time.

@tmcw @1ec5

1ec5:
@ajashton previously [presented and blogged](https://www.mapbox.com/blog/multilingual-character-spacing/) about an approach in CartoCSS that uses word spacing instead of letter spacing in scripts such as Arabic and Bengali.

See also:

* “[Arabic Layout Requirements](https://w3c.github.io/alreq/#h_letterspacing)”, part of a series of W3C documents on various writing systems

nickidlugash:
Also interesting: W3C documents on [letter spacing](https://drafts.csswg.org/css-text/#letter-spacing-property) and [typographic character units](https://drafts.csswg.org/css-text/#typographic-character-unit).

As a first pass, might it be ok to simply disable letter-spacing for Arabic (and other affected scripts)? I think that not having broken text on maps using our platform is an acceptable first goal, and that supporting any alternative formatting (word spacing, swashes, etc) could be done in a separate update.

> It still might be a surprise if increased letter spacing didn't do anything to Arabic text

Perhaps – personally, though, I don't think this is a big deal, at least compared to the alternative of inadvertently displaying broken Arabic text. We could perhaps just add a disclaimer to the documentation on this style spec property.

>  For now this is a problem that can always be addressed with a style modification

While this is true, and internally we've avoided using `text-letter-spacing` for styling text referencing Arabic or "local" language (e.g. the `name` field in Mapbox Streets vector tiles), even in our core styles there are occasional instances of broken Arabic because some of our "English" labels still fall back to local language. It would be nice if we could continue using letter spacing in English/latin script maps without breaking the small percentage of Arabic labels, since this type of text formatting provides really helpful context in basemaps.

ChrisLoer:
Fixed with #4727

PR #4727, opened 2017-05-19
Disable letter-spacing for Arabic text (issue #4208)

ChrisLoer:
Disables letter-spacing for Arabic text as discussed in issue #4208.

Two things to note:

* The decision to space or not to space is done on a per-feature level, instead of being done at word boundaries, which might make more sense for diglossic labels. I could make an argument that consistent spacing within the feature is a good thing, but mainly this is just significantly easier to implement.
* The PR only addresses Arabic, where there's a known problem. I haven't done an inventory of other scripts where this might be a problem. To be fully aware of when letter spacing is allowed, we'd need to work with the font (i.e. using Harfbuzz to render text).

/cc @1ec5 @nickidlugash @jcsg


1ec5:
> The PR only addresses Arabic, where there's a known problem.

That’s fine, since Arabic is the only writing system for which we support ligatures at the moment.

PR #4588, opened 2017-04-14
Roadmap last updated

asantos3026:
## Launch Checklist

 - [X] briefly describe the changes in this PR
- removed hard coded last updated date in the roadmap's page
- added an updated_at key and configured yml file
- accessed value of updated_at key in the roadmap page
- iterated through items in the roadmap yml to show in the roadmap's page

 - [ ] write tests for all new functionality

 - [ ] post benchmark scores

PR #4749, opened 2017-05-24
keep roadmap last updated date in sync, update roadmap contents

### LINKED ISSUE #4464: Keep "Last updated" value on roadmap page up to date
Currently, the "Last updated" value on the roadmap page is baked into the HTML, making it easy to forget to update it.  Ideally, it would be generated based on when the roadmap changes, but barring that, perhaps we could change the `roadmap.yml` format to include this value, so that it's easy to edit in the same place.

mollymerp:
Folding in @asantos3026's awesome addition of auto-updating date ( #4588 ) with a content refresh of the roadmap. Thanks again @asantos3026! 

close #4464

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/docs/_data/roadmap.yml
+++ b/docs/_data/roadmap.yml
@@ -5,8 +5,8 @@ roadmap_items:
     description: "Things actively being built right now."
     items:
       -
-        name: Client-side terrain rendering
-        issues: ["3730"]
+        name: Client-side hillshade rendering
+        issues: ["3730", "4701"]
       -
         name: Arbitrary expressions for property functions
         issues: ["4715"]
--- a/docs/_data/roadmap.yml
+++ b/docs/_data/roadmap.yml
@@ -1,4 +1,4 @@
-updated_at: '04/13/2017'
+updated_at: '05/24/2017'
 roadmap_items:
   -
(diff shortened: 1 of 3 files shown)


docs/_data/roadmap.yml
```diff
@@ -1,29 +1,34 @@
...
-    -
-      name: <code>Map#setCamera</code> method
-      issues: ["3583", "2801"]
-    -
-      name: Custom source type API
-      project: 2
-    - 
-      name: Patterns for loading and errored tiles
-      issues: [3280]
+updated_at: '05/24/2017'
```
andrewharvey: This can be confusing for dates like 2/10/2017 is it 2nd of Oct or 10th of Feb? See https://en.wikipedia.org/wiki/Date_format_by_country#/media/File:Date_format_by_country_(new).png

I think it'll be safer to use something like 10th Feb 2017 or Feb 10th 2017 to avoid confusion.

Scarysize: Thumbs up from Europe 😆  👍

mollymerp: Fair enough!

Issue #4708, opened 2017-05-12
Adjust line inset value for lines with `line-gap-width` in shader?


nickidlugash:
### Steps to Trigger Behavior

 1. Create a line style layer
 2. Create another line style layer with a `line-gap-width` value that equals the `line-width` value of the first line style layer

### Expected Behavior

There is no visible gap between the two lines

### Actual Behavior

There is a small visible gap between the two lines:

<img width="610" alt="screen shot 2017-05-11 at 10 12 39 pm" src="https://cloud.githubusercontent.com/assets/5607844/25979515/fb16a8b2-3696-11e7-9d89-a8005aeb8557.png">

The current shader behavior is to adjust the inset value of lines with a gapwidth to account for the antialiasing, but I think that this is causing the visible gap.

If you replace [this line](https://github.com/mapbox/mapbox-gl-js/blob/master/src/shaders/line.vertex.glsl#L58) in the line vertex shader:
`float inset = gapwidth + (gapwidth > 0.0 ? ANTIALIASING : 0.0);`

with

`float inset = gapwidth;`

There is no longer a visible gap. This does make lines a bit visibly thicker when there isn't a corresponding line whose `line-width` equals its `line-gap-width` drawn on top of it. However, I think the main use case of the `line-gap-width` property is to use it with another line drawn on top of it, so I think it may be better to optimize for that use case. If there's a solution that would be optimized for both, that would be even better.

Here's what the current behavior looks like:

<img width="870" alt="line-gap-width-current" src="https://cloud.githubusercontent.com/assets/5607844/25979459/92558988-3696-11e7-8baa-7999245d4173.png">

- Green lines have `line-width: A`
- Double maroon lines drawn under green lines have `line-gap-width: A` and `line-width: B`
- Double maroon lines not under green lines also have `line-gap-width: A` and `line-width: B`, for comparison
- Single maroon lines have `line-width: B`, for comparison

Here's what the altered behavior looks like:

<img width="880" alt="line-gap-width-no-visible-gap" src="https://cloud.githubusercontent.com/assets/5607844/25979820/ab2f433e-3698-11e7-81e3-85c97804f827.png">

We've used `line-gap-width` extensively in our core styles for a while, for roads, but I think the color contrast between the casing and the background color has always been so low that this issue wasn't really visible. We've recently been testing more high-contrast styling, and we are now finding that this is an issue. (I tested older version of mapbox-gl-js, I think this has probably always been consistent behavior).

[...]

PR #4750, opened 2017-05-24
Shift gap line inset in by .75 px to eliminate gap

lbud:
This fixes https://github.com/mapbox/mapbox-gl-js/issues/4708 by shifting gap lines in by 0.75 px — 0.5px by eliminating the addition of `ANTIALIASING`, and then an extra .5 ANTIALIASING because:
This is when we only shift it in by 0.5 (by changing it to `float inset = gapwitdth`):
screen shot 2017-05-24 at 3 07 07 pm
If you zoom in a lot, you can see white px.
This is shifting in by 0.75:
screen shot 2017-05-24 at 3 08 25 pm

Remaining:
- [ ] @nickidlugash @ansis does shifting it by a full `1.5 * ANTIALIASING` seem okay?
- [x] update render tests


lbud:
I pushed updated render tests in https://github.com/mapbox/mapbox-gl-js/pull/4750/commits/430df594d0a87e16389461199211239248e776ff — you can see that the lines do become noticeably wider (but as @nickidlugash pointed out, we rarely use gapped lines without another line layer on top).

lbud:
(We could also shift the outset inward by the same amount so that the lines don't become any wider, but if line-gap-width is always used with another line on top its intended width wouldn't be fully visible…)

ansis:
> This fixes #4708 by shifting gap lines in by 0.75 px — 0.5px by eliminating the addition of ANTIALIASING, and then an extra .5 ANTIALIASING because:
This is when we only shift it in by 0.5 (by changing it to float inset = gapwitdth):

I guess this is because if you have a 0.5 opacity pixel on top of a 0.5 opacity pixel you end up 0.75 opacity instead of 1.0.

@nickidlugash What are the use cases for a solid line on top of a gap-width line? When is #4708 encountered?

nickidlugash:
> What are the use cases for a solid line on top of a gap-width line? When is #4708 encountered?

This is how we style roads and traffic lines in our cores styles.

mollymerp:
@ansis just out of curiosity why this true? 🤔 
> I guess this is because if you have a 0.5 opacity pixel on top of a 0.5 opacity pixel you end up 0.75 opacity instead of 1.0.

ansis:
> > I guess this is because if you have a 0.5 opacity pixel on top of a 0.5 opacity pixel you end up 0.75 opacity instead of 1.0.

> @ansis just out of curiosity why this true? 🤔

If you have two sheets of tinted glass that block 50% of light and let the other 50% through, then 50% of light gets through the first sheet and of that remaining amount of light only 50% makes it through the second sheet. The total amount of light that gets through both is 50% * 50% = 25%.

Now let's say you have two sheets of clear glass that don't block light, but you cover half of one with black paper and the opposite half of the other. We know that 0% of the light will get through both sheets of glass because all light on one half is blocked by the first sheet, and all light on the other half is blocked by the other sheet. BUT, if the sheet of glass is the size of one pixel and we don't have the resolution to record which part of the sheet let light through all we can do is treat them as the tinted glass in the first example and record that they both let through 50%, which gives us the incorrect 25%.

With infinite resolution this wouldn't be a problem.

---

Thinking a bit more about the proposed change, I think there might be some problems. Something will break slightly. 😬
- if the inside edge is shifted by 1px and outside edge is shifted by 0px
    - the line width gets 1px wider, which is visible in all cases when line-gap-width is used on a standalone line
- if inside edge is shifted by 1px and the outside edge is shifted by 1px
    - all casings in existing styles that use `line-gap-width` will get 1px thinner

Are either of those ok to break?

---

> This is how we style roads and traffic lines in our cores styles.

Saman clarified that `line-gap-width` is often used used for casings to separate the concerns of how wide the line is and how wide the casing is. `line-gap-width` for the casing is made to match `line-width` of the main line, letting you style the width of the casing independently. The fact that the gap is transparent isn't the goal in this case.

### Alternative fixes:

Add `line-gap-colour` to let people set the colour for the gap and avoid this problem. (pretty much gives us `line-fill-width`, `line-fill-color`, `line-stroke-width` and `line-stroke-color` with different names and slightly weirder behaviour)

Make map designers manually subtract `1` from their `line-gap-width` values if they want to use it for this.

Provide some way for people to express values as sums of other values (casing `line-width` is fill `line-width` + new value).

nickidlugash:
> Add line-gap-colour to let people set the colour for the gap and avoid this problem

@ansis so to get the intended result for roads, would the `line-gap-color` need to be the same color as the `line-color` of that layer, or the same color as the `line-color` of the road fill layer above it? (For this use cases we would still need 2 layers for proper drawing order)

ansis:
> @ansis so to get the intended result for roads, would the line-gap-color need to be the same color as the line-color of that layer, or the same color as the line-color of the road fill layer above it? (For this use cases we would still need 2 layers for proper drawing order)

Yep

Another possibility: leaves it how it is and revisit this later. This rendering artifact doesn't look that severe. Maybe at some point far down the line we can have expressions that can use values from other layers?

nickidlugash:
>  This rendering artifact doesn't look that severe.

It's fairly obvious with a combination of high-contrast colors and lower resolution screens – carto team first noticed it for an earlier version of the [red traffic lines in this style](https://api.mapbox.com/styles/v1/mapbox/navigation-preview-day-v1.html?title=true&access_token=pk.eyJ1Ijoibmlja2kiLCJhIjoiczVvbFlXQSJ9.1Gegt3V_MTupW6wfjxq2QA#11.72/38.8877/-77.0123) on WQHD resolution monitors (good resolution for a large external monitor, but not retina).

> Another possibility: leave it how it is and revisit this later.

 For our (navigation styles) use case, it's probably ok to punt on this for a bit, in order for us to first better understand how users use `line-gap-width` before we make any decisions. My sense right now is that the most prevalent use case for it is what's been outline above: road casings (or similar), with the `line-gap-width` value matching the `line-width` value of a line drawn on top of it. Ideally then, we would support this use cases as simply as possible, without additional style properties to assign or additional calculations to keep track of.

PR #4751, opened 2017-05-25
anchor Markers at their center

### LINKED ISSUE #2900: Make markers anchored at their center by default

andrewharvey:
Closes #2900 (breaking change!)

Not sure if this is the best approach (tweaking transform translate values) or a CSS approach is best.

Currently if you change the HTML Element's size, it will loose centering until the map camera changes and triggers `Marker._update`.

## Launch Checklist

anchor Markers at their center by default per #2900 
~~~Tests aren't working, needs more work.~~~ Not sure how to test due to DOM used for tests is not behaving the same as a browser.
 - [ ] post benchmark scores
Need to check this extra calculation at each camera change doesn't add much overhead.
looks okay (including when the element has a border).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## master
 
+#### :warning: Breaking changes
+
+- Markers are now anchored at their center by default ([#4751](https://github.com/mapbox/mapbox-gl-js/pull/4751))
+
 ## 0.38.0 (June 9, 2017)
 
 #### New features :sparkles:
--- a/docs/_posts/examples/3400-01-24-custom-marker-icons.html
+++ b/docs/_posts/examples/3400-01-24-custom-marker-icons.html
@@ -88,7 +88,7 @@ geojson.features.forEach(function(marker) {
     });
 
     // add marker to map
-    new mapboxgl.Marker(el, {offset: [-marker.properties.iconSize[0] / 2, -marker.properties.iconSize[1] / 2]})
+    new mapboxgl.Marker(el)
         .setLngLat(marker.geometry.coordinates)
         .addTo(map);
 });
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -254,9 +254,7 @@ class GeolocateControl extends Evented {
         if (this.options.showUserLocation) {
             this._dotElement = DOM.create('div', 'mapboxgl-user-location-dot');
 
-            // until https://github.com/mapbox/mapbox-gl-js/issues/2900, the offset is used to ensure the Marker is centered at the
-            // user's location, half of the .mapboxgl-user-location-dot width (including border).
-            this._userLocationDotMarker = new Marker(this._dotElement, { offset: [-10, -10] });
+            this._userLocationDotMarker = new Marker(this._dotElement);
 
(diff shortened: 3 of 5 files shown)


jfirebaugh:
Seems good to me. We could switch to a CSS approach later (including adding a `anchor` option like `Popup`) if it's needed.

mollymerp:
Thanks @andrewharvey ! Can you also round the offsets to the nearest pixel? We had issues w subpixel transforms causing `Popup`s to render blurry on some screens.

andrewharvey:
> Can you also round the offsets to the nearest pixel? We had issues w subpixel transforms causing Popups to render blurry on some screens.

Actually they are already rounded at 
https://github.com/mapbox/mapbox-gl-js/blob/a266c43e42c8687b2bd7e1821adba5624707c58c/src/ui/marker.js#L166

I've pulled out this test to get it passing, for some reason in the testing infrastructure the DOM element.offsetWidth isn't working so the test fails.

```
t.test('marker centered by default', (t) => {
    const map = createMap();
    const element = window.document.createElement('div');
    element.style.width = '10px';
    element.style.height = '10px';
    element.style.border = '6px solid black';

    const marker = new Marker(element).setLngLat([0, 0]).addTo(map);
    const translate = Math.round((map.getContainer().offsetWidth / 2) - ((6 + 10 + 6) / 2));
    t.equal(marker.getElement().style.transform, `translate(${translate}px, ${translate}px)`, 'Marker centered');
    t.end();
});
```

I've manually tested and it works well even when the Marker has a padding or border.

mollymerp:
@andrewharvey I believe the failing test is due to the changes from upgrading JSDOM (16c75b8). You'll have to use `Object.defineProperty` as in https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/ui/control/attribution.test.js#L45

andrewharvey:
> @andrewharvey I believe the failing test is due to the changes from upgrading JSDOM (16c75b8). You'll have to use Object.defineProperty as in https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/ui/control/attribution.test.js#L45

Thanks for the pointer @mollymerp. I've updated the unit test now, though it is far away from a real world browser but it's better than nothing I suppose.

mollymerp:
Ooo one thing I missed @andrewharvey – I think we should pass the `offset` to the popup when its bound to a `Marker` with `setPopup`, but allow it to be overridden if the user specifies different offsets.

andrewharvey:
> Ooo one thing I missed @andrewharvey – I think we should pass the offset to the popup when its bound to a Marker with setPopup, but allow it to be overridden if the user specifies different offsets.

I think I have this working now, but I'd like to test it more and I'll need to look into a unit test.

mollymerp:
Ooh sorry for not being more clear – I meant offset the popup by the same additional amount as the marker `{x: -offsetWidth / 2, y: -offsetHeight / 2}` unless they specify alternate offsets for the `Popup`, so that its also anchored at the `Marker`'s center

PR #4752, opened 2017-05-25
Require node 6; use more ES6 features

jfirebaugh:
Require node 6, allowing us to permit use of additional ES6 features:

* default parameters
* rest parameters
* destructuring

### DIFF
--- a/src/style-spec/function/index.js
+++ b/src/style-spec/function/index.js
@@ -196,9 +196,9 @@ function evaluateExponentialFunction(parameters, propertySpec, input) {
     const interp = interpolate[propertySpec.type] || identityFunction;
 
     if (typeof outputLower === 'function') {
-        return function() {
-            const evaluatedLower = outputLower.apply(undefined, arguments);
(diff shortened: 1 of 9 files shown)


jfirebaugh:
Reverted the changes to default parameters; they turned out to be too dangerous in existing code, which may be receiving `null` rather than `undefined`. We can start using them in new code.

src/style-spec/function/index.js
```diff
@@ -196,9 +196,9 @@ function evaluateExponentialFunction(parameters, propertySpec, input) {
     const interp = interpolate[propertySpec.type] || identityFunction;
 
     if (typeof outputLower === 'function') {
-        return function() {
-            const evaluatedLower = outputLower.apply(undefined, arguments);
-            const evaluatedUpper = outputUpper.apply(undefined, arguments);
+        return function(...args) {
+            const evaluatedLower = outputLower.apply(undefined, args);
+            const evaluatedUpper = outputUpper.apply(undefined, args);
```
mourner: You can do `outputLower(...args)`.

PR #4757, opened 2017-05-25
light.anchor: zoom (piecewise-constant) functions supported

### LINKED ISSUE #4743: Light anchor documentation lacks function, transition icons
The [light anchor documentation](https://www.mapbox.com/mapbox-gl-js/style-spec/#light-anchor) in the style specification lacks the icons that denote that the property can take a function or be transitioned. This omission led to the misinterpretation in https://github.com/mapbox/mapbox-gl-native/pull/9043#discussion_r117386443 that led the iOS and macOS SDKs to make the property only accept a constant value: mapbox/mapbox-gl-native#9085.

/cc @jfirebaugh @fabian-guerra @lbud

lbud:
Fixes #4743 by adding appropriate key/value pairs to `light.anchor` documentation .

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -1705,6 +1705,9 @@
         }
       },
       "transition": false,
+      "zoom-function": true,
+      "property-function": false,
+      "function": "piecewise-constant",
       "doc": "Whether extruded geometries are lit relative to the map or viewport.",
       "example": "map",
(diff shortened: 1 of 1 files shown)


lbud:
@jfirebaugh in GL JS current master (and release) this is actually not supported because spec-based validation fails, so I think this would not be appropriate to target `mb-pages` before the next release.

Issue #4740, opened 2017-05-23
Resource transforms: alter URLs before requesting


kkaefer:
In Native, [we have added "Resource transforms"](https://github.com/mapbox/mapbox-gl-native/pull/8084), which allows you to supply a function that is called every time a URL is requested. That function can perform arbitrary operations on the URL string, and the returned URL will be requested. We should implement this in GL JS for API parity.

Scarysize:
Will this feature make it possible to have relative URLs for certain properties in the map style JSON? E.g.:

```js
{
  ...
  "spritesheet": "my/local/spritesheet"
}
```

A callback could resolve this relative url and return an absolute one.

asheemmamoowala:
@Scarysize - the changes proposed in #4763 do not explicitly allow pure relative URLs in the `style.json`. 
Instead, you can provide a custom protocol and resolve that in the transform callback:
```
{
...
"sprite": "app-local://my/local/spritesheet"
}
```

Scarysize:
That was my intention. Thanks for confirming that 🙌

volkergraf:
Will this kinda make the following Possible:

To load "local"-Tiles like 
```
"sources": {
                "openmaptiles": {
                    "type": "vector",
                    "tiles":  [location.href + "/../germany/{z}/{x}/{y}.pbf" ],
                    "name": "OpenMapTiles",
                    "format": "pbf",
                    "basename": "germany.mbtiles",
```

on iOS, I have to transform the FileURL from e.g. **location.href + "/../germany/14/1111/{2222.pbf"**
to the corresponding blob-id (_blob:file:///c415dbb6-3d57-4354-83c0-a33c0b6ddb4f_) on iOs using the function  _window.resolveLocalFileSystemURL_ provided by the **cordova-plugin-file**-Plugin. 

Thanx for any Help

asheemmamoowala:
@volkergraf I believe you are asking for #4917. 

#5021 will allow transforming the URL as you've mentioned above. Not being familiar with Cordova,  I do not know if a blob url from Cordova,  returned from `window.resolveLocalFuleSystemURL`, can be retrieved using an `XMLHttpRequest`.

volkergraf:
@asheemmamoowala Well THIS I can definively can confirm (using the cordova-file-plugin and the local-URl-Fix).. to test it I created such an URL form an Object like

[...]

PR #4763, opened 2017-05-26
Add global mapboxgl.setURLTransform

asheemmamoowala:
This PR addresses #4740 by adding a global `mapboxgl.setURLTransform` method that allows setting a single callback to transform URLs before they are requested.

The callback needs to be set before a new Map object is created so that it can be applied to requests that are generated as part of Map initialization.

  ~post benchmark scores~

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -25,12 +25,12 @@ mapboxgl.Point = require('point-geometry');
 
 mapboxgl.Evented = require('./util/evented');
 mapboxgl.supported = require('./util/browser').supported;
+mapboxgl.setURLTransform = require('./util/mapbox.js').setURLTransform;
 
 const config = require('./util/config');
 mapboxgl.config = config;
 
 const rtlTextPlugin = require('./source/rtl_text_plugin');
-
 mapboxgl.setRTLTextPlugin = rtlTextPlugin.setRTLTextPlugin;
 
  /**
@@ -78,4 +78,20 @@ Object.defineProperty(mapboxgl, 'accessToken', {
  * @example
  * mapboxgl.supported() // = true
  * @see [Check for browser support](https://www.mapbox.com/mapbox-gl-js/example/check-for-support/)
+ */
+
+/**
+ * Sets the map's URL Transform callback function
+ * Call this method to set a callback to transform URLs before they are requested
+ * from the internet. The callback can be used to add or remove custom parameters, or reroute
+ * certain requests to other servers or endpoints.
+ *
+ * @function setURLTransform
+ * @param {Function} callback Callback function called with the original URL to be transformed. Return the transformed URL
+ * @example
+ * mapboxgl.setURLTransform( (url)=> {
+ *  if(url.startsWith('https://myHost') {
+ *   return url += `&_ts=${Date.now()}`;
+ *  }
+ * });
  */
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -3,6 +3,7 @@
 
 const config = require('./config');
 const browser = require('./browser');
+const window = require('../util/window');
 
 const help = 'See https://www.mapbox.com/api-documentation/#access-tokens';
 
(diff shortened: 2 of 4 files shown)


src/index.js
```diff
@@ -25,12 +25,12 @@ mapboxgl.Point = require('point-geometry');
 
 mapboxgl.Evented = require('./util/evented');
 mapboxgl.supported = require('./util/browser').supported;
+mapboxgl.setURLTransform = require('./util/mapbox.js').setURLTransform;
```
kkaefer: Nit: We leave out the `.js` part of the filename in the `require` call

src/index.js
```diff
@@ -79,3 +79,19 @@ Object.defineProperty(mapboxgl, 'accessToken', {
...
+ * Call this method to set a callback to transform URLs before they are requested
+ * from the internet. The callback can be used to add or remove custom parameters, or reroute
+ * certain requests to other servers or endpoints.
+ *
+ * @function setURLTransform
+ * @param {Function} callback Callback function called with the original URL to be transformed. Return the transformed URL
+ * @example
+ * mapboxgl.setURLTransform( (url)=> {
+ *  if(url.startsWith('https://myHost') {
+ *   return url += `&_ts=${Date.now()}`;
```
kkaefer: I know this is just an example, but can we change it so that we also check for presence of `?` before blindly appending `&`?

src/util/mapbox.js
```diff
@@ -113,5 +129,13 @@ function parseUrl(url: string): UrlObject {
...
-    return `${obj.protocol}://${obj.authority}${obj.path}${params}`;
+    const url = `${obj.protocol}://${obj.authority}${obj.path}${params}`;
+    return transformURL(url);
 }
+
+exports.resolveURL = function(url: string): string {
+    url = transformURL(url);
+    const a = window.document.createElement('a');
+    a.href = url;
+    return transformURL(a.href);
```
kkaefer: Do we need to go through `a.href` again with the transformed URL? Or only with the transformed URL?

kkaefer: AFAIK, this is used to resolve relative URLs to absolute URLs. I think both versions (resolve only after, or resolve both before and after the transform) could be valid, and I'm not sure which one is best.

jfirebaugh:
> The callback needs to be set before a new Map object is created so that it can be applied to requests that are generated as part of Map initialization.

It could be a [map option](https://github.com/mapbox/mapbox-gl-js/blob/cb34199b947c929f87da1d574b22b830c52bf903/src/ui/map.js#L73). This is more flexible than a global (different maps could have different transforms).

src/util/mapbox.js
kkaefer: I think it might make sense to only do the URL transform *before* resolving it. This allows developers to use any URLs they want in the stylesheet, and requires them to be a resolvable URL only after they were transformed. E.g. on Native, we sometimes use custom URL *schemes* to indicate that this URL needs to be transformed. Going through the resolver before transformation means that they are going to be invalid, or at least in an unknown state.

src/index.js
asheemmamoowala: I wondered about that. Makes sense to have it be complete.

src/util/mapbox.js
asheemmamoowala: @kkaefer - I just realized that this snuck through incorrectly. The intent was for it to be a post resolve transform only. I imagine that developers might want to take advantage of the transform, inspect, or validate fully resolved URLs. With a partial URL, I would find it harder to transform URLs if I don't know the host they might resolve to. This is useful when deploying code to different environments, or bundling a plugin where you don't always know what environment it will be running in.

asheemmamoowala:
> It could be a map option. This is more flexible than a global (different maps could have different transforms).
My initial thought was to  set the transform on the `Map` object as well. After going down that path for a bit, I found that it disperses the calls to `transform` over many different source files and "dirties" the source IMO. 
Using a single global centralizes this logic and allows reliable application to URLs.

jfirebaugh:
It's easier today to write it using what's effectively a global variable, yes. But one almost always regrets the global variable shortcut later...

src/util/mapbox.js
kkaefer: We're now transforming the URL twice, once before resolving, and once after resolving. `resolveURL` is only used for GeoJSON sources that can reference GeoJSON files relative to the web page's location. Using the same `transformURL` call means that they sometimes get relative URLs (pre-resolving), and sometimes resolved URLs with a full scheme and hostname. It also means that we're calling `transformURL` twice for a single URL when it's a GeoJSON source, compared to only once for all other URLs.

Issue #4158, opened 2017-02-01
Change units of line-dasharray to pixels


lucaswoj:
_From @mollymerp on December 22, 2016 16:45_

After discussing with @ansis and @jfirebaugh it seems that the most straight-forward way to make data-driven `line-width` and `line-dasharray` possible, would be to change the units of `line-dasharray` to absolute pixels (from relative units of `line-width`) because
>Dashed lines rely on [scaling the dash length](https://github.com/mapbox/mapbox-gl-js/blob/d5c33fdb806a118946c8167c3412e5fba49d8cf3/js/style/style_layer/line_style_layer.js#L14-L19) by the line width at the previous round zoom level.

This would introduce some complexity in stying lines with `line-dasharray` because they would need additional zoom functions to stay in sync with line-width to make sure the `line-dasharray` appears consistent as the map zooms. This seems like a reasonable tradeoff to keeping property function evaluation consistent across all paint properties (we'd need to introduce many exceptions to the current code path if `line-dasharray` stays a relative value). 

Interested to hear thoughts on this! cc @lucaswoj @nickidlugash

Previous discussion on the possible approaches to solving this problem here: https://github.com/mapbox/mapbox-gl-js/issues/3682

_Copied from original issue: mapbox/mapbox-gl-style-spec#633_

lucaswoj:
_From @1ec5 on December 22, 2016 19:22_

This would also be good for parity with MapKit in the iOS and macOS SDKs. (MapKit’s `MKOverlayPathRenderer.lineDashPattern` property is measured in screen points.)

lucaswoj:
_From @nickidlugash on December 22, 2016 19:34_

@mollymerp sounds like this is the best option. 

If we do this, is it possible to also refactor the function behavior of `line-dasharray` to interpolate at integer zoom levels (similar to the behavior of layout number properties), as opposed to being piecewise? I believe that the reason this wasn't done in the first place is that line dasharrays are rendered like line patterns, so we matched the function behavior of line patterns. However, since we are generating the line-dasharray images in the renderer, would it not be possible to generate a new line-dasharray image at every integer zoom level, based on the interpolated value?

If it is possible to implement this ^, I think it would greatly reduce the burden on designers for matching `line-dasharray` to `line-width`, because it would then be possible to match the stops and base values. I think this would also more closely match user expectations for the `line-dasharray` values (I think it's currently the only number property that is piecewise).

[...]

Issue #3045, opened 2016-08-22
Implement property functions for "line-dasharray" and "line-cap"


lucaswoj:
Implementing property functions for `line-dasharray` will require sharing information about `LineAtlas` with the worker, much like the `*-pattern` properties https://github.com/mapbox/mapbox-gl-js/issues/2732

lucaswoj:
This is currently blocked by implementation difficulties. `line-dasharray` is measured in units relative to `line-width`. We may need to change the units of `line-dasharray` to pixels #4158 before proceeding.

lbud:
Re: `line-dasharray`: from https://github.com/mapbox/mapbox-gl-js/issues/4434
> High priority; high difficulty. Perhaps not feasible to support.

One thing we actually might consider, now that the line-width scaling happens in the shader and acts essentially like its own property function, is adding a new property like `line-dasharray-units: ["width", "pixels"]` (default: width). This would allow users using data-driven line-width to maintain full control over line-dasharray. I think this would be pretty easy to support after https://github.com/mapbox/mapbox-gl-js/pull/4773 — `floorwidth` would just need to be 1 instead of the line-width values evaluated at floor(z).

This would enable something like
![image](https://user-images.githubusercontent.com/3170312/27657467-9c572ece-5c01-11e7-8dea-6784eb7b33e1.png)
(where line-width is property-driven (5, 10, 15); line-dasharray is always [10px, 10px])

mollymerp:
@lbud we talked about a variation of this (adding a new `line-dasharray-pixels`) property, but I like your idea better.

lbud:
(we could also use something like `[relative, absolute]`, which is more recognizable/memorable but less explicit…)

lbud:
I edited the title of this because `line-width` was implemented, and `line-cap` implementation is intertwined with `line-dasharray`.

jfirebaugh:
Is it possible to have one line atlas and line atlas texture for each tile, like we're planning to do for [glyph and icon atlases](https://github.com/mapbox/mapbox-gl-js/issues/4876)?

lbud:
@jfirebaugh probably. That might be helpful in the approach I'm working on for data-driven line-caps then (which is to create both round and non-round lines in the line atlas for each dash pattern and be able to pick the right one in the shader, which will increase the line atlas area used…) 🕵️‍♀️

lbud:
[...]

PR #4773, opened 2017-06-01
Enable property functions for line-width

lbud:
Ref #3045 

The challenge in implementing data-driven `line-width` has been that `line-dasharray` is scaled by `line-width`, so dasharray calculations require the value of a line's width at floor(zoom). This PR accomplishes data-driven `line-width` by moving that calculation to the shader and adding an additional, sort of phantom paint attribute that calculates the value of line-width at floor(zoom). This required a bit of special casing in program_configuration but given the existing situation seemed to me the best option. This would nullify #4158.  @ansis @jfirebaugh @nickidlugash @anandthakker — does this seem reasonable?
 
I verified manually with a matrix of 
`{ line-width: [constant, zoom, property, zoom-and-property] }` 
x 
`{ line-dasharray: [none, constant, zoom] }` 
and added render tests here to the same effect (the naming in the render suite is a bit verbose 💁).

## Launch Checklist

 - [x] gl-native companion PR -> https://github.com/mapbox/mapbox-gl-native/pull/9214/


src/data/bucket/line_bucket.js
```diff
@@ -50,6 +50,8 @@ const lineInterface = {
         {property: 'line-opacity', multiplier: 10, type: 'Uint8'},
         {property: 'line-gap-width', multiplier: 10, type: 'Uint8', name: 'a_gapwidth'},
         {property: 'line-offset', multiplier: 1, type: 'Int8'},
+        {property: 'line-width', multiplier: 10, type: 'Uint8', name: 'a_width'},
+        {property: 'line-width', multiplier: 10, type: 'Uint8', name: 'a_floorwidth'},
```
ansis: This approach looks great!

I wonder if adding a `floorZoom: true` to the property definition here and using that to making the `if (name === 'a_floorwidth')` exceptions less exceptional would be good. Or if that would just hide the exception in a bad way.

lbud: 👍 I like that idea

mollymerp:
@lbud awesome 🕵️‍♀️ !!

keyofj:
Thanks for making this a priority. Dynamic line-width is an important feature.

test/integration/render-tests/line-dasharray/function--line-width-composite-function/style.json
```diff
@@ -0,0 +1,150 @@
...
+    52.499167
+  ],
+  "zoom": 14,
+  "sources": {
+    "mapbox": {
+      "type": "vector",
+      "maxzoom": 14,
+      "tiles": [
+        "local://tiles/{z}-{x}-{y}.mvt"
+      ]
```
anandthakker: I'd be in favor of simplifying these render tests using some bare-bones GeoJSON, as that'll make em easier to understand/assess.

lbud:
@anandthakker thanks for the suggestion — simplified geojson fixtures in  https://github.com/mapbox/mapbox-gl-js/pull/4773/commits/467ed46c71aea0dd638140b8618e4e955623ea93

Issue #4361, opened 2017-03-03
Style spec: Add ability to sort features


kkaefer:
With data driven styling, we can render more features in a single layer. However, this means that the order of the features in the source dictates the rendering order. It'd be useful to add a capability to sort/group features before they are parsed into a bucket.

The sorting step would come after the filter step and before the features are transformed into actual buffers. Here's a style syntax suggestion:

```json
{
  "id": "place-city-lg-s",
  "type": "symbol",
  "source": "composite",
  "source-layer": "place_label",
  "filter": [
    "all",
    ["<=", "scalerank", 2],
    ["in", "type", "city"],
    ["in", "ldir", "S", "SE", "SW", "E"]
  ],
  "sort": [
    ["localrank", ">"]
  ]
}
```

```json
{
  "id": "landcover",
  "type": "fill",
  "source": "composite",
  "source-layer": "landcover",
  "filter": [
    "all",
    ["in", "class", "crop", "grass", "scrub", "wood", "snow"]
  ],
  "sort": [
    ["class", ["crop", "grass", "scrub", "wood", "snow"]],
    ["area", "<"]
  ]
}
```

Sort rules are applied in order, so if items have an identical `class` , we're going to sort them by `area` in descending order. In the array, we're first supplying the sort field, then the sort order. We can either specify `<` for descending, `>` for ascending, and an array with explicitly listed items for discrete value ordering (all items that do not match a discrete value are going last and will have subsequent sort rules applied).

See https://github.com/mapbox/mapbox-gl-js/issues/1349 for a similar discussion.

stevage:
Small point, I'd suggest using a string like "asc" rather than subverting comparison operators (">"). They look like the filter attributes (which are comparisons) but the behaviour is unrelated.

It might even be better to use a different term which makes clearer the relationship between the sort order and the display order. I don't necessarily know whether an "ascending" order means that ID 5 is drawn above or below ID 3. I don't know what that term would be though :)

1ec5:
In the iOS and macOS SDKs, it may be desirable to expose this sort option to the runtime styling API as a `sortDescriptors` property on MGLVectorSourceLayer that holds an array of [NSSortDescriptor](https://developer.apple.com/reference/foundation/nssortdescriptor)s, for consistency with how we’re already representing filters as NSPredicates.

nextstopsun:
[...]

PR #4777, opened 2017-06-02
Project Arbex (Arbitrary Expressions)

anandthakker:
## Intent
Design and implement a style specification syntax for defining "style functions" using arbitrary expressions.  Such functions should be usable for:
 - Style layer property values, wherever zoom functions and property functions are currently supported.
 - Style layer filters
 - [future] Intra-layer sorting and grouping of features https://github.com/mapbox/mapbox-gl-js/issues/4361

## Background:

Initial discussion: https://github.com/mapbox/mapbox-gl-js/pull/4715
Development of working draft: https://github.com/mapbox/mapbox-gl-js/pull/4754

## Important use cases that the final design should cover:
- unit conversion
- number formatting
  - decimal precision
  - thousands separator
- conditionals (for the token defaults use case)
- concatenation (for the token defaults use case)
- customizable interpolation
- arithmetic on feature properties within functions
- localization

## Plan
- [x] Develop working draft of spec
- [x] GL JS Implementation #4841 
- Tail work for GL JS implementation:
  - [x] Implement cubic-bezier interpolation type for curve expressions
  - [x] Drop new `bezier-easing` dependency in favor of pre-existing `unitbezier`
  - [ ] Reinstate support for color space interpolation option
  - [x] Check out how native handles validation for a property function whose stop outputs are of the wrong type (see https://github.com/mapbox/mapbox-gl-js/pull/4841#issuecomment-310491857)
  - [x] Refactor style function call sites to pass (globalProperties, feature) rather than (globalProperties, featureProperties) (needed for geometry_type and id)
  - [x] Generalize `match` expression to support multiple keys per output value
  - [x] Guard against JS keywords in let-bound variable names
  - [x] Update {text,icon}-size DDS path to use expressions
  - [ ] Benchmark.
- [x] Add documentation for each expression into style-spec JSON and the generated docs pages.
- [x] Cross-platform test suite
- [ ] GL Native Implementation https://github.com/mapbox/mapbox-gl-native/pull/9439
- [ ] SDK Implementations

# Benchmarks

Benchmark scores:

benchmark | master 45fde16 | feature/expressions 3fc994c
--- | --- | ---
**map-load** | 148 ms  | 121 ms 
**style-load** | 76 ms  | 132 ms 
**buffer** | 1,058 ms  | 1,091 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 6 ms, 0% > 16ms  | 8 ms, 1% > 16ms 
**query-point** | 0.85 ms  | 0.84 ms 
**query-box** | 51.25 ms  | 62.33 ms 
**geojson-setdata-small** | 7 ms  | 9 ms 
**geojson-setdata-large** | 135 ms  | 151 ms 

<del>Looking at the dds-specific benchmark being added in https://github.com/mapbox/mapbox-gl-js/pull/5207, which exercises tile layout with many zoom-and-property functions, looks like about a 2x hit using expressions compared to existing functions:</del>

**update:** After a couple quick fixes, it's more like 1.5x instead of 2x:

screen shot 2017-08-30 at 11 55 30 am

@mapbox/gl @mapbox/studio @mapbox/cartography-cats @mapbox/support


docs/style-spec/expressions.md
```diff
@@ -0,0 +1,174 @@
...
+- `["length", e: JSONArray|Vector<T>|String] -> Number`
+
+###Feature data:
+- `["properties"] -> JSONObject` the feature's `properties` object
+- `["geometry_type"] -> String` the string value of `feature.geometry.type`
+- `[ "id" ] -> Value` returns the value of `feature.id`.
+
+###Decision:
+- `["case", cond1: Boolean, result1: T, cond2: Boolean, result2: T, ..., cond_m, result_m: T, result_otherwise: T] -> T`
+- `["match", x: T, a_1: T, y_1: U, a_2: T, y_2: U, ..., a_m: T, y_m: U, y_else: U]` - `a_1`, `a_2`, ... must be _literal_ values of type `T`.
```
jfirebaugh: Can we generalize this so that multiple literal values can be mapped to the same output expression?

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,174 @@
...
+- `Null`
+  - Literal: `null`
+- `String`
+  - Literal: `"my string value"`
+- `Number`
+  - Literal: `1729`
+- `Boolean`
+  - Literal: `true` or `false`
+- `Color`
+- `JSONObject`
```
jfirebaugh: Let's just call it `Object`.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,174 @@
...
+- `Array<T, N>`: an array of N values of type T
+  - Literal: `["array", v0, v1, ...]`
+- `Vector<T>`: a dynamically sized array of values of type T
+  - Literal: `["vector", v0, v1, v2, ... vN]`
+  - TODO: without type inference, 0-length arrays and vectors can't be typed
+- `Value`: A "variant" type representing the set of possible values retrievable from a feature's `properties` object (`Null | String | Number | Boolean | JSONObject | Vector<Value>`)
+- `Error`: a subtype of all other types. Used wherever an expression is unable to return the appropriate supertype. Carries diagnostic information such as an explanatory message and the expression location where the Error value was generated.
+
```
jfirebaugh: What about enumerated types? I think we'll find that leaving them stringly-typed is less than ideal.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,174 @@
...
+- `["string", e: Value] -> String`
+- `["number", e:Value] -> Number`
+  - Uses platform-default string-to-number conversion. (TBD: parse locale-specificformatted number strings)
+- `["boolean", e:Value] -> Boolean`
+  - `0`, `''`, and `null` are converted to `false`, all other values to `true`
+
+###Lookup:
+- `["get", obj: JSONObject, key: String ] -> Value`
+- `["has", obj: JSONObject, key: String ] -> Boolean`
+- `["at", arr: JSONArray, index: Number] -> Value`
```
jfirebaugh: This is redundant to the polymorphic version on the next line, right?

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,174 @@
...
+###Math:
+All of the following take `Number` inputs and produce a `Number`.
+- +, -, \*, /, %, ^ (e.g. `["+", expr1, expr2, expr3, …]`, `["-", expr1, expr2 ]`, etc.)
+- log10, ln, log2
+- sin, cos, tan, asin, acos, atan
+- ceil, floor, round, abs
+- min, max
+
+###String:
+- `["concat", expr1: T, expr2: U, …] -> String`
```
jfirebaugh: `["concat", expr: String, …] -> String`?

docs/style-spec/expressions.md
anandthakker: Whoops, yep thanks.

docs/style-spec/expressions.md
anandthakker: Good idea 👍

docs/style-spec/expressions.md
anandthakker: I was thinking it might be okay, as a convenience, for `concat` to implicitly convert all of its arguments, on the premise that this should always be possible without error as long as the value isn't an Error.

jfirebaugh: Is there a particular reason to support implicit conversion for `concat` but not other expressions (`+`, `&&`, `up/downcase`, ...)?

anandthakker: Not a hugely compelling one... just the combination of:
 - Combining string literals, string property values, and numeric property values will be a very common use case (i.e. the equivalent of using `{}` token replacement for `text-field`
 - Automatic conversions to `String` will never fail for values that aren't already `Error`, so the explicit cast doesn't seem to me to add any value.

ansis: Would there by any advantage to having a string formatting function instead of concat?

On one hand it would add complexity because it would add function-specific syntax. On the other hand you could handle the entire "this is what I want the displayed text to be" in a single function, which might be more UI-friendly.

docs/style-spec/expressions.md
anandthakker: @jfirebaugh Can you say more about why you think we’ll want these over strings?  specifically: are there cases where such types would be useful for checking what a _subexpression_‘s evaluates to? 

So far, the use cases I can think of for enums are for checking the _final_ output of an expression that's being used for an enum-typed style property like `symbol-placement`.  If it's true that this is the main reason for them, then the question is which is simpler/better: adding enums to the type system here, or treating the validation of style property values as a separate step?

Since there are, e.g., numeric properties like `*-opacity` that can't really be validated via types (at least not normal ones), seems like we're gonna have the extra validation step anyway, which makes me lean towards the latter.

jfirebaugh: Could be wrong, just a gut feeling that we'll run into issues in practice, or as we implement it (especially in native). I don't have any concrete examples.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,173 @@
...
+      0, [ "get", ["properties"], "rating", "Number" ],
+      // zoom is 20 and "rating" is 0 -> circle radius will be 4 * 0 = 0px
+      // zoom is 20 and "rating" is 5 -> circle radius will be 4 * 5 = 20px
+      10, [ "*", 4, [ "get", ["properties"], "rating", "Number" ] ]
+    ]
+  }
+}
+```
+
+## Property Expressions
```
1ec5: This document refers exclusively to “expressions”, but the title of this PR refers to “arbitrary expressions”. To avoid confusion, we should settle on one name for the feature and eliminate the other name. As catchy as “Arbex” sounds, I prefer simply “expressions” for these reasons:

* For Studio users unfamiliar with the style specification, “arbitrary” may not be perceived as a positive trait.
* “Arbitrary expressions” would seem to imply a non-arbitrary expression feature that doesn’t exist.
* “Expressions” aligns with common usage on the iOS platform and in other kinds of software.

/cc @dasulit

docs/style-spec/expressions.md
anandthakker: Simplest syntax for this would be:

`["match", x: T, [a_1: T, a_2: T, ... a_N: T], y_a: U, [b_1: T, b_2: T, ..., b_M], y_b: U, ..., y_else: U]`

Only bad thing about this is that it would take us away from the purity of every array in the JSON representing an expression (since we could consider the `curve` types like `["linear"]` to be expressions of an private `CurveType` type).  Is maintaining such syntax purity worthwhile?  Some ways we could do that:

1. `["match", x: T, [ "vector", a_1: T, a_2: T, ... a_N: T], y_a: U, [ "vector", b_1: T, b_2: T, ..., b_M], y_b: U, ..., y_else: U]`.  But allowing a `"vector"` constructor may not be a great idea, since we don't have a way for users to specify arbitrary types.  

2. We could use `"json_array"` instead -- but that's not great type-wise, since it specifically produces `Vector<Value>`, not `Vector<T>`.

3. A specialized `"match_case"` expression.

jfirebaugh: I think we should use the simplest syntax. It's fine if `"match"` is a special form. (`case` is in scheme too.)

ivovandongen:
@anandthakker Could we consider using a (formal) specification for the expression syntax? Something like [ABNF](https://tools.ietf.org/html/rfc5234) would let us use existing parser generators like [ANTLR](http://www.antlr.org/) and others to generate platform specific bindings and also parts of the core implementations and tests.

anandthakker:
@ivovandongen Yeah, I've definitely been pondering this.  There are some things that would make it challenging to specify expressions fully and formally:
 - It's a JSON 'syntax', not text, so I don't think existing methods (like BNF) would work.
 - Thus far, we've had no need to include a way to represent every _type_ in the system.  Types exist, but only in the implementation--not in the language itself.  In order to formally specify each expression in a spec, we'd need to either introduce a serialized representation of types, so that we could specify, e.g., that the `'length'` expression is typed `(Array<T>, String) => Number`.
 - Each expression has not only unique evaluation behavior, but also potentially custom parsing needs.  That's not a dealbreaker for a formal spec, but it does limit how much useful code we can generate from such a spec.

Beyond that, it's also just more machinery in general.  As such, when @jfirebaugh and I discussed this yesterday we landed on the notion that, at least for now, we could just use the module in GL JS that defines each expression (found at `src/style-spec/function/definitions/index.js`) as the "source of truth".  Even though that module's first purpose is to actually provide the GL JS implementation of each expression, we can also use it to build scripts to generate docs and core / SDK code.  If/when that becomes problematic, we can always revisit the question of moving the 'source of truth' into a more complete specification within the style spec reference JSON.

src/style-spec/function/expression.js
```diff
@@ -0,0 +1,362 @@
...
+
+        const bindings: Array<[string, Expression]> = [];
+        for (let i = 0; i < args.length - 1; i += 2) {
+            const name = args[i];
+            const key = context.path.concat(i + 1).join('.');
+            if (typeof name !== 'string')
+                throw new ParsingError(key, `Expected string, but found ${typeof name} instead`);
+
+            if (context.definitions[name])
+                throw new ParsingError(key, `"${name}" is reserved, so it cannot not be used as a "let" binding.`);
```
jfirebaugh: Since this is getting compiled to `function (name, ...) { ... }`, we need to either prohibit JS keywords as well, or sanitize names, say by prefixing them with `_`.

anandthakker: 👍  yep, tracking that on my list of tail work but forgot to add it to the top of this ticket.  Adding now.

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,195 @@
...
+- `Value`: A "variant" type representing the set of possible values retrievable from a feature's `properties` object (`Null | String | Number | Boolean | Object | Array`)
+- `Error`: a subtype of all other types. Used wherever an expression is unable to return the appropriate supertype. Carries diagnostic information such as an explanatory message and the expression location where the Error value was generated.
+
+### Constants:
+- `[ "ln2" ] -> Number`
+- `[ "pi" ] -> Number`
+- `[ "e" ] -> Number`
+
+### Variable binding:
+- `[ "let", name1: String, e1, name2: String, e2, ..., e_result ]` - Bind expression `e1` to the string `name1`, `e2` to `name2`, etc., before evaluating `e_result`.  The bound expressions may be referenced within `e_result` with `[ name1 ]`, `[ name2 ]`, etc. (E.g.: `["let", "a", 1, "b", ["number", ["get", "blah"]], [ "+", ["a"], ["b"] ]`.)
```
jfirebaugh: I don't think we want to use `[name1]` as the syntax for this -- that would mean that any new expression form we want to add in a future version would potentially break existing expressions that used that name as an identifier.

anandthakker: 🤦‍♂️ yeah.  `["ref", name1]`? `["var", name1]`?  Less concise, but actually might allow us to avoid having special logic for var reference expressions.

jfirebaugh:
Continuing to look at how to remove `TypeName` from the `Type` variant list...

Besides `match`, `case`, `curve`, and `coalesce`, the expression forms with "special" typing rules are:

* The forms that take a variable number of arguments of a single type: `concat`, `+`, `*`, `&&`, and `||`. These are fairly easy to handle with a special case `varargs` signature type, a simplified version of `nargs` which only `CompoundExpression#typecheck` needs to handle (`type Signature = Array<Type> | Varargs`).
* `length`: the array signature wants to match an array of any type. This seems like it should be handled by the array subtyping rules somehow.
* `at`: the output type is the element type of the input array type. This is the closest to "inference" we seem to get. Maybe it should just have a custom `typecheck` override?

Another thing I noticed is that `typecheck` operates one level "deeper" than typing judgements are usually expressed. Typing judgements are usually of the form

> let t1 be the type of subexpression a, t2 be the type of subexpression b, if \<some predicate on t1 and t2\>, then the type of `[expr, a, b]` is type t3

Note that there's no "expected" type input to this judgement: if you need to check that t3 is a subtype of some expected type, than you do that externally to checking `expr` itself. I think our typechecking code would be clearer if it was written like this.

anandthakker:
@jfirebaugh 👍 I came to similar tentative conclusions re: `Varargs` and `length`.  For `at`, I was thinking of two possibilities:
 - `at: (Number, Array) => Value` (`Array` being `Array<Value>`).  As with `length`, array subtyping should allow any array here (though I think currently, `Value` might, incorrectly, fail to match `Array<T>` for `T != Value`).
 - Promote `at` to be a special form with custom typecheck.

> typecheck operates one level "deeper" than typing judgements are usually expressed

Oh yeah, good point -- this was originally the case to help make type inference of generics more straightforward, but now that we're eliminating that need, we can simplify it 🙌

src/style-spec/function/definitions/index.js
```diff
@@ -0,0 +1,409 @@
...
+            // or ['let', ..., ['coalesce', ['curve', interp, ['zoom'], ...], ... ] ]
+            if (
+                !/\[2\]$/.test(context.key) ||
+                !/^(let\.result:|coalesce:)*curve$/.test(ancestors)
+            ) {
+                throw new ParsingError(
+                    context.key,
+                    'The "zoom" expression may only be used as the input to a top-level "curve" expression.'
+                );
+            }
```
anandthakker: This check should happen in `function/index.js` rather than here, because this restriction on `[zoom]` only applies to expressions used for style property values (as opposed to, say, filters)

src/style-spec/function/index.js
```diff
@@ -1,307 +1,181 @@
...
+const {CompoundExpression} = require('./compound_expression');
+const Curve = require('./definitions/curve');
+const Coalesce = require('./definitions/coalesce');
+const Let = require('./definitions/let');
+
+import type {Expression} from './expression';
+
+export type StyleFunction = (zoom?: number, featureProperties?: {}) => any;
+
+type FunctionSpecification = {|
```
jfirebaugh: I just opened #5113 to add something similar to `flow-typed/style-spec.js`.

src/style-spec/function/definitions/match.js
```diff
@@ -0,0 +1,121 @@
...
+            branches.push([(labels: any), result]);
+        }
+
+        const input = parseExpression(args[0], context.concat(1, inputType));
+        if (!input) return null;
+
+        const otherwise = parseExpression(args[args.length - 1], context.concat(args.length, outputType));
+        if (!otherwise) return null;
+
+        assert(inputType && outputType);
```
jfirebaugh: `match` with only two arguments triggers this assertion -- `inputType` is undefined.

src/style-spec/function/definitions/match.js
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            const labelContext = context.concat(i + 1);
+            if (labels.length === 0) {
+                return labelContext.error('Expected at least one branch label.');
+            }
+
+            for (const label of labels) {
+                if (label !== null && typeof label !== 'number' && typeof label !== 'string' && typeof label !== 'boolean') {
+                    return labelContext.error(`Branch labels must be null, numbers, strings, or booleans.`);
```
jfirebaugh: I think we can remove `null` from the allowed types. There's no assertion expression for null -- you can't write `['match', ['null', ['get', 'x']], ...]`, so the only valid case is `['match', null, null, x, otherwise]`, which isn't useful.

(This comment doesn't apply if we end up allowing the first argument type to be `Value` -- and I think we should consider that as one of the first "relax typing rules"/"allow more inference" enhancements we make.)

src/style-spec/function/definitions/match.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    compile() {
+        const input = this.input.compile();
+        const outputs = [`function () { return ${this.otherwise.compile()} }.bind(this)`];
+        const lookup = {};
+
+        for (const [labels, expression] of this.branches) {
+            for (const label of labels) {
+                lookup[`${typeOf(label).name}-${String(label)}`] = outputs.length;
```
jfirebaugh: Now that `match` is strictly-typed, we can replace the type-prefixing of keys in the lookup object with a preflight check of `i` against the desired type.

Additionally, we should build the lookup table at parse time and verify that all branches are reachable (i.e. no label is repeated).

jfirebaugh:
Added some `match` tests in https://github.com/mapbox/mapbox-gl-js/commit/b3e853a502becc0e02e770c17f19e162f7f862a5; feel free to merge once passing.

src/style-spec/function/convert.js
```diff
@@ -0,0 +1,192 @@
...
+        expression = ['curve', ['step'], ['zoom']];
+    } else if (type === 'exponential') {
+        const base = parameters.base !== undefined ? parameters.base : 1;
+        expression = ['curve', ['exponential', base], ['zoom']];
+    } else {
+        throw new Error(`Unknown zoom function type "${type}"`);
+    }
+
+    for (const stop of parameters.stops) {
+        expression.push(stop[0], stop[1]);
```
anandthakker: Ah - interesting!

src/style-spec/function/types.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+
+export type Type = PrimitiveType | ArrayType // eslint-disable-line no-use-before-define
```
anandthakker: Agreed! Intended to bring this over from native, just hadn't done it yet.

src/style-spec/function/expression.js
```diff
@@ -0,0 +1,253 @@
...
+ * @private
+ */
+function match(
+    expected: Type,
+    t: Type,
+    context?: ParsingContext
+): ?string {
+    let error = `Expected ${expected.name} but found ${t.name} instead.`;
+
+    // a `null` literal is allowed anywhere.
```
anandthakker: Yeah, I've been uneasy about this from the start.  The main reason I put it in place was that I think we need a way for an expression to (intentionally) fail to produce a value of its output type, e.g. `[match, input, 'a', A, 'b', B, 'c', C, null]` (where you want an enclosing `coalesce` to handle the case where the input isn't 'a', 'b', or 'c').

One alternative would be to type such an expression as `Value`, though this would probably then mean more type annotations, `[number, [match, input, 'a', A, 'b', B, 'c', C, null]]`.

Another would be to introduce an explicit `[error]` expression (and a corresponding bottom type `Error`).

src/style-spec/function/types.js
```diff
@@ -0,0 +1,66 @@
...
+    'String': StringType,
+    'Number': NumberType,
+    'Boolean': BooleanType,
+    'Object': ObjectType,
+    'Color': ColorType,
+    'Value': ValueType
+};
+
+const arrayPattern = /^Array(<([^,>]+)(\s*,\s*([0-9])+)?>)?$/;
+function parseType (type: string): Type | null {
```
anandthakker: Ahh.  Yeah, this is what I really wanted, deep down :)

src/style-spec/function/definitions/match.js
anandthakker: Tangentially: I liked your earlier suggestion that we only allow number and string as match labels, since `boolean` labels would be better represented in a `case` expression.  Did you have second thoughts about that?

src/style-spec/function/expression.js
jfirebaugh: > a way for an expression to (intentionally) fail to produce a value of its output type, e.g. `[match, input, 'a', A, 'b', B, 'c', C, null]` (where you want an enclosing `coalesce` to handle the case where the input isn't 'a', 'b', or 'c')

If the input isn't 'a', 'b', or 'c', but is still a string, then you can just use whatever value you would have used in the coalesce fallback in place of `null`. So the case I think you're getting as is where `input` may produce a string other than 'a', 'b', or 'c' _or_ may produce an error. In that case, you could write `[coalesce, [match, input, ..., O], O]` or `[let, o, O, [coalesce, [match, input, ..., o], o]]`. Do you think this is a common enough situation to require the slightly more succinct syntax with `null`?

If we do need this, I think `[error]` is the best mechanism.

src/style-spec/function/definitions/match.js
jfirebaugh: No second thoughts, let's disallow that too.

src/style-spec/function/expression.js
anandthakker: Yeah, the case you spelled out is the one I had in mind.  Hard to be anything like sure, but my guess is that this will be relatively common, because datasets with missing or poorly typed property values are quite common.  That said, with the `let` option allowing us to avoid writing the fallback expression more than once, I'm okay with our just recommending `[let, o, O, [coalesce, [match, input, ..., O], O]` -- we can always revisit if this turns out to be common and too inconvenient.

anandthakker: Oh, wait -- what if the user wants the whole expression to produce an error if the input isn't one of 'a', 'b', or 'c'?

jfirebaugh: What would be a context where that would be desired?

anandthakker: E.g. when you want the style property to fall back to its spec-defined default.  This case occurs when trying to translate a `type: "categorical"` function to an expression (particularly in `native`, where we don't have access to the spec default at conversion time)

docs/style-spec/expressions.md
```diff
@@ -0,0 +1,196 @@
...
+- `["length", e: Vector<T>|String] -> Number`
+
+### Feature data:
+- `["properties"] -> Object` the feature's `properties` object
+- `["geometry_type"] -> String` the string value of `feature.geometry.type`
+- `[ "id" ] -> Value` returns the value of `feature.id`.
+
+### Decision:
+- `["case", cond1: Boolean, result1: T, cond2: Boolean, result2: T, ..., cond_m, result_m: T, result_otherwise: T] -> T`
+- `["match", x: T, a_1: T, y_1: U, a_2: T, y_2: U, ..., a_m: T, y_m: U, y_else: U]` - `a_1`, `a_2`, ... must be _literal_ values of type `T`.
```
nickidlugash: @anandthakker Do you think it makes sense to consider making `y_else: U` optional? Could the renderer maybe treat `y_1` (or maybe `y_m`) as this value if it's not explicitly assigned? A common usage of "match" in our core styles is using it to assign style values across `a` values that have already been explicitly filtered for in that style layer, so we know that those are the only possible `a` values. Including an `y_else: U` value in all of these expressions is fine, but seems a bit inelegant, since they contain values that will never be used.

waissbluth:
I the usage of Object properties (e.g. Arrays, Hashes) being considered in this issue? Could it be?

For example, to style a layer based on a specific element (position) of an array property or using a key when the properties are nested.

docs/style-spec/expressions.md
jfirebaugh: @nickidlugash How about omitting `a_m` in that case instead (effectively making `y_m` into `y_else`)?

jfirebaugh:
@waissbluth Yes, `"get"` and `"at"` can be composed, e.g. `["at", 3, ["get", "temperature"]]`.

docs/style-spec/expressions.md
nickidlugash: @jfirebaugh I think that would make the expression more concise, but the styling less understandable (e.g. "motorway: orange, trunk: yellow" would become "motorway:orange, else: yellow"). But maybe I'm overthinking this – readability perhaps will mainly depend on how this type of expression would be displayed in the Studio UI.

jfirebaugh: For reasons of robustness, I place a pretty high value on ensuring that `match` is exhaustive (can never produce "no value"). If there was a way to statically verify that the the match labels were exhaustive with respect to potential inputs, then that would be a situation I'd want the "otherwise" value to be omitted. But for the majority of cases, that isn't possible -- the expression evaluator doesn't know that vector tile property contains only certain values. (And even if we are pretty certain it's true for Mapbox tilesets, I think there's still value in writing expressions that are robust to being copy and pasted for use with other datasets.)

src/style-spec/function/definitions/index.js
```diff
@@ -0,0 +1,188 @@
...
+    'geometry_type': [ StringType, [], () =>
+        'this.get(this.get(feature, "geometry", "feature"), "type", "feature.geometry")'
+    ],
+    'id': [ ValueType, [], () =>
+        'this.get(feature, "id", "feature")'
+    ],
+    'zoom': [ NumberType, [], () => 'mapProperties.zoom' ],
+    '+': defineBinaryMathOp('+', true),
+    '*': defineBinaryMathOp('*', true),
+    '-': defineBinaryMathOp('-'),
```
anandthakker: Ah good catch

src/style-spec/function/definitions/literal.js
```diff
@@ -0,0 +1,89 @@
...
+            expected.kind === 'Array' &&
+            (typeof expected.N !== 'number' || expected.N === 0)
+        ) {
+            type = expected;
+        }
+
+        return new Literal(context.key, type, value);
+    }
+
+    typecheck() {
```
anandthakker: Heh yep.

src/style-spec/function/definitions/literal.js
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+            return ["literal", this.value];
+        }
+    }
+
+    accept(visitor: Visitor<Expression>) { visitor.visit(this); }
+}
+
+function hasInvalidNumber (value: mixed, context: ParsingContext) {
+    if (typeof value === 'number' && Math.abs(value) > Number.MAX_SAFE_INTEGER) {
```
anandthakker: Whoops, good point.

anandthakker: @jfirebaugh hm, so should we only guard `match` label values, as they're the only ones we're explicitly treating as integers?

jfirebaugh: It would be nice to check that the input contains no integer literals greater than `MAX_SAFE_INTEGER` (±), but such a validation would need to be before the lossy `JSON.parse`.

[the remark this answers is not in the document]
anandthakker: Eh, I dunno if I'd expect to see an error for using, e.g., `1e20` (rather than `1.0e20`, which I suppose would be how one would mark intent not to treat the number as an integer).  I'm okay having large numbers be 'use at your own risk'... we can document that numbers will be treated as `Number` / `double`.

anandthakker:
@mourner @jfirebaugh posted some benchmarks in the top of the ticket - looks like DDS is about 2x slower with expressions, but hopefully quick profiling will yield some low-hanging fixes that will reclaim some of that.

mourner:
@anandthakker thanks! The frame-duration 6ms to 8ms change is also worth looking into.

ChrisLoer:
> The frame-duration 6ms to 8ms change is also worth looking into.

#5150 adds to frame-duration as well (because it's doing collision detection during rendering). It's mainly an issue for zooms/maps with very dense line labeling, which is kind of the worst case for collision detection. We should try to keep an eye on how the two PRs interact.

anandthakker:
@mourner @ChrisLoer After 6f4a6c4 the impact on `frame-duration` is approximately 0.84 ms:

screen shot 2017-08-30 at 11 53 57 am

src/style-spec/function/compile.js
```diff
@@ -0,0 +1,122 @@
...
+        const fn = new Function('$globalProperties', '$feature', `
+$globalProperties = $globalProperties || {};
+var $props = $feature && $feature.properties || {};
+var $this = this;
+return this.unwrap(${compiled})
+`);
+
+        return {
+            result: 'success',
+            function: fn.bind(evaluationContext()),
```
mourner: might be worth trying a custom `bind` or an arrow fn — calls of natively binded functions are expensive in many browsers

src/style-spec/function/evaluation_context.js
```diff
@@ -0,0 +1,245 @@
...
+        assert(expectedType.kind, `Invalid type ${JSON.stringify(expectedType)}`);
+
+        let type;
+        let typeError = false;
+        if (expectedType.kind === 'Null') {
+            typeError = value === null;
+        } else if (expectedType.kind === 'Value') {
+            typeError = true;
+        } else if (expectedType.kind !== 'Array') {
+            typeError = typeof value !== expectedType.kind.toLowerCase();
```
mourner: nitpick: `toLowerCase` will be called on almost every `get`; might be preferable to make a lookup table for kinds instead

src/style-spec/function/evaluation_context.js
```diff
@@ -0,0 +1,245 @@
...
+    coalesce: function (...thunks: Array<Function>) {
+        while (true) {
+            try {
+                if (thunks.length === 0) return null;
+                const result = (thunks.shift())();
+                if (result !== null) return result;
+            } catch (e) {
+                if (thunks.length === 0) throw e;
+            }
+        }
```
mourner: Could we do away with a simple `for` loop here? Here, we create an array from arguments, and then pop values one by one from it when we can do away with just iterating over them, which is much cheaper.

anandthakker: Ah, this is actually obsolete code after df245c1 - will remove.

gabrielmoncea:
I see that issue #4820 **Support contains, begins with, and ends with filter option for strings** is referenced in this package. But I cannot figure out the syntax for the expression to filter strings that begin with a given value. Can anyone help me with it?

Issue #4052, opened 2017-01-26
Make tile cache size customizable to prevent iOS crashes


kristfal:
Hey,

one of our core target devices are iPads with 1GB RAM and more. We're running a single MGL instance covering the full viewport. In that environment, we're experiencing a lot of crashes on iPads due to memory pressure, and some crashes on iPhones. 

The [sourceCache](https://github.com/mapbox/mapbox-gl-js/blob/67fc7dff69d512db92a30605ca533acfbb7100fb/js/source/source_cache.js#L279) assigns a cache size based on the viewport size of the device. 

Our current workaround is to reduce the cache size from the predefined 45 tiles to 8 on iPads. We do a similar but not as drastic reduction on iPhones. This workaround reduces memory related crashes quite a bit, but they are not fully sufficient to prevent a 1GB RAM iPad to handle zoom and pan interactions for more than a couple of minutes before crashing.

With that workaround, an emulated iPad in Chrome keeps a stable heap size of 60 - 70 MB in our test bench. 

In Mobile Safari, we see the test starting at 35 MB heap size, and then gradually growing up to 130+ MB (as the image bellow illustrates). Eventually either the inspector or safari crashes.

![skjermbilde 2017-01-26 17 48 04](https://cloud.githubusercontent.com/assets/5638107/22341265/746822b2-e3f0-11e6-809d-b0cead8d2c89.png)

**mapbox-gl-js version**: Master

### Steps to Trigger Behavior

 1. Run mapbox-gl-js in full screen on an iPad with 1 GB RAM
 2. Zoom and pan (for instance use flyTo) inside or between larger cities
 3. Crash

### Expected Behavior

iOS devices should not crash after prolonged use.

### Actual Behavior

Retained memory builds on iOS, triggering crashes after prolonged use.

### Suggestions

– Reduce cache size on iOS
– Identify potential memory leaks in Safari (let me know if any heap snapshots would help)

Side note: We are only able to record timelines in Safari when the map is not moving. The best approach seems to be to interact with the map for 30 seconds, pause and record with the inspector for 10 seconds and then repeat until either the inspector or safari crashes.

stackTom:
I have a similar issue: https://github.com/mapbox/mapbox-gl-js/issues/4063

kristfal:
@stackTom Try replacing `updateCacheSize` with:

```js
        const widthInTiles = Math.ceil(transform.width / transform.tileSize) + 1;
[...]

PR #4778, opened 2017-06-02
Add the ability to customize the tile cache size

jczaplew:
This pull request addresses #4052. Currently the tile cache size is determined by the size of the viewport and there is no way for the user to manually override this size. This pull requests exposes a new option on the map, `cacheSize`, which can be used to explicitly set the tile cache size.

### DIFF
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -60,7 +60,7 @@ class SourceCache extends Evented {
 
     onAdd(map) {
         this.map = map;
-        this._maxTileCacheSize = map._maxTileCacheSize
+        this._maxTileCacheSize = map._maxTileCacheSize;
         if (this._source && this._source.onAdd) {
             this._source.onAdd(map);
         }
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -60,7 +60,7 @@ class SourceCache extends Evented {
 
     onAdd(map) {
         this.map = map;
-        this._maxTileCacheSize = map._maxTileCacheSize;
+        this._maxTileCacheSize = map ? map._maxTileCacheSize : null;
         if (this._source && this._source.onAdd) {
             this._source.onAdd(map);
         }
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -53,14 +53,14 @@ class SourceCache extends Evented {
         this._cache = new Cache(0, this.unloadTile.bind(this));
         this._timers = {};
         this._cacheTimers = {};
-        this._tileCacheSize = null;
+        this._maxTileCacheSize = null;
 
         this._isIdRenderable = this._isIdRenderable.bind(this);
     }
 
     onAdd(map) {
         this.map = map;
-        this._tileCacheSize = map._tileCacheSize
+        this._maxTileCacheSize = map._maxTileCacheSize
         if (this._source && this._source.onAdd) {
             this._source.onAdd(map);
         }
@@ -281,7 +281,7 @@ class SourceCache extends Evented {
 
     /**
      * Resizes the tile cache based on the current viewport's size
-     * or the tileCacheSize option passed during map creation
+     * or the maxTileCacheSize option passed during map creation
      *
      * Larger viewports use more tiles and need larger caches. Larger viewports
      * are more likely to be found on devices with more memory and on pages where
@@ -295,7 +295,8 @@ class SourceCache extends Evented {
         const approxTilesInView = widthInTiles * heightInTiles;
         const commonZoomRange = 5;
 
-        const maxSize = this._tileCacheSize || Math.floor(approxTilesInView * commonZoomRange);
+        const viewDependentMaxSize = Math.floor(approxTilesInView * commonZoomRange);
+        const maxSize = typeof this._maxTileCacheSize === 'number' ? Math.min(this._maxTileCacheSize, viewDependentMaxSize) : viewDependentMaxSize;
 
         this._cache.setMaxSize(maxSize);
     }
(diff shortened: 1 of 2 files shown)


anandthakker:
Thanks for this PR< @jczaplew !

While do I agree that providing some control over the cache size seems like a good idea (especially in light of the iOS crashes), I find myself hesitant to solve this issue by exposing SourceCache's cache size as a direct option.

I'm going to spend a little time thinking about possible alternative designs and will post back here soon.

jczaplew:
Sounds good, and understandable!

src/ui/map.js
```diff
@@ -57,7 +57,9 @@ const defaultOptions = {
 
     renderWorldCopies: true,
 
-    refreshExpiredTiles: true
+    refreshExpiredTiles: true,
+
+    cacheSize: null
```
mourner: I would name it `tileCacheSize` for less ambiguity.

src/source/source_cache.js
```diff
@@ -291,7 +294,10 @@ class SourceCache extends Evented {
         const heightInTiles = Math.ceil(transform.height / transform.tileSize) + 1;
         const approxTilesInView = widthInTiles * heightInTiles;
         const commonZoomRange = 5;
-        this._cache.setMaxSize(Math.floor(approxTilesInView * commonZoomRange));
+
+        const maxSize = this._tileCacheSize || Math.floor(approxTilesInView * commonZoomRange);
```
anandthakker: Could we change this logic to:
```js
const viewDependentMaxSize = Math.floor(approxTilesInView * commonZoomRange);
const maxSize = typeof this._tileCacheSize === 'number' ? Math.min(this._tileCacheSize, viewDependentMaxSize) : viewDependentMaxSize;
```

This way, the new option would be setting an upper bound on the cache size, rather than forcing it to be fixed.

src/ui/map.js
```diff
@@ -122,6 +124,7 @@ const defaultOptions = {
  * @param {number} [options.bearing=0] The initial bearing (rotation) of the map, measured in degrees counter-clockwise from north. If `bearing` is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
  * @param {number} [options.pitch=0] The initial pitch (tilt) of the map, measured in degrees away from the plane of the screen (0-60). If `pitch` is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
  * @param {boolean} [options.renderWorldCopies=true]  If `true`, multiple copies of the world will be rendered, when zoomed out.
+  * @param {number} [options.tileCacheSize=null]  The number of tiles stored in the tile cache. If omitted, the cache will be dynamically sized based on the current viewport.
```
anandthakker: In light of the comment on the logic in `SourceCache`, would it make sense to rename this to `maxTileCacheSize`?  (Sorry, I know you already renamed once! 😬 )

jczaplew: Yes, I like that better in light of the above change

jczaplew:
@anandthakker - I renamed all instances of `tileCacheSize` to `maxTileCacheSize` per your recommendation, and also updated the logic in `src/source/source_cache.js` so that the parameter functions as a maximum value instead of simply a fixed value.

anandthakker:
Thanks @jczaplew!  Looks like there's a lint error preventing CI from passing.

anandthakker:
@jczaplew hm looks like some unit tests are failing.  Could you check that out?  You should be able to run them locally with `yarn test` or, slightly more conveniently `tap -b test/` to have it stop after the first failure.

jczaplew:
Unit tests check out. Please let me know if there are any other tweaks you would like.

src/ui/map.js
```diff
@@ -122,6 +124,7 @@ const defaultOptions = {
  * @param {number} [options.bearing=0] The initial bearing (rotation) of the map, measured in degrees counter-clockwise from north. If `bearing` is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
  * @param {number} [options.pitch=0] The initial pitch (tilt) of the map, measured in degrees away from the plane of the screen (0-60). If `pitch` is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
  * @param {boolean} [options.renderWorldCopies=true]  If `true`, multiple copies of the world will be rendered, when zoomed out.
+  * @param {number} [options.maxTileCacheSize=null]  The maxiumum number of tiles stored in the tile cache. If omitted, the cache will be dynamically sized based on the current viewport.
```
anandthakker: Ah, just noticed this -- we should probably specify here that it's the maximum number of tiles in the cache **_for a given source_**

jczaplew: Good catch. Fixed.

anandthakker:
Thanks @jczaplew 🎉 !

volkergraf:
Hmm I tried the following; (mapbox GL JS  0.39.1)

```
  map = new mapboxgl.Map({
            container: 'map',
            ......
            maxBounds: [
                [8.9651, 48.691],
                [9.0089, 48.705]
            ], // approximate bounds of tiles
            style: mapStyle,
            maxTileCacheSize: 2
           ....
        });
```

But when panning/tilting the whole (Cordova App) crashes and restarts .. any Ideas what I'm doing wrong ?

anandthakker:
@volkergraf I'm unable to reproduce this with the information above.  If you're able to create a minimal test case reproducing this issue, could you please open a new issue and share it there?

PR #4779, opened 2017-06-02
Apply text-keep-upright after text-offset

### LINKED ISSUE #2350: Fix behaviour when using keep-text-upright in combination with text-offset places text
**GL-JS version: Whatever was in use in Mapbox Studio on 2016-03-21**:

I have direction-specific line geometry in some tiles.  I want to offset labels to the right of the lines.

If I specify `text-offset: [[0,1]]`, it works, but a bunch of the labels are upside-down (as expected):

screen shot 2016-03-21 at 4 53 02 pm

If I enable "Keep text upright", then the text is upright, but the Y offset is interpreted incorrectly:

screen shot 2016-03-21 at 4 53 28 pm

The expected behaviour would be for the labels to rotate around their offset anchor until they're "upright".  It seems that the rotation to keep upright needs to be done _before_ the label is offset, rather than _after_.

anandthakker:
Closes #2350


ansis:
I think this change makes sense but are there any cases where the current behavior is the desired one? Is this a breaking change?

anandthakker:
@ChrisLoer looks like the collision problem already exists -- filed as a separate issue here: https://github.com/mapbox/mapbox-gl-js/issues/4798

anandthakker:
@jfirebaugh got confirmation from @nickidlugash that it's reasonable from the carto team's point of view to consider this a bug fix.

jfirebaugh:
Do the integration tests pass in -native?

anandthakker:
@jfirebaugh just chatted with @ansis about this when I realized I'd forgotten to mark them ignored -- since the shaders here are already ahead of native, he's going to make sure this fix is included in the stuff he's working on.

PR #4781, opened 2017-06-02
improve pitched legibility of labels that follow lines

### LINKED ISSUE #4718: transform symbol vertices outside of shaders
Broader context: https://github.com/mapbox/mapbox-gl-js/issues/4704

I'm working on transforming symbol vertices in javascript. Each frame the vertices are shifted, offset, projected and added to buffers as gl coordinates. Doing the transform outside of shaders let's us have full access to the line geometry which let's us render better in pitched views.

---

I'm working on this in the [cpu-label-transform](https://github.com/mapbox/mapbox-gl-js/tree/cpu-label-transform) branch.

I've hacked in a rough implementation that is mostly working, but it's currently taking ~20ms to run. I'm working on a bunch of relatively low-hanging optimizations that will hopefully cut that down significantly.

@ChrisLoer @kkaefer

ansis:
fixes #4718

This improves legibility of labels that follow lines in pitched views. The previous approach used the limited information in the shader to calculate put the glyph in approximatelyright place. The new approach does this more accurately by doing it on the cpu where we have access to the entire line geometry.

The core part of the pr is `src/symbol/projection.js` which does two things:
- calculates the matrices used for projecting symbols
- the find-the-correct-point-along-the-line-for-this-glyph calculations

See the comment in `src/symbol/projection.js` for an overview of the new approach.

This pr also gets rid of the the [sliding-glyphs-for-each-segment-that-get-toggled-on-and-off](https://camo.githubusercontent.com/5e1f9d174172207ae5a840dfae2ffec8b420661a/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f313432313635322f313535343730382f62373964623831342d346534622d313165332d393864612d3438646233336438353463652e676966) approach, which simplifies things a lot.

This fixes:
- garbled text near sharp corners
- upside down labels caused by perspective sloping
- glyph stretching and squishing
- letter spacing with pitch-scaling

### before / after

screen shot 2017-06-02 at 6 25 10 pm
screen shot 2017-06-02 at 6 23 44 pm
screen shot 2017-06-02 at 6 36 43 pm
screen shot 2017-06-02 at 6 37 00 pm
screen shot 2017-06-02 at 6 44 23 pm
screen shot 2017-06-02 at 6 43 25 pm

## Benchmarks
Frame duration is about 1ms slower in benchmarks. In full page maps it's about 1-2 ms slower, and a bit more in pitched full page maps.

benchmark | master 943dc39 | along-line-label-legibility b5b6f9a
--- | --- | ---
**map-load** | 243 ms  | 138 ms 
**style-load** | 122 ms  | 119 ms 
**buffer** | 1,075 ms  | 1,069 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | **4.2 ms**, 0% > 16ms  | **5.2 ms, 0% > 16ms**
**query-point** | 1.55 ms  | 1.41 ms 
**query-box** | 73.66 ms  | 73.58 ms 
**geojson-setdata-small** | 12 ms  | 5 ms 
**geojson-setdata-large** | 137 ms  | 130 ms 

## Launch Checklist

 - [x] post benchmark scores

@ChrisLoer @kkaefer @anandthakker


ansis:
and a couple wip shots:
screen shot 2017-05-12 at 11 17 33 am
screen shot 2017-05-12 at 11 45 20 am
screen shot 2017-05-12 at 12 50 14 pm

ChrisLoer:
This looks SOO good!

I won't have a chance to get into too much review until Monday, but to give you an idea for how the old `max-camera-distance` filter might still be relevant, look at:

`14.27/34.004/-118.4845/-46.6/60` with viewport aligned road labels:

screenshot 2017-06-02 16 47 34

src/data/bucket/symbol_bucket.js
```diff
@@ -655,55 +710,62 @@ class SymbolBucket {
...
         const elementArray = arrays.elementArray;
         const layoutVertexArray = arrays.layoutVertexArray;
+        const dynamicLayoutVertexArray = arrays.dynamicLayoutVertexArray;
 
         const zoom = this.zoom;
         const placementZoom = Math.max(Math.log(scale) / Math.LN2 + zoom, 0);
 
+        const glyphOffsetArrayStart = this.glyphOffsetArray.length;
+
+        const labelAngle = Math.abs((labelAnchor.angle + placementAngle) % Math.PI);
```
ChrisLoer: Maybe add a comment that `labelAngle` is over a 180 degree range relative to line orientation? I assumed 360 degrees at first and got confused reading the code.

src/render/painter.js
```diff
@@ -293,19 +293,27 @@ class Painter {
...
     }
 
-    translatePosMatrix(matrix, tile, translate, anchor) {
+    translatePosMatrix(matrix, tile, translate, anchor, inPixelUnits) {
         if (!translate[0] && !translate[1]) return matrix;
 
-        if (anchor === 'viewport') {
-            const sinA = Math.sin(-this.transform.angle);
-            const cosA = Math.cos(-this.transform.angle);
+        const angle = inPixelUnits ?
```
ChrisLoer: It's not immediately obvious here what `inPixelUnits` has to do with the anchor angle -- I had to go back and look at the use of `getGlCoordMatrix` in `draw_symbol` to puzzle it out. Maybe just an explanatory comment would help?

src/symbol/projection.js
```diff
@@ -0,0 +1,247 @@
...
+    const lineVertexArray = bucket.lineVertexArray;
+    const placedSymbols = isText ? bucket.placedGlyphArray : bucket.placedIconArray;
+
+    for (let s = 0; s < placedSymbols.length; s++) {
+        const symbol = placedSymbols.get(s);
+
+        const anchorPos = [symbol.anchorX, symbol.anchorY, 0, 1];
+        vec4.transformMat4(anchorPos, anchorPos, posMatrix);
+
+        // Don't bother calculating the correct point for invisible labels. Move them offscreen.
```
ChrisLoer: Maybe add a note here that we have to move them offscreen instead of just dropping them because we're not updating the paired/non-dynamic buffers?

src/shaders/symbol_icon.vertex.glsl
```diff
@@ -35,60 +33,45 @@ void main() {
...
-        layoutSize = u_size;
     }
 
-    float fontScale = u_is_text ? size / 24.0 : size;
+    vec4 projectedPoint = u_matrix * vec4(a_pos, 0, 1);
+    highp float camera_to_anchor_distance = projectedPoint.w;
+    highp float distance_ratio = u_pitch_with_map ?
+        camera_to_anchor_distance / u_camera_to_center_distance :
+        u_camera_to_center_distance / camera_to_anchor_distance;
+    highp float perspective_ratio = 0.5 + 0.5 * clamp(distance_ratio, 0.1, 10.0);
```
ChrisLoer: Just out of curiosity, what case is the `clamp` addressing? Or I just a way of saying "we expect a distance ratio of 10 to be effectively infinite"?

src/shaders/symbol_icon.vertex.glsl
```diff
@@ -35,60 +33,45 @@ void main() {
...
-        layoutSize = u_layout_size;
     } else {
         size = u_size;
-        layoutSize = u_size;
     }
 
-    float fontScale = u_is_text ? size / 24.0 : size;
+    vec4 projectedPoint = u_matrix * vec4(a_pos, 0, 1);
+    highp float camera_to_anchor_distance = projectedPoint.w;
+    highp float distance_ratio = u_pitch_with_map ?
```
ChrisLoer: The corresponding logic in `projection.js` is easier to understand, but here the inversion of `distance_ratio` is pretty hard to figure out since it's further removed from the rest of the projection logic. I would add a comment saying something like "If the label is pitched with the map, layout is done in pitched space, which makes labels in the distance smaller relative to viewport space. We counteract part of that effect by multiplying by the perspective ratio. If the label isn't pitched with the map, we do layout in viewport space, which makes labels in the distance larger relative to the features around them. We counteract part of that effect by dividing by the perspective ratio."

src/shaders/symbol_sdf.vertex.glsl
```diff
@@ -73,89 +52,38 @@ void main() {
...
-        gl_Position = u_matrix * vec4(a_pos, 0, 1) + vec4(extrude, 0, 0);
-        gl_Position.z += clipUnusedGlyphAngles(v_size * perspective_ratio, layoutSize, a_minzoom, a_maxzoom) * gl_Position.w;
-    // pitch-alignment: viewport
-    // rotation-alignment: viewport
-    } else {
-        vec2 extrude = fontScale * u_extrude_scale * perspective_ratio * (a_offset / 64.0);
-        gl_Position = u_matrix * vec4(a_pos, 0, 1) + vec4(extrude, 0, 0);
-    }
+    highp float distance_ratio = u_pitch_with_map ?
+        camera_to_anchor_distance / u_camera_to_center_distance :
```
ChrisLoer: Same comment as for `symbol_icon`

ChrisLoer:
Example of broken collision detection for vertically-oriented viewport-pitch-aligned road labels:

image

src/shaders/symbol_icon.vertex.glsl
ansis: I was seeing things like [this](https://cloud.githubusercontent.com/assets/1421652/26834130/78565fec-4aa2-11e7-9904-10a3d3d388db.png) and thought it was an infinite/precision issue. Since I'm not seeing it now I think it might have been something else (broken inversion). Removing

ansis:
@ChrisLoer yeah, the collision boxes are a problem. I just tried adding an adjustment to `maxScale` to account for this and it seems like it mostly works: https://github.com/mapbox/mapbox-gl-js/pull/4781/commits/6a789ecff129cae92d107cbe85d14a537340d206. Do you think this is good enough to last until we rewrite this?

src/symbol/collision_tile.js
```diff
@@ -150,6 +150,16 @@ class CollisionTile {
             box.bbox2 = x2;
             box.bbox3 = y2;
 
+            // When the map is pitched the distance covered by a line changes.
+            // Adjust the max scale by (approximatePitchedLength / approximateRegularLength)
+            // to compensate for this.
+            const yStretch2 = yStretch;
+            const xSqr = box.offsetX * box.offsetX;
```
ChrisLoer: Don't these offsets need to have a rotation applied to them? I think as is this only works for labels that are relatively horizontal relative to tile orientation.

ChrisLoer:
This is a nit, and it could go in another PR, but maybe it would be better if we detected the orientation of a line label based on the beginning and the end of the label instead of the orientation of the middle line segment? The current approach can lead to labels that look "upside down" even though technically the segment their anchor is on is "right side up":

image

src/symbol/projection.js
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    // For each glyph, find the point `offsetX` distance from the anchor.
+    for (const glyph of glyphs) {
+        const offsetX = Math.abs(glyph.offsetX) * fontScale;
+
+        // If the current segment doesn't have enough remaining space, iterate forward along the line.
+        // Since all the glyphs are sorted by their distance from the anchor you never need to iterate backwards.
+        // This way line vertices are projected at most once.
+        while (offsetX >= segmentDistance + previousDistance && Math.abs(vertexIndex) < numVertices) {
```
ChrisLoer: This while loop exits if we reach the end of the line (`vertexIndex >= numVertices`), but we keep adding glyphs onto whatever the last segment was. This leads to some funky behavior:

- Ignoring collisions (because collision boxes only go as long as the line)
- Road labels extending past the end of the road
- Road labels wrapping back on themselves (see Ocean Park Blvd in the screen capture)... although this one might be better addressed by a `max-curvature` property.

turnabout is fair play

We should hide the label instead if we hit this limit.

src/symbol/collision_tile.js
ansis: Yeah, definitely

ChrisLoer:
With an eye on https://github.com/mapbox/mapbox-gl-native/pull/9009#issuecomment-306276107, we should try to see if we can stay under the 8 vertex attribute limit when we make these changes on the JS side.

> since we should move towards treating `a_size` as a regular data-driven property we might not want to pack into it.

Do we have a choice? On this PR `a_data` is using 40 bits and `a_size` is using 32 bits, and I believe the only extra space is 16 bits in `a_projected_pos`. If we keep `a_size` and try to get rid of `a_data`, we'd have 40 - 16 -> 24 bits to pack in somewhere else. We can drop a few bits of precision from the offset in `a_pos_offset` (dropping from 1/64 precision to 1/8 saves 6 bits and doesn't look too noticeably different although the render tests do change), and a few bits from the anchor position. You could also go back to dropping 8 bits of precision from `segment_angle` (but that'd be a bummer, because the low precision angles meant that labels sometimes didn't quite line up with their lines).

ansis:
I just rebased on master which includes the recent `*-offset` fix (https://github.com/mapbox/mapbox-gl-js/pull/4779). The new implementation is [here](https://github.com/mapbox/mapbox-gl-js/blob/75f0765dfded4afbb893952449fc9d76923d2f35/src/symbol/projection.js#L251) for alongLine labels and [here](https://github.com/mapbox/mapbox-gl-js/blob/75f0765dfded4afbb893952449fc9d76923d2f35/src/symbol/quads.js#L157) for horizontal labels.

> Out of curiosity, why did you pack the label angle into the third part of a_projected_pos instead of just adding a fourth item to the vector? Do we think there's a performance benefit? I was a little sad to see the increased label angle precision go away (and a little surprised that text-writing-mode/chinese didn't have to be reverted after the angle change)

I'm not sure, I guess the impulse was to just pack everything as tightly as possible instead of using an extra 4 bytes. Should I switch to using a fourth item?

my remaining items:
- fix the `x` component of `text-offset` for labels that follow lines
- fix failing query test

ansis:
@ChrisLoer I also made a query test less sensitive to avoid a failure. The test *shouldn't* be less sensitive but I don't think we can fix this properly before working on all the collision stuff.

PR #4790, opened 2017-06-06
Normal map (normal) layer type

ibesora:
Adds a new layer type called _normal_ that given a source layer with raster type pointing to a normal map pyramid, is shaded with the light defined on the style object.

I've created a new draw_normal.js where the webgl initialization is done and two files: normal.fragment.js and normal.vertex.js with the normal mapping shader. The remaining files are modified to update the spec to support the new type and its parameters (the same as a raster layer)

### DIFF
--- a/src/render/draw_normal.js
+++ b/src/render/draw_normal.js
@@ -1,0 +1,171 @@
+'use strict';
+
+const util = require('../util/util');
+const glMatrix = require('@mapbox/gl-matrix');
+
+module.exports = drawNormalMap;
+
+function setLight(program, painter) {
+    const gl = painter.gl;
+    const light = painter.style.light;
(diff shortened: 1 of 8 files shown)


mourner:
Is this similar to #4701 @kkaefer @mollymerp?

anandthakker:
@ibesora thank you very much for submitting this PR!

As @mourner notes above, we are working on similar (but not identical) functionality over in #4701, with the `hillshade` layer type.

With that new type already being added to the spec, I think that the `normal` layer you've created here would be fit better as a [custom layer](https://github.com/mapbox/mapbox-gl-js/issues/281).  The custom layer API is still in early design/development, but I think this is a great use case for us to take into account while working on it!

PR #4809, opened 2017-06-08
Added style field

tomscholz:
Many modules in npm are starting to expose their css entry files in their package.json files. This allows tools like npm-css or rework-npm to import `mapbox-gl.css` from the node_modules directory and not from the cdn


mourner:
Thanks!

PR #4811, opened 2017-06-09
Introduce Texture class to dry up code

mourner:
Abstracts our texture logic (which was repeated in many places) in a `Texture` class, reducing the amount of code and the error surface. A work in progress — just wanted to get this out to show the direction this is going and get some feedback.

## Launch Checklist

 - [ ] fix a few render test failures
 - [ ] post benchmark scores


anandthakker:
@mourner nice!  Approach looks good to me overall.  One thing I wonder: would it be worthwhile to try to avoid having `Texture` keep the reference to the GL context, by having it passed in to `setData`, `setImage`, etc?

jfirebaugh:
Added something similar in #5190; tracking future work in #5277.

PR #4819, opened 2017-06-10
Only necessary files are published to npm

tomscholz:
Switched from .npmignore to files array inside package.json
- https://github.com/npm/npm/wiki/Files-and-Ignores
- https://docs.npmjs.com/files/package.json


mourner:
Thanks for the PR! Is there any advantage to specifying included files in `package.json` as opposed to excluded files in `.npmignore`? It seems to me that it would be much easier to accidentally screw up a release with the first approach by forgetting to update `package.json` after a change in file paths.

DannyDelott:
The general reason is that dotfiles are easy to miss when downloading/sharing source code or copying directories around.

There is also an argument to suggest that dotfiles are better for machine-specific config (eg: `.git`, `.vimrc`) and essential project config (eg: `package.json`, `yarn.lock`) belongs in files with more visibility.

package.json
```diff
@@ -128,5 +128,12 @@
...
     "prepublish": "in-publish && run-s build-dev build-min || not-in-publish",
     "postinstall": "cd src/style-spec && yarn || npm install"
-  }
+  },
+  "files": [
+    "src/",
+    "dist/",
+    "plugins/",
+    "LICENSE.txt",
+    "ARCHITECTURE.md"
```
mollymerp: I don't think we'd want to include `ARCHITECTURE.md` in the npm package. We should include `README.md` and `CHANGELOG.md` though. 

Also the `plugins/` folder isn't needed (it actually shouldn't even be in `master` anymore)

tomscholz: Readme and changelog are automatically added see https://docs.npmjs.com/files/package.json#files

tomscholz: I also remove the `LICENSE.txt`.
In the npm documentation it states 
> Certain files are always included, regardless of settings:
[...]
LICENSE / LICENCE
README, CHANGES, LICENSE & NOTICE can have any case and extension.

mollymerp:
You've still got `/plugins` in there 😸

tomscholz:
Done! 💪🏻

mollymerp:
@anandthakker do you know if we need to include the `flow_interfaces` directory in the files published to the npm registry? From what I can tell flow is used primarily for testing and we `unflowify` before we build the `dist` files, but I'm not sure if we need the flow files for people using mapbox-gl with another bundler 💭

anandthakker:
> do you know if we need to include the flow_interfaces directory in the files published to the npm registry? From what I can tell flow is used primarily for testing and we unflowify before we build the dist files, but I'm not sure if we need the flow files for people using mapbox-gl with another bundler

Yeah, IMO we should keep these in so that people who want to bundle our source directly have a fighting chance.

LitoMore:
We should not publish the `src` to npm.

anandthakker:
@LitoMore we intentionally publish `src` to NPM so that developers who prefer to use flow-annotated sources are able to do so.

Issue #4434, opened 2017-03-15
Master ticket for remaining data-driven style properties


jfirebaugh:
Property | Status
------------ | -------------
`visibility` | #4288. Needs discussion; overlaps with filters; we may elect to satisfy the use case via other features.
`fill-antialias` | Not ticketed. Lowest priority; we are considering removing support for this property (#4093).
`fill-translate` | #2731. Low priority.
`fill-translate-anchor` | #2731. Low priority.
`fill-pattern` | #2732. Medium priority.
`line-cap` | Not ticketed. Low priority; probably not difficult.
`line-join` | Not ticketed. Low priority; probably not difficult.
`line-miter-limit` | Not ticketed. Lowest priority; we are considering removing support for this property (#4092).
`line-round-limit` | Not ticketed. Lowest priority; we are considering removing support for this property (#4092).
`line-translate` | #2731. Low priority.
`line-translate-anchor` | #2731. Low priority.
`line-width` | #3045. High priority; high difficulty. May require backwards incompatible changes to the style specification to support (#4158).
`line-dasharray` | #3045. Low priority; high difficulty. Perhaps not feasible to support.
`line-pattern` | #2732. Medium priority.
`symbol-placement` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`symbol-spacing` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`symbol-avoid-edges` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`icon-allow-overlap` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`icon-ignore-placement` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`icon-optional` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`icon-rotation-alignment` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`icon-size` | #4228. High priority; currently in development.
`icon-text-fit` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`icon-text-fit-padding` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`icon-image` | #4304. Medium priority.
`icon-padding` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`icon-keep-upright` | Not ticketed. Needs compelling use cases / investigation of feasibility.
`text-pitch-alignment` | Not ticketed. Needs compelling use cases / investigation of feasibility.
[...]

PR #4836, opened 2017-06-15
draft of expressions examples doc

nickidlugash:
@anandthakker as discussed, here's a doc with some practical examples of expressions. If any of the syntax is incorrect please update 🙏   

I wasn't sure where to put the file, so please move it if you wish.

### DIFF
--- a/docs/style-spec/expressions-examples.md
+++ b/docs/style-spec/expressions-examples.md
@@ -75,7 +75,13 @@ Example output value:
 {
 	"icon-image": {
 		"expression" : [
-				"concat",["string", ["get", ["properties"], "shield"]], "-", ["string", ["get", ["properties"], "ref"]]
+				"concat",
+				["string", ["get", ["properties"], "shield"]],
+				"-",
+				["case",
+					["==", ["length", ["string", ["get", ["properties"], "ref"]]], 1],
+					["2"]
+					["string", ["length", ["string", ["get", ["properties"], "ref"]]]]
 		]
 	}
 }
@@ -112,13 +118,15 @@ Mountain peak label displays name and elevation if elevation value exists, other
 				["match", ["get", ["properties"], "class"], 
 					"park", "hsl(100, 58%, 76%)",
 					"school", "hsl(50, 47%, 81%)",
-					"hospital", "hsl(340, 37%, 87%)"
+					"hospital", "hsl(340, 37%, 87%)",
+					"hsl(35, 12%, 89%)"
 				],
 				16,
 				["match", ["get", ["properties"], "class"], 
 					"park", "hsl(100, 58%, 76%)",
 					"school", "hsl(50, 63%, 84%)",
-					"hospital", "hsl(340, 63%, 89%)"
+					"hospital", "hsl(340, 63%, 89%)",
+					"hsl(35, 12%, 89%)"
 				],
 		]
 	}
@@ -155,20 +163,20 @@ Based on the following data properties:
 {
 	"icon-image": {
 		"expression": [
-			"case",
-			["<=", ["number", ["get", ["properties"], "scalerank"]], 2],
+			"curve", ["step"], ["number", ["get", ["properties"], "scalerank"]],
+			0,
 			["case", 
 				["==", ["number", ["get", ["properties"], "capital"]], 2],
 				"capital-city-dot-large",
 				"city-dot-large" 
 			],
-			["<=", ["number", ["get", ["properties"], "scalerank"]], 5],
+			3,
 			["case", 
 				["==", ["number", ["get", ["properties"], "capital"]], 2],
 				"capital-city-dot-medium",
 				"city-dot-medium" 
 			],	
-			[">", ["number", ["get", ["properties"], "scalerank"]], 5],
+			6,
 			["case", 
 				["==", ["number", ["get", ["properties"], "capital"]], 2],
 				"capital-city-dot-small",
@@ -182,25 +190,15 @@ Based on the following data properties:
 {
 	"text-size": {
 		"expression": [
-			"curve", "exponential", 0.85, ["zoom"],
+			"curve", ["exponential", 0.85], ["zoom"],
 				4,
-				["case",
-					["<=", ["number", ["get", ["properties"], "scalerank"]], 2],
-					12,
-					["<=", ["number", ["get", ["properties"], "scalerank"]], 5],
-					9,
-					[">", ["number", ["get", ["properties"], "scalerank"]], 5],
-					6
+				["curve", ["step"], ["number", ["get", ["properties"], "scalerank"]],
+					0, 12,
+					3, 9,
+					6, 6
 				],
 				14,
-				["case",
-					["<=", ["number", ["get", ["properties"], "scalerank"]], 2],
-					22,
-					["<=", ["number", ["get", ["properties"], "scalerank"]], 5],
-					22,
-					[">", ["number", ["get", ["properties"], "scalerank"]], 5],
-					22
-				],
+				22
 		]
 	}
 }
@@ -211,7 +209,7 @@ Based on the following data properties:
 		"expression": [
 			"curve", "step", ["zoom"],
 				7, 
-				["match", ["get", ["properties"], "ldir"],
+				["match", ["string", ["get", ["properties"], "ldir"]],
 					"N", "center",
 					"NE", "left",
 					"E", "left",
@@ -219,7 +217,8 @@ Based on the following data properties:
 					"S", "center",
 					"SW", "right",
 					"W", "right",
-					"NW", "right"
+					"NW", "right",
+					"center"
 				],
 				8,
 				"center"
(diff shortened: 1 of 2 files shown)


docs/style-spec/expressions-examples.md
```diff
@@ -0,0 +1,298 @@
...
+
+Example output value: 
+`us-interstate-3`
+
+```
+{
+	"icon-image": {
+		"expression" : [
+				"concat",["string", ["get", ["properties"], "shield"]], "-", ["string", ["get", ["properties"], "ref"]]
+		]
```
anandthakker: Is the after the `"-"` supposed to be wrapped in `length`, like  `["length", ["string", ["get", ["properties"], "ref"]]`?

docs/style-spec/expressions-examples.md
```diff
@@ -0,0 +1,298 @@
...
+				"city-dot-medium" 
+			],	
+			[">", ["number", ["get", ["properties"], "scalerank"]], 5],
+			["case", 
+				["==", ["number", ["get", ["properties"], "capital"]], 2],
+				"capital-city-dot-small",
+				"city-dot-small" 
+			]
+		]	
+	}
```
anandthakker: You could use a `"step"` curve here as well:

```js
[
  "curve", ["step"],
  0, ["case", ["==", ["number", ["get", ["properties"], "capital"]], 2], "capital-city-dot-large", "city-dot-large" ],
  2, ["case", ["==", ["number", ["get", ["properties"], "capital"]], 2], "capital-city-dot-medium", "city-dot-medium" ],
  5, ["case", ["==", ["number", ["get", ["properties"], "capital"]], 2], "capital-city-dot-small", "city-dot-small" ]
]
```

docs/style-spec/expressions-examples.md
```diff
@@ -0,0 +1,298 @@
...
+		"expression": [
+			"curve", "exponential", 0.85, ["zoom"],
+				4,
+				["case",
+					["<=", ["number", ["get", ["properties"], "scalerank"]], 2],
+					12,
+					["<=", ["number", ["get", ["properties"], "scalerank"]], 5],
+					9,
+					[">", ["number", ["get", ["properties"], "scalerank"]], 5],
+					6
```
anandthakker: Similarly, this `case` could be a `step` curve instead.

docs/style-spec/expressions-examples.md
anandthakker: Ah, except for one thing: as of now, `step` curves (like their predecessor, `interval` functions) use intervals that are inclusive of the left-hand boundary and exclusive of the right-hand one.  We could make this configurable (`["step", ">"]` / `["step", ">="]`), or we could just rely on a workaround like using `2.001` instead of `2` as the step value.

docs/style-spec/expressions-examples.md
```diff
@@ -0,0 +1,298 @@
...
+	}
+}
+```
+```
+{
+	"text-justify": {
+		"expression": [
+			"curve", "step", ["zoom"],
+				7, 
+				["match", ["get", ["properties"], "ldir"],
```
anandthakker: the `["get", ...` here should be wrapped in a `["string", ...`

docs/style-spec/expressions-examples.md
```diff
@@ -0,0 +1,298 @@
...
+					"N", "center",
+					"NE", "left",
+					"E", "left",
+					"SE", "left",
+					"S", "center",
+					"SW", "right",
+					"W", "right",
+					"NW", "right"
```
anandthakker: Needs a final output value in case `ldir` doesn't match one of these values.

docs/style-spec/expressions-examples.md
nickidlugash: oops, no – that property should be `reflen`, which is the same thing as ^, pre-calculated in the data, and with the exception that values of `1` become `2` (because we use the same icon for both). I don't know yet if we would consider dropping this property in favor of using `length` (since we have users of our tiles that don't use Mapbox GL), but I'll try rewriting this expression to use `length` anyway, as an example.

docs/style-spec/expressions-examples.md
nickidlugash: @anandthakker ah, just realizing now that `y_else` is specified for [`match`](https://github.com/mapbox/mapbox-gl-js/pull/4777/files#diff-31711ce21b059455779238e7ddf5de21R136). Should I add a final output value to all the `match` expressions in this file? Is it an invalid expression if the final output value is not included?

anandthakker: Yah, it'll fail validation if not included

docs/style-spec/expressions-examples.md
nickidlugash: for this example (since the values are integers), I think the steps: 0, 3, 6 would be the same as the current `case` expression?

jfirebaugh:
Thanks for these examples @nickidlugash -- very instructive to look at some real use cases. The first thing that jumps out at me is that while our current syntax enables a lot of flexibility, it leave a lot to be desired in terms of ergonomics and concision. It's never too early to begin thinking about how we could improve this, so here are the things that come to my mind:

* Reverse the order of arguments to `"has"` and `"get"`, and make `["properties"]` the default for what becomes the final argument. So `["has", ["properties"], "foo"]` ⇢ `["has", "foo"]`. This simple change would probably halve the length of many of these examples.
* As previously discussed, it's probably worthwhile to have a `"coalesce"` expression (returns the first non-error value in a sequence of subexpressions) as sugar for chained `"has"` inside `"case"`.
* We're almost certainly going to want to add local identifier bindings -- `"let"` -- at some point, so that common subexpressions don't have to be repeated within more complex expressions. This in turn may require us to relax the restriction on where `["zoom"]` can appear. For example, in the city labels examples, you'd want to be able to write:
  ```
  ["let", "scalerank", ["number", ["get", ["properties"], "scalerank"]],
     ["curve", "exponential", 0.85, ["zoom"], ... expressions using "scalerank" ...]]
  ```

docs/style-spec/expressions-examples.md
```diff
@@ -0,0 +1,298 @@
...
+				],
+				14,
+				["case",
+					["<=", ["number", ["get", ["properties"], "scalerank"]], 2],
+					22,
+					["<=", ["number", ["get", ["properties"], "scalerank"]], 5],
+					22,
+					[">", ["number", ["get", ["properties"], "scalerank"]], 5],
+					22
+				],
```
nickidlugash: @anandthakker Can I drop this second case expression and just make this value: `14, 22` ? Will the renderers be able to calculate an exponential curve like this?

anandthakker: Ah hah -- yep! that should work!

nickidlugash:
> Reverse the order of arguments to "has" and "get", and make ["properties"] the default for what becomes the final argument.

@anandthakker @jfirebaugh what are the other available options besides `["properties"]` for these expressions?

> it's probably worthwhile to have a "coalesce" expression (returns the first non-error value in a sequence of subexpressions) as sugar for chained "has" inside "case".

I anticipate that this will be a fairly commonly used syntax pattern for our core styles.

I'm happy to keep these examples updated as we make any improvements to the syntax. (@anandthakker up to you if you want to merge this in once all the expressions are valid/optimal according to the latest syntax definitions, or if you want to keep this separate for now). This exercise has been super helpful for me for thinking through the future of our core styles as well 🌅

anandthakker:
> what are the other available options besides ["properties"] for these expressions?

The only immediate example I know of would be if a feature property value were to be a nested object (technically possible in GeoJSON, and maybe in the future in vector tiles as well).  E.g. `{"properties": { "my_data": { "x": 1, "y": 2 } }`, and something like `["get", ["object", ["get", ["properties"], "my_data"], "x"]`.

> up to you if you want to merge this in once all the expressions are valid/optimal according to the latest syntax definitions, or if you want to keep this separate for now

I wonder if it would be worthwhile to make these into integration test cases like the ones added in https://github.com/mapbox/mapbox-gl-js/pull/4841/commits/63e4c642724256a339c1a0fec0be72ac7586f55a.  That would let us keep checking the syntax, and also to include expected input/output values to confirm that the functions really do behave the way we think.

nickidlugash:
> I wonder if it would be worthwhile to make these into integration test cases

Could be a good idea! Based on [the master DDS-tracking ticket](https://github.com/mapbox/mapbox-gl-js/issues/4434), some of the style properties used in this example page (`text-anchor`, `text-justify`) have not been ported to DDS yet though.

nickidlugash:
>  some of the style properties used in this example page (text-anchor, text-justify) have not been ported to DDS yet though.

Noting that `text-anchor` and `text-justify` have now been ported to DDS. All of these examples should now be usable.

jfirebaugh:
Closing; this was very useful as a design aid but I don't think we're planning to keep expressions.md around much longer.

Issue #4786, opened 2017-06-06
Full screen control should not be displayed on unsupported device


ethan-ho:
**mapbox-gl-js version**: 0.37.0

### Steps to Trigger Behavior

 1. Visit [Fullscreen](https://www.mapbox.com/mapbox-gl-js/example/fullscreen/) example on iOS device, I am using iPad  
 2. Click the fullsceen button on the map

### Expected Behavior
If fullscreen JavaScript API is not supported on current device, the button should not be added to the map instance, like the GeolocateControl [currently does](https://github.com/mapbox/mapbox-gl-js/blob/c3de2d2a681ce69e61089db3a22e5225175e217a/src/ui/control/geolocate_control.js#L70).

### Actual Behavior
The button is appearing but do nothing.

mollymerp:
closed by #4838

jerrygarciame123:
Can't we just use JS to check and add/remove accordingly?  Thanks, John

PR #4838, opened 2017-06-15
do not display fullscreen control on unsupported devices #4786

stepankuzmin:
Do not display fullscreen control on unsupported devices according to #4786

I think, that it would be useful to warn the user if the device does not support fullscreen mode.

### DIFF
--- a/src/ui/control/fullscreen_control.js
+++ b/src/ui/control/fullscreen_control.js
@@ -4,6 +4,8 @@ const DOM = require('../../util/dom');
 const util = require('../../util/util');
 const window = require('../../util/window');
 
+const className = 'mapboxgl-ctrl';
+
 /**
  * A `FullscreenControl` control contains a button for toggling the map in and out of fullscreen mode.
  *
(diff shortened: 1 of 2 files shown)


src/ui/control/fullscreen_control.js
```diff
@@ -4,6 +4,8 @@ const DOM = require('../../util/dom');
 const util = require('../../util/util');
 const window = require('../../util/window');
 
+const className = 'mapboxgl-ctrl';
```
mollymerp: instead of another global variable could you store this as `this._className`?

src/ui/control/fullscreen_control.js
```diff
@@ -33,15 +35,15 @@ class FullscreenControl {
...
-        this._mapContainer = map.getContainer();
-        window.document.addEventListener(this._fullscreenchange, this._changeIcon);
-        return container;
+        this._map = map;
+        this._mapContainer = this._map.getContainer();
+        this._container = DOM.create('div', `${className} mapboxgl-ctrl-group`);
+        if (this._checkFullscreenSupport()) {
+            this._setupUI();
+        } else {
+            util.warnOnce('This device does not support fullscreen mode.');
```
mollymerp: we should also set `display:none;` for the container when fullscreen is not supported

src/ui/control/fullscreen_control.js
stepankuzmin: Sure!

stepankuzmin:
Done!

mollymerp:
Thanks for the contribution @stepankuzmin! I think this is almost ready to go, but it would be great to see a unit test that confirms the control container gets added to the map when fullscreen is enabled.

stepankuzmin:
Thanks for the review, @mollymerp! I've updated the PR.

mollymerp:
Thanks again @stepankuzmin!

dieterdreist:
There's a problem now on Firefox as on Safari mobile: does not display the Fullscreen icon any more and logs "This device does not support fullscreen mode." to the console. Still working on Safari (desktop). Until api v. 0.38.0 it worked in Firefox as well.

aparajita:
@dieterdreist Fullscreen was never supported on mobile devices. Mapbox was mistakenly showing the control, but it couldn't have worked, because the fullscreen API isn't supported (except for video and audio elements).

http://caniuse.com/#feat=fullscreen

jfirebaugh:
Fixed Firefox in #5272.

PR #4841, opened 2017-06-15
JS arbex implementation: parse, typecheck, compile expressions

anandthakker:
Todo:
 - [x] Implement `match` expressions (blocked by syntax design https://github.com/mapbox/mapbox-gl-js/pull/4777#discussion_r122568073)
 - [x] Implement `curve` expressions
 - [x] Implement `color` expressions
 - [x] Add unit tests for each function type
 - [x] Revise `get`, `has` to be more concise per https://github.com/mapbox/mapbox-gl-js/pull/4836#issuecomment-309521472
 - [x] Add constructors for `Vector<T>` and `Array<T, N>`
 - [x] Implement runtime error handling
 - [x] Add `coalesce`
 - [x] Write stop function => expression function conversion, and use it to replace current function implementation
 - [x] Add constraint on use of `zoom` expression, and add stop/interpolation metadata to compiled result
 - [x] OO-ify
 - [x] Implement `let`
 - [x] <del>Type checking a `number->number` against `number->variant<number|string>` should yield an expression of type `number->number`.</del> (punt on this for now)
 - [x] Update style spec to accept `{expression: [...]}` as a function.
 - [ ] Refactor style function call sites to pass `(globalProperties, feature)` rather than `(globalProperties, featureProperties)` (needed for `geometry_type` and `id`)
 - [x] Include expression type checking in spec validation (e.g. in `setPaintProperty`, etc.) -- but see https://github.com/mapbox/mapbox-gl-js/pull/4841#issuecomment-310491857; we should look into what native does in this case.
 - [x] Remove `=>` (and any other es6 stuff) from generated code, since it won't be transpiled.
 - [x] Update `expressions.md` to accurately reflect the implementation here.
 - [x] <del>Implement `cubic-bezier` interpolation type for `curve` expressions</del> (doing this in a followup)


anandthakker:
@jfirebaugh @ansis plenty of work still left to do here, but a base implementation is in place in case you want to take an early look.

src/style-spec/function/expressions.js
```diff
@@ -0,0 +1,354 @@
...
+    'json_array': {
+        name: 'json_array',
+        type: lambda(vector(ValueType), ValueType),
+        compile: fromContext('asArray')
+    },
+    'object': {
+        name: 'object',
+        type: lambda(ObjectType, ValueType),
+        compile: fromContext('asObject')
+    },
```
anandthakker: `json_array` and `object` are assertions (yielding an error if the input isn't the right type), whereas `number`, `string`, `boolean` are coercions.  Should we _also_ have assertions for number/string/boolean?

src/style-spec/function/compile.js
```diff
@@ -0,0 +1,155 @@
...
+ * ```
+ * {
+ *   result: 'error',
+ *   errors: Array<CompileError>
+ * }
+ * ```
+ *
+ * @private
+ */
+function compileExpression(expr: any) {
```
jfirebaugh: Looks like it's possible to narrow `any` to `mixed` here without too much trouble.

src/style-spec/function/compile.js
```diff
@@ -0,0 +1,155 @@
...
+ import type { ExpressionName } from './expressions.js';
+
+ export type CompiledExpression = {|
+     result: 'success',
+     js: string,
+     type: Type,
+     isFeatureConstant: boolean,
+     isZoomConstant: boolean,
+     expression: TypedExpression,
+     function?: any
```
jfirebaugh: This `any` can be `Function`.

src/style-spec/function/expressions.js
```diff
@@ -0,0 +1,354 @@
...
+    'case': {
+        name: 'case',
+        type: lambda(typename('T'), nargs(BooleanType, typename('T')), typename('T')),
+        compile: (_, args) => {
+            args = [].concat(args);
+            const result = [];
+            while (args.length > 1) {
+                const c = args.splice(0, 2);
+                result.push(`${c[0].js} ? ${c[1].js}`);
```
anandthakker: 👍 yep -- it's done globally at the end of `compile()`, https://github.com/mapbox/mapbox-gl-js/pull/4841/commits/04dbb977691d2a1710e0587c6e8cb058138ebf12#diff-442324b3174c25e635f18536d7732df2R147, but this reminds me, we should include a test that targets this since it's be very easy to regress

src/style-spec/function/expressions.js
```diff
@@ -0,0 +1,354 @@
...
+    const args = [];
+    while (arity-- > 0) args.push(NumberType);
+    return {
+        name: name,
+        type: lambda(NumberType, ...args),
+        compile: (_, args) => ({ js: `Math.${key}(${args.map(a => a.js).join(', ')})` })
+    };
+}
+
+function defineBinaryMathOp(name, isAssociative) {
```
anandthakker: Ah, whoops!  This was intended for allowing  `+`, `&&`, etc. to take > 2 args as a convenience.

src/style-spec/function/evaluation_context.js
```diff
@@ -0,0 +1,170 @@
...
+
+module.exports = () => ({
+    get: function (obj, key) {
+        if (!obj) throw new RuntimeError(`Cannot get property ${key} from null object`);
+        if (!(key in obj)) throw new RuntimeError(`Property ${key} not found in object with keys: [${Object.keys(obj)}]`);
+        return obj[key];
+    },
+
+    typeOf: function (x) {
+        if (Array.isArray(x)) return 'Vector<Value>';
```
anandthakker: It's not quite the only place it's exposed publicly: type checking error messages will need to refer to types by name, and probably some runtime error messages will, too.  That said, I'm also with you in that I've been uneasy/unsatisfied with having `Vector<Value>` as a possible output of `typeof` (because of both the generics syntax and the `Value` alias).

Would it be confusing for `[typeof, [get, 'foo']]` to equal `Array`, but then to have `[number,
 [at, [get, 'foo'], 0]]` to produce a runtime error message like `Expected Vector<number> but properties.foo was an object instead`?

src/style-spec/function/expressions.js
```diff
@@ -0,0 +1,354 @@
...
+ export type ExpressionName = "literal" | "ln2" | "pi" | "e" | "string" | "number" | "boolean" | "json_array" | "object" | "get" | "has" | "at" | "typeof" | "length" | "zoom" | "properties" | "geometry_type" | "id" | "case" | "match" | "is_error" | "==" | "!=" | ">" | ">=" | "<=" | "<" | "&&" | "||" | "!" | "curve" | "step" | "exponential" | "linear" | "cubic-bezier" | "+" | "-" | "*" | "/" | "%" | "^" | "log10" | "ln" | "log2" | "sin" | "cos" | "tan" | "asin" | "acos" | "atan" | "ceil" | "floor" | "round" | "abs" | "min" | "max" | "concat" | "upcase" | "downcase" | "rgb" | "rgba" | "color" | "color_to_array"
+
+ type Definition = {
+     name: ExpressionName,
+     type: Type,
+     compile: (expr: TypedExpression, args: Array<CompiledExpression>) => ({ js?: string, errors?: Array<string>, isFeatureConstant?: boolean, isZoomConstant?: boolean })
+ }
+ */
+
+const expressions: { [string]: Definition } = {
```
anandthakker: Ack.  Seems like this isn't compatible with accessing this map using `string` values.  There's no good way to take `myRawJSONExpression[0]` and refine it so that flow knows that it's an `ExpressionName`.

src/style-spec/function/expressions.js
anandthakker: Done in 013a59e06236c988347787d1ab0d40d890b174a3

src/style-spec/function/expressions.js
anandthakker: Done in 2b91368f134c2abbf5066d0874e8d0407e7e5d7a

src/style-spec/function/evaluation_context.js
anandthakker: Maybe we could always print names like:
 - Value: `JSONValue`
 - Vector<T>: `T[]`
 - Array<T, N>: `T[N]`

anandthakker: Or, if we think `T[]` is _too_ concise and easy to misread, we could do:
- Vector<T>: `Array of T's`
- Array<T, N>: `Array of N T's`

jfirebaugh:
For consistency with `get`/`has`, we should also reorder `at`'s arguments

docs/style-spec/expressions.md
```diff
@@ -134,7 +134,7 @@ Every expression evaluates to a value of one of the following types.
 ###Decision:
 - `["case", cond1: Boolean, result1: T, cond2: Boolean, result2: T, ..., cond_m, result_m: T, result_otherwise: T] -> T`
 - `["match", x: T, a_1: T, y_1: U, a_2: T, y_2: U, ..., a_m: T, y_m: U, y_else: U]` - `a_1`, `a_2`, ... must be _literal_ values of type `T`.
-- `["is_error", expr: T]` - `true` if `expr` is an `Error` value, `false` otherwise
+- `["coalesce", e1: T, e2: T, e3: T, ...] -> T` - evaluates each expression in turn until the first non-error value is obtained, and returns that value.
```
anandthakker: This should probably be "non-error **or non-null**"

anandthakker:
If a user specifies a property function with, say, numeric stop outputs, for a style property whose spec definition specifies a string, we accept the mistyped function and just yield the default.  IMO, this should be a validation error instead.

anandthakker:
@jfirebaugh @ansis okay, there's still some remaining tail work (colorSpace support, actually accepting `{expression:}` values, validation, etc.), but I think this is ready for a substantive review.  I'd love to get some feedback on the approach here before I get going on the native port.

### Notes

 - [`parse`](https://github.com/mapbox/mapbox-gl-js/pull/4841/commits/9a3cfc5d939d7f5cd4938c8154a87933c585db0e#diff-5915208d8f6c6e5a02efac320df843d5R38) converts a raw JSON expression into a TypedExpression tree where each node is tagged with the possibly-generic `Type` specified in that expression's definition.
 - [`typecheck`](https://github.com/mapbox/mapbox-gl-js/pull/4841/commits/9a3cfc5d939d7f5cd4938c8154a87933c585db0e#diff-1436f258974cf90a75f4fc014c108843) actually checks that each expression's result type matches the expected, and also returns a new copy of the expression tree with typenames resolved. 
 Inference is fairly naive:
   - If we're checking `e: (T, ...)->T` against an expected concrete type `c`, then we infer `T = c`.
   - If checking `e: (T, ...)->T` against a generic type, we make no inference and proceed to checking e's arguments without knowing `T`.
     - For any of parameter of `e` whose type is `T`, if the corresponding subexpression has a concrete result type `r` (or is a literal or type `r`), then we infer `T = r`.  If the corresponding subexpression is itself generic, we make no inference.
     - If, after checking all the arguments, we've failed to infer `e`'s result type, type checking fails.
   - These same rules apply if `T` in the above is replaced with a complex generic type like `Array<T>`.  If we're checking a generic vs. a generic, we still make no inference, but we _do_ check the structure of the type (e.g. `Array<T, 2>` would fail to match against `Array<U, 4>` or just `U`).  And if we're checking, say `Array<T, 2>` vs `Array<Number, 2>`, then we infer `T=Number`.
- [`expressions.js`](https://github.com/mapbox/mapbox-gl-js/pull/4841/commits/9a3cfc5d939d7f5cd4938c8154a87933c585db0e#diff-f809a6e3c491945cc761b48ca037a76a) declares 'definitions' of all our expressions, for each one indicating its type and providing the details of how it's compiled to a JS expression.
   - To properly handle certain cases where compiling to a pure JS expression doesn't suffice, we use an [evaluation context](https://github.com/mapbox/mapbox-gl-js/pull/4841/commits/9a3cfc5d939d7f5cd4938c8154a87933c585db0e#diff-a16de8c3a293513214b60e54bb438324) to provide helper functions, which is bound as `this` to the generated function.

src/style-spec/function/expression_name.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+// This would ideally be in expressions.js, but pulled into separate file
+// to avoid circular imports, due to https://github.com/facebook/flow/issues/3249
+
+/*::
+ export type ExpressionName = "literal" | "ln2" | "pi" | "e" | "string" | "number" | "boolean" | "json_array" | "object" | "get" | "has" | "at" | "typeof" | "length" | "zoom" | "properties" | "geometry_type" | "id" | "case" | "match" | "coalesce" | "==" | "!=" | ">" | ">=" | "<=" | "<" | "&&" | "||" | "!" | "curve" | "step" | "exponential" | "linear" | "cubic-bezier" | "+" | "-" | "*" | "/" | "%" | "^" | "log10" | "ln" | "log2" | "sin" | "cos" | "tan" | "asin" | "acos" | "atan" | "ceil" | "floor" | "round" | "abs" | "min" | "max" | "concat" | "upcase" | "downcase" | "rgb" | "rgba" | "color" | "color_to_array"
```
1ec5: `json_array`, `geometry_type`, and `color_to_array` are snake-case, but `cubic-bezier` is train-case and `typeof` is a run-in. Would it be helpful to make some of these identifiers consistent?

anandthakker: 👍 good point.  The style spec makes heavy (exclusive?) use of train-case, so perhaps we should stick with that?

src/style-spec/function/type_check.js
```diff
@@ -0,0 +1,300 @@
...
+                    // If we've consumed a full sequence, append a copy of this
+                    // list of types to the expanded argument list
+                    if (j === 0) {
+                        sequenceCount++;
+                        expandedArgTypes.push.apply(expandedArgTypes, t.types);
+                        util.extend(typenames, argTypeNames);
+                        argTypeNames = {};
+                    }
+                }
+                vi -= j;
```
anandthakker: This loop/logic could be simplified if we make the (pretty reasonable) restriction that an expression can only have one `NArgs` pack in its parameter list.  With that restriction, we could simply calculate the number of times we need to repeat the NArgs parameter pack to form an unrolled parameter list that matches the number of arguments actually provided.

src/style-spec/function/types.js
```diff
@@ -0,0 +1,118 @@
...
+        N
+    };
+}
+
+function lambda(result: Type, ...args: Array<Type>) /*: LambdaType */ {
+    return {
+        kind: 'lambda',
+        name: `(${args.map(a => a.name).join(', ')}) => ${result.name}`,
+        result,
+        args
```
anandthakker: I think using the word `parameters` for function input _types_ and `arguments` for input _values_ would make the type checking code a lot more readable.

src/style-spec/function/type_check.js
```diff
@@ -0,0 +1,290 @@
...
+        //  - collect typename mappings when ^ succeeds or type errors when it fails
+        for (let i = 0; i < argValues.length; i++) {
+            const param = expandedParams[i];
+            const arg = argValues[i];
+            const error = match(
+                resolveTypenamesIfPossible(param, typenames),
+                arg.type,
+                typenames
+            );
+            if (error) errors.push({ key: arg.key, error });
```
anandthakker: Is it valuable to collect & return multiple type checking errors?  If we didn't do this and instead failed at the first error, we could just `throw`, and it would probably make things cleaner.  On the other hand, for expression authors, being able to see & fix multiple errors at once might be nice.

src/style-spec/function/type_check.js
```diff
@@ -0,0 +1,290 @@
...
+            expectedTypenames[expected.typename] = t;
+        }
+        return null;
+    }
+
+    if (t.kind === 'typename' && !isGeneric(expected)) {
+        if (!tTypenames[t.typename] && t !== NullType) {
+            tTypenames[t.typename] = expected;
+        }
+        t = expected;
```
anandthakker: Oops - we can just return `null` here.

src/style-spec/function/types.js
```diff
@@ -0,0 +1,118 @@
...
+const BooleanType = primitive('boolean');
+const ColorType = primitive('color');
+const ObjectType = primitive('object');
+
+const ValueType = variant(
+    NullType,
+    NumberType,
+    StringType,
+    BooleanType,
+    ColorType,
```
jfirebaugh: Why is `ColorType` included here?

test/integration/expression-tests/typeof/basic/test.json
```diff
@@ -0,0 +1,75 @@
...
+  ],
+  "expected": {
+    "compileResult": {
+      "result": "success",
+      "isFeatureConstant": false,
+      "isZoomConstant": true,
+      "type": "string"
+    },
+    "evaluateResults": [
+      "Null",
```
jfirebaugh: What's the rationale for when the first letter of a type name is upper case versus lower case? Here it's upper, but in error messages it's (sometimes) lower.

jfirebaugh:
I think we should simplify array typing. Pushed a prototype in f249fca74cde7e2e8dbb850f24e117800d41586f; the remaining issue is the ambiguity of whether `"array"` is a constructor or a coercion. Seems like we need two different forms:

* `["array", ...]` -- constructor. Example: `["array", 1, 2]`.
* `["->array", o]` -- "coercer". Examples: `["->array", ["get", "foo"]]` (which would never actually do any coercing, only runtime type checking); `["->array", ["color", "green"]]` (replaces `"color_to_array"`).

This isn't perfect though. Elsewhere we're using `"type"` to be the coercer, and we need to distinguish between a coercer and a simple type assertion, which is not the same as a constructor. Do we need actually need _three_ different forms for each type: constructor, coercion, and assertion?

src/style-spec/function/types.js
anandthakker: Oops, it shouldn't be.

test/integration/expression-tests/typeof/basic/test.json
anandthakker: No rationale--we should be consistent, and I think uppercase first letter is probably the way to go.

anandthakker:
Yeah, I've been stumped, design-wise, on the construction / coercion / assertion issue.

Array assertions and coercers are problematic -- either you can only assert that an expression is `Array<T>` (or maybe `Array<T, N>`), or else we have to get into providing a syntax for expressing arbitrary types like `Array<number>` or `Array<Array<number, 2>>`, which seems... undesirable.

Do we actually need array assertion/coercion for any case other than (a) annotating a JSON value as an array or (b) Color => components?   If not, I would be inclined to use:
 - `json_array: Value -> Array<Value>` for (a)
 - something like `components: (Color, colorSpace: string) -> Array<number, 4>` for (b)
 - and leave `[array, a1, a2, ...]` as a constructor.

For the primitive types, I think coercion will be used much more commonly than assertion, which makes me want coercion to be as concise as possible -- e.g., `string` preferable to `to_string`.   One possibility would be to have coercion/assertion behavior depend on a parameter, e.g., `['string', val: T, strict: boolean = false]`, so for `{x: 5}`, `['string', ['get', 'x']]` evaluates to `"5"`, but `['string', ['get', 'x'], true]` would produce a runtime error.

jfirebaugh:
When I look at real world examples like in #4836, I want to be able to extract array and object literals as constants, assigned in `"let"` expressions, and then write the guts of the expression with a data-driven dispatch into the constant. So, yes, if we make them possible/convenient, I think there will be use cases for arrays and objects outside of (a) and (b).

If/when we do support this for both objects and arrays, they should be parallel in syntax. Here are some possibilities:

* `["array", [...]]` and `["object", {...}]`, with no evaluation "inside" the argument.
* `["array", value, value...]`, and `["object", string, value, string, value...]`, with evaluation of arguments.
* `["quote", [...]]` and `["quote", {...}]`, with no evaluation "inside" the argument. A lispish single special form; provides the option to have `"quasiquote"` form in the future for partial inner evaluation. I also like this form because it leaves the names `"array"` and `"object"` free for assertion/coercion. We could choose less lispish names if you prefer, for example "quote" ⇢ "literal", "quasiquote" ⇢ "template", "unquote" ⇢ "substitute".

jfirebaugh:
> For the primitive types, I think coercion will be used much more commonly than assertion, which makes me want coercion to be as concise as possible

Funny, I have the opposite instinct. I would think assertion/annotation to be the more common (especially if we expect Studio to auto-insert annotations based on tilestats), and coercion to be the more dangerous, and hence better suited to the less concise name.

src/style-spec/function/expressions.js
```diff
@@ -0,0 +1,425 @@
...
+    'ln2': defineMathConstant('ln2'),
+    'pi': defineMathConstant('pi'),
+    'e': defineMathConstant('e'),
+    'string': {
+        name: 'string',
+        type: lambda(StringType, ValueType),
+        compile: args => ({ js: `String(${args[0].js})` })
```
jfirebaugh: Is it wise to inherit all of JS's coercion quirks?

* `["string", ["^", 2, 70]]` ⇢ "1.1805916207174113e+21"
* `["string", ["properties"]]` ⇢ "[object Object]"
* `["string", ["array", 1, 2]]` ⇢ "1,2"
* `["string, ["get", "weird"]]` + `{"properties": {"weird": {"toString": "asdf"}}` ⇢ TypeError

anandthakker:
> Funny, I have the opposite instinct. I would think assertion/annotation to be the more common (especially if we expect Studio to auto-insert annotations based on tilestats)

Good point -- I was thinking more that coercion would be more common in _manually-authored_ expressions, which is why I was going for conciseness...

> coercion to be the more dangerous, and hence better suited to the less concise name.

... but this is convincing.  👍 for making coercion less concise / more explicit.

src/style-spec/function/expressions.js
anandthakker: I do think it seems reasonable to lean on the platform's number->string conversion...  as for objects and arrays... would it be reasonable to just punt, requiring primitive types only?

jfirebaugh: 👍 👍

1ec5:
> Good point -- I was thinking more that coercion would be more common in manually-authored expressions, which is why I was going for conciseness...

This was my thought as well. Runtime styling doesn’t have the benefit of auto-annotation via tilestats. Per https://github.com/mapbox/mapbox-gl-native/issues/8074#issuecomment-311261083, I don’t think developers on iOS will mind a little danger if they can avoid mandatory type annotations (which would have major usability issues).

src/style-spec/function/evaluation_context.js
```diff
@@ -97,13 +98,8 @@ module.exports = () => ({
...
     array: function(type, items) {
         if (!type) {
-            let itemtype;
-            for (const item of items) {
-                if (!itemtype) itemtype = typeof item;
-                else if (itemtype !== typeof item) itemtype = 'Value';
-            }
-            type = (!itemtype || itemtype === 'Value') ?
-                'Array' : `Array<${titlecase(itemtype)}>`;
+            const t = ArrayLiteral.inferArrayType(items);
```
anandthakker: @jfirebaugh This change causes the runtime typing of `[get, foo]` when `foo` is an array to be as specific as possible, based on its actual contents:
 - the array is homogenous of type number, string, or boolean, then => `Array<typeof item, properties.foo.length>`
 - otherwise => `Array<Value, properties.foo.length>`

src/style-spec/function/expression.js
```diff
@@ -80,39 +73,22 @@ class BaseExpression {
...
-                break;
-            }
-        }
+    static parse(args: Array<mixed>, context: ParsingContext) {
+        if (args.length !== 1)
+            throw new ParsingError(context.key, `'literal' expression requires exactly one argument, but found ${args.length} instead.`);
 
-        return array(itemType || ValueType, value.length);
-    }
+        if (!isValue(args[0]))
```
anandthakker: As of now, there's actually no way to express a color literal -- `rgba` and `color` are functions -- but this would allow `['literal', new Color(...)]`.

src/style-spec/function/expression.js
```diff
@@ -145,12 +145,12 @@ function parseExpression(expr: mixed, context: ParsingContext) : Expression {
         return new LiteralExpression(key, NumberType, expr);
     } else if (Array.isArray(expr)) {
         if (expr.length === 0) {
-            throw new ParsingError(key, `Expected an array with at least one element.`);
+            throw new ParsingError(key, `Expected an array with at least one element. If you wanted a literal array, use ["literal", []].`);
```
anandthakker: 👍  to this and the others like it.

jfirebaugh:
Now that we have an AST made up of `Expression` instances, we should define things like `isFeatureConstant` and `isZoomConstant` as independent recursive functions rather than lumping them in with the compilation step.

`Expression#compile` itself should be typed `() => string | Array<CompileError>`.

src/style-spec/function/expression.js
jfirebaugh: Yeah, I figured that would be okay -- since we're not likely to export `Color` publicly, there no way to construct such an expression. And even if there were, it'd be harmless to do so.

anandthakker:
> Now that we have an AST made up of Expression instances, we should define things like isFeatureConstant and isZoomConstant as independent recursive functions rather than lumping them in with the compilation step.
> Expression#compile itself should be typed () => string | Array<CompileError>.

💯 I'm doing this as part of implementation of `let`.

anandthakker:
Few notes/questions about `let`:
 - Uses [let](https://docs.racket-lang.org/reference/let.html) (as opposed to `let*` or `letrec`) semantics.
 - We _could_ make `Reference` part of `LiteralExpression` and `LetExpression` a subclass of `LambdaExpression`.  I kept the former separate for semantic reasons, and the latter also because it requires specialized type checking.
 - To keep things simple for now, a `[zoom]` curve is not allowed in a let binding; it can only appear in the _result_ of a `let`, even though we theoretically could accept, e.g. `[let, 'a', [curve, [linear], [zoom], 0, 0, 10, 10], ['a']]`
 - Currently, a let binding that produces an error will cause the entire let expression to be an error, even if it would have been unused.  E.g. `[let, 'a', [number, [get, x]], [case, false, [a], 0]]`.  What's more natural: the present behavior, or only having an error propagated if the erroneous binding is actually referenced?

src/style-spec/function/expression.js
```diff
@@ -168,6 +198,103 @@ class LambdaExpression extends BaseExpression {
     }
 }
 
+class Reference extends BaseExpression {
+    name: string;
+    constructor(key: string, name: string, type: Type) {
+        super(key, type);
+        if (!/^[a-zA-Z_]+[a-zA-Z_0-9]*$/.test(name))
+            throw new ParsingError(key, `Invalid identifier ${name}.`);
```
anandthakker: This check should be moved to `LetExpression.parse()`, and it should also either check against JS reserved words or else 'escape' them (e.g. replace `function` with `var_function` or something)

anandthakker:
Squashed this down, FF merging into `feature/expressions` branch

PR #4842, opened 2017-06-16
LogoControl cleanup

### LINKED ISSUE #4813: There was an addtional div when LogoControl was removed
**mapbox-gl-js version**:
0.37.0
### Steps to Trigger Behavior

https://jsfiddle.net/jingsam/ue99mkgy/1/

### Expected Behavior

there should not exsit a `mapbox-ctrl` div

mollymerp:
fix #4813 

This changes a couple things. 

I added a `sourcedata` event firing when a source is removed. This seems in line with the current description in our docs:
>Fired when one of the map's sources loads or changes

I also follow @jingsam's recommendation of hiding the LogoControl when no Mapbox sources are present instead of removing it from the DOM. 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/control/logo_control.js
+++ b/src/ui/control/logo_control.js
@@ -19,10 +19,16 @@ class LogoControl {
 
     onAdd(map) {
         this._map = map;
+        this._logoDisplayed = true;
+
         this._container = DOM.create('div', 'mapboxgl-ctrl');
+        const anchor = DOM.create('a', 'mapboxgl-ctrl-logo');
+        anchor.target = "_blank";
+        anchor.href = "https://www.mapbox.com/";
+        anchor.setAttribute("aria-label", "Mapbox logo");
+        this._container.appendChild(anchor);
 
         this._map.on('sourcedata', this._updateLogo);
-        this._updateLogo();
         return this._container;
     }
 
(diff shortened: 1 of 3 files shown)


src/ui/control/logo_control.js
```diff
@@ -38,15 +45,12 @@ class LogoControl {
...
-                this._container.appendChild(anchor);
-                this._map.off('data', this._updateLogo);
-            } else if (this._container.childNodes.length && !this._logoRequired()) {
-                this.onRemove();
+            if (!this._logoDisplay && this._logoRequired()) {
+                this._container.style.display = 'block';
+                this._logoDisplay = true;
+            } else if (this._logoDisplay && !this._logoRequired()) {
+                this._container.style.display = 'none';
+                this._logoDisplay = false;
```
jingsam: We could simply toggle display like this:
```
this._container.style.display = this._logoRequired() ? 'block' : 'none'
```

The state variable `this._logoDisplay` seems unnecessary.

src/ui/control/logo_control.js
```diff
@@ -21,9 +21,14 @@ class LogoControl {
...
         this._container = DOM.create('div', 'mapboxgl-ctrl');
+        const anchor = DOM.create('a', 'mapboxgl-ctrl-logo');
+        anchor.target = "_blank";
+        anchor.href = "https://www.mapbox.com/";
+        anchor.setAttribute("aria-label", "Mapbox logo");
+        this._container.appendChild(anchor);
+        this._container.style.display = 'none';
 
         this._map.on('sourcedata', this._updateLogo);
-        this._updateLogo();
```
jfirebaugh: I think updating on add was a good idea. Instead of removing it, how about changing the condition in `_updateLogo` to `!e || e.sourceDataType === 'metadata')`?

mollymerp: Sure! I actually had that for a second, but after I moved the `DOM.create` stuff to `onAdd` I didn't think it was necessary anymore. But there is a potential race condition if  source fires its `metadata` event before the LogoControl is added 👍

PR #4843, opened 2017-06-16
include MouseEvents polyfill to allow map rotation via compass button on IE11

### LINKED ISSUE #4784: Navigation Control compass click event IE 11 throws: Object doesn't support this action
mapboxgl - v.0.37.0
Internet Explorer - 11.0.38

When you click the compass the following error is thrown only in Internet Explorer:

Object doesn't support this action
mapbox-gl.js, Line: 370, Column: 41

If you click on the error it takes me to line 105 in function copyMouseEvent(e) 

This can be duplicated using the examples page: [Navigation Control](https://www.mapbox.com/mapbox-gl-js/example/navigation/)

Open IE Debugger Tools (F12) and you will see the error above.

In the example page, the map will go ahead and return to due North, but in  applications that use IE 11 embedded the JavaScript error raises an alert window and will not allow the application to continue.

andrewharvey:
Closes #4784 

## Launch Checklist

This PR include a MouseEvents polyfill to fix #4784. What's the feeling towards using polyfills in GL JS to maintain support for older browsers?

This change seems to break a whole lot of tests, I can look into that once I get feedback on if this approach will be accepted or not.

 - [ ] post benchmark scores
tested before and after on IE11.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "brfs": "^1.4.0",
     "bubleify": "^0.7.0",
     "earcut": "^2.0.3",
+    "events-polyfill": "^1.1.9",
     "geojson-rewind": "^0.1.0",
     "geojson-vt": "^2.4.0",
     "grid-index": "^1.0.0",
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -6,6 +6,8 @@ const util = require('../../util/util');
 
 const className = 'mapboxgl-ctrl';
 
+require('events-polyfill');
+
 /**
  * A `NavigationControl` control contains zoom buttons and a compass.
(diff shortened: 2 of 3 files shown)


jfirebaugh:
> What's the feeling towards using polyfills in GL JS to maintain support for older browsers?

If the dependency is minimal, it can be ok to use local conditionals like `const MouseEvent = window.MouseEvent || require('mouse-event');`, but GL JS shouldn't modify any globals, as that could introduce conflicts with polyfills an application using GL JS wants to install.

andrewharvey:
sorry I haven't had the time to work on this, good to see progress in https://github.com/mapbox/mapbox-gl-js/pull/5157

PR #4847, opened 2017-06-16
Update docs and test to newer version of mapbox-gl-rtl-text plugin

ChrisLoer:
Upgrade to `mapbox-gl-rtl-text` v0.1.1 to take advantage of smaller package size due to newer version of Emscripten (1.37) stripping more out of the runtime.

I've done sanity checks of the example page, and the Arabic unit tests pass with the new build. I've also tested "time to first tile load" by logging a timestamp from `VectorTileSource#done`, and don't see a noticeable performance change. The plugin isn't used by our benchmarks, so they're unaffected.

### DIFF
--- a/docs/_posts/examples/3400-01-31-mapbox-gl-rtl-text.html
+++ b/docs/_posts/examples/3400-01-31-mapbox-gl-rtl-text.html
@@ -9,7 +9,7 @@ tags:
 <div id='map'></div>
 
 <script>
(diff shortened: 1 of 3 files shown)


ChrisLoer:
`tile.test.js` initially failed on 'expires header set date as expected'... which seemed potentially timing related, so I restarted the build and the error went away. Nothing to see here....

Issue #4825, opened 2017-06-12
Base label "keep-upright" orientation based on start and end of label


ChrisLoer:
With https://github.com/mapbox/mapbox-gl-js/pull/4781, labels will now re-orient themselves based on perspective. 🎉 

The current approach determines label orientation by looking at whether the segment the anchor of the label is on is upright: this can lead to labels that look "upside down" even though technically the segment their anchor is on is "right side up".

![image](https://user-images.githubusercontent.com/375121/26845174-8130bbdc-4aaa-11e7-886d-e8c8933fc766.png)

I think it would probably be a better heuristic if we detected the orientation of a line label based on the beginning and the end of the label instead of the orientation of the middle line segment (although there could certainly be cases where this looked less natural).

/cc @nickidlugash @ansis

nickidlugash:
I think this makes sense 👍  While there will be instances where the middle segment is more representative of a larger proportion of the label, I think it's probably easier to read a label that _starts_ upright, all else being equal.

PR #4853, opened 2017-06-19
Orient line labels based on the relative position of their first and last glyphs

ChrisLoer:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/4825

Before:

image

After:
screenshot 2017-06-19 14 15 47

This turned out to add a little more complexity to the code than I would have liked, but it shouldn't have too much of a performance impact. Because we have to project the first and last glyph to find out whether to do the flip, there's some extra work in the case that we flip (because we have to reproject the first and last glyph once they're flipped). On the other hand, there's a small savings here in that when there's not enough room on a line for a label, we exit earlier because we test the beginning and end before projecting all the glyphs in the middle.

/cc @ansis @nickidlugash


ChrisLoer:
@ansis that sounds like a good idea, what do you think of the new flipping approach I tried? Instead of just returning a boolean I return two bits, so that we can still exit early in the case the glyphs don't fit on the line unflipped (it is _possible_ they would fit if we flipped them, but the vast majority of the time they won't so we might as well save the extra work).

ansis:
Looks good!

PR #4868, opened 2017-06-22
Enhance render test results page

anandthakker:
- Include copyable list of failed tests at top to make it easier to
   rerun only the failed tests
 - If there are failing tests, filter out passed tests, with a button to reveal them


asheemmamoowala:
I made the same change yesterday too ! 💯

Issue #4859, opened 2017-06-21
RTL text plugin is not applied to styles initialized after the plugin is loaded


petrsloup:
**mapbox-gl-js version**: 0.38.0

### Steps to Trigger Behavior

 1. `mapboxgl.setRTLTextPlugin('https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.1.1/mapbox-gl-rtl-text.js');`
 2. Initialize the map (or even `setStyle` is enough) **after** the plugin is loaded

Compare:
https://jsbin.com/kepafevoho/edit?html,output (style passed via constructor)
https://jsbin.com/lixutelitu/edit?html,output (style set via `setStyle` after 1 s timeout)

### Expected Behavior
The RTL texts should be displayed correctly even when using `setStyle`.
The two demos above should produce visually equal results.

### Actual Behavior
The RTL does not work (the map does not seem to know about the plugin, but calling `setRTLTextPlugin` throws `setRTLTextPlugin cannot be called multiple times.`.

Looks like the plugin is only "applied" on the existing map and style **once** after it is loaded... ?

jfirebaugh:
cc @ChrisLoer

ChrisLoer:
Thanks for the super-clear report, @petrsloup.

There's a race condition here. The first time you create a style, we start running background web workers. When you register the RTL text plugin, we send a notification to the web workers to load the code. The code assumes the workers will be running before the plugin becomes available, but in your example that's not the case, because the plugin has a chance to fully load before your initial call to `setStyle`. In that case, the "load plugin" notification we sent to the workers gets ignored.

We'll fix the bug, but until then note that the bug only applies to the ordering of the initial `setStyle` vs. `setRTLTextPlugin`. After the background workers are running, changing styles should work just fine.

petrsloup:
Oh, I see. Thanks for the workaround.

ChrisLoer:
Err, my race condition explanation was totally bogus. 😅 Turns out to be a simple regression I introduced in https://github.com/mapbox/mapbox-gl-js/commit/171db6177a6a5e7246fcef4d74ed297ee5a196d8 by passing the wrong arguments to the plugin registration callback in the case the plugin was already available.

hakimelek:
Hello, I am still experiencing the same issue on rerenders. Is there a way to detect that `setRTLTextPlugin` was already being called?

arindam1993:
[...]

PR #4870, opened 2017-06-22
Fix issue #4859: allow loading RTL text plugin before first style load.

ChrisLoer:
Fixes issue #4859: this was a regression I introduced in https://github.com/mapbox/mapbox-gl-js/commit/171db6177a6a5e7246fcef4d74ed297ee5a196d8. Unfortunately I still haven't figured out a good way to exercise this code in our test suite, since it depends on running in the browser. :/

/cc @anandthakker @mollymerp

### DIFF
--- a/src/source/rtl_text_plugin.js
+++ b/src/source/rtl_text_plugin.js
@@ -23,6 +23,11 @@ module.exports.registerForPluginAvailability = function(
     return callback;
 };
 
+// Exposed so it can be stubbed out by tests
+module.exports.createBlobURL = function(response: Object) {
+    return window.URL.createObjectURL(new window.Blob([response.data]), {type: "text/javascript"});
+};
+
 module.exports.setRTLTextPlugin = function(pluginURL: string, callback: ErrorCallback) {
     if (pluginRequested) {
         throw new Error('setRTLTextPlugin cannot be called multiple times.');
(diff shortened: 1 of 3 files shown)


ChrisLoer:
@mollymerp I had to hack things up a bit to do the stubbing (I'm not very experienced with Sinon), but I extracted the hard-to-test `createBlobURL` and put in a test that verifies a bogus `pluginBlobURL` gets passed down to the workers. Do you think that makes sense?

mollymerp:
Thanks for adding that test!
Would adding a `calledWith` check to make sure the correct arguments are passed to the test here https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/style/style.test.js#L59 have caught this specific regression? Sorry if this is a dumb question – I'm not too familiar with the RTL plugin 😬

ChrisLoer:
@mollymerp A `calledWith` check on the original test (https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/style/style.test.js#L59) wouldn't have worked originally because it didn't have a stubbed out `createBlobURL` (one confusing thing here is that the the URL we ultimately care about is the "blob URL", not the "plugin URL" that gets passed in to `setRTLTextPlugin`). I'll try adding a `setRTLTextPlugin` call to the test now that we've got the stubbing mechanism.

ChrisLoer:
I had to add a `clearRTLTextPlugin` method for the tests to allow both tests to modify the plugin... global singletons don't play well with tests. :/

PR #4878, opened 2017-06-23
Update yarn.lock

jfirebaugh:
Missing from #4847.


jfirebaugh:
Nah, we can wait until a release.

Issue #4081, opened 2017-02-01
Document "*-transition" properties


lucaswoj:
_From @jfirebaugh on September 22, 2014 18:19_

Document how `*-transition` properties work.

_Copied from original issue: mapbox/mapbox-gl-style-spec#171_

lucaswoj:
Both gl-js and gl-native support setting paint-property specific transition options. For example,the `circle-radius` transition may be controlled by setting the property `circle-radius-transition`. This should be clearly documented. 

cc @danswick

lucaswoj:
I just identified a minor outstanding bug related to this https://github.com/mapbox/mapbox-gl-js/issues/1996

lucaswoj:
_From @peterqliu on September 8, 2016 20:34_

I'm interested in working on this, as it's required for setting transition durations at runtime with `.setPaintProperty`. Is there some way we can generate these (very repetitive) entries in an automated fashion, or are we stuck writing these manually?

lucaswoj:
@peterqliu I think we can document this feature with a paragraph and some examples in the in the ["Transition" section](https://www.mapbox.com/mapbox-gl-style-spec/#transition). Reading an entry per paint property would be as repetitive as writing them!

lucaswoj:
_From @mollymerp on October 21, 2016 0:4_

This is an easy fix and something we should hit asap. @peterqliu do you want to knock this out ? Happy to jump in if not.

lucaswoj:
_From @lbud on January 25, 2017 19:58_

Came up again in https://github.com/mapbox/mapbox-gl-js/issues/4045

ibesora:
As a _victim_ of this issue who implemented my own version of fill-extrusion-height transition (see https://github.com/mapbox/mapbox-gl-js/pull/4888), may I help to document these properties so nobody does extra work again?

anandthakker:
@ibesora sorry for your wasted effort.  A documentation PR is always very welcome!

Perhaps the most straightforward way to document this would be to add a note in the [**Transition** section of the style spec documentation](https://www.mapbox.com/mapbox-gl-js/style-spec/#transition) indicating that in addition to the style's global `transition`, individual an individual paint property's transition behavior can be controlled by setting `*-transition`.

urschrei:
Hi, is feature working on JS? Using 42.2, and the following:

        map.addLayer({
            'id": 'foo-layer',
            'type": 'fill-extrusion',
            'source': 'foo',
[...]

PR #4888, opened 2017-06-26
Extrude animation

ibesora:
Adds some new parameters to the fill-extrusion (animation, duration, reversed) to animate the extrusion instead of painting it fully extruded


jfirebaugh:
Hey @ibesora, thanks for the contribution! However, I think Mapbox GL JS already provides the functionality you've implemented here. You can achieve animated transitions for fill extrusions by setting the `fill-extrusion-height-transition` paint property in the extrusions layer to a value like `{ "duration": 300 }`. This isn't documented well right now; we have an issue for improving the documentation in #4081.

PR #4893, opened 2017-06-26
Pause tile reloads when worker layer update is pending

### LINKED ISSUE #4738: Runtime styling sometimes doesn't take effect
When using a GeoJSON source, and setting runtime styling properties, then updating the GeoJSON source data, the changes to the style don't always take effect.

https://jsfiddle.net/ygxuk5zy/ contains a reduced test case (replace your access token). This is how it *should* look:

screen shot 2017-05-22 at 23 05 13

This is how it actually looks:

screen shot 2017-05-22 at 23 05 20

anandthakker:
Fixes #4738

### DIFF
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -35,7 +35,7 @@ class SourceCache extends Evented {
 
             // for sources with mutable data, this event fires when the underlying data
             // to a source is changed. (i.e. GeoJSONSource#setData and ImageSource#serCoordinates)
-            if (this._sourceLoaded && e.dataType === "source" && e.sourceDataType === 'content') {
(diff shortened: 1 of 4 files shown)


anandthakker:
This isn't pretty, but (a) it can hold us over until we can take on larger refactors, and (b) it's probably good to wire in this dependency--i.e. the need for batched-updated processing to control tile loading--before those refactors, so that we have to take it into account when we do them.

jfirebaugh:
Unfortunately this doesn't fix the issue. I'll add the test file to this branch so you can see.

anandthakker:
Shoot, my unit test was bogus.  Amended / force-pushed and I _think_ it should now be fixed. 

(Note: temporarily leaving the test file in for verification despite lint errors.)

PR #4902, opened 2017-06-27
Upgrade flow and eslint-plugin-flowtype

jfirebaugh:


jfirebaugh:
No, just making sure we're up to date.

Issue #4876, opened 2017-06-23
Port per-tile glyph/icon atlases


jfirebaugh:
Port https://github.com/mapbox/mapbox-gl-native/pull/9213 / https://github.com/mapbox/mapbox-gl-native/pull/9355.

jfirebaugh:
The gl-native implementation relies on transferring references to immutable objects across threads. Unfortunately, we don't have that option in JS -- we can only copy JSON or move ArrayBuffers. For icons and glyphs, that means copying image and glyph data from the main thread to the worker. Might be expensive. :(

[`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) is a possible future solution, but it'll be a while before support is widespread.

PR #4903, opened 2017-06-27
Flow type workers

jfirebaugh:
This adds flow annotations to the majority of the `Worker`-related code. I started on #4876, wanted to make some refactors, and realized that it would be better to have the safety of some static type checking as I did.

The first commit here switches the eslint `sourceType` option to `module`, so that we can use [module types](https://flow.org/en/docs/types/modules/) without resorting to flow comments.

Q: Why not use flow comments?
A: They're ugly, and they're not as good as native type annotations: not syntax highlighted, not linted (e.g. no warnings if you import something and don't use it), no documentation.js support, no IDE integration.

Q: What are the side effects of `"sourceType": "module"`?
A: AFAICT, it doesn't force us to actually use non-type `import`/`export` rather than CommonJS `require`, it just means that we can't `'use strict'` in each file -- this is enforced by eslint because modules are always strict. I worked around this by writing a simple `strictify` transform that puts them back in. (I didn't use the published one because it's [buggy and unmaintained](https://github.com/jsdf/strictify/pull/4).)

Q: Shouldn't you be able to use `import type` without requiring `"sourceType": "module"`?
A: Yes, I think you should be. If you agree, 👍  [this bug](https://github.com/babel/babel-eslint/issues/474).

Beyond that, the eslint + flow combo pretty much works. The only issue I've found is that eslint chokes on unnamed parameters in function types. So just name them:

**Works**
```
type Callback = (error: ?Error, result: ?Result) => void;
```

**Doesn't Work**
```
type Callback = (?Error, ?Result) => void;
```

### DIFF
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -114,7 +114,7 @@ class Tile {
         }
 
         this.collisionBoxArray = new CollisionBoxArray(data.collisionBoxArray);
-        this.collisionTile = new CollisionTile(data.collisionTile, this.collisionBoxArray);
+        this.collisionTile = CollisionTile.deserialize(data.collisionTile, this.collisionBoxArray);
         this.featureIndex = new FeatureIndex(data.featureIndex, this.rawTileData, this.collisionTile);
         this.buckets = Bucket.deserialize(data.buckets, painter.style);
     }
@@ -129,7 +129,7 @@ class Tile {
     reloadSymbolData(data: any, style: any) {
         if (this.state === 'unloaded') return;
 
-        this.collisionTile = new CollisionTile(data.collisionTile, this.collisionBoxArray);
+        this.collisionTile = CollisionTile.deserialize(data.collisionTile, this.collisionBoxArray);
         this.featureIndex.setCollisionTile(this.collisionTile);
 
         for (const id in this.buckets) {
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -1,4 +1,3 @@
-'use strict';
 
 const util = require('../util/util');
 const Bucket = require('../data/bucket');
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -1,3 +1,4 @@
+// @flow
 
 const util = require('../util/util');
 const Bucket = require('../data/bucket');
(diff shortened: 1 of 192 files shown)


src/source/tile.js
```diff
@@ -19,11 +20,44 @@ const CLOCK_SKEW_RETRY_TIMEOUT = 30000;
...
+    vtLayers: any;
+
     /**
      * @param {TileCoord} coord
-     * @param {number} size
+     * @param size
+     * @param sourceMaxZoom
      */
-    constructor(coord, size, sourceMaxZoom) {
+    constructor(coord: any, size: number, sourceMaxZoom: number) {
```
anandthakker: should this be `coord: TileCoord`?

src/source/tile.js
```diff
@@ -38,20 +72,12 @@ class Tile {
...
-        // - `unloaded`:  Tile data has been deleted.
-        // - `errored`:   Tile data was not loaded because of an error.
-        // - `expired`:   Tile data was previously loaded, but has expired per its HTTP headers and is in the process of refreshing.
         this.state = 'loading';
 
         this.placementThrottler = new Throttler(300, this._immediateRedoPlacement.bind(this));
     }
 
-    registerFadeDuration(animationLoop, duration) {
+    registerFadeDuration(animationLoop: any, duration: any) {
```
anandthakker: why not `duration: number`?

src/source/worker.js
```diff
@@ -8,12 +9,30 @@ const assert = require('assert');
...
+        registerWorkerSource: (string, Class<WorkerSource>) => void,
+        registerRTLTextPlugin: (any) => void
+    };
+    actor: Actor;
+    layerIndexes: { [string]: StyleLayerIndex };
+    workerSourceTypes: { [string]: Class<WorkerSource> };
+    workerSources: { [string]: { [string]: WorkerSource } };
+
+    constructor(self: WorkerGlobalScope) {
+        this.self = (self: any);
```
anandthakker: 👍  I don't see a cleaner way to do this.  Could we stick a `//` comment here explaining the cast to `any`?

src/symbol/collision_tile.js
```diff
@@ -91,7 +119,7 @@ class CollisionTile {
         ];
     }
 
-    serialize(transferables) {
+    serialize(transferables /*: ?Array<Transferable> */): SerializedCollisionTile {
```
anandthakker: Why a comment for the parameter type?

src/source/tile.js
jfirebaugh: This was a rough pass; for many types I just used `any` where flow would otherwise complain. `tile_coord.js` hasn't been typed yet.

src/symbol/collision_tile.js
jfirebaugh: Oversight.

src/source/worker.js
jfirebaugh: Yeah, would be nice if `Object.assign(self, {registerWorkerSource: ..., registerRTLTextPlugin: ...})` worked, but it doesn't. I'll comment.

PR #4907, opened 2017-06-28
Use default flow-typed directory name

jfirebaugh:
While here, consistify ignores and remove one that seems unnecessary.


jfirebaugh:
Will update and merge #4819 after this.

PR #4909, opened 2017-06-28
Type browser

jfirebaugh:
* Switched to object literal so we can use `get` rather than `Object.defineProperty`.
* Add `@module` declaration to util.js to work around a documentation.js bug.
* Require mapbox-gl-supported in index.js directly

### DIFF
--- a/src/util/browser.js
+++ b/src/util/browser.js
@@ -1,87 +1,82 @@
-
-/**
- * @module browser
- * @private
- */
+// @flow
 
 const window = require('./window');
 
-/**
- * Provides a function that outputs milliseconds: either performance.now()
- * or a fallback to Date.now()
- */
-module.exports.now = (function() {
-    if (window.performance &&
-        window.performance.now) {
-        return window.performance.now.bind(window.performance);
-    } else {
-        return Date.now.bind(Date);
-    }
-}());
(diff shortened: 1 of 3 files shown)


src/util/browser.js
```diff
@@ -1,97 +1,86 @@
...
+     * Provides a function that outputs milliseconds: either performance.now()
+     * or a fallback to Date.now()
+     */
+    now: (() => {
+        if (window.performance &&
+            window.performance.now) {
+            return window.performance.now.bind(window.performance);
+        } else {
+            return Date.now.bind(Date);
+        }
```
anandthakker: Why use the immediately invoked function expression (IIFE) here (and below in `frame`, `cancelFrame`) rather than declaring `const now = (window.performance && window.performance.now) ? window.performance.now.bind(window.performance) : Date.now.bind(Date);` prior to defining the exported object?

No strong objection to this, but all else being equal, I think IIFE is maybe a bit less readable.

jfirebaugh: Keeps the initialization scoped to the relevant property and avoids extra variables at module scope. Although, given that other parts of the module want to use `now`, `frame`, etc., it probably is better to have local variables. I'll change this.

PR #4912, opened 2017-06-28
Fix geojson serialize to include all option keys.

brandonbloom:
This was causing style diffing to always fail when using additional
option keys, such as `cluster: true`.

There are probably other serialize functions that are insufficient in the same way. Serialization for all style types should probably be systematically tested for setStyle diffing friendliness.

### DIFF
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -72,6 +72,7 @@ class GeoJSONSource extends Evented {
         this.setEventedParent(eventedParent);
 
         this._data = options.data;
+        this._options = util.extend({}, options);
 
         if (options.maxzoom !== undefined) this.maxzoom = options.maxzoom;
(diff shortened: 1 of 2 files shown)


mollymerp:
@brandonbloom thank you for your contribution! Would you be able to add a unit test that would have caught this bug? Let me know if you need help with our test suite.

brandonbloom:
The right thing to do to test this is probably a fair bit more work than a targeted unit test.

In order to catch this class of issue systematically, one would need to enumerate all the possible style settings, construct a style object for each, serializing them, then run styleDiff and validate that it produces zero output. I haven't looked at the test suite at all yet, but does such a repository of example style configurations exist? Maybe this could be accomplished by generating random styles from the style spec?

Such a testing effort is probably beyond the time commitment I can make right now.

PR #4918, opened 2017-06-30
Type painter.js and shaders.js

jfirebaugh:


src/render/painter.js
```diff
@@ -25,14 +26,66 @@ const draw = {
...
+    lineAtlas: LineAtlas;
+    spriteAtlas: SpriteAtlas;
+    glyphSource: GlyphSource;
+    depthRange: number;
+    isOpaquePass: boolean;
+    currentLayer: number;
+    id: string;
+    _showOverdrawInspector: boolean;
+    cache: { [string]: Program };
+    currentProgram: Program;
```
anandthakker: I wonder if we should standardize on alphabetizing these 🤔

src/render/painter.js
```diff
@@ -400,24 +454,30 @@ class Painter {
...
 
         const numAttributes = gl.getProgramParameter(program, gl.ACTIVE_ATTRIBUTES);
         const result = {program, numAttributes};
 
         for (let i = 0; i < numAttributes; i++) {
             const attribute = gl.getActiveAttrib(program, i);
-            result[attribute.name] = gl.getAttribLocation(program, attribute.name);
+            if (attribute) {
+                result[attribute.name] = gl.getAttribLocation(program, attribute.name);
+            }
```
anandthakker: Alternative:

```js
if (!attribute) assert(false);
result[attribute.name] = gl.getAttribLocation(program, attribute.name);
```

It's a little weird, but (a) closer to what we really want to say (which is presumably just `assert(attribute)`) and easier to migrate if/when flow supports more flexible guards, (b) unassertify will make this into `if(!attribute) {}` which minification would remove, and (c) two LOC instead of 3 😏

src/render/painter.js
jfirebaugh: For now I'm either listing them in the order that flow complains about, or grouping by functionality. I think the latter is preferable. For classes like `Painter`, we clearly just have way too many members and should refactor to a smaller set.

src/render/painter.js
jfirebaugh: This doesn't work, unfortunately:

```
src/render/painter.js:465
465:             result[attribute.name] = gl.getAttribLocation(program, attribute.name);
                                  ^^^^ property `name`. Property cannot be accessed on possibly null value
465:             result[attribute.name] = gl.getAttribLocation(program, attribute.name);
                        ^^^^^^^^^ null
```

Flow doesn't know to treat `assert(false)` as diverging.

anandthakker: Oh, dang / duh.

PR #4919, opened 2017-06-30
Update script detection for Unicode 10

1ec5:
This change updates the script detection files to reflect changes in [Unicode 10](http://www.unicode.org/Public/10.0.0/ucd/Blocks.txt) and [UTR #50 revision 17](http://www.unicode.org/Public/vertical/revision-17/VerticalOrientation-17.html).

### DIFF
--- a/src/util/is_char_in_unicode_block.js
+++ b/src/util/is_char_in_unicode_block.js
@@ -1,6 +1,6 @@
 // @flow
 
-// The following table comes from <http://www.unicode.org/Public/9.0.0/ucd/Blocks.txt>.
+// The following table comes from <http://www.unicode.org/Public/10.0.0/ucd/Blocks.txt>.
 // Keep it synchronized with <http://www.unicode.org/Public/UCD/latest/ucd/Blocks.txt>.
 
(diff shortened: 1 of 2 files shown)


src/util/is_char_in_unicode_block.js
```diff
@@ -244,6 +248,8 @@ const unicodeBlockLookup: UnicodeBlockLookup = {
     // 'Tangut': (char) => char >= 0x17000 && char <= 0x187FF,
     // 'Tangut Components': (char) => char >= 0x18800 && char <= 0x18AFF,
     // 'Kana Supplement': (char) => char >= 0x1B000 && char <= 0x1B0FF,
+    // 'Kana Extended-A': (char) => char >= 0x1B100 && char <= 0x1B12F,
+    // 'Nushu': (char) => char >= 0x1B170 && char <= 0x1B2FF,
```
1ec5: Nushu and CJK Unified Ideographs Extension F consist of characters that should be given ideographic breaking and vertical layout, but for our lack of astral plane support: #4001.

PR #4923, opened 2017-06-30
Type feature_index.js

jfirebaugh:


src/source/worker_tile.js
```diff
@@ -66,7 +66,7 @@ class WorkerTile {
         const sourceLayerCoder = new DictionaryCoder(Object.keys(data.layers).sort());
 
         const featureIndex = new FeatureIndex(this.coord, this.overscaling);
-        featureIndex.bucketLayerIDs = {};
+        featureIndex.bucketLayerIDs = {}; // Should this be an array?
```
anandthakker: Good point - looks to me like it should.

PR #4932, opened 2017-07-03
GeoJSON and VectorTile types

jfirebaugh:
Includes some code changes to support the type annotations:

* `GeoJSONWrapper` implements the `VectorTile` interface as well as the `VectorTileLayer` interface. This eliminates special `_geojsonTileLayer` cases from `GeoJSONWorkerSource` and `WorkerTile`.
* Use composition instead of augmenting vector tile objects with `rawData`, `expires`, and `cacheControl` properties. (We still augment individual features with `index` and `sourceLayerIndex`; this should be cleaned up in a subsequent refactor.)

### DIFF
--- a/flow-typed/geojson.js
+++ b/flow-typed/geojson.js
@@ -1,0 +1,39 @@
+type Position = [number, number] | [number, number, number];
+type Geometry<T, C> = { type: T, coordinates: C }
+
+declare type      GeoJSONPoint = Geometry<     'Point',       Position>;
+declare type GeoJSONMultiPoint = Geometry<'MultiPoint', Array<Position>>;
+
+declare type      GeoJSONLineString = Geometry<     'LineString',       Array<Position>>;
+declare type GeoJSONMultiLineString = Geometry<'MultiLineString', Array<Array<Position>>>;
+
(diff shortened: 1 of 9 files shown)


flow-typed/geojson.js
```diff
@@ -0,0 +1,39 @@
...
+    properties: {},
+    id?: number | string
+};
+
+declare type GeoJSONFeatureCollection = {
+    type: 'FeatureCollection',
+    features: Array<GeoJSONFeature>
+};
+
+declare type GeoJSON = GeoJSONGeometry | GeoJSONFeature | GeoJSONFeatureCollection;
```
anandthakker: 😍

anandthakker:
> GeoJSONWrapper implements the VectorTile interface as well as the VectorTileLayer interface. This eliminates special _geojsonTileLayer cases from GeoJSONWorkerSource and WorkerTile.

>Use composition instead of augmenting vector tile objects with rawData, expires, and cacheControl properties. (We still augment individual features with index and sourceLayerIndex; this should be cleaned up in a subsequent refactor.)

💯

PR #4933, opened 2017-07-03
Type rest of data directory

jfirebaugh:


src/data/bucket.js
```diff
@@ -27,16 +56,41 @@ const util = require('../util/util');
...
+    index: number;
+    zoom: number;
+    overscaling: number;
+    layers: Array<StyleLayer>;
+    buffers: BufferGroup;
+    arrays: ArrayGroup;
+
+    +addFeature: (feature: VectorTileFeature) => void;
+
+    static deserialize(input: Array<SerializedBucket>, style: Style) {
```
anandthakker: Return type `: {[string]: Bucket}`

src/data/bucket/symbol_bucket.js
```diff
@@ -175,7 +192,40 @@ function addCollisionBoxVertex(layoutVertexArray, point, anchor, extrude, maxZoo
  * @private
  */
 class SymbolBucket {
-    constructor(options) {
+    static programInterfaces: {
+        ['glyph' | 'icon' | 'collisionBox']: ProgramInterface
+    };
```
anandthakker: I think we want `{ glyph: ProgramInterface, icon: ProgramInterface, collisionBox: ProgramInterface }` here, so that all three are required.

PR #4934, opened 2017-07-03
Rewrite ProgramConfiguration

### LINKED ISSUE #3954: Support extrusion heights over 65km
Hello,
I'm trying to create a 3D hexbin map, similar to the example you showed a while ago (https://www.mapbox.com/bites/00304/). I'm trying to do this at a country level rather than city level. I tried to raise the `fill-extrusion-height` to an appropriate level, as this is set in meters I'd need something in the region of 300+Km but I guess I'm hitting a limitation and I can really go above ~50Km. Do you think that this could be an error on my behalf or this is indeed a limitation perhaps do to the type of Integer that is used to set the `fill-extrusion-height`?

### LINKED ISSUE #4438: DDS extrusion-heights get Math.floor'ed
chekkit: https://jsfiddle.net/xgs8oyrc/

**mapbox-gl-js version**: v0.33.1

### Steps to Trigger Behavior

 1. Add a fill-extrusion layer
 2. Apply a categorical, DDS styling rule on `fill-extrusion-opacity` (might be other DDS types as well)
 3. For the value, use a fractional height value (e.g. 1.5)

### Expected Behavior

An extrusion of height 1.5

### Actual Behavior

An extrusion of height 1.

From the linked fiddle:

screen shot 2017-03-16 at 1 27 58 pm
_Red extrusion is DDS styled to 1.5. Blue extrusion is explicitly styled to 1.5_

cc/ @lbud

jfirebaugh:
* Expand `#pragma`s to `#ifndef`s early on. This is written in such a way that it can be reused by gl-native's generate-shaders.js. It also avoids generating unnecessary `varying`s (#4542).
* Binder interface for paint properties, with a concrete implementation for each binding strategy.
* Always use floats for paint property attributes, without a multiplication factor. (Fixes #4438. Fixes #3954.)
* When binding composite functions, interpolate between two stop values only: zoom and zoom + 1.

benchmark | master 83fb52d | rewrite-program-configuration 96e5fb8
--- | --- | ---
**map-load** | 99 ms  | 83 ms 
**style-load** | 67 ms  | 60 ms 
**buffer** | 1,002 ms  | 945 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 5.1 ms, 0% > 16ms  | 5.4 ms, 0% > 16ms 
**query-point** | 0.78 ms  | 0.87 ms 
**query-box** | 50.73 ms  | 50.52 ms 
**geojson-setdata-small** | 4 ms  | 4 ms 
**geojson-setdata-large** | 129 ms  | 130 ms 

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/data/program_configuration.js
+++ b/src/data/program_configuration.js
@@ -1,8 +1,8 @@
 // @flow
 
-const assert = require('assert');
 const createVertexArrayType = require('./vertex_array_type');
-const util = require('../util/util');
+const interpolationFactor = require('../style-spec/function').interpolationFactor;
+const packUint8ToFloat = require('../shaders/encode_attribute').packUint8ToFloat;
 
 import type StyleLayer from '../style/style_layer';
 import type {ViewType, StructArray} from '../util/struct_array';
@@ -14,50 +14,9 @@ type LayoutAttribute = {
 }
 
 type PaintAttribute = {
+    property: string,
     name?: string,
-    type: ViewType,
-    property: string,
-    components?: number,
-    multiplier?: number,
-    dimensions?: number,
-    zoomStops?: Array<number>,
     useIntegerZoom?: boolean
-}
-
-type Attribute = {
-    name: string,
-    type: ViewType,
-    property: string,
-    components: number,
-    multiplier: number,
-    dimensions: number,
-    zoomStops: Array<number>,
-    useIntegerZoom: boolean
-}
-
-type Uniform = {
-    name: string,
-    property: string
-}
-
-type InterpolationUniform = {
-    name: string,
-    property: string,
-    stopOffset: number,
-    useIntegerZoom: boolean
-}
-
-type Pragmas = {
-    define: Array<string>,
-    initialize: Array<string>,
-    vertex: {
-        define: Array<string>,
-        initialize: Array<string>,
-    },
-    fragment: {
-        define: Array<string>,
-        initialize: Array<string>,
-    }
 }
 
 export type PaintPropertyStatistics = {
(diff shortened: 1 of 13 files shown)


mourner:
Awesome! Do you know if there's a performance impact from the float type change? Would it cause more data to be uploaded?

src/data/program_configuration.js
```diff
@@ -76,6 +35,162 @@ export type Program = {
     [string]: any
 }
 
+function packColor(color: [number, number, number, number]): Array<number> {
```
anandthakker: I think the return type here should be `[number, number]`.

src/shaders/index.js
```diff
@@ -75,3 +75,63 @@ module.exports = {
...
+
+const re = /#pragma mapbox: ([\w]+) ([\w]+) ([\w]+) ([\w]+)/g;
+
+for (const programName in module.exports) {
+    const program = module.exports[programName];
+    const fragmentPragmas = {};
+
+    program.fragmentSource = program.fragmentSource.replace(re, (match, operation, precision, type, name) => {
+        fragmentPragmas[name] = true;
+        return operation === 'define' ? `
```
anandthakker: I'd favor the more explicit `if (operation === 'define') ... else if (operation === 'initialize')` (or the equivalent `switch`) here and below, so as to (re-)articulate how our custom `#pragma` syntax works.

jfirebaugh:
> Do you know if there's a performance impact from the float type change? Would it cause more data to be uploaded?

Yes, in some cases -- I believe the majority -- the buffers for data-driven properties will be larger. For source functions, non-color data-driven properties that were previously typed as `Uint8` will use 4x space; `Uint16` and color properties, 2x. For composite functions, it depends on how many stops the function uses and the type of the property. Color properties will generally use less space. Functions with four or more stops will generally use less space. Other composite functions will generally use the same amount of space or more space. The space for non-data-driven property values is unchanged.

Overall, I expect the space increase to have a minor impact, and I think it's worth it to simplify the implementation, make gl-js consistent with gl-native, and eliminate bugs like #4438 where using a 16-bit integer results in truncation and limited range.

PR #4938, opened 2017-07-05
Relax lint indent settings

jfirebaugh:
Fixes build by allowing indentation like:

```
    addCurrentVertex(currentVertex: Point,
                     distance: number,
                     normal: Point,
                     endLeft: number,
                     endRight: number,
                     round: boolean,
                     segment: Segment) {
...
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -27,7 +27,15 @@
     "space-before-function-paren": "off",
     "template-curly-spacing": "error",
     "no-useless-escape": "off",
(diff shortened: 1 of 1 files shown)


jfirebaugh:
(The build failures were because #4933 was run in CI prior to #4936, and then rebase-merged.)

mourner:
@jfirebaugh I'd slightly prefer to reindent the affected lines instead; this rule helped me catch a lot of 2-space indents in function parameters, and it's nice to enforce consistency. Not a strong preference though.

PR #4939, opened 2017-07-05
migrate to ci 2.0

### LINKED ISSUE #4905: Switch to Circle CI 2.0
We should switch to Circle CI 2.0 to take advantage of custom docker images and faster build machines. It's likely we can substantially decrease the build time by doing so.

Because Circle CI 2.0 does not yet support macOS, this will mean that we do switch back to testing on linux. We had previously switched to macOS as a workaround for undiagnosed test stability issues. Hopefully those are resolved on the 2.0 infrastructure, or if not we should invest the time to track down and fix them.

mollymerp:
close #4905 

there are a couple failing render tests but the diff is very small (looks like tile boundary is 1 px off), so I may increase the allowable diff for those tests unless someone else has another idea. 

I also still need to test the deploy script with a test tag

cc @jfirebaugh @kkaefer 

## Launch Checklist

 - [ ] post benchmark scores


mollymerp:
Looks like the build is down to about 8 minutes for v2 from 11-12 minutes in v1

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
+version: 2
+
+jobs:
+  build:
+    docker:
+      - image: mbgl/ci:r4-linux-clang-3.9
```
jfirebaugh: Does this image have awscli pre-installed? We should use one that does, so we can remove that from `dependencies.sh`.

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
...
+    working_directory: ~/mapbox-gl-js
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-mapbox-gl-js-cache
+      - run:
+          name: install dependencies
+          command: |
+            bash ./.circleci/dependencies.sh
```
jfirebaugh: Let's move the `yarn install` part of this to its own `run` step, to measure the time independently.

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
...
+            bash ./.circleci/test.sh
+      - store_artifacts:
+          path: "test/integration/render-tests/index.html"
+      - store_artifacts:
+          path: "test/integration/query-tests/index.html"
+      - run:
+          name: deploy if tagged
+          when: on_success
+          command: |
+            if [[ $CIRCLE_TAG =~ v[0-9]+\.[0-9]+\.[0-9]+(\-dev)? ]]; then
```
jfirebaugh: It looks like something is off about this conditional; https://circleci.com/gh/mapbox/mapbox-gl-js/8468 ran `deploy.sh` but it wasn't a tagged build.

mollymerp: I think it was a tagged build unless I'm missing something 🤔  I made (and subsequently deleted) a `v0.0.0-dev` tag to test out the conditional.

screen shot 2017-07-06 at 8 55 14 am

jfirebaugh: Oooh, ok, good idea!

mollymerp:
Switched to a gl-js specific image – PR here https://github.com/mapbox/mbgl-ci-images/pull/8

mollymerp:
blerg I do not know why this test is failing – its passing when I run the render tests inside the docker image locally 💭

mollymerp:
I'm fully perplexed by this [remaining failing test](https://8673-8629417-gh.circle-artifacts.com/0/root/mapbox-gl-js/test/integration/render-tests/index.html) 😞  it was passing in [this build](https://circleci.com/gh/mapbox/mapbox-gl-js/8468) and it passes locally in the docker container and on CI when `yarn run test-render runtime-styling` is run in isolation from the other tests. 

I've tried disabling the cache in `suite_implementation` and clearing the `SpriteAtlas` before the `addImage` tests are run to no avail. 

Anyone have any ideas? Should I ignore the test to get this out? @kkaefer

asheemmamoowala:
@mollymerp I wonder if the issue is related to that specific image file, although I can't posit why that would be. Maybe running with a different image - like `marker.png` or the `1x.png` sprite would show that.
Another thing to try is to use a larger size and set `"debug": true` to see if the entire tile is shifted or just the image is off.

mollymerp:
😂 moving to workflows with the newly added 2.0 tag filter support fixed the failing test 🤷‍♀️ and increasing our `resource_class` to 6GB ram got the builds to pass without failing out. I'm not sure why we need to up the RAM because I got the build to run locally on a 4GB RAM container just fine.

mollymerp:
I ran a test build tagged as v0.0.2-dev [here](https://circleci.com/gh/mapbox/mapbox-gl-js/9013) and the regular non-tagged build [here](https://circleci.com/gh/mapbox/mapbox-gl-js/9012)

mollymerp:
we'll have to un-require the old ci/circleci build in the settings in order to merge this.

PR #4949, opened 2017-07-06
Type style_layer_index.js

jfirebaugh:


src/source/vector_tile_worker_source.js
```diff
@@ -66,7 +66,7 @@ function loadVectorTile(params: WorkerTileParameters, callback: LoadVectorDataCa
  */
 class VectorTileWorkerSource implements WorkerSource {
     actor: Actor;
-    layerIndex: number;
```
anandthakker: funny that we got away with this beofre 😬

PR #4952, opened 2017-07-06
Link to mapbox-gl-language

1ec5:
Link to [Mapbox GL Language](https://github.com/mapbox/mapbox-gl-language/) from the plugins page. Also link to the mapbox-gl-rtl-text example.

After merging, this change can be cherry-picked into the mb-pages branch to go live immediately instead of waiting until the next GL JS release.

/cc @lukasmartinelli


1ec5:
Ah, for some reason I always thought it was the other way around. I’ve rebased and retargeted this PR.

PR #4955, opened 2017-07-07
Type ajax.js and evented.js

jfirebaugh:


src/util/mapbox.js
```diff
@@ -12,7 +12,7 @@ type UrlObject = {|
     params: Array<string>
 |};
 
-function makeAPIURL(urlObject: UrlObject, accessToken): string {
+function makeAPIURL(urlObject: UrlObject, accessToken?: ?string): string {
```
anandthakker: Why both optional _and_ maybe-string?

src/util/evented.js
```diff
@@ -124,11 +132,9 @@ class Evented {
      * Bubble all events fired by this instance of Evented to this parent instance of Evented.
      *
      * @private
-     * @param {parent}
-     * @param {data}
      * @returns {Object} `this`
      */
-    setEventedParent(parent, data) {
+    setEventedParent(parent: ?Evented, data?: Object) {
```
anandthakker: Oh, just saw: should `data` be `Object | () => Object`?

src/util/mapbox.js
jfirebaugh: It needs (and is able) to accept both `null` and `undefined`. It would be clearer to write this as `accessToken: string | null | void` though -- I'll switch.

PR #4971, opened 2017-07-11
Increase significantRotateThreshold, fix #4970

### LINKED ISSUE #4970: TouchRotate is seems too sensitive. Difficult to only PinchZoom the map.
After a lot of testing and feedback from people using our Mapbox map on mobile, everyone says that the rotation is too sensitive. People want to be able to pinch zoom the map without rotating the map unless they really mean to do it.

I have found that the implementation in the TouchZoomRotate handler is fairly good, and the only thing that needs to be fixed is to increase the `significantRotateThreshold` value a bit.

It is currently set to `4` degrees, which is very little. After experimenting, I have found that `10` is a more sensible default value.

Preferably, developers would want to customize this value, as there might be cases where you want to precisely always rotate a map, but I believe in most cases it is better to up this value a bit.

- Google Maps for iOS seem to have a threshold similar to 10.
- Apple Maps for iOS seem to have a low threshold similar to 4 or a bit higher.

However, Apple Maps will automatically rotate the map BACK to north, if the total rotation change after pinchZoom is less than some extra threshold value. So that is an extra piece of logic that keeps the map from rotating away from north when the user don't want to rotate.

dagjomar:
Ref #4970 

From user testing and feedback I have found that the old threshold value of 4 is too low, causing map to rotate when users are trying to simply pinch zoom the map.

Tested with higher values than 10, but found that 10 is a good value.

Changed the `significantRotateThreshold` from `4` to `10`

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/handler/touch_zoom_rotate.js
+++ b/src/ui/handler/touch_zoom_rotate.js
@@ -8,7 +8,7 @@ const inertiaLinearity = 0.15,
     inertiaDeceleration = 12, // scale / s^2
     inertiaMaxSpeed = 2.5, // scale / s
     significantScaleThreshold = 0.15,
-    significantRotateThreshold = 4;
+    significantRotateThreshold = 10;
 
 /**
  * The `TouchZoomRotateHandler` allows the user to zoom and rotate the map by


dagjomar:
Regarding also changing `bearingSnap` value, it sounds like a good idea, as currently, it is difficult to actually perceive that there exist such a value. (As a user, I am trying to make the map snap north, but unable to make it perceivably happen)

However, that is secondary in importance, and requires some more manual user testing before a value can be confidently set. I can also see some potential caveats with a high value for that, and it begs for it to be an customizable property. Or at least possible to have `touchZoomSnapToNorth: false` - or even `0` to turn it off and other values to set it custom.

mourner:
> and it begs for it to be an customizable property

It is already — it's a public `bearingSnap` map option.

dagjomar:
@mourner  Wow, that is great! Sorry, I didn't check if it was already an optional value, I just assumed it wasn't because all the values looked like private consts when I glimsed on the code.

Well, anyway, great!

PR #4984, opened 2017-07-14
Update draw to fix bug with safari

mcwhittemore:
Bumping draw's version to fix a [bug with Safari](https://github.com/mapbox/mapbox-gl-draw/issues/665)


jfirebaugh:
🙇

PR #4987, opened 2017-07-17
fix Map#areTilesLoading from not returning true on final sourcedata event 

### LINKED ISSUE #4975: areTilesLoaded() doesn't return true in some situations
v0.38.0

### Steps to Trigger Behavior
https://jsfiddle.net/ujxun8kp/7/
 1. Go the example, wait for it to fully load
 2. shift-drag a small box and zoom in to a place where there are points.

tile-load-event-bug

### Expected Behavior
The map zooms in, 'data' events fire as requested vector tiles come in.  Eventually map.areTilesLoaded() will return true when called within a 'data' event handler, when the last tile arrives.  Now we can ensure we are querying the complete set of map features in view.

### Actual Behavior
At times this works, but in certain situations (such as when using a box zoom after the initial map load in the fiddle), areTilesLoaded() consistently never returns true within the 'data' event handler when the last tile arrives.  In project code, I've seen this behavior on the initial map load as well but I've only been able to reproduce in a simple example as described.

What's the use case?
Store locator, with lots of stores, published as vector tiles.  A list of locations on the page is populated using features queried from the current map view.  List must be populated after the map loads *and* all tiles have arrived.  The list must be kept up to date with the map view (subsequent pans, zooms, resizes).  Features should not be queried from the map again until all tiles arrive.

Workaround
If I check map.areTilesLoaded() by polling with an interval, it will return true after the last tile 'data' event.  But, I'd rather use a map event to act on that state change as soon as possible.

I'm open to other ideas.  I discussed a similar issue with @mollymerp earlier in the year using older API methods to check map state.

mollymerp:
Fixes #4975 

When tiles are `reloading` they don't fire a `sourcedata` event when they finish loading unless debug collision boxes are enabled (https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/tile.js#L232 🤔 )

this approach changes the `areTilesLoaded` check to accept `reloading` tiles as loaded, but an alternative approach could be removing the conditional in `tile.js` linked above so that a sourcedata event would be fired when all `reloading` tiles switch back to `loaded`. 

cc @ksummerill  
## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/test/unit/source/tile.test.js
+++ b/test/unit/source/tile.test.js
@@ -175,7 +175,7 @@ test('Tile#redoPlacement', (t) => {
         const tile = new Tile(new TileCoord(1, 1, 1));
         tile.loadVectorData(createVectorData(), createPainter());
         t.stub(tile, 'reloadSymbolData').returns(null);
-        const options = util.extend(new Evented(), {
+        const source = util.extend(new Evented(), {
             type: 'vector',
             dispatcher: {
                 send: (name, data, cb) => {
@@ -188,7 +188,7 @@ test('Tile#redoPlacement', (t) => {
             }
         });
 
-        tile.redoPlacement(options);
+        tile.redoPlacement(source);
         tile.placementSource.on('data', ()=>{
             if (tile.state === 'loaded') t.end();
         });
--- a/test/unit/source/tile.test.js
+++ b/test/unit/source/tile.test.js
@@ -11,6 +11,7 @@ const FeatureIndex = require('../../../src/data/feature_index');
 const CollisionTile = require('../../../src/symbol/collision_tile');
 const CollisionBoxArray = require('../../../src/symbol/collision_box');
 const util = require('../../../src/util/util');
+const Evented = require('../../../src/util/evented');
 
 test('querySourceFeatures', (t) => {
     const features = [{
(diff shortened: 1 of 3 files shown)


anandthakker:
> so that a sourcedata event would be fired when all reloading tiles switch back to loaded

I would lean towards this approach, because of the use case where someone does `setData()` and then wants to know when the new data is loaded/rendered.... but that's a pretty mysterious conditional 😬

mollymerp:
yeah – I think that makes sense @anandthakker and it doesn't seem to break anything 🤞 
finding this tough to test though 💭

anandthakker:
@mollymerp On a second look, I'm starting to wonder whether it makes sense to be firing this event via `Tile#_immediateRedoPlacement`.  Do you know why [this line in SourceCache](https://github.com/mapbox/mapbox-gl-js/blob/8567504a9b52e6d12773e73168ef92fe9f6b4bf8/src/source/source_cache.js#L195) doesn't already give us the sourcedata event we need?

mollymerp:
@anandthakker it does fire the event, `areTilesLoaded` won't be `true` yet for tiles where `tile.redoWhenDone=true` because https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/vector_tile_source.js#L135 sets the tile's state to `reloading` and kicks off the async reloading, so it's not `loaded` when the `sourcedata` event fires. We could potentially invoke `_tileLoaded` when a tile finishes reloading though 💭

anandthakker:
Ohhhh, I see 🤦‍♂️ .  In that case, I withdraw my previous comment -- since it's `Tile` that's responsible for invoking the redoPlacement action, it does make sense for it to fire the associated events.

mollymerp:
ready for your 👓 ✅ @anandthakker !

test/unit/source/tile.test.js
```diff
@@ -170,6 +171,29 @@ test('Tile#redoPlacement', (t) => {
         t.end();
     });
 
+    test('reloaded tile fires a data event on completion', (t)=>{
+        const tile = new Tile(new TileCoord(1, 1, 1));
+        tile.loadVectorData(createVectorData(), createPainter());
+        t.stub(tile, 'reloadSymbolData').returns(null);
+        const options = util.extend(new Evented(), {
```
anandthakker: nit: could we call this `source` rather than `options`?

volkanunsal:
This is still happening.

ryanhamley:
@volkanunsal if you're still seeing this happening, can you please open a new issue with a minimal reproduction of the problem so that we can better diagnose it? thanks!

PR #4990, opened 2017-07-17
Draw debug tiles for source with largest maxZoom

### LINKED ISSUE #4830: showTileBoundaries debug option should read from a source whose min/maxzoom include current zoom if available
**mapbox-gl-js version**: 0.38.0

### Steps to Trigger Behavior

 1. Visit [the openmaptiles style page](https://openmaptiles.org/styles/), open the javascript console.
 2. Paste `maps['osm-liberty'].showTileBoundaries = true;` in the console.
 3. Note the red tile boundaries appear in osm-liberty map on that page. If you zoom that map beyond zoom 6 the tile boundaries disappear.

### Expected Behavior

The boundary tiles should not disappear.

### Actual Behavior

The boundary tiles disappear.

While the instructions above reference a page that's using mapbox-gl 0.31.0, I have a private test case that's using 0.38.0 that's exhibiting the same behavior. Unfortunately I can't link to it because it's behind firewalls.

I've opened a ticket on the style maker's github here: lukasmartinelli/osm-liberty#13. Here's a relevant excerpt from that ticket:

I found this out because I've copied this style to a map I'm working on as a proof-of-concept for the use of mapbox-gl for my company. I've noticed this problem is tied to the natural_earth layer in style.json. If I completely remove this layer definition and reload the page, setting `showTileBoundaries = true` does absolutely nothing. If I remove the source declaration for `natural_earth_shaded_relief` after removing the `natural_earth` layer, everything works as expected.

### LINKED ISSUE #5535: overzooming tiles in studio not reporting correct zxy in deubg mode
**mapbox-gl-js version**: latest in mapbox-studio 

studio-zoom

in studio when you turn on debug mode the z/x/y helper doesn't update after z16;  guessing it is reflecting the data that is being displayed, and at z16+ its still displaying data from z16 mvt. which makes sense, but showing a corresponding z## with that x,y coordinate can cause confusing.

dropping ticket here, per @samanpwbb this is a bug in gl-js and not studio 

### Steps to Trigger Behavior

 1. turn on debug->tile boundaries in Mapbox studio 
 2. zoom in past z16+

### Expected Behavior
debug's tile coordinates for x,y at z16+ reflect the z16 tile

### Actual Behavior
label debug tiles with z16 consistently since that's the vt's data that's being show or to update to proper z-x-y coordinates in debug mode

asheemmamoowala:
Closes #4830 
Closes #5535

### DIFF
--- a/src/css/mapbox-gl.css
+++ b/src/css/mapbox-gl.css
@@ -64,11 +64,14 @@
 
 .mapboxgl-ctrl-group {
     border-radius: 4px;
+    overflow: hidden;
+    background: #fff;
+}
+
+.mapboxgl-ctrl-group:not(:empty) {
     -moz-box-shadow: 0 0 2px rgba(0, 0, 0, 0.1);
     -webkit-box-shadow: 0 0 2px rgba(0, 0, 0, 0.1);
     box-shadow: 0 0 0 2px rgba(0, 0, 0, 0.1);
-    overflow: hidden;
-    background: #fff;
 }
 
 .mapboxgl-ctrl-group > button {


mourner:
I did the single source debug intentionally at some point, because it can get very confusing visually when you have multiple sources drawn at the same time (especially if they have different tile sizes etc.). Would it be better to keep the logic but just make sure we use a source that's visible on the current zoom (rather than just the first source)?

asheemmamoowala:
If we have multiple sources with difference tile sizes, this same problem would arise if we pick just one source, some sources might be using smaller tiles, and we may not be drawing those even if they are visible at the current zoom level.

asheemmamoowala:
@mourner - I've changes this to draw tiles for just a single source -> the visible source with the largest `maxZoom`. This ensures that closest used tile boundary will be visible.

src/source/tile_coord.js
```diff
@@ -39,8 +39,12 @@ class TileCoord {
         (this: any).posMatrix = null;
     }
 
-    toString() {
-        return `${this.z}/${this.x}/${this.y}`;
+    toString(sourceMaxZoom: number | void) {
+        if (sourceMaxZoom && sourceMaxZoom < this.z) {
+            return `${sourceMaxZoom}/${this.x}/${this.y} => ${this.z}`;
```
lbud: 🤔 is there any way I can talk you into writing a render test that shows this?

PR #4995, opened 2017-07-19
Handle configured API_URLs that have a path:

durran:
For the case when a user has modified `mapboxgl.config.API_URL` that
includes a path such as http://test.example.com/api.mapbox.com the
path was getting removed when building the full URL as it was being
overwritten by the original path from the request.

This pull request checks if the generation of the `apiUrlObject` has a
path other than `'/'` (which is the default), and if so we need to prepend it to the
original `urlObject` path.

This fixes the case where users using a reverse proxy to mapbox have
filtered on the specific path set in their API_URL, and when it has been
removed the lack of a match causes the proxy to not forward the requests
.

Debug page:

screen shot 2017-07-19 at 9 34 43 pm

Benchmarks:

```shell
benchmark | master 8567504 | api-url-with-path d0feffa
--- | --- | ---
**map-load** | 298 ms  | 304 ms 
**style-load** | 1,048 ms  | 95 ms 
**buffer** | 1,173 ms  | 1,163 ms 
**fps** | 57 fps  | 60 fps 
**frame-duration** | 8.1 ms, 2% > 16ms  | 7.9 ms, 1% > 16ms 
**query-point** | 1.15 ms  | 1.24 ms 
**query-box** | 86.21 ms  | 95.22 ms 
**geojson-setdata-large** | 170 ms  | 167 ms 

```

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -17,6 +17,10 @@ function makeAPIURL(urlObject: UrlObject, accessToken: string | null | void): st
     urlObject.protocol = apiUrlObject.protocol;
     urlObject.authority = apiUrlObject.authority;
 
+    if (apiUrlObject.path !== '/') {
+        urlObject.authority = `${urlObject.authority}${apiUrlObject.path}`;
+    }
+
     if (!config.REQUIRE_ACCESS_TOKEN) return formatUrl(urlObject);
 
     accessToken = accessToken || config.ACCESS_TOKEN;
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -18,7 +18,7 @@ function makeAPIURL(urlObject: UrlObject, accessToken: string | null | void): st
     urlObject.authority = apiUrlObject.authority;
 
     if (apiUrlObject.path !== '/') {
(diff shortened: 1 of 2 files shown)


src/util/mapbox.js
```diff
@@ -17,6 +17,10 @@ function makeAPIURL(urlObject: UrlObject, accessToken: string | null | void): st
     urlObject.protocol = apiUrlObject.protocol;
     urlObject.authority = apiUrlObject.authority;
 
+    if (apiUrlObject.path !== '/') {
+        urlObject.authority = `${urlObject.authority}${apiUrlObject.path}`;
```
jfirebaugh: Could this prepend `apiUrlObject.path` to `urlObject.path`, rather than appending it to `authority`? That would be clearer and more robust.

durran: More than happy to make that change.

durran: @jfirebaugh Changes made - we now prepend to the `urlObject.path` instead of appending to `urlObject.authority`.

durran:
Thanks!

PR #4997, opened 2017-07-19
Implement DDS for 'text-anchor' property.

ChrisLoer:
@nickidlugash and I started work on implementing some of the less-difficult/high-priority DDS features needed for core styles (https://github.com/mapbox/mapbox-gl-js/issues/4434). This PR implements `text-anchor` as a zoom-and-property function.

This one seems pretty simple. What are we missing?

/cc @mollymerp


nickidlugash:
As discussed, there seems to be a bug in identity functions where strings in sources are not converted to enums, so going to merge this with the test as is.

ChrisLoer:
The identity function bug is: https://github.com/mapbox/mapbox-gl-js/issues/4999

Issue #2412, opened 2016-04-08
Support property aggregation on clustered features


vibze:
Hello, I've started exploring this lib and have a feature request:
Provide reduce function(s) to clustering configuration, so it's possible to filter clusters by aggregated properties rather than just `point_count`.

mourner:
Interesting feature. Can you describe your use case in detail? How should it aggregate the properties? E.g. suppose we cluster a million points where each has different property keys and values — how should the top cluster properties look like then?

vibze:
User must provide an aggregation function, so in a case of points belonging to a cluster have different property keys and values or value types everything should be handled in that function. Mapbox-gl just applies it to the dataset.

Well, let's say I want to build a population heatmap using population data for every human settlement in area.

Feature example:
```javasript
{
    "type": "Feature", 
    "properties": { 
        "osm_id": "26544289", 
        "name": "Алматы", 
        "type": "city", 
        "population": 1475400 
    }, 
    "geometry": { 
        "type": "Point", 
        "coordinates": [ 76.9458522, 43.2355947 ] 
    } 
}
```

Now, I want every cluster to have a sum population of all included points as a property. I configure the source like this:
```javascript
map.addSource("points", { 
    "type": "geojson", 
    "data": "data/points.geojson",
    "cluster": true, 
    "clusterMaxZoom": 14, 
    "clusterRadius": 20,
    "clusterAggregations": {
        "population_sum": function(prevValue, feature, i, arr) {
            if (!('population' in feature) || isNaN(parseFloat(feature.population))) {
                return prevValue;  // handle the case of missing keys or wrong type
            }

            return prevValue + parseFloat(feature.population);
        }
    }
});
```
... and every cluster from this source receives the `population_sum` property to work with.

mourner:
I believe we won't be able to accept a function, since the mapbox style specification only allows valid JSON. Would it be enough to have something like this?

```
clusterAggregate: ["population"]
```

Then it would just do a sum for all given properties.

vibze:
It would be enough for certain cases, but definitely not as flexible. Do you think something like this http://bl.ocks.org/gisminister/10001728 could be possible with Mapbox GL API?

lavishmantri:
Can also have some common mathematical functions
`clusterAggregateFunction: "SUM"` or `clusterAggregateFunction: "AVG"`?

That would be helpful.

ryanbaumann:
[...]

PR #5002, opened 2017-07-20
Increase minimum supercluster version

johnlaur:
Please consider increasing the supercluster version to ^2.3.0. This version of supercluser populates a cluster_id property of all clusters that it generates. This property is a useful cross reference when using a second instance of the supercluster object to aggregate properties or explore cluster members. Such functionality appears to be planned for integration into mapbox-gl-js per discussion in #2412 at which point the newer supercluster release would become a requirement.

In any case, supercluster 2.0.1 lacks the large performance improvement in version 2.1.0, and the release version of mapbox-gl-js is built against 2.2.0. So this ought to be bumped for other reasons as well.

👀 @jfirebaugh

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "pbf": "^1.3.2",
     "point-geometry": "^0.0.0",
     "quickselect": "^1.0.0",
-    "supercluster": "^2.0.1",
+    "supercluster": "^2.3.0",
     "through2": "^2.0.3",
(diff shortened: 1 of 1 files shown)


mourner:
Thank you!

andrewharvey:
Should this also bump yarn.lock? https://github.com/mapbox/mapbox-gl-js/blob/master/yarn.lock#L6077

PR #5011, opened 2017-07-20
Fix bad if statement in vector_tile_work_souce

### LINKED ISSUE #5005: Code bug in vector_tile_source
There is a bug that might lead to worker number 0 not being utilized properly.
```javascript
if (!tile.workerID || tile.state === 'expired') {
   tile.workerID = this.dispatcher.send('loadTile', params, done.bind(this));
} else if (tile.state === 'loading') {
   // schedule tile reloading after it has been loaded
   tile.reloadCallback = callback;
} else {
   this.dispatcher.send('reloadTile', params, done.bind(this), tile.workerID);
}
```
If the worker is assigned index 0, it will never be re-utilized for reloading. I believe the tile will constantly be loaded instead of reloaded if it was being handled by worker 0.
Correct if statement should be `if (tile.workerID !== undefined && tile.workerID !== null ... )`

sanjin-saric:
Closes #5005

### DIFF
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -187,7 +187,7 @@ class GeoJSONSource extends Evented implements Source {
     }
 
     loadTile(tile: Tile, callback: Callback<void>) {
-        const message = !tile.workerID || tile.state === 'expired' ? 'loadTile' : 'reloadTile';
+        const message = tile.workerID === undefined || tile.state === 'expired' ? 'loadTile' : 'reloadTile';
         const params = {
             type: this.type,
             uid: tile.uid,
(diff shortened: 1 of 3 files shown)


mollymerp:
great! thank you!

I think we also need to change the condition here https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/geojson_source.js#L188

would you be able to add some unit tests that check for this error?

sanjin-saric:
Sure. Will do.

mollymerp:
@sanjin-saric do you think you'll have time to add the requested changes and rebase this PR? If not I'd be happy to wrap it up for you. 

Thanks!

sanjin-saric:
i was on vacation and I couldn't find time to finish this up. I'll try to wrap this up within a week.

sanjin-saric:
I fixed GeoJSON source but I do not know how to write test for this. I cannot write a test that tests the validity of code.

Issue #4917, opened 2017-06-29
suggestion: resource Loader with a callback mechanism to provide locally served files. 


Yermo:
## Motivation

In my use case, I need to be able to render a map on a device in a completely offline context. As such, json styles, mbtiles, etc all have to be served from localstorage on the phone.

mapbox-gl-js calls ajax methods directly whenever it needs to pull a resource making the assumption the resource will be remotely hosted.

I noticed the semi-complete "addSourceType" features and read the discussion. I don't want to create a new source type. I merely want to pull the 'vector' source from the local phone. 

This may be related to issue #4740 Resource Transforms.

## Design Alternatives

While adding new source types to address this is an option, abstracting the loading of resources would avoid some code duplication and be much simpler.

I just want to be able to:

1. host the json style file locally and refer to it using a "file:///" url in map constructor options.
2. reference all resources specified in the JSON files using a file:/// prefix. 
3. host a small mbtiles file locally, also specified via a file:/// url.

## Design

As a proof of concept, I have implemented a "src/util/resourceLoader,js" shim that mirrors the interface in src/util/ajax.js and replaced all ajax calls in the code with the corresponding calls to resourceLoader.

(eg. ajax.getJSON() becomes resourceLoader.getJSON()).

There is a map level configuration option for a user provided object that expects two methods: getJSON() and getArrayBuffer(). In my case I'm using Cordova File and Sqlite plugins.

In src/utils/resourceLoader I check the URL. If it's a remote URL I pass on the request to the ajax handler, business as usual. If the URL is not prefixed by http:// https:// or mapbox://, I forward the request to the caller supplied resourceLoader.

A complication is that Cordova plugins can only be called from the main thread so I've modified the interface to LoadVectorData to include the actor so I can send a message to the main thread. I've modified style.js to receive those calls and forward them to the user provided loader. 

Doing it this way would provide a great deal of flexibility for developers to host offline maps or pull data from various sources without having to maintain separate source implementations.

### Implementation

Since this is just a proof of concept and needs some cleanup work, I didn't want to send a pull request yet but I have put up a fork with this work. 

https://github.com/Yermo/mapbox-gl-js/tree/resource-loader

At present, there is an ugly hack in map.js to make the reference for the user supplied loader available in src/util/resourceLoader.js.

Ideally, I'd like to get something like this merged into the code. With a little guidance, I am happy to improve upon this work and package it up for a pull request.

anandthakker:
[...]

Issue #3874, opened 2016-12-30
Allow Cross-site Source Data Request with Credentials


jasonpepper:
## Motivation

What use cases are we trying to accommodate? 

Source data may contain private data that requires authentication. Furthermore, the data may be hosted at a different http origin than the map application. Currently, cross-site MapBox ajax requests will not pass authentication tokens.

## Design Alternatives

Add a boolean option to the various types of sources called "withCredentials", where if it is true the XMLHttpRequest will have the withCredentials option set to true. For example:

"mapbox-streets": {
  "type": "vector",
  "url": "mapbox://mapbox.mapbox-streets-v6",
  "withCredentials": true
}

### Concepts

How will we teach this design?

In the style spec, add a description of the new option.

### Implementation

How you would implement the design in Javascript?

    var xhr = new window.XMLHttpRequest();
    xhr.open('GET', url, true);
    xhr.responseType = 'arraybuffer';
    xhr.withCredentials = withCredentials;

lucaswoj:
For the same reasons as https://github.com/mapbox/mapbox-gl-js/pull/2918, I'm not sure this feature fits the TileJSON ethos. This may need to depend on #3326.

huaruiwu:
@lucaswoj those two issues look closed. Any plans on allowing credentials?

mollymerp:
this feature is not on our current roadmap @huaruiwu.

huaruiwu:
Without this feature, it isn't possible to have secure map data. Signing URLs have too much of an overhead. I think @x9xjdzz9 's proposed design makes sense.

rayterrill:
We're looking to swing json and image hosting over to Cloudfront using Signed Cookies, and after struggling to track down what was happening, it appears we're running into this exact same issue. :(

asheemmamoowala:
Closed in #5021

Issue #3123, opened 2016-09-01
Allow custom done callback for RasterTileSource.loadTile


timiyay:
I have a use case that requires me to make a custom `SourceType`, for which I was hoping to inherit from `RasterTileSource`.

I'm integrating an external WMS into our `mapbox-gl-js` project, as they provide high-quality aerial imagery for my area of interest.

The issue is that the WMS will always return a `200`. When tiles are missing for high zoom levels, the WMS return a `200` and a blank white tile.

I need a custom `SourceType` so that I can programatically-detect these blank tiles, and prevent them being rendered.

I realise there are current issues and PRs looking at custom `SourceType`s. Nonetheless, for my use case, I would be greatly aided if `RasterTileSource.loadTile` was refactored into separate methods: https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/raster_tile_source.js#L49

If I was able to override the `done()` function and then, on success, pass my `img` object to a separate function that called the GL rendering code from line 59 (https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/raster_tile_source.js#L59), then it'd be an easy switch.

As it stands, I'll have to cut'n'paste all of code from `loadTile` into my own version of the function.

Would there be any support for this change within the master `mapbox-gl-js` repo, or do you see this use case as being better-solved by existing/upcoming capabilities of  custom `SourceType`s?

I'm happy to whip up a PR, if there's support for this change.

anandthakker:
@timiyay thanks for raising this.  Ergonomic extensibility of `RasterTileSource` is definitely something that wasn't really addressed in the current custom source type setup.

I'm thinking that a possible variation to your proposal might be to extend the `RasterSource` constructor to accept a fourth, optional `loadTileImage` function with a signature something like `(TileCoordinate, callback) => void`, which, if provided, is responsible for fetching the raw image data (and i guess maybe the content type?) for a given tile and calling the callback with it.

Then a slightly reworked version of [these lines](https://github.com/mapbox/mapbox-gl-js/blob/master/js/source/raster_tile_source.js#L45-L47) would essentially be the default implementation of loadTileImage.

Once this was in place, then implementing a custom raster source could look like:

```js
var RasterSource = require('mapbox-gl/js/source/raster_source')
function MyRasterSource (id, options, dispatcher) {
   var rasterSource = new RasterSource(id, options, dispatcher, myLoadTileImage);
   // delegate Source interface methods to the RasterSource instance:
[...]

PR #5021, opened 2017-07-21
URL Transform callback as a Map option

asheemmamoowala:
We have quite a few tickets (#4740, #2918, #4917, and [gl-native/#7455](https://github.com/mapbox/mapbox-gl-native/issues/7455)) related to modifying XHR requests made by mapbox-gl-js for various resources. While these tickets are each slightly different, they all include an element of inspecting and modifying the XHR request before it is sent.  
The common use cases break down to:
- sending authentication headers/cookies with cross-origin requests
- adding custom headers to requests
- altering URLs dynamically

This PR addresses all these concerns by providing a per-map option for a `transformRequest` callback.

### Example usage
```
let Map = new Map({    
  transformRequest: (url, resourceType) => { 
    if (url.startsWith('http://myHost') {
      return {
        url: url.replace('http:', 'https:'),
        headers: { 'my-custom-header': true},
        withCredentials: true
    }
});
```
----

benchmark | master 93a9922 | 4740-per-map-transform 0933fb7
--- | --- | ---
**map-load** | 160 ms  | 103 ms 
**style-load** | 91 ms  | 80 ms 
**buffer** | 937 ms  | 905 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 6 ms, 0% > 16ms  | 6 ms, 0% > 16ms 
**query-point** | 0.95 ms  | 0.98 ms 
**query-box** | 52.74 ms  | 53.28 ms 
**geojson-setdata-small** | 5 ms  | 4 ms 
**geojson-setdata-large** | 134 ms  | 131 ms 

### Checklist
 - [x] Allow transforming URLs before they are requested #4740 
 - [X] Allow adding custom headers (Similar to #2918, partial of #3123)
 - [x] Allow sending request with cookie/credentials #3874 
 - [x] post benchmark scores

// **cc** @anandthakker @jfirebaugh @kkaefer

### DIFF
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -4,6 +4,7 @@ const Evented = require('../util/evented');
 const util = require('../util/util');
 const window = require('../util/window');
 const EXTENT = require('../data/extent');
+const ResourceType = require('../util/ajax').ResourceType;
 
 import type {Source} from './source';
 import type Map from '../ui/map';
@@ -138,8 +139,8 @@ class GeoJSONSource extends Evented implements Source {
     }
 
     onAdd(map: Map) {
-        this.load();
         this.map = map;
+        this.load();
     }
 
     /**
@@ -170,7 +171,7 @@ class GeoJSONSource extends Evented implements Source {
         const options = util.extend({}, this.workerOptions);
         const data = this._data;
         if (typeof data === 'string') {
-            options.url = resolveURL(data);
+            options.request = this.map._transformRequest(resolveURL(data), ResourceType.Source);
         } else {
             options.data = JSON.stringify(data);
         }
--- a/src/style/image_sprite.js
+++ b/src/style/image_sprite.js
@@ -5,6 +5,8 @@ const ajax = require('../util/ajax');
 const browser = require('../util/browser');
 const normalizeURL = require('../util/mapbox').normalizeSpriteURL;
 
+import type {RequestTransformFunction} from '../ui/map';
+
 class SpritePosition {
     x: number;
     y: number;
@@ -27,19 +29,22 @@ class ImageSprite extends Evented {
     base: string;
     retina: boolean;
 
+    transformRequestFn: RequestTransformFunction;
     data: ?{[string]: SpritePosition};
     imgData: ?Uint8ClampedArray;
     width: ?number;
 
-    constructor(base: string, eventedParent?: Evented) {
+    constructor(base: string, transformRequestCallback: RequestTransformFunction, eventedParent?: Evented) {
         super();
         this.base = base;
         this.retina = browser.devicePixelRatio > 1;
         this.setEventedParent(eventedParent);
+        this.transformRequestFn = transformRequestCallback;
 
         const format = this.retina ? '@2x' : '';
-
-        ajax.getJSON(normalizeURL(base, format, '.json'), (err, data) => {
+        let url = normalizeURL(base, format, '.json');
+        const jsonRequest = transformRequestCallback(url, ajax.ResourceType.SpriteJSON);
+        ajax.getJSON(jsonRequest, (err, data) => {
             if (err) {
                 this.fire('error', {error: err});
             } else if (data) {
@@ -47,8 +52,9 @@ class ImageSprite extends Evented {
                 if (this.imgData) this.fire('data', {dataType: 'style'});
             }
         });
-
-        ajax.getImage(normalizeURL(base, format, '.png'), (err, img) => {
+        url = normalizeURL(base, format, '.png');
+        const imageRequest = transformRequestCallback(url, ajax.ResourceType.SpriteImage);
+        ajax.getImage(imageRequest, (err, img) => {
             if (err) {
                 this.fire('error', {error: err});
             } else if (img) {
@@ -71,7 +77,7 @@ class ImageSprite extends Evented {
 
     resize(/*gl*/) {
         if (browser.devicePixelRatio > 1 !== this.retina) {
-            const newSprite = new ImageSprite(this.base);
+            const newSprite = new ImageSprite(this.base, this.transformRequestFn);
             newSprite.on('data', () => {
                 this.data = newSprite.data;
                 this.imgData = newSprite.imgData;
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -25,6 +25,7 @@ const isSupported = require('mapbox-gl-supported');
 
 import type {LngLatLike} from '../geo/lng_lat';
 import type {LngLatBoundsLike} from '../geo/lng_lat_bounds';
+import type {RequestParameters} from '../util/ajax';
 
 /* eslint-disable no-use-before-define */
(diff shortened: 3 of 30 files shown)


src/source/geojson_source.js
```diff
@@ -2,8 +2,8 @@
 
 const Evented = require('../util/evented');
 const util = require('../util/util');
-const window = require('../util/window');
 const EXTENT = require('../data/extent');
+const resolveURL = require('../util/mapbox').resolveURL;
```
asheemmamoowala: @jfirebaugh `util.js` already depends on `window`, so putting `resolveURL` in `util.js` creates a cyclic dependency. `resolveURL` could be moved to `ajax.js`, or left in `geojson_source.js` which was its only usage anyways

src/ui/map.js
```diff
@@ -599,6 +608,49 @@ class Map extends Camera {
...
+     * Map.setRequestTransform( (url, resourceType)=> {
+     *  if(resourceType == 'Source' && url.startsWith('http://myHost') {
+     *   return {
+     *    url: url.replace('http', 'https'),
+     *    headers: { 'my-custom-header': true},
+     *    withCredentials: true
+     *  }
+     * });
+     */
+    setRequestTransform(transform: Function) {
```
asheemmamoowala: Thanks for this suggestion. I didn't even consider the timing and order-of-execution issue with regards to `Map#setRequestTransform.` 
`reousrceType` is optional for catching calls that don't provide a type.

Yermo:
FWIW, the addition of this feature solves a big problem for me as I need to send an Authorization header with my requests. (I was on my way to implementing something like this, so thanks!)

src/util/ajax.js
```diff
@@ -2,6 +2,38 @@
...
+ * A `RequestParameters` object to be returned from Map.options.transformRequest callbacks.
+ * @typedef {Object} RequestParameters
+ * @property {string} url The URL to be requested.
+ * @property {Object} headers The headers to be sent with the request.
+ * @property {boolean} withCredentials Whether cross-site requests should be made with cookies.
+ */
+export type RequestParameters = {
+    url: string,
+    headers?: Object,
+    withCredentials? : Boolean
```
jfirebaugh: Let's make this a bit more extensible, and consistent with the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch) (which we may eventually switch to internally):

```
    credentials?: 'same-origin' | 'include'
```

(It doesn't look like it's possible to emulate Fetch's `'omit'` value using XHR, so I left that out.)

src/source/geojson_source.js
```diff
@@ -170,7 +170,7 @@ class GeoJSONSource extends Evented implements Source {
         const options = util.extend({}, this.workerOptions);
         const data = this._data;
         if (typeof data === 'string') {
-            options.url = resolveURL(data);
+            options.request = this.map._transformRequest(resolveURL(data));
```
jfirebaugh: Is this the only place where the type isn't provided? Native [uses the `Source` type here](https://github.com/mapbox/mapbox-gl-native/blob/175e7e26e2008d1e55664bd1064536c23b6de86a/src/mbgl/style/sources/geojson_source.cpp#L53); let's do the same and then make this parameter required.

src/style/image_sprite.js
```diff
@@ -31,24 +31,26 @@ class ImageSprite extends Evented {
     imgData: ?HTMLImageElement;
     width: ?number;
 
-    constructor(base: string, eventedParent?: Evented) {
+    constructor(base: string, eventedParent?: Evented, transformRequestCallback?: Function) {
```
jfirebaugh: Can the added parameter be required (in second position)?

asheemmamoowala:
Fixes #4740, #3874

rayterrill:
Any idea when this will be released? This looks like it'll fix some of the CloudFront Signed Cookies issues we're currently experiencing.

kkaefer:
@rayterrill It'll be part of the next GL JS release. We're typically releasing new versions ~6 weeks. [Mapbox GL JS v0.39](https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.39.0) was released on July 21, so the next release will likely be end of August/beginning of September.

rayterrill:
@kkaefer Thanks mate!

volkergraf:
Just a Question .. the Transformation of the URL seems to work ok .. but now I'm getting a "Error: Unimplemented type: 3" 

The Code looks like:


```
transformRequest: (url, resourceType) => {
              console.error(resourceType);
              alert("url: "+ url+"->"+ a_urls_to_blobs[url],);
              return {
                url: a_urls_to_blobs[url],
                headers: { "content-encoding" :"gzip"},
              }
            }
```

Do I have to set any type or something like that ?

asheemmamoowala:
@volkergraf  can you provide more context on the error you are seeing ? A reproducible case with jsfiddle would be helpful in debugging.

kkaefer:
It sounds like the response is zipped but not decompressed correctly. Why are you setting `"content-encoding": "gzip"` for the request header? GET requests like this one don't have a body, so it doesn't make sense to specify a content encoding. Browsers automatically specify an Accept-Encoding that includes `gzip`, so there's no header that you need to set to benefit from transport compression.

volkergraf:
@kkaefer Well this was only a test, trying to get rid of the "Error: Unimplemented type: 3" which did not work =)

volkergraf:
@asheemmamoowala Ok .. well the Scenario is the following: 

We are developing an iOS Application using Apache-Cordova. 

WebKit has a Problem concerning AJAX-Calls using **file://..-**Urls .. but there is a workaround, using the **cordova-file-plugin** and the **WXWebKitView** (with a modified Initializer), allowing you to kinda use **file://...**-Urls  to access the deployd files (that will reside in www/foo/bar), but you can NOT use the MapBox-GL.js source URL like "...www/foo/bar{z}/{x]/{y}.pbf, you have to transform it to a blob:file://var......11-22-33-66-767-88/1122.pbf blob-URL.

Therefore I created an associative-Array, containing the Original-URLs that the mapbox-gl.js-map requests and it's corresponding blob:-URLs .. _**a_urls_to_blobs[url]**_

**I can definitively confirm that the mapbox-gl.js CAN load the Tiles trough the blob-URLs** .. but it seems that using **transformRequest:** might require a few more Parameters to return .. to make the whole Thing run .. at the Moment I am returning simply the "blob:"-URLs .. perhapes I might have to add a type or something like that .. any Ideas ?

kkaefer:
`Error: Unimplemented type: 3` means that PBF decoding failed. It likely fails because the file wsa not decompressed correctly because the headers aren't set. Mapbox GL itself doesn't decompress anything; it relies on the browser infrastructure to do so.

rayterrill:
I integrated this and implemented the transformRequest credentials functionality today using CloudFront signed cookies - Everything appears to be working great! Thanks all!

Issue #4942, opened 2017-07-06
Icon and text symbols disappearing on a specific text/icon-size


edpop:
Hello.
There is a symbol layer unexpected disappearing on a specific text/icon-size zoom function. You may see it also in Mapbox Studio.

**mapbox-gl-js version**: v0.38.0

### Steps to Trigger Behavior

 1. Go to https://jsbin.com/nacagabewe/1/edit?html,output
 2. Look at 17.9 zoom
 3. Look at 18.0 zoom
 4. Look at 18.1 zoom

### Expected Behavior
The layer should be displayed on 18.0 zoom

### Actual Behavior
The layer is not displayed on 18.0 zoom

jfirebaugh:
Thanks for the report and the example. Looks like this stopped working sometime between v0.34.0 and v0.35.0. #4455 looks like a likely candidate.

gilazr:
As a temporary workaround, we added 0.1 to the stops zoom levels. Hope this helps

PR #5028, opened 2017-07-24
Fix zoom-function

edpop:
Good day.
Fix a case when interpolationFactor returned NaN (:heart: javascript)
Fixes https://github.com/mapbox/mapbox-gl-js/issues/4942


edpop:
Done. The icons look good.
I don't know, what better to do with this test https://github.com/mapbox/mapbox-gl-js/blob/v0.39.0/test/unit/style-spec/function.test.js#L59.

jfirebaugh:
You fixed the issue with NaN, so I believe the assertion should be changed to `t.equal(f(17), 11)` and the comment should be deleted.

PR #5035, opened 2017-07-24
Copy style-spec package.json dependencies to root

jfirebaugh:
npm does not support subdirectories as dependencies in published packages.

Reroll of #5034 because flow can't cope with a branch named `package.json`.


anandthakker:
> Reroll of #5034 because flow can't cope with a branch named package.json

lol

Issue #2929, opened 2016-08-02
Support fractional minzoom values


vicapow:
minzoom and maxzoom on layers don't seem to respect floating point values and instead round up.

Here's a quick demo: https://jsfiddle.net/oopkjeak/16/

If this is expected, maybe the docs should mention this and style validators should check for integers?

Otherwise, it would be awesome to support floating point values!

shortsleeves:
FWIW - The style spec seems to indicate floats should be valid: [#layer-minzoom](https://www.mapbox.com/mapbox-gl-style-spec/#layer-minzoom)

lucaswoj:
This is not but could be supported 😄

lucaswoj:
Let me know if there is anything I can to do help you put together a PR @vicapow. 🙇

mollymerp:
fwiw `maxzoom` seems to respect float values https://jsfiddle.net/cmozpyyk/

vicapow:
@lucaswoj I spent some time digging but wasn't able to find where this is happening. Any pointers on where to look? really odd that this happens for min but not max.

lucaswoj:
My theory is that the library that creates vector tiles, `geojson-vt`, is generating tiles from `floor(minzoom)` to `floor(maxzoom)`. We can overzoom to render tiles > maxzoom but we cannot "underzoom" to render tiles < minzoom. The fix may be to explicitly `ceil(minzoom)` before passing it to `geojson-vt`

asheemmamoowala:
Close in #5039

PR #5039, opened 2017-07-25
Support fractional minzoom on layers

asheemmamoowala:
Addresses : #2929 

GeoJSON and Vector Tile sources are using a floor'ed zoom value when loading new tiles. This blocks tile data loaded for layers with fractional minzoom as it doesn't pass the test in [worker_tile.js](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/worker_tile.js#L97)

The solution requires two changes
- [x] Pass the current zoom level to the worker tile
- [x] Worker tile should parse tiles that are at the integer min zoom floor  

benchmark | master b528524 | 2929-fractional-minzoom 6d1c21b
--- | --- | ---
**map-load** | 180 ms  | 113 ms 
**style-load** | 107 ms  | 88 ms 
**buffer** | 1,057 ms  | 926 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 6.1 ms, 0% > 16ms  | 6.3 ms, 1% > 16ms 
**query-point** | 0.94 ms  | 0.98 ms 
**query-box** | 53.86 ms  | 49.46 ms 
**geojson-setdata-small** | 4 ms  | 3 ms 
**geojson-setdata-large** | 145 ms  | 137 ms 

## Checklist
 - [x] post benchmark scores


src/source/geojson_source.js
```diff
@@ -190,7 +190,7 @@ class GeoJSONSource extends Evented implements Source {
             type: this.type,
             uid: tile.uid,
             coord: tile.coord,
-            zoom: tile.coord.z,
+            zoom: this.map.transform.zoom,
```
anandthakker: @asheemmamoowala I'd love to avoid another point of dependency of sources on the `map` object (see also https://github.com/mapbox/mapbox-gl-js/issues/3350).

Would an alternative to using the current map's zoom level be to keep using `tile.coord.z`, but change the test to something like `if (this.zoom < floor(layer.minzoom)) continue;`?

If I'm understanding this right, we want:

layer minzoom | tile zoom | tile needed?
---------------|----------|-------------
4|3 (and below)|no
4|4 (and above)|yes
4.5|4 (and above)|yes
4.5|3 (and below)|no

test/integration/render-tests/geojson/fractional-minzoom/style.json
```diff
@@ -0,0 +1,32 @@
+{
+  "version": 8,
+  "metadata": {
+    "test": {
+      "width": 512,
+      "height": 512
+    }
```
anandthakker: Could we simplify this test a tad by:
 - width & height: 64
 - center 0,0
 - data: literal geojson point

test/integration/render-tests/geojson/fractional-minzoom/style.json
```diff
@@ -0,0 +1,38 @@
+{
```
anandthakker: Shoot, I should have thought of this before, but: could we move this test to `render-tests/regressions/mapbox-gl-js#2929`?

test/integration/render-tests/geojson/fractional-minzoom/style.json
```diff
@@ -0,0 +1,38 @@
...
+    {
+      "id": "line",
+      "type": "circle",
+      "source": "geojson",
+      "minzoom": 4.3,
+      "paint": {
+        "circle-color": "#EC8D8D",
+        "circle-radius": 4
+      }
+    }
```
anandthakker: Would it make sense to have two more layers with `minzoom` 4.5 and 5, that we would expect _not_ to appear?  (could use `circle-translate` so each layer's dot would be in a distinct location)

asheemmamoowala: That's a good improvement to the test, Thanks !

kkaefer:
This needs implementation in Native.

PR #5047, opened 2017-07-25
Type glyph and sprite atlases

jfirebaugh:


src/symbol/sprite_atlas.js
```diff
@@ -205,30 +251,26 @@ class SpriteAtlas extends Evented {
         this.dirty = true;
     }
 
-    setSprite(sprite) {
-        if (sprite && this.canvas) {
-            this.canvas.width = this.width;
-            this.canvas.height = this.height;
-        }
```
anandthakker: ^ this is obsolete?

src/symbol/glyph_source.js
```diff
@@ -89,40 +105,39 @@ class GlyphSource extends Evented {
...
                 this.fire('error', { error: err });
-                return;
-            }
+            } else if (typeof range === 'number' && data) {
+                const stack = this.stacks[fontstack].ranges[range] = data.stacks[0];
+                for (let i = 0; i < missingRanges[range].length; i++) {
+                    const glyphID = missingRanges[range][i];
+                    const glyph = stack.glyphs[glyphID];
+                    const rect  = atlas.addGlyph(uid, fontstack, glyph, buffer);
+                    if (glyph && rect) glyphs[glyphID] = new SimpleGlyph(glyph, rect, buffer);
```
anandthakker: Checking for `rect` is causing some rendering tests to fail due to a missing space between words -- looks like maybe we rely on `rect` being optional in the case of a space character?  (I don't know enough about how we're managing glyph data to know what makes the most sense here.)

src/symbol/sprite_atlas.js
jfirebaugh: Yeah, `this.canvas` is never set. I think it was related to some debug code that got removed.

src/symbol/glyph_source.js
jfirebaugh: Thanks for pinpointing that. Fixed.

PR #5051, opened 2017-07-26
Type style.js

jfirebaugh:


src/style/style.js
```diff
@@ -85,65 +127,78 @@ class Style extends Evented {
             }
         });
 
-        const stylesheetLoaded = (err, stylesheet) => {
+        const stylesheetLoaded = (err, stylesheet: ?StyleSpecification) => {
```
anandthakker: 🤔 should we typed this as `mixed` and then cast it to `StyleSpecification` after it passes validation?

src/style/style.js
```diff
@@ -46,12 +53,47 @@ const ignoredDiffOperations = util.pick(diff.operations, [
...
+    sprite: ImageSprite;
+    spriteAtlas: SpriteAtlas;
+    lineAtlas: LineAtlas;
+    glyphSource: GlyphSource;
+    light: Light;
+
+    _layers: {[string]: StyleLayer};
+    _order: Array<string>;
+    sourceCaches: {[string]: SourceCache};
+    zoomHistory: ZoomHistory | {};
```
anandthakker: Is this typed as `ZoomHistory | {}` rather than `?ZoomHistory` to reduce invasiveness of the change?  ( 👍 if so, just making sure )

jfirebaugh: Yes, exactly. `?ZoomHistory` would be preferable but requires some refactoring.

src/style/style.js
jfirebaugh: https://github.com/mapbox/mapbox-gl-js/issues/5064

Issue #2395, opened 2016-04-05
Fix bug caused by interactions between "Popup#remove" and "closeOnClick" 


timiyay:
**Codepen repro**: http://codepen.io/timiyay/pen/LNOWrr
**mapbox-gl-js version**: 0.16.0 (on `earcut-final` branch)
### Steps to Trigger Behavior
1. create an instance of `mapboxgl.Popup`, with options `{ closeButton: false, closeOnClick: true }`
2. create a click listener that calls `queryRenderedFeatures`, and displays any popups under the clicked point, or calls `popup.remove()` if no features found
3. use the previously-created instance of `mapboxgl.Popup` to render the popup
### Expected Behavior

Clicking on a feature once: should render the popup
Clicking on the same feature again: should keep the popup on the screen
Clicking on a map area with no features: should remove the popup
Clicking on a different feature: should reuse the popup, and render it on the new feature
### Actual Behavior

Clicking on a feature once: renders the popup :+1: 
Clicking on the same feature again: removes the popup :cry: 
Clicking on a map area with no features: removes the popup, or simply doesn't render a popup :+1: 
Clicking on a different feature: doesn't render any popup :cry: 

There's definitely scope for this to be programmer error, but our use is (to my eyes) identical to the Mapbox example at https://www.mapbox.com/mapbox-gl-js/example/popup-on-click/

mourner:
Hmm, can you figure out what's the main difference between your codepen and the Mapbox example?

timiyay:
Comparing the Codepen vs the Mapbox example...

The clear difference is that the Codepen create a single `mapboxgl.Popup()` instance, and attempts to reuse it, like so:

```js
// Create the map

// Create a single instance for our popup
var popup = new mapboxgl.Popup({
  closeButton: false,
  closeOnClick: true
});

map.on("click", function(e) {
  // get a list of features under the clicked point, using map.queryRenderedFeatures()
  
  // Note that we're reusing the popup object we created above, and calling remove()
  // on it if we've clicking a featureless area
  if (!features.length) {
    popup.remove();
    return;
  }
  
  // Again, we're reusing the same instance of mapboxgl.Popup
  popup.setLngLat(feature[0].geometry.coordinates)
    .setHTML(feature[0].properties.name)
    .addTo(map);
});

```

...whereas the Mapbox example is creating a new instance of `mapboxgl.Popup()` every time

```js
// When a click event occurs near a marker icon, open a popup at the location of
[...]

PR #5059, opened 2017-07-27
Check if parentNode exists before trying to removechild.

NealHumphrey:
Fixes bug relating to popup removal if popup has already been removed. 

Popups include an option to closeOnClick. If the user manually removes the popup by some other method (a timeout, clicking a clear button etc,), and then clicks the map, the console logs an error because the popup has already been removed:

```
popup.js:99 Uncaught TypeError: Cannot read property 'removeChild' of null
    at o.remove (popup.js:99)
    at o._onClickClose (popup.js:295)
    at e.Evented.fire (evented.js:94)
    at h (bind_handlers.js:138)
    at HTMLDivElement.s (bind_handlers.js:113)
```

Corresponds to #2395 

@jfirebaugh - I tested this locally and it fixed the problem in my code and popups behaved normally otherwise. I'm not familiar enough with the code to know if there's any other functionality that this might interfere with, but seems unlikely given that anything that didn't pass that if statement would error anyways.

### DIFF
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -95,7 +95,7 @@ class Popup extends Evented {
             this._content.parentNode.removeChild(this._content);
         }
 
-        if (this._container) {
+        if (this._container && this._container.parentNode) {
             this._container.parentNode.removeChild(this._container);
             delete this._container;
         }


jfirebaugh:
Thanks @NealHumphrey. It turned out that it was actually `this._content` that needed the parent check, rather than `this._container`. This was fixed in 65afc08ef02f5eb2081c2e3ee83f23e61722ac14 and I'm adding a test in #5114.

PR #5067, opened 2017-08-01
Implement DDS for `text-max-width` property

nickidlugash:
Implements DDS for `text-max-width` property

## Launch Checklist


ChrisLoer:
Awesome! This looks great. Do you think you could add a test for the zoom-and-property case? Like the other symbol layout properties, `text-max-width` will be a little bit funky in that case and it'd be nice to have the expected behavior in a test case. My expectation is that if you have zoom stops at 5 and 7, and you render a tile at zoom 6.5, you should get max-width interpolated as if it were at zoom level 6.

nickidlugash:
@ChrisLoer Added, thanks 👍

PR #5069, opened 2017-08-01
Type most of render directory

jfirebaugh:


src/geo/transform.js
```diff
@@ -29,7 +29,7 @@ class Transform {
     angle: number;
     rotationMatrix: Float64Array;
     zoomFraction: number;
-    pixelsToGLUnits: [number, number];
+    pixelsToGLUnits: Array<number>;
```
jfirebaugh: Because `gl.uniform2fv` doesn't accept a tuple.

PR #5073, opened 2017-08-01
Use separate attribute elements for line normals

jfirebaugh:
Broadcom GPUs don't cope well with using the least significant bit for this.

Refs https://github.com/mapbox/mapbox-gl-native/issues/7583.

### DIFF
--- a/src/data/bucket/line_bucket.js
+++ b/src/data/bucket/line_bucket.js
@@ -5,6 +5,7 @@ const createElementArrayType = require('../element_array_type');
 const loadGeometry = require('../load_geometry');
 const EXTENT = require('../extent');
 const vectorTileFeatureTypes = require('@mapbox/vector-tile').VectorTileFeature.types;
+const {packUint8ToFloat} = require('../../shaders/encode_attribute');
 
 import type {BucketParameters} from '../bucket';
 import type {ProgramInterface} from '../program_configuration';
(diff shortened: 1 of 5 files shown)


lbud:
Given that the line program is one of the programs that has hit snags by exceeding `GL_MAX_VERTEX_ATTRIBS` on some hardware, should we try to minimize space used here at least by packing the two normal elements into one number?

jfirebaugh:
Good thought, but this PR doesn't increase the number of attributes -- it uses the previously unused `z` and `w` components of an existing attribute. Packing both bits into a single component is possible, and would reduce the buffer size slightly, but it wouldn't affect the attribute count, and I'm wary of bitpacking now that I've seen it fail on at least one GPU.

lbud:
Aha — I had skimmed something somewhere that said
> Arrays of vertex shader inputs (vertex attributes) are supported, but each element of the array counts against your GL_MAX_VERTEX_ATTRIBS limit.

but I see now based on official docs ("the maximum number of 4-component generic vertex attributes accessible to a vertex shader") that that source is either mistaken or poorly worded.

mourner:
I'm worried that we once again increase the amount of data uploaded to the GPU, this time by 50% for line layers (which are by far the heaviest in our maps).

jfirebaugh:
@mourner Do you have an alternative in mind?

jfirebaugh:
Ok, going to go with adding a single extra component -- tested, works on device, only increases size by 25%.

PR #5088, opened 2017-08-02
More flow

jfirebaugh:
Up to 89% coverage now!

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -386,7 +386,7 @@ class Camera extends Evented {
             return 0;
         }), ["bottom", "left", "right", "top"])) {
             util.warnOnce("options.padding must be a positive number, or an Object with keys 'bottom', 'left', 'right', 'top'");
-            return;
+            return this;
         }
 
         bounds = LngLatBounds.convert(bounds);
@@ -410,7 +410,7 @@ class Camera extends Evented {
 
         if (scaleY < 0 || scaleX < 0) {
             util.warnOnce('Map cannot fit within canvas with the given bounds, padding, and/or offset.');
-            return;
+            return this;
         }
 
         options.center = tr.unproject(nw.add(se).div(2));
--- a/src/ui/handler/drag_pan.js
+++ b/src/ui/handler/drag_pan.js
@@ -1,7 +1,11 @@
+// @flow
 
 const DOM = require('../../util/dom');
 const util = require('../../util/util');
 const window = require('../../util/window');
+
+import type Map from '../map';
+import type Point from '@mapbox/point-geometry';
 
 const inertiaLinearity = 0.3,
     inertiaEasing = util.bezier(0, 0, inertiaLinearity, 1),
@@ -15,7 +19,15 @@ const inertiaLinearity = 0.3,
  * @param {Map} map The Mapbox GL JS map to add the handler to.
  */
 class DragPanHandler {
-    constructor(map) {
+    _map: Map;
+    _el: HTMLElement;
+    _enabled: boolean;
+    _active: boolean;
+    _pos: Point;
+    _startPos: Point;
+    _inertia: Array<[number, Point]>;
+
+    constructor(map: Map) {
         this._map = map;
         this._el = map.getCanvasContainer();
 
@@ -74,7 +86,7 @@ class DragPanHandler {
         this._enabled = false;
     }
 
-    _onDown(e) {
+    _onDown(e: MouseEvent | TouchEvent) {
         if (this._ignoreEvent(e)) return;
         if (this.isActive()) return;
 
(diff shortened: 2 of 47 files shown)


src/util/find_pole_of_inaccessibility.js
```diff
@@ -35,7 +34,7 @@ module.exports = function (polygonRings, precision, debug) {
     // a priority queue of cells in order of their "potential" (max distance to polygon)
     const cellQueue = new Queue(null, compareMax);
 
-    if (cellSize === 0) return [minX, minY];
+    if (cellSize === 0) return new Point(minX, minY);
```
jfirebaugh: Flow found this bug.

src/ui/handler/drag_rotate.js
```diff
@@ -212,14 +227,14 @@ class DragRotateHandler {
...
-            if (typeof InstallTrigger !== 'undefined' && e.button === 2 && e.ctrlKey &&
+            if (typeof window.InstallTrigger !== 'undefined' && e.button === 2 && e.ctrlKey &&
                 window.navigator.platform.toUpperCase().indexOf('MAC') >= 0) {
                 // Fix for https://github.com/mapbox/mapbox-gl-js/issues/3131:
                 // Firefox (detected by InstallTrigger) on Mac determines e.button = 2 when
                 // using Control + left click
                 eventButton = 0;
             }
-            return (e.type === 'mousemove' ? e.buttons & buttons === 0 : !this.isActive() && eventButton !== button);
+            return (e.type === 'mousemove' ? e.buttons && buttons === 0 : !this.isActive() && eventButton !== button);
```
jfirebaugh: I changed this because flow was complaining, but I'm not sure it's correct... the intent is pretty obscure.

src/ui/handler/drag_pan.js
```diff
@@ -197,7 +209,7 @@ class DragPanHandler {
             if (e.ctrlKey) return true;
             const buttons = 1,  // left button
                 button = 0;   // left button
-            return (e.type === 'mousemove' ? e.buttons & buttons === 0 : e.button && e.button !== button);
+            return (e.type === 'mousemove' ? e.buttons && buttons === 0 : e.button && e.button !== button);
```
jfirebaugh: Ditto.

src/ui/camera.js
```diff
@@ -386,7 +386,7 @@ class Camera extends Evented {
             return 0;
         }), ["bottom", "left", "right", "top"])) {
             util.warnOnce("options.padding must be a positive number, or an Object with keys 'bottom', 'left', 'right', 'top'");
-            return;
+            return this;
```
jfirebaugh: Flow found this bug.

src/ui/camera.js
```diff
@@ -410,7 +410,7 @@ class Camera extends Evented {
 
         if (scaleY < 0 || scaleX < 0) {
             util.warnOnce('Map cannot fit within canvas with the given bounds, padding, and/or offset.');
-            return;
+            return this;
```
jfirebaugh: Flow found this bug.

src/ui/handler/drag_rotate.js
anandthakker: Hmm, yeah.  Peeking at the history, I actually suspect that this was intended to be `(e.buttons & buttons) === 0`, but I'm still not exactly sure what it's meant to be doing.

src/ui/map.js
```diff
@@ -1156,7 +1184,7 @@ class Map extends Camera {
         image: HTMLImageElement | $ArrayBufferView,
         options?: {width: number, height: number, pixelRatio: number}
     ) {
-        this.style.spriteAtlas.addImage(name, image, options);
+        this.style.spriteAtlas.addImage(name, image, (options : any));
```
anandthakker: Should we just type `options` as optional on `SpriteAtlas.addImage` as well?

jfirebaugh: I'm rewriting this code as part of #4876; will be addressed there.

jfirebaugh:
Going to go ahead and revert that; we should do a general pass over `any` casts as we finish up flow typing.

src/ui/handler/drag_rotate.js
jfirebaugh: I think `(e.buttons & buttons) === 0` was indeed the intention, but in fact that prevents drag pan/rotate operations from working on Safari, which does not support the `buttons` property. Whereas `e.buttons & buttons === 0` was simply always false, so `mousemove` events were never ignored, which is harmless!

To fix this, I'm going to remove the logic for ignoring `mousemove` events. Since it was never ignoring them, and we never noticed, it must be unnecessary.

PR #5092, opened 2017-08-03
Enable map recycling option when running render tests in GL JS

brunoabinader:
Enables map recycling option when running render tests in GL JS.

Depends on https://github.com/mapbox/mapbox-gl-js/pull/5091 to be merged first.

This PR:
- Causes the `load` event to be fired again when setting a style after a previous set style.
- Enables map recycling in GL JS render tests via `options.recycleMap`.


jfirebaugh:
Changing the semantics of the `load` event would be a significant disruption for existing code; let's find another way to accomplish this.

brunoabinader:
> Changing the semantics of the load event would be a significant disruption for existing code; let's find another way to accomplish this.

@anandthakker suggested we could listen to the `'data'` event instead, and check the event payload for `e.dataType === 'style'`.

anandthakker:
^ Yep, that should work in the `{diff: true}` case, but note that if you do `map.setStyle(newStyle)` without diff: true, that event would only be fired if the diff-and-patch operation fails and we have to fall back to wholesale replacement of the style.

package.json
```diff
@@ -131,10 +131,11 @@
     "lint-css": "stylelint 'dist/mapbox-gl.css'",
     "open-changed-examples": "git diff --name-only mb-pages HEAD -- docs/_posts/examples/*.html | awk '{print \"http://127.0.0.1:4000/mapbox-gl-js/example/\" substr($0,33,length($0)-37)}' | xargs open",
     "test": "run-s lint lint-css test-unit test-flow",
-    "test-suite": "run-s test-render test-query",
+    "test-suite": "run-s test-render test-render-recycle-map test-query",
     "test-suite-clean": "find test/integration/*-tests -mindepth 2 -type d  -not \\( -exec test -e \"{}/style.json\" \\; \\) -print | xargs -t rm -r",
     "test-unit": "tap --reporter dot --no-coverage test/unit",
     "test-render": "node --max-old-space-size=2048 test/render.test.js",
+    "test-render-recycle-map": "node --max-old-space-size=2048 test/render.test.js --recycle-map",
```
jfirebaugh: I think `yarn test-render --recycle-map` should work, which would make this unnecessary.

brunoabinader: This works, but I had to append `--` before the script arguments:
`yarn test-render -- --recycle-map`

circle.yml
```diff
@@ -38,6 +38,12 @@ workflows:
           filters:
             tags:
               only: /.*/
+      - test-render-recyrcle-map:
```
ChrisLoer: `recyrcle-map` looks like a typo

springmeyer:
Is this ready to merge?

brunoabinader:
> Is this ready to merge?

This is stale - requires rebase + logic update to avoid changing `load` event semantics, as suggested by @jfirebaugh.

jfirebaugh:
Closing here but without prejudice towards updating with the needed changes and resubmitting -- would be nice to have this.

PR #5096, opened 2017-08-04
More flow

jfirebaugh:
Type the remaining .js files and add a lint rule that requires `// @flow`. Excluding `style-spec` directory for now. If we start adding annotations there, we need to figure out how that will work with publishing mapbox-gl-style-spec as a separate module.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,11 @@
 {
-  "extends": "mourner",
+  "extends": [
+    "mourner",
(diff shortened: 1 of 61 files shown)


anandthakker:
> Excluding style-spec directory for now. If we start adding annotations there, we need to figure out how that will work with publishing mapbox-gl-style-spec as a separate module

Good point.  Note that the expressions branch includes a lot of flow annotations in the `style-spec` dir.

Issue #4601, opened 2017-04-18
Outline drawing doesn't work in headless tests


kkaefer:
While investigating https://github.com/mapbox/mapbox-gl-js/issues/4600, I discovered that drawing polygon outlines (antialiasing) isn't working correctly in headless tests. It does, however, work correctly when rendering in the browser.

Cf.:

JS headless:

![unknown](https://cloud.githubusercontent.com/assets/52399/25134260/6e8028b4-244f-11e7-90dc-7a7c12e88561.png)

Native headless:

![unknown-1](https://cloud.githubusercontent.com/assets/52399/25134340/9734c792-244f-11e7-8759-728e766b10cb.png)

This example is `zoomed-fill/default` from our [render tests](https://7602-8629417-gh.circle-artifacts.com/0/Users/distiller/mapbox-gl-js/test/integration/render-tests/index.html). Other examples in the render tests include `tms/tms`, `extent/1024-fill`, and any other tests that render fill polygons.

kkaefer:
See https://github.com/mapbox/mapbox-gl-native/issues/6927#issuecomment-266606578 for a possible cause.

jfirebaugh:
I think I saw it wasn't working 100% in the browser either -- it looked like `glLineWidth` didn't support  any values other than 1.

http://webglstats.com/webgl/parameter/ALIASED_LINE_WIDTH_RANGE

ansis:
ref #2080

> I think I saw it wasn't working 100% in the browser either

Since I *think* ANGLE doesn't support anything > 1 I *think* it's broken on windows in firefox and chrome.

jfirebaugh:
Duplicate of #2080.

PR #5099, opened 2017-08-04
Update fill-outline-color test expectations

kkaefer:
https://github.com/mapbox/mapbox-gl-native/issues/6927 mentions that these tests are different in JS vs. Native. Looking more at the code, I believe that the test expectations are wrong, and they should have the fuzzy antialiasing edges. JS *headless* doesn't have these due to https://github.com/mapbox/mapbox-gl-js/issues/4601. However, when rendering these stylesheets in the browser, they do look like Native versions.


test/ignores.json
```diff
@@ -9,7 +9,6 @@
   "render-tests/fill-extrusion-pattern/literal": "https://github.com/mapbox/mapbox-gl-js/issues/3327",
   "render-tests/fill-extrusion-pattern/missing": "https://github.com/mapbox/mapbox-gl-js/issues/3327",
   "render-tests/fill-extrusion-pattern/opacity": "https://github.com/mapbox/mapbox-gl-js/issues/3327",
-  "render-tests/fill-outline-color/fill": "https://github.com/mapbox/mapbox-gl-js/issues/4601",
```
jfirebaugh: `fill-outline-color/fill` isn't included in the image updates. Was it passing everywhere already?

kkaefer: Yes

Issue #5090, opened 2017-08-03
Glyph x-offset calculation incorrect for vertical glyphs


ChrisLoer:
When we introduced CPU placement of line labels in `0.39` (#4781), we changed the center we used for rotating vertical glyphs (see: https://github.com/mapbox/mapbox-gl-js/blob/master/src/symbol/quads.js#L169). The new center isn't actually the center of the glyph, so rotation ends up shifting the x-position of the glyph. This isn't very noticeable in all-vertical text because everything's just shifted a little bit along the line. However, if you mix vertical and non-vertical text, you start seeing overlap:

![screenshot 2017-08-02 16 30 03](https://user-images.githubusercontent.com/375121/28900489-728c8588-77a6-11e7-85e3-dad5c7eb3d61.png)

We introduced the same bug into `gl-native` at the same time.

/cc @1ec5 @nickidlugash @mollymerp @ansis

ChrisLoer:
I may be misunderstanding something about what this code is trying to accomplish, but it looks like we end up rotating the glyph around what is _approximately_ the left center side. For a CJK glyph with advance of 24, we'll rotate around `(-12,12)`. You might think of the ideal CJK glyph as having a top-left of `(-12,0)` and a bottom-right of `(12,24)` although the actual metrics will be different. 

Rotating around this point has the effect of moving the glyph up (up == negative y) and to the left (negative x). At the same time, in the [shaping code](https://github.com/mapbox/mapbox-gl-js/blob/master/src/symbol/shaping.js#L296), we don't put in the default `yOffset` of `-17` for vertical glyphs. The two y changes just about cancel each other out, so the glyphs ends up showing correctly on the line, but the x effect means we're shifted about a half advance (12 points for CJK) to the left.

I think the right approach here is actually much simpler: we just need to keep the same `yOffset` and rotate the glyph around whatever its actual center is.

As I'm looking at this code, I think we should change `text-rotate` to work the same way. Right now, it rotates the glyph around the glyph origin (which is not the center), so that if you use it with a line label you'll end up being offset from the line (on the other hand, I'm not really familiar with how `text-rotate` is used, this example case doesn't make much sense):

<img width="122" alt="screenshot 2017-08-03 11 28 03" src="https://user-images.githubusercontent.com/375121/28937158-dd681014-783e-11e7-9452-1df741e8edda.png">

ChrisLoer:
[...]

PR #5100, opened 2017-08-04
Correct x-offset introduced by vertical glyph rotation.

ChrisLoer:
Fixes issue #5090.

The render test `text-writing-mode/mixed` shows the core of the problem this PR is trying to fix -- when vertical and horizontal text are mixed, it's especially important to make sure we're not changing the x-offset of only the vertical text. Look at "eight" (which should have a space after it) and "33" (which shouldn't have a space before it). You can also see the vertically oriented text in the upper-right of `text-writing-mode/punctuation` now agrees better with the horizontally oriented equivalent in the bottom right.

/cc @ansis @1ec5 @nickidlugash @mollymerp


cs09g:
It's very far away from the day of PR, could you explain why is `5` in line `const xOffsetCorrection = new Point(5, 0);`? because for me as user of CJK fonts, it's not aligned well with the value. 9 or 10 is good for my font.

ansis:
@cs09g could you open a new issue with the problem you are seeing?

Issue #4134, opened 2017-02-01
Correct depth rendering for multiple fill extrusions


lucaswoj:
_From @lbud on October 20, 2016 22:33_

## Motivation

Currently we render individual extruded layers to a separate texture, and then render that texture at an opacity specified by that layer's `paint.fill-opacity` value back to the map. We do this in order to e.g. not render interior walls in buildings.

This causes problems when creating multiple extruded layers, as the full layers are painted back to the map separately with no respect to other properties, defying the laws of physics and rendering wonky af. This problem will continue to snowball as we add more 3D features.
## Design Alternatives

❌ do nothing 🙅⛄️
❌ abandon the texture approach and render only at full opacity
❌ abandon the texture approach and render interior walls
✅ render all extruded/z-offset layers to the same separate texture, and then render back to the map at a root-level opacity

👇 
#### options (flexible on wording), all starting at root level:

A.

```
{
  "3d-opacity": <number>
}
```

B.

```
{
  "3d-properties": {
    "opacity": <number>
  }
}
```

C.

```
{
  "3d-properties": {
    "opacity": <number>,
    "light": {
        ... light properties folded in here, since they only affect extrusions
    }
}
```
## Design

I'd prefer not to do A, as further 3D properties may continue to clutter the root object; this feels short-sighted.
- **Breaking**: `fill-opacity` in extruded layers already created will need to be ignored

I _want_ to prefer B. The corollary to the above is that using a `3d-properties` object will leave us the option to easily add more 3d properties later without further cluttering the root object.
- **Breaking**: (same as above)

_However_, I instinctively think C is the right answer: `light` properties already only affect 3d layers, so in the long run this makes the most sense, but is also the most breaking option.
- **Breaking**: (same as above)
- **Breaking**: `light` properties get moved around
### Implementation

I think it shouldn't be too hard to switch to a single 3d texture that we only render back to the map after all layers are rendered.
### Considerations

Calling `setPaintProperty` to transition from a flat fill layer with any degree of transparency to an extruded fill layer will lose its specified transparency.

It sucks that these are breaking changes, but it would be best to do this ASAP as at this point relatively few people (early adopters) have created styles with extrusions, they're not in Studio yet, and they aren't in default styles yet.

@mapbox/gl 

_Copied from original issue: mapbox/mapbox-gl-style-spec#545_

lucaswoj:
_From @jfirebaugh on October 21, 2016 0:5_

[...]

Issue #4875, opened 2017-06-23
Port Render* split and immutability from native


jfirebaugh:
We should replace the current smart `setStyle` implementation in gl-js with one more similar to native, based on a split between:

* immutable, pure style-spec domain objects, independent of rendering, and
* parallel `Render*` objects that handle rendering-specific concerns and can update themselves by diffing prior and new immutable objects

This approach eliminates the complex state tracking needed to track "pending" changes (`Style#_updatedLayers`, `_updatedSymbolOrder`, `_updatedSources`, etc.).

anandthakker:
I've been thinking a bit about how we should express immutability semantics for this work.  gl-native uses a set of generic classes [`Mutable<T>`, `Immutable<T>`](https://github.com/mapbox/mapbox-gl-native/blob/7c1d79fb95882003a7424a181bcdfb95bbfe23a9/include/mbgl/util/immutable.hpp) (for immutable objects), and [`Collection<ClassWithImmutableImpl>`](https://github.com/mapbox/mapbox-gl-native/blob/7c1d79fb95882003a7424a181bcdfb95bbfe23a9/include/mbgl/util/immutable.hpp) (for a collection capable of perf benefits from immutability-awareness).  Together, these provide:
1. Static enforcement of immutable state management
2. Zero runtime overhead (relative to pre-existing implementation -- I think)
3. DRY implementation, abstracted into the general-purpose classes linked above

So far, it looks to me like we can probably only get two out of three of these with JS/Flow (specifically, I don't see a way to get 3 without compromising 2).  I'm leaning towards an approach where we:
 - Treat the `Style` > `StyleLayer` > `Layout`/`Transitionable` class hierarchy as private implementations (`StyleImpl` > `StyleLayerImple` > etc.)
 - Expose each one to rest of the codebase through a pair of interfaces:
```js
interface StyleLayer {
  /* getters, but no setters */
  getMutable(): MutableStyleLayer;
}
interface MutableStyleLayer extends StyleLayer {
  /* setters */
}
class StyleLayerImpl implements MutableStyleLayer {
   /* getters and setters */
   getMutable() { this.shallowClone(); }
}
```
 - Where we need similar semantics for raw objects or arrays, use `type MutableBlah = 
 ...; type Blah = $ReadOnly<MutableBlah>`, together with a `asMutable<T>(x: $ReadOnly<T>): T { return shallowClone(x); }`

Updating state would then look like:
```js
[...]

PR #5101, opened 2017-08-04
Preserve depth buffer between fill-extrusion layers + optimize render order

lbud:
This PR changes the way we render fill-extrusion layers:
* Optimizes render order by rendering all fill-extrusion textures first, so that we don't have to restore any framebuffer contents (this optimization is outlined in https://github.com/mapbox/mapbox-gl-native/issues/9286 but not ticketed in gl-js)
* Preserves depth buffer contents between different fill-extrusion layers in the same render pass so that multiple fill-extrusion layers can be rendered on the same map (fixes https://github.com/mapbox/mapbox-gl-js/issues/4134)

Before + after (the blue layer is on top of the green layer in the render stack):
image

Technically this is a "breaking" rendering change ⚠️ , though I think we can all agree that it was already broken…

One odd quirk of this PR is the changes made in https://github.com/mapbox/mapbox-gl-js/commit/b1559131a5f58d03a47a26c137e9e1ca17007115: on my live maps, this worked whether clearing the depth buffer before binding or rendering any textures, or after rendering and unbinding all textures. However, all fill-extrusion tests were failing to render any fill-extrusions at this point. I'm not sure why clearing the depth buffer only once a texture (color target) is bound fixes the issue for the render tests, because it seems to me the depth clearing should work once the depth renderbuffer is attached whether or not there is a render target bound… 🤷‍♀️ 

## Launch Checklist

 - [ ] corresponding mapbox-gl-native PR


src/render/painter.js
```diff
@@ -118,14 +124,20 @@ class Painter {
         this.height = height * browser.devicePixelRatio;
         gl.viewport(0, 0, this.width, this.height);
 
-        if (this.viewportTexture) {
-            this.gl.deleteTexture(this.viewportTexture);
-            this.viewportTexture = null;
+        for (let i = 0; i < this.viewportTextures.length; i++) {
```
jfirebaugh: Use a `for ... of` loop here.

lbud:
> if fill-extrusion-opacity is 1, can it copy the texture over in the opaque pass?)

No — the opaque pass relies on fragments being able to retain their place through subsequent draw calls based on their depth. Since we can't write to `gl_FragDepth`, the full viewport would be assigned the same depth value and would block all layers underneath:
image

src/render/draw_circle.js
```diff
@@ -11,7 +11,8 @@ import type TileCoord from '../source/tile_coord';
 module.exports = drawCircles;
 
 function drawCircles(painter: Painter, sourceCache: SourceCache, layer: CircleStyleLayer, coords: Array<TileCoord>) {
-    if (painter.isOpaquePass) return;
+    const pass = 'translucent';
+    if (painter.renderPass !== pass) return;
```
jfirebaugh: In places (x4) where there's no conditionals involved in computing `pass`, just compare directly against the string literal.

src/render/painter.js
```diff
@@ -261,6 +274,87 @@ class Painter {
...
+        {
+            const gl = this.gl;
+            // We'll wait and only attach the framebuffer if we think we're
+            // rendering something.
+            let fboAttached = false;
+
+            let sourceCache;
+            let coords = [];
+
+            for (let i = 0; i < layerIds.length; i++) {
```
jfirebaugh: Are there ways we could simplify the control flow in the body of this loop? For example, move checks such as `layer.paint['fill-extrusion-opacity'] === 0` into `has3DPass` and set `this._prerenderedTextures[layer.id] = null` up front for every layer? Or even eliminate `has3Dpass`, have the layers without a 3D pass simply do nothing, and have the 3D layers do the texture setup inside their `render` functions? Not sure what exactly is kosher performance-wise, but this code seems like it could be simpler.

src/render/painter.js
```diff
@@ -261,6 +274,87 @@ class Painter {
...
+
+                    if (!fboAttached) {
+                        // This is the first 3D layer we're rendering: attach
+                        // the framebuffer now.
+                        this._setup3DFramebuffer();
+                        // Wait to flip the boolean until after we attach the first
+                        // texture and clear the depth buffer a few lines down.
+                    }
+
+                    let renderTarget = this.viewportTextures.pop();
```
jfirebaugh: It looks like `this.viewportTextures` will always be an empty array. Am I missing somewhere where textures get released back (say when the layer is removed from the style)?

Instead of having a `this._prerenderedTextures` map keyed on layer ID, could fill extrusion layers have a `prerenderedTexture` property?

src/render/painter.js
```diff
@@ -269,58 +363,97 @@ class Painter {
...
-    }
 
-    renderPass() {
-        const layerIds = this.style._order;
+            for (this.currentLayer; this.currentLayer >= 0; this.currentLayer--) {
+                const layer = this.style._layers[layerIds[this.currentLayer]];
 
-        let sourceCache;
-        let coords = [];
+                if (layer.source !== (sourceCache && sourceCache.id)) {
```
jfirebaugh: (I know this bit of code is duplicated in similar form in all three of the passes -- not ideal, but okay for now IMO. Hopefully as part of #4875 we can bring over the system in native where the coordinates are precomputed once for all layers ahead of time.)

src/render/render_texture.js
```diff
@@ -0,0 +1,48 @@
...
+        array.emplaceBack(1, 0);
+        array.emplaceBack(0, 1);
+        array.emplaceBack(1, 1);
+        this.buffer = Buffer.fromStructArray(array, Buffer.BufferType.VERTEX);
+        this.vao = new VertexArrayObject();
+    }
+
+    attachToFramebuffer() {
+        const gl = this.gl;
+        gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, this.texture, 0);
```
kkaefer: Ideally, we'd be using one *framebuffer* for each layer (with different color attachments, and the same depth attachment), rather than detaching/reattaching textures to the framebuffer. The reason for this is that it's expensive to "validate" that a framebuffer has correctly bound attachments. It looks like we don't do that yet in JS; in Native, we're [running `glCheckFramebufferStatus`](https://github.com/mapbox/mapbox-gl-native/blob/e8bfcd04740bbc35a1e25038fe7b7d2577792b66/src/mbgl/gl/context.cpp#L365-L392). Therefore, the API in Native is modeled so that you can't change the attachments of a framebuffer after creation.

http://www.songho.ca/opengl/gl_fbo.html claims that the *opposite* of what I said is true:

> Framebuffer object (FBO) provides an efficient switching mechanism; detach the previous framebuffer-attachable image from a FBO, and attach a new framebuffer-attachable image to the FBO. Switching framebuffer-attachable images is much faster than switching between FBOs.

However, further research suggests that this is outdated information, and no longer true for modern GPUs:

* [Intel built-in GPUs](https://software.intel.com/en-us/articles/opengl-performance-tips-swap-frame-buffer-objects-fbo-instead-of-surfaces-using-a-single)

  > Switching an entire FBO is more efficient than switching individual surfaces one at a time.

* [PowerVR](http://cdn.imgtec.com/sdk-documentation/PowerVR.Performance+Recommendations.pdf)

  > For optimal performance, [...] attachments should not be added or removed once the FBO has been created.

* [nVidia](https://developer.nvidia.com/sites/default/files/akamai/gamedev/docs/Porting%20Source%20to%20Linux.pdf)

  > Do not create a single FBO and then swap out attachments on it. This causes lots of validation in the driver, which in turn leads to poor performance.

[ARM/Mali](https://community.arm.com/graphics/b/blog/posts/mali-performance-2-how-to-correctly-handle-framebuffers) also has a very good article on framebuffer handling and how it interacts with buffer swapping that also confirms the behavior we've seen of PowerVR GPUs which prompted https://github.com/mapbox/mapbox-gl-native/issues/9286

lbud: 👀 interesting — thank you for the links @kkaefer! 🛠

src/render/painter.js
lbud: > have the 3D layers do the texture setup inside their `render` functions

Conceptually I don't prefer this — as it is, the `render` functions are mostly just responsible for rendering whatever they have to render to whatever target is bound, and the painter is responsible for managing render targets and switching between them. Especially if we add more types with `3d` passes I'd prefer to have the render target management for all 3D types happen in the painter's 3D pass loop. I did some simplification in https://github.com/mapbox/mapbox-gl-js/pull/5101/commits/677c2610a27c73df45f897a666bc3a69be082669.

src/render/painter.js
lbud: > Am I missing somewhere where textures get released back (say when the layer is removed from the style)?

Ah, I forgot this — added in https://github.com/mapbox/mapbox-gl-js/pull/5101/commits/0651dcfa32fec3bd4d453f5cbf8008ff25ddf1ac. 

In order for fill extrusion layers to have a `prerenderedTexture` property, we'd have to make fbos/textures members of style layers, right? Which feels weird…

src/render/render_texture.js
lbud: @kkaefer I implemented this in https://github.com/mapbox/mapbox-gl-js/pull/5101/commits/0651dcfa32fec3bd4d453f5cbf8008ff25ddf1ac. We still have to detach/reattach the depth renderbuffer between framebuffers, so I do wonder about performance gains — before there was one framebuffer with an always-attached renderbuffer and we switched texture attachments, and now there are many framebuffers with their own texture but we switch renderbuffer attachments, so I assume they still do validation every time we attach the depth rbo… 🤔

kkaefer: Hmm, why are we attaching the depth renderbuffer for every render? Can we leave the same renderbuffer attached to multiple framebuffers?

lbud: Oh, interesting — okay, I guess I had conceptually assumed that since `gl.framebufferRenderbuffer` attaches a renderbuffer to a framebuffer, that attachment was unique to the bound framebuffer. It seems instead based on what you're implying (and manual testing confirms this) that that attaches a renderbuffer to whatever is bound at the `gl.FRAMEBUFFER` attachment point at the time of the render?
To me [this](https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glFramebufferRenderbuffer.xml) doesn't read super clearly either way —
> attach a renderbuffer object to a framebuffer object

> glFramebufferRenderbuffer attaches the renderbuffer specified by renderbuffer as one of the logical buffers of the currently bound framebuffer object

but it does seem to work when I only attach the depth RBO on the first fill-extrusion draw…

kkaefer: hmm, what could work is upon creation of a framebuffer, attach both the depth RBO + color texture to the FBO, with every FBO having a different texture, but the same depth RBO?

lbud: Right — this is how it currently works

src/render/render_texture.js
```diff
@@ -0,0 +1,55 @@
...
+        const array = new PosArray();
+        array.emplaceBack(0, 0);
+        array.emplaceBack(1, 0);
+        array.emplaceBack(0, 1);
+        array.emplaceBack(1, 1);
+        this.buffer = Buffer.fromStructArray(array, Buffer.BufferType.VERTEX);
+        this.vao = new VertexArrayObject();
+    }
+
+    attachRenderbuffer(depthRbo: WebGLRenderbuffer) {
```
asheemmamoowala: The name of this method and `detachRenderBuffer` is a bit misleading. Can we make `DEPTH_ATTACHMENT` a parameter, or change to `attachDepthBuffer`/`detachDepthBuffer` ?

src/render/painter.js
```diff
@@ -257,72 +266,166 @@ class Painter {
...
+                    first = false;
+                }
+
+                this.renderLayer(this, (sourceCache: any), layer, coords);
+
+                renderTarget.detachDepthRenderbuffer();
+                this.prerenderedFrames[layer.id] = renderTarget;
             }
+
+            if (this.depthRboAttached) gl.bindRenderbuffer(gl.RENDERBUFFER, null);
```
jfirebaugh: `_setup3DRenderbuffer` always unbinds the renderbuffer, so is this necessary?

kkaefer: Given that we're binding/unbinding framebuffers, I believe we shouldn't be unbinding renderbuffers manually, since this state is part of the FBO.

VNuhaan:
I have a problem with the new render order.

I have a base map at low detail and an overlay with high detail. With this change all low detail buildings are drawn on top and thru the high detail map.

Is it possible to add an option to select if the extrusion mixes with other extrusions or overrides other extrusion?

Issue #5102, opened 2017-08-04
collision-pitched render tests fail locally


jfirebaugh:
I think this has been discussed before, but nobody has opened an issue yet. The `debug/collision-pitched` and `debug/collision-pitched-wrapped` test fail when run locally on macOS.

![image](https://user-images.githubusercontent.com/98601/28985321-d3cc8fde-7916-11e7-962b-3d208ac0f681.png)
![image](https://user-images.githubusercontent.com/98601/28985329-dddbb4a0-7916-11e7-95ee-aafa82b7ae0f.png)

cc @ChrisLoer

PR #5103, opened 2017-08-04
Add ignores for https://github.com/mapbox/mapbox-gl-js/issues/5102

jfirebaugh:


ChrisLoer:
I wasn't aware we were getting local failures of this test -- it passes locally on my machine (macOS 10.12.6). Looks like the cause is something like different implementations of gl.LINES for the tile boundaries? It's already ignored on the native side in part for that reason (https://github.com/mapbox/mapbox-gl-native/issues/3841).

I don't think the tile boundaries are an important part of this test -- I'd rather drop them and keep the test running on at least the gl-js side than have the test ignored on both platforms.

PR #5106, opened 2017-08-05
Fix initialization of CollisionTile causing exception

mlmassey:
In the constructor for a CollisionTile, if the collisionBoxArray parameter passed in was null or undefined, the constructor will throw an exception of length is not defined for null.  Since the code is later initializing to a temporary CollisionBox, create a temporary collisionBoxArray so code can continue without exception.

The problem I was running into is that a javascript exception was being thrown, although the code continued to appear to work.  The source of the problem was the Tile.loadVectorData() function.  The Tile object can be created without a collisionBoxArray member, which is only later initialized when loadVectorData() is called.  The slightly strange thing is in Tile.js:119, a Tile can be set this.state = 'loaded' even though the this.collisionTitle and this.collisionBoxArray parameters are never initialized.  (in the case the data parameter passed in is null/undefined).  It would seem better to initialize these members to default values in the constructor or not allow the state to be loaded if these are not set.  Otherwise, all functions/objects have to handle the case where the values may be null/undefined (such as I pointed out above).

Since this is a very minor change and could not really cause a side-effect, I did not run all your build checks and unit tests.  My apologies.

### DIFF
--- a/src/symbol/collision_tile.js
+++ b/src/symbol/collision_tile.js
@@ -4,6 +4,7 @@ const Point = require('@mapbox/point-geometry');
 const EXTENT = require('../data/extent');
 const Grid = require('grid-index');
 const intersectionTests = require('../util/intersection_tests');
+const CollisionBoxArray = require('../symbol/collision_box');
 
 import type CollisionBoxArray, {CollisionBox} from './collision_box';
 
--- a/src/symbol/collision_tile.js
+++ b/src/symbol/collision_tile.js
@@ -92,6 +92,10 @@ class CollisionTile {
         // boxes at render time.
         this.yStretch = Math.max(1, cameraToTileDistance / (cameraToCenterDistance * Math.cos(pitch / 180 * Math.PI)));
 
+        if (!collisionBoxArray) {
+            // If collisionBoxArray was not initialize, create a new object
+            collisionBoxArray = new CollisionBoxArray();
+        }
         this.collisionBoxArray = collisionBoxArray;
         if (collisionBoxArray.length === 0) {
             // the first time collisionBoxArray is passed to a CollisionTile


ChrisLoer:
Thanks for your contribution, @mlmassey! The dependency between `CollisionBoxArray` and `CollisionTile` is definitely awkward and we should do some clean-up here. Because it's so complicated, I want to make sure before we fix the exception thrown from the constructor that we're not hiding a case where our assumptions were being violated. 

Were you seeing the exception thrown from `reloadSymbolData`? That's the only place I see where it looks like we could create a `CollisionTile` without having created a `CollisionBoxArray` first. If there's no `CollisionBoxArray` in that code, we assume that means we're basically loading an empty tile. In that case, we shouldn't be trying to re-do collision detection (or placement) at all, because there aren't any symbols in the tile for us to place. It looks like we already try to avoid doing placement in the "empty tile" case with [this check](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/tile.js#L190). Do you know why placement is still happening in your code?

Incidentally, it looks like the build failed because the added require collided with the `import type CollisionBoxArray` (short of running all the tests, you can quickly run our JS linter with `yarn lint`). I think if we identify the fundamental problem that allows a `CollisionTile` to be created without a `CollisionBoxArray`, it may be unnecessary to make any changes at all to `CollisionTile` (or maybe replace the exception on undefined behavior with an explicit assertion).

mlmassey:
I'll get back to you on our code sequence and how we are sequencing calls to get this error to occur.  That should enlighten us more.

mlmassey:
Closing this PR as I cannot reproduce the issue anymore.  If it resurfaces, I will reopen

PR #5111, opened 2017-08-07
Clarify documentation on stop functions for input < lowest stop

### LINKED ISSUE #5098: Clarify documentation about stop function behavior for input < lowest stop value
**mapbox-gl-js version**:
0.39.1

According to the [style specification for Function Types](https://www.mapbox.com/mapbox-gl-js/style-spec/#function-type):

> interval
> **functions return the output value of the stop just less than the function input**. The domain (input value) must be numeric. Any style property may use interval functions. For properties marked with ,  the "interval" symbol, this is the default function type.

This means that any feature with a value **less than the first stop** should take the default value if it exists, and any feature with a value **greater than the last stop** should take the output value of the last stop.

### Steps to Trigger Behavior
https://jsbin.com/naciqocuja/1/edit?js,output

### Expected Behavior
Maryland should be transparent, not red.

### Actual Behavior
Maryland is red, despite it's value being less than the lowest declared stop.

This behavior has existed since version 0.32.0, when `default` was first introduced to the style specification.

anandthakker:
Closes #5098

### DIFF
--- a/docs/style-spec/_generate/index.html
+++ b/docs/style-spec/_generate/index.html
@@ -850,12 +850,14 @@ navigation:
                 <dd>functions return their input as their output.</dd>
                 <dt><var>exponential</var></dt>
                 <dd>functions generate an output by interpolating between stops just less than and just greater than the
-                    function input. The domain (input value) must be numeric, and the style property must support
+                  function input. (For input values that lie below the first stop, the function outputs the value of the first stop.)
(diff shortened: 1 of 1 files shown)


anandthakker:
screen shot 2017-08-07 at 8 47 07 am

screen shot 2017-08-07 at 8 49 09 am

PR #5117, opened 2017-08-08
Remove diff handicap for text-pitch-{alignment,scaling} render tests

brunoabinader:
These tests should fail in GL native, as per https://github.com/mapbox/mapbox-gl-native/issues/9732.


jfirebaugh:
Based on the CI output it looks like gl-js may have the same issue as described for native in https://github.com/mapbox/mapbox-gl-native/issues/9732?

brunoabinader:
> Based on the CI output it looks like gl-js may have the same issue as described for native in mapbox/mapbox-gl-native#9732?

Precisely - I'm adding them to GL JS ignore list for now.

brunoabinader:
I've added an issue in GL JS to track in the render test ignore list: https://github.com/mapbox/mapbox-gl-js/issues/5118

Issue #5112, opened 2017-08-07
Street labels overlapping at high zooms with pitched views


asheemmamoowala:
https://api.mapbox.com/styles/v1/mapbox/navigation-guidance-day-v2.html?access_token=<ACCESS_TOKEN>#19.63/33.86818/-118.28172/-176/45
 
**mapbox-gl-js version**: 0.39.1

### Expected Behavior
Street label text does not overlap

### Actual Behavior
![screen shot 2017-08-07 at 11 49 07 am](https://user-images.githubusercontent.com/11619675/29041104-8199f6b4-7b66-11e7-980f-b04a12562af7.png)

ChrisLoer:
It took me a while looking at that to visually figure out what was going on -- looks like the labels basically start doubling back on themselves at the midpoint. Could be a bug in the `keepUpright` logic. I'll take a look.

/cc @ansis

ChrisLoer:
Wow I had a tough time figuring this one out...

It turns out that if you're zoomed in far enough to a highly pitched tile, the bottom edge of the tile can come *behind* the camera, and our projection only works for points in front of the camera (there's a discontinuity as points cross the plane of the camera, then they flip orientation as their w values go negative). The flipping effect that shows up in the screenshot @asheemmamoowala posted comes because the first half of the label is oriented from the anchor (which is within the viewport and projected just fine) to the top vertex of the line segment (which is in the distance and projected just fine). The second half of the label is oriented from the anchor to the bottom vertex of the line segment -- whose x value will get projected more or less correctly, but whose y value will be flipped so it's past the top of the viewport.

Narrowing the field-of-view of the camera to ten degrees moves the camera back and fixes the problem, but of course changes the way pitched views look... I'm not sure what the best solution is here. We might have to check vertices to see if they cross the plane of the camera -- and if so, instead of using that vertex, we take a unit vector from tile space, project it, and then extend it to the edge of the viewport.

asheemmamoowala:
Fixed in #9865

PR #5120, opened 2017-08-08
Fix issue #5112 (line labels can render incorrectly on overzoomed tiles)

ChrisLoer:
Adds special handling for (rare) case that line geometry projects to a point behind the plane of the camera. Fixes issue #5112.

Before merging the PR, I want to build tests for the two cases here (multiple glyph and single-glyph labels). In development, I've been testing with the coordinates @asheemmamoowala provided in #5112, and modifying the style to make single-glyph labels.

/cc @asheemmamoowala @ansis @kkaefer


ChrisLoer:
Building tests was tricky using a GeoJSON source. For the single glyph special case I ended up just finding a vector tile that could reproduce the issue and I used the vector tile in the test case. Is that a reasonable approach?

src/symbol/projection.js
```diff
@@ -276,11 +290,19 @@ function placeGlyphsAlongLine(symbol,
     return {};
 }
 
+function projectTruncatedLineSegment(previousTilePoint: Point, currentTilePoint: Point, previousProjectedPoint: Point, minimumLength: number, projectionMatrix: mat4) {
+    const projectedUnitVertex = project(previousTilePoint.add(previousTilePoint.sub(currentTilePoint)._unit()), projectionMatrix).point;
```
anandthakker: Would it be possible for `previousTilePoint + unitVector(previousTilePoint - currentTilePoint)` to fall behind the camera, if `previousTilePoint` happened to be within one unit of the camera plane?  If so, would that cause a similar problem to the one we're solving here?

ChrisLoer: 🤔  That's a good question. I think the answer is "not under any reasonable conditions," but let me try to reason why:

 - Any point that's near the plane of the camera is very far "away" (in x/y terms) from the viewport
 - We're only placing glyphs whose anchor is within the viewport
 - For a label to reach from within the viewport to a line vertex at the edge of camera plane, it would need to be _very_ long (much longer than could be shown on one screen)
 - A label with `text-pitch-alignment: map` would be the best candidate for being way longer than expected, because it would get stretched out more and more the closer it got to the plane of the camera. But that case wouldn't be affected by the "plane of the camera" logic, because it would be projected using a tile-based matrix (see logic in `getLabelPlaneMatrix`).
 - Therefore any label we're placing should fit on the line _before_ the hypothetical vertex that's right at the edge of the camera plane, and we should never reach a point where we try to project past the hypothetical vertex.

I guess another way of thinking about it is: if we needed to draw a glyph that was on a line just next to the plane of the camera, we'd be very close to needing to draw a glyph that was just behind the plane of the camera, and we'd be hosed either way.

src/symbol/projection.js
```diff
@@ -246,22 +252,30 @@ function placeGlyphsAlongLine(symbol,
...
         // Only a single glyph to place
         // So, determine whether to flip based on projected angle of the line segment it's on
         if (keepUpright && !flip) {
-            const a = project(lineVertexArray.get(symbol.lineStartIndex + symbol.segment), posMatrix);
-            const b = project(lineVertexArray.get(symbol.lineStartIndex + symbol.segment + 1), posMatrix);
+            const a = project(tileAnchorPoint, posMatrix).point;
+            const tileSegmentEnd = lineVertexArray.get(symbol.lineStartIndex + symbol.segment + 1);
+            const projectedVertex = project(tileSegmentEnd, posMatrix);
+            // We know the anchor will be in the viewport, but the end of the line segment may be
+            // past the plane of the camera, in which case we can use a point at any arbitrary (closer)
```
ChrisLoer: :+1:

src/symbol/projection.js
```diff
@@ -107,10 +107,14 @@ function getGlCoordMatrix(posMatrix: mat4,
...
 
-function project(point: Point, matrix: mat4): Point {
+function project(point: Point, matrix: mat4) {
     const pos = [point.x, point.y, 0, 1];
     vec4.transformMat4(pos, pos, matrix);
-    return new Point(pos[0] / pos[3], pos[1] / pos[3]);
+    const w = pos[3];
+    return {
+        point: new Point(pos[0] / w, pos[1] / w),
+        distanceToCamera: w
```
ChrisLoer: :+1:

src/symbol/projection.js
anandthakker: 👍 sounds legit to me -- thanks for spelling it out!

PR #5122, opened 2017-08-09
Fix erroneous validation error caused by usage in environment with extended prototypes

bcherny:
`a in b` returns `true` if `a` is defined as a key in object `b` or if it's defined as a key anywhere in `b`'s prototype chain.

This is probably not the intention here, and it is causing errors in my environment (where `Object.prototype.filter` is defined).

Please let me know if there's anywhere else this change should be made.

### DIFF
--- a/src/style-spec/validate/validate_layer.js
+++ b/src/style-spec/validate/validate_layer.js
@@ -31,9 +31,9 @@ module.exports = function validateLayer(options) {
         }
     }
 
-    if ('ref' in layer) {
+    if (layer.hasOwnProperty('ref')) {
         ['type', 'source', 'source-layer', 'filter', 'layout'].forEach((p) => {
-            if (p in layer) {
+            if (layer.hasOwnProperty(p)) {
                 errors.push(new ValidationError(key, layer[p], '"%s" is prohibited for ref layers', p));
             }
         });


jfirebaugh:
Sorry, we're not going to support environments where `Object.prototype` has been extended. That's extremely bad practice and will likely cause you a wide variety of issues.

bcherny:
@jfirebaugh Thanks for the quick response, but I couldn't disagree more. We use extensions extensively in our project with no issues - MapboxGL is the only one of dozens of libraries we use that gave us issues.

On our side, we make extensions typesafe with TypeScript (similar to what Scala or Swift do). That extensions are bad practice in JS is due to them not being safe historically; with static type systems, JS catches up to other languages where extensions are common and useful.

On your side, the "in" code is almost certainly a bug. What if the consumer passed an object created via Object.create? You don't want to be in the business of traversing the object's prototype - it's exactly the reason why the hasOwnProperty API exists!

On another note, it makes me sad that we can't use MapboxGL. I've been using your product for years on various projects and I love it, including recommending it to many friends. This is not the sort of support I expected.

jfirebaugh:
I don't believe static typing makes `Object.prototype` extension less dangerous or inadvisable. On the contrary, extending `Object.prototype` _subverts_ static typing. [Here's an example in TypeScript](https://www.typescriptlang.org/play/#src=interface%20MaybeHasFilter%20%7B%0D%0A%20%20%20%20filter%3F%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Afunction%20test(o%3A%20MaybeHasFilter)%3A%20string%20%7B%0D%0A%20%20%20%20return%20o.filter%20%7C%7C%20%22filter%22%3B%0D%0A%7D%0D%0A%0D%0A(%3Cany%3EObject.prototype).filter%20%3D%20function%20()%20%7B%20%7D%3B%0D%0A%0D%0Aconsole.log(typeof%20test(%7B%7D)%20%3D%3D%3D%20'string')%3B%0D%0A). [Here's an example in flow](https://flow.org/try/#0C4TwDgpgBAYglgG2BATlAvFA3gHygM0WRQC4oB5AIwCsIBjYKHAXwG4AoUSWAex42x4ewABaoA7nADOEMlVoMmbdu3wBXAHYM4PDVGRTgACh5l4SVE148AlNnZQocfFBMA6QhZR2sDx69NYIlQbDn8oOl0pHgQINwQeAHMjAHJPYhTQv2YoCAQZe3CTMz4s8MiNaNj4pNThMRRJGUywqGZ2dvYjeXpgNzAUYWFwWSgAQw0QGw9gtEx1LWAdPSMfZU4IQyMseolpUaxmZlCgA). In both cases, the guarantees supposedly given by the type system are invalidated at runtime because of the `Object.prototype` extension.

Note that in both type systems, you have to disable type checking via an `any` cast to actually be allowed to extend `Object.prototype` -- that's the type system telling you it's unsafe.

bcherny:
@jfirebaugh Since it doesn't look like you normally use TypeScript, here's a corrected example: 

```ts
// tell TS about extension
interface Object {
    filter<T>(this: Object, f: (a: T) => boolean): Object
}

// extend prototype
Object.prototype.filter = <T>(this: Object, f: (a: T) => boolean) => this // stub

// use it
function test(o: Object) {
    return o.filter(a => true)
}
```

This is the basis for [ESlib](https://github.com/eslib/lodash), a project I'm working on (see example output [here](https://github.com/eslib/cli)).

In any case, this is not the point I want to argue. The Mapbox library should not assume what I have on my prototype. Some users might have extensions, some might not, but the less assumptions Mapbox needs to make about the environment it's running in the better.

This is the point Douglas Crockford makes in JS the Good Parts:

> The other form (called `for in`) enumerates the property names (or keys) of an object. On each iteration, another property name string from the object is assigned to the variable. It is usually necessary to test `object.hasOwnProperty(variable)` to determine whether the property name is truly a member of the object or was found instead on the prototype chain.

You're using the same `in` operator as is used in a `for in`. When using `in`, it is good practice to add a `hasOwnProperty` check to make sure you don't accidentally pick up properties from the object's prototype. In your case, you can forego the `in` altogether in favor of a `hasOwnProperty` check.

Some more benefits of `hasOwnProperty` over `in`, if you're still unconvinced:

- `hasOwnProperty` is the intended meaning in your case. You should not care what's on the prototype.
- Mapbox currently checks for 'type', 'source', 'source-layer', 'filter', and 'layout' on the prototype chain. If in the future you guys decide to add `get`, `set`, or `toString` (or any other properties that are on Object's prototype chain), you will need the `hasOwnProperty` check anyway.
- It's faster to check own properties instead of traversing up the prototype chain.

bcherny:
I have to say, this negative response is pretty unexpected. I thought this would be a quick fix, and while the response isn't totally sound to me, I do appreciate your guys time spent responding to this issue. Cheers.

Issue #4772, opened 2017-06-01
document mouseenter/mouseleave


peterqliu:
As used in the [hover popup example](https://www.mapbox.com/mapbox-gl-js/example/popup-on-hover/). 

Looks like these are different from the other mouse events, as they pertain to specific style layers, rather than the map object as a whole.

mollymerp:
Oh good catch – it's only mentioned here https://www.mapbox.com/mapbox-gl-js/api/#map#on

If you want to add it it would go at the bottom of `src/ui/map.js` with the other map events https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/map.js#L1718-L1760 😸

jfirebaugh:
Added in #5126.

PR #5126, opened 2017-08-09
More flow

jfirebaugh:
I set out to remove as many `any`s as I could from popup.js and marker.js, and discovered that I needed a `MapMouseEvent` type first. So I reorganized map event types and docs, then did the original task.

What's nice about this organization is the documentation for each map event type is next to its entry in the `MapEvent` enum, so we can see where we're missing documentation (e.g. #4772), or missing enum values.

### DIFF
--- a/src/ui/events.js
+++ b/src/ui/events.js
@@ -162,8 +162,39 @@ export type MapEvent =
      */
     | 'dblclick'
 
+    /**
+     * Fired when a pointing device (usually a mouse) enters a visible portion of a specified layer from
+     * outside that layer or outside the map canvas. This event can only be listened for via the three-argument
+     * version of {@link Map#on}, where the second argument specifies the desired layer.
+     *
+     * @event mouseenter
+     * @memberof Map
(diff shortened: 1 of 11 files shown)


src/ui/events.js
```diff
@@ -0,0 +1,576 @@
...
+    | 'mouseover'
+
+    /**
+     * Fired when a point device (usually a mouse) leaves the map's canvas.
+     *
+     * @event mouseout
+     * @memberof Map
+     * @instance
+     * @property {MapMouseEvent} data
+     * @see [Highlight features under the mouse pointer](https://www.mapbox.com/mapbox-gl-js/example/hover-styles/)
```
mollymerp: hmm this example actually uses the `mouseleave` event, not `mouseout`.  Are we not defining the events on L165-167 because they're self explanatory or for some other reason?

jfirebaugh:
Yeah, I think I'll just go ahead and add docs for `mouseleave` etc., and move the example.

PR #5132, opened 2017-08-10
Eliminate LayerData type

jfirebaugh:
Its non-`ProgramConfiguration` members are better suited as properties of `ProgramConfiguration` itself.


jfirebaugh:
Now eliminates the other `LayerData` type, in `BufferGroup`.

PR #5137, opened 2017-08-11
Yet another Bucket refactor!?

jfirebaugh:
Yeah, I know, I've been refactoring `Bucket` for years. But it needs it!

This series of refactors culminates in the elimination of `ArrayGroup` and `BufferGroup`. These classes were a poor abstraction: they required highly conditional logic to satisfy the needs of all their consumers, and they were tightly coupled to each other. It's better to have each bucket class contain exactly the array/buffer properties it needs, and to keep corresponding arrays and buffers close to each other in code.

This brings the bucket classes much closer to their corresponding gl-native implementations.

benchmark | master 4eb5d98 | bucket-interface 7998c6d
--- | --- | ---
**map-load** | 121 ms  | 111 ms 
**style-load** | 64 ms  | 58 ms 
**buffer** | 983 ms  | 976 ms 
**fps** | 59 fps  | 60 fps 
**frame-duration** | 5.8 ms, 0% > 16ms  | 5.9 ms, 0% > 16ms 
**query-point** | 0.76 ms  | 0.73 ms 
**query-box** | 41.95 ms  | 42.29 ms 
**geojson-setdata-small** | 3 ms  | 2 ms 
**geojson-setdata-large** | 93 ms  | 109 ms


mourner:
Weird, the Circle build log keeps stuck in loading for me...

jfirebaugh:
In debugging the query test failure, I think I found a flow soundness bug! https://github.com/facebook/flow/issues/4580

jfirebaugh:
There's definitely some commonality that deserves extraction, but I want to avoid creating an abstraction until it's clearer that we've got the right one. Some of the issues are:

* `SymbolBucket` is an obvious outlier.
* `FillBucket` is also an outlier -- it maintains two index buffers per vertex buffer.
* I think we'll want to make the flow types involved here generic on the layout attributes, like they are in native.

@kkaefer had some good ideas in this vein in https://github.com/mapbox/mapbox-gl-native/commit/2d36983bbb1131282241bd4d8b5240bf2c86d0ec; see my comments on https://github.com/mapbox/mapbox-gl-native/pull/9468. (These changes were written out of that PR but I've pushed a branch to keep them around.)

PR #5142, opened 2017-08-14
Introduce a Program class

jfirebaugh:
Bring another concept over from native. Here it improves flow coverage (by eliminating a `{[string]: any}` map) and removes `createVAOs` calls from bucket constructors (slowly reducing duplicated code).

benchmark | master b9d9ef2 | program-class afe8a8e
--- | --- | ---
**map-load** | 120 ms  | 130 ms 
**style-load** | 46 ms  | 54 ms 
**buffer** | 974 ms  | 967 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 5.8 ms, 0% > 16ms  | 6.4 ms, 0% > 16ms 
**query-point** | 0.74 ms  | 0.76 ms 
**query-box** | 40.80 ms  | 43.64 ms 
**geojson-setdata-small** | 2 ms  | 1 ms 
**geojson-setdata-large** | 108 ms  | 111 ms

### DIFF
--- a/src/data/bucket/circle_bucket.js
+++ b/src/data/bucket/circle_bucket.js
@@ -74,7 +74,6 @@ class CircleBucket implements Bucket {
             this.elementBuffer = new Buffer(options.elementArray, ElementArrayType.serialize(), Buffer.BufferType.ELEMENT);
             this.programConfigurations = ProgramConfigurationSet.deserialize(circleInterface, options.layers, options.zoom, options.programConfigurations);
             this.segments = new SegmentVector(options.segments);
-            this.segments.createVAOs(options.layers);
         } else {
             this.layoutVertexArray = new LayoutVertexArrayType();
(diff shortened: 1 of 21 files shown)


mourner:
Is the 5.8ms -> 6.4ms frame duration change a random fluke, or a real 10% perf hit?

jfirebaugh:
It's really hard to tell. If I run it in isolation several times, `program-class` usually seems to take about 0.1 to 0.2 ms longer (not 0.6). But times for `master` tend vary by that much too, and even when I bench master against master, the second run seems to come up slower more often than not.

anandthakker:
I think we should go ahead with this.  Looking at the raw `frame_duration` samples, it does look like there's a real effect here, but it looks more like 0.1-0.2, not 0.6:

screen shot 2017-08-21 at 4 32 33 pm

Issue #2706, opened 2016-06-09
Fix label clipping across tile boundaries


lucaswoj:
Presently, if a source has any of the following properties set to `true`, labels may be clipped at tile boundaries:
- `text-allow-overlap`
- `icon-allow-overlap`
- `text-ignore-placement`
- `icon-ignore-placement`

This behaviour is intended to avoid double-drawing labels on tile boundaries when the feature is in one tile and in an adjacent tile's buffer. This behaviour can also cause a label to be clipped when the feature is not in the adjacent tile's buffer (because the buffer is too small). 

The relevant code is in [`draw_symbol`](https://github.com/mapbox/mapbox-gl-js/blob/master/js/render/draw_symbol.js#L13-L14)

I don't have any specific solutions in mind yet.

cc @ansis @yhahn @noemiwalzebuck @mourner

mourner:
Could we still do label collision for those symbols but only with tile boundaries in the collision tree?

lucaswoj:
Very good idea @mourner 😄. That approach would be a nice follow-up to #2703.

pathmapper:
What are the drawbacks having double-drawed labels? Is there a big impact on the rendering performance? Depending on the drawbacks a double-drawed label (if perfectly drawed one upon the other) might be preferable compared to a clipped one.

ChrisLoer:
> This behaviour can also cause a label to be clipped when the feature is not in the adjacent tile's buffer (because the buffer is too small).

The clipping can also happen whenever labels placed with `symbol-placement: line` get placed at a boundary in one tile but due to symbol-spacing logic don't get placed with the exact same anchor in another tile. This behavior [shows up in our current symbol spacing tests](https://github.com/mapbox/mapbox-gl-js/blob/2313006d8ea76e31e39d6d5f1e638ba094f17b53/test/integration/render-tests/symbol-spacing/line-close/expected.png) (look for the discontinuous line across the bottom).

> What are the drawbacks having double-drawed labels?

The main drawback I'm aware of is that it messes with the ordering of labels that are allowed to overlap. Whatever gets drawn last will be shown "on top" of whatever was drawn earlier. Each tile has an internally consistent ordering of overlapping labels based on the y-position of their anchor (see issue #470).

[...]

Issue #5086, opened 2017-08-02
Reduce density for labels based on distance from camera


nickidlugash:
Continuing the discussion from https://github.com/mapbox/mapbox-gl-js/issues/4704 and from chat:

In pitched maps, labels often appear denser in the background than in the foreground, because more vector tiles' worth of data is being shown in the background for the same number of pixels on the screen. This is the *opposite* of what would typically make sense from an information design perspective: generally, information in the foreground is more important and/or more in need of immediate attention by the viewer. Denser background labels distract from the higher-priority foreground labels:

E.g. Place labels in Mapbox Streets are often denser in the background than in the foreground:

<img width="900" alt="screen shot 2017-08-02 at 4 44 42 pm" src="https://user-images.githubusercontent.com/5607844/28894156-f5ca921e-77a1-11e7-936c-53e372ce5c0e.png">

[Attenuating label size scaling with distance](https://github.com/mapbox/mapbox-gl-js/pull/4547) (i.e. halving the scale factor used to adjust label size based on distance) lessened this issue, because using proportionally larger labels/collision boxes in the background naturally reduces their density due to more labels being hidden through collisions (and conversely, using propertionally smaller labels in the foreground increases their density).

However, reducing density through collisions in this way is mainly helpful for a mid-range of label sizes and densities. If labels are very small and/or sparse (e.g. highway shields in Mapbox Streets, and other icons along lines), background labels don't necessarily face  collisions. If labels are very large and/or dense (e.g. zoom level 16 POI labels in Mapbox Streets for very dense urban areas), there are already a lot of collisions in the foreground so the density ends up pretty even in the foreground and background. 

So far we've discussed two approaches to improving the situation (that don't require redoing anchor placement when pitch changes):

## 1. For labels along line at a distance, remove every other anchor

See discussions starting from https://github.com/mapbox/mapbox-gl-js/issues/4704#issuecomment-319192573. 

One advantage of pursuing the reduction of labels along a line specifically (instead or in addition to a more general approach) is that this just reduces density of *duplicate* labels, rather than hiding potentially useful labels that only appear once. (Of course this only helps with labels along lines, though).

Reposting @ChrisLoer's [first results into using this approach](https://github.com/mapbox/mapbox-gl-js/issues/4704#issuecomment-319485157), which look promising:

> Here's the status quo in a pitched view:

[...]

Issue #5038, opened 2017-07-24
Variable text placement for point labels


nickidlugash:
The purpose of this is ticket is to discuss and document our experimentation with variable text placement for point labels, and to decide if we want to implement this in Mapbox GL as an alternative option to fixed label placements.

Generally, variable label placement is the ability for the renderer to assign different placements for a label in order to improve the likelihood that the label will be visible. 

Implementing [viewport label placement](https://github.com/mapbox/mapbox-gl-js/pull/4972) will unblock this ability. (Caveat: we can't implement this for api-gl, as cross-tile placements would not be predictable.)

Example placement workflow:

For each label:

1. Try Placement Option 1.
2. If using Option 1 will cause the label to not be placed (because it collides with an already placed label), try Placement Option 2.
3. And so on through all specified placement options.
4. If last placement option does not allow label to be placed, do not place label and move on to next label. 

# Benefits 

Allowing variable label placements should result in label visibility that more closely matches a map designer's intent, as defined by the zoom level visibility and placement order of features in the data source + stylesheet. Specifically:

### 1. Improved label hierarchy

With a fixed label placement, it's common for a label that was attempted to be placed later (and therefore generally less important) to be visible while a label that was attempted to be placed earlier (and therefore generally more important) to be hidden. This is due to a collision between the label with earlier placement and another label with even earlier placement.

For example:

![better-hierarchy](https://user-images.githubusercontent.com/5607844/28546386-7ebf2474-7098-11e7-9176-3faa210717c5.png)

In this example, Label 1 is placed first. Label 2 is attempted to be placed second, but with the fixed placement, it collides with Label 1 and does not get placed. With a variable placement, using the second placement option (the right side) gives Label 2 space to be placed. Label 3 is attempted to be placed third, and with the fixed placement there is space for it to be placed since Label 2 was not placed. With the variable placement, Label 3 collides with Label 2 and does not get placed. Even though the number of labels visible is the same with fixed and variable placement in this example, variable placement improves the visibility of more important labels. 

### 2. Improved label density

In addition, variable label placement can allow more labels overall to be visible, to more closely match the density of the data.

For example:

![better-density](https://user-images.githubusercontent.com/5607844/28546393-832bd76e-7098-11e7-8a70-3b8f0ea65d19.png)

In this example, a fixed placement allows 2 out of 3 labels to be visible, while a variable placement allows all 3 labels to be visible.

# Spec implementation

[...]

PR #5150, opened 2017-08-15
Viewport collision detection

ChrisLoer:
Switching from tiled label placement to global viewport label placement. https://github.com/mapbox/mapbox-gl-js/issues/4704 WIP.

This PR dramatically changes our approach to collision detection. Instead of trying to pre-calculate collisions across a range of zooms in the background, we now calculate collisions synchronously in the foreground. Fade animations are no longer tied to zoom level, but instead based on time elapsed since collision occurred. This approach helps solve many problems:

 - More reliable collision detection between tiles (buffering/avoid-edges are no longer necessary)
 - Collision detection now works on labels from different sources (e.g. a GeoJSON set of labels added on top of a base map)
 - Calculating collision boxes at (or near) render time is much more accurate for pitched maps, line labels, sizes based on zoom functions, etc.
 - Fade in/out animations are no longer limited to zoom operations (i.e. collisions related to pitch/pan/rotate will all be animated now)

The main structural changes are:

 - `CollisionTile` is now a global object that lives only in the foreground. Insertions/queries into the collision index are now done entirely in viewport coordinates instead of tile coordinates.
 - The "placement" step no longer happens in the background thread, and static buffers for symbols are created during the "prepare" step.
 - The new foreground "placement" step is primarily responsible for updating a dynamic array of opacities that are used to animate collision states. For each vertex, the opacity state is expressed as an "opacity at time of last update" and a "target opacity". The symbol shaders animate towards the "target opacity" based on an animation start time set for each bucket.
 - It's now necessary to place a hard time limit on how long an individual placement call can take, to avoid frame stutter -- so full placement can now be split across multiple frames.
 - Collision features for lines are now represented in the collision grid as an array of circles, instead of as an array of boxes. This makes their behavior more stable under pitch/rotation operations.
 - Because fading is now tied to "collision time" instead of to zoom level, we need a way to tell when two symbols in tiles of two different zoom levels are "the same symbol", so that we can avoid triggering a fade in/fade out when we cross between zoom levels. This is primarily handled by the new `CrossTileSymbolIndex`.

TODO:

- [x] Handle re-ordering for overlapping symbols (requires re-generating static buffers on rotation -- might be time to address issue #2706?)
- [x] Fix remaining test failures/update tests to reflect new behavior
  - [x] Render tests
    - [x] 10 remaining render test failures for symbol-avoid-edges/tile-clipping (current behavior is actually better)/overlap-sorting
  - [x] Query tests
  - [x] Unit tests
- [ ] Performance improvements:
  - [ ] ~~Don't require full placement when a new tile is added -- instead, start progressive placement for the tile as soon as it's added, but don't start rendering until the first full placement has completed.~~
  - [x] Allow progressive placement to pause at the bucket level. (Currently placement pauses at the layer level, where a single dense layer of street labels on a large viewport can take ~8ms on my MacBook Pro)
- [x] Implement anti-aliasing for collision circle debugging (need them to look good for Studio)
- [x] Extend collision detection to a buffer around the viewport (so that animations for changes around the edge can be mostly finished animating before they're actually visible)
- [x] Update comments, rationalize class names (e.g. `CollisionTile`), remove dead code
- [x] Rebase on master
- [x] Flow-ify changes
- [x] Handle wrapped tiles in `CrossTileSymbolIndex`
- [x] Figure out home for `grid_index_experimental` (adds support for circle geometry queries; removes support for cross-thread serialization): either update https://github.com/mapbox/grid-index, make a new library, or just continue to include the code directly.
- [ ] Port to `gl-native`
  - [ ] Figure out how to continue doing tile-based collision detection in `api-gl`

This PR will enable further changes that we're not tackling yet:

 - Collision detection for labels with rotation-alignment map.
 - `text-offset` applied to collision detection
 - Adding collision features with arbitrary geometry
 - Cross-fading for layout property changes
 - Multiple label placement (https://github.com/mapbox/mapbox-gl-js/issues/5038 / https://github.com/mapbox/mapbox-gl-js/tree/multiplace)
 - Label density reduction: #5086
 - Reduce data usage by creating vector tiles without collision buffers
 - Fade out symbols from removed tiles (requires holding on to tile or at least symbol buckets over period of fade animation)

## Latest Benchmark

screenshot 2017-09-29 11 15 44

/cc @ansis @nickidlugash @mollymerp


nickidlugash:
@ChrisLoer A few notes based on my visual review of this branch today:

- Overall I think this is looking awesome 💯  Super excited to see this improved behavior on all our styles!

- I'm noticing a lot of labels that are initially visible and then disappear while panning. It looks like that is what would be addressed in this TODO item:
  > Extend collision detection to a buffer around the viewport (so that animations for changes around the edge can be mostly finished animating before they're actually visible)

  How will we calculate how big this buffer needs to be? Would we approximate it based on the default collision fade duration, and a reasonable panning speed? Would we be able to make a buffer large enough to resolve this issue for the background in very pitched maps?

- `text-offset` for labels along lines isn't taken to account in collision. Is this going to be part of further work listed?:

  >  Collision detection for labels with rotation-alignment map.

- As I mentioned in chat, might be worth us thinking about whether `collisionFadeDuration` should be a map property (as currently implemented in this branch), or a global style property (or both)? I could see this being a useful property in both.

- Now that you've identified exactly what updates will and won't be part of this initial PR, I'll open a couple more tickets for specific future changes so we can start tracking those discussions separately.

ChrisLoer:
> How will we calculate how big this buffer needs to be? Would we approximate it based on the default collision fade duration, and a reasonable panning speed? Would we be able to make a buffer large enough to resolve this issue for the background in very pitched maps?

These are all good questions, and I don't know the answer. We should experiment with different values to find the best stability/performance tradeoff. I can try to make the padding a configurable value just so we can experiment more quickly.

> text-offset for labels along lines isn't taken to account in collision

Actually, I think that's just a bug -- there are two failing test cases for that too. It's on my list to look into, I've just been mentally filing it under "fix tests".

> As I mentioned in chat, might be worth us thinking about whether collisionFadeDuration should be a map property (as currently implemented in this branch), or a global style property (or both)? I could see this being a useful property in both.

:+1:

ChrisLoer:
OK, the query tests are now passing and the ten remaining render test failures are all caused by the known avoid-edges/tile-clipping/overlapping-symbol-sorting issues. If we removed support for symbol-avoid-edges and stopped doing tile clipping for overlapping symbols, we'd be down to just two failures, both in `icon-translate-anchor`, caused by changing the sort order of overlapped symbols on a rotated tile.

> > text-offset for labels along lines isn't taken to account in collision

> Actually, I think that's just a bug -- there are two failing test cases for that too. It's on my list to look into, I've just been mentally filing it under "fix tests".

I was too optimistic here. The cause of those test failures was a bug in handling collision boxes for single-glyph labels. Actually handling text-offset correctly in collisions for line labels isn't included in this PR, but it should be much more tractable after this PR goes in. I've added it to "further work".

ChrisLoer:
Benchmark results overall look pretty good, but `frame-duration` has gone up. A modest increase in the average may be acceptable, but for the zoom 13 test it caused at least one frame to take longer than 16ms:

master b203a40 | viewport-collision 1cff6c7
--- | ---
7.6 ms, 0% > 16 ms at zoom 13 | 8.4 ms, 1% > 16 ms at zoom 13

benchmark | master b203a40 | viewport-collision 1cff6c7
--- | --- | ---
**map-load** | 168 ms  | 126 ms 
**style-load** | 70 ms  | 67 ms 
**buffer** | 1,002 ms  | 1,007 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 5.3 ms, 0% > 16ms  | 6 ms, 0% > 16ms 
**query-point** | 0.55 ms  | 0.59 ms 
**query-box** | 39.98 ms  | 39.50 ms 
**geojson-setdata-small** | 1 ms  | 2 ms 
**geojson-setdata-large** | 123 ms  | 109 ms

ChrisLoer:
I have a local stash that replaces the "flattened" arrays of collision boxes and collision circles with arrays of javascript objects. It's much easier to read. However, it also seems to be slower. My test methodology was to rotate a pitched full screen map with lots of road labels while recording with the CPU profiler, and then look at the percent of render time spent on the relevant code (looking at ratios is more reproducible than looking at absolute render time):

Before:

 Run| redoPlacement | placeCollisionCircles (self)
--- |--- | ---
First | 18.2% | 8.5% (2.9%)
Second | 16.7% | 7.5% (2.4%)
Third | 16.7% | 7.4% (2.5%)

After:

Run | redoPlacement | placeCollisionCircles (self)
--- |--- | ---
First | 21.0% | 11.5% (3.5%)
Second | 18.6% | 8.5% (2.7%)
Third | 23.4% | 13.5% (4.1%)

I don't have a good explanation for why this makes such a noticeable difference, but it looks like in this particular example it pushes up placement time by at least a millisecond.

ChrisLoer:
I did four more test runs, this time with a build that represented circles/boxes as javascript objects, but with only number data members (i.e. instead of containing a `Point` member, they just contained an `x` and a `y`). Results looked pretty similar -- still a noticeable win for the earlier "flat" approach.

mourner:
@ChrisLoer I also noticed that working with flat number arrays generally gives a huge boost to performance compared to arrays of objects. V8 treats them differently. Earcut and Delaunator both take advantage of this.

ChrisLoer:
D'oh. I found that at least part of the cause of the flattened array being faster was that it had a bug that caused it not to place some collision circles that it should have (of course the bug derived from the awkwardness of the code!).

After fixing the bug, the difference is looking narrower, but the "flattened" approach is still reliably faster...

Before:

 Run| redoPlacement | placeCollisionCircles (self)
--- |--- | ---
First | 22.1% | 11.8% (2.3%)
Second | 21.4% | 11.0% (2.0%)
Third | 20.7% | 11.3% (2.0%)

After:

Run | redoPlacement | placeCollisionCircles (self)
--- |--- | ---
First | 22.7% | 12.7% (4.0%)
Second | 23.9% | 13.5% (4.0%)
Third | 22.6% | 12.8% (2.5%)

ChrisLoer:
@mourner That's good to know, thanks! @kkaefer had theorized that v8 would be able to be clever about recognizing that objects were "plain ol' data" and do the flattening internally, but I guess not, or at least not yet...

ChrisLoer:
@nickidlugash I pushed a change that extends collision detection past the edge of the viewport by 100 pixels in all directions. Unfortunately, there's no getting around that doing collision over a broader area makes it significantly more expensive (the cost scales at least linearly with the number of symbols included in the detection area). I ended up at padding of 100px because the hit to collision times wasn't *that* bad, and it seemed to eliminate most of the notable cases of label instability at the edge of the viewport. Maybe you could evaluate if it's good enough? If we can get away with an even lower value, that'd be great, too (just change the `viewportPadding` constant in `collision_index` to try different values).

ChrisLoer:
@nickidlugash I just pushed a change that draws the collision circles with anti-aliasing, so they should be "ready" for use in Studio. I could use your help figuring out the color strategy though. The current black and blue (with fading to indicate unused collision boxes/circles) works well for me, and I think it looks reasonable. We could go back to green and red, but those are really difficult for my kind to figure out. Overall, I think the decision is a lot easier because we used to have four colors for the various states, but now there are just two colors.

ChrisLoer:
Thanks @anandthakker for making it easy to do a density plot (although I don't have the R skills you have, were you using ggplot?).

Overall, frame durations are looking pretty similar, but there's definitely a bit of a long tail that's been introduced with these changes.

Before:

image

After:

image

ChrisLoer:
I pushed a change to allow placement to pause per-bucket instead of just per-layer (although the pause/restart logic is a bit ugly 😬 ). On my machine even the most expensive/dense buckets don't take much more than a millisecond to place, so we can pretty effectively constrain how long placement takes. I've updated the benchmark at the top of the ticket, along with a density plot showing that the behavior is now looking pretty close to `master`.

The remaining flat-lined long tail represent the frames on which a new tile loads, in which case full placement has to take place (in this run, the longest frame-duration came out to 16.035ms, so that probably would have been a one frame stutter). We can't do a partial placement when new tiles come in because they may be higher or lower-zoom versions of tiles we're already showing and to avoid flicker we need to transfer opacities from the previous zoom level.

What we could potentially do to solve the tile loading problem is:

- Split collision detection apart from updating the opacity buffers
- Add a single-frame "commit" that updates all of the opacity buffers at once whenever a collision detection pass is finished
- Mark new tiles as non-renderable until they've been included in a completed collision detection pass

ChrisLoer:
`reloading` and `expired` tiles present an extra challenge for the "don't render until you've done first placement" strategy, because in the case of reloading tiles, we want to keep rendering the old data until the updated tile is fully placed. That would require us to keep around two copies of the tile data during that period (the old one for rendering, the new one for processing the next placement).

@kkaefer (or @mourner), do you think we can get away with doing a full placement in those two (relatively uncommon) cases?

ChrisLoer:
@ansis When you get back, I'd like to brainstorm with you about how to solve the overlapping-symbol-ordering problem, which I think is the only _big_ thing left on the JS side. Ideally, I'd like to fix #2706 _and_ not reopen #470.

- **current `viewport-collision` implementation**: Sort symbols on y-order using bearing of 0, and then never change them. Don't use tile clipping, only draw symbols whose anchor is in the tile we're currently drawing.
  - Pros:
    - Easy!
    - Fixes #2706.
    - Consistent ordering across all tiles (previously, asynchronous placement meant that tiles could change ordering independently)
  - Cons:
    - Rolls back #470: although ordering is consistent, it doesn't maintain the "higher icons are rendered on top" behavior during rotation
    - Ordering can get messed up at tile edges (a lower symbol in one tile can be drawn over a higher tile in an adjacent one)
- **Re-generate sorted buffers in background**: This would basically be a re-implementation of master's `redoPlacement` behavior, except its only purpose would be to sort buffers
  - Pros:
    - Preserve current behavior
  - Cons:
    - Asynchronous `redoPlacement` adds a ton of complexity to the code that we'd be preserving for just this one relatively minor case.
- **Re-generate sorted buffers in foreground**: We would serialize buffers with enough metadata to make it easy to do a sort/re-generate operation on the foreground at placement time (not more often because we don't want to constantly be re-uploading static buffers).
  - Pros:
    - Preserves `master` behavior, with modest improvement that sorting for all tiles would happen at the same time.
    - Preserves `viewport-collision` simplification of loading symbol tiles
  - Cons:
    - Still can't simultaneously fix #2706 and preserve #470 
  - **Variant 1**: Keep `master`'s tile-clipping behavior, thus ensuring that overlapping symbol order is consistent even at tile boundaries.
  - **Variant 2**: Stop tile-clipping and instead only draw symbols whose anchor is within the current tile. Solves all the problems of #2706, but introduces potential ordering inconsistencies at tile boundaries.
- **Combine and sort symbol buffers on the foreground**: Whenever we're finishing a placement on the foreground, we combine all of the symbol buffers for currently loaded tiles into one sorted buffer with a common coordinate system, and `draw_symbols` would use the unified buffer instead of per-tile buffers.
  - Pros:
    - Fixes #2706 and preserves #470 behavior
    - Reducing GL draw calls should be a performance win
    - Provides a relatively simple way to keep rendering current tiles/symbols while a new placement operation is in progress.
  - Cons:
    - Performance: sorting/transforming buffers could be expensive (this part might be amenable to backgrounding)
    - Performance: requires uploading more static buffers -- any time a tile is added or symbol ordering changes, the entire buffer has to be re-uploaded, vs. current behavior where only new tiles have to be uploaded.
    - Not clear what the right transformed coordinate system would be -- I'm guessing it would be something like "a tile zoomed out far enough to include all the tiles currently in the tree", but for each zoom level we'd lose a bit of precision. We can get three bits of precision back because we wouldn't need tile buffers any more (and what used to be the line normal bit is free). We can probably get away with dropping another couple bits of precision, but at some point we'd risk symbols starting to jump around as the coordinates get transformed.
    - Have to do all the implementation work just to find out if the hypothetical performance issues are a problem!

My inclination is to do "per-tile re-sort on foreground, without tile-clipping". I think the problems in #2706 are overall more severe than the potential for overlapping symbols to have their order messed up at tile boundaries -- but I may not have enough context for why that would be a serious regression. I think "combine and sort buffers on the foreground" is promising, but it would be a big change and this PR is already a lot to digest.

nickidlugash:
>  I pushed a change that extends collision detection past the edge of the viewport by 100 pixels in all directions. Unfortunately, there's no getting around that doing collision over a broader area makes it significantly more expensive (the cost scales at least linearly with the number of symbols included in the detection area). I ended up at padding of 100px because the hit to collision times wasn't that bad, and it seemed to eliminate most of the notable cases of label instability at the edge of the viewport. Maybe you could evaluate if it's good enough? If we can get away with an even lower value, that'd be great, too

@ChrisLoer I reviewed this in one of the areas of the Mapbox Streets style with the most collisions (Manhattan at z16), with a 60deg pitch, with the default collision fade duration of 300ms, and I think it looks good enough 👍  I only tested it with manually panning/zooming/rotating/pitching the map, but for all practical speeds of doing these actions, the map handled this issue pretty well.

I haven't tested camera animations, but as we discussed offline, fly to animations where the zoom level + center is changing rapidly probably won't load new tiles fast enough to be very affected by this (but I guess we should test this?). The nature of animation used for turn-by-turn guidance seems like it is slow enough (and typically uses styles that have sparse enough labels) for this to not be an issue, despite the  extreme pitch typically used. There may be use cases we haven't discussed that need very fast panning on pitched maps, but I think this viewport padding solution is fine to proceed with for now.

If performance is a big concern, we could decrease the padding a bit – a 50px padding was noticeably worse than 100px, but I thought 75px was still pretty reasonable.

ChrisLoer:
@nickidlugash Thanks for taking the time to run those tests! Sounds like your impressions were broadly similar to mine. I think I'll stick with 100px vs 75px for now just to give the padding a little padding. ;)

ChrisLoer:
My latest change tries to be more methodical about choosing an optimal number of cells for the grid-index on the size of the viewport. The results at the end of all the tweaking don't come out all that different on my device/viewport from what I was seeing with the earlier choice of "make a 20x20 grid", but they should be a little more robust to different viewport sizes. I evaluated choices by collecting the average time to complete a full placement (i.e. one placement split across multiple frames) during a 10 second 180 degree rotation in a 60 degree pitched map full of line labels using Nicki's label-placement-debug style (`#13.86/33.9612/-118.3133/0/60`):

screenshot 2017-09-01 10 26 32

Basically anything between 15px and 40px seemed to work pretty similarly, so I went with a 25px cell size.

ChrisLoer:
After rebasing on master and combining with the expressions PR (#4777), the benchmark results look problematic. 😞 

In both density plots, the flatter line is `viewport-collision`.

Both branches seem to have a longer tail than we'd like -- I'm not sure what we expect is causing that on the expressions side (I'm assuming the changes on master are because of #4777, that could be wrong). On the `viewport-collision` side, I know the tail can be long because it's sometimes necessary to synchronously perform a full placement. The tail also seemed to get fatter, and I'm not sure why. I haven't found any obvious regressions in performance otherwise (i.e. average placement time hasn't increased), but I'll poke into that more.

## Before expressions

benchmark | master c48a56b | viewport-collision 67e3ca6
--- | --- | ---
**map-load** | 114 ms  | 82 ms 
**style-load** | 55 ms  | 54 ms 
**buffer** | 1,006 ms  | 977 ms 
**tile_layout_dds** | 962 ms  | n/a 
**fps** | 59 fps  | 59 fps 
**frame-duration** | 5.2 ms, 0% > 16ms  | 5.7 ms, 0% > 16ms 
**query-point** | 0.55 ms  | 0.61 ms 
**query-box** | 41.54 ms  | 42.29 ms 
**geojson-setdata-small** | 1 ms  | 2 ms 
**geojson-setdata-large** | 106 ms  | 52 ms 


image

## After expressions

benchmark | master 8a674b8 | viewport-collision 669c9c3
--- | --- | ---
**map-load** | 87 ms  | 83 ms 
**style-load** | 110 ms  | 108 ms 
**buffer** | 1,052 ms  | 1,083 ms 
**tile_layout_dds** | 1,493 ms  | 1,453 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 6 ms, 0% > 16ms  | 7.1 ms, 4% > 16ms 
**query-point** | 0.83 ms  | 1.03 ms 
**query-box** | 68.71 ms  | 74.89 ms 
**geojson-setdata-small** | 7 ms  | 9 ms 
**geojson-setdata-large** | 141 ms  | 104 ms 
**filter** | n/a  | n/a 

image

/cc @anandthakker

ChrisLoer:
I experimented with trying to cut the time spent on updating the dynamic opacity buffer by reducing 8 uint16s per glyph (opacity + target opacity for each of the four corners) to one uint32 per glyph. I packed the opacity information into one uint8 per vertex, and then I packed all four vertices into one uint32 (and just cast to uint8s when I bound the buffer), so that there'd be fewer assignments/multiplications on the javascript side.

It worked just fine, but it didn't make a noticeable dent in the time spent updating the buffers (watching in the profiler). Also, if making the buffers smaller improved rendering times, it was too small of a difference for me to be able to reliably pick up in the benchmarks.

mourner:
@ChrisLoer me and @anandthakker are currently working on a refactor of expressions code generation system which should make it many times faster (see [PoC here](https://gist.github.com/mourner/88286de4d89dd1306272ce6c50860d06)). Hopefully that'll help with performance.

mourner:
The perf effect is probably due to #5224 (stop functions).

ChrisLoer:
@ansis I'm thinking it might make sense to move the collision circle generation into the foreground too (leaving `CollisionFeature` without much purpose in life). The idea behind the current approach was:

 - Following line geometry and creating circles along the way has lots of steps: best to avoid doing that again and again
 - Instead, store off the tile-unit "distance from anchor" of each (background-generated) circle, and then when we project the line labels, we calculate where the ends are and then reverse that back to a tile-unit distance. Then, we can use that distance to quickly filter which circles should be used in collision detection

The problem with this approach is that "reversing back to tile-unit distance" isn't very well defined for interpolated points in the middle of a segment. The approximation we use ("if it's 20% along the length of the projected segment, treat it as 20% along the tile segment") breaks down for overzoomed tiles (causing the `text-pitch-alignment/viewport-overzoomed` test to fail).

Instead of going to greater lengths to get a correct tile-unit distance, it might be best to just generate the collision circles in viewport space on every run of the collision algorithm:

 - We're already traversing the line geometry to find where the first and last glyph are, but we'll have to build up the array of collision circles along the way.
 - Previously, we had to project every used collision circle from tile to viewport coordinates. Now, we'll only have to project the line vertices (which we're already doing), and we'll be using direction vectors plus interpolation to find the circle coordinates (which seems like it could be a bit faster).
 - We'd have to switch to dynamically updating the layout array for the debug collision circles, since the layout array would change every time placement happened.
 - We'd have to re-implement support for circles in `queryRenderedSymbols`, since its logic currently depends on the `grid-index` pointing back to an underlying (background-generated) `CollisionBoxArray`.

Assuming the performance impact were neutral-to-positive, I think this change would generally make collision circles fit their rendered labels somewhat better/more-predictably.

ChrisLoer:
I revisited the opacity update optimizations I mentioned in https://github.com/mapbox/mapbox-gl-js/pull/5150#issuecomment-327588508. With a couple tweaks and one significant addition (don't update the "hidden" state of symbols in the placedGlyphArray unless the state has actually changed), `commitPlacement` roughly halved its CPU time from 1 to .5% in the 180 degree rotation scenario from https://github.com/mapbox/mapbox-gl-js/pull/5150#issuecomment-326641731.

Although it's a small chunk of CPU time overall, the opacity updates are only being run on frames where a placement is being completed, and those are the problematic frames that end up taking longer than 16ms. On the other hand, I still can't easily detect the effect of these changes in the frame-duration benchmark, so it's very possible this is premature optimization.

To the extent that these changes help, the same types of optimization (switch from using StructArray to flat javascript arrays where possible, reduce number of StructArray insertions) could have a lot more effect in the `updateLineLabels` code (which takes ~25% of CPU time in the same profiling runs). For example, in the profiles, `hideGlyphs` (which is nothing more than effectively nulling a bunch of values in a StructArray) is taking ~7% of CPU time. It's not strictly related to collisions, but optimizing that could buy us efficiency we could spend on collisions.

ChrisLoer:
@ansis and I just synced up on our strategy for getting this PR ready to merge. His immediate next steps will be:

- [ ] Implement "variant 2" of "regenerate sorted symbols in foreground" (from https://github.com/mapbox/mapbox-gl-js/pull/5150#issuecomment-326438377)
- [ ] Do a round of review on the changes so far

My immediate next step is:

 - [ ] Fix overzoomed collision test failure by implementing foreground collision circle generation (as described in https://github.com/mapbox/mapbox-gl-js/pull/5150#issuecomment-327636880)

After that, we'll take a look at:

 - [ ] Revisit `SymbolBucket` serialization/use of `StructArray`. In general, we think the code could be both more readable and faster with more of the data in plain javascript objects/arrays (i.e. `SymbolInstance`). Specifically, `updateLineLabels` loads a lot of data out of `StructArray`s that I suspect could be faster with the same data represented by flat arrays of numbers.
 - [ ] Come up with a firmer plan for `api-gl`. We think it makes sense to merge this PR before the native changes go in, but at a minimum we should have a clear strategy in mind for making the changes work with `api-gl`.
 - [ ] Consider a change to `CrossTileSymbolIndex` to handle the current edge cases in the grid used to detect duplicate symbols between different zoom levels. We're leaning towards ignoring this problem as long as it doesn't seem too noticeable.
 - [ ] Performance: consider optionally using a GL extension that allows us to bind the same data to multiple vertices (to improve cases like dynamic layout and opacity updates, where we write the same data for all four vertices of a glyph).

Due to high complexity, we plan to punt on:

 - Making placement interrupt-able even when tiles have been added or removed. Although this does introduce potential for skipped frames, we hope overall performance improvements will mitigate this concern.
 - Combining overlapping symbols from multiple tiles into a single global bucket. This means that symbol sorting may not always follow y-order right at tile boundaries. We are gambling that this will not be that noticeable, and is a worthwhile trade-off in exchange for closing our current tile-clipping bugs.

ChrisLoer:
I gave up on the foreground/viewport collision circle generation described in https://github.com/mapbox/mapbox-gl-js/pull/5150#issuecomment-327636880. I still think it's probably the eventual solution we should move to, but as I started implementing it the scope of the change kept growing. The biggest fundamental thing I underestimated was that the current line-label projection logic runs in the "label plane", which can be different from the viewport plane -- so there would have to be additional logic to convert interpolated anchor points to the viewport plane.

Instead, I added code to make the estimation of tile-unit "distance from anchor" more accurate. It adds a little bit of extra work but it doesn't look like a big hit in the profiler.

ChrisLoer:
I experimented with moving the "projectionCache" from being created per-symbol to being created per-bucket -- since the cache is indexed into the `lineVertexArray` shared by the entire bucket, and it's possible for multiple symbols to be placed on the same line segment (or line segments that share a vertex). It's a simple enough change, but it doesn't seem to make much of a difference. After seven runs of both versions, placement appeared to be ~1% faster, and a t-test gave a "same distribution" probability of 51%. In theory this would help more in over-zoomed tiles where the lines are longer and straighter, but it doesn't seem like it's worthwhile.

ChrisLoer:
Updating `CrossTileSymbolIndex` to check adjacent grid positions turned out to be easy, and it doesn't seem to make a noticeable performance difference.

You can see an example of the reduced flicker by looking at Portland between zooms 11 and 12 (`#12/45.5162/-122.671/7.2`). "Pearl District" will flicker before the changes.

ChrisLoer:
So I think the fatness of the tail that showed up after the rebase is being caused by `evaluateSizeForFeature`. Here's the density plot with that short circuited to just return the lower covering stop:

image

The tail is still longer than we want, but at least it's nice and thin.

It's nice to be able to evaluate the font size exactly at collision time, because it gives us the most accuracy, but we _could_ go back to the previous behavior where we base the collision box on the font size for `tileZoom + 1`.

ChrisLoer:
I take back my speculation about `evaluateSizeForFeature` -- the clear distinction between the short-circuited version and the unmodified version disappeared after multiple runs of the benchmark.

ChrisLoer:
@mourner / @kkaefer, can either of you help us define the "finish line" in terms of performance for this PR? At this point, average frame duration and "% > 16ms" is looking pretty close to master, although still usually somewhere from .3 to .7ms slower on my machine. The biggest impact shows up around zoom level 13 where line labels are most dense, and we've partially offset the extra collision detection cost with some optimizations of line label projection, which conveniently costs the most in exactly the same situations.

The frame-duration benchmark doesn't really capture the effects of requiring a "full" (non-interruptible) placement every time a tile is added to the render tree (since the benchmark doesn't change transform parameters as it renders). In my "high stress" case (60 degree pitch, dense road network, near full screen), I see the average time for a full placement coming in at 10ms --
 probably enough to to cause a skipped frame every time a tile is added. That sounds bad, and in fact there is noticeable stutter when I rotate the map in this scenario... but there's also noticeable stutter on `master` in the same scenario, so there may not be that much practical difference.

src/source/source_cache.js
```diff
@@ -44,6 +45,9 @@ class SourceCache extends Evented {
     _maxTileCacheSize: ?number;
     _paused: boolean;
     _shouldReloadOnResume: boolean;
+    _needsFullPlacement: boolean;
+    _placementIDs: Array<number>;
+    _placementIndex: number;
```
ansis: I'm wondering if it would be helpful to separate the placement state (and maybe even the placement methods) into a separate class or object. If it is, it could also be done at the style and tile level.

src/source/source_cache.js
```diff
@@ -573,10 +591,12 @@ class SourceCache extends Evented {
         if (tile.uses > 0)
             return;
 
-        tile.stopPlacementThrottler();
+        this._updatePlacement();
+        tile.removed();
 
         if (tile.hasData()) {
-            const wrappedId = tile.coord.wrapped().id;
+            tile.coord = tile.coord.wrapped();
```
ansis: When does this change the value of `tile.coord`? Does this affect anything else later on?

src/source/source_cache.js
```diff
@@ -652,11 +678,29 @@ class SourceCache extends Evented {
         return tileResults;
     }
 
-    redoPlacement() {
+    placeLayer(collisionIndex: CollisionIndex, showCollisionBoxes: boolean, layer: any, posMatrices: any, shouldPausePlacement: any) {
+        while (this._placementIndex < this._placementIDs.length) {
+            const tile = this.getTileByID(this._placementIDs[this._placementIndex]);
+            if (!posMatrices[this._placementIndex]) {
+                posMatrices[this._placementIndex] = collisionIndex.transform.calculatePosMatrix(tile.coord, tile.sourceMaxZoom);
+            }
```
ansis: Would it make sense to move this caching to the inside of `transform.calculatePosMatrix(...)`? not sure

src/source/tile.js
```diff
@@ -73,6 +76,7 @@ class Tile {
     texture: any;
     refreshedUponExpiration: boolean;
     reloadCallback: any;
+    crossTileSymbolIndex: any;
```
ansis: Do tiles need to hold a reference to the `CrossTileSymbolIndex`? I added this because it was the easiest way to get access to it in the added/removed callbacks but it might be good to get rid of this if possible.

src/source/tile.js
```diff
@@ -206,78 +181,58 @@ class Tile {
...
-        this.showCollisionBoxes = source.map.showCollisionBoxes;
-        this.placementSource = source;
+    placeLayer(showCollisionBoxes: boolean, collisionIndex: CollisionIndex, layer: any, posMatrix: Float32Array) {
+        const bucket = this.getBucket(layer);
+        const collisionBoxArray = this.collisionBoxArray;
 
-        this.state = 'reloading';
-        this.placementThrottler.invoke();
+        if (bucket && bucket instanceof SymbolBucket && collisionBoxArray) {
+            collisionIndex.setMatrix(posMatrix);
```
ansis: Maybe we should remove `collisionIndex.setMatrix` and instead pass the matrix as an argument to the relevant method.

src/source/tile.js
```diff
@@ -206,78 +181,58 @@ class Tile {
...
+                PlaceSymbols.updateOpacities(bucket, collisionFadeTimes);
+                bucket.sortFeatures(angle);
             }
-        }, this.workerID);
+        }
+
+        // Don't update the collision index used for queryRenderedFeatures
+        // until all layers have been updated to the same state
+        if (this.featureIndex) {
+            this.featureIndex.setCollisionIndex(collisionIndex);
```
ansis: Does this need to be set once for every tile? Or should this be done at the style level?

src/source/source_cache.js
ChrisLoer: The tile is being added to the cache with the wrapped ID. If we don't modify the tile's ID to match the wrapped ID, then when the tile is pulled out of the cache, the ID stored in the tile won't match the ID that was used to fetch it. This messes up the `CrossTileSymbolIndex` logic, which is based on the ID stored in the tile.

The sequence is something like this (I'm going from memory here). Imagine tile A and its wrapped copy tile B.

 - Tile B added
 - Tile B removed (now in cache with Tile A's id)
 - Tile A added (fetches tile B from cache). CrossTileSymbolIndex adds entries for Tile B.
 - Tile B added. CrossTileSymbolIndex doesn't do anything because tile B is already added.
 - Tile A removed. CrossTileSymbolIndex removes entries for Tile B, even though they're still relevant for Tile B. Unpredictable label madness ensues.

src/source/tile.js
ChrisLoer: `SourceCache` is already holding a reference to the map (and thus the painter and the `CrossTileSymbolIndex`), so we could pass this in as an argument to `added` and `removed`. I'll do that.

src/source/source_cache.js
ChrisLoer: I like that idea, I'll give it a try.

src/source/source_cache.js
ChrisLoer: That sounds like a good idea, but I don't have a clear idea of where it would be best to draw the boundaries. Could you push an implementation?

src/source/tile.js
ChrisLoer: Yeah, sounds good.

src/source/tile.js
ChrisLoer: Are you thinking you would pass the appropriate `CollisionIndex` in as a parameter to `FeatureIndex#query`? The chain of calls for `queryRenderedFeatures/Symbols` is somewhat long and confusing, and I feel like it might be kind of hard to keep track of what the "right" collision index is.

src/source/source_cache.js
ChrisLoer: Done.

src/source/tile.js
ChrisLoer: Done.

src/source/tile.js
ChrisLoer: Done.

src/symbol/place_symbols.js
```diff
@@ -0,0 +1,260 @@
...
+        if (opacityState.targetOpacity !== targetOpacity) {
+            collisionFadeTimes.latestStart = opacityUpdateTime;
+        }
+        const increment = collisionFadeTimes.duration ? ((opacityUpdateTime - opacityState.time) / collisionFadeTimes.duration) : 1;
+        opacityState.opacity = Math.max(0, Math.min(1, opacityState.opacity + (opacityState.targetOpacity === 1 ? increment : -increment)));
+        opacityState.targetOpacity = targetOpacity;
+        opacityState.time = opacityUpdateTime;
+    }
+    return (!initialVisible && !initialTargetVisible && opacityState.targetOpacity !== 0) ||
+           ((initialVisible || initialTargetVisible) && (opacityState.opacity === 0 && opacityState.targetOpacity === 0));
```
ansis: I think this is *right* but it could be clearer as
```js
const initialVisible = opacityState.opacity !== 0 && opacityState.targetOpacity !== 0;
const nowVisible = opacityState.opacity !== 0 && opacityState.targetOpacity !== 0;
return initialVisible !== nowVisible
```

also, the return value is only used for text. Should it be moved out of this function and just around the text?

src/symbol/place_symbols.js
```diff
@@ -0,0 +1,260 @@
...
+    } else if (opacityState.opacity === 1 && opacityState.targetOpacity === 1) {
+        return 4294967295;
+    }
+    const targetBit = opacityState.targetOpacity === 1 ? 1 : 0;
+    const opacityBits = Math.floor(opacityState.opacity * 127);
+    return opacityBits * shift25 + targetBit * shift24 +
+        opacityBits * shift17 + targetBit * shift16 +
+        opacityBits * shift9 + targetBit * shift8 +
+        opacityBits * shift1 + targetBit;
+}
```
ansis: what's the advantage of the multiplication instead of using instead of the bitshift operator?

src/symbol/cross_tile_symbol_index.js
```diff
@@ -0,0 +1,249 @@
...
+            symbolInstance.isDuplicate = false;
+            // If we don't pick up an opacity from our parent or child tiles
+            // Reset so that symbols in cached tiles fade in the same
+            // way as freshly loaded tiles
+            symbolInstance.textOpacityState = new OpacityState();
+            symbolInstance.iconOpacityState = new OpacityState();
+        }
+    }
+
+    getScaledCoordinates(symbolInstance: SymbolInstance, coord: TileCoord, z: number): any {
```
ansis: A comment explaining the relevant coordinate spaces would be great.

Also, inlining this in `get` might be useful since it's short and only used there.

src/symbol/cross_tile_symbol_index.js
```diff
@@ -0,0 +1,249 @@
...
+        const scale = Math.pow(2, Math.min(this.sourceMaxZoom, z) - Math.min(this.sourceMaxZoom, coord.z)) * roundingFactor;
+        const anchor = symbolInstance.anchor;
+        return {
+            x: Math.floor((coord.x * EXTENT + anchor.x) * scale),
+            y: Math.floor((coord.y * EXTENT + anchor.y) * scale)
+        };
+    }
+
+    get(otherTileSymbol: SymbolInstance, coord: TileCoord) {
```
ansis: any ideas for a better name for this method?

src/symbol/collision_index.js
```diff
@@ -0,0 +1,346 @@
...
+            /*flip*/ false,
+            labelPlaneAnchorPoint,
+            tileUnitAnchorPoint,
+            symbol,
+            lineVertexArray,
+            labelPlaneMatrix,
+            projectionCache,
+            /*return tile distance*/ true);
+
+        if (projectedAnchor.perspectiveRatio < .2 || !firstAndLastGlyph) {
```
ansis: 0.2 comes from just trying things out and seeing what works?

src/symbol/collision_index.js
```diff
@@ -0,0 +1,346 @@
...
+            const projectedPoint = this.projectPoint(posMatrix, collisionCircles[k], collisionCircles[k + 1]);
+            const x = projectedPoint.x;
+            const y = projectedPoint.y;
+
+            const radius = tileUnitRadius / tileToViewportScaled;
+
+            if (placedCollisionCircles.length) {
+                const dx = x - placedCollisionCircles[placedCollisionCircles.length - 4];
+                const dy = y - placedCollisionCircles[placedCollisionCircles.length - 3];
+                if (radius * radius * 2 > dx * dx + dy * dy) {
```
ansis: This is `if (radius * sqrt(2) > distance(prev, current)` right? Is the `sqrt(2)` just an experimentally derived value?

src/symbol/collision_index.js
ChrisLoer: Err... actually this is mostly left over from an earlier version of `approximateTileDistance` that had more trouble with this (overzoomed) situation. It's still true, I think, that anything with a perspective ratio of less than .2 is anchored way off screen, but it's probably not an important performance optimization so it's probably better just to remove the logic.

src/symbol/collision_index.js
ChrisLoer: Yeah, I ended up at this number with just a few rough experiments looking at how the circle density turned out. We could probably come up with a more theoretically optimal number.

src/symbol/cross_tile_symbol_index.js
ChrisLoer: `getMatchingSymbol`?

src/symbol/cross_tile_symbol_index.js
ChrisLoer: I'll add a comment. `getScaledCoordinates` is also used in the constructor when it's populating the index.

I didn't really see this as a performance hot spot, but it occurs to me it's called frequently enough maybe it makes sense to use bitshifting to generate the scale factor instead of `Math.pow`

src/symbol/place_symbols.js
ChrisLoer: The problem was that the high bits could be interpreted as negative numbers, and that threw off the addition logic.

src/symbol/cross_tile_symbol_index.js
ansis: I was thinking about maybe getting a tiny improvement in clarity not performance by inlining but it's probably not significant

src/symbol/place_symbols.js
ChrisLoer: "visible" would be `opacity != 0 || targetOpacity != 0`, but I changed it instead to just `initialHidden` and `nowHidden`, and moved them out to the test label call as you suggested.

src/symbol/cross_tile_symbol_index.js
ChrisLoer: Comments added. I also switched to bit shifting, which probably wasn't necessary but forced me to think a little more explicitly about the order of look-up (that is: we look up "child" symbols in the "parent" `TileLayerIndex`).

src/symbol/collision_index.js
ChrisLoer: I just removed the perspectiveRatio logic, and rolled the `firstAndLastGlyph` logic in with the later tile distance check. This means we do some extra work projecting collision circles for line labels way outside the viewport, but I don't think that was ever too big a deal (and if we really wanted to optimize for that case, I think I'd base it directly on the anchor position relative to the viewport rather than the perspective ratio).

src/symbol/collision_index.js
ChrisLoer: I updated this section with comments and hopefully more self-explanatory variable names, and I revisited the √2 value experimentally. Using a value of 1 is significantly slower, and by the time I upped the value to √3, I started to see missed collision detections (usually with orthogonal streets that "passed through" each other).

ChrisLoer:
@anandthakker I did a few runs of the benchmarks on Firefox 52, and `viewport-collision` actually came out faster on the paint benchmark (although not significantly, ~.3-.5 sd). It's an interesting data point, although I'm not sure what conclusions to draw. Safari looked pretty similar to Chrome.

ChrisLoer:
Porting to native made me realize there's some confusing and unnecessary double-scaling going on in this PR. We use `pixels_to_tile_units.js` to get a pixel ratio which is `EXTENT / (tileSize * scale)`, but then we almost always multiply the result by the tile's `scale`. The reason we're essentially ignoring scale is that we're dealing with text, whose size is specified in pixels, not tile units (and perspective effects are handled separately). We only need to include the scale when we're working with tile geometry (like line shapes).

I started work on a change to get rid of the extra division/multiplication, but it triggered a test failure in `text-size/composite-function-line-placement`. I think the trigger was just a tiny change from precision limitations, but it looks like there's a real problem underneath. I'll try to get a fix in tomorrow.

src/symbol/collision_index.js
```diff
@@ -0,0 +1,353 @@
...
+            const y1 = blocking.y1 / tileToViewport + projectedPoint.point.y;
+            const x2 = blocking.x2 / tileToViewport + projectedPoint.point.x;
+            const y2 = blocking.y2 / tileToViewport + projectedPoint.point.y;
+            const bbox = [
+                new Point(x1, y1),
+                new Point(x2, y1),
+                new Point(x2, y2),
+                new Point(x1, y2)
+            ];
+            if (!intersectionTests.polygonIntersectsPolygon(query, bbox)) continue;
```
anandthakker: Is this the only place where we use the `x1, y1, x2, y2` (box) data for the along-line-labels/collision-circle case?  If we added circle-aware handling here, could we then split `collisionBoxArray` into two separate arrays, one for boxes and one for (sets of) circles?

src/data/bucket/symbol_bucket.js
```diff
@@ -58,22 +41,47 @@ type SymbolBucketParameters = BucketParameters & {
...
 
-type SymbolInstance = {
+export type SingleCollisionBox = {
+    x1: number;
+    y1: number;
+    x2: number;
+    y2: number;
+    anchorPointX: number;
+    anchorPointY: number;
+};
```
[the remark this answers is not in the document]
ChrisLoer: The `offsetX, offsetY` in `CollisionBox` is vestigial, I'll remove that. It was previously used for line collision boxes (it referred to how far the box was from the anchor, not to the `text-offset` property), but is unnecessary with foreground placement.

src/data/bucket/symbol_bucket.js
```diff
@@ -294,31 +341,34 @@ class SymbolBuffers {
  *
  * 2. WorkerTile asynchronously requests from the main thread all of the glyphs
  *    and icons needed (by this bucket and any others). When glyphs and icons
- *    have been received, the WorkerTile creates a CollisionTile and invokes:
+ *    have been received, the WorkerTile creates a CollisionIndex and invokes:
  *
- * 3. SymbolBucket#prepare(stacks, icons) to perform text shaping and layout,
- *    populating `this.symbolInstances` and `this.collisionBoxArray`.
+ * 3. PrepareSymbols#prepare(bucket, stacks, icons) perform texts shaping and
```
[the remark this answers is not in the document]
ChrisLoer: Yeah, that seems reasonable, along with an analogous `symbol_placement.js` and `performSymbolPlacement()`?

src/data/bucket/symbol_bucket.js
```diff
@@ -903,7 +637,8 @@ class SymbolBucket implements Bucket {
             addVertex(layoutVertexArray, labelAnchor.x, labelAnchor.y, bl.x, y + bl.y, tex.x, tex.y + tex.h, sizeVertex);
             addVertex(layoutVertexArray, labelAnchor.x, labelAnchor.y, br.x, y + br.y, tex.x + tex.w, tex.y + tex.h, sizeVertex);
 
-            addDynamicAttributes(dynamicLayoutVertexArray, labelAnchor, 0, placementZoom);
+            addDynamicAttributes(dynamicLayoutVertexArray, labelAnchor, 0);
+            arrays.opacityVertexArray.emplaceBack(0);
```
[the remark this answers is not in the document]
ChrisLoer: Yes, but... this is actually an unnecessary step, so I'll just remove it instead of adding an explanatory comment.

src/symbol/collision_box.js
```diff
@@ -13,16 +13,11 @@ export type CollisionBox = {
...
     featureIndex: number,
     sourceLayerIndex: number,
     bucketIndex: number,
-    bbox0: number,
-    bbox1: number,
-    bbox2: number,
-    bbox3: number,
-    placementScale: number
+    radius: number,
+    distanceToAnchor: number
```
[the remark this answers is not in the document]
ChrisLoer: :+1:

src/symbol/collision_box.js
```diff
@@ -31,31 +26,10 @@ export type CollisionBox = {
...
- * https://cloud.githubusercontent.com/assets/1421652/8060094/4d975f76-0e91-11e5-84b1-4edeb30a5875.png
- * slightly higher zoom:
- * https://cloud.githubusercontent.com/assets/1421652/8060061/26ae1c38-0e91-11e5-8c5a-9f380bf29f0a.png
- * In the zoomed in image the two grey boxes on either side don't cover the
- * label anymore. Their maxScale is smaller than the current scale.
- *
+ * Line labels are represented by a set of these boxes spaced out along the
+ * line. When we calculate collision geometries, we use the circle inscribed
+ * in the box, rather than the box itself. This makes collision detection more
+ * stable during rotation.
```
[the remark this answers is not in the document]
ChrisLoer: :+1:

src/symbol/collision_index.js
```diff
@@ -0,0 +1,353 @@
...
+        ignoredGrid: Grid = new Grid(transform.width + 2 * viewportPadding, transform.height + 2 * viewportPadding, 25)
+    ) {
+        this.transform = transform;
+
+        this.grid = grid;
+        this.ignoredGrid = ignoredGrid;
+        this.pitchfactor = Math.cos(transform._pitch) * transform.cameraToCenterDistance;
+    }
+
+    placeCollisionBox(collisionBox: SingleCollisionBox, allowOverlap: boolean, textPixelRatio: number, posMatrix: mat4): Array<number> {
```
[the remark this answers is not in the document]
ChrisLoer: Yeah, that name is left over from the previous behavior. How about instead of "placeX" which returns true if the item _can_ be placed, we name the functions "collisionTextX", and return true if there's a collision?

src/symbol/collision_index.js
```diff
@@ -0,0 +1,353 @@
...
+                return [];
+            }
+        }
+        return [tlX, tlY, brX, brY];
+    }
+
+    approximateTileDistance(tileDistance: any, lastSegmentAngle: number, pixelsToTileUnits: number, cameraToAnchorDistance: number, pitchWithMap: boolean): number {
+        // This is a quick and dirty solution for chosing which collision circles to use (since collision circles are
+        // laid out in tile units). Ideally, I think we should generate collision circles on the fly in viewport coordinates
+        // at the time we do collision detection.
```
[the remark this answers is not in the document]
ChrisLoer: Added link to https://github.com/mapbox/mapbox-gl-js/issues/5474

src/symbol/collision_index.js
```diff
@@ -0,0 +1,353 @@
...
+                (boxDistanceToAnchor < -firstTileDistance) ||
+                (boxDistanceToAnchor > lastTileDistance)) {
+                // The label either doesn't fit on its line or we
+                // don't need to use this circle because the label
+                // doesn't extend this far. Either way, mark the circle unused.
+                markCollisionCircleUsed(collisionCircles, k, false);
+                continue;
+            }
+
+            const projectedPoint = this.projectPoint(posMatrix, collisionCircles[k], collisionCircles[k + 1]);
```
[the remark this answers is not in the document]
ChrisLoer: Yup, good suggestion.

src/symbol/collision_index.js
```diff
@@ -0,0 +1,353 @@
...
+    /**
+     * Because the geometries in the CollisionIndex are an approximation of the shape of
+     * symbols on the map, we use the CollisionIndex to look up the symbol part of
+     * `queryRenderedFeatures`. Non-symbol features are looked up tile-by-tile, and
+     * historically collisions were handled per-tile.
+     *
+     * For this reason, `queryRenderedSymbols` still takes tile coordinate inputs and
+     * converts them back to viewport coordinates. The change to a viewport coordinate
+     * CollisionIndex means it's now possible to re-design queryRenderedSymbols to
+     * run entirely in viewport coordinates, saving unnecessary conversions.
```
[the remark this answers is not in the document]
ChrisLoer: Ticketed in https://github.com/mapbox/mapbox-gl-js/issues/5475

src/symbol/collision_index.js
ChrisLoer: Yes, we could, although I don't think having the extra bit of unused data in the `collisionBoxArray` is much of a performance concern. If we switch to foreground circle generation (#5474), then the `collisionBoxArray` as a whole will become almost entirely vestigial (it would really just hold indices for `queryRenderedSymbols`). At that point I think it would make sense to just remove it entirely.

PR #5154, opened 2017-08-16
Remove fast-stable-stringify

### LINKED ISSUE #5152: dependency: fast-stable-stringify
Hi,

currently mapbox-gl-js has the dependency fast-stable-stringify, which is licensed under a GPL-3 license. This is problematic for corporate developers (like me), as we are not allowed to build apps with copyleft based licenses.

Do you think that it, would it be possible to get rid of that dependency?

mourner:
Closes #5152 by implementing a simpler key generation for layer specifications. Input is pre-validated so we don't have to worry about circular links or weird objects, and the output doesn't have to be valid JSON, just good enough to be used as a unique key.

## Launch Checklist

 - [x] ~~write tests for all new functionality~~ covered by `group_by_layout.test.js`
 - [x] post benchmark scores

### DIFF
--- a/src/style-spec/group_by_layout.js
+++ b/src/style-spec/group_by_layout.js
@@ -1,11 +1,34 @@
 
-const refProperties = require('./util/ref_properties'),
-    stringify = require('fast-stable-stringify');
+const refProperties = require('./util/ref_properties');
 
-function key(layer) {
-    return stringify(refProperties.map((k) => {
-        return layer[k];
-    }));
+function stringify(obj) {
+    const type = typeof obj;
+    if (type === 'number' || type === 'string' || obj === undefined || obj === null)
+        return JSON.stringify(obj);
+
+    if (Array.isArray(obj)) {
+        let str = '[';
+        for (const val of obj) {
+            str += `${stringify(val)},`;
+        }
+        return `${str}]`;
+    }
+
+    const keys = Object.keys(obj).sort();
+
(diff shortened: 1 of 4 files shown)


src/style-spec/group_by_layout.js
```diff
@@ -1,11 +1,34 @@
...
-    stringify = require('fast-stable-stringify');
+const refProperties = require('./util/ref_properties');
 
-function key(layer) {
-    return stringify(refProperties.map((k) => {
-        return layer[k];
-    }));
+function stringify(obj) {
+    const type = typeof obj;
+    if (type === 'number' || type === 'string' || type === 'undefined')
```
jfirebaugh: Also `type === null`.

mourner:
benchmark | master 12aa05d | remove-stable-stringify a9c76c6
--- | --- | ---
**map-load** | 228 ms  | 269 ms 
**style-load** | 38 ms  | 44 ms 
**buffer** | 628 ms  | 592 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 7.7 ms, 0% > 16ms  | 7.5 ms, 0% > 16ms 
**query-point** | 0.37 ms  | 0.38 ms 
**query-box** | 24.68 ms  | 27.03 ms 
**geojson-setdata-small** | 1 ms  | 1 ms 
**geojson-setdata-large** | 65 ms  | 65 ms

PR #5155, opened 2017-08-16
Copy canvas data into intermediary ImageData buffer

### LINKED ISSUE #4262: Canvas source errors when larger than some environment-specific size
per @kronick 

> It seems like canvases over a certain size (which varies by browser, but as small as 256x256 in Chrome) are failing and causing the entire map layer to go white-- this doesn't happen consistently and @lbud wasn't seeing the same issues so my guess is that it has something to do with video memory or the way the texture is created from the canvas

We're going to create some minimal test cases and do some research into known limitations today.

lbud:
I believe this fixes #4262: I've confirmed with a few colleagues who have affected GPUs that multiple test cases were fixed with this build.

I suspect affected GPUs have a bug in using an `HTMLCanvasElement` as the `pixels` argument in `texImage2D`/`texSubImage2D`. I haven't been able to corroborate this theory anywhere on the internet, unfortunately.

This PR circumvents that by copying canvas data into an intermediary `ImageData` object. It does necessitate the addition of a `CanvasSource#contextType` parameter.

## Launch Checklist

### DIFF
--- a/src/source/canvas_source.js
+++ b/src/source/canvas_source.js
@@ -118,7 +118,7 @@ class CanvasSource extends ImageSource {
             const gl = this.context;
             const data = new Uint8Array(this.width * this.height * 4);
             gl.readPixels(0, 0, this.width, this.height, gl.RGBA, gl.UNSIGNED_BYTE, data);
-            if (!this.secondaryContext) this.secondaryContext = document.createElement('canvas').getContext('2d');
+            if (!this.secondaryContext) this.secondaryContext = window.document.createElement('canvas').getContext('2d');
             const imageData = this.secondaryContext.createImageData(this.width, this.height);
             imageData.data.set(data);
             return imageData;
--- a/src/source/canvas_source.js
+++ b/src/source/canvas_source.js
@@ -22,7 +22,8 @@ import type Evented from '../util/evented';
  *        [-76.52, 39.18],
  *        [-76.52, 39.17],
  *        [-76.54, 39.17]
- *    ]
+ *    ],
+ *    contextType: '2d'
  * });
  *
  * // update
@@ -40,6 +41,7 @@ class CanvasSource extends ImageSource {
     options: CanvasSourceSpecification;
     animate: boolean;
     canvas: HTMLCanvasElement;
+    context: (CanvasRenderingContext2D | WebGLRenderingContext);
     width: number;
     height: number;
     play: () => void;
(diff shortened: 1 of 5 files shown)


src/source/canvas_source.js
```diff
@@ -101,6 +106,21 @@ class CanvasSource extends ImageSource {
...
+        // texImage2D/texSubImage2D (in ImageSource#_prepareImage), but for
+        // some reason this breaks the map on certain GPUs (see #4262).
+
+        if (this.context instanceof CanvasRenderingContext2D) {
+            return this.context.getImageData(0, 0, this.width, this.height);
+        } else if (this.context instanceof WebGLRenderingContext) {
+            const gl = this.context;
+            const data = new Uint8Array(this.width * this.height * 4);
+            gl.readPixels(0, 0, this.width, this.height, gl.RGBA, gl.UNSIGNED_BYTE, data);
+            return new window.ImageData(new Uint8ClampedArray(data), this.width, this.height);
```
jfirebaugh: [According to MDN](https://developer.mozilla.org/en-US/docs/Web/API/ImageData), the `ImageData` constructor is not available on IE (and Safari support is unknown).

One possible way around this would be to return `{width: number, height: number, data: Uint8ClampedArray}` (which `ImageData` will also satisfy), and have `_prepareImage` accept that.

src/source/canvas_source.js
```diff
@@ -22,7 +22,8 @@ import type Evented from '../util/evented';
  *        [-76.52, 39.18],
  *        [-76.52, 39.17],
  *        [-76.54, 39.17]
- *    ]
+ *    ],
+ *    contextType: '2d'
```
jfirebaugh: If we have to make a breaking change here, maybe it's better to replace the `canvas` property with a `context` property of type `CanvasRenderingContext2D | WebGLRenderingContext`. On the other hand, that would be a property that's not JSON-serializable...

lbud: Intuitively I think I prefer the JSON-serializable string, and feel okay with the string enum options since it must match the [string](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext#Parameters) that was used to access the drawing context in the first place 🤔 though I'm happy to defer to stronger opinions…

src/source/canvas_source.js
lbud: I started going down this path but ran into complications with different `texImage2D`/`texSubImage2D` signatures and with flow 😕 instead I went with using `CanvasRenderingContext2D.createImageData`, which is much cleaner but requires creating an extra canvas context for webgl-based canvas sources…

PR #5156, opened 2017-08-16
Fix drag pan/rotate on Safari

jfirebaugh:
They were broken since 65afc08ef02f5eb2081c2e3ee83f23e61722ac14; see https://github.com/mapbox/mapbox-gl-js/pull/5088#discussion_r133591435 for details.

### DIFF
--- a/src/ui/handler/drag_rotate.js
+++ b/src/ui/handler/drag_rotate.js
@@ -224,8 +224,7 @@ class DragRotateHandler {
         if (e.touches) {
             return (e.touches.length > 1);
         } else {
-            const buttons = (e.ctrlKey ? 1 : 2),  // ? ctrl+left button : right button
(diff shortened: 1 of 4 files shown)


src/ui/handler/drag_rotate.js
```diff
@@ -94,7 +94,7 @@ class DragRotateHandler {
         if (this._ignoreEvent(e)) return;
         if (this.isActive()) return;
 
-        window.document.addEventListener('mousemove', this._onMove);
+        window.document.addEventListener('mousemove', this._onMove, {capture: true});
```
jfirebaugh: This was necessary for the "DragRotateHandler prevents mousemove events from firing during a drag" test to pass.

jfirebaugh: Ugh, somehow it breaks mouseup in a real browser though.

PR #5165, opened 2017-08-17
Specialize linting in test/integration and docs/ directories

anandthakker:


docs/.eslintrc
```diff
@@ -0,0 +1,13 @@
+{
+    "parser": "espree",
```
jfirebaugh: Ah, so that's the key!

Issue #5129, opened 2017-08-10
icon-offset scales in integer steps


stevage:
**mapbox-gl-js version**: 0.37.0

### Steps to Trigger Behavior

I have an icon image which is a pin (where the bottom end of the pin represents the precise location). It gets bigger as you zoom in. I'm trying to line up a text label so the label always sits at the bottom end of the pin.

The relevant part of the pin style looks like this:

```
    layout: {
        'icon-image': 'Markers%20for%20Mapbox-01',
        'icon-size': {
            stops: [
                [13, 1 * pinScale],
                [17, 3 * pinScale]
            ]
        },
        'icon-allow-overlap': true,
        'icon-offset': { stops: [
            [13, [0, -30]],
            [17, [0, -90]]
        ]}
    },

```

And the label:

```
    layout: {
        'text-size': 16,
        'text-font': ['CoM Regular','Arial Unicode MS Regular'],
        'text-field': '{Title}',
        'text-anchor': 'left'
   }
```

Observe what happens when I zoom in or out across an integer zoom boundary:

<img width="525" alt="screenshot 2017-08-10 14 51 49" src="https://user-images.githubusercontent.com/678344/29154964-e97b6a34-7ddb-11e7-8f9e-244bdbaa7ed2.png">

<img width="523" alt="screenshot 2017-08-10 14 52 01" src="https://user-images.githubusercontent.com/678344/29154966-ebc74948-7ddb-11e7-88f4-b18e95d5df2a.png">

![faces-marker](https://user-images.githubusercontent.com/678344/29155051-8d85cb6a-7ddc-11e7-967c-5d0ed0ea2bd5.gif)

### Expected Behavior

Smooth transitioning, so the offset is scaling at exactly the same rate as the icon is getting bigger, so the illusion of the label being attached to a specific point on the icon works.

### Actual Behavior

The offset seems to jump in whole integer amounts, leading to the icon jumping around distractingly.

The same thing happens with `text-offset`

kristfal:
Just to chime in: Our current workaround is to create a buffer at the bottom of the marker so the actual center of the sprite matches the tip of the marker. This isn't really optimal, and I'd say an `icon-anchor` property that behaves similar to `text-anchor` would be the best solution as it wouldn't require any intricate scale / offset matching in the style to make this behavior work.

stevage:
Thanks. In this particular case that's awkward to implement as the marker asset is loaded dynamically from an asset library that I don't control.

[...]

PR #5183, opened 2017-08-23
Proposal: add `icon-anchor` property

lbud:
This PR adds `icon-anchor`, which behaves exactly like `text-anchor`. There is no existing ticket on adding `icon-anchor` 🙊 but I think it's useful in doing things like aligning SDF pins/markers to point to their source Point rather than always being centered/requiring extra space. Incidentally, this was also discussed recently in https://github.com/mapbox/mapbox-gl-js/issues/5129#issuecomment-321502768. The implementation is very straightforward as it just reuses `text-anchor` code.

If everyone is 👍 on this I'll work on a native PR as well.
## Launch Checklist

 - [x] mapbox-gl-native PR https://github.com/mapbox/mapbox-gl-native/pull/9849
 - [x] update SDK support tables


lbud:
> Just curious, why is the branch called iconOffset

🤦‍♀️ just a typo 😬 @ChrisLoer — this was totally unrelated to https://github.com/mapbox/mapbox-gl-js/issues/5129, have not touched that.

> Is there any case where we'd want something like an "auto" value for icon-anchor that followed the value of `text-anchor`? I'm thinking of a situation where you're using `icon-text-fit` to use an icon as background for a label.

Good question — @nickidlugash? I'd be inclined to say no as it seems a bit prescriptive about combining text and icons — when using an icon as a background for a label you'd want both to use the same anchor, but for e.g. placing a label underneath an icon (in the y direction rather than z) you could use `icon-anchor: bottom` and `text-anchor: top` (this could potentially be an improvement on the way I assume designers currently do it, using `-offset` properties).

nickidlugash:
Awesome! 🎉 

> Is there any case where we'd want something like an "auto" value for icon-anchor that followed the value of text-anchor? I'm thinking of a situation where you're using icon-text-fit to use an icon as background for a label.
>> Good question — @nickidlugash? I'd be inclined to say no as it seems a bit prescriptive about combining text and icons

I think I would agree with @lbud here –  icon+text position relationships are varied enough across use cases that I think it's better to stay agnostic. I think an `auto` value is most useful in the spec when there is a clear case for it to be the default value, which I don't think there is here.

Issue #702, opened 2014-08-15
Coalesce texture uploads at the beginning of a frame


kkaefer:
See mapbox/mapbox-gl-native#423.

PR #5184, opened 2017-08-23
Introduce an explicit "upload" step to rendering

jfirebaugh:
In this step, all pending vertex and index buffers are uploaded, rather than them being uploaded in the course of rendering. This brings gl-js closer to the native behavior, and in theory is more efficient.

benchmark | master b203a40 | upload e386312
--- | --- | ---
**map-load** | 221 ms  | 105 ms 
**style-load** | 85 ms  | 84 ms 
**buffer** | 1,033 ms  | 1,022 ms 
**fps** | 59 fps  | 60 fps 
**frame-duration** | 5.3 ms, 0% > 16ms  | 5 ms, 0% > 16ms 
**query-point** | 0.68 ms  | 0.69 ms 
**query-box** | 46.08 ms  | 41.86 ms 
**geojson-setdata-small** | 3 ms  | 4 ms 
**geojson-setdata-large** | 129 ms  | 111 ms 

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/gl/index_buffer.js
+++ b/src/gl/index_buffer.js
@@ -1,35 +1,38 @@
 // @flow
 
-import type {SerializedStructArray} from '../util/struct_array';
+import type {TriangleIndexArray, LineIndexArray} from '../data/index_array_type';
 
 class IndexBuffer {
-    arrayBuffer: ?ArrayBuffer;
-    length: number;
     gl: WebGLRenderingContext;
-    buffer: ?WebGLBuffer;
+    buffer: WebGLBuffer;
 
-    constructor(array: SerializedStructArray) {
-        this.arrayBuffer = array.arrayBuffer;
-        this.length = array.length;
+    constructor(gl: WebGLRenderingContext, array: TriangleIndexArray | LineIndexArray) {
+        this.gl = gl;
+        this.buffer = gl.createBuffer();
+
+        // The bound index buffer is part of vertex array object state. We don't want to
+        // modify whatever VAO happens to be currently bound, so make sure the default
+        // vertex array provided by the context is bound instead.
+        if (gl.extVertexArrayObject === undefined) {
+            (gl: any).extVertexArrayObject = gl.getExtension("OES_vertex_array_object");
+        }
+        if (gl.extVertexArrayObject) {
+            (gl: any).extVertexArrayObject.bindVertexArrayOES(null);
+        }
+
+        gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, this.buffer);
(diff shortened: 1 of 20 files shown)


jfirebaugh:
Fixes #702.

src/gl/index_buffer.js
```diff
@@ -1,35 +1,35 @@
...
-
-            // dump array buffer once it's bound to gl
-            this.arrayBuffer = null;
-        } else {
-            gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, this.buffer);
+        if (gl.extVertexArrayObject === undefined) {
+            (gl: any).extVertexArrayObject = gl.getExtension("OES_vertex_array_object");
+        }
+        if (gl.extVertexArrayObject) {
+            (gl: any).extVertexArrayObject.bindVertexArrayOES(null);
```
anandthakker: Why was this `extVertexArrayObject` stuff not necessary in this class before this change?  (I haven't had occasion to learn what "VAO"s are all about)

jfirebaugh: Vertex array objects are essentially a mechanism to group together some pieces of OpenGL state such that that state can be switched more efficiently, all at once, by changing the bound VAO, rather than piece by piece. One of the pieces of VAO state is the bound element (index) buffer. That means that if you call `gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, ...)`, you're _modifying the currently bound VAO_. There's no way to bind an element buffer without modifying some VAO.

One way to expose this reality would be to expose element buffer binding only through a wrapper method on our `VertexArrayObject` class, and never call `gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, ...)` directly. But this is awkward because when we're creating the index buffer in the first place, there's not an obvious single `VertexArrayObject` instance available to use. (In fact, when the number of vertices being indexed exceeds `UINT16_MAX`, the created index buffer will eventually be bound in _multiple_ VAOs, one for each segment.) And we certainly don't want to mutate whatever VAO happens to currently be bound. So instead, we make sure we're modifying a special "global" VAO -- in OpenGL, this is done by binding VAO ID 0, in WebGL by binding `null`.

Where the VAO extension isn't available, you can think of it as an implementation where this "global" VAO is the only VAO that can exist -- and that's in fact [how we unify the API](https://github.com/mapbox/mapbox-gl-native/blob/6cbeeff6898211d3c3955e34503358e8b2ee127b/src/mbgl/gl/context.cpp#L304-L306) in native.

I'll add a comment here to explain.

jfirebaugh: Oh, to answer why wasn't this necessary before: because we created index buffers on the fly, when (one of) the relevant VAOs was guaranteed to be bound.

PR #5185, opened 2017-08-23
Fix redoPlacement race condition

### LINKED ISSUE #3700: Uncaught TypeError: Cannot read property 'length' of null (CollisionTile in Tile.reloadSymbolData)

mourner:
Closes #3700. The condition occurred when a tile requested redoPlacement (which sent a redoPlacement request to the worker) and then shortly got removed (before getting response to redoPlacement back from the worker)  — `stopPlacementThrottler` did not prevent redoPlacement from running in this case because it was past the throttling phase.

### DIFF
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -235,6 +235,8 @@ class Tile {
             cameraToTileDistance: this.cameraToTileDistance,
             showCollisionBoxes: this.showCollisionBoxes
         }, (_, data) => {
+            if (this.state !== 'reloading') return;
+
             this.state = 'loaded';
             this.reloadSymbolData(data, this.placementSource.map.style);
             this.placementSource.fire('data', {tile: this, coord: this.coord, dataType: 'source'});
@@ -242,6 +244,7 @@ class Tile {
             if (this.placementSource.map) this.placementSource.map.painter.tileExtentVAO.vao = null;
 
             if (this.redoWhenDone) {
+                this.state = 'reloading';
                 this.redoWhenDone = false;
                 this._immediateRedoPlacement();
             }


mourner:
Not sure if it's worth adding a test for this tricky condition if the placement code is fully rewritten in the `viewport-collision` branch anyway.

ChrisLoer:
I think this change would turn the `redoWhenDone` functionality into a no-op. Looking back even before the throttling changes went in, the behavior used to be that the tile state would be 'loaded', but another placement request would happen. I think that was a bug -- the state should have always been switched to 'reloading'.

Did the throttling changes cause #3700, or did they just change the timing so that it was easier to expose this condition?

mourner:
@ChrisLoer fixed it to add the `reloading` state on `redoWhenDone`. If you think we'll do one more release before the viewport labeling PR is finalized, this is worth merging, otherwise we should close to reduce merge collisions. Not sure what caused the regression in the first place but throttling could be related.

Issue #4113, opened 2017-02-01
Add expression operator testing whether a string contains a substring


lucaswoj:
_From @vicapow on April 5, 2016 22:46_

This would be useful for showing query results as the user types.

_Copied from original issue: mapbox/mapbox-gl-style-spec#438_

lucaswoj:
Closely related to #233

1ec5:
Now that expressions have landed and can be used as filters, this is actually a request to add an expression function. Specifically, there should be a function that tests whether a string contains another string. By analogy, the most likely syntax would be `["match", needleString, haystackString]`. However, the word “match” is more closely associated with wildcards and regular expressions, which are covered by #4089.

arethasamuel:
Hi When will this feature be ready? autocomplete search is one of our main use cases, is there an alternative way of doing a "contains" search?

andrewharvey:
@arethasamuel if your data isn't too big you could keep an external list of the query properties from your features and filter based on the full list of names which match the user input.

anandthakker:
@arethasamuel the best place to track this at the moment is via the discussion about regexp expressions happening at https://github.com/mapbox/mapbox-gl-js/pull/6228.  As you can see from that conversation, we're still hashing out the right API design, so there's no concrete timeline for when this functionality will be available.

pshavr:
so
when substr?)

dwohlfahrt:
I'll second @pshavr 's request for a substring function and associated expression syntax!

1ec5:
A substring test expression would make it easy to implement [these road maps color-coded by street name suffix](https://erdavis.com/2019/07/27/the-beautiful-hidden-logic-of-cities/) in GL.

ryanhamley:
>A substring test expression would make it easy to implement these road maps color-coded by street name suffix in GL.

Here's an example of the color-coded road maps mentioned by @1ec5 
https://jsfiddle.net/rsh412/Ldn39w0x/4/

Issue #4089, opened 2017-02-01
Support "regex" filters


lucaswoj:
_From @ansis on January 15, 2015 20:32_

We've talked about this before but it was never implemented. How would it be specified?

Carto uses `=~`. We could also use `regex`.

_Copied from original issue: mapbox/mapbox-gl-style-spec#233_

lucaswoj:
_From @divya1c on March 15, 2016 21:26_

Hi! is the regex feature added to mapbox gl yet?

lucaswoj:
_From @tmcw on March 15, 2016 21:33_

If the issue is open, the task isn't done yet.

lucaswoj:
_From @1ec5 on July 10, 2016 22:53_

`regex` (or `like`) would be more discoverable/memorable, since `~` has wildly different meanings in every language.

lucaswoj:
_From @tmcw on November 22, 2016 21:6_

To unpack what's necessary to implement this feature:

1. Naming - `regex` or `~`
2. Compatibility across GL JS and GL Native

1 will likely be 10% of the work or less. Problem number 2 is more complex.

If the JavaScript port (GL JS) uses [JavaScript's built-in RegExp object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp), then GL Native will need to include a compatible implementation of regular expressions in order to ensure that maps render exactly the same in a native environment. There are many particular flavors of regular expressions, so picking a feature-filled Native regular expression engine would mean that styles break in less feature-filled GL JS RegExp implementations.

There's also the question of whether GL Native should use _platform-provided_ regular expression libraries, or bring its own on the C++ level. From diving into my handy V8 source checkout (get your own today! they're great), V8's implementation is [Irregexp](https://blog.chromium.org/2009/02/irregexp-google-chromes-new-regexp.html)).

lucaswoj:
_From @1ec5 on November 22, 2016 21:59_

@jfirebaugh points out that we can use [`std::regex`’s ECMAScript regex support](http://en.cppreference.com/w/cpp/regex/ecmascript) in gl-native to ensure compatibility between GL JS and the native SDKs. However, note that `std::regex` supports some features that browsers don’t, like `[:alnum:]`.

[...]

Issue #4490, opened 2017-03-24
Custom filter functions


indus:
## Motivation

I want to visualize a big dataset that should get filtered depending on the records in an array property. The array has a variable number of timestamps that all belong to the very same geometry (the variable size of the array is the reason why I can´t split it to seperate key value pairs ```timestamp1, timestamp2, ... , timestampN```).
With the current filter-functions it is not possible to get this kind of filtering done.
I already did a monky-hack to to give it a test and it works great (see "design" + disabling style validation ;-).
But I would wish this would come to mainline.

## Design Alternatives

> Is  "do nothing" an option?

That is always an option. But I could think others would like custom filters as well. Some uses and posibilities I can imagine:
-  using ```Math``` functions; e.g. ```Math.abs``` using the value as a filter but the sign determins the paint color (a lot of this is possible by preprocessing the data; but if I wouln´t have to I wouldn´t do)
- using string-functions like ```substring```, ```lowercase```, ```charAt```, ....
- topographic relations (e.g. point features within a certain sector and distance from a center point; imagine a radar monitor)

## Design

allow a new type of filter where the operator is ```custom``` and the value is a function-body as a string. Use the function-body directly in the [createFilter](https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/feature_filter/index.js#L16) function like so:

``` JS
function createFilter(filter) {
   var fnBody = filter[0]=="custom"? filter[1] : `var p = (f && f.properties || {}); return ${compile(filter)}`
   return new Function('f', fnBody );
}
```

### Mock-Up

``` JS
{ //layer
id: "myLyr",
...
filter: ["custom","for(var i=0,d=JSON.parse(f.properties.d),l=d.length;i<l;i++) if( d[i]>1490338653308 )return true"],
...
}

// using it to set a "dynamic" timebased filter

map.setFilter('myLyr',["custom","for(var i=0,d=JSON.parse(f.properties.d),l=d.length;i<l;i++) if( d[i]>"+(+new Date)+")return true]"
// feature gets visible when at least one of its timestamps is in the future"

```

### Concepts
> How will we teach this design?

a short addition in docs. The only tricky aspect is that you have to know and trust blindly that ```f``` is there and it is the feature

> What terminology will work best for the new concepts introduced by this design?

The name of the operator should be one of those (if there has to be one):
- ```["custom", "<fnBody>"]```
- ```["fn", "<fnBody>"]```

[...]

PR #5193, opened 2017-08-26
Support expressions in filters

### LINKED ISSUE #4077: Allow arbitrary expressions in functions and filters

### LINKED ISSUE #4410: Allow null as a filter value
mapbox-gl-js: 0.33.1
When I tried to filter out null values like this:
```
filter: [
  "!=",
  "name",
  null
]
```
But  it warns:
```
string, number, or boolean expected, null found
```

anandthakker:
Closes #4077
Closes #4410

- Converts classic filters to expressions under the hood
 - Also accepts `{expression: [...]}`, requiring that the given
 expression be of type Boolean
 - [ ] Benchmark to assess performance impact acainst classic filters
 - [ ] Add fixtures for `validation.test.js`
 - [ ] Manually test debug page
 - [ ] Add some kind of partial string matching? ( https://github.com/mapbox/mapbox-gl-js/issues/4089 / https://github.com/mapbox/mapbox-gl-js/issues/4113 )

 cc @jfirebaugh @aparlato

### DIFF
--- a/bench/benchmarks/filter_evaluate.js
+++ b/bench/benchmarks/filter_evaluate.js
@@ -39,7 +39,7 @@ module.exports = class FilterEvaluate extends Benchmark {
         for (const layer of this.layers) {
             for (const filter of layer.filters) {
                 for (const feature of layer.features) {
-                    if (typeof filter(feature) !== 'boolean') {
+                    if (typeof filter({zoom: 0}, feature) !== 'boolean') {
                         assert(false, 'Expected boolean result from filter');
                     }
                 }
--- a/src/data/bucket/circle_bucket.js
+++ b/src/data/bucket/circle_bucket.js
@@ -80,7 +80,7 @@ class CircleBucket implements Bucket {
 
     populate(features: Array<IndexedFeature>, options: PopulateParameters) {
         for (const {feature, index, sourceLayerIndex} of features) {
-            if (this.layers[0].filter(feature)) {
+            if (this.layers[0]._featureFilter({zoom: this.zoom}, feature)) {
                 const geometry = loadGeometry(feature);
                 this.addFeature(feature, geometry);
                 options.featureIndex.insert(feature, geometry, index, sourceLayerIndex, this.index);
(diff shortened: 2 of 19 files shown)


indus:
Will there be an expression for "hasOwnProperty". I'm living with a patched version since #4490

anandthakker:
@indus Good question - there is a `["has", "propertyName"]` expression (see https://github.com/mapbox/mapbox-gl-js/pull/4777), but we should also add `[ "contains" , ... ]` for checking presence in an array.

anandthakker:
@aparlato @nickidlugash updated with:
 - support for `[zoom]`
 - `[contains]`

Still working on performance but it should be usable for developing styles.

anandthakker:
Rebased onto master and reworked to take advantage of #5234 -- performance is much better, but evaluation performance is still about 8x compared to non-expressions-based filters:

screen shot 2017-09-28 at 3 52 25 pm

And creation performance is much slower:
screen shot 2017-09-28 at 3 52 10 pm

anandthakker:
Along the lines of #5379 , @mourner suggests leaving the existing filter implementation in place, and making expressions support additive / opt-in.

anandthakker:
Back to being a bit worse after rebasing onto the recent updates on master (no-eval, etc.):

Bench | master | expressions-filter | gist
------ | -- | -- | --
FilterCreate | 0.303 ms | 9.81 ms ( +9.51 ms / 15.1 std devs ) | https://bl.ocks.org/anonymous/raw/f98f920d15e3df367eb5175987548939/ 
FilterEvaluate | 0.136 ms | 2.93 ms ( +2.79 ms / 37.3 std devs ) | https://bl.ocks.org/anonymous/raw/b42d635f2eda75b0627f95801985d7ba/

anandthakker:
cc @mourner @jfirebaugh I was hoping the recent changes would mean we could stick with converting old-style filters, but looks like we should drop the conversion for now.

jfirebaugh:
https://github.com/mapbox/mapbox-gl-js/compare/expressions-filter-perf gets it down to 3.5-4x slower.

jfirebaugh:
I merged the alternative without conversion, #5434. We'll hopefully be resurrecting the conversion code from this PR at some point, but closing for now.

PR #5199, opened 2017-08-28
prevent double draw in render-tests

asheemmamoowala:
Move text and camera off tile boundaries to prevent double drawing of symbols.

In support of passing tests in gl-native [port](https://github.com/mapbox/mapbox-gl-native/pull/9870)

> Chris Loer @asheemmamoowala The double-drawing is actually a larger underlying bug that I think we should address, but for now with the test, the right approach might be to just make it irrelevant by shifting the labels so they’re not near a tile boundary and thus don’t get double-drawn


ChrisLoer:
It would be interesting to know why the double-drawing _isn't_ happening in gl-native -- I thought the logic was the same.

Issue #3320, opened 2016-10-07
Fix z-fighting on overlapping fills within the same layer


tgecho:
**mapbox-gl-js version**: 0.25.1
### Steps to Trigger Behavior
1. Add a geojson source with overlapping features and a fill layer
2. Adjust the pitch of the map
### Expected Behavior

The fill layers should stay smooth
### Actual Behavior

The fill layers flicker as the pitch changes and looks jagged when it's not moving. This happens in at least Chrome, Firefox and Safari

https://jsfiddle.net/tgecho/brhb9wng/
![image](https://cloud.githubusercontent.com/assets/49615/19197612/c9de3d7c-8c88-11e6-9642-d219c6996fa8.png)

jfirebaugh:
This looks like the GL JS equivalent of https://github.com/mapbox/mapbox-gl-native/issues/6604.

snkashis:
Yes, also happening for me in 0.26.0 with Chrome 54.0.2840.71.They are jagged during moves and also while static.

gena:
I'm experiencing the same when the map is tilted: [![screen capture](https://img.youtube.com/vi/k5lWRefq96M/0.jpg)](https://www.youtube.com/watch?v=k5lWRefq96M) 

The map is here: [test map](https://api.mapbox.com/styles/v1/gennadiy/ciuwljd1d015u2hpbmoluv1qg.html?fresh=true&title=true&access_token=pk.eyJ1IjoiZ2VubmFkaXkiLCJhIjoiMm9WN3VWQSJ9.b9s_EXvxcqiAbaf5GzrEnA#3.21/-57.60/55.12/-36/60). The water polygons are based on the standard layers.

nextstopsun:
Same issue here.

Do mapbox base maps also have this problem on overlapping landcover features?

<img width="679" alt="screen shot 2017-06-01 at 10 40 33" src="https://cloud.githubusercontent.com/assets/966321/26669548/fc06e7d4-46b6-11e7-93a6-bc276ff117a0.png">

kermeat:
![image pasted at 2017-6-2 23-50](https://cloud.githubusercontent.com/assets/5646358/26744411/7da1b386-47ee-11e7-853a-c07d93c92915.png)

nextstopsun:
Any news here? Everything looks great until you pitch the map. I hope feature rendering order won't become random when this gets fixed.

jfirebaugh:
[...]

PR #5206, opened 2017-08-30
Fix z-fighting of translucent fills

jfirebaugh:
Partial fix for #3320, as described in https://github.com/mapbox/mapbox-gl-js/issues/3320#issuecomment-325845228.


src/render/draw_fill.js
```diff
@@ -26,6 +26,7 @@ function drawFill(painter: Painter, sourceCache: SourceCache, layer: FillStyleLa
         // Once we switch to earcut drawing we can pull most of the WebGL setup
         // outside of this coords loop.
         painter.setDepthSublayer(1);
+        painter.depthMask(painter.isOpaquePass);
```
kkaefer: good catch!

PR #5209, opened 2017-08-30
Corrected lat/lng syntax on the Zoom-to Linestring example

joshuapowell:
## Changes
When using this example online, I got hung up and kept getting an error about my `LatLngBounds` not being correct.

I got the example to work once I changed the example code to use `coordinates[0], coordinates[1]` instead of using the longitude twice.

## Tests
Didn't write any tests ... are there tests that I missed?


mourner:
The example on https://www.mapbox.com/mapbox-gl-js/example/zoomto-linestring/ works without issues. Can you make a minimal test case that demonstrates the issue your seeing? The current code for this example is correct — it's not using to longitude values, it instead starts the reduce with an initial value of a single-point (zero-area) bounds.

PR #5210, opened 2017-08-30
Update style of collision debug geometries

nickidlugash:
@ChrisLoer This updates colors of collision debug geometries, and stroke width of collision debug circles:

screen shot 2017-08-30 at 3 30 11 pm

- Geometries for non-visible labels are red, geometries for visible labels are blue (but lower opacity than the red). For legibility, especially in areas with a lot of collisions, I think it's best for the visible labels to have much lighter/more subtle collision geometries than non-visible labels.

- I reduced the stroke width of collision circles from `5.0` to `3.0`, which I think works best for non-retina screens. `2.0` would look better on retina screens, but would be too illegible on non-retina. I tried different values in the `smoothstep` function, as you suggested, but didn't seem to get better results. I think it may just be hard to find values that work equally well for retina and non-retina.

Please review! Does this work for your 👀  ?

### DIFF
--- a/src/shaders/collision_box.fragment.glsl
+++ b/src/shaders/collision_box.fragment.glsl
@@ -9,7 +9,7 @@ void main() {
     // Red = collision, hide label
     gl_FragColor = vec4(1.0, 0.0, 0.0, 1.0) * alpha;
 
-    // Blue = collision, label is showing
+    // Blue = no collision, label is showing
     if (v_placed > 0.5) {
         gl_FragColor = vec4(0.0, 0.0, 1.0, 0.5) * alpha;
     }
--- a/src/shaders/collision_box.fragment.glsl
+++ b/src/shaders/collision_box.fragment.glsl
@@ -6,12 +6,12 @@ void main() {
 
     float alpha = 0.5;
 
-    // Blue = collision, hide label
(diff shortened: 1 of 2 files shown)


src/shaders/collision_box.fragment.glsl
```diff
@@ -6,16 +6,16 @@ void main() {
 
     float alpha = 0.5;
 
-    // Blue = collision, hide label
-    gl_FragColor = vec4(0.0, 0.0, 1.0, 1.0) * alpha;
+    // Red = collision, hide label
+    gl_FragColor = vec4(1.0, 0.0, 0.0, 1.0) * alpha;
 
-    // Black = collision, label is showing
+    // Blue = collision, label is showing
```
ChrisLoer: Comment should say "no collision", right?

nickidlugash: Yep

nickidlugash:
@ChrisLoer 👍 

> The tests that use debug collision circles will need to have their expected.png regenerated. If you want to push the commits straight to viewport-collision, I don't mind taking care of that.

Thank you! :bowing_woman:

Btw, you had mentioned that collision boxes are now never unused, so it looks like there is some extraneous code in `collision_box.fragment` related to this. I didn't want to accidentally strip out necessary code, so I didn't touch it.

PR #5216, opened 2017-08-31
Shuffle seed support (replaces 'shuffle-array' with 'shuffle-seed')

brunoabinader:
Based on @jfirebaugh's proposal for simplifying test re-runs: By using the same shuffling seed, we'll end up with the same quasi-random sequence.

See example below:

1. First run:
```
$ node test/render.test.js --shuffle
* shuffle seed: vIqRzf4MNX
* passed icon-text-fit height-padding
* passed icon-pitch-alignment auto-rotation-alignment-map
* passed text-offset literal-multiline-anchorcenter-justifyright-offsetnegative
...
```

2. Re-run the same sequence: 
```
$ node test/render.test.js --shuffle --seed=vIqRzf4MNX
* shuffle seed: vIqRzf4MNX
* passed icon-text-fit height-padding
* passed icon-pitch-alignment auto-rotation-alignment-map
* passed text-offset literal-multiline-anchorcenter-justifyright-offsetnegative
```

Caveats:
- The sequence stays the same as long as the amount of tests remains the same.
- Only works when shuffling: `--shuffle` needs to be specified for `--seed=<seed>` to be used.
- If `--shuffle` is specified but not `--seed` (as example 1), a random seed is generated.


test/integration/package.json
```diff
@@ -5,14 +5,13 @@
...
-    "mapbox-gl-styles": "2.0.2",
     "colors": "^1.1.2",
     "d3-queue": "^3.0.3",
     "diff": "^3.0.0",
     "lodash": "^4.17.4",
+    "mapbox-gl-styles": "2.0.2",
     "pixelmatch": "^4.0.2",
     "pngjs": "^3.0.0",
-    "st": "^1.1.0",
-    "shuffle-array": "^1.0.1"
```
jfirebaugh: The `shuffle-seed` dependency needs to be duplicated here, due to how we manage dependencies in native.

PR #5221, opened 2017-08-31
Separate fill-extrusion segments if necessary

### LINKED ISSUE #5214: Fill-extrusion breaks komplex polygons at some zoom levels
We use fill-extrusion layers to visualize overlapping geojson polygons. At some zoom levels komplex polygons get completely broken. This does not happen using fill layers. The polygons itself have no errors (like self intersection).

**mapbox-gl-js version**: 39.1

### Steps to Trigger Behavior

 1. Open https://jsfiddle.net/gtm6jdka/4/
 2. Wait ca 30 sec for the polygons to load.
 3. The polygon is broken at the starting zoom level.

### Expected Behavior

 Non-broken polygons.
 Open https://jsfiddle.net/gtm6jdka/4/ and zoom a few levels in and out to see the non-broken polygons.

### Actual Behavior
 
 See https://jsfiddle.net/gtm6jdka/4/

lbud:
This PR
* Separates fill-extrusion vertex/element creation into two separate loops: once for the footprint and once for the walls. In this way a segment needs to prepare at most `numVertices`, rather than, as it was previously, `numVertices * 5`, and then additional segments can be prepared as needed during wall creation. (I think we _could_ be really tricky about this and try to keep everything in the same loop by maintaining separate cursors for the footprint and walls, but this seems too much unnecessary complication.)
* Adds a warning to `prepareSegment` when a segment of length > 65535 is requested.

Fixes #5214.

## Launch Checklist

### DIFF
--- a/src/data/bucket/fill_extrusion_bucket.js
+++ b/src/data/bucket/fill_extrusion_bucket.js
@@ -1,6 +1,6 @@
 // @flow
 
-const {SegmentVector} = require('../segment');
+const {SegmentVector, MAX_VERTEX_ARRAY_LENGTH} = require('../segment');
 const VertexBuffer = require('../../gl/vertex_buffer');
 const IndexBuffer = require('../../gl/index_buffer');
 const {ProgramConfigurationSet} = require('../program_configuration');
@@ -129,7 +129,7 @@ class FillExtrusionBucket implements Bucket {
                 numVertices += ring.length;
             }
 
-            const segment = this.segments.prepareSegment(numVertices * 5, this.layoutVertexArray, this.indexArray);
+            let segment = this.segments.prepareSegment(numVertices, this.layoutVertexArray, this.indexArray);
 
             const flattened = [];
             const holeIndices = [];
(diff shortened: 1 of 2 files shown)


src/data/bucket/fill_extrusion_bucket.js
```diff
@@ -173,10 +166,34 @@ class FillExtrusionBucket implements Bucket {
                             segment.primitiveLength += 2;
                         }
                     }
+                }
+            }
+
+            if (segment.vertexLength + numVertices > MAX_VERTEX_ARRAY_LENGTH) {
+                segment = this.segments.prepareSegment(4, this.layoutVertexArray, this.indexArray);
```
jfirebaugh: `4` → `numVertices`

src/data/bucket/fill_extrusion_bucket.js
```diff
@@ -173,10 +166,34 @@ class FillExtrusionBucket implements Bucket {
...
+
+                if (ring !== polygon[0]) {
+                    holeIndices.push(flattened.length / 2);
+                }
+
+                for (let i = 0; i < ring.length; i++) {
+                    const p = ring[i];
+
+                    addVertex(this.layoutVertexArray, p.x, p.y, 0, 0, 1, 1, 0);
+                    indices.push(segment.vertexLength++);
```
jfirebaugh: You can avoid accumulating the `indices` array by caching `const startingIndex = segment.vertexLength` prior to the `ring of polygon` loop, and then calculating indexes like `startingIndex + triangleIndices[j]` after earcut. (See `FillBucket` for example.)

src/data/bucket/fill_extrusion_bucket.js
```diff
@@ -127,28 +127,21 @@ class FillExtrusionBucket implements Bucket {
...
 
-                if (ring !== polygon[0]) {
-                    holeIndices.push(flattened.length / 2);
-                }
-
                 let edgeDistance = 0;
 
                 for (let p = 0; p < ring.length; p++) {
-                    const p1 = ring[p];
+                    if (segment.vertexLength + 4 > MAX_VERTEX_ARRAY_LENGTH) {
```
jfirebaugh: Should this be inside the `p >= 1` and `!isBoundaryEdge(p1, p2)` conditionals, so we only prepare a new segment if's definitely needed?

PR #5240, opened 2017-09-05
attach onerror callback to the image used for ajax#getImage

kturney:
I ran into there being no error callback while trying to map#loadImage a url to an svg instead of a png.

The error message could probably be better, and I'm not sure how y'all feel about attaching the original error as the domErr property.

The `getImage, invalid image` test is currently failing with
```
Error: Fake server request processing threw exception: URL.createObjectURL is not a function
      at TypeError: URL.createObjectURL is not a function
      at     at exports.getArrayBuffer (src/util/ajax.js:136:53)
      at     at FakeXMLHttpRequest.xhr.onload (src/util/ajax.js:94:13)
      at     at Array.forEach (<anonymous>)
      at     at Object.window.server.respondWith.request [as response] (test/unit/util/ajax.test.js:100:21)
      at     at Array.forEach (<anonymous>)
      at     at Test.t.test (test/unit/util/ajax.test.js:107:23)
```

so I'm not sure why that wouldn't be there.

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -125,6 +125,11 @@ exports.getImage = function(requestParameters: RequestParameters, callback: Call
                 callback(null, img);
                 URL.revokeObjectURL(img.src);
             };
+            img.onerror = function(domErr) {
+                const err = new Error('failed to load image. make sure the image is a png');
+                (err: any).domErr = domErr;
+                callback(err);
+            };
             const blob: Blob = new window.Blob([new Uint8Array(imgData.data)], { type: 'image/png' });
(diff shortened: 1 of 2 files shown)


jfirebaugh:
We use jsdom in the tests, and it [doesn't implement `URL.createObjectURL` or `URL.revokeObjectURL`](https://github.com/tmpvar/jsdom/issues/1721). Unless there's some creative way to bypass those with mocks, we might not be able to test this currently.

kturney:
So what is the desired solution?

Delete the tests or (somewhat hackily) stub out URL.createObjectURL and URL.revokeObjectURL if they do not exist?

PR #5241, opened 2017-09-05
3D buildings example: labels on top, hash option

1ec5:
This PR makes two modifications to the “[Display buildings in 3D](https://www.mapbox.com/mapbox-gl-js/example/3d-buildings/)” example based on [this gist](https://gist.github.com/1ec5/09234e60976ad968dc3fd499fb893717):

* The extrusion layer is added beneath any symbol layers that are currently on top – that is, any labels, icons, and shields.
* The `hash` option is set, so that the “[Simple 3D buildings](https://wiki.openstreetmap.org/wiki/Simple_3D_buildings#Demo_areas)” page on the OpenStreetMap Wiki can link to this example page instead of my gist. The wiki page needs to link to specific places on the map, and I think this example page would be more effective than pointing people to [this block](https://bl.ocks.org/1ec5/raw/09234e60976ad968dc3fd499fb893717/) that asks for an access token on the first visit.

/cc @lbud @mourner

### DIFF
--- a/docs/_posts/examples/3400-01-30-3d-buildings.html
+++ b/docs/_posts/examples/3400-01-30-3d-buildings.html
@@ -14,6 +14,7 @@ var map = new mapboxgl.Map({
     zoom: 15,
     pitch: 45,
     bearing: -17.6,
+    hash: true,
     container: 'map'
 });
 
--- a/docs/_posts/examples/3400-01-30-3d-buildings.html
+++ b/docs/_posts/examples/3400-01-30-3d-buildings.html
@@ -17,8 +17,12 @@ var map = new mapboxgl.Map({
     container: 'map'
 });
 
-// the 'building' layer in the mapbox-streets vector source contains building-height
-// data from OpenStreetMap.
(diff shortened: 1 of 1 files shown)


docs/_posts/examples/3400-01-30-3d-buildings.html
```diff
@@ -14,11 +14,17 @@
...
     bearing: -17.6,
+    hash: true,
     container: 'map'
 });
 
-// the 'building' layer in the mapbox-streets vector source contains building-height
-// data from OpenStreetMap.
+// The 'building' layer in the mapbox-streets vector source contains building-height
+// data from OpenStreetMap. Insert the layer beneath any symbol layer.
+var layers = map.getStyle().layers.reverse();
```
1ec5: This call needs to be inside the `load` handler. Fixed in 6d6c5d92d8e89e8d6c4e608cddaf1febe43c47b1.

1ec5:
The OSM Wiki [now links](https://wiki.openstreetmap.org/wiki/Special:Diff/1504804) to the example page.

PR #5242, opened 2017-09-05
document Marker#getElement and add Popup#getElement

mollymerp:
per a question on stackoverflow, `Marker#getElement` was not documented, and I also noticed `Popup` did not have an equivalent function.
 
## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -107,6 +107,10 @@ class Marker {
         return this;
     }
 
+    /**
+     * Returns the `Marker`'s HTML element.
+     * @returns {HTMLElement} element
+     */
     getElement() {
         return this._element;
     }


src/ui/popup.js
```diff
@@ -213,6 +213,15 @@ class Popup extends Evented {
     }
 
     /**
+     * Returns the `Popup`'s HTML element.
+     * @returns {HTMLElement} element
+     */
+    getElement() {
+        return this._content ? this._content : null;
+
+    }
```
anandthakker: 🤔  there's a a slight mismatch here between the element returned by `getElement()` -- which is the wrapping `<div class="mapboxgl-popup-content">...</div>` -- versus the element provided by, say, `setDOMContent()` / `setHTML()` / `setText()`, which would set the _content_ of that div.

If someone does `setHTML(<div class="my-popup">...</div>)`, they might reasonably expect to get the `"my-popup"` div back from `getElement()`.  But since the the content provided by the setters need not be a single element, we can't use that as the return value for `getElement()`.  Maybe the mismatch is fine, but we specify it in the docs?

mollymerp: ah you're right @anandthakker – it is weird especially bc this will include the little close button we include by default 💭

mollymerp: removing `Popup` portion of this PR – will revisit if we get user requests.

PR #5243, opened 2017-09-05
Pause animation loop when removing canvas source

### LINKED ISSUE #5097: Removing canvas source with animate=true leaves map in render loop
**mapbox-gl-js version**: 39.1

### Steps to Trigger Behavior

 1. Add canvas source / layer (with default animate=true)
 2. Remove canvas source / layer

### Expected Behavior

Map should revert to state it was in prior to adding canvas source / layer initially.

### Actual Behavior

Map is now in an infinite render loop and eating CPU.

### Notes

This relates to a [comment in another](https://github.com/mapbox/mapbox-gl-js/issues/5052#issuecomment-318209858) issue but I think warrants its own space.

In the following codepen, if you add and then remove a canvas layer, you can see that the map is stuck in a loaded = false state indefinitely. This is indicative of the infinite render loop mentioned in the linked comment above. If you want your CPU cycles back, you have to pause the canvas source before removing it. I think it should be auto-paused when being removed.

https://codepen.io/gpbmike/pen/prEMrX

gpbmike:
### Problem

When removing a canvas source, the map can be left in a infinite animation loop. 

### Solution

When removing a canvas source, pause the animation.

### Additional Changes

Protect against setting more than more animation loop per canvas source.

fixes #5097

### DIFF
--- a/src/source/canvas_source.js
+++ b/src/source/canvas_source.js
@@ -66,12 +66,16 @@ class CanvasSource extends ImageSource {
         let loopID;
 
         this.play = function() {
-            loopID = this.map.style.animationLoop.set(Infinity);
-            this.map._rerender();
+            if (!loopID) {
+                loopID = this.map.style.animationLoop.set(Infinity);
+                this.map._rerender();
+            }
         };
 
         this.pause = function() {
-            this.map.style.animationLoop.cancel(loopID);
+            if (loopID) {
+                loopID = this.map.style.animationLoop.cancel(loopID);
+            }
         };
 
         this._finishLoading();
@@ -93,6 +97,10 @@ class CanvasSource extends ImageSource {
         if (this.canvas) {
             if (this.animate) this.play();
         }
+    }
+
+    onRemove() {
+        this.pause();
     }
 
     /**
--- a/src/source/canvas_source.js
+++ b/src/source/canvas_source.js
@@ -66,14 +66,14 @@ class CanvasSource extends ImageSource {
         let loopID;
 
         this.play = function() {
-            if (!loopID) {
+            if (loopID === undefined) {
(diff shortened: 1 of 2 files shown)


gpbmike:
I would love some help writing tests for this PR. I don't feel like I'm knowledgable enough about the internals to know how to write a good test.

I would like to add a canvas source and then remove it and check for rerenders.

lbud:
Hey @gpbmike, thanks for your contribution!

The canvas source unit tests are [here](https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/source/canvas_source.test.js). All unit tests are run with `<yarn|npm> run test-unit`, but you can run just the one file with `node test/unit/source/canvas_source.test.js`. Adding a `'remove source'` test [within the larger CanvasSource test](https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/source/canvas_source.test.js#L96), following the patterns of the other subtests, would be good here.

Right now the canvas source tests [stub](https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/source/canvas_source.test.js#L33) `source.map.style.animationLoop`. What I'd recommend is changing that to include an actual `AnimationLoop` instance, like we do in [animation_loop.test.js](https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/style/animation_loop.test.js). Then you'll be able to use the [`AnimationLoop#stopped`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/style/animation_loop.js#L13-L18) method to check to make sure removing the canvas source cancels the animation loop.

Let me know if you have any more questions — happy to help further!

gpbmike:
@lbud Thanks for your help again. I have added tests for `onRemove`, `pause`, and `play`.

Issue #5208, opened 2017-08-30
Zoom performance with cluster of points


sthomp:
I thought this might be related to https://github.com/mapbox/mapbox-gl-js/issues/5041 but I've tried on both 0.38.0 and 0.39.1

### Steps to Trigger Behavior

 1. Using geojson points (1 point = 1 feature, as described in https://www.mapbox.com/mapbox-gl-js/example/geojson-markers/)
 2. Have a densely clustered area of points
 3. Zoom in
 4. Wait for points to load

### Expected Behavior

Points should render smoothly

### Actual Behavior

Points look sparse then after a few seconds will load

Here is an example of a dense area of about 13k points. When I first zoom in it looks like theres only a small number of points.

<img width="474" alt="screen shot 2017-08-30 at 10 32 26 am" src="https://user-images.githubusercontent.com/1254471/29878334-41ad3946-8d70-11e7-9f26-bca9b87e59bf.png">

After waiting a few seconds the rest of the points render:

<img width="403" alt="screen shot 2017-08-30 at 10 32 32 am" src="https://user-images.githubusercontent.com/1254471/29878343-46164a04-8d70-11e7-89d2-c5e0b9ece5b9.png">

jfirebaugh:
Thanks for using Mapbox, and sorry to hear you're running into an issue. In order for us to diagnose this issue, we'll need to be able to reproduce it ourselves. Could you please create a minimal self-contained JSFiddle that demonstrates the issue?

sthomp:
Hi, thanks for the quick response. 

I've created a minimal example to reproduce. When the example loads it will zoom into the points. Then you'll need to wait for about 5-10 seconds for the points to re-appear.

https://jsfiddle.net/w2frv0ns/

https://github.com/sthomp/mapbox_test

Here's a screen recording:

![mapbox_zoom](https://user-images.githubusercontent.com/1254471/29890004-1cd60638-8d94-11e7-93ea-f10142ca8833.gif)

jfirebaugh:
[...]

PR #5244, opened 2017-09-06
Avoid intermediate representation of geometry in FeatureWrapper

jfirebaugh:
Less GC, better performance on https://github.com/mapbox/mapbox-gl-js/issues/5208.

benchmark | master 78e8e74 | perf-5208 1dc290b
--- | --- | ---
**map-load** | 86 ms  | 89 ms 
**style-load** | 87 ms  | 89 ms 
**buffer** | 1,008 ms  | 988 ms 
**tile_layout_dds** | 1,376 ms  | 1,386 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 6.6 ms, 1% > 16ms  | 6.7 ms, 2% > 16ms 
**query-point** | 1.10 ms  | 1.05 ms 
**query-box** | 69.43 ms  | 80.24 ms 
**geojson-setdata-small** | 12 ms  | 11 ms 
**geojson-setdata-large** | 159 ms  | 127 ms 
**filter** | n/a  | n/a

### DIFF
--- a/src/source/geojson_wrapper.js
+++ b/src/source/geojson_wrapper.js
@@ -64,29 +64,6 @@ class FeatureWrapper implements VectorTileFeature {
         }
     }
 
-    bbox() {
-        const rings = this.loadGeometry();
-        let x1 = Infinity,
-            x2 = -Infinity,
-            y1 = Infinity,
-            y2 = -Infinity;
-
-        for (let i = 0; i < rings.length; i++) {
-            const ring = rings[i];
-
-            for (let j = 0; j < ring.length; j++) {
-                const coord = ring[j];
-
-                x1 = Math.min(x1, coord.x);
-                x2 = Math.max(x2, coord.x);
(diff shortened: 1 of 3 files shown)


src/source/geojson_wrapper.js
```diff
@@ -48,28 +42,30 @@ class FeatureWrapper implements VectorTileFeature {
...
+            return geometry;
         }
-        return this.geometry;
     }
 
     bbox() {
-        if (!this.geometry) this.loadGeometry();
-
-        const rings = this.geometry;
+        const rings = this.loadGeometry();
```
mourner: This will mean rebuilding the geometry from scratch on every `bbox()` call — previously it was cached. Maybe we should calculate bbox as a part of `loadGeometry` and then just return the calculated value here?

jfirebaugh: `bbox` is never used; I removed it from the interface type to enforce that.

Issue #5234, opened 2017-09-01
Revise approach to handling expression type errors


anandthakker:
Right now, `coalesce` not only coalesces `null` values, but also serves as a try/catch.  The idea behind this latter behavior was that `coalesce` could be used to provide fallback values for expressions where feature data didn't conform to expected types.  However, @mourner points out that these error-handling semantics are awkward and dangerous (in the sense of swallowing rather than failing hard on errors).  Capturing from chat an revised design for handling mis-typed data:

 - `"get"` returns `null` (rather than producing an error) for a missing property value.
 - Type assertions (`"string"`, `"number"`, ...) and coercions (`"to-string"`, `"to-number"`, ...) accept one or more fallback values in case of mis-typed input, e.g. `["number", ["get", "x"], ["get", "y"], 0]`: if `properties.x` is not numeric, then evaluate to `properties.y` -- if _that's_ not numeric, then evaluate to 0.
 - Either:
   - `"coalesce"` no longer catches errors -- it only coalesces `null`.
   - We remove `"coalesce"` altogether, instead allowing a single, literal fallback value for `"get"` and `"id"`, e.g. `["get", "x", 5]`: evaluate to `properties.x` or, if it is null, to 5.

cc @mourner @jfirebaugh @nickidlugash @aparlato

jfirebaugh:
>  - Either:
>   - `"coalesce"` no longer catches errors -- it only coalesces `null`.
>   - We remove `"coalesce"` altogether, instead allowing a single, literal fallback value for `"get"` and `"id"`, e.g. `["get", "x", 5]`: evaluate to `properties.x` or, if it is null, to 5.

It's probably not a good idea to have two independently-optional arguments to `get`. It's contrived, but consider `["get", "x", ["literal", {"x": "foo"}]`. Is the second argument the object to `get` from, or a fallback for a feature property lookup?

anandthakker:
@jfirebaugh good point.  We could do:

 - `get` no longer defaults to `["properties"]` -- i.e., it requires the object argument: `["get", key, object]` / `["get", key, object, defaultValue]`.
 - introduce `["prop", key]` / `["prop", key, defaultValue]` as an alias for `["get", key, ["properties"]` / `["get", key, ["properties", defaultValue]`

Or we could just keep `coalesce`.  The reason the above alternative is attractive is that it would allow us to remove `Null` altogether.

anandthakker:
[...]

PR #5250, opened 2017-09-07
Refactor generated expression code to avoid allocations

anandthakker:
h/t @mourner for fleshing out this approach


anandthakker:
@mourner as expected, the refactor wasn't too bad.  you can quickly see the generated code for various expressions on the results page if you run `node test/expression.test.js`.   Currently performance is roughly the same as `master`, but that's without having simplified the generated code for most expressions yet.

mourner:
This is a great start! Let's try fighting back the performance based on this, starting with things that go into stop functions generated code.

anandthakker:
screen shot 2017-09-07 at 9 35 34 am

Getting pretty close to pre-expressions.  @mourner I think this is ready for a review -- I took a stab at removing try/catch, but it's a pretty intensive change, so I think it's out of scope for this PR.

src/style-spec/function/definitions/curve.js
```diff
@@ -17,6 +16,11 @@ export type InterpolationType =
     { name: 'exponential', base: number } |
     { name: 'cubic-bezier', controlPoints: [number, number, number, number] };
 
+export type InterpolationCache = {
+    pow: {[number]: {[number]: number}},
+    unitBezier: {[string]: UnitBezier}
+}
```
mourner: I'm wondering if this kind of cache gives us any perf wins given the increased complexity... If it's marginal, maybe we should take this out and revisit later? E.g. UnitBezier [doesn't do much in the constructor](https://github.com/mapbox/unitbezier/blob/master/index.js#L33-L39) compared to the main approximation code. And `Math.pow` values are cached by non-integer `exp` values which are known to be much slower in V8 than integer keys.

src/style-spec/function/definitions/match.js
```diff
@@ -95,26 +95,30 @@ class Match implements Expression {
...
+        }
+
+        const lookupObject = ctx.addVariable(`{${lookup}}`);
+
+        const inputType = this.inputType.kind !== 'Array' ?
+            `$this.types.${this.inputType.kind}` : JSON.stringify(this.inputType);
+
+        return `
+var i = $this.as(${input}, ${inputType});
+return (${lookupObject}[i] || ${ctx.addExpression(this.otherwise.compile(ctx))})();`;
```
mourner: Maybe let's indent lines like this — we want to prioritize source code readability over generated one's.

src/style-spec/function/expression.js
```diff
@@ -123,8 +125,87 @@ class ParsingContext {
...
+        let id = this._cache[body];
+        if (!id) {
+            id = `e${this._id++}`;
+            this._cache[body] = id;
+
+            // convenience: allow simple expressions to omit 'return ' in
+            // compiled string
+            if (!/return/.test(body)) {
+                body = `return ${body}`;
+            }
```
mourner: I think it might be better to write out `return` explicitly than introduce non-obvious shortcuts like this...

mourner:
To expand on my caching comment:

- `UnitBezier` caching could probably be more effective if it was cached as a variable in codegen rather than looked up at runtime with `controlPoints.join(',')` key.
- `Math.pow(difference)` values could potentially be cached more effectively by saving them as an array variable in codegen for each curve, with indices corresponding to stops in the curve for easy lookup (instead of the complex `{<base>: {<difference>: value}}` object).

src/style-spec/function/definitions/curve.js
anandthakker: @mourner ah, yeah: just realzied that the UnitBezier caching is a holdover from when I was using https://github.com/gre/bezier-easing, which does some precomputation upon construction.

Re: `Math.pow`, I did, in fact see a small but definitely visible gain from adding this caching -- something 2% on the tile_layout_dds benchmark.

src/style-spec/function/expression.js
anandthakker: Agreed.  I never did feel good about this 😄

anandthakker: @mourner I'm going back on this.  An advantage to having something like this is that, when it's possible for an expression to be compiled to a pure JS expression that can be easily inlined (rather than a function body, which would be more complicated to inline), we could do so.  I know you said that JS engines will inline pure functions w/o side effects, but I'd still like to leave this open for further investigation.

anandthakker: (That said, the current regex check is definitely not the way to do this.)

mourner: How about converting all keywords to be pure expressions and then always adding a `return` here? This way all expression would potentially be inlinable and no need for this conditional logic. I went through all expressions in this ticket and it looks like they all could be converted.

anandthakker: I'm not sure they can all be pure expressions, _unless_ we commit to providing a helper method in `EvaluationContext` for any case where we need to use a subexpression's value more than once.

anandthakker:
@mourner okay, removed the interpolation caching, and fixed up the regex check for `return` -- you were right, every expression except for "coalesce" could be compiled to a pure JS expression.  I added a bit of special case handling for `coalesce`, but we can remove it as soon as we implement https://github.com/mapbox/mapbox-gl-js/issues/5234 (which I'll work on as soon as this is merged)

src/style-spec/function/definitions/curve.js
```diff
@@ -31,34 +26,23 @@ class Curve implements Expression {
...
-            const key = interpolation.controlPoints.join(',');
-            let ub = cache.unitBezier[key];
-            if (!ub) {
-                ub = new UnitBezier(...interpolation.controlPoints);
-                if (cache.unitBezier) {
-                    cache.unitBezier[key] = ub;
-                }
-            }
-            t = ub.solve(exponentialInterpolation(input, 1, lower, upper, cache.pow));
+            const ub = new UnitBezier(...interpolation.controlPoints);
```
mourner: Let's change this to explicit `p[0], p[1], p[2], p[3]`. I just realized that this can be a huge hit because Buble [compiles this](http://buble.surge.sh/#const%20ub%20%3D%20new%20UnitBezier(...interpolation.controlPoints)%3B) to the following atrocity:

```js
var ub = new (Function.prototype.bind.apply( UnitBezier, [ null ].concat( interpolation.controlPoints) ));
```

anandthakker: oof good call

PR #5253, opened 2017-09-07
Heatmaps layer

### LINKED ISSUE #4756: Proposal: Heatmap layer type
## Motivation

Heatmap support is one of the most requested features for Mapbox GL, and an extremely common map visualization type, especially prevalent in weather, travel, fitness and social apps.

Heatmaps are typically used to visualize density of a point dataset. It's in a way similar to clustering, but with a very different look that shows areas points are crowded in very well.

image
image

### Option A: Implement it as an external plugin

We could use the heatmaps use case to drive the design of custom layers API and custom sources API, both of which are not fleshed out yet, and then package it as an external plugin. The drawbacks of this approach are:

1. Heatmaps are a perfect candidate for a Mapbox Studio feature, since the process of designing them (picking the color ramp, adjusting radius, weights etc.) highly benefits from instant visual feedback. However, we can only add Studio features that seamlessly work across platforms without additional configuration and are a part of the style specification, so Studio won't be able to support heatmaps as a plugin.
2. The native plugin for heatmaps will have to be implemented independently, and we have no fleshed out plugin architecture on the native side either.
3. We want to reuse a lot of the existing Mapbox GL features such as functions, data-driven properties, clustering (very important for heatmapping big datasets), and (in near future) arbitrary expressions and property aggregations. Because of that, having to implement heatmaps as a plugin (and designing flexible enough custom layers/sources API) might be technically very difficult.
4. Heatmaps won't be available through the static maps API and the raster tile fallback.
5. It will likely be much harder to use compared to heatmaps as a core layer type.

### Option B: `colorize-alpha` property 

Prior discussion in #3182. The idea is to introduce a `colorize-alpha` property for existing layers like `circle`, then render the circles with blur as is and recolorize according to the specified color ramp that matches each pixel alpha value to a color. Drawbacks:

1. It's not a "true" heatmap as is. Mathematically, heatmap is a [bivariate kernel density estimation](https://en.wikipedia.org/wiki/Multivariate_kernel_density_estimation)). For a good result, you have to blend circles differently (with additive blending — `ONE, ONE` blend mode), and KDEs usually use a different sprite for each point — a [Gaussian kernel](https://en.wikipedia.org/wiki/Gaussian_function), which may also look like a blurred circle but is still different.
2. This would mean introducing a lot of conditional code in circle drawing because of the need to render into a temporary framebuffer, then recolorizing and rendering a texture to the screen (instead of drawing circles directly).
3. It breaks down when you pitch the map, since circles in our rendering don't get the perspective distortion (they remain circles on screen, whether they scale with pitch or not).
4. Some circle properties don't make sense with `colorize-alpha` on — `circle-stroke-*`, `circle-
pitch-scale`, `circle-translate-*`, and `circle-blur` should be always `1.0`. Having these options makes heatmaps harder to use and easier to screw up.

This also feels very similar to the case of extrusion-related properties as a part of the `fill` type vs a separate `fill-extrusion` type. We chose the latter option, and I'm inclined towards the same in the choise of additional `circle` properties vs new `heatmap` layer type.

## Design

Given considerations above, I propose to implement a new `heatmap` layer type in the core. It will have its own set of shaders and drawing code along other types. We'll be able to render either vector tile points, GeoJSON points or GeoJSON clusters as heatmaps, just like we do with circles.

The main drawback of this approach are code and specification bloat, since we are unlikely to remove a feature once we add it. I think the feature is important enough to justify being in the core, but not yet fully convinced — let's discuss.

### Mock-Up

```json
"type": "heatmap",
"paint": {
    "heatmap-radius": 30,
    "heatmap-color": {
        "colorSpace": "lab",
        "stops": [[0, "yellow"], [0.4, "green"], [1.0, "red"]]
    },
    "heatmap-weight": {
        "type": "identity",
        "property": "score"
    },
    "heatmap-max-weight": 100,
    "heatmap-opacity": 0.7,
    "heatmap-mode": "density"
}
```

- **heatmap-radius**: controls the radius of the kernel. The bigger the radius, the smoother the heatmap (but with less detail). Radius can be a zoom function.
- **heatmap-color**: the color ramp used for coloring pixels. Using stop functions for this is very convenient — you can make a smooth gradient, or a contour-like rendering using an interval function. The only weird thing about it is that the input stop values are pixel `alpha` values, not `zoom`, so this would be an outlier in the spec. Is this OK? And how would an equivalent function look in the arbitrary expressions proposal?
- **heatmap-weight**: for weighted heatmaps where each point has an importance score. 1.0 by default (all points are treated equally). Weight === 3.0 should be visually the same as having three 1.0-weight points in the same spot.
- **heatmap-max-weight**: a global layer property for scaling the heatmap intensity.
- **heatmap-opacity**: global heatmap layer opacity (applied to the final result).
- **heatmap-mode**: by default, heatmaps render according to point density (so two points in the same spot are twice as intense as one). But there are cases where we want a different behavior — imagine an air pollution map that shows measurements from spots around the country. Multiple measure points clustered close together shouldn't cause the area to appear many times more polluted compared to an area with the same pollution but only one measure spot. We could implement this behavior by introducing a property that governs the underlying blending GL equation (for the pollution case, we would use `EXT_blend_minmax.MAX` rather than `GL_FUNC_ADD`).

## Implementation

The only way I've seen heatmaps implemented in the GPU is the same — draw a kernel ("blurred circle") for each point of the dataset with additive blending (storing the result in a temporary framebuffer), then colorize each pixel using a color ramp (based on alpha values). We'll have to use the same approach — I haven't found a more efficient one for the GPU.

As for performance, we have the advantage of being able to cluster points (which is already implemented), which enables us to render huge datasets as heatmaps on the client.

cc @kkaefer @anandthakker @ansis @1ec5 @ryanbaumann @peterqliu @mapbox/gl

Previous tickets: #3994 #3182

mourner:
Closes #4756. Still a work in progress, but pretty close.

### Technical tasks

 - [x] rebase on top of master with expressions
 - [x] refactor to take advantage of the new `RenderTexture` class (edit: not applicable for now)
 - [x] make sure `viewportTexture` caching works properly and is aligned with master
 - [x] possibly increase the number of values in the color ramp lookup array (update: no need, very small difference)
 - [x] add comments for tricky heatmap fragment shader math
 - [x] fix halos around points
 - [x] commit a small GeoJSON fixture for the `heatmaps.html` debug page
 - [x] handle the need to have the first color stop with zero alpha value
 - [x] settle sensibly on all default property values
 - [x] fix issues when resizing the map
 - [x] cache color ramp GL texture object
 - [x] make sure all texture preparations happen in the upload step rather than in the middle (edit: nothing to do in the upload step)
 - [x] ~~take advantage of the new tile mask feature to~~ get rid of occasional flickering while zooming (used a simpler hack)
 - [x] don't create viewport buffer/vao on every frame
 - [x] ~~investigate prerendering a reusable kernel texture for performance~~ no need for now, heatmap rendering is already fast enough
 - [x] make sure environments without half-float/half-float-linear extensions fall back gracefully
 - [x] fix flow errors
 - [x] put labels on top in the debug page
 - [x] adjust kernel sizes depending on weight to minimize artifacts
 - [x] ~~`Uint8Array` -> `Uint8ClampedArray`~~ WebGL doesn't like it
 - [x] use `getPaintValue` instead of `layer.paint` where possible
 - [x] use `pixelsToTileUnits`
 - [x] extract `sourceCache._tiles[parentId] && !sourceCache._coveredTiles[parentId]` into a method
 - [x] explain `blendFunc(ONE, ONE)` in the comments in detail
 - [x] explain the half-float texture with the fallback use in the comments
 - [x] explain the tile flicker hack better in the comments
 - [x] fall back to non-float texture on incomplete framebuffer to make sure heatmaps work on Android
 - [x] add version numbers to spec sdk-support property
 - [x] make sure the docs are great
 - [x] rebase against master again
 - [x] add heatmap render tests

### Things I need help with

 - [x] review #5328 to unblock color ramp texture caching
 - [x] find a less hacky way to set up the non-zoom curve property (for the color ramp) @anandthakker #5350 
 - [x] refactor `StyleLayer` to handle functions as default properties @anandthakker 
 - [x] ~~render heatmap intermediate texture in a separate pass (`3d`?)~~ decided with @lbud @mollymerp to follow up on this after both this and raster DEM PRs are merged
 - [ ] add an official example @ryanbaumann


anandthakker:
@mourner FYI https://github.com/mapbox/mapbox-gl-js/issues/5308

mourner:
Performance is looking pretty good at this point — here's a sample JS profiler breakdown (a heavy clustered heatmaps layer takes only 2x as much as the background): 

image

debug/heatmap.html
```diff
@@ -0,0 +1,66 @@
...
+                    [0.3, "cyan"],
+                    [0.5, "lime"],
+                    [0.7, "yellow"],
+                    [1, "red"]
+                ],
+                "default": "rgba(0, 0, 0, 0)"
+            }
+        }
+    });
+});
```
kkaefer: Can we add the layer below labels please?

src/render/draw_heatmap.js
```diff
@@ -0,0 +1,137 @@
...
+
+    gl.clearColor(0, 0, 0, 0);
+    gl.clear(gl.COLOR_BUFFER_BIT);
+
+    gl.blendFunc(gl.ONE, gl.ONE);
+
+    for (let i = 0; i < coords.length; i++) {
+        const coord = coords[i];
+
+        // skip tiles that have uncovered parents to avoid flickering
```
kkaefer: Can you please expand on the reasoning of this a little more? And potentially add a render test?

src/render/draw_heatmap.js
```diff
@@ -0,0 +1,137 @@
...
+    const gl = painter.gl;
+
+    painter.setDepthSublayer(0);
+    painter.depthMask(false);
+
+    // Allow circles to be drawn across boundaries, so that
+    // large circles are not clipped to tiles
+    gl.disable(gl.STENCIL_TEST);
+
+    renderToTexture(gl, painter, layer);
```
kkaefer: Can we please move this call to a different render pass to improve performance? Extrusion fills use a `3d` render pass to render the texture before rendering to the main FBO. See https://github.com/mapbox/mapbox-gl-native/issues/9286 for a description of this optimization, and https://github.com/mapbox/mapbox-gl-js/pull/5101 for @lbud implemented this for extrusion fills.

src/style-spec/reference/v8.json
```diff
@@ -2961,6 +2969,99 @@
...
+    "heatmap-radius": {
+      "type": "number",
+      "default": 30,
+      "minimum": 1,
+      "function": "interpolated",
+      "zoom-function": true,
+      "property-function": false,
+      "transition": true,
+      "units": "pixels",
+      "doc": "Heatmap radius.",
```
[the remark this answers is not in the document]
mourner: Yes, that's still in the checklist as you can see.

src/style-spec/reference/v8.json
```diff
@@ -376,6 +376,13 @@
             }
           }
         },
+        "heatmap": {
+          "doc": "A heatmap.",
+          "sdk-support": {
+            "basic functionality": {
+            }
```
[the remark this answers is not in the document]
mourner: Should I include the number of the next would-be release, e.g. `0.41` here for JS?

src/style/style_layer/heatmap_style_layer.js
```diff
@@ -0,0 +1,58 @@
...
+        super._applyPaintDeclaration(name, declaration, options, globalOptions, animationLoop, zoomHistory);
+        if (name === 'heatmap-color') {
+            const len = this.colorRampData.length;
+            for (let i = 4; i < len; i += 4) {
+                const pxColor = this.getPaintValue('heatmap-color', {zoom: i / len});
+                const alpha = pxColor[3];
+                this.colorRampData[i + 0] = Math.floor(pxColor[0] * 255 / alpha);
+                this.colorRampData[i + 1] = Math.floor(pxColor[1] * 255 / alpha);
+                this.colorRampData[i + 2] = Math.floor(pxColor[2] * 255 / alpha);
+                this.colorRampData[i + 3] = Math.floor(alpha * 255);
```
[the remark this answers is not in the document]
mourner: Thanks for the explanation — I'll look more closely at the premultiplying stuff.

> It'd also be cool to have unit tests for generating color ramp textures/buffers from a style specification.

I think it would be enough to cover this with a render test, since the results are visual, not something where a difference would be useful to see as numbers in the console. What do you think?

debug/heatmap.html
mourner: Doesn't matter since it's just a debug page. The official example definitely should have labels on top though.

src/render/draw_heatmap.js
mourner: @kkaefer @lbud should we call this pass differently then? e.g. `prerender` or something?

src/render/draw_heatmap.js
mourner: Certainly needs a render test, although it will be tricky to make because of the need to manually make 4 new tile fixtures. The situation here is similar to #5004. The difference is that heatmaps don't need such a complex fix (tile masks) — the difference between tiles of different zooms for heatmaps is much more subtle since it's still vector data, so waiting for all 4 children to load before putting them in place of a parent is an acceptable solution to the problem that's also faster and much simpler.

src/style/style_layer/heatmap_style_layer.js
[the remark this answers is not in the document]
mourner: Per chat, the code works correctly — leaving it as is as a weird special case because we need to un-premultiply the premultiplied colors returned by `getPaintValue` because the `Texture` class later premultiplies them when uploading again.

debug/heatmap.html
kkaefer: Yeah, but people still look at this, and we should set a good example here.

src/render/draw_heatmap.js
lbud: Yes, @mollymerp and I were just talking about this yesterday, as hillshades use a pre-render pass as well — it would make sense to rename the `3d` pass to `prerender` and carefully move relevant parts of the current `3d` pass (I suppose for now we could move them into `draw_fill_extrusion.js`, although eventually if we add other 3D types it'll be moved to a file where it can be shared between types).

src/shaders/heatmap.fragment.glsl
```diff
@@ -0,0 +1,21 @@
...
+
+// Gaussian kernel coefficient: 1 / sqrt(2 * PI)
+#define GAUSS_COEF 0.3989422804014327
+
+void main() {
+    #pragma mapbox: initialize highp float weight
+
+    // Kernel density estimation with a Gaussian kernel of size 5x5
+    float d = -0.5 * 5.0 * 5.0 * dot(v_extrude, v_extrude);
+    float val = weight * u_intensity * GAUSS_COEF * exp(d);
```
anandthakker: I don't think the 5^2 factor here technically translates to a kernel size of 5x5.  Here's my read/analysis to make sure I'm understanding this right:

The "bandwidth" of the Gaussian kernel (which is also its standard deviation) would be the value `h` in `1/2pi * exp( -0.5 * (u/h)^2 )`

Since `v_extrude` is essentially in units of radii (i.e. at a distance of `u_radius` from the point, `v_extrude` is a unit vector), the calculation above is equivalent to `weight * u_intensity * GAUSS_COEF * exp( (distance_in_pixels / (radius / 5)^ )`, making the bandwidth `radius/5`.

Three standard deviations is a pretty reasonable definition for the practical "size" of a Gaussian kernel.  In this case, that would be: `3/5 * radius`.

src/style/style_layer/heatmap_style_layer.js
```diff
@@ -0,0 +1,58 @@
...
+        this.colorRampData = new Uint8Array(256 * 4);
+        if (!this.getPaintProperty('heatmap-color')) {
+            this.setPaintProperty('heatmap-color', this._paintSpecifications['heatmap-color'].default, '');
+        }
+    }
+
+    _applyPaintDeclaration(name: any, declaration: any, options: any, globalOptions: any, animationLoop: any, zoomHistory: any) {
+        super._applyPaintDeclaration(name, declaration, options, globalOptions, animationLoop, zoomHistory);
+        if (name === 'heatmap-color') {
+            const len = this.colorRampData.length;
```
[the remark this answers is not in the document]
mourner: I actually started with `setPaintProperty` but later moved to this, I don't remember why — will recheck.

src/render/draw_heatmap.js
```diff
@@ -0,0 +1,137 @@
...
+    gl.blendFunc(gl.ONE, gl.ONE);
+
+    for (let i = 0; i < coords.length; i++) {
+        const coord = coords[i];
+
+        // skip tiles that have uncovered parents to avoid flickering
+        const parentTile = sourceCache.findLoadedParent(coord, 0, {});
+        if (parentTile) {
+            const parentId = parentTile.coord.id;
+            if (sourceCache._tiles[parentId] && !sourceCache._coveredTiles[parentId]) continue;
```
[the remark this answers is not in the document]
mourner: Yes, good idea.

src/render/draw_heatmap.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        const tile = sourceCache.getTile(coord);
+        const bucket: ?HeatmapBucket = (tile.getBucket(layer): any);
+        if (!bucket) continue;
+
+        const programConfiguration = bucket.programConfigurations.get(layer.id);
+        const program = painter.useProgram('heatmap', programConfiguration);
+        programConfiguration.setUniforms(gl, program, layer, {zoom: painter.transform.zoom});
+        gl.uniform1f(program.uniforms.u_radius, layer.paint['heatmap-radius']);
```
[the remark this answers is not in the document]
mourner: You mean `getPaintValue`, right? Is there an advantage in case of non-DDS properties? I thought that would be faster.

src/shaders/heatmap.fragment.glsl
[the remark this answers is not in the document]
mourner: Yes, exactly. I thought about this a lot and considered making this one property, but it's just too inconvenient to use — you have to construct complex expressions or zoom+property functions and it's just hard to do intuitively. This separation though fits pretty well into how I think about heatmaps. Another more practical benefit is that you can change heatmap intensity in real time without recomputing all the buffers.

mourner: Yeah, I noted that "radius" is quite a blurry term when referring to kernel density Gaussian kernels, but haven't thought this through further. Given the thinking above, how do you think we should change the formula in the shader?

src/render/draw_heatmap.js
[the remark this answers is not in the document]
anandthakker: Yep sorry `getPaintValue`.  No advantage for non-DDS, but at some point I think we want to get rid of `StyleLayer.{paint,layout}` (or hide it), see https://github.com/mapbox/mapbox-gl-js/issues/3044.  I don't think the performance difference would be enough to matter in code that's executed only once per draw.

src/shaders/heatmap.fragment.glsl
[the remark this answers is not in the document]
anandthakker: Yeah, I guess the only alternative would be that expressions would have to be of the form: `feature-only-expression` or `zoom-only-expression` or `['*', feature-only-expression, zoom-only-expression]`... which is probably just as confusing (or more) as having two properties, heh

anandthakker: I think changing `5.0` to `3.0` would be pretty reasonable.  Then `radius` would have a pretty clean meaning:
 - For advanced users familiar w/ the underlying math, "3 standard deviations of the Gaussian kernel"
 - For most users, "the distance around each point where it influences the heatmap" (or something like that)

anandthakker: Hmm, experimenting with this a bit, one issue is that with if we use `3.0`, then for values of `intensity * weight > 0.5`, the output for a point ends up getting clipped by the square mesh for that point.  Whereas with `5.0`, it takes `intensity * weight > 5000` to make that happen.  

I guess the problem is that, even at x = 3 standard deviations, the gaussian curve evaluates to `0.0018`, which is just half of the 1/255 resolution of the color ramp.

mourner: Yeah, now I remember that empirically 5.0 is reasonable enough to make clipping artifacts not to be noticeable. Picking more, the performance will suffer because of the number of fragments. Any ideas on how to do better here?

anandthakker: Do you think it'd be reasonable to restrict `heatmap-weight` to values between 0 and 1?  If so, then we could choose kernel bandwidth based on the intensity in such a way as to ensure that the kernel output drops to below `1/255/2` at ~ 1 radius:

```
float ZERO = 1.0/255.0/2.0;
float h = sqrt(-2.0 * log(ZERO / u_intensity / GAUSS_COEF));
float d = -0.5 * h * h * dot(v_extrude, v_extrude);
float val = weight * u_intensity * GAUSS_COEF * exp(d);
```

If `weight` can be arbitrarily large, though, then there's sort of no way around the problem (regardless of what value we choose, 3, 5, etc.)

anandthakker: ^ btw that `h` calculation need not happen in the shader -- could also be a uniform itself.  (Dunno if WebGL/OpenGL would optimize such a computation since it only relies on uniform values.)

src/render/draw_heatmap.js
```diff
@@ -0,0 +1,137 @@
...
+    // Allow circles to be drawn across boundaries, so that
+    // large circles are not clipped to tiles
+    gl.disable(gl.STENCIL_TEST);
+
+    renderToTexture(gl, painter, layer);
+
+    gl.clearColor(0, 0, 0, 0);
+    gl.clear(gl.COLOR_BUFFER_BIT);
+
+    gl.blendFunc(gl.ONE, gl.ONE);
```
anandthakker: Would be good to add a comment here to explain the significance of this blend function, since it's essential to how the kernel density method works (and maybe some comments in general on this method)

src/render/draw_heatmap.js
```diff
@@ -0,0 +1,137 @@
...
+
+    if (!texture) {
+        texture = layer.heatmapTexture = gl.createTexture();
+        gl.bindTexture(gl.TEXTURE_2D, texture);
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB, painter.width / 4, painter.height / 4, 0, gl.RGB,
+            painter.extTextureHalfFloat ? painter.extTextureHalfFloat.HALF_FLOAT_OES : gl.UNSIGNED_BYTE, null);
```
anandthakker: What's the purpose of using `HALF_FLOAT_OES` when it's supported?  Is it to increase the precision we use for kde values?

src/shaders/heatmap.fragment.glsl
```diff
@@ -0,0 +1,21 @@
+#pragma mapbox: define highp float weight
+
+uniform float u_intensity;
+varying vec2 v_extrude;
+
+// Gaussian kernel coefficient: 1 / sqrt(2 * PI)
+#define GAUSS_COEF 0.3989422804014327
+
+void main() {
```
anandthakker: Suggest renaming this program from `heatmap` to `heatmap_density`, and then renaming `heatmap_texture` to just `heatmap`

src/shaders/heatmap.fragment.glsl
mourner: We can't restrict the weight, otherwise you won't be able to use clustering with heatmaps. If there's a cluster with 1000 points, it's weight should be 1000 times the single point weight.

mourner: Other option would be to pick another type of kernel (not Gaussian) for the heatmap, or modify the Gaussian to fade out to 0 at the borders regardless of the weight.

src/render/draw_heatmap.js
mourner: Yes. Try disabling it and see how the heatmap looks. Better precision allows for a much smoother look.

src/shaders/heatmap.fragment.glsl
mourner: Not sure those names would be clearer. Also, I named them similarly to the extrusion names — `fill_extrusion` and `extrusion_texture`.

src/shaders/heatmap.fragment.glsl
mourner: Trie Epanechnikov and the heatmap looks pretty bad. Maybe we could both calculate bandwidth and increase the size of the kernel depending on weight in the vertex shader? That way small intensities wouldn't require as many fragments, which would increase performance, and big intensities which cause big kernels wouldn't appear too often.

anandthakker: @mourner Actually if we use a `weight`-based adjustment to the size of the kernel (i.e., the size of the mesh produced by the vertex shader), then we could go back to just using a simple, fixed value of `3.0` for the bandwidth.  I.e., in the vertex shader,

```glsl
    v_extrude = vec2(mod(a_pos, 2.0) * 2.0 - 1.0) * scale_based_on_weight;
```

And then in fragment

```glsl
       float d = -0.5 * h * h * dot(v_extrude / scale_based_on_weight, v_extrude / scale_based_on_weight);
```

Where `scale_based_on_weight` is a scale factor that grows the kernel large enough beyond `radius` so that the kernel falls effectively to zero at the edge

mourner: @anandthakker exactly, that's what I was thinking too! Now we need to determine that scale properly...

src/style/style_layer/heatmap_style_layer.js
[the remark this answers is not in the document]
mourner: Oh, I think I remember now — the problem is you can't call `getPaintValue` immediately after `setPaintProperty`, because it's being set up in a separate "apply paint transitions" stage.

src/style/style_layer/heatmap_style_layer.js
```diff
@@ -0,0 +1,58 @@
...
+    colorRamp: RGBAImage;
+    colorRampTexture: ?Texture;
+
+    createBucket(options: any) {
+        return new HeatmapBucket(options);
+    }
+
+    constructor(layer: LayerSpecification) {
+        super(layer);
+        this.colorRampData = new Uint8Array(256 * 4);
```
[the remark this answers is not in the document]
mourner: Interestingly, WebGL doesn't like `Uint8ClampedArray` as texture input — it produces `texture.js:47 WebGL: INVALID_OPERATION: texImage2D: type UNSIGNED_BYTE but ArrayBufferView not Uint8Array`.

src/style/style_layer/heatmap_style_layer.js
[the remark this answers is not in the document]
anandthakker: Ah, okay.  Could we add a comment about that here, so that no one is tempted to try such a refactor later on?

src/style-spec/reference/v8.json
```diff
@@ -1972,6 +2001,10 @@
         "doc": "Gets the current zoom level.  Note that in style layout and paint properties, [\"zoom\"] may only appear as the input to a top-level [\"curve\"] expression.",
         "group": "Zoom"
       },
+      "heatmap-density": {
+        "doc": "Gets the density of a pixel in a heatmap layer. Can only be used in `heatmap-color` property.",
```
1ec5: Nit: add “the” before `heatmap-color`.

1ec5: As a lay reader, I’d have little idea what it means for a pixel to have density – other than [pixel density](https://en.wikipedia.org/wiki/Pixel_density), which is a different concept. Any idea how to explain density in the context of a heatmap?

src/style-spec/reference/v8.json
```diff
@@ -3022,6 +3056,100 @@
...
+          "stops": [
+              [0, "rgba(0, 0, 255, 0)"],
+              [0.1, "royalblue"],
+              [0.3, "cyan"],
+              [0.5, "lime"],
+              [0.7, "yellow"],
+              [1, "red"]
+          ]
+      },
+      "doc": "Defines the color of each pixel based on its density value in a heatmap. Should be either a stop function with input values ranging from `0` to `1`, or a curve expression with a special `[\"heatmap-density\"]` keyword as the input.",
```
1ec5: Noting that this explanation is a bit specific to the GL JS function and expression syntaxes, so the iOS/macOS SDK will need to override it with the appropriate syntax.

src/render/draw_circle.js
```diff
@@ -45,8 +44,6 @@ function drawCircles(painter: Painter, sourceCache: SourceCache, layer: CircleSt
             gl.uniform2fv(program.uniforms.u_extrude_scale, painter.transform.pixelsToGLUnits);
         }
 
-        gl.uniform1f(program.uniforms.u_devicepixelratio, browser.devicePixelRatio);
```
kkaefer: Why is this no longer necessary?

src/render/draw_heatmap.js
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    renderTextureToMap(gl, painter, layer);
+}
+
+function renderToTexture(gl, painter, layer) {
+    gl.activeTexture(gl.TEXTURE1);
+
+    // Use a 4x downscaled screen texture for better performance
+    gl.viewport(0, 0, painter.width / 4, painter.height / 4);
```
kkaefer: Do we need to account for the device pixel ratio here? E.g. do we need a "retina" version of the downscaled texture?

src/render/draw_heatmap.js
```diff
@@ -0,0 +1,149 @@
...
+
+function bindTextureFramebuffer(gl, painter, texture, fbo) {
+    // Use the higher precision half-float texture where available (producing much smoother looking heatmaps);
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, painter.width / 4, painter.height / 4, 0, gl.RGBA,
+        painter.extTextureHalfFloat ? painter.extTextureHalfFloat.HALF_FLOAT_OES : gl.UNSIGNED_BYTE, null);
+
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 0);
+
+    // If using half-float texture as a render target is not supported, fall back to a low precision texture
```
kkaefer: Is there any way we can test unavailability of this extension? For VAOs, I think we have code that specifically prevents reporting that this extension is available so that we can run the test with it enabled and disabled.

src/render/draw_circle.js
mourner: It's an unused leftover — we now define DEVICE_PIXEL_RATIO as a shader constant in the shader generation code.

src/render/draw_heatmap.js
mourner: I think not because `painter.width` and `painter.height` already have "retina" values.

src/render/draw_heatmap.js
mourner: Unfortunately the `gl` module which we use for tests doesn't seem to support the extension, so we're forced to only test the "fallback" state.

Also, this check is not about availability of the halfFloat extension, it's about being able to use halfFloat textures as a render target (which some GPUs simply don't support while they support the extension). So the only way to know that is to try, and then check the framebuffer attachment status (this is even officially documented somewhere).

mourner: Found this thread on the matter https://groups.google.com/forum/#!topic/webgl-dev-list/ms21hnWk0tg

Issue #5261, opened 2017-09-08
Add expression operator to determine style image availability


1ec5:
The style specification should have an `image-exists` or `has-image` operator for use in expressions. It would accept one argument, the name of an image, and return a Boolean indicating whether the image is part of the style (e.g., via `addImage()`). Typically, this operator would be used inside a `case` operator.

This operator would allow style designers to significantly condense any symbol layers that display route shields using the Mapbox Streets source by no longer hard-coding a list of known `shield` values: https://github.com/mapbox/mapbox-gl-js/issues/4149#issuecomment-276818537. It would also future-proof those layers for any additional `shield` values that we would add to the source down the line; currently such changes are breaking changes because the hard-coded list has no fallback mechanism.

An `image-exists` operator would be more flexible than the “image stacks” proposed in #4149: it could be used on any property, not only an image-typed property. The existence of an image with a given name could determine not only the icon to display but also the color of the text to superimpose. A designer would still need to employ the `-white` hack described at the bottom of https://github.com/mapbox/mapbox-gl-js/issues/4149#issuecomment-276818537 until we get a substring expression operator (#4113), but at least the long and fragile list would go away and we’d be able to use a single layer for both black- and white-lettered shields.

For context, a shield fallback was a key use case that led to the implementation of expressions in the first place: https://github.com/mapbox/mapbox-gl-style-spec/issues/104#issuecomment-263762518. In https://github.com/mapbox/mapbox-gl-js/pull/4715#issuecomment-302826118, we decided to punt on this use case for the initial PR that implemented expressions.

/cc @anandthakker @jfirebaugh @mapbox/cartography

jfirebaugh:
With an eye towards future introduction of expression forms that combine, lighten, darken, recolor, or otherwise manipulate images, I would design this slightly differently, providing an `["image", id]` expression form whose result type is `Image | Null`, where `Image` is a new type. Then `coalesce` is convenient for finding the first available image in a stack, and `case` can still be used as well.

1ec5:
Would properties like `icon-image` continue to take a string, or would they migrate over to the new Image type?

1ec5:
Moving `icon-image` over to a new Image type seems like a pretty long-range proposal. In the meantime, it wouldn’t be possible to use `coalesce` with the `image` operator for `icon-image`. Instead, you’d have to write `["match", ["image", "us-oh-sci-3"], null, "us-oh-sci-3"]`.

[...]

PR #5263, opened 2017-09-08
Revise approach to evaluation-time type errors

### LINKED ISSUE #5234: Revise approach to handling expression type errors

anandthakker:
Closes #5234

Changes:
 * `get` now returns `null` for a missing property value, rather than throwing an error
 * `coalesce` now does NOT handle type errors -- it only coalesces `null`
 * Basic type assertions (`string`, `number`, `boolean`) now accept any number of arguments -- each one is tested in turn until the first value of the desired type is found.  If the final value doesn't match, then an error is thrown.
 * `to-number` and `to-color` work the same way as the assertions, accepting any number of arguments and attempting to coerce each one in turn until one works

benchmark | master 6e2fbf4 | expressions-type-errors 6d78dbe
--- | --- | ---
**map-load** | 181 ms  | 132 ms 
**style-load** | 150 ms  | 190 ms 
**buffer** | 1,176 ms  | 1,158 ms 
**tile_layout_dds** | 1,372 ms  | 1,381 ms 
**fps** | 60 fps  | 59 fps 
**frame-duration** | 7.1 ms, 1% > 16ms  | 7.3 ms, 2% > 16ms 
**query-point** | 1.00 ms  | 1.18 ms 
**query-box** | 80.64 ms  | 80.97 ms 
**geojson-setdata-small** | 10 ms  | 10 ms 
**geojson-setdata-large** | 191 ms  | 125 ms 
**filter** | Done  | Done 

cc @jfirebaugh @mourner


src/style-spec/function/evaluation_context.js
```diff
@@ -99,72 +95,107 @@ module.exports = () => ({
...
+        let error;
+        for (let i = 0; i < arguments.length; i++) {
+            input = arguments[i];
+            error = null;
+            if (typeof input === 'string') {
+                const c = this._parseColor(input);
+                if (c) return c;
+            } else if (Array.isArray(input)) {
+                error = this._validateRGBA(input[0], input[1], input[2], input[3]);
+                if (!error) return this.rgba(input[0], input[1], input[2], input[3]);
```
mourner: This looks redundant because `this.rgba` does the same validation check.

src/style-spec/function/evaluation_context.js
```diff
@@ -99,72 +95,107 @@ module.exports = () => ({
...
-    parseColor: function (input: string) {
+    _parseColorCache: ({}: {[string]: ?Color}),
+    _parseColor: function (input: string): ?Color {
         let cached = this._parseColorCache[input];
         if (!cached) {
             const c = parseColor(input);
-            if (!c)
-                throw new RuntimeError(`Could not parse color from value '${input}'`);
-            cached = this._parseColorCache[input] = new Color(...c);
+            cached = this._parseColorCache[input] = c ? new Color(...c) : null;
```
mourner: spread arguments with new alert

src/style-spec/function/evaluation_context.js
```diff
@@ -99,72 +95,107 @@ module.exports = () => ({
...
+        // use arguments object instead of rest args because buble translation
+        // of rest args allocates an array
+        let value;
+        for (let i = 0; i < arguments.length; i++) {
+            value = arguments[i];
+            if (value === null) continue;
+            const num = Number(value);
+            if (isNaN(num)) continue;
+            return num;
+        }
```
mourner: Is there any way to avoid relying on `arguments` in a runtime-called method? Perhaps we could cache the possible values in an array like we do for stop outputs (and get rid of this method in the eval context)? Let's check and maybe benchmark to be sure it's faster.

src/style-spec/function/index.js
```diff
@@ -51,52 +52,76 @@ type FunctionParameters = DataDrivenPropertyValueSpecification<StylePropertyValu
...
+        if (compiled.result === 'success') {
+            const warningHistory: {[key: string]: boolean} = {};
+            evaluate = function (globalProperties: {+zoom?: number}, feature?: Feature) {
+                try {
+                    const val = compiled.function(globalProperties, feature);
+                    if (val === null || val === undefined) {
+                        return defaultValue;
+                    }
+                    return val;
+                } catch (e) {
```
mourner: Is there any way we could avoid a full `try/catch` here? Maybe passing an `errors` array to the compiled function and accumulating errors there somehow?

anandthakker: Definitely possible, but I think it's outside the scope of this PR -- I experimented with a non-try catch approach recently, and it's definitely a nontrivial change.  The naive approach (checking for errors before evaluating any expression) was slower than try/catch, and the non-naive approach will take some careful work.

src/style-spec/function/evaluation_context.js
anandthakker: Oops, thanks.

src/style-spec/function/evaluation_context.js
anandthakker: 👍, good call, will try that out.

@mourner btw, I was going off of the rule of thumb that using `arguments` is fine if you never use the object itself, only access `arguments.length` and `arguments[i]` -- what's your take on that?

src/style-spec/function/index.js
mourner: @anandthakker I'm not sure how much exactly it affects performance (I guess it depends on browser engine/version), but as a rule of thumb, I avoid `arguments` keyword altogether in perf-sensitive code paths. At least it is easier to make sure a certain function is monomorphic.

anandthakker:
cc @aparlato @nickidlugash

src/style-spec/function/evaluation_context.js
anandthakker: @mourner fixed -- perf was equivalent, this way seems cleaner/better anyway

src/style-spec/function/evaluation_context.js
```diff
@@ -118,47 +118,80 @@ module.exports = () => ({
...
     rgba: function (r: number, g: number, b: number, a?: number) {
-        ensure(r >= 0 && r <= 255 &&
-            g >= 0 && g <= 255 &&
-            b >= 0 && b <= 255, `Invalid rgba value [${[r, g, b, a || 1].join(', ')}]: 'r', 'g', and 'b' must be between 0 and 255.`);
-        ensure(typeof a === 'undefined' ||
-            (a >= 0 && a <= 1), `Invalid rgba value [${[r, g, b, a || 1].join(', ')}]: 'a' must be between 0 and 1.`);
+        const error = this._validateRGBA(r, g, b, a);
+        if (error) throw new RuntimeError(error);
         return new Color(r / 255, g / 255, b / 255, a);
     },
```
mourner: Is this method still used anywhere?

anandthakker: yeah, it's used for evaluating the `["rgba", r, g, b, a]` (and `"rgb"`) expression

src/style-spec/function/evaluation_context.js
```diff
@@ -99,72 +95,103 @@ module.exports = () => ({
...
+        let error;
+        for (let i = 0; i < args.length; i++) {
+            input = args[i]();
+            error = null;
+            if (typeof input === 'string') {
+                const c = this._parseColor(input);
+                if (c) return c;
+            } else if (Array.isArray(input)) {
+                error = this._validateRGBA(input[0], input[1], input[2], input[3]);
+                if (!error) return new Color(input[0] / 255, input[1] / 255, input[2] / 255, input[3]);
```
mourner: Can you replace those 2 lines with a call to `rgba`?

anandthakker: No, because `rgba` throws a `RuntimeError` on failure.

mourner: Oh, nevermind — I just realized `toColor` now acts like coalesce through multiple args...

nickidlugash:
@anandthakker for type assertions and coercions, is `null` an acceptable argument?

anandthakker:
@nickidlugash for assertions, `null` is never accepted: `["string", null]` will produce an error; `["string", null, "foo"]` will fall back go `"foo"`.

For coercions, it depends. `["to-boolean", null]` => false, `["to-string", null]` => `"null"`, but `to-number` and `to-color` will both fail on `null` input value.

src/style-spec/function/evaluation_context.js
```diff
@@ -208,6 +235,15 @@ module.exports = () => ({
         }
 
         return interpolate[resultType](outputLower, outputUpper, t);
+    },
+
+    coalesce(args: Array<Function>) {
+        for (let i = 0; i < args.length - 1; i++) {
```
jfirebaugh: Could be written as:

```js
let result = null;
for (const arg of args) {
    result = arg();
    if (result !== null) break;
}
return result;
```

Which handles the empty array case. (`coalesce` requires one argument currently, but we could consider relaxing this.)

src/style-spec/function/evaluation_context.js
```diff
@@ -208,6 +235,15 @@ module.exports = () => ({
         }
 
         return interpolate[resultType](outputLower, outputUpper, t);
+    },
+
+    coalesce(args: Array<Function>) {
```
jfirebaugh: `Function` → `()=>Value`

src/style-spec/function/definitions/index.js
```diff
@@ -49,46 +53,61 @@ CompoundExpression.register(expressions, {
...
-    'string': defineAssertion(StringType),
-    'number': defineAssertion(NumberType),
-    'boolean': defineAssertion(BooleanType),
-    'object': defineAssertion(ObjectType),
     'to-string': [ StringType, [ValueType], fromContext('toString') ],
-    'to-number': [ NumberType, [ValueType], fromContext('toNumber') ],
-    'to-boolean': [ BooleanType, [ValueType], ([v]) => `Boolean(${v})` ],
-    'to-rgba': [ array(NumberType, 4), [ColorType], ([v]) => `${v}.value` ],
-    'to-color': [ ColorType, [ValueType], fromContext('toColor') ],
+    'to-number': [ NumberType, varargs(ValueType), (ctx, args) => {
```
jfirebaugh: `to-number` and `to-color` now do not necessarily evaluate all their arguments -- i.e. they are special forms. Let's write them as such. I think this will also allow you to avoid all the other `ctx` and `cache` changes to this file.

test/integration/expression-tests/to-string/basic/test.json
```diff
@@ -15,16 +15,6 @@
...
-      "false",
-      "null",
-      {
-        "error": "Expected a primitive value in [\"string\", ...], but found Array<Number, 2> instead."
-      },
-      {
-        "error": "Expected a primitive value in [\"string\", ...], but found Object instead."
-      }
-    ]
+    "outputs": ["1", "false", "null", "[1,2]", "{\"y\":1}"]
```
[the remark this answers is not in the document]
anandthakker: I am, yeah.  It feels to me like `to-string` should always work, and since, with the exception of `Color`, all the values in our system can be represented in JSON just fine, I think it seems fine to commit to `JSON.stringify`.  (Once `Color`s can be contained within objects and arrays, we can use the `replacer` parameter of `JSON.stringify` to deal with those.)

src/style-spec/function/definitions/index.js
anandthakker: Another reason these `ctx` / `cache` changes may be useful is that they'll let us experiment with inlining arguments (rather than always caching each one as a `function eN() { ... }`)

test/integration/expression-tests/to-string/basic/test.json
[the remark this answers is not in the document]
jfirebaugh: I don't think we want to commit to having a string representation of every type we ever add to the expression type system. What, for example, would be the result of `["to-string", ["image", "foo"]]` (https://github.com/mapbox/mapbox-gl-js/issues/5261)?

src/style-spec/function/definitions/match.js
```diff
@@ -107,15 +107,12 @@ class Match implements Expression {
...
-            lookup += `${String(label)}: ${outputs[this.cases[label]]}`;
+            lookup += `${JSON.stringify(label)}: ${outputs[this.cases[label]]}`;
         }
 
         const lookupObject = ctx.addVariable(`{${lookup}}`);
 
-        const inputType = this.inputType.kind !== 'Array' ?
-            `$this.types.${this.inputType.kind}` : JSON.stringify(this.inputType);
-
-        return `(${lookupObject}[$this.as(${input}, ${inputType})] || ${ctx.addExpression(this.otherwise.compile(ctx))})();`;
```
jfirebaugh: What was the purpose of using `$this.as` here before?

src/style-spec/function/definitions/match.js
```diff
@@ -107,15 +107,12 @@ class Match implements Expression {
         for (let i = 0; i < labels.length; i++) {
             if (i > 0) lookup += ', ';
             const label = labels[i];
-            lookup += `${String(label)}: ${outputs[this.cases[label]]}`;
+            lookup += `${JSON.stringify(label)}: ${outputs[this.cases[label]]}`;
```
jfirebaugh: Why `JSON.stringify` rather than `String`? The input is going to implicitly go through `String` in `${lookupObject}[${input}]`, so if they behave differently that would seem to be an issue.

test/integration/expression-tests/to-string/basic/test.json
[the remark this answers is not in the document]
anandthakker: Eh, I think it's okay to come up with some (obviously lossy) way to stringify `Image` values -- `{"image": "foo"}` or something -- but even if not, I'd still be in favor of being as liberal as possible.  If nothing else, I could imagine it being super convenient while debugging to set a symbol's text-field property to `["to-string", ["get", "offset"]]` (when we finally have support for array values in feature properties)

src/style-spec/function/definitions/match.js
anandthakker: The `JSON.stringify()` here is to get a JS string literal for the key in the JS object literal we're building.  If not, then `label === "foo-bar"` would result in invalid code `{ foo-bar: output }`

src/style-spec/function/definitions/match.js
anandthakker: I believe it was always redundant, since type checking should already ensure that the input value is the right type.

src/style-spec/function/definitions/match.js
jfirebaugh: Looks like `JSON.stringify(number)` is equivalent to `String(number)`, so that should be fine. For strings, [U+2028 and U+2029 need special handling](http://timelessrepo.com/json-isnt-a-javascript-subset).

anandthakker: Oy.  We'll need to handle those in `Literal`, too, then.

src/style-spec/function/evaluation_context.js
```diff
@@ -99,72 +95,103 @@ module.exports = () => ({
...
     toString: function(value: Value) {
-        const type = this.typeOf(value);
-        ensure(value === null || /^(String|Number|Boolean)$/.test(type), `Expected a primitive value in ["string", ...], but found ${type} instead.`);
-        return String(value);
+        const type = typeof value;
+        if (value === null || type === 'string' || type === 'number' || type === 'boolean') {
+            return String(value);
+        } else if (value instanceof Color) {
+            const [r, g, b, a] = value.value;
+            return `rgba(${r * 255}, ${g * 255}, ${b * 255}, ${a})`;
```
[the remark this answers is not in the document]
anandthakker: Added in https://github.com/mapbox/mapbox-gl-js/pull/5263/commits/bae93ca3089f6fcfd657b502de910712bd0e4016

src/style-spec/function/definitions/index.js
anandthakker: Moved `to-number` and `to-color` to a special form, `Coercion`.  If/when it becomes possible for `to-string` to fail, we can move it there, too.

src/style-spec/function/parse_expression.js
```diff
@@ -65,33 +64,23 @@ function parseExpression(expr: mixed, context: ParsingContext): ?Expression {
...
-    'Number': 'number',
-    'String': 'string',
-    'Boolean': 'boolean',
-    'Color': 'to-color'
-};
-
 function wrapForType(expected: Type, expression: Expression, context: ParsingContext) {
-    const wrapper = typeWrappers[expected.kind];
-    if (!wrapper) {
+    if (expected.kind === 'Color') {
```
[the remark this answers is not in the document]
anandthakker: 👍 did so.

PR #5266, opened 2017-09-11
Add expressions.html debug page with barebones expression editor

anandthakker:
Run `yarn start-debug` and open http://localhost:9966/debug/expressions.html to view the page.


anandthakker:
FYI @nickidlugash @aparlato  -  this debug page will let you see how old-style stop functions get converted to expressions under the hood

src/style-spec/function/index.js
```diff
@@ -90,6 +90,9 @@ function createFunction(parameters: FunctionParameters, propertySpec: StylePrope
                 throw new Error(f.zoomCurve ? f.zoomCurve.error : 'Invalid zoom expression');
             }
         }
+
+        // useful for debugging, especially for converted stop functions
+        f.rawExpression = expr;
```
mollymerp: should we conditionally save this property (i.e. add a debug argument/option to `createFunction`)? not sure it would make a big difference beyond clarity on what `rawExpression` is used for.

src/style-spec/function/index.js
```diff
@@ -90,6 +90,9 @@ function createFunction(parameters: FunctionParameters, propertySpec: StylePrope
...
         return f;
     } else {
         console.log(JSON.stringify(expr, null, 2));
```
mollymerp: same as above perhaps 💭

anandthakker:
> Also checking that you meant to keep the changes to circles.html from your presentation this morning?

ya, I figured we might as well

debug/expression_editor.js
```diff
@@ -0,0 +1,147 @@
+const createFunction = require('../src/style-spec/function/index');
+const util = require('../src/util/util');
+const spec = require('../src/style-spec/reference/v8.json');
+
+class ExpressionEditor extends React.Component {
```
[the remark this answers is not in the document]
anandthakker: Good point.  You _can_ edit expressions here -- it'll accept constants, old-style functions, _or_ new-style expressions... so maybe `StylePropertyEditor` would be a more appropriate name

src/style-spec/function/index.js
anandthakker: Yeah, I considered that briefly, but AFAIK the current build process doesn't have an easy way to do `if (DEBUG BUILD) {...}`.  We could use `envify`, which we're already using for the benchmarks, but since this seems pretty harmless, I thought maybe it wasn't worth adding a whole transform to the browserify build...

src/style-spec/function/index.js
anandthakker: Oops, yeah this is leftover, will remove

src/style-spec/function/index.js
mollymerp: yeah let's go with not over-complicating things 👌

anandthakker:
@mollymerp addressed your comments ( 🙇 ) -- going to merge now, but also setting a reminder for 3 months from now to remove this: it's useful right now because expressions are new (and with a few parts of the spec still shifting slightly), but once they're "mainstream", this will probably be stale / clutter up the debug directory

Issue #4078, opened 2017-02-01
Make "icon-offset" units be pixels


lucaswoj:
_From @ansis on July 8, 2014 23:26_

from #83 

> The only thing that may cause confusion is that text-offset is in ems, while icon-offset is in pixels. But maybe that's not important.

The current units for icon-offset aren't even exactly pixels. If the icon gets scaled with icon-size then the offset also gets scaled. An icon with `"icon-offset": 6` and `"icon-size": 0.5` actually has an offset of 3 pixels.

For now the units need to be relative to the unscaled size, not absolute values, but maybe we should changes the units of icon-offset to a fraction of the original size.

_Copied from original issue: mapbox/mapbox-gl-style-spec#84_

jfirebaugh:
Closing here; see #6584.

PR #5271, opened 2017-09-12
document units for `icon-offset`

ansis:
Documents that the units for `icon-offset` are px. ~~`em` and explains what `1 em` is for an icon.~~

## Launch Checklist

 - [NA] write tests for all new functionality
 - [NA] document any changes to public APIs
 - [NA] post benchmark scores

@ericdeveloper @mollymerp

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -1026,7 +1026,7 @@
     "icon-offset": {
       "type": "array",
       "value": "number",
-      "units": "pixels",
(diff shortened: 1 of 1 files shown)


ansis:
@mollymerp thanks for catching that they are actually pixels! 😬

jfirebaugh:
#4661 says they are px * icon-size. Is this right? (Also, the commit that purportedly fixed that issue seems to have actually been an update to `icon-size` units, not `icon-offset`.

jfirebaugh:
See also #4078.

PR #5274, opened 2017-09-12
Fixed #2784

### LINKED ISSUE #2784: Make double-tapping on make zoom in by a factor of 2 on iOS
In WebKit on iOS, double-tapping the map zooms in by a very small amount, probably a factor of 1.1. Double-tapping should zoom in by a factor of 2, for consistency with double-clicks in desktop browsers and double-taps in the Mapbox iOS SDK.

Reproduced in v0.19.1 in Safari on iOS 9.3.

/cc @lucaswoj

aparajita:
Previously, a zoom initiated by a double tap on a touchscreen device would stop as soon as the user lifted his/her finger from the second tap. This was due to the map's mousedown handler calling `map.stop()`.

This PR sets a flag when the user initiates a double click/tap, which is then checked before calling `map.stop()` in the map's mousedown handler.

mousedown events are still received and passed to any handlers during a user-initiated zoom animation, but they will not stop the zoom.

### DIFF
--- a/src/ui/bind_handlers.js
+++ b/src/ui/bind_handlers.js
@@ -46,7 +46,7 @@ module.exports = function bindHandlers(map: Map, options: {}) {
     }
 
     function onMouseDown(e: MouseEvent) {
-        if (!map._isUserDoubleClick) {
+        if (!map.doubleClickZoom.isActive()) {
             map.stop();
         }
 
--- a/src/ui/bind_handlers.js
+++ b/src/ui/bind_handlers.js
@@ -46,7 +46,10 @@ module.exports = function bindHandlers(map: Map, options: {}) {
     }
 
     function onMouseDown(e: MouseEvent) {
-        map.stop();
+        if (!map._isUserDoubleClick) {
+            map.stop();
+        }
+
         startPos = DOM.mousePos(el, e);
         fireMouseEvent('mousedown', e);
 
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -244,7 +244,6 @@ class Map extends Camera {
     _refreshExpiredTiles: boolean;
     _hash: Hash;
     _delegatedListeners: any;
-    _isUserDoubleClick: boolean;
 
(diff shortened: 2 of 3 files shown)


src/ui/bind_handlers.js
```diff
@@ -46,7 +46,10 @@ module.exports = function bindHandlers(map: Map, options: {}) {
     }
 
     function onMouseDown(e: MouseEvent) {
-        map.stop();
+        if (!map._isUserDoubleClick) {
```
mollymerp: instead of adding this property to the map object, you could create a `DoubleClickZoomHandler#isActive` method as we do in with some of the other handles (ex. https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/handler/box_zoom.js#L52), so you could call `map.doubleClickZoom.isActive()` for this conditional

aparajita: Good idea, I'll switch to using `isActive`. As for the test, there aren't any existing tests for the DoubleClickZoomHandler, and I'm kind of a little too busy to figure out how to write one to catch this specific bug. Better someone who knows the test framework well should do it.

mollymerp:
Thanks for the update @aparajita. I started writing a test but it turns out that the `TouchEvent` constructor is not implemented in JSDOM, so I'm not sure if writing a meaningful test is possible at this time. 

@jfirebaugh do you have other ideas for testing this?

src/ui/map.js
```diff
@@ -244,6 +244,7 @@ class Map extends Camera {
     _refreshExpiredTiles: boolean;
     _hash: Hash;
     _delegatedListeners: any;
+    _isUserDoubleClick: boolean;
```
mollymerp: I don't think we need this flag anymore

mollymerp:
@aparajita once you remove the unneeded flag on the `map` object, I will merge this as is and open a follow-up ticket on testing.

src/ui/map.js
aparajita: Done.

PR #5279, opened 2017-09-13
Fold compound expressions with literal arguments into literals

anandthakker:
When parsing an expression, if its arguments are all Literals, then evaluate it immediately and yield a Literal instead.

Yields a modest perf improvement in `tile_layout_dds`:
 - before: 51.37629 - 52.20937 ms / iteration
 - after: 47.81951 - 49.86712 ms / iteration

More importantly, this means we get validation errors at compile time rather than evaluation time for the common case of color curves defined like `[curve, [linear], [zoom], 0, "red", 10, "bleu mispelled"]`

benchmark | master 2200752 | expressions-fold-constant 0567a8f
--- | --- | ---
**map-load** | 111 ms  | 119 ms 
**style-load** | 128 ms  | 168 ms 
**buffer** | 1,167 ms  | 1,120 ms 
**tile_layout_dds** | 1,329 ms  | 1,255 ms 
**fps** | 60 fps  | 60 fps 
**frame-duration** | 7.1 ms, 1% > 16ms  | 6.9 ms, 2% > 16ms 
**query-point** | 0.98 ms  | 1.24 ms 
**query-box** | 73.43 ms  | 75.72 ms 
**geojson-setdata-small** | 17 ms  | 0 ms 
**geojson-setdata-large** | 88 ms  | 175 ms 
**filter** | Done  | Done

### DIFF
--- a/src/style-spec/function/definitions/literal.js
+++ b/src/style-spec/function/definitions/literal.js
@@ -4,7 +4,7 @@ const { Color, isValue, typeOf } = require('../values');
 
 import type { Type } from '../types';
 import type { Value }  from '../values';
-import type { Expression, ParsingContext }  from '../expression';
+import type { Expression, ParsingContext, CompilationContext }  from '../expression';
 
 const u2028 = /\u2028/g;
 const u2029 = /\u2029/g;
@@ -45,9 +45,19 @@ class Literal implements Expression {
         return new Literal(context.key, type, value);
     }
 
-    compile() {
-        const value = Literal.compile(this.value);
-        return typeof this.value === 'object' ?  `(${value})` : value;
+    compile(ctx: CompilationContext) {
+        let value;
+        if (this.type.kind === 'Color') {
+            value = `(new $this.Color(${(this.value: any).join(', ')}))`;
+        } else {
+            value = Literal.compile(this.value);
+        }
+
+        if (typeof this.value === 'object' && this.value !== null) {
+            return ctx.addVariable(value);
+        } else {
+            return value;
+        }
     }
 
     static compile(value: Value) {
--- a/src/style-spec/function/definitions/literal.js
+++ b/src/style-spec/function/definitions/literal.js
@@ -72,7 +72,7 @@ class Literal implements Expression {
         }
     }
 
-    accept(visitor: Visitor<Expression>) { visitor.visit(this); }
+    eachChild() {}
 }
 
 module.exports = Literal;
--- a/src/style-spec/function/evaluation_context.js
+++ b/src/style-spec/function/evaluation_context.js
@@ -62,6 +62,7 @@ function ensure(condition: any, message: string) {
 module.exports = () => ({
     types: types,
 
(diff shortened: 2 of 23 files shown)


anandthakker:
Example generated code for a color curve with `"red"` as the stop output value (excerpt of the `curve/linear-color` integration test output)

## before

```js
function e4() {   return "red" }
var v5 = [e4];
function e6() {   return $this.toColor(v5) }
```

## after

```js
var v4 = (new $this.Color(1, 0, 0, 1));
function e5() {   return v4 }
```

mourner:
Very cool!

anandthakker:
Rebased, generalized, and updated benchmarks.

src/style-spec/function/evaluation_context.js
```diff
@@ -13,7 +13,6 @@ const {
     toString,
     checkSubtype} = require('./types');
 const {Color, typeOf} = require('./values');
-const Curve = require('./definitions/curve');
```
jfirebaugh: What's the reason for these changes?

src/style-spec/function/definitions/literal.js
```diff
@@ -45,9 +45,20 @@ class Literal implements Expression {
...
+    compile(ctx: CompilationContext) {
+        let value;
+        if (this.type.kind === 'Color') {
+            value = `(new $this.Color(${(this.value: any).join(', ')}))`;
+        } else {
+            value = Literal.compile(this.value);
+        }
+
+        if (typeof this.value === 'object' && this.value !== null) {
+            const v = ctx.addVariable(value);
```
jfirebaugh: Nit: `return ctx.addVariable(value);`

src/style-spec/function/parse_expression.js
```diff
@@ -68,4 +87,21 @@ function parseExpression(expr: mixed, context: ParsingContext): ?Expression {
     }
 }
 
+const nonConstantExpressions = [ 'error', 'get', 'has', 'properties', 'id', 'geometry-type', 'zoom' ];
```
jfirebaugh: This looks similar to `isFeatureConstant`/`isZoomConstant`. Can they share more code?

src/style-spec/function/evaluation_context.js
anandthakker: Breaking a new dependency cycle, which was introduced by `parse_expression` now depending on `evaluation_context`.  I guess it's better to break it there instead (especially since we're already doing so in that module)

src/style-spec/function/parse_expression.js
anandthakker: 👍 I almost did that, and then stopped myself because I thought I was prematurely abstracting.

anandthakker: Oh, actually the code itself is a bit different. `isConstant` only looks one level deep (because that's sufficient for the way we're using it).  We could consolidate the base case code, though

anandthakker:
@jfirebaugh updated to address your notes/questions

PR #5282, opened 2017-09-13
Sync changes from 0.40 release back to master

anandthakker:
All but the last three are bringing over changes from `mb-pages`.  Normally, that would have happened in the release process, but it didn't this time because I used a non-master branch.  Pulling these in now so that we can safely force-push onto mb-pages next time.

Last couple commits are just changelog & package.json


jfirebaugh:
I advocate a regular merge of `mb-pages` to `master` instead: https://github.com/mapbox/mapbox-gl-js/compare/merge-mb-pages.

anandthakker:
@jfirebaugh 👍 took your merge commit and just added b23226c (the latter was a fixup that I tacked onto `release-0.40` but hadn't bothered pushing back to mb-pages)

anandthakker:
> Will need to be merged from the command line rather than GitHub

@jfirebaugh GitHub seems to be offering a Squash and merge option here... but are you thinking we should merge manually to preserve the history?

jfirebaugh:
Yeah, both of the merge button options would discard the actual merge commit, which we don't want to do. In this case we just want to fast forward `master` on the command line to b23226c, and then push it -- I went ahead and did so.

Issue #4807, opened 2017-06-08
client-side terrain: scale slope calculations by latitude


ajashton:
@mollymerp @kkaefer I brought this up recently in chat but figured I should capture details in a ticket.

Slope calculations for terrain visualization from mercator-projected tiles should ideally to take into account the varying scale along latitude. The ground distance covered by 1 pixel in mercator shrinks as you move away from the equator, so hills at extreme latitudes will appear more flat than they really are if this is not accounted for.

[This StackExchange answer](https://gis.stackexchange.com/questions/14750/using-srtm-global-dem-for-slope-calculation/40464#40464) explains the situation well and includes some math that can be used to adjust the terrain visualization to be more accurate.

> …assuming scale is correct at the Equator, its distortion equals the secant of the latitude. That is, distances on the map appear to be multiplied by the secant.

| Horizontal scale factor | approx. latitude |
| --- | --- |
| 1.25× | ±37° |
| 1.5× | ±48° |
| 2× | ±60° | 
| 3× | ±70° |
| 4× | ±75° | 
| 5× | ±78.5° |

Options:

- Compensate for latitude on a per-pixel basis. Most accurate but probably difficult or expensive.
- Compensate for latitude on a per-tile basis (eg adjust all pixels in a tile based on the lat of the tile center). This would require Some gradual falloff at the most zoomed out levels to avoid obvious seams between tiles. A mostly-acceptable tradeoff, but unfortunate for places like Svalbard.
    - This is the option we went with for vector terrain
- Do nothing. A hillshade style designed with 20-40° of latitude in mind would slightly over-exaggerate slopes in equatorial regions and under-exaggerate them at more extreme latitudes, but the error would not be very noticeable at most populated latitudes. Northern North America & northern Europe would lose out on accuracy in the most noticeable way, with perhaps also southern South America and New Zealand.

mollymerp:
Thanks @ajashton for capturing in a ticket. Very interesting stack exchange article! I think per-pixel basis might be more trouble than its worth but I don't think per-tile scaling would be too difficult. 

@kkaefer let me know if you have thoughts on this or if you have explored any of these options previously.

kkaefer:
[...]

PR #5286, opened 2017-09-14
add raster-dem source type and hillshade layer type

mollymerp:
Moving to a fresh PR in place of #4701 

This PR implements client-side hillshading using the Mapbox Terrain RGB tileset – right now it is not extensible to other DEM encoding formats. 

Todo:
- [x] implement performant framebuffer binding per https://github.com/mapbox/mapbox-gl-js/pull/4701#issuecomment-308182066 with the new types of render passes ( #5101 )
- [x] remove shader debug code
- [x] scale slope by latitude #4807 
- [x] rebase once #5105 is merged to solve flickering on tile load 🙏 
- [x] use new `Texture` + maybe `RenderTexture` classes
- [x] add more unit tests
- [x] profile to see whether worker source is necessary
- [x] fix handling of missing tiles (persistent border artifacts)
- [x] improve solution for border flicker when tiles' neighbors haven't been completely backfilled
- [x] debug tile draw failures
- [x] debug android render failure
- [x] decide how we want to handle ensuring we use `.pngraw` when requesting terrain-rgb tiles
- [x] get shader/rendering input from @nickidlugash 
- [x] code reviews 🔍 📝 

cc @kkaefer 

## Launch Checklist

 - [x] post benchmark scores


mollymerp:
@kkaefer making progress with testing + getting through the backlog. The raster masking did fix the overlapping tile flashes, but there is still flickering coming from the texture getting updated with more border data as neighboring tiles load. Tomorrow, I will be focusing on alternative solutions to that tile flickering and writing more unit tests, but whenever you have time to review the PR in its current state that would be much appreciated. Thanks! 

I may have questions about performance + profiling as well as I try to get this through the finish line.

mollymerp:
note that hillshade layer types aren't rendering and don't produce console errors on the 2 android devices that I've tested it on. will try and get my hands on a test device to debug more thoroughly ASAP

mollymerp:
> In Safari and Firefox, rendering looks weird (Chrome looks fine):

> When loading the map, we seem to be double-drawing the tiles for a few frames:

these were both issues w the debug page (not ensuring the file extension of the terrain source was `pngraw` and changing the light on `load`.

> There appears to be a drawing problem (possibly related with the double draw?) when overzooming tiles. These frames are select from a zoom in animation:

fix for this is up at #5380 

(i've collapsed some of the rendering images already being addressed into `details` for scrolling brevity)

will be working on the tile border issues and the disappearing child tiles tomorrow.

mollymerp:
I can't reproduce the parent tile eviction on _zoom in_ but I did find an issue where we were evicting child tiles prematurely on _zoom out_ if the zoom out is > 1 zoom level. I think this is a bug in native as well.

mollymerp:
I tested out a new approach to tile borders that I think looks pretty good compared to the flashing white borders – when a tile first loads I create a 1px border around the original image populated with the elevation data from the nearest pixel with real data, and then I replace this value with the real neighboring tile data once it loads. I think it looks a lot better than the previous approach, and it solves the issue of flashing / persistent white seams.

mollymerp:
@nickidlugash when you have a sec to look at the shaders / default style-spec properties that would be great. you'll need to pull down this branch and then you can run `yarn start-debug` and load http://localhost:9966/debug/terrain-rendering.html in your browser. I'm particularly interested in what you think about the intensity by zoom level and what happens at points of extreme slope (there are some craters in the Himalayas that are good examples of this as well as the grand canyon).

To modify the intensity / zoom function you'll want to play around with [this line](https://github.com/mapbox/mapbox-gl-js/pull/5286/files#diff-2869d624151d9150a5f8f68cf567cb6fR38), and then the final shading color is [set here](https://github.com/mapbox/mapbox-gl-js/pull/5286/files#diff-39cc0956b79a7315e53d81a9a3efe2bfR48)

let me know if you have any questions! 

_craters in the Himalayas_
pasted image at 2017_10_03 05_40 pm

mollymerp:
@kkaefer today I was able to isolate the missing tiles to an issue with the `terrainPrepare` step where apparently the data we're putting in the texture is incomplete/invalid/not ready on the first frame that tries to render a tile. 

If you change the code so that `terrainPrepare` runs on every frame, the missing tiles eventually appear. 

When I render out the raw pixel data we're passing into the `tile.demTexture` I get black squares, but when I log out the `pixelData` I'm not seeing any tiles with missing/weird looking values. 

image

src/source/raster_dem_tile_source.js
```diff
@@ -0,0 +1,130 @@
...
+                    uid: tile.uid,
+                    coord: tile.coord,
+                    source: this.id,
+                    rawImageData: rawImageData
+                };
+
+                if (!tile.workerID || tile.state === 'expired' || tile.state === 'unloaded') {
+                    tile.workerID = this.dispatcher.send('loadDEMTile', params, done.bind(this));
+                }
+            }
```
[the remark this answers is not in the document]
mollymerp: I'm pretty sure that `ajax#getImage` prevents a case where there'd be no `img` _and_ no `err`

mollymerp:
💀  the missing/flickering tiles were due to `UNPACK_PREMULTIPLY_ALPHA_WEBGL` being set to `true` elsewhere in the rendering loop. 

I also was able to fix the rendering failure on android by setting precision to `highp` in the shaders. 

as I see it the outstanding items (copied from above) are 

- [ ] decide how we want to handle ensuring we use `.pngraw` when requesting terrain-rgb tiles
- [ ] get shader/rendering input from @nickidlugash 
- [ ] code reviews 🔍 📝 
- [ ] bonus: benchmarking?

MythicLionMan:
Sorry to butt in here as a user, but I installed the demo and tried this out on a few areas that have low hills that are of particular interest to myself and had a bit of feedback on the intensity of the shading. I tried with the default scaling of pow(2.0, 14.5 - u_zoom * 0.5) and found that low hills almost disappeared at high zoom levels. I played around a bit and found pow(2.0, 12.0 - u_zoom * 0.5) worked much better. 

For instance see [here](http://localhost:9966/debug/terrain-rendering.html#10.68/46.083/-81.452) and [here](http://localhost:9966/debug/terrain-rendering.html#10.6/47.1587/-80.1874). With a base value of 14.5 the terrain effect is very subtle, whereas with a value of 12.0 it is much more noticeable. For comparison see the same two areas in Google terrain [here](https://www.google.ca/maps/@46.0773038,-81.4341036,11.96z/data=!5m1!1e4) and [here](https://www.google.ca/maps/@47.1225866,-80.2276705,11.75z/data=!5m1!1e4).

Figuring out what the function should be to work at all zoom levels is beyond me, but I do think that taking the needs of rendering gentler terrain into account is important.

BTW: Overall the shaded terrain is a huge improvement and I'm looking forward to it's being deployed. I've used Mapbox for some rendering in the areas specified and was never happy with the results, but this is much more promising.

Killarney In Mapbox Hillshade with pow(2.0, 12.0 - u_zoom * 0.5):
Killarney In Mapbox Hillshade

Killarney In Mapbox Hillshade with pow(2.0, 14.5 - u_zoom * 0.5):
screen shot 2017-10-06 at 11 59 11 am

Killarney In Google Maps:
Killarney In Google Maps

mollymerp:
@MythicLionMan thanks for the feedback! definitely not butting in. Our cartography team hasn't yet reviewed the shaders for this layer type so they are definitely not final, and we will take your comments into consideration!

src/source/raster_dem_tile_source.js
```diff
@@ -0,0 +1,128 @@
...
+            neighboringTiles[getTileId({ z: z, x: x, y: y + 1, w: w  })] = {backfilled: false};
+            neighboringTiles[getTileId({ z: z, x: nx, y: y + 1, w: nxw  })] = {backfilled: false};
+        }
+
+        return neighboringTiles;
+
+        function getTileId(coord) {
+            const tilecoord = new TileCoord(coord.z, coord.x, coord.y, coord.w);
+            return tilecoord.id;
+        }
```
anandthakker: ^ instead of constructing all these new objects, let's extract the TileCoord constructor code that calculates the id from `(z,x,y,w)` as a static method so we can just do `neighboringTiles[TileCoord.idFromCoordinates(z, px, y - 1, pxw)]`

src/source/tile.js
```diff
@@ -74,12 +75,17 @@ class Tile {
...
     vtLayers: {[string]: VectorTileLayer};
     mask: Mask;
+
+    neighboringTiles: ?Object;
+    dem: ?DEMData;
     aborted: ?boolean;
     maskedBoundsBuffer: ?VertexBuffer;
     maskedIndexBuffer: ?IndexBuffer;
     segments: ?SegmentVector;
+    prepared: ?boolean
```
anandthakker: With an eye towards https://github.com/mapbox/mapbox-gl-js/issues/5444, could we name this flag something specific to DEM tiles, so that it's easier to identify / pull out later?  E.g. `needsHillshadePrepare` or something

test/integration/render-tests/hillshade-accent-color/literal/style.json
```diff
@@ -0,0 +1,38 @@
...
+      "paint": {
+        "background-color": "white"
+      }
+    },
+    {
+      "id": "hillshade",
+      "type": "hillshade",
+      "source": "source",
+      "paint": {
+        "hillshade-accent-color": "purple"
```
anandthakker: Just for fun I tried changing this to "red" and discovered that, while I can see the difference in the image, it still passed the render test -- I guess the difference was within the threshold.  Is there an image/location we could use for the hillshade-accent  that would have more pronounced accents so that they're more likely to catch subtle regressions?

src/data/dem_data.js
```diff
@@ -0,0 +1,170 @@
...
+    idx(x: number, y: number) {
+        if (x < -this.border || x >= this.width + this.border ||  y < -this.border || y >= this.height + this.border) throw new RangeError('out of range source coordinates for DEM data');
+        return (y + this.border) * this.stride + (x + this.border);
+    }
+}
+
+class DEMData {
+    uid: string;
+    scale: number;
+    data: Array<Level>;
```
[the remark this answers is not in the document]
mollymerp: yeah I kept going back and forth on this – but it is confusing as is, so I'll change it 👍

src/data/dem_data.js
```diff
@@ -0,0 +1,170 @@
...
+    constructor(uid: string, scale: ?number, data: ?Array<Level>) {
+        this.uid = uid;
+        this.scale = scale || 1;
+        this.data = data ? data : [];
+        this.loaded = !!data;
+    }
+
+    loadFromImage(data: RGBAImage) {
+        // Build level 0
+        this.data = [new Level(data.width, data.height, data.width / 2)];
```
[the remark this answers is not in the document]
mollymerp: data.width and data.height should always be the same, and the border is needed because texture dimensions have to be power of 2, so we put a border of 1/2 the original dimension on each side to effectively double the height and width. this should be clearer.

src/render/draw_hillshade.js
```diff
@@ -0,0 +1,168 @@
...
+        // tiles will appear blank, because as you can see above the alpha value for these textures
+        // is always 0
+        gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, (false: any));
+
+
+        tile.demTexture = tile.demTexture || painter.getTileTexture(tile.tileSize);
+        if (tile.demTexture) {
+            tile.demTexture.update(pixelData[0], false);
+            // for some reason flow thinks tile.demTexture can be null or undefined here :sob:
+            (tile.demTexture: any).bind(gl.NEAREST, gl.CLAMP_TO_EDGE);
```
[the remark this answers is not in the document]
mollymerp: awesome! thank you

src/data/dem_data.js
```diff
@@ -0,0 +1,170 @@
...
+        for (let x = 0; x < data.width; x++) {
+            // left vertical border
+            level.set(-1, x, level.get(0, x));
+            // right vertical border
+            level.set(data.width, x, level.get(data.width - 1, x));
+            // left horizontal border
+            level.set(x, -1, level.get(x, 0));
+            // right horizontal border
+            level.set(x, data.width, level.get(x, data.width - 1));
+        }
```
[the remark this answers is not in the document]
mollymerp: updated to ensure Level is always square!

src/data/dem_data.js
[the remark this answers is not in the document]
anandthakker: Ah, interesting!

nickidlugash:
@mollymerp This looks great 👍  I reviewed the branch with a focus on both overall visual display and user needs. I've made some suggestions in case it's helpful!

### Spec properties

As discussed previously with @mollymerp, my initial reaction was that it would make more sense from a user perspective to make the spec properties consistent with the extrusion properties (assigning a single `-color` property, and then using the global `light: {color}`) to adjust the colors of the highlights and lowlights). After spending some time with this branch, however, I think these hillshade properties generally make sense. The extrusion properties are consistent with how a user would expect to interact with a 3D rendering, and the hillshade properties are consistent with how a user would expect to interact with a terrain respresentation calculated from a DEM. There are different precedents for these two layer types, so I think the inconsistencies between the two are fine.  

### Accent color

Perhaps `hillshade-accent-color` would be more usefully named `hillshade-slope-color`? I think the specificity would be helpful to users, both those with and without familiarity of terrain rendering from other software.


### Highlights and shadows

I think overall the shading looks great. 

Regarding the [extreme areas that you pointed out](https://github.com/mapbox/mapbox-gl-js/pull/5286#issuecomment-334023276), @ajashton noted that these seem to be areas with missing data.

I noticed two (related) areas where there could be some visual improvement:

- The mix of the shade and highlight color can often result in intermediary colors that look gray or brown, which doesn't always look that good:

screen shot 2017-10-25 at 9 03 44 pm

- White (or very light) highlight colors don't seem visible enough (unless the color underneath is very dark). Here's white highlights over a light green vs a dark green background:

screen shot 2017-10-25 at 5 26 24 pm
screen shot 2017-10-25 at 5 27 04 pm


  I think that it'll be fairly common for users to want a light-to-medium toned map with dark shadows and light highlights, and expect the highlights to be fairly visible over the background color.

Right now the final shade color is a linear mix of the shadow and highlight color. I think this is causing the muddiness in intermediary colors (since the rgb mix of two opposite colors is a neutral color), and also the lack of visibility for white/very light highlights (since intermediary colors will appear more similar to the less neutral color, and also more similar the the color that has more contrast with the background color). 

I'm not sure if these suggestions are at all practical, but just to throw out a few ideas:

- Right now the alpha value of a pixel is adjusted based on the slope and polar angle. Perhaps we can include an adjustment based on the shade value, so that the more equally interpolated a color gets, the lower the alpha value? The idea would be that the intermediary colors look less gray and more similar to the background.

- Instead of having two shade color properties, perhaps we could have a single property that requires a color ramp with an arbitrary number of stops (2 minimum)? Perhaps similar to the heatmap color implementation with a custom expressions curve? This could allow users to move the stop positions of the highlight and shadow, add intermediary stops, or make the interpolation exponentional. Since the visual display is so dependent on which particular colors a user chooses for both the shading and the rest of the map, this could allow more control in making different color combinations look good.

### Different default colors?

Unless there's a strong rationale behind the current default colors, I wonder if changing them to just be grayscale might be a simpler default:

- `hillshade-highlight-color`: #ffffff
- `hillshade-shadow-color`: #000000
- `hillshade-accent-color`: #000000

### Vertical exaggeration by zoom level

I think right now the way this is written, when zl < 9, there is a vertical exaggeration, and when zl >= 9, there is a vertical reduction – is this intentional (to minimize the appearance of hillshading at higher zl, since not all regions have consistently high-res source data)? If not, then I think that perhaps z14 could have no exaggeration/reduction, and lower zooms could have even more exaggeration than they do currently. I'll comment more in the specific line in the code.

### Light position

The default azimuth is 210deg (which I believe is roughly SW or lower-left of viewport). I'm not sure about this, but to me it almost looks like the light is coming from the opposite direction (upper-right)?

### Light intensity / visibility of terrain

Right now the light intensity only affects `hillshade-accent-color` (the slope color). Would it make sense for it to affect `hillshade-highlight-color` and `hillshade-shadow-color` as well, so that users can control how bold/intense to make the hillshading? You can already adjust both the alpha values of the 3 colors as well as `light-position[2]` (the polar angle)  to adjust the proportion of shading to slope.

I tested a basic implementation, replacing:

```
vec4 shade_color = mix(u_shadow, u_highlight, shade) * slope * sin(polar);
```
with:

```
vec4 shade_color = mix(u_shadow, u_highlight, shade) * slope * sin(polar) * mix(0.0, 2.0, u_light.x);
```
in `hillshade.fragment.glsl`, and it seems to work pretty well. I think this would be essentially the same as providing a scale-factor property that enables arbitrary linear scaling of elevation. (The other option might be to have an explicit "scale-factor" property and to disable the light intensity?). Either way, I think it's useful to enable users to adjust the visibility of the terrain.

### Visual tests?

@lbud wrote a really helpful visual test when we were working on the fill-extrusion shaders, that took arrays of values for several properties and outputted an image for each combination – it was thousands of images, but could be scanned fairly quickly for any particularly ugly or unexpected results. If you want to do something similar, I've provided a list of properties and values that could be used, and I'd be happy to review the results. I couldn't find the code for the test, but maybe @lbud can share it?  

- light-position:
	- default, 0deg polar, 90deg polar	 
- light-intensity:
	- 0, 0.25, 0.5, 0,75, 1 
- hillshade-highlight-color:
	- white, white 50% opacity, yellow, light yellow, pink, light blue 
- hillshade-shadow-color:
	- black, black 50% opacity, dark blue, bright blue, dark red, bright red
- hillshade-accent-color:
	- black, black 50% opacity, dark blue, bright blue, dark red, bright red
- background-color:
	- white, light green, dark green, black

```
light-position: [
	[1.15, 210, 30],
	[1.15, 210, 0],
	[1.15, 210, 90]
]

light-intensity: [
	0, 0.25, 0.5, 0.75, 1
]

hillshade-highlight-color: [
	rgba(255, 255, 255, 1),
	rgba(255, 255, 255, 0.5),
	rgba(255, 255, 0, 1),
	rgba(255, 255, 210, 1),
	rgba(255, 160, 220, 1),
	rgba(230, 230, 255, 1)
]

hillshade-shadow-color: [
	rgba(0, 0, 0, 1),
	rgba(0, 0, 0, 0.5),
	rgba(0, 0, 128, 1)
	rgba(0, 0, 255, 1),
	rgba(85, 0, 0, 1),
	rgba(255, 0, 0, 1)	
]

hillshade-accent-color: [
	rgba(0, 0, 0, 1),
	rgba(0, 0, 0, 0.5),
	rgba(0, 0, 128, 1)
	rgba(0, 0, 255, 1),
	rgba(85, 0, 0, 1),
	rgba(255, 0, 0, 1)
]

background-color: [
	rgba(255, 255, 255, 1),
	rgba(180, 255, 210, 1)
	rgba(30, 100, 60, 1),
	rgba(0, 0, 0, 1)
]
```

1ec5:
> As discussed previously with @mollymerp, my initial reaction was that it would make more sense from a user perspective to make the spec properties consistent with the extrusion properties (assigning a single `-color` property, and then using the global `light: {color}`) to adjust the colors of the highlights and lowlights). After spending some time with this branch, however, I think these hillshade properties generally make sense. The extrusion properties are consistent with how a user would expect to interact with a 3D rendering, and the hillshade properties are consistent with how a user would expect to interact with a terrain respresentation calculated from a DEM. There are different precedents for these two layer types, so I think the inconsistencies between the two are fine. 

Do the differences between extrusion lighting and terrain lighting stem from a need for different defaults or from a need for different concepts entirely? If we keep two lighting concepts around, are there ways to assist the developer in keeping them in sync, so that they don’t fall into the trap of producing inconsistent lighting between the ground and buildings?

nickidlugash:
> Do the differences between extrusion lighting and terrain lighting stem from a need for different defaults or from a need for different concepts entirely?

Different concepts, IMO. Pixel colors for extrusions are mainly based on the surface color of the object, the light color, and amount of light hitting that face of the object. Opacity is uniform. Pixel colors for terrain are assigned based on the horizontal angle of the hill face relative to the light and an additional accent color based on slope, with opacity affected by the slope and polar angle of light. There isn't a surface color per se, as hillshading allows the color of the background or map features underneath to show through. (@mollymerp pls correct as needed). I'm not that familiar with hillshading techniques, but my sense is that they are designed to make the terrain as understandable as possible rather than to closely mimic what you might actually see in the real world.

> If we keep two lighting concepts around, are there ways to assist the developer in keeping them in sync, so that they don’t fall into the trap of producing inconsistent lighting between the ground and buildings?

Terrain uses the light anchor property, and part of the light position (azimuth and polar angles). The use of these properties is consistent across terrain and extrusions (or at least they should be? I'm not sure if the azimuth angle is consistent between the two right now). I had also proposed light intensity being made consistent between the two. I'm not sure if it's confusing that terrain only uses some of the light properties, but not all.

I think there will be some playing around needed for assigning the various colors across extrusions and terrain, but it seems relatively easy. The hue of the light color roughly corresponds to that of the `hillshade-highlight-color`, and the hue of the `fill-extrusion-color` roughly corresponds to that of the `hillshade-shadow-color`. Since the concepts are different though, I don't think there's necessarily any "inconsistent" way to assign these colors.

mollymerp:
thank you for the feedback @nickidlugash @anandthakker @1ec5 !!

with respect to light position discrepancies – I think this is a mistake on my part. I think I thought that extrusion lighting defaulted to the upper right corner (for viewport alignment) and modified the hillshade layer to behave the same way. although I'm not sure how the default of lower left will look for hillshading, I think it makes most sense to have the lighting position unified across all layers that incorporate lighting into their rendering.

@nickidlugash with respect to renaming `hillshade-accent-color` 👉  `hillshade-slope-color` I find this more ambiguous (is it the shaded part of the slope or the lit part of the slope? In this case I think accent is more clear, but I'm open to other ideas. 

seeing the following as next actions on my end to get this across the finish line:

### spec: 
- [ ] keep current hillshade-specific color properties
- [ ] test alternate defaults:
    - `hillshade-highlight-color`: #ffffff
    - `hillshade-shadow-color`: #000000
    - `hillshade-accent-color`: #000000
- [ ] fix light rendering to match global light object
- [ ] add scale-factor / exaggeration property to adjust terrain contrast

### shader modifications
- [ ] investigate solution to highlight/shadow muddiness and low visibility
    - alpha value adjustment based on the shade value, so that the more equally interpolated a color gets, the lower the alpha value? The idea would be that the intermediary colors look less gray and more similar to the background.
    - try using a color ramp/custom interpolation instead of having two shade color proporties (@nickidlugash what 2 properties are you referring to in this case?)
- [ ] make `light-intensity` property impact `-highlight-color` and `-shadow-color`
    ```glsl
    vec4 shade_color = mix(u_shadow, u_highlight, shade) * slope * sin(polar) * mix(0.0, 2.0, u_light.x);
    ```

### testing
- [ ] script up some automatic testing that outputs images with many different combinations of style-spec properties
- [ ] change render tests to be more sensitive / in a more extreme or subtle terrain area to eliminate false positive results


### final reviews
- [ ] @anandthakker 
- [ ] @kkaefer 
- [ ] @nickidlugash

src/shaders/hillshade_prepare.fragment.glsl
```diff
@@ -0,0 +1,64 @@
...
+    float f = getElevation(v_pos + vec2(epsilon.x, 0), 0.0);
+    float g = getElevation(v_pos + vec2(-epsilon.x, epsilon.y), 0.0);
+    float h = getElevation(v_pos + vec2(0, epsilon.y), 0.0);
+    float i = getElevation(v_pos + vec2(epsilon.x, epsilon.y), 0.0);
+
+    vec2 deriv = vec2(
+        (c + f + f + i) - (a + d + d + g),
+        (g + h + h + i) - (a + b + b + c)
+        // decrease the value u_zoom is subtracted from to intensify terrain appearance
+    ) /  pow(2.0, 14.5 - u_zoom * 0.5);
```
nickidlugash: This number seemed a bit cryptic to me, so I separated out the calculations and made a suggestion for a tweaked value based on an arbitrary value for `a` that looked good to me:

Divide the differences by `8 * pixel size`
where pixel size (aka meters/pixel) is:  
`circumference of the world / (pixels per tile * number of tiles)`

`= 8 * 40075016.6855785 / (512 * pow(2, u_zoom))`
`= pow(2, 3) * pow(2, 25.25619978527) * pow(2, -9) * pow(2, -u_zoom)`
`= pow(2, 19.25619978527 - u_zoom)`

Multiply this number by a scale factor for vertical exaggeration by zoom level,
where the highest zl (z14) is the base and the exaggeration is proportional to zoom level: 
`pow(2, (u_zoom - 14) * a)`

`= pow(2, 19.25619978527 - u_zoom) * pow(2, u_zoom * a - (14 * a))`
`= pow(2, 19.25619978527 - (14 * a) + u_zoom * (a - 1))`

when `a = 0.3`:
`= pow(2.0, 15.06 - u_zoom * 0.7)`

mollymerp: brilliant – thank you for the awesome breakdown @nickidlugash. I implemented what you suggest with a slight variation because there was barely any hillshading showing up at low zoom levels. take a look here https://github.com/mapbox/mapbox-gl-js/pull/5286/files#diff-2869d624151d9150a5f8f68cf567cb6fR61

nickidlugash:
> with respect to renaming hillshade-accent-color 👉 hillshade-slope-color I find this more ambiguous (is it the shaded part of the slope or the lit part of the slope? In this case I think accent is more clear, but I'm open to other ideas.

I don't have strong opinions on this, but in reviewing the shaders, it seemed like the calculation for how this color affects the final display is equivalent to what some other platforms refer to as a "slope map" (the change in color is directly correlated to the steepness of the slope, and no other variable). If you assigned color values to `hillshade-shadow-color` and `hillshade-highlight-color` with zero alpha (e.g. `hsla(0, 0%, 0%, 0)`) the final display will be a slope map. 

My (fairly uneducated) guess is that `hillshade-slope-color` might be more understandable to users familiar with terrain rendering from other software, but I can see how `accent` might be generally less confusing. Perhaps if we stick with `hillshade-accent-color`, we can add to the property's style spec description something like "Based on steepness of the slope."

> although I'm not sure how the default of lower left will look for hillshading, I think it makes most sense to have the lighting position unified across all layers that incorporate lighting into their rendering.

Hmm, yeah – northwest lighting is usually typical for hillshading, but lighting from that direction would make the extrusions backlit. I'm going to dig in a bit more on this issue.

mgvarley:
I note that the scope of this PR says "...right now it is not extensible to other DEM encoding formats" but it would be awesome if this could be extended to support the public terrain datasets that Mapzen have made available on AWS https://aws.amazon.com/public-datasets/terrain/

mollymerp:
@mgvarley thanks for the feedback! we plan on adding extensibility for DEM decoding after this initial PR ships 🚢

kkaefer:
> Perhaps `hillshade-accent-color` would be more usefully named `hillshade-slope-color`? I think the specificity would be helpful to users, both those with and without familiarity of terrain rendering from other software.

The wording of "accent" is from an earlier iteration that included terrain openness into the visualization as well, rather than just the slope. We've cut this feature from the initial release but are hoping to bring it back once this has merged. The questions is whether we should change the behavior of this property to include openness once this lands later on, since it'd alter the appearance of the rendering for the same stylesheet.

nickidlugash:
> The questions is whether we should change the behavior of this property to include openness once this lands later on, since it'd alter the appearance of the rendering for the same stylesheet.

@kkaefer @mollymerp Oh awesome – I didn't realize that you were planning to add back openness calculations. Would it make sense for openness to be a separate property, so that users can independently control the amount of hillshading, slope, and openness in the final display? (Not sure if this avoids the issue of altering the appearance of the rendering for existing stylesheet though, since the default value for a new openness property would presumably create a change in appearance as well).

nickidlugash:
I think there are still a couple of unresolved issues around the global lighting. The global light object was originally designed to be a global light for all extrusions layers. Adapting it to also be used for hillshading layers (which have shading calculated differently) is currently causing a few hiccups:

### 1. The optimal value for azimuthal angle of the light (`position[1]`) is different for hillshading and extrusion style layers.

For hillshading, a NW or NNW angle is generally thought to be ideal (around 315–335degs). [Relief inversion (or terrain reversal)](http://www.tandfonline.com/doi/abs/10.1080/15230406.2016.1185647) is a well-documented effect where bottom-lit hillshading tends to invert the perception of concave and convex landforms. From visual review, @mollymerp and I think this holds true for our implementation.

For extrusions on a pitched map, front-lit shading (around 210deg or 135deg) is generally more useful, since back-lit shading prevents visible faces of the extrusions from receiving light.

Extrusions on an *unpitched* map, however, look generally good at any angle:

|  | azimuth=210 | azimuth=335 |
| --- | --- | ---|
| Hillshading | screen shot 2017-11-09 at 3 54 45 pm |  screen shot 2017-11-09 at 3 54 27 pm |
| Fill-extrusion, pitched | screen shot 2017-11-09 at 3 56 35 pm | screen shot 2017-11-09 at 3 56 58 pm |
| Fill-extrusions, unpitched | extrusion-unpitched-215 | extrusion-unpitched-335 |

I don't think there is much of a use case for combining hillshading and extrusions on a pitched map, so users probably won't run into the issue of needing to assign different light positions in the same style. If a users wants to combine hillshading and extrusions on a map that won't be pitched, assigning a value that is optimal for hillshading should be fine.

**I think the main issue then is that the default value for light position can't be optimal for the most common use cases of both style layer types.** The current default value is 210deg, which was originally optimized for extrusions. Without making breaking changes, that means that the default light postion for terrain will not be optimal.


### 2. Hillshading doesn't use the radial distance (`position[0]`) or light color properties. 

Is it confusing that hillshading is affect by some of the light properties but not others?


### Options

- Use the current global light for both hillshading and extrusions, and keep the current default azimuthal angle (as currently implemented). Document in spec both suggested optimal azimuthal angle for hillshading, and the fact that certain light properties are disabled for hillshading.

- Create a separate global hillshading light that affects all hillshading layers (and the current light can continue to be used for fill-extrusions, and other future extrusion layer types).

- Add light properties within each hillshading layer.

- ???

1ec5:
> I don't think there is much of a use case for combining hillshading and extrusions on a pitched map, so users probably won't run into the issue of needing to assign different light positions in the same style.

Just to state the obvious, some developers may encounter this hillshading effect on an untilted map and expect to see 3D terrain (#1489) when tilting it. So there may be an _initial_ expectation that hillshading and extrusions would coexist on a tilted map, though it won’t last too long. 😉

nickidlugash:
> Just to state the obvious, some developers may encounter this hillshading effect on an untilted map and expect to see 3D terrain (#1489) when tilting it.

Perhaps even more reason to consider decoupling lighting of hillshading from that of extrusions?

src/source/source_cache.js
```diff
@@ -477,17 +518,8 @@ class SourceCache extends Evented {
...
-                        const childCoord = children[j];
-                        const childTile = childCoord ? this.getTile(childCoord) : null;
-                        if (!!childTile && childTile.hasData()) {
-                            retain[childCoord.id] = true;
-                        } else {
-                            covered = false;
-                        }
-                    }
+                    this._findLoadedChildren(coord, maxCoveringZoom, retain);
+                    covered = false;
```
anandthakker: What does this change fix?  Would it make sense to pull this (and the corresponding changes to the source cache unit test) out into a small, separate PR?

test/unit/data/dem_data.test.js
```diff
@@ -0,0 +1,213 @@
...
+                if (level0.get(x, y) === -65536) {
+                    nonempty = false;
+                    break;
+                }
+            }
+        }
+        t.true(nonempty, 'pixel data populates DEM data level');
+
+        let verticalBorderMatch = true;
+        for (let x = -1; x < 5; x += 5) {
```
anandthakker: `for (const x of [-1, 4])` would be a little clearer

test/unit/data/dem_data.test.js
```diff
@@ -0,0 +1,213 @@
...
+
+        t.end();
+    });
+
+    t.test('backfillBorder correctly populates borders with neighboring data', (t)=>{
+        const level0 = dem0.level;
+
+        dem0.backfillBorder(dem1, -1, 0);
+        for (let y = 0; y < 4; y++) {
+            t.true(level0.get(-1, y) !== level0.get(0, y), 'backfills neighbor -1, 0');
```
anandthakker: This looks like it's checking that the border no longer matches the closest pixel from the content, but it doesn't check whether the new backfilled data has correct values.  Is there a good way to check the latter?

mollymerp: 😛 good point! yes I think it would be p simple to add this.

src/source/source_cache.js
mollymerp: Ah yes – I forgot to split this out. thanks for the catch! This fixes an issue where the deleted code only checks one ZL below the current ZL so any children that are more than 1 ZL greater get released, so rapid zooming out causes a blink where child tiles are deleted before the ideal parent tiles are loaded

mollymerp: 👉  #5671

mollymerp:
Thanks for the additional review @anandthakker and the detailed feedback @nickidlugash!

> I think the main issue then is that the default value for light position can't be optimal for the most common use cases of both style layer types.

I agree but I'm not 100% sold on how to get around this. I'm torn between:
> Use the current global light for both hillshading and extrusions, and keep the current default azimuthal angle (as currently implemented). Document in spec both suggested optimal azimuthal angle for hillshading, and the fact that certain light properties are disabled for hillshading.

and
> Add light properties within each hillshading layer.

reverting back to `hillshade-illumination-direction` and `hillshade-illumination-intensity` ? The reasoning for the latter could be that the global `light` object is reserved for true 3d scenes.

what are your preferences @nickidlugash @kkaefer ?

I think the remaining blocker on this is the muddy-looking intermediary colors @nickidlugash pointed out in https://github.com/mapbox/mapbox-gl-js/pull/5286#issuecomment-339519505 
I don't have any great ideas on how to mitigate this but I will plan on playing around w the shaders this week.

mollymerp:
I tried out LAB interpolation in d1b5833 and removed dependence on the global light object in 4a84194. Will sync up with @nickidlugash about the new color space.

nickidlugash:
Synced up with @mollymerp yesterday about LAB interpolation and generally on remaining color considerations:

### LAB vs RGB interpolation
I compared the output of the two, with @ajashton as additional 👀 . While the LAB interpolation is definitely less muddy and overall nicer-looking in some cases, the range of additional hues it introduces might be unexpected and undesirable, especially since the rest of the color interpolation on the map is RGB by default. E.g., here's what blue/yellow hillshading looks like:

RGB:
blue-yellow-rgb

LAB:
blue-yellow-lab

My guess is that users would prefer the tighter control over hues offered by RGB. However, I could also see situations where LAB would be desirable, especially since we don't allow users to set multiple color stops. 

Perhaps we can add a `hillshade-colorspace` property, with an RGB default?

### White/light highlights not prominent enough over light backgrounds

I previously flagged this as a potential issue with the color output of this shader. While I do still have some concerns about this, for subtle hillshading styles we can get around the issue by increasing the `hillshade-exaggeration` and decreasing the alpha values for the `hillshade-shadow-color` and `hillshade-accent-color` so that the highlight color is more prominent:

screen shot 2017-11-21 at 7 17 38 pm

---

I think the implementation as it stands would be fine for what we might need for future iterations of our core styles (at least from a visual perspective – there are other considerations that still need to be evaluated). The output for more high-contrast or stylistic color choices could maybe still be improved, but perhaps it makes sense to release this new feature and get more user feedback on it first.

src/render/render_texture.js
```diff
@@ -22,7 +22,7 @@ class RenderTexture {
         gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
         gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
         gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
-        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, painter.width, painter.height, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, width || painter.width,  height || painter.height, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
```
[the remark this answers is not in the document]
mollymerp: no, we guard against this in the Level constructor

mollymerp:
@nickidlugash thanks for the feedback. per chat with konstantin we're going to proceed with rgb interpolation in the initial implementation and potentially add in other color spaces at a later time. I think I've addressed all your other visual concerns. thanks again for your help!

@kkaefer @anandthakker @nickidlugash this is ready for final review 🎈 i'm shooting to get this merged into master by EOW. 

benchmarks – https://bl.ocks.org/anonymous/raw/620431bd26bc2d35126364ee2bf19c77/
image

lbud:
image

@mollymerp I just merged https://github.com/mapbox/mapbox-gl-js/pull/5739, which has implications for this PR — see e.g. https://github.com/mapbox/mapbox-gl-js/pull/5739/files#diff-efa3fb7fa712ba847de8d82efdd2964d for the kinds of changes you'll need to make here 😬

flow-typed/style-spec.js
```diff
@@ -294,7 +294,7 @@ declare type HeatmapLayerSpecification = {|
         "heatmap-radius"?: PropertyValueSpecification<number>,
         "heatmap-weight"?: DataDrivenPropertyValueSpecification<number>,
         "heatmap-intensity"?: PropertyValueSpecification<number>,
-        "heatmap-color"?: PropertyValueSpecification<ColorSpecification>,
+        "heatmap-color"?: ColorSpecification,
```
kkaefer: Merge error?

src/shaders/hillshade.fragment.glsl
```diff
@@ -0,0 +1,51 @@
+
+#ifdef GL_ES
+precision highp float;
+#else
+#define lowp
+#define mediump
+#define highp
+#endif
```
kkaefer: We're defining this in `_prelude.{fragment,vertex}.glsl`, which is automatically prepended to all shaders in both GLJS and GL Native, so there's no need to include this definition here.

src/source/raster_dem_tile_source.js
```diff
@@ -0,0 +1,128 @@
...
+import type {Callback} from '../types/callback';
+
+
+class RasterDEMTileSource extends RasterTileSource implements Source {
+    _options: RasterSourceSpecification;
+
+    constructor(id: string, options: RasterSourceSpecification, dispatcher: Dispatcher, eventedParent: Evented) {
+        super(id, options, dispatcher, eventedParent);
+        this.type = 'raster-dem';
+        this.maxzoom = 15;
```
kkaefer: Why is the maxzoom hard coded?

src/style-spec/reference/v8.json
```diff
@@ -448,6 +451,14 @@
             }
           }
         },
+        "hillshade": {
+          "doc": "Client-side hillshading visualization based on DEM data. Currently, the implementation only supports Mapbox Terrain RGB tiles",
+          "sdk-support": {
+            "basic functionality": {
+              "js": "next"
```
kkaefer: update before merging

src/style-spec/reference/v8.json
```diff
@@ -1785,6 +1797,30 @@
...
+        }
+      },
+      "default": "visible",
+      "doc": "Whether this layer is displayed.",
+      "sdk-support": {
+        "basic functionality": {
+          "js": "0.10.0",
+          "android": "2.0.1",
+          "ios": "2.0.0",
+          "macos": "0.1.0"
```
kkaefer: this is incorrect ;)

flow-typed/style-spec.js
mollymerp: addressed with #5789

src/shaders/hillshade.fragment.glsl
mollymerp: I have to keep highp precision for hillshade_prepare.fragment for it to work on some android devices.

src/source/raster_dem_tile_source.js
mollymerp: terrain-rgb tiles stop at z14, but I changed it to 22 for consistency with other source types.

mollymerp:
thanks for sticking with me y'all ⛰ 🗻 ⛰ 
thanks @anandthakker and @kkaefer for continued reviews and feedback!

note that there may be some caching on the api side that causes the endpoint to still return webp files so you may need to use the following transform function for the next few hours

``` js 
transformRequest: (url, resourceType) =>{
    if (resourceType === 'Tile' && url.search('terrain')) {
        return {
            url: url.replace('@2x', '').replace(/(\.webp)|(\.png)/, '.pngraw')
        };
    }
}
```

stevewillard:
I have been following this PR for awhile. This is some really great work @mollymerp Nice job! 🎉 

I look forward to trying this layer type out

PR #5293, opened 2017-09-14
Refactor checkSubtype

jfirebaugh:
Some simplifications I noticed while doing another review pass.


src/style-spec/function/types.js
```diff
@@ -54,6 +54,44 @@ function toString(type: Type): string {
...
+            return null;
+        }
+    } else if (expected.kind === t.kind) {
+        return null;
+    } else if (expected.kind === 'Value') {
+        for (const memberType of valueMemberTypes) {
+            if (!checkSubtype(memberType, t)) {
+                return null;
+            }
+        }
```
anandthakker: Hm. I'm realizing now: according to this logic (together with array subtyping), `Value` is effectively the `top` type... should `Color` really be a member type of `Value`?

jfirebaugh: Let's ticket this separately.

anandthakker: Ticketed #5296

PR #5298, opened 2017-09-14
symbol sorting for viewport labelling

ansis:
This sorts symbols that may overlap based on their vertical position. The sorting happens at the end of symbol placement within `commitPlacement`.

The sorting happens in each tile independently. This means there might be incorrect sorting near tile edges but I think that fixing that minor drawback would be not worth the added complexity. Unless we switch back to the clipping approach we used before.

@ChrisLoer


src/gl/index_buffer.js
```diff
@@ -21,14 +22,22 @@ class IndexBuffer {
...
+            delete array.arrayBuffer;
+        }
     }
 
     bind() {
         this.gl.bindBuffer(this.gl.ELEMENT_ARRAY_BUFFER, this.buffer);
     }
 
+    updateData(array: TriangleIndexArray | LineIndexArray) {
+        this.bind();
```
ChrisLoer: Does it make sense to assert here that it's in `dynamicDraw` mode?

src/data/bucket/symbol_bucket.js
```diff
@@ -752,6 +761,78 @@ class SymbolBucket implements Bucket {
...
+        // additional segments. In this case we need to make sure that vertices
+        // are referenced from their matching index array segment. This is done by
+        // adding all the indexes from the first segment, then the second, etc.
+        const numTextSegments = this.text.segments.get().length;
+        for (let segmentIndex = 0; segmentIndex < numTextSegments; segmentIndex++) {
+            for (const i of symbolInstanceIndexes) {
+                const symbolInstance = this.symbolInstances[i];
+
+                for (const placedTextSymbolIndex of symbolInstance.placedTextSymbolIndices) {
+                    const placedSymbol = this.placedGlyphArray.get(placedTextSymbolIndex);
```
ChrisLoer: Am I right in reading this as "we sort symbols within a segment, but symbols can't reorder between segments"? And if that's the case, aren't we basically giving up on sorting when there are multiple segments? That seems OK since this is an edge case, but if we're going to give up, why not just early-exit the entire sorting process?

ansis: Yep, it's kind of giving up. I hadn't thought about an early-exit but that's probably cleaner. This approach sorts them so that they look mostly right, but is more complicated and probably error-prone. I'll switch to an early exit!

src/gl/index_buffer.js
ansis: yep!

ChrisLoer:
Oh also as part of these changes we should un-ignore the tests tagged 5150 in `ignores.json` and update the expected images as appropriate.

ChrisLoer:
Err, another concern: doesn't this break the debug collision boxes/circles for sorted symbols? Their index arrays will need the same re-sorting...

ansis:
> Err, another concern: doesn't this break the debug collision boxes/circles for sorted symbols? Their index arrays will need the same re-sorting...

Nope, this leaves collision box rendering unchanged. Their boxes are rendered un an unsorted order, but that seems ok to consider it's just a debug view and these symbols are all allowed to overlap.

ChrisLoer:
> Nope, this leaves collision box rendering unchanged.

🤦‍♂️  I didn't think that through all the way -- forgot you're not changing the sort of the `symbolInstances` themselves.

ChrisLoer:
Looks great to me, let's merge it into `viewport-collision`.

We can do it on either branch, but we should be able to remove these three ignores now:

    "render-tests/symbol-spacing/line-close": "WIP Ignore for PR #5150 (no more tile clipping artifacts)",
    "render-tests/symbol-spacing/line-far": "WIP Ignore for PR #5150 (no more tile clipping artifacts)",
    "render-tests/text-tile-edge-clipping/default": "WIP Ignore for PR #5150 (no more tile clipping artifacts)"

We'll have to update the expected results, but the new results should be unambiguously better. 😃

PR #5302, opened 2017-09-14
Experiment: flatten heavily-accessed StructArrays

ChrisLoer:
This is a proof-of-concept of a potential optimization for collision detection and line-label rendering. The idea is to allow `StructArray`s to convert themselves into flat arrays of javascript numbers after they've been deserialized. For performance hotspots that do a lot of array lookups, indexing straight into flat arrays of numbers seems to give noticeable performance wins.

I added a "flatten" method to `StructArray` that converts the struct array into one flat array of plain javascript numbers, and generates getters that are just a thin wrapper on indexing into the flat array (e.g. `getThirdMember(index) { return this.flattenedArray[index + 2]; }`. I switched to using the flattened version for `glyphOffsetArray` and `lineVertexArray`, the two arrays that are most heavily accessed in `updateLineLabels`.

In three runs of a 180 degree rotation in a pitched/densely labeled map, I saw average time spent in `updateLineLabels` go from 28.22% down to 25.95% (p = .014). So, it looks like there's a measurable win here, although it's not that dramatic.

@mourner and @ansis, any thoughts on whether this approach is worth pursuing?


ansis:
I wonder how getting members directly from the underlying array (`getThirdMember(index) { return this.uint16[index * size + offset]; }`) would compare. It would avoid having that big array of arrays. 

Another thing to check would be to just change `struct.thirdMember` to `struct.getThirdMember()`. Maybe the special getters are slower than regular functions.

It might also be worth checking performance of having `StructArray.get(i)` return a regular object instead of this lazy stuff.

If we're storing the stuff in regular js objects why not just get rid of the StructArrays here and just transfer json? I think StructArrays slightly optimize transfer time and memory usage at the cost of slightly more expensive access.

ChrisLoer:
> Instead of an explicit flatten stage, would it make sense to consider refactoring all structArrays to be flat?

The current set-up of `StructArray` seems like it's pretty well optimized for _most_ of the cases where it's used (i.e. can be bound directly to a WebGL buffer, can be transferred quickly, uses less memory for low precision types). It's only the symbol code that does lots of heavy accessing of the arrays from the CPU.

> Also, it looks like this could be refactored to use instance attributes instead of eval-type code generation — would it have similar performance this way?

I may be misunderstanding what you're asking... The code generation only happens at the time the type is generated, so I wouldn't expect that to be too expensive. By refactoring to "instance attributes", do you mean make a "get" that returns an object that knows its index and then uses the index to look up its member properties in the underlying array? That would be a nicer interface, but I assumed it would be harder for the JS engine to convert into plain array access. I'll try to experiment with that.

> I wonder how getting members directly from the underlying array (getThirdMember(index) { return this.uint16[index * size + offset]; }) would compare. It would avoid having that big array of arrays.

That's a good idea, I'll try that too.

> If we're storing the stuff in regular js objects why not just get rid of the StructArrays here and just transfer json? I think StructArrays slightly optimize transfer time and memory usage at the cost of slightly more expensive access.

Yeah, this is a good question. Do we have a technique for evaluating transfer time differences?

ChrisLoer:
> I wonder how getting members directly from the underlying array (getThirdMember(index) { return this.uint16[index * size + offset]; }) would compare. It would avoid having that big array of arrays.

This turned out to work just as well as flattening the array, and has the advantage of no extra memory usage or time spent flattening.

I did 4 runs of `bench/#frame-duration` and collected the delta from master to these versions.

| | Direct Access | Current `viewport-collision`
--- | --- | ---
| |.4ms | .9ms
| | .4ms | .7ms
| | .4ms | .5ms
| | .4ms | .6ms
| Mean | .4ms | .675ms |

p = .018

I like that this requires almost no change to any of the other uses of `StructArray` -- it just adds a slightly more cumbersome but faster accessor.

> It might also be worth checking performance of having StructArray.get(i) return a regular object instead of this lazy stuff.

The view `get(i)` returns is used by `populatePaintArrays` for writing into the underlying buffer, so if we did something like that we'd need an alternate way to do the setting.

ChrisLoer:
I rebased the `viewport-collision` branch and cherry-picked a version of this onto it.

Issue #5301, opened 2017-09-14
WebGL CanvasSource performance drop in 0.40.0


gpbmike:
**mapbox-gl-js version**: 0.40.0

### Steps to Trigger Behavior

Add a webgl canvas source/layer.

### Expected Behavior

Should be a buttery 60fps like 0.39.1.

### Actual Behavior

Severe drop in framerate. In the example codepens below I'm using [plotty](https://github.com/santilland/plotty/blob/master/src/plotty.js) to generate a colorful webgl canvas. The only difference between the codepens is the `mapbox-gl-js` version.

The performance is heavily dependent on the size of the canvas being used. In this case it's 2000x2000, which is not uncommon for me.

It's probably the overhead of creating a large image on every render. 😢 
https://github.com/mapbox/mapbox-gl-js/pull/5155

#### 0.39.1
https://codepen.io/gpbmike/pen/oGXmRJ

#### 0.40.0
https://codepen.io/gpbmike/pen/jGPvpo

gpbmike:
I can't seem to create a reduced test case for this so I'm going to close it and do more digging.

gpbmike:
I have updated the description with a reduced test case.

lbud:
🙁 right, likely a side effect of having to copy all pixels into an intermediary buffer #5155 — we can mitigate this for non-animated canvases actually (added in #5303), but unfortunately since #5155 is a workaround for a hardware bug I don't have ideas for making animated, WebGL-based canvas sources much faster easily. (Eventually down the road when Mapbox GL adds support for ["custom layers"](https://github.com/mapbox/mapbox-gl-js/issues/281) — that is, layers that define their own layout and shaders, using our existing pipeline — that'll be the way forward for fast custom WebGL overlays.)

gpbmike:
I've been doing some additional tinkering in `mapbox-gl-js` with my broadwell processor machine (https://github.com/mapbox/mapbox-gl-js/issues/4262#issuecomment-320343557).

- I **reconfirmed the problem with `0.39.1`** when adding a webgl canvas source. 
- I **confirmed that the problem is fixed in `0.40.1`** when adding a webgl canvas source.
- I pointed to a local copy of `0.40.1` where [`readCanvas`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/canvas_source.js#L118-L140) simply sets `this.canvasData = this.canvas` and it still works. 🤔 

[...]

PR #5303, opened 2017-09-14
Fix for webgl-based canvas sources rendering flipped in 0.40.0 + non-animated optimization

### LINKED ISSUE #5300: CanvasSource coordinates inverted in 0.40.0
**mapbox-gl-js version**: 0.40.0

### Steps to Trigger Behavior

 1. Add canvas layer to map with coordinates order top left, top right, bottom right, bottom left per [spec](https://www.mapbox.com/mapbox-gl-js/style-spec/#sources-canvas).

### Expected Behavior

Canvas layer should render in desired location.

### Actual Behavior

Canvas layer is rotated and flipped. If you use the following order the canvas layer will render correctly: bottom left, bottom right, top right, top left.

Here's a giphy that compares v0.37.0 (first, smooth fps) to v0.40.0 (second, choppy fps). Notice the colored layer is positioned differently in the second example.

Framerate issues addressed here: https://github.com/mapbox/mapbox-gl-js/issues/5301

lbud:
WebGL APIs and Canvas APIs are vertically flipped (gl reads from the lower left, canvas from the upper left), so I introduced https://github.com/mapbox/mapbox-gl-js/issues/5300 in #5155. Eventually we should try to write render tests with a secondary webgl canvas 🤔 to catch bugs like this, but in the meantime this should make it into the 0.40.1 patch release.

Edit: this PR also modifies the intermediary read step to only reread if a canvas is supposed to be animated, if it hasn't been read yet, or if it is resized, to mitigate performance concerns for non-animated canvases (see https://github.com/mapbox/mapbox-gl-js/issues/5301).

Fixes #5300.
Refs #5301.

## Launch Checklist

### DIFF
--- a/src/source/canvas_source.js
+++ b/src/source/canvas_source.js
@@ -129,7 +129,7 @@ class CanvasSource extends ImageSource {
             const imageData = this.secondaryContext.createImageData(this.width, this.height);
             const flipped = new Uint8Array(this.width * this.height * 4);
             for (let i = this.height - 1, j = 0; i >= 0; i--, j++) {
-                flipped.set(data.slice(i * this.width * 4, (i + 1) * this.width * 4), j * this.width * 4);
+                flipped.set(data.subarray(i * this.width * 4, (i + 1) * this.width * 4), j * this.width * 4);
             }
             imageData.data.set(flipped);
             return imageData;
--- a/src/source/canvas_source.js
+++ b/src/source/canvas_source.js
@@ -45,6 +45,7 @@ class CanvasSource extends ImageSource {
     secondaryContext: ?CanvasRenderingContext2D;
     width: number;
     height: number;
+    canvasData: ?ImageData;
     play: () => void;
     pause: () => void;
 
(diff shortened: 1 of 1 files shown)


src/source/canvas_source.js
```diff
@@ -127,7 +127,11 @@ class CanvasSource extends ImageSource {
             gl.readPixels(0, 0, this.width, this.height, gl.RGBA, gl.UNSIGNED_BYTE, data);
             if (!this.secondaryContext) this.secondaryContext = window.document.createElement('canvas').getContext('2d');
             const imageData = this.secondaryContext.createImageData(this.width, this.height);
-            imageData.data.set(data);
+            const flipped = new Uint8Array(this.width * this.height * 4);
+            for (let i = this.height - 1, j = 0; i >= 0; i--, j++) {
+                flipped.set(data.slice(i * this.width * 4, (i + 1) * this.width * 4), j * this.width * 4);
```
mourner: I'd suggest avoiding many `slice` operations since they create a new array every time. 4 lines setting each of the 4 values of the array will be faster.

ansis: The array looks like its an entire row, not just four values. It might make sense to try `data.subarray(begin, end)` which is like `.slice()` except it provides a view into the existing array without copying it.

mourner: Oh, you're right. In that case a secondary inner loop might get also work well.

lbud: Oh cool, did not know about `subarray` — thanks @ansis @mourner. (I tried the secondary loop in this test as well for good measure: )
image

src/source/canvas_source.js
```diff
@@ -114,21 +115,29 @@ class CanvasSource extends ImageSource {
...
+            if (!this.canvasData || resize) {
+                this.canvasData = this.secondaryContext.createImageData(this.width, this.height);
+            }
+
+            // WebGL reads pixels bottom to top, but for our ImageData object we need top to bottom: flip here
+            const flipped = new Uint8Array(this.width * this.height * 4);
+            for (let i = this.height - 1, j = 0; i >= 0; i--, j++) {
+                flipped.set(data.subarray(i * this.width * 4, (i + 1) * this.width * 4), j * this.width * 4);
+            }
+            this.canvasData.data.set(flipped);
```
mourner: One more thing — is it possible to do `this.canvasData.data.set` inside the loop instead of using a temporary `flipped` array?

PR #5313, opened 2017-09-16
serialize use url not urls

wandergis:
`serialize` use `url` not `urls`


jfirebaugh:
Thanks @wandergis! Can you update [the corresponding test](https://github.com/mapbox/mapbox-gl-js/blob/1cb1cffbdc80020e80cf7db91b105943f7646c75/test/unit/source/image_source.test.js#L86-L95) as well so that CircleCI passes?

PR #5314, opened 2017-09-16
use translate3d replace translate

wandergis:
For my app,when map container is scaled, box will dispear when it is out of the origin bounds,this is caused by map container's css `overflow:hidden`,so i replace `translate` to `translate3d`.


jfirebaugh:
I'm not sure that `translate3d` should be required here. Can you create a standalone example that demonstrates the original issue you had?

PR #5330, opened 2017-09-19
Fix for infinite flight path length

pkang:
The total flight path length can sometimes be infinite. Using `!isFinite(S)` handles `Infinity` and `NaN` values for `S`.


pkang:
Looks like #5331 would fix this issue. Closing.

Issue #4914, opened 2017-06-28
setStyle should use setData in diff when it can


brandonbloom:
I'm making use of the new-ish "smart" `setStyle` in a React application. Modulo some small issues (see #4912) , it works quite well!

One thing I noticed was that geojson layers are compared in a way that includes their data property. If the data changes, a `removeSource`/`addSource` pair is executed, which results in reloading data and a visible flicker. A better thing to do would be to emit a `setData` operation to ensure smooth animation.

anandthakker:
👍 

related (but not the same): https://github.com/mapbox/mapbox-gl-js/issues/3692

anandthakker:
Done in #5332

brandonbloom:
Thanks!

PR #5332, opened 2017-09-20
Use setData operation when diffing geojson sources

zjernst:
https://github.com/mapbox/mapbox-gl-js/issues/4914

When diffing the styles, if a geojson source is found with different data, perform a setData operation. Previously, the diff would cause a ```removeSource```/```addSource``` pair when a change was found, causing the source to flicker.

I'm migrating my project from performing all diffs myself to keeping a style state tree and utilizing the smarter ```setStyle``` that performs the necessary diffs, however the flickering was a noticeable downgrade to the experience.

First contribution, so feel free to tear it apart/offer a better implementation!

### DIFF
--- a/src/source/source.js
+++ b/src/source/source.js
@@ -57,6 +57,7 @@ export interface Source {
 
     +onAdd?: (map: Map) => void;
     +onRemove?: (map: Map) => void;
+    +setData?: (data: GeoJSON | string) => Source;
 
     loadTile(tile: Tile, callback: Callback<void>): void;
     +hasTile?: (coord: TileCoord) => boolean;
--- a/src/source/source.js
+++ b/src/source/source.js
@@ -57,7 +57,6 @@ export interface Source {
 
     +onAdd?: (map: Map) => void;
     +onRemove?: (map: Map) => void;
-    +setData?: (data: GeoJSON | string) => Source;
 
     loadTile(tile: Tile, callback: Callback<void>): void;
     +hasTile?: (coord: TileCoord) => boolean;
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -102,6 +102,12 @@ class SourceCache extends Evented {
         }
     }
 
+    setData(data: GeoJSON | string) {
+        if (this._source && this._source.setData) {
+            this._source.setData(data);
+        }
+    }
+
     /**
      * Return true if no tile data is pending, tiles will not change unless
      * an additional API call is received.
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -102,12 +102,6 @@ class SourceCache extends Evented {
         }
     }
 
-    setData(data: GeoJSON | string) {
-        if (this._source && this._source.setData) {
-            this._source.setData(data);
-        }
-    }
-
     /**
      * Return true if no tile data is pending, tiles will not change unless
      * an additional API call is received.
--- a/src/style-spec/diff.js
+++ b/src/style-spec/diff.js
@@ -44,7 +44,7 @@ const operations = {
     removeSource: 'removeSource',
 
     /*
-     * { command: 'setData', args: ['sourceId'. data] }
+     * { command: 'setData', args: ['sourceId', data] }
      */
     setData: 'setData',
 
--- a/src/style-spec/diff.js
+++ b/src/style-spec/diff.js
@@ -44,6 +44,11 @@ const operations = {
     removeSource: 'removeSource',
 
     /*
+     * { command: 'setData', args: ['sourceId'. data] }
+     */
+    setData: 'setData',
+
+    /*
      * { command: 'setLayerZoomRange', args: ['layerId', 0, 22] }
      */
     setLayerZoomRange: 'setLayerZoomRange',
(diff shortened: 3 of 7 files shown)


src/style-spec/diff.js
```diff
@@ -44,6 +44,11 @@ const operations = {
     removeSource: 'removeSource',
 
     /*
+     * { command: 'setData', args: ['sourceId'. data] }
+     */
+    setData: 'setData',
```
anandthakker: Let's call this `setGeoJSONSourceData`

src/source/source_cache.js
```diff
@@ -102,6 +102,12 @@ class SourceCache extends Evented {
         }
     }
 
+    setData(data: GeoJSON | string) {
+        if (this._source && this._source.setData) {
+            this._source.setData(data);
+        }
+    }
```
anandthakker: Instead of adding this geojson-source-specific method to `SourceCache`, we can use `sourceCache.getSource().setData(...)` in the `Style` method handling this diff operation.

src/style/style.js
```diff
@@ -512,9 +512,10 @@ class Style extends Evented {
         this._checkLoaded();
 
         if (this.sourceCaches[id] === undefined) {
-            throw new Error('There no source with this ID');
+            throw new Error('There is no source with this ID');
         }
```
anandthakker: Since this is an internal invariant, please replace with `assert(this.sourceCaches[id] === undefined, 'There is no source with this ID')`

test/unit/style/style.test.js
```diff
@@ -644,6 +644,64 @@ test('Style#removeSource', (t) => {
...
+                {
+                    "type": "Feature",
+                    "geometry": { "type": "Point", "coordinates": [125.6, 10.1] }
+                }
+            ]
+        };
+        style.on('style.load', () => {
+            const sourceCache = style.sourceCaches['source-id'];
+            t.spy(sourceCache, 'setData');
+            style.setData('source-id', geoJSONSourceData);
```
anandthakker: Instead of testing `style.setData()`, could you please rework this test to use `style.setState()`?   The latter is part of the interface of `Style` from the perspective of the rest of the codebase, whereas the former is essentially private.  (Admittedly, this isn't very clearly documented or enforced at the moment.)

src/style/style.js
```diff
@@ -501,6 +502,27 @@ class Style extends Evented {
...
+    * @param {GeoJSON|string} data GeoJSON source
+    * @throws {Error} if no source is found with the given ID
+    */
+
+    setData(id: string, data: GeoJSON | string) {
+        this._checkLoaded();
+
+        if (this.sourceCaches[id] === undefined) {
+            throw new Error('There is no source with this ID');
+        }
```
anandthakker: Since this is an internal invariant, please replace this check with just `assert(this.sourceCaches[id] === undefined, 'There is no source with this ID')`

src/style-spec/diff.js
```diff
@@ -44,6 +44,11 @@ const operations = {
     removeSource: 'removeSource',
 
     /*
+     * { command: 'setData', args: ['sourceId', data] }
+     */
+    setData: 'setData',
```
anandthakker: Let's call this `setGeoJSONSourceData`

zjernst:
@anandthakker Thanks for the review! Made changes to reflect your comments. Let me know how it looks

src/source/source.js
```diff
@@ -57,6 +57,7 @@ export interface Source {
 
     +onAdd?: (map: Map) => void;
     +onRemove?: (map: Map) => void;
+    +setData?: (data: GeoJSON | string) => Source;
```
anandthakker: Leftover -- this can be removed

anandthakker:
@zjernst changes look good, thanks!  One leftover: https://github.com/mapbox/mapbox-gl-js/pull/5332#pullrequestreview-64593158

@jfirebaugh @mourner FYI, this change makes `setStyle()` smarter by using `GeoJSONSource#setData()` when possible, rather than always removing/re-adding the source.  It _doesn't_ currently defer the `setData()` work to the once-per-frame batch update (`Style#update()`), but I think we should go ahead with it and leave that for #3692

src/source/source.js
zjernst: Correct me if I'm wrong, but I believe I still need to add this to the ```Source``` interface so ```setData``` can be called in ```Style#setGeoJSONSourceData```. I'm getting a flow error of ```property setData of unknown type``` when I remove this line.

anandthakker: Ah, I see.  `setData` is defined on [GeoJSONSource](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/geojson_source.js).  So, instead of adding it to the general `Source` interface, you can do the following in `setGeoJSONSourceData`:

```js
const geojsonSource: GeoJSONSource = (this.sourceCaches[id].getSource(): any);
geojsonSource.setData()
```

Note that the use of `: any` essentially disables Flow's checking, so it should be used with caution.  As a general practice, I try to include some sort of runtime check to accompany any use of `: any` -- e.g., `assert(geojsonSource.type === 'geojson')`.

zjernst: :+1: Gotcha -- that makes sense

zjernst:
@anandthakker Thanks for you help, think it should be good now.

anandthakker:
Thanks again @zjernst 🎉 !

Issue #5333, opened 2017-09-20
Map flickers/bounces when animating in unfocused iframe on Chrome 61


lamuertepeluda:
Hi there,

When zooming in/out (wheel, or animations like flyTo) the map flickers and "bounces" in and out.

### Steps to Trigger Behavior

the bug is visible with the [main example](https://www.mapbox.com/mapbox-gl-js/api/) (or also other examples, e.g. [geocode](https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/)) since v0.40.1. Using v0.39.1 I don't have this behavior.

My System:
iMac 16,2 (late 2015)
Browser: Chrome 61.0.3163.91 64bit

The problem is less evident when using Safari (10.1.2 ) and Firefox (55.0.3), although minimal flickering still occurs.

Please look at the [screen capture at this link](https://drive.google.com/open?id=0B8iH6OUV3dt6cExDS251SkpWbGM) (I tried to make a gif but it cuts some frames out and renders poorly).

### Expected Behavior

Smooth animations like in v0.39.1.

### Actual Behavior

The animation is jerky, the map flickers and "bounces" in and out while zooming.

ChrisLoer:
I can reproduce this using the example page, but not in a map outside of an `iframe`. I see the same behavior between v0.39.1 and v0.40.1 -- for me the change that made the problem appear was to upgrade Chrome from v60 to v61. I can't reproduce on Safari or Firefox (although there are some other potential causes of flickering you might be noticing?).

As in #5335, the problem goes away when I pan the map (but returns every time I switch tabs or switch apps and then come back to the problematic map). No clue yet what's causing this, but we're investigating.

ChrisLoer:
@lamuertepeluda Have you seen this behavior anywhere outside of our examples?

lamuertepeluda:
@ChrisLoer it's even stranger! It happens only on the example website page! 😮 

After you pointed me to the possible `iframe` issue, I tried the following:  
* upgrade mapbox from 0.39.1 to 0.40.1 on some sample code of mine without `iframe`: **works without flickering** ✅ 
* upgrade mapbox from 0.39.1 to 0.40.1 on some sample code of mine inside an `iframe`: **works without flickering** ✅ 
* example code served locally outside an `iframe`: **works without flickering** ✅ 
* example code served locally *inside* an `iframe`: **works without flickering** ✅ 
[...]

PR #5343, opened 2017-09-21
Workaround for example flickering documented in issue #5333.

ChrisLoer:
Removes `overflow: hidden` from div containing the example iframe.
I don't have a clear theory for why this works, but my hypothesis is that it's some interaction with Chrome's throttling of animation for frames that it thinks aren't visible.

Running the docs locally, I no longer see the flicker documented in #5333.

@mollymerp I can't see any other change in the example page after removing the clipping for the example div, but I don't know why it was put there in the first place. Do you have any idea who might know?

### DIFF
--- a/docs/_layouts/example.html
+++ b/docs/_layouts/example.html
@@ -2,7 +2,7 @@
 layout: pages
 ---
 
-<div class='round doc clip fill-white keyline-all'>
+<div class='round doc fill-white keyline-all'>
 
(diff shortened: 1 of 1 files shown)


ChrisLoer:
:tada: Looks fixed on mapbox.com!

ryanbaumann:
That made a huge difference - thanks, @chrisloer!

PR #5348, opened 2017-09-25
Make compass and zoom controls optional

### LINKED ISSUE #1554: Decouple zoom and compass UI controls
Adding 'navigation' gives zoom in and out and compass; would it be possible to have just the compass (thus saving 'mobile screen estate')?

matijs:
Add options to `NavigationControl`. `showZoom` and `showCompass` both default to `true` can be used to hide either the zoom buttons or the compass button from the Navigation Control.

- [x] conditionally add dragRotateHandler

Fixes #1554

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -7,7 +7,8 @@ const DragRotateHandler = require('../handler/drag_rotate');
 import type Map from '../map';
 
 const defaultOptions = {
-    showCompass: true
+    showCompass: true,
+    showZoom: true
 };
 
 /**
@@ -39,9 +40,10 @@ class NavigationControl {
         this._container = DOM.create('div', 'mapboxgl-ctrl mapboxgl-ctrl-group');
         this._container.addEventListener('contextmenu', (e) => e.preventDefault());
 
-        this._zoomInButton = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-zoom-in', 'Zoom In', () => this._map.zoomIn());
-        this._zoomOutButton = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-zoom-out', 'Zoom Out', () => this._map.zoomOut());
-
+        if (this.options.showZoom) {
+            this._zoomInButton = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-zoom-in', 'Zoom In', () => this._map.zoomIn());
+            this._zoomOutButton = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-zoom-out', 'Zoom Out', () => this._map.zoomOut());
+        }
         if (this.options.showCompass) {
             this._compass = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-compass', 'Reset North', () => this._map.resetNorth());
             this._compassArrow = DOM.create('span', 'mapboxgl-ctrl-compass-arrow', this._compass);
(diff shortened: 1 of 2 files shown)


src/ui/control/navigation_control.js
```diff
@@ -15,6 +15,9 @@ const defaultOptions = {
  * A `NavigationControl` control contains zoom buttons and a compass.
  *
  * @implements {IControl}
+ * @param {Object} [options]
+ * @param {Object} [options.showCompass=true] If `false` will omit the compass button from the Navigation Control. The default is `true` to show a compass button.
```
anandthakker: Small edit:
> @param {boolean} [options.showCompass=true] If `true`, the compass button is included.
> @param {boolean} [options.showZoom=true] If `true`, the zoom-in and zoom-out buttons are included.

Note the type should be marked `{boolean}`, not `{Object}`.

  The default value is shown in the generated docs, so no need to state it.

src/ui/control/navigation_control.js
```diff
@@ -35,13 +41,18 @@ class NavigationControl {
 
         this._zoomInButton = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-zoom-in', 'Zoom In', () => this._map.zoomIn());
         this._zoomOutButton = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-zoom-out', 'Zoom Out', () => this._map.zoomOut());
-        this._compass = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-compass', 'Reset North', () => this._map.resetNorth());
-        this._compassArrow = DOM.create('span', 'mapboxgl-ctrl-compass-arrow', this._compass);
+
+        if (this.options.showCompass) {
+            this._compass = this._createButton('mapboxgl-ctrl-icon mapboxgl-ctrl-compass', 'Reset North', () => this._map.resetNorth());
+            this._compassArrow = DOM.create('span', 'mapboxgl-ctrl-compass-arrow', this._compass);
+        }
```
anandthakker: If the `this._compass` element is not created for `showCompass: false`, then we should also make sure in `onAdd()` not to create the `DragRotateHandler`, whose constructor accepts that element as an argument.

matijs: I thought the same, but if no element it passed in the `DragRotateHandler` will default to the `canvasContainer` :)

src/ui/control/navigation_control.js
matijs: Fixed.

src/ui/control/navigation_control.js
anandthakker: > if no element it passed in the DragRotateHandler will default to the canvasContainer

Ah, I see!  Good point...  but even so, I think we should avoid it: that default behavior of `DragRotateHandler` isn't documented as part of its internal API, and it could easily end up being removed in a refactor.  

> Especially not to break stuff that's already out there.

What do you see as potentially breaking?

jfirebaugh: As @anandthakker suggests, we do need this change. In addition to the potential future compatibility issues, there is also the fact that the map creates and registers a `DragRotateHandler` of its own. If we let the navigation control create a second one bound to the same element, that will result in duplicate event handlers and probably produce unexpected behavior.

matijs:
@anandthakker is there anything else I need to do for this? :)

anandthakker:
Hi @matijs - yes, this comment https://github.com/mapbox/mapbox-gl-js/pull/5348#discussion_r142161439 still stands -- the `DragRotateHandler` should only be added if the `this._compass` element exists.

matijs:
@anandthakker Ugh… I'm so sorry. That conversation was hidden from me by GitHub 😔

I'll probably have some time to look into this later this week.

anandthakker:
No problem - thanks for following up!

matijs:
@anandthakker finally got round to fixing this :)

Issue #3904, opened 2017-01-08
A way to limit animation duration on flyTo or fitBounds


nicooprat:
Hi there. Is there any way to retrieve a transition duration before or after it begins? I'd like to avoid too long transitions on camera actions (`flyTo`, `fitBounds`, etc.) when the target point is too far away, like Google Maps does. I'd rather avoid to redo the entire calculations done internally... For example:

1. Map is centered on point A
2. Programmatically make it centers on point B
3. Transition begins
4. Check transition duration (looks like it's set [here](https://github.com/mapbox/mapbox-gl-js/blob/7427bfcf85777cc45123b27affb5ff4774bc8a79/js/ui/camera.js#L695))
5. If duration > 3s, trigger 2. again with `{animate: false}`

Can't find any proper or hacky way to do it :) For sure the best scenario would be an option like `maxDurationForAnimation`!

Thanks.

stevage:
Maybe just an option, `"maxTime"` to those functions, since that's your actual goal? What do you want to happen if the transition would take longer than that, though? Jump to?

nicooprat:
Yes that would be perfect!

mourner:
I like the `maxDuration` option suggestion. We would then either set the duration to 0 if calculated `flyTo` `duration` exceeds `maxDuration`, or we could do `duration = Math.min(duration, maxDuration)` to limit the duration but still animate every time.

nicooprat:
Yes but the 2nd option could mean super fast transition from Pekin to New York at zoom 15 for example, doesn't it? (Should I edit the title ?)

mourner:
@nicooprat yeah, it looks like the first option would be more practical, and in line with what some other mapping APIs do.

stepankuzmin:
I've created PR for this issue https://github.com/mapbox/mapbox-gl-js/pull/5349

nicooprat:
Thanks guys :)

PR #5349, opened 2017-09-25
limit animation duration on flyTo with maxDuration option

stepankuzmin:
Hi there! I've added the `maxDuration` option to limit animation duration on `flyTo` according to https://github.com/mapbox/mapbox-gl-js/issues/3904.

If this is the right way to do it, I'll add tests and documentation.

## Launch Checklist

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -665,6 +665,7 @@ class Camera extends Evented {
      *     It does not correspond to a fixed physical distance, but varies by zoom level.
      * @param {number} [options.screenSpeed] The average speed of the animation measured in screenfuls
      *     per second, assuming a linear timing curve. If `options.speed` is specified, this option is ignored.
+     * @param {number} [options.maxDuration] The animation's maximum duration, measured in milliseconds.
      * @param eventData Additional properties to be added to event objects of events triggered by this method.
      * @fires movestart
      * @fires zoomstart
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -803,6 +803,10 @@ class Camera extends Evented {
             options.duration = 1000 * S / V;
         }
 
+        if (options.maxDuration) {
+            options.duration = Math.min(options.duration, options.maxDuration);
+        }
+
         this.zooming = true;
         this.rotating = (startBearing !== bearing);
         this.pitching = (pitch !== startPitch);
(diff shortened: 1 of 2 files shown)


src/ui/camera.js
```diff
@@ -665,6 +665,7 @@ class Camera extends Evented {
      *     It does not correspond to a fixed physical distance, but varies by zoom level.
      * @param {number} [options.screenSpeed] The average speed of the animation measured in screenfuls
      *     per second, assuming a linear timing curve. If `options.speed` is specified, this option is ignored.
+     * @param {number} [options.maxDuration] The animation's maximum duration, measured in milliseconds.
```
mourner: Let's clarify what happens if duration exceeds maxDuration in the docs.

stepankuzmin: Thanks for the review! I've updated the PR.

src/ui/camera.js
```diff
@@ -803,6 +805,10 @@ class Camera extends Evented {
             options.duration = 1000 * S / V;
         }
 
+        if (options.maxDuration && options.duration > options.maxDuration) {
+            options.duration = 0;
```
mollymerp: I think it would make more sense to set `options.duration = options.maxDuration` in this case.

stepankuzmin: Thanks for the review, @mollymerp! According to the last @mourner's comment in  https://github.com/mapbox/mapbox-gl-js/issues/3904, it was decided to reset `duration` to `0`. So should I change it, or not?

mourner: @mollymerp yeah, that was intentional — consider a use case where you need to switch to a view from another part of the world, but you don't want it to take forever to animate. Setting to `maxDuration` would make the animation too fast, since it's a very long fly path — instead, just switching to the new view in a moment is usually a better experience IMO.

stepankuzmin: So everything is fine, then? Should I write tests and add a new example?

mourner: @stepankuzmin definitely worth adding a test! I don't think a small option like this needs a separate example though.

stepankuzmin:
Thanks for the review! I've added the test.

src/ui/camera.js
mollymerp: @mourner alright fair enough! I would think if a user had a case with a very long "flight path" they would rather explicitly set duration to 0 but I can definitely see both as valid 🤷‍♀️

mourner: > I would think if a user had a case with a very long "flight path" they would rather explicitly set duration to 0
@mollymerp the user doesn't know beforehand how much the flight will take between two given positions (e.g. say you have a list of locations around the world and you click to switch between them in arbitrary order) — this is the primary reason for introducing `maxDuration` in the first place, and is similar to how other libs like Leaflet and Google Maps do it, disabling animation automatically if the flight path would be too long.

PR #5350, opened 2017-09-25
Add ['heatmap-density'] expression

anandthakker:
This adds a `["heatmap-density"]` expression and special-case handling when converting stop functions to use that instead of `["zoom"]` for the `heatmap-color` style property.

### DIFF
--- a/src/style-spec/function/compile.js
+++ b/src/style-spec/function/compile.js
@@ -12,7 +12,7 @@ const definitions = require('./definitions');
 const evaluationContext = require('./evaluation_context');
 const {
     isFeatureConstant,
-    isZoomConstant
+    isGlobalPropertyConstant
 } = require('./is_constant');
 
 import type { Type } from './types.js';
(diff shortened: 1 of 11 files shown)


anandthakker:
cc @mourner

anandthakker:
Oops, yep will fix.

Re: default values hack -- I sorta think the existing hack is alright / involves the minimal amount of special-case code.  Can we keep that as is?

mourner:
> Re: default values hack -- I sorta think the existing hack is alright / involves the minimal amount of special-case code. Can we keep that as is?

I mean, the requirement to have `default` in the stops function provided (otherwise the code breaks).

anandthakker:
Oh, right, sorry!  Yeah, let me try to tack on a fix here.

anandthakker:
@mourner when I remove the `default:` from the stops function for `heatmap-color` in `heatmap.html`, it still works for me.

mourner:
Heh, maybe you or me accidentally fixed this somehow in other commits :)

PR #5363, opened 2017-09-27
Prevent tests from outputting to console.error or console.warn

### LINKED ISSUE #3900: Make tests fail if any extraneous log output is produced
Our tests are producing more and more extraneous log output. In the long term, this will make our test output harder to read may be indicative of log spew while using GL JS too.

### LINKED ISSUE #5352: Functions converted to expressions should not produce warnings
It seems that in various situations, evaluating an expression-converted function can produce a console warning. Among other things, this pollutes test output (#3900) with warnings like "No default property value available" and "Expected value to be of type Number, but found Null instead."

While expressions may have safer semantics, functions should continue to behave how they have always behaved. Unless I'm missing something about the purpose of these warnings, I don't think it's necessary, at this point, to call out the existing quirks of function evaluation behavior with warnings.

cc @anandthakker

jfirebaugh:
* Prevent current tests from outputting to `console.{error,warn}`. Fixes #5352.
* Fail future tests that do so. Fixes #3900.
* Various other test cleanups; see commits for details.

### DIFF
--- a/bench/lib/create_style.js
+++ b/bench/lib/create_style.js
@@ -13,7 +13,8 @@ class StubMap extends Evented {
 
 module.exports = function (styleJSON: StyleSpecification): Promise<Style> {
     return new Promise((resolve, reject) => {
-        const style = new Style(styleJSON, (new StubMap(): any), {});
+        const style = new Style((new StubMap(): any));
+        style.loadJSON(styleJSON);
 
         style
             .on('style.load', () => resolve(style))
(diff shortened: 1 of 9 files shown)


anandthakker:
> Added 6d5f1ad 

Err, meant to say that I pushed it, not that I added it to this branch.

jfirebaugh:
Nice, thanks!

jfirebaugh:
There's one left -- style_transition.test.js is triggering "No default property value available". (It's hard to tell what this test is testing and whether it's important to preserve the behavior.)

jfirebaugh:
@anandthakker https://github.com/mapbox/mapbox-gl-js/commit/d5c3ed0821df0fa96f2eb5ab38ce5125ebd0030f look ok?

anandthakker:
@jfirebaugh ah - yep 👍.  I think that must be vestigial or something, LGTM

PR #5364, opened 2017-09-27
split placement control into separate file

ansis:
#### Background

A single round of global symbol placement sometimes gets split over multiple frames for performance reasons. This means we need to keep track of how far into the placement we got in the previous step. es6 generators would be ideal for managing this control flow but unfortunately we can't use them. Instead we need to manually keep track of the progress.

#### Changes
The first commit separates the placement stuff into a separate file so that it's less intertwined with `Map` and `Style`.

The second commit is unrelated cleanup that generalizes dynamic VAO buffer binding. We don't need an arg for each type of dynamic buffer because to the VAO they all look the same. This removes the `collisionVertexBuffer` arg from VAOs and renames `opacityVertexBuffer` to make it more general.

@ChrisLoer

### DIFF
--- a/src/render/vertex_array_object.js
+++ b/src/render/vertex_array_object.js
@@ -13,8 +13,7 @@ class VertexArrayObject {
     boundIndexBuffer: ?IndexBuffer;
     boundVertexOffset: ?number;
     boundDynamicVertexBuffer: ?VertexBuffer;
-    boundOpacityVertexBuffer: ?VertexBuffer;
-    boundCollisionVertexBuffer: ?VertexBuffer;
+    boundDynamicVertexBuffer2: ?VertexBuffer;
     vao: any;
     gl: WebGLRenderingContext;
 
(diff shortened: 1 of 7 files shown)


src/render/vertex_array_object.js
```diff
@@ -13,7 +13,7 @@ class VertexArrayObject {
     boundIndexBuffer: ?IndexBuffer;
     boundVertexOffset: ?number;
     boundDynamicVertexBuffer: ?VertexBuffer;
-    boundOpacityVertexBuffer: ?VertexBuffer;
+    boundDynamicVertexBuffer2: ?VertexBuffer;
     boundCollisionVertexBuffer: ?VertexBuffer;
```
ChrisLoer: `boundCollisionVertexBuffer` should be removed here, right?

ansis: 👍

PR #5365, opened 2017-09-27
Benchmarks: use robust estimators and report effect size

### LINKED ISSUE #5361: Improve benchmark statistical analysis
Recording @anandthakker's followup ideas from #5340:

> * Replace our mean and standard deviation statistics with a trimmed mean and Windsorized standard deviation. This will make them more robust to outliers, so that we can report the difference between the (trimmed) means as one metric of effect size. (We certainly don't want to ignore outliers. However, throwing them into the mean doesn't really help us to assess them. Instead, we're better off handling them separately: e.g., by inspecting the plots for suspicious changes in the number or pattern of outliers.)
> * Also calculate and report the "probability of superiority" (or, in our case, "inferiority"): i.e., the probability that a randomly chosen sample from the non-master branch is slower than a randomly chosen sample from the master branch. This is a simple, intuitive-to-understand score that's insensitive to outliers and I think would support the difference-between-trimmed-means metric nicely.

anandthakker:
Fixes #5361

### DIFF
--- a/bench/benchmarks.js
+++ b/bench/benchmarks.js
@@ -13,6 +13,7 @@ window.mapboxglVersions.push(version);
 function register(Benchmark) {
     window.mapboxglBenchmarks[Benchmark.name] = window.mapboxglBenchmarks[Benchmark.name] || {};
     window.mapboxglBenchmarks[Benchmark.name][version] = new Benchmark();
+    window.mapboxglBenchmarks[Benchmark.name].master = new Benchmark();
 }
 
 register(require('./benchmarks/layout'));
--- a/bench/benchmarks.js
+++ b/bench/benchmarks.js
@@ -13,7 +13,6 @@ window.mapboxglVersions.push(version);
 function register(Benchmark) {
     window.mapboxglBenchmarks[Benchmark.name] = window.mapboxglBenchmarks[Benchmark.name] || {};
     window.mapboxglBenchmarks[Benchmark.name][version] = new Benchmark();
-    window.mapboxglBenchmarks[Benchmark.name].master = new Benchmark();
 }
 
 register(require('./benchmarks/layout'));
--- a/bench/benchmarks_view.js
+++ b/bench/benchmarks_view.js
@@ -261,15 +261,17 @@ class BenchmarkRow extends React.Component {
     render() {
         const endedCount = this.props.versions.filter(version => version.status === 'ended').length;
 
-        let effectSize = '';
+        let master;
+        let current;
+        if (/master/.test(this.props.versions[0].name)) {
+            [master, current] = this.props.versions;
+        } else {
+            [current, master] = this.props.versions;
+        }
+
+        let change = '';
+        let pInferiority = '';
         if (endedCount === 2) {
-            let master;
-            let current;
-            if (this.props.versions[0].name === 'master') {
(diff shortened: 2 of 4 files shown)


anandthakker:
@jfirebaugh draft implementation of outlier-robust mean/deviation and effect size summaries.  Eg:

screen shot 2017-09-27 at 4 58 34 pm

anandthakker:
The probability of inferiority needs to be labeled more clearly.  It is NOT the probability that the branch is slower, but rather the probability that a random run of the benchmark from current branch is slower than a random run from master.  (We'll have to get some experience with this metric to learn what a good critical values for it are.)

jfirebaugh:
Can we arrange the layout to incorporate these:

<table>
<tr><th><th>HEAD<th>MASTER
<tr><td>R² Slope / Correlation<td>x<td>y
<tr><td>(Windsorized) Deviation <td> x <td> y
<tr><td>(Trimmed) Mean <td> x <td> y
<tr><td>Change <td colspan=2>+z (y std devs)</td>
<tr><td>P(HEAD > master) <td colspan=2>p</td>
</table>

(Dropping minimum since it's easily read from the plot.)

anandthakker:
@jfirebaugh good idea, moved it into the table.  I'd be for keeping minimum in the table for a while, until we get a feel for which of these stats we find most meaningful in practice.

anandthakker:
We should also add emoji indicators for Change / P(inferiority) -- different than the regression ones, though, as these would indicate content, not reliability, of results

jfirebaugh:
How about (https://github.com/mapbox/mapbox-gl-js/commit/43059ae0179707f732d06627d2082ab2425626c6):

image

bench/benchmarks.js
```diff
@@ -13,6 +13,7 @@ window.mapboxglVersions.push(version);
 function register(Benchmark) {
     window.mapboxglBenchmarks[Benchmark.name] = window.mapboxglBenchmarks[Benchmark.name] || {};
     window.mapboxglBenchmarks[Benchmark.name][version] = new Benchmark();
+    window.mapboxglBenchmarks[Benchmark.name].master = new Benchmark();
```
jfirebaugh: Replace with https://github.com/mapbox/mapbox-gl-js/commit/e3bec20d71a8096118df8e8219a09758560f2394

jfirebaugh:
@ChrisLoer Are these colors accessible for deuteranopes?

anandthakker:
@jfirebaugh okay, I'm punting on the emoji annotations.  7aed18c tweaks the presentation a bit more -- see the commit message re: probability.   Also moved regression back down the table, because I think with outliers influencing the regression, the trimmed mean is generally going to be a more reliable estimate

ChrisLoer:
> Are these colors accessible for deuteranopes?

They're not great on that front -- I have to squint and look back and forth a lot to make sure which line is which. Something like a blue would have a lot more contrast for me.

anandthakker:
Good point re: colors -- changed it to a pair chosen from http://colorbrewer2.org/#type=qualitative&scheme=Dark2&n=3 -- LMK if that's not ideal.

anandthakker:
screen shot 2017-09-28 at 6 29 13 am

PR #5368, opened 2017-09-27
Style test cleanup

jfirebaugh:
Couple small changes I noticed in the course of #5367.

### DIFF
--- a/test/unit/style/style.test.js
+++ b/test/unit/style/style.test.js
@@ -386,38 +386,6 @@ test('Style#_resolve', (t) => {
         });
     });
 
-    t.test('handles ref layer preceding referent', (t) => {
-        const style = new Style(new StubMap());
-        style.loadJSON({
-            "version": 8,
-            "sources": {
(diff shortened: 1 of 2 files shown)


test/unit/style/style.test.js
```diff
@@ -360,67 +384,6 @@ test('Style#update', (t) => {
...
-
-        style.on('error', (error) => { t.error(error); });
-
-        style.on('style.load', () => {
-            t.ok(style.getLayer('fill') instanceof StyleLayer);
-            t.end();
-        });
-    });
-
-    t.test('handles ref layer preceding referent', (t) => {
```
mourner: What was that test about, is it OK to kill?

jfirebaugh: It's tested here: https://github.com/mapbox/mapbox-gl-js/blob/b9e10b939c6a3fe5d7ecac209f751b4871970ede/test/unit/style-spec/deref.test.js#L29

PR #5372, opened 2017-09-28
Bypass expression parsing and compiling for top-level constants

jfirebaugh:
image

### DIFF
--- a/src/style-spec/function/index.js
+++ b/src/style-spec/function/index.js
@@ -53,19 +53,37 @@ type StylePropertySpecification = {
     value: 'number' | 'string' | 'boolean',
     length?: number,
     default?: Array<mixed>
+} | {
+    type: 'color',
+    default?: string
 };
 
 type StylePropertyValue = null | string | number | Array<string> | Array<number>;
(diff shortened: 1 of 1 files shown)


src/style-spec/function/index.js
```diff
@@ -53,19 +53,37 @@ type StylePropertySpecification = {
...
+    if (typeof parameters === 'string' && propertySpec.type === 'color') {
+        const color = parseColor(parameters);
+        return {
+            isFeatureConstant: true,
+            isZoomConstant: true,
+            evaluate() { return color; }
+        };
+    }
 
+    if (parameters === null || typeof parameters !== 'object' || Array.isArray(parameters)) {
```
edpop: Is it ever possible, when a variable is not an object, but is an array?

jfirebaugh: Not sure I understand the question. `typeof [] === 'object'`, right?

edpop: Yes, I can't imagine, when `typeof x !== 'object' && Array.isArray(x)` is true

PR #5376, opened 2017-09-28
Fix benchmarks page when there's only a single version

jfirebaugh:


bench/benchmarks_view.js
```diff
@@ -307,7 +307,7 @@ class BenchmarkRow extends React.Component {
                     {this.renderStatistic('(20% trimmed) Mean',
                         (version) => <p>
                             {formatSample(version.summary.trimmedMean)} ms
-                            {version.name === current.name && change}
+                            {change && version.name === current.name}
```
anandthakker: This should still evaluate to `change` (rather than a boolean expression).

`{current && version.name === current.name && change}`?

jfirebaugh: 🤦‍♂️

PR #5380, opened 2017-09-28
fix raster-masking vertex indexing

mollymerp:
There was a bug in the raster clipping masks with the order of the mask vertices causing rendering artifacts with overlapping triangles which were only noticeable on zooming with translucent raster tiles. 

This PR changes to use a `SegmentVector` with an explicit `IndexArray` which solves the overlapping triangles when I manually tested the debug page, but I'm not sure the best way to write a unit/rendering test for this. 

## Launch Checklist

 - [ ] post benchmark scores


mollymerp:
this very closely matches the [native implementation](https://github.com/mapbox/mapbox-gl-native/blob/ba7574ec51e2cfd4a55565d5a8337c47f1784bc2/src/mbgl/renderer/buckets/raster_bucket.cpp#L48-L104)

mollymerp:
added a render test – here's the corresponding test on master with the artifacts
actual

PR #5386, opened 2017-09-29
Use old function code rather than converting to expressions

### LINKED ISSUE #5379: Reinstate direct implementation for old-style functions
Per chat w/ @mourner and @jfirebaugh, the plan is to remove the automatic migration of old-style functions to expressions, instead reinstating a direct implementation for them. (We may be able to at least share some of the stops-interpolating logic between this and `evaluation_context`).

This would make expressions a fully opt-in feature.

jfirebaugh:
Fixes #5379.

Clear improvements on `LayoutDDS`, `Paint`, and `StyleLayerCreate`. `QueryBox` seems to be slightly slower (?). The benchmarks I added in #5384 turn out to be too micro to be effective (not that they should show any changes here).

image

### DIFF
--- a/src/style-spec/expression/index.js
+++ b/src/style-spec/expression/index.js
@@ -62,9 +62,7 @@ type StylePropertySpecification = {
 type StylePropertyValue = null | string | number | Array<string> | Array<number>;
 type FunctionParameters = DataDrivenPropertyValueSpecification<StylePropertyValue>
 
-function createExpression(expression: mixed,
-                          propertySpec: StylePropertySpecification,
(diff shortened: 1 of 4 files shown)


src/style-spec/expression/index.js
```diff
@@ -87,7 +85,7 @@ function createExpression(expression: mixed,
             }
             return val;
         } catch (e) {
-            if (!options.isConvertedFunction && !warningHistory[e.message]) {
+            if (!warningHistory[e.message]) {
```
anandthakker: Just noting, we'll possibly need something like this again if/when we reinstate legacy function => expression conversion.

PR #5388, opened 2017-09-30
Accommodate microbenchmarks

anandthakker:
Accommodate microbenchmarks by (possibly) running >1 iteration per measurement.

Attached a couple riders, since they address issues that are particularly prevalent with microbenchmarks
 - Clamp the t-axis of the statistics plot to prevent outliers from scaling away much of the detail in KDE / scatter plots
 - Visualize both mean and trimmed mean

### DIFF
--- a/bench/lib/benchmark.js
+++ b/bench/lib/benchmark.js
@@ -2,9 +2,22 @@
 
 'use strict';
 
+// According to https://developer.mozilla.org/en-US/docs/Web/API/Performance/now,
+// performance.now() should be accurate to 0.005ms. Set the minimum running
+// time for a single measurement at 5ms, so that the error due to timer
+// precision is < 0.1%.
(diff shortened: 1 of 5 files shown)


anandthakker:
Example of how outliers look:

screen shot 2017-09-30 at 12 28 31 pm

(and when the mean gets skewed into outlier territory:)

screen shot 2017-09-30 at 12 28 21 pm

anandthakker:
@jfirebaugh LMK what you think of this cropping stuff -- I left it as an option for now in case you wanted to look at cropped & uncropped at the same time

bench/lib/benchmark.js
```diff
@@ -2,9 +2,18 @@
 
 'use strict';
 
+const minTimeForMeasurement = 0.005 * 1000;
```
jfirebaugh: Can you comment the rationale for this value?

PR #5396, opened 2017-10-03
Non-eval expressions preparation

jfirebaugh:
All the preparatory work for removing `new Function` from expression compilation, in a separate PR for benchmarking isolation. See individual commits for details.

Not sure why, but this alone improved evaluation performance:

image

[Full benchmark results](https://bl.ocks.org/anonymous/raw/4ada00ef61424e6c723a7027d5f733da/)


src/style-spec/validate/validate_expression.js
```diff
@@ -1,32 +1,21 @@
...
-        }
-
-        const curve = findZoomCurve(compiled.expression);
-        if (curve instanceof Curve) {
-            return [];
-        } else if (curve) {
-            return [new ValidationError(`${key}${curve.key}`, options.value, curve.error)];
-        } else {
-            return [new ValidationError(`${key}`, options.value, '"zoom" expression may only be used as input to a top-level "curve" expression.')];
-        }
```
anandthakker: I think this check makes sense here, rather than in `creatExpression`, because the latter will be used in contexts where the zoom curve restriction isn't applicable -- for example, expressions used for filters can use `[zoom]` anywhere.

jfirebaugh: I think it's better to have validation and error generation centralized, so that the `"zoom"` check uses the same conventions and types as other static errors. To control where `"zoom"` is allowed we can use an option. I assume that's what the (currently unused) `disallowNestedZoom` option was -- when it's needed again, it can be passed on to `createExpression`.

PR #5398, opened 2017-10-04
 Non-eval'ing expressions

### LINKED ISSUE #5347: Assess & improve performance of creating expressions
We've made some good improvements on expression _evaluation_ performance, but we should also look into the cost of creating them (parsing, typechecking, compiling), and also of first converting stop functions into expressions.  It's surely slower than creating pre-expression functions, and so we should expect the performance of parsing a style layer to have been degraded.

jfirebaugh:
Fixes #5347. Faster across the board:

Chrome | Safari | Firefox
--- | --- | ---
[Create](https://bl.ocks.org/anonymous/raw/a49040a329bb25522f826a39126fcac2/) | [Create](https://bl.ocks.org/anonymous/raw/1e108c7b64d7d14aa9401bce452bad38/) | [Create](https://bl.ocks.org/anonymous/raw/b25406e0a972cdc780056c64297848d9/)
[Evaluate](https://bl.ocks.org/anonymous/raw/e396eb525055a0f84469322e44d0bedd/) | [Evaluate](https://bl.ocks.org/anonymous/raw/a0f7db1a6cd1287006f225228e4ce9d9/) | [Evaluate](https://bl.ocks.org/anonymous/raw/8d329da579d1b3925a0bc187978d4f1f/)

(Benchmarking page is pretty busted in Firefox. I'll work on fixing that.)


src/style-spec/expression/compound_expression.js
```diff
@@ -2,37 +2,37 @@
...
         this.type = type;
-        this.compileFromArgs = compileFromArgs;
+        this._evaluate = evaluate;
         this.args = args;
     }
 
-    compile(ctx: CompilationContext) {
-        return this.compileFromArgs(ctx, this.args);
+    evaluate(ctx: EvaluationContext) {
+        return this._evaluate.apply(ctx, this.args.map(arg => arg.evaluate(ctx)));
```
anandthakker: Evaluating arguments ahead of time means that `&&` and `||` won't stop early, i.e. `["||", true, ["error", ""]]` would produce an error rather than evaluating to `true`.  We should make these a special form (and I should have had tests for this).  I'm gonna rebase the `expressions-filters` to pick up these changes, and can pick this up as part of that work.

jfirebaugh: 👍  Filed https://github.com/mapbox/mapbox-gl-js/issues/5403 to track this.

PR #5401, opened 2017-10-04
error if invalid 'before' argument is passed to Map#addLayer

### LINKED ISSUE #3868: Unexpected behaviour when calling "Map#addLayer" with invalid "before" argument

mollymerp:
fix #3868

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -568,7 +568,11 @@ class Style extends Evented {
 
         layer.setEventedParent(this, {layer: {id: id}});
 
-        const index = before ? this._order.indexOf(before) : this._order.length;
+
+        const beforeIndex = before ? this._order.indexOf(before) : -1;
+        const index = before && beforeIndex !== -1 ? beforeIndex : this._order.length;
+        if (before && beforeIndex === -1) util.warnOnce(`Layer with id "${before}" does not exist on this map. Adding layer to the top of the stack.`);
+
         this._order.splice(index, 0, id);
 
         this._layers[id] = layer;
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -569,9 +569,11 @@ class Style extends Evented {
         layer.setEventedParent(this, {layer: {id: id}});
 
 
-        const beforeIndex = before ? this._order.indexOf(before) : -1;
-        const index = before && beforeIndex !== -1 ? beforeIndex : this._order.length;
-        if (before && beforeIndex === -1) util.warnOnce(`Layer with id "${before}" does not exist on this map. Adding layer to the top of the stack.`);
+        const index = before ? this._order.indexOf(before) : this._order.length;
+        if (before && index === -1) {
+            this.fire('error', { message: new Error(`Layer with id "${before}" does not exist on this map.`)});
+            return;
+        }
 
         this._order.splice(index, 0, id);
 
--- a/test/unit/style/style.test.js
+++ b/test/unit/style/style.test.js
@@ -1006,6 +1006,28 @@ test('Style#addLayer', (t) => {
         });
     });
 
+    t.test('warns if before layer does not exist', (t) => {
+        const style = new Style(new StubMap());
+        style.loadJSON(createStyleJSON({
+            layers: [{
+                id: 'a',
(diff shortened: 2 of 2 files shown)


src/style/style.js
```diff
@@ -568,7 +568,10 @@ class Style extends Evented {
 
         layer.setEventedParent(this, {layer: {id: id}});
 
-        const index = before ? this._order.indexOf(before) : this._order.length;
+        const beforeIndex = before ? this._order.indexOf(before) : 0;
```
peterqliu: @mollymerp is `0` the top or bottom of the stack? Hitting `map.style._order` in my console, the first layer in the array is visually under all other layers on the map.

mollymerp: I suppose this is misleading – `beforeIndex` isn't actually used unless `before` exists so the `0` value never gets assigned to `index`. I made it 0 instead of `null` to appease flow but I could make this clearer.

src/style/style.js
```diff
@@ -568,7 +568,11 @@ class Style extends Evented {
 
         layer.setEventedParent(this, {layer: {id: id}});
 
-        const index = before ? this._order.indexOf(before) : this._order.length;
+
+        const beforeIndex = before ? this._order.indexOf(before) : -1;
+        const index = before && beforeIndex !== -1 ? beforeIndex : this._order.length;
+        if (before && beforeIndex === -1) util.warnOnce(`Layer with id "${before}" does not exist on this map. Adding layer to the top of the stack.`);
```
jfirebaugh: Let's report this via `this.fire('error', ...)` for consistency with non-existent layer IDs in `moveLayer`, `removeLayer`, etc.

test/unit/style/style.test.js
```diff
@@ -1006,6 +1006,28 @@ test('Style#addLayer', (t) => {
...
+            layers: [{
+                id: 'a',
+                type: 'background'
+            }, {
+                id: 'b',
+                type: 'background'
+            }]
+        }));
+        const layer = {id: 'c', type: 'background'};
+        t.stub(console, 'error');
```
jfirebaugh: Bind a handler for the `error` event rather than stubbing `console.error` (see the subsequent test for example).

src/style/style.js
```diff
@@ -568,7 +568,12 @@ class Style extends Evented {
 
         layer.setEventedParent(this, {layer: {id: id}});
 
-        const index = before ? this._order.indexOf(before) : this._order.length;
+
+        const beforeIndex = before ? this._order.indexOf(before) : -1;
+        const index = before && beforeIndex !== -1 ? beforeIndex : this._order.length;
+        if (before && beforeIndex === -1) this.fire('error', {
```
jfirebaugh: Instead of adding to the top of the stack, return early without making changes in the case of an error.

src/style/style.js
```diff
@@ -568,7 +568,12 @@ class Style extends Evented {
 
         layer.setEventedParent(this, {layer: {id: id}});
 
-        const index = before ? this._order.indexOf(before) : this._order.length;
+
+        const beforeIndex = before ? this._order.indexOf(before) : -1;
```
jfirebaugh: `this._order.length` can never be `-1`, so the conditionals can be simplified:

```
const index = before ? this._order.indexOf(before) : this._order.length;
if (index === -1) {
    // error
}

this._order.splice(index, ...);
```

mollymerp: ah good point! I still had to use `if (before && index === -1)` for flow to allow me to  include `before` in the error message.

PR #5404, opened 2017-10-05
More expression performance improvements

jfirebaugh:
https://bl.ocks.org/anonymous/raw/b0cfb339684b3b8a4d33edbc009b0dd0/

### DIFF
--- a/src/style-spec/expression/definitions/index.js
+++ b/src/style-spec/expression/definitions/index.js
@@ -29,7 +29,6 @@ const Case = require('./case');
 const Curve = require('./curve');
 const Coalesce = require('./coalesce');
 
-import type { Value } from '../values';
 import type { Expression } from '../expression';
 
 const expressions: { [string]: Class<Expression> } = {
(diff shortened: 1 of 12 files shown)


src/style-spec/expression/evaluation_context.js
```diff
@@ -9,7 +9,7 @@ import type { Feature } from './index';
 import type { Expression } from './expression';
 
 class EvaluationContext {
-    globals: {+zoom?: number};
+    globals: {zoom: number};
```
anandthakker: Is the purpose here to help V8 optimize by providing a fixed object structure?  Note that for `heatmap-color`, we intend to pass `{heatmapDensity: number}`, rather than `{zoom: number}` as the globals -- but we could make it `{heatmapDensity: number, zoom: -1}` or something, since this optimization seems worth it.

src/style-spec/expression/definitions/index.js
```diff
@@ -339,32 +341,32 @@ CompoundExpression.register(expressions, {
     '&&': [
         BooleanType,
         varargs(BooleanType),
-        (...args) => args.reduce((a, b) => a && b, true)
+        (ctx, args) => args.reduce((a, b) => a && b.evaluate(ctx), true)
```
anandthakker: With evaluation now happening within these definitions, `&&` and `||`  could be implemented so as not to evaluate all their arguments, without needing to be defined in a separate class.   I'm in favor of doing that -- it's true that `CompoundExpression` would then include both normal and special forms, but is there any advantage to avoiding that?

src/style-spec/expression/expression.js
```diff
@@ -10,7 +9,7 @@ export interface Expression {
     +type: Type;
 
     static parse(args: Array<mixed>, context: ParsingContext): ?Expression;
-    evaluate(ctx: EvaluationContext): Value;
+    evaluate(ctx: EvaluationContext): any;
```
anandthakker: Why not `Value`?

jfirebaugh: It forces every use of `evaluate` to cast the result through `any` to the actual concrete type. The right way to fix this is to make `Expression` a generic type on the result type, but I wanted to avoid getting into that for now.

src/style-spec/expression/evaluation_context.js
jfirebaugh: It's mainly to avoid an extra check in the implementation of `zoom` to satisfy flow.

src/style-spec/expression/definitions/index.js
jfirebaugh: No big advantage that I see. This change itself does fix short circuiting, though we should still add tests for that.

src/style-spec/expression/expression.js
anandthakker: Ah, right.  Yeah we should try the generic type approach, but definitely later since there's a decent chance that effort will stumble on flow limitations, heh

src/style-spec/expression/definitions/index.js
anandthakker: > This change itself does fix short circuiting

Oy -- yes, that it does 🤦‍♂️

Issue #5370, opened 2017-09-28
Style#setPaintProperty needs update for expressions


jfirebaugh:
This code hasn't been updated for expressions and will not do the right thing if passed a feature-varying expression:

https://github.com/mapbox/mapbox-gl-js/blob/ef5e3890f6e8a54c8fb5ffa2e1c96f6a8f0234b6/src/style/style.js#L791-L796

The ad-hoc check should be replaced with a call to parse the expression and check the `isFeatureConstant` property on the result.

cc @anandthakker

Issue #5222, opened 2017-09-01
Drop the `{ expression: }` wrapper object for expressions used in layout/paint properties


anandthakker:
@jfirebaugh had  💡realized 💡  that, except for font stack arrays, expressions are always distinguishable from other literal arrays that might currently be used as style properties.  Since it's very unlikely that anyone's using a font whose name clashes with one of our expressions, this means that we can drop the wrapper object, e.g.: `circle-color: { expression: [ 'to-color', ['get', 'x'] ] }` => `circle-color: [ 'to-color', ['get', 'x'] ]`

anandthakker:
cc @nickcordella @aparlato

mourner:
BTW, in a similar vein, could we drop array wrappers for some predefined values like `["properties"]`, `["zoom"]` etc.? So that array would still be accepted in case we want to add options in future, but a string literal too.

jfirebaugh:
My worry with that is that there may be contexts (either with the expression forms we have now, or with ones we add in the future) where both a string literal and `["properties"]`/`["zoom"]` are plausible arguments. Then the user would have to disambiguate by using brackets or not, and that would be confusing for someone who is not aware of the ambiguity. I'd prefer to keep it unambiguous all the time.

PR #5407, opened 2017-10-05
Preparatory expression refactors

anandthakker:


anandthakker:
@jfirebaugh couple of preparatory changes that I'd like to get merged to reduce need for invasive rebasing in #5253 and #5193

src/style-spec/expression/index.js
```diff
@@ -107,8 +141,9 @@ function createExpression(expression: mixed, expectedType: Type | null): StyleEx
     };
 }
 
-function createExpressionWithErrorHandling(expression: mixed, expectedType: Type | null, defaultValue: Value | null): StyleExpression {
-    expression = createExpression(expression, expectedType);
+function createExpressionWithErrorHandling(expression: mixed, options: StyleExpressionOptions): StyleExpression {
+    expression = createExpression(expression, options);
+    const defaultValue = options.defaultValue === undefined ? null : options.defaultValue;
```
anandthakker: Should we have `errorHandling: boolean` as an option and then expose only `createExpression()`?

src/style-spec/expression/index.js
```diff
@@ -24,19 +24,31 @@ export type Feature = {
     +properties: {[string]: any}
 };
 
+export type StyleExpressionContext = 'declaration' | 'filter';
```
jfirebaugh: `'declaration'` → `'property'`?

src/style-spec/expression/index.js
jfirebaugh: Yeah, we need to unify the two somehow for #5222 and #5370. I was trying to figure out a clean way to do it. Could we use an error return type by default, and push the try/catch up to `StyleDeclaration#calculate`? I avoided doing this in previous performance PRs because it would make the expression benchmarks not strictly comparable, but I'm 👍  on doing it in an independent PR.

Also, we should use `Color` ubiquitously so `unwrap` becomes unnecessary.

jfirebaugh: > Could we use an error return type by default, and push the try/catch up to `StyleDeclaration#calculate`?

Hmm, nope, that regresses layout performance, presumably because constant "evaluation" is also wrapped. https://bl.ocks.org/anonymous/raw/2b247bab7e59c3f83fa75ad8c7f2921d/

anandthakker: @jfirebaugh could we just decide whether to do error handling based on the presence or absence of `defaultValue`

PR #5414, opened 2017-10-05
Fix texture#update

mollymerp:
fixes a problem introduced by #5406 because `RGBAImage` and `AlphaImage` are declared as `Classes` but not used as such so `image instanceof RGBAImage` was always returning false. 

benchmarks work again:
https://bl.ocks.org/anonymous/raw/6bf016e150a227d2f59703c816100e87/

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/render/texture.js
+++ b/src/render/texture.js
@@ -1,7 +1,7 @@
 // @flow
 
 const {RGBAImage, AlphaImage} = require('../util/image');
-const {HTMLImageElement, HTMLCanvasElement, HTMLVideoElement, ImageBitmap, ImageData} = require('../util/window');
+const {HTMLImageElement, HTMLCanvasElement, HTMLVideoElement, ImageData, ImageBitmap} = require('../util/window');
 
 export type TextureFormat =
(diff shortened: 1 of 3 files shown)


src/render/texture.js
```diff
@@ -55,10 +55,10 @@ class Texture {
             gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, (true: any));
         }
 
-        if (image instanceof RGBAImage || image instanceof AlphaImage) {
-            gl.texImage2D(gl.TEXTURE_2D, 0, this.format, width, height, 0, this.format, gl.UNSIGNED_BYTE, image.data);
-        } else {
+        if (image instanceof HTMLImageElement || image instanceof HTMLCanvasElement || image instanceof HTMLVideoElement || image instanceof HTMLCanvasElement || image instanceof ImageData || image instanceof ImageBitmap) {
```
jfirebaugh: `HTMLCanvasElement` is checked twice here.

PR #5418, opened 2017-10-05
Fix isFeatureConstant test in Style#setPaintProperty

### LINKED ISSUE #5370: Style#setPaintProperty needs update for expressions

jfirebaugh:
Fixes #5370.


src/style/style_layer.js
```diff
@@ -193,7 +194,10 @@ class StyleLayer extends Evented {
         return false;
     }
 
-    updatePaintTransitions(options: any, globalOptions: any, animationLoop: any, zoomHistory: any) {
+    updatePaintTransitions(options: {transition?: boolean},
+                           globalOptions?: TransitionSpecification,
+                           animationLoop?: AnimationLoop,
+                           zoomHistory?: any) {
```
anandthakker: is zoomHistory here still `any` because of the `ZoomHistory | {}` in Style?

jfirebaugh: Yeah... we should fix that eventually.

PR #5423, opened 2017-10-06
Add expression validation special cases

### LINKED ISSUE #5415: line-dasharray must not accept expression with non-step curves
`line-dasharray` is a special case: it's type is `Array<number>`, but unlike other properties with that type, it's not interpolatable. That's why it's designated `"function": "piecewise-constant"` in the spec. It needs to disallow zoom curves with interpolation other than `["step"]`.

### LINKED ISSUE #5416: allow feature-varying expressions only for DDS-enabled properties
Non-DDS-enabled properties can't support expressions that use `get`, `has`, etc.

### LINKED ISSUE #5417: disallow expressions in "visibility" properties
This property must only be a literal `"visible"` or `"none"`.

jfirebaugh:
* visibility property can't use expressions (fixes #5417)
* Non-DDS properties can't use feature expressions (fixes #5416)
* line-dasharray must use step interpolation (fixes #5415)

https://bl.ocks.org/anonymous/raw/823b83f6fa8cc92f4bc2f4a73c44f8c1/

### DIFF
--- a/src/style-spec/expression/index.js
+++ b/src/style-spec/expression/index.js
@@ -163,7 +163,7 @@ function createExpression(expression: mixed,
     }
 
     const isFeatureConstant = isConstant.isFeatureConstant(parsed);
-    if (!isFeatureConstant && !propertySpec['property-function']) {
+    if (!isFeatureConstant && context === 'property' && !propertySpec['property-function']) {
         return {
             result: 'error',
             errors: [new ParsingError('', 'property expressions not supported')]
--- a/src/style-spec/expression/index.js
+++ b/src/style-spec/expression/index.js
@@ -30,12 +30,6 @@ export type GlobalProperties = {
 
 export type StyleExpressionContext = 'property' | 'filter';
 
-export type StyleExpressionOptions = {
-    context: StyleExpressionContext,
-    expectedType: Type | null,
-    defaultValue?: Value | null
-}
-
 export type StyleExpressionErrors = {
     result: 'error',
     errors: Array<ParsingError>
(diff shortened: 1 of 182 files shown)


src/style-spec/expression/index.js
```diff
@@ -131,18 +163,24 @@ function createExpression(expression: mixed, options: StyleExpressionOptions): S
     }
 
     const isFeatureConstant = isConstant.isFeatureConstant(parsed);
-    const isZoomConstant = isConstant.isGlobalPropertyConstant(parsed, ['zoom']);
+    if (!isFeatureConstant && !propertySpec['property-function']) {
```
anandthakker: `&& context === 'property'`

src/style-spec/validate/validate_expression.js
```diff
@@ -1,25 +1,15 @@
...
-    const defaultValue = options.key.endsWith('heatmap-color') ? [0, 0, 0, 0] :
-        getDefaultValue(options.valueSpec);
-    const expression = createExpression(
-        unbundle.deep(options.value), {
-            context: options.expressionContext,
-            expectedType: getExpectedType(options.valueSpec),
-            defaultValue
-        });
-
+    const expression = createExpression(unbundle.deep(options.value), options.valueSpec, 'property');
```
anandthakker: Shouldn't `'property'` be part of `options`, since this function will be used for validating filters, too?

test/integration/expression-tests/heatmap-density/basic/test.json
```diff
@@ -1,5 +1,5 @@
 {
-  "expectExpressionType": {"kind": "Color"},
+  "propertySpec": {"type": "color", "default": "rgb(0,0,0)"},
```
anandthakker: Might be more appropriate to have `default` here be a function (to exercise the special case handling in `getDefaultValue()`)

PR #5429, opened 2017-10-09
Return early from draw calls on layers with constant opacity=0

lbud:
We're completing full draw calls for layers even when their opacity is set to zero. This cuts back on render passes by bailing early from draw functions if a layer's opacity is constant at 0. One common use case for this is switching very quickly between several visualization layers without having to reparse buckets/re-download tiles.

Locally I replaced the Streets style in the Paint benchmark with a version where _half_ of the layers are set to opacity=0 — obviously this example is extremely contrived, but
image


src/render/draw_circle.js
```diff
@@ -12,6 +12,7 @@ module.exports = drawCircles;
 
 function drawCircles(painter: Painter, sourceCache: SourceCache, layer: CircleStyleLayer, coords: Array<TileCoord>) {
     if (painter.renderPass !== 'translucent') return;
+    if (layer.paint['circle-opacity'] === 0 && layer.paint['circle-stroke-opacity'] === 0) return;
```
anandthakker: With an eye towards https://github.com/mapbox/mapbox-gl-js/issues/3044, we should use `layer.isPaintValueFeatureConstant()` and `layer.getPaintValue()` instead of directly accessing `layer.paint`

lbud:
Good idea @anandthakker — moved these in https://github.com/mapbox/mapbox-gl-js/pull/5429/commits/09ed3de68501a68f1736cef550e567ff4653a11c.

jfirebaugh:
> I wonder if it makes sense to fold these checks into `isHidden()` for each layer type?

I seem to recall that's what we used to do, but we had reason to stop. Can you please review the history and see why?

lbud:
Good call @jfirebaugh — looks like this was removed in https://github.com/mapbox/mapbox-gl-js/pull/3120. It seems like the cause of that underlying bug was that we were also trying to be smart about the alpha channel in the color, which I think isn't necessary here — users trying to show/hide layers with opacity would do so with explicit `-opacity` properties. From that PR:

> The latest commit removes all opacity tests from the `StyleLayer#isHidden` method. This is unlikely to cause performance regressions in core styles because we don't often have 0 opacity layers in core styles. It's possible a 3rd party style may see a perf regression due to this change.

¯\_(ツ)_/¯

jfirebaugh:
Yeah, that's the one I was thinking of. In addition to that, we also tell people to use opacity 0 if they don't want the layer to be visible, but do want it to be queryable with `queryRenderedFeatures`. I believe including an opacity check in `isHidden` would break that, because `isHidden` affects whether the tiles are loaded. So I think we should short circuit in drawing, but not change `isHidden`.

lbud:
Okay, this now short-circuits only in draw functions with `isOpacityZero` methods per StyleLayer subclass. I've also added two query tests for invisible features (one `opacity=0` and one `visibility=none`) and a render test for circles with strokes but zero-opacity fills, since that's a slightly weird edge case here.

mourner:
@lbud let's also run the bench for a style with no opacity=0 layers since this may affect performance (many additional `getPaintValue` calls)

lbud:
@mourner with the regular Streets style there's no statistically significant change. Here are bench results from #Paint — ran it several times, always ± just a bit in either direction:

image

Full bench run:
image

@anandthakker and I discussed this yesterday as well — we could theoretically just use computed `layer.paint[<type>-opacity]`, but as he noted in https://github.com/mapbox/mapbox-gl-js/pull/5429#discussion_r143462457 we'd like to get away from doing that.

src/render/draw_background.js
```diff
@@ -9,6 +9,8 @@ import type StyleLayer from '../style/style_layer';
 module.exports = drawBackground;
 
 function drawBackground(painter: Painter, sourceCache: SourceCache, layer: StyleLayer) {
```
jfirebaugh: Change `layer: StyleLayer` to `layer: BackgroundStyleLayer`...

src/render/draw_raster.js
```diff
@@ -12,6 +12,7 @@ module.exports = drawRaster;
 
 function drawRaster(painter: Painter, sourceCache: SourceCache, layer: StyleLayer, coords: Array<TileCoord>) {
```
jfirebaugh: ...`layer: RasterStyleLayer`...

src/style/style_layer.js
```diff
@@ -41,6 +41,7 @@ class StyleLayer extends Evented {
     _layoutFunctions: {[string]: boolean};
 
     +createBucket: (parameters: BucketParameters) => Bucket;
+    +isOpacityZero: (zoom: number, property?: string) => boolean;
```
jfirebaugh: ...then you can remove this...

src/style/style_layer/symbol_style_layer.js
```diff
@@ -48,6 +48,13 @@ class SymbolStyleLayer extends StyleLayer {
         return (new SymbolBucket((parameters: any)): any);
     }
 
+    isOpacityZero(zoom: number, property?: string) {
```
jfirebaugh: ...and make `property` non-optional.

lbud: 👍 thanks @jfirebaugh!

PR #5434, opened 2017-10-09
Support expressions in filters (without conversion)

jfirebaugh:
Like #5193 but continue to use the old code for non-expression filters.

### DIFF
--- a/test/unit/style-spec/feature_filter.test.js
+++ b/test/unit/style-spec/feature_filter.test.js
@@ -3,6 +3,45 @@
 const test = require('mapbox-gl-js-test').test;
 const filter = require('../../../src/style-spec').featureFilter;
 
+test('expression, zoom', (t) => {
+    const f = filter(['>=', ['number', ['get', 'x']], ['zoom']]);
+    t.equal(f({zoom: 1}, {properties: {x: 0}}), false);
+    t.equal(f({zoom: 1}, {properties: {x: 1.5}}), true);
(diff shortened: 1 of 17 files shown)


jfirebaugh:
https://bl.ocks.org/anonymous/raw/c15be57dd06c39b2f5b1c71b7f1dd513/

The `FilterEvaluate` result from this run seems to be an anomaly. There's some difference, but not that bad.

https://bl.ocks.org/anonymous/raw/0bdb3084a58885a63f316d73bccf8030/

test/unit/style-spec/feature_filter.test.js
```diff
@@ -3,6 +3,45 @@
...
+test('expression, compare two properties', (t) => {
+    t.stub(console, 'warn');
+    const f = filter(['==', ['string', ['get', 'x']], ['string', ['get', 'y']]]);
+    t.equal(f({zoom: 0}, {properties: {x: 1, y: 1}}), false);
+    t.equal(f({zoom: 0}, {properties: {x: '1', y: '1'}}), true);
+    t.equal(f({zoom: 0}, {properties: {x: 'same', y: 'same'}}), true);
+    t.equal(f({zoom: 0}, {properties: {x: null}}), false);
+    t.equal(f({zoom: 0}, {properties: {x: undefined}}), false);
+    t.end();
+});
```
anandthakker: Let's add a test for the case where `any` and `all` are used as expressions

PR #5441, opened 2017-10-10
Reinstate function docs

### LINKED ISSUE #5223: Reinstate stop function documentation
In #4777, I _replaced_ the stop function documentation in the style spec docs with documentation for expression-based functions, but this was a mistake.  This should have been an additive change, and we should wait to remove stop function docs until (a) we have parity not just in gl-native but in each of the platform SDKs, (b) the expressions API has stabilized and been validated a bit by the community, and (c) our other guides and tutorials are updated to use expressions.

anandthakker:
Closes #5223

 - [x] Add deprecation notice for Function
 - [x] Add documentation for `curve` interpolation types, and then update the "supports {interpolated,piecewise constant} functions" links to point to `curve`.


anandthakker:
@jfirebaugh note some subsequent changes fleshing out `curve`

screen shot 2017-10-10 at 6 02 38 pm

jfirebaugh:
Remove `["step"]` from the final code block there.

PR #5448, opened 2017-10-11
Canvas Source Performance Increase

gpbmike:
A [recent fix](https://github.com/mapbox/mapbox-gl-js/pull/5155) for GPU related bug introduced a large performance hit to animated canvas sources. The intermediary step of copying data to an `ImageData` object, particularly with large canvases (my use case is 3000x3000), caused an unacceptable frame drop for my users.

This PR removes the intermediate step. The result is a dramatic increase in performance, back to 60fps.

I have confirmed that the change works with previously affected broadwell line GPUs referenced here: https://github.com/mapbox/mapbox-gl-js/issues/4262

`contextType` is no longer needed when creating a canvas source.

Tests do not need an update since there is not a good way to test rendering at the moment (see https://github.com/mapbox/mapbox-gl-js/pull/5303). 

I appreciate all the effort that was put into confirming and working around the GPU issue when it was first presented. It would seem as though the work around is not needed anymore. I don't know if that's the result of a change somewhere else in the mapbox codebase or something else entirely. 

Happy to answer questions.

## Launch Checklist

fixes https://github.com/mapbox/mapbox-gl-js/issues/5301

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "mapbox-gl",
+  "name": "@3drobotics/mapbox-gl",
   "description": "A WebGL interactive maps library",
   "version": "0.40.1",
   "main": "dist/mapbox-gl.js",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@3drobotics/mapbox-gl",
+  "name": "mapbox-gl",
   "description": "A WebGL interactive maps library",
   "version": "0.40.1",
   "main": "dist/mapbox-gl.js",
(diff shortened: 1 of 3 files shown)


lbud:
I'm mystified by what fixed this on affected GPUs, but not having easy access to one myself I'm going to have to leave that mystery for someone else. @gpbmike I'll take your word for it that those GPUs are fine now without the intermediary image buffer. 

I checked out this branch and there are just a few minor tweaks to make to fully revert https://github.com/mapbox/mapbox-gl-js/pull/5155. I hope you don't mind if I push to this branch myself in the interest of expediency so we can get this in the next release — we're _just_ about to cut a new release, so you opened this in the nick of time :)

lbud:
Moving to https://github.com/mapbox/mapbox-gl-js/pull/5449.

gpbmike:
Thanks @lbud 🙏

PR #5449, opened 2017-10-11
Revert CanvasSource intermediary image buffer fix

lbud:
Moving #5448 here to 🚢 in time for the next release. Thanks so much @gpbmike!

⚠️ **BREAKING** ⚠️: awkwardly, this reverts the breaking change introduced in #5155 and canvas sources will no longer have a `contextType` parameter, which is ultimately for the better, but… 😬 

> A recent fix for GPU related bug introduced a large performance hit to animated canvas sources. The intermediary step of copying data to an `ImageData` object, particularly with large canvases (my use case is 3000x3000), caused an unacceptable frame drop for my users.
> 
> This PR removes the intermediate step. The result is a dramatic increase in performance, back to 60fps.
> 
> I have confirmed that the change works with previously affected broadwell line GPUs referenced here: #4262
> 
> `contextType` is no longer needed when creating a canvas source.
> 
> Tests do not need an update since there is not a good way to test rendering at the moment (see #5303).
> 
> I appreciate all the effort that was put into confirming and working around the GPU issue when it was first presented. It would seem as though the work around is not needed anymore. I don't know if that's the result of a change somewhere else in the mapbox codebase or something else entirely.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "mapbox-gl",
+  "name": "@3drobotics/mapbox-gl",
   "description": "A WebGL interactive maps library",
   "version": "0.40.1",
   "main": "dist/mapbox-gl.js",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@3drobotics/mapbox-gl",
+  "name": "mapbox-gl",
   "description": "A WebGL interactive maps library",
   "version": "0.40.1",
   "main": "dist/mapbox-gl.js",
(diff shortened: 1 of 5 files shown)


gpbmike:
@lbud does this line need to be removed? https://github.com/mapbox/mapbox-gl-js/blob/master/flow-typed/style-spec.js#L115

I forgot to do that in my PR.

lbud:
Oops, thanks for pointing that out @gpbmike! Fixed in https://github.com/mapbox/mapbox-gl-js/pull/5454.

PR #5451, opened 2017-10-11
Move expression docs to their own section

jfirebaugh:
And demote function/filter to "Other", in preparation for stronger deprecation in the future.

### DIFF
--- a/docs/style-spec/_generate/index.html
+++ b/docs/style-spec/_generate/index.html
@@ -58,8 +58,22 @@ navigation:
     - title: Boolean
     - title: Number
     - title: Array
+  - title: Expressions
+    subnav:
+    - title: Types
+    - title: Feature data
+    - title: Lookup
+    - title: Decision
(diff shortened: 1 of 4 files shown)


docs/style-spec/_generate/index.html
```diff
@@ -820,7 +834,143 @@ <h3 class='space-bottom1'><a href='#types-array' title='link to array'>Array</a>
...
+        {% highlight js %}
+      {
+      "circle-color": [
+        'rgb',
+        // red is higher when feature.properties.temperature is higher
+        ["get", "temperature"],
+        0,
+        // blue is higher when feature.properties.temperature is lower
+        ["-", 100, ["get", "temperature"]]
+      ]
```
anandthakker: Let's either indent or remove the outer `{}`

anandthakker: Likewise for the "zoom-only" and "zoom-and-property" examples below

Issue #5373, opened 2017-09-28
Flow type test source files


jfirebaugh:
I've encountered several bugs in the tests in the last few days that would have been caught by type checking the test files. We should work towards enabling flow for all of our test sources.

PR #5454, opened 2017-10-12
Generate flow types from recent style-spec changes

lbud:
As pointed out in https://github.com/mapbox/mapbox-gl-js/pull/5449#issuecomment-335952705, that PR failed to remove the `contextType` parameter from the `CanvasSourceSpecification` flow type. I ran the build script to generate it and it looks like our type definitions are out of date for a few things. @jfirebaugh or @anandthakker — is it weird that flow testing on local machines/CI hasn't complained about these? 🤔

### DIFF
--- a/flow-typed/style-spec.js
+++ b/flow-typed/style-spec.js
@@ -78,10 +78,10 @@ declare type TileSourceSpecification = {
     "type": "vector" | "raster",
     "url"?: string,
     "tiles"?: Array<string>,
+    "bounds"?: [number, number, number, number],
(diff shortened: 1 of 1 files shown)


anandthakker:
> is it weird that flow testing on local machines/CI hasn't complained about these? 

Good question.  I think the reasons there were no complaints is that we're never explicitly calling `new CanvasSource(...)` except in a unit test (see also #5373) -- so there was no file that flow was checking where we actually omitted the `contextType` property.

Issue #5443, opened 2017-10-10
Actors don't properly clean up callback references


tgecho:
**mapbox-gl-js version**: 0.40.1

### Steps to Trigger Behavior

 1. Load any map (for example: https://www.mapbox.com/mapbox-gl-js/examples/)
 2. Count the number of callbacks retained by the worker actors:
    On the example page, run this in the console: `document.getElementById('demo').contentWindow.map.style.dispatcher.actors.reduce((total, actor) => Object.keys(actor.callbacks).length + total, 0)`

### Expected Behavior
The number of callbacks should rise and fall back to 0

### Actual Behavior
The number of callbacks continues to climb as you interact with the map

#### Background / Details
I'm investigating some an apparent memory leaks (in a much more complicated scenario with auto refreshing data), and I noticed references were being held open by this issue. Basically, [the actor saves a callback](https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/actor.js#L45), but in some cases the worker never calls back since many if its methods don't use the `done` param. The primary culprits seem to be setLayers and updateLayers in [source/worker.js](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/worker.js), though I noticed others.

Experimentally clearing them with `map.style.dispatcher.actors.forEach(a => a.callbacks = {})` seems to somewhat bring memory usage back down, though not completely.

jfirebaugh:
Good catch!

tgecho:
I'm trying my hand at a fix for this and I've noticed that `loadTile` is another leaky event type. It has a callback, so retaining it is correct, but apparently it doesn't always get called back, especially when zooming/panning.

PR #5455, opened 2017-10-12
Ensure worker callbacks are cleaned up

tgecho:
Fix for #5443 

The main change is adding a no callback path to `Dispatcher.broadcast`. I originally started to do this in `util.asyncAll`, but the required type changes were rippling further than I was comfortable with.

My primary testing method was to run this from the console: `map.style.dispatcher.actors.reduce((total, actor) => Object.keys(actor.callbacks).length + total, 0)`

With this set of fixes, the count always settled back down to zero across a variety of layer types.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -223,11 +223,11 @@ class GeoJSONSource extends Evented implements Source {
 
     unloadTile(tile: Tile) {
         tile.unloadVectorData();
-        this.dispatcher.send('removeTile', { uid: tile.uid, type: this.type, source: this.id }, () => {}, tile.workerID);
+        this.dispatcher.send('removeTile', { uid: tile.uid, type: this.type, source: this.id }, null, tile.workerID);
     }
 
     onRemove() {
-        this.dispatcher.broadcast('removeSource', { type: this.type, source: this.id }, () => {});
+        this.dispatcher.broadcast('removeSource', { type: this.type, source: this.id });
     }
 
     serialize() {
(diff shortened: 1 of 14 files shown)


anandthakker:
Thanks for working on this, @tgecho!

> The main change is adding a no callback path to Dispatcher.broadcast

This gives me pause -- keeping track of which worker messages call back and which ones don't, and adjusting use of `dispatcher` to suit, seems likely to be error prone.   🤔  Might it be better/safer to simply make all worker methods consistently take a callback?

jfirebaugh:
Or make _no_ worker methods take a callback, and instead always send a named message back. This is more like how gl-native works.

tgecho:
Totally understand the hesitation (which is partly why I called it out). I don't think it's too bad (or at least not really worse), since the dispatcher wouldn't need to change again. However, you could still have a caller that requests a callback that the worker doesn't support. It *seems* nice at first glance to not require callbacks for performance reasons, but I suppose it's not actually that expensive.

> Or make no worker methods take a callback, and instead send a named message back when necessary. This is more like how gl-native works.

That sounds like a bigger refactoring story than I'm comfortable biting off at the moment! :)

I'm happy to remove the dispatcher split and add callback handling to the worker methods as needed if that's what you would prefer. Or you can drop this PR and fix it another way. Just let me know!

anandthakker:
> I'm happy to remove the dispatcher split and add callback handling to the worker methods as needed if that's what you would prefer. 

@tgecho that would be great - yes please!  

I agree that going full-on pure message passing would be a much more involved refactor.  We may take it on at some point in the future, but I think it'd be great to get this fix in now.

tgecho:
Sorry about the delay. I think I got all the worker methods cleaned up. I ran through each layer type in debug and wasn't able to reproduce the memory leak.

jfirebaugh:
Thanks @tgecho!

PR #5456, opened 2017-10-13
fix link to expressions documentation in CHANGELOG

andrewharvey:
The current links in CHANGELOG.md to expressions documentation don't seem to use the right anchor.


jfirebaugh:
I did this in #5451 as well, though that's the `mb-pages` branch.

andrewharvey:
> I did this in #5451 as well, though that's the mb-pages branch.

:+1: Sorry I didn't see that. Either way better to have this fixed in master now as most people read the CHANGELOG from master, instead of waiting for mb-pages to be merged in the next release.

PR #5476, opened 2017-10-17
Fix bug in multi-signature expression type inference

anandthakker:
`overload[0][0]` was one level too deep; since `overload[0][0]` actually referred to the first parameter of the first signature, `Array.isArray(overload[0][0])` was always true, so we checked both the 1-ary and 2-ary versions of of `"-"` even when there were two arguments.

### DIFF
--- a/src/style-spec/expression/compound_expression.js
+++ b/src/style-spec/expression/compound_expression.js
@@ -51,12 +51,14 @@ class CompoundExpression implements Expression {
         const type = Array.isArray(definition) ?
             definition[0] : definition.type;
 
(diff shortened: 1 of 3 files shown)


jfirebaugh:
What was the bug?

anandthakker:
Oh hrm, I had an explanation in the commit comment, not sure how it got lost.  But was that `overload[0][0]` was one level too deep; since `overload[0][0]` actually referred to the first parameter of the first signature, `Array.isArray(overload[0][0])` was always true, so we checked both the 1-ary and 2-ary versions of of `-` even when there were two arguments.

Issue #5481, opened 2017-10-18
Can't access feature_filter createFilter as stand-alone function anymore


lamuertepeluda:
## Motivation

I was using the wonderful filter library also outside the map context. Not so long ago[ it was a standalone repository](https://github.com/mapbox/feature-filter) that was recently merged into this repository. 

However, **before** mapbox 0.41

`import createFilter from 'mapbox-gl/src/style-spec/feature_filter';`

According to this [README](https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/feature_filter/README.md) you could then use createFilter to create wonderful GeoJSON property filters that exploit that clever style-spec of yours.

I did it and everything fine! 😃 🌈 

**But since** mapbox **0.41**... 😭 🌧 

```
node_modules/mapbox-gl/src/style-spec/feature_filter/index.js Unexpected token (5:12)
You may need an appropriate loader to handle this file type.
| const {createExpression} = require('../expression');
| 
| import type {GlobalProperties} from '../expression';
| export type FeatureFilter = (globalProperties: GlobalProperties, feature: VectorTileFeature) => boolean;
| 
 @ ./src/js/components/Home/components/mapTab/MapTab.jsx 40:0-57
 @ ./src/js/components/Home/components/mapTab/index.js
 @ ./src/js/components/Home/components/index.js
 @ ./src/js/components/Home/Home.jsx
 @ ./src/js/components/Home/index.js
 @ ./src/js/components/index.js
 @ ./src/js/startup/routesConfig.js
 @ ./src/js/startup/index.js
 @ ./src/main.js
```

In my setup I'm using webpack 3 (1.), and I'm also using React and it's a nightmare configuring/finding a decent flowtype plugin (2.) and compile mapbox sources.
1. I know, I know, you like more browserify than webpack and you won't support it, it's fine... 
2. I know, I know, you guys hate Typescript that makes life a bit easier on the configuration side... 
But since I use the compiled mapbox-gl file, however, I bypassed that an I was having no problem, relying on the `mapboxgl` object. in the global namespace, as if I use a CDN. But sadly this does not export the `createFilter` function that I need so much!!! 😭 😭 😭 

## Design/Implementation

Just export the `createFilter` function in the global namespace, such as, for instance, Map, Style etc. in a way that somebody could use `maboxgl.createFilter` to filter his own stuff.

jfirebaugh:
→ #5502

PR #5484, opened 2017-10-19
Export createFilters from style-spec

lamuertepeluda:
## Launch Checklist

Address #5481

It now exports `createFilters` from `style-spec` to the global `mapboxgl object`.

It is now possible to create custom filters for a GeoJSON collection to be used outside the map using the same powerful style-spec filter language.

### Example

```javascript
const filterDef = ['==', 'type', 'bar'];
const filterFunction = mapboxgl.createFilter(filterDef).bind(null, null); // see comments below
const filteredFeature = myGeoJSONData.features.filter(filterFunction); // will hold only features of type bar
```

#### Why bind(null, null)
I have to set thisArg null and the first param null, because now the compiled filter has a first parameter 'g' which is not clear to me, but wasn't present in v0.40.1, but perhaps has something to do with the new expression filters syntax.

## Possible use cases for this API

An application where I want to show or make available for download only the features visible on the map, by filtering the original data source, using the same filter definition, even in a page that does not display the map.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ const Point = require('@mapbox/point-geometry');
 const Evented = require('./util/evented');
 const config = require('./util/config');
 const rtlTextPlugin = require('./source/rtl_text_plugin');
+const createFilter = require('./style-spec/feature_filter');
 
 module.exports = {
     version,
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Thanks for the contribution @lamuertepeluda! However, I think we want to continue to maintain the separation between the published `mapbox-gl` and `mapbox-gl-style-spec` modules. #5502 tracks restoring the working `mapbox-gl-style-spec` functionality.

Issue #2419, opened 2016-04-11
Support dragPan.disable() mid-drag


kristfal:
**mapbox-gl-js version**: Master branch
### Steps to Trigger Behavior
1. Start dragging the map (mouse or touch input)
2. Set `map.dragPan.disable()` during the drag event
### Expected Behavior

Dragging should no longer move the map.
### Actual Behavior

Map will continue being draggable until `mouseup` or `touchend`

Afaik, the root issue is that event listeners can't be removed until the ongoing `mousedown` / `touchstart` event is finished.

Workaround:

Update the following drag_pan.js handlers:

``` js

_onDown: function (e) {
        if (this._ignoreEvent(e)) return;
        if (this.isActive()) return;
        if (!this._enabled) { // Added check for _enabled
            this._inertia = [];
            return;
        };

        if (e.touches) {
            document.addEventListener('touchmove', this._onMove);
            document.addEventListener('touchend', this._onTouchEnd);
        } else {
            document.addEventListener('mousemove', this._onMove);
            document.addEventListener('mouseup', this._onMouseUp);
        }

        this._active = false;
        this._startPos = this._pos = DOM.mousePos(this._el, e);
        this._inertia = [[Date.now(), this._pos]];
    },

    _onMove: function (e) {
        if (this._ignoreEvent(e)) return;
         if (!this._enabled) { // Added check for _enabled
            this._inertia = [];
            return;
        };

        if (!this.isActive()) {
            this._active = true;
            this._fireEvent('dragstart', e);
            this._fireEvent('movestart', e);
        }

        var pos = DOM.mousePos(this._el, e),
            map = this._map;

        map.stop();
        this._drainInertiaBuffer();
        this._inertia.push([Date.now(), pos]);

        map.transform.setLocationAtPoint(map.transform.pointLocation(this._pos), pos);

        this._fireEvent('drag', e);
        this._fireEvent('move', e);

        this._pos = pos;

        e.preventDefault();
    },

```

lucaswoj:
What are your thoughts, @bhousel?

kristfal:
The reason for wanting/needing to use `dragPan.disable()` mid-drag is that it is currently not possible to replicate the [draggable marker](https://www.mapbox.com/mapbox-gl-js/example/drag-a-point/) functionality on touch devices.

[...]

PR #5486, opened 2017-10-19
Allow disable() mid-drag (#2419)

krabbypattified:
This code allows the user to call dragPan.disable() mid-drag.
This is intended as a fix for touch (mobile) devices.
It enables draggable markers on touch devices.


mollymerp:
@krabbypattified thanks for your contribution! the tests are failing because of linting errors right now. you can fix this by running `yarn lint --fix` or `npm run lint -- --fix`.

do you have a test case that you used to manually determine if this fix was working? I can investigate adding a unit test to test this fix as well.

krabbypattified:
I've added a test for my code, but since there's [no `simulation.touch` events](https://github.com/mapbox/mapbox-gl-js/blob/master/test/node_modules/mapbox-gl-js-test/simulate_interaction.js), I wasn't able to include touch tests.

I've run `yarn lint --fix` which has come back with no errors. Granted, I did copy/paste from my local Atom files into the GitHub online commit system.

krabbypattified:
Why is Flow doing this to me?
` if (e.touches) return this._onTouchEnd(e);
                                                        ^ MouseEvent. This type is incompatible with the expected param type of
195:     _onTouchEnd(e: TouchEvent)`

mollymerp:
Ooo @krabbypattified sorry for the delay here. What's happening is that since you're not explicitly ensuring that `e` is of type `TouchEvent` before invoking `_onTouchEnd` flow thinks that the `e` passed to `_onTouchEnd` could possibly be of type `MouseEvent.`

I belive you can get rid of this error by changing your condition on L111 from `if (e.touches)` to `if (e instanceof TouchEvent)`

jfirebaugh:
Thanks for the contribution @krabbypattified, and sorry that we weren't more responsive on this PR. It got overtaken by other event-related work, so I've opened an alternative fix for the issue in #6232.

PR #5498, opened 2017-10-22
Using transform instead of box-shadow for user location dot pulse animation

jasonbarry:
In Safari 11.0, I noticed that the [Locate the user example](https://www.mapbox.com/mapbox-gl-js/example/locate-user/) looked like this: 

old

I modified the CSS to use transforms instead of box-shadow for the pulse animation, which not only fixes the square-flash issue, but is also more performant because it's hardware-accelerated. 

new17

- Changed the dot diameter from 16px to ~~17px~~ 15px — odd diameter widths look better when centering
- Used shorthand for `animation` properties to eliminate the need for `animation-iteration-count` properties

### DIFF
--- a/dist/mapbox-gl.css
+++ b/dist/mapbox-gl.css
@@ -354,16 +354,16 @@ a.mapboxgl-ctrl-logo {
 
 .mapboxgl-user-location-dot {
     background-color: #1da1f2;
-    width: 17px;
-    height: 17px;
+    width: 15px;
+    height: 15px;
     border-radius: 50%;
     box-shadow: 0 0 2px rgba(0, 0, 0, 0.25);
 }
 .mapboxgl-user-location-dot::before {
     background-color: #1da1f2;
     content: '';
-    width: 17px;
-    height: 17px;
+    width: 15px;
+    height: 15px;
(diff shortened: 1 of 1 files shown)


jasonbarry:
@andrewharvey Thanks for the feedback! Fixed that issue and made some minor adjustments, see here: https://jsbin.com/qudibixica/edit?html,output

Tested in latest Safari, Firefox, and Opera on macOS. Would definitely appreciate more cross-browser testing.

andrewharvey:
@jasonbarry Looks better now in Chrome. I tested on IE11, it looks like it's off centre a bit (inside dot vs the animated halo). It's not a big issue, but does look like a step back compared to how it currently looks in IE11.

mollymerp:
Thanks @andrewharvey – I just tried to test in IE using browserstack but I ran into the same issue I was having in October with the Geolocation plugin not working.

andrewharvey:
@mollymerp I think browserstack is a bit hit and miss, but I did managed to see it looking good in both browserstack Windows 7 IE11 and on a real Windows 7 IE11, along with ios safari 11 and a bunch of latest android and desktop browsers.

PR #5507, opened 2017-10-23
Individual layer benchmarks

lbud:
Adds individual benchmarks for all layer types _except heatmaps_, because my heatmaps bench implementation kept pausing as it was about to OOM (cc @mourner if you're up for adding a simple heatmap benchmark). Also adds rejection handlers to these and other benchmarks that invoke `createMap()` to catch errors more easily in the future.


mourner:
> my heatmaps bench implementation kept pausing as it was about to OOM

Oh no, that doesn't sound good... Do you experience any crashes / slowness when using heatmaps in the real world? E.g. https://www.mapbox.com/mapbox-gl-js/example/heatmap-layer/

jfirebaugh:
👍  what @mourner said -- the `bench` method should benchmark rendering, and the map creation should be done in `setup` (see the `Paint` benchmark for example). For convenience, you may want to put all the benchmarks in a single file with a base class for common setup ([example](https://github.com/mapbox/mapbox-gl-js/blob/c5554ff48335a9afdaba0453ef503618683146e9/bench/benchmarks/expressions.js)).

bench/benchmarks/map_load.js
```diff
@@ -10,6 +10,9 @@ module.exports = class MapLoad extends Benchmark {
                 sources: {},
                 layers: []
             }
-        }).then(map => map.remove());
+        }).then(map => map.remove())
+            .catch(error => {
```
jfirebaugh: Let's put the error handling in the `Benchmark` base class, wherever any of the overridable methods (`setup`, `bench`, and `teardown`) are called. That way it's consistent for all benchmarks.

PR #5508, opened 2017-10-24
Port docs to Batfish

jfirebaugh:
Replace our Jekyll+liquid+handlebars+ejs+jQuery+vanillajs conglomerate with Batfish pages.

Huge thanks 🙏  to @davidtheclark for continuous guidance.

TODO:

* [x] Rebase, resolve conflicts
* [x] Remove `convert-examples.js`
* [x] Check that page weight is acceptable
* [x] Change the way GL JS gets its page shell updated, so it receives the JS files instead of the Jekyll template
* [x] Test on staging


davidtheclark:
Another item for the checklist: Change the way GL JS gets its page shell updated, so it receives the JS files instead of the Jekyll template.

If you want a zippy loading indicator at the top of the page when changing routes, like https://www.mapbox.com/android-docs/, here's how you could implement that: https://github.com/mapbox/android-docs/blob/bed73b1a6e4f819f2dedba9751fbbfa619c050aa/src/components/page-shell.js#L19-L27

Let me know if there are other ways I can help!

jfirebaugh:
> On the local site (`yarn start-docs`), hash links like ☝️ do not scroll to the correct place on the page.

My guess is that this is related to the flash of unstyled content that you can see on the initial page load: the page is rendered without CSS, the browser jumps to where that anchor is currently located, then the CSS is applied and the anchor moves to a different location. This seems like an issue that would need to be addressed in Batfish itself (cc @davidtheclark).

davidtheclark:
@jfirebaugh Yeah, your diagnosis is correct. Things are a little different on the dev server — it's not going through all the work of building HTML pages, inlining CSS, etc. I'm not exactly sure the best way to smooth this out — but it would happen with Batfish.

Issue #1970, opened 2016-01-22
Add a way to set initial map bounds


peckjon:
Is there no way to set the initial bounds of a Map?

[options.maxBounds](https://www.mapbox.com/mapbox-gl-js/api/#Map) allows one to set the initial bounding box, but also permanently constrains the view to this box.

A combination of .center and .zoom would effectively allow one to set initial bounds, but also requires some computation, and is inexact.

Is there an easy way to set an initial options.bounds ?

ravraponi:
+1

lucaswoj:
You are correct. That feature does not exist but should :+1:

ryanbaumann:
Currently I'm suing the 'geojsonExtent' from the mapbox.js library to facilitate zooming to bounds.  Agreed it would be great to have as a standard feature in the API.

```javascript
<script src='https://api.mapbox.com/mapbox.js/plugins/geojson-extent/v0.0.1/geojson-extent.js'></script>
function fit() {
    //fit gl map to a geojson file bounds
    try {
        map.fitBounds(geojsonExtent( geojson_layer_id ));
    }
    catch (err) {
        console.log(err);
    }
}
```

averas:
@ryanbaumann I think there's a misunderstand there. I believe what @peckjon is asking for is not a method to to zoom to a bounding box (Map#fitBounds) per se, because this is actually already part of the library. The problem is that the current function is animated so you can't instantly change the center/zoom of a map like you can with Map#jumpTo, using a bounding box. What this ticket asks for is a fitBounds without a transition (and possibly also as input when instantiating the map).

Also, getting the bounding box for a geometry I would argue is out of scope for the library, but can easily done with for example turf.js (http://turfjs.org/static/docs/module-turf_extent.html).

jfirebaugh:
Unanimated `Map#fitBounds` is #1473. This issue is about bounds-based map initialization.

averas:
@jfirebaugh Gotcha, but I still got the impression that @ryanbaumann was asking for functionality in the API to actually calculate the bounds for you based on a given geometry, i.e. something like "<code>Map#fitGeojson(geojson)</code>"

ryanbaumann:
@averas I was off-topic on my comment. Apologies.

My intent was to communicate that from a user perspective, I want to initialize a map to fit the bounds of a data layer.  Zooming to bounds after the data layer is added to the map is acceptable.

anx-ckreuzberger:
[...]

PR #5518, opened 2017-10-25
add initial bounds as map constructor option #1970

stepankuzmin:
This PR will add initial bounds as map constructor option according to #1970.

I have a problem implementing the test for this case. Then I manually test new functionality using the debug page, I'm getting the expected result. But the same steps reproduced in test somehow outcomes with different zoom level in the result (`1.113` instead of `2.469`). What can cause this?

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -196,6 +196,7 @@ const defaultOptions = {
  * @param {number} [options.zoom=0] The initial zoom level of the map. If `zoom` is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
  * @param {number} [options.bearing=0] The initial bearing (rotation) of the map, measured in degrees counter-clockwise from north. If `bearing` is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
  * @param {number} [options.pitch=0] The initial pitch (tilt) of the map, measured in degrees away from the plane of the screen (0-60). If `pitch` is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
+ * @param {LngLatBoundsLike} [options.bounds] The initial bounds of the map. If `bounds` is specified, it overrides `center` and `zoom` constructor options.
  * @param {boolean} [options.renderWorldCopies=true]  If `true`, multiple copies of the world will be rendered, when zoomed out.
  * @param {number} [options.maxTileCacheSize=null]  The maximum number of tiles stored in the tile cache for a given source. If omitted, the cache will be dynamically sized based on the current viewport.
  * @param {string} [options.localIdeographFontFamily=null] If specified, defines a CSS font-family
@@ -369,12 +370,17 @@ class Map extends Camera {
         this._hash = options.hash && (new Hash()).addTo(this);
         // don't set position from options if set through hash
         if (!this._hash || !this._hash._onHashChange()) {
-            this.jumpTo({
-                center: options.center,
-                zoom: options.zoom,
-                bearing: options.bearing,
-                pitch: options.pitch
-            });
+            if (options.bounds) {
+                this.resize();
+                this.fitBounds(options.bounds, { duration: 0 });
+            } else {
+                this.jumpTo({
+                    center: options.center,
+                    zoom: options.zoom,
+                    bearing: options.bearing,
+                    pitch: options.pitch
+                });
+            }
         }
 
         this.resize();
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -367,12 +367,11 @@ class Map extends Camera {
 
         bindHandlers(this, options);
 
-        this.resize();
-
         this._hash = options.hash && (new Hash()).addTo(this);
         // don't set position from options if set through hash
         if (!this._hash || !this._hash._onHashChange()) {
             if (options.bounds) {
+                this.resize();
                 this.fitBounds(options.bounds, { duration: 0 });
             } else {
                 this.jumpTo({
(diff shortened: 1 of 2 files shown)


src/ui/map.js
```diff
@@ -324,6 +325,22 @@ class Map extends Camera {
...
+                    dimensions = this._containerDimensions(),
+                    width = dimensions[0],
+                    height = dimensions[1],
+                    scaleX = width / size.x,
+                    scaleY = height / size.y;
+
+                options.center = tr.unproject(nw.add(se).div(2));
+                options.zoom = Math.min(tr.scaleZoom(tr.scale * Math.min(scaleX, scaleY)), options.maxZoom);
+            }
+
```
mourner: Can we reuse the code from `fitBounds` somehow and avoid copying it?

test/unit/ui/map.test.js
```diff
@@ -77,6 +77,16 @@ test('Map', (t) => {
...
+    t.test('initial bounds in constructor options', (t) => {
+        const bounds = [[-133, 16], [-68, 50]];
+        const map = createMap({bounds});
+
+        t.deepEqual(fixedLngLat(map.getCenter(), 4), { lng: -100.5, lat: 34.7171 });
+        t.equal(fixedNum(map.getZoom(), 3), 2.469);
+
+        t.end();
+    });
+
```
mourner: > But the same steps reproduced in test somehow outcomes with different zoom level

The zoom level depends on the size of the map container.

src/ui/map.js
stepankuzmin: Actually, we can try to use `fitBounds(bounds, { duration: 0 })` instead of manually reproducing `fitBounds` logic, but we have to resize the map before.

```js
this.resize();
this.fitBounds(options.bounds, { duration: 0 });
```

Unfortunately, the problem with the test is still there.

test/unit/ui/map.test.js
stepankuzmin: I took https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/ui/camera.test.js#L1355 as an example. And it works fine with the zoom level.

src/ui/map.js
stepankuzmin: I have updated the PR using `fitBounds`

stepankuzmin:
Oh, well, It seems that I mixed up rebasing. Should I create separate PR on top of the current master branch?

mourner:
@stepankuzmin I think you can reset the PR branch to the master, apply your commit on top and force-push — no need to create a new PR.

stepankuzmin:
Thanks for help, @mourner! It seems that tests are passing now.

src/ui/map.js
```diff
@@ -368,12 +369,17 @@ class Map extends Camera {
...
         // don't set position from options if set through hash
         if (!this._hash || !this._hash._onHashChange()) {
-            this.jumpTo({
-                center: options.center,
-                zoom: options.zoom,
-                bearing: options.bearing,
-                pitch: options.pitch
-            });
+            if (options.bounds) {
+                this.resize();
```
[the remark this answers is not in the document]
stepankuzmin: Without resize, Transform's `pixelMatrixInverse` is undefined here: https://github.com/mapbox/mapbox-gl-js/blob/11d09c7d3f19377d05755b7bcef4ad30ee4d07fd/src/geo/transform.js#L372

Maybe we can move following resize call before condition? https://github.com/mapbox/mapbox-gl-js/blob/11d09c7d3f19377d05755b7bcef4ad30ee4d07fd/src/ui/map.js#L379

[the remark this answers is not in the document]
mourner: @stepankuzmin yes, let's do that — seems to be more logical.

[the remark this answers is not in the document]
stepankuzmin: Done!

mourner:
Hmm, now the query tests fail for some reason :(

stepankuzmin:
That's strange, because `test-query` fails on `upstream/master` too.

mourner:
@stepankuzmin maybe let's rebase on top of the current master and revisit this?

stepankuzmin:
I've rebased on top of the current master branch, but the tests are still failing. And the problem is that I can't reproduce it locally — all 14 376 tests are passing.

mourner:
@stepankuzmin the unit tests are passing; it's the query and render tests that fail. I can reproduce failures locally, e.g. by running  `./build/run-node test/render.test.js 5911`. I guess moving `resize` before the `jumpTo` somehow messes up map initialization in certain circumstances.

stepankuzmin:
Hey @mourner and @ryanhamley! Sorry for the late reply, I've updated the PR according to the last comment — moved resize back to conditional and set expected zoom value in the test to `2.113`.

GeovationMax:
@mourner Just stumbled upon this new feature which would be quite useful to us. When is the next release planned for this to be included as it's not in the current latest 0.50.0?

Thank you!

ryanhamley:
@GeovationMax This will be included in v0.51.0. The beta will be released next week and the final version will be released the following week, around November 7th.

SoLoGHoST:
I am using the latest version js and css v0.51.0.  Can someone please help me understand how to initialize the map with bounds?  Doesn't work for me and I get no errors.  It just doesn't seem to work.

andrewharvey:
See https://www.mapbox.com/mapbox-gl-js/api/#map there is a map option called `bounds` which accepts a [`LngLatBoundsLike`](https://www.mapbox.com/mapbox-gl-js/api/#lnglatboundslike)

```
var map = new mapboxgl.Map({
    container: 'map',
    bounds: [left, bottom, right, top]
});
```

SoLoGHoST:
So, I'm misunderstood.  How can I set a bunch of LngLat markers than have the map show with them centered in there?

andrewharvey:
@SoLoGHoST Better to ask that at https://stackoverflow.com/tags/mapbox-gl-js

SoLoGHoST:
Ok thank you, I have created a question here:  https://stackoverflow.com/questions/53330202/set-map-bounds-based-on-multiple-marker-lng-lat

PR #5520, opened 2017-10-25
Fixing padding issues in fitBounds

### LINKED ISSUE #4846: fitBounds padding is not zooming correctly
Using fitBounds with padding, I wanted to increase the padding on the left. I noticed that it seemed way off though for larger padding values.

http://jsbin.com/nadukifexa/edit?html,output

In the example, I drew a 400px wide box, then set the padding on the left to 420 (400 + 20). You would expect the fit map to be just off of the right side of the box. Instead it is way off. To achieve the expected result, I had to use half and set it to 220 (200 + 20).

GunnarSturla:
This fixes #4846 where the offset from center was incorrectly calculated when there was a difference between top and bottom, or left and right padding. The offset should be half the difference between padding when finding the center.

The example from the original issue showing the error:
http://jsbin.com/nadukifexa/edit?html,output

Same example with this fix:
http://jsbin.com/rihunorico/edit?html,output

## Launch Checklist

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -395,7 +395,7 @@ class Camera extends Evented {
         // (lateral and vertical padding), and the part that does (paddingOffset). We add the padding offset
         // to the options `offset` object where it can alter the map's center in the subsequent calls to
         // `easeTo` and `flyTo`.
-        const paddingOffset = [options.padding.left - options.padding.right, options.padding.top - options.padding.bottom],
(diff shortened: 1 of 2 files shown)


GunnarSturla:
All tests are passing 👍

mollymerp:
Thanks again for the contribution @GunnarSturla !

GunnarSturla:
Woohoo! Thanks!

PR #5534, opened 2017-10-26
Fix packaging issues with style-spec package

### LINKED ISSUE #5502: Fix bugs preventing mapbox-gl-style-spec package on NPM from being used externally
Current issues with the mapbox-gl-style-spec package:

- Flow types in `expression/*.js` prevent the package from being used directly: `require('@mapbox/mapbox-gl-style-spec')` fails with syntax errors
- The absence of a `browserify: { transform: ["unflowify", "bubleify", "./build/strictify"] }` setting in `package.json` prevents browserify from being able to bundle a program that uses `@mapbox/mapbox-gl-style-spec` as a dependency (except maybe if that build used the "global transform" antipattern).
- Published npm package does not include `flow-typed` directory from the root repo.  I haven't actually checked whether any types declared there are used within `src/style-spec`, but even if they aren't now, it is likely that they will be in the future.
- (from #5522) Usage of `let` and `const` without `"use strict";` is forbidden in Node v4.0.  More broadly: as of now, the style-spec package needs to be usable in Node v4.0 and IE 11, both of which have limited ES6 support.

After resolving the above issues, we should also add a build test that simulates installing @mapbox/mapbox-gl-style-spec outside of the gl-js repo and attempts to (a) require it in node, (b) bundle it.

anandthakker:
Fixes #5502

The style-spec's prepublish step now:
- Compiles the style-spec module to `dist/index.js`. Note that this is still a *Node* module: it does not include external dependencies or browser implementations of built-in node modules.
- Copies `flow-typed/` from the root mapbox-gl-js repo. As @davidtheclark mentions in https://github.com/mapbox/mapbox-gl-js/issues/5502#issuecomment-338691744, this may not be sufficient to make flow types work out of the box, but it at least ensures that the necessary type definitions are distributed with the style-spec package.

The majority of users should simply be able to do `require("@mapbox/mapbox-gl-style-spec")`, and it should work fine both in Node and standard browser bundles that understand commonjs.

But it's also possible to require individual modules like `require("@mapbox/mapbox-gl-style-spec/expression")`, so long as the client either (a) uses `flow-remove-types` and Node >= 6.0, or (b) respects the transform configuration in the `browserify` field of package.json.

TODO:
 - [ ] Publish a prerelease/test version to npm and manually test a fresh install.


anandthakker:
cc @davidtheclark

anandthakker:
> Publish a prerelease/test version to npm and manually test a fresh install

Update: not ready to merge yet, b/c somehow the package isn't including `dist/index.js` -- even when I add a blank `dist/.npmignore` to ensure `npm` isn't using `dist/.gitignore`.

PR #5537, opened 2017-10-27
update example in geojson layer in stack html

VictorHom:
## Launch Checklist

 - [X] briefly describe the changes in this PR - updated example in a geojson layer in stack html

### DIFF
--- a/docs/pages/example/geojson-layer-in-stack.html
+++ b/docs/pages/example/geojson-layer-in-stack.html
@@ -11,10 +11,9 @@ map.on('load', function () {
 
     var layers = map.getStyle().layers;
     // Find the index of the first symbol layer in the map style
-    var firstSymbolId = undefined;
     for (var i = 0; i < layers.length; i++) {
         if (layers[i].type === 'symbol') {
-            firstSymbolId = layers[i].id;
+            var firstSymbolId = layers[i].id;
             break;
         }
     }
--- a/docs/pages/example/geojson-layer-in-stack.html
+++ b/docs/pages/example/geojson-layer-in-stack.html
@@ -9,6 +9,13 @@ var map = new mapboxgl.Map({
 
 map.on('load', function () {
 
+    // Find the index of the bottom-most type: symbol layer in the map style
+    // Insert the layer beneath any symbol layer.
+    var layers = map.getStyle().layers.reverse();
+    var labelLayerIdx = layers.findIndex(function (layer) {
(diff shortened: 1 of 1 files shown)


docs/pages/example/geojson-layer-in-stack.html
```diff
@@ -9,6 +9,13 @@
 
 map.on('load', function () {
 
+    // Find the index of the bottom-most type: symbol layer in the map style
```
mollymerp: > // Find the index of the bottom-most type: symbol layer in the map style
👇 
> // Find the index of the bottom-most symbol layer in the map style

(sorry I know this was my wording before but I forgot that you can't do code highlighting in js comments so this looks kind of confusing in plain text)

mollymerp: also this comment should be right before line 15 (the line where we're actually finding the symbol layer)

docs/pages/example/geojson-layer-in-stack.html
```diff
@@ -9,6 +9,13 @@
...
+    // Insert the layer beneath any symbol layer.
```
mollymerp: this comment should be included after the comment on L31-36. Something like "in this case, we are inserting the new layer immediately before the first symbol layer in the style"

docs/pages/example/geojson-layer-in-stack.html
```diff
@@ -9,6 +9,13 @@
...
+    var layers = map.getStyle().layers.reverse();
+    var labelLayerIdx = layers.findIndex(function (layer) {
```
jfirebaugh: [`findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Browser_compatibility) isn't available in IE, and don't you want to insert the layer before the first symbol layer, rather than before the last non-symbol layer?

```js
var before = undefined;
for (var i = 0; i < layers.length; i++) {
    if (layers[i].type === symbol) {
        before = layers[i].id;
        break;
    }
}
```

mollymerp: 🤦‍♀️ good catch(es) @jfirebaugh!

VictorHom:
The pr is updated to find the first symbol id and insert that instead @mollymerp :)

mollymerp:
Thanks again for your contribution @VictorHom!

PR #5541, opened 2017-10-27
Remove gl.lineWidth

lbud:
This PR removes the use of `gl.lineWidth` in the few places where we use it: the [MDN doc for `WebGLRenderingContext.lineWidth()`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/lineWidth) states

> `gl.lineWidth()` has no effect in common modern browsers. The corresponding [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=60124) and [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=634506) are closed as Won't Fix and the WebGL specification now defines that `gl.lineWidth()` does not change the line width anymore. If you need lines of any width other than 1px, consider drawing a narrow strip of triangles.

### DIFF
--- a/src/gl/context.js
+++ b/src/gl/context.js
@@ -7,7 +7,7 @@ import DepthMode from './depth_mode';
 import StencilMode from './stencil_mode';
 import ColorMode from './color_mode';
 import { deepEqual } from '../util/util';
-import { ClearColor, ClearDepth, ClearStencil, ColorMask, DepthMask, StencilMask, StencilFunc, StencilOp, StencilTest, DepthRange, DepthTest, DepthFunc, Blend, BlendFunc, BlendColor, Program, LineWidth, ActiveTextureUnit, Viewport, BindFramebuffer, BindRenderbuffer, BindTexture, BindVertexBuffer, BindElementBuffer, BindVertexArrayOES, PixelStoreUnpack, PixelStoreUnpackPremultiplyAlpha } from './value';
+import { ClearColor, ClearDepth, ClearStencil, ColorMask, DepthMask, StencilMask, StencilFunc, StencilOp, StencilTest, DepthRange, DepthTest, DepthFunc, Blend, BlendFunc, BlendColor, Program, ActiveTextureUnit, Viewport, BindFramebuffer, BindRenderbuffer, BindTexture, BindVertexBuffer, BindElementBuffer, BindVertexArrayOES, PixelStoreUnpack, PixelStoreUnpackPremultiplyAlpha } from './value';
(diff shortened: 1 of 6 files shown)


jfirebaugh:
Refs #2080. The only reason I'm hesitant to approve this is http://webglstats.com/webgl/parameter/ALIASED_LINE_WIDTH_RANGE:

[image](http://webglstats.com/webgl/parameter/ALIASED_LINE_WIDTH_RANGE)

I.e., something happened recently that appears to have increased >1 line width support to 70%+ (although declining since then).

lbud:
It sounds like [WebGL Stats](http://webglstats.com/) gets its data by embedding a script in websites that have chosen to participate (specifically [these](http://webglstats.com/contributors)). Given that MDN's [`lineWidth` compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/lineWidth) shows only IE12 as supporting this, I have to wonder if that chart only indicates that in mid-2017 a new site with a lot of IE users opted in to sharing stats. Given the caveats — 

> The webgl spec, based on the OpenGL ES 2.0/3.0 specs points out that the minimum and maximum width for a line is implementation defined. The maximum minimum width is allowed to be 1.0. The minimum maximum width is also allowed to be 1.0. Because of these implementation defined limits it is not recommended to use line widths other than 1.0 since there is no guarantee any user's browser will display any other width.
> 
> As of January 2017 most implementations of WebGL only support a minimum of 1 and a maximum of 1 as the technology they are based on has these same limits.

([source: MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/lineWidth))

> Professional applications should not use gl.lineWidth. First of all, there is zero guarantee that wide lines are supported on all WebGL implementations. So you must have a fallback anyway. Secondly, wide lines can be implemented by the driver in a variety of ways that are spec compliant (or not). To make sure they look the same across devices, polygons should be used instead. Thirdly, it is likely that a driver implements wide lines using polygons anyway. So there's no performance advantage. It has been deprecated from core desktop OpenGL for these reasons.

(source: a Swiftshader maintainer's [comment](https://bugs.chromium.org/p/chromium/issues/detail?id=60124#c20) on the Chromium issue, marked Wontfix)


Given what we know (and the fact that rendering antialiased polygons with different line widths based on whether or not they're on IE12 seems like a suboptimal experience) I think we may as well merge this — @jfirebaugh what do you think? I'll rebase + merge if you agree.

PR #5542, opened 2017-10-27
Split "curve" into "step" and "interpolate"

### LINKED ISSUE #5528: Split "curve" expression into "step" and "interpolate" expressions
In retrospect, `"curve"` is overgeneralized:

* `["curve", input, ["step"], ...]` and `["curve", input, anything else, ...]` have different stop signatures -- step has an extra output value. This makes it awkward for them to share the same expression name.
* `step` is universally supported by paint and layout properties, while other interpolation types are supported only by certain properties. This is something we need to call out in documentation, and it would be simpler to be able to say "Supports `"interpolate"` expressions" rather than "Supports all `curve` interpolation types"/"Supports only `step` interpolation".

Instead, there should be two expressions:

* `["step", input, stop_output_0: OutputType,
    stop_input_1: number, stop_output_1: OutputType,
    stop_input_n: number, stop_output_n: OutputType, ... ]`
* `["interpolate", interpolation, input, 
    stop_input_1: number, stop_output_1: OutputType,
    stop_input_n: number, stop_output_n: OutputType, ...]`

Note that I've also swapped the position of the `interpolation` and `input` arguments for `interpolate`. I think `["interpolate", ["exponential", base], input, ...]` reads better than `["interpolate", input, ["exponential", base], ...]`.

This would be a breaking change, so it's best to get it made now, before adoption is widespread. The following will need to be updated:

* GL JS implementation
* GL JS docs
* Expression blog post
* Expression help guide
* Styles

jfirebaugh:
Fixes #5528. Will make sure all the right people get a heads up before this is merged.

### DIFF
--- a/bench/benchmarks/layers.js
+++ b/bench/benchmarks/layers.js
@@ -115,7 +115,7 @@ class LayerHeatmap extends LayerBenchmark {
                     },
                     "heatmap-intensity": 0.9,
                     "heatmap-color": [
-                        "curve",
+                        "interpolate",
                         ["linear"],
                         ["heatmap-density"],
                         0, "rgba(0, 0, 255, 0)",
--- a/debug/circles.html
+++ b/debug/circles.html
@@ -42,7 +42,7 @@ map.on('load', function() {
         "source": "circles",
         "paint": {
             "circle-radius": [
-                "curve",
+                "interpolate",
                 ["exponential", 2.0],
                 ["zoom"],
                 0, 5,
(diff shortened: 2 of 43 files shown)


jfirebaugh:
Blocked on posting benchmark results by https://github.com/mapbox/mapbox-gl-js/issues/5543.

mourner:
@anandthakker I'd say no, better break hard early. Alias would be good to help with the transition, but there's not enough adoption of expressions yet to delay the break.

jfirebaugh:
I kept it an error, but with a custom error message indicating the new operators.

jfirebaugh:
https://bl.ocks.org/anonymous/raw/0faddf4a07f92862b32b0d2dfeff39bf/ (no significant changes)

PR #5559, opened 2017-10-30
don't call `_render()` directly, fix #5543

### LINKED ISSUE #5543: Error in Paint benchmark: Uncaught TypeError: Cannot read property 'lineAtlas' of undefined
Run `yarn start`, open http://localhost:9966/bench?no-master#Paint, and let it run. Open the console:

```
Uncaught TypeError: Cannot read property 'lineAtlas' of undefined
    at Painter.render (painter.js:276)
    at Map._render (map.js:1460)
```

Bisect says the first bad commit could be any of:
135561825addae07b253bab2167cee72f00f3ac9
3e51b6cf588e47284d9e714477d72a2d476355eb

cc @ChrisLoer @ansis

ansis:
`_rerender()` requests a deduped animation frame. `_render()` does the actual rendering. Using the non-deduped version resulted in overwriting the `_frameId` which prevented the frames from being cancelled.

I'm not sure *exactly* how the placement changes made this bug visible, but I'm guessing some extra rerendering made it more likely.

### DIFF
--- a/bench/benchmarks/paint.js
+++ b/bench/benchmarks/paint.js
@@ -23,7 +23,7 @@ module.exports = class Paint extends Benchmark {
         for (const map of this.maps) {
             map._styleDirty = true;
             map._sourcesDirty = true;
-            map._render();
+            map._rerender();
         }
     }
 


ansis:
Oh, I guess https://github.com/mapbox/mapbox-gl-js/commit/302180335550511773092ccd5ccd17ca63877029 is the right fix?

Issue #3879, opened 2017-01-02
Make error handling in style consistent


jfirebaugh:
Erroneous uses of the runtime styling API in some cases propagate errors by throwing (e.g. [here](https://github.com/mapbox/mapbox-gl-js/blob/fb173257e734cffa04f6636bc03b3798bdcd8a4f/js/style/style.js#L343-L343), [here](https://github.com/mapbox/mapbox-gl-js/blob/fb173257e734cffa04f6636bc03b3798bdcd8a4f/js/style/style.js#L364-L364), [here](https://github.com/mapbox/mapbox-gl-js/blob/fb173257e734cffa04f6636bc03b3798bdcd8a4f/js/style/style.js#L368-L368), [here](https://github.com/mapbox/mapbox-gl-js/blob/fb173257e734cffa04f6636bc03b3798bdcd8a4f/js/style/style.js#L396-L396)) and in other cases by emitting an `error` event (e.g. [here](https://github.com/mapbox/mapbox-gl-js/blob/fb173257e734cffa04f6636bc03b3798bdcd8a4f/js/style/style.js#L681-L681), [here](https://github.com/mapbox/mapbox-gl-js/blob/fb173257e734cffa04f6636bc03b3798bdcd8a4f/js/style/style.js#L134-L134), [here](https://github.com/mapbox/mapbox-gl-js/blob/3e56c777ca384e8443c2af889630b9a8fc0664bb/js/style/validate_style.js#L8-L8)).

We have [previously agreed](https://github.com/mapbox/mapbox-gl-js/issues/3208) that all errors should be propagated via the `error` event, so should convert the cases that throw.

PR #5562, opened 2017-10-30
Throw an error when a used source is removed and not readded

### LINKED ISSUE #1711: Display an error message when a used source is removed and not re-added
Code: http://jsfiddle.net/rwccfmej/

cc/ @lucaswoj

lucaswoj:
Currently, the below code does not throw an error. Instead it quietly fails to render `mapbox-layer`.

```js
map.setStyle({
  'version': 8,
  'sources': {
    'mapbox-source': {...}
  },
  'layers': [{
    'id': 'mapbox-layer',
    'source': 'mapbox-source',
    'type': 'circle'
  }]
});

map.removeSource('mapbox-source');
// THROWS
```

This PR causes an error to be fired in the above case while still allowing the below case where a source is removed and then re-added before the next call to `Style#update`

```js
map.removeSource('mapbox-source');
map.addSource('mapbox-source', {...});
// DOES NOT THROW
```

fixes #1711 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -510,7 +510,7 @@ class Style extends Evented {
         }
         for (const layerId in this._layers) {
             if (this._layers[layerId].source === id) {
-                throw new Error(`Source "${id}" cannot be removed while layer "${layerId}" is using it.`);
+                return this.fire('error', `Source "${id}" cannot be removed while layer "${layerId}" is using it.`);
             }
         }
 
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -89,7 +89,6 @@ class Style extends Evented {
     _rtlTextPluginCallback: Function;
     _changed: boolean;
     _updatedSources: {[string]: 'clear' | 'reload'};
-    _removedUsedSources: {[string]: boolean};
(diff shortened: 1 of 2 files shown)


lucaswoj:
@jfirebaugh I simplified per your feedback. I prefer this way but it may cause headaches for some users relying on the edge case. What do you think about using a warning & deprecation notice rather than an error?

jfirebaugh:
Signal the error via `this.fire('error', ...)` rather than an exception -- that's the [intent](https://github.com/mapbox/mapbox-gl-js/issues/3879) for all style errors, and will automatically make the error behavior softer. It'll be a console error rather than stack unwinding.

Secondarily, I favor making the change now to return immediately in the error case (i.e. not removing the layer), but if you want to delay making that change for a release or two that's ok too. Let's leave the issue open if you do though.

lucaswoj:
✅ I made the suggested changes to the error handling flow. We'll should come back to this file and get rid of existing uses of `throw` with some kind of migration strategy for users who depend on catching those.

Issue #2920, opened 2016-07-29
Document plugin API for sources


mourner:
Spin-off from #929. Now that it's implemented in #2667, let's make sure it's well documented and with examples.

anandthakker:
Just a note for anyone tracking this: a fair amount of the custom source type API is documented under the hood, just [marked private for now](https://github.com/mapbox/mapbox-gl-js/pull/2667/commits/c991549515d688919d7362feaf8e5a2f953dd98b): 

> We should validate the custom source architecture and build some demo sources before documenting the internals publicly.

So, with that in mind, some specific steps before going public w/ this API:
 - [ ] Fixup [this topojson source example](https://github.com/developmentseed/mapbox-gl-topojson), including a nice readme.  This is an example of a fairly minimal custom source type that leverages much of the existing geojson source.
 - [ ] Complete the "dynamic vector source" type (proof of concept [here](https://github.com/anandthakker/mapbox-gl-js/tree/dynamic-data-source), relies on #2812 )
 - [ ] Actually use ^ in production (Dev Seed has a project that will definitely be doing this)
 - [ ] Wait X weeks

@lucaswoj @mourner any additional suggestions, or thoughts on the right value of X?

lucaswoj:
That plan looks good to me @anandthakker 😄 

X could be 0 weeks. Building the "dynamic vector source" is what I'm after.

pwilczynski:
We also are interested in exploring some of the custom source ideas for our map but we're worried about the instability of the existing API. Do you have a sense of how "beta" it is at this point? @anandthakker, how did it work for your project?

anandthakker:
@pwilczynski it's working well in our project, but I do think that there are very likely changes on the horizon, esp relating to some architectural changes in how the WebWorker code works (tracked here: #3034 ).  I am not sure what the likely timeline is for that, but I've been holding off on the 'dynamic vector source' example that's blocking this ticket until after that change.

lucaswoj:
@pwilczynski There are some big changes on the horizon for this API. I recommend looking at designs that don't require creating custom sources for the time being.

nickpeihl:
[...]

PR #5572, opened 2017-10-31
Expose the actor parent (Worker) on wrapped actor.

deanm:
Worker sources are created with a wrapped actor with a send() that has the last argument bound to the mapId.  The wrapper is just a raw object literal that didn't have any of the normal Actor properties on it.  Add the actor parent to the object so that the worker source has access to it.

### DIFF
--- a/src/source/worker.js
+++ b/src/source/worker.js
@@ -141,6 +141,7 @@ class Worker {
             // use a wrapped actor so that we can attach a target mapId param
             // to any messages invoked by the WorkerSource
             const actor = {
+                parent: this.actor.parent,
                 send: (type, data, callback, buffers) => {
                     this.actor.send(type, data, callback, buffers, mapId);
                 }


deanm:
@anandthakker I know the 3rd party sources API is still under work, so I tried to make the most minimal change.  Maybe it seems a bit strange but having the actor parent gives me access to the Worker which I don't think there would be a way to otherwise.

deanm:
Not sure if I can add reviewers, maybe it's something for @mourner to review?

Thanks!

deanm:
@anandthakker Isn't parent already normally exposed on the Actor interface?

https://github.com/deanm/mapbox-gl-js/blob/25e5f1dd9d55835d9572e15893740aa90aae9eeb/src/util/actor.js#L18

The point here is just that the Actor that gets passed in is not a real Actor, just an object literal with `send`... If you passed it the real Actor instance then accessing `parent` wouldn't be a problem.  So I didn't understand wrapping the Actor as a way to somehow protect against the WorkerSources accessing the Actor, but just so that it could bind the last parameter to send, no?

anandthakker:
@deanm yep, you're right that this was the original intent of the wrapping object, and that `parent` is a _de facto_ part of the Actor interface.  What I'm wondering, however, is whether it really _should_ be part of `Actor`'s public interface, or whether we should treat it as a private member instead.

deanm:
@anandthakker I don't want to give a huge long winded explanation of what I'm trying to do, but basically I'm creating a 3rd party source, so that means adding some functionality on both sides — the main thread and the worker.  There are some things I would like to do inside of the worker, but the global scope basically doesn't have anything on it besides registerWorkerSource.  That is fine if all you want to do is register your source of course, but if I want to access anything else in the mapbox world, I at least need a reference to something.  Just as an example, I might want to communicate with another builtin source (accessible via Worker.getWorkerSource).  That's just an example, but anyway, by getting to the Worker (in what I realize is a bit of a monkey patch way) this solves my problems.  I didn't want to make a big deal out of it and ask for any bigger API changes, etc, so I just looked for the most minimal way to get to something and this seemed like the simplest thing to do.

deanm:
@anandthakker you could also just put the `Worker` instance or some way of getting it on the global scope like `registerWorkerSource` and `registerRTLTextPlugin`.  It sort of makes sense to me to be able to get at the `Worker` from within the worker, but I don't have a specific way I feel like it should be organized, so perhaps maybe you feel this alternative is better.

jfirebaugh:
Hey @deanm -- you're probably aware of this, but this part of the codebase is not a public API and we make no guarantees of API stability from release to release. For now, not making guarantees about what will change and won't change, and not imposing requirements on the code beyond what is needed internally, maximizes our ability to improve performance, implement new features, and fix bugs. I realize this particular change might seem like a small thing, but small things have a tendency to build up over time and in aggregate become a meaningful drag on development velocity. Therefore I think we must regretfully decline this PR, though we appreciate the contribution, and you're of course welcome to continue to use this change locally.

deanm:
@jfirebaugh Okay, I understand, but at the same time it does seem like there is some intention and support for 3rd party sources, including the functionality I mentioned (registering a source on the browser side, and registering a URL to load for the web worker side, and registering from within the worker, etc).  So I understand it's still a work in progress, but some of it is there, you're saying we shouldn't use it at all?

deanm:
@jfirebaugh From the public documentation, it does say:

https://www.mapbox.com/mapbox-gl-js/api/#map#addsource

```
The source type, which must be either one of the core Mapbox GL source types defined in the style specification or a custom type that has been added to the map with Map#addSourceType .
```

So there are some pieces around the documentation suggesting support for custom source types, which is how I came down this road...

jfirebaugh:
@deanm Thanks for pointing that out -- that was an oversight in the docs and I removed the reference. We do eventually intend to offer a public custom source API, but it's not ready yet. Issue #2920 tracks that work.

PR #5575, opened 2017-11-01
Add render tests for expressions

anandthakker:
Generated with a script using `convertFunction` to migrate `style.json` for every render test that has `/function/` in its name.


anandthakker:
> How much does this increase the runtime?  

Surprisingly, it only adds about 14s on my laptop: 2m7s vs 1m53s.

> wait until we cut over to an all-expression implementation, and then remove most of the function tests

If we did this, the function-to-expression conversion path wouldn't get as much test coverage... but maybe the `function.test.js` (perhaps with some additions) will suffice to guard against regressions.

> Is there a reason the conversion generated ["exponential", 1] in some places and ["linear"] in others?

It's because of how we're converting composite functions: https://github.com/mapbox/mapbox-gl-js/blob/e6f3399b02b0bba0c55d224df14a2151686e942f/src/style-spec/function/convert.js#L107-L113

anandthakker:
Ah - my conversion script didn't catch the `-to-*-function` tests -- I'll need to convert those manually.

anandthakker:
☝️ `to-*-function` tests converted too

anandthakker:
Closing in favor of #5584.  Here's the conversion script I hacked together, so that we can use it again later once we're ready to migrate the other function-based tests https://gist.github.com/anandthakker/b0b4e2e42c45cfdb3b748fc1582c37e7

PR #5584, opened 2017-11-02
Add expression versions of runtime styling tests

anandthakker:
Extracting just the runtime-styling tests from #5575


anandthakker:
@jfirebaugh see my comment over https://github.com/mapbox/mapbox-gl-js/pull/5575#issuecomment-341214760 -- if that feels like adding too much, how about just adding these converted runtime-styling tests for now, and saving the full batch for later?

test/integration/render-tests/runtime-styling/layout-property-default-to-property-expression/style.json
```diff
@@ -0,0 +1,78 @@
...
+  "metadata": {
+    "test": {
+      "width": 64,
+      "height": 64,
+      "operations": [
+        [
+          "setLayoutProperty",
+          "symbol",
+          "icon-rotate",
+          [
```
jfirebaugh: Can we improve the conversion so this is just `["get", "rotate"]`?

anandthakker: 👍  sneaking that in here, as well as exporting function-related modules in the main style-spec package (i'd done this locally to get a build of the style spec that i could use in the node conversion script)

PR #5588, opened 2017-11-02
fix #5491 symbol fading across tile reloads

### LINKED ISSUE #5491: Labels and icons fade out and in
**mapbox-gl-js version**:

3e51b6c

### Steps to Trigger Behavior

 1. Call `setFilter` (and open a popup which seems to be irrelevant)

### Expected Behavior

The map remains the same, the popup opens.

### Actual Behavior

All labels and icons in one specific data source fade out with a transition animation and instantly back in.

mapbox-bug

The purple circle and its text as well as the gray labels are in different data source (vector tile) than the blinking elements which are all in the same vector tile.

ansis:
This fixes fading across tile reloads by:
- copying opacities from the previous version of the bucket
- removing the previous version of the bucket from the index

@jfirebaugh @ChrisLoer

### DIFF
--- a/src/symbol/cross_tile_symbol_index.js
+++ b/src/symbol/cross_tile_symbol_index.js
@@ -153,10 +153,18 @@ class CrossTileSymbolLayerIndex {
             }
         }
 
+        const oldTileIndex = this.indexes[coord.z] && this.indexes[coord.z][coord.id];
+        if (oldTileIndex) {
+            // mark labels in the old version of the tile as blocked
+            this.blockLabels(tileIndex, oldTileIndex, true);
+
+            // remove old version of the tile
+            this.removeTile(coord, sourceMaxZoom);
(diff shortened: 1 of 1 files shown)


ansis:
@ChrisLoer do the changes address your concerns?

> A similar change with removeTile would be nice too 

I wasn't sure exactly what you were thinking here so I moved the removeTile call next to where `blockLabels` is called for the tile and commented it.

ChrisLoer:
Yeah, that's a lot easier to read, thanks. 🍏 

You have any idea what's going on with that test failure?

ansis:
I'm not sure about the test failure. I couldn't reproduce it locally and ci passed after rerunning

ChrisLoer:
Yeah, it seemed unrelated to these changes -- I say go ahead and merge.

PR #5589, opened 2017-11-03
Validate expression values for enum-typed properties

### LINKED ISSUE #5576: Expressions for "enum"-typed properties need a post-evaluation validation step
After evaluating an enum-typed property, we need to check that the resulting string is a valid value, and if not issue a warning and fall back to the default value.

anandthakker:
Fixes #5576

Benchmarks: https://bl.ocks.org/anonymous/raw/7a3a7ef8190610345b3c20fdb57abd2c/

(QueryPoint in the above looks like a fluke: https://bl.ocks.org/anonymous/raw/0e24f2c171d51b42356270c3663883b2/)


src/style-spec/expression/index.js
```diff
@@ -148,6 +153,9 @@ function createExpression(expression: mixed,
                 if (val === null || val === undefined) {
                     return defaultValue;
                 }
+                if (enumValues && enumValues.indexOf(val) < 0) {
```
jfirebaugh: Any particular reason to use `indexOf` here rather than `val in values`?

test/unit/style-spec/expression.test.js
```diff
@@ -19,3 +19,26 @@ test('createExpression', (t) => {
 
     t.end();
 });
+
+test('evaluate expression', (t) => {
```
jfirebaugh: Can this be an integration test?

anandthakker: It started out that way, but the expression tests use the non-error-checking `createExpression()` so they can check error messages.  Since this is a post-evaluation check specific to the error-checking path, it's not a great fit for those tests.  

🤔 We could add a _render_ test, though.

anandthakker:
@jfirebaugh added a render test and switched to `val in enumValues`

PR #5594, opened 2017-11-03
Implement queryRadius and queryIntersectsFeature in heatmap layer

### LINKED ISSUE #5473: queryRenderedFeatures throws error if map has layer type 'heatmap'
**mapbox-gl-js version**:

### Steps to Trigger Behavior
 1.   Copy heatmap example
 2.  Create event that triggers queryRenderedFeatures(e.point). I used mousemove
 3.  See Console Log

Plunkr: https://plnkr.co/edit/i8eyFV4ZvHEED16jg4DM?p=preview
Add access token and run.
### Expected Behavior
No error thrown.
### Actual Behavior
Throws uncaught TypeError: e[n].queryRadius is not a function.  While I'm not looking to query features in the actual heatmap layer, I do wish to query other created layers that share the map with the heatmap layer.  If the heatmap layer is 'visible', this error will always be thrown, preventing the results of the function from ever returning.  Setting the heatmap's layer property 'visibility' to 'none' allows queryRenderedFeatures to return as expected.

asheemmamoowala:
Prevent errors when accidentally invoking `queryRenderedFeatures` on heatmap layers. 

Closes #5473


jfirebaugh:
Can you add a regression test for this please? A query test with a heatmap layer that expects no results is fine.

PR #5601, opened 2017-11-06
sort renderable tiles by y and x to match -native

ansis:
This changes the tile sort order to match -native. This is important to make sure the render order of symbols that cross tile boundaries is the same in -native and -js.

This will let us fix `symbol-spacing/line-close` for the viewport collision pr in -native.

`regressions/mapbox-gl-native#7357/expected.png` changes. I think the new rendering is fine, but if we shouldn't change this I can change this so that only symbol rendering order is affected.

## Launch Checklist

 - [x] write tests for all new functionality (built into existing tests and render tests)
 - [x] post benchmark scores https://bl.ocks.org/anonymous/raw/1963f7f03b48d92742d3c17f05ed76d9/

@ChrisLoer


ChrisLoer:
This looks good to me -- my one concern was that `compareKeyZoom` ends up getting called pretty heavily, so the more expensive calculation might make a difference. I ran the "paint" and "layout" benchmarks several times and saw this branch as between 53% and 71% likely to be slower than `master`. I also ran a test on an animated full screen map with the Chrome javascript profiler running and saw `getIds` taking between .4-1% of CPU time on both branches. So.. seems fine.

Issue #5597, opened 2017-11-06
mapbox-gl-js can not render pyramid images


shiweihappy:
hi, when i use mapbox-gl-js to render pyramid image and image not find correct.

0. source url is `https://github.com/shiweihappy/mapbox-gl-js-source`

[source](https://github.com/shiweihappy/mapbox-gl-js-source)

1. i use gdal2tile.py to create pyramid image:

`python gdal2tiles.py -r near -s EPSG:3857 -z "0-15" -w openlayers a4.tif ./a4/`

2. then i put pyramid images to local nginx static source folder

`xxxx/tiles/a4/xxxxx`

3. nginx already set cors and other vector tiles use correctly

4. html render code:

```
var ndvi_source = {
    "type": "raster",
    'tiles': [
        "http://localhost:7777/tiles/a4/{z}/{x}/{y}.png"
    ],
    'tileSize': 256,
    'bounds': [121.849365234, 41.428569989, 121.947212219, 41.4838910427]
};
var ndvi_layer = {
    "id": "ndvi_layer_id",
    "type": "raster",
    "source": "ndvi_source",
    'paint': {}
};
map.addSource('ndvi_source', ndvi_source);
map.addLayer(ndvi_layer);
```

5. then open html i found that, mapbox-gl-js can not find pyramid image, but i use leaflet or openlayer3 is correct. so i do not know why mapbox-gl-js not find correct image x/y/z?

jfirebaugh:
gdal2tiles generates tiles with TMS numbering scheme, so you need to set `"scheme": "tms"` in the source options. Sorry for not having documentation for this property yet -- it'll be documented on https://www.mapbox.com/mapbox-gl-js/style-spec/#sources-raster in the next release.

shiweihappy:
@jfirebaugh , thank you very much!

PR #5604, opened 2017-11-06
Split `source_tile` in schema

jfirebaugh:
In the course of answering #5597, I went to add `scheme` to the documentation. I didn't want it to be documented as an option for vector sources, so I first refactored things to allow that. Then I discovered that the property had in fact already been added. Still, I think it's best to split up `source_tile` anyway. The PR also includes a merge from `mb-pages` to pick up ff1c4c305ef326403b8290bace1cba918863ad8b -- I'll merge this PR on the command line.


jfirebaugh:
This also removes support for validating and migrating v6 styles, and support for validating v7 styles, because otherwise the validation code couldn't handle splitting `source_tile`. Both those versions should be ancient history, but just in case, I left in support for migrating v7 to v8.

PR #5608, opened 2017-11-07
Fix geojson Position element issue (3rd element)

oguiter:
Following RFC 7946, position is an array of number with two OR MORE
elements (Implementations SHOULD NOT extend positions beyond three
elements). The 3rd element is usually Altitude. We could ignore it
but we must not fail.
This small fix allows well-formed geojson file to be fully processed

Signed-off-by: Olivier Guiter <oguiter@free.fr>

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/geo/lng_lat.js
+++ b/src/geo/lng_lat.js
@@ -107,7 +107,7 @@ class LngLat {
         if (input instanceof LngLat) {
             return input;
         }
-        if (Array.isArray(input) && input.length === 2) {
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Thanks @oguiter! Could you please add a test for this?

oguiter:
Hi @jfirebaugh,
Tests added... feel free to change anything :)

PR #5613, opened 2017-11-07
Throttle calls to "window.history.replaceState" to fix Mobile Safari error message

### LINKED ISSUE #5612: "history.replaceState" error on Mobile Safari
We're seeing a spike in occurrences of an error message on our embed page (i.e. https://api.mapbox.com/styles/v1/mapbox/outdoors-v9.html)

```
SecurityError: Attempt to use history.replaceState() more than 100 times per 30.000000 seconds
```

Research suggests this error message is new to iOS 9.3.  

**mapbox-gl-js version**: v0.40.1

### Steps to Trigger Behavior

 - Navigate to "https://api.mapbox.com/styles/v1/mapbox/outdoors-v9.html?access_token=..." in Safari on an iPhone running iOS 9.3 or later 
 - Zoom and pan around the map continuously for 30 seconds 

### Expected Behavior

No error messages

### Actual Behavior

```
SecurityError: Attempt to use history.replaceState() more than 100 times per 30.000000 seconds
```

lucaswoj:
As of iOS 9.3, Mobile Safari throws an error if `window.history.replaceState` is called more than 100 times per 30 seconds. This PR throttles all calls to `window.history.replaceState` to avoid triggering this error message. 

Benchmark scores: https://bl.ocks.org/anonymous/raw/9a02ed1c37970d9d6c697f667b56527a/

fixes #5612

## Next Steps

 - [x] Fix failing unit tests
 - [x] (Debug custom `util.throttle` & write tests for `util.throttle`) or find acceptable `throttle` implementation
 - [x] Post benchmark scores
 - [x] Debug `moveend` deboucing

## Test Harness

```html
<!DOCTYPE html>
<html>
<head>
    <title>Mapbox GL JS debug page</title>
    <meta charset='utf-8'>
    <link rel='stylesheet' href='/dist/mapbox-gl.css' />
    <style>
        html, body, #map { height: 100%; }
    </style>
</head>

<body>
<div id='map'></div>

<script>

    container: 'map',
    zoom: 12.5,
    center: [-77.01866, 38.888],
    hash: true
});

setInterval(() => {
    map.flyTo({
        bearing: map.getBearing() + 1
    });
}, 100)


</script>
</body>
</html>
```


lucaswoj:
> I wonder, why is it called so often in the first place? `_updateHash` is called on `moveend`. If `moveend` has been called 100 times in 30 seconds, and given that `moveend` itself is debounced (called only once in case of many quick successive movements of the map), there might be a bug lurking there of which this is only a symptom.

The provided test harness demonstrates a realistic use case that would trigger the error message even if we revamp the panning `moveend` debouncing. I will investigate further to figure out why panning produces so many `moveend` events.

> Also, I'd prefer bringing back the little `util.throttle` we had before but removed at some point than including a part of lodash for the first time in the GL JS build.

I experienced a bug in our `util.throttle` implementation (`replaceState` was still being invoked more often that it should've) which went away when I switched `lodash.throttle`. I couldn't find any tests for `util.throttle`. 

I will put more time into debugging `util.throttle` but we may want to consider a not-invented-here solution.

lucaswoj:
I must confess I'm disappointed by the size of `lodash.throttle`: 16 KB, 430 LOC (unminified and with comments).

lucaswoj:
> I wonder, why is it called so often in the first place? `_updateHash` is called on `moveend`. If `moveend` has been called 100 times in 30 seconds, and given that `moveend` itself is debounced (called only once in case of many quick successive movements of the map), there might be a bug lurking there of which this is only a symptom.

The problem is that `DragPanHandler` calls `Camera#panBy` which calls `Camera#panTo` which calls `Camera#easeTo` which calls `Camera#fire('moveend', ...)` in an unthrottled way. My view is that the real root issue here is https://github.com/mapbox/mapbox-gl-js/issues/3357.

src/util/throttle.js
```diff
@@ -0,0 +1,36 @@
...
+
+    const throttledFunction = () => {
+        const time = Date.now();
+
+        if (nextTime <= time && !pending) {
+            nextTime = time + period;
+            unthrottledFunction();
+        } else if (!pending) {
+            pending = true;
+            // eslint-disable-next-line no-use-before-define
```
mourner: The [ESLint docs](https://eslint.org/docs/rules/no-use-before-define) on the rule state that:

> In ES6, block-level bindings (let and const) introduce a “temporal dead zone” where a ReferenceError will be thrown with any attempt to access the variable before its declaration.

So if this code weren't transpiled (which may happen in future) it would through an error in theory. So perhaps let's rewrite to either use `function` declaration instead of arrows, or declare one of the vars at the top with `let`?

lucaswoj: Whoa! I've been out in the woods too long. Fixed in 20f3fe0.

PR #5616, opened 2017-11-08
Fix validation for [zoom] expression used with no curve

anandthakker:


anandthakker:
Missed this in the review of #5611

Issue #2739, opened 2016-06-14
Refactor to create StyleProperty class


lucaswoj:
There are many methods within `StyleLayer` that only operate on a single style property.
- `setLayoutProperty`
- `getLayoutProperty`
- `getLayoutValue`
- `setPaintProperty`
- `getPaintProperty`
- `getPaintValue`
- `getPaintValueStopZoomLevels`
- `getPaintInterpolationT`
- `isPaintValueFeatureConstant`
- `isLayoutValueFeatureConstant`
- `isPaintValueZoomConstant`
- ... 

A cleaner design would be to have a `StyleProperty` object per property with all of these methods

``` js
type StyleValue<T> {
  isZoomConstant: boolean;
  isFeatureConstant: boolean;
  calculate(zoom:number, featureProperties: {}) => T;
  getInterpolationT(zoom: number) =>;
  ...
}
```

At the same time we might make `StyleTransition` conform to the same interface as `StyleProperty`

ref https://github.com/mapbox/mapbox-gl-js/pull/2629#discussion_r64939804

jfirebaugh:
👍  -- what gl-native does might serve as a model here:

* [`PaintProperty`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/style/paint_property.hpp)
* [`LayoutProperty`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/style/layout_property.hpp)
* [`PropertyValue`](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/property_value.hpp)

Issue #5531, opened 2017-10-26
Revise terminology for classifying expressions


1ec5:
We lack an unambiguous vocabulary for classifying expressions. [The style specification says](https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions):

> The `expression_name` is the name of a specific expression, e.g. `'*'` or `'case'`. Each `argument_i` is either a literal value (a string, number, boolean, or `null`), or else another expression in the array form above.

The wording “a specific expression” conflates the concept of an _expression_ (an instance of the value that a paint/layout property or filter is set to) with the subset of expressions being described. Even “expression name” implies that “expression” is the entire subset of expressions. Meanwhile, the “[Get started with Mapbox GL JS expressions](https://www.mapbox.com/help/mapbox-gl-js-expressions/)” guide refers to “expression types”, which risks confusion with an expression’s evaluated data type.

Making matters worse, the style specification also describes a subset of expressions as “property expressions”. This term is supposed to refer to expressions that contain references to feature properties, but you set a paint or layout property to an expression whether or not the expression refers to a feature property.

I propose the following terminology:

* _expression_ — an instance of the value that a paint or layout property or filter is set to
* _expression operator_ — the first part of the expression array; by metonymy, the subset of expressions that begin with a given operator
* _data expression_ — an expression that contains the `get` operator somewhere within
* _camera expression_ — an expression that contains the `zoom` operator somewhere within (with a nod to #4154)

One thing going in favor of _expression operator_ is that filters are classified by “operators” too.

/cc @mapbox/gl @colleenmcginnis

davidtheclark:
> an instance of the value that a paint or layout property or filter is set to

@1ec5 I'm having some trouble understanding this phrase. Could you clarify a bit more?

1ec5:
Sure: you set `text-field` or `filter` to an expression (which may contain subexpressions).

colleenmcginnis:
>Meanwhile, the “Get started with Mapbox GL JS expressions” guide refers to “expression types”, which risks confusion with an expression’s evaluated data type.

I did struggle to find a neutral word to use here that would also be intuitive. I think "expression operator" is a good fit.

stevage:
Are `curve` and `let` and `concat` examples of "expression operators" too? That seems a bit awkward, as they would be statements or functions in other languages.

1ec5:
[...]

PR #5628, opened 2017-11-08
Expression kinds

jfirebaugh:
Introduce the same sort of discriminated union as is used in native: an expression (or function, or generally, any property value) has one of the following kinds: `constant`, `camera`, `source`, `composite`.

This is a prerequisite for #2739 / #3044.

### DIFF
--- a/src/style-spec/expression/index.js
+++ b/src/style-spec/expression/index.js
@@ -29,43 +29,16 @@ export type GlobalProperties = {
     heatmapDensity?: number
 };
 
-export type StyleExpressionContext = 'property' | 'filter';
-
 export type StyleExpressionErrors = {
     result: 'error',
     errors: Array<ParsingError>
 };
 
-type ZoomConstantExpression = {
+export type StyleExpression = {
     result: 'success',
-    context: StyleExpressionContext,
-    isZoomConstant: true,
-    isFeatureConstant: boolean,
     evaluate: (globals: GlobalProperties, feature?: Feature) => any,
-    // parsed: Expression
+    parsed: Expression
 };
-
-export type StyleDeclarationExpression = ZoomConstantExpression | {
-    result: 'success',
-    context: 'property',
-    isZoomConstant: false,
-    isFeatureConstant: boolean,
-    evaluate: (globals: GlobalProperties, feature?: Feature) => any,
-    // parsed: Expression,
-    interpolationFactor: (input: number, lower: number, upper: number) => number,
-    zoomStops: Array<number>
-};
-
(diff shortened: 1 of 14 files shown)


src/style-spec/expression/index.js
```diff
@@ -135,73 +100,86 @@ function createExpression(expression: mixed,
...
-        return {
-            result: 'error',
-            errors: [new ParsingError('', 'property expressions not supported')]
-        };
+    return success({ evaluate, parsed });
+}
+
+type ConstantExpression = {
+    kind: 'constant',
+    evaluate: (globals: GlobalProperties, feature?: Feature) => any
```
jfirebaugh: It's not feasible yet, but I hope to be able to specialize the `evaluate` method appropriately for the various kinds in a later commit. For example, constant expression evaluation should require no arguments, and the `Feature` argument should not be options for source and composite expression evaluation.

anandthakker: Once this happens, we'll need a way to heatmap-color.

src/style-spec/expression/index.js
```diff
@@ -135,37 +107,65 @@ function createExpression(expression: mixed,
...
+
+type ConstantExpression = {
+    result: 'constant',
+    evaluate: (globals: GlobalProperties, feature?: Feature) => any
+};
+
+type SourceExpression = {
+    result: 'source',
+    evaluate: (globals: GlobalProperties, feature?: Feature) => any,
+};
```
anandthakker: What do you think about `FeatureExpression`?  Seems more direct to me, but maybe there was a reason we opted for `SourceFunction` over `FeatureFunction` before...

src/symbol/symbol_size.js
```diff
@@ -31,17 +31,17 @@ export type SizeData = {
...
+    if (!declaration || declaration.isZoomConstant()) {
         return isFeatureConstant ? {
             functionType: 'constant',
             layoutSize: layer.getLayoutValue(sizeProperty, {zoom: tileZoom + 1})
         } : { functionType: 'source' };
     }
 
     // calculate covering zoom stops for zoom-dependent values
-    const levels = declaration.expression.zoomStops;
+    const levels = (declaration.expression: any).zoomStops;
```
anandthakker: This `any` cast is justified by the `!declaration || declaration.isZoomConstant()` check above, right?  Let's include a line comment indicating that.

src/style-spec/expression/index.js
jfirebaugh: #5531 suggests `DataExpression` -- let's hash out terminology there. For now, I'm happy just getting JS consistent with current native terms. 😄

PR #5632, opened 2017-11-09
Fix rendering for icon-text-fit + data-driven text-size

### LINKED ISSUE #5631: Wrong rendering for data-driven text-size + icon-text-fit
This is a specific example of #3044:

https://github.com/mapbox/mapbox-gl-js/blob/3b7187a68e33c72408e34af82615ecf6a582f92c/src/symbol/quads.js#L74

If `text-size` is data-driven, this code is incorrect. What needs to happen is that the evaluated value for the current feature is passed in. Here's native implementing that:

https://github.com/mapbox/mapbox-gl-native/blob/a71756ea8d62cb448963a59e215e1324cfd57d9f/src/mbgl/text/quads.cpp#L18

jfirebaugh:
Fixes #5631.


test/integration/render-tests/regressions/mapbox-gl-js#5631/style.json
```diff
@@ -0,0 +1,74 @@
...
+          "type": "identity",
+          "property": "name"
+        },
+        "text-size": {
+          "type": "identity",
+          "property": "size"
+        },
+        "icon-image": "label",
+        "icon-text-fit": "both",
+        "icon-ignore-placement": true,
```
jfirebaugh: I'm not sure why `icon-ignore-placement` and `icon-allow-overlap` were needed here, because the two symbols don't appear to collide. But without them, only "A" shows up. Might be another bug lurking here.

ChrisLoer: Yeah, it looks like there is a bug -- the icons have collision boxes that are much bigger than the text:

screenshot 2017-11-08 17 27 30

I'll look into that tomorrow.

src/symbol/quads.js
```diff
@@ -71,7 +71,7 @@ function getIconQuads(anchor: Anchor,
     if (layout['icon-text-fit'] !== 'none' && shapedText) {
         const iconWidth = (right - left),
             iconHeight = (bottom - top),
-            size = layout['text-size'] / 24,
+            size = layer.getLayoutValue('text-size', globalProperties, feature) / 24,
```
ChrisLoer: Why isn't this using `{zoom: bucket.zoom + 1}` like we do in `symbol_layout#addFeature`? That seems like the best match. @anandthakker understands that logic better than I do, though, I think.

anandthakker: Woah, good point @ChrisLoer.   This raises an interesting issue.  It's not just that we need to evaluate `text-size` with `{zoom: bucket.zoom + 1}` during layout (though I think you're right about that) -- if text-size is zoom depentent, we also need to scale the icon at render time according to the current text-size value.  And if text-size is a composite function, that's going to be problematic -- we _could_ try to use the text-size attributes for the icon rendering pass in that case, but that would only help for `icon-text-size: both`.

In any case, I think this goes well beyond the scope of this PR.  I'll ticket it separately, 👍 for now!

anandthakker: https://github.com/mapbox/mapbox-gl-js/issues/5656

test/integration/render-tests/regressions/mapbox-gl-js#5631/style.json
```diff
@@ -0,0 +1,74 @@
...
+      "source": "geojson",
+      "layout": {
+        "text-font": [
+          "Open Sans Semibold",
+          "Arial Unicode MS Bold"
+        ],
+        "text-field": {
+          "type": "identity",
+          "property": "name"
+        },
```
anandthakker: `["get", "name"]`

test/integration/render-tests/regressions/mapbox-gl-js#5631/style.json
```diff
@@ -0,0 +1,74 @@
...
+        "text-size": {
+          "type": "identity",
+          "property": "size"
+        },
```
anandthakker: `["get", "size"]`

PR #5633, opened 2017-11-09
Switch to default tap test reporter

lucaswoj:
With 10164 tests, the `dot` reporter isn't very ergonomic. The default test reporter works much better.

screen shot 2017-11-08 at 4 40 24 pm
screen shot 2017-11-08 at 4 38 40 pm


jfirebaugh:
Oh yes, much better! It doesn't seem to work on CI though -- instead it outputs raw TAP output, which is hard to read.

lucaswoj:
Good catch @mourner @jfirebaugh. Fixed in b33f1a7.

PR #5636, opened 2017-11-09
Increase max maxZoom from 20 to 22

jingsam:
I think maxZoom should be 22, instead of 20

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -197,8 +197,7 @@ const defaultOptions = {
  * @example
  * var map = new mapboxgl.Map({
  *   container: 'map',
- *   center: [-122.4
- 679, 37.772537],
+ *   center: [-122.4679, 37.772537],
  *   zoom: 13,
  *   style: style_object,
  *   hash: true,
@@ -1174,8 +1173,8 @@ class Map extends Camera {
      * Sets the zoom extent for the specified style layer.
      *
      * @param {string} layerId The ID of the layer to which the zoom extent will be applied.
-     * @param {number} minzoom The minimum zoom to set (0-20).
-     * @param {number} maxzoom The maximum zoom to set (0-20).
+     * @param {number} minzoom The minimum zoom to set (0-22).
+     * @param {number} maxzoom The maximum zoom to set (0-22).
      * @returns {Map} `this`
      * @example
      * map.setLayerZoomRange('my-layer', 2, 5);
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -197,7 +197,8 @@ const defaultOptions = {
  * @example
  * var map = new mapboxgl.Map({
  *   container: 'map',
- *   center: [-122.420679, 37.772537],
+ *   center: [-122.4
+ 679, 37.772537],
  *   zoom: 13,
  *   style: style_object,
  *   hash: true,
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -197,7 +197,8 @@ const defaultOptions = {
  * @example
  * var map = new mapboxgl.Map({
  *   container: 'map',
- *   center: [-122.420679, 37.772537],
+ *   center: [-122.4
+ 679, 37.772537],
```
mourner: Accidental?

jingsam: Ops. Updated

andrewharvey:
Why limit this to 22?

The Style Spec supports layers up to zoom 24 -> https://www.mapbox.com/mapbox-gl-js/style-spec/#layer-maxzoom
GL JS seems to work fine up to zoom 24 -> https://jsbin.com/monenufivo/edit?html,console,output

So I think the JSDoc should say the upper limit is 24 (while possibly still documenting the default map maxzoom value as 22, which is what it is currently).

src/ui/map.js
```diff
@@ -197,7 +197,7 @@ const defaultOptions = {
  * @example
  * var map = new mapboxgl.Map({
  *   container: 'map',
- *   center: [-122.420679, 37.772537],
+ *   center: [-122.4679, 37.772537],
```
andrewharvey: Is there any reason to change the coordinates of the example?

jingsam: Accidental change 😂

jingsam:
@anandthakker To my understanding, there are no validation for `zoom`, `maxZoom` and other options When initializing an instance of Map. Try to set `zoom: 100; maxZoom: 10000`, no errors complains out!

andrewharvey:
> To my understanding, there are no validation for zoom, maxZoom and other options When initializing an instance of Map. Try to set zoom: 100; maxZoom: 10000, no errors complain out!

More realistically a maxzoom of 26 will work just the same in my example as a maxzoom of 24. Granted, given the style spec maxzoom for a layer is 24 and given https://github.com/mapbox/geojson-vt/issues/87 seems to give a current limit at 24, I think it makes sense for the documentation here to hint you can go up to zoom 24, unless I missing any bugs which mean maps don't work up to z24?

jingsam:
Should we change `defaultMaxZoom` to 24? I concerned that mapbox-streets do not have tiles exceeding zoom 22. And ios and android sdk may have different limits on the maxZoom.

We should unified the `maxZoom` to avoid confusion.

andrewharvey:
> Should we change defaultMaxZoom to 24? I concerned that mapbox-streets do not have tiles exceeding zoom 22. And ios and android sdk may have different limits on the maxZoom.

I don't think so because of exactly the reason you mentioned, when looking at the map at zoom 24 you can see artefacts from mapbox-streets as it's overzoomed from 22.

I think it makes sense only to update the JSDoc with regards to the maxzoom you can practically use (which I think is 24), but leave the default maxzoom at 22.

jingsam:
Changed layerZoomRange to 0-24

src/ui/map.js
```diff
@@ -481,7 +481,7 @@ class Map extends Camera {
      * If the map's current zoom level is lower than the new minimum,
      * the map will zoom to the new minimum.
      *
-     * @param {number | null | undefined} minZoom The minimum zoom level to set (0-20).
+     * @param {number | null | undefined} minZoom The minimum zoom level to set (0-22).
```
andrewharvey: I think this should also be zoom 24 to match the minzoom 24 later at the bottom of this PR. Looks like the ship on this one has sailed.

PR #5652, opened 2017-11-10
Symbol layer order no longer matters to the worker

jfirebaugh:
With viewport label placement, the ordering dependency is handled during foreground placement. The worker no longer needs to be concerned about layer order, and relayout doesn't need to happen just because symbol layers are reordered.


jfirebaugh:
I've seen that one failing intermittently for a while.

Issue #3557, opened 2016-11-07
Provide default contents for Marker


jfirebaugh:
For closer parity with Leaflet markers, the `element` parameter to the `Marker` class should be optional, and the default should be render a traditional marker pin icon similar to ![image](https://cloud.githubusercontent.com/assets/98601/20068106/67b546de-a4cc-11e6-9340-a8ec330220f1.png).

For implementation, I suggest that this use some inline SVG source, so as not to require an external asset.

mikehatfield:
Any progress on this?

jfirebaugh:
@mikehatfield Nope -- want to work on a PR?

rahuldhawani:
Anyone working on this? If not, then I would like to pick it up.

jfirebaugh:
Go for it @rahuldhawani!

samanpwbb:
I recommend this marker, for consistency across gl native / mapbox's static image API: https://github.com/mapbox/makiwich/blob/master/sources/marker-large.svg

Qwertie-:
This would be very helpful. It's overly complex to add a basic marker to the map currently.

PR #5661, opened 2017-11-12
Provide default element for Marker class per #3557

Syntaf:
This PR creates a basic html element for when the `element` parameter to the `Marker` class is null. The icon is pulled from https://www.mapbox.com/maki-icons/ (marker-15), so there shouldn't be any licensing issues.

If this is PR is a valid solution to #3557, maybe something in the docs could be added to show a default marker exists? I can add that if necessary.

Here is a sample `index.html` to test this:

```html
<!DOCTYPE html>
<html>
<head>
    <title>Mapbox GL JS debug page</title>
    <meta charset='utf-8'>
    <link rel='stylesheet' href='/dist/mapbox-gl.css' />
    <style>
        html, body, #map { height: 100%; }
    </style>
</head>

<body>
<div id='map'></div>

<script>

    container: 'map',
    zoom: 2.5,
    center: [-77.01866, 38.888],
    hash: true
});

map.on('load', function() {
    // default marker
    var marker = new mapboxgl.Marker()
    .setLngLat([30.5, 50.5])
    .addTo(map);

    // custom element marker
    var el = document.createElement('div');
    el.className = 'marker';
    el.style.backgroundImage = 'url(http://via.placeholder.com/25x25)';
    el.style.width = '25px';
    el.style.height = '25px';

    var marker2 = new mapboxgl.Marker(el)
    .setLngLat([41, 50.5])
    .addTo(map);
});

</script>
</body>
</html>
```

(Updated) Example of icon:
marker

### DIFF
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -39,21 +39,21 @@ class Marker {
             element = DOM.create('div');
 
             // create svg marker icon based on the Maki icon marker-15
-            var svg = DOM.createNS('http://www.w3.org/2000/svg', 'svg');
+            const svg = DOM.createNS('http://www.w3.org/2000/svg', 'svg');
             svg.setAttributeNS(null, 'height', '30');
             svg.setAttributeNS(null, 'width', '30');
             svg.setAttributeNS(null, 'x', '0');
             svg.setAttributeNS(null, 'y', '0');
             svg.setAttributeNS(null, 'viewBox', '0 0 19 19');
 
-            var rect = DOM.createNS('http://www.w3.org/2000/svg', 'rect');
+            const rect = DOM.createNS('http://www.w3.org/2000/svg', 'rect');
             rect.setAttributeNS(null, 'fill', 'none');
             rect.setAttributeNS(null, 'x', '0');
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Thanks for the contribution @Syntaf, this is a much-requested feature!

Would you mind changing the default to use the SVG content @samanpwbb suggested [here](https://github.com/mapbox/mapbox-gl-js/issues/3557#issuecomment-320000273)?

Syntaf:
Ah, I must have missed that comment. Seems like a much better icon for this feature, I'll modify it to use that SVG when I'm home today.

Syntaf:
Modified the code to use the marker suggested, and updated the example picture in the original PR.

jfirebaugh:
This looks great @Syntaf, thanks again for the contribution. Can you make two small additions and then I'll merge?

* Add a test in [`marker.test.js`](https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/ui/marker.test.js). This can just be a copy of the `constructor` test that omits the parameter.
* Where the JSDoc comment currently says "creates a div element by default", update that to give a brief description of the new default.

PR #5662, opened 2017-11-12
Re-export Transform for external use

mike-marcacci:
This simply exports the `Transform` class for external use. For apps with unidirectional data flow (such as a large React app), it's much cleaner to do these calculations at a higher level than a map component. This would allow a root component to, for example, calculate map bounds given a center/zoom/pitch from the URL and and window resolution.

This is especially important for server side rendering, where results of a geo-bounded query shouldn't have to wait for a map to be instantiated.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -14,6 +14,7 @@ const Marker = require('./ui/marker');
 const Style = require('./style/style');
 const LngLat = require('./geo/lng_lat');
 const LngLatBounds = require('./geo/lng_lat_bounds');
+const Transform = require('./geo/transform');
 const Point = require('@mapbox/point-geometry');
 const Evented = require('./util/evented');
 const config = require('./util/config');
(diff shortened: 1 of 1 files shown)


mike-marcacci:
Also, in case anyone else is attempting to get to this class, it's possible to use this (nasty) little hack:

```es6
import mapboxgl from 'mapbox-gl/dist/mapbox-gl.js';
	container: document.createElement('div'),
	style: { version: 8, sources: {}, layers: [] },
	interactive: false,
});
const Transform = map.transform.constructor;
map.remove();
```

anandthakker:
Thanks @mike-marcacci. 

I feel hesitant about this change at present.  The `Transform` class contains a fair number of implementation details that really ought to remain private; exposing them in the public API risks inhibiting our flexibility to make internal refactors.

Would you mind opening an issue that describes the pain point / use case that you're trying to address here?  I do think it's an important one, I just think it would be good to step back from this particular solution and discuss the problem

mike-marcacci:
Hi @anandthakker, thanks so much for the thoughtful reply. I definitely sympathize with not wanting to tie your public API to implementation details! I'll go ahead and open an issue detailing my use case.

PR #5671, opened 2017-11-14
retain all loaded children

mollymerp:
... even those at ZL more than one greater than the ideal tile.

popping this out of #5286. right now we're only checking for loaded children that are exactly one ZL higher from the ideal tile. this change makes it so we retain all loaded children of not-yet-loaded ideal tiles so that in the event of a fast zoom out we don't drop deep descendent child tiles before the parents are loaded. 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/test/unit/source/source_cache.test.js
+++ b/test/unit/source/source_cache.test.js
@@ -708,6 +708,39 @@ test('SourceCache#_updateRetainedTiles', (t)=> {
         t.end();
     });
 
+    t.test('retains all loaded children ', (t) => {
+        const sourceCache = createSourceCache({
+            loadTile: function(tile, callback) {
+                tile.state = 'errored';
+                callback();
+            }
+        });
+
+        const idealTile = new TileCoord(3, 1, 2);
+        sourceCache._tiles[idealTile.id] = new Tile(idealTile);
+        sourceCache._tiles[idealTile.id].state = 'errored';
+
+        const loadedChildren = [
+            new TileCoord(4, 2, 4, 0),
+            new TileCoord(4, 3, 4, 0),
+            new TileCoord(4, 2, 5, 0),
+            new TileCoord(5, 6, 10, 0),
+            new TileCoord(5, 7, 10, 0),
+            new TileCoord(5, 6, 11, 0),
+            new TileCoord(5, 7, 11, 0)
+        ];
+
+        for (const t of loadedChildren) {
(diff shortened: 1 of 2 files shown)


anandthakker:
@mollymerp looking back at https://github.com/mapbox/mapbox-gl-native/blob/3bef7593a64a51e86dd5a2ed9fd36b4a143350b0/src/mbgl/algorithm/update_renderables.hpp, I think this is not what -native does -- will this need a port?

mollymerp:
yeah, this isn't the same behavior as native, but it was the behavior in js prior to #5119. I'm not sure if there is a mobile-related reason as to why we'd want to purge these tiles from the cache during a zoom out, but if there isn't I think porting over would be a good idea. cc @kkaefer

test/unit/source/source_cache.test.js
```diff
@@ -724,12 +757,6 @@ test('SourceCache#_updateRetainedTiles', (t)=> {
         stateCache[idealTiles[0].id] = 'loaded';
         const retained = sourceCache._updateRetainedTiles(idealTiles, 1);
         t.deepEqual(getTileSpy.getCalls().map((c)=>{ return c.args[0]; }), [
-            // checks all child tiles to see if they're loaded before loading parent
-            new TileCoord(2, 0, 2),
-            new TileCoord(2, 1, 2),
-            new TileCoord(2, 0, 3),
-            new TileCoord(2, 1, 3),
```
anandthakker: At first, I was going to ask whether there's some other way we could preserve testing of this logic, checking for child tiles before trying to load parent...  But, if I'm understanding things right, it looks like the `'prefer loaded child tiles to parent tiles'` at least partially confirms that the right thing is happening, so we're probably okay.  Does that sound right to you, @mollymerp? 

Do you think it would be worthwhile to add a test that's similar to 
the `'prefer loaded child tiles to parent tiles'` one, but where _all_ the child tiles are loaded, to confirm that we don't load the parent in that case?

mollymerp: done!

mollymerp:
per chat with @kkaefer this morning – he is 👍 on restoring this behavior in JS and porting to native. apparently the native behavior was originally implemented to avoid having to render too many child tiles (if they were highly under-zoomed), but because we're only retaining already loaded child tiles, it is not likely that we'd end up with an unreasonable number of tiles to render.

PR #5673, opened 2017-11-14
Add update call when showCollisionBoxes disabled #5653

### LINKED ISSUE #5653: Turning off showCollisionBoxes doesn't repaint
On a map with collision boxes visible, run `map.showCollisionBoxes = false`. Expected result: the boxes disappear immediately. Actual result: they disappear only after something else causes a repaint.

Syntaf:
It seems `showCollisionBoxes()` inside `map.js` was missing a call to `this._update()` when collisionboxes are set to false. This PR adds that in.

Should fix and close #5653


jfirebaugh:
Great, thanks! This fixes #5653 I presume?

PR #5679, opened 2017-11-15
check if before layer exists

jingsam:
I think we should check if `before` layer exists when move a layer

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -622,6 +622,10 @@ class Style extends Evented {
         this._order.splice(index, 1);
 
         const newIndex = before ? this._order.indexOf(before) : this._order.length;
+        if (before && newIndex === -1) {
+            this.fire('error', { message: new Error(`Layer with id "${before}" does not exist on this map.`)});
+            return;
+        }
         this._order.splice(newIndex, 0, id);
 
         this._layerOrderChanged = true;
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -568,7 +568,7 @@ class Style extends Evented {
 
         const index = before ? this._order.indexOf(before) : this._order.length;
         if (before && index === -1) {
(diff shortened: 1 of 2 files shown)


src/style/style.js
```diff
@@ -622,6 +622,10 @@ class Style extends Evented {
         this._order.splice(index, 1);
 
         const newIndex = before ? this._order.indexOf(before) : this._order.length;
+        if (before && newIndex === -1) {
+            this.fire('error', { message: new Error(`Layer with id "${before}" does not exist on this map.`)});
```
mourner: The key here should be `error` since you're passing an error object, not a message.

mourner:
Otherwise looks good!

src/style/style.js
mourner: Oh, I see it was mixed up in the original code.

andrewharvey:
@jingsam did you want to fix up the JSDoc just a few lines above your change while you're at it? https://github.com/mapbox/mapbox-gl-js/blob/master/src/style/style.js#L601

src/style/style.js
mollymerp: I think you're right, the key here should be `error` as well as here https://github.com/mapbox/mapbox-gl-js/pull/5679/files#diff-0fcf9ef37a1614185cad0dcbd5ec5529L571

jingsam: Shall I change it to a `{error: ....}`?

mollymerp: yes please @jingsam and it would be great if you could change the it to `{error: new Error...` on line 571 as well. thank you!

jingsam:
fixed tests. but this seems odd:
```
t.match(error.error, /does not exist on this map/);
```

I prefer `error.message`

Issue #5677, opened 2017-11-14
Query results should contain evaluated style properties


jfirebaugh:
The documentation for `queryRenderedFeatures` says:

> Each feature includes a top-level layer property whose value is an object representing the style layer to which the feature belongs. Layout and paint properties in this object contain values which are fully evaluated for the given zoom level and feature.

The second sentence doesn't reflect the current behavior: currently, the `layer` objects in the results are just [`serialize`d layers](https://github.com/mapbox/mapbox-gl-js/blob/cb42e3930a76d5034bd32f03a18d71ee449928a1/src/data/feature_index.js#L235), and `serialize` doesn't produce property values evaluated for the given zoom level or feature. You can confirm this via [this](https://www.mapbox.com/mapbox-gl-js/example/queryrenderedfeatures/) example.

I agree with the intent of the documentation -- since you can already obtain the unevaluated property values via runtime styling APIs, it's more useful to get the evaluated properties. We should fix the behavior to match what's documented.

asheemmamoowala:
Closed by https://github.com/mapbox/mapbox-gl-js/pull/9198

PR #5692, opened 2017-11-16
Evaluate query results on queryRenderedFeatures #5677

Syntaf:
#### Preface: the code in this PR likely needs to change, but I wanted to show a 'working' example and mention a possible fix if i'm on the right track.

If I understand the issue correctly, this PR will 'fix' #5677, and properly evaluate the current values before showing them. 

The solution I wrote however is a bit hacky, as I realized a couple of things:

- Proper evaluation of the layout and paint declaration rely on the global property `zoom`, which is not available during the `serialize` method.
- Modifying `serialize` to use the global property zoom and pass it along during the mapping of layout and paint would solve one problem and create another, as serialize is also used to grab non-evaluated results.

I think `filterMatching` needs to separate some logic out for `queryRenderedFeatures` somehow, such that you don't need to check when the global context is valid to pass or not.  

My first impression is to maybe create a flag within `filterMatching` which defines whether to serialize **and** evaluate, or _just_ serialize. That way it won't have to rely on checking whether `this.coord.z` is defined or not to determine if the serialization should also perform evaluation.

Something along the lines of:
```javascript
    filterMatching(
        result: {[string]: Array<{ featureIndex: number, feature: GeoJSONFeature }>},
        matching: Array<any>,
        array: any,
        queryGeometry: Array<Array<Point>>,
        filter: FeatureFilter,
        filterLayerIDs: Array<string>,
        styleLayers: {[string]: StyleLayer},
        bearing: number,
        pixelsToTileUnits: number,
        serializeAndEvaluate?: boolean // make optional maybe?
    ) {

            // ....

           // if asked to evaluate, we know the global context is defined and valid
           if (serializeAndEvaluate) {
               (geojsonFeature: any).layer = styleLayer.serializeWithEvaluation(globals);
           } else {
               // otherwise, perform a normal serialization
               (geojsonFeature: any).layer = styleLayer.serialize();
           }
```

Thoughts?

P.S. I used [this](https://www.mapbox.com/mapbox-gl-js/example/queryrenderedfeatures/) example from the ticket to test this.

### DIFF
--- a/src/data/feature_index.js
+++ b/src/data/feature_index.js
@@ -232,7 +232,11 @@ class FeatureIndex {
                 }
 
                 const geojsonFeature = new GeoJSONFeature(feature, this.z, this.x, this.y);
-                (geojsonFeature: any).layer = styleLayer.serialize();
+                let globals;
+                if (this.coord.z !== undefined) {
+                    globals = { zoom: this.coord.z };
+                }
+                (geojsonFeature: any).layer = styleLayer.serialize(globals);
                 let layerResult = result[layerID];
                 if (layerResult === undefined) {
                     layerResult = result[layerID] = [];
--- a/src/style/style_layer.js
+++ b/src/style/style_layer.js
@@ -233,7 +233,7 @@ class StyleLayer extends Evented {
         }
     }
 
-    serialize() {
+    serialize(globals?: GlobalProperties) {
         const output : any = {
             'id': this.id,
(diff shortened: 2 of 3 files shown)


ryanhamley:
This section of the library has changed significantly since this PR was submitted so I'm going to close it. We'd be happy to revisit a new PR. Sorry for the lack of response.

PR #5694, opened 2017-11-17
Revise terminology for classifying expressions

jfirebaugh:
Per #5531.

### DIFF
--- a/docs/pages/style-spec.js
+++ b/docs/pages/style-spec.js
@@ -1115,60 +1115,41 @@ export default class extends React.Component {
                             <h2><a href='#expressions' title='link to expressions'>Expressions</a></h2>
 
                             <p>The value for any <a href="#layout-property">layout property</a>, <a
-                                href="#paint-property">paint property</a>, or <a
-                                href="#layer-filter">filter</a> may be specified as an <em>expression</em>.
-                                Expressions define how one or more feature
-                                property values and/or the current zoom level are combined using
-                                logical, mathematical, string, or color operations to produce the
-                                appropriate style property value or filter decision.</p>
+                                href="#paint-property">paint property</a>, or <a href="#layer-filter">filter</a> may be
+                                specified as an <em>expression</em>. An expression defines a formula for computing the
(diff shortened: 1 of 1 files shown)


docs/pages/style-spec.js
```diff
@@ -1115,60 +1115,40 @@ export default class extends React.Component {
...
-                            <p>A <a id="property-expression" className="anchor"></a><strong>property expression</strong> is
-                                any expression defined using an expression that includes a reference to feature property
-                                data. Property expressions allow the appearance of a feature to change with its
-                                properties. They can be used to visually differentate types of features within the same
-                                layer or create data visualizations.</p>
-
-                            <h4>Example: a property expression</h4>
+
+                            <h3>Data expressions</h3>
+                            <p>A <em>data expression</em> is any expression that uses one of the data operators:
```
anandthakker: For some readers, it may be a bit easier to grasp this if we couple the list-based definition with a conceptual definition.  Something like:

```
A <em>data expression</em> is any expression that access feature data -- that is, any expression that uses one of the data operators:
```

docs/pages/style-spec.js
```diff
@@ -1256,9 +1242,19 @@ export default class extends React.Component {
...
+                            <p>There is an important difference between layout and paint properties in
+                                the timing of camera expression evaluation. Paint property camera expressions are
+                                re-evaluated whenever the zoom level changes, even fractionally. For example, a paint
+                                property camera expression will be re-evaluated continuously as the map moves between
+                                zoom levels 4.1 and 4.6. On the other hand, a layout property camera expression is
+                                evaluated only at integer zoom levels. It will <em>not</em> be re-evaluated as the zoom
+                                changes from 4.1 to 4.6 -- only if it goes above 5 or below 4.
+                            </p>
+
+                            <h3>Composite expressions</h3>
```
1ec5: Do we still need a dedicated term for expressions that compose multiple kinds of operators? Or would it suffice for a section titled “Composition” to point out that a single expression can use a mixture of subexpression operators?

PR #5695, opened 2017-11-17
Validate container option

jingsam:


jingsam:
Relate to #5635

jingsam:
Seems lead to wrong way. Shall I import HTMLElement from `util/window.js`

mollymerp:
Yes, I believe importing `HTMLElement` will fix it @jingsam

PR #5698, opened 2017-11-17
Enable data-driven values for text-font

### LINKED ISSUE #5568: Add data-driven-styling support for `text-font`
Data-driven styling for font stacks would be very useful, especially now that expressions have landed. Per chat with @jfirebaugh, this was infeasible before per-tile glyph atlases, but feasibility has not been reviewed since per-tile glyph atlases landed.

We are also now working on [expression constraints](https://github.com/mapbox/mapbox-gl-js/issues/5393) that will allow the use of data-driven expressions with limited output values.

/cc @anandthakker @aparlato

jfirebaugh:
Enables support for data-driven `text-font` values. Fixes #5568.

The port to native will need to take into account https://github.com/mapbox/mapbox-gl-native/issues/9939 -- my proposed resolution is described in https://github.com/mapbox/mapbox-gl-native/issues/9939#issuecomment-335895131.


anandthakker:
The port to native will need to take into account mapbox/mapbox-gl-native#9939 -- my proposed resolution is described in mapbox/mapbox-gl-native#9939 (comment).

@jfirebaugh I was thinking we would impose this restriction at the level of spec validation (and so in GL JS as well), not just in -native.

jfirebaugh:
Oh, yeah, that's probably a good idea.

test/unit/style-spec/fixture/text-font.input.json
```diff
@@ -17,6 +17,106 @@
...
+    },
+    {
+      "id": "invalid expression - let/var",
+      "type": "symbol",
+      "source": "vector",
+      "source-layer": "layer",
+      "layout": {
+        "text-font": ["let", "p", ["array", "string", ["get", "text-font"]], ["var", "p"]],
+        "text-field": "{foo}"
+      }
```
anandthakker: Nit: let's include `"text-font"` in the id for these layers, since that's what they're testing.

jfirebaugh: Do you think this is necessary given that `text-font` is in the fixture file name?

anandthakker: Oh! Whoops -- I'd assumed without looking that this was in the `functions` fixture.  👍 as is

anandthakker:
Oof, I always forget about those.  Another option would be to require data-driven values to be expressions (not functions)

jfirebaugh:
I thought about that, but I think it might be confusing for a long-existing property to have that restriction. We did do it for `heatmap-color`, but that was (essentially) a new property, and it has a special-case input.

anandthakker:
> I think it might be confusing for a long-existing property to have that restriction.

Hm, yeah good point -- I suppose if someone was already using a zoom-only stop function, they'd expect to be able to upgrade it to a zoom-and-property stop function after this.

nickidlugash:
@jfirebaugh looks great – thanks for pushing on this. Can't wait to test out some expression values for `text-font`! /cc @aparlato

jplante:
Just verifying: 

1. This being data-driven does not appear to be updated on the API documentation Style Specification.  `text-font` is listed as "data-driven styling" and "Not yet supported" across platforms.

2. Even if it does work, does it support `identity` functions?

anandthakker:
@jplante Yes, the API docs need to be updated -- thanks for noting that!

> does it support identity functions

No, identity functions aren't supported for `text-font`. (And, to generalize this, text-font [expressions](https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions) must contain all possible output values as literals within the expression)

Issue #5219, opened 2017-08-31
Add detail to Expressions section of style-spec documentation


anandthakker:
- [x] Narrative intro section with a couple examples
- [x] High level description of type system, assertions vs. coercions
- [x] `curve` details
- [x] `match` details
- [x] `let` details
- [x] `coalesce` details
- [x] details about type coercion rules (per https://github.com/mapbox/mapbox-gl-js/issues/5219#issuecomment-339164651)

davidtheclark:
@anandthakker How about more explanation in the coercion section about what types get coerced in what ways? I'm wondering about questions like the following: Will the boolean `true` be coerced to the number `1`, or error? Will the string `700px` be coerced to the number `700`, or error? Do all non-empty strings coerced to booleans become `true`? Basically: How close are these rules to the rules of coercion in JavaScript?

anandthakker:
👍 good call @davidtheclark, i'll add it to the list in the top of the ticket

jfirebaugh:
I think these are all addressed now -- followup issues are #5500 and #5685.

PR #5701, opened 2017-11-17
High level description of type system, assertions vs. conversion

jfirebaugh:
Per #5219, add documentation regarding the expression type system.

### DIFF
--- a/docs/pages/style-spec.js
+++ b/docs/pages/style-spec.js
@@ -1197,7 +1197,8 @@ export default class extends React.Component {
                                     }`)}
                             </div>
 
-                            <p>This example uses the <a href="#expressions-interpolate"><code>interpolate</code></a>
+                            <p>This example uses the <a
+                                href="#expressions-interpolate"><code>interpolate</code></a>
                                 operator to define a linear relationship between zoom level and circle size using a set
                                 of input-output pairs. In this case, the expression indicates that the circle radius should
                                 be 1 pixel when the zoom level is 5 or below, and 5 pixels when the zoom is 10 or above.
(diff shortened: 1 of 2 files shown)


docs/pages/style-spec.js
```diff
@@ -1275,6 +1276,57 @@ export default class extends React.Component {
...
+                                identifying a valid <a href="#types-color">color</a>. If this check fails, an error will
+                                be indicated in an SDK-specific way (typically a log message), and the default value for
+                                the property will be used instead.
+                            </p>
+
+                            <p>In most cases, this verification will occur automatically wherever it is needed. However,
+                                in certain situations, the SDK may be unable to automatically determine the expected
+                                result type of a data expression from surrounding context. For example, it is not clear
+                                whether the expression <code>["&lt;", ["get", "a"], ["get", "b"]]</code> is attempting
+                                to compare strings or numbers. In situations like this, you must use one of
```
anandthakker: "must" could be misleading, since using a conversion rather than an assertion is also an option.

anandthakker: How about:

> In situations like this, the simplest option is to use <em>type assertions</em>: <code>["&lt;", ["number", ["get", "a"]], ["number", ["get", "b"]]]</code>.  A type assertion <em>checks</em> that the feature data actually matches the expected type of the data expression.  If this check fails, it produces an error and causes the whole expression to fall back to the default value for the property being defined.  The assertion operators are ...
>
> Another option is to use a <em>type conversion</em>.  Expressions only perform one kind of implicit type conversion...

This repeats the description of runtime type checking from the previous paragraph, but I think the redundancy is worthwhile in order to explicitly state what an assertion does.

jfirebaugh: Hmm, not convinced about this change. I think we should keep the use cases clearly separated here: conversion is for when you have one type, but need another. Assertion is for when the implementation needs you to tell it what type you have. If you are using a conversion, then you're not in a situation where the implementation is unable to automatically determine the type.

anandthakker: > If you are using a conversion, then you're not in a situation where the implementation is unable to automatically determine the type.

You could be in such a situation -- e.g., say you want to do `["==", ["id"], ["get", "name"]]`.  Depending on your data, you really might want to use conversion rather than assertion as your way out of the type checking error.

That said, I'm 👍 for focusing on assertions as the primary way to deal with this, since, though conversions _can_ be used in this way, their typical use cases are much broader.    I still think the word "must" is misleading, and also that we should include an explicit description of assertions' behavior.

PR #5702, opened 2017-11-17
Introduce generic serializer for web worker transfers

### LINKED ISSUE #5143: Introduce a generic object serializer/deserializer for WebWorker transfers
See [sketch here](https://github.com/mapbox/mapbox-gl-js/compare/web-worker-transfer).

The idea is to replace all our manually-defined `serialize` methods with a generic serializer function that recursively serializes the own-properties of an object and records the constructor name, and all our manually-defined `deserialize` methods with a generic deserializer that looks up the constructor name in a class registry, then uses `Object.create` to reconstitute an instance without triggering the constructor.

Properties that should not be serialized (e.g. cached computed values) shall be marked as non-enumerable.

In the sketch above, `Tranferable`s that should be transferred can be flagged by setting a `transfer: true` property on the `Transferable`. But I think it would be better to make `serialize` a transfer-only operation, and require you to copy ahead of time of any `Transferable`s you need to copy.

anandthakker:
Closes #5143

- [x] Add serialize, deserialize, register
- [x] Replace Bucket#serialize() with generic serialization
- [x] Fixup https://github.com/mapbox/mapbox-gl-js/pull/5702/commits/bbcedde13245729fd41ecc7f9d031f052897d968 to include compiled stop functions
- [x] Make FeatureIndex serializable
- [x] Remove `StructArray#serialize()`
- [x] <del>? Use generic serialization for `setLayers` message</del> (this is pretty involved, punt for now)
- [x] Update unit tests
- [x] Benchmark, profile, recover performance

Benchmarks (no significant change)
https://bl.ocks.org/anonymous/raw/5858aa59155120157d4ba4b938835434/

### DIFF
--- a/src/data/bucket/heatmap_bucket.js
+++ b/src/data/bucket/heatmap_bucket.js
@@ -1,6 +1,9 @@
 // @flow
 
 const CircleBucket = require('./circle_bucket');
+const {register} = require('../../util/web_worker_transfer');
+
+import type StyleLayer from '../../style/style_layer';
 
 const heatmapInterface = {
     layoutAttributes: CircleBucket.programInterface.layoutAttributes,
@@ -11,7 +14,12 @@ const heatmapInterface = {
     ]
 };
 
-class HeatmapBucket extends CircleBucket {}
+class HeatmapBucket extends CircleBucket {
+    // Needed for flow to accept omit: ['layers'] below, due to https://github.com/facebook/flow/issues/4262
+    layers: Array<StyleLayer>;
+}
+
+register(HeatmapBucket, {omit: ['layers']});
 
 HeatmapBucket.programInterface = heatmapInterface;
 
--- a/src/data/feature_index.js
+++ b/src/data/feature_index.js
@@ -12,12 +12,11 @@ const Protobuf = require('pbf');
 const GeoJSONFeature = require('../util/vectortile_to_geojson');
 const arraysIntersect = require('../util/util').arraysIntersect;
 const TileCoord = require('../source/tile_coord');
+const {register} = require('../util/web_worker_transfer');
 
 import type CollisionIndex from '../symbol/collision_index';
 import type StyleLayer from '../style/style_layer';
-import type {SerializedStructArray} from '../util/struct_array';
 import type {FeatureFilter} from '../style-spec/feature_filter';
-import type {Transferable} from '../types/transferable';
 
 const FeatureIndexArray = createStructArrayType({
     members: [
@@ -45,14 +44,6 @@ type QueryParameters = {
     sourceID: string
 }
 
-export type SerializedFeatureIndex = {
-    coord: TileCoord,
-    overscaling: number,
-    grid: ArrayBuffer,
-    featureIndexArray: SerializedStructArray,
-    bucketLayerIDs: Array<Array<string>>
-}
-
 class FeatureIndex {
     coord: TileCoord;
     overscaling: number;
@@ -69,21 +60,6 @@ class FeatureIndex {
     sourceLayerCoder: DictionaryCoder;
 
     collisionIndex: CollisionIndex;
-
-    static deserialize(serialized: SerializedFeatureIndex,
-                       rawTileData: ArrayBuffer) {
-        const coord = serialized.coord;
-        const self = new FeatureIndex(
-            new TileCoord(coord.z, coord.x, coord.y, coord.w),
-            serialized.overscaling,
-            new Grid(serialized.grid),
-            new FeatureIndexArray(serialized.featureIndexArray));
-
-        self.rawTileData = rawTileData;
-        self.bucketLayerIDs = serialized.bucketLayerIDs;
-
-        return self;
-    }
 
     constructor(coord: TileCoord,
                 overscaling: number,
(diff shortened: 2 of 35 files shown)


anandthakker:
Originally, I'd hoped to centralize serialization/deserialization in `Dispatcher` or `Actor`, but that's not quite possible yet (and I'm not actually 100% sure it's a good idea)

src/style-spec/expression/index.js
```diff
@@ -31,10 +33,72 @@ export type GlobalProperties = {
     heatmapDensity?: number
 };
 
-export type StyleExpression = {
-    evaluate: (globals: GlobalProperties, feature?: Feature) => any,
-    parsed: Expression
-};
+class StyleExpression {
+    expression: Expression;
+    propertySpec: StylePropertySpecification;
```
jfirebaugh: Unused.

src/style-spec/expression/index.js
```diff
@@ -31,10 +33,72 @@ export type GlobalProperties = {
...
+        }
+
+        this._evaluator.globals = globals;
+        this._evaluator.feature = feature;
+        return this.expression.evaluate(this._evaluator);
+    }
+}
+
+class StyleExpressionWithErrorHandling extends StyleExpression {
+    defaultValue: Value;
```
jfirebaugh: Prefix with underscore?

src/style-spec/expression/index.js
```diff
@@ -60,70 +124,75 @@ function createExpression(expression: mixed,
...
-        const warningHistory: {[key: string]: boolean} = {};
-        const defaultValue = getDefaultValue(propertySpec);
-        let enumValues;
-        if (propertySpec.type === 'enum') {
-            enumValues = propertySpec.values;
+        return success(new StyleExpressionWithErrorHandling(parsed, propertySpec));
+    }
+}
+
+class ZoomConstantExpression<Kind> {
```
jfirebaugh: I'd prefer to convert `ConstantExpression`, `SourceExpression`, etc. to classes, remove `kind`, and use `instanceof` for refinement.

src/style-spec/expression/index.js
```diff
@@ -164,26 +233,50 @@ function createPropertyExpression(expression: mixed,
...
+// serialization wrapper for old-style stop functions normalized to the
+// expression interface
+class StylePropertyFunction<T> {
+    _parameters: PropertyValueSpecification<T>;
+    _specification: StylePropertySpecification;
+
+    kind: 'constant' | 'source' | 'camera' | 'composite';
+    evaluate: (globals: GlobalProperties, feature?: Feature) => any;
+    interpolationFactor: ?(input: number, lower: number, upper: number) => number;
+    zoomStops: ?Array<number>
```
jfirebaugh: `;`

src/data/feature_index.js
```diff
@@ -70,25 +61,11 @@ class FeatureIndex {
...
-        self.bucketLayerIDs = serialized.bucketLayerIDs;
-
-        return self;
-    }
-
     constructor(coord: TileCoord,
                 overscaling: number,
                 grid?: Grid,
                 featureIndexArray?: FeatureIndexArray) {
+        require('assert')(coord instanceof TileCoord);
```
jfirebaugh: Is this leftover debug code?

src/style/query_utils.js
```diff
@@ -11,8 +12,15 @@ function getMaximumPaintValue(property: string, layer: StyleLayer, bucket: Bucke
     if (value.kind === 'constant') {
         return value.value;
     } else {
-        return (bucket: any).programConfigurations.get(layer.id)
-            .paintPropertyStatistics[property].max;
+        const binders = (bucket: any).programConfigurations.get(layer.id).binders;
+        for (const name in binders) {
```
jfirebaugh: 😱  Why isn't `binders` keyed on property name? I remember trying this before but running into some issue that I don't remember now.

src/data/bucket/heatmap_bucket.js
```diff
@@ -11,7 +14,11 @@ const heatmapInterface = {
     ]
 };
 
-class HeatmapBucket extends CircleBucket {}
+class HeatmapBucket extends CircleBucket {
+    layers: Array<StyleLayer>;
```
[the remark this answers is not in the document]
anandthakker: It's needed for flow to accept `omit: ['layers']` below (`omit` is typed `Array<$Keys<T>>`, and $Keys apparently doesn't include super's keys: https://github.com/facebook/flow/issues/4262)

src/data/feature_index.js
anandthakker: Oops, thanks

src/style/query_utils.js
anandthakker: It's keyed on _attribute_ name rather than property name.

src/util/web_worker_transfer.js
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * @private
+ */
+function register<T: any>(klass: Class<T>, options: RegisterOptions<T> = {}) {
+    const name: string = klass.name;
+    assert(name);
+    assert(!registry[name], `${name} is already registered.`);
+    registry[name] = {
+        klass,
+        omit: (options.omit: any) || [],
```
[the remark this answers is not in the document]
anandthakker: Oh, nice 👍

src/style-spec/expression/index.js
anandthakker: I considered this, but backed off from it mostly because normalizing stop functions to expressions would be sort of weird -- i.e., would we have `class XxxFunction extends XxxExpression`?   (Also, though more minor, `CameraExpression` and `CompositeExpression` would be identical (but distinct) classes)

src/style/query_utils.js
anandthakker: From a surface look, I don't see any obvious reason why we couldn't key it by property name instead.

anandthakker:
@jfirebaugh should be good to go

Issue #5172, opened 2017-08-19
QueryRenderedFeatures and icon-allow-overlap


ttsirkia:
**mapbox-gl-js version**: 0.39.1

### Steps to Trigger Behavior

 1. a symbol with an icon assigned to it and `icon-allow-overlap` is `true`
 2. another layer which draws a text label for the symbol, set a different min-zoom
 3. when the label becomes visible, the original symbol is not returned anymore by `queryRenderedFeatures` although still visible (symbol would be hidden in this case without `icon-allow-overlap`)

![kuva](https://user-images.githubusercontent.com/2797729/29484966-58456a12-84d2-11e7-8373-f6c179cb0518.png)

### Expected Behavior

`queryRenderedFeatures` returns always the visible elements

### Actual Behavior

`queryRenderedFeatures` acts as `icon-allow-overlap` is `false` and ignores some visible elements

Erutan409:
I'm also encountering this EXACT same issue.  Is there any headway on looking into this?

anandthakker:
We should investigate this bug after #5150 lands in master

ChrisLoer:
@ttsirkia or @Erutan409 can you test this with the latest version of `master` now that #5150 has landed? Or provide a JSFiddle showing the problem?

ttsirkia:
@ChrisLoer  I'll try to reproduce the issue soon and report my findings.

ttsirkia:
I can still replicate the bug but the behavior is a bit different. Now the rendered icon is sometimes returned and sometimes not depending on the map position. If I drag the map, I get different results although the icon is always visible. I have to see if I can create a minimal JSFiddle.

ChrisLoer:
@ttsirkia Thanks! I can take a look once you've got the fiddle.

Erutan409:
@ChrisLoer I think I "solved" my issue by dynamically adding the icons after loading the style.  So, I'm not experiencing it anymore.  But, that's not to say the root cause is fixed.  And I unfortunately can't create a test case with the data/style I'm using.  I'm not at liberty to make public until it's actually approved to merge into production.  Sorry :\

ttsirkia:
Here https://jsfiddle.net/0b9m5y2e/ is a simple example to reproduce the issue. Zoom in and see how the counter becomes zero.

ChrisLoer:
[...]

PR #5703, opened 2017-11-17
Include bucketIndex in queryRenderedSymbols duplicate feature filter

ChrisLoer:
Fixes issue #5172: items from multiple layers but sharing a common source feature would only show up in queryRenderedSymbols results for one of the layers.

Although this doesn't cause any of our existing query tests to fail, I worry a _little_ that we might break assumptions for some users of `queryRenderedSymbols` with this change. Before, we'd never return the same feature twice in a result set -- now it's possible to return the feature multiple times (once for each layer it's part of). Should we change maybe something like `Style#_flattenRenderedFeatures` to remove duplicates?

/cc @mollymerp @anandthakker

### DIFF
--- a/src/symbol/collision_index.js
+++ b/src/symbol/collision_index.js
@@ -267,13 +267,18 @@ class CollisionIndex {
             const blocking = collisionBoxArray.get(thisTileFeatures[i]);
             const sourceLayer = blocking.sourceLayerIndex;
             const featureIndex = blocking.featureIndex;
+            const bucketIndex = blocking.bucketIndex;
 
             // Skip already seen features.
             if (sourceLayerFeatures[sourceLayer] === undefined) {
                 sourceLayerFeatures[sourceLayer] = {};
             }
-            if (sourceLayerFeatures[sourceLayer][featureIndex]) continue;
-
+            if (sourceLayerFeatures[sourceLayer][featureIndex] === undefined) {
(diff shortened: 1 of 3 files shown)


ttsirkia:
Docs state: "Because features come from tiled vector data or GeoJSON data that is converted to tiles internally, feature geometries may be split or duplicated across tile boundaries and, as a result, features may appear multiple times in query results." Actually, the newest versions do not have this behavior as for example 0.39 had.

src/symbol/collision_index.js
```diff
@@ -267,13 +267,18 @@ class CollisionIndex {
...
                 sourceLayerFeatures[sourceLayer] = {};
             }
-            if (sourceLayerFeatures[sourceLayer][featureIndex]) continue;
-
+            if (sourceLayerFeatures[sourceLayer][featureIndex] === undefined) {
+                sourceLayerFeatures[sourceLayer][featureIndex] = {};
+            }
+            if (sourceLayerFeatures[sourceLayer][featureIndex][bucketIndex]) {
+                continue;
+            }
```
anandthakker: Since there are generally fewer bucket indices than feature indices, would it be (slightly) beneficial to switch the order to `sourceLayerFeatures[sourceLayer][bucketIndex][featureIndex]`?

ChrisLoer: 🤔 maybe? I wouldn't expect it to make too big a difference...

PR #5704, opened 2017-11-17
"Clear" stencil buffer manually; workaround for #5490

### LINKED ISSUE #5490: Map blink when zooming
When I zoom the map, map became blinking.

**mapbox-gl-js version**:

Mapbox Studio version

MacBook Air 2017
macOS High Sierra
Chrome dev 63

### Steps to Trigger Behavior

 1. login Mapbox Studio
 2. Create a dark map
 3. Zooming

### Expected Behavior

Zooming would be smooth

### Actual Behavior
qq20171020-165813-hd

lbud:
As of Chrome v62, depth reads in certain Intel Iris graphics cards (at least Intel Iris Graphics 6100) all fail on intermittent frames, creating a flashing effect as in https://github.com/mapbox/mapbox-gl-js/issues/5490. I believe this bug is some kind of race condition or incorrect vector access related to stencil buffer clearing. This PR mitigates this issue pending an upstream fix by never using `gl.clear(gl.STENCIL_BUFFER_BIT)` and instead "clearing" it manually by drawing a stencil mask of all 0s to the full viewport. I've confirmed on maps where this effect is most dramatic (maps with many sources) that this fix eliminates flashing entirely.

Fixes #5490.
Upstream issue: https://bugs.chromium.org/p/chromium/issues/detail?id=784030

### DIFF
--- a/src/render/painter.js
+++ b/src/render/painter.js
@@ -219,9 +219,40 @@ class Painter {
      */
     clearStencil() {
         const gl = this.gl;
-        gl.clearStencil(0x0);
+
+        // As a temporary workaround for https://github.com/mapbox/mapbox-gl-js/issues/5490,
+        // pending an upstream fix, we draw a fullscreen stencil=0 clipping mask here,
+        // effectively clearing the stencil buffer: restore this code for native
+        // performance and readability once an upstream patch lands.
+
+        // gl.clearStencil(0x0);
+        // gl.stencilMask(0xFF);
+        // gl.clear(gl.STENCIL_BUFFER_BIT);
+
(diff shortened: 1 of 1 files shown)


SBRK:
Does this impact performance ?

lbud:
@SBRK in theory it _might_ impact performance ever so slightly but it is a necessary workaround to a critical upstream bug.

SBRK:
@lbud yep I know we had this problem and it was affecting half the computers in our company :) But it seemed to me the performance was slightly worse after updating. It could just be an impression though

PR #5707, opened 2017-11-17
make buildings grow

mollymerp:
converted to expressions and also removed `findIndex` to restore IE support per https://github.com/mapbox/mapbox-gl-js/pull/5537#discussion_r148134674


mtirwin:
🙏  y'all. I get that we're mixing two concepts here, but I like this approach as best practice.

PR #5721, opened 2017-11-20
Fix rendering in Safari

jfirebaugh:
Previously, tile clipping masks were drawn with an empty `ProgramConfiguration`, which when used with the "fill" shader, rendered with color and opacity attributes that had not been set to a definite value. We instead apparently want to draw with color and opacity _uniforms_ that aren't set to a definite value. 🤷‍♂️


jfirebaugh:
👍, switched.

mourner:
@jfirebaugh quick question — I remember that we previously avoided creating programConfiguration dynamically on every frame, and used a cached value for things like this instead for performance. Was it intentional to switch to dynamically created ones?

PR #5573, opened 2017-11-01
Use Color ubiquitously

jfirebaugh:
Fixes #5410.

PR #5739, opened 2017-11-22
Introduce GL state tracking

lbud:
This PR introduces GL state tracking (similar to gl-native behavior) as a first step in https://github.com/mapbox/mapbox-gl-js/issues/145. 

Updated benchmark results: [link](https://bl.ocks.org/anonymous/raw/a667bcd4143795417ff548e16501dc4c/)
image

## Remaining

- [x] unit tests
- [x] re-run benchmarks once everything is ready


src/gl/index_buffer.js
```diff
@@ -29,30 +31,34 @@ class IndexBuffer {
         // The bound index buffer is part of vertex array object state. We don't want to
         // modify whatever VAO happens to be currently bound, so make sure the default
         // vertex array provided by the context is bound instead.
-        if (this.gl.extVertexArrayObject === undefined) {
-            (this.gl: any).extVertexArrayObject = this.gl.getExtension("OES_vertex_array_object");
```
jfirebaugh: Glad to eliminate this `any`.

src/gl/index_buffer.js
```diff
@@ -29,30 +31,34 @@ class IndexBuffer {
...
+            this.context.bindVertexArrayOES.set(null);
         }
     }
 
     bind() {
-        this.gl.bindBuffer(this.gl.ELEMENT_ARRAY_BUFFER, this.buffer);
+        const gl = this.context.gl;
+        gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, this.buffer);
+        // this.context.bindElementBuffer.set(this.buffer);
+        // TODO not sure why this doesn't work, but using bindElementBuffer throws:
```
jfirebaugh: My hunch would be that a necessary call to `gl.bindBuffer()` is getting incorrectly deduped away somehow.

src/gl/state.js
```diff
@@ -0,0 +1,35 @@
...
+        this.value = v;
+        this.current = v.constructor.default(v.context);
+    }
+
+    get(): T {
+        return this.current;
+    }
+
+    set(t: T) {
+        if (this.current !== t) {
```
jfirebaugh: We have to be careful about using the `!==` operator here. It will always use object identity when `T` is an object or array type like `BlendFuncType` or `StencilFuncType`. That probably isn't what we want. We have a couple deep equal implementations that might work, or we could add an `equal(a: T, b: T): boolean` method to the `Value` interface.

src/gl/types.js
```diff
@@ -0,0 +1,60 @@
...
+    | $PropertyType<WebGLRenderingContext, 'ONE_MINUS_DST_ALPHA'>
+    | $PropertyType<WebGLRenderingContext, 'CONSTANT_COLOR'>
+    | $PropertyType<WebGLRenderingContext, 'ONE_MINUS_CONSTANT_COLOR'>
+    | $PropertyType<WebGLRenderingContext, 'CONSTANT_ALPHA'>
+    | $PropertyType<WebGLRenderingContext, 'ONE_MINUS_CONSTANT_ALPHA'>
+    | $PropertyType<WebGLRenderingContext, 'BLEND_COLOR'>;
+
+export type BlendFuncType = [BlendFuncConstant, BlendFuncConstant];
+
+export type ColorType = [number, number, number, number];
```
jfirebaugh: Follow #5573 and use `Color` instead.

yarn.lock
```diff
@@ -188,11 +188,7 @@
   version "3.0.0"
   resolved "https://registry.yarnpkg.com/@mapbox/whoots-js/-/whoots-js-3.0.0.tgz#c1de4293081424da3ac30c23afa850af1019bb54"
 
-"@types/node@*":
```
jfirebaugh: Are there actual dependency updates in here or is this an accidental change (different version of yarn maybe?)

lbud:
I rolled back use of `context.bindTexture` after debugging a few failing render tests because the bound texture state is also dependent on the active texture unit — I'm going to address this in my next PR (focusing on framebuffer + renderbuffer + texture state).

Issue #5716, opened 2017-11-20
symbol layer fades in upon setData


andrewharvey:
**mapbox-gl-js version**: git bisect reveals this is a regression in https://github.com/mapbox/mapbox-gl-js/commit/135561825addae07b253bab2167cee72f00f3ac9 (#5150)

When updating a source with setData any symbol layers using that source will fadein again after the setData. This causes the symbol to flicker when setData is fired frequently such as from a cursor event.

See this in action at https://jsbin.com/kezucog/edit?html,output (just move the cursor around the screen). This wasn't a problem in 0.41.0, but is a problem with 0.42.0 and I pinpointed it to the commit mentioned.

I can see that GL JS now does cross source symbol placement so previously the "Mapbox DC" (from the example) would overlap existing labels, now it will actually fade out the other labels to make room for it. To get the 0.41 behaviour back it's just a matter of using `            "text-ignore-placement": true, "icon-ignore-placement": true` in that layer's layout properties. However even with this, the label still fades in again every time it is updated.

/cc @ChrisLoer

ryanbaumann:
I'm seeing this too as described by @andrewharvey .  Examples:

1. [0.41.0 example](https://bl.ocks.org/ryanbaumann/9b9b52e09ff86d1ce8346fb76b681427/2e18829c0faa4bc4ffe80917121295fb5c4785e4) - setData() on a symbol layer animates symbols on setData() without fade.
2. [0.42.0 example](https://bl.ocks.org/ryanbaumann/9b9b52e09ff86d1ce8346fb76b681427/1cc5b3c3fea1cd06e61b4401f148b079e61796ba) - setData() on a symbol layer causes significant icon-image fade.

![](https://cl.ly/2A1m0o2s1r18/download/Screen%20recording%202017-11-20%20at%2007.37.54%20AM.gif)

ChrisLoer:
@ansis This sounds like it's probably another issue with how `CrossTileSymbolIndex` updates when a tile reload happens (like #5491, except this time it's from `setData` instead of `setFilter`)? I'm not sure why the fix for #5491 isn't also working here, but this seems like something that the back-port from native would address. Do you see a quick fix we can do now for this, or is the back-port coming so soon it doesn't matter?

@ryanbaumann I shouldn't make light of broken functionality, but that fleet of ghost-planes crossing the South Pacific is... _kind of awesome_.

lbud:
[...]

PR #5741, opened 2017-11-22
skip symbol fading for tiles immediately after reload

ansis:
fix https://github.com/mapbox/mapbox-gl-js/issues/5716

The new symbol fading was causing flickering for symbols animated with `setData`. When the new location of the symbol was far enough away from the old it would not identify them as being the same icon, resulting in fading. Solving this properly would include figuring out a good way to fix this matching problem.

Instead, this commit works around the problem by skipping fading for any tiles that have just been reloaded.

### DIFF
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -81,6 +81,7 @@ class Tile {
     texture: any;
     refreshedUponExpiration: boolean;
     reloadCallback: any;
+    justReloaded: boolean;
 
     /**
      * @param {TileCoord} coord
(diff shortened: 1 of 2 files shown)


jfirebaugh:
LGTM, but tagging @ChrisLoer since he's more familiar with this.

ChrisLoer:
I filed https://github.com/mapbox/mapbox-gl-native/issues/10536 for the native side. It occurs to me that maybe within a single tile we could do a better job of detecting duplicates based on ID or feature index.

I looked at @ryanbaumann's example, and the points don't have IDs, but they do have stable indices in each call to setData -- so at least in that case, an algorithm that said "you're a duplicate if your index/id matches the previous index/id" could work.

Issue #4670, opened 2017-05-03
Popups and markers lag behind the map during pan animation


bicubic:
Issue can be seen at https://www.mapbox.com/mapbox-gl-js/example/popup/
Perform a pan gesture with the mouse and lift the mouse button while still moving the cursor so the map continues to pan for a bit. Alternatively, use the arrow buttons to pan the map. 

During the animation, the popup appears to be located a few pixels (one frame?) behind the map. Quite jarring on touch screen experiences.

kkaefer:
Here's a slowed down version of the bug.
![out2](https://cloud.githubusercontent.com/assets/52399/25662819/30c3fb18-3016-11e7-81f0-16987f787ead.gif)

dagjomar:
This is true not only for popups for markers also. Generally, things that are DOM elements placed on top of the rendered canvas.

bicubic:
Tested synchronizing a pop-up to a a moving element today, there was no latency. So it looks like the render timing and DOM update are synchronized properly. The issue may be due to popup update fetching viewport bounds prior to them being updated during that frame. I'm just making guesses at the moment, don't have any grasp of the mapbox codebase yet.

bicubic:
Looks like it may be a timing issue after all. The DOM appears to be a frame ahead of the canvas when the pan easing animation is in effect.
 
The attached gist is a proof of concept which injects one frame of latency to the popup transform. The latency is added only if the `_update()` was called from `browser.tick()` and not if it originated from a mouse event. 

It doesn't completely solve the misalignment but does significantly improve it for me. I believe the remainder of the misalignment may be due to the `camera._ease` logic which is panning the camera by sub-pixel amounts. During mouse panning, the `point` is always integer. I can't figure out how to fix the remaining error, but hopefully this gives some clue where to look. 

https://gist.github.com/bicubic/272f56341b55689982c60e32c7b28242

bicubic:
Please take a look at #4681

I was able to get popups fully aligned with two changes:

DOM transform updates are not applied immediately but buffered via new function `util.pushTransform()` and applied via `util.applyTransforms()` which is invoked by `painter.render()`. This ensures that layout changes happen at the same time as canvas painting. This involved a change to `DOM.setTransform()`, but it is backwards compatible and callers must explicitly pass in a new parameter to buffer the change, otherwise it is invoked immediately as before. 

[...]

PR #5742, opened 2017-11-22
Fix marker/popup lag by firing move events 1 frame earlier

mourner:
This changes the camera event logic so that `move`/`zoom`/`rotate`/`pitch` events fire before the transform is updated (instead of after). Miraculously, this fixes the longstanding issue with lagging markers/popups on animations (#4670, #4681). 

I'm not entirely sure why, but it seems like the logic that makes the map canvas reflect the transform state happens before the frame handlers for camera animations, which means that when transform is changed in a frame as a part of an animation, it won't be reflected until the next frame while changes in marker positions are reflected instantly, hence the off-by-one-frame sync issues.

@jfirebaugh maybe you have an idea of how to fix that on a fundamentally upper level? Or we could accept this fix for now — it shouldn't be breaking.

cc @bicubic

## Launch Checklist

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -562,6 +562,8 @@ class Camera extends Evented {
         clearTimeout(this._onEaseEnd);
 
         this._ease(function (k) {
+            this._fireMoveEvents(eventData);
+
             if (this.zooming) {
                 tr.zoom = interpolate(startZoom, zoom, k);
             }
@@ -583,8 +585,6 @@ class Camera extends Evented {
                 const newCenter = tr.unproject(from.add(delta.mult(k * speedup)).mult(scale));
                 tr.setLocationAtPoint(tr.renderWorldCopies ? newCenter.wrap() : newCenter, pointAtOffset);
             }
-
-            this._fireMoveEvents(eventData);
 
         }, () => {
             if (options.delayEndEvents) {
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Won't changing the order leave the final position of popups and markers one frame off from the desired position? On the final easing frame, the `move` event will be fired, the popup will update to the current position, then the transform will be updated. Then there's no subsequent `move` event to update the overlays to the final position (`_easeToEnd` only fires `moveend`).

I think the root cause here is that camera easing and map rendering happen on separate `requestAnimationFrame` loops, so relative timing is at the whim of the browser implementation. In native, by contrast, updating the camera during an ease is something that is integrated into the render loop:

https://github.com/mapbox/mapbox-gl-native/blob/e3f1d8e3a0beb648ec90ac4d9baa5f27c6cf3935/src/mbgl/map/map.cpp#L753

https://github.com/mapbox/mapbox-gl-native/blob/e3f1d8e3a0beb648ec90ac4d9baa5f27c6cf3935/src/mbgl/map/map.cpp#L203-L205

mourner:
> Then there's no subsequent move event to update the overlays to the final position (_easeToEnd only fires moveend).

Marker and popup update their position on `moveend` as well, so it's not a problem. But I agree that your solution is better.

mourner:
-> #5743

PR #5743, opened 2017-11-22
Integrate camera easing into main render loop

### LINKED ISSUE #4670: Popups and markers lag behind the map during pan animation

jfirebaugh:
Fixes #4670. Alternative to #5742.

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -72,9 +72,12 @@ class Camera extends Evented {
 
     _bearingSnap: number;
     _onEaseEnd: number;
-    _abortFn: any;
-    _finishFn: any;
+    _easeStart: number;
+    _easeOptions: {duration: number, easing: (number) => number};
+    _easeFn: (number) => void;
+    _finishFn: () => void;
     _prevEase: any;
+    +_update: () => void;
 
     constructor(transform: Transform, options: {bearingSnap: number}) {
         super();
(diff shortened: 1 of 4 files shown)


andy-mcdonald:
@jfirebaugh - Apologies for asking, but can you provide any timeline for when the next version (v0.42.3?) of `mapbox-gl` will be released? My team is working on a Mapbox-powered mobile web application (for a high profile client) that is due to launch on December 16th and we really need Markers to be stable.

I tried to run the build scripts from the `master` branch and manually patch them into the dist folder within the node module, but `mapbox-gl.js` is throwing an `Uncaught Error: can't deserialize object of anonymous class` and the map tiles are not displayed. Swapping in `mapbox-gl-dev.js` doesn't seem to have the new fix. Any ideas how I can get this working?

Again, sorry for pushing on this - I realize it's not practical to release a new version for every issue that gets resolved.

bicubic:
I'm not 100% sure yet, but I think this change caused a regression where the `map.transform` updates lag by 1 frame behind the `update` event only during camera easing. I have an application that works fine on 0.34.0 but transforms are 1 frame late during easing on 0.43.0. 

@mourner Can you please help me figure out where map.transform is being currently updated so I can investigate further?

mourner:
@bicubic mostly in https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/camera.js, but also in some handler code like https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/handler/drag_pan.js#L126

PR #5745, opened 2017-11-23
Only call update if no other source prop other than data differs

### LINKED ISSUE #5731: setStyle diffing does not work if source gets clustered on the fly
**mapbox-gl-js version**: 0.42.1
### Steps to Trigger Behavior

 1. have a map with a source with cluster false
 2. set the source to cluster true on the fly after the map is initialised
 3. apply the new style on the map with setStyle

### Expected Behavior
The points should appear clustered

### Actual Behavior
The points do not appear clustered

If I start with cluster: true from the beginning all is fine. Also disabling the style diffing kind of works (ofcourse I get flashes on the map, but the clustering behaviour is as I would expect).

bartvde:
This PR fixes #5731

### DIFF
--- a/src/style-spec/diff.js
+++ b/src/style-spec/diff.js
@@ -100,6 +100,36 @@ const operations = {
 
 };
 
+function addSource(sourceId, after, commands) {
+    commands.push({ command: operations.addSource, args: [sourceId, after[sourceId]] });
+}
+
+function removeSource(sourceId, commands, sourcesRemoved) {
+    commands.push({ command: operations.removeSource, args: [sourceId] });
+    sourcesRemoved[sourceId] = true;
+}
+
+function updateSource(sourceId, after, commands, sourcesRemoved) {
+    removeSource(sourceId, commands, sourcesRemoved);
+    addSource(sourceId, after, commands);
+}
+
+function canUpdateGeoJSON(before, after, sourceId) {
+    let prop;
+    for (prop in before[sourceId]) {
+        if (!before[sourceId].hasOwnProperty(prop)) continue;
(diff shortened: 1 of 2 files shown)


bartvde:
Good point, will adapt

bartvde:
@mourner I added https://github.com/mapbox/mapbox-gl-js/pull/5745/commits/a6f57dae933ede6422b22dcca380abd711d35b85 but I'm unsure if it's okay to use union from lodash

bartvde:
thanks for the guidance @mourner I updated the PR

src/style-spec/diff.js
```diff
@@ -111,25 +141,25 @@ function diffSources(before, after, commands, sourcesRemoved) {
...
+            addSource(sourceId, after, commands);
         } else if (!isEqual(before[sourceId], after[sourceId])) {
             if (before[sourceId].type === 'geojson' && after[sourceId].type === 'geojson') {
-                // geojson sources use setGeoJSONSourceData command to update
-                commands.push({ command: operations.setGeoJSONSourceData, args: [sourceId, after[sourceId].data] });
+                if (canUpdateGeoJSON(before, after, sourceId)) {
+                    commands.push({ command: operations.setGeoJSONSourceData, args: [sourceId, after[sourceId].data] });
+                } else {
+                    updateSource(sourceId, after, commands, sourcesRemoved);
+                }
```
mourner: Nit: you can move the `canUpdateGeoJSON` condition to the parent `if`, which will remove the need for the `else` above.

bartvde: good point will do

bartvde: changes made

Issue #5758, opened 2017-11-28
add setter for offset to marker


bartvde:
Currently you can only provide offset for a marker in the constructor. I'm creating a div as the element for the marker, and inside of this div a React popup component is rendered. I only know the size after addTo is called on the marker, since it will only get added tot the DOM at that point, which is too late to know the offset if it can only be provided at construction time. I'm able to work around this by using private variables [1] but I'd rather have a public method for updating the offset of a marker dynamically.

[1]
```
marker._offset = offset;
marker._update();
```

Just for reference, this is the full code:

```
  addPopup(popup) {
    const id = uuid.v4();
    const elem = document.createElement('div');
    elem.setAttribute('class', 'sdk-mapbox-gl-popup');
    const overlay = new mapboxgl.Marker(elem);
    const self = this;
    // render the element into the popup's DOM.
    ReactDOM.render(popup, elem, (function addInstance() {
      self.popups[id] = this;
      self.elems[id] = elem;
      this.setMap(self);
    }));

    // set the popup id so we can match the component
    //  to the overlay.
    overlay.popupId = id;
    const coord = popup.props.coordinate;
    const lngLat = new mapboxgl.LngLat(coord['0'], coord['1']);
    this.overlays.push(overlay.setLngLat(lngLat).addTo(this.map));
    const size = ReactDOM.findDOMNode(elem).getBoundingClientRect();
    const yTransform = size.height / 2 + 11;
    const xTransform = size.width / 2 - 48;
    const offset = new mapboxgl.Point.convert([xTransform, -yTransform]);
    // TODO do not use mapbox internals here
    overlay._offset = offset;
    overlay._update();
  }
```

I'd be willing to work on a PR for this ofcourse if the idea is accepted.

mourner:
Sounds good! A PR would be welcome.

PR #5759, opened 2017-11-28
Add getter and setter for offset on marker

bartvde:
This adds a setter (and getter) for offset on marker as discussed in #5758


bartvde:
thanks for the quick turn-around on the PRs!

Issue #5755, opened 2017-11-27
Coalesce expression doesn't work as expected


muesliq:
**mapbox-gl-js version**: 0.42.2

### Steps to Trigger Behavior

Use this code:

`layout: {"text-field": ["coalesce", ["get", "fieldA"], ["get", "fieldB"]]}
`

where fieldA is null and fieldB is not null

### Expected Behavior

Display fieldB.

### Actual Behavior

Empty label. (It works though when fieldA is not null, displaying fieldA.)

### Note: 

A work-around with `case` did the job. However this seems to be a bug.  Sorry if I somehow misunderstood the Style Spec and it isn't!

anandthakker:
Thanks for the report @muesliq!  Could you by any chance provide a live jsfiddle example reproducing this issue?

jfirebaugh:
This is an outcome of type inference and automatic type assertions. What's happening is that since `text-field` must be a string, the result of `coalesce` must be a string, so it's inferred that each of the two `get`s must also be a string. So there is a `string` assertion automatically inserted, as if you had written:

```
["coalesce", ["string", ["get", "fieldA"]], ["string", ["get", "fieldB"]]]
```

Then when the expression is evaluated, it's found that the first `get` produces `null` rather than a string, which causes an error and aborts the evaluation. (The visual result is that you get the default value of the property, an empty string.)

This is _arguably_ working as intended, but I think the behavior is so surprising that we will need to revisit the design.

anandthakker:
Nice analysis @jfirebaugh.

I think the best fix here is to treat `coalesce` as a special case when performing inference, not inserting assertions for its arguments, and instead inserting an assertion around the whole coalesce expression where necessary.

PR #5765, opened 2017-11-29
Omit inferred type annotations for 'coalesce' arguments

anandthakker:
As described in #5755, such inferred annotations cause a runtime type error for `null` inputs, thus preempting the desired null-coalescing behavior of something like `["coalesce", ["get", "a"], ["get", "b"]]`.

### DIFF
--- a/src/style-spec/expression/definitions/coalesce.js
+++ b/src/style-spec/expression/definitions/coalesce.js
@@ -1,6 +1,7 @@
 // @flow
 
 const assert = require('assert');
+const {checkSubtype, ValueType} = require('../types');
 
 import type { Expression } from '../expression';
 import type ParsingContext from '../parsing_context';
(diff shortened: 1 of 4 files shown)


anandthakker:
> Rather than wrapping and unwrapping, did you consider an approach where parse gets a flag from Coalesce.parse indicating not to insert inferred assertions?
 
I considered that, but `Coalesce.parse` also needs to know if an assertion "should" have been added, in order to know whether or not to wrap itself in an outer assertion.  Given that, it seemed simpler to contain the logic within `Coalesce.parse`.

Realizing now that an alternative would be to just _always_ wrap the coalesce in an outer assertion, even if none of its arguments are typed `value`.  Maybe that's better.

jfirebaugh:
👍  always wrapping with an outer assertion.

src/style-spec/expression/definitions/coalesce.js
```diff
@@ -21,18 +22,31 @@ class Coalesce implements Expression {
...
         assert(outputType);
-        return new Coalesce((outputType: any), parsedArgs);
+
+        // Above, we parse arguments without inferred type annotation so that
+        // they don't produce a runtime error for `null` input, which would
+        // preempt the desired null-coalescing behavior.
+        // Thus, if any of our arguments would have needed an annotation, we
+        // need to wrap the enclosing coalesce expression with it instead.
+        const needsAnnotation = expectedType &&
+            parsedArgs.some(arg => checkSubtype(expectedType, arg.type));
```
anandthakker: This bit isn't strictly necessary -- we could just always return `new Coalesce(ValueType, parsedArgs)` -- I figured including it would be more explicit & therefore clearer.

src/style-spec/expression/parsing_context.js
```diff
@@ -68,26 +81,36 @@ class ParsingContext {
             if (Expr) {
                 let parsed = Expr.parse(expr, context);
                 if (!parsed) return null;
-                const expected = context.expectedType;
-                const actual = parsed.type;
-                if (expected) {
-                    // When we expect a number, string, or boolean but have a
```
jfirebaugh: How about:

```js
                    // When we expect a number, string, boolean, or array but
                    // have a Value, we can wrap it in a refining assertion.
                    // When we expect a Color but have a String or Value, we
                    // can wrap it in "to-color" coercion.
                    // Otherwise, we do static type-checking.
                    if ((expected.kind === 'string' || expected.kind === 'number' || expected.kind === 'boolean') && actual.kind === 'value') {
                        if (!options.omitTypeAnnotations) {
                            parsed = new Assertion(expected, [parsed]);
                        }
                    } else if (expected.kind === 'array' && actual.kind === 'value') {
                        if (!options.omitTypeAnnotations) {
                            parsed = new ArrayAssertion(expected, parsed);
                        }
                    } else if (expected.kind === 'color' && (actual.kind === 'value' || actual.kind === 'string')) {
                        if (!options.omitTypeAnnotations) {
                            parsed = new Coercion(expected, [parsed]);
                        }
                    } else if (context.checkSubtype(context.expectedType, parsed.type)) {
                        return null;
                    }
```

anandthakker: 👌

anandthakker: Updated with this approach.

PR #5769, opened 2017-11-29
Bump geojson-rewind dependency

pelotom:
Currently Yarn complains

```ts
warning mapbox-gl > geojson-rewind > geojson-area@0.1.0: This module is now under the @mapbox namespace: install @mapbox/geojson-area instead
```

This dependency is fixed in the latest `geojson-rewind`.


jfirebaugh:
Thanks! Can you check in the updated yarn.lock as well?

pelotom:
Yep.

Issue #5762, opened 2017-11-29
`data` event from style mutations should be fired synchronously


jfirebaugh:
Currently, if you mutate the style via `setPaintProperty` or some other runtime styling API, the corresponding `data` event (with `dataType` property equal to `style`) is not fired until the subsequent repaint, which is asynchronous. I think this event should be fired immediately instead:

* The change is observable with respect to `getPaintProperty` or comparable accessor immediately, and it seems like the event should fire as soon as the change becomes observable.
* It's necessary for #2741: `Map` will use the `data` event as the consistent trigger for a new frame, rather than wrapping each `Style` method and calling `_update(true)` manually.
* The current behavior is causing each style mutation to trigger two repaints: one from the manual `_update(true)`, and a second one from the `data` event that's triggered during the first repaint.

jfirebaugh:
Another reason: if the `data` event is triggered from within `style.update()`, as it is now, calling `map.remove()` in response to that event causes a crash.

PR #5771, opened 2017-11-29
Don't fire sourcedata event with sourceDataType: 'metadata' when a source is removed

jfirebaugh:
Noticed this while working on #5762 -- this was added in #4842, but it seems odd to me to fire this event when the source itself isn't changing. Instead, have the LogoControl listen for the styledata event that's triggered when a source is removed.

### DIFF
--- a/src/ui/control/logo_control.js
+++ b/src/ui/control/logo_control.js
@@ -32,6 +32,7 @@ class LogoControl {
         this._container.appendChild(anchor);
         this._container.style.display = 'none';
 
+        this._map.on('styledata', this._updateLogo);
         this._map.on('sourcedata', this._updateLogo);
         this._updateLogo();
         return this._container;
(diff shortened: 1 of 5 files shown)


src/ui/control/logo_control.js
```diff
@@ -32,6 +32,7 @@ class LogoControl {
         this._container.appendChild(anchor);
         this._container.style.display = 'none';
 
+        this._map.on('styledata', this._updateLogo);
```
mollymerp: you'll also need to remove this event listener in the `onRemove` function

src/ui/control/logo_control.js
```diff
@@ -32,6 +32,7 @@ class LogoControl {
...
         this._map.on('sourcedata', this._updateLogo);
```
mollymerp: I think you are safe to remove this listener all together. Keeping it here and removing the `metadata` condition causes `_updateLogo` to be called on every tile load 🙃

jfirebaugh: I think this might still be necessary to ensure that the control updates when TileJSON is loaded.

mollymerp: good catch, thank you! it was working on the debug page because the spritesheet was loading after the tilejson.

jfirebaugh:
Stale; closing for now.

PR #5773, opened 2017-11-30
 Control the clock in integration tests

### LINKED ISSUE #3966: Add the ability to control the clock in integration tests
ref #3912

jfirebaugh:
By default, `browser.now()` always returns 0 in render tests. To control the clock, include a `wait` operation in the test with an argument for how long, in milliseconds, to advance the clock.

Fixes #3966. @ChrisLoer, I'm not sure whether your comment there is still applicable; which animations did you have in mind? I think I would argue that it's appropriate for `load` to fire when the _data_ is complete, even if some of the fades or transitions are still in progress. Where the test wants to avoid prematurely completing before that, it should either disable fading (which is mostly what it already does), or use the `wait` facility introduced here.

https://bl.ocks.org/anonymous/raw/5cfc21495d62134956471a8afca33476/

Speeds up `yarn test-render` a bit: 159.15s vs 130.90s.


jfirebaugh:
Note that Github is showing an incorrect ordering of commits above. Consult a local checkout for an order that makes more sense.

src/style/style.js
```diff
@@ -280,7 +280,7 @@ class Style extends Evented {
         const parameters = {
             zoom: z,
             now: browser.now(),
-            defaultFadeDuration: 300,
```
ChrisLoer: Just to make sure I'm reading this right: 

- The previous behavior was that transitions that didn't specify a duration would default to 300ms
- The new behavior is that transitions that don't specify a duration will inherit the map's (configurable) `fadeDuration`, which also controls collision fading, and which defaults to 300

ChrisLoer:
> I think I would argue that it's appropriate for load to fire when the data is complete, even if some of the fades or transitions are still in progress.

I was interpreting `loaded` as "has reached steady-state" which in my mind would have all animations completed, but basing it on data loading sounds reasonable too. In  that case, I think we can ignore my earlier comment.

src/style/style.js
jfirebaugh: Not quite -- this parameter affects only the cross-fading of `line-dasharray` and `*-pattern` when the zoom crosses an integer threshold. It does not affect property transitions; those are controlled by the top-level `transition` property in the style, overridden by individual `*-transition` properties.

ChrisLoer:
This looks 💯  to me -- and it looks to me like it should work as the basis for doing high-fidelity video (i.e. render one frame at a time with all assets loaded), which could potentially allow us to test not just a single point in a transition but a whole animated sequence?

Before I 🍏  let me play around with it a little bit tomorrow and try making my own transition test.

jfirebaugh:
The one gotcha with integration tests I've seen so far is that you'll typically need two `wait` events, due to the way that style changes are queued up and applied in batch on the subsequent frame. The first one triggers actually getting the style mutations you've queued up applied internally, and then the second one increments the clock.

PR #5774, opened 2017-11-30
in Style's addLayer and moveLayer the before param is optional

andrewharvey:
@jingsam Sorry I missed this when looking over #5679 before it was merged but did you mean to remove that `=`? Or am I missing something? This PR makes before optional again.

Inspecting the code, `before` is an optional param so the JSDoc should reflect this. Flow types already have it as optional.

Low impact since this is for the Style class not map.addLayer so doesn't affect end users.

/cc @mollymerp

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -547,7 +547,7 @@ class Style extends Evented {
     /**
      * Add a layer to the map style. The layer will be inserted before the layer with
      * ID `before`, or appended if `before` is omitted.
-     * @param {string} before  ID of an existing layer to insert before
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Documentation.js gives precedence to Flow annotations over JSDoc type information, so this wouldn't have a user-visible impact even if it was a public API. At some point we should do a pass over the JSDoc comments and remove all types.

andrewharvey:
I see.

I'm all for only have types defined in one place.

jingsam:
Sorry, I didn't aware `=?` has meaning in JsDoc.

Issue #5735, opened 2017-11-21
Expose map.getImage


alex3165:
Hello,

I would need to know if an image has already been added or not, would it be possible to expose a `map.getImage` function? I can see that `addImage` is already "protected":
```
  if (this.imageManager.getImage(id)) {
    return this.fire('error', {error: new Error('An image with this name already exists.')});
  }
``` 
But I would like to avoid the error message and prevent from calling `map.addImage` upfront by checking is the image is already there in my application.

**mapbox-gl-js version**: 0.42.2

ArturAntonov:
I found some error. And 'try-catch' not helping.
```
try {
          this.map.addImage('arrow-head', image);
        } catch (e) {
          console.log('image with id "arrow-head" is already exists');
        }

```
Even if i use 'try-catch' i still get error.
```
vendor.bundle.js:44292 Error: An image with this name already exists.
    at t.addImage (vendor.bundle.js:44178) [angular]
    at e.addImage (vendor.bundle.js:44266) [angular]
```
Is the method "addImage" async?

mollymerp:
@alex3165 we would gladly accept a PR implementing this method!

PR #5775, opened 2017-11-30
Expose getImage to public API

alex3165:
This PR expose `getImage` method from image manager to the public API. It should address https://github.com/mapbox/mapbox-gl-js/issues/5735.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -449,7 +449,7 @@ class Style extends Evented {
     }
 
     addImage(id: string, image: StyleImage) {
-        if (this.imageManager.getImage(id)) {
+        if (this.getImage(id)) {
             return this.fire('error', {error: new Error('An image with this name already exists.')});
         }
         this.imageManager.addImage(id, image);
@@ -461,7 +461,7 @@ class Style extends Evented {
     }
 
     removeImage(id: string) {
-        if (!this.imageManager.getImage(id)) {
+        if (!this.getImage(id)) {
             return this.fire('error', {error: new Error('No image with this name exists.')});
         }
         this.imageManager.removeImage(id);
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -456,6 +456,10 @@ class Style extends Evented {
         this.fire('data', {dataType: 'style'});
     }
 
+    getImage(id: string): ?StyleImage {
+        return this.imageManager.getImage(id);
+    }
+
     removeImage(id: string) {
         if (!this.imageManager.getImage(id)) {
             return this.fire('error', {error: new Error('No image with this name exists.')});
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1063,6 +1063,22 @@ class Map extends Camera {
     }
 
     /**
+     * Get an image from the style (such as one used by `icon-image` or `background-pattern`).
+     *
+     * @param id The ID of the image.
+     */
+    getImage(id: string): ?StyleImage {
+        if (!id) {
+            this.fire('error', {
+                error: new Error('Missing required image id')
+            });
+            return;
+        }
+
+        return this.style.getImage(id);
+    }
+
+    /**
      * Remove an image from the style (such as one used by `icon-image` or `background-pattern`).
      *
      * @param id The ID of the image.
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1072,7 +1072,7 @@ class Map extends Camera {
             this.fire('error', {
                 error: new Error('Missing required image id')
             });
-            return;
+            return false;
         }
 
         return !!this.style.getImage(id);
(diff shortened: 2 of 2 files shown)


src/ui/map.js
```diff
@@ -1062,6 +1062,15 @@ class Map extends Camera {
         this.style.addImage(id, { data: ((data: any): RGBAImage), pixelRatio, sdf });
     }
 
+    /**
+     * Get an image from the style (such as one used by `icon-image` or `background-pattern`).
+     *
+     * @param id The ID of the image.
+     */
+    getImage(id: string) {
+        this.style.getImage(id);
```
mollymerp: you need a return statement here

jfirebaugh:
What's the return type of `getImage`? Is that something we're comfortable making part of the public API?

src/ui/map.js
alex3165: my bad 👍

alex3165:
Thanks, I addressed all the comments and squashed my commits

mollymerp:
Good point @jfirebaugh – @alex3165 would returning a boolean and renaming the method `hasImage` suffice for your use-case? I'm not sure how useful it would be to return what is stored in the image manager: `StyleImage`

src/style/style.js
```diff
@@ -456,6 +456,10 @@ class Style extends Evented {
         this.fire('data', {dataType: 'style'});
     }
 
+    getImage(id: string): ?StyleImage {
+        return this.imageManager.getImage(id);
+    }
+
     removeImage(id: string) {
         if (!this.imageManager.getImage(id)) {
```
mollymerp: you can use the new `Style#getImage` method here.

src/ui/map.js
```diff
@@ -1062,6 +1062,22 @@ class Map extends Camera {
         this.style.addImage(id, { data: ((data: any): RGBAImage), pixelRatio, sdf });
     }
 
+    /**
+     * Get an image from the style (such as one used by `icon-image` or `background-pattern`).
+     *
+     * @param id The ID of the image.
+     */
+    getImage(id: string): ?StyleImage {
```
mollymerp: the build is failing because the type `StyleImage` was not imported so flow is erroring. I think we should rename the method `hasImage` and change the return type to `boolean`

alex3165:
Yes `hasImage` is good enough! 👍

mollymerp:
Thanks @alex3165 !

alex3165:
Hey, @mollymerp when do you plan to release this?

mollymerp:
This will go out with the v0.43.0 release – i'd estimate sometime in the next 3-5 weeks

areichman:
@mollymerp Is there an equivalent method that could be exposed for controls? If so, I can create a new ticket. 

The use case is switching map styles. Since there is no "change base layer" ability like in Leaflet, we store references to all active layers in our app state and then re-add those sources/layers in GL. It appears that controls are not actually removed during this process so when we re-add them we get duplicate copies of the control.

Jinkwon:
Great!!

mollymerp:
@areichman – no we don't expose a method for Controls. For DOM elements added to the map (see https://github.com/mapbox/mapbox-gl-js/issues/5821#issuecomment-349774516), we generally don't keep copies in the `Map` state to minimize memory use and API sprawl. Like `Marker`s, `Control`s are not part of the `Map`'s stylesheet, so changing styles won't remove the controls you've added unless you create a `new mapboxgl.Map(...` instance.  If you do need to remove the controls when you switch styles, I suggest you save a copy of the `Control` instance and then invoke `Map#removeControl` as needed.

areichman:
@mollymerp That's what I ended up doing. I am storing the control reference for other reasons and can just check for the presence of `control._map` to know if it has already been added. Thanks!

alpha-beta-soup:
I'm not sure what this internal `getImage` method returns, but just putting it out there that it would be great to use something like that to obtain the symbol used by a Mapbox style, so that a legend can be populated within an application that depends on Mapbox GL JS.

alpha-beta-soup:
That's a handy method actually. My use case is to create automated legends from a layer. I do this with circle, symbol and polygon layers at the moment using my own (brittle) method that works because I have tightly-defined, known inputs (and graceful failure). But point `symbol` layers only work with a reference to the name of the symbol in the sprite sheet—and then I use that name to select the correct SVG to include in the legend. That means I need to bundle the sprites/symbols into both my application and into the spritesheet.

I don't know if what I want to do is at all possible, but I just think it would be useful to get the actual image used in the spritesheet (or some kind of remote reference) to include in a legend, without having that separation between the symbols in Mapbox and those in the application. Then someone can add symbols to the spritesheet via Mapbox Studio without rebuilding the application.

I'll include a gif. The legend for the area layer is determined dynamically, entirely based on the content of the style loaded when the application starts. Ideally I'd achieve the same for symbol layers too. In this example, all layers are pre-included into the style, and the UI simply toggles layer `visibility`.

peek 2018-11-15 10-36

Issue #2769, opened 2016-06-20
Fix transitions from property default values


jfirebaugh:
Noticed while working on #2762.

`map.setPaintProperty("circle", "circle-color", "red")` when called on this layer will interpolate black to red correctly:

```
                {
                    "id": "circle",
                    "type": "circle",
                    "source": "geojson",
                    "paint": {
                        "circle-color": "black"
                    }
                }
```

However, with this layer it'll just jump from black to red without a transition:

```
                {
                    "id": "circle",
                    "type": "circle",
                    "source": "geojson",
                    "paint": {}
                }
```

The expected behavior is for both to have an interpolated transition.

stepankuzmin:
Hi! I've done a little research and found, that `StyleLayer#_paintDeclarations` is empty for layers with default paint properties, while layers with specified paint properties have `_paintDeclarations`.

I've tried to set paint declaration to default after https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style_layer.js#L111 like this:

```js
this._paintDeclarations[klass || ''][name] = new StyleDeclaration(this._paintSpecifications[name]);
```

but it does not affect transition. So am I missing something? Should we explicitly define default paint properties in StyleLayer constructor? (something like this)

```js
for (paintName in this._paintSpecifications) {
    this.setPaintProperty(paintName, this._paintSpecifications[paintName].default);
}
```

mourner:
@stepankuzmin I would suggest exploring `style_transition.js` for why it doesn't transition from a default property. There's an early bailout there somewhere probably.

stepankuzmin:
It seems that there is no transition from defaults because there is no `oldTransition` to transition from. `oldTransition` for `StyleLayer` is set in `_applyPaintDeclaration`, which is called from `updatePaintTransitions` if `StyleLayer._paintDeclarations` is not empty. For layer with default paint properties, there is no `_paintDeclarations`, so there is no `oldTransition` also.

Defining default paint properties in `StyleLayer` constructor solves the problem, but I am not sure about this.

Issue #2467, opened 2016-04-19
Raster Flicker


dflynn15:
This is more or less just a question. I have been trying to track down a raster "flickering" that happens whenever the Map is animated using `pitch` and noticed that there was a [previous issue](https://github.com/mapbox/mapbox-gl-js/issues/2212) and [PR](https://github.com/mapbox/mapbox-gl-js/pull/2282) that in theory should have fixed this. Is there a possibility this has been re-introduced by the upgrades to `0.17.0`

If there is anything else I need to provide let me know!

**mapbox-gl-js version**: `0.17.0`
### Steps to Trigger Behavior
1. Go to this CodePen [here](http://codepen.io/dflynn15/full/aNKBKp/)
2. Click the "Fly In" button
3. Click the "Fly out" button after animation
4. You should notice the raster flickering on the top of the map

ansis:
I'm not seeing any raster layers in the codepen so I think this is different than https://github.com/mapbox/mapbox-gl-js/issues/2212

Which layers are flickering? lines? fills? text?

dflynn15:
Ah, I was wrong in the terminology. My bad :sob:

It appears to be the map tiles (?). I do want to emphasize this is intermittent and best "found" on Chrome. Firefox seems to be quick clean :grimacing: You can de-prioritize this until I can get a consistent or reproducible screen capture for you :disappointed:

lucaswoj:
Thanks for reporting @dflynn15!

In the name of keeping our issue queue lean and mean, I'm going to close this until we can get a consistent or reproducible screen capture. Please re-open once you have one!

mtirwin:
I'm able to reproduce a raster flicker using production mapbox.mapbox-satellite, mapbox-gl 0.18.0, Chrome 50.0.2661.86, on current OSX. 

Flicker behavior happens consistently on [fly-to events with with pitch 0 and > 0 in this demo](https://www.mapbox.com/labs/digitalglobe-rio/#), and [this screen vid catches the issue](https://www.dropbox.com/s/12mna61vaghx4mm/flicker.mp4?dl=0) (can't get a gif with high enough frame rate) in the upper left after the initial fly-to event.

mtirwin:
Much better capture with this video:

https://www.dropbox.com/s/efywvcw1594zs1r/flicker2.mov?dl=0

kkaefer:
I'm seeing this as well, and it's pretty jarring.

mapsam:
[...]

PR #5780, opened 2017-11-30
Transition regression tests

### LINKED ISSUE #3912: Add integration tests for transitions
Migrated from https://github.com/mapbox/mapbox-gl-test-suite/issues/168

> Per https://github.com/mapbox/mapbox-gl-style-spec/issues/171, let's add tests for transition properties. Sticking point is somehow taking a screenshot after a certain elapsed time: @lucaswoj pointed me to a [wait operation](https://github.com/mapbox/mapbox-gl-test-suite/blob/a0b0e3be927dff9a01b0869a4d0b41c830370388/render-tests/runtime-styling/filter-default-to-false/style.json#L7-L19) that waits til rendering is complete, but it [doesn't support a duration argument yet](https://github.com/mapbox/mapbox-gl-js/blob/95b0ca42bf8376561db7e96c616d76696d8f65f7/test/suite_implementation.js#L83-L102).

> - [ ] extend `wait` to include a duration parameter in `gl-js`
> - [ ] add tests with durations to test transitions

Migrated from https://github.com/mapbox/mapbox-gl-test-suite/issues/116

> We should to figure out how to add tests for property transitions, in order to prevent bugs like:
> - https://github.com/mapbox/mapbox-gl-js/issues/2762
> - https://github.com/mapbox/mapbox-gl-js/issues/2769
>
> A prerequisite for this is that the test harness has control over the clock used for rendering transitions, so it can deterministically render, say, 50% of the way through a transition. Then we could add that percentage as a parameter somewhere in the test case metadata.
> 
> gl-js internals which use Date.now() directly that need to be converted to using a passed in parameter:
> 
> https://github.com/mapbox/mapbox-gl-js/blob/062f52d836e8991f694c6742ab24ecfdc643151b/js/style/style_declaration.js#L53-L53
> https://github.com/mapbox/mapbox-gl-js/blob/8eaeebbcaeea012fe7876c06fb558a0a2ea99475/js/style/style_transition.js#L54-L54
> https://github.com/mapbox/mapbox-gl-js/blob/8eaeebbcaeea012fe7876c06fb558a0a2ea99475/js/style/style_layer.js#L311-L311
> https://github.com/mapbox/mapbox-gl-js/blob/1df146627092b58c8db64d45530c74cd12a3fa02/js/render/frame_history.js#L17-L17
> https://github.com/mapbox/mapbox-gl-js/blob/7840d9c2265c1d9be52a329d9a9c38784b75b3ca/js/style/style.js#L250-L257

_Migrator's note: sinon has some built-in utilities for mocking out the system clock http://sinonjs.org/docs/#clock_

jfirebaugh:
Adds regression render tests for #2762, #2769, and #2467. Fixes #3912.


jfirebaugh:
Added descriptions.

PR #5782, opened 2017-11-30
switch TileCoord to -native's TileID types

ansis:
TODO
- [x] fix query tests
- [x] rebase on master

-----

This switches from a single `TileCoord` type to the three types used in -native: `CanonicalTileID`, `UnwrappedTileID` and `OverscaledTileID`. This is very roughly a port of https://github.com/mapbox/mapbox-gl-native/pull/3834.

There are some parts of `SourceCache` and `Tile` that don't match -native so some of the uses of each ID type don't match 100% in -native and -js. Those differences should be eliminated in separate prs.

## Launch Checklist

 - [x] post benchmark scores https://bl.ocks.org/anonymous/raw/9b20a0d3bb856c4c074542b7683987c2/


src/source/worker_source.js
```diff
@@ -11,7 +11,7 @@ export type TileParameters = {
 };
 
 export type WorkerTileParameters = TileParameters & {
-    coord: { z: number, x: number, y: number, w: number },
+    tileID: { overscaledZ: number, wrap: number, canonical: { z: number, x: number, y: number } },
```
jfirebaugh: I see `OverscaledTileID` is [registered](https://github.com/mapbox/mapbox-gl-js/blob/c0ded76f6c4fd4f87fe4e38ee36d6bf1ea6f6520/src/util/web_worker_transfer.js#L52), so can this just be `tileID: OverscaledTileID`?

bench/benchmarks/layout.js
```diff
@@ -11,7 +11,8 @@ const promisify = require('pify');
 const WorkerTile = require('../../src/source/worker_tile');
 const StyleLayerIndex = require('../../src/style/style_layer_index');
 const deref = require('../../src/style-spec/deref');
-const TileCoord = require('../../src/source/tile_coord');
+const OverscaledTileID = require('../../src/source/tile_id').OverscaledTileID;
```
jfirebaugh: I think we're migrating to `const {OverscaledTileID} = require(...);` for this type of require.

ansis:
I've fixed the require style, the type simplification and rebase on the dem rendering.

PR #5790, opened 2017-12-01
Refactor StructArray to use statically defined classes

anandthakker:
- [x] Decouple generated Struct and StructArray classes to reduce redundancy
- [x] Lots of code cleanup
- [x] Assess & reduce bundle size impact
- [x] Fix up tests
- [x] Benchmarks!
- [x] Add internal docs about StructArray

---

### Bundle size

I was initially concerned that the move to static, generated struct array types would cause a large increase in bundle size, but, considering gzipped size, it doesn't look bad:

```
675K Dec  4 10:53 mapbox-gl.master.js
731K Dec  4 11:11 mapbox-gl.struct-array-refactor.js

167K Dec  4 10:53 mapbox-gl.master.js.gz
171K Dec  4 11:11 mapbox-gl.struct-array-refactor.js.gz
```

TBD whether the 8% increase in _non_-gzipped size shows up as a significant increase in script parsing/compiling time (and, relatedly, web worker init)


src/data/array_type/circle_blur_paint_vertex.js
```diff
@@ -0,0 +1,51 @@
...
+class CircleBlurPaintVertexStructArray extends StructArrayLayout {
+    uint8: Uint8Array;
+    float32: Float32Array;
+
+    geta_circle_blur(index: number) {
+        return this.float32[index * 1 + 0];
+    }
+
+    static deserialize(input: SerializedStructArray): CircleBlurPaintVertexStructArray {
+        const structArray = Object.create(CircleBlurPaintVertexStructArray.prototype);
```
anandthakker: This shouldn't really have to be duplicated each individual subclass, but it's currently necessary in order to create the deserialized object with the right prototype.  Maybe worth adding an `afterDeserialize` hook, which could be defined on the base class, so that the standard web worker transfer can take care of the `Object.create()` part.

anandthakker:
Initial benchmarks: https://bl.ocks.org/anonymous/raw/7c6b3f6ff5f562a678be3b2e6adc2050/

anandthakker:
@jfirebaugh @ansis There's still some cleaning up to do, but this could use a first review.

src/data/index_array_type.js
```diff
@@ -1,25 +1,12 @@
...
-            name: 'vertices',
-            components
-        }]
-    });
-}
-
 module.exports = {
-    LineIndexArray: createIndexArrayType(2),
-    TriangleIndexArray: createIndexArrayType(3)
+    LineIndexArray: require('./array_type/line_index'),
```
jfirebaugh: I'm tempted to say these could just be straight `Uint16Array`s, or light wrappers around them.

build/generate-struct-arrays.js
```diff
@@ -0,0 +1,325 @@
...
+        name: string,
+        type: ViewType,
+        +components?: number,
+    }>,
+    alignment?: number
+};
+
+
+const structArrayLayoutJs = ejs.compile(fs.readFileSync('src/util/struct_array_layout.js.ejs', 'utf8'), {strict: true});
+const structArrayJs = ejs.compile(fs.readFileSync('src/util/struct_array.js.ejs', 'utf8'), {strict: true});
```
ansis: What is the advantage of separating the layout and the rest of the struct array? Is there a lot of duplication that is avoided? What's the downside of the duplication?

anandthakker: There's a fair amount of duplication.  One definite effect of the duplication is that it increases the code size; gzip probably makes this minimal for downloading, but the code still has to be parsed `1 + workerCount` times in the browser.  I also suspect that reducing the redundancy might make it a little easier for the optimizer to do its thing, but that's just a hypothesis.

anandthakker: Maybe more importantly, though, is that this separation is a step towards eliminating "the rest of the struct array" part entirely for the array types where we don't actually make use of the struct accessors -- i.e., all the layout arrays and paint property arrays.

ansis: Thanks, that makes sense!

anandthakker:
> Do you have any ideas what's causing this or how we can improve it?

Since the `Paint` benchmark doesn't rely on any data-driven style properties, there's good reason to believe that the cost is on the JS side (rather than the WebGL side).  My first guess is that the overhead of creating the (possibly empty or one-item) array of vertexBuffers to pass to `VertexArrayObject#bind()` each frame, rather than just passing the one buffer for layout and (maybe) a second for paint, could be a key factor.  Caching the container may help here.

anandthakker:
Updated bundle sizes:
```
682K mapbox-gl.master.js
729K mapbox-gl.struct-array-refactor.js

168K mapbox-gl.master.js.gz
172K mapbox-gl.struct-array-refactor.js.gz
```

> TBD whether the 8% increase in _non_-gzipped size shows up as a significant increase in script parsing/compiling time (and, relatedly, web worker init).

The ungzipped delta is now down to  +6.8%.  We no longer have a benchmark for map load (it's difficult to measure reliably), but I'm not _too_ concerned about the parse time at this point.

anandthakker:
Updated benchmarks after 05069ab and 8813f7b, looking better: https://bl.ocks.org/anonymous/raw/057f7a01280ae9876b42535fdb635b5b/.

- The increase in `Paint` is now down to ~0.7ms (less than 0.5 standard deviations), and `Layout`, `LayoutDDS` are clearly faster.
- `LayerRaster` +1.2 standard deviations looks like an anomaly, didn't reproduce despite multiple retries: https://bl.ocks.org/anonymous/raw/c249897a4816c12b3310a2e9adf7efdf/
- The 2x speedup in `QueryBox` is exciting enough that I'd like to understand better why it's so much faster.  The query tests are passing and the query_features debug page works, but I'm still curious.  The statically defined classes for `CollisionBoxArray` and `FeatureIndexArray` likely mean that the `array.get(index)` operation is now faster, and it does happen per feature in `filterMatching()`.

anandthakker:
@jfirebaugh This is ready for another review.

> I'm tempted to say these could just be straight Uint16Arrays, or light wrappers around them.

Didn't get to this bit -- will ticket out as follow-up work if I'm not able to sneak it in tomorrow AM.

anandthakker:
Hmm, looking into simplifying `{Triangle,Line}IndexArray`s, I'm not sure it's worth it: since typed arrays aren't resizable, we'd still need the "light" wrapper to implement `resize()`, at which point it's not really much simpler than `StructArrayLayout_6_3ui` / `StructArrayLayout_4_2ui`

src/util/struct_array.js
```diff
@@ -3,13 +3,9 @@
...
-const {register} = require('./web_worker_transfer');
-const Point = require('@mapbox/point-geometry');
 
 import type {Transferable} from '../types/transferable';
 
-module.exports = createStructArrayType;
-
 const viewTypes = {
     'Int8': Int8Array,
     'Uint8': Uint8Array,
```
jfirebaugh: The `StructMember` typedef a few lines down from here is obsolete.

src/util/struct_array.js
```diff
@@ -139,35 +130,18 @@ class StructArray {
...
-            cacheKey: array.constructor.prototype._cacheKey
         };
     }
 
-    static deserialize(input: SerializedStructArray): StructArray {
-        const Klass = input.cacheKey && input.cacheKey in structArrayTypeCache ?
-            structArrayTypeCache[input.cacheKey] :
-            createStructArrayType(input.type);
-        const structArray = Object.create(Klass.prototype);
+    static deserialize(input: SerializedStructArray) {
```
jfirebaugh: What work remains to eliminate the need for custom serialization for `StructArray`?

src/data/program_configuration.js
```diff
@@ -2,38 +2,31 @@
...
-    name: string,
-    type: ViewType,
-    components?: number
-}
-
-type PaintAttribute = {
-    property: string,
-    name?: string,
-    useIntegerZoom?: boolean
+export type PaintPropertyStatistics = {
```
jfirebaugh: Unused.

src/data/program_configuration.js
```diff
@@ -2,38 +2,31 @@
...
+    [property: string]: { max: number }
 }
 
 export type ProgramInterface = {
```
jfirebaugh: Used only by `symbol_bucket.js`, should move there.

src/data/program_configuration.js
```diff
@@ -123,40 +117,60 @@ class SourceExpressionBinder<T> implements Binder<T> {
     type: string;
     statistics: { max: number };
 
-    constructor(expression: SourceExpression, name: string, type: string) {
+    PaintVertexArray: Class<StructArray>;
```
jfirebaugh: Not used outside the constructor; make it a local variable and remove the `_omit`.

src/data/program_configuration.js
```diff
@@ -247,53 +278,41 @@ class CompositeExpressionBinder<T> implements Binder<T> {
...
+    static createDynamic<Layer: TypedStyleLayer>(layer: Layer, zoom: number, filterProperties: (string) => boolean) {
         const self = new ProgramConfiguration();
-        const attributes = [];
-
-        for (const attribute of programInterface.paintAttributes || []) {
-            const property = attribute.property;
-            const name = attribute.name || property.replace(`${layer.type}-`, '').replace(/-/g, '_');
-            const value: PossiblyEvaluatedPropertyValue<any> = layer.paint.get(property);
+        for (const property in layer.paint._values) {
+            if (!filterProperties(property)) continue;
```
jfirebaugh: The combination of `filterProperties` plus needing `instanceof` and `property-function` conditionals seems a bit cumbersome. Could we drive this loop via a passed in array of property names or something instead?

src/util/struct_array.js
anandthakker: Really the main issue is the need for `_refreshViews()` after deserializing.  If we added some kind of `afterDeserialize` (or maybe `initialize`) hook, I think that would be sufficient.

src/util/struct_array.js
anandthakker: 👍 removed

anandthakker:
@jfirebaugh great comments -- most of them are addressed.  Over to you for these remaining ones:

 - [ ]  The capitalization is off for these generated accessors. Are they actually used anywhere?
 - [ ] FillExtrusionLayoutArray uses StructArrayLayout_16_6i, while Collision{Box,Circle}LayoutArray use StructArrayLayout_12_6i, I guess because of the layout of the extrusion attributes. We should adjust the packing so that extrusion attributes are on a 12 byte stride too (can ticket for later).
 - [ ] The combination of filterProperties plus needing instanceof and property-function conditionals seems a bit cumbersome. Could we drive this loop via a passed in array of property names or something instead?

jfirebaugh:
First one is done, latter two ticketed as https://github.com/mapbox/mapbox-gl-js/issues/5863 and https://github.com/mapbox/mapbox-gl-js/issues/5864.

PR #5791, opened 2017-12-01
Fix cross-fading of *-pattern and line-dasharray

### LINKED ISSUE #5740: fill patterns don't finish crossfading
from https://github.com/mapbox/mapbox-gl-js/issues/5716#issuecomment-346177237

@pathmapper:

---

There is also an effect on a fill-pattern:

image

Here are two examples, just zoom in to see the issue:

[v.0.41.0 example](https://jsfiddle.net/z73Lywrv/1/) -> the blurry lines are only a short moment visible when zooming

[v.0.42.2 example](https://jsfiddle.net/z73Lywrv/2/) -> the blurry lines are staying visible at certain zoom levels

---

**mapbox-gl-js version**: 0.42.2

### Steps to Trigger Behavior

 1. Zoom on a map with a fill pattern
 2. Stop zooming

### Expected Behavior
Pattern finishes crossfading.

### Actual Behavior
Pattern gets stuck in the middle of a fade.

jfirebaugh:
Fixes #5740. Quite a bit of refactoring was needed to make this feasible:

* Needed to avoid serializing `PossiblyEvaluatedPropertyValue` (1st commit)
* Needed to rework the interaction and parameter passing between `Map#_render`, `Style#update`, and `Style#_recalculate` (2nd through 5th commits)
* Needed to remove `_placementDirty` from `Map#loaded()`, otherwise the test case hangs (6th commit)

https://bl.ocks.org/anonymous/raw/17fa5e060a121799adc5f80f6d722e06/


ChrisLoer:
I only looked at the `_placementDirty` change, which I think is fine using the interpretation of `loaded()` to mean "all data has loaded", because in the current code any data loading will have triggered an accompanying full placement.

@ansis this is something to watch out for with the placement back-port. In the context of tests, I think we almost always want to just do placement before each frame (unless we're specifically testing a transition where placement timing matters).

Post-back-port, I'm not sure what the right behavior for `loaded()` should be in the case where we've loaded a new tile but haven't yet run placement on it. If you consider the placement just part of a transition animation, then maybe it's OK to return "true" just like we would if there were some other style transition in progress (although even there, we'd only return "true" if you called `loaded` from within a handler for the `load` signal, because we'd set `_styleDirty` right after firing the signal). But placement doesn't just affect rendering: it also affects the results of `queryRenderedFeatures`.

PR #5793, opened 2017-12-02
Framebuffer refactor

lbud:
Primary changes:
* Refactors all render calls that render to offscreen framebuffers (`fill-extrusion`, `heatmap`, `hillshade`) into one `offscreen` pass (fixes https://github.com/mapbox/mapbox-gl-js/issues/5803)
* Makes `Context` responsible for all framebuffer + renderbuffer creation
* Introduces `Framebuffer` class, responsible for managing its color and/or depth attachment states, retaining size info, and cleaning up its attachments with `gl.delete*` calls when destroyed
  * Framebuffers still need to be bound and unbound (by setting context.bindFramebuffer) in render code in order to specify the intended rendering target, but do not need to be manually bound in order to change their attachments: attachment values handle the necessary bindings when set
* Eliminates `RenderTexture` class

Minor changes:
* Moves GL extensions stored on the painter to the context object
* ~Caches tile-sized framebuffers on the painter when unloaded (from `hillshade` layers) for reuse in new loaded hillshade tiles~
* Updates `heatmap-radius/antimeridian` render test (https://github.com/mapbox/mapbox-gl-js/pull/5793#issuecomment-349437843)

Benchmark results: https://bl.ocks.org/anonymous/raw/3d9a8aa509b5dc8d0c1d97c6e5f6110f/

Refs https://github.com/mapbox/mapbox-gl-js/issues/145.


lbud:
Note that the test updated in https://github.com/mapbox/mapbox-gl-js/pull/5793/commits/3577b7cbe1d1b72f523ac9720d4c0f2798c9055b changed because it had been incorrectly overdrawing before: `drawHeatmap` was previously using `painter.isOpaquePass`, which we deprecated a little while ago in favor of a `painter.pass` enum.

mourner:
@lbud the heatmap drawing changes look great to me!

lbud:
> Can the the `Framebuffer` itself own the attachments and  be responsible for releasing them in a `Framebuffer#destroy` method comparable to `Texture#destroy` and `VertexArrayObject#destroy`? Or does the need for shared attachments prevent that?

Yes, good idea — addressed this in https://github.com/mapbox/mapbox-gl-js/pull/5793/commits/470e764fe02ffeb2fcadf3777046647bd23f41ba. Per the [WebGL spec](https://www.khronos.org/registry/webgl/specs/1.0/#5.14.7), "If the renderbuffer has already been deleted the call has no effect," so shared renderbuffer attachments won't cause a problem (confirmed on a debug map).

src/style/style_layer.js
```diff
@@ -42,7 +42,7 @@ class StyleLayer extends Evented {
     _transitioningPaint: Transitioning<any>;
     +paint: mixed;
 
-    viewportFrame: ?RenderTexture;
+    viewportFrame: ?Framebuffer;
```
jfirebaugh: Can this declaration be pushed down hierarchy, into `FillExtrusionStyleLayer`?

src/source/raster_dem_tile_source.js
```diff
@@ -116,6 +116,10 @@ class RasterDEMTileSource extends RasterTileSource implements Source {
 
     unloadTile(tile: Tile) {
         if (tile.demTexture) this.map.painter.saveTileTexture(tile.demTexture);
+        if (tile.fbo) {
+            this.map.painter.saveTileFramebuffer(tile.fbo);
```
jfirebaugh: How much of a performance difference does caching framebuffers make?

mourner:
@lbud are you planning to do a similar refactor (consolidating offscreen rendering into an "offscreen" pass) on the gl-native side?

lbud:
@mourner we should use the same process (an offscreen pass to render to all offscreen framebuffers) on native. I haven't yet started something like this there, although at the moment fill-extrusion layers are the only offscreen layer type on master, so if I were to do so right now it would just be a minor refactor (moving [extrusion depth buffer setup](https://github.com/mapbox/mapbox-gl-native/blob/a170c18bb228a12cf0cb2d041d3c91fc99d19600/src/mbgl/renderer/renderer_impl.cpp#L438-L443) into render_fill_extrusion.cpp and renaming "3d" to "offscreen"). I know you and @mollymerp have heatmap/hillshade branches in progress — not sure how far along you are; if it's convenient for you to render those in the 3d pass feel free to make those changes, but if not it shouldn't be a big deal to refactor afterward as we did in gl-js.

PR #5799, opened 2017-12-04
Upgrade geojson-vt to 3.0.0

mourner:
Makes GeoJSON indices take 2x less memory and generate tiles 20%–100% faster.

The external interface in the [new geojson-vt release](https://github.com/mapbox/geojson-vt/releases/tag/v3.0.0) is the same, so this upgrade shouldn't be breaking, but let's watch out for any edge cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "csscolorparser": "~1.0.2",
     "earcut": "^2.0.3",
     "geojson-rewind": "^0.3.0",
-    "geojson-vt": "^2.4.0",
+    "geojson-vt": "^3.0.0",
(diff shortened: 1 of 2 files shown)


mourner:
I have a feeling that benchmarks are broken on my machine due to some obscure GPU/OS issue, but here are the results of the only benchmark that uses GeoJSON (two runs):

image

image

mourner:
Update: just tried upgrading to macOS 10.12.2 Beta 6 and the issue seems gone! I have snappy GL maps again. Woohoo! Updated bench (notice 2x smaller times):

image

PR #5801, opened 2017-12-04
Update to @mapbox/mapbox-gl-supported v1.3.0

### LINKED ISSUE #4358: Security Error on WebWorker IE11 Win8.1
Hello , we have found some security issues on windows explorer 11.
the specific error is on the blob url initialization for web worker. The bug is [documented here. ](https://connect.microsoft.com/IE/feedback/details/801810/web-workers-from-blob-urls-in-ie-10-and-11), it seems to be a bug on previous versions of IE11. The problem is that the `mapboxgl.supported()` still returns true. So, I see two options ,   `mapboxgl.supported()` should return false or patch WebWorker initialization with an strategy similar to [https://github.com/parallel-js/parallel.js/pull/16/files](https://github.com/parallel-js/parallel.js/pull/16/files)

I think this is actually related with #2504

jfirebaugh:
Fixes #4358.

## Launch Checklist

 - [x] manually test the debug page (also tested [the example](https://www.mapbox.com/mapbox-gl-js/example/check-for-support/))

### DIFF
--- a/docs/components/example.js
+++ b/docs/components/example.js
@@ -6,7 +6,7 @@ import PageShell from './page_shell';
 import LeftNav from './left_nav';
 import TopNav from './top_nav';
 import {highlightMarkup} from './prism_highlight';
-import supported from 'mapbox-gl-supported';
+import supported from '@mapbox/mapbox-gl-supported';
 import {copy} from 'execcommand-copy';
 import examples from '@mapbox/batfish/data/examples';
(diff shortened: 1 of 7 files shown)


jlaine:
It looks like the typings in `flow-typed/mapbox-gl-supported.js` also need updating

jfirebaugh:
@jlaine Good catch! Fixed.

PR #5804, opened 2017-12-04
Use custom class property for serialization registry

### LINKED ISSUE #5800: Map fails to load with minified build
With the minified build -- `dist/mapbox-gl.js`, built by `yarn run build-min` -- all maps fail to load. On the console is the following error:

```
Uncaught Error: can't deserialize object of anonymous class
    at deserialize (web_worker_transfer.js:212)
    at deserialize (web_worker_transfer.js:228)
    at Tile.loadVectorData (tile.js:157)
    at t.i (vector_tile_source.js:125)
    at Actor.receive (actor.js:81)
````

Bisects to #5702. cc @anandthakker

### PR DESCRIPTION
**Pull request by anandthakker**
Fixes #5800

It's possible for named classes to become anonymous after minification,
so relying on `constructor.name` isn't safe.  Instead, we attach a
custom class property when the class is registered and use that rather
than `constructor.name` to recover the correct class for
deserialization.

### DIFF
--- a/src/data/bucket/circle_bucket.js
+++ b/src/data/bucket/circle_bucket.js
@@ -56,7 +56,6 @@ const LayoutVertexArrayType = createVertexArrayType(circleInterface.layoutAttrib
  * @private
  */
 class CircleBucket<Layer: CircleStyleLayer | HeatmapStyleLayer> implements Bucket {
-    static classRegistryName;
     static programInterface: ProgramInterface;
     index: number;
     zoom: number;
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: src/data/bucket/circle_bucket.js [hunk 1]
```diff
@@ -56,6 +56,7 @@ const LayoutVertexArrayType = createVertexArrayType(circleInterface.layoutAttrib
  * @private
  */
 class CircleBucket<Layer: CircleStyleLayer | HeatmapStyleLayer> implements Bucket {
+    static classRegistryName;
```
jfirebaugh: Unused?

src/source/source_cache.js
```diff
@@ -301,10 +301,7 @@ class SourceCache extends Evented {
         }
     }
     /**
-     * Get a specific tile by TileCoordinate
-=======
-     * Get a specific tile by TileID 
->>>>>>> switch TileCoord to -native's TileID types
+     * Get a specific tile by TileID
```
anandthakker: Leftover from #5782

PR #5805, opened 2017-12-04
docs for hillshade layer and raster-dem source

andrewharvey:
@mollymerp This isn't ready for merging, but posting it here for feedback. This PR includes:

1. added hillshade example
I'm not sure if we should have a dedicated example for the hillshade layer, but I put one together.

2. added `raster-dem` source and `hillshade` layer to docs.

3. included `raster-dem` source in the style spec rather than it being a type of the `raster` source.

 - [ ] post benchmark scores

### DIFF
--- a/docs/pages/example/hillshade.html
+++ b/docs/pages/example/hillshade.html
@@ -1,0 +1,31 @@
+<div id='map'></div>
+<script>
+var map = new mapboxgl.Map({
+    container: 'map',
+    style: 'mapbox://styles/mapbox/outdoors-v9',
+    center: [-118.8291, 37.4266],
+    zoom: 9
+});
+
+map.on('load', function () {
+    map.removeLayer('hillshade_shadow_extreme');
+    map.removeLayer('hillshade_shadow_dark');
+    map.removeLayer('hillshade_shadow_med');
+    map.removeLayer('hillshade_shadow_faint');
+    map.removeLayer('hillshade_highlight_med');
+    map.removeLayer('hillshade_highlight_bright');
+
+    map.addSource('dem', {
+        "type": "raster-dem",
+        "url": "mapbox://mapbox.terrain-rgb"
+    });
+    map.addLayer({
+        "id": "hillshading",
+        "source": "dem",
+        "type": "hillshade"
+    }, 'waterway-river-canal-shadow');
+    // insert below waterway-river-canal-shadow as that is where Hillshading
+    // sits in the Mapbox Outdoors style
+});
+
+</script>
--- a/docs/pages/example/hillshade.html
+++ b/docs/pages/example/hillshade.html
@@ -2,19 +2,12 @@
 <script>
 var map = new mapboxgl.Map({
     container: 'map',
-    style: 'mapbox://styles/mapbox/outdoors-v9',
+    style: 'mapbox://styles/mapbox/cjaudgl840gn32rnrepcb9b9g', // the outdoors-v10 style but without Hillshade layers
     center: [-118.8291, 37.4266],
     zoom: 9
 });
 
 map.on('load', function () {
-    map.removeLayer('hillshade_shadow_extreme');
-    map.removeLayer('hillshade_shadow_dark');
-    map.removeLayer('hillshade_shadow_med');
-    map.removeLayer('hillshade_shadow_faint');
-    map.removeLayer('hillshade_highlight_med');
-    map.removeLayer('hillshade_highlight_bright');
-
     map.addSource('dem', {
         "type": "raster-dem",
         "url": "mapbox://mapbox.terrain-rgb"
--- a/docs/pages/example/hillshade.html
+++ b/docs/pages/example/hillshade.html
@@ -3,7 +3,7 @@
 var map = new mapboxgl.Map({
     container: 'map',
     style: 'mapbox://styles/mapbox/cjaudgl840gn32rnrepcb9b9g', // the outdoors-v10 style but without Hillshade layers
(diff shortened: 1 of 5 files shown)


src/style-spec/validate/validate_source.js
```diff
@@ -40,7 +40,7 @@ module.exports = function validateSource(options) {
         errors = errors.concat(validateObject({
```
mollymerp: I think we can actually get rid of this case now and add `raster-dem` into the previous case with `vector` and `raster` sources. you will have to edit line 26 to something like 
```js
valueSpec: styleSpec[`source_${type.replace('-','_'}`],
```
to get it to work.

docs/pages/example/hillshade.html
```diff
@@ -0,0 +1,32 @@
+<div id='map'></div>
+<script>
+var map = new mapboxgl.Map({
+    container: 'map',
+    style: 'mapbox://styles/mapbox/outdoors-v9',
+    center: [-74.50, 40],
```
mollymerp: let's change the map center to somewhere with ⛰ s. Z11, center: -119.5857, 37.7221 is yosemite valley for example

src/style-spec/validate/validate_source.js
andrewharvey: Good point, that reduces code duplication.

docs/pages/example/hillshade.html
andrewharvey: Agreed!

docs/pages/example/hillshade.html
```diff
@@ -0,0 +1,32 @@
...
+map.on('load', function () {
+    map.removeLayer('hillshade_shadow_extreme');
+    map.removeLayer('hillshade_shadow_dark');
+    map.removeLayer('hillshade_shadow_med');
+    map.removeLayer('hillshade_shadow_faint');
+    map.removeLayer('hillshade_highlight_med');
+    map.removeLayer('hillshade_highlight_bright');
+
+    map.setLight({
+        position: [1.15, 335, 60]
```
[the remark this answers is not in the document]
andrewharvey: I did this since the debug map used a light position, but I see what you mean now.

docs/pages/example/hillshade.html
```diff
@@ -0,0 +1,31 @@
...
+    zoom: 11
+});
+
+map.on('load', function () {
+    map.removeLayer('hillshade_shadow_extreme');
+    map.removeLayer('hillshade_shadow_dark');
+    map.removeLayer('hillshade_shadow_med');
+    map.removeLayer('hillshade_shadow_faint');
+    map.removeLayer('hillshade_highlight_med');
+    map.removeLayer('hillshade_highlight_bright');
```
andrewharvey: @mollymerp This was one thing I'm not sure about. In terms of an example for the hillshade layer, it might make better sense to just use something like the basic style and add hillshading to that. It would be a simpler example.

This example here is more complex due to it removing those existing hillshade layers from outdoors client side. Especially given any real map wanting Outdoors with raster hill shading could first cut the Hillshade layers from Outdoors in Studio as a new style rather than doing it client side... :thought_balloon:

mollymerp: yeah good point – I made a new public style in the Mapbox account that is Outdoors minus the hillshade layers,so you can use this style id and explain with a comment what it is ?

mapbox://styles/mapbox/cjaudgl840gn32rnrepcb9b9g

andrewharvey: Great. Assuming you're okay that people will copy the example and rely on that style not disappearing in the future?

mollymerp: yep – and we can replace it with a permanent style whenever our core styles start using this layer type.

mollymerp:
Thanks for the fixes @andrewharvey – I added a couple tweaks to the example, and added in the explicit properties the raster-dem source type currently supports for consistency with the other source types, as well as to remove the `scheme` property, which isn't currently supported because this source only works with the mapbox terrain-rgb tileset at the moment.

src/style-spec/reference/v8.json
```diff
@@ -227,6 +225,58 @@
...
+    "url": {
+      "type": "string",
+      "doc": "A URL to a TileJSON resource. Supported protocols are `http:`, `https:`, and `mapbox://<mapid>`."
+    },
+    "tiles": {
+      "type": "array",
+      "value": "string",
+      "doc": "An array of one or more tile source URLs, as in the TileJSON spec."
+    },
+    "bounds": {
```
andrewharvey: That's interesting, I tested bounds and found it had no affect. I better re-test it.

mollymerp: in order for it to work you have to use the `tiles: [...]` property to specify the url because otherwise the `bounds` returned by the source's `TileJSON` will override the ones you specify. which I suppose we should also document 😅

mollymerp: this worked as expected for me:
```es6
        "rasterTerrain": {
            "type": "raster-dem",
            "tiles": [`https://a.tiles.mapbox.com/v4/mapbox.terrain-rgb/{z}/{x}/{y}.png?access_token=${mapboxgl.accessToken}`],
            "bounds" : [-40, -40, 40, 40]
        }
```

andrewharvey: ahh that makes sense now :grin:

andrewharvey: Agreed, it should be documented but I don't think it's critical for this PR. The same applies to the rest of these properties and raster sources too.

PR #5812, opened 2017-12-05
 Convert functions to expressions; add interpolation color spaces

### LINKED ISSUE #5326: Implement LAB/HSL color interpolation for expressions
The expressions PR is already merged into master, but it [removed support](https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/function/convert.js#L30-L32) for LAB/HSL `colorSpace` property.

We should either add it back before a release (since it's a breaking removal), or remove its support fully (along with `color_spaces.js` and related code) if we don't plan on also implementing it on the Native side. 

cc @anandthakker @kkaefer

### LINKED ISSUE #5811: Should hcl interpolation use shortest path?
Our current HCL interpolation implementation does a naive numeric interpolation of the hue value, which does not take the shortest path between hues:

image

I propose to change this so that the default is shortest path interpolation:

image

jfirebaugh:
Convert functions to expressions for evaluation (reverses #5379). Adds support for interpolate-hcl and interpolate-lab expression operators to replace the colorSpace function property (fixes #5326). Switches to shortest-path interpolation for HCL (fixes #5811).

Evaluation of converted expressions is slightly faster:
https://bl.ocks.org/anonymous/raw/ea7442aff28940abb9a358501aa9067b/

However, conversion+creation is slower. I may try to reimplement conversion to directly construct expression instances, rather than converting to a JSON representation.
https://bl.ocks.org/anonymous/raw/63b5d4b704f55e8e50dd5871297a1490/

What's still not working currently is the `default` property for functions. @anandthakker, I noticed the function tests for this were [skipped](https://github.com/mapbox/mapbox-gl-js/blob/b860950b2862bd425913ccb67e420826280f55dc/test/unit/style-spec/function.test.js#L292). It looks like the skips were added as part of #4777. Do you recall if we ever came up with a plan for how to convert `default`s into expression syntax?


anandthakker:
> Do you recall if we ever came up with a plan for how to convert defaults into expression syntax?

I think now that we're handling spec-defined defaults outside of expression evaluation (i.e. in a `catch` block in the error-handling wrapper function), it would probably be simplest to handle user-defined defaults that way, too.  Including functions' defaulting semantics in the converted expression would be possible, but undesirably complicated.

jfirebaugh:
Shelving this for now. Conversion to JSON intermediate format isn't fast enough, and it turned out to be difficult and ugly to implement direct conversion to `Expression` subclasses.

PR #5814, opened 2017-12-05
Register {Alpha,RGBA}Image for transfer

jfirebaugh:
Now that #5702 landed, we can remove the workarounds we had in place where we were avoiding using classes, `instanceof`, etc. for things that needed to be transferred to and from workers. One such case was `AlphaImage`/`RGBAImage`, which is nice because not having these be actual classes was confusing and led to issues like #5414.

### DIFF
--- a/src/data/dem_data.js
+++ b/src/data/dem_data.js
@@ -96,7 +96,7 @@ class DEMData {
     }
 
     getPixels() {
-        return RGBAImage.create({width: this.level.dim + 2 * this.level.border, height: this.level.dim + 2 * this.level.border}, new Uint8Array(this.level.data.buffer));
+        return new RGBAImage({width: this.level.dim + 2 * this.level.border, height: this.level.dim + 2 * this.level.border}, new Uint8Array(this.level.data.buffer));
     }
 
     backfillBorder(borderTile: DEMData, dx: number, dy: number) {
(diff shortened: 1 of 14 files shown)


jfirebaugh:
I'll fix the code duplication, good catch. What do you mean by making them true instances of their respective classes? I think they will be, after this change -- the [deserializer]() will handle "rehydrating" the instances from a pure JSON representation:

https://github.com/mapbox/mapbox-gl-js/blob/ffb73f1eec6e5281eae3f5406b38b17318b398d0/src/util/web_worker_transfer.js#L225

mollymerp:
hmm yes, but it doesn't look like the data for the icon/glyph atlas images are actually deserialized in that function unless I'm mistaken. I came across this when playing around with the branch and the glyph and icon atlases were not instances of `RGBAImage` or `AlphaImage` when creating their `Texture`s

jfirebaugh:
Oh, you're totally right. I had a mistaken assumption about how serialize/deserialize were used: I thought they were applied transparently at the level of actor.js or so, but it looks like they need to be called explicitly at a higher level. @anandthakker -- what's your thinking here? Could we make serialization fully automatic?

anandthakker:
@jfirebaugh yeah, I think we should eventually get to fully automatic serialization.  I punted on it in #5702 because using the new serialization mechanism for the `setLayers` message was looking like a pretty big project on its own.  Just added a ticket to track this: https://github.com/mapbox/mapbox-gl-js/issues/5823

PR #5822, opened 2017-12-06
Handle thrown errors gracefully in integratin tests

anandthakker:
Currently, if a bug causes an uncaught error, it crashes the integration tests. (And since a test's results are only printed to the console after it completes, determining which test caused the error means searching through the alphabetical directory listing).

### DIFF
--- a/src/util/web_worker.js
+++ b/src/util/web_worker.js
@@ -49,8 +49,12 @@ class MessageBus implements WorkerInterface, WorkerGlobalScopeInterface {
 
     postMessage(data: Object) {
         setImmediate(() => {
-            for (const listener of this.postListeners) {
-                listener({data: data, target: this.target});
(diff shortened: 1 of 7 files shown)


src/util/web_worker.js
```diff
@@ -49,8 +49,12 @@ class MessageBus implements WorkerInterface, WorkerGlobalScopeInterface {
...
     postMessage(data: Object) {
         setImmediate(() => {
-            for (const listener of this.postListeners) {
-                listener({data: data, target: this.target});
+            try {
+                for (const listener of this.postListeners) {
+                    listener({data: data, target: this.target});
+                }
+            } catch (e) {
+                console.error(e);
```
anandthakker: Note that this is the non-browser stub for the web worker.

PR #5840, opened 2017-12-11
Improve typing for ==, != expressions

### LINKED ISSUE #5761: How to test for nulls with new expression syntax?
I'm not sure if this is a support request, a comment on the new API or a feature request.

I'm loading a GeoJSON layer which contains some nulls. I want to create a color property which displays the null values differently from other values. (Yes, I could create two layers and a filter, as a workaround). My code looks like this:

```
        paint: {
            "circle-color": 
                ["case",
                    ["all", 
                        ["!=", ["get", "Enrolments_2017_Total"], null],
                        ["!=", ["get", "Capacity_Total"], null]],
                    [
                        "rgb",
                        ["max",
                            0,                    
                            ["min", 
                                255, 
                                ["+", 
                                    -200, 
                                    ["*", 
                                        400, 
                                        ["/", 
                                            ["get", "Enrolments_2017_Total"], 
                                            ["get", "Capacity_Total"]]]]]],
                        0,
                        200
                    ],
                    "yellow"
                ],
```

### Problem:

If I use `["!=", ["get", "Enrolments_2017_Total"], null],`, I get an error. 

> Error: layers.school-points.paint.circle-color[1][1]: Expected arguments of type (number, number) | (string, string) | (boolean, boolean) | (null, null), but found (value, null) instead.

Can this be right? Is it literally saying you can only compare a null literal to another null literal (`["!=", null, null]`)? Um, why would you want to do that? :)

If I use `["has", "Enrolments_2017_Total"]` then that test always passes, because every feature "has" that property, it's just sometimes `null`.

If I use `[">", ["get", "Enrolments_2017_Total"], 0],`

I get: 

> Error: layers.school-points.paint.circle-color[1][1]: Expected arguments of type (number, number) | (string, string), but found (value, number) instead.`

Is there something I'm missing here?

### LINKED ISSUE #5835: Automatically type-annotate arguments to "==", "!=" expressions
## Motivation

After 2 hours of experimentation, I relaised that a simple tag filter looks like this:
 ["all",["has","amenity"], ["==", ["string",["get", "amenity"]], "toilets"]]

because I have to ensure first that amenity exists, than making its value string, in order to avoid type errors.
This should not be so difficult, and definitely people will not love expresssions, if this very common use case is so difficult to achieve.

## Design Alternatives

["==","amenity","toilets"]
I think, this should work out of the box. Because properties is a special object, so I suggest a direct access, and various type issues should not be handled in the forntend code.

(Why on Earth, "get","something" is a Value, when it is in fact a "String"??? :) )

I could get along with:
["==",["get","amenity"],"toilets"]

## Design
no idea

### Mock-Up
no idea

### Concepts
no idea

### Implementation
no idea

anandthakker:
Closes #5761
Closes #5835

The overloads of == and != are now effectively:
 - `(T1: Comparable, T2: Comparable) => boolean { T1 == T2 }`
 - `(Comparable, value) => boolean`
 - `(value, Comparable) => boolean`

Where `Comparable = string | number | boolean | null`, and evaluation semantics for the `value` cases are equivalent to Javascript's strict equality (`===`/`!==`), i.e., when the `value` argument's type doesn't match that of the `Comparable` argument, `==` evaluates to `false`, `!=` to true.

An alternative, stricter approach we discussed was to keep the signatures above, but to produce an runtime error in the case of a type mismatch between the `value` and `Comparable` arguments -- essentially inserting a type assertion wrapper around the `value` argument.  We went with the less strict approach because:
 - it's consistent with the semantics of the pre-existing `==`, `!=` filters
 - it enables using `==`/`!=` to test for `null` (#5761)


src/style-spec/expression/definitions/equals.js
```diff
@@ -0,0 +1,76 @@
...
+        return new Equals(op, lhs, rhs);
+    }
+
+    eachChild(fn: (Expression) => void) {
+        fn(this.lhs);
+        fn(this.rhs);
+    }
+
+    possibleOutputs() {
+        return this.lhs.possibleOutputs().concat(this.rhs.possibleOutputs());
```
jfirebaugh: Should be `return [true, false]`.

anandthakker: 🤦‍♂️

PR #5845, opened 2017-12-12
Fix image source performance regression

### LINKED ISSUE #5752: performance regression between v0.32 and v0.42
On migrating my webapp from mapbox gl version 0.32 to 0.42 I recognized a significant performance drop. The update/frame rate is even worse than before just for panning the map around. On iOS it is below acceptance. Apart from replacing the mapbox engine there is no other modification in my JS code. Any thoughts?

Here is a video to demonstrate the difference: https://vimeo.com/244512199

**mapbox-gl-js version**: 0.32.1 vs. 0.42.2

### Steps to Trigger Behavior

 1. Open Chronoscope with v0.32.1: https://mprove.de/chronoscope/map.html?q=53.51850,9.67088&z=11.53&y=1937&o=0
 2. Open Chronoscope with v0.42.2:  https://mprove.de/chronoscope/map42.html?q=53.51850,9.67088&z=11.53&y=1937&o=0
 3. Move the maps around and compare the update performance

### Expected Behavior
faster or at least same update performance / frame rate for newer mapbox gl version

### Actual Behavior
the update performance is slower for 0.42.2 compared to 0.32.1 before

jfirebaugh:
Fixes #5752. Static image sources were getting the image texture uploaded on every frame. It's unclear to me why the `instanceof window.ImageData` condition was included in `_prepareImage` -- `gl.texSubImage2D` should be needed only for video and animated canvas sources. Does anyone know?

I manually tested:

* canvas.html debug page
* "Add an image" example
* "Add a video" example
* "Animate a series of images" example


src/source/image_source.js
```diff
@@ -207,7 +201,7 @@ class ImageSource extends Evented implements Source {
             this.texture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
         } else if (resize) {
             this.texture.update(image);
-        } else if (image instanceof window.HTMLVideoElement || image instanceof window.ImageData || image instanceof window.HTMLCanvasElement) {
+        } else if (image instanceof window.HTMLVideoElement || image instanceof window.HTMLCanvasElement) {
```
mollymerp: per your comment, should we also check if `animate: true` for `CanvasSource` in this conditional?

jfirebaugh: Good call, added a fix for that as well.

src/source/canvas_source.js
```diff
@@ -135,6 +135,8 @@ class CanvasSource extends ImageSource {
 
         if (Object.keys(this.tiles).length === 0) return; // not enough data for current position
 
+        if (!this._playing && this.texture) return; // Already prepared the static image
```
jfirebaugh: Hmm, actually, this is wrong -- `_prepareImage` needs to be called if the coordinates have changed, to re-initialize `boundsBuffer`.

jfirebaugh:
Ok, I took a different approach (inlining `_prepareImage`), and it should now work correctly for paused canvas and video sources.

PR #5849, opened 2017-12-12
Mark layer "type" property as required

### LINKED ISSUE #5772: Layer "type" property is incorrectly documented as "optional"
[From StackOverflow:](https://stackoverflow.com/questions/47397217/why-is-type-or-ref-required-for-layer-in-mapbox-gl-js/47399228?noredirect=1#comment81769907_47399228)

 > [mapbox-gl-js throws an error if a layer does not have a type. However, type is optional in the spec, and ref is not mentioned in the spec (was apparently taken out at some point). Can somebody clarify this? Should type be required then for the layer in the spec? Or something else?](https://stackoverflow.com/questions/47397217/why-is-type-or-ref-required-for-layer-in-mapbox-gl-js/47399228?noredirect=1#comment81769907_47399228)

As reported, the layer `type` property is documented as "optional."

screen shot 2017-11-29 at 1 47 07 pm

(Naively changing the style reference causes a few failing unit tests.)

jfirebaugh:
But special-case it for validation, for backward compatibility with "ref" mechanism.

Fixes #5772.

### DIFF
--- a/src/style-spec/validate/validate_layer.js
+++ b/src/style-spec/validate/validate_layer.js
@@ -5,6 +5,7 @@ const validateObject = require('./validate_object');
 const validateFilter = require('./validate_filter');
 const validatePaintProperty = require('./validate_paint_property');
 const validateLayoutProperty = require('./validate_layout_property');
+const validateSpec = require('./validate');
 const extend = require('../util/extend');
 
 module.exports = function validateLayer(options) {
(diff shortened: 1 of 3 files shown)


src/style-spec/validate/validate_layer.js
```diff
@@ -71,10 +71,14 @@ module.exports = function validateLayer(options) {
         }
     }
 
+    // https://github.com/mapbox/mapbox-gl-js/issues/5772
+    const valueSpec = extend({}, styleSpec.layer,
+        {type: extend({}, styleSpec.layer.type, {required: false})});
```
lucaswoj: This is a little on the hacky side for my taste...

What do you think about adding a no-op function for `type` in `objectElementValidators` instead?

jfirebaugh: It can't be a no-op because we need to validate the other schema properties, and it requires an additional modification, so I think it comes out as a wash in terms of hackiness. Happy to go with either approach though -- take a look at see which one you prefer.

src/style-spec/validate/validate_object.js
```diff
@@ -47,6 +47,11 @@ module.exports = function validateObject(options) {
     }
 
     for (const elementSpecKey in elementSpecs) {
+        // Don't check `required` when there's a custom validator for that property.
+        if (elementValidators[elementSpecKey]) {
+            continue;
+        }
```
lucaswoj: This makes intuitive sense to me. I like this approach. 👍

PR #5850, opened 2017-12-12
Add test for pitched and zoomed image source

asheemmamoowala:
Render test for https://github.com/mapbox/mapbox-gl-native/pull/10677


asheemmamoowala:
It's zoomed way in on the image. In native, without a fix, the image will not be displayed.

PR #5870, opened 2017-12-15
[WIP] updates to hillshade shader to improve opaque hillshade styling

nickidlugash:
@mollymerp per chat, I made a few adjustments to the hillshade shader to allow for a fuller range of opacities. The changes mainly involve reworking the `intensity` value, and adjusting some of the other calculations that were causing the color values to be out of range, making the hillshading look "over-exposed."

Other changes in addition to the ones noted in the comments:

- Remove sin(polar) from `shade_color` calculation for now – the way we calculate hillshading doesn't seem to benefit from using this value, and the hardcoded value did not allow users to create more opaque styling. If we do want to add it back, we could maybe make it a variable that affects the balance between shade and accent color?

- Adjusted initial `slope` calculation to move the scale factor inside the arctan function. Not totally sure if this is correct, but this way the upper limit of slope values is maintained at 0.5PI.

- Adjusted `shade_color` calculcation to use `sin(slope)` instead of `slope`, since slope values can be >1, which was making very steep areas look overexposed.

- I thought the accent color was overall too prominent for the default, so I changed:
 `clamp((1.0 - accent) * 2.0, 0.0, 1.0)` to 
`1.0 - accent`
(You can still adjust the proportion of accent to shade by adjusting alpha values in the style).

Please review! (Code still needs some clean-up.)


src/shaders/hillshade.fragment.glsl
```diff
@@ -26,10 +27,23 @@ void main() {
     float azimuth = u_light.y + PI;
     float polar = u_light.z;
 
+    // We scale the slope exponentially based on intensity, using a calculation similar to 
+    // the exponential interpolation function in the style spec: 
+    // https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/expression/definitions/interpolate.js#L217-L228
+    // so that higher intensity values create more opaque hillshading.
+    slope = intensity != 0.5 ?
```
mollymerp: i'm in favor of creating a new variable name instead of mutating `slope`

src/shaders/hillshade.fragment.glsl
```diff
@@ -26,10 +27,23 @@ void main() {
     float azimuth = u_light.y + PI;
     float polar = u_light.z;
 
+    // We scale the slope exponentially based on intensity, using a calculation similar to 
+    // the exponential interpolation function in the style spec: 
+    // https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/expression/definitions/interpolate.js#L217-L228
+    // so that higher intensity values create more opaque hillshading.
+    slope = intensity != 0.5 ? 
+        ((pow(1.875 - intensity * 1.75, slope) - 1.0) / (pow(1.875 - intensity * 1.75, (0.5 * PI)) - 1.0)) * (0.5 * PI)
```
mollymerp: I think it'd be helpful to have more variables here to make this more readable( i.e `float base = 1.875 - intensity * 1.75;`)

src/shaders/hillshade.fragment.glsl
```diff
@@ -26,10 +27,23 @@ void main() {
...
+    // The accent color is calculated with the cosine of the slope while the shade color is calculated with the sine
+    // so that the accent color's rate of change eases in while the shade color's eases out.
     float accent = cos(slope);
-    vec4 accent_color = intensity * clamp((1.0 - accent) * 2.0, 0.0, 1.0) * u_accent;
+    // We multiply both the accent and shade color by a clamped intensity value
+    // so that intensities >= 0.5 do not additionally affect the color values
+    // while intensity values < 0.5 make the overall color more transparent.
+    vec4 accent_color = (1.0 - accent) * u_accent * clamp(intensity * 2.0, 0.0, 1.0);
     float shade = abs(mod((aspect + azimuth) / PI + 0.5, 2.0) - 1.0);
-    vec4 shade_color = mix(u_shadow, u_highlight, shade) * slope * sin(polar) * mix(0.0, 2.0, intensity);
```
mollymerp: if we're not using the polar angle anymore for rendering, we should remove it (L26)

debug/terrain-rendering.html
```diff
@@ -51,7 +51,13 @@
         {
             "id": "terrain",
             "source": "rasterTerrain",
-            "type": "hillshade"
+            "type": "hillshade",
+            "paint": {
+                //"hillshade-shadow-color": "hsla(230, 80%, 30%, 1)",
```
mollymerp: ✂️ before merging

src/shaders/hillshade.fragment.glsl
```diff
@@ -14,9 +14,10 @@ void main() {
 
     vec2 deriv = ((pixel.rg * 2.0) - 1.0);
 
-    // we multiply the slope by a scale factor based on the cosin of the pixel's approximate latitude
+    // We multiply the length by a scale factor based on the cosin of the pixel's approximate latitude
     float scaleFactor = cos(radians((u_latrange[0] - u_latrange[1]) * (1.0 - v_pos.y) + u_latrange[1]));
-    float slope = atan(1.25 * length(deriv)) / scaleFactor;
+    // We also multiply the length by an arbitrary z-factor of 1.25
+    float slope = atan(1.25 * length(deriv) / scaleFactor);
```
mollymerp: Great catch! I think this is correct based on https://gis.stackexchange.com/questions/14750/using-srtm-global-dem-for-slope-calculation/40464#40464 which AJ referenced in #4807 
>Compute the slope (as rise:run or a percent) using a Mercator projection, then divide the result by the cosine of the latitude.

and before we were dividing the slope as an _angle_ not a percent 😳

nickidlugash:
@mollymerp thanks for the comments and making the updates! Looks great!

Not sure if this matters or not, but all the "combination" render tests are named as though they expect the hillshading to be translucent, but these styles use the value `"hillshade-exaggeration": 1`, which is now basically fully opaque. /cc @lbud

PR #5888, opened 2017-12-19
Allocate static modes just once

lbud:
Changes static modes (`[ColorMode/StencilMode/DepthMode].disabled` / `ColorMode.alphaBlended` / `ColorMode.unblended`) to be allocated on the relevant mode class export rather than reallocated with methods on every use.

Sadly this doesn't move the paint benchmark at all 😢

### DIFF
--- a/src/gl/color_mode.js
+++ b/src/gl/color_mode.js
@@ -20,19 +20,15 @@ class ColorMode {
 
     static Replace: BlendFuncType;
 
-    static disabled(): ColorMode {
-        return new ColorMode(ColorMode.Replace, Color.transparent, [false, false, false, false]);
-    }
-
-    static unblended(): ColorMode {
-        return new ColorMode(ColorMode.Replace, Color.transparent, [true, true, true, true]);
-    }
-
-    static alphaBlended(): ColorMode {
-        return new ColorMode([ONE, ONE_MINUS_SRC_ALPHA], Color.transparent, [true, true, true, true]);
-    }
+    static disabled: ColorMode;
+    static unblended: ColorMode;
+    static alphaBlended: ColorMode;
 }
 
 ColorMode.Replace = [ONE, ZERO];
+
+ColorMode.disabled = new ColorMode(ColorMode.Replace, Color.transparent, [false, false, false, false]);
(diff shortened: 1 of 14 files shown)


jfirebaugh:
This is mutating one of the values in place:

https://github.com/mapbox/mapbox-gl-js/blob/2124161ebba0c17d764e2e7f78e2d34a283e9bae/src/render/draw_heatmap.js#L39

Flow can catch some cases like this if you declare the type as, say, `$ReadOnly<ColorMode>`. This is kind of like `const` in C++. It doesn't happen to catch this particular case because of how `util.extend` is typed, but we should do it anyway.

src/gl/color_mode.js
```diff
@@ -20,19 +20,15 @@ class ColorMode {
...
-    }
-
-    static unblended(): ColorMode {
-        return new ColorMode(ColorMode.Replace, Color.transparent, [true, true, true, true]);
-    }
-
-    static alphaBlended(): ColorMode {
-        return new ColorMode([ONE, ONE_MINUS_SRC_ALPHA], Color.transparent, [true, true, true, true]);
-    }
+    static disabled: ColorMode;
```
jfirebaugh: Type this (and `unblended`, `alphaBlended`, `DepthMode.disabled`, etc.) as `$ReadOnly` too.

src/render/draw_heatmap.js
```diff
@@ -29,14 +29,14 @@ function drawHeatmap(painter: Painter, sourceCache: SourceCache, layer: HeatmapS
...
-        context.setStencilMode(StencilMode.disabled());
+        context.setStencilMode(StencilMode.disabled);
 
         bindFramebuffer(context, painter, layer);
 
         context.clear({ color: Color.transparent });
 
         // Turn on additive blending for kernels, which is a key aspect of kernel density estimation formula
-        context.setColorMode(util.extend(painter.colorModeForRenderPass(), {
+        context.setColorMode(util.extend({}, painter.colorModeForRenderPass(), {
```
jfirebaugh: This is better written as `new ColorMode([gl.ONE, gl.ONE], Color.transparent, [true, true, true, true])` I think.

Issue #5887, opened 2017-12-18
'queryRenderedSymbols' results invalid between tile reload and next rendered frame


ChrisLoer:
Consider a currently loaded tile A, with symbol X displaying on screen and in the collision index, then trigger a reload of tile A.

1. When the reloaded tile result arrives on the foreground, `Tile#loadVectorData` will get called, which will reset the tile's `featureIndex` to a new index from the result. The new index won't have a `collisionIndex` set yet.
2. `SourceCache#_tileLoaded` immediately schedules a "full placement"
3. On the next `Map#render`, full placement will run, and update the `collisionIndex` of all tiles

If someone queries the map between steps 2 and 3, symbol X _will_ be present in tile A's featureIndex, but it won't show up in the query results because the featureIndex isn't linked to a collisionIndex.

The invalid state should only last milliseconds, but if something is querying the map frequently (for instance, the `mouseover` handler), and it happens to hit this gap, you'll see some form of flickering.

/cc @ansis @andrewharvey @livemixlove

tobrun:
Linking https://github.com/mapbox/mapbox-gl-native/issues/8799 as the gl-native equivalent.

ChrisLoer:
> The invalid state should only last milliseconds, but if something is querying the map frequently (for instance, the mouseover handler), and it happens to hit this gap, you'll see some form of flickering.

This is what happened in closely-related issue #5506. We should get this fixed in the next release.

ChrisLoer:
Fixed in #6497.

PR #5890, opened 2017-12-19
port CollisionIndex and Placement changes from -native

ansis:
This ports changes to `CrossTileSymbolIndex` and `Placement` from -native (https://github.com/mapbox/mapbox-gl-native/pull/10436) back to -js. 

---

Previously symbol opacities were stored directly on `SymbolInstance`s. Whenever a tile was added or removed `CrossTileSymbolIndex` would do a bunch of calculations and copy these opacities directly between `SymbolInstance`s, which it would hold references to. The approach was really stateful. The new approach untangles things into four more distinct steps:

- match symbols across tiles and assign matching `crossTileIDs` to matching symbols (`CrossTileIndex`)
- calculate boolean placement values for each `crossTileID` in the current set of tiles (`Placement.placeLayer`)
- combine these boolean placement values with the previous values to calculate opacities (`Placement.commit`)
- write these placement and opacity values to the buffers of current tiles by matching based on `crossTileID` (`Placement.updateLayerOpacities`)

A stale `Placement` can be used to update newly loaded buckets. This lets us avoid doing a full, expensive placement on every frame where a tile gets loaded. It uses the old values and updates them after a new Placement (split over several frames) has been produced.

---

## Launch Checklist

 - [x] post benchmark scores https://bl.ocks.org/anonymous/raw/f5ce771d18384774bead7f8a06f1af7f/
The most critical benchmark for these changes is `Paint` and the results look good.
 - [x] manually test a `setData` symbol example

I still need to finish writing new tests but the pr is at the point where it could be reviewed.


ChrisLoer:
Awesome! I'm just starting to work through the PR, should get more comments in tomorrow.

It appears to me that this fixes part of https://github.com/mapbox/mapbox-gl-js/issues/5887 (in that tiles never "lose" their collision index because they don't hold references to the collision index any more), but maybe makes more of a problem in the case when a `FeatureIndex` changes during tile reload? Because we may now keep using the old collision index for many frames (until the next placement happens), it could be more problematic if the indices in the collision index don't match up with the feature index. I wonder if it might even trigger an error if the new feature index has fewer features and `filterMatching` indexes off the end of the array...

ansis:
> I wonder if it might even trigger an error if the new feature index has fewer features and filterMatching indexes off the end of the array...

Yes... I think so. Would adding a `bucketInstanceId` check be good enough? In the case of a bucket update and stale placement it would return nothing.

src/style/placement.js
```diff
@@ -1,25 +1,26 @@
 // @flow
```
ChrisLoer: Can we rename this file to `pauseable_placement.js` so the name doesn't collide with `src/symbol/placement.js`?

ChrisLoer:
Haven't figured out the problem yet, but just panning around I see sometimes placement doesn't seem to finish (?) until some later map interaction:

missing_placement

Also on zooming out I'll see symbols disappear:

spurious_duplicates


(This is running against 7bb3d455e64b7cf0554bfec2dcadf3e8d0831cfd)

src/symbol/placement.js
```diff
@@ -0,0 +1,399 @@
...
+
+        const iconWithoutText = !bucket.hasTextData() || layout.get('text-optional');
+        const textWithoutIcon = !bucket.hasIconData() || layout.get('icon-optional');
+
+        for (const symbolInstance of bucket.symbolInstances) {
+            if (!seenCrossTileIDs[symbolInstance.crossTileID]) {
+
+                let placeText = false;
+                let placeIcon = false;
+                //let offscreen = true;
```
ChrisLoer: Is the plan to hook this back up later?

src/symbol/cross_tile_symbol_index.js
```diff
@@ -1,11 +1,12 @@
...
-const assert = require('assert');
 
 import type {OverscaledTileID} from '../source/tile_id';
 import type {SymbolInstance} from '../data/bucket/symbol_bucket';
+import type SymbolBucket from '../data/bucket/symbol_bucket';
+import type StyleLayer from '../style/style_layer';
+import type Tile from '../source/tile';
 
 /*
     The CrossTileSymbolIndex generally works on the assumption that
```
ChrisLoer: This comment needs updating.

ansis:
70eb21581b32ff870de262caec48bf355060345b should fix the missing labels

src/symbol/cross_tile_symbol_index.js
```diff
@@ -85,60 +82,75 @@ class TileLayerIndex {
...
         for (const zoom in this.indexes) {
-            minZoom = Math.min((zoom: any), minZoom);
-            maxZoom = Math.max((zoom: any), maxZoom);
+            minZoom = Math.min(+zoom, minZoom);
+            maxZoom = Math.max(+zoom, maxZoom);
         }
 
-        const tileIndex = new TileLayerIndex(tileID, symbolInstances);
-
         // make all higher-res child tiles block duplicate labels in this tile
```
ChrisLoer: These two loops (child index and parent index) still have structure and comments from when child and parent indexes had different blocking logic. Now that the logic is really just "run `findMatches` across all zoom levels", maybe we could express this more clearly with one loop that just has a branch for looking up child vs. parent indices?

src/style/style.js
```diff
@@ -937,44 +942,76 @@ class Style extends Evented {
...
+            if (styleLayer.type !== 'symbol') continue;
+
+            if (!layerTiles[styleLayer.source]) {
+                const sourceCache = this.sourceCaches[styleLayer.source];
+                layerTiles[styleLayer.source] = sourceCache.getRenderableIds()
+                    .map((id) => sourceCache.getTileByID(id))
+                    .sort((a, b) => (b.tileID.overscaledZ - a.tileID.overscaledZ) || (a.tileID.isLessThan(b.tileID) ? -1 : 1));
+            }
+
+            if (this.crossTileSymbolIndex.addLayer(styleLayer, layerTiles[styleLayer.source])) symbolBucketsChanged = true;
```
ChrisLoer: How about `symbolBucketsChanged |= addLayer(...` instead?

ChrisLoer:
> 70eb215 should fix the missing labels
:+1: Looking good now.

ChrisLoer:
> Would adding a bucketInstanceId check be good enough? In the case of a bucket update and stale placement it would return nothing.

It would work to prevent errors, but wouldn't that still exhibit the behavior from https://github.com/mapbox/mapbox-gl-js/issues/5716#issuecomment-352582854? Ideally, I think the results of `queryRenderedFeatures` should only change when a new placement is committed. For this PR, though, it's probably OK to just make sure we don't error out.

src/symbol/cross_tile_symbol_index.js
ansis: updated in 8e20dd3

src/style/style.js
ansis: I think a regular `if` is clearer than cleverly using a bitwise or

ansis: and flow expects numbers not booleans for `|=`

ChrisLoer: OK, how about just bringing the `symbolBucketsChanged = true` down to its own line? My reasoning is that when someone is scanning the logic looking for where assignments happen, they're easiest (most common) to see when they start the line they're on.

ansis: good point, would this alternative be better than a separate line?
```js
const layerBucketsChanged = this.crossTileSymbolIndex.addLayer(styleLayer, layerTiles[styleLayer.source]);
symbolBucketsChanged = symbolBucketsChanged || layerBucketsChanged;
```

ChrisLoer: Yeah that's good.

ChrisLoer:
It looks like getting rid of full placement on tile loading is working to make the ten-second pitched rotation animation case significantly smoother (looks like 2 instances of stutter after vs. about 15 before):

[fps_meter_during_rotation.zip](https://github.com/mapbox/mapbox-gl-js/files/1576674/fps_meter_during_rotation.zip)

🎉 !

ansis:
@ChrisLoer can you review c47acaa, d44241b, ba0dc8a, 98dbc7f which are meant to fix flickering bugs related to data and style updates?

src/style/style.js
```diff
@@ -937,44 +942,81 @@ class Style extends Evented {
...
+
+            if (this.pauseablePlacement.isDone()) {
+                const placement = this.pauseablePlacement.placement;
+                placementChanged = placement.commit(this.placement, browser.now());
+                if (!this.placement || placementChanged || symbolBucketsChanged) {
+                    this.placement = placement;
+                    this.collisionIndex = this.placement.collisionIndex;
+                    this.placement.setRecent(browser.now());
+                } else {
+                    this.placement.setRecent(browser.now());
```
ChrisLoer: This was tricky for me to read -- it looks like `setRecent` is just being duplicated, it's easy to miss that `setRecent` is going to set `stale` to false, and confusing that we want to undo it right away (because we want to update `_recentUntil` for throttling, right?).

Could we just save something like `inProgressPlacementStale` at the top and set it at the bottom? It might be clearer.

Or would it make sense to simplify this just by always committing the pausable placement? Do we expect there's some performance win from keeping the original placement object unmodified in the case nothing changes?

src/symbol/placement.js
```diff
@@ -0,0 +1,414 @@
...
+    }
+    isHidden() {
+        return this.text.isHidden() && this.icon.isHidden();
+    }
+}
+
+class JointPlacement {
+    text: boolean;
+    icon: boolean;
+    // offscreen = outside viewport, but within CollisionIndex::viewportPadding px of the edge
```
ChrisLoer: By the time it gets here "offscreen" now means something like "skip fade animation" since it's the combination of "offscreen" and "justReloaded". We should come up with a new name to account for that once the two concepts are combined.

PR #5898, opened 2017-12-20
DDS heatmap-radius

jfirebaugh:
Allow `heatmap-radius` to use a property function or data expression.


mourner:
Ha, I was looking into the same thing today but you beat me to it :)

Issue #4271, opened 2017-02-14
Drawing example very jittery in Firefox 


jaapster:
Drawing and updating a polygon in Firefox using the mapbox-gl-draw example code is very jittery at larger map sizes.

Platform: macOS Sierra on a MacBook Pro mid 2015

Mapbox SDK version: 0.32.1

### Steps to trigger behavior

1. Create an HTML file containing the code of the mapbox-gl-draw example
2. Open this file in Firefox and set browser size to 1700 x 1000 pixels
3. Draw a polygon
4. Update the polygon by dragging one of it's coordinates

### Expected behavior
A smooth drawing experience

### Actual behavior
The polygon is not changing smoothly. The change is (very) jittery. It feels like the rendered polygon is jumping back and forth between the current state and previous states.

1ec5:
Possibly related? #4060

ravraponi:
I have found that the Firefox version 51 is very jittery with the drawing tool.  The beta or developer version is better.

jaapster:
I just tested in FF Developer Edition on MacOS and it performs even worse.

ChiefKleef:
Experiencing the same issue with 2 different mapbox web applications. Any change to the drawing or updating of data in-session (ex: for hover effects) is slow and jittery in Firefox and extremely slow internet explorer (but is performant in Chrome).

eltonjuan:
Just downloaded Firefox v.48b01 and can confirm that this issue was introduced at some point after that. Likely related to WebGL 2.

jaapster:
I just tested in Firefox 57.0 (Quantum) and drawing is not jittery anymore but very slow. The geometry updates seem to be lagging and have to "catch up" with the cursor. This is surprising as tilting and rotating a full screen map in FF 57 is considerably smoother than in Chrome.

I have included screencaptures using my own drawing module.

Chrome:

![chrome](https://user-images.githubusercontent.com/1597572/32946595-73fbb2fe-cb98-11e7-8ece-ba37e90981ee.gif)

Firefox Quantum:

![ff](https://user-images.githubusercontent.com/1597572/32946607-8128b3aa-cb98-11e7-98ef-4649b8a15eab.gif)

The more setData requests a source receives, the more it appears to "choke up" in Firefox.

mourner:
I just tested this on Firefox 61 and it seems to work smoothly. 🎉 Let's close this for now then.

mourner:
[...]

PR #5902, opened 2017-12-20
Coalesce high-frequency 'setData' calls

ChrisLoer:
This is an exploratory PR motivated by various example animations provided in https://github.com/mapbox/mapbox-gl-js/issues/5716. The common theme was using high-frequency `setData` calls to animate items in a GeoJSON source.

When you send `setData` calls to the source faster than they can be processed, you end up with an unbounded backlog of requests on the GeoJSON worker, and every request gets processed in order (also several `reloadTile` requests get added to the backlog for every `setData`) no matter how far behind you are. Non-GeoJSON requests to the same worker may also starve.

This PR introduces a coalescing mechanism on `setData` so that the worker will always finish any `loadData` call it starts, and it will always process the last `loadData` it receives, but it is allowed to discard calls that arrive while it is already in the middle of processing an earlier request.

I couldn't figure out a good way to self-send a "coalesced" message so I ended up round-tripping a "coalesce" message back to the parent, which adds a little latency but I think should still be correct.

The `set_data.html` debug page contains a relatively slow `setData` animation that exercises this functionality, and there are currently log statements that record how many calls are getting coalesced.

On my machine, this demo page works _ok_ without the coalescing, and coalescing only seems to skip roughly half of the `setData` calls, so maybe this isn't that important an optimization. But just upping the size of the GeoJSON feature set is enough to make the animation start developing a large backlog if it doesn't have coalescing enabled.

/cc @ansis @jfirebaugh @andrewharvey

### DIFF
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -183,6 +183,14 @@ class GeoJSONSource extends Evented implements Source {
         // implementation
         this.workerID = this.dispatcher.send(`${this.type}.loadData`, options, (err) => {
             this._loaded = true;
+            // Any `loadData` calls that piled up while we were processing
+            // this one will get coalesced into a single call when this
+            // 'coalesce' message is processed.
+            // We would self-send from the worker if we had access to its
+            // message queue. Waiting instated for the 'coalesce' to round-trip
+            // through the foreground just means we're throttling the worker
+            // to run at a little less than full-throttle.
+            this.dispatcher.send(`${this.type}.coalesce`, null, null, this.workerID);
             callback(err);
         }, this.workerID);
     }
--- a/src/source/geojson_worker_source.js
+++ b/src/source/geojson_worker_source.js
@@ -66,6 +66,11 @@ function loadGeoJSONTile(params: WorkerTileParameters, callback: LoadVectorDataC
     });
 }
 
+export type SourceState =
+    | 'Idle'            // Source empty or data loaded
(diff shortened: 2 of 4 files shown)


ChrisLoer:
> Sounds related to #4271 and #3692?

Yup. I didn't know about #3692, but it sounds like I've implemented something similar to https://github.com/mapbox/mapbox-gl-js/issues/3692#issuecomment-267403692. Maybe this is enough to solve all of that problem?

ChrisLoer:
> I don't notice much improvement in terms of the draggable symbol being 100% in sync/as fast as the cursor

Yeah, fundamentally using GeoJSON for this kind of animation can be problematic (re-uploading/re-parsing the entire GeoJSON data set for every motion). Longer term we're interested in adding some sort of interface that would be more efficient for this type of animation.

ryanbaumann:
Nice @ChrisLoer.  Going to test this branch for common animation use cases cc @cmtoomey

anandthakker:
👏

mb12:
@ChrisLoer and @anandthakker Would throttling of setdata calls be added to native as well?

ChrisLoer:
@mb12 I haven't looked at it closely, but I think the architecture on native is fairly different (I think the equivalent `GeoJSONSource::setGeoJSON` actually runs synchronously?). So I don't think we'd do a direct port of this change -- although there are probably analogous situations where frequent calls to something like (on iOS) `MGLShapeSource::setShape` could build up a backlog of requests.

PR #5903, opened 2017-12-21
Refactor symbol buffer related types

jfirebaugh:
Better to have a separate `CollisionBuffers` type, rather than trying to shoehorn those duties into `SymbolBuffers`.

### DIFF
--- a/src/symbol/symbol_layout.js
+++ b/src/symbol/symbol_layout.js
@@ -292,12 +292,11 @@ function addTextVertices(bucket: SymbolBucket,
         writingMode,
         anchor,
         lineArray.lineStartIndex,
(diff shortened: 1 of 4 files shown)


src/symbol/symbol_layout.js
```diff
@@ -292,12 +292,11 @@ function addTextVertices(bucket: SymbolBucket,
         writingMode,
         anchor,
         lineArray.lineStartIndex,
-        lineArray.lineLength,
-        bucket.placedGlyphArray);
+        lineArray.lineLength);
 
     // The placedGlyphArray is used at render time in drawTileSymbols
```
ChrisLoer: This comment needs updating.

Issue #5911, opened 2017-12-21
Collision detection perspective ratios are incorrect


ChrisLoer:
If you look at a dense map in the distant field, you can find places where the collision boxes for two labels clearly overlap, but both labels are still showing. Look at "Kenneth Hahn State Recreational Area" and "VIEW PARK-WINDSOR HILLS":

<img width="126" alt="screenshot 2017-12-21 14 42 21" src="https://user-images.githubusercontent.com/375121/34278048-4b302b72-e65d-11e7-9e87-0ca4ff2dde0c.png">

The collision boxes themselves look right, and if you look at their rendering code, you'll see they adjust their size by:

https://github.com/mapbox/mapbox-gl-js/blob/79f0a84383456dc3ee09c668e34ea895aba8e9b2/src/shaders/collision_box.vertex.glsl#L16

On the other hand, the code in `CollisionIndex` is using the opposite ratio (`p[3]` is the same as `camera_to_anchor_distance`):

https://github.com/mapbox/mapbox-gl-js/blob/79f0a84383456dc3ee09c668e34ea895aba8e9b2/src/symbol/collision_index.js#L359

Why is this not completely broken? Well, we end up _dividing_ by the perspective ratio instead of multiplying by it:

https://github.com/mapbox/mapbox-gl-js/blob/79f0a84383456dc3ee09c668e34ea895aba8e9b2/src/symbol/collision_index.js#L62-L66

But `1/(.5 + .5 * y/x)` is not the same as `(.5 + .5 * x/y)`! It is however a close enough approximation when x and y are relatively near each other that we didn't notice! 😅  😬 

Dividing by `tileToViewport` always should have felt fishy -- it turned out to mostly work because we inverted the `textPixelRatio` as well. Inversions all the way down!

I've got a fix that un-inverts everything and makes "collision detection" + "collision debug boxes" + "actually rendered labels" all be consistent (and I believe correct against the expected "attentuates 50% of the effect of perspective scaling" behavior). I'm working through the associated test changes.

Same problem exists in gl-native.

/cc @ansis @jfirebaugh

ChrisLoer:
Fixed with #5913.

PR #5913, opened 2017-12-22
Fix collision detection perspective ratios

ChrisLoer:
Fixes issue #5911.

 - Box and circle collision detection were using incorrectly inverted perspective ratios. [(1/(.5 + .5 * y/x) instead of (.5 + .5 * x/y)].
 - Debug collision circles were using the same incorrect ratio (so they matched actual collision detection).
 - Debug collision boxes were right the whole time (this is how we noticed there was a problem).
 - pixelsToTileUnits incorrectly included perspective ratio: it's used for measuring geometry, not text, so "scale" is relevant but "perspective ratio" isn't.

'collision-pitched-lines' exposed the behavior, but it was too complicated a test for us to be able to understand the problem.

I'll try to construct a regression test that clearly exercises the old vs. new behavior, but probably won't get around to it until the new year.

/cc @ansis @anandthakker @jfirebaugh


ChrisLoer:
Pushed a regression test that isolates the behavior from all the noise in the other collision render tests that are affected by this change.

PR #5938, opened 2018-01-03
Use template strings for validation errors

mourner:
Eliminates the need for the browserified `util` module — we don't need `format` since we have ES6 template strings. Reduces the build a bit:

```bash
# before
orig: 706.92KB
gzip: 171.55KB (24.27%)

#after
orig: 696.23KB
gzip: 168.29KB (24.17%)
```

### DIFF
--- a/src/style-spec/error/validation_error.js
+++ b/src/style-spec/error/validation_error.js
@@ -1,8 +1,6 @@
 
-const format = require('util').format;
-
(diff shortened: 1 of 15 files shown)


mourner:
The build failure looks like a flaky test, but I can't rebuild — Circle doesn't show me that button for some reason 😕

PR #5940, opened 2018-01-03
Minimize circle render tests

jfirebaugh:
After #5927, the next best thing I've found to make render tests faster is to keep them small (64x64) where possible. This PR does so for the circle tests.


jfirebaugh:
Added a couple tests. I think they exposed an alpha blending bug.

image

jfirebaugh:
Alpha bug fixed in #6275. @mourner want to take another look?

PR #5944, opened 2018-01-04
add getter and setter for renderWorldCopies

### LINKED ISSUE #4039: Add setter for renderWorldCopies option to allow it to be changed after Map instantiation
Follow up to https://github.com/mapbox/mapbox-gl-js/pull/3885 

cc @cammanderson

andrewharvey:
Closes #4039

## Launch Checklist

Adds a getter and setter for renderWorldCopies so it can be changed after map initialisation.

Unit tests confirm the getter and setters are in sync.

JS doc included

 - [ ] post benchmark scores

I've manually tested flipping renderWorldCopies and it updates the visuals correctly.

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -97,7 +97,13 @@ class Transform {
     }
 
     get renderWorldCopies(): boolean { return this._renderWorldCopies; }
-    set renderWorldCopies(renderWorldCopies: boolean) {
+    set renderWorldCopies(renderWorldCopies?: ?boolean) {
+        if (renderWorldCopies === undefined) {
+            renderWorldCopies = true;
+        } else if (renderWorldCopies === null) {
+            renderWorldCopies = false;
+        }
+
         this._renderWorldCopies = renderWorldCopies;
     }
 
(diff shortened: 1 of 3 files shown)


andrewharvey:
I think I need some help here!

I guess Flow doesn't like that you can call `setRenderWorldCopies` with null/undefined given it expects a boolean.

However it's important to have the setter consistant with setting via map initialisation, which is ensured by this PR.

`setRenderWorldCopies(null)` is like `false` just like `renderWorldCopies: null`
`setRenderWorldCopies(undefined)` is like `true` just like `renderWorldCopies: undefined`
`setRenderWorldCopies()` is like `true` just like omitting `renderWorldCopies` as the map option.

I just added a second commit to make Flow happy but now eslint doesn't like it. Any suggestions on what I can do?

andrewharvey:
I've added another commit which explicitly handles `null` and `undefined` in the same way as the initialiser. Not sure if this is the right way of doing it but it makes Flow happy and gives the expected outcomes.

I didn't want to get too bogged down in corner cases like passing null and undefined and would be happy with forcing passing either a `true` or a `false` to the setter, but I guess it's good to keep it consistent with the initialiser.

andrewharvey:
> Maybe indicate in the docs what happens in the cases of null or undefined ?

I've added this to the documentation and also included test cases for this since it's now documented.

PR #5955, opened 2018-01-05
Guard against undefined stylesheet.

vicapow:
We're seeing issues in production with stylesheet being undefined. There also seem to be other places were it's access is guarded. see: https://github.com/mapbox/mapbox-gl-js/blob/ca01a1edfc4b0a1b10b85db1a722cc2aa4f6163a/src/style/style.js#L776

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1481,7 +1481,7 @@ class Map extends Camera {
                 now,
                 fadeDuration: this._fadeDuration,
                 zoomHistory: this.style.zoomHistory,
-                transition: util.extend({ duration: 300, delay: 0 }, this.style.stylesheet.transition)
+                transition: this.style.getTransition()
             });
 
             const factor = parameters.crossFadingFactor();


src/ui/map.js
```diff
@@ -1481,7 +1481,7 @@ class Map extends Camera {
                 now,
                 fadeDuration: this._fadeDuration,
                 zoomHistory: this.style.zoomHistory,
-                transition: util.extend({ duration: 300, delay: 0 }, this.style.stylesheet.transition)
+                transition: util.extend({ duration: 300, delay: 0 }, this.style.stylesheet && this.style.stylesheet.transition)
```
jfirebaugh: How about just doing `transition: this.style.getTransition()`?

vicapow: sure

jfirebaugh:
Thanks!

pavelsl:
How can one know when it's going to be released?

pavelsl:
Or may be there is a way to operate another way with map to prevent this issue?

vicapow:
@pavelsl  If you'd like, you can use my folk until this gets fixed. just do `yarn add git+https://github.com/vicapow/mapbox-gl-js.git#6c72c102f0f40ff160ec05ec0b4d6093c71f8d4a`

PR #5956, opened 2018-01-05
Add documentation clarifying stop output value type

davidtheclark:
Earlier today I was trying an expression as the output of a stop value. There were cryptic errors, but I didn't find in the spec whether this was *supposed* to work or not. @jfirebaugh clarified in chat that this shouldn't work, so I thought I'd try amending the docs.

I also corrected a spacing problem.

Before:

screen shot 2018-01-05 at 3 29 10 pm

After:

screen shot 2018-01-05 at 3 27 24 pm

### DIFF
--- a/docs/pages/style-spec.js
+++ b/docs/pages/style-spec.js
@@ -1481,9 +1481,10 @@ export default class extends React.Component {
                                             <div><em className='quiet'>Required (except
                                                 for <var>identity</var> functions) <a href='#types-array'>array</a>.</em></div>
                                             <div>Functions are defined in terms of input and output values. A set of one
-                                                input value and one output value is known as a "stop." Output values can
-                                                be strings, numbers, and booleans. They cannot be functions or
-                                                expressions.
+                                                input value and one output value is known as a "stop." Stop output values
+                                                must be literal values (i.e. not functions or expressions), and appropriate
+                                                for the property. For example, stop output values for a function used in
+                                                the `fill-color` property must be <a href="#types-color">colors</a>.
                                             </div>
                                         </div>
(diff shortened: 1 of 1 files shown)


jfirebaugh:
To be a bit more precise, we could say "Output values must be literal values (i.e. not functions or expressions), and appropriate for the property. For example stop output values for a function used in the `fill-color` property must be colors."

davidtheclark:
@jfirebaugh Thanks for the suggestion. I made those changes.

docs/pages/style-spec.js
```diff
@@ -1478,10 +1478,13 @@ export default class extends React.Component {
...
-                                            <div><em className='quiet'>Required (except for <var>identity</var>
-                                                functions) <a href='#types-array'>array</a>.</em></div>
+                                            <div><em className='quiet'>Required (except
+                                                for <var>identity</var> functions) <a href='#types-array'>array</a>.</em></div>
                                             <div>Functions are defined in terms of input and output values. A set of one
-                                                input value and one output value is known as a "stop."
+                                                input value and one output value is known as a "stop." Stop output values
+                                                must be literal values (i.e. not functions or expressions), and appropriate
+                                                for the property. For example, stop output values for a function used in
+                                                the `fill-color` property must be <a href="#types-color">colors</a>.
```
jfirebaugh: Replace the markdown around `fill-color` with markup.

davidtheclark:
@jfirebaugh It was unclear to me whether to use `<code>` or `<var>` here. I opted for `<code>`, since that seems to be the majority of the references to style property names.

jfirebaugh:
Thanks!

Issue #5941, opened 2018-01-03
Viewport aspect ratio affects debug collision circle thickness


ChrisLoer:
Very narrow (~1:6) viewports -> thin collision circles:
<img width="145" alt="screenshot 2018-01-03 14 47 14" src="https://user-images.githubusercontent.com/375121/34543298-059f6bb2-f095-11e7-9061-485b4b8eaaa5.png">

Very wide (~6:1) viewports -> thick collision circles:
<img width="203" alt="screenshot 2018-01-03 14 46 59" src="https://user-images.githubusercontent.com/375121/34543299-05c82214-f095-11e7-93d4-1506ad8a03f0.png">

It's not a very noticeable effect in most cases -- it only grabbed my attention because I tried to write a test case with a very tall/narrow viewport.

I think the problem is that the stroke-width is hardwired here:

https://github.com/mapbox/mapbox-gl-js/blob/2f16075259d37e5179ea7dcaa50bff60fa0488d1/src/shaders/collision_circle.fragment.glsl#L26

... and needs to be modified based on the varying extrude scale.

/cc @ansis @nickidlugash

ChrisLoer:
Fixed with #5957.

PR #5957, opened 2018-01-05
Make debug collision circle stroke width independent of aspect ratio.

ChrisLoer:
Fixes issue #5941.

12.0 * extrude_scale_length chosen to match previous stroke width for square viewports.
Regression test #5911a tests new functionality because of its narrow viewport.

@nickidlugash If you get a chance could you test out how the circles look? For square viewports, they should look the same as when we originally implemented the collision circles. We might want to tweak the stroke width a little if it turns out we were originally tuning against a non-square viewport.

This should have a trivial port to gl-native -- just need to pull over the shader update.

/cc @ansis


src/shaders/collision_circle.fragment.glsl
```diff
@@ -23,7 +23,7 @@ void main() {
 
     float extrude_scale_length = length(v_extrude_scale);
     float extrude_length = length(v_extrude) * extrude_scale_length;
-    float stroke_width = 3.0;
+    float stroke_width = 12.0 * extrude_scale_length;
```
nickidlugash: @ChrisLoer The viewports I was using when we first implemented this were probably closer to 1:1.5–1:2 than to a square, and subsequently looking at this now, the stroke does seem a bit too thin (it looks good on retina, but doesn't render well on non-retina). I tested out `15.0 * extrude_scale_length` here and it looks better to me.

ChrisLoer:
Updated to `15.0 * extrude_scale_length` based on @nickidlugash 's feedback, and updated debug render tests accordingly.

PR #5959, opened 2018-01-06
Minor corrections in style spec documentation

1ec5:
Made two small changes to the style specification JSON to facilitate better codegen results in mapbox/mapbox-gl-native#10850:

* `icon-offset` uses backticks around the reference to `icon-size` so that `icon-size` can become `iconScale` on iOS/macOS.
* Indicated that `{token}` replacement is also unavailable in data expressions. At least that’s the case in the native implementation; not sure about GL JS.

/cc @jfirebaugh

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -1162,7 +1162,7 @@
     "icon-offset": {
       "type": "array",
       "value": "number",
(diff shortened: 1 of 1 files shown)


1ec5:
Closing in light of #5958 and #5600.

PR #5966, opened 2018-01-08
fix values setting for the Light object

mollymerp:
noticed this when `map.getLight` was returning an empty object – it turns out when a style used all defaults for the `Light` properties, we were never initializing their values with `setLight` because `lightOptions` was undefined. 

open to alternative approaches if this is not the best way to go about this

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style/light.js
+++ b/src/style/light.js
@@ -78,6 +78,12 @@ class Light extends Evented {
     constructor(lightOptions?: LightSpecification) {
         super();
         this._transitionable = new Transitionable(properties);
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Hmm, what exactly is the issue with `getLight()` returning an empty object if all the values are defaulted? Omitting default values is the behavior we have for `paint` and `layout` -- it seems like `light` should behave the same way.

mollymerp:
ah ok – I missed/forgot that this is expected behavior how.

PR #5967, opened 2018-01-09
Introduce 'gpuTiming' map options

ChrisLoer:
I've been working on an "animation benchmarker" that replays animated scenes to collect performance statistics. @jfirebaugh pointed me at the [disjoint_query_timer extension](https://developer.mozilla.org/en-US/docs/Web/API/EXT_disjoint_timer_query) as a way to try to capture meaningful GPU timing stats to combine with CPU timing stats collected using `window.performance`. This PR tries to generalize the frame-timing part of that work. It introduces a new "gpuTiming" Map option which can be 'none' (default), 'frame', or 'layer'.

 - 'frame' records total CPU and GPU time spent rendering every frame

screenshot 2018-01-08 16 27 18

 - 'layer' breaks down CPU and GPU draw call time by layer. Draw calls alone miss a lot of the costs of rendering, so interpret these numbers with caution.

screenshot 2018-01-08 16 25 57

When timing is on, the map fires a `frame-timing` event after each frame.

The 'frame' method gives results that seem to line up decently well with what I see in the Chrome performance timeline. I'm much less confident in what the 'layer' method is measuring -- although the results seem to correlate with other measures of layer cost, the measured costs seem to be an order of magnitude too low. Adding bucket upload times to the per-layer measurements might help, but I suspect it's either a limitation of the query API or my misuse of it.

I haven't handled "disjoint" errors at all, I don't know if that's likely to make a big difference in the "performance testing on a desktop" case:

> Disjoint operations occur whenever a change in the GPU occurs that will
    make the values returned by this extension unusable for performance 
    metrics. An example can be seen with how mobile GPUs need to proactively
    try to conserve power, which might cause the GPU to go to sleep at the 
    lower levers. This means disjoint states will occur at different times on
    different platforms and are implementation dependent.

You can play around with this on the debug page by adding a 'gpuTiming' option and something like:

    map.on('frame-timing', function(data) {
        console.log(data);
    });

/cc @jfirebaugh @lbud @nickidlugash @kkaefer

### DIFF
--- a/src/gl/context.js
+++ b/src/gl/context.js
@@ -61,10 +61,12 @@ class Context {
     extTextureFilterAnisotropic: any;
     extTextureFilterAnisotropicMax: any;
     extTextureHalfFloat: any;
+    timerQueryExt: any;
 
     constructor(gl: WebGLRenderingContext) {
         this.gl = gl;
         this.extVertexArrayObject = this.gl.getExtension('OES_vertex_array_object');
+        this.timerQueryExt = this.gl.getExtension('EXT_disjoint_timer_query');
         this.lineWidthRange = gl.getParameter(gl.ALIASED_LINE_WIDTH_RANGE);
 
         this.clearColor = new ClearColor(this);
--- a/src/render/painter.js
+++ b/src/render/painter.js
@@ -58,13 +58,15 @@ import type VertexBuffer from '../gl/vertex_buffer';
 import type {DepthMaskType, DepthFuncType} from '../gl/types';
 
 export type RenderPass = 'offscreen' | 'opaque' | 'translucent';
+export type GPUTiming = 'none' | 'frame' | 'layer';
 
 type PainterOptions = {
     showOverdrawInspector: boolean,
     showTileBoundaries: boolean,
     rotating: boolean,
     zooming: boolean,
-    fadeDuration: number
+    fadeDuration: number,
+    gpuTiming: GPUTiming
 }
 
 /**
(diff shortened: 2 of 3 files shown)


ChrisLoer:
@jfirebaugh or @lbud, any thoughts on whether to go forward with this?

jfirebaugh:
I think we should go forward with it. Can you post benchmark results?

ChrisLoer:
Huh, I rebased and started to test, and I can't get the extension to work at all across Chrome/Firefox/Safari. Is there a chance they're getting clipped as part of some spectre/meltdown defence? I did switch to a new laptop, but it's pretty similar graphics hardware and I wouldn't expect the drivers to remove support for the extension.

Can anyone else try `[some canvas].getContext("webgl").getExtension('EXT_disjoint_timer_query')` and see if they get a non-null object back?

jfirebaugh:
Yeah, it worked in Chrome 64 but not in 65 (I just updated).

jfirebaugh:
https://bugs.chromium.org/p/chromium/issues/detail?id=820891

ChrisLoer:
😞 I guess we'll just have to leave this shelved for a few months to see if they work something out...

jfirebaugh:
Closing for now; we can revive this if the timer extension is restored.

asheemmamoowala:
The chromium bug was [fixed](https://chromiumdash.appspot.com/commit/cc1732e0386125c0c3aa04690663bfb43e6d7f68) in Chrome 70. 

LGTM

ansis:
Rebasing and tweaking in https://github.com/mapbox/mapbox-gl-js/pull/8829

PR #5973, opened 2018-01-09
don't check if a geojson tile expired (they never do)

ansis:
This removes a condition which is never true.
@ChrisLoer


ChrisLoer:
Oh, sorry I missed this! Thanks for PR-patrolling @mollymerp!

Issue #5776, opened 2017-11-30
Label collision detection preventing labels not shown on the map but according to clamped min-zoom


dieterdreist:
There's maybe a bug in mapbox-gl-js/v0.42.2 (and maybe earlier versions), where the collission detection for labels prevents the rendering of labels in a lower layer where there is a different layer with the same features above, but the features are not rendered (excluded by float Min-Zoom of 8.8)

Description of the situation:
There is a layer B which selects a subset of the features in Layer A.
Layer A is above Layer B in MB Studio.
Layer A has its zoom limited from z8.8 to z14.
Layer B has its zoom limited from z6.5 to z15.

I've triple checked the zoom-dependent styling, and it has nothing to do with it.

Features from Layer B display from z6.5 to z.8.
Then they disappear and reappear when zoom 8.8 is reached (because now it is Layer A which displays them). I can see that disappearing of features in Layer B is caused by Layer A because the features re-appear in Layer B (in Zoom 8-8.8 and above) when I disable Layer A or disable collision prevention in Layer B. Unfortunately, disabling collision prevention is not an option for me. For now I workaround by setting the min-zoom for the source to 9 rather than 8.8

dieterdreist:
Looking at it, the problem could be more concisely rephrased like this:
a layer minzoom float is treated as a clamped integer in collision detection when interacting with other layers, but working fine when it comes to visibility in rendering.

The specs state: 

> minzoom
> Optional number between 0 and 24 inclusive.
> The minimum zoom level on which the layer gets parsed and appears on.

A number is "A number value, often an integer **or floating point** (decimal number).", so this is likely a bug.

eulo:
Bug seems to be present in all 0.42.*

Reproducible below, change version to 41 to see desired outcome.

```HTML
<script src='https://api.tiles.mapbox.com/mapbox-gl-js/v0.42.2/mapbox-gl.js'></script>
<link href='https://api.tiles.mapbox.com/mapbox-gl-js/v0.42.2/mapbox-gl.css' rel='stylesheet' />
<style>
	body { margin:0; padding:0; }
	#map { position:absolute; top:0; bottom:0; width:100%; }
</style>
<div id='map'></div>
<script>
[...]

PR #5974, opened 2018-01-10
Don't place layers hidden by min/maxzoom constraints.

ChrisLoer:
Fixes issue #5776.

`StyleLayer#isHidden` prevents us from drawing layers outside of `minzoom`/`maxzoom`, but there was no equivalent filtering for placing layers. The layers would be filtered out at tile parsing time:

https://github.com/mapbox/mapbox-gl-js/blob/5e44e625ccf6e9d5e6d48011bb11281d1a4a197a/src/source/worker_tile.js#L96-L97

... but that code would only filter at integer zoom levels.

/cc @ansis @asheemmamoowala


src/style/pauseable_placement.js
```diff
@@ -63,7 +63,10 @@ class PauseablePlacement {
         while (this._currentPlacementIndex >= 0) {
             const layerId = order[this._currentPlacementIndex];
             const layer = layers[layerId];
-            if (layer.type === 'symbol') {
+            const placementZoom = this.placement.collisionIndex.transform.zoom;
+            if (layer.type === 'symbol' &&
+                (!layer.minzoom || layer.minzoom <= placementZoom) &&
+                (!layer.maxzoom || layer.maxzoom > placementZoom)) {
```
ansis: a small thing: when `layer.maxzoom === 0` wouldn't be handled correctly because `!layer.maxzoom` would`true`. This seems like a super unlikely but technically valid value.

ChrisLoer: Good point -- but `style_layer.js` and `worker_tile.js` both have the same behavior (treating `maxzoom === 0` as if there's no maxzoom set), so I think I'll err on the side of consistency for now.

Issue #5934, opened 2018-01-03
Changing hillshade-illumination-direction near 0/360 triggers page refresh


stevage:
**mapbox-gl-js version**: 0.43.0

### Steps to Trigger Behavior

 1. Create a hilshade layer with `hillshade-illumination-direction` of 359
 2. Call `map.setPaintProperty(layername, 'hillshade-illumination-direction', 0)`

### Expected Behavior
Almost no visible effect.

### Actual Behavior
Entire map repaints with large visible flicker.

Easily seen here: https://bl.ocks.org/stevage/raw/a4569595d8592181ab09a5a13166be8d

(The current direction is shown in the URL)

Code: https://gist.github.com/stevage/a4569595d8592181ab09a5a13166be8d

mollymerp:
yes I see what you mean – I don't think it's triggering a page refresh though, its just transitioning from 359 -> 0 counterclockwise so the light source does almost a full rotation instead of just moving slightly clockwise to 360/0. 

A workaround right now is to set `transition.duration` to 0 https://www.mapbox.com/mapbox-gl-js/style-spec#root-transition

stevage:
Ah, that seems so obvious now :) Thanks.

PR #5981, opened 2018-01-10
temporarily disable transitions for hillshade-illumination-direction

mollymerp:
temporary workaround for https://github.com/mapbox/mapbox-gl-js/issues/5934
cc @stevage 

light-position wasn't actually transitioning, despite having `transition: true` in the style spec, so this won't change behavior from the current version. 

when I have some spare time I will implement an interpolation function that accounts for light-position and hillshade-illumination-direction for a better solution

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2302,7 +2302,7 @@
       "default": [1.15, 210, 30],
       "length": 3,
       "value": "number",
-      "transition": true,
+      "transition": false,
       "function": "interpolated",
       "zoom-function": true,
       "property-function": false,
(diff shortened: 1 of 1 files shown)


lbud:
`light-position` is supposed to be interpolated using its cartesian coordinates https://github.com/mapbox/mapbox-gl-js/blob/da7471353bbd1ec4226c5db2a307b48b7537a41b/src/style/light.js#L45-L51 so that degree thresholds don't cause a problem there. I would rather fix them than declare them non-transitionable 🙃  https://github.com/mapbox/mapbox-gl-js/issues/5982

Issue #5970, opened 2018-01-09
mapbox-gl does not render all GeoJSON layers


vicapow:
This is only on master and from what I can tell is not an issue with the latest published version, 0.43.0. 

Here's the full example after modifying setstyle.html.

Expected: A red circle layer to be rendered as in https://jsbin.com/kurufojuvi/edit?html,output.

<img width="154" alt="screen shot 2018-01-08 at 8 06 44 pm" src="https://user-images.githubusercontent.com/583385/34704789-6bf12aee-f4b0-11e7-931d-f664a9a1a88f.png">

Actual: No red layer is ever rendered.

<img width="132" alt="screen shot 2018-01-08 at 8 07 35 pm" src="https://user-images.githubusercontent.com/583385/34704791-6e46c9ac-f4b0-11e7-8d90-ff27555bf42d.png">

I was able to track this down to [this specific diff](https://github.com/mapbox/mapbox-gl-js/commit/fa95c19c0160949dec1b482ee7f259d04e51db90): from @ChrisLoer 19 days ago. Can we revert?

````html
<!DOCTYPE html>
<html>
<head>
    <title>Mapbox GL JS debug page</title>
    <meta charset='utf-8'>
    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
    <link rel='stylesheet' href='/dist/mapbox-gl.css' />
    <style>
        body { margin: 0; padding: 0; }
        html, body, #map { height: 100%; }
    </style>
</head>

<body>
<div id='map'></div>
<script src='/dist/mapbox-gl-dev.js'></script>
<script src='/debug/access_token_generated.js'></script>
<script>

var data = {
    type: 'FeatureCollection',
    features: [
        {
            type: 'Feature',
            properties: {},
            geometry: { type: 'Point', coordinates: [ -122.41848707199095, 37.77554439836295 ] }
        }
    ]
};
var style = {
    version: 8,
    name: 'example',
    minzoom: 0,
    maxzoom: 22,
    sources: {
        source1: { type: "geojson", data: data },
        source2: { type: "geojson", data: data },
        source3: { type: "geojson", data: data },
        source4: { type: "geojson", data: data },
        source5: { type: "geojson", data: data }
    },
    layers: [
        {
            id: 'source1-layer',
            source: 'source1',
            type: 'circle',
            paint: { 'circle-radius': 50, 'circle-color': '#ccc' }
        }, {
            id: 'source2-layer',
[...]

PR #5983, opened 2018-01-10
Don't mix coalesce calls from multiple GeoJSON sources (issue #5970)

ChrisLoer:
Fixes issue #5970.

Also: send back an "abandoned" callback for coalesced calls so that foreground doesn't leak callback handles.

Adds 'loadData' unit test for geojson_worker_source that simulates building up a queue of loadData messages.

/cc @ansis @jfirebaugh @anandthakker

### DIFF
--- a/src/source/geojson_worker_source.js
+++ b/src/source/geojson_worker_source.js
@@ -31,6 +31,10 @@ export type LoadGeoJSONParameters = {
     geojsonVtOptions?: Object
 };
 
+export type CoalesceParameters = {
+    source: string
+};
+
 export type LoadGeoJSON = (params: LoadGeoJSONParameters, callback: Callback<mixed>) => void;
 
 export interface GeoJSONIndex {
@@ -84,8 +88,8 @@ export type SourceState =
 class GeoJSONWorkerSource extends VectorTileWorkerSource {
     _geoJSONIndexes: { [string]: GeoJSONIndex };
     loadGeoJSON: LoadGeoJSON;
-    state: SourceState;
-    pendingCallback: Callback<void>;
+    sourceStates: { [string]: SourceState };
+    pendingCallback: Callback<boolean>;
     pendingLoadDataParams: LoadGeoJSONParameters;
 
     /**
@@ -100,7 +104,7 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
         }
         // object mapping source ids to geojson-vt-like tile indexes
         this._geoJSONIndexes = {};
-        this.state = 'Idle';
+        this.sourceStates = {};
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Hrm... not too thrilled with the additional state tracking this solution requires.

What would it take to be able to truly self-send the coalesce method?

ChrisLoer:
I did manual testing against the 2000 circle example from https://github.com/mapbox/mapbox-gl-js/issues/5716#issuecomment-352607067 -- performance feels similar.

ChrisLoer:
> What would it take to be able to truly self-send the coalesce method?

I'm not sure -- I don't think we have a way to directly access the worker's message queue. Maybe we could build a wrapper/shadow of the message queue and push the self-send there? I don't see how we could manage the order correctly though, since we don't have a good way to know when we've temporarily emptied the real message queue (i.e. when it's time to insert our "self sent" messages).

Self-sending would be a simplification, but it wouldn't address either of the issues in this PR -- we'd still have to maintain state separately for sources that share a worker, and we'd still have to send back "abandoned" callbacks to the foreground.

src/source/geojson_worker_source.js
```diff
@@ -120,13 +124,19 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
      * @param params.source The id of the source.
      * @param callback
      */
-    loadData(params: LoadGeoJSONParameters, callback: Callback<void>) {
+    loadData(params: LoadGeoJSONParameters, callback: Callback<boolean>) {
```
jfirebaugh: Don't the `params` and `callback` need to be stored on a per-source basis as well?

I think the root problem here is that `WorkerSource`s should be per-source, not per-source-type. The mapping from source ID to `WorkerSource` instance should be handled by `Worker`.

ChrisLoer: > Don't the params and callback need to be stored on a per-source basis as well?

Oh, of course. Good catch.

> I think the root problem here is that WorkerSources should be per-source, not per-source-type. The mapping from source ID to WorkerSource instance should be handled by Worker.

🤔 That sounds convincing, although I'd like to handle it in a different PR. On a quick inspection, it looks like the `StyleLayerIndex` is the main thing that would remain shared between multiple `WorkerSource`s?

ChrisLoer:
I addressed the params/callback problem, and in the process noticed a further race condition between `removeSource` and `coalesce`. I'm looking now at modifying `worker.js` to store one `WorkerSource` per source, although I'm a little worried about unintended consequences of that change (especially if we're going to do it in v.44).

anandthakker:
@ChrisLoer @jfirebaugh when a worker sends a message to the main thread, the response is queued right along with other main-thread-initiated workers, right?  So another approach to the coalesce message would be to have the worker source send `coalesce` to the main thread and wait for it to bounce back.  Still not a true self-coalesce... but maybe that simulates it slightly better?  Not sure.

ChrisLoer:
@anandthakker My first implementation worked with the worker sending an explicit coalesce message to the main thread -- I think it's pretty similar to the current approach of having the main thread `loadData` callback send `coalesce` to the worker, but just has one extra message (and extra handling code).

anandthakker:
@ChrisLoer this is subsumed by https://github.com/mapbox/mapbox-gl-js/pull/5988 now, right?

PR #5988, opened 2018-01-11
per-SourceType WorkerSource -> per-source-instance WorkerSource

ChrisLoer:
Based on @jfirebaugh 's feedback on #5983, I started exploring instantiating `WorkerSource`s on a per-source basis instead of a per-source-type basis.

The initial set of changes were not too bad, and it allows for a nice simplification of geojson_worker_source that still supports the `setData` coalescing behavior.

My biggest concern is how this will affect custom sources -- requiring a source name to be included (either in parameters or in the "data.type" argument) is probably breaking the implied API?

`removeSource` also needs some re-thinking to handle messages that arrive after a source has been removed (or, moving even more into edge-case territory, to handle messages that were sent to a _previous_ instantiation of a source, which has been removed and then re-added).

The unit tests that handle multiple sources are currently broken, but render/query tests pass.

/cc @jfirebaugh @anandthakker

### DIFF
--- a/src/source/geojson_worker_source.js
+++ b/src/source/geojson_worker_source.js
@@ -33,24 +33,19 @@ export type LoadGeoJSONParameters = {
     geojsonVtOptions?: Object
 };
 
-export type CoalesceParameters = {
-    source: string
-};
-
 export type LoadGeoJSON = (params: LoadGeoJSONParameters, callback: Callback<mixed>) => void;
 
 export interface GeoJSONIndex {
 }
 
 function loadGeoJSONTile(params: WorkerTileParameters, callback: LoadVectorDataCallback) {
-    const source = params.source,
-        canonical = params.tileID.canonical;
+    const canonical = params.tileID.canonical;
 
-    if (!this._sources[source] || !this._sources[source].geoJSONIndex) {
+    if (!this._geoJSONIndex) {
         return callback(null, null);  // we couldn't load the file
     }
 
-    const geoJSONTile = this._sources[source].geoJSONIndex.getTile(canonical.z, canonical.x, canonical.y);
+    const geoJSONTile = this._geoJSONIndex.getTile(canonical.z, canonical.x, canonical.y);
     if (!geoJSONTile) {
         return callback(null, null); // nothing in the given tile
     }
@@ -89,12 +84,10 @@ export type SourceState =
  */
 class GeoJSONWorkerSource extends VectorTileWorkerSource {
     loadGeoJSON: LoadGeoJSON;
-    _sources: { [string]: {
-        state?: SourceState,
-        pendingCallback?: Callback<boolean>,
-        pendingLoadDataParams?: LoadGeoJSONParameters,
-        geoJSONIndex?: GeoJSONIndex // object mapping source ids to geojson-vt-like tile indexes
-    }};
+    _state: SourceState;
+    _pendingCallback: Callback<boolean>;
+    _pendingLoadDataParams: LoadGeoJSONParameters;
+    _geoJSONIndex: GeoJSONIndex
 
     /**
      * @param [loadGeoJSON] Optional method for custom loading/parsing of
@@ -106,7 +99,6 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
         if (loadGeoJSON) {
             this.loadGeoJSON = loadGeoJSON;
         }
-        this._sources = {};
     }
 
     /**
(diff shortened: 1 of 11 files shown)


ChrisLoer:
Since we reverted the GeoJSON coalescing changes, I cherry picked them into this branch and re-based it on top of `master`. After talking with @anandthakker, I went for the narrowest fix I could figure out for the potential `removeSource` race condition -- I just make `GeoJSONSource` track if it's been removed, and if so it will skip sending further messages in response to a `loadData` callback.

`GeoJSONSource` appears to be the only source implementation that actually sends `removeSource`, which helps keep the scope of this change small. It does however imply that we're leaking `WorkerSource` objects on the background whenever non-GeoJSON sources are removed. This is probably (?) not too big a deal -- previously, we were leaking entries in the shared `loaded`/`loading` maps, whereas now we're leaking an object which is basically just those entries along with some references to shared state (`actor`, `styleLayerIndex`, etc.).

ChrisLoer:
@mollymerp, I made analogous changes to `raster_dem_tile_worker_source`, maybe you could take a glance at them? It seems pretty simple, but I'm almost totally unfamiliar with that source (TIL: DEM = Digital Elevation Model).

src/source/geojson_worker_source.js
```diff
@@ -77,8 +82,11 @@ function loadGeoJSONTile(params: WorkerTileParameters, callback: LoadVectorDataC
  * @private
  */
 class GeoJSONWorkerSource extends VectorTileWorkerSource {
-    _geoJSONIndexes: { [string]: GeoJSONIndex };
     loadGeoJSON: LoadGeoJSON;
+    _state: SourceState;
+    _pendingCallback: Callback<boolean>;
+    _pendingLoadDataParams: LoadGeoJSONParameters;
+    _geoJSONIndex: GeoJSONIndex // object mapping source ids to geojson-vt-like tile indexes
```
anandthakker: nit: the line comment is now obsolete

src/util/actor.js
```diff
@@ -86,10 +86,10 @@ class Actor {
             // data.type == 'loadTile', 'removeTile', etc.
             this.parent[data.type](data.sourceMapId, deserialize(data.data), done);
         } else if (typeof data.id !== 'undefined' && this.parent.getWorkerSource) {
-            // data.type == sourcetype.method
+            // data.type == sourcetype.sourcename.method
```
anandthakker: Nit: also update the doc comment above `send()` to reflect this change

ChrisLoer:
OK, updated comments and squashed to two commits: (1) re-introduces GeoJSON loadData coalescing, and (2) does the WorkerSource re-factoring.

My plan is still to hold off merging this until after we've released .44, just in case...

ChrisLoer:
I just rebased on top of master, which turned out to be a bit messy because these changes collided with the new resource timing collection from PR #5991.

@sbma44 can you point me to a good way to sanity-check that this PR won't break yours?

sbma44:
@ChrisLoer sure thing--hopefully the unit tests capture a bunch of this, but for integration testing/sanity checking here's what I used:

```
'use strict';

const TOKEN = 'pk.xyz123'

let gd = {
  type: "Feature",
  properties: {},
  geometry: {
    type: "Point",
    coordinates: [
      -88.6,
      36.6
    ]
  }
};

mapboxgl.accessToken = TOKEN;
    container: 'map',
    collectResourceTiming: true
});

map.on('load', () => {
	map.addSource('blah', {
	    type: 'geojson',
	    data: 'http://localhost:5000/test.geojson'
	});

	window.setTimeout(() => {
		map.getSource('blah').setData('http://localhost:5000/test2.geojson');

		window.setTimeout(() => {
			map.getSource('blah').setData(gd);
		}, 2000);
	}, 1000);
});

map.on('data', (mde) => {
	console.log(mde);
});
```

to adapt it you will need to:
- populate the access token
- make a local geojson available at the two embedded URLs (doesn't matter what)
- load the page and check console output for the presence of `resourceTiming` on:
  - the tile property of the source data event associated with vector tile loads
  - the event itself of geojson source metadata events
  - the event itself of geojson source content events (for updates)

ChrisLoer:
Thanks @sbma44! Looks like it's coming through fine. :+1:

test/unit/source/geojson_worker_source.test.js
```diff
@@ -197,3 +150,90 @@ test('resourceTiming', (t) => {
...
+            "type": "Point",
+            "coordinates": [0, 0]
+        }
+    };
+
+    const layerIndex = new StyleLayerIndex(layers);
+    function createWorker() {
+        const worker = new GeoJSONWorkerSource(null, layerIndex);
+
+        // Making the call to loadGeoJSON to asynchronous
```
mollymerp: nit - tiny typo
>Making the call to loadGeoJSON ~to~ asynchronous

ChrisLoer:
Benchmark results don't seem to show any big changes:

https://bl.ocks.org/anonymous/raw/9187a9a85d29a307fe03fe1f86f3d61c/

Issue #5948, opened 2018-01-04
expose Resource Timing API information for web worker XHRs


sbma44:
## Motivation

Characterizing performance as resources are fetched is useful for a variety of purposes. Although some of GL JS's resource-fetching can already be scrutinized via the [Resource Timing API](https://developer.mozilla.org/en-US/docs/Web/API/Resource_Timing_API/Using_the_Resource_Timing_API), requests made by web workers, including tile fetches, [are monitored](https://developer.mozilla.org/en-US/docs/Web/API/Resource_Timing_API/Using_the_Resource_Timing_API) but their measurements remain stranded in the worker's DedicatedWorkerGlobalScope.

## Design Alternatives

Measuring request time is possible using existing GL JS events, but not with the level of specificity available via the Resource Timing API (cf https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming).

## Design

An `options.collectTiming` (or similar) boolean is passed when the `Map` object is instantiated. If true, `MapDataEvent`s received by event listeners should include the `PerformanceEntry` object associated with each XHR, as appropriate (e.g. when `isSourceLoaded` is true). My read of the code indicates that this property will need to be passed down to web workers, and when `true`, workers will collect and pass back `PerformanceEntry`s to `WorkerTileCallback` (as an additional property alongside `rawTileData`, I think?).

I have been examining this problem from a vector-tile-fetching perspective; apologies if there are other resource types that should be included here that I haven't considered. At a glance it appears that repeating this pattern across worker-fetched resource types shouldn't be too bad.

### Mock-Up

Developers will be able to enable this flag and collect timing information from event callbacks.
End users should not notice any difference.

### Concepts

The Resource Timing API is widely used and supported browser API, so many resources already exist to familiarize developers with the relevant concepts. Updating documentation to make users aware of this option and to explain why they can't access tile timings from the application's scope will be useful.

### Implementation

See above for Javascript implementation proposal.

I suspect that C++ idioms for measuring network performance are distinct and should be respected. Sharing property names across implementations might be nice. I don't see an immediate need for implementing in other Mapbox GL libraries, though. I don't see any broad impacts on the Mapbox GL ecosystem.

sbma44:
Noting briefly that thanks to spirit guidance from @anandthakker I've got a working prototype -- adding support for GeoJSON sources and tests now, should have something for you all in a day or two.

mollymerp:
closed in #5991 – thanks @sbma44

PR #5991, opened 2018-01-12
Expose Resource Timing API

sbma44:
Implements https://github.com/mapbox/mapbox-gl-js/issues/5948. Much thanks to @anandthakker for his patient guidance!

Since the only source types that make network requests in workers are VT and GeoJSON, those are the only two source types with changes. Both now return `resourceTiming` properties on `data` events if the `Map` is instantiated with `{ collectResourceTiming: true }`.

Because of differences between these sources, the location of the timing data differs between them. For VT workers, network requests are associated with individual tiles; for GeoJSON sources, they are associated with source addition and data update events. Consequently the location of the `resourceTiming` property varies depending on the source type -- it's on tiles for VT and is a top-level property for GeoJSON events.

## Launch Checklist

 - [x] document any changes to public APIs (JSDoc only right now)
 - [x] manually test the debug page (done w/ custom local page with `collectResourceTiming` and a geoJSON source).
 - [x] post benchmark scores

~~Benchmarks are running now but a looming ✈️  means I'll probably need to post the results next week.~~ Because this feature is intended to be used primarily in diagnostic or development contexts and is disabled by default, I don't expect to see differences in the bench suite. Even when enabled, this code simply calls a browser API and stringifies a small amount of scalar data (and does so in a worker). If anyone feels that writing a bench test for this is appropriate, though, please let me know and I'll give it a shot.


sbma44:
2018-01-12-15-33-localhost_9966

src/source/geojson_source.js
```diff
@@ -134,9 +140,18 @@ class GeoJSONSource extends Evented implements Source {
                 this.fire('error', {error: err});
                 return;
             }
+
+            const data = {};
+            data.dataType = 'source';
+            data.sourceDataType = 'metadata';
```
anandthakker: Nit: 
```js
const data = {
    dataType: 'source',
    sourceDataType: 'metadata'
}
```

(also below)

src/source/geojson_source.js
```diff
@@ -76,8 +77,10 @@ class GeoJSONSource extends Evented implements Source {
     map: Map;
     workerID: number;
     _loaded: boolean;
+    _collectResourceTiming: boolean;
+    _resourceTimings: {[string]: Array<PerformanceResourceTiming>};
```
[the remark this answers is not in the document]
sbma44: ah ok. I had gotten the impression from the `coalesce` operation (and a superficial read of the https://github.com/mapbox/mapbox-gl-js/pull/5988 issue title) that a single worker could wind up handling multiple sources. If that's not the case I agree that it should be simplified 👍

src/source/geojson_source.js
sbma44: That's actually what I wrote originally, but it didn't survive `test-flow`. I think this has to stay this way because of the subsequent assignment of `data.resourceTiming`.  Flow complains about

```js
const data = { x: 'y' };
data.i = 'j';
```

but not

```js
const data = {};
data.x = 'y';
data.i = 'j';
```

My knowledge of Flow is pretty rudimentary so if there's a better way to accommodate this case please lmk.

src/source/geojson_source.js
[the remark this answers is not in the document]
anandthakker: That's true -- a single worker can handle multiple sources, but this here is the main-thread side (not the worker side) of the source

src/source/geojson_source.js
sbma44: That's actually what I wrote originally, but it didn't survive `test-flow`. I think this has to stay this way because of the subsequent potential assignment of `data.resourceTiming`.  Flow complains about

```js
const data = { x: 'y' };
data.i = 'j';
```

but not

```js
const data = {};
data.x = 'y';
data.i = 'j';
```

My knowledge of Flow is pretty rudimentary so if there's a better way to accommodate this case please lmk.

src/source/geojson_source.js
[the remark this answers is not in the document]
sbma44: ah, right you are. sorry to have missed that!

src/source/geojson_source.js
anandthakker: Ah, gotcha.  Does the following avoid the flow error?

```js
const data: MapDataType = {
    dataType: 'source',
    sourceDataType: 'metadata'
}
```

(you might have to import `MapDataType` from `ui/events.js`)

sbma44:
Noting here: this behavior fails to return resource timing information in Firefox Quantum 58 due to https://bugzilla.mozilla.org/show_bug.cgi?id=1425458 . It looks like this will be resolved when Firefox Quantum 60 ships.

Issue #5993, opened 2018-01-12
Multiple symbols in the same tile can share the same crossTileID


ChrisLoer:
This is the GL JS version of https://github.com/mapbox/mapbox-gl-native/issues/10844#issuecomment-357128942. You can reproduce by modifying the "chinese.html" map to give every one of the points the same text label. Then by starting at z0 and zooming in, you can get to the state where all of the points are marked as duplicate against the first point.

The fix we plan is:
 - Maintain a set of `crossTileID`s used by each zoom level
 - On detecting duplicates, check to see if the about-to-be-copied `crossTileID` is already in use for your zoom layer. If so, move on to the next potential duplicate. If no more potential duplicates are found, assign a new `crossTileID`
 - On removing a bucket from the index, remove the set of `crossTileID`s used by that bucket.

Example scenario imagining a z0 tile with symbols A and B, and a z1 tile with symbols A', B', and C' (all at the same location and with the same "key"):

> - A' matches with A
> - A' sees that the id from A is not used and copies it
> - B' matches with A
> - B' sees that the id from A is used and continues the search
> - B' matches with B
> - B' sees that the id from B is not used and copies it
> - C' matches with A
> - C' sees that the id from A is used and continues the search
> - C' matches with B
> - C' sees that the id from B is used and continues the search
> - C' reaches the end of the search without copying and gets assigned a new id
> both A and B are duplicate and don't get rendered
> both A' and B' are instantly visible (since they used the opacities from A and B)
> C' fades in

/cc @ansis @friedbunny

ChrisLoer:
Fixed with #5994.

PR #5994, opened 2018-01-12
Prevent multiple symbols at the same zoom from sharing a crossTileID

ChrisLoer:
Fixes issue #5993, which would allow multiple symbols in a tile to share the same crossTileID as one of their duplicate parent-tile symbols. Once the symbols shared a crossTileID, they would permanently remain "duplicates" of each other, even after increasing zoom level allowed the CrossTileSymbolIndex to distinguish between them.

This change introduces some extra work on removing a bucket (since we have to iterate through all symbol instances removing their crossTileIDs from the index), but it doesn't really show up in benchmarks. However, running the Chrome javascript profiler while doing pan and zoom operations, I'm seeing `removeStaleBuckets` taking .5-1.5% of CPU time, which seems tolerable but still pretty concerning. Since there's no time urgency on the pruning, maybe we can set up a separate pruning function that we just run once every N minutes/iterations/frames?

On top of the unit test, I tested by manually exercising the "chinese.html" debug page with the keys all set to the same value. I watched to see how crossTileIDs got assigned and watched that `usedCrossTileIDs` wasn't growing without bound.

@friedbunny once we get this in we'll port it over to native to fix the original annotation view bug.

/cc @ansis

### DIFF
--- a/src/symbol/cross_tile_symbol_index.js
+++ b/src/symbol/cross_tile_symbol_index.js
@@ -71,7 +71,7 @@ class TileLayerIndex {
         };
     }
 
-    findMatches(symbolInstances: Array<SymbolInstance>, newTileID: OverscaledTileID) {
+    findMatches(symbolInstances: Array<SymbolInstance>, newTileID: OverscaledTileID, zoomCrossTileIDs: {[crossTileID: number]: boolean}) {
         const tolerance = this.tileID.canonical.z < newTileID.canonical.z ? 1 : Math.pow(2, this.tileID.canonical.z - newTileID.canonical.z);
 
         for (const symbolInstance of symbolInstances) {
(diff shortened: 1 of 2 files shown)


ChrisLoer:
I put together a commit that does periodic pruning instead of doing the pruning during `removeStaleBuckets`, but it didn't actually make `removeStaleBuckets` run that much faster -- I guess what that probably means is the "there's nothing to remove" part of `removeStaleBuckets` is more expensive than the case where we actually have to remove something. So I think the deferred pruning commit probably isn't worth the (mild) extra complexity

src/symbol/cross_tile_symbol_index.js
```diff
@@ -168,6 +181,11 @@ class CrossTileSymbolLayerIndex {
             const zoomIndexes = this.indexes[z];
             for (const tileKey in zoomIndexes) {
                 if (!currentIDs[zoomIndexes[tileKey].bucketInstanceId]) {
+                    for (const key of Object.keys(zoomIndexes[tileKey].indexedSymbolInstances)) {
```
ansis: is there an advantage to using `for (const key of Object.keys(z))` instead of `for (const key in z))` in this case?

ChrisLoer: I don't think so, it's just left over from an earlier implementation I tried iterating over `Object.entries`. I'll update it to the simpler form.

PR #5999, opened 2018-01-16
queryRenderedFeatures crash the map

Scarysize:
Not a feature, but an example which completely crashes the map. I wasn't able to reproduce this in a jsfiddle, that's why I'm providing the sample as a debug page in a PR.

**mapbox-gl-js version**: v0.43.0/current master

### Steps to Trigger Behavior

0. (start the debug server, `npm run start-debug`)
1. Navigate to http://localhost:9966/debug/1178.html
2. Zoom close to the netherlands
3. Move your mouse over the rendered geojson polygons
4. The map & chrome tab get stuck/crash

### Expected Behavior

The map does not crash.

### Actual Behavior

The map crashes.

Hopefully this is reproducible for someone else. I don't think that this is an issue of calling `queryRenderedFeatures` in quick succession, this also happens with a similar setup listening for `click` events.

### DIFF
--- a/debug/1178.html
+++ b/debug/1178.html
@@ -33,7 +33,7 @@
     });
 
     map.on('load', async () => {
-      const geojson = await fetch('https://storage.googleapis.com/storage.ubidev.net/meteogroup/dWU5OHd6Zmg4NzI5cmZjMDh6/geojson/geojsonsample-1178.json').then(res => res.json());
+      const geojson = await fetch('https://storage.googleapis.com/storage.ubidev.net/mapbox-gl-js/329034912902/geojsonsample-1178.json').then(res => res.json());
       map.addSource('src', {
         type: 'geojson',
         data: geojson
--- a/debug/1178.html
+++ b/debug/1178.html
@@ -1,0 +1,59 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <title>Mapbox GL JS debug page</title>
+  <meta charset='utf-8'>
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+  <link rel='stylesheet' href='/dist/mapbox-gl.css' />
+  <style>
+    body {
+      margin: 0;
+      padding: 0;
+    }
+
+    html,
+    body,
+    #map {
+      height: 100%;
+    }
+  </style>
+</head>
+
+<body>
+  <div id='map'></div>
+
+  <script src='/dist/mapbox-gl-dev.js'></script>
+  <script src='/debug/access_token_generated.js'></script>
+  <script>
+
+    const map = window.map = new mapboxgl.Map({
+      container: 'map',
+      style: 'mapbox://styles/mapbox/streets-v10'
+    });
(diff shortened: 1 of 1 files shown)


ansis:
Which version of Chrome are you encountering this with? I was not able to reproduce with `63.0.3239.132`.

How long do you have to move your mouse over the polygons? Does it reliably crash or just every once in a while?

Scarysize:
I just reproduced this on my private MBP 13":
- High Sierra 10.13.2
- Intel Iris Graphics 6100 1536 MB

**Browsers I was able to reproduce this:**
- Safari 11.0.2
- Chrome 63.0.3239.132
- Firefox 57.0.4

Also our customer has this issue on a Windows machine with Chrome, unsure about the version, but probably the most recent one.

> How long do you have to move your mouse over the polygons? Does it reliably crash or just every once in a while?

It crashes pretty fast. If you zoom in close enough so that the polygons roughly cover your viewport, it crashes before you hovered over them even once.

kkaefer:
I just tried reproducing this as well with a 15" MBP, High Sierra 10.13.2, AMD Radeon Pro 450 and Intel HD Graphics 530, with Safari 11.0.2, Firefox 58.0, and Chrome 63.0.3239.132, but couldn't find any crash or hang with this example. I also confirmed that `queryRenderedFeatures` returns the correct results.

ansis:
>  If it's of any help, here is what the call stack looks like, when the map is stuck:

@Scarysize does it always get stuck with the same call stack?

ChrisLoer:
Just for the record: I just tried reproducing this with a 15" MBP, High Sierra 10.13.2, AMD Radeon R9 M370X 2 GB and Intel Iris Pro 1536 MB, with Safari 11.0.2, Firefox 58.0, and Chrome 63.0.3239.132, and just like @kkaefer didn't see any problems.

Scarysize:
Very weird. I updated to the recent `master` and could reproduce this on multiple devices:
- iPhone 7, iOS 11 (Chrome)
- Win 10, recent Chrome (GTX 980Ti)

I will provide a screencast as soon as possible.

Scarysize:
https://i.imgur.com/ut75goV.mp4

Here is a video showing the issue. As you can see, after hovering of the polygons, I'm trying to pan & zoom the map. But the page isn't responding anymore. Shortly after I get the "Page Unresponsive" dialogue from Chrome.

I'm trying to get a colleague to reproduce this next week.

Scarysize:
> @Scarysize does it always get stuck with the same call stack?

@ansis Yes it does.

It crashes reliably on all the systems I checked. My colleague can't reproduce the issue though. I'm pretty clueless, what's responsible for reproducing this...

tolgaozkan:
For me reproducible as well:

Win 10 Enterprise Version: 1703 OS Build: 15063.850
Chrome Version 63.0.3239.132 (Official Build) (64-bit)
node 8.9.4
npm 5.6.0

CPU: Intel Core i5-5200U
GPU: Intel HD 5500

PR #6005, opened 2018-01-17
Sync continuously-updating interaction handlers with render loop

### LINKED ISSUE #5390: SVG overlay using on("move") and project() out of sync in IE/Edge/Safari
I'm trying to use an SVG overlay like this https://bl.ocks.org/shimizu/5f4cee0fddc7a64b55a9 or this http://bl.ocks.org/enjalot/0d87f32a1ccb9a720d29ba74142ba365.

I use a map.on("move") event to call a function which updates the location of the SVG elements, together with map.project() to get the on screen co-ordinates (similar to the first example - except without hiding the SVG graphics when moving).

In Chrome and Firefox this works well with the SVG graphics appearing seamlessly with mapbox, but in IE11, Edge and to some extent Safari there is a significant lag between the mouse move and the SVG features being updated.

Possibly this is related to this issue: https://github.com/mapbox/mapbox-gl-js/pull/4681 because I see similar behaviour with markers in those browsers. Not sure if that PR is being progressed?

**Example in Edge:**

mapbox-svg

anandthakker:
Closes #5390 

Quoting from that issue:

> the DragPan handler updates the camera directly and fires a move event. Updates to the DOM that happen in move listeners happen synchronously, but the map renderer doesn't update until the next animation frame.

This PR aims to solve this issue by having the `DragPan` handler (and the other continuously-updating handlers) cede control to the camera, which in turn is driven by the render loop.

I think it's also a step towards further refactoring of camera of `Camera` that eliminates some of the complicated internal state, simplifies/clarifies `move`-related event firing, etc.

## Launch Checklist

 - [x] <del>write tests for all new functionality</del> (blocked by https://github.com/mapbox/mapbox-gl-js/issues/1550)
 - [ ] post benchmark scores

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -309,9 +309,6 @@ class Map extends Camera {
 
         this.on('move', this._update.bind(this, false));
         this.on('zoom', this._update.bind(this, true));
-        this.on('move', () => {
-            this._rerender();
-        });
 
         if (typeof window !== 'undefined') {
             window.addEventListener('online', this._onWindowOnline, false);
@@ -1528,7 +1525,7 @@ class Map extends Camera {
         // Even though `_styleDirty` and `_sourcesDirty` are reset in this
         // method, synchronous events fired during Style#update or
         // Style#_updateSources could have caused them to be set again.
-        if (this._sourcesDirty || this._repaint || this._styleDirty || this._placementDirty || this.isEasing()) {
+        if (this._sourcesDirty || this._repaint || this._styleDirty || this._placementDirty) {
             this._rerender();
         }
 
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1458,9 +1458,7 @@ class Map extends Camera {
      * @private
      */
     _render() {
-        if (this.isEasing()) {
-            this._updateEase();
-        }
+        this._updateCamera();
 
         let crossFading = false;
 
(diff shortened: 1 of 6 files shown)


anandthakker:
@mourner and others on @mapbox/gl-core, could you give this a quick look to validate the approach, before I propagate it to the other handlers (`DragRotate`, etc.)?

anandthakker:
@mourner removed closures and attempted to clarify the code a bit more.  I still lean towards calling it `_startAnimation`, even if it's slightly stretching the idea of 'animation'... but I don't feel strongly about it / happy to change it.

This is ready for a real review now @mollymerp @mourner

src/ui/map.js
```diff
@@ -1528,7 +1525,7 @@ class Map extends Camera {
         // Even though `_styleDirty` and `_sourcesDirty` are reset in this
         // method, synchronous events fired during Style#update or
         // Style#_updateSources could have caused them to be set again.
-        if (this._sourcesDirty || this._repaint || this._styleDirty || this._placementDirty || this.isEasing()) {
+        if (this._sourcesDirty || this._repaint || this._styleDirty || this._placementDirty) {
```
jfirebaugh: If we're planning to fully convert the handlers to use render-loop-driven camera, then we should:

* Have handlers call `map._update` explicitly when they need to trigger a render, as in #6068
* Remove the `move`/`zoom` => `this._update` bindings
* Restore this `this.isEasing()` trigger

PR #6012, opened 2018-01-18
default marker offset

andrewharvey:
## Launch Checklist

#5661 implemented a default Marker element, this PR applies an offset for that default Marker, so the tip of the marker image is located at the marker lng/lat.

I don't think it needs any special documentation as I'd consider it to be expected behaviour so I consider this to be a bug fix rather than a breaking change.

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -31,8 +31,6 @@ class Marker {
     _pos: ?Point;
 
     constructor(element: ?HTMLElement, options?: {offset: PointLike}) {
-        this._offset = Point.convert(options && options.offset || [0, 0]);
-
         bindAll(['_update', '_onMapClick'], this);
 
         if (!element) {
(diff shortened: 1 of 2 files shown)


src/ui/marker.js
```diff
@@ -127,8 +125,22 @@ class Marker {
             svg.appendChild(page1);
 
             element.appendChild(svg);
+
+            // if no element and no offset option given apply an offset for the default marker
+            const defaultMarkerOffset = [0, -14];
```
mollymerp: did you determine the -14 through trial and error or some other way? seeing that the SVG is 41px tall makes this number non-obvious to me, although it looks great when I tested the branch. it would be nice to avoid hardcoding this constant and get the offset from some of the SVG properties.

andrewharvey: I actually opened the SVG in Inkscape and measured the distance from the image center to where the tip is, but you could calculate it based on the center of the shadow ellipse, from the SVG, but it would add a fair bit more code and not sure how useful that would be. I've added some comments on where the -14 magic number came from.

What do you think, should be leave it as is with the comment or go ahead and calculate it on the fly from the SVG?

mollymerp: gotcha – I think we can leave it hardcoded for now with the comment.

andrewharvey:
> Thanks @andrewharvey – this works for me. One thing I thought of was maybe adding a comment to the element param of Marker constructor to indicate that, if an element is provided, the marker will be placed such that the center of the element is on the LngLat they set so users will know they need to offset if they want different behavior.

Thanks for revewieng @mollymerp .

@jay-manday is working on Marker anchor at #6031 which also brings in Flow and better JSDoc for Marker. I think once the anchor option is introduced it makes it a bit clearer that by default they are anchored to the middle, so I'm tempted to leave the comment out and have it addressed in #6031, which also helps reduce the conflicts which have arisen between both PRs.

Does that work?

anandthakker:
Thanks @andrewharvey !

PR #6026, opened 2018-01-19
Align projection matrix to pixel grids to draw crisp raster tiles

kkaefer:
This changes our projection matrix to be aligned to the pixel grid to fix blurry raster tiles at integer zoom levels (https://github.com/mapbox/mapbox-gl-js/issues/4552).

While this change works great

before:
expected

after:
actual

and produces the same level of crispness as a Leaflet map of the same raster tiles, it also changes the behavior of some of our APIs: it is no longer possible to roundtrip a point through our transform object and get the same result back, since all transformations are now aligned to this pixel grid. It also slightly shifts rendering of all other items (shapes, lines, symbols) by up to half a pixel, which results in hundreds of rendering "errors".

One way to solve this would be to use the pixel-aligned projection matrix only for raster (and terrain?) tiles. However, this means that our raster tiles might be shifted by up to half a pixel compared to the rest of the data we are drawing on the screen. In reality, I doubt that this will be noticeable, but something we should consider before merging this.

A notable discovery is that pixel alignment of our map depends on whether the viewport has even or odd dimensions.


kkaefer:
Update: changed this PR so that it only pixel-grid-aligns raster tiles. On my machine, this produces perfectly crisp output on regular and retina screens in all browsers and for all viewport dimensions.

kkaefer:
Testing-wise, it'd be cool to have a few people with different GPUs test this change. For this purpose, I've checked in a http://localhost:9966/debug/raster-streets.html#3/0/0 file. Use (shift-)double-click to zoom in(out) to remain on integer zoom levels (or manually manipulate the hash).

kkaefer:
Also added the ability to draw crisp rasters at 90 and 270 degrees rotation + tests.

src/geo/transform.js
```diff
@@ -398,10 +401,11 @@ class Transform {
      * Calculate the posMatrix that, given a tile coordinate, would be used to display the tile on a map.
      * @param {UnwrappedTileID} unwrappedTileID;
      */
-    calculatePosMatrix(unwrappedTileID: UnwrappedTileID): Float32Array {
+    calculatePosMatrix(unwrappedTileID: UnwrappedTileID, aligned: boolean = false): Float32Array {
         const posMatrixKey = unwrappedTileID.key;
-        if (this._posMatrixCache[posMatrixKey]) {
-            return this._posMatrixCache[posMatrixKey];
+        const cache = aligned ? this._alignedPosMatrixCache : this._posMatrixCache;
+        if (!aligned && cache[posMatrixKey]) {
```
lbud: 🤔 would we not want to use the `alignedPosMatrixCache` result if it's available?

lbud: 🤔 would we not want to use the `alignedPosMatrixCache` result if it's available?

kkaefer: Ah, good catch. That’s an artifact from a different approach I tried.

ryanbaumann:
cc @pdgoodman @aag6z

ibesora:
Amazing work @kkaefer 
Time to ditch Leaflet!

jaapster:
Thanks @kkaefer!

Issue #5640, opened 2017-11-09
Add 'anchor' option for Markers


anandthakker:
Following up from https://github.com/mapbox/mapbox-gl-js/pull/4751, it would be nice to add an 'anchor' option for `Marker`, similar to the one in `Popup`: https://github.com/mapbox/mapbox-gl-js/blob/f11b7640a2b6d589ef8d2d7297c4d9167ae94f47/src/ui/popup.js#L37-L42

jay-manday:
Happy to work on this if no one has it in their queue! I can see this being pretty helpful for myself and other users as well.

anandthakker:
@jay-manday I don't think anyone's claimed it yet, so please feel free!

> Do we want to export a MarkerOptions type as done in popup as well?

Couldn't hurt!

jay-manday:
Hi @anandthakker I've been working on this today and I think I made a bit of progress though a little stumped as to why my tests are failing. 

I'm getting an `Cannot read property 'anchor' of undefined` in testing. 

When you get a chance would you mind taking a look? I've tried a few things like setting default anchor and also did some snooping in the test file for marker.js but a little unsure of how best to proceed.

[Here's where i'm at :/](https://github.com/mapbox/mapbox-gl-js/commit/c9e6b40bb43542c3b550d8b26cb58b16ce5cae10)

Thanks for taking the time to review!

jay-manday:
Hi @anandthakker 

I thiiiink i have things working as needed though would love a review of my work before making a PR to the master branch on your end. I did some digging on the origins of the issue was a pretty cool learning experience! 🌀💡🎉✨ 

Heres the work done on my brach:
https://github.com/jay-manday/mapbox-gl-js/pull/2/commits/47d1243057b5b4e6324ee0508051ad4f7499eca0

I also went ahead and manually wrote the test for the marker file too. n_n

andrewharvey:
Hi @jay-manday nice to see you implementing this! Feel free to go ahead and open that PR as that's where the review typically takes place, even if you feel it's not 100% ready, you can just mention it's a WIP.

The options for Popup's anchor are   `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`, `'top-right'`, `'bottom-left'`, and `'bottom-right'. I think that's what we should implement for the Marker anchor.

jay-manday:
[...]

PR #6031, opened 2018-01-21
WIP: added support for anchor option on Marker

jay-manday:
## Launch Checklist
I've been working on adding support for anchor options to the marker element as requested in #5640 and #4751. It's a work in progress and borrows a bit from the `popup` where I thought appropriate. I did some testing and all the cosmetics are clean though getting a little stumped when it hits the tests for marker.js. 

would love some feedback on my progress so far though! n_n

 - [ ] post benchmark scores


src/ui/marker.js
```diff
@@ -11,26 +11,54 @@ import type Popup from './popup';
...
+ * @param {string} [options.anchor] - A string indicating the markers's location relative to
+ *   the coordinate set via {@link Marker#setLngLat}.
+ *   Options are `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`,
+ *   `'top-right'`, `'bottom-left'`, and `'bottom-right'`. If unset the anchor will be
+ *   dynamically set to ensure the marker falls within the map container with a preference
+ *   for `'bottom'`.
+ * @param {number|PointLike|Object} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @example
+ * var markerHeight = 50, markerRadius = 10, linearOffset = 25;
+ * var markerOffsets = {
```
andrewharvey: Popup has this since the anchor of the popup changes to ensure it's visible within the map (ie. moves below the point when there is no space above it), the Popup offset needs to be different for each anchor in some use cases.

Is there a use case for having a variable anchor for a single marker and hence needing different offsets set for each anchor?

jay-manday: mmm I was thinking about exposing as a variable so users could define their own offset per marker/anchor. This could be for styling considerations on static maps or maps with markers in close proximity to eachother. 

One possible route to take would have popup inherit the anchor from marker as a default for when users don't define?

andrewharvey: Good point, I guess you could have different marker images anchored at different locations.

andrewharvey: I see, for example the default Marker in Mapbox with an anchor of bottom still needs an offset due to the shadow, so if you have a marker that goes upsidedown when it can't fit at the top of the map and it has a shadow it needs an offset per anchor value. Got it.

jay-manday: yup! I think its good to have built in flexibility but include batteries for desired behaviors if not specified. These were great questions though, thank you for taking the time to review my work so far ✨🎷🐛

anandthakker: As @andrewharvey noted above, the reason we need this in `Popup` is that it dynamically decides the anchor based on the popup's position relative to the edges of the viewport.  I don't think `Marker` should include such logic -- it's lower-level than `Popup`, and the use case for such automatic anchoring magic seems less clear.

src/ui/marker.js
```diff
@@ -11,26 +11,54 @@ import type Popup from './popup';
 import type {LngLatLike} from "../geo/lng_lat";
 import type {MapMouseEvent} from './events';
 
+export type Anchor = 'top' | 'bottom' | 'left' | 'right' | 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
```
anandthakker: Also needs `center`

src/ui/marker.js
```diff
@@ -268,16 +296,82 @@ class Marker {
...
+            } else if (pos.x > this._map.transform.width - width / 2) {
+                anchor.push('right');
+            }
+
+            if (anchor.length === 0) {
+                anchor = 'bottom';
+            } else {
+                anchor = anchor.join('-');
+            }
+        }
```
anandthakker: Similar to the discussion above re: offset, I'd be in favor of keeping the default behavior simpler, i.e., simply defaulting to `center` (since that's the current behavior)

src/ui/marker.js
jay-manday: sorry to keep things held up!! i've been working my way through testing and will make another commit soon but would definitely welcome another pair of eyes on this n_n

src/ui/marker.js
```diff
@@ -11,26 +11,54 @@ import type Popup from './popup';
...
- * @param options.offset The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
+ * @param {Object} [element] DOM element to use as a marker. If left unspecified a default SVG will be created as the DOM element to use.
+ * @param {Object} [options]
+ * @param {string} [options.anchor] - A string indicating the markers's location relative to
+ *   the coordinate set via {@link Marker#setLngLat}.
+ *   Options are `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`,
+ *   `'top-right'`, `'bottom-left'`, and `'bottom-right'`. If unset the anchor will be
+ *   dynamically set to ensure the marker falls within the map container with a preference
+ *   for `'bottom'`.
+ * @param {number|PointLike|Object} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
```
anandthakker: > relative to the element's center

This will now be "relative to the anchor"

src/ui/marker.js
andrewharvey: `middle` should be an option, and it should be the default option (it what happens currently in master)

src/ui/marker.js
andrewharvey: > "relative to the element's center"

I think with the anchor option it's clearer to say relative to the element's anchor.

src/ui/marker.js
anandthakker: Ah, yeah I think @andrewharvey 's right, it should be `middle`, not `center`

src/ui/marker.js
andrewharvey: Whops, sorry didn't see @anandthakker's comment

jay-manday: i'm just seeing those as well X_x thank you both for taking the time to review and for your patience

jay-manday:
I've made the requested changes to marker, though still getting stumped with the proper testing procedure to mock the new anchor option. This has been a good learning experience for me though, I don't have too much experience writing tests though always strive for succinct code. Thank you so so much for your time, patience and mentorship @anandthakker @andrewharvey I super appreciate it.

src/ui/marker.js
```diff
@@ -11,27 +11,43 @@ import type Popup from './popup';
...
 /**
  * Creates a marker component
- * @param element DOM element to use as a marker. If left unspecified a default SVG will be created as the DOM element to use.
- * @param options
- * @param options.offset The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
+ * @param {Object} [element] DOM element to use as a marker. If left unspecified a default SVG will be created as the DOM element to use.
+ * @param {Object} [options]
+ * @param {string} [options.anchor] - A string indicating the markers's location relative to
+ *   the coordinate set via {@link Marker#setLngLat}.
+ *   Options is  `'middle'` by default
```
andrewharvey: See https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/map.js#L171 for an example of how to set the default value directly in the JSDoc. The documentation generator automatically picks up this default value as you can see https://www.mapbox.com/mapbox-gl-js/api/#map

andrewharvey: Also like https://github.com/mapbox/mapbox-gl-js/blob/d11cf304f8cb4554fdf46a95bd20eeb3df090c59/src/ui/map.js#L171 I'd include a list of all the valid options for 'anchor'

jay-manday: omg, if this works i'm probably going to cry lol 😂

src/ui/marker.js
```diff
@@ -11,27 +11,43 @@ import type Popup from './popup';
 import type {LngLatLike} from "../geo/lng_lat";
 import type {MapMouseEvent} from './events';
 
+export type Anchor = 'middle';
```
andrewharvey: You had it right the first time enumerating the valid options like https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/popup.js#L19

@anandthakker any thoughts on if Marker should import the Anchor type from Popup? That would mean adding `middle` as a popup anchor option...

andrewharvey:
@jay-manday Keep it up! I think it might be helpful to re-look at how the API should work.

One way to approach it is to write all your unit tests first, as that defines how you expect the API to work, then you can start chipping away at the code to make them pass.

It looks like you're getting Anchor (a string like 'middle', 'bottom', 'top', etc) and Offset (an x,y pixel offset to move the Marker image relative to the anchor) mixed up in the code, and if we get the test cases right first in means we have defined how the API should work.

jay-manday:
ty soo much @andrewharvey! 

I made a ton of progress last night, only a handful of tests failing at this point. 
When i've got all the tests clearing I'll push my changes 🌮 🎉

jay-manday:
Still working on things and fleshed out the testing for marker. It seems the only major hurdle remaining is testing through the enumerated anchors and offsets, i'm going to keep working on this though of would welcome reviews of my progress. 🐌 

Thank you!

src/ui/marker.js
```diff
@@ -11,26 +11,53 @@ import type Popup from './popup';
 import type {LngLatLike} from "../geo/lng_lat";
 import type {MapMouseEvent} from './events';
 
+export type Anchor = 'middle' | 'top' | 'bottom' | 'left' | 'right' | 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
```
andrewharvey: :+1: this looks like all the options that we should support

src/ui/marker.js
```diff
@@ -11,26 +11,53 @@ import type Popup from './popup';
...
+}
+
 /**
  * Creates a marker component
- * @param element DOM element to use as a marker. If left unspecified a default SVG will be created as the DOM element to use.
- * @param options
- * @param options.offset The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
+ * @param {Object} [element] DOM element to use as a marker. If left unspecified a default SVG will be created as the DOM element to use.
+ * @param {Object} [options]
+ * @param {Array<string> | string} [options.anchor = Anchor] - A string indicating the markers's location relative to
```
andrewharvey: See "An optional parameter and default value" at http://usejsdoc.org/tags-param.html the default value should be `middle`.

See https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/popup.js#L37, the options.anchor type is only `string`. Each marker can only have one anchor so the type should just be string.

src/ui/marker.js
```diff
@@ -127,11 +154,31 @@ class Marker {
             svg.appendChild(page1);
 
             element.appendChild(svg);
+
+            // if no element and no offset option given apply an offset for the default marker
```
andrewharvey: best to rebase from master so these changes are brought in that way. that should also resolve the merge conflicts

src/ui/marker.js
```diff
@@ -268,16 +315,84 @@ class Marker {
...
+            }
+
+            if (anchor.length === 0) {
+                anchor = 'middle';
+            } else {
+                anchor = anchor.join('-');
+            }
+        }
+
+        const offsetedPos = pos.add(offset[anchor]);
```
jay-manday: for some reason, this part right here seems to be breaking all my tests, usually in the form of add `undefined` or expecting a `truthy value`

src/ui/marker.js
jay-manday: sure thing, and apologies for consuming so much time with this, i feel like i've learned a lot going through this process if it's any consolation 🌀  👍

jay-manday: sure thing, and apologies for taking so much time with this, i think

jay-manday:
so I've written more tests for marker's anchor and offset options and things are passing! I think I can safely say I now know how _*not*_ to do unit testing 😅 

Thank you so much for your patience and words of encouragement @andrewharvey @anandthakker, I really appreciate it, please let me know if anything i've done sticks out. 🎉

src/ui/marker.js
```diff
@@ -1,36 +1,71 @@
...
  * Creates a marker component
- * @param element DOM element to use as a marker. If left unspecified a default SVG will be created as the DOM element to use.
- * @param options
- * @param options.offset The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
+ * @param {Object} [element] DOM element to use as a marker. If left unspecified a default SVG will be created as the DOM element to use.
+ * @param {Object} [options]
+ * @param {string} [options.anchor] - A string indicating the markers's location relative to
+ *   the coordinate set via {@link Marker#setLngLat}.
+ *   Options are `'middle'`, `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`, `'top-right'`, `'bottom-left'`, and `'bottom-right'`.
+ *   If unset the anchor will be dynamically set to ensure the marker falls within the map container with a preference for `'middle'` by default
```
andrewharvey: in line with https://github.com/mapbox/mapbox-gl-js/pull/6031#discussion_r163733579 I don't think this should be the default behaviour. the anchor should default to middle and not do any dynamic choosing of an anchor based on where it will fit best within the map container.

src/ui/marker.js
```diff
@@ -1,36 +1,71 @@
...
+ * @param {number|PointLike|Object} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's anchor.
  * @example
+ * var markerHeight = 50, markerRadius = 10, linearOffset = 25;
```
andrewharvey: Even if it ends up supporting the same dynamic anchor logic as Popup, I don't think such a complicated use case should be the default example.

src/ui/marker.js
```diff
@@ -280,30 +318,98 @@ class Marker {
...
+            classList.remove(`mapboxgl-marker-anchor-${key}`);
+        }
+        classList.add(`mapboxgl-marker-anchor-${anchor}`);
+
+        DOM.setTransform(this._element, `${anchorTranslate[anchor]} translate(${offsetedPos.x}px,${offsetedPos.y}px)`);
+    }
+
+    /**
+     * Get the marker's anchor.
+     * @returns {Point}
```
andrewharvey: At the top you have set the Anchor type as `export type Anchor = 'middle' | 'top' | 'bottom' | 'left' | 'right' | 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';` yet your `getAnchor` returns `Point`?

test/unit/ui/marker.test.js
```diff
@@ -17,62 +17,201 @@ function createMap() {
 
 test('Marker', (t) => {
     t.test('constructor', (t) => {
-        const el = window.document.createElement('div');
+        const el = window.document.createElement(`mapboxgl-marker`);
```
andrewharvey: this will try to make an Element like `<mapboxgl-marker>` let's stick with a `<div>` element.

test/unit/ui/marker.test.js
```diff
@@ -17,62 +17,201 @@ function createMap() {
...
         const marker = new Marker(el);
         t.ok(marker.getElement(), 'marker element is created');
         t.end();
     });
 
     t.test('default marker', (t) => {
```
andrewharvey: this test case was originally written to test that when no element is passed to Marker that it will use the default SVG, let's leave that test in place.

test/unit/ui/marker.test.js
```diff
@@ -17,62 +17,201 @@ function createMap() {
...
-        t.ok(marker.getOffset().equals(new Point(0, -14)), 'default marker with no offset uses default marker offset');
+        const el = window.document.createElement(`mapboxgl-marker`);
+        const marker = new Marker(el);
+        t.ok(marker.getElement(this.options), 'default marker is created');
+        t.ok(marker.getAnchor(), `marker set with anchor to default`);
+        t.ok(marker.getOffset(), 'default marker with no offset uses default marker offset');
         t.end();
     });
 
     t.test('default marker with some options', (t) => {
```
andrewharvey: let's leave the original test in place, you can write a new test for the functionality you've added.

test/unit/ui/marker.test.js
```diff
@@ -17,62 +17,201 @@ function createMap() {
...
-        t.ok(marker.getOffset().equals(new Point(0, -14)), 'default marker with no offset uses default marker offset');
+        t.ok(marker.getAnchor(), 'marker set with anchor options');
+        t.ok(marker.getOffset(), 'default marker with no offset uses default marker offset');
         t.end();
     });
 
     t.test('default marker with custom offest', (t) => {
-        const marker = new Marker(null, { offset: [1, 2] });
+        const el = window.document.createElement(`mapboxgl-marker`);
+        const marker = new Marker(el, { offset: [1, 2] });
```
andrewharvey: this test was originally written for the default marker, ie. null must be passed as the element, I'd suggest leaving this test in place and writing a new one for anchor

test/unit/ui/marker.test.js
```diff
@@ -83,23 +222,14 @@ test('Marker', (t) => {
...
-        new Marker(window.document.createElement('div'))
+        new Marker()
             .setPopup(popup)
             .setLngLat([-77.01866, 38.888])
             .addTo(map);
         t.deepEqual(popup.getLngLat(), new LngLat(-77.01866, 38.888));
         t.end();
     });
 
-    t.test('marker centered by default', (t) => {
```
andrewharvey: we still need to test that the marker is centered by default

test/unit/ui/marker.test.js
```diff
@@ -109,16 +239,5 @@ test('Marker', (t) => {
         t.end();
     });
 
-    t.test('marker\'s offset can be changed', (t) => {
```
andrewharvey: we still need to test that the marker's offset can be changed

src/ui/marker.js
jay-manday: this is a good point, i'll make an example based on defaults

src/ui/marker.js
jay-manday: gotcha, thank you for going through this! I might have missed this while resetting git to avoid the conflicts.

test/unit/ui/marker.test.js
jay-manday: will do 👍

test/unit/ui/marker.test.js
jay-manday: Gotcha, thank you for going through this!  I'm still getting used to working with flow and will be making the necessary changes and pushing shortly.

src/ui/marker.js
```diff
@@ -280,30 +318,98 @@ class Marker {
...
+     */
+    getAnchor() {
+        return this.options.anchor;
+    }
+
+    /**
+     * Sets the anchor of the marker
+     * @param {PointLike} [anchor] The anchor in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
+     * @returns {Marker} `this`
+     */
```
jay-manday: is the PointLike an issue here? this smells kind of funny, unless this is interpolating the anchor

andrewharvey: if we refer back to the original ticket #5640, the anchor option should be a string like bottom, top-left, etc. A PointLike is an x,y pixel value.

jay-manday:
Sorry for the delay guys, things got a little hectic at the start of the semester and then got hit hard with the flu 😷 , feeling better though! 

I went through this and things should be working properly now! Please take a look and let me know about any further tweaks and changes 👍

test/unit/ui/marker.test.js
```diff
@@ -17,62 +17,211 @@ function createMap() {
...
         t.end();
     });
 
     t.test('default marker', (t) => {
-        const marker = new Marker();
+        const el = window.document.createElement(`div`);
+        const marker = new Marker(el);
         t.ok(marker.getElement(), 'default marker is created');
-        t.ok(marker.getOffset().equals(new Point(0, -14)), 'default marker with no offset uses default marker offset');
+        t.ok(marker.getOffset(), 'default marker with no anchor and offset options, uses default marker options');
```
andrewharvey: We should leave in a test case for the default marker, ie. no element passed to `new Marker()`.

test/unit/ui/marker.test.js
```diff
@@ -17,62 +17,211 @@ function createMap() {
...
         t.end();
     });
 
     t.test('default marker with some options', (t) => {
```
andrewharvey: in this case, "default marker" is when no element is passed to `Marker`. We should leave in those unit tests.

test/unit/ui/marker.test.js
```diff
@@ -17,62 +17,211 @@ function createMap() {
...
-        t.ok(marker.getOffset().equals(new Point(0, -14)), 'default marker with no offset uses default marker offset');
+        t.ok(marker.getAnchor(), 'marker set with anchor options');
+        t.ok(marker.getOffset(), 'default marker with no offset uses default marker offset');
         t.end();
     });
 
     t.test('default marker with custom offest', (t) => {
-        const marker = new Marker(null, { offset: [1, 2] });
+        const el = window.document.createElement(`div`);
+        const marker = new Marker(el, { offset: [1, 2] });
```
andrewharvey: same as above, this unit test case is for the "default marker", ie. no element passed to `Marker`

test/unit/ui/marker.test.js
```diff
@@ -17,62 +17,211 @@ function createMap() {
...
-        t.ok(marker.getOffset().equals(new Point(1, 2)), 'default marker with supplied offset');
+        t.ok(marker.getAnchor(), 'marker sets with anchor ');
+        t.ok(marker.getOffset(), 'default marker with supplied offset');
         t.end();
     });
 
     t.test('marker is added to map', (t) => {
         const map = createMap();
-        const marker = new Marker(window.document.createElement('div')).setLngLat([-77.01866, 38.888]);
+        const el = window.document.createElement(`div`);
```
andrewharvey: I think we should leave in the existing test as is, and add a new unit test(s) with the anchor option. this helps to ensure backwards compatibility is retained and not broken with this PR.

jay-manday: Gotcha, I added the mentioned tests to my latest push and reverted to the `window.document.createElement('div')` pattern for the rest of the tests

test/unit/ui/marker.test.js
```diff
@@ -16,31 +16,60 @@ function createMap() {
 }
 
 test('Marker', (t) => {
+
     t.test('constructor', (t) => {
-        const el = window.document.createElement('div');
-        const marker = new Marker(el);
-        t.ok(marker.getElement(), 'marker element is created');
+        const marker = new Marker();
+        t.ok(marker, 'marker is created');
```
andrewharvey: the original test case here isn't affected by the introduction of the anchor option so best to leave it as is and add new tests if you have new functionality to test.

test/unit/ui/marker.test.js
```diff
@@ -16,31 +16,60 @@ function createMap() {
...
         t.end();
     });
 
-    t.test('default marker', (t) => {
+    t.test('default marker without element', (t) => {
         const marker = new Marker();
-        t.ok(marker.getElement(), 'default marker is created');
-        t.ok(marker.getOffset().equals(new Point(0, -14)), 'default marker with no offset uses default marker offset');
```
andrewharvey: we should still test that the offset and anchor of the default marker is as expected

test/unit/ui/marker.test.js
```diff
@@ -16,31 +16,60 @@ function createMap() {
...
+        t.ok(marker.getElement(), 'marker returns default element');
+        t.end();
+    });
+
+    t.test('default marker with element', (t) => {
```
andrewharvey: It's not as clear as it could be, but I used the term "default marker" to mean when you create a marker without passing an element, so it uses the default Marker SVG element instead. So "default marker with element" doesn't make sense in that context.

What do you mean by default marker with element?

test/unit/ui/marker.test.js
```diff
@@ -16,31 +16,60 @@ function createMap() {
...
+        t.ok(marker.getOffset(), 'marker returns default offset option');
         t.end();
     });
 
     t.test('default marker with some options', (t) => {
-        const marker = new Marker(null, { foo: 'bar' });
+        const marker = new Marker(null, { anchor: 'bottom',  foo: 'bar' });
         t.ok(marker.getElement(), 'default marker is created');
-        t.ok(marker.getOffset().equals(new Point(0, -14)), 'default marker with no offset uses default marker offset');
+        t.ok(marker.getAnchor(), 'marker set with anchor options');
```
andrewharvey: this only tests that some anchor is returned by getAnchor, we should test that the anchor returned by getAnchor is the same anchor that was set in the initialisation options ie. `bottom`.

test/unit/ui/marker.test.js
```diff
@@ -16,31 +16,60 @@ function createMap() {
...
+        t.ok(marker.getOffset(), 'default marker with no offset uses default marker offset');
         t.end();
     });
 
     t.test('default marker with custom offest', (t) => {
-        const marker = new Marker(null, { offset: [1, 2] });
+        const marker = new Marker(window.document.createElement('div'), { offset: [1, 2] });
```
andrewharvey: default marker in this test means when no element is passed, let's leave the original test in place and you can add new tests for the new anchor functionality.

test/unit/ui/marker.test.js
```diff
@@ -54,25 +83,177 @@ test('Marker', (t) => {
...
+        t.end();
+    });
+
+    t.test('marker centered by default', (t) => {
+        const map = createMap();
+        const marker = new Marker(window.document.createElement('div'))
+            .setLngLat([0, 0])
+            .addTo(map);
+
+        t.ok(marker.setAnchor('middle'));
```
andrewharvey: Did you mean to check in this test that getAnchor is set to middle by default?

andrewharvey:
hi @jay-manday the "automatic anchoring magic" (see https://github.com/mapbox/mapbox-gl-js/pull/6031#discussion_r163733579) is still in this PR. If you strongly feel that it's needed, did you want to put together a use case demo? It seems like a very uncommon use case and we're not sure if it justifies the increased API complexity.

jay-manday:
yes I agree, looking back on things and the prior conversations, allowing users to define a marker offset would complicate things. The edge cases for static maps are really minor compared with ho this is usually used. Also please pardon my over enthusiastic language describing the proposed offset behavior 😅 

I'll go ahead and remove that functionality, thanks again for your patience with me on this PR, I've really learnt a ton, this was my first time writing unit tests and have grown to appreciate the rigor of this work flow. Looking forward to getting this merged (as per approval) so we can move onto fixing other issues : ~ )

andrewharvey:
hi @jay-manday I think there may be a bit of misunderstanding of what the original ticket #5640 is for. We still want to support Marker offset, for example so Markers like this https://github.com/mapbox/makiwich/blob/master/sources/marker-large.svg can be anchored at the tip of the blue dot.

jay-manday:
i'm sorry @andrewharvey, i'm feeling properly silly now. I've reverted to the previous commit. I think the confusion came from mis-interpreting the previous message.

PR #6036, opened 2018-01-22
fix querying features that extend across tile boundaries

### LINKED ISSUE #5756: mouseenter event not firing correctly on circle
When you enter the circle from the **LEFT** side (works fine on other sides, just not the left), the `mouseenter` event doesn't get fired until you reach closer to the center.

Reproducer:
https://jsfiddle.net/0mknjj30/6/

screen shot 2017-11-27 at 1 16 02 pm

I'd expect it to trigger **AS SOON** as my cursor touches the left edge of the circle

ansis:
fix #5756

Some rendered features (circles, symbols, lines) can extend outside of the boundaries of the tile they are stored in. When querying we have to check all tiles within `additionalRadius` of the query.

## Benchmarks
https://bl.ocks.org/anonymous/raw/9762d28049090d90438b81a109918626/
https://bl.ocks.org/anonymous/raw/d69b9f2d20bb2720849c029571929611/

`QueryPoint` is measurably slower. In order to fix the bug we have to do more work so it makes sense and isn't huge change.

## Alternative

I tried a different approach (delaying the short-circuiting until later) that is slightly nicer but a bit [slower](https://bl.ocks.org/anonymous/raw/66e797775ee3016ab78b67f9fe8a5d24/): https://github.com/mapbox/mapbox-gl-js/commit/06d2dd39107646222e57e9683def79adec6839c0

## Launch Checklist

 - [x] post benchmark scores
 - [N/A] manually test the debug page

## Next steps
- porting to -native after review

cc @lilykaiser


src/source/source_cache.js
```diff
@@ -704,14 +704,16 @@ class SourceCache extends Evented {
         for (let i = 0; i < ids.length; i++) {
             const tile = this._tiles[ids[i]];
             const tileID = tile.tileID;
+            const scale = Math.pow(2, this.transform.zoom - tile.tileID.overscaledZ);
+            const additionalRadius = tile.additionalRadius * EXTENT / tile.tileSize / scale;
```
ChrisLoer: Can't getIds() return tiles that haven't loaded yet, in which case `tile.additionalRadius` will be undefined, and this expression will result in `NaN`?

src/source/tile.js
```diff
@@ -166,6 +167,12 @@ class Tile {
             }
         }
 
+        this.additionalRadius = 0;
+        for (const id in this.buckets) {
+            const bucket = this.buckets[id];
+            this.additionalRadius = Math.max(this.additionalRadius, painter.style.getLayer(bucket.layerIds[0]).queryRadius(bucket));
```
ChrisLoer: Since we're storing the `additionalRadius` value here, can't we re-use the value in `Tile#queryRenderedFeature` instead of re-calculating? It's probably not that expensive to calculate, but it still seems like a good code simplification.

src/source/tile.js
```diff
@@ -91,6 +91,7 @@ class Tile {
     refreshedUponExpiration: boolean;
     reloadCallback: any;
     resourceTiming: ?Array<PerformanceResourceTiming>;
+    additionalRadius: number;
```
anandthakker: Maybe rename to something like `renderedFeatureBuffer`

src/source/source_cache.js
anandthakker: Yeah - I think it makes sense to initialize `Tile#additionalRadius` to `0` in the `Tile` constructor. (And if we weren't going to do that, we'd probably want to type it `?number` -- this is a common issue in our flow typing)

src/source/tile.js
ChrisLoer: "buffer" is like "layer" or "source" -- it means too many things in our code base! Maybe "padding" or "query padding"?

anandthakker: Ugh you're so right

ansis:
I made the suggested changes in ea8a2ee and dbe3c23 and then expanded the scope of the pr by implementing support for `circle-pitch-alignment` and `circle-pitch-scale` for querying (
3346dc6, dbe3c23). Would it be better to split off the last two commits into a separate pr?

src/geo/transform.js
```diff
@@ -559,6 +559,13 @@ class Transform {
         this._posMatrixCache = {};
         this._alignedPosMatrixCache = {};
     }
+
+    getPitchScaleFactor() {
```
ChrisLoer: Maybe we could change the name to something like `maxPitchScaleFactor` along with a comment something along the lines of "returns the pitch scaling factor for the very top of the viewport"?

src/data/feature_index.js
```diff
@@ -17,15 +17,19 @@ import type CollisionIndex from '../symbol/collision_index';
...
 import type {FeatureFilter} from '../style-spec/feature_filter';
 import type {CollisionBoxArray} from './array_types';
+import type Transform from '../geo/transform';
 
 const {FeatureIndexArray} = require('./array_types');
 
 type QueryParameters = {
     scale: number,
     bearing: number,
+    cameraToCenterDistance: number,
```
ChrisLoer: Why do we split `cameraToCenterDistance` out as its own parameter in all these calls? It seems like if we're passing in the transform, we can just get it from the transform when we need it?

src/style/style_layer/circle_style_layer.js
```diff
@@ -41,15 +44,56 @@ class CircleStyleLayer extends StyleLayer {
...
+                           posMatrix: Float32Array,
+                           transform: Transform): boolean {
         const translatedPolygon = translate(queryGeometry,
             this.paint.get('circle-translate'),
             this.paint.get('circle-translate-anchor'),
             bearing, pixelsToTileUnits);
-        const radius = this.paint.get('circle-radius').evaluate(feature) * pixelsToTileUnits;
-        const stroke = this.paint.get('circle-stroke-width').evaluate(feature) * pixelsToTileUnits;
-        return multiPolygonIntersectsBufferedMultiPoint(translatedPolygon, geometry, radius + stroke);
+        const radius = this.paint.get('circle-radius').evaluate(feature);
```
ChrisLoer: This all looks right to me but it took me a while to reason through it. Would this sort of explanatory comment help?

```
// For pitch-alignment: map, compare feature geometry to query geometry in the plane of the tile
// Otherwise, compare geometry in the plane of the viewport
// A circle with fixed scaling relative to the viewport gets larger in tile space as it moves into the distance
// A circle with fixed scaling relative to the map gets smaller in viewport space as it moves into the distance
```

ansis:
- comment added
- getPitchFactor renamed to maxPitchScaleFactor
- redundant `cameraToCenterDistance` parameter removed

ChrisLoer:
🎉

Issue #6041, opened 2018-01-23
Collision circles too thick for overzoomed tiles


ChrisLoer:
This was introduced with the fix to issue #5941:

<img width="233" alt="screenshot 2018-01-23 12 48 48" src="https://user-images.githubusercontent.com/375121/35299943-cd289360-003b-11e8-87c9-c8455aa9898c.png">

🤦‍♂️ 

/cc @ansis @nickidlugash

ChrisLoer:
Fixed in #6041.

PR #6060, opened 2018-01-25
Fix debug collision circle drawing on overzoomed tiles (#6041).

ChrisLoer:
Fixes issue #6041.

This just fixes a straightforward oversight of the overscale factor in #5957. However, fixing that exposed a more subtle bug, which is that on an overzoomed tile, when we render circles for a label whose anchor is near the plane of the camera (i.e. _way_ below the bottom of the viewport), the "collision perspective ratio" tends toward infinity, and you can end up with these flashes of blue or red across the screen as one of those gigantic circles momentarily crosses the viewport.

Clamping the perspective ratio to [0,4] is a quick and easy way to correctly render any circles that are reasonably supposed to show up in the viewport. Lots of weird things happen when `w` becomes very different from `camera_to_center_distance` (we call it `camera_to_anchor_distance`, but that becomes less and less true as the two diverge). We end up putting giant geometries into the `CollisionIndex` -- but they're outside the bounds of the grid so they shouldn't matter. The line rendering code in `projection.js` _would_ make giant glyphs, but it has code to hide glyphs that are anchored off-screen. The point label rendering code will make giant glyphs but they won't (I think) cross the viewport so the fragment shader will never run for them.

I started looking into clamping values in all those situations and decided it's best to just leave those monsters sleeping for now. 😬 

/cc @ansis @nickidlugash


mourner:
Nitpick: could we make the render test smaller?

ChrisLoer:
> Nitpick: could we make the render test smaller?

:+1: I pushed a simplified version of the test that just draws a label on one line. Here's what the test would have shown before the fix:

screenshot 2018-01-25 15 37 20

ChrisLoer:
> I pushed a simplified version...

Ack, I hadn't actually pushed it! Updated now. @ansis can you review?

PR #6068, opened 2018-01-26
Ensure drag handlers request render frames on move

### LINKED ISSUE #6063: Pausing with the mousedown while panning prevents any further panning (0.44.0)
Hi,

I'm noticing an issue after upgrading to 0.44.0. Took a while to work out what was going on but I believe it is the following:

v0.44.0

### Steps to Trigger Behavior

 1. Start panning around the map
 2. Stop/pause panning for a moment, but keep holding down the mouse button
 3. Move the mouse to start panning again

### Expected Behavior

Panning should continue after the pause if the user is still holding the mouse button down without having to let go.

### Actual Behavior

The map does not continue to pan (and appears frozen) until the user stops panning (let go of the mouse button) then starts again.

It can look like the map randomly freezes if the user happens to pause for a moment e.g. they pan left, pause, then pan right.

I've tested this in Chrome, Safari and Edge on a couple of computers. I was able to replicate using the mapbox examples.

Thanks,
Joe

anandthakker:
Closes #6063

Caused by https://github.com/mapbox/mapbox-gl-js/pull/6005 because, after that change, the drag handlers relied on the `move` event to trigger a map rerender to pick up the next batch of mouse movements.

This worked fine while dragging was in progress, but after the user paused and the render frame triggered by the last `move` event was complete, there was nothing to re-initiate the render loop once the mouse moved again.

## Launch Checklist

### DIFF
--- a/test/unit/ui/handler/drag_pan.test.js
+++ b/test/unit/ui/handler/drag_pan.test.js
@@ -1,0 +1,34 @@
+'use strict';
+
+const test = require('mapbox-gl-js-test').test;
+const util = require('../../../../src/util/util');
+const window = require('../../../../src/util/window');
+const Map = require('../../../../src/ui/map');
+const DOM = require('../../../../src/util/dom');
+const simulate = require('mapbox-gl-js-test/simulate_interaction');
+
+function createMap(options) {
+    return new Map(util.extend({
+        container: DOM.create('div', '', window.document.body),
+        style: {
+            "version": 8,
+            "sources": {},
+            "layers": []
+        }
+    }, options));
+}
+
+test('DragPanHandler requests a new render frame after each mousemove event', (t) => {
+    const map = createMap();
+    const update = t.spy(map, '_update');
+
+    simulate.mousedown(map.getCanvas(), {bubbles: true, buttons: 2});
+    simulate.mousemove(map.getCanvas(), {bubbles: true, buttons: 2});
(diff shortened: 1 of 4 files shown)


test/unit/ui/handler/drag_pan.test.js
```diff
@@ -0,0 +1,34 @@
...
+    const update = t.spy(map, '_update');
+
+    simulate.mousedown(map.getCanvas(), {bubbles: true, buttons: 2});
+    simulate.mousemove(map.getCanvas(), {bubbles: true, buttons: 2});
+    t.ok(update.callCount > 0);
+
+    // https://github.com/mapbox/mapbox-gl-js/issues/6063
+    update.reset();
+    simulate.mousemove(map.getCanvas(), {bubbles: true, buttons: 2});
+    t.equal(update.callCount, 1);
```
mollymerp: can do `t.ok(update.calledOnce())` as well but this works too!

anandthakker: I like `t.equal(update.callCount, 1)` because if it fails, you see the actual/incorrect value in the test log

pathmapper:
Sorry if it's a stupid question, but I wonder why it has been merged into `0.44 `instead of `master`

anandthakker:
We're using `0.44` as a release branch for putting out patch releases (which we'll do next week).  Backported to master #6073

PR #6071, opened 2018-01-28
CSS linting: Adding stylelint-config-standard

jasonbarry:
This project already uses [`stylelint`](https://github.com/stylelint/stylelint) to enforce that CSS classes are prefixed with `mapboxgl-`. However, no other rules are enforced. 

This PR configures the CSS to adhere to the [`stylelint-config-standard`](https://github.com/stylelint/stylelint-config-standard) configuration, a common set of rules for stricter CSS linting. The result is a neater `dist/mapbox-gl.css` file.

- Upgrades stylelint from 7.10.1 to 8.4.0
- `yarn lint-css` results in 0 errors
- No change in style rules, properties, or selector specificity


dist/mapbox-gl.css
```diff
@@ -1,433 +1,556 @@
...
-.mapboxgl-ctrl-attrib.mapboxgl-compact:after {
-    content: '';
-    cursor: pointer;
-    position: absolute;
-    bottom: 0;
-    right: 0;
-    background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg%20viewBox%3D%270%200%2020%2020%27%20xmlns%3D%27http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%27%3E%0D%0A%09%3Cpath%20fill%3D%27%23333333%27%20fill-rule%3D%27evenodd%27%20d%3D%27M4%2C10a6%2C6%200%201%2C0%2012%2C0a6%2C6%200%201%2C0%20-12%2C0%20M9%2C7a1%2C1%200%201%2C0%202%2C0a1%2C1%200%201%2C0%20-2%2C0%20M9%2C10a1%2C1%200%201%2C1%202%2C0l0%2C3a1%2C1%200%201%2C1%20-2%2C0%27%20%2F%3E%0D%0A%3C%2Fsvg%3E");
-    background-color: rgba(255, 255, 255, .5);
-    width: 24px;
-    h
```
andrewharvey: I don't like the idea of mixing these all together, since they are unrelated, and just happen to have the same values.

It also mixes up the stylesheet document, instead of having attribution styling together and then user-location-dot styling together it's now mixed.

jasonbarry:
@andrewharvey Makes sense — I've reverted that change.

dist/mapbox-gl.css
```diff
@@ -171,55 +236,61 @@ a.mapboxgl-ctrl-logo {
...
-    color: rgba(0,0,0,0.75);
+    color: rgba(0, 0, 0, 0.75);
     text-decoration: none;
 }
+
 .mapboxgl-ctrl-attrib a:hover {
     color: inherit;
     text-decoration: underline;
 }
-/* stylelint-disable */
```
mollymerp: we still need to disable stylelint here and revert the `mapboxgl-improve-map` –> `mapbox-improve-map` per my previous comment

.stylelintrc
```diff
@@ -1,5 +1,7 @@
 {
+  "extends": "stylelint-config-standard",
   "rules": {
-    "selector-class-pattern": "mapboxgl-[a-z-]+"
+    "indentation": 4,
+    "selector-class-pattern": "mapbox(gl)?-[a-z-]+"
```
jasonbarry: @mollymerp I've done that, just in a different location: the .stylelintrc file now has a more flexible selector pattern so that those disable comments aren't needed in the CSS.

mollymerp:
thanks again for the contribution @jasonbarry !

Issue #5728, opened 2017-11-21
Demystifying the black magic of addLayer


jingsam:
I know `addLayer` have a black magic dealing with `source`.

When you want to add a layer and related source to the map, you are supposed to do like this:
```
// 1. add a source
map.addSource('aaa', source)

// 2. add a layer ref to the source
map.addLayer({
  id: 'bbb',
  source: 'aaa'
})
```

In fact, you can do like this:
```
map.addLayer({
  id: 'bbb',
  source: {
     type: 'geojson',
     data: './a.geojson'
  }
})
```

Why you can do that? When you inline a `source` definition in layer, the `addLayer` will create a new source with same `id` of layer.

The docs says:
```
Adds a [Mapbox style layer] to the map's style.
```
The proble is:
1. inline source in layer style is not confront to mapbox-style-spec
2. if we indeed allow inline source in layer, we should make it explict in the docs, as to avoid some confusion in edge case such as conflict with the existing sources.

mollymerp:
For context, support for inline sources was added in #3861

jfirebaugh:
I think inline sources might have been an API design mistake. It was probably responsible for the confusion in #5440.

andrewharvey:
> I think inline sources might have been an API design mistake. It was probably responsible for the confusion in #5440.

Personally I avoid using inline sources as I feel it makes the code less readable, more confusing and less refactorable, especially when you start having multiple layers built from the 1 source.

I go so far as un-inlining sources when copying the examples to build upon. I'm not advocating removing inline source support, but I'd be :+1: for changing the examples back to un-inlined sources. I feel this is especially important for the examples which need to be immediately and easily understood and be a clear extension of the GL JS architecture where layers and sources are probably the most important concepts.

stevage:
Huh. I wasn't really aware you could inline sources like that. Although it looks convenient, I think I'd agree with Andrew that it could get confusing. One thing I like about the Mapbox-GL-JS API is that there is generally only one way to do things, and it has resisted many temptations to add too many conveniences and syntactic sugars etc. (For contrast, try the Tangram API - eep!)

andrewharvey:
Thanks @pathmapper. Given all this and given your comment @jfirebaugh "I think inline sources might have been an API design mistake". Should we change the examples back to the original style?

jimutt:
[...]

PR #6091, opened 2018-02-02
Documentation

george-silva:
Adding a note on the docs about using direct source objects when creating a layer.

## Launch Checklist

I've removed the items I considered to not be relevant for such a small change.


jfirebaugh:
Thanks for the contribution @george-silva. There's an ongoing discussion about this feature in #5728, and current consensus is leaning towards perhaps removing it. If we were to keep it, I think we'd describe it in different terms that you've used here, so I'm going to close this PR, but we do appreciate your work.

Issue #3552, opened 2016-11-07
Feature request: camera elevation


IvanSanchez:
## Motivation

With extruded geometries, there are use cases where the focus should be on a specific elevation.

The most obvious is  to focus on a building rooftop with a call to `Map#flyTo`. With more 3D data, users might want to have a control to travel "up" and "down" besides moving N-S-W-E.

## Design

I suggest a new property in [`CameraOptions`](https://www.mapbox.com/mapbox-gl-js/api/#CameraOptions), named `height` or `elevation`, e.g.:

```
map.flyTo({
  center: [ 63.5, 10.4 ],
  zoom: 17,
  bearing: 15,
  pitch: 45,
  elevation: 100  // assumed meters, or CRS units, as in the extrude height
})
```

### Implementation

I guess the internal camera state shall store the camera elevation somewhere, and modify the 3D transform matrix as appropriate.

This might change the viewport bounds, so the calculations for the tile ranges might change as well (my knowledge of the internals of this project is low, so I don't know if the 3D transform matrix is reused for this purpose).

Any easing function for the camera state (`flyTo` animations and the like) shall also ease and interpolate the elevation.

1ec5:
First of all, any `elevation` option would be mutually exclusive with the existing `zoom` option, because they’re competing ways of scaling the map. Zoom level–based APIs are appropriate for aligning with a given function stop in a style layer, whereas altitude-based APIs are appropriate for aligning to real-world features. (This is, after all, a map rather than an image viewer.)

Specifically, imagine you want to set a camera that comfortably clears the tallest building in a city. You could do it in one line by saying you want a camera with a certain altitude. Or you can guess and check until you come up with the right zoom level. But guessing and checking only works if you hard-code a specific city and a specific building.

[`flyTo`](https://www.mapbox.com/mapbox-gl-js/api/#Map#flyTo) demonstrates this point nicely. A developer can only intuit a value for the `minZoom` option if they’ve designed the map themselves and want to ensure that a certain function stop is reached. But if the goal is to simulate aircraft flying at a certain altitude – say, for a game – a [`peakAltitude`](https://www.mapbox.com/ios-sdk/api/3.3.6/Classes/MGLMapView.html#/c:objc(cs)MGLMapView(im)flyToCamera:withDuration:peakAltitude:completionHandler:) option would be far more intuitive.

[...]

PR #6093, opened 2018-02-02
Animate camera position helpers

elifitch:
This PR adds two helper functions that should be useful for camera animation.  The first `getCameraPosition` returns the lat/long underneath the camera (accounting for pitch), the camera's altitude in meters, pitch and bearing.  The second helper function `setCameraPosition` accepts the same data structure outputted by `getCameraPosition` and sets the map transform to match.

Wrote tests, but looks like they're failing on circle due to a difference in floating point precision I think. Is it reasonable to slightly limit precision with `toFixed`?

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -403,7 +403,7 @@ class Transform {
 
     /**
      * Get the latitude and longitude under the camera, the altitude of the camera in meters, bearing & pitch
-     * @returns {Object} containing longitude, latitude, altitude and pitch
+     * @returns {Object} containing longitude, latitude, altitude and pitch(in degrees)
      */
     getCameraPosition(): { lng: number, lat: number, altitude: number, pitch: number, bearing: number } {
         const pitch = this._pitch;
@@ -413,24 +413,26 @@ class Transform {
         const latLong = this.pointLocation(latPosPointInPixels);
         const verticalScaleConstant = this.worldSize / (2 * Math.PI * 6378137 * Math.abs(Math.cos(latLong.lat * (Math.PI / 180))));
         const altitudeInMeters = altitude / verticalScaleConstant;
+        const pitchInDegrees = pitch * (180 / Math.PI);
 
-        return { lng: latLong.lng, lat: latLong.lat, altitude: altitudeInMeters, pitch, bearing: this.bearing };
+        return { lng: latLong.lng, lat: latLong.lat, altitude: altitudeInMeters, pitch: pitchInDegrees, bearing: this.bearing };
     }
 
     /**
-     * Set camera position to desired latitude, longitude, altitude, pitch and bearing
+     * Set camera position to desired latitude, longitude, altitude, pitch (in degrees) and bearing
      * @param {Object} camPos Object containing the latitude and longitude under the camera, the altitude in meters, bearing & pitch
      */
     setCameraPosition(camPos: { lng: number, lat: number, altitude: number, pitch: number, bearing: number }) {
         const { lng, lat, altitude, pitch, bearing } = camPos;
 
+        const pitchInRadians = pitch * (Math.PI / 180);
         const cameraToCenterDistance = 0.5 / Math.tan(this._fov / 2) * this.height;
-        const pixelAltitude = Math.abs(Math.cos(pitch) * cameraToCenterDistance);
+        const pixelAltitude = Math.abs(Math.cos(pitchInRadians) * cameraToCenterDistance);
         const metersInWorldAtLat = (2 * Math.PI * 6378137 * Math.abs(Math.cos(lat * (Math.PI / 180))));
         const worldsize = pixelAltitude / altitude * metersInWorldAtLat;
         const zoom = Math.log(worldsize / this.tileSize) / Math.LN2;
 
-        const latOffset = Math.tan(pitch) * cameraToCenterDistance;
+        const latOffset = Math.tan(pitchInRadians) * cameraToCenterDistance;
         const newPixelPoint = new Point(this.width / 2, this.height / 2 + latOffset);
         const newLongLat = new LngLat(lng, lat);
 
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -402,10 +402,10 @@ class Transform {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


elifitch:
Benchmark results 
2018-02-02-17-41-localhost_9966

ryanhamley:
Still interested in having these methods @elifitch? I think fixing the decimal places to 7 or 8 digits is fine since that corresponds to ~1 cm to 1 mm accuracy; any decimals more than that are basically superfluous.

src/geo/transform.js
```diff
@@ -385,6 +385,47 @@ class Transform {
...
+    /**
+     * Get the latitude and longitude under the camera, the altitude of the camera in meters, bearing & pitch
+     * @returns {Object} containing longitude, latitude, altitude and pitch(in degrees)
+     */
+    getCameraPosition(): { lng: number, lat: number, altitude: number, pitch: number, bearing: number } {
+        const pitch = this._pitch;
+        const altitude = Math.cos(pitch) * this.cameraToCenterDistance;
+        const latOffset = Math.tan(pitch) * this.cameraToCenterDistance;
+        const latPosPointInPixels = this.centerPoint.add(new Point(0, latOffset));
+        const latLong = this.pointLocation(latPosPointInPixels);
```
ryanhamley: Let's call this `lngLat` to correspond to our naming conventions elsewhere.

src/geo/transform.js
```diff
@@ -385,6 +385,47 @@ class Transform {
...
+        const verticalScaleConstant = this.worldSize / (2 * Math.PI * 6378137 * Math.abs(Math.cos(latLong.lat * (Math.PI / 180))));
```
ryanhamley: Can we add a comment here explaining the magic number `6378137` and what this constant represents?

src/geo/transform.js
```diff
@@ -385,6 +385,47 @@ class Transform {
...
 
+    /**
+     * Get the latitude and longitude under the camera, the altitude of the camera in meters, bearing & pitch
+     * @returns {Object} containing longitude, latitude, altitude and pitch(in degrees)
+     */
+    getCameraPosition(): { lng: number, lat: number, altitude: number, pitch: number, bearing: number } {
+        const pitch = this._pitch;
+        const altitude = Math.cos(pitch) * this.cameraToCenterDistance;
+        const latOffset = Math.tan(pitch) * this.cameraToCenterDistance;
+        const latPosPointInPixels = this.centerPoint.add(new Point(0, latOffset));
```
ryanhamley: I think we can simplify this a bit and not define this variable at all by doing `const lngLat = this.pointLocation(this.centerPoint.add(new Point(0, latOffset));`

src/geo/transform.js
```diff
@@ -385,6 +385,47 @@ class Transform {
...
+     */
+    setCameraPosition(camPos: { lng: number, lat: number, altitude: number, pitch: number, bearing: number }) {
+        const { lng, lat, altitude, pitch, bearing } = camPos;
+
+        const pitchInRadians = pitch * (Math.PI / 180);
+        const cameraToCenterDistance = 0.5 / Math.tan(this._fov / 2) * this.height;
+        const pixelAltitude = Math.abs(Math.cos(pitchInRadians) * cameraToCenterDistance);
+        const metersInWorldAtLat = (2 * Math.PI * 6378137 * Math.abs(Math.cos(lat * (Math.PI / 180))));
+        const worldsize = pixelAltitude / altitude * metersInWorldAtLat;
+        const zoom = Math.log(worldsize / this.tileSize) / Math.LN2;
```
ryanhamley: Let's just set `this.zoom = Math.log(worldsize / this.tileSize) / Math.LN2` instead on line 423.

src/geo/transform.js
```diff
@@ -385,6 +385,47 @@ class Transform {
...
+
+        const latOffset = Math.tan(pitchInRadians) * cameraToCenterDistance;
+        const newPixelPoint = new Point(this.width / 2, this.height / 2 + latOffset);
+        const newLongLat = new LngLat(lng, lat);
```
ryanhamley: `const newLngLat`

src/geo/transform.js
elifitch: @ansis it's been ages since we whipped this up; do you recall what was up with this magic # ?

mourner:
Closing as stale — feel free to reopen if there's still interest in shipping this.

src/geo/transform.js
andrewharvey: fixed

src/geo/transform.js
andrewharvey: fixed

src/geo/transform.js
andrewharvey: fixed

src/geo/transform.js
andrewharvey: fixed

src/geo/transform.js
andrewharvey: I've expanded this and added a comment. Per https://github.com/mapbox/mapbox-gl-js/issues/8777#issuecomment-533246706 we could eventually move this to a global constant, but right now I'm just cleaning up this stale PR to investigate if it could still be merged or not.

andrewharvey:
I'm interested in this feature, due to https://github.com/isaacs/github/issues/361 it can't be reopened so I've opened a new draft PR #8830 

> Still interested in having these methods @elifitch? I think fixing the decimal places to 7 or 8 digits is fine since that corresponds to ~1 cm to 1 mm accuracy; any decimals more than that are basically superfluous.
>
> Edit: What's the use case for these?

Both getting and setting the camera's elevation is useful, when working with extrusions.

Being able to set the camera elevation (or equivalent zoom) is critical for being able to have the camera at an extrusion from above without being zoomed in too so that you're inside the extrusion, or too far out that the extrusion is just a small dot on the map is useful.

I'm also interested in tying this into #3552 to be able to look at the top of an extrusion. 

> Should they be exposed on the Map object/class? Or are you imagining other methods that would make use of these internal methods?

Given Map.transform isn't exposed publicly, I do think we should expose them via the Map class as public methods.

Issue #6074, opened 2018-01-28
[regression] 0.44.0 map.queryRenderedFeatures() completely ignores symbols


mike-marcacci:
Upon upgrading to 0.44.0, map.queryRenderedFeatures() fails to return any `symbol` layers that use GeoJSON sources specified in the second argument. I've gone through my layers' layout properties one-by-one to see if there is one particular trigger, and there doesn't seem to be.

Do note, though, that rarely map.queryRenderedFeatures *will* return a symbol layer, but after just another second of map interaction, stops again. Downgrading to 0.43.0 fixes this.

[VIEW SCREENCAST](https://youtu.be/QSMfCEOEnAA)

Check out #6075 for a screencast of "expected" behavior.

pathmapper:
Running into the same issue with **overzoomed** vector tile sources hosted by Mapbox.

Steps to reproduce using this [jsfiddle](https://jsfiddle.net/5550r1mc/):

1. Hover over the green circle -> Properties are shown on the right side of the map (using the the queryRenderedFeatures API)

2. Zoom further in

3. Hover again over the green circle -> Properties are NOT shown on the right side of the map

Please note: 
Sometimes it is necessary to repeat steps 2 and 3 to reproduce.
Zoom extend of the used vector tileset for the point: z11-z12


![query_symbol](https://user-images.githubusercontent.com/20856381/35633383-5fe988f8-06a9-11e8-8b78-1722d603bebc.gif)

anandthakker:
@mike-marcacci Thanks for reporting and for the screencast.  Is there any chance you'd be able to share a live page that reproduces the issue?  So far, I haven't managed to reproduce it locally. 
 (@pathmapper's jsfiddle above does reliably show an issue [thanks!], but it's not clear that it's the same as the one you're seeing)

mike-marcacci:
Hi @anandthakker, thanks for the reply. I don't have anything live running this version (I'm just running the previous one instead), but I'll try to get a simplified working demo live today.

anandthakker:
I believe the overzooming problem that [@pathmapper described above](https://github.com/mapbox/mapbox-gl-js/issues/6074#issuecomment-361981622) happens like so:

- [render frame 1] Map starts at z12, so initial symbol placement occurs using tile `12/1283/1587`.

- Zoom to z13; this means that we have a new, overzoomed version of the above tile (`13/1283/1587` in the debug view)

[...]

PR #6104, opened 2018-02-07
Fix bug preventing symbol features in overscaled from appearing in queryRenderedFeature results

anandthakker:
This resolves the issue described in https://github.com/mapbox/mapbox-gl-js/issues/6074#issuecomment-363237163 by unconditionally updating `style.placement` and `style.collisionIndex`, rather than only updating them if `placementChanged`.

This necessitated two other changes:
1. Replace `placementChanged` with `placementCommitted`, and use it to make sure that we invoke `updateLayerOpacities` any time `style.placement` is updated.
2. Add a `Placement#lastPlacementChangeTime` property, which is updated only in a `commit()` where placement was actually changed, and otherwise inherited from the previous placement.  Use this new property, rather than `commitTime` for the `hasTransitions()` check.  This wasn't necessary before, because previously we simply threw away the new placement (with its newer commit time) if `commit` resulted in no changes -- whereas now, since we save the new placement even when it's a no-op, we want to treat it as an older commit in that case so that `hasTransitions()` yields false and we stop asking for a new render.

Benchmarks: https://bl.ocks.org/anonymous/raw/ea32279a8eaada909b3111ad2682579a/

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -954,7 +954,7 @@ class Style extends Evented {
 
     _updatePlacement(transform: Transform, showCollisionBoxes: boolean, fadeDuration: number) {
         let symbolBucketsChanged = false;
-        let placementChanged = false;
+        let placementCommitted = false;
 
         const layerTiles = {};
 
@@ -985,17 +985,20 @@ class Style extends Evented {
             this._layerOrderChanged = false;
         }
 
-        if (!this.pauseablePlacement.isDone()) {
+        if (this.pauseablePlacement.isDone()) {
+            // the last placement finished running, but the next one hasn’t
+            // started yet because of the `stillRecent` check immediately
+            // above, so mark it stale to ensure that we request another
+            // render frame
+            this.placement.setStale();
+        } else {
             this.pauseablePlacement.continuePlacement(this._order, this._layers, layerTiles);
 
             if (this.pauseablePlacement.isDone()) {
-                const placement = this.pauseablePlacement.placement;
-                placementChanged = placement.commit(this.placement, browser.now());
-                if (!this.placement || placementChanged || symbolBucketsChanged) {
-                    this.placement = placement;
-                    this.collisionIndex = this.placement.collisionIndex;
-                }
-                this.placement.setRecent(browser.now(), placement.stale);
+                this.pauseablePlacement.placement.commit(this.placement, browser.now());
+                this.placement = this.pauseablePlacement.placement;
+                this.collisionIndex = this.placement.collisionIndex;
+                placementCommitted = true;
             }
 
             if (symbolBucketsChanged) {
@@ -1004,12 +1007,9 @@ class Style extends Evented {
                 // placement is already stale while it is in progress
                 this.pauseablePlacement.placement.setStale();
             }
-
-        } else {
-            this.placement.setStale();
         }
 
-        if (placementChanged || symbolBucketsChanged) {
+        if (placementCommitted || symbolBucketsChanged) {
             for (const layerID of this._order) {
                 const styleLayer = this._layers[layerID];
                 if (styleLayer.type !== 'symbol') continue;
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -33,7 +33,6 @@ import type Transform from '../geo/transform';
 import type {Source} from '../source/source';
 import type {StyleImage} from './style_image';
 import type {StyleGlyph} from './style_glyph';
-import type CollisionIndex from '../symbol/collision_index';
 import type {Callback} from '../types/callback';
 import type EvaluationParameters from './evaluation_parameters';
 import type Placement from '../symbol/placement';
@@ -91,7 +90,6 @@ class Style extends Evented {
     _updatedPaintProps: {[layer: string]: true};
     _layerOrderChanged: boolean;
 
-    collisionIndex: CollisionIndex;
     crossTileSymbolIndex: CrossTileSymbolIndex;
     pauseablePlacement: PauseablePlacement;
     placement: Placement;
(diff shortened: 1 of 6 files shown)


anandthakker:
@ChrisLoer @ansis you can check out the original bug using the update to the debug page in the first commit here: load `debug/index.html` and click on the green dot.  On some page loads, you'll get no features, and on others, you'll get one.  I'll revert that debug page once we're happy with the changes here.

ChrisLoer:
@anandthakker I'm afraid I may have led you down the wrong path. 😞 

It doesn't work to keep the old placement commit time as the code stands now because the "commit" logic increments the opacities based on the last commit time -- if you keep the same "commit time" through multiple commits, you end up with double counting. This will be mostly hidden with our default settings because the fade duration is only 300ms, but if you up the fade duration you should be able to see fades get short circuited.

I see a couple ways forward:

- Give up on this whole approach and try the fix you were originally considering that I pushed you away from (setting a kind of "dirty" flag if symbolBucketsChanged is true when you pause a placement).
- Push forward with the "always commit" approach, but instead of messing with the "commit time", track a separate "lastChange" time and use that for calculating `hasTransitions`. I _think_ this could land at a place where the names felts pretty logical, but there is a bit of a smell to adding the extra "lastChange" parameter.

Random other points:

- Without doing much reorganization, we could just flip the `if (!this.pauseablePlacement.isDone())` clause to put the common `setStale` case on top -- so that would happen close to the `stillRecent` logic and be a little easier to understand (maybe with a one line comment explaining what it's doing).
- We can simplify the logic in `_updatePlacement` by pushing more of it into `commit`. Like, `commit` should also take care of `setRecent`, and it should return `this` so that we can just say something like `this.placement = pausablePlacement.commit(...`

anandthakker:
> It doesn't work to keep the old placement commit time as the code stands now because the "commit" logic increments the opacities based on the last commit time -- if you keep the same "commit time" through multiple commits, you end up with double counting.

Is this a problem even though we'd only be keeping the old placement commit time in the case where `commit()` had no effect, which, in the current implementation, would result in simply throwing away the new placement (with its new commit time) anyway?

I.e.:

placement | master | branch
-----------|--------|--------
`placement.commit(null, 0)` | Update `this.placement`, so `this.placement.commitTime === 0` | Update `this.placement`, so `this.placement.commitTime === 0`
`placement.commit(previous, 100)` => some placement changes | update `this.placement`, so `this.placement.commitTime === 100` | update `this.placement`, and use fresh commit time, so `this.placement.commitTime === 100`
`placement.commit(previous, 200)` => no placement changes | DON'T update `this.placement`, so `this.placement.commitTime` is still 100 | DO update `this.placement`, but copy previous commitTime, so `this.placement.commitTime === 100`

ChrisLoer:
That table is :100:! So the `commitTime` is the same in both contexts, but the thing to look at is the entries in `placement.opacities`. When we get to the end of the 3rd placement:

- On `master`, you're still using the 2nd placement, which means you're using its opacities, which have been modified by `increment` once since the first placement. They're trued up to a 100ms commit time, and when you render them at 200ms we effectively add 100ms to them at render time (via setting a uniform for the shader) without modifying the opacity stored in the placement.
- On `6104`, you're using the 3rd placement, which is still using a commitTime of 100, but has had an extra `increment` applied to all the opacities. So the `increment` will have trued the opacities up to their expected value at 200, but when you go to render you're going to play them forward an extra 100ms because the placement still has the old commit time. They look like they would at 400ms.

anandthakker:
Ah hah, gotcha @ChrisLoer.  Okay, now that I understand better, I'm inclined to push through with the second approach... I'll give that a whirl now

anandthakker:
@ChrisLoer updated and ready for another 👁‍🗨

src/symbol/placement.js
```diff
@@ -242,7 +240,15 @@ class Placement {
             }
         }
 
-        return placementChanged;
+        // this.lastPlacementChangeTime is the time of the last commit() that
+        // resulted in a placement change -- in other words, the start time of
+        // the last symbol fade animation
+        assert(!prevPlacement || typeof prevPlacement.lastPlacementChangeTime === 'number');
```
ChrisLoer: Why `typeof ... === 'number'` instead of just using the boolean cast of "undefined"?

anandthakker: The latter would yield false for `0`... which would never happen for real `browser.now()` values, but be an annoyance in unit tests or something

src/style/style.js
```diff
@@ -985,17 +985,20 @@ class Style extends Evented {
...
 
             if (this.pauseablePlacement.isDone()) {
-                const placement = this.pauseablePlacement.placement;
-                placementChanged = placement.commit(this.placement, browser.now());
-                if (!this.placement || placementChanged || symbolBucketsChanged) {
-                    this.placement = placement;
-                    this.collisionIndex = this.placement.collisionIndex;
-                }
-                this.placement.setRecent(browser.now(), placement.stale);
+                this.pauseablePlacement.placement.commit(this.placement, browser.now());
```
ChrisLoer: _"placement placement placement placement!"_ 😉 

This is already a great improvement, but maybe we can go further here by giving `PausablePlacement` its own commit wrapper that handles some of this state tracking? Ideally these three lines could be one line.

src/symbol/placement.js
```diff
@@ -362,16 +368,13 @@ class Placement {
...
-        return this.symbolFadeChange(now) < 1 || this.stale;
+        return this.stale ||
+            now - this.lastPlacementChangeTime < this.fadeDuration;
     }
 
     stillRecent(now: number) {
-        return this.recentUntil > now;
-    }
-
-    setRecent(now: number, stale: boolean) {
```
ChrisLoer: Love how this change gets rid of the need for `setRecent/recentUntil` -- no net increase in internal state!

src/symbol/placement.js
ChrisLoer: Ok fair enough. How about using `!== undefined`? I'm just looking for a form that expresses as simply as possible that this is about whether the initial value is set, not about the value possibly having multiple types.

anandthakker:
@ChrisLoer 👍 updated to address your most recent comments.

anandthakker:
> as we don't see any regression in the benchmarks 

😢 alas:

screen shot 2018-02-12 at 10 24 05 am

anandthakker:
Fortunately, that Paint performance regression turned out to be due to a simple typo.  Fixed, and now the benchmarks look good.

ChrisLoer:
> Fortunately, that Paint performance regression turned out to be due to a simple typo.

Phew! 😅

PR #6114, opened 2018-02-08
Remove unnecessary files from package files list

LitoMore:
The `src` files should not publish to the npm registry, we should only publish the `dist` files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,7 +162,6 @@
     "build/",
     "dist/",
     "flow-typed/",
-    "src/",
     ".flowconfig"
   ]
 }


anandthakker:
@LitoMore we intentionally publish src to NPM so that developers who prefer to use flow-annotated sources are able to do so.

LitoMore:
@anandthakker But this will make the package became twice big as only dist files. It's not a good idea for production use.

jfirebaugh:
I don't think we're likely to change our stance here; disk space is cheap. If you really need a super-minimal package, you can publish `LitoMore:remove-unnecessary-files` as a [scoped package](https://docs.npmjs.com/getting-started/scoped-packages).

PR #6123, opened 2018-02-09
Fix `layer_properties` template in order to generate valid-flow empty PaintProps

Jesus89:
This PR adds a fix in the `layer_properties.js.ejs` file to prevent the generation of non-flow-compatible code when the `paintProperties` array is empty. Instead of the following code:

```js
export type PaintProps = {|
|};
```

now the following code is generated:

```js
export type PaintProps = {};
```

Thanks!


mourner:
Thank you!

Issue #5574, opened 2017-11-01
SourceCache checks status 404 but status code not sent in postMessage


deanm:
When one of the web workers makes an AJAX request that fails, it will throw an AJAXError (with status: Number) back to the callback.  Ex:

https://github.com/mapbox/mapbox-gl-js/blob/25e5f1dd9d55835d9572e15893740aa90aae9eeb/src/source/vector_tile_source.js#L107

However, when this goes through the actor, this AJAXError is simply converted to a string for the message posted back to the main thread:

https://github.com/mapbox/mapbox-gl-js/blob/25e5f1dd9d55835d9572e15893740aa90aae9eeb/src/util/actor.js#L68

It looks like _tileLoaded however is specifically trying to handle the condition for 404:

https://github.com/mapbox/mapbox-gl-js/blob/25e5f1dd9d55835d9572e15893740aa90aae9eeb/src/source/source_cache.js#L228

However the `err` object will have just been a generic `Error` recreated from the string:

https://github.com/mapbox/mapbox-gl-js/blob/25e5f1dd9d55835d9572e15893740aa90aae9eeb/src/util/actor.js#L77

So it seems like there is some logic to handle 404 errors specially, however as the network fetches are usually done by the source workers, the status code doesn't make it back to that code.

mollymerp:
fixed in #6126

Issue #6094, opened 2018-02-03
Map unresponsive to events: `Error at Actor.receive (actor.js:81)` and `XHR failed loading: GET "https://a.tiles.mapbox.com/v4/chanversus.40louhe4/10/300/384.vector.pbf...`


elainechan:
**mapbox-gl-js version**: 0.44.0
### Steps to Trigger Behavior
 1. Load the map app
 2. Scrolling, dragging, clicking, or hovering using the mouse.
 3. Reload the app and repeat.
### Expected Behavior
- The map should respond to these events:	
1. Symbol `on click`: display information on the symbol in a popup. Feature names: `Bike parking`, `Public benches`.
4. Polygon `on hover`: when the `Neighborhoods` _vector tile_ layer is active, hovering over a neighborhood polygon highlights it and prints its information in a popup in the bottom of the screen.
5. Map `on scroll`: zoom in or out of a map by scrolling.
6. Map `on drag`: move to a different place on the map by dragging.
### Actual Behavior
- The map renders _styles_ and custom _vector tile_ layers successfully, but after it finishes rendering, the map does not respond to any user events.
- All the features of the map app depend entirely upon _vector tiles_ responding to events. As a result, the app is completely broken. It is equivalent to displaying a static image on a browser.
- Perhaps similar to [this issue](https://github.com/mapbox/mapbox-gl-js/issues/4858) with loading PBF files, although not the same. Also the problem does not only affect _vector tile_ PBF files; it affects the main _style_ files as well. 

## Additional Information
- The [static app](https://elainechan.github.io/mapping/) is hosted on GitHub Pages and the source code is [here](https://github.com/elainechan/mapping). 
- A detailed debugging report can be found [here](https://github.com/elainechan/mapping/tree/master/debug).

### Logged Errors
- Recurring error: this error comes up a dozen of times during each run.
```
Error: Error
    at Actor.receive (actor.js:81)
Evented.fire	@	evented.js:109
Evented.fire	@	evented.js:103
Evented.fire	@	evented.js:103
t._tileLoaded	@	source_cache.js:238
i	@	vector_tile_source.js:125
Actor.receive	@	actor.js:81
```
- Loading failures: it seems like most of the offending files are `pbf` custom vector tiles I created on Mapbox Studio by uploading data.
[Screenshot](https://github.com/elainechan/mapping/blob/master/debug/Screen_2018-02-03_1.50.06AM.png)

```
[...]

PR #6126, opened 2018-02-09
fix web worker error handling

mollymerp:
changes from #4446 regressed with the new Web Worker transfer, and this was causing hard to diagnose errors like 

```
Error: Error
    at Actor.receive (actor.js:81)
```

because `String(err)` was just outputting "Error" with no status code. 

Related to #6094


jfirebaugh:
@mollymerp Does this fix #5574?

mollymerp:
yes! good catch.

Issue #6116, opened 2018-02-09
Enum type missing from mapbox-gl-style-spec doc page


samanpwbb:
I expect to see the enum type in this list at https://www.mapbox.com/mapbox-gl-js/style-spec/#types, but it is not present.

![screen shot 2018-02-08 at 4 40 23 pm](https://user-images.githubusercontent.com/108094/36005935-f6f243c0-0cee-11e8-9b2a-1189aa4f3527.png)

In the spec, [properties like `text-transform`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/reference/v8.json#L1679) use the enum type.

jay-manday:
Hi @samanpwbb I can take care of this, I've been meaning to review the style specs as per #5438 and this is a good place to start!

jfirebaugh:
These were intentionally removed; see #5577.

Issue #5577, opened 2017-11-01
Remove "enum" type from documentation


jfirebaugh:
We don't really have enums in the expression type system, and I don't think they're adding much clarity in the documentation either. We should just document these properties as having type "string" with some requirements. Our documentation text already lists the possible values, it should just replace "enum" with "string" and format the possible values as string literals:

> #### visibility
> Layout property. Optional [_string_](https://www.mapbox.com/mapbox-gl-js/style-spec/#types-string). One of `"visible"`, `"none"`. Defaults to `"visible"`.

jingsam:
what about "color",?JSON have no color types, they are strings.

anandthakker:
Good point, @jingsam  --  since color is a distinct type in the expression type system, when we encounter string values where a color is expected, we automatically coerce them to color values (using [`to-color`](https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-types-to-color)), so the problem described here fortunately doesn't extend to colors.

jfirebaugh:
It's true we're using a string representation for colors in JSON. But it is a distinct, full-fledged type in the style specification domain -- certain properties are colors, not just arbitrary strings, and there are expressions operators that require string arguments, or return string results.

hyperknot:
The docs are still linking to the non-existing enum types though, on this link: https://www.mapbox.com/mapbox-gl-js/style-spec/#types-enum

Issue #5438, opened 2017-10-10
Add unit test checking that each expression type is documented


anandthakker:
Check that each expression has an entry in the style spec JSON and in `docs/style_spec/_generate/expression-metadata.js`, and vice versa

jay-manday:
hi Anand, been doing some experimenting with mapbox gl and really into it, is this issue still open?

anandthakker:
Yes, still open!

jay-manday:
Cooool sure thing, does this entail cross reffing [style-spec.js](https://github.com/mapbox/mapbox-gl-js/blob/6465c04107f34dceb2c16bfd824fdf8c3380fc53/docs/pages/style-spec.js) with [the expression tests](https://github.com/mapbox/mapbox-gl-js/tree/f839a8c25725cf799bf9026e76792128893024f2/test/integration/expression-tests) just curious on where is a good place to start? is there some type of reg-ex search you'd typically use to make the process go quicker? to clarify this issue is just making sure there is parity between the style-specs for methods and the docs, correct?

anandthakker:
> to clarify this issue is just making sure there is parity between the style-specs for methods and the docs, correct?

@jay-manday the docs are generated _from_ the style spec (and [expression_metadata.js](https://github.com/mapbox/mapbox-gl-js/blob/ea28e719f80c760347deabfc799c4bcc7e1feb05/docs/components/expression-metadata.js)), so this issue is about making sure that each expression in the 'registry' exported by [style-spec/expression/definitions/index.js](https://github.com/mapbox/mapbox-gl-js/blob/ea28e719f80c760347deabfc799c4bcc7e1feb05/src/style-spec/expression/definitions/index.js) has a corresponding entry in the [style spec JSON](https://github.com/mapbox/mapbox-gl-js/blob/ea28e719f80c760347deabfc799c4bcc7e1feb05/src/style-spec/reference/v8.json#L2015) and expression-metadata.js

jay-manday:
Thanks for clarifying! Happy to work on it :)

PR #6127, opened 2018-02-09
added support for enum type in style-spec as per #6116

jay-manday:
## Launch Checklist

I added support for Enum types in the style spec as requested by @samanpwbb in #6116, this feels right but if anyone sees anything funky please lmk! 👍 

### DIFF
--- a/src/style-spec/expression/index.js
+++ b/src/style-spec/expression/index.js
@@ -246,6 +246,7 @@ const {
     NumberType,
     BooleanType,
     ValueType,
+    EnumType,
     array
 } = require('./types');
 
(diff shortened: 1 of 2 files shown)


anandthakker:
Thanks for contributing, @jay-manday !

This PR adds an `enum` type to the _expression_ type system, but (at least for now), we've chosen to treat enums as strings during expression evaluation.  So while, for example, `line-join` is defined in the _style specification_ as an enum, an expression for `line-join` would just be typed `string`.

I believe @samanpwbb 's intent in #6116 was that the enum type that's used for various properties in the style specification should be listed in the documentation page under "Types".

Sorry for the confusion!

jay-manday:
@anandthakker, of course! n_n

ahhhh I see, I see, no worries! Thanks for explaining, I think I read to deeply into your comment on #5438 mentioning how the docs are generated from the style spec. I'll reset this branch and push the changes @samanpwbb wanted :)

jfirebaugh:
Hold off on that too actually -- we removed enums there intentionally. See #5577 and #5578.

jay-manday:
ahhh gotcha sounds good

PR #6128, opened 2018-02-09
add Map#isZooming method

### LINKED ISSUE #6119: We need isZooming api
**mapbox-gl-js version**: 0.44.0

### Steps to Trigger Behavior
- no Steps :)

Hello mapboxgl.
Thanks for your good library.
I know that mapbox has isMoving(), but no isZooming.
Is there any reason ?

mollymerp:
Resolves #6119 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -83,6 +83,7 @@ class Camera extends Evented {
     constructor(transform: Transform, options: {bearingSnap: number}) {
         super();
         this.moving = false;
+        this.zooming = false;
         this.transform = transform;
         this._bearingSnap = options.bearingSnap;
     }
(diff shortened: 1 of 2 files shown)


src/ui/camera.js
```diff
@@ -855,7 +855,17 @@ class Camera extends Evented {
      * @returns A Boolean indicating whether the camera is moving.
      */
     isMoving(): boolean {
-        return this.moving;
+        return !!this.moving;
```
lbud: Since `moving` and `zooming` are both boolean members of the Camera class already, let's remove the double negation here and below and just return `this.moving` / `this.zooming`.

mollymerp: yep, `this.zooming` wasn't initialized in the constructor, so it could be `undefined` but I've initialized it as false in a new commit and removed the `!!`

PR #6135, opened 2018-02-12
Fix incorrect textPixelRatio argument

### LINKED ISSUE #6075: [bug] map.queryRenderedFeatures() doesn't account for {text,icon}-offset or {text,icon}-rotate
NOTE: this bug cannot be reproduced in mapbox-gl-js 0.44.0, because it is blocked by [this other, more serious bug](https://github.com/mapbox/mapbox-gl-js/issues/6074).

Using v0.43.0, map.queryRenderedFeatures() doesn't account for the `icon-rotate` layout property of `symbol` layers, and returns runs as if it were set to `0`.

[VIEW SCREENCAST](https://youtu.be/SvuS24twwKs)

anandthakker:
Closes #6075

`textPixelRatio` is used to scale tile coordinates to viewport coordinates; the value that was being passed here, EXTENT / tileSize, was the reciprocal of the ratio needed -- we were passing 16 instead of 0.0625.

As a result, the corners of the viewport-projected collision box that we reconstructed here, https://github.com/mapbox/mapbox-gl-js/blob/f13c86ea356c384fdab31855b9152f5bf5ef97b8/src/symbol/collision_index.js#L322-L325, were 16^2 times farther from the anchor than they should have been. With `icon-offset: [0, 0]`, this didn't matter, because the incorrectly large box would contain the (much smaller) correct box.  But when there's an offset, it too gets scaled too much, and so we run out of luck.

### DIFF
--- a/src/data/feature_index.js
+++ b/src/data/feature_index.js
@@ -122,7 +122,7 @@ class FeatureIndex {
         this.filterMatching(result, matching, this.featureIndexArray, queryGeometry, filter, params.layers, styleLayers, args.bearing, pixelsToTileUnits);
 
         const matchingSymbols = args.collisionIndex ?
-            args.collisionIndex.queryRenderedSymbols(queryGeometry, this.tileID, EXTENT / args.tileSize, args.collisionBoxArray, args.sourceID, args.bucketInstanceIds) :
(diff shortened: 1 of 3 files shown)


ChrisLoer:
Although this change is the simplest fix, it's worth noting that on gl-native we skipped this issue entirely by storing viewport coordinates in the `GridIndex` and just comparing the viewport-projected query directly against those coordinates instead of re-projecting tile coordinates the way we do in gl-js.

Issue #6137, opened 2018-02-13
Update ScaleControl unit of distance on the fly


ryanhamley:
## Motivation

It should be possible to programatically change the ScaleControl's unit of distance and have the ScaleControl instantly update. This is helpful for applications that may need to show scale's in different units based on a site's location.

## Design Alternatives

You could remove the scale and add it back when you want to change the unit.

Alternatively, you could provide a method to update the scale's internal unit and update the scale.

## Design

Removing and re-adding the control is inefficient and can result in layout issues if multiple controls are in the same container. Updating the internal unit is fairly straightforward and only requires calling the existing `updateScale` method. I believe exposing this method is the proper way to achieve this goal.

### Mock-Up

N/A

### Concepts

N/A

### Implementation

Proposed [pull request here](https://github.com/mapbox/mapbox-gl-js/pull/6138)

kriscarle:
In case it helps anyone, I've put together a custom scale control that shows metric and imperial at the same time, along with cursor coordinates https://github.com/maphubs/mapbox-gl-dual-scale-control

![image](https://user-images.githubusercontent.com/2431570/36232057-4e6517cc-11ae-11e8-9e55-58a2bd5b4513.png)

jfirebaugh:
This was added in #6274.

PR #6138, opened 2018-02-13
Scale update units

ryanhamley:
Potential fix for #6137

## Launch Checklist

Adds a method to the ScaleControl to update the scale's unit of distance on the fly

screen capture on 2018-02-13 at 00-32-33

### DIFF
--- a/src/ui/control/scale_control.js
+++ b/src/ui/control/scale_control.js
@@ -67,7 +67,7 @@ class ScaleControl {
     /**
      * Set the scale's unit of the distance
      *
-     * @param {string} Unit of the distance (`'imperial'`, `'metric'` or `'nautical'`).
+     * @param {string} unit Unit of the distance (`'imperial'`, `'metric'` or `'nautical'`).
      */
     setUnit(unit: string) {
         this.options.unit = unit;
--- a/src/ui/control/scale_control.js
+++ b/src/ui/control/scale_control.js
@@ -67,7 +67,7 @@ class ScaleControl {
     /**
      * Set the scale's unit of the distance
      *
-     * @param {String} unit
+     * @param {string} Unit of the distance (`'imperial'`, `'metric'` or `'nautical'`).
      */
     setUnit(unit: string) {
         this.options.unit = unit;
--- a/src/ui/control/scale_control.js
+++ b/src/ui/control/scale_control.js
@@ -18,10 +18,12 @@ const defaultOptions = {
  * @param {number} [options.maxWidth='100'] The maximum length of the scale control in pixels.
  * @param {string} [options.unit='metric'] Unit of the distance (`'imperial'`, `'metric'` or `'nautical'`).
  * @example
- * map.addControl(new mapboxgl.ScaleControl({
+ * var scale = map.addControl(new mapboxgl.ScaleControl({
  *     maxWidth: 80,
  *     unit: 'imperial'
  * }));
+ *
+ * scale.updateUnit('metric');
  */
 class ScaleControl {
     _map: Map;
@@ -39,11 +41,6 @@ class ScaleControl {
 
     getDefaultPosition() {
         return 'bottom-left';
-    }
-
-    updateUnit(unit: string) {
-        this.options.unit = unit;
-        updateScale(this._map, this._container, this.options);
     }
 
     _onMove() {
@@ -64,6 +61,16 @@ class ScaleControl {
         DOM.remove(this._container);
         this._map.off('move', this._onMove);
         this._map = (undefined: any);
+    }
+
+    /**
+     * Update the scale's unit of the distance
+     *
+     * @param {String} unit
+     */
+    updateUnit(unit: string) {
+        this.options.unit = unit;
+        updateScale(this._map, this._container, this.options);
     }
 }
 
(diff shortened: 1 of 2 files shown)


test/unit/ui/control/scale.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+const test = require('mapbox-gl-js-test').test;
+const window = require('../../../../src/util/window');
+const Map = require('../../../../src/ui/map');
+const config = require('../../../../src/util/config');
+const ScaleControl = require('../../../../src/ui/control/scale_control');
+
+function createMap() {
+    const container = window.document.createElement('div');
+    config.ACCESS_TOKEN = 'pk.123';
```
anandthakker: The dummy access token shouldn't be necessary, and nor should the `owner` or `id` properties below.  (I think it exists in the `AttributionControl` tests because that specific control needs it.)

test/unit/ui/control/scale.test.js
```diff
@@ -0,0 +1,40 @@
...
+    t.end();
+});
+
+test('ScaleControl appears in the position specified by the position option', (t) => {
+    const map = createMap();
+    map.addControl(new ScaleControl(), 'top-left');
+
+    t.equal(map.getContainer().querySelectorAll('.mapboxgl-ctrl-top-left .mapboxgl-ctrl-scale').length, 1);
+    t.end();
+});
```
anandthakker: 🙇  thank you for adding these tests!  Could you also add one that tests the new `updateUnit()` method?

anandthakker:
> an open question, should the naming of this method be setUnit to be consistent with the GL JS API which already extensivly uses get and set?

👍 yep, agreed @andrewharvey  -- I meant to mention that in the review, in fact!

test/unit/ui/control/scale.test.js
ryanhamley: Hi @anandthakker I will work on this in the next day or so. Thanks for the feedback.

ryanhamley:
I'll update the name of the method to `setUnit` @anandthakker @andrewharvey

ryanhamley:
@anandthakker @andrewharvey I pushed the requested changes.

src/ui/control/scale_control.js
```diff
@@ -54,6 +63,16 @@ class ScaleControl {
         this._map.off('move', this._onMove);
         this._map = (undefined: any);
     }
+
+    /**
+     * Set the scale's unit of the distance
+     *
+     * @param {String} unit
```
andrewharvey: `String` is causing the CI to fail, I believe it should be `string`.

andrewharvey: Also I think it would be helpful to include the documentation for `unit`, listing out what the supported values are.

ryanhamley:
I fixed the param documentation and it looks like the build ci is passing now @andrewharvey

ryanhamley:
@anandthakker i added the requested test and everything is building now

test/unit/ui/control/scale.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+test('ScaleControl should change unit of distance after calling setUnit', (t) => {
+    const map = createMap();
+    const scale = new ScaleControl();
+    map.addControl(scale);
+
+    t.equal(scale.options.unit, 'metric');
+    scale.setUnit('imperial');
+    t.equal(scale.options.unit, 'imperial');
```
anandthakker: The fact that the unit is stored in `scale.options.unit` is an implementation detail that could be changed in a future refactor.  Let's update this test so that it wouldn't be affected by such a change by having these assertions actually check the content of the rendered scale control, e.g. maybe by searching for an appropriate string within `map.getContainer().querySelectorAll('.mapboxgl-ctrl-top-left .mapboxgl-ctrl-scale').innerHTML`

ryanhamley:
I pushed a change to the scale test that reads the unit from the scale's HTML @anandthakker Does that look good? I'm pretty sure it's not best practice to abstract getting the unit from the map into a function but I wasn't sure.

test/unit/ui/control/scale.test.js
```diff
@@ -40,8 +40,13 @@ test('ScaleControl should change unit of distance after calling setUnit', (t) =>
     const scale = new ScaleControl();
     map.addControl(scale);
 
-    t.equal(scale.options.unit, 'metric');
+    let unit = map.getContainer().querySelectorAll('.mapboxgl-ctrl-bottom-left .mapboxgl-ctrl-scale')[0].innerHTML.match(/[a-zA-Z]+|[0-9]+/g)[1];
+
+    t.equal(unit, 'km');
```
anandthakker: @ryanhamley small nitpick: let's simplify this just a tad to:
```js
let contents = map.getContainer().querySelector('.mapboxgl-ctrl-bottom-left .mapboxgl-ctrl-scale').innerText; // (not 100% sure jsdom supports innerText, but worth a shot
t.match(contents, /\bkm\b/);
```

ryanhamley:
@anandthakker I pushed a simpler version of the test. it doesn't appear that jsdom has innerText so I used innerHTML and the word boundary character in the regex doesn't distinguish between letters and numbers so I had to remove that from the regex to match the correct characters.

anandthakker:
--> #6274

PR #6147, opened 2018-02-13
fix duplicate layer test, enforce layer id uniqueness

### LINKED ISSUE #5989: addLayer doesn't enforce id uniqueness
**mapbox-gl-js version**: 0.43.0beta

### Steps to Trigger Behavior

 1. `map.addLayer({id: 'foo', ...})`
 2.`map.addLayer({id: 'foo', ...})`

### Expected Behavior

An error.

### Actual Behavior

The extra layer is added, which can cause some very surprising behaviour. Notably, calls like `removeLayer()` and some others don't work properly, causing odd error messages.

mollymerp:
fix #5989 

the test that was supposed to verify this was erroneously passing due to an extra `t.end()` and validation in validate_layer doesn't work because we're passing in  `arrayIndex = -1` so it never enters the loop to check duplicate ids. 

https://github.com/mapbox/mapbox-gl-js/blob/e077b780a525aa5c39bf9d1489f9bd1f458856ba/src/style-spec/validate/validate_layer.js#L27-L32

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -521,6 +521,11 @@ class Style extends Evented {
 
         const id = layerObject.id;
 
+        if (this.getLayer(id)) {
+            this.fire('error', {error: new Error(`Layer with id "${id}" already exists on this map`)});
+            return;
+        }
+
         if (typeof layerObject.source === 'object') {
             this.addSource(id, layerObject.source);
             layerObject = util.clone(layerObject);
--- a/test/unit/style/style.test.js
+++ b/test/unit/style/style.test.js
@@ -992,20 +992,19 @@ test('Style#addLayer', (t) => {
     });
 
     t.test('emits error on duplicates', (t) => {
+        t.stub(console, 'error');
         const style = new Style(new StubMap());
         style.loadJSON(createStyleJSON());
         const layer = {id: 'background', type: 'background'};
 
         style.on('error', (e) => {
-            t.deepEqual(e.layer, {id: 'background'});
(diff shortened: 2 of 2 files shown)


src/style/style.js
```diff
@@ -521,6 +521,11 @@ class Style extends Evented {
 
         const id = layerObject.id;
 
+        if (this.getLayer(id)) {
+            this.fire('error', {error: new Error(`Layer with id "${id}" already exists on this map`)});
```
asheemmamoowala: `Map.addSource`  will throw an error when attempting to add a source with a duplicate ID. Is there a reason to fire an error event instead of throwing for `addLayer` ?

jfirebaugh: Firing an event is correct; see #3879.

mollymerp: We made a decision awhile back to emit errors across the board instead of throwing https://github.com/mapbox/mapbox-gl-js/issues/3879#issue-198365050

mollymerp: whoops didn't reload and see john's comment 😬

test/unit/style/style.test.js
```diff
@@ -992,20 +992,19 @@ test('Style#addLayer', (t) => {
     });
 
     t.test('emits error on duplicates', (t) => {
+        t.stub(console, 'error');
```
jfirebaugh: Is this necessary?

mollymerp: hmm I guess not! left over from earlier debugging I think. removed!

PR #6149, opened 2018-02-13
add sdk support tables for hillshade layer properties

mollymerp:
does it make sense to add the ios and android versions now bc they're still in beta and not fully released? 

cc @tristen @jmkiley @langsmith 

## Launch Checklist

 - [ ] post benchmark scores


jfirebaugh:
I believe consensus is to include entries only when a final release is available.

PR #6158, opened 2018-02-14
Parallelize Circle CI execution

jfirebaugh:
Use Circle CI workflows to parallelize most steps of the build:

image

This looks like it will reduce build time from 15-16 minutes to 6-7 minutes, with further improvements possible if we somehow parallelize the test-unit and test-render steps.

TODO:

* [x] Ensure deploy step works (push v0.0.0 branch)
* [x] Restore/refactor benchmark.js deployment
* [x] Restore/abandon coveralls use (I vote abandon; AFAIK it's ignored by all of us)


mourner:
@jfirebaugh 👍 on abandoning Coveralls, tracking coverage on every commit is not very useful at this point. We can still get a coverage report manually when needed.

jfirebaugh:
* [Branch build](https://circleci.com/workflow-run/3ec2b3d7-4bb6-4072-a9d2-f0865fd082a40)
* [Release tag build](https://circleci.com/workflow-run/be503707-f9f7-4c76-b0c3-6c3648bdf5f1)

Will update required status checks once this is merged to master.

circle.yml
```diff
@@ -0,0 +1,208 @@
...
+            - test-render
+            - test-query
+            - test-expressions
+          filters:
+            tags:
+              only: /v[0-9]+.[0-9]+.[0-9]+(-.+)?/
+            branches:
+              only:
+                - master
+                - /release-.*/
```
anandthakker: Why do we need this on `release-` branches if we already have it for version tags?

circle.yml
```diff
@@ -0,0 +1,208 @@
...
+          at: .
+      - restore_cache:
+          keys:
+            - v1-lint-{{ .Branch }}
+            - v1-lint
+      - run: yarn run lint
+      - run: yarn run lint-docs
+      - run: yarn run lint-css
+      - save_cache:
+          key: v1-lint-{{ .Branch }}-{{ .Revision }}
```
anandthakker: Looks like we never restore a cache with this key.  Should this match the `restore_cache` key above?

circle.yml
jfirebaugh: So that when you're benchmarking local changes that will be targeted to a release branch, you can compare `HEAD` to the current state of the target branch.

circle.yml
jfirebaugh: No; restoration uses prefix matching, with the most recent winning in case of ties, and caches are immutable once written.

anandthakker: Oh, that's nice!

PR #6162, opened 2018-02-14
Improve heatmap render tests; support multiple expected images

mourner:
Some platforms on GL native support half-float-textures-based rendering of heatmaps, for which we provide a separate expected image alongside ubyte-textures-based.

The test harness then compares with each expected image and picks the one with the least amount of difference. 

This PR will allow us to fully cover https://github.com/mapbox/mapbox-gl-native/pull/11046 with render tests on different platforms.


test/integration/lib/render.js
```diff
@@ -125,26 +140,27 @@ exports.run = function (implementation, ignores, render) {
 
             png.data = data;
 
-            if (process.env.UPDATE) {
+            let writeExpected = !!process.env.UPDATE;
+            try {
+                stats = fs.statSync(expected, fs.R_OK | fs.W_OK);
+                if (!stats.isFile()) throw new Error();
```
kkaefer: Let's add an error message here

test/integration/lib/render.js
```diff
@@ -125,26 +140,27 @@ exports.run = function (implementation, ignores, render) {
...
+
+            } catch (e) { // no expected.png, create it
+                writeExpected = true;
+            }
```
anandthakker: I think we should keep the requirement for an explicitly setting `UPDATE` rather than always writing `expected.png` when it's missing

test/integration/lib/render.js
```diff
@@ -5,24 +5,39 @@ const path = require('path');
 const PNG = require('pngjs').PNG;
 const harness = require('./harness');
 const pixelmatch = require('pixelmatch');
+const glob = require('glob');
 
-function compare(path1, path2, diffPath, callback) {
+function compare(actualPath, expectedPaths, diffPath, callback) {
```
anandthakker: Let's add a brief comment about how we take the best comparison (and why)

test/integration/lib/render.js
mourner: I didn't change the logic, just deduplicated some code here. You can see that it creates expected.png when its missing if you don't set UPDATE.

anandthakker: 🙈  comment withdrawn

test/integration/lib/render.js
```diff
@@ -125,26 +140,27 @@ exports.run = function (implementation, ignores, render) {
...
             png.data = data;
 
-            if (process.env.UPDATE) {
+            let writeExpected = !!process.env.UPDATE;
+            try {
+                stats = fs.statSync(expected, fs.R_OK | fs.W_OK);
+                if (!stats.isFile()) throw new Error();
+
+            } catch (e) { // no expected.png, create it
+                writeExpected = true;
```
kkaefer: Can we add a message to the console when we're updating this to at least give the user some indication of what is happening

mourner:
@kkaefer @anandthakker after renaming `expected.png` to `expected-ubyte.png`, the "create expected if missing" logic broke the tests, so now I agree with @anandthakker that we should simply remove that logic and only create it explicitly on `UPDATE=true`. Added that in the latest commit.

mourner:
@anandthakker on a second thought, maybe we should revert the renaming — we're mostly updating render tests from GL JS side, and running tests with `UPDATE=1` for any matter will create a bunch of duplicate `expected.png` for heatmap tests which have to be deleted, which is very inconvenient.

anandthakker:
Ah, yeah I guess that makes sense @mourner -- in that case maybe we should make `expected.png` be the half-float version, since it's the more "correct" one?

mourner:
@anandthakker no, I think we should leave ubyte ones there because that's what GL JS will generate on UPDATE=true with headless-gl.

PR #6164, opened 2018-02-14
do not hide icons if text is an empty string

### LINKED ISSUE #6160: Symbol layer items disappear when text-field is an empty string

### Steps to Trigger Behavior

 1. Create a symbol layer with icon-image and text-field.
 2. Conditionally set text-field to empty string.

### Expected Behavior

All the symbols are visible.

### Actual Behavior

Symbols that have text-field set as empty string disappear on close enough zoom-levels.

Simple example which reproduces the issue: https://jsbin.com/weraqecesa/edit?html,output

See for example the area around Muir Beach where we have one symbol without text which disappears when zooming in. If we change the symbol to have non-empty text-field it stays visible.

ansis:
fix #6160

Empty text strings do not have any collision boxes so the collision index check was never done and the text was not counted as "placed". This fixes that by making "placed" the default when the feature has text.

## Alternative fix
Another possibility would be to create zero-sized collision boxes for empty strings. If we did this, should these zero-size boxes collide with other boxes if they are contained within them?

## Launch Checklist

 - [N/A] post benchmark scores


mb12:
@ansis Can you please clarify when should one use 'text-optional' and 'icon-optional' layout properties? Shouldn't this require setting text-optional to true in the corresponding style?

dagjomar:
Would this also fix the inverse case, "do not hide text if icon is an empty string" ?

ChrisLoer:
@dagjomar Mostly I think yes... from the point of view of the code, there's nothing special about an `icon-image` that evaluates to an empty string: it will be just the same as any other `icon-image` that's not available in the sprite sheet. Here's the relevant logic:

https://github.com/mapbox/mapbox-gl-js/blob/25f9afa30bed8d425c66c9ab92df8092cdbd5cdb/src/symbol/symbol_layout.js#L124-L147

Along with:

https://github.com/mapbox/mapbox-gl-js/blob/25f9afa30bed8d425c66c9ab92df8092cdbd5cdb/src/symbol/placement.js#L162

So what that means is something like: "If the feature has an icon, but that icon's image is not in the image map, we won't create a shapedIcon and we won't add a symbol instance for just that icon. But if the same feature also has text, we'll bring the icon along for the ride, and `placeIcon` will default to true even though there's no collision feature for the empty icon"

azadeh18:
hi. i have an apposite problem. when i add a text field the symbols disappear. i searched a lot and i used a lot of suggestions. but none of them didn't work for me. please help me. that's my code:
`
 private void addSymbolToMap(String name, LatLng point) {
  mapStyle.addImage(name, getResources().getDrawable(R.drawable.mapbox_marker_icon_default));
        SymbolManager sampleSymbolManager = new SymbolManager(mapView, map, mapStyle);
   sampleSymbolManager.setIconAllowOverlap(true);
        sampleSymbolManager.setTextAllowOverlap(true);
        sampleSymbolManager.setIconRotationAlignment(ICON_ROTATION_ALIGNMENT_VIEWPORT);
 SymbolOptions sampleSymbolOptions = new SymbolOptions()
                .withLatLng(point)
                .withIconImage(name)
                .withIconSize(1.5f);
                .withTextField("test")
                .withTextFont(new String[] { "Arial Unicode MS Regular"});
 Symbol sampleSymbol = sampleSymbolManager.create(sampleSymbolOptions);
}

`

PR #6166, opened 2018-02-14
always hide duplicate labels instantly

ansis:
Duplicate labels were not always hidden instantly, resulting in the same label being rendered twice by different tiles. This fixes that.

port of https://github.com/mapbox/mapbox-gl-native/pull/11207

## Launch Checklist

 - [N/A] post benchmark scores

### DIFF
--- a/src/symbol/placement.js
+++ b/src/symbol/placement.js
@@ -284,9 +284,6 @@ class Placement {
                 opacityState = duplicateOpacityState;
             } else if (!opacityState) {
                 opacityState = defaultOpacityState;
-            }
-
-            if (!this.opacities[symbolInstance.crossTileID]) {
                 // store the state so that future placements use it as a starting point
                 this.opacities[symbolInstance.crossTileID] = opacityState;
             }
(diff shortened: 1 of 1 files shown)


src/symbol/placement.js
```diff
@@ -279,12 +280,15 @@ class Placement {
...
 
             let opacityState = this.opacities[symbolInstance.crossTileID];
-            if (!opacityState) {
+            if (isDuplicate) {
+                opacityState = duplicateOpacityState;
+            } else if (!opacityState) {
                 opacityState = defaultOpacityState;
+            }
+
+            if (!this.opacities[symbolInstance.crossTileID]) {
```
ChrisLoer: Can't we just move this into the `else if (!opacityState)` clause above and avoid a second lookup? Because it shouldn't be possible for us to have gone down the `isDuplicate` branch without having first made an entry for this `this.opacities[crossTileID]` (and if it were possible, it'd be a problem, because we'd be making our initial entry into `duplicateOpacityState` instead of `defaultOpacityState`).

ansis: Yep! fixed

PR #6167, opened 2018-02-14
Handle world wrapping better in popup examples

### LINKED ISSUE #6118: Wrong popup position when zoom out mapbox
**mapbox-gl-js version**: 0.44.0

### Steps to Trigger Behavior

 1.  Go to https://www.mapbox.com/mapbox-gl-js/example/popup-on-hover/
 2. Zoom out all the way until can't zoom out any more. Now we can see 2 points that will popup when hover.
 3. Hover the left point, popup position is correct. Hover the right point, popup still pop on the left point

### Expected Behavior
- Popup on correct position
### Actual Behavior
- Hover to the right but popup pop on the left:
untitled

jfirebaugh:
Fixes #6118.

### DIFF
--- a/docs/pages/example/popup-on-click.html
+++ b/docs/pages/example/popup-on-click.html
@@ -125,9 +125,19 @@ map.on('load', function () {
     // When a click event occurs on a feature in the places layer, open a popup at the
     // location of the feature, with description HTML from its properties.
     map.on('click', 'places', function (e) {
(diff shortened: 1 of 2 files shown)


jfirebaugh:
That's the "if the coordinate is off screen, wrap it" behavior from `Popup` kicking in. If the popup were anchored at the coordinates of the copy on the right, it would be partially off-screen. I think it's acceptable behavior.

Issue #6169, opened 2018-02-15
Don't pass RTL text plugin to workers as a blob URL


ChrisLoer:
`loadWorkerSource` supports loading worker sources with blob URLs so that we can construct the source without requiring any extra depedency. In the case of `setRTLTextPlugin`, we're always loading an external dependency, so that's not an issue. I'm not sure why I implemented it to use a blob, but my guess is that I was just figuring out how everything worked and I followed the example of `loadWorkerSource`.

Some CSP policies don't allow loading script from blob sources. Since this isn't a strict requirement for us, we should get rid of it and just run `importScripts` directly on the source URL.

/cc @anandthakker @jfirebaugh

PR #6170, opened 2018-02-15
Don't create RTL text plugin blob on foreground.

ChrisLoer:
Avoids CSP script-src constraints.
Fixes issue #6169.

Now that I built and tested this, I remember the motivation for creating a blob on the foreground -- we had hoped that the browser could coalesce the duplicate requests for the same plugin from the workers, but it looks like on the first map load (with the cache clear), all four workers download the blob in parallel. So that's 800KB to load instead of 200KB.

Does it make sense to turn this into a configurable "CSP friendly" option? What did we do in the worker source case?

I added a `setRTLTextPlugin` call to `csp.html` to test this out, but it currently points to the eval-free "0.1.2" version which I haven't published yet.

/cc @jfirebaugh @anandthakker

### DIFF
--- a/src/source/rtl_text_plugin.js
+++ b/src/source/rtl_text_plugin.js
@@ -1,51 +1,38 @@
 // @flow
-
-const ajax = require('../util/ajax');
 const Evented = require('../util/evented');
-const window = require('../util/window');
 
 let pluginRequested = false;
-let pluginBlobURL = null;
+let pluginURL = null;
 
 module.exports.evented = new Evented();
 
 type ErrorCallback = (error: Error) => void;
 
 module.exports.registerForPluginAvailability = function(
-    callback: (args: {pluginBlobURL: string, errorCallback: ErrorCallback}) => void
+    callback: (args: {pluginURL: string, errorCallback: ErrorCallback}) => void
 ) {
-    if (pluginBlobURL) {
-        callback({ pluginBlobURL: pluginBlobURL, errorCallback: module.exports.errorCallback});
+    if (pluginURL) {
+        callback({ pluginURL: pluginURL, errorCallback: module.exports.errorCallback});
     } else {
         module.exports.evented.once('pluginAvailable', callback);
     }
     return callback;
 };
 
-// Exposed so it can be stubbed out by tests
-module.exports.createBlobURL = function(response: Object) {
-    return window.URL.createObjectURL(new window.Blob([response.data], {type: "text/javascript"}));
-};
 // Only exposed for tests
 module.exports.clearRTLTextPlugin = function() {
     pluginRequested = false;
-    pluginBlobURL = null;
+    pluginURL = null;
 };
 
-module.exports.setRTLTextPlugin = function(pluginURL: string, callback: ErrorCallback) {
+module.exports.setRTLTextPlugin = function(url: string, callback: ErrorCallback) {
     if (pluginRequested) {
         throw new Error('setRTLTextPlugin cannot be called multiple times.');
     }
     pluginRequested = true;
+    pluginURL = url;
     module.exports.errorCallback = callback;
-    ajax.getArrayBuffer({ url: pluginURL }, (err, response) => {
-        if (err) {
-            callback(err);
-        } else if (response) {
-            pluginBlobURL = module.exports.createBlobURL(response);
-            module.exports.evented.fire('pluginAvailable', { pluginBlobURL: pluginBlobURL, errorCallback: callback });
-        }
-    });
+    module.exports.evented.fire('pluginAvailable', { pluginURL: pluginURL, errorCallback: callback });
 };
 
 module.exports.applyArabicShaping = (null: ?Function);
--- a/debug/csp.html
+++ b/debug/csp.html
@@ -27,6 +27,8 @@ var map = window.map = new mapboxgl.Map({
     hash: true
 });
 
+mapboxgl.setRTLTextPlugin('https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.1.2/mapbox-gl-rtl-text.js');
+
 </script>
 </body>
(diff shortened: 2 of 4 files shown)


anandthakker:
> Does it make sense to turn this into a configurable "CSP friendly" option? What did we do in the worker source case?

An alternative would be to ensure it's in the browser cache by requesting it on the main thread, waiting for a response, and only then sending the `loadPlugin` message to all the workers.

ChrisLoer:
@anandthakker you're a genius. I'll try that.

anandthakker:
Ah, bummer. I don’t suppose you have “disable cache” on in dev tools?  I usually do..

ChrisLoer:
I didn't have "disable cache" set in Firefox, and I'm still seeing the same "load five times" behavior. I _did_ in Chrome 😳, and after I turned that off and cleared my cache I saw the behavior we hoped for. Also Safari behaved as we'd hope.

@1ec5 should we go ahead and make Firefox users load a more expensive version of the map? 😜

anandthakker:
@ChrisLoer when i look at this simple test in Firefox, I get the expected caching behavior:

```html
<script src="main.js?t=0"></script>
<script src="main.js?t=0"></script>
<script>
    new Worker('main.js?t=0');
</script>
```

screen shot 2018-02-15 at 11 04 31 am

anandthakker:
Yep, same thing using `xhr` and `importScripts` just like GL JS does https://gist.github.com/anandthakker/352001a6482d43d20aea5807ef7aaf4f

screen shot 2018-02-15 at 11 24 19 am

ChrisLoer:
Hmm, I must be doing something wrong, but haven't figured it out yet. Let me see if I get the same behavior with your gist... Here's what I'm seeing after clearing the cache in Firefox 59.0b9 (look for mapbox-gl-rtl-text.js):

screenshot 2018-02-15 10 06 19

ChrisLoer:
I see the same thing as you loading from localhost, but take your test setup and replace the local URL with something pointing to `api.mapbox.com`, eg:

```
const url = `https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.1.2/mapbox-gl-rtl-text.js?t=${Math.random()}`;
```

screenshot 2018-02-15 10 21 40

I notice the inferred type is `octet-stream` instead of `js`, not sure if that would affect caching behavior...

anandthakker:
Hmm, I'm still getting the right caching with that URL:

screen shot 2018-02-15 at 1 27 46 pm

ChrisLoer:
I get the same behavior loading `mapbox-gl.js` from `api.tiles.mapbox.com`, which is correctly tagged with content-type 'application/js':

screenshot 2018-02-15 10 46 20

Not sure if there's some other reason I should make sure to update our hosted plugins to have a content type tagged on them, but doesn't seem like it'll help here...

jfirebaugh:
> Not sure if there's some other reason I should make sure to update our hosted plugins to have a content type tagged on them

Yes, pass `--content-type application/javascript` on the `aws s3 cp` command.

ChrisLoer:
🤦‍♂️ This is pretty embarassing, but after realizing I had "disable cache" enabled in Chrome devtools, I never revisited my initial assumption that the workers couldn't intelligently coordinate their cache use. Just reverting the "try loading on the foreground" first commit gives me the behavior we'd hope for (one load, three from cache) on Chrome/Firefox/Safari.

I'm not sure I have a total explanation for everything I saw, but I think I figured out a couple things:

 - At least in the tests _I_ was running against localhost, I was using `st --no-cache`, and it looks like the `cache-control:none` caused the browser to attach `If-Modified-Since` headers to subsequent requests. The server would then respond with a 304, and _then_ we'd look in the cache and find the file that we had loaded on the foreground.
- In the tests pointing to `api.mapbox.com` with a generous cache-control max-age, we wouldn't include the `If-Modified-Since` header, and it looks like the foreground and the workers effectively worked against two different caches (even in @anandthakker's simplified "caching is working right" screengrab (https://github.com/mapbox/mapbox-gl-js/pull/6170#issuecomment-366018225) note that the file got downloaded once on the foreground and once for the workers).
- The behavior I saw where increasing the file size seemed to increase the chance of cache misses only applied to the first case where we were getting 304s... so something about that pathway allowed for some sort of non-instant synchronization between two caches 🤷‍♂️ ?

At any rate, now that we're no longer trying to do anything fancy to trick browser caching, and I'm seeing the behavior I'd hoped for in the first place, I feel OK about my current level of ignorance.

src/source/rtl_text_plugin.js
```diff
@@ -1,51 +1,38 @@
...
     module.exports.errorCallback = callback;
-    ajax.getArrayBuffer({ url: pluginURL }, (err, response) => {
-        if (err) {
-            callback(err);
-        } else if (response) {
-            pluginBlobURL = module.exports.createBlobURL(response);
-            module.exports.evented.fire('pluginAvailable', { pluginBlobURL: pluginBlobURL, errorCallback: callback });
-        }
-    });
+    module.exports.evented.fire('pluginAvailable', { pluginURL: pluginURL, errorCallback: callback });
```
anandthakker: Since we're no longer requesting the URL here, a malformed URL that would previously have resulted in an error callback would now just get passed along to the worker (as seen int he "some bogus url" unit test below).  Now that we don't have the AJAX request effectively validating the url for us, should we do so directly here?

ChrisLoer: We do still feed errors from the workers to the foreground, but the caller needs to pass in an error callback and while that's documented we don't actually pass a callback in our example code -- maybe we should just change the example to include a simple logging error callback? The actual loading error will hopefully be more informative than whatever we come up with from URL validation.

anandthakker: Ah, I see -- yeah, that seems fine

Issue #6106, opened 2018-02-07
Simultaneous map zoom and drag throws TypeError


dm261395:
**mapbox-gl-js version**:
v0.44.0

### Steps to Trigger Behavior

 1. On the Mapbox GL JS simple map example at https://www.mapbox.com/mapbox-gl-js/example/simple-map/, start zooming in or out on the map via the mouse wheel scroll.
 2. Before finishing zooming, click and drag the map in some direction. 
 3. Type error is thrown (Chrome console output): 

> Uncaught TypeError: Cannot read property 'x' of undefined

GIF demonstration:
![bug](https://user-images.githubusercontent.com/34164181/35911484-d61969d8-0bf1-11e8-85cc-2433273a8209.gif)

I am on macOS High Sierra Version 10.13.3. I have encountered this error on Chrome Version 64.0.3282.140, Firefox Quantum v58.0 and Safari Version 11.0.3 since the mapbox-gl upgrade v0.44.0.

mourner:
Fixed in #6113

PR #6171, opened 2018-02-15
Don't use _onDragFinished as a "finish" callback for _startAnimation

jfirebaugh:
I believe that this is a better fix for #6106 than #6113. #6113 doesn't address the related issues that an "interruption" by another handler would trigger inertial panning and a `dragend` event, despite the drag remaining in progress until the mouse or touch is released.

@anandthakker, if you agree, I will port this approach to `DragRotateHandler` as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "glob": "^7.0.3",
     "in-publish": "^2.0.0",
     "is-builtin-module": "^1.0.0",
-    "jsdom": "^11.5.1",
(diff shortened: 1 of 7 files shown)


anandthakker:
Yeah, you’re totally right—this is the right fix.

jfirebaugh:
@anandthakker Want to 👀  the equivalent changes for `DragRotateHandler`?

PR #6175, opened 2018-02-15
add originalEvent on user-initiated scroll zoomend and moveend #1975

stepankuzmin:
`originalEvent` is missing on `zoomend` and `moveend` user-initiated scroll events. This PR adds `_lastWheelEvent` as `originalEvent` to `zoomend` and `moveend` events.

See #1975 for more info.

## Launch Checklist

### DIFF
--- a/src/ui/handler/scroll_zoom.js
+++ b/src/ui/handler/scroll_zoom.js
@@ -251,8 +251,8 @@ class ScrollZoomHandler {
         if (!this.isActive()) return;
         this._active = false;
         this._finishTimeout = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


stepankuzmin:
Thanks for the reply! The main point is to detect if an `*end` event was initiated by a user. Consider using [Mapbox GL JS in a Reactive application](https://blog.mapbox.com/mapbox-gl-js-in-a-reactive-application-e08eecf0221b) as an example. Using the set of `*end` event handlers, in that case, seems more logical than using single `move` event handler with debouncing.

PR #6181, opened 2018-02-16
Native tests for expression serialization

ChrisLoer:
Native test implementation for https://github.com/mapbox/mapbox-gl-native/pull/11156:

 - Round-trip all expression tests through a serialization and check that they return the same value after their second compilation/evaluation.
 - Bootstrap an "expected serialization" for each expression test. I've gone through this whole list at least lightly once, but not very exhaustively. For making sure these initial values are right, I'm depending on doing a spot-check for each expression type, and especially relying on the round-trip tests.

The order of the JSON output for the expression tests changed when I ran the update -- I'm not sure why, but the new order looks more rational to me (alphabetical within JSON objects).

The GL JS tests themselves shouldn't change.

The most interesting changes are down at the bottom in `lib/expression.js`.

/cc @anandthakker @1ec5


test/integration/expression-tests/array/basic/test.json
```diff
@@ -2,12 +2,13 @@
...
     "compiled": {
       "result": "success",
-      "isZoomConstant": true,
       "isFeatureConstant": true,
+      "isZoomConstant": true,
       "type": "array"
-    }
+    },
+    "outputs": [[1, 2, 3]],
+    "serialized": ["literal", [1, 2, 3]]
```
anandthakker: Huh, I'm surprised to see the outer `"array"` assertion missing here -- I don't see any reason over at https://github.com/mapbox/mapbox-gl-native/pull/11156/files#diff-14232703a3132993aadc29177a139a62 why this would happen (and I don't think it would happen on the parsing side either)

test/integration/lib/expression.js
```diff
@@ -70,19 +70,35 @@ exports.run = function (implementation, options, runExpressionTest) {
 
             if (process.env.UPDATE) {
                 fixture.expected = result;
+                // We save the expected serialization, but the round-tripped compilation results
```
anandthakker: This would strip `expected.serialized` if we ran the update from GL JS.  Since we don't have `Expression.serialize()` here, maybe something like:

```js
  fixture.expected = Object.assign({
    // if we're updating from GL JS, where `result.serialized` will not exist, just copy the existing expected value, or default to expected the serialized form to be identical to the input expression
    serialized: (fixture.expected && fixture.expected.serialized) ? fixture.expected.serialized : fixture.expression
  }, result);
```

test/integration/lib/expression.js
```diff
@@ -70,19 +70,35 @@ exports.run = function (implementation, options, runExpressionTest) {
...
+                // and outputs should be identical, so no need to save them.
+                delete fixture.expected.recompiled;
+                delete fixture.expected.roundTripOutputs;
```
anandthakker: These properties get added by the native test runner but not the JS one, which might make this a bit confusing to maintain later... instead of copying `result` wholesale, let's opt-in to the specific properties we want:
```js
fixture.expected = {
    compiled: result.compiled,
    outputs: result.outputs,
    serialized: result.serialized !== undefined ? result.serialized :
        (fixture.expected && fixture.expected.serialized) ? fixture.expected.serialized :
        fixture.expression
}
```

test/integration/expression-tests/array/basic/test.json
anandthakker: Oh wait, this is just due to constant folding. 👍

test/integration/lib/expression.js
ChrisLoer: Of course! Good catch...

test/integration/lib/expression.js
ChrisLoer: :+1:

ChrisLoer:
OK, the update mode is now consistent between the two versions. Trying to enforce decimal precision rules with floating point numbers gets pretty weird.

PR #6183, opened 2018-02-16
Don't reuse Map#{moving,zooming} state

### LINKED ISSUE #6173: Incorrect result for isMoving() if drag/rotate and scroll interleave
`DragPanHandler`, `DragRotateHandler`, and `ScrollZoomHandler` each set and unset `map.moving` independently. But zooms and drags can interleave:

1. `dragstart`
2. `zoomstart`
3. `zoomend`
4. `dragend`

In this scenario, `map.isMoving()` will return `false` between 3 and 4. It should return `true`.

jfirebaugh:
Instead, implement isMoving and isZooming using the independent state that the relevant handlers already track.

Fixes #6173.

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -584,7 +584,7 @@ class Map extends Camera {
      * Returns true if the map is panning, zooming, rotating, or pitching due to a camera animation or user gesture.
      */
     isMoving(): boolean {
-        return this.moving ||
+        return this._moving ||
             this.dragPan.isActive() ||
             this.dragRotate.isActive() ||
             this.scrollZoom.isActive();
@@ -594,8 +594,16 @@ class Map extends Camera {
      * Returns true if the map is zooming due to a camera animation or user gesture.
      */
     isZooming(): boolean {
-        return this.zooming ||
+        return this._zooming ||
             this.scrollZoom.isActive();
+    }
+
+    /**
+     * Returns true if the map is rotating due to a camera animation or user gesture.
+     */
+    isRotating(): boolean {
+        return this._rotating ||
+            this.dragRotate.isActive();
     }
 
     /**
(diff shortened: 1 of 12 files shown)


anandthakker:
There's at least one place where we're reading the `map.zooming`, etc. properties directly: https://github.com/mapbox/mapbox-gl-js/blob/38c7250da7b5153c531d0b5b8ff4a2f07f9ab863/src/ui/map.js#L1529-L1530

As a way to sweep for those and also since `map.{zooming,rotating,moving}` should be behind getters anyway, should we rename these to `map._{zooming,rotating,moving}`?

jfirebaugh:
@anandthakker Ready for second-pass review here.

test/unit/ui/map/isMoving.test.js
```diff
@@ -0,0 +1,145 @@
...
+    map._updateCamera();
+
+    now += 400;
+    map._updateCamera();
+});
+
+test('Map#isMoving returns true when drag panning and scroll zooming interleave', (t) => {
+    const map = createMap();
+
+    map.on('dragstart', () => {
```
anandthakker: Add a quick comment here that we're expecting the events to come in the order they're listed here?

src/ui/map.js
```diff
@@ -580,6 +580,32 @@ class Map extends Camera {
...
+            this.dragRotate.isActive() ||
+            this.scrollZoom.isActive();
+    }
+
+    /**
+     * Returns true if the map is zooming due to a camera animation or user gesture.
+     */
+    isZooming(): boolean {
+        return this._zooming ||
+            this.scrollZoom.isActive();
```
mollymerp: should `this.doubleClickZoom.isActive()` also be called here?

jfirebaugh: `DoubleClickZoomHandler` implements the zoom using `zoomTo`, which will set `_zooming`, so it isn't strictly necessary. (In fact I think `DoubleClickZoomHandler` should probably not have an `isActive()` method at all. Unlike drag panning/zooming, where there's an interval bookended by down/up events, double-clicking is an instantaneous action.)

jfirebaugh: `DoubleClickZoomHandler #isActive()` turns out to be used internally, so I'll abandon that idea for now.

Issue #6056, opened 2018-01-25
Remove `worker:source blob` CSP requirement (iframe sandbox `allow-same-origin` support)


ryanbaumann:
# Issue:

Running Mapbox GL JS in a sandboxed iframe with only the `allow-scripts` permission blocks GL JS from running on IE11, Edge, and Safari browsers.

EX: 
`<iframe src="https://www.mapbox.com/bites/00384" sandbox="allow-scripts"></iframe>`

# Minimal test case:
https://bl.ocks.org/ryanbaumann/c28a57618ecd60d23f59d24d91d016fd

If the map in the example above does not load tiles in a browser, that means that the  **allow-same-origin** sandbox permissions are required for the browser iframe sandbox.  In IE11 and Edge, the example above does not load for me.  

The root cause of the map not loading in the minimal example above using sandboxed iframes on IE11 and Edge _seems_ to be due to fetching same-origin blob URL requests at web worker boot time:

Error messages loading the minimal example above:

**On Edge:**
![](https://cl.ly/2x361Z2Y0X47/download/Image%202018-01-25%20at%208.01.59%20AM.png)

**On IE11:**
![](https://cl.ly/2E3I0T1c291v/download/Image%202018-01-25%20at%208.06.51%20AM.png)

# Desired Solution 

If a new build of GL JS that eliminates the `worker:src blob` CSP, loading the test page above should work on IE11, Edge, and Safari.

See https://github.com/mapbox/mapbox-gl-js/issues/5939 for a potential solution.

jfirebaugh:
Since this is so closely related to #6058, let's centralize discussion there.

Issue #855, opened 2014-12-01
Switch to 6to5


mourner:
https://github.com/mapbox/mapbox-gl-online/issues/558 — we should probably do this here too.

jfirebaugh:
Yeah, going to revert to es5. Tooling support for es6 in node is too premature.

jfirebaugh:
Also, to answer "why not io.js" -- it doesn't yet support the ES6 features we use (arrow functions and object literal extensions).

mourner:
Yes, sounds good.

jfirebaugh:
Recording for posterity the reasons to switch back:

* There are no node debuggers that understand source maps, so breakpoints etc. don't work correctly.
* Couldn't get istanbul to work under traceur or 6to5.

thejameskyle:
> Couldn't get istanbul to work under traceur or 6to5.

You can use [isparta](https://github.com/douglasduteil/isparta) as your instrumentor and it should work fine.

mourner:
@jfirebaugh just wondering, did you try https://github.com/evanw/node-source-map-support?

jfirebaugh:
Yes. It's only for exception stack traces though, doesn't help with debugging or code coverage.

PR #6196, opened 2018-02-20
Switch to ES modules and Rollup

### LINKED ISSUE #5939: Switch to ES Modules with Rollup
I've been looking into this today, and to sum up, I think we should do it. 

Up till now, the main blocker for this has been our desire to keep time-to-first-render low. According to research in #3153 by @jfirebaugh, the amount of code evaluated on the worker side is the biggest factor in that, so we've been assuming we need to automatically create slim worker bundles from the main bundle, stripping out all the unnecessary code. So far, this has only been possible with Browserify because it keeps all the code in a dependency tree which we can traverse at runtime. There's no way to do that in Rollup — see https://github.com/rollup/rollup/issues/1029.

Note that when we implemented slimming down the worker bundle (https://github.com/browserify/webworkify/pull/30), the overall effect on TTFR [was not that big](https://github.com/mapbox/mapbox-gl-js/issues/3153#issuecomment-245352316) — only ~200ms despite the bundle becoming ~2x smaller.

Today, I decided to take a small part of GL JS and convert it to Rollup to see how it would look — specifically, [the expression code](https://github.com/mapbox/mapbox-gl-js/tree/master/src/style-spec/expression). When uglified/gzipped, the ES modules based build is approximately 10% smaller than the browserify-based one. Additionally, I wrapped the whole code with a simple `console.time/timeEnd` — the browserify-based build executed in 5.5ms in Node but the rollup one in just 1.2ms. Then I also remembered [this article by Nolan Lawson](https://nolanlawson.com/2016/08/15/the-cost-of-small-modules/) that benchmarked the runtime cost of different bundlers (and showed Browserify having a huge hit compared to Rollup). 

All this made me realize that maybe we should have looked at Browserify itself as the culprit of long evaluation times, and not just the size of the bundle. What if the switch from Browserify to ES Modules / Rollup reduces TTFR so dramatically that we no longer have to worry about worker bundle size much? Then, we could just make a custom rollup prelude that would launch the whole bundle as workers without traversing the deps at runtime.

My hunch is that we should take the plunge with ES Modules. The only big drawback I see is that we'll have to complicate the unit test setup (possibly relying on something like `@std/esm`), but advantages are great:

- smaller bundle
- possibly better time-to-first-render
- modern code base without the awkward mix of ES-like flow type imports and CommonJS
- more strict compile-time checking of dependencies

What do you think?

cc @jfirebaugh @anandthakker

anandthakker:
Closes #5939 
Refs #6056 

This PR:
 - Refactors the codebase to use ES modules
 - Changes the bundler from browserify to [Rollup](https://rollupjs.org/guide/en#introduction)

**Note on building the Worker bundle blob:**
There's a substantial change here in our approach to constructing the `Blob` URL for the worker script, because `webworkify` depends on specific `browserify` implementation details.  The way we're doing it now:
 - Use rollup's "code splitting" feature to run the build against two entrypoints: `src/index.js` and `src/source/worker.js`.  The result is _three_ bundles: `rollup/build/index.js`, `rollup/build/worker.js`, and `rollup/build/chunk1.js`.  The first two correspond to the two entrypoints, and the third consists of all the code that is shared by the main thread and worker scripts.
 - Use a second pass to assemble these three chunks into a single bundle, with some webworkify-inspired wrapper code that, at runtime, stringifies the worker source and uses it to create our worker blob URL.

**Note on running our code in Node:**
In order to run our Flow-annotated ES module code in a Node environment, we have to use two `require` hooks.  Previously, we would do this for flow by doing `require('flow-remove-types/register')` in our common unit test module -- this allowed us to `require` flow-annotated files from tests, but did _not_ permit flow in the test files themselves.   Instead, we're now telling `node` to require `flow-remove-types` (and `@std/esm`, for modules) directly via command-line arguments; you can run `node` or `tap` with these args already set using the `build/run-{tap,node}` scripts.

### Benchmarks

[Full](https://bl.ocks.org/anonymous/raw/5f5a2922bca2d544393cb816afe818d1/)
[Re-run of Paint](https://bl.ocks.org/anonymous/raw/69b6b6fb6ca71886cdc4cc75362ea70b/) 

#### rollup-anandthakker:

screen shot 2018-03-07 at 3 43 55 pm

#### master

screen shot 2018-03-07 at 3 44 19 pm

### Minified bundle size:

```
-rw-r--r--  1 anand  staff   709K Mar  7 15:37 dist/mapbox-gl--master.js
-rw-r--r--  1 anand  staff   572K Mar  7 15:42 dist/mapbox-gl--rollup-anandthakker.js
```

```
-rw-r--r--  1 anand  staff   169K Mar  7 15:37 dist/mapbox-gl--master.js.gz
-rw-r--r--  1 anand  staff   148K Mar  7 15:42 dist/mapbox-gl-rollup-anandthakker.js.gz
```

### Todo

- [x] <del>Check whether "minifying" the style-spec (as in https://github.com/mapbox/mapbox-gl-js/blob/master/build/minifyify_style_spec.js) eliminates the [perf regression](https://github.com/mapbox/mapbox-gl-js/issues/5939#issuecomment-366744544) in the worker-gets-the-whole-bundle approach</del> NOPE.
- [x] Get code splitting to work
- [ ] Polish build
  - [x] Prod / dev variations
  - [ ] Ensure fast watch/rebuild cycle (upstream PR submitted: https://github.com/rollup/rollup/pull/2035)
  - [x] Sourcemaps 😨  
  - [x] style-spec package build
  - [x] benchmarks build
  - [x] npm scripts / CI
  - [x] debug pages build (mainly `access_token_generated.js`)
- [x] Determine the best way to execute es module code in Node environment and apply it to:
  - [x] Unit tests
  - [x] Integration tests
  - [x] Codegen scripts
- [x] Update docs site
- [x] Update lint rules
  - Use https://github.com/benmosher/eslint-plugin-import h/t @mourner 
- [x] Remove single module object export pattern (`const exported = {...}; export default exported;`) -- ticketed https://github.com/mapbox/mapbox-gl-js/issues/6293
- [x] Remove browserify, (+ unused browserify transforms, derequire)
- [x] Post benchmark results: [Full](https://bl.ocks.org/anonymous/raw/5f5a2922bca2d544393cb816afe818d1/) / [Re-run of Paint](https://bl.ocks.org/anonymous/raw/69b6b6fb6ca71886cdc4cc75362ea70b/) 
- [x] Post final time-to-first-render comparison


src/data/bucket.js
```diff
@@ -98,3 +98,9 @@ module.exports = {
         return output;
     }
 };
+
+export default exported;
+
+export const {
+    deserialize
+} = exported;
```
mourner: We should probably get rid of all instances of exporting object literals like this — I suppose this is a legacy require/codemod thing? Ideally we would always export symbols directly, and only export default for classes/functions. This makes tree-shaking more effective.

bench/benchmarks_generated-v0.43.0.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable */
-(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
+(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw (f.code="MODULE_NOT_FOUND", f)}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
```
mourner: Committed accidentally?

bench/unit/bench_transform.js
```diff
@@ -1,25 +0,0 @@
...
-    t.locationPoint(lnglat);
-
-}).add('pointLocation', () => {
-    const point = new Point(1000 * Math.random(), 1000 * Math.random());
-    t.pointLocation(point);
-
-}).on('cycle', (event) => {
-    console.log(String(event.target));
-
-}).run();
```
mourner: Should we remove those benchmarks as a part of this PR?

build/generate-struct-arrays.js
```diff
@@ -10,14 +10,15 @@
 
 'use strict'; // eslint-disable-line strict
 
-require('flow-remove-types/register');
-
 const fs = require('fs');
+
+require = require('@std/esm')(module, {"cjs":true, "esm":"js"});
+
```
mourner: Can you expand on why `flow-remove-types` and `@std/esm` don't play well together? It would be great if we could avoid generating the source to a separate `src_untyped` folder...

bench/unit/bench_transform.js
anandthakker: We can reinstate them or remove them in a separate PR -- I was just getting distracted by the error messages coming from these files :)

src/data/bucket.js
anandthakker: 👍 yeah these cases were either codemod outputs (or just the quickest manual fix for the time being).  I'm not sure how much tree-shaking we'll enjoy from within our codebase, since I don't think we have too much dead code within `src/`... but if we do end up going the code-splitting route, this could make a notable difference in the worker blob size.   It'll also just be simpler/clearer code.

mourner:
BTW how's the total bundle size (minified/gzipped) looking?

anandthakker:
@mourner pretty good!

### master
```
-rw-r--r--   1 anand  staff   682K Dec 12 23:23 mapbox-gl.js
-rw-r--r--   1 anand  staff   168K Dec 12 23:23 mapbox-gl.js.gz
```

### rollup-anandthakker
```
-rw-r--r--   1 anand  staff   555K Feb 22 11:13 mapbox-gl.js
-rw-r--r--   1 anand  staff   144K Feb 22 11:13 mapbox-gl.js.gz
```

build/generate-struct-arrays.js
anandthakker: Some context here: https://github.com/standard-things/esm/issues/119

anandthakker: > It would be great if we could avoid generating the source to a separate src_untyped folder...

I agree, I'm really hoping we can avoid a build step for tests 🤞

anandthakker:
Alright, using Rollup's relatively new code splitting feature, we're now at a point where we have a smaller bundle that loads as fast as our browserify build.  Before I embark on the longish tail of tasks that would be needed to finish this and merge it into `master`, it's time to ask: should I?

The advantages:
- A smaller bundle, with the potential for (possibly significant?) further optimizations via "tree shaking"
- Modernized codebase:
  - language-level module declarations (rather than runtime ones via `require()`/`module.exports`)
  - better handling of cyclic dependencies
  - cleaner code (IMO, anyway)
  - consistent with our ubiquitous Flow `import type` / `export type` declarations

The disadvantages:
- Browserify has been around a long time, is well hardened, we know it works.
- Requires the use of Rollup's new "code splitting" functionality, which is still new and behind a feature flag (`experimentalCodeSplitting`).  (On the other hand, there are a couple of active core maintainers who have already been very responsive.)
- Might require something like a build step to handle the combination of ES modules + Flow before we can run code in Node (for unit tests, integration tests, codegen)

@mapbox/gl-core what do you think?  Please weigh in!  It's definitely not too late to abandon this effort if we don't feel that the time is right.

ChrisLoer:
I can't really comment on the overall approach although it sounds like progress! Two questions:

- How would "dynamic imports to fetch modules on demand" work? I'm wondering if we'd end up building out infrastructure that would make it convenient to have mapbox-gl-rtl-text auto-load instead of requiring additional configuration. Even closer to the dream would be to dynamically choose whether to load a WASM or ASM.js version of the plugin based on browser capability.
- You've been doing a lot of bechmarking of time-to-first-render, right? Can you document any of those numbers here?

jfirebaugh:
> Might require something like a build step to handle the combination of ES modules + Flow before we can run code in Node (for unit tests, integration tests, codegen)

I think this is the most important unanswered question. [Way back in the day](https://github.com/mapbox/mapbox-gl-js/issues/855) we reverted an early attempt to introduce ES6 syntax because tooling wasn't ready -- specifically debuggers and code coverage. I rely on running individual test files pretty frequently, often attaching a node debugger to track down test failures. If switching to ES6 modules would mean I can no longer do that, I'd prefer to hold off on it.

anandthakker:
🎉 looks like `flow-remove-types` and `@std/esm` can cooperate after all!  @jfirebaugh I confirmed that I'm able to attach a debugger with `node --inspect` and step through original sources.

anandthakker:
Another open question: how should we build the benchmarks?  Options:

1. Reproduce the code-splitting trick we're using for the main bundle.  Pro: we continue to have a single `benchmarks.js` bundle that gets published on each `master` CI build and for each release.  Con: more complicated build setup.
2. Generate a separate `worker.js` script, and, in `benchmarks.js`, set `mapboxgl.workerUrl' to point to it.  This would mean that the web workers would be loaded from that `worker.js` file rather than a blob URL.  Pro: less complicated build setup.  Con: we would have to publish two files--`benchmarks.js` and `worker.js`.

I was initially thinking that another downside to option 2 is that it would preclude being able to benchmark anything that included worker load time, because we'd be loading the worker in a different manner than in the shipped library.  But in fact, I don't think it would really be practical to measure this within our benchmarking system regardless of how we load the workers, so I don't think this should be a factor.

anandthakker:
> How would "dynamic imports to fetch modules on demand" work? I'm wondering if we'd end up building out infrastructure that would make it convenient to have mapbox-gl-rtl-text auto-load instead of requiring additional configuration. Even closer to the dream would be to dynamically choose whether to load a WASM or ASM.js version of the plugin based on browser capability.

@ChrisLoer I haven't dug into the details here, but I _think_ dynamic imports might indeed make this a possibility. (But we'd likely have to also have to keep supporting the existing approach for the foreseeable future.)

> You've been doing a lot of bechmarking of time-to-first-render, right? Can you document any of those numbers here?

Not much more than what's documented in https://github.com/mapbox/mapbox-gl-js/issues/5939 -- I'll do another pass at this again soon, along with running our normal benchmarks, and post results here.

yarn.lock
```diff
@@ -6957,78 +6906,62 @@ node-gyp@^3.0.3, node-gyp@^3.3.1:
...
     tar "^2.2.1"
     tar-pack "^3.4.0"
 
 "nomnom@>= 1.5.x":
-  version "1.8.1"
-  resolved "https://registry.yarnpkg.com/nomnom/-/nomnom-1.8.1.tgz#2151f722472ba79e50a76fc125bb8c8f2e4dc2a7"
-  dependencies:
-    chalk "~0.4.0"
-    underscore "~1.6.0"
+  version "2.0.0"
```
jfirebaugh: Revert this to fix the build (https://github.com/zaach/jsonlint/issues/103#issuecomment-367416531).

anandthakker:
> Apart from the remaining bullet points and making the build green, I'd love to add a bunch of eslint-plugin-import rules to our ESLint config — used this plugin and it is fantastic. Can help catch a lot of subtle bugs.

@mourner oooh nice, I was hoping to find a good set of import-related eslint rules, and this looks perfect.

> Also, would be great to fix a few remaining export const exported = { ... } cases, but I'm fine with leaving it for a follow-up PR if it makes it easier to ship.

👍 yeah, I'll either fix up the rest of these or ticket separately

jfirebaugh:
I played around with debugging unit tests in [WebStorm](https://www.jetbrains.com/webstorm/), my environment of choice, and it worked great. Just had to add `-r ./build/flow-remove-types -r @std/esm` to the run configuration -- you can add this in a template and then run/debug any individual `*.test.js` file with a keyboard shortcut.

anandthakker:
Time to first render:

## rollup-anandthakker:

screen shot 2018-03-07 at 3 39 42 pm

## master

screen shot 2018-03-07 at 3 39 54 pm

anandthakker:
Minified bundle size:

```
-rw-r--r--  1 anand  staff   709K Mar  7 15:37 dist/mapbox-gl--master.js
-rw-r--r--  1 anand  staff   572K Mar  7 15:42 dist/mapbox-gl--rollup-anandthakker.js
```

```
-rw-r--r--  1 anand  staff   169K Mar  7 15:37 dist/mapbox-gl--master.js.gz
-rw-r--r--  1 anand  staff   148K Mar  7 15:42 dist/mapbox-gl-rollup-anandthakker.js.gz
```

anandthakker:
This is ready for a full review.  The history is broken up into fairly meaningful commits, with automated or mostly-superficial changes marked with `[codemod]` and `[lint]`.  See the top of the ticket for a couple notes.

src/gl/value.js
```diff
@@ -658,7 +659,9 @@ module.exports = {
...
     PixelStoreUnpack,
     PixelStoreUnpackPremultiplyAlpha,
-
     ColorAttachment,
-    DepthAttachment,
+    DepthAttachment
 };
+
+export default exported;
+export { ClearColor, ClearDepth, ClearStencil, ColorMask, DepthMask, StencilMask, StencilFunc, StencilOp, StencilTest, DepthRange, DepthTest, DepthFunc, Blend, BlendFunc, BlendColor, Program, LineWidth, ActiveTextureUnit, Viewport, BindFramebuffer, BindRenderbuffer, BindTexture, BindVertexBuffer, BindElementBuffer, BindVertexArrayOES, PixelStoreUnpack, PixelStoreUnpackPremultiplyAlpha, ColorAttachment, DepthAttachment };
```
lbud: @anandthakker are these intentionally exported twice (and if so what does that mean)?

src/render/glyph_atlas.js
```diff
@@ -70,6 +71,9 @@ function makeGlyphAtlas(stacks: {[string]: {[number]: ?StyleGlyph}}): GlyphAtlas
     return {image, positions};
 }
 
-module.exports = {
+const exported = {
     makeGlyphAtlas
 };
+
+export default exported;
+export { makeGlyphAtlas };
```
lbud: same question here as above (I assume it's intentional/related to default exports?)

src/render/image_atlas.js
```diff
@@ -95,7 +96,10 @@ function makeImageAtlas(images: {[string]: StyleImage}): ImageAtlas {
...
 }
 
-module.exports = {
+const exported = {
     imagePosition,
     makeImageAtlas
 };
+
+export default exported;
+export { imagePosition, makeImageAtlas };
```
lbud: double export

bench/benchmarks.js
```diff
@@ -13,19 +15,36 @@ function register(Benchmark) {
...
+register(Validate);
+register(StyleLayerCreate);
+register(QueryPoint);
+register(QueryBox);
+ExpressionBenchmarks.forEach(register);
+register(FilterCreate);
+register(FilterEvaluate);
 
 // Ensure the global worker pool is never drained. Browsers have resource limits
 // on the max number of workers that can be created per page.
```
jfirebaugh: Attach this comment to `getWorkerPool().acquire(-1);`.

package.json
```diff
@@ -40,28 +38,25 @@
...
   "devDependencies": {
-    "@mapbox/batfish": "^0.13.3",
+    "@mapbox/batfish": "^1.0.4",
     "@mapbox/mapbox-gl-rtl-text": "^0.1.1",
     "@mapbox/mapbox-gl-test-suite": "file:test/integration",
+    "@std/esm": "^0.21.7",
     "babel-eslint": "^7.0.0",
     "benchmark": "~2.1.0",
-    "browserify": "^14.0.0",
+    "browserify": "^16.1.0",
```
jfirebaugh: No longer necessary?

src/data/extent.js
```diff
@@ -1,19 +1,2 @@
 // @flow
-
-/**
```
jfirebaugh: Please preserve this comment.

src/data/index_array_type.js
```diff
@@ -1,12 +1,8 @@
 // @flow
 
-/**
```
jfirebaugh: And this one. Was their removal a codemod effect? If so, please sweep for all cases.

src/source/tile.js
```diff
@@ -42,8 +39,8 @@ export type TileState =
     | 'reloading' // Tile data has been loaded and is being updated. Tile can be rendered.
     | 'unloaded'  // Tile data has been deleted.
     | 'errored'   // Tile data was not loaded because of an error.
-    | 'expired';  /* Tile data was previously loaded, but has expired per its
```
jfirebaugh: Lost some significant whitespace here.

src/style-spec/expression/parsing_context.js
```diff
@@ -177,14 +183,10 @@ class ParsingContext {
     }
 }
 
-module.exports = ParsingContext;
+export default ParsingContext;
 
 function isConstant(expression: Expression) {
     // requires within function body to workaround circular dependency
-    const {CompoundExpression} = require('./compound_expression');
```
jfirebaugh: 😌

src/style-spec/migrate.js
```diff
@@ -1,22 +1,11 @@
 
-/**
```
jfirebaugh: Lost comment.

src/style-spec/validate/latest.js
```diff
@@ -1,9 +1,3 @@
 
-/*
```
jfirebaugh: Lost comment.

src/data/bucket.js
jfirebaugh: Can you ticket this as a followup item?

package.json
anandthakker: Still needed for a build unit test that confirms that the built file can be used with browserify.

src/data/index_array_type.js
anandthakker: Ack, yes -- I caught this happening with the `// @flow` comments and fixed it, but didn't realize it was affecting subsequent comments as well.  Will fix.

src/data/bucket.js
anandthakker: Yep, already did: https://github.com/mapbox/mapbox-gl-js/issues/6293

src/util/performance.js
```diff
@@ -1,12 +1,12 @@
 // @flow
-
-// Wraps performance.getEntriesByName to facilitate testing
-// Not incorporated into browser.js because the latter is poisonous when used outside the main thread
```
anandthakker: Dropped comment

PR #6205, opened 2018-02-21
add trigger method to GeolocateControl

### LINKED ISSUE #5464: Allow GeolocateControl's geolocation to be triggered programmatically
## Motivation

The [Geolocate control](https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/control/geolocate_control.js) allows the user's location to be obtained when the user clicks on the associated UI button. However, there is currently no way to have the geolocate functionality trigger when the map first loads.

At present, the map loads and displays the default location it is given. It would be useful to trigger the geolocate functionality on load so the map automatically centers on the user's location without the user first having to click the Geolocate control button.

## Design Alternatives

This is my current solution:

```javascript
const geolocate = new mapboxgl.GeolocateControl(...)
map.addControl(geolocate)
setTimeout(geolocate._onClickGeolocate.bind(geolocate), 5)
```

The logic performed by `addControl` is not entirely synchronous, therefore, it is necessary to delay calling `_onClickGeolocate` until the relevant DOM elements exist.

This solution is not ideal as it requires calling a private method and introduces a race condition mediated by an arbitrary time delay.

## Design

**Two proposals:**
- Add option to have the geolocation functionality trigger when the control is added to the map
- Expose public method on `GeolocateControl` allowing the programmer to trigger the geolocation functionality, along with a `ready` event that is fired once the `GeoLocate` control has been added to the map and all DOM elements have been created.

The latter proposal offers more flexibility to the programmer, while the former is more restrictive and only affords the programmer the additional ability to have the geolocate functionality trigger when the control is first added.

I am in favor of the second proposal as it enables the programmer to trigger geolocation at any time after the control has been added to the map as opposed to just initially.

andrewharvey:
closes #5464

## Launch Checklist
adds a public trigger method to the GeolocateControl, allowing it to be invoked programatically.

 - n/a post benchmark scores

manually testing triggering both with and without having added the control to the map

### DIFF
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -88,6 +88,7 @@ class GeolocateControl extends Evented {
     _watchState: string;
     _lastKnownPosition: any;
     _userLocationDotMarker: Marker;
+    _setup: boolean; // set to true once the control has been setup
 
     constructor(options: any) {
         super();
@@ -99,8 +100,7 @@ class GeolocateControl extends Evented {
             '_finish',
             '_setupUI',
             '_updateCamera',
-            '_updateMarker',
-            '_onClickGeolocate'
+            '_updateMarker'
         ], this);
     }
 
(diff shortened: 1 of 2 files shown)


andrewharvey:
The workaround in #5464 says:

> The logic performed by addControl is not entirely synchronous, therefore, it is necessary to delay calling _onClickGeolocate until the relevant DOM elements exist.

I still need to confirm if this PR has the same issue.

andrewharvey:
> I still need to confirm if this PR has the same issue.

I think there is a race condition as the control could be added to the map before the UI elements are added, since onAdd only does the UI setup via a callback so it could return first.

trigger doesn't crash, just returns false if this happens. I haven't run into the issue so think it should be okay.

PR #6208, opened 2018-02-21
Fit Camera to Feature

### LINKED ISSUE #2245: Replace fitBounds with fitFeature
"bounds" is an ill-defined concept in mapbox-gl-js due to pitch and rotation. A less ambiguous and more useful API would fit the map to an arbitrary GeoJSON feature. 

ref https://github.com/mapbox/mapbox-gl-js/issues/2112

asheemmamoowala:
Closes #2245.

### TODO
 - [ ] description
 - [ ] tests
 - [ ] documentation

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -342,6 +342,65 @@ class Camera extends Evented {
         return this;
     }
 
+    /**
+     * Pans and zooms the map to contain its visible area within the geographical bounds of the feature geometry.
+     * This function will also reset the map's bearing to 0 if bearing is nonzero.
+     *
+     * @memberof Map#
+     * @param feature Center the bounds of the feature's geometry in the viewport and use the highest
+     *      zoom level up to and including `Map#getMaxZoom()` that fits it in the viewport.
+     * @param options
+     * @param {number | PaddingOptions} [options.padding] The amount of padding in pixels to add to the given bounds.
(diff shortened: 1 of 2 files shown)


src/ui/camera.js
```diff
@@ -342,6 +342,65 @@ class Camera extends Evented {
         return this;
     }
 
+    /**
+     * Pans and zooms the map to contain its visible area within the geographical bounds of the feature geometry.
+     * This function will also reset the map's bearing to 0 if bearing is nonzero.
```
andrewharvey: does this mean fit to feature is pitch aware and will retain the existing pitch?

andrewharvey: > This function will also reset the map's bearing to 0 if bearing is nonzero.

What are the implications of this for backwards compatibility? In the future if we can support retaining the bearing as part of `fitFeature` would it just be a breaking change?

asheemmamoowala: It does not alter the pitch, but it also doesn't account for it and will not give the correct result. 
There is no resolution yet for the pitch aware fitting.

There is more work needed to maintain the bearing or change it to some other value as part of the fitting operation.

navenedrob:
@ asheemmamoowala What is the status of this? I've run into a use case where this would be very valuable.

mourner:
Closing as stale — feel free to reopen if there's still interest in shipping this.

PR #6209, opened 2018-02-21
adjust empty string fix to match -native

ansis:
This should not change behaviour. This just brings -js slightly closer to the -native fix: https://github.com/mapbox/mapbox-gl-native/pull/11206

benchmarks:
The benchmarks looked good except for Layout. But I wasn't able to reproduce the that difference on [later runs](https://bl.ocks.org/anonymous/raw/e9a09f3cfaa256552ee0c67572b381f9/).

## Launch Checklist

 - [N/A] write tests for all new functionality
 - [x] post benchmark scores

### DIFF
--- a/src/symbol/placement.js
+++ b/src/symbol/placement.js
@@ -112,14 +112,11 @@ class Placement {
 
         const partiallyEvaluatedTextSize = symbolSize.evaluateSizeForZoom(bucket.textSizeData, this.transform.zoom, symbolLayoutProperties.properties['text-size']);
 
-        const iconWithoutText = !bucket.hasTextData() || layout.get('text-optional');
-        const textWithoutIcon = !bucket.hasIconData() || layout.get('icon-optional');
-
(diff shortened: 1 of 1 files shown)


src/symbol/placement.js
```diff
@@ -170,6 +167,11 @@ class Placement {
                     offscreen = offscreen && placedIconBoxes.offscreen;
                 }
 
+                const hasTextData = Boolean(symbolInstance.collisionArrays.textBox || symbolInstance.collisionArrays.textCircles);
+                const hasIconData = Boolean(symbolInstance.collisionArrays && symbolInstance.collisionArrays.iconBox);
```
ansis: The `symbolInstance.collisionArrays &&` was needed to satisfy flow complaining about `symbolInstance.collisionArrays` being possible undefined. I'm not sure why flow didn't complain for the line above.

mourner:
@ChrisLoer looks like this might have gotten lost in notifications — can you revisit?

PR #6218, opened 2018-02-23
Event improvements

### LINKED ISSUE #2891: Refactor interaction handlers to never unbind their event listeners

### Steps to Trigger Behavior

https://jsbin.com/havivaruwe/edit?html,output

 1. map.doubleClickZoom.disable();
 2. on double click
     * map.fitBounds(...)
     *  map.doubleClickZoom.enable();
 3. zoom back out
 4. on double click
      * map.fitBounds(...)

### Expected Behavior

I expect the fit bounds call to bring me to the bounds

### Actual Behavior

the double click zoom happens.

### LINKED ISSUE #3290: dragPan.disable() does not work first time it's run
This is based on the latest Mapbox GL JS (v0.25.1).  I've posted a demo repo which has one file (index.html) which when loaded in Chrome fully replicates the issue.  That file is an almost complete replica of: https://www.mapbox.com/mapbox-gl-js/example/drag-a-point/

The only differences are the places where `map.dragPan.enable()` and `map.dragPan.disable()` are called.  Instead of calling those functions during `mousemove`, I call them during `mouseup` and `mousedown`, but that should work fine too...

Repro file at: https://github.com/ntilwalli/mapboxDragPanIssue

### Steps to Trigger Behavior

 1. Load the index.html file from the repo
 2. Try dragging the circle (it won't work)
 3.  Try dragging a second time (it will work)

### Expected Behavior

Dragging circle should always work

### Actual Behavior

Dragging circle only works from second attempt onwards

jfirebaugh:
* Introduces preventDefault for mousedown/touchstart/dblclick (fixes #3290)
* Never unbind mousedown/touchstart handlers (fixes #2891)

Tested in Mac Chrome, Mac Firefox, iOS Safari.


src/ui/events.js
```diff
@@ -25,38 +27,141 @@ export type MapMouseEvent = {
...
+    lngLat: LngLat;
+
+    /**
+     * Prevents subsequent default processing of the event by the map.
+     *
+     * Calling this method will prevent the following default map behaviors:
+     *
+     *   * On `mousedown` events, the behavior of {@link DragPanHandler}
+     *   * On `mousedown` events, the behavior of {@link DragRotateHandler}
+     *   * On `mousedown` events, the behavior of {@link BoxZoomHandler}
```
anandthakker: I think we should also have:
 - On `wheel` events, the behavior of `{@link ScrollZoomHandler}`

src/ui/handler/drag_pan.js
```diff
@@ -85,23 +82,22 @@ class DragPanHandler {
...
+    onDown(e: MouseEvent | TouchEvent) {
+        if (!this.isEnabled()) return;
         if (this._map.boxZoom.isActive()) return;
         if (this._map.dragRotate.isActive()) return;
         if (this.isActive()) return;
 
         if (e.touches) {
             if ((e.touches: any).length > 1) return;
-            window.document.addEventListener('touchmove', this._onMove);
+            window.document.addEventListener('touchmove', this._onMove, {capture: true});
```
anandthakker: Include comment mentioning why `capture: true`

jfirebaugh:
> what's your thinking re: *move events: does it still make sense for handlers to manage listening to them directly, or should we have the bindHandlers listeners dispatch to the appropriate handlers for these like we're doing for *{start,end}, etc.?

The latter was what I tried initially, but I ended up reverting. The issue is that the drag handlers need to _capture_ move events by binding the listener to the window (for the duration of the drag), while the general move event handler should receive only canvas container move events (but is permanent). So it's easier to keep them separated.

lucaswoj:
Looks great @jfirebaugh! Thank you for improving the clarity & robustness of our event type system while working on this new feature.

src/ui/events.js
jfirebaugh: #6225

PR #6227, opened 2018-02-25
add box-sizing to user-location-dot css

andrewharvey:
## Launch Checklist

if the map div has inherited a `box-sizing: border-box` for example applied to a whole application outside of mapbox gl js then the geolocation dot will look wrong. This PR explicitly sets the geolocation dot to use `box-sizing: content-box;`.

 - n/a write tests for all new functionality
 - n/a post benchmark scores

### DIFF
--- a/dist/mapbox-gl.css
+++ b/dist/mapbox-gl.css
@@ -439,6 +439,7 @@ a.mapboxgl-ctrl-logo {
     position: absolute;
     top: -2px;
     width: 15px;
+    box-sizing: content-box;
 }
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Is it possible to adjust the sizing to work with `box-sizing: border-box`? That's what we use elsewhere in the CSS (still with an explicit declaration, of course).

andrewharvey:
> Is it possible to adjust the sizing to work with box-sizing: border-box? That's what we use elsewhere in the CSS (still with an explicit declaration, of course).

Good point, just changed it

PR #6232, opened 2018-02-26
 Support disable() mid drag or rotate gesture

### LINKED ISSUE #2419: Support dragPan.disable() mid-drag

jfirebaugh:
Moves `DragPanHandler` and `DragRotateHandler` closer to a state machine based implementation, so that, when disabled, they can perform the appropriate actions in any state.

These two handlers have four states:

* `disabled`: The handler is disabled, and should not respond to any input.
* `enabled`: The handler is enabled but has not started receiving any input.
* `pending`: The handler has received a `mousedown` or `touchstart` event, but not a subsequent `move` event.
* `active`: The handler has received a `mousedown` or `touchstart` event and at least one subsequent `move` event.

Disabling a handler in the `enabled` state doesn't generally require anything special. Disabling a handler in the `pending` state requires unbinding some event listeners. Disabling a handler in the `active` state requires unbinding some event listeners, sending appropriate `*end` events, and _not_ triggering an inertial camera animation.

Fixes #2419.

### DIFF
--- a/src/ui/handler/drag_pan.js
+++ b/src/ui/handler/drag_pan.js
@@ -87,8 +87,6 @@ class DragPanHandler {
 
     onDown(e: MouseEvent | TouchEvent) {
         if (!this.isEnabled()) return;
-        if (this._map.boxZoom.isActive()) return;
-        if (this._map.dragRotate.isActive()) return;
         if (this.isActive()) return;
 
         // Bind window-level event listeners for move and up/end events. In the absence of
--- a/src/ui/handler/drag_pan.js
+++ b/src/ui/handler/drag_pan.js
@@ -38,7 +38,9 @@ class DragPanHandler {
 
         util.bindAll([
             '_onMove',
-            '_onUp',
+            '_onMouseUp',
+            '_onTouchEnd',
+            '_onBlur',
             '_onDragFrame'
         ], this);
     }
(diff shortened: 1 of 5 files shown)


src/ui/handler/drag_pan.js
```diff
@@ -164,43 +180,82 @@ class DragPanHandler {
...
+    }
 
-        if (!this.isActive()) return;
-
-        this._active = false;
+    _deactivate() {
         delete this._lastMoveEvent;
         delete this._previousPos;
         delete this._pos;
+    }
```
anandthakker: Just noting: because there's a state called `'active'`, I (briefly) found it confusing to have a method called `_deactivate` that doesn't affect `this.state`... but I can't think of any better names.

PR #6246, opened 2018-02-28
fix line-pattern sampling from outside icon tex coords

### LINKED ISSUE #5978: Line-pattern rendering pieces of other sprites
**mapbox-gl-js version**: 0.43.0

[Similar/same issue appears to have come up and been resolved in native](https://github.com/mapbox/mapbox-gl-native/issues/9976). Hopefully I'm not making a repeat issue here, but didn't see it in gl-js issue queue.

### Steps to Trigger Behavior

 1. Create a line pattern on a thicker line (use motorway in Streets for instance)
 2. Add an icon (try `de-federal-4`, not all icons behave this way, but enough do)
 3. Zoom in anywhere between z 17 and z 22 (Also works if you just increase width significantly)
 4. See other icons partially rendered in addition to desired icon

image

### Expected Behavior

Scale icon as needed for `line-pattern`.

### Actual Behavior

Scales icon, but shows pieces of other sprites at high zooms or large widths.

mollymerp:
fixes #5978 

not sure if we should repeat the pattern instead of stretching 💭 

test on master:
actual

this branch: 

actual

_updated with render test after d78d242_

## Launch Checklist

 - [ ] post benchmark scores


mollymerp:
hmm whoops – this is not the correct fix. it is flipping the icon, and not stretching the whole pattern (duh 🤦‍♀️ )
expected

mollymerp:
Ok, I think the fix is correct now. Added some explanatory comments as well.

ansis:
Hmm, another option besides repeating and stretching would be just to draw nothing in the extra space. So if you have a 24px pattern and a 36px wide line it would look the same as a 24px pattern and 24px line. 

I'm not sure which would be preferable, I guess it depends on the pattern?

mollymerp:
mm good point. @aparlato @nickidlugash any opinions on desired behavior for line patterns with width > pattern height?

aparlato:
> mm good point. @aparlato @nickidlugash any opinions on desired behavior for line patterns with width > pattern height?

Oy this is a tough call. I'm inclined to say stretching meets more needs overall? The instances that I have used this property, this was the behavior I wanted. Although, @mollymerp, it looks like this only stretches it by width rather than treating the whole icon as larger. What's the reason behind just stretching by width rather than maintaining aspect ratio?

> Hmm, another option besides repeating and stretching would be just to draw nothing in the extra space. So if you have a 24px pattern and a 36px wide line it would look the same as a 24px pattern and 24px line.

@ansis I think this makes sense in many cases, but since this can also be accomplished by just maintaining a 24px line, I think from a styling perspective this is less flexible unless I've misunderstood the comment.

ansis:
> Oy this is a tough call. I'm inclined to say stretching meets more needs overall? 

Sounds good, let's do this.

>  it looks like this only stretches it by width rather than treating the whole icon as larger. What's the reason behind just stretching by width rather than maintaining aspect ratio?

Scaling horizontally is a bit problematic. If the pattern is scaled at anything other than the rate the line is scaled at then the pattern will look like it is sliding along the line.

> @ansis I think this makes sense in many cases, but since this can also be accomplished by just maintaining a 24px line, I think from a styling perspective this is less flexible unless I've misunderstood the comment.

Yep, that's a good point.

src/render/draw_line.js
```diff
@@ -75,7 +75,7 @@ function drawLineTile(program, painter, tile, bucket, layer, coord, programConfi
             imagePosB = painter.imageManager.getPattern(image.to);
             if (!imagePosA || !imagePosB) return;
 
-            gl.uniform2f(program.uniforms.u_pattern_size_a, imagePosA.displaySize[0] * image.fromScale / tileRatio, imagePosB.displaySize[1]);
+            gl.uniform2f(program.uniforms.u_pattern_size_a, imagePosA.displaySize[0] * image.fromScale / tileRatio, imagePosA.displaySize[1]);
```
ansis: good catch

PR #6259, opened 2018-03-01
[test] Back to URL transform in HTTP server

brunoabinader:
In GL JS, the `node_modules` directory is found by traversing 3 levels of subdirectories, in GL native it requires one extra level because `mapbox-gl-js` is a git submodule in GL native.


brunoabinader:
@jfirebaugh partially reverted 097c8cc52e3dbe56f59dd312fda0b9ce63c78ee4 as per discussion - using `require.resolve` to address node modules relative path, thus no longer requiring `..\..` paths.

Issue #5786, opened 2017-12-01
Make "Missing Mapbox GL JS CSS" a console warning


jfirebaugh:
Putting "Missing Mapbox GL JS CSS" in the DOM has confused at least one user (#5785) and causes other issues (https://github.com/mapbox/mapbox-gl-js/issues/5359#issuecomment-342325200).

Instead, it should be a console warning that's logged when a map is created and there's no CSS rule targeting `.mapboxgl-map` present. This can be checked using [CSSOM APIs](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model).

rtwalz:
For now the workaround I'm using is adding `display: none` to the class `mapboxgl-missing-css`

stevage:
Just encountered another confused user. (He told me that "the CSS isn't loading", on the strength of that DOM element.) Another possible solution would be deleting the DOM element once the CSS has indeed loaded.

jczaplew:
FWIW, the CSSOM APIs only function as expected with cross-loaded CSS resources when the attribute `crossorigin="anonymous"` is added to the `<link>`. 

If you load the CSS as most people do, with `<link href='https://api.tiles.mapbox.com/mapbox-gl-js/v0.45.0/mapbox-gl.css' rel='stylesheet' />`, the rules of the stylesheet are not available via `document.styleSheets`. I'm uncertain as to whether this is a chromium bug or feature, but you could hack around this by making an additional request for all stylesheets that do not contain a `rules` or `cssRules` attribute, but that seems like a heavy-handed solution for a fairly small problem.
 
In the absence of a better solution, simply checking the computed style of `.mapboxgl-map` might suffice. Something like 

````javascript
let mapboxglMap = document.querySelector('.mapboxgl-map')
if (!mapboxglMap || window.getComputedStyle(mapboxglMap).overflow != 'hidden') {
    // CSS error
}
````

The nasty bit is that it depends on a hardcoded CSS value that could change in the future. Thoughts as to whether this might be an appropriate method of checking if the CSS has loaded?

Issue #5785, opened 2017-12-01
Missing Mapbox GL JS CSS


tetsuo:
I am trying to make this work with choojs and Browserify.

The error I'm getting: `Missing Mapbox GL JS CSS` Actually I need to open Chrome devtools to see the error message, it is written in a tag with `display: none`: `<div class="mapboxgl-missing-css">Missing Mapbox GL JS CSS</div>`

I have included a link to the CSS file in HTML, but still all I'm seeing is this error message.

What is `Mapbox GL JS CSS`? Is that a file name?

How do you determine that the CSS file is missing by the way? and why is that so important?

mourner:
It's not an error, it's just a placeholder that becomes visible only when CSS is actually missing. Are you experiencing any problems with the map itself?

tetsuo:
@mourner Thanks for the reply!

Following is the example I'm trying to make work:

```
const cache = require('cache-element')
const html = require('choo/html')
const choo = require('choo')

mapboxgl.accessToken = 'xxx';

const app = choo()
app.route('/', mainView)
app.mount('body')

const nav = cache(() => html`<div id="map" />`)

function mainView() {
  return html`
    <body onload=${() => {
      const map = new mapboxgl.Map({
        container: 'map',
        style: 'mapbox://styles/mapbox/dark-v9',
        center: [-93.261, 44.971]
      });
    }}>
      ${nav.render()}
    </body>
  `
}
```

I'm `browserify`ing this with `sheetify` btw:

```
yarn run watchify app.js -t sheetify -o public/bundle.js -dv  
```

And this is the end result I'm getting:

![x](https://i.imgur.com/oaAkIRN.png)

tetsuo:
> It's not an error, it's just a placeholder that becomes visible only when CSS is actually missing.

So the CSS file is not missing, the error is not visible, and the map is not showing :)

mourner:
Can you set up a minimal JSFiddle test case with all the already generated code? We can't afford debugging all possible configurations of front-end bundling, but with an isolated test case, it would be easy to see where the problem is.

tetsuo:
Hmm, I'm not sure if `browserify` works on JSfiddle, so I can't create a proper isolated test-case on it.

I thought maybe it is a known issue, since the example I used is super easy and the browserify usage is very common. Thanks for the answer. Unfortunately I cannot afford writing a test-case right now as well :\ Good luck.

zavan:
[...]

Issue #5359, opened 2017-09-27
Display some kind of warning if mapbox-gl CSS is is missing


anandthakker:
We should somehow issue a warning if the mapbox-gl CSS is missing.  As suggested by @nfarina in https://github.com/mapbox/mapbox-gl-js/issues/4048#issuecomment-332560458, a div added to the map's target element and hidden by our default CSS should do the trick.

eltonjuan:
A random callout but if you're delivering your CSS in JS (e.g. local development with Webpack) the injection of this div causes Mapbox's calculation for determining the canvas size to be thrown off by the amount of the div. 

Not really an issue since you should be generating a CSS bundle in production, but was driving me a little mad (thinking my CSS rules were messed up). Figured I'd call it out in case anyone else goes searching...

tywoodpav:
@eltonjuan How did you get around this? V annoying

Srounsroun:
@tywoodpav Make sure to reference css link in index.html headers 

```<link href='https://api.mapbox.com/mapbox-gl-js/v0.42.0/mapbox-gl.css' rel='stylesheet' />```

and load the correct **style** in mapbox:

```
var map = new mapboxgl.Map({
container: 'YOUR_CONTAINER_ELEMENT_ID',
style: 'mapbox://styles/mapbox/streets-v10'
});
```

as describe in https://www.mapbox.com/install/js/bundler-install/

PR #6261, opened 2018-03-01
Use console warning rather than DOM element to indicate missing stylesheet

lbud:
This uses CSSOM interfaces to check for the presence of a `.mapboxgl-map` CSS rule upon map creation and display a warning in the console if it's missing, rather than adding and hiding a DOM element (ref https://github.com/mapbox/mapbox-gl-js/issues/5785, https://github.com/mapbox/mapbox-gl-js/issues/5359#issuecomment-342325200).

Fixes https://github.com/mapbox/mapbox-gl-js/issues/5786.

Having to stub `console.warn` so many places in the unit tests feels weird, and treacherous for future writers of tests that include `Map` — if there's a better way to do this I'm all ears!

## Launch Checklist

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1474,7 +1474,7 @@ class Map extends Camera {
     _detectMissingCSS(): void {
         const computedColor = window.getComputedStyle(this._missingCSSCanary).getPropertyValue('background-color');
         if (computedColor !== 'rgb(250, 128, 114)') {
-            util.warnOnce('Missing Mapbox GL JS CSS: map may not display correctly.');
+            warnOnce('Missing Mapbox GL JS CSS: map may not display correctly.');
         }
     }
 
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1474,7 +1474,10 @@ class Map extends Camera {
     _detectMissingCSS(): void {
         const computedColor = window.getComputedStyle(this._missingCSSCanary).getPropertyValue('background-color');
         if (computedColor !== 'rgb(250, 128, 114)') {
-            warnOnce('Missing Mapbox GL JS CSS: map may not display correctly.');
+            warnOnce('This page appears to be missing CSS declarations for ' +
+                'Mapbox GL JS, which may cause the map to display incorrectly. ' +
+                'Please ensure your page includes mapbox-gl.css, as described ' +
+                'in https://www.mapbox.com/mapbox-gl-js/api/.');
         }
     }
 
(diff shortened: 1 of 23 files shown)


src/ui/map.js
```diff
@@ -1388,12 +1387,20 @@ class Map extends Camera {
         return [width, height];
     }
 
+    _hasCSS(): boolean {
+        for (const sheet of (window.document.styleSheets: any)) {
```
jfirebaugh: How fast is this if a page has a bajillion CSS rules?

If iterating over them all in JS is expensive, another approach would be:

* Add `.mapboxgl-canary { background-color: salmon; }` to mapbox-gl.css. (Unfortunately `canary` is not a CSS color keyword so I chose another animal.)
* Map adds `<div class="mapboxgl-canary" style="visibility: hidden;"/>` and then checks the computed background color.

lbud:
Oof, looks like there was a bug (possibly introduced in Chrome 64) causing cross-origin stylesheets not only to be missing a `cssRules` property but to error upon calling that getter (https://bugs.chromium.org/p/chromium/issues/detail?id=813826). Discovered while digging through studio investigating https://github.com/mapbox/mapbox-gl-js/pull/6261#discussion_r171751284.

lbud:
Switched to checking the computed color of a hidden div as suggested in https://github.com/mapbox/mapbox-gl-js/pull/6261#discussion_r171751284 (thank you @jfirebaugh!), both for performance and because of https://github.com/mapbox/mapbox-gl-js/pull/6261#issuecomment-369810390.

jfirebaugh:
> Having to stub `console.warn` so many places in the unit tests feels weird, and treacherous for future writers of tests that include `Map` — if there's a better way to do this I'm all ears!

Agreed. I can think of a couple possible approaches:

* Actually add a `mapbox-gl.css` `<link>` element to the test DOM. This would be the most "real-world" way to do it, but I don't think it'll work because AFAIK jsdom does not fully implement CSS and just has a stub implementation of `getComputedStyle`.
* Move the canary check into a private method, and create a common `createMap` function used by all tests that stubs it out.

src/ui/map.js
```diff
@@ -1471,12 +1471,20 @@ class Map extends Camera {
         return [width, height];
     }
 
+    _detectMissingCSS(): void {
+        const computedColor = window.getComputedStyle(this._missingCSSCanary).getPropertyValue('background-color');
+        if (computedColor !== 'rgb(250, 128, 114)') {
+            warnOnce('Missing Mapbox GL JS CSS: map may not display correctly.');
```
jfirebaugh: > This page appears to be missing CSS declarations for Mapbox GL JS, which may cause the map to display incorrectly. Please ensure your page includes mapbox-gl.css, as described in https://www.mapbox.com/mapbox-gl-js/api/.

Issue #2732, opened 2016-06-13
Enable property functions for *-pattern properties


lucaswoj:
Enabling property functions for *-pattern properties will require syncing `SpriteAtlas` information to the `WebWorker`s. 
## Properties
- `line-pattern`
- `fill-pattern`
- `fill-extrusion-pattern`

mollymerp:
closed in #6289

PR #6262, opened 2018-03-01
allow multiple attributes per style-spec property

mollymerp:
This refactor is necessary for #2732 because `*-pattern` properties are cross-faded and require more data to be passed to the shaders – the x,y coordinates of the top-left and bottom-right corners of up to four icons in an image_atlas, and the pixel ratio of the atlas – that can be packed in one `vec4` attribute.

I added a function that takes a `property` and `type` and returns the `PaintVertexArray` constructor for the layout needed for that property/type/expression type. This is where we will be able to add exceptions to the default paint vertex array layouts for properties with multiple attributes. 

I haven't settled on a good way to extend the `populatePaintArrays` functionality to accommodate special cases yet, but it might make sense to work that out in the follow up PR implementing `line-pattern` property functions. 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/data/program_configuration.js
+++ b/src/data/program_configuration.js
@@ -29,13 +29,16 @@ function packColor(color: Color): [number, number] {
 
 /**
  *  `Binder` is the interface definition for the strategies for constructing,
- *  uploading, and binding paint property data as GLSL attributes.
+ *  uploading, and binding paint property data as GLSL attributes. Most style-
+ *  spec properties have a 1:1 relationship to shader attribute/uniforms, but
+ *  some require multliple values per feature to be passed to the GPU, and in
+ *  those cases we bind multiple attributes/uniforms.
  *
  *  It has three implementations, one for each of the three strategies we use:
  *
  *  * For _constant_ properties -- those whose value is a constant, or the constant
  *    result of evaluating a camera expression at a particular camera position -- we
- *    don't need a vertex buffer, and instead use a uniform.
+ *    don't need a vertex attribute buffer, and instead use a uniform.
  *  * For data expressions, we use a vertex buffer with a single attribute value,
  *    the evaluated result of the source function for the given feature.
  *  * For composite expressions, we use a vertex buffer with two attributes: min and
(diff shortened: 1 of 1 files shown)


src/data/program_configuration.js
```diff
@@ -68,19 +71,19 @@ interface Binder<T> {
...
         this.value = value;
-        this.name = name;
+        this.names = names;
         this.type = type;
         this.statistics = { max: -Infinity };
     }
 
     defines() {
-        return [`#define HAS_UNIFORM_u_${this.name}`];
+        return this.names.map((name) => { return `#define HAS_UNIFORM_u_${name}`; });
```
mourner: Nit: you can omit braces and `return` here.

src/data/program_configuration.js
```diff
@@ -93,36 +96,40 @@ class ConstantBinder<T> implements Binder<T> {
...
-        } else {
-            gl.uniform1f(program.uniforms[`u_${this.name}`], value);
-        }
+        this.names.forEach((name)=>{
+            if (this.type === 'color') {
+                gl.uniform4f(program.uniforms[`u_${name}`], value.r, value.g, value.b, value.a);
+            } else {
+                gl.uniform1f(program.uniforms[`u_${name}`], value);
+            }
+        });
```
mourner: Nit: let's avoid using `forEach` and closures in favor of a `for .. of` loop.

src/data/program_configuration.js
```diff
@@ -93,36 +96,40 @@ class ConstantBinder<T> implements Binder<T> {
...
-        }];
+        const PaintVertexArray = layout;
+        this.paintVertexAttributes = names.map((name)=> {
+            return {
+                name: `a_${name}`,
+                type: 'Float32',
+                components: type === 'color' ? 2 : 1,
+                offset: 0
+            };
+        });
```
mourner: Nit: you can omit braces and `return` here too if you surround the object with `(...)`.

mollymerp:
folding this into #6289 – I incorporated your requested changes @mourner 😸

Issue #6255, opened 2018-03-01
Make Expressions fully deserializable


anandthakker:
Currently, `CompoundExpression#_evaluate` doesn't survive serialization/deserialization via our web worker transfer scheme: https://github.com/mapbox/mapbox-gl-js/blob/f87d7e3401af5902a1d9e6ded2600aa21e4c633f/src/util/web_worker_transfer.js#L101

This was fine until now, because we only evaluate a `{Source,Composite}Binder`'s expression on the worker side, during `populatePaintArray` https://github.com/mapbox/mapbox-gl-js/blob/f87d7e3401af5902a1d9e6ded2600aa21e4c633f/src/data/program_configuration.js#L139

For https://github.com/mapbox/mapbox-gl-js/issues/6020, however, we'll need to evaluate again on the main thread when making updates, and so we should refactor `CompoundExpression` so that it can be fully rehydrated on the main thread.

anandthakker:
cc @asheemmamoowala

jfirebaugh:
It would be great if we could use this as an opportunity to refactor/untangle the following types:

* `Expression`
* `StyleExpression`
* `StyleExpressionWithErrorHandling`
* `ZoomConstantExpression`
* `ZoomDependentExpression`
* `ConstantExpression`
* `SourceExpression`
* `CameraExpression`
* `CompositeExpression`
* `StylePropertyExpression`
* `StylePropertyFunction`

That seems like more types than is ideal.

anandthakker:
The performance considerations in https://github.com/mapbox/mapbox-gl-js/issues/6367 suggest that we might want to solve this another way.

Issue #6019, opened 2018-01-18
RFC: Feature identity for interactivity


ansis:
## Motivation

Vector tile sources split features across multiple tiles.

How do we tell if two features in different tiles are actually the same feature?
How do we highlight a feature across tile boundaries?
How do we only dispatch a feature event listener once, not once-per-tile?

We need some concept of what makes things "the same feature".

*For GeoJSON sources we have access to whole features so we can assign them ids guaranteed to be unique.*

### Option 1: do nothing

Treat vector tile features from different tiles as different features (like we do currently).

### Option 2: use the `id` field

Vector tile sources support an `id` field for features that [*SHOULD* be unique](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features). We could treat all features with the same `id` as the same feature. Users would need to provide this in order for interactivity to work well.

It could be possible to have multiple features with the same `id` but different `properties`, which could create some weird edge cases. How would a theoretical `map.getFeatureProperties(sourceID, featureID)` handle this case?

Would features with missing `id` fields get merged into one feature? Or never get merged?

### Option 3: use the combination of the `id` field and `properties`

We could adapt the previous option and use the combination of the `id` and `properties` as a unique identifier. Features with the same `id` but different `properties` would be treated as different features. The nice thing about this is that a combined feature always has a well-defined `properties`. Also, merging could mostly work in many cases even if `id` is not provided.

### Option 4: Using a user defined function

We could extend the style spec to let users specify an expression to calculate a unique id. This expression could default to `["id"]`. This would provide a way to merge features together but it would also add extra complexity to the style spec.

This would have the same problem as Option 2, where a single combined feature could have multiple different `properties` objects.

Also, would this apply to just vector tile sources, or also geojson?

## Design

I think we should implement *Option 3*.
*Option 2* could be fine too.

Advantages of Option 3 over Option 2:
- the combined feature has a well-defined set of `properties`
- features without `id`s don't get merged into one giant feature

Disadvantages of Option 3 compared to Option 2:
- calculating the id could be more expensive since it has to process all the `properties`
- 3 could be harder to explain than option 2

Option 4 provides extra power but also extra complexity. I think that users that need this could implement it with event listeners and data joining.

#### vector tiles
```
TileA contains: [ { id: 123, type: 'multipoint', coords: [[10, 20]], properties: { key: value }]
TileB contains: [ { id: 123, type: 'multipoint', coords: [[40, 30]], properties: { key: value }]

[...]

Issue #4006, opened 2017-01-18
Allow to keep certain styles and layers on setStyle()


indus:
## Motivation

> Injecting custom layers into Mapbox core styles is a common use case and confusing to users, especially when a user wants to swap out the core style and keep the custom layers.

## Design Alternatives

#3979 current approach
#4000 alternative approach (draw-back: need for predefined nested styles)

## Design

I would like to suggest an option that would allow the user to copy some layers/sources over to a new style to keep them when switching.

### Mock-Up

```JS
//layers appended to the end of the new layer array (in order)
setStyle('mapbox://styles/mapbox/' + layerId + '-v9', {
  copySources: ['states'],
  copyLayers: ['states']
})

// layer added before another layer in the new layer array (like addLayer(layer, before))
setStyle('mapbox://styles/mapbox/' + layerId + '-v9', {
  copySources: ['states'],
  copyLayers: [{'states':'waterway-label' }]
})
```

### Concepts

it works like a normal setStyle but after the new style is loaded the copying from the old to the new style takes place.

### Implementation

the copying-code likely could get called at the end of the style-diffing procedure.

anandthakker:
Thanks for this proposal @indus!

I'm not sure how I feel about the `copyLayers` part, but I think that `copySources` (or something like it) might fill a current shortcoming of the `setStyle` API that also isn't addressed by the nested styles proposal in #4000 

`GeoJSONSource` in particular may cary a large amount of state if its `data` is set to literal JSON.  As @averas describes [here](https://github.com/mapbox/mapbox-gl-js/pull/3643#issuecomment-275741172), copying this over from the existing style into a new target style when using `map.setStyle()` may be inconvenient and potentially a performance problem.

While it's true that using a reactive paradigm and immutable objects could mostly eliminate these problems, we can't assume that this is necessarily a valid option for all/most users.  Thus, I think it makes sense to provide a means for developers to short-circuit the diff for particular sources.

lucaswoj:
Another design that may address the same problem without requiring the reintroduction of imperative concepts is something akin to React's [`key` attribute](https://facebook.github.io/react/docs/lists-and-keys.html#keys) -- some developer-provided metadata that hints to GL JS when a source can be copied.

anandthakker:
[...]

PR #6263, opened 2018-03-02
Arbitrary states for feature interactivity

asheemmamoowala:
Based on @ansis' work in [`data-update-proof-of-concept`](https://github.com/mapbox/mapbox-gl-js/commit/1bdd3130a39a386b11a78f69ef360af8ea24c881)

Implements `map.setFeatureState()`, a new method to set `key: value` pairs on specific features in a source. The source must provide unique feature id's per source-layer for this to work correctly.

Also implemented `['state', '<stateName>']` expression syntax. Similar to `get`, this allows accessing a feature's state instead of source data property. 

## todo
 - [x] Apply state to all symbol instances
 - [ ] ~Improve [ExpressionCreate](https://bl.ocks.org/anonymous/raw/f82c913b555326fd44b16d1864b7a23c) benchmark~
 - [x] Implement feature id checking per source-layer
 - [x] Only evaluate state-dependent layers 
     - [x] Conditionally enable `dynamicDraw` on vertex and index buffers
- [ ] Don't reparse expressions (#6255)

## Launch Checklist
 - [x] post benchmark scores
 - [x] create debug page

cc @ansis @mollymerp


src/data/bucket.js
```diff
@@ -89,6 +93,7 @@ module.exports = {
             // look up StyleLayer objects from layer ids (since we don't
             // want to waste time serializing/copying them from the worker)
             (bucket: any).layers = layers;
+            (bucket: any).stateDependentLayers = layers.filter((l) => l.isStateDependent());
```
asheemmamoowala: Pre-compute the layers with state-dependent expressions during deserialize. These are needed to re-compute the expressions in `ProgramConfiguration` until expression evaluators can be re-created when deserialized (#6255)
After that change, layer ids will be sufficient

src/style/style.js
```diff
@@ -780,6 +780,23 @@ class Style extends Evented {
         return this.getLayer(layer).getPaintProperty(name);
     }
 
+    setFeatureState(source: string, feature: string, key: string, value: any, sourceLayer?: string) {
+        if (this.sourceCaches[source] === undefined) {
+            throw new Error('There is no source with this ID');
```
mollymerp: for a more helpful error message, maybe print the passed `source` id?

src/style/style.js
```diff
@@ -780,6 +780,23 @@ class Style extends Evented {
...
+        }
+
+        const sourceCache = this.sourceCaches[source];
+        sourceCache.setFeatureState(feature, key, value, sourceLayer);
+    }
+
+    getFeatureState(source: string, feature: string, key?: string, sourceLayer?: string) {
+        if (this.sourceCaches[source] === undefined) {
+            throw new Error('There is no source with this ID');
```
mollymerp: same comment about error as above

src/source/source_cache.js
```diff
@@ -52,6 +54,7 @@ class SourceCache extends Evented {
     transform: Transform;
     _isIdRenderable: (id: number) => boolean;
     used: boolean;
+    _state: SourceFeatureState
```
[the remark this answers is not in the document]
anandthakker: I think it does make sense to _store_ a `SourceFeatureState` here on the `SourceCache`, since it belongs to a source and should be discarded precisely when a source is discarded.

But I agree that the state-aware _logic_ in below feels a bit out of place here.  What if we:
 - move the responsibility for calling `Tile#updateFeatureState` during `prepare()` directly into `SourceFeatureState#coalesceChanges()` (i.e., take a set of tiles as a parameter to `coalesceChanges`)?
 - Add a `SourceFeatureState#initializeTileState(Tile)` method and replace the other two cases of `tile.updateFeatureState(this._state.state);` with `this._state.initializeTileState(tile)`?

src/style-spec/expression/definitions/index.js
```diff
@@ -180,6 +180,15 @@ CompoundExpression.register(expressions, {
...
     },
+    'state': {
+        type: ValueType,
+        overloads: [
+            [
+                [StringType],
+                (ctx, [key]) => get(key.evaluate(ctx), ctx.state())
+            ]
+        ]
+    },
```
anandthakker: Since there's only one implementation, use:
```js
ValueType,
[StringType],
(ctx, [key]) => get(key.evaluate(ctx), ctx.state())
```

src/style-spec/expression/is_constant.js
```diff
@@ -8,6 +8,8 @@ function isFeatureConstant(e: Expression) {
     if (e instanceof CompoundExpression) {
         if (e.name === 'get' && e.args.length === 1) {
             return false;
+        } else if (e.name === 'state' && e.args.length === 1) {
```
anandthakker: Just `e.name === 'state'`, no need to check argument count

src/style-spec/expression/is_constant.js
```diff
@@ -28,6 +30,19 @@ function isFeatureConstant(e: Expression) {
     return result;
 }
 
+function isStateConstant(e: Expression) {
+    if (e instanceof CompoundExpression) {
+        if (e.name === 'state' && e.args.length === 1) {
```
anandthakker: No need to check `e.args.length`

src/ui/map.js
```diff
@@ -1339,6 +1339,34 @@ class Map extends Camera {
...
+     * 
+     * @param {string} source The source identifier for the feature.
+     * @param {string} feature The unique feature identifier from the source data.
+     * @param {string} key The name of the state to be set.
+     * @param {string | number | boolean} value The value to be set.
+     * @param {string} sourceLayer (optional) The source-layer identifier for Vector Tile sources.
+     */
+    setFeatureState(source: string, feature: string, key: string, value: any, sourceLayer?: string) {
+        this.style.setFeatureState(source, feature, key, value, sourceLayer);
+        this._rerender();
```
anandthakker: Use `this._update()`

src/source/source_cache.js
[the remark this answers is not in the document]
asheemmamoowala: > are there any alternatives to storing state for all features on all loaded tiles in the `SourceCache`?

Since `SourceCache` is responsible for creating sources and managing their tiles, it is the appropriate place to manage the feature state of the source as well. 
@anandthakker, that same reasoning is what led me to make `SourceFeatureState` a data-only class and have `SourceCache`  continue to be responsible for managing tiles and their state. 

How about adding `SourceCache#initializeTile`, called for newly loaded tiles, or tiles fetched from the cache? It would update feature state and backfill raster DEM tiles.

src/data/bucket/circle_bucket.js
```diff
@@ -70,20 +73,33 @@ class CircleBucket<Layer: CircleStyleLayer | HeatmapStyleLayer> implements Bucke
...
-                this.addFeature(feature, geometry);
+                this.addFeature(feature, geometry, index);
                 options.featureIndex.insert(feature, geometry, index, sourceLayerIndex, this.index);
             }
         }
     }
 
+    update(states: FeatureStates, vtLayer: VectorTileLayer) {
+        if (!this.stateDependentLayers.length) return;
+        this.changed = this.programConfigurations.updatePaintArrays(states, vtLayer, this.stateDependentLayers);
```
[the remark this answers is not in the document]
anandthakker: > If we follow that idea, the correct place to track changes would be in the individual ProgramConfiguration's

We can put a `dirty` property on `ProgramConfigurationSet`, which is set during `updatePaintArrays`, checked in `upload()` to decide whether to actually do work (and then cleared).  I suppose it might not be simpler per se, but it has the advantage of localizing and centralizing this state tracking in `program_configuration.js`, rather than having it replicated in each bucket.

src/source/source_cache.js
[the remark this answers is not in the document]
anandthakker: > Since SourceCache is responsible for creating sources and managing their tiles, it is the appropriate place to manage the feature state of the source as well

I disagree.  `SourceCache` is already very difficult to reason about, in part because it's doing too much. (See also https://github.com/mapbox/mapbox-gl-js/issues/2291.)  Ideally, it would not manipulate the "content" of its `Tile`s at all (except perhaps setting `tile.state`, since that's directly related to its job of caching tiles).

src/data/program_configuration.js
```diff
@@ -316,10 +371,50 @@ class ProgramConfiguration {
...
+            const featureId = String(feature.id);
+            this._idMap[featureId] = this._idMap[featureId] || [];
+            this._idMap[featureId].push({
+                index: index,
+                start: this._bufferPos,
+                length: length
+            });
+        }
+
+        this._bufferPos = length;
```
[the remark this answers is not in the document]
anandthakker: Oh, yeah that's confusing!  Maybe we should rename the `length` parameter to `targetLength` or something?

src/source/source_state.js
```diff
@@ -0,0 +1,74 @@
+// @flow
+const util = require('../util/util');
+const Tile = require('./tile');
+
+export type FeatureStates = {[feature_id: string]: {[key: string]: string | number | boolean }};
```
lucaswoj: Do we intend to only allow `string | number | boolean` in feature states? Aren't GeoJSON properties allowed to include `null`, arrays, and objects now?

src/data/bucket/symbol_bucket.js
```diff
@@ -414,16 +418,28 @@ class SymbolBucket implements Bucket {
         }
     }
 
+    update(states: FeatureStates, vtLayer: VectorTileLayer) {
```
lucaswoj: I would assume that a method called `Bucket#update` did something more general than this (like update all the data in the bucket). Would a name like `Bucket#updateFeatureState` be more clear?

src/data/bucket/symbol_bucket.js
```diff
@@ -169,15 +169,17 @@ class SymbolBuffers {
...
-        this.indexBuffer = context.createIndexBuffer(this.indexArray, dynamicIndexBuffer);
+    upload(context: Context, dynamicIndexBuffer: boolean, update?: boolean) {
+        if (!update) {
+            this.layoutVertexBuffer = context.createVertexBuffer(this.layoutVertexArray, symbolLayoutAttributes.members);
+            this.indexBuffer = context.createIndexBuffer(this.indexArray, dynamicIndexBuffer);
+            this.dynamicLayoutVertexBuffer = context.createVertexBuffer(this.dynamicLayoutVertexArray, dynamicLayoutAttributes.members, true);
+            this.opacityVertexBuffer = context.createVertexBuffer(this.opacityVertexArray, shaderOpacityAttributes, true);
+            // This is a performance hack so that we can write to opacityVertexArray with uint32s
+            // even though the shaders read uint8s
+            this.opacityVertexBuffer.itemSize = 1;
```
lucaswoj: Can you explain how this works in more detail? Is this prone to break in the future?

ChrisLoer: Yeah, this is a relatively brittle performance hack. We might be able to make it more future proof by pushing the hack into `StructArray` so the external interface was always clearly "this is an array of uint8s" even if uint32s were used internally. But it's a weird special case and it's not clear to me how we'd define that for `StructArray`.

Background: we do _lots_ of opacity vertex updates for symbols: for each symbol, we have one packed opacity value that we need to update on all four vertices for every glyph. Profiling in Chrome, the actual cost of building the array showed up as a significant part of rendering time. Pushing n/4 uint32s turned out to be significantly faster than pushing n uint8s (see https://github.com/mapbox/mapbox-gl-js/pull/5150#issuecomment-327880914, there's an additional wrinkle that the same change included packing "target opacity" into the same value). At some point it will probably make sense to re-test this optimization, since testing this kind of thing is so squirrelly and changes in the javascript engine could render it useless.

src/data/bucket/symbol_bucket.js
asheemmamoowala: `Bucket#update` updates the dynamic buffers in the bucket with the passed in `states`. The layout and index buffers are static, and so cannot be updated. The more specific name is more descriptive, but it doesn't distinguish itself from any other type of `update`.

src/data/bucket/symbol_bucket.js
lucaswoj: > We might be able to make it more future proof by pushing the hack into `StructArray`

I think that's a good approach 👍

src/ui/map.js
```diff
@@ -1339,6 +1339,34 @@ class Map extends Camera {
...
+    /**
+     * Sets state values on a feature, given a source and [optional]source-layer.
+     * 
+     * @param {string} source The source identifier for the feature.
+     * @param {string} feature The unique feature identifier from the source data.
+     * @param {string} key The name of the state to be set.
+     * @param {string | number | boolean} value The value to be set.
+     * @param {string} sourceLayer (optional) The source-layer identifier for Vector Tile sources.
+     */
+    setFeatureState(source: string, feature: string, key: string, value: any, sourceLayer?: string) {
```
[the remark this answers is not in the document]
asheemmamoowala: This seems like an improvement. I'd like to make the API error out if a source layer is not provided for Vector sources. It would mean checking the type of the source id, but it would prevent erroneous use of the API.

src/source/source_state.js
asheemmamoowala: I think everything but `object` is OK to use. Nested objects are not supported in expressions, so it wouldn't be usable for runtime styling, and possibly cause confusion.

lucaswoj: @anandthakker Are nested objects supported in GeoJSON feature properties? I suspect the least confusing thing will be parity between GeoJSON feature properties and feature state.

anandthakker: Yep, they're supported.  E.g., the following works:

```js
map.on('load', () => {
    map.addLayer({
        id: 'test',
        source: {
            type: 'geojson',
            data: {
                type: 'Feature',
                properties: {
                    x: { y: 'hello' }
                },
                geometry: {
                    type: 'Point',
                    coordinates: [0, 0]
                }
            }
        },
        type: 'symbol',
        layout: {
            'text-field': ["get", "y", ["get", "x"]]
        }
    })
})
```

anandthakker: Although note that in `queryRenderedFeatures`, array or object property values currently get stringified.

asheemmamoowala: Thanks for the correction @anandthakker! In that case, feature states should accept all JSON types. 
For `queryRenderedFeatures`, this means that arrays and objects would be stringified in `properties` but not in `state`.

anandthakker: Yeah, that's fine -- qRF _shouldn't_ stringify them, it's just a limitation of a current implementation detail: https://github.com/mapbox/mapbox-gl-js/issues/2434

src/data/program_configuration.js
```diff
@@ -129,31 +140,49 @@ class SourceExpressionBinder<T> implements Binder<T> {
...
+            const color = packColor(value);
+            for (let i = start; i < end; i++) {
+                paintArray.emplace(i, color[0], color[1]);
+            }
+        } else {
+            for (let i = start; i < end; i++) {
+                paintArray.emplace(i, value);
+            }
+
+            this.statistics.max = Math.max(this.statistics.max, value);
```
ansis: This doesn't handle the case where the maximum value is replaced with a smaller value and the max should decrease. I think this is fine though, because having a bigger max won't cause bugs the way we use it and maintaining the real max value would be more complicated and expensive. This should be fine and safe the way it is, right?

src/ui/map.js
```diff
@@ -1341,6 +1341,34 @@ class Map extends Camera {
...
+     * Gets the state of a feature, given a source and [optional]source-layer.
+     * 
+     * @param {string | {sourceId: string, sourceLayer: string}} source *For vector tile sources, the sourceLayer is
+     *  required.* For GeoJSON sources, just the source id.
+     * @param {string} feature The unique feature identifier from the source data.
+     * @param {string} key (optional) The name of the state. When null, all state values are returned.
+     * 
+     * @returns {any} The value of the specified specified state or all states.
+     */
+    getFeatureState(source: string | { sourceId: string; sourceLayer: string; }, feature: string, key?: string): any {
```
ansis: The optional `key` param changes the return type. Should returning the full state object be split into a separate method? If it's useful to return the entire state as an object, should it be possible to set state with an object instead of individual calls for each key?

src/data/program_configuration.js
```diff
@@ -367,18 +460,27 @@ export default class ProgramConfiguration {
...
         for (const key in this.programConfigurations) {
-            this.programConfigurations[key].populatePaintArrays(length, feature);
+            this.programConfigurations[key].populatePaintArrays(length, feature, index);
+        }
+        this.needsUpload = true;
+    }
+
+    updatePaintArrays(featureStates: FeatureStates, vtLayer: VectorTileLayer, layers: $ReadOnlyArray<TypedStyleLayer>) {
+        for (const layer of layers) {
+            this.needsUpload =  this.needsUpload || this.programConfigurations[layer.id].updatePaintArrays(featureStates, vtLayer, layer);
```
ansis: If `this.needsUpload` is `true` before is gets called then the `||` will short-circuit and `this.programConfigurations[layer.id].updatePaintArrays(...)` will not get called. It should always get called, right?

src/source/source_state.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    coalesceChanges(tiles: {[any]: Tile}) {
+        const changes: LayerFeatureStates = {};
+        for (const sourceLayer in this.stateChanges) {
+            this.state[sourceLayer]  = this.state[sourceLayer] || {};
+            const layerStates = {};
+            for (const id in this.stateChanges[sourceLayer]) {
+                this.state[sourceLayer][id] = extend(
+                                                {},
```
ansis: Is it necessary to create and extend a new object? or could it just extend the existing `this.state[sourceLayer][id]`?

src/data/program_configuration.js
ansis: Also, the max changing means the cached `queryPadding` value might be invalid: https://github.com/mapbox/mapbox-gl-js/blob/bd5a65380b1ffb961bd2e3173b0b96f8aa445f8f/src/source/tile.js#L170-L174

I think we need some way of either invalidating it, or if it is cheap enough, recalculating it every time we need to use it instead of caching it.

src/data/program_configuration.js
asheemmamoowala: This is a great catch! I didn't run into the short-circuiting, but it is possible.

src/data/program_configuration.js
asheemmamoowala: Keeping track of the `max` value correctly may not be worth it. When a property value is reduced, the update logic would need to filter the previously populated values to check if there is a new lower `max`. 

Leaving this as is won't return incorrect results, but it will increase the number of features that are matched in the initial query (with `queryPadding`) and then have to be filtered out. 

If there are use cases that significantly affect the performance of `queryRenderedFeatures`, this can be revisited.

asheemmamoowala: Added this in 
https://github.com/mapbox/mapbox-gl-js/blob/25d23f946bbb65329fcbc8c5d07dc3ca8bb68717/src/source/tile.js#L435-L438

`Tile.queryPadding` is updated after each` bucket update.

lucaswoj:
I haven't been this excited about a new feature in a loooong time!

ansis:
This looks good to me. The only remaining question I have are details about the external interface. The current interface seems fine but I think it's worth looking at this extra carefully since it will be public.

The current public signature is:
```
setFeatureStateValue(
    source: string | { sourceId: string; sourceLayer: string; },
    feature: string,
    key: string,
    value: any)
```

`setFeatureStateValue` seems like a good name, but would `setFeatureState` be just as clear?

What if the method signature was:
```
setFeatureStateValue(
    feature: {source: string; sourceLayer: string; id: string},
    key: string,
    value: any)
```
The advantages I'm seeing:
- (very minor) the first argument is always an object
- if we set `source` and `sourceLayer` on geojson features returned by `queryRenderedFeatures` and on the features event handlers are called with we could pass those features directly to this method:
`setFeatureStateValue(feature, 'key', 'value')` instead of
`setFeatureStateValue({ source: 'id', sourceLayer: 'layerId' }, feature.id, 'key', 'value')`.

---

@lucaswoj @ChrisLoer @anandthakker @mollymerp @jfirebaugh did any of you want to review this any further?

asheemmamoowala:
> queryPadding possibly being invalid if the statistics.max value changed

@ansis I opened #6536 to make `queryPadding` computation on-demand.

asheemmamoowala:
> Should there be a method that allows updating multiple state values at once? E.g., updateFeatureState(feature, { key1: 'value1', key2: 'value2' })

@anandthakker  Were you thinking of this as a replacement for `setFeatureStateValue`, or in addition to it?

asheemmamoowala:
> if we set source and sourceLayer on geojson features returned by queryRenderedFeatures and on the features event handlers are called with we could pass those features directly to this method:
setFeatureStateValue(feature, 'key', 'value') instead of
setFeatureStateValue({ source: 'id', sourceLayer: 'layerId' }, feature.id, 'key', 'value').

@ansis The `source` and `source-layer` properties are already available under the `layer`  property of a feature object returned from the query methods. Only issue I see with this is that it would make it less convenient to create 'feature' objects when they are not already returned from some existing API (data join may be an example of that)

asheemmamoowala:
@anandthakker `updateFeatureState ` breaks from the convention of `get`/`set` methods used on `Map`.  I do like the idea of a single set/update method for multiple key/value pairs though.

lucaswoj:
The original API design, `setFeatureState(feature, key, value)`, is much more consistent with the rest of our APIs, such as `setPaintProperty` and `["feature-state"]`. If we are going to adopt a new naming paradigm, let's do it separately and everywhere. 

I don't have a strong opinion about whether or not we also allow `setFeatureState(feature, values)`.

https://github.com/mapbox/mapbox-gl-js/issues/2741 is about internal architecture not naming conventions.

anandthakker:
I see this not as a new naming convention, but instead a different behavior for which there isn't an existing precedent in our existing APIs.  There's a broader convention that `setXxxx()` methods are merely setters, which would lead me to think that `setFeatureState(feature, {p1: v1})` would _replace_ the given feature's state with `{p1: v1}`, rather than merging `{p1: v1}` into the existing state.

jfirebaugh:
React's `setState` does a merge. I think the name we had was fine.

anandthakker:
Another inconsistency: our own `setStyle()` does _not_ do a merge. See also https://github.com/mapbox/mapbox-gl-js/issues/4006

asheemmamoowala:
<details>
<summary> Benchmark results </summary>

screencapture-localhost-9966-bench-2018-04-19-21_35_24

screencapture-localhost-9966-bench-2018-04-19-21_35_24-2
</details>

thunderkid:
This looks like nice technology, but I'm trying to understand exactly when it should be used. It's generally been encouraged to have a separate layer for features that change rapidly (eg a layer that contains only a copy of the hovered item but with some different styling.) That approach is pretty performant, and as long as the hovered item obscures the original it works well. So when should the separate hover-layer approach be used and when should one use a single layer with `setFeatureState`? Are there any general guidelines?

lucaswoj:
> So when should the separate hover-layer approach be used and when should one use a single layer with `setFeatureState`? Are there any general guidelines?

`setFeatureState` was designed specifically  with the hover use case in mind. I would not recommend using the "hover layer approach" now that `setFeatureState` has been implemented.

The only case in which you may want to use a second layer is if the _geometry_ of a feature needs to change rapidly (i.e. for a drawing plugin) as `setFeatureState` cannot change the feature's geometry.

shayke:
I'm not sure if there is a bug in this feature or I'm using it wrong
but please take a look at those (hopefully) equivalent examples:
1. Using setFeatureState - http://jsfiddle.net/01b97w0L/672/
2. Using extra layer - http://jsfiddle.net/01b97w0L/673/

The idea is to selected all "related" features to the selected one and color them as red.
If you click on the left most line and zoom out
1. setFeatureState - Doesn't color the zoomed out features
2. "extra layer" - properly colors *every* feature on the map (regardless of zoom level)

ansis:
@shayke your version using `setFeatureState` isn't working because `querySourceFeatures` only returns features in the current viewport. This might be something we should fix for geojson sources where we do have all the features. In the mean time, you could filter the source geojson data yourself to get all the ids of features with the correct county

Amit-Gore:
@lucaswoj Does setFeatureState always require feature.id? Can it work on feature.properties.name?

asheemmamoowala:
@Amit-Gore Currently `setFeatureState` does require the top-level `feature.id` attribute. See #6019 for a discussion on future improvements to this.

Issue #4136, opened 2017-02-01
Case and diacritic folding for expression operators


lucaswoj:
_From @1ec5 on October 21, 2016 10:26_

As discussed in mapbox/mapbox-gl-native#6781, it can often be necessary for a filter to perform a case-insensitive or diacritic-insensitive comparison. Variations on the existing operators should be added to support these options.

Here’s an example design (but by no means the best):

``` json
[
  {
    "operator": "in",
    "case-sensitive": false,
    "diacritic-sensitive": false,
  },
  "key",
  "value 1",
  "value 2"
]
```

Implementing case-insensitive comparisons should be trivial on all the platforms supported by Mapbox GL. On the other hand, while there are fine options for diacritic folding on the native platforms, JavaScript would have to rely on a library for diacritic-insensitive comparisons.

/cc @incanus @lucaswoj @jfirebaugh

_Copied from original issue: mapbox/mapbox-gl-style-spec#548_

mollymerp:
TIL what diacritic means

incanus:
_Everyone's a diacritic._

1ec5:
Over in https://github.com/mapbox/mapbox-gl-js/pull/4715#discussion_r117343669, we’re discussing a potential syntax for equality comparisons in expressions. Although this issue talks about filters, getting case and diacritic folding into expressions would address some of the main use cases for this functionality (for bilingual labeling in particular, but not for highlighting “fuzzy-matched” search results).

https://github.com/mapbox/mapbox-gl-js/pull/4715#discussion_r117343669 asks whether we should incorporate case and diacritic folding into the proposal as some sort of modifier on equality comparisons, or whether we should rely on the style author to run each operand through functions like `lowercase` or `strip-diacritics` beforehand.

Is a case- and diacritic-folded equality comparison equivalent to an equality comparison between two normalized strings, or are there aspects to folding that require both original strings as context? To put it another way, are there strings that should be considered equal after case- and diacritic-folding but would differ after each string is run through `lowercase` and `strip-diacritics`?

1ec5:
> are there strings that should be considered equal after case- and diacritic-folding but would differ after each string is run through `lowercase` and `strip-diacritics`?

[...]

Issue #6197, opened 2018-02-20
Add expression operators for locale matching (system languages)


1ec5:
There should be a simple way for the style author to specify that a `text-field` should be set to the `name_*` feature property that best fits the system’s preferred languages. Secondarily, it would be great if the most appropriate locale could be used on its own in expressions.

## Motivation

[Localizing a style’s labels](https://www.mapbox.com/help/change-language/) currently entails iterating over all the layers, manually replacing references to `name_*` feature properties within each `text-field` value. If these values are expressions, replacing the references can be an involved, recursive step. The iOS and macOS map SDKs have a built-in option, `MGLStyle.localizesLabels`, that applies these changes automatically based on the system language and region preferences. There’s [a plugin for GL JS](https://github.com/mapbox/mapbox-gl-language/) and a forthcoming plugin for the Android map SDK (mapbox/mapbox-plugins-android#74) that do likewise.

While this approach is effective, it operates at such a high level that the localizing code doesn’t have a good way to reason about the style author’s intentions. Should `{name} ({name_en})` be replaced by `{name_es} ({name_en})` or just `{name}`? The style’s author has no opportunity to react to changes that could radically alter the style’s appearance, for instance by increasing the font size when the system language is Chinese. Moreover, the localization feature implicitly opts the map into runtime styling–specific behaviors like disabling automatic style refreshes.

## Design

The style specification would be extended with two expression operators:

* `user-locales` takes no arguments and evaluates to an array of locale identifiers corresponding to the user’s preferences.
* `match-locales` has the signature `["match-locales", inputLocales, availableLocales]` and evaluates to the item in `availableLocales` (an unordered array of locale identifiers) that corresponds to the first item in `inputLocales` (e.g., `user-locales`) that matches one of `availableLocales`.

For the purposes of these operators, a locale identifier could include a language code, script code, or region code, or some combination thereof. I would be in favor of specifying BCP 47 as the locale identifier standard to follow.

In typical usage, a style author would opt into localization by setting `text-field` to a value such as:

```json
[
  "let",
  "streets-languages", ["ar", "de", "en", "es", "fr", "pt", "ru", "zh", "zh-Hans"],
  [
    "coalesce",
    ["concat", "name_", ["match-locales", ["user-locales"], ["var", "streets-languages"]]],
    "name"
  ]
]
```

[...]

PR #6270, opened 2018-03-02
'Collator' expression for controlling case and diacritic sensitivity in string comparisons

ChrisLoer:
Potential implementation of a fix for https://github.com/mapbox/mapbox-gl-js/issues/4136.

For "locale", we could:

- Get rid of the argument, and always use the runtime's default locale. Advantage: easiest from the point of view of the expression author.
- Allow a single optional string as an argument (currently implemented). Advantage: not _too_ complicated, and allows basic localization (e.g. "let's use German collation rules for this map").
- Allow array of locale strings as an argument. Advantage: allows definition of fallback rules when the set of locally supported locales isn't known ahead of time. Disadvantage: more cumbersome to specify.

It does feel odd to have this operator implemented separately from `==`,`!=`,`<`,`<=`,`>`, and`>=`, (and also `filter-*-in`) *but*:

- The equality operator (along with `!`) gives us what we need for the specific use case of changing map behavior when two values are "near duplicates"
- Greater/less than operators probably (?) make more sense in the context of being able to specify general purpose collation rules.

An alternative approach might be to make `Equals`, `lt`, `gt`, etc. all take an optional "Collation" as an argument and then add expression support for defining a collation. One risk of this approach is that the more customizability we expose the harder it'll be to maintain consistent behavior across platforms (although we expect the underlying implementations to _usually_ be based on some form of ICU).

/cc @1ec5 @anandthakker @kkaefer @nickidlugash

### DIFF
--- a/src/style-spec/expression/definitions/collator.js
+++ b/src/style-spec/expression/definitions/collator.js
@@ -1,0 +1,130 @@
+// @flow
+
+import { StringType, BooleanType, CollatorType } from '../types';
+
+import type { Expression } from '../expression';
+import type EvaluationContext from '../evaluation_context';
+import type ParsingContext from '../parsing_context';
+import type { Type } from '../types';
+
+declare var Intl: {
+  Collator: Class<Collator>
+}
+
+declare class Collator {
+  constructor (
+    locales?: string | string[],
+    options?: CollatorOptions
+  ): Collator;
+
+  static (
+    locales?: string | string[],
+    options?: CollatorOptions
+  ): Collator;
+
+  compare (a: string, b: string): number;
+
+  resolvedOptions(): any;
+}
+
+type CollatorOptions = {
+  localeMatcher?: 'lookup' | 'best fit',
+  usage?: 'sort' | 'search',
+  sensitivity?: 'base' | 'accent' | 'case' | 'variant',
+  ignorePunctuation?: boolean,
+  numeric?: boolean,
+  caseFirst?: 'upper' | 'lower' | 'false'
+}
+
+export class CollatorInstantiation {
+    locale: string | null;
+    sensitivity: 'base' | 'accent' | 'case' | 'variant';
+
+    constructor(caseSensitive: boolean, diacriticSensitive: boolean, locale: string | null) {
+        if (caseSensitive)
+            this.sensitivity = diacriticSensitive ? 'variant' : 'case';
+        else
+            this.sensitivity = diacriticSensitive ? 'accent' : 'base';
+
+        this.locale = locale;
+    }
+
+    compare(lhs: string, rhs: string): number {
+        return new Intl.Collator(this.locale ? this.locale : [],
+                                 { sensitivity: this.sensitivity, usage: 'search' })
+            .compare(lhs, rhs);
+    }
+
+    resolvedLocale(): string {
+        return new Intl.Collator(this.locale ? this.locale : [])
+            .resolvedOptions().locale;
+    }
+
+    serialize() {
+        const serialized = ["collator"];
+        serialized.push(this.sensitivity === 'variant' || this.sensitivity === 'case');
+        serialized.push(this.sensitivity === 'variant' || this.sensitivity === 'accent');
+        if (this.locale) {
+            serialized.push(this.locale);
+        }
+        return serialized;
+    }
+}
+
+export class CollatorExpression implements Expression {
+    type: Type;
+    caseSensitive: Expression;
+    diacriticSensitive: Expression;
+    locale: Expression | null;
+
+    constructor(caseSensitive: Expression, diacriticSensitive: Expression, locale: Expression | null) {
+        this.type = CollatorType;
+        this.locale = locale;
+        this.caseSensitive = caseSensitive;
+        this.diacriticSensitive = diacriticSensitive;
+    }
+
+    static parse(args: Array<mixed>, context: ParsingContext): ?Expression {
+        if (args.length !== 3 && args.length !== 4)
+            return context.error(`Expected two or three arguments.`);
+
+        const caseSensitive = context.parse(args[1], 1, BooleanType);
+        if (!caseSensitive) return null;
+        const diacriticSensitive = context.parse(args[2], 2, BooleanType);
+        if (!diacriticSensitive) return null;
+
+        let locale = null;
+        if (args.length === 4) {
+            locale = context.parse(args[3], 3, StringType);
+            if (!locale) return null;
+        }
+
+        return new CollatorExpression(caseSensitive, diacriticSensitive, locale);
+    }
+
+    evaluate(ctx: EvaluationContext) {
+        return new CollatorInstantiation(this.caseSensitive.evaluate(ctx), this.diacriticSensitive.evaluate(ctx), this.locale ? this.locale.evaluate(ctx) : null);
+    }
+
+    eachChild(fn: (Expression) => void) {
+        fn(this.caseSensitive);
+        fn(this.diacriticSensitive);
+        if (this.locale) {
+            fn(this.locale);
+        }
+    }
+
+    possibleOutputs() {
+        // Technically the set of possible outputs is the combinatoric set of Collators produced
+        // by all possibleOutputs of locale/caseSensitive/diacriticSensitive
+        // But for the primary use of Collators in comparison operators, we ignore the Collator's
+        // possibleOutputs anyway, so we can get away with leaving this undefined for now.
+        return [undefined];
+    }
+
+    serialize() {
+        const serialized = ["collator"];
+        this.eachChild(child => { serialized.push(child.serialize()); });
+        return serialized;
+    }
+}
--- a/src/style-spec/expression/definitions/collator.js
+++ b/src/style-spec/expression/definitions/collator.js
@@ -61,13 +61,13 @@ export class CollatorInstantiation {
     }
 
     serialize() {
-        const serialized = ["collator"];
-        serialized.push(this.sensitivity === 'variant' || this.sensitivity === 'case');
-        serialized.push(this.sensitivity === 'variant' || this.sensitivity === 'accent');
+        const options = {};
+        options['caseSensitive'] = this.sensitivity === 'variant' || this.sensitivity === 'case';
+        options['diacriticSensitive'] = this.sensitivity === 'variant' || this.sensitivity === 'accent';
         if (this.locale) {
-            serialized.push(this.locale);
+            options['locale'] = this.locale;
         }
-        return serialized;
+        return ["collator", options];
     }
 }
 
@@ -85,17 +85,24 @@ export class CollatorExpression implements Expression {
     }
 
     static parse(args: Array<mixed>, context: ParsingContext): ?Expression {
-        if (args.length !== 3 && args.length !== 4)
-            return context.error(`Expected two or three arguments.`);
+        if (args.length !== 2)
+            return context.error(`Expected one argument.`);
 
-        const caseSensitive = context.parse(args[1], 1, BooleanType);
+        const options = (args[1]: any);
+        if (typeof options !== "object" || Array.isArray(options))
+            return context.error(`Collator options argument must be an object.`);
+
+        const caseSensitive = context.parse(
+            options['caseSensitive'] === undefined ? false : options['caseSensitive'], 1, BooleanType);
         if (!caseSensitive) return null;
-        const diacriticSensitive = context.parse(args[2], 2, BooleanType);
+
+        const diacriticSensitive = context.parse(
+            options['diacriticSensitive'] === undefined ? false : options['diacriticSensitive'], 2, BooleanType);
         if (!diacriticSensitive) return null;
 
         let locale = null;
-        if (args.length === 4) {
-            locale = context.parse(args[3], 3, StringType);
+        if (options['locale']) {
+            locale = context.parse(options['locale'], 3, StringType);
             if (!locale) return null;
         }
 
@@ -123,8 +130,12 @@ export class CollatorExpression implements Expression {
     }
 
     serialize() {
-        const serialized = ["collator"];
-        this.eachChild(child => { serialized.push(child.serialize()); });
-        return serialized;
+        const options = {};
+        options['caseSensitive'] = this.caseSensitive;
+        options['diacriticSensitive'] = this.diacriticSensitive;
+        if (this.locale) {
+            options['locale'] = this.locale;
+        }
+        return ["collator", options];
     }
 }
--- a/src/style-spec/expression/definitions/collator.js
+++ b/src/style-spec/expression/definitions/collator.js
@@ -7,20 +7,23 @@ import type EvaluationContext from '../evaluation_context';
 import type ParsingContext from '../parsing_context';
 import type { Type } from '../types';
 
+// Flow type declarations for Intl cribbed from
+// https://github.com/facebook/flow/issues/1270
+
 declare var Intl: {
-  Collator: Class<Collator>
+  Collator: Class<Intl$Collator>
 }
 
-declare class Collator {
+declare class Intl$Collator {
   constructor (
     locales?: string | string[],
     options?: CollatorOptions
-  ): Collator;
+  ): Intl$Collator;
 
   static (
     locales?: string | string[],
     options?: CollatorOptions
-  ): Collator;
+  ): Intl$Collator;
 
   compare (a: string, b: string): number;
 
@@ -36,9 +39,10 @@ type CollatorOptions = {
   caseFirst?: 'upper' | 'lower' | 'false'
 }
 
-export class CollatorInstantiation {
+export class Collator {
     locale: string | null;
     sensitivity: 'base' | 'accent' | 'case' | 'variant';
+    collator: Intl$Collator;
 
     constructor(caseSensitive: boolean, diacriticSensitive: boolean, locale: string | null) {
         if (caseSensitive)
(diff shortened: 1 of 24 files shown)


ChrisLoer:
Hmm, the expression test failure points out how easy it will be to shoot yourself in the foot authoring styles that depend on locally installed locales (I wrote the test with the "de" locale present on my machine, but apparently it's not installed on our CI image).

Browser support may also be an issue (https://caniuse.com/#search=Intl.Collator), although it looks pretty widespread, and I think it's probably reasonable to just fall back to plain equality when we don't have browser support.

jfirebaugh:
If we were to provide a more comprehensive set of collation-related expression operators:

* What else would we likely want to provide?
* Would `diacritic-insensitive-equal` extend neatly into a design that encompasses these additional features, or would it tend to become an API wart?

ChrisLoer:
I think if we had a `Collator` type that included:

 - Locale (string)
 - case-sensitivity (bool)
 - diacritic-sensitivity (bool)

We could make that an optional third argument to the equality/comparison operators (I'm not sure how awkward it would be to tack on an optional argument there?). If we had that, `diacritic-insensitive-equal` would just be a left-over wart, I think.

`Intl.Collator` exposes some features that could potentially be useful (like "numeric" sorting or controlling whether to sort lower or upper case first), although they wouldn't be relevant for equality checks. We could add them to our Collator in the future, if we knew a consistent way to implement them (I don't think there's a direct analog to those exposed by `NSString compare`).

test/integration/expression-tests/diacritics/locale/test.json
```diff
@@ -1,5 +1,5 @@
 {
-  "expression": ["diacritic-insensitive-equal", "de", ["get", "lhs"], ["get", "rhs"]],
+  "expression": ["==", ["string",["get", "lhs"]], ["get", "rhs"], ["collator", true, false, "de"]],
```
anandthakker: Maybe we should consider accepting an single object literal for collator options, e.g. `["collator", { caseSensitive: false, diacriticSensitive: false, locale: 'de' }]`.  It's not how any existing expressions work, but this is the first case where we really need a set of options, and a dictionary is much clearer than a positional arguments for this.

anandthakker: And while we're considering such a departure, we might as well ask: should we simply accept a collator options object directly as an optional fourth argument to collating expressions?  

```js
["==", ["string",["get", "lhs"]], ["get", "rhs"], { caseSensitive: false, diacriticSensitive: false, locale: 'de' }]
```

nickidlugash:
@ChrisLoer So excited to see explorations on this issue!

Implementing collation parameters together as one set of options seems to make sense, since 
 I think users will likely want to use case-insensitivity and diacritic-insensitivity at the same time (or at least, we'll want to in our core styles).

> What else would we likely want to provide?

From the js [intl.collator doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator), `ignorePunctuation` also seems like it could be useful. I don't know how support for this [varies across platforms](https://github.com/mapbox/mapbox-gl-js/issues/4136#issuecomment-369768348).

1ec5:
> From the js [intl.collator doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator), `ignorePunctuation` also seems like it could be useful. I don't know how support for this [varies across platforms](https://github.com/mapbox/mapbox-gl-js/issues/4136#issuecomment-369768348).

Ignoring punctuation is actually supported publicly by Qt, but not by iOS or macOS.

ChrisLoer:
OK, this branch currently has:

- `collator` expression that can create a collator based on locale/case-sensitivity/diacritic-sensitivity
- Support for `collator` in `==`,`!=`,`<`,`<=`,`>`,`>=`
- `["resolved-locale", collator]` to inspect if you're getting the locale you're asking for

Most of the tests are written to use the default locale. If they're run on a system where the default locale isn't `en` they may break. 🤔 I could make all of them specify `en` and count on that at least being installed everywhere they run? Unfortunately hard to test the "default" behavior without making tests that are bound to behave differently on different platforms.

There's one test that runs against the `de` locale, and it uses `resolved-locale` to change what it expects based on whether `de` is installed. This is weird because it means that it tests something different on CI (where it's not installed) vs. locally on OS X (where it is installed). I think this is OK? Although right now it assumes that if you don't have `de` you must be using `en`, probably better to just make it pass no matter what if you don't have `de`.

I haven't added collation support to the old filters, but it should always be possible to simply author a filter with a new expression to get the functionality, right?

Current style authors use `downcase`/`upcase` to do case-insensitive comparisons. This is probably a preferable alternative, but I'm not sure if we need to do anything to push existing users towards this.

Before merging this, I'd like to be at least relatively confident in the equivalent native behavior. Some issues:

- Uncertain support across platforms for `ignorePunctuation` seems to me like a good argument for holding off on that, at least for now.
- Currently "locale" is specified as a "BCP 47 language tag" because that's what `Intl.Collator` takes, but I'm worried this allows over-specification relative to the platform support we'll end up with. If we can get away with it, it would be nice to restrict the locale to a ISO 639-1 Code, as a kind of "least common denominator". That would however rule out specifying regional differences within languages. I can imagine a future in which we generate "locale" objects and use them for formatters (where region would matter more) as well as collation, but I don't want to get too far ahead of myself.
- On Qt, as far as I can tell we can do case-insensitive comparison or default-locale comparison but we can't combine them and we can't control the locale. [Some (old) discussion of the issue...](https://blog.qt.io/blog/2011/06/14/string-collation-with-locales/). We _do_ have a build of Qt that bundles ICU for Arabic support, but bundling collation data as well would significantly increase the build size.
- Where we end up depending on bundled ICU collation support (thankfully not iOS or Android), we'll have to choose what locales to bundle. Should we try to bundle a minimal set and leave it at that? Or expose it as a build time configuration option?

ChrisLoer:
I think this is ready to go, pending review.

The previously "default" tests now specify the "en" locale -- so the tests require "en" to be installed, but don't require it to be the default locale on the system. The "de" locale test is now hardwired to pass (without doing any comparisons) if "de" isn't installed.

@1ec5 has convinced me to stick with BCP 47 language tags as the interface for defining a locale, with the idea that we'll implement a BCP 47 parser in native and then use as much as we can based on the platform we're running on (e.g. we'd parse the "script" tag but ignore it on ICU-based systems).

@kkaefer thinks that where we're bundling ICU (linux, Qt), install size is not a major concern and we can get away with just including the entire resource bundle (so we can skip worrying about _which_ locales to bundle).

ChrisLoer:
> For the "variant" option, MDN docs state "other differences may also be taken into consideration." Do we know what those might be and whether we care about them?

At least for the Firefox implementation, "variant" maps to ICU's locale-dependent "tertiary strength":

https://dxr.mozilla.org/mozilla-central/source/js/src/builtin/intl/Collator.cpp#342-343

ICU has a further "IDENTICAL" option which is based directly on unicode code point equality -- the impression I get is that the difference is likely to be in whether characters that "look the same" but have different code points get treated as equal. "IDENTICAL" isn't exposed by `Intl.Collator` but it should be equivalent to plain-ol' javascript string comparison operators.

It is a little weird that a case-and-diacritic-sensitive collator equality will still be subtly different from "==" without a collator, but I think it makes sense to say "even if your collator is case/diacritic-sensitive, the fact that you're using it implies you want some form of locale-aware comparison".

> Can we consider some alternatives with better code readability?

Yeah, of those three suggestions I think my preferences are in the same order you wrote them. @anandthakker any thoughts on what fits in best with the rest of expressions?

ChrisLoer:
@anandthakker and I discussed this in more detail today and decided on the syntax:

`['collator', { caseSensitive: <boolean>, diacriticSensitive: <boolean>, locale: <string> }]`

We went back and forth on whether we could avoid introducing a collator type/expression entirely and just pass the configuration object as an argument to operators that supported collators. The use-case that pushed us towards keeping an explicit collator type is wanting to support defining collators in "let" expressions. Because we think it's very possible collator expressions could themselves depend on something like [`match-locales`](https://github.com/mapbox/mapbox-gl-js/issues/6197), we want to be able to use "let" to keep expressions that use collators from getting too verbose.

There's a little bit of an open question on whether we should make `locale` optional or require an explicit "default" parameter. The idea is that if locale is implicit, it's too easy for an expression author to think that case/diacritic-sensitivity doesn't depend on locale, and then be surprised when the behavior changes depending on user locale. On the other hand, it adds more verbosity and requires mixing a mapbox-specific enum with an otherwise BCP-47 argument. Also, `Intl.Collator` uses the convention of "unspecified locale => user default locale". So I could be convinced either way but right now my inclination is to keep it as an optional parameter.

ChrisLoer:
@anandthakker I just added serialization support for the collator functionality (haven't yet switched to object notation). One thing that came up from bootstrapping the tests is that we should consider whether `collator` is feature constant. The issue is that _technically_ what the locale resolves to can change at any time. On those grounds I made it non-feature constant. But as I think about it more, that's probably enough of an edge case that we don't want to sacrifice the performance benefits of allowing constant folding?

ChrisLoer:
Updated to new `{ "caseSensitive": true, "diacriticSensitive": false, "locale": "en" }` syntax, with default values for all of them (case/diacritic insensitive in system locale). I removed the non-feature-constant flag for `collator` as well.

src/style-spec/expression/definitions/literal.js
```diff
@@ -58,6 +59,8 @@ class Literal implements Expression {
             return ["literal", this.value];
         } else if (this.value instanceof Color) {
             return ["rgba"].concat(this.value.toArray());
+        } else if (this.value instanceof CollatorInstantiation) {
+            return this.value.serialize();
```
anandthakker: This is here for the same reason as the `Color` branch above, right?  I.e., because constant folding can cause an expression tree to contain `Literal` with `value: Color` or `value: CollatorInstantiation`?

It's pre-existing, but I think this is a bit of a wart, that `Literal` can hold a value for which there's no actual literal, and thus has to know (or, in this case, the value itself has to know) how to serialize an expression that it wouldn't be responsible for parsing.  @ChrisLoer would you mind adding a quick comment in each of these two branches just briefly noting that they're here to cover the constant-folding case?

src/style-spec/expression/definitions/collator.js
```diff
@@ -0,0 +1,141 @@
...
+type CollatorOptions = {
+  localeMatcher?: 'lookup' | 'best fit',
+  usage?: 'sort' | 'search',
+  sensitivity?: 'base' | 'accent' | 'case' | 'variant',
+  ignorePunctuation?: boolean,
+  numeric?: boolean,
+  caseFirst?: 'upper' | 'lower' | 'false'
+}
+
+export class CollatorInstantiation {
```
anandthakker: Suggested renames:
 - `declare class Collator` -> `declare class Intl$Collator` (this should be okay, since it's only being declared at top-level so that it can be referenced within the `Intl` class declaration, right?)
 - `CollatorInstantiation` -> `Collator`

anandthakker: On a different note: `values.js` will need to become aware of this class, e.g., for `isValue`, `typeOf`

src/style-spec/reference/v8.json
```diff
@@ -2193,6 +2193,15 @@
           }
         }
       },
+      "collator": {
+        "doc": "Returns a `collator` for use in locale dependent comparison operations. The `caseSensitive` and `diacriticSensitive` options default to `false`. The `locale` argument specifies the IETF language tag of the locale to use. If none is provided, the default locale is used. If the requested locale is not available, the `collator` will use a system-defined fallback locale. Use `resolved-locale` to test the results of locale fallback behavior.",
```
anandthakker: Nit: `locale dependent` -> `locale-dependent`?

src/style-spec/expression/definitions/collator.js
```diff
@@ -0,0 +1,141 @@
...
+        const options = (args[1]: any);
+        if (typeof options !== "object" || Array.isArray(options))
+            return context.error(`Collator options argument must be an object.`);
+
+        const caseSensitive = context.parse(
+            options['caseSensitive'] === undefined ? false : options['caseSensitive'], 1, BooleanType);
+        if (!caseSensitive) return null;
+
+        const diacriticSensitive = context.parse(
+            options['diacriticSensitive'] === undefined ? false : options['diacriticSensitive'], 2, BooleanType);
```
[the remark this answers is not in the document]
ChrisLoer: That sounds like a good idea, but OK if I just correct the index to "1" for now?

src/style-spec/expression/definitions/collator.js
```diff
@@ -0,0 +1,141 @@
...
+        if (options['locale']) {
+            locale = context.parse(options['locale'], 3, StringType);
+            if (!locale) return null;
+        }
+
+        return new CollatorExpression(caseSensitive, diacriticSensitive, locale);
+    }
+
+    evaluate(ctx: EvaluationContext) {
+        return new CollatorInstantiation(this.caseSensitive.evaluate(ctx), this.diacriticSensitive.evaluate(ctx), this.locale ? this.locale.evaluate(ctx) : null);
```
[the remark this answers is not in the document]
ChrisLoer: You mean the cost of `new Intl.Collator` (which we do on each `compare`)? It would be easy enough to move into the `CollatorInstantiation` constructor (allowing all of the uses in one expression to share the same constructor). But doesn't seem like something we should bother doing unless we see it showing up as a hot spot...

src/style-spec/expression/definitions/collator.js
ChrisLoer: I updated `values.js`, but I haven't actually figured out an expression test that would behave differently with the old `isValue => false` and `typeOf => ObjectType` behavior...

hsivonen:
I'm a developer working on Firefox's Gecko engine and on the collator component of [ICU4X](https://github.com/unicode-org/icu4x/). I have identified the data for search collations as having a disproportionately large contribution to binary size and, therefore, am exploring ways to make the data size smaller.

The vast, vast majority of the usage of `Intl.Collator` with `usage: "search"` on the Web can be traced back to this PR, so understanding the use case here is relevant to understanding the impact of tweaks to the underlying data.

What kind of user-visible feature does mapbox-gl-js use the accent-insensitive comparison for? From the source code, it seems that not only are equal/not-equal filtering supported but also less-than/greater-than. The semantics of `usage: "search"` are such that the application is only supposed to pay attention to equal/not-equal. Less-than/greater-than is OK as an implementation detail e.g. in a data structure, but not OK as a user-visible ordering (`usage: "sort"` is appropriate for that). How does mapbox-gl-js use greater-than/less-than with `usage: "search"`?

Also, the comments above talk only about accent-insensitivity. Does mapbox-gl-js rely on the specifics of `usage: "search"` for scripts that don't have accents?

zcorpan:
I helped @hsivonen with collecting the data.

> Just by number of code appearances, or by how often it is actually executed / ran?

By code appearances. In particular, response bodies for requests in [httparchive](https://httparchive.org/faq) (11.5m pages) that match the regexp `\b(Intl\.Collator\([^\(\)]+\))` and where the matched group also matches `\busage[\'"]?\s*:\s*[\'"`]search\b`.

hsivonen:
> I have a hard time understanding the technical differences between "sort" and "search" (or why usage is even necessary).

My inference from the evidence that I've seen is that it started with the notion that for various Latin-script languages collations happen to already provide a diacritic-insensitive equality definition such that some things that are technically diacritics but on a per-language basis aren't treated as such are _not_ ignored so that e.g. o and ö in Finnish, Swedish, Icelandic and Turkish can be considered unequal even when comparing diacritic-insensitively. Therefore, if you want diacritic-insensitivity where some diacritics are _not_ treated as diacritics in a language-dependent way, you can reuse collation data.

Then it was observed that the concept could be extended to Arabic and Thai _scripts_ in a way that doesn't reuse a sorting tailoring, so search collations become a separate thing with the script-level aspects going into a common search root. (Also, once sort and search got separated, some Latin-script languages got search behavior that diverges from sorting.)

> Why would "search" contribute more to binary size than "sort"?

In sorting, there are two layers of data: The root collation and, optionally, a language-specific tailoring overlay. In search, there are logically three layers of data: the root for sorting, a search root overlaid on that, and then, optionally, a language-specific tailoring. However, implementations only admit two layers, so for each language that's supposed to reuse its sort tailoring for searching, you instead carry a search tailoring that contains a merged _copy_ of the search root and a _copy_ of the sort tailoring for the language.

The obvious solution of "allow three layers of data, then" would make ctrl/cmd-f perform worse in apps that use search collations for interactive incremental search (which motivates the data but isn't supported by the Web API). I don't think it's realistic for _me_ to pursue that in the context of ICU4C, but depending on what I learn, I might pursue it in the context of ICU4X.

> Is this just a hint for performance reasons, or are there actual differences for the result?

There are actual differences, but I'm trying to figure out if this mapbox-gl-js use case is meant to care about them. Depending on language, the differences for what becomes greater-than and what becomes less-than are so notable that `usage: "search"` is very inappropriate to use for user-visible sorting.

> I could imagine that this collator feature is being used to compare feature names to that in some textbox or to highlight a specific building/country/.. by matching the name.
...
> However, I also doubt that collators are ever used for sorting

Thanks. So far, it sounds like it would be tolerable for Arabic-script and Thai matching to become stricter (less insensitive to certain marks) when the resolved locale is a Latin-script language with its language-specific rules for what diacritics are _not_ ignored in diacritic-insensitive search. (Again, English and French are do not have such language-specific rules.)

PR #6275, opened 2018-03-03
Write PNGs with unassociated alpha

### LINKED ISSUE #3771: Fringing artifacts in circle layers
The changes in support of `circle-stroke` introduced some fringing artifacts around the outer edge of the circle. This can be seen in several of the test cases on the `native-circle-stroke` branch:

[circle-stroke-width/default](https://github.com/mapbox/mapbox-gl-test-suite/blob/276f03fd17d7a3a515f1dc0eb69910a69b7d6837/render-tests/circle-stroke-width/default/expected.png):

image

[circle-stroke-color/zoom-and-property-function](https://github.com/mapbox/mapbox-gl-test-suite/blob/d9919c4ed3abaa3585309070920142931257bd84/render-tests/circle-stroke-color/zoom-and-property-function/expected.png):

image

### LINKED ISSUE #3913: Fix "fill-pattern/opacity" integration render test
Migrated from https://github.com/mapbox/mapbox-gl-test-suite/issues/165

> Our current rendering for [`fill-pattern` with non-opaque opacity](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/fill-pattern/opacity) doesn't look right to me:
> 
> image
> 
> It's grayish in that way that usually indicates incorrect premultiplication somewhere.
> 
> Here's the opacity 1 version:
> 
> image

jfirebaugh:
The test harness was sending the GL premultiplied pixel data directly to PNG, which expects unassociated pixel data. This led to test results that looked wrong visually, even though the rendering was correct. See the `expected.png` updates for examples. 

(I ran an `UPDATE=1` across the board, but only committed the `expected.png`s for tests that were failing after the `render.js` change in either JS or native.)

Fixes #3771.
Fixes #3913.


jfirebaugh:
Among the expecteds that I didn't commit, there are a bunch that now are slightly different from (correct) actual, but not enough to trigger pixelmatch. Maybe we should just do an across-the-board update?

lbud:
> Maybe we should just do an across-the-board update?

👍 that sounds like a good idea to maintain a reasonable error threshold for those tests.

PR #6283, opened 2018-03-05
provide helpful message if access token triggers 401: Unauthorized

### LINKED ISSUE #5852: Friendly message when malformed accessToken provided
I copied an example while not logged in:

```...
<script>
mapboxgl.accessToken = '<your access token here>';
var map = new mapboxgl.Map({
...
```

Not surprisingly it didn't work. More surprisingly, there's not a helpful error message in the console, just:

```
Object { status: 401, stack: "" }
```

Prior to version 0.35.0, this produced a slightly more (maybe) useful error message:

```
Error: Unauthorized
Stack trace:
[191]</exports.getJSON/n.onload@https://api.tiles.mapbox.com/mapbox-gl-js/v0.34.0/mapbox-gl.js:392:513
```

Suggestion: throw the same error as when no `accessToken` is set at all:

```
Error: An API access token is required to use Mapbox GL. See https://www.mapbox.com/api-documentation/#access-tokens
```

Currently the "helpful" line of code ` mapboxgl.accessToken = '<your access token here>';` is actually making things worse for the newbie.

An alternative is to simply comment that line out in every example.

(Nitpick: technically the error message is not accurate. Mapbox-GL-JS works just fine without an access token, as long as you're not accessing any Mapbox services.)

mollymerp:
fix #5852 

## Launch Checklist

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -85,7 +85,7 @@ exports.getJSON = function(requestParameters: RequestParameters, callback: Callb
             }
             callback(null, data);
         } else {
-            if (xhr.status === 401) {
+            if (xhr.status === 401 && requestParameters.url.match(/mapbox.com/)) {
                 callback(new AJAXError(`${xhr.statusText}: you may have provided an invalid Mapbox access token. See https://www.mapbox.com/api-documentation/#access-tokens`, xhr.status, requestParameters.url));
             } else {
                 callback(new AJAXError(xhr.statusText, xhr.status, requestParameters.url));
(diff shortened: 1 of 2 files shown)


src/util/ajax.js
```diff
@@ -85,7 +85,11 @@ exports.getJSON = function(requestParameters: RequestParameters, callback: Callb
             }
             callback(null, data);
         } else {
-            callback(new AJAXError(xhr.statusText, xhr.status, requestParameters.url));
+            if (xhr.status === 401) {
```
lucaswoj: What do you think about also testing the URL is on the `mapbox.com` domain? So long as we promise [compatibility with third party vector tile sources](https://www.mapbox.com/mapbox-gl-js/example/third-party/) we must be careful with logic that assumes we're talking to Mapbox servers.

mollymerp: good idea!

Issue #4879, opened 2017-06-23
Validate maximum number of data-driven paint properties per layer


jfirebaugh:
The maximum number of properties that can be data-driven in a single layer simultaneously is effectively limited by `GL_MAX_VERTEX_ATTRIBS`, which can be as low as 8. As we grow the number of data-driven properties, it becomes more likely that a style author could go over this limit unknowingly, test on devices that support `GL_MAX_VERTEX_ATTRIBS` > 8, see nothing wrong, and then deploy and have the style break on lower-end devices.

We should introduce style validation to prevent this. The validation rule should enforce a limit on the number of data-driven paint properties used per layer. The limit varies by layer type, and is defined by (8 - maximum number of _non_-paint-property attributes required by any shader used for that layer type).

jfirebaugh:
This is, unfortunately, a validation that may become _stricter_ over time, if we introduce features that require adding new non-paint-property attributes to a shader. Not sure there's any solution for that other than to avoid such features.

anandthakker:
Should this be something that we flag in Studio as well?  @mapbox/studio

jfirebaugh:
Current limits: circle: 7, fill-extrusion: 6, fill: 7, heatmap: 7, line: 6, symbol: 4. For symbol layers, the limit is independent in `text-*` properties and `icon-*` properties. That is, you can make up to 4 `text-*` properties data driven, and simultaneously, up to 4 `icon-*` properties.

These numbers are subject to change, but effectively, given the _current_ set of data-driven properties, the only way to hit this limit is to make _all_ the `icon-*` paint properties or _all_ the `text-*` paint properties data-driven simultaneously.

PR #6289, opened 2018-03-07
Data-driven styling support for *-pattern properties

mollymerp:
todos:
- [x] figure out flow typing for `CrossFadedDataDrivenProperty` or potentially refactor `CrossFaded` out
- [x] preserve precision when passing pattern texture coordinates to shaders
- [x] add support for composite expressions
- [x] identify bug causing flashing/incorrect texture sampling
- [x] figure out how to deal w layers that share a bucket but have different `line-pattern` properties
- [x] decide on how to pack the attributes / manage the attribute buffers

## Launch Checklist

 - [x] post benchmark scores


mollymerp:
this is ready for an initial review, with the caveat that there are two open questions:
- because `line-pattern` is a paint property, it is possible that there would be more than one layer with different `line-pattern` expressions on the same bucket. not sure how we want to handle this case, because it will come up again with the other `*-pattern` properties. 
- right now I'm not doing any attribute packing on the `line-pattern` PaintArrays. There are a couple of options I see:
  - we could get down to only needing a single `vec4` attribute for line-pattern, but it would require creating two paint arrays, one with the packed coordinates for the `min` and `mid` CrossFaded pattern locations, and the other with `mid` and `max` pattern locations, and then manually binding the correct paint buffer depending on if we're zooming in or out. 
    - this has the advantage of only requiring one attribute which is important for keeping overall attribute #s down for device compatibility (see #4879)
    - disadvantage is it would require more special casing in `ProgramConfiguration` and/or manual buffer binding in `drawLine`
  - we could have 2 attributes, one 4 component holding the coordinates of `min` and `max` (we only need one for rendering, but don't know which at layout time) and one 2 component with `mid` coordinates. 

luckily, cross-faded properties are already interpolated along zoom-levels and thus require the same data to be passed for `Source` and `Composite` functions, so we don't need to worry about packing in double the data for the `Composite` case. 


cc @ansis @jfirebaugh @kkaefer

src/style/style_layer/line_style_layer.js
```diff
@@ -50,18 +50,32 @@ class LineStyleLayer extends StyleLayer {
...
+    getCrossfadeParameters(): {fromScale: number, toScale: number, t: number} {
+        const p = this.transitionParameters;
+        const z = p.zoom;
+        const fraction = z - Math.floor(z);
+        const t = p.crossFadingFactor();
+
+        return z > p.zoomHistory.lastIntegerZoom ?
+            { fromScale: 2, toScale: 1, t: fraction + (1 - fraction) * t } :
+            { fromScale: 0.5, toScale: 1, t: 1 - (1 - t) * fraction };
+    }
```
mollymerp: When I implement `fill-pattern` and `fill-extrusion-pattern` dds in a follow-up PR, I will probably move this function to the parent `StyleLayer` class.

src/style/cross_faded.js
```diff
@@ -3,6 +3,9 @@
 export type CrossFaded<T> = {
     from: T,
     to: T,
+    min: T,
+    mid: T,
+    max: T,
```
mollymerp: when I implement the remaining `*-pattern` properties, I will be able to remove the `from` and `to` components from this type in favor of the `StyleLayer` getter for cross fade parameters.

jfirebaugh:
> because `line-pattern` is a paint property, it is possible that there would be more than one layer with different `line-pattern` expressions on the same bucket.

Yep, this is something that's true of all data-driven paint properties. We handle it in `ProgramConfigurationSet`, which maintains a map of layer ID to `ProgramConfiguration`, and `ProgramConfiguration` contains all the per-layer data. Will that strategy work for `line-pattern` or is there additional nuance?

mollymerp:
🤦‍♀️ well, that's embarrassing. I think I confused myself about this step where we determine which glyphs to add to the tile's `iconAtlas` and how it relates to the same step in `SymbolBucket` where all properties that reference a sprite are layout specific.

https://github.com/mapbox/mapbox-gl-js/pull/6289/files#diff-b1205e64508e847d415d42243c10212bR158

it shouldn't be an issue to loop through each layer with data-driven `line-pattern` here and add all relevant glyphs to the atlas 👍

jfirebaugh:
I think we should aim to handle _non_-data-driven `*-pattern` properties via `ProgramConfiguration` + `ConstantBinder` + `Tile#iconAtlasImage`, rather than keeping a separate atlas in `ImageManager` for them and manually setting the uniforms.

mollymerp:
Getting some diffs after moving uniform bindings to the `Pattern*Binders`

image

they're tiny and I haven't looked too much into them – if they're acceptable diffs I will update the tests, and if not I will spend more time trying to fix. 

I also have a version of this PR that reduces the # attributes to 1, but requires dynamic checking and possibly binding of the vertex buffer on each frame/tile (there is a separate vertex buffer per tile depending on if the user is zooming in or zooming out for crossfading to work correctly) – should I pursue that route or stick with the current approach?

jfirebaugh:
Updating the tests for those diffs sounds good. Or, if you want to, minimize the tests so that they don't show those minor differences (use inline GeoJSON with simple geometries, 64x64 size, etc.).

Reducing the number of attributes to one sounds great! It should be fairly cheap to switch vertex buffers on the fly, and it'll get cheaper once we have state tracking that allows us to rebind only the attributes that have changed.

mollymerp:
Thanks @anandthakker for helping me better understand floating point precision 🔢 
I haven't come up with a way to pack two integers (with greater than 255 maximum value), but I have reduced the number of attributes to two and implemented a method to update the vertex buffer that is used for data-driven line-patterns at runtime. 

This is ready for review, with the caveat that I still need to fix this PR for non-integer zoom stops because that is currently broken for both data-driven styling and camera functions. DIdn't notice this was broken until late this afternoon, so I will aim to address that tomorrow.

src/data/bucket/line_attributes.js
```diff
@@ -1,10 +1,17 @@
...
-export const {members, size, alignment} = layout;
+// these attributes are only needed if using data-driven line-pattern
+export const linePatternAttributes = createLayout([
+    // [tl.x, tl.y, br.x, br.y]
+    {name: 'a_pattern_from', components: 4, type: 'Float32'},
+    {name: 'a_pattern_to', components: 4, type: 'Float32'}
+]);
+
+export default lineLayoutAttributes;
+export const {members, size, alignment} = lineLayoutAttributes;
```
anandthakker: Since this module now exports more than one set of attributes, let's drop the default export in favor of two named exports (See https://github.com/mapbox/mapbox-gl-js/issues/6336)

src/data/bucket/line_bucket.js
```diff
@@ -110,20 +124,78 @@ class LineBucket implements Bucket {
...
+        this.dataDrivenPatternLayers = [];
+
+
+        for (let i = 0; i < this.layerIds.length; i++) {
+            const id = this.layerIds[i];
+            const programConfiguration = this.programConfigurations.get(id);
+            if (programConfiguration.binders && programConfiguration.binders['line-pattern'] && programConfiguration.binders['line-pattern'].isDataDriven()) {
+                this.dataDrivenPatternLayers.push(i);
+            }
+        }
```
anandthakker: Rather than storing this on `this`, could we compute it locally in `populate()`?   (Also: why is this an array of indices into `this.layers` rather than an array of layers themselves?)

src/data/bucket/line_bucket.js
```diff
@@ -110,20 +124,78 @@ class LineBucket implements Bucket {
...
+            if (typeof feature.id !== 'undefined') {
+                lineFeature.id = feature.id;
             }
+
+            this.features.push(lineFeature);
+            options.featureIndex.insert(feature, geometry, index, sourceLayerIndex, this.index);
+        }
+    }
+
+    // used if line-pattern is data-driven
```
anandthakker: Isn't `addFeatures()` also used if line-pattern is constant?

src/data/bucket/line_bucket.js
anandthakker: One more question: why use `binder.isDataDriven()` rather than checking `layer.paint.get('line-pattern')`?

src/data/program_configuration.js
```diff
@@ -52,83 +59,143 @@ function packColor(color: Color): [number, number] {
...
+                gl.uniform4f(program.uniforms[`u_${name}`], value.r, value.g, value.b, value.a);
+            } else {
+                gl.uniform1f(program.uniforms[`u_${name}`], value);
+            }
+        }
+    }
+
+}
+
+class PatternConstantBinder<T> extends ConstantBinder<T> {
```
anandthakker: I'm still inclined to say `PatternConstantBinder` should be a top-level class rather than a subclass of `ConstantBinder` (and likewise for `PatternCompositeExpressionBinder`).

src/source/tile.js
```diff
@@ -224,9 +225,8 @@ class Tile {
 
         const gl = context.gl;
 
-        if (this.iconAtlasImage) {
-            this.iconAtlasTexture = new Texture(context, this.iconAtlasImage, gl.RGBA);
-            this.iconAtlasImage = null;
```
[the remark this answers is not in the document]
mollymerp: yes – I had meant to come back to this. setting `iconAtlas.image` to null was upsetting flow but I will revisit! 👍

src/data/bucket/line_bucket.js
mollymerp: yep, sorry – an artifact from an earlier iteration 🙈

src/render/draw_line.js
```diff
@@ -18,17 +18,16 @@ export default function drawLine(painter: Painter, sourceCache: SourceCache, lay
...
     const context = painter.context;
-
     context.setDepthMode(painter.depthModeForSublayer(0, DepthMode.ReadOnly));
     context.setColorMode(painter.colorModeForRenderPass());
 
+    const linePattern = layer.paint.get('line-pattern');
     const programId =
         layer.paint.get('line-dasharray') ? 'lineSDF' :
-        layer.paint.get('line-pattern') ? 'linePattern' : 'line';
+        linePattern && linePattern.constantOr((1: any)) ? 'linePattern' : 'line';
```
mollymerp: open to better ways of doing this check – basically if `line-pattern` is undefined, `constantOr(1)` will return undefined, but `linePattern` itself is no longer falsy if `line-pattern` is undefined because it is now a `PossiblyEvaluatedPropertyValue`

src/style/properties.js
```diff
@@ -576,6 +576,62 @@ export class DataDrivenProperty<T> implements Property<T, PossiblyEvaluatedPrope
...
 }
 
+/**
+ * An implementation of `Property` for  data driven `line-pattern` which are transitioned by cross-fading
+ * rather than interpolation.
+ *
+ * @private
+ */
+
+export class CrossFadedDataDrivenProperty<T> extends DataDrivenProperty<?CrossFaded<T>> {
```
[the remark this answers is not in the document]
mollymerp: per chat – we decided not to convert `CrossFadedDataDrivenProperty` to a top-level implementation of `Property` because it caused issues with `PossiblyEvaluatedPropertyValue` where the following wouldn't work

```es6
export class PossiblyEvaluatedPropertyValue<T> {
    property: DataDrivenProperty<T> | CrossFadedDataDrivenProperty<T>;
```

because we need the `<T>` in `property` to be `CrossFaded<T>` but the base `<T>` to be `string`

src/render/draw_line.js
anandthakker: I'm okay with this as-is, since it's analogous to how we do other such checks.

src/data/program_configuration.js
```diff
@@ -246,8 +322,131 @@ class CompositeExpressionBinder<T> implements Binder<T> {
...
+
+        this.zoomInPaintVertexArray = new PaintVertexArray();
+        this.zoomOutPaintVertexArray = new PaintVertexArray();
+    }
+
+    defines() {
+        return [];
+    }
+
+    populatePaintArray(length: number, feature: Feature, imagePositions: ?{[string]: ImagePosition}) {
```
anandthakker: I'd rather have `imagePositions` be required parameter, both here and on the `Binder` interface, and a parameter defaulting to `{}` in `ProgramConfigurationSet#populatePaintArray()`

src/render/draw_line.js
jfirebaugh: Is there a way to eliminate the `any` cast?

mollymerp:
@jfirebaugh thanks for the feedback

removing `from` and `to` from `CrossFaded` would require implementing DDS for `fill-extrusion-pattern` and `line-dasharray` in addition to `fill-pattern`. because this dds support for this property is high priority, I would like to address this in a follow-up PR so line-pattern can be supported in the next release. 

> Extend this to fill-pattern, remove from and to in CrossFaded, move getCrossfadeParameters to StyleLayer.

similarly, I don't think #6389 getting resolved should block this PR and I think its reasonable to postpone resolution til the other `CrossFaded` properties add dds support. 

> Figure out how to avoid /line-pattern/ special case in createDynamic. Maybe the responsibility for creating an appropriate binder can be delegated to the Property interface.

👍 I'll look into this

mollymerp:
> Resolve #6389 in a way that avoids the /-pattern$/.test in generate-style-code.js

also fwiw there is already a `/-pattern$/.test` condition in master – I agree we should get rid of it but perhaps in a separate PR

https://github.com/mapbox/mapbox-gl-js/blob/866ab7b46635f619347334da6e1baa2cc10e7ecf/build/generate-style-code.js#L44-L46

src/render/draw_fill.js
```diff
@@ -60,11 +55,20 @@ function drawFill(painter: Painter, sourceCache: SourceCache, layer: FillStyleLa
 }
 
 function drawFillTiles(painter, sourceCache, layer, coords, drawFn) {
-    if (isPatternMissing(layer.paint.get('fill-pattern'), painter)) return;
+    // if (isPatternMissing(layer.paint.get('fill-pattern'), painter)) return;
 
+    // if (layer.paint.get('fill-pattern').constantOr(1) )
```
ansis: Are these two commented out lines safe to remove?

src/style/properties.js
```diff
@@ -14,7 +14,7 @@ import type EvaluationParameters from './evaluation_parameters';
 
 import type {
     Feature,
-    GlobalProperties,
+    // GlobalProperties,
```
ansis: Is this safe to remove?

src/style/properties.js
```diff
@@ -425,9 +425,9 @@ type PossiblyEvaluatedValue<T> =
 export class PossiblyEvaluatedPropertyValue<T> {
     property: DataDrivenProperty<T>;
     value: PossiblyEvaluatedValue<T>;
-    globals: GlobalProperties;
+    globals: EvaluationParameters;
```
ansis: I'm not very familiar with this part of the code, but does it make sense to rename `globals` now that the type isn't `GlobalProperties` before?

mollymerp:
I have incorporated the updating feature state changes from #6263 and have everything passing except one hold out `fill-pattern` render test. 

some notable benchmark results:

image

image

image

querying has regressed quite a bit for reasons unbeknownst to me – will dig into this more 😕 

image

mollymerp:
hmm when I run the query benchmarks by themselves the results are much different and fluctuate quite a bit
﻿﻿
image

image

image

image

🐟 🐠 🐟

anandthakker:
Yeah @mollymerp I think the Query* benchmarks are pretty volatile in general

mollymerp:
this is ready for another look @anandthakker @ansis – I've added support for fill and fill-extrusion pattern dds

image

src/data/program_configuration.js
```diff
@@ -299,8 +388,171 @@ class CompositeExpressionBinder<T> implements Binder<T> {
...
+
+                zoomOutArray.emplace(i,
+                    imageMid.tl[0], imageMid.tl[1], imageMid.br[0], imageMid.br[1],
+                    imageMax.tl[0], imageMax.tl[1], imageMax.br[0], imageMax.br[1]
+                );
+            }
+        }
+    }
+
+    getVertexBuffer(crossfade: CrossfadeParameters) {
```
anandthakker: I think it would be clearer to inline this into `updatePatternPaintBuffers` below -- do we expect to use it anywhere else?

src/style/cross_faded.js
anandthakker: It looks like it should be possible now for `CrossFaded<T>` to be pared down to `{min: T, mid: T, max: T}`.  Maybe adding `from: 'min' | 'max'` to `CrossfadeParameters` so that in places where we're doing `pattern.from`, we can do `pattern[crossfade.from]` instead?

src/style-spec/validate/validate_expression.js
```diff
@@ -14,9 +14,9 @@ export default function validateExpression(options: any) {
         });
     }
 
-    if (options.expressionContext === 'property' && options.propertyKey === 'text-font' &&
+    if (options.expressionContext === 'property' && (options.propertyKey === 'text-font' || options.propertyKey === 'line-pattern') &&
         (expression.value: any)._styleExpression.expression.possibleOutputs().indexOf(undefined) !== -1) {
-        return [new ValidationError(options.key, options.value, 'Invalid data expression for "text-font". Output values must be contained as literals within the expression.')];
+        return [new ValidationError(options.key, options.value, `Invalid data expression for "${options.propertyKey}". Output values must be contained as literals within the expression.`)];
```
anandthakker: Is it true that this in fact applies to any cross-faded-data-driven property?  If so, let's do this check using use the `property-type` value instead of checking for each -pattern property... though unfortunately there'll be a bit of a naming clash adding this to the validation options: https://github.com/mapbox/mapbox-gl-js/blob/5b3af8e1515ee311967661b5aabdab65f1376e36/src/style-spec/validate/validate_property.js#L62

src/style/style_layer.js
```diff
@@ -32,6 +38,7 @@ class StyleLayer extends Evented {
     maxzoom: ?number;
     filter: FilterSpecification | void;
     visibility: 'visible' | 'none';
+    transitionParameters: EvaluationParameters;
```
anandthakker: Prefix with `_` to indicate that this is private.  Maybe also rename to just `_parameters`, since this is used to evaluate all zoom-driven properties, not just transitionable ones.

src/style/style_layer/fill_style_layer.js
```diff
@@ -24,6 +24,7 @@ class FillStyleLayer extends StyleLayer {
     }
 
     recalculate(parameters: EvaluationParameters) {
+        super.recalculate(parameters);
         this.paint = this._transitioningPaint.possiblyEvaluate(parameters);
```
anandthakker: This already happens in the base class's `recalculate`, which we're now calling.

src/shaders/_prelude.vertex.glsl
```diff
@@ -48,12 +48,18 @@ float unpack_mix_vec2(const vec2 packedValue, const float t) {
...
     vec4 minColor = decode_color(vec2(packedColors[0], packedColors[1]));
     vec4 maxColor = decode_color(vec2(packedColors[2], packedColors[3]));
     return mix(minColor, maxColor, t);
 }
 
+vec4 unpack_mix_vec4(const vec4 patternVertices, const float t) {
+    // line pattern location vertices are not packed, and they use a different t
+    // for interpolation that is specific to cross-faded properties
+    return patternVertices;
+}
```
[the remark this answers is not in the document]
mollymerp: I agree it is confusing, but since cross faded properties are more the exception than the rule, I don't know if adding a new slew of `#define`s will be clearer in the long run I took an alternate approach – let me know if this works for you!

src/shaders/index.js
```diff
@@ -176,13 +186,24 @@ uniform ${precision} ${type} u_${name};
 #endif
 `;
             } else /* if (operation === 'initialize') */ {
-                return `
+                if (unpackType === 'vec4') {
+                    // vec4 attributes are only used for cross-faded properties, and are not packed
```
anandthakker: 👍 on this approach

anandthakker: Can we duplicate this comment above, too?

src/render/draw_fill.js
```diff
@@ -60,11 +54,11 @@ function drawFill(painter: Painter, sourceCache: SourceCache, layer: FillStyleLa
 }
 
 function drawFillTiles(painter, sourceCache, layer, coords, drawFn) {
-    if (isPatternMissing(layer.paint.get('fill-pattern'), painter)) return;
-
     let firstTile = true;
     for (const coord of coords) {
         const tile = sourceCache.getTile(coord);
+        if (layer.paint.get('fill-pattern').constantOr((1: any)) && !tile.patternsLoaded()) continue;
```
anandthakker: Just confirming: with this nice & clean patternsLoaded() check, we're no longer checking whether the particular patterns needed by the fill-pattern value are present in the image atlas -- is that okay?

charandas:
We have been running with this on one of our MVPs for last couple months, and it's performing well and without issues. Any reason why it has not been merged?

charandas:
By the way, just want to report that `feature-state` currently is not working with DDS `fill-pattern` on a given fill layer, when using this branch.

mollymerp:
@charandas sorry for the delay. we have to wait to merge until the implementation in mapbox-gl-native is complete because the two libraries share shaders. Its hopefully almost done 🙏⏳https://github.com/mapbox/mapbox-gl-native/pull/12284

thanks for the bug report – working on a fix for that now!

mollymerp:
notable benchmarks
<details>

image
image
image
image

image
image

</details>

mourner:
@mollymerp looks like a pretty huge regression for Layout — is that spurious or reproducible with repeated runs?

mollymerp:
@mourner 😞 yep its reproducable and kind of expected because there is an additional worker-main thread round trip for *-pattern layers to fetch the icons needed to layout. I'm trying something quickly to avoid this if the pattern property is not used, and hopefully that will reduce the regression.

mollymerp:
layout benchmark looking much better after 3cd1985 
image

src/style/style_layer.js
```diff
@@ -162,6 +186,7 @@ class StyleLayer extends Evented {
     }
 
     recalculate(parameters: EvaluationParameters) {
+        this._parameters = parameters;
```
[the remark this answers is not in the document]
mollymerp: we can get rid of it if we pass the evaluation parameters from `Map#_render` to `painter#render` and through to the draw calls. Is that preferable?

src/data/bucket/fill_bucket.js
```diff
@@ -65,18 +68,49 @@ class FillBucket implements Bucket {
...
+                    geometry: geometry,
+                    properties: feature.properties,
+                    type: feature.type
+                };
+
+                if (typeof feature.id !== 'undefined') {
+                    fillFeature.id = feature.id;
+                }
+
+                this.features.push(fillFeature);
```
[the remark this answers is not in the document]
mollymerp: this follows the pattern that is also used in symbol layout. we could split it into separate steps, but this way avoids repeating the work  of parsing features out of the vector layers here:
https://github.com/mapbox/mapbox-gl-js/blob/2d0c4bc9872b51efefa77d41cf71ed806c8eb3c8/src/source/worker_tile.js#L77-L94

src/style/style_layer.js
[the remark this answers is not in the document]
jfirebaugh: I took a deeper look and understand things a lot better now: `CrossFadedDataDrivenProperty` wants access to the same state -- `fromScale`, `toScale`, and `t` -- that `CrossFadedProperty` returns via `CrossFaded`. So here's what you could do support them both in a common way:

* Rename `this._parameters` to `this._crossfadeParameters` and change the type to `CrossfadeParameters`.
* Move `getCrossfadeParameters()` to `EvaluationParameters`, so the above line in the PR becomes `this._crossfadeParameters = parameters.getCrossfadeParameters()`. (This isn't strictly necessary, but `getCrossfadeParameters()` seems like it should live on `EvaluationParameters` per [Feature Envy](https://refactoring.guru/smells/feature-envy).)
* Remove `fromScale`, `toScale`, and `t` from `CrossFaded<T>` and the duplicate calculation of these values from `CrossFadedProperty<T>::_calculate`. Instead, have layers that use either `CrossFadedProperty` or `CrossFadedDataDrivenProperty` obtain the calculated values from the same source: `this._crossfadeParameters`.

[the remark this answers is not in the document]
jfirebaugh: Seems like `CrossFaded<T>` should become `{min: T, mid: T, max: T}`. This would allow it to be used in more places, and maybe DRY up some of the duplication in the bucket classes.

src/data/bucket/fill_bucket.js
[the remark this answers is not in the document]
jfirebaugh: I was going to say, since we're restricting `*-pattern` properties to values where the possible outputs are contained as literals within the expression, we can take advantage of that to calculate the needed pattern images without reference to the actual features. But it looks like you've lifted that restriction in the latest revision. Do you remember the original rationale for it? On the one hand, it's nice that it would allow this optimization, but on the other, I can definitely envision someone wanting to use e.g. `"fill-pattern": ["get", "pattern-name"]`, since that's been so useful for `icon-image`.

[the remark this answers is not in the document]
mollymerp: I think the original reason for the restriction was to avoid the per-feature evaluation before the layout step, but yes it has been lifted and I agree, the ["get", "pattern-name"] use-case is probably going to be one users will appreciate and use.

src/style/style_layer.js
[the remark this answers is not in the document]
mollymerp: @jfirebaugh `{min, mid, max}` are only needed for data-driven property evaluation, and `to`, `from` is still useful for constant and camera expressions, and since the expression evaluation in the binders returns `T`, not `CrossFaded<T>` I've kept the `to`, `from` values for now. 

I DRYd up the bucket code in https://github.com/mapbox/mapbox-gl-js/pull/6289/commits/f1a06a333e061634ba658251fb5cdf277146f10b and took your suggestions for reworking `getCrossfadeParameters` in https://github.com/mapbox/mapbox-gl-js/pull/6289/commits/9f6bac9453f66c163d43b3dc824a18fa33228826. Thank you!!

src/data/bucket/pattern_bucket_features.js
```diff
@@ -0,0 +1,87 @@
...
+                    // save for layout
+                    patternFeature.patterns[layer.id] = { min, mid, max };
+                }
+            }
+            patternFeatures.push(patternFeature);
+        } else {
+            addFeature(patternFeature, geometry, index, {});
+        }
+
+        options.featureIndex.insert(feature, geometry, index, sourceLayerIndex, bucketIndex);
```
ansis: This method handles a couple non-pattern related things, like loading the geometry and indexing features. Is there anything here that can be separated and shared with buckets that don't have patterns (like circles?). It's very possible that doing this wouldn't make sense. Either way I don't see this as something that needs to block this pr.

charandas:
Thanks @jfirebaugh @mollymerp. Did something significant change as to the interface for this using `feature-state` in the commits that came after this [message](https://github.com/mapbox/mapbox-gl-js/pull/6289#issuecomment-413018575)?

I was able to get `fill-pattern` DDS work using `feature-state` prior to that day using code from this PR, and now it just silently fails, when I use the code from master?

I am doing something as simple as:

```
map.setPaintProperty('ctracts', 'fill-pattern', ["step", ["number", ["feature-state", "lmi"], 0], "pattern-1", 1, "pattern-2", 3, "pattern-3"])
```

I can also confirm that the feature-state is loaded with the `lmi` property. I can make a different issue with a JSBin once master is released to npm.

mollymerp:
yep @charandas – thanks for the flag. this was a bug I introduced as part of a performance related refactor. working on a fix now.

Issue #4095, opened 2017-02-01
Support gradient-colored shapes


lucaswoj:
_From @lucaswoj on August 21, 2015 19:8_

It would be useful to support gradient-colored lines, fills, and backgrounds. What would the stylesheet syntax look like? How would this work with data-driven styling?

ref https://github.com/mapbox/mapbox-gl-native/issues/2142

cc @webventil

![](https://cloud.githubusercontent.com/assets/1701915/9403260/b7a7eb82-47e5-11e5-84b6-7550d4c15edc.png)

_Copied from original issue: mapbox/mapbox-gl-style-spec#349_

lucaswoj:
_From @jfirebaugh on August 21, 2015 19:14_

See also https://github.com/mapbox/mapbox-gl-js/issues/748.

lucaswoj:
_From @tmcw on August 21, 2015 19:14_

I think the style spec side here is totally doable but the hard part is going to be data representation: GeoJSON and lots of other formats are derived at a basic level from the [Simple Features Spec](http://www.opengeospatial.org/standards/sfs), which, in its definition of 'simple' does not include the concept of values differing _along_ lines, which is how I'd assume you would want to control the color of the line.

The GPX and OSM formats do support this, and I have a [proposal for an add-on spec for GeoJSON](https://github.com/mapbox/geojson-coordinate-properties). Vector tiles, as far as I know, don't support values along lines.

lucaswoj:
_From @tomtaylor on August 24, 2015 12:58_

While this is being considered, it would also be useful to be able to have polylines where the dashing varies along the line too. So you can make certain sections as dotted, for example.

lucaswoj:
_From @jordi-adame on August 29, 2015 17:30_

+1

lucaswoj:
_From @cowgp on February 2, 2016 22:26_

+1

lucaswoj:
_From @isimonescu on September 7, 2016 11:0_

Any progress on this?

kaligrafy:
Any progress?

espy1122:
Any progress??

andrewharvey:
> I think the style spec side here is totally doable but the hard part is going to be data representation: GeoJSON and lots of other formats are derived at a basic level from the Simple Features Spec, which, in its definition of 'simple' does not include the concept of values differing along lines, which is how I'd assume you would want to control the color of the line.
>
[...]

Issue #4194, opened 2017-02-03
Replace property spec "function" property with "interpolatable" property


jfirebaugh:
The "function" property in the style specification JSON for a style property currently takes on the enumerated values "interpolated" and "piecewise-constant". These terms are a historical accident, not used elsewhere in our function vocabulary.

The "function" property is used:

* To select a default function type if the function itself does not have a `"type"` property. (If `"function": "interpolated"`, then `"type": "exponential"`, else `"type": "interval"`.)
* In function validation, to prohibit the use of exponential functions when the property is not interpolatable.
* To [check if interpolation is possible for purposes of style transitions](https://github.com/mapbox/mapbox-gl-js/blob/371240e078ed51bf525f6711eafe7e22ee290dc5/js/style/style_transition.js#L21-L21).
* In function validation, to mandate the use of integer stops when the property is not interpolatable (though this is [proposed for removal](https://github.com/mapbox/mapbox-gl-js/issues/4147)).

In essence, what's really being checked in all these cases is "is it possible to interpolate between two values of this property?" This would be more clearly served by a boolean property in the style specification: `"interpolatable": true/false`.

Whether or not a property is interpolatable can _almost_ be deduced from the property type, with `number` and `color` being interpolatable; `enum`, `string`, and `boolean` not; and `array` reducing to the interpolatability of the element type. But `line-dasharray` is the exception -- it's not interpolatable, even though it's an array of numbers.

PR #6303, opened 2018-03-08
Add line-gradient property

lbud:
This PR adds support for gradient lines, with the following limitations:
* `line-gradient` can only be used for lines with GeoJSON sources
  * the GeoJSON source used for these lines must specify the `"lineMetrics": true` option
* `line-gradient` cannot be used in conjunction with `line-dasharray` or `line-pattern`
* like `heatmap-color`, `line-gradient` must be specified using an expression with a special `line-progress` property: for example,
``` json
"line-gradient": [
    "interpolate",
    ["linear"],
    ["line-progress"],
    0, "rgba(0, 0, 255, 0.5)",
    0.5, "red",
    1, "#ffff00"
]
```

### Overview of changes:
* Adds a `lineMetrics` option to GeoJSON sources, which is passed through to geojson-vt in order to track distance metrics for linestring features
* Adds a `line-gradient` line paint property, which specifies a gradient and is used in the same way `heatmap-color` is used to write a 256x1px texture containing the gradient
  * Changes `HeatmapColorProperty` to `ColorRampProperty` to be used for both, and abstracts color ramp image creation into a utility module for use in both
* `LineBucket`:
  * checks for the presence of `mapbox_clip_start` and `mapbox_clip_end` properties, which are special properties added by geojson-vt when `lineMetrics=true`, and if so
  * precalculates the total distance of the tiled feature in tile units, and
  * scales the `distance` attribute of each vertex to [0, 2<sup>15</sup>) relative to the length of the entire original line geometry
* Adds new `line_gradient` shaders which sample from the line gradient texture based on the distance varying.

Refs https://github.com/mapbox/mapbox-gl-js/issues/4095.

### Benchmarks: 
http://bl.ocks.org/lbud/raw/e96f30170bfc39a6e3f1ddd2c0cc8fa1/ — the only benchmark I'd expect to potentially be affected by this is `LayerLine`, which in the linked benchmarks looks like it improves but when repeatedly running benchmarks actually stays just about equal. I also added a temporary `LayerLineGradient` benchmark (bottom of page) and it is, expectedly, just a little bit slower than the plain `LayerLine`, which doe not render to or sample from a texture.

## Launch Checklist

 - [x] depends on https://github.com/mapbox/geojson-vt/pull/93 being merged and released
 - [x] post benchmark scores


build/generate-style-code.js
```diff
@@ -44,8 +44,8 @@ global.propertyType = function (property) {
         return `DataDrivenProperty<${flowType(property)}>`;
     } else if (/-pattern$/.test(property.name) || property.name === 'line-dasharray') {
         return `CrossFadedProperty<${flowType(property)}>`;
-    } else if (property.name === 'heatmap-color') {
-        return `HeatmapColorProperty`;
+    } else if (property.name === 'heatmap-color' || property.name === 'line-gradient') {
```
jfirebaugh: It's great that we have a second property with the same semantics as `heatmap-color` -- that increases the likelihood that `heatmap-color` is not some odd exception (and as we know, the odd exceptions are disproportionately expensive to maintain).

Since we now have two similarly-behaving properties, it's worth finding a generic way to capture that the style spec and replace the property name comparisons here with the use of a semantic property like `isDataDriven` uses. Such a property would probably be useful for Studio as well. As a schema design question, it intersects with #4194 -- although in light of this new information, I'm not sure my proposal there is still the best choice. @lbud do you want to think about this and come up with a proposal?

lbud:
> It's great that we have a second property with the same semantics as heatmap-color -- that increases the likelihood that heatmap-color is not some odd exception (and as we know, the odd exceptions are disproportionately expensive to maintain).
> 
> Since we now have two similarly-behaving properties, it's worth finding a generic way to capture that the style spec and replace the property name comparisons here with the use of a semantic property like isDataDriven uses. Such a property would probably be useful for Studio as well. As a schema design question, it intersects with #4194 -- although in light of this new information, I'm not sure my proposal there is still the best choice. @lbud do you want to think about this and come up with a proposal?

I'm working on this (https://github.com/mapbox/mapbox-gl-js/issues/6389) but ended up splitting it off this branch as it complicates things quite a bit and conflates two different issues. I'm thinking maybe we should merge this as is (_not_ as part of the 0.45 release) and follow it up immediately with a PR accomplishing https://github.com/mapbox/mapbox-gl-js/issues/6389#issuecomment-376187545. Does that sound alright?

danpe:
👍

Issue #1928, opened 2016-01-13
Fix touch events in EDGE browser


vmikeska:
Touch screen functionality (moving, resizing) is not working with EDGE. 
-Win 10 + EDGE
-Win 10 mobile + EDGE
-Win 8.1 mobile + IE

bhousel:
Cool, I'm getting a device to test this next week.

mb12:
@mourner and @lucaswoj  Is it possible to prioritize this or suggest a workaround to make at least panning work using touch on win 10 laptops (Edge browser only )?  At the moment on Windows laptops, even simple panning does not work with touch. (Its okay if panning does not work on windows mobile.)

I am sure this should be blocking others also from switching entirely to mapbox vector base map.

vmikeska:
My original complaint was about the Mapbox JS and it seems working now in 2.4

mourner:
@mb12 we will get to this eventually, but if you need this urgently, we'll appreciate a pull request!

mb12:
@mourner Thanks for the response.  I have found a workaround that does not require any changes to GL JS.  javascript lets you change prototype of any class  that can be used to convert pointer events to touch events. Using the workaround for now. Will wait for the official fix.

jfirebaugh:
Might be fixed with #4259.

fmairesse:
No it is not fixed. Please reconsider the priority: it has been almost 2 years since the bug declaration and more than 2000 people starred it. This bug makes mapbox-gl.js just unusable on a Windows touch screen, particularly if you make an application with a webview (cannot change the browser in this case).

One thing I noticed: in Edge, if you go to about:flags and set the options "Enable touch events" to "Only when a touch screen is detected" and "Fire compatible mouse events in response to the tap gesture" to "Always off", then you can pan.

fmairesse:
A very simple fix to make the pan work:
body {
    -ms-content-zooming: none;
    overflow: hidden;
}

sethb75:
There's a very large number of Surface Pro tablets out there ... many people use touch with Edge. I have customers complaining that touch events don't work for them.

aptogo:
I can confirm that this is still a real issue with Win 10 + EDGE (41.16299.15.0) on Surface devices and touchscreen laptops. For example, https://www.mapbox.com/mapbox-gl-js/example/center-on-symbol/ - dragging pans the entire page, pinch&zoom zooms the entire page and tapping on an icon doesn't trigger a click event.

[...]

PR #6325, opened 2018-03-13
Fix touch panning in Edge

jfirebaugh:
Touch events are behind a flag in Edge, even on touchscreen devices. 🤷‍♂️ However, Edge _will_ dispatch mouse events in response to touch input -- in certain circumstances which are not well documented. It seems that it depends on the `touch-action` CSS property being present on particular elements. We previously set `touch-action: none` on the `.mapboxgl-canvas-container` element, but this is apparently not enough for Edge. In my tests setting `touch-action: none` on the body element works, as does setting `touch-action: none` on both the canvas container _and_ the canvas. Since we don't want to affect touch-action on the entire page, here's the latter solution.

Note that this doesn't fix pinch zooming or touch rotating. Until touch events are unflagged by default, the only way to support those is via pointer events.

Refs #1928.

### DIFF
--- a/dist/mapbox-gl.css
+++ b/dist/mapbox-gl.css
@@ -28,16 +28,19 @@
     cursor: grabbing;
 }
 
-.mapboxgl-canvas-container.mapboxgl-touch-zoom-rotate {
+.mapboxgl-canvas-container.mapboxgl-touch-zoom-rotate,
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Yep, the `-ms-` prefix is necessary only for IE10.

PR #6335, opened 2018-03-14
Replace exported objects with named exports

### LINKED ISSUE #6293: Replace exported objects with named exports
Follow-up to https://github.com/mapbox/mapbox-gl-js/pull/6196: we should remove all (or as many as possible) instances of the CJS-like exported object pattern with named exports.

Probably the main hurdle in doing this is the fact that we rely on stubbing some methods (e.g. `browser.getImageData()`).

jfirebaugh:
Fixes #6293.

Remaining cases of `export default exported` are the main `index.js` and `style-spec.js` which I am afraid to touch, and `browser.js` and `performance.js` which rely on stubbing.

### DIFF
--- a/src/data/dem_data.js
+++ b/src/data/dem_data.js
@@ -4,7 +4,7 @@ import { RGBAImage } from '../util/image';
 import { warnOnce, clamp } from '../util/util';
 import { register } from '../util/web_worker_transfer';
 
-class Level {
+export class Level {
     dim: number;
     border: number;
     stride: number;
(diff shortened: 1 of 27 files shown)


src/data/dem_data.js
```diff
@@ -44,7 +44,7 @@ register('Level', Level);
 // surrounding pixel values to compute the slope at that pixel, and we cannot accurately calculate the slope at pixels on a
 // tile's edge without backfilling from neighboring tiles.
 
-class DEMData {
+export default class DEMData {
```
anandthakker: I think we should avoid mixing default and named exports https://github.com/mapbox/mapbox-gl-js/issues/6336

PR #6337, opened 2018-03-15
Update CONTRIBUTING.md for ES6 module use

### LINKED ISSUE #6336: Use a consistent style for named vs. default exports
My proposal:
 - No exported "namespace objects" (#6335) -- modules should export either classes or functions, with an occasional exception as needed for stubbing.
 - If a module exports one thing, use a default export 
 - If a module exports multiple things, use named exports only

Whatever we decide, it should be backed up by lint rules.

jfirebaugh:
Refs #6336.


jfirebaugh:
👍

PR #6338, opened 2018-03-15
Don't stop camera on mousedown/touchstart when non-interactive

### LINKED ISSUE #3150: Fix bug allowing camera easings to be stopped with a click on a non-interactive map
**mapbox-gl-js v0.23.0**:
### Steps to Trigger Behavior
1. Create a non-interactive map (interactive: false)
2. Initiate an animated transition, for example with flyTo
3. Click on the map during the transition

Demo: http://codepen.io/anon/pen/xEbZQm
### Expected Behavior

That the map does not react to the mouse click.
### Actual Behavior

The transition is aborted.

jfirebaugh:
Fixes #3150.

## Launch Checklist

### DIFF
--- a/src/ui/bind_handlers.js
+++ b/src/ui/bind_handlers.js
@@ -1,11 +1,8 @@
 // @flow
 
 import { MapMouseEvent, MapTouchEvent, MapWheelEvent } from '../ui/events';
-
 import DOM from '../util/dom';
-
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Agreed. I think we're able to slim down map.js through other refactors, we should merge bind_handlers.js into it.

PR #6341, opened 2018-03-15
drop the mapbox-gl-supported script include

andrewharvey:
nitpicking here, it's not a big deal but mapboxgl.supported() is included with mapbox-gl-js already so in this example loading the plugin doesn't do anything

the plugin can be loaded on it's own when someone want's to test support without or before loading mapbox-gl-js but this example already loads gl js by default so it's not needed.


anandthakker:
👍 good call, thanks

Issue #6309, opened 2018-03-09
Improve build time


anandthakker:
Follow-up to https://github.com/mapbox/mapbox-gl-js/pull/6196

On my laptop, rebuilding takes about 7s:
- 3.2s to build `rollup/build/{index,worker,chunk1}.js`, and then
- 4s just to assemble these into the final bundle

That second step is much slower than it needs to be.  It's using Rollup again, but only to make sure the sourcemaps for `{index,worker,chunk1}.js` get properly consumed/composed into the final sourcemap.  This means we're wasting lots of time re-parsing these ast's even though it's not at all necessary.

mourner:
Here's a perf profile for a build from scratch: https://gist.github.com/mourner/0acdc68648627ac431229d2ad601eefc

A quick scan seems to suggest there are some nice low hanging fruits there.

anandthakker:
See also https://github.com/rollup/rollup/issues/1553

anandthakker:
Tracking with a project board here: https://github.com/mapbox/mapbox-gl-js/projects/11

jfirebaugh:
Rollup 0.58.0 and later includes https://github.com/rollup/rollup/pull/2119. We're currently on 0.57.1; however upgrading to 0.58.2 seems to break our solution for Web Workers:

```
$ /Users/john/Development/mapbox-gl-js/node_modules/.bin/rollup -c --environment BUILD:production

src/index.js, src/source/worker.js → rollup/build/mapboxgl...
(!) Circular dependency: src/style-spec/expression/parsing_context.js -> src/style-spec/expression/compound_expression.js -> src/style-spec/expression/parsing_context.js
(!) Circular dependency: src/style-spec/validate/validate.js -> src/style-spec/validate/validate_function.js -> src/style-spec/validate/validate.js
(!) Circular dependency: src/style-spec/validate/validate.js -> src/style-spec/validate/validate_function.js -> src/style-spec/validate/validate_object.js -> src/style-spec/validate/validate.js
(!) Circular dependency: src/style-spec/validate/validate.js -> src/style-spec/validate/validate_function.js -> src/style-spec/validate/validate_array.js -> src/style-spec/validate/validate.js
(!) Circular dependency: src/style-spec/validate/validate.js -> src/style-spec/validate/validate_layer.js -> src/style-spec/validate/validate_paint_property.js -> src/style-spec/validate/validate_property.js -> src/style-spec/validate/validate.js
[...]

PR #6346, opened 2018-03-15
Low-hanging rollup perf tweaks

anandthakker:
* Don't use rollup-plugin-sourcemaps for main build, since we don't need to ingest incoming sourcemaps there
* Don't use treeshaking for the final build step, since it's not relevant: we're basically just concatenating files here.

### DIFF
--- a/build/rollup_plugins.js
+++ b/build/rollup_plugins.js
@@ -8,7 +8,6 @@ import json from 'rollup-plugin-json';
 import browserifyPlugin from 'rollup-plugin-browserify-transform';
 import brfs from 'brfs';
 import uglify from 'rollup-plugin-uglify';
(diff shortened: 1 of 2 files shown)


anandthakker:
Refs https://github.com/mapbox/mapbox-gl-js/issues/6309

anandthakker:
> Don't use treeshaking for the final build step, since it's not relevant: we're basically just concatenating files here

This change makes the second step of the build run just as fast as doing it manually using `MagicString.Bundle` without Rollup.

mourner:
Upgrading Rollup after #2062 lands (along with other recent improvements) and using the upgraded `remove-flow-types` fork should get us to quite a good level.

PR #6349, opened 2018-03-15
Convert shaders to ES module

mourner:
A better alternative to #6347.


jfirebaugh:
Please ensure this works with mapbox-gl-native too.

anandthakker:
If we go this route we’ll need to use a separate version of
shaders/index.js for node and browser

mourner:
Oh right, I completely forgot this was being used under Node with gl-native. Sorry! I think it's better to leave it as is then.

PR #6350, opened 2018-03-15
Add anchor option to Marker

### LINKED ISSUE #5640: Add 'anchor' option for Markers

jfirebaugh:
Fixes #5640.

## Launch Checklist

### DIFF
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -6,16 +6,25 @@ import LngLat from '../geo/lng_lat';
 import Point from '@mapbox/point-geometry';
 import smartWrap from '../util/smart_wrap';
 import { bindAll } from '../util/util';
+import { type Anchor, anchorTranslate, applyAnchorClass } from './anchor';
 
 import type Map from './map';
 import type Popup from './popup';
 import type {LngLatLike} from "../geo/lng_lat";
 import type {MapMouseEvent} from './events';
 
+type Options = {
+    offset?: PointLike,
+    anchor?: Anchor
+};
+
 /**
  * Creates a marker component
  * @param element DOM element to use as a marker. If left unspecified a default SVG will be created as the DOM element to use.
  * @param options
+ * @param options.anchor A string indicating the Marker's location relative to the coordinate set via {@link Marker#setLngLat}.
+ *   Options are `'middle'`, `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`, `'top-right'`, `'bottom-left'`, and `'bottom-right'`.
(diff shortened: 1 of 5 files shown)


src/ui/marker.js
```diff
@@ -6,34 +6,46 @@ import LngLat from '../geo/lng_lat';
...
     _element: HTMLElement;
     _popup: ?Popup;
     _lngLat: LngLat;
     _pos: ?Point;
 
-    constructor(element: ?HTMLElement, options?: {offset: PointLike}) {
+    constructor(element: ?HTMLElement, options?: Options) {
         bindAll(['_update', '_onMapClick'], this);
 
+        this._anchor = options && options.anchor || 'middle';
```
mollymerp: should we add a check here so that if a user provides something like `new mapboxgl.Marker({anchor: 'left'})` it is valid instead of having to use `new mapboxgl.Marker(null, {anchor: 'left'})`?

if we don't want the former to be valid, we should fire a better error because right now it is just `Uncaught TypeError: Cannot read property 'add' of undefined`

mollymerp:
Hmm actually I think the offset's might be backwards
In these screenshots the red dots are markers with [0, 0] offset and `anchor: middle`
_anchor: left_
screen shot 2018-03-16 at 3 25 23 pm

_anchor: right_
image

the docs state: 
>A string indicating the Marker's location relative to the coordinate set via {@link Marker#setLngLat}.

I would expect the results to be flipped 🤔

jfirebaugh:
The behavior is chosen for consistency with `icon-anchor`. The anchor is the part of the popup/marker/icon that's closest to its coordinate.

Though now looking at [`icon-anchor`](https://www.mapbox.com/mapbox-gl-js/style-spec/#layout-symbol-icon-anchor), I see that it should be `'center'` rather than `'middle'`.

mollymerp:
gotcha – makes sense when looking with a popup that has the same anchor setting (blue dot is the one with the `top-left` anchor):

image

src/ui/marker.js
jfirebaugh: Gonna send a separate PR for this in a sec.

cjbara:
Any chance this will be released soon?

Issue #1531, opened 2015-09-23
Fix rendering of long symbol text lines


bsudekum:
I'm seeing a rendering issue associated with long single line labels.

Text anchored centered:

![](https://cldup.com/KSOISpDvJH.png)

Text anchor left/right:
![](https://cldup.com/75zGclhV0Y.png)

I thought this was a tile bounds issue, but decreasing the font size does not fix the issue.

Here is a copy of the style: https://cloudup.com/confY1gZ6Oe

jfirebaugh:
This was first encountered in gl-native, right? Is it an issue in both native and js?

bsudekum:
@jfirebaugh yes, these screenshots are in gl-js

bsudekum:
To clarify, the above screenshots are from a gl-js map.

ansis:
When the offset from the anchor is added to the buffers there aren't enough bits to store the full value so it ends up wrapping. https://github.com/mapbox/mapbox-gl-js/blob/971b956670d0e89039c22618410adb0d33cd1156/js/data/symbol_bucket.js#L45

```
maxOffset = 32768
unitsPerPixel = 64
referenceFontSize = 24
maximumTextMaxWidth = maxOffset / unitsPerPixel / referenceFontSize = 21.33
```

We could either:
- add a maximum max-text-width value of 21em
- reduce the subpixel precision (currently 1/64 pixels) to enable wider labels

@bsudekum do you actually need wider labels or is it ok to fix this with forced wrapping?

ramunasd:
@ansis FYI, event max-text-width value of 10 doesn't helps, at least with non-latin strings

jfirebaugh:
@ChrisLoer @ansis Is this fixed with recent text rendering changes?

ChrisLoer:
@jfirebaugh We've still got the same overflow issue in the rendering. Collision detection can now handle labels that are significantly wider than their tile, so if we were to fix the rendering issue by decreasing subpixel precision, I don't think we'd introduce a new collision problem.

ChrisLoer:
#5150 has landed, so it's now possible for us to drop subpixel precision to support wider labels. I still don't know what the optimal trade off between max-label-width and minimum-precision is. Dropping from 1/64px to 1/32px in order to double the possible width of labels seems like it would probably be fine.

james-wagner11:
[...]

PR #6375, opened 2018-03-21
Increase maximum label width by factor of 2

ChrisLoer:
This is a partial fix for #1531 -- wide labels can still overflow (at which point they stop rendering correctly), but they can now be twice as wide, which will hopefully cover a much wider variety of use cases. #5150 took care of the collision detection problem, so there's no added work necessary for doing collision detection on labels that are wider than the tile's symbol buffers. Also, we no longer use tile clipping for symbols, so there's no problem with labels bleeding far outside their origin tile (err, except for api-gl, wide labels will get rendered-correctly per-tile, but a very wide label in a neighboring tile may be clipped if it doesn't make it into the symbol buffers).

This decreases our glyph-placement precision from 1/64th pixel to 1/32nd pixel, which still seems plenty high. I can't notice any difference in rendering, and all of the render tests pass without change. Doing another halving to 1/16th pixel precision still looks fine to my eyes, but it causes two render tests to fail on (very slight) glyph position changes. For now, 1/32nd seems like a good conservative choice.

Test result before fix:

screenshot 2018-03-21 14 10 25

/cc @ansis @jfirebaugh @james-wagner11


james-wagner11:
Before:

After:


Thank you @ChrisLoer! I haven't found any labels that still overflow. The additional 2x width makes a big difference for us.

ChrisLoer:
@james-wagner11 that's great! Thanks for posting those pictures, and happy this turned out to be (knock on wood) an easy one.

PR #6380, opened 2018-03-22
Don't use 'const' in bundle boilerplate code

anandthakker:
This would break the bundle in IE11


jfirebaugh:
There's still a `let` a few lines up. Isn't that an issue too?

anandthakker:
Ah dang it, yeah. Good catch.

anandthakker:
I’ll add a build test for this too.

Issue #1953, opened 2016-01-20
Android devices, performance and ppi


kristfal:
Hey,

Summary: 

On (mobile) devices with very high ppi, enable a method to render at an lower resolution than the device's native ppi and upscale it.

Long story:

Mapbox-gl-js fares very well on iOS devices performance wise (with #1606).

The story is a bit different on Android devices. 

The most interesting findings we've had so far is that a Samsung Galaxy S3 (Released March 2013) is roughly 10% slower FPS wise than a Samsung Galaxy S6 (Released Q2 2015) under similar circumstances.

The GL draw calls take roughly equal time on the S3 and S6, despite a much faster CPU and GPU on the S6. The reason for this lack of performance gains over the S3 seems to be the newer device's PPI. The S6 has a PPI of 577 vs 306 for the S3, and for ref, 326 for iPhone 6. 

This pattern repeats itself for almost all new Android devices, [and the race is still going to 800 ppi and beyond](http://www.sonymobile.com/global-en/products/phones/xperia-z5-premium/). Android screen resolution is often 'unnecessary' high, which in turn leads to poor web-gl performance.

It is possible to downscale the rendering of mapbox-gl-js in html/css without touching the internal js code, but it also requires highjacking the touch / mouse handlers and scale down symbols and text in the style to make it work properly. We're following this approach, and downscaling to an effective 326 ppi on any device with a higher native ppi. In the case of the S6, we're seeing an effective 3.5x increase in performance without any notices from users.

This workaround is probably outside the scope of your average mapbox-gl-js user. For gl-js not to be a performance hog on mobile (android) devices by default, device PPI should probably handled.

mourner:
That's a valid feature request. Want to look into turning the suggestion into a pull request? I wonder how easy it is to make the ratio configurable as opposed to always on based on the screen.

kristfal:
@mourner I was leaning towards 'max PPI / devicePixelRatio' option in the map config that would downscale only if the threshold was surpassed, but do think it would be better to add a 'target PPI / devicePixelRatio' option that would force a ratio regardless of native screen spec?

I haven't done any changes in the mapbox-gl code yet, just hacks in the app containing the map, but I'll look into it – probably starting this weekend.

SandroGrzicic:
Hi,

we are using mapbox gl js for both android and iOS devices and we just want to add a +1 on this feature request.

wprater:
Is this solution one that's still being considered?

[...]

PR #6395, opened 2018-03-25
Add devicePixelRatio option to config

sniok:
Makes devicePixelRatio configurable #1953 

By default config gets actual devicePixelRatio from window if no value has been set to devicePixelRatio config property.

Changes

- [x] Add a devicePixelRatio setting to config.js which defaults to browser.devicePixelRatio
- [x] Add a devicePixelRatio getter and setter to index.js
- [x] Change all uses of browser.devicePixelRatio to config.devicePixelRatio

## Launch Checklist

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -60,7 +60,7 @@ const exported = {
     /**
      * Gets and sets map's pixel ratio
      * 
-     * @var {number} devicePixelRation
+     * @var {number} devicePixelRatio
      * @example
      * mapboxgl.devicePixelRatio = 2
      */
@@ -69,7 +69,7 @@ const exported = {
     },
 
     set devicePixelRatio(devicePixelRatio: number) {
-        Object.defineProperty(config, 'DEVICE_PIXEL_RATIO', { value: devicePixelRatio });
+        config.DEVICE_PIXEL_RATIO = devicePixelRatio;
     },
 
     workerUrl: ''
--- a/src/index.js
+++ b/src/index.js
@@ -57,6 +57,21 @@ const exported = {
         config.ACCESS_TOKEN = token;
     },
 
+    /**
+     * Gets and sets map's pixel ratio
+     * 
+     * @var {number} devicePixelRation
+     * @example
+     * mapboxgl.devicePixelRatio = 2
+     */
+    get devicePixelRatio() {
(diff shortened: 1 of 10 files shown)


src/index.js
```diff
@@ -55,6 +55,21 @@ const exported = {
...
+     * @var {number} devicePixelRation
+     * @example
+     * mapboxgl.devicePixelRatio = 2
+     */
+    get devicePixelRatio() {
+        return config.DEVICE_PIXEL_RATIO;
+    },
+
+    set devicePixelRatio(devicePixelRatio: number) {
+        Object.defineProperty(config, 'DEVICE_PIXEL_RATIO', { value: devicePixelRatio });
```
ryanhamley: is there a reason to use `Object.defineProperty` here instead of simply assigning to `config.DEVICE_PIXEL_RATIO` the way that `accessToken` does? I'd say it's better to avoid having two separate ways of assigning to config properties here and unless there's a specific reason to favor `defineProperty`, I'd side with the simplest code possible.

src/util/config.js
```diff
@@ -1,15 +1,21 @@
...
+  ACCESS_TOKEN: ?string,
+  DEVICE_PIXEL_RATIO: number,
 |};
 
 const config: Config = {
     API_URL: 'https://api.mapbox.com',
     REQUIRE_ACCESS_TOKEN: true,
-    ACCESS_TOKEN: null
+    ACCESS_TOKEN: null,
+    get DEVICE_PIXEL_RATIO() {
```
ryanhamley: Why use a getter method instead of just assigning the property directly? Again, I'd err on the side of consistency and simplicity here unless there's a specific reason not to.

sniok: Hi! The reason behind this was to save the behavior of browser.getDevicePixelRatio:
```js
get devicePixelRatio() { return window.devicePixelRatio; },
```
so that if pixel ratio wasn't set in config it would use current value of window.devicePixelRatio and not the value that was present at the moment when config object was created.

It's probably not a big deal. I can't think of a use case when pixel ratio would suddenly change (apart from some unit tests). So I am going to stick to the simpler solution.

ryanhamley:
This hasn't had any activity in quite awhile but there seems to be interest in this feature, so I'm going to close this and create a new PR.

PR #6417, opened 2018-03-29
offset the popup around the default marker

andrewharvey:
## Launch Checklist

if you create a default marker with a popup it looks like this:
```js
var marker = new mapboxgl.Marker()
    .setLngLat([0, 0])
    .setPopup(new mapboxgl.Popup().setText('hello hi');)
    .addTo(map);
```

selection_864

with this PR it sets the offset of the popup so the tip touches the outside of the marker icon:

selection_863
selection_862
selection_861
selection_860
selection_859

This is only the default when not providing your own Marker element and no Popup offset. It can still be overridden:

```js
var marker = new mapboxgl.Marker()
    .setLngLat([0, 0])
    .setPopup(new mapboxgl.Popup({offset: [0, 0]}).setText('hello hi');)
    .addTo(map);
```

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -44,6 +44,7 @@ export default class Marker {
     _lngLat: LngLat;
     _pos: ?Point;
     _color: ?string;
+    _defaultMarker: boolean;
 
     constructor(options?: Options) {
         // For backward compatibility -- the constructor used to accept the element as a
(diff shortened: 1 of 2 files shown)


andrewharvey:
> Thanks! Can you add some tests please?

@jfirebaugh I've added unit tests to ensure that the offsets for the different anchors are within a reasonable range, plus some more unit tests to test all anchor positions. This is on top of testing the debug page manually.

Issue #6372, opened 2018-03-21
Canvas source initialization from HTML element, instead of element id


ryanbaumann:
**mapbox-gl-js version**: 0.44.1

# Background

The Canvas source type supports initialization from a canvas HTML element `id` only.  https://www.mapbox.com/mapbox-gl-js/style-spec#sources-canvas-canvas

# Feature Request

Allow initialization of Canvas source from an HTML canvas element, the same way that a mapboxgl map object can be bound directly to an HTML element https://www.mapbox.com/mapbox-gl-js/api/#map.

# Use case

Canvas elements in an application framework do not have an HTML `id` element.  Adding HTML unique `id`s to each Canvas element would add complexity to the application, instead of just creating the canvas element and passing to the Canvas source.

cc @lbud

lbud:
This is definitely feasible. It would mainly entail:
* Changing CanvasSource so that it can optionally [read directly from the element](https://github.com/mapbox/mapbox-gl-js/blob/6b848e5f6bcb1188d51691c456054dc8095c053d/src/source/canvas_source.js#L78) if `options.canvas` is an `HTMLCanvasElement`, and omitting the canvas element from [`serialize`](https://github.com/mapbox/mapbox-gl-js/blob/6b848e5f6bcb1188d51691c456054dc8095c053d/src/source/canvas_source.js#L182) in that case
* Modifying the [`CanvasSource#canvas` specification](https://github.com/mapbox/mapbox-gl-js/blob/6b848e5f6bcb1188d51691c456054dc8095c053d/src/style-spec/reference/v8.json#L429) to accept _either_ a string or HTMLCanvasElement. Currently there are no other union types in the spec — all `type` values are either a single string-represented type or an enum type represented by an object — so we'd need to determine the best way to represent a union type and validate it

PR #6403, opened 2018-03-26
Canvas source initialization from HTML element

lbud:
* Modifies `CanvasSource#canvas` property to accept _either_ a string denoting the ID of the canvas element from which to read (as it previously functioned) or an HTMLCanvasElement (fixes https://github.com/mapbox/mapbox-gl-js/issues/6372)
* Introduces the concept of union types to the style spec, denoted similarly to enum types ([see `canvas`](https://github.com/mapbox/mapbox-gl-js/blob/7f5b208cca2db34a39bb76d39be99d63a7dcb1fb/src/style-spec/reference/v8.json#L428-L449))

cc @ryanbaumann 

## Launch Checklist

 - [x] briefly describe the changes in this PR
 - [x] write tests for all new functionality
 - [x] document any changes to public APIs (will be autogenerated in mb-pages update)
 - [x] manually test the debug page

Issue #5545, opened 2017-10-27
Discourage developers from adding canvas sources to style JSON


1ec5:
The [`canvas` source type](https://www.mapbox.com/mapbox-gl-js/style-spec/#sources-canvas) documentation should state that `canvas` sources are only to be used at runtime in JavaScript code, never in style JSON. This source type is inherently platform-specific: some platforms like iOS have no concept of a rendering surface associated with a unique identifier. A platform-specific source type has no place in a file format that’s intended to be portable across platforms.

Ideally, `canvas` would be omitted entirely from the style specification. However, GL JS uses the style specification to document its runtime styling API, so that isn’t an option right now.

/cc @anandthakker

jfirebaugh:
1c719f159c9e6ce0695d7f67020caeec45017338

PR #6424, opened 2018-03-29
Canvas source initialization from HTML element

lbud:
* Modifies `CanvasSource#canvas` property to accept _either_ a string denoting the ID of the canvas element from which to read (as it previously functioned) or an HTMLCanvasElement (fixes https://github.com/mapbox/mapbox-gl-js/issues/6372)
* Introduces validation warnings — while I doubt we have many users initializing canvas sources from stylesheets rather than via `map.addSource`, it's possible (and in fact this was done by [our own debug example](https://github.com/mapbox/mapbox-gl-js/blob/7bc23b9b9df28afc54252a2d68d450aa24580c0b/debug/canvas.html)). It seems abrupt to deprecate this quickly and without warning, so validation now warns and does not fail on canvas sources added in the stylesheet.

Closes https://github.com/mapbox/mapbox-gl-js/issues/6372.
Supersedes https://github.com/mapbox/mapbox-gl-js/pull/6403.
Ref https://github.com/mapbox/mapbox-gl-js/issues/5545.

Docs now show a separate `CanvasSourceOptions` typedef (🙇‍♀️ thank you @jfirebaugh):
image

## Launch Checklist


lbud:
> One of those advantages is clarity: it's simpler to describe what style is allowed to be if it's just "a style spec document", rather than "a style spec document, plus canvas sources, custom sources, and custom layers".

I agree the clarity would be nice, but I think since at the moment our current best/only recommendation for "how do I persist my overlays while changing the basemap" is to use `setStyle` well, we should at least for now allow canvas sources in `setStyle` without emitting warnings 🛠

anandthakker:
Yeah, I guess deprecating this without yet having an alternative doesn't really make sense

PR #6425, opened 2018-03-30
Round-trip linear interpolators

1ec5:
Linear interpolators should be serialized as such rather than as exponential interpolators with a base of 1.0. These test fixture changes are required for mapbox/mapbox-gl-native#11562.

## Launch Checklist

 - [ ] post benchmark scores

/cc @anandthakker @ChrisLoer

### DIFF
--- a/test/integration/expression-tests/let/zoom/test.json
+++ b/test/integration/expression-tests/let/zoom/test.json
@@ -32,7 +32,7 @@
       30,
       [
         "interpolate",
-        ["exponential", 1],
+        ["linear"],
         ["zoom"],
         0,
         ["var", "z0_value"],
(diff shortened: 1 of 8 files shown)


test/integration/expression-tests/let/zoom/test.json
```diff
@@ -23,22 +23,6 @@
       "isZoomConstant": false,
       "type": "number"
     },
-    "outputs": [20],
```
ChrisLoer: Why did the whole "serialized" section disappear here (and in zoom/basic)?

1ec5: Oh, I thought `serialized` is only needed when it differs from `expression`. I can add it back.

ChrisLoer: That's a good idea! But no it doesn't do that (yet)...

Issue #6253, opened 2018-03-01
Expression tests should check "serialized"


jfirebaugh:
https://github.com/mapbox/mapbox-gl-js/pull/6181 added functionality to the expression tests to verify that the actual "serialized" representation matches an expected value. However, it omitted a serialization mechanism in the JS implementation; the JS test harness ignores the expected value for existing tests, and bootstraps newly added tests with an "expected" serialization that's copied from the input expression.

This introduces a new source of friction in the gl-js/gl-native workflow: it's easy to create a new expression test in gl-js that fails unexpectedly when pulled into gl-native because the serialized representation is not exactly equivalent to the input expression. When this happens, you have to either ignore that test in gl-native (even if it would otherwise be passing), or pause your current work and update the expected serialization in a gl-js PR.

/cc @ChrisLoer

anandthakker:
Yeah -- it wouldn't be too hard to add `Expression#serialize()` to GL JS, but as of now, we'd be doing it solely for the purpose of these tests.  Still, it's not that much code to add, so it's probably worth doing.

PR #6427, opened 2018-03-30
to-rgba of rgba serializes as a literal

1ec5:
#6388 causes mapbox/mapbox-gl-native#11565 to fail because the `to-rgba` operator cancels out the `rgba` operator, at least in mbgl. This change updates the test fixture so that mapbox/mapbox-gl-native#11565 can pass.

## Launch Checklist

 - [ ] post benchmark scores

/cc @jfirebaugh @ChrisLoer


anandthakker:
Another vote in favor of #6253

Issue #6298, opened 2018-03-07
Implement `SymbolStyleLayer#queryRadius()` (queryRenderedFeatures fix)


ChrisLoer:
The current implementation just returns "0":

https://github.com/mapbox/mapbox-gl-js/blob/31d92db36398cddbf8a511af261861460770d894/src/style/style_layer/symbol_style_layer.js#L76-L78

This means that symbols that "bleed over" into the tile next to them won't show up in a `queryRenderedFeatures` query that intersects their geometry but doesn't intersect their original tile. This wasn't much of an issue before #5150 because the symbol would generally be included in both tile's collision indices.

Once #6036 goes in, we'll have an upper bound on the amount of "pitch scaling" that can apply to text. We can implement `queryRadius` here by going through all the symbols in the bucket and keeping the max dimensions.

/cc @ansis

ChrisLoer:
Actually, I think I'd rather fix this with https://github.com/mapbox/mapbox-gl-js/issues/5475

ChrisLoer:
Fixed in #6497: `queryRadius == 0` is now correct.

PR #6429, opened 2018-03-30
only index features strictly within tile boundaries

ansis:
Instead of indexing all features for querying, index only those that are actually within the tile boundaries. This prevents duplicate results from being returned for point features. This is only a partial fix for the duplicate results problem (lines and polygons can still return
duplicate results).

Previously we relied on tile buffers for cross-tile indexing. But now that we have https://github.com/mapbox/mapbox-gl-js/pull/6036 we don't need that anymore.

I noticed this while porting https://github.com/mapbox/mapbox-gl-js/pull/6036 to -native. These test changes would be useful to have for that port.

## Launch Checklist

 - [x] post benchmark [scores](https://user-images.githubusercontent.com/1421652/38152270-fa525cae-3434-11e8-819d-a3bd1b21c31c.png)

### DIFF
--- a/src/data/feature_index.js
+++ b/src/data/feature_index.js
@@ -83,7 +83,12 @@ class FeatureIndex {
                 bbox[3] = Math.max(bbox[3], p.y);
             }
 
-            this.grid.insert(key, bbox[0], bbox[1], bbox[2], bbox[3]);
+            if (bbox[0] < EXTENT &&
+                bbox[1] < EXTENT &&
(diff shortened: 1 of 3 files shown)


ansis:
> My reading is that we'll still be in the same situation described in #6298: we may miss querying some tiles altogether, but if we query a tile for symbols, it won't be affected by this change because symbol queries don't touch the non-symbol grid.

Yep, this is how I understand it as well.

Issue #6389, opened 2018-03-23
Proposal: add `expression-dependency` key to spec properties that depend on internal calculations


lbud:
Following from https://github.com/mapbox/mapbox-gl-js/pull/6303#discussion_r173597559, this is a proposal to add a new spec key, `expression-dependency`, for paint/layout properties whose display is linked to a special calculated property. These keys have string vales that specify the name of the internally calculated property upon which they depend, so currently those would be (including https://github.com/mapbox/mapbox-gl-js/pull/6303):
```
"heatmap-color": {
    "expression-dependency": "heatmap-density"
}
"line-gradient": {
    "expression-dependency": "line-progress"
}
```
By adding a spec key we can remove special casing code, pave the way for future related properties, and move these specific instructions out of the docstrings ([e.g.](https://github.com/mapbox/mapbox-gl-js/blob/3c07b72a59d78bc6240a8e3d77a54bec7619af5d/src/style-spec/reference/v8.json#L3607)) and into generated documentation.

Thoughts, semantic or otherwise? @mapbox/gl-core

lbud:
We'd also want to add a special key (`"calculated"`?) to the "dependencies" in the expressions section (https://github.com/mapbox/mapbox-gl-js/blob/3c07b72a59d78bc6240a8e3d77a54bec7619af5d/src/style-spec/reference/v8.json#L2322-L2329)

anandthakker:
I wonder if it makes sense to consider the other expression-related style-spec properties at the same time:
 - `function: "interpolated" | "piecewise-constant"`
 - `zoom-function`
 - `property-function`

In particular, the latter two say something similar to what we're trying to say for `heatmap-density` / `line-progress`: `zoom-function: true` means an expression may refer to `["zoom"]`; `property-function: true` means an expression may refer to `["get", key]`, `["has", key]`, `["geometry-type"]`, `["id"]`, or `["properties"]`.

What about something like:

```js
  expression: {
    interpolated: true | false // replaces function: "interpolated" | "piecewise-constant", 
                               // and defaults to true for number, color, or array
                               // of number/color
    parameters: Array<"zoom" | "feature" | "heatmap-density" | "line-progress">
  }
```

So, e.g.:

```js
"circle-color": {
    "expression": {
        "parameters": ["zoom", "feature"]
    }
}
"heatmap-color": {
    "expression": {
        "parameters": ["heatmap-density"]
    }
}
"line-gradient": {
    "expression": {
[...]

PR #6430, opened 2018-03-30
Only evaluate CrossFaded properties at integer zooms

### LINKED ISSUE #6390: Incorrect cross-fading at non-integer zoom levels
http://jsbin.com/gatejucoti/edit?html,output

crossfade

This is because the crossfade interpolation factor `t`  is designed to cross-fade the property across integer zoom levels, so when you have a non-integer zoom level, the `from` and `to` values of the `CrossFadedProperty` will be incorrect at integer zooms after a non-integer stop. 

I think we should disallow non-integer stops for these properties, because they're non-interpolatable so they won't work anyways with data-driven expressions, so I think it may cause confusion if they're supported for camera expressions. 

@nickidlugash @natslaughter do you have strong feelings about this?

cc @ansis @anandthakker

mollymerp:
fixes #6390 

⚠️ this might be a breaking change for some people even though it is also a bug fix

question: should we have a style spec field that indicates if the property is cross-faded or not to avoid having to hard code the property name string check?

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2860,7 +2860,7 @@
       "function": "piecewise-constant",
       "zoom-function": true,
       "transition": true,
-      "doc": "Name of image in sprite to use for drawing image fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512).",
+      "doc": "Name of image in sprite to use for drawing image fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512). Note that this property does not support zoom-dependent expressions with non-integer zoom inputs.",
       "sdk-support": {
         "basic functionality": {
           "js": "0.10.0",
@@ -2975,7 +2975,7 @@
       "function": "piecewise-constant",
       "zoom-function": true,
       "transition": true,
-      "doc": "Name of image in sprite to use for drawing images on extruded fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512).",
+      "doc": "Name of image in sprite to use for drawing images on extruded fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512). Note that this property does not support zoom-dependent expressions with non-integer zoom inputs.",
       "sdk-support": {
         "basic functionality": {
           "js": "0.27.0",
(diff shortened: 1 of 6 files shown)


src/style-spec/validate/validate_expression.js
```diff
@@ -18,5 +18,12 @@ export default function validateExpression(options: any) {
         return [new ValidationError(options.key, options.value, 'Invalid data expression for "text-font". Output values must be contained as literals within the expression.')];
     }
 
+    if (options.expressionContext === 'property' && expression.value.kind === 'camera' && options.propertyKey.match(/(-pattern)|(-dasharray)/)) {
```
ansis: Would it be slightly clearer to explicitly list all the properties this applies to instead of using a regular expression?

jfirebaugh: It's another case where #6389 (specifically https://github.com/mapbox/mapbox-gl-js/issues/6389#issuecomment-376245679) would help.

jfirebaugh:
I'm slightly worried about the consequences of changing this. We had at least one case where we made validation stricter in a similar area and later had to roll it back because existing styles stopped working (https://github.com/mapbox/mapbox-gl-style-spec/pull/413).

anandthakker:
> then we also can (and, I think, should) address this by making sure that in CrossFadedProperty#possiblyEvaluate, we only evaluate the expression at integer zoom levels

@mollymerp is the above correct?  If so, and we want to avoid risking breaking existing styles, we could just do that instead.

src/style-spec/reference/v8.json
```diff
@@ -2965,7 +2965,7 @@
       "function": "piecewise-constant",
       "zoom-function": true,
       "transition": true,
-      "doc": "Name of image in sprite to use for drawing images on extruded fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512).",
+      "doc": "Name of image in sprite to use for drawing images on extruded fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512). Note that this property does not support zoom-dependent expressions with non-integer zoom inputs.",
```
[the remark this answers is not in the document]
mollymerp: 👍 yep that's more concise.

PR #6441, opened 2018-04-02
Force touchstart, touchmove, and wheel listeners to be non-passive

anandthakker:
Cherry pick of 02d652847c01083e9317b6773f91f4fd6edf00e3

Yes, Chrome, yes WebKit: we really do intend to prevent the default scroll behavior for these events.

This is the only way to silence the console warnings from Chrome and support iOS Safari 11.3+.

### DIFF
--- a/src/ui/bind_handlers.js
+++ b/src/ui/bind_handlers.js
@@ -1,7 +1,7 @@
 // @flow
 
-const DOM = require('../util/dom');
 const Point = require('@mapbox/point-geometry');
+const DOM = require('../util/dom');
 
 import type Map from './map';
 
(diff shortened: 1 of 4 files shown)


anandthakker:
Hmm, something is still going wrong here.  Change the css in `debug/index.html` to:

```css
        html, body { height: 200vh; }
        #map { height: 50vh; }
```

and then try to pan the map: the page scrolls too.  Adding `console.log(e.defaultPrevented)` after the `preventDefault()` call in `DragPanHandler#_onMove()` prints `false`.

If I change the `touchmove` listener in `bind_handlers.js` to also use `{passive: false}` (even though AFAICT it shouldn't need to), then scrolling is correctly prevented.

anandthakker:
More puzzling: the following _doesn't_ reproduce the problem:

```html
<html>
    <head>
        <meta name="viewport" content="width=device-width, initial-scale=1">
    </head>
    <body>
        <div style="width: 100%; height: 300px; background-color: #aa00aa"></div>
        <div style="width: 100%; height: 300px; background-color: #00aaaa"></div>
        <div style="width: 100%; height: 300px; background-color: #aa00aa"></div>
        <div style="width: 100%; height: 300px; background-color: #00aaaa"></div>
        <div style="width: 100%; height: 300px; background-color: #aa00aa"></div>
        <div style="width: 100%; height: 300px; background-color: #00aaaa"></div>
        <div style="width: 100%; height: 300px; background-color: #aa00aa"></div>
        <div style="width: 100%; height: 300px; background-color: #00aaaa"></div>
        <div style="width: 100%; height: 300px; background-color: #aa00aa"></div>
        <script>
            const divs = document.querySelectorAll('div')
            for (let i = 0; i < divs.length; i++) {
                const div = divs[i];
                div.addEventListener('touchstart', function () {}, {passive: true});
                div.addEventListener('touchmove', function (e) {
                    console.log(`div ${e.type}`, e.cancelable);
                }, {passive: true});
            }

            window.document.addEventListener('touchstart', handleTouchEvent, {passive: false});
            window.document.addEventListener('touchmove', handleTouchEvent, {passive: false});

            function handleTouchEvent(e) {
                const cancelable = e.cancelable;
                e.preventDefault();
                const defaultPrevented = e.defaultPrevented;
                console.log(`document ${e.type}`, {cancelable, defaultPrevented});
            }
        </script>
    </body>
</html>
```

anandthakker:
Filed: https://bugs.webkit.org/show_bug.cgi?id=182521

Issue #6440, opened 2018-04-02
RENDER WARNING: texture bound to texture unit 0 is not renderable. It maybe non-power-of-2 and have incompatible texture filtering.


jfirebaugh:
https://jsfiddle.net/m77nkgme/1/
https://github.com/mapbox/mapbox-gl-js/issues/3312#issuecomment-377937751

PR #6444, opened 2018-04-02
only use mipmaps for power-of-two textures

ansis:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/6440

The WMS source in that example returns 400x400 tiles. WebGL only supports mipmaps for power-of-two size textures. This fix just disables mipmap generation and filtering for textures that are not power-of-two.

Alternative: we could either upscale the image to a power-of-two texture or add padding and draw only part of the texture but both of these solutions are significantly more complex and might not look better. Mipmapping only helps rasters look a bit better when they are significantly underzoomed (zooming out and the tiles get smaller).

I wrote a render test for this but it doesn't fail without the fix (I guess the node module implementation doesn't error the same way) so I didn't include it here. I don't have any other ideas on how to test this.

## Launch Checklist

 - [tried] write tests for all new functionality
 - [ ] post benchmark scores

### DIFF
--- a/src/render/texture.js
+++ b/src/render/texture.js
@@ -40,39 +40,52 @@ class Texture {
     format: TextureFormat;
     filter: ?TextureFilter;
     wrap: ?TextureWrap;
+    useMipmap: boolean;
     powerOfTwoSize: boolean;
 
-    constructor(context: Context, image: TextureImage, format: TextureFormat, premultiply: ?boolean) {
+    constructor(context: Context, image: TextureImage, format: TextureFormat, premultiply: ?boolean, useMipmap: ?boolean) {
         this.context = context;
-
-        const {width, height} = image;
-        this.size = [width, height];
         this.format = format;
-
         this.texture = context.gl.createTexture();
-        this.update(image, premultiply);
+        this.update(image, premultiply, useMipmap);
     }
 
-    update(image: TextureImage, premultiply: ?boolean) {
+    update(image: TextureImage, premultiply: ?boolean, useMipmap: ?boolean) {
         const {width, height} = image;
-        this.size = [width, height];
-
+        const resize = !this.size || this.size[0] !== width || this.size[1] !== height;
         const {context} = this;
         const {gl} = context;
+
+        this.useMipmap = Boolean(useMipmap);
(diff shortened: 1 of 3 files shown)


src/render/texture.js
```diff
@@ -40,12 +40,14 @@ class Texture {
...
     filter: ?TextureFilter;
     wrap: ?TextureWrap;
+    powerOfTwoSize: boolean;
 
     constructor(context: Context, image: TextureImage, format: TextureFormat, premultiply: ?boolean) {
         this.context = context;
 
         const {width, height} = image;
         this.size = [width, height];
+        this.powerOfTwoSize = width === height && (Math.log(width) / Math.LN2) % 1 === 0;
```
jfirebaugh: Is there a reason to use state rather than compute this as needed? As state it could be wrong if the size is changed via `update`.

mourner: BTW a common neat and fast way to check if a number is a power of two is `x & (x - 1) === 0`

ansis:
@jfirebaugh I moved setting `powerOfTwoSize` from the constructor to the `update` to fix the bug (
3936aaa). I can switch it to a method if you think that could avoid future bugs.

I also added 8b5ebb0 which moves update-without-reallocation logic into `Texture` so that we can handle that and mipmap generation within `Texture` instead of within `RasterTileSource`. I'm not completely sure if this is better of it's just less direct.

src/render/texture.js
```diff
@@ -40,35 +40,51 @@ class Texture {
     format: TextureFormat;
     filter: ?TextureFilter;
     wrap: ?TextureWrap;
+    useMipmap: boolean;
+    powerOfTwoSize: boolean;
 
-    constructor(context: Context, image: TextureImage, format: TextureFormat, premultiply: ?boolean) {
+    constructor(context: Context, image: TextureImage, format: TextureFormat, premultiply: ?boolean, useMipmap: ?boolean) {
```
jfirebaugh: Rather than add a second optional boolean parameter, let's move `premultiply` and `useMipmap` into an options object so they get descriptive names at the call site and there's no need for an `undefined` placeholder argument.

Issue #5506, opened 2017-10-23
Hover effects when layers are stacked (z-index)


ethanKlein:
Hi. I'm struggling to create a hover effect I believe should be pretty strait-forward and also a common use case for developers using mapbox gl js. 

I have a map with 2 layers. 
Layer 1 is a symbol layer that uses icons to mark points on the map. 
Layer 2 is a symbol layer that shows a score (text) on top of the icon from layer 1.

[Here is a video depicting the issue](http://ethanklein.org/videos/mapbox-hover.mp4)

I use filtering ala `https://www.mapbox.com/mapbox-gl-js/example/hover-styles/` to listen for mouse events and hide and show Layer 2 when appropriate. 

In my code I'm listening for `mouseenter` and `mouseleave` events on Layer 1. My problem is that as Layer 2 gets shown via a change in filter, it then causes a mouseleave event to get fired from Layer 1 because Layer 2 sits above Layer 1. As the mouse moves across Layer 2 the `mouseenter` and `mouseleave` events for Layer 1 get continually fired as the mouse event reaches through the lines of text of Layer 2 down to Layer 1.  This causes flickering of Layer 2 as it gets shown and hidden. 

Things I've tried:

- I've attempted using the 'before' argument `https://www.mapbox.com/mapbox-gl-js/example/geojson-layer-in-stack/` 

- I've attempted to create a new layer (Layer 3) that sits on top of Layer 1 and Layer 2 containing a transparent icon of the same size as Layer 1 and listened for mouse events on Layer 3.

- I've tried toggling the visibility of Layer 2 via setStyle (rather than using filters).

- I've tried including the Layer 2 content as the text-field of Layer 1 and looking for a way to show-on-hover the text-field of a mapbox layer (couldn't figure out... is this possible?)

It seems that Mapbox gl js redraws the layer order when a layer is modified via filtering or style changes.  The redraw seems to always place the affected layer on top!? 

Any direction or advice would be greatly appreciated. Pretty stumped!! 

Also apologies if this is not the correct forum.

Thanks for an amazing product! 

Ethan

jfirebaugh:
Thanks for using Mapbox, and sorry to hear you're running into an issue. In order for us to diagnose this issue, we'll need to be able to reproduce it ourselves. Could you please create a minimal self-contained JSFiddle that demonstrates the issue?

stevage:
[...]

Issue #5475, opened 2017-10-17
Convert queryRenderedSymbols to run entirely in viewport coordinates


ChrisLoer:
Follow up to PR #5150.

`queryRenderedFeatures` converts a viewport-geometry query into tile coordinates and then queries the `FeatureIndex` of each tile for results. That used to include a query to a tile-specific `CollisionTile` for symbol results, but it now forwards symbol queries to the global/viewport-aligned `CollisionIndex`, where the tile unit query geometry is converted back to viewport coordinates.

The current approach works fine and I don't think the round-trip coordinate transformation is a performance bottleneck, but we can probably refactor this to be simpler by moving `queryRenderedSymbols` out of the tile-based `FeatureIndex`.

ChrisLoer:
This doesn't matter that much, but while doing the native port of #5150 it occurred to me that even with the current tile-based query architecture, it's unnecessary to re-project the collision boxes, since the projected box coordinates are already stored in the `GridIndex` and it's easy to modify `GridIndex` to return them as part of a query.

ChrisLoer:
Marking this as a bug because the current tile based implementation of `queryRenderedFeatures` won't include tiles with symbols that "bleed over" into adjacent tiles (if the query only intersects the adjacent tile). Doing the query entirely against the global collision index would solve this problem.

geografa:
Adding this test case as a good example of  where this is failing:
https://jsfiddle.net/x1fLbnzg/1/

Zooming in to markers labeled `F` and `1` demonstrate how the markers that are not clickable when spanning over into adjacent tiles:

![markerthing](https://user-images.githubusercontent.com/134753/38433967-a6a42180-3981-11e8-96fc-f2ca29bd4cba.gif)

![ohnoes](https://user-images.githubusercontent.com/134753/38434032-da5f0986-3981-11e8-95b1-47c9b94446d0.gif)

cc @ChrisLoer @malwoodsantoro

ChrisLoer:
Fixed in #6497.

PR #6461, opened 2018-04-05
Fix race conditions in symbol querying

ChrisLoer:
Addresses hover flicker from issues #5887 and #5506.
Moves all data necessary for symbol querying into FeatureIndex.
Retains old FeatureIndexes if they're associated with a bucket that was used in the current placement.

The approach to fixing these bugs is unfortunately a bit complicated: because GL JS supports "incremental placement", it's possible that one `CollisionIndex` (which we use for querying rendered symbols) is built with data from multiple different versions of the same tile. Placement can't be interrupted for a single bucket of a single tile, so what we do here is hold on to all the data we need to query each bucket that gets included in a given placement. We piggy-back on the existing `bucketInstanceId` mechanism in `CrossTileSymbolIndex` that guarantees a unique ID number for each new version of a bucket.

In the case where one tile has multiple `FeatureIndex`es (because tile updates have arrived in the middle of an incremental placement), we run our symbol query against each of those indices separately, filtering `CollisionIndex` results against the set of bucket ids associated with each `FeatureIndex`. 

Although this is a pretty inefficient way to store/query what is for the most part a bunch of duplicate data, I'm not too worried about it because it's just a temporary added cost during the period of time between a tile update and the next placement. This PR does introduce a lot of shared pointers to data, though -- one way it could go wrong would be to start leaking that data.

I've manually tested this branch against the repro case in #5506, but I haven't come up with any automated tests yet. Although they'll be a bit contrived, I think I can at least add some cases to the `Tile` unit test that try to exercise the functionality.

## Launch Checklist

 - [x] Ran benchmarks, nothing jumped out. How do we post now that anonymous gists are broken?

/cc @ansis @malwoodsantoro @ryanbaumann @lilykaiser

### DIFF
--- a/debug/debug.html
+++ b/debug/debug.html
@@ -82,6 +82,39 @@ map.on('load', function() {
             "text-font": ["Open Sans Semibold", "Arial Unicode MS Bold"]
         }
     });
+
+    map.addLayer({
+        id: "marker",
+        type: "symbol",
+        source: "points",
+        layout: {
+            "icon-image": "bus-11",
+            "icon-size": 2,
+            "icon-ignore-placement": true,
+            "icon-allow-overlap": true
+        }
+    });
+
+    map.addLayer({
+        id: "marker-hover",
+        type: "symbol",
+        filter: ["==", "name", ""],
+        source: "points",
+        layout: {
+            "icon-image": "bus-15",
+            "icon-size": 2,
+            "icon-ignore-placement": true,
+            "icon-allow-overlap": true
+        }
+    });
+
+    map.on('mouseenter', 'marker', function(e) {
+        map.setFilter('marker-hover', ['==', 'name', e.features[0].properties.name]);
+    });
+    map.on('mouseleave', 'marker', function(e) {
+        map.setFilter('marker-hover', ['==', 'name', '']);
+    });
+
 });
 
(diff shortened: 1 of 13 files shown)


ChrisLoer:
I added a hover effect to four bus icons in `debug/debug.html` to make it easier to manually test this functionality. I also set up verbose local logging of the bucket retention/pruning behavior and through that found a problem in my initial implementation: data that had arrived while one placement was in progress but wasn't included in that placement _would_ get used successfully in the next placement, but it would get pruned from `retainedFeatureIndexes` before that placement happened, which would make it disappear from future query results.

ChrisLoer:
I've been working on addressing issue #5475 (stop doing `queryRenderedSymbols` on a per-tile basis) which looks like it may make a fair bit of this code (the tile-based querying and feature index pruning) obsolete. The approach that looks promising to me is to double down on `bucketInstanceId` as the way to track which data should be retained and included in queries, but with all the ownership managed by `CrossTileSymbolIndex` or some other mostly-global data structure. I think it should still be worthwhile to merge this as-is while I work on that larger refactoring.

ChrisLoer:
Looks like this will be superseded by PR #6497.

Issue #6464, opened 2018-04-06
Road name label density badly calculated based on text-size


watbywbarif:
v0.44.1 @ Chrome 65 @ Windows 10:

### Steps to Trigger Behavior

 1. I have created custom map data and style in Casablanca, and as road name labels in Mapbox examples look bad in that area it can most likely be reproduced with your data also.
 2. I have excluded mostly every other layer for this to be more obvious.
 3. I keep everything as a constant, only thing that changes is text-size : stops for label layer. 

### Expected Behavior
![image](https://user-images.githubusercontent.com/2914274/38420078-ef6515b6-39a2-11e8-8588-fddbe33a1383.png)
with:
```
				"text-size": {
					"stops": [ 
						[ 12, 10 ],
						[ 19, 10 ],
						[ 20, 16 ]
					]
				}
```
### Actual Behavior
Now i only change this:

```
                               "text-size": {
					"stops": [ 
						[ 12, 10 ],
						[ 14, 10 ],
						[ 20, 16 ]
					]
				}
```

And result is this:

![image](https://user-images.githubusercontent.com/2914274/38420576-87ddc6a2-39a4-11e8-866e-9d822a035022.png)

How is this possible on zoom level 12, especially with stop [ 14, 10 ]?
And even better how does stop [ 19, 10 ], fixes this problem?
This is simplified example but this effect destroys map quality in certain areas of the globe where there are less POIs and other labels.

watbywbarif:
CollisionBoxes reveal no information.
![image](https://user-images.githubusercontent.com/2914274/38423841-58e097c0-39af-11e8-827a-0d639be4aaaf.png)

I have created similar example in mapbox studio, will this do?

https://api.mapbox.com/styles/v1/watbywbarif/cjfnzy9c62i6n2sk0b660w778.html?fresh=true&title=true&access_token=pk.eyJ1Ijoid2F0Ynl3YmFyaWYiLCJhIjoiY2lzc3c3YjVyMDAxZzJvbnpwem56MGc2NyJ9.WsKJJLOudPa4iZvAJhpwmw#13.2/33.592870/-7.621778/0

![image](https://user-images.githubusercontent.com/2914274/38424434-5aea58ec-39b1-11e8-813c-ac3413b9f044.png)

![image](https://user-images.githubusercontent.com/2914274/38424474-75cd8b02-39b1-11e8-9be7-eb4815fdad6b.png)

mourner:
[...]

PR #6468, opened 2018-04-06
Fix constant folding

### LINKED ISSUE #6253: Expression tests should check "serialized"

anandthakker:
Replaces #6464 
Refs https://github.com/mapbox/mapbox-gl-native/issues/10849#issuecomment-378971127

### DIFF
--- a/src/style-spec/expression/definitions/let.js
+++ b/src/style-spec/expression/definitions/let.js
@@ -17,10 +17,7 @@ class Let implements Expression {
     }
 
     evaluate(ctx: EvaluationContext) {
-        ctx.pushScope(this.bindings);
-        const result = this.result.evaluate(ctx);
-        ctx.popScope();
-        return result;
+        return this.result.evaluate(ctx);
     }
 
     eachChild(fn: (Expression) => void) {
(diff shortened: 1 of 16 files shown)


anandthakker:
I included #6253 in this work because I realized belatedly that it's the most natural way to test the changes.

anandthakker:
> I included #6253 in this work because I realized belatedly that it's the most natural way to test the changes.

Actually I'll just extract this into a separate PR

anandthakker:
Alright, rebased now that `Expression#serlialize()` is in master.

anandthakker:
> I have a lingering instinct that isConstant and friends are missing some simplifying abstraction that would make them more straightforward and less dependent on checking a bunch of highly specific conditionals

Yeah, I've had the same feeling, but haven't had an idea for the right abstraction click into place yet

PR #6472, opened 2018-04-07
compact mapbox wordmark on narrow maps

### LINKED ISSUE #4282: collapsed Mapbox workmark on narrow maps

The static maps API uses a collapsed M workmark in lieu of the full Mapbox workmark when width < 250px (the full workmark is used when width >= 250px).

240x120

250x120

Would it be acceptable for the same logic to apply to the [Mapbox workmark LogoControl](https://github.com/mapbox/mapbox-gl-js/pull/3933) in GL JS?

/cc @mollymerp

andrewharvey:
## Launch Checklist

closes #4282

selection_876

same styling as the static maps api uses:

249x100 2x

 - [ ] post benchmark scores


andrewharvey:
> where is the SVG asset you've inlined from? Just want to make sure we're using the most up to date version and check with our design team.

I took the current inlined svg for the normal wordmark and edited it to only contain the icon part then re inlined it. I didn't include the SVG asset since the full wordmark SVG asset isn't included currently.

It would be nice to have those inline SVGs inserted automatically from the source SVG asset, but I think we should leave that to another issue. (just opened at #6493)

> is the 250 cutoff from API GL only? We're using 640px as the cutoff between compact and normal attribution, and I wonder if it makes sense to stay consistent there.

Yes it's the same as what API GL (static map api) uses. It feels right to use a smaller cutoff than the normal attribution as attribution is usually much wider to begin with.

andrewharvey:
- [x] rebased against master since #6513 merged.
- [x] tested again on the debug page

src/css/mapbox-gl.css
```diff
@@ -196,6 +196,12 @@ a.mapboxgl-ctrl-logo {
     background-image: svg-load('svg/mapboxgl-ctrl-logo.svg');
 }
 
+a.mapboxgl-ctrl-logo.mapboxgl-compact {
+    width: 21px;
+    height: 21px;
+    background-image: svg-load('svg/mapboxgl-ctrl-logo-compact.svg');
```
jfirebaugh: Rather than adding another SVG, could we hide the wordmark portion of the existing logo SVG by toggling a CSS class?

src/ui/control/logo_control.js
```diff
@@ -35,6 +36,10 @@ class LogoControl {
 
         this._map.on('sourcedata', this._updateLogo);
         this._updateLogo();
+
+        this._map.on('resize', this._updateCompact);
```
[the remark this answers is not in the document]
andrewharvey: good catch, fixed.

src/css/mapbox-gl.css
andrewharvey: Nice idea but since the logo is loaded as a background-image I don't think it's possible to control that through standard CSS, I think it would need to be an embedded SVG for that to work :thinking:. Given it doesn't add much weight to the file size, I'd be tempted to leave it as is

Issue #5853, opened 2017-12-13
Add some missing math expressions


anandthakker:
- abs
- round
- floor
- ceiling

Scarysize:
Would be a `random` expression something which fits into this issue?

jfirebaugh:
I don't think we should allow expressions to be non-deterministic. That would produce unpredictable results across tiles and complicate static rendering and caching.

anandthakker:
> I don't think we should allow expressions to be non-deterministic. That would produce unpredictable results across tiles and complicate static rendering and caching.

Hm, yeah, I agree that the static rendering / caching issues are problematic, but would unpredictable rendering across tiles be problematic? 

An (admittedly weird) alternative would be to provide (a) a `["get-random-integer", index: number]` that (in spirit) just looks up an entry in a (static) random number table, and (b) some sort of hashing expression that users could use to convert `["id"]` or `["get", "some-property"]` into an index to use with (a).

jfirebaugh:
> would unpredictable rendering across tiles be problematic? 

If `random` was used in any fill or line property, a likely expectation would be that it produces the same value for an entire logical feature, when in fact it will not match up when that feature is split across tiles.

Scarysize:
Make sense! Thanks for the explanation :)

1ec5:
In addition to the operators mentioned above, the following expression operators would improve [compatibility with NSExpression](https://developer.apple.com/documentation/foundation/nsexpression/1413747-init#discussion), which would benefit the iOS and macOS SDKs:

* `avg` returns the arithmetic mean (average) value in an array.
* `median` returns [the median value](https://en.wikipedia.org/wiki/Mode_%28statistics%29#Comparison_of_mean,_median_and_mode) in an array.
* `mode` returns the most frequent value in an array.
* `stddev` returns the standard deviation in an array.
* `exp` returns _e_ raised to a number.

mapbox/mapbox-gl-native#10726 [synthesizes `avg`, `exp`, `abs`, `trunc`, `floor`, and `ceiling`](https://github.com/mapbox/mapbox-gl-native/blob/1ae15d5d9115f13f711a6d6668346c20069c6d48/platform/darwin/docs/guides/Predicates%20and%20Expressions.md#predefined-functions) based on existing operators, but the others require native support.

kekscom:
A practical use case would be variance to 'fill-extrusion-height' in order to avoid z-fighting for colored roof faces.
[...]

PR #6496, opened 2018-04-11
Add abs, round, floor, ceil operators

anandthakker:
Refs #5853 

TODO: 
 - [x] Round away from 0 instead of towards +Infinity for halfway values. 
 - [x] Add doc entries and SDK support entries to v8.json

### DIFF
--- a/src/style-spec/expression/definitions/index.js
+++ b/src/style-spec/expression/definitions/index.js
@@ -317,7 +317,13 @@ CompoundExpression.register(expressions, {
     'round': [
         NumberType,
         [NumberType],
-        (ctx, args) => Math.round(args[0].evaluate(ctx))
+        (ctx, args) => {
+            const v = args[0].evaluate(ctx);
+            // Javascript's Math.round() rounds towards +Infinity for halfway
+            // values, even when they're negative. It's more common to round
+            // away from 0 (e.g., this is what python and C++ do)
+            return v < 0 ? -Math.round(-v) : Math.round(v);
+        }
     ],
     'floor': [
         NumberType,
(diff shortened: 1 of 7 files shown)


anandthakker:
`"ceil"` matches Javascript's `Math.ceil` -- should we use `"ceiling"` instead?

anandthakker:
cc @samanpwbb

src/style-spec/expression/definitions/index.js
```diff
@@ -309,6 +309,32 @@ CompoundExpression.register(expressions, {
         varargs(NumberType),
         (ctx, args) => Math.max(...args.map(arg => arg.evaluate(ctx)))
     ],
+    'abs': [
+        NumberType,
+        [NumberType],
+        (ctx, args) => Math.abs(args[0].evaluate(ctx))
```
jfirebaugh: Prefer argument destructuring as in the other definitions, i.e. `(ctx, [n]) => ...`.

src/style-spec/reference/v8.json
```diff
@@ -2490,6 +2490,42 @@
           }
         }
       },
+      "round": {
+        "doc": "Rounds the input to the nearest integer. Halflway values are rounded away from zero. E.g., `[\"round\", -1.5]` evaluates to -2.",
```
jfirebaugh: Halflway → Halfway

1ec5: Nit: It’s somewhat unusual to begin a sentence with “E.g.” “For example” is more easily understood.

src/style-spec/reference/v8.json
```diff
@@ -2490,6 +2490,42 @@
...
+        "group": "Math",
+        "sdk-support": {
+          "basic functionality": {
+            "js": "0.45.0"
+          }
+        }
+      },
+      "abs": {
+        "doc": "Returns the absolute value of the input.",
```
tobrun: when implementing I noticed abs on `-1.1` returns `1`, could we include in the documentation that abs is an integer type function/not floating point? (from a java background this was a bit confusing for me).

anandthakker:
Oh shoot, that’s a bug! Will fix

Issue #6184, opened 2018-02-16
queryRenderedSymbols does not sort features


mollymerp:
**mapbox-gl-js version**:

### Steps to Trigger Behavior

 1. load a symbol layer with collision detection disabled + overlapping symbols
 2. use `on('click', layer, ...` to invoke `queryRenderedFeatures` on the symbol layer
 3. 

### Expected Behavior

`e.features[0]` is the topmost rendered featured (and subsequent features are sorted in z order) per
https://github.com/mapbox/mapbox-gl-js/blob/38c7250da7b5153c531d0b5b8ff4a2f07f9ab863/src/ui/map.js#L747-L748

### Actual Behavior

`e.features[0]` is not always the topmost feature

![motivate](https://user-images.githubusercontent.com/2425307/36289640-da432168-1275-11e8-9208-c6c958d581c0.gif)

looks like we sort non-symbol features, but don't sort symbol features:

https://github.com/mapbox/mapbox-gl-js/blob/38c7250da7b5153c531d0b5b8ff4a2f07f9ab863/src/data/feature_index.js#L118-L126

ChrisLoer:
@ansis Do you know the origins of that documentation? I'm not sure what it means by "topmost rendered feature" -- it looks like the sorting is based on the order in which items are inserted into the `FeatureIndex` by `*Bucket.populate`... which is closely related to the draw/z-order but not strictly the same especially across tiles?

But the recent divergence of symbol behavior from non-symbol behavior is because of the sorting here:

https://github.com/mapbox/mapbox-gl-js/blob/859d4a93d639359c64edeb345cfbb2922a9d97e2/src/data/bucket/symbol_bucket.js#L635

I think we'll need to do something like store the sort indices generated there and use them for sorting the `matchingSymbols` results in `queryRenderedFeatures`

ansis:
@ChrisLoer yep, you're right about it being similar but not always right. The tile boundary bug might have existed when it was added, or we might have changed how querying works at tile boundaries (do we index points outside of tile edges or do we do padded queries across tile boundaries).

Your suggested fix sounds right

ChrisLoer:
I think I can build a good fix for this on top of PR #6497...

ChrisLoer:
Fixed in #6497.

PR #6497, opened 2018-04-11
Global symbol query

ChrisLoer:
This is an alternative to PR #6461 that goes further and also fixes issue #5475 and #6298 (switches to global instead of per-tile symbol querying, which fixes edge cases around tile boundaries). It also addresses the original hover flicker from issues #5887 and #5506.

I split it out into a separate PR because it's even more invasive and we might want to choose the smaller fix. However, at this point I'm in favor of going with this version because it has some nice simplifications:

 - No round-tripping of viewport query coordinates through tile space
 - No more need to merge duplicate results from the same symbol showing up in multiple tiles
 - All querying-related data can now be indexed with a bucket instance id and a feature index.
 - CrossTileSymbolIndex manages lifetime of data needed for querying symbols, Tile is only responsible for "latest" data
 - CollisionBoxArray no longer involved in querying at all

I've just realized a subtle problem with this PR -- the order of symbol results can now depend on the (arbitrary) instance ID of the bucket they belong to. I'll need to figure out a fix for that. Do we have to go as far as fully reproducing the previous order based on tileID? (We should be able to do that, since we can work back from bucket to tile).

 - [x] post benchmark scores

/cc @ansis @malwoodsantoro @ryanbaumann @lilykaiser @jfirebaugh


ChrisLoer:
I just pushed a change to make the symbol results sort first by tileID, the same way they did before. There is still a small difference:

- Previously, within a tile/layer, we sorted symbol results by their index into the `CollisionBoxArray` (basically the order in which they got created during `performSymbolLayout`)
- Now, since we're not using the `CollisionBoxArray`, we sort them by their feature index in the original vector tile.

As far as I know, these should be pretty much equivalent since the layout order is derived relatively directly from the order of features in the tile, but I'm not sure I could guarantee it.

ChrisLoer:
Benchmarks look fine except for QueryPoint and QueryBox... which look like they need work!

screenshot 2018-04-11 16 47 51

src/symbol/collision_index.js
```diff
@@ -317,41 +278,41 @@ class CollisionIndex {
...
-            const y2 = blocking.y2 * tileToViewport + projectedPoint.point.y;
             const bbox = [
-                new Point(x1, y1),
-                new Point(x2, y1),
-                new Point(x2, y2),
-                new Point(x1, y2)
+                new Point(feature.x1, feature.y1),
+                new Point(feature.x2, feature.y1),
+                new Point(feature.x2, feature.y2),
+                new Point(feature.x1, feature.y2)
```
ansis: Does the bbox polygon need five points (with the last point being the same as the first) in order for the intersection test to work properly?

src/source/tile.js
```diff
@@ -148,13 +146,20 @@ class Tile {
...
+            this.latestFeatureIndex = data.featureIndex;
+            if (data.rawTileData) {
+                // Only vector tiles have rawTileData, and they won't update it for
+                // 'reloadTile'
+                this.latestRawTileData = data.rawTileData;
+                this.latestFeatureIndex.rawTileData = data.rawTileData;
+            } else if (this.latestRawTileData) {
+                // If rawTileData hasn't updated, hold onto a pointer to the last
+                // one we received
+                this.latestFeatureIndex.rawTileData = this.latestRawTileData;
```
ansis: In what cases does a the new `data` have a `featureIndex` but not a `rawTileData`?

src/data/bucket/symbol_bucket.js
```diff
@@ -593,7 +598,8 @@ class SymbolBucket implements Bucket {
             // An icon can only have one box now, so this indexing is a bit vestigial...
             const box: CollisionBox = (collisionBoxArray.get(k): any);
             if (box.radius === 0) {
-                collisionArrays.iconBox = { x1: box.x1, y1: box.y1, x2: box.x2, y2: box.y2, anchorPointX: box.anchorPointX, anchorPointY: box.anchorPointY };
+                collisionArrays.iconBox = { x1: box.x1, y1: box.y1, x2: box.x2, y2: box.y2, anchorPointX: box.anchorPointX, anchorPointY: box.anchorPointY, featureIndex: box.featureIndex };
```
ansis: Is the `featureIndex` stored on `iconBox` used anywhere?

src/symbol/placement.js
```diff
@@ -81,7 +81,10 @@ class Placement {
 
     placeLayerTile(styleLayer: StyleLayer, tile: Tile, showCollisionBoxes: boolean, seenCrossTileIDs: { [string | number]: boolean }) {
         const symbolBucket = ((tile.getBucket(styleLayer): any): SymbolBucket);
-        if (!symbolBucket) return;
+        if (!symbolBucket || !tile.latestFeatureIndex)
```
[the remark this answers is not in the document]
ChrisLoer: None that I know of, that was for flow.

src/data/bucket/symbol_bucket.js
ChrisLoer: Oops! That's vestigial from an earlier implementation.

src/source/tile.js
ChrisLoer: `loadTile` in `VectorTileWorkerSource` returns a `rawTileData`, but `reloadTile` doesn't (we use a new `FeatureIndex` but with the previous data since it hasn't changed).

src/symbol/cross_tile_symbol_index.js
```diff
@@ -230,11 +255,23 @@ class CrossTileSymbolIndex {
...
+            // the matching FeatureIndex/CollisionBoxArray/data for querying purposes
+            this.lastUsedBuckets[symbolBucket.bucketInstanceId] = true;
+            this.retainedBuckets[symbolBucket.bucketInstanceId] = new RetainedBucket(
+                symbolBucket.bucketInstanceId,
+                bucketFeatureIndex,
+                symbolBucket.sourceLayerIndex,
+                symbolBucket.index,
+                tile.tileID
+            );
+
```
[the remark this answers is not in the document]
ChrisLoer: This logic is more about holding onto data during incremental placement than it is about holding onto data between frames. `lastUsedBuckets` is only cleared during the `prune` operation that happens on committing a placement. That allows us to hold onto `RetainedBuckets` that weren't used in the previous/current frame, but will still be part of the next placement that gets committed. Does that make sense?

`RetainedBuckets` is probably a misleading name, because it makes it sound like we're holding on to the buckets themselves (and thus makes it sound like something related to rendering). What we're actually holding onto is all the bucket-specific data necessary for symbol querying (basically, the raw vector tile data plus some metadata). Maybe `RetainedBucketQueryData`?

A further refactoring that might make sense is to stop holding on to the `FeatureIndex` entirely here. We're using it for shared code that loads features from the underlying vector data, but we don't need to hold onto all the data it has in its non-symbol `GridIndex`, since we always query the latest version for non-symbol features.

src/symbol/collision_index.js
ChrisLoer: I don't think so? I actually haven't looked closely at how `intersection_tests.js` implements this, but it's the same behavior as before, right?

src/symbol/placement.js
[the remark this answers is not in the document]
ChrisLoer: Ah, but it's also vestigial -- in an earlier implementation, I had the collision box data in the feature index. I'll get rid of that.

src/symbol/cross_tile_symbol_index.js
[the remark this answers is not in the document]
ChrisLoer: 🤔 Not sure if it would have a noticeable performance impact, but the current implementation of `lastUsedBuckets` holds on to more than it needs to, since it will hold onto any bucket that gets loaded during any frame of an incremental placement, even if that bucket didn't actually end up used in the placement (because of where it was in the layer order). I'll look at changing that behavior so the intent is clearer.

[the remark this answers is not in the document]
ansis: Thanks for explaining, I think I had confused `prune` with `removeStaleBuckets`. This makes sense.

Renaming to `RetainedBucketQueryData` sounds good to me.

What is the advantage of loading the exact feature used for placement instead of loading features with matching `crossTileIDs`?

ansis:
I have one high-level question about this approach: The results are wrong (and already were wrong before this pr) if the query happens after the map moves but before placement runs. This is generally ok because this window is very small. Do you think it will be important to fix this at some point? And if we do fix it, would a global approach still be useful or would it make sense to switch back to tiled indexing at that point?

(I don't think we have to think too much about this right now)

src/symbol/grid_index.js
```diff
@@ -91,12 +91,32 @@ class GridIndex {
...
+            if (hitTest) {
+                return true;
+            }
+            for (let boxUid = 0; boxUid < this.boxKeys.length; boxUid++) {
+                result.push({
+                    key: this.boxKeys[boxUid],
+                    x1: this.bboxes[boxUid * 4],
+                    y1: this.bboxes[boxUid * 4 + 1],
+                    x2: this.bboxes[boxUid * 4 + 2],
+                    y2: this.bboxes[boxUid * 4 + 3]
```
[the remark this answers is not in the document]
ansis: I was thinking `query(geometry: Array<Point>)` but a filter function could also work. How hot is this part of querying? Does creating extra objects add noticeable overhead? The current approach seems fine to me and I don't think it necessarily needs to change.

src/symbol/cross_tile_symbol_index.js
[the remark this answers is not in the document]
ChrisLoer: Hrm, that's a good question, I haven't thought through what it would be like to implement that. So instead of storing bucket/feature indices in the `CollisionIndex`, we'd just store a `crossTileID` and maintain a reverse-lookup index to go from a `crossTileID` to the latest matching bucket/`symbolInstance`?

Initial reactions:
 - There can be multiple symbol instances per feature, and our querying logic is feature-based, so we'd have to be careful to handle duplicates/merging correctly there.
 - We'd have to figure out something reasonable to do with `crossTileID`s that were in the latest placement, but point to a tile removed from the pyramid since the placement happened.
 - Using the latest data instead of the data that was included in the latest placement opens up some room for some subtle mismatches that could in some cases be hard to debug (e.g. your query shows a symbol as being within your query geometry, but the returned `GeoJSONFeature` appears to be outside your query geometry).

I think the honest answer is "I didn't think of doing it that way". What would be the major advantage of switching to that approach? Do you think it might be a simplicity win, or are you more thinking about being able to discard data earlier?

In most cases, I don't think the current implementation leads to much extra data being retained, but the worst case is probably when there's something like a `setData`-style GeoJSON animation causing tiles with symbols to update really frequently (say, every 20ms). In that case, we'd be holding on to ~15 copies of the tile data over the course of a 300ms placement.... which sounds less-than-optimal, although I still don't think it'd be that bad (not much data, the data would have to wait for GC either way, etc.).

src/symbol/grid_index.js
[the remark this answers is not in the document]
ChrisLoer: For rendering, the 'hitTest' part of the logic is hot. For querying... I need to do more profiling anyway to figure out why the "Query" benchmarks have slowed down so much. I like this idea, but I think it seems fine to defer to later for now.

src/symbol/cross_tile_symbol_index.js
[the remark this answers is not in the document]
ansis: I was thinking about it from a simplicity point of view (not having to deal with retaining logic), but I think those are some good points. Especially the third one. I think your current approach is 👍

ChrisLoer:
> The results are wrong (and already were wrong before this pr) if the query happens after the map moves but before placement runs. This is generally ok because this window is very small. Do you think it will be important to fix this at some point?

Yeah, good point. I think we should ticket it and fix it, although it should be mostly separate from this PR. I can imagine it cropping up in something like an interactive article that uses animations to give a tour of the map and also implements hover functionality in the map -- it would be easy to get kind of bizarre hover behavior.

> And if we do fix it, would a global approach still be useful or would it make sense to switch back to tiled indexing at that point?

My first thought on how to fix it would be to convert the query geometry based on the current transform to match the transform used in the placement. Wouldn't that solve most of the problem? The current screen coordinates might map to something that was entirely outside the scope of the last placement, but in that case most of the symbols wouldn't be placed on screen either (there are exceptions like `text-allow-overlap: true`, but it's already the case that those can show up on screen before they show up in results).

ansis:
> My first thought on how to fix it would be to convert the query geometry based on the current transform to match the transform used in the placement. Wouldn't that solve most of the problem?

This would be closer but would also not handle labels that change size or position (relative to the map plane) when zooming, rotating or when panning in perspective view. I think that completely fixing it would involve not using Placement's grid index for querying and instead using an index we can create per-tile in the work or something. We'd need to be able to apply the zoom and perspective transforms to the labels at query time. But making that fast enough would probably be hard and maybe not worth the effort right now.

ChrisLoer:
Yeah, if we wanted symbol querying to always perfectly match what was on screen, I think we'd need a totally different approach... re-calculating everything at query time could work but also be expensive. I guess the "dream" is still that we somehow get the shaders to generate query information for us on every frame.

But I'm less concerned about general problems of the form "what we're rendering at this lat/lon has changed since the last placement" than I am concerned about the specific problem where the effective lat/lon of the query changes.

mb12:
@ansis If the caller re-runs the query upon receiving event REGION_DID_CHANGE/REGION_DID_CHANGE_ANIMATED, would that fix the problem? If the map is moving/animating when the query is made, it is okay for the results to be inconsistent, but they should be consistent when the client receives the change done events.

src/symbol/cross_tile_symbol_index.js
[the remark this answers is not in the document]
ChrisLoer: I got rid of the pruning entirely by moving `RetainedQueryData` into `Placement` -- it's a much better fit since the lifetime of the data needs to be the same as the lifetime of the `Placement`. This also avoids retaining data for buckets that never get included in a placement, which means the `setData`-animation worst case should go from ~15x tile data storage to ~2x storage.

src/symbol/collision_index.js
ansis: Yep, it's the same behaviour as before.

https://github.com/mapbox/mapbox-gl-js/blob/13ea16df5eabc4c8867d69e81da180cdcd6cc6b2/src/util/intersection_tests.js#L112-L124

`lineIntersectsLine`, which is used by `polygonIntersectsPolygon`, *seems* to expect a repeated vertex when used this way (the last segment is never tested). BUT I think it's still (accidentally?) correct, because it's impossible for this missing edge to be crossed without either crossing a different edge or having a point within the polygon. So, I'm not sure if this a bug or not

src/source/tile.js
ansis: Thanks!

ansis:
@ChrisLoer Moving the bucket retaining to placement (dd20158) looks great. Thanks for clearing up all these questions for me. This looks ready to me if the performance regression can be fixed

@mb12 thanks for your input. If I'm understanding you correctly, that would involve introducing new events? that would probably be nice to avoid.

mb12:
@ansis  These events are already available on both iOS and Android. We should add them to js as well. I copy pasted those names from the Android version (that's why they were all caps).

src/data/feature_index.js
```diff
@@ -125,79 +114,105 @@ class FeatureIndex {
...
-        pixelsToTileUnits: number,
-        posMatrix: Float32Array,
-        transform: Transform
-    ) {
-        let previousIndex;
-        for (let k = 0; k < matching.length; k++) {
-            const index = matching[k];
+        intersectionTest?: (feature: VectorTileFeature, styleLayer: StyleLayer) => boolean) {
 
-            // don't check the same feature more than once
```
ChrisLoer: 🤦‍♂️ This turned out to be the cause of the performance regression -- we definitely still need to filter out duplicate entries. I misread `FeatureIndex#insert`, which can actually make _very many_ entries in the grid for a single feature. I spent half a day messing around trying to reliably profile QueryBox before I finally realized this...

ChrisLoer:
Fixed the perf regression (take these screenshots with a grain of salt -- I've had a lot of trouble getting QueryBox especially to reproduce results reliably... but the >10x slowdown is definitely fixed):

screenshot 2018-04-12 15 23 04
screenshot 2018-04-12 15 20 33

src/symbol/collision_index.js
ChrisLoer: 🙈 Do you have any ideas for tests or assertions we could put in to make it clear what types of geometry are supported? I haven't wrapped my head around it and for now my assumption is it's OK.

ChrisLoer:
I couldn't resist pushing a fix for issue #6184, since it's closely related and I think we should take the opportunity to fix things in `queryRenderedSymbols` when we've got all of its complexity loaded up in our heads. 🙂 

@mollymerp, is there any chance you have the reproduction case hanging around from #6184? I bootstrapped the test for getting the rotation sort order right, but I keep getting tripped up when I stare at all those coordinates and try to think what order they're supposed to be in.

src/data/feature_index.js
```diff
@@ -125,79 +114,112 @@ class FeatureIndex {
...
+                    return styleLayer.queryIntersectsFeature(queryGeometry, feature, geometry, this.z, args.transform, pixelsToTileUnits, args.posMatrix);
+                }
+            );
+        }
 
         return result;
     }
 
-    filterMatching(
+    generateGeoJSONFeature(
```
anandthakker: Naming nit: feels to me like the most important thing this method is doing is still deciding whether or not to include the feature, moreso than generating the geojson result if the answer is 'yes'.  Bad suggested rename: `maybeAddFeatureToResult()`

src/data/bucket/symbol_bucket.js
```diff
@@ -656,8 +657,11 @@ class SymbolBucket implements Bucket {
         this.text.indexArray.clear();
         this.icon.indexArray.clear();
 
+        this.featureSortOrder = [];
+
         for (const i of symbolInstanceIndexes) {
             const symbolInstance = this.symbolInstances[i];
+            this.featureSortOrder.push(symbolInstance.featureIndex);
```
ansis: Would line symbol instances that share the same feature have the same `featureIndex`? Would all features be sorted based on the y of the first instance?

src/source/query_features.js
```diff
@@ -56,9 +57,31 @@ export function queryRenderedSymbols(styleLayers: {[string]: StyleLayer},
...
+                // Match topDownFeatureComparator from FeatureIndex, but using
+                // most recent sorting of features from bucket.sortFeatures
+                const featureSortOrder = queryData.featureSortOrder;
+                if (featureSortOrder) {
+                    // queryRenderedSymbols documentation says we'll return features in
+                    // "top-to-bottom" rendering order (aka last-to-first).
+                    // Actually there can be multiple symbol instances per feature, so
+                    // we sort each feature based on the first matching symbol instance.
+                    const sortedA = featureSortOrder.indexOf(a.featureIndex);
+                    const sortedB = featureSortOrder.indexOf(b.featureIndex);
```
ansis: Would it be better to use an object for the lookup instead of an array with a linear indexOf here? The feature numbers could be low enough that this doesn't matter at all though

src/symbol/placement.js
```diff
@@ -393,6 +394,9 @@ export class Placement {
         }
 
         bucket.sortFeatures(this.transform.angle);
+        if (bucket.featureSortOrder) {
+            this.retainedQueryData[bucket.bucketInstanceId].featureSortOrder = bucket.featureSortOrder.slice();
+        }
```
ansis: Is the copy of the array necessary?
Would passing `RetainedQueryData` to `bucket.sortFeatures(...)` avoid the need to store the array on the bucket at all?

src/data/feature_index.js
ChrisLoer: 🤔You're right, when I first named this I thought the two pathways would end up sharing a smaller chunk of shared code. Maybe I should just go back to `filterMatching`? Or maybe `loadMatchingFeatures`?

src/data/bucket/symbol_bucket.js
ChrisLoer: Yup -- and this could be a little weird in "hover effect" type interactions, where you're hovering over a particular symbol instance that looks topmost, but you're getting the feature behind it first. But since we return unique features, not symbol instances, it would also be kind of weird to sort based on symbol instance: the same query results would get sorted differently just based on how you generated them.

src/source/query_features.js
ChrisLoer: I thought a little about this and then decided it would be premature optimization to do that without this showing up as a hot spot. Worst case here is when we sort results from a really dense layer of potentially overlapping results, but even with a thousand features I don't think it gets that bad.

src/symbol/placement.js
ChrisLoer: Yeah, having the sort array always live with `RetainedQueryData` is a good idea, and we can avoid the copy.

There is a subtle oddity here -- the sort order will be based on the last time we update opacities, instead of being based on the last time placement happened. It's a little weird that it's not all synced to the same time point, but at least right now, I can't think of why it would be a problem to use a somewhat newer sort order for the same data.

ChrisLoer: Changed.

src/data/feature_index.js
ChrisLoer: Went with `loadMatchingFeature`

src/symbol/placement.js
ChrisLoer: Actually, I realized this change was wrong in some last minute manual testing! Moving to `RetainedQueryData` is actually awkward because `sortFeatures` doesn't necessarily run on every placement (it will early-exit if the bearing hasn't changed). We'd have to make sure to copy `featureSortOrder` from previous placements, and it's actually much simpler to just use the last `featureSortOrder` from the bucket itself. We still get rid of the array copy.

ChrisLoer:
@mollymerp I ended up using the new hover-bus behavior in `debug/debug.html` for manually sanity-checking the results, so no need to worry about finding the original repro case from #6184. My test was just to make two of the icons overlap, check that `queryRenderedSymbols` put the bottom-y (top-most rendered) first in the list, rotate the map 180 degrees, and make sure the sort order is flipped.

PR #6499, opened 2018-04-11
Merge release-0.45 back to master

anandthakker:
We originally made this release branch because we'd [planned to hold `line-gradient` until the next release](https://github.com/mapbox/mapbox-gl-js/pull/6303#issuecomment-377641568).  Since we decided to go ahead and include line-gradient in 0.45, I no longer see any need to exclude changes on `master`.


anandthakker:
> We're going to keep the branch around for beta release purposes, right?

I was thinking we'd delete this branch, release from master, and only re-create `release-0.45` on demand.  @jfirebaugh do you think we should do something different?

jfirebaugh:
I think there will be changes after the beta release that we want to go into master but not 0.45 final.

anandthakker:
FYI @mollymerp

Issue #6433, opened 2018-04-01
test/build/style-spec.test.js is not run in CI


jfirebaugh:
It was added in #5534 but AFAICT has never been executed with any of our automated test commands.

PR #6505, opened 2018-04-12
Rework spec function/expression taxonomy 

lbud:
This PR reworks the taxonomy to describe how a property can use functions/expressions, as described in https://github.com/mapbox/mapbox-gl-js/issues/6389#issuecomment-376187545 and https://github.com/mapbox/mapbox-gl-js/issues/6389#issuecomment-379861063:

``` ts
type ExpressionType = 'data-driven' | 'cross-faded' | 'cross-faded-data-driven' | 'color-ramp' | 'data-constant' | 'constant';

type ExpressionParameters = Array<'zoom' | 'feature' | 'heatmap-density' | 'line-progress'>;

type ExpressionSpecification = {
    interpolated: boolean,
    parameters: ExpressionParameters
}

StylePropertySpecification = {
    ...
    'property-type': ExpressionType,
    'expression': ExpressionSpecification
}
```

so now any property all properties include a `property-type` key, and for all property-types except `constant` also an `expression` property, i.e.
```
    "type": "number",
    "property-type": "data-driven",
    "expression": {
        "interpolated": true,
        "parameters": [
            "zoom",
            "feature"
        ]
    }
```

Benchmarks: http://bl.ocks.org/lbud/raw/2ac3b8869074c112802166978cea8345/ (I couldn't reproduce the Layer discrepancies on subsequent isolated bench runs).

Closes https://github.com/mapbox/mapbox-gl-js/issues/6389
Supersedes / closes https://github.com/mapbox/mapbox-gl-js/issues/4194
Refs https://github.com/mapbox/mapbox-gl-js/pull/6430
Refs https://github.com/mapbox/mapbox-gl-js/pull/6303

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/build/generate-style-code.js
+++ b/build/generate-style-code.js
@@ -43,8 +43,10 @@ global.propertyType = function (property) {
         case 'color-ramp':
             return `ColorRampProperty`;
         case 'data-constant':
-        default:
+        case 'constant':
             return `DataConstantProperty<${flowType(property)}>`;
+        default:
+            throw new Error(`unknown property-type "${property['property-type']}" for ${property.name}`);
     }
 };
 
@@ -97,8 +99,10 @@ global.propertyValue = function (property, type) {
         case 'color-ramp':
             return `new ColorRampProperty(styleSpec["${type}_${property.layerType}"]["${property.name}"])`;
         case 'data-constant':
-        default:
+        case 'constant':
             return `new DataConstantProperty(styleSpec["${type}_${property.layerType}"]["${property.name}"])`;
+        default:
+            throw new Error(`unknown property-type "${property['property-type']}" for ${property.name}`);
     }
 };
 
--- a/build/generate-style-code.js
+++ b/build/generate-style-code.js
@@ -12,10 +12,6 @@ global.camelize = function (str) {
     });
 };
 
-global.isDataDriven = function (property) {
-    return property['property-function'] === true;
-};
-
 global.flowType = function (property) {
     switch (property.type) {
         case 'boolean':
@@ -39,14 +35,16 @@ global.flowType = function (property) {
 };
 
 global.propertyType = function (property) {
-    if (isDataDriven(property)) {
-        return `DataDrivenProperty<${flowType(property)}>`;
-    } else if (/-pattern$/.test(property.name) || property.name === 'line-dasharray') {
-        return `CrossFadedProperty<${flowType(property)}>`;
-    } else if (property.name === 'heatmap-color' || property.name === 'line-gradient') {
-        return `ColorRampProperty`;
-    } else {
-        return `DataConstantProperty<${flowType(property)}>`;
+    switch (property.expression['property-type']) {
+        case 'data-driven':
+            return `DataDrivenProperty<${flowType(property)}>`;
+        case 'cross-faded':
+            return `CrossFadedProperty<${flowType(property)}>`;
+        case 'color-ramp':
+            return `ColorRampProperty`;
+        case 'data-constant':
+        default:
+            return `DataConstantProperty<${flowType(property)}>`;
     }
 };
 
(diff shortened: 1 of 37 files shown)


src/style-spec/style-spec.js
```diff
@@ -1,51 +1,46 @@
 // @flow
 
+type ExpressionType = 'data-driven' | 'cross-faded' | 'cross-faded-data-driven' | 'color-ramp' | 'data-constant';
+type ExpressionParameters = Array<'zoom' | 'feature' | 'heatmap-density' | 'line-progress'>;
+
+type ExpressionSpecification = {
+    type: ExpressionType,
```
anandthakker: Looks like field is currently called `property-type`, not `type`

anandthakker: It feels slightly weird to me to have `property-type` nested under `expression`.  Would it make sense to move it up to the layout/paint property specification, and also add `constant` as one of the options for properties that can't take any expression?   E.g.:

```js
"visibility": {
  "property-type": "constant",
  // no "expression" field
  "type": "enum",
   /* ... */
}
```

```js
"line-cap": {
  "property-type": "data-constant",
  "expression": {
    "interpolated": false,
    "parameters": ["zoom"]
   },
  "type": "enum",
   /* ... */
}
```

Then we could always use `propertySpec["property-type"]` and never have to rely on the presence/absence of the "expression" field

src/style-spec/util/properties.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+
+export function isPropertyFunction(spec: Object): boolean {
+    const dataDrivenTypes = new Set(['data-driven', 'cross-faded-data-driven']);
```
anandthakker: I don't think we can use Set / Map, since they'd require polyfills for IE 😢

anandthakker: Ah, this would have showed up in the style-spec build test, but it's not getting run on CI 😬 https://github.com/mapbox/mapbox-gl-js/issues/6433

src/style-spec/util/properties.js
```diff
@@ -0,0 +1,15 @@
...
+
+    return spec.expression && dataDrivenTypes.has(spec.expression['property-type']);
+}
+
+export function isZoomFunction(spec: Object): boolean {
+    return spec.expression && spec.expression.parameters.indexOf('zoom') > -1;
+}
```
anandthakker: Rename to `isPropertyExpression` and `isZoomExpression`?

src/style-spec/style-spec.js
lbud: Sounds good to me @anandthakker!

src/style-spec/function/convert.js
```diff
@@ -236,8 +236,8 @@ function appendStopPair(curve, input, output, isStep) {
 function getFunctionType(parameters, propertySpec) {
     if (parameters.type) {
         return parameters.type;
-    } else if (propertySpec.function) {
-        return propertySpec.function === 'interpolated' ? 'exponential' : 'interval';
+    } else if (propertySpec.expression) {
```
jfirebaugh: Should this be asserting `propertySpec.expression`? We shouldn't be trying to create a function for something that doesn't support expressions.

build/generate-style-code.js
```diff
@@ -39,14 +35,16 @@ global.flowType = function (property) {
...
+    switch (property['property-type']) {
+        case 'data-driven':
+            return `DataDrivenProperty<${flowType(property)}>`;
+        case 'cross-faded':
+            return `CrossFadedProperty<${flowType(property)}>`;
+        case 'color-ramp':
+            return `ColorRampProperty`;
+        case 'data-constant':
+        default:
+            return `DataConstantProperty<${flowType(property)}>`;
```
anandthakker: Just for completeness, should we also list `case 'constant':` here?  (Alternatively, since `DataConstantProperty<T>` isn't really intended to represent a totally constant property, it could be `case 'constant': throw new Error("Constant properties unimplemented");`)

build/generate-style-code.js
```diff
@@ -91,14 +89,16 @@ global.defaultValue = function (property) {
...
+    switch (property['property-type']) {
+        case 'data-driven':
+            return `new DataDrivenProperty(styleSpec["${type}_${property.layerType}"]["${property.name}"])`;
+        case 'cross-faded':
+            return `new CrossFadedProperty(styleSpec["${type}_${property.layerType}"]["${property.name}"])`;
+        case 'color-ramp':
+            return `new ColorRampProperty(styleSpec["${type}_${property.layerType}"]["${property.name}"])`;
+        case 'data-constant':
+        default:
+            return `new DataConstantProperty(styleSpec["${type}_${property.layerType}"]["${property.name}"])`;
```
anandthakker: Same as above

src/style-spec/function/convert.js
lbud: 👍 this was a bad workaround for a flow error — will fix.

Issue #6448, opened 2018-04-03
Snapshot for Mapbox GL JS


samsammurphy:
## Motivation

Create a png/jpeg image of a map and all its visible layers using GL JS, similar too:

- [ios snapshot](https://www.mapbox.com/ios-sdk/examples/map-snapshotter/)
- [react snapshot](https://github.com/react-community/react-native-maps#take-snapshot-of-map)
- [static maps](https://www.mapbox.com/help/define-static-api/)

My aim is to allow users to find awesome content on a map and readily share it using social media posts that we help generate for them, with an image of what they were looking at.

## Design Alternatives

The current [working solution](https://codepen.io/samsammurphy/pen/VXdOPv) is a bit clunky and hard to find online. In summary, set `preserveDrawingBuffer = true` when initializing the map, then use `map.getCanvas().toDataURL()` to get a base64 encoded image.

### Mock-Up

.. I love sugar.. I would suggest having a method for developers called `map.snapShot()`

### Concepts

Then I would make it easy to find by adding it to the [examples](https://www.mapbox.com/mapbox-gl-js/example/simple-map/).

### Implementation

`map.snapShot()` would effectively

```
set preserveDrawingBuffer to true (if not already)
get image using map.getCanvas().toDataURL()
set preserveDrawingBuffer back to false (if that's what it was before the call).

return the image
```

samsammurphy:
.. I love sugar.. I would suggest having a method called `map.snapShot()` which effectively 

1) enables `preserveDrawingBuffer` if not already true
2) runs `map.getCanvas().toDataURL()`
3) returns `preserveDrawingBuffer` to previous state.

jfirebaugh:
I'm not sure this is possible to encapsulate.

`preserveDrawingBuffer` can't be modified on the fly. It has to be set at the time the WebGL context is created, and that can have a negative effect on performance.

It's rumored that you can grab the the canvas data URL immediately after rendering, without needing `preserveDrawingBuffer`, but I haven't verified that, and I suspect it's not guaranteed by the spec.

samsammurphy:
My suggestion would then be to add an official [example](https://www.mapbox.com/mapbox-gl-js/example/simple-map/). Is it ok to suggest that here?

It might look something [like this](https://codepen.io/samsammurphy/pen/VXdOPv).

mollymerp:
[...]

PR #6513, opened 2018-04-14
use postcss to inline svg files into css

andrewharvey:
## Launch Checklist
 - [x] briefly describe the changes in this PR

1. closes #6493 by using postcss-cli to build `dist/mapbox-gl.css`.
2. Uses https://github.com/TrySound/postcss-inline-svg to keep svg assets as plain svg files inlined to css automatically. Currently working with base64 encoded and/or url encoded svg files in the `dist/mapbox-gl.css` makes editing svg assets or introducing new ones more complex than it needs to be.
3. https://github.com/egoist/rollup-plugin-postcss exists but it looks like it will automatically bundle the style with the JS distributeable and auto insert into the document at run time. I'm not sure if that's desirable so I've just used postcss-cli for now to build the `dist/mapbox-gl.css` to minimise change to end users.

 - n/a write tests for all new functionality

4. stylelint still runs

 - n/a document any changes to public APIs
 - [x] post benchmark scores

before:
40K dist/mapbox-gl.css
11K dist/mapbox-gl.css.gz

after:
31K dist/mapbox-gl.css
8.1K dist/mapbox-gl.css.gz

5. I think file sizes decreased due to svg's being url encoded rather than a mix of url and base64 encoded.

 - [x] manually test the debug page
6. tested all the icons on the debug page on Chrome, Firefox, Android Browser, IE11, iOS Safari everything looks good, just like existing.

7. This conflicts with #6472 and #6506 so some rebasing and merge conflict resolution will need to happen.
8. I've updated the `yarn run` targets in package.json as best I can, but I would appreciate some :eyes: on that part in particular.

/cc @anandthakker @davidtheclark

PR #6518, opened 2018-04-16
snapshot example (basic outline)

samsammurphy:
## Launch Checklist

An example of snapshot for GL JS (#6448)


samsammurphy:
**NB. change of tactic**: snapshot to canvas (previously to div)

Q1. I am not sure where the original map canvas is getting the logo from. Is there a standard href I should use? For now, I have included a base64 encoded image in the example code.

Q2. Why is the height of the map not 100% when width is < 300px? For example, the green and blue regions  below.
width_lt_300px

samsammurphy:
misclick on close and comment (reopening)

samsammurphy:
@andrewharvey :tada:  all checks passed. ready for inspection

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,102 @@
...
+
+var map = new mapboxgl.Map({
+    container: 'map',
+    style: 'mapbox://styles/mapbox/dark-v9',
+    center: {
+        lat: 37.4036,
+        lng: -122.0325
+    },
+    zoom: 10,
+    preserveDrawingBuffer: true// not guaranteed to work if false
```
andrewharvey: I would just add a comment like " required for the map's canvas to be exported to a PNG"

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,102 @@
...
+ctx.fillText(text, snapshot.width / 2, snapshot.height / 2);
+
+map.on('click', function() {
+    var png = map.getCanvas().toDataURL();
+    var copy = new Image();
+    copy.src = png;
+    copy.onload = function() {
+        ctx.drawImage(copy, 0, 0);
+        logo();
+        if (snapshot.width > 400) {
```
andrewharvey: The attribution should always be shown regardless of the image width, similar to [how the static API works](https://www.mapbox.com/api-documentation/#retrieve-a-static-map-from-a-style), shortening to just (c) OSM should be okay on very small images where that is all that will fit.

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,102 @@
...
+    ctx.fillRect(snapshot.width - width, snapshot.height - height, width, height);
+    ctx.font = fontsize + 'px Arial';
+    ctx.fillStyle = 'black';
+    ctx.textAlign = 'left';
+    ctx.textBaseline = 'bottom';
+    ctx.fillText(text, snapshot.width - width, snapshot.height - 3);
+}
+
+function logo() {
+    // Please attribute mapbox if using non-commercially
```
andrewharvey: I think it's better if we find the Mapbox logo from the DOM, if it exists then add it, if not, then don't. This means the snapshot will always reflect the interactive map and it avoids hardcoding a second copy of the logo.

docs/pages/example/snapshot.js
```diff
@@ -0,0 +1,10 @@
+/*---
+title: Snapshot
```
andrewharvey: In line with the rest of the examples, I think we should use something like "Snapshot an image of a map" as the title, "snapshot" on it's own is a bit unclear.

andrewharvey:
@samsammurphy Sorry I missed your questions...

> Q1. I am not sure where the original map canvas is getting the logo from. Is there a standard href I should use? For now, I have included a base64 encoded image in the example code.

It's HTML (svg) added on top, so it's not part of the map canvas.

andrewharvey:
Also this PR should [target the `mb-pages` branch instead of `master`](https://github.com/mapbox/mapbox-gl-js/blob/master/docs/README.md#committing-and-publishing-documentation) as it doesn't need any other features from master not yet released (means it will go live when merged, rather than at the next release). I don't think you can change the branch on a PR, but to avoid scattering the PR, let's hold off for now.

docs/pages/example/snapshot.html
samsammurphy: I have updated the comment to this suggestion.

docs/pages/example/snapshot.html
samsammurphy: a text attribution will now always be displayed on the map, the short version (c) OSM kicks in if the canvas width is < 300px

docs/pages/example/snapshot.js
samsammurphy: I would make the case that the description 'capture a snapshot image of a map' and the interactive example 'click on the map' make this clear. Shorter titles also make it easier and faster to scan through the example titles..

However, I understand that this is your project (styles, decisions, aesthetics, etc.).

screenshot from 2018-04-25 10-03-02

andrewharvey: I'm just working off the instructions at https://github.com/mapbox/mapbox-gl-js/tree/master/docs#writing-examples which say: "A short title for the example in sentence case as a verb phrase".

samsammurphy:
@andrewharvey 

1) Title has been updated 
2) Logo is now read from DOM. 

The approach in (2) works for Chrome. However it seems like [Firefox](https://github.com/samsammurphy/mapbox-gl-js/blob/0938863e0e67aa0a192b218c53d49fcb3316eef6/docs/pages/example/snapshot.html#L99) might require quite a bit of (experimental) [work](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Drawing_DOM_objects_into_a_canvas) or a [third party library](https://github.com/canvg/canvg) to draw SVGs on a canvas?

I tried the top two solutions on this [SO thread](https://stackoverflow.com/questions/3768565/drawing-an-svg-file-on-a-html5-canvas/19699830).

docs/pages/example/snapshot.html
samsammurphy: Now loading the Mapbox logo from the DOM

docs/pages/example/snapshot.js
samsammurphy: Title updated as suggested.

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,120 @@
...
+}
+
+function mapboxlogo() {
+
+    // logo image
+    var a = document.querySelector('a.mapboxgl-ctrl-logo');
+    var style = window.getComputedStyle(a, false);
+    var base64 = style.backgroundImage.slice(31, -2);
+    var b64Start = 'data:image/svg+xml;base64,';
+    var image64 = b64Start + base64;
```
mollymerp: we're not actually using base64 svg anymore since #6513, so I think you'll want to use `"data:image/svg+xml;charset=utf-8,"`

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,120 @@
...
+}
+
+function logo() {
+    // Please attribute mapbox if using non-commercially
+    // https://www.mapbox.com/help/how-attribution-works
+    var img = new Image();
+    var logo = mapboxlogo();
+    img.src = logo.image64;
+    img.onload = function() {
+    //(!) Doesn't work in Firefox
```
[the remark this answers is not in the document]
samsammurphy: It now works in Firefox. :tada:

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,120 @@
...
+    ctx.fillRect(snapshot.width - width, snapshot.height - height, width, height);
+    ctx.font = fontsize + 'px Arial';
+    ctx.fillStyle = 'black';
+    ctx.textAlign = 'left';
+    ctx.textBaseline = 'bottom';
+    ctx.fillText(text, snapshot.width - width + 3, snapshot.height - 3);
+}
+
+function logo() {
+    // Please attribute mapbox if using non-commercially
```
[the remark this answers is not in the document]
samsammurphy: suggestion implemented

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,120 @@
...
+    top:50%;
+    transform: translateY(-50%);
+    margin:0;
+  }
+ 
+</style>
+
+<div id='map' class='window'></div>
+
+<canvas id='snapshot' class='window'></div>
```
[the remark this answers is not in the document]
samsammurphy: added suggestions:
- border
- message overlaying existing snapshot

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,120 @@
...
+
+// draw snapshot canvas 
+window.addEventListener('load', function() {
+    var text = 'Click on the map';
+    snapshot.height = window.innerHeight;
+    snapshot.width = window.innerWidth / 2;
+    ctx.fillStyle = 'white';
+    ctx.textAlign = 'center';
+    ctx.textBaseline = 'middle';
+    ctx.font = Math.round(snapshot.width / 10) + 'px Arial';
```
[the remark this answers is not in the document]
samsammurphy: my understanding is that they cannot in this case, which is a canvas context object

docs/pages/example/snapshot.html
```diff
@@ -0,0 +1,120 @@
...
+    var fontsize = 14;
+    var height = fontsize + 6;
+    var width = (text.length + 3) / fontsize * 100;
+    ctx.fillStyle = 'rgba(255,255,255,0.4)';
+    ctx.fillRect(snapshot.width - width, snapshot.height - height, width, height);
+    ctx.font = fontsize + 'px Arial';
+    ctx.fillStyle = 'black';
+    ctx.textAlign = 'left';
+    ctx.textBaseline = 'bottom';
+    ctx.fillText(text, snapshot.width - width + 3, snapshot.height - 3);
```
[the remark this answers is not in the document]
samsammurphy: ditto

docs/pages/example/snapshot.html
samsammurphy: I think I am missing something. When I console.log the backgroundImage it seems to indicate it is a base64.

https://github.com/samsammurphy/mapbox-gl-js/blob/de8d14bc7d6c544b47950f68aa60ac3e2990e721/docs/pages/example/snapshot.html#L137

ryanhamley:
Hey @andrewharvey just FYI, you can change the target branch of a PR by clicking the Edit button at the top of the page.

docs/pages/example/snapshot.html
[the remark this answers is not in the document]
samsammurphy: added firefoxBugFix() to insert height and width.

docs/pages/example/snapshot.html
samsammurphy: simplified accordingly, just using dataURLs as standard approach (with firefox bugfix tacked on later)

mollymerp:
I've opened a new PR at #6846 – thanks for all your work on this @samsammurphy and @andrewharvey for reviwing

closing here to focus conversation in the new PR 👉

PR #6535, opened 2018-04-18
Add native sdk support entries for expressions

anandthakker:


anandthakker:
cc @mapbox/studio

anandthakker:
Does it make sense to go ahead and merge this, since we know these are the versions where expression support will land, or should we wait until we're out of beta/rc?

PR #6556, opened 2018-04-23
fix #6548, never group symbol layers by layout

### LINKED ISSUE #6548: Collision detection incorrect for symbol layers that share the same layout properties
When we generate symbol buckets for a tile, we use the `groupByLayout` function to group together all style layers that share a common set of layout properties and a common source layer. Since the two layers share common layout data, they can also share a `SymbolBucket`, even though they may render differently due to different paint properties.

The shared `SymbolBucket` breaks the current assumptions of our `CrossTileSymbolIndex`, because for each layer, we store each symbol's `crossTileID` in the associated bucket's `SymbolInstances`. When two symbols from two different layers share the same `SymbolInstance`, they can overwrite each other's `crossTileID`s.

One way this could manifest is:

- Layer1 and Layer2 share the same bucket, and the layers are allowed to collide with each other.
- Layer1 gets placed
- Layer2 gets placed, but matches all the symbols for Layer1, so every `SymbolInstance` is marked collided
- At render time, symbols in Layer1 pick up opacities from the collided Layer2, so that no symbols are shown from either layer.

I think the fix here is just to have `SymbolInstance` store crossTileIDs on a per-layer basis.

/cc @ansis

ansis:
@ChrisLoer this is a possible fix for #6548 and alternative to https://github.com/mapbox/mapbox-gl-js/pull/6550

Instead of storing `CrossTileIDs` per-layer this just makes it so that symbol buckets never have more than one layer. The layer grouping is an attempted optimization to reduce bucket creation time but this doesn't really matter for symbol layers because there aren't really very many (any?) production uses for having multiple layers with the same layout properties.

EDIT: at first this seemed like a nicer approach than https://github.com/mapbox/mapbox-gl-js/pull/6550 but I'm starting to lean back towards https://github.com/mapbox/mapbox-gl-js/pull/6550. The other pr seems like a cleaner more logical exception, but this approach could possibly avoid some future bugs from layer sharing.

## Launch Checklist

 - [N/A] post benchmark scores

### DIFF
--- a/src/data/bucket/symbol_bucket.js
+++ b/src/data/bucket/symbol_bucket.js
@@ -18,6 +18,7 @@ import Anchor from '../../symbol/anchor';
 import { getSizeData } from '../../symbol/symbol_size';
 import { register } from '../../util/web_worker_transfer';
 import EvaluationParameters from '../../style/evaluation_parameters';
+import assert from 'assert';
 
 import type {Feature as ExpressionFeature} from '../../style-spec/expression';
 import type {
@@ -294,6 +295,8 @@ class SymbolBucket implements Bucket {
         this.zoom = options.zoom;
         this.overscaling = options.overscaling;
         this.layers = options.layers;
+        // symbol layers are never grouped
+        assert(this.layers.length === 1);
         this.layerIds = this.layers.map(layer => layer.id);
         this.index = options.index;
         this.pixelRatio = options.pixelRatio;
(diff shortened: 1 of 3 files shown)


ChrisLoer:
I was thinking of an approach like this too, but I realized that this approach changes behavior from pre-#5150, while #6550 (I think) restores pre-#5150 behavior.

@jfirebaugh gave an example of something you might want to do that takes advantage of grouping symbol layers by layout: add a text layer, and then add a second with `text-translate` in order to implement a drop-shadow. Grouping by layout allows the two layers to share collision detection behavior. I don't think that's something we _have_ to support, but I think it's a good argument that, other things being equal, we should avoid changing behavior here.

Issue #6555, opened 2018-04-22
querySourceFeatures throws Uncaught TypeError in loadMatchingFeature


andrewharvey:
**mapbox-gl-js version**: regression in v0.45.0-beta.1, works fine on v0.44.2.

I'm labelling this as a release blocker as it seems like a pretty major bug in the beta release, please adjust if it's not. Thanks.

### Steps to Trigger Behavior

http://jsbin.com/wixaxim/2/edit?html,output

start clicking some points and the error below should popup in the console

### Actual Behavior

```
feature_index.js:163
Uncaught TypeError: Cannot read property 'decode' of undefined
    at ja.loadMatchingFeature (VM89 mapbox-gl.js:28)
    at x (VM89 mapbox-gl.js:28)
    at ja.query (VM89 mapbox-gl.js:28)
    at lt.queryRenderedFeatures (VM89 mapbox-gl.js:32)
    at at (VM89 mapbox-gl.js:32)
    at i.queryRenderedFeatures (VM89 mapbox-gl.js:32)
    at n.queryRenderedFeatures (VM89 mapbox-gl.js:32)
    at n.delegates.o.(/wixaxim/anonymous function) (https://api.tiles.mapbox.com/mapbox-gl-js/v0.45.0-beta.1/mapbox-gl.js:32:281769)
    at n.I.fire (VM89 mapbox-gl.js:28)
    at HTMLDivElement.<anonymous> (VM89 mapbox-gl.js:32)
```

/cc @ChrisLoer it looks related to https://github.com/mapbox/mapbox-gl-js/commit/fc992ec5997b9d4828bae42d342e6d420f2d91c2

ChrisLoer:
Yikes, thanks for the catch @andrewharvey! Yup, https://github.com/mapbox/mapbox-gl-js/commit/fc992ec5997b9d4828bae42d342e6d420f2d91c2 introduced a path where `vtLayers` could be set by not `sourceLayerIndex`. Should be an easy fix.

Sorry I didn't notice this earlier, it didn't make it into my "Mention" queue -- did you edit the original post to tag me?

andrewharvey:
> Sorry I didn't notice this earlier, it didn't make it into my "Mention" queue -- did you edit the original post to tag me?

@ChrisLoer Yeah I did, I never realised that meant the notification doesn't go through.

jfirebaugh:
Needs backporting to `release-0.45`.

PR #6559, opened 2018-04-24
Don't load FeatureIndex#vtLayers from outside FeatureIndex code.

ChrisLoer:
Fixes issue #6555 (a possible crash when calling queryRenderedFeatures after querySourceFeatures)
The actual contents of the regression test are not that important -- before this fix, the test would crash.

The problem was that `vtLayers` would be loaded without loading the `sourceLayerCoder`, and the FeatureIndex would assume both were loaded if `vtLayers` was loaded.

I've checked and this doesn't appear to have an analog on gl-native.

/cc @ansis @mollymerp @andrewharvey

### DIFF
--- a/src/data/feature_index.js
+++ b/src/data/feature_index.js
@@ -84,12 +84,17 @@ class FeatureIndex {
         }
     }
 
-    // Finds non-symbol features in this tile at a particular position.
-    query(args: QueryParameters, styleLayers: {[string]: StyleLayer}): {[string]: Array<{ featureIndex: number, feature: GeoJSONFeature }>} {
+    loadVTLayers(): {[string]: VectorTileLayer} {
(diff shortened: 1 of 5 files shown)


ChrisLoer:
I fixed the `Tile` unit test by switching to using a deserialized version of the FeatureIndex/CollisionBoxArray (necessary because `querySourceFeature` now needs to call the `loadVTLayers` method of `FeatureIndex`). I looked back a ways through the commit history and didn't figure out why we were using a serialized version in the first place, but I kept a serialize/deserialize round trip so that the unit test still exercises some of the expected pathway for data being provided by the web workers.

jfirebaugh:
@ChrisLoer Can you please backport this to the 0.45 release branch?

PR #6564, opened 2018-04-25
Allow integration tests to be organized in an arbitrarily deep directory structure

### LINKED ISSUE #3920: Allow integration tests to be organized in an arbitrarily deep directory structure
Migrated from https://github.com/mapbox/mapbox-gl-test-suite/issues/102

> There is currently a fixed test hierarchy: `[style property name]/[style property value]`. This system has not scaled elegantly as we've added tests for other sorts of things: tile extents, regression tests, query tests, etc. 
> - [ ] refactor to allow for an arbitrary test hierarchy
> - [ ] organize existing tests into a new-and-improved heirarchy
> - [ ] refactor query tests to use the operations API and merge them into the new hierarchy
> ----
> Sounds great @lucaswoj! It'd be useful when storing similar type tests like, geojson's nested here

jfirebaugh:
Fixes #3920.


test/integration/render-tests/result_item.html.tmpl
```diff
@@ -8,7 +8,7 @@
     <td></td><td></td>
 <% } %>
     <td>
-        <h2 style="background: <%- r.color %>"><a href="<%- r.group %>/<%- r.test %>/style.json"><%- r.group %>/<%- r.test %> <% if (!r.ok) { %>(<%=r.status%>)<% } %></a></h2>
```
anandthakker: Need this change in `{expression,query}-tests/result_item.html.tmpl` also

Issue #6290, opened 2018-03-07
Rollover effect with 0.45


phiphou:
I re-open https://github.com/mapbox/mapbox-gl-js/issues/6278 because I don't understand the answer. 

I want to be able to have multiple markers, not overlapping, and a rollover effect on either the markers or the text of the markers when displayed

Here is a sample : https://jsfiddle.net/d8qs32pw/15/ (zoom on the marker to see the second one)

It worked fine with` v0.41` and older versions.

I know there is a bug with `text-offset` in https://github.com/mapbox/mapbox-gl-js/issues/6146

@jfirebaugh told me to set `icon-ignore-placement` and `text-ignore-placement` on the markers layer to `true` but I don't want them to be overlapped. 

Setting  `text/icon-allow-overlap` to `true` on the hover layer as @ChrisLoer told me does not solved the problem of the weird hovering behavior.

So, I cannot use https://github.com/mapbox/mapbox-gl-js/issues/6163#issuecomment-365672211

If possible, can you update my example or tell me if `v0.45` will let me have the same behavior of `v0.41`

It's important for me and my clients to know witch version we can use to have this needed multiple markers (not overlapping) rollover effect.

Sorry for the flood, I know you're not stackoverflow, but I think it's currently a bug. 

Please don't close this before I can check the answer correctly fits my needs (and sorry for the bad english, I'm french)

ChrisLoer:
Hi @phiphou, oops we've got different time-zones making it harder to talk too!

Yes, we agree it's a regression from `v.41`, it was introduced by the architectural changes to support "cross-source collision detection" in PR #5150. We are tracking the issue as #5836. I don't know if it will make it into the `v.45` release, but I'm working on it -- and we'll keep everyone updated on that ticket. I used your fiddle to test that my proposed fix for #5836 would address your use case. I don't think you should try to use my fix yet, because it's not released yet and there's a good chance the implementation will change before release.

phiphou:
Thanks for your answer, I know understand better the problem, and will follow https://github.com/mapbox/mapbox-gl-js/issues/5836. 

For now I will stay with `v0.41`and wait for future releases to see when I can switch to an higher version.

PR #6566, opened 2018-04-25
Add "crossSourceCollisions" option to restore per-source collision detection

ChrisLoer:
This is an alternative PR to #6028 that removes `text-collision-group` and `icon-collision-group` from the style spec, and instead optionally groups layers based on their source. Setting `map.crossSourceCollisions = false` will restore the collision detection behavior from before cross-source collision detection (PR #5150).

The only change versus #6028 is the public interface -- the implementation under-the-hood is the same. The reasons for deliberately reducing the functionality are:

- Studio team is uncomfortable with supporting the previous "collision group" proposal. The concern is not so much the immediate cost of building UI to support it as the concern that similar functionality will be available in the future as part of a larger "style components" feature, and backwards compatibility would require us to maintain two separate ways of doing the same thing (a serious design smell).
- Since the use-cases we have in mind for wanting separate collision detection are generally of the form "put user-supplied data on top of a basemap", we expect that using "source" as the de-facto grouping mechanism for collision detection will be workable, even if it's not how we'd think about the problem to begin with. Also, we know it's backwards-compatible with pre-#5150 behavior.
- Although we don't escape backwards compatibility concerns by moving to the map, the situation improves:
  - Because the scope of the functionality is limited to "cross source placement", it should be relatively easy to implement continuing support for this option using whatever collision system we develop in the future.
  - _If_ we have to make a breaking change, it's not _as_ painful for the SDK as it is for the style-spec
  - By limiting this to a runtime option, we don't have to worry about dependence on this behavior working its way into widely used styles.

/cc @ansis @anandthakker @ttsirkia @samanpwbb @nickidlugash


anandthakker:
This compromise & rationale mostly sounds good to me.  One question: will @mapbox/studio support a way to toggle this runtime option on/off?  Without that, a designer building a map style intended for use with `crossSourceCollisions = false` wouldn't be able to do all of their design work in Studio.

ChrisLoer:
> One question: will @mapbox/studio support a way to toggle this runtime option on/off? Without that, a designer building a map style intended for use with crossSourceCollisions = false wouldn't be able to do all of their design work in Studio.

Good question: @samanpwbb?

samanpwbb:
> One question: will @mapbox/studio support a way to toggle this runtime option on/off? Without that, a designer building a map style intended for use with crossSourceCollisions = false wouldn't be able to do all of their design work in Studio.

Right now Studio auto-composites all tilesets by default, and doesn't un-composite even if you turn off the autocomposite feature, so I could see this causing more confusion than anything else if we were to add the option. It is possible, and if it's a frequent feature request we could consider doing it.

phiphou:
Does this PR solve my problem in #6290 ?

ChrisLoer:
I believe so: since the markers you want to have their own collision detection are already in the separate "markers" GeoJSON source, you would just have to set the map option `crossSourceCollisions: false`.

src/symbol/grid_index.js
```diff
@@ -117,17 +117,20 @@ class GridIndex {
...
+            }
         } else {
             const queryArgs = {
                 hitTest,
                 seenUids: { box: {}, circle: {} }
             };
-            this._forEachCell(x1, y1, x2, y2, this._queryCell, result, queryArgs);
+            this._forEachCell(x1, y1, x2, y2, this._queryCell, result, queryArgs, predicate);
         }
         return hitTest ? result.length > 0 : result;
```
anandthakker: Can this return now be moved into the `else` block?

src/symbol/placement.js
```diff
@@ -83,6 +82,45 @@ export class RetainedQueryData {
...
+    get(groupName?: string) {
+        if (groupName && this.maxGroupID === 0) {
+            // Keep the predicate null until the first collision
+            // group gets added to avoid overhead in the case
+            // everything's in the default group.
+            this.collisionGroups[undefined].predicate =
+                (key) => {
+                    return key.collisionGroup === 0;
+                };
+        }
```
anandthakker: This could cause a name collision if a source id (and thus the corresponding `groupName`) were `"undefined"`. Could the default group just be a separate `defaultGroup` property on the class, rather than stored in the `collisionGroups` map?

anandthakker: We could then also just reference `defaultGroup` instead of doing `collisionGroups.get()`

src/symbol/collision_index.js
```diff
@@ -298,17 +299,17 @@ class CollisionIndex {
         return result;
     }
 
-    insertCollisionBox(collisionBox: Array<number>, ignorePlacement: boolean, bucketInstanceId: number, featureIndex: number) {
+    insertCollisionBox(collisionBox: Array<number>, ignorePlacement: boolean, bucketInstanceId: number, featureIndex: number, collisionGroup: number) {
         const grid = ignorePlacement ? this.ignoredGrid : this.grid;
 
-        const key = { bucketInstanceId: bucketInstanceId, featureIndex: featureIndex };
+        const key = { bucketInstanceId: bucketInstanceId, featureIndex: featureIndex, collisionGroup: collisionGroup };
```
anandthakker: Nit: it might be helpful to rename this (both the key property and function parameter) to `collisionGroupID`, since `collisionGroup` could also refer to the group name

test/integration/render-tests/icon-collision-group/default/style.json
```diff
@@ -0,0 +1,145 @@
...
+                        }
+                    }
+                ]
+            }
+        }
+    },
+  "sprite": "local://sprites/sprite",
+    "layers": [
+        {
+            "id": "defaultGroup1",
```
anandthakker: Would this really use the "default" group?  My understanding is that with crossSourceCollisions: false, we're giving each source a named group.

src/symbol/placement.js
```diff
@@ -83,6 +82,45 @@ export class RetainedQueryData {
...
+                predicate: null
+            }
+        };
+    }
+
+    get(groupName?: string) {
+        if (groupName && this.maxGroupID === 0) {
+            // Keep the predicate null until the first collision
+            // group gets added to avoid overhead in the case
+            // everything's in the default group.
```
anandthakker: Will this ever happen in the current implementation?  It looks to me like if `crossSourceCollisions` is set `false`, then every layer gets a named group, and otherwise, every layer uses the default group.

ansis:
> Studio team is uncomfortable with supporting the previous "collision group" proposal. The concern is not so much the immediate cost of building UI to support it as the concern that similar functionality will be available in the future as part of a larger "style components" feature

@samanpwbb how do you see components and collisions working? I haven't been following those discussions closely but my gut reaction is that sometimes you'd want two components to collide with each other and sometimes not. It seems like these are two separate concerns?

> map.crossSourceCollisions = false

I pretty hesitant about adding a map option that affects the style of the map. If we're willing to make this a public ui I *think* we should put it in the spec.

> If we have to make a breaking change, it's not as painful for the SDK as it is for the style-spec

It's easier on our side because we don't need to handle migration, but for users adapting to the change would be the same no matter what. Even if we don't migrate styles.

---

My gut reaction is that we should try to figure out how collision groups would work with style components. If collisions are not handled at the style component level we should probably do https://github.com/mapbox/mapbox-gl-js/pull/6028 instead of this intermediate solution.

samanpwbb:
> @samanpwbb how do you see components and collisions working? I haven't been following those discussions closely but my gut reaction is that sometimes you'd want two components to collide with each other and sometimes not. It seems like these are two separate concerns?

Here's an example component, with a `crossComponentCollisions` property:

```js
{
  "id": "my-component",
  "name": "My Component",
  "sources": { ... },
  "layers": [
    {"id": "layer-id-1", "type": "fill", "layout": {}, "paint": {} },
    {"id": "layer-id-2", "type": "fill", "layout": {}, "paint": {} },
    ...
  ]
  // Options allows user to configure layers/sources in component
  "options": {
      "water": {
          "title": "Water",
          "type": "color",
          "doc": "Color of oceans, lakes, rivers, and other bodies of water.",
          "default": "#8ccbf7"
      },
      "language": {
          "title": "Language",
          "type": "enum",
          "values": ["default", "en", "fr", "ar"]
      }
  },
  // This property would act the same as colliisionGroup property,
  // But for all layers within the component
  "crossComponentCollisions": false
}
```

Imagine an app that uses the `collisionGroup` layer property. The app involves frequently adding and removing layers. The implementor would need to manage layers twice: the adding/removing of layers themselves, and then they'd need to update which layers belong to which collision group.

Components could also serve the purpose of managing which layers belong to which collision group. 

I'm most concerned about a future where we have both _components_, and _collision groups_ – essentially two competing state management systems.

src/symbol/placement.js
ChrisLoer: Yeah, this is probably weird dead code to leave around if we're going to use the `crossSourceCollisions: bool` approach. We might try to add something to skip the expense if `crossSourceCollisions` is `true` and there's only one source, but it's probably not an important optimization.

src/symbol/grid_index.js
ChrisLoer: Yeah, that might be a cleaner way to write it, but we need to make sure to return something in the first clause if `predicate` is undefined.

ansis:
Thanks @samanpwbb

> Imagine an app that uses the collisionGroup layer property. The app involves frequently adding and removing layers. The implementor would need to manage both the adding/removing of layers themselves, and then they'd need to update which layers belong to which collision group. That makes the amount of state the app needs to grapple with higher than it needs to be, imo.



If I'm understanding the example correctly, the advantage of of specifying the collision group on the component instead of the layer is that you can set the property once and have it automatically apply to a whole bunch of layers. I think this reasoning could apply to other properties as well. What makes `collision-group` special when compared to other properties like `text-font`? Would setting the `collision-group` for each layer in the example be different than setting the font for each layer that gets added and removed?

I'm not sure I understand why collision group should be treated as an exception

One possible approach alternative approach to reducing verbosity/redundancy could be to provide some way of specifying default values at the component level. The component creator could then do `"defaults": { "text-collision-group": "my-component-name", "text-font": "My Font" }` and add layers without touching these. This adds complexity though

samanpwbb:
> What makes collision-group special when compared to other properties like text-font?

The conponent abstraction does benefit both properties. The difference is that a Collision group string property would be the only property that has interdependencies with other layers. Changing text-font on a layer is immediately apparent and won’t be effected by changes to other layers. Components would make the concept behind a collision group simpler to understand and work with.

ChrisLoer:
Hi @cammanderson sorry for the delay, we're still trying to figure out the most consistent way to implement this so that we don't introduce inconsistencies in the style-spec that may conflict with future functionality. It would help us reason about the problem if you could provide some detail about what you're trying to accomplish and which part gets blocked by the current behavior.

cammanderson:
Thanks for the update. We are placing symbols on a map across a number of locations. We used to have symbols sit on top of towns labels etc, and the town labels would still be present as our symbol data was from another source (think store locator when zoomed out you still want to see the spread of stores but maintain the contextual labeling). Now all context is lost as all the map location labels have disappeared due to the collision behaviour changing. We can’t really change the location label behaviour because they were working as intended to resolve their conflicts with other location labels as desired. We also can’t use marker controls as we are using a lot of symbols from our data sources (eg vector tiles in some cases) and rely also on those symbols collision behaviour also. Our maps aren’t static enough for us to hardcode or have a reasonable workaround or a scalable workaround. My only option would be to revert to a version of mapbox that supports data driven expressions and then manually apply the major iOS bug fix that came in later releases by itself, which came after this global collision behaviour change. I’m not even sure if that’s possible but I guess I just need to know how long as we have several deadlines fast approaching. I can appreciate the challenge of updating the style spec and we don’t want to create junk, I’m an advocate of the collision groups PR. What I’m unclear about is why Studio UI support would drive the style spec? Many of us affected are using the API and various JS from your examples to build additional data sources and layers dynamically on top of base styles. I can appreciate the PR to change the map property in one place as a quicker work around, but either way it’s a tricky scenario where due to major compliance with iOS releases we are forced to adopt the versions where the global collision behaviour has changed with a dramatic effect of some existing use cases that I suspect are not niche.

src/symbol/placement.js
ChrisLoer: I changed `CollisionGroups` to explicitly work with source IDs now, which avoids the dead code and any worries about how to manage a "default" group.

test/integration/render-tests/icon-collision-group/default/style.json
ChrisLoer: Updated the names of the tests and the IDs of these layers to get rid of the vestigial grouping logic.

src/symbol/grid_index.js
ChrisLoer: Updated.

ChrisLoer:
I just rebased on master and addressed @anandthakker's review comments, but we're still looking for consensus on the design direction.

Thanks for the extra context @cammanderson! From the input I've seen, it seems like a common thread is thinking of basemap labels as part of the "background" for symbols added at runtime to be displayed on top of. I think our default response to the "marker on top of a location label" case has been something along the lines of "isn't it better for legibility if the location label gets hidden since it's partially occluded". From your description of your case, it sounds like (1) it's frequently the case that _enough_ of the location label is visible for it to be legible, (2) marker/location-label overlap isn't a visual problem because the marker has an opaque background, (3) if markers _are_ allowed to collide out location labels, the location-label density ends up too low to be useful?

>  What I’m unclear about is why Studio UI support would drive the style spec?

We want Studio to be capable of creating/editing any Mapbox style. Blocking style spec changes that don't have a matching Studio UI design helps us achieve that goal, but I think it also forces us to have a higher bar for what we accept into the style spec (i.e. if we're not sure what the matching Studio UI for a change is, that implies we're introducing a change to the style spec that departs from existing conventions and may cause trouble for other users as well).

ChrisLoer:
For the archaeological record: we deferred implementing this on gl-native because we weren't totally comfortable with the design direction and we were hoping that there might not be large gl-native users dependent on the old behavior. I think "wait and see" was a reasonable strategy, but in this case it only took ~2.5 months before the native version of this showed up as a high priority customer issue.

Issue #6536, opened 2018-04-18
Map.queryRenderedFeatures uses stale tile padding


asheemmamoowala:
The `Tile.queryPadding` value is not updated after a call to `Map.setPaintProperty()`. 

The feature states PR (#6263), introduces an additional API through which style layer paint properties can be updated dynamically. This was also highlighted [here(comment)](https://github.com/mapbox/mapbox-gl-js/pull/6263#pullrequestreview-110201228)

Dynamically computing `Tile.queryPadding` would address these bugs, and ensure that the padding reflects current tile state.

Issue #6236, opened 2018-02-26
Explain that filters are evaluated only at integer zoom levels


anandthakker:
We explain this for layer layout properties, but now that we support `["zoom"]` expressions in filters, we should explain it in the filter docs as well.

PR #6569, opened 2018-04-26
queryRenderedFeatures doesnt take into account feature states

asheemmamoowala:
Addresses #6536 .

When updating a feature's paint properties using `Map.setFeatureState`, each tile should refresh its `queryPadding.`  Additionally, feature state needs to be present on the `feature` object when evaluating paint properties for the intersection test.

### DIFF
--- a/src/source/query_features.js
+++ b/src/source/query_features.js
@@ -24,7 +24,7 @@ export function queryRenderedFeatures(sourceCache: SourceCache,
             wrappedTileID: tileIn.tileID.wrapped().key,
             queryResults: tileIn.tile.queryRenderedFeatures(
                 styleLayers,
-                sourceCache.getFeatureState.bind(sourceCache),
+                sourceCache._state,
                 tileIn.queryGeometry,
                 tileIn.scale,
                 params,
--- a/src/source/query_features.js
+++ b/src/source/query_features.js
@@ -24,6 +24,7 @@ export function queryRenderedFeatures(sourceCache: SourceCache,
             wrappedTileID: tileIn.tileID.wrapped().key,
             queryResults: tileIn.tile.queryRenderedFeatures(
                 styleLayers,
+                sourceCache.getFeatureState.bind(sourceCache),
                 tileIn.queryGeometry,
                 tileIn.scale,
                 params,
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -168,7 +168,7 @@ class SourceCache extends Evented {
             this._source.prepare();
         }
 
-        this._state.coalesceChanges(this._tiles);
+        this._state.coalesceChanges(this._tiles, this.map ? this.map.painter : null);
         for (const i in this._tiles) {
             this._tiles[i].upload(context);
         }
@@ -251,7 +251,7 @@ class SourceCache extends Evented {
         if (previousState === 'expired') tile.refreshedUponExpiration = true;
         this._setTileReloadTimer(id, tile);
         if (this.getSource().type === 'raster-dem' && tile.dem) this._backfillDEM(tile);
-        this._state.initializeTileState(tile);
+        this._state.initializeTileState(tile, this.map ? this.map.painter : null);
 
         this._source.fire(new Event('data', {dataType: 'source', tile: tile, coord: tile.tileID}));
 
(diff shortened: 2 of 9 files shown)


src/source/source_cache.js
```diff
@@ -168,7 +168,7 @@ class SourceCache extends Evented {
             this._source.prepare();
         }
 
-        this._state.coalesceChanges(this._tiles);
+        this._state.coalesceChanges(this._tiles, this.map ? this.map.painter : null);
```
asheemmamoowala: @anandthakker, in [#6236(comment)](https://github.com/mapbox/mapbox-gl-js/pull/6263#discussion_r177534083), we discussed (and agreed upon) : 
>-  move the responsibility for calling Tile#updateFeatureState during prepare() directly into SourceFeatureState#coalesceChanges() (i.e., take a set of tiles as a parameter to coalesceChanges)?
> - Add a SourceFeatureState#initializeTileState(Tile) method and replace the other two cases of tile.updateFeatureState(this._state.state); with this._state.initializeTileState(tile)?

What do you think of moving back the tile-specific code from `this._state.coalesceChanges` into `SourceCache`, to reduce the `painter` spaghetti created here. 
Not a burning issue though.

src/source/query_features.js
```diff
@@ -24,6 +24,7 @@ export function queryRenderedFeatures(sourceCache: SourceCache,
             wrappedTileID: tileIn.tileID.wrapped().key,
             queryResults: tileIn.tile.queryRenderedFeatures(
                 styleLayers,
+                sourceCache.getFeatureState.bind(sourceCache),
```
asheemmamoowala: Passing the function because it maintains different feature states for each `sourceLayer`.

src/source/source_cache.js
anandthakker: I'm thinking we should keep this as-is, because I expect this bit of  🍝 to be temporary: via https://github.com/mapbox/mapbox-gl-js/issues/4875 and https://github.com/mapbox/mapbox-gl-js/issues/4012, I'm hoping we'll end up in a place where each tile can easily hold on to a view of the style layers that were used to lay it out, and that will let us eliminate the need to thread `painter` through here.

src/source/tile.js
```diff
@@ -418,7 +419,7 @@ class Tile {
         }
     }
 
-    setFeatureState(states: LayerFeatureStates) {
+    setFeatureState(states: LayerFeatureStates, painter: any) {
```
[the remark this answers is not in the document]
asheemmamoowala: I used `painter` to maintain similarity with `loadVectorData`, but it doesn't have to be that way - for both these methods. The only advantage I see of passing `painter` as the argument is that it communicates that the _render style_ is being used. This becomes important when considering the Immutable Styl/Render-classes split.

src/source/query_features.js
anandthakker: Hmm -- there are several other places in the code that already have awareness of `_geojsonTileLayer`...  as such, I'd be inclined to say we should just dupe the logic, or else move it into `SourceFeatureState`.

src/source/tile.js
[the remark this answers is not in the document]
anandthakker: Oh, yeah that's odd that `loadVectorData` takes a painter rather than a style object.  I lean slightly towards passing `Style` here, but since this will get refactored in the nearish future with the immutable style state changes, it's not a big deal either way.  

> The only advantage I see of passing painter as the argument is that it communicates that the render style is being used. This becomes important when considering the Immutable Styl/Render-classes split

I think the correct style to use here will be the one that the `tile` will already hold a reference to -- namely, the one that was used at layout time, since that's the one that will be consistent with the data in the buckets.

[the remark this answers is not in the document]
asheemmamoowala: >  since this will get refactored in the nearish future with the immutable style state changes, it's not a big deal either way.

I'd prefer to keep it as-is in that case.

> I think the correct style to use here will be the one that the tile will already hold a reference to

👍

src/source/query_features.js
asheemmamoowala: @anandthakker What is the advantage of duplicating the logic over passing the bound function? This PR has already reduced the _spred_ of `_geojsonTileLayer` by consolidating it in `SourceCache.`

anandthakker: It just seems clearer to me: I find a callback as a method of passing data to be harder to read and understand as compared to passing an object.  Would moving the `_geojsonTileLayer` logic into `SourceFeatureState` give us the best of both worlds?

asheemmamoowala: > Would moving the _geojsonTileLayer logic into SourceFeatureState give us the best of both worlds?

I don't think so. I could change the name of the callback from `getState` to `queryState`, but all of these changes seem minor and not that significant advantages to whats already here.

ansis: I agree with @anandthakker that moving the logic into `SourceFeatureState` or duping the logic would be a bit clearer than passing the callback here

PR #6573, opened 2018-04-27
List gl-accessibility plugin on plugins page

1ec5:
Added the [gl-accessibility](https://github.com/mapbox/mapbox-gl-accessibility/) plugin to the “Plugins” listing. Also realphabetized the UI plugins section.

@tristen, do you think we should hold off on this until the plugin is no longer considered experimental? Or should we add it now since it’s already functional?

### DIFF
--- a/docs/pages/plugins.js
+++ b/docs/pages/plugins.js
@@ -15,6 +15,10 @@ const meta = {
 
 const plugins = {
     "User Interface Plugins": {
+        "mapbox-gl-accessibility": {
+            "website": "https://github.com/mapbox/mapbox-gl-accessibility/",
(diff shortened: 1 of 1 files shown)


andrewharvey:
> oh wow @1ec5 ￼ I think including this makes sense to do. It works(ish) and more exposure to encourage contributions is

@1ec5 @tristen I'm so glad this was done, 100% agree with @tristen more exposure will encourage contribution. I'm sure there are people out there implementing this themselves (I was one of them) and after seeing this would likely contribute.

andrewharvey:
I just realised this probably should have been merged against `mb-pages` to go [live straight](https://www.mapbox.com/mapbox-gl-js/plugins/) away rather than next release. Not a big deal.

PR #6580, opened 2018-04-30
Updated Examples to use Map#setFeatureState

asheemmamoowala:
Updated the tilesets/geojson sources to include top level feature Ids. These are required for using `Map#setFeatureState` at this time.

Updated the Data Join example to use a single `feature-state` dependent expression and using `Map#setFeatureState` to join the JSON data with the vector tile source. The result is a fixed size expression, as opposed to an expression with individual match entries for each entry in the JSON data.

Update the Hover Styles example to use a single data-driven styling fill layer with `feature-state` dependent `fill-opacity`. Use `Map#setFeatureState` to set/unset the `hover` state on the feature under the mouse.

cc: @ryanbaumann

### DIFF
--- a/docs/pages/example/data-join.html
+++ b/docs/pages/example/data-join.html
@@ -12,58 +12,58 @@ var map = new mapboxgl.Map({
 // Join local JSON data with vector tile geometry
 // USA unemployment rate in 2009
 // Source https://data.bls.gov/timeseries/LNS14000000
-var maxValue = 13;
+var maxValue = 14;
 var data = [
-    {"STATE_ID": "01", "unemployment": 13.17},
-    {"STATE_ID": "02", "unemployment": 9.5},
-    {"STATE_ID": "04", "unemployment": 12.15},
-    {"STATE_ID": "05", "unemployment": 8.99},
-    {"STATE_ID": "06", "unemployment": 11.83},
-    {"STATE_ID": "08", "unemployment": 7.52},
-    {"STATE_ID": "09", "unemployment": 6.44},
-    {"STATE_ID": "10", "unemployment": 5.17},
-    {"STATE_ID": "12", "unemployment": 9.67},
-    {"STATE_ID": "13", "unemployment": 10.64},
-    {"STATE_ID": "15", "unemployment": 12.38},
-    {"STATE_ID": "16", "unemployment": 10.13},
-    {"STATE_ID": "17", "unemployment": 9.58},
-    {"STATE_ID": "18", "unemployment": 10.63},
-    {"STATE_ID": "19", "unemployment": 8.09},
-    {"STATE_ID": "20", "unemployment": 5.93},
-    {"STATE_ID": "21", "unemployment": 9.86},
-    {"STATE_ID": "22", "unemployment": 9.81},
-    {"STATE_ID": "23", "unemployment": 7.82},
-    {"STATE_ID": "24", "unemployment": 8.35},
-    {"STATE_ID": "25", "unemployment": 9.1},
-    {"STATE_ID": "26", "unemployment": 10.69},
-    {"STATE_ID": "27", "unemployment": 11.53},
-    {"STATE_ID": "28", "unemployment": 9.29},
-    {"STATE_ID": "29", "unemployment": 9.94},
-    {"STATE_ID": "30", "unemployment": 9.29},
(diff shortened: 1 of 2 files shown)


ryanbaumann:
We'll want to provide solid documentation around this feature if it only supports feature-level (not property-level) unique identifiers.  Currently, developer workflows using the data-join technique all involve exploiting property expressions.  This would be a significant change from that workflow.

docs/pages/example/data-join.html
```diff
@@ -72,30 +72,35 @@
     // https://www.census.gov/geo/maps-data/data/cbf/cbf_state.html
     map.addSource("states", {
         type: "vector",
-        url: "mapbox://mapbox.us_census_states_2015"
+        url: "mapbox://mapbox.10asuvo4",
```
mollymerp: I'd rename the tileset to have a self-explanatory name and/or add a comment explaining what data the tileset contains.

docs/pages/example/data-join.html
```diff
@@ -72,30 +72,35 @@
...
     map.addLayer({
         "id": "states-join",
         "type": "fill",
         "source": "states",
         "source-layer": "states",
         "paint": {
-            "fill-color": expression
+            "fill-color": ["rgba",
+                0,
+                [ "*", [ "/", ["to-number", ["feature-state", "unemployment"]], maxValue], 255],
```
mollymerp: Would be nice to include a comment here (as we had before) along the lines of 'calculate green color band value based on unemployment percentage'

docs/pages/example/data-join.html
```diff
@@ -72,30 +72,35 @@
...
+            map.setFeatureState({source: 'states', sourceLayer: 'states', id: row["STATE_ID"]}, row);
+        });
+    }
+
+    if (map.isSourceLoaded('states')) {
+        setStates();
+    } else map.on('sourcedata', function(e) {
+        if (e.sourceId === 'states' && e.isSourceLoaded) {
+            setStates();
+        }
```
mollymerp: Do you want to turn this event listener off after the states have been set?

docs/pages/example/data-join.html
```diff
@@ -72,30 +72,35 @@
...
+                0,
+                [ "*", [ "/", ["to-number", ["feature-state", "unemployment"]], maxValue], 255],
+                0,
+                0.9]
         }
     }, 'waterway-label');
+
+    function setStates() {
+        data.forEach(function(row) {
+            map.setFeatureState({source: 'states', sourceLayer: 'states', id: row["STATE_ID"]}, row);
```
mollymerp: I think a comment here describing what this does would be useful – I don't think it would necessarily be obvious to a user that you're associating the JSON data to the features on the map.

docs/pages/example/hover-styles.html
```diff
@@ -35,27 +40,22 @@
...
-        },
-        "filter": ["==", "name", ""]
-    });
-
     // When the user moves their mouse over the states-fill layer, we'll update the filter in
     // the state-fills-hover layer to only show the matching state, thus making a hover effect.
     map.on("mousemove", "state-fills", function(e) {
-        map.setFilter("state-fills-hover", ["==", "name", e.features[0].properties.name]);
+        if (e.features.length > 0) {
+            if (hoveredStateId) {
```
mollymerp: should we check here if `e.features[0].id == hoverstateid` and if so return early?

docs/pages/example/hover-styles.html
```diff
@@ -35,27 +40,22 @@
...
+            }
+            hoveredStateId = e.features[0].id;
+            map.setFeatureState({source: 'states', id: hoveredStateId}, { hover: true});
+        }
     });
 
     // Reset the state-fills-hover layer's filter when the mouse leaves the layer.
     map.on("mouseleave", "state-fills", function() {
-        map.setFilter("state-fills-hover", ["==", "name", ""]);
+        map.setFeatureState({source: 'states', id: hoveredStateId}, { hover: false});
```
mollymerp: is `hoveredStateId` guaranteed to be defined? It seems like yes, but not 100% sure

docs/pages/example/data-join.html
ryanbaumann: We can re-upload the tileset with a specific `tilesetid` for improved self-documenting names.

PR #6582, opened 2018-04-30
Fix merge collision between PR#6559 and PR#6263.

ChrisLoer:
- Query tests now have extra data in them
- Tile#setFeatureState needs to switch to getting vtLayers from FeatureIndex.

For what it's worth, this is the first breaking merge collision I've seen in the last year and a half...

/cc @asheemmamoowala @jfirebaugh

### DIFF
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -413,17 +413,19 @@ class Tile {
     }
 
     setFeatureState(states: LayerFeatureStates) {
-        if (!this.latestRawTileData || Object.keys(states).length === 0) return;
-
-        if (!this.vtLayers) {
-            this.vtLayers = new vt.VectorTile(new Protobuf(this.latestRawTileData)).layers;
+        if (!this.latestFeatureIndex ||
+            !this.latestFeatureIndex.rawTileData ||
+            Object.keys(states).length === 0) {
+            return;
         }
+
+        const vtLayers = this.latestFeatureIndex.loadVTLayers();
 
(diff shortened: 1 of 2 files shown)


src/source/tile.js
```diff
@@ -413,17 +413,19 @@ class Tile {
     }
 
     setFeatureState(states: LayerFeatureStates) {
-        if (!this.latestRawTileData || Object.keys(states).length === 0) return;
-
-        if (!this.vtLayers) {
-            this.vtLayers = new vt.VectorTile(new Protobuf(this.latestRawTileData)).layers;
+        if (!this.latestFeatureIndex ||
+            !this.latestRawTileData ||
```
asheemmamoowala: Does it need to check for `this.latestRawTileData` since `this.latestFeatureIndex` has its own copy of the raw tile data

asheemmamoowala: Does it need to check for `this.latestRawTileData` since `this.latestFeatureIndex` has its own copy of the raw tile data

ChrisLoer: Ah good point. It should check `this.latestFeatureIndex.rawTileData`. They'll have the same value because of the logic in `loadVectorData`, but the state of `FeatureIndex` is what we care about directly. Also, the tile workers should never generate a `FeatureIndex` that doesn't have an accompanying `rawTileData` (unless they're doing an update)... but since that's outside of the control of `Tile`, I think we still want to maintain the check.

PR #6586, opened 2018-05-01
Ensure style-spec is published with mapbox-gl

anandthakker:
Includes a copy of the compiled style-spec bundle under
`dist/style-spec/index.js` in the published mapbox-gl package

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -131,6 +131,8 @@ jobs:
       - run: yarn run build-min
       - run: yarn run build-dev
       - run: yarn run build-css
+      - run: yarn run build-style-spec
+      - run: yarn run test-build
       - persist_to_workspace:
           root: .
           paths:
--- a/package.json
+++ b/package.json
@@ -120,6 +120,7 @@
     "watch-dev": "rollup -c --environment BUILD:dev --watch",
     "build-min": "rollup -c --environment BUILD:production && build/run-tap --no-coverage test/build/min.test.js",
     "build-css": "postcss -o dist/mapbox-gl.css src/css/mapbox-gl.css",
+    "build-style-spec": "cd src/style-spec && npm run build && cd ../.. && cp -R src/style-spec/dist dist/style-spec && build/run-tap --no-coverage test/build/style-spec.test.js",
     "watch-css": "postcss --watch -o dist/mapbox-gl.css src/css/mapbox-gl.css",
     "build-token": "node build/generate-access-token-script.js",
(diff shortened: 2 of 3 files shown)


anandthakker:
cc @samanpwbb

package.json
```diff
@@ -141,13 +142,14 @@
     "test-suite": "run-s test-render test-query",
     "test-suite-clean": "find test/integration/{render,query}-tests -mindepth 2 -type d  -not \\( -exec test -e \"{}/style.json\" \\; \\) -print | xargs -t rm -r",
     "test-unit": "build/run-tap --reporter classic --no-coverage test/unit",
+    "test-build": "build/run-tap --no-coverage test/build/**/*.test.js",
```
jfirebaugh: Does this get run in CI?

package.json
```diff
@@ -116,10 +116,11 @@
     "sourceCache": true
   },
   "scripts": {
-    "build-dev": "rollup -c --environment BUILD:dev && build/run-tap --no-coverage test/build/dev.test.js",
+    "build-dev": "rollup -c --environment BUILD:dev",
     "watch-dev": "rollup -c --environment BUILD:dev --watch",
-    "build-min": "rollup -c --environment BUILD:production && build/run-tap --no-coverage test/build/min.test.js",
+    "build-min": "rollup -c --environment BUILD:production",
     "build-css": "postcss -o dist/mapbox-gl.css src/css/mapbox-gl.css",
+    "build-style-spec": "cd src/style-spec && npm run build && cd ../.. && cp -R src/style-spec/dist dist/style-spec",
```
jfirebaugh: Any particular reason for `npm` rather than `yarn`?

anandthakker: Nope, just habit, will change.

package.json
anandthakker: Oops, good catch -- accidentally omitted the circle.yml change

circle.yml
```diff
@@ -131,6 +131,8 @@ jobs:
       - run: yarn run build-min
       - run: yarn run build-dev
       - run: yarn run build-css
+      - run: yarn run build-style-spec
+      - run: yarn run test-build
```
anandthakker: @jfirebaugh do you see any reason to do `test-build` as a separate workflow step?  I thought it kinda made sense to keep it here, but 🤷‍♂️

jfirebaugh: No, adding it here is what I would have done too.

package.json
anandthakker: Reverted to `npm` due to https://github.com/mapbox/mbgl-ci-images/issues/17

PR #6588, opened 2018-05-01
Prevent CollatorExpression from evaluating as "constant"

ChrisLoer:
Although the results shouldn't change between evaluations, they may change between environments, so we don't want constant folding to make serializations that are inconsistent on different machines.
`CollatorExpression#serialize` was not previously used by the test suite because of constant folding, and was broken -- it needed to explicitly serialize its children.

/cc @anandthakker @jfirebaugh

### DIFF
--- a/test/integration/expression-tests/collator/accent-equals-de/test.json
+++ b/test/integration/expression-tests/collator/accent-equals-de/test.json
@@ -37,7 +37,17 @@
     "outputs": [true, false],
     "serialized": [
       "case",
-      false,
+      [
+        "==",
+        [
+          "resolved-locale",
+          [
+            "collator",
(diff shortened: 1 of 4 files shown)


test/integration/expression-tests/collator/accent-equals-de/test.json
```diff
@@ -37,7 +37,17 @@
...
     "serialized": [
       "case",
-      false,
+      [
+        "==",
+        [
+          "resolved-locale",
+          [
+            "collator",
+            {"caseSensitive": true, "diacriticSensitive": false, "locale": "de"}
```
jfirebaugh: Not related to this change, but these should be spelled `case-sensitive` and `diacritic-sensitive`.

PR #6594, opened 2018-05-02
Flow 0.69

jfirebaugh:
Upgrade flow to [0.69](https://github.com/facebook/flow/releases/tag/v0.69.0). Not upgrading past that due to facebook/flow#6151.

One of the nice features in [0.68](https://github.com/facebook/flow/releases/tag/v0.68.0) is:

> Previously, Flow would allow you to write `if (foo.unknownProp) { ... }`.
> Now Flow disallows testing unknown properties in conditionals. If foo is a union type like `{ x: string } | { y: number }`, `x` and `y` are known properties and `z` would be an unknown property

I remember finding a few bugs that would have been caught by this.


src/style-spec/expression/definitions/index.js
```diff
@@ -172,7 +172,7 @@ CompoundExpression.register(expressions, {
     'feature-state': [
         ValueType,
         [StringType],
-        (ctx, [key]) => get(key.evaluate(ctx), ctx.state())
+        (ctx, [key]) => get(key.evaluate(ctx), ctx.featureState || {})
```
asheemmamoowala: Should there be a check for `feature` when using `featureState`? Feature state is only valid if there is a current feature. As it stands, it's possible to call `Expression#evaluate` with a `null` feature, but non-null `featureState.`

asheemmamoowala: Should there be a check for `feature` when using `featureState`? Feature state is only valid if there is a current feature. As it stands, it's possible to call `Expression#evaluate` with a `null` feature, but non-null `featureState.`

jfirebaugh: I'd like to fix it this way:

```diff
diff --git a/src/style-spec/expression/index.js b/src/style-spec/expression/index.js
index 1310db3f2..269f718ec 100644
--- a/src/style-spec/expression/index.js
+++ b/src/style-spec/expression/index.js
@@ -179,18 +179,18 @@ export class ZoomDependentExpression<Kind> {
 
 export type ConstantExpression = {
     kind: 'constant',
-    +evaluate: (globals: GlobalProperties, feature?: Feature) => any,
+    +evaluate: (globals: GlobalProperties) => any,
 }
 
 export type SourceExpression = {
     kind: 'source',
     isStateDependent: boolean,
-    +evaluate: (globals: GlobalProperties, feature?: Feature, featureState?: FeatureState) => any,
+    +evaluate: (globals: GlobalProperties, feature: Feature, featureState?: FeatureState) => any,
 };
 
 export type CameraExpression = {
     kind: 'camera',
-    +evaluate: (globals: GlobalProperties, feature?: Feature, featureState?: FeatureState) => any,
+    +evaluate: (globals: GlobalProperties, feature: Feature, featureState?: FeatureState) => any,
     +interpolationFactor: (input: number, lower: number, upper: number) => number,
     zoomStops: Array<number>
 };
@@ -198,7 +198,7 @@ export type CameraExpression = {
 export type CompositeExpression = {
     kind: 'composite',
     isStateDependent: boolean,
-    +evaluate: (globals: GlobalProperties, feature?: Feature, featureState?: FeatureState) => any,
+    +evaluate: (globals: GlobalProperties, feature: Feature, featureState?: FeatureState) => any,
     +interpolationFactor: (input: number, lower: number, upper: number) => number,
     zoomStops: Array<number>
 };
```

But this requires https://github.com/mapbox/mapbox-gl-js/issues/6255#issuecomment-369439525, so I'd like to defer it.

However, in looking at this, I noticed that I'd failed to update `ZoomDependentExpression#evaluate[WithoutErrorHandling]` -- I've added additional render tests to cover that case.

asheemmamoowala: This seems reasonable. #6255 is needed soon, and I don't see an actual code path at this time where this issue would come up.

PR #6602, opened 2018-05-03
Use named exports in style-spec.js

### LINKED ISSUE #6601: Use named exports for style-spec entrypoint module
Currently, because style-spec.js uses a default export, the compiled bundle's exports are behind `default` -- i.e., you have to do `const validate = require('style-spec').default.validate` instead of `const validate = require('style-spec').validate`.

Switching to named exports should restore the expected bundling behavior, and anyway is in line with https://github.com/mapbox/mapbox-gl-js/issues/6336

anandthakker:
Fixes #6601


anandthakker:
cc @mapbox/maps-api @mapbox/studio

ian29:
thx @anandthakker <3

PR #6610, opened 2018-05-04
Fix ajax when loading local files on iOS web view

oscarfonts:
When running in iOS webview (cordova app), we detected that loading local resources (`file:///...` urls) is returning a "0" status code (instead of "200 OK").

Included this case in util.ajax, so local files are fetched without checking the response status code value.

Tried to add a unit test, but sinon's FakeXMLHttpRequest doesn't provide the standard responseURL property we are reading. At least made sure actual unit tests are passing. :)


mourner:
@oscarfonts looks like this PR has gone askew. Can you rebase on master and make sure the PR only contains the commit you added?

I'm also not sure this should be fixed on the GL JS side since it follows the specs, and requesting `file://` files with Ajax is usually prohibited due to security restrictions. Could there be a workaround on the Cordova/app side instead?

oscarfonts:
Cleared PR with rebase. Thanks!

Didn't find a simpler way to pass a local (offline) resource to mapbox-gl.

Alternatives were: either forking the library to replace ajax calls with some other (non-http, probably environment specific) mechanism to access local resources, or embedding a local http server into the App.

I can keep my fork if you feel this too hacky, I just thought it could go upstream for a more offline-friendly library, so to speak.

mourner:
@oscarfonts is the way Cordova behaves with Ajax + local files (e.g. the 0 status code) documented somewhere?

oscarfonts:
The description of this behavior can be found on different forums (pouchdb, jquery, phonegap):

https://github.com/pouchdb-community/pouchdb-load/issues/11
https://forum.jquery.com/topic/get-status-0-on-local-page-load-with-cordova
https://groups.google.com/forum/#!topic/phonegap/1e0z4I9Ps68
https://stackoverflow.com/questions/26556604/phonegap-ios-http-request-always-return-status-0-on-local-files
https://stackoverflow.com/questions/17964383/phonegap-ajax-call-fails-everytime/19498463#19498463

See how jQuery deals with it:
https://github.com/jquery/jquery/blob/master/src/ajax/xhr.js#L17

In fact, different browsers behave differently. For instance, in Android you'll get a 200 OK code. In general, the HTTP response status code is meaningless if there was no HTTP transaction at all. So, this PR will ignore status code when the `file://` protocol is used. Just check for content.

mourner:
@oscarfonts sorry for a delay — maybe let's just accept `xhr.status === 0` in `onload` without checking the URL? If this works for jQuery, it should probably be safe for GL JS..

> Before the request is complete, the value of status will be 0 . It is worth noting that browsers report a status of 0 in case of XMLHttpRequest errors too.

The [docs](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status) suggest that `status` is `0` by default and only changes to an appropriate number when it gets a response (in `onload`), but if it doesn't, it implies that the XHR is "fake" and we can simply accept the response (since errors will already be handled in `onerror`).

oscarfonts:
@mourner ok, done

Issue #6612, opened 2018-05-04
Failing query test: circle-radius/feature-state 


anandthakker:
Due to a merge conflict involving https://github.com/mapbox/mapbox-gl-js/commit/de4648f1cf9e69d3663cf9df30b7f103913e5255

PR #6615, opened 2018-05-04
Add featureState parameter to evaluate methods

asheemmamoowala:
Addresses #6612.

Updates `StyleLayer#queryIntersectsFeature` to include feature state as a separate parameter.

### DIFF
--- a/src/style/properties.js
+++ b/src/style/properties.js
@@ -446,7 +446,7 @@ export class PossiblyEvaluatedPropertyValue<T> {
         }
     }
 
-    evaluate(feature: Feature, featureState?: FeatureState): T {
+    evaluate(feature: Feature, featureState: FeatureState): T {
         return this.property.evaluate(this.value, this.globals, feature, featureState);
     }
 }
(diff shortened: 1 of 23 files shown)


src/style/properties.js
```diff
@@ -445,8 +446,8 @@ export class PossiblyEvaluatedPropertyValue<T> {
         }
     }
 
-    evaluate(feature: Feature): T {
-        return this.property.evaluate(this.value, this.globals, feature);
+    evaluate(feature: Feature, featureState?: FeatureState): T {
```
jfirebaugh: Should `featureState` be a required parameter everywhere that `feature` is required?

asheemmamoowala: The `feature-state` expression is only usable in data-driven paint properties, so I restricted the change to that code path. The one place where this might be useful is  `Binder<T>#populatePaintArray`. But that method is run on the worker, and we don't have state available there, so it would also be unused.

jfirebaugh: The benefit of making it required is that it leans on the type system to enforce consideration of of whether state needs to be provided.

* We can be sure we've looked at all the existing code paths.
* Code paths where state isn't available make that explicit by passing an empty object.
* If we add another evaluation code path, we can be sure we don't forget state where it should be supplied.

PR #6627, opened 2018-05-07
Restore target property for mouse{enter,over,leave,out}

### LINKED ISSUE #6623: 0.45.0-beta.1 doesn't provide e.target for mouseenter/mouseleave
**mapbox-gl-js version**: 0.45.0-beta.1
**browser**: Chrome 66 / macOS

### Steps to Trigger Behavior
 1. use e.target in a mouseenter callback

### Link to Demonstration
https://codesandbox.io/s/7wyv3pwkyj

### Expected Behavior
I didn't find this breaking change in 0.45 release notes, so it might be a bug.

jfirebaugh:
Fixes #6623.

### DIFF
--- a/src/ui/events.js
+++ b/src/ui/events.js
@@ -33,7 +33,7 @@ export class MapMouseEvent extends Event {
     /**
      * The `Map` object that fired the event.
      */
-    map: Map;
+    target: Map;
 
     /**
      * The DOM event which caused the map event.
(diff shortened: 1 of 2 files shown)


src/ui/events.js
```diff
@@ -33,7 +33,7 @@ export class MapMouseEvent extends Event {
     /**
      * The `Map` object that fired the event.
      */
-    map: Map;
+    target: Map;
```
jfirebaugh: It doesn't look like map mouse events have ever actually had a `map` property.

Issue #5641, opened 2017-11-09
Icon collision boxes don't account for `icon-text-fit`


ChrisLoer:
See discussion at https://github.com/mapbox/mapbox-gl-js/pull/5632/#discussion_r149846100.

`CollisionFeature` takes a `Shaping` as an input for determining the size of collision boxes. This works fine for text, but for icons with `icon-text-fit: true`, the quads generated for the icon by `getIconQuads` may not match the `Shaping`.

I suspect the right fix here is to move most of the `icon-text-fit` logic into `shapeIcon`.

My guess is this hasn't been a big deal in practice because when people are using `icon-text-fit` they also use `icon-allow-overlap` and `icon-ignore-placement`, and rely on the matching text's collision box. But it'd be better not to have that gotcha.

/cc @jfirebaugh @ansis @anandthakker

kkaefer:
This will be fixed by https://github.com/mapbox/mapbox-gl-js/pull/8741

Issue #4861, opened 2017-06-21
Collision boxes don't account for symbol-placement:point/rotation-alignment: map


ChrisLoer:
As far as I can tell this has always been the case:

<img width="490" alt="screenshot 2017-06-21 13 30 56" src="https://user-images.githubusercontent.com/375121/27405308-e65e5a5a-5685-11e7-80f8-9216eb11f450.png">

Labels with `rotation-alignment: map` are usually more likely to use `symbol-placement: line`, and they're mostly fine because collision detection is done with a series of small boxes (icons with lopsided aspect ratios might be an exception here).

This would be straightforward to fix if we moved to an approach of global collision detection using current camera parameters (discussed somewhat in https://github.com/mapbox/mapbox-gl-js/issues/4704).

/cc @ansis

ChrisLoer:
> This would be straightforward to fix if we moved to an approach of global collision detection using current camera parameters (discussed somewhat in #4704).

It turns out not to be totally straightforward to fix with global collision detection because we still don't have an easy/fast way to represent rectangular geometries that aren't viewport-axis-aligned.

For text line labels, we've started representing their geometry with a series of circles (which have the advantage of keeping the same shape during rotation operations) -- it's possible we could solve this problem by breaking map-aligned rectangles (whether for text or icons) into a series of circles. However, we're not tackling that problem yet in the global collision PR #5150).

/cc @nickidlugash

ChrisLoer:
When we fix this, we should also switch icons with `symbol-placement: line` to use the same behavior (with rotation alignment based on the tangent to the line). We attempted to do that in the past with https://github.com/mapbox/mapbox-gl-js/pull/1981, but removed that behavior as part of #5150.

smerchek:
I just ran into this issue. We use a symbol rotated in this way as a hover/click target and it's very disorienting for users once the map is rotated, because they can only click on a small part of the symbol (since the rest of the click target is below the symbol). 

Curious if there are any workarounds or fix in sight? 

I thought maybe I could simulate a point placement but with a small LineString since symbol-placement=line works with map rotation, but it doesn't really give me the precision I need (and there is also a bug preventing this workaround since the symbol will easily collide https://github.com/mapbox/mapbox-gl-js/issues/9715).

Issue #4798, opened 2017-06-07
Respect text-offset in symbol collision handling


anandthakker:
We currently do not respect the value of `text-offset` when determining whether two symbol instances collide. 

Example -- in this style:

```js
{
  "version": 8,
  "metadata": {
    "test": {
      "height": 64,
      "collisionDebug": true
    }
  },
  "center": [ 0, 0 ],
  "zoom": 0,
  "sources": {
    "geojson": {
      "type": "geojson",
      "data": {
        "type": "Feature",
        "properties": {},
        "geometry": {
          "type": "LineString",
          "coordinates": [
            [-20, -20],
            [20, -20],
            [20, 20],
            [-20, 20]
          ]
        }
      }
    }
  },
  "glyphs": "mapbox://fonts/mapbox/{fontstack}/{range}.pbf",
  "layers": [
    {
      "id": "guid",
      "type": "line",
      "source": "geojson"
    },
    {
      "id": "text",
      "type": "symbol",
      "source": "geojson",
      "layout": {
          "symbol-placement": "line",
          "symbol-spacing": 115,
          "text-size": 10,
          "text-allow-overlap": false,
          "text-ignore-placement": false,
          "text-field": "A",
          "text-font": [
              "Open Sans Semibold",
              "Arial Unicode MS Bold"
          ],
          "text-keep-upright": false,
          "text-offset": [0, -3]
      }
    }
  ]
}
```

We'd expect only one of the 'A' instances to be rendered, but both are:

<img width="100" alt="screen shot 2017-06-07 at 8 28 18 am" src="https://user-images.githubusercontent.com/4974151/26878400-4da5ca3a-4b5b-11e7-95d3-853861e9de31.png">

anandthakker:
The above JSON should work verbatim as an integration test -- I'm holding off on adding it because I'm not sure whether it's well defined which of those two labels should be shown.

cc @ansis @ChrisLoer

ChrisLoer:
It's not totally ignored: it feeds into the "top" and "bottom" values of the shaped text, which feeds into `CollisionFeature` as the `shaped` parameter. For point labels, those values are fed into the single collision box, but they're essentially ignored for line labels (they feed into "height" and "width" but the offset won't make a difference).

Empirically, it's still partially broken for point labels:

![screenshot 2017-06-07 09 09 31](https://user-images.githubusercontent.com/375121/26888825-0a86dce8-4b61-11e7-8cca-a4c50849af45.png)

[...]

PR #6631, opened 2018-05-07
Handle *-rotate for point-placed symbols.

ChrisLoer:
Fixes issue #6075 -- or at least fixes what I think is the chief complaint remaining there: that the combination of `*-offset` and `*-rotate` doesn't work for point labels.

This PR does not fix:
 - Collision detection for point symbols with `*-rotation-alignment: map` (#4861)
 - Collision detection or for line-placed symbols with `*-offset` (and/or `*-rotate`): (#4798)
 - Combination with `icon-text-fit`: (#5641)

## Launch Checklist

 - [ ] post benchmark scores

/cc @anandthakker @ansis


ChrisLoer:
One of the render tests kept failing because the collision box would draw slightly differently on the CI machine (I think this is just built into the variation in `gl.LINES` implementations). The part that was failing was not actually the thing under test (it was just a reference to show where the origin of the offset was), so I just removed it.

An alternative solution would be to write the tests as query tests, but I prefer render tests because they're more visual/simpler to understand.

anandthakker:
Nice un-overcomplication of the problem 😄

PR #6636, opened 2018-05-08
don't stop animation on map resize

### LINKED ISSUE #4041: Flying ends on window resize and browser zoom change
**0.31.0**:

### Steps to Trigger Behavior

 1. start a flying event
 2. resize the window or change the browser zoom level

### Expected Behavior
I expect that the flying event would continue as normal

### Actual Behavior
The flying event ends mid-flight

Flight end on window resize: https://gyazo.com/6fc4e12952b42a6ac1ffe5e790bbb39e

### LINKED ISSUE #4481: Geocoder control's flyTo animation on result selection gets interrupted on Android devices
I have copied and pasted the sample code from the [Mapbox geocoder example](https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/) into a [test example](https://anna.ps/sandbox/test.html), with no changes or additions. 

In the test example, in Chrome on Android, selecting an item from the dropdown does not zoom or pan the map. 

The issue does not occur on the desktop browsers I've tried, so I assume it is related to touch events. I've tried running the Chrome remote debugger, and I cannot see any console warnings or errors in the mobile browser. Switching between HTTP and HTTPS also doesn't seem to make a difference. 

This may be related to the closed issue #4277 or the open issue #3720, but I feel this probably needs its own issue (or at least to reprioritise #3720), since rather than a feature request it is an actual user-facing bug. On my current project several users have reported it.

Most odd is that the [Mapbox geocoder example](https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/) itself seems to work fine in Chrome/Android.

andrewharvey:
## Launch Checklist
closes #4041 
closes #4481 

allows animations from flyTo to continue even when the map resizes. You can end up with the point you're flyingTo no longer being inside the map window if the map is resized a lot, but I feel it's no worse than the current behaviour, something than can be fixed in a future PR, and addresses the root cause of mapbox-gl-geocoder not working on Android.

 - [ ] check new unit test fails on master without this change

 - n/a post benchmark scores
 - n/a tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec changes

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -449,11 +449,12 @@ class Map extends Camera {
         this.transform.resize(width, height);
         this.painter.resize(width, height);
 
-        return this
-            .fire(new Event('movestart', eventData))
+        this.fire(new Event('movestart', eventData))
             .fire(new Event('move', eventData))
             .fire(new Event('resize', eventData))
             .fire(new Event('moveend', eventData));
+
+        return this;
     }
 
     /**


andrewharvey:
I'm not too familiar with the animation/interaction code, so not sure if there are any implications of this change I've overlooked.

mollymerp:
hmm yeah – it does temporarily bork the transform state but it does seem to me that this behavior is preferable considering the drawbacks of the current approach. @mourner @ansis are there undesirable side-effects of this that I'm missing?

src/ui/bind_handlers.js
```diff
@@ -177,6 +177,10 @@ export default function bindHandlers(map: Map, options: {interactive: boolean})
     }
 
     function onWheel(e: WheelEvent) {
+        if (options.interactive) {
+            map.stop();
+        }
+
```
ansis: Is this related to the resize fix or something separate?
No need to split it out if it's separate, I'm just wondering if I'm missing a relation between this and the resize bug.

ansis:
@mollymerp nah, I think this is fine. If it doesn't finish in the right location we should try to fix that but this seems fine to me.

src/ui/bind_handlers.js
andrewharvey: I think I needed this so that scrolling on the map cancelled an active flyTo.

andrewharvey:
test-flow is failing with:

```
Cannot call `this.resize()._update` because property `_update` is missing in `Evented` [1].

   src/ui/map.js:1738:13
   1738|             this.resize()._update();
                     ^^^^^^^^^^^^^^^^^^^^^^^

References:
   src/util/evented.js:47:14
     47| export class Evented {
                      ^^^^^^^ [1]
```

:thinking:

mollymerp:
@andrewharvey I think flow was confused by returning `this.fire....` from `map.resize()` 🤷‍♀️

jdaudier:
Any idea when this fix will be released?

allthesignals:
This is a little tangential, but what is the behavior _called_ when MapboxGL (or WebGL) pauses animation (or rendering) when the window/tab isn't "active" (isn't visible in the user's desktop)? Having an automation issue, and trying to find the correct language to understand...

ryanhamley:
AFAIK, we (and WebGL) do not pause rendering on background tabs. That is the default behavior of many modern browsers for performance reasons.

allthesignals:
@ryanhamley thank you! Makes sense that it's way lower level than a MapboxGL thing.

PR #6644, opened 2018-05-09
Backport: Add `worker-src blob:` to CSP directives

### LINKED ISSUE #6626: Update documentation on necessary CSP directives
Hi, thanks for the great, feature rich map you provide.

I found one issue with current documentation available on:

https://www.mapbox.com/mapbox-gl-js/api/

The `child-src` directive is deprecated, it should address it as `worker-src`, to avoid collision with `frame-src`.

Also, `image-src` is lacking the `*.tiles.mapbox.com`. `blob:` is not enough in itself with current version.

**browser**: Chrome 66.0.3359.139 (Official Build) (64-bit)

jfirebaugh:
Backports #6640
----
Fixes #6626.


jfirebaugh:
Hey, it works!

PR #6670, opened 2018-05-15
PopupOptions are optional

TimPetricola:
The flow annotation for `PopupOptions` makes them mandatory at the moment. 

In fact they are all optional.


TimPetricola:
@jfirebaugh oh sorry about that, it is now fixed :)

PR #6684, opened 2018-05-16
Align match behavior with case/==

### LINKED ISSUE #6680: Fallback value in "match" expression is not applied to features with missing property
Hi,

In Mapbox Studio, I am using data conditions to style features based on a property value. (in my case: lots based on their land use category).

Here is a sample use case:
I set a yellow color for all lots that fall within the "01" land use category. All lots that belong to another category are colored using the fallback color. However, unlike before, the fallback color is only applied to features that actually have the property. So lots with missing LandUse property are being colored in black:
image

By default, Mapbox Studio uses a "match" expression:
image

---

On the other hand, if I use a "case" expression,  the "fallback" value is assigned to all features that don't match the "01" condition (even if the property is missing). Here is how it looks:

image
when a "case" expression is used:
image

Personally I expect to see the fallback color assigned to all lots (like in the second example - when using the "case" expression). This is how data-driven styling used to work: there was a "default" value which was assigned to all features no matter if they had the property or not.

I wonder if this is the expected behavior of the "match" expression ? Is there a way to assign a color value to lots with a missing property through the UI (without replacing the match expression with a case)?

Best regards,
Kiro

jfirebaugh:
Makes `["match", ["get", k], label, match, otherwise]` equivalent to `["case", ["==", ["get", k], label], match, otherwise]`. This changes the behavior of match expressions where the runtime type of the input does not match the type of the labels: previously such expressions produced a runtime type error and then fell back to the property default value; now they produce the fallback value from the match expression.

Fixes #6680.

cc @mapbox/studio @mapbox/maps-design @mapbox/ios @mapbox/android 

## Launch Checklist

### DIFF
--- a/src/style-spec/expression/definitions/match.js
+++ b/src/style-spec/expression/definitions/match.js
@@ -3,11 +3,11 @@
 import assert from 'assert';
 
 import { typeOf } from '../values';
+import { ValueType, type Type } from '../types';
 
 import type { Expression } from '../expression';
 import type ParsingContext from '../parsing_context';
 import type EvaluationContext from '../evaluation_context';
-import type { Type } from '../types';
 
 // Map input label values to output expression index
 type Cases = {[number | string]: number};
@@ -84,19 +84,25 @@ class Match implements Expression {
             outputs.push(result);
         }
 
-        const input = context.parse(args[1], 1, inputType);
+        const input = context.parse(args[1], 1, ValueType);
         if (!input) return null;
 
         const otherwise = context.parse(args[args.length - 1], args.length - 1, outputType);
         if (!otherwise) return null;
 
         assert(inputType && outputType);
+
+        if (input.type.kind !== 'value' && context.concat(1).checkSubtype((inputType: any), input.type)) {
+            return null;
+        }
+
         return new Match((inputType: any), (outputType: any), input, cases, outputs, otherwise);
     }
 
     evaluate(ctx: EvaluationContext) {
         const input = (this.input.evaluate(ctx): any);
-        return (this.outputs[this.cases[input]] || this.otherwise).evaluate(ctx);
+        const output = (typeOf(input) === this.inputType && this.outputs[this.cases[input]]) || this.otherwise;
+        return output.evaluate(ctx);
     }
 
     eachChild(fn: (Expression) => void) {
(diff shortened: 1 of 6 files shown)


jfirebaugh:
It doesn't look like our [documentation](https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-match) was specific enough to cover this nuance, so it technically doesn't require an update. Does anyone think we should make it more specific?

> Selects the output whose label value matches the input value, or the fallback value if no match is found. The input can be any string or number expression (e.g. `["get", "building_type"]`). Each label can either be a single literal value or an array of values.

1ec5:
Yes, I think we should make the documentation more specific, assuming we don’t intend to revisit this decision anytime soon. Without additional documentation, I think some developers may still be initially surprised that `["match", ["get", k], label, match, otherwise]` resolves to `otherwise` instead of `match`. 

I agree with this change, but the previous error message at least gave the developer a clue as to where the problem might lay. Without either the error message or documentation, the developer is left to guess that `match` fell through to the else case.

anandthakker:
Agreed re: clarifying the docs.  How about:

> Selects the output whose label value matches the input value, or the fallback value if no match is found. The input can be any expression (e.g. `["get", "building_type"]`).  Each label must either be a single literal value or an array of literal values (e.g. `'a'`, `['c', 'b']`), and those values must be all strings or all numbers. (The values `'1'` and `1` cannot both be labels in the same match expression).

src/style-spec/expression/definitions/match.js
```diff
@@ -84,19 +84,25 @@ class Match implements Expression {
...
+            return null;
+        }
+
         return new Match((inputType: any), (outputType: any), input, cases, outputs, otherwise);
     }
 
     evaluate(ctx: EvaluationContext) {
         const input = (this.input.evaluate(ctx): any);
-        return (this.outputs[this.cases[input]] || this.otherwise).evaluate(ctx);
+        const output = (typeOf(input) === this.inputType && this.outputs[this.cases[input]]) || this.otherwise;
```
anandthakker: Too bad we can't just use `Map`

jfirebaugh:
@anandthakker Want to review the conversion addition here?

PR #6687, opened 2018-05-17
Feature/draggable markers

### LINKED ISSUE #3087: Add "draggable" property to "Marker" class
**mapbox-gl-js version**:Master
### Steps to Trigger Behavior

It would be really nice if there were a draggable property on ui/marker.js. Would make it easy porting over web apps reliant on G Maps. Adding a geojson source and using setData every time a move event is emitted  for modeling dragging is unacceptably  slow. It looks like Geojsonsources aren't able to handle data that changes as frequently as the mouse move event.

Markers that are regular html divs would likely be more performant (for a small number of them) since they can just translate the position themselves upon dragging.

Here is the specific property I am referring to.

https://developers.google.com/maps/documentation/javascript/markers#draggable

### LINKED ISSUE #4597: `on` should be idempotent
Like [`addEventListener`](https://dom.spec.whatwg.org/#dom-eventtarget-addeventlistener), `Evented#on` (and `Map#on` with delegation) should be idempotent, and only do something if there wasn't already a listener for the given event type and callback.

For example, the sequence `on('a', cb)`, `on('a', cb)`, `off('a', cb)` should leave no `cb`'s listening for `a`.

### LINKED ISSUE #6102: change "drag a point" example to use Marker
Performance isn't great on https://www.mapbox.com/mapbox-gl-js/example/drag-a-point/ and this is a pain point for users (https://github.com/mapbox/mapbox-gl-js/issues/2334#issuecomment-352612642). Using `Marker` would drastically simplify the code in the example and have better performance because we wouldn't be using `setData` to change the location.

ryanhamley:
## Launch Checklist

- Adds drag functionality to markers either via a `draggable` option or a `setDraggable` method
- Adds `dragstart`, `drag` and `dragend` events on Marker
- Updates JSDocs
- Adds new drag-a-marker example (leaves drag-a-point example as is)
- Adds tests for dragging a marker
- Makes `evented._addEventListener` idempotent to avoid adding the same listener multiple times

Closes #6102 
Closes #3087 
Closes #4597

### DIFF
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -7,35 +7,36 @@ import Point from '@mapbox/point-geometry';
 import smartWrap from '../util/smart_wrap';
 import { bindAll, extend } from '../util/util';
 import { type Anchor, anchorTranslate, applyAnchorClass } from './anchor';
-
+import { Event, Evented } from '../util/evented';
 import type Map from './map';
 import type Popup from './popup';
 import type {LngLatLike} from "../geo/lng_lat";
-import type {MapMouseEvent} from './events';
+import type {MapMouseEvent, MapTouchEvent} from './events';
 
 type Options = {
     element?: HTMLElement,
     offset?: PointLike,
     anchor?: Anchor,
-    color?: string
+    color?: string,
+    draggable?: boolean
 };
 
 /**
  * Creates a marker component
- * @param options
- * @param options.element DOM element to use as a marker. The default is a light blue, droplet-shaped SVG marker.
- * @param options.anchor A string indicating the part of the Marker that should be positioned closest to the coordinate set via {@link Marker#setLngLat}.
+ * @param {Object} [options]
+ * @param {HTMLElement} [options.element] DOM element to use as a marker. The default is a light blue, droplet-shaped SVG marker.
+ * @param {string} [options.anchor='center'] A string indicating the part of the Marker that should be positioned closest to the coordinate set via {@link Marker#setLngLat}.
  *   Options are `'center'`, `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`, `'top-right'`, `'bottom-left'`, and `'bottom-right'`.
- *   The default is `'center'`.
- * @param options.offset The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
- * @param options.color The color to use for the default marker if options.element is not provided. The default is light blue.
+ * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
+ * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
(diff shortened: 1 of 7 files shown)


ryanhamley:
I am unsure about the final two tests in `marker.test.js` and would love some feedback on them. The idea was to have tests where we would create a marker, simulate a mouse/touch drag through the combination of `mousedown`, `mousemove` and `mouseup` events then compare the start and end positions to see that the marker was actually moved. The tests pass, but I'm not convinced that `Marker with draggable:true moves to new position in response to a mouse-triggered drag` is doing what I think it's doing. Setting `clientX` and `clientY` when creating a new `mousemove` event works, but is it actually moving the mouse around? AFAIK, it's not possible to move the mouse with Javascript due to a variety of security concerns. Is there a better way to write this test? Should it be included at all? Is this type of interaction really better suited for an integration test (which we don't have) than a unit test?

Note that I also deleted the equivalent TouchEvent test because TouchEvents don't have `clientX` and `clientY` properties or anything like them to set when creating the event.

src/ui/marker.js
```diff
@@ -346,4 +362,116 @@ export default class Marker {
...
+     * @returns {Marker} `this`
+     */
+    setDraggable(shouldBeDraggable: boolean) {
+        this._draggable = !!shouldBeDraggable; // convert possible undefined value to false
+
+        // handle case where map may not exist yet
+        // e.g. when setDraggable is called before addTo
+        if (this._map) {
+            if (shouldBeDraggable) {
+                this._map.on('mousedown', this._addDragHandler);
```
jfirebaugh: We'll need to either fix #4597 or check that `this._draggable` state is actually changing, so as not to register duplicate handlers if `setDraggable(true)` is called when the marker is already draggable.

src/ui/marker.js
```diff
@@ -346,4 +362,116 @@ export default class Marker {
...
+                this._map.off('touchstart', this._addDragHandler);
+            }
+        }
+
+        return this;
+    }
+
+    /**
+     * Returns the marker's draggable property
+     * @returns {boolean} `this._draggable`
```
jfirebaugh: Don't reference the private property here.

src/ui/marker.js
ryanhamley: a5da36c addresses #4597 which would resolve this comment. Is that an acceptable solution? It definitely works.

src/util/evented.js
```diff
@@ -6,6 +6,7 @@ type Listener = (Object) => any;
 type Listeners = { [string]: Array<Listener> };
 
 function _addEventListener(type: string, listener: Listener, listenerList: Listeners) {
+    _removeEventListener(type, listener, listenerList);
```
jfirebaugh: Rather than remove and re-add, make this an affirmative check for the current presence of listener, so that a redundant `on` doesn't change the relative order of listeners. E.g. given `map.on('click', listenerA); map.on('click', listenerB); map.on('click', listenerA);`, `listenerA` should be the first one called.

Issue #6689, opened 2018-05-17
Better documentation for global transition option


lbud:
Documentation for the [global `transition` property](https://www.mapbox.com/mapbox-gl-js/style-spec#root-transition) (as well as the [general `transition` type](https://www.mapbox.com/mapbox-gl-js/style-spec#transition)) is pretty thin and could be clarified.

Related: https://github.com/mapbox/mapbox-gl-js/issues/4081.

Ref: https://github.com/mapbox/mapbox-gl-js/issues/6519 (closed in favor of this ticket).

Issue #6519, opened 2018-04-16
Transition root property seems to have no effect for first rendering of the map


pathmapper:
**mapbox-gl-js version**: v0.44.2

The transition root property seems to have no effect (or I'm misunderstanding the concept of transition).

### Steps to Trigger Behavior

1. Set a [transition root property](https://www.mapbox.com/mapbox-gl-js/style-spec/#root-transition) for a style and use this style for a map.
2. Change the values for `duration` and `delay` of the transition root property significantly and compare with the behavior before the change.

Example Style: [style_5000.zip](https://github.com/mapbox/mapbox-gl-js/files/1916334/style_5000.zip)

### Expected Behavior

The transition behavior changes according to the chosen values for `duration` and `delay`.

### Actual Behavior

Transition behavior stays the same, regardless which values are chosen for `duration` and `delay`.

This [jsbin example](http://jsbin.com/gekoqaxadu/1/edit?html,output) uses the example style  [(style_5000.zip)](https://github.com/mapbox/mapbox-gl-js/files/1916334/style_5000.zip) with the transition root property

```
"transition": {
  "duration": 5000,
  "delay": 5000
}
```
and the symbol layer itself has no transition settings, but the transition behavior of the text labels looks like the [default](https://www.mapbox.com/mapbox-gl-js/style-spec/#transition):
```
"transition": {
  "duration": 300,
  "delay": 0
}
```

pathmapper:
Updated demonstrations (v0.45.0 and style included):

Transition root property:
http://jsbin.com/fanihuqeni/edit?html,output

Transition root property + map.setPaintProperty
http://jsbin.com/bumajivoxa/edit?html,output

maxwang7:
I am working on something similar where I would like to be able to change the text shading when you hover over an icon. Currently the text fades in and out which is visually jarring, so would ideally be able to turn off those transitions. Thanks!

lbud:
Hm. There are two things at play here @pathmapper:
[...]

PR #6691, opened 2018-05-17
Clarify transition type + global transition

lbud:
This adds a bit more description around what a `transition` is, how the global `transition` works, and notes the possibility of property-specific transitions.

`#root-transition`:
image

`#transition`:
image

Closes https://github.com/mapbox/mapbox-gl-js/issues/6689.
Ref https://github.com/mapbox/mapbox-gl-js/issues/4081, https://github.com/mapbox/mapbox-gl-js/issues/6519.

### DIFF
--- a/docs/pages/style-spec.js
+++ b/docs/pages/style-spec.js
@@ -860,7 +860,11 @@ export default class extends React.Component {
                             <a id='transition' className='anchor'></a>
                             <h2><a href='#transition' title='link to transition'>Transition</a></h2>
                             <p>
-                                A style's <code>transition</code> property provides global transition defaults for that style.
+                                A <code>transition</code> property controls timing for the interpolation between a transitionable style
+                                property's previous value and new value. A style's <a href='#root-transition' title='link to root-transition'>
+                                root <code>transition</code></a> property provides global transition defaults for that style. Any transitionable
+                                style property may also have its own <code>-transition</code> property that defines specific transition timing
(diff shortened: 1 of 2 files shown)


docs/pages/style-spec.js
```diff
@@ -966,7 +966,11 @@ export default class extends React.Component {
                             <a id='transition' className='anchor'></a>
                             <h2><a href='#transition' title='link to transition'>Transition</a></h2>
                             <p>
-                                A style's <code>transition</code> property provides global transition defaults for that style.
+                                A <code>transition</code> property controls timing for the interpolation between a transitionable style
```
mollymerp: It might be nice to include same styling for `transitionable` that we do in the rest of the docs here 🤔
```html
<em className='quiet'><span className='icon opacity inline quiet' />transitionable </em>
```

lbud: Hm, I think 
_transitionable_
denotes that a property can visually/functionally transition between values, whereas
`transition`
is a literal property key that dictates how to perform a transition, so I'm inclined to keep this as `<code>transition</code>`

lbud:
Thanks for the suggestion @pathmapper — I'm inclined to omit mention of the `fadeDuration` option from the style spec as it is at this point only a JS implementation-specific option rather than being mandated by all implementers of the spec (i.e. not present in mapbox-gl-native), but I've added a hint that the style's `transition` property has no effect on collision-based symbol fading.

PR #6697, opened 2018-05-18
don't fire click event if default is prevented on mousedown for a drag event

### LINKED ISSUE #6642: v0.45.0 breaks mapbox-gl-directions plugin/example
**mapbox-gl-js version**: v0.45.0

**browser**: Chrome

### Steps to Trigger Behavior

 1. Open https://jsbin.com/gipahagera/edit?html,output
 2. Click somewhere on the route between A and B and drag the route

auswahl_277

### Expected Behavior

Destination (B) is not moved and the route is adapted as with v0.44.2
https://jsbin.com/zidemesofe/edit?html,output

auswahl_276

### Actual Behavior

Destination (B) is moved

image

Also reproducible with:
https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-directions/

mollymerp:
fix #6642

in #6218 we stopped checking if the mouse position had changed before firing a click event. the `DragPanHandler` suppresses its associated `click` event on `mouseup`

https://github.com/mapbox/mapbox-gl-js/blob/a5a8dfe5b90737dbb295eaca7f41c667ae4060a8/src/ui/handler/drag_pan.js#L191

but if `preventDefault()` is called on `mousedown` the click was no longer being suppressed after a drag which caused the change in behavior in #6642 

I added a `drag` simulation that allows you to specify different options for mousedown/mouseup but I'm open to other ideas on how to test this. 

## Launch Checklist

### DIFF
--- a/src/ui/bind_handlers.js
+++ b/src/ui/bind_handlers.js
@@ -25,6 +25,7 @@ export default function bindHandlers(map: Map, options: {interactive: boolean}) 
     const el = map.getCanvasContainer();
     let contextMenuEvent = null;
     let mouseDown = false;
+    let startPos = null;
 
     for (const name in handlers) {
         (map: any)[name] = new handlers[name](map, options);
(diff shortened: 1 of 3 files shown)


mollymerp:
yeah I agree its an odd condition to have had in the first place 🤷‍♀️

PR #6703, opened 2018-05-21
DragRotateHandler should register mouseDown when used with non-map elements 

### LINKED ISSUE #6650: Rotate and pitch not possible with navigationControl compass
**mapbox-gl-js version**: v0.45.0

**browser**: Chrome on Linux

### Steps to Trigger Behavior

 1. Try to use the navigationControl compass to rotate and pitch the map

### Link to Demonstration

https://www.mapbox.com/mapbox-gl-js/example/navigation/

### Expected Behavior

Rotate and pitch is possible with the navigationControl compass as with v0.44.2:

https://jsbin.com/juzupaneju/edit?html,output

### Actual Behavior

Rotate and pitch is not possible with navigationControl compass.

Similar issue in the past:
https://github.com/mapbox/mapbox-gl-js/issues/5711 with fix https://github.com/mapbox/mapbox-gl-js/pull/5725

asheemmamoowala:
Fixes #6650 

This regressed as a result of  #6218. When moving the mouse down event registration from the drag handlers to `Map`, the `NavigationControl` button lost the mouse down handling inherited from `DragRotateHandler`

cc @anandthakker @jfirebaugh

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -69,7 +69,7 @@ class NavigationControl {
             this._map.on('rotate', this._rotateCompassArrow);
             this._rotateCompassArrow();
             this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass});
-            DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown.bind(this._handler));
+            DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
             this._handler.enable();
         }
         return this._container;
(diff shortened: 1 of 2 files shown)


anandthakker:
What if instead we made this change in `NavigationControl`?  I.e., when setting up the compass, add a mousedown listener on `this._compass` like:

```js
DOM.addEventListener(this._compass, 'mousedown', e => this._handler.onMouseDown);
```

This essentially replicates what `bind_handlers.js` does for `Map`, which feels right to me: `Map` takes responsibility for dispatching events to its handlers, and so now `NavigationControl` should take on the analogous responsibility for dispatching events to its single DragRotateHandler.

src/ui/control/navigation_control.js
```diff
@@ -78,6 +79,7 @@ class NavigationControl {
         DOM.remove(this._container);
         if (this.options.showCompass) {
             this._map.off('rotate', this._rotateCompassArrow);
+            DOM.removeEventListener(this._compass, 'mousedown', this._handler.onMouseDown.bind(this._handler));
```
mourner: This won't actually remove anything because `bind` returns a new function, and the newly crested listener isn't among the added ones for the event. You should have `onMouseDown` already bound in the handler instead.

PR #6704, opened 2018-05-21
Add unit tests to ensure docs list all available expressions

### LINKED ISSUE #5438: Add unit test checking that each expression type is documented

ryanhamley:
## Launch Checklist

    - Adds unit tests to ensure that all expression types are documented

Closes #5438

### DIFF
--- a/test/unit/style-spec/expression.test.js
+++ b/test/unit/style-spec/expression.test.js
@@ -1,5 +1,26 @@
 import { test } from 'mapbox-gl-js-test';
 import { createPropertyExpression } from '../../../src/style-spec/expression';
+import definitions from '../../../src/style-spec/expression/definitions';
+import v8 from '../../../src/style-spec/reference/v8';
+import {expressions as definitionMetadata} from '../../../docs/components/expression-metadata';
+
+// filter out interal "error" and "filter-*" expressions from definition list
+const filterExpressionRegex = /filter-/;
+const definitionList = Object.keys(definitions).filter((expression) => {
+    return expression !== 'error' && !filterExpressionRegex.exec(expression);
+}).sort();
+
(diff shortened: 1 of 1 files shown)


test/unit/style-spec/expression.test.js
```diff
@@ -1,5 +1,48 @@
...
+
+    // filter out interal "error" and "filter-*" expressions from definition list
+    const filterExpressionRegex = /filter-/;
+    const definitionList = Object.keys(definitions).filter((expression) => {
+        return expression !== 'error' && !filterExpressionRegex.exec(expression);
+    });
+
+    test('v8.json includes all definitions from style-spec', (t) => {
+        const v8List = Object.keys(v8.expression_name.values);
+        t.ok(equals(definitionList, v8List));
```
anandthakker: minor nit: this can be `t.deepEquals(definitionList.sort(), v8List.sort())` and thus avoid the need to implement `equals()` above

test/unit/style-spec/expression.test.js
```diff
@@ -1,5 +1,48 @@
...
+        return true;
+    };
+
+    // filter out interal "error" and "filter-*" expressions from definition list
+    const filterExpressionRegex = /filter-/;
+    const definitionList = Object.keys(definitions).filter((expression) => {
+        return expression !== 'error' && !filterExpressionRegex.exec(expression);
+    });
+
+    test('v8.json includes all definitions from style-spec', (t) => {
```
anandthakker: I think this (and the one below) should be `t.test(...)` if we want each of these to be a "subtest".  It's also probably fine to just drop the subtest wrapper and do both of these inline.

PR #6720, opened 2018-05-23
Avoid TypedArray#slice (not available in IE11)

### LINKED ISSUE #6718: Official example 'add-image-generated' broken in IE since 0.43+
**mapbox-gl-js version**: 0.43+
**browser**: IE 11 on Win 10

### Steps to Trigger Behavior
 1. Open add-image-generated example

### Link to Demonstration
https://www.mapbox.com/mapbox-gl-js/example/add-image-generated/

0.42.0
https://jsbin.com/wewejiy/3/edit?html,js,output

### Expected Behavior
Should show icon

### Actual Behavior
Doesn't show icon

jfirebaugh:
Fixes #6718.

See #5858 for a previous instance of this issue. In the past I've looked into whether there are any static analysis tools that could catch browser compatibility issues (e.g. https://github.com/amilajack/eslint-plugin-compat), but not found anything sophisticated enough to be able to catch issues like this.

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1147,7 +1147,7 @@ class Map extends Camera {
                 'or object with `width`, `height`, and `data` properties with the same format as `ImageData`')));
         } else {
             const {width, height, data} = image;
-            this.style.addImage(id, { data: new RGBAImage({width, height}, data.slice(0)), pixelRatio, sdf });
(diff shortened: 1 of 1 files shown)


anandthakker:
> In the past I've looked into whether there are any static analysis tools that could catch browser compatibility issues (e.g. https://github.com/amilajack/eslint-plugin-compat), but not found anything sophisticated enough to be able to catch issues like this

Could we stub `*Array.slice` to throw an error before running unit/integration tests?

PR #6721, opened 2018-05-23
Fallback resource timing in web workers

sbma44:
This PR adds a fallback method to `collectResourceTiming: true` for browsers that don't implement the `performance.getEntries()` method properly in web workers. It provides as much data as can be collected now, and should return full data when these browsers correct their implementations.

The availability of the `performance` API in web workers remains unpleasantly murky. Despite tickets like https://bugzilla.mozilla.org/show_bug.cgi?id=1425458 and [badges declaring web worker availability](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntries) bugs remain in major browsers. Specifically, it appears that in both Firefox and Safari `performance.getEntries`, `performance.getEntriesByName` and `performance.getEntriesByType` return empty arrays in web workers regardless of whether the worker has made network requests. 

`performance.mark`, `performance.measure` and subsequent use of `performance.getEntries*` _do_ work, however. This is an inferior solution, as `PerformanceMeasure` objects compare unfavorably to `PerformanceResourceTiming` objects:
- no DNS, TLS or connection timing data
- no division between request and response
- no data on payload size

Still, the basic timing data that can be collected is better than nothing.

This change does not alter any codepaths when `collectResourceTiming` is false. When it is true, a check on array length and an invocation of `performance.mark()` are added to the status quo codepath. Obviously there's a bit more code for the fallback case, but it's pretty slight.

## Launch Checklist

 - [ ] ~~document any changes to public APIs~~ n/a
 - [ ] post benchmark scores
 - [x] ~~manually test the debug page~~ tested on my instrumentile integration test page instead (GL JS debug page doesn't collect resource timing)

### DIFF
--- a/src/source/geojson_worker_source.js
+++ b/src/source/geojson_worker_source.js
@@ -2,7 +2,7 @@
 
 import { getJSON } from '../util/ajax';
 
-import perf from '../util/performance';
+import performance from '../util/performance';
 import rewind from 'geojson-rewind';
 import GeoJSONWrapper from './geojson_wrapper';
 import vtpbf from 'vt-pbf';
@@ -152,6 +152,10 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
         const params = this._pendingLoadDataParams;
         delete this._pendingCallback;
         delete this._pendingLoadDataParams;
+
+        const perf = (params && params.request && params.request.collectResourceTiming) ?
+            new performance.Performance(params.request) : false;
+
         this.loadGeoJSON(params, (err, data) => {
             if (err || !data) {
                 return callback(err);
@@ -171,8 +175,8 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
                 this.loaded = {};
 
                 const result = {};
-                if (params.request && params.request.collectResourceTiming) {
-                    const resourceTimingData = perf.getEntriesByName(params.request.url);
+                if (perf) {
+                    const resourceTimingData = perf.finish();
                     // it's necessary to eval the result of getEntriesByName() here via parse/stringify
                     // late evaluation in the main thread causes TypeError: illegal invocation
                     if (resourceTimingData) {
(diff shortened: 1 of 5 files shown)


sbma44:
Please let me know if you'd like to see benchmarks here -- I haven't bothered running them yet given how slight this change is.

.flowconfig
```diff
@@ -8,6 +8,7 @@
 .*/node_modules/htmltojsx/.*
 .*/node_modules/documentation/.*
 .*/node_modules/module-deps/.*
+.*/node_modules/bcryptjs/src/bower.json
```
jfirebaugh: What's the rationale for this change? Doesn't look like bcryptjs is in the dependency tree.

ryanhamley: I ran into something similar to this the other day when I was using `npm install` rather than `yarn install`. I can't say for sure that's where this was introduced but I kept getting a test error in bcrypt until I removed node modules and reinstalled with yarn. This shouldn't be needed.

sbma44: It's present in my node_modules after a fresh `npm i` and causes flow test failures thanks to [some templating wackiness](https://github.com/dcodeIO/bcrypt.js/blob/684fac6814a81d974c805a15e22fd69922c7ca6e/src/bower.json.template#L4). Not related to this PR and I'm happy to do whatever's most appropriate here. Replicated with node 6 and 8.

ryanhamley: Tom, if you remove this line from `.flowconfig` then `rm -rf node_modules` and reinstall with `yarn install` does the bcrypt error go away?

jfirebaugh: `npm install` doesn't use `yarn.lock`, so it installs a newer version of `@mapbox/batfish`, which pulls in bcryptjs via some transitive dependencies.

Let's revert here and consider using https://www.npmjs.com/package/use-yarn to avoid this gotcha in the future.

sbma44: @ryanhamley yes, confirmed that using yarn avoids the problem

@jfirebaugh sounds good, I've removed the relevant commit

ryanhamley: I opened a ticket for the npm bug https://github.com/mapbox/mapbox-gl-js/issues/6774

sbma44:
Rebasing away the bower config commit did something weird to my last merge from master, prompting another merge and a weird-looking commit history in this PR. LMK if this is an issue and you need a fresh branch.

Issue #6725, opened 2018-05-24
Improve feature-state expression documentation


samanpwbb:
## Motivation

I find it hard to understand the `feature-state` expression based on the description in the spec. See here:

https://github.com/mapbox/mapbox-gl-js/blob/1a8e9fec61af6569cb60d074bb07ad7fbc1e9704/src/style-spec/reference/v8.json#L2609

### Implementation

Fully document how `feature-state` works. What is "current feature's state" and how do I set it?

asheemmamoowala:
The `feature-state` is set programmatically, but the API will be different on each platform. Given that the API is only available on gl-js at this time, I think it would be hard to describe in the style-spec how it is set. More importantly, the style-spec should not be trying to specify implementation or API details.

That said, there is still room to improve the documentation for this expression.

anandthakker:
Y, the spec shouldn't discuss API details, but indicating that feature state is set at runtime in the GL JS (and platform SDKs, once they support it) probably makes sense

PR #6737, opened 2018-05-24
Improve documentation for feature-state expression in style-spec

asheemmamoowala:
Addresses #6725.

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2606,7 +2606,7 @@
         }
       },
       "feature-state": {
-        "doc": "Retrieves a property value from the current feature's state. Returns null if the requested property is not present on the feature's state object.",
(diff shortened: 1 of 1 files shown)


src/style-spec/reference/v8.json
```diff
@@ -2606,7 +2606,7 @@
         }
       },
       "feature-state": {
-        "doc": "Retrieves a property value from the current feature's state. Returns null if the requested property is not present on the feature's state object.",
+        "doc": "Retrieves a property value from the current feature's state. Returns null if the requested property is not present on the feature's state. A feature's state is not part of the GeoJSON or vector tile data, and must be set programmatically on each feature. Note that [\"feature-state\"] can only be used with paint properties that support data-driven styling.",
```
asheemmamoowala: > A feature's state is not part of the GeoJSON or vector tile data, and must be set programmatically on each feature.

@anandthakker @samanpwbb Does this statement work for capturing that feature state must be set programmatically on each session? Not sure if the docs should also mention that the state doesn't persist across sessions/reloads of the map.

PR #6747, opened 2018-05-29
container option must be provided

jingsam:


jingsam:
Or should we delete the `Invalid type:` part

PR #6756, opened 2018-05-29
Fix send / remove timing bug in Dispatcher

uforic:
## Description
I _believe_ there is a JavaScript worker timing bug in dispatcher.js. On our Flexport Sentry, we are seeing infrequent bugs like this:
`Unable to get property 'send' of undefined or null reference`, with a fuller stack trace:

```
TypeError: Unable to get property 'send' of undefined or null reference
  at e(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:517:481)
  at t.prototype.loadTile(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:231:2184)
  at t.prototype._loadTile(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:223:2168)
  at h(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:223:8194)
  at n(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:223:7400)
  at o(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:223:6566)
  at r(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:223:795)
  at this(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:521:839)
  at e(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:521:1045)
  at o(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:231:1210)
  at t(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:207:438)
  at n.onload(/home/app/flexport/node_modules/mapbox-gl/dist/mapbox-gl.js:503:1103)
```

The TypeError is occurring on this line [here](https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/dispatcher.js#L58).

It appears that the way this error is getting caused is:
* In componentWillUnmount, the implementor calls "map.remove()"
* A web worker running a task calls "Dispatcher.send", or a previously running broadcast call called `asyncAll(Dispatcher.send(payload...`.
* Dispatcher.send tries to access the 0th index of an empty array, because remove() has already been completed.
TypeError.

Open to suggestions on how to improve this PR. One thing I'm curious about is:
**Is it possible for Dispatcher.remove() to be called, but an async process then tries to call Dispatcher.send()?**

It's totally possible that this is somehow user error, and that the user is somehow dispatching an action after having called map.remove(), but looking through the code (especially the Dispatcher broadcast function earlier in the file), it seems like like the answer is "yes".

Another question: Does anyone depend on the actor ID being returned? Will -1 have any negative downstream impact?

### DIFF
--- a/src/util/dispatcher.js
+++ b/src/util/dispatcher.js
@@ -50,6 +50,10 @@ class Dispatcher {
      * @returns The ID of the worker to which the message was sent.
      */
     send(type: string, data: mixed, callback?: ?Function, targetID?: number): number {
+        if (this.actors.length === 0) {
+            // an async send call is getting executed but remove has been called
+            return -1;
+        }
         if (typeof targetID !== 'number' || isNaN(targetID)) {
             // Use round robin to send requests to web workers.
             targetID = this.currentActor = (this.currentActor + 1) % this.actors.length;


jfirebaugh:
Hi @uforic. Thanks for the contribution! It looks to me like the proposed change here, while it will silence the TypeError, isn't really addressing the root cause, which is that something is attempting to send a message to to the worker after `map.remove()` is called. One of the things that `map.remove()` is supposed to do is cancel any pending async work, so that this doesn't happen.

I don't _think_ it's related `broadcast` _per se_, because `broadcast` calls `actor.send` for all actors synchronously. The asyncAll you see there is to handle the fact that `actor.send` is itself async, so `broadcast` needs to collect all the async results before calling `cb`.

Can you dig a little more into why something is attempting to send even after `map.remove()`? From the stack trace I suspect it's related to tile loading, but with uglified function names in the stack it's hard to tell more.

uforic:
Howdy @jfirebaugh , thanks for getting back to us. We were able to get some more information on this one.

Another raw stack with occasional function names coming through:
```
TypeError: Unable to get property 'send' of undefined or null reference
  at i.prototype.send(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:654324)
  at t.prototype.loadTile(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:370297)
  at t.prototype._loadTile(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:350246)
  at t.prototype._addTile(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:356189)
  at t.prototype._updateRetainedTiles(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:355406)
  at t.prototype.update(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:354762)
  at Anonymous function(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:349075)
  at o.prototype.fire(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:657088)
  at o.prototype.fire(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:657292)
  at Anonymous function(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:369255)
  at s(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:339754)
  at n.onload(/packs/ClientApp.e9f7c8958d58adb5da75.js:33:650379)
  at r(/packs/vendor.dad33dd45589c0fbc421.js:55:12493)
```

I dug into a couple frames, and got their line nos in GitHub:

Bottom of stack:
https://github.com/mapbox/mapbox-gl-js/blob/d3f8ffc305dce86d5b9f488ad54739105fb8e141/src/util/ajax.js#L82

Next up:
https://github.com/mapbox/mapbox-gl-js/blob/0a53a149f00032e5f85f93de0df2c0720402fa61/src/source/raster_tile_source.js#L58

Totally possible I'm misdiagnosing things, but it seems like the onload to getJSON is getting called after map.remove has been called, which is then triggering send action via loadTile.

Let me know what you think! Perhaps those promises need to be cancellable? I notice that the return type of getJSON is never stored off anywhere.

jfirebaugh:
Thanks, that was perfect detail. I opened a PR that should fix this: #6826.

dmnd:
Excellent, thanks for the fix @jfirebaugh! 🙇‍♂️

PR #6764, opened 2018-05-30
Allow for relative paths when loading RTL plugin

### LINKED ISSUE #6719: 0.45.0 broke setRTLTextPlugin on local source
**mapbox-gl-js version**: 0.45.0
**browser**: various

### Steps to Trigger Behavior
 1. Save the RTL example into a static html: https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-rtl-text/
 2. Save `mapbox-gl-rtl-text.js` next to it
 3. Change the code to load locally: 
```
mapboxgl.setRTLTextPlugin('/mapbox-gl-rtl-text.js.min')
```
 4. Start `node serve` or `python3 -m http.server`

### Expected Behavior
Should work.

### Actual Behavior
- Doesn't load RTL plugin in Chrome.
- **Breaks all vector layers in IE 11**.

OK, so this wasn't an easy one to diagnose. In IE 11 all vector layer renderings were broken. Not just text, every layer, as in nothing was displayed at all, only raster layers. Reverting to 0.44.2 makes everything work.

It turns out, the culprit is 0.45.0's `setRTLTextPlugin()`, but only when used with local sources. With https CDN it works well.

ryanhamley:
## Launch Checklist

    - Allows for relative paths when loading RTL plugin

Closes #6719

### DIFF
--- a/src/source/rtl_text_plugin.js
+++ b/src/source/rtl_text_plugin.js
@@ -1,6 +1,7 @@
 // @flow
 
 import { Event, Evented } from '../util/evented';
+import browser from '../util/browser';
 
 let pluginRequested = false;
 let pluginURL = null;
@@ -34,7 +35,7 @@ export const setRTLTextPlugin = function(url: string, callback: ErrorCallback) {
         throw new Error('setRTLTextPlugin cannot be called multiple times.');
     }
     pluginRequested = true;
-    pluginURL = url;
+    pluginURL = browser.resolveURL(url);
     _completionCallback = (error?: Error) => {
         if (error) {
             // Clear loaded state to allow retries
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -3,7 +3,6 @@
 import { Event, ErrorEvent, Evented } from '../util/evented';
 
 import { extend } from '../util/util';
-import window from '../util/window';
 import EXTENT from '../data/extent';
 import { ResourceType } from '../util/ajax';
 import browser from '../util/browser';
@@ -197,7 +196,7 @@ class GeoJSONSource extends Evented implements Source {
         const options = extend({}, this.workerOptions);
         const data = this._data;
         if (typeof data === 'string') {
-            options.request = this.map._transformRequest(resolveURL(data), ResourceType.Source);
+            options.request = this.map._transformRequest(browser.resolveURL(data), ResourceType.Source);
             options.request.collectResourceTiming = this._collectResourceTiming;
         } else {
             options.data = JSON.stringify(data);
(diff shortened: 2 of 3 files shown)


src/source/rtl_text_plugin.js
```diff
@@ -1,11 +1,30 @@
...
+    if (isAbsoluteUrl(path)) {
+        return path;
+    }
+    // resolves from root URL
+    // if path begins with a '/', it will be appended to the domain
+    // otherwise, it will be appended to the full path
+    const a = window.document.createElement('a');
+    a.href = path;
+    return a.href;
+};
```
kkaefer: Looks like we're doing something similar in https://github.com/mapbox/mapbox-gl-js/blob/6255e9e449d1d2cffb9810de57e9a8d97275a500/src/source/geojson_source.js#L288-L292

Can we move that code to a more central place to avoid duplicating it?

src/source/rtl_text_plugin.js
```diff
@@ -1,11 +1,30 @@
...
+const isAbsoluteUrl = (url) => {
+    // handles https, http, and protocol-relative URLs
+    const isAbsolute = new RegExp('^([a-z]+://|//)', 'i');
+    return isAbsolute.exec(url);
+};
+
+const resolvePluginUrl = (path) => {
+    if (isAbsoluteUrl(path)) {
+        return path;
+    }
```
kkaefer: Do we need to special case absolute URLs? It looks like we could also just happily pipe absolute/protocol-relative URLs through the `a`/`href` resolver. If we need this, it'd be a good idea to add the absolute URL special case to the GeoJSON resolver (see below) as well)

src/source/rtl_text_plugin.js
ryanhamley: Ah thanks for pointing that out. Yeah, we should be able to centralize this somewhere.

ryanhamley:
I'm not sure how to test this other than manually. There's no straightforward way to test this in the unit tests since Node doesn't have an `window.location.origin` and I don't think you can set it since `origin` is `readonly`. The render tests hardcode the plugin through an import statement and changing this to load the plugin via the `setRTLTextPlugin` method seems to break all the render tests. Any ideas? Am I overthinking this? It definitely works since it was tested on the debug page `csp.html`. @kkaefer

ChrisLoer:
> I believe we've been relying on manual testing for the RTL plugin.

Mainly, yes, and I think testing manually with csp.html is probably the most important/effective test for this pathway. The style unit test has a "register plugin listener" test case that could easily be modified to test the handling of a relative path -- but as @ryanhamley pointed out that gets tripped up on not having a `window.location.origin`. I think you could probably use Sinon's stubbing to handle that, but you'd be testing a pretty narrow/contrived case.

Thanks for fixing this, Ryan!

Issue #5654, opened 2017-11-10
CollisionIndex stores boxes outside of grid range


ChrisLoer:
We only store collision boxes in a grid that covers the area of the viewport and a little bit of padding outside of it. When we run collision detection, we may place symbols that end up completely outside the range of that grid. These symbols will never detect any collisions, so they'll always get placed. This doesn't usually matter, because they're offscreen. However, if you pan the map to bring them onscreen, they'll be visible until the next placement occurs, which might then detect collisions and cause them to fade out.

In practice, this doesn't seem to be very disruptive -- placement usually happens quickly enough that by the time the too-densely-placed labels come on screen, they're already mostly faded out.

/cc @ansis @jfirebaugh @anandthakker

ChrisLoer:
This is more correct but less performant than I thought on my first reading of the code. 😬 

Items outside the grid range get added to the cells right along the edge of the grid. This makes the grid slower than it needs to be, but also prevents the problem I originally described. Essentially it makes the range of the grid infinite, but the performance gets worse the more things you place outside the defined range.

ChrisLoer:
This is more performant but less correct than I thought on my second reading of the code. 😬

It's true that we store boxes for items outside the grid range (which does slow queries along the edges down), but when we query _entirely_ outside the grid range, we'll short circuit the query results to "false":

https://github.com/mapbox/mapbox-gl-js/blob/13ea16df5eabc4c8867d69e81da180cdcd6cc6b2/src/symbol/grid_index.js#L90-L93

This means that collision detection still works for features that have any part of their geometry touching the grid, but not for features entirely outside the grid.

> In practice, this doesn't seem to be very disruptive -- placement usually happens quickly enough that by the time the too-densely-placed labels come on screen, they're already mostly faded out.

It may not be terrible, but it was at least ugly enough to get reported as #6489.

dagjomar:
Amy progress on this?

ChrisLoer:
@dagjomar I put together a PR at #6784, we should merge it soon. If you're interested, you might try checking out that branch to see if you get the behavior you'd expect after the changes.

mourner:
This was fixed in #6784.

PR #6784, opened 2018-06-06
Don't place symbols entirely outside range of collision grid.

ChrisLoer:
Fixes issue #5654, brings gl-js behavior in line with gl-native.
symbol-placement/line-overscaled functions as a regression test.

Performance-wise, this change means:

- Mild loss: We do an extra bounds check for every box or circle we add to the collision index
- Mild win: We avoid adding items to the index that are outside of the grid range, which makes grid queries for boxes that are right at the edge of the index cheaper.

... I don't know if that's an overall win or loss, but it doesn't look large enough to show up in our benchmarking.

I played around with the debug page to check that rapid panning no longer creates temporarily-too-dense labeling (which generated a few bug reports).

 - [x] post benchmark scores

screenshot 2018-06-06 11 27 30

/cc @ansis @dagjomar @watbywbarif


ChrisLoer:
Oops didn't notice that I left this sitting around with a broken merge on top of the cross-source collision changes. Should be good to go now.

PR #6788, opened 2018-06-08
Add render test for collator expressions.

ChrisLoer:
The expression test suite is a more natural home for collator tests, but we don't have the expression test suite hooked up on Android and the collator expression has Android-specific functionality that needs testing.

My plan is to cover collator expressions using the Android render test suite that @tobrun just implemented.

/cc @anandthakker @tobrun @LukasPaczos


test/integration/render-tests/collator/default/style.json
```diff
@@ -0,0 +1,225 @@
...
+                    ["==", ["var", "op"], "<"],
+                    ["<", ["var", "a"], ["var", "b"], ["var", "collation"]],
+                    ["==", ["var", "op"], "<="],
+                    ["<=", ["var", "a"], ["var", "b"], ["var", "collation"]],
+                    ["==", ["var", "op"], ">"],
+                    [">", ["var", "a"], ["var", "b"], ["var", "collation"]],
+                    ["==", ["var", "op"], ">="],
+                    [">=", ["var", "a"], ["var", "b"], ["var", "collation"]],
+                    false
+                ]]
```
anandthakker: This is a nice test expression! :)

Nit: could it be a little simpler with `match`?

```js
                ["to-string", ["match",
                    ["var", "op"],
                    "==", ["==", ["var", "a"], ["var", "b"], ["var", "collation"]],
                    "!=", ["!=", ["var", "a"], ["var", "b"], ["var", "collation"]],
                    "<", ["<", ["var", "a"], ["var", "b"], ["var", "collation"]],
                    "<=", ["<=", ["var", "a"], ["var", "b"], ["var", "collation"]],
                    ">", [">", ["var", "a"], ["var", "b"], ["var", "collation"]],
                    ">=", [">=", ["var", "a"], ["var", "b"], ["var", "collation"]],
                    false
                ]]
```

test/integration/render-tests/collator/default/style.json
```diff
@@ -0,0 +1,225 @@
+{
+  "version": 8,
+  "metadata": {
+    "test": {
+      "height": 512,
```
anandthakker: Can this work at a smaller size (w/ a corresponding smaller font size)?  Or does that make the image illegible?

test/integration/render-tests/collator/default/style.json
```diff
@@ -0,0 +1,225 @@
...
+          },
+          {
+            "type": "Feature",
+            "properties": {
+                "a": "a",
+                "b": "ä",
+                "operator": "<",
+                "case-sensitive": true,
+                "diacritic-sensitive": false
+            },
```
anandthakker: This combination looks like a duplicate of the one at line 67

test/integration/render-tests/collator/default/style.json
ChrisLoer: Good idea! Yeah it was actually nice to be forced to think a little bit about a non-trivial collator expression.

test/integration/render-tests/collator/default/style.json
ChrisLoer: Oops, meant that to be diacritic-sensitive!

test/integration/render-tests/collator/default/style.json
ChrisLoer: Yeah I brought it down to 256x256 at 11 point font, still easy to read.

PR #6790, opened 2018-06-08
Move CircleCI config to proper location.

felicianotech:
Hey Folks,

I noticed that the CircleCI config has been update for 2.0 (great) but the file wasn't moved to proper location. In order to prevent any issues from creeping up after CircleCI 1.0 is EOL'd, I decided to go ahead and open a PR moving the config file to the correct location.

Please let me know if there are any questions. :smile:


jfirebaugh:
Thanks @felicianotech!

I personally prefer having the configuration file at the top level, as opposed to being the only file within a hidden directory. Let's keep it this way for now -- if it stops working at some point, it's easy to adapt.

PR #6792, opened 2018-06-10
Fix DEPLOY_ENV for website staging builds

davidtheclark:
@samanpwbb noticed that the staging docs are being included in Google search results.

Although our staging site has a `robots.txt` that discourages crawlers, that will not always prevent indexing if the page is linked to or otherwise attracts the great spider. I think the next measure to take is to add a `<meta name="robots" content="noindex">` to the page.

dotcom-page-shell is set up to add this meta tag if `DEPLOY_ENV=staging`. However, right now for this site `DEPLOY_ENV` is set to `production` for every build, even on staging.

This PR adds a script that determines whether to set `DEPLOY_ENV=staging` based on the Git branch name. I used a shell script in order to avoid adding an npm dependency — but if you'd prefer, I can use the git-branch package or something and make it a Node script.

Notice that this branch merges into dev-pages. I'm not sure whether you'd rather modify dev-pages first, or modify mb-pages then update dev-pages from it. 🤷‍♂️

### DIFF
--- a/build/set-deploy-env.sh
+++ b/build/set-deploy-env.sh
@@ -1,0 +1,21 @@
+# Establish environment variables based on the branch name.
+#
+# This is an sh script (instead of Node or bash) because it needs to run in the
+# top-level process so the environment variables it creates are available to
+# all children processes (e.g. `x && y && z`). `npm run` executes in an `sh`
+# environment -- so, sh it is.
+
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Thanks! I merged this and then reset `dev-pages` to `mb-pages`.

Issue #6796, opened 2018-06-11
fill-outline-color docs display bug


lbud:
The `fill-outline-color` [entry](https://www.mapbox.com/mapbox-gl-js/style-spec/#paint-fill-fill-outline-color) on the style spec documentation page is omitting a word:
> Requires fill-antialias to be . 

![image](https://user-images.githubusercontent.com/3170312/41247578-37dcfb9e-6d63-11e8-8bd3-599ac528551f.png)

PR #6797, opened 2018-06-11
Fix omitted word on the style spec documentation page

pathmapper:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/6796


jfirebaugh:
Thanks @pathmapper. This really should be a boolean value though, so I fixed this in a different way in 2e08111da56873ea4db2e42d7258d42699a24e9d.

Issue #6768, opened 2018-05-31
Fix vector tile overdrawing with overlapping tiles from different zoom levels


jfirebaugh:
The tile loading logic in Mapbox GL JS attempts to load a parent tile in certain cases when a request for a child tile fails, e.g. the tile response is a 404, or a 200 with a content length of 0. This is intended to compensate for situations where tiles are not available at the current zoom level, but may be available at lower zoom levels, which could be overzoomed.

Unfortunately, this leads to overdraw in certain situations. I believe the primary situation (there may be others) involves:

* A vector tileset that does not cover the whole world, i.e. tiles exist only for a certain bounding box. Outside that bounding box, the Mapbox tile API will respond with 404s.
* Layers sourced from this tileset that use semi-transparent styling.
* A map viewport that intersects the bounds of the tileset.

Under these conditions, the map succeeds in loading one or more tiles at zoom `z` inside the bounds, but fails on one or more tiles outside the bounds. This triggers requests for tiles at `z-1` that succeed, returning tiles that overlap the child tiles. Both parent and child tiles are rendered, producing overdraw. (The fallback loading behavior was originally implemented for fully opaque raster tiles, and semi-transparent vector tiles was not taken into account.)

Multiple issues have been filed relating to this behavior:

* #4691, involving a transparent circle layer ([JSFiddle](https://jsfiddle.net/nlindblad/nucky1bm/)):
  ![mapbox-bug](https://cloud.githubusercontent.com/assets/164991/25857886/42fcb304-34d2-11e7-8ee4-dbf6e3f1636f.gif)
* #5829, similarly ([demo](https://www.mapbox.com/help/demos/heatmap-gljs-dds/demo-3.html)):
   ![circle-zoom](https://user-images.githubusercontent.com/8291663/33771091-635bce58-dbf5-11e7-958b-fef415a340d4.gif)
* #6386, similarly ([jsbin](http://jsbin.com/yobagowica/edit?html,output)
   ![v44dds](https://user-images.githubusercontent.com/17933661/37838375-1c89dad8-2e8e-11e8-8e4a-cfcc05a5e94e.gif)
* #6560, with a heatmap layer ([example](https://bl.ocks.org/j5kay/a50925604a76c892826e5d3a7db77ddd))
* #6754, with a fill-extrusion layer ([JSFiddle](https://jsfiddle.net/9ou7r119/5/))

[...]

Issue #6783, opened 2018-06-06
Mapbox not able to change the style of vector tiles at run time


Amit-Gore:
I am using tippecanoe command line utility to create my application vector tileset. This is creating a directory structure as per the z/x/y coordinates which is perfectly fine. I have a certain group of features(allocated with a layer) which do not need z to be up to 21 zoom level so it's creating the tiles up to zoom-level 14. It is critical in my case to not to waste the memory space by increasing the max-zoom to 21 for certain layer of features.

As per my understanding, mapbox gl-js queries for the vector tiles as per its coordinate space.

So during my zoom-in from 6 to 21, although zoom-level > 14 tile queries are responding with 404, gl-js is adopting the same tile which is available at 14.

The problem is,

If at zoom-level > 14, I trigger a style change by any mean upon current layer-tile it's not able to do that.

Is it trying to change the style of a tile which has responded with 404? Is it possible to make it change the available tile source obtained at zoom-level 14?

jfirebaugh:
> If at zoom-level > 14, I trigger a style change by any mean upon current layer-tile it's not able to do that.

I'm having trouble understanding this. Can you explain in more detail what you're trying to do, what you expected to happen, and what happens instead? If you can create a small code example that shows it, that would be great.

Amit-Gore:
For example, If I click on any feature, I need that feature to be highlighted. I am doing it by filtering out the layers as : 
//hiding the current layer 
`mapBox.setFilter(currentLayer, ["==",'gid', "_none_"]);` 

//showing only the clicked feature by filtering it out with a unique id it has
`mapBox.setFilter(highlightedLayer, ["==",'gid', feature_gid]);`

This works as expected for zoom < 14 but if zoom exceeds 14 (the max-zoom while tile creation) then it is not able to render the tile with applied layer style as it tries to fetch a tile which is not there on my server. So my question is if a source tile at particular zoom level is giving 404 then why not apply the layer style to whichever tile is available at zoom level 14 ?
at zoom < max-zoom used while tile creation
![image](https://user-images.githubusercontent.com/7140231/41052056-712eab54-69d5-11e8-9aaa-0538c736a04f.png)
at zoom > max-zoom used while tile creation
![image](https://user-images.githubusercontent.com/7140231/41052257-e9b49692-69d5-11e8-97ed-2860beb927b5.png)

jfirebaugh:
[...]

Issue #5013, opened 2017-07-21
Overzooming partially not working 


pathmapper:
**mapbox-gl-js version**:
v0.38.0

### Steps to Trigger Behavior

 1. open https://jsfiddle.net/f8kf4ps2/
 2. zoom in

### Expected Behavior

Lines/labels are drawn when overzooming

### Actual Behavior

Some lines/labels are not drawn when overzooming / some tiles are missing (Console: GET receives 404)

The tileset layers used for the lines and labels are having a zoom range z12 - z12.

Screenshot Mapbox Studio z12.7
![image](https://user-images.githubusercontent.com/20856381/28454964-6a544b56-6dfd-11e7-9018-aa6a0b0d0b9b.png)

Screenshot Mapbox Studio z13.1
![image](https://user-images.githubusercontent.com/20856381/28454990-8c2193ec-6dfd-11e7-9b51-8700b3b8fb36.png)

pathmapper:
This issue is related to the custom tileset I am using for the style.

What I've done to create it: 
Created several tilesets (.mbtiles) with Mapbox Studio Classic. These tilesets are having different zoom ranges and the layers are having different buffer sizes.
These separate tilesets are all working like expected with a style.

Then I've used `tile-join` of tippecanoe to combine these separate .mbtiles files into one.
Two layers of this single tileset (not used for the style linked with the jsfiddle above) are having a zoom extend of z9-z13. 

When I replace these two layers, using a zoom range of z9-z12,  overzooming works as expected. 

I am unsure if this issue described above is related to:

- GL JS
- tippecanoe  `tile-join`
- Mapbox Studio Classic
- my approach using different buffer and zoom ranges for layers in one single tileset is violation of some specs
- something else

If it's not related to GL JS, sorry for the noise.

pathmapper:
With regard to https://github.com/mapbox/mapbox-gl-js/pull/6803#issuecomment-396757100 :

> I wonder if there are many cases of users relying on the current behavior of overzooming parent tiles when a 204/404 occurs.

and 

https://github.com/mapbox/mapbox-gl-js/pull/6803#issuecomment-396769363 : 

> We didn't anticipate this use case.

The motivation for merging tilesets with different maxzoom and buffer sizes in one single source was the following:

[...]

PR #6803, opened 2018-06-12
6768 overdraw fix

ryanhamley:
## Launch Checklist

    - This partially solves #6768 by treating 404'd tiles as renderable empty tiles. This is similar to how GL-Native treats these tiles as seen [here](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/tile/tile_loader_impl.hpp#L97). This is a compromise position between not retaining parent tiles for tiles that don't exist, which produced a noticeable flash at each zoom level as tiles drew in, and creating a mask for vector tiles similar to the one used with raster tiles, which forced circle features to be cut off at tile boundaries. Several variations of these approaches were attempted without achieving satisfactory results, but not retaining parent tiles came the closest; the trade off was a flash at every zoom level for every tile set which seems unacceptable as a solution to a more limited bug.
    - This proposed solution does not introduce a flash of tile loading, cut off features or allow the overdraw to be indefinitely displayed on the screen, but it does produce a flash of overdraw at new zoom levels. This flash goes away once the tiles are cached. [See this example](https://www.dropbox.com/s/bgn2fqkrdyvuoz3/overdraw.mov?dl=0).
    - Eventually, GL-Native's tile loading/caching logic will need to be ported over to GL-JS and that may improve this functionality a bit more.


ryanhamley:
@mollymerp Do you have anything to add to this? Anything I'm forgetting to mention?

jfirebaugh:
> This is similar to how GL-Native treats these tiles as seen here.

Are you sure this is how native works? I was under the impression that it also overzooms parent tiles when receiving a 204/404 and the same issue exists there. (Note that a tile 404 actually does not produce `Reason::NotFound`. For historical reasons, [tiles that 404 are treated as "no content"](https://github.com/mapbox/mapbox-gl-native/blob/4943260d3b94139452ef53e13ca520ceb1cb8045/platform/default/http_file_source.cpp#L457-L458).)

I wonder if there are many cases of users relying on the current behavior of overzooming parent tiles when a 204/404 occurs. It isn't as important for vector sources as it is for raster sources, where heterogenous maximum zoom levels are common, but there's probably cases where it's expected -- #6783 was a recent example, and possibly #5013.

mollymerp:
@jfirebaugh thanks for the clarification around the 404 handling – TIL. 

In that code you linked, the 404s then do not cause `reponse->error` to be set at all, and the `noContent` flag causes [`data` to be a `nullptr` for that tile ](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/tile/tile_loader_impl.hpp#L109), but eventually the empty tile still gets marked as `loaded = true; renderable = true` which prevents the `updateRenderables` algorithm from retaining the empty tile's parent.

mollymerp:
I think this is the relevant change in native where we started rendering errored/empty tiles https://github.com/mapbox/mapbox-gl-native/pull/10011

> this part of the code seems to have a history of a fix for one issue introducing another issue.

trueee 😅

I do see what you mean about users relying on parent tile overzooming when tiles 404. maybe the check we discussed before (don't retain parent tiles if at least one child successfully loads and one child 404s/errors) will have fewer side-effects than this change

ryanhamley:
Yeah I can confirm actually that #5013 is worse with this solution. The overzoomed tiles fail completely 😞 I assume the same is true for the other issue John pointed out. So this actually isn't going to work. We didn't anticipate this use case.

mollymerp:
picking this back up in favor of #7051. per chat, this is what native does, and we should keep behavior consistent cross-platform. note that we'll still want to revise our strategy if/when we implement server-side 204 responses for `no content`.

mollymerp:
note that #5013 was an issue with how the tileset was generated, overzooming at ZL > tileset maxzoom still works w this PR.

jfirebaugh:
Let's see if we can write a render test for this. Here's a sketch:

* Add a subdirectory to `test/integration/tiles` named `sparse` or something like that. Add a 0/0/0 vector tile that covers the world with a solid feature.
* Add an integration test that renders a fill layer at z1 using a source pointing to this "tileset", with a contrasting background color.
* The expected result should be the background color: the server will send a 404 for z1 tiles, and we don't want to try loading z0. (Test without this PR applied, to make sure the test is failing as expected.)

There might be a reason why a render test of this type _doesn't_ work, but let's give it a shot. It would be nice for maintaining parity in behavior between js and native.

ryanhamley:
@jfirebaugh What makes you say that a render test like the one you described may not work? The test I have set up produces the expected output of just a background color, but it does this with or without the fix. I'm unable to produce a failing test.

jfirebaugh:
> What makes you say that a render test like the one you described may not work?

It's out of the norm for render tests, which tends to make things tricky. Not being able to produce a failing test is a common way for this to manifest. It means you'll have to debug the "expected to be failing" case to see why it's not failing. In this case, we expect it to be failing because it will load and render the parent tile, so the first thing to check is whether it's actually loading the parent tile. Or is it, say, indicating that all tiles are loaded as soon as the z1 tile fails, and stopping there?

Sometimes when you figure out why it's not failing, it's something you can fix with a tweak to the test case or the harness. But sometimes it's not, and you have to settle for a gl-js-specific unit test (or if even that's a disproportionate effort, manual testing).

jfirebaugh:
Actually, the issue here may just be that your layer order is swapped. The background layer should be first.

ryanhamley:
I did need to reverse the layer order, but there was also an inconsistency between `src/util/ajax#getArrayBuffer` and `test/ajax_stubs#getArrayBuffer`. The stubbed version was not returning an `error` object with a `status` property which is what the library is using to check for 404 status so I patched that and confirmed that the test fails without this PR and passes with it.

Without PR
screen shot 2018-08-06 at 4 27 48 pm

With PR
screen shot 2018-08-06 at 4 28 14 pm

All other tests pass except for `render-tests/mixed-zoom` which seems to be testing some label placement when zooming in. @ChrisLoer looks like you wrote this test. Can you elaborate on exactly what it was this is testing so we can figure out how to fix it?

Without PR
screen shot 2018-08-06 at 5 11 55 pm

With PR
screen shot 2018-08-06 at 5 11 47 pm

ChrisLoer:
> Can you elaborate on exactly what it was this is testing so we can figure out how to fix it?

The key thing that test case was focusing on was collision detection in the case where symbols are being included from both the overdrawn child and from the parent. The basic approach is to place both tiles but rely on the `CrossTileSymbolIndex` to detect duplicates between parent and child and hide the duplicate parent symbols:

- There shouldn't be any double-draw in the overlapping area (visually, you'd notice double-draw as labels looking bolder as they crossed the tile border)
- Parent symbols _can_ show in the child tile area, as long as nothing else is there to collide them out
- Collision boxes/circles should be seamless at tile boundaries

danielfornies:
Hi guys, just tested this, with tiles generated via tippecanoe which only go until zoom level 14, while I use the map until zoom level 18. After zoom level 15 the tiles aren't rendered anymore.
What should I do to overcome this? I could invoke tippecanoe forcing it until zoom level 18, but that's also a lot of extra data I don't really need (zoom level 14 already contains all the details)
Let me know, thanks!
Daniel

mollymerp:
@danielfornies you need to se tthe `maxzoom` on your source to 14.

danielfornies:
> @danielfornies you need to se tthe `maxzoom` on your source to 14.

Omg, silly me, it was just that, thanks!

shayke:
Hello,
We are using a custom tileserver which serves multiple mbtiles for the same source
(Users can choose which mbtiles they want to include since it's local).
Is there a workaround other than putting maxzoom for the source?
For example I have 2 mbtiles served from the same source:
1. planet - z0-z7
2. usa - z0-z14

If setting maxzoom 7 then mapbox won't request tiles which are available from usa
If setting maxzoom 14 then mapbox won't use the parent tile for z8-z14 for places which I don't have usa tiles (for example europe).

shayke:
> Hello,
> We are using a custom tileserver which serves multiple mbtiles for the same source
> (Users can choose which mbtiles they want to include since it's local).
> Is there a workaround other than putting maxzoom for the source?
> For example I have 2 mbtiles served from the same source:
> 
> 1. planet - z0-z7
> 2. usa - z0-z14
> 
> If setting maxzoom 7 then mapbox won't request tiles which are available from usa
> If setting maxzoom 14 then mapbox won't use the parent tile for z8-z14 for places which I don't have usa tiles (for example europe).

Any pointers here please? This prevents me from upgrading mapbox from 0.47 and I'm not really sure how I could tackle this.

Issue #6806, opened 2018-06-12
heatmap tile clipping when layers are ordered above it


tristen:
Could be related to https://github.com/mapbox/mapbox-gl-js/issues/6768? When a heatmap layer is not ordered above other layers it results the bottom rightmost tile being rendered and no others.

![ordering-issue](https://user-images.githubusercontent.com/61150/41320305-3efec3b6-6e6d-11e8-8f26-a9958011acfe.gif)

**mapbox-gl-js version**:
`v0.45.0`

**browser**:
Chrome `Version 67.0.3396.79 (Official Build) (64-bit)`

### Steps to Trigger Behavior

Order a symbol layer above a heatmap layer.

### Link to Demonstration
https://jsfiddle.net/tristen/wuc56tne/

PR #6807, opened 2018-06-12
Explicitly set heatmap stencil mode

lbud:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/6806. We had not been setting the stencil mode explicitly in the heatmap texture copy pass, meaning it inherited the stencil mode from whatever the previous layer was — presumably heatmaps have been most used immediately above symbol layers (which disable stencil testing) but when they follow a stencil-enabled layer (e.g. fill layers) they inherit a single (most recently used) tile mask from that layer.

## Launch Checklist

 - [ ] @mourner how did you generate the alternate images for heatmap render tests? Or I suppose I could also just set a higher diff threshold here, since failing tests look very different: …?
badheat


mourner:
> how did you generate the alternate images for heatmap render tests? Or I suppose I could also just set a higher diff threshold here, since failing tests look very different: …?

@lbud hmm, not sure I follow... Alternate images?

lbud:
@mourner it looks like many of the heatmap tests have two possible expected images depending on whether the platform supports half-float-textures-based rendering as of https://github.com/mapbox/mapbox-gl-js/commit/d61850a6bd84ce6d697be392c7ef6d4b5555c20e — is that something I need to worry about/should expect to be an issue for some gl-native platforms when this is ported?

mourner:
@lbud the non-half-float version were generated with GL JS render tests because headless gl doesn't support the half-float extension, and the half-float version were generated with GL Native render tests locally. Ideally there would be two versions for every test that has heatmaps, but a higher diff is probably OK too.

Issue #6132, opened 2018-02-12
Minimum size for boxzoom (suggestion)


stevage:
In my map I have an interaction where shift-clicking does something. One issue is that if the mouse moves even slightly during the process of shift-clicking, it triggers a box-zoom, to a tiny box.

My suggestion: don't trigger a box-zoom if the box is less than X by X pixels, where X is something like 20.

There doesn't seem to be any way for me intercept the boxzoom handler and reject it, so my workaround for now is to disable boxzoom any time that the shift-click feature is available.

jfirebaugh:
> intercept the boxzoom handler and reject it

Thinking about a potential way to do this in #3290.

PR #6809, opened 2018-06-13
configurable drag pan threshold

### LINKED ISSUE #1832: Add a movement threshold for drag panning
As a user, very easy to trigger drag event when you really meant for a click event. Can listen to dragstart+end and fire click if drag was only a pixel or two, but wondering if there is an option to set drag threshold directly.. if not, should there be?

msbarry:
This is an attempt to fix #1832. I added a `dragThreshold` option which lets you set the threshold above which a drag is considered a drag and below which a drag is considered a click. It defaults to 0, which retains the existing behavior.  This is my first pull request here, so let me know if there's a better way to do this or anything that I'm missing, thanks!

 - [ ] post benchmark scores


src/ui/map.js
```diff
@@ -170,6 +172,8 @@ const defaultOptions = {
  *   bearing will snap to north. For example, with a `bearingSnap` of 7, if the user rotates
  *   the map within 7 degrees of north, the map will automatically snap to exact north.
  * @param {boolean} [options.pitchWithRotate=true] If `false`, the map's pitch (tilt) control with "drag to rotate" interaction will be disabled.
+ * @param {number} [options.dragThreshold=0] The threshold, measured in pixels, that determines how far a drag must be before it is considered a drag
```
ryanhamley: I have two related suggestions here:

1) I think `clickTolerance` might be a more accurate name for this option. For one thing, that's what this option is called in Leaflet (even though it's part of the `Draggable` class) which might help people who are migrating to GL-JS or searching for a solution to this problem. Additionally, the original issue for this PR was how to get a click event instead of a drag event when the shift was only a couple pixels, so really this is about setting a tolerance limit around the click event to prevent it turning into a drag.

2) I think the description could be clearer. Consider how Leaflet describes `clickTolerance`: "The max number of pixels a user can shift the mouse pointer during a click for it to be considered a valid click (as opposed to a mouse drag)." You may want to make it clear in your description that the value of the option is the max number *inclusive* since you're checking for `<=` while Leaflet just checks `<`.

test/unit/ui/map_events.test.js
```diff
@@ -553,3 +553,35 @@ test(`Map#on mousedown doesn't fire subsequent click event if mousepos changes`,
     map.remove();
     t.end();
 });
+
+test(`Map#on mousedown fires subsequent click event if mousepos changes less than drag threshold`, (t) => {
```
ryanhamley: Minor again, but I'd spell out `mouse position` instead of `mousePos` in these two test descriptions.

ryanhamley:
Hi @msbarry thanks for submitting this PR! I added a few comments but overall I think it looks pretty good.

msbarry:
Thanks @ryanhamley!  I renamed the option to `clickTolerance` and used the same description from leaflet. I also made the check exclusive instead of inclusive for consistency (mapbox-gl-draw appears to do the same).

What do you think for the default value? Leaflet uses 3, do you think we should switch to that or retain existing behavior for users who don't set the option?

Also, should we apply the threshold to other built-in handlers that look for drags like box zoom and drag rotate? It seems less critical since if you control or shift click it doesn't fire a click event so the drag handlers don't need to differentiate from a click.

mourner:
Awesome PR! No one complained about Leaflet's default in years, so I'd vote for `3` as a default here as well — will make clicking GL JS maps much easier, especially on a trackpad.

ryanhamley:
Switching the default from 0 to 3 would be a breaking change in the API but I'm not sure if there's really a use case where you'd depend on this value being 0 so it seems like a fairly safe breaking change to me.

It doesn't seem like Leaflet uses any type of tolerance detection in `BoxZoom` and it doesn't have a concept of rotation of the map. That being said, I think including a tolerance could be useful. Accidentally turning a click into a drag would be a pretty minor nuisance with drag rotate, but on box zoom it might be possible to inadvertently zoom in close on a very small portion of the map which could confuse a user. Thoughts @mourner?

msbarry:
updated default click tolerance to 3px. Let me know what you decide about box zoom, I think it's fairly straightforward to add if needed.

jfirebaugh:
I think using the same threshold for box zoom is a good idea. It was requested in #6132.

msbarry:
Sounds good. I applied click tolerance to box zoom handler as well.

PR #6812, opened 2018-06-13
Add half float image for heatmap stencil regression test

lbud:
Adds half float test (I _think_ — generated in gl-native) for https://github.com/mapbox/mapbox-gl-js/pull/6807.

## Launch Checklist


mourner:
image
image

@lbud the images look different — in theory, they should be the same (matching shapes, etc.) except non-half-float having lower "resolution", while the above half-float version is more "intensive". Here's an example from `heatmap-color/default` of how the difference should look:

image
image

mourner:
Looks like we can't merge without figuring out the mismatch so I'm going to close this as stale for now.

PR #6813, opened 2018-06-13
do not reload errored tiles

### LINKED ISSUE #6783: Mapbox not able to change the style of vector tiles at run time

mollymerp:
fix #6783 

errored tiles were being reloaded, and their state was being changed from `errored` -> `reloading` which results in `tile.hasData()` to return `true`, preventing the existing parent tiles from being retained correctly. 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -215,7 +215,7 @@ class SourceCache extends Evented {
         this._cache.reset();
 
         for (const i in this._tiles) {
-            if (this._tiles[i] && this._tiles[i].state !== "errored") this._reloadTile(i, 'reloading');
+            if (this._tiles[i].state !== "errored") this._reloadTile(i, 'reloading');
         }
     }
 
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -215,7 +215,7 @@ class SourceCache extends Evented {
         this._cache.reset();
 
         for (const i in this._tiles) {
-            this._reloadTile(i, 'reloading');
+            if (this._tiles[i] && this._tiles[i].state !== "errored") this._reloadTile(i, 'reloading');
         }
     }
 
--- a/test/unit/source/source_cache.test.js
+++ b/test/unit/source/source_cache.test.js
@@ -401,6 +401,32 @@ test('SourceCache / Source lifecycle', (t) => {
         sourceCache.onAdd();
     });
 
+    t.test('does not reload errored tiles', (t) => {
+        const transform = new Transform();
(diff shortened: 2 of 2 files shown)


ryanhamley:
Does this affect #6768 at all?

mollymerp:
@ryanhamley no, this doesn't change the behavior of #6768 as far as I can tell

src/source/source_cache.js
```diff
@@ -215,7 +215,7 @@ class SourceCache extends Evented {
         this._cache.reset();
 
         for (const i in this._tiles) {
-            this._reloadTile(i, 'reloading');
+            if (this._tiles[i] && this._tiles[i].state !== "errored") this._reloadTile(i, 'reloading');
```
anandthakker: Is the first part of the condition necessary, since we're already iterating through the keys of `this._tiles`?

test/unit/source/source_cache.test.js
```diff
@@ -401,6 +401,32 @@ test('SourceCache / Source lifecycle', (t) => {
...
+            if (e.dataType === 'source' && e.sourceDataType === 'metadata') {
+                sourceCache.update(transform);
+                sourceCache.getSource().fire(new Event('data', {dataType: 'source', sourceDataType: 'content'}));
+            }
+        });
+        sourceCache.onAdd();
+
+        t.ok(reloadTileSpy.notCalled);
+
+        t.end();
```
anandthakker: I had to dig in a bit to convince myself that test would cause `reload()` to be executed.  To make it more obvious, could we have `loadTile` load one successful tile and one errored tile, and then make sure that we have only one call to `_reloadTile`?

anandthakker:
👍  🐑

Issue #5061, opened 2017-07-29
Add placement logic for a single centered line label?


nickidlugash:
A fairly common technique for formatting line geometries for text labeling is to create short lines that are intended to have just a single text label centered along the line. We currently use this technique for creating some of our marine labels in Mapbox Streets:

![screen shot 2017-07-28 at 8 09 16 pm](https://user-images.githubusercontent.com/5607844/28740271-bfe8726c-73d0-11e7-8af3-456b886a60f8.png)

 and @ajashton is looking to do this more widely in our next big Mapbox Streets data update. We've come across this need in other contexts too, such as these place labels in AJ's [polar projection map](https://api.mapbox.com/styles/v1/aj/cif75dzfp04ogs3kr6l3dflod.html?title=true&access_token=pk.eyJ1IjoiYWoiLCJhIjoiY2lrZW4zZnE1MDA2NHRvbHl3Y2psejM1dSJ9.L2ShDU7WcmDhj5tZdBLD5w#4.33/-27.164/21.423) :

![screen shot 2017-07-28 at 7 25 44 pm](https://user-images.githubusercontent.com/5607844/28740251-85ecfa6a-73d0-11e7-8b36-02153fccc398.png)

There isn't a great way to display these currently in our renderers: if you play around with text size, symbol spacing, etc, you can sometimes approximate this behavior, but not reliably, given the various logic we have in place to distribute multiple anchors as nicely as possible along a line. The labeling in the map linked above wasn't used in the final version because the placement wasn't reliable enough. The labeling for our marine labels in Streets only looks reasonable because of the sparsity of labels.

### Add `symbol-placement: line-center` and disable clipping lines to tile boundaries?

However, we do already have logic to attempt [placing an anchor at the center of a line](https://github.com/mapbox/mapbox-gl-js/blob/master/src/symbol/get_anchors.js#L91-L98) if the first placement attempt fails. Perhaps we can implement a new value for `symbol-placement` called `line-center` that *only* uses the centering placement logic? (Or another property for defining placement along a line, but the former idea seems simpler).

In order to acheive the desired effect of a single label on a line regardless of tile boundaries, this could be implemented in conjunction with [disabling the clipping lines to tile boundaries](https://github.com/mapbox/mapbox-gl-js/issues/1995) (assuming there is an adequate buffer in the vector tiles).

The reason we currently clip lines to tile boundaries, according to @ansis:

[...]

PR #6821, opened 2018-06-14
`symbol-placement: line-center`

ChrisLoer:
This is a potential implementation of #5061. Currently, it does the simplest thing possible, which is to just place labels in the middle of the line, ignoring `text-max-angle` limitations. I think we should actually implement some form of `text-max-angle` support (either hiding labels that don't pass or moving them away from center until they do pass).

I added a simple test case using one of our Berlin test tiles:

screenshot 2018-06-14 14 52 23

I think it highlights one of the potential problems with using "line-center" on data in existing sources: the data is kind of organized by features (I have feature IDs mapped to colors in this example), but those features can be broken down into individual lines in ways that might not make much sense to a style designer. And of course the "center" they have in mind could easily be the center of multiple joined features.

I know @ajashton and @nickidlugash had in mind using this specifically on source layers designed for center labeling (e.g. a center line for a body of water)... to get a better feel for that use case, it would be helpful for us to design an actual vector tile taking advantage of the large buffers to do center-placing across tile boundaries. We can't just use GeoJSON like we do in most tests because that'll automatically get chopped up at tile boundaries (huh, I guess that could be an issue for someone trying to add cross-tile center-placed labels at runtime, too).

 - [ ] post benchmark scores
 - [x] tagged @mapbox/studio and @mapbox/maps-design if this PR includes style spec changes

/cc @ansis


ChrisLoer:
@mourner pointed out that maybe it makes sense to use `mergeLines` on features with `symbol-placement: line-center`. I tried it out on the same Berlin tile and ended up with:

screenshot 2018-06-19 11 28 02

Just looking at the image, I don't think there's much of a win in terms of developing an intuition for what the "center" of a line really means. On the other hand, using `mergeLines` could undermine explicit decisions made in the data layer to support `line-center` placement... so I think it probably makes sense to skip the merge step?

ajashton:
> On the other hand, using mergeLines could undermine explicit decisions made in the data layer to support line-center placement... so I think it probably makes sense to skip the merge step?

Agree. I think it's fine that line-center will require special consideration / prep of the underlying data. We already have a similar prep requirement for the much more basic task of labeling a polygon (without repeated labels across tiles).

nickidlugash:
> to get a better feel for that use case, it would be helpful for us to design an actual vector tile taking advantage of the large buffers to do center-placing across tile boundaries. 

@ChrisLoer We have line geometries in the `marine_label` layer of `mapbox-streets-v7` that are formatted for this purpose – will that work for testing? (I pulled this branch and took a look at these on the debug page – looks 👌 !)

nickidlugash:
>  if you could point me to a good tile in streets-v7 to test against (preferably one in which the label will go well past the edge of the tile), I can download that as a fixture and write a test case around it.

Tiles 4/4/7 and 4/5/7 are adjacent tiles that have a relevant line crossing the two of them. 

> For text-max-angle, does the "don't show anything if max angle check doesn't pass" behavior sound right to you?

I _think_ this sounds right. What is the alternative – disabling `text-max-angle` for `symbol-placement: line-center`? Although I imagine that anyone using this placement method would be pretty lenient on the max angle (since the label is not repeated), I could still see wanting the option to hide any labels that look really distorted. I think I would just want to double check that the `text-max-angle` default of 45 is reasonable for center lines. If necessary, we could make the default different depending on the symbol placement (since we do have precedent for dependencies in default values), but hopefully we wouldn't have to add this complexity.

By the way, some food for thought: the `symbol-placement` property is not data-driven, and my understanding is that we're not likely to make it DDS anytime soon due to the level of complexity (since it has many cascading changes to the rendering of a symbol). If we are interested in making different line placement options data-driven, it may be simpler to add a new style property for line placement behavior (so center line behavior would look something like `symbol-placement:line` + `line-placement:center`). I raised this question to others on the cartography team, and we all don't have any strong feelings right now towards needing to make line placement behavior data-driven. So this current implementation of the style spec seems fine (all else equal, adding an additional value to an existing property rather than a new property seems desirable), but I just wanted to raise the issue for others to consider as well.

ChrisLoer:
> What is the alternative – disabling text-max-angle for symbol-placement: line-center?

Either disabling it, or relaxing the "center" part and trying to re-sample until we find something _close_ to the center that still fits. I think just hiding sounds right, though, I'm implementing it now.

> So this current implementation of the style spec seems fine (all else equal, adding an additional value to an existing property rather than a new property seems desirable), but I just wanted to raise the issue for others to consider as well.

Yeah that is good food for thought, although I don't feel like I could say anything very informed about it. I think implementation-wise it would not be too hard to make something like `line-placement: center` data-driven since there's not much change to the rendering code. Do we have any use cases in mind where we'd want this to be data-driven?

ChrisLoer:
Here's a render of just 4/5/7:

screenshot 2018-06-21 13 47 33

You can see the line for the `marine_label` gets clipped at the tile boundary, but the center-placed label itself is allowed to render far off the edge of the tile.

Writing the test made me realize some problems, though... if we allow `line-center` symbols to be added outside of the drawing area:
 - If collision detection is off, we can get double-draw for the same label present in two adjacent tiles.
 - If collision detection is on, only one of the symbols will draw :+1:. But the two symbols from adjacent tiles won't have the same `crossTileID` (because their tiles don't have a parent/child relationship). So if you rotate the map and thus change which tile gets placed first, the "placed" one can swap states with the "not-placed" one, and that's going to show up as a flickery fade-in/out.

🤔 I don't have an immediate strategy in mind for addressing those issues -- I think it would require something like a `CrossTileSymbolIndex` that could handle arbitrary tile relationships (any other ideas @ansis?). On the other hand, the problems _might_ not be that bad... 

We could also try to avoid including the "same" label in multiple tiles, but then you'd have the downside that you'd depend on one particular tile getting loaded to show the label.

ChrisLoer:
Err, after thinking about this for a second more, it occurs to me that I probably just imagined the "draw symbols anchored outside the tile boundaries" requirement. I think the actual requirement may just be that the labels themselves should be able to draw across tile boundaries -- in which case we'd be able to skip the problems I just brought up.

The only downside would be the same as in the "only include _data_ in a single tile" case... The tile containing the center-point of the label would have to be loaded before a label could show, so for example you might expect a label that looked like "Carr..." clipped against the right edge of the screen, but it wouldn't start displaying until you moved a bit further to the right, you loaded the tile that contained the center of the label, and "Carribe..." started showing clipped against the right edge.

nickidlugash:
> The tile containing the center-point of the label would have to be loaded before a label could show

@ChrisLoer – @ajashton and I both think this behavior is fine 👍

ansis:
> The tile containing the center-point of the label would have to be loaded before a label could show

Yep, I agree. We already have this behavior for all non-centered labels so if we need to fix it we'll need to find a way that works for all labels.

ChrisLoer:
OK, I think this is ready to go pending review/edits, then!

@samanpwbb do you or anyone from Studio want to weigh in before we publish this? I think from a Studio UI point of view this should be pretty simple, my only concern would be that using `symbol-placement: line-center` effectively requires more awareness/coupling with the underlying data than just using `symbol-placement: line`, so it has a bit of that "advanced feature" smell.

src/style-spec/reference/v8.json
```diff
@@ -861,6 +861,9 @@
         },
         "line": {
           "doc": "The label is placed along the line of the geometry. Can only be used on `LineString` and `Polygon` geometries."
+        },
+        "line-center": {
+          "doc": "The label is placed at the center of the line of the geometry. Can only be used on `LineString` and `Polygon` geometries. Note that a single feature in a vector tile may contain multiple line geometries. The center is calculated based on the entirety of the geometry included in the tile buffers, and the label may extend past the edge of the tile."
```
nickidlugash: @samanpwbb you may want to review this description, since Studio will display this text for the property value tooltip.

src/symbol/get_anchors.js
```diff
@@ -8,7 +8,68 @@ import checkMaxAngle from './check_max_angle';
...
+        lineLength += line[k].dist(line[k + 1]);
+    }
+    return lineLength;
+}
+
+function getAngleWindowSize(shapedText: ?Shaping,
+                            glyphSize: number,
+                            boxScale: number): number {
+    return shapedText ?
+        3 / 5 * glyphSize * boxScale :
```
mollymerp: a comment explaining the magic 3/5 number might be a good idea 🔢

ChrisLoer: @ansis or @nickidlugash do you know the origin of this magic number? I guess I just assumed it was something "experimentally reasonable".

ChrisLoer:
Good idea on unit tests for `getCenterAnchor`, I'll put some in.

src/style-spec/reference/v8.json
```diff
@@ -861,6 +861,9 @@
         },
         "line": {
           "doc": "The label is placed along the line of the geometry. Can only be used on `LineString` and `Polygon` geometries."
+        },
+        "line-center": {
```
samanpwbb: Can we add more specific compatibility info? See
https://github.com/mapbox/mapbox-gl-js/blob/a8cb4c48fb66868ccca00309da30518cf7637eef/src/style-spec/reference/v8.json#L1035-L1040 for example.

src/style-spec/reference/v8.json
samanpwbb: This description is a little on the long side. Is the last sentence necessary?

src/style-spec/reference/v8.json
ChrisLoer: Good idea -- this look right?

screenshot 2018-07-03 14 00 05

ChrisLoer:
Oops, yeah, thanks for catching that @andrewharvey! https://github.com/mapbox/mapbox-gl-js/pull/6943

PR #6828, opened 2018-06-18
Fix flyTo not zooming to exact given zoom

### LINKED ISSUE #6191: flyTo doesn't reach integral zoom needed for cluster expansion

### Steps to Trigger Behavior

 1. Start with a map with clusters and map.getZoom() == 3
 2. map.flyTo({zoom: 4})

### Expected Behavior

1. Some clusters expand.
2. Math.floor(map.getZoom()) == 4

### Actual Behavior

1. Clusters remain the same after zoom.
2. Math.floor(map.getZoom()) == 3
3. map.getZoom() == 3.999999999999999

mourner:
Currently, if you call `flyTo` with an integer `zoom`, you're not guaranteed to land on the exact zoom after the animation because of floating point errors. E.g. `map.flyTo({zoom: 2})` when starting on zoom 0.6 will end up on zoom `1.999999999996`, which is problematic if you wish to get on the exact zoom for corresponding tiles to load. Discovered this while implementing the `getClusterExpansionZoom` API.

## Launch Checklist

 - [x] (n/a) ~~document any changes to public APIs~~
 - [x] (n/a) ~~post benchmark scores~~

closes #6191

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -852,7 +852,7 @@ class Camera extends Evented {
             // s: The distance traveled along the flight path, measured in ρ-screenfuls.
             const s = k * S;
             const scale = 1 / w(s);
-            tr.zoom = startZoom + tr.scaleZoom(scale);
+            tr.zoom = k === 1 ? zoom : startZoom + tr.scaleZoom(scale);
 
             if (this._rotating) {
                 tr.bearing = interpolate(startBearing, bearing, k);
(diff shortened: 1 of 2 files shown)


andrewharvey:
@mourner I edited your comment to include closes #6191, since I think it needs to be in the top comment to auto close when merged.

mourner:
@andrewharvey great, I didn't realize that was already reported.

@mollymerp I couldn't come up with a comment that would make the lines clearer...

mollymerp:
oh I was just thinking something like "avoid floating point errors by using zoom directly when (whatever k represents)`k===1`"

PR #6829, opened 2018-06-18
Add methods for inspecting GeoJSON clusters

### LINKED ISSUE #3318: See features in cluster
Hello! 
I know this question was asked several times already, but there is still no workaround!
Using clustering is the only solution in case you have a big amount of pins on map. Unfortunately Supercluster allows you only to "aggregate" points in clusters, but there is no way to return the features inside the cluster back. This makes current clustering algorithm useless for the end user. Simple example: ten points with the same location are clustered...or the ten points with slightly different locations are clustered - how can I get the list of points inside the cluster on click? 
I do understand that potentially it can be millions of points inside the cluster...so I can suggest a few solutions:

1) Have a look at Leaflet Clustering ([screenshot](https://camo.githubusercontent.com/e0d878ddb6a4900830ba2ff5b19a41173c1d2c78/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f34353734302f313933303335392f34336466353066362d376561612d313165332d393463312d3466636131373063316166312e706e67) ) - it calculates the approximate bounds of the cluster - it can be the solution to do the same, at least developers will be able to query the DB to return the original points...

2) Or you can store the ID's of the original points in each cluster and even limit the number to 100 for example - in most cases it will be more then enough - devs can at least code something like "if num_of_points_in_cluster > 100, then Zoom In the map else return IDs"

Really hope you will be able to find ANY solution, though the current clustering algorithm is useless and potentially buggy...

mourner:
Closes #3318. Introduces 3 new methods to clustered GeoJSON sources:

* `getChildren` for returning cluster's immediate children (from the next zoom)
* `getLeaves` for returning original points that belong to the given cluster, with pagination support.
* `getClusterExpansionZoom` for returning the zoom the cluster expands on.

```js
map.on('click', 'cluster', (e) => {
  var clusterId = e.features[0].properties.cluster_id;

  map.getSource('geojson').getClusterExpansionZoom(clusterId, (err, zoom) => {
    ...
  });
});
```

Misc. open questions: 

- [x] Is it worth adding unit tests for this, and if so, how? The methods are just a thin layer between the GL JS API and the Supercluster library, and the functionality is tested there directly. We could add tests similar to those for `GeoJSONSource#setData`, but they don't actually test the roundtrip — they just mock the dispatcher `send` method, so for this case I doubt such tests will be useful.
- [ ] Is the minimal flow typing in `geojson_source_worker` enough, or should we try doing something elaborate (e.g. setting up Supercluster/GeoJSON-VT typings, casting the index to either etc.)? Currently `GeoJSONIndex` interface in the master branch is empty and typing checks for it don't run at all for some reason — e.g. there's `this._geoJSONIndex.getTile(...)`, this method isn't typed anywhere but Flow doesn't complain.
- [x] Is it worth packing the features through vt-pbf when sending them back? In most use cases, the payload will be tiny (a few point features), so getting maximum performance didn't seem necessary here, so for now I'm just sending GeoJSON directly. Should we leave it like this and get a round of feedback before trying to optimize?

## Launch Checklist

 - [x] n/a ~~post benchmark scores~~

### DIFF
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -192,12 +192,43 @@ class GeoJSONSource extends Evented implements Source {
     /**
      * For clustered sources, fetches the zoom at which the given cluster expands.
      *
-     * @param clusterId the value of the cluster's `cluster_id` property.
+     * @param clusterId The value of the cluster's `cluster_id` property.
      * @param callback A callback to be called when the zoom value is retrieved (`(error, zoom) => { ... }`).
      * @returns {GeoJSONSource} this
      */
     getClusterExpansionZoom(clusterId: number, callback: Function) {
         this.dispatcher.send('geojson.getClusterExpansionZoom', { clusterId, source: this.id }, callback, this.workerID);
+        return this;
+    }
+
+    /**
+     * For clustered sources, fetches the children of the given cluster on the next zoom level (as an array of GeoJSON features).
+     *
+     * @param clusterId The value of the cluster's `cluster_id` property.
+     * @param callback A callback to be called when the features are retrieved (`(error, features) => { ... }`).
+     * @returns {GeoJSONSource} this
+     */
+    getClusterChildren(clusterId: number, callback: Function) {
+        this.dispatcher.send('geojson.getClusterChildren', { clusterId, source: this.id }, callback, this.workerID);
+        return this;
+    }
+
+    /**
+     * For clustered sources, fetches the original points that belong to the cluster (as an array of GeoJSON features).
+     *
+     * @param clusterId The value of the cluster's `cluster_id` property.
+     * @param limit The maximum number of features to return.
+     * @param offset The number of features to skip (e.g. for pagination).
+     * @param callback A callback to be called when the features are retrieved (`(error, features) => { ... }`).
+     * @returns {GeoJSONSource} this
+     */
+    getClusterLeaves(clusterId: number, limit: number, offset: number, callback: Function) {
+        this.dispatcher.send('geojson.getClusterLeaves', {
+            source: this.id,
+            clusterId,
+            limit,
+            offset
+        }, callback, this.workerID);
         return this;
     }
 
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -175,7 +175,8 @@ class GeoJSONSource extends Evented implements Source {
         this.fire(new Event('dataloading', {dataType: 'source'}));
         this._updateWorkerData((err) => {
             if (err) {
-                return this.fire(new ErrorEvent(err));
+                this.fire(new ErrorEvent(err));
+                return;
             }
 
             const data: Object = { dataType: 'source', sourceDataType: 'content' };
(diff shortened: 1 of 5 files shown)


src/source/geojson_source.js
```diff
@@ -189,6 +189,49 @@ class GeoJSONSource extends Evented implements Source {
...
     }
 
+    /**
+     * For clustered sources, fetches the zoom at which the given cluster expands.
+     *
+     * @param clusterId The value of the cluster's `cluster_id` property.
+     * @param callback A callback to be called when the zoom value is retrieved (`(error, zoom) => { ... }`).
+     * @returns {GeoJSONSource} this
+     */
+    getClusterExpansionZoom(clusterId: number, callback: Function) {
```
jfirebaugh: Type the callbacks more precisely, e.g. `Callback<number>`.

src/source/geojson_source.js
```diff
@@ -189,6 +189,49 @@ class GeoJSONSource extends Evented implements Source {
...
+    }
+
+    /**
+     * For clustered sources, fetches the children of the given cluster on the next zoom level (as an array of GeoJSON features).
+     *
+     * @param clusterId The value of the cluster's `cluster_id` property.
+     * @param callback A callback to be called when the features are retrieved (`(error, features) => { ... }`).
+     * @returns {GeoJSONSource} this
+     */
+    getClusterChildren(clusterId: number, callback: Function) {
```
jfirebaugh: `Callback<Array<GeoJSONFeature>>`

src/source/geojson_worker_source.js
```diff
@@ -37,6 +37,12 @@ export type LoadGeoJSONParameters = {
 export type LoadGeoJSON = (params: LoadGeoJSONParameters, callback: Callback<mixed>) => void;
 
 export interface GeoJSONIndex {
+    getTile(z: number, x: number, y: number): GeoJSON;
+
+    // supercluster methods
+    getClusterExpansionZoom(clusterId: number): number;
+    getChildren(clusterId: number): GeoJSON[];
```
jfirebaugh: We're using the `Array<GeoJSON>` convention rather than `GeoJSON[]`.

src/source/geojson_worker_source.js
```diff
@@ -37,6 +37,12 @@ export type LoadGeoJSONParameters = {
 export type LoadGeoJSON = (params: LoadGeoJSONParameters, callback: Callback<mixed>) => void;
 
 export interface GeoJSONIndex {
+    getTile(z: number, x: number, y: number): GeoJSON;
```
jfirebaugh: Let's replace `export type GeoJSON = Object` above with types from [`flow-typed/geojson.js`](https://github.com/mapbox/mapbox-gl-js/blob/20c0d13b9bb412fc18a00404212de06337350711/flow-typed/geojson.js).

lewis500:
Is there any good way to find the cluster_id associated with a given piece of data?

I have a map with clusters on it. When people search for something and select it, I want to highlight which cluster it's a part of.

andrewharvey:
> Is there any good way to find the cluster_id associated with a given piece of data?
> I have a map with clusters on it. When people search for something and select it, I want to highlight which cluster it's a part of.

@lewis500 I guess you could use `map.querySourceFeatures` to get all the `cluster_id`'s of visible clusters, then do `getClusterLeaves` on each to build an index of features => cluster_id's. Certainly not ideal and only works for clusters in the view, otherwise you'll probably need to use private undocumented methods and properties.

PR #6834, opened 2018-06-19
use absolute URLs for example assets for portability

### LINKED ISSUE #6511: using relative paths for assets in examples makes them less portable
The [3D indoor mapping](https://www.mapbox.com/mapbox-gl-js/example/3d-extrusion-floorplan/) example does not work locally out of the box. 

As mentioned on [SO](https://stackoverflow.com/questions/42096737/extrude-polygons-for-3d-indoor-mapping-in-mapbox/49815556#49815556) it results in this error:
`Failed to load file:///mapbox-gl-js/assets/indoor-3d-map.geojson: Cross origin requests are only supported for protocol schemes:`

The line causing this CORS issue is [here](https://github.com/mapbox/mapbox-gl-js/blob/13ea16df5eabc4c8867d69e81da180cdcd6cc6b2/docs/pages/example/3d-extrusion-floorplan.html#L20).

A simple workaround might be to use inline json?

andrewharvey:
## Launch Checklist

closes #6511

Examples are designed to be copy pasted, but all the asset URLs were relative which most of the time means the examples didn't work when running locally or on jsbin, etc.

I'd still like the ability for new examples being developed before being published to be run so don't think we should simply hardcode the absolute URLs. This PR instead transforms the assets URLs to be absolute for the copy paste code only (not the code that's driving the live example).

It looks like the ACAO header is present on the assets now, so I couldn't see any issues.

### DIFF
--- a/docs/components/example.js
+++ b/docs/components/example.js
@@ -53,10 +53,7 @@ export default function (html) {
 </head>
 <body>
 
-${html
-        .replace("<script>", `<script>\nmapboxgl.accessToken = '${this.state.token}';`)
-        .replace("/mapbox-gl-js/assets/", "https://www.mapbox.com/mapbox-gl-js/assets/")
-}
+${html.replace("<script>", `<script>\nmapboxgl.accessToken = '${this.state.token}';`)}
 </body>
 </html>`;
         }
--- a/docs/components/example.js
+++ b/docs/components/example.js
@@ -53,7 +53,10 @@ export default function (html) {
 </head>
 <body>
 
-${html.replace("<script>", `<script>\nmapboxgl.accessToken = '${this.state.token}';`)}
+${html
+        .replace("<script>", `<script>\nmapboxgl.accessToken = '${this.state.token}';`)
+        .replace("/mapbox-gl-js/assets/", "https://www.mapbox.com/mapbox-gl-js/assets/")
+}
 </body>
 </html>`;
         }
--- a/docs/pages/example/3d-extrusion-floorplan.html
+++ b/docs/pages/example/3d-extrusion-floorplan.html
@@ -17,7 +17,7 @@ map.on('load', function() {
             // GeoJSON Data source used in vector tiles, documented at
             // https://gist.github.com/ryanbaumann/a7d970386ce59d11c16278b90dde094d
             'type': 'geojson',
-            'data': '/mapbox-gl-js/assets/indoor-3d-map.geojson'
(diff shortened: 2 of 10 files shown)


pathmapper:
https://www.mapbox.com/mapbox-gl-js/assets/radar.gif

from https://www.mapbox.com/mapbox-gl-js/example/image-on-a-map/

failed to load using the absolute URL because of 
`No 'Access-Control-Allow-Origin' header is present on the requested resource.`

Example:
https://jsbin.com/coyegamoyo/edit?html,output

jfirebaugh:
It's working for me. You sure you don't have a cached version?

```
$ curl -s -I -H'Origin: http://example.com/' https://www.mapbox.com/mapbox-gl-js/assets/radar.gif | grep Access-Control
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, HEAD
Access-Control-Max-Age: 3000
Vary: Origin, Access-Control-Request-Headers, Access-Control-Request-Method
```

pathmapper:
@jfirebaugh it's working now, I think you are right and it was caused by a cached version.

ryanhamley:
I was able to recreate the CORS error but the headers indicated that it was being served up from the CDN so it was definitely just a cache issue. I think this is 👍

andrewharvey:
> Rather than do this as a search and replace though, let's just use absolute URLs in the source.

I did that originally, but swapped for the search and replace because:

1. `/mapbox-gl-js/assets/` is highly qualified so low risk of catching something by mistake.
2. It allows examples to be run at any given git tag or commit without depending on published assets, like while developing. 
3. Helps allow docs to be run on other domains like a staging or testing without cross linking assets.

Are you sure you'd prefer hard coding the full URL everywhere?

jfirebaugh:
The problem with search and replace is that there's no hint in the example code itself that the source is going to be modified; it's hidden magic. One likely pitfall is that someone will use the full URL in an example, the `/mapbox-gl-js/assets/` portion of the path within it will get prefixed with `https://www.mapbox.com`, and then they'll have a mysteriously broken URL.

andrewharvey:
Okay I've swapped it to hard coding `https://www.mapbox.com`.

People developing new examples can still use a relative path while developing then swap over to the full url before pushing, and anyone pushing docs to another domain or staging area can add their own search and replace on the mapbox domain.

andrewharvey:
PS. assets like https://www.mapbox.com/mapbox-gl-js/assets/us_states.geojson don't get served with gzip encoding, could that be enabled?

It would help people on slow connections in particular, as currently they might think the example is broken when it actually loading still.

PR #6835, opened 2018-06-19
[WIP] open popup on a marker from keyboard

andrewharvey:
## Launch Checklist

allows popups bound to markers to be opened via the keyboard
todo
 - n/a post benchmark scores
 - n/a tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec changes
 - [ ] doesn't do anything without a tabindex on the element so it can be active via the keyboard, need to think about how that should play into this PR


src/ui/marker.js
```diff
@@ -289,6 +289,14 @@ export default class Marker extends Evented {
...
             this._popup = popup;
             if (this._lngLat) this._popup.setLngLat(this._lngLat);
+
+            this._element.addEventListener('keypress', ((e) => {
+                const code = e.charCode || e.keyCode;
+
+                if ((code === 32) || (code === 13)) { // space or enter
+                    this.togglePopup();
+                }
+            }).bind(this));
```
mourner: No need to `bind` because this is an arrow function.

mourner:
> doesn't do anything without a tabindex on the element so it can be active via the keyboard, need to think about how that should play into this PR

Leaflet sets `tabIndex` to `'0'` for markers by default so that they're accessible, with an ability to turn it off in options. Maybe we should do the same?

ryanhamley:
@mourner @andrewharvey It looks like if a popup is already bound to the marker, it's being removed when `setPopup` is called again in line 270. We could just move the listener function into a variable then call `this._element.removeEventListener('keypress', listenerFn);` inside that if statement to avoid the listeners piling up.

src/ui/marker.js
andrewharvey: Thanks! I originally used a regular function and needed bind, but swapped to the arrow function when I was reminded of the style guide by lint, good to know arrow functions don't need to bind this.

mourner:
> PS. This PR is still very much a WIP.

@andrewharvey Any updates on this one?

andrewharvey:
> @andrewharvey Any updates on this one?

No sorry, I haven't been able to prioritise picking this up again, but still keen on seeing a solution.

asheemmamoowala:
@andrewharvey Still WIP?

andrewharvey:
Thanks for the reminders @mourner @asheemmamoowala. I've taken another look at this, and tested it out again, and it's ready for review/merge now please.

src/ui/marker.js
```diff
@@ -277,9 +279,26 @@ export default class Marker extends Evented {
...
+            const code = e.code;
+            const legacyCode = e.charCode || e.keyCode;
+
+            if (
+                (code === 'Space') || (code === 'Enter') ||
+                (legacyCode === 32) || (legacyCode === 13) // space or enter
+            ) {
+                this.togglePopup();
+            }
+        };
```
mourner: For consistency, can we move this into its own `_onKeyPress` method (and add it to `bindAll` in the constructor)?

andrewharvey:
> As far as I remember, generic click event is triggered when you press either enter or space when the element is focused — the only thing it needs is a tabIndex: 0. E.g. that's how Leaflet handles marker keyboard events. So do we have to explicitly put keypress listeners then?

Probably, but because of :point_down: there is no `click` listener on the element.

https://github.com/mapbox/mapbox-gl-js/blob/8f0998dc447c81de6322b79b0430dfc3ff1b1ca3/src/ui/marker.js#L207-L210


For that reason the popup's aren't being activated on enter or space, so we workaround the fact that we're not using a `click` listener on the element by watching for a space or enter keypress.

src/ui/marker.js
andrewharvey: I could, but for reasons I don't understand I need to do:

```js
this._element.addEventListener('keypress', (e: KeyboardEvent) => { this._onKeyPress(e) });
```

as the following doesn't work:

```js
this._element.addEventListener('keypress', this._onKeyPress);
```

is that okay?

andrewharvey: Actually, forget that, I must have been doing something wrong, it's working now.

mourner:
> If we attached the `click` listener to the marker element, the popup 
 // would close once the event propogated to `map` due to the 
 // `Popup#_onClickClose` listener.

mourner:
> If we attached the `click` listener to the marker element, the popup would close once the event propogated to `map` due to the `Popup#_onClickClose` listener. 

Hmm... I wonder if we could try fixing the root problem here instead of the hacky workaround? E.g. involving `stopPropagation` to prevent the click propagating to map?

andrewharvey:
> when tabbing through markers with a keyboard, if a marker is not fully within the current view, the canvas container will scroll to accommodate it (default focus action), breaking the map. We should add some code to prevent this (likely involving focus event and e.preventDefault).

Good catch. Since [focus is not cancelable](https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event) I couldn't work out how to stop the scroll.

I was able to get the map to `panBy` just enough to move the element into view on `focusin` so that it no longer to scrolls the page to fit the Marker as the Marker is now already within view. Although this does change behaviour, I think this is reasonable.

If that's okay, I suppose I should extend that to work for the left, top and right borders too to be consistent.

There's still a question on if we should disable this for Markers completely outside the view, but I'm hesitant to disable that since tab cycling through Markers, I think should continue even if outside the viewport, so then that would mean the map might pan a lot to get these Markers into view, but I think that's okay.

andrewharvey:
... experimenting with ^ in https://github.com/mapbox/mapbox-gl-js/commit/d1a2a5b8eae87da7a701264e109c6fda73a95ad3 but it does have some downsides when the Marker is larger than the canvas.

mourner:
@andrewharvey the focus event is not preventable, but we might be able to prevent the scrolling action itself — e.g. there's this `preventScroll` polyfill we can take some clues from: https://github.com/calvellido/focus-options-polyfill

andrewharvey:
Thanks for the tips, I investigated and as far as I can tell, we can only use preventScroll option or the polyfill when manually triggering a focus with `Element.focus()`, you can't do that when the browser has invoked the `focus` event, since there's no way to preventDefault when a `focus` event is triggered we can't seem to stop the browser scrolling.

The polyfill grabs the window.scrollY then sets the window.scroll to the same, but that doesn't help here (at least for full screen maps) because the page isn't actually scrolled, and scrollY is still 0 before and after the focus.

Screenshot from 2019-10-17 12-23-40

It feels like a browser bug since it should respect the absolute position of the map at 0,0 regardless if we focus on a child that's positioned outside...

Maybe there is a solution, but I can't work it out.

mourner:
OK, I think this is ready to merge:

- Rebased on master and fixed linting issues.
- Found a way to fix the scrolling issue by resetting the container scroll position on focus.
- Made it so that the markers aren't focused when you click on them — the outline may be distracting for non-keyboard users and this seems to be consistent with how browsers behave with native elements like button (not focusing unless you explicitly tab to it).

mourner:
Merged — let's follow up with any remaining issues about keyboard accessibility of markers in other issues/PRs.

andrewharvey:
@mourner That's smart, I didn't think of setting the scroll of the map Element itself. Thank you.

Issue #5768, opened 2017-11-29
Feature request: make fill-extrusion vertical gradient configurable


nmccrea-db:
Hey folks. Thanks for making such an amazing platform.

I'm using **mapbox-gl-js 0.42.2**. I am representing a stack of blocks on a map using fill extrusions. The problem I'm having is that the lowest blocks are shaded slightly darker than the blocks above them. Here is an example:

<img width="1041" alt="funky shading 01" src="https://user-images.githubusercontent.com/32711966/33390183-2b65ff8a-d503-11e7-86f6-d2573ef64c1b.png">

In this example, all four blocks that are pointed out are given the same paint color (in this case, `white`). But the lowest two are rendered slightly darker than the higher two. Above the first two levels, there are no other inconsistencies.

It appears to have something to do with the elevation of the blocks. For example, if I make the bottom-most block thicker, and thus raise the elevation of the second block, then the inconsistency only affects the first block:

<img width="1045" alt="funky shading 02" src="https://user-images.githubusercontent.com/32711966/33390305-7e9caa6e-d503-11e7-9d94-6636e59c30d5.png">

The problem exists regardless of what color paint is being used:

<img width="493" alt="screen shot 2017-11-29 at 1 04 02 pm" src="https://user-images.githubusercontent.com/32711966/33391069-e5e2848a-d505-11e7-8348-49bed0e7ad3e.png">

And regardless of opacity:

<img width="524" alt="screen shot 2017-11-29 at 1 03 40 pm" src="https://user-images.githubusercontent.com/32711966/33391084-f3e951b2-d505-11e7-8810-0748399cbdd1.png">

<img width="489" alt="screen shot 2017-11-29 at 1 03 27 pm" src="https://user-images.githubusercontent.com/32711966/33391095-f964b758-d505-11e7-8d4d-24f330b6b52e.png">

This is in a geoJSON layer. The geoJSON is produced like this:

```javascript
      {
        'type': 'FeatureCollection',
        'features': geometries.map( geometry => ({
          'type': 'Feature',
          'properties': {
            'height':       geometry.height + geometry.baseElevation,
            'base_height':  geometry.baseElevation + VISUAL_SPACE_BETWEEN_BLOCKS,
            'color':        colorForType(geometry.type)
          },
          'geometry': {
            'type': 'Polygon',
            'coordinates': [
              geometry.polygon.concat([geometry.polygon[0]]),
            ],
          },
        }))
      }
```

[...]

PR #6841, opened 2018-06-20
Add fill-extrusion-vertical-gradient property

lbud:
This PR adds a new boolean data-constant paint property, `fill-extrusion-vertical-gradient` ( 🤔 a bit verbose — I'm open to better ideas). It defaults to `true` (the current behavior), but if set to `false`, does not apply a vertical gradient along the sides of fill-extrusion features.

Closes https://github.com/mapbox/mapbox-gl-js/issues/5768.

I only ran benchmarks for `LayerFillExtrusion` since nothing else should be affected, and it shows no real change: https://bl.ocks.org/lbud/raw/d353b63e34c7fcb6e4ec2698370fd6b9/

## Launch Checklist

 - [x] document any changes to public APIs (docs page update will be generated in release)
 - [x] post benchmark scores
 - [x] tag @mapbox/studio and/or `@mapbox/maps-design` if this PR includes style spec changes


ryanhamley:
Is @mapbox/studio ok with including this? Seems like this is a desired option.

elifitch:
In general we're a little concerned about the spec getting bloated, but we're good with this addition; it'll be easy enough for us to support

ryanhamley:
I rebased this PR against `master` and updated it to use uniform bindings. I think it's good to go; the test results look good.

screen shot 2018-09-24 at 2 13 21 pm

src/render/program/fill_extrusion_program.js
```diff
@@ -24,7 +24,8 @@ export type FillExtrusionUniformsType = {|
     'u_matrix': UniformMatrix4f,
     'u_lightpos': Uniform3f,
     'u_lightintensity': Uniform1f,
-    'u_lightcolor': Uniform3f
+    'u_lightcolor': Uniform3f,
+    'u_vertical_gradient': Uniform1f
```
mollymerp: this is just 1 or 0 right? in that case we can use an integer uniform `Uniform1i` to follow the convention of other uniforms used as booleans such as `u_is_zoom_constant` for symbol icons/sdfs

src/render/program/fill_extrusion_program.js
```diff
@@ -80,7 +84,8 @@ const extrusionTextureUniforms = (context: Context, locations: UniformLocations)
 
 const fillExtrusionUniformValues = (
     matrix: Float32Array,
-    painter: Painter
+    painter: Painter,
+    verticalGradient: number
```
mollymerp: For clarity, I lean towards passing this value to the `*UniformValues` functions as a boolean, and casting only as the value is bound to its uniform location.

src/shaders/fill_extrusion.vertex.glsl
```diff
@@ -47,7 +48,11 @@ void main() {
 
     // Add gradient along z axis of side surfaces
     if (normal.y != 0.0) {
-        directional *= clamp((t + base) * pow(height / 150.0, 0.5), mix(0.7, 0.98, 1.0 - u_lightintensity), 1.0);
+        // This avoids another branching statement, but multiplies by a constant of 0.84 if no vertical gradient,
```
mollymerp: do we have any idea where the magic -0.84 comes from? maybe nicki would know

ryanhamley: why would Nicki know? i have no idea where it comes from

src/render/program/fill_extrusion_program.js
ryanhamley: `Uniform1i` requires a number input so casting the boolean to a number only when binding to the location causes a Flow error. Is there a way around that?

src/shaders/fill_extrusion.vertex.glsl
mollymerp: Nicki worked on the extrusion shaders w lbud

ryanhamley: @nickidlugash do you remember where this magic number came from?

src/render/program/fill_extrusion_program.js
mollymerp: yep you use the `+` like here https://github.com/mapbox/mapbox-gl-js/blob/master/src/render/program/symbol_program.js#L116

ryanhamley: Ah yeah, I was mistakenly casting to a number elsewhere in the program, causing a flow error. This is fixed now.

src/render/program/fill_extrusion_program.js
ryanhamley: This actually needs to be a float because the calculation in the shader has float constants.

ryanhamley:
I updated the way the gradient is calculated (removed the 0.84 constant) to allow the full spectrum of light intensity to be used in determining the final color.

screen shot 2018-09-25 at 2 49 30 pm

PR #6846, opened 2018-06-21
add snapshot example

mollymerp:
this replaces #6518

@samsammurphy thanks for your contribution! I had to open a new PR because #6513 caused some of your work to break, and I didn't want to bother you further with more back and forth (I can't push to your branch). I also changed the styling around a bit and simplified the code a bit because we try to keep our examples as minimal as possible. 

we'll have to be careful with this example because it will break if we change our CSS for the logo control at all 😕 

## Launch Checklist

### DIFF
--- a/docs/pages/example/snapshot.html
+++ b/docs/pages/example/snapshot.html
@@ -41,7 +41,7 @@
     <div id="container">
       <div id="map"></div>
       <div>
-        <h3 id="instructions">Drag and pan the map on the left to change the snapshot area, then click to generate a static image on the left</h3>
+        <h3 id="instructions">Drag and pan the map on the left to change the snapshot area, then click to generate a static image on the right.</h3>
         <canvas id='snapshot'></canvas>
       </div>
     </div>
--- a/docs/pages/example/snapshot.html
+++ b/docs/pages/example/snapshot.html
@@ -7,45 +7,44 @@
     height: 100%;
   }
 
-  h3 {
-    font-size: 16px;
-    line-height: 20px;
-    font-family: "Open Sans", sans-serif;
-    margin: 5px 10px;
-    color: white;
-    padding: 10px;
-    background: rgba(100,100,100, 0.4);
+  #container {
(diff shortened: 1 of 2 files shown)


mollymerp:
_before clicking_
image

_after clicking_
image

ryanhamley:
Sorry to do a little rewrite on the CSS. I think it looks good. Super cool demo. 👍

mollymerp:
@jfirebaugh I agree this is a bit too brittle to be comfortable with, but the `LogoControl` is private now and I wasn't sure we would want to change that  – do you suggest we make `map.logoControl` public (we explicitly made it private in the past to encourage TOS compliance) or have `getLogoImage` on the `Map` object?

andrewharvey:
While I agree relying on internal implementation details isn't a good idea, I think it's better to include this example as is and leave the door open to change/refine it later on.

That aside, any kind of implementation of the Logo and Attribution, even a no so ideal one makes complying with the TOS much easier.

> // Firefox requires SVG to have height and width specified
> // in the root SVG object when drawing to canvas

Can we not just add the `width` and `height` to the [`mapboxgl-ctrl-logo.svg`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/css/svg/mapboxgl-ctrl-logo.svg)?

ryanhamley:
I'm going to close this PR because we're uncomfortable with some of the implementation details here. @ChrisLoer and I are hoping to work on a video export feature soon and that could probably expose a snapshot method as well then we can revisit adding examples for these features.

andrewharvey:
This has been handy in a few cases to demonstrate how to take a static snapshot of a map, hope it or a reincarnation of it does eventually make it into the examples at some point.

andrewharvey:
@ryanbaumann I know some of the implementation details here quite deep, but I still find myself digging up this PR to use as a reference for new implementations.

Even if it gets changed later from a future video export feature or refactor, my 2c are it would be still useful for others to have this published in the meantime.

Issue #6839, opened 2018-06-20
Memory leak calling map.setFeatureState() frequently [0.46-beta1]


ryanbaumann:
**mapbox-gl-js version**: 0.46.0-beta1

**browser**: `Chrome v67.0`

**OS**: Windows 10

**GPU**: Geforce GTX 660 with latest driver 398.11

### Steps to Trigger Behavior

 1. Load the [example](https://bl.ocks.org/ryanbaumann/733ba99c5ca1d9d15259081b395e4b00/34d3bb174e9ac3b5e0c7f2d0cc3c41ed139849b7) in Chrome on a Windows 10 device using a Nvidia GPU (GTX GeForce 660 series with latest drivers).
 2. Move the mouse cursor repeatedly over features in the map using `map.setFeatureState()` on a vector tile source to change the hovered feature's style.  Do not zoom or pan the map.  
3. Each map.setFeatureState() call, the browser memory allocation increases.  If you call setFeatureState() enough before zooming to new tiles, the browser crashes.

### Link to Demonstration

* https://bl.ocks.org/ryanbaumann/733ba99c5ca1d9d15259081b395e4b00/34d3bb174e9ac3b5e0c7f2d0cc3c41ed139849b7

* [Video](https://www.dropbox.com/s/xn6bfdk4yx7i6pv/feature-state-crash-chrome-windows10-180620.mp4?dl=0)

* [chrome://gpu report from machine](https://gist.github.com/ryanbaumann/dd00b48287ab5aa93e6a34dc254461a0)

* [Chrome performance profile](https://www.dropbox.com/s/lcz8bwlitgqywkm/chrome-profile-mapboxgl-180620.zip?dl=0)

* ![](https://cl.ly/3F0U3a0G182P/download/Image%202018-06-20%20at%209.53.54%20AM.png)

### Expected Behavior

* Memory is reclaimed / minimally expanded for each new call of `map.setFeatureState`

### Actual Behavior

* Memory increases proportionately for each `map.setFeatureState()` call triggered
* Memory is mostly reclaimed when loading and rendering a new zoom level of tiles

cc/ @asheemmamoowala

ChrisLoer:
@ryanbaumann Looking at that performance profile, the thing that solidly trends up is "Nodes", which I think is DOM nodes -- makes me suspect the actual problem is all the popups getting created. Can you remove the popup part of the code to see if you can still reproduce the leak?

[...]

PR #6853, opened 2018-06-22
Use updateData instead of re-creating buffers for repopulated paint arrays

ChrisLoer:
Fixes issue #6839 (memory leak on frequent calls to setFeatureState). The leak was caused by mismatched calls to "createVertexBuffer" and "destroy". We didn't see the leak on most machines, presumably due to differences in how different GPU drivers handled re-creating the buffers. @ryanbaumann could reproduce the problem on one of his Windows machines, and verified the problem went away with these changes.

Improves performance of PaintStates benchmark. The performance improvement is presumably because we're able to take advantage of the DYNAMIC_DRAW option intended for frequently updated buffers. I ran the benchmarks multiple times and saw the same pattern: it looks like the PaintStates benchmark has much less variability after this change (and it looks like before the change there was actually a pattern in which progressive iterations of the bench seemed to slow down -- note the way the dots slope up to the right).

screenshot 2018-06-22 13 11 18

There are a couple things that bothered me a little but don't seem too serious:

- Duplication between CompositeExpressionBinder and SourceExpressionBinder
- Peaking inside the `VertexBuffer` to test if it's "live" (e.g. `this.paintVertexBuffer.buffer`) -- it kind of feels like this should be handled by `VertexBuffer` somehow. If `buffer` is null, that means `destroy()` has already been called, and it needs to be re-recreated (I don't actually know of a pathway that would cause this to happen).
- We call `updateData` regardless of whether DYNAMIC_DRAW is enabled (it's cued off `expression.isStateDependent`). But I believe `updateData` is still fine if it ends up getting called on a STATIC_DRAW buffer (should be no worse than destroying and re-creating the buffer).

## Launch Checklist

 - [ ] ~~write tests for all new functionality~~ (no explicit behavior change)
 - [x] post benchmark scores

/cc @asheemmamoowala @ryanbaumann @mollymerp @ansis

### DIFF
--- a/src/data/program_configuration.js
+++ b/src/data/program_configuration.js
@@ -185,7 +185,11 @@ class SourceExpressionBinder<T> implements Binder<T> {
 
     upload(context: Context) {
         if (this.paintVertexArray && this.paintVertexArray.arrayBuffer) {
-            this.paintVertexBuffer = context.createVertexBuffer(this.paintVertexArray, this.paintVertexAttributes, this.expression.isStateDependent);
+            if (this.paintVertexBuffer && this.paintVertexBuffer.buffer) {
+                this.paintVertexBuffer.updateData(this.paintVertexArray);
+            } else {
+                this.paintVertexBuffer = context.createVertexBuffer(this.paintVertexArray, this.paintVertexAttributes, this.expression.isStateDependent);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


asheemmamoowala:
> I can imagine cases where we might occasionally want to update a buffer but still expect it to be rare enough that STATIC_DRAW was more appropriate than DYNAMIC_DRAW

The difference between those buffer creation modes is driver dependent. Without profiling individual drivers it is not very useful to try an optimize the use of `STATIC_DRAW` vs `DYNAMIC_DRAW` especially given that we won't always know _if_ the runtime APIs will actually be used to update buffers, and what each drivers optimization for this looks like.

PR #6860, opened 2018-06-25
Fix popup shaking

hyperknot:
Removed round() from transform: translate.

This round() was causing the random shaking after camera animations (panBy, panTo, easeTo, etc.)

I believe there is no need to have this rounded, the browser will handle it. In current Chrome this gets rounded to 3 decimal digits.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -314,7 +314,7 @@ export default class Popup extends Evented {
             }
         }
 
-        const offsetedPos = pos.add(offset[anchor]).round();
+        const offsetedPos = pos.add(offset[anchor]);
 
         DOM.setTransform(this._container, `${anchorTranslate[anchor]} translate(${offsetedPos.x}px,${offsetedPos.y}px)`);
         applyAnchorClass(this._container, anchor, 'popup');


jfirebaugh:
As mentioned in the issue, this `round()` is intentional -- we found that not all browsers handle it in all situations, sometimes making the popup borders and content non-pixel aligned.

In order to remove it, we would need to to a comprehensive round of browser testing and validate that all supported browsers no longer show this behavior.

hyperknot:
Leaflet has been using fractional `transform: translate3d` values with ancient browsers, is translate handled that much different compared to `translate3d`?

If we cannot get rid of round, our only option is to round to 0.1 first and then round to integer values. That way a jumping 0.499 vs 0.5 wont make a full pixel jump at least.

hyperknot:
Actually doing a rounding to 0.1 before would be the same as moving the values by -0.05.
Since the jumping happens only at very close to 0.499-0.501 values moving it away might be a good solution actually.

hyperknot:
OK, that would just make different cases broken. 

The only proper solution then - if rounding is required - is to store the initial fractional part of the position using a debounced function, and substract it from the new position till the end of the animation.

hyperknot:
I've tested, and on pixel level, even Chrome and Firefox does make slight changes when the div is not aligned to integer pixels. I'm closing this.

Would you be interested in a debounce based pull request?

jfirebaugh:
By debounce, do you mean short-circuiting `panBy([0,0])`? I think investigating what's going on there would be promising.

hyperknot:
No, I mean in popup's _update.

What I'm seeing is that the fluctuations are very small, on the range of 0.001 pixels.

What I'm thinking of doing is:
1. Save the fractional parts of _pos in the first _update event of an animation. I'd need debounce to detect this.
2. subtract the fractional parts at the end of calculation, just before rounding.

In theory this would make the fluctuations very close to integer values.

p.s.: panBy([0,0]) is equal to `map.easeTo({center: map.getCenter(),})` which is what introduces these fluctuations.

jfirebaugh:
No, I don't think that approach would be worth the added complexity. I'd rather figure out how to short-circuit what should be no-op pans.

hyperknot:
That seems very simple, why not just limit pan to 1 pixel or bigger movements? The question is whether anyone would like to pan by fractional pixels?

PR #6868, opened 2018-06-26
Rotated filtered symbol sort test

ChrisLoer:
This PR basically just adds a symbol sorting test designed to catch https://github.com/mapbox/mapbox-gl-native/issues/12104. However, because of precision issues, I had to make a small change to get native and JS to have the same results for this (previously untested) case.

Our symbol sorting algorithm (only used for figuring out how to display overlapping symbols) is "sort by y-position first, and if the y-position is the same, sort by the order the underlying feature appeared in the data". Before, y-position could only be "the same" if it was _exactly_ the same, which meant that due to precision issues there could be small differences in order between gl-native and gl-js. This PR rounds to the nearest integer (in tile coordinates) to solve that problem.

/cc @julianrex @asheemmamoowala

### DIFF
--- a/src/data/bucket/symbol_bucket.js
+++ b/src/data/bucket/symbol_bucket.js
@@ -671,8 +671,8 @@ class SymbolBucket implements Bucket {
         symbolInstanceIndexes.sort((aIndex, bIndex) => {
             const a = this.symbolInstances[aIndex];
             const b = this.symbolInstances[bIndex];
-            const aRotated = (sin * a.anchor.x + cos * a.anchor.y) | 0;
-            const bRotated = (sin * b.anchor.x + cos * b.anchor.y) | 0;
+            const aRotated = Math.round(sin * a.anchor.x + cos * a.anchor.y) | 0;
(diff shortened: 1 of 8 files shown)


asheemmamoowala:
@ChrisLoer on low zoom tiles could this result in different labels taking precedence than previously? Would it adversely affect existing styles?

ChrisLoer:
I like where your head's at @asheemmamoowala, overzooming and underzooming are always waiting to make fools out of us! 😉 

These values are in tile units (e.g. 0-8192). Assuming 512px to a tile at base zoom, that means you'd have to be overzoomed by a factor of 16 (4 zoom levels) until the units reached par with a pixel. Beyond that point you could potentially have a situation where a symbol one pixel above another symbol was treated as being at "the same" y position, and we'd switch to sorting based on feature order in the data, which might swap the order of symbols for overlapping purposes.

Which is to say: technically, yes this makes a difference for existing styles, but it should be trivial to the point of being difficult to detect.

Issue #6876, opened 2018-06-27
feature-state expression is always evaluated to null in filter


stepankuzmin:
Consider this example (https://jsbin.com/rogidup/9/edit?js,output):

```js
map.addLayer({
  id: 'points',
  type: 'circle',
  source: 'points',
  paint: {
    'circle-color': '#fff'
  },
  filter: ["feature-state", "display"]
});

map.setFeatureState({ id: 1, source: 'points' }, { display: true });
```

It seems, that `featureState` is `undefined` when evaluating `feature-state` expression context in `filter`. So `feature-state` expression is always evaluated to `null` in `filter`.

https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/expression/definitions/index.js#L175

**mapbox-gl-js version**: v0.46.0

**browser**: Google Chrome 67.0.3396.87

### Steps to Trigger Behavior

 1. Set feature state
 2. Filter features according to this state

### Link to Demonstration

https://jsbin.com/rogidup/9/edit?js,output

### Expected Behavior

Features can be filtered according to their state

### Actual Behavior

Features can't be filtered according to their state

asheemmamoowala:
@stepankuzmin Thank you for using Mapbox. The `feature-state` expression was added specifically to enable interactivity with data-driven styling. Can you speak to more of the use case you are expecting to use the `feature-state` expression in a layer filter?

stepankuzmin:
Thanks for the reply, @asheemmamoowala!

The ability to filter features based on their state seems consistent to me.

Imagine the following use case. I've got a set of features on the map and I want to add values to them from some external API. But this API may not have some values for specific features and it seems logical to filter them out.

asheemmamoowala:
@stepankuzmin `feature-state` expressions should be supported only in the data-driven styling enabled paint properties. Adding support for them in filter expressions, while possible, is outside of their intended use case. Maybe you could use the layer's `-opacity` property with a `feature-state` expression to get the desired filtering effect?

strech345:
@asheemmamoowala 
setting `opacity` per feature is not possible for `fill-extrusion`. So in this case only filter is possible to hide elements. 
And i don't understand why the join example now change to feature state if this should only be used for interaction.  https://www.mapbox.com/mapbox-gl-js/example/data-join/
If it is now the prefered way to use it also for joining data, filter should be also worked - from my side.

[...]

PR #6878, opened 2018-06-27
WIP: add feature-state expression support in filter #6876

stepankuzmin:
I've started digging into #6876 issue, but I'm not sure, is this a right direction.

### DIFF
--- a/src/style-spec/feature_filter/index.js
+++ b/src/style-spec/feature_filter/index.js
@@ -2,8 +2,9 @@
 
 import { createExpression } from '../expression';
 
+import type {FeatureState} from '../';
 import type {GlobalProperties} from '../expression';
-export type FeatureFilter = (globalProperties: GlobalProperties, feature: VectorTileFeature) => boolean;
+export type FeatureFilter = (globalProperties: GlobalProperties, feature: VectorTileFeature, featureState: FeatureState) => boolean;
 
 export default createFilter;
 export { isExpressionFilter };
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Thanks for the PR @stepankuzmin -- however, there are some architectural fundamentals that make supporting `feature-state` for filters problematic. In order to make `setFeatureState` fast and synchronous, it does its work on the main thread. But filter evaluation happens on a web worker at layout time. So right now they are not a good fit for each other.

stepankuzmin:
Oh Okay! Thanks for the reply, @jfirebaugh!

annehulsey:
At @jfirebaugh do you have a suggestion for filtering polygons based on a cursor hover? 

Fill-extrusions don't allow for data driven opacity so I need another way to make objects appear or disappear based on user input. Since I can't use setFeatureState for the expression and it also doesn't work as a filter, I looked into joining a dummy attribute to the data, which could then be used for the filter. However, from what I can tell, data joins only work for expressions that drive the layer symbology. Which leads back to the original problem of fill-extrusions not allowing data driven opacity.

Suggestions would be much appreciated. (I want to use the mouse hover to overlay a transparent black on the hovered building in [this map](https://annehulsey.github.io/SF_Tall_Buildings/). In theory, temporarily darkening the building under the cursor will reduce confusion about which building's info is being displayed on the left.)

asheemmamoowala:
@annehulsey Check out [this](https://github.com/mapbox/mapbox-gl-js/blob/master/debug/extrusion-query.html) demo to see how the `fill-extrusion-color` can be changed on hover. Alternatively, you can copy the hovered over geometry to a separate layer and apply a translation.

PR #6880, opened 2018-06-27
Upgrade test/integration to ES6 modules

### LINKED ISSUE #6563: Upgrade integration test suite to ES2015 and ES6 modules
The [integration suite code](https://github.com/mapbox/mapbox-gl-js/tree/master/test/integration/lib) follows different conventions than the rest of the codebase: it doesn't use any ES2015 syntax or ES6 modules (via esm). This is because this code needs to support being run in node 4, which is part of gl-native's CI suite, because it's what's currently used for server-side rendering.

Once we upgrade node on the server, then we can remove node 4 from the gl-native matrix, and then we can make the test suite follow the same conventions as the rest of the codebase.

jfirebaugh:
Fixes #6563. Not passing over in mapbox-gl-native yet due to a variety of issues.


mourner:
What steps are left on the native side?

jfirebaugh:
Spent some more time on this today and everything is now working, except there are some expression test failures in native due to needing to port #6961.

PR #6886, opened 2018-06-28
[WIP] fixes #6884 Option to only enable zooming by scrolling while holding ctrl

### LINKED ISSUE #6884: Allowing "zoom by scrolling while holding ctrl key" interaction
## Motivation

Currently the `scrollZoom` option turns off `onWheel` handler. This prohibits not only zooming by scrolling but zooming by pinching on a touchpad, and also zooming by scrolling while pressing CTRL key.

## Design

We would need an option to override scroll zooming while CTRL key is pressed, in which case it would be fully backward compatible.

### Mock-Up

The option to achieve the behaviour that disables zooming while scrolling but enables zooming  with scrolling while holding the CTRL key (or pinching on touchpad). So the example below would turn scrolling off, but would allow ctrl + scroll.

```javascript
const mapOptions = {
  scrollZoom: {
   ctrl: true
  }
};
```

This change might also need some kind of UI change, like GoogleMaps does on a touch device. But it's not necessary IMO.

### Implementation

This all should be changed on the client side, by extending the `ScrollZoomHandler` class. It bares the `_enabled` property currently, but would need a `_ctrlEnabled` property too. The name could be changed tho.

In which case the `onWheel` handler would need an extra condition, something like this:

```javascript
// scroll_zoom.js

onWheel(e: WheelEvent) {
  if (!this.isEnabled() && !(this.isCtrlEnabled() && e.ctrlKey)) return;

  // ...
}
```

On how to handle this property I'm not quite sure, please advise.

abelvarga:
## Launch Checklist

Fixes #6884 

While enabling `ScrollZoomHandler`, if an option `Object` is passed, it'll look for the `ctrl` property which if set to `true`, will only allow zooming with scrolling while CTRL key is down.

Some TODO still left, did not want to write documentation until general acceptance.

### DIFF
--- a/src/ui/handler/scroll_zoom.js
+++ b/src/ui/handler/scroll_zoom.js
@@ -31,6 +31,7 @@ class ScrollZoomHandler {
     _map: Map;
     _el: HTMLElement;
     _enabled: boolean;
+    _ctrlEnabled: boolean;
     _active: boolean;
     _aroundCenter: boolean;
     _around: Point;
--- a/src/ui/handler/scroll_zoom.js
+++ b/src/ui/handler/scroll_zoom.js
@@ -95,6 +95,7 @@ class ScrollZoomHandler {
      * Enables the "scroll to zoom" interaction.
      *
      * @param {Object} [options]
+     * @param {string} [options.ctrl] If scrolling zooming should work while holding ctrl key
      * @param {string} [options.around] If "center" is passed, map will zoom around center of map
      *
      * @example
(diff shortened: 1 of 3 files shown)


ansis:
@abelvarga sorry about not getting back to you earlier. It looks like this slipped through the cracks.

> Is this design even right?

Yes, it looks good.

>  And would be a cover image needed to get this feat in?

I think so! showing text across the map when scrolling without the modifier would be good.

- should we use `command` on mac instead of `ctrl`?
- if we do ^, should we rename to option to something more general? `withModifierKey`?
- eventually we might add two-finger panning on mobile to address this same issue. What would we name that option? Is there a name that could work well for both that and this?

@ryanhamley any opinions on these questions?

@abelvarga I'm happy to help wrap up this pr if you no longer have time or interest.

ryanhamley:
> should we use command on mac instead of ctrl?

I don't know. That would add complexity because you'd have to listen for the various browser dependent keycodes for `command` instead of just listening for `e.ctrlKey`.

`withModifierKey` might be more future-proof and clear regardless.

> eventually we might add two-finger panning on mobile to address this same issue. What would we name that option? Is there a name that could work well for both that and this?

Something like `modifyScroll`?

abelvarga:
Hi @ansis, thanks for getting back. I almost forgot about this PR.

I currently have little time, as we moved forward in our application, but would like to see this feature get going. So I'm glad to take you up on your offer.

Also, this [reply](https://github.com/mapbox/mapbox-gl-js/issues/6884#issuecomment-428778923) from @nigelsim on #6884 might need some consideration.

vakila:
:wave: Hi folks, just jumping in here to check in as part of some PR gardening we're doing. 

It looks like there hasn't been much movement on this, and since it's still labeled WIP and @abelvarga doesn't have time to work on it further, I think it makes sense to close this for now. Also we may be making some changes to our overall interaction architecture, so might make sense to reevaluate this as/after that work is completed. @ansis @arindam1993 if one of you wants to pick up work on this in the meantime, we can continue to push changes to this branch and re-open the PR down the line when it's ready for review. 

Hope that sounds OK, and if I've misunderstood and there actually has been progress on this please re-open/let me know! :heart:

moll:
I think it certainly requires some solution. Last I checked, it made the user experience rather abysmal as Mapbox hijacked scrolling down a page.

olivierhill:
Agreed, even more the code change is even attached to this. Is there anything we can help with?

PR #6894, opened 2018-06-29
Add CameraForBoxAndBearing function

taraadiseshan:
fixes https://github.com/mapbox/mapbox-gl-js/issues/4221

Working with @ChrisLoer on adding CameraForBoxAndBearing() and fitScreenCoordinates() to have box zoom behavior without reset map bearing. We tested this manually on the debug page and ran the unit and integration tests. Still working on tests for new functionality and documentation for changes. 

## Launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## master
+### ✨ Features and improvements
+* Add `Map#fitScreenCoordinates`: fits viewport to two points, similar to `Map#fitBounds` but uses screen coordinates and supports non-zero map bearings. ([#6894](https://github.com/mapbox/mapbox-gl-js/pull/6894))
+
+
 ## 0.47.0
 
 ## ✨ Features and improvements
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -357,9 +357,9 @@ class Camera extends Evented {
 
     /**
      * @memberof Map#
-     * @param bounds Calculate the center for these bounds in the viewport and use
+     * @param {LatLngBoundsLike} bounds Calculate the center for these bounds in the viewport and use
      *      the highest zoom level up to and including `Map#getMaxZoom()` that fits
-     *      in the viewport.
+     *      in the viewport. LatLngBounds represent a box that is always axis-aligned with bearing 0.
      * @param options
      * @param {number | PaddingOptions} [options.padding] The amount of padding in pixels to add to the given bounds.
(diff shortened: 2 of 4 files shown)


ChrisLoer:
Thanks @taraadiseshan, this is awesome!

I followed up with some small stuff:
 - Rebased onto the current version of master
 - Added documentation and changelog entry
 - Added unit tests to camera.js. I started down the path of making a render test, but the unit test ended up being easier to write and I think it offers pretty good coverage. I bootstrapped the expected values by running the test once, although at a glance they look like they match my expectations (e.g. the finished zoom is 1.5x when the destination is rotated 225 degrees, but 2x when kept at bearing 0).

🤔 At first I was thinking that we didn't need to document the change to `BoxZoomHandler` as a breaking change, because it seems to me the new behavior is obviously what people would expect...

@mourner you might be able to make suggestions about whether this is a good way to expand the API?

- The `fitScreenCoordinates` interface is the easiest fit for `BoxZoomHandler`, and we didn't want to touch the existing `LatLngBounds`-based `fitBounds`. I can imagine wanting something that's `LatLng` based but still allows you to control the target bearing (think ["fit to kenya"](https://www.mapbox.com/mapbox-gl-js/example/fitbounds/) but with bearing 180 because we want to use south-up map orientation). We weren't eager to add a _third_ method just to explicitly support that, though.
- `cameraForBoxAndBearing` doesn't _need_ to be public, it's just following the example of `cameraForBounds`, and it's nice that someone could use `cameraForBoxAndBearing` to implement their own bearing-supporting variant of `fitBounds`. _BUT_ it's adding another method to the Map object and we probably want to be careful about how many methods we end up with.

ChrisLoer:
Yeah, I think we can figure out better terminology, but I don't have a great idea yet. "bounds" is already taken to mean (aligned-to-north) "LatLngBounds", and I don't think we want to change that.

Two points and a bearing _do_ define a specific box -- once you're axis-aligned to the bearing, any two points fit to a single box (right?). If you added more points, that would introduce the possibility of the points not fitting _any_ box (or at least not to the corners). But maybe the right way to talk about this is something that sounds more like "fit these two points (to a box)".

`cameraForTwoPointsAndBearing`? 🤷‍♂️

asheemmamoowala:
> proposed something like "fit to a geojson feature" as a more natural API for that.

I started work on a `fitToFeature`  method in #6208. It doesn't yet account for `bearing` and `pitch`.

PR #6909, opened 2018-07-03
Set tile#queryPadding per layer

### LINKED ISSUE #6896: Regression: mouseenter event is fired on wrong layer
**mapbox-gl-js version**: v0.46.0

**browser**: Chrome on Linux

Beginning with v.0.45.0 a `mouseenter` event is fired on the wrong layer if two layers are using the same GeoJSON source and the `mouseenter` listener is on the second layer. 

### Steps to Trigger Behavior

 1. Add two line layers using the same source to a map.
 2. Set up a `mouseenter` listener on the second layer to change the `line-color` of this layer.
 3. Hover over the second layer.

### Link to Demonstration
https://jsbin.com/resezeyosu/edit?html,output

### Expected Behavior
`line-color` of the second layer changes when hovering over the second layer. 

Demonstration with v0.44.2: https://jsbin.com/bicatukoxa/edit?html,output
### Actual Behavior
`line-color` of the second-layer changes when hovering over the first layer

ryanhamley:
## Launch Checklist

    - fixes #6896 
    - `queryPadding` was always being calculated based on the first layer in a bucket which resulted in a bug in which two layers using the same source could introduce the bug seen in the ticket
    - this PR changes `queryPadding` to be calculated for each layer in a bucket


src/source/tile.js
```diff
@@ -434,7 +435,8 @@ class Tile {
 
             bucket.update(sourceLayerStates, sourceLayer);
             if (painter && painter.style) {
-                this.queryPadding = Math.max(this.queryPadding, painter.style.getLayer(bucket.layerIds[0]).queryRadius(bucket));
+                const index = bucket.layerIds.indexOf(id);
+                this.queryPadding = Math.max(this.queryPadding, painter.style.getLayer(bucket.layerIds[index]).queryRadius(bucket));
```
mourner: `layerIds[layerIds.indexOf(id)]` is simply `id`, so no need to do `indexOf` lookups in both places.

ryanhamley:
The query test isn't working as is @ansis. I committed it just to have my most up to date work in the PR. I haven't written one before and was confused about some of the behavior and what would be the best way to write it. I'm definitely open to suggestions on the best way to test this.

I assumed that `actual.json` would have `state: { "hover": true }` but state is actually an empty object. I'm guessing that I'm not fully understanding the nature of writing a query test.

ryanhamley:
@ansis I finally got around to fixing the query test for this. It's just checking that a feature is found under the query geometry in the case where you have two features in the same bucket. I verified that this test is correct by reverting the fix and confirming that the test fails. So I think this PR is good to go now.

PR #6918, opened 2018-07-05
Clamp latitude value to valid Web Mercator range when projecting to y

### LINKED ISSUE #6906: fitBounds with -90 results in uncaught error
This must have been introduced in a recent version but calling fitBounds with a bounding box containing `-90` results in an uncaught error.

screen shot 2018-07-03 at 9 37 20 am

**mapbox-gl-js version**:
`v0.46.0`

**browser**:
Chrome `Version 67.0.3396.99 (Official Build) (64-bit)`

### Steps to Trigger Behavior

Call `fitBounds` with a sw latitude coordinate of `-90`

### Link to Demonstration
https://jsfiddle.net/tristen/bytzwu08/ Click "Fit bounds" and view the console

ryanhamley:
Fixes #6906 

## Launch Checklist

    - clamps latitude values to valid Web Mercator range
    - handles large, wrapped longitude values by reducing them to a +-180 range and avoiding world spin animation

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -21,6 +21,7 @@ class Transform {
     tileZoom: number;
     lngRange: ?[number, number];
     latRange: ?[number, number];
+    maxValidLatitude: number;
     scale: number;
     width: number;
     height: number;
@@ -47,6 +48,7 @@ class Transform {
 
     constructor(minZoom: ?number, maxZoom: ?number, renderWorldCopies: boolean | void) {
         this.tileSize = 512; // constant
+        this.maxValidLatitude = 85.051129; // constant
 
         this._renderWorldCopies = renderWorldCopies === undefined ? true : renderWorldCopies;
         this._minZoom = minZoom || 0;
@@ -284,7 +286,7 @@ class Transform {
     get point(): Point { return new Point(this.x, this.y); }
 
     /**
-     * latitude to absolute x coord
+     * longitude to absolute x coord
      * @returns {number} pixel coordinate
      */
     lngX(lng: number) {
@@ -295,6 +297,7 @@ class Transform {
      * @returns {number} pixel coordinate
      */
     latY(lat: number) {
+        lat = clamp(lat, -this.maxValidLatitude, this.maxValidLatitude);
         const y = 180 / Math.PI * Math.log(Math.tan(Math.PI / 4 + lat * Math.PI / 360));
         return (180 - y) * this.worldSize / 360;
     }
@@ -433,7 +436,7 @@ class Transform {
             this._constrain();
         } else {
             this.lngRange = null;
-            this.latRange = [-85.05113, 85.05113];
+            this.latRange = [-this.maxValidLatitude, this.maxValidLatitude];
         }
     }
 
--- a/test/unit/ui/camera.test.js
+++ b/test/unit/ui/camera.test.js
@@ -503,7 +503,7 @@ test('camera', (t) => {
             const camera = createCamera({ zoom: 0 });
             camera.rotateTo(90, { around: [5, 0], duration: 0 });
             t.equal(camera.getBearing(), 90);
-            t.deepEqual(fixedLngLat(camera.getCenter()), fixedLngLat({ lng: 4.999999999999972, lat: 0.000014144426558004852 }));
+            t.deepEqual(fixedLngLat(camera.getCenter()), fixedLngLat({ lng: 4.999999999999972, lat: 0.000002552471840999715 }));
             t.end();
         });
 
@@ -519,7 +519,7 @@ test('camera', (t) => {
             const camera = createCamera({ zoom: 0 });
             camera.rotateTo(90, { offset: [100, 0], duration: 0 });
             t.equal(camera.getBearing(), 90);
-            t.deepEqual(fixedLngLat(camera.getCenter()), fixedLngLat({ lng: 70.3125, lat: 0.000014144426558004852 }));
+            t.deepEqual(fixedLngLat(camera.getCenter()), fixedLngLat({ lng: 70.3125, lat: 0.000002552471840999715 }));
             t.end();
         });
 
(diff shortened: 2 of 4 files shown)


hyperknot:
That should take care of latitude (as long as clamp works correctly with Infnity as well).

Longitude on the other hand is tricky:
- It should support wrapped bounds
- Yet, if a too big lon value is given, it starts a crazy world rotating animation.
Like:
```
fitBounds([0,-40,10000,40])
```

Actually, fitBounds([0,-40,10000,40]) is wrong for an other reason: it zooms out the map, thus breaking the lat constraints. I believe it should do a mod 360 for exampe.

My idea is that for longitude, there should be some kind of reasonable clamping so that Infinity works, as well as a mod 360 operation.

src/geo/transform.js
```diff
@@ -295,6 +295,12 @@ class Transform {
      * @returns {number} pixel coordinate
      */
     latY(lat: number) {
+        const validLatitude = 85.05113;
```
asheemmamoowala: Declare as static constant outside this method, maybe in `util.js` or on `LngLat`

hyperknot: Actually, you have `this.latRange = [-85.05113, 85.05113];` in the same file, as well as couple of `-85.0511` values in style-spec/reference/v8.json.

ryanhamley: Good catch. I'll refactor these references to a single place.

hyperknot: You probably cannot refactor the json one, but maybe it's a good idea to have the same precision level in all 3 places (that is, probably stick with `-85.0511`)

mourner:
I recently tried to tackle the same longitude issue described by @hyperknot in Supercluster — perhaps you could reuse some of that logic: https://github.com/mapbox/supercluster/blob/master/index.js#L72-L84

src/geo/transform.js
ryanhamley: I don't think it should be declared in LngLat because the way I understand it is that the LngLat is based around EPSG84 coordinates which are valid up to +-90° so the issue is more in projecting from those coordinates to valid Web Mercator ones. I'm leaning towards using the `latRange` method in Transform.

asheemmamoowala: Using `latRange` would prevent general transform/projection calls from using otherwise valid calls with `LatLng` outside the bounds specified in `Map#setMaxBound`. Im not sure that would be clear/expected.

ryanhamley: Ahhhhh, yeah I was misunderstanding what `latRange` was. Thanks for the clarification.

hyperknot:
Ah, I looked into this, the silly negative modulo of JS really makes it complicated. Picking @mourner's solution verbatim is probably the best idea. 

(BTW,  why the 180 special case? This way 180 maps to 180 and 540 maps to -180. Was the original aim to make min between -180 .. 179.99 and max between -179.99 .. 180?)

src/geo/transform.js
```diff
@@ -295,6 +297,9 @@ class Transform {
      * @returns {number} pixel coordinate
      */
     latY(lat: number) {
+        if (lat > this.maxValidLatitude || lat < this.maxValidLatitude * -1) {
```
hyperknot: Is there any point in using an `if` here? Why not just call clamp directly?

ryanhamley: That's probably fine. Originally, I had a console warning inside the if statement but I think we've decided to favor documentation over a warning so this can be removed.

hyperknot:
Is there anything blocking this? I thought copying @mourner's snippet from supercluster would solve longitude. We just need to fix the 540 bug.

ryanhamley:
No, I just have been pushing to get some other stuff finished and this was put on the back burner. I can finish up this PR for just the latitude fix if you're interested in putting together a PR for the longitude fix separately @hyperknot. Otherwise, it will probably be next week before I finish this.

hyperknot:
I'm happy to do a PR for lon fix from @mourner's snippet.

hyperknot:
@ryanhamley I think @mourner's code is having two cases to make sure that -180 gets mapped to -180 when it's the min side and and +180 to +180 when it's the max side.

src/geo/transform.js
```diff
@@ -47,12 +48,13 @@ class Transform {
...
     constructor(minZoom: ?number, maxZoom: ?number, renderWorldCopies: boolean | void) {
         this.tileSize = 512; // constant
+        this.maxValidLatitude = 85.051129; // constant
 
         this._renderWorldCopies = renderWorldCopies === undefined ? true : renderWorldCopies;
         this._minZoom = minZoom || 0;
         this._maxZoom = maxZoom || 22;
 
-        this.setMaxBounds();
+        this.latRange = [this.maxValidLatitude * -1, this.maxValidLatitude];
```
mourner: Let's move this back to `setMaxBounds` to make sure the behavior is consistent (and `setMaxBounds(null)` resets the map to the state it was by default).

Also nit: `-this.maxValidLatitude` is more readable than `this.maxValidLatitude * -1`.

ryanhamley:
> I'm now hesitating about the longitude wrapping — it can bring some unintended consequences. Maybe let's leave that for a different PR?

@mourner Agreed. I'm reverting that change and will just close out the original latitude clamping issue with this PR. I think there's potential to combine longitude wrapping with #6985 in a single PR.

ryanhamley:
@mourner I think this PR is good to go from my standpoint if you can look it over and give it a 👍 if it looks good.

test/unit/ui/camera.test.js
```diff
@@ -503,7 +503,7 @@ test('camera', (t) => {
             const camera = createCamera({ zoom: 0 });
             camera.rotateTo(90, { around: [5, 0], duration: 0 });
             t.equal(camera.getBearing(), 90);
-            t.deepEqual(fixedLngLat(camera.getCenter()), fixedLngLat({ lng: 4.999999999999972, lat: 0.000014144426558004852 }));
+            t.deepEqual(fixedLngLat(camera.getCenter()), fixedLngLat({ lng: 4.999999999999972, lat: 0.000002552471840999715 }));
```
mourner: Quick question — why is value here different? Latitude is close to 0 so isn't supposed to be affected by clamping here...

ryanhamley: This is caused by a slight change in the `maxValidLatitude` value from `85.05113` to `85.051129`. Should I revert to the older value?

Issue #6782, opened 2018-06-06
Fixing matrix errors


hyperknot:
Failed to invert matrix errors are quite bad, as they get the UI in a "stuck" state, where no interaction have any effect. Effectively the map part of the application is frozen.

I've been debugging how do they happen using Sentry in a live production site, as otherwise I couldn't reproduce it on my machine, but they happen very regularly to users (across all kind of browsers).

There are multiple ways a map can get into this illegal state, here is the most common one:

1. **scroll_zoom / _onScrollFrame /  delta !== 0** is not run first. This means that `this._targetZoom` is not initialized.
https://github.com/mapbox/mapbox-gl-js/blob/b9e32bf6cebfe830bf6faa9be8635dc31709a4ff/src/ui/handler/scroll_zoom.js#L207

2. **interpolate** receives a value with `undefined` as the second parameter. The way interpolate is written, it'll turn this into `NaN`, even if k is 0. 
https://github.com/mapbox/mapbox-gl-js/blob/b9e32bf6cebfe830bf6faa9be8635dc31709a4ff/src/ui/handler/scroll_zoom.js#L235

3. **geo / transform / set zoom** has a guard for making sure a value is between min and max zoom, but this guard doesn't protect against NaN. 
https://github.com/mapbox/mapbox-gl-js/blob/b9e32bf6cebfe830bf6faa9be8635dc31709a4ff/src/geo/transform.js#L161

4. **NaN** goes through setZoom and just kills the matrix calculations.

Recommendations:

1. By far the safest and simplest option is to guard **set zoom**. This'll protect against all kind of errors, not just this.  I recommend adding an 

```
if (typeof zoom !== 'number' || !isFinite(zoom)) return
```

guard at the first line.

2. Interpolate could return a value without calculation if t is 0 or 1, both as an optimization as well as making it more reliable.

3. scroll_zoom should be fixed so that _targetZoom and _easing is always initialised.

@anandthakker @mourner @jfirebaugh

bravecow:
We are getting same errors after upgrade to v45:

![image](https://user-images.githubusercontent.com/988471/41221350-8b8b1164-6d6c-11e8-8f63-0da69009e2f8.png)

Downgrading to 44 😞

andrewharvey:
see also #6486

hyperknot:
@bravecow I think this was present in 0.44 as well, look at #6486. 
[...]

PR #6921, opened 2018-07-06
set zoom guard, fixing illegal matrix errors

hyperknot:
## Launch Checklist
 - [ ] post benchmark scores

No changes, no new functionality, just a guard to make sure that zoom cannot be set to NaN, and thus turn the map into a frozen state.

Detailed description about how can set zoom receive a NaN value:
https://github.com/mapbox/mapbox-gl-js/issues/6782

Those issues can be fixed in the future, if needed, but this one makes sure that the map never ends up in a broken "illegal matrix" state.

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -158,6 +158,7 @@ class Transform {
 
     get zoom(): number { return this._zoom; }
     set zoom(zoom: number) {
+        if (typeof zoom !== 'number' || !isFinite(zoom)) return;
         const z = Math.min(Math.max(zoom, this.minZoom), this.maxZoom);
         if (this._zoom === z) return;
         this._unmodified = false;


anandthakker:
Thanks for the PR @hyperknot (and sorry for the delay in responding to https://github.com/mapbox/mapbox-gl-js/issues/6782).  

My concern is that this fix will hide the underlying errors that result in zoom being set to a NaN/non-number value.  I'd be more in favor of making the check in `set zoom()` into an assert, i.e. `assert(typeof zoom === 'number' && isFinite(zoom))`, and fixing the root causes you identified in https://github.com/mapbox/mapbox-gl-js/issues/6782#issue-329845314

hyperknot:
Yes, I agree that the underlying cause needs to be fixed, but I wanted to make sure that the map is never frozen for any user. I'm running a custom build just for this reason.

How behaviour does a failed assert in core means for users? Everything keeps working and it's just an error in console? If so I'm happy to change the code.

anandthakker:
A failed assert would be equivalent to throwing an exception, so no, everything would not keep working -- but it would mean that "failed to invert matrix" would be replaced with a more precise error, thus making future bugs of this sort quicker to diagnose and fix.

I agree that the map shouldn't be frozen for any user, but I'm still not convinced that placing a blanket guard in `set zoom()` is a good idea.  It could hide other existing bugs or allow new ones to be introduced, making them subtler and more difficult to find.

The particular error from the ScrollZoom handler should really have been caught by Flow.  It wasn't, because `_targetZoom` is incorrectly typed as `number` instead of `?number`: https://github.com/mapbox/mapbox-gl-js/blob/b9e32bf6cebfe830bf6faa9be8635dc31709a4ff/src/ui/handler/scroll_zoom.js#L47

hyperknot:
Then it's up to you how do you see the timeline of properly fixing the handler bugs. My solution would be a temporary fix for user experience, until everything is polished out in scroll_zoom.

anandthakker:
Fixed in https://github.com/mapbox/mapbox-gl-js/pull/6924

Thanks for helping to move this forward, @hyperknot !

hyperknot:
@anandthakker so what's with transform.js? I feel it's important to put the guard in there, as there'll definitely be bugs in the future which can propaget to set zoom.

PR #6923, opened 2018-07-06
Adds warning to geolocate control when unsupported

aendrew:
## Launch Checklist

 - [x] post benchmark scores

This is a really basic pull request that adds a simple warning to the geolocate control when it's added to a map in a context that is unsupported.

Mainly submitting this because I couldn't figure out for the life of me why the control wasn't displaying, turns out I was serving the page over HTTP. Adding a warning informs the developer that this is indeed the expected behaviour.

### DIFF
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -259,7 +259,10 @@ class GeolocateControl extends Evented {
     }
 
     _setupUI(supported: boolean) {
-        if (supported === false) return;
+        if (supported === false) {
+            warnOnce('Geolocation not supported. The Mapbox geolocate control will not be rendered.');
+            return;
+        }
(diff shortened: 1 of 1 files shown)


src/ui/control/geolocate_control.js
```diff
@@ -259,7 +259,10 @@ class GeolocateControl extends Evented {
     }
 
     _setupUI(supported: boolean) {
-        if (supported === false) return;
+        if (supported === false) {
+            warnOnce('Geolocation not supported. The Mapbox geolocate control will not be rendered.');
```
andrewharvey: +1 for adding this warning, since the developer has asked for the GeolocateControl, we should at least warn when that hasn't happened.

Could we change the warning to `Geolocation support is not available, the GeolocateControl will not be visible.`?

aendrew: @andrewharvey Absolutely, will update that now.

PR #6924, opened 2018-07-06
Fix ScrollZoom handler setting tr.zoom = NaN

### LINKED ISSUE #6486: Failed to invert matrix error
The error still in version v0.44.2.

If it helps, I also get the same error on **random** occasions when making zoom in/out by using mouse scroll for instance. First is raised a `failed to invert matrix` error, and then, on every action in the map, for instance moving the mouse over, is raised this error: `Invalid LngLat object: (NaN, NaN)`. Call Stack for this last error: 
```
lng_lat.js:29 Uncaught Error: Invalid LngLat object: (NaN, NaN)
    at new LngLat (lng_lat.js:29)
    at Transform.coordinateLocation (transform.js:352)
    at Transform.pointLocation (transform.js:329)
    at e.unproject (map.js:559)
    at h (bind_handlers.js:155)
    at HTMLDivElement.a (bind_handlers.js:81)
```
Seems that the problem in this last error is when trying to convert the point to coordinate, in the line `return this.coordinateLocation(this.pointCoordinate(p))` the `this.pointCoordinate(p)` is returning `Coordinate {column: NaN, row: NaN, zoom: NaN}`. It is difficult to reproduce the error, because it is random, if any new detail I will update this post or create a new one.

Details of `failed to invert matrix` error: 

```
Uncaught Error: failed to invert matrix
    at Transform._calcMatrices (transform.js:549)
    at Transform.prototypeAccessors.zoom.set (transform.js:165)
    at ScrollZoomHandler._onScrollFrame (scroll_zoom.js:237)
    at e.i._updateCamera (camera.js:901)
    at e._render (map.js:1464)
    at map.js:1567
```

The line:
```
    // inverse matrix for conversion from screen coordinaes to location
    m = mat4.invert(new Float64Array(16), this.pixelMatrix);
    if (!m) throw new Error("failed to invert matrix");
    this.pixelMatrixInverse = m;
```
Question, what about if `this.pixelMatrix` isn't invertible? Has that any logic?

### LINKED ISSUE #6782: Fixing matrix errors

anandthakker:
Closes #6782
Closes #6486
Replaces #6921

### DIFF
--- a/src/ui/handler/scroll_zoom.js
+++ b/src/ui/handler/scroll_zoom.js
@@ -1,5 +1,6 @@
 // @flow
 
+import assert from 'assert';
 import DOM from '../../util/dom';
 
 import { ease as _ease, bindAll, bezier } from '../../util/util';
@@ -43,11 +44,11 @@ class ScrollZoomHandler {
     _lastWheelEvent: any;
     _lastWheelEventTime: number;
 
-    _startZoom: number;
-    _targetZoom: number;
+    _startZoom: ?number;
+    _targetZoom: ?number;
     _delta: number;
-    _easing: (number) => number;
-    _prevEase: {start: number, duration: number, easing: (number) => number};
+    _easing: ?((number) => number);
+    _prevEase: ?{start: number, duration: number, easing: (number) => number};
 
     _frameId: ?TaskID;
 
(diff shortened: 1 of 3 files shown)


hyperknot:
What about the `assert(typeof zoom === 'number' && isFinite(zoom))` line you recommended? That's the one which actually makes sure that illegal matrices do not happen.

PR #6925, opened 2018-07-06
Fix issue #6919: bring gl-js collision handling closer to gl-native

### LINKED ISSUE #6919: line labels with 0 "collision circles" still get placed
Discovered in the native port: the `symbol-placement/line-center-buffer` test case actually doesn't generate any collision circles because the logic in `CollisionFeature` aborts for line features that cross tile boundaries. The test case didn't notice on gl-js because there's nothing to collide against and gl-js treats "no collision circles" as "ok to place", while gl-native treats that as "not placeable".

/cc @nickidlugash

ChrisLoer:
Fixes #6919. 

- Don't place features that _have_ text but don't have collision boxes

This change partially rolls back https://github.com/mapbox/mapbox-gl-js/pull/6164 and replaces it with the equivalent logic on gl-native, which calculates `textWithoutIcon` per-symbolInstance instead of per-bucket. Placement is performance sensitive, but hopefully this still isn't too big a cost -- doesn't seem to show up in the benchmark at least.

- Create a single collision box even when the length of a line label is less than half the box size

Before this change the label "." wouldn't get a collision box because it was less than 1/2 a collision box wide. Although this can change label placement on existing maps, I expect the impact will generally be pretty limited since almost all labels are longer than the cutoff and the new behavior should conform to expectations better.

Test updates:
- Add debug collision circles to line-center and line-center-buffer to verify collision behavior
- Add regression test that exercises case where a line label is almost exactly the same length as its line

screenshot 2018-07-06 12 32 15

 - [x] post benchmark scores

/cc @ansis @nickidlugash @mollymerp


src/symbol/placement.js
```diff
@@ -256,6 +256,9 @@ export class Placement {
                     offscreen = offscreen && placedIconBoxes.offscreen;
                 }
 
+                const iconWithoutText = textOptional || (symbolInstance.numGlyphVertices === 0 && symbolInstance.numVerticalGlyphVertices === 0);
```
asheemmamoowala: These variable names tripped me up quite a bit. Would `textInstanceOptional` and `iconInstanceOptional` make more sense?

Issue #6944, opened 2018-07-10
`text-max-angle` check uses wrong scale for `symbol-placement: line-center`


ChrisLoer:
When we run the `checkMaxAngle` code, we pass in a `labelLength` that's supposed to be in tile geometry units, which is `shapingUnits * textBoxScale`. Instead, we pass the label length in plain shaping units, which makes the label seem much shorter than it really is (w/ a 16 point font, ~10x smaller). This means we only check the angle change for a small area right around the center of the label, so we miss sharp curves that happen further out.

We have a `text-max-angle/line-center` test, but I copied the 60 degree limit from `text-max-angle/literal` and the problem wasn't obvious in that case. With a 15 degree limit the problem manifests pretty clearly.

/cc @ansis @mollymerp

PR #6945, opened 2018-07-10
Use correct labelLength scale for text-max-angle/line-center

ChrisLoer:
Fixes issue #6944: text-max-angle check for `symbol-placement: line-center` was too permissive because it checked too short a section of the line.

/cc @nickidlugash @ansis @mollymerp @asheemmamoowala


ChrisLoer:
Ack, this PR needs to wait for https://github.com/mapbox/mapbox-gl-js/pull/6925 to go in first -- by fixing the `checkMaxAngle` test we also correctly determine that there's not enough room on the "Carribean Sea" line for its collision boxes (in `symbol-placement/line-center-buffer`).

ChrisLoer:
> we also correctly determine that there's not enough room on the "Carribean Sea" line for its collision boxes

It's a little bit worse than that. It turns out that checkMaxAngle` calculates its line length test a little differently from `collision_feature.js`, because the collision box starts counting from a slightly offset position:

https://github.com/mapbox/mapbox-gl-js/blob/5dd68bb38d59bd18d8b628641f1ffdfa5f55e3d6/src/symbol/collision_feature.js#L130-L152

vs.

https://github.com/mapbox/mapbox-gl-js/blob/5dd68bb38d59bd18d8b628641f1ffdfa5f55e3d6/src/symbol/check_max_angle.js#L28-L39

Ideally these would be perfectly in sync, but after #6925 it should be OK for one to be stricter than the other (depending on the label).

I adjusted the test to pass the `checkMaxAngle` length test and still cross the tile boundary by reducing the font size and shifting the label to the right a little bit.

src/symbol/get_anchors.js
```diff
@@ -59,12 +59,9 @@ function getCenterAnchor(line: Array<Point>,
 
             const anchor = new Anchor(x, y, b.angleTo(a), i);
             anchor._round();
-            if (angleWindowSize &&
-                !checkMaxAngle(line, anchor, labelLength, angleWindowSize, maxAngle)) {
-                return;
+            if (!angleWindowSize || checkMaxAngle(line, anchor, labelLength, angleWindowSize, maxAngle)) {
+                return anchor;
             }
```
ansis: Should this still always return once it reaches the middle segment?

ChrisLoer: Oh yeah, good point. I think without that return it's going to just waste cycles iterating to the end of the line.

ChrisLoer: Fixed.

PR #6949, opened 2018-07-11
popup: closeOnClick can be dynamic

magicdawn:
## Launch Checklist

 - [ ] post benchmark scores

## popup
get the `popup.options.closeOnClick` value when click happened. 
This makes the popup `closeOnClick` dynamic and can be changed later.

### DIFF
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -95,9 +95,7 @@ export default class Popup extends Evented {
     addTo(map: Map) {
         this._map = map;
         this._map.on('move', this._update);
-        if (this.options.closeOnClick) {
-            this._map.on('click', this._onClickClose);
-        }
+        this._map.on('click', this._onClickClose);
         this._update();
 
         /**
@@ -321,7 +319,9 @@ export default class Popup extends Evented {
     }
 
     _onClickClose() {
-        this.remove();
+        if (this.options.closeOnClick) {
+            this.remove();
+        }
     }
 }
 
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -323,6 +323,15 @@ export default class Popup extends Evented {
             this.remove();
         }
     }
+
+    getCloseOnClick() {
+        return this.options.closeOnClick;
+    }
+
+    setCloseOnClick(val: boolean) {
+        this.options.closeOnClick = val;
+        return this;
+    }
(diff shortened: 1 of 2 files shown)


andrewharvey:
I think convention is to use getters and setters rather than mutating properties directly, especially since `Popup.options` isn't part of the public API.

So this is only needed if you want to change the closeOnClick behaviour of an already open popup.

I'm curious what the use case is for needing to change it after the Popup is opened...?

magicdawn:
@andrewharvey 

I'm building an application that has a feature tracking mode. When in tracking mode, I want the popup shows always(closeOnClick = false), when exit tracking mode, use the default behavior(closeOnClick = true).

Anyway. It's easy and reasonable to make this option changeable/dynamic, and why not?

ryanhamley:
Hey @magicdawn thanks for submitting this. We'd definitely like to see a getter and setter method for this option. Otherwise, I think this is probably fine since it's not a breaking change in any way.

magicdawn:
ping, jsdoc & getter/setter added

ryanhamley:
Hey @magicdawn sorry for the delay getting back to you. This is looking really good. Can you please update the getter/setter methods to their ES6 equivalents and update the test to reflect this usage? Thanks!

magicdawn:
Hi @ryanhamley , do you mean `get closeOnClick() { ... }` / `set closeOnClick(){ ... }` ?

peterqliu:
this sounds like a reasonable feature addition, involving minimal addition to the codebase. but thinking about this more, we should strive to keep the API as lean and dry as possible until there's a need that it truly doesn't cover. There are existing ways to close popups based on a condition: to use the above use case

>When in tracking mode, I want the popup shows always(closeOnClick = false), when exit tracking mode, use the default behavior(closeOnClick = true).

since you're already tracking the `tracking` state, 

```
map.on('click', () => {
 if (!tracking) popup.remove()
})
```
would do the same thing, and can handle even more complex condition statements than a built-in method ever can.

In general, **it's best to track as much state as possible in the app**, outside the map. This keeps only one source of truth for the whole application, and reduces the chance for state to fall out of sync between the map and the broader application. Because of this, I think getters should return state objects outside the map as well.

Closing for now, but open to pushback here.

Issue #934, opened 2015-01-14
Implement cross-fading of vector tiles


mourner:
Currently the awesome smoothness of vector tile rendering in GL JS is hindered by new tiles popping up instantly, sometimes with an unpleasant effect (e.g. lots of labels suddenly appearing).

What does it take to implement cross-fading of vector tiles? Do we need #193 fist? @ansis

ljbade:
We should do something like goodle, they don't fade tiles in, but they must somehow pre-fetch tiles along the boundary of the map, and neighbouring zoom levels.

When you pan or zoom it always has some tile there (usually blurry) while detailed tile loads.

There map is certainly a lot more smooth feeling then ours.

mourner:
If you look closely, you can see that what Google prefetches are actually raster tiles. We can't do that — we need all the rendering to happen on the client with vectors. We had the notion of "pan tile" earlier to simulate the effect, but it was removed for a reason I can't remember.

I'm still interested in cross-fading tiles, so they don't just pop in instantly, but are faded in.

mourner:
Here's what I mean: ![gl-fade](https://cloud.githubusercontent.com/assets/25395/5755456/4e5d8e40-9cad-11e4-8c6d-4332893863df.gif)

ljbade:
Ah I see. Interesting that Google use a raster tile trick.

ansis:
Last week I was working on fading labels across multiple tiles (when a new tile loads), and on rotation. 

https://github.com/mapbox/mapbox-gl-js/tree/placement-fading

I'm putting it on hold for now because:
- it's blocked by having layer-by-layer rendering
- it's a big chunk of code that is mostly separate from actual label placement, so I should wrap that up first

How this currently works:
When labels are placed for a tile it creates a typed array of label positions. Each label position has (id, x, y, placementZoom). The array is sorted by id. This array is transferred to the main thread. In the main thread it then compares these label positions with positions from other tiles and figure out which ones were in one/the other/both. It figures out which labels should be shown and whether they need to be shown instantly or faded in/out. It updates the gl buffers with new the new opacities.

johnlaur:
[...]

Issue #5726, opened 2017-11-20
Implement fading for features that hit minzoom/maxzoom limits


ChrisLoer:
The `isHidden()` method of `StyleLayer` prevents us from drawing any layer that's outside its min/maxzoom limits (or has its visibility toggled to 'none'). Now that we're doing explicit time-based fade animations, it should be possible for us to animate these transitions -- we just need to (1) make sure the min/maxzoom calculation is fed into collision detection, and (2) update the drawing code to keep drawing any layers in the middle of fading (assuming this is a significant savings over just drawing the layer with nothing in it).

I noticed this while zooming in on a city label: at z14 in streets v10, `place-city-label-lg` will disappear abruptly. It doesn't look that bad, but it'd still be a nice improvement to fade it.

/cc @ansis @nickidlugash @jfirebaugh

ChrisLoer:
I'm looking at the native behavior that we haven't yet back-ported to JS, but with cross-fading between tiles, I think this behavior may have actually regressed a little bit from before we started on viewport collision. If the zoom is right around the integer boundary and moving back and forth quickly (I sometimes accidentally do this with the two-finger zoom gesture), you can get pretty noticeable flicker -- I think the reason it's worse is that we're rendering both the z13 and z14 tiles at the same time, so you might have a fade operation running on the z13 tile that's getting intermittently "flickered out" by the maxzoom limit of z14. The perceptual effect can be something like "fade in but disappear for a frame in the middle". Before all of our recent changes, it was just visible or not, so even though you could get flicker swapping back between zoom levels, there weren't any faulty-looking animations.

PR #6951, opened 2018-07-11
Introduce symbol cross fading when crossing integer zoom levels.

ChrisLoer:
Fixes issue #934: when tiles containing symbols are removed from the render tree, hold onto the tiles until `fadeDuration` ms have elapsed, and continue rendering _just_ the symbol layers so that the fade animation can finish.

**Before**
js-no-symbol-cross-fade

**After**
js-symbol-cross-fade 2

Rough port of logic in https://github.com/mapbox/mapbox-gl-native/pull/10468: although the idea is the same, the logic ends up looking pretty different because (1) painter/SourceCache are so different from renderer/TilePyramid, and (2) gl-js supports variable fade durations.

My testing so far has been entirely manual, and I never figured out an automated test for this on the native side. I think I might be able to do something similar to `regressions/mapbox-gl-js#5740` that exercises some of this logic, but most of the things I'd worry about getting wrong (like holding on to too many tiles) wouldn't really show up. To do my manual testing, I entered log statements to track tile rentention, and watched that tiles were being added and removed when expected as I crossed integer zoom boundaries. I also tracked the total number of tiles in the source cache before and after the change during a few animations. As expected: the count of tiles was the same at rest, and peaked a little bit higher during animation.

The paint benchmark doesn't show a noticeable change, but I'd expect the cost here to show up during animations, so I may have to come up with some sort of before on after. On native, the cost never seemed that bad, but it's hard to quantify -- and there's definitely some added cost from drawing extra symbol layers (and some extra memory used holding onto tiles). As in native, we hold onto tiles regardless of the reason they were removed from the render tree, just because the logic is simpler. We could try to exclude tiles that we know fall outside the viewport, but they should also be pretty cheap to render because their fragment shaders won't run?

screenshot 2018-07-11 11 51 37

 - [x] post benchmark scores

/cc @mourner @ansis @brunoabinader @kkaefer


ChrisLoer:
I also used log statements to watch how long it took to go from "last map movement" to "last frame rendered" -- my concern was that these changes could cause us to keep rendering longer because we're holding onto (and doing placement) on extra tiles. It looks like in most cases we're still looking at ~300ms of rendering after movement stops (excluding renders triggered by tiles loading from the network). The basic logic of why we don't incur extra cost is something like:

`placementChanged` only gets triggered when a symbol (identified by `crossTileID`):
 - Goes from placed to not-placed or vice versa
   - There's no change happening for the "hold for fade" tiles, since they're hard-wired to always be not-placed
 - Is missing from current placement and was placed in previous placement
   - When the held for fade tile gets dropped, it'll be missing from the next placement, but it will have been not-placed in the previous placement, so no change.
 - Was not in the previous placement and is placed now
   - The held-for-fade tiles don't add anything new

ChrisLoer:
To get an idea for the performance impact on zoom, I set up an 1103x929px viewport centered on `16/34/-118.4` (LA streets), and then counted the frames rendered over the course of a 10 second zoom animation out to 7, another 10 seconds back to 16, and then repeated again. With the changes, the animation averaged 59.05 FPS. Without the changes, the animation averaged 59.45 FPS. Although it sounds small written that way, it _is_ roughly a doubling of dropped frames... There's enough noise in FPS counting that it's hard to be too precise about what this means, but the pattern of a slight decrease in frame rate has so far been consistent. My interpretation is:

- There's probably a non-negligible performance hit
- There's probably not a huge/immediately noticeable performance hit
- From a UX point of view, I think the benefit of smoother label transitions will outweigh the potential cost in extra dropped frames

zoom-fps-test

ChrisLoer:
> One minor semi-related thing: layers with minzoom aren't faded. They are instantly hidden once you cross the zoom threshold.

Yeah, that's filed as #5726, I noticed that in the native cross-fading but forgot about it. I don't think it's urgent to fix, but there is maybe a _little_ UX regression if you're crossing back and forth over one of the `minzoom` limits because showing/hiding a symbol that's in the middle of a fade animation looks worse than just plain showing/hiding.

>  I think I might be able to do something similar to regressions/mapbox-gl-js#5740 that exercises some of this logic

I've been trying and failing to get a render test working using the `["wait", <ms>]` syntax. I don't think it's critical for this PR, but I'm actually kind of confused about the way the test harness is failing and want to spend a little time trying to figure it out.

ChrisLoer:
I got the cross-fade render test working and pushed it. The interaction between the test harness's clock-shifting and the placement logic is kind of hard to reason about and I worry the test might end up being a bit brittle, but I think it's also good to have at least one render test that exercises symbol fading.

ChrisLoer:
@nickidlugash just a heads-up this is coming down-stream to a gl-js renderer near you. I don't think it should be very disruptive since we already have this behavior on gl-native and I think it's the behavior people would usually expect in the first place.

Issue #4359, opened 2017-03-03
Build fails when using mapbox-gl webpack 2 and UglifyJSPlugin


henningko:
**mapbox-gl-js version**:
0.32.1

### Steps to Trigger Behavior

 Set up Mapbox with webpack 2  via [vuejs-template](https://github.com/vuejs-templates/webpack)

Configuration is as follows:
Base:
https://gist.github.com/henningko/5a76d12a14e485bfc4dd62c2e50ad08e
Production:
https://gist.github.com/henningko/3387f41a55558af60205ea00bbc6b530

Mapbox is used with `import mapboxgl from 'mapbox-gl'`.

### Expected Behavior

Mapbox can be successfully uglified.

### Actual Behavior

When building for production, [UglifyJS2](https://github.com/mishoo/UglifyJS2) has Mapbox throw a `Uncaught ReferenceError: n is not defined`.
This error also occurs when using Google Closure instead of UglifyJS.
May be related to #1649 , which mainly treats webpack v1 and is closed.

Thanks for your help!

stevewillard:
Seeing this as well. Anyone know which version I can downgrade to, which minifies properly?

henningko:
Fixed the problem by setting `devtool: 'eval'` instead of `source-map`. Not sure why source-map causes trouble for Mapbox, though.

mourner:
Could this be a WebPack 2 bug? This doesn't seem to be a fault on the GL JS side.

stevewillard:
@mourner yeah definitely. The weird thing is that this was working about a month ago for me. I'm trying to narrow down which dependency is causing the issue.

@henningko eval works for me too, but what about production? `cheap-module-source-map` has the same issue

henningko:
@stevewillard I was just happy to get a minified version out. 😐 I have opened an issue with webpack, too. It's still a gnarly workaround to get the to function together.

plugn:
Just met the same issue after migrating from webpack 1.x to 2.x. 
I have the same application built with `vue-template/webpack` 1.x working  properly: 
[wepack1 app](http://auctioner-raincoat-83602.netlify.com/)

Here is a deployed app with webpack 2 (Errors in console) 
[webpack2 app](http://telephone-operator-yell-70606.netlify.com/)

kachkaev:
Having this in production build since the end of last week, after upgrading mapbox-gl and its react wrapper. Could you please share your workaround guys?

[...]

PR #6956, opened 2018-07-12
Fix Webpack production mode compatibility

mourner:
A proposal for fixing #4359. Sets `ignoreGlobal: true` option for `rollup-plugin-commonjs` plugin, turning off its handling of the `global` keyword in CommonJS modules, and upgrades the plugin to the latest version (because the option didn't work properly in the previous one). In particular, it stops transforming top-level `this` reference into a variable that contains the `typeof global` check (which consequently triggered Webpack issues). 

One consequence of the fix is that GL JS will throw a runtime error when it encounters a reference to the `global` keyword. However, there are no such instances in the bundle currently — the only real reference I found (in the browserified node `assert` module) gets treeshaken away because it's in unused methods. So in theory this fix won't bring much trouble unless we plan on using Node-targeted modules with the `global` keyword in future.

Thanks a lot to @aendrew who provided a minimal repro that helped me find this fix.

### DIFF
--- a/build/rollup_plugins.js
+++ b/build/rollup_plugins.js
@@ -29,6 +29,8 @@ export const plugins = () => [
         include: 'src/shaders/index.js'
     }),
     commonjs({
+        // global keyword handling causes Webpack compatibility issues, so we disabled it:
+        // https://github.com/mapbox/mapbox-gl-js/pull/6956
         ignoreGlobal: true,
         namedExports: {
             '@mapbox/whoots-js': ['getTileBBox']
--- a/build/rollup_plugins.js
+++ b/build/rollup_plugins.js
@@ -29,6 +29,7 @@ export const plugins = () => [
         include: 'src/shaders/index.js'
     }),
     commonjs({
+        ignoreGlobal: true,
         namedExports: {
             '@mapbox/whoots-js': ['getTileBBox']
         }
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "rollup": "^0.57.0",
     "rollup-plugin-browserify-transform": "^1.0.1",
     "rollup-plugin-buble": "^0.18.0",
-    "rollup-plugin-commonjs": "^8.2.6",
+    "rollup-plugin-commonjs": "^9.1.3",
     "rollup-plugin-json": "^2.3.0",
     "rollup-plugin-node-resolve": "3.0.0",
     "rollup-plugin-replace": "^2.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3843,9 +3843,9 @@ estree-walker@^0.3.0:
   version "0.3.1"
   resolved "https://registry.yarnpkg.com/estree-walker/-/estree-walker-0.3.1.tgz#e6b1a51cf7292524e7237c312e5fe6660c1ce1aa"
 
-estree-walker@^0.5.0:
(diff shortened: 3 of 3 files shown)


mourner:
I must also mention that an alternative way to achieve a working Webpack build would be to release new versions of `whoots-js` and `shelf-pack` (cc @bhousel), adding a `module` field to `package.json`. Currently they're treated as CommonJS by Rollup despite being authored as ES modules because instead of `module` field, they use `jsnext:main` which is considered deprecated and is no longer handled by modern bundlers. We should definitely do this, but I'd still merge this PR because we might want to use some CommonJS modules in future, even after eliminating the current uses.

bhousel:
> I must also mention that an alternative way to achieve a working Webpack build would be to release new versions of whoots-js and shelf-pack (cc @bhousel), adding a module field to package.json.

Thanks for flagging @mourner, I'll take care of this now 👍

anandthakker:
> I must also mention that an alternative way to achieve a working Webpack build would be to release new versions of whoots-js and shelf-pack (cc @bhousel), adding a module field to package.json. 

@mourner can you explain the relationship between `ignoreGlobal: true` and rollup-plugin-commonjs's handling of these cjs dependencies that expose ES modules?

mourner:
@anandthakker sure. If you look into `dist/mapbox-gl-dev.js` on `master`, you can see the following line at the top introduced by `rollup-plugin-commonjs`:

```js
var commonjsGlobal = typeof window !== 'undefined' ? window : typeof global !== 'undefined' ? global : typeof self !== 'undefined' ? self : {};
```

When Rollup encounters a commonjs-style UMD build such as previous versions of `whoots-js` and `shelf-pack` (now updated thanks to @bhousel), it sees a top-level `this` keyword (used by the UMD wrapper) and replaces it with a `commonjsGlobal` reference:

```js
var index_umd = createCommonjsModule(function (module, exports) {
(function (global, factory) {
	'object' === 'object' && 'object' !== 'undefined' ? module.exports = factory() :
	typeof undefined === 'function' && undefined.amd ? undefined(factory) :
	(global.ShelfPack = factory());
}(commonjsGlobal, (function () {
```

Setting `ignoreGlobal: true` disables that top-level `this` substitution, so the bundle no longer needs the `commonjsGlobal` definition and it gets treeshaken away in the production build.

build/rollup_plugins.js
```diff
@@ -29,6 +29,7 @@ export const plugins = () => [
         include: 'src/shaders/index.js'
     }),
     commonjs({
+        ignoreGlobal: true,
```
jfirebaugh: Add a comment here outlining why we're using this option and linking to this PR.

george3447:
@jfirebaugh, @mourner

Could you please tell, is this PR included in latest npm release (0.47.0-beta.1), if not is there any tentative date planned for next release including this PR. It would be really helpful if you can provide some info regarding the same. Thanks.

jfirebaugh:
It's not included in 0.47.0-beta.1, but if some of the folks here who have been affected by the issue can try it out (using a build from `master`) and confirm that it resolves the issue for them and doesn't cause any new issues, we'll include it in 0.47.0 final (due out Wednesday 7/18).

aendrew:
@jfirebaugh @mourner Just tried it with my current project, LGTM :+1:

Issue #6959, opened 2018-07-12
Map#setFeatureState should reject feature identifier without an id field


asheemmamoowala:
The `Map#setFeatureState` should require the `id` field in the first argument and reject invocations missing said field.

**mapbox-gl-js version**: 0.46.1

Discovered this while debugging an issue where string ids in geojson data do not work with `Map#setFeatureState`. That is related to : #2716.

asheemmamoowala:
Addressed by #6974

PR #6960, opened 2018-07-12
Revert Data Join example until feature.id type issues are resolved

asheemmamoowala:
#6580 updated the Data Join example to use the new `Map#setFeatureState` API. While the updated example shows the power of using the new API for data joins, it causes confusion with developers trying to use it in two ways:
1. Input features must have numeric ids assigned. Non numeric/string ids (valid in GeoJSON) are not supported and break silently. ref: #2716, #6959
2. Mapbox's own APIs sometimes discard the feature ids in input data when creating a Tileset.

cc @ryanbaumann @samgehret

### DIFF
--- a/docs/pages/example/data-join.html
+++ b/docs/pages/example/data-join.html
@@ -12,58 +12,58 @@ var map = new mapboxgl.Map({
 // Join local JSON data with vector tile geometry
 // USA unemployment rate in 2009
 // Source https://data.bls.gov/timeseries/LNS14000000
-var maxValue = 14;
+var maxValue = 13;
 var data = [
-    {"STATE_ID": 1, "unemployment": 13.17},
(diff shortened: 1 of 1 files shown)


andrewharvey:
>2. Mapbox's own APIs sometimes discard the feature ids in input data when creating a Tileset.

ref https://github.com/mapbox/mapbox-gl-js/issues/6019#issuecomment-403262981

asheemmamoowala:
@andrewharvey we have discovered a few other ways that the ids are re-created or dropped. Overall the demo creates more trouble for developers with data outside of their control. It should be back soon as a few PR fixes land.

A couple PRs related to this:
- https://github.com/mapbox/supercluster/pull/94
- https://github.com/mapbox/geojson-vt/pull/109

PR #6964, opened 2018-07-13
Support two-argument shorthand for "match" expression

### LINKED ISSUE #6950: `match` expression shorthand for "true/false" evaluation?
## Motivation

`match` expressions in filters need to explicitly output true/false assignments:

`[ "match", [ "get", "class" ], [ "path", "pedestrian" ], true, false ]`

Whereas comparison expressions directly evaluate to true/false:

 `[ "==", [ "get", "class" ], "path" ]`

Even though this `match` pattern makes sense given how the syntax for this operator works, it still seems quite unintuitive/clunky in filters. I'm used to seeing data querying syntax and data filtering UI in other languages/programs treat "is true, otherwise false" as implied.

## Design Alternatives

The current design does not block any filtering abilities, so I think it would be ok if we left it as is. I'm also not sure if this is something that Studio would abstract out (cc @tristen @samanpwbb)

## Design

I have no idea how practical/feasible this is, but could we allow `match` expressions to only have three arguments (operator, input, label_1), and default these expressions to have output_1: `true`, default: `false`? I think this would basically mimic the `in` operator functionality in our deprecated filter syntax.

This seems mainly useful for filters, but perhaps it would be helpful for boolean style properties as well. 

Perhaps `case` expressions could have the same shorthand (I haven't thought about this much because I haven't used this operator much in filters).

### Mock-Up

This would be valid:

`"filter": [ "match", [ "get", "class" ], [ "path", "pedestrian" ] ]`

### Concepts

I think we have precedent for this type of shorthand with `get` and `has` expressions (where the last argument is optional, and defaults to the current feature's properties).

### Implementation

I'm not sure what implications this may have for Studio or other parts of the Mapbox ecosystem.

/cc @anandthakker @jfirebaugh @mapbox/maps-design

jfirebaugh:
Support two-argument shorthand for "match" expression. Fixes #6950.

## Launch Checklist

### DIFF
--- a/docs/components/expression-metadata.js
+++ b/docs/components/expression-metadata.js
@@ -144,6 +144,12 @@ const types = {
             'label_n: InputType | [InputType, InputType, ...], output_n: OutputType, ...',
             'default: OutputType'
         ]
+    }, {
+        type: 'boolean',
+        parameters: [
+            'input: InputType (number or string)',
+            'label: InputType'
+        ]
     }],
     var: [{
         type: 'the type of the bound expression',
--- a/src/style-spec/expression/definitions/match.js
+++ b/src/style-spec/expression/definitions/match.js
@@ -32,6 +32,8 @@ class Match implements Expression {
     }
 
     static parse(args: Array<mixed>, context: ParsingContext) {
+        if (args.length === 3)
+            args = args.concat(true, false);
         if (args.length < 5)
             return context.error(`Expected at least 4 arguments, but found only ${args.length - 1}.`);
         if (args.length % 2 !== 1)
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2359,6 +2359,9 @@
             "android": "6.0.0",
             "ios": "4.0.0",
             "macos": "0.7.0"
+          },
+          "two-argument shorthand": {
+            "js": "0.48.0"
           }
         }
       },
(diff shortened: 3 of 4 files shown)


nickidlugash:
@jfirebaugh Just wanted to check up on the status of this PR. Thanks!

jfirebaugh:
It's pretty much ready to go, I'm just not totally convinced it's worth it -- it's a modest syntax improvement that fixes an ergonomic papercut, but it adds a special case internally, and in similar-feeling situations that's been a source of future bugs and a potential snag to hit when making changes. My main concern is that the syntax doesn't round trip through the internal representation -- in native, if you serialize the result of getting an expression of this form from the runtime styling API, it'll use `true, false`. In https://github.com/mapbox/mapbox-gl-native/pull/12415#issuecomment-405857234 @1ec5 details some of the nuances associated with that. We could introduce additional code so that it round trips, but then it's even more of a special case internally.

samanpwbb:
> I'm just not totally convinced it's worth it -- it's a modest syntax improvement that fixes an ergonomic papercut, but it adds a special case internally, and in similar-feeling situations that's been a source of future bugs and a potential snag to hit when making changes.

This change also adds complexity and a special case to Studio. It'd be fine to support it, but if it makes any difference, I am a soft 👎 here.

ryanhamley:
@nickidlugash @samanpwbb How are we feeling about this change? The general consensus seems to be a resounding "Meh" but I'd like to make sure this doesn't just end up in purgatory so we should probably either merge it or close it. FWIW, @1ec5 said on Slack that he was not opposed to the change.

nickidlugash:
If others are not sure that this is worth supporting, I'm fine with not moving forward with this. I've seen match statements trip up style authors, given the way data filtering is typically defined (both in other programs/languages, and in our previous filter syntax), but it's probably not a big deal. If and when Studio supports a UI for expressions-based filters, perhaps this is something that can be glossed over in the UI instead?

ryanhamley:
Closing since no one seems to be a strong yes on this feature

PR #6974, opened 2018-07-16
Require feature id in Map#setFeatureState

asheemmamoowala:
`Map#setFeatureState` should validate the input feature identifier object for the `id` field, in addition to `source` and `sourceLayer`.

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -794,6 +794,10 @@ class Style extends Evented {
             this.fire(new ErrorEvent(new Error(`The sourceLayer parameter must be provided for vector source types.`)));
             return;
         }
+        if (feature.id == null || feature.id === "") {
+            this.fire(new ErrorEvent(new Error(`The feature id parameter must be provided.`)));
+            return;
+        }
 
         sourceCache.setFeatureState(sourceLayer, feature.id, state);
     }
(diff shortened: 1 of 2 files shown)


src/style/style.js
```diff
@@ -794,6 +794,10 @@ class Style extends Evented {
             this.fire(new ErrorEvent(new Error(`The sourceLayer parameter must be provided for vector source types.`)));
             return;
         }
+        if (feature.id == null) {
```
ryanhamley: What happens if you pass an undefined value or empty string as the feature id? Is that considered valid?

asheemmamoowala: `== null` works for `undefined` and `null`, but not for `""`. Ive added that additional check. Using `===` isn't appropriate here because `0` is a valid id value.

ryanhamley: 👍 good explanation

strech345:
does setFeatureState only works if the id-propertyname is "id"? whats happend if it is a different name?

asheemmamoowala:
@strech345 the `id` field required for this API is the top level `feature.id` field, not one of the fields in its property collection (`feature.propertis.id`).

strech345:
@asheemmamoowala 
thanks. and how to create mbtiles with id by using tippicanoe? Do i only need a geojson which does have the id property?

ericfischer:
https://github.com/mapbox/tippecanoe/pull/615 adds the most basic `--generate-ids` option (using the input feature sequence for the ID), with the disclaimer that the IDs are not stable and that their format may change in the future.

PR #6980, opened 2018-07-17
User turnstile event

asheemmamoowala:
### What is in this change?
Just like the mapbox-gl-native mobile SDKs, mapbox-gl-js will send a turnstile event, once per calendar day, when (and only when) the Mapbox access token is set and resources are loaded from a Mapbox API. No turnstile event will be sent for maps that use only third-party styles and tiles. This is the same behavior as the mapbox-gl-native mobile SDKs.

The turnstile event generates a random id that is persisted in the browser's [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) (when available). No other personal identifiable information is sent as part of this event. The subdomain that processes the event is distinct from the tiles API servers and Mapbox does not connect events between the two. There is no way for us to connect style and tile requests to an individual user.

You can view the part of our privacy policy that covers random identifiers [here](https://www.mapbox.com/privacy/#%5BAaMSWwdt%5D)

### Why is this needed? 
Today, the way that Mapbox determines distinct users of the gl-js sdk is imprecise. A turnstile event helps get more accurate information about things like:

- How many users are active on an account monthly. Our licensing incorporates end user limits for commercial applications; this change will allow us to closely track these numbers and make them visible to customers.
- What versions of gl-js are being used by end-users, and what browsers are being used. This will help us prioritize browser compatibility issues and support for cutting-edge technologies such as WebGL 2.

Additionally, as a company we want to use consistent metrics across all our platforms. This change brings the web platform in sync with our mobile platforms.

### Can I opt out of this tracking?
If your use of mapbox-gl-js does not set the access token (using `mapboxgl.accessToken`) then this code is always disabled. When the access token is set, there is no option to disable the turnstile event.

## Launch Checklist

 - [ ] ~post benchmark scores~

cc @jfirebaugh @anandthakker

### DIFF
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -134,7 +134,7 @@ export const postTurnstileEvent = function(tileUrls: Array<string>) {
     // mapbox tiles.
     if (!config.ACCESS_TOKEN ||
         !tileUrls ||
-        !tileUrls.some((url) => { return /mapbox.com/i.test(url); })) {
+        !tileUrls.some((url) => { return /mapbox.c(n)|(om)/i.test(url); })) {
         return;
     }
 
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -3,6 +3,12 @@
 import config from './config';
 
 import browser from './browser';
+import window from './window';
+import { version } from '../../package.json';
+import { uuid, validateUuid, storageAvailable } from './util';
+import { postData } from './ajax';
+
+import type { RequestParameters } from './ajax';
 
 const help = 'See https://www.mapbox.com/api-documentation/#access-tokens';
 
@@ -119,3 +125,72 @@ function formatUrl(obj: UrlObject): string {
     const params = obj.params.length ? `?${obj.params.join('&')}` : '';
     return `${obj.protocol}://${obj.authority}${obj.path}${params}`;
 }
+
+const STORAGE_TOKEN = 'mapbox.userTurnstileData';
+const localStorageAvailable = storageAvailable('localStorage');
+
+export const postTurnstileEvent = function(tileUrls: Array<string>) {
+    //Enabled only when Mapbox Access Token is set and a source uses
+    // mapbox tiles.
+    if (!config.ACCESS_TOKEN ||
+        !tileUrls ||
+        !tileUrls.some((url) => { return /mapbox.com/i.test(url); })) {
+        return;
+    }
+
+    let anonId = null;
+    let lastUpdateTime = null;
+    let pending = false;
+    //Retrieve cached data
+    if (localStorageAvailable) {
+        const data = window.localStorage.getItem(STORAGE_TOKEN);
+        if (data) {
+            const json = JSON.parse(data);
+            anonId = json.anonId;
+            lastUpdateTime = Number(json.lastSuccess);
+        }
+    }
+
+    if (!validateUuid(anonId)) {
+        anonId = uuid();
+        pending = true;
+    }
+
+    // Record turnstile event once per calendar day.
+    if (lastUpdateTime) {
+        const lastUpdate = new Date(lastUpdateTime);
+        const now = new Date();
+        const daysElapsed = (+now - Number(lastUpdateTime)) / (24 * 60 * 60 * 1000);
+        pending = pending || daysElapsed >= 1 || daysElapsed < 0 || lastUpdate.getDate() !== now.getDate();
+    }
+    if (!pending) {
+        return;
+    }
+
+    const evenstUrlObject: UrlObject = parseUrl(config.EVENTS_URL);
+    evenstUrlObject.params.push(`access_token=${config.ACCESS_TOKEN || ''}`);
+    const request: RequestParameters = {
+        url: formatUrl(evenstUrlObject),
+        headers: {
+            'Content-Type': 'text/plain' //Skip the pre-flight OPTIONS request
+        }
+    };
+
+    const payload = JSON.stringify([{
+        event: 'appUserTurnstile',
+        created: (new Date()).toISOString(),
+        sdkIdentifier: 'mapbox-gl-js',
+        sdkVersion: `${version}`,
+        'enabled.telemetry': false,
+        userId: anonId
+    }]);
+
+    postData(request, payload, (error) => {
+        if (!error && localStorageAvailable) {
+            window.localStorage.setItem(STORAGE_TOKEN, JSON.stringify({
+                lastSuccess: Date.now(),
+                anonId: anonId
+            }));
+        }
+    });
+};
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -151,12 +151,18 @@ class TurnstileEvent {
         if (this.pendingRequest || this.queue.length === 0) {
             return;
         }
-        let dueForEvent = false;
+        const storageKey = `${turnstileEventStorageKey}:${config.ACCESS_TOKEN || ''}`;
+        let dueForEvent = (this.eventData.accessToken !== config.ACCESS_TOKEN);
+
+        //Reset event data cache if the access token changed.
+        if (dueForEvent) {
+            this.eventData.anonId = this.eventData.lastSuccess = null;
+        }
         if (!this.eventData.anonId || !this.eventData.lastSuccess &&
             isLocalStorageAvailable) {
             //Retrieve cached data
             try {
-                const data = window.localStorage.getItem(turnstileEventStorageKey);
+                const data = window.localStorage.getItem(storageKey);
                 if (data) {
                     this.eventData = JSON.parse(data);
                 }
@@ -178,8 +184,6 @@ class TurnstileEvent {
             const daysElapsed = (nextUpdate - this.eventData.lastSuccess) / (24 * 60 * 60 * 1000);
             dueForEvent = dueForEvent || daysElapsed >= 1 || daysElapsed < 0 || lastUpdate.getDate() !== nextDate.getDate();
         }
-
-        dueForEvent = dueForEvent || (this.eventData.accessToken !== config.ACCESS_TOKEN);
 
         if (!dueForEvent) {
             return this.processRequests();
@@ -210,7 +214,7 @@ class TurnstileEvent {
                 this.eventData.accessToken = config.ACCESS_TOKEN;
                 if (isLocalStorageAvailable) {
                     try {
-                        window.localStorage.setItem(turnstileEventStorageKey, this.eventData);
+                        window.localStorage.setItem(storageKey, JSON.stringify(this.eventData));
                     } catch (e) {
                         warnOnce('Unable to write to LocalStorage');
                     }
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -128,7 +128,7 @@ function formatUrl(obj: UrlObject): string {
     return `${obj.protocol}://${obj.authority}${obj.path}${params}`;
 }
 
-class TurnstileEvent {
+export class TurnstileEvent {
     eventData: { anonId: ?string, lastSuccess: ?number, accessToken: ?string};
     queue: Array<number>;
     pending: boolean
@@ -139,6 +139,16 @@ class TurnstileEvent {
         this.queue = [];
         this.pending = false;
         this.pendingRequest = null;
+    }
+
+    postTurnstileEvent(tileUrls: Array<string>) {
+        //Enabled only when Mapbox Access Token is set and a source uses
+        // mapbox tiles.
+        if (config.ACCESS_TOKEN &&
+            Array.isArray(tileUrls) &&
+            tileUrls.some((url) => { return /(mapbox\.c)(n|om)/i.test(url); })) {
+            this.queueRequest(browser.now());
+        }
     }
 
     queueRequest(date: number) {
@@ -182,7 +192,7 @@ class TurnstileEvent {
             const lastUpdate = new Date(this.eventData.lastSuccess);
             const nextDate = new Date(nextUpdate);
             const daysElapsed = (nextUpdate - this.eventData.lastSuccess) / (24 * 60 * 60 * 1000);
-            dueForEvent = dueForEvent || daysElapsed >= 1 || daysElapsed < 0 || lastUpdate.getDate() !== nextDate.getDate();
+            dueForEvent = dueForEvent || daysElapsed >= 1 || daysElapsed < -1 || lastUpdate.getDate() !== nextDate.getDate();
         }
 
         if (!dueForEvent) {
@@ -227,12 +237,4 @@ class TurnstileEvent {
 
 const turnstileEvent_ = new TurnstileEvent();
 
-export const postTurnstileEvent = function (tileUrls: Array<string>) {
-    //Enabled only when Mapbox Access Token is set and a source uses
-    // mapbox tiles.
-    if (config.ACCESS_TOKEN &&
-        Array.isArray(tileUrls) &&
-        tileUrls.some((url) => { return /(mapbox\.c)(n|om)/i.test(url); })) {
-        turnstileEvent_.queueRequest(browser.now());
-    }
-};
+export const postTurnstileEvent = turnstileEvent_.postTurnstileEvent.bind(turnstileEvent_);
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -128,16 +128,14 @@ function formatUrl(obj: UrlObject): string {
 }
 
 class TurnstileEvent {
-    STORAGE_TOKEN: string;
-    localStorageAvailable: boolean;
+    static STORAGE_TOKEN: string;
+    static localStorageAvailable: boolean;
     eventData: Object;
     queue: Array<number>;
     pending: boolean
     pendingRequest: ?Cancelable;
 
     constructor() {
-        this.STORAGE_TOKEN = 'mapbox.turnstileEventData';
-        this.localStorageAvailable = storageAvailable('localStorage');
         this.eventData = { anonId: null, lastSuccess: null };
         this.queue = [];
         this.pending = false;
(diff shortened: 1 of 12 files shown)


src/util/mapbox.js
```diff
@@ -119,3 +125,72 @@ function formatUrl(obj: UrlObject): string {
     const params = obj.params.length ? `?${obj.params.join('&')}` : '';
     return `${obj.protocol}://${obj.authority}${obj.path}${params}`;
 }
+
+const STORAGE_TOKEN = 'mapbox.userTurnstileData';
+const localStorageAvailable = storageAvailable('localStorage');
```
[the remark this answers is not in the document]
asheemmamoowala: There is no fallback for saving the last update time locally - it would send multiple requests. Are you thinking that there should be a local variable for `lastUpdateTime` so that in a given instantiation of `Map`, this still maintains the _once per day_ frequency?

src/util/mapbox.js
```diff
@@ -119,3 +125,72 @@ function formatUrl(obj: UrlObject): string {
...
+
+export const postTurnstileEvent = function(tileUrls: Array<string>) {
+    //Enabled only when Mapbox Access Token is set and a source uses
+    // mapbox tiles.
+    if (!config.ACCESS_TOKEN ||
+        !tileUrls ||
+        !tileUrls.some((url) => { return /mapbox.c(n)|(om)/i.test(url); })) {
```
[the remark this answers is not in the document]
asheemmamoowala: `parseUrl` throws, and I'd like to avoid having a `try..catch`. Even with the `authority` parsed out of the url, it will need a check for `mapbox.com` and `mapbox.cn`.

mb12:
@asheemmamoowala Can you please kindly clarify the following?

1.) What is the difference between turnstile event and telemetry data that is collected on the native side  (if any)? GL Native collects information pertaining to touches, gestures, geo location when tracking is enabled etc.  Does the GL JS also track these and post this with each turnstile event?

2) What is the estimated bandwidth usage that can be attributed to information posted per turnstile event?  Recently the following issue was filed on the native side regarding data usage attributed to telemetry.
https://github.com/mapbox/mapbox-gl-native/issues/11979

asheemmamoowala:
> What is the difference between turnstile event and telemetry data that is collected on the native side (if any)? GL Native collects information pertaining to touches, gestures, geo location when tracking is enabled etc. Does the GL JS also track these and post this with each turnstile event?

@mb12 the turnstile event is a single event when a Map first requests a tile from a Mapbox API server, limited to one per calendar day. GL-JS does not track user interactions (touches and gestures) or  geo location. The [telemetry data](https://www.mapbox.com/telemetry/#telemetry) collected by Mobile SDKs includes many more events including intermittent location updates. 

> What is the estimated bandwidth usage that can be attributed to information posted per turnstile event?

Each turnstile event includes a `POST` with 6-7 parameters that combine to <200 bytes in size.

src/util/mapbox.js
```diff
@@ -119,3 +126,108 @@ function formatUrl(obj: UrlObject): string {
...
+        }
+
+        if (!validateUuid(this.eventData.anonId)) {
+            this.eventData.anonId = uuid();
+            dueForEvent = true;
+        }
+        const nextUpdate = this.queue.shift();
+
+        // Record turnstile event once per calendar day.
+        if (this.eventData.lastSuccess) {
```
anandthakker: What's the expected behavior if localStorage is not available?  As of now, it looks like we'd only send a turnstile event once per page in that case, since `lastSuccess` would never be set.

test/unit/util/mapbox.test.js
```diff
@@ -272,5 +273,66 @@ test("mapbox", (t) => {
...
+            //Depend on having a successful POST in a prior test.
+            config.ACCESS_TOKEN = 'pk.*';
+
+            mapbox.postTurnstileEvent(['a.tiles.mapbox.com']);
+
+            const req = window.server.requests[0];
+            t.false(req);
+
+            t.end();
+        });
```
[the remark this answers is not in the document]
anandthakker: They should also test the expected behavior:
 - With and without `localStorage` available (currently, I think they're testing without it)
 - For a page that remains open > 1 day
 - For a page that is closed and reopened a day later (i.e., the load-from-localStorage path)

src/util/mapbox.js
```diff
@@ -119,3 +126,108 @@ function formatUrl(obj: UrlObject): string {
...
+            const lastUpdate = new Date(this.eventData.lastSuccess);
+            const nextDate = new Date(nextUpdate);
+            const daysElapsed = (nextUpdate - this.eventData.lastSuccess) / (24 * 60 * 60 * 1000);
+            dueForEvent = dueForEvent || daysElapsed >= 1 || daysElapsed < 0 || lastUpdate.getDate() !== nextDate.getDate();
```
anandthakker: What's the reason for the `lastUpdate.getDate() !== nextDate.getDate()` condition?  When _would_ these two dates be equal?

src/util/mapbox.js
asheemmamoowala: `eventData` caches the id and success values within a single session. But I see the problem, this value needs to be updated irrespective of `localStorageAvailable` when a request succeeds.

If `localStorage` is not available, there may be more than one event sent per calendar day, but the events server de-duplicates those.

src/util/mapbox.js
asheemmamoowala: `daysElapsed` is not perfectly named. It calculates how many 24-hr periods have elapsed since the last update. Notifications should be sent once per calendar day, so the date check allows testing for situations where a user loads the map at 11pm and again at 1am. 

`Date#getDate()` returns just the day of the month, so it would be equal on 6/20 and 7/20. The combined check returns true if the clock has gone backwards (fly east), or forward ( by 24+ hrs or by a calendar date).

src/util/mapbox.js
```diff
@@ -119,3 +126,108 @@ function formatUrl(obj: UrlObject): string {
...
+        this.pendingRequest = postData(request, payload, (error) => {
+            if (!error && this.localStorageAvailable) {
+                this.eventData.lastSuccess = nextUpdate;
+                window.localStorage.setItem(this.STORAGE_TOKEN, JSON.stringify({
+                    lastSuccess: this.eventData.lastSuccess,
+                    anonId: this.eventData.anonId
+                }));
+                this.pendingRequest = null;
+                this.pending = false;
+                this.processRequests();
```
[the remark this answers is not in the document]
asheemmamoowala: Definitely a bug that the `pending..` variables aren't cleared. For `processRequests` I think it should stop attempting to send requests if there is an error, until the next event is queued - in case of network connection loss, or a server outage.

test/unit/util/mapbox.test.js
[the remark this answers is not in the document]
asheemmamoowala: > For a page that remains open > 1 day
This is currently treated as a single event until the page requests a new mapbox vector/raster source.

src/util/mapbox.js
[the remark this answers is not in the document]
anandthakker: Ah, yeah that seems reasonable

src/util/mapbox.js
```diff
@@ -119,3 +126,113 @@ function formatUrl(obj: UrlObject): string {
...
+            headers: {
+                'Content-Type': 'text/plain' //Skip the pre-flight OPTIONS request
+            }
+        };
+
+        const payload = JSON.stringify([{
+            event: 'appUserTurnstile',
+            created: (new Date(nextUpdate)).toISOString(),
+            sdkIdentifier: 'mapbox-gl-js',
+            sdkVersion: `${version}`,
```
jfirebaugh: No need for string template.

test/unit/util/mapbox.test.js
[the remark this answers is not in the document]
jfirebaugh: Should also test the queuing mechanism.

src/util/util.js
```diff
@@ -196,6 +197,27 @@ export function uniqueId(): number {
...
+    return b();
+}
+
+/**
+ * Validate a string to match UUID(v4) of the
+ * form: xxxxxxxx-xxxx-4xxx-[89ab]xxx-xxxxxxxxxxxx
+ * @param str string to validate.
+ */
+export function validateUuid(str: ?string): boolean {
+    return str ? /^[0-9a-f]{8}-[0-9a-f]{4}-[4][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(str) : false;
```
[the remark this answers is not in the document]
asheemmamoowala: Flow complaints that

src/util/util.js
```diff
@@ -196,6 +197,27 @@ export function uniqueId(): number {
     return id++;
 }
 
+/**
+ * Return a random UUID (v4). Taken from: https://gist.github.com/jed/982883
+ */
+export function uuid(): string {
+    function b(a) {
+        return a ? (a ^ Math.random() * 16 >> a / 4).toString(16) :
+        //$FlowFixMe
```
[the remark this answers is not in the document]
asheemmamoowala: Flow doesn't like the implied array literal conversion in : `([1e7] + -[1e3] + -4e3 + -8e3 + -1e11)`.

> array literal [1] is not a number.

test/unit/util/mapbox.test.js
```diff
@@ -272,5 +273,137 @@ test("mapbox", (t) => {
...
+        t.test('POSTs consecutive appuserTurnstile events on different calendar days', (t) => {
+            config.ACCESS_TOKEN = 'pk.*';
+            const today = Date.now();
+            const laterToday = today + 1;
+            const tomorrow = today + (25 * 60 * 60 * 1000); // Add a day
+
+            const stub = t.stub(browser, 'now');
+            stub.onCall(0).returns(today);
+            stub.onCall(1).returns(laterToday);
+            stub.onCall(2).returns(tomorrow);
```
anandthakker: This approach to stubbing could be fragile, because it assumes that `browser.now` is only called once per `postTurnstileEvent` call.  Alternative:

```js
let now = Date.now();

t.stub(browser, 'now').callsFake(() => now);

mapbox.postTurnstileEvent(['a.tiles.mapbox.com']);
const reqToday = window.server.requests[0];
reqToday.respond(200);
let reqBody = JSON.parse(reqToday.requestBody)[0];
t.ok(reqBody.created, new Date(now).toISOString());

now += 1; // later today
mapbox.postTurnstileEvent(['b.tiles.mapbox.com']);
t.ok(window.server.requests.length === 1);

now += (25 * 60 * 60 * 1000); // tomorrow
mapbox.postTurnstileEvent(['c.tiles.mapbox.com']);
const reqTomorrow = window.server.requests[1];
reqTomorrow.respond(200);
reqBody = JSON.parse(reqTomorrow.requestBody)[0];
t.ok(reqBody.created, new Date(now).toISOString());
```

test/unit/util/mapbox.test.js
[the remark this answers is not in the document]
anandthakker: Still need tests for the load-from-localStorage path (which will require stubbing localStorage)

test/unit/util/mapbox.test.js
asheemmamoowala: The point of this test is to test the queuing mechanism(as suggested by @jfirebaugh [previously(comment)](https://github.com/mapbox/mapbox-gl-js/pull/6980/files/ab94794b5a354d4805f6d1875c3948fb78082636#r204520204). Events are queued correctly triggered consecutively and/or responses are returned with a higher latency. The alternative you suggest is a serial approach that is being tested already.

anandthakker: Ah, I see.  We should be able to use the more explicit approach to manipulating mock-time though, right?

test/unit/util/mapbox.test.js
```diff
@@ -322,82 +304,219 @@ test("mapbox", (t) => {
...
+            });
+
+            t.test('POSTs event when previously stored timestamp is in the future', (t) => {
+                config.ACCESS_TOKEN = 'pk.*';
+                const now = +Date.now();
+
+                window.localStorage.setItem(`mapbox.turnstileEventData:${config.ACCESS_TOKEN}`, JSON.stringify({
+                    anonId: uuid(),
+                    lastSuccess: now + ms25Hours // 24-hours later
+                }));
```
anandthakker: I think it would be better to test this by calling `postTurnstileEvent` twice, with the (stubbed) time being set back between calls, rather than setting `localStorage` directly

test/unit/util/mapbox.test.js
```diff
@@ -322,82 +304,219 @@ test("mapbox", (t) => {
...
+
+                const req = window.server.requests[0];
+                req.respond(200);
+                const reqBody = JSON.parse(req.requestBody)[0];
+                t.ok(reqBody.created, new Date(now).toISOString());
+
+                t.end();
+            });
+
+            t.test('does not POST appuserTurnstile event second time within same calendar day', (t) => {
```
anandthakker: Rather than having some tests rely on state from previous tests, I'd be in favor of making each test self-contained. 
 This could be done by moving the logic from `mapbox.postTurnstileEvent` into a method on `TurnstileEvent`, exporting the `TurnstileEvent` class, and having each test construct a new instance to test against.

test/unit/util/mapbox.test.js
asheemmamoowala: Are you implying that the test should cover the case of the queue receiving events in an out-of-order time sequence? In a single threaded environment, that situation seems highly unlikely. Even if it does happen, it should be counted as a single session IMHO. 

Testing with `LocalStorage` allows verifying that clocks going back in time between sessions are accounted for, where it should be posted as a separate session correctly.

anandthakker: Oh -- no, I just meant that if possible, we should do this test using calls to `postTurnstileEvent` (the api being tested) rather than by directly manipulating `localStorage` (which requires baking more implementation details, like the storage key, into the test)

test/unit/util/mapbox.test.js
asheemmamoowala: I understand that it will make the tests self contained, but I think it is better to have a singleton `TurnstileEvent` for each instance of the map so that there can be a local cache of the eventData. Exposing the entire class without `postTurnstileEvent` being static would require the callers (Map in this case) to manage an instance of the event which seems unnecessary

test/unit/util/mapbox.test.js
asheemmamoowala: > we should do this test using calls to postTurnstileEvent

I believe that within a single session, the clock going backwards should not report a new event -  so this cannot be tested with calls to the method directly. To simulate calls across sessions, the best way is to change the access token and clear out the cached `eventData`

test/unit/util/mapbox.test.js
anandthakker: Exposing the class doesn't preclude having a singleton:

```js
export class TurnstileEvent {
   ...
   postTurnstileEvent(...) { /* actual logic */ }
}

const _turnstileEvent = new TurnstileEvent();

export function postTurnstileEvent() {
  return _turnstileEvent.postTurnstileEvent(...);
}
```

test/unit/util/mapbox.test.js
anandthakker: > this cannot be tested with calls to the method directly

Why not, with something like:

```js
let now = Date.now();
t.stub(browser, 'now').callsFake(() => now);

mapbox.postTurnstileEvent();

// simulate system clock being set backwards
now = now - 1000 * 60 * 60 * 24;

mapbox.postTurnstileEvent();
```

This is a more direct simulation of the situation we're intending to test, and allows the test to avoid a dependency on implementation details.

asheemmamoowala: That makes sense to add as a negative test, `when clock moves bacward, new event is not set.`

test/unit/util/mapbox.test.js
```diff
@@ -272,5 +278,313 @@ test("mapbox", (t) => {
...
+                const req = window.server.requests[0];
+                req.respond(200);
+
+                const reqBody = JSON.parse(req.requestBody)[0];
+                t.equal(req.url, `${config.EVENTS_URL}?access_token=key`);
+                t.equal(req.method, 'POST');
+                t.equal(reqBody.event, 'appUserTurnstile');
+                t.equal(reqBody.sdkVersion, version);
+                t.ok(reqBody.userId);
+                t.ok(reqBody.created, new Date(tomorrow).toISOString());
```
anandthakker: This should be `t.equal`. (`t.ok` only checks that the first argument is truthy).  Also, should `reqBody` be `JSON.parse(req.requestBody)[1]` rather than `[0]`?

anandthakker: A few other `t.ok(x, y)`s should be `t.equals` as well

asheemmamoowala: Good 👀 @anandthakker !

> should reqBody be JSON.parse(req.requestBody)[1] rather than [0]?

Actually `req` should be `window.server.requests[1]`. Found and fixed after the `t.equal` change.

jingsam:
There always should have an opt-out option, no matter set `access-token`  or not. Privacy does matter and you have the responsibility to tell users the purpose of data collection.

lilykaiser:
Thanks for your comment @jingsam . We take the responsibility of protecting users' privacy seriously. That's why there is no way for us to connect style and tile requests to an individual end user, and no personally identifiable information is collected in this event other than a random identifier.


> you have the responsibility to tell users the purpose of data collection.

As @asheemmamoowala mentioned in the original comment, we will use this data to track how many people are using different versions of gl-js to determine the severity of bugs and performance issues. It will also improve our tracking of how many users a commercial application has. Lastly, it brings our internal calculation of active users on gl-js in line with how we calculate mobile users.

We want to facilitate clear communication between us, developers who use our products, and end users. I hope the above points clarify why Mapbox wants to collect an end user number to our developer community.

We unfortunately can't offer opting out while using Mapbox map tile APIs, because this (anonymous) usage information is important for us to sustainably maintain this service.  If opting out is a hard requirement for you, note that Mapbox GL JS may be used with other tile services.

scottgerring:
Will these `localStorage` items be cleaned out at some point? We're seeing more and more items stashed in there over time but apparently no removal.

asheemmamoowala:
@scottgerring Removal is not in our current implementation or upcoming plans. Please open an issue with reproducible steps if you think that there are an unreasonable number of items being added to the local storage.

PR #6988, opened 2018-07-18
android-v6.3.0, ios-v4.2.0, macos-v0.9.0 in style spec

1ec5:
Updated the style specification compatibility tables to reflect mapbox/mapbox-gl-native#12176 mapbox/mapbox-gl-native#11869 mapbox/mapbox-gl-native#12329 in Android map SDK v6.3.0, iOS map SDK v4.2.0, and macOS map SDK v0.9.0.

/cc @ChrisLoer @mollymerp @friedbunny

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -5121,6 +5121,7 @@
       "sdk-support": {
         "basic functionality": {
           "js": "0.47.0",
+          "android": "6.3.0",
           "ios": "4.2.0",
           "macos": "0.9.0"
         },
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2501,7 +2501,9 @@
         "group": "Types",
         "sdk-support": {
           "basic functionality": {
-            "js": "0.45.0"
+            "js": "0.45.0",
+            "ios": "4.2.0",
+            "macos": "0.9.0"
           }
         }
       },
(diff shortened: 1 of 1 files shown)


src/style-spec/reference/v8.json
```diff
@@ -2971,6 +2973,11 @@
             "android": "6.0.0",
             "ios": "4.0.0",
             "macos": "0.7.0"
+          },
+          "collator": {
+            "js": "0.45.0",
+            "ios": "4.2.0",
+            "macos": "0.9.0"
```
1ec5: @ChrisLoer, I think we should call out the collator syntax separately for each of these comparison operators, since it was introduced after the comparison operators themselves. Not sure if there need to be any changes in docs/pages/style-spec.js.

ChrisLoer: Yeah, this should work and I agree it makes sense to call out separately.

src/style-spec/reference/v8.json
```diff
@@ -2983,6 +2990,11 @@
             "android": "6.0.0",
             "ios": "4.0.0",
             "macos": "0.7.0"
+          },
+          "collator": {
+            "js": "0.45.0",
```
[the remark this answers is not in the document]
tobrun: would prefer leaving it out until we have a specific api for it in Expression.java (technically people can already use it with a raw expression).

Issue #6643, opened 2018-05-08
Raster tile loading blocking UI, FPS < 2.0


hyperknot:
**mapbox-gl-js version**: 0.45.0
**browser**: Chrome 66 / macOS

 1. On an empty style, add a raster tile layer, especially with smaller tile sizes, like "fake-retina" 256 pixels as 128.
 2. Zoom around the map or use flyTo animations
 3. Things can get _really_ choppy.

### Link to Demonstration
(I tested using Chrome FPS meter)
example: https://zl761nnr3m.codesandbox.io/
code: https://codesandbox.io/s/zl761nnr3m

### Expected Behavior
UI interactions should be smooth, independent of tile loadings in the background.

### Actual Behavior
The UX is really bad with raster layers now. Chrome FPS meter regularly gets under 2.0 fps on my Macbook Pro.

I've almost finished porting a big codebase over to Mapbox GL viewer, but this one is puzzling me. Both Leaflet and OpenLayers can display raster maps at constant 58+ FPS. How is it possible that Mapbox GL can render the most complicated vector tile regions smoothly, but is breaking with simple raster tiles?

mourner:
This seems to be some kind of regression with the source cache logic — most of the time is spend figuring out which tiles to show, not actual rendering or loading. Investigating now.

mourner:
Looks like it's not a regression, just an algorithm that dramatically slows down quadratically when the tile size goes down — it was only optimized for vector tiles, where you typically only see 6-7 tiles on the screen. Thinking about a faster one now...

mourner:
OK, I have a good lead on this. Will submit a PR tomorrow.

mourner:
Turns out that the jank on raster layers is currently dominated by XHR requests, which we use to load images since #1472 because of an issue with browsers not fully respecting `Cache-Control: no-cache` (#1470). You can see how much snappier the raster test case becomes in 9b372d96de012520d01b740ef26de99e08d832d7 (`git co bench-raster-tile-load`, `yarn run start-debug`, then load `http://localhost:9966/debug/`). If you switch `getImage` back to XHR-based requests in `ajax.js`, it becomes unusably janky, as reported.

[...]

PR #6998, opened 2018-07-20
Optimize visible coordinates retrieval

mourner:
Simplifies and significantly optimizes visible coordinates retrieval. A part of my performance push for fixing #6643. Contains two independent changes (please rebase instead of squashing):

- The first commit simplifies and optimizes visible coordinates retrieval during the three painter rendering passes. Instead of getting the same coords multiple times, we retrieve them once for each source at the top and then use the cached objects.
- The second commit fixes a performance regression introduced by #5601, which made the sorting logic for renderable ids more expensive. The commit makes it only use fancy sorting for symbol layers, and simple zoom sorting for everything else.

Combined, on a test case from #6643, the PR makes retrieving coordinates 3 times faster — on a single fly animation profile, down from 700ms (21.8% of `_render`) to 230ms (7.8%).

- [ ] Revert test changes in #5601 to make the build pass.

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/render/painter.js
+++ b/src/render/painter.js
@@ -6,7 +6,6 @@ import { mat4 } from 'gl-matrix';
 import SourceCache from '../source/source_cache';
 import EXTENT from '../data/extent';
 import pixelsToTileUnits from '../source/pixels_to_tile_units';
-import { filterObject } from '../util/util';
 import VertexArrayObject from './vertex_array_object';
 import { RasterBoundsArray, PosArray } from '../data/array_types';
 import rasterBoundsAttributes from '../data/raster_bounds_attributes';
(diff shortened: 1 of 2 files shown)


src/render/painter.js
```diff
@@ -276,140 +275,111 @@ class Painter {
...
+            this.renderLayer(this, sourceCache, layer, coords);
+            prevSourceId = layer.source;
         }
 
         if (this.options.showTileBoundaries) {
-            const sourceCache = this.style.sourceCaches[Object.keys(this.style.sourceCaches)[0]];
-            if (sourceCache) {
-                draw.debug(this, sourceCache, sourceCache.getVisibleCoordinates());
+            for (const id in sourceCaches) {
+                draw.debug(this, sourceCaches[id], coordsAscending[id]);
```
ChrisLoer: It doesn't look like any behavior changes here, but it got me thinking -- would it be better/more consistent if we drew debug tile boundaries as the union of tiles in all source caches?

mourner: We used to draw debug boundaries for all source caches, but I felt it looked very confusing with multiple sources. You can try it but I think it's more useful to have clear, legible boundaries for at least one source.

mourner:
Pushed a fix that makes the tests pass without reverting them. Benchmarks look unaffected, because they don't exercise heavy zooming or small tiles.

image

PR #7008, opened 2018-07-23
Avoid XHR when possible for faster raster tile loading

mourner:
A proposal for addressing #6643. If we see that a raster tile we requested with XHR should be cached for a long time (currently set to 1 hr), we switch from XHR to simple `new Image` tile loads for that source. This fixes a lot of the jank when using raster tile sources.

## Launch Checklist

 - [x] ~~write tests for all new functionality~~ (we don't have tests for image loading, and they probably won't be useful anyway as long as we use `jsdom` and stubbing)
 - [x] post benchmark scores

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -130,8 +130,21 @@ function sameOrigin(url) {
 
 const transparentPngUrl = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAAC0lEQVQYV2NgAAIAAAUAAarVyFEAAAAASUVORK5CYII=';
 
-export const getImage = function(requestParameters: RequestParameters, callback: Callback<HTMLImageElement>): Cancelable {
-    // request the image with XHR to work around caching issues
+export const getImage = function(requestParameters: RequestParameters, callback: Callback<HTMLImageElement>, avoidXHR?: boolean): Cancelable {
+    // if we know for sure that the tile is cached for a long time, avoid XHR for better performance
+    // https://github.com/mapbox/mapbox-gl-js/issues/6643
+    if (avoidXHR && requestParameters.headers === undefined && requestParameters.credentials === undefined) {
+        const img: HTMLImageElement = new window.Image();
+        const url = requestParameters.url;
+        if (!sameOrigin(url)) {
+            img.crossOrigin = 'Anonymous';
+        }
+        img.onload = () => callback(null, img);
+        img.src = url;
+        return {cancel: () => { img.src = transparentPngUrl; }};
+    }
+
+    // otherwise request the image with XHR to work around caching issues
     // see https://github.com/mapbox/mapbox-gl-js/issues/1470
     return getArrayBuffer(requestParameters, (err, imgData) => {
         if (err) {
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -139,6 +139,7 @@ export const getImage = function(requestParameters: RequestParameters, callback:
         if (!sameOrigin(url)) {
             img.crossOrigin = 'Anonymous';
         }
+        img.onerror = () => callback(new Error(`Could not load image: ${url}`));
         img.onload = () => callback(null, img);
         img.src = url;
         return {cancel: () => { img.src = transparentPngUrl; }};
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -142,7 +142,7 @@ export const getImage = function(requestParameters: RequestParameters, callback:
         img.onerror = () => callback(new Error(`Could not load image: ${url}`));
         img.onload = () => callback(null, img);
         img.src = url;
-        return {cancel: () => { img.src = transparentPngUrl; }};
+        return {cancel: () => { img.onload = null; img.src = transparentPngUrl; }};
     }
 
     // otherwise request the image with XHR to work around caching issues
--- a/src/source/raster_dem_tile_source.js
+++ b/src/source/raster_dem_tile_source.js
@@ -1,9 +1,7 @@
 // @flow
 
-import { getImage, ResourceType } from '../util/ajax';
 import { extend } from '../util/util';
 import { Evented } from '../util/evented';
-import { normalizeTileURL as normalizeURL } from '../util/mapbox';
 import browser from '../util/browser';
 import { OverscaledTileID } from './tile_id';
 import RasterTileSource from './raster_tile_source';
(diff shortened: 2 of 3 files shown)


src/util/ajax.js
```diff
@@ -130,8 +130,21 @@ function sameOrigin(url) {
...
+    // https://github.com/mapbox/mapbox-gl-js/issues/6643
+    if (avoidXHR && requestParameters.headers === undefined && requestParameters.credentials === undefined) {
+        const img: HTMLImageElement = new window.Image();
+        const url = requestParameters.url;
+        if (!sameOrigin(url)) {
+            img.crossOrigin = 'Anonymous';
+        }
+        img.onload = () => callback(null, img);
+        img.src = url;
+        return {cancel: () => { img.src = transparentPngUrl; }};
```
hyperknot: Shouldn't `callback(err)` be handled here as well?

hyperknot: I think an [onerror](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement#Errors) handler would be enough.

mourner: Good point, added.

mourner:
Ran `LayerRaster` benchmarks — they are unaffected because they don't exercise the performance impact of tile loading.

hyperknot:
XHR heuristics works perfectly, both cases are working well.

jfirebaugh:
What are you using to evaluate this versus the prior behavior? In light testing, I'm not sure I can distinguish the difference.

mourner:
@jfirebaugh I'm using the original test case here https://codesandbox.io/s/zl761nnr3m — the difference is very noticeable, especially if you have #6995 (another big perf hit) pulled in when comparing.

hyperknot:
Here here a source with short expires time:
```
https://a.tile.opentopomap.org/{z}/{x}/{y}.png
```

And Mapbox satellite for example is with long expires time.

jfirebaugh:
I'm comparing e9e08e2e9d15804d2c2f74822623fc5143a3dbba and f3f9357eaf73b1a65a138deecfcfe1bfe535e85f side-by-side on each half of an external monitor and I can't really tell the difference. You're saying it should be really obvious?

hyperknot:
@jfirebaugh make sure you don't have developer tools open, as that introduced janks for me in Chrome.

old: https://codesandbox.io/s/zl761nnr3m
new (both PRs merged): https://codepen.io/hyperknot/full/vaxvJN/

jfirebaugh:
This is with developer tools closed. Subjectively, the behavior I see is:

* Approximately equal jank between the two. With this PR, maybe slightly less, but that also might be conditioning.
* This PR shows less blank white space when zooming out.
* This PR shows more pixelated, overzoomed tiles when zooming in.

mourner:
@jfirebaugh on the examples above (make sure both are viewed fullscreen), I get heavy stuttering on the old example (on my MBP Retina 15'' screen), especially in the middle of zooming in and zooming out, and almost no jank on the new one.

Perf recording before (`optimize-tile-retention`) (notice the red dots and big gaps):

image

After (`optimize-tile-retention` + `faster-raster-tiles-load` merged in):

image

I also noticed like @hyperknot that it still stutters when devtools are open, but stops stuttering when you start recording a timeline perf profile — I guess Chrome disables some debug stuff while recording.

jfirebaugh:
With developer tools open, looking at a performance recording, it looks like the time spent sending the requests might just be delayed until the end of the frame, rather than interspersed with JS execution:

image

mourner:
Here's another comparison, JS profiler. Before:

screen shot 2018-07-23 at 23 53 16

After:

screen shot 2018-07-23 at 23 53 25

Also, subjectively, if I disable the fly animation and just browse the map fullscreen, it's barely usable before the PR with unbearable stuttering, and smooth and pleasant after.

I think you closed the PR prematurely. Let's have a discussion first.

jfirebaugh:
Yeah, I don't see results as definitive as that. I'm testing locally with this file: https://gist.github.com/jfirebaugh/7cf3e57c359513f980d6aad56283d75e.

image

I wonder what causes such different results for us?

mourner:
@jfirebaugh I think it's important to have all the other raster perf PRs merged in for this to have pronounced effect. I just rebased both PRs (this and #6995) on master. Now can you compare this PR vs this PR + #6995 merged?

jfirebaugh:
I've been testing with `optimize-tile-retention` rebased on `master` and `faster-raster-tiles-load` rebased on top of that.

Does someone else want to step in and try? Must be some other variable unaccounted for here.

mourner:
@hyperknot @jfirebaugh I just pushed a new WIP branch based on this one that adds a basic image queue (https://github.com/mapbox/mapbox-gl-js/commit/cbb059fd1ce8847d47719db3f45d0e3cdcb0edcd), limiting parallel image loads to 16 at a time. Can you try it out and see if it makes a difference for performance?

mourner:
@hyperknot any comment on the above?

hyperknot:
Hi, sorry for not responding earlier. So do I understand right that the now stable release (0.49) has the previously discussed improvements, but not the tile queue one? And for tile queue I should check out https://github.com/mapbox/mapbox-gl-js/commit/cbb059fd1ce8847d47719db3f45d0e3cdcb0edcd?

mourner:
@hyperknot the last release includes all improvements except this PR (avoiding XHR for image loading). And additionally cbb059f explores the concept of limiting concurrency for image loads — let's see if it's any faster than this branch.

hyperknot:
@mourner OK, I'll check out the improvements and report back today.

mourner:
@hyperknot any updates?

hyperknot:
Oh, sorry, I promised to get back but I didn't. I'll do it now.

hyperknot:
@mourner this is absolutely wonderful performance wise! The 2 fps parts are now totally gone, even with console open.

I did a comparison between 0.50 and cbb059fd here:
https://codepen.io/hyperknot/pen/bmvzQx
https://codepen.io/hyperknot/pen/xyWMMO

My only problem now is that I'm yet 100% confident in how the code is implemented. What isn't clear is both handleImageLoad and getImage checks for `numImageLoads < > maxParallelImageLoads`. 
Shouldn't one of them be the reference for this test?

Also, are you sure it can never happen that the queue explodes? I mean getImage adds handlers for handleImageLoad which calls getImage.

Having said that, if all the delicate parts of these are thought out by you then I'm happy with the added huge performance improvements!

mourner:
@hyperknot awesome, thanks a lot for checking! This was a quick proof of concept so likely pretty buggy — and now since this seems promising, I'll probably explore it more. Note that while FPS is improved, the branch codepen version seems to have more blurry tiles when zooming in because new ones aren't loaded fast enough. 

We're also exploring raster loading performance from another angle in #7405, which will hopefully bring more improvements too.

hyperknot:
But blurry tiles are good, aren't they? I mean blurry means that smooth
animation is more important than title loading, which is exactly what
should happen.

Maybe it can be polished so that the tiles at the center of the screen take
higher priority then the ones at the  edges though.

kkaefer:
I can confirm that cbb059fd1ce8847d47719db3f45d0e3cdcb0edcd is much smoother on my computer as well. It seems to load the maximum number of image loads that are loaded in parallel. However, I'm still seeing more than one image per frame being _uploaded_ (call to `texImage2D`). That's something that we should limit to 1 image per second as well, i.e. only fire one `onload` event per animation frame.

kkaefer:
@hyperknot thank you for these wonderful test cases; they're very useful for debugging this issue!

mourner:
I have a feeling this PR won’t have much effect after #7497, so I’m going to close for now.

Issue #7012, opened 2018-07-23
Don't serialize SymbolFeature as part of SymbolInstance


ChrisLoer:
When we transfer a symbol bucket from the workers to the foreground, we include a copy of the source feature for each `SymbolInstance` we generate:

https://github.com/mapbox/mapbox-gl-js/blob/83b147b0f08365a588cd393dc007978793a3e565/src/data/bucket/symbol_bucket.js#L76-L77

These features are bulky and expensive to transfer, and as far as I know they're unused on the foreground. If we need access to the source feature on the foreground, we can use the "feature index" to look the feature up directly in the source data for the tile.

/cc @anandthakker @ansis @mourner

ChrisLoer:
This showed up as a significant part of the size of the JSON we transfer in: https://github.com/mapbox/mapbox-gl-js/issues/7011.

ChrisLoer:
Fixed in #7013.

PR #7013, opened 2018-07-23
Stop unnecessary serialization of symbol source features.

ChrisLoer:
Fixes issue #7012.
This data is expensive to serialize and isn't even being used.

I did a brief manual check to see that this didn't break anything. I'm still not sure why this was here in the first place, but I think it's (1) vestigial, and (2) got more expensive as part of adding expressions.

None of our benchmarks exercise the tile serialization pathway, but I did some quick-and-dirty benchmarking by zooming from 7 to 16 and back twice (40 seconds total animation) and counting frames. For each recording 600 == no dropped frames. The first run is presumably slower because the tiles are being loaded for the first time instead of from cache.

**Before**
562
594
600
598

**After**
563
594
589
597

🤷‍♂️  Not much of a difference? Let me try to focus on just first load since we'd expect the difference to mainly be there:

**Before**
562
558
562
534

-> 46 dropped frame avg

**After**
563
567
570
565

-> 34 dropped frame avg

If I squint enough that looks like a win!

## Launch Checklist

 - [ ] post benchmark scores

/cc @mourner @anandthakker @ansis


ChrisLoer:
I did another four runs of before/after and got generally slower results (not sure why: I did switch to a different internet connection and I suspect it being faster meant it loaded more tiles during the zoom animation?). But the difference between before and after got more pronounced. Across all 8 runs, "before" is now 67 dropped frames, stdev 29, "after" is 40 dropped frames, stdev 7.

Hooray for low-hanging fruit? :tada:

Issue #7011, opened 2018-07-23
Reevaluate worker payload and serialization


mourner:
Spent some time investigating noticeable jank (abnormally long frames) that occurs when browsing a typical Mapbox Streets map. Here's a typical flame chart for these frames:

![image](https://user-images.githubusercontent.com/25395/43089578-ae414ef0-8ead-11e8-86f2-01f91d9708b2.png)

In my observation, a lot of the frames where jank occurs are heavily occupied by `Actor` `receive`. Most of it is reported as `self` time (so deserialization isn't the bottleneck). Since there's nothing special happening in that method other than the fact it gets called when getting a response back from a worker thread, I assume that it spends most of the time natively getting the JSON data in the message. This is plausible given that we've seen [very bad performance](https://github.com/mapbox/mapbox-gl-js/issues/1504#issuecomment-179934774) of structured cloning before.

I stringified a typical message that the main thread gets when a tile loads and cut out all values that are transferable (typed arrays) to get a hang of how much structured cloning actually happens. Here's the sample: https://gist.github.com/mourner/ae43942f402f8e5346974945da83a748

The sample is about 500kb of JSON (minified), with tons of nested objects. I think we have a very good opportunity to reduce the jank by looking closer at the payload, with the following things in mind:

- What objects can we fully omit from the payload?
- What objects can we recreate from scratch on the main thread without serialized properties from the worker?
- Is there anything we can do with the serialization/deserialiazation implementation to make the payload smaller?
- What objects can we pack into typed array for transfering? (e.g. feature state id map, symbol instance geometries)

cc @anandthakker @jfirebaugh @asheemmamoowala @ChrisLoer

mourner:
I should also note that we don't capture jank in our benchmarks — they either measures something isolated in a single thread (e.g. layout), or rendering a static view without much tile loads (various paint benchmarks).

ChrisLoer:
Yikes, I didn't realize we were serializing the whole `SymbolFeature` as part of `SymbolInstance`! I don't think there's any reason we need to do that. When we look up features in symbol querying we just use an index into the "rawTileData" to get the feature. I just quickly removed it from the serialization and render/query/flow are still passing. 

I don't know the full history there (presumably it was needed for something at some point), but it looks like when we added expression support we promoted an old list of feature properties to be an entire feature?

[...]

PR #7015, opened 2018-07-24
WIP: Use struct array for feature id maps and deserialize on the main thread

asheemmamoowala:
Follow up from #7011. The feature id map adds significant size to the JSON transfer from WebWorkers to the main thread.

Push all feature id and paint buffer offsets into a struct array and deserialize into an object/map on the main thread.

cc @ChrisLoer @ansis @mourner

### DIFF
--- a/build/generate-struct-arrays.js
+++ b/build/generate-struct-arrays.js
@@ -28,7 +28,8 @@ const typeAbbreviations = {
     'Uint16': 'ui',
     'Int32': 'l',
     'Uint32': 'ul',
-    'Float32': 'f'
+    'Float32': 'f',
+    'Float64': 'lf'
 };
 
 const arraysWithStructAccessors = [];
@@ -174,6 +175,18 @@ createStructArrayType('feature_index', createLayout([
     // the bucket the feature appears in
     { type: 'Uint16', name: 'bucketIndex' }
 ]), true);
+
+createStructArrayType('paint_buffer_offset', createLayout([
+    // the index of the feature in the original vectortile
+    { type: 'Float64', name: 'featureId' },
+    // the index of the feature in the original vectortile
+    { type: 'Uint32', name: 'featureIndex' },
+    // the starting offset of the feature in the paint buffer
+    { type: 'Uint16', name: 'start' },
+    // the end offset of the feature in the paint buffer
+    { type: 'Uint16', name: 'end' }
+]), true);
+
 
 // triangle index array
 createStructArrayType('triangle_index', createLayout([
--- a/src/gl/vertex_buffer.js
+++ b/src/gl/vertex_buffer.js
@@ -22,7 +22,8 @@ const AttributeType = {
     Uint16: 'UNSIGNED_SHORT',
     Int32:  'INT',
     Uint32: 'UNSIGNED_INT',
-    Float32: 'FLOAT'
+    Float32: 'FLOAT',
+    Float64: 'FLOAT'
 };
 
 /**
--- a/src/data/array_types.js
+++ b/src/data/array_types.js
@@ -637,6 +637,56 @@ register('StructArrayLayout1ul2ui8', StructArrayLayout1ul2ui8);
 
 /**
  * Implementation of the StructArray layout:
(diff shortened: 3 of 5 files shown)


build/generate-struct-arrays.js
```diff
@@ -175,6 +176,18 @@ createStructArrayType('feature_index', createLayout([
     { type: 'Uint16', name: 'bucketIndex' }
 ]), true);
 
+createStructArrayType('paint_buffer_offset', createLayout([
+    // the index of the feature in the original vectortile
+    { type: 'Float64', name: 'featureId' },
```
asheemmamoowala: Corresponds to `uint64_t` in native

src/gl/vertex_buffer.js
```diff
@@ -22,7 +22,8 @@ const AttributeType = {
     Uint16: 'UNSIGNED_SHORT',
     Int32:  'INT',
     Uint32: 'UNSIGNED_INT',
-    Float32: 'FLOAT'
+    Float32: 'FLOAT',
+    Float64: 'FLOAT'
```
asheemmamoowala: Added this for Flow. `Float64` shouldn't be used for vertex buffers.

build/generate-struct-arrays.js
ChrisLoer: I still confused by this. What's the underlying type defined in the vector tile spec? A Float64 can't represent every uint64_t.

asheemmamoowala: The [vector tile spec](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features) doesn't define a type AFAICT. Native vt libraries use a `unint64_t` to represent ids. The vectortile.js library uses a `number` type to represent the id, which I believe has the same precision issue as `Float64` does here.

ericfischer: The feature IDs in the `.proto` file of the current vector tile spec are [64-bit unsigned integers](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto#L32), although this should have been mentioned in the text too. GeoJSON prefers string IDs, so for VT3 we are discussing [generalizing IDs to any representable type](https://github.com/mapbox/vector-tile-spec/pull/120) and would appreciate input on what is feasible in client code.

ChrisLoer: Huh yeah JS's `Number.MAX_SAFE_INTEGER == 2^53 - 1` so technically that's a bug in our current implementation... although apparently we've gotten this far without hearing anyone complain about the feature ID 2^54 not working.

asheemmamoowala: @ChrisLoer that limit is for Integers. But `Number` is always a double, with `Number.MAX_VALUE == 1.79E+308` . In order to use a large Number in a typed array, as is the case here, the array type should be large enough to accommodate `uint64_t`. `Float64` is the only type that can do that here.

asheemmamoowala: Created #7020 to track the feature ID bit-size discrepancy.

asheemmamoowala:
I'm not sure if this change has performance benefits. Even though it does reduce the JSON (non-TypedArray) size for higher zoom tiles, I couldnt measure or perceive improvements (less jank) in panning or zooming animations. 

Pan and Zoom animations over 10s both seem to drop more frames (as compared to `master`). My guess is that deserializing the `idMap` in `ProgramConfiguration` affects the main thread, but I havent figured out a way to measure that additional deserialization cost vs any savings from transferring the id map as a `TypedArray.`

Tested with the map centered over a dense part of NYC (`[-73.95327, 40.58377]`) with the 3d-buildings dataset that includes unique ids for all building polygons.

@mourner @ChrisLoer any thoughts on how else to measure this change's performance impact?

ChrisLoer:
> any thoughts on how else to measure this change's performance impact?

Can you do before/after JS profiler captures of the same 10 second animation, and look for changes in time spent in `Actor.receive` vs time spent in the feature deserialization code?

Might also be interesting to look at the flame chart, find long frames, and just look for qualitative differences before/after.

Issue #6963, opened 2018-07-13
Publish Flow type definitions alongside compiled bundle


anandthakker:
There isn't currently a straightforward way for a downstream consumer of Mapbox GL JS to benefit from the Flow type annotations that exist throughout our source code (and, specifically, on our public API).  We should find a way to distribute `dist/mapbox-gl.js.flow` alongside the main bundle.

https://github.com/facebook/flow/issues/5871 suggests that the `flow gen-flow-types` tool may get deprecated soon and so we probably don't want to rely on that. 

Since we _do_ include `src/` in our npm package, we could use the approach mentioned in https://github.com/facebook/flow/issues/5871#issuecomment-376018369 -- i.e., have `dist/mapbox-gl.js.flow` look like:

```js
// @flow
export * from '../src/index.js';
```

This almost works, except that the downstream package would need to add many of the ignores that we use: https://github.com/mapbox/mapbox-gl-js/blob/f55e5489e249da575b7cd0ff80dc71ac970b714d/.flowconfig#L2-L12

shanshanzhu:
Can you try this in the .flowconfig? `<PROJECT_ROOT>/node_modules/.*`

PR #7031, opened 2018-07-26
Move local type definitions out of flow-typed

anandthakker:
Prepares for #6963


anandthakker:
Note also that I replaced our GeoJSON declarations with https://github.com/bigslycat/flow-geojson, which seems like a pretty well-written set of Flow types for GeoJSON.

jfirebaugh:
Eh, I'm not a fan of those GeoJSON definitions. They're much more complicated than is necessary, and there are at least a couple of issues (feature geometry should be nullable, `JSONArray` type is wrong).

anandthakker:
Dang, I missed those.  Alright, alternative: we could keep the ones we had and publish them as a separate npm pacakge `@mapbox/geojson-types`

anandthakker:
> Edit: also properties should not be nullable

@jfirebaugh looks like the spec allows null for `properties`:

>      A Feature object has a member with the name "properties".  The
>      value of the properties member is an object (any JSON object or a
>      JSON null value).
https://tools.ietf.org/html/rfc7946#section-3.2

anandthakker:
@jfirebaugh updated

PR #7033, opened 2018-07-26
Add customAttribution option to AttributionControl

mklopets:
Adds a way to add custom attribution strings to AttributionControl. Also exposes the `customAttribution` option on Map.

### DIFF
--- a/src/ui/control/attribution_control.js
+++ b/src/ui/control/attribution_control.js
@@ -7,7 +7,8 @@ import config from '../../util/config';
 import type Map from '../map';
 
 type Options = {
-    compact?: boolean
+    compact?: boolean,
+    customAttribution?: string | string[]
 };
 
 /**
@@ -16,6 +17,7 @@ type Options = {
  * @implements {IControl}
  * @param {Object} [options]
  * @param {boolean} [options.compact] If `true` force a compact attribution that shows the full attribution on mouse hover, or if `false` force the full attribution control. The default is a responsive attribution that collapses when the map is less than 640 pixels wide.
+ * @param {string | Array<string>} [options.customAttribution] String or strings to show ahead of any other attributions.
  * @example
  * var map = new mapboxgl.Map({attributionControl: false})
  *     .addControl(new mapboxgl.AttributionControl({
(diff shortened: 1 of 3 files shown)


src/ui/control/attribution_control.js
```diff
@@ -111,6 +113,13 @@ class AttributionControl {
     _updateAttributions() {
         if (!this._map.style) return;
         let attributions: Array<string> = [];
+        if (this.options.customAttribution) {
+            if (this.options.customAttribution instanceof Array) {
```
mourner: Let's use `Array.isArray` instead — `instanceof Array` can sometimes fail.

src/ui/map.js
```diff
@@ -61,6 +61,7 @@ type MapOptions = {
     container: HTMLElement | string,
     bearingSnap?: number,
     attributionControl?: boolean,
+    customAttribution?: string | string[],
```
mourner: Let's use `Array<string>` form for consistency.

mklopets:
btw, I noticed that running build locally adds `"scheme"?: "xyz" | "tms"` to `VectorSourceSpecification` in `flow-typed/style-spec.js` – is this intentionally not committed in?

mourner:
@mklopets not intentionally, don't worry about it — we'll commit.

src/ui/control/attribution_control.js
```diff
@@ -16,6 +17,7 @@ type Options = {
  * @implements {IControl}
  * @param {Object} [options]
  * @param {boolean} [options.compact] If `true` force a compact attribution that shows the full attribution on mouse hover, or if `false` force the full attribution control. The default is a responsive attribution that collapses when the map is less than 640 pixels wide.
+ * @param {string | Array<string>} [options.customAttribution] String or strings to show ahead of any other attributions.
```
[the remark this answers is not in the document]
mklopets: Actually, this is incorrect, everything is sorted by length anyway – I'll change the doc. Thanks!

mklopets:
@mourner could we get this released? Would be much appreciated – needed in production.

mourner:
@mklopets unfortunately not at the moment — we're on a monthly release cadence so you should expect the new version in about 3 weeks. However, you can always build the `master` version and use it in production.

mklopets:
Oh, alright :/ I tried looking up the release cadence from the docs here in the repo but couldn't find anything. Might be good to add a little note to `CONTRIBUTING.md` about it.

Thanks for the quick response(s)!

Issue #3996, opened 2017-01-17
Add support for rendering 3d models 


lucaswoj:
Continued from https://github.com/mapbox/DEPRECATED-mapbox-gl/issues/17

> We are adding support for [simple extrusions](https://github.com/mapbox/mapbox-gl-style-spec/issues/456). For buildings, there are some complex and high-profile landmarks that can't be rendered accurately with simple extrusion:
> 
> ![eiffel-tower](https://cloud.githubusercontent.com/assets/5607844/16136340/b62eef22-33f7-11e6-9982-736ec86cbd42.png)
> _Haiii, Eiffel Tower._
> 
> For these features, it seems like we would want to include 3D models. How would/could this work? How much of a lift is this? 
> 
> /cc @lbud @ajashton @mapbox/gl 

cc @mourner @tmcw @jfirebaugh

allthesignals:
My team is interested in learning more about the status of this feature - we are also in a place where simple extrusions will not work, and we have a wealth of 3d model data that we think would fit better inside our MapboxGL stack. Any chance there's direct access to the scene interface? Is a pathway for this related to https://github.com/mapbox/mapbox-gl-js/issues/6456? Thanks!

lbud:
> Any chance there's direct access to the scene interface? Is a pathway for this related to #6456?

Exactly @allthesignals, custom layers (#6456) will be the soonest you'll be able to accomplish this, and that feature will likely land ~this summer — native Mapbox GL support for 3D models isn't on our short- or medium-term roadmap at this time.

allthesignals:
Thanks, @lbud - I'll keep an eye on this PR here: https://github.com/mapbox/mapbox-gl-js/pull/6124

andrewharvey:
Over on the PR I posted some example code which helps place 3D models at given long, lats

https://github.com/mapbox/mapbox-gl-js/pull/7039#issuecomment-410621717

![selection_999 017](https://user-images.githubusercontent.com/117278/43704356-bb7296ec-99a2-11e8-93ec-a6ff0083cb55.png)

allthesignals:
Fabulous - thanks @andrewharvey!

andrewharvey:
I'll closing this ticket given custom layers support adding 3D models to a Mapbox GL JS map.

An official example will be followed up in #7330

3D terrain mesh is at #1489

femski:
[...]

PR #7039, opened 2018-07-27
implement custom style layers

ansis:
This implements https://github.com/mapbox/mapbox-gl-js/issues/6456. It follows the approach taken in -native and keeps the scope fairly tight. It's an alternative to #6124.

Custom layers can be created by implementing `CustomLayerInterface` and can be added to the map with `map.addLayer(customLayer)`. An [example](https://github.com/mapbox/mapbox-gl-js/blob/custom-layers/debug/custom3d.html). Another [example using ThreeJS](https://github.com/mapbox/mapbox-gl-js/blob/custom-layers/debug/threejs.html).

### The interface

```js
type CustomRenderMethod = (gl: WebGLRenderingContext, matrix: Array<number>) => void;

type CustomLayerInterface = {
    id: string,
    type: "custom",
    renderingMode: "2d" | "3d",
    render: CustomRenderMethod, // for rendering directly into the texture
    prerender: ?CustomRenderMethod, // for rendering into a texture
    onAdd: ?(map: Map, gl: WebGLRenderingContext) => void,
    onRemove: ?(map: Map) => void
}
```

Custom layers trigger repaints with the newly exposed `map.triggerRepaint()`.

### The matrix
The matrix projects spherical mercator coordinates to gl coordinates. The spherical mercator coordinate  [0, 0] represents the top left corner of the mercator world and  [1, 1] represents the bottom right corner. The z coordinate is conformal. A box with identical x, y, and z lengths in mercator units would be rendered as a cube.

### Shared depth
Custom 3D layers share depth space with our 3D layers. `/debug/threejs.html` contains an example of ThreeJS sharing depth space with fill extrusions:

screen shot 2018-07-27 at 12 51 43 pm

### API design questions

Currently a layer implements `render(...)` if it wants to render a 2D layer and `render3D(...)` if it wants to render in 3D. Layers can implement both. Would it be better to only have `render(...)` and differentiate based on something else? For example, having two types: `"type": "custom-3d"` and `"type": "custom"`

When a custom layer takes over the gl context the state is in a mostly undefined state. Should we reset all (or a subset) of the state to default values?

### Things left for later
- an api for projecting to/from mercator and custom coordinate spaces
- convenience methods for producing matrices for custom defined coordinate spaces
- exposing support or helpers for tiled rendering

### Checklist 
 - [ ] post benchmark scores

@davidmanzanares @ibgreen does this meet your basic needs?


andrewharvey:
This is huge in what it will unlock! :heart: Thanks @ansis 

As someone writing a custom layer do I need to worry about if I'm drawing off screen or not? Is that my responsibility or does GL JS worry about that for me?

Surely there's performance considerations with your THREE example if I'm making boxes which are halfway across the world and not in the current view.

> an api for projecting to/from mercator and custom coordinate spaces

I think the ultimate convenience is working with WGS84 `longitude, latitude`. For example from your https://github.com/mapbox/mapbox-gl-js/blob/custom-layers/debug/threejs.html example many use cases would be I have the long,lat coordinates and want to use THREE to draw something here.

mourner:
@ansis the API and the implementation look really great! Exploring it in `custom-layers-wind`, having some troubles...

- No layers draw after the wind layer. E.g. if I place it before `building`, no layers that come after `building` will be drawn. So the wind layer somehow corrupts the gl state I suppose.
- I still need to do `gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);` in `onAdd`. I guess we should reset the context before `onAdd` too, and also expose a method to reset it, e.g. if the GL context is initialized asynchronously.
- The map doesn't draw properly if I don't do `gl.viewport(0, 0, gl.canvas.width, gl.canvas.height)` at the end of the render call — I guess viewport isn't being properly reset at the beginning of GL rendering?
- The particles now exhibit some weird striping artifacts that I haven't noticed before (especially noticeable when zooming to a certain zoom), and I have no idea what might be causing them and why we haven't seen them with the previous implementation.

src/style/style_layer/custom_style_layer.js
```diff
@@ -0,0 +1,182 @@
...
+ * top left corner of the mercator world and `[1, 1]` represents the bottom right corner.
+ */
+export type CustomLayerInterface = {
+    id: string,
+    type: "custom",
+    render: CustomRenderMethod,
+    render3D: CustomRenderMethod,
+    onAdd: (map: Map, gl: WebGLRenderingContext) => void,
+    onRemove(map: Map): void
+}
```
anandthakker: This type definition says that `render` and `render3D` are both defined. Should they both be typed `?CustomRenderMethod` instead?

src/style/style.js
```diff
@@ -554,21 +557,26 @@ class Style extends Evented {
...
         }
 
-        if (typeof layerObject.source === 'object') {
-            this.addSource(id, layerObject.source);
-            layerObject = clone(layerObject);
-            layerObject = (extend(layerObject, {source: id}): any);
-        }
+        let layer;
+        if (layerObject.type === 'custom') {
+            layer = createStyleLayer(layerObject);
```
anandthakker: Since this isn't going through the style spec validation path, the provided implementation needs to be validated 1) for layer id uniqueness, and 2) conformance to the CustomLayerInterface.  Probably 1) makes sense to do here, and maybe 2) in the CustomStyleLayer constructor?

ryanhamley:
> I think the ultimate convenience is working with WGS84 longitude, latitude. For example from your https://github.com/mapbox/mapbox-gl-js/blob/custom-layers/debug/threejs.html example many use cases would be I have the long,lat coordinates and want to use THREE to draw something here.

Backing up @andrewharvey's point here with a specific example, this is exactly the kind of thing that would have been a game changer at my previous company. We had all these meshes and point clouds of real-world sites that we rendered with Three.js but we weren't able to geolocate them on a map so we simply rendered them in their own separate viewer. Being able to place them on a map would have given the users a ton of added value.

markusjohnsson:
This looks very promising! 

It looks as if it will allow us to move our custom canvas-based layers to this API instead. I think we will need to keep some of our scene handling logic though, like rendering two world widths to allow graphics crossing the +/-180 meridian. It would be nice if that could be handled by mapbox in the future, though I don't know how it would work.

markusjohnsson:
I tested this branch to feel it out, but I got problems with label layers: 

image

It only happened after I added a texture (using `THREE.TextureLoader`) and subsequently used it in a shader (`THREE.ShaderMaterial`). Prior to using textures it worked fine.

andrewharvey:
> Backing up @andrewharvey's point here with a specific example, this is exactly the kind of thing that would have been a game changer at my previous company. We had all these meshes and point clouds of real-world sites that we rendered with Three.js but we weren't able to geolocate them on a map so we simply rendered them in their own separate viewer. Being able to place them on a map would have given the users a ton of added value.

This function works for me to transform from WGS84 long, lat into a unit vector, anchored at the top left, as needed by this PR.

```js
 var fromLL = function (lon,lat) {
     // derived from https://gist.github.com/springmeyer/871897
     var extent = 20037508.34;
 
     var x = lon * extent / 180;
     var y = Math.log(Math.tan((90 + lat) * Math.PI / 360)) / (Math.PI / 180);
     y = y * extent / 180;
 
     return [(x + extent) / (2 * extent), 1 - ((y + extent) / (2 * extent))];
 }
```

Providing a helper function for this would make it much easier for people to do #3996 (custom 3d models).

Here's an example built on top of @ansis's code, but placing a model with long, lat coords. We 

https://bl.ocks.org/andrewharvey/7b61e9bdb4165e8832b7495c2a4f17f7

selection_999 017

davidmanzanares:
Hi,

First of all, thank you so much for pushing this, this is an amazing feature.

Regarding your question, we've only made one small PoC of integrating this with CARTO VL, but it looked quite good. We found two small issues regarding WebGL state sharing( https://github.com/mapbox/mapbox-gl-js/pull/7080 and https://github.com/mapbox/mapbox-gl-js/pull/7081) but apart from that, we haven't seen any other rendering issue.

However, we do have seen a performance regression compared to our previous approach. After some profiling, I saw that the `actor.js:58 receive()` function took much more time when using this integration: from a total time of 9% to a total time of 19%. We also saw an important increase in execution time in `web_worker_transfer.js:197 deserialize()`. To be honest, I don't know if this makes any sense, I don't see any WebGL related calls in those functions. I could be seeing another performance regression not related to this branch since our integration is not aligned with `master`. If you think this is the case, we could update our integration with `master` and see if the regression is there.

We also have thought of another way to share the WebGL state in a safe way without changing this new API. I think that monkey-patching the WebGL context to provide a patched version to custom layers could be a possible way of knowing when you need to mark some state as dirty. This will remove the need of calling `Context.setDirty` in every frame and state would be marked as dirty as needed. The number of WebGL drawing calls could be decreased that way, but I don't know if the performance advantage is worthwhile.

Thanks!

mourner:
@davidmanzanares yes, the performance regression you're seeing is likely in master too — see #7011 for more context. There's also #7015 as an attempt to win some perf back, but we're not sure it's the right approach yet.

andrewharvey:
I ran into an issue where a THREE.js scene wasn't showing up with an empty map style:

```
style: {
    version: 8,
    sources: {},
    layers: []
}
```

I needed to add a background layer first. It might have been a problem on my end, but I'm not sure.

TripleMap:
@andrewharvey, 

Your example inspired me to use models instead of circles. But the question arises. I can not find the input point in  src mapbox, to work with the current source. For example, I have about 20 point (lat, lng, height). How can I get data from source? (I belive, that it is possible).

ansis:
I've pushed a bunch of fixes including some api changes:

- the`render3D` method was removed and a `"renderingMode": "2d" | "3d"` property was added
- an optional `prerender` method was added so that users can separate rendering into textures from rendering into the main framebuffer. This was added to work around cases where the user changes the blend and depth settings when drawing into the texture and would then need to reset these settings when drawing onto the map

Two render tests were added.

> I think the ultimate convenience is working with WGS84 longitude, latitude. For example from your https://github.com/mapbox/mapbox-gl-js/blob/custom-layers/debug/threejs.html example many use cases would be I have the long,lat coordinates and want to use THREE to draw something here.

Yep! I think we should add this but outside of this pr so that we can get the api right for that.

> I think we will need to keep some of our scene handling logic though, like rendering two world widths to allow graphics crossing the +/-180 meridian. It would be nice if that could be handled by mapbox in the future, though I don't know how it would work.

Yeah, this is tricky to implement but it's outside the scope of the current custom layers work. I'm not sure how this would work on the mapbox-gl side.

> I tested this branch to feel it out, but I got problems with label layers:
@markusjohnsson thanks for the test case. It should be fixed!

> We also have thought of another way to share the WebGL state in a safe way without changing this new API. I think that monkey-patching the WebGL context to provide a patched version to custom layers could be a possible way of knowing when you need to mark some state as dirty.

@davidmanzanares yep, that could be a good approach! If the `setDirty()` approach turns out to be a performance or safety problem we should try this.

> I ran into an issue where a THREE.js scene wasn't showing up with an empty map style:

@andrewharvey Thanks for catching this. It should now be fixed!

Jesus89:
Hi @ansis,

I have tested the integration of this branch with CARTO VL, adding the new interface (attributes and functions) to our `carto.Layer` object to match the API.

I'm a bit worried about exposing these attributes and functions in the `Layer` because a user can be confused using them as part of the public API.

Could you add something to differentiate this internal API from the public API, for example, adding '$' at the beginning?

```
type CustomLayerInterface = {
    $id: string,
    $type: "custom",
    $render: CustomRenderMethod,
    $render3D: CustomRenderMethod,
    $onAdd: (map: Map, gl: WebGLRenderingContext) => void,
    $onRemove(map: Map): void
}
```

Thanks for your work!

Regards.

ansis:
> I'm a bit worried about exposing these attributes and functions in the Layer because a user can be confused using them as part of the public API.
>
> Could you add something to differentiate this internal API from the public API, for example, adding '$' at the beginning?

@Jesus89 from the mapbox-gl-js perspective this will be a public API that. I'm not sure if I'm understanding your issue correctly, but if adding these members to `carto.Layer` is an issue, could you implement them on a separate object and then add that to the map, instead of the carto.Layer itself?

markusjohnsson:
@ansis I think that what @Jesus89 refers to is that if you get a reference to the layer you would be able to make a call to one of these methods e.g. `map.getLayer().render()`, in that sense, the methods should not be public (but rather `protected`, in the TypeScript/C#/Java sense).

src/geo/transform.js
```diff
@@ -560,6 +565,10 @@ class Transform {
         mat4.rotateZ(m, m, this.angle);
         mat4.translate(m, m, [-x, -y, 0]);
 
+        // The mercatorMatrix can be used to transform points from mercator coordinates
+        // ([0, 0] nw, [1, 1] se) to GL coordinates.
+        this.mercatorMatrix = mat4.scale([], m, [this.worldSize, this.worldSize, this.worldSize]);
```
mourner: Any reason we're using regular JS array here for the matrix instead of typed one like in other places, e.g. `new Float64Array(16)`?

debug/custom3d.html
```diff
@@ -0,0 +1,117 @@
...
+        const d = 0.125;
+
+        const vertexArray = new window.Float32Array([
+            x, y, 0,
+            x + d, y, 0,
+            x, y + d, z,
+            x + d, y + d, z,
+            x, y + d + d, 0,
+            x + d, y + d + d, 0]);
+        const indexArray = new window.Uint16Array([
```
mourner: Nit: why do we access those with `window`? If linter complains, this probably needs a fix on the lint rules side.

src/render/draw_custom.js
```diff
@@ -0,0 +1,58 @@
...
+            implementation.render(context.gl, painter.transform.customLayerMatrix());
+
+            context.setDirty();
+            painter.setBaseState();
+        }
+
+    } else if (painter.renderPass === 'translucent') {
+
+        if (implementation.renderingMode === '3d') {
+            drawOffscreenTexture(painter, layer, 1);
```
mourner: Let's add a way to control this — e.g. instead of `1`, `implementation.opacity === undefined ? 1 : implementation.opacity`.

src/render/draw_custom.js
```diff
@@ -0,0 +1,58 @@
...
+    const context = painter.context;
+    const implementation = layer.implementation;
+
+    if (painter.renderPass === 'offscreen') {
+
+        const prerender = implementation.prerender;
+        if (prerender) {
+            painter.setCustomLayerDefaults();
+
+            prerender.call(implementation, context.gl, painter.transform.customLayerMatrix());
```
mourner: Nit: no need for `call` here, `implementation.prerender(...)`

src/render/draw_custom.js
```diff
@@ -0,0 +1,58 @@
...
+
+            context.setDirty();
+            painter.setBaseState();
+        }
+
+        if (implementation.renderingMode === '3d') {
+            painter.setCustomLayerDefaults();
+
+            prepareOffscreenFramebuffer(painter, layer);
+            implementation.render(context.gl, painter.transform.customLayerMatrix());
```
mourner: Should it be possible to have both `prerender` and `renderingMode === '3d'`?

src/style/style.js
```diff
@@ -847,7 +867,9 @@ class Style extends Evented {
             glyphs: this.stylesheet.glyphs,
             transition: this.stylesheet.transition,
             sources: mapObject(this.sourceCaches, (source) => source.serialize()),
-            layers: this._order.map((id) => this._layers[id].serialize())
+            layers: this._order.map((id) => this._layers[id])
+                .filter((layer) => layer.type !== 'custom')
+                .map((layer) => layer.serialize())
```
mourner: This is duplicated with `_serializeLayers`. Let's DRY up. Maybe also write with a `for` loop / `push` for less short-lived allocations?

src/geo/transform.js
ansis: Yep, the gl uniform commands (`uniformMatrix4fv(...)`) don't accept `Float64Array`. Using them would force the user to convert them into a regular array or `Float32Array`. I'm not sure there would be any advantage to using `Float64Array`s instead of regular arrays.

We should review whether there is a good reason to be using Float64Array elsewhere. I know we've been doing it for a long time but I'm not sure if there is an actual advantage.

src/render/draw_custom.js
ansis: I'd rather not expose this right now.

Opacity for our 3d layers is problematic because we can't get ordering correct for semi-transparent layers. I think we may have to change how this is done and I'd rather not expose something we would have to break.

src/render/draw_custom.js
ansis: This was done to satisfy flow complaining about `implementation.prerender` being potentially undefined. The alternative is a cast through `any`. Would this be better?

mourner: @ansis this is surprising — Flow should detect the check in `if (implementation.prerender)` and not complain in this case.

jfirebaugh: https://flow.org/en/docs/lang/refinements/#toc-refinement-invalidations

SakeenaQureshee:
@markusjohnsson wrt [this](https://github.com/mapbox/mapbox-gl-js/pull/7039#issuecomment-410257180) I was facing the same issue with the label layers, did you come across any resolution for it?

markusjohnsson:
@SakeenaQureshee no, I just tested with the latest changes, and I still have the same problem (at least at certain zoom levels).

andrewharvey:
> @SakeenaQureshee no, I just tested with the latest changes, and I still have the same problem (at least at certain zoom levels).

I think it affects other textures too including raster sources.

markusjohnsson:
image

Actually, I have to take back my last comment.. perhaps I had a cache issue or something, because now it works, both raster and labels.

andrewharvey:
> Actually, I have to take back my last comment.. perhaps I had a cache issue or something, because now it works, both raster and labels.

Heh, actually me to, with the latest code it's all working now.

measuredweighed:
Thanks for all of the hard work on this so far, it's looking really promising. We're tracking this and have begun experimenting with custom layer rendering. Is there an ETA (however vague) for when/if this is likely to be merged/considered ready for production use?

ansis:
@Jesus89 thanks for the feedback but we've decided to leave the interface as is

@measuredweighed the current implementation should be pretty stable. The first production release it will be included in will be released October 10th.

Jesus89:
OK. Thanks for your work! :+1:

tibotiber:
This feature is huge, thanks for the amazing work. A quick question: if I wanted to add a lot of independent objects (e.g. threejs 3D models), each with separate locations, would adding a custom layer (with its own scene, camera and renderer) for each object pose any performance issue? Or would you recommend I create a single custom layer in which I add the different objects computing location based on camera matrix and intended position?

In the second case, I'd imagine the location computation will be a bit more complex. Any pointer on this? Probably something like `gl_Position = projectionMatrix * viewMatrix * modelMatrix * vec4( position, 1.0 );` as seen in https://github.com/mrdoob/three.js/issues/1188#issuecomment-3666286 but not sure where I'd get `viewMatrix` from?

Cc @pviejo

tibotiber:
Update on my previous comment. I have gone the second way and implemented one single custom layer centered in `(lat, lng)=(0,0)` in which I add multiple objects for which the position is computed using the `fromLL` method shared by @andrewharvey. Works really well and I believe more efficient than having a separate renderer for each object. Thanks again for the great work 👍.

tibotiber:
Re-update FWIW. threejs would have rendering artifacts on large scenes, so it's better locating the custom layer scene near your area of interest and offsetting objects in the scene accordingly.

wandergis:
> Update on my previous comment. I have gone the second way and implemented one single custom layer centered in `(lat,lng)=(0,0)` in which I add multiple objects for which the position is computed using the `fromLL` method shared by @andrewharvey. Works really well and I believe more efficient than having a separate renderer for each object. Thanks again for the great work 👍.

Does Raycaster works also works well ?

tibotiber:
@wandergis no, completely out of sync 😬

andrewharvey:
@rveciana See the code snippet in my comment at https://github.com/mapbox/mapbox-gl-js/pull/7039#issuecomment-410621717

lileialg:
For the first,this feature is very great,I can upload my shader to mapbox-gl,So I can display unlimited visualization; but I have a question, when I increase the zoom to 17,18,19,20,21,22....., the position of point is not stable, the position of points are shaking with the zoom increasing or map rotating, the attachment is my code file,I am looking for your response, thanks very much
[custom3d-4.txt](https://github.com/mapbox/mapbox-gl-js/files/2532870/custom3d-4.txt)

measuredweighed:
@lileialg This is related to a lack of floating point precision. I opened an issue about this (#7268) and also provided one [possible workaround](https://github.com/mapbox/mapbox-gl-js/issues/7268#issuecomment-420722456).

lileialg:
@ansis @markusjohnsson I covert position to pixel in javascript instead of in shader script,It looks like working fine.
[custom3d-13.txt](https://github.com/mapbox/mapbox-gl-js/files/2537790/custom3d-13.txt)

bhison:
> Re-update FWIW. threejs would have rendering artifacts on large scenes, so it's better locating the custom layer scene near your area of interest and offsetting objects in the scene accordingly.

Thanks for your above comments, they've really set me on the right path for rendering multiple 3D objects across my map. I've started from the [Add a 3D model](https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/), set the center of my map as the translation for the Three projection matrix and then applied an offset transform on the object to have it positioned correctly. 

My model is now in the correct position, with the right scale, a slightly off rotation (which I assume is the symptom of floating point issues), but the major issue is for some reason it now looks like the normals are inverted on my model. I was wondering if you came up against anything like this? I've tried fiddling with a lot of options but I'm perplexed, I can't see how this could be happening.

tibotiber:
@bhison glad the comments could help :). And no I haven't met any issue concerning normals. Sorry, can't be of much help on this.

bhison:
> @bhison glad the comments could help :). And no I haven't met any issue concerning normals. Sorry, can't be of much help on this.

Thanks for your response anyway. Knowing you haven't had the same issue is kind of helpful from a process of elimination perspective anyway.

PR #7040, opened 2018-07-27
updating example to forward geocoding with custom data

kbauhaus:
## Launch Checklist

Changing this example from reverse geocoding to forward geocoding with custom data. This is a more common use case based on feedback from users (via support tickets).

cc @andrewharvey if you think we should keep the reverse geocoding example in addition to this forward geocoding version, let me know and I can add as a separate example.


andrewharvey:
hey @kbauhaus I agree that forward geocoding based on your own data eg. from a Mapbox Dataset or other source is a common use case, so glad to see you put together this example!

The original example which accepts `lat,long` style queries is still something I've encountered people wanting to do, so I'm in favour of having this as a new example instead of replacing the current one.

docs/pages/example/mapbox-gl-geocoder-local-geocoder.html
```diff
@@ -57,11 +60,167 @@
...
+
+
+function forwardGeocoder(query) {
+    var matchingFeatures = [];
+    for (var i = 0; i < customData.features.length; i++) {
+        var feature = customData.features[i];
+        // handle queries with different capitalization than the source data by calling toLowerCase()
+        if (feature.properties.title.toLowerCase().search(query.toLowerCase()) !== -1) {
+            // add a tree emoji as a prefix for custom data results
+            feature['place_name'] = '🌲 ' + feature.properties.title;
```
andrewharvey: I'd add in a note about this being in the carmen geojson format https://github.com/mapbox/carmen/blob/master/carmen-geojson.md to explain these properties.

kbauhaus:
Thanks for the reviews @andrewharvey and @mollymerp!

PR #7043, opened 2018-07-28
Enable auto generation of ids for GeoJSON sources

### LINKED ISSUE #6849: Automatically generate feature `id` values for geojson & cluster sources
## Motivation

Examples using map.setFeatureState() like this https://www.mapbox.com/mapbox-gl-js/example/hover-styles/ require that a user include an integer parent `id` value in their geojson or vector tile source.

## Design Alternatives

* Create more documentation about parent `id` requirement to use featureStates

## Design

* Automatically generate parent a `id` value for a geojson source in geojson-vt, so that all geojson source features support featureStates.  
* Automatically generate a `parent` id for `clustered` geojson datasources

* Not sure if a similar approach is possible for vector tile sources.

### Mock-Up

Geojson-vt auto-generates feature `id` values for geojson and cluster sources.

### Concepts

* Vector tile `id` values
* The difference between how vector tile sources and geojson sources are created and how they support featureStates.

asheemmamoowala:
Closes #6849.

Upgrades geojson related libraries to include auto-generation of ids for geojson features and clusters.

cc @ryanbaumann


andrewharvey:
It looks like this is only done when you initialise a new GeoJSON source, if you subsequently use `setData` I assume `generateId` won't apply `id`s then? I think the behaviour should be made explicit somehow in the docs.

asheemmamoowala:
> It looks like this is only done when you initialise a new GeoJSON source

@andrewharvey it should generate ids every time. `workerOptions` is built up in the constructor and is reused in calls to `_updateWorkerData`.

I just realized that this is somewhat of a problem though. `id`'s are based on the index of elements in the `features` array, so they can change on subsequent calls to `setData`. This would mean any previously cached feature ids would point to a different feature after the new tiles are rendered. This would need documentation, but I also think that it is not a good API experience.

andrewharvey:
> I just realized that this is somewhat of a problem though. id's are based on the index of elements in the features array, so they can change on subsequent calls to setData. This would mean any previously cached feature ids would point to a different feature after the new tiles are rendered.

Previously cached internally to GL JS or by the user outside GL JS? I should be fine if ID's change since you're resetting the data anyway, and anyone who needs stable IDs can provide their own.

asheemmamoowala:
> Supercluster doesn't have an equivalent of generateId option, so if the original GeoJSON doesn't have id in features, they will only be present for the clusters, not the points. Should we add it there as well?
@mourner -> https://github.com/mapbox/supercluster/issues/97

asheemmamoowala:
> Previously cached internally to GL JS or by the user outside GL JS? 

@andrewharvey this would be both. `setData` needs to clear the corresponding feature state cache internally, and any feature ids cached or used externally will also need to be invalidated.

hoogw:
How to use auto-generated-ID?
feature.id ?

asheemmamoowala:
@hoogw You can enable id generation by setting the [generateId](https://docs.mapbox.com/mapbox-gl-js/style-spec/#sources-geojson-generateId) flag on your GeoJSON source.

hoogw:
@asheemmamoowala 
excellent! well done !

What if I have  **type: vector** instead of type: geojson  

I want to use map.setFeatureState(), have to have a unique id, 
but the source from mapbox-tilesets ( or mapserver .... .pbf) do not have a id field, no unique value column.

How to get "generateid" ?

hoogw:
this works fine.


                ` map.addSource("_geojson_source", {
              "type": "geojson",
              "data": ___GeoJson,
              
              
              // https://docs.mapbox.com/mapbox-gl-js/style-spec/#sources-geojson-generateId
              //https://github.com/mapbox/geojson-vt/pull/109
              //https://github.com/mapbox/mapbox-gl-js/pull/7043
              "generateId":true
              
              
              });`



 "generateId":true  also works in vector source?
If not, how to deal with vector source, to use map.setFeatureState() ???

                       `map.addSource("_tilesets_source",  {
                        type: 'vector',
                        url: 'mapbox://hoogw.aiuo72ew'

                         // ????
                           "generateId":true   // ??????
                        });`


and this ???


                     map.addSource("_tileserver_source",  {
                        type: 'vector',
                      
                        "tiles": ["http://localhost:10/tileserver/news-freedom-rank-2019/{z}/{x}/{y}.pbf"],
                        
                          // ????
                       "generateId":true,   // ??????

                        "minzoom": 0,
                        "maxzoom": 22
                        
                        });

PR #7047, opened 2018-07-30
Style benchmarks

ryanhamley:
## Launch Checklist

    - Adds the ability to run benchmark tests against Mapbox styles by supplying a styleURL
    - Does not change functionality of existing benchmark suites that run on various versions of the code

`Layout` and `Paint` tests run once per tile over a list of tiles/locations (center and zoom) supplied by @nickidlugash and the Carto team to exercise a number of different relevant areas of style performance (CJK text, urban, rural, zoom levels, etc).
screen shot 2018-07-30 at 12 20 07 pm
screen shot 2018-07-30 at 12 20 38 pm

`QueryPoint`, `QueryBox`, `StyleLayerCreate` and `StyleLayerValidate` run the same as the existing non-style benchmark suite with the exception that the two Query tests use the list of locations that Carto defined. The other benchmarks are not used by the style suite.
screen shot 2018-07-30 at 12 20 58 pm


bench/lib/benchmark.js
```diff
@@ -11,8 +12,20 @@ export type Measurement = {
...
+    zoom: number,
+    center: LngLatLike
+};
+
 class Benchmark {
-    constructor() {
+    styleURL: string;
+    locations: ?Location;
+
+    constructor(styleURL: string, locations: ?Location) {
```
ChrisLoer: I'm confused by the plurality/singularity of `locations: ?Location`. The type and the way it's used in paint/layout makes it seem like this should be a singular "location" member. Where it's plural (non-paint/layout), it seems like the type would actually be `Array<Location>` (although I guess there's also the separate case of one "location" with multiple tiles, that doesn't look like what gets returned from `import { locations } from './lib/style_locations';`

bench/lib/parameters.js
```diff
@@ -0,0 +1,33 @@
+const isStyleBench = process.env.STYLE_BENCHMARK;
```
ChrisLoer: Why do we need two separate "isStyleBench" flags set? (`process.env.STYLE_BENCHMARK` and also `?style_bench=true`) Shouldn't one of them be enough?

ryanhamley: Basically, because this flag needs to be available both in `index.html` and all the benchmark `.js` files. You can see in `index.html` that we need to somehow determine the type of test in order to know how to set the tests up. The only way I could think to make this available in a script tag on an html page was to put it in the URL. I recognized this redundancy and I hate it, so I will gladly change it if we can think of a simpler way to make the test type parameter available in both places.

https://github.com/mapbox/mapbox-gl-js/blob/69335765d0f125f56f9f7ca8484ea903b70d992f/bench/index.html#L20

bench/lib/benchmark.js
ChrisLoer: I would guess that it's easy to get flow errors because if it can be singular or plural that has to be explicitly checked all the time before you work with it. If it can be one or more, standardizing on plural seems good -- probably "plural but only one item" doesn't need to be handled as a special case?

bench/lib/parameters.js
ChrisLoer: Hmm yeah I see the challenge. I guess you _could_ wait for the local script to resolve and then look up the `isStyleBench` used there, but that would be a weird timing dependency? Or you could make index.html into something templated and have the watch script populate the parameter accordingly? It's probably not too big a deal if it's hard...

bench/lib/parameters.js
```diff
@@ -1,4 +1,4 @@
-const isStyleBench = process.env.STYLE_BENCHMARK;
+const isStyleBench = getURLParameter('style_bench');
```
ChrisLoer: Nice. 🙂

bench/lib/parameters.js
ryanhamley: This is one of those things that seems like a huge issue when you've been deep in the weeds for a couple weeks. Then you take a step back and realize the solution was obvious. Just using the URL parameter works fine and simplifies the API considerably. D'oh!

jfirebaugh: It looks like there's more divergent branches than shared code in some of these files, so rather than trying to share all the same files between the two benchmarking modes, what about an approach where we use different paths/`index.html`/`benchmarks.js` files?

For example, we could have the version-comparing benchmarks at `/bench/versions/` and the style-comparing benchmarks at `/bench/styles/`.

Where we do want to share code between the modes, we can do that by extracting components or functions and importing them into both.

bench/lib/benchmark.js
ryanhamley: By "standardizing on plural", do you mean making `this.locations` always be an array of Location objects, even if it's just an array with a length of 1? That seems like the most logical way to square this circle in my brain.

ChrisLoer: Yup. :+1:

bench/benchmarks_view.js
```diff
@@ -423,41 +432,103 @@ for (const name in window.mapboxglBenchmarks) {
...
+                            status: 'waiting',
+                            logs: [],
+                            samples: [],
+                            style: {},
+                            summary: {}
+                        });
+                    }
+                });
+
+                promise = promise.then(() => {
```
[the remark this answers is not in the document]
ryanhamley: This seems to be in tension with @jfirebaugh's suggestion above.

> It looks like there's more divergent branches than shared code in some of these files, so rather than trying to share all the same files between the two benchmarking modes, what about an approach where we use different paths/index.html/benchmarks.js files?

The biggest changes are this section and the `register` function in `bench/benchmarks.js`. For the individual benchmark tests, the simplest thing from a code standpoint might be to just standardize on the `style_locations` as input. We don't have to run the tests individually since that takes a long time, but we could still use the `style_locations` as opposed to a simple array of zoom numbers with a hardcoded center all averaged together into a single test. You can see this in action in the `QueryBox` and `QueryPoint` benches because that's exactly what they do in a style test. This might give us better real-world tests in version benchmarking while simplifying the code base some.

As for this section, is it better to attempt to factor out the similar code here or break it apart the way John suggests?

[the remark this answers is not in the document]
ChrisLoer: :thinking: Yeah there is some tension, but maybe not that much? To me it makes sense that specifying the set of benchmarks (e.g. `benchmarks.js`) would have different implementations, but rendering a benchmark (e.g. `benchmarks_view.js`) would have a shared/generalized implementation.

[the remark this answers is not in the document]
ryanhamley: What do you think about standardizing on the locations/tiles in `style_locations.js`? I updated the handful of tests that rely on iterating over a few different zoom levels to use the ones in `style_locations` and it all still works. The code is simpler as well. The `setup` method in `Paint`, for example, went from 25 to 11 lines of code. The tradeoff is that the tests take longer now, which you'd expect from benching 15 maps as opposed to 4 or 5. It's most dramatic in `Layout`.

screen shot 2018-07-31 at 3 25 55 pm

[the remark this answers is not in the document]
ChrisLoer: I like the idea of using a standardized code pathway/definition format for specifying multiple locations. I think we should probably have separate location definitions for style benchmarking vs. the regular benchmarks because:
 - We'll probably want to change the style locations more often
 - In style benchmarking we'll typically care about more locations and be willing to spend more time testing them
 - Changing the location set isn't a big deal when you're comparing multiple styles with the same code version, but since it's baked into the benchmark, changing the location set is a big deal for comparing between code versions (as shown in your screenshot above: it breaks comparisons for Layout whenever different versions have different location sets).

ryanhamley:
Everything still works as before, but I've done some significant refactoring to create separate pathways for style and version tests. It's no longer required to add any additional parameter to the URL which makes switching between the types of tests easier because it's dependent on the npm command you run. The newest commits also refactors the business logic from the view.

bench/lib/benchmark.js
```diff
@@ -11,8 +12,20 @@ export type Measurement = {
...
+    zoom: number,
+    center: LngLatLike
+};
+
 class Benchmark {
-    constructor() {
+    styleURL: string;
+    locations: ?Array<Location>;
+
+    constructor(styleURL: string, locations: ?Array<Location>) {
```
jfirebaugh: Putting this in the base class feels a little over-generalized to me -- not all benchmarks have a style URL. I'd rather have concrete subclasses initialize the properties they use. There'll be a bit of repetition, but it's more explicit and flexible.

bench/version-tests/benchmarks.js
```diff
@@ -1,33 +1,34 @@
...
 mapboxgl.accessToken = accessToken;
 
 window.mapboxglVersions = window.mapboxglVersions || [];
 window.mapboxglBenchmarks = window.mapboxglBenchmarks || {};
 
+const url = styleURL();
 const version = process.env.BENCHMARK_VERSION;
 window.mapboxglVersions.push(version);
 
 function register(Benchmark) {
```
jfirebaugh: Feel free to change the parameter type from a class to an instance, so that different subclasses can accept different constructor parameters. And keep pulling out things that differ between version and style benchmarks.

```
const style = 'mapbox://styles/mapbox/streets-v10';
const center = [-77.032194, 38.912753];
const locations = [4, 8, 11, 13, 15, 17].map(zoom => ({zoom, center, style});
register(new Layout(url));
register(new Paint(locations));
register(new QueryPoint(locations));
```

package.json
```diff
@@ -122,11 +122,15 @@
     "build-benchmarks": "BENCHMARK_VERSION=${BENCHMARK_VERSION:-\"$(git rev-parse --abbrev-ref HEAD) $(git rev-parse --short=7 HEAD)\"} rollup -c bench/rollup_config_benchmarks.js",
     "build-benchmarks-view": "BENCHMARK_VERSION=${BENCHMARK_VERSION:-\"$(git rev-parse --abbrev-ref HEAD) $(git rev-parse --short=7 HEAD)\"} rollup -c bench/rollup_config_benchmarks_view.js",
     "watch-benchmarks": "BENCHMARK_VERSION=${BENCHMARK_VERSION:-\"$(git rev-parse --abbrev-ref HEAD) $(git rev-parse --short=7 HEAD)\"} rollup -c bench/rollup_config_benchmarks.js --watch",
+    "watch-style-benchmarks": "BENCHMARK_VERSION=${BENCHMARK_VERSION:-\"$(git rev-parse --abbrev-ref HEAD) $(git rev-parse --short=7 HEAD)\"} rollup -c bench/rollup_config_style_benchmarks.js --watch",
     "watch-benchmarks-view": "BENCHMARK_VERSION=${BENCHMARK_VERSION:-\"$(git rev-parse --abbrev-ref HEAD) $(git rev-parse --short=7 HEAD)\"} rollup -c bench/rollup_config_benchmarks_view.js --watch",
     "start-server": "st --no-cache -H 0.0.0.0 --port 9966 --index index.html .",
+    "start-bench-server": "st --no-cache -H 0.0.0.0 --port 9966 --index index.versions.html .",
```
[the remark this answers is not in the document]
ryanhamley: I've reduced the scripts to just `start-server` and `start-bench` but switching between `bench/versions/` and `bench/styles/` doesn't rebuild the generated files because Rollup isn't run again. Is there a straightforward way to run Rollup when you change URLs? Or is it ok to have to run `start-bench` again?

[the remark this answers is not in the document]
jfirebaugh: I guess `bench/versions/` and `bench/styles/` should load different bundles, both of which are watched by `start-bench`?

bench/styles/benchmarks.js
```diff
@@ -0,0 +1,118 @@
...
+
+const filter = window.location.hash.substr(1);
+
+function register(Benchmark) {
+    const name = Benchmark.name;
+
+    if (filter && name !== filter)
+        return;
+
+    switch (name) {
```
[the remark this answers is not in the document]
ryanhamley: Doing `locations.forEach(location => { register(Layout...); register(Paint...); });` works but the result is that you when you run all the tests, you get a pattern in the UI of `Layout, Paint, Layout, Paint...`. This could be useful since you can see the Layout and Paint tests for each tile grouped together. Or would we expect that all Layout and all Paint tests are grouped together?

bench/styles/benchmarks.js
```diff
@@ -0,0 +1,118 @@
...
+
+setTimeout(() => {
+    // Ensure the global worker pool is never drained. Browsers have resource limits
+    // on the max number of workers that can be created per page.
+    // We do this async to avoid creating workers before the worker bundle blob
+    // URL has been set up, which happens after this module is executed.
+    getWorkerPool().acquire(-1);
+}, 0);
+
+export default mapboxgl;
```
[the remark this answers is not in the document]
ryanhamley: Removing this causes an error in the console `Cannot set property 'workerUrl' of undefined` that's thrown from `benchmarks_generated.js`, specifically a line `mapboxgl.workerUrl = window.URL.createObjectURL(new Blob([workerBundleString], { type: 'text/javascript' }));`. This doesn't actually seem to cause any issues in running the tests though so I'm not sure what this error is really about.

bench/styles/benchmarks.js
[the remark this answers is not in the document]
jfirebaugh: If you want to keep the existing order, this is fine too:

```js
locations.forEach(location => register(Layout...));
locations.forEach(location => register(Paint...));
```

Issue #6560, opened 2018-04-24
Rendering issue with map position/zoom + tileset w/Features per zoom level


ksummerill:
**mapbox-gl-js version**: 0.44.2

### Steps to Trigger Behavior

 1. Open this map: https://bl.ocks.org/j5kay/a50925604a76c892826e5d3a7db77ddd
 2. Click and drag to move around the map, observe rendering behavior.

### Expected Behavior
Map rendering is smooth and consistent as you move around the map

### Actual Behavior
Rendering is jumpy, appearing and disappearing - https://ookla.d.pr/GoSLAX

jfirebaugh:
Is `mapbox://ookla-ops.coverage-carrier-3221` a partial tilset, not covering the whole world? If so, I suspect this is a variant of #5829 / #4691, where when GL JS encounters a 404 response for a tile outside but adjacent to the boundary of the tileset, which triggers loading the parent tile, which also covers tiles at the natural zoom level that _do_ exist.

j5kay:
@jfirebaugh - Yes, it has data not only within the LngLatBounds specified in the tileset's TileJSON, but also has features on a per-zoom basis between its z=0 to z=10 zoom extents (also specified in the TileJSON).

Not sure if the per-zoom tile data also plays a role per se, but when using `circle`-based layers for rendering data, it seems that multiple zoom levels are merged versus the exclusive rendering for `heatmap` or `extrusion`-based layers. Maybe a separate issue entirely, but I could make an example that illustrates the behavior with `circle` layers, if it's useful.

jfirebaugh:
Thanks for the report! I'm consolidating all the issues related to this bug in #6768.

PR #7051, opened 2018-07-30
prevent overdraw by not retaining parent tiles for sparse tilesets

mollymerp:
This is one possible _interim_ solution for #6768. A long term fix will likely involve changes on the server side that distinguish between `not found` and `no content` responses. 

I'm posting this early without tests to gather feedback on whether we want to move forward with this approach. I used #6995 as the base branch to avoid future conflicts. 

The change boils down to:
 - if an ideal tile is unavailable (hasn't loaded yet or has errored) and if the underlying source is not raster 
   - start ascending the tile pyramid to find a valid parent tile,
   - if the parent tile has _both_ a successfully loaded child, and an errored child tile
     - infer that the tileset is sparse and set `sourceCache.sparse = true`
     - do not retain the parent tile to prevent overdraw on the loaded child tile
     - and for future tile pyramid ascents for other unavailable tiles, only ascend one zoom level above the ideal zoom to prevent overdraw from grandparent tiles or other ancestors

Pros:
- low-lift approach
- removes overdraw on most rendered frames
 
Cons: 
 - overdraw can still flicker on zoom-level changes (similar to what happens in native right now)
 - fast zooming will purge loaded parent tiles before children have loaded causing a flicker on sparse tilesets
 - not yet fully tested, will change existing behavior for some users of sparse tilesets

My main question is whether folks think this is an OK interim solution and is worth polishing and testing more thoroughly for the next release

cc @ryanhamley @jfirebaugh @ryanbaumann @lilykaiser 

overdraw

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -56,6 +56,7 @@ class SourceCache extends Evented {
     _isIdRenderable: (id: number, symbolLayer?: boolean) => boolean;
     used: boolean;
     _state: SourceFeatureState;
+    sparse: boolean;
 
     static maxUnderzooming: number;
     static maxOverzooming: number;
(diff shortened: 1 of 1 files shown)


mollymerp:
another note: this almost completely fixes the issue for heatmaps seen in  #6560 (afaict)  
_before this PR_
before

_after this PR_
aftersmall

cc @ksummerill

mollymerp:
closing this in favor of #6803 in the interest of keeping gl-js and gl-native consistent. #6803 still fixes #6560 and eliminates persistent overdrawing on sparse tilesets.

PR #7060, opened 2018-08-01
Map class cleanup

### LINKED ISSUE #2741: Refactor to simplify Map

mourner:
A cleanup sweep for the `Map` class. See individual commits. Fixes #2741. I don't think there's much more logic left to move downstream unless we break the API.

## Launch Checklist

### DIFF
--- a/test/unit/style/style.test.js
+++ b/test/unit/style/style.test.js
@@ -1782,13 +1782,13 @@ test('Style#queryRenderedFeatures', (t) => {
         style._updateSources(transform);
 
         t.test('returns feature type', (t) => {
-            const results = style.queryRenderedFeatures([{column: 1, row: 1, zoom: 1}], {}, transform);
+            const results = style.queryRenderedFeatures([{x: 0, y: 0}], {}, transform);
             t.equal(results[0].geometry.type, 'Line');
             t.end();
         });
 
         t.test('filters by `layers` option', (t) => {
-            const results = style.queryRenderedFeatures([{column: 1, row: 1, zoom: 1}], {layers: ['land']}, transform);
+            const results = style.queryRenderedFeatures([{x: 0, y: 0}], {layers: ['land']}, transform);
             t.equal(results.length, 2);
             t.end();
         });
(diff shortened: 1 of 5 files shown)


test/unit/style/style.test.js
```diff
@@ -1782,13 +1782,13 @@ test('Style#queryRenderedFeatures', (t) => {
         style._updateSources(transform);
 
         t.test('returns feature type', (t) => {
-            const results = style.queryRenderedFeatures([{column: 1, row: 1, zoom: 1}], {}, transform);
```
jfirebaugh: First argument was supposed to be `{viewport, worldCoordinate}`, so apparently this was just working by accident before.

mourner:
@jfirebaugh yeah, but that interpretation is only mentioned as a small comment at the bottom of the issue. So I think we should make a new issue focused on the API change.

PR #7061, opened 2018-08-01
Fix querySourceFeatures returning bad results on zooms past maxZoom

### LINKED ISSUE #2868: Fix incorrect z coordinate on overzoomed tiles

### LINKED ISSUE #3722: querySourceFeatures returns invalid feature coordinates on high zoom levels
If the map zoom level > 16 the features returned from map.querySourceFeatures have invalid geometry coordinates (negative or otherwise nonsense).

https://jsbin.com/sesafumovi/edit?html,console,output

Can be reproduced on current master branch.

mourner:
Fixes #2868. Fixes #3722. 

~~I also removed the `reparseOverscaled` source option because it doesn't seem useful — we always want to reparse overscaled tiles for correct layout values, and we don't need to worry about performance because parsing overscaled tiles is very fast anyway (there's only a few on the screen and they're small).~~

## Launch Checklist

### DIFF
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -284,12 +284,13 @@ class Tile {
         if (!layer) return;
 
         const filter = featureFilter(params && params.filter);
-        const coord = { z: this.tileID.overscaledZ, x: this.tileID.canonical.x, y: this.tileID.canonical.y };
+        const {z, x, y} = this.tileID.canonical;
+        const coord = {z, x, y};
 
         for (let i = 0; i < layer.length; i++) {
             const feature = layer.feature(i);
             if (filter(new EvaluationParameters(this.tileID.overscaledZ), feature)) {
-                const geojsonFeature = new GeoJSONFeature(feature, coord.z, coord.x, coord.y);
+                const geojsonFeature = new GeoJSONFeature(feature, z, x, y);
                 (geojsonFeature: any).tile = coord;
                 result.push(geojsonFeature);
             }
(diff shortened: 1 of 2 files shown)


src/geo/transform.js
```diff
@@ -248,7 +246,7 @@ class Transform {
             this.pointCoordinate(new Point(this.width, this.height), z),
             this.pointCoordinate(new Point(0, this.height), z)
         ];
-        return tileCover(z, cornerCoords, options.reparseOverscaled ? actualZ : z, this._renderWorldCopies)
```
ansis: Does this work for raster sources past their max zoom level? I think the previous behaviour was that it used the existing tiles from the max zoom level. Does this end up creating new tiles?

mourner:
I also noticed (after the PR) that #4990 inadvertently contains the same fix. Sorry for not noticing earlier @asheemmamoowala!

ralphking:
I'm not sure that this fixes https://github.com/mapbox/mapbox-gl-js/issues/3722

it sets reparseOverscaled to true, which seemingly does the opposite

PR #7078, opened 2018-08-03
Upgrade flow to 0.77

anandthakker:


src/render/draw_background.js
```diff
@@ -42,7 +42,7 @@ function drawBackground(painter: Painter, sourceCache: SourceCache, layer: Backg
         painter.tileExtentPatternVAO.bind(context, program, painter.tileExtentBuffer, []);
     } else {
         program = painter.useProgram('background');
-        gl.uniform4fv(program.uniforms.u_color, [color.r, color.g, color.b, color.a]);
+        gl.uniform4fv(program.uniforms.u_color, ([color.r, color.g, color.b, color.a]: Array<number>));
```
anandthakker: These are needed because Flow doesn't like the ambiguity caused by https://github.com/facebook/flow/pull/5893 (👋 @lbud !)

src/style/style_layer/layer_properties.js.ejs
```diff
@@ -47,4 +47,6 @@ const paint: Properties<PaintProps> = new Properties({
 <% } -%>
 });
 
-export default { paint<% if (layoutProperties.length) { %>, layout<% } %> };
+export default ({ paint<% if (layoutProperties.length) { %>, layout<% } %> }: $Exact<{
+  paint: Properties<PaintProps><% if (layoutProperties.length) { %>, layout: Properties<LayoutProps><% } %>
+}>);
```
anandthakker: Without adding this explicit annotation, Flow was inferring weaker types for these objects from their use in the constructor to `StyleLayer`, as `{layout?: Properties<...>, paint: Properties<...>}`

anandthakker:
Many of the new annotations are (I think) due to this change in [Flow v0.70.0](https://github.com/facebook/flow/blob/master/Changelog.md#0700): 
> New missing annotation errors might be reported. These requirements were missed before because of implementation bugs. We now use the polarity-sensitive type visitor to walk exported types, which reports errors in the expected places. There are several exceptions, which can be exported without annotations: e.g., object and array literals; initialized class properties (instance & static); and this parameters.

src/render/draw_background.js
jfirebaugh: This particular case could be written `gl.uniform4f(program.uniforms.u_color, color.r, color.g, color.b, color.a)` (`4f`, without the `v`).

src/style/style_layer/layer_properties.js.ejs
jfirebaugh: 👍 Add this as a comment.

PR #7079, opened 2018-08-03
Publish flow types

### LINKED ISSUE #6963: Publish Flow type definitions alongside compiled bundle

anandthakker:
Closes #6963

Downstream projects using Flow should now get typing for gl-js. However,
you'll need to include libdefs for some of our dependencies. Our libdefs are
included with the `mapbox-gl` NPM package, so you can get them by adding:

```
[libs]
node_modules/mapbox-gl/flow-typed
```

You'll also need to add an ignore for one of our dependencies that causes
Flow errors:

```
[ignore]
.*/node_modules/@mapbox/jsonlint-lines-primitives/.*
```

### DIFF
--- a/build/mapbox-gl.js.flow
+++ b/build/mapbox-gl.js.flow
@@ -1,0 +1,3 @@
+// @flow
+/* eslint-disable */
+export * from '../src/index.js';
--- a/circle.yml
+++ b/circle.yml
@@ -132,6 +132,7 @@ jobs:
       - run: yarn run build-dev
       - run: yarn run build-css
       - run: yarn run build-style-spec
+      - run: yarn run build-flow-types
       - run: yarn run test-build
       - persist_to_workspace:
           root: .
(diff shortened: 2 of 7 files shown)


anandthakker:
Depends on https://github.com/mapbox/mapbox-gl-js/pull/7078

PR #7081, opened 2018-08-06
Mark WebGL state as dirty on `Map._render()`

davidmanzanares:
After https://github.com/mapbox/mapbox-gl-js/pull/7080 we still saw some WebGL warnings in Chrome's console:
mgl-bug

When debugging it (see picture), we saw that Mapbox GL were doing some WebGL calls before setting the state as dirty.

This PR moves the "setDirty" code from `Painter` to `Map`.

Thanks!


davidmanzanares:
You're welcome!

Issue #5246, opened 2017-09-06
Switch to flat representations of geometry


mourner:
As @jfirebaugh confirmed while investigating #5208, using `Array<Array<Point>>` for internal representations of feature geometries, while convenient, brings a lot of memory/GC overhead.

We could improve performance considerably by switching to a flat array representation of geometry — JS engines are extremely good at working with flat arrays of numbers. E.g. a polygon geometry could look like `[num_rings, num_ring1_points, x, y, x, y, ...]`, inspired by the way it is encoded in Geobuf.

We'll probably have to give up the convenience of using [point-geometry](https://github.com/mapbox/point-geometry) with chained math methods, relying on simple functions instead, but I think it's worth it.

A corresponding task for GeoJSON-VT is tracked at https://github.com/mapbox/geojson-vt/issues/85. 

We might want to do the same in [vector-tile-js](https://github.com/mapbox/vector-tile-js) too (a major breaking change). If so, maybe we should create a small specification for flat GeoJSON geometries to keep the interop among different modules well documented.

cc @jfirebaugh @anandthakker

mourner:
The first step in this will be to get rid of all uses of [point-geometry methods](https://github.com/mapbox/point-geometry/blob/master/index.js) in the codebase, replacing them with plain number calculations (without intermediary point objects). This will be especially painful in `LineBucket` code because of its math-heavy complexity, and the code should become much more verbose, but I'm going to try (without changing any interfaces yet).

mourner:
-> #10501

PR #7085, opened 2018-08-06
Remove point-geometry methods and avoid unnecessary point allocations

mourner:
This PR starts our preparation for #5246. To open up a way to a different geometry storage, first we need to get rid of all uses of [math methods from @mapbox/point-geometry](https://github.com/mapbox/point-geometry/blob/master/index.js) when processing VT geometry, so that we're not tied to all points being `Point` objects from that module. This has additional benefits:

- Those methods made it easy to make unnecessary allocations during vector calculations without realizing it. Operating on raw numbers directly makes all allocations explicit.
- It encourages passing coordinates as numbers in arguments explicitly instead of creating intermediary point objects where possible, further saving on allocations.
- It is less prone to bugs due to immutability — the current approach of using underscored methods for in-place calculations is fragile (e.g. I discovered at least one bug caused by it while refactoring).
- We can optimize calculations like rotation of many points by the same angle (by calculating cos/sin only once).

In my test load of Mapbox Streets on DC z11, doing this in line_bucket.js reduces overall worker-side `Point` allocations from 1579279 to 415265, or by almost 4 times. This doesn't yet have a noticeable effect on the layout benchmark, but I'll try to capture perf more substantially after finishing the PR.

This change also makes the code a bit more verbose and less descriptive, but I think the future trade-off is worth it.

Note that I'm not adding equivalent util methods to replace the Point methods, preferring to inline calculations, to avoid intermediary objects in return values. Once the full elimination pass is complete, I'll see if I can reintroduce some util methods for cleaner code if it's possible to do without the overhead.

cc @ansis @jfirebaugh @ChrisLoer 

### VT-related

- [x] line_bucket
- [x] fill_extrusion_bucket
- [x] line_style_layer
- [x] query_utils
- [x] intersection_tests
- [x] symbol_bucket
- [x] symbol_layout
- [x] clip_line
- [x] get_anchors
- [x] check_max_angle
- [x] projection
- [x] collision_feature
- [x] quads

### Optional

- [ ] transform
- [ ] source_cache
- [ ] camera
- [ ] events
- [ ] popup
- [ ] marker
- [ ] bind_handlers
- [ ] box_zoom
- [ ] drag_pan
- [ ] drag_rotate
- [ ] touch_zoom_rotate
- [ ] smart_wrap
- [ ] eliminate `@mapbox/point-geometry` dependency, replace with internal `Point` type

## Launch Checklist

 - [x] post benchmark scores


mourner:
A simple `console.time` measurement of `WorkerTile` `parse` on a dense DC tile already shows a consistent 10% improvement vs master. Really looking forward to seeing how much we'll win on top of that after switching to flat geometries on the `vector-tile-js` side.

src/data/bucket/fill_extrusion_bucket.js
```diff
@@ -225,8 +212,8 @@ function isBoundaryEdge(p1, p2) {
 }
 
 function isEntirelyOutside(ring) {
-    return ring.every(p => p.x < 0) ||
-        ring.every(p => p.x > EXTENT) ||
-        ring.every(p => p.y < 0) ||
-        ring.every(p => p.y > EXTENT);
+    for (const p of ring) {
```
jfirebaugh: This isn't testing the same thing. For example, a ring with xs all < 0 should return `true`, but this logic will return `false`, because x < 0 ⇒ x <= EXTENT.

mourner: Good catch, fixed.

mourner:
> It is less prone to bugs due to immutability — the current approach of using underscored methods for in-place calculations is fragile (e.g. I discovered at least one bug caused by it while refactoring).

@ansis @jfirebaugh To illustrate this point, here's the nasty bug I found: 

https://github.com/mapbox/mapbox-gl-js/blob/855b850e62196c46f0987ab73600fffddc3e22bc/src/style/style_layer/line_style_layer.js#L134-L136

Notice that on the third line, `aToB` is mutated (because it uses `_` methods). But for the first point of the ring, `aToB` is simply a reference to `zero`, so by the time `zero` is referenced again on the last point, it is no longer zero.

Took quite a while to figure out why a rewritten, flat version was failing query tests.

mourner:
@jfirebaugh I completed the first stage — the map still works if I replace `Point` with `{x, y}` in `vector-tile`. The next stage is to try outputting flat geometries in VT, update the GL JS code accordingly and benchmark, although I'm not sure whether this should be done as a part of this PR rather than a follow-up — I think it can be merged as is.

jfirebaugh:
Can you make a benchmark that shows the improvement so far? It would be useful to have one that's not dominated by earcut anyway.

mourner:
Closing as stale. I'm breaking some of these changes up into different PRs. The first and most important one (for line bucket) is #8303.

PR #7095, opened 2018-08-08
Add migration to expressions into gl-style-migrate

### LINKED ISSUE #6927: Add script to migrate a style from legacy functions and filters to expressions
We have a module for converting stop functions to expressions ([src/style-spec/function/convert.js](https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/function/convert.js)), so what we need is:

 - [ ] A method to convert a legacy filter into an equivalent expression.  This doesn't exist internally right now, because we use a specialized code path to handle legacy filters.  Note that filters converted to expressions might be slightly less performant than their legacy counterparts because of the overhead necessary to represent the exact legacy type semantics in expressions.
- [ ] A function that takes in a style and traverses & converts all of its layer properties and filters.

cc @nickidlugash @mapbox/studio

anandthakker:
Closes #6927 

@mapbox/studio @mapbox/maps-design.

Updates the style-spec's old `gl-style-migrate` script to include conversion of legacy functions and filters to their expression equivalents.  Note that because of some subtle differences in how legacy filters and expressions handle type mismatches, the converted expressions for complex filters (particularly "any" or "none" filters) may end up being overly complicated and under-performant.  For styles that have those, it would be worth manually inspecting and revising the auto-generated expressions.

```
cd src/style-spec
npm run build
bin/gl-style-migrate YOUR_STYLE.json > OUTPUT.json
```


src/style-spec/feature_filter/convert.js
```diff
@@ -0,0 +1,197 @@
...
+ *     ["any", ["all", [">", ["get", "y"], 0], [">", ["get", "z"], 0]], [">", ["get", "x"], 0]]
+ *
+ * But if we tried to evaluate this against, say `{x: 1, y: null, z: 0}`, the
+ * [">", ["get", "y"], 0] would cause an evaluation error, leading to the
+ * entire filter returning false. Legacy filter semantics, though, ask for
+ * [">", "y", 0] to simply return `false` when `y` is of the wrong type,
+ * allowing the subsequent terms of the outer "any" expression to be evaluated
+ * (resulting, in this case, in a `true` value, because x > 0).
+ *
+ * We account for this by inserting a preflight type-checking expression before
```
jfirebaugh: Can you add a few sentences outlining the motivation for this particular approach as opposed to alternatives? I'm assuming it's to minimize the cases where typechecking is needed (only when "any" is used; other cases are fine to let error out [Are they? Will this produce annoying error messages?]).

jfirebaugh:
> What is the official status now of tokens (`{data-field}`) for text-field and icon-image?

Tokens will continue to work, but they're about the same level of deprecated as legacy functions. @anandthakker, want to add conversion for literal string values? (Token strings in legacy function stops are already handled.)

anandthakker:
> We couldn't think of a straightforward way to translate the semantics of default to expressions, and in native, we resorted to supporting it via a special case in the evaluator. Do you have any new ideas on that front?

Ah, good point @jfirebaugh -- I just relied on the existing conversion implementation in https://github.com/mapbox/mapbox-gl-js/blob/855b850e62196c46f0987ab73600fffddc3e22bc/src/style-spec/function/convert.js, but that implementation doesn't fully handle legacy functions' default semantics. (Specifically, the expressions it produces don't fall back to the user-provided default when there's a type mismatch on the input property.)  No new ideas as such--I'd say we could just introduce a similar approach as for filters, including type checks explicitly in the expression when they're needed.

src/style-spec/feature_filter/convert.js
anandthakker: > I'm assuming it's to minimize the cases where typechecking is needed (only when "any" is used; other cases are fine to let error out [Are they? Will this produce annoying error messages?])

Yep, that was the reason.  Other cases are fine to let error out, in the sense that they'll produce the correct behavior.  It's true that they'll produce annoying warnings in that case. But this (and the other issues that have come up re: evaluation error warnings) makes me wonder whether it might make sense to simply disable these warnings (or some subset of them?) by default, only showing them if a `map.expressionDebug` property is set to true...

anandthakker: > simply disable these warnings (or some subset of them?)

Another possibility: we currently only log each distinct evaluation error once, with that uniqueness scoped to each expression. If we expanded the uniqueness scope to the whole map, that would dramatically reduce the number of these warnings, while still giving developers a clue if they're trying to figure out why their style isn't doing what they want.

jfirebaugh:
The tricky case for defaults turns out to be arrays. For e.g. `{"type": "identity", "property": "p", "default": [-1]}` we want to convert to something like `["array", "number", ["get", "p"], ["literal", [-1]]`. This poses two problems:
* Unlike other assertion operators, `"array"` doesn't currently accept fallbacks.
* It's not obvious how to add them due to how it is overloaded. With three arguments, the second argument is an expected length, but in this case we don't want to make assertions about the length. `text-font` doesn't have a fixed length, and legacy functions don't use the `default` value for an incorrect length on other array-valued properties. Supporting a `null` placeholder, i.e. `["array", "number", null, ["get", "p"],  ["literal", [-1]]`, is an option, but I'm not fond of it.

The other missing piece is `"interpolate-hsla"`, `"interpolate-lab"`, and `"interpolate-hcl"` expression operators, so that `colorSpace` can be converted.

jfirebaugh:
> Supporting a `null` placeholder, i.e. `["array", "number", null, ["get", "p"],  ["literal", [-1]]`, is an option...

I ended up just doing this.

This is now ready for review.

jfirebaugh:
Yeah, that's kind of my hope. If I was redesigning type assertions I'd do it like `["assert", {"type": "array", "element-type": "number"}, input, fallbacks...]` or something like that, but it's too late to change now.

jfirebaugh:
FYI @mapbox/studio @mapbox/maps-design (who got de-tagged ☝️in the description due to Anand's departure) -- this merged. Please try it out and let us know if you notice any issues with the migration.

PR #7098, opened 2018-08-08
Use absolute asset URL in hover-styles example

ryanhamley:
## Launch Checklist

I think this was missed in https://github.com/mapbox/mapbox-gl-js/pull/6834 but we should use an absolute URL for assets on all the example pages so that they can be copied over to JSBin or wherever


andrewharvey:
> I think this was missed in #6834 but we should use an absolute URL for assets on all the example pages so that they can be copied over to JSBin or wherever

Whoops, thanks for catching this.

shouldn't it target `mb-pages`?

ryanhamley:
@andrewharvey Yep. I was just waiting until the release was finished to merge this.

ryanhamley:
Oops, I borked this PR by rebasing against master instead of mb-pages so I created a new branch and PR https://github.com/mapbox/mapbox-gl-js/pull/7103 I'm closing this one

PR #7106, opened 2018-08-10
Correct feature state id type to number

bfrengley:
Currently, string feature ids are not propagated through the transformation from GeoJSON to vector tiles (#2716, #6960) and are [not supported by vector tiles at all](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto#L32). However, `setFeatureState` / `getFeatureState` indicate that they require string ids.

This pull request updates `setFeatureState` / `getFeatureState` to require unsigned numeric ids.

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -803,7 +803,7 @@ class Style extends Evented {
             return;
         }
         if (feature.id == null || feature.id < 0) {
-            this.fire(new ErrorEvent(new Error(`The feature id parameter must be provided and positive.`)));
+            this.fire(new ErrorEvent(new Error(`The feature id parameter must be provided and non-negative.`)));
             return;
         }
 
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -787,7 +787,7 @@ class Style extends Evented {
         return this.getLayer(layer).getPaintProperty(name);
     }
 
-    setFeatureState(feature: { source: string; sourceLayer?: string; id: string; }, state: Object) {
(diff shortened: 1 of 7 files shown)


src/style/style.js
```diff
@@ -802,15 +802,15 @@ class Style extends Evented {
             this.fire(new ErrorEvent(new Error(`The sourceLayer parameter must be provided for vector source types.`)));
             return;
         }
-        if (feature.id == null || feature.id === "") {
-            this.fire(new ErrorEvent(new Error(`The feature id parameter must be provided.`)));
+        if (feature.id == null || feature.id < 0) {
+            this.fire(new ErrorEvent(new Error(`The feature id parameter must be provided and non-negative.`)));
```
asheemmamoowala: I think `Map#setFeatureState` should still accept string types that can be coerced to integers. This would allow what is otherwise a valid `id` for a GeoJSON feature to be input to the API without explicit conversion. 

Based on https://github.com/mapbox/mapbox-gl-js/pull/4581:

https://github.com/mapbox/mapbox-gl-js/blob/3b0a1a9c3fceaefe1b0cd6f2d36cf92138621715/src/source/geojson_wrapper.js#L38-L46

asheemmamoowala: @bfrengley that seems like a fix worth putting in as well.

mourner:
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

@bfrengley can you please rebase on top of master for clean merging?

PR #7124, opened 2018-08-14
Reduce worker payload

mourner:
Reduces the size of the JSON (excluding transferred arrays) that gets sent from the worker to the main thread, reducing lags during tile loading (in theory). Partially addresses #7011.

For a dense z11 Mapbox Streets v10 tile in DC, the payload is reduced by ~44%. For the expressions-based Basic style, the reduction is 39%.

See individual commits. The first one has the biggest impact (~28%) — it converts serialized form from `{name: 'Object', properties: {foo: 'bar'}}` to `{foo: 'bar}` for simple objects and `{$name: 'Foo', foo: 'bar'}` for both registered classes and custom-serialized ones.

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/data/program_configuration.js
+++ b/src/data/program_configuration.js
@@ -70,7 +70,7 @@ function packColor(color: Color): [number, number] {
  */
 
 interface Binder<T> {
-    statistics: { max: number };
+    maxValue: number;
     uniformName: string;
 
     populatePaintArray(length: number, feature: Feature): void;
@@ -89,7 +89,7 @@ interface Binder<T> {
 class ConstantBinder<T> implements Binder<T> {
     value: T;
     name: string;
-    statistics: { max: number };
+    maxValue: number;
     type: string;
     uniformName: string;
 
@@ -98,7 +98,7 @@ class ConstantBinder<T> implements Binder<T> {
         this.name = name;
         this.uniformName = `u_${this.name}`;
         this.type = type;
-        this.statistics = { max: -Infinity };
+        this.maxValue = -Infinity;
     }
 
     defines() {
@@ -127,7 +127,7 @@ class SourceExpressionBinder<T> implements Binder<T> {
     name: string;
     uniformName: string;
     type: string;
-    statistics: { max: number };
+    maxValue: number;
 
     paintVertexArray: StructArray;
     paintVertexAttributes: Array<StructArrayMember>;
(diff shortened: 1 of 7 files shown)


src/data/program_configuration.js
```diff
@@ -386,7 +394,7 @@ export default class ProgramConfiguration {
             const useIntegerZoom = value.property.useIntegerZoom;
 
             if (value.value.kind === 'constant') {
-                self.binders[property] = new ConstantBinder(value.value, name, type);
+                self.binders[property] = new ConstantBinder(value.value.value, name, type);
```
mourner: This one's weird — I guess this was a buggy code in an unused code path, since this doesn't break any tests.

mourner:
image

image

src/data/program_configuration.js
ChrisLoer: And flow isn't helping us here because `ConstantBinder` is deducing the type from that first argument? 😕

src/util/web_worker_transfer.js
```diff
@@ -68,16 +69,18 @@ export function register<T: any>(name: string, klass: Class<T>, options: Registe
 
 register('Object', Object);
 
-Grid.serialize = function serializeGrid(grid: Grid, transferables?: Array<Transferable>): Serialized {
-    const ab = grid.toArrayBuffer();
+type SerializedGrid = { buffer: ArrayBuffer };
+
+Grid.serialize = function serializeGrid(grid: Grid, transferables?: Array<Transferable>): SerializedGrid {
```
ChrisLoer: I'm missing something... what was the JSON savings in changing from serializing as an `ArrayBuffer` to serializing as a `SerializedGrid`?

mourner: If you serialize just the `ArrayBuffer` which is transferred, any custom properties on it (in this case `$name`) don't survive the transfer. Before the PR we implicitly transferred in a wrapper (`{name: 'Grid', properties: buffer}`), and now we explicitly transfer as `{$name: 'Grid', buffer}`. I'm now throwing an error is `serialize` in case the serialized object is fully transferred just so that we don't trip up on this in future.

src/data/program_configuration.js
mourner: Yeah, I guess it's because of flow templating (`ConstantBinder<T>`) — I'm not fully sure how it works (@jfirebaugh may know better) but looks like if there's a template or an `any` in Flow, you can basically do anything in the code the touches those places without any type safety.

src/util/web_worker_transfer.js
ChrisLoer: Ah got it, thanks.

PR #7125, opened 2018-08-14
fix icon-size for small data-driven values

### LINKED ISSUE #7017: symbol layout 'icon-size' < 0.1 is treated as 0
**mapbox-gl-js version**: 0.47.0

**browser**: chrome, firefox

### Steps to Trigger Behavior

 1. create symbol layer with 'icon-size' as expression (not constant)
 2. make sure the expression evaluates to 0.09

### Link to Demonstration

https://jsfiddle.net/og5tbcLn/31/

In the demonstration, there is two features with size 0.1 and 0.09 calculated from property.
The icon with size 0.1 is displayed while the 0.09 is not.

If i set 'icon-size': 0.09 (not expression) it works as expected.
It also works as expected if the expression does not contain ['get', ...] expression

### Expected Behavior
Display the symbol with 0.09 scale

### Actual Behavior
Symbol is displayed with 0 scale

### LINKED ISSUE #7066: Stops vs Literal value: Icon size is different for same value.
Symbol layer with
``` 
   "icon-size": 0.27,
```
``` 
   "icon-size": {
      type: 'interval',
      default: iconSize,
      property: "__resize_by__",
      stops: [[0, iconSize]]
    },
```
are rendering images at different size.

[~~jsfiddle~~](http://jsfiddle.net/bewithjonam/d0ust2nr/6/) [jsfiddle](https://jsfiddle.net/bewithjonam/Lf1prz26/12/) explaining the issue. In the fiddle I render a icon
at size 0.27 and then after 2 seconds I set the same size using stops. 

### Expected Behavior
It should render with the same size

### Actual Behavior
When using stops, it is rendering smaller than when given a literal value

ansis:
fix #7017, fix #7066

Fixes both issues by scaling icon-size and text-size by 256 instead of 10. It can now represent values from 0 - 256 with precision of 1/256. I think this is enough and saves the bytes compared to using floats.

## Launch Checklist

 - [ ] post benchmark scores


ChrisLoer:
This changes makes a lot of sense to me and your implementation looks good. The one thing I'm worried about is text/icon-size overflow. I've seen icon-size overflow at 6554, and now it would overflow at 256. Admittedly, 256 already seems like a big value. In the case I saw a user hitting overflow, they were trying to exponentially size an icon so it matched map scale as you zoomed in. Probably something like a fill layer with a pattern would be a better tool for that particular use case, but when the overflow happens the behavior is really confusing/hard-to-debug. Can we do anything to improve that? Like maybe generate intelligible console errors at expression evaluation time?

ansis:
93a40f9 adds runtime warnings for icon-size and text-size values that are too big

PR #7128, opened 2018-08-14
Convert offsetHeight/Width to clientHeight/Width

### LINKED ISSUE #6848: consider replacing offset{Height/Width} with client{Height/Width} for map canvas sizing
We currently use the map `container`'s `offsetHeight` and `offsetWidth` to size the map canvas element, which leads to odd behavior when using a border, including `Map#getCenter`, `Map#getBounds`, etc. returning incorrect values for what is visible to the user because some of the map canvas will be hidden by the container.

https://github.com/mapbox/mapbox-gl-js/blob/bf88cd735b06eacc607c188a2508c9912140768d/src/ui/map.js#L1474-L1477

image

image

If we used `clientHeight`/`clientWidth` to size the canvas, we would eliminate this behavior, but I'm not sure what the original reasoning was behind the choice to use `offset*`

This would qualify as a breaking change. If we decide not to make this change, we could specify in the documentation that a map's `container` cannot have a border. 

### Expected Behavior
with `client{Height/Width}` sizing, canvas remains within the main content area of the container element

expected

### Actual Behavior

with current behavior, canvas overflows the main content area and is partially hidden due to a border being used on the container element

actual

cc @mourner @jfirebaugh @kkaefer

ryanhamley:
## Launch Checklist

    - Converts `offset{Height/Width}` to `client{Height/Width}`
    - Fixes #6848  

`master with offset{Height/Width}`
> screen shot 2018-08-14 at 1 16 23 pm

`canvas-hw with client{Height/Width}`
> screen shot 2018-08-14 at 1 17 08 pm

### DIFF
--- a/src/style-spec/types.js
+++ b/src/style-spec/types.js
@@ -408,3 +408,4 @@ export type LayerSpecification =
     | RasterLayerSpecification
     | HillshadeLayerSpecification
     | BackgroundLayerSpecification;
+
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1414,8 +1414,8 @@ class Map extends Camera {
         let height = 0;
 
         if (this._container) {
-            width = this._container.offsetWidth || 400;
-            height = this._container.offsetHeight || 300;
+            width = this._container.clientWidth || 400;
+            height = this._container.clientHeight || 300;
         }
 
         return [width, height];
--- a/test/suite_implementation.js
+++ b/test/suite_implementation.js
@@ -30,8 +30,8 @@ module.exports = function(style, options, _callback) { // eslint-disable-line im
     window.devicePixelRatio = options.pixelRatio;
 
     const container = window.document.createElement('div');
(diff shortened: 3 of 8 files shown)


src/ui/popup.js
```diff
@@ -290,8 +290,8 @@ export default class Popup extends Evented {
         const offset = normalizeOffset(this.options.offset);
 
         if (!anchor) {
-            const width = this._container.offsetWidth;
-            const height = this._container.offsetHeight;
+            const width = this._container.clientWidth;
+            const height = this._container.clientHeight;
```
mourner: This is measuring the popup's container rather than the map one, and we might want to leave `offset` sizes here, since for positioning the popup, it's important to take the whole popup into account (including borders/etc) rather than just its internal area.

ryanhamley: I reverted popups and the attribution and logo controls to use offsetWidth/Height

PR #7129, opened 2018-08-14
don't clear state on reload, just on setData

mollymerp:
in #6849 we added a fresh state on source `reload` but reload is also called when a style property or layout property is changed that requires relayout, so we were inadvertently expunging the state whenever that happened. originally reported in https://github.com/mapbox/mapbox-gl-js/pull/6289#issuecomment-411186302

## Launch Checklist


src/source/source_cache.js
```diff
@@ -75,6 +75,7 @@ class SourceCache extends Evented {
             // for sources with mutable data, this event fires when the underlying data
             // to a source is changed. (i.e. GeoJSONSource#setData and ImageSource#serCoordinates)
             if (this._sourceLoaded && !this._paused && e.dataType === "source" && e.sourceDataType === 'content') {
+                if (this._source.type === "geojson") this.resetState();
```
asheemmamoowala: nit: {}

PR #7130, opened 2018-08-14
espresso cherry-pick: don't clear state on reload, just on setData

mollymerp:
ref #7129


mollymerp:
superceded by #7141  #7142

PR #7132, opened 2018-08-15
Use a special data structure for transferring feature id maps to the main thread

### LINKED ISSUE #7110: Performance regression: feature-state _idMap building is expensive
Since version 0.46, when feature-state was added, time-to-first-render (TTFR) drastically increased.

Not building _idMap in populatePaintArrays in ProgramConfiguration, by getting rid of the if-block there, gets you back the old performance.

Could you make it so _idMap doesn't get built unless feature-state is actually used? Or maybe delay it until after first render?

See also: https://github.com/mapbox/mapbox-gl-js/pull/7039#issuecomment-410701578

BTW, #7015 helps reduce TTFR, but slowdown is still significant compared to version 0.45.

mourner:
Fixes #7110 (not fully but almost there); a part of addressing #7011. An alternative to #7015.

Instead of a struct array that gets deserialized into a JS object on the main thread (which is expensive and causes stutter), this PR introduces a special data structure for storing feature id to position mappings which consists of two typed arrays — ids and positions. Both get sorted by id before transferring, which allows using binary search on the main thread for id lookup (avoiding building a JS object completely).

This brings TTFR for the difficult #7110 case to nearly pre-feature-state levels (10-20% slower instead of 4x slower), while also theoretically reducing stutter (due to elimination of big non-transferable id maps which are very expensive to clone when sending to the main thread). @ryanbaumann's [feature state choropleth](https://bl.ocks.org/ryanbaumann/733ba99c5ca1d9d15259081b395e4b00/414e974f854c20c263edddc6c9cc934d0956a9ac) also speeds a little (20% faster to fully load).

Note that I used `Uint32Array` for buffer offsets instead Uint16 (like in #7015) because the latter overflows (e.g. on Ryan's example).

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/data/feature_position_map.js
+++ b/src/data/feature_position_map.js
@@ -1,0 +1,112 @@
+// @flow
+
+import { register } from '../util/web_worker_transfer';
+
+type SerializedFeaturePositionMap = {
+    ids: Float64Array;
+    indices: Uint32Array;
+    offsets: Uint16Array;
+};
+
+type FeaturePosition = {
+    index: number;
+    start: number;
+    end: number;
+};
+
+// A transferable data structure that maps feature ids to their indices and buffer offsets
+export default class FeaturePositionMap {
+    ids: Array<number>;
+    indices: Array<number>;
+    offsets: Array<number>;
+    _bufferLen: number;
+
+    constructor() {
+        this.ids = [];
+        this.indices = [];
+        this.offsets = [];
+        this._bufferLen = 0;
+    }
+
+    add(id: number, index: number, newLength: number) {
+        this.ids.push(id);
+        this.indices.push(index);
+        this.offsets.push(this._bufferLen, newLength);
+        this._bufferLen = newLength;
+    }
+
+    getPositions(id: number): Array<FeaturePosition> {
+        // binary search for the first occurrence of id in this.ids
+        let i = 0;
+        let j = this.ids.length - 1;
+        while (i < j) {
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Can we avoid even generating an id map for layers that don't use any `feature-state` expressions?

asheemmamoowala:
@jfirebaugh that is possible, but would require re-building the paint arrays, or building the map some other way if `setPaintProperty` is used with a `feature-state` expression.

mourner:
@jfirebaugh we should definitely explore this, but independently of this PR (which will be a benefit in any case I think).

jfirebaugh:
Did none of our existing benchmarks detect the performance regression in #7110? What would it take to have a benchmark that can alert us to regressions in TTFR?

mourner:
> Did none of our existing benchmarks detect the performance regression in #7110? What would it take to have a benchmark that can alert us to regressions in TTFR?

@jfirebaugh it might help to add a worker roundtrip benchmark that parses a few heavy tiles and then just sends them back and forth between the main thread and a worker (to specifically target `postMessage` performance and transferring). Let's do it in a follow-up PR.

jfirebaugh:
@mourner We should have a benchmark that shows this change is a performance benefit though. (And unit tests.)

mourner:
Results for the new WorkerTransfer benchmark:

image

Issue #6889, opened 2018-06-29
setFeatureState and the need for removeFeatureState


cammanderson:
Hi Guys, 

Working with the new setFeatureState, and I noticed that there isn't a way to remove the state supplied. You must always provide a value to merge it, and can't reset to the original values. 

Use cases are varied, and not sure of the overall impact, but I am updating a react project to offer the function, but I can't 'reset' the state back to 'pre' application of the setFeatureState. Ideally, I would be able to remove the feature state completely, and then assumingly it would source from the original value of the underlying source data. If not, you'd have to track the original property value of the object to work around it. Not sure if we need to add a removeFeatureState() ?

Cheers
Cam

**mapbox-gl-js version**:
0.46.0

**browser**:
N/A

### Steps to Trigger Behavior

 1. setFeatureState({ source: 'composite', layer: 'water', id: 0 }, { value: true })
 2. setFeatureState({ source: 'composite', layer: 'water', id: 0 }, null)

### Expected Behavior
Remove feature state entirely

### Actual Behavior
Because underlying it is using an extend, the null || {} values have no effect.

featureState: { value: true }

asheemmamoowala:
@cammanderson this should be possible by getting the current state with `Map#getFeatureState()` and setting all values to `null` or `undefined` before calling `Map#setFeatureState()`.

peterqliu:
@cammanderson @asheemmamoowala from a design perspective, should the `removeFeatureState` reset the entire state object, or allow users to reset specific keys within the object?

peterqliu:
Making some progress for a key-level `removeFeatureKey` over at https://github.com/mapbox/mapbox-gl-js/commit/37409713d205d6be97712ca4d3a0f0f4a863423b

asheemmamoowala:
Closed by #7761

PR #7143, opened 2018-08-16
Espresso: Revert feature state reset

asheemmamoowala:
#7043 added a reset of a source's feature state every time it was reloaded. This was meant to target only `GeoJSONSource#setData` but inadvertently also reset the feature data when a source went through a relayout. This was originally reported in [#62889(comment)(https://github.com/mapbox/mapbox-gl-js/pull/6289#issuecomment-411186302) and a fix was put into `master` via #7129.

For the espresso release, revert this behavior entirely.  A more inclusive fix is being researched for `master` along the lines of #6889.

cc @jfirebaugh @mollymerp

### DIFF
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -217,9 +217,6 @@ class SourceCache extends Evented {
         }
 
         this._cache.reset();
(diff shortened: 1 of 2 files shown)


asheemmamoowala:
Beaten to the 👊 by #7142

PR #7151, opened 2018-08-17
Adds example demonstrating the format expression 

malwoodsantoro:
This example demonstrates the new `format` expression by showing country labels in both English and local language.

### DIFF
--- a/docs/pages/example/display-and-style-bilingual-labels.html
+++ b/docs/pages/example/display-and-style-bilingual-labels.html
@@ -7,10 +7,10 @@ var map = new mapboxgl.Map({
     zoom: 4 // starting zoom
 });
 
-map.on('load', function(){
+map.on('load', function() {
     var layers = ['country-label-lg', 'country-label-md', 'country-label-sm'];
 
-    layers.forEach(function(layer){
+    layers.forEach(function(layer) {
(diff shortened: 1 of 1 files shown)


docs/pages/example/display-and-style-bilingual-labels.html
```diff
@@ -0,0 +1,20 @@
+<div id='map'></div>
+<script>
+var map = new mapboxgl.Map({
+    container: 'map', // container id
+    style: 'mapbox://styles/mapbox/streets-v9', // stylesheet location
```
ChrisLoer: nit: should probably use streets-v10 as the basemap in new examples?

docs/pages/example/display-and-style-bilingual-labels.html
```diff
@@ -0,0 +1,20 @@
...
+    center: [9.49, 49.01], // starting position [lng, lat]
+    zoom: 4 // starting zoom
+});
+
+map.on('load', function(){
+    map.setLayoutProperty('country-label-lg', 'text-field', ['format',
```
ChrisLoer: The example will probably have a more consistent appearance if we set the same (or similar) styling for all three of the country layers? I think it's -lg, -md, and -sm, right?

PR #7164, opened 2018-08-21
Edit title of format expression example

malwoodsantoro:
Per chat with @1ec5 and @nickidlugash, this is a more fitting and descriptive title for this example.


1ec5:
I personally find “Display and style rich text labels” to be descriptive because “rich text” specifically refers to being able to piece together differently-formatted runs of text. All our other labels are based on plain text, formatted uniformly. The search bar on the left also searches through descriptions, so we can reserve some details for the description. Some terms people might search for are “format”, “rich text”, “style”, and “markup”.

PR #7173, opened 2018-08-23
Auto set .cn EVENTS_URL endpoint by api url

macro-shen:
**Auto set cn EVENTS_URL endpoint by api url**

This PR made a configuration change on our backend that as long as the customer points to `api.mapbox.cn` then their turnstile events will automatically be set to `events.mapbox.cn`, without needing to ask them change this by themselves. This logic is same with that in iOS sdk.

### DIFF
--- a/src/util/config.js
+++ b/src/util/config.js
@@ -6,7 +6,6 @@ const EventsUrlOptions = {
 };
 
 type Config = {|
-  _API_URL: string,
   API_URL: string,
   EVENTS_URL: string,
   REQUIRE_ACCESS_TOKEN: boolean,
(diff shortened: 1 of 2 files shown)


macro-shen:
@jfirebaugh - Thank you for your advice, I will do it. The reasons of configuration of `EVENTS_URL` are (1) we don't need ask developers/clients change this by themselves.(2) There has been a same logic in iOS sdk.

chriswu42:
> Please add a unit test for this as well.

@springmeyer for visibility.

springmeyer:
> @springmeyer for visibility.

Thanks @chriswu42. @macro-shen as you work to create unit tests please see https://github.com/mapbox/mapbox-gl-js/blob/master/CONTRIBUTING.md#writing--running-tests

Issue #6722, opened 2018-05-23
Pinch & Zoom is off center on touch devices (v0.45.0)


dave11mj:
**mapbox-gl-js version**: `v0.45.0`

**browser**:
- Chrome (Windows 10) with a touch monitor (Dell S2240T)
- Safari (iPhone 7)
- Safari (iPad)
- Any other touch enabled device (specially large screen displays)

### Steps to Trigger Behavior

 1. Place your fingers around the intended target to initiate zoom
 2. Expand your fingers in a pinch & zooming gesture
 3. Observe target moves away from the center of your fingers

### Link to Demonstration

![v0-45-0-demo](https://user-images.githubusercontent.com/6619371/40448684-39133d2a-5ea4-11e8-8dfb-bd7552f4b065.gif)

[Demo Github Page Using v0.45.0](https://dave11mj.github.io/github-issues-demos/mapbox/zooming-out-of-center-v0-45-0.html)
[Demo Source](https://github.com/dave11mj/github-issues-demos/blob/master/mapbox/zooming-out-of-center-v0-45-0.html)

Note that in the demo above when you try to pinch and zoom on the "Capitol Hill Fountain"
it goes off center instead of remaining between the fingers.

![v0-44-2-demo](https://user-images.githubusercontent.com/6619371/40447887-f4623fc0-5ea1-11e8-9f26-689dfc5eff95.gif)

[Demo Github Page Using v0.42.2](https://dave11mj.github.io/github-issues-demos/mapbox/zooming-out-of-center-v0-42-2.html)
[Demo Source](https://github.com/dave11mj/github-issues-demos/blob/master/mapbox/zooming-out-of-center-v0-42-2.html)

Note that in the demo above when you try to pinch and zoom on the "Capitol Hill Fountain"
it stays between fingers. However, you are unable to pan with more than one finger.

### Expected Behavior

Target should remain centered between the user's fingers without removing the ability to pan using more than one finger.

### Actual Behavior

Target moves off center as the user zooms into it.

andest01:
I am able to reproduce this error on my setup as well.
- Chrome Version 66.0.3359.181 (Official Build) (64-bit)
- (Windows 10) 
- Dell P2418HT touch device

jliebrand:
This makes using pinch zoom pretty crappy on mobile. Any update on a potential fix here?

jliebrand:
Friendly ping. Any plans for a fix?

pakastin:
[...]

PR #7191, opened 2018-08-24
Keep multitouch center at start center location when pinching (#6722)

pakastin:
This is a simple solution for #6722. Just save start center location and keep multicouch center point at that location while pinching. Works with and without rotation without issues.


pakastin:
The fix is in use here and works flawlessly: https://beta.ilmailukartta.fi/
(https://beta.ilmailukartta.fi/js/mapbox-gl.js)

PR #7208, opened 2018-08-27
Exponential interpolation example

malwoodsantoro:
@mzdraper and I are working on filling in some of the gaps in GL JS examples this quarter. We only have one example utilizing exponential interpolation and thought we should add another. The idea is that we will eventually be able to link to these examples next to each expression in the style spec.


docs/pages/example/create-color-ramp-for-building-layer.html
```diff
@@ -0,0 +1,52 @@
+<style>
+    #zoom {
```
mollymerp: Let's center the button and style it like those in https:www.mapbox.com/mapbox-gl-js/example/language-switch/ and https://www.mapbox.com/mapbox-gl-js/example/flyto/ for consistency.

docs/pages/example/create-color-ramp-for-building-layer.js
```diff
@@ -0,0 +1,10 @@
+/*---
+title: Create color and opacity ramp for building layer
```
mollymerp: Maybe use a more descriptive title like "Change building color based on zoom level"

docs/pages/example/create-color-ramp-for-building-layer.js
```diff
@@ -0,0 +1,10 @@
...
+description: Uses exponential interpolation to ease-in the building layer and smoothly fade from one color to the next.
```
mollymerp: Let's explicitly call out that this is an example using expressions and link to those docs here.

PR #7210, opened 2018-08-28
Adds elm-mapbox integration

gampleman:
Adds a link to elm-mapbox, which is a library for the Elm programming languages that wraps a map and also creates a type safe wrapper around the style spec language.


ryanhamley:
This should go to our `mb-pages` branch rather than `master` so it appears immediately in the docs. Can you rebase your branch against `mb-pages` before I merge this?

gampleman:
@ryanhamley Done.

Issue #7076, opened 2018-08-02
Style spec: add option to disable y-position sort for symbol features


ChrisLoer:
We've automatically sorted symbols based on their projected y-position since https://github.com/mapbox/mapbox-gl-js/issues/470 -- the reasoning was roughly (1) they're often not likely to be in any meaningful order in the source data, and (2) sorting them makes them look more visually coherent.

https://github.com/mapbox/mapbox-gl-js/issues/4361 tracks adding a general purpose feature sorting functionality - something that would allow you to generate arbitrary sort keys via expressions.

The status quo is that non-symbol features are drawn in the order they appear in the underlying data. It can be cumbersome to maintain your data in such a way that you get the draw order you want, but for some use cases it works out pretty well. We have some users that would like to be able to use the same source-based ordering for symbols.

I think implementing this is just adding a style spec property something like `symbol-sort-by-y` that defaults to `true`. If it's `false`, then we just skip the y-sorting step.

/cc @lilykaiser @mapbox/maps-design @mapbox/studio

nickidlugash:
> they're often not likely to be in any meaningful order in the source data

Just want to flag that all data in Mapbox Streets Source (including for symbols) has meaningful sort order. I am pretty sure that removing the y sorting would be beneficial to our core styles (but would definitely want to visually confirm this).

> I think implementing this is just adding a style spec property something like symbol-sort-by-y that defaults to true. If it's false, then we just skip the y-sorting step.

Since sorting by y-position is only beneficial for symbol layers that allow overlap, should we consider setting the default to `auto`, where it is `true` if either `*-allow-overlap` or `*-ignore-placement` is `true`, and `false` otherwise?

ChrisLoer:
> Since sorting by y-position is only beneficial for symbol layers that allow overlap, should we consider setting the default to auto, where it is true if either *-allow-overlap or *-ignore-placement is true, and false otherwise?

This is pretty much already the case: sorting only happens for symbols that can overlap with other symbols in the same layer, because otherwise sorting wouldn't make any difference. Because the core styles rarely use overlapping symbols, I think this is mostly irrelevant for them.

[...]

PR #7219, opened 2018-08-29
Add symbol-z-order symbol layout property to style spec

ryanhamley:
## Launch Checklist

    - Adds symbol-sort-by-y symbol layout property to style spec
    - closes https://github.com/mapbox/mapbox-gl-js/issues/7076
    - Native port: https://github.com/mapbox/mapbox-gl-native/pull/12783


ryanhamley:
See https://github.com/mapbox/mapbox-gl-js/issues/7076 for background on this PR.

I can think of two alternatives to adding a new style spec property.

1) We can make this a runtime setting.
    - Downside: I don't think it makes much sense semantically and organizationally since the sorting only applies to symbols, not all layers on a map. Disabling this sorting is closely linked to the existing style spec symbol layout properties `*-allow-overlap` and `*-ignore-placement` so it's logical to include this as another style spec property.

2) We could push forward with adding a mechanism such as some expression syntax to allow for generalized manual sorting of all feature types as discussed in https://github.com/mapbox/mapbox-gl-js/issues/4361.
    - Downside: This is significantly more work and possibly overkill for this particular use case.

The downside of this PR's approach obviously is a new property on the style-spec which is something we have to be prepared to maintain for awhile. If/when #4361 is implemented, does this feature become obsolete? I feel like symbol sorting by y is a specialized case that users may want a simple way to disable even if they plan on doing further sorting. And of course this concern assumes we eventually implement general sorting.

Two further thoughts/questions:

1) Is `symbol-sort-by-y` an appropriate name or would we prefer something more descriptive like `sort-overlapping-symbols` (which doesn't specify *how* the symbols are sorted).

2) Is the description of the property in `v8.json` accurate and informative enough for users? And is it true that non-skeumorphic icons won't benefit from symbol sorting and are therefore good candidates for `symbol-sort-by-y: false`?

Tagging Carto and Studio into this discussion: @nickidlugash @samanpwbb

samanpwbb:
Overall approach looks good to me. I think it's fine to make this a new spec property since it's a  straightforward boolean. if/when we take on #4361 we can deprecate or de-emphasize the property.

> Is symbol-sort-by-y an appropriate name or would we prefer something more descriptive like sort-overlapping-symbols (which doesn't specify how the symbols are sorted).

What about `sort-relative-to-viewport`? I find `sort-by-y` a little unclear - I needed to read the description to understand what it meant.

ryanhamley:
Renamed `symbol-sort-by-y` to more descriptive `symbol-sort-relative-to-viewport`

ryanhamley:
screen shot 2018-08-31 at 4 38 12 pm

jfirebaugh:
Let's make this `symbol-z-order`, an enum-valued property with possible values `"viewport-y"` and `"source"`. That gives us the flexibility to extend the property with other options. (I can imagine a request for `"map-y"` for example.)

samanpwbb:
> Let's make this symbol-z-order, an enum-valued property with possible values "viewport-y" and "source". That gives us the flexibility to extend the property with other options. (I can imagine a request for "map-y" for example.)

I like this option 👍

ryanhamley:
Yep. Makes sense. I'll change this and the native PR over to `symbol-z-order`.

ryanhamley:
Ah yeah. I think most front end developers will understand it pretty intuitively as being analogous to the CSS z-order property though.

PR #7220, opened 2018-08-29
Add example that utilizes jumpTo function

malwoodsantoro:
We don't have an example that demonstrates this function, so I thought I would create one.

### DIFF
--- a/docs/pages/example/jump-to.html
+++ b/docs/pages/example/jump-to.html
@@ -83,10 +83,10 @@ var map = new mapboxgl.Map({
 
 map.on('load', function() {
 
-    cities.features.forEach(function(city, index){
-        setTimeout(function(){
+    cities.features.forEach(function(city, index) {
+        setTimeout(function() {
             map.jumpTo({center: city.geometry.coordinates});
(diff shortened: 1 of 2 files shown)


docs/pages/example/jump-to.html
```diff
@@ -0,0 +1,95 @@
...
+    container: 'map',
+    style: 'mapbox://styles/mapbox/streets-v10',
+    center: [100.507, 13.745],
+    zoom: 9
+});
+
+map.on('load', function() {
+
+    var i = -1;
+    (function jumpToCity () {
```
mollymerp: I think this [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) might be a little over-complicated for the purpose of the examples page – begginers to JS might not know what's going on here. 
You could simplify this so that its just a simple loop that only visits each location once, or even simpler just have a button that will jump trigger a jump to a random location or a random location in your list like the `flyTo` example

malwoodsantoro:
Thanks @mollymerp :) Great advice. I went with the approach of looping through the cities once. Let me know if it's good to merge.

Issue #7222, opened 2018-08-30
Fix #6828 break the flyTo end position in some cases


benoitbzl:
**mapbox-gl-js version**: 0.48

**browser**: Chrome

### Steps to Trigger Behavior

At the end of animation, when final scale does not produce the correct zoom value, the zoom value is forced to the desired one (purpose of the fix #6828). But the scale value is not modified accordingly, so that the center computation goes wrong at 

https://github.com/mapbox/mapbox-gl-js/blob/8dd624d4e09f68d4a8739f1d140a983cae7e23e0/src/ui/camera.js#L970

The scale value does not match the one of the transform object, so the unproject() function produces wrong values.

mourner:
Would you be open to submitting a PR for this?

benoitbzl:
I am not sure yet how to compute the scale value based on the zoom value, but I will give it a try this afternoon.

benoitbzl:
I submitted PR #7223.

PR #7223, opened 2018-08-30
Fix flyTo when the final zoom value is not the requested one (#7222)

benoitbzl:
This complete the fix for #6828 when the final zoom value is different from the requested zoom.

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -957,8 +957,8 @@ class Camera extends Evented {
         this._ease((k) => {
             // s: The distance traveled along the flight path, measured in ρ-screenfuls.
             const s = k * S;
-            const scale = k === 1 ? tr.zoomScale(zoom - startZoom) : 1 / w(s);
-            tr.zoom = startZoom + tr.scaleZoom(scale);
+            const scale = 1 / w(s);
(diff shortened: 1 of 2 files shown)


src/ui/camera.js
```diff
@@ -957,8 +957,8 @@ class Camera extends Evented {
         this._ease((k) => {
             // s: The distance traveled along the flight path, measured in ρ-screenfuls.
             const s = k * S;
-            const scale = 1 / w(s);
-            tr.zoom = k === 1 ? zoom : startZoom + tr.scaleZoom(scale);
+            const scale = k === 1 ? tr.zoomScale(zoom - startZoom) : 1 / w(s);
+            tr.zoom = startZoom + tr.scaleZoom(scale);
```
mourner: Let's keep the old `k === 1` condition here because zoomScale+scaleZoom may introduce loss of floating point precision.

benoitbzl: Good point. I will update the PR with your recommendation.

benoitbzl:
I found example code in the tests to simulate the animation for flyTo, as I think the problem I saw requires the computation of a fly path over several points. I will give it a try.

PR #7232, opened 2018-09-04
update platform support tables for *-pattern

mollymerp:


src/style-spec/reference/v8.json
```diff
@@ -3501,7 +3501,10 @@
           "macos": "0.1.0"
         },
         "data-driven styling": {
-          "js": "0.46.0"
+          "js": "0.49.0",
+          "android": "6.5.0",
+          "macos": "0.11.0",
+          "ios": "4.4.0"
```
1ec5: Typically, native releases aren’t added to the compatibility tables until the releases are published. These releases are in development but aren’t quite there yet.

mollymerp: This is just updating master – it won't be published on mapbox.com til the `f` final release date

1ec5: Ah, I missed that. Making this change on master ahead of the release is a great idea. That’ll save me from having to follow up as part of the macOS map SDK release process.

PR #7234, opened 2018-09-05
Update yarn install instructions for Linux in CONTRIBUTING.md

anderco:
Since Ubuntu version 18.04 (and possibly as early as 17.04), the package
cmdtest has a "Provides: yarn" entry, so it gets installed when one runs
`sudo apt-get install yarn`. It also doesn't seen Ubuntu ever had a
package for yarn, so add instructions to install it from the right
sources.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,11 +31,17 @@ yarn install
 
 ### Linux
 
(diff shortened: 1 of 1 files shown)


anderco:
> I wonder if this is simpler?

It is, but with a slightly different end result since yarn will get installed in the user home. The rest of the instructions seen to assume Debian or Ubuntu so I took the relevant instructions for Yarn's install page.

In the end, this is a matter of preference, so perhaps we should just let contributors figure it out by themselves by pointing to the Yarn install page?

anderco:
I replaced the inline instructions with the simpler ones, but added a note that packages are also available so people can choose.

PR #7252, opened 2018-09-07
Enable face culling for fill extrusion layers

anderco:
Follow up to #7178. See https://github.com/mapbox/mapbox-gl-js/pull/7178#issuecomment-419453791

This implements face culling as a new argument to Program::draw so that it can be enabled on layer by layer basis and also enables it for fill extrusion layers. Because of that, only the winding order of that type of layer is changed. If there would be a decision to enforce winding order for other layers that could be done incrementally.


ansis:
Closing in favour of https://github.com/mapbox/mapbox-gl-js/pull/7178

PR #7253, opened 2018-09-07
Follow-up fixes for atlas packing

### LINKED ISSUE #7241: gl.texImage2D "Invalid argument" in IE 11
Many examples started failing in IE 11 after #7171 with an "Invalid argument" error here:

https://github.com/mapbox/mapbox-gl-js/blob/93fef1aea9b242c02203f71f86dd9e838adf1a84/src/render/texture.js#L73

Guessing here, but maybe IE 11 does not like `0` for width and height, which didn't happen before #7171 but now it is?

cc @mourner

mourner:
I don't have a way to test on IE11, but let's check if this fixes #7241.


mourner:
Thanks a lot for confirming!

PR #7261, opened 2018-09-11
["to-array", <item type>, <empty array>] should work for any item type

jfirebaugh:
Previously it was a type error because the runtime type of an empty array is inferred as `array<value, 0>`.


jfirebaugh:
Yeah, there are several changes that need porting; I'll do them all at once and roll into https://github.com/mapbox/mapbox-gl-native/pull/12864.

PR #7262, opened 2018-09-11
Update Benchmarks readme

nickidlugash:
Update Benchmarks readme to include access token in command for running style benchmarks.

(@ryanhamley: we also discussed the other day changing the commands to `yarn start-bench`. I'm not sure why `yarn start` wasn't working for me previously, but it does work now so I don't have a preference here).

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -21,7 +21,7 @@ By default, the benchmark page will compare the local branch against `master` an
 Start the benchmark server
 
 ```bash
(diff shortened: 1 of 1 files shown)


ryanhamley:
Looks good to me. `yarn start-bench` and `yarn start` should both work just fine for running these tests. `yarn start` was broken after merging this to master because I forgot to update the script after deprecating a couple of sub-scripts, but it was fixed last week.

PR #7273, opened 2018-09-12
fix missing repeats of CanvasSource when it crosses the antimeridian

### LINKED ISSUE #7271: Canvas source doesn't render in world repeats at edges
> The example above shows our requirement of syncing a canvas source (represented by a green border with blue background) such that it is always centered in the map container and is equal to the pixel width of one world. The bug can be seen in the rightmost world repeat, where the canvas is not visible. If you pan the map east past the antimeridian, the canvases render correctly. This also happens when zooming. For example, zooming into the left instance of India from the initial view will cause the canvas to disappear once the map bounds no longer contains the antimeridian.

**browser**: Chrome

### Steps to Trigger Behavior

 1. Add canvas source that covers large part of the world and crosses antimeridian
 2. Zoom out
 3. See missing copy at right edge of map

### Link to Demonstration

https://codepen.io/anon/pen/pOVPgE

### Expected Behavior

Renders canvas on right side.

### Actual Behavior

Doesn't render canvas.

ansis:
This fixes #7271 by including an extra copy of image sources and canvas sources on either side of the world. In almost all cases this extra copy is necessary, but it shouldn't add any overhead because rendering two offscreen triangles should be negligible.

It might be slightly nicer to have an approach that actually calculates how many copies we need to render but I think this would be significantly more complex and probably slower.

## Launch Checklist

 - [ ] post benchmark scores


ryanhamley:
This is probably a major edge case, but I noticed this behavior while just randomly playing around with the map.

screen shot 2018-09-12 at 10 27 22 am

screen shot 2018-09-12 at 10 43 47 am


In fact, it happens on `master` too so it's an existing issue that is not tied to this change. Should this be addressed with this change or maybe just ticket it and investigate it later?

ansis:
@ryanhamley good catch, a54e774 adds a fix for that as well

src/geo/transform.js
```diff
@@ -200,13 +200,21 @@ class Transform {
...
-        const ur = this.pointCoordinate(new Point(this.width, 0), 0);
-        const w0 = Math.floor(ul.column);
-        const w1 = Math.floor(ur.column);
         const result = [new UnwrappedTileID(0, tileID)];
         if (this._renderWorldCopies) {
-            for (let w = w0; w <= w1; w++) {
+            const utl = this.pointCoordinate(new Point(0, 0), 0);
+            const utr = this.pointCoordinate(new Point(this.width, 0), 0);
+            const ubl = this.pointCoordinate(new Point(this.width, this.height), 0);
+            const ubr = this.pointCoordinate(new Point(0, this.height), 0);
```
ryanhamley: Can you explain how this fixes the rotation issue, @ansis?

ansis: Using all four corners is necessary since the leftmost/rightmost points could be any of the corners. The following two lines with the Math.min and Math.max handle the case when the map is upside down

PR #7280, opened 2018-09-13
Auto-coerce top-level results and concat arguments to strings

### LINKED ISSUE #6190: Automatic string coercion for "concat" and text-field
**mapbox-gl-js version**: `0.44`

It is common to want to use numbers as strings, especially in the `text-field` input. One use-case is to display number values on contour lines. This doesn't work:

```json
["get", "ele"]
```

The proper expression is as follows:
```json
["to-string", ["get", "ele"]]
```

For all users, this is pretty confusing, and doesn't seem necessary:

- JS developers are used to numbers being treated like strings in cases like this, so I expect this will be a common stumbling block when implementing any expression on a string field.
- In Studio, we can work around the current design by automatically wrapping all return values in the UI in `to-string`, but it adds complexity and just feels unnecessary.

jfirebaugh:
Fixes #6190, implementing auto-coercion for `"concat"` and string-valued properties. The latter change affects the behavior of coalesce as well, for instance `"text-field": ["coalesce", ["get", "a"], ["get", "b"]]` will now never produce a type error.

## Launch Checklist


src/style-spec/function/convert.js
```diff
@@ -36,7 +36,9 @@ function convertIdentityFunction(parameters, propertySpec): Array<mixed> {
     const get = ['get', parameters.property];
 
     if (parameters.default === undefined) {
-        return get;
+        // By default, expressions for string-valued properties get coerced. To preserve
+        // legacy function semantics, insert an explicit assertion instead.
+        return propertySpec.type === 'string' ? ['string', get] : get;
```
samanpwbb: Should the `to-string` expression be dropped here: https://github.com/mapbox/mapbox-gl-js/pull/7280/files#diff-4b96c63e8fcfc96468a92c7a73e53b5eR229

src/style-spec/expression/index.js
```diff
@@ -116,7 +116,11 @@ export function isExpression(expression: mixed) {
  */
 export function createExpression(expression: mixed, propertySpec: StylePropertySpecification): Result<StyleExpression, Array<ParsingError>> {
     const parser = new ParsingContext(definitions, [], getExpectedType(propertySpec));
-    const parsed = parser.parse(expression);
+
+    // For string-valued properties, coerce to string at the top level rather than asserting.
+    const parsed = parser.parse(expression, undefined, undefined, undefined,
+        propertySpec.type === 'string' ? {typeAnnotation: 'coerce'} : undefined);
+
```
samanpwbb: 🎉

src/style-spec/function/convert.js
jfirebaugh: 👍 , updated.

src/style-spec/expression/index.js
```diff
@@ -347,22 +351,23 @@ function findZoomCurve(expression: Expression): Step | Interpolate | ParsingErro
...
-function getExpectedType(spec: StylePropertySpecification): Type | null {
+function getExpectedType(spec: StylePropertySpecification): Type {
     const types = {
         color: ColorType,
         string: StringType,
         number: NumberType,
         enum: StringType,
-        boolean: BooleanType
+        boolean: BooleanType,
+        formatted: FormattedType
```
ChrisLoer: I think with this change, every place in the code that checks `text instanceof Formatted` has a dead branch. I can go clean those up in a later PR. There may also be some performance impact here because of the extra bookkeeping the "formatted" path uses, although I'd guess it's pretty minimal.

ChrisLoer:
No obvious change in the Layout benchmark, which should exercise the extra string->formatted coercion (although I've been getting pretty variable results between runs):

screenshot 2018-09-17 10 18 13

ChrisLoer:
I went ahead and made the changes to strip out the `instanceof Formatted` dead code, and as part of testing that realized that the implicit coercion strategy John used in https://github.com/mapbox/mapbox-gl-js/pull/7280/commits/85fce0183a9d2423c17c1289b1bc6f485db45982#diff-7341a1d97068bd8c5a153f89bc9e4b1aL117 wouldn't work on this relatively simple case:

```
[ "coalesce", [ "get", "name_en" ], [ "get", "name" ] ]
```

The reason was that the coalesce introduced a `typeAnnotation` of `omit`, triggering the assert at:

```
 assert(!options.typeAnnotation);
 parsed = new FormatExpression([{text: parsed, scale: null, font: null}]);
```

I'm not sure of the right answer here -- it seems like allowing `omit` might make sense in this case. But I think we have to support any type of annotation, because for instance if someone defines a `text-field` to be `['number', 0]` that should generation an expression parsing error, not an assertion failure.

I tried going back to the approach of using `Coercion`, but added explicit format-supporting code to the `Coercion` so that it wouldn't try to serialize as `to-formatted`.

@mollymerp since I've started modifying this PR, can I ask you to step in as reviewer?

I'm going to try to add some test cases for the different "type annotation" cases.

test/integration/expression-tests/format/implicit-assert/test.json
```diff
@@ -0,0 +1,14 @@
+{
+  "expression": ["number", ["get", "p"]],
+  "propertySpec": {
+    "type": "formatted"
+  },
+  "expected": {
+    "compiled": {
+        "result": "error",
+        "errors": [
+          {"key": "", "error": "Expected formatted but found number instead."}
```
mollymerp: This seems like a rather unhelpful error message, considering that most `"text-field"` values won't be actually using the `"format"` expression. Should we special case this type checking error for `formatted` or punt on this for a later error message improvement project?

src/style-spec/expression/parsing_context.js
```diff
@@ -61,20 +60,29 @@ class ParsingContext {
...
     }
 
-    _parse(expr: mixed, options: {omitTypeAnnotations?: boolean}): ?Expression {
-
+    _parse(expr: mixed, options: {typeAnnotation?: 'assert' | 'coerce' | 'omit'}): ?Expression {
         if (expr === null || typeof expr === 'string' || typeof expr === 'boolean' || typeof expr === 'number') {
             expr = ['literal', expr];
         }
 
+        function annotate(parsed, type, typeAnnotation: 'assert' | 'coerce' | 'omit') {
```
mollymerp: I find `typeAnnotation` to be a very confusing name for this modifier (`typeModifier` 🤔 ) but we don't have to deal with that in this PR 😶

test/integration/expression-tests/format/implicit-assert/test.json
ChrisLoer: 🤔 I think defer? The way I kind of think about it is: it's confusing because people expect to be able to return a 'string', and have it implicitly-coerce to 'formatted' without them having to even know 'formatted' exists. So you'd kind of like it to say something like `Expected formatted or string but found number instead.` -- but that's still actually confusing because (1) things other than 'formatted' or 'string' can still successfully coerce to 'formatted', (2) even 'number' can coerce to 'formatted' - the problem is that the assertion disables the implicit coercion.

PR #7281, opened 2018-09-14
[docs] Update SDK compatibility versions for line-gradient

captainbarbosa:
Updates the SDK compatibility table for the new `line-gradient` property introduced in https://github.com/mapbox/mapbox-gl-native/pull/12575/.

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -4082,7 +4082,10 @@
       ],
       "sdk-support": {
         "basic functionality": {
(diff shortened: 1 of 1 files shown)


captainbarbosa:
@1ec5 I think I may have editing the wrong file (`src/style-spec/reference/v8.json` vs. your edits on `docs/pages/style-spec.js` in https://github.com/mapbox/mapbox-gl-js/pull/7282/. Can you confirm that I should be editing `style-spec.js` instead?

Issue #7247, opened 2018-09-06
3d fill-extrusion-heights on iOS Safari are flat when height is over 65,536 meters


levelsio:
- versionGlJs: 0.47.0 

Platform: Safari iOS (and also Android 3rd-party reports)

Example Link: https://nomadlist.com/3d

### Expected Behavior
Fill extrusion height is set from a features geoJSON dataset: 
'fill-extrusion-height': {type: 'identity', property: 'height'}, 

### Actual Behavior
fill-extrusion-heights in 3d ignore height on Safari iOS regardless of height value and all show as flat.

I think it's because I am showing fill-extrusion-heights on a world level, so they're quite big. And it's making some kind of integer overflow (65,536?) on iOS. I've seen Safari iOS handle integers differently before.

On any desktop browser like Safari MacOS and Chrome MacOS they show up properly as large bars.

How it looks on iOS:
![photo_2018-09-07_03-52-13](https://user-images.githubusercontent.com/7150848/45184549-6f0cb900-b251-11e8-9377-7c5c09933035.jpg)

How it should look (in Chrome desktop responsive mode): <img width="557" alt="screen shot 2018-09-07 at 03 51 41" src="https://user-images.githubusercontent.com/7150848/45184558-746a0380-b251-11e8-842d-ccdb6cd55d38.png">

And how it looks on Chrome desktop:
<img width="1055" alt="screen shot 2018-09-07 at 03 51 50" src="https://user-images.githubusercontent.com/7150848/45184557-73d16d00-b251-11e8-8e61-82cc53f00bb8.png">

levelsio:
It seems to be reported before:
https://github.com/mapbox/mapbox-gl-js/issues/3954

"yes, fill-extrusion-height uses a Uint16 type, so the maximum possible value is 65535. At the moment we don't have plans to use bigger numbers as the increase in memory usage would be substantial for what seems to be a pretty limited use case — from what I've seen so far most use cases don't come near 65535 meters." by @lbud on 2017-01-13

It's stated the limit is removed:
https://github.com/mapbox/mapbox-gl-js/pull/5338

Is it though?

jfirebaugh:
Duplicate of #7104.

levelsio:
> Duplicate of #7104.

Sorry but this is wrong. It's a different issue. The examples in #7104 work fine on my iPhone 8 Plus in Safari iOS, but the fill-extrusion-height in my example works on no mobile devices at all, only desktop. 

[...]

Issue #7104, opened 2018-08-10
Extrusions don't work on mobile safari on iPhone 10 running iOS 11.4.1


brsbl:
**mapbox-gl-js version**: v0.47.0

**browser**: mobile safari on iPhone 10

### Steps to Trigger Behavior

 1. Open a map with extrusions on mobile safari on an iPhone 10

### Link to Demonstration

Here's an example url: https://api.mapbox.com/styles/v1/saman/cjkn8e4ki85r62qmw46cqgy2k.html?fresh=true&title=true&access_token=pk.eyJ1Ijoic2FtYW4iLCJhIjoiS1ptdnd0VSJ9.19qza-F_vXkgpnh80oZJww#0.0/0.000000/0.000000/0

And a picture of the linked style working on iPhone 8 and not iPhone 10 (both running iOS 11.4.1):
![image from ios](https://user-images.githubusercontent.com/19296229/43932487-45eb077a-9bf9-11e8-9332-7562b892666c.jpg)

Also found it worked on iPhone 7.

Style sheet for this style:
```
{
    "version": 8,
    "name": "Blank",
    "center": [
        1.4622770665675944,
        42.27278859987004
    ],
    "zoom": 5.14961062357403,
    "bearing": -20.000000000000114,
    "pitch": 36.50000000000002,
    "sources": {
        "composite": {
            "url": "mapbox://mapbox.mapbox-streets-v7",
            "type": "vector"
        }
    },
    "sprite": "mapbox://sprites/saman/cjkn8e4ki85r62qmw46cqgy2k",
    "glyphs": "mapbox://fonts/saman/{fontstack}/{range}.pbf",
    "layers": [
        {
            "id": "background",
            "type": "background",
            "paint": {
                "background-color": "hsl(237, 76%, 64%)"
            }
        },
        {
            "id": "country-label",
            "type": "fill-extrusion",
            "source": "composite",
            "source-layer": "water",
            "layout": {},
            "paint": {
                "fill-extrusion-height": 100000,
                "fill-extrusion-color": "hsl(0, 0%, 100%)"
            }
        }
    ],
    "created": "2018-08-10T00:06:23.155Z",
    "id": "cjkn8e4ki85r62qmw46cqgy2k",
    "modified": "2018-08-10T00:09:34.783Z",
    "visibility": "private"
}
```

### Expected Behavior

Extrusions to appear

### Actual Behavior

Extrusions don't appear, just the background color of the map

[...]

Issue #2096, opened 2016-02-09
Improve shader attribute precision on mobile devices


lucaswoj:
There are two outstanding bugs related to poorly calibrated shader attribute precision on mobile devices (#2055, #2060 and #2020)

@kristfal kindly provided a PR (https://github.com/mapbox/mapbox-gl-js/pull/2091) with one potential solution to the problem. 

We should all take a step back and come up with our ideal blue-sky solution to the problem. Should we

 - keep adding `highp` annotations to the shader source when bugs are reported 
 - design a "smarter" (than [this](https://github.com/mapbox/mapbox-gl-js/blob/master/js/render/gl_util.js#L23-L26)) shader attribute precision rewriting algorithm
 - other?

cc @kristfal @ansis @kkaefer @jfirebaugh

jfirebaugh:
Previously: #1852, #750.

I think we should use `highp` universally and unconditionally in vertex shaders. There appears to be little or no downside to doing so.

For fragment shaders, [it's more complicated](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices). Do we know of any cases where our fragment shaders produce incorrect results on mobile with `mediump`?

ansis:
> I think we should use highp universally and unconditionally in vertex shaders. There appears to be little or no downside to doing so.

> For fragment shaders, it's more complicated. Do we know of any cases where our fragment shaders produce incorrect results on mobile with mediump?

This sounds right to me. I don't think there are any cases we need high precision in fragment shaders.

lucaswoj:
Naively switching most shader attributes to `highp` has led to many rendering errors on desktop devices, even though they claim `highp` support

```js
        var isHighpSupported = (
            context.getShaderPrecisionFormat &&
            context.getShaderPrecisionFormat(type, context.HIGH_FLOAT).precision > 0
        );
```

Do we want to continue restricting `highp` use to mobile devices?

jfirebaugh:
> Naively switching most shader attributes to highp has led to many rendering errors on desktop devices

Can you elaborate on this?

lucaswoj:
In my prototype branch, #2117, circles and lines fail to render in the browser and in headless testing. I'm not yet sure what the underlying cause is.

ansis:
done https://github.com/mapbox/mapbox-gl-js/issues/2096

PR #7292, opened 2018-09-18
Increase fill_extrusion base/height precision.

ChrisLoer:
Fixes issue #7247: heights over 65,536 meters don't render on some devices (tested against iOS Safari). We were using the `lowp` precision qualifier for the `fill-extrusion-base` and `fill-extrusion-height` attributes. GLSL allows implementations to use _really_ low precision (and small range) for lowp floats, although most implementations don't go down to the limits.

See discussion in issue #2096: as a rule of thumb we want to standardize on using `highp` precision qualifiers in our shaders unless there's a specific reason we know (1) low precision won't break anything we depend on, and (2) low precision is likely to be a useful performance optimization.

I tested this against a simple "landcover" extrusion layer set to 70,000 meters on iOS 11 Safari running on an iPhone 8.

**Before:**
image from ios 1

**After:**
image from ios

## Launch Checklist

 - [ ] ~~write tests for all new functionality~~ (Failing test depends on a hardware configuration we can't do on CI)
 - [x] post benchmark scores

/cc @mollymerp @mourner @mzdraper

### DIFF
--- a/src/shaders/fill_extrusion.fragment.glsl
+++ b/src/shaders/fill_extrusion.fragment.glsl
@@ -1,13 +1,6 @@
 varying vec4 v_color;
-#pragma mapbox: define lowp float base
-#pragma mapbox: define lowp float height
-#pragma mapbox: define highp vec4 color
 
 void main() {
-    #pragma mapbox: initialize lowp float base
-    #pragma mapbox: initialize lowp float height
-    #pragma mapbox: initialize highp vec4 color
-
     gl_FragColor = v_color;
 
 #ifdef OVERDRAW_INSPECTOR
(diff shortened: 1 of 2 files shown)


ChrisLoer:
Running on Desktop Chrome:
screenshot 2018-09-18 13 02 56

Running on iPhone 8:
image from ios 2

ryanhamley:
Does this also close https://github.com/mapbox/mapbox-gl-js/issues/7104 or is that a separate problem?

ChrisLoer:
Yeah looks like the same problem -- it's an extrusion layer set to 100,000 meters... and I just now see that John actually had a branch that did _almost_ the same thing as what I did here, except he went to mediump instead of all the way to highp.

Issue #7295, opened 2018-09-19
Doc disambiguation for mouseover event


Fractaliste:
Into [current documentation](https://www.mapbox.com/mapbox-gl-js/api/#map.event:mouseover) the mouseover event appears twice with different explanation:
![doc_error](https://user-images.githubusercontent.com/5181419/45760868-5c1fbe80-bc2b-11e8-8ce0-ba373cc2a05c.png)

The second one refers to mouseenter event which seems to be much more restricted than mouseover one's

ryanhamley:
Thanks for the heads up! Closed by https://github.com/mapbox/mapbox-gl-js/pull/7299

PR #7299, opened 2018-09-19
Remove 'mouseover' as synonym for 'mouseenter'

ryanhamley:
## Launch Checklist

    - removes duplicative `mouseover` event that says its a synonym for `mouseenter`

Closes https://github.com/mapbox/mapbox-gl-js/issues/7295

This looks like an oversight to me. The synonym was added in August 2017 https://github.com/mapbox/mapbox-gl-js/commit/8a68233f4371a6ff3b65c0a1822586b15d8688bd while the actual `mouseover` event was added in January 2018 https://github.com/mapbox/mapbox-gl-js/commit/35138df78e1200109940fac68ad588ff2a58d9e0 I'm guessing the synonym should have been removed at that time.

### DIFF
--- a/src/ui/events.js
+++ b/src/ui/events.js
@@ -26,7 +26,6 @@ export class MapMouseEvent extends Event {
         | 'mouseover'
         | 'mouseenter'
         | 'mouseleave'
-        | 'mouseover'
         | 'mouseout'
         | 'contextmenu';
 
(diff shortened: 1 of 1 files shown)


ChrisLoer:
Although it shouldn't make any difference, we should probably clean up the duplicate entry here too, right?

https://github.com/mapbox/mapbox-gl-js/blob/a1233cb33e3b55cde90b52b908cb4d8f4b6864e6/src/ui/events.js#L21-L31

ryanhamley:
Oh wait...should this go to `mb-pages` rather than `master`?

Issue #7300, opened 2018-09-19
"let" expression strips expected type during parsing


ChrisLoer:
This showed up in a bug report as a somewhat cryptic "Type string is not interpolatable" error for a style property of this form:

```
"fill-color": ["let", 
   "someNumber",
   500,
   [
          "interpolate",
          ["linear"],
          ["var", "someNumber"],
          274,
          "#edf8e9",
          1551,
          "#006d2c"
    ]
```

The same interpolate expression without using a "let/var" worked just as expected. It turns out the reason was that the color strings (eg "#edf8e9" were getting implicitly coerced to the "color" type because that's the expected type for the `fill-color` property. Inside the let expression though, the implicit coercion didn't happen.

I think this is just a mistake here:

https://github.com/mapbox/mapbox-gl-js/blob/a1233cb33e3b55cde90b52b908cb4d8f4b6864e6/src/style-spec/expression/definitions/let.js#L52

That `undefined` should actually be `context.expectedType`, in order to correctly set the let expression's result type. It looks like we're already doing that in native:

https://github.com/mapbox/mapbox-gl-native/blob/a991eec597541c32682946b541e22ccad0f98dd1/src/mbgl/style/expression/let.cpp#L61

cc @malwoodsantoro @mollymerp @ryanhamley

PR #7301, opened 2018-09-19
Preserve expected type when parsing 'let' expressions

ryanhamley:
## Launch Checklist

Fixes https://github.com/mapbox/mapbox-gl-js/issues/7300

### DIFF
--- a/src/style-spec/expression/definitions/let.js
+++ b/src/style-spec/expression/definitions/let.js
@@ -49,7 +49,7 @@ class Let implements Expression {
             bindings.push([name, value]);
         }
 
-        const result = context.parse(args[args.length - 1], args.length - 1, undefined, bindings);
+        const result = context.parse(args[args.length - 1], args.length - 1, context.expectedType, bindings);
         if (!result) return null;
 
         return new Let(bindings, result);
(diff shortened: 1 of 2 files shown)


ryanhamley:
Confirmed that the error goes away with this change

screen shot 2018-09-19 at 3 26 28 pm

cc: @ChrisLoer @malwoodsantoro

ChrisLoer:
Awesome! This is a really good candidate for an expression regression test -- you can start from one of the existing "let" tests, and use add a "propertySpec" entry to the test.json to specify the expected type.

ryanhamley:
@ChrisLoer I added a test that's passing. Does that look right to you?

ChrisLoer:
Yeah, that looks right to me! Did you double-check that the test also fails if you revert your change?

You could probably make the test a little more focused/easier to read by getting rid of the nested interpolate? Just one interpolate expression ought to be enough to exercise the functionality. I've also started trying to annotate some of our render/query/expression tests with a "description" field that describes what's going on in the test (in this case something like "'let' should preserve expected type for child expressions. If it didn't, this test case would fail because the expected "color" type would not get passed into the interpolate and the literal strings wouldn't automatically coerce to the color type").

If you think this is ready for review, I'm ready to 🍏.

ryanhamley:
I simplified the test, added a description and confirmed that the test fails if I revert the change in the `let` definition.

PR #7309, opened 2018-09-20
premultiply canvas images when updating texture

### LINKED ISSUE #7302: Canvas source not respecting alpha values set on canvas element
**mapbox-gl-js version**:
post-v0.49.0 (this bug has not hit a release yet but is reproducible on `master`)

**browser**:
any

### Steps to Trigger Behavior
 1. Draw/render image on a canvas element with some alpha transparency
 2. Add canvas element to a canvas source

### Link to Demonstration

https://codepen.io/rsh412/pen/ZMVNJq?editors=1010#0

Note that this is an example of the canvas source working as expected. I can't get a buggy example to work in JSBin or Codepen because the bug isn't in a release. But at least this shows a testable setup.

### Expected Behavior
Canvas source should render alpha transparency set by canvas element

### Actual Behavior
Canvas source renders alpha channel with opacity of 1

v0.49.0 (correct alpha opacity)
screen shot 2018-09-19 at 5 20 13 pm

master (incorrect alpha opacity)
screen shot 2018-09-19 at 5 20 43 pm

`git bisect` points to something in https://github.com/mapbox/mapbox-gl-js/pull/7039 as the culprit. cc @ansis

ansis:
Fix #7302 by premultiplying canvas images when updating their textures.

## Launch Checklist

 - [] document any changes to public APIs
 - [] post benchmark scores
 - [] tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec changes


ansis:
I added some render tests that fake canvases using images so that I could add a regression test for this. The approach taken to fake the canvas isn't *great* but I couldn't think of a better approach.

ansis:
> Can you just explain to me real quick why the texture.bind and gl.texSubImage2D calls aren't necessary anymore?

The call to `texture.update(...)` contains a gl.texSubImage2D. The other bind calls... I'm not sure that we needed them before. I'm guessing the might be left over from previous refactoring

PR #7310, opened 2018-09-20
Make 'formatted' serialization slightly less verbose.

ChrisLoer:
Omit "font-scale" and "text-font" from serialization if they're unused.

This is a pretty minor change, but I think it brings us closer to the behavior of the rest of our serializations. I noticed it because on the native side the expression type checking would fail on a `null` `font-scale` because it expects a number type if the field is defined.

Technically this changes our public interface, so tagging in studio and design, but I wouldn't expect this to be a breaking change for most consumers of the serialization (maybe I'm not thinking broadly enough?).

## Launch Checklist

 - [x] tagged @mapbox/studio and/or @mapbox/maps-design 
 if this PR includes style spec changes

/cc @mollymerp @ryanhamley

### DIFF
--- a/src/style-spec/expression/types/formatted.js
+++ b/src/style-spec/expression/types/formatted.js
@@ -31,10 +31,14 @@ export default class Formatted {
         const serialized = ["format"];
         for (const section of this.sections) {
             serialized.push(section.text);
-            const fontStack = section.fontStack ?
-                ["literal", section.fontStack.split(',')] :
-                null;
-            serialized.push({ "text-font": fontStack, "font-scale": section.scale });
+            const options = {};
+            if (section.fontStack) {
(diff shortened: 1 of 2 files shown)


ryanhamley:
It looks ok to me and I generally favor simplicity in code as much as possible. I don't know much about the serialization of expressions and when they might be consumed by a user, but this change seems like it brings the actual behavior of the code more in line with the documentation:

>  If set, the text-font argument overrides the font specified by the root layout properties. If set, the font-scale argument specifies a scaling factor relative to the text-size specified in the root layout properties.

"If set" implies that `text-font` and `font-scale` are optional and I wouldn't expect to have to pass `null` for an optional variable to work.

Issue #7312, opened 2018-09-20
getImage will fail silently if an SVG is requested


ryanhamley:
**mapbox-gl-js version**:
current

**browser**:
any

### Steps to Trigger Behavior

 1. Add image source with a url pointing to an SVG file

### Expected Behavior
Nothing appears on the screen and no warnings or errors are thrown

### Actual Behavior
An error with information on what's wrong should be thrown

PR #7313, opened 2018-09-20
Add error message when trying to use SVG in image source

ryanhamley:
## Launch Checklist

Fixes https://github.com/mapbox/mapbox-gl-js/issues/7312
Superseding https://github.com/mapbox/mapbox-gl-js/pull/5240

#5240 is out of date and it's easier just to open a new one and close that. JSDom still does not implement `URL.createObjectURL` or `URL.revokeObjectURL` so there's not a good way to add a test for this. It's pretty simple though and worked as expected in manual debugging so I think no test is fine.

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -158,9 +158,14 @@ export const getImage = function(requestParameters: RequestParameters, callback:
         } else if (imgData) {
             const img: HTMLImageElement = new window.Image();
             const URL = window.URL || window.webkitURL;
+            console.log('get image', imgData);
             img.onload = () => {
                 callback(null, img);
                 URL.revokeObjectURL(img.src);
+            };
+            img.onerror = (e) => {
+              const err = new Error('Image sources cannot use SVGs');
+              callback(err);
             };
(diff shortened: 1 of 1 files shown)


src/util/ajax.js
```diff
@@ -158,10 +158,15 @@ export const getImage = function(requestParameters: RequestParameters, callback:
         } else if (imgData) {
             const img: HTMLImageElement = new window.Image();
             const URL = window.URL || window.webkitURL;
+            console.log('get image', imgData);
```
mourner: Leftover `console.log`

src/util/ajax.js
```diff
@@ -158,10 +158,15 @@ export const getImage = function(requestParameters: RequestParameters, callback:
...
             img.onload = () => {
                 callback(null, img);
                 URL.revokeObjectURL(img.src);
             };
+            img.onerror = () => {
+                const err = new Error('Image sources cannot use SVG files');
+                callback(err);
+            };
```
mourner: Maybe let's do a shorter version?

```js
img.onerror = () => callback(new Error('Image sources cannot use SVG files'));
```

PR #7314, opened 2018-09-21
4830 debug tiles

### LINKED ISSUE #4830: showTileBoundaries debug option should read from a source whose min/maxzoom include current zoom if available

### LINKED ISSUE #5535: overzooming tiles in studio not reporting correct zxy in deubg mode

ryanhamley:
This PR implements two improvements to debug tile data:
- Use the source with the highest `maxzoom` so that more granular tile debug data is always available
- Describe the actual and overscale tile ID on debug tiles

Closes #4830
Closes #5535

This re-opens https://github.com/mapbox/mapbox-gl-js/pull/4990. Somehow when trying to rebase that PR, I accidentally removed the commits and closed out #4990 🤷‍♂️  Please refer to that PR for discussion and code review.


src/render/draw_debug.js
```diff
@@ -40,7 +40,7 @@ function drawDebugTile(painter, sourceCache, coord) {
         debugUniformValues(posMatrix, Color.red), id,
         painter.debugBuffer, painter.tileBorderIndexBuffer, painter.debugSegments);
 
-    const vertices = createTextVertices(coord.toString(), 50, 200, 5);
+    const vertices = createTextVerticies(coord.canonical.toString(), 50, 200, 5);
```
asheemmamoowala: typo: `createTextVertices`

src/render/painter.js
```diff
@@ -395,9 +397,22 @@ class Painter {
...
-            for (const id in sourceCaches) {
-                draw.debug(this, sourceCaches[id], coordsAscending[id]);
-                break;
+            //Use source with highest maxzoom
+            let selectedSource;
+            let sourceCache;
+            const layers = values(this.style._layers);
+            layers.forEach((layer) => {
+                if (layer.source && !layer.isHidden(this.transform.zoom)) {
+                    if (layer.source !== (sourceCache && sourceCache.id)) {
```
asheemmamoowala: nit: the null check for `sourceCache` makes this code a bit hard to read.

andrewharvey:
This is closely related to https://github.com/mapbox/mapbox-gl-js/issues/8376#issuecomment-504619187, which also would create conflicts. I'm happy to help rebase this so these changes can be applied cleanly but it might make sense to wait for #8380.

asheemmamoowala:
@andrewharvey Want to take the changes from this into #8380 and help close this out ?

andrewharvey:
Thanks for refactoring this @asheemmamoowala, it looks good to me from just reading the code.

vakila:
Hi folks, just jumping in here to check in as part of some PR gardening we're doing.

@ryanhamley @andrewharvey Where does this stand at the moment? AFAICT it's not waiting on any requested changes, so is it in need of another round of review, or ready to merge?

src/render/painter.js
```diff
@@ -430,9 +431,22 @@ class Painter {
...
 
         if (this.options.showTileBoundaries) {
-            for (const id in sourceCaches) {
-                draw.debug(this, sourceCaches[id], coordsAscending[id]);
-                break;
+            //Use source with highest maxzoom
+            let selectedSource;
+            let sourceCache;
+            const layers = values(this.style._layers);
+            layers.forEach((layer) => {
```
arindam1993: nit: Use a `for..of`?

Also would it be possible to clean up the loop somehow? seems its a bit unnecessarily complex for what it needs to do

ryanhamley: Simplified it a good bit 👍

vakila:
@ryanhamley Thanks for the update! Looks like the `debug/tile` render test is now failing (I tried rerunning the tests in case it was a fluke but got the same failure), and there's also a small lint error to be fixed. I'll leave it to @arindam1993 to speak to the latest changes per his comment.

ryanhamley:
Yep, working on those now. Not sure why my change would affect the test

ryanhamley:
I tried a couple approaches to simplifying the loop and they don't really work well. I ended up having to add back all of the same checks that are in the current PR. I think it's fine as is so I'm going with the original PR.

src/render/painter.js
vakila: Looks like this has become moot as we've reverted to the original `forEach`, but for future reference just noting here that IE can't handle `for...of` :grimacing:

PR #7317, opened 2018-09-21
Add variable binding expression example

malwoodsantoro:
This example uses `let` and `var` expressions to demonstrate variable binding syntax. To test this out locally, you'll need to replace the url to the GeoJSON with the raw data. 

:)


docs/pages/example/visualize-population-density.js
```diff
@@ -0,0 +1,10 @@
+/*---
+title: Visualize population density
+description: Using a [`varible binding expression`](/mapbox-gl-js/style-spec#expressions-variable-binding) to calculate and display population density.
```
ryanhamley: nit: `variable` is misspelled here

docs/pages/example/visualize-population-density.html
```diff
@@ -0,0 +1,53 @@
...
+                        1551,
+                        ["to-color", "#006d2c"]
+                    ],
+                    10,
+                    [
+                        "interpolate",
+                        ["linear"],
+                        ["var", "density"],
+                        274,
+                        ["to-color", "#eff3ff"],
```
ryanhamley: I *think* @ChrisLoer 's recent work makes the explicit cast to a color here unnecessary

ChrisLoer: I think the cast is necessary until https://github.com/mapbox/mapbox-gl-js/pull/7301 has merged and is being used by the example page, right?

ryanhamley: LOL, oh yeah. I was working on that. Fridays.

docs/pages/example/visualize-population-density.html
```diff
@@ -0,0 +1,53 @@
+<div id='map'></div>
+<script>
+var map = new mapboxgl.Map({
+    container: 'map', // container id
+    style: 'mapbox://styles/mapbox/streets-v9', // stylesheet location
```
ChrisLoer: I have a personal vendetta against streets-v9 in examples. The flattened road labels are 🙅‍♂️

ChrisLoer:
> The code is almost identical to our choropleth example but with the added overhead of declaring the variable syntax

🤔 Yeah I see your point although it seems like using `let` and `var` will always be _part_ of accomplishing something else (hopefully with less duplication in the expression code). I guess this seems like a pretty reasonable choice to me.

ryanhamley:
I'm ok with it then. Just a couple little nits to clean up. I'll update the example to remove the explicit casts to color when I finish up the test in https://github.com/mapbox/mapbox-gl-js/pull/7301

PR #7324, opened 2018-09-21
Adding search playground link

samely:
### Context
Our `Search Playground` tool is live in `www.mapbox.com/search-playground`

### Changes
:heavy_check_mark: Added a link to  `www.mapbox.com/search-playground`.

### DIFF
--- a/docs/pages/example/mapbox-gl-geocoder.js
+++ b/docs/pages/example/mapbox-gl-geocoder.js
@@ -3,6 +3,7 @@ title: Add a geocoder
 description: >-
   Use the [mapbox-gl-geocoder](https://github.com/mapbox/mapbox-gl-geocoder)
   control to search for places using Mapbox Geocoding API.
+  Checkout our [Search Playground](https://www.mapbox.com/search-playground/) to test out geocoding quality.
 tags:
   - controls-and-overlays
   - geocoder
--- a/docs/pages/example/mapbox-gl-geocoder.js
+++ b/docs/pages/example/mapbox-gl-geocoder.js
@@ -3,7 +3,7 @@ title: Add a geocoder
 description: >-
   Use the [mapbox-gl-geocoder](https://github.com/mapbox/mapbox-gl-geocoder)
   control to search for places using Mapbox Geocoding API.
-  Checkout our [Search Playground](https://www.mapbox.com/search-playground/) to test out geocoding quality.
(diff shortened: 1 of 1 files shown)


docs/pages/example/mapbox-gl-geocoder.js
```diff
@@ -3,6 +3,7 @@ title: Add a geocoder
 description: >-
   Use the [mapbox-gl-geocoder](https://github.com/mapbox/mapbox-gl-geocoder)
   control to search for places using Mapbox Geocoding API.
+  Checkout our [Search Playground](https://www.mapbox.com/search-playground/) to test out geocoding quality.
```
ChrisLoer: Nit: "check out" instead of "checkout".

"to test out geocoding quality" also seems like it could be phrased a little differently. It's not like going to the search playground gives you an immediate quality metric or comparison to other tools, right? Put another way, what's the difference between the "search playground" and the geocoder results the reader will already be able to see in the geocoder example on the docs page? Seems like if we're going to point people to the playground we want to highlight what extra capability it exposes.

samely: > `Nit: "check out" instead of "checkout".`
:sweat_smile:

Yeah... you are right. We thought about adding it here since this is the starting point to add a geocoder, we also we to show what other options can be set up for a better experience using our geocoder.

ChrisLoer: Ah OK, how about something like "Check out our Search Playground to explore geocoding configuration options and test how they might affect results for your users"?

As a separate piece of feedback -- I didn't discover the "settings" option in the search playground until my third visit to the page, and only once I was explicitly looking for a way to control options. I was naturally just drawn to the search interface, and although I did try clicking a few of the buttons in the upper right, the first two I clicked didn't really grab my interest so I stopped exploring. Is the "settings" functionality a central enough part of the "Search Playground" that it makes sense for it to be showing/selected from the beginning?

samely: Thanks for your suggestion! Already fixed ^^

> As a separate piece of feedback -- I didn't discover the "settings" option in the search playground until my third visit to the page,
Your feedback about the tool is really valuable! We use it frequently and didn't realize this is not intuitive enough. I'm gonna open an issue to fix this! Thank you

PR #7333, opened 2018-09-26
Pixel ratio

### LINKED ISSUE #1953: Android devices, performance and ppi

ryanhamley:
## Launch Checklist

Makes `devicePixelRatio` configurable and fixes #1953

By default, `config` gets the `devicePixelRatio` from `window` if no value is explicitly set.

tip of the 🎩 to @sniok

- [x] Add a devicePixelRatio setting to config.js which defaults to browser.devicePixelRatio
- [x] Add a devicePixelRatio getter and setter to index.js
- [x] Change all uses of browser.devicePixelRatio to config.devicePixelRatio

### DIFF
--- a/docs/documentation.yml
+++ b/docs/documentation.yml
@@ -1,6 +1,7 @@
 toc:
   - Map
   - accessToken
+  - devicePixelRatio
   - supported
   - version
   - setRTLTextPlugin
--- a/src/index.js
+++ b/src/index.js
@@ -76,7 +76,7 @@ const exported = {
     },
 
     set devicePixelRatio(devicePixelRatio: number) {
-        config.DEVICE_PIXEL_RATIO = devicePixelRatio;
+        Object.defineProperty(config, 'DEVICE_PIXEL_RATIO', { value: devicePixelRatio });
     },
 
     workerUrl: ''
--- a/src/index.js
+++ b/src/index.js
@@ -64,6 +64,21 @@ const exported = {
         WorkerPool.workerCount = count;
     },
 
+    /**
+     * Gets and sets map's pixel ratio
+     *
+     * @var {number} devicePixelRation
+     * @example
+     * mapboxgl.devicePixelRatio = 2
+     */
+    get devicePixelRatio() {
+        return config.DEVICE_PIXEL_RATIO;
+    },
+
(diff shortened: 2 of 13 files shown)


AskAlice:
please merge this. When I resize my renderer with three, the mapbox gl resizes at double scale and looks very bugged

Mechazawa:
Setting `window.devicePixelRatio` and calling `map.resize()` this seems to work fine for now untill this PR has been finished/merged. Do note that the devicePixelRatio can change if the browser is moved to a different display.

asheemmamoowala:
This PR has gotten very stale, and there are still many open questions from https://github.com/mapbox/mapbox-gl-js/pull/7333#issuecomment-424800855. Closing for now, but @ryanhamley please re-open if/when this is planned to work on next.

PR #7337, opened 2018-09-26
Less verbose function -> expression conversion

### LINKED ISSUE #7336: Output values in expressions get wrapped in literal expression when they shouldn't
**mapbox-gl-js version**: `@mapbox/mapbox-gl-style-spec` v13.2.0

**browser**: N/A 

### Steps to Trigger Behavior

Convert a function to an expression. Here's an example function that I will use throughout this issue:

```js
'background-opacity': {
  base: 1.0,
  stops: [[0, 1], [10, 0.72]]
}
``` 

### Expected Behavior

```js
'background-opacity': [
  'interpolate',
  ['exponential', 1],
  ['zoom'],
  0,
  1,
  10,
  0.72
]
```

### Actual Behavior
```js
'background-opacity': [
  'interpolate',
  ['exponential', 1],
  ['zoom'],
  0,
  ['literal', 1],
  10,
  ['literal', 0.72]
]
```

While not technically _wrong_ this migration approach results in less readable expressions. These expressions are not handled elegantly in Studio's UI by default. 

The behavior I expect is that array spec property values (like `text-font`), would be wrapped in literals, but non-arrays should not.

Looks like this changed was introduced in https://github.com/mapbox/mapbox-gl-js/pull/7095, specifically, in convertFunction (https://github.com/mapbox/mapbox-gl-js/pull/7095/files#diff-4b96c63e8fcfc96468a92c7a73e53b5e).

mourner:
Fixes #7336. Note that `typeof obj === 'object'` should be true for all non-primitive values in the spec (including arrays), so we don't need more exhaustive checks.

## Launch Checklist

 - [x] ~~write tests for all new functionality~~ covered by existing tests
 - [x] ~~post benchmark scores~~ n/a

### DIFF
--- a/src/style-spec/function/convert.js
+++ b/src/style-spec/function/convert.js
@@ -5,6 +5,10 @@ import type {StylePropertySpecification} from '../style-spec';
 
 export default convertFunction;
 
+function convertLiteral(value) {
+    return typeof value === 'object' ? ['literal', value] : value;
+}
+
 function convertFunction(parameters: any, propertySpec: StylePropertySpecification) {
     let stops = parameters.stops;
     if (!stops) {
@@ -20,7 +24,7 @@ function convertFunction(parameters: any, propertySpec: StylePropertySpecificati
         if (!featureDependent && propertySpec.tokens && typeof stop[1] === 'string') {
             return [stop[0], convertTokenString(stop[1])];
         }
-        return [stop[0], ['literal', stop[1]]];
+        return [stop[0], convertLiteral(stop[1])];
     });
 
     if (zoomAndFeatureDependent) {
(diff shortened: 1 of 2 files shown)


ChrisLoer:
This looks right to me -- it's basically specifically string/boolean/number that we're allowing without the "literal" wrapper, right?

Can we add a legacy expression test that uses @samanpwbb 's function from #7336 ?

mourner:
@ChrisLoer there are already legacy tests like this, e.g. [this one](https://github.com/mapbox/mapbox-gl-js/blob/master/test/integration/expression-tests/legacy/exponential/implicit/test.json), it's just that the tests don't test conversion output itself — only the output after creating a style expression object and serializing it, which somehow strips the unnecessary "literal" wrappers. Not sure how to go about this.

mourner:
@ChrisLoer tried adding comparison of raw-converted legacy function with serialized one, but there are many expressions which don't roundtrip like this — e.g. `["linear"]` will turn into `["exponential", 1]`, and `"black"` will turn into `["rgba", 0, 0, 0, 1]`. It's reassuring that 44 failing tests with the test change get down to 22 in this branch.

PR #7342, opened 2018-09-26
Add ImageSource.updateImage()

### LINKED ISSUE #4050: Add update functionality to ImageSource objects
## Motivation

ImageSource objects only have a create mechanism, not an update. So, if you wanted to swap out the image on an image source with a new image, you have to delete the existing source and create a new one. This is inefficient and can result in unwanted popping when re-rendering.

## Design Alternatives

Add an update method to ImageSource objects.

## Design

Similar to how images are created, the update() method would take as an argument an object with the following optional properties:
```
{
    url: <the image URL>, // if not provided, the image does not change
    coordinates: <optional coordinates to place the image> // if not provided, the existing coords do not change
} 
```
### Mock-Up

Here's an example of how it would look like in code, based of the documented example:

```
// add to map
map.addSource('some id', {
   type: 'image',
   url: 'https://www.mapbox.com/images/foo.png',
   coordinates: [
       [-76.54, 39.18],
       [-76.52, 39.18],
       [-76.52, 39.17],
       [-76.54, 39.17]
   ]
});

// update
var myImageSource = map.getSource('some id');

// update the image source with a new image and new coordinates
myImageSource.update({
    url: 'https://www.mapbox.com/images/bar.png',
    coordinates: [
        [-76.54335737228394, 39.18579907229748],
        [-76.52803659439087, 39.1838364847587],
        [-76.5295386314392, 39.17683392507606],
        [-76.54520273208618, 39.17876344106642]
    ]
});

map.removeSource('some id');  // remove
```

### Implementation

Implementation is fairly straightforward as it is similar to addSource(). Care would need to be taken if the image size changes during the update. The following is what it could look like in code (this is based of v0.28.0:

```
    ImageSource.prototype.updateImage = function(options) {
        if (!this.image || !options.url) {
            return;
        }

        var updateCoords = Boolean(options.coordinates);

        ajax.getImage(options.url, (err, image) => {
            // @TODO handle errors via event.
            if (err) return this.fire('error', {error: err});

            if (image.width != this.image.width ||
                image.height != this.image.height) {
                // set a resized flag
                this._resized = true;
            } else {
                // set an update flag
                this._updated = true;
            }

            this.image = image;

            if (updateCoords && this.map) {
                // update coordinates
                this.setCoordinates(options.coordinates);
                this.fire('change');
            }
        });
    };

    ImageSource.prototype._setTile = function(tile) {
        const gl = this.map.painter.gl;
        
        // the tile holds onto resources used for rendering
        // images, so if the tile changes with the update
        // then the resources on the previous tile should be removed
        if (this.tile) {
            if (tile !== this.tile) {
                delete this.tile.texture;
                delete this.tile.buckets;
                delete this.tile.boundsBuffer;
                delete this.tile.boundsVAO;
            }
        }

        this.tile = tile;
        const maxInt16 = 32767;
        const array = new RasterBoundsArray();
        array.emplaceBack(this._tileCoords[0].x, this._tileCoords[0].y, 0, 0);
        array.emplaceBack(this._tileCoords[1].x, this._tileCoords[1].y, maxInt16, 0);
        array.emplaceBack(this._tileCoords[3].x, this._tileCoords[3].y, 0, maxInt16);
        array.emplaceBack(this._tileCoords[2].x, this._tileCoords[2].y, maxInt16, maxInt16);

        if (!this.boundsBuffer) {
            this.boundsBuffer = Buffer.fromStructArray(array, Buffer.BufferType.VERTEX);
            this.boundsVAO = new VertexArrayObject();
        } else {
            // the tile didn't change with the updates, so
            // reset the tile's boundsBuffer to the new tile coords
            const bufferObj = this.boundsBuffer;
            const data = array.serialize();
            const type = gl[bufferObj.type];

            // NOTE: when in here, the array structure hasn't changed, just the
            // data underneath, so there's no need to recreate the
            // VAO or anything like that.
            gl.bindBuffer(type, bufferObj.buffer);
            gl.bufferSubData(type, 0, data.arrayBuffer);
        }

        // set the tile's render resources
        this.tile.buckets = {};
        this.tile.boundsBuffer = this.boundsBuffer;
        this.tile.boundsVAO = this.boundsVAO;
    };

    ImageSource.prototype._prepareImage = function(gl, image) {
        if (!this.texture) {
            // create the texture resource if it doesn't exist.
            this.texture = gl.createTexture();
            gl.bindTexture(gl.TEXTURE_2D, this.texture);
            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, image);
        } else if (this._resized) {
            // Image was updated and its dimensions changed.
            gl.bindTexture(gl.TEXTURE_2D, this.texture);
            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, image);
            this._resized = false;
        } else if (this._updated || image instanceof window.HTMLVideoElement || image instanceof window.ImageData || image instanceof window.HTMLCanvasElement) {
            gl.bindTexture(gl.TEXTURE_2D, this.texture);
            gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, image);
        }

        if (this.tile.state !== 'loaded') {
            this.tile.state = 'loaded';
            this.tile.texture = this.texture;
        }
    };
```

dcervelli:
Fixes #4050.

update-image

H/t to the folks on the #4050 thread who showed the way to do this.

## Launch Checklist

 - [ ] post benchmark scores


dcervelli:
As pointed out by the original submitter of the bug, I need to test changing the image size before merging.

src/source/image_source.js
```diff
@@ -111,6 +128,42 @@ class ImageSource extends Evented implements Source {
...
+     * @returns {ImageSource} this
+     */
+    updateImage(options) {
+        if (!this.image || !options.url) {
+            return this;
+        }
+
+        var updateCoords = Boolean(options.coordinates);
+
+        getImage(this.map._transformRequest(options.url, ResourceType.Image), (err, image) => {
```
ryanhamley: There's a lot of repetition of code in this. We could DRY it up quite a bit by using the existing `load` and `_finishLoading` functions which has the added bonus of firing the appropriate events and updating `this.url`, etc. You have to change `_finishLoading` so it can accept the coordinates as a parameter and it looks like you'd have to change `load` to accept a callback so you can pass in the necessary functionality like `this.texture = null` but otherwise it seems to work pretty well.

src/source/image_source.js
```diff
@@ -111,6 +128,42 @@ class ImageSource extends Evented implements Source {
...
+            if (err) {
+                this.fire(new ErrorEvent(err));
+            } else if (image) {
+                this.image = browser.getImageData(image);
+                this.texture = null;
+                this.fire(new Event('data', {dataType:'source', sourceDataType: 'content'}));
+            }
+
+            if (updateCoords && this.map) {
```
ryanhamley: Just check `options.coordinates` here instead of casting to a Boolean and saving the result as a variable.

debug/update_image.html
```diff
@@ -0,0 +1,103 @@
...
+    container: 'map',
+    minZoom: 1,
+    zoom: 6,
+    center: [center.lng, center.lat],
+    style: 'mapbox://styles/mapbox/streets-v10',
+    hash: false
+});
+
+const path = "/test/integration/image/";
+const images = [
```
ryanhamley: I'm not sure if we need quite so many gifs to be added to the library. You can illustrate this with just a few. In fact, you can probably modify this to use `'https://www.mapbox.com/mapbox-gl-js/assets/radar' + currentImage + '.gif'` as seen [here](https://www.mapbox.com/mapbox-gl-js/example/animate-images/) and eliminate the need for these gifs entirely.

debug/update_image.html
```diff
@@ -0,0 +1,103 @@
...
+]
+
+map.on("load", () => {
+    let currentImage = 0;
+    const getPath = () => path + images[currentImage];
+
+    map.addSource("radar", {
+        type: "image",
+        url: getPath(),
+        coordinates: [
```
ryanhamley: Might be good to also update the coordinates on each iteration so that can easily be debugged.

src/source/image_source.js
```diff
@@ -44,15 +44,32 @@ import type {
...
  *     [-76.54335737228394, 39.18579907229748],
  *     [-76.52803659439087, 39.1838364847587],
  *     [-76.5295386314392, 39.17683392507606],
  *     [-76.54520273208618, 39.17876344106642]
  * ]);
- *
+ * 
+ * // update url. 
+ * // Note: this is most effective if your layer has the `raster-fade-duration` paint property set to 0.
+ * mySource.updateImage({
```
ryanhamley: We probably only need one example here. Just show it with `url` and `coordinates`. The docs should note that `coordinates` is optional.

dcervelli:
@ryanhamley I think I got all of the comments addressed except for a unit test around an image update with a size change. Not really sure how to do that one. FWIW, the code in `image_source.js` doesn't ever do anything with width/height -- all of that is handled in the `Texture`/`ImageData` code -- and I've covered this pretty thoroughly in manually testing.

debug/update_image.html
```diff
@@ -0,0 +1,78 @@
...
+
+    setInterval(() => {
+        currentImage = (currentImage + 1) % frameCount;
+        map.getSource("radar").updateImage({
+            url: getPath(),
+            coordinates
+        });
+
+        // To illustrate moving coordinates at the same time
+        coordinates.forEach(c => {
```
ryanhamley: This moves the image off the screen pretty rapidly and eventually will cause an error when the latitude goes above 90°. I know I suggested this but maybe it would be better to remove it. Sorry about that.

dcervelli: No problem.

PR #7348, opened 2018-09-28
add 'empty' tile state to prevent unnecessary reloads

### LINKED ISSUE #7346: map.loaded() returns "false" when setFilter() is used on a layer with missing tiles within the viewport
When using custom vector tiles, empty tiles will return a 404, which is expected behavior. However, if a we have missing tiles within the viewport and we use setFilter() on click on a layer using the missing tiles, map.loaded() will continuously return `false`. 

**mapbox-gl-js version**: 0.49

**browser**: Chrome, FireFox, Safari

### Steps to Trigger Behavior

 1. Create custom vector tiles
 2. Add a layer to filter and a layer to set the on click event on. 
 3. Add an on click event to the layer
 4 . Add a `setFiler()` on click
 5. Set an interval logging `map.loaded()`

### Link to Demonstration

Here is a jsFiddle: https://jsfiddle.net/f27j8rqs/

screen shot 2018-09-27 at 11 14 18 am

### Expected Behavior

`map.loaded()` should return `true`. 

### Actual Behavior

`map.loaded()` returns `false`.

mollymerp:
fix #7346 

Since #6803 changed 404 tiles to not be treated as errored, the benefits of #6813 regressed for tiles that error with 404, and this causes unnecessary tile reloads that prevent map state from returning to `true`. 

Alternatives:
I think adding an `empty` state adds some clarity to TileState, but this issue could also be fixed in one line by checking if a tile has any buckets before reloading. Curious if the team would prefer that approach. 

cc @lobenichou 

## Launch Checklist

 - [x briefly describe the changes in this PR

### DIFF
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -121,7 +121,7 @@ class SourceCache extends Evented {
         if (!this._sourceLoaded) { return false; }
         for (const t in this._tiles) {
             const tile = this._tiles[t];
-            if (tile.state !== 'loaded' && tile.state !== 'errored')
+            if (tile.state !== 'loaded' && tile.state !== 'errored' && tile.state !== 'empty')
                 return false;
         }
         return true;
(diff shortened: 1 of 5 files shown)


ChrisLoer:
Where would the empty bucket check for reloading be if we used that approach? I don't have strong feelings about this, but my initial reaction is that `empty` and `loaded` seem like really similar concepts (I think of them sort of like "can be rendered").

mollymerp:
closing in favor of a simpler solution at #7355

PR #7355, opened 2018-10-02
make sure worker sends response after reloading 'empty' tile

### LINKED ISSUE #7346: map.loaded() returns "false" when setFilter() is used on a layer with missing tiles within the viewport

mollymerp:
This is an alternative to #7348 to fix #7346 that does not require adding an additional tile `state` value, and prevents us from guessing whether a reload is necessary based on data returned from the initial worker parse. This is also nice because now the WorkerTile's `loaded:{}` object should match the tiles marked as `state: 'loaded'` in the foreground.

Prior to this, a reload request for a tile that 404'd or otherwise returned no data from the AJAX request would never call its callback and prevented `map.loaded()` from ever returning `true` due to tiles hanging with `state: 'reloading'`.
## Launch Checklist

### DIFF
--- a/src/source/vector_tile_worker_source.js
+++ b/src/source/vector_tile_worker_source.js
@@ -110,6 +110,8 @@ class VectorTileWorkerSource implements WorkerSource {
             delete this.loading[uid];
 
             if (err || !response) {
+                workerTile.status = 'done';
+                this.loaded[uid] = workerTile;
                 return callback(err);
             }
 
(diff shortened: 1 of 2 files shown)


src/source/vector_tile_worker_source.js
```diff
@@ -110,6 +110,8 @@ class VectorTileWorkerSource implements WorkerSource {
             delete this.loading[uid];
 
             if (err || !response) {
+                workerTile.status = 'done';
+                this.loaded[uid] = workerTile;
```
mollymerp: hmm maybe we only want to do this `if (!response)` 💭

PR #7358, opened 2018-10-04
Remove code repetition in value.js

mourner:
Made a base class for value classes to remove code repetition and improve clarity.

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/gl/value.js
+++ b/src/gl/value.js
@@ -19,7 +19,6 @@ import type {
 } from './types';
 
 export interface Value<T> {
-    context: Context;
     current: T;
     default: T;
     dirty: boolean;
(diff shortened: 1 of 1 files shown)


mourner:
image

src/gl/value.js
```diff
@@ -28,920 +27,434 @@ export interface Value<T> {
...
-    set(v: any): void {
-        if (this.current !== v && this.context.extVertexArrayObject || this.dirty === true) {
-            this.context.extVertexArrayObject.bindVertexArrayOES(v);
-            this.current = v;
-            this.dirty = false;
+    getDefault(): any {
+        return null;
+    }
+    _set(v: any) {
+        if (this.vao) {
```
ChrisLoer: Thinking out loud to document the change here...

Before, if we didn't have `context.extVertexArrayObject`, we would rely on code in `context.js` to never set `this.dirty` to `true`, so we'd never try to bind the object. Now, we don't depend on that code in `context.js` to keep us from attempting to bind, but it still prevents us from storing any value in `this.current` (although it also looks like all the calls to `bindVertexArrayOES.set` also check for the extension before running).

So looks like just one more layer of belt-and-suspenders protection against accidentally using a missing extension. :+1:.

src/gl/value.js
```diff
@@ -28,920 +27,434 @@ export interface Value<T> {
...
-
-    constructor(context: Context, parent: WebGLFramebuffer) {
-        super(context, parent);
-        this.dirty = false;
-    }
-
-    setDefault(): void { this.set(this.default); }
-
-    set(v: ?WebGLTexture): void {
-        if (this.dirty || this.current !== v) {
```
ChrisLoer: Was there a reason that the dirty/inequality check order was reversed here before? I'm not actually sure what the equality operator for `WebGLTexture` does although it seems unlikely to make any difference that we can now call it more frequently (when the `ColorAttachment` is dirty).

mollymerp: I think the base class `set` function does the dirty/inequality check before calling each child class' `_set` function.

mourner: I think there's no difference — comparing boolean shouldn't be slower than comparing object references.

mollymerp: ooh I missed the key word `reversed` – nvm!

mourner:
It looks like the _set/_equals methods don't get inlined for some reason, so while there's a few percent slowdown in profiles which might not reflect on the paint benchmark, I'm not comfortable merging it. So I'll try to find out why V8 doesn't inline, and if it can't be helped, roll back halfway so that at least `set` methods are written out in full (but the PR will still shave off a lot of repetition).

mourner:
OK, I reintroduced explicit `set` methods for each type, which makes it more verbose but still better than `master` — profiles and Paint benchmark look good.

image

PR #7363, opened 2018-10-04
fix custom layers example in safari

ansis:
It was relying on the ability of some browsers to render without any buffers or attributes. I guess that isn't allowed by the spec. This expands the example to use attributes and buffers, which is probably better anyways.

## Launch Checklist


kkaefer:
(the `SecurityError`s are related to `window.history.replaceState` calls and seem unrelated to GL)

ansis:
[It works for me in IE](https://user-images.githubusercontent.com/1421652/46504298-882a6a80-c7fb-11e8-928b-ba179c6a7a45.png). Maybe it was a caching issue?

PR #7364, opened 2018-10-04
Add `SymbolLayout` benchmark

ChrisLoer:
This benchmark has the same setup as `Layout`, but narrows in to focus on `performSymbolLayout` -- essentially text shaping and line breaking. These show up as a pretty big chunk of layout time, so I think it makes sense to try to bench them separately so we can detect changes that might be harder to see in the Layout benchmark as a whole. I may try to retroactively apply this bench to the recent `format` changes, and I expect this bench would be a good one to watch with any changes we make for "dynamic label placement".

I had to change the `WorkerTile` interface a little in order to pass the glyph/icon dependencies back to the benchmark -- hopefully that doesn't have any discernible effect on how the callbacks work in production.

## Launch Checklist

 - [x] post benchmark scores

screenshot 2018-10-04 12 17 45
screenshot 2018-10-04 12 17 37

cc @mourner @mollymerp @ryanhamley

### DIFF
--- a/bench/benchmarks/symbol_layout.js
+++ b/bench/benchmarks/symbol_layout.js
@@ -1,0 +1,47 @@
+// @flow
+
+import Layout from './layout';
+import SymbolBucket from '../../src/data/bucket/symbol_bucket';
+import { performSymbolLayout } from '../../src/symbol/symbol_layout';
+import { OverscaledTileID } from '../../src/source/tile_id';
+
+export default class SymbolLayout extends Layout {
+    parsedTiles: Array<any>;
+
+    constructor(style: string, locations: ?Array<OverscaledTileID>) {
+        super(style, locations);
+        this.parsedTiles = [];
+    }
+
+    setup(): Promise<void> {
+        return super.setup().then(() => {
+            // Do initial load/parse of tiles and hold onto all the glyph/icon
+            // dependencies so that we can re-do symbol layout in isolation
+            // during the bench step.
+            return Promise.all(this.tiles.map(tile =>
+                this.parser.parseTile(tile, true).then((tileResult) => {
+                    this.parsedTiles.push(tileResult);
+                })
+            )).then(() => {});
+        });
+    }
+
+    bench() {
+        let promise = Promise.resolve();
+        for (const tileResult of this.parsedTiles) {
+            promise = promise.then(() => {
+                for (const bucket of tileResult.buckets) {
+                    if (bucket instanceof SymbolBucket) {
+                        performSymbolLayout(bucket,
+                                            tileResult.glyphMap,
+                                            tileResult.glyphPositions,
+                                            tileResult.iconMap,
+                                            tileResult.imageAtlas.iconPositions,
+                                            false);
+                    }
+                }
+            });
+        }
+        return promise;
+    }
+}
--- a/bench/versions/benchmarks.js
+++ b/bench/versions/benchmarks.js
@@ -2,6 +2,7 @@ import mapboxgl from '../../src';
 import accessToken from '../lib/access_token';
 import { summaryStatistics, regression } from '../lib/statistics';
 import updateUI from '../benchmarks_view';
+import styleLocations from '../lib/style_locations';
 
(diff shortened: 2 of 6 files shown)


ChrisLoer:
I switched the benchmark to use the locations from `style_locations` because I wanted to make sure to capture CJK-heavy tiles (which have much more expensive line breaking). Since this benchmark is significantly narrower than `Layout`, I think we can get away with benching more tiles without the benchmark taking _too_ much longer to run.

screenshot 2018-10-04 12 51 16

bench/versions/benchmarks.js
```diff
@@ -47,6 +49,7 @@ register(new PaintStates(center));
 LayerBenchmarks.forEach((Bench) => register(new Bench()));
 register(new Load());
 register(new LayoutDDS());
+register(new SymbolLayout(style, styleLocations.map(location => location.tileID[0])));
```
ryanhamley: This seems like it would be useful in `styles/benchmarks` too

src/source/worker_tile.js
```diff
@@ -196,7 +198,10 @@ class WorkerTile {
                     featureIndex,
                     collisionBoxArray: this.collisionBoxArray,
                     glyphAtlasImage: glyphAtlas.image,
-                    imageAtlas: imageAtlas
+                    imageAtlas: imageAtlas,
+                    glyphMap: this.returnDependencies ? glyphMap : null,
+                    iconMap: this.returnDependencies ? iconMap : null,
+                    glyphAtlas: this.returnDependencies ? glyphAtlas : null
```
ryanhamley: Does it make sense to just pass through `glyphAtlas.positions` to reduce the amount that's transferred between threads?

ChrisLoer: 🤔 That's probably a good idea since the positions are all that's needed, although none of this actually transfers between threads in the bench case and `returnDependencies` shouldn't be true in the "live" case.

bench/versions/benchmarks.js
ChrisLoer: Yeah, it could be useful, but I'm kind of worried that the style benchmarks are already really long running and kind of intimidating to browse through -- I'm wary of adding lots more benchmarks to them. My thinking is kind of that for the style benchmarking we want to be able to catch changes in overall performance for specific areas (given changes in the style), whereas for version benchmarking we want to catch changes in performance of specific parts of the code over a broad set of areas (given changes in the code).

cc @nickidlugash

src/source/worker_tile.js
ChrisLoer: Made the change you suggested.

bench/versions/benchmarks.js
ryanhamley: I think that makes sense

bench/benchmarks/symbol_layout.js
```diff
@@ -0,0 +1,51 @@
...
+            // dependencies so that we can re-do symbol layout in isolation
+            // during the bench step.
+            let promise = Promise.resolve();
+            for (const tile of this.tiles) {
+                promise = promise.then(() => {
+                    return this.parser.parseTile(tile, true).then((tileResult) => {
+                        this.parsedTiles.push(tileResult);
+                    });
+                });
+            }
```
mourner: No need to load tiles serially in setup: 

```js
Promise.all(this.tiles.map(tile => this.parser.parseTile ...`
```

src/source/worker_tile.js
```diff
@@ -196,7 +198,10 @@ class WorkerTile {
                     featureIndex,
                     collisionBoxArray: this.collisionBoxArray,
                     glyphAtlasImage: glyphAtlas.image,
-                    imageAtlas: imageAtlas
+                    imageAtlas: imageAtlas,
+                    glyphMap: this.returnDependencies ? glyphMap : null,
+                    iconMap: this.returnDependencies ? iconMap : null,
+                    glyphPositions: this.returnDependencies ? glyphAtlas.positions : null
```
mourner: Nit: add a comment about benchmarking use here too.

PR #7366, opened 2018-10-05
Filter validation should enforce lack of feature-state support

### LINKED ISSUE #7352: setFilter allows invalid filters (feature-state)
As a follow-on from this ticket: https://github.com/mapbox/mapbox-gl-js/issues/6876

If for implementation reasons using `feature-state` is unsupported in `setFilter`, a `setFilter` call using that expression should throw an error, rather than setting up a filter that does not work.

Demo: https://jsbin.com/xutawinuso/1/edit?html,js,output

Expected behaviour: either a filter that works, or an error.

Actual behaviour: a filter that does not work.

asheemmamoowala:
Closes #7352.

Style Filters cannot accept `feature-state` expressions and should not validate when used with such expressions.

### DIFF
--- a/src/style-spec/validate/validate_expression.js
+++ b/src/style-spec/validate/validate_expression.js
@@ -23,5 +23,10 @@ export default function validateExpression(options: any) {
         (!isStateConstant((expression.value: any)._styleExpression.expression))) {
         return [new ValidationError(options.key, options.value, '"feature-state" data expressions are not supported with layout properties.')];
     }
+
+    if (options.expressionContext === 'filter' && !isStateConstant((expression.value: any).expression)) {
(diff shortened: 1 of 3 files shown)


mourner:
@asheemmamoowala anything else to do here? Seeing the `[WIP]` in the title...

asheemmamoowala:
@mourner The unit tests weren't passing for me locally earlier, but it seems to have been resolved with a re-install of node modules.

ansis:
Should the validation be based on the presence of the `feature-state` ExpressionParameter in the spec for that property instead of whether it is a layout property or filter?

asheemmamoowala:
Tracking style-spec based validation in #7382

PR #7368, opened 2018-10-05
Switch shaders to ES modules, minify them for smaller bundle

mourner:
A more elaborate attempt at #6349. Advantages:

- Stripping the remains of Browserify and CommonJS from the codebase, making things more modern and consistent.
- Switching to the Rollup plugin system allows us more flexibility. In this particular case, I added a simple regex-based GLSL minification that strips 30kb from the minified bundle (5kb gzipped).

The big number of added lines comes from updated Yarn.lock where a new Yarn version added a new `integrity` field (SHA256 checksum) to each package.

## Launch Checklist


mourner:
> after bumping the esm dependency on the native side.

Hmm what do you mean? There's no `esm` dep in native, only in GL JS.

PR #7369, opened 2018-10-05
Allow browser context menu when possible

### LINKED ISSUE #2301: Ability to re-enable browser-native contextmenu
**mapbox-gl-js version**:

### Steps to Trigger Behavior

 1. Optionally call `map.off('contextmenu')`
 2. Right-click on the map.

### Expected Behavior

The browser's context menu will appear.

### Actual Behavior

The [_onContextMenu handler still prevents the event from having a default action](https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/interaction.js#L149).

mourner:
Closes #2301. Enables standard browser context menu on maps where rotation is disabled and there are no `contextmenu` listeners. 

Unfortunately I don't see a way to enable it by default because it's not compatible with rotation — on Mac, `contextmenu` is fired together with `mousedown` (not `mouseup` like in other browsers), and at this point we don't know if it's a start of a rotation gesture or a simple right click. cc @jplante

## Launch Checklist

### DIFF
--- a/src/ui/bind_handlers.js
+++ b/src/ui/bind_handlers.js
@@ -178,7 +178,11 @@ export default function bindHandlers(map: Map, options: {interactive: boolean, c
             contextMenuEvent = e;
         }
 
-        e.preventDefault();
+        // prevent browser context menu when necessary; we don't allow it with rotation
+        // because we can't discern rotation gesture start from contextmenu on Mac
+        if (map.dragRotate.isEnabled() || map.listens('contextmenu')) {
+            e.preventDefault();
+        }
     }
 
     function onWheel(e: WheelEvent) {


src/ui/bind_handlers.js
```diff
@@ -178,7 +178,11 @@ export default function bindHandlers(map: Map, options: {interactive: boolean, c
             contextMenuEvent = e;
         }
 
-        e.preventDefault();
+        // prevent browser context menu when necessary; we don't allow it with rotation
+        // because we can't discern rotation gesture start from contextmenu on Mac
+        if (map.dragRotate.isEnabled() || map.listens('contextmenu')) {
```
ryanhamley: I understand the first part of this conditional but am I thinking about it correctly that `map.listens('contextmenu')` would be `true` if a `contextmenu` listener has been set up on the map? Why would having a listener prevent the event? It seems like odd behavior that I could disable drag rotate, set up a listener for `contextmenu` and then the context menu wouldn't actually open.

Also, the docs say that `listens` should return `false` if no listener is set up, but I actually get `undefined`.

mourner: The original reason for having Map `contextmenu` event is to allow specifying custom behavior on right click, such as opening your own HTML-styled app-specific context menu, and in this case it doesn't make sense to also show the browser default one. 

Leaflet [does the same thing](https://github.com/Leaflet/Leaflet/blob/9fda888df2791ca0df6f182a9d91d393ee49034c/src/map/Map.js#L1369) and no one raised concerns about it yet, so I think it's safe to do the same logic here.

ryanhamley: Ah, ok that makes sense. I tested this on Mac with Safari, Firefox and Chrome and they all worked as expected so I think this is good.

ansarikhurshid786: This is not working in ios WKWebView.

PR #7371, opened 2018-10-05
Use Fetch API in browsers that support it

kkaefer:
We're currently using the [XmlHttpRequest](https://github.com/mapbox/mapbox-gl-js/blob/53cf0d6f565e7dbb9bac803e8e067b7cb614895d/src/util/ajax.js#L64-L73) to retrieve data like tiles from the backend server.

This PR switches to the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) where it is supported to enable sending correct referrers from the worker bundle. XHR requests don't send a correct referrer for code from a blob URL worker bundle in Safari, Chrome and Edge (and the `Referer` header is disallowed in this API), while the Fetch API allows us to correctly set the page's referrer URL, as demonstrated by this small snippet:

```html
<script>
let worker = new Worker(window.URL.createObjectURL(new Blob([`
const referrer = "${document.location.href}";
const url = 'https://example.com'

const xhr = new XMLHttpRequest();
xhr.open('GET', url, true);
xhr.send();

fetch(new Request(url, { referrer }))
`], { type: 'text/javascript' })));
</script>
```

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -11,7 +11,7 @@ import GlyphManager from '../render/glyph_manager';
 import Light from './light';
 import LineAtlas from '../render/line_atlas';
 import { pick, clone, extend, deepEqual, filterObject, mapObject } from '../util/util';
-import { getJSON, ResourceType } from '../util/ajax';
+import { getJSON, getReferrer, ResourceType } from '../util/ajax';
 import { isMapboxURL, normalizeStyleURL } from '../util/mapbox';
 import browser from '../util/browser';
 import Dispatcher from '../util/dispatcher';
@@ -143,6 +143,8 @@ class Style extends Evented {
         this._loaded = false;
 
         this._resetUpdates();
+
+        this.dispatcher.broadcast('setReferrer', getReferrer());
 
         const self = this;
         this._rtlTextPluginCallback = Style.registerForPluginAvailability((args) => {
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -188,12 +188,12 @@ class Style extends Evented {
         url = normalizeStyleURL(url, options.accessToken);
         const request = this.map._transformRequest(url, ResourceType.Style);
 
-        this._request = getJSON(request, (error, json) => {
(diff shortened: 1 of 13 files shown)


src/style/style.js
```diff
@@ -144,6 +144,8 @@ class Style extends Evented {
 
         this._resetUpdates();
 
+        this.dispatcher.broadcast('setReferrer', getReferrer());
```
kkaefer: We are dispatching the referrer to the workers when we set up the Style object. On single page applications, the pathname may change, but we'll continue to use the initial pathname. Should we rebroadcast a new pathname?

test/unit/util/ajax.test.js
```diff
@@ -17,18 +17,6 @@ test('ajax', (t) => {
...
-        window.server.respondWith(request => {
-            request.respond(200, {'Content-Type': 'image/png'}, '');
-        });
-        getArrayBuffer({ url:'' }, (error) => {
-            t.pass('called getArrayBuffer');
-            t.ok(error, 'should error when the status is 200 without content.');
-            t.end();
-        });
-        window.server.respond();
-    });
```
kkaefer: I've removed the special casing for empty array buffers. There's nothing inherently wrong with empty 200 responses. The reason we originally added this special case is for images that had an empty response, but that is already handled directly in the `getImage` call.

test/unit/util/mapbox.test.js
```diff
@@ -300,14 +322,14 @@ test("mapbox", (t) => {
         t.test('does not POST when mapboxgl.ACCESS_TOKEN is not set', (t) => {
             config.ACCESS_TOKEN = null;
 
-            event.postTurnstileEvent([' a.tiles.mapxbox.com']);
+            event.postTurnstileEvent(mapboxTileURLs);
```
kkaefer: We're passing full tile URLs from the TileJSON file here, not just host names.

Issue #7115, opened 2018-08-12
Tooltip for navigation controls


mmmatthew:
## Motivation
The compass navigation function is not well-known, and could use an explanation. Just clicking the button does not do anything unless the map was rotated.

## Design
A simple tooltip could clarify this, for desktop users at least.

andrewharvey:
I agree we should have `title` attributes on the button which browsers will render as a tooltip on hover.

bravecow:
For compass it would be better to hide like google maps do, I think:

https://bravecow.github.io/mapbox-gl-controls/#p-zoom-compass

ansis:
Thanks for noticing! https://github.com/mapbox/mapbox-gl-js/pull/7373 adds tooltips

PR #7376, opened 2018-10-08
add `showCompassWhenNorth` option to NavigationControl

ansis:
When the option is false the compass is hidden when the map is not rotated. This option keeps the UI cleaner while still providing a way to reset bearing after a user has rotated.

Suggested in #7115.

**I'm not sure we should add this. I think it is a minor improvement but I'm not sure whether we should support this in core. I threw this together so it would be easier to 👍 or 👎 this feature request.**

If we do want to do this, is the naming good? Any other ides? Would negating it (`hideCompassWhenNorth`) be clearer?

screen shot 2018-10-08 at 3 52 07 pm screen shot 2018-10-08 at 3 52 22 pm
unrotated and rotated appearance

## Launch Checklist

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -13,7 +13,8 @@ type Options = {
 
 const defaultOptions: Options = {
     showCompass: true,
-    showZoom: true
+    showCompassWhenNorth: true,
+    showZoom: true,
 };
 
 /**
@@ -22,6 +23,7 @@ const defaultOptions: Options = {
  * @implements {IControl}
  * @param {Object} [options]
  * @param {Boolean} [options.showCompass=true] If `true` the compass button is included.
+ * @param {Boolean} [options.showCompassWhenNorth=true] If `true` the compass button is shown even when the map is not rotated.
  * @param {Boolean} [options.showZoom=true] If `true` the zoom-in and zoom-out buttons are included.
  * @example
  * var nav = new mapboxgl.NavigationControl();
@@ -61,6 +63,15 @@ class NavigationControl {
     _rotateCompassArrow() {
         const rotate = `rotate(${this._map.transform.angle * (180 / Math.PI)}deg)`;
         this._compassArrow.style.transform = rotate;
+
+        if (!this.options.showCompassWhenNorth) {
+            const display = this._map.getBearing() === 0 ? 'none' : '';
+            if (this.options.showZoom) {
+                this._compass.style.display = display;
+            } else {
+                this._container.style.display = display;
+            }
+        }
     }
 
     onAdd(map: Map) {


ryanhamley:
I think the naming is confusing. `hideCompassWhenNorth` or `hideCompassOnNorth` are clearer in my opinion. Maybe something like `alwaysShowCompass` with `true` as default and the docs make it clear that `false` hides the compass when pointing north?

I don't actually see that Google Maps does this but maybe I'm missing something. The UX is a little weird to click the button to return to north then have the button disappear but I think it's probably fine.

ansis:
> I don't actually see that Google Maps does this but maybe I'm missing something. The UX is a little weird to click the button to return to north then have the button disappear but I think it's probably fine.

I think the behavior the user was referring to only happens on mobile.

>  The UX is a little weird to click the button to return to north then have the button disappear but I think it's probably fine.

Yep, I'm not entirely sure this option is necessary.

pathmapper:
There might be another use case where such a `hideCompassWhenNorth` option would make sense:

[Disable map rotation](https://www.mapbox.com/mapbox-gl-js/example/disable-rotation/
) and [display map navigation controls](https://www.mapbox.com/mapbox-gl-js/example/navigation/
) at the same time.

Because it is possible to rotate the map on desktop also with a "hold left click" on the compass icon, the user is still able to rotate the map:

https://jsbin.com/pawejonaya/edit?html,output

btw: the tooltip added for the compass in https://github.com/mapbox/mapbox-gl-js/pull/7373 is not describing the whole functionality.

bravecow:
By the way, why compass was merged with NavigationControl? Why not to split them to ZoomControl and CompassControl? (like we do in mapbox-gl-controls)

1. You will not need `showCompass` option.
2. You can reorder buttons. We use CompassControl at the top of ZoomControl.

bravecow:
@pathmapper I never knew before that you can drag compass control 😏such weird behavior, especially on mobile. and it goes against options, like your example.

ryanhamley:
> By the way, why compass was merged with NavigationControl? Why not to split them to ZoomControl and CompassControl? (like we do in mapbox-gl-controls)
> 
> 1. You will not need `showCompass` option.
> 2. You can reorder buttons. We use CompassControl at the top of ZoomControl.

Splitting the controls strikes me as something that might be worth considering. The main downside I see is that it would be a fairly significant breaking change. Although, I don't think simply splitting out the compass into its own control would solve the use cases brought up by @pathmapper so something like this PR might still be desired. Do you know if there was any particular rationale behind merging these two controls into one, @ansis?

bravecow:
@ansis ? 🙄

vakila:
Hi folks, just jumping in here to check in as part of some PR gardening we're doing.

It looks like a) there hasn't been any progress on this in the last year, and b) there wasn't a ton of agreement in the first place as to whether or not we need this to be implemented. So for now I'm going to close this PR out, and we can still continue the high-level discussion on the closed PR until we've got a clear decision re whether/how to do this. @ansis if you decide to pick this up again we can reopen when it's ready for another round of review. 

Thanks! :heart:

PR #7383, opened 2018-10-09
explain that filters are only evaluated at integer zoom levels

### LINKED ISSUE #6236: Explain that filters are evaluated only at integer zoom levels

ansis:
Tweak docs to fix #6236.

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -581,7 +581,7 @@
     },
     "filter": {
       "type": "filter",
-      "doc": "A expression specifying conditions on source features. Only features that match the filter are displayed. The `feature-state` expression is not supported in filter expressions."
+      "doc": "A expression specifying conditions on source features. Only features that match the filter are displayed. Filters are only evaluated at integer zoom levels. The `feature-state` expression is not supported in filter expressions."
     },
     "layout": {
       "type": "layout",


src/style-spec/reference/v8.json
```diff
@@ -581,7 +581,7 @@
     },
     "filter": {
       "type": "filter",
-      "doc": "A expression specifying conditions on source features. Only features that match the filter are displayed. The `feature-state` expression is not supported in filter expressions."
+      "doc": "A expression specifying conditions on source features. Only features that match the filter are displayed. Filters are only evaluated at integer zoom levels. The `feature-state` expression is not supported in filter expressions."
```
ryanhamley: This makes it sound like filters are only evaluated when a zoom level changes, but that's not always the case right? The filter in [our example](https://www.mapbox.com/mapbox-gl-js/example/filter-features-within-map-view/) evaluates on pan as well. Would it be more accurate to say something like `zoom-dependent filters will only be evaluated at integer zoom levels"?

ansis:
@ryanhamley is `Zoom expressions in filters are only evaluated at integer zoom levels.` better?

ryanhamley:
Yeah that seems a lot clearer

PR #7396, opened 2018-10-10
tighten validation to disallow expressions as stop values

### LINKED ISSUE #7387: Zoom stop functions permit `literal` expressions, which trigger font API 404 status code
**mapbox-gl-js version**: 0.46.0

**browser**: Google Chrome v69.0.3497.100

### Steps to Trigger Behavior

1. Create a stylesheet with a zoom stop function containing literal expressions (see snippet below).
1. Observe that stylesheet passes validation.
1. Zoom into/out of the map so that the zoom stops are breached.
1. Observe API error in console with format `https://api.mapbox.com/fonts/v1/{username}/literal,{encoded_fonts}/{range}.pbf?access_token={token}`. It looks like the `literal` expression reference is making its way into the font request.

```js
{
  "base": 1,
  "stops": [
    [
      6,
      [
        "literal",
        [
          "Open Sans Regular",
          "Arial Unicode MS Regular"
        ]
      ]
    ],
    [
      7,
      [
        "literal",
        [
          "Open Sans Bold",
          "Arial Unicode MS Regular"
        ]
      ]
    ]
  ]
}
```

### Link to Demonstration

1. Click on [this style link](https://api.mapbox.com/styles/v1/emilymdubois-1/cjn1tbzdo5qtn2rpccyy7atu3.html?fresh=true&title=true&access_token=pk.eyJ1IjoiZW1pbHltZHVib2lzLTEiLCJhIjoiY2piZmF1YjltMnVvYjJ3bXIzbzUzaWpmZCJ9.3rcEI0saBslzhRjCpXzzlA#1.61/0/0).
1. Observe font API error in browser network console.

### Expected Behavior

The `literal` expression reference should be removed before making the font API request.

### Actual Behavior

The font API request contains the `literal` expression reference.

ansis:
This fixes #7387 by tightening validation to not allow expressions within the old-style functions. It does this by passing a more limited valueSpec which disallows expressions while validation stop values.

@emilymdubois @asheemmamoowala @mollymerp could I get a pretty careful review of this? This is my first time working on style validation. I'm not sure if this could have unintended consequences. I know we have had issues with tightened validation in the past.

## Launch Checklist

### DIFF
--- a/src/style-spec/validate/validate_function.js
+++ b/src/style-spec/validate/validate_function.js
@@ -1,6 +1,7 @@
 
 import ValidationError from '../error/validation_error';
 import getType from '../util/get_type';
+import extend from '../util/extend';
 import validate from './validate';
 import validateObject from './validate_object';
 import validateArray from './validate_array';
@@ -138,10 +139,15 @@ export default function validateFunction(options) {
             }, value));
         }
 
+        // Remove `expression` from the value spec because expressions
+        // and functions are not allowed as values within functions.
+        const stopValueSpec = extend({}, functionValueSpec);
+        delete stopValueSpec.expression;
+
         return errors.concat(validate({
             key: `${key}[1]`,
             value: value[1],
-            valueSpec: functionValueSpec,
+            valueSpec: stopValueSpec,
             style: options.style,
             styleSpec: options.styleSpec
         }));
--- a/src/style-spec/validate/validate_function.js
+++ b/src/style-spec/validate/validate_function.js
@@ -1,7 +1,6 @@
 
 import ValidationError from '../error/validation_error';
 import getType from '../util/get_type';
-import extend from '../util/extend';
 import validate from './validate';
 import validateObject from './validate_object';
 import validateArray from './validate_array';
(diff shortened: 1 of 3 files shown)


emilymdubois:
Confirming that this change produces the expected result in the Studio UI when I attempt to configure an expression within a stop value:

screen shot 2018-10-12 at 10 13 29 am

I can also confirm that the expression @asheemmamoowala articulated passes validation, but still triggers an invalid GL JSON error modal in the Studio UI:

screen shot 2018-10-12 at 10 17 13 am

Ideally, our validation would catch this before publishing to the stylesheet. If this level of validation is too much of an edge case to incorporate into the function validator, we can also add custom validation on top of the style spec's in Studio.

ansis:
@emilymdubois thanks for checking!

@mollymerp that fix looks cleaner and clearer. Switching to it. It should also handle the edge case that @asheemmamoowala is thinking about. Thanks!

src/style-spec/validate/validate_function.js
```diff
@@ -1,11 +1,13 @@
 
 import ValidationError from '../error/validation_error';
 import getType from '../util/get_type';
+import extend from '../util/extend';
```
mollymerp: I don't think you need `extend` anymore 🙃

ansis: yepp

PR #7403, opened 2018-10-11
Add an unminified production build to the NPM package

mourner:
Provides an alternative build in the published NPM package which is like the production build but unminified (e.g. for use with custom minifiers).

## Launch Checklist

 - [ ] ~~document any changes to public APIs~~ not sure if we should document anywhere

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
   "scripts": {
     "build-dev": "rollup -c --environment BUILD:dev",
     "watch-dev": "rollup -c --environment BUILD:dev --watch",
+    "build-alt": "rollup -c --environment BUILD:production-unminified",
     "build-min": "rollup -c --environment BUILD:production",
     "build-flow-types": "cp build/mapbox-gl.js.flow dist/mapbox-gl.js.flow && cp build/mapbox-gl.js.flow dist/mapbox-gl-dev.js.flow",
     "build-css": "postcss -o dist/mapbox-gl.css src/css/mapbox-gl.css",
@@ -141,7 +142,7 @@
     "test-flow": "build/run-node build/generate-flow-typed-style-spec && flow .",
     "test-flow-cov": "flow-coverage-report -i 'src/**/*.js' -t html",
     "test-cov": "nyc --require=@mapbox/flow-remove-types/register --reporter=text-summary --reporter=lcov --cache run-s test-unit test-expressions test-query test-render",
-    "prepublishOnly": "run-s build-flow-types build-dev build-min build-css build-style-spec test-build",
+    "prepublishOnly": "run-s build-flow-types build-dev build-min build-alt build-css build-style-spec test-build",
     "codegen": "build/run-node build/generate-style-code.js && build/run-node build/generate-struct-arrays.js"
   },
   "files": [
--- a/package.json
+++ b/package.json
@@ -110,8 +110,8 @@
   "scripts": {
     "build-dev": "rollup -c --environment BUILD:dev",
     "watch-dev": "rollup -c --environment BUILD:dev --watch",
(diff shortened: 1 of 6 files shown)


rollup.config.js
```diff
@@ -3,9 +3,12 @@ import sourcemaps from 'rollup-plugin-sourcemaps';
 import {plugins} from './build/rollup_plugins';
 
 const version = JSON.parse(fs.readFileSync('package.json')).version;
-
-const production = process.env.BUILD === 'production';
-const outputFile = production ? 'dist/mapbox-gl.js' : 'dist/mapbox-gl-dev.js';
+const {BUILD} = process.env;
+const minified = BUILD === 'production';
+const production = BUILD === 'production' || BUILD === 'production-unminified';
```
kkaefer: Can we remove the conflation of debug/release and minification? We could have two separate variables: `BUILD` controls dev vs. release, while `UGLIFY` (defaulting to true) controls minification

package.json
```diff
@@ -110,6 +110,7 @@
   "scripts": {
     "build-dev": "rollup -c --environment BUILD:dev",
     "watch-dev": "rollup -c --environment BUILD:dev --watch",
+    "build-alt": "rollup -c --environment BUILD:production-unminified",
```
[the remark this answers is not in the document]
mourner: Just a short name for the package script (which shouldn't be called directly anyway), "alternative" build. Didn't want to name it `build-production-unminified`.

rollup.config.js
```diff
@@ -3,9 +3,12 @@ import sourcemaps from 'rollup-plugin-sourcemaps';
...
 const version = JSON.parse(fs.readFileSync('package.json')).version;
-
-const production = process.env.BUILD === 'production';
-const outputFile = production ? 'dist/mapbox-gl.js' : 'dist/mapbox-gl-dev.js';
+const {BUILD, MINIFY} = process.env;
+const minified = MINIFY === 'true';
+const production = BUILD === 'production';
+const outputFile =
+    production && minified ? 'dist/mapbox-gl.js' :
+    production ? 'dist/mapbox-gl-unminified.js' : 'dist/mapbox-gl-dev.js';
```
kkaefer: This still doesn't account for the matrix of `BUILD`/`MINIFY` yet :) How about
```js
const suffix = (production ? '' : '-dev') + (minified ? '' : '-unminified');
const outputFile = 'dist/mapbox-gl' + suffix + '.js';
```

mourner: @kkaefer I wouldn't want to change names of existing builds. This suggestion would rename `mapbox-gl-dev.js` to `mapbox-gl-dev-unminified.js`.

kkaefer: Let's reverse the MINIFY flag and default to true, then specify `MINIFY:false` for the non-minified production build

mourner: @kkaefer and rename the current `mapbox-gl.js` build to `mapbox-gl-min.js`?

kkaefer: No, the suffix only gets added for unminified builds

mourner: @kkaefer so it would still be `mapbox-gl-dev-unminified.js`? I'm confused.

kkaefer: Sorry for the confusion; I didn't realize that the dev version is unminified. What I was trying to get at is that we have 4 possible combinations of `BUILD` and `MINIFY`, but we're only producing 3 different file names; the one for `BUILD:development,MINIFY:true` produces the same output file.

mourner: Since it doesn't make sense to have a minified dev build anyway, I don't think we need to worry about it. I added another commit that hopefully makes build naming logic clearer. Feel free to add any changes you feel are needed here and I'll review, and otherwise we can merge.

PR #7404, opened 2018-10-11
Add geojson linting

arlyn-bones:
What it says on the tin. We should be properly formatting geojson. Linter doesn't actually yell on any existing code with this change but we should proactively encourage it.


arlyn-bones:
Nevermind, I don't actually see any specific benefit for this. Thanks for your time!

PR #7426, opened 2018-10-16
only align raster sources to pixel grid when map is idle to prevent shaking

### LINKED ISSUE #6150: Tiny map shake while zooming in on raster tiles
In `v0.44` on the ["raster tile source" example](https://www.mapbox.com/mapbox-gl-js/example/map-tiles/), if you zoom in you can notice (at least on Mac Chrome/Firefox/Safari) a slight jitter back and forth:

raster shake on zoom

My initial guess is that this has something to do with the [recent change to align raster tile projection matrices to a pixel grid](https://github.com/mapbox/mapbox-gl-js/pull/6026).

/cc @anandthakker @kkaefer @mollymerp

mollymerp:
fix #6150

~~one note: `map.isZooming()` was never returning true inside the `map._render()` function when using the `ScrollZoomHandler` because that handler returns to `active: false` before render is called, whereas `zoomend` is on a timeout so it is only called once if many scroll frames are triggered in succession. I mutated the `map._zooming` state to account for that, but I'm open to other ideas on how to handle it.~~ removed this change for a separate PR

Any ideas on how to test this? 💭

## Launch Checklist

### DIFF
--- a/src/render/draw_raster.js
+++ b/src/render/draw_raster.js
@@ -28,8 +28,7 @@ function drawRaster(painter: Painter, sourceCache: SourceCache, layer: RasterSty
     const stencilMode = StencilMode.disabled;
     const colorMode = painter.colorModeForRenderPass();
     const minTileZ = coords.length && coords[0].overscaledZ;
-    const align = !painter.options.moving && (Math.round(painter.transform.zoom) === painter.transform.zoom);
-
+    const align = !painter.options.moving;
     for (const coord of coords) {
         // Set the lower zoom level to sublayer 0, and higher zoom levels to higher sublayers
         // Use gl.LESS to prevent double drawing in areas where tiles overlap.
--- a/src/render/draw_raster.js
+++ b/src/render/draw_raster.js
@@ -28,6 +28,7 @@ function drawRaster(painter: Painter, sourceCache: SourceCache, layer: RasterSty
     const stencilMode = StencilMode.disabled;
     const colorMode = painter.colorModeForRenderPass();
     const minTileZ = coords.length && coords[0].overscaledZ;
+    const align = !painter.options.moving && (Math.round(painter.transform.zoom) === painter.transform.zoom);
 
     for (const coord of coords) {
         // Set the lower zoom level to sublayer 0, and higher zoom levels to higher sublayers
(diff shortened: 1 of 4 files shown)


ansis:
In addition to only aligning when the map is moving should we only align the raster matrix when at integer zooms and quarter rotations?

mollymerp:
@ansis I don't think so – prior to the pixel-grid alignment change, raster maps looked blurry at non-integer ZLs and that is why we added the alignment in the first place.

mollymerp:
@ansis @kkaefer can you a take a look at this again?

src/render/draw_raster.js
```diff
@@ -28,6 +28,7 @@ function drawRaster(painter: Painter, sourceCache: SourceCache, layer: RasterSty
     const stencilMode = StencilMode.disabled;
     const colorMode = painter.colorModeForRenderPass();
     const minTileZ = coords.length && coords[0].overscaledZ;
+    const align = !painter.options.moving && (Math.round(painter.transform.zoom) === painter.transform.zoom);
```
kkaefer: I think the ` && (Math.round(painter.transform.zoom) === painter.transform.zoom)` part isn't necessary. Choosing the aligning matrix isn't affecting nearest neighbor vs. bilinear filtering, so it's safe to use the aligned matrix even for fractional zoom levels (in fact, this produces better bilinear filtering, as indicated by the old test case image of the render test updated in this PR).

Another behavior that I'm observing at least in Firefox is that `moving` isn't reset to false correctly when the mouse is released. Another possibility is that it _is_ reset correctly, but no render is triggered when its state changes, leaving the map in a blurry state when it is rotated.

kkaefer: (Noting that this wasn't a problem before because we always used aligned rendering)

mollymerp:
@kkaefer 
>Another behavior that I'm observing at least in Firefox is that moving isn't reset to false correctly when the mouse is released. Another possibility is that it is reset correctly, but no render is triggered when its state changes, leaving the map in a blurry state when it is rotated.

I can't reproduce this on FF with my trackpad (raster layer always returns to render an `aligned` frame after animation completes) – are you using a physical mouse?

kkaefer:
@mollymerp: Yes, I'm using a physical mouse. This issue only manifests itself when the browser window has a pixel width that is odd. It's also not happening on every pan, but frequently enough to see it at least once among 10 pans.

mollymerp:
@kkaefer @ChrisLoer can one of you take another look here?

Issue #7407, opened 2018-10-11
Unused worker threads cause slowdown on machines with > 8 cores


mollymerp:
While investigating performance issues on a user's map load time, I noticed that we were spinning up many more workers than tiles that could be viewed on the current viewport. This user's had 32 cores, so it was spinning up 16 worker threads (`WorkerPool.workerCount = Math.max(Math.floor(browser.hardwareConcurrency / 2), 1);`) but only actually using ~6 of them to parse tiles, and capping the `workerCount` at 6 reduced page load time by ~25% 

@mapbox/gl-core question for the team: 
Should we be using a different metric to set our worker count (for example viewport size/tile size, or  hard cap it at an arbitrary number for machines with many cores) to reduce unnecessary slowdown caused by spinning up workers?

mourner:
I think we should simply cap the number of workers, maybe at 4? Having 16 certainly feels excessive no matter what the machine is.

PR #7430, opened 2018-10-16
Limit the number of workers that can be spawned

ryanhamley:
## Launch Checklist

    - Fixes https://github.com/mapbox/mapbox-gl-js/issues/7407
    - This came up recently when a user had a machine with 32 cores which spawned 16 workers, most of which weren't used but which degraded performance significantly
    - Do we need tests for this? And how would we test it?

### DIFF
--- a/src/util/worker_pool.js
+++ b/src/util/worker_pool.js
@@ -43,4 +43,5 @@ export default class WorkerPool {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ryanhamley:
What's our consensus on the max number of workers that should be spun up? @mollymerp suggested 6 and @mourner suggested 4. I assume the fewer is better as long as there isn't a limit below which we aren't spinning up enough workers.

PR #7431, opened 2018-10-17
Add map.load telemetry event

### LINKED ISSUE #7297: Add a Map Load event to GL-JS
**What is requested in this change?**
mapbox-gl-js will send a map load event when (and only when) the Mapbox access token is set, a Mapbox map is instantiated, and resources are loaded from a Mapbox API. This behavior will be the same as in the mapbox-gl-native SDKs for iOS and Android.

For each instantiation of a Mapbox GL map object, a map load will be accrued upon its first tile request to the Mapbox Maps API. Further requests made by that instance, e.g. those caused by user interactions such as zooming and panning, or by usage of runtime styling methods, will not generate additional map loads.

Other than the IP address, which is retained for a limited period of time per our [privacy policy](https://www.mapbox.com/privacy/#%5BAaMSWwdt%5D), no device identifiers will be sent as part of this event.

**Why is this needed?**
Today, we track how many maps we serve to our customers via API requests. We are adding map loads as a metric that isn’t biased by implementation of/interaction with maps, as API requests are. Additionally, tracking map loads will make our internal web usage metrics consistent with our internal mobile usage metrics.
We think carefully about all tracking we add to our developer tools, and only add what we believe helps us provide a better customer experience.

**Can I opt out of this tracking?**
If your use of mapbox-gl-js does not set the access token (using `mapboxgl.accessToken`) then this code is always disabled. When the access token is set, there is no option to disable this event.

mollymerp:
Adds a `map.load` telemetry event for Mapbox analytics if and only if:
- a mapbox access token is set AND
- tiles are being requested from a mapbox api endpoint

I'm open to better ideas for keeping track of distinct `Map` instances than adding a unique id on the map state 💭

fix #7297 

cc @chloekraw 

## Launch Checklist

### DIFF
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -12,7 +12,7 @@ import type { RequestParameters } from './ajax';
 import type { Cancelable } from '../types/cancelable';
 
 const help = 'See https://www.mapbox.com/api-documentation/#access-tokens';
-const turnstileEventStorageKey = 'mapbox.turnstileEventData';
+const telemEventKey = 'mapbox.eventData';
 
 type UrlObject = {|
     protocol: string,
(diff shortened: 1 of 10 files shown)


kkaefer:
> I'm open to better ideas for keeping track of distinct Map instances than adding a unique id on the map state 💭
I think this is already a pretty straightforward implementation. I have a hard time seeing advantages of another approach, or how it would simplify the current implementation.

src/util/mapbox.js
```diff
@@ -133,19 +133,122 @@ function formatUrl(obj: UrlObject): string {
...
+
+        if (!this.eventData.anonId) {
+            this.fetchEventData();
+        }
+
+        if (!validateUuid(this.eventData.anonId)) {
+            this.eventData.anonId = uuid();
+        }
+
+        const eventsUrlObject: UrlObject = parseUrl(config.EVENTS_URL);
```
asheemmamoowala: Code to send the request is common for all telemetry events, maybe it could be pulled into `TelemetryEvent` with something like:
```
sendTelemetryEvent(eventId, timestamp, { ..other event properties }, callback).
```
It would also ensure that all events correctly save event data and process the queue

src/util/mapbox.js
```diff
@@ -133,19 +133,122 @@ function formatUrl(obj: UrlObject): string {
...
+                warnOnce('Unable to write to LocalStorage');
+            }
+        }
+
+    }
+
+    processRequests() {}
+
+    queueRequest(date: number | {id: number, timestamp: number}) {
+        this.queue.push(date);
```
asheemmamoowala: `id` is not being pushed into the queue. so the `shift` in `MapLoadEvent#processRequests` will always get `undefined`.

src/util/mapbox.js
```diff
@@ -133,19 +133,122 @@ function formatUrl(obj: UrlObject): string {
...
+        }
+    }
+
+    saveEventData() {
+        const isLocalStorageAvailable = storageAvailable('localStorage');
+        const storageKey = `${telemEventKey}:${config.ACCESS_TOKEN || ''}`;
+
+        if (isLocalStorageAvailable) {
+            try {
+                window.localStorage.setItem(storageKey, JSON.stringify(this.eventData));
```
asheemmamoowala: This would override any previously stored event data. In the case of a turnstile request finishing before the map load request, the `lastSuccess` property of the turnstile event would be removed, and more than one turnstile request may be processed. 

The event data could include separate objects for each event type or use a different key for each event type.

test/unit/util/mapbox.test.js
```diff
@@ -616,6 +615,248 @@ test("mapbox", (t) => {
 
         t.end();
     });
+    t.test('MapLoadEvent', (t) => {
```
asheemmamoowala: I think a test of both events together would also be helpful, since they share state/data.

src/util/mapbox.js
mollymerp: mmm I don't think that's true. date is one argument, for turnstile events it is just a number, for map load events it is the object w id and timestamp. its working correctly for me locally and is also verified in the tests 🕵️‍♀️

test/unit/util/mapbox.test.js
```diff
@@ -616,6 +614,271 @@ test("mapbox", (t) => {
...
+        let event;
+        let turnstileEvent;
+        t.beforeEach((callback) => {
+            window.useFakeXMLHttpRequest();
+            event = new mapbox.MapLoadEvent();
```
asheemmamoowala: nit: rename to `mapLoadEvent`

Issue #7338, opened 2018-09-26
What are the effects, if any, of including the mapbox global JS after the map has already begun loading?


tstirrat15:
**mapbox-gl-js version**: Any

### Question
In [the docs](https://www.mapbox.com/mapbox-gl-js/api/), there are instructions to include the mapbox global CSS. However, the map seems to work all right without having that CSS available (as it did in my project, when I was including it incorrectly and the CSS was not available when it needed to be).

First, what are the ramifications of that CSS being unavailable? Are there performance differences? Are there features that will break? I ask because the fact that the map loads at all without that CSS, despite it being "required," is confusing.

Second, if the CSS is subsequently added to the page (such as by JS that adds a `<link />` to the head to the relevant CSS), will it be picked up and used, or will the fact that it was unavailable when the check was made cause the map to behave differently?

I ask because we make use of code splitting and CSS provided by and referenced by JS in our application, and it'd be nice if we didn't have to provide this CSS to the entire site by having it statically in the head when only one subapplication requires the CSS.

### Links to related documentation
https://www.mapbox.com/mapbox-gl-js/api/

mourner:
Good question! The GL map itself should work if CSS is unavailable — only DOM-related things like controls, popups, markers and logo/attribution will be broken, but they'll recover as soon as the CSS loads — so you can do it async, just be mindful of a possible flicker of DOM elements.

The "include CSS in the head section" suggestion in the docs is merely a beginner's quickstart guide, not a requirement.

tstirrat15:
Awesome. Thank you for the quick response. Is there an appropriate place to document this? I'd be happy to submit a PR that includes this information.

mourner:
@tstirrat15 I'm not sure — if you find a good spot to put this in the docs (but not in quickstart so that we don't overwhelm beginners), we'll gladly accept a PR.

PR #7437, opened 2018-10-17
Add documentation about CSS inclusion

tstirrat15:
## Motivation
Our project uses React, Webpack, code-splitting, and ES6 import syntax. Including the Mapbox CSS via the CDN wouldn't have been terrible, but there's a possibility of version mismatch, and we prefer to provide our own assets.

There weren't previously docs for how to go about this. I asked about this in #7338 and received a good answer, and figured I'd document it for people down the road.

## Changes
Add a section after the CSP that documents how Mapbox's CSS works and can/should be included.

## Note
I see I put this on the wrong base; I'll rebase and push shortly.


mollymerp:
thanks for the contribution @tstirrat15! it looks like you have a linting error – should be fixable by running `yarn lint --fix` locally.

tstirrat15:
It was a trailing space >.<

PR #7442, opened 2018-10-18
deregister marker mouseup and touchend when removing marker

vbud:
We have run into scenarios where we remove a draggable marker before `mouseup` is ever fired and, because mouseup/touchend never fired once, `_onUp` is called but the marker has already been torn down, leading to errors.

Ensuring these listeners are deregistered when tearing down the marker will ensure that doesn't happen.

~I didn't see any tests in the marker unit test file that ensure that listeners are being deregistered, so I did not add any tests with this. Let me know if I missed anything.~ Tests added.


vbud:
@mollymerp tests added!

PR #7463, opened 2018-10-22
Upgrade lint-related dependencies

mourner:
We haven't upgraded our lint-related dependencies for a while, so we're missing out on many new fixes/improvements and new rules.

This PR upgrades all these deps (the most major one being ESLint 4->5), and temporarily disables a bunch of rules that the new setup violates. We can either merge now and re-enable those rules in follow-up PRs, or re-enable them here with separate commits to merge all at once.


mourner:
Looks like the builds failed because Circle is using an old version of Node for some reason...

src/symbol/shaping.js
```diff
@@ -209,29 +209,29 @@ function shapeText(text: Formatted,
...
-    [0x20]: true, // space
+    0x09: true, // tab
+    0x0a: true, // newline
+    0x0b: true, // vertical tab
+    0x0c: true, // form feed
+    0x0d: true, // carriage return
+    0x20: true, // space
 };
 
 const breakable: {[number]: boolean} = {
```
ryanhamley: Looks like we'd have to update the Flow declaration here and in line 211.

src/ui/camera.js
```diff
@@ -294,7 +294,7 @@ class Camera extends Evented {
      */
     rotateTo(bearing: number, options?: AnimationOptions, eventData?: Object) {
         return this.easeTo(extend({
-            bearing: bearing
+            bearing
```
ryanhamley: There's a few places like this and line 210 that could be moved onto one line with this change. I don't know if ESLint can be configured to do that automatically.

.eslintrc
```diff
@@ -13,8 +13,24 @@
...
+    "implicit-arrow-linebreak": "off",
+    "arrow-parens": "off",
+    "arrow-body-style": "off",
+    "no-confusing-arrow": "off",
+    "no-control-regex": "off",
+    "no-invalid-this": "off",
+    "no-buffer-constructor": "off",
+
     "array-bracket-spacing": "off",
-    "block-scoped-var": "error",
```
ryanhamley: Why get rid of `block-scoped-var` in addition to `no-var`? I'm assuming some parts of the library still need `var` which is fine, but then shouldn't we want the block scope check?

src/geo/lng_lat_bounds.js
```diff
@@ -28,7 +28,7 @@ class LngLatBounds {
     // This constructor is too flexible to type. It should not be so flexible.
     constructor(sw: any, ne: any) {
         if (!sw) {
-            return;
+            // noop
```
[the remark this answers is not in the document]
mourner: It's equivalent in this case.

package.json
```diff
@@ -127,7 +127,7 @@
     "build-docs": "documentation build --github --format json --config ./docs/documentation.yml --output docs/components/api.json src/index.js",
     "build": "run-s build-docs && . build/set-deploy-env.sh && batfish build # invoked by publisher when publishing docs on the mb-pages branch",
     "start-docs": "run-s build-prod-min build-css build-docs && DEPLOY_ENV=local batfish start",
-    "lint": "eslint --cache --ignore-path .gitignore src test bench docs docs/pages/example/*.html debug/*.html",
+    "lint": "eslint --fix --cache --ignore-path .gitignore src test bench docs docs/pages/example/*.html debug/*.html",
```
mollymerp: ✨🙇‍♀️

src/data/program_configuration.js
```diff
@@ -650,7 +650,7 @@ export default class ProgramConfiguration {
     defines(): Array<string> {
         const result = [];
         for (const property in this.binders) {
-            result.push.apply(result, this.binders[property].defines());
+            result.push(...this.binders[property].defines());
```
mollymerp: I don't think IE supports the spread operator – is this getting transpiled this now?

mourner: Yes: https://buble.surge.sh/#foo.bar(...args)

Issue #3437, opened 2016-10-21
Can't render a simple map example on Android Chrome


gvieiradot:
mapbox-gl.js version: 0.26.0
Chrome version: 53.0
Android version: 4.1.1
Device: Galaxy X
### Steps to Trigger Behavior
1. I created a page with that simple demo: https://www.mapbox.com/mapbox-gl-js/example/check-for-support/;
2. Try to open it via my android device.
### Expected Behavior

Open on my device with no errors;
### Actual Behavior

When I try to open via Android Chrome the map imagery crashes. I can see it for some milliseconds and then it crashes, leaving only the labels;
At my computer it work just great.

lucaswoj:
@gvieiradot Do you see anything in your phone's web browser's console?

gvieiradot:
How can I open the Console in mobile chrome?

lucaswoj:
https://www.google.com/search?q=How+can+I+open+the+Console+in+mobile+chrome

benparkergit:
I have the same issue on latest Chrome, Mac: Version 56.0.2924.87 (64-bit)

ChrisBellew:
I have the same issue using a Samsung Galaxy S6 on BrowserStack.

Android: 5.0.2
Chrome: 55.0.2883.91

It looks like this: http://imgur.com/a/SqOPl
Devtools is mostly empty: http://imgur.com/a/MojYw

andrewharvey:
I noticed that on the Note 4 Chrome browser on BrowserStack too. http://mapbox.github.io/mapbox-gl-supported/diagnostics.html doesn't seem to indicate any issues:

![selection_515](https://cloud.githubusercontent.com/assets/117278/24653365/e3518192-1978-11e7-97de-224139575dc8.png)
![selection_514](https://cloud.githubusercontent.com/assets/117278/24653366/e36e8472-1978-11e7-93dc-13f00c0b220a.png)
![selection_513](https://cloud.githubusercontent.com/assets/117278/24653367/e3d5f814-1978-11e7-9c72-55d2e448a1d6.png)

quillcraft:
I have the same problem on android devices. on the first moment I see map background, and after that showing only text layer with labels.

Android ver. / Chrome ver.: https://www.dropbox.com/s/wtlfskbfu1guqeh/IMAGE%202017-06-21%2012%3A37%3A51.jpg?dl=0

dagjomar:
[...]

PR #7471, opened 2018-10-23
Workaround for failure to clear depth buffer on some Android devices.

ChrisLoer:
Fixes issue #3437.

We don't expect `gl.depthRange` to affect the behavior of `gl.clearDepth`, but experimentally, it appears that we're failing to clear the depth buffer on some Android devices, and resetting the depth range before the clear call appears to fix the problem (maybe because some drivers implement `clearDepth` in terms of just drawing a full screen quad to the depth buffer?). This only happens once per frame, so should be negligibly expensive.

The only place I could reproduce this issue was on a BrowserStack Galaxy S6 image running Chrome 68.0.3440.85/Android 5.0.2 -- we'll need to collect feedback from other people who have encountered #3437 to see how full a fix this is. 

/cc @ansis @mollymerp @chloekraw

### DIFF
--- a/src/gl/context.js
+++ b/src/gl/context.js
@@ -189,6 +189,11 @@ class Context {
 
         if (typeof depth !== 'undefined') {
             mask |= gl.DEPTH_BUFFER_BIT;
+
+            // Workaround for platforms where clearDepth doesn't seem to work
+            // without reseting the depthRange. See https://github.com/mapbox/mapbox-gl-js/issues/3437
+            this.depthRange.set([0, 1]);
+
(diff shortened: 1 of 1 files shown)


ansis:
It might be worth checking whether mapbox-gl-native has any issues that could be solved by this fix.

ChrisLoer:
@tobrun wasn't aware of any Android rendering issues that sounded like this one. Initial feedback from #3437 is that this fix is working, and I also confirmed on the BrowserStack Note 4 image... so I think this makes sense to merge, even though I'd still like to learn more about what's going on.

Issue #7449, opened 2018-10-19
Prototype "bucket merging" for Circle layer


ChrisLoer:
Layers that aren't clipped at tile boundaries (e.g. Circle and Symbol) can't guarantee consistent draw-order sorting of features that cross the tile boundary. See https://github.com/mapbox/mapbox-gl-js/issues/7111.

One of our proposals for addressing this problem is to try to merge all of the tiles for a layer into a single "mega tile" so that all of the features can be drawn in a single draw call. This would give us the technical capability to control the draw order across all tiles, although we'd still need a separate mechanism to actually determine which order to use (e.g. sorting based on y-position would work fine, but the current approach of "sort based on order in the data" wouldn't work outside of a single-tile context).

As a first step, we plan to prototype "bucket merging" for circle layers: our goal is to figure out the performance characteristics of this approach. The approach will require re-uploading the entire "global" buffer every time any tile changes, and it will also require converting tile coordinates into some kind of global coordinate system -- we need to see if we can keep those costs low enough to make this functionality performant.

cc @ansis @tmpsantos @chloekraw

ansis:
dumping some thoughts

### Position precision

> and it will also require converting tile coordinates into some kind of global coordinate system

We'll need to store positions with higher precision if we merge tiles. Buffer size and number of attributes matter, but the largest limitation is the precision of floats in vertex shaders:

<img width="550" alt="screen shot 2018-10-19 at 1 18 36 pm" src="https://user-images.githubusercontent.com/1421652/47233712-92269e80-d3a1-11e8-894e-b5b48f18050d.png">

I think our best bet is using [relative to eye rendering](https://github.com/mapbox/mapbox-gl-js/issues/7268#issuecomment-420722456). Pass 32 bit position components to the shader as two 16 bit integers. Pass the position of the center of the map the same way. Translate the position to a relative-to-center position that preserves precision for positions close to the center. Project this with a matrix that is translated to be relative-to-center.

```glsl
uniform vec2 u_center_pos_high;
uniform vec2 u_center_pos_low;
attribute vec4 a_pos;
...
vec2 pos_high = a_pos.xy;
vec2 pos_low = a_pos.zw;
vec2 pos_relative_to_center = (pos_high - u_center_pos_high) + (pos_low - u_center_pos_low);
gl_FragCoord = u_matrix * vec4(pos_relative_to_center, 0.0, 1.0);
```

[...]

PR #7487, opened 2018-10-26
[WIP] Global coordinates for circle layers

ChrisLoer:
WIP PR for prototyping work described in https://github.com/mapbox/mapbox-gl-js/issues/7449.

TODO:

- [x] Handle pitch, pitch-alignment, and pitch-scaling
- [x] Handle world-wrapping. Since circle layers don't change between wrapped versions of a tile, I'm thinking I can get away with just doing extra draw calls for whichever wrapped versions of the world are showing and apply a different translation matrix for each.
- [x] Initial merging into a single bucket: just concatenate all the individual buffers together, and upload them in their entirety on any changes.
- [ ] More performant bucket-merging? Come up with an allocation scheme for doing partial vertex buffer updates with `bufferSubData`. I'm not sure how hard dealing with fragmentation will be (and also how bad the performance implications will be if we have chunks of "unused" areas in the vertex buffer due to removed tiles/fragmentation).
- [ ] Implement an index-buffer based sorting mechanism for circles, similar to `SymbolBucket#sortFeatures`. We'll also have the limitation that sorting can only be done within a single segment (so ~16,384 circles). I don't think there's a good way around this unless we re-upload everything on every sort change?
- [ ] Get heatmap layer ported to new circle base layer, fix flow, tests, etc.

/cc @ansis

### DIFF
--- a/build/generate-struct-arrays.js
+++ b/build/generate-struct-arrays.js
@@ -124,6 +124,7 @@ createStructArrayType('pos', posAttributes);
 createStructArrayType('raster_bounds', rasterBoundsAttributes);
 
 const circleAttributes = require('../src/data/bucket/circle_attributes').default;
+const heatmapAttributes = require('../src/data/bucket/heatmap_attributes').default;
 const fillAttributes = require('../src/data/bucket/fill_attributes').default;
 const fillExtrusionAttributes = require('../src/data/bucket/fill_extrusion_attributes').default;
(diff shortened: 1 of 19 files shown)


ChrisLoer:
🎩 -tip to @measuredweighed for showing us the "relative to eye" technique we're trying here. 🙂

mourner:
Closing as stale — this turned out not to have too many drawbacks to be shipped. If we ever revisit this, it'll be in a different PR.

PR #7488, opened 2018-10-26
add `MercatorCoordinate` type

### LINKED ISSUE #7485: expose mercator projection for custom layers
## Motivation

The matrix we give custom layers expects "mercator coordinates" as input. We provide no way of converting LngLat to mercator coordinates. We have this projection implemented internally but we don't expose this to users. Exposing it is mainly a design question.

If we can settle on a design I think we could sneak this into the next release.

@mourner can you help pick a design for this?

Relevant comments and workarounds:
- https://github.com/mapbox/mapbox-gl-js/issues/7481
- https://github.com/mapbox/mapbox-gl-js/issues/7268
- https://github.com/peterqliu/threebox/blob/2b4805927c72fe8cee6e47edbf95d8043ad989ad/src/Threebox.js#L131-L144
- https://github.com/mapbox/mapbox-gl-js/pull/7039#issuecomment-432110054

## Design Alternatives

I'm splitting this into two parts: a type for mercator coordinates and the methods used to convert.

### Coordinate type

```js
//1A: a three component array
typedef MercatorCoordinate = Array<number>;
[x, y, z]

// 1B: a class
class MercatorCoordinate {
    constructor(x, y, z) {
        this.x = x;
        this.y = y;
        this.z = z;
    }
}
```

### projection method signature

Mercator coordinates are three dimensional. Should we support projecting and unprojecting features with altitude. How?

```js
// 2A: using existing LngLat type
// downside: unprojecting and getting the altitude requires a separate method
projectMercator(LngLat: number, altitudeMeters?: number) : MercatorCoordinate
unprojectMercator(coord: MercatorCoordinate) : LngLat
// and maybe:
unprojectMercatorAltitude(coord: MercatorCoordinate) : number

// 2B: add optional altitude to LngLat type
LngLat(lng: number, lat: number, altitude?: number)
projectMercator(lngLatAlt: LngLat) : MercatorCoordinate
unprojectMercator(coord: MercatorCoordinate) : LngLatAlt

// 2C: like 2B except add a new LngLatAlt type
LngLatAlt(lng: number, lat: number, altitude?: number)
projectMercator(lngLatAlt: LngLat | LngLatAlt) : MercatorCoordinate
unprojectMercator(coord: MercatorCoordinate) : LngLatAlt
```

####
```js

// 3A on map class (requires an instantiated map)
map.projectMercator(...);
map.unprojectMercator(...);

// 3B without instantiated map
mapboxgl.projectMercator(...);
mapboxgl.unprojectMercator(...)

// 3C without instantiated map but namespaced
mapboxgl.mercator.project(...);
mapboxgl.mercator.unproject(...);

// 3D a mercator class
const mercator = new mapboxgl.MercatorProjection(optionalWorldSize);
mercator.project(...);
mercator.unproject(...);

// 3E like 3D but a bit more and with a helper for constructing the appropriate matrix for an area. Broad sketch:
const origin = LngLat({ lng: -79, lat: 42 });
const mercator new mapboxgl.MercatorProjection(origin)
mercator.project(...);
mercator.unproject(...);
map.getViewProjectionMatrix(mercator)

// 3F static method on MercatorCoordinate type
MercatorCoordinate.project(...)
MercatorCoordinate.unproject(...)
// alternative named `.fromLngLat(...)` and `toLngLat(...)`

// 3G static method for projection, regular method for unprojection
const coord = MercatorCoordinate.project(...);
const lngLat = coord.uproject();

// 3H constructor on MercatorCoordinate type
// constructor(xOrLngLat: LngLat | number, yOrAltitude?: number, z?: number)
const coord = new MercatorCoordinate(lngLat, altitude);
const lngLat = coord.unproject();
```

## Design

I'm not sure which is better. @mourner I need your design expertise.

I'm slightly leaning towards 1B over 1A because it makes the typing specific in user code but I could definitely be swayed towards 1A if someone has a strong preference for that.

The design adding `altitude` to `LngLat` but that might be a big change that we want to spend more time considering. For example, is the altitude the altitude above the surface? or above the wgs84 spheroid? how does this work for flat maps? how would this work if we added 3d terrain? `2A` is nice because it sidesteps these questions but it is not as nice.

For part 3, I think it's good to be able to use this without an instantiated map. I think I like `3G` because it provides separation and keeps the projection close to the type. I like `3D` because it namespaces the methods, allows for future expansion to 3E-like things. I like 3A and 3B because they are simple.

### Concepts

Documentation for the mercator type will explain the coordinate space. Documentation for custom layers will reference these methods. Custom layer examples will be switched to use this projection.

### Implementation

Factoring `worldSize` out of [here](https://github.com/mapbox/mapbox-gl-js/blob/ab7d06a3fff9c67564cf49eb60b83e84bda49121/src/geo/transform.js#L301-L320) and exposing it. And adding tests and documentation.

ansis:
This adds a `MercatorCoordinate` to expose a way to convert from from LngLat to the mercator coordinates used by custom layers.

fixes #7485

```js
class MercatorCoordinate {
    x: number;
    y: number;
    z: number;

    constructor(x: number, y: number, z?: number) { ... }
    static fromLngLat(lngLatLike: LngLatLike, altitude?: number) { ... }
    toLngLat(){ ... }
    toAltitude() { ... }
}
```

## Launch Checklist

# TODO

- [ ] switch `Transform` to use `MercatorCoordinate` for projection

### DIFF
--- a/src/geo/mercator_coordinate.js
+++ b/src/geo/mercator_coordinate.js
@@ -1,0 +1,88 @@
+// @flow
+
+import LngLat from '../geo/lng_lat';
+import type {LngLatLike} from '../geo/lng_lat';
+
+/*
+ * The circumference of the world in meters at the given latitude.
+ */
+function circumferenceAtLatitude(latitude: number) {
+    const circumference = 2 * Math.PI * 6378137;
+    return circumference * Math.cos(latitude * Math.PI / 180);
+}
+
+/**
+ * A `MercatorCoordinate` object represents a 3 dimensional position in projected web mercator coordinates.
+ *
+ * The "world size" used by `MercatorCoordinate` is 1, meaning `MercatorCoordinate(0, 0, 0)` is the north-west
+ * corner of the mercator world and `MercatorCoordinate(1, 1, 0)` is the south-east corner of the mercator world.
+ *
+ * @param {number} x The x component of the position.
+ * @param {number} y The y component of the position.
+ * @param {number} z The z component of the position.
+ * @example
+ * var nullIsland = new mapboxgl.MercatorCoordinate(0.5, 0.5, 0);
+ *
+ * @see [Add a custom style layer](https://www.mapbox.com/mapbox-gl-js/example/custom-style-layer/)
+ */
+class MercatorCoordinate {
+    x: number;
+    y: number;
+    z: number;
+
+    constructor(x: number, y: number, z?: number) {
+        this.x = +x;
+        this.y = +y;
+        this.z = z === undefined ? 0 : +z;
+    }
+
+    /**
+     * Project a `LngLat` to a `MercatorCoordinate`.
+     *
+     * @param {LngLatLike} lngLatLike The location to project.
+     * @param {number} altitude The altitude in meters of the position.
+     * @returns {MercatorCoordinate} The projected mercator coordinate.
+     * @example
+     * var coord = mapboxgl.MercatorCoordinate.fromLngLat({ lng: 0, lat: 0}, 0);
+     * coord; // MercatorCoordinate(0.5, 0.5, 0)
+     */
+    static fromLngLat(lngLatLike: LngLatLike, altitude?: number) {
+        const lngLat = LngLat.convert(lngLatLike);
+
+        const x = (180 + lngLat.lng) / 360;
+        const y = (180 - (180 / Math.PI * Math.log(Math.tan(Math.PI / 4 + lngLat.lat * Math.PI / 360)))) / 360;
+        const z = altitude === undefined ? 0 : (altitude / circumferenceAtLatitude(lngLat.lat));
+        return new MercatorCoordinate(x, y, z);
+
+    }
+
+    /**
+     * Returns the `LatLng` for the coordinate.
+     *
+     * @returns {LngLat} The `LngLat` object.
+     * @example
+     * var coord = new mapboxglMercatorCoordinate(0.5, 0.5, 0);
+     * var latLng = coord.toLatLng(); // LngLat(0, 0)
+     */
+    toLngLat(){
+        const lng = this.x * 360 - 180;
+        const y2 = 180 - this.y * 360;
+        const lat = 360 / Math.PI * Math.atan(Math.exp(y2 * Math.PI / 180)) - 90;
+        return new LngLat(lng, lat);
+    }
+
+    /**
+     * Returns the altitude in meters of the coordinate.
+     *
+     * @returns {number} The altitude in meters.
+     * @example
+     * var coord = new mapboxgl.MercatorCoordinate(0, 0, 0.02);
+     * coord.toAltitude(); // 6914.281956295339
+     */
+    toAltitude() {
+        const lat = this.toLngLat().lat;
+        return this.z * circumferenceAtLatitude(lat);
+    }
+}
+
+export default MercatorCoordinate;
--- a/docs/documentation.yml
+++ b/docs/documentation.yml
@@ -19,6 +19,7 @@ toc:
   - LngLatBoundsLike
   - Point
   - PointLike
+  - MercatorCoordinate
   - name: User Interface
     description: |
       Controls, markers, and popups add new user interface elements to the map.
--- a/docs/pages/example/custom-style-layer.html
+++ b/docs/pages/example/custom-style-layer.html
@@ -2,8 +2,8 @@
 <script>
 var map = window.map = new mapboxgl.Map({
     container: 'map',
-    zoom: 4,
-    center: [0, 0],
+    zoom: 3,
+    center: [7.5, 58],
     style: 'mapbox://styles/mapbox/light-v9'
 });
 
@@ -17,7 +17,7 @@ var highlightLayer = {
         "uniform mat4 u_matrix;" +
         "attribute vec2 a_pos;" +
         "void main() {" +
-        "    gl_Position = vec4(a_pos, 0.0, 1.0);" +
+        "    gl_Position = u_matrix * vec4(a_pos, 0.0, 1.0);" +
         "}";
 
         var fragmentSource = "" +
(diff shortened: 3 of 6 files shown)


src/geo/mercator_coordinate.js
```diff
@@ -0,0 +1,88 @@
...
+function circumferenceAtLatitude(latitude: number) {
+    const circumference = 2 * Math.PI * 6378137;
+    return circumference * Math.cos(latitude * Math.PI / 180);
+}
+
+/**
+ * A `MercatorCoordinate` object represents a 3 dimensional position in projected web mercator coordinates.
+ *
+ * The "world size" used by `MercatorCoordinate` is 1, meaning `MercatorCoordinate(0, 0, 0)` is the north-west
+ * corner of the mercator world and `MercatorCoordinate(1, 1, 0)` is the south-east corner of the mercator world.
```
andrewharvey: This documentation provides great clarity, I really appreciate it.

A lot of geospatial people might be confused by this definition though since the [web mercator projection](https://epsg.io/3857) is in meters and has a projected bounds of

```
-20026376.39 -20048966.10
20026376.39 20048966.10
```

Since `MertactorCoordinate` here is a a unit vector, I'm wondering if we need to be more explict that this is not ESPG:3857 web mercator :thinking: ?

ansis: @andrewharvey thats for flagging this! I'll have to think more closely whether we should match that or if we should just clarify the documentation.

ansis:
The main thing that would be useful to have reviewed so far are the first couple commits.

The last commit (8856fbf) is a big one and it replaces the previous internal `Coordinate` type with `MercatorCoordinate`. This isn't related to the main goal of exposing a coordinate type. When I started using `MercatorCoordinate` in `Transform` I realized that it would work better than `Coordinate` in most places we are using it. The commit is still a bit broken. I can split this into a different pr if that helps.

src/geo/mercator_coordinate.js
ansis: I clarified this a bit in 83cda37. Do you think this is clear enough?

*Also, I think the projected bounds on that website are wrong. I think it should be `20037508`.*

> I like the simplicity of a unit vector as the underlying representation compared to EPSG:3857, however I very much see it as an internal implementation detail.

Yep, it makes conversion from vector tiles to these units and back simpler. It also allows vector tile positions to be stored without loss of precision. I think it makes sense to keep it as is.

Do you think renaming it would help? `UnitMercatorCoordinate` to emphasize the unitness of it? or `TiledMercatorCoordinate` to nod in the direction of the tiled coordinate spaces this borrows from? I think I'm leaning towards just `MercatorCoordinate`

andrewharvey: I like the new documentation in https://github.com/mapbox/mapbox-gl-js/commit/83cda371cd4369f5caab946823bb4aff67a56f4e :+1: 

-1 to `TiledMercatorCoordinate`, that makes me think xyz tiles which vary by zoom level, but in fact this `MercatorCoordinate` doesn't vary my zoom level and isn't broken into tiles.

I like `UnitMercatorCoordinate` since it's more explicit that these are unit coordinates (0 to 1),

src/geo/mercator_coordinate.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+class MercatorCoordinate {
+    x: number;
+    y: number;
+    z: number;
+
+    constructor(x: number, y: number, z?: number) {
+        this.x = +x;
+        this.y = +y;
+        this.z = z === undefined ? 0 : +z;
```
mourner: Nit: we can use ES default parameters for things like this (below too for altitude).

src/geo/transform.js
```diff
@@ -278,50 +279,26 @@ class Transform {
     scaleZoom(scale: number) { return Math.log(scale) / Math.LN2; }
 
     project(lnglat: LngLat) {
-        return new Point(
-            this.lngX(lnglat.lng),
-            this.latY(lnglat.lat));
+        const lat = clamp(lnglat.lat, -this.maxValidLatitude, this.maxValidLatitude);
+        const clamped = new LngLat(lnglat.lng, lat);
+        const coord = MercatorCoordinate.fromLngLat(clamped);
+        return new Point(coord.x * this.worldSize, coord.y * this.worldSize);
```
mourner: My only concern with this change is that we significantly increase the number of allocations in potentially hot code, like here — instead of a single `Point` allocation, we do 3 — `Point`, `MercatorCoordinate`, `Point` again.

src/geo/transform.js
```diff
@@ -492,41 +459,43 @@ class Transform {
...
+            maxY = this.project(new LngLat(0, latRange[0])).y;
             sy = maxY - minY < size.y ? size.y / (maxY - minY) : 0;
         }
 
         if (this.lngRange) {
             const lngRange = this.lngRange;
-            minX = this.lngX(lngRange[0]);
-            maxX = this.lngX(lngRange[1]);
+            minX = this.project(new LngLat(lngRange[0], 0)).x;
+            maxX = this.project(new LngLat(lngRange[1], 0)).x;
```
mourner: To add to my point above, here, instead of zero allocations, we do 4 * (1 + 3) = 16, along with double-projection.

src/geo/transform.js
ansis: Do you think this is significant enough to warrant splitting out the projection into `lngToMercatorX(...)` and `latToMercatorY(...)` for internal use?

ansis: Does 0d4c2ca look good?

src/geo/mercator_coordinate.js
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * `MercatorCoordinate` uses the web mercator projection ([EPSG:3857](https://epsg.io/3857)) with slightly different units:
+ * - the size of 1 unit is the width of the projected world instead of the "mercator meter"
+ * - the origin of the coordinate space is at the north-west corner instead of the middle
+ *
+ * For example, `MercatorCoordinate(0, 0, 0)` is the north-west corner of the mercator world and
+ * `MercatorCoordinate(1, 1, 0)` is the south-east corner. If you are familiar with
+ * [vector tiles](https://github.com/mapbox/vector-tile-spec) it may be helpful to think
+ * of the coordinate space as the `0/0/0` tile with an extent of `1`.
+ *
```
asheemmamoowala: Does there need to be a caveat about precision here? I.e. When using a mercator co-ordinate to place something on a custom layer, it may not match the exact location at higher zooms.

src/geo/transform.js
```diff
@@ -203,12 +203,12 @@ class Transform {
     getVisibleUnwrappedCoordinates(tileID: CanonicalTileID) {
         const result = [new UnwrappedTileID(0, tileID)];
         if (this._renderWorldCopies) {
-            const utl = this.pointCoordinate(new Point(0, 0), 0);
-            const utr = this.pointCoordinate(new Point(this.width, 0), 0);
-            const ubl = this.pointCoordinate(new Point(this.width, this.height), 0);
-            const ubr = this.pointCoordinate(new Point(0, this.height), 0);
-            const w0 = Math.floor(Math.min(utl.column, utr.column, ubl.column, ubr.column));
-            const w1 = Math.floor(Math.max(utl.column, utr.column, ubl.column, ubr.column));
+            const utl = this.pointCoordinate(new Point(0, 0));
```
asheemmamoowala: Does `pointCoordinate` need to use `Point` for input? It could avoid the temporary object creation.

PR #7493, opened 2018-10-27
Add ability to use custom shaders for circle and line type layers + add timestamp uniform to fragment shaders 

kaligrafy:
## Launch Checklist

 - [ ] post benchmark scores


ryanhamley:
`performance.now()` is not available in our Node test suite. You'll have to stub it out for the query and render tests to work properly. JSDOM implements the performance API so it should be possible to do. I think you'd want to implement it in the query and render files here https://github.com/mapbox/mapbox-gl-js/tree/master/test/integration/lib

PR #7497, opened 2018-10-29
Limit parallel image requests for raster performance

### LINKED ISSUE #7460: Limit the number of concurrent image loads
A part of addressing #7405. Loading many tiles at once in the browser leads to a lot of stutter due image decoding and texture uploads on the main thread. If we limit the number of parallel image loads, this improves performance significantly, as seen in [this discussion](https://github.com/mapbox/mapbox-gl-js/pull/7008#issuecomment-423928601). This is also what some other mapping engines such as OpenLayers do.

There's a proof of concept implementation here https://github.com/mapbox/mapbox-gl-js/commit/cbb059fd1ce8847d47719db3f45d0e3cdcb0edcd, but it needs to be cleaned up / prepared for a proper PR.

mourner:
Fixes #7460, improving raster sources performance on big screens by setting a global limit on how many image requests can happen in parallel, and queueing all the rest.

Similarly to `workerCount`, the PR exposes the limit as `mapboxgl.maxParallelImageRequests` which you can change (`16` by default).

## Launch Checklist

 - [ ] ~~document any changes to public APIs~~ (probably not worth publicly documenting?)
 - [ ] ~~post benchmark scores~~ benchmarks don't exercise big-screen tile loading

### DIFF
--- a/src/util/config.js
+++ b/src/util/config.js
@@ -4,7 +4,8 @@ type Config = {|
   API_URL: string,
   EVENTS_URL: string,
   REQUIRE_ACCESS_TOKEN: boolean,
-  ACCESS_TOKEN: ?string
+  ACCESS_TOKEN: ?string,
+  MAX_PARALLEL_IMAGE_REQUESTS: number
 |};
 
 const config: Config = {
@@ -17,7 +18,8 @@ const config: Config = {
         }
     },
     REQUIRE_ACCESS_TOKEN: true,
-    ACCESS_TOKEN: null
+    ACCESS_TOKEN: null,
+    MAX_PARALLEL_IMAGE_REQUESTS: 16
 };
 
 export default config;
--- a/src/index.js
+++ b/src/index.js
@@ -48,7 +48,7 @@ const exported = {
      * mapboxgl.accessToken = myAccessToken;
      * @see [Display a map](https://www.mapbox.com/mapbox-gl-js/examples/)
      */
-    get accessToken() {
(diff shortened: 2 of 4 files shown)


src/util/config.js
```diff
@@ -17,7 +18,8 @@ const config: Config = {
         }
     },
     REQUIRE_ACCESS_TOKEN: true,
-    ACCESS_TOKEN: null
+    ACCESS_TOKEN: null,
+    MAX_PARALLEL_IMAGE_REQUESTS: 16
```
kkaefer: Why default to 16?

mourner: I had to pick a number anyway, and that's what [OpenLayers defaults to](https://github.com/openlayers/openlayers/blob/1035ba7bf2285ca5842505b5bdf5ec28ab62b64f/src/ol/source/Raster.js#L345) — it seemed to work well in practice, so I left it at that. We can adjust the default if you think 16 isn't a good value.

hyperknot:
So this one is a properly implemented https://github.com/mapbox/mapbox-gl-js/commit/cbb059fd1ce8847d47719db3f45d0e3cdcb0edcd? And you are saying that the XHR overhead shouldn't make a difference after this?

hyperknot:
Ah, so this actually shipped with 0.51 but was missed out from the Changelog? 
I was checking the changelog and I didn't see anything related to rasters! So this is great, actually it's the biggest performance update for raster layers in a very long time! Congrats.

PR #7516, opened 2018-10-30
Stub local and session storage in JSDOM

ryanhamley:
## Launch Checklist

This replaces https://github.com/mapbox/mapbox-gl-js/pull/7455

JSDOM 11.12 introduced the `localStorage` API. Our codebase copies properties from one JSDOM instance to another which worked fine but when accessing `localStorage` in a Node context where the JSDOM's base URL is not set, an error is thrown. `localStorage` is not used in our tests (we stub it out in a couple of places) so I think this is probably an acceptable workaround. Setting the JSDOM's base URL can cause issues with some tests and creates potential CORS errors that makes testing a pain.

Tip of the 🎩 to @fc for reporting the original bug and helping figure out what exactly was going wrong.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "gl": "^4.0.1",
     "glob": "^7.0.3",
     "is-builtin-module": "^1.0.0",
-    "jsdom": "11.11.0",
+    "jsdom": "^11.11.0",
     "json-stringify-pretty-compact": "^1.0.4",
     "jsonwebtoken": "^8.3.0",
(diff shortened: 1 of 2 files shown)


fc:
Would it be desirable to upgrade JSDom along with this?

ryanhamley:
I've updated `package.json` so JSDOM can freely upgrade and stubbed out `[local/session]Storage` with a warning log in case anyone attempts to access either property.

PR #7518, opened 2018-10-30
Fix #7517 - asymmetrical cameraForBounds

### LINKED ISSUE #7517: cameraForBounds returns incorrect CameraOptions with asymmetrical padding/offset
**mapbox-gl-js version**: v0.50.0

**browser**: all

### Steps to Trigger Behavior

 1. Use [`cameraForBounds`](https://www.mapbox.com/mapbox-gl-js/api#map#cameraforbounds) to get a [`CameraOptions`](https://www.mapbox.com/mapbox-gl-js/api#cameraoptions) object, using asymmetrical `padding` and/or `offset`.
 2. Use the center, zoom, and bearing to move the map.
 3. The map is in the wrong place.

### Link to Demonstration
https://jsfiddle.net/mike_marcacci/Lnh9fx6z

### Expected Behavior
image

### Actual Behavior
image

mike-marcacci:
The `_cameraForBoxAndBearing` method had several shortcomings that caused `cameraForBounds` to return incorrect values [when `padding` or `offset` are asymmetrical](https://github.com/mapbox/mapbox-gl-js/issues/7517). Additionally, because the documented return value is a [`CameraOptions`](https://www.mapbox.com/mapbox-gl-js/api/#cameraoptions) object, we can't rely on the `offset` property which was leaked through in the previous implementation.

## Launch Checklist
 - [ ] post benchmark scores

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -434,17 +434,9 @@ class Camera extends Evented {
             return;
         }
 
-        // we separate the passed padding option into two parts, the part that does not affect the map's center
-        // (lateral and vertical padding), and the part that does (paddingOffset). We add the padding offset
-        // to the options `offset` object where it can alter the map's center in the subsequent calls to
-        // `easeTo` and `flyTo`.
-        const paddingOffset = [(options.padding.left - options.padding.right) / 2, (options.padding.top - options.padding.bottom) / 2],
-            lateralPadding = Math.min(options.padding.right, options.padding.left),
-            verticalPadding = Math.min(options.padding.top, options.padding.bottom);
-        options.offset = [options.offset[0] + paddingOffset[0], options.offset[1] + paddingOffset[1]];
-
         const tr = this.transform;
-        // we want to calculate the upper right and lower left of the box defined by p0 and p1
+
+        // We want to calculate the upper right and lower left of the box defined by p0 and p1
         // in a coordinate system rotate to match the destination bearing.
         const p0world = tr.project(LngLat.convert(p0));
         const p1world = tr.project(LngLat.convert(p1));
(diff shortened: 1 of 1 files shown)


mike-marcacci:
The existing test was incorrect given my understanding of how the method should work based on documentation. I fixed the test and added more to explicitly test these scenarios.

mike-marcacci:
I added a debug file to test the behavior visually. I also used it as a fiddle:

- [before (broken)](https://jsfiddle.net/mike_marcacci/Lnh9fx6z)
- [after (fixed)](https://jsfiddle.net/mike_marcacci/10zwb67n/)

mike-marcacci:
Linter now passes. Ready for review 👍

src/ui/camera.js
```diff
@@ -454,22 +446,32 @@ class Camera extends Evented {
...
-        options.center =  tr.unproject(p0world.add(p1world).div(2));
-        options.zoom = Math.min(tr.scaleZoom(tr.scale * Math.min(scaleX, scaleY)), options.maxZoom);
-        options.bearing = bearing;
 
-        return options;
+        const zoom = Math.min(tr.scaleZoom(tr.scale * Math.min(scaleX, scaleY)), options.maxZoom);
+
+        // Calculate center: apply the zoom, the configured offset, as well as offset that exists as a result of padding.
+        const paddingOffset = [(options.padding.left - options.padding.right) / 2, (options.padding.top - options.padding.bottom) / 2];
+        const offsetAtInitialZoom = Point.convert([options.offset[0] + paddingOffset[0], options.offset[1] + paddingOffset[1]]);
```
mourner: Nitpick: let's avoid temporary arrays here by using `new Point(x, y)` instead of `Point.convert([x, y])`, and maybe declaring separate `paddingOffsetX` and `paddingOffsetY` variables since we don't really need it being in an array.

mourner: Also, `options.offset` is of `PointLike` type so this code will break if you pass e.g. `offset: new Point(10, 10)` instead of `offset: [10, 10]`. So let's bring back the `const offset = Point.convert(options.offset)` line and use that.

mike-marcacci: Great points! It looks like the previous implementation was also accessing array indices directly, so I added a test to make sure `PointLike` objects work.

PR #7528, opened 2018-10-31
make yarn test-unit command more cross-platform

booo:
The first variable is the script name. Don't pass the script name as
argument. If we pass the script name as argument the command hangs with:

  Reading TAP data from stdin (use "-" argument to suppress)

Don't pass the fourth argument ("test/unit") to the tap script.
Otherwise we execute all test scripts in "test/unit".

The shell sets BASH_ARGV only when in extended debugging mode. Use the
argument indices to access them. This increases readability too.

## Launch Checklist

What about test-build? It uses run-tap too. Calling test-build fails in my environment. I guess I have trigger an actual build first.

### DIFF
--- a/build/run-tap
+++ b/build/run-tap
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 if [ "$#" == 5 ]; then
-  arg="${@:0:4} ${BASH_ARGV[1]}/${BASH_ARGV[0]}"
+  arg="${@:1:3} ${4}/${5}"
 else
   arg="${@}"
 fi


booo:
If I run `yarn test-unit ui/map.test.js on my system the script tries to interpret build/run-tap as test and hangs "forever" after a timeout and the execution of ui/map.test.js:

```
➜  mapbox-gl-js git:(master) yarn test-unit ui/map.test.js
yarn run v1.12.1
$ build/run-tap --reporter classic --no-coverage test/unit ui/map.test.js
Reading TAP data from stdin (use "-" argument to suppress)
build/run-tap ......................................... 0/1 31s
  not ok timeout!
    expired: build/run-tap

test/unit/ui/map.test.js .......................... 181/181
total ............................................. 181/182
  

  181 passing (41s)
  1 failing
```

Sorry for the bad problem description. Can I provide you with additional information?

ryanhamley:
What OS and shell are you using? I can't reproduce this on my Mac with ZSH or Bash. The only reference I can find to what you're seeing is this issue https://github.com/tapjs/node-tap/issues/349 It says the warning you see results from not being able to find tests, but clearly it's finding and running the tests.

booo:
My bash version is *4.4.23(1)-release (x86_64-unknown-linux-gnu)*. My zsh version is *zsh 5.6.2 (x86_64-pc-linux-gnu)*. I distribution is ArchLinux.

I tested this with bash and zsh. No matter what shell I use I have problems with the argument handling of the script.

ryanhamley:
Weird. I'm guessing this is some ArchLinux specific bug but honestly 🤷‍♂️ . Regardless, your changes work on my computer so I'm inclined to say that if this is more cross-platform, then great. What do you think @mollymerp?

mollymerp:
👌 works for me!

PR #7539, opened 2018-11-01
Prevent calls to getLayoutProperty from causing fatal error when layer doesn't exist

### LINKED ISSUE #7537: Calling map.getLayoutProperty will throw an undefined error when layer doesn't exist
## Motivation
Calling map.getLayoutProperty should not throw an undefined error when layer name doesn't exist. 

## Design Alternatives
It should return null.

## Design
I thought I was having issues with my map object because an undefined error was being thrown.

### Implementation
javascript

I was using version 0.50.0

ryanhamley:
Fixes #7537 
## Launch Checklist

     - When calling `map.getLayoutProperty()` for a layer that does not exist, a fatal error would fire. `setLayoutProperty()` fires a warning error to handle this situation. This PR updates `getLayoutProperty` to do the same thing.
    - Also updates the `layer` argument to `layerId` to match `setLayoutProperty`


ryanhamley:
I pulled the linter fixes out to https://github.com/mapbox/mapbox-gl-js/pull/7541

Issue #7544, opened 2018-11-02
Allow FullscreenControl to specify a source


carrieVant:
## Motivation

There is currently no easy way to display additional features other than just the map in FullscreenControl.  We display buttons below our map that allow users to toggle between different maps (ex: street, satellite) and add additional layers to maps that display/filter pins.  We would like these to appear in Fullscreen view also.

## Design Alternatives

Add a source property to FullscreenControl to specify which DOM element you would like to display in FullscreenControl.  This would allow you to display an element that includes other functionality in FullscreenControl (similar to what OpenLayers.js currently supports).

## Design

map.addControl(new mapboxgl.FullscreenControl({source: 'map-container'}));

### Implementation

map.addControl(new mapboxgl.FullscreenControl({source: 'map-container'}));

PR #7548, opened 2018-11-02
Allow user to define element other than map container for full screen control

ryanhamley:
fixes https://github.com/mapbox/mapbox-gl-js/issues/7544

## Launch Checklist

The intended use case is for displaying a map with an overlaid element that works with the map, such as radio buttons to toggle styles or a banner element with a legend. The current behavior of making the map container element full screen will not show these elements. I can't really think of any particular reason to disallow this behavior but let me know if there's a reason not to implement this.

Full screen with `{source: body}`
screen shot 2018-11-02 at 12 40 43 pm

Default behavior (notice that the debug control in the upper left corner is not displayed)
screen shot 2018-11-02 at 12 42 00 pm

### DIFF
--- a/src/ui/control/fullscreen_control.js
+++ b/src/ui/control/fullscreen_control.js
@@ -7,26 +7,36 @@ import window from '../../util/window';
 
 import type Map from '../map';
 
+type Options = {
+    source?: HTMLElement
+};
+
 /**
  * A `FullscreenControl` control contains a button for toggling the map in and out of fullscreen mode.
  *
  * @implements {IControl}
+ * @param {Object} [options]
+ * @param {HTMLElement} [options.source] `source` is the [compatible DOM element](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen#Compatible_elements) which should be made full screen. By default, the map container element will be made full screen.
+ *
  * @example
- * map.addControl(new mapboxgl.FullscreenControl());
+ * map.addControl(new mapboxgl.FullscreenControl({source: document.querySelector('body')}));
  * @see [View a fullscreen map](https://www.mapbox.com/mapbox-gl-js/example/fullscreen/)
  */
 
 class FullscreenControl {
     _map: Map;
-    _mapContainer: HTMLElement;
     _container: HTMLElement;
     _fullscreen: boolean;
     _fullscreenchange: string;
     _fullscreenButton: HTMLElement;
     _className: string;
+    _source: HTMLElement;
 
-    constructor() {
+    constructor(options: Options) {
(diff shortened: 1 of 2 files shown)


src/ui/control/fullscreen_control.js
```diff
@@ -7,26 +7,36 @@ import window from '../../util/window';
...
     _fullscreenchange: string;
     _fullscreenButton: HTMLElement;
     _className: string;
+    _source: HTMLElement;
 
-    constructor() {
+    constructor(options: Options) {
         this._fullscreen = false;
+        if (options && options.source) {
+            this._source = window.document.querySelector(options.source);
```
mourner: Our other APIs that need an HTML element reference, such as Map `container` and Marker `element`, accept an `HTMLElement` rather than a selector — I'd suggest keeping this consistent.

src/ui/control/fullscreen_control.js
```diff
@@ -45,7 +59,7 @@ class FullscreenControl {
 
     onAdd(map: Map) {
         this._map = map;
-        this._mapContainer = this._map.getContainer();
+        if (!this._source) this._source = this._map.getContainer();
```
mollymerp: I think `container`/`mapContainer` is a clearer word for this member / option

ryanhamley:
Added a unit test and changed `source` option to `container` and the previous `_container` internal property to `_controlContainer` to disambiguate them.

test/unit/ui/control/fullscreen.test.js
```diff
@@ -27,3 +27,18 @@ test('FullscreenControl does not appears when fullscreen is not enabled', (t) =>
...
+test('FullscreenControl makes optional container element full screen', (t) => {
+    window.document.fullscreenEnabled = true;
+
+    const map = createMap(t);
+    const fullscreen = new FullscreenControl({container: window.document.querySelector('body')});
+    map.addControl(fullscreen);
+    const control = map._controls.find((ctrl) => {
+        return ctrl.hasOwnProperty('_fullscreen');
+    });
+    control._onClickFullscreen();
```
mollymerp: maybe set a spy on `body.requestFullscreen` and then check `t.calledOnce` after this?

ryanhamley: I might be missing something but it doesn't seem like JSDOM supports the Full Screen API. I don't see any documentation or code that mentions it and `window.document.requestFullScreen` and `body.requestFullScreen` are both `undefined` (along with their prefixed alternatives).

mollymerp: ahh ok nvm then!

PR #7550, opened 2018-11-03
Avoid unnecessary image decode.

dcervelli:
`image_source.js` unnecessarily decodes image data via JavaScript before creating a `Texture` from the image data.

The `Texture` class can handle a native `HTMLImageElement` directly and then the browser can decode the image much more quickly and with less memory.

For my use case, this results in about a 10-20% reduction in scripting time.

### DIFF
--- a/src/source/image_source.js
+++ b/src/source/image_source.js
@@ -79,7 +79,7 @@ class ImageSource extends Evented implements Source {
     dispatcher: Dispatcher;
     map: Map;
     texture: Texture | null;
-    image: ImageData;
+    image: HTMLImageElement;
     tileID: CanonicalTileID;
     _boundsArray: RasterBoundsArray;
     boundsBuffer: VertexBuffer;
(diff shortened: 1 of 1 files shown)


mourner:
It was originally like this but changed in this commit https://github.com/mapbox/mapbox-gl-js/commit/2d359e97d58c85bf87326c50b381d501f3ab29d0 — can you make sure this won't break the use cases that necessiated that change?

dcervelli:
I pored over that commit and a few other related ones to determine what the intent was behind that change but don't have a definitive answer.

It seems like the change to `image_source.js` is unrelated to the broader commit (changing the public API of `Map#addImage`).

A couple of threads I tried to track down:
* Was the `browser.getImageData` call added to correspond to the desired change to the type in `_prepareImage`? This doesn't make a ton of sense though since the original `HTMLImageElement` should have been compatible with `TexImageSource`.
* Was the `browser.getImageData` call needed because most of the guts of dealing with texture were in `image_source.js` at the time (as opposed to `texture.js`, where they are now)? Again, this isn't an adequate explanation because the `HTMLImageElement` appears to be a sufficient type for the texture methods.

Given that `image` is private, the types are all compatible, and the changed line appears to be unrelated to the commit message where it was changed, this seems like a safe change.

@jfirebaugh, any background you could add?

PR #7558, opened 2018-11-06
fix shaky sprites on scrollZoom

mollymerp:
Sprites wiggle around on scroll zoom because map.isZooming() always returns false due to the order of operations in the render loop, because it checks `ScrollZoomHandler.isActive()` which is set back to false by the time it is called here

https://github.com/mapbox/mapbox-gl-js/blob/61ae8d29f1c4377463c8f411bfb1f1ad80c52074/src/ui/map.js#L1634

and `painter.options.zooming` is used to choose the texture filter 
https://github.com/mapbox/mapbox-gl-js/blob/e682aa65da789df11fe7e76dbc56ada33e6e56ab/src/render/draw_symbol.js#L111-L112

I don't have many good ideas on how to test this 😬
## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/handler/scroll_zoom.js
+++ b/src/ui/handler/scroll_zoom.js
@@ -33,6 +33,7 @@ class ScrollZoomHandler {
     _el: HTMLElement;
     _enabled: boolean;
     _active: boolean;
+    _zooming: boolean;
     _aroundCenter: boolean;
     _around: Point;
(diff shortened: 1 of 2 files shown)


mollymerp:
for posterity: the shakyness is hard to see on retina screens, but very visible on lower-resolution devices.

PR #7561, opened 2018-11-06
Load embedded metadata from sprite PNG files

kkaefer:
This PR adds support for embedded sprite JSON manifests in PNG files. https://github.com/mapbox/spritezero/pull/59 adds support for embedding the metadata encoded as PBF in an ancillary chunk in the PNG file with the type `mbSM`. The goal is to remove the chance of the JSON and the PNG file going out of sync, which sometimes happens and is described in https://github.com/mapbox/mapbox-gl-native/issues/4117.

This PR implements support in a backwards-compatible way: We'll first load the sprite image, and if it is a PNG, we're looking for the `mbSM`, and parse it if present. Otherwise, we fall back to requesting the JSON manifest.

This could entail a slight degradation for legacy sprite files without an embedded manifest as the two requests for the image and the manifest file are now in serial rather than in parallel. However, it'll reduce the number of requests for new sprite images and eliminate the chance of going out of sync. Existing APIs are advised to continue to serve both the sprite image with the metadata embedded, as well as a separate manifest JSON for backwards compatibility.

## Launch Checklist

 - [ ] post benchmark scores


kkaefer:
I opted for using Protocol buffer for encoding rather than JSON for two reasons:
- The ancillary chunks are not compressed, and the JSON is pretty big. We can't easily compress the chunk because we'd manually have to decompress it in GL JS, or use `Content-Encoding`/`Transfer-Encoding` and recompress the PNG file
- Parsing JSON from an `ArrayBuffer` requires jumping through a few hoops, like converting the buffer to a string. This requires correct UTF-8 handling which is available in some browsers, but not IE. Our Pbf library already handles UTF-8 encoded strings internally.

src/style/load_sprite.js
```diff
@@ -11,57 +11,100 @@ import type {RequestTransformFunction} from '../ui/map';
...
+    return new Protobuf(data).readFields(function(tag: number, manifest: Object, pbf: Protobuf) {
+        if (tag === 1) {
+            let id;
+            const img = pbf.readMessage(function(tag: number, img: Object, pbf: Protobuf) {
+                if (tag === 1) id = pbf.readString();
+                else if (tag === 2) img.width = pbf.readVarint();
+                else if (tag === 3) img.height = pbf.readVarint();
+                else if (tag === 4) img.x = pbf.readVarint();
+                else if (tag === 5) img.y = pbf.readVarint();
+                else if (tag === 6) img.y = pbf.readFloat();
```
mourner: Typo: `img.pixelRatio`

src/util/ajax.js
```diff
@@ -178,7 +178,8 @@ const transparentPngUrl = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAA
 const imageQueue = [];
 let numImageRequests = 0;
 
-export const getImage = function(requestParameters: RequestParameters, callback: Callback<HTMLImageElement>): Cancelable {
+export const getImage = function(requestParameters: RequestParameters, callback: (error: ?Error, img: ?HTMLImageElement, data: ?ArrayBuffer) => void): Cancelable {
+    console.warn('getImage');
```
ryanhamley: Did you mean to leave a warning log here?

src/style/load_sprite.js
```diff
@@ -11,57 +11,100 @@ import type {RequestTransformFunction} from '../ui/map';
...
+        }
+        image = img;
 
-    let imageRequest = getImage(transformRequestCallback(normalizeSpriteURL(baseURL, format, '.png'), ResourceType.SpriteImage), (err, img) => {
-        imageRequest = null;
-        if (!error) {
-            error = err;
-            image = img;
-            maybeComplete();
+        console.warn(data);
```
ryanhamley: console.warn here too

src/style/load_sprite.js
```diff
@@ -11,57 +11,100 @@ import type {RequestTransformFunction} from '../ui/map';
...
+        // Try to extract the JSON manifest from the PNG image in a special 'mbSM' (Mapbox sprite manifest) chunk.
+        const png = new DataView(data);
+        if (png.getUint32(0) === 0x89504E47 && png.getUint32(4) === 0xD0A1A0A) {
+            for (let i = 8; i < png.byteLength;) {
+                var length = png.getUint32(i);
```
ryanhamley: `var` should be `let`

ansis:
Do you think we will eventually support including the images directly in the style? If yes, should we try to avoid adding this new way of specifying metadata within pngs that will be made redundant by that?

A temporary solution that doesn't involve any spec or sdk changes could be to refer to specific sprite versions within the style:

```json
"sprite": "mapbox://sprites/mapbox/bright-v8?version=123"
or
"sprite": "mapbox://sprites/mapbox/bright-v8?hash=abcdef"
```

There is also a related issue where the sprite is out of sync with the style. It's less critical because if the sprites are mostly the same there should be few visual bugs. Referring to sprite versions or embedding the sprite within the style would fix this issue as well.

PR #7574, opened 2018-11-09
Update style spec compatibility tables for Horchata

1ec5:
Updated the style specification’s compatibility tables for Android map SDK v6.7.0, iOS map SDK v4.5.0–v4.6.0, and macOS map SDK v0.12.0.

/cc @fabian-guerra @ChrisLoer

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -973,7 +973,8 @@
       "sdk-support": {
         "basic functionality": {
           "js": "0.49.0",
-          "ios": "4.5.0"
+          "ios": "4.5.0",
+          "macos": "0.12.0"
         },
         "data-driven styling": {}
       },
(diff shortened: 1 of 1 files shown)


src/style-spec/reference/v8.json
```diff
@@ -2566,7 +2567,10 @@
         "group": "Types",
         "sdk-support": {
           "basic functionality": {
-            "js": "0.48.0"
+            "js": "0.48.0",
+            "android": "6.7.0",
+            "ios": "4.6.0",
+            "macos": "0.12.0"
```
1ec5: The iOS and macOS do support the `format` operator in style JSON and at runtime via `MGL_FUNCTION` in NSExpression. mapbox/mapbox-gl-native#12739 tracks full support via an idiomatic NSExpression syntax.

As a general rule, we agreed in https://github.com/mapbox/mapbox-gl-js/pull/6988#discussion_r203505507 to leave out an entry when the SDK supports a property in style JSON but not fully at runtime. But Studio relies on this file for its compatibility options, so I’m coming around to the idea that we should go ahead and acknowledge support for a property even if it’s only supported in style JSON. Otherwise, Studio users could be prevented from using a property until it gets runtime styling support, which seems unfair.

/ref @tobrun

PR #7581, opened 2018-11-10
Fix Query point translation for multi-point geometry

### LINKED ISSUE #6833: queryRenderedFeatures bbox broken with circle-translate
**mapbox-gl-js version**: 0.45.0
**browser**: Chrome latest

### Steps to Trigger Behavior
 1. Have a circle layer with `circle-translate`
 2. Run `queryRenderedFeatures` with bbox on that layer

### Link to Demonstration
https://jsbin.com/viqafex
(based on official example "using-box-queryrenderedfeatures")

### Expected Behavior
All features should be selected

### Actual Behavior
Some features are randomly missing. If you move the map around, different features are missing.

screenflow

asheemmamoowala:
Fixes #6833 

This wasn't an issue with `circle-translate` but with how query geometry is translated when it contains more than a single point. 

The `Point#_mult` method modifies the point, which causes the query geometry to be multiplied by `pixelsToTileUnits` multiple times here:
https://github.com/mapbox/mapbox-gl-js/blob/c708474eb65d9c6a117fe232b677db19525f70b4/src/style/query_utils.js#L44


src/style/query_utils.js
```diff
@@ -29,8 +29,7 @@ export function translate(queryGeometry: Array<Array<Point>>,
     if (!translate[0] && !translate[1]) {
         return queryGeometry;
     }
-
-    const pt = Point.convert(translate);
+    const pt = Point.convert(translate);._mult(pixelsToTileUnits);
```
mourner: typo: `;` before `_mult`

Issue #7196, opened 2018-08-24
Single finger pan events overlaps with multi touch pan


pakastin:
There is no touch point count filtering in single finger pan. When you start multi touch pan there’s single and multi touch pan functions doing DOM updates simultaneously.

I noticed this when was fixing the multi touch center issue. Will debug more if there’s an easy fix. I’m not sure how the events should work in such case, so didn’t want to rush with a fix..

pakastin:
Here's a good description how it affects the UX: https://github.com/mapbox/mapbox-gl-js/issues/6722#issuecomment-421928604

jliebrand:
sub

PR #7582, opened 2018-11-11
begin touch zoom immediately when rotation disabled

msbarry:
## Launch Checklist

When touch rotation is disabled, there is a lag between when you start pinching to zoom in or out and when the zoom starts happening. This is because the touch zoom rotate handler has a zoom and rotate threshold to disambiguate between the two cases. This threshold is not needed if rotation is disabled. This makes the pinch to zoom behavior comparable with leaflet. There's still a bit of weirdness due do #7196 but it seems to be an improvement.


ryanhamley:
Thanks!

PR #7584, opened 2018-11-12
Introduce clusterProperties option for aggregated cluster properties

### LINKED ISSUE #2412: Support property aggregation on clustered features

mourner:
Closes #2412. An alternative to #7004.

Finally got my hands on exploring this feature, building on an extremely helpful implementation by @redbmk (sorry for not jumping in earlier!). Here I'm proposing a different API, which is very terse and easy to use, although it might look weird at first:

```json
"clusterProperties": {
  "max": ["max", 0, ["get", "scalerank"]],
  "sum": ["+", 0, ["get", "scalerank"]],
  "has_island": ["any", false, ["==", ["get", "featureclass"], "island"]]
}
```

The syntax is `property: [operator, initialExpression, mapExpression]`. The array is itself a valid expression, though using it for cluster properties adds additional limitations.

- Initial value is calculated by evaluating `initialExpression`
- Single point value is calculated with `mapExpression` (that can reference feature properties)
- Reduction is basically `accumulated = evaluate([operator, accumulated, value])`.

To implement this, I had to introduce an internal-use only `["accumulated"]` expression. An alternative would be to expose it in the syntax, making it more verbose but also maybe more self-descriptive:

```json
"max": ["max", ["accumulated", 0], ["get", "scalerank"]]
```

Regarding `Infinity` and `-Infinity`, a point raised in #7004 (it produces `null` because JSON format doesn't support `Infinity`), — I think it would make sense to introduce an `["infinity"]` expression, similar to how we have `["pi"]` and `["e"]`.

Regarding access to `["zoom"]` in the accumulating expression — I think this wouldn't be useful because of how map/reduce works: imagine that out of 10 points, 5 merge into one cluster on z16, and 5 remaining get merged into that cluster on z15. The accumulated value in such cluster would depend on multiple zoom values (depending on where it acquired new points), which doesn't semantically makes much sense.

Opening up the PR to start a discussion, but it still needs some work before a proper review:

## TODO

- [x] Proper Flow types
- [x] Proper error handling
- [x] Validation of cluster property expressions (specific to this use case)
- [x] Documentation

## Launch Checklist

 - [ ] ~~post benchmark scores~~ we don't benchmark clustering


sheerun:
Looks awesome, I hope you'll have time to push this feature through :)

behuda:
Maybe off topic, but can i change geojson clusterRadius property without creating new geojson source.
Wanted to change clusterRadius property through  a input slider.
But the problem old geojson src has to be destroyed and corresponding layer should be remove.
Only if the same geojson source could be updated retaining the source id. 
So that error: source cannot be removed since layers are using it can be avoided.

src/source/geojson_worker_source.js
```diff
@@ -293,4 +296,70 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+            const message = parsed.value.map(e => e.message).join('; ');
+            throw new Error(`Error parsing expression for cluster property "${key}": ${message}`);
+
+        } else if (!isGlobalPropertyConstant(parsed.value.expression, ['zoom'])) {
+            throw new Error(`Error parsing expression for cluster property "${key}": zoom expressions not supported.`);
+        }
+
+        const initialExpressionParsed = createExpression(initialExpression);
+        const mapExpressionParsed = createExpression(mapExpression);
+        const reduceExpressionParsed = createExpression([operator, ['accumulated'], ['get', key]]);
```
asheemmamoowala: Do you think the map/reduce expressions need to support the feature-state operator as well? Given that these expressions are evaluated just once it probably doesn't make sense to support that operator - it can change through interactivity on the map. An additional `isStateConstant` check would be required. 

If we want to support `feature-state`, it would require re-computing the map/reduce on every change of a feature's state, or at least the clusters affected by a feature. Not sure that there is a need for it right now.

src/style-spec/reference/v8.json
```diff
@@ -373,6 +373,10 @@
       "type": "number",
       "doc": "Max zoom on which to cluster points if clustering is enabled. Defaults to one zoom less than maxzoom (so that last zoom features are not clustered)."
     },
+    "clusterProperties": {
+      "type": "*",
+      "doc": "An object defining custom properties on the generated clusters if clustering is enabled, aggregating values from clustered points. Has the form `{\"property_name\": [operator, initial_expression, map_expression]}`. `operator` is any expression function that accepts at least 2 operands (e.g. `\"+\"` or `\"max\"`) — it accumulates the property value from clusters/points the cluster contains; `initial_expression` evaluates the initial value of the property before accummulating other points/clusters; `map_expression` produces the value of a single point.\n\nExample: `{\"sum\": [\"+\", 0, [\"get\", \"scalerank\"]]}`."
+    },
```
asheemmamoowala: Do you think it would be useful to add an expression support block for each of the expressions - initial,map, reduce ?Something similar to what is in the the layer properties.
https://github.com/mapbox/mapbox-gl-js/blob/1208cfc217ca90fa0a4d7fb1f2a1cc073d319b94/src/style-spec/reference/v8.json#L3500-L3508

mourner: I'm not sure how to separate those into separate properties. There are two expressions (map, initial) and an operator that accepts them as parameters. Map expression would be just `["feature"]`, and initial expression would be `[]`.

Given the non-standard syntax and that support block is dictated more by property aggregation semantics rather than technical limitations (like in case of many properties), perhaps we should leave that to validation, and maybe add a note clarifying this in the doc string.

src/source/geojson_worker_source.js
```diff
@@ -293,4 +296,70 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+
+    superclusterOptions.initial = () => {
+        const properties = {};
+        for (const key of propertyNames) {
+            properties[key] = initialValues[key];
+        }
+        return properties;
+    };
+    superclusterOptions.map = (pointProperties) => {
+        feature.properties = pointProperties;
```
redbmk: Why is this (and in `reduce below`) using a shared `feature` and overriding the `properties` of it instead of just defining a new `const feature = { properties: pointProperties }` here?

redbmk:
Thanks so much for taking this on! I'm looking forward to having native mapbox support for cluster agg! I left a couple comments regarding some of the code, but overall it looks great and I think this is easier to read than what I had come up with in #7004

src/source/geojson_worker_source.js
```diff
@@ -293,4 +296,70 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+        const properties = {};
+        for (const key of propertyNames) {
+            properties[key] = mapExpressions[key].evaluate(globals, feature);
+        }
+        return properties;
+    };
+    superclusterOptions.reduce = (accumulated, clusterProperties) => {
+        feature.properties = clusterProperties;
+        for (const key of propertyNames) {
+            globals.accumulated = accumulated[key];
```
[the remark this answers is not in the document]
mourner: Currently not, but I think this could be added (check if `operator` is string and if it's not, consider it as a reduce expression). Would add quite a bit of complexity to the docs, but add some flexibility as well. Not sure how to go here.

src/source/geojson_worker_source.js
mourner: The map function and expression evaluation is synchronous, meaning there are no side effects from using mutable state here (which is reset at the beginning). At the same time, we avoid constantly creating new feature objects, improving performance and lots of garbage collection passes.

redbmk: Ah, OK that makes sense. I thought it might have something to do with performance optimization.

src/source/geojson_worker_source.js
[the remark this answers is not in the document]
redbmk: Ah ok, I see - the reduce expression has to be something that fits into `[operator, ['accumulated'], ['get', key]]`. So that's why `max` and `any` work.

👍

mourner:
@asheemmamoowala moved the validation logic to style-spec and added the feature-state check — ready for another review.

@redbmk also added support custom expressions (referencing `["accumulated"]`) in place of `operator`. The example above won't work because of type mismatches but I guess you could still do the same with something more verbose.

docs/components/expression-metadata.js
```diff
@@ -199,6 +199,7 @@ for (const name in CompoundExpression.definitions) {
 }
 
 delete types['error'];
+delete types['accumulated']; // skip documenting `accumulated` since it is internal use only
```
asheemmamoowala: Should this still be deleted, given that it is now usable as part of a custom reduce expression?

mourner: Right, just readded it.

mourner:
I'm thinking that maybe I made a mistake of requiring an initial value, since for most cases, we could make it optional — see https://github.com/mapbox/supercluster/pull/114. If it's merged, we could simplify the minimal syntax to:

```js
"clusterProperties": {
  "max": ["max", ["get", "scalerank"]],
  "sum": ["+", ["get", "scalerank"]],
  "has_island": ["any", ["==", ["get", "featureclass"], "island"]]
}
```

And make the initial value the third optional argument. We could still make that change if we hurry to get this in before the next GL JS beta.

mourner:
Just realized that it would mean we can define or omit initial values on a per-property basis, which doesn't fit the implementation in Supercluster (which is all or nothing). So our options are to either leave things as is, or ditch initial values completely (which I'm increasingly leaning towards).

redbmk:
I'm inclined to leave `initial` as an optional param after the map expression. I shared an example in https://github.com/mapbox/supercluster/pull/114#issuecomment-454613799 of using an `initial` value that wouldn't work by just using the first mapped value. However, I think in _most_ cases you wouldn't need `initial`, so it makes sense to have it be optional.

behuda:
```
"clusterProperties": { 
    "my_point_count" : [["+", ["accumulated"], 1], 1]
}
```
when i'm comparing it with point_count the my_point_count property seems to different on some cluster

Also, 
how to calculate average from a field lets say salary.

src/source/geojson_worker_source.js
willviles: @asheemmamoowala @mourner I'm just wondering if you're looking to support `feature-state` on `clusterProperties` in the near future?

I've got a use case where I need to highlight the marker cluster for a given feature when a reference of it is selected in a list view outside the map canvas.

This works nicely for non-clustered points, but I need a work around for this situation:

```javascript
map.addSource('foos', {
  type: 'geojson',
  data,
  cluster: true,
  clusterMaxZoom: 14,
  clusterRadius: 50,
  clusterProperties: {
    'has_hovered_foo': ['==', ['feature-state', 'hover'], true]
  }
})

map.addLayer({
  id: 'clusters',
  type: 'circle',
  source: 'foos',
  filter: ['has', 'point_count'],
  paint: {
    'circle-color': [
      'case',
      ['get', 'has_hovered_foo'],
      '#000',
      '#FFF'
    ]
  }
})
```

Empty2k12:
Sorry to revive this old thread, I have another interesting reduce problem where the reduce-syntax does not work like I expected it to.

Basically, all my markers have a custom property which is a string. Let's stay the property is `ethnicity` and the values are all from the set [`Hispanic`, `White`, `Black`, `Asian`]. Let's also assume this set is dynamic and I don't know the types of ethnicities in my data. Let's also assume I know all possible variations once I work with the data. I would like to aggregate the types of contained ethnicities.

What I am doing right now, is the following:
```
clusterProperties: {
    ethnicities: [
        [
            "concat",
            ["accumulated"],
            [
                "case",
                [
                    "!",
                    [
                        "in",
                        ["get", "ethnicity"],
                        ["accumulated"]
                    ]
                ],
                ["concat", "-", ["get", "ethnicity"]],
                ""
            ]
        ],
        ["get", "ethnicity"]
    ]
}
```

Basically in pseudo code:
```
concat to "accumulated" (if "ethnicity" is not substring of "accumulated" use "-" + "ethnicity" else emptystring)
```

Basically, add all the ethnicities in the cluster into a String, but do not repeat an ethnicity if it is already in the string. I want to use this to generate a fitting icon for the cluster.

However, it appears only the first `ethnicity` is appended into the string, and for each member of the cluster only the minus is appended, not `-${ethnicity}`. Additionally, I'd only expect as many minuses as there are ethnicities in the cluster, not one for every cluster member. 

How can I reduce to properties without repeating?

Issue #7451, opened 2018-10-19
Stagger texture uploads across multiple consecutive frames


tmpsantos:
On a big map, specially on high resolution screen,uploading textures might take a lot of time, specially for rendering the first frame. This will cause frame drops.

One strategy for mitigating the issue would be not upload all the textures at one and postpone some for the next frame. We would need to set a hard limit on how many textures we upload per frame.

mourner:
Closing as duplicate of #7405

Issue #7405, opened 2018-10-11
Reduce slowness when loading many raster tiles at once


kkaefer:
We're bottlenecked on too many texture uploads occurring at the same time, and Webkit's timeline recorder confirms this:

<img width="951" alt="image" src="https://user-images.githubusercontent.com/52399/46813736-f1702780-cd7f-11e8-83be-0a7ed481f312.png">

It seems that uploading a single texture takes ~7.5ms per raster tile. On smaller screens, this is typically split up across multiple frames, but if you have a fast network and a big screen, many of the load events can happen within the same frame.

We should:
* Investigate why the calls to `texImage2D` and `generateMipmap` raster tile take a combined ~7.5ms. The tiles in the screenshot above `@2x` tiles (512×512), but even then ~7.5ms seems excessive.
* Distribute raster tile loads across multiple frames and upload at most 1-2 tiles per frame.
* Don't generate the mipmaps immediately after loading, but wait a frame to spread the load.

Further avenues for investigation:
* Check whether using `texSubImage2D` to upload the bitmap in batches, and spread the upload of large bitmaps across multiple frames improves the performance.
* A [comment on the chromium mailing list](https://groups.google.com/a/chromium.org/d/msg/chromium-dev/0j74Nd7i7IA/NmYiPkClKAAJ) claims:
  > Image decompression happens after the onload handler fires. [...] The ImageBitmap API will fix this problem.

  The [ImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap) API is supported in Chrome and Firefox at the time of this writing, but not Safari/Webkit, IE and Edge. It looks like we could use this to make image decoding asynchrous: The `createImageBitmap` constructor method returns a promise that will be resolve to a `ImageBitmap` object, so the call to `texImage2D` doesn't need to decode the image first. (See example in https://gist.github.com/ahem/d19ee198565e20c6f5e1bcd8f87b3408). Note that [at least Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=580202) still decodes on the current, thread, so we need to do this in a webworker.

ansis:
Is it faster later on when it can reuse old textures and upload with `texSubImage2D`?

Issue #7555, opened 2018-11-05
Use ImageBitmap API in Chrome and Firefox to decode images


kkaefer:
The [ImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap) API is supported in Chrome and Firefox at the time of this writing, but not Safari/Webkit, IE and Edge. It looks like we could use this to make image decoding asynchrous: The `createImageBitmap` constructor method returns a promise that will be resolve to a `ImageBitmap` object, so the call to `texImage2D` doesn't need to decode the image first. (See example in https://gist.github.com/ahem/d19ee198565e20c6f5e1bcd8f87b3408). Note that [at least Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=580202) still decodes on the current, thread, so we need to do this in a webworker.

See https://github.com/mapbox/mapbox-gl-js/issues/7405 for the master ticket around image decoding/uploading performance.

PR #7592, opened 2018-11-14
Upload one raster tile texture per render frame

ryanhamley:
## Launch Checklist

This closes https://github.com/mapbox/mapbox-gl-js/issues/7451

It also completes some of the work in https://github.com/mapbox/mapbox-gl-js/issues/7405

    - introduces a queue to limit raster tile texture upload to the GPU to one texture per render frame
 - [x] post benchmark scores

screen shot 2018-11-14 at 3 42 03 pm

Benchmarks don't show any major performance improvement or degradation

screen shot 2018-11-20 at 4 31 39 pm
screen shot 2018-11-20 at 4 31 49 pm
screen shot 2018-11-20 at 4 32 19 pm
screen shot 2018-11-20 at 4 32 27 pm

### DIFF
--- a/src/source/raster_tile_source.js
+++ b/src/source/raster_tile_source.js
@@ -36,11 +36,11 @@ class RasterTileSource extends Evented implements Source {
     dispatcher: Dispatcher;
     map: Map;
     tiles: Array<string>;
+    textureQueue: Array<Object>;
 
     _loaded: boolean;
     _options: RasterSourceSpecification | RasterDEMSourceSpecification;
     _tileJSONRequest: ?Cancelable;
-    _textureQueue: Array<Object>;
 
     constructor(id: string, options: RasterSourceSpecification | RasterDEMSourceSpecification, dispatcher: Dispatcher, eventedParent: Evented) {
         super();
@@ -56,7 +56,7 @@ class RasterTileSource extends Evented implements Source {
         this.tileSize = 512;
         this._loaded = false;
 
-        this._textureQueue = [];
+        this.textureQueue = [];
         this._options = extend({}, options);
         extend(this, pick(options, ['url', 'scheme', 'tileSize']));
     }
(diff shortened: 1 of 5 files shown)


src/source/raster_tile_source.js
```diff
@@ -117,22 +120,25 @@ class RasterTileSource extends Evented implements Source {
...
+                    tile.texture = this.map.painter.getTileTexture(img.width);
+                    if (tile.texture) {
+                        tile.texture.update(img, { useMipmap: true });
+                    } else {
+                        tile.texture = new Texture(context, img, gl.RGBA, { useMipmap: true });
+                        tile.texture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE, gl.LINEAR_MIPMAP_NEAREST);
+
+                        if (context.extTextureFilterAnisotropic) {
+                            gl.texParameterf(gl.TEXTURE_2D, context.extTextureFilterAnisotropic.TEXTURE_MAX_ANISOTROPY_EXT, context.extTextureFilterAnisotropicMax);
+                        }
```
kkaefer: Given that we're now deferring the action of uploading the tile, `abortTile` no longer cancels everything: There could be a situation where the request responds with a tile, we store in the queue, then the tile gets aborted/deleted. We now still hold on to the tile and upload it to the GPU, but it'll never be used and gets GCed once the callback/tile object is removed from the `textureQueue`. We should add a way to either remove requests from the `textureQueue` that reference this tile, or use another mechanism to prevent tiles from remaining alive longer than they need to.

src/source/source.js
```diff
@@ -48,6 +48,7 @@ export interface Source {
     tileID?: CanonicalTileID;
     reparseOverscaled?: boolean,
     vectorLayerIds?: Array<string>,
+    +textureQueue?: Array<Object>,
```
kkaefer: Why is the `textureQueue` per-source rather than per-style?

src/source/raster_tile_source.js
ryanhamley: Is it enough to check that the tile pulled from `textureQueue` doesn't have the `aborted` property before uploading to the GPU? Or would that be unreliable and not cover all potential cases?

kkaefer: That works; another option would be to remove the object from the queue when the tile is aborted.

mourner: There's similar logic here if that helps: https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/ajax.js#L196-L197

ryanhamley:
My manual testing seems to indicate that 2 is an optimum number of textures to upload at once. 1 and 2 textures both show a roughly 12% dropped frame rate while at 4 textures per frame, the dropped frame rate increases to roughly 20%. 1 texture per frame is too slow, especially on pitched maps while 2 textures per frame looks much snappier and uses half as many frames.

andycalder:
This looks like a great performance improvement to me! I tested this on a satellite map with 60 degree pitch and the frame rate difference is very noticeable on initial map load when the tiles have been cached by the browser. I'd love to see this PR move forward. Is it a work in progress?

Some things I noticed:
* This breaks the tile fade-in transition because the `_tileLoaded()` callback sets `tile.timeAdded` immediately, and this starts the opacity interpolation before the tile is even rendered. I quick-fixed this by moving the callback to `textureCallback()`. Visual comparison with master is more favourable when the fade works.
* For the raster-dem case, it looks like you're queueing the DEM decode operation, which happens in a web worker, instead of the actual texture upload. Texture upload happens in `draw_hillshade.js`.

ryanhamley:
Thanks for the feedback @andycalder! It's a work in progress, yes. We've been investigating a few different ways of making performance better with raster image loading. https://github.com/mapbox/mapbox-gl-js/issues/7555 is the master ticket for this work. A related improvement I've been working on is using the Image Bitmap API in browsers that support it which you can see in https://github.com/mapbox/mapbox-gl-js/tree/image-bitmap-prototype (very rough prototype).

I actually hadn't seen a huge improvement with this PR and in fact, when combined with https://github.com/mapbox/mapbox-gl-js/pull/7497 the performance seemed worse since raster requests were essentially being rate limited twice. It definitely requires more investigation. Are you able to share a minimal example in which you saw a performance improvement?

ryanhamley:
Also, I don't think the version of this branch on Github had been updated with #7497 previously so I'd recommend trying it now that I've rebased against `master` @andycalder

vakila:
Closing this as stale/no longer WIP per chat with @ryanhamley

PR #7594, opened 2018-11-16
Canonicalize tile url

### LINKED ISSUE #7400: "canonicalize" tile urls from TileJSON so they use config.API_URL
gl-native has a step where it converts mapbox.com tile urls to use the mapbox:// prefix so that they ultimately fire requests at the `baseApiUrl` domain. we should add behavior in mapbox-gl-js as well, I think.

```cpp
// Return a "mapbox://tiles/..." URL (suitable for normalizeTileURL) for the given Mapbox tile URL.
std::string canonicalizeTileURL(const std::string& url, style::SourceType, uint16_t tileSize);
```
https://github.com/mapbox/mapbox-gl-native/blob/f3341dd589c6330c0cfd6e8e381398d08493b48a/src/mbgl/util/mapbox.hpp#L21-L22

we should also consider adding a setter/getter for API_URL (native exposes these functions [here](https://github.com/mapbox/mapbox-gl-native/blob/d4f41364140cc37820f9a9bbff58e5e124097469/include/mbgl/storage/default_file_source.hpp#L45-L46))

mollymerp:
close #7400 

This PR does a couple 🆕 things to improve parity between gl-native and gl-js:
- adds setter and getter for `config.API_URL`
- transforms tile URLs to use the config.API_URL if using a `mapbox://` url, like the rest of the assets requested by the library

this is technically changing existing behavior, but it shouldn't be breaking, unless a user was creating their own TileJSON endpoint to respond with an object where the `tiles` urls were from a different domain than the TileJSON/other assets 😬 🙏 So I guess it maybe technically is breaking?

Also, as far as I know the reason for providing multiple urls like `a.tiles.mapbox.com`, `b.tiles.mapbox.com` was to workaround the max open connections the browser will allow for any given origin for HTTP 1.0. When profiling locally on a large monitor, I do notice that some tiles requests are queued briefly (~10-50ms) when many tiles are being requested at once, but I'm not noticing a real big in time to first render (TTFR) or zoom/pan performance. @kkaefer any thoughts on how we should verify this? 

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/source/load_tilejson.js
+++ b/src/source/load_tilejson.js
@@ -28,8 +28,7 @@ export default function(options: any, requestTransformFn: RequestTransformFuncti
 
             // only canonicalize tile tileset if source is declared using a tilejson url
             if (options.url) {
-                const canonicalTileUrls = canonicalizeTileset(result, options.url);
-                result.tiles = canonicalTileUrls;
+                result.tiles = canonicalizeTileset(result, options.url);
             }
             callback(null, result);
         }
--- a/src/source/load_tilejson.js
+++ b/src/source/load_tilejson.js
@@ -4,7 +4,7 @@ import { pick } from '../util/util';
 
 import { getJSON, ResourceType } from '../util/ajax';
 import browser from '../util/browser';
-import { normalizeSourceURL as normalizeURL } from '../util/mapbox';
+import { normalizeSourceURL as normalizeURL, canonicalizeTileset } from '../util/mapbox';
 
 import type {RequestTransformFunction} from '../ui/map';
 import type {Callback} from '../types/callback';
@@ -26,6 +26,11 @@ export default function(options: any, requestTransformFn: RequestTransformFuncti
                 result.vectorLayerIds = result.vectorLayers.map((layer) => { return layer.id; });
             }
 
+            // only canonicalize tile tileset if source is declared using a tilejson url
+            if (options.url) {
+                const canonicalTileUrls = canonicalizeTileset(result, options.url);
+                result.tiles = canonicalTileUrls;
+            }
             callback(null, result);
         }
     };
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -10,6 +10,7 @@ import { postData } from './ajax';
 
 import type { RequestParameters } from './ajax';
 import type { Cancelable } from '../types/cancelable';
+import type {TileJSON} from '../types/tilejson';
 
 const help = 'See https://www.mapbox.com/api-documentation/#access-tokens';
 const telemEventKey = 'mapbox.eventData';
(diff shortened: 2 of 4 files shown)


mollymerp:
Here are the benchmark results. I didn't notice anything unusual, and really wouldn't expect to since the benchmarks don't actually request many tiles.

<details>
screencapture-localhost-9966-bench-versions-2018-11-15-16_30_01
</details>

src/source/load_tilejson.js
```diff
@@ -26,6 +26,11 @@ export default function(options: any, requestTransformFn: RequestTransformFuncti
                 result.vectorLayerIds = result.vectorLayers.map((layer) => { return layer.id; });
             }
 
+            // only canonicalize tile tileset if source is declared using a tilejson url
+            if (options.url) {
+                const canonicalTileUrls = canonicalizeTileset(result, options.url);
+                result.tiles = canonicalTileUrls;
```
ryanhamley: Could just be `result.tiles = canonicalizeTileset(result, options.url);`

src/util/mapbox.js
```diff
@@ -100,18 +101,49 @@ export const normalizeTileURL = function(tileURL: string, sourceURL?: ?string, t
     const suffix = browser.devicePixelRatio >= 2 || tileSize === 512 ? '@2x' : '';
     const extension = browser.supportsWebp ? '.webp' : '$1';
     urlObject.path = urlObject.path.replace(imageExtensionRe, `${suffix}${extension}`);
+    urlObject.path = `/v4${urlObject.path}`;
```
ryanhamley: Are we sure that `urlObject.path` will always begin with a `/`?

src/util/mapbox.js
```diff
@@ -100,18 +101,49 @@ export const normalizeTileURL = function(tileURL: string, sourceURL?: ?string, t
...
-    replaceTempAccessToken(urlObject.params);
-    return formatUrl(urlObject);
+    return makeAPIURL(urlObject);
 };
 
-function replaceTempAccessToken(params: Array<string>) {
-    for (let i = 0; i < params.length; i++) {
-        if (params[i].indexOf('access_token=tk.') === 0) {
-            params[i] = `access_token=${config.ACCESS_TOKEN || ''}`;
+export const canonicalizeTileURL = function(url: string) {
```
ryanhamley: This can be an arrow function I think. Same in line 135.

src/util/mapbox.js
```diff
@@ -100,18 +101,49 @@ export const normalizeTileURL = function(tileURL: string, sourceURL?: ?string, t
...
+    const version = "/v4/";
+
+    const urlObject = parseUrl(url);
+    // Make sure that we are dealing with a valid Mapbox tile URL.
+    // Has to be /v4/, with a valid filename + extension
+    if (!urlObject.path.match(version) || !(urlObject.path.match(imageExtensionRe) || urlObject.path.match('.pbf'))) {
```
ChrisLoer: The logic below depends on `version` matching at the beginning of the path, right? We should probably explicitly require that? Also we should check that the extensions are at the end of the path (I think the image extension regex already does that)?

src/util/mapbox.js
```diff
@@ -100,18 +101,49 @@ export const normalizeTileURL = function(tileURL: string, sourceURL?: ?string, t
...
+        return url;
+    }
+    // Reassemble the canonical URL from the parts we've parsed before.
+    let result = "mapbox://tiles/";
+    result +=  urlObject.path.replace(version, '');
+
+    // Append the query string, minus the access token parameter.
+    let ampersand = false;
+    for (const param of urlObject.params) {
+        if (!param.match('access_token')) {
```
ChrisLoer: We should match something like `/^access_token=/`, right? You could write this as this as `params.filter(param => !param.match(/^access_token=/))`, and then if filteredParams.length > 0 then add `?${filteredParams.join("&")}`

src/util/mapbox.js
mollymerp: following convention here.

src/util/mapbox.js
mollymerp: yes, see `urlRe`

src/util/mapbox.js
ryanhamley: 👍 I thought maybe that was the case. Good to know.

src/util/mapbox.js
kkaefer: We should check for string equality here instead of just an arbitrary match, since we only want to exclude the `access_token` parameter, not other parameters that might include the string too.

src/util/mapbox.js
```diff
@@ -100,18 +101,49 @@ export const normalizeTileURL = function(tileURL: string, sourceURL?: ?string, t
...
+    const canonical = [];
+    if (isMapboxURL(sourceURL)) {
+        for (const url of tileJSON.tiles) {
+            const canonicalUrl = canonicalizeTileURL(url);
+            canonical.push(canonicalUrl);
+        }
+    } else {
+        return tileJSON.tiles;
+    }
+    return canonical;
```
kkaefer: Let's eliminate creating `canonical` for the `!isMapboxURL` case

src/util/mapbox.js
```diff
@@ -100,18 +102,43 @@ export const normalizeTileURL = function(tileURL: string, sourceURL?: ?string, t
...
+    return makeAPIURL(urlObject);
+};
+
+export const canonicalizeTileURL = function(url: string) {
+    const version = "/v4/";
+
+    const urlObject = parseUrl(url);
+    // Make sure that we are dealing with a valid Mapbox tile URL.
+    // Has to begin with /v4/, with a valid filename + extension
+    if (!urlObject.path.match(/(^\/v4\/)/) || !(urlObject.path.match(imageExtensionRe) || urlObject.path.match(vectorExtensionRe))) {
```
kkaefer: Any reason we're checking the extensions? Looking at the native code, we're not doing that there and it seems like a restricting that might cause problems down the road if we decide to change the extensions.

src/util/mapbox.js
mourner: Nit: I think the convention is `export function fn(...)`, not `export const fn = function(...)`. So this looks like a pre-existing omission.

src/util/mapbox.js
mollymerp: I'm not sure I understand what you mean – the `param` string is both the query param and its value, so checking for string equality wouldn't be possible without splitting up the string. I took Chris' suggestion of making the regex specify the start of the string with an `=`

src/util/mapbox.js
mollymerp: hmm, yeah I guess the native version is just checking that the extension length is greater than 1.

src/util/mapbox.js
mollymerp: yeah I was following convention of the functions in this file – do you want me to change all of them ?

src/util/mapbox.js
kkaefer: Ah sorry, I thought it was a key/value object

src/util/mapbox.js
```diff
@@ -100,18 +103,40 @@ export const normalizeTileURL = function(tileURL: string, sourceURL?: ?string, t
...
+        // Not a proper Mapbox tile URL.
+        return url;
     }
-}
+    // Reassemble the canonical URL from the parts we've parsed before.
+    let result = "mapbox://tiles/";
+    result +=  urlObject.path.replace(version, '');
+
+    // Append the query string, minus the access token parameter.
+    const params = urlObject.params.filter(p => !p.match(/^access_token=/));
```
kkaefer: 👍

PR #7596, opened 2018-11-17
[WIP] add dynamic text placement for point labels

mollymerp:
Implements https://github.com/mapbox/mapbox-gl-js/issues/5038

To do:
- [ ] implement icon-aware offsets of text-labels that are placed dynamically
- [ ] add `dynamic-text-offset` radial distance property
- [ ] fix drifty label issue on zoom
- [ ] tighten validation on `dynamic-text-placement`
- [ ] fix broken tests / flow
- [ ] add render tests
- [ ] add docs/examples

cc @nickidlugash 

## Launch Checklist

 - [ ] post benchmark scores


ChrisLoer:
I'm seeing collision detection sometimes fail when `dynamic-text-anchor` is set to `["auto"]`.

screenshot 2018-12-20 16 49 30

Investigating.

ChrisLoer:
The issue here is that pitch and rotation aren't being correctly handled by the dynamic anchor offset code, because the offset is applied before projection happens in the shader. I think we could:

- Do the projection in the CPU at render time, and change the `u_label_plane_matrix` uniform to the identity matrix -- hopefully as close as possible to the way we currently handle line labels.
- Take advantage of the fact that for point labels `a_pos` and `a_projected_pos.xy` are currently duplicate, change `a_projected_pos.xy` to hold a tile-unit anchor offset, and then apply the anchor offset in the shader.

I'm leaning towards the first solution because I'm hoping it will allow us to avoid adding yet more confusing logic to the already-too-complex shader. It's extra CPU work, but hopefully it should be relatively small compared to the amount of projection we're already doing for line labels.

cc @ansis

nickidlugash:
@ChrisLoer I noticed a couple of minor irregularities/unexpected behavior with text offsetting:

- Left/right placement: the vertical alignment looks slightly off (text seems too high)
  _dynamic placement:_  
  screen shot 2019-01-04 at 5 19 34 pm
  
  _static placement:_
  screen shot 2019-01-04 at 5 19 03 pm

- Right placement: horizontal offset looks too large (see left placement for comparison)
  screen shot 2019-01-04 at 5 57 06 pm

- Bottom placement: vertical offset looks too large (see top placement for comparison):
   screen shot 2019-01-04 at 6 36 34 pm

- Diagonal placements (top-left, etc): offset looks too large (see left and top placements for comparisons)
    screen shot 2019-01-04 at 7 01 50 pm
  I think this particular issue is an expected visual result of having the same radial offset value everywhere: when text is placed horizontally or vertically, the closest part of the text box to the anchor is an edge, whereas when text is placed diagonally, the closest part of the text box to the anchor is a corner. The second case results in a larger apparent space between icon and text. I wonder if we should very slightly reduce the calculated offset values for diagonal placement to account for this?

ChrisLoer:
Thanks for documenting these cases! The current offset logic just doesn't work in a lot of cases, and I think we need to apply the offset at render time instead of at collision detection time the way the current code runs. As I work on it, I'll check against these situations. Rotation and pitch handling for the offset are also broken.

> The second case results in a larger apparent space between icon and text. I wonder if we should very slightly reduce the calculated offset values for diagonal placement to account for this?

🤔Instead of doing a radial offset, should we just apply the "dynamic-text-offset" in both the x and y directions for the diagonal cases? Or we could for the corner cases we could use the full x offset and some smaller y offset, to avoid making the radial distance increase so much (although this could be hard to explain clearly).

nickidlugash:
@ChrisLoer Just wanted to note here, per chat, that the first three offset issues I noted [in my comment above](https://github.com/mapbox/mapbox-gl-js/pull/7596#issuecomment-451606806) seem to have been resolved via your [recent commit](https://github.com/mapbox/mapbox-gl-js/pull/7596/commits/78519a9d5b52be26a408123b6bbd5bd153457da1) that changed the base anchor of offsets 🎉 

For the last issue regarding diagonal placements, looking at this again I think an additional problem here is that the text itself within a text box is further away from the corner of the text box than it is from the sides, so I think that the shortest distance from the anchor to closest point of the text is actually larger.

> Instead of doing a radial offset, should we just apply the "dynamic-text-offset" in both the x and y directions for the diagonal cases? Or we could for the corner cases we could use the full x offset and some smaller y offset, to avoid making the radial distance increase so much (although this could be hard to explain clearly).

Wouldn't these two options make the diagonal offsets larger, rather than smaller? 🤔

ChrisLoer:
> Wouldn't these two options make the diagonal offsets larger, rather than smaller? 🤔

Ah, I was thinking about this the wrong way -- I was thinking that keeping the alignment of the left/right edges of the boxes consistent would make it more legible. Yeah, you're right.

> the text itself within a text box is further away from the corner of the text box than it is from the sides, so I think that the shortest distance from the anchor to closest point of the text is actually larger.

Yeah, although I think for a single-line label this should be a pretty small effect. For a multi-line label it could be more dramatic if for instance the bottom line is significantly shorter than the top line. I'm somewhat uncomfortable with just arbitrarily shortening the radial offset in the diagonal case, 'cause I'm not sure it would always get the effect you want and it just seems harder to explain?

ChrisLoer:
OK, I think I have viewport-pitch/rotation-aligned labels working in pitched views now. There are a bunch of loose ends I need to track down:

- [ ] Tests exercising the cases I've fixed since I picked this up
- [ ] Disable support for "map" pitch-rotation alignment? I've implemented support on the rendering side, but we already had pretty bad support for collision detection with map-rotated point labels, and it just gets more confusing with dynamic anchors added.
- [ ] Enforce "text-offset" vs "dynamic-text-offset" at style validation time
- [ ] Public example
- [ ] Simplify code: I haven't focused on this yet, but I think there are a fair number of places where we transform/copy things more times than we need to
- [ ] Consider alternate approach: do the anchor offset in the shader? Current approach of calculating the anchor offsets in the CPU for each frame seems workable, but _if_ we found a good way to get the necessary data to the shader, it'd be great to be able to piggyback on existing logic (and take advantage of the GPU)
- [ ] Consider alternate approach: use a different shaping/set of glyphs for every single position. If we do this we have a maximum of 9 sets of glyphs, instead of the current maximum of 3 (one for each justification). On the other hand, all we have to do is select which one to use, and then we can rely entirely on existing rendering logic.
- [ ] Consider perf optimization: for single line labels, there's no need to generate/upload 3 shapings for the three different justifications, since they're all the same.
- [ ] Consider "stability" improvement: for any given dynamic anchor placement, start with the anchor position that was last placed successfully, before moving on to the rest of the list provided in `dynamic-text-anchor`. (e.g. if "left" is first in our list, but we were successfully placed in "top-left" in the last placement, try "top-left" first in the current placement before going on to "left")
- [ ] Consider cross-fading
- [ ] Consider naming and/or style-spec changes? Molly and @nickidlugash already chewed on this for a while... but I'm still not super comfortable with "dynamic" as a descriptive name. What about "multiple-text-anchor" or "auto-text-anchor"? Also -- do we need a separate radial offset property, or could we get away with just using the existing `text-offset` property? This may be too hacky an idea, but we could automatically take the magnitude of the `text-offset` vector to convert it into a radial offset.

cc @ansis

ChrisLoer:
I did some initial profiling of a large-screen NYC dynamically placed POI case. It should be a pretty good stress test of a very dynamic-anchor heavy map:

screenshot 2019-01-11 16 37 08

The actual cost of the extra collision detection doesn't end up looking too large (`attemptAnchorPlacement` is where most of the dynamic work takes place):

screenshot 2019-01-11 16 25 47


On the other hand, at render time, in this example, the dynamic placement of anchors looks like it's costing twice as much as the dynamic updating of road labels (this example does have many more POIs than roads).

screenshot 2019-01-11 16 24 47

That render-time cost is dominated by re-building and re-uploading the dynamic layout arrays.

`updateBucketOpacities` and shaping are presumably three times as expensive for the dynamic anchor layer because there are three times as many glyphs, but that cost isn't split out in the profiling.

Overall, this seems tolerable? An alternative approach would be to upload a full set of glyphs for every possible anchor placement. The tradeoffs are roughly:

- More static data to upload to GPU, and more vertices to render in the shader (although we could still choose to move the bulk of the unused vertices offscreen). More time spent updating opacities
- Makes cross-fading more straightforward (since we have a set of glyphs to render simultaneously for both positions)
- No render-time dynamic layout updates necessary
- Possibly simpler?

I'm not sure it's likely to make a big enough difference to be worth prototyping.

ChrisLoer:
I implemented the "only do multiple shapings for multi-line labels" optimization, and in the NYC dense POI label test case from above, it looks like a pretty significant improvement. I wouldn't put too much weight on the absolute numbers, but what looks promising is the dynamic-anchor:line-label cost ratio went from 2:1 down to 1:1.

screenshot 2019-01-14 18 31 35

From discussion with @chloekraw @nickidlugash and @elifitch, we're currently planning these changes to the style-spec side:

- `variable-text-anchor` and `variable-text-offset` instead of `dynamic-`
- Get rid of the "auto" option. Rationale: users can set it up if they want, but we don't want to make it the "easy" default, since it (a) often doesn't look that great (too hard to tell what label belongs with what anchor), (b) is the least performant option, and (c) doesn't quite match the way the rest of our "auto" parameters work.
- Keep the distinction between `variable-text-offset` and `text-offset`, with a style spec validation error if you use the wrong one.

ChrisLoer:
I just pushed a small change to the rendering of debug collision boxes that makes them a little more legible in my mind:

- If an unplaced symbol was _previously_ placed, we show the red collision box in the last placed location (so you can see where the collision happened, and also the most likely place for it to reappear).
- If an unplaced symbol was _never_ placed (this gets reset when a tile moves in/out of viewport), then show the center box to indicate something was there, but fade it out because it's mostly distracting.

ChrisLoer:
Nothing stood out in the benchmarks (I didn't make a new benchmark specifically for variable anchors, so this is mainly looking for regression that some of the overhead might have introduced):

screenshot 2019-01-17 17 07 14
screenshot 2019-01-17 17 06 41
screenshot 2019-01-17 17 05 13

ChrisLoer:
@ansis I think this is ready for review. I edited Molly's original post to include a little introduction.

cc @mapbox/studio @mapbox/map-design (see style-spec changes in v8.json or quick description at top of PR).

chloekraw:
@bradjonesca - this won't make it into v0.53, but it will be in v0.54 (~early March). Thanks for using Mapbox!

ChrisLoer:
I talked this over with @ansis and I’m OK with removing the animated label transitions -- does this sound OK to you @nickidlugash and @chloekraw?

My thinking is roughly:
1) _I_ think it’s an improvement for the case where just a few important things are showing on the screen — which I think is the FB case. It’s obviously bad in the “show a high density of POIs” use case, but honestly I think using variable placement for high density POIs looks bad no matter what (Apple Maps seems to feel differently). At any rate, this animation is not a must-have feature anyone has requested.
2) I don’t think there’s a lot of code complexity in the feature, but that also means it’s pretty easy to split out into a separate PR we could choose to accept later.
3) The strongest argument against it in my mind is that adding it to the style-spec just because we can locks us into supporting it in the future. Even though it’s easy to support now, if we decided to support cross-fading in the future, it would actually be difficult to have parallel support for both modes. So if we avoid implementing now, we have more flexibility in the future.

I'll try to do this tomorrow.

As for collapsing `text-anchor` and `text-variable-anchor` into a single property that could be either a singular anchor position or an array of positions -- I like the idea, but I think it might be a little tricky to implement in the style-spec. I don't think I'll have a chance to do it. If we do decide to come back and do it, I don't think we'd be in a terrible position style-spec wise because we could just deprecate `text-variable-anchor` but maintain support and just internally convert it to the new syntax.

PR #7611, opened 2018-11-21
[WIP] Support benchmarking styles from file paths

tristen:
Detects stylesheets originating from local paths at the rollup config stage and inlines them into the environment variable.

Also:

- Renamed `MAPBOX_STYLE_URL` to `MAPBOX_STYLES` which aligns better with what is now supported.
- Flow flagged `layer.ref` not being type defined which I think is accurate (it should not exist) so I removed the conditional in https://github.com/mapbox/mapbox-gl-js/pull/7611/commits/a8a0d78c682aeed53653d697f5e8d58bcfaa80ad
- I can leave it out of this PR but I noticed an [environment variable defined](https://github.com/mapbox/mapbox-gl-js/pull/7611/files#diff-eba31f50800e4aa9b1dca219296e584aL12) that is never used so I removed it.

---

Lastly, I created a branch from this that fetched all stylesheets during the rollup config step which omitted the need for `bench/lib/fetch_style.js` or a distinction between string urls or style objects but ran into a few difficulties getting it to work and punted.

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -21,9 +21,9 @@ By default, the benchmark page will compare the local branch against `master` an
 Start the benchmark server
 
 ```bash
-MAPBOX_ACCESS_TOKEN={YOUR MAPBOX ACCESS TOKEN} MAPBOX_STYLE_URL={YOUR STYLES HERE} yarn start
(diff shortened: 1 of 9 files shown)


ryanhamley:
Awesome! One thing I'd suggest before finalizing this is updating to the newest version of Yarn before installing the `pify` module. This should prevent the integrity fields of lots of unchanged packages from being changed in the `yarn.lock` file. I ran into something similar recently and upgrading did the trick.

Issue #2074, opened 2016-02-05
Fix rendering of symbol layers under solid fill layers


ansis:
It looks like depth testing is disabled for symbol rendering. Is there a good reason?

@scothis

lucaswoj:
Not as far as I know. Looking back at git history suggests that this existed in your original layer-by-layer branch. https://github.com/mapbox/mapbox-gl-js/blame/aaa7e63bc88bada9b967062826161decf1106c94/js/render/draw_symbol.js#L29

ansis:
When rendering we draw all solid fills top down. We then render everything else bottom-up. We use the depth buffer to drop all fragments that are behind solid fills. This is faster than rendering everything bottom-up because we can drop fragments.

Depth testing is disabled for symbols because it breaks `rotation-alignment: viewport` labels in perspective view. Since the extrudes are added after projection the depth values don't match up. I don't think there is a simple way to fix the depth values in this case.

![screen shot 2016-02-05 at 2 43 11 pm](https://cloud.githubusercontent.com/assets/1421652/12861248/e3002ac8-cc16-11e5-935e-eeb0d8660f6f.png)

We can enable depth testing for `rotation-alignment: map` labels and fix rendering order in all those cases. **We should definitely do this**. This would fix oneway arrows drawn under bridges.

We can enable depth testing for `rotation-alignment: viewport` labels when pitch is 0. **Should we do this?** The labels would pop up above the fill layers as soon as you tilt.

If we really want to obey layer order for `rotation-alignment: viewport` labels in tilted views we need to treat all fill layers above the symbol layers as if they are transparent and draw them in the bottom-up pass. **Should we do this?** Are there any cases when this would matter? I'm not sure this would make sense conceptually since in perspective view these labels exist in 3d space not the 2d layer space.

lucaswoj:
The "[highlight features under the mouse pointer](https://www.mapbox.com/mapbox-gl-js/example/hover-styles/)" example demonstrates this bug well

andrewharvey:
This bug causes masks to not really work as intended

![selection_290](https://cloud.githubusercontent.com/assets/117278/15706828/ea5e5c76-2838-11e6-901d-df389f2f6e0d.png)

Ideally the labels would also be masked, but due to this bug they appear on top.

scothis:
A hacky workaround is to set the fill layer's opacity to `0.99`

jfirebaugh:
[...]

PR #7612, opened 2018-11-21
Fix symbol rendering under opaque fill layers

ryanhamley:
## Launch Checklist

    - fixes https://github.com/mapbox/mapbox-gl-js/issues/2074
    - enables depth testing for symbol layers so that opaque fill layers can cover symbol layers

Without fix
screen shot 2018-11-21 at 2 49 19 pm

With fix
screen shot 2018-11-21 at 2 47 42 pm


ryanhamley:
It didn't seem to matter which depth calculation I chose in `painter.depthModeForSublayer` so I arbitrarily chose `farDepth` and renamed it `depth`. I can change this if I'm overlooking something.

Also just want to ensure that this is the behavior we're looking for in this situation, where the underlying symbol layer is cut off and obscured by the opaque fill layer. It seems like the expected behavior to me.
screen shot 2018-11-21 at 2 47 28 pm

ansis:
This looks good. Can you confirm that symbols are not [clipped](https://cloud.githubusercontent.com/assets/1421652/12861248/e3002ac8-cc16-11e5-935e-eeb0d8660f6f.png) when the map is pitched?

PR #7615, opened 2018-11-22
Fix sporadical "cannot read property _values of undefined" error

### LINKED ISSUE #7523: JS errors while zooming (Cannot read property '_values' of undefined)
**mapbox-gl-js version**: 0.50

**browser**: Chrome 69

### Steps to Trigger Behavior

 1. Zoom in and out rapidly

### Link to Demonstration

I can try to provide one if this isn't immediately obvious from the stack trace

https://jsbin.com/

### Expected Behavior

No errors

### Actual Behavior

This stack trace (pardon the formatting, this was extracted manually from our production error reporter with sourcemaps turned on)

```
Error: Cannot read property '_values' of undefined

at in in webpack:///src/style/style_layer.js
Line 248 : Column 49
247 | isStateDependent() {
248 | for (const property in (this: any).paint._values) {
249 | const value = (this: any).paint.get(property);

at filter in webpack:///src/data/bucket.js
Line 110 : Column 68
109 | (bucket: any).layers = layers;
110 | (bucket: any).stateDependentLayers = layers.filter((l) => l.isStateDependent());
111 | for (const layer of layers) {

at Array.filter 

at filter in webpack:///src/data/bucket.js
Line 110 : Column 52
109 | (bucket: any).layers = layers;
110 | (bucket: any).stateDependentLayers = layers.filter((l) => l.isStateDependent());
111 | for (const layer of layers) {

at CollisionBoxArray in webpack:///src/source/tile.js
Line 171 : Column 41
170 | this.collisionBoxArray = data.collisionBoxArray;
171 | this.buckets = deserializeBucket(data.buckets, painter.style);
172 |  

at callback in webpack:///src/source/vector_tile_source.js
Line 143 : Column 17
142 | if (this.map._refreshExpiredTiles && data) tile.setExpiryData(data);
143 | tile.loadVectorData(data, this.map.painter);
144 |  

at callback(deserialize in webpack:///src/util/actor.js
Line 83 : Column 12
82 | } else if (callback) {
83 | callback(null, deserialize(data.data));
84 | }

```

mourner:
Closes #7523. I'm not able to trace the exact race condition that causes the error, but it should be fixed with this change. Also slightly improves typing, which would catch the potential error if we didn't use the `(obj: any).property` hack there.

## Launch Checklist

### DIFF
--- a/src/style/style_layer.js
+++ b/src/style/style_layer.js
@@ -17,7 +17,7 @@ import type { FeatureState } from '../style-spec/expression';
 import type {Bucket} from '../data/bucket';
 import type Point from '@mapbox/point-geometry';
 import type {FeatureFilter} from '../style-spec/feature_filter';
-import type {TransitionParameters} from './properties';
+import type { TransitionParameters, PossiblyEvaluated } from './properties';
 import type EvaluationParameters, {CrossfadeParameters} from './evaluation_parameters';
 import type Transform from '../geo/transform';
 import type {
@@ -42,11 +42,11 @@ class StyleLayer extends Evented {
     _crossfadeParameters: CrossfadeParameters;
 
     _unevaluatedLayout: Layout<any>;
-    +layout: mixed;
+    +layout: ?PossiblyEvaluated<any>;
 
     _transitionablePaint: Transitionable<any>;
     _transitioningPaint: Transitioning<any>;
-    +paint: mixed;
+    +paint: ?PossiblyEvaluated<any>;
 
     _featureFilter: FeatureFilter;
 
@@ -245,8 +245,11 @@ class StyleLayer extends Evented {
     }
 
     isStateDependent() {
-        for (const property in (this: any).paint._values) {
-            const value = (this: any).paint.get(property);
+        if (!this.paint)
+            return false;
+
+        for (const property in this.paint._values) {
+            const value = this.paint.get(property);
             if (!(value instanceof PossiblyEvaluatedPropertyValue) || !supportsPropertyExpression(value.property.specification)) {
                 continue;
             }


src/style/style_layer.js
```diff
@@ -245,8 +245,11 @@ class StyleLayer extends Evented {
     }
 
     isStateDependent() {
-        for (const property in (this: any).paint._values) {
-            const value = (this: any).paint.get(property);
+        if (!this.paint)
+            return false;
```
kkaefer: Why is `paint` (or `layout`) optional for this `StyleLayer`?

mourner: Because it's calculated on `recalculate`, not in constructor. Alternatively, we can try setting it to empty objects there, but I think the PR solution is fine.

kkaefer: Hm, I think we should investigate this further rather than covering up the symptoms. Returning `false` from `isStateDependendent` sounds wrong (since it's in an undefined state). Instead, we should not call `isStateDependent` when `recalculate` has never been called, or ensure that it is called in the constructor.

asheemmamoowala: > we should not call isStateDependent when recalculate has never been called, or ensure that it is called in the constructor.

I wasn't able to reproduce this error condition even after a bunch of experimentation. In [36270d](https://github.com/mapbox/mapbox-gl-js/commit/36270d8381b2c30d40ef75a26e330beaf87ea4af) I took the approach of moving the paint property evaluation to the worker. This should have the additional benefit of not needing to parse the expressions on the main thread.

anshumanbora:
I'm currently facing a similar issue. Would love to get an update about the pending PR.

mourner:
@asheemmamoowala let's resurrect that [fix-7523](https://github.com/mapbox/mapbox-gl-js/compare/fix-7523) branch into a PR maybe?

asheemmamoowala:
@mourner - its up #7790.

PR #1340, opened 2015-06-26
don't reset bearing in fitbounds

kelvinabrokwa:
https://github.com/mapbox/mapbox-gl-js/issues/1225

Issue #1338, opened 2015-06-25
Allow custom bearing/rotation while using fitBounds()


friedbunny:
The `bearing` option in `fitBounds()` is currently [hard-coded to `0`](https://github.com/mapbox/mapbox-gl-js/blob/37145b8b9bb0de022811ad46f3e89a2714896b18/js/ui/camera.js#L365), but it would be nice to support eased rotation in `fitBounds()`, too. Removing that `bearing = 0` line will rotate the map, but without easing and with uncorrected bounds.

Here's an test-case around Wyoming:

```js
// Wyoming, slightly offset because of simplification
map.fitBounds([[45, -111],[41, -104]], { bearing: 45 });
```

```js
// Draw Wyoming polygon
map.on('style.load', function() {
  map.addSource("wyoming", {
    "type": "geojson",
    "data": {"type":"FeatureCollection","features":[{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-111,41],[-111,45],[-104,45],[-104,41],[-111,41]]]}}]}
  });

  map.addLayer({
    "id": "wyoming",
    "type": "fill",
    "source": "wyoming",
    "layout": {
    },
    "paint": {
      "fill-color": "blue",
      "fill-outline-color": "gray",
      "fill-opacity":  0.5
    }
  });
});
```

Which should ultimately end up like this:
![screen shot 2015-06-25 at 1 49 36 pm](https://cloud.githubusercontent.com/assets/1198851/8365318/3ffae68c-1b41-11e5-8fbb-6071e3e231af.png)

1ec5:
Dupe of #1225.

kelvinabrokwa:
should be addresses in https://github.com/mapbox/mapbox-gl-js/pull/1340

1ec5:
Reopening because #1340 has been pared down to just keep the bearing but not account for it.

1ec5:
Doh, I forgot we already had #1225.

lucaswoj:
Merged with #2112

dagjomar:
Not sure what happened with this, but cannot find it in the actual mapbox-gl-js latest release.

I ended up implementing my own version of `fitBounds`, called `fitBoundsRotated`, which has a small part of code that transforms the box and then uses that to calculate the ending bounds, plus using the map bearing or given bearing options.

```
// Copied code from MapBoxGL's Camera#fitBounds method
	// Adding in rotation logic
[...]

PR #7618, opened 2018-11-23
Allow passing bearing as an option ot fitBounds

dentuzhik:
## Launch Checklist

I know there have been other attempts to get it to `fitBounds`, but I got it working fairly easiy with a following snippet of code (react based), hence decide to file a PR.
```
mapInstance._fitInternal(
  mapInstance._cameraForBoxAndBearing(
    bounds.getNorthWest(),
    bounds.getSouthEast(),
    mapInstance.getBearing(),
    {},
  ),
);
```
I decided not to go with all the other checklist items, before getting any feedback, but surely can add docs and tests if necessary.

Previous art:
https://github.com/mapbox/mapbox-gl-js/pull/1340
https://github.com/mapbox/mapbox-gl-js/issues/1338

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -362,6 +362,7 @@ class Camera extends Evented {
      *      in the viewport. LatLngBounds represent a box that is always axis-aligned with bearing 0.
      * @param options
      * @param {number | PaddingOptions} [options.padding] The amount of padding in pixels to add to the given bounds.
+     * @param {number} [options.bearing=0] Desired map bearing at end of animation, in degrees.
      * @param {PointLike} [options.offset=[0, 0]] The center of the given bounds relative to the map's center, measured in pixels.
      * @param {number} [options.maxZoom] The maximum zoom level to allow when the camera would transition to the specified bounds.
      * @returns {CameraOptions | void} If map is able to fit to provided bounds, returns `CameraOptions` with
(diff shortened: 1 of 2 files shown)


mvriel:
Mentioning @dentuzhik in the hopes he will incorporate the suggested changes :)

dentuzhik:
Not sure why occasionally `test-flow` fails, tried locally on my machine, it passes:
image

Anyway I've fixed an issue with passing no options, and added a warning for passing only numeric value. Plus some trivial test cases to cover that. I didn't quite understand why setting bearing changes transform zoom value, but since it seems like it's out of scope of this PR I have adapted tests.

Wasn't sure about perf benchmarks in this case, seems unnecessary

dentuzhik:
In regards to shorter way of writing, decided to keep it unchanged in favour of adding a warning about bearing being numeric, hope it's ok @mourner

mvriel:
An interesting observation I made when testing with workarounds in our product is that outcome of the calculations in `_cameraForBoxAndBearing` are not correct when a bearing and pitch is provided. With a bearing of 0 and a bound determined with 2 LatLng coordinates it exactly fits in the viewport but when you apply a bearing it frequently happens that the coordinates are outside of the viewport.

Same goes when you have a pitch; the pitch makes the bounding box a trapezium shape instead of a rectangle and the `_cameraForBoxAndBearing` does not (and cannot) compensate for this

mvriel:
To give an idea of the errors that we encounter:

image

dentuzhik:
@mvriel it seems that it deserves a separate issue and PR. Zoom level changes based on bearing change also seems suspicious, but let's see what @mourner says

asheemmamoowala:
> Same goes when you have a pitch; the pitch makes the bounding box a trapezium shape instead of a rectangle and the _cameraForBoxAndBearing does not (and cannot) compensate for this

`_cameraForBoxAndBearing` does not take into account the pitch when computing the camera. This is pending implementation.

asheemmamoowala:
Closing this as stale. @dentuzhik feel free to re-open the PR when you've had a chance to address the issues mentioned at https://github.com/mapbox/mapbox-gl-js/pull/7618#pullrequestreview-177940972

spencerthayer:
Has this been resolved?

PR #7626, opened 2018-11-27
"number-format" expression

### LINKED ISSUE #4119: Add expression operator for formatting numbers
_From @1ec5 on May 6, 2016 23:21_

It’s possible for a style to [localize string labels](https://www.mapbox.com/help/change-language/) by choosing a language-specific field in the source. But when it comes to numeric types (elevations of peaks, for instance), the viewer is stuck with an unformatted number of the form ####.## using European digits. Ideally, it should be possible for a topo map to show elevations in #,###.## form for English, in #&nbsp;###,## form for French, and in #,##,###.## form for various Indic languages. Meanwhile, an Arabic or Gujarati map would use the appropriate numeral system.

I’m not sure whether the number format should be controlled by the style (and thus defined in this specification) or determined by the system locale (and thus be implementation-specific). But I figured I’d start here because the conventional approach to localized strings involves changing style properties.

/cc @mikemorris @kkaefer @lucaswoj

_Copied from original issue: mapbox/mapbox-gl-style-spec#452_

ChrisLoer:
Add "number-format" expression, to fix #4119. Example:

`["number-format", 1234.5678, { "currency": "EUR" } ]` -> `"€ 1,234.56"`

Format options are:

- "locale": BCP 47 language tag, defaults to current locale. May or may not succeed in loading the locale, currently the only way to inspect this is to use the `resolved-locale` expression on a collator.
- "currency": If set, ISO 4217 currency code, used for currency formatting. Returned string includes a symbol representing the currency (e.g. "JP¥" or "€").
- "min-fraction-digits": Formatter will include at least this many digits after the decimal point. Defaults to 0.
- "max-fraction-digits": Formatter will include at most this many digits after the decimal point. Defaults to 3.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/style-spec/expression/definitions/number_format.js
+++ b/src/style-spec/expression/definitions/number_format.js
@@ -1,0 +1,142 @@
+// @flow
+
+import { StringType, NumberType } from '../types';
+
+import type { Expression } from '../expression';
+import type EvaluationContext from '../evaluation_context';
+import type ParsingContext from '../parsing_context';
+import type { Type } from '../types';
+
+declare var Intl: {
+    NumberFormat: Class<Intl$NumberFormat>
+};
+
+declare class Intl$NumberFormat {
+    constructor (
+        locales?: string | string[],
+        options?: NumberFormatOptions
+    ): Intl$NumberFormat;
+
+    static (
+        locales?: string | string[],
+        options?: NumberFormatOptions
+    ): Intl$NumberFormat;
+
+    format(a: number): string;
+
+    resolvedOptions(): any;
+}
+
+type NumberFormatOptions = {
+    style?: 'decimal' | 'currency' | 'percent';
+    currency?: null | string;
+    minimumFractionDigits?: null | string;
+    maximumFractionDigits?: null | string;
+};
+
+export default class NumberFormat implements Expression {
+    type: Type;
+    number: Expression;
+    locale: Expression | null;   // BCP 47 language tag
+    currency: Expression | null; // ISO 4217 currency code, required if style=currency
+    minFractionDigits: Expression | null; // Default 0
+    maxFractionDigits: Expression | null; // Default 3
+
+    constructor(number: Expression,
+                locale: Expression | null,
+                currency: Expression | null,
+                minFractionDigits: Expression | null,
+                maxFractionDigits: Expression | null) {
+        this.type = StringType;
+        this.number = number;
+        this.locale = locale;
+        this.currency = currency;
+        this.minFractionDigits = minFractionDigits;
+        this.maxFractionDigits = maxFractionDigits;
+    }
+
+    static parse(args: Array<mixed>, context: ParsingContext): ?Expression {
+        if (args.length !== 3)
+            return context.error(`Expected two arguments.`);
+
+        const number = context.parse(args[1], 1, NumberType);
+        if (!number) return null;
+
+        const options = (args[2]: any);
+        if (typeof options !== "object" || Array.isArray(options))
+            return context.error(`NumberFormat options argument must be an object.`);
+
+        let locale = null;
+        if (options['locale']) {
+            locale = context.parse(options['locale'], 1, StringType);
+            if (!locale) return null;
+        }
+
+        let currency = null;
+        if (options['currency']) {
+            currency = context.parse(options['currency'], 1, StringType);
+            if (!currency) return null;
+        }
+
+        let minFractionDigits = null;
+        if (options['min-fraction-digits']) {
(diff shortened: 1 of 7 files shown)


ChrisLoer:
This would fix https://github.com/mapbox/mapbox-gl-js/issues/4119

ChrisLoer:
cc @chloekraw and @mapbox/maps-design -- I know this isn't near the top of our prioritized list, but it's potential low-hanging fruit and could be pretty useful for displaying contextual information, especially with POIs.

Caveat on "low-hanging fruit" claim: just like the collator expressions, this requires us to come up with uniform behavior across multiple differing platform implementations, so there are pitfalls. The good news is that unlike the collator work, no further changes to the expression language itself are necessary.

src/style-spec/expression/definitions/number_format.js
```diff
@@ -0,0 +1,114 @@
...
+    style?: 'decimal' | 'currency' | 'percent';
+    currency?: null | string;
+};
+
+export default class NumberFormat implements Expression {
+    type: Type;
+    number: Expression;
+    style: Expression;
+    locale: Expression | null;   // BCP 47 language tag
+    currency: Expression | null; // ISO 4217 currency code, required if style=currency
```
1ec5: > `Intl.NumberFormat` requires this in "currency" mode, it doesn't just guess based on the locale

Yeah, that’s the right approach. Operating systems allow you to choose the preferred currency independently of the system region. By default, the system region typically affects mechanics like decimal separators, but not semantics or content.

1ec5: Here’s a correspondence to other implementation languages:

Style specification | Objective-C
----|----
`NumberFormat.style` | `NSNumberFormatter.numberStyle`
`NumberFormat.locale` | `NSNumberFormatter.locale`
`NumberFormat.currency` | `NSNumberFormatter.currencyCode`

In Java, use `NumberFormat.get{Currency,Number,Percent}Instance()` and `NumberFormat.setCurrency()`.

src/style-spec/expression/definitions/number_format.js
```diff
@@ -0,0 +1,114 @@
...
+type NumberFormatOptions = {
+    style?: 'decimal' | 'currency' | 'percent';
+    currency?: null | string;
+};
+
+export default class NumberFormat implements Expression {
+    type: Type;
+    number: Expression;
+    style: Expression;
+    locale: Expression | null;   // BCP 47 language tag
```
1ec5: Would we support any extensions, such as `zh-Hans-CN-u-nu-hanidec` to specify Chinese decimal numbering? Perhaps it would be an “if it works, it works” situation, with no availability guarantees from one platform to another.

src/style-spec/expression/definitions/number_format.js
```diff
@@ -0,0 +1,114 @@
...
+        options?: NumberFormatOptions
+    ): Intl$NumberFormat;
+
+    format(a: number): string;
+
+    resolvedOptions(): any;
+}
+
+type NumberFormatOptions = {
+    style?: 'decimal' | 'currency' | 'percent';
```
1ec5: `NSNumberFormatterStyle` actually supports a number of styles, including `NSNumberFormatterSpellOutStyle` and `NSNumberFormatterOrdinalStyle`. However, I recognize the need to stick to JavaScript’s `Intl` as a common denominator, without implementing our own custom types that could conflict with future JavaScript types.

This does mean that the style author will need to hard-code some logic to implement [short formats](https://www.unicode.org/cldr/charts/26/by_type/numbers.compact_decimal_formatting.html#Short_Formats_), such as the “1k” for 1,000 that Supercluster implements (only for English) in the `point_count_abbreviated` property (mapbox/supercluster#110).

/ref tc39/ecma402#37 tc39/ecma402#215.

src/style-spec/expression/definitions/number_format.js
ChrisLoer: > Perhaps it would be an “if it works, it works” situation, with no availability guarantees from one platform to another.

That sounds right to me.

ChrisLoer:
I think this is ready. The tests may break if run on a machine that doesn't have a US locale installed, which makes me feel squirmy, although I think so far that has worked out OK with the collator tests. If it becomes a problem we can make the tests more verbose (check for "resolved-locale" and just pass the test the locale isn't available).

@asheemmamoowala would you have a chance to review?

Also tagging @mapbox/studio b/c I don't think they got the original notification.

ChrisLoer:
> Would there be a need to use data-driven styling for the currency field?

I could imagine DDS being used for currency in something like a map showing hotel costs in local currencies, where the currency to use would depend on data in the map.

PR #7628, opened 2018-11-28
Add narrow Y (height < tile size) render tests

brunoabinader:
This helps to ensure GL native renders the same non-standard aspect ratios.


brunoabinader:
> Looks good, although I'd generally prefer for new render tests to be smaller when possible because every new one slows down the suite. E.g. 1024x480 could maybe be 512x240?

Sure - I'm going to update it :+1: thanks @mourner!

brunoabinader:
> Looks good, although I'd generally prefer for new render tests to be smaller when possible because every new one slows down the suite. E.g. 1024x480 could maybe be 512x240?

On a second thought, having a repeated map being shown is part of the test verification. This is only achievable when e.g. the width is bigger than tile size (>512). We could, however, leave only `basic-v9/z0-narorw-y` and remove duplicates in `satellite-v9` and `bright-v9`. Does that sounds good @mourner ?

mourner:
@brunoabinader yes, if they don't test different things. I'd also give preference to a narrower test, e.g. 600x128 maybe?

PR #7633, opened 2018-11-28
merge mb-pages into master

ansis:
Merge `mb-pages` into `master`


ryanhamley:
I think this should be squashed into a single commit before merging. From the [release process doc](https://github.com/mapbox/gl-internal/wiki/Javascript-release-process): 

>If you have more than the one commit on sync-mb-pages, you can rebase against master to squash them into a single commit, but you must use the --preserve-merge option. This commit should have two parents (master and mb-pages) if everything went correctly.

ansis:
@ryanhamley I was wondering about that line as well but I think this branch is as expected. There is just one new commit on `sync-mb-pages` (the merge commit). The rest already exist in mb-pages but show up here because they aren't in master.

the previous releases' prs look the same https://github.com/mapbox/mapbox-gl-js/pull/7515, https://github.com/mapbox/mapbox-gl-js/pull/7239 and https://github.com/mapbox/mapbox-gl-js/pull/6937 look the same.

ryanhamley:
Ah ok, 👍 then!

PR #7641, opened 2018-11-29
advance image request queue when request aborted

### LINKED ISSUE #7637: raster tiles fail to load in safari

**browser**: Safari on mac

### Steps to Trigger Behavior

 1. open debug/satellite.html
 2. wait for the tiles load
 3. zoom out and witness tiles failing to load

### Link to Demonstration

http://localhost:9966/debug/satellite.html#12.5/38.888/-77.01866

### Expected Behavior

Tiles load.

### Actual Behavior

Tiles don't load.

ansis:
fixes #7637

Aborted requests never decremented the number of active requests because the main callback is never called. This fixes that by making `getImage` return a wrapped version of the `getArrayBuffer` return value that also advances the queue.

Within `advanceImageRequestQueue` we need to change that it hasn't already been advanced for that request because sometimes `cancel()` is called after the request has completed.

## Launch Checklist

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -4,6 +4,7 @@ import window from './window';
 import { extend } from './util';
 import { isMapboxHTTPURL } from './mapbox';
 import config from './config';
+import assert from 'assert';
 
 import type { Callback } from '../types/callback';
 import type { Cancelable } from '../types/cancelable';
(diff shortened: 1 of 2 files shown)


ansis:
@mourner can I get a review of the new commit?

Issue #7639, opened 2018-11-29
symbol flickering related to wrapped tiles


ansis:
Symbols occasionally flicker at the lowest zoom level, likely related to wrapping.

**mapbox-gl-js version**: master and v0.51

**browser**: chrome

### Steps to Trigger Behavior

 1. open https://www.mapbox.com/mapbox-gl-js/example/geojson-markers/
 2. wait until it loads
 3. quickly zoom out to the lowest zoom

### Link to Demonstration

https://www.mapbox.com/mapbox-gl-js/example/geojson-markers/

### Expected Behavior

no flickering

### Actual Behavior

flickering

---

cc @ChrisLoer 

I'm not marking this as a release blocker because it is not that significant and because it affects the previous version.

ChrisLoer:
This looks like a pretty simple bug:

https://github.com/mapbox/mapbox-gl-js/blob/16a0c1de245e91a1a779829629d11945af560b51/src/source/tile_id.js#L94-L101

`isChildOf` doesn't account for wrap, so the cross tile symbol index picks up "child" tiles in different world copies. Looks like we have the same problem on native.

ChrisLoer:
Fixed in #7644.

PR #7644, opened 2018-11-29
OverscaledTileID#isChildOf shouldn't include tiles with different wrap.

ChrisLoer:
Fixes issue #7639, which happened because the `CrossTileSymbolIndex` would incorrectly treat a tile in a different world copy as a child, causing two different symbols to share the same `crossTileID`.

## Launch Checklist

 - [ ] post benchmark scores

cc @ansis


ChrisLoer:
I didn't see anything jump out in the benchmark scores. I'd imagine this is a pretty small cost although spread all over the place.

PR #7645, opened 2018-11-29
Add check for layer 'type' property

ryanhamley:
## Launch Checklist

    - filters out layers using the deprecated `ref` property. this check is necessary to avoid a crash in certain tests that check for the `type` property that older layers lack. (this used to check for `layer.ref` but a PR that allows for using local style JSON files in the style benchmarks changed the Flow type of `layer` so that checking for `layer.ref` explicitly will cause a Flow error. inverting the check achieves the same goal while avoiding the Flow issue)


ryanhamley:
I added a comment. I'll squash when I merge the PR.

PR #7655, opened 2018-11-30
Fix images not being aborted when dequeued

### LINKED ISSUE #7650: image requests that used to be queued can't be aborted
Image requests that are queued return a Cancelable object that only cancels the request if the request is still queued. Once the request has been unqueued it is not possible to abort the network request. Calling `cancel` at that point will do nothing.

https://github.com/mapbox/mapbox-gl-js/blob/16a0c1de245e91a1a779829629d11945af560b51/src/util/ajax.js#L186

@mourner I don't think this should block the beta because it is already in the previous version. should it block the final release?

**mapbox-gl-js version**: v0.50 and master

**browser**: all

### Steps to Trigger Behavior

 1. open http://localhost:9966/debug/satellite.html#7.85/58.009/-113.643 in large viewport to queue up image requests
 2. wait a bit so that some requests complete and queue requests start
 3. zoom quickly to abort requests

### Link to Demonstration

http://localhost:9966/debug/satellite.html#7.85/58.009/-113.643

### Expected Behavior

Requests for previously queued images abort.

### Actual Behavior

Requests for previously queued images do not abort.

mourner:
Theoretically fixes #7650. cc @ansis 

## Launch Checklist

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -190,9 +190,14 @@ resetImageRequestQueue();
 export const getImage = function(requestParameters: RequestParameters, callback: Callback<HTMLImageElement>): Cancelable {
     // limit concurrent image loads to help with raster sources performance on big screens
     if (numImageRequests >= config.MAX_PARALLEL_IMAGE_REQUESTS) {
-        const queued = {requestParameters, callback, cancelled: false};
+        const queued = {
+            requestParameters,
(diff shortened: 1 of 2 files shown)


ansis:
Looks good. Do you think it would be possible to add a unit test for this?

PR #7662, opened 2018-12-03
Clarify docs that branch labels must be unique in `match` expression

chloekraw:
PR motivated by a user-reported bug that we investigated, through which we determined that the parsing of a `match` expression was failing because the branches were not unique. Clarifies this requirement in https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-match

## Launch Checklist

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2390,7 +2390,7 @@
         }
       },
       "match": {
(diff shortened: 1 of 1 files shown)


radhikavm:
Hello,
I am getting this error: `layers.postcodes.filter[2]: Branch labels must be unique.` I am using match expression :
`map.setFilter('postcodes', ['match', ['get', 'name'], highlighting, true, false]);`

I may be probably missing something obvious but can some please explain what exactly does that mean with reference to the above context.

PR #7664, opened 2018-12-03
Include bench directory in package files published

tristen:
Not sure what the consequences are including the `bench/` in the files array but I would like to consume it for another project so I'm proposing a change that its added to the published dependency.


tristen:
👍

Issue #7647, opened 2018-11-30
Video source repaint after seeked


dwilson2547:
## Motivation

I need the ability to set or update the currentTime of a video in mapbox-gl while the video is paused. As the video source is written now the video source is only painted while the video is playing, so after pausing and changing the time, no visible change is seen in mapbox-gl.

## Design Alternatives

You could add another event listener to the video for the "seeked" event and when that event fires, set a flag to repaint the video which is then removed after being painted once.

## Design

Potential issues: The hasTransition() and prepare() functions might not always execute in the correct order due to a race condition, one additional flag may be needed to ensure that prepare() is run first, then hasTransition()

### Mock-Up

Potential solution, see patch attached to this request or the demo [here](https://dwilson2547.github.io/mapbox-video-controls-example/index.html)

### Concepts

This shouldn't affect the existing functionality of the video source, I'm hoping to add to it in such a way that no one has to modify existing code using the video source.

### Implementation

[videofix.txt](https://github.com/mapbox/mapbox-gl-js/files/2631392/videofix.txt)

asheemmamoowala:
@dwilson2547 Thank you for using Mapbox! We'd happy to accept a patch if you could open a formal PR with your changes for review.

hayley-starr:
@asheemmamoowala @dwilson2547 This functionality would be really useful. Right now I am trying to get the frame to repaint with map.triggerRepaint() to no avail. I want to be able to see the video frames change as the user drags the video scrubber. Any workaround right now?

PR #7665, opened 2018-12-04
Video source update

dwilson2547:
## Launch Checklist

Potential fix for issue #7647 this would allow for a video to be redrawn once it has been seeked to a different time regardless of whether the video is playing or not.

 - [ ] post benchmark scores

### DIFF
--- a/src/source/video_source.js
+++ b/src/source/video_source.js
@@ -181,7 +181,7 @@ class VideoSource extends ImageSource {
     }
 
     hasTransition() {
-        var result = this.video && (!this.video.paused || this.videoSeeked);
+        let result = this.video && (!this.video.paused || this.videoSeeked);
         if (this.videoSeeked && this.textureDrawn) {
             this.videoSeeked = false;
             this.textureDrawn = false;
--- a/src/source/video_source.js
+++ b/src/source/video_source.js
@@ -181,7 +181,7 @@ class VideoSource extends ImageSource {
     }
 
     hasTransition() {
-        let result = this.video && (!this.video.paused || this.videoSeeked);
+        const result = this.video && (!this.video.paused || this.videoSeeked);
         if (this.videoSeeked && this.textureDrawn) {
             this.videoSeeked = false;
             this.textureDrawn = false;
--- a/src/source/video_source.js
+++ b/src/source/video_source.js
@@ -50,6 +50,7 @@ class VideoSource extends ImageSource {
     urls: Array<string>;
     video: HTMLVideoElement;
     roundZoom: boolean;
+    videoSeeked: boolean;
 
     /**
      * @private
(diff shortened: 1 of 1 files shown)


src/source/video_source.js
```diff
@@ -166,7 +181,12 @@ class VideoSource extends ImageSource {
     }
 
     hasTransition() {
-        return this.video && !this.video.paused;
+        const result = this.video && (!this.video.paused || this.needsUpdate);
+        if (this.needsUpdate && this.textureDrawn) {
```
ryanhamley: This check seems superfluous. If these variables are already `false`, setting them to `false` won't change anything and if they're `true` then they get set to `false` anyway.

ryanhamley: In fact, we could just do

```
this.needsUpdate = false;
this.textureDrawn = false;
return this.video && (!this.video.paused || this.needsUpdate);
```

src/source/video_source.js
```diff
@@ -143,11 +154,15 @@ class VideoSource extends ImageSource {
...
             this.texture = new Texture(context, this.video, gl.RGBA);
             this.texture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
-        } else if (!this.video.paused) {
+        } else if (!this.video.paused || this.needsUpdate) {
             this.texture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
             gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, this.video);
         }
 
+        if (this.needsUpdate) {
+            this.textureDrawn = true;
```
ryanhamley: What is this property for? It seems like it's only used in the check in `hasTransition`?

src/source/video_source.js
dwilson2547: Hey Ryan, in the first update I made to the code i just added what's now the needsUpdate variable and it worked most of the time, but i noticed that occasionally the video wouldn't be rendered after skipping. I assumed that this was due to a race condition between the prepare and hasTransition functions where the hasTransition function would be called after the video was seeked but before the prepare function. This would set the needsUpdate variable back to false before reaching the prepare function so it wouldn't draw the texture for the updated position of the video.  The textureDrawn variable was added to prevent this race condition so I could guarantee that the updated texture has been drawn before the hasTransition function is called so the painter has an updated texture to draw. Once that's complete, the needsUpdate and textureDrawn variables can be set back to false so the painting can resume normally

ryanhamley: Hey @dwilson2547 did you notice this in a specific environment or anything like that? I've been playing with the code and I see it working (at least in Firefox) as expected with no changes to `hasTransition` at all. I've tried quite a few times and haven't been able to produce a race condition where the video fails to update when skipping forward. Is there a specific way to produce the bug?

`needsUpdate` is necessary in lines 92 and 157, but I can't see why it's necessary in `hasTransition` or why `textureDrawn` is necessary at all.

dwilson2547: @ryanhamley I haven't been able to purposely reproduce the bug, I actually didn't notice it in my first few days of testing. For me it happened about once in every 50 tries, I can't definitively prove that it was a race condition where hasTransition was run before the prepare function, but that was the only thing that made sense to me. I did my testing in chrome and firefox and I saw the bug 3 times. I checked the console for errors and never saw any so I'm fairly certain it wasn't just a slow buffering issue

asheemmamoowala:
Closing this PR as stale. @dwilson2547 feel free to re-open once you've addressed the remaining review feedback.

PR #7669, opened 2018-12-05
removeFeatureState() (not ready)

peterqliu:
Addresses https://github.com/mapbox/mapbox-gl-js/issues/6889

Depending on parameters specified, enables feature state removal at the source, feature, and property levels.

screen shot 2018-12-11 at 1 35 34 pm
screen shot 2018-12-11 at 1 36 02 pm
screen shot 2018-12-11 at 1 39 01 pm

Effect on `setFeatureState` perf:

screen shot 2018-12-10 at 4 20 54 pm

cc/ @asheemmamoowala 

 - [x] post benchmark scores


peterqliu:
Just realized that the `removeFeatureState` example above removes state at the _source_ level. Doing the removal at the feature or prop level is much faster, and comparable to the previous `setFeatureState` (to default) in performance. Will investigate to see if we can optimize the source-wide removal further.

Also noticed that the previously hovered features reappear when I zoom in/out, suggesting that cached tiles get the `set` but not the `remove` action. @asheemmamoowala can you point me to what I might be missing?

peterqliu:
Turns out feature state removal is slow only if:

- it removes state at the source level, and
- many features in the source already have some feature state set (though not necessarily the prop being removed)

Since mouse interactivity is probably the most performance sensitive use case, and it rarely requires making changes to more than one or two features at a time, I think this situational slowness is acceptable.

src/source/source_state.js
```diff
@@ -22,11 +22,42 @@ class SourceFeatureState {
...
+            this.deletedStates[sourceLayer][feature] = {};
+            for (key in this.state[sourceLayer][feature]) this.deletedStates[sourceLayer][feature][key] = null;
+        }
+
+        else  {
+            this.deletedStates[sourceLayer] = {};
+            for (feature in this.state[sourceLayer]) {
+                this.deletedStates[sourceLayer][feature] = {};
+                for (key in this.state[sourceLayer][feature]) {
+                    this.deletedStates[sourceLayer][feature][key] = null;
```
asheemmamoowala: Is it necessary to iterate over all features/keys here? Can't presence of an empty `this.deleteStates[sourceLayer]` object be used to indicate that all items should be deleted? Similarly for all items for a specific feature id ?

src/source/source_state.js
```diff
@@ -22,11 +22,42 @@ class SourceFeatureState {
...
+        extend(this.stateChanges[sourceLayer][feature], newState);
+
+        const deletionInQueue = this.deletedStates && this.deletedStates[sourceLayer] && this.deletedStates[sourceLayer][feature]
+        if (deletionInQueue) delete this.deletedStates[sourceLayer][feature];
+    }
+
+    removeFeatureState(sourceLayer: string, featureId: number, key: string) {
+        const feature = String(featureId);
+
+        this.deletedStates = {};
```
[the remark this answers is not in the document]
peterqliu: 👍 it's usually `null` when this line evaluates, but not if we have several `rFS`'s queued up

src/source/source_state.js
peterqliu: 🤔it appears necessary, because in the coalesce step `layerStates` needs to iterate over features from `deletedStates` to populate itself: https://github.com/mapbox/mapbox-gl-js/blob/dbbf41cf26acd1903e8c39f313fd1979f37924b7/src/source/source_state.js#L98

In turn, layerStates is used to extend `currentChanges`

peterqliu: If I understand correctly, this whole workflow works by an "absence of evidence is not evidence of absence" philosophy, which necessitates the explicit `null`s

src/source/source_state.js
```diff
@@ -16,50 +16,142 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
...
+
+
+        if (this.deletedStates && Object.keys(this.deletedStates).length > 0) {
+
+
+            for (const sourceLayer in this.deletedStates) {
+                this.state[sourceLayer]  = this.state[sourceLayer] || {};
+                const layerStates = {};
+                for (const feature in this.deletedStates[sourceLayer]) {
+
```
asheemmamoowala: nit: empty line(s)

peterqliu:
➡️ https://github.com/mapbox/mapbox-gl-js/pull/7669

Issue #6748, opened 2018-05-29
style option must be provided in Map constructor


jingsam:
Currently, the style option is optional in Map constructor. Initializing a map without style may be useless. And also `addLayer`, `addSource`, etc don't check the existence of style, which depend on a valid style. So should we check the style option in Map constructor, or use a empty style  as the default style option?

jfirebaugh:
It should default to an empty style, for consistency with native.

PR #7672, opened 2018-12-05
set default map style option to an empty style #6748

stepankuzmin:
This PR sets the default map style option to an empty style according to https://github.com/mapbox/mapbox-gl-js/issues/6748
I'm not sure about hardcoding `version`, maybe there is a more convenient way?

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -98,6 +98,11 @@ type MapOptions = {
 const defaultMinZoom = 0;
 const defaultMaxZoom = 22;
 const defaultOptions = {
+    style: {
+        version: 8,
+        sources: {},
+        layers: []
+    },
     center: [0, 0],
     zoom: 0,
     bearing: 0,


stepankuzmin:
I've noticed, that unit test that checks if the load event is emitted after a style is set is now broken because a style is now always set.

```js
t.test('emits load event after a style is set', (t) => {
    t.stub(Map.prototype, '_detectMissingCSS');
    const map = new Map({ container: window.document.createElement('div') });

    map.on('load', fail);

    setTimeout(() => {
        map.off('load', fail);
        map.on('load', pass);
        map.setStyle(createStyle());
    }, 1);

    function fail() { t.ok(false); }
    function pass() { t.end(); }
});
```

Should I rewrite this test, or it should work as conceived?

andrewharvey:
I've :+1: for this change.

> Should I rewrite this test, or it should work as conceived?

I would say rewrite the test, as if the style would now have a default empty style then map.on('load') can be triggered even without setStyle.

vakila:
Closing this out as it has moved to https://github.com/mapbox/mapbox-gl-js/pull/8924, thanks @stepankuzmin @arindam1993 !

PR #7680, opened 2018-12-06
[docs]Add mapbox events domain to CSP directives

asheemmamoowala:
In #6980 and #7431, new events POST to `https://events.mapbox.com`. This domain needs to be whitelisted in the CSP directives to allow event requests to go through.

cc @springmeyer @mollymerp


andrewharvey:
PS. The Map Embed API needs this fixed too https://api.mapbox.com/styles/v1/mapbox/streets-v9.html

PR #7683, opened 2018-12-07
Use 1px border for DEM backfilling

andycalder:
This PR removes the empty 256 pixels of padding around DEM data. This padding means that the DEMData array buffer is ~4x as big as it needs to be, and we're uploading a ~4x demTexture to the GPU. A one pixel border is all we need for DEM backfilling. The demTexture is not mipmapped so I don't think it needs to be a power-of-two texture. A quick look at Chrome Dev Tools shows a big drop in memory usage on the hillshade debug page (119MB to 47MB). Benchmarks don't show anything notable.

## Launch Checklist

 - [x] post benchmark scores

paint
layout
layerhillshade

### DIFF
--- a/src/data/dem_data.js
+++ b/src/data/dem_data.js
@@ -1,7 +1,7 @@
 // @flow
 import { RGBAImage } from '../util/image';
 
-import { warnOnce, clamp } from '../util/util';
+import { warnOnce } from '../util/util';
 import { register } from '../util/web_worker_transfer';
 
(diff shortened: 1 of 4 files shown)


mollymerp:
yes – we initially were going to use mipmapping for the "openness" visualization, but had to scrub that because of hardware issues on some devices but the original power-of-two constraint snuck in. I don't think this should cause any problems.

I have ported this over to gl-native here https://github.com/mapbox/mapbox-gl-native/pull/13550

thanks!

PR #7687, opened 2018-12-11
avoid incomplete webp support in Edge 18

### LINKED ISSUE #7671: Studio Map won't render in Edge 2018 Release
**mapbox-gl-js version**: 0.51.0

**browser**: Microsoft Edge October 2018 (
Microsoft Edge 44.17763.1.0, Microsoft EdgeHTML 18.17763, and Microsoft Windows Version 10.0.17763 Build 17763.

### Steps to Trigger Behavior

 1. Go to https://jsbin.com/yakopis/edit?html,js,output in Chrome
 2. Map renders okay
 3. Go to https://jsbin.com/yakopis/edit?html,js,output in Edge
 4. Unable to render map fully.

### Link to Demonstration
https://jsbin.com/yakopis/edit?html,js,output

### Expected Behavior
Map would work the same in Edge as other browsers

### Actual Behavior
Map is nonfunctional in Edge Oct 2018.
edgeerror

ansis:
Edge 18 supports WebP but not uploading WebP images to gl textures. Avoid this bug by testing support for this before enabling WebP loading.

fix #7671 

The one downside of this approach is that once per page load we create a gl context and throw it away. I haven't seen this be a problem though. We could wait until a map is initialized and use that context but that would be more complex.

## Launch Checklist

We can't add automated testing because we don't do any of that in Edge. Manually testing the satellite map example does test this.

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -25,7 +25,7 @@ import { RGBAImage } from '../util/image';
 import { Event, ErrorEvent } from '../util/evented';
 import { MapMouseEvent } from './events';
 import TaskQueue from '../util/task_queue';
-import testWebp from '../util/test_webp';
+import webpSupported from '../util/webp_supported';
 
 import type {PointLike} from '@mapbox/point-geometry';
 import type {LngLatLike} from '../geo/lng_lat';
@@ -1541,7 +1541,7 @@ class Map extends Camera {
 
         this.painter = new Painter(gl, this.transform);
 
-        testWebp(gl);
+        webpSupported.testSupport(gl);
     }
 
     _contextLost(event: *) {
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -25,6 +25,7 @@ import { RGBAImage } from '../util/image';
 import { Event, ErrorEvent } from '../util/evented';
 import { MapMouseEvent } from './events';
 import TaskQueue from '../util/task_queue';
+import testWebp from '../util/test_webp';
 
 import type {PointLike} from '@mapbox/point-geometry';
 import type {LngLatLike} from '../geo/lng_lat';
@@ -1539,6 +1540,8 @@ class Map extends Camera {
         }
 
         this.painter = new Painter(gl, this.transform);
+
+        testWebp(gl);
     }
 
     _contextLost(event: *) {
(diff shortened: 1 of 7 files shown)


ansis:
I reworked the implementation to avoid creating a context and instead using the one already created for a map. That is slightly better but the implementation is more complex in order to properly deal with asynchness.

> @ansis This approach makes every browser pay the price for the check that only the Edge browser really needs.

Yep, the goal is to avoid user agent sniffing and instead check for the capability.

> could this supportsWebp be exported as a function and the support value cached after the first call?

In the first implementation the function gets run as soon as the mapbox-gl-js script gets loaded (once per page load). In the new implementation it gets run on map creation (to avoid creating it's own context). Is this what you meant?

> It would also mean that if there are any issues with webgl context creation, they do not surface in this code.

Yep, this needs to change! I've changed it in the latest commit.

ansis:
@ChrisLoer would it be possible to get a gut review from you? Does https://github.com/mapbox/mapbox-gl-js/pull/7687/commits/e88ecae897dbe766a68d89fe8e98d30cba449daf or the current HEAD of the branch seem like a better way to check for webp support?

asheemmamoowala:
> In the first implementation the function gets run as soon as the mapbox-gl-js script gets loaded (once per page load). In the new implementation it gets run on map creation (to avoid creating it's own context). Is this what you meant?

I was thinking more that it would be a function that is invoked when needed by the usage in `mapbox.js#normalizeTileURL`. This method, from what I can tell is called from `loadTIle` on the main thread before passing to a worker. So there shouldn't be any synchronization issue with the way we use it.

ChrisLoer:
> Does e88ecae or the current HEAD of the branch seem like a better way to check for webp support?

I don't have a strong opinion. Can you write a tiny benchmark to get an idea of the cost of the create context and texture upload check? My guess would be it's pretty trivial, in which case we might as well go with the e88ecae approach just because it's simpler. But the "do it with the gl context provided by setupPainter" approach doesn't seem too bad to me if there's any measurable performance win. Is there some tiny chance that _while_ `webpImgTest` is loading, the GL context could be lost/restored, and then the test would fail from having stored the old context?

ansis:
@asheemmamoowala I think I like keeping the test method separate from the way of accessing the result because otherwise we'd have to pass a gl context through to normalizeTileURL which seems a bit more invasive.

@ChrisLoer It looks like context creation is measurable but not that expensive. So keeping this approach? I added a context loss check to avoid that bug.

PR #7696, opened 2018-12-12
Remove feature state (very maybe possibly ready)

peterqliu:
retry of https://github.com/mapbox/mapbox-gl-js/pull/7669

-----

Addresses https://github.com/mapbox/mapbox-gl-js/issues/6889

Depending on parameters specified, enables feature state removal at the source, feature, and property levels.

screen shot 2018-12-11 at 1 35 34 pm
screen shot 2018-12-11 at 1 36 02 pm
screen shot 2018-12-11 at 1 39 01 pm

Effect on `setFeatureState` perf:

screen shot 2018-12-10 at 4 20 54 pm

cc/ @asheemmamoowala 

 - [x] post benchmark scores


src/ui/map.js
```diff
@@ -25,7 +25,6 @@ import { RGBAImage } from '../util/image';
 import { Event, ErrorEvent } from '../util/evented';
 import { MapMouseEvent } from './events';
 import TaskQueue from '../util/task_queue';
-import webpSupported from '../util/webp_supported';
```
ChrisLoer: Looks like the recent `webpSupported` changes got accidentally stripped, probably on a rebase or something?

peterqliu: whoops yep, the original PR was branched off an earlier `master`, and i copied it straight from there. fixing

src/ui/map.js
```diff
@@ -1833,12 +1846,7 @@ class Map extends Camera {
      * @memberof Map
      */
     get repaint(): boolean { return !!this._repaint; }
-    set repaint(value: boolean) {
```
ChrisLoer: I think this repaint change and also the `isMoving` change above are also unintended?

test/unit/ui/map.test.js
```diff
@@ -1518,30 +1728,6 @@ test('Map', (t) => {
         });
     });
 
-    t.test('no render after idle event', (t) => {
```
ChrisLoer: These tests also look accidentally removed.

src/source/source_state.js
```diff
@@ -16,50 +16,144 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
...
+        }
     }
 
     getState(sourceLayer: string, featureId: number) {
         const feature = String(featureId);
         const base = this.state[sourceLayer] || {};
         const changes = this.stateChanges[sourceLayer] || {};
-        return extend({}, base[feature], changes[feature]);
+        const deletions = this.deletedStates && this.deletedStates[sourceLayer] ? this.deletedStates[sourceLayer] : {};
+        return extend({}, base[feature], changes[feature], deletions[feature]);
```
peterqliu: @asheemmamoowala running into a surprising snag here when trying to delete state keys instead of setting them to null. The getState() method essentially has to run a coalesce of its own to ensure that it's returning a reconciled version of state, updates, and deletions. Deletions are represented by `null`s, and applied to the state via `extend()`. 

Unfortunately, the `const` syntax (immutable?) is preventing me from deleting the nulls from the `extend`ed object. I can declare another root-level variable alongside `state`, `stateChanges`, and `deletedStates`, or write a custom version of `extend()` that can actually delete stuff, but both of these feel like overkill

bench/benchmarks/remove_paint_state_feature.js
```diff
@@ -0,0 +1,76 @@
+
```
ryanhamley: The three benchmark tests here look to be almost identical with the exception of what's passed to `setFeatureState` and `removeFeatureState` in `bench`. You could probably create an overarching `remove_paint_state` benchmark that has sub-tests for each individual test, in the same vein as the Expressions and Layers benchmarks.

peterqliu: @ryanhamley  was looking to do just that! thanks for the tip, was able to crib the format from Layers. consolidated the benches here https://github.com/mapbox/mapbox-gl-js/pull/7696/commits/d048d22d4a6f3224e811b31426333c78bdb2c718

src/source/source_state.js
```diff
@@ -96,8 +96,16 @@ class SourceFeatureState {
...
         const changes = this.stateChanges[sourceLayer] || {};
-        const deletions = this.deletedStates && this.deletedStates[sourceLayer] ? this.deletedStates[sourceLayer] : {};
-        return extend({}, base[feature], changes[feature], deletions[feature]);
+
+        const reconciledState = extend({}, base[feature], changes[feature]);
+
+        const propsToDelete = this.deletedStates && this.deletedStates[sourceLayer] && this.deletedStates[sourceLayer][featureId];
+
+        if (propsToDelete) {
+            for (const prop in this.deletedStates[sourceLayer][featureId]) delete reconciledState[prop];
```
asheemmamoowala: 👍

src/source/source_state.js
```diff
@@ -130,9 +128,16 @@ class SourceFeatureState {
                 this.state[sourceLayer]  = this.state[sourceLayer] || {};
                 const layerStates = {};
                 for (const feature in this.deletedStates[sourceLayer]) {
-                    for (const key in this.deletedStates[sourceLayer][feature]) {
-                        if (this.state[sourceLayer][feature][key]) this.state[sourceLayer][feature][key] = null;
+
+                    const featureProperties = Object.keys(this.deletedStates[sourceLayer][feature]);
```
asheemmamoowala: `Object.keys` will throw when `this.deletedStates[sourceLayer][feature] == null`. Needs a null check instead of the `length == 0` test in ln133-134

src/source/source_state.js
```diff
@@ -56,9 +56,7 @@ class SourceFeatureState {
             const updateInQueue = this.stateChanges[sourceLayer] && this.stateChanges[sourceLayer][feature];
 
             if (featureStateExists) {
-                for (key in this.state[sourceLayer][feature]) {
-                    this.deletedStates[sourceLayer][feature][key] = null;
-                }
+                this.deletedStates[sourceLayer][feature] = null;
```
asheemmamoowala: The same `null` treatment can be used at the `sourceLayer` level on ln70-ln80.

src/source/source_state.js
peterqliu: 👍 not caught by tests -- only surfaces when the remove happens on a separate coalesce from the original `set`

src/source/source_state.js
```diff
@@ -16,50 +16,147 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
...
+        for (const key in newState) {
+            const deletionInQueue = this.deletedStates && this.deletedStates[sourceLayer] && this.deletedStates[sourceLayer][feature] && this.deletedStates[sourceLayer][feature][key] === null;
+            if (deletionInQueue) {
+                delete this.deletedStates[sourceLayer][feature][key];
+            }
+        }
+    }
+
+    removeFeatureState(sourceLayer: string, featureId?: number, key?: string) {
+
```
asheemmamoowala: nit: unnecessary new lines

src/source/source_state.js
```diff
@@ -16,50 +16,147 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
...
+            if (updateInQueue) {
+                this.deletedStates[sourceLayer][feature] = {};
+
+                for (key in this.stateChanges[sourceLayer][feature]) {
+                    this.deletedStates[sourceLayer][feature][key] = null;
+                }
+            } else this.deletedStates[sourceLayer][feature] = null;
+
+        } else {
+            this.deletedStates[sourceLayer] = {};
```
asheemmamoowala: Will you be using the `null` shortcut for whole sourceLayer deletion as well?

src/source/source_state.js
```diff
@@ -16,50 +16,147 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
...
+        const featuresChanged: LayerFeatureStates = {};
+
         for (const sourceLayer in this.stateChanges) {
             this.state[sourceLayer]  = this.state[sourceLayer] || {};
             const layerStates = {};
-            for (const id in this.stateChanges[sourceLayer]) {
-                if (!this.state[sourceLayer][id]) {
-                    this.state[sourceLayer][id] = {};
+            for (const feature in this.stateChanges[sourceLayer]) {
+                if (!this.state[sourceLayer][feature]) this.state[sourceLayer][feature] = {};
```
asheemmamoowala: nit: use `{ .. }` blocks for single line `if..else` clauses

src/source/source_state.js
```diff
@@ -16,50 +16,147 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
...
+                extend(this.state[sourceLayer][feature], this.stateChanges[sourceLayer][feature]);
+                layerStates[feature] = this.state[sourceLayer][feature];
+            }
+            featuresChanged[sourceLayer] = layerStates;
+        }
+
+
+        if (this.deletedStates && Object.keys(this.deletedStates).length > 0) {
+
```
asheemmamoowala: nit: unnecessary newlines

test/unit/ui/map.test.js
```diff
@@ -1466,6 +1466,281 @@ test('Map', (t) => {
...
+                    "layers": []
+                }
+            });
+            map.on('load', () => {
+                map.setFeatureState({ source: 'geojson', id: 1}, {'hover': true, 'foo': true});
+                map.setFeatureState({ source: 'geojson', id: 2}, {'hover': true, 'foo': true});
+
+                map.removeFeatureState({ source: 'geojson'});
+
+                const fState1 = map.getFeatureState({ source: 'geojson', id: 1});
```
asheemmamoowala: After clearing all the state in the source/source layer, there should be no feature objects remaining. Ideally `fState1` returns `null`/`undefined`.

peterqliu: @asheemmamoowala `getFeatureState()` returns `{ }` before any state gets set, and I think it makes sense to return that after add+remove, for the sake of API symmetry

test/unit/ui/map.test.js
ChrisLoer: This is just a nit now, but it looks like these two tests changed location in the file. Let's put them back where they started so it's easy to tell in the diff that nothing's changed.

src/ui/map.js
```diff
@@ -1409,6 +1409,25 @@ class Map extends Camera {
...
+     * Removes feature state, setting it back to the default behavior. If only
+     * source is specified, removes all feature states of that source. If
+     * feature.id is also specified, removes all keys for that feature's state.
+     * If key is also specified, removes that key from that feature's state.
+     *
+     * @param {Object} feature Feature identifier. Feature objects returned from
+     * {@link Map#queryRenderedFeatures} or event handlers can be used as feature identifiers.
+     * @param {string | number} feature.id (optional) Unique id of the feature. Optional if key is not specified.
+     * @param {string} feature.source The Id of the vector source or GeoJSON source for the feature.
+     * @param {string} [feature.sourceLayer] (optional)  *For vector tile sources, the sourceLayer is
```
ChrisLoer: If someone's using a GeoJSON source, and they provide a `sourceLayer` here, it'll break, won't it? (it would be an easy mistake to make, after all it looks like you're supposed to put something, and it's super easy to mix up "style layer" and "source layer").

Could we change the interface in some way to make it harder to make the mistake? Or alternatively, we could just change the code in SourceCache to _always_ hardwire a source layer of `_geojsonLayer` for geojson sources, even if another argument is provided?

debug/highlightpoints.html
```diff
@@ -49,16 +49,21 @@
...
             if (!f.state.hover) {
-                map.setFeatureState(f, {'hover': true});
-                if (hoveredFeature) {
-                    map.setFeatureState(hoveredFeature, {'hover': false});
-                }
+                map.setFeatureState(f, {'hover': true, 'visited': true, 'visits':f.state.visits ? f.state.visits + 1 : 3});
                 hoveredFeature = f;
+
             }
+        } else if (hoveredFeature) {
```
ChrisLoer: Can we tweak the demo to _always_ remove the previous hover state, even if a new item is being hovered (assuming it's not the same item)? I know it's not critical for actually testing the functionality, but it's helpful for future developers if the "meaning" of the demos is easy to figure out by playing with them.

src/ui/map.js
```diff
@@ -1409,6 +1409,25 @@ class Map extends Camera {
...
+     *  required.*
+     * @param {string} key (optional) The key in the feature state to reset.
+    */
+    removeFeatureState(feature: { source: string; sourceLayer?: string; id?: string | number; }, key?: string) {
```
ChrisLoer: Can we rename the "feature" argument to something broader here (and in the matching style.js setFeatureState)? On my first reading I totally missed that you could leave out a feature ID in order to remove an entire layer.

src/source/source_state.js
```diff
@@ -16,50 +16,137 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
...
+        if (this.deletedStates[sourceLayer] === null) {
+            this.deletedStates[sourceLayer] = {};
+            for (const ft in this.state[sourceLayer]) {
+                if (ft !== feature) this.deletedStates[sourceLayer][ft] = null;
+            }
+        } else {
+            const featureDeletionQueued = this.deletedStates[sourceLayer] && this.deletedStates[sourceLayer][feature] === null;
+            if (featureDeletionQueued) {
+                this.deletedStates[sourceLayer][feature] = {};
```
ChrisLoer: IMO using `true`/`false` to indicate "deleted"/"not deleted" would be clearer than using `null`/`{}`.

src/source/source_state.js
```diff
@@ -16,50 +16,137 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
...
+        const sourceLayerDeleted = this.deletedStates[sourceLayer] === null;
+        if (sourceLayerDeleted) return;
+
+        const feature = String(featureId);
+
+        this.deletedStates[sourceLayer] = this.deletedStates[sourceLayer] || {};
+
+        if (key && featureId) {
+            if (this.deletedStates[sourceLayer][feature] !== null) {
+                this.deletedStates[sourceLayer][feature] = this.deletedStates[sourceLayer][feature] || {};
```
ChrisLoer: Translating this to `true`/`false` instead of `null`/`{}` would make this simpler to express I think, as just:

`this.deletedStates[sourceLayer][feature] = false`

src/source/source_state.js
```diff
@@ -16,50 +16,137 @@ export type LayerFeatureStates = {[layer: string]: FeatureStates};
 class SourceFeatureState {
     state: LayerFeatureStates;
     stateChanges: LayerFeatureStates;
+    deletedStates: {};
```
ChrisLoer: The goal of the merging/reconciling logic in this class isn't immediately clear, I think an introductory comment at the top could help make the rest a lot easier to read. Something along the lines of:

- we can’t just change the state directly because we want to batch changes to the tiles
- we store all updates in a single map, doing an overwrite/merge with previous updates
- we store deletes in a separate map that we use to “mask out” any changes in the update map that have since been removed
- whenever we update something, we make sure to take it out of the deleted map if it’s there.

src/ui/map.js
peterqliu: 👍gonna make it throw an error on superfluous `sourceLayer`s for geojson. preferable to hardwiring, in case user meant to use a vector source

src/source/source_state.js
peterqliu: @ChrisLoer Just realized the `false` syntax won't work at the feature level, because we need a place to record the specific key deletion, which only `{}` can do

src/ui/map.js
peterqliu: now called `target` 👍

src/source/source_state.js
peterqliu: after some thinking on this, 

1) realized that there isn't really a case for `false` here -- states not deleted simply wouldn't appear in the object
2) `true` doesn't have apparent benefits over `null`, and seems more liable to confusion with an actual state set to `true`
3) will add comment explaining the use of null, to further elucidate the usage

peterqliu:
got a bit carried away with squashing-- moving things over to https://github.com/mapbox/mapbox-gl-js/pull/7761

PR #7703, opened 2018-12-13
Use account information instead of access token for local storage 

### LINKED ISSUE #7602: turnstile event localStorage cleaning
Following up on @scottgerring's question at https://github.com/mapbox/mapbox-gl-js/pull/6980#issuecomment-436502564 ...

@asheemmamoowala if you're using temporary access tokens which are fresh for each user session, recurring users can quite easily gather quite a few of these localStorage items per day, as the days go on this will just keep growing.

Not only that but it looks like under the current implementation Mapbox will see those as new unique users when actually they are recurring users.

What if instead of using the access token as the localStorage key you only use the account username of the access token?

asheemmamoowala:
## Launch Checklist

Closes #7602 

Access tokens are not always permanent and some apps may rotate tokens quite frequently. In these cases the turnstile event is being sent more frequently than is necessary, and access token based local storage is being polluted.

This PR changes the way the turnstile event is triggered, and it's information stored, by basing it on the account owner instead of the encoded token.

cc @mick @amyjin7

### DIFF
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -5,7 +5,7 @@ import config from './config';
 import browser from './browser';
 import window from './window';
 import { version } from '../../package.json';
-import { uuid, validateUuid, storageAvailable, warnOnce, extend } from './util';
+import { uuid, validateUuid, storageAvailable, b64DecodeUnicode, b64EncodeUnicode, warnOnce, extend } from './util';
 import { postData } from './ajax';
 
 import type { RequestParameters } from './ajax';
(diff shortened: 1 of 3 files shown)


src/util/mapbox.js
```diff
@@ -267,7 +298,7 @@ export class MapLoadEvent extends TelemetryEvent {
         // mapbox tiles.
         if (config.ACCESS_TOKEN &&
             Array.isArray(tileUrls) &&
-            tileUrls.some(url => isMapboxHTTPURL(url))) {
+            tileUrls.some(url => isMapboxURL(url) || isMapboxHTTPURL(url))) {
```
ChrisLoer: Is this `isMapboxURL` bit a separate change, or am I missing how this is connected to the rest of the PR?

asheemmamoowala: It is separate I guess. https://github.com/mapbox/mapbox-gl-js/pull/7594 regressed the ability to detect mapbox tile URLs.

ChrisLoer: Ah, got it.

src/util/mapbox.js
```diff
@@ -198,12 +229,12 @@ class TelemetryEvent {
...
         const isLocalStorageAvailable = storageAvailable('localStorage');
-        const storageKey = `${telemEventKey}:${config.ACCESS_TOKEN || ''}`;
-        const uuidKey = `${telemEventKey}.uuid:${config.ACCESS_TOKEN || ''}`;
+        const storageKey =  this.getStorageKey();
+        const uuidKey = this.getStorageKey('uuid');
         if (isLocalStorageAvailable) {
             try {
                 window.localStorage.setItem(uuidKey, this.anonId);
-                if (this.eventData.lastSuccess) {
+                if (Object.keys(this.eventData).length >= 1) {
```
ChrisLoer: Why is this change here? Isn't `lastSuccess` still set in the same place?

asheemmamoowala: It is unrelated, but it ensures that state is saved even if it doesn't include `lastSuccess`

PR #7706, opened 2018-12-17
Don't use fetch API when protocol is file:

### LINKED ISSUE #7603: Use of fetch API breaks cordova apps
**mapbox-gl-js version**: 0.51.0

**browser**: Chrome 70 (webview)

#7371 introduces a bug in cordova apps loading local assets. The fetch API doesn't support the `file:` scheme whereas xhr does (https://github.com/github/fetch/pull/92#issuecomment-140665932) so when style specification or images are stored locally, they won't load after 0.51.0 (works fine up to 0.50.0).

For reference, I commented on the commit here https://github.com/mapbox/mapbox-gl-js/commit/eec8cd5e43ee861b5202a1e1eb2e7d5739e0dfef#commitcomment-31351193

gpiffault:
## Launch Checklist

Closes #7603 

This PR adds the condition that the protocol is not `file:` to use the fetch API

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -158,7 +158,12 @@ function makeXMLHttpRequest(requestParameters: RequestParameters, callback: Resp
     return { cancel: () => xhr.abort() };
 }
 
-const makeRequest = window.fetch && window.Request && window.AbortController ? makeFetchRequest : makeXMLHttpRequest;
+const makeRequest = window.fetch &&
+                    window.Request &&
+                    window.AbortController &&
+                    window.location.protocol !== 'file:' ?
+    makeFetchRequest :
+    makeXMLHttpRequest;
 
 export const getJSON = function(requestParameters: RequestParameters, callback: ResponseCallback<Object>): Cancelable {
     return makeRequest(extend(requestParameters, { type: 'json' }), callback);


src/util/ajax.js
```diff
@@ -158,7 +158,12 @@ function makeXMLHttpRequest(requestParameters: RequestParameters, callback: Resp
     return { cancel: () => xhr.abort() };
 }
 
-const makeRequest = window.fetch && window.Request && window.AbortController ? makeFetchRequest : makeXMLHttpRequest;
+const makeRequest = window.fetch &&
+                    window.Request &&
+                    window.AbortController &&
+                    window.location.protocol !== 'file:' ?
+    makeFetchRequest :
+    makeXMLHttpRequest;
```
kkaefer: This change doesn't fix the issue. It switches to XHR when the document is loaded from a file URL. However, in Safari, Chrome and Firefox, using the Fetch API to load resources works even if the document was loaded from a file URL. The way I read the original Cordova issue, the fact that the document is loaded from a file URL doesn't matter, it's only the original of the resource that is loaded _with_ `makeRequest` that matters. This means that `makeRequest` needs to be a function that discriminates between file and non-file requests and chooses XHR/Fetch (if available) at request time, and not at initialization time.

gpiffault: Well it does fix the issue in cordova (I did test it in a cordova app) but you're right that you might still be able to use the Fetch API to load https: resources (but not http:) in this context.

As I understand it, you do prefer a more specific solution at the expense of increased code complexity ?

gpiffault: I should add that indeed it doesn't fix the issue if in cordova context and the page shown in the webview is not loaded over file: (not the typical use case).
I assumed in this case you shouldn't be loading file: resources but I might miss some use cases.

mourner: @kkaefer as far as I understand, the assumption here is that it's very unlikely when setting up a Cordova app to load the main HTML file externally but resources locally, so this PR will generally fix the use case of using local resources while remaining simple and minimal. Thus it might make sense to accept it in this form.

kkaefer: That may be, but it's still incorrect, and the proposed correct solution is viable and terse as well.

kkaefer: Besides, a potential application would be to load the main UI from a website, but load e.g. fonts locally since they never change.

kkaefer:
Alternative implementation landed in https://github.com/mapbox/mapbox-gl-js/pull/7818

PR #7725, opened 2018-12-22
Add support for #map= hash prefix

JesseWeinstein:
This enables URLs used by OSM to be copied directly, without breakage.

## Launch Checklist

 - [x] document any changes to public APIs **(there are none)**
 - [ ] post benchmark scores
 - [x] tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec changes **(it doesn't)**

### DIFF
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -78,7 +78,7 @@ class Hash {
     }
 
     _onHashChange() {
-        const loc = window.location.hash.replace('#map=','').replace('#', '').split('/');
+        const loc = window.location.hash.replace('#map=', '').replace('#', '').split('/');
         if (loc.length >= 3) {
             this._map.jumpTo({
                 center: [+loc[2], +loc[1]],
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -78,7 +78,7 @@ class Hash {
     }
 
     _onHashChange() {
-        const loc = window.location.hash.replace('#', '').split('/');
+        const loc = window.location.hash.replace('#map=','').replace('#', '').split('/');
         if (loc.length >= 3) {
             this._map.jumpTo({
                 center: [+loc[2], +loc[1]],
--- a/test/unit/ui/hash.test.js
+++ b/test/unit/ui/hash.test.js
@@ -58,7 +58,7 @@ test('hash', (t) => {
         t.equal(map.getBearing(), 0);
         t.equal(map.getPitch(), 0);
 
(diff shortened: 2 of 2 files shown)


JesseWeinstein:
I know various of the Launch Checklist aren't done yet -- this seemed worth providing, at a minimum as documentation of the problem for the benefit of others who hit it.

mourner:
> This enables URLs used by OSM to be copied directly, without breakage.

Can you elaborate on this? I'm not sure I'm following why we should add special code for easier URL copying from osm.org. Can you just copy the part after `map=` for the same effect?

JesseWeinstein:
Thank you for asking!

I noticed this when using OpenInfraMap, e.g. https://openinframap.org/#8.09/34.281/-118.48 . With this (tiny) code change, I can go from an OSM link like https://www.openstreetmap.org/#map=5/45.525/-124.201 to the equivalent on OpenInfraMap (or anything else using mapbox-gl-js) by simply changing the host. Without it, I have to remember to strip out the "map=" part, or it crashes with a URL like: https://openinframap.org/#NaN/NaN/NaN , losing the details and requiring me to dig them back up from wherever I previously had them.

Ideally, everyone could settle on *ONE* standard way to represent this same info; but until we get there, it seems basic neighborliness to at least accept the format used by the main source of Mapbox's data, OSM. I plan to make a similar PR on their end to accept the format without the "map=".

I hope that clarifies things.

andrewharvey:
See also #8596 a proposal to let the developer choose any hash name.

asheemmamoowala:
Closing this in favor of https://github.com/mapbox/mapbox-gl-js/pull/8603

JesseWeinstein:
@asheemmamoowala Does #8603 actually accomplish the same purposes as this? It doesn't look like it does, from my (cursory) review of it...

andrewharvey:
> @asheemmamoowala Does #8603 actually accomplish the same purposes as this? It doesn't look like it does, from my (cursory) review of it...

No #8603 is not the same, but it does allow the developer to choose what hash prefix they want, it's un-opinionated. So if you want your maps to use the `#map=` prefix you can tell it too.

JesseWeinstein:
Right, that's very different from what this achieves, which is to enable *all* maps using this software to transparently, automatically handle the format used by OSM, the primary source for the data displayed with this software.

@asheemmamoowala -- please re-open this, as it is not handled by the other PR.

PR #7740, opened 2019-01-03
Update dotcom-page-shell

davidtheclark:
:wave: Please merge to update the page shell, cc @mapbox/gl @mapbox/docs


kkaefer:
Does this require testing from our side?

davidtheclark:
@kkaefer The same changes have already been pushed to staging, so you could double-check things on staging or locally, if you'd like. The only noticeable change should be that items in the upper-right user menu, when you're signed in, are slightly different.

katydecorah:
https://github.com/mapbox/mapbox-gl-js/pull/7742 supersedes the changes in this PR

PR #7764, opened 2019-01-11
Add options for NavigationControl button titles

rendrom:


rendrom:
I can not understand why the test-flow failed. Please tell me what I miss

ryanhamley:
What's the use case for this change @rendrom? Are you hoping to localize the button labels to the user's local language? If so, this might be better handled in a more comprehensive manner rather than piecemeal through various optional API additions.

rendrom:
@ryanhamley thanks 
yes, I tried to add options for localization because it is good for my purposes.
now for translation i'm doing [this way](https://github.com/nextgis/nextgisweb_frontend/blob/master/packages/mapboxgl-map-adapter/src/controls/ZoomControl.ts#L22)
Please tell me how to provide localization. I'll rewrite it as it should.

ryanhamley:
Localization isn't something that GL JS handles internally. However, the buttons you are attempting to change the title of are regular HTML elements so you are free to change the title attribute of them in your application like so `document.querySelector('.mapboxgl-ctrl-zoom-in').setAttribute('title', 'testing 1234');`.

For more extensive localization, you can check out [this example](https://www.mapbox.com/mapbox-gl-js/example/language-switch/) and the [Mapbox GL Language plugin](https://github.com/mapbox/mapbox-gl-language/).

I'm going to close this PR since localization isn't functionality we're going to build into the main library, but thanks for using and contributing to Mapbox!

Issue #7775, opened 2019-01-15
map.getLayer() throws TypeError on map unmount instead of returning undefined if layer does not exist


mattgif:
**mapbox-gl-js version**: 0.52.0

**browser**: Chrome

### Steps to Trigger Behavior
1. Mount Map component (React)
2. Unmount Map component

Occurring since updating to react 16.6.0 and react-redux 6.0.0.

### Expected Behavior
`this.style.getLayer(o)` returns undefined, component unmounts gracefully. 

### Actual Behavior
`this.style.getLayer(o)` throws TypeError

mourner:
GL JS doesn't have to conform to React app conventions — it's the responsibility of the developer, so if `getLayer` throws an error, you should wrap it with a try/catch. #7776 is not a suitable fix because there are many other style-related map methods that it doesn't touch, and we want to be consistent — either throw an error or fail silently in all of them.

PR #7776, opened 2019-01-15
map.getLayer() handles undefined style 7775

mattgif:
## Launch Checklist

map.getLayer throws TypeError if map style is undefined. Added check for this.style before calling this.style.getLayer.


mourner:
See https://github.com/mapbox/mapbox-gl-js/issues/7775#issuecomment-454413507

Issue #7777, opened 2019-01-15
Track popups to mouse cursor


peterqliu:
## Motivation

I've seen many use cases (in Studio's data inspector, in [my own demos](https://www.mapbox.com/bites/00273/), and [from others](https://github.com/mapbox/mapbox-gl-js/issues/7772)) where we'd want a popup "inspector" to follow the mouse, rather than any static lnglat. I've usually built these from scratch, but would be ideal to extend the current popup API for this.

## Design Alternatives

Can simulate this with the current popup by detecting mouse position on every `mousemove`, projecting that to lnglat, and `setLngLat` to the popup. On `mouseout` and `mouseenter` of the map element, the popup would hide and show itself, respectively.

But it would be nice to one-line this.

## Design/Mockup

Some API ideas:

- `popup.setLngLat({trackCursor: true})` : would expand the current method accept object literals
-  `popup.trackCursor()`: would add a new method to the Popup class, that along with `setLngLat` would mutually overwrite each other.
- Have mouse tracking as the default behavior if `setLngLat` is not called. May be the easiest to implement, but weirdest to use in practice.

ryanhamley:
I think this would be pretty similar to the draggable marker code, the meat of which is https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/marker.js#L371-L375 It's also important to account for any offset between the cursor position and the position of the element being moved around the screen which is calculated here https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/marker.js#L441

Another use case for this would be working with multiple, stacked fill extrusions and querying them to show a popup with information as you move to various "floors" of the stacked extrusions. Could be useful for things like real estate apps or terrain applications with raster-dem tiles.

peterqliu:
>I think this would be pretty similar to the draggable marker code

@ryanhamley exactly, though can we short-circuit the `mouse pixel` ➡️ `lnglat` ➡️ `popup translation pixel` dance, and just keep everything in px? Cursor offset straight to popup's CSS translate 🤔

ryanhamley:
You mean because there wouldn't be a LatLng for the popup in the case where it's following the cursor? I think that makes sense. I just wonder if leaving latLng undefined or null would cause unexpected issues elsewhere. It's definitely something you could explore. My guess is the only argument against cutting out the conversion to/from LatLng in this case is one of consistency within the API.

bravecow:
mouse cursor popup I do this way:

```javascript
[...]

PR #7786, opened 2019-01-18
Track popup to cursor [ready]

peterqliu:
ticket: https://github.com/mapbox/mapbox-gl-js/issues/7777

Adds a method `.trackPointer()` to the Popup API that repositions the popup to the mouse cursor on every mousemove event, and hides/shows the popup when cursor leaves/enters the map. Mutually overwritable with `.setLngLat`.

### Design

Opted not to extend setLngLat, as it doesn't quite fit those words. Also opted not to extend the popup `options` object, for cases when users need to change behavior after adding to map.

 ## Launch Checklist


ryanhamley:
> Touchscreen behavior: show popup only on touchmove? Not show at all? Console warning when invoking .trackCursor() on touch devices?

I'm trying to imagine a use case for this on mobile. It seems like it would be a very awkward UX as compared to simply touching on each new point to see a new popup. On the other hand, having a desktop only API method seems odd. Do we have other methods that don't work on mobile? If we do enable this on `touchmove`, maybe it would be better to call the method a more generic term like `trackPointer`.

> Close button: won't be clickable when popup shadows mouse. Should we hide? Does that affect any corner cases?

This is a fun consequence. Maybe there needs to be a separate action to close the popup in this case. Would clicking the map to toggle the popup be too weird?

peterqliu:
#### Mobile handling

In lieu of hover, I thought about adding cursor-tracked popups to a tap on the map. The interaction pattern for mobile is unfortunately quite complex: `click` is the event that typically corresponds to this, but is nondeterministic. On Webkit browsers where the user has also implemented a `hover` listener, `click` is invoked only on the *second* tap, and not a double-tap at that. Because of this unpredictability, I've removed touchscreen support entirely for the time being.

#### Other changes

- renamed `trackCursor` to `trackPointer`, to reflect the mouse-specific compatibility
- move the show/hide logic to CSS, greatly simplifying the event handler logic
- opted to keep the closeButton by default, as is the current Popup default. The change is easy enough on the user's side

peterqliu:
@KravMaguy that is correct, you will lose that functionality if you use a cursor-tracked popup on mobile. In these contexts, most apps display the information in a panel separate from the map: it's a much better practice given the limited screen space, and preserves visibility/interactivity of the map itself.

some examples ⬇️ 

<details>

img_3115
img_3117
img_3116

</details>

peterqliu:
@KravMaguy you can still `unproject`/`setLngLat` the popup on each touch event, to move the popup there

src/ui/popup.js
```diff
@@ -95,13 +96,18 @@ export default class Popup extends Evented {
...
             this._map.on('click', this._onClickClose);
         }
+
         this._map.on('remove', this.remove);
         this._update();
 
+        if (this._trackPointer) {
+            this._map.on('mousemove', (e) => { this._update(e.point); });
+            this._container.classList.add('mapboxgl-popup-track-pointer');
+        } else this._map.on('move', this._update);
```
ryanhamley: I'd wrap the else block in brackets

ryanhamley:
> Agree that this is weird, but the specific solution seems like it would vary by user/use case/preference. I'm inclined to let people modify CSS as appropriate.

👍 I do think it makes sense to change the cursor from a grabber to a pointer when in `trackPointer` mode. The idea of this feature is to point at various locations on the map and get information about them, so the pointer feels semantically valid as well as it visually covering less of the popup.

peterqliu:
- opted to hide popup when dragging the map -- tracking popup on drag made it harder to see the map when repositioning it takes priority

- cursor is now `pointer` when tracking popup

peterqliu:
>If you move the mouse quickly enough, you can end up inside the popup. I'm not entirely sure what the answer here could be, if anything.

Also removed interactivity with pointer-tracked popups, to avoid getting trapped in it 👍

Plantain:
Is this likely to make it into a release in the near future? We'd like to use this feature rather than the workarounds in #7777

peterqliu:
@Plantain this branch is ready to the extent that I can tell -- looping in some reviewers to see if we can get this across the line.

src/ui/popup.js
```diff
@@ -283,18 +322,18 @@ export default class Popup extends Evented {
...
             this._closeButton.innerHTML = '&#215;';
             this._closeButton.addEventListener('click', this._onClickClose);
         }
+
     }
 
-    _update() {
-        if (!this._map || !this._lngLat || !this._content) { return; }
+    _update(cursor: PointLike) {
+        if (!this._map || !this._lngLat && !this._trackPointer || !this._content) { return; }
```
ryanhamley: for clarity's sake, you may want to explicitly group the middle term here: `!this._map || (!this._lngLat && !this._trackPointer) || !this._content`

ryanhamley:
This is too late to be in the next release of GL JS which comes out this week, but it should be in May's release

peterqliu:
clarified a bit of the code per https://github.com/mapbox/mapbox-gl-js/pull/7786#issuecomment-479154849 and squashed as much as I can (not sure if I can squash past a previous pull from master?)

ryanhamley:
You don't have to squash ahead of time @peterqliu. Once all the checks have passed, just pressing the `Squash and merge` button will take care of everything.

PR #7812, opened 2019-01-24
Add docs for workerCount & maxParallelImageRequests

### LINKED ISSUE #7810: How to limit the number of web Worker create by mapbox-gl-js?
I run into a problem that when testing on UC browser (device: iphone6s) , the code below:
new mapboxgl.Map(opt);

It will throw an error: QuotaExceededError:Failed to construct 'Worker': Maximum worker number has been reached.
Is there any way to control or limit the number of workers for mapboxjs to solve this problem? I cant find any relative option from the doc

mourner:
Closes #7810.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -57,10 +57,11 @@ const exported = {
     set accessToken(token: string) {
         config.ACCESS_TOKEN = token;
     },
+
     /**
      * Gets and sets the map's default API URL for requesting tiles, styles, sprites, and glyphs
      *
-     * @var {string} url
+     * @var {string} baseApiUrl
      * @example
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -72,6 +73,15 @@ const exported = {
         config.API_URL = url;
     },
 
+    /**
+     * Gets and sets the number of web workers instantiated on a page with GL JS maps.
+     * By default, it equals half the number of CPU cores (but maximum 6).
```
asheemmamoowala: nit: ... By default, it is set to half the number of CPU cores, capped at 6.

PR #7817, opened 2019-01-25
Fix Safari bug which mistakenly requested WebP images

ryanhamley:
## Launch Checklist

I uncovered a bug while testing iOS Safari in which WebP images would occasionally be requested. This PR fixes that bug by waiting until the image's onload event fires to check for WebP support.

 - [ ] post benchmark scores

### DIFF
--- a/src/util/webp_supported.js
+++ b/src/util/webp_supported.js
@@ -12,12 +12,14 @@ export default exported;
 let glForTesting;
 let webpCheckComplete = false;
 let webpImgTest;
+let webpImgTestOnloadComplete = false;
 
 if (window.document) {
     webpImgTest = window.document.createElement('img');
     webpImgTest.onload = function() {
         if (glForTesting) testWebpTextureUpload(glForTesting);
         glForTesting = null;
+        webpImgTestOnloadComplete = true;
     };
     webpImgTest.onerror = function() {
         webpCheckComplete = true;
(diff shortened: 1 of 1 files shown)


src/util/webp_supported.js
```diff
@@ -34,7 +36,15 @@ function testSupport(gl: WebGLRenderingContext) {
...
-    testWebpTextureUpload(gl);
+    // HTMLImageElement.complete is set when an image is done loading it's source
+    // regardless of whether the load was successful or not.
+    // It's possible for an error to set HTMLImageElement.complete to true which would trigger
+    // testWebpTextureUpload and mistakenly set exported.supported to true in browsers which don't support webp
+    // To avoid this, we set a flag in the image's onload handler and only call testWebpTextureUpload
+    // after a successful image load event
+    if (webpImgTestOnloadComplete) {
+        testWebpTextureUpload(gl);
+    }
```
ansis: I think there is still a possible race condition here. `testSupport` could run when the image has successfully loaded (`webpImgTest.complete === true`) but `onload` has not run yet. In this case `testWebpTextureUpload` would never run and it would never start loading webps. That isn't as bad as loading then when we shouldn't but it might be good to fix.

I think this could be fixed by adding 
```
} else {
    glForTesting = gl;
}
```
here and removing the `if (!webpImgTest.complete) {` section completely.

ryanhamley: I think I pushed the changes you're asking for

Issue #3636, opened 2016-11-16
Relative URLs in sources don't work anymore (0.27.0)


nicooprat:
With version 0.26.0 I had something like this:

```js
"sprite": "/mapbox-gl-styles/sprites/bright-v9",
"glyphs": "/mapbox-gl-styles/glyphs/{fontstack}/{range}.pbf",
``` 

in my style definition, because I wanted it to work in all environments. Now with 0.27.0, the function `parseUrl` throws an error because the URL doesn't match the expected syntax (protocol missing):

https://github.com/mapbox/mapbox-gl-js/blob/44727a70955a77ef5fbf3c46aa52160af0b2e086/js/util/mapbox.js#L93

I guess the protocol could be retrieved, if missing, with `location.protocol` ; but I'm not good enough with regexs to fix it myself... :) Here, if the regex doesn't match, it returns `null`.

Thanks.

jfirebaugh:
It wasn't intended for `sprite` and `glyphs` properties to support relative URLs; if it worked in past versions, it was an accident.

> because I wanted it to work in all environments

Can you say what specifically you mean by this?

nicooprat:
I mean that each environment retrieves its own glyphs and fonts: local from local, staging from staging, prod from prod... So I didn't need to hard code each URLs for each environment.

lucaswoj:
It is a reasonable expectation that relative URLs would work.

jfirebaugh:
I don't think it's something we should support.

* There's ambiguity about what the URL should be relative to. Should it be the style URL or the page URL?
* If it's the style URL, then it needs explicit code and acts differently than relative URLs elsewhere in Web APIs.
* If it's the page URL, then it's a style spec feature that can never be replicated in gl-native.

lucaswoj:
Those are good points. 👍 Let's close this ticket and open another to consider adding this rule to the validator.

brianssheldon:
Since my production environment is not connected to the internet, this was keeping me from upgrading mapbox-gl.js...

My fix for this turned out to be the same one I used for the tiles. 

from the json file in my project- 
"sprite": "https://replacemewiththehostnameplease:4567/myfolder/style/bright/sprite",

then in the js code:
jQuery.get('/myfolder/tweakedstyle.json', function(theTweakedJson) {
        var theStringifiedTweakedJson = JSON.stringify(theTweakedJson);
[...]

PR #7818, opened 2019-01-25
Perform XHR in the main context 

kkaefer:
...rather than the worker's to preserve referrer information.

IE11 doesn't send the correct referrer for XHRs in a web worker. Instead, we're now doing the request from the main context and transfering the result back to the worker. Most modern browsers will continue to use the Fetch API introduced in https://github.com/mapbox/mapbox-gl-js/pull/7371 and will not be affected by this change.

## Launch Checklist

 - [ ] post benchmark scores

I looked at the impact of transferring the ArrayBuffer/Image object in IE and it's very small, typically below 1ms. In any case, this code path will only be triggered in IE11 (and other older browsers that don't support the Fetch API).

I added a `Cancelable` mode to the Actor as well so that we can cancel requests across workers.

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -71,14 +71,17 @@ class AJAXError extends Error {
     }
 }
 
+function isWorker() {
+    return typeof WorkerGlobalScope !== 'undefined' && typeof self !== 'undefined' &&
+           self instanceof WorkerGlobalScope;
+}
+
 // Ensure that we're sending the correct referrer from blob URL worker bundles.
 // For files loaded from the local file system, `location.origin` will be set
 // to the string(!) "null" (Firefox), or "file://" (Chrome, Safari, Edge, IE),
 // and we will set an empty referrer. Otherwise, we're using the document's URL.
 /* global self, WorkerGlobalScope */
-export const getReferrer = typeof WorkerGlobalScope !== 'undefined' &&
-                           typeof self !== 'undefined' &&
-                           self instanceof WorkerGlobalScope ?
+export const getReferrer = isWorker() ?
     () => self.worker && self.worker.referrer :
     () => {
         const origin = window.location.origin;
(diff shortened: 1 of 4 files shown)


src/util/ajax.js
```diff
@@ -158,7 +161,15 @@ function makeXMLHttpRequest(requestParameters: RequestParameters, callback: Resp
...
 }
 
-const makeRequest = window.fetch && window.Request && window.AbortController ? makeFetchRequest : makeXMLHttpRequest;
+export const makeRequest = function(requestParameters: RequestParameters, callback: ResponseCallback<any>): Cancelable {
+    if (window.fetch && window.Request && window.AbortController && !/^file:/.test(requestParameters.url)) {
+        return makeFetchRequest(requestParameters, callback);
+    } else if (isWorker() && self.worker.actor) {
+        return self.worker.actor.send('getResource', requestParameters, callback);
+    } else {
+        return makeXMLHttpRequest(requestParameters, callback);
```
mourner: Is it intended that the worker-main hack will be used not only in IE11 context but also for file protocol requests in modern browsers, or am I misreading the conditions?

kkaefer:
Also fixes https://github.com/mapbox/mapbox-gl-js/issues/7603

src/util/ajax.js
```diff
@@ -158,7 +161,22 @@ function makeXMLHttpRequest(requestParameters: RequestParameters, callback: Resp
...
 }
 
-const makeRequest = window.fetch && window.Request && window.AbortController ? makeFetchRequest : makeXMLHttpRequest;
+export const makeRequest = function(requestParameters: RequestParameters, callback: ResponseCallback<any>): Cancelable {
+    // We're trying to use the Fetch API if possible. However, in some situations we can't use it:
+    // - IE11 doesn't support it at all. In this case, we dispatch the request to the main thread so
+    //   that we can get an accruate referrer header.
+    // - Requests for resources with the file:// URI scheme don't work with the Fetch API either. In
+    //   this case we unconditionally use XHR on the current thread since referrers don't matter.
+    if (!/^file:/.test(requestParameters.url)) {
```
gpiffault: Doesn't work with relative URLs

CookieCookson: I can confirm this, you have to specify the absolute path with the file protocol before passing it in, else it thinks it is an http request and throws an error about the file protocol like before.

kkaefer: We don't support relative URLs intentionally — see the discussion at #3636 and linked tickets. That said,  https://github.com/mapbox/mapbox-gl-js/pull/8612 should now correctly classify these URLs as file URLs.

kkaefer: We don't support relative URLs intentionally — see the discussion at #3636 and linked tickets. That said, #8612 should now correctly classify these URLs as file URLs.

PR #7840, opened 2019-01-29
Update docs-related deps

mourner:
Updates docs-related dependencies to their latest versions, including `documentation` v9 and newer `remark` deps. Note that I removed `object-entries` because we now target Node v8+.

Started the docs server successfully and there seem to be no apparent issues, but worth double-checking just in case.


docs/components/sdk_support_table.js
```diff
@@ -24,7 +23,7 @@ export default class SDKSupportTable extends React.Component {
                         </tr>
                     </thead>
                     <tbody>
-                        {entries(this.props).map(([key, entry], i) => <tr key={i}>
+                        {Object.entries(this.props).map(([key, entry], i) => <tr key={i}>
```
ryanhamley: `Object.entries` is not supported in IE. Do we have a polyfill somewhere for it?

mourner: Oh, I didn't realize this is executing at runtime rather than compiling.

ryanhamley: Ah I could be wrong. I just know that we had an issue in the current release where docs weren't working in IE11 because of an Object.assign call so I assumed this could be an issue as well.

mourner:
Hmm, now before we merge, I'm really curious about why it works in IE11! It doesn't support `Object.entries` according to online docs, and the JS calls are present in the generated code, so I'm at loss for why it would work...

PR #7843, opened 2019-01-29
update style benchmark location descriptions and zoom levels

nickidlugash:
This PR updates some of the style benchmark location descriptions, so that they more accurately reflect what aspect of the style is being reviewed:

- Better description of what map features are prominent in the location
- Explicitly calls out use of overzoomed tiles in location description 

This also changes the zoom level for one location ("High zoom (pedestrian polygon fills, roads, paths, landuse, labels) – Paris, z16.25") from z17 to z16. The original intention for this location was to be at a non-integer zoom level, but since it turned out that that isn't possible, z16 would be preferable to z17.

## Launch Checklist

@ryanhamley is there anything else I should do for this PR?

 - [ ] post benchmark scores

/cc @dasulit @tristen

### DIFF
--- a/bench/lib/style_locations.js
+++ b/bench/lib/style_locations.js
@@ -2,13 +2,13 @@ import { OverscaledTileID } from '../../src/source/tile_id';
 
 export default [
     {
-        "description": "Road labels – Houston, z12",
+        "description": "Roads – Houston, z12",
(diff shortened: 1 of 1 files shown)


nickidlugash:
@ryanhamley Thanks for the review!

PR #7859, opened 2019-01-31
rename example for accepting coordinates in the geocoder

andrewharvey:
## Launch Checklist

There is already a geocoder example which does a good job of showing how to include search results from your own custom data in the geocoder -> https://docs.mapbox.com/mapbox-gl-js/example/forward-geocode-custom-data/

This PR, renames the original demo for that function to better reflect exactly what the example does, ie. accepting coordinates as input to the geocoder and including these as part of the results.

This change will break links in the wild, but I think that's okay.


andrewharvey:
I'll wait for feedback on if this change is acceptable or not, before I address the conflict introduced by #7924. So let's not let the conflict hold up a review.

PR #7861, opened 2019-01-31
Much nicer integration tests pages

mourner:
Our current render/query/expression tests result pages were becoming hard to use because of the table layout and some of the tests being super-wide, so I took a stab at fixing this and fully redesigned the pages in the process. Check it out:

- [ Render tests before](https://37329-8629417-gh.circle-artifacts.com/0/root/mapbox-gl-js/test/integration/render-tests/index.html)
- [Render tests after](https://37348-8629417-gh.circle-artifacts.com/0/root/mapbox-gl-js/test/integration/render-tests/index.html)
- [Query tests before](https://37328-8629417-gh.circle-artifacts.com/0/root/mapbox-gl-js/test/integration/query-tests/index.html)
- [Query tests after](https://37353-8629417-gh.circle-artifacts.com/0/root/mapbox-gl-js/test/integration/query-tests/index.html)

Notes:

- Made the layout list-like rather than columned.
- Overall cleaner look that's much easier to skim.
- Removed a bunch of unnecessary context such as width/height/center/pitch/bearing — you can look it up (along with other style-defining things) in corresponding `style.json` if needed.
- Added stats to see how many tests are passed/failed/ignored-passed/ignored-failed.
- Put shuffle seed and the list of failed tests to the "test sequence" block where it belongs (it shows when you click "Toggle test sequence").
- Fixed diff rendering in query tests.

### DIFF
--- a/test/integration/results.html.tmpl
+++ b/test/integration/results.html.tmpl
@@ -1,76 +1,73 @@
 <style>
-    body { font-family: Helvetica; }
-    div { color: black; }
-    h2 { text-align: center; margin-bottom: 10px; padding: 2px 5px; font-weight: normal; }
-    h2 a { color: white; text-decoration: none; }
-    table td { vertical-align: top; }
-    ul { margin-top: 0; }
-    .hide { display: none; }
-    del { background-color: #ffeef0; }
-    ins { background-color: #e6ffed; }
-    del,ins { text-decoration: none; }
+body { font: 18px/1.2 -apple-system, BlinkMacSystemFont, "Helvetica Neue", Helvetica, Arial, sans-serif; padding: 10px; }
+h1 { font-size: 32px; margin-bottom: 0; }
+button { vertical-align: middle; }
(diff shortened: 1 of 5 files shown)


test/integration/results.html.tmpl
```diff
@@ -1,76 +1,73 @@
...
+        el.dataset.altSrc = tmp;
+    }
+}
 
-const sequenceButton = document.getElementById('toggle-sequence');
-sequenceButton.addEventListener('click', function (e) {
-    const sequence = document.getElementById('test-sequence');
-    sequence.style.display = sequence.style.display == "none" ? "block" : "none";
+document.getElementById('toggle-passed').addEventListener('click', function (e) {
+    for (const row of document.querySelectorAll('.test.passed')) {
```
asheemmamoowala: Should this also hide `ignored passed` tests?

mourner: @asheemmamoowala these test will have two classes — `ignored` and `passed`, so the selector above will include them.

Issue #7572, opened 2018-11-08
Mousewheel zoom very slow - Chrome/Linux


pathmapper:
**mapbox-gl-js version**: v0.51.0

**browser**: Chrome 70.0.3538.77 (64-Bit) on Linux

### Steps to Trigger Behavior

 Use the mousewheel for zooming in and out.

### Link to Demonstration

 https://www.mapbox.com/mapbox-gl-js/example/simple-map/ with `hash: true`

### Expected Behavior

Zooming in and out happens faster, at least 0.25 zoom level per mouswheel "click".

### Actual Behavior

Only 0.08 - 0.09 zoom level per mouswheel "click".

---

Tried it on 2 different machines.

For comparision:

|   |  zoom level / mouswheel "click" |
| ------------- | ------------- |
| Google Maps | 0.25  |
| www.openstreetmap.org  | 1  |

andrewharvey:
I notice this too.

Using my laptop trackpad two finger scroll I can go from z0 to z24 in about full 4 scroll motions, compared to on the mousewheel it takes 24 motions.

It's so painful with the mouse scroll, by habit I either use the trackpad or use the shift zoom box feature, which takes 3 motions to get from z0 to z24.

asheemmamoowala:
@andrewharvey What OS are you using? 
@pathmapper & @andrewharvey What type of mouse devices are you using and do you have any additional device driver software installed to configure scroll wheels?

Our implementation uses these magic numbers to implement wheel zoom:

https://github.com/mapbox/mapbox-gl-js/blob/c708474eb65d9c6a117fe232b677db19525f70b4/src/ui/handler/scroll_zoom.js#L17-L22

This [archived link](https://archive.fo/ZV8gz) seems to show that the `wheelZoomDelta` should be OS specific, but I'm not sure if that is true any longer.

andrewharvey:
Debian, just a very stock standard mouse.

I never reported the issue since it's very depended on OS and the mouse hardware, if you change the values to make it perfect for me, it will be way to fast for the next person. Plus at the moment you can make small adjustments in zoom if you make it faster you'll loose that precision.

It partly might just be down to my OS settings, I could probably make the scroll faster on my end already.

pathmapper:
Logitech MX ERGO and Logitech G402, Debian, no additional driver and both with the same behavior.

> wheelZoomDelta should be OS specific

It is also dependent on the browser, here the values for Firefox 60.3.0esr:

[...]

Issue #4723, opened 2017-05-19
Allow configuring zoom speed/step


juho-hanhimaki:
## Motivation
Zooming the map via scroll or double tap feels too slow most of the time. There is too much scrolling or double tapping to do. This is especially painful on devices with only touchpad.

Our use scenario has large fleet of assets that are situated around the globe. Users will often start with wide view of globe and then go into deep zoom levels to see more of the local context around specific asset. This process involves a lot of zooming in (and sometimes out too). It is common complaint for us that there is too much zooming to be done to get to desired level of zoom.

## Design
Make zoom step/speed was configurable via map options or in the zoom handlers. At the moment it seems the scroll_zoom.js and dblclick_zoom.js use hard coded values instead of configurable values.

andrewharvey:
I've found this varies by the pointer device/os a fair bit. For me it's the opposite the scroll wheel on the mouse is slow but the touchpad scroll is nice and fast.

Perhaps in your UI you could hint your users to use Shift+drag a box to speed up zooming in? as a workaround for it being configurable.

stevage:
Also running into this. My experience is like Andrew's - with the touchpad it's easy to do a big swipe that zooms forever, whereas the wheel is much more work.

It seems to be related to how much data is being displayed. An empty map scrolls in and out quickly, while a very heavy one feels like it's skipping scroll events.

Caerbannog:
The title of this issue is about adding a configuration option.
I suspect that fixing the related issue #7572 (inconsistent scroll speed across browsers) might be a better solution for some of the people affected here.

wisyr:
Current scrolling speed is way too low. Looking forward for this feature.

mourner:
Here are the constants that have to be made configurable: 

https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/handler/scroll_zoom.js#L19-L22

Adding a "good first issue" label — a PR would be welcome! You can see how handler options are implemented in other handlers like drag pan or drag rotate.

ryanhamley:
Closed by https://github.com/mapbox/mapbox-gl-js/pull/7863 🎉

PR #7863, opened 2019-01-31
ScrollZoomHandler: allow to set zoomRate

sf31:
Hello there, first PR here, I hope I'm doing it right.

I noticed that scroll zoom on desktop browsers is really slow when using mouse since you need to 'wheel' so many times. I didn't find a way to set 'scroll/zoom sensibility' so I've edited ScrollZoomHandler class and added setZoomRate(number) and setWheelZoomRate(number) methods.
These basically allow to set/override default values of 'defaultZoomRate' and 'wheelZoomRate'.

It seems to work pretty well, I'm sending this PR to let you know about this possiblity in case you want to include that in the library.

Thanks for your work!

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/handler/scroll_zoom.js
+++ b/src/ui/handler/scroll_zoom.js
@@ -87,7 +87,7 @@ class ScrollZoomHandler {
 
     /**
      * Set the zoom rate of a mouse wheel
-     * @param {number} [zoomRate = 1/450]
+     * @param {number} [wheelZoomRate = 1/450]
      */
     setWheelZoomRate(wheelZoomRate: number) {
         this._wheelZoomRate = wheelZoomRate;
--- a/src/ui/handler/scroll_zoom.js
+++ b/src/ui/handler/scroll_zoom.js
@@ -16,6 +16,7 @@ import type {TaskID} from '../../util/task_queue';
 
 // deltaY value for mouse scroll wheel identification
 const wheelZoomDelta = 4.000244140625;
+
 // These magic numbers control the rate of zoom. Trackpad events fire at a greater
 // frequency than mouse scroll wheel, so reduce the zoom rate per wheel tick
 const defaultZoomRate = 1 / 100;
@@ -65,8 +66,6 @@ class ScrollZoomHandler {
 
         this._delta = 0;
 
-        // These magic numbers control the rate of zoom. Trackpad events fire at a greater
-        // frequency than mouse scroll wheel, so reduce the zoom rate per wheel tick
         this._defaultZoomRate = defaultZoomRate;
         this._wheelZoomRate = wheelZoomRate;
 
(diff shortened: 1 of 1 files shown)


andrewharvey:
:+1: here's the tickets this relates to

- likely closes #4723: Allow configuring zoom speed/step
- related to #7572: Mousewheel zoom very slow - Chrome/Linux, I think to close this ticket it would need to change the defaults to be faster, not just make it configurable.

sf31:
> I think to close this ticket it would need to change the defaults to be faster, not just make it configurable

In this case it could be enough change those constants to a higher value, but then you are just moving the problem to those people who want an high precision/small steps in scroll zoom. Why do you think it should not be configurable?

pathmapper:
#7572 is about making the mousewheel zoom behavior consistent across different OS/browser/mouse combinations.

ryanhamley:
Yeah it looks like to close #7572, we'd need to add something like https://github.com/facebookarchive/fixed-data-table/blob/master/src/vendor_upstream/dom/normalizeWheel.js but I think that's not necessarily in scope of this PR. If the author wants to tackle that, great, but this PR doesn't need to include scroll normalization to be considered "done".

andrewharvey:
> Why do you think it should not be configurable?

Oh, I think it should be configurable :100:%

> Yeah it looks like to close #7572, we'd need to add something like https://github.com/facebookarchive/fixed-data-table/blob/master/src/vendor_upstream/dom/normalizeWheel.js but I think that's not necessarily in scope of this PR. If the author wants to tackle that, great, but this PR doesn't need to include scroll normalization to be considered "done".

Absolutely, sorry I didn't mean to derail this into #7572.

src/ui/handler/scroll_zoom.js
```diff
@@ -70,6 +78,22 @@ class ScrollZoomHandler {
         ], this);
     }
 
+    /**
+     * Override default zoomRate value
```
ryanhamley: How about something like `Set the zoom rate of a trackpad`?

We should also add the default here so that people reading the docs can know what sort of number to enter. You can include a default like this `@param {number} [zoomRate=1/100]`

src/ui/handler/scroll_zoom.js
```diff
@@ -70,6 +78,22 @@ class ScrollZoomHandler {
...
+     * @param {number} zoomRate
+     */
+    setZoomRate(zoomRate: number) {
+        this._defaultZoomRate = zoomRate;
+    }
+
+    /**
+     * Override default wheelZoomRate value
```
ryanhamley: `Set the zoom rate of a mouse wheel` and add the default here as well.

sf31:
Hi, I've added/updated missing commets, let me know if more is needed!

ryanhamley:
Thanks!

PR #7868, opened 2019-02-01
Add text-variable-anchor-animation property

ChrisLoer:
This PR is split out of #7596 because of concerns that adding this style spec property commits us to ongoing support, and if we decide that cross-fade is the more important/useful animation, having to support this animation as well would get in the way. We think cross-fading is much better suited to handling many location changes on the map at the same time without overwhelming the user visually.

Nonetheless, I think this translation-based animation actually works pretty well in the case you're displaying a few high-priority labels, and just because I'm fond of it, I'm splitting it into its own PR. Once variable-label-placement goes into master, it should be easy to rebase this on top and merge -- the actual code change involved is quite small.

cc @ansis @nickidlugash @chloekraw @asheemmamoowala


ChrisLoer:
This is easy to do because the current approach to variable anchor placement is to prepare one set of glyphs (or up to three depending on line justification), and then move the glyphs around in the CPU at run-time, similar to the way we handle line labels. Since we're moving them on each frame already, it's easy to add the interpolation logic.

If we decide to implement cross-fading for variable anchor placement, we can't use that approach any more because we need to show two (partially faded) copies of the glyphs at the same time. So we could switch to a different approach, where we generate/upload glyphs for each possible position, and just control opacity for them along the lines of our current collision fading logic. In that approach, we'd have more overhead in glyph upload, but we'd avoid the cost of dynamic layout at render time... except that continuing to support the translation-style animation in this PR would require us to hold onto all the render-time dynamic layout logic.

mourner:
There doesn't seem to be a consensus on whether we should support this, so let's close the PR for now and reopen when we reconsider.

PR #7871, opened 2019-02-03
Improve wording and syntax definition of "case" and "match".

### LINKED ISSUE #7779: Spell out that `match` performs the role of `in`
I was trying to work out how to implement an old-style `"in"` filter using an expression:

`["in", "myfield", "val1", "val2", "val3"]`

After looking through the expression doc, it seemed there wasn't a way. Only by using Studio did I work out that this is the new way:

`["match", ["get", "myfield"], ["val1", "val2", "val3"], true, false]`

So I'd suggest improving this paragraph, which doesn't actually specify the behaviour when an array of values is provided:

>Selects the output whose label value matches the input value, or the fallback value if no match is found. The input can be any expression (e.g. ["get", "building_type"]). Each label must either be a single literal value or an array of literal values (e.g. "a" or ["c", "b"]), and those values must be all strings or all numbers. (The values "1" and 1 cannot both be labels in the same match expression.) Each label must be unique. If the input type does not match the type of the labels, the result will be the fallback value.

stevage:
Fixes #7779.
Using 'fallback' rather than 'default' in code to avoid syntax highlighting
issue where 'default' is detected as a reserved word.

Also I tweaked `case` a bit to make the two as consistent as possible. Using `...` rather than the `repeat: []` construct seems to give better output, with better controlled line breaks.

### DIFF
--- a/docs/components/expression-metadata.js
+++ b/docs/components/expression-metadata.js
@@ -88,7 +88,7 @@ const types = {
     case: [{
         type: 'OutputType',
         parameters: [
-            'condition: boolean, output: OutputType', 
+            'condition: boolean, output: OutputType',
             'condition: boolean, output: OutputType',
             '...',
             'fallback: OutputType']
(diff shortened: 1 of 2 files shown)


stevage:
Fixed.

PR #7875, opened 2019-02-04
[docs] remove duplicated sources-geojson section

### LINKED ISSUE #7873: Wrong StyleSpec Docs for Geojson Source
Not sure if this is the right repo to report bug about the mapbox gl js doc, but doing anyway.
(StyleSpec Docs for Geojson Source)[https://docs.mapbox.com/mapbox-gl-js/style-spec/#sources-geojson]

There are two sections of geojson source in the doc. The first geojson source doc list properties like 
`tiles`/`scheme`/`url` (properties from TileJSON). And I guess the first one is wrong.

**mapbox-gl-js version**:

**browser**:

### Steps to Trigger Behavior

 1.
 2.
 3.

### Link to Demonstration

https://jsbin.com/

katydecorah:
The sources-geojson found itself duplicated and nested inside the vector-sources section. This PR removes the duplicated section.

Fixes #7873

### DIFF
--- a/docs/pages/style-spec.js
+++ b/docs/pages/style-spec.js
@@ -368,64 +368,6 @@ export default class extends React.Component {
                                             "url": "mapbox://mapbox.mapbox-streets-v6"
                                         }`)}
                                 </div>
-                                <SectionH3 id='sources-geojson' title='geojson'>
-                                    <p>
-                                        A <a href="http://geojson.org/">GeoJSON</a> source. Data must be provided via a <code>"data"</code>
-                                        property, whose value can be a URL or inline GeoJSON.
-                                    </p>
-                                    <div className='mb6'>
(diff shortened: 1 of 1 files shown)


docs/pages/style-spec.js
```diff
@@ -539,11 +481,16 @@ export default class extends React.Component {
                                         android: '4.2.0',
                                         ios: '3.4.0',
                                         macos: '0.3.0'
+                                    },
+                                    'line distance metrics': {
```
colleenmcginnis: Where is this coming from @katydecorah? Was this addition intentional?

katydecorah: Yes, I pulled it from the duplicated content and checked the git history to make sure `line distance metrics` should be included here.

PR #7880, opened 2019-02-06
[docs] update dr-ui 0.5.0

katydecorah:
* [x] update dr-ui to 0.5.0
* [x] replace prism.css with `require.resolve('@mapbox/dr-ui/css/prism.css')`
* [x] move h1 into MarkdownPageShell
* [x] replace `themeIcon` on `Icon`
* [x] remove styled scrollbar css


ryanhamley:
@katydecorah You may need to update your Yarn installation. `yarn.lock` shouldn't have so many changes to the `integrity` field. For example, updating the `dr-ui` version in `package.json` probably shouldn't change the `integrity` field of various Babel plugins.

katydecorah:
Thanks for pointing that out @ryanhamley, I did as you suggested and the diff for the yarn.lock looks a lot less noisy now

docs/pages/examples.js
```diff
@@ -53,7 +53,7 @@ class ExamplesLandingPage extends React.PureComponent {
         const gettingStartedSection = (
             <div className="">
                 <div className='prose'>
-                    <h1 className='mt24 mt0-mm' id='getting-started'>Getting started</h1>
+                    <h1 className='mt24 mt0-mm txt-fancy' id='getting-started'>Getting started</h1>
```
colleenmcginnis: I'm not sure about using `txt-fancy` here.

katydecorah: To keep the headings consistent, what do you think of making "Examples" the h1 on this page and dropping "Getting started" to h2:

image

This would follow the pattern of the Plugins page:

image

colleenmcginnis: Yes. I think that makes more sense -- I guess I think of `h1` as describing the whole page and `Examples` is more accurate than `Getting started` if we follow that guideline. 🙂

katydecorah:
@colleenmcginnis  I found a few other layout things that I think we should make consistent with other sites and I just pushed those commits. Here's what the changes look like:

### Examples

Before | After
-------|------
image | image


### Plugins

Before | After
-------|------
image | image

colleenmcginnis:
Yesss. Looks great to me @katydecorah 🚀

PR #7891, opened 2019-02-08
[docs] style fix for examples code blocks

katydecorah:
In #7880, we adjusted the use of `prose` which turned off the class on example pages, as a result the examples pages lost some styling on the code block. This PR fixes the cascade.

image

@colleenmcginnis should we also debug the warning note flash that we're seeing on https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-directions/ in this PR or separate one? I'm not seeing it locally, but we could test on staging.

### DIFF
--- a/docs/components/example.js
+++ b/docs/components/example.js
@@ -68,7 +68,7 @@ ${html}
             const {frontMatter} = this.props;
             return (
                 <PageShell meta={frontMatter}>
-                    <div className='relative'>
+                    <div className='relative prose'>
                         <div className='round bg-white'>
                             <div className='prose'>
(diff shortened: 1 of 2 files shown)


colleenmcginnis:
>@colleenmcginnis should we also debug the warning note flash that we're seeing on https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-directions/ in this PR or separate one?

@katydecorah let's address this in a follow up PR. I can open a ticket this afternoon so we don't forget next week! @ryanhamley feel free to merge here 🙂

Issue #7894, opened 2019-02-12
examples filter doesn't seem to work while the page is still loading


andrewharvey:
I noticed on the examples page https://docs.mapbox.com/mapbox-gl-js/examples/ if you start typing in the filter list input box while the page is still loading it won't filter anything even after the page has loaded. You need to then force it by changing the input.

/cc @katydecorah

colleenmcginnis:
I'm transferring this issue to the new home for the Mapbox GL JS documentation ([/mapbox-gl-js-docs](https://github.com/mapbox/mapbox-gl-js-docs)). See https://github.com/mapbox/mapbox-gl-js/pull/8582 for details on this move.

PR #7900, opened 2019-02-14
[docs] use appropriate-images for examples to increase page speed

katydecorah:
ref https://github.com/mapbox/mapbox-gl-js/issues/7894

The new example images were adding significant weight to the [examples page](https://docs.mapbox.com/mapbox-gl-js/api/). As a result, on slower connections it could make the user experience delayed if you only want to search to find an example. 

This PR use the appropriate-images module to compress and generate different image sizes/formats for each thumbnail. As a result, this cut the page weight by more than half (originally 12.2 MB transferred, now 5.4 MB). I also tested by throttling my page speed and found I was able to search the examples better than before.

I've updated documentation in the docs/README.md and tests to assert that when a new example is added, that the unit test will fail if the contributor hasn't added an image to the correct directory and run the appropriate-images script to generate the image sizes.

@andrewharvey could you also test this PR (`yarn install` then `npm run start-docs` and then finally navigating to /examples to try the search box)? I'd love to double check that the searching experience has improved for you!

## Launch Checklist

 - [x] document any changes


andrewharvey:
This is great @katydecorah, I didn't realise such large images were being requested only to be scaled down for the thumbnail.

Unless I misunderstand things, this just mitigates the likely hood of #7894 rather than completely preventing it? It seems the underlying issue is you can start typing in the filter/search input, but it won't do anything unless all the thumbnails have finished loading, even though the actual list of examples has already loaded. It would be nice if this underlying issue could be fixed.

I did try to test locally but `yarn run start-docs` gets to Webpack compiled, but when requesting the site it eventually crashes with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`.

ryanhamley:
@andrewharvey Are you doing anything in particular to trigger this out of memory warning? I've had the docs running for maybe 20 minutes and haven't seen it yet. Also, what version of Webpack are you using? `yarn install` should install Webpack 3.x.x but it looks like there's a [known memory leak in Webpack 4](https://github.com/webpack/webpack/issues/6929)

andrewharvey:
> @andrewharvey Are you doing anything in particular to trigger this out of memory warning? I've had the docs running for maybe 20 minutes and haven't seen it yet. Also, what version of Webpack are you using? yarn install should install Webpack 3.x.x but it looks like there's a known memory leak in Webpack 4

It happens straight away while trying to load the first examples page.

./node_modules/.bin/webpack --version
3.11.0

^ installed in mapbox-gl-js by yarn.

colleenmcginnis:
@katydecorah this looks good re: the implementation of appropriate-images. Two notes: 

1. I had a hard time installing the dependencies of appropriate-images. I was able to track it down to attempting to use node v9.10.0, which was one of several versions suggested when I ran `yarn install` (`The engine "node" is incompatible with this module. Expected version "^6.14.0 || ^8.10.0 || >=9.10.0"`). It looks like `sharp` won't work with any node v9.x.x. (It works with 6, 7, 8, 10, and 11). I'm wondering if we should add something to the README about this restriction?
2. In all other uses of appropriate-images we use a naming pattern `likeThis` instead of `like-this` in the image config. I'm not sure there's a reason to prefer one over the other besides consistency.

ryanhamley:
@andrewharvey I've been unable to reproduce this error. Maybe it's environment-specific. I think you typically use Linux, right?

@colleenmcginnis Are you 100% sure you were on Node 9.10? I can't reproduce the error you saw during `yarn install` on 9.10 but I can on 9.0

colleenmcginnis:
@ryanhamley sorry I wasn't clear! Here's what I'm seeing: 

- I was on node `v9.4.0` and ran `yarn install`.
- I got that error: `1. The engine "node" is incompatible with this module. Expected version "^6.14.0 || ^8.10.0 || >=9.10.0"`.
- I used nvm to switch to `v9.10.0`.
- I ran `yarn install` again. It starts to run, but it fails to install `sharp` (a dependency of `appropriate-images`).
- Without `sharp`, I can't successfully run `node docs/bin/appropriate-images.js --all`. 
- I bump up to node `v10+`. 
- Everything installs correctly and the appropriate-images script works. 🎉 

My conclusion: You need to use node v10+ in order to install all dependencies and run the appropriate-images script.

ryanhamley:
Weird. I just followed the steps you outlined @colleenmcginnis and it worked fine on 9.10 for me. Are you doing `rm -r node_modules` before running `yarn install`? I wonder if it's possible that the Node gyp bindings are incorrect if you're not starting fresh on the Node modules when you switch to a new version.

screen shot 2019-02-15 at 4 44 51 pm

colleenmcginnis:
Very weird. Yes I removed all `node_modules` before running `yarn install`. @katydecorah didn't have problems either, though we didn't talk about what node version she was using.

andrewharvey:
> @andrewharvey I've been unable to reproduce this error. Maybe it's environment-specific. I think you typically use Linux, right?

Correct. Though don't let me hold up the :ship:

katydecorah:
> Unless I misunderstand things, this just mitigates the likely hood of #7894 rather than completely preventing it? It seems the underlying issue is you can start typing in the filter/search input, but it won't do anything unless all the thumbnails have finished loading, even though the actual list of examples has already loaded. It would be nice if this underlying issue could be fixed.

@andrewharvey right, this PR mitigates the likelihood but doesn't prevent it. I've updated the PR so that it references the original issue instead of fixes it and we'll circle back after this PR is complete.

@colleenmcginnis per chat, I made @mapbox/appropriate-images an optionalDependencies as that's the pattern we were using in other sites. Hopefully this cuts down on errors. A couple follow-ups:


>  It looks like `sharp` won't work with any node v9.x.x. (It works with 6, 7, 8, 10, and 11). I'm wondering if we should add something to the README about this restriction?

I've been in the habit of running `nvm use 8` before loading batfish sites. That may be a good thing for us to add to the README or CONTRIBUTING

> In all other uses of appropriate-images we use a naming pattern `likeThis` instead of `like-this` in the image config. I'm not sure there's a reason to prefer one over the other besides consistency.

I think that as long as it's consistent in the repo then it's ok. I went with kebab case so we didn't have to add an image id to each markdown file to map the image with the file, instead we use the file name as the image id.

asheemmamoowala:
@katydecorah Is there a reason that the contents of the `img/dist` folder need to be checked in to the repo? Couldn't this just be run as needed to build the docs? It would slow down the first run on a clean repo, but after that would only need to update images that are added/modified.

katydecorah:
> Couldn't this just be run as needed to build the docs

@asheemmamoowala thanks for pointing that out. I've reworked how we're building the images to remove the dist folder from the repo. I tested this branch on `publisher-staging` and the images are generated as expected.

Does anyone on the gl team have time to review? @asheemmamoowala @ryanhamley

docs/README.md
```diff
@@ -33,10 +33,18 @@ In the `.html` file, write the HTML and JavaScript constituting the example.
...
+
+1. Save an image (1200 x 500 pixel, PNG, under 500 KB) in `docs/img/src/`. The file name of the image **must** match the example's file name.
+2. In `docs/conf/image.config.js`, add an entry for the image. Example:
+```js
+"3d-buildings": {  // file-name
+    basename: "3d-buildings.png", // file-name.png
+    sizes: [{ width: 800 }, { width: 500 }] // always these sizes for examples
+}
+```
+3. Run `node docs/bin/appropriate-images.js file-name` to generate the appropriate images for the image or `node docs/bin/appropriate-images.js --all` to generate all images. Alternatively, when you run `npm run start-docs` the site will generate the images.
```
ryanhamley: Nit: You could replace `node docs/bin/appropriate-images.js --all` with `npm run build-images` now.

docs/conf/image.config.js
```diff
@@ -0,0 +1,389 @@
+module.exports = { // eslint-disable-line import/no-commonjs
+    "3d-buildings": {
```
ryanhamley: If the sizes are always the same and the basename is always just the name of the file in `docs/img/src` is it possible to just iterate over that folder and generate this list dynamically to avoid having to maintain a huge, ever growing list of image names?

katydecorah: That's smart! I'll rework this so that the image config is built automatically 🙇‍♀️

docs/bin/build-image-config.js
```diff
@@ -0,0 +1,24 @@
...
+        const sizes = [{ width: 800 }, { width: 500 }];
+        // set additional sizes for specific images
+        if (key === 'simple-map') sizes.push({ width: 1200 });
+        obj[key] = {
+            basename: image,
+            sizes
+        };
+    }
+    return obj;
+}, {});
```
ryanhamley: 👍 nice!

katydecorah:
Thanks for your help, everyone! I'm glad we were able to improve how we're working with images on the examples page and come up with a build system that we'll be able to use in other docs site. 

I've tested this branch on staging and the images are getting generated as expected.

mourner:
@katydecorah it seems like `yarn run build-docs` now takes 30 seconds because all images get optimized/regenerated regardless whether they changed. Can we somehow avoid that work, e.g. only optimizing newly-generated images?

andrewharvey:
> @andrewharvey Are you doing anything in particular to trigger this out of memory warning? I've had the docs running for maybe 20 minutes and haven't seen it yet. Also, what version of Webpack are you using? `yarn install` should install Webpack 3.x.x but it looks like there's a [known memory leak in Webpack 4](https://github.com/webpack/webpack/issues/6929)

@ryanhamley I'm still running into `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory` with `yarn run start-docs`, however I managed to workaround this with:

1. Run `yarn run start-docs` to build first, then Ctrl+C to exit.
2. Then run batfish with a larger max-old-space-size `node --max-old-space-size=4096 node_modules/.bin/batfish start`. After that I no longer get the heap out of memory error.

Strange that I'm the only one running into this, I wonder if we should build `--max-old-space-size=4096` into the package.json scripts...

PR #7901, opened 2019-02-14
[WIP] Advanced label formatting using ["text-section"] expression

alexshalamov:
POC for advanced label formatting proposal in https://github.com/mapbox/mapbox-gl-native/pull/13904

This PR adds `["text-section"]` expression that can be used for SymbolLayer paint properties in order to style formatted text sections.

## Launch Checklist
 - [ ] post benchmark scores

### DIFF
--- a/src/data/bucket/symbol_bucket.js
+++ b/src/data/bucket/symbol_bucket.js
@@ -523,8 +523,7 @@ class SymbolBucket implements Bucket {
         const glyphOffsetArrayStart = this.glyphOffsetArray.length;
         const vertexStartIndex = segment.vertexLength;
 
-        for (const symbol of quads) {
-
+        const addSymbol = (symbol: SymbolQuad) => {
             const tl = symbol.tl,
                 tr = symbol.tr,
                 bl = symbol.bl,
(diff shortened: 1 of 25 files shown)


alexshalamov:
@ansis could you quickly check this proposal? Maybe there is a better way of implementing advanced label styling. I was experimenting with different options:

### 1. Add paint properties to format expression
Didn't feel right, as the `["format"]` expression would need to know about all paint properties, tricky to merge / sync properties coming from `"paint": ` section with the ones from `["format"]` expression. 

### 2. Introduce `["text-section"]` expression
This PR. New expression returns formatting options for rendered section. Simplest solution (IMO), however, once many paint properties are specified for multiple sections, style spec inflates rather quickly.

```json
"layout": {
  "text-field": ["format",
                  ["get", "name_en"], { "id": "header" },
                  ["get", "description_en"], { "id": "description" }
                ],
},
"paint": {
  "text-color": ["match", 
                  ["get", "id", ["text-section"]],
                    "header", "hsl(35, 8%, 85%)",
                    "description", "#0F9D58",
                    "black"
                ]
}
```

### 3. Introduce new `"text-section"` object for SymbolLayer paint properties
Similar to `["text-section"]` expression, yet, hides section matching details from style definition and allows to define multiple text section specific paint properties in one block.

Complexity-wise, similar to (1), as there would be need to merge / sync paint property expressions, that would internally create something similar to `match` expression in option (2).

```json
"layout": {
    "text-field": ["format",
                      "First line\n", { "id": "header" },
                      "second line",  { "id": "footer" }
    ]
},
"paint": {
    "text-section": {
        "header": {
            "text-color": "red",
            "text-halo-color": "blue"
        },
        "footer": {
            "text-color": "green",
            "text-halo-width": 1
        }
    },
    "text-color": "hsl(26, 25%, 32%)",
    "text-halo-color": "hsl(0, 0%, 100%)",
    "text-halo-width": 0.5,
   }
}
```

alexshalamov:
@mourner could you please take a look at the proposal? I talked to Nicki and she was fine with the proposal (2).

ansis:
@alexshalamov sorry about blocking you on this! I agree that having `format` know about most paint properties feels weird. I agree that **2** sounds like a more extendible and maintainable path forward.

How does **2** interact with previously implemented formatting properties? Would `text-scale` and `text-font` within `[format, ...]` be deprecated?  I'm slightly worried about consistency if we don't deprecate the previous formatting approach.

It looks like in your proposal `["text-section"]` returns an object with all the formatting properties. Do you think it would be a good idea to have `["text-section"]` return just the id? That way we could use `["text-section"]` instead of `["get", "id", ["text-section"]]`.

What does `["text-section"]` return if the `text-field` has no sections?

Does `id` need to be unique?

alexshalamov:
Thanks for comments @ansis :bowing_man:

> How does **2** interact with previously implemented formatting properties? Would `text-scale` and `text-font` within `[format, ...]` be deprecated? I'm slightly worried about consistency if we don't deprecate the previous formatting approach.

It doesn't affect previously implemented properties, except adding new field for `Formatted(Options)` type. At the moment, (**2**) only applicable for paint properties. However, I'm in favor of allowing `["text-section"]` for layout properties and deprecating `text-scale` / `text-font`. Chris also mentioned that we need proper per-section `text-size` instead of `text-scale`, as there are visible artifacts when large scale factors are used.

@ansis Do you think it would be acceptable to introduce feature in 2 stages? First, only for paint properties, second layout properties + deprecation of `text-scale` and `text-font`.

> That way we could use `["text-section"]` instead of `["get", "id", ["text-section"]]`

That's a very good point, I will try rewriting POC and see how it looks.

> What does `["text-section"]` return if the `text-field` has no sections?

At the moment, empty object.

ansis:
> @ansis Do you think it would be acceptable to introduce feature in 2 stages? First, only for paint properties, second layout properties + deprecation of text-scale and text-font.

Completely acceptable!

> I think it should not be unique, would be easier to text with multiple sections that have similar style, for example multiple `<h3>` headers.

I agree. This isn't that important, but if `id` can be used for non-unique things but I'm wondering if there is a better name. `class`, `section`? I'm completely fine with `id`. It might be useful if this matched the expression, for example: `[text-section]` and `section`

alexshalamov:
@ansis Thanks for feedback, based on discussion in https://github.com/mapbox/mapbox-gl-native/pull/13904, I need to rewrite quite a few things. Will close this PR and start from scratch.

PR #7906, opened 2019-02-14
Add max width for popups

### LINKED ISSUE #1862: Add a max-width to popup containers
The default style should perform well. Without a max-width it looks like this:

popup-max-width

mzdraper:
## Launch Checklist

This adds a max width parameter for popups.

Closes #1862

### DIFF
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -67,6 +67,7 @@ export type PopupOptions = {
  * var popup = new mapboxgl.Popup({offset: popupOffsets, className: 'my-class'})
  *   .setLngLat(e.lngLat)
  *   .setHTML("<h1>Hello World!</h1>")
+ *   .setMaxWidth("300px")
  *   .addTo(map);
  * @see [Display a popup](https://www.mapbox.com/mapbox-gl-js/example/popup/)
  * @see [Display a popup on hover](https://www.mapbox.com/mapbox-gl-js/example/popup-on-hover/)
@@ -233,6 +234,12 @@ export default class Popup extends Evented {
         return this.setDOMContent(frag);
     }
 
+    /**
+     * Sets the popup's max width.
+     *
+     * @param pixel A string representing the pixel value for the maximum width.
+     * @returns {Popup} `this`
+     */
     getMaxWidth() {
       return this._container;
     }
@@ -242,19 +249,7 @@ export default class Popup extends Evented {
       this._update();
       return this;
     }
-/*
-    setMaxWidth(maxWidth: string) {
 
-        var width = document.querySelector('.mapboxgl-popup');
-        width.style.()
-        while (true) {
-
-        }
-    }
-
-    getter & setter so it can change dynamically
-    test in debug pages with complex conditions & permutations
-*/
     /**
      * Sets the popup's content to the element provided as a DOM node.
      *
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -16,7 +16,8 @@ import type {PointLike} from '@mapbox/point-geometry';
 const defaultOptions = {
     closeButton: true,
     closeOnClick: true,
-    className: ''
+    className: '',
+    maxWidth: "240px"
 };
 
 export type Offset = number | PointLike | {[Anchor]: PointLike};
@@ -26,7 +27,8 @@ export type PopupOptions = {
     closeOnClick?: boolean,
     anchor?: Anchor,
     offset?: Offset,
-    className?: string
+    className?: string,
+    maxWidth?: string
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/ui/popup.js
```diff
@@ -231,6 +234,22 @@ export default class Popup extends Evented {
         return this.setDOMContent(frag);
     }
 
+    /**
```
ryanhamley: This has to be moved above the `setMaxWidth` (or `set maxWidth` if you go the getter/setter route) method. You should also add a doc comment for `getMaxWidth`.

src/ui/popup.js
```diff
@@ -231,6 +234,22 @@ export default class Popup extends Evented {
...
+     * Sets the popup's max width.
+     *
+     * @param pixel A string representing the pixel value for the maximum width.
```
ryanhamley: The name of the parameter here in the doc should be the same as the argument to the method, which in this case is `maxWidth`. I'd also make it clear that this is setting the CSS property `maxWidth` and clarify that the expected string format is "Npx" where N is some number.

src/ui/popup.js
```diff
@@ -231,6 +234,22 @@ export default class Popup extends Evented {
...
+     * @returns {Popup} `this`
+     */
+    getMaxWidth() {
+      return this._container;
+    }
+
+    setMaxWidth(maxWidth: string) { // should this be number?
```
ryanhamley: I think it makes sense to keep this as a string since it's setting a CSS property which takes a string. I could envision taking a number as input and converting it to a string but this property doesn't have to be in pixels. It could be ems, rems, points or any other valid CSS unit of measure and we wouldn't know what the user intended if they supplied a number. It may be reasonable to default to pixels, but I think it's clearer to be explicit and let the user define the units as part of the input.

src/ui/popup.js
```diff
@@ -231,6 +234,22 @@ export default class Popup extends Evented {
...
+      this.options.maxWidth = maxWidth;
+      this._update();
+      return this;
```
ryanhamley: If you convert this to a setter method, there's no reason to `return this` at the end (because it won't be a chainable method anymore).

src/ui/popup.js
```diff
@@ -231,6 +234,30 @@ export default class Popup extends Evented {
...
+    /**
+     * Sets the popup's max width.
+     *
+     * This is setting the CSS property maxWidth. It expects a string in "Npx" format, where N is some number.
+     *
+     * @param maxWidth A string representing the pixel value for the maximum width.
+     * @returns {Popup} `this`
+     */
+    set maxWidth(maxWidth: string) {
+        console.log(this.options.maxWidth);
```
ryanhamley: Remove these two console.log statements

src/ui/popup.js
```diff
@@ -275,13 +302,16 @@ export default class Popup extends Evented {
...
             this._tip       = DOM.create('div', 'mapboxgl-popup-tip', this._container);
             this._container.appendChild(this._content);
-
             if (this.options.className) {
                 this.options.className.split(' ').forEach(name =>
                     this._container.classList.add(name));
             }
         }
 
+        if (this._container.style.maxWidth != this.options.maxWidth) {
```
ryanhamley: This should be `!==`

ryanhamley:
@mzdraper there's a Flow error that you'll have to fix before you can merge

```
/root/mapbox-gl-js/src/ui/popup.js
  310:1  error  Expected indentation of 12 spaces but found 10  indent
```

viniciuskneves:
I think it could be achieved by setting a class to `Popup`, like

```javascript
new mapboxgl.Popup({
  ...
  className: 'some-class'
})...
```

And then you could customize this class on your CSS as

```css
.some-class {
  max-width: 240px;
}
```

Just updated Mapbox and it breaks a lot my map. Now you enforce some `max-width` which wasn't required before and as a breaking change it should have been released a major version and not minor.

I would like to updated Mapbox to get all the benefits from it without such a burden as it covers only a specific case where there is a nice and supported workaround already.

I had to add `maxWidth: 'none'` to all my popups now as it broke all my use cases 😢

mourner:
> as a breaking change it should have been released a major version and not minor

The accepted convention on semver is that 0.x versions may introduce breaking changes with a minor version change. This was released as a part of v0.54.0 and [documented as a breaking change](https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.54.0). So we did this right.

viniciuskneves:
> > as a breaking change it should have been released a major version and not minor
> 
> The accepted convention on semver is that 0.x versions may introduce breaking changes with a minor version change. This was released as a part of v0.54.0 and [documented as a breaking change](https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.54.0). So we did this right.

Sorry @mourner my bad, I was with 1.0 release in mind. Nevertheless I don't want to focus on it here.

My main input is regarding the feature released that breaks stuffs and I think it doesn't add much to the library as there was another solution for the very same problem already implemented.

Issue #3610, opened 2016-11-14
map.getCenter() returns an internal reference, not a clone


blq:
Here's a, I'd say, reasonable use case:
```
var center = map.getCenter();
// I  want to create a marker or popup or such _at an offset_
center.lat += 1;
popup.setLngLat(center);
...
```
=> map's _internal_  transform center is now affected (compromised even), since popup now accidentaly share state with the map.

I suggest returning a cloned `LngLat `from` getCenter() ` would be more intuitive and less risk of subtle bugs.

This issue might be present in other API calls too, and yes, references are a generic JavaScript "problem" of course.

/ Fredrik Blomqvist

PR #7922, opened 2019-02-18
Return clone instead of a reference on `map.getCenter()` #3610

stepankuzmin:
Return clone instead of a reference on `map.getCenter()` according to #3610
Is it correct to use `util.clone` in this case?


stepankuzmin:
I've changed `util.clone` to `new LngLat()`.

Issue #7927, opened 2019-02-19
Benchmarks are broken by RemovePaintState benchmark


andycalder:
It appears that the "version" benchmarks have been broken since b01a5d3bc78662de493f800ec9430cb7ea42555f. The bug made it in to the v0.53.0 release so it's impossible to benchmark feature work against master or release. I have traced the problem to a bad export [here](https://github.com/mapbox/mapbox-gl-js/blob/c58b24c3e07f192dbe5d7ea9909f6b30f80e9520/bench/benchmarks/remove_paint_state.js#L113) (should be `export default`). The benchmark looks like it might be incomplete because there are three unused classes.

**mapbox-gl-js version**: master and v0.53.0

**browser**: All

### Steps to Trigger Behavior

 1. Start the benchmarks server.
 2. Go to http://localhost:9966/bench/versions.

### Expected Behavior

No error.

### Actual Behavior

`TypeError: RemovePaintState$1 is not a constructor`.

ryanhamley:
Thanks for reporting this @andycalder. This is fixed now but the release branch will contain the bug until our next official release in a couple weeks. You can work around the release branch for the time being by changing https://github.com/mapbox/mapbox-gl-js/blob/master/bench/versions/index.html#L22 to `.then(pkg => [ 'master']))`. This will then run your current branch only against `master` and you shouldn't see any further errors.

PR #7930, opened 2019-02-20
Fix RemovePaintState benchmarks

ryanhamley:
## Launch Checklist

    - Iterates over individual removePaintState benchmark tests
    - fixes https://github.com/mapbox/mapbox-gl-js/issues/7927

screen shot 2019-02-19 at 4 12 49 pm


ryanhamley:
@peterqliu You should see the error twice now, once for `master` and once for the latest release branch. Merging this to `master` should publish a new version of that to S3 so we'll be down to 1 error. Unfortunately, that won't be resolved until we publish our next release in a couple weeks.

You can work around the release branch for the time being by changing https://github.com/mapbox/mapbox-gl-js/blob/master/bench/versions/index.html#L22 to `.then(pkg => [ 'master']))`

PR #7934, opened 2019-02-20
Update featureFilter README

fc:
Update require statement in example for how to use `featureFilter`.  Now this can be used as a reference for writing tests against the `featureFilter`.


fc:
@mourner agreed, updated

Issue #7865, opened 2019-02-01
setLayerZoomRange causing rasters to be re-requested


nreese:
Whenever `setLayerZoomRange` is called with new min/max values on a layer containing raster tiles, all tiles are requested. Is this expected behavior? I am using mapbox-gl version `0.45.0` in Chrome 71.0.3578.98. 

Below is the code used to create the layer.

```
    mbMap.addSource(sourceId, {
      type: 'raster',
      tiles: [url],
      tileSize: 256,
      scheme: 'xyz',
    });

    mbMap.addLayer({
      id: mbLayerId,
      type: 'raster',
      source: sourceId,
      minzoom: 0,
      maxzoom: 22,
    });
```

Then if `mbMap.setLayerZoomRange(mbLayerId, 0, 15);` is called, all tiles are re-fetched. Why are the tiles re-fetched? The map is at zoom level 3 so the change should have no effect on the current tiles.

mourner:
@nreese 
- Did you make sure the behavior is the same on the latest version, v0.52.0 or v0.53.0-beta.1?
- Can you set up a minimal reproducible test case to demonstrate this (e.g. with JSFiddle)?

nreese:
@mourner The problem exists in `v0.52.0`. Here is a js-fiddle that demonstrates the problem https://jsfiddle.net/reesenathan/pwb6nz74/. Adjust the slider in the top left corner and notice that when you release the slider and `setLayerZoomRange` is triggered, the tiles flicker because they are getting re-fetched.

peterqliu:
Looks like killing this line removes the flicker, while preserving the method functionality:

https://github.com/mapbox/mapbox-gl-js/blob/ead961f7c6b903bb5bbfbc3f128362bb46d421eb/src/style/style.js#L735

does anything rely on this line running, that I'm not seeing?

PR #7942, opened 2019-02-22
[ready] address raster flickering

peterqliu:
Solves https://github.com/mapbox/mapbox-gl-js/issues/7865 by removing source reload after `setLayerZoomRange`.  Seems to solve flickering without breaking anything, or any extant tests

## Launch Checklist


peterqliu:
Added integration tests that **lower** and **raise** zoom ranges, **into** and **out of** the current zoom. For each test, also added a parallel test for raster layers.

peterqliu:
Turns out that I couldn't just remove the line -- without it, vector layers rendered fine only when `setLayerZoomRange` invokes in the same render loop as when the layer was added. But when executed a later time, the vector layer does rely on the update function.

Instead of outright deletion, I'm skipping `_updateLayer` only for raster layers.

peterqliu:
>If that is happening already, what alternative update mechanism is affecting this change?

@asheemmamoowala the update is mediated upstream in map.js [here](https://github.com/mapbox/mapbox-gl-js/blob/5dae56219bcf6942b81b4c455748f3754f7cdeba/src/ui/map.js#L1356). The map `render` event fires because of this (and doesn't fire in absence of this), and does so regardless of whether [_updateLayer()](https://github.com/mapbox/mapbox-gl-js/blob/5dae56219bcf6942b81b4c455748f3754f7cdeba/src/ui/map.js#L1356) gets invoked in style.js.

Through various tests, the latter command shows to be nearly completely redundant, _except_ when updating only _vector_ sources to only  _less restrictive_ zoom ranges. To stay on the safe side, this PR bypasses that line only for raster sources.

peterqliu:
further digging reveals that `this._updatedSources` is usually an empty object, causing this line to evaluate to true https://github.com/mapbox/mapbox-gl-js/blob/5dae56219bcf6942b81b4c455748f3754f7cdeba/src/style/style.js#L942, which in turn causes the reload that causes the raster flicker.

Instead of conditionally calling the `_updateLayer()` function, I'm culling only the lines that set the `changed` and `update` flags, and invoking those for all source types https://github.com/mapbox/mapbox-gl-js/pull/7942/commits/1a1a1dd73c91b51623cfa57006a888b0ddd248e4#diff-0fcf9ef37a1614185cad0dcbd5ec5529L741. This avoids the raster flicker, while ensuring that all zoom range updates work as expected.

debug/zoomrange.html
```diff
@@ -0,0 +1,100 @@
+<!DOCTYPE html>
```
asheemmamoowala: This file doesn't seem like a good addition to the `debug/` folder. Mind removing it?

test/integration/render-tests/zoom-visibility/was-above/style.json
```diff
@@ -1,22 +1,35 @@
 {
   "version": 8,
+  "center": [
```
asheemmamoowala: Why are changes needed to this test? the unchanged test seems to work correctly even with the fix introduced in this PR.

peterqliu:
>What is the change in test/integration/render-tests/zoom-visibility/in-range/expected.png ? Is it expected or just a small delta that should be ignored/reverted?

reverted 👍 

>Also noticed that the expected results for the new now-above and now-below tests are empty transparent pngs. These tests seem to pass even without your changes in style.js.

I think this should be fine, as there was never an issue around flipping off layers that are now out of range. The tests are meant to guard against future regressions

asheemmamoowala:
Did some further testing of this PR, and it introduces a subtle bug that breaks for vector layers where tiles do need to be reloaded when the zoom range changes.

Closing in favor of #8854

PR #7950, opened 2019-02-25
mapbox-gl-geocoder@latest in examples

scottsfarley93:
## Launch Checklist

Bumps the version of mapbox-gl-geocoder in the examples to `@latest` to more easily propagate bug fixes and new features.


andrewharvey:
> Using `latest` for versioned JS libraries is an anti-pattern, because all examples will break after of an incompatible (sometimes non-obvious) change. It's always preferable to have strictly versioned CDN links and upgrade explicitly.

I agree we should continue to pin explicit versions. It helps protect against bugs that slip through releases and allows us to make breaking changes in the future without breaking downstream applications.

For example v3.0.0 introduced a breaking API change and v3.1.0 broke the proximity option, both of which pinning version numbers helped mitigate any damage.

I'm :+1: for either 1. still requiring a review by someone other than the submitter to bump versions, or 2. ensure that all mapbox-gl-geocoder releases go through a review. Yes it does mean you can't move as fast, but it helps catch issue before they are released instead of scrambling to fix them up after releasing.

scottsfarley93:
👍 good points all around. Will close this in favor of a PR that updates to the latest version.

PR #7951, opened 2019-02-25
add docs for CustomLayerInterface#renderingMode

ansis:
Add some more documentation for custom layers. Document:
- `renderingMode` and the difference between 2d and 3d
- blending functions

### DIFF
--- a/src/render/draw_custom.js
+++ b/src/render/draw_custom.js
@@ -19,6 +19,7 @@ function drawCustom(painter: Painter, sourceCache: SourceCache, layer: CustomSty
         const prerender = implementation.prerender;
         if (prerender) {
             painter.setCustomLayerDefaults();
+            context.setColorMode(painter.colorModeForRenderPass());
 
             prerender.call(implementation, context.gl, painter.transform.customLayerMatrix());
 
--- a/src/style/style_layer/custom_style_layer.js
+++ b/src/style/style_layer/custom_style_layer.js
@@ -20,6 +20,11 @@ type CustomRenderMethod = (gl: WebGLRenderingContext, matrix: Array<number>) => 
  * and they should appropriately handle {@link Map.event:webglcontextlost} and
  * {@link Map.event:webglcontextrestored}.
  *
+ * The `renderingMode` property controls whether the layer is treated as a `"2d"` or `"3d"` map layer. Use:
+ * - `"renderingMode": "3d"` to use the depth buffer and share it with other layers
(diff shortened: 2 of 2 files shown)


src/render/draw_custom.js
```diff
@@ -19,6 +19,7 @@ function drawCustom(painter: Painter, sourceCache: SourceCache, layer: CustomSty
         const prerender = implementation.prerender;
         if (prerender) {
             painter.setCustomLayerDefaults();
+            context.setColorMode(painter.colorModeForRenderPass());
```
mourner: Nit: this isn't a docs change.

ansis: Ahh, woops. I was going to commit that separately. This adds a default blending value for `prerender` so that it is more consistent with `render`. Should I split this into a separate commit?

src/style/style_layer/custom_style_layer.js
```diff
@@ -20,6 +20,11 @@ type CustomRenderMethod = (gl: WebGLRenderingContext, matrix: Array<number>) =>
  * and they should appropriately handle {@link Map.event:webglcontextlost} and
  * {@link Map.event:webglcontextrestored}.
  *
+ * The `renderingMode` property controls whether the layer is treated as `"2d"` or `"3d"` map layer. Use:
```
ryanhamley: very minor nit: `whether the layer is treated as **a** "2d"....`

I have a question on semantics. We don't typically talk about "3d" because as far as I know we don't really support true 3d rendering right now. Have we purposefully avoided the terminology to prevent confusion over the library's capabilities or is the difference between "3d" and something like "extruded" mostly meaningless?

ansis: > Have we purposefully avoided the terminology to prevent confusion over the library's capabilities or is the difference between "3d" and something like "extruded" mostly meaningless?

I think it maybe helps prevent confusion when we talk about the core layer types because fill-extrusion is limited enough that saying "3d" could be misleading. But custom layers can be fully 3d. In this context it refers to whether the things drawn by the layer exist in 3d space, which is the case for both extrusions and some custom layers

ryanhamley: In other words, "does this layer have a meaningful Z coordinate"? That makes sense. I assumed that custom layers were flexible enough to make 3D possible, but just wanted to clarify for my own knowledge.

ansis: Yep, that's a clearer way of putting it

PR #7967, opened 2019-02-28
Optimize matrix calculation for labels

mourner:
I noticed that `getLabelPlaneMatrix` and `getGlCoordMatrix` functions were taking considerable time on the main thread, and made matrix caching more aggressive to avoid a lot of this work, at least for screen-aligned labels which dominate the map. After the changes they take 10x less time.

We could cache even more (and I might follow up on this), but this PR gets us most of the way there with minimal changes.

Before:
image

After:
image

Bench:
image

Curiously, paint benchmark is nearly unaffected.

 - [x] post benchmark scores

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -37,6 +37,8 @@ class Transform {
     alignedProjMatrix: Float64Array;
     pixelMatrix: Float64Array;
     pixelMatrixInverse: Float64Array;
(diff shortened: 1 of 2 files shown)


pakastin:
Wow, this makes such a huge difference in Aviamaps!
https://aviamaps.com/map

Awesome work! 👌

PR #7969, opened 2019-02-28
Optimize setUniforms

mourner:
Eliminates unnecessary loops and object lookups in a performance-critical method (`setUniforms`). Measuring this with DevTools profiler, the method seems to get 2.5x faster — from taking 9.1% of CPU time to 3.7%.

Before:
image

After:
image

However, the Paint benchmark doesn't show this improvement — once again implying that it's not representative of real world performance (which needs to be investigated separately).

image

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/data/program_configuration.js
+++ b/src/data/program_configuration.js
@@ -14,7 +14,6 @@ import {
     Uniform1f,
     UniformColor,
     Uniform4f,
-    type UniformBindings,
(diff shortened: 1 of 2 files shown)


mourner:
Surprisingly fails one render test (regressions #5978) — investigating.

PR #7978, opened 2019-03-01
update mapbox-gl-geocoder to @3.1.4 in examples

scottsfarley93:
## Launch Checklist

Upgrades mapbox-gl-geocoder in examples to v3.1.4. 
 
\cc @mapbox/docs


mourner:
No idea why the `test-flow` test fails, even after rerunning a few times. Probably OK merging anyway.

PR #7985, opened 2019-03-04
Fix Transform#getCameraPoint

andycalder:
This method should use sin instead of tan.

@ansis


andycalder:
Investigating the failing query test...

andycalder:
My mistake, it looks like the code is correct

Issue #6231, opened 2018-02-26
Adding symbol icon layer before adding icons causes icons not to render at initial zoom


BartCorremans:
Apologies for the long title, this is a bit of a strange one and it took me a while to pin down and consistently reproduce.

Reproduction: https://jsbin.com/wigirelagu/edit?html,output (add your access token at the top of the script)

This is based on https://www.mapbox.com/mapbox-gl-js/example/add-image/ with a few changes.

**mapbox-gl-js version**: 0.44.1

### Steps to Trigger Behavior

* Create a symbol layer that uses image icons
* Add this layer to your map before you add the images that that layer uses

### Expected Behavior

Icons appear at any zoom level and regardless of when the layer was added.

### Actual Behavior

Icons don't appear on the map unless you zoom away from the zoom level that the map initially loads with. This applies to the entire integer zoom level; since the reproduction has an initial zoom of 7, the icons will be hidden when your zoom level is anywhere between 7 and 8.

Adding the layer after the icons are added works around the issue (you can try this in the repro), but it seems like this should work either way.

mollymerp:
@BartCorremans thanks for the report and reproduction!

mollymerp:
I don't think that adding a layer with a symbol that has not been added to the map _should_ work, but we can definitely make error reporting better here, and avoid the behavior you're seeing where it just skips rendering the symbol on the initial zoom level.

BartCorremans:
Thanks for the reply! I understand that it might indeed not make sense for this to actually work in the first place, though I've found it convenient to be able to just add layers at any point without having to take into account that the loadImage/addImage workflow is async, and know that the icons will be rendered as soon as the images load.

anandthakker:
> I don't think that adding a layer with a symbol that has not been added to the map should work

I'd have expected this to work: an icon layer might refer mostly to icons that exist in the sprite sheet, but to a few that have to be added at runtime for some reason.

jfirebaugh:
In general, the style specification and runtime styling APIs requires that ID-based references (`source` IDs in layers, `icon-image` and `*-pattern` image IDs in layer properties, and source and layer IDs passed to runtime styling APIs) are valid -- referring to objects that actually exist. This helps catch programming mistakes and avoids undue complexity in the implementation.

[...]

PR #7987, opened 2019-03-04
add `styleimageneeded` event for on-demand images

### LINKED ISSUE #7587: feature request: a way to conditionally provide icon images
## Motivation

There are currently two ways to provide images for icons and patterns: the sprite and `addImage`. These work well when you have a limited number of images or when you know ahead of time which images you need. These approaches do not work when there are many possible images and you don't know which you need, for example user icons.

We have a way to *push* images. We need a way to *pull* images.

Example:
You have a vector tile source that contains user locations. You want to set a unique icon image for each user depending on a *userid* data property. This is currently not possible there are too many icons to add in the sprite and because you can't use `addImage` because you don't know which icons are needed.

A solution to this problem could be also useful for rendering custom icons based on feature properties.

### A) Do nothing

Force users to load the data separately as geojson and have them manually iterate through features and add the correct icons before setting the data.

Cons:
- does not work at all for vector tile sources
- adds complexity to geojson setting and updating, especially on -native where json isn't as simple to use

### B) Support loading images from urls

```
"icon-image": ["load-image", ["concat", "http://example.com/", ["get", "userid"], ".png"]]
```

The `transformRequest` map option could be used by users to transform custom uri schemes to regular urls. For example, `my-scheme://images/123` to `http://example.com/images/`.

Cons:
- doesn't allow for local icon rendering
- can't use external code (that handles auth, etc) for loading images
- lets people use separate images everywhere in the style which could let people hurt their load performance too easily

### C) event for missing images

An event that gets fired *after* noticing we need an icon but *before* actually trying to use it. This lets the user add an image within the callback.

```
"icon-image": ["load-image", ["concat", "my-user-image", ["get", "userid"]]
```

```js
map.on('imagemissing', function(name) {
    if (name.indexOf('my-user-image') === 0) {
        const img = document.createElement('img');
        const userid = name.replace('my-user-image', '');
        img.src = 'http://example.com/image/' + userid + '.png';
        map.addImage(name, image);
    }
});
```
```js
map.on('imagemissing', function(name) {
    if (name.indexOf('my-user-image') === 0) {
        const data = new Uint8Array(...);
        // render icon into `data`
        map.addImage(name, { width: ..., height: ..., data: data });
    }
});
```

We would need some way of cleaning up images after they aren't needed anymore. One possible way is to add a third `addImage` parameter that lets the map drop it when it isn't used anymore: `addImage(name, img, dropWhenUnused)`.

We would also need support for async images.

Cons:
- the developer needs to encode all data into a string and then parse it back out of a string. This gets more complicated if your icon depends on multiple properties

### D) add image rendering callback with structured data

```
"icon-image": ["render-image", "mynamespace", ["get", "userid"], ["feature-state", "hover"]]
```
```
map.addImageRenderer("mynamespace", function(userid, hover) {
    ...
    return img;
});
```

Cons:
- new expression, new api method
- typing could be complicated on -native

### E) user-implemented expressions

D) allows for user code to be called by expressions indirectly. We could take this further and allow users to register their own expressions.
```
map.registerExpression("generate-user-image": {
    evaluate: function(arguments, evalContext) {
        // return image
    }
});
```

```
"icon-image": ["generate-user-image", ["get", "userid"]]
```

Cons:
- a much broader impact and riskier than the other options
- a larger interface
- tricky to run user code in workers without breaking csp compliance

## Design

I think we should implement *C) event for missing images*. It does what we need with the least amount of changes. It seems like something similar could be added on iOS (cc @1ec5) with a delegate and on Android with listener. There is already precedent for this on iOS with `-mapView:imageForAnnotation:`.

Options D and E feel over designed. B works only in some cases and the possible performance trap it would introduce worries me.

### Concepts

C) also has the least amount of new concepts introduced. We would teach this design by adding an example that fetches a custom image based on data and an example that renders an image locally. No new terminology would need to be introduced.

### Implementation

ImageManager#getImages would fire the event when images are missing. ImageManager would also check if the image has loaded, and if not, delay until it has. The implementation in -native should be very similar.

---

@ChrisLoer does `C)` sound good?

ansis:
This is the first of several dynamic icon prs.

The event gets fired when a layer needs an image that the map doesn't have. The developer has a chance to listen for this event and add an image within the callback in order to have it included.

This event can be used to dynamically generate icons based on feature properties.

fix #7587

```js
map.on('styleimageneeded', function(e) {
    var image = generateImage(e.id);
    map.addImage(e.id, image);
});
```

See the committed `add-image-missing-generated` example for a full version.

## Naming

I'm pretty ok with `styleimageneeded`. Other alternatives were:
- `styleimagemissing`
- `imageneeded`
- `imagemissing`

Do you prefer any of these?

## Launch Checklist

 - [n/a] post benchmark scores

### DIFF
--- a/docs/pages/example/add-image-missing-generated.js
+++ b/docs/pages/example/add-image-missing-generated.js
@@ -1,0 +1,11 @@
+/*---
+title: Generate and add a missing icon to the map
+description: Dynamically generate a missing icon at runtime and add it to the map.
+tags:
+  - styles
+  - layers
+pathname: /mapbox-gl-js/example/add-image-missing-generated/
+---*/
+import Example from '../../components/example';
+import html from './add-image-missing-generated.html';
+export default Example(html);
--- a/src/ui/events.js
+++ b/src/ui/events.js
@@ -779,6 +779,20 @@ export type MapEvent =
     | 'sourcedataloading'
 
     /**
+     * Fired when an icon or pattern needed by the style is missing. The missing image can
+     * be added with {@link Map#addImage} within this event listener callback to prevent the image from
+     * being skipped. This event can be used to dynamically generate icons and patterns.
+     *
+     * @event styleimagemissing
+     * @memberof Map
+     * @instance
+     * @property {string} id The id of the missing image.
+     *
+     * @see [Generate and add a missing icon to the map](https://mapbox.com/mapbox-gl-js/example/add-image-missing-generated/)
+     */
+    | 'styleimagemissing'
+
+    /**
      * @event style.load
      * @memberof Map
      * @instance
--- a/docs/pages/example/add-image-missing-generated.html
+++ b/docs/pages/example/add-image-missing-generated.html
@@ -1,0 +1,81 @@
+<div id='map'></div>
+
+<script>
+
+var map = new mapboxgl.Map({
+    container: 'map',
+    style: 'mapbox://styles/mapbox/streets-v9'
+});
+
+map.on('styleimagemissing', function(e) {
+    var id = e.id; // id of the missing image
+
+    // check if this missing icon is one this function can generate
+    var prefix = 'square-rgb-';
+    if (id.indexOf(prefix) !== 0) return;
+
+    // extract the color from the id
+    var rgb = id.replace(prefix, '').split(',').map(Number);
+
+    var width = 64; // The image will be 64 pixels square
+    var bytesPerPixel = 4; // Each pixel is represented by 4 bytes: red, green, blue, and alpha.
+    var data = new Uint8Array(width * width * bytesPerPixel);
+
+    for (var x = 0; x < width; x++) {
+        for (var y = 0; y < width; y++) {
+            var offset = (y * width + x) * bytesPerPixel;
+            data[offset + 0] = rgb[0]; // red
+            data[offset + 1] = rgb[1]; // green
+            data[offset + 2] = rgb[2]; // blue
+            data[offset + 3] = 255;    // alpha
+        }
+    }
+
+    map.addImage(id, {width: width, height: width, data: data});
+});
+
+map.on('load', function () {
+    map.addLayer({
+        "id": "points",
+        "type": "symbol",
+        "source": {
+            "type": "geojson",
+            "data": {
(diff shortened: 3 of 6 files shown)


src/ui/events.js
```diff
@@ -778,6 +778,20 @@ export type MapEvent =
      */
     | 'sourcedataloading'
 
+    /**
+     * Fired when an icon or pattern needed by the style is missing. The missing image can
+     * be added with {@link Map#addImage} within this callback to prevent the image from
```
ryanhamley: Maybe better to be explicit here -> `within the event listener callback`

docs/pages/example/add-image-missing-generated.js
```diff
@@ -0,0 +1,11 @@
+/*---
+title: Generate and add a missing icon to the map
+description: Add a missing icon to the map that was generated at runtime.
```
ryanhamley: Nit: in `the map that was generated at runtime`, "that" appears to reference "map". Maybe something like `Dynamically generate a missing icon at runtime and add it to the map`?

src/ui/events.js
ansis: yep, good idea!

docs/pages/example/add-image-missing-generated.js
ansis: yep, that's better. Thanks!

ansis:
Thanks for the suggestions and feedback. I'll merge this as `styleimagemissing`

joewoodhouse:
#6231 for me renders this feature unusable (though I love the idea of it). Currently that bug still exists for me in my use-case - the icons never properly render at the initial zoom level when I add them via the `styleimagemissing` event. As the initial reports say, they seem to mostly render fine at all other zoom levels, which makes me think this must surely just be a bug that needs fixing? Even if I heavily delay adding the layers that have the missing icons, they still don't render properly.

mourner:
@joewoodhouse can you submit a new bug report with a minimal reproducible test case for this (e.g. using JSFiddle)?

joewoodhouse:
@mourner I've created https://github.com/mapbox/mapbox-gl-js/issues/8335 which includes a codepen to recreate the issue.

Issue #7904, opened 2019-02-14
"Uncaught TypeError: e is not a function at vu.receive" in Chrome 65 or earlier


aleksejleonov:
**mapbox-gl-js version**:

0.53.0

**browser**:

Chrome 65 or earlier version. Could be reproduced both on Win10 and Mac OS Mojave.
Starting Chrome 66 can't reproduce error.

### Steps to Trigger Behavior

 1. Open single-map example https://docs.mapbox.com/mapbox-gl-js/example/simple-map/ in Chrome 65 (or earlier version).
 2. Make several rapid zoom in/out (with touchpad) with large zoom level change (from city to planet level).

### Link to Demonstration

https://docs.mapbox.com/mapbox-gl-js/example/simple-map/

### Expected Behavior

No JS errors in console. 

### Actual Behavior

`Uncaught TypeError: e is not a function at vu.receive (actor.js94)` errors in browser console.

<img width="1498" alt="screen shot 2019-02-14 at 4 33 29 pm" src="https://user-images.githubusercontent.com/241582/52794605-cae93280-3078-11e9-8671-e5733f2cfddf.png">

ryanhamley:
Also reported here with specific information on the cause of this bug: https://github.com/mapbox/mapbox-gl-js/issues/7905#issue-410354816

I believe the bug was introduced in https://github.com/mapbox/mapbox-gl-js/commit/913d3752444556d72460d1060048c9b428ef1715

asheemmamoowala:
cc @mourner

mourner:
> I believe the bug was introduced in 913d375

Nope — I ran `git bisect` while testing on older Chrome (using [these instructions](https://stackoverflow.com/a/19672729/244789)), and the culprit is #7818 (cc @kkaefer). Which is plausible given changes in the `actor.js` code, although I have no idea what changed in Chrome v65 that made the bug disappear.

mourner:
Fixed in #7988. Looks like IE11 was affected too.

Issue #7905, opened 2019-02-14
v0.53.0 JavaScript error mapbox-gl.js:29 Uncaught TypeError: e is not a function


jwoodwardtfx:
It seems that when fast panning the map on v0.53.0 I sometimes get a Javascript error. It seems to be related to cancelling tile requests that are in progress.

_mapbox-gl.js:29 Uncaught TypeError: e is not a function
    at vu.receive (mapbox-gl.js:29)_

It appears that within "actor.js" it gives an error in the "receive" function:

```
if (data.type === '<response>' || data.type === '<cancel>') {
        callback = this.callbacks[data.id];
        delete this.callbacks[data.id];
        if (callback && data.error) {
            callback(deserialize(data.error));
        } else if (callback) {
**-- ERROR HERE WHEN TRYING TO USE OBJECT CALLBACK AS A FUNCTION ---**
              callback(null, deserialize(data.data));
        }
    }
```

Here **_callback_** is actually an object which contains a single function called "cancel".

It seems that this may have originally come from "makeXMLHttpRequest" (which returns an object containing a single method called cancel).

This didn't happen on v0.52.0.

Thanks,
Joe

ryanhamley:
Are you seeing this in a specific browser(s)? https://github.com/mapbox/mapbox-gl-js/issues/7904 reports the same issue but specifically on Chrome > 66

jwoodwardtfx:
I reproduced this bug on a Chromium build, reporting itself as "Chrome/62.0.3202.9".

PR #7988, opened 2019-03-05
ensure we're calling the correct cancellation function

kkaefer:
Fix for https://github.com/mapbox/mapbox-gl-js/issues/7904

### DIFF
--- a/src/util/actor.js
+++ b/src/util/actor.js
@@ -102,7 +102,7 @@ class Actor {
             if (cancelable && this.callbacks[data.id] === null) {
                 // Only add the cancelable callback if the done callback wasn't already called.
                 // Otherwise we will never be able to delete it.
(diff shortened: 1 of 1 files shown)


jacknkandy:
Could someone confirm whether this fix should be available in the latest release 0.53.1. I have checked the commit history for 0.53.1 and can see this fix in the list.

Some of our customers are experiencing the issue reported #7904 / #7905. It appears that these users are running Chrome 64 whereas our other customers are running more recent versions. The update to 0.53.1 has not seemed to fix this issue.

chloekraw:
@jacknkandy - you are correct. This fix will be out in the 0.54-beta this week and final release next week.

jacknkandy:
Hi @chloekraw Thats awesome! Thanks for confirming and providing these details.

PR #7999, opened 2019-03-06
add support for animated images

### LINKED ISSUE #4804: Add updateImage API, or allow addImage to update
There's currently no API for updating the image associated with a particular ID. gl-native allows updating an image via `addImage`, but in gl-js that triggers an error.

When implementing this, care should be taken to avoid the gl-js equivalents of https://github.com/mapbox/mapbox-gl-native/issues/6536 and https://github.com/mapbox/mapbox-gl-native/issues/7618.

### LINKED ISSUE #7588: updatable and animatable images
## Motivation

Images can be used as icons, fill-patterns and line-patterns. Images can currently only be static (with some update support on -native). Being able to animate these images could help make the map more dynamic, emphasize certain features and invite interaction.

Dynamic images could be used for:
- playing a video as a polygon background
- playing a video as an icon
- animating an icon using custom rendering code
- modifying icons in response to user interaction

### A) push update + beforerender event

Native supports updating images by calling `map.addImage` again with the same name. We probably want to do the same in -js. Or something similar and add a separate method for updating images:

This could be used for animating icons if we also add a `renderstart` event to let users push updates before each frame.

```js
map.addImage(name, image);
map.on('renderstart', function() {
    if (animating) {
        map.updateImage(name, newimage);
    }
});
```

This is the simplest design alternative but leaves more to the user. The user needs to handle adding and removing event listeners when playing/pausing animation.

An alternative name for the even could be `renderstart`.

### B) pull updates with events

We could define a more fleshed out `UserImage` interface. Users could (optionally) inherit from `Evented` when defining their image. We could listen to `play`, `pause`, `seek`, `reset` events and pull in updates when we need to by calling a `getImageData` method that would render the image.

```js
class MyAnimatedIcon extends mapboxgl.Evented {
    constructor() { this.paused = true; }
    play() { this.fire('play'); this.paused = false;}
    pause() { this.fire('pause'); this.paused = true; }
    getImageData() { /* return rendered image. Called on each frame while icon is playing. */ }
}

const icon = new MyAnimatedIcon(...);
let playing = false;
map.addImage('animated-icon', icon);

...

function onClick() {
    if (icon.paused) {
        icon.play();
    } else {
        icon.pause();
    }
}
```

### C) pull updates with polling

Similar to B), but instead of events have a single boolean `animating` property that the map polls on every frame. This might be slightly less performant for us but simpler for users to implement.

```js

const image = {
    getImageData: function() { /* render image */ },
    animating: false
}
map.addImage('animated-icon', image);

...

function onClick() {
    image.animating = !image.animating;
}
```

## GL icon rendering design

All the above options could be extended to support rendering icons with gl. This could provide better performance for some types of animation. The implementation would be similar to custom layers.

Since framebuffer binding is relatively expensive we would want to be able to render multiple icons into a single texture. 

### GL A) user provides texture and coordinates

The user would create an atlas themselves and provide us with a texture as well as the coordinates of the icon. We would then copy the icons into our internal atlas.

```
const image = {
    width: 10,
    height: 10,
    data: glTexture,
    x: 125,
    y: 300
};
```

### GL B) user provides render method

The user would provide a method for rendering the icon with gl, similar to how custom layers work. We would allocate space for the icon in a texture and use `gl.scissor()` and `gl.viewport()` to restrict drawing to that space.

```
const image = {
    width: 10,
    height: 10,
    render: function(gl) {
    }
};
```

## Design

I think we should implement design `1A)`. The update image method would be useful one of updates and -native parity. The `renderstart` event would be simple to add and maintain.

B and C could be useful later but adding them now could be overdesigning. I'd prefer C to B right now because it is simpler.

I think we can initially not support rendering with GL and only consider this if cpu rendering turns out to be a performance issue for our users. GL rendering would come with the same concerns as custom layers on -native (how switching to metal would be handled, etc).

### Mock-Up

```js
const image = {
    width: 10,
    height: 10,
    data: renderIcon(userid);
};

map.addImage("asdf", image);

map.on('renderstart', function() {
    image.data = renderIcon(userid, Date.now());
    map.updateImage("asdf", image);
}
```

example with video:
```js
const video = document.createElement('video');
video.src = '...';
map.addImage('asdf', video);
function updateVideo() {
    map.updateImage('asdf', video);
}
video.on('play', () => map.on('renderstart', updateVideo));
video.on('pause', () => map.off('renderstart', updateVideo));
```

### Concepts

Design `1A` does not introduce any new concepts. We'll teach this design with examples that use `beforerender` to update an image.

### Implementation

The `renderstart` event would fire before the rendering begins for a frame.

`map.updateImage` would update `ImageManager` as well as any sprites that are already using that image. It would also need to correctly handle sprites that are being created in workers during that time.

---

@ChrisLoer does design option 1) with no gl rendering sound good for an initial implementation?

ansis:
This adds support for updating and animating images. fixes #7588 and fixes #4804

There are two parts to this:

### StyleImageInterface

```js
export type StyleImageInterface = {
    width: number,
    height: number,
    data: Uint8Array | Uint8ClampedArray,
    render?: () => void,
    onAdd?: (map: Map, id: string) => void,
    onRemove?: () => void
};
```
`map.addImage(...)` now accepts images that implement the `StyleImageInterface`. This interface adds three new and optional methods:
- `render` gets called once for every map frame that the icon is used in the visible map. The user can update the icon in this method to animate the icon.
- `onAdd` gets called when the icon gets added to the map to allow for resource initialization, etc.
- `onRemove` mirrors onAdd and will be useful if/when we add image eviction

This pull-based approach to image updating allows the user to only rerender the image when it is necessary. The image is not rerendered for cached tiles. Image renders are synchronized with map renders.

### map.updateImage(id, image)

This updates and image and immediately updates map. This is a lighter alternative to the above `render` approach and should be useful for instant, infrequent updates. For example, changing an icon on click. Or changing an icon when a resource has loaded.

Both approaches use the same underlying implementation.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/render/image_atlas.js
+++ b/src/render/image_atlas.js
@@ -63,40 +63,13 @@ export default class ImageAtlas {
     uploaded: ?boolean;
 
     constructor(icons: {[string]: StyleImage}, patterns: {[string]: StyleImage}) {
-        const iconPositions = {}, patternPositions = {}, haveRenderCallbacks = [];
+        const iconPositions = {}, patternPositions = {};
+        this.haveRenderCallbacks = [];
 
         const bins = [];
-        for (const id in icons) {
-            const src = icons[id];
-            const bin = {
-                x: 0,
-                y: 0,
-                w: src.data.width + 2 * padding,
-                h: src.data.height + 2 * padding,
-            };
-            bins.push(bin);
-            iconPositions[id] = new ImagePosition(bin, src);
 
-            if (src.hasRenderCallback) {
-                haveRenderCallbacks.push(id);
-            }
-        }
-
-        for (const id in patterns) {
-            const src = patterns[id];
-            const bin = {
-                x: 0,
-                y: 0,
-                w: src.data.width + 2 * padding,
-                h: src.data.height + 2 * padding,
-            };
-            bins.push(bin);
-            patternPositions[id] = new ImagePosition(bin, src);
-
-            if (src.hasRenderCallback) {
-                haveRenderCallbacks.push(id);
-            }
-        }
+        this.addImages(icons, iconPositions, bins);
+        this.addImages(patterns, patternPositions, bins);
 
         const {w, h} = potpack(bins);
         const image = new RGBAImage({width: w || 1, height: h || 1});
@@ -126,7 +99,24 @@ export default class ImageAtlas {
         this.image = image;
         this.iconPositions = iconPositions;
         this.patternPositions = patternPositions;
-        this.haveRenderCallbacks = haveRenderCallbacks;
+    }
+
+    addImages(images: {[string]: StyleImage}, positions: {[string]: ImagePosition}, bins: Array<Rect>) {
+        for (const id in images) {
+            const src = images[id];
+            const bin = {
+                x: 0,
+                y: 0,
+                w: src.data.width + 2 * padding,
+                h: src.data.height + 2 * padding,
+            };
+            bins.push(bin);
+            positions[id] = new ImagePosition(bin, src);
+
+            if (src.hasRenderCallback) {
+                this.haveRenderCallbacks.push(id);
+            }
+        }
     }
 
     patchUpdatedImages(imageManager: ImageManager, texture: Texture) {
--- a/src/render/image_atlas.js
+++ b/src/render/image_atlas.js
@@ -5,6 +5,8 @@ import { register } from '../util/web_worker_transfer';
 import potpack from 'potpack';
 
 import type {StyleImage} from '../style/style_image';
+import type ImageManager from './image_manager';
+import type Texture from './texture';
 
 const padding = 1;
 
@@ -19,10 +21,12 @@ type Rect = {
 export class ImagePosition {
     paddedRect: Rect;
     pixelRatio: number;
+    version: number;
 
-    constructor(paddedRect: Rect, {pixelRatio}: StyleImage) {
+    constructor(paddedRect: Rect, {pixelRatio, version}: StyleImage) {
         this.paddedRect = paddedRect;
         this.pixelRatio = pixelRatio;
+        this.version = version;
     }
 
     get tl(): [number, number] {
@@ -55,10 +59,11 @@ export default class ImageAtlas {
     image: RGBAImage;
     iconPositions: {[string]: ImagePosition};
     patternPositions: {[string]: ImagePosition};
+    haveRenderCallbacks: Array<string>;
     uploaded: ?boolean;
 
     constructor(icons: {[string]: StyleImage}, patterns: {[string]: StyleImage}) {
-        const iconPositions = {}, patternPositions = {};
+        const iconPositions = {}, patternPositions = {}, haveRenderCallbacks = [];
 
         const bins = [];
         for (const id in icons) {
@@ -71,6 +76,10 @@ export default class ImageAtlas {
             };
             bins.push(bin);
             iconPositions[id] = new ImagePosition(bin, src);
+
+            if (src.hasRenderCallback) {
+                haveRenderCallbacks.push(id);
+            }
         }
 
         for (const id in patterns) {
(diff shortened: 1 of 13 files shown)


src/render/image_atlas.js
```diff
@@ -113,7 +126,27 @@ export default class ImageAtlas {
...
     }
+
+    patchUpdatedImage(position: ?ImagePosition, image: ?StyleImage, texture: Texture) {
+        if (!position || !image) return;
+
+        if (position.version === image.version) return;
+
+        position.version = image.version;
+        const tl = position.tl;
+        texture.update(image.data, undefined, { x: tl[0], y: tl[1] });
```
mourner: ES6 nit:

```js
const [x, y] = position.tl;
texture.update(image.data, undefined, {x, y});
```

src/render/image_atlas.js
```diff
@@ -83,6 +92,10 @@ export default class ImageAtlas {
             };
             bins.push(bin);
             patternPositions[id] = new ImagePosition(bin, src);
+
+            if (src.hasRenderCallback) {
+                haveRenderCallbacks.push(id);
+            }
```
mourner: These two blocks (icons and patterns loop) start to become too repetitive. Let's cut this out into a helper function to DRY it up.

src/render/image_manager.js
```diff
@@ -25,7 +26,7 @@ type Pattern = {
 const padding = 1;
 
 /*
-    ImageManager does two things:
+    ImageManager does three things:
 
         1. Tracks requests for icon images from tile workers and sends responses when the requests are fulfilled.
         2. Builds a texture atlas for pattern images.
```
asheemmamoowala: Could you add the third thing to this comment for completeness

src/style/style_image.js
```diff
@@ -1,9 +1,126 @@
...
+ *  }
+ *
+ *  map.addImage('flashing_square', flashingSquare);
+ */
+
+/**
+ * This method is called once before every frame where the icon will be used.
+ * The method can optionally update the image's `data` member with a new image.
+ *
+ * If the method updates the image it must return `true` to commit the change.
```
asheemmamoowala: It would probably be helpful to clarify that this method has to return synchronously, and provide clarity on how to deal with the situation where the `render()` call needs to load resources asynchronously., i.e. - can it just call `updateImage` when ready

src/render/texture.js
```diff
@@ -72,10 +72,11 @@ class Texture {
             }
 
         } else {
+            const {x, y} = position || { x: 0, y: 0};
             if (image instanceof HTMLImageElement || image instanceof HTMLCanvasElement || image instanceof HTMLVideoElement || image instanceof ImageData) {
-                gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, image);
+                gl.texSubImage2D(gl.TEXTURE_2D, 0, x, y, gl.RGBA, gl.UNSIGNED_BYTE, image);
             } else {
-                gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, image.data);
+                gl.texSubImage2D(gl.TEXTURE_2D, 0, x, y, width, height, gl.RGBA, gl.UNSIGNED_BYTE, image.data);
```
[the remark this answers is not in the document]
ansis: This is good to know!

I'm going to merge this as is but we should keep an eye on the performance this. Profiling this with a low number of images seemed to work ok. It's likely that for larger numbers of images your approach could work better. I initially implemented this by copying the images into the image and then calling `texSubImage2D` once per texture. The cost of the manual copies was measurable. It's very possible that by switching to the current approach I just made the cost more hidden.

Hopefully if this API turns out to be useful we can optimize based on users' use cases.

One of my motivations for patching images individually is that it lets you supply dom elements as the input. We aren't using this right now but I'm hoping to extend image support to include `HTMLVideoElement`. In a prototype, updating a image by going the `canvas drawImage --> canvas getImageData --> texture update` was much slower and this approach avoids that. Implementing both approaches shouldn't add more complexity.

Issue #6823, opened 2018-06-15
symbol layer not displayed on raster TMS layer


mikelmaron:
**mapbox-gl-js version**:

0.45.0

**browser**:

Google Chrome | 66.0.3359.181 (Official Build) (64-bit)

### Steps to Trigger Behavior

 1. Toggle between `symbol-osm` (style based on OSM raster TMS) and `symbol-streets` (mapbox streets style)

### Link to Demonstration

https://mapkibera.github.io/counties/test.html

### Expected Behavior

symbol layer visible whether OSM TMS style or mapbox streets style

### Actual Behavior

symbol layer only visible with mapbox streets style

jfirebaugh:
The `symbol-osm` style needs a `sprite` property. I believe you can reuse the value from Mapbox Streets, `"sprite": "mapbox://sprites/mapbox/streets-v9"`.

Leaving this open because although we have a validation check for missing `glyphs` property when `text-field` is used, it looks like we're missing the equivalent for `sprite`/`icon-image`. (Though we have to be careful about making validations stricter to avoid breaking existing styles.)

PR #6831, opened 2018-06-18
Warn user if required sprite property is not present or non-existent image is requested from sprite

ryanhamley:
## Launch Checklist

 - [x] briefly describe the changes in this PR
    - logs a console warning if sprite URL is not defined and a property that requires sprites is used
    - logs a console warning if an image not in the sprite is requested
 - [ ] write tests for all new functionality
    - i don't see tests for the existing glyph validation. do we think unit tests are necessary for this? if so, i can add them for both sprite and glyph validation
 - [x] manually test the debug page
screen shot 2018-11-01 at 5 42 27 pm
screen shot 2018-11-01 at 5 48 21 pm

Closes #6823

PR #8000, opened 2019-03-06
warn about missing style images

ansis:
Adds a missing fix to the previous icon pr. Cherry picked from https://github.com/mapbox/mapbox-gl-js/pull/6831/

### DIFF
--- a/src/render/image_manager.js
+++ b/src/render/image_manager.js
@@ -8,6 +8,7 @@ import { ImagePosition } from './image_atlas';
 import Texture from './texture';
 import assert from 'assert';
 import {renderStyleImage} from '../style/style_image';
+import { warnOnce } from '../util/util';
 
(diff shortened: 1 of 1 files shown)


ansis:
Yep, changed to mention `styleimagemissing`

chloekraw:
@ansis @ryanhamley does this close https://github.com/mapbox/mapbox-gl-js/issues/6823?

ryanhamley:
@chloekraw #6823 is technically asking for something slightly different (warn when the sprite URL is not defined in a style that uses sprite-dependent properties vs warn when an image that is requested is not found) but this might cover the use case well enough for it to be considered closed. @ansis do you think we need specific sprite validation logic?

chloekraw:
👍 thanks @ryanhamley, to me it sounds like something that would still be good to have.

Issue #7648, opened 2018-11-30
Setting the same value with setLayoutProperty triggers render event


DannyDelott:
**mapbox-gl-js version**:
0.51.0

**browser**:
Chrome

### Background

My application subscribes to the map's `render` event for many reasons, eg: showing a spinner while the map renders tiles/layers.

I notice we get flickers of a spinner when calling `setLayoutProperty` on a layer and passing the same value that's already in the layer.

### Steps to Trigger Behavior

 1. Call `setLayoutProperty` on a known, visible layer
 2. Set the visibility property to `visible`
 3. Set the visibility property to `visible` again

### Link to Demonstration
Click the "Test" button in this jsbin:
https://jsbin.com/faqupawuho/1/edit?console,output

### Expected Behavior

Calling `map.setLayoutProperty('some-layer', 'visibility', 'visible')` when the layer is already visible should be a no-op.

### Actual Behavior
Two `render` events fire (see console logs in the JSBin example)

weiliang903:
same issue for me.
`this.map.setPaintProperty(this._rainfallId, 'raster-opacity', visible ? 0.6 : 0);`
 it's working

mourner:
This was likely fixed by #7459. I'll reopen if it's not the case.

Issue #7459, opened 2018-10-22
`map.setLayoutProperty('layerID', 'visibility', 'none')` will trigger duplicate  http request


yangdonglai:
**mapbox-gl-js version**: 0.49.0

### Question
simple hidden a raster layer `map.setLayoutProperty('layerID', 'visibility', 'none')` will duplicate  http request`
### step
1. add a raster layer with google map source.and clear network console
![image](https://user-images.githubusercontent.com/17786362/47284571-6feb8580-d61a-11e8-8471-254f4c0040c6.png)

2. hidden raster layer by  `map.setLayoutProperty('layerID', 'visibility', 'none')`,and you will find it trigger duplicate http 
![image](https://user-images.githubusercontent.com/17786362/47284697-e2f4fc00-d61a-11e8-8b96-ee94ab62888f.png)

yangdonglai:
### Is it a special feature or something need improved? 
 
I think `visibility` should not trigger source reload.  Could somebody identify which ones layout property should trigger  `sourceCache.reload` like paint property. 
![image](https://user-images.githubusercontent.com/17786362/47285053-2439db80-d61c-11e8-8efd-e93d98fe047e.png)

Niumeng1:
**I found that the branch of 0.20.0 version of the same code, the request is not made, the new code on request**

mourner:
This was likely fixed by #7459. I'll reopen if it's not the case.

igor-dv:
Hey, we have the same issue with the latest version (1.1.0), I see it was mentioned this was fixed by #7459, but #7459 is actually **this** issue...

mourner:
Oh, I meant #8005. Let's reopen and check again then.

PR #8005, opened 2019-03-07
Improve diffing of styles containing "visibility": "visible"

lucaswoj:
When I execute this code:

```js
const style = {
    version: 8,
    sources: {},
    layers: [{id: 'background', layout: {visibility: 'visible'}]
}
const map = new Map({style, ...});
map.setStyle(style)
```

I expect `setStyle` to do a diff-based update which reduces to a no-op. 

However what actually happens is that it produces a `['setLayoutProperty', 'visibility', null]` command because we store `visibility: 'visible'` the same way as `visibility: undefined` and always serialize into the latter. `map.getStyle().layers[0].layout.visibility` is `undefined`. 

It is expensive to run a large number of `['setLayoutProperty', 'visibility', null]` commands.

Other properties do not behave like this. Visibility is a special case. 

This PR correctly stores and serializes the state of `visibility` into either `visible` or `undefined`.

This may `fix` https://github.com/mapbox/mapbox-gl-js/issues/7459
This may `fix` https://github.com/mapbox/mapbox-gl-js/issues/7648

## Launch Checklist

### DIFF
--- a/src/style/style_layer.js
+++ b/src/style/style_layer.js
@@ -39,7 +39,7 @@ class StyleLayer extends Evented {
     minzoom: ?number;
     maxzoom: ?number;
     filter: FilterSpecification | void;
-    visibility: 'visible' | 'none';
+    visibility: 'visible' | 'none' | void;
     _crossfadeParameters: CrossfadeParameters;
 
     _unevaluatedLayout: Layout<any>;
@@ -69,7 +69,6 @@ class StyleLayer extends Evented {
 
         this.id = layer.id;
         this.type = layer.type;
-        this.visibility = 'visible';
         this._featureFilter = () => true;
 
         if (layer.type === 'custom') return;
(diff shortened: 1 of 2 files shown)


lucaswoj:
All callsites that use the visibility property check it as `.visbility !== 'none'` so this change should be okay.

lucaswoj:
... It's definitely not working. Investigating further.

src/style/style_layer.js
```diff
@@ -210,7 +209,7 @@ class StyleLayer extends Evented {
             'paint': this._transitionablePaint && this._transitionablePaint.serialize()
         };
 
-        if (this.visibility === 'none') {
+        if (this.visibility) {
```
asheemmamoowala: This will always set `output.layout.visibility` to `'none'` regardless of the value of `this.visibility`

lucaswoj: I think I’ve fixed this in a subsequent commit. Or I misunderstand. There’s a test that ensures it can be set to “visible”.

asheemmamoowala: You're right, I was viewing a single commit instead of all the changes.

lucaswoj:
Thanks @asheemmamoowala! I think you need to press the big green button 😄.

PR #8008, opened 2019-03-09
Enable local ideograph generation by default 

asheemmamoowala:
Enable local ideograph generation for glyphs in the 'CJK Unified Ideographs' and 'Hangul Syllables' ranges, using the system default sans-serif font.

I haven't been able to find a good way to add tests for this PR. The glyph rendering depends on a font being consistently available on the system. While a font can be installed on the docker containers for CI, it would also have to be made available for every developer running the test locally. 

@mapbox/studio @mapbox/maps-design This change affects GL-JS by default, so for designers in Studio working on styling with custom fonts - it might be useful think of some messaging if the map is bringing in data in those glyph ranges, or the map is hovering over data in that range. 

## Launch Checklist
 - [ ] post benchmark scores

cc @ChrisLoer 👋 @chloekraw @lloydsheng @chriswu42

### DIFF
--- a/test/suite_implementation.js
+++ b/test/suite_implementation.js
@@ -52,6 +52,7 @@ module.exports = function(style, options, _callback) { // eslint-disable-line im
         axonometric: options.axonometric || false,
         skew: options.skew || [0, 0],
         fadeDuration: options.fadeDuration || 0,
+        localIdeographFontFamily: options.localIdeographFontFamily || false,
         crossSourceCollisions: typeof options.crossSourceCollisions === "undefined" ? true : options.crossSourceCollisions
     });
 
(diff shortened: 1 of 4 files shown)


test/suite_implementation.js
```diff
@@ -156,7 +157,9 @@ module.exports = function(style, options, _callback) { // eslint-disable-line im
             map.addLayer(new customLayerImplementations[operation[1]](), operation[2]);
             map._render();
             applyOperations(map, operations.slice(1), callback);
-
+        } else if (operation[0] === 'setStyle') {
+            map.setStyle(operation[1], { localIdeographFontFamily: false });
+            applyOperations(map, operations.slice(1), callback);
```
mourner: Nit: let's add a comment here about why this is necessary.

Issue #7701, opened 2018-12-13
CSS transform issue 


jingsam:
When emit click, zoom, touch event on map, we calculate the mouse position in map as follows:

https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/dom.js#L104-L111

```
DOM.mousePos = function (el: HTMLElement, e: any) {
    const rect = el.getBoundingClientRect();
    e = e.touches ? e.touches[0] : e;
    return new Point(
        e.clientX - rect.left - el.clientLeft,
        e.clientY - rect.top - el.clientTop
    );
};
```

This process is simple like this:

![image](https://user-images.githubusercontent.com/1522494/49931810-a6e76480-ff02-11e8-9006-045994735416.png)

The calculation is correct on most cases, except the case that we add css transform property on .mapboxgl-canvas-container.  For example, we add `rotate(10deg)`

![image](https://user-images.githubusercontent.com/1522494/49931985-fc237600-ff02-11e8-8fd2-8b8c544731a3.png)

If we scale or skew or even transform3d the canvas container, we also can not get the right position from `mousePos()`. The zoom, pan, touch, click behaviour will be affected:

https://jsfiddle.net/jingsam/3wh5grko/25/

Try to click somewhere on the map.

This [issue](https://github.com/mapbox/mapbox-gl-js/issues/6079) is also related to the same problem.

# Solution

What we want from `mousePos` is the position is **map space**, not the viewport space or other space. Luckily, we have [offsetX](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX) [offsetY](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX). [CSSOM](https://drafts.csswg.org/cssom-view/#dom-mouseevent-offsetx) says:

> The offsetX attribute must follow these steps:
> 1. If the event’s dispatch flag is set, return the x-coordinate of the position where the event occurred **relative to the origin of the padding edge of the target node**, ignoring the transforms that apply to the element and its ancestors, and terminate these steps.
> 2. Return the value of the event’s pageX attribute.

offsetX and offsetY are just we want. so `mousePos` should changed as follows:

```
DOM.mousePos = function (el: HTMLElement, e: any) {
    e = e.touches ? e.touches[0] : e;
    return new Point(
        e.offsetX,
        e.offsetY
    );
};
```

jingsam:
@mourner @ansis any idea on this issue?

[...]

Issue #6079, opened 2018-01-29
Mouse events are shifted after a css transform scale on the map container


fxi:
**mapbox-gl-js version**: 0.44.0

### Steps to Trigger Behavior

 1. Create a map,
 2. Set `transform scale(0.9)` on map container,
 3. Use `map.resize()` to update the canvas.

Example : https://jsfiddle.net/fxi/gb5jqssu/

### Expected Behavior

All mouse events should match the updated map.

### Actual Behavior

All mouse events are slightly shifted.

azei:
I'm running in to this issue as well. Scaling both shifts zooming in/out as well making the map hard to pan.

Has anyone found a workaround?

adamcbrewer:
Same here, also struggling with this and the only way around this currently is to write custom interactions for all scroll and touch events.

Perhaps a scale-offset property could be added to the API?

adamcbrewer:
Just a note, i've posted a possible solution to this issue in https://github.com/mapbox/mapbox-gl-js/issues/7701#issuecomment-465951671.

adamcbrewer:
So I've just submitted a [proposed PR](https://github.com/mapbox/mapbox-gl-js/pull/8017) ☝️  for this issue. The functionality works, but once I've had feedback I'll continue with the implementation and writing tests. Any eyes-on would be appreciated!

adamcbrewer:
@rgazelot unfortunately not as the conversation in #8017 means not using css transforms and rather going down the line of using `offsetX` and `offsetY`.  However I don't believe this is a solid solution until there's a way to prevent panning/dragging when the mouse/cursor leaves the map bounds.

rgazelot:
Thanks for the update @adamcbrewer

anvaka:
I have experimented with `offsetX` and `offsetY` based solution. While it does work and correctly returns local coordinates of an element inside CSS-transformed tree, it doesn't work for touch devices.

One promising direction to explore would be to use pointer events, like [pointerdown](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/pointerdown_event). MDN reports this as not supported by Safari, however this simple test https://output.jsbin.com/garuceh worked in both iOS simulator's Safari and desktop Safari. It also worked in Firefox, Chrome for Android and desktop.

I understand that changing input model to [pointers events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) could be big switch, but hopefully the right one in the long term.

[...]

PR #8017, opened 2019-03-11
Fixes shifted mouse events when CSS transform-scale applied

adamcbrewer:
This fixes [6079](https://github.com/mapbox/mapbox-gl-js/issues/6079) and part of [7701](https://github.com/mapbox/mapbox-gl-js/issues/7701) where mouse and touch events are incorrect when CSS transforms are applied to the map or any parent element wrapping it.

**NOTE: The PR is a draft until someone can confirm this could be a sensible fix.**

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/events.js
+++ b/src/ui/events.js
@@ -4,7 +4,7 @@ import { Event } from '../util/evented';
 
 import DOM from '../util/dom';
 import Point from '@mapbox/point-geometry';
-import { extend } from '../util/util';
+import { extend, cssTransformPoint } from '../util/util';
 
 import type Map from './map';
 import type LngLat from '../geo/lng_lat';
@@ -77,8 +77,9 @@ export class MapMouseEvent extends Event {
      */
     constructor(type: string, map: Map, originalEvent: MouseEvent, data: Object = {}) {
         const point = DOM.mousePos(map.getCanvasContainer(), originalEvent);
-        const lngLat = map.unproject(point);
-        super(type, extend({ point, lngLat, originalEvent }, data));
+        const transformedPoint = cssTransformPoint(point, map._cssTransforms);
+        const lngLat = map.unproject(transformedPoint);
+        super(type, extend({ transformedPoint, lngLat, originalEvent }, data));
         this._defaultPrevented = false;
         this.target = map;
     }
@@ -156,8 +157,9 @@ export class MapTouchEvent extends Event {
      */
     constructor(type: string, map: Map, originalEvent: TouchEvent) {
         const points = DOM.touchPos(map.getCanvasContainer(), originalEvent);
-        const lngLats = points.map((t) => map.unproject(t));
-        const point = points.reduce((prev, curr, i, arr) => {
+        const transformedPoints = points.map(p => cssTransformPoint(p, map._cssTransforms));
+        const lngLats = transformedPoints.map((t) => map.unproject(t));
+        const point = transformedPoints.reduce((prev, curr, i, arr) => {
             return prev.add(curr.div(arr.length));
         }, new Point(0, 0));
         const lngLat = map.unproject(point);
--- a/src/ui/handler/drag_pan.js
+++ b/src/ui/handler/drag_pan.js
@@ -1,7 +1,7 @@
 // @flow
 
 import DOM from '../../util/dom';
-import { bezier, bindAll } from '../../util/util';
+import { bezier, bindAll, cssTransformPoint } from '../../util/util';
 import window from '../../util/window';
 import browser from '../../util/browser';
 import { Event } from '../../util/evented';
@@ -9,7 +9,7 @@ import assert from 'assert';
 
 import type Map from '../map';
 import type Point from '@mapbox/point-geometry';
-import type {TaskID} from '../../util/task_queue';
+import type { TaskID } from '../../util/task_queue';
 
 const inertiaLinearity = 0.3,
     inertiaEasing = bezier(0, 0, inertiaLinearity, 1),
@@ -93,20 +93,20 @@ class DragPanHandler {
         if (!this.isEnabled()) return;
         this._el.classList.remove('mapboxgl-touch-drag-pan');
         switch (this._state) {
-        case 'active':
-            this._state = 'disabled';
-            this._unbind();
-            this._deactivate();
-            this._fireEvent('dragend');
-            this._fireEvent('moveend');
-            break;
-        case 'pending':
-            this._state = 'disabled';
-            this._unbind();
-            break;
-        default:
-            this._state = 'disabled';
-            break;
+            case 'active':
+                this._state = 'disabled';
+                this._unbind();
+                this._deactivate();
+                this._fireEvent('dragend');
+                this._fireEvent('moveend');
+                break;
+            case 'pending':
+                this._state = 'disabled';
+                this._unbind();
+                break;
+            default:
+                this._state = 'disabled';
+                break;
         }
     }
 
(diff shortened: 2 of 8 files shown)


adamcbrewer:
For some reason simulated mouse events are failing in `test/unit/ui/marker.test.js`, but I'm not sure what could be causing the issue. Is this an existing issue or a dependancy one? Pretty sure it's failing on [`simulate.mousedown(el);`](https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/ui/marker.test.js#L294) before any equality checks are even performed.

adamcbrewer:
I've updated to v1.0.0, but I still need someone to review these changes before looking into fixing the unit tests

adamcbrewer:
@mourner not sure who to ask, but would be great if someone from the team could review this proposal. I might need some help with the tests, but I don't want to spent time on fixing them if this isn't something that's likely to make it in.

ryanhamley:
Hi @adamcbrewer thanks for the PR and for being so patient. I think that this approach probably is not the right way to solve this issue. The `cssTransformPoint` function seems like it would put us on the hook to implement and support all 21 current CSS transform functions (plus any new ones that are added in the future). In effect, GL JS would have to maintain an implementation of CSS `transform`. Limiting it to certain transforms seems like a half-measure that leads to a bad user experience. Additionally, this approach requires the developer to implement the transform in two places, the CSS and the map constructor.

I think that @jingsam was on the right track with their example in https://github.com/mapbox/mapbox-gl-js/issues/7701#issue-390615747 even if that particular implementation has some edge cases such as the ones you mentioned in that issue. The map shouldn't need to know what, if any, transforms have been applied to it. Using `offsetX` and `offsetY` seems like the most promising approach; we'll just have to finesse how we handle panning and the mouse entering other elements. I'm not sure yet what the best way to achieve that is, but I think this is where I'd personally focus my efforts.

@peterqliu any thoughts on the best way forward on this issue?

peterqliu:
🤔 to take a step back, using transforms on the map verges on an antipattern to me, and is at best an edge case to support. Scaling is best done by resetting dimensions directly.

That said, I think this feature is limited in its upside, and in doing so introduces leaky abstractions (support for only certain transforms, on elements no higher in the DOM hierarchy) with the potential to add confusion and break expectations of intended behavior. 

I second @ryanhamley  that the `offsetX` route is most promising, if we can iron out mouseexit weirdness without adding significant complexity.

jingsam:
@peterqliu transform map does have its use cases. For example, If we create a dashboard that have maps, when presenting it on super big screen (I mean 4K X 2, or 1K X 6), it would cause two major problem:
1. poor performance: too many request to fulfill the map and memory consumption is huge.
2.too small symbol: the distance from a viewer to a big screen is far from desktop PC, so the text and icon is difficult to distinguish.

On this case, we scale the map using transform, perfectly solve the problem.

`offsetX` and `offsetY` are indeed a promising solution. However, `touch` events do not support `offsetX` and `offsetY`, so we should find another way on the mobile.

adamcbrewer:
Thanks for the reviews @peterqliu and @ryanhamley. I actually agree that it's an antipattern and also trying to maintain a set of css transforms is something to avoid.

I really liked the `offsetX` and `offsetY` solution by @jingsam as well, but as he mentioned there are additional considerations (such as mobile) and also the ones I mentioned in [this comment for 7701](https://github.com/mapbox/mapbox-gl-js/issues/7701#issuecomment-464021353).

I agree that handling transforms isn't the responsibility of GL JS, but in the comment above the `offsetX` and `offsetY` solution only works well if the mouse is still over the scaled node containing the css scaled/transformed map. If someone pans a map and drags the mouse outside of the scaled area it's almost completely unusable and you might get a lot of people creating issues for it 😄 

Perhaps there's a way to kill all the events if the map has been scaled and the mouse/pointer has left the map area? Maybe supporting we find some way of supporting `offsetX` and `offsetY`, but create an online example for handling css transform cases?

peterqliu:
@jingsam tangent regarding big screens: you make a great point that big screens tend to also be farther away, where we can see less detail. This hints at the actual best practice: *reducing screen resolution* as they get bigger, as users can't see as finely anyway. We see this pattern in display resolutions between phones, tablets, and laptops.

Handling this in the display settings is better than in CSS, because we can build the map once for all screens, without adding conditional logic to apply different transform scale for different sizes.

jingsam:
@peterqliu I agree with you that adjusting screen resolution is a feasible way. However, sometimes we share this big screen with other applications. Adjusting its resolution maybe break other application appearance. Also, our dashboard need to deploy to different big screen, like 4k, 6k, 8k. Adjusting resolution maybe a little trivial. In our practice, we calculate the ratio between screen and dashboard size and scale the dashboard to autofit to the big screen. That's why we depend on transform feature a lot. In out fork of mapbox-gl, we use `offsetX, offsetY` in mouse events and fallback to the original way in touch events.

ryanhamley:
>you make a great point that big screens tend to also be farther away, where we can see less detail. This hints at the actual best practice: reducing screen resolution as they get bigger, as users can't see as finely anyway. We see this pattern in display resolutions between phones, tablets, and laptops.

@peterqliu @jingsam Given this paradigm, is it even a big deal if `offsetX` and `offsetY` don't work on mobile devices? The assumption would be that transforms should only really be necessary on large screens which by definition are not mobile devices. This doesn't solve the issue of mouse events outside the transformed element but it would simplify the implementation a bit by avoiding the concerns with mobile. Or am I missing a mobile use case that would require a transform of some kind?

ryanhamley:
Also, would being able to set the `devicePixelRatio` help with larger screen resolutions instead of scaling the map? https://github.com/mapbox/mapbox-gl-js/pull/7333

jingsam:
@ryanhamley I think we could use `offsetX, offsetY` in mouse events. It has better performance as it is native supported. Also, it compatible  to all kinds of transform. When mouse moves outside of the transformed map, I assume that the mouse is no longer interact with the map. 

@ryanhamley thanks for your advice. Using `devicePixelRatio` maybe a tricky way to go around this problem. I guess that will increase the burden of GPU. I will test against it.

adamcbrewer:
@jingsam just to comment on what you said: 

> When mouse moves outside of the transformed map, I assume that the mouse is no longer interact with the map.

In fact Mapbox _does_ allow panning the map outside the map bounds as long as the mouse/pointer is still held down. This is where I've noticed the issues with `offsetX` and `offsetY` where once you're moving over elements with different transforms the behaviour becomes very erratic to say the least.

So really if we use `offsetX` and `offsetY` we need to ask 

1. should we allow panning outside of the map area (if the user is still interacting) and 
2. if so we should have at least show an example of this case, or at least how to kill the events once pointer has 'left' the map area

adamcbrewer:
I think this PR should be closed and a new one created if `offsetX` and `offsetY` are going to be the way forward.

ryanhamley:
agreed @adamcbrewer are you interested in working on a new PR for this? i'm not sure when someone on the GL JS team would get to it

adamcbrewer:
@ryanhamley I'd be happy to, but it would be good to get some thoughts on my previous comment: https://github.com/mapbox/mapbox-gl-js/pull/8017#issuecomment-519407188

> In fact Mapbox does allow panning the map outside the map bounds as long as the mouse/pointer is still held down. This is where I've noticed the issues with offsetX and offsetY where once you're moving over elements with different transforms the behaviour becomes very erratic to say the least.

> So really if we use offsetX and offsetY we need to ask:

> 

> 1. should we allow panning outside of the map area (if the user is still interacting) and
> 2. if so we should have at least show an example of this case, or at least how to kill the events once pointer has 'left' the map area"

PR #8018, opened 2019-03-11
Avoid reapplying marker anchor class on every frame

mourner:
`applyAnchorClass` was taking noticeable time when profiling a map with a lot of HTML markers. The PR moves this code to happen once when creating a marker, which should be fine because marker anchors are not mutable.

## Launch Checklist


jonnenauha:
When can this be expect in a release?

mourner:
@jonnenauha we're cutting a beta release today, with the expected final release next week.

PR #8019, opened 2019-03-11
Add HTML clusters + aggregated properties example

### LINKED ISSUE #4491: Support clustering for HTML "Marker"s
Hello!
As we can see here https://www.mapbox.com/mapbox-gl-js/example/custom-marker-icons/
we could create a custom Marker and add it to map.
e.g. new mapboxgl.Marker(...).setLngLat(...).addTo(map);

I'm trying to find the way how to add marker to the layer and organize into the Cluster. 
Smth like https://www.mapbox.com/mapbox-gl-js/example/cluster/ but with custom Markers:
1. var map = new mapboxgl.Map...
2. map.addSource(...);
3. map.addLayer(...);
4. and here map.getLayer(...).addMarker(Marker). or marker.addto(Layer/Source) or ...
As a result we should see a Cluster with custom markers.

mourner:
Closes #4491. An official example for using custom HTML (or SVG) elements for clusters, along with the new `customProperties` API. It's an advanced example with quite a bit going on, but I think it's still worth adding. 

Most of the complexity is in the donut chart generation logic, which is not important for the key features covered, but makes for a pretty and useful example.

image


docs/pages/example/cluster-html.js
```diff
@@ -0,0 +1,10 @@
+/*---
+title: Display HTML clusters with custom properties
+description: An advanced example of using Mapbox GL JS clustering with HTML markers and custom property expressions.
```
asheemmamoowala: Outlining the general strategy that the example takes would be very useful here to set context for the code.

mourner: Can you elaborate? Even the general strategy would need description that seems too long for the purpose.

asheemmamoowala: Maybe it can be added in the code section then?

Strategy would Outline that this example is using clustered GeoJSON with filtered layers and continuously updating Markers.

mourner: Right — I've extended the description.

mourner:
Disregard the failed build — flow randomly acting up.

docs/pages/example/cluster-html.js
```diff
@@ -0,0 +1,10 @@
+/*---
+title: Display HTML clusters with custom properties
+description: An advanced example of using Mapbox GL JS clustering with HTML markers and custom property expressions. To be able to use HTML or SVG for clusters in place of a Mapbox GL layer, we have to manually synchronize the clustered source with a pool of marker objects that we update continuously while the map view changes.
```
colleenmcginnis: A few suggestions that (hopefully) simplify the language without making anything untrue: 

>This advanced example uses Mapbox GL JS clustering with HTML markers and custom property expressions. To use HTML or SVG for clusters in place of a Mapbox GL layer, you have to manually synchronize the clustered source with a pool of marker objects that updates continuously while the map view changes.

mourner: @colleenmcginnis thanks, updated!

PR #8022, opened 2019-03-12
Example for `fill-pattern`

### LINKED ISSUE #8004: Example for using `fill-pattern`
The docs for the iOS SDK contains [an example](https://docs.mapbox.com/ios/maps/examples/fill-pattern/) of using the `fill-pattern` property along with an image of our choice. However the Mapbox GL JS doc doesn't, so I thought it would be great to add one, especially given that it's pretty simple to do once you know how, but it's not the same when there's no example and you have to figure out by yourself by reading the API documentation.

I could even submit a PR to get you started if you want.

flawyte:
This is a new example that shows how to use the `fill-pattern` paint property.

I've used the same title as the [iOS example](https://docs.mapbox.com/ios/maps/examples/fill-pattern/).

Happy to be contributing 🎉

Fixes #8004


ryanhamley:
I forgot that our new doc styles require a 1200x500 screenshot of the example to be saved into `/docs/img/src`. Our docs team is working on a fix so that a missing image doesn't break the docs, but if you want to fix the failing unit tests and make this work, you can add that screenshot. You can read more details [here](https://github.com/mapbox/mapbox-gl-js/tree/publisher-production/docs#writing-examples).

flawyte:
I tried to use the `create-image` command after merging `publisher-production` but the result image was always blank, so I took a screenshot.

I've also updated the code indentation to match yours.

Should be ok now.

docs/pages/example/fill-pattern.js
```diff
@@ -0,0 +1,10 @@
+/*---
+title: Add a pattern to a polygon
+description: Load an image and use it as a repeating pattern to draw a polygon.
```
ryanhamley: The description should include the term `fill-pattern` so that if the property is searched in the examples, this example will come up. Something like `Use [fill-pattern](https://docs.mapbox.com/mapbox-gl-js/style-spec/#paint-fill-fill-pattern) to draw a polygon from a repeating image pattern` would be good and I like linking to the property documentation.

docs/pages/example/fill-pattern.html
```diff
@@ -0,0 +1,49 @@
...
+                    [30, -25],
+                    [-30, -25]
+                ]]
+            }
+        }
+    });
+
+    // Load the image to use as pattern
+    map.loadImage('https://upload.wikimedia.org/wikipedia/commons/thumb/6/60/Cat_silhouette.svg/64px-Cat_silhouette.svg.png', function(err, image) {
+        // Throw if something wrong happened
```
ryanhamley: Nit on grammar: `Throw an error if something went wrong`

docs/pages/example/fill-pattern.html
```diff
@@ -0,0 +1,49 @@
...
+                    [-30, 35],
+                    [30, 35],
+                    [30, -25],
+                    [-30, -25]
+                ]]
+            }
+        }
+    });
+
+    // Load the image to use as pattern
```
ryanhamley: Nit: `Load an image to use as the pattern`

PR #8023, opened 2019-03-12
Gray out zoom buttons at maxZoom/minZoom

andycalder:
This is a simple UI enhancement. When the map is at `maxZoom`, the `NavigationControl` looks like this:
zoomingrayed
And likewise at `minZoom`:
zoomoutgrayed

This is standard behaviour for zoom controls in many applications. I was going to implement my own buttons because I wanted this functionality, but then I thought "why not just add it to the library?"

## Launch Checklist

 - [ ] write tests for all new functionality (`NavigationControl` has no unit tests. Should it?)
 - [x] fix flickering issue on Safari/Firefox

On Safari and Firefox there is an almost imperceptible flicker when the `background-image` changes. The geolocate icon has the same problem. Chrome seems to render perfectly. I haven't tested IE/Edge. Does anyone have any advice on the best way to fix this?


ryanhamley:
A [CSS blend mode](https://css-tricks.com/basics-css-blend-modes/) could be helpful

andycalder:
Thanks for the suggestions. I managed to get it working using opacity, which fixes the flickering issue and avoids duplicating SVGs.

PR #8025, opened 2019-03-12
Set default thumbnail for examples page for development

katydecorah:
## Launch Checklist

When adding a new example, if you didn't set a thumbnail then the examples page would not load due to the failed dependency. This PR sets a default thumbnail image for examples so that the contributor is able to load the examples page locally.


ryanhamley:
This works well but do you think it might be better to use a true placeholder image kind of like this?

Screen Shot 2019-03-12 at 12 53 31 PM

My thinking is that it's easier to quickly find the broken image this way when scanning the examples page. Maybe that's not a huge deal but it could help catch the problem.

katydecorah:
@ryanhamley that makes sense. I just pushed a placeholder image with a solid background.

At your suggestion, I started playing around with Puppeteer and I think I'll be able to create a script to automatically create the image at the correct dimensions and naming format. I'll create a PR once I have the script cleaned up.

PR #8029, opened 2019-03-12
Add create-image script to generate example thumbnails

katydecorah:
refs https://github.com/mapbox/mapbox-gl-js/pull/8025

To make it easier for contributors who are adding new examples, I created `npm run create-image` that will take a screenshot of the new example's map and save it to the correct directory. The script also saves the image using our specifications of size and format.

For the script to run, the contributor will need to pass an Mapbox access token to the command so that the map can load for the screenshot. While this adds a small amount of friction, I think it still streamlines the process of generating these thumbnail images. I'd be happy to hear any suggestions on implementation.


katydecorah:
Thanks for the feedback @ryanhamley! I've incorporated your suggestions.

> The script doesn't recreate the current images as seen here:

That's a good point! The original screenshots were also retina, so I played around with puppeteer's viewport and was able to force a very similar scale. I think that should keep us consistent.

docs/bin/create-image.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+
+const puppeteer = require('puppeteer'); // eslint-disable-line
+const path = require('path'); // eslint-disable-line
+const pack = require('../../package.json'); // eslint-disable-line
+
+const fileName = process.argv[2];
+const token = process.argv[3] || process.env.MAPBOX_ACCESS_TOKEN;
```
ryanhamley: Ah sorry, I meant that it would be good to check for `MAPBOX_ACCESS_TOKEN` in addition to `MapboxAccessToken` since people may have it saved either way. That's what the benchmark tests do as seen [here](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/lib/access_token.js#L2-L3).

docs/README.md
```diff
@@ -35,12 +35,10 @@ In the `.html` file, write the HTML and JavaScript constituting the example.
 
 Every example **must** have an accompanying image:
 
-1. Save the image in `docs/img/src/`. Image requirements:
-  - The file name of the image **must** match the example's file name
-  - 1200 x 500 pixel
-  - png format
-2. Run `npm run build-images` to generate the appropriate images. Alternatively, when you run `npm run start-docs` the site will generate the images.
-3. Run the site locally to verify that your example image is loading as expected.
+1. Run `npm run create-image <example-file-name> <mapbox-access-token>`. The script will take a screenshot of map in the example and save it to `docs/img/src/`. Commit the image.
```
ryanhamley: Nit: should read `The script will take a screenshot of the map`

mourner:
Just a note that `puppeteer` is quite a heavy dev dependency (180 MB) — one more argument to cut out the docs system from the main repo.

PR #8035, opened 2019-03-14
Cache temporary link for url resolution

### LINKED ISSUE #8032: The number of DOM nodes rises when setData(URL) is called frequenty
**mapbox-gl-js version**: 0.53.1

**browser**: tested in chrome, but possibly in all browsers

### Steps to Trigger Behavior

 1. minimal code is following: create source and update the source using setData(URL) periodicaly

```
#HTML code
<div id="map"></div>
<button onclick="updateSource()">UPDATE</button>

#JS code
map.addSource("src", {
   "type" : "geojson",
   "data" : {
        "type": "Feature",
        "geometry": {
          type : "Point",
          coordinates : [14.5, 50.5]
        }
    }
});

function createJsonUrl (json = '') {
    return window.URL.createObjectURL(new Blob([json], {type: 'application/json'}));
};

function setSourceData (sourceId, data) {
   const src = map.getSource(sourceId);
   if (src) src.setData(data);
};

function updateSource () {
  const url = createJsonUrl(
    JSON.stringify({
      "type": "Feature",
      "geometry": {
        type : "Point",
        coordinates : [14.5 + (Math.random()-0.5)/4, 50.5 + (Math.random()-0.5)/4]
      }
    })
  );
  setSourceData('src', url);
};
```
2. open in chrome dev tools - performace tab
3. start recording
4. click several times to UPDATE button
5. stop recording 
6. check number of DOM nodes (green line in performance report)
 
### Expected Behavior
The number of DOM nodes should be constant. 

### Actual Behavior
Number of DOM nodes (green line in performance report ) constantly rises with each source update. In case of fast animation using requestAnimationFrame, after some time DOM garbage collector is called, caused noticeably longer animation frame ("freezing").
The cause is the method browser.resolveURL(), which creates new A element out of the document tree only for the purpose of the url normalization. 
The easiest solution would be to make url resolving optional (I guess it's not needed especially in case of object URLs). Or rewrite the method resolveURL() to reuse the anchor element node, if possible....?

Anyway, thank you for the great piece of software!

mourner:
Closes #8032. Avoids unnecessary GC freezes when collecting unattached DOM objects by caching the link element for URL resolution. @poletani can you check if this fixes your issue?

## Launch Checklist

 - [ ] ~~write tests for all new functionality~~ we don't test browser-dependent methods

### DIFF
--- a/src/util/browser.js
+++ b/src/util/browser.js
@@ -17,6 +17,8 @@ const cancel = window.cancelAnimationFrame ||
     window.webkitCancelAnimationFrame ||
     window.msCancelAnimationFrame;
 
+let linkEl;
+
 /**
  * @private
  */
@@ -45,9 +47,9 @@ const exported = {
     },
 
     resolveURL(path: string) {
-        const a = window.document.createElement('a');
-        a.href = path;
-        return a.href;
+        if (!linkEl) linkEl = window.document.createElement('a');
+        linkEl.href = path;
+        return linkEl.href;
     },
 
     hardwareConcurrency: window.navigator.hardwareConcurrency || 4,


ryanhamley:
I _think_ the underlying issue is that the anchor's `href` property was saved to an `options` object in `geojson_source._updateWorkerData` then a different property in the `options` object was used in dispatching the update to a worker. Since we reuse workers, it seems like the options object was never being garbage collected and in turn the `<a>` tags weren't being garbage collected. Either way, this should fix the issue.

mourner:
@ryanhamley `a.href` is a string (a primitive value), so it shouldn't have any connection to the anchor. I think I misread the original report — the DOM object was indeed garbage collected, but not immediately because browsers prefer to do that in batches, and apparently if there were a ton of these objects to collect during a garbage sweep, it could cause micro-stutters.

poletani:
@mourner unfortunately now I am unable to check it (I'll try later just to be sure), but it appears to me like the ideal solution. Thanks a lot for fixing it!

poletani:
@mourner I can confirm that the number of DOM nodes remains constant under the same test method, using the new fix.

pakastin:
Perfect, this was a huge issue for me! Great work! 👌

PR #8044, opened 2019-03-15
Introduce static worker version of the bundle for strict CSP environments

### LINKED ISSUE #6058: Start web worker from static URL on startup
# Issue

Currently, Mapbox GL JS dynamically generates a worker on startup and hosts it in a[ local window URL](https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/browser/web_worker.js#L5).  This startup method causes issues with:

1. Applications that require a `worker:src` CSP
2. Mapbox GL startup in iframe sandbox permissions `allow-same-origin` on IE11, Safari, and Edge browsers.

# Possible implementation

1. Add another build step that browserifies https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/worker.js
2. Modify https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/browser/web_worker.js so that instead of using WebWorkify, it takes a real URL (by which that browserified worker build will be accessible)

# Ideal solution

Ideally, we'd be able to detect CSP on the client and load the same JS file again, with a different entry point.  The entry point can be a static worker URL if CSP `worker:src` is enforced.

cc @kkaefer 

Related: https://github.com/mapbox/mapbox-gl-js/issues/6056

mourner:
Closes #6058. Adds a `build-csp` script that builds a static worker version of the GL JS bundle — `dist/mapbox-gl.csp.js` and `dist/mapbox-gl.worker.js`, which would be included in the distribution of the next release. The only additional step to using this bundle is setting `mapboxgl.workerUrl` manually before instantiating the map:

```html
<script src='/dist/mapbox-gl.csp.js'></script>
<script>
    mapboxgl.workerUrl = '/dist/mapbox-gl.worker.js';
    var map = new mapboxgl.Map(...);
```

You can test this out locally with the `debug/csp-static.html` page.

## Bundle size comparison

bundle | minified | gzipped
--- | --- | ---
mapbox-gl.js | 656.74KB | 166.76KB
mapbox-gl.csp.js | 597.66KB | 147.30KB
mapbox-gl.worker.js | 324.46KB | 85.42KB

## Launch Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,7 @@
 /rollup/build/
 /docs/components/api.json
 /dist/
-/docs/pages/dist/mapbox-gl-dev.js
-/docs/pages/dist/mapbox-gl.js
-/docs/pages/dist/mapbox-gl-unminified.js
+/docs/pages/dist/
 *.js.map
 node_modules
 package-lock.json
--- a/build/banner.js
+++ b/build/banner.js
@@ -1,0 +1,4 @@
+import fs from 'fs';
+
+const version = JSON.parse(fs.readFileSync('package.json')).version;
+export default `/* Mapbox GL JS is licensed under the 3-Clause BSD License. Full text of license: https://github.com/mapbox/mapbox-gl-js/blob/v${version}/LICENSE.txt */`;
--- a/build/rollup_plugins.js
+++ b/build/rollup_plugins.js
@@ -9,14 +9,10 @@ import { terser } from 'rollup-plugin-terser';
 import minifyStyleSpec from './rollup_plugin_minify_style_spec';
 import { createFilter } from 'rollup-pluginutils';
 
-const {BUILD, MINIFY} = process.env;
-const minified = MINIFY === 'true';
-const production = BUILD === 'production';
-
 // Common set of plugins/transformations shared across different rollup
 // builds (main mapboxgl bundle, style-spec package, benchmarks bundle)
 
-export const plugins = () => [
+export const plugins = (minified, production) => [
     flow(),
     minifyStyleSpec(),
     json(),
--- a/debug/csp-static.html
+++ b/debug/csp-static.html
@@ -21,7 +21,7 @@
 <script src='/debug/access_token_generated.js'></script>
 <script nonce="app-js">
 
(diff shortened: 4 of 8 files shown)


mourner:
Feel free to suggest different naming — maybe `mapbox-gl.static.js` would be better?

ansis:
This looks good to me.

Is the script loaded only when the first worker is initialized? Is this ok or would first load be faster if we prefetch it? This is something we don't need to address now.

> The only additional step to using this bundle is setting mapboxgl.workerUrl manually before instantiating the map

Is it possible to default this to the path the current script was loaded from? So that if a user puts both files in the same directory and doesn't rename them they don't need to specify this?

> Feel free to suggest different naming — maybe mapbox-gl.static.js would be better?

I think both files should share a name to show that they go together. For example,
`mapbox-gl.ASDF.js` and `mapbox-gl.ASDF.worker.js` (or `mapbox-gl.ASDF-worker.js`).

Other possible names are `csp-worker-src` to be more specific about what this solves. Or `no-worker-blob`. Or just `split`?

`mapbox-gl.split.js` and `mapbox-gl.split.worker.js`?

mourner:
> Is the script loaded only when the first worker is initialized? Is this ok or would first load be faster if we prefetch it? This is something we don't need to address now.

If the map isn't instantiated on load, this can be addressed on the app side with:

```html
<link rel="preload" href=".../mapbox-gl-worker.js" as="script">
```

> Is it possible to default this to the path the current script was loaded from? So that if a user puts both files in the same directory and doesn't rename them they don't need to specify this?

We could, but I think that with the current prevalence of bundlers and fragility of regexp script heuristics, it's not worth doing. If someone wants to switch to this special bundle, I think they won't have a problem adding a short line of code, and it will be reliable.

> I think both files should share a name to show that they go together.

Good point! I can rename to `mapbox-gl-csp.js` and `mapbox-gl-csp-worker.js` (using hyphens for consistency with other existing builds). I think we should keep the names simple (`mapbox-gl-no-worker-blob-worker.js` looks scary), but in that case at least `csp` gives a clue about what the bundle is about (`split` without context seems more cryptic).

ansis:
👍

thomasneirynck:
Thanks @mourner!

Issue #3709, opened 2016-11-28
Using dynamically-rendered vector tiles (from cartodb)


chriswhong:
Hello Mapbox Team!  

I am experimenting with the undocumented windshaft (cartodb) vector tile feature, and have successfully rendered carto vector tiles in mapboxGL.  Here's an example:  http://bl.ocks.org/chriswhong/2695b75fd1936bd034df83c91738648d

My question is about dynamically updating the vector tiles using the same workflow that `carto.js` uses to update raster tiles.  This probably sounds like I am _doing it wrong_ , but hear me out.

With carto.js, you can call `setSQL()` on a raster layer, which triggers leaflet to request new tiles from windshaft.  It removes the current raster tile layer and replaces it with new tiles rendered off the new SQL query.  I am wondering if it would be possible to update the `url` of a mapboxGL vector source in the same way, prompting mapboxGL to grab new tiles but leave the layer/style rules in place, essentially refreshing the data much like you can do with geoJSON layers.

I have a lot of bespoke-to-a dataset filtering UI that is essentially generating SQL filters.  One solution would be to only have carto serve the vector tiles once with a `SELECT *`, and then just have my filter UI filter the mapboxGL layers and never talk to Carto.  I assume this makes the most benefit of vector tiles, as there is no need to constantly interact with the server. 

However, I also have a bunch of charts, each of which is executing SQL against the carto database table, and the SQL is dependent on the filters applied in the filtering UI.  With the traditional carto workflow, I can just get new SQL from the filter UI, and use it to quickly update the raster tiles AND each of the charts without too much effort.   Everything that is dependent on the new filters does its own AJAX calls and updates itself accordingly, and I can keep everything modular. 

I could set a filter on the vector tile source and then use `queryRenderedFeatures()` on the vector tile layer, filter for unique features, and then use the resulting features to update the charts.  I am hesitant to do this for a number of reasons, one being that I don't want to couple the charts too closely with the map's filtering so I can re-use them elsewhere, and another being that I don't seem to get the same totals (`SELECT count(*)` in the database is usually higher than `getUnique(queryRenderedFeatures()).length` for some reason, possibly because vector tiles are hiding some features that should not be rendered at certain zoom levels?)  So, hitting the DB is best way for me to be 100% sure I have everything when doing aggregations.

[...]

PR #8048, opened 2019-03-17
add support for dynamic vector tiles rendering #3709

stepankuzmin:
Hey everyone! This PR adds the equivalent of GeoJSON Source `setData` to the Vector Sources, so a user can change sources without updating the whole map style.

Usage:

```js
map.addSource('example', {
  type: 'vector',
  url: 'mapbox://stepankuzmin.cjtd6d7i8107f3vqtku9o5jqn-7xl1l'
});

const source = map.getSource('example');
source.setSourceProperty('url', 'mapbox://stepankuzmin.cjtd6f5jh2eyk2xqtsyuo1kbz-3w7xz');
```

What do you think about it?

I'm not sure how to properly reload cached tiles, maybe someone can point me a direction for this?

### DIFF
--- a/src/source/vector_tile_source.js
+++ b/src/source/vector_tile_source.js
@@ -66,8 +66,6 @@ class VectorTileSource extends Evented implements Source {
             throw new Error('vector tile sources must have a tileSize of 512');
         }
 
-        // this._updateOptions(options);
-
         this.setEventedParent(eventedParent);
     }
 
@@ -107,28 +105,29 @@ class VectorTileSource extends Evented implements Source {
         this.load();
     }
 
-    _updateOptions(options: VectorTileSourceOptions) {
-        extend(this, pick(options, ['url', 'scheme', 'tileSize']));
-        this._options = extend({ type: 'vector' }, options);
-
-        this._collectResourceTiming = !!options.collectResourceTiming;
-
-        if (this.tileSize !== 512) {
-            throw new Error('vector tile sources must have a tileSize of 512');
-        }
-    }
-
-    setSourceProperty(name: string, value: mixed) {
+    setSourceProperty(callback: Function) {
         if (this._tileJSONRequest) {
             this._tileJSONRequest.cancel();
         }
 
-        const options = { [name]: value};
-        this._updateOptions(options);
+        callback();
 
         const sourceCache = this.map.style.sourceCaches[this.id];
         sourceCache.clearTiles();
         this.load();
+    }
+
+    setTiles(tiles: Array<string>) {
+        this.setSourceProperty(() => {
+            this._options.tiles = tiles;
+        });
+    }
+
+    setUrl(url: string) {
+        this.setSourceProperty(() => {
+            this.url = url;
+            this._options.url = url;
+        });
     }
 
     onRemove() {
--- a/src/source/vector_tile_source.js
+++ b/src/source/vector_tile_source.js
@@ -121,7 +121,7 @@ class VectorTileSource extends Evented implements Source {
         });
     }
 
-    setURL(url: string) {
+    setUrl(url: string) {
         this.setSourceProperty(() => {
             this.url = url;
             this._options.url = url;
--- a/src/source/vector_tile_source.js
+++ b/src/source/vector_tile_source.js
@@ -116,6 +116,8 @@ class VectorTileSource extends Evented implements Source {
             throw new Error('vector tile sources must have a tileSize of 512');
         }
 
+        const sourceCache = this.map.style.sourceCaches[this.id];
+        sourceCache.clearTiles();
         this.load();
     }
 
--- a/src/source/vector_tile_source.js
+++ b/src/source/vector_tile_source.js
@@ -19,6 +19,32 @@ import type {Callback} from '../types/callback';
 import type {Cancelable} from '../types/cancelable';
 import type {VectorSourceSpecification, PromoteIdSpecification} from '../style-spec/types';
 
+/**
+ * A source containing vector tiles in [Mapbox Vector Tile format](https://docs.mapbox.com/vector-tiles/reference/).
+ * (See the [Style Specification](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#vector) for detailed documentation of options.)
+ *
+ * @example
+ * map.addSource('some id', {
+ *     type: 'vector',
+ *     url: 'mapbox://mapbox.mapbox-streets-v6'
+ * });
+ *
+ * @example
+ * map.addSource('some id', {
+ *     type: 'vector',
+ *     tiles: ['https://d25uarhxywzl1j.cloudfront.net/v0.1/{z}/{x}/{y}.mvt'],
+ *     minzoom: 6,
+ *     maxzoom: 14
+ * });
+ *
+ * @example
+ * map.getSource('some id').setUrl("mapbox://mapbox.mapbox-streets-v6");
+ *
+ * @example
+ * map.getSource('some id').setTiles(['https://d25uarhxywzl1j.cloudfront.net/v0.1/{z}/{x}/{y}.mvt']);
+ * @see [Add a vector tile source](https://docs.mapbox.com/mapbox-gl-js/example/vector-source/)
+ * @see [Add a third party vector tile source](https://docs.mapbox.com/mapbox-gl-js/example/third-party/)
+ */
 class VectorTileSource extends Evented implements Source {
     type: 'vector';
     id: string;
@@ -115,17 +141,33 @@ class VectorTileSource extends Evented implements Source {
         this.load();
     }
 
+    /**
+     * Sets the source `tiles` property and re-renders the map.
+     *
+     * @param {string[]} tiles An array of one or more tile source URLs, as in the TileJSON spec.
+     * @returns {VectorTileSource} this
+     */
     setTiles(tiles: Array<string>) {
         this.setSourceProperty(() => {
             this._options.tiles = tiles;
         });
+
+        return this;
     }
 
+    /**
+     * Sets the source `url` property and re-renders the map.
+     *
+     * @param {string} url A URL to a TileJSON resource. Supported protocols are `http:`, `https:`, and `mapbox://<Tileset ID>`.
+     * @returns {VectorTileSource} this
+     */
     setUrl(url: string) {
         this.setSourceProperty(() => {
             this.url = url;
             this._options.url = url;
         });
+
+        return this;
     }
 
     onRemove() {
--- a/src/source/vector_tile_source.js
+++ b/src/source/vector_tile_source.js
@@ -19,8 +19,6 @@ import type {Callback} from '../types/callback';
 import type {Cancelable} from '../types/cancelable';
 import type {VectorSourceSpecification, PromoteIdSpecification} from '../style-spec/types';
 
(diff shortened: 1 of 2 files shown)


andrewharvey:
Hey @stepankuzmin I haven't looked in detail, but... 

Does this just replace a source's definition, ie. it's TileJSON?

ie. you can't say, change the geometry of this feature or change this feature's attributes?

If that's the case, I'd prefer not calling it setData as it's a bit different to a GeoJSON setData.

Just thinking out loud, is it better to have a more general setSourceProperty? Similar to setLayoutProperty and setPaintProperty to change any source values.

stepankuzmin:
Hey @andrewharvey! Yes, `setData` actually allows to redefine TileJSON endpoint, so I like the `setSourceProperty` approach more. I'll update the PR accordingly.

stepankuzmin:
I've changed `setData` to a more general `setSourceProperty`:

```js
source.setSourceProperty('url', 'mapbox://stepankuzmin.cjtd6f5jh2eyk2xqtsyuo1kbz-3w7xz');

source.setSourceProperty('tiles', [
  "mapbox://tiles/stepankuzmin.cjtd6f5jh2eyk2xqtsyuo1kbz-3w7xz/{z}/{x}/{y}.vector.pbf",
  "mapbox://tiles/stepankuzmin.cjtd6f5jh2eyk2xqtsyuo1kbz-3w7xz/{z}/{x}/{y}.vector.pbf"
]);
```

stepankuzmin:
I've added tiles cache clearing on `setSourceProperty`:

Kapture 2019-03-19 at 17 34 00

Akiyamka:
Great feature. Without this for a simple update of tile url will cause 
sequential call: 
```
map.addSource(<newSourceID>, {
  type: 'vector',
  url: <newTileURL>
});

// For each layer connected with old source
map.setLayoutProperty(<layerUsedOldSource>, 'source', <newSourceID> });

map.removeSource(<oldSourceID>);
```

aMoniker:
This will be great for our app, which changes source url parameters based on which filters a user chooses.

When replacing a source and clearing the cached tiles, will this prevent the flicker that currently happens when using `setStyle` to make a source change? It seems like that is what's happening in the demo gif.

stepankuzmin:
Hi, @aMoniker! Yes, this PR prevents flickering when updating sources.

src/source/vector_tile_source.js
```diff
@@ -94,6 +94,24 @@ class VectorTileSource extends Evented implements Source {
         this.load();
     }
 
+    setSourceProperty(name: string, value: mixed) {
+        if (this._tileJSONRequest) {
+            this._tileJSONRequest.cancel();
+        }
+
+        const options = { [name]: value};
+        extend(this, pick(options, ['url', 'scheme', 'tileSize']));
```
andrewharvey: I think it would be nicer if we could avoid the duplication of code between here and the constructor.

src/source/vector_tile_source.js
```diff
@@ -94,6 +94,24 @@ class VectorTileSource extends Evented implements Source {
...
+        this._options = extend({ type: 'vector' }, options);
+
+        if (this.tileSize !== 512) {
+            throw new Error('vector tile sources must have a tileSize of 512');
+        }
+
+        const sourceCache = this.map.style.sourceCaches[this.id];
+        sourceCache.clearTiles();
```
asheemmamoowala: The `load()` method should fire the `metadata` and `content` events that trigger a `SourceCache#reload()` and clears tiles when the new tilejson is available. See: 
https://github.com/mapbox/mapbox-gl-js/blob/4eb7d4e58c2fb4483994e7d364016c2188709e1a/src/source/source_cache.js#L69-L83

Is there a reason this additional call to `clearTiles` is needed here?

stepankuzmin: `SourceCache#reload()` doesn't remove existing tiles. It causes that after changing the source URL, you still have cached tiles from previous source URL, because they are not in `errored` state (see https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/source_cache.js#L225).

So I call `clearTiles` to force remove previously cached tiles.

stepankuzmin: We can force `errored` state thought, but it looks not so good as clearing tiles with `clearTiles `

src/source/vector_tile_source.js
stepankuzmin: I've moved options setter into `_updateOptions` method

stepankuzmin:
> @stepankuzmin Thanks for putting together this concept.
> 
> The current implementation of `setSourceProperty` leaves a bit unclear how
> 
> * `setSourceProperty` can be used to override properties (like min/maxzoom) of a source after it is set using a URL. The call to `load` would re-fetch the tilejson and write over those properties.
> * `load()` is asynchronous, but `this.options` would already have been updated with new values and referenced by `SourceCache#update` before the async load is complete.
> 
> This type of change is better suited to be handled via #4875

I think about passing an optional callback in `load()` method, so we can use it to update properties after TileJSON is loaded. Something like this:

```js
setSourceProperty(name: string, value: mixed) {
    if (this._tileJSONRequest) {
        this._tileJSONRequest.cancel();
    }

    const sourceCache = this.map.style.sourceCaches[this.id];
    sourceCache.clearTiles();

    this.load(() => {
        const options = { [name]: value};
        this._updateOptions(options);
    });
}
```

robinsummerhill:
This will be an absolute game-changer for us if this gets merged. We have a platform that uses 'parameterized' vector tile URLs to apply time-based filtering etc. To be able to change source URLs without producing the flicker associated with a full map repaint will be amazing.

arindam1993:
@stepankuzmin , thanks a lot for this! I'm closing this for now since there hasn't been any activity for a month or so. Feel free to repoen it once you have more progress on it.

andrewharvey:
> @stepankuzmin , thanks a lot for this! I'm closing this for now since there hasn't been any activity for a month or so. Feel free to repoen it once you have more progress on it.

@arindam1993 I think it's better left open (even as a draft if not completely ready yet) so it can continue to be worked on in a collective way. I'll try to help review if that's what it needs.

ansis:
@stepankuzmin thanks for working on this! Let me know if you would like help pushing any of these last things over the finish line

I think there are a couple tasks remaining before we can merge this:

- setting any property currently reloads the tilejson which can can overwrite the just-set property
- both this method and VectorTileSource need to be documented (I'm not sure why we haven't documented VectorTileSource yet...)
- settle on naming

### naming

It looks like existing methods to update source properties are one method per property. For example, `ImageSource#setCoordinates(...)` and `GeoJSONSource#setData(...)`. Should we follow this convention and have a `VectorTileSource#setUrl(...)`? I think this convention could be broken, but if we aren't sure that breaking it would be better I'd default to keeping with the existing convetion

stepankuzmin:
Hey @ansis!

Thanks for the clarification. I agree with one method per property and I'll get back to this issue soon.

stevage:
I've just found this issue as it's likely to be useful for a client I'm working with. The use case there is slightly different: updating server-side data, and wanting to force a tile refresh, without a change in URL. (Although likely to be implemented with a slightly different URL [ `?update=1` etc] to defeat caching.)

>It looks like existing methods to update source properties are one method per property. For example, ImageSource#setCoordinates(...) and GeoJSONSource#setData(...). Should we follow this convention and have a VectorTileSource#setUrl(...)? I think this convention could be broken, but if we aren't sure that breaking it would be better I'd default to keeping with the existing convetion

It looks like all the interest here is really just about `setUrl()` and `setTiles()`. Is there really any call for `setScheme()`, `setTileSize()`, `setMinzoom()` etc? If not, I'd suggest only implementing those two high-value methods, to avoid API bloat?

stepankuzmin:
Hi everyone! I've updated PR adding separate methods for setting source tiles and URL.

```js
source.setURL('mapbox://...');
```

```js
source.setTiles(['https://api.mapbox.com/v4/...']);
```

stevage:
Awesome to see!

Comment: I think maybe `setUrl` rather than `setURL`. It's a hard one to decide, but there is the `mapboxgl.baseApiUrl` property to give an indication...?

ogix:
Any updates on this PR? We also need this feature because all our layers are time-enabled.
@stepankuzmin can you do that change @stevage asked?
Is it the only thing that stops from merging?

stepankuzmin:
Hi everyone! Sorry for the late response. I've renamed `setURL` to `setUrl`, added unit tests and docs for this. Is there any chance for this to get merged?

/cc @ansis @arindam1993 @asheemmamoowala

ogix:
Are you guys still alive?

kukiel:
@stepankuzmin that's a great addition I must say. Would be nice to also support raster source with the same functionality.

andrewharvey:
> Any updates on this PR? We also need this feature because all our layers are time-enabled.

@ogix Just an aside that this shouldn't be needed for your use case (as I've understood it), GL JS will automatically re-request tiles which have expired based on the maxage cache header you set, check it out with the `traffic-day-v2` style.

andrewharvey:
I gave this a test drive, I couldn't find any issues,  seems to work as intended.

mourner:
Thanks everyone for the help landing this, and apologies for the delay!

underbluewaters:
Having a `setTiles()` method available for raster sources would be very helpful. I'd be willing to put together a PR if I could get a couple pointers. My use case is loading WMS services with support for toggling layers. Completely replacing the source gets tricky when there are referencing layers. More importantly, toggling additional layers causes all tiles to disappear before loading new ones which makes for a pretty awful UI compared to other tools.

I've found that following the same steps of calling `clearTiles()` on the source cache causes old tiles to disappear before the new ones are loaded. What I would like is for them to only be replaced when the new tile is ready (much like updating the url on an ImageSource). Calling `reload()` on the source cache gives me the behavior I want. Does that seem like a reasonable approach?

jonseppanen:
Hi @stepankuzmin, I noticed in comment https://github.com/mapbox/mapbox-gl-js/pull/8048#issuecomment-493739359 you mention this new feature prevents tile flickering.

When reloading a tile source using setTiles, the tiles vector layer is still flickering for us. Is this the way we should be using it to avoid flickering? I can't seem to find any other method.

Using latest mapbox 2.x.

stepankuzmin:
Hi @jonseppanen

Can you, please, provide some reproducible example for this?

robinsummerhill:
Hi @stepankuzmin. I'm also seeing flickering like @jonseppanen above.

You can see a minimal example here: https://codepen.io/rsummerhill/pen/dyOGXpQ?editors=0010
(this is using a tilejson source from our own GIS system)

From looking at the source code I'm guessing that features are removed from the map when `sourceCache.clearTiles()` is called in `setSourceProperty` and are not replaced until new tiles are loaded. This leads to flicker because there is a split second between removing the old tiles and loading the new ones.

stevage:
FWIW, I have found this to be an effective and flicker-free way of refreshing tiles:

```
map.style.sourceCaches['mylayer'].clearTiles();
map.style.sourceCaches['mylayer'].update(map.transform);
```

mrozema:
Is there a solution for the flickering when calling `setTile`? The minimal example from @robinsummerhill  is exactly what I'm observing. Using latest mapbox-gl (2.15.0). None of the workarounds seem to work for me.

PR #8053, opened 2019-03-19
Mapbox-gl-geocoder v4 examples

scottsfarley93:
## Launch Checklist

This PR adds several examples to demonstrate the functionality added in [v4](https://github.com/mapbox/mapbox-gl-geocoder/issues/197) of the mapbox-gl-geocoder. Specifically, new examples include:
- [x] A mapbox-gl-geocoder with a custom rendering function
- [x] A mapbox-gl-geocoder with language localization
- [x] A mapbox-gl-geocoder with custom flyTo animation properties

Additionally, several examples will be updated, including:
- [x] Updating the `Set a point after geocode result` to use the new built-in marker functionality in the gl-geocoder
- [x] ~Updating~ Remove the `Bias geocoder results around the map view` ~to use the built-in trackProximity functionality in the gl-geocoder~

**Note:** This is a WIP and these examples will not work until we have a working release candidate for version 4 available on api.mapbox.com.

\cc @katydecorah @yuletide @kbauhaus


andrewharvey:
> Updating the Bias geocoder results around the map view to use the built-in trackProximity functionality in the gl-geocoder

Whops, I forgot that's now done with `trackProximity`. I'd suggest we just remove this example since per https://github.com/mapbox/mapbox-gl-geocoder/issues/195, it does this by default now.

docs/pages/example/point-from-geocoder-result.html
```diff
@@ -13,7 +13,12 @@
     container: 'map',
     style: 'mapbox://styles/mapbox/streets-v9',
     center: [-79.4512, 43.6568],
-    zoom: 13
+    zoom: 13,
+    addToMap: true,
+    markerOptions: {
+        color: 'orange',
+        draggable: true
```
andrewharvey: What's the reasoning behind `draggable: true`? I get it shows options you can pass in, but I'm not sure if this makes sense as a default.

scottsfarley93: Entirely as an example of being able to pass in options. Perhaps just the `color` option would suffice as an illustrative example.

andrewharvey:
- [ ] There's a couple other examples which need the version tag updated.
- [ ] consider changing https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-accept-coordinates/ to not insert a comma which causes this:

Screenshot from 2019-03-27 14-24-06
Screenshot from 2019-03-27 14-24-18

Yes, in a proper implementation you could write your own render function, but in the interests of keeping the example simple and minimal we should just drop the comma.

docs/pages/example/mapbox-gl-geocoder-custom-render.html
```diff
@@ -0,0 +1,39 @@
...
+<link rel='stylesheet' href='https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v4.0.0/mapbox-gl-geocoder.css' type='text/css' />
+<style>
+.geocoder-dropdown-item{
+    padding: 5px
+}
+.geocoder-dropdown-icon{
+    margin-right: 10px;
+}
+.geocoder-dropdown-text{
+    color: darkgray;
```
katydecorah: This color value's contrast is very low, which will make it difficult for users with low vision to read. Maybe go with a darker gray? Or a fun purple `#5a3fc0` ?

```suggestion
    color: #666;
```

docs/pages/example/point-from-geocoder-result.js
```diff
@@ -1,10 +1,8 @@
 /*---
 title: Set a point after Geocoder result
 description: >-
-  Listen to the
-  [`result`](https://github.com/mapbox/mapbox-gl-geocoder/blob/master/API.md#on)
-  event from [mapbox-gl-geocoder](https://github.com/mapbox/mapbox-gl-geocoder)
-  and place a point on the coordinate results.
+  Style a [Marker](https://docs.mapbox.com/mapbox-gl-js/api/#marker) to use to mark the location of the [mapbox-gl-geocoder](https://github.com/mapbox/mapbox-gl-geocoder)
```
katydecorah:
```suggestion
  Style a [Marker](https://docs.mapbox.com/mapbox-gl-js/api/#marker) used to mark the location of the [mapbox-gl-geocoder](https://github.com/mapbox/mapbox-gl-geocoder)
```

PR #8063, opened 2019-03-21
Use @mapbox/gazetteer for benchmark coordinates

### LINKED ISSUE #8059: Replace style_locations with @mapbox/gazetteer
Right now, there's a locally managed [`bench/lib/style_locations.js`](https://github.com/mapbox/mapbox-gl-js/blob/master/bench/lib/style_locations.js) file that is the mirrored version of style-benchmark-locations.json in [`@mapbox/gazetteer`](https://github.com/mapbox/gazetteer/blob/mb-pages/mapbox-streets/style-benchmark-locations.json). To keep everything in sync so that other platforms (native) are using the same location data in their benches we should replace what's here with that.

tristen:
Replace `bench/lib/style_locations.js` with the same place data made available in `@mapbox/gazetteer`. 

@ansis thanks for the tip to use `MercatorCoordinate.fromLngLat` to derive tile index. Could I get you to review this?

---
Closes #8059

### DIFF
--- a/bench/lib/style_locations.js
+++ b/bench/lib/style_locations.js
@@ -1,12 +1,10 @@
-/* eslint camelcase: 0 */
-
 import styleBenchmarkLocations from '@mapbox/gazetteer/mapbox-streets/style-benchmark-locations.json';
 import MercatorCoordinate from '../../src/geo/mercator_coordinate';
 import { OverscaledTileID } from '../../src/source/tile_id';
 
 export default styleBenchmarkLocations.features.map(feature => {
     const { coordinates } = feature.geometry;
-    const { zoom, place_name } = feature.properties;
+    const { zoom } = feature.properties;
     const { x, y } = MercatorCoordinate.fromLngLat({
         lng: coordinates[0],
         lat: coordinates[1]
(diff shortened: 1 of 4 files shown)


tristen:
🤔 It's not clear to me why tests are failing.

tristen:
> The bench changes look good (aside from the test failures), but did you accidentally ran `yarn upgrade`?

@mourner hmm I don't _think_ I did? I just tried blowing away `node_modules/` + `yarn.lock`,  running `yarn` again and I don't any difference against the lock file :/

ryanhamley:
You may need to just update Yarn. That fixed this issue for me.

tristen:
> You may need to just update Yarn. That fixed this issue for me.

@ryanhamley that seemed promising but there are still no changes against `yarn.lock` after upgrading yarn to `1.15.2`.

mourner:
@tristen do not delete the `yarn.lock` file. Run `yarn` when it's reset to its master state.

bench/lib/style_locations.js
```diff
@@ -1,3 +1,5 @@
+/* eslint camelcase: 0 */
```
mourner: Just do `feature.properties['place_name']` instead.

package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@mapbox/appropriate-images-react": "^1.0.0",
     "@mapbox/dr-ui": "0.6.0",
+    "@mapbox/gazetteer": "^3.1.0",
```
mourner: Needs to be in devDeps.

tristen: Oy ... me and yarn today :/

Issue #6224, opened 2018-02-23
remove anonymous gist dependency from benchmarks 


mollymerp:
per https://github.com/blog/2503-deprecation-notice-removing-anonymous-gist-creation

![image](https://user-images.githubusercontent.com/2425307/36615378-d845a4a6-1894-11e8-9d18-f05cf71afd97.png)

jfirebaugh:
Alternatives:

* Create an oauth'd GitHub app that can gist on behalf of the user (cons: oauth is a pain, this will pollute folks' gist collections)
* Upload directly to S3 bucket (cons: requires access to AWS credentials)
* Find some other anonymous hosting service

anandthakker:
Darn, just tried to fix this by generating a data URI that could be pasted into a PR, but GH markdown sanitizes that away (which makes sense)

mourner:
There are no viable and easy alternatives, so we'll remove the share button for now -> #8066

PR #8066, opened 2019-03-22
Fix and refactor the benchmark suite

### LINKED ISSUE #8064: Fix broken benchmarks build (or drop React)
After merging `publisher-production` branch with the docs upgrades, `yarn run build-benchmarks` started failing, with `rollup-plugin-commonjs` complaining about unexpected token in `react-dom` (although the exact location is bogus — there's nothing suspicious there). This package upgraded from v16.0.0 to v16.7.0, and something in between those versions broke the setup. I tracked it down specifically to v16.4.0, starting from which the builds fail. Also confimed that the builds still fail in v16.8.2.

This is really cryptic and I don't have yet good ideas about what causes this, but I think the best fix would be to refactor and cut out all this React bloat from benchmarks for good. 😡

mourner:
Fixes benchmarks builds, closes #8064. Gets the uploaded build down from 10MB to 719k.

- [x] Fix failing benchmark builds on master
- [x] Make source maps external to avoid huge benchmark builds (previously, 10MB)
- [x] Turn on production mode for benchmark builds so that they're minified/unasserted (had to switch to explicit benchmark names for that because class names are mangled in minified build)
- [x] Remove broken sharing button (that we have no replacement for, #6224)
- [x] Simplify benchmark build configs and watch scripts
- [x] Introduce separate `benchmark_view` build again so that React/d3 and UI logic isn't built into benchmark builds
- [x] DRY up code between styles and versions benchmarks


bench/styles/index.html
```diff
@@ -8,12 +8,15 @@
...
     <link rel='stylesheet' href='/dist/mapbox-gl.css'/>
-    <script href='https://unpkg.com/url-search-params@0.10.0/build/url-search-params.js'></script>
 </head>
 
 <body>
     <div id="benchmarks"></div>
     <script src="/bench/styles/benchmarks_generated.js"></script>
+    <script src="/bench/benchmarks_view_generated.js"></script>
+    <script>
+        window.Benchmarks.run(benchmarks);
```
tristen: Is the `benchmarks` variable provided by benchmarks_view_generated.js?

tristen: Trying to think through any consequences of calling `Benchmark.run` for my submodule use case of the benchmark suite. I think its probably fine but I'm just curious the reasoning here.

bench/benchmarks_view.js
```diff
@@ -1,12 +1,89 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import * as d3 from 'd3';
-import Axis from './lib/axis';
```
tristen: I kind of wish all the React components weren't defined in one script but 🤷‍♂️

bench/styles/index.html
mourner: @tristen it's provided by `index.html` and `benchmarks.js` of each respective benchmark. Styles and versions benchmarks need different treatment here — versions benchmarks group by name in a global variable because versions are ran as separate builds (and we need to keep compat with previous benchmark builds), and styles benchmarks build the benchmarks array in the build itself (because we don't upload them, and they need additional metadata like location).

bench/benchmarks_view.js
mourner: I thought it would be nice to put all UI in one file, so that there's clear separation — `lib` only contains logic that gets insider the different benchmarks builds, and `bencharks_view` encapsulates all the view logic. As a follow-up, we could split the view components again in a separate folder (e.g. `view` or `components`).

ryanhamley: Splitting up the components may be a good opportunity to remove React too. Could be easier to do if the UI is split into smaller, discrete pieces

ryanhamley:
The `styles` folder is missing a rollup config and the style benchmarks are currently only running against `master`

mourner:
> The styles folder is missing a rollup config

All the configs are centralized in one file (`bench/rollup_config_benchmarks.js`) for less duplication. I left the config file in `versions` (although it reuses the one above) so that CI can build it without building other benchmarks. See this commit https://github.com/mapbox/mapbox-gl-js/pull/8066/commits/ba571a04ba162b6bb3a5192877ba6759da5cd60d

bench/benchmarks_view.js
mourner: Sounds good! Let's follow up with splitting components in a different PR (don't want to blow up the scope of this one too much).

mourner:
Here's an example timing of loading a benchmark without cache — the first line is v0.53.1, the second is `master` from S3 (after merging this PR), and the third is the local bench build. It's smaller than S3 master because gzip is turned off for S3 bench uploads for some reason. We should probably look into that some time.

image

PR #8068, opened 2019-03-22
Allow overriding SymbolLayer's paint properties using format expression options

alexshalamov:
Port of https://github.com/mapbox/mapbox-gl-native/pull/14062

## Launch Checklist

 - [x] write tests for all new functionality. Render tests (https://github.com/mapbox/mapbox-gl-js/pull/8093, https://github.com/mapbox/mapbox-gl-js/pull/8046)
 - [x] post benchmark scores

### DIFF
--- a/src/data/bucket/symbol_bucket.js
+++ b/src/data/bucket/symbol_bucket.js
@@ -535,8 +535,7 @@ class SymbolBucket implements Bucket {
 
         const angle = (this.allowVerticalPlacement && writingMode === WritingMode.vertical) ? Math.PI / 2 : 0;
 
-        for (const symbol of quads) {
-
+        const addSymbol = (symbol: SymbolQuad) => {
             const tl = symbol.tl,
                 tr = symbol.tr,
                 bl = symbol.bl,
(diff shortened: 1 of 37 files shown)


alexshalamov:
Benchmark results:

Screen Shot 2019-04-02 at 20 49 43

Screen Shot 2019-04-02 at 20 50 32

src/data/bucket/symbol_bucket.js
```diff
@@ -324,6 +325,14 @@ class SymbolBucket implements Bucket {
     }
 
     createArrays() {
+        const layout = this.layers[0].layout;
+        this.hasPaintOverrides = SymbolStyleLayer.hasPaintOverrides(layout);
+        if (this.hasPaintOverrides) {
+            for (const layer of this.layers) {
+                layer.setPaintOverrides();
```
ansis: It looks like a SymbolStyleLayer can exist either with overrides applied or not. And since the state depends on this method call it is hard to tell elsewhere what the state of the style layer is. Your current implementation looks correct but I think someone less familiar with this could easily make changes that either result in `setPaintOverrides()` being called more or less than once, or before `recalculate()`. 

Would triggering it in `styleLayer.recalculate(...)` make sense?

src/style/style_layer/symbol_style_layer.js
```diff
@@ -77,6 +98,83 @@ class SymbolStyleLayer extends StyleLayer {
...
+
+            const expr: ZoomConstantExpression<'source'> = ((textField.value): any);
+            if (expr._styleExpression) {
+                checkExpression(expr._styleExpression.expression);
+                if (!sections.length) {
+                    expr._styleExpression.expression.eachChild(checkExpression);
+                }
+            }
+        }
+
```
ansis: I think at this point `sections` is set to the sections of a single formatted expression. In the case where there is a ZoomConstantExpression with multiple possible formatted expressions it will only hold the last one. If the last expression has no overrides but an earlier one does, does this return the wrong value?

For example:

```json
["case",
    ["get", "colourful"], ["format", "red text", { "text-color": "red" }],
    ["format", "default text", {}]]
```

src/style/properties.js
```diff
@@ -535,9 +535,11 @@ export class DataConstantProperty<T> implements Property<T, T> {
  */
 export class DataDrivenProperty<T> implements Property<T, PossiblyEvaluatedPropertyValue<T>> {
     specification: StylePropertySpecification;
+    overrides: ?Object;
 
-    constructor(specification: StylePropertySpecification) {
+    constructor(specification: StylePropertySpecification, overrides?: Object) {
```
ansis: Would it be possible to make the type of `overrides` more specific?

Issue #8071, opened 2019-03-23
The default marker is positioned incorrectly


dmytro-gokun:
I'm creating a default droplet-shaped marker (by omitting options.element as described here: https://docs.mapbox.com/mapbox-gl-js/api/#marker). As a result the marker is positioned slightly higher then it should be. You can see it here:

https://jsbin.com/ragudog/5/edit?html,output

The reason is that <svg> is an inline element and the size of the div is determined by the size of the svg + line-height setting (which is 20px for the map by default).

So, in order to fix this, we need to change that default droplet svg to be a "block" element. If everybody agrees, I will submit a PR to fix this issue.

dmytro-gokun:
After some thinking ... one could argue that when using the default marker, anchor & offset should be overridden as well, so that the droplet's pin point to exact location. Any thoughts?

mourner:
Yes, we'll welcome a PR! The sharp end point definitely should point to exact location.

dmytro-gokun:
@mourner Deal. I'll work on this.

PR #8074, opened 2019-03-25
Fix default marker positioning

dmytro-gokun:
As described here: https://github.com/mapbox/mapbox-gl-js/issues/8071


mourner:
Thank you!

Issue #8010, opened 2019-03-09
Abilitry to set tooltips on standard controls/change them on the fly.


dmytro-gokun:
This will give us ability to translate our apps properly w/o need to hack into the standard controls's guts and changing attributes on them.

dmytro-gokun:
Closed by https://github.com/mapbox/mapbox-gl-js/pull/8012

dmytro-gokun:
@mourner @andrewharvey 
Now, when https://github.com/mapbox/mapbox-gl-js/pull/8012 is merged in, I'd like to work on this one. But before I go ahead and create a pull request (s), I'd like to hear your opinion on this.

First of all, is it better to have a separate pull request per a control or do them all at once? (I'd like to start with FullscreenControl first, just to use that as a prototype for the rest of controls).

Secondly, my idea was to add new optional parameters to controls' constructors. Do you like it?

Another idea was to give ability to change the tooltip(s) after the control is created (this is for advanced scenarios, like when UI language is changed during the app's lifetime). But after some thinking, i decided to leave this out for now.

So, plz let me know what you think here. If you are okay with the idea, I'll go ahead and create a PR.

andrewharvey:
I'm :-1: to implementing fine grain ability to set tooltip values. These controls are standard out of the box features and the default tooltips should be sufficient.

If you're trying to re-purpose the controls to do something else, best to create a new control entirely.

If it's for language localisation, I think it's better we implement that across the library, rather than by passing strings directly like this.

For example this could be a uiLanguage option to Map, and we have a fixed table of language strings to lookup, or the user could provide their own translation by passing an object to uiLanguage.

dmytro-gokun:
@andrewharvey You are right, the main purpose here is localization. We want to support different languages in our app and mapbox kind of sticks out with its English-only text.

The approach you're suggesting is pretty neat and clean. However, it looks like it is going to be a pretty big task for a single PR. May be it has to be split into several steps.

A concern here is the library size. If we're going to support a bunch of languages and have a bunch of strings to translate, all that may end up increasing bundle size significantly. Even for those customers who are not interested in particular languages. Then, may be it's better to have a separate package per UI language?

Another issue I can see with this approach is that it is pretty hard to support. Say, we have support for 50 languages and then we need to add one new string to the UI.

Then what about custom controls. Should they use this central "string DB" (may be, by extending it on  the fly?) or use their own localization strategy?

andrewharvey:
[...]

PR #8076, opened 2019-03-25
Add ability to customize FullscreenControl's tooltips

dmytro-gokun:
As described here: https://github.com/mapbox/mapbox-gl-js/issues/8010
If this is accepted, I'd like to add similar options for the rest of the standard controls (where applicable).


andrewharvey:
If this is done so they can be localised, I think it's better to implement localisation across the library, instead of implementing this on text string. I'll leave more comments on that at #8010

Issue #6217, opened 2018-02-23
Two taps (within 300ms) at different locations should not trigger zoom


jfirebaugh:
A "double tap zoom" should be triggered only when the touch locations are relatively close to each other.

Also, it should be triggered on the second `touchend`, not the second `touchstart`.

asheemmamoowala:
Closed by  #8086

PR #8086, opened 2019-03-27
Fix double-tap zoom (#6217)

vakila:
Hi folks! Per chat with @asheemmamoowala I'm jumping in on a `good first issue`, trying to fix the double-tap zoom issues described in #6217. Thanks in advance for the feedback :pray: 

### Changes:
- Add tests for existing double-click zoom functionality & fix up existing spuriously-passing test
- Prevent zooming when two taps are within 300ms but in different locations (based on best-guess threshold; see below), & add tests
- Zoom on the second `touchend` event (not `touchstart`) of double-tap within the time threshold, & add tests

### Known issues/questions:
- I've set the `maxDelta` between tap locations (in both x & y) to 30, rather arbitrarily based on my manual tests - happy to use a different value if there is some standard for this. 
- On the Android device I manually tested with, in addition to touch events being fired a dblclick event is also fired when the double-tap points are within a distance threshold, which is somewhat higher than the threshold I set :point_up: - this means it's possible for two slightly-farther-apart touch events to trigger zoom via the dblclick handler, rather than the double-tap handler. That behavior seems OK to me, but just wanted to flag it. 
- The "Size" check shows this as adding +483 B - Is that normal/OK for such a small fix? If there are ways I can improve this I'm all ears

## Launch Checklist

 - ~document any changes to public APIs~ N/A afaict
 - [x] post benchmark scores :point_right: [screenshot](https://user-images.githubusercontent.com/5424927/55045363-c65e5500-4ffa-11e9-8013-af528a2c77f0.png)  (charts seem broken in Firefox :cry: )
 - [x] manually test the debug page - didn't notice any breakage on debug or events pages on desktop (Chrome & Firefox on Linux) or mobile (Chrome on Android)

### DIFF
--- a/src/ui/handler/dblclick_zoom.js
+++ b/src/ui/handler/dblclick_zoom.js
@@ -14,6 +14,7 @@ class DoubleClickZoomHandler {
     _enabled: boolean;
     _active: boolean;
     _tapped: ?TimeoutID;
+    _tappedPoint: ?{x: number, y: number};
 
     /**
      * @private
@@ -71,12 +72,22 @@ class DoubleClickZoomHandler {
         if (!this.isEnabled()) return;
         if (e.points.length > 1) return;
 
+        const maxDelta = 0;
+
         if (!this._tapped) {
-            this._tapped = setTimeout(() => { this._tapped = null; }, 300);
+            this._tappedPoint = e.points[0];
+            this._tapped = setTimeout(() => { this._tapped = null; this._tappedPoint = null; }, 300);
         } else {
+            const newTap = e.points[0];
+            const firstTap = this._tappedPoint;
+
             clearTimeout(this._tapped);
             this._tapped = null;
-            this._zoom(e);
+            this._tappedPoint = null;
+
+            if (firstTap && Math.abs(firstTap.x - newTap.x) <= maxDelta && Math.abs(firstTap.y - newTap.y) <= maxDelta) {
+                this._zoom(e);
+            }
         }
     }
 
--- a/src/ui/handler/dblclick_zoom.js
+++ b/src/ui/handler/dblclick_zoom.js
@@ -81,14 +81,34 @@ class DoubleClickZoomHandler {
             const newTap = e.points[0];
             const firstTap = this._tappedPoint;
 
-            clearTimeout(this._tapped);
-            this._tapped = null;
-            this._tappedPoint = null;
-
             if (firstTap && Math.abs(firstTap.x - newTap.x) <= maxDelta && Math.abs(firstTap.y - newTap.y) <= maxDelta) {
-                this._zoom(e);
+                const onTouchEnd = () => {
+                    if (this._tapped) { // make sure we are still within the timeout window
+                        this._zoom(e); // pass this touchstart event, as touchend events have no points
+                    }
+                    this._map.off('touchend', onTouchEnd);
+                    this._resetTapped();
+                };
+
+                const onTouchCancel = () => {
+                    this._map.off('touchend', onTouchEnd);
+                    this._map.off('touchcancel', onTouchCancel);
+                    this._resetTapped();
+                };
+
+                this._map.on('touchend', onTouchEnd);
+                this._map.on('touchcancel', onTouchCancel);
+
+            } else { // touches are too far apart, don't zoom
+                this._resetTapped();
             }
         }
+    }
+
+    _resetTapped() {
+        clearTimeout(this._tapped);
+        this._tapped = null;
+        this._tappedPoint = null;
     }
 
     onDblClick(e: MapMouseEvent) {
--- a/src/ui/handler/dblclick_zoom.js
+++ b/src/ui/handler/dblclick_zoom.js
@@ -72,7 +72,7 @@ class DoubleClickZoomHandler {
         if (!this.isEnabled()) return;
         if (e.points.length > 1) return;
 
-        const maxDelta = 0;
+        const maxDelta = 30;
 
         if (!this._tapped) {
             this._tappedPoint = e.points[0];
--- a/src/ui/handler/dblclick_zoom.js
+++ b/src/ui/handler/dblclick_zoom.js
@@ -82,6 +82,8 @@ class DoubleClickZoomHandler {
             const firstTap = this._tappedPoint;
 
             if (firstTap && Math.abs(firstTap.x - newTap.x) <= maxDelta && Math.abs(firstTap.y - newTap.y) <= maxDelta) {
+                e.originalEvent.preventDefault(); // prevent duplicate zoom on dblclick
+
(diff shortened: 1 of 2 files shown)


src/ui/handler/dblclick_zoom.js
```diff
@@ -71,15 +72,47 @@ class DoubleClickZoomHandler {
...
+            this._tappedPoint = e.points[0];
+            this._tapped = setTimeout(() => { this._tapped = null; this._tappedPoint = null; }, 300);
         } else {
-            clearTimeout(this._tapped);
-            this._tapped = null;
-            this._zoom(e);
+            const newTap = e.points[0];
+            const firstTap = this._tappedPoint;
+
+            if (firstTap && Math.abs(firstTap.x - newTap.x) <= maxDelta && Math.abs(firstTap.y - newTap.y) <= maxDelta) {
```
mourner: You could also use `firstTap.dist(newTap) <= maxDelta`.

src/ui/handler/dblclick_zoom.js
```diff
@@ -71,15 +72,47 @@ class DoubleClickZoomHandler {
...
+                };
+
+                const onTouchCancel = () => {
+                    this._map.off('touchend', onTouchEnd);
+                    this._map.off('touchcancel', onTouchCancel);
+                    this._resetTapped();
+                };
+
+                this._map.on('touchend', onTouchEnd);
+                this._map.on('touchcancel', onTouchCancel);
```
mourner: To simplify this a bit, you can do `this._map.once(...)` instead of `on` — this way listener doesn't have to manually remove itself. It also ensures `touchcancel` listeners are not accumulated if they never happen.

test/unit/ui/handler/dblclick_zoom.test.js
```diff
@@ -18,9 +33,152 @@ test('DoubleClickZoomHandler does not zoom if preventDefault is called on the db
...
+    t.notOk(zoom.called, 'should not trigger zoom before second touchend');
+
+    simulate.touchcancel(map.getCanvas());
+    simulate.touchend(map.getCanvas());
+    map._renderTaskQueue.run();
+    t.notOk(zoom.called, 'should not trigger zoom if second touch is canceled');
+
+    simulate.touchstart(map.getCanvas(), {touches: [{clientX: 0.5, clientY: 0.5}]});
+    simulate.touchend(map.getCanvas());
+    simulate.touchstart(map.getCanvas(), {touches: [{clientX: 0.5, clientY: 0.5}]});
```
mourner: Small nit: maybe extract `canvas` into a variable instead of repeated `getCanvas()` calls? Also the event data (`{touches: ...}`) that gets repeated.

src/ui/handler/dblclick_zoom.js
mourner: Ah, disregard the last sentence — it seems like we still need to make sure `touchcancel` listeners don't accumulate with every tap when touches are not cancelled.

vakila:
Thanks for the reviews @mourner @asheemmamoowala ! I'll take all your comments into account & push a new set of changes that incorporates both.

src/ui/handler/dblclick_zoom.js
```diff
@@ -71,15 +72,47 @@ class DoubleClickZoomHandler {
...
-            this._zoom(e);
+            const newTap = e.points[0];
+            const firstTap = this._tappedPoint;
+
+            if (firstTap && Math.abs(firstTap.x - newTap.x) <= maxDelta && Math.abs(firstTap.y - newTap.y) <= maxDelta) {
+                e.originalEvent.preventDefault(); // prevent duplicate zoom on dblclick
+
+                const onTouchEnd = () => {
+                    if (this._tapped) { // make sure we are still within the timeout window
+                        this._zoom(e); // pass this touchstart event, as touchend events have no points
```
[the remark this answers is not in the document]
vakila: Was hoping the comment would convey this but: `e` here is actually the `touchstart` event, which I'm passing through to the `touchend` handler created _within_ the `touchstart` handler here, and just ignoring the actual `touchend` event because it has no points we can use.

a) Is there a better, less-hacky solution? 
b) If not, is there a better way I can phrase this comment to describe what's going on?

test/unit/ui/handler/dblclick_zoom.test.js
```diff
@@ -18,9 +33,152 @@ test('DoubleClickZoomHandler does not zoom if preventDefault is called on the db
...
+
+});
+
+test('DoubleClickZoomHandler does not zoom on double tap if touchstart events are > 300ms apart', (t) => {
+    const map = createMap(t);
+
+    const zoom = t.spy();
+    map.on('zoom', zoom);
+
+    const simulateDelayedDoubleTap = () => {
```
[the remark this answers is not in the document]
vakila: :+1: yep as you noticed above, I had this pulled out into a separate function at one point, but then decided that as I was only using it twice it would be more readable to simply inline it each time (and in that process, missed removing the old params in L63). I've now pulled it back out :heavy_check_mark:

src/ui/handler/dblclick_zoom.js
vakila: :+1: thanks for catching the missing `touchcancel` removal! I've now used `.once()` for both listeners and still manually remove the other listener in each handler, so I think all bases should be covered.

src/ui/handler/dblclick_zoom.js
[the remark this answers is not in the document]
asheemmamoowala: I had to re-read the comment to correct my understanding of what is happening here. What tripped me up is that `onTouchEnd` has no event parameter. 

Would `//pass the original touchstart event...` make it easier to understand? If not, I am okay with keeping the existing wording.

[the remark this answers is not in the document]
vakila: :+1: thanks @asheemmamoowala, yep I think "original" makes it clearer - edited in https://github.com/mapbox/mapbox-gl-js/pull/8086/commits/61362b03c557a37164e457c518fd92bb276f56fa. I also added a comment calling attention to the fact that we're ignoring the `touchend` event. Hope that clears things up!

src/ui/handler/dblclick_zoom.js
```diff
@@ -4,6 +4,10 @@ import { bindAll } from '../../util/util';
 
 import type Map from '../map';
 import type {MapMouseEvent, MapTouchEvent} from '../events';
+import Point from '@mapbox/point-geometry';
```
mourner: `import type Point` if it's only used for typing.

vakila: :+1: thanks for the catch! Fixed in https://github.com/mapbox/mapbox-gl-js/pull/8086/commits/ae80d83b06da759e149e560c667f58da6e1b9bf2

PR #8087, opened 2019-03-27
Fix removeFeatureState bug

### LINKED ISSUE #8026: removeFeatureState does not remove state from the original zoom level
### Steps to reproduce

https://jsfiddle.net/u3cvf8sx/

1. Click on Africa to highlight (notice we are default to zoom level 1).
2. Now zoom in via mouse scroll wheel until zoom level changes to 2 or higher.
3. Now click anywhere in ocean area, and the highlight on Africa will disappear as expected. This is achieved by calling removeFeatureState on the layer.
4. Now without clicking anything but use mouse scroll wheel to zoom out again back to zoom level 1.

### Expected behavior

highlight is removed for all zoom levels

### Actual behavior

The highlight is back again for zoom level 1 only. As if the feature's state at zoom level 1 hasn't been removed.

### Configuration
**Browser:** Chrome
**Operating systems + versions:**

asheemmamoowala:
Closes #8026.

Feature state is applied only to those features identified by an `id` that have state set. The `removeFeatureState` method, when called to clear state on all features in a source/source-layer, didn't clear the state of each feature, it cleared all the state for that source.  i.e. `
{ idA: {..}, idB:{...} }` became `{}` instead of `{ idA: {}, idB: {}}`.

This PR fixes this by preserving the feature id that have had state applied previously, and clearing out that state. This allows the empty state to be applied to tiles that are being 'refreshed' when restored from the cache.

<details>
<summary>Benchmark results</summary>
Screen Shot 2019-03-26 at 10 05 41 PM
Screen Shot 2019-03-26 at 10 03 47 PM
Screen Shot 2019-03-26 at 10 04 08 PM
Screen Shot 2019-03-26 at 10 04 18 PM
</details>

## Launch Checklist

 - [x] post benchmark scores

cc @peterqliu @kbauhaus


kkaefer:
Does this need porting to Native?

asheemmamoowala:
No, because there is no native implementation of feature state as yet.

PR #8093, opened 2019-03-28
Add render test for format expression nested under decision expression

alexshalamov:
Render test for https://github.com/mapbox/mapbox-gl-native/issues/14254
GL native fix: https://github.com/mapbox/mapbox-gl-native/pull/14259

## Launch Checklist


alexshalamov:
> Are separate tests needed for nested format expressions with text-scale and text-font as well?

I don't think so, I quickly made few tests with font-scale and text-font options and these are not exercising code that I'm adding in https://github.com/mapbox/mapbox-gl-js/pull/8068, Chris made quite a good test suite for layout specific properties. If I find that something is missing, I would definitely add more unit / render tests.

PR #8095, opened 2019-03-28
Attempt to localize mapbox UI

dmytro-gokun:
This is second attempt to implement UI localization, as described here: https://github.com/mapbox/mapbox-gl-js/issues/8010

@andrewharvey reviewed my previous PR and suggested a better idea, which i'm trying to implement here. For starter, I only implemented localization for FullscreenControl. If the overall approach is approved by contributors, I'll work to add localization for the rest of controls. Also, I'm able to add Russian, Ukrainian & Danish translation (but those should, probably, go as a separate PR or even separate NPM packages).

Any feedback is welcome.

### DIFF
--- a/src/ui/control/fullscreen_control.js
+++ b/src/ui/control/fullscreen_control.js
@@ -93,9 +93,13 @@ class FullscreenControl {
     }
 
     _updateTitle() {
-        const title = this._isFullscreen() ? "Exit fullscreen" : "Enter fullscreen";
+        const title = this._getTitle();
         this._fullscreenButton.setAttribute("aria-label", title);
         this._fullscreenButton.title = title;
+    }
+
+    _getTitle() {
+        return this._map._getUIString(this._isFullscreen() ? 'FullscreenControl.Exit' : 'FullscreenControl.Enter');
     }
 
     _isFullscreen() {
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -44,6 +44,7 @@ import type DragPanHandler, {DragPanOptions} from './handler/drag_pan';
 import type KeyboardHandler from './handler/keyboard';
 import type DoubleClickZoomHandler from './handler/dblclick_zoom';
 import type TouchZoomRotateHandler from './handler/touch_zoom_rotate';
+import defaultLocale from './default_locale';
 import type {TaskID} from '../util/task_queue';
 import type {Cancelable} from '../types/cancelable';
 import type {
@@ -96,7 +97,8 @@ type MapOptions = {
     renderWorldCopies?: boolean,
     maxTileCacheSize?: number,
     transformRequest?: RequestTransformFunction,
-    accessToken: string
+    accessToken: string,
+    locale?: Object
 };
 
 const defaultMinZoom = 0;
@@ -235,7 +237,7 @@ const defaultOptions = {
  * @param {number} [options.fadeDuration=300] Controls the duration of the fade-in/fade-out animation for label collisions, in milliseconds. This setting affects all symbol layers. This setting does not affect the duration of runtime styling transitions or raster tile cross-fading.
  * @param {boolean} [options.crossSourceCollisions=true] If `true`, symbols from multiple sources can collide with each other during collision detection. If `false`, collision detection is run separately for the symbols in each source.
  * @param {string} [options.accessToken=null] If specified, map will use this token instead of the one defined in mapboxgl.accessToken.
-
+ * @param {string} [options.locale=null] A patch to apply to the default localization table for UI strings, e.g. control tooltips. The `locale` object maps namespaced UI string IDs to translated strings in the target language; see `src/ui/default_locale.js` for an example with all supported string IDs. The object may specify all UI strings (thereby adding support for a new translation) or only a subset of strings (thereby patching the default translation table).
  * @example
  * var map = new mapboxgl.Map({
  *   container: 'map',
@@ -293,6 +295,7 @@ class Map extends Camera {
     _mapId: number;
     _localIdeographFontFamily: string;
     _requestManager: RequestManager;
+    _locale: Object;
 
     /**
      * The map's {@link ScrollZoomHandler}, which implements zooming in and out with a scroll wheel or trackpad.
@@ -374,6 +377,7 @@ class Map extends Camera {
         this._renderTaskQueue = new TaskQueue();
         this._controls = [];
         this._mapId = uniqueId();
+        this._locale = extend({}, defaultLocale, options.locale);
 
         this._requestManager = new RequestManager(options.transformRequest, options.accessToken);
 
@@ -1166,6 +1170,15 @@ class Map extends Camera {
             this._localIdeographFontFamily = options.localIdeographFontFamily;
             return this._updateStyle(style, options);
         }
+    }
+
+    _getUIString(key: string) {
+        const str = this._locale[key];
+        if (str == null) {
+            throw new Error(`Missing UI string '${key}'`);
+        }
+
+        return str;
     }
 
     _updateStyle(style: StyleSpecification | string | null,  options?: {diff?: boolean} & StyleOptions) {
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -263,8 +263,9 @@ class GeolocateControl extends Evented {
                 this._geolocateButton.classList.remove('mapboxgl-ctrl-geolocate-background');
                 this._geolocateButton.classList.remove('mapboxgl-ctrl-geolocate-background-error');
                 this._geolocateButton.disabled = true;
-                this._geolocateButton.title = 'Location not available';
-                this._geolocateButton.setAttribute('aria-label', 'Location not available');
+                const title = this._map._getUIString('GeolocateControl.LocationNotAvailable');
+                this._geolocateButton.title = title;
+                this._geolocateButton.setAttribute('aria-label', title);
 
                 if (this._geolocationWatchID !== undefined) {
                     this._clearWatch();
@@ -293,13 +294,17 @@ class GeolocateControl extends Evented {
         this._geolocateButton = DOM.create('button', `mapboxgl-ctrl-geolocate`, this._container);
         DOM.create('span', `mapboxgl-ctrl-icon`, this._geolocateButton).setAttribute('aria-hidden', true);
         this._geolocateButton.type = 'button';
-        this._geolocateButton.title = 'Find my location';
-        this._geolocateButton.setAttribute('aria-label', 'Find my location');
+
         if (supported === false) {
             warnOnce('Geolocation support is not available so the GeolocateControl will be disabled.');
+            const title = this._map._getUIString('GeolocateControl.LocationNotAvailable');
             this._geolocateButton.disabled = true;
-            this._geolocateButton.title = 'Location not available';
-            this._geolocateButton.setAttribute('aria-label', 'Location not available');
+            this._geolocateButton.title = title;
+            this._geolocateButton.setAttribute('aria-label', title);
+        } else {
+            const title = this._map._getUIString('GeolocateControl.FindMyLocation');
+            this._geolocateButton.title = title;
+            this._geolocateButton.setAttribute('aria-label', title);
         }
 
         if (this.options.trackUserLocation) {
--- a/src/ui/control/logo_control.js
+++ b/src/ui/control/logo_control.js
@@ -31,7 +31,7 @@ class LogoControl {
         anchor.target = "_blank";
         anchor.rel = "noopener nofollow";
         anchor.href = "https://www.mapbox.com/";
-        anchor.setAttribute("aria-label", "Mapbox logo");
+        anchor.setAttribute("aria-label", this._map._getUIString('LogoControl.Title'));
         anchor.setAttribute("rel", "noopener nofollow");
         this._container.appendChild(anchor);
         this._container.style.display = 'none';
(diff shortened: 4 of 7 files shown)


src/ui/map.js
```diff
@@ -215,6 +218,8 @@ const defaultOptions = {
...
  * @param {number} [options.fadeDuration=300] Controls the duration of the fade-in/fade-out animation for label collisions, in milliseconds. This setting affects all symbol layers. This setting does not affect the duration of runtime styling transitions or raster tile cross-fading.
  * @param {boolean} [options.crossSourceCollisions=true] If `true`, symbols from multiple sources can collide with each other during collision detection. If `false`, collision detection is run separately for the symbols in each source.
+ * @param {string} [options.uiLanguage='en'] A language to use in the UI. The only supported languages for now are 'en' & 'uk'. If an unsupported language is provided, 'en' will be used.
+ * @param {object} [options.uiLanguagePatch] A patch to apply to the localization table. You can replace all UI strings (thereby adding support for a new translation) or just part of them (thereby patching the default table for `uiLanguage`).
```
andrewharvey: These two could be combined. If an object is passed to uiLanguage it acts as the language string lookup, or if just a string it's the language code.

On the other hand you could say how it is now is clearer :thought_balloon: . Interested to see what other people think.

dmytro-gokun: @andrewharvey Yes, you are right having it as one of a string or an object might be an option as well. The reasoning behind the route I've taken was that it allows the following scenarios:

1) Only change the default translation for some strings. So, say i specify Swedish as a language and then only "patch" the "Exit fullcsreen" tooltip. If I had to provide the entire list of strings, that would make my code more brittle and susceptible to bugs when new controls/strings added or existing default text is changed. With this "patch" approach i can either change all the strings (thus, providing support for a completely new language/translation) or only few of them.

2) By allowing to "patch" the dictionary, it is easier to add localization support for custom controls. So, ideally, when developing a custom control, one would expose language "packs" for it (and perform the actual translation through this._map._translateUiString('') . Later, a consumer of such control could pass that language "pack" as "uiLanguagePatch" objects. Moreover, if you consume several such controls, you may merge their translation packs, while preserving translation for the default controls.

So, this was my reasoning. May be it is too complicated and a bit simpler approach has to be taken. For example, we could have a central 'localization' repository + some API to patch it. Then, custom controls will have a chance to install their supported languages.

I will wait for some more feedback.

src/ui/control/fullscreen_control.js
```diff
@@ -94,11 +94,19 @@ class FullscreenControl {
...
     _updateTitle() {
-        const title = this._isFullscreen() ? "Exit fullscreen" : "Enter fullscreen";
+        const title = this._getTitle();
         this._fullscreenButton.setAttribute("aria-label", title);
         this._fullscreenButton.title = title;
     }
 
+    _getTitle() {
+        if (this._isFullscreen()) {
+            return this._map._translateUiString("FullscreenControl.ExitTooltip") || "Exit fullscreen";
```
vakila: Ultimately I think we should follow what I understand to be the [general best practice](https://docs.microsoft.com/en-us/dotnet/standard/globalization-localization/globalization#use-resource-files) of pulling all the default (en) strings out of the code and into an `en` resource module analogous to the `uk` one you've created. Then `._translateUiString()` would become something more like `._getUiString()` and would be responsible for returning the default-or-localized version.

src/ui/localization/ui-translator.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+
+import { clone, extend, warnOnce } from '../../util/util';
+import uk from './uk';
```
vakila: As there aren't a lot of strings to deal with (yet), what about putting all the localizations in a single module, as [mapbox-gl-geocoder has done](https://github.com/mapbox/mapbox-gl-geocoder/pull/201/files#diff-76473b4e518211fdbb427d6d2e0127d3)?

Importing that one module would then give us the equivalent of the `SupportedUiLanguages` object here, and in the future we wouldn't have to touch this file to add new imports every time we add a language.

src/ui/map.js
vakila: I agree with @andrewharvey that it would be nice to keep this a single option, but I also agree with the rationale for 1&2 here. 

What if we had a single option, which always expects an object with a `language` string and optionally a `strings` object with the string mappings? Defaulting to `{ language: 'en' }`

In that case I could pass `{ language: 'uk' }` to use the built-in Ukrainian translations, without any overrides, or something like this to provide a partial or full translation patch for a given language:
```
{ language: 'en', strings: { 'FullscreenControl.ExitTooltip': 'Minimize', ... } }
```

In that case perhaps renaming the option `uiLocalization` or something would avoid the awkwardly repetitive `uiLanguage: { language: ... }`.

Thoughts?

1ec5:
> Personally I think we don't have that many localized strings to implement a full-fledged internationalization system. Instead, we could just extract the strings as control options, just like Leaflet does (e.g. [here](https://leafletjs.com/reference-1.5.0.html#control-zoom-zoomintext)). This would be much simpler, and cover most use cases.

An approach that includes built-in translations has the advantage of not requiring the developer to obtain (read: possibly pay for) translations into each of the languages they want to fully support. It avoids duplication of effort.

One disadvantage would be the potential for translations to take up more space as the number of languages and translatable strings increases. But GL JS is far from that, as the check-size build would probably show. (It isn’t running on CI because this branch lives on a fork.) Meanwhile, combining the translations into a single module could address the concern about complexity, but anyways most of the PR’s logic pertains to the override options, not the vended translations.

A _full-fledged_ internationalization system would include [gl-language](https://github.com/mapbox/mapbox-gl-language/) and also pull strings in from Transifex like the native SDKs do. 😉 This PR doesn’t go quite that far and doesn’t need to, but maybe someday it might be nice to use Transifex, so that translators can more easily keep translations in sync between the different platform SDKs.

dmytro-gokun:
@1ec5 Thank you for your feedback. Honestly, i did not update this PR because i was a bit confused about the path that folks prefer to take. Do you think it makes sense to implement what @vakila suggests here: https://github.com/mapbox/mapbox-gl-js/pull/8095#discussion_r289464255?

interDist:
@dmytro-gokun , @1ec5 , @andrewharvey  I hope this PR moves forward soon! The English-only tooltips are an annoying UI bug.

mourner:
@dmytro-gokun hey, sorry for the delay with responses and thanks a lot for taking on this! Even though localization doesn't change a lot of code, we have to take time coming up with the right direction because committing to a particular approach will have major long-term implications for maintaining the project.

Here are my thoughts personally (other team members might have different opinions):

- We're probably not ready to commit to maintaining a set of official translations for different languages as a part of this repo. But we can set up the API in a way so that it will be easy to plug in community translations some time later — for example, maintaining them in a different repo.
- Since the UI strings are not a critical part of the map experience and we don't have that many to translate, I'd strive for the most minimal and simple API / code possible, avoiding sophisticated solutions.
- The simplest API I can imagine that seeminly still has enough flexibility (e.g. supports partial overrides) and convenience (e.g. all strings in one place) is having a single object with all the UI strings as a map option — e.g. like this:

```js
  ...
  locale: {
    zoomIn: 'Приблизити',
    zoomOut: 'Віддалити'
  }
});
```

Then we'd merge those over the default English strings like we do for options in general. Then we can support the use case while not having to commit to a particular translation maintenance model right now (given limited bandwidth). Will be happy to hear your thoughts on this.

vakila:
Hi @dmytro-gokun, thanks again for this work and sorry again for our slow & at times confusing responses - as @mourner wrote we are excited about seeing this get over the finish line, and we've been discussing as a team how we can best support a localization strategy going forward.

Thanks @mourner for capturing the team discussion! I - and I believe the rest of the team - have come to agree that an API like the one above, i.e. where the map has no internal "knowledge" of locale codes or their translations, but only receives a locale object with strings to patch, will be the best path forward. At some point after we get this merged, we can tackle the longer-term question of how to manage translations going forward - noting some of the team's suggestions below for future reference. 

@dmytro-gokun and anyone else following along who'd like to weigh in - does that sound reasonable to you? If so, are you still available to work on moving this in that direction, so we can get it merged in? 

---


>  plug in community translations some time later — for example, maintaining them in a different repo.

Some of the ideas suggested so far:
- Do not maintain translations in a centralized place, and merely allow/require developers to pass in the strings they want at map creation
  - This will be the initial situation once we merge this PR, but as @1ec5 pointed out above, in the long run we'd like a better solution to avoid reduplication of effort and support translations being shared and reused in the community. Balancing that with our team's limited ability to support sourcing & maintaining translations will be the trick.
- A plugin or separate repo collecting string patch objects for various locales (e.g. `import { ru } from "mapbox-gl-ui-locales"`). We'd need to determine whether that would be maintained by Mapbox or the community.
- Separate community-maintained modules for each locale, which users are able to pick and choose from.

dmytro-gokun:
@vakila @dmytro-gokun Thanks for the update, guys.

To summarize what have be done now with this PR: 
1. Add a new optional property to Map's options named "locale". It is expected to hold a string map.
2. If "locale" is supplied, it will be used. If it is not supplied (or a particular key is missing), the default locale (hard-coded within mapbox-gl-js) will be contacted.

If everyone is okay with this, i'll go ahead and update the PR accordingly. Please let me know.

mourner:
> If it is not supplied (or a particular key is missing), the default locale (english, hardcoded within mapbox-gl-js) will be consulted.

@dmytro-gokun the supplied keys will override the default ones, like this:

```js
this._locale = extend({}, defaultLocale, options.locale);
```

dmytro-gokun:
@mourner Got it. Thanks.

dmytro-gokun:
@mourner @vakila Please check it. Is that like you wanted it to be? If it's okay, i'll go ahead and do this for the rest of control (or should that be done in separate PR(s)?)

src/ui/map.js
```diff
@@ -1168,6 +1172,10 @@ class Map extends Camera {
         }
     }
 
+    _translateUiString(key: string) {
+        return this._locale[key] || key;
```
mourner: Nit: is there a case where we want `key` to be returned if the input is supposed to be an id rather than a final string? If not, perhaps we should rename to `_getUIString`, and throw an error if `this._locale[key]` is not defined.

src/ui/map.js
```diff
@@ -44,6 +44,7 @@ import type DragPanHandler, {DragPanOptions} from './handler/drag_pan';
 import type KeyboardHandler from './handler/keyboard';
 import type DoubleClickZoomHandler from './handler/dblclick_zoom';
 import type TouchZoomRotateHandler from './handler/touch_zoom_rotate';
+import defaultLocale from './localization/default_locale';
```
mourner: Nit: do we need to introduce a new folder if it will only contain a single file? Maybe worth moving to just `src/ui/default_locale.js`.

src/ui/localization/default_locale.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+const defaultLocale = {
+    'FullscreenControl.EnterTooltip': 'Enter fullscreen',
+    'FullscreenControl.ExitTooltip': 'Exit fullscreen'
```
mourner: Also wondering whether we need long namespaced ids if there are not that many strings to define, but it will be clearer when we define them for all controls.

src/ui/map.js
```diff
@@ -235,7 +237,7 @@ const defaultOptions = {
  * @param {number} [options.fadeDuration=300] Controls the duration of the fade-in/fade-out animation for label collisions, in milliseconds. This setting affects all symbol layers. This setting does not affect the duration of runtime styling transitions or raster tile cross-fading.
  * @param {boolean} [options.crossSourceCollisions=true] If `true`, symbols from multiple sources can collide with each other during collision detection. If `false`, collision detection is run separately for the symbols in each source.
  * @param {string} [options.accessToken=null] If specified, map will use this token instead of the one defined in mapboxgl.accessToken.
-
+ * @param {string} [options.locale=null] A patch to apply to the default localization table. You can replace all UI strings (thereby adding support for a new translation) or just part of them (thereby patching the default translation table).
```
vakila: This will take an `Object` value, right? Do we need to be more specific about the type of object and define (elsewhere?) what a `Locale` object is shaped like (e.g. which keys are valid)? 

In any case I think the docs entry here could be a bit more specific/descriptive, here's my attempt to flesh it out a bit.

```suggestion
 * @param {Object} [options.locale=null] A patch to apply to the default localization table for UI strings, e.g. control tooltips. The `locale` object maps namespaced UI string IDs to translated strings in the target language; see `src/ui/default_locale.js` for an example with all supported string IDs. The object may specify all UI strings (thereby adding support for a new translation) or only a subset of strings (thereby patching the default translation table).
```

src/ui/map.js
```diff
@@ -374,6 +377,7 @@ class Map extends Camera {
         this._renderTaskQueue = new TaskQueue();
         this._controls = [];
         this._mapId = uniqueId();
+        this._locale = extend({}, defaultLocale, options.locale);
```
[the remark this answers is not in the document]
dmytro-gokun: I'd say it is not needed. Having unrecognized keys here is okay, as the same object might be used to keep translations for custom controls. So, the core library may not be aware of all possible keys here.

dmytro-gokun:
Thanks, @mourner! I've now fixed lint errors, but it appear that 'test-render' fails:

```
failed text-variable-anchor/pitched-with-map
```

I'm not if this might have been caused by this commit or if it's unrelated..

mourner:
@dmytro-gokun it's unrelated, just fails sometimes (see #8679). I've restarted the test, should be green soon.

dmytro-gokun:
Thanks, guys

PR #8113, opened 2019-04-03
Include source name in invalid GeoJSON error

Zirak:
When supplying invalid data to a GeoJSON source, include the source name in the
error message for easier debugging.

Before PR:
Error and stack trace without a source name, before the PR
As the stacktrace is heavily async, we don't get an immediate indication of where in userland code the error originated. It can be any one of the user-defined sources. The only way to figure which source is the erroneous one, one needs to put a breakpoint on the line logging the error (as it's not an exception), and search up the async stack trace. A bit cumerbose for a bug that's usually "whoops I gave the wrong object".

After PR:
Error and stack trace showing the source name, after the PR
Nothing but sunshine and kittens, as we can see the source name being supplied invalid GeoJSON we can search our code, and only if needed debug up the stack trace.

This PR unfortunately only covers *some* of the invalid GeoJSON errors, there is another one in a mapbox dependency ([geojson-vt](https://github.com/mapbox/geojson-vt)), which of course does not know about sources and layers and all that fun. So the following call:

```js
map.addSource('awesomesauce', { type: 'bazinga' });
```

Will still show the less fun errors. As there are many ways to solve that error (daisy-chaning an error message defined in the caller with the original error, change the error to a null return value and check that, ...), how would you think it best be done?

Thanks.

## Launch Checklist

 - [ ] write tests for all new functionality (n/a? no existing tests for this error AFAICT)

### DIFF
--- a/src/source/geojson_worker_source.js
+++ b/src/source/geojson_worker_source.js
@@ -167,7 +167,7 @@ class GeoJSONWorkerSource extends VectorTileWorkerSource {
             if (err || !data) {
                 return callback(err);
             } else if (typeof data !== 'object') {
-                return callback(new Error("Input data is not a valid GeoJSON object."));
+                return callback(new Error(`Input data given to '${params.source}' is not a valid GeoJSON object.`));
             } else {
                 rewind(data, true);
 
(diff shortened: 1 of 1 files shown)


mourner:
Thanks!

PR #8118, opened 2019-04-04
Fix blurry icon rendering when icon-pitch-alignment=map

### LINKED ISSUE #8110: Symbol interpolation issue when icon-pitch-alignment is "map"
**mapbox-gl-js version**: v0.54.0-beta.1

**browser**: any

### Steps to Trigger Behavior

 1.  create a symbol layer with icon-pitch-alignment set to "map"
 2. do not rotate or pitch the map

### Expected Behavior
Icon should be rendered with NEAREST interpolation

### Actual Behavior
Icon is rendered with LINEAR interpolation

Hint: there is a condition in draw_symbol.js:197
```js
pitchWithMap || tr.pitch !== 0
```
I think it should be just
```js
tr.pitch !== 0
```
because when  icon-pitch-alignment is set to "map", and neither the icon nor the map is rotated, NEAREST should be used. (same as "viewport" alignment)

brncsk:
## Launch Checklist

Fixes #8110 and updates the corresponding render tests.

 - [ ] post benchmark scores


ryanhamley:
The `symbol-spacing/line-close`, `symbol-spacing/line-far`, and `symbol-spacing/line-overscaled` render tests are failing on CI

brncsk:
@ryanhamley They all pass for me locally. Do the render tests depend on the graphics card being used? If so, shall I replace the expected results with those from the CI output?

brncsk:
@ryanhamley 
- I updated the expected results of the three failing render tests to match the actual results generated on CI.
- I also increased the allowed difference for these tests to `0.0005` to make them pass both locally and on CI.

This makes the build green.

chloekraw:
cc: @asheemmamoowala

mourner:
@ansis can you take a look? I think the original logic goes back to https://github.com/mapbox/mapbox-gl-js/commit/79c897f8f6e54dffbad9c419c879cb97c1cf4e46

ansis:
@brncsk thanks for looking into this!

You are right that `pitchWithMap` shouldn't be a condition here but I think we might need to add a couple new conditions for when to use linear. The render tests with the current changes are visibly more pixelated than before.

This bug was introduced in https://github.com/mapbox/mapbox-gl-js/commit/73f102ad510406f09717a2a5f931c422e2ee61b6 where the condition changed from `alignedWithMap` (`icon-rotation-alignment`) to `pitchWithMap` (`icon-pitch-alignment`).

I think linear rendering might need to be enabled depending on the value of `icon-rotation-alignment` and `icon-rotate`. I can make these changes if you want.

ahk:
Hi @brncsk, we were waiting on feedback about this comment since May 7: https://github.com/mapbox/mapbox-gl-js/pull/8118#issuecomment-490219424

We should have checked back since then to see if we could resolve the concerns there, but we didn't and in the meantime we are working on this: https://github.com/mapbox/mapbox-gl-js/issues/8730

If we move to always use `GL_LINEAR` interpolation during icon rendering as in the linked issue, this logic to determine if symbols should be interpolated as `GL_NEAREST` or `GL_LINEAR` will no longer apply.

Because of that I don't think your bugfix (which was good!) will be necessary anymore. I'm going to close this ticket and its associated pull request, but we really appreciate your help and hope the new change we're landing soon is a strict improvement for you.

Please let me know if I've missed something here, and apologies for not getting this fixed up and pushed through sooner.

brncsk:
@ahk Thanks for the heads up!

PR #8119, opened 2019-04-04
Fix location dot shadow

bravecow:
Location dot shadow was applied to blue `div` with 2px width. It was covered by 2px width `::after` pseudo element.

Before:

image

After:

image

### DIFF
--- a/src/css/mapbox-gl.css
+++ b/src/css/mapbox-gl.css
@@ -9,7 +9,7 @@
     width: 100%;
     height: 100%;
 }
-
+.mapboxgl-user-location-dot::after
 .mapboxgl-canary {
     background-color: salmon;
 }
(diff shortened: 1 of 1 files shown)


andrewharvey:
Looks good to me. Thanks @bravecow.

I think while we're at it, we should make that shadow more visible, because to me 2px, 0.25 opacity is not distinguishable from no shadow. I think at least 8px with 0.5 opacity. What do you think?

Screenshot from 2019-04-05 11-17-42

bravecow:
Large shadow makes it more "dirty". I suggest `0 0 3px rgba(0,0,0,.35)`

Slightly increasing opacity and 1px more shadow:

image

mourner:
@bravecow agreed — let's update to this one.

PR #8122, opened 2019-04-04
[] add a cache to style-layer-index so we don't recalculate keys

vallendm:
## Launch Checklist

Test case: 33000 post codes are displayed on a map in a single fill layer. The tileset that its based upon contains postcodes for other countries so to reduce memory usage - a filter is applied on the layer so that only the 33000 postcodes that we're interested in are displayed. 
Next we add a second layer for region outlines. When we mouseover a region - we set this "highlight" layer to show the single item. 
Issue is that the speed of updating the "highlight" layer is slow. Profiling has shown this is due to the filter applied on the original region layer. When the "highlight" layer is changed - a broadcast event is sent to each tileworker - with the updated style. The tileworker then updates its styles based on the broadcast - and calculates a key for each layer. The key includes the filter - so to make the key it requires iterating over the 33000 items.

However this layer isn't one which is changing. In this case it seem a waste to have to regenerate the key.So add a key cache to the style_layer_index. When the key is calculated - save it in the cache. For updated layers - remove the key from the cache so we get a new key generated. For layers which stay the same - just use the existing key from the cache. 

With this change - the testcase average time to display a hightlight was 381.03ms - whereas without the change the average highlight time is 830.58

 - [x] post benchmark scores

### DIFF
--- a/src/style-spec/group_by_layout.js
+++ b/src/style-spec/group_by_layout.js
@@ -53,11 +53,11 @@ function groupByLayout(layers, cachedKeys) {
 
     for (let i = 0; i < layers.length; i++) {
 
-        const k = (cachedKeys && cachedKeys[layers[i].id] ) || getKey(layers[i]);
+        const k = (cachedKeys && cachedKeys[layers[i].id]) || getKey(layers[i]);
         // update the cache if there is one
-        if ( cachedKeys )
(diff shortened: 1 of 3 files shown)


vallendm:
Sorry about that - I'm used to a different bracket/space pattern. Fixed now.

Issue #7425, opened 2018-10-16
Add getElement method to Popup (like Marker)


cantacell:
Adding a getElement method (like Marker or Leaflet Popup) on Popup gives the possibility to change dynamically the css className.
Regards,
Francesco

DeusVultSVK:
CAN I work on this issue?

ryanhamley:
@DeusVultSVK yep, see our contributor's guide https://github.com/mapbox/mapbox-gl-js/blob/master/CONTRIBUTING.md

mollymerp:
@DeusVultSVK Thank you! Let us know if you have questions or need help implementing this feature.

andrewharvey:
This was done in #8123

shubhuhidev:
Popup.getElement() method giving undefined

RicardoTactiql:
@shubhuhidev If the marker is closed it gives undefined. 

Is there a solution to that?

PR #8123, opened 2019-04-04
Add Popup.getElement method (#7425)

bravecow:
Implements #7425 

Added test. Previous tests instead of `_container` use new `.getElement`. (Like marker test: https://github.com/mapbox/mapbox-gl-js/blob/master/test/unit/ui/marker.test.js#L26 and below)


mourner:
Thanks!

PR #8130, opened 2019-04-08
update one more frame after canvas source paused

ansis:
Calling `canvasSource.pause()` lets us know we can stop pulling in updates from it. Since it is possible that changes were made to the canvas source since the last render we should pull in one more update
before stopping. This also lets you call play and pause immediately to render just one frame:

```
ctx.fillRect(10, 10, 50, 50);
canvasSource.play();
canvasSource.pause();
```

## Launch Checklist

 - [N/A] post benchmark scores


ansis:
> did you accidentally forget to commit it?

Yep, 6f8feba adds that.

> As a developer I would expect that calling CanvasSource#pause allows me to make updates to the underlying canvas without updating the rendered map. 

95dc749 fixes that

test/suite_implementation.js
```diff
@@ -157,6 +157,16 @@ module.exports = function(style, options, _callback) { // eslint-disable-line im
             map.addLayer(new customLayerImplementations[operation[1]](), operation[2]);
             map._render();
             applyOperations(map, operations.slice(1), callback);
+        } else if (operation[0] === 'updateFakeCanvas') {
+            const canvasSource = map.getSource(operation[1]);
+            canvasSource.play();
+            // update before pause should be rendered
+            updateFakeCanvas(window.document, options.addFakeCanvas.id, operation[2]);
```
asheemmamoowala: This is putting the logic to be tested in the suite implementation instead of in a specific test. It is not a good precedent, but there is probably not a better way to do this with the current render test framework. Maybe an additional note here tying it back to a specific ticket or render-test would make it easier to remember why it is done this way.

ansis: Yeah, I agree that expanding the suite implementation further is not great but I'm not seeing a better way. I'm hesitant to add a note pointing to a specific test because notes tend to get outdated. I think searching for "updateFakeCanvas" within tests should be good enough.

PR #8131, opened 2019-04-08
add `animationFrameProvider` map option

### LINKED ISSUE #7893: Synchronous redraw API
## Motivation

In react-map-gl, when the viewport updates, we call `Map.jumpTo()` inside `componentDidUpdate()`, which schedules a Mapbox rerender in the next animation frame. This causes the canvas rerender to always occur one step behind the React updates.

You can see this issue in https://github.com/uber/react-map-gl/pull/720 and https://github.com/uber/deck.gl/issues/2458.

## Design

Provide a synchronous API e.g. `redraw()` that immediately flushes the dirty state.

### Implementation

Here's how we are currently forcing rerender:

```js
// map render will throw error if style is not loaded
if (map.isStyleLoaded()) {
  map._render();
  // cancel the scheduled update
  if (map._frame) {
    map._frame.cancel();
    map._frame = null;
  }
}
```

Private methods and properties are manipulated here, which is not a reliable solution.

ansis:
This option allows the user to implement `requestAnimationFrame` and `cancelAnimationFrame` in order to synchronize map rendering with their external rendering. While it is generally recommended to let the browser schedule map rendering frames this may be useful for synchronizing with external content in some cases (fix #7893).

This could also be useful for synchronizing multiple maps.

```js

const animationFrameProvider = {
    requestAnimationFrame: fn => { ... },
    cancelAnimationFrame: handle => { ... }
};
const map = new Map({
    animationFrameProvider: animationFrameProvider,
    ...
});
```

### As an alternative to synchronous render frames

https://github.com/mapbox/mapbox-gl-js/issues/7893 explains the reasons why synchronous rendering is necessary in some cases. An `animationFrameProvider` option seemed preferable to implementing a `map.synchronousRepaint()` because:
- it prevents all renders that aren't directly triggered by use (instead of just allowing renders to be triggered by you)
- it could be harder to abuse

```js
class SynchronizedFrameProvider {
    constructor() {
        this._requests = [];
        this._nextHandle = 0;
    }

    requestAnimationFrame(fn) {
        this._requests.push({
            handle: this._nextHandle++,
            fn
        });
    }

    cancelAnimationFrame(handle) {
        this._requests = this._requests.filter(req => req.handle !== handle);
    }

    // called by you to trigger synchronous rendering
    dispatchFrame() {
        // clear array before dispatching to handle case where user calls
        // `requestAnimationFrame` from within the dispatched function.
        const requests = this._requests;
        this._requests = [];

        const now = performance.now();
        for (const req of requests) {
            req.fn(now);
        }
    }
}

const provider = new SynchronizedFrameProvider();

const map = new Map({
    animationFrameProvider: provider
    ...
});

...

provider.dispatchFrame();
```

## Launch Checklist

 - [ ] post benchmark scores

@mourner @ryanhamley does this approach seem preferable? Or would it be better just implement a `map.synchronousRender()`?

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -35,6 +35,7 @@ import type {StyleOptions, StyleSetterOptions} from '../style/style';
 import type {MapEvent, MapDataEvent} from './events';
 import type {CustomLayerInterface} from '../style/style_layer/custom_style_layer';
 import type {StyleImageInterface} from '../style/style_image';
+import type {AnimationFrameProvider} from '../util/browser';
 
 import type ScrollZoomHandler from './handler/scroll_zoom';
 import type BoxZoomHandler from './handler/box_zoom';
@@ -95,7 +96,8 @@ type MapOptions = {
     pitch?: number,
     renderWorldCopies?: boolean,
     maxTileCacheSize?: number,
-    transformRequest?: RequestTransformFunction
+    transformRequest?: RequestTransformFunction,
+    animationFrameProvider?: AnimationFrameProvider
 };
 
 const defaultMinZoom = 0;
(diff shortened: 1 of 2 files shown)


jahow:
Hi!

The OpenLayers community is looking forward to this improvement as it will allow keeping an OpenLayers map and Mapbox in sync, thus giving people the best of both worlds! There is currently a working example but it uses an undocumented API, as was described in https://github.com/mapbox/mapbox-gl-js/issues/7893#issue-408992184

Please let me know if there is anything we can do to help, inc. contribution proposal if necessary. Thanks a lot

mourner:
It appears that there's no consensus on whether we should commit to this API — it [doesn't solve](https://github.com/mapbox/mapbox-gl-js/issues/7893#issuecomment-482243898) the original issue that prompted it according to the author, and there are concerns over its ergonomics. Let's close the PR for now and re-evaluate the problem.

PR #8134, opened 2019-04-08
Fixes bugs in findStopLessThanOrEqualTo algorithm

Mike96Angelo:
## Old behaviour
- When the input > all stops it returned the second-last stop, should have been the last stop.
```
findStopLessThanOrEqualTo([0, 1, 2, 3, 4, 5, 6, 7], 8) // 6 but should have been 7
```
- When more than one stop had the same value it didn't always return the last stop
```
findStopLessThanOrEqualTo([0.5, 0.5], 0.5) // 0 but should have been 1
```

## New behaviour
- When the input > all stops it returns the last stop.
```
findStopLessThanOrEqualTo([0, 1, 2, 3, 4, 5, 6, 7], 8) // 7
```
- When more than one stop has the same value it always returns the last stop
```
findStopLessThanOrEqualTo([0.5, 0.5], 0.5) // 1
```

## Launch Checklist

 - [ ] post benchmark scores


src/style-spec/expression/stops.js
```diff
@@ -11,19 +11,22 @@ export type Stops = Array<[number, Expression]>;
  * @private
  */
 export function findStopLessThanOrEqualTo(stops: Array<number>, input: number) {
-    const n = stops.length;
+    const n = stops.length - 1;
     let lowerIndex = 0;
-    let upperIndex = n - 1;
+    let upperIndex = n;
```
mourner: Nit: let's rename `n` to `lastIndex` if we change its meaning.

src/style-spec/expression/stops.js
```diff
@@ -11,19 +11,22 @@ export type Stops = Array<[number, Expression]>;
...
-        upperValue = stops[currentIndex + 1];
-        if (input === currentValue || input > currentValue && input < upperValue) { // Search complete
+        nextValue = stops[currentIndex + 1];
+
+        if (currentValue <= input && (currentIndex === n || input < nextValue)) { // Search complete
             return currentIndex;
-        } else if (currentValue < input) {
+        }
+
+        if (currentValue <= input) {
```
mourner: Nit: can we put the "search complete" condition inside this `if` so that we don't check `currentValue <= input` twice?

Mike96Angelo:
@mourner have you been able to look at the changes I've made?

PR #8135, opened 2019-04-09
Studio supports rtl

samanpwbb:
Tiny documentation fix. Studio has RTL plugin support built-in 🎉 

@colleenmcginnis for review plz 🙏

### DIFF
--- a/docs/pages/example/mapbox-gl-rtl-text.js
+++ b/docs/pages/example/mapbox-gl-rtl-text.js
@@ -2,8 +2,8 @@
 title: Add support for right-to-left scripts
 description: >-
   Use the [mapbox-gl-rtl-text](https://github.com/mapbox/mapbox-gl-rtl-text)
-  plugin to support scripts that use right-to-left layout (such as Arabic or
-  Hebrew). Note that Mapbox Studio does not yet support RTL text.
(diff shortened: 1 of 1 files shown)


docs/pages/example/mapbox-gl-rtl-text.js
```diff
@@ -3,7 +3,7 @@ title: Add support for right-to-left scripts
 description: >-
   Use the [mapbox-gl-rtl-text](https://github.com/mapbox/mapbox-gl-rtl-text)
   plugin to support scripts that use right-to-left layout (such as Arabic or
-  Hebrew). Note that Mapbox Studio does not yet support RTL text.
+  Hebrew).
```
1ec5: By the way, what’s left of this paragraph can be misleading. Arabic and Hebrew are the only right-to-left scripts that gl-language reliably supports. Moreover, Arabic is the only Arabic-script language it supports.

/ref https://github.com/mapbox/documentation/issues/156#issuecomment-477664649

samanpwbb: That's really interesting @1ec5, but I'd prefer to leave this change for a separate content update – sounds like we need to workshop the precise language we want to use when describing RTL and apply it across our docs.

PR #8173, opened 2019-04-22
clamp rather than exclude south pole features

peterqliu:
## Launch Checklist


mourner:
I don't think this is right. `(x < 0 || x >= EXTENT || y < 0 || y > EXTENT)` condition makes sure that a point that is present on the boundary (and thus also included in adjacent tile) doesn't get rendered twice. 

This should be reverted and solved in some other way (also, there seem to be no context for the PR — is there a specific issue it addresses?).

mourner:
You can see the problem in the render test that was updated to a wrong rendering — it's a single point nowhere near the pole, and it gets double-rendered after this PR.

peterqliu:
@mourner thanks for the catch! the context was from a customer in a private repository, who noticed dropped circles in points of latitude < 85.05. The original PR clamps it down to that value, but can totally see why it'd render bottoms of every tile doubly, when there's a tile below it.

Will have to determine whether the tile is on the bottom row for its given zoom level, and do a special rendering case for those

PR #8195, opened 2019-04-25
add a cache of serialized style layers

vallendm:
## Launch Checklist

QueryRenderedFeatures has poor performance when returning lots of features. This appears to be caused by having to serialize the layer into the response. Rather than serializing the layer once - its is serialized N times if there are N features returned. This isn't normally a problem - but for a layer which has a large color expression it can take alot of time. 

Rather than serialize it may times - just keep an optional cache so we only serialize each layer once. 

 - [ ] post benchmark scores

### DIFF
--- a/src/data/feature_index.js
+++ b/src/data/feature_index.js
@@ -128,6 +128,7 @@ class FeatureIndex {
         matching.sort(topDownFeatureComparator);
 
         const result = {};
+        const layerCache = {};
         let previousIndex;
         for (let k = 0; k < matching.length; k++) {
             const index = matching[k];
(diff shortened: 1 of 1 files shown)


mourner:
Hi @vallendm, apologies for such a long delay looking into this PR. 

This appears to be a more difficult matter than I anticipated — digging into 5 year commit history around this particular feature (query results including serialized `layer`), I discovered that the original intent was to return _evaluated_ layout/paint values, not original definitions — see #879. At some point during many refactorings, this changed without anyone noticing.

We could apply a temporary fix to mitigate performance issues, but an ideal solution would be to track down where that change happened, and see if we can avoid full serialization by reverting back to outputting final values of properties evaluated at the time of the query. I'll dig into this further.

mourner:
I've created a new issue for this #8373 — let's track it there.

PR #8200, opened 2019-04-25
Add contains method to LngLatBounds

### LINKED ISSUE #7512: Add LatLngBounds.contains() method
## Motivation

I'm splitting this request out from https://github.com/mapbox/mapbox-gl-js/issues/7090 in which several different `LatLng` and `LatLngBounds` methods were brought up. GL Native has a [`LatLngBounds.contains`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/geo.cpp#L37) utility method which is missing in JS. Adding this method to JS would increase parity between GL implementations and provide a useful method to users.

## Design

I can see two possibilities.

1) A static method on LatLngBounds
```
LatLngBounds.contains(bounds, point)
```

2) An instance method
```
const bounds = new LatLngBounds([sw, ne]);
bounds.contains(point);
```

I'd say the instance method is preferable since all other methods on `LatLngBounds` are also instance methods. The static method would mostly be useful as a pure utility method in which you may want to check arbitrary bounds for a point, but does that have many use cases? I'm unsure.

`contains` should be able to check that a bounding box contains a [point](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/geo.cpp#L41) or a [CanonicalTileID](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/geo.cpp#L37), as the Native implementation does. Native also checks [unwrapped bounds](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/geo.cpp#L68), but that concept isn't well defined in JS and I think we can side step that implementation for now.

### Implementation

We should be able to just port the C++ implementation
https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/tile_range.hpp#L40

mzdraper:
I've been working with @ryanhamley to add a `contains` method on `LngLatBounds` that checks to see if a point is within a bounding box. Would love some eyes on how we're handling wrapped coordinates.

For a quickstart, you can add the following to a debug page to test it locally.

```js
var bounds = [[190, -10], [170, 10]];
var bounds = [[190, -10], [170, 10]];
var llb = new mapboxgl.LngLatBounds(bounds[0], bounds[1]);
map.fitBounds(bounds);
var containsPoint = llb.contains([180, 0]); // true
```

Closes #7512.

### DIFF
--- a/src/geo/lng_lat_bounds.js
+++ b/src/geo/lng_lat_bounds.js
@@ -1,8 +1,6 @@
 // @flow
 
 import LngLat from './lng_lat';
-import Point from '@mapbox/point-geometry';
-import Transform from './transform';
 
 import type {LngLatLike} from './lng_lat';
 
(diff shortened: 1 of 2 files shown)


ryanhamley:
Would close https://github.com/mapbox/mapbox-gl-js/issues/7512

mzdraper:
This will bring GL JS into parity with this [GL Native code](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/geo.cpp#L37-L65) for `LatLngBounds.contains(point)`. Though the GL Native implements `contains` for `LatLngBounds` and `CanonicalTileID`, this PR does not because `LatLngBounds` can be implemented with Turf.js and the use case for `CanonicalTileID` isn't clear.

@ansis can you check the wrapped coordinates logic? Let me know if you think there are uses for `LatLngBounds` and `CanonicalTileID` to be included in GL JS too!

src/geo/lng_lat_bounds.js
```diff
@@ -209,6 +209,25 @@ class LngLatBounds {
         return !(this._sw && this._ne);
     }
 
+    /**
+    * Check if the point is within the bounding box.
+    *
+    * @returns {boolean} True if the point is within the bounding box.
+    */
+    contains(lnglat: LngLatLike) {
+        lnglat = lnglat instanceof LngLat ? new LngLat(lnglat.lng, lnglat.lat) : LngLat.convert(lnglat);
```
mourner: `LngLat` case is [already covered](https://github.com/mapbox/mapbox-gl-js/blob/master/src/geo/lng_lat.js#L108-L110) in `convert`.

```suggestion
        lnglat = LngLat.convert(lnglat);
```

PR #8221, opened 2019-05-06
8079 followup

tristen:


tristen:
I think I'm seeing predictable results now. 

Screen Shot 2019-05-06 at 3 26 13 PM

Issue #3405, opened 2016-10-18
add mobile gesture handling for pitch 


incanus:
We need gesture support on mobile web to allow the map to be pitched. The way this works gesturally on native mobile is two fingers swiped up or down, both in parallel with each other and approximately the same y-value. 

Without this, there isn't a way to pitch on mobile without an external control button or similar that toggles between pitch values a là `View elevation profile` in https://www.mapbox.com/blog/dc-bikeshare-revisited/. 

Related: https://github.com/mapbox/mapbox-gl-js/issues/2449

1ec5:
This is a special case of #2618. As far as I can tell, gesture event support exists in mobile browsers and Internet Explorer, Edge, and Safari on the desktop. Blink and Gecko appear to lack gesture event support, but given the importance of mobile Web browsing, I don’t think that should matter.

ansarikhurshid786:
Any update or work around?

bFlood:
bump - Any update or work around?

ansarikhurshid786:
@bFlood Good job. it's working as expected.

cs09g:
beyond @bFlood 's solution would be to check distance between two points. not only it's y-distance.
@bFlood 's solution disables rotation when two fingers start rotating from horizontal positions.
```
...
var diff = Math.abs(data.points[0].y - data.points[1].y);
if (diff <= 50) {
...
```
to
```
...
const diffY = data.points[0].y - data.points[1].y;
const diffX = data.points[0].x - data.points[1].x;
const distance = Math.sqrt(Math.pow(diffX, 2) + Math.pow(diffY, 2)); alert(distance);
if (distance <= 120) {
...
```

fishead:
@cs09g Good job. How about remove the `alert(distance);` ?
```javascript
const distance = Math.sqrt(Math.pow(diffX, 2) + Math.pow(diffY, 2)); alert(distance);
```

fishead:
yet another solution base on @bFlood one.

```javascript
function addMobilePitchGesture(mbMap) {
  const minDiffX = 70; // min x distance to recognize pitch gesture
  const maxDiffY = 100; // max y distance to recognize pitch gesture
  const minDiff = 30; // min distance to recognize zoom gesture
  const delay = 160; // delay for pitch, in case it's a zoom gesture

  let dpPoint;
  let dpPitch;
  let startTiming;
  let startDistance;
  let startEventData;

  mbMap
    .on("touchstart", data => {
[...]

Issue #8208, opened 2019-04-29
Tilting compass as map tilted


cs09g:
## Motivation
When you rotate a map, compass on the navigation control rotates as much as the degree map rotated. but when you tilt the map, the compass doesn't interact for the pitch.

Also, when you click the compass, it makes the map degree to 0 as its default value but pitch would never be back to default even when you click the compass again.

I think when it interacts with both its bearing and pitch, will give better UX.
Implementation would look like(drag mouse on the MAP labeled element, and see compass below. and click the compass):
https://jsfiddle.net/cs09g/5an14p6r/

mourner:
That's a cool idea! Want to make a PR with the change?

cs09g:
Sure, one more discussion left about compass button clicking. One options is to reset bearing and pitch at once, the other is two-step reset-first click: reset bearing, second click(when bearing is 0): reset pitch. Which option would be better?

cs09g:
@mourner Could you share your idea for the options?

mourner:
I'm not sure, both seem good... @samanpwbb what do you think?

andrewharvey:
I think one click to reset both is simpler.

ryanhamley:
This is basically closed by https://github.com/mapbox/mapbox-gl-js/pull/8226 The only thing not implemented in that PR is reseting the pitch when the compass button is clicked

samanpwbb:
I checked Apple and Google maps for precedent and looks like _neither_ use the compass to control pitch. The compass's sole purpose is to communicate the cardinal directions and to allow user to reset direction. Pitching the compass makes directionality less clear. 

Therefore, I am actually not sure a rotating compass control is a good idea. At the very least, this should be an optional feature.

If we do implement it, one click to reset would likely match user expectations most closely (double-click interactions like this are rare on the web).

ryanhamley:
#8226 does implement pitching on the compass control but we could make the functionality optional in an additional PR that implements the on-click pitch reset.

pakastin:
What would you suggest for the name of the pitching setting?

pakastin:
Hmm.. but it already controls the pitch by default, so wouldn't make much sense 😂

Maybe `visualizePitch`?

ryanhamley:
@samanpwbb You're saying that the compass icon shouldn't move at all by default right? That both pitch and bearing visualization should be optional.

pakastin:
[...]

PR #8226, opened 2019-05-08
Add touch support for NavigationControl's compass button + add pitch indication to icon

pakastin:
Related to #3405

NavigationControl's compass button was missing touch drag support. Added that.

You can try the functionality here: https://aviamaps.com/map

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -83,6 +83,7 @@ class NavigationControl {
             this._rotateCompassArrow();
             this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass});
             DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
+            DOM.addEventListener(this._compass, 'touchstart', this._handler.onMouseDown);
             this._handler.enable();
         }
         return this._container;
@@ -96,6 +97,7 @@ class NavigationControl {
         if (this.options.showCompass) {
             this._map.off('rotate', this._rotateCompassArrow);
             DOM.removeEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
+            DOM.removeEventListener(this._compass, 'touchstart', this._handler.onMouseDown);
             this._handler.disable();
             delete this._handler;
         }
--- a/src/ui/handler/drag_rotate.js
+++ b/src/ui/handler/drag_rotate.js
@@ -149,8 +149,9 @@ class DragRotateHandler {
         // to prevent map move events from being fired during a drag.
         const moveEvent = touchEvent ? 'touchmove' : 'mousemove';
         const upEvent = touchEvent ? 'touchend' : 'mouseup';
+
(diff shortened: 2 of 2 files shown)


pakastin:
Tapping broke, will fix that now

src/ui/handler/drag_rotate.js
```diff
@@ -258,6 +270,8 @@ class DragRotateHandler {
     _unbind() {
         window.document.removeEventListener('mousemove', this._onMouseMove, {capture: true});
         window.document.removeEventListener('mouseup', this._onMouseUp);
+        window.document.removeEventListener('touchmove', this._onMouseMove, { capture: true });
```
cs09g: should be followed the coding convention `{ capture: true }` to  `{capture: true}` as line no.271

pakastin:
Tapping bug now fixed, so everything is working as expected!

src/ui/handler/drag_rotate.js
pakastin: Didn't change coding convention it already had..

pakastin:
Linting issues now fixed!

src/ui/handler/drag_rotate.js
cs09g: I see, there's already used in mixed way.

pakastin: However I fixed that ;)

pakastin:
Bonus feature: compass icon now pitches when pitch is changed

pakastin:
> The second feature is also discussing on #8208. I think one PR should be covered one feature.
 
that was the plan, but accidentally came into this PR.. 😕

cs09g:
How could you skip the issue already registered. I'd like you to divide this PR for each feature.

andrewharvey:
> How could you skip the issue already registered. I'd like you to divide this PR for each feature.

@pakastin I think it's okay for both these features to be done in this single PR given they are related, and the touch support is fairly minor. I don't think it's worth the overhead of effort separating them, unless you want to, or unless the maintainers ask...

pakastin:
Ok, perfect! They're both basically patches IMO..

src/ui/control/navigation_control.js
```diff
@@ -68,7 +68,7 @@ class NavigationControl {
     }
 
     _rotateCompassArrow() {
-        const rotate = `rotate(${this._map.transform.angle * (180 / Math.PI)}deg)`;
+        const rotate = `rotateX(${this._map.transform._pitch * (180 / Math.PI)}deg) rotateZ(${this._map.transform.angle * (180 / Math.PI)}deg)`;
```
asheemmamoowala: It would be better to use the `transform#pitch` property instead of the internal variable.

pakastin: Ok, fixed now! 👍

pakastin:
Any news on merging this?

pakastin:
Good, I can update to v1.0.0 then 😎

PR #8227, opened 2019-05-08
Update mapbox-gl.css

mc100s:
When the map is inside a tag with `text-align: center` (or `right`), the map is not displayed properly. An easy fix is to add `text-align: left`.

Screenshot with `text-align: center`: https://i.imgur.com/t1XU5uw.png


mourner:
Thank you!

Issue #8192, opened 2019-04-25
tileSize in source option can not overwrite tilejson


jingsam:
Suppose I want to add a raster source through a tilejson url:
```
map.addSource('a', {
  type: 'raster',
  url: 'http://a.com/tilejson'
})
```

However, this tilejson is not configured correctly. The actual tileSize is 512, but in tilejson is configured as 256. So I overwrite the tileSize like this:
```
map.addSource('a', {
  type: 'raster',
  url: 'http://a.com/tilejson',
  tileSize: 512
})
```
The expected behavior is that I can load the tile service as 512 tileSize.

However, the code tells me I can't:

https://github.com/mapbox/mapbox-gl-js/blob/88a0526833d9417fc4d5090ddfff790a59659618/src/source/raster_tile_source.js#L58-L69

As the code shows that tilejson always loads after Source initialization. So if tilejson have `tileSize` key, it will always overwrite source option.

I think `scheme` option will face same issue.

mourner:
Yeah, I think it would be reasonable for explicit properties to take precedence over TileJSON, although it's arguable. Want to explore how a PR could look for the fix?

jingsam:
@mourner I have send a PR #8232 to enable this feature.

htrex:
I'm not using tilejson but seeing an issue with the 512 tileSize on a TMS source, please let me understand if I'm missing something or there's a more general bug in mapbox gl and the tileSize 512 setting is always ignored.

I've configured a source just like in following example and mapbox is trying to load the tiles with X and Y coordinates which are exactly double of what is on the file system despite the "tileSize": 512 setting.
```
    "drone_image_512": {
      "type": "raster",
      "tiles": [
        "https://example.com/maps/drone_image_512/{z}/{x}/{y}.png"
      ],
      "tileSize": 512,
      "minzoom": 14,
      "maxzoom": 23,
      "bounds": [13.716698756, 42.360647913, 13.736451649, 42.369613033],
      "scheme": "tms"
    }
```
I've received the same raster image also exported with 256 tilesize, in fact the X and Y coordinates double on the file system and it loads perfectly with the following style source:

```
    "drone_image_256": {
      "type": "raster",
      "tiles": [
        "https://example.com/maps/drone_image_256/{z}/{x}/{y}.png"
      ],
      "tileSize": 256,
      "minzoom": 14,
      "maxzoom": 23,
[...]

PR #8232, opened 2019-05-09
Explicit source options overwrite

jingsam:
Explicit source options should taken precedence over TileJSON. See details in #8192

# Launch Checklist

 - [ ] post benchmark scores


test/unit/style-spec/fixture/sources.output.json
```diff
@@ -7,14 +7,6 @@
     "message": "sources.invalid-type.type: expected one of [vector, raster, raster-dem, geojson, video, image], \"invalid\" found",
     "line": 7
   },
-  {
-    "message": "sources.no-tilejson-properties-with-url.tiles: a source with a \"url\" property may not include a \"tiles\" property",
-    "line": 12
-  },
-  {
-    "message": "sources.no-unknown-properties-with-url.foo: a source with a \"url\" property may not include a \"foo\" property",
```
asheemmamoowala: Why is unknown properties support needed?

jingsam: I think foreign members (unknown properties) in source options should be leaved as it is. Foreign members does not affect source's functionality and add extensibility for metadata or custom sources. As we don't invalidate foreign members in geojson、image、video sources, invalidate foreign members in  vector, raster, raster-dem seems inconsistent.

asheemmamoowala:
cc @mapbox/studio

test/unit/style-spec/fixture/sources.output.json
asheemmamoowala: @jingsam Sorry for the delay in getting back to this. Allowing foreign members in the TileJSON does not seem necessary as part of this PR. It is preferable to have strict validation to ensure that unnecesary or incorrect properties are not being added to such files.

jingsam:
@asheemmamoowala I have invalid foreign members in source options when `url` is defined. As you can see, I have enum so many possible valid members. This code is fragile as someday we may need new property in test cases or application. If we now restrict properties allowed in sources options, that may make sources options can not be extended. Actually, in mapbox-style-spec, there is no limitation on foreign properties. So limit in here is sounds unreasonable. Forregin properties are hurtless, if mapbox-gl-js does not use them, take as it is but not restrict them.
https://github.com/mapbox/mapbox-gl-js/blob/58fa110a3360b2954d06af9ce43995631542586d/src/style-spec/validate/validate_source.js#L32-L38

jingsam:
@asheemmamoowala I add a test, please check out.

PR #8247, opened 2019-05-14
Provide an es modules build in dist/ for mapbox-gl-style-spec

ahocevar:
This pull request adds an es modules bundle to the `@mapbox/mapbox-gl-style-spec` package. With the accompanying changes in `package.json`, es module capable bundlers like webpack will automatically pick up the es module instead of the umd one. The result will be a smaller application bundle thanks to tree shaking.

### DIFF
--- a/src/style-spec/rollup.config.js
+++ b/src/style-spec/rollup.config.js
@@ -6,12 +6,20 @@ import unassert from 'rollup-plugin-unassert';
 import json from 'rollup-plugin-json';
 import {flow} from '../../build/rollup_plugins';
 
+// Build es modules?
+const esm = 'esm' in process.env;
+
+const transforms = {
+    dangerousForOf: true,
+    modules: esm ? false : undefined
+};
+
 const config = [{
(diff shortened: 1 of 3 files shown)


src/style-spec/rollup.config.js
```diff
@@ -6,12 +6,20 @@ import unassert from 'rollup-plugin-unassert';
 import json from 'rollup-plugin-json';
 import {flow} from '../../build/rollup_plugins';
 
+// Build es modules?
+const esm = 'esm' in process.env;
+
+const transforms = {dangerousForOf: true};
+if (esm) {
+    transforms.modules = false;
```
arindam1993: could you add this param using ternary syntax in the object declaration itself? just for consistency.

ahocevar: Thanks for the review @arindam1993, I have made the change you requested.

arindam1993:
Did a quick test with studio to validate the build it looks good, thanks @ahocevar.

PR #8255, opened 2019-05-16
prefer breaking lines at a zero width space.

ansis:
Mapbox data sources use zero width spaces to suggest better break points for Japanese labels. This somewhat hacky approach avoids adding more complex line breaking logic to -gl-js.

@asheemmamoowala @lily-chai 

## Launch Checklist

 - [x] tagged @mapbox/gl-native if this PR includes shader changes or needs a native port

### DIFF
--- a/src/symbol/shaping.js
+++ b/src/symbol/shaping.js
@@ -276,17 +276,18 @@ function calculateBadness(lineWidth: number,
     return raggedness + Math.abs(penalty) * penalty;
 }
 
-function calculatePenalty(codePoint: number, nextCodePoint: number) {
+function calculatePenalty(codePoint: number, nextCodePoint: number, ideographicBreak: boolean) {
     let penalty = 0;
     // Force break on newline
     if (codePoint === 0x0a) {
         penalty -= 10000;
     }
-    // Prioritize zero width spaces because the are used by mapbox as a way
-    // to mark ideal break points in Japanese text.
-    if (codePoint === 0x200b) {
-        penalty -= 50;
+    // Penalize breaks between characters that allow ideographic breaking because
+    // they are less preferable than breaks at spaces (or zero width spaces).
+    if (ideographicBreak) {
+        penalty += 150;
     }
+
     // Penalize open parenthesis at end of line
     if (codePoint === 0x28 || codePoint === 0xff08) {
         penalty += 50;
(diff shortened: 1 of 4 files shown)


alexshalamov:
@ansis I think I should port it to gl-native, right?

test/unit/symbol/shaping.test.js
```diff
@@ -55,6 +55,15 @@ test('shaping', (t) => {
     if (UPDATE) fs.writeFileSync(path.join(__dirname, '/../../expected/text-shaping-newlines-in-middle.json'), JSON.stringify(shaped, null, 2));
     t.deepEqual(shaped, expectedNewLinesInMiddle);
 
+    // Prefer zero width spaces when breaking lines. Zero width spaces are used by Mapbox data sources as a hint that
+    // a position is ideal for breaking.
+    const expectedZeroWidthSpaceBreak = JSON.parse(fs.readFileSync(path.join(__dirname, '/../../expected/text-shaping-zero-width-space.json')));
+
+    shaped = shaping.shapeText(Formatted.fromString('A B​C D​E F​G H I J K L​M N'), glyphs, fontStack, 5 * oneEm, oneEm, 'center', 'center', 0, [0, 0], WritingMode.horizontal);
```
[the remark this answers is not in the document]
ansis: Done in 8326234

src/symbol/shaping.js
```diff
@@ -282,6 +282,11 @@ function calculatePenalty(codePoint: number, nextCodePoint: number) {
     if (codePoint === 0x0a) {
         penalty -= 10000;
     }
+    // Prioritize zero width spaces because the are used by mapbox as a way
```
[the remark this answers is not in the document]
ansis: Changed in 8326234

ansis:
@lily-chai found that prioritizing zero width spaces couldn't effectively both break at zws and not break too often. Instead I've implemented penalization for ideographic breaks which seems to work better.

@alexshalamov yes!

src/symbol/shaping.js
```diff
@@ -366,18 +372,19 @@ function determineLineBreaks(logicalInput: TaggedString,
...
-            potentialLineBreaks.push(
-                evaluateBreak(
-                    i + 1,
-                    currentX,
-                    targetWidth,
-                    potentialLineBreaks,
-                    calculatePenalty(codePoint, logicalInput.getCharCode(i + 1)),
-                    false));
+        if ((i < logicalInput.length() - 1)) {
+            const ideographicBreak = !breakable[codePoint] && charAllowsIdeographicBreaking(codePoint);
```
asheemmamoowala: I dont see any overlap between `charAllowsIdeographicBreaking`and `breakable` codepoints. This check could be just  `charAllowsIdeographicBreaking`;

src/symbol/shaping.js
```diff
@@ -276,12 +276,18 @@ function calculateBadness(lineWidth: number,
...
+function calculatePenalty(codePoint: number, nextCodePoint: number, penalizableIdeographicBreak: boolean) {
     let penalty = 0;
     // Force break on newline
     if (codePoint === 0x0a) {
         penalty -= 10000;
     }
+    // Penalize breaks between characters that allow ideographic breaking because
+    // they are less preferable than breaks at spaces (or zero width spaces).
+    if (penalizableIdeographicBreak) {
+        penalty += 150;
```
1ec5: From the rendering test failures, it looks like this change regresses the balanced line breaking in #3743. The penalty is probably too high.

src/symbol/shaping.js
```diff
@@ -276,12 +276,18 @@ function calculateBadness(lineWidth: number,
...
+    }
+
     // Penalize open parenthesis at end of line
     if (codePoint === 0x28 || codePoint === 0xff08) {
         penalty += 50;
```
1ec5: The text-max-width/ideographic-punctuation-breaking test failure shows that this penalty is too low compared to the new penalty introduced above.

ansis:
I've fixed the render tests by enabling the penalty only when a zws is present in the string.

Capturing a conversation with @1ec5 from chat: This exception isn't great but shouldn't pose any problems if we change our approach in the future.

PR #8259, opened 2019-05-17
Fixes #8256

### LINKED ISSUE #8256: Map zoom fires "movestart" many times, but "moveend" only once
I'm not sure if this is a bug, but it was unexpected behavior to me.

I'm attaching event listeners to `movestart` and `moveend`. 

I'm finding that while dragging / panning the map, the event listeners trigger when I would expect them to trigger (when user starts moving the map `movestart` fires, then when they release the map `moveend` fires) 

However while zooming it looks like `movestart` is firing many many times while `moveend` is still only firing once. 

This seems somewhat related to #6512, but with the zoom api instead of resize.

**mapbox-gl-js version**:
0.53.1
**browser**:
Chrome

### Steps to Trigger Behavior

 1. attach map.on('movestart', () =>{}) and map.on('moveend', () => {}) event listeners.
 2. zoom the map in or out.

### Link to Demonstration

https://jsfiddle.net/bambielli767/Lc5yutef/9/

### Expected Behavior

`movestart` and `moveend` are both triggered once, when zooming starts and when zooming finishes.

### Actual Behavior

`movestart` is triggered many many times, while `moveend` is still only triggered once.

bambielli-flex:
## Launch Checklist

This PR is to fix #8256. By wrapping the movestart event emitter in a check to see if `_zooming` has already started, it will only fire once. 

This affects both movestart and zoomstart, both of which suffered from the problem outlined in the issue. 

 - [x] ~document any changes to public APIs~
 - [x] ~post benchmark scores~ <-- couldn't get these to run
 - [x] ~tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec changes~
 - [x] ~tagged `@mapbox/gl-native` if this PR includes shader changes or needs a native port~

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -12,7 +12,8 @@ There are two test suites associated with Mapbox GL JS
 
  To run individual tests:
 
- - Unit tests: `yarn test-unit path/to/file.test.js` where the path begins within the `/test/unit/` directory
+ - Unit tests: `yarn test-unit path/to/file.test.js` where path *does not include* `test/unit/` 
+   - e.g. `yarn test-unit ui/handler/scroll_zoom.test.js`
  - Render tests: `yarn test-render render-test-name` (e.g. `yarn test-render background-color/default`)
 
 ## Writing Unit Tests
--- a/test/unit/ui/handler/scroll_zoom.test.js
+++ b/test/unit/ui/handler/scroll_zoom.test.js
@@ -4,6 +4,7 @@ import window from '../../../../src/util/window';
 import Map from '../../../../src/ui/map';
 import DOM from '../../../../src/util/dom';
 import simulate from 'mapbox-gl-js-test/simulate_interaction';
+import sinon from 'sinon';
 
 function createMap(t) {
     t.stub(Map.prototype, '_detectMissingCSS');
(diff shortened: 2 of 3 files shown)


bambielli-flex:
Oops... linting rules workspace collision...

bambielli-flex:
Huh for some reason I can't get the benchmarks to run, even on master... running out of memory.

bambielli-flex:
Sorry I'm not super familiar with tap, but Is there a way to flush timers? 

I want to write a test like this:

```
t.test('emits one movestart event and one moveend event while zooming', (t) => {
        const map = createMap(t);
        let startCount = 0;
        map.on('movestart', () => {
            startCount += 1;
        });

        let endCount = 0
        map.on('moveend', () => {
            console.log('ending');
            endCount += 1;
            console.log('endconunt', endCount);
        });

        map._renderTaskQueue.run();
        const startZoom = map.getZoom();

        const events = [
            [2, {type: 'trackpad', deltaY: -1}],
            [7, {type: 'trackpad', deltaY: -2}],
            [30, {type: 'trackpad', deltaY: -3}],
            [1, {type: 'wheel', deltaY: -4}],
            [5, {type: 'wheel', deltaY: -5}],
            [20, {type: 'wheel', deltaY: -6}],
            [22, {type: 'wheel', deltaY: -7}],
        ];

        const end = now + 2000;
        let lastWheelEvent = now;

        // simulate the above sequence of wheel events, with render frames
        // interspersed every 20ms
        while (now++ < end) {
            if (events.length && lastWheelEvent + events[0][0] === now) {
                const [, event] = events.shift();
                simulate.wheel(map.getCanvas(), event);
                lastWheelEvent = now;
            }
            if (now % 20 === 0) {
                map._renderTaskQueue.run();
            }
        }
        console.log('startcount', startCount);
        console.log('endcount', endCount);
        t.equal(startCount, 1);
        t.equal(endCount, 1); 
        t.end();

    });
```
This test simulates many wheel / trackpad events. Previously this test would have emitted 6 movestart events, now it only emits 1.

Since the "moveend" is wrapped in a setTimeout in the scroll_zoom handler, my assertion on moveend not pass. Is there a way to flush any pending timers?

bambielli-flex:
I also tried using sinon's fake timers, but that didn't solve the problem either.

test/README.md
```diff
@@ -12,7 +12,8 @@ There are two test suites associated with Mapbox GL JS
 
  To run individual tests:
 
- - Unit tests: `yarn test-unit path/to/file.test.js` where the path begins within the `/test/unit/` directory
+ - Unit tests: `yarn test-unit path/to/file.test.js` where path *does not include* `test/unit/` 
+   - e.g. `yarn test-unit ui/handler/scroll_zoom.test.js`
```
bambielli-flex: I found this instruction confusing, took me a while to realize that I needed to *not* include /test/unit/ in the test path to be able to run a test.

test/unit/ui/handler/scroll_zoom.test.js
```diff
@@ -148,5 +149,93 @@ test('ScrollZoomHandler', (t) => {
         t.end();
     });
 
+    t.test('emits one movestart event and one moveend event while zooming', (t) => {
```
bambielli-flex: I confirmed that these tests would fail on master without my update.

PR #8262, opened 2019-05-19
ensure fetch/ajax image requests accept webp when supported

### LINKED ISSUE #8261: ensure image requests send an accept header
In modern browsers when you request an image if the browser supports webp it will add this to the request accept header, so if the server supports serving different content based on the client support it can.

However, when making a Fetch or XMLHttpRequest, the browser will pass `*/*` as the accept header, in which case the server won't know if the client supports webp or not.

Mapbox raster sources have built in logic at https://github.com/mapbox/mapbox-gl-js/blob/f5c2ae0f3d39d11ab622e09d1523a24e11c2887c/src/util/mapbox.js#L96-L110 to ask for a webp tile when supported, but this change could help some 3rd party raster sources.

I've drafted a PR to address this at #8262

andrewharvey:
## Launch Checklist

fixes #8261

See further detail at #8261 , but in essence this PR ensures image requests from GL JS to servers, include accepts image/webp where supported, so the server knows if it can respond with a webp image or not
 - [ ] post benchmark scores
 - [x] ~~check this webp support test is only run once, instead of at each tile request, and what impact it has on performance~~ tested, only run once
tested

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -6,6 +6,7 @@ import {isMapboxHTTPURL, hasCacheDefeatingSku} from './mapbox';
 import config from './config';
 import assert from 'assert';
 import {cacheGet, cachePut} from './tile_request_cache';
+import webpSupported from './webp_supported';
 
 import type {Callback} from '../types/callback';
 import type {Cancelable} from '../types/cancelable';
@@ -266,6 +267,13 @@ export const resetImageRequestQueue = () => {
 resetImageRequestQueue();
 
 export const getImage = function(requestParameters: RequestParameters, callback: Callback<HTMLImageElement>): Cancelable {
+    if (webpSupported.supported) {
+        if (!requestParameters.headers) {
+            requestParameters.headers = {};
+        }
+        requestParameters.headers.accept = 'image/webp,*/*';
+    }
+
     // limit concurrent image loads to help with raster sources performance on big screens
     if (numImageRequests >= config.MAX_PARALLEL_IMAGE_REQUESTS) {
         const queued = {
--- a/test/unit/util/ajax.test.js
+++ b/test/unit/util/ajax.test.js
@@ -195,24 +195,16 @@ test('ajax', (t) => {
         });
 
         // mock webp support
-        webpSupported.supported = true
+        webpSupported.supported = true;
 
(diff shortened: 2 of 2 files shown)


andrewharvey:
@andrewharvey Are you still working on this PR? The change itself looks fine to accept, but I want to understand how this affects the serving of non-tile resources if webp is added to the accept header, before moving forward with it.

Yes, I can pick this up again.

> @samanpwbb @jseppi Would the sprite images endpoint be affected by the addition of an accepts header with image/webp?

For more context, GL JS will always request a sprite image with the extension `.png`:

https://github.com/mapbox/mapbox-gl-js/blob/87cac3be79880fe6b6cf532baa023c52b112cf1f/src/style/load_sprite.js#L28

However as explained at #8261, when making a `Fetch` or `XMLHttpRequest`, the browser will pass */* as the accept header and hence the server won't be able to detect if the client supports webp or not.

I tested Mapbox's API requesting the `sprite.png` with `accepts: image/webp` and there was no difference, the response was still the PNG sprite, which is fine. So it doesn't look like this breaks anything for sprites. Even if the server was smart and detected the accepts header and returned a webp sprite in spite of the `.png` extension that should also be fine, it should work.

andrewharvey:
I did further testing,

On Chrome, it works fine, only tests for webp support once and requests and accepts webp resources from the start.
On Firefox, it would re-check webp support for each request, and there was a lag for the webp support check coming back positive, so the first few requests wouldn't pass the webp header but later requests would. It looks like this issue stems from the webp support check.

asheemmamoowala:
> On Firefox, there was a lag for the webp support check coming back positive, so the first few requests wouldn't pass the accepts webp header but later requests would. It looks like this issue stems from the webp support check.

@andrewharvey how much of an issue is this? Does it matter if a few requests don't use `webp` images, or should the requests be blocked until the check is completed?

andrewharvey:
> @andrewharvey how much of an issue is this? Does it matter if a few requests don't use webp images, or should the requests be blocked until the check is completed?

Nothing more than a minor annoyance. I wouldn't hold up the requests because of this.

andrewharvey:
> @andrewharvey Are you still working on this? I think we just need a unit test to merge it.

@ryanhamley yes! unit test committed :heavy_check_mark:

PR #8269, opened 2019-05-21
v1.0.0

ryanhamley:
## Launch Checklist

Update version to 1.0.0


jingsam:
wow, this is a big thing.

ifzm:
wow, nice

mourner:
@jingsam yeah, we think the library is way past being stable enough for production use, and there are no big reasons to put this off any more. :)

PR #8274, opened 2019-05-22
[docs] Remove /map-tiles GL JS example

katydecorah:
## Launch Checklist

This PR removes the example https://docs.mapbox.com/mapbox-gl-js/example/map-tiles/ since it's using a classic style. The path will redirect to https://docs.mapbox.com/mapbox-gl-js/example/satellite-map/ 


asheemmamoowala:
> The path will redirect 

The redirect is put in place elsewhere right?

katydecorah:
@asheemmamoowala  correct! The redirect will happen in https://github.com/mapbox/subdomain-docs/pull/29 I will merge here after the redirect PR is merge

andrewharvey:
:wave: @katydecorah I find the "Add a raster tile source" example very helpful, it demonstrates how to add a 3rd party raster tile layer to a map.

So I think it's useful to keep. If there's a concern about using the classic style, perhaps we could point it to a different raster tile source?

The satellite example is good, but since it just uses the default mapbox satellite style, it doesn't quite show the same thing.

asheemmamoowala:
> So I think it's useful to keep. If there's a concern about using the classic style, perhaps we could point it to a different raster tile source?

If the value of this example is to show how to add a raster tile source that does not have a TileJSON file, then maybe we can show how to add the mapbox satellite tile layer with in-lined `tiles` property ?

colleenmcginnis:
>If the value of this example is to show how to add a raster tile source that does not have a TileJSON file, then maybe we can show how to add the mapbox satellite tile layer with in-lined tiles property?

@asheemmamoowala I think this could get confusing for the reader since we have a separate example for using the satellite _style_ https://docs.mapbox.com/mapbox-gl-js/example/satellite-map/, which I believe is the preferred method. A reader looking for _the_ way to add Mapbox satellite imagery to their application may not be sure which to use.

docs/pages/example/map-tiles.html
```diff
@@ -8,8 +8,9 @@
         "sources": {
             "raster-tiles": {
                 "type": "raster",
-                "url": "mapbox://" + tileset,
-                "tileSize": 256
+                "tiles": ["http://tile.stamen.com/watercolor/{z}/{x}/{y}.jpg"],
+                "tileSize": 256,
+                "attribution": '<a id="home-link" target="_top" href="../" class="">Map tiles</a> by <a target="_top" href="http://stamen.com">Stamen Design</a>, under <a target="_top" href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a target="_top" href="http://openstreetmap.org">OpenStreetMap</a>, under <a target="_top" href="http://creativecommons.org/licenses/by-sa/3.0" class="">CC BY SA</a>'
```
andrewharvey: how about this?

`Map tiles by <a target="_top" rel="noopener" href="http://stamen.com">Stamen Design</a>, under <a target="_top" rel="noopener" href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a target="_top" rel="noopener" href="http://openstreetmap.org">OpenStreetMap</a>, under <a target="_top" rel="noopener" href="http://creativecommons.org/licenses/by-sa/3.0">CC BY SA</a>`

colleenmcginnis:
👋 @katydecorah can you take a look at the changes I made, rebase, and merge if all looks good?

Issue #8243, opened 2019-05-14
Reduce the amount of data uploaded to the GPU


mourner:
One of the things that affects GL JS rendering performance linearly and that we can track reliably is the amount of data we upload to the GPU for rendering (given the same style). Specifically, the total size of data given to `bufferData` calls.

There are two directions we can explore for this — fixing older regressions and finding new ways to upload less.

### Fixing regressions

Looking at historical metrics of how much data we upload throughout previous GL JS version, I found 3 regressions:

1. As a result of #3061 and #3120, we upload data for zero-opacity layers, which can significantly increase the amount of unnecessarily uploaded data for some style/zoom combinations. On z11.1 Mapbox Streets v10 sample, it's a ~12% regression.
2. In #5073, we had to introduce a ~25% regression to work around a GPU bug on old Android phones. @ansis believes we can win this back while keeping the bug fixed by packing normals differently.
3. #5184 has a ~5% regression but I'm not sure why. Needs to be investigated.

### Adding new savings

Some things we could explore from the top of my mind:

1. Using instanced rendering (see also #1898) to significantly reduce vertex buffers. Should be pretty straightforward for instancing things like circles and heatmap kernels (4x reduction). It will be harder to take advantage of this for lines/fills, but we should explore possibilities.
2. Revisiting how we tesselate lines and seeing if we can reduce the amount of triangles generated for line joins and line caps.

cc @ansis @kkaefer

mourner:
> #5184 has a ~5% regression but I'm not sure why. Needs to be investigated.

Turns out the overhead is in fill layers with `fill-antialias: false`. Previously, since we uploaded buffers on render, we didn't upload line index data for outline rendering if it's not drawn. Now, since we upload buffers beforehand, line index data is uploaded regardless whether we need it or not. cc @ansis

mourner:
> As a result of #3061 and #3120, we upload data for zero-opacity layers, which can significantly increase the amount of unnecessarily uploaded data for some style/zoom combinations. On z11.1 Mapbox Streets v10 sample, it's a ~12% regression.

This is also something that's now hard to fix because of the coalescing of buffer uploads independently of drawing (#5184). However, it's probably not worth fixing because it's rare to have affected layers in newer, expression-based styles (which usually have feature-based rather than global zoom-based opacity set).

PR #8275, opened 2019-05-22
A more accurate algorithm for approximating round line joins

mourner:
The algorithm we used for approximating round joins with triangles had several issues:

1. Since the number of slices was linearly based on cosine of half angle between normals (rather than the angle itself), it led to wildly inconsistent approximation — sharper corners were over-approximated while flatter corners were under-approximated. Here, the left turn uses 2 triangles while the right one 6 (!):

image

2. We used linear (rather than angular) interpolation between normals to generate triangles, which meant that triangles were inconsistent in size — those closer to the join normal were bigger and those closer to prev/next normals smaller:

Fig. 4: uneven slices

3. Since we always generated a pie slice vertex in the middle, the number of approximating triangles could only be even, which lead to sometimes missing an opportunity for better approximation (e.g. using 2 where 3 would be better or using 6 when 5 would suffice).

These issues combined meant that we 1) generated more triangles than needed overall, 2) had inconsistent quality of approximation.

This PR changes the algorithm to use a more mathematically correct spherical interpolation, solving both issues. To avoid trigonometric calculations in this perf-sensitive section of the code, I adopted polynomial curve fitting approximation [demonstrated here](https://observablehq.com/@mourner/approximating-geometric-slerp). The quality of approximation is now controlled by the constant `DEG_PER_TRIANGLE`, which is 30 degrees by default — this seems to be an OK value quality-wise, and reduces the number of GPU buffer size on a typical streets z11 view by ~4-5% (#8243). We could definitely tweak this though. Here's how this looks for oversized lines at the moment (3 triangles on the left, 2 on the right):

image

Pending benchmark results, review and a discussion of the constant.

## Launch Checklist

 - [x] ~~write tests for all new functionality~~ covered by render tests
 - [x] post benchmark scores

### DIFF
--- a/src/data/bucket/line_bucket.js
+++ b/src/data/bucket/line_bucket.js
@@ -52,6 +52,9 @@ const EXTRUDE_SCALE = 63;
 const COS_HALF_SHARP_CORNER = Math.cos(75 / 2 * (Math.PI / 180));
 const SHARP_CORNER_OFFSET = 15;
 
+// Angle per triangle for approximating round line joins.
+const DEG_PER_TRIANGLE = 30;
+
 // The number of bits that is used to store the line distance in the buffer.
 const LINE_DISTANCE_BUFFER_BITS = 15;
 
@@ -331,11 +334,16 @@ class LineBucket implements Bucket {
              *
              */
 
-            // Calculate the length of the miter (the ratio of the miter to the width).
-            // Find the cosine of the angle between the next and join normals
-            // using dot product. The inverse of that is the miter length.
+            // calculate cosines of the angle (and its half) using dot product
+            const cosAngle = prevNormal.x * nextNormal.x + prevNormal.y * nextNormal.y;
             const cosHalfAngle = joinNormal.x * nextNormal.x + joinNormal.y * nextNormal.y;
+
+            // Calculate the length of the miter (the ratio of the miter to the width)
+            // as the inverse of cosine of the angle between next and join normals
             const miterLength = cosHalfAngle !== 0 ? 1 / cosHalfAngle : Infinity;
+
+            // approximate angle from cosine
+            const approxAngle = 2 * Math.sqrt(2 - 2 * cosHalfAngle);
 
             const isSharpCorner = cosHalfAngle < COS_HALF_SHARP_CORNER && prevVertex && nextVertex;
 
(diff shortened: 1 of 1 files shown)


mourner:
No noticeable changes in the benchmarks:

image
image
image

mourner:
@asheemmamoowala render tests should stay the same I think — these changes should be mostly not noticeable by design. I measured the impact locally with a debug page. How would you like to measure this instead?

mourner:
cc @mapbox/gl-native — will need a port to native if we land this.

asheemmamoowala:
> How would you like to measure this instead?

The OP for this PR states that:

> 1) generated more triangles than needed overall,

I think it would be enough to document the approximate/average reduction in vertices or buffer size as part of the PR, until we have an automated way to measure that.

mourner:
> I think it would be enough to document the approximate/average reduction in vertices or buffer size as part of the PR, until we have an automated way to measure that.

@asheemmamoowala still confused by what you mean by "document" — there's the following bit in the PR description:

> reduces the number of GPU buffer size on a typical streets z11 view by ~4-5%

asheemmamoowala:
> reduces the number of GPU buffer size on a typical streets z11 view by ~4-5%

D'oh! I didnt see that. Sorry for the trouble!

mourner:
@ansis on my local test of streets-v11 with line layers only (less is better):

- `master`: 13,296,410
- 30 degrees: 12,591,730
- 25 degrees: 12,839,554
- 20 degrees: 13,119,920
- 15 degrees: 13,653,826

So if we want to be close to original precision for the 90 degrees case while not regressing on the buffer size, we could change the value to 20 degrees for now, and look at ways to make that adaptive in a future PR.

> Trying to think about other ways we could reduce artifacts without increasing the number of total triangles. Would the total number of geometry vertices in a layer be something we could branch on? If there are a lot of vertices reduce the quality. If there are fewer, increase it?

It might be hard to come up with a good heuristic for basing precision on number of points in a layer, since it depends on the style and data so much, but at least we can assume that if there are a lot of vertices, the lines are unlikely to have long straight segments where approximation differences would be very visible.

Perhaps we could come up with a heuristic based on max distance between points in a particular line?

Anyway, I think it's better to merge the PR without a heuristic and then explore this in a different PR.

PR #8278, opened 2019-05-22
Fix broken benchmarks

mourner:
This was broken since #8220 because of a type mismatch.


mourner:
@tristen that's the current master failing; in theory, it should be fixed once this PR is merged.

PR #8281, opened 2019-05-23
Start adopting Flow strict mode

mourner:
This PR demonstrates how we can start tightening up our typing with Flow strict mode:
https://flow.org/en/docs/strict/

Strict mode enables stronger safety guarantees by banning unsafe types such as `any`, which currently plague our codebase with a false sense of safety, allowing for serious bugs to slip through. `any` is kind of a virus — any code that touches it gets "tainted", and we used it a lot, leaving its removal for later. As a result, basically the majority of our codebase is unsafely typed (which is often not better than untyped).

Strict mode on the other hand is viral too — you can enable it on a file-by-file basis, and it enforces every import to be strictly typed too. That means we can start with typing "leaves" and external dependencies, and go up the tree until we type out everything.

In this PR, I started with typing a few utility files and a few dependencies. We can merge it to get the config in, and then gradually type the rest of the code base with a succession of minimal PRs.


src/util/window.js
```diff
@@ -82,7 +81,7 @@ function restore(): Window {
     window.ImageData = window.ImageData || function() { return false; };
     window.ImageBitmap = window.ImageBitmap || function() { return false; };
     window.WebGLFramebuffer = window.WebGLFramebuffer || Object;
-    extend(_window, window);
+    Object.assign(_window, window); // eslint-disable-line no-restricted-properties
```
asheemmamoowala: This will break in IE11. I misunderstood what the bublé `objectAssign` configuration does.

mourner: It won't because this file is a Node stub for testing, replaced by `util/browser/window.js`.

asheemmamoowala: 🤕

PR #8289, opened 2019-05-28
Fix #8273: Layer#setPaintProperty returns true for CrossFadedDataDriv…

### LINKED ISSUE #8273: setPaintProperty not work on click event
**mapbox-gl-js version**: 0.54.0

**browser**: chrome

### Steps to Trigger Behavior
click map to excute setPaintProperty function, but it don't work, the property line-pattern not changed

### Link to Demonstration
https://jsbin.com/rukoji/edit?js,output

### Expected Behavior
change paint property, the line-pattern will be greenArrow
### Actual Behavior
don't work, the line-pattern is still redArrow

ahk:
This patch fixes a bug reported in issue #8273. Users were unable to update `line-pattern` property images via `Map#setPaintProperty.` There is only one change and it removes an unnecessary check that the previous value for this property didn't exist. As long as the value has changed (which is checked in `Style#setPaintProperty` before deferral to the layer) then the layer needs to be updated, or for example the image used for the property value isn't added to the texture atlas.

It's possible this affects performance, but benchmarks didn't seem to indicate that. I can imagine that some updates from a non-image value would only need re-layout on changing to or from a null value. Any opinions on that question are greatly appreciated.

## Launch Checklist

 - [x] post benchmark scores


peterqliu:
👍  want to add a new test ensuring this keeps working moving forward?

src/style/style_layer.js
```diff
@@ -155,13 +155,13 @@ class StyleLayer extends Evented {
             // so a call to _updateLayer is necessary, and we return true from this function so it gets called in
             // Style#setPaintProperty
             const prop = this._transitionablePaint._values[name];
-            const newCrossFadedValue = prop.property.specification["property-type"] === 'cross-faded-data-driven' && !prop.value.value && value;
+            const isCrossFadedValue = prop.property.specification["property-type"] === 'cross-faded-data-driven';
```
asheemmamoowala: Strictly speaking, the variable is representing `isCrossFadedProperty`.

ahk: Ohhhh yes thank you. Was following the existing name but I agree with you.

ahk:
> 👍 want to add a new test ensuring this keeps working moving forward?

Is it correct to try to land a test in `test/unit/style/style_layer.test.js` ?

peterqliu:
@ahk yep, set it up to reproduce the erstwhile error state and make sure getPaintProperty returns what you expect

can also do a render test to make sure the proper pattern applies, but seems like a unit test should suffice

ahk:
@asheemmamoowala I read through a ton of the render tests, wanted to make sure my thinking was correct before I created one. I think I should land a test into the `regressions` directory. I feel [this test](
https://github.com/mapbox/mapbox-gl-js/blob/master/test/integration/render-tests/regressions/mapbox-gl-js%232533/style.json) is the best example of what I'm trying to do (exercise StyleLayer#setPaintProperty), I just want to switch out `fill-pattern` for `line-pattern` as the property I'm updating.

mourner:
@ahk correct! One small note is that we're trying to add small-sized tests if possible (so not 512x512 like the one above, but e.g. 64x64 if it still can exercise the bug).

ahk:
fix-#8273-benchmarks

PR #8291, opened 2019-05-29
Fix touch listeners

pakastin:
Forgot to make listeners passive for compass drag events in #8226 – fixed that. Also fixed https://github.com/mapbox/mapbox-gl-js/commit/27ac83a63eecdec3c9f5597d87fddcf1f3bdd660#r33719348 to make master work again in https://aviamaps.com/map on iOS.


mourner:
Oops, let me try to fix that Flow error first...

mourner:
Landed in #8292. Sorry, looks like we'll have to redo #8226 :(

PR #8292, opened 2019-05-29
Revert broken compass touch support; fix flow issues and passive events on iOS

mourner:
#8281 (just merged) accidentally broke passive events on iOS, and also broke Flow tests for the last master commit. The PR should fix both issues. cc @pakastin — note that I use `mousePos` for touch events too because this function already works with them and should be equivalent.

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -68,7 +68,7 @@ class NavigationControl {
     }
 
     _rotateCompassArrow() {
-        const rotate = `rotateX(${this._map.transform.pitch}deg) rotateZ(${this._map.transform.angle * (180 / Math.PI)}deg)`;
+        const rotate = `rotate(${this._map.transform.angle * (180 / Math.PI)}deg)`;
         this._compassArrow.style.transform = rotate;
     }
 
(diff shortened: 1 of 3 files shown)


mourner:
I just tested both this PR and master pre-Flow commit and it looks like merged #8226 was broken anyway. :( After dragging the compass, it 1) resets to north after dragging, 2) makes the touch interaction "stuck" afterwards when dragging the map as if one additional finger touches the map.

pakastin:
I already fixed those issues in #8291 – just forgot to add the `{ passive: false }`. You can try #8226 + #8291 in https://aviamaps.com/map

mourner:
@pakastin my apologies for the mess! Would you mind opening a new PR that combines those on top of master (+ the `mousePos` change to make Flow happy)?

pakastin:
Sure yeah, no worries!

ryanhamley:
Should we take this opportunity to make the compass pitching an optional feature and implement the click reset behavior as per https://github.com/mapbox/mapbox-gl-js/issues/8208#issuecomment-496664716?

pakastin:
Sounds good!

pakastin:
It's done! #8296

You can try it out here: https://aviamaps.com/map

Issue #8293, opened 2019-05-29
removeFeatureState doesn't handle id '0'


asheemmamoowala:
Moving forward from [#7632(comment)](https://github.com/mapbox/mapbox-gl-js/issues/7632#issuecomment-494985798)

> Looking at removeFeatureState function in the SourceFeatureState class, it does use the number feature id, but the if statements would fail if the id is 0. Lines 65, 70. Will need to test this to verify.

https://github.com/mapbox/mapbox-gl-js/blob/a67c21368ad363ae48fece3cf8937327dbdfe172/src/source/source_state.js#L61-L70

peterqliu:
@asheemmamoowala that was by design 🤔  _should_ it handle zero? or negative values, or even strings as the ticket suggests?

asheemmamoowala:
It need to handle `0` because that is a valid feature id. A string 0 can be cooerced to a number and will be handled before it gets to the `SourceState#removeFeatureState` method.

peterqliu:
🚢

PR #8302, opened 2019-05-30
accept id's >=0 for removeFeatureState

peterqliu:
## Launch Checklist

solves https://github.com/mapbox/mapbox-gl-js/issues/8293


test/unit/ui/map.test.js
```diff
@@ -1485,6 +1485,25 @@ test('Map', (t) => {
     });
 
     t.test('#removeFeatureState', (t) => {
+
+        t.test('accepts "0" id', (t) => {
+            const map = createMap(t, {
```
asheemmamoowala: This test passes even without the code change. That happens because the checks in `removeFeatureState` fall to the final `else` clause and remove all the state for the source.

src/source/source_state.js
```diff
@@ -62,12 +62,12 @@ class SourceFeatureState {
 
         this.deletedStates[sourceLayer] = this.deletedStates[sourceLayer] || {};
 
-        if (key && featureId) {
+        if (key && typeof featureId === 'number' && featureId >= 0) {
```
asheemmamoowala: Line 61 already converts the `featureId` function parameter to a string. It should be enough to switch from `featureId` to `feature` in these if clauses.

mourner: `typeof feature` will be `string` then. Generally, for a targeted fix of `removeFeatureState` not working with `0` id, it would be enough to do `featureId !== undefined` for consistency (we don't check validity in `setFeatureState` either).

peterqliu: >it would be enough to do featureId !== undefined

was originally gonna say that wouldn't catch id's like boolean types, but both implementations would treat booleans as invalid id's and therefore interpret the deletion as source-wide deletion 👍

test/unit/ui/map.test.js
peterqliu: 👍 added nuance to the test to make sure 0 id's aren't interpreted as source-wide state deletion, by adding a second property that should stay put

peterqliu:
@asheemmamoowala 👍 looks like this one can be squashed automatically

PR #8303, opened 2019-05-30
Optimize and simplify line tessellation code

mourner:
This PR refactors the line tessellation code, removing code repetition and some unnecessary logic, making it both faster and easier to follow.

The potential performance improvement comes from reducing the amount of temporary `Point` object allocations by ~2 times, which theoretically reduces the amount of subsequent GC pauses. I'll try to find a way to measure real-world performance impact, but roughly it should be 5-10% less time spent tessellating lines on the worker side.

It brings some ideas from #7085 but in a less intrusive way.

The PR is made on top of #8275 so needs to be applied after it lands.

## Launch Checklist

 - [x] post benchmark scores
 - [x] tagged @mapbox/gl-native — might need a port to native to make the code between platforms more similar.

### DIFF
--- a/src/data/bucket/line_bucket.js
+++ b/src/data/bucket/line_bucket.js
@@ -78,7 +78,6 @@ class LineBucket implements Bucket {
 
     e1: number;
     e2: number;
-    e3: number;
 
     index: number;
     zoom: number;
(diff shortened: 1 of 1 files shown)


src/data/bucket/line_bucket.js
```diff
@@ -231,15 +216,21 @@ class LineBucket implements Bucket {
     }
 
     addLine(vertices: Array<Point>, feature: BucketFeature, join: string, cap: string, miterLimit: number, roundLimit: number, index: number, imagePositions: {[string]: ImagePosition}) {
-        let lineDistances = null;
+        this.distance = 0;
+        this.scaledDistance = 0;
+        this.totalDistance = 0;
+
```
ansis: Do you think creating a line for each object is a significant cost? putting per-line values directly on the bucket seems a bit less nice but probably ok if there is a significant benefit.

mourner: @ansis we had `distance` as a bucket state before this PR, so the latter doesn't change the status quo in that sense — I moved the line stats to shared state too for simplicity/consistency, but I think we can revisit this in a follow-up PR.

mourner:
Layout benchmarks are not affected:

image

Issue #8304, opened 2019-05-30
Reduce line buffers back from 12 to 8 bytes per vertex


mourner:
A part of #8243. PRs #5073 followed by #5258, aiming to fix [an obscure Broadcom GPU bug](https://github.com/mapbox/mapbox-gl-native/issues/7583), also increased the amount of data we upload to the GPU for line layers by 50% — from 8 to 12 bytes. This directly affects rendering performance.

We should explore alternative packing approaches to get this back to 8 bytes while avoiding the GPU bug trigger. One thing to try would be to try packing the `round`/`up` attributes into most significant bit rather than least significant. cc @ansis

mourner:
Fixed in #8306

PR #8306, opened 2019-05-30
Switch back to a more compact line attributes layout

mourner:
Switch back from 12 to 8 bytes per vertex for line layers. See #8304. 

I tried to decode them in a potentially safer way this time (avoiding `mod` and `/` in shader math) to see if it fixes the Broadcom GPU bug — ~~we need to verify this before merging~~ we did verify it works properly on Broadcom VideoCore IV GPU, and faster than master 🎉. ~~If not, this PR will be a good starting point for exploring other packing strategies (e.g. packing in most significant bit).~~

## Launch Checklist

 - [x] ~~write tests for all new functionality~~ covered by render tests
 - [x] ~~post benchmark scores~~ ran Layout/Paint/LayerLine, they don't seem to be affected since this is a GPU thing
 - [x] tagged @mapbox/gl-native if this PR includes shader changes or needs a native port
 - [x] test on the affected Broadcom GPU device
 - [x] test GL Native counterpart on a Broadcom GPU device https://github.com/mapbox/mapbox-gl-native/pull/14851

### DIFF
--- a/src/data/array_types.js
+++ b/src/data/array_types.js
@@ -115,12 +115,12 @@ register('StructArrayLayout2i4i12', StructArrayLayout2i4i12);
 
 /**
  * Implementation of the StructArray layout:
- * [0]: Int16[4]
- * [8]: Uint8[4]
+ * [0]: Int16[2]
+ * [4]: Uint8[4]
  *
  * @private
  */
-class StructArrayLayout4i4ub12 extends StructArray {
+class StructArrayLayout2i4ub8 extends StructArray {
     uint8: Uint8Array;
     int16: Int16Array;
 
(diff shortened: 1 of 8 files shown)


ansis:
Would a new buffer size benchmark be useful to track these kinds of changes?

mourner:
> The testing so far looks promising but I think we should try to test this in -native with the buggy device before merging.

Yep, already submitted a PR https://github.com/mapbox/mapbox-gl-native/pull/14851 — will wait for confirmation of that one.

> Do you think using the most significant bit instead of the least significant bit could make it even more robust?

I don't know why it wasn't robust in the first place so not sure whether it will make a difference, but if the current approach is confirmed to work with those GPUs, we shouldn't have any problem on newer ones.

> Would a new buffer size benchmark be useful to track these kinds of changes?

Yes, I have it locally and want to PR, but it doesn't fit the benchmark format because it's about qualitative values, not timing. Might be worth pushing just as a debug page at first.

PR #8312, opened 2019-06-03
[DOCS] Update Popup maxWidth description

viniciuskneves:
It updates the `maxWidth` property description as it accepts not only pixel values but all available `max-width` CSS values. I've also added the default value to the documentation which was missing.
Also added some small info as a tip to users that want to have a fluid container size.

### DIFF
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -54,8 +54,8 @@ export type PopupOptions = {
  * @param {string} [options.className] Space-separated CSS class names to add to popup container
  * @param {string} [options.maxWidth='240px'] -
  *  A string that sets the CSS property of the popup's maximum width, eg `'300px'`.
- *  If you want a fluid container which sizes to the inner size, you can set this property to `'none'`.
- *  All available values can be found here: https://developer.mozilla.org/en-US/docs/Web/CSS/max-width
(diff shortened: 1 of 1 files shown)


src/ui/popup.js
```diff
@@ -52,7 +52,10 @@ export type PopupOptions = {
  *   - an object of {@link Point}s specifing an offset for each anchor position
  *  Negative offsets indicate left and up.
  * @param {string} [options.className] Space-separated CSS class names to add to popup container
- * @param {string} [options.maxWidth] A string that sets the CSS property of the popup's maxWidth in pixels, eg "300px"
+ * @param {string} [options.maxWidth='240px'] -
+ *  A string that sets the CSS property of the popup's maximum width, eg `'300px'`.
+ *  If you want a fluid container which sizes to the inner size, you can set this property to `'none'`.
```
ryanhamley: How about something like "To ensure the popup resizes to fit its content, set this property to `'none'`"?

src/ui/popup.js
```diff
@@ -52,7 +52,10 @@ export type PopupOptions = {
...
+ *  All available values can be found here: https://developer.mozilla.org/en-US/docs/Web/CSS/max-width
```
ryanhamley: Very minor nit but let's make this `Available values....` to match line 252

PR #8314, opened 2019-06-03
Update Flow to 0.100 and improve typing

mourner:
Updates Flow version from 0.95 to 0.100. Also improves typing in a bunch of files — after soundness fixes in recent Flow versions triggered a lot of errors.

Currently the failing build is unrelated to the changes (and rerunning a job with SSH, then removing node_modules and installing deps again make it pass) — I'll try to fix this by bumping yarn cache key.


mourner:
I also tried to unfork `remove-flow-types` by switching to the one now maintained in the main flow repo, but it has a blocking bug, so will need to wait: https://github.com/facebook/flow/issues/7779

src/source/worker_source.js
```diff
@@ -9,7 +9,6 @@ import type {Bucket} from '../data/bucket';
 import type FeatureIndex from '../data/feature_index';
 import type {CollisionBoxArray} from '../data/array_types';
 import type DEMData from '../data/dem_data';
-import type {PerformanceResourceTiming} from '../types/performance_resource_timing';
```
asheemmamoowala: Why is this file being deleted?

mourner: The type is present in the standard Flow definitions library, here: https://github.com/facebook/flow/blob/master/lib/bom.js#L250

PR #8315, opened 2019-06-03
Add base-62 test for sku tokens

ryanhamley:
## Launch Checklist

    - adds a test that sku tokens are composed of base-62 characters

### DIFF
--- a/test/unit/util/sku_token.test.js
+++ b/test/unit/util/sku_token.test.js
@@ -19,6 +19,12 @@ test('sku token generation', (t) => {
         t.end();
     });
 
+    t.test('token only contains base-62 characters', (t) => {
+        const regex = /^[0-9a-zA-Z]+$/;
+        t.ok(regex.exec(skuToken));
+        t.end();
+    });
+
     t.test('createSkuToken generates a unique token each time it\'s called', (t) => {
         const secondSkuToken = createSkuToken().token;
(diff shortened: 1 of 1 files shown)


test/unit/util/sku_token.test.js
```diff
@@ -19,6 +19,12 @@ test('sku token generation', (t) => {
         t.end();
     });
 
+    t.test('token only contains base-62 characters', (t) => {
+        const regex = /^[0-9a-zA-Z]+$/;
+        t.ok(regex.exec(skuToken));
```
asheemmamoowala: This test may not fail consistently if something was to break in the sku token generation. Is there a better way to test/verify that the base-62 character constraint is enforced?

ryanhamley: I don't think I understand what this test might not catch. We just want to check that the token consists only of characters 0-9, a-z, and A-Z right?

asheemmamoowala: If tokens are generated randomly, and there was a bug - only some tokens would fail this test and it would not be easy top replicate the test in any way to understand why it fails.

asheemmamoowala:
@ryanhamley this PR isn't robust enough to catch errors in the code it is testing. Closing for now. Please re-open with an alternative approach once you have addressed https://github.com/mapbox/mapbox-gl-js/pull/8315#pullrequestreview-247867345

PR #8336, opened 2019-06-11
Unit tests refactoring for less magic

mourner:
Addresses the concerns raised by @vakila — having a fake `node_modules` in the `test` folder to get magic path resolution was annoying (editors hide `node_modules` by default) and confusing for contributors.

This PR makes it much less magic. See individual commits when reviewing changes.


test/unit/ui/popup.test.js
```diff
@@ -4,7 +4,7 @@ import { createMap as globalCreateMap } from '../../util';
 import Popup from '../../../src/ui/popup';
 import LngLat from '../../../src/geo/lng_lat';
 import Point from '@mapbox/point-geometry';
-import { click, mousemove } from 'mapbox-gl-js-test/simulate_interaction';
+import simulate from '../../util/simulate_interaction';
```
vakila: This feels much nicer imo :heart:

test/util/index.js
```diff
@@ -28,3 +28,11 @@ export function createMap(t, options, callback) {
 
     return map;
 }
+
+export function equalWithPrecision(test, expected, actual, multiplier, message, extra) {
```
vakila: What's the rationale for moving `equalWithPrecision` to a standalone function? Since this is wrapping tap's built-in `equals` assertion, to me it makes sense to leave it as a custom assertion method called like all other tap assertions, instead of switching to this functional API - unless we have a strong reason that I'm missing.

test/unit/style-spec/declass.test.js
```diff
@@ -1,7 +1,7 @@
-import { test as t } from 'mapbox-gl-js-test';
+import { test } from '../../util/test';
```
vakila: Thanks a lot for fixing this naming inconsistency! :heart:

test/util/index.js
mourner: My main motivation is moving away from Tap-specific things — e.g. I might try switching to `tape` in future to fix conflicting deps like `esm`, and `tape` is much more minimalistic in design and encourages making all helpers explicit rather than augmenting the prototype. And if we ever consider switching to a different testing ecosystem (e.g. to Jest), this will make it easier.

vakila: OK, fair enough - I personally would wait until we actually _do_ make the change away from tap, but if you think that'll be coming down the pike soon then sure. No biggie either way, anyway :smile:

PR #8337, opened 2019-06-11
Update Flow ignores for faster start

mourner:
Makes Flow startup ~4.5 times faster by updating Flow ignore list with the worst offenders that we should ignore (mostly from non-direct deps in `node_modules`).

With this PR, `npx flow check` takes 17 seconds instead of 76. `npx flow ls | wc -l` shows the number of files Flow processes — the PR brings it down from 40565 to 9213.


.flowconfig
```diff
@@ -1,20 +1,50 @@
...
-.*/node_modules/htmltojsx/.*
+.*/node_modules/postcss.*/.*
+.*/node_modules/prismjs.*/.*
 .*/node_modules/documentation/.*
 .*/node_modules/module-deps/.*
-.*/node_modules/bcryptjs/.*
-.*/node_modules/@mapbox/geojson-types/fixtures/.*
 .*/test/unit/style-spec/fixture/invalidjson.input.json
-.*/test/integration/render-tests/.*
+.*/render-tests/.*
```
asheemmamoowala: The `render-tests`, `query-tests`, and `expression-tests` folders are all in `test/integration`. Can that folder be ignored, but the `integration/lib` folder be included?

mourner: I'm not sure how `include`/`ignore` sections in Flow config interact, but I'd rather have it one way (and a more explicit one), not mixed, for clarity and less chance of accidental mistakes in the future.

PR #8356, opened 2019-06-15
Add mapbox-gl-framerate to plugins list

tristen:
Quick addition that adds [`mapbox-gl-framerate`](https://github.com/mapbox/mapbox-gl-framerate) to the plugins list.

### DIFF
--- a/docs/pages/plugins.js
+++ b/docs/pages/plugins.js
@@ -155,6 +155,10 @@ const plugins = {
         "mapbox-gl-fps": {
             "website": "https://github.com/MazeMap/mapbox-gl-fps",
             "description": "A frames-per-seconds GUI control and measurer with statistic report output."
+        },
+        "mapbox-gl-framerate": {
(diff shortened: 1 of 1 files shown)


tristen:
Oh hmm, should the base branch be different?

andrewharvey:
> Oh hmm, should the base branch be different?

Yes, if you target `publisher-production` then the change can go live when merged instead of next release, per https://github.com/mapbox/mapbox-gl-js/blob/master/docs/README.md#committing-and-publishing-documentation.

andrewharvey:
I tried to change the base branch for you in GitHub, but it added a whole bunch of other commits, I changed it back and it seems just like before, hope I didn't break anything.

tristen:
@andrewharvey ah yeah thanks. I think I needed to rebase with `git rebase --onto publisher-production master` and force push before setting the base branch.

tristen:
\o/

PR #8363, opened 2019-06-18
[WIP] use browser `Cache` interface to cache tiles

ansis:
The per-session token recently added to each tile request defeats the browser's caching. This adds caching that ignores the token using `Cache` interface.

- [x] push eviction improvements
- [x] write in-browser tests that can be run in all the browsers
- [ ] document challenges encountered and potential future improvements

## Launch Checklist

 - [x] post benchmark scores

### DIFF
--- a/src/util/tile_request_cache.js
+++ b/src/util/tile_request_cache.js
@@ -3,8 +3,13 @@
 import { parseCacheControl } from './util';
 import window from './window';
 
+import type Dispatcher from './dispatcher';
+
 const CACHE_NAME = 'mapbox-tiles';
-const CACHE_LIMIT = 50;
+let cacheLimit = 500; // 50MB / (100KB/tile) ~= 500 tiles
+let cacheCheckThreshold = 50;
+
+const MIN_TIME_UNTIL_EXPIRY = 1000 * 60 * 7; // 7 minutes. Skip caching tiles with a short enough max age.
 
 export type ResponseOptions = {
     status: number,
@@ -12,9 +17,8 @@ export type ResponseOptions = {
     headers: window.Headers
 };
 
-export function cachePut(request, response, cacheHeaders, requestTime) {
+export function cachePut(request: Request, response: Response, requestTime: number) {
     if (!window.caches) return;
-    console.log('put');
 
     const options: ResponseOptions = {
         status: response.status,
@@ -28,20 +32,23 @@ export function cachePut(request, response, cacheHeaders, requestTime) {
         return;
     }
     if (cacheControl['max-age']) {
-        options.headers.set('Expires', new Date(requestTime + cacheControl['max-age'] * 1000).toUTCString())
-        console.log(options.headers.get('Expires'));
+        options.headers.set('Expires', new Date(requestTime + cacheControl['max-age'] * 1000).toUTCString());
     }
 
-    const clonedResponse = new window.Response(response.clone().body, options);
+    const timeUntilExpiry = new Date(options.headers.get('Expires')).getTime() - requestTime;
+    if (timeUntilExpiry < MIN_TIME_UNTIL_EXPIRY) return;
 
-    window.caches.open(CACHE_NAME)
-        .then(cache => {
-            cache.put(request, clonedResponse).then(() => {
-            });
-        });
+    const clonedResponse = new window.Response(response.body, options);
+
+    window.caches.open(CACHE_NAME).then(cache => cache.put(stripQueryParameters(request.url), clonedResponse));
 }
 
-export function cacheGet(request, callback) {
+function stripQueryParameters(url: string) {
+    const start = url.indexOf('?');
+    return start < 0 ? url : url.slice(0, start);
+}
+
+export function cacheGet(request: Request, callback: (error: ?any, response: ?Response, fresh: ?boolean) => void) {
     if (!window.caches) return callback(null);
 
     window.caches.open(CACHE_NAME)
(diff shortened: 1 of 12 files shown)


src/util/tile_request_cache.js
```diff
@@ -0,0 +1,97 @@
...
+    return expires > Date.now() && !cacheControl['no-cache'];
+}
+
+
+let globalEntryCounter = Infinity;
+
+export function cacheEntryPossiblyAdded(dispatcher: Dispatcher) {
+    globalEntryCounter++;
+    if (globalEntryCounter > CACHE_CHECK_THRESHOLD) {
+        dispatcher.send('enforceCacheSizeLimit');
```
mourner: Trying to follow the flow here... Is the cache shared between the main thread and the workers? If so, why do we need to send a signal to enforce the size limit to a random worker (through dispatcher)? Or is this because of cache operations blocking the main thread and affecting performance? If so, can we add a comment about this?

src/util/tile_request_cache.js
```diff
@@ -0,0 +1,97 @@
...
+    if (cacheControl['max-age']) {
+        options.headers.set('Expires', new Date(requestTime + cacheControl['max-age'] * 1000).toUTCString())
+    }
+
+    const clonedResponse = new window.Response(response.body, options);
+
+    window.caches.open(CACHE_NAME)
+        .then(cache => {
+            cache.put(request, clonedResponse).then(() => {
+            });
```
mourner: Leftover empty callback?

src/util/tile_request_cache.js
```diff
@@ -0,0 +1,97 @@
...
+
+    window.caches.open(CACHE_NAME)
+        .catch(callback)
+        .then(cache => {
+            cache.match(request, { ignoreSearch: true })
+                .catch(callback)
+                .then(response => {
+                    const fresh = isFresh(response);
+
+                    // reinsert into cache so
```
mourner: Unfinished comment

src/util/tile_request_cache.js
```diff
@@ -0,0 +1,97 @@
...
+
+function isFresh(response) {
+    if (!response) return false;
+    const expires = new Date(response.headers.get('Expires'));
+    const cacheControl = parseCacheControl(response.headers.get('Cache-Control') || '');
+    return expires > Date.now() && !cacheControl['no-cache'];
+}
+
+
+let globalEntryCounter = Infinity;
```
mourner: Is this `Infinity` to enforce cache size limit on an initial page load (anticipating already existing cache)? Might need a comment too.

ansis: Added a comment in 54ac863d074268a67d87d4201db44b21e9d75c1b

src/util/tile_request_cache.js
ansis: finished in 54ac863d074268a67d87d4201db44b21e9d75c1b

src/util/tile_request_cache.js
ansis: removed in 9033e2850cc8088156842166c435838eb80cf743

src/util/tile_request_cache.js
ansis: Yep, blocking. I'm not fully confident whether it is misleading profiling or a real issue but I think it's safer to do it on the worker for now. I'm seeing it sometimes and not seeing it other times. Added a comment 54ac863d074268a67d87d4201db44b21e9d75c1b

src/util/tile_request_cache.js
```diff
@@ -0,0 +1,95 @@
...
+        headers: new window.Headers()
+    };
+    response.headers.forEach((v, k) => options.headers.set(k, v));
+
+    const cacheControl = parseCacheControl(response.headers.get('Cache-Control') || '');
+    if (cacheControl['no-store']) {
+        return;
+    }
+    if (cacheControl['max-age']) {
+        options.headers.set('Expires', new Date(requestTime + cacheControl['max-age'] * 1000).toUTCString())
```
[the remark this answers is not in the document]
ansis: Implemented in 
d936bf7

ansis:
> I also agree about needing control over custom caching that ignores query parameters — we should definitely not enforce it for non-Mapbox resources, because some servers will have query-based tile url scheme — e.g. example.com/tile?z=10&x=100&y=300. Perhaps let's only hardcode-enable this for Mapbox resources, and think about ways to explicitly turn this on for other sources programmatically later (e.g. in a different PR)?

Yep. I've changed this in https://github.com/mapbox/mapbox-gl-js/pull/8363/commits/de81a8eae6a305a740955ab3f19d645faffcf3b7 (messy commit with some other changes as well). It now runs the caching only when the url is both:
- mapbox, and
- contains `sku=`

ansis:
Some notes on things I've learned along the way:

- `If-None-Modified` is not considered an allowed header for simple cors requests. Using it to revalidate expired tiles results in the cors with preflight requests which doesn't work for us.
- Aborting requests after adding them to cache raises errors and crashes the firefox tab
- Adding a key that already exists in the cache modifies the order in some browsers but no others

The current implementation only runs when the fetch api is used. Due to a limitation with aborting fetch requests in safari this doesn't run in safari. Mobile safari doesn't work on browserstack due to the same fetch api limitation. But mobile safari works on my device?! I'm not sure exactly what is going on.

This has been also tested to work correctly in browsers without support for this api.

andrewharvey:
This is sneaky, but I understand why it's needed, it just had me pulling my hair out wondering why I couldn't see any network requests for tiles in Chrome's DevTools under Network... Until I realised this bypasses that entirely and uses the "Cache" interface.

Issue #6331, opened 2018-03-14
Allow for different Map instances to use different Mapbox access tokens


anandthakker:
## Motivation

Currently, a Mapbox access token must be set on the global `mapboxgl` library object.  There are some use cases that require multiple maps on a page, each using a different access token.

## Design Alternatives

1. Do nothing

It's technically possible to use the Map's [`transformRequest` option](https://www.mapbox.com/mapbox-gl-js/api/#map) to get different maps using different access tokens via something like:

```js
mapboxgl.accessToken = '<access token>' // placeholder
var map1 = new mapboxgl.Map({
    container: container,
    style: style,
    transformRequest: (url) => ({url: url.replace('<access token>', token_1)})
});
var map2 = new mapboxgl.Map({
    container: container,
    style: style,
    transformRequest: (url) => ({url: url.replace('<access token>', token_2)})
});
```

2. Accept `accessToken` as a `Map` option that takes precedence over `mapboxgl.accessToken`.

## Design

While option 1 above works, I think it would be better to address this use case more directly via option 2.

### Mock-Up

```js
mapboxgl.accessToken = token_1;
var map1 = new mapboxgl.Map({ // defaults to `mapboxgl.accessToken`
    container: container,
    style: style,
});
var map2 = new mapboxgl.Map({
    container: container,
    style: style,
    accessToken: token_2 // overrides `mapboxgl.accessToken`
});
```

stevage:
It's also arguably possible for someone to want to have two different Mapbox tokens in the *same* map, to mash up layers from two different accounts. Assuming that doesn't violate any Mapbox policy, might be worth supporting at the same time.

andrewharvey:
> It's also arguably possible for someone to want to have two different Mapbox tokens in the same map, to mash up layers from two different accounts. Assuming that doesn't violate any Mapbox policy, might be worth supporting at the same time.

Licensing aside, I assume you know that by making the tileset public, then it's no longer tied to an access token from that account and you can use any token from any account. Of course this only works when you're happy to make it public.

asheemmamoowala:
Closed by #8364

3zzy:
@andrewharvey I have the exact same use case where I need to show maps from different accounts using their accessToken and the maps cannot be made public. I guess my only option is to set the token on the global object?

andrewharvey:
[...]

PR #8364, opened 2019-06-18
map-specific access token

peterqliu:
## Launch Checklist

solves https://github.com/mapbox/mapbox-gl-js/issues/6331

Adds an `accessToken` parameter to Map options, that optionally overwrites the token specified in mapboxgl.accessToken for just that map.

 - [ ] post benchmark scores


test/unit/ui/map.test.js
```diff
@@ -50,6 +50,19 @@ test('Map', (t) => {
         t.end();
     });
 
+    // t.test('map-specific token', (t) => {\
```
peterqliu: How can we instantiate a map with a bad token?

Would like to add a test for the positive case, where it starts with `mapboxgl.accessToken = badtoken` but gets overwritten by a valid token inside the map options.

debug/token.html
```diff
@@ -0,0 +1,35 @@
...
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+
+<body>
+<div id='map'></div>
+
+<script src='/dist/mapbox-gl-dev.js'></script>
+<script src='/debug/access_token_generated.js'></script>
```
asheemmamoowala: Better to exercise this in the `multiple.html` debug page.

test/unit/ui/map.test.js
asheemmamoowala: > How can we instantiate a map with a bad token?
The unit tests don't exercise any access token specific logic. 

Map options are set-only with no ability to read back to options that were used to instantiate a map object. Since the access token is passed through to the `RequestManager` unit tests are probably suited to the behavior of that class.

src/util/mapbox.js
```diff
@@ -42,8 +42,10 @@ export class RequestManager {
     _skuTokenExpiresAt: number;
     _transformRequestFn: ?RequestTransformFunction;
 
-    constructor(transformRequestFn?: RequestTransformFunction) {
+    constructor(transformRequestFn?: RequestTransformFunction, customAccessToken?: string) {
         this._transformRequestFn = transformRequestFn;
+        if (customAccessToken) config.ACCESS_TOKEN = customAccessToken;
```
asheemmamoowala: `config.ACCESS_TOKEN` is used directly to read the `mapboxgl.accessToken` value. Modfiying it here will modify the `mapboxgl.accessToken` and `config.ACCESS_TOKEN` value for all other maps in the same page. The map-specific access token needs to be a member variable of the `RequestManager`

peterqliu: >The map-specific access token needs to be a member variable of the RequestManager

This would work well, but turns out that both `RequestManager` and `TelemetryEvent` need this token. The former is invoked by map.js, which is great because it directly handles the custom token from the options object. 

However, the latter is invoked before the map itself gets added, by its child class `TurnstileEvent` here https://github.com/mapbox/mapbox-gl-js/blob/3474c64ba2987d92a2bf38aec24a30a7bb69e155/src/util/mapbox.js#L484. Where is a good place to stow the custom token so that `TelemetryEvent` can refer to it at `event` time?

peterqliu: cc @ryanhamley ^ you might have some insight, having worked on turnstile stuff

ryanhamley: You can store it in the `RequestManager` and then reference it in the `postTurnstileEvent` function since everywhere that `postTurnstileEvent` is called also has access to `RequestManager`. You can see that [I did the same thing with the `postMapLoadEvent`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/source/vector_tile_source.js#L77) which needed the SKU token.

debug/token.html
peterqliu: that would definitely make less work for this PR, but debugging this requires breaking a few of the maps on purpose that I think would make `multiple.html` less useful for its other intended purposes? 😄

src/util/mapbox.js
```diff
@@ -448,8 +447,8 @@ export class TurnstileEvent extends TelemetryEvent {
             this.fetchEventData();
         }
 
-        const tokenData = parseAccessToken(config.ACCESS_TOKEN);
-        const tokenU = tokenData ? tokenData['u'] : config.ACCESS_TOKEN;
+        const tokenData = parseAccessToken(RequestManager._customAccessToken || config.ACCESS_TOKEN);
```
ryanhamley: This isn't going to work. `_customAccessToken` isn't a static property; it's set in the constructor function so it will only be available on instances of `RequestManager` and `RequestManager._customAccessToken` will always be `undefined`. Why not use the `customAccessToken` parameter being passed into the function? Or even create `this._customAccessToken` in the `TurnstileEvent` constructor and set it in `postTurnstileEvent` so you can reference it in `processRequest` instead of passing the variable down the function chain? That's what I did with the `MapLoad` event needing `skuToken`.

peterqliu:
bit stumped by the no-unused-vars lint error -- the flagged variable is used at https://github.com/mapbox/mapbox-gl-js/pull/8364/files#diff-3bdcadb35be47763a1abccc9707d22e0R375

src/util/mapbox.js
```diff
@@ -183,7 +185,7 @@ const normalizeTileURL = function(tileURL: string, sourceURL?: ?string, tileSize
         urlObject.params.push(`sku=${skuToken}`);
     }
 
-    return makeAPIURL(urlObject);
+    return makeAPIURL(urlObject, customAccessToken);
```
asheemmamoowala: Line 182/184 above will not add the `sku` parameter to the url if `config.ACCESS_TOKEN` is not provided. With the per-Map access token option, it is foreseeable that someone uses it without setting `mapboxgl.accessToken`.

src/index.js
```diff
@@ -70,6 +70,7 @@ const exported = {
     },
 
     set baseApiUrl(url: string) {
+        console.log('setting');
```
ryanhamley: console.log

ryanhamley:
>IS there a reason to separate the implementation of the `normalize...URL` methods from the `RequestManager`? This PR would be a lot simpler if those were flattened, and all access to a token was through a `RequestManager#accessToken()` method.

@asheemmamoowala I'm not sure I follow. the `normalize...URL` methods aren't being separated from `RequestManager` in this PR. Flattening them is hard because they have the same or similar signatures and it would be hard to know which method to call internally based on inputs. You could definitely have a `RequestManager#accessToken()` getter/setter but it would have to be an instance method which doesn't really help in getting the token in the `normalize...URL` methods.

src/util/mapbox.js
ryanhamley: `config.REQUIRE_ACCESS_TOKEN && (config.ACCESS_TOKEN || customAccessToken) && skuToken)` should work

asheemmamoowala:
> Flattening them is hard because they have the same or similar signatures and it would be hard to know which method to call internally based on inputs.

Why do the methods need to exist outside the `RequestManager` class? They are no longer exported, so the implementation could live within the `RequestManager` entirely.

> You could definitely have a RequestManager#accessToken() getter/setter but it would have to be an instance method which doesn't really help ...

The `RequestManager#accessToken` getter will need to be public so that access token can be shared with Telemetry event types

ryanhamley:
Ohhh I get what you're saying @asheemmamoowala. Yes, they could be moved into the `RequestManager` class. This PR didn't move them out of the class though. They just were never in it. Even if we moved the implementations into `RequestManager`, I'm not sure we could flatten them though.

peterqliu:
sounds like something we could tackle in a future PR 👍 @asheemmamoowala @ryanhamley do you see any other blockers to this one?

ryanhamley:
Agreed that we can refactor in a separate PR

peterqliu:
👍 done ^

PR #8365, opened 2019-06-19
Include Hiragana and Katakana code blocks for local glyph generation

asheemmamoowala:
This PR adds support for locally generating glyphs in the Hiragana and Katakana code blocks. This allows layers with mixed japaense syllabary to be able to use a single font. Without this change - kanji would use the locally generated glyphs but hiragana and katakana would use glyphs based on the style. 

The `localIdeographFontFamily` map option that is used to control this behavior is now not the best name given that Hiragana and Katakana are not ideographic. The name should be changed in a future API breaking release.

 - ~[ ] post benchmark scores~
 - [x] tagged @mapbox/gl-native  if this PR includes shader changes or needs a native port

cc @1ec5 @chloekraw @tristen

### DIFF
--- a/test/unit/render/glyph_manager.test.js
+++ b/test/unit/render/glyph_manager.test.js
@@ -60,3 +60,41 @@ test('GlyphManager generates CJK PBF locally', (t) => {
         t.end();
     });
 });
+
+test('GlyphManager generates Katakana PBF locally', (t) => {
+    t.stub(GlyphManager, 'TinySDF').value(class {
+        // Return empty 30x30 bitmap (24 fontsize + 3 * 2 buffer)
+        draw() {
+            return new Uint8ClampedArray(900);
+        }
+    });
+
+    const manager = new GlyphManager((url) => ({url}), 'sans-serif');
+    manager.setURL('https://localhost/fonts/v1/{fontstack}/{range}.pbf');
+
+    // Katakana letter te
+    manager.getGlyphs({'Arial Unicode MS': [0x30c6]}, (err, glyphs) => {
+        t.ifError(err);
+        t.equal(glyphs['Arial Unicode MS'][0x30c6].metrics.advance, 24);
+        t.end();
+    });
+});
+
+test('GlyphManager generates Hiragana PBF locally', (t) => {
+    t.stub(GlyphManager, 'TinySDF').value(class {
+        // Return empty 30x30 bitmap (24 fontsize + 3 * 2 buffer)
+        draw() {
+            return new Uint8ClampedArray(900);
+        }
+    });
+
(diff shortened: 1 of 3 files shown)


test/unit/render/glyph_manager.test.js
```diff
@@ -60,3 +60,41 @@ test('GlyphManager generates CJK PBF locally', (t) => {
...
+
+    // Katakana letter te
+    manager.getGlyphs({'Arial Unicode MS': [0x30c6]}, (err, glyphs) => {
+        t.ifError(err);
+        t.equal(glyphs['Arial Unicode MS'][0x30c6].metrics.advance, 24);
+        t.end();
+    });
+});
+
+test('GlyphManager generates Hiragan PBF locally', (t) => {
```
ryanhamley: Typo `Hiragan`

test/unit/render/glyph_manager.test.js
```diff
@@ -60,3 +60,41 @@ test('GlyphManager generates CJK PBF locally', (t) => {
...
+        // Return empty 30x30 bitmap (24 fontsize + 3 * 2 buffer)
+        draw() {
+            return new Uint8ClampedArray(900);
+        }
+    });
+
+    const manager = new GlyphManager((url) => ({url}), 'sans-serif');
+    manager.setURL('https://localhost/fonts/v1/{fontstack}/{range}.pbf');
+
+    //Hiragan letter te
```
ryanhamley: Same typo

test/unit/render/glyph_manager.test.js
```diff
@@ -60,3 +60,41 @@ test('GlyphManager generates CJK PBF locally', (t) => {
...
+    manager.getGlyphs({'Arial Unicode MS': [0x3066]}, (err, glyphs) => {
+        t.ifError(err);
+        t.equal(glyphs['Arial Unicode MS'][0x3066].metrics.advance, 24);
```
ryanhamley: I'm not sure I fully understand what the tests do. I get that you're stubbing the `draw` method of the static `TinySDF` object. But how does this test ensure that `TinySDF.draw` was called as opposed to going through the regular pathway? In either case, wouldn't we expect the glyph to be 24px wide?

asheemmamoowala: `GlyphManager#loadGlyphRange` would be the other code path, and that is not stubbed, so the test will error out if it tries to go down that path. The `metrics` test is to ensure that the code path called `_tinySDF` successfully and returned a `StyleGlyph` successfully.

ryanhamley: Ah ok, that makes sense.

PR #8380, opened 2019-06-22
fix tile debug rendering on satellite tiles

### LINKED ISSUE #2223: Tile number can't display on satellite style

andrewharvey:
## Launch Checklist

closes #2223 by improving the readability of the debug tiles (showTileBoundaries) on satellite tiles.

This change also increases the halo to 2px to improve readability, but I can change this back if you think it's now too bold.

Screenshot from 2019-06-22 14-02-04

I'll update the render test fixtures once we decide if the increase to 2px halo is okay or not.


ryanhamley:
The `debug/tile-overscaled` render test is failing now so you'll have to update that

tristen:
@andrewharvey thanks for making this change!

> This change also increases the halo to 2px to improve readability, but I can change this back if you think it's now too bold.

My 2c: It does feel a bit too bold.

andrewharvey:
> The debug/tile-overscaled render test is failing now so you'll have to update that.

I thought that was because I increased the halo to 2px, but it wasn't, the render test actually helped pickup an error in my code!

I've fixed that now but a new unit test I created for testing the debug tiles on satellite tiles, passes on my local machine but fails on circleci, it doesn't look like I can access circleci's artifacts from the job, so don't know what it's rendering on how it's different...

> If you want to finish this PR and then pick up #7314 that's fine with me. I won't get to it anytime soon so the help is appreciated.

Yep I think I'll try to get this one merged first, then I'll try to work in your changes from #7314.

asheemmamoowala:
> I've fixed that now but a new unit test I created for testing the debug tiles on satellite tiles, passes on my local machine but fails on circleci, it doesn't look like I can access circleci's artifacts from the job, so don't know what it's rendering on how it's different...

Here's a link to the results: https://45149-8629417-gh.circle-artifacts.com/0/root/mapbox-gl-js/test/integration/render-tests/index.html

andrewharvey:
Thanks for fixing that @asheemmamoowala, I could never work out why when I generated the fixture locally it looked fine and hence passed locally but for some reason that same test running on CI was different and hence failed. :thinking:

asheemmamoowala:
> I could never work out why when I generated the fixture locally it looked fine and hence passed locally 

If I remember correctly you run a non-OSX setup. I believe the difference is subtle in the graphics driver and happens to change the output just enough to go past the tolerance for render tests.

PR #8383, opened 2019-06-24
Add line-sort-key and fill-sort-key

ahk:
This pull request adds support for two new `style-spec` layout properties: `line-sort-key` and `fill-sort-key`. 

These properties mirror the `symbol-sort-key` in terms of function and allow for features of either `line` or `fill` type to specify their own sort order. Sort keys are expected to be float values in ascending order (back to front). Features with larger sort keys will be drawn in front of features with smaller sort keys.

Addition of the sort key property was motivated by cross-tile draw order issues for `symbol` and `circle` layers. `line` and `fill` do not suffer from the same cross-tile sorting issues, but these properties are still useful for designers and are supported for flexibility and consistency.

## Launch Checklist

 - [ ] post benchmark scores

cc @mapbox/studio @mapbox/map-design-team @mapbox/gl-native


src/style-spec/reference/v8.json
```diff
@@ -632,6 +632,21 @@
     }
   },
   "layout_fill": {
+    "fill-sort-key": {
+      "type": "number",
+      "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above features with a lower sort key.",
+      "sdk-support": {
+          "js": "1.0.0"
```
ansis: This should be 1.2.0

src/style-spec/reference/v8.json
```diff
@@ -850,6 +865,21 @@
       },
       "property-type": "data-constant"
     },
+    "line-sort-key": {
+      "type": "number",
+      "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above features with a lower sort key when they overlap. Features with a lower sort key will have priority over other features when doing placement.",
+      "sdk-support": {
+          "js": "1.0.0"
```
ansis: This should be 1.2.0 as well

tmpsantos:
> * tagged `@mapbox/gl-native` if this PR includes shader changes or needs a native port

Thanks for the heads up

andrewharvey:
:+1: 

Though it looks like this won't support sorting based on feature-state:

> `fill-sort-key: "feature-state" data expressions are not supported with layout properties.`

That's okay, but it would open up more use cases of this feature in the future if it was supported.

chloekraw:
> Though it looks like this won't support sorting based on feature-state. That's okay, but it would open up more use cases of this feature in the future if it was supported.

@andrewharvey my understanding is that the limitation you describe is rooted in the fact that feature-state broadly speaking has not been extended to layout properties, as opposed to a specific choice that we made for the implementation in this pr. but let me know if I'm missing something

andrewharvey:
:point_right: #8467

shayke:
Is there a reason why `feature-state` is not supported?
In our case, we color line features per state and it's mandatory to sort the features per status (selected ones should be on top). Currently we solve that by separating the layer into hovered, selected, normal lines to support hover -> selected -> normal z-index which kind of defeats the purpose of `feature-state` anyway (since before all you could do is separate layers and apply filters).

andrewharvey:
@shayke I believe the reason is in https://github.com/mapbox/mapbox-gl-js/pull/8383#issuecomment-507943718, it's just because no one has implemented feature-state for layout properties. Your workaround is exactly what I would suggest.

PR #8386, opened 2019-06-24
Bump `esm` package version

tmpsantos:
Fix security vulnerability pointed out by GitHub


mourner:
More background on why this was closed: unfortunately `esm` versions after `3.0.84` break our testing setup, and upgrading is difficult (I attempted it several times). However, the "vulnerability" GitHub warns us about doesn't affect us in any way so can be dismissed. We will address this eventually but it's not a priority.

Issue #8419, opened 2019-07-02
Benchmarks server is broken on master and publisher-production


ryanhamley:
**mapbox-gl-js version**:
current `master` and `publisher-production`

**browser**:
n/a

### Steps to Trigger Behavior

 1. run `yarn run start-bench`

### Link to Demonstration
n/a

### Expected Behavior
The benchmarks server should run without error

### Actual Behavior
on `master`, the server will often fail while complaining that the Javascript heap is out of memory
<img width="2990" alt="Screen Shot 2019-07-01 at 6 12 20 PM" src="https://user-images.githubusercontent.com/4523080/60475378-de61f800-9c2b-11e9-89b6-9557e587f84e.png">

on `publisher-production`, the server will die with a React DOM error
<img width="920" alt="Screen Shot 2019-07-01 at 6 13 26 PM" src="https://user-images.githubusercontent.com/4523080/60475396-f9346c80-9c2b-11e9-9312-10dab1055d0c.png">

ryanhamley:
cc @mapbox/docs I believe the `publisher-production` issue is related to the update to React DOM in https://github.com/mapbox/mapbox-gl-js/pull/8362

mourner:
While I'm investigating the out of memory issue, for us to fix the React-DOM-related update, I'll need to push for a new release of `buble` and `rollup-plugin-buble` that include my fix in https://github.com/bublejs/buble/pull/194 — afterwards we'll be able to update deps freely without build issues.

mourner:
@ryanhamley can you say more about the out-of-memory issue — e.g. provide your version of Node, a full log of the error, make sure Yarn was run, etc.? I can't reproduce on my machine with Node v10.16.0.

ryanhamley:
@mourner I can produce it on Node 8.11.4 with Yarn 1.16.0 when running `yarn run start-bench` or `yarn run watch-benchmarks`. It doesn't happen with `yarn run build-benchmarks` so it's obviously something with the view.

Full stack trace
<img width="3004" alt="Screen Shot 2019-07-02 at 11 05 19 AM" src="https://user-images.githubusercontent.com/4523080/60535630-6e9a4e80-9cb9-11e9-9108-3aa3ac75b523.png">

mourner:
@ryanhamley yeah, I managed to reproduce after switching to Node v8. Can you try https://github.com/mapbox/mapbox-gl-js/pull/8422?

ryanhamley:
[...]

PR #8422, opened 2019-07-02
Optimize benchmark_view build & update rollup deps

mourner:
Fixes half of #8419 (Node v8 running out of memory when running `yarn run start-bench`) by simplifying the view config, excluding all the plugins we use for the rest of the source and only using what's necessary. Also updates Rollup dependencies just in case.

Update: a side benefit is the `benchmarks_view` build time gets down from 17 to 6 seconds.


ryanhamley:
This looks good now. I don't get the memory error and the full benchmark suite works as expected.

Two things I noticed:
1) a `FunctionConvert` test that doesn't run against `master` or the current branch
2) all the benchmarks are running against `v1.0.0` which seems weird given that we released `v1.1.0` last night

mourner:
> The FunctionConvert test only running on 1/3 versions seems to be the same on master so I think we can ticket/investigate that separately from this PR

That bench was removed — it was just shown for v1.0.0 but we've fixed that to be v1.1.0 now.

PR #8434, opened 2019-07-05
Manually strip urls for faster cache matching

### LINKED ISSUE #8431: 1.1.0 release tile load tiems performs poorly compared to 0.54
I had been using the v0.54 release and have a version of a map I'd been deving here:
https://app-dev.simplelayers.com/~staging/apps/public/simple_map/8406

The same map using 1.1.0
https://app-dev.simplelayers.com/~art/apps/public/simple_map/8406

I know you prefer a simple example, but hopefully the same complex environment with minor differences can show the performance hit.

I know you are switching up the token scheme in the new version so I don't know if there is added latency in that process or what, but the performance difference is uncomfortably noticable.

**mapbox-gl-js version**:
0.54. vs. 1.1.0
**browser**:
I have been testing in Chrome on Windows 10
### Steps to Trigger Behavior
It really shows if you pitch and rotate the map so new tiles have to load. The 0.54 link will render them zippily but the 1.1.0 will take what seem-like a second per new tile.

 1. Load the map you'll see the roads+sat map and then a image with different polygons will appear
 2. Right-click pitch and rotate the map
Note the speed at which tiles fill in the view.

Do that for both links provided and note the performance difference.
The 1.1.0 performance is non-viable., the .54 performance was excellent.

mourner:
Closes #8431. A workaround for a [known Chrome performance issue](https://bugs.chromium.org/p/chromium/issues/detail?id=682677) with `match` calls that use `ignoreSearch: true`.

### DIFF
--- a/src/util/tile_request_cache.js
+++ b/src/util/tile_request_cache.js
@@ -73,17 +73,20 @@ function stripQueryParameters(url: string) {
 export function cacheGet(request: Request, callback: (error: ?any, response: ?Response, fresh: ?boolean) => void) {
     if (!window.caches) return callback(null);
 
+    const strippedURL = stripQueryParameters(request.url);
+
     window.caches.open(CACHE_NAME)
         .catch(callback)
         .then(cache => {
-            cache.match(request, { ignoreSearch: true })
+            // manually strip URL instead of `ignoreSearch: true` because of a known
+            // performance issue in Chrome https://github.com/mapbox/mapbox-gl-js/issues/8431
+            cache.match(strippedURL)
                 .catch(callback)
                 .then(response => {
                     const fresh = isFresh(response);
 
                     // Reinsert into cache so that order of keys in the cache is the order of access.
                     // This line makes the cache a LRU instead of a FIFO cache.
-                    const strippedURL = stripQueryParameters(request.url);
                     cache.delete(strippedURL);
                     if (fresh) {
                         cache.put(strippedURL, response.clone());


mourner:
> Is there a way to measure this performance improvement?

I did this by putting `console.time`/`console.timeEnd` in `tile_request_cache.js` `cacheGet`. You can see the problem by browsing the `satellite.html` debug page for a while — eventually cached tile loads start to take ~500ms each in Chrome. With this PR, it goes down to 20-50ms.

> Does it show up in the benchmarks or the unit tests?

No, since we don't exercise tile loading in benchmarks and don't have unit tests run in Chrome.

asheemmamoowala:
> I did this by putting console.time/console.timeEnd in tile_request_cache.js cacheGet. You can see the problem by browsing the satellite.html debug page for a while 

I tried this on Chrome and see the opposite results from what I would expect:

| Change | Avg. time (ms) |
| -- | -- |
| With no changes | 127.858224 | 
| With strippedURL + `ignoringSearch` | 157.0550753 |
| With this PR | 189.2290455 |

> **eventually** cached tile loads start to take ~500ms each in Chrome. With this PR, it goes down to 20-50ms.

The above numbers are for ~1000-1500 tiles in the cache. What range of cached tile size should the performance improvement apply to?

mourner:
@asheemmamoowala are you sure you're timing correctly? Note that you have to time by `request.url`, otherwise you might capture the wrong time invervals (relating to different tiles). I just opened `satellite.html` in master, adding `console.time('get' + request.url)` + `timeEnd` in the relevant place, and I'm seeing ~2500ms per tile. This is with just 525 entries. After switching to this branch, it gets down to `200-300ms`.

mourner:
> The above numbers are for ~1000-1500 tiles in the cache. What range of cached tile size should the performance improvement apply to?

The bigger the number, the bigger the perf difference, as you can read in the linked Chrome issue — time increases linearly with the amount of stored objects. We have a manual limit of 500 tiles for the cache, but the regression is still severe.

ansis:
> I did this by putting console.time/console.timeEnd in tile_request_cache.js cacheGet. You can see the problem by browsing the satellite.html debug page for a while — eventually cached tile loads start to take ~500ms each in Chrome. With this PR, it goes down to 20-50ms.

I'm seeing similar numbers! Looks great.

asheemmamoowala:
Was able to reproduce the results with pitched satellite maps. 

| Change | Avg. time (ms) |
| -- | -- |
| With no changes | 20000.858224 | 
| With strippedURL + `ignoringSearch` | 3000.0550753 |
| With this PR | 500.2290455 |

All of these are with a cache that is full. So the results are worst case scenario for frequent map users.

PR #8467, opened 2019-07-12
Add *-sort-key layout property for circle, fill, line

ahk:
This pull request adds support for three new style-spec layout properties: circle-sort-key, line-sort-key, and fill-sort-key.

These properties mirror the symbol-sort-key in terms of function and allow for features of circle, line, or fill type to specify their own sort order. Sort keys are expected to be float values in ascending order (back to front). Features with larger sort keys will be drawn in front of features with smaller sort keys.

Addition of the sort key property was motivated by cross-tile draw order issues for symbol and circle layers. Line and fill do not suffer from the same cross-tile sorting issues, but these properties are still useful for designers and are supported for flexibility and consistency.

 - [ ] post benchmark scores


src/data/bucket/line_bucket.js
```diff
@@ -119,8 +120,11 @@ class LineBucket implements Bucket {
         this.layers = options.layers;
         this.layerIds = this.layers.map(layer => layer.id);
         this.index = options.index;
-        this.features = [];
         this.hasPattern = false;
+        this.patternFeatures = [];
+
+        const sortKey = this.layers[0].layout.get('line-sort-key');
+        this.sortFeaturesByKey = sortKey.constantOr(1) !== undefined;
```
asheemmamoowala: I see this pattern of `constantOr` used elsewhere, but don't understand how/when it is expected to evaluate to `false` if `constantOr` returns `1` as the fallback value

More importantly, what is the need for this instance variable? In `populate()`, cant `lineSortKey` be used to determine if the features need to be sorted?

src/style-spec/reference/v8.json
```diff
@@ -850,6 +850,21 @@
       },
       "property-type": "data-constant"
     },
+    "line-sort-key": {
+      "type": "number",
+      "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above features with a lower sort key when they overlap. Features with a lower sort key will have priority over other features when doing placement.",
```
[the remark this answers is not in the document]
ahk: Nope, it's not relevant. Good catch.

ahk:
I think the flow-type errors related to `layout.get('circle-sort-key')` are true positives for a bug. `CircleBucket` supports both `CircleStyleLayer` and `HeatmapStyleLayer` as layer types (completely didn't notice the heatmap support in the bucket) and `HeatmapStyleLayer` doesn't have layout properties beyond visibility, and definitely not `circle-sort-key`. All lookups of layout properties within `CircleBucket` and `drawCircles` need to be cased on the layer type being used which seems a bit hacky, but I think is the best path forward. There's still the majority of the bucket and draw logic being shared, and I think appropriately shared.

src/data/bucket/line_bucket.js
ahk: Removed the instance variable in CircleBucket and will modify this patchset soon to remove from Fill and Line. `constantOr` I believe could return `undefined` as checked here if the value is a constant, and that value is `undefined` though I'm not sure why that would happen.

PR #8472, opened 2019-07-12
fix race condition between caching and aborting

### LINKED ISSUE #8470: Regression: unbounded memory growth on the worker side since v1.1.0
**mapbox-gl-js version**: v1.1.0
**browser**: Chrome

### Steps to Trigger Behavior

 1. Browse a simple map (e.g. `/debug`), quickly zooming back and forth.
 2. Watch workers memory footprint in the "Memory" tab.

### Expected Behavior

Memory fluctuates around a certain size.

### Actual Behavior

Memory grows indefinitely.

Since v1.1.0, quick map browsing introduces some race condition that makes some tiles to be retain on the worker side in `this.loaded` object of `VectorTileWorkerSource` indefinitely. You can track this by setting `mapboxgl.workerCount = 1`, setting Map `maxTileCacheSize` to `1`, and adding `console.log(Object.keys(this.loaded).length)` after `this.loaded[uid] = workerTile` in `vector_tile_worker_source.js`.

I bisected this down to #8363 @ansis — something in there introduced a race condition with worker cache clearing logic. Let's investigate further.

ansis:
If a request was aborted while it was checking the cache the request would still complete leading to a parsed tile that would never get cleared. The fix is to not call back when the request is aborted.

fix #8470

Any idea if I should / how I could add a test for this?

## Launch Checklist

### DIFF
--- a/src/util/ajax.js
+++ b/src/util/ajax.js
@@ -102,6 +102,7 @@ function makeFetchRequest(requestParameters: RequestParameters, callback: Respon
         signal: controller.signal
     });
     let complete = false;
+    let aborted = false;
 
     const cacheIgnoringSearch = hasCacheDefeatingSku(request.url);
 
@@ -110,6 +111,8 @@ function makeFetchRequest(requestParameters: RequestParameters, callback: Respon
     }
 
     const validateOrFetch = (err, cachedResponse, responseIsFresh) => {
+        if (aborted) return;
+
         if (err) {
             // Do fetch in case of cache error.
             // HTTP pages in Edge trigger a security error that can be ignored.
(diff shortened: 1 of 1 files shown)


src/util/ajax.js
```diff
@@ -172,6 +176,7 @@ function makeFetchRequest(requestParameters: RequestParameters, callback: Respon
     }
 
     return { cancel: () => {
+        aborted = true;
         if (!complete) controller.abort();
```
asheemmamoowala: Can't the controller's abort signal be used directly to reject or escape from `validateOrFetch` early instead of using an additional `aborted` variable?

ansis: from discussion in chat: let's leave this for now

Issue #8397, opened 2019-06-26
setStyle updates `text-variable-anchor` inconsistently


samanpwbb:
**mapbox-gl-js version**: 1.0.0

**browser**: any

### Steps to Trigger Behavior

 1. Render a mapbox gl map with a text-variable-anchor value
 2. Use setStyle to apply new text-variable-anchor values 

The result is that Studio cannot provide a realistic preview of the user's style as the user changes the `text-variable-anchor` value.

### Link to Demonstration

https://jsbin.com/riwasomula/1/edit?html,output

### Expected Behavior

 Expect label placement to change when preferred text anchor changes.

### Actual Behavior

text placement does not change.

samanpwbb:
Note that `{ diff: false }` does properly update text position – but this shouldn't be necessary.

peterqliu:
@samanpwbb I can't reproduce the bug in the linked test case 🤔 changes are subtle, but if we disable the color change and bump the interval to 1000, labels are clearly updating:

<img src='https://media.giphy.com/media/KDbWMslRUv4sXz5oQy/giphy.gif'>

samanpwbb:
> @samanpwbb I can't reproduce the bug in the linked test case 🤔 changes are subtle, but if we disable the color change and bump the interval to 1000, labels are clearly updating:

Yes, there are some subtle changes. I would expect label placement after a setStyle to be deterministic though. Right now, results of setStyle vary based on past label placement calculations. It appears as though once a given label finds a position that works, it will "prefer" that position in the future. Would it be possible to reset the "memory" of past positions after setStyle calls that effect `text-variable-anchor`?

dereklieu:
This is a pretty simplistic example, but I think it helps illustrate why someone using this property in Studio might think the current behavior is buggy. I made the screengrab on `0.54.x` but confirmed it on `1.1.0` as well.

**Regular text-anchor**
![text-anchor](https://user-images.githubusercontent.com/1590802/60931376-4210a480-a26e-11e9-9a02-764555ae0fe0.gif)

**text-variable-anchor**
![text-variable-anchor](https://user-images.githubusercontent.com/1590802/60931385-489f1c00-a26e-11e9-9da0-3d968ca70040.gif)

asheemmamoowala:
The `styleDiff` is working correctly. It generates the correct `setLayoutProperty` change operation in the jsbin provided.

The problem (I think) stems from preferring previous placement in: 

[...]

Issue #8445, opened 2019-07-08
De-duplicate array enums for symbol layer properties when parsing the style


asheemmamoowala:
Moved from https://github.com/mapbox/mapbox-gl-js/pull/8399#discussion_r299234134

`text-variable-anchor` values are not current de-duped.
In #8399, `text-writing-mode` values are de-duped for every bucket.

The array of values for `text-variable-anchor` and `text-writing-mode` should be de-duplicated as part of parsing the style. This would do it just once, allow better style diffing, and reduce per-bucket work.

Issue #8370, opened 2019-06-20
Should validator accept same values for an array-enum property?


dereklieu:
**mapbox-gl-js version**: 0.54.0

### Question

👋 I'm working on supporting `text-variable-anchor` in Studio, and I noticed that this passed the validator.

![Screen Shot 2019-06-19 at 7 08 36 PM](https://user-images.githubusercontent.com/1590802/59813246-ad95d080-92c5-11e9-885d-a0e42613ec3c.png)

Should the style validator accept same values in an array-enum property? Asking because this feels like a bug, but I also don't know how much leeway is built into the validator for what I would consider low-priority things like this.

asheemmamoowala:
Ticketed in https://github.com/mapbox/mapbox-gl-js/issues/8370

PR #8473, opened 2019-07-13
Favor previous anchor only when still in the `text-variable-anchor` options

asheemmamoowala:
Addresses #8397 

The `text-variable-anchor` implementation favored the previous anchor placement for symbols, even if the previous anchor was no longer an option in the `text-variable-anchor` value (when changed by `setStlye` or `setlayoutProperty`)

The benchmarks are unaffected because they don't exercise the `text-variable-anchor` property codepath. They use streets-v10 style which predates this property.

This PR also needs to be ported to gl-native:

https://github.com/mapbox/mapbox-gl-native/blob/a0e2064f75a229a0bc369ce023e2eaaccb6af074/src/mbgl/text/placement.cpp#L198-L202

cc @mapbox/studio  @mapbox/gl-native @vakila


asheemmamoowala:
> Prevent duplicate values: #8370, which would make step (2) from the second example impossible.

The de-duplication is also ticketed out in https://github.com/mapbox/mapbox-gl-js/issues/8445

> Always clear all preferences whenever text-variable-anchor changes via setPaintProperty / setStyle.

With https://github.com/mapbox/mapbox-gl-js/pull/8399 this will require clearing out the preferred placement for multiple properties.

pozdnyakov:
Port on gl-native: https://github.com/mapbox/mapbox-gl-native/pull/15129

PR #8479, opened 2019-07-15
Fix extent warning and clamp exceeding coords

### LINKED ISSUE #8477: Vector tile line rendering regression
After upgrading from version 1.0.0 to 1.1.0 we experienced a rendering regression while rendering lines from a vector tile. See screenshot below for comparison between the two versions. An extra almost vertical line is rendered that should not be there.

In #2854 a similar issue has been posted, but in our case we _don't_ get the warning `Geometry exceeds allowed extent`.

**mapbox-gl-js version**: 1.1.0

**browser**: latest Chrome, latest Firefox, latest Edge and latest Internet Explorer

### Screenshot

Screenshot of rendering with two versions

### Steps to trigger behavior

1. Visit https://jkoelewijn.github.io/issues/mapbox/mapbox-gl-js/line-rendering-regression/mapbox-gl-js-1.1.0.htm.
2. Observe that an extra almost vertical line is rendered compared to https://jkoelewijn.github.io/issues/mapbox/mapbox-gl-js/line-rendering-regression/mapbox-gl-js-1.0.0.htm.

### Source code of example

https://github.com/jkoelewijn/jkoelewijn.github.io/tree/master/issues/mapbox/mapbox-gl-js/line-rendering-regression

mourner:
Closes #8477:

- Fixes the extent warning, for which the threshold wasn't updated to reflect #8306 (we only have 15 bits for the coordinates again).
- Clamps the coordinates that exceed allowed extent so that when this does happen, it's unlikely to cause any artifacts.

## Launch Checklist

 - [ ] ~~document any changes to public APIs~~ no changes
 - [ ] ~~post benchmark scores~~ not necessary

### DIFF
--- a/src/data/load_geometry.js
+++ b/src/data/load_geometry.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { warnOnce } from '../util/util';
+import { warnOnce, clamp } from '../util/util';
 
 import EXTENT from './extent';
 
@@ -17,7 +17,7 @@ function createBounds(bits) {
     };
 }
 
-const bounds = createBounds(16);
+const bounds = createBounds(15);
 
 /**
  * Loads a geometry from a VectorTileFeature and scales it to the common extent
@@ -39,6 +39,8 @@ export default function loadGeometry(feature: VectorTileFeature): Array<Array<Po
 
             if (point.x < bounds.min || point.x > bounds.max || point.y < bounds.min || point.y > bounds.max) {
                 warnOnce('Geometry exceeds allowed extent, reduce your vector tile buffer size');
+                point.x = clamp(point.x, bounds.min, bounds.max);
+                point.y = clamp(point.y, bounds.min, bounds.max);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/data/load_geometry.js
```diff
@@ -39,6 +39,8 @@ export default function loadGeometry(feature: VectorTileFeature): Array<Array<Po
 
             if (point.x < bounds.min || point.x > bounds.max || point.y < bounds.min || point.y > bounds.max) {
                 warnOnce('Geometry exceeds allowed extent, reduce your vector tile buffer size');
+                point.x = clamp(point.x, bounds.min, bounds.max);
```
asheemmamoowala: Will clamping here change the current behavior (especially with respect to geometries split across tile boundaries? Or is it always broken when points are beyond the bounds?

mourner: It's broken. It's just that the bounds were increased 2x when we switched to bigger line layout, and got back when we reverted that. The clamping at minimum doesn't get things worse, but also often makes things better.

PR #8490, opened 2019-07-16
fix #8481, unbounded memory growth in IE

### LINKED ISSUE #8481: Regression: unbounded memory growth on the worker side in IE11 since v0.53.0
**mapbox-gl-js version**: v0.53.0+
**browser**: IE11

Very similar issue to #8470, but with a different root cause in a different code path — some tile data gets retained indefinitely on the worker side when the XHR code path is used (the fetch one is fixed in #8472). You can track this in Chrome by commenting out the `fetch` code path and then following the same steps as in #8470.

This regressed in v0.53.0, and I believe the most likely culprit is #7818.

ansis:
Cancellable requests to workers should never call back after they have been cancelled.

## Launch Checklist

### DIFF
--- a/src/util/actor.js
+++ b/src/util/actor.js
@@ -56,12 +56,16 @@ class Actor {
         }, buffers);
         if (callback) {
             return {
-                cancel: () => this.target.postMessage({
-                    targetMapId,
-                    sourceMapId: this.mapId,
-                    type: '<cancel>',
-                    id: String(id)
-                })
+                cancel: () => {
+                    // Set the callback to null so that it never fires after the request is aborted.
+                    this.callbacks[id] = null;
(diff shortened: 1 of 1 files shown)


src/util/actor.js
```diff
@@ -56,12 +56,16 @@ class Actor {
...
             return {
-                cancel: () => this.target.postMessage({
-                    targetMapId,
-                    sourceMapId: this.mapId,
-                    type: '<cancel>',
-                    id: String(id)
-                })
+                cancel: () => {
+                    // Set the callback to null so that it never fires after the request is aborted.
+                    this.callbacks[id] = null;
```
kkaefer: Why do we set it to null instead of deleting it?

Issue #8288, opened 2019-05-28
Support prefers-reduced-motion


CharlesBelov:
## Motivation

Please change your product so that prefers-reduced-motion is automatically supported. Specifically, if the browser indicates that the site visitor has specified that they don't want animation, flyTo, jumpTo, and easeTo would automatically not animate the transition, but simply display the new view. Chrome, Firefox, and Safari support this setting.
This is important because these effects can make some people, myself included, feel ill. That's probably not something you want associated with your product.

The advantage is that you can still provide animation for people who are not bothered by it without making people who are bothered by it feel ill.

## Design Alternatives

This could be accommodated by querying the CSS media query [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion). This is currently supported in Firefox, Chrome, and Safari.

The no-change option would be to expect websites to reduce animation for everyone or else leave people to learn to avoid a website if it makes them feel ill.

### Mock-Up

What will this design look like to developers?

The intent is that MapBox would implement this feature automatically so that developers do not need to be concerned with it.

What will this design look like to end users?

If an end user has set their operating system to reduce or eliminate animation, then maps will always go from one state to another with no movement transition.

If an end user has not set their operating system to reduce or eliminate animation, then maps will display with whatever movement transition the developer has programmed.

### Concepts

How will we teach this design?

Since this would be implemented universally, no training would be needed. That said, you might link to an article on the subjects such as https://alistapart.com/article/designing-safer-web-animation-for-motion-sensitivity/ (warning: some people have reported that the image at the top of this article triggers symptoms similar to those caused by animation)

What terminology will work best for the new concepts introduced by this design?

vestibular disorders
prefers-reduced-motion

What existing precedents support the new concepts?

https://www.drupal.org home page detects prefers-reduced-motion in supporting browsers and either animates or not depending on that setting.

Where do the concepts set new precedents?

None, although this is an emerging best practice.

### Implementation

How you would implement the design in Javascript?

If prefers-reduced-motion is set to reduce, 
then transition between two map states without animation.

How you would implement the design in C++?

Same pseudocode.

What parts of the Mapbox GL ecosystem will need to change to accommodate this design?

flyTo, jumpTo, and easeTo

Are there any important edge cases?

There is disagreement on how to handle browsers which do not support prefers-reduced-motion, such as IE or Edge. Some recommend treating such browsers as requesting no animation; some provide the animation.

[...]

PR #8494, opened 2019-07-16
Support for prefers-reduced-motion

arindam1993:

Addresses #8288. 
Defaults the animation duration to 0 when `prefer-reduced-motion` is set to `reduce`.

How to set this on your local machine:
https://developers.google.com/web/updates/2019/03/prefers-reduced-motion


src/util/browser.js
```diff
@@ -19,6 +19,8 @@ const cancel = window.cancelAnimationFrame ||
 
 let linkEl;
 
+let _reducedMotionQuery: MediaQueryList;
```
mourner: Nit: maybe remove the dangling `_` in the name for consistency? This will be private anyway since it's not exposed.

CharlesBelov:
@mourner: Actually, that is the whole point of prefersReducedMotion, that the developer can use as much animation as they want and the person who doesn't want to see animation will never see it but others who are okay with animation will see it. Yes, my original issue specifically mentioned:

What parts of the Mapbox GL ecosystem will need to change to accommodate this design?

flyTo, jumpTo, and easeTo

arindam1993:
Thanks for the quick review @mourner !
I interpreted the desired behavior to be as @CharlesBelov mentioned, a global `force-disabling` of all animations.

I added the change to `flyTo` as well!

andrewharvey:
> @mourner: Actually, that is the whole point of prefersReducedMotion, that the developer can use as much animation as they want and the person who doesn't want to see animation will never see it but others who are okay with animation will see it.
> 
> So, no, it's not appropriately a default, it is an override, similar to using !important on a user CSS style sheet.

I agree, the whole point is a developer can build with full animations and if the user asks for no animations via prefersReducedMotion, they don't need to do any checks themself, mapboxgl will automatically comply with the users request.

src/ui/camera.js
```diff
@@ -857,6 +857,18 @@ class Camera extends Evented {
...
+        // Fall throwugh to jumpTo is user has set prefers-reduced-motion
+        if (browser.prefersReducedMotion) {
+            this.jumpTo({
+                center: options.center,
+                zoom: options.zoom,
+                bearing: options.bearing,
+                pitch: options.pitch,
+                around: options.around
+            }, eventData);
+            return;
```
mourner: Should be `return this` (or `return this.jumpTo(...)`) for consistency/chaining.

arindam1993:
Thanks @mourner ! changes made 👍

src/ui/camera.js
```diff
@@ -857,6 +857,17 @@ class Camera extends Evented {
      * @see [Fly to a location based on scroll position](https://www.mapbox.com/mapbox-gl-js/example/scroll-fly-to/)
      */
     flyTo(options: Object, eventData?: Object) {
+        // Fall throwugh to jumpTo is user has set prefers-reduced-motion
```
asheemmamoowala: nit: typos

arindam1993:
> Is there a way to add a unit/render test for this?

I think I can write a test and model it after the `can be called from within a moveend event handler` in `camera.test.js`, it does asynchronous timing based things as well.

arindam1993:
Added unit tests 👍

CharlesBelov:
Possibly one more typo:
// Fall through to jumpTo is user has set prefers-reduced-motion
might correctly be
// Fall through to jumpTo if user has set prefers-reduced-motion

test/unit/ui/camera.test.js
```diff
@@ -869,6 +881,14 @@ test('camera', (t) => {
             }, 0);
         });
 
+        t.test('duration is 0 when prefers-reduced-motions: reduce is set', (t) => {
```
asheemmamoowala: nit: typo `motions`

dieterdreist:
I agree it can be very consistent to remove flyTo and easeTo animations, in certain circumstances (particularly, rather short animations, "eye gimmicks", when the animations are not required for functional reasons.
But if you use the animations as integral part of your app (e.g. to fly somewhere in many seconds, to give the user an overview over a bigger area with an automatic flight...), then this will break the app.

andrewharvey:
> But if you use the animations as integral part of your app (e.g. to fly somewhere in many seconds, to give the user an overview over a bigger area with an automatic flight...), then this will break the app.

But still the user has requested reduced motion, I would think that means no more than 1 frame per second. What about just reducing the animation to larger one second steps? That avoids animations but still gives the user a view of the the fly operation?

dieterdreist:
"reduced motion" is not about frames per second. It is about reducing "useless" motion (eye candy).

If you activate the function, would you expect your videos to playback with 1 frame per second?

Did you notice how Apple values this setting? They use less "effects" (e.g. no parallax effects or  icon animations on the home screen, or simpler text animations in the keyboard), but some UI elements like the drawers for notifications (from top) or settings (from bottom) animate just the same. The latter don't "jump" or pop up instantly, they smoothly move into the screen as if you hadn't requested reduced motion.

I had activated the reduced motion, because it offered my a cleaner and less distracting experience while supposedly also saving on battery. I read this often as advice in the web, it is not just a feature for people with visual problems. Core functionality should keep working.

If you still believe activating "reduced motion" means that any flyTo and easeTo animation in mapbox-gl must be converted into a jumpTo, then please make this customizable / a setting.

andrewharvey:
According to https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

> Indicates that user has notified the system that they prefer an interface that **minimizes the amount of movement or animation, preferably to the point where all non-essential movement is removed.**

My interpretation of this is it depends on the app, if you're building an animation with a slow ease fly to different locations, then I could consider that essential as it's core to the experience or app, if you're building an app where you search for a location and the map flys to that location, then I could consider that animation non-essential as you can just skip straight to the destination. Sure the animation is nicer and is a better UX, but still non-essential.

With that in mind it could be we expose some kind of option so the app developer can say if this transition is considered essential or not?

The smooth zoom in you get when you click the navigation zoom in button is another example, I would classify as non-essential and should respect prefers-reduced-motion.

CharlesBelov:
In short, the three issues that are covered by this are:
1) Motion sickness 
2) Attention disruption
3) Battery life

Distinguishing between essential motion and non-essential motion:

1) User has requested a different view - non-essential
2) User has requested an animation going from this view to that view - essential

So, a developer might offer two options:

1) Go to point B
2) Fly from point A to point B

The problem becomes if the developer doesn't give a choice and assumes the user wants the fly option.

So, yes, it would be fine if the developer can specify that a particular animation is essential, but it would not properly be the default and there needs to be a non-animating option.

As for the Apple UX drawer slide, it is non-essential, so I view it as a failure in implementation.

andrewharvey:
See the proposal at #8743.

CharlesBelov:
I have an update to my immediately previous comment.

Actually, with iOS 13, Apple has now added a separate preference, Settings > Accessibility > Motion > Prefer cross-fade transitions, described as "Reduce the motion for user interface controls that slide in when appearing and disappearing."

I'm not sure how they expose this to applications or how applications would in turn expose this to websites.

PR #8495, opened 2019-07-17
fix class toggling on control for IE

cs09g:
The second argument of `classList.toggle` is not supported in IE.
So, the control displays as disabled in IE at default.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11865865/

## Launch Checklist


cs09g:
I'm going to fix the convention from the ternary operator to if~else condition.

mourner:
Thanks!

PR #8502, opened 2019-07-17
set further near plane to fix depth precision for deckgl

### LINKED ISSUE #7573: Set better near/far planes
## Motivation

To reduce z-fighting effect when rendering 3d objects in custom layers

Reproduce steps: 
- Click on https://codepen.io/Pessimistress/pen/aQNKKg
- The left is rendered using deck.gl canvas, the right is rendering a deck.gl layer as custom Mapbox layer
- Rotate each map. The right one has visibly more flickering due to z-fighting.

## Cause

deck.gl's projection matrix has its near plane set to `0.1`, and far plane set to `10 * distanceToTopEdge`. This was implemented to match the projection matrix used prior to Mapbox 0.29: https://github.com/uber/deck.gl/blob/master/modules/core/src/viewports/web-mercator-viewport.js#L57

When using deck.gl with Mapbox's custom layer, we set the near plane to `1 / height` and far plane to `1 * distanceToTopEdge`, so that it matches the projection matrix used in Mapbox 0.50: https://github.com/uber/deck.gl/blob/master/modules/mapbox/src/mapbox-layer.js#L58

I wonder what was the motivation behind Mapbox's change in its projection matrix calculation. In our use cases, the old settings definitely perform better.

ansis:
fix #7573

This fixes precision issues in custom layers that use deckgl. While the old near plane worked well for our fill-extrusion layers it was causing issues in deckgl layers that do more complex projection in the shaders. This moves the near plane further away to a distance that seems to fix both the z-fighting in deckgl while not creating camera-close-to-building clipping issues.

deckgl would have to change `1 / deck.height ` to `1 / 50` here to match our change:
https://github.com/uber/deck.gl/blob/bddc36902a2535fa9430a1b450944f83cf4d9d11/modules/mapbox/src/deck-utils.js#L118

@Pessimistress can you confirm this would work for you?

### Testing
The [debug-z-fighting-deckgl](https://github.com/mapbox/mapbox-gl-js/tree/debug-z-fighting-deckgl) branch contains a debug page that can be used to test this. In the branch I've both patched deckgl and mapbox-gl-js. The third map pane shows both a deckgl layer and a fill-extrusion layer coexisting. To start the example: `yarn run start-debug` and then open http://localhost:9966/debug/

## Launch Checklist

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -537,9 +537,18 @@ class Transform {
         // Add a bit extra to avoid precision problems when a fragment's distance is exactly `furthestDistance`
         const farZ = furthestDistance * 1.01;
 
+        // The larger the value of nearZ is
+        // - the more depth precision is available for features (good)
+        // - clipping starts appearing sooner when the camera is close to 3d features (bad)
+        //
+        // Smaller values worked well for mapbox-gl-js but deckgl was encountering precision issues
+        // when rendering it's layers using custom layers. This value was experimentally chosen and
+        // seems to solve z-fighting issues in deckgl while not clipping buildings too close to the camera.
+        const nearZ = this.height / 50;
+
         // matrix for conversion from location to GL coordinates (-1 .. 1)
         let m = new Float64Array(16);
-        mat4.perspective(m, this._fov, this.width / this.height, 1, farZ);
+        mat4.perspective(m, this._fov, this.width / this.height, this.height / 50, farZ);
 
         mat4.scale(m, m, [1, -1, 1]);
         mat4.translate(m, m, [0, 0, -this.cameraToCenterDistance]);
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -548,7 +548,7 @@ class Transform {
 
         // matrix for conversion from location to GL coordinates (-1 .. 1)
         let m = new Float64Array(16);
-        mat4.perspective(m, this._fov, this.width / this.height, this.height / 50, farZ);
+        mat4.perspective(m, this._fov, this.width / this.height, nearZ, farZ);
 
         mat4.scale(m, m, [1, -1, 1]);
         mat4.translate(m, m, [0, 0, -this.cameraToCenterDistance]);
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1952,6 +1952,8 @@ class Map extends Camera {
      * @memberof Map
      * @var {string} version
      */
+
+     get version(): string { return version; }
 }
 
 export default Map;
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1,5 +1,6 @@
 // @flow
 
+import { version } from '../../package.json';
 import { extend, bindAll, warnOnce, uniqueId } from '../util/util';
 import browser from '../util/browser';
 import window from '../util/window';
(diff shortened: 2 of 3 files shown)


chloekraw:
cc @tgorkin @chrisirhc @heshan0131 @ibgreen

Pessimistress:
@ansis Thank you!
For backward compatibility, is there a way we can get the version of mapbox-gl from the Map instance?

ansis:
> For backward compatibility, is there a way we can get the version of mapbox-gl from the Map instance?

The version isn't on the map instance but it is exported by the library (`mapboxgl.version`). Does that work?

src/geo/transform.js
```diff
@@ -537,9 +537,18 @@ class Transform {
...
         const farZ = furthestDistance * 1.01;
 
+        // The larger the value of nearZ is
+        // - the more depth precision is available for features (good)
+        // - clipping starts appearing sooner when the camera is close to 3d features (bad)
+        //
+        // Smaller values worked well for mapbox-gl-js but deckgl was encountering precision issues
+        // when rendering it's layers using custom layers. This value was experimentally chosen and
+        // seems to solve z-fighting issues in deckgl while not clipping buildings too close to the camera.
+        const nearZ = this.height / 50;
```
astojilj: LGTM
On native side, [near set to `0.1 * state.getCameraToCenterDistance()`](https://github.com/mapbox/mapbox-gl-native/blob/286bc66ba4eef043af1d67858f8f5ec540717f31/src/mbgl/renderer/paint_parameters.cpp#L26) evaluates to height / 90.

Pessimistress:
@ansis We need to account for both the script tag and node_modules. So we end up having to do something like

```js
function getMapboxVersion() {
  if (typeof mapboxgl === 'undefined') {
    return mapboxgl.version;
  }
  try {
    return require('mapbox-gl').version;
  } catch (err) {
    return 'unknown';
  }
}
```

And also change bundler configs to avoid bundling mapbox-gl with deck.

Pessimistress:
@ansis Just brainstorming, can `layer.render` receive a third argument that are miscellaneous parameters? Then we can do something like

```
render(gl, matrix, {near, far}) {
  ...
}
```

ansis:
I've added the version to the map instance. Exposing the actual parameters is a good idea but in the interest of just getting this merged I've added the version.

src/geo/transform.js
asheemmamoowala: Is there an existing render test that exercises this with extrusion layers?

asheemmamoowala:
> We need to account for both the script tag and node_modules. 

@Pessimistress can you say more about what the problem with your current approach is? It looks like you are trying to support multiple ways of importing mapboxgl in your project (or it is demand loaded?)

When a [CustomLayerInterface](https://docs.mapbox.com/mapbox-gl-js/api/#customlayerinterface) is added to a map, and `onAdd` is invoked, you should be able to capture the `mapboxgl.version` and use it to determine how the layer behaves.

Pessimistress:
@asheemmamoowala Our custom layer class is published to npm and it does not depend on mapbox-gl. This is clearly allowed as long as it implements the custom layer interface.

From the layer's perspective, we cannot tell how the user is including mapbox. If they are using a bundler, then there will not be a global `mapboxgl` object. Conditionally requiring an external module module is dangerous, because

- It requires additional bundler setup for users who wish to include mapbox using the script tag
- It breaks the esm build by mixing require and export
- It generally violates our coding policy by requiring something that is not in the dependencies list

ansis:
> Is there an existing render test that exercises this with extrusion layers?

There are tests with extrusions but none that exercise this because all the values work well for us in mapbox-gl. It's only in deckgl that the issues arise.

> When a CustomLayerInterface is added to a map, and onAdd is invoked, you should be able to capture the mapboxgl.version and use it to determine how the layer behaves.

onAdd provides the map, not `mapboxgl` which is what is motivating adding it to the map

PR #8513, opened 2019-07-18
Move `esm` to devDependencies

### LINKED ISSUE #8450: esm dependency vunerability
Getting some github alerts while using this lib due to  `"esm": "~3.0.84",`

Screen Shot 2019-07-09 at 4 34 33 PM

Do you think you could give them a bump?

DatGreekChick:
## Launch Checklist

 - [ ] post benchmark scores

I simply ran `npm i -D esm` to move the package to `devDependencies`. This is the output I get from running tests:

```
// unit tests: my branch
total ......................................... 16491/16491

  16491 passing (3m)

  ok

// integration tests - my branch
440 passed (40.8%)
1 passed but were ignored (0.1%)
6 ignored (0.6%)
631 failed (58.5%)
1 errored (0.1%)

// integration tests - master
440 passed (40.8%)
1 passed but were ignored (0.1%)
6 ignored (0.6%)
631 failed (58.5%)
1 errored (0.1%)
```

I don't believe running anything else is relevant since I'm only moving the package location as per the issue tagged below. If this is not the case, please let me know, and I'll be happy to close more items from the checklist.

CLOSES #8450


ryanhamley:
I think you just need to change `^` back to `~`. Newer versions of ESM break our test suite for some reason (see https://github.com/mapbox/mapbox-gl-js/issues/8450) so we have to keep it locked to patch changes, hence the tilde semver range.

mourner:
Also, please use Yarn instead of NPM so that Yarn.lock is updated properly as well.

DatGreekChick:
Ok will do and will either reopen or open a new PR. Thanks @ryanhamley and @mourner 
However, when I changed it to a `~` I still got the same amount of integration tests failing locally. I'll try again and update you both.

ryanhamley:
Did you try with `yarn install` then `yarn run test`? It's a known issue that `npm install` will not likely break things.

asheemmamoowala:
@DatGreekChick can you share the failure result that you see locally?

Issue #8246, opened 2019-05-14
Focused control button doesn't defer visually from non-focused


life777:
**mapbox-gl-js version**: 0.54.0

**browser**: Chome 74 for Linux

### Steps to Trigger Behavior

If you have some controls on the map (like zoom controls) and you navigate to them with Tab button you can't distinguish what button is focused now. There are no styles for focused state on the buttons. There should be either outline (like on links in attributions control) or apply hover styles when button is focused.

 1. Open map with zoom control
 2. Navigate with Tab to any zoom button

### Link to Demonstration
Example with mapbox:
https://jsbin.com/qemokezepu/edit?html,output 

### Expected Behavior

Focused buttons should have different styles (or just outline) than non-focused.

### Actual Behavior

User can't distinguish what button he/she is currently focused on.

andrewharvey:
I'm +1 to having a blue highlight around focused buttons, exactly like GitHub does, which seems to be best practice. Are you interested in putting together a PR?

life777:
Yes, I would love to help you with this case during this week. Thanks.

PR #8520, opened 2019-07-19
Display outline on control buttons when focused

ryanhamley:
## Launch Checklist

Fixes https://github.com/mapbox/mapbox-gl-js/issues/8246

    - this creates an outline on the control buttons when they are tab focused, which helps with accessibility issues
    - this avoids using the CSS `outline` property because outlines do not follow rounded borders. `box-shadow` does follow rounded borders, leading to a cleaner looking UI while simulating the outline effect (not sure why the GIF is so slow but you get the idea)

buttons

### DIFF
--- a/src/css/mapbox-gl.css
+++ b/src/css/mapbox-gl.css
@@ -75,7 +75,6 @@
 
 .mapboxgl-ctrl-group {
     border-radius: 4px;
-    overflow: hidden;
     background: #fff;
 }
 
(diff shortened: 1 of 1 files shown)


life777:
@ryanhamley  I agree, your solution looks better. Thanks

ryanhamley:
I fixed the lint issue and added a couple new rules that will only display the outline when the buttons are focused by keyboard input using the `:focus-visible` selector. This is currently only supported out of the box in Firefox. It's also supported in Chrome behind the `enable-experimental-web-platform-features` flag. Normally I wouldn't use an experimental feature, but this is part of the W3C spec and seems very likely to be implemented widely in the not-too-distant future; it's harmless for browsers that don't implement it and makes the UI a touch nicer for browsers that do. That being said, I'm not real attached to it so we can remove those two rules if anyone has an objection.

src/css/mapbox-gl.css
```diff
@@ -111,6 +110,30 @@
     background-color: rgba(0, 0, 0, 0.05);
 }
 
+.mapboxgl-ctrl-group > button:focus {
+    box-shadow: 0 0 2px 2px rgba(0, 150, 255, 1);
+}
+
+.mapboxgl-ctrl-group > button:focus-visible {
+    box-shadow: 0 0 2px 2px rgba(0, 150, 255, 1);
+}
```
mourner: Nit: you can group these two selectors.

ryanhamley: I actually can't because I just realized Firefox uses the `-moz-focusring` selector instead of `focus-visible` so combining them makes the entire rule invalid in Firefox which breaks all focusing on the buttons. It turns out that Firefox and Safari don't appear to register clicks as focusing the button so they just work without this rule. I'll still leave it though because it's a progressive enhancement in Chrome and presumably newer versions of Edge.

PR #8542, opened 2019-07-24
8504 duplicate match exp

### LINKED ISSUE #8504: Style spec feature_filter should dedupe values for legacy `in` statements
## Motivation

A customer reported an issue migrating their styles using Studio in order to use expression syntax. Their legacy filters contained an `in` statement that referenced duplicate values, for example:

`filter: [ 'in', 'parent_0', 'AA', 'AA', 'AB', 'AC', ...  ]`

This seems to be a valid legacy filter, however upon migration it fails validation for the newer syntax with error output `GL JSON Invalid: Branch labels must be unique`.

I propose that the [`convertInOp`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/feature_filter/convert.js#L164) method to convert legacy filter syntax include a line to remove duplicate string or numerical values. This can only benefit users of this utility by easing migration from a looser validation requirement to a tougher one.

## Design Alternatives

Alternatively, we could tighten legacy validations, but that would break lots of working maps. We could also surface a warning in GL when we detect duplicate filter values in legacy syntax, however this could be confusing as the filter would still be valid.

dereklieu:
Closes #8504 

The utility to convert legacy filters to expressions currently outputs a `match` expression as an optimization for certain legacy `in, !in` operators. This PR removes duplicate values from the result to ensure that the expression is valid.

### DIFF
--- a/src/style-spec/feature_filter/convert.js
+++ b/src/style-spec/feature_filter/convert.js
@@ -188,13 +188,8 @@ function convertInOp(property: string, values: Array<any>, negate = false) {
 
     if (uniformTypes && (type === 'string' || type === 'number')) {
         // Match expressions must have unique values.
-        const seen = [];
-        const uniqueValues = values.filter(v => {
-            if (seen.indexOf(v) === -1) {
-                seen.push(v);
-                return true;
-            }
-            return false;
+        const uniqueValues = values.sort().filter((value, i) => {
+            return i === 0 || values[i - 1] !== value;
         });
         return ['match', get, uniqueValues, !negate, negate];
     }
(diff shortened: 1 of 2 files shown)


src/style-spec/feature_filter/convert.js
```diff
@@ -187,7 +187,16 @@ function convertInOp(property: string, values: Array<any>, negate = false) {
...
-        return ['match', get, values, !negate, negate];
+        // Match expressions must have unique values.
+        const seen = [];
+        const uniqueValues = values.filter(v => {
+            if (seen.indexOf(v) === -1) {
+                seen.push(v);
+                return true;
+            }
+            return false;
+        });
```
mourner: Note that this is `O(n^2)`, so might get really slow when the filter has a lot of values (although not sure how likely we are to hit this in practice). An alternative would be to sort the values and then filter linearly, since it doesn't matter what order they are in the `match` expression: 

```js
const uniqueValues = values.sort().filter((v, i) => values[i - 1] !== v);
```

dereklieu: 👍

mourner: Nit: you can write this shorter without curly brackets, `return` and on one line.

PR #8565, opened 2019-07-27
Fix usage of String#includes which is es6 but not transformed by buble

ahk:
Fix a usage of `String#includes` which breaks in any non-es6 browser and is also not handled by buble transformation.

### DIFF
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -178,7 +178,7 @@ export class RequestManager {
         if (accessToken[0] === 's')
             throw new Error(`Use a public access token (pk.*) with Mapbox GL, not a secret access token (sk.*). ${help}`);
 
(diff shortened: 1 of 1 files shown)


src/util/mapbox.js
```diff
@@ -178,7 +178,7 @@ export class RequestManager {
         if (accessToken[0] === 's')
             throw new Error(`Use a public access token (pk.*) with Mapbox GL, not a secret access token (sk.*). ${help}`);
 
-        urlObject.params = urlObject.params.filter((d) => !d.includes('access_token'));
+        urlObject.params = urlObject.params.filter((d) => d.indexOf('access_token') !== -1);
```
mourner: Invert the comparison — "not includes" will translate to `indexOf(...) === -1`.

ahk: Thank you Vlad. Classic.

PR #8507, opened 2019-07-18
[docs] adds feedback component

katydecorah:
This PR adds a feedback component to bottom of pages and sections. This component is currently on all docs.mapbox.com sites and it looks like this:

image

The component will appear at the end of:
- /mapbox-gl-js/overview/
- each subsection in /mapbox-gl-js/api/
- each example at /mapbox-gl-js/example/*/
- each section /mapbox-gl-js/style-spec/

Note: since the Style spec's subsections tend to be shorter than subsections on the API page, we decided to only place the component at the end of each Style spec section.

After the launch of this final site, look out for an internal post from the docs team where we'll share how teams can review ratings for given sites and pages. We'll also share how the docs team is triaging this feedback on a daily basis.

Ref https://github.com/mapbox/documentation/issues/209

## Launch Checklist
 - [x] briefly describe the changes in this PR
 - [x] check with gl-js team before merging since this PR makes changes to yarn.lock

### How to test

1. `yarn run start-docs`
1. click yes/no on one of the feedback components
1. open the Segment mapbox-staging debugger and search for "sent docs feedback" to see if your events came through (events should come through regardless of ad blocker)

PR #8580, opened 2019-07-30
Merge publisher-production into master

mourner:
Get `master` up-to-date with `publisher-production` to prepare for the docs split. I rebased the changes to reduce the noise, and I didn't preserve the merge commits because we'll get rid of `publisher-production` anyway.


docs/constants.json
```diff
@@ -0,0 +1,6 @@
+{
+  "FORWARD_EVENT_WEBHOOK": {
+    "staging": "https://evj5gwoa8j.execute-api.us-east-1.amazonaws.com/hookshot/webhook",
```
asheemmamoowala: What PR did these land from and do they correctly belong in the repo?

mourner: @asheemmamoowala this landed for the feedback component https://github.com/mapbox/mapbox-gl-js/pull/8507

PR #8584, opened 2019-08-02
Running query tests in browser

arindam1993:
This PR adds a bunch of new tooling to allow the `query` tests to be run in the browser.
New tools added:
- `testem`: this is a test-runner that that handles 
   - launching browsers
   - serving test files to launched browsers, and receiving test results back via websocket
   - persisting test results to disk once tests have finished in `xunit` format
- `tape`: browser compatible version of `tap`, used to write the tests.
- `browserify`: tape is not browser compatible ootb, so we need to build it ourselves, and rollup does not handle tape's commonjs style well and fails at trying to inject node globals and builtins. So we build it with browserify.

### 1.Loading fixture data to the browser
This is done via a preprocessor that generates `fixtures.json` file within `test/integration/dist` that contains all the data within the directory inlined within it,  this file can then be imported into the test root file and the `rollup` will inline everything so all the data is available synchronously.

### 2.Bundling resources for the browser
There are 3 bundles that get built in parallel before the test suite starts
-`tape_config.js`: Done via browserify, and exposed as a global `tape`.
- `mapboxgl`: Done via `rollup`, exposed as a global `mapboxgl`.
- `query-browser.js`: this is the entrypoint of the test suite, built via `rollup`as an iife so that tests start executing as soon as the script is loaded.

### 3. Test running
`testem` provides hooks to run custom code before and after tests. The `before_tests` hook is used to run the build, and start an http server thats used to serve up certain resources such as spritesheets and fonts( this is the same as the existing implementation).
`testem` then launches its own server with a static html page which server up the built javascript resources. An extra script injected by testem onto the page captures the `tap` formatted `console.log` messages and relays them over websocket to be interpreted on the server to detect when the test suite has finished. 
It then converts the gathered `tap` formatted report and convertes it to `xunit` format for integration with CircleCI's test reporting.

### 4.Test reporting
testem finishes running tests and persists the test report in xunit format to disk.
I've intentionally left a failing query test so we can see how it shows up in circle.

## Guide:
Adds a bunch of npm scripts:
`yarn run test-query-browser` runs all the tests in headless chrome and exits
`yarn run watch-query` starts build watchers and launches a chrome window to run tests so you can work iterate on tests.
`yarn run start-query-debug` starts both the test server and the debug server while sharing the saem build-watcher.


package.json
```diff
@@ -35,10 +35,12 @@
     "quickselect": "^2.0.0",
     "rw": "^1.3.3",
     "supercluster": "^6.0.1",
+    "tape": "^4.11.0",
```
ryanhamley: Why isn't `tape` a `devDependency`?

.circleci/config.yml
```diff
@@ -99,10 +99,10 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v2-yarn-{{ checksum "yarn.lock" }}
-      - run: yarn
+            - v3-yarn-{{ checksum "yarn.lock" }}
+      - run: sudo npm install -g yarn & yarn install
```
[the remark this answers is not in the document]
arindam1993: There aren't any tools that we use other OS level than yarn and browsers, everything else gets bundled into `node_modules` and cached. Browsers are already provided with this image and this just updates the version of yarn.

build/rollup_plugins.js
```diff
@@ -5,31 +5,44 @@ import resolve from 'rollup-plugin-node-resolve';
...
-    }),
-    commonjs({
-        // global keyword handling causes Webpack compatibility issues, so we disabled it:
-        // https://github.com/mapbox/mapbox-gl-js/pull/6956
-        ignoreGlobal: true
-    })
-].filter(Boolean);
+export const plugins = (minified, production, test) => {
+    let env = 'development';
+    if(production){
```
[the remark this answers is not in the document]
arindam1993: Removed this, its unnecessary now.

build/test/build-tape.js
```diff
@@ -0,0 +1,10 @@
+var browserify = require('browserify');
+var fs = require('fs');
+
+
+var b = browserify(require.resolve('../../test/util/tape_config.js'), { standalone: 'tape' })
```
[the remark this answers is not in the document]
arindam1993: `rollup` struggles to inject node builtins into tape. 
https://github.com/substack/tape/issues/410
https://github.com/substack/tape/issues/415
And `tape` devs prescribe handling it via browserify on user end.

Modifying `tape` is also proving to be a challenge, mainly because it relies on `through`, the github repo for which is read-only.

Also the use for `browserify` in this case is minimal, since we're using it to only package `tape`, our code still goes through rollup as usual.

test/integration/lib/generate-fixture-json.js
```diff
@@ -0,0 +1,97 @@
...
+ * the contents into a single json file which can then we imported and built into a bundle
+ * to be shipped to the browser.
+ *
+ * @param {string} directory
+ */
+module.exports = function (directory) {
+    const basePath = directory;
+    const jsonPaths = path.join(basePath, '/**/*.json');
+    const imagePaths = path.join(basePath, '/**/*.png');
+    const ignoreOutputPath = path.join(basePath, OUTPUT_FILE);
```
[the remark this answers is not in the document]
arindam1993: Moved to `test/integration/dist`

src/util/config.js
```diff
@@ -22,7 +22,13 @@ const config: Config = {
         }
     },
     FEEDBACK_URL: 'https://apps.mapbox.com/feedback',
-    REQUIRE_ACCESS_TOKEN: true,
+    REQUIRE_ACCESS_TOKEN: ((() => {
```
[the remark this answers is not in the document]
arindam1993: removed

kkaefer:
> I was discussing this with @asheemmamoowala, latest browser versions could unexpectedly cause some tests to break, causing development hassle, but then this is behaviour our users would see too, so is it worth catching in tests?

Our main reason for not using container names that change (like `*-latest`, or `*-dev`) is that it makes it hard to build historical branches on CI, e.g. when releasing a bugfix for an old branch. However, this is more of a problem on gl-native than it is on gl-js. In any case, Circle CI supports pinning to a specific image by appending `@sha256:<hash>` to the image name.

build/rollup_plugins.js
```diff
@@ -12,24 +12,26 @@ import { createFilter } from 'rollup-pluginutils';
 // Common set of plugins/transformations shared across different rollup
 // builds (main mapboxgl bundle, style-spec package, benchmarks bundle)
 
-export const plugins = (minified, production) => [
```
asheemmamoowala: nit: Seems like no change in this file (other than whitespace)

flow-typed/environment.js
```diff
@@ -0,0 +1,2 @@
+// @flow strict
+declare var GL_JS_ENV: 'production' | 'development' |'test';
```
asheemmamoowala: nit: Not seeing this used anywhere

testem.js
```diff
@@ -0,0 +1,66 @@
...
+let server;
+
+module.exports =  {
+    "framework": "tap",
+    "src_files": [
+        "test/integration/dist/*.js"
+    ],
+    "serve_files": [
+        "test/integration/dist/tape.js",
+        "test/integration/dist/mapbox-gl-test.js",
```
asheemmamoowala: Remove this rollup target destination and just use `dist/mapbox-gl-dev.js` instead. It would allow a simpler dev workflow where the debug pages and the test pages can be open simulatneously.

arindam1993: Done, you can now have both open simultaneously

arindam1993:
> Would be good to add flow typing to the new test lib scripts.

Unfortunately testem does not play nice with `esm`, and flow for commonjs modules is not great.


> I found a couple issues with running the testem server locally and trying to build new query tests or modify code to fix test output. Looks like you are missing watchers for the test fixture and gl-js targets.
> 
> Addtionally, if a browser stays open to the testem page, it holds on to the page (and reruns it) even if the testem server is re-launched in the terminal. This gives a false sense of whether the old/new code is being run.

I added the watchers you can now do.
`yarn run watch-query`: runs only the query server in watch mode
`yarn run watch-query-debug`: runs both the query and the debug server working off the same gl-js build.

.circleci/config.yml
kkaefer: Why `sudo`? Docker builds typically run as the root user anyway.

arindam1993:
> This looks nearly ready. A few items to wrap up before merging:
> 
> * Update `test-query` and related `package.json` scripts to use this version of the test instead of the node based tests
> * Update integration tests [readme](https://github.com/mapbox/mapbox-gl-js/blob/master/test/integration/README.md) with instructions to run the node-based tests, and to run/debug the testem based tests.

Done 👍 , I also made it so that it starts the testem server when running `start`/`start-debug` both the debug pages on `:9966/debug` and the test page on `:7357` run off the same bundle and build watcher.

PR #8585, opened 2019-08-02
Expose mapboxgl.rtlTextPluginRequested boolean

### LINKED ISSUE #7869: Expose a "clearRTLTextPlugin" or a give a way to check if the rtl plugin is requested
## Motivation

Right now there is no way to check if the RTL plugin has been requested or not, as such if we try to set it while it was already requested the following error occurs: ` setRTLTextPlugin cannot be called multiple times.`. which is an expected behavior.

however, if this error is not caught it may block the controls from rendering. 
In my case, i encountered this error in a vuejs component where I always initialize the RTL plugin when the component is loaded even tho I make sure to destroy the map using `map.remove()` after each time leaving the component the error still persists whenever I navigate back to that route (which loads that component again)

## Design Alternatives

expose [this function](https://github.com/mapbox/mapbox-gl-js/blob/ead961f7c6b903bb5bbfbc3f128362bb46d421eb/src/source/rtl_text_plugin.js#L28) in the `index.js` or create a getter to `pluginRequested` variable to check if the RTL plugin has been requested

ryanhamley:
## Launch Checklist

Fixes #7869 

  This PR exposes `mapboxgl.rtlTextPluginRequested` which returns a boolean that is `true` if the RTL plugin has previously been requested. This allows users to avoid requesting the plugin multiple times which causes an error.
  Added an RTL debug page

I considered exposing the existing [`clearRTLTextPlugin`](https://github.com/mapbox/mapbox-gl-js/blob/ead961f7c6b903bb5bbfbc3f128362bb46d421eb/src/source/rtl_text_plugin.js#L28) method, but ultimately decided that this made more sense code-wise. I'm open to differing opinions though.

```
if (!mapboxgl.rtlTextPluginRequested) {
  mapboxgl.setRTLTextPlugin(url);
}
```

### DIFF
--- a/docs/documentation.yml
+++ b/docs/documentation.yml
@@ -7,6 +7,7 @@ toc:
   - supported
   - version
   - setRTLTextPlugin
+  - rtlTextPluginRequested
   - clearStorage
   - AnimationOptions
   - CameraOptions
--- a/debug/rtl.html
+++ b/debug/rtl.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Mapbox GL JS debug page</title>
+    <meta charset='utf-8'>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+    <link rel='stylesheet' href='/dist/mapbox-gl.css' />
+    <style>
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+
+<body>
+<div id='map'></div>
+
+<script src='/dist/mapbox-gl-dev.js'></script>
+<script src='/debug/access_token_generated.js'></script>
+<script>
+
+mapboxgl.setRTLTextPlugin('https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.0/mapbox-gl-rtl-text.js');
+
+var map = window.map = new mapboxgl.Map({
+    container: 'map',
+    zoom: 16.5,
+    center: [44.355435, 33.258814],
+    style: 'mapbox://styles/mapbox/streets-v11',
+    hash: true
+});
+
+map.on('load', function () {
+    if (!mapboxgl.rtlTextPluginRequested) {
+        mapboxgl.setRTLTextPlugin('https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.0/mapbox-gl-rtl-text.js');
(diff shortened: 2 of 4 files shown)


docs/documentation.yml
```diff
@@ -7,6 +7,7 @@ toc:
   - supported
   - version
   - setRTLTextPlugin
+  - rtlTextPluginRequested
```
mourner: Note that this file is being moved over to https://github.com/mapbox/mapbox-gl-js-docs — should we keep it in this repo? We could either keep it, or move it over but PR ToC changes to the docs repo after a release separately. cc @colleenmcginnis

ryanhamley: I'm willing to defer to you and Colleen on whether this file should be moved or not since you have both thought about this way more than I have. This isn't super urgent so I can wait until we get the docs separated in order to merge this PR. That gives us time to figure out what we want to do with the ToC.

asheemmamoowala: > should we keep it in this repo? We could either keep it, or move it over but PR ToC changes to the docs repo after a release separately.

It would be better to move this over to docs, and add a step to the post-release PR for docs repo to ensure that these types of changes get added to the ToC.

ryanhamley:
I'm not sure if an event listener would really solve the use cases that were reported for this feature. A typical use case is something like a React component which always tries to initialize the plugin whenever a user navigates to a route that uses that plugin. An event listener would theoretically be cleaned up and removed each time the component is removed. Even if you persisted the event listener, once the plugin is instantiated, the events wouldn't be firing anymore so the component wouldn't always be receiving the event to alert it if the plugin is available or not. We need a more active way of checking the component's availability.

An option that I also considered that we could do instead of or even in tandem with this PR is to expose the `clearRTLTextPlugin` method which could be called when cleaning up a component or route which would make re-initializing the plugin safe the next time the component is instantiated.

asheemmamoowala:
> 'm not sure if an event listener would really solve the use cases that were reported for this feature. 

@ryanhamley the event handler was a suggestion  _in addition to_ a status check API:
>  to expose the status of the plugin - being one of : unavailable, pending/loading, available/loaded, error

Would that not satisfy the reported issue? 
Your sample code would then change to:
```
if (mapboxgl.rtlTextPluginStatus === 'unavailable') {
  mapboxgl.setRTLTextPlugin(url);
}

```

ryanhamley:
This is fairly out of date so I'm closing in favor of a slightly different implementation https://github.com/mapbox/mapbox-gl-js/pull/8864

PR #8550, opened 2019-07-24
Document type constrain on feature.id for feature state methods.

asheemmamoowala:
Closes #7632 

Even though the `feature.id` constraint applies to the data, it is encountered most often in the context of trying to use the `feature state` APIs.

PR #8592, opened 2019-08-02
Document type constrain on feature.id for feature state methods

asheemmamoowala:
Rework of #8550 targeting `master` instead of `publisher-production`


peterqliu:
🚢

PR #8600, opened 2019-08-06
fix polygons with incorrect winding orders

arindam1993:
Refer Issue:
https://github.com/openmaptiles/openmaptiles/issues/639

Looks a bunch of data-sources provide extrusion polygons with inconsistent winding orders, pre 0.50 we had backface culling turned off, causing these extrusions to render somewhat-kinda properly.

We can simulate the previous behavior on the CPU side by implementing a winding order detection and reversal, which is what this PR does.

After:
Screen Shot 2019-08-05 at 7 33 03 PM

 


src/util/util.js
```diff
@@ -385,6 +385,28 @@ export function isClosedPolygon(points: Array<Point>): boolean {
     return Math.abs(calculateSignedArea(points)) > 0.01;
 }
 
+/**
+ * Retruns the maximum value of an element in an array,
```
ryanhamley: Typo `Retruns` and on the second line `retured`. You should also mark this function as `@private` in the JSDoc header

src/data/bucket/fill_extrusion_bucket.js
```diff
@@ -281,3 +278,15 @@ function isEntirelyOutside(ring) {
         ring.every(p => p.y < 0) ||
         ring.every(p => p.y > EXTENT);
 }
+
+function cleanupPolygonWindingOrder(polygon) {
+    // reverse all rings because have a case wherein the polygons are provided
```
ryanhamley: comment is a bit unclear `reverse all rings because have`

src/util/util.js
```diff
@@ -385,6 +385,28 @@ export function isClosedPolygon(points: Array<Point>): boolean {
...
+ * based on the value retured by the comparator function. Returns null if the
+ * array is empty
+ *
+ * @export
+ * @param {Array<any>} array
+ * @param {(elem: any) => number} comparator
+ * @returns {*}
+ */
+export function max(array: Array<any>, comparator: (elem: any) => number): ?any {
```
ryanhamley: Where is this function being called?

src/util/util.js
```diff
@@ -385,6 +385,28 @@ export function isClosedPolygon(points: Array<Point>): boolean {
     return Math.abs(calculateSignedArea(points)) > 0.01;
 }
 
+/**
+ * Retruns the maximum value of an element in an array,
+ * based on the value retured by the comparator function. Returns null if the
+ * array is empty
+ *
+ * @export
```
ryanhamley: `@export` isn't a valid JSDoc tag and is breaking the linter FYI

src/data/bucket/fill_extrusion_bucket.js
```diff
@@ -219,11 +221,6 @@ class FillExtrusionBucket implements Bucket {
                 segment = this.segments.prepareSegment(numVertices, this.layoutVertexArray, this.indexArray);
             }
 
-            //Only triangulate and draw the area of the feature if it is a polygon
-            //Other feature types (e.g. LineString) do not have area, so triangulation is pointless / undefined
-            if (vectorTileFeatureTypes[feature.type] !== 'Polygon')
-                continue;
```
ahk: The position of this check prevents constructing geometry for caps (roofs, ceilings), and also prevents attempting to triangulate this cap geometry. Do we remove this check because the two following for loops and earcut() effectively `noop` in this case?

src/util/util.js
ahk: Agree that it doesn't seem used.

src/data/bucket/fill_extrusion_bucket.js
ahk: I ran the render tests with this check re-included and it produces the old test-case output that has been updated elsewhere in this PR. So this check does change the output. No idea if we like that output but we did make a test explicitly for this case of `MultiLineString` feature type in a fill-extrusion bucket.

src/util/util.js
arindam1993: Ah, yes. I was using a different approach to determine the `outer shell` polygon earlier. Which was picking the polygon with the largest `area`. I later realized that that was necessary since, that always seems to be at `polygons[0]`. I left the util function in there just in case we need it later.

src/data/bucket/fill_extrusion_bucket.js
arindam1993: I don't have the context for this either, but `openmaptiles` does have some of their buildings as `LineString` and this check was preventing only the caps from rendering.

I'm thinking we can either not generate the entire `fill-extrusion` when the feature type isn't `Polygon` or, we can do what this PR does and tries to do and ignores the check, or keep the status quo.

ahk: I see, so we have closed `Polygon` data mislabeled as `LineString`. I'm assuming this happens in the GeoJSON or tiles themselves. I think if you saw more output that looks like correct buildings (newly generated with caps) by removing this check, then maybe removing the check is the right thing to do. I think I agree that we should either render the cap or render no extrusion at all. We should get some more feedback from people who know the user input data better.

src/util/util.js
ahk: I think we should avoid adding code we aren't either:
- using internally
or
- exposing as an API for map developers

src/data/bucket/fill_extrusion_bucket.js
ryanhamley: This check was added in https://github.com/mapbox/mapbox-gl-js/pull/7685 to avoid visual artifacts. @mourner or @ansis should have more context

arindam1993: Thanks for the issue link @ryanhamley, I reverted the change. It seems like it seems like its a good idea to leave it in.

ahk: I still want to know if these caps we're avoiding would be wrong artifacts or potentially representing data meant as a polygon that wasn't correctly labeled, but agree that reverting is probably best.

arindam1993:
> This looks fine, but I'd like to point out that [the vector tile spec requires a certain winding order](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#4344-polygon-geometry-type) of encoded polygons, and GeoJSON winding order is fixed [at another place](https://github.com/mapbox/mapbox-gl-js/blob/9e6c6d95ac9491fba0abfef0e1e9ea1356755192/src/source/geojson_worker_source.js#L172) in code, so this PR is fixing a symptom of a bug specifically in OpenMapTiles. I would personally prefer for it to be fixed on their side.

I agree, do we have an agreed on mechanism in our SDK to warn users of errors in their data if we silently fix them? We can trigger a warning.

pathmapper:
Is there a noticeable performance impact when fixing many polygons with a lot of vertices?

arindam1993:
> Is there a noticeable performance impact when fixing many polygons with a lot of vertices?
> 
> (e.g. extrusions of buildings on a tilted map or use cases like shown in #7247)

I think it should be ok, its not adding to the big-O complexity of the routine, and extrusion polygons generally have few vertices. It also does the fix in-place and off the main thread.

Issue #8596, opened 2019-08-05
Make hash configurable


SebCorbin:
## Motivation

Hash option is a boolean, if we set it, we can't use `location.hash` for anything else other than the mapbox hash. It would be nice to implement a query parameter style, if defined, like `myurl.org/mymap#map=10.37/14.589/-61.0225&config=myotheroptions`.
This is useful for permalinks for application containing a configurable map of anything else actually.

## Design Alternatives

Add an optional (to prevent breaking change) `hashName` option, that if defined, make the hash a query parameter-styled string.

### Implementation

Add and document this option, modify `ui/hash.js` so that it replaces (if defined), only the `hashName` part of the map.

asheemmamoowala:
@SebCorbin Thanks for the proposal. Would you be interested in implementing it as well? The changes should be limited to https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/hash.js and fairly isolated to test.

SebCorbin:
Sure, I have a PR in progress, I think I'll drop the `hashName` in favor of a `hash` option that can be boolean or string. I see that you support IE11, so I may not use [URLSearchParams](https://caniuse.com/#search=URLSearchParams), right?

andrewharvey:
Related to #7725. I like this approach to make it more flexible and let the developer choose the name they want to use.

JesseWeinstein:
I'm very glad to see this -- please consider also allowing the named form as a *input only* alternative, ideally on by default. That way all maps using mapbox-gl will automatically be able to accept pasted-in URLs from OSM without barfing.

andrewharvey:
was closed by #8603

PR #8603, opened 2019-08-06
Configurable hash #8596

SebCorbin:
Refs https://github.com/mapbox/mapbox-gl-js/issues/8596

## Launch Checklist

 - [x] post [benchmark scores](https://user-images.githubusercontent.com/645207/62539432-7cf9fe00-b855-11e9-8bbe-5072584a3a36.png)

### DIFF
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -17,8 +17,8 @@ class Hash {
     _updateHash: () => ?TimeoutID;
     _hashName: ?string;
 
-    constructor(hashName: string | boolean) {
-        this._hashName = (typeof hashName === 'string' && hashName) || null;
+    constructor(hashName: ?string) {
+        this._hashName = hashName;
         bindAll([
             '_getCurrentHash',
             '_onHashChange',
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -18,7 +18,7 @@ class Hash {
     _hashName: ?string;
 
     constructor(hashName: ?string) {
-        this._hashName = hashName;
+        this._hashName = hashName && encodeURIComponent(hashName);
         bindAll([
             '_getCurrentHash',
             '_onHashChange',
@@ -78,18 +78,19 @@ class Hash {
         if (bearing || pitch) hash += (`/${Math.round(bearing * 10) / 10}`);
         if (pitch) hash += (`/${Math.round(pitch)}`);
 
-        if (this._hashName && !mapFeedback) {
+        if (this._hashName) {
+            const hashName = this._hashName;
             let found = false;
             const parts = window.location.hash.slice(1).split('&').map(part => {
                 const key = part.split('=')[0];
-                if (key === this._hashName) {
+                if (key === hashName) {
                     found = true;
                     return `${key}=${hash}`;
                 }
                 return part;
             }).filter(a => a);
             if (!found) {
-                parts.push(`${this._hashName || ''}=${hash}`);
+                parts.push(`${hashName}=${hash}`);
             }
             return `#${parts.join('&')}`;
         }
@@ -98,6 +99,7 @@ class Hash {
     }
 
     _getCurrentHash() {
+        // Get the current hash from location, stripped from its number sign
         const hash = window.location.hash.replace('#', '');
         if (this._hashName) {
             // Split the parameter-styled hash into parts and find the value we need
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -15,9 +15,12 @@ import type Map from './map';
 class Hash {
     _map: Map;
     _updateHash: () => ?TimeoutID;
+    _hashName: ?string;
 
-    constructor() {
+    constructor(hashName: string | boolean) {
+        this._hashName = (typeof hashName === 'string' && hashName) || null;
         bindAll([
+            '_getCurrentHash',
             '_onHashChange',
             '_updateHash'
         ], this);
@@ -67,18 +70,46 @@ class Hash {
         if (mapFeedback) {
             // new map feedback site has some constraints that don't allow
             // us to use the same hash format as we do for the Map hash option.
-            hash += `#/${lng}/${lat}/${zoom}`;
+            hash += `/${lng}/${lat}/${zoom}`;
         } else {
-            hash += `#${zoom}/${lat}/${lng}`;
+            hash += `${zoom}/${lat}/${lng}`;
         }
 
         if (bearing || pitch) hash += (`/${Math.round(bearing * 10) / 10}`);
         if (pitch) hash += (`/${Math.round(pitch)}`);
-        return hash;
+
+        if (this._hashName && !mapFeedback) {
+            let found = false;
+            const parts = window.location.hash.slice(1).split('&').map(part => {
+                const key = part.split('=')[0];
+                if (key === this._hashName) {
+                    found = true;
+                    return `${key}=${hash}`;
+                }
+                return part;
+            }).filter(a => a);
+            if (!found) {
+                parts.push(`${this._hashName || ''}=${hash}`);
+            }
+            return `#${parts.join('&')}`;
+        }
+
+        return `#${hash}`;
+    }
+
+    _getCurrentHash() {
+        const hash = window.location.hash.replace('#', '');
+        if (this._hashName) {
+            const keyval = hash.split('&').map(
+                part => part.split('=')
+            ).find(part => part[0] === this._hashName);
+            return (keyval ? keyval[1] || '' : '').split('/');
+        }
+        return hash.split('/');
     }
 
     _onHashChange() {
-        const loc = window.location.hash.replace('#', '').split('/');
+        const loc = this._getCurrentHash();
         if (loc.length >= 3) {
             this._map.jumpTo({
                 center: [+loc[2], +loc[1]],
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -100,6 +100,7 @@ class Hash {
     _getCurrentHash() {
         const hash = window.location.hash.replace('#', '');
         if (this._hashName) {
+            // Split the parameter-styled hash into parts and find the value we need
             const keyval = hash.split('&').map(
                 part => part.split('=')
             ).find(part => part[0] === this._hashName);
@@ -110,7 +111,7 @@ class Hash {
 
     _onHashChange() {
         const loc = this._getCurrentHash();
-        if (loc.length >= 3 && !loc.some(v => isNaN(+v))) {
+        if (loc.length >= 3 && !loc.some(v => isNaN(v))) {
             this._map.jumpTo({
                 center: [+loc[2], +loc[1]],
                 zoom: +loc[0],
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -110,7 +110,7 @@ class Hash {
 
     _onHashChange() {
         const loc = this._getCurrentHash();
-        if (loc.length >= 3) {
+        if (loc.length >= 3 && !loc.some(v => isNaN(+v))) {
             this._map.jumpTo({
                 center: [+loc[2], +loc[1]],
                 zoom: +loc[0],
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -173,8 +173,7 @@ const defaultOptions = {
  *
  * @param {boolean} [options.hash=false] If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
  *   For example, `http://path/to/my/page.html#2.59/39.26/53.07/-24.1/60`.
- *   If a string is provided, it will used as the name for a parameter-styled hash, for example:
- *   `http://path/to/my/page.html#map=2.59/39.26/53.07/-24.1/60&foo=bar`
+ *   If a string is provided, it will be used as the name for a parameter-styled hash, e.g. `http://path/to/my/page.html#map=2.59/39.26/53.07/-24.1/60&foo=bar`.
  * @param {boolean} [options.interactive=true] If `false`, no mouse, touch, or keyboard listeners will be attached to the map, so it will not respond to interaction.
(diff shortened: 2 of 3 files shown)


src/ui/map.js
```diff
@@ -173,6 +173,7 @@ const defaultOptions = {
  *
  * @param {boolean} [options.hash=false] If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
```
andrewharvey: the JSDoc can be updated here to show it also accepts a string type. See https://devhints.io/jsdoc

src/ui/map.js
```diff
@@ -171,8 +171,9 @@ const defaultOptions = {
  * Tilesets hosted with Mapbox can be style-optimized if you append `?optimize=true` to the end of your style URL, like `mapbox://styles/mapbox/streets-v9?optimize=true`.
  * Learn more about style-optimized vector tiles in our [API documentation](https://www.mapbox.com/api-documentation/maps/#retrieve-tiles).
  *
- * @param {boolean} [options.hash=false] If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
+ * @param {(boolean\|string)} [options.hash=false] If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
```
asheemmamoowala: nit: type `\|`

src/ui/hash.js
```diff
@@ -15,9 +15,12 @@ import type Map from './map';
 class Hash {
     _map: Map;
     _updateHash: () => ?TimeoutID;
+    _hashName: ?string;
 
-    constructor() {
+    constructor(hashName: string | boolean) {
```
asheemmamoowala: What does `hashName` of boolean type do here? Shouldn't the `hashName` be an optional string parameter

test/unit/ui/hash.test.js
```diff
@@ -4,8 +4,8 @@ import window from '../../../src/util/window';
 import { createMap as globalCreateMap } from '../../util';
 
 test('hash', (t) => {
-    function createHash() {
-        const hash = new Hash();
+    function createHash(name) {
```
asheemmamoowala: Could you add an additional test for `Hash`, so that there is a test with and without a `hashName` parameter.

SebCorbin: I actually _added_ a named version for most of the tests, so it should be enough IMHO. See lines `createHash` usages on lines 22, 34, 47, 77, 125, 173 for unnamed and 105, 143, 219 for named.

That or I did not understand your comment.

src/ui/hash.js
SebCorbin: Exact, I updated `map.js` in consequence

src/ui/map.js
SebCorbin: Removed, was fooled by a badly written cheatsheet https://devhints.io/jsdoc instead of looking at https://jsdoc.app/tags-type.html

chloekraw:
@SebCorbin thanks for this contribution! We're doing the branch cut for the next release tomorrow, so if you have any bandwidth to make the changes today, we can take another look and try to get it into v1.3

SebCorbin:
> @SebCorbin thanks for this contribution! We're doing the branch cut for the next release tomorrow, so if you have any bandwidth to make the changes today, we can take another look and try to get it into v1.3

Sure, all changes have been made, except for the last of @asheemmamoowala's comments which I either did not understand or is not required.

test/unit/ui/hash.test.js
asheemmamoowala: My mistake @SebCorbin , I misread this function in my previous read.

ryanhamley:
I'm seeing a couple issues while testing this out.

Step 1: If I have a page with `hash: true`, I start out in the expected hash configuration, e.g. `http://localhost:9966/debug/#12.5/38.888/-77.01866`

Step 2: But if I change the map to `hash: 'foo'` and reload the page, I get `http://localhost:9966/debug/#12.5/38.888/-77.01866&foo=12.5/38.888/-77.01866` with the new hash just appended to the old one.

Step 3: Now if I go back to `hash: true` and reload, I get a bunch of errors. `Invalid LngLat object: (NaN, 38.888)` followed by `Cannot read property '0' of undefined`.

Likewise, if after step 2, I delete the entire hash and reload the page, I get the correct hash `http://localhost:9966/debug#foo=12.5/38.888/-77.01866`. However, if I then set the map back to `hash: true` and reload, I get `http://localhost:9966/debug#NaN/NaN/NaN` and the same errors as seen in step 3.

SebCorbin:
I would say this is fairly normal, with the current version if you put a wrongly formed hash, you would get a “Failed to invert matrix” error.

As for the reload with another option value, we could at least drop the hash if the parsing fails, but we can’t just assume what would be a previous option, just ask and I’ll update the pull request to add the dropping mechanism.

SebCorbin:
@ryanhamley @chloekraw PR updated 👋

src/ui/hash.js
```diff
@@ -15,9 +15,12 @@ import type Map from './map';
 class Hash {
     _map: Map;
     _updateHash: () => ?TimeoutID;
+    _hashName: ?string;
 
-    constructor() {
+    constructor(hashName: ?string) {
+        this._hashName = hashName;
```
ryanhamley: We should probably do some validation here. If I set `hash: 'hello world'` or `hash: 'hello\nworld'`, the hash is appended twice so it makes sense to ensure the hash is what we expect.

src/ui/hash.js
```diff
@@ -67,19 +70,47 @@ class Hash {
...
         } else {
-            hash += `#${zoom}/${lat}/${lng}`;
+            hash += `${zoom}/${lat}/${lng}`;
         }
 
         if (bearing || pitch) hash += (`/${Math.round(bearing * 10) / 10}`);
         if (pitch) hash += (`/${Math.round(pitch)}`);
-        return hash;
+
+        if (this._hashName && !mapFeedback) {
```
ryanhamley: Nit: can we just move this block into the `else` statement on line 74 to avoid checking `mapFeedback` a second time?

src/ui/hash.js
```diff
@@ -67,19 +70,47 @@ class Hash {
...
+            return (keyval ? keyval[1] || '' : '').split('/');
+        }
+        return hash.split('/');
     }
 
     _onHashChange() {
-        const loc = window.location.hash.replace('#', '').split('/');
-        if (loc.length >= 3) {
+        const loc = this._getCurrentHash();
+        if (loc.length >= 3 && !loc.some(v => isNaN(+v))) {
```
ryanhamley: `isNaN` will coerce the value automatically from a string to a number, so I don't think there's any need to do `+v`. It's also notable that an empty string of any length (e.g. `'    '`) is coerced to 0, regardless of whether it's in `isNaN` or with the `+` operator so perhaps we should check explicitly for that condition? See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Confusing_special-case_behavior

src/ui/hash.js
```diff
@@ -67,19 +70,47 @@ class Hash {
...
+        return `#${hash}`;
+    }
+
+    _getCurrentHash() {
+        const hash = window.location.hash.replace('#', '');
+        if (this._hashName) {
+            const keyval = hash.split('&').map(
+                part => part.split('=')
+            ).find(part => part[0] === this._hashName);
+            return (keyval ? keyval[1] || '' : '').split('/');
```
ryanhamley: is the `|| ''` part necessary?

src/ui/hash.js
```diff
@@ -67,19 +70,47 @@ class Hash {
...
+            const parts = window.location.hash.slice(1).split('&').map(part => {
+                const key = part.split('=')[0];
+                if (key === this._hashName) {
+                    found = true;
+                    return `${key}=${hash}`;
+                }
+                return part;
+            }).filter(a => a);
+            if (!found) {
+                parts.push(`${this._hashName || ''}=${hash}`);
```
ryanhamley: Do we need the `|| ''` part here? If there's no hash name, then shouldn't we skip the `=` as well?

src/ui/hash.js
```diff
@@ -67,19 +70,47 @@ class Hash {
...
+            }
+            return `#${parts.join('&')}`;
+        }
+
+        return `#${hash}`;
+    }
+
+    _getCurrentHash() {
+        const hash = window.location.hash.replace('#', '');
```
ryanhamley: Could you add a comment on this just to make it a little easier to grasp at a quick glance?

src/ui/map.js
```diff
@@ -171,8 +171,9 @@ const defaultOptions = {
  * Tilesets hosted with Mapbox can be style-optimized if you append `?optimize=true` to the end of your style URL, like `mapbox://styles/mapbox/streets-v9?optimize=true`.
  * Learn more about style-optimized vector tiles in our [API documentation](https://www.mapbox.com/api-documentation/maps/#retrieve-tiles).
  *
- * @param {boolean} [options.hash=false] If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
+ * @param {(boolean|string)} [options.hash=false] If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
  *   For example, `http://path/to/my/page.html#2.59/39.26/53.07/-24.1/60`.
+ *   If a string is provided, it will be used as the name for a parameter-styled hash, e.g. `http://path/to/my/page.html#map=2.59/39.26/53.07/-24.1/60&foo=bar`.
```
ryanhamley: Nit, but I think it's probably best to leave the `&foo=bar` off the example just to be clear that something like that isn't added by using a string in this property. You could expand the comment though to include an explanation like "This allows the developer to use arbitrary hashes in addition to the map hash`.

cc @chloekraw on thoughts/wording for this?

test/unit/ui/hash.test.js
```diff
@@ -67,6 +67,14 @@ test('hash', (t) => {
         t.equal(map.getBearing(), 30);
         t.equal(map.getPitch(), 60);
 
+        window.location.hash = '#4/wronlgy/formed/hash';
```
ryanhamley: nit: typo `wronlgy`

src/ui/map.js
chloekraw: Agree with @ryanhamley that as-is, there is some potential for confusion. I would suggest explicitly flagging this as optional and adding a clause at the end of what you've written to explain what `foo` and `bar` refer to. My proposal:

> An additional string may optionally be provided to indicate a parameter-styled hash, e.g. `http://path/to/my/page.html#map=2.59/39.26/53.07/-24.1/60&foo=bar`, where `foo` is a custom parameter and `bar` is an arbitrary hash distinct from the map hash.

Also, could we fix up the linebreaking throughout the PR? We generally keep comments to the 80 column mark.

src/ui/hash.js
SebCorbin: What do you recommend, display a warning/error or sanitize the hashName?

src/ui/hash.js
SebCorbin: This is hardly possible, as we need `bearing` and `pitch` to be added to hash if needed. By the way, `mapFeedback` is not well documented so I don't understand if I need to take it into account when `hashName` is set.

src/ui/hash.js
SebCorbin: Yes, we must ensure there is a string value to be splitted, if you remove it, this test fails with error:
```
        window.location.hash = '#map&foo=bar';

        t.false(hash._onHashChange());
```

SebCorbin: If you really want this out, we can do this:
```diff
diff --git a/src/ui/hash.js b/src/ui/hash.js
index 056d07fbf..de7d30967 100644
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -103,14 +103,14 @@ class Hash {
             const keyval = hash.split('&').map(
                 part => part.split('=')
             ).find(part => part[0] === this._hashName);
-            return (keyval ? keyval[1] || '' : '').split('/');
+            return keyval && keyval.length === 2 && keyval[1].split('/');
         }
         return hash.split('/');
     }
 
     _onHashChange() {
         const loc = this._getCurrentHash();
-        if (loc.length >= 3 && !loc.some(v => isNaN(+v))) {
+        if (loc && loc.length >= 3 && !loc.some(v => isNaN(v))) {
             this._map.jumpTo({
                 center: [+loc[2], +loc[1]],
                 zoom: +loc[0],
```

Your call

src/ui/hash.js
SebCorbin: I don't think an empty string of any length will cause any problem as it is coerced to 0. It seems there haven't been any issue with this until now.

src/ui/hash.js
SebCorbin: `test-flow` fails without this, see https://circleci.com/gh/mapbox/mapbox-gl-js/50268
```
Cannot coerce `this._hashName` to string because null or undefined [1] should not be coerced.

   src/ui/hash.js:92:31
   92|                 parts.push(`${this._hashName}=${hash}`);
                                     ^^^^^^^^^^^^^^
```

SebCorbin:
@ryanhamley weekly reminder for this PR :)

ryanhamley:
I promise I will review this in the next day or two @SebCorbin!

src/ui/hash.js
ryanhamley: We can get around this by assigning `this._hashName` to a variable at the top of the if statement and then changing the two references to `this._hashName` to the variable. That lets us get rid of the unnecessary and confusing `|| ''` part of this while keeping Flow happy.

src/ui/hash.js
ryanhamley: OK, I think leave it as is

PR #8609, opened 2019-08-08
Introduce two component text radial offset

### LINKED ISSUE #8598: text-radial-offset does not account for icon size or bounding boxes
Stub ticket to capture customer feedback:

`text-radial-offset` does not take into account symbol icon dimensions. Text with `text-radial-offset` may overlap wide or horizontally stretched icons with a `"*-left"`/`"*-right"` `text-variable-anchor` position.

text_right_shift_broken

cc @pozdnyakov @alexshalamov

pozdnyakov:
This PR introduces two-component text radial offset for variable text placement. This feature is useful when the text is placed around a symbol icon with different dimensions. The new behavior is described below:

* if `text-radial-offset` is initialized with a number (to keep the existing styles working) or with an array of one element, we'll use the circle offset (same way we do now).

* if `text-radial-offset` is initialized with an array of two elements, these two numbers will stand for the radius on the x and y axes of the ellipse offset.

Note: a new `legacy-type` field is added to the specification JSON file in order to allow usage of legacy types - in this particular case we still allow using plain `number` type with the `text-radial-offset`, so that the existing styles remain working.

Tagging `@mapbox/studio`  `@mapbox/maps-design`  `@mapbox/gl-native` 

Fixes #8598


src/style-spec/validate/validate.js
```diff
@@ -62,7 +62,18 @@ export default function validate(options) {
         return validateExpression(options);
 
     } else if (valueSpec.type && VALIDATORS[valueSpec.type]) {
-        return VALIDATORS[valueSpec.type](options);
+        const errors = VALIDATORS[valueSpec.type](options);
```
alexshalamov: Can this be simplified, if I understand we need to concatenate errors with legacy errors if there is legacy type, right?

asheemmamoowala:
cc @mapbox/studio @mapbox/maps-design @mapbox/gl-native

src/symbol/symbol_layout.js
```diff
@@ -64,48 +64,69 @@ export type TextAnchor = 'center' | 'left' | 'right' | 'top' | 'bottom' | 'top-l
...
-        x = -radialOffset;
+        x = -radialOffsetX;
         break;
     }
 
     return [x, y];
 }
 
+function handleRadialTextOffset(value) {
+    if (!value) return [0, 0];
```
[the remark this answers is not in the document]
pozdnyakov: The value has already passed `validate()`, also `isNaN` does not look quite reliable here as value can be both arrays or numbers

src/style-spec/validate/validate.js
pozdnyakov: yeah, this is what's happening below. We first check the spec type, if fail we check legacy type, if fail again - concatenate the errors.

ansis:
Is placing around an ellipse what we're looking for? It looks like placing around a rectangle might work better for https://github.com/mapbox/mapbox-gl-js/issues/8598

pozdnyakov:
Closing in favor of #8642

Issue #8069, opened 2019-03-22
Referrer error in iframe


Pessimistress:
**mapbox-gl-js version**:
0.53.1

**browser**:
Chrome 72.0.3626

### Steps to Trigger Behavior

 1. Save this page https://gist.github.com/Pessimistress/ae04af5ee5537652174f36a3a54cf508
 2. Serve with local dev server
 3. Map does not load. Error:

```
Uncaught TypeError: Failed to construct 'Request': Referrer 'http://127.0.0.1:8000http://127.0.0.1:8000/4b78be26-cdf7-42a9-a59f-097945b09779' is not a valid URL.
    at ajax.js:95
```

I only see this error when using dynamically generated iframe content.
This was working with mapbox-gl v0.50 and prior.

Looks like the issue was introduced here:
https://github.com/mapbox/mapbox-gl-js/blob/eec8cd5e43ee861b5202a1e1eb2e7d5739e0dfef/src/util/ajax.js#L82

If `window.top.location` is used instead of `window.location` the issue will go away.

kkaefer:
Thanks for reporting this issue. It looks like `window.top.location` works if the embedding document is loading the map document via a blob URL or data URL, but fails if it's loaded via a http URL that isn't the same origin as the embedding document.

Pessimistress:
I suspect this is a bug on Chrome's end. Though `window.location.pathname.startsWith(origin)` should be sufficient to catch it?

kkaefer:
Safari and Firefox have the same behavior. When a URL is loaded via blob URL, the _pathname_ is the full blob URL (e.g. `http://127.0.0.1:8000/4b78be26-cdf7-42a9-a59f-097945b09779`), which doesn't seem sensible. Using `window.top.location` would fail if the embedded document was loaded from a different origin than the host document, so I think we'll need to detect `blob`/`data` URLs and only use `window.top.location` (or `window.parent.location`?) in that case.

kkaefer:
https://github.com/mapbox/mapbox-gl-js/pull/8612 now uses the creating document's referrer in iframes loaded from a blob URL.

PR #8612, opened 2019-08-09
Fix resource loading edge cases

kkaefer:
This PR bundles a few fixes for edge cases in resource loading:

- Changes URLs in our debug pages to be relative, so that we can test them when loaded via the `file://` URL scheme as well (as long as they don't depend on other local resources)
- Fixes a bug that broke map loading when the document containing the map view was loaded from a blob URL: https://github.com/mapbox/mapbox-gl-js/issues/8069
- Classifies relative URLs as file:// URLs when the original document was loaded from a file URL to avoid using Fetch for those.

### DIFF
--- a/debug/2762.html
+++ b/debug/2762.html
@@ -4,7 +4,7 @@
     <title>Mapbox GL JS debug page</title>
     <meta charset='utf-8'>
     <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
-    <link rel='stylesheet' href='/dist/mapbox-gl.css' />
+    <link rel='stylesheet' href='../dist/mapbox-gl.css' />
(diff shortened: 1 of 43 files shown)


kkaefer:
Things to test:
- Are we sending correct referrers in all browsers?
- Does the new `debug/iframe-blob.html` page work in all browsers?
- Does the `debug/index.html` page work when loaded via a `file://` URL?

kkaefer:
IE doesn't support HTML documents loaded from data or blob URLs.

kkaefer:
Changelog entry:
* Fix map loading for documents loaded from Blob URLs
* Improve classification of relative URLs as file:// URLs when the document was loaded from a file URL

PR #8615, opened 2019-08-09
Number validation should fail on NaN

ryanhamley:
## Launch Checklist

It's not hard to imagine an expression or some other function unexpectedly producing `NaN` as an output to a style-spec property. Something like `text-max-width: ["sqrt", -1]` is a trivial example. Currently, `NaN` passes number validation because `typeof NaN === 'number'`. This could result in failures further downstream that may be hard to debug. This PR adds an explicit check for `NaN` values so that they fail number validation and throw a descriptive error.

This check is based on [a solution from MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Confusing_special-case_behavior) to filter out confusing behavior of `isNaN`. `Number.isNan` is more robust but is not supported in IE11.


ryanhamley:
@kkaefer I added a unit test and some runtime checks for this. Can you take another look?

ryanhamley:
This is a little weird to test. You can't set a property to `NaN` directly because it's not valid JSON and if you do something like `"text-size": ["sqrt", -1]`, it returns the default value from the style spec so there's no error. I did add a render test which shows that when an expression evaluates to `NaN` at runtime, it returns the default value and confirmed that the test fails without these changes.

Issue #8527, opened 2019-07-22
Enable variable label placement when {text|icon}-allow-overlap properties are set to true


tristen:
In the example below, if `text-allow-overlap` is set to `true` I would still expect variable label placement to work on finding the optimal display. Instead, it appears clustered like this:

![variable placement](https://user-images.githubusercontent.com/61150/61638298-1aa2da00-ac67-11e9-99dc-a376ea34570d.gif)

Here's what my layer looks like:

```js
{
  id: 'foo',
  type: 'symbol',
  source: 'composite',
  'source-layer': 'housenum_label',
  layout: {
    'text-field': 'I am a label',
    'text-size': 12.5,
    'text-allow-overlap': true,
    'text-variable-anchor': [
      'center',
      'left',
      'right',
      'top',
      'bottom',
      'top-left',
      'top-right',
      'bottom-left',
      'bottom-right'
    ],
    'text-justify': 'auto'
  }
}
```

---

Mapbox GL JS Version: **v1.1.1**
And a link to the demo: https://jsfiddle.net/tristen/tpaygf4w/

ansis:
Are you expecting it to try different positions looking for one without collisions and only allow a collision if there is no position without a collision?

tristen:
> Are you expecting it to try different positions looking for one without collisions and only allow a collision if there is no position without a collision?

@ansis right. Even if `*-allow-overlap` is true I would expect variable label placement to still try and find an optimized position before concluding there will be overlap.

chloekraw:
cc: @pozdnyakov @alexshalamov

pozdnyakov:
Implementation for gl-native https://github.com/mapbox/mapbox-gl-native/pull/15354

tristen:
Whoa! Amazing @pozdnyakov. Should this be left open for the GL JS portion of the work?

pozdnyakov:
> Should this be left open for the GL JS portion of the work?

Is it fine, if I make a separate issue in GL Native to track mapbox/mapbox-gl-native#15354

chloekraw:
@tristen https://github.com/mapbox/mapbox-gl-js/pull/8620 is in gl-js!

PR #8620, opened 2019-08-12
Enable variable placement with allowed text overlap

pozdnyakov:
The engine is trying to place the text first, and allows collision only if there is no position without a collision.

Fixes: https://github.com/mapbox/mapbox-gl-js/issues/8527


pozdnyakov:
@ansis thanks for review!

Issue #8583, opened 2019-08-01
icon-text-fit should work with text-variable-anchor


tristen:
### Expected Behavior

When using these properties on a layer:

```js
{
  'text-variable-anchor': ['left', 'top', 'right', 'bottom'],
  'text-justify': 'auto',
  'icon-text-fit': 'width'
}
```

I would expect both the icon that's been fitted under the label and the label to be variably placed.

### Actual Behavior

Here's a label using a icon-text-fit background 

| Without  `text-variable-anchor` | With `text-variable-anchor` |
| --- | --- |
| <img width="294" alt="Screen Shot 2019-08-01 at 9 08 34 AM" src="https://user-images.githubusercontent.com/61150/62296601-6c085180-b43d-11e9-8507-8967cbca989f.png"> | <img width="296" alt="Screen Shot 2019-08-01 at 9 17 17 AM" src="https://user-images.githubusercontent.com/61150/62296602-6c085180-b43d-11e9-9991-52d2999c65b6.png"> |

---

**mapbox-gl-js version**: v1.1.0
**browser**: Chrome `Version 75.0.3770.142 (Official Build) (64-bit)`

### Link to Demonstration

https://jsfiddle.net/tristen/830prsyz/

asheemmamoowala:
cc @alexshalamov

alexshalamov:
@asheemmamoowala yes, it is a bug, we need to offset an icon together with a text. Also, would be nice to document that in a spec as well.

chloekraw:
@alexshalamov, would be nice to document in the spec now that this property doesn't work with `icon-text-fit`, or to document that it _does_ after we close this issue?

alexshalamov:
@chloekraw please disregard my comment, after checking the spec, I think it should be clear how `icon-text-fit` and variable anchors work together, e.g., icon would be shifted together with the text.

alexshalamov:
gl-native fix is in https://github.com/mapbox/mapbox-gl-native/pull/14932/commits/19a49a5abbf53d40846af114ada0a9b1d25f3da0

Also, I had to it for fix combination of `text-writing-mode` with `icon-text-fit`

chloekraw:
@alexshalamov thanks for the investigation and proposed fix. per chat, I'm tracking that your next steps here are to split those commits out into a separate PR on GL-Native and to open a PR with render tests on GL-JS.

alexshalamov:
[...]

PR #8625, opened 2019-08-13
Render tests for icon-text-fit combined with text-variable-anchor and text-writing-mode

alexshalamov:
Render tests required for https://github.com/mapbox/mapbox-gl-native/pull/15367 and mapbox/mapbox-gl-js#8583


alexshalamov:
@ansis what do you think about updated tests with 1px border diff? Also, since the collision box for a icon-text-fit icon has proper size (bigger), it affects symbols with line placement ([#4861](https://github.com/mapbox/mapbox-gl-js/issues/4861)), therefore, I've updated one line placement test.

If you think these changes are acceptable and I can proceed with https://github.com/mapbox/mapbox-gl-native/pull/15367, we can land new render tests and I will continue to work with native PR tomorrow.

ansis:
I checked with Saman and he thinks this should be ok from the map design side! Seems ok to me as well.

PR #8626, opened 2019-08-13
[style-spec] export validateExpression globally

rmrice:
## Launch Checklist

Expressions are hard and it would be nice to validate an expression without validating an entire style.json. This PR globally exports `validateExpression`, so it can be used in projects like:

```js
const { validateExpression } = require('@mapbox/mapbox-gl-style-spec')

validateExpression({'value': ["==", "some", "expression" ]})
```

cc @mapbox/map-design-team @mapbox/api-gl @mapsam 

 - [ ] post benchmark scores

### DIFF
--- a/src/style-spec/style-spec.js
+++ b/src/style-spec/style-spec.js
@@ -75,6 +75,7 @@ import convertFunction from './function/convert';
 import { eachSource, eachLayer, eachProperty } from './visit';
 
 import validate from './validate_style';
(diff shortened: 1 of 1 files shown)


asheemmamoowala:
Had an offline conversation with @rmrice. Their use case is for validating HTTP requests with parameters to be passed to  `Map#addLayer` and `Map#setFilter` implementations. These methods already perform validation, and a separate validation should not be required.

PR #8633, opened 2019-08-14
Improve cancellation of tile loads

kkaefer:
Our `Actor` object is responsible for communicating with other threads, either by sending messages from the main thread to web workers, or vice versa. When calling `Actor#send()`, calls that provided a callback returned a `Cancelable` object that supported calling `.cancel()` for aborting the request. This only worked in a very small number of instances: When the action that was triggered in the corresponding thread was asynchronous, and supported cancellation itself.

This PR changes this to allow cancellation of a much wider array of requests. During high pressure workload phases, e.g. while zooming rapidly, tasks could pile up in the worker: in particular `loadTile` messages, that were followed by `abortTile` messages for the same tile uid. This PR now allows canceling messages while the _wait_ in the worker's queue to be processed. This keeps the task from ever starting in the worker thread. When zooming quickly while loading tiles, this means that we now issue fewer HTTP requests that get cancelled immediately. Similarly, canceling HTTP requests on the worker were delayed by computing tasks on the worker.

Before this PR, the Actor processed messages sent from the counterpart via `postMessage` synchronously. This means that starting a task, followed immediately by a cancellation never actually canceled the task, because the worker first processed the task, then the cancelation. In this PR, I decoupled the message event from processing by putting tasks in a queue, then triggering processing with `setTimeout(..., 0)`. This means that at most 1 task is processed per event loop iteration. This allows message processing run way more frequently, and consequently allows us to process cancel messages more quickly.

I opted for this model instead of a more generic priority queue because we don't have any use cases for a priority queue other than canceling tasks.

This PR also contains the addition of a `loaded` method to `Source` objects. This is necessary to correctly classify the loaded state of a source because we can now no longer rely on all tasks sent to a worker being processed

### DIFF
--- a/src/source/canvas_source.js
+++ b/src/source/canvas_source.js
@@ -111,6 +111,7 @@ class CanvasSource extends ImageSource {
      */
 
     load() {
+        this._loaded = true;
         if (!this.canvas) {
             this.canvas = (this.options.canvas instanceof window.HTMLCanvasElement) ?
                 this.options.canvas :
--- a/src/source/geojson_source.js
+++ b/src/source/geojson_source.js
@@ -11,6 +11,7 @@ import type {Source} from './source';
 import type Map from '../ui/map';
 import type Dispatcher from '../util/dispatcher';
 import type Tile from './tile';
+import type Actor from '../util/actor';
 import type {Callback} from '../types/callback';
 import type {GeoJSON, GeoJSONFeature} from '@mapbox/geojson-types';
 import type {GeoJSONSourceSpecification} from '../style-spec/types';
@@ -74,9 +75,8 @@ class GeoJSONSource extends Evented implements Source {
     _data: GeoJSON | string;
     _options: any;
     workerOptions: any;
-    dispatcher: Dispatcher;
     map: Map;
-    workerID: number;
+    actor: Actor;
     _loaded: boolean;
     _collectResourceTiming: boolean;
     _resourceTiming: Array<PerformanceResourceTiming>;
@@ -102,7 +102,7 @@ class GeoJSONSource extends Evented implements Source {
         this._removed = false;
         this._loaded = false;
 
-        this.dispatcher = dispatcher;
+        this.actor = dispatcher.getActor();
         this.setEventedParent(eventedParent);
 
         this._data = (options.data: any);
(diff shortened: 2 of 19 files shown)


ansis:
This is really great!

> Read through the changes and didn't find any obvious things to fix, but there's a ton of logic so very easy to miss something. But hopefully we'd catch any serious issues during manual and automated testing

Yep, I agree. I did some manual testing on a bunch of the debug pages and didn't notice anything.

mannnick24:
@mourner and @kkaefer I believe there is now a race condition related to triggerRepaint in the map.render

emikoshi:
> @mourner and @kkaefer I believe there is now a race condition related to triggerRepaint in the map.render
> 
> in map.js is this
> if (this._sourcesDirty || this._repaint || this._styleDirty || this._placementDirty) {
> this.triggerRepaint();
> } else if (!this.isMoving() && this.loaded()) {
> this.fire(new Event('idle'));
> }
> and I think the tiles may not have loaded, but sourcesDirty could be false,
> 
> anyhoo, on IE we may or may not get a render after the map reports loaded() our product always waits for a a render with the map loaded (if dataloading) so our viz hangs in IE
> 
> Here is a fiddle, that may or may not trigger the fitbounds in IE - chrome seems to always pass
> [loadedrender.zip](https://github.com/mapbox/mapbox-gl-js/files/3551920/loadedrender.zip)

I'm getting this to happen pretty consistently in Chrome (Version 76.0.3809.132) with dev tools opened and cache disabled. We've been using version 1.2.1 of Mapbox for a while now and never had such issue.

Similar to what @mannnick24 was reporting, we have calls that happen after map load/render and they are not being called since it's visibly clear some tiles aren't loaded. I've done side by sides in our project between 1.2.1 and 1.3 with nothing else updated and still only happens in 1.3.

Looking at the changelog this commit/issue seemed to be the most relevant, hopefully this is in the right place. I've attached a screenshot just for context. Some other notes:

- No errors in dev tools.
- Doesn't seem to happen in Firefox with/without dev tools.
- We use webpack as a module bundler so mapbox is imported via `import mapboxgl from "mapbox-gl";`

Screen Shot 2019-09-04 at 2 37 53 PM

Issue #8328, opened 2019-06-07
Port asymmetric viewport PR#14664 from gl-native to gl-js


astojilj:
Port https://github.com/mapbox/mapbox-gl-native/pull/14664 to gl-js.

andrewharvey:
Done in #8638

Issue #4268, opened 2017-02-14
Add a global "padding" option


lucaswoj:
We allow users to pass an [`around` option](https://www.mapbox.com/mapbox-gl-js/api/#CameraOptions) to some camera methods. 

This ticket proposes a `around` option be added to `Map` as a constructor option, getter, and setter. 

This option would specify a default value for `around` on all camera mutations so that

 - it becomes possible to specify `around` for implicit camera mutation, such as those triggered by use of the `NavigationControl`
 - it becomes more ergonomic to use the same `around` value for all camera mutations

# Open Questions

 - Does `around` make sense semantically in this case? Many users will be using this feature because they have a translucent element over the map. In that use case, `padding` is a more intuitive option.
 - Might users be confused about the difference between `around` and `center`? Can we find a naming convention that eliminates that confusion?
 - Are there any other parameters that should have global defaults like this?

1ec5:
A `padding` option would avoid the potential confusion between `center` and `around`. We do need to make sure `padding` accepts all four edges independently, both for consistency with related style specification properties and so that we address the use case that originally prompted this cluster of tickets.

Another important aspect is that it becomes all the more important to provide a method that returns the current CameraOptions, accepting a padding option. The developer should be able to call something like `jumpTo(getCamera(padding), padding)` and count on the viewport to stay constant.

mourner:
I don't think it makes sense to have a global `around` option, because the value you need for it depends on context (e.g. the current map view). I have a hard imagining a use case where it would be static.

`padding` makes sense though because it's usually static. I would introduce independent padding for 4 sides because it's one of the most common use case — e.g. having a semi-transparent sidebar on the map.

musicformellons:
It's been a while since I brought up this issue referring then to it as 'shifted center'. I am now trying to establish what current status is. I note that in the meantime padding added, but am I right that then you would have to set all flyTo etc kind of operation explicitly with offset to the 'shifted center', whereas this issue would have this setup instantly for all kinds of zoom operations?

mourner:
@musicformellons yes, that's right. I still think a global map padding option makes more sense than individual `offset` options in various camera methods, but we don't have a timeline for implementing this at the moment.

jaapster:
I would also prefer global map padding as this would also apply to box zooming.

stevage:
Thanks @mourner for noticing this duplicate. Copying my text in from that issue since I think it covers another use case, with much the same proposed solution:

[...]

PR #8638, opened 2019-08-15
Port asymmetric viewport rendering to gl-js

arindam1993:
Addresses the following issues:
https://github.com/mapbox/mapbox-gl-js/issues/8328
https://github.com/mapbox/mapbox-gl-js/issues/4268
and ports @astojilj 's amazing work to the js side.

### API:
Users can now specify global `left`, `right`, `top` and `bottom` padding values (in pixels) in in a `padding` field in camera functions  i.e `jumpTo`, `easeTo` and `flyTo`.
Example:
```
map.easeTo({{
    padding: {
         left: 400,
         right: 100
    } 
)};
```
Similar to other camera options the existing padding value is retained if a parameter is unspecified in the call.
Also, similar to @astojilj 's design, the rendering changes are implemented accounting for the offset center elements 8 and 9 of `Transform#projMatrix`, in `Transform#_calcMatrices()`, while ensuring that everything that reads from `Transform#getCenter` doesn't break with the center being offset.

### Testing:
- Unit tests for `EdgeInsets` class that tests its behavior in isolation
- Unit tests in `camera` for testing if its interface oh setting `padding` behave as expected.
- Added debug rendering for padding: 
Screen Shot 2019-08-19 at 3 21 36 PM
Left padding rendered in blue,
Bottom padding rendered in green
Right padding rendered in magenta
Top padding rendered in red
This can be turned on with the map option `showPadding`.
- Add render tests with `showPadding` set to `true`

## Launch Checklist

 - [x] address overlap between `padding` in `fitBounds` and padding for camera options for asymmetric viewport
<changelog>Map API functions such as `easeTo` and `flyTo` now support `padding: PaddingOptions` which lets developers shift the center of perspective for a map when building floating sidebars.</changelog>


astojilj:
@arindam1993 
There is a follow up work that needs to be ported:
https://github.com/mapbox/mapbox-gl-native/pull/15195

It prevents out of memory (too many tiles loaded) with large top padding:

src/geo/edge_insets.js
```diff
@@ -0,0 +1,105 @@
...
+     * This is in pixels and with the top left being (0.0) and +y being downwards.
+     *
+     * @param {number} width
+     * @param {number} height
+     * @returns {Point}
+     * @memberof EdgeInsets
+     */
+    getCenter(width: number, height: number): Point {
+        // Clamp insets so they never overflow width/height and always calculate a valid center
+        const totalXInset = Math.min(this.left + this.right, width);
```
astojilj: could we do this when values for size and padding are set, as opposite to handling it here?

src/geo/transform.js
```diff
@@ -185,6 +190,52 @@ class Transform {
         this._calcMatrices();
     }
 
+    get padding(): EdgeInsetJSON { return this._edgeInsets.toJSON(); }
```
astojilj: Serving the existing API in native, i didn't like that in Android we use "padding" and in iOS "edgeInsets". is padding here added to keep it close to native implementation?

arindam1993: I just did it so the public facing term for is it simpler, but happy to change it to whatever.
Maybe @chloekraw , do you have any thoughts?

chloekraw: For this feature, I think `padding` is the term that's going to be closest to what web developers who use CSS would expect. `inset` is most commonly used as a type of border style:
- https://developer.mozilla.org/en-US/docs/Web/CSS/padding
- https://developer.mozilla.org/en-US/docs/Web/CSS/border-style

I did find this https://developer.mozilla.org/en-US/docs/Web/CSS/inset "experimental technology" but it sounds like this `inset` is more about defining margins around blocks of text and is less analogous to a "viewport." (related: https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block, https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline)

@arindam1993 are there other parts of the GL-JS code base that also use this idea of padding?

test/unit/ui/camera.test.js
```diff
@@ -316,6 +316,34 @@ test('camera', (t) => {
...
+    t.test('#setPadding', (t) => {
+        t.test('sets padding', (t) => {
+            const camera = createCamera();
+            const padding = { left: 300, top: 100, right: 50, bottom: 10};
+            camera.setPadding(padding);
+            t.deepEqual(camera.getPadding(), padding);
+            t.end();
+        });
+
+        t.test('doesnt change padding thats already present if new value isnt passed in', (t) => {
```
astojilj: test that padding is retained when no padding is set to camera might be useful, too

src/geo/transform.js
astojilj: @arindam1993 , related to this - what's your opinion about propagating CSS style padding specified to Map's parent HTML container to transform?

arindam1993: @chloekraw I don't think so.

@astojilj I don't think we should, I think devs waning to create floating UI elements, which this feature is designed for would want to set a `padding` on the container, since it'd offset all their UI elements which I assume would be appended as children fo the `map` container.

However I think it might be useful to push the current state of `padding` to our own CSS classes and DOM elements which we use for nesting `NavigationControl` and `AttributionControl`
Screen Shot 2019-08-20 at 3 21 37 PM
`top`,`left`, `right` and `bottom` attributes are always 0, i think we can update them from within `Map`.

Having `Map` be the only API entrypoint seems more intuitive to me, with plugins automatically respecting it as well.

arindam1993: Here's a demo of what I mean.

src/geo/edge_insets.js
arindam1993: It depends on the `width` and `height` of the current viewport, which can change as the user resizes the map. 
The transform could explicitly issue recalculate calls, but that seems somewhat fragile.

test/unit/ui/camera.test.js
arindam1993: Added 👍

src/geo/edge_insets.js
```diff
@@ -0,0 +1,105 @@
...
+            top: this.top,
+            bottom: this.bottom,
+            left: this.left,
+            right: this.right
+        };
+    }
+}
+
+export type EdgeInsetLike = EdgeInsets | {top?: number, bottom?: number, right?: number, left?: number} | EdgeInsetJSON;
+export type EdgeInsetJSON = {top: number, bottom: number, right: number, left: number}
```
astojilj: Why not using PaddingOptions?

src/render/draw_debug.js
```diff
@@ -9,15 +9,60 @@ import SegmentVector from '../data/segment';
...
 import Color from '../style-spec/util/color';
+import browser from '../util/browser';
 
 import type Painter from './painter';
 import type SourceCache from '../source/source_cache';
 import type {OverscaledTileID} from '../source/tile_id';
 
 export default drawDebug;
 
+const topColor = new Color(1, 0, 0, 0.4);
```
astojilj: Does draw_debug.js go to production code?
I'm not sure how useful is to including additional shader and this code for rendering different colors or padding values.
For manual debugging, propose to add additional debug/***.html page (doesn't go to production code AFAIK) with overlays (in the way the customers are supposed to use it)...

> - [x] manually test the debug page

and in that that page would be used for this step.

Please doublecheck my opinion with other reviewers - I might be missing something...

ansis: Thanks for adding a debug view. They can be pretty useful for understanding things. If the size and complexity of the shader another approach could be to use a gl scissor and clear to draw a line at each edge.

```js
    gl.enable(gl.SCISSOR_TEST);
    gl.scissor(x, y, width, height);
    context.clear({ color });
    gl.disable(gl.SCISSOR_TEST);
```

arindam1993:
I'm closing this for now since I haven't been actively working on it. Will reopen once I have active progress on it again.

asheemmamoowala:
Reopening this ticket since it is pending completion while blocked by #8975

ryanhamley:
@arindam1993 #8975 was merged. Is anything else blocking this PR?

arindam1993:
Nope, its just addressing comments and testing and getting it back upto speed.

arindam1993:
@mpulkki-mapbox 
Your LOD PR integrates with this really well! 🎉  nice work! No more runaway tile downloads or performance slowdows when using a  large top padding!

src/ui/camera.js
```diff
@@ -646,11 +684,17 @@ class Camera extends Evented {
                 .fire(new Event('pitchend', eventData));
         }
 
+        if (paddingChanged) {
+            this.fire(new Event('paddingstart', eventData))
```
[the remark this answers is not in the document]
arindam1993: Some kind of UI state synchronization?
for example, if you want a floating sidebar to be extended out exactly as much as the map's current value of padding, you can use the events to synchronize that state.

src/geo/transform.js
```diff
@@ -550,6 +603,10 @@ class Transform {
         let m = new Float64Array(16);
         mat4.perspective(m, this._fov, this.width / this.height, nearZ, farZ);
 
+        //Apply center of perspective offset
+        m[8] = -offset.x * 2 / this.width;
+        m[9] = offset.y * 2 / this.height;
+
```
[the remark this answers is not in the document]
arindam1993: the only api method that allows that I think can work is the asymmetric view frustum method but that takes 4 different fov's as input, and back calculating that seemed more janky to me.

src/geo/edge_insets.js
```diff
@@ -0,0 +1,105 @@
...
+     */
+    getCenter(width: number, height: number): Point {
+        // Clamp insets so they never overflow width/height and always calculate a valid center
+        const totalXInset = Math.min(this.left + this.right, width);
+        const totalYInset = Math.min(this.top + this.bottom, height);
+
+        const x = Math.min(this.left, width) + 0.5 * (width - totalXInset);
+        const y = Math.min(this.top, height) + 0.5 * (height - totalYInset);
+
+        return new Point(x, y);
```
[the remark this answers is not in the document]
arindam1993: I think the simpler one you suggested is better! 
Do you think we should throw an error/warning when the overflow happens?

arindam1993:
@asheemmamoowala  @astojilj  @ansis , I think I've addressed your previous comments, could you all please take another pass at this, I think this is ready to go now

arindam1993:
Aso, @chloekraw  its all called padding now

src/geo/edge_insets.js
```diff
@@ -0,0 +1,103 @@
...
+     * @param {PaddingOptions} target
+     * @param {number} t
+     * @returns {EdgeInsets}
+     * @memberof EdgeInsets
+     */
+    interpolate(start: PaddingOptions | EdgeInsets, target: PaddingOptions, t: number): EdgeInsets {
+        if (target.top != null && start.top != null) this.top = number(start.top, target.top, t);
+        if (target.bottom != null && start.bottom != null) this.bottom = number(start.bottom, target.bottom, t);
+        if (target.left != null && start.left != null) this.left = number(start.left, target.left, t);
+        if (target.right != null && start.right != null) this.right = number(start.right, target.right, t);
```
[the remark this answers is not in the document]
arindam1993: this logic ensures that the  `if the key is not present the value does not change change from the existing value` logic for adding padding. Defaulting to 0 would remove the padding. and having it here ensures that that logic does not have to be implemented in every public interface function that accepts padding as input.

src/render/draw_debug.js
```diff
@@ -11,13 +11,63 @@ import StencilMode from '../gl/stencil_mode';
...
+function drawDebugSSRect(painter: Painter, x: number, y: number, width: number, height: number, color: Color) {
+    const context = painter.context;
+    const gl = context.gl;
+
+    gl.enable(gl.SCISSOR_TEST);
+    gl.scissor(x * browser.devicePixelRatio, y * browser.devicePixelRatio, width * browser.devicePixelRatio, height * browser.devicePixelRatio);
+    context.clear({color});
+    gl.disable(gl.SCISSOR_TEST);
+}
+
```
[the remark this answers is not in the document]
arindam1993: Not super often, no actually, but it seems like an useful tool to have for render tests, examples and debug pages. I could make it so it is only available in the `dev` build, but it will break our usual convention.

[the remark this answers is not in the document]
arindam1993: I've also added some new render tests in the newest commit that represents the distortion caused by asymmetric viewport much better.
https://github.com/mapbox/mapbox-gl-js/pull/8638/files#diff-633f8e084d5e1ba0bb38d2c145e1b771

src/ui/camera.js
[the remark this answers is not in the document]
arindam1993: tru, removed them for simplicity!

src/geo/edge_insets.js
```diff
@@ -0,0 +1,103 @@
...
+            left: this.left,
+            right: this.right
+        };
+    }
+}
+
+export type PaddingOptions = {top: ?number, bottom: ?number, right: ?number, left: ?number};
+export type SerializedPadding = {top: number, bottom: number, right: number, left: number};
+
+export default EdgeInsets;
```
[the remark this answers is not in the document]
arindam1993: It could be, but this seems easier to write unit tests for, and handles a lot of the validation and updating logic without adding additional private methods to the transform class.

arindam1993:
Added some fresh render tests that visualize the effect of the distortion better.
expected
expected

arindam1993:
YAY merging!!!! thanks everyone for the detailed review! 🙇

Issue #8618, opened 2019-08-10
Possible to rotate / tilt a map with interactive: false via compass icon


pathmapper:
**mapbox-gl-js version**: v1.2.1

**browser**: Chrome/Firefox, likely any

### Steps to Trigger Behavior

https://docs.mapbox.com/mapbox-gl-js/example/animate-camera-around-point/ with 
`interactive: false` and
```
map.addControl(new mapboxgl.NavigationControl({
  visualizePitch: true,
  showZoom: false
}));
```

### Link to Demonstration

https://jsbin.com/yatuvegeti/1/edit?html,output

### Expected Behavior

It's not possible for the user to interrupt/alter the animation. Also the compass tooltip _Reset bearing to north_ should not be displayed.

### Actual Behavior

It's possible to rotate / tilt the map dragging the compass icon and the tooltip is displayed.

---

One could argue that this is not a bug, because the compass is a navigation control. In this case it would be appreciated if this issue would be considered as feature request. It would be useful to be able to use the compass for indicating bearing/pitch while animating the camera and the user is not able to interrupt/alter the animation.

peterqliu:
I'm unclear about what this issue is requesting. 

>It would be useful to be able to use the compass for indicating bearing/pitch while animating the camera and the user is not able to interrupt/alter the animation.

This is the current behavior.

>One could argue that this is not a bug, because the compass is a navigation control.

Agree, preserving compass functionality despite `interactive: false` is intentional.

pathmapper:
> This is the current behavior.

Don't think so. Currently it is possible for the user to interupt/alter the animation via the compass icon (gif from the demonstration linked above):

![compass_interrupt_animation](https://user-images.githubusercontent.com/20856381/62897015-936ef080-bd52-11e9-81fe-9c6e3dc14d96.gif)

> I'm unclear about what this issue is requesting.

Show the compass while animating the camera and this animation cannot be interrupted/altered by the user. Hope it is clearer now.

peterqliu:
>Hope it is clearer now.

It's still fairly confusing, but if this is solely about removing compass interactivity when map is `interactive: false`, then we'll decline for now.

peterqliu:
Per chat with several other developers, this could be desirable. Reopening

danafillip:
It will stop on its own

danafillip:
Disable compass interactivity for `interactive: false` maps[/+]

danafillip:
[...]

PR #8643, opened 2019-08-15
Disable compass button mouse/touch handlers when map#interactive is false

ryanhamley:
## Launch Checklist

    - fixes https://github.com/mapbox/mapbox-gl-js/issues/8618
    - when `map#interactive` is set to `false`, we should disable map interactivity via the Navigation Control's compass button as well

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -104,10 +104,12 @@ class NavigationControl {
             }
             this._map.on('rotate', this._rotateCompassArrow);
             this._rotateCompassArrow();
-            this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass});
-            DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
-            DOM.addEventListener(this._compass, 'touchstart', this._handler.onMouseDown, { passive: false });
-            this._handler.enable();
+            if (this._map._interactive) {
+                this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass});
+                DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
+                DOM.addEventListener(this._compass, 'touchstart', this._handler.onMouseDown, { passive: false });
+                this._handler.enable();
+            }
         }
         return this._container;
     }
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -174,7 +174,7 @@ const defaultOptions = {
  *
  * @param {boolean} [options.hash=false] If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
  *   For example, `http://path/to/my/page.html#2.59/39.26/53.07/-24.1/60`.
- * @param {boolean} [options.interactive=true] If `false`, no mouse, touch, or keyboard listeners will be attached to the map, so it will not respond to interaction.
+ * @param {boolean} [options.interactive=true] If `false`, no mouse, touch, or keyboard listeners will be attached to the map, so it will not respond to user interaction. This option also disables user interaction with the map through the Navigation Control's compass button.
  * @param {number} [options.bearingSnap=7] The threshold, measured in degrees, that determines when the map's
  *   bearing will snap to north. For example, with a `bearingSnap` of 7, if the user rotates
  *   the map within 7 degrees of north, the map will automatically snap to exact north.


asheemmamoowala:
@ryanhamley Will the compass button look differently when the map is not interactive? How would a user know that the button is not going to do anything?

andrewharvey:
I think that when interactive is false:

1. the button should be `disabled`, so that the cursor doesn't change to the pointer, and so that you can't tab select it

2. We should reset the `title="Reset bearing to north" aria-label="Reset bearing to north"` since these are no longer true when you can't click on the button.

3. We should disable its hover effect. The hover effect indicates you can click on the button, but not when it's disabled.

ryanhamley:
I think @andrewharvey is right. I got a little ahead of myself on this PR. I'd say that this should apply to all 3 navigation control buttons (zoom in/out and compass) as well as geolocate, but it should not apply to the full screen control or any other built in control.

peterqliu:
@ryanhamley there are many ways to do this and your current one totally works, but the most concise method is probably a new CSS class that conditionally append to the whole navigation module with has `pointer-events:none;`, and fades out/hides/somehow indicates the zoom buttons are inactive.

This way, all you have to do in JavaScript is toggle the class and set the title aria.

andrewharvey:
> We should reset the title="Reset bearing to north" aria-label="Reset bearing to north" since these are no longer true when you can't click on the button.

Except now I'm second guessing myself, maybe the title and aria-label should be left in, as they help tell the user what the button would do, if it were enabled. Otherwise you get this disabled button which sighted users can see as a compass/zoom in/zoom out and see that's it is disabled, but unsighted users see there is a disabled button there but don't know what it is for without the title/aria-label.

So maybe just disabled is enough and the label should remain.

pathmapper:
> So maybe just disabled is enough and the label should remain.

Then there would be still a tooltip for the compass icon, right? 

It could confuse sighted users if there is a tooltip "Reset bearing to north" but it's not possible to reset the bearing. So if a title/label should remain, it should be changed to something else.

There should be no fade out for the compass icon, because it still has a purpose when `map#interactive` is set to `false` (visualising bearing and pitch, the latter if the `visualizePitch` option is set).

andrewharvey:
> Then there would be still a tooltip for the compass icon, right?
>
> It could confuse sighted users if there is a tooltip "Reset bearing to north" but it's not possible to reset the bearing. So if a title/label should remain, it should be changed to something else.

Yes that was my original thinking that the label was confusing since it says the button will reset the bearing, but if you try to click nothing happens so it doesn't really do that.

But on the other hand having no label isn't an option because sighted users can see the + - and compass icon, but even when disabled we need to communicate through text those symbols for accessibility.

> There should be no fade out for the compass icon, because it still has a purpose when map#interactive is set to false (visualising bearing and pitch, the latter if the visualizePitch option is set).

But if it's a `button` then if you can't click on it or clicking does nothing, then it should be have the `disabled` attribute right?

peterqliu:
>But if it's a button then if you can't click on it or clicking does nothing, then it should be have the disabled attribute right?

the compass still has value as an indicator -- the `reset north` functionality is arguably ancillary, even-- so visibility is still important. `disabled` would fade it out a bit

pathmapper:
> disabled would fade it out a bit

Maybe this could be avoided using the  [`:disabled` CSS pseudo-class ](https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled) to style the `disabled` compass the same as the `enabled` one.

ryanhamley:
I'm going to close this for the time being while we consider https://github.com/mapbox/mapbox-gl-js/issues/9036. This issue/PR would be moot if we simply removed the NavigationControl's interactivity entirely

pathmapper:
> This issue/PR would be moot if we simply removed the NavigationControl's interactivity entirely

@ryanhamley agree, but if I understood it right, #9036 is about removing only the drag rotation on NavigationControl compass, not about removing the interactivity entirely.

So after #9036 the issue (#8618) would still exist with the _Reset bearing to north_ functionality of the compass:
https://jsfiddle.net/2j7qp86h/

reset_bearing_to_north

ryanhamley:
@pathmapper Good point, I forgot about that functionality. I do think I'll keep this closed for the time being because we have a number of refactors planned in the near-to-mid-term around mobile gestures and accessibility so this may work better as part of that push. This *probably* should be revived when we do those refactors though so that the reset bearing functionality is disabled with `map#interactive: false`.

PR #8650, opened 2019-08-16
Fix debug pages

ryanhamley:
## Launch Checklist

    - The changes to the debug page in https://github.com/mapbox/mapbox-gl-js/pull/8612 broke the debug pages when served locally via the `yarn start-debug` command. This is the only configuration I've found to allow the debug pages to work properly when served locally and when tested with a file:// URL

### DIFF
--- a/debug/2762.html
+++ b/debug/2762.html
@@ -15,7 +15,7 @@
 <div id='map'></div>
 
 <script src='../dist/mapbox-gl-dev.js'></script>
-<script src='access_token_generated.js'></script>
+<script src='../debug/access_token_generated.js'></script>
 <script>
 
 var map = new mapboxgl.Map({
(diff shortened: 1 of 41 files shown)


mourner:
Debug pages through `yarn run start-debug` work fine for me on `master` — I wonder why they're broken on your side.

ryanhamley:
Well that's weird. I always see `http://localhost:9966/access_token_generated.js net::ERR_ABORTED 404` and the page fails to load. I've tried on Node 8 and 10 just to rule out that possibility and I tried in multiple browsers.

ryanhamley:
Ah so, I have always just gone to `localhost:9966/debug` in my browser. That doesn't work anymore. The URL has to conclude in `/` or the path to an html file. So `localhost:9966/debug/` works fine. My mistake.

PR #8657, opened 2019-08-19
Don't cache server requested glyphs in the local glyph ranges

### LINKED ISSUE #8653: Locally generated glyphs shift on over-zoomed tiles
When zooming from z17 to z18, I am seeing some locally generated glyphs being replaced by server retrieved glyphs. This is demonstrated in the gifs below. The second gif shows the box/size of the glyph made by adjusting the fragment shader. 

On z18 tiles, the glyphs have variable size which is a property of the server generated glyphs. The z17 tiles have fixed size (24/32 px sq.) size that is characteristic of locally generated glyphs.

**mapbox-gl-js version**: v1.2.1

**browser**: Chrome

### Steps to Trigger Behavior

 1. With the local debug server open http://localhost:9966/debug/debug.html#18.04/35.681234/139.770108/-0.1
 2. Zoom between z17 and z18

### Link to Demonstration
localIdeo_shift

localIdeo_shift_debug

### Expected Behavior
Locally generated glyphs are not overridden by server generated ones.

### Actual Behavior
At z18 server glyphs are taking over. Causing a change in the baseline and font of some characters.

cc @ansis @chloekraw @tristen

asheemmamoowala:
Fixes #8653.

Font glyphs from the server are requested in packed in ranges of 256 characters. In some rare cases a single range may contain characters that can be locally generated (using tinySDF) and that need to be fetched from the server.

The issue described in #8653 stems from a character from such an overlapping range causing the remote glyphs being cached for even those characters that should be locally generated. This would be a non-issue if both local generation and the server request were resolving to the same font.

The fix is to check each character in the fetched range and skip it if it is should be generated locally. An additional improvement in this PR is to cache the locally generated glyphs so that subsequent tiles don't need to re-generate them.

I also found a bug in `Map#setStyle` while debugging this issue that uses the wrong `default` font family for local glyph generation. It should use the value resolved in the map constructor.

## Launch Checklist

 - [ ] post benchmark scores
 - [x] tagged @mapbox/map-design-team 
 - [x] tagged @mapbox/gl-native for native port

cc @chloekraw @tmpsantos

### DIFF
--- a/src/render/glyph_manager.js
+++ b/src/render/glyph_manager.js
@@ -64,6 +64,7 @@ class GlyphManager {
 
             glyph = this._tinySDF(entry, stack, id);
             if (glyph) {
+                entry.glyphs[id] = glyph;
                 callback(null, {stack, id, glyph});
                 return;
             }
--- a/src/render/glyph_manager.js
+++ b/src/render/glyph_manager.js
@@ -82,7 +82,7 @@ class GlyphManager {
                     (err, response: ?{[number]: StyleGlyph | null}) => {
                         if (response) {
                             for (const id in response) {
-                                if (!this._doesCharSupportLocalGlyph(id)) {
+                                if (!this._doesCharSupportLocalGlyph(+id)) {
                                     entry.glyphs[+id] = response[+id];
                                 }
                             }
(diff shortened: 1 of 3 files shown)


src/render/glyph_manager.js
```diff
@@ -64,6 +64,7 @@ class GlyphManager {
 
             glyph = this._tinySDF(entry, stack, id);
             if (glyph) {
+                entry.glyphs[id] = glyph;
```
ansis: What does this line change? Were we previously not caching generated glyphs?

asheemmamoowala: > Were we previously not caching generated glyphs?

That's right. Locally generated glyphs were not being cached before this PR in gl-js. Native was already doing it correctly. I considered creating a separate stack for local fonts - so that each local font would only have one cache entry. The current approach can lead to one entry per font stack (in the style definition), but is already an improvement over re-generating each local glyph per tile. Those changes could probably be addressed in a separate PR.

PR #8659, opened 2019-08-20
Fix icon/text collisionBox translation

zmiao:
Fix: js part of https://github.com/mapbox/mapbox-gl-native/issues/13526

The root cause for it is that when rendering the symbols, the translate offset is only applied to icon/text, but the corresponding `collisionBox` is ignored. Also `collisionBox` in current `SymbolBucket` needs to be split for icon and text symbol separately, in case they have different translate parameter.

## Launch Checklist


zmiao:
This change leads to one failing render test debug/collision. As the rendering order may be changed for icon and text collisionBox, as they are now separated. Results are depicted as following:
expected: 
expected 

actual: 
actual

diff: 
diff

asheemmamoowala:
@zmiao It looks safe to update the fixture for that test.

zmiao:
@ansis Whitespace is moved + changelog is added. Thanks for heads up.

PR #8663, opened 2019-08-20
Add api-supported-mode to spec.validate

samanpwbb:
The style API has stricter stylesheet validation than Mapbox GL JS. This PR adds a new mode to the validator that will inform the user if their style is fit to be uploaded to the style API. The purpose of this PR is to add transparency to the style API and unify test suites for both validation steps.

Changelog suggestion:
- Add a new validation function that returns the same error responses as the Mapbox Styles API. Use as a CLI tool with `gl-style-validate --api-supported-mode`. Use in Javascript by importing `validateMapboxApiSupported` from the `@mapbox/mapbox-gl-style-spec` package.

## Launch Checklist

 - [x] Finish the code
 - [ ] post benchmark scores


src/style-spec/validate/validate_mapbox_api_supported.js
```diff
@@ -0,0 +1,95 @@
...
+            errors.push(new ValidationError('sources.url', null, 'Style must reference sources hosted by Mapbox'));
+        }
+
+        validateKeys(sources[q], SOURCE_KEYS, errors, 'source');
+    }
+
+}
+
+function validate(style: Object, errors: Array<?Error>): void {
+    if (style.version !== SUPPORTED_SPEC_VERSION) {
```
asheemmamoowala: What happens to older styles - are they no longer valid, or auto updated on the backend?

src/style-spec/validate/validate_mapbox_api_supported.js
```diff
@@ -0,0 +1,95 @@
...
+        errors.push(new ValidationError('version', null, `style version must be ${SUPPORTED_SPEC_VERSION}`));
+    }
+
+    if (style.sprite && !isValidGlyphsURL(style.glyphs)) {
```
asheemmamoowala:
```suggestion
    if (style.glyhps && !isValidGlyphsURL(style.glyphs)) {
```

src/style-spec/validate/validate_mapbox_api_supported.js
samanpwbb: They are no longer supported by the style API.

samanpwbb:
Okay, this is ready for review. I opted for creating a validation function that is fully independent of the existing recursive validation function. The existing system highly depends on the style-spec config file, while my new validator is a little more imperative. I considered writing api-style-supposed mode as a new version of spec instead (or rather, as a set of overrides that would be merged into the main spec before tests run). I think this would be cool, but realized that it add quite a few new challenges. For example, I would need to add new pattern-matching validators to the style spec, so I backed out and kept things simple.

I would love a review from @jseppi in addition to a review from the gl team, since this PR significantly impacts upstream work in api-styles.

samanpwbb:
Oh one more thing: How am I supposed to test the function in style-spec/bin ? Trying to run them directly didn't work :|

samanpwbb:
> Minor nitpick, also more of a question, any reason to use .`forEach` instead of `for..of`?

Here: https://github.com/mapbox/mapbox-gl-js/pull/8663/files#diff-abf4b5c9003a59059bbce3d73ab97182R65 – I need index to report a nice key for the error, and forEach is a little nicer than tracking an i variable. Since I was using forEach there I stuck with it. I wasn't able to discern from GL's codebase a preferred style for iterating over objects but I'd be happy to switch to `for..of` if that's more inline with the direction you all are headed in.

src/style-spec/bin/gl-style-validate
```diff
@@ -39,4 +46,5 @@ function help() {
     console.log('');
     console.log('options:');
     console.log('--json  output errors as json');
+    console.log('--mapbox-api-supported validate compatibility with Mapbox Styles API');
```
jseppi: nit: the `--json` help has two spaces between the flag and the description. do the same here for consistency?

```suggestion
    console.log('--mapbox-api-supported  validate compatibility with Mapbox Styles API');
```

src/style-spec/read_style.js
```diff
@@ -0,0 +1,15 @@
+import ParsingError from './error/parsing_error';
+import jsonlint from '@mapbox/jsonlint-lines-primitives';
+
+export default function readStyle(style) {
+    let s = style;
+    if (s instanceof String || typeof s === 'string' || s instanceof Buffer) {
+        try {
+            s = jsonlint.parse(s.toString());
```
jseppi: Could we return here instead of the new variable assignment?

```suggestion
            return jsonlint.parse(style.toString());
```

Then we could also remove the `let s = style;` and change the final return to `return style`

src/style-spec/read_style.js
```diff
@@ -0,0 +1,15 @@
...
+        } catch (e) {
+            throw new ParsingError(e);
+        }
+    }
+
+    return s;
```
jseppi: What's the case for returning the original input without attempting to parse? Could we remove the `if` statement and simplify this function to:

```js
export default function readStyle(style) {
  try {
    return jsonlint.parse(style.toString());
  } catch (e) {
    throw new ParsingError(e);
  }
}

File: src/style-spec/validate_mapbox_api_supported.js [hunk 6]
```diff
@@ -0,0 +1,172 @@
...
+const MAX_SOURCES_IN_STYLE = 15;
+
+function isValid(value: ?string, regex: RegExp): boolean {
+    if (!value || getType(value) !== 'string') return true;
+    return !!value.match(regex);
+}
+
+function getSourceCount(source: Object) {
+    if (source.url) {
+        return source.url.replace('mapbox://', '').split(',').length;
```
jseppi: Does the first instance of `'mapbox://'` need to be removed before `split`ting? I think the `.replace` call could be removed probably.

src/style-spec/validate_mapbox_api_supported.js
```diff
@@ -0,0 +1,172 @@
...
+    if (!isValid(style.sprite, spriteUrlPattern)) {
+        errors.push(new ValidationError('sprite', style.sprite, 'Styles must reference sprites hosted by Mapbox'));
+    }
+
+    /*
+     * "visibility" is optional. If present, valid examples:
+     * "private"
+     * "public"
+     */
+    const visibilityPattern = /^(public|private)/;
```
jseppi: I think this needs a `$` at the end to make sure strings like `'private-boop'` are not admitted:

```suggestion
    const visibilityPattern = /^(public|private)$/;
```

src/style-spec/validate_style.min.js
```diff
@@ -26,8 +26,7 @@ import validateLayoutProperty from './validate/validate_layout_property';
  *   var validate = require('mapbox-gl-style-spec/lib/validate_style.min');
  *   var errors = validate(style);
  */
-function validateStyleMin(style, styleSpec) {
-    styleSpec = styleSpec || latestStyleSpec;
+function validateStyleMin(style, styleSpec = latestStyleSpec) {
```
jseppi: In the preceding file ([link](https://github.com/mapbox/mapbox-gl-js/pull/8663/files#diff-d51d6c6dc5a36091136162e69a7323d4R34)), `options` (which allows setting `mapboxApiSupported`) gets passed to this function, but I don't see it in the signature of this function. Should it be here?

jseppi: Oh, maybe the change in `validate_style.js` was a holdover from a previous approach at combining the two validation modes?

samanpwbb: Oh yeah good catch – I need to remove that

src/style-spec/read_style.js
samanpwbb: I imagine it's just a performance optimization. It was in the code before this PR so I am inclined to leave it since validate _is_ run at runtime by GL JS.

kkaefer:
@samanpwbb could you please draft a note for our changelog?

samanpwbb:
Here:
- Add a new validation function that returns the same error responses as the Mapbox Styles API. Use as a CLI tool with gl-style-validate --api-supported-mode. Use in Javascript by importing validateMapboxApiSupported from the @mapbox/mapbox-gl-style-spec package.

If that’s too long, just use the first sentence

PR #8667, opened 2019-08-21
clarification of description for step expression

nextstopsun:
Style spec:
```
Returns the output value of the stop just less than the input, or the first input if the input is less than the first stop.
```

Should really be

Returns the output value of the stop just less than the input, or the first **_output_** if the input is less than the first stop.


mourner:
Good catch, thanks!

PR #8683, opened 2019-08-23
Backport extended build uploads to CDN

mourner:
CP of #8681.


mourner:
Ah damn, forgot to merge this one before tagging v1.3.0 — so this PR will have to wait until we do a patch release (which is likely given a [pending Earcut fix](https://github.com/mapbox/earcut/issues/119)).

Issue #8052, opened 2019-03-18
Default icon-image value if not present in a sprite


11sma:
## Motivation

I've searched in docs and in issues (quick search) and in google but I didn't find a possible solution. I think that simply didn't exists (sorry if yes...). so...
I'm trying to put in a map a serie of icon-image depending a property icon on tiles (or geojson). These icons are in sprite. I'm using: 

`"icon-image": "{icon}"`

It works fine. The problem (and the motivation) is if "icon" in properties take a value that is not in the sprite, obviously, shows nothing. Is there some way to put a default icon if mapbox cannot use the icon of sprite because it doen't exists?

## Design Alternatives

Nowadays I can create a new set of sprites faster I can to don't have this missmatch. But it's not automatic, nor always can do it so faster. 
Another measure is to use match expression, to create some similar to default, but the inconvenient is that I need to specify to the style all the possible values of icon and, in fact, apply the changes faster as the option above.

I think this is not the correct way...

### Mock-Up

I think it should be similar than match expression existent.

`'icon-image': ['{icon}',
{'default': 'default-icon-image'}
]`
(sorry if the acuracy of code doesn't match)

### Implementation

simply adding the "default" value if the value of a property is an array

peterqliu:
🤔 this would be useful, but trying to think of how we can fit this in our existent types without a breaking change. So far, we don't have any properties that could take multiple types ("string or array"). Expressions and categorical DDS do support defaults, but as you said, users must list out the whole sprite list.

- could add a `default-icon-image` prop
- at the stylesheet's root level, could add a prop specifying one global fallback sprite for all layers using any sprite property (`*-pattern` and `icon-image`)

pathmapper:
If I understood it right, another option is suggested in https://github.com/mapbox/mapbox-gl-js/issues/5261 (Add expression operator to determine style image availability).

ryanhamley:
## Default icon-image options

### A) `icon-default` symbol style spec property

```
    "layout": {
      "icon-default": "myDefault",
      "icon-image": "cat",
      "icon-size": 0.25
    }
```

**Pros:**

- easy for users to understand
- flexible enough to allow for different default icons depending on the layer

**Cons:**

- requires a new style spec property
- does not allow for data-driven or more complex, branching fallback checks

### B) root-level `default-sprite` or `default-icon` style spec property

**Pros:**

- users can quickly apply a single default icon to the entire map

**Cons:**

- requires a new style spec property
[...]

PR #8684, opened 2019-08-23
Image expression

ryanhamley:
## Launch Checklist

This PR implements a new `Image` expression type and `image` operator as discussed in https://github.com/mapbox/mapbox-gl-js/issues/8052# particularly https://github.com/mapbox/mapbox-gl-js/issues/8052#issuecomment-511622996

    - adds an `image` expression operator to allow for determining an image's availability before requesting it
 - [ ] post benchmark scores
    - will need a native port @mapbox/gl-native 

**Changelog**
<changelog>Add `image` expression operator which determines if a requested image is available in the map's style or not</changelog>

### DIFF
--- a/src/render/draw_background.js
+++ b/src/render/draw_background.js
@@ -25,7 +25,7 @@ function drawBackground(painter: Painter, sourceCache: SourceCache, layer: Backg
     const transform = painter.transform;
     const tileSize = transform.tileSize;
     const image = layer.paint.get('background-pattern');
-    if (painter.isPatternMissing(image)) return;
+    if (painter.isPatternMissing((image: any))) return;
 
     const pass = (!image && color.a === 1 && opacity === 1 && painter.opaquePassEnabledForLayer()) ? 'opaque' : 'translucent';
     if (painter.renderPass !== pass) return;
@@ -46,9 +46,8 @@ function drawBackground(painter: Painter, sourceCache: SourceCache, layer: Backg
     const crossfade = layer.getCrossfadeParameters();
     for (const tileID of tileIDs) {
         const matrix = painter.transform.calculatePosMatrix(tileID.toUnwrapped());
-
         const uniformValues = image ?
-            backgroundPatternUniformValues(matrix, opacity, painter, image, {tileID, tileSize}, crossfade) :
+            backgroundPatternUniformValues(matrix, opacity, painter, (image: any), {tileID, tileSize}, crossfade) :
             backgroundUniformValues(matrix, opacity, color);
 
         program.draw(context, gl.TRIANGLES, depthMode, stencilMode, colorMode, CullFaceMode.disabled,
--- a/src/render/draw_background.js
+++ b/src/render/draw_background.js
@@ -25,7 +25,7 @@ function drawBackground(painter: Painter, sourceCache: SourceCache, layer: Backg
     const transform = painter.transform;
     const tileSize = transform.tileSize;
     const image = layer.paint.get('background-pattern');
-    if (painter.isPatternMissing((image: any))) return;
+    if (painter.isPatternMissing(image)) return;
 
     const pass = (!image && color.a === 1 && opacity === 1 && painter.opaquePassEnabledForLayer()) ? 'opaque' : 'translucent';
     if (painter.renderPass !== pass) return;
--- a/src/render/draw_background.js
+++ b/src/render/draw_background.js
@@ -47,7 +47,7 @@ function drawBackground(painter: Painter, sourceCache: SourceCache, layer: Backg
     for (const tileID of tileIDs) {
         const matrix = painter.transform.calculatePosMatrix(tileID.toUnwrapped());
         const uniformValues = image ?
-            backgroundPatternUniformValues(matrix, opacity, painter, (image: any), {tileID, tileSize}, crossfade) :
+            backgroundPatternUniformValues(matrix, opacity, painter, image, {tileID, tileSize}, crossfade) :
             backgroundUniformValues(matrix, opacity, color);
 
         program.draw(context, gl.TRIANGLES, depthMode, stencilMode, colorMode, CullFaceMode.disabled,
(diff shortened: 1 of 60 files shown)


src/style-spec/expression/definitions/image.js
```diff
@@ -0,0 +1,54 @@
...
+
+        return null;
+    }
+
+    eachChild(fn: (Expression) => void) {
+        fn(this.input);
+    }
+
+    possibleOutputs() {
+        // Technically the combinatoric set of all children
```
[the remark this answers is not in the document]
ryanhamley: I'm not sure what `possibleOutputs` is for. It haven't needed it for this. This comment was copied over from elsewhere. I can update the comment.

src/render/image_atlas.js
```diff
@@ -140,4 +140,3 @@ export default class ImageAtlas {
 
 register('ImagePosition', ImagePosition);
 register('ImageAtlas', ImageAtlas);
-
```
asheemmamoowala: nit: no change needed

src/style-spec/expression/definitions/image.js
```diff
@@ -0,0 +1,52 @@
...
+        this.input = input;
+    }
+
+    static parse(args: $ReadOnlyArray<mixed>, context: ParsingContext): ?Expression {
+        if (args.length !== 2) {
+            return context.error(`Expected two arguments.`);
+        }
+
+        const name = context.parse(args[1], 1, StringType);
+        if (!name) return context.error(`No icon name provided.`);
```
asheemmamoowala: nit: Error should refer to `image` not `icon`

src/style-spec/reference/v8.json
```diff
@@ -2804,6 +2804,15 @@
           }
         }
       },
+      "image": {
+        "doc": "Returns `image` type for use in `icon-image` and `*-pattern` entries. If set, the `image` argument will check that the requested image exists in the style and will return either the resolved image name or `null`, depending on whether or not the image is currently in the style. This validation process is synchronous and requires the image to have been added to the style before requesting it in the `image` argument.",
```
asheemmamoowala:
```suggestion
        "doc": "Returns an `image` type for use in `icon-image` and `*-pattern` entries. This expression validates that a named image is available in the style, returning `null` if it is not found. This validation process is synchronous and requires the image to have been added to the style before requesting it in the `image` argument.",
```

src/render/draw_background.js
```diff
@@ -25,7 +25,7 @@ function drawBackground(painter: Painter, sourceCache: SourceCache, layer: Backg
     const transform = painter.transform;
     const tileSize = transform.tileSize;
     const image = layer.paint.get('background-pattern');
-    if (painter.isPatternMissing(image)) return;
+    if (painter.isPatternMissing((image: any))) return;
```
asheemmamoowala: Is Flow complaining because the expected type in `isPatternMissing` is `CrossFaded<string>` but `layer.paint.get` is now returning a `CrossFaded<Image>`. If so would it make sense to update `isPatternMissing` and undo the changes in `draw_line`, `draw_fill`, and `draw_fill_extrusion`

ryanhamley: `isPatternMissing` expects `CrossFaded<string>`. I can change that to `CrossFaded<Image>` and then in `isPatternMissing` call `image.to.toString()` and it works well and gets rid of the Flow error.

If I try that same approach in the other `draw_*` files though, I get a weird Flow error. 

```
        if (constantPattern && tile.imageAtlas) {
            const posTo = tile.imageAtlas.patternPositions[constantPattern.to.toString()];
            const posFrom = tile.imageAtlas.patternPositions[constantPattern.from.toString()];
            if (posTo && posFrom) programConfiguration.setConstantPatternPositions(posTo, posFrom);
        }
```

gets rid of the `Image is incompatible with string` error (patternPositions expects a string key) but Flow then complains that it `Cannot get tile.imageAtlas.patternPositions because property patternPositions is missing in null or undefined` which doesn't make sense since we're already checking that `tile.imageAtlas` isn't null or undefined.

ryanhamley: Actually, I'm wrong. I had forgotten to remove the cast to `any` in `draw_background`. If I do that and update `isPatternMissing` I get this error:

`Cannot call painter.isPatternMissing with image bound to image because Image [1]
is incompatible with Image [2] in property from.`

which doesn't make sense since `to` and `from` are just defined as the type passed to `CrossFaded`

https://github.com/mapbox/mapbox-gl-js/blob/3d6d2b440c89199dfc38f0a2b109588f1f264ac8/src/style/properties.js#L28-L31

src/render/image_atlas.js
ryanhamley: This was automatically added by my editor. It's pretty standard to end files with a new line isn't it? Most of our other files have a trailing new line.

src/render/draw_background.js
ryanhamley: @ahk helped me fix the casts to `any` in `draw_{fill/fill-extrusion/line}` but we haven't come up with a way around this cast yet. We're still trying.

ryanhamley: Turns out that type `Image` accidentally shadows the Javascript constructor `Image` leading to a very difficult to debug Flow error. To avoid this, I'm going to change the `Image` type to `ResolvedImage`

ryanhamley: All casts to `any` have been removed. The Flow typing should be solid now

ryanhamley:
>Question, do you plan to add support for default image that is not in a sprite sheet? Or, would it be handled via onStyleImageMissing callback?

I'm adding an additional step in a follow-up PR to determine the first requested image in an expression (reading from left to right and assuming that the first image requested is the "most desirable" option) which evaluates to `null` and returning that image's name so that `styleimagemissing` will always work as expected.

>Another question is that, what happens when developer removes image from a sprite sheet while worker is parsing tile? It looks like availableImages and images that are in sprite sheet could be out-of-sync and image operator would not fall-back to default value (e.g., when used in coalesce).

I think this is analogous to what happens if you asynchronously load an image with `addImage`. Before this PR, asynchronously loading an image will not cause it to appear in a tile that had requested the image previous to it being loaded. This PR does not change that behavior because it would require reloading tiles every time an image is added or removed, which likely would be distracting and probably a performance hit. I think the best way to deal with both of these situations is still an open question.

Issue #8691, opened 2019-08-26
load event not fired after style change with setStyle


stepankuzmin:
**mapbox-gl-js version**:

**browser**:

### Steps to Trigger Behavior

 1. create map with `load` event handler
 2. change style with setState

### Link to Demonstration

https://jsbin.com/

stepankuzmin:
I've created a PR for this https://github.com/mapbox/mapbox-gl-js/pull/8692

mourner:
I think it's working as designed — from the docs:

> Fired immediately after all necessary resources have been downloaded and the **first** visually complete rendering of the map has occurred.

Maybe you're looking for the `styledata` event instead?

stepankuzmin:
Hey @mourner! Thanks for the reply!

The problem with `styledata` is that after changing map style with `setStyle ` there is no way (as far as I know), to catch map in loaded state.

Calling `map.isStyleLoaded()` in `styledata` event handler always resolves to `false`.

Consider example: https://jsfiddle.net/StepanKuzmin/goxjspc4t/

mourner:
@stepankuzmin should we fix `isStyleLoaded` instead of changing the semantics of `load` then?

stepankuzmin:
Well, yes. If it is intended to fire `load` event only once during map lifecycle, then it would be good to detect if map style is loaded after changing it using `styledata` event handler.

gmaclennan:
This would be really good to fix. I have been running into variants of this bug for the past two years of developing maps with mapbox-gl-js. I can confirm the current issue is that `map.isStyleLoaded()` is very unreliable, and if it is `false` then it is not clear what event you can listen to for when it is `true` again. E.g. if I listen to `styledata` then sometimes the last event `isStyleLoaded()` is `false`, and if I listen to `sourcedata` then `isStyleLoaded()` seems to randomly switch between `true` and `false`

DannyDelott:
I've been able to attach a handler on the `style.load` event instead of `load` to know when it is safe to update the map after calling `map.setStyle()`.

It's been working fine for us since early 2018, but it is an undocumented event type.

mourner:
Related issue we need to cover when fixing this: #4849

leigeber:
[...]

PR #8692, opened 2019-08-26
fix: reset map loaded state to false on setStyle #8691

stepankuzmin:
Map loaded state should be reseted to false after style changed with `setStyle`, so `load` event can be emitted.


mourner:
Closing as this would need a different fix per discussion in https://github.com/mapbox/mapbox-gl-js/issues/8691#issuecomment-525251837

Issue #6405, opened 2018-03-27
`event` passed from map.on('moveend' missing `originalEvent` on some interactions


hryanjones:
**Context**: I'm looking for a way to tell whether user interaction caused the map to move. The use case is that we use `.fitBounds` to set the initial state and when the user changes the bounding box (via scroll zoom, double click zoom, drag to pan, the navigation controls, keyboard commands, etc.) we want to persist the bounds, but we don't want to do this when it's a `fitBounds`.

Is there a way to tell that it was user initiated? The only time I've seen where I can distinguish is when I also see an `.originalEvent` property in the event in addition to `.target` which happens on some user interactions, but not on all of them.

**mapbox-gl-js version**: (using version that was in the examples on mapboxGL website)

### Steps to Trigger Behavior

1. Using HTML below with console open
2. notice that `.fitBounds` triggers the `event` and it doesn't have an `.originalEvent` property
3. notice that the event logged from these interactions have an `.originalEvent`:
  - drag
  - double click
  - if you use the keyboard +/- buttons
4. notice that `.originalEvent` is missing if you do the following:
  - click on the navigation +/- zoom buttons
  - if you scroll to change zoom

```html
<script src='https://api.tiles.mapbox.com/mapbox-gl-js/v0.44.1/mapbox-gl.js'></script>
<link href='https://api.tiles.mapbox.com/mapbox-gl-js/v0.44.1/mapbox-gl.css' rel='stylesheet' />
<style>
    body { margin:0; padding:0; }
    #map { position:absolute; top:0; bottom:0; width:100%; }
</style>

<div id='map'></div>
<script>
mapboxgl.accessToken = ''; // Put your own mapbox token here

var map = new mapboxgl.Map({container: 'map', style: 'mapbox://styles/mapbox/streets-v9'});
map.fitBounds([[-133.4261895047858, 27.268363255653668], [-89.21205651435507, 46.38482102526038]]);
map.addControl(new mapboxgl.NavigationControl());

map.on('moveend', function(event) {
    console.log(event);
});
</script>
```

soldt:
A solution would be to add eventData to the fitBounds method.

`eventData(Object)Additional properties to be added to event objects of events triggered by this method.`

Example:
```
map.fitBounds(lngLatBounds, {}, {breakonfitbounds: true});
```

Return foorbar
```
map.on('moveend', function (e) {
  if (e.breakonfitbounds) {
    return;
  }
  // Do something
});
```

[...]

PR #8693, opened 2019-08-26
Add `originalEvent` property to NavigationControl events

stepankuzmin:
This PR adds `originalEvent` property to events initiated using `NavigationControl` (zoom in, zoom out, reset bearing). Relates to https://github.com/mapbox/mapbox-gl-js/issues/6405

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -52,18 +52,18 @@ class NavigationControl {
             bindAll([
                 '_updateZoomButtons'
             ], this);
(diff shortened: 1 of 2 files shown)


stepankuzmin:
> @stepankuzmin Thanks for this fix. Looks good , but could you also add a note in the docs to signal that the `MouseEvent` can be a click when the zoom is triggered by a UI event.

@asheemmamoowala I've added a note for this

mourner:
Also, could you please rebase on `master`? This might fail the build once merged because of stricter linting rules about curly braces.

stepankuzmin:
@mourner done!

Issue #6027, opened 2018-01-19
raster-dem sub-meter hillshade?


halset:
When zooming far in, it looks like both the `mapbox.terrain-rgb` and my own bathymetry based terrain rgb tiles have only meter vertical resolution. However, at least my bathymetry data set has higher resolution.

Here is an example with the bathymetry tiles. 
<img width="497" alt="screen shot 2018-01-19 at 23 29 30" src="https://user-images.githubusercontent.com/445229/35174706-a2cf5cd4-fd70-11e7-819a-0425415832b3.png">

Here is an example with `mapbox.terrain-rgb`.
<img width="606" alt="screen shot 2018-01-19 at 23 31 40" src="https://user-images.githubusercontent.com/445229/35174813-0337f14e-fd71-11e7-9537-4798ea187efa.png">

Do you have an example where `mapbox.terrain-rgb` shows shadows from vertical sub-meter resolution?

mollymerp:
the terrain-rgb _encoding_ is precise to 0.1 m but the _underlying data_ vary in their resolution. 
what zoom level are these screenshots from – I see this kind of jagginess and pixelation with overzoomed tiles (the maxzoom of the terrain-rgb tileset is 15).

halset:
When looking at the raw dem tile for 14/8431/4555, there seem to be smooth transitions, but when looking at the rendered map in mapbox-gl-js (and -native), the rendering seem to be rounded to whole meters. This screen shot is from terrain-rgb 14/8431/4555 rendered as hillshade.
<img width="881" alt="screen shot 2018-01-27 at 13 52 12" src="https://user-images.githubusercontent.com/445229/35472130-8c912a34-0369-11e8-8b12-62e0efe45cd7.png">

tomass:
@halset is zoom 15 built-in limit in mapbox-gl-js?
I'm also getting visually bad results with higher zooms (with terrain-rgb tiles created by rio rgbify with --interval 0.01 from a very detailed source GeoTIFF).

At zoom 15.42 everything is fine:
![paveikslas](https://user-images.githubusercontent.com/969513/35675965-b057a1b8-0752-11e8-9950-b0bfa1064f92.png)

If I zoom a little bit to 15.6 I'm starting to get artificial visual lines:
![paveikslas](https://user-images.githubusercontent.com/969513/35676015-e2d631ea-0752-11e8-8309-09b4187a940e.png)

And if I zoom to 18, I get this:
[...]

PR #8694, opened 2019-08-27
DEMData: save one CPU copy and do decode on GPU

astojilj:
Save one CPU copy and avoid decoding DEM data on CPU (do it on GPU).

Pad DEMData, for backfill, in context.getImageData. This avoids the need to pad it in DEMData constructor.

DEM decode to height is a single dot operation - no need to do it on CPU since we already have another decode (from internal format) in src/shaders/hillshade_prepare.fragment.glsl.

Time spent, averaged in milliseconds, in `getImageData` and `new DEMData` is measured using [this patch](https://gist.github.com/astojilj/96915e2b9931ad5384335847b6a4e11e).

When opening debug/hillshade.html, page loads 7 256x256 tiles, with total 368ms (~52ms per tile) spent in new DEMData. This is reduced to 1 ms per tile with the patch.

```
                    master branch       |          with this patch
----------------------------------------|-------------------------------           
           new DEMData  | getImageData  |  new DEMData  | getImageData     
------------------------|---------------|---------------|---------------           
iPhoneX       45 ms     |     12 ms     |       1 ms    |     12 ms 

```
            

cc @mapbox/gl-native

### DIFF
--- a/src/data/dem_data.js
+++ b/src/data/dem_data.js
@@ -16,56 +16,53 @@ import { register } from '../util/web_worker_transfer';
 
 export default class DEMData {
     uid: string;
-    data: Int32Array;
+    data: Uint32Array;
     stride: number;
     dim: number;
+    encoding: "mapbox" | "terrarium";
 
+    // RGBAImage data has uniform 1px padding on all sides: square tile edge size defines stride
+    // and dim is calculated as stride - 2.
     constructor(uid: string, data: RGBAImage, encoding: "mapbox" | "terrarium") {
         this.uid = uid;
         if (data.height !== data.width) throw new RangeError('DEM tiles must be square');
         if (encoding && encoding !== "mapbox" && encoding !== "terrarium") return warnOnce(
             `"${encoding}" is not a valid encoding type. Valid types include "mapbox" and "terrarium".`
         );
-        const dim = this.dim = data.height;
-        this.stride = this.dim + 2;
-        this.data = new Int32Array(this.stride * this.stride);
-
-        const pixels = data.data;
-        const unpack = encoding === "terrarium" ? this._unpackTerrarium : this._unpackMapbox;
-        for (let y = 0; y < dim; y++) {
-            for (let x = 0; x < dim; x++) {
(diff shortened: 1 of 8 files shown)


andrewharvey:
Oh hey it looks like this partially fixes #6027 :tada: This is awesome. If you look closely you can still see some artefacts described in #6027 with this PR, but they don't appear for a few extra zooms further than currently in master, and without more testing I can't say if that's due to a limitation in the data or in the codebase... I can help try to answer that.

Two screen shots below, exact same data and style:

## master
Screenshot from 2019-08-27 23-53-07

## this PR
Screenshot from 2019-08-27 23-52-41

andrewharvey:
This performance improvement is awesome.

Though I wanted to point some tangential analysis I did a while back at https://github.com/mapbox/rio-rgbify/issues/3#issuecomment-381330102. I know this is a bit off topic and not something that can be addressed on the client side, but might be worth investigating.

For a large map that takes 25 256x256 terrain tiles for the map, that's 3.3MB and took about 4.3 seconds just to download those tiles, this would be worse over a bad mobile connection.

Asking for terrain-rgb `webp` tiles, Mapbox still only serves `png`s anyway.

    https://a.tiles.mapbox.com/v4/mapbox.terrain-rgb/9/85/197@2x.webp
    > identify 197@2x.webp
    PNG 512x512 512x512+0+0 8-bit sRGB

I tried a few compression options to see what affect that would have:

    > for f in tiles/*; do b=`basename $f`; pngcrush -brute $f pngcrush/$b; done
    > for f in tiles/*; do b=`basename $f`; cwebp -lossless -q 100 -m 6 $f -o cwebp/$b; done

method | total size
--|--
original | 3.3M
pngcrush | 2.4M ~30% improvement
cwebp | 1.8M ~50% improvement

Keep in mind both of these compressions are lossless, so there's big gains to be made by spending more compute on the API side to reduce the size sent over the wire.

astojilj:
@andrewharvey 

Thanks for checking #6027 and letting me know.

PR #8697, opened 2019-08-27
Set eslint object-curly-spacing to `never`

asheemmamoowala:
I've seen PRs with inconsistent rules applied to spacing inside curly brace objects. This PR sets the strict lint to never allow leading and trailing spaces inside curly braces.

This changes a large number of files, but enforces consistency so that it does not need to be `nit`ted in code reviews.

cc @ansis


mourner:
Needs rebase after #8694

Issue #8575, opened 2019-07-29
newline character not respected if text-max-width is set to zero


samanpwbb:
**mapbox-gl-js version**: 1.1.0

**browser**: _not browser specific_

### Steps to Trigger Behavior

 1. Use a newline character `\n` in a text field value
 2. Set `text-max-width` to zero

### Link to Demonstration

https://jsbin.com/zawiyegehe/1/edit?html,output

### Expected Behavior

Text wraps to second line after newline character:
<img width="249" alt="Screen Shot 2019-07-29 at 1 23 05 PM" src="https://user-images.githubusercontent.com/108094/62079661-09784100-b204-11e9-9ae0-e675f9d9e263.png">

### Actual Behavior

Text does not wrap:
<img width="273" alt="Screen Shot 2019-07-29 at 1 23 08 PM" src="https://user-images.githubusercontent.com/108094/62079671-0da45e80-b204-11e9-9585-a773b1305400.png">

mourner:
Good catch! Looks like we skip determining line breaks entirely if `max-width` is not set (regressed in #3725): 

https://github.com/mapbox/mapbox-gl-js/blob/4ffd0e2c6cbdf4f2d96b89f79a816012eba66329/src/symbol/shaping.js#L353-L354

ryanhamley:
@malwoodsantoro and I have been working on this bug together and I wanted to capture what we've discovered so far.

The proximate culprit is the check noted above

```
if (!maxWidth)
    return [];
```

Removing this check however runs into a related issue caused by this bit of code.

https://github.com/mapbox/mapbox-gl-js/blob/master/src/symbol/symbol_layout.js#L186-L188

If `symbol-placement` is set to `line` or `line-center`, `text-max-width` is set to zero. This makes sense since text that follows a line shouldn't have breaks, but removing the `if (!maxWidth)` check causes the line to break on every whitespace character - exactly the opposite of what we want. This is confirmed by numerous failing line rendering tests if the `if` statement is removed.

A perhaps more correct way to write the `if` statement would be something like this:
```
// disable line breaking entirely on text-along-a-line
// point placed symbols will always be shaped
if (symbolPlacement !== 'point') {
    return [];
}
```

The only problem with this is that `shaping.js` does not have access to the `layout` instance or the `symbol-placement` property so we'll have to pass that in. Mal and I will work on submitting a PR in our meeting next week.

cc @asheemmamoowala

PR #3725, opened 2016-11-30
Line break refactoring in preparation for bidirectional text support.

ChrisLoer:
Separates line-breaking logic from glyph placement. Introduces a few small behavior changes:
* Adds U+2027 "Interpunct" to ideographic breaking character set
* Trims whitespace from beginning and end of line before calculating width for alignment purposes.
* Fixes a crash on labels that generated lines with a single whitespace glyph.

shaping.test.js updated to reflect invisible behavior changes (e.g. invisible glyphs no longer added to shaping).

cc @lucaswoj @jfirebaugh @mollymerp @1ec5

PR #8706, opened 2019-08-29
Only shape point placement symbols

malwoodsantoro:
## Launch Checklist

Fixes https://github.com/mapbox/mapbox-gl-js/issues/8575

    - This PR fixes a bug in which newline characters were being ignored if `text-max-width` was set to zero. This was caused by a faulty check in the shaping code which equated "no max width" with "no shaping should occur". This PR fixes the issue and corrects the check by explicitly checking for `line` and `line-center` symbol placement and not shaping those symbols (e.g. only shape text when the symbol is placed as a point).
 - [x] post benchmark scores
    - this will need a port to native cc: @mapbox/gl-native 

<changelog>Fix bug in which explicit newline characters were ignored when text-max-width was set to zero</changelog>


ryanhamley:
This is working really well now. The only failing test is a render test `text-max-width/unlimited` which I believe to be an incorrect test. [The test](https://github.com/mapbox/mapbox-gl-js/tree/master/test/integration/render-tests/text-max-width/unlimited) uses `text-max-width: 0` and equates it with "unlimited" width is wrong. My recommendation is to convert this to an explicit test of `text-max-width: 0` with the expectation that there are line breaks on every possible whitespace character.

ryanhamley:
https://github.com/mapbox/mapbox-gl-js/pull/3725 is the PR that introduced the `if (!maxWidth)` check that is being changed here. I don't really understand why Chris went this particular direction. Making this change seems to be more consistent with the expected behavior and removes a confusing semantic equivalence between 0 max width and effectively infinite max width. I think this change is ok, but I just wanted to note the original PR in case.

asheemmamoowala:
> The only failing test is a render test text-max-width/unlimited which I believe to be an incorrect test.
> Making this change seems to be more consistent with the expected behavior and removes a confusing semantic equivalence between 0 max width and effectively infinite max width. I think this change is ok

@malwoodsantoro @ryanhamley It is unclear how this change affects existing styles that may be depending on this behavior. Maybe someone from @mapbox/map-design-team can comment on whether this change will require style changes for common use cases.

ryanhamley:
@mapbox/map-design-team Does anyone have feedback on this change?

ryanhamley:
Benchmarks look good

Screen Shot 2019-09-17 at 11 48 31 AM
Screen Shot 2019-09-17 at 11 48 40 AM
Screen Shot 2019-09-17 at 11 49 18 AM
Screen Shot 2019-09-17 at 11 49 26 AM

ryanhamley:
*Visual Changes*

This style with `text-max-width: 0` and an explicit newline character in the text:
Screen Shot 2019-09-17 at 11 54 57 AM

Produces this output on `master`. Note that the newline is ignored.
Screen Shot 2019-09-17 at 11 54 23 AM

With this PR, this is the output. Since the text max width is 0, we break on every whitespace character.
Screen Shot 2019-09-17 at 11 53 34 AM

Issue #8734, opened 2019-09-06
Tooltip for All Icon-only Control Buttons


BAByrne:
## Motivation
Zoom in/out and bearing buttons have descriptions that display on hover. Icon-only control buttons (e.g. Geolocation) can be ambiguous to casual users of interactive maps. 

## Design Alternatives
Add a custom tooltip or title attribute to all icon-only core control buttons. 

## Design
A title attribute: solves the problem; adds to accessibility; is supported by all modern browsers; requires minimal changes to the codebase

### Mock-Up
N/A

### Concepts
The geolocator's tooltip should be descriptive. Maybe something like, "Find my location".

### Implementation
Should be implemented the same way the other button titles are implemented.

PR #8735, opened 2019-09-06
Fixes #8731 by adding tooltip to the geolocation control button.

### LINKED ISSUE #8731: style the shared outlines between rendered features in a layer
**mapbox-gl-js version**:

### Question

My req is to use fill-outline-color for  rendered features in a layer and make transparent the outline's which are shared between features ( I looked into documentation of layer type fill and line), but couldn't find a way to achieve this.

an example to be make it clear: My layer filtering California and Nevada should have an outline color but the shared border between two states should be invisible.

BAByrne:
Fixes issue #8734 (wrong issue # in commit msg) - Added a title attribute to geolocation control button. 

### DIFF
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -270,8 +270,8 @@ class GeolocateControl extends Evented {
             `${className}-icon ${className}-geolocate`,
             this._container);
         this._geolocateButton.type = 'button';
-        this._geolocateButton.title = 'Show my location';
-        this._geolocateButton.setAttribute('aria-label', 'Geolocate');
+        this._geolocateButton.title = 'Find my location';
+        this._geolocateButton.setAttribute('aria-label', 'Find my location');
 
         if (this.options.trackUserLocation) {
             this._geolocateButton.setAttribute('aria-pressed', 'false');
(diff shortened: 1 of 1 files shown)


ryanhamley:
Thanks for the PR @BAByrne! It looks good. I'm just going to tag in a teammate for their thoughts on semantics.

@chloekraw This PR seems like a good idea but do you have any thoughts on what would be an appropriate tooltip text? `Show my location` is suggested here. `Find my location` might be a more accurate representation of the geolocation function. Also, in the navigation control, the buttons' `title` and `aria-label` are the same thing. I think we should probably follow that pattern here. The `aria-label` "Geolocate" seems a bit uninformative though so we should probably match it to whatever we decide on for a title.

andrewharvey:
I added the "Geolocate" label back in #3492, I agree that "Geolocate", while it might be clear for web developers, it's probably not the clearest for the average user.

I think "Show my location" or "Find my location" are both fine. Sometimes the control will add a blue dot, sometimes it won't (depending on options) but I still think either are fine for both scenarios.

I agree we should keep title and aria-label in sync.

chloekraw:
@ryanhamley's suggestion of `Find my location` sounds good to me 👍

Issue #8730, opened 2019-09-05
Always use GL_LINEAR interpolation for icon rendering


kkaefer:
Our goal is to render icons as crisp and sharp as possible. Since our icons are raster based, the best case is doing a 1:1 blit where we transfer the pixels directly, without any kind of interpolation, from the icon atlas texture. To that extent, we are disabling texture interpolation (by using `GL_NEAREST`) in situations where the map is not in the process of being rotated or zoomed, it's not tilted, and only for icons that are not rotated, transformed, or resized in any way. If all of those conditions are met, our map looks like this:

![interpolation-nearest](https://user-images.githubusercontent.com/52399/64354845-79969580-d000-11e9-920d-f83d2ecd5dc0.png)

However, when one of these isn't met, we resort to linear interpolation, which makes them look like this:

![interpolation-linear](https://user-images.githubusercontent.com/52399/64354895-8f0bbf80-d000-11e9-84d1-39329100f017.png)

Note how some icons, like the Metro signs, are a bit blurry. To make this easier to see, here's a GIF that swaps both back and forth:

![interpolation](https://user-images.githubusercontent.com/52399/64355133-f3c71a00-d000-11e9-9d67-dd73d139d911.gif)

However, there are a lot of caveats with our current approach:

- We've hit lots of rounding errors where nearest-neighbor texture lookups produced weird results
- We can't render pixel-grid aligned icons with interpolated icons in the same draw call. This e.g. happens when using data-driven styling for `icon-scale` or `icon-rotate` ([a user hit this problem](https://github.com/mapbox/mapbox-gl-native/issues/15537)), or when the icon layer is tilted
- We can't render pixel-grid aligned icons when they're rotated by right angles (90°, 180° 270°)
- The switch between linearly-interpolated and nearest-neighbor texture lookups when the user starts or stops rotating or zooming is quite jarring:
    ![images](https://user-images.githubusercontent.com/52399/64355967-6389d480-d002-11e9-86ea-731c8a5410ce.gif)

I think we should remove our use of `GL_NEAREST` and always use linear interpolation. To keep icons crips, we could ensure that icons that can be rendered crisply use texture coordinates that are aligned to the texture's pixel grid. This would allow us to mix pixel-grid aligned and scaled icons within the same draw call, as well as "fade" between pixel-grid alignment and linearly interpolated icons when rotating or zooming.

[...]

PR #8738, opened 2019-09-08
Prototype pixel snapping icons with GL_LINEAR

kkaefer:
Implements https://github.com/mapbox/mapbox-gl-js/issues/8730.

We're currently switching between pixel-grid aligned icons in a binary way. In a follow-up patch, we could implement a gradual switch by making `u_crisp` a float, and scaling it between 0 and 1 whenever the user interacts with the map.

* [ ] More visual tests
* [ ] Verify that it works correctly at other pixel ratios. I've observed that neighboring icons from the image atlas bleed into the icon at the sides and we may potentially have to scale the 1 px border with the pixel ratio (i.e. 2 pixels) to account for the pixel grid alignment.
* [ ] Verify on a large number of devices. I already tested a few different ones and they all work correctly. However, since this relies on floating point precision and rounding, there's a chance that some devices may behave erratically. I tried to take this into account when designing the calculation by using the projected center point, before the extrusion vectors are applied to get identical results for all four vertices of the icon.
* [ ] Potentially rework the way we use the transformation matrices to reduce the number of calculations we have to do; we currently have 3 4-dimensional projection matrices in the shader and I'm sure we can reduce that a bit. The code currently uses the canvas size to "unproject" from GL space (-1...1) to pixel space so that we can obtain the fractional pixel value to shift the texture coordinates.


src/shaders/symbol_icon.vertex.glsl
```diff
@@ -29,6 +32,13 @@ varying float v_fade_opacity;
 
 #pragma mapbox: define lowp float opacity
 
+vec2 absfract(vec2 value) {
+    vec2 result = fract(value);
+    if (result.x > 0.5) result.x -= 1.0;
+    if (result.y > 0.5) result.y -= 1.0;
+    return result;
```
astojilj: Shorter way to write it `return fract(value) - vec2(greaterThan(color, vec2(0.5)));`. Shouldn't matter, though.

src/shaders/symbol_icon.vertex.glsl
```diff
@@ -84,7 +94,15 @@ void main() {
     vec4 projected_pos = u_label_plane_matrix * vec4(a_projected_pos.xy, 0.0, 1.0);
     gl_Position = u_coord_matrix * vec4(projected_pos.xy / projected_pos.w + rotation_matrix * (a_offset / 32.0 * fontScale), 0.0, 1.0);
 
-    v_tex = a_tex / u_texsize;
+    v_tex = a_tex;
+    if (u_crisp && (segment_angle + symbol_rotation == 0.0)) {
+        // move texture loopup to snap to the pixel grid
```
astojilj: loopup/lookup

src/shaders/symbol_icon.vertex.glsl
```diff
@@ -84,7 +94,15 @@ void main() {
...
+        vec4 icon_pos = u_coord_matrix * vec4(projected_pos.xy / projected_pos.w, 0.0, 1.0);
+        v_tex += absfract((icon_pos.xy + 1.0) * u_canvas_size * vec2(1.0, -1.0) / 2.0);
+        v_tex += fract(a_offset / 32.0 * fontScale * u_device_pixel_ratio);
```
astojilj: Is there a need to multiply this by u_coord_matrix before fract, as it is in gl_Position calculation?

src/shaders/symbol_icon.vertex.glsl
kkaefer: I can try, but many OpenGL implementations I tested are _very_ picky about subtle changes like this and perform rounding differently.

kkaefer:
Related issue: https://github.com/mapbox/mapbox-gl-js/issues/8110 (and https://github.com/mapbox/mapbox-gl-js/pull/8118)

hrach:
Is this going to be fixed in release-c? Really looking for seing this fixed :) Thanks.

ryanhamley:
Closing this as stale

Issue #8722, opened 2019-09-03
Symbols with `icon-text-fit` and `text-offset`


tristen:
### Expected Behavior

I would expect collision rules to be governed by `text-offset` when a symbol layer is using `icon-text-fit`.

### Actual Behavior

Here's an example of a layer using `text-offset` vs a layer using `text-offset` and `icon-text-fit`. I would expect the text-offset value to continue to not force a collision with the icon below:

![text-offset-with-icon-fit](https://user-images.githubusercontent.com/61150/64210047-37d6e500-ce70-11e9-912a-867d008645cf.gif)

---

**mapbox-gl-js version**: v1.2.0
**browser**: Chrome `Version 76.0.3809.132 (Official Build) (64-bit)`

### Link to Demonstration

https://api.mapbox.com/styles/v1/tristen/cjzu99z8h00qp1co0srxwywwn.html?fresh=true&title=true&access_token={TOKEN}#7.9/43.797847/-79.658695

chloekraw:
We discovered that we weren't manipulating the collision boxes correctly when `icon-text-fit` was used with `text-variable-anchor` and `text-writing-mode` when working on the fix for https://github.com/mapbox/mapbox-gl-js/issues/8583. 

It looks like we didn't add a render test for `text-offset` (https://github.com/mapbox/mapbox-gl-js/pull/8625/files); thus, I'm not sure that we can confirm whether https://github.com/mapbox/mapbox-gl-native/pull/15367 fixes this issue too. (@ansis is working on the JS port.)

@alexshalamov could you add two render tests: one for using `icon-text-fit` with `text-offset` and another one for `icon-text-fit` with `text-variable-anchor` and `text-radial-offset` ?

lukasmartinelli:
@chloekraw What's the ETA for this?

asheemmamoowala:
@lukasmartinelli There is a potential fix for this already in GL-Native, and the JS port is in progress. Confirmation is needed that they correctly address this issue as well.

Both those should be available in the upcoming release.

alexshalamov:
@chloekraw added new render tests in https://github.com/mapbox/mapbox-gl-js/pull/8747

kkaefer:
This is fixed by https://github.com/mapbox/mapbox-gl-js/pull/8741

PR #8741, opened 2019-09-09
Fix icon text fit

kkaefer:
We're adding a 1 pixel border around all icons when adding them to the image atlas texture to avoid having neighboring icons bleed into the current image when using linear texture interpolation. When stretching icons to fit the text's dimensions with `icon-text-fit`, however, we didn't account for this correctly: We didn't add the border in this case, resulting in stretched icons that didn't quite cover the whole text. When the difference between the icon's original size and the stretched size was small, this wasn't really noticeable, especially in cases where we were using nearest neighbor texture interpolation instead of linear interpolation. When accounting for the 1 pixel border, have to to take into account that there's no longer a 1:1 mapping between vertex units and pixel units, and expand the vertices adequately.

Here's an example:

| Before | After |
| --- | --- |
| actual | actual |

This example uses a really small icon icon-text-fit so that the discrepancy between the true size and the scaled size is big, which leads to a large error.

Native patch is forthcoming.


kkaefer:
This PR changes the way icons are aligned with text when the `*-anchor` doesn't match. Here are all combinations _before_ the PR:



and with this PR applied:



As you can see, the previous code essentially ignored the `icon-anchor` value when at least one stretch direction was set (i.e. `icon-text-fit != none`), while the new PR respects the icon's position in the direction that wasn't stretched.

This makes me believe that we should  better define how `icon-text-fit` is supposed to work:
- Should `icon-text-fit` only affect the _sizing_ of the icon, or should it also affect the positioning? At the moment, it's a hybrid of both.
- Does this also apply to icons/text that is moved relative to the anchor point with `icon/text-offset`?
- Are there valid use cases for fitting an icon to the text size, but not also positioning it with the text?
- When applying a `icon-text-fit-padding`, should we respect the padding for the dimension that is not stretched in case of `width`/`height`?

/cc @mapbox/design @mapbox/studio

tristen:
> Are there valid use cases for fitting an icon to the text size, but not also positioning it with the text?

I don't think there are valid use cases for that.

---

@kkaefer My 2c would be this: When `icon-text-fit` is set, positioning properties by icon properties should no longer work: it should now be controlled by text positioning.

asheemmamoowala:
@kkaefer Does this also address the collision issue described in https://github.com/mapbox/mapbox-gl-js/issues/8722 ?

cc @zmiao

kkaefer:
@asheemmamoowala it didn't initially, but I changed the behavior so that the text-fitting happens earlier in the process and is now respected in the collision boxes it generates.

@tristen I made an example page that allows playing with the various values and how they're affecting positioning:

- [**Unfixed**/current version](https://mapbox.s3.amazonaws.com/kkaefer/symbols/unfixed-icon-text-fit.html)
- [**Fixed**/new version](https://mapbox.s3.amazonaws.com/kkaefer/symbols/fixed-icon-text-fit.html)

I increased the size so that it's easier to see. In Firefox and Chrome, antialiasing for the resized canvas is disabled (Safari doesn't have support for that).

A few other things that are fixed:
- The `icon-offset` is now applied _after_ fitting the icon to the text.
- The icon collision box is now correctly reflecting the actual icon position, after fitting + translation, and padding

There are also some changes that I'm not sure about:
- When `icon-text-fit` is `height`, we're not applying the left and right padding; only the bottom and top padding. Same applies to `width` and top/bottom padding. Should this behavior be reverted? I.e. should we always apply the `icon-text-fit-padding` in both axes regardless of whether the icon is fitted on that axis?
- `icon-anchor` is now without function if `icon-text-fit` is set to `width`/`height`/`both` and the icon is always _centered_ on the text. We could reuse `icon-anchor` in this case to describe where the icon should be anchored to _within_ the text frame. However, it also seems a bit weird to reuse the same name for this slightly different feature, and maybe use `icon-text-fit-anchor` for that instead if we decide to implement it.

Using `icon-size` with non-center `text-align` as well as a `icon-text-fit`ting produces a kind of parallax effect and the icon isn't centered anymore on the text. This is unchanged compared to the existing state. There's no easy way to fix this because the scale anchor would have to be relative to the text's position, which is already relative to the object's anchor position. Should we disable `icon-size` when `icon-text-fit` is set?

tristen:
Oh awesome @kkaefer "Fixed/new version" is 👨‍🍳 💋

asheemmamoowala:
> When `icon-text-fit` is `height`, we're not applying the left and right padding; only the bottom and top padding. Same applies to `width` and top/bottom padding. Should this behavior be reverted? I.e. should we always apply the `icon-text-fit-padding` in both axes regardless of whether the icon is fitted on that axis?

This seems like reasonable behavior to me, and provides decent outcomes, but I wonder if there is enough of a use-case to add support and documentation for this nuance. `icon-text-fit` is not data-driven yet, so these style parameters are fixed at design time. It would make more sense to add this if/when `icon-text-fit` is data driven. 

> `icon-anchor` is now without function if `icon-text-fit` is set to `width`/`height`/`both` and the icon is always _centered_ on the text. We could reuse `icon-anchor` in this case to describe where the icon should be anchored to _within_ the text frame. However, it also seems a bit weird to reuse the same name for this slightly different feature, and maybe use `icon-text-fit-anchor` for that instead if we decide to implement it.

> Using `icon-size` with non-center `text-align` as well as a `icon-text-fit`ting produces a kind of parallax effect and the icon isn't centered anymore on the text. This is unchanged compared to the existing state. There's no easy way to fix this because the scale anchor would have to be relative to the text's position, which is already relative to the object's anchor position. Should we disable `icon-size` when `icon-text-fit` is set? 

👍 to disabling `icon-size` and `icon-anchor` when using `icon-text-fit`.

src/symbol/quads.js
```diff
@@ -58,43 +58,50 @@ export function getIconQuads(anchor: Anchor,
...
-        tr = new Point(textLeft + offsetX + padR + width, textTop + offsetY - padT);
-        br = new Point(textLeft + offsetX + padR + width, textTop + offsetY + padB + height);
-        bl = new Point(textLeft + offsetX - padL,         textTop + offsetY + padB + height);
-    // Normal icon size mode
+    const size = layout.get('text-size').evaluate(feature, {}) / 24;
+    const stretchX = layout.get('icon-text-fit') === 'width' || layout.get('icon-text-fit') === 'both';
+    const stretchY = layout.get('icon-text-fit') === 'height' || layout.get('icon-text-fit') === 'both';
+
+    let left, right;
+    if (shapedText && stretchX) {
```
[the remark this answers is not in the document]
kkaefer: Yes, I've realized this too and moved most of the code to the beginning of `addFeature` as well to mirror where native is doing the stretching.

kkaefer:
For the sake of getting this merged let's defer dealing with the `icon-size` issue when `icon-text-fit` is set to another PR.

src/symbol/quads.js
```diff
@@ -44,58 +44,35 @@ export type SymbolQuad = {
...
-        br = new Point(right, bottom);
-        bl = new Point(left, bottom);
-    }
-
-    const angle = layer.layout.get('icon-rotate').evaluate(feature, {}) * Math.PI / 180;
+    // Expand the box to respect the 1 pixel border in the atlas image. We're using `image.paddedRect - border`
+    // instead of image.displaySize because we only pad with one pixel for retina images as well, and the
+    // displaySize uses the logical dimensions, not the physical pixel dimensions.
+    const iconWidth = shapedIcon.right - shapedIcon.left;
+    const expandX = (iconWidth * image.paddedRect.w / (image.paddedRect.w - 2 * border) - iconWidth) / 2;
```
alexshalamov: nit: border is always 1, can be omitted?

kkaefer: Yes, it's always `1` in our current implementation. However, I'd like to keep it as is to keep the code readable and avoid using magic numbers.

alexshalamov: maybe we can keep it, easier to understand that it is a 1px border on both sides.

kkaefer:
> One question, not directly related to changes in this PR. In some of the tests, for instance, [both-text-anchor-1x-image-2x-screen](https://github.com/mapbox/mapbox-gl-js/blob/897129a97a71a512e0191648c49e03b8f25f592f/test/integration/render-tests/icon-text-fit/both-text-anchor-1x-image-2x-screen/expected.png), text is not center-aligned to the anchor, is it a bug?

Can you say more? I believe that this test image is correct.

alexshalamov:
> Can you say more? I believe that this test image is correct.

Visually, looks like text should be shifted few pixels to the left. Letter 't' is touching right edge of a stretched icon, while there is quite a big gap in front if 'Ü'
image

kkaefer:
Hm, I believe that this can be attributed to the glyph metrics of this particular letter. For comparison, here's a rendering with a different letter which goes all the way to the edge:

actual-1

kkaefer:
This PR sizes and positions the collision boxes correctly, except when using `text-variable-anchor`. In that case, the size is correct, but the position isn't. @ansis, how can we ensure that icons shifted by the variable text-anchor placement is reflected in the collision index as well?

ansis:
I pushed two commits to [`fix-icon-text-fit-variable`](https://github.com/mapbox/mapbox-gl-js/tree/fix-icon-text-fit-variable) that fix collisions for icon-text-fit when combined with variable text anchors. The first commit just shifts the icon collision box to match the chosen text position. From what I can understand this is what -native does. This doesn't work ideally in the case the text has no collisions but the `icon-text-fit` icon does. In this case the label might be hidden instead of shown in a different position. The second commit tries to avoid this by testing for icon collisions while picking the text anchor. That may be an improvement.

Both commits fail the icon-text-fit/text-variable-anchor render tests. I haven't investigated whether this are good or bad differences.

kkaefer:
Thanks @ansis! I only picked the first commit because this puts us at par with GL Native. Let's land the second patch in a separate PR.

asheemmamoowala:
> I only picked the first commit because this puts us at par with GL Native. Let's land the second patch in a separate PR.

@kkaefer is there a ticket or PR for the second patch mentioned here?

kkaefer:
Yes, it's at https://github.com/mapbox/mapbox-gl-js/pull/8803

PR #8747, opened 2019-09-10
Add icon-text-fit collision tests

alexshalamov:
Three new render tests requested in https://github.com/mapbox/mapbox-gl-js/issues/8722#issuecomment-527672938


test/integration/render-tests/icon-text-fit/both-collision-variable-anchor/style.json
```diff
@@ -0,0 +1,69 @@
...
+  "layers": [
+    {
+      "id": "symbol",
+      "type": "symbol",
+      "source": "geojson",
+      "layout": {
+        "text-field": "ABC",
+        "text-anchor": "top",
+        "text-offset": [0, 0.5],
+        "text-variable-anchor": ["left", "right"],
```
asheemmamoowala: I believe this test is showing a bug. The expected image is showing both left and right placements when the radial offset is provided. The correct expectation would be with only one of left or right rendered.

alexshalamov: I think it is correct, 2 features with same coordinates that are rendered by 2 layers. Only 1 feature from layer `"symbol_with_offset"` is rendered, as the second is hidden due to the collision. Layer with variable anchors is able to render both by trying placing these with radial offset.

Test verifies that icon collision box from `"symbol_with_offset"` layer is correctly shifted with a text box and does not collide with labels from `"symbol"` layer.

alexshalamov:
@asheemmamoowala thanks for review!

PR #8748, opened 2019-09-10
[docs] Address user feedback on resize and *RenderedWorldCopies

colleenmcginnis:
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/19
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/51
Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/65

This PR is the first in a series that will address user feedback on the contents of the API docs including adding inline examples to all sections of the API docs and auditing `Related` links for relevance. 

cc @mapbox/docs

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -667,6 +667,9 @@ class Map extends Camera {
      *
      * @param {LngLatLike} lnglat The geographical location to project.
      * @returns {Point} The {@link Point} corresponding to `lnglat`, relative to the map's `container`.
+     * @example
+     * var coordinate = [-122.420679, 37.772537];
+     * var point = map.project(coordinate);
      */
     project(lnglat: LngLatLike) {
         return this.transform.locationPoint(LngLat.convert(lnglat));
@@ -678,7 +681,11 @@ class Map extends Camera {
      *
      * @param {PointLike} point The pixel coordinates to unproject.
      * @returns {LngLat} The {@link LngLat} corresponding to `point`.
-     * @see [Show polygon information on click](https://www.mapbox.com/mapbox-gl-js/example/polygon-popup-on-click/)
+     * @example
+     * map.on('click', function(e) {
+     *   // When the map is clicked, get the geographic coordinate.
+     *   var coordinate = map.unproject(e.point);
+     * });
      */
     unproject(point: PointLike) {
         return this.transform.pointLocation(Point.convert(point));
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -573,6 +573,8 @@ class Map extends Camera {
      * @param {number | null | undefined} minZoom The minimum zoom level to set (0-24).
      *   If `null` or `undefined` is provided, the function removes the current minimum zoom (i.e. sets it to 0).
      * @returns {Map} `this`
+     * @example
+     * map.setMinZoom(12.25);
      */
     setMinZoom(minZoom?: ?number) {
 
@@ -593,6 +595,8 @@ class Map extends Camera {
      * Returns the map's minimum allowable zoom level.
      *
      * @returns {number} minZoom
+     * @example
+     * map.getMinZoom();
      */
     getMinZoom() { return this.transform.minZoom; }
 
@@ -604,6 +608,8 @@ class Map extends Camera {
      * @param {number | null | undefined} maxZoom The maximum zoom level to set.
      *   If `null` or `undefined` is provided, the function removes the current maximum zoom (sets it to 22).
      * @returns {Map} `this`
+     * @example
+     * map.setMaxZoom(18.75);
      */
     setMaxZoom(maxZoom?: ?number) {
 
@@ -619,6 +625,16 @@ class Map extends Camera {
 
         } else throw new Error(`maxZoom must be greater than the current minZoom`);
     }
+
+    /**
+     * Returns the map's maximum allowable zoom level.
+     *
+     * @returns {number} maxZoom
+     * @example
+     * map.getMaxZoom();
+     */
+    getMaxZoom() { return this.transform.maxZoom; }
+
 
     /**
      * Returns the state of renderWorldCopies.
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -276,39 +276,45 @@ class Map extends Camera {
     _requestManager: RequestManager;
 
     /**
-     * The map's {@link ScrollZoomHandler}, which implements zooming in and out with a scroll wheel or trackpad.
+     * The map's `ScrollZoomHandler`, which implements zooming in and out with a scroll wheel or trackpad.
+     * Find more details and examples using `scrollZoom` in the {@link ScrollZoomHandler} section.
```
mourner: Maybe worth clarifying all the handler comments with what you can actually do with the property? E.g. "Allows you to control scroll zooming behavior dynamically (toggle on and off, adjust options, etc.)".

src/ui/map.js
```diff
@@ -482,11 +498,19 @@ class Map extends Camera {
      * Resizes the map according to the dimensions of its
      * `container` element.
      *
-     * This method must be called after the map's `container` is resized by another script,
+     * Checks if the map container size changed and updates the map if it has changed.
+     * This method must be called after the map's `container` is resized by another script
```
mourner: Nit: "another script" might not be a little confusing — perhaps change to "programmatically"?

src/ui/map.js
```diff
@@ -507,13 +531,17 @@ class Map extends Camera {
     /**
      * Returns the map's geographical bounds. When the bearing or pitch is non-zero, the visible region is not
      * an axis-aligned rectangle, and the result is the smallest bounds that encompasses the visible region.
+     * @example
+     * map.getBounds();
```
mourner: Perhaps worth adding `var bounds = ...` in examples like this to show that it returns a value?

src/ui/map.js
```diff
@@ -203,7 +203,7 @@ const defaultOptions = {
  * @param {number} [options.pitch=0] The initial pitch (tilt) of the map, measured in degrees away from the plane of the screen (0-60). If `pitch` is not specified in the constructor options, Mapbox GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
  * @param {LngLatBoundsLike} [options.bounds] The initial bounds of the map. If `bounds` is specified, it overrides `center` and `zoom` constructor options.
  * @param {Object} [options.fitBoundsOptions] A [`fitBounds`](#map#fitbounds) options object to use _only_ when fitting the initial `bounds` provided above.
- * @param {boolean} [options.renderWorldCopies=true]  If `true`, multiple copies of the world will be rendered, when zoomed out.
+ * @param {boolean} [options.renderWorldCopies=true]  If `true`, multiple copies of the world will be rendered side by side when the map is zoomed out far enough that a single representation of the world does not fill the map's entire container.
```
andrewharvey: I know this was an issue with the docs even before you're change here, but technically it's not just zoomed out. zoomed right in to z14 looking showing a map of Fiji you want renderWorldCopies to true otherwise half the country will be cut off.

Perhaps a better way instead of saying when zoomed out is, saying outside of the -180 to 180degrees longitude bounds?

src/ui/map.js
```diff
@@ -276,39 +276,45 @@ class Map extends Camera {
...
-     * The map's {@link ScrollZoomHandler}, which implements zooming in and out with a scroll wheel or trackpad.
+     * The map's `ScrollZoomHandler`, which implements zooming in and out with a scroll wheel or trackpad.
+     * Find more details and examples using `scrollZoom` in the {@link ScrollZoomHandler} section.
      */
     scrollZoom: ScrollZoomHandler;
 
     /**
-     * The map's {@link BoxZoomHandler}, which implements zooming using a drag gesture with the Shift key pressed.
+     * The map's `BoxZoomHandler`, which implements zooming using a drag gesture with the Shift key pressed.
+     * Find more details and examples using `boxZoom` in the {@link BoxZoomHandler} section.
```
andrewharvey: I think having the @link at the very first mention of BoxZoomHandler is better. Anytime you first encounter a new technical term, having a link to further explain it is very helpful as a beginner. Moving the link further down makes it less likely I'd see and click the link.

src/ui/map.js
```diff
@@ -531,6 +559,14 @@ class Map extends Camera {
      *
      * @param {LngLatBoundsLike | null | undefined} bounds The maximum bounds to set. If `null` or `undefined` is provided, the function removes the map's maximum bounds.
      * @returns {Map} `this`
+     * @example
+     * // Define bounds that conform to the `LngLatBoundsLike` object.
+     * var bounds = [
+     *   [-74.04728, 40.68392], // Southwest coordinates
```
andrewharvey: People often get thrown off by the coordinates order, I know we already mention this is LngLat, but for absolute clarity it might be best to change the comment here to something which is more explicit, eg.

```
// [west, south]
// [east, north]
```

Or something like that?

The numbers in there are helpful for someone working in the geographic area of the coordinates as they'll likely recognise the order from the numbers alone, but generally not helpful for an international audience, so we can't rely on those alone. eg. if I see "150, -34" I know that's `long, lat` order, but to someone in the US that's probably not helpful.

andrewharvey:
A lot of good changes here, thanks! I just made a few comments.

src/ui/map.js
colleenmcginnis: I started doing this, but it felt repetitive. What do you think about skipping this for now and reevaluating if we continue to get feedback on these sections?

src/ui/map.js
colleenmcginnis: I think a visual would go a long way here. https://github.com/mapbox/mapbox-gl-js-docs/pull/72 for your consideration. 💁

colleenmcginnis: Also -- I didn't update the text for `getRenderWorldCopies` and `setRenderWorldCopies` yet, but once we're comfortable with the language here I can update those, too.

Issue #8679, opened 2019-08-22
Fix flaky text-variable-anchor/pitched-with-map render test


mourner:
This render test [failed in a commit](https://circleci.com/gh/mapbox/mapbox-gl-js/49591) that just changed the changelog, so it's apparently flaky (it passed after rerunning the tests) — let's investigate. cc @alexshalamov

alexshalamov:
@mourner Just tried running gl-native tests

`for i in {1..20}; do build/linux-x86_64/Release/mbgl-render-test text-variable-anchor -r --shuffle --seed=$RANDOM; done`

![image](https://user-images.githubusercontent.com/1001009/63546731-2407b680-c533-11e9-9fbf-850de0affc85.png)

Really strange that it is flaky on gl-gs, we had few flaky tests in native, as the rendering results were created in gl-js and native uses floats in placement code.

mourner:
[Another example of a flaky test](https://circleci.com/gh/mapbox/mapbox-gl-js/49862), this time it's `text-variable-anchor/pitched` — the pattern of failure is very similar:

![image](https://user-images.githubusercontent.com/25395/63866680-429b0100-c9bc-11e9-8260-422a39f92d4e.png)

kkaefer:
Got another instance of this. @arindam1993, @alexshalamov let's fix this.

PR #8751, opened 2019-09-11
Another attempt at fixing AWS on Circle

mourner:
Follow-up to #8750 
cc @kkaefer


mourner:
Render failure is #8679 (should be ignored)

PR #8754, opened 2019-09-14
[docs] Address user feedback for /api: sources, images, layers

colleenmcginnis:
This PR is one in a series that will address user feedback on the contents of the API docs including adding inline examples to all sections of the API docs and auditing `Related` links for relevance. 

As always, all feedback is welcome! But, there are some specific things that need a close look:

- Am I talking about sprites + images added using `addImage` correctly? 
- Thoughts on adding context to `@see` resources (for example, ` * @see Vector source: [Show and hide layers](https://docs.mapbox.com/mapbox-gl-js/example/toggle-layers/)`)?
- Style layer zoom extents vs. source layer zoom extents in the `setLayerZoomRange` description.

cc @mapbox/docs

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1361,6 +1361,7 @@ class Map extends Camera {
      * 
      * @param id The ID of the image.
      *
+     * @returns {Boolean}  A Boolean indicating whether the image exists.
      * @example
      * // Check if an image with the ID 'cat' exists in 
      * // the style's sprite.
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1244,7 +1244,7 @@ class Map extends Camera {
      * @returns {?Object} The style source with the specified ID, or `undefined`
      *   if the ID corresponds to no existing sources.
      * @example
-     * var sourceId = map.getSource('points');
+     * var sourceObject = map.getSource('points');
      * @see [Create a draggable point](https://www.mapbox.com/mapbox-gl-js/example/drag-a-point/)
      * @see [Animate a point](https://www.mapbox.com/mapbox-gl-js/example/animate-point-along-line/)
      * @see [Add live realtime data](https://www.mapbox.com/mapbox-gl-js/example/live-geojson/)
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -1405,13 +1509,42 @@ class Map extends Camera {
...
     }
 
+    /**
+     * Sets the zoom extent for the specified style layer. The zoom extent includes the
+     * [minimum zoom level](https://docs.mapbox.com/mapbox-gl-js/style-spec/#layer-minzoom)
+     * and [maximum zoom level](https://docs.mapbox.com/mapbox-gl-js/style-spec/#layer-maxzoom))
+     * at which the layer will be rendered.
+     *
+     * Note: For style layers using vector sources, style layers cannot be rendered at zoom levels lower than the
+     * minimum zoom level of the _source layer_ because the data does not exist at that zoom level. If the minimum
```
colleenmcginnis:
```suggestion
     * minimum zoom level of the _source layer_ because the data does not exist at those zoom levels. If the minimum
```

src/ui/map.js
```diff
@@ -1144,9 +1144,29 @@ class Map extends Camera {
      * {@link CanvasSourceOptions}.
      * @fires source.add
      * @returns {Map} `this`
-     * @see [Draw GeoJSON points](https://www.mapbox.com/mapbox-gl-js/example/geojson-markers/)
-     * @see [Style circles using data-driven styling](https://www.mapbox.com/mapbox-gl-js/example/data-driven-circle-colors/)
-     * @see [Set a point after Geocoder result](https://www.mapbox.com/mapbox-gl-js/example/point-from-geocoder-result/)
```
andrewharvey: ha, I just discovered this independently, was about to submit my PR when I noticed this.
Since https://github.com/mapbox/mapbox-gl-js/commit/6e6a0057bd4d6db3757c8775821115c44e778987#diff-44befe964c5c5c476298e33cc55ce3e7 theo geocoder example no longer uses addSource.

PR #8737, opened 2019-09-07
port fix for icon-text-fit combined with variable text placement

ansis:
Addresses : #8583 

This is still very much a WIP that doesn't work yet. https://github.com/mapbox/mapbox-gl-native/pull/15367.

When I got to porting the changes in `placement.cpp` I tried adding the all at once because I didn't see a clear way of doing it incrementally. But that may have not been the best plan since it doesn't work and I'm not sure if it's close or if there are many large bugs remaining.

The next steps for working on this could be to try run `yarn run test-render icon-text-fit` and debug the crashes and problems one by one as they appear. But it may actually be better to do a careful readthrough of the original PR and make sure everything is included. Finally, if that doesn't work, it might be easiest to start with a fresh branch and gradually cherry-pick changes from this branch over or use it as a guide.

PR #8755, opened 2019-09-14
Fix icon-fit with variable label placement

arindam1993:
This is an attempt to isolate and cherry pick just the fix required for #8583 from #8737 .
The issue seems to be caused by the fact that `draw_symbol.js:#updateVariableAnchors` applies an offset to the text glyphs , but never touches the icon buffer at all.

This PR adds an attribute to the `placedSymbol` buffers, called `associatedSymbolIndex`, this lets the text `placedSymbol` instances know which icon they're associated with.
So at render time the offsets can be applied in both buffers.
It also moves the offset calculation out of the paint loop, since icons and text need to share the same data.

This still leaves out making `icon-text-fit` update the collision box (#8722) and making it work with vertical labels. 

UPDATE: 
demo, 1 frame desynch fixed

Added a render test
Before (on master):
actual

After:
expected

## Launch Checklist

 - [ ] post benchmark scores
 - [ ] ~tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec changes~


src/data/bucket/symbol_bucket.js
```diff
@@ -539,7 +539,8 @@ class SymbolBucket implements Bucket {
                writingMode: any,
                labelAnchor: Anchor,
                lineStartIndex: number,
-               lineLength: number) {
+               lineLength: number,
+               associatedIconIndex: number) {
```
mourner: Looking at the `addSymbols` calls, `associatedIconIndex` seems to be unused (always undefined). So it's added as `0`?

debug/8583.html
```diff
@@ -0,0 +1,133 @@
+<!DOCTYPE html>
```
asheemmamoowala: Are you planning to commit this file? It would be more useful for the next developer if it was named more appropriately.

src/data/bucket/symbol_bucket.js
arindam1993: No, there are two calls in `symbol_layout.js` both have an index added. Its fixed to `-1` for the icon bucket, and the `placedIconIndex` for the the text vertices in `addTextVertices`

src/render/draw_symbol.js
```diff
@@ -233,18 +301,19 @@ function drawLayerSymbols(painter, sourceCache, layer, coords, isText, translate
...
+            bucket.hasIconData();
+
         if (alongLine) {
             symbolProjection.updateLineLabels(bucket, coord.posMatrix, painter, isText, labelPlaneMatrix, glCoordMatrix, pitchWithMap, keepUpright);
-        } else if (isText && size && variablePlacement) {
-            const tileScale = Math.pow(2, tr.zoom - tile.tileID.overscaledZ);
-            updateVariableAnchors(bucket, rotateWithMap, pitchWithMap, variableOffsets, symbolSize,
-                                  tr, labelPlaneMatrix, coord.posMatrix, tileScale, size);
-        } else if (isText && size && bucket.allowVerticalPlacement) {
+        } else if (isText && size && bucket.allowVerticalPlacement && !variablePlacement) {
```
vakila: Perhaps I'm missing something but why the `&& !variablePlacement` here? IIRC we can have both vertical & variable placement enabled at the same time, are we accounting for that case?

src/render/draw_symbol.js
```diff
@@ -55,6 +55,17 @@ function drawSymbols(painter: Painter, sourceCache: SourceCache, layer: SymbolSt
     // Disable the stencil test so that labels aren't clipped to tile boundaries.
     const stencilMode = StencilMode.disabled;
     const colorMode = painter.colorModeForRenderPass();
+    const variablePlacement = layer.layout.get('text-variable-anchor');
+
+    //Compute variable-offsets before painting since icons and text data positioning
+    //depend on each other in this case.
```
vakila: Nice :+1:

src/render/draw_symbol.js
arindam1993: No, we're not. In the interest of time time this PR only handles `icon-text-fit` rendering for horizontal variable anchor labels.

arindam1993:
> LGTM! Thanks for adding the render test @arindam1993 , both that & the debug page (on Linux Firefox/Chrome) look good 💚
> 
> One question below, and relatedly: do we need tests that this works also when combined with multi-line and/or vertical labels?
> 
> Some of those (currently ignored) tests are failing, for example `text-writing-mode/point_label/mixed-multiline-vertical-horizontal-mode-icon-text-fit`, and the failure doesn't look to me like it has to do with the 1px issue addressed in #8741. Thoughts?

Yes, will address that and shifting of icon collision boxes in a separate PR. This was intended as a fix to the biggest visual glitch.

src/render/draw_symbol.js
vakila: :ok_hand: Maybe we should ticket that out

src/data/array_types.js
```diff
@@ -422,10 +422,11 @@ register('StructArrayLayout2ub2f12', StructArrayLayout2ub2f12);
  * [28]: Float32[2]
  * [36]: Uint8[3]
  * [40]: Uint32[1]
+ * [44]: Int16[1]
  *
  * @private
  */
-class StructArrayLayout2i2ui3ul3ui2f3ub1ul44 extends StructArray {
+class StructArrayLayout2i2ui3ul3ui2f3ub1ul1i48 extends StructArray {
```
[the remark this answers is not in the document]
arindam1993: Thats what I did the change is in `symbol_attributes.js`

arindam1993:
> Is the intention to fix `text-writing-mode/point_label/cjk-variable-anchors-vertical-horizontal-mode-icon-text-fit` as well? Regardless of the `icon-text-fit` aspect, it seems like the
> 
> Here's the native version with collision boxes activated (which arguably seems wrong too):
> 
> actual
> 
> And this is the JS version in this branch:
> 
> actual
> 
> for reference, this is the JS version _before_ this patch:
> 
> actual

No I didn't intend to fix that, this change only ensures that the offset thats calculated and applied to `text-variable-anchor` at render time `updateVariableAnchors` gets applied to the icon quad as well, it doesnt touch placement or layout.


> It looks like there are no tests that are unignored, which means there are render test cases missing that were broken in the old behavior, but work with this patch applied. Could you please add some?

I have added one(see original PR body), but its not fully representative of this fix since its doing collision. Updating so `icon/text-overlap` is `true` so collision is ignored but the probelm with un-offset icons is still surfaced.

Master:
actual
This branch:
expected

arindam1993:
@asheemmamoowala 
Cleaner render tests added:
Master:
`icon-text-fit/text-variable-anchor`:
actual

`icon-text-fit/text-variable-anchor-overlap`:
actual


This branch:
`icon-text-fit/text-variable-anchor`:
expected

`icon-text-fit/text-variable-anchor-overlap`:
expected

arindam1993:
I've fixed the vertical label getting hidden issue. this is what the `text-writing-mode/point_label/cjk-variable-anchors-vertical-horizontal-mode-icon-text-fit` looks like now:
actual

The box sizing is still incorrect, but this PR isn't intended to fix that.

Issue #4872, opened 2017-06-23
stop camera animation when constrained by maxBounds


andrewharvey:
When using `flyTo` with `maxBounds`, the camera will keep trying to animate when hitting the bounds.

https://jsbin.com/ritikupaqe/edit?html,output

Should we change this so that when the transform becomes constrained, the animation is stopped at that point?

stevage:
Interesting. In this case, you're doing something totally unreasonable: saying "I never want to see anything outside this box", and then trying to move the camera a long long way outside the box. Are there reasonable use cases where this is a problem? I kind of think it would be fair for mapbox-gl-js to return an error here.

andrewharvey:
I think it's fair enough to have either option

a) let the application developer do their own checks before calling any of the camera setters, and/or
b) the camera setters to do their own internal check and return an error when the requested camera is outside the maxBounds

It should be clear who is responsible for that check.

> Are there reasonable use cases where this is a problem?

The GeolocateControl with a maxBounds set on the map when a user is located outside the maxBounds.

The problem here is the [geolocate event](https://www.mapbox.com/mapbox-gl-js/api/#geolocatecontrol.event:geolocate) is triggered after setting the camera to the new view so you can't currently deal with this case in your application.

We could build in checks into the GeolocateControl to detect and report back on this, but I'm not sure of the best place in the "stack" for the check to happen.

stevage:
Oh, yeah, that use case - happens a lot when someone in Europe makes a cool map and you try to view it from overseas and...yep. 

Maybe the issue that "max bounds" isn't really a meaningful concept for a *map*, it's really a constraint that should be applied to individually to different actions, like user scroll, geolocating, manually triggered by code...

andrewharvey:
Another design is build the check into all the camera setters and if a request comes in which would end with the camera outside maxBounds, then don't change the camera and trigger a 'outside max bounds' event.

That way no matter what's making the changes to camera, the application developer can detect via the event.

Either way we're going of a tangent from the "stop camera animation when constrained by maxBounds" issue :wink:

mollymerp:
yeah, I think this is undesired behavior/a bug. I am in favor of emitting an error and bailing if someone wants to animate to a place outside of the constrained bounds. @mourner @lbud does that make sense to you?

andrewharvey:
closely related to #3231

andrewharvey:
[...]

PR #8756, opened 2019-09-14
Check map maxBounds in Geolocate control

MoradiDavijani:
## Launch Checklist
Changes:
The GeolocateControl with a `maxBounds` set on the map when a user is located outside of the `maxBounds`, keeps trying to move the camera outside of the `maxBounds` and causes a bad experience for users. In this PR, I checked map `maxBounds` on geolocate success callback and if the located position is out of `maxBounds`, remove location dot marker, update control `classList`, emit an event called `outofmaxbounds` and return to avoid camera update.

Related issue: #4872

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -156,6 +156,7 @@ class GeolocateControl extends Evented {
             this._geolocateButton.classList.add('mapboxgl-ctrl-geolocate-active-error');
 
             this.fire(new Event('outofmaxbounds', position));
+            this._updateMarker()
             this._finish();
             
             return;
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -137,7 +137,30 @@ class GeolocateControl extends Evented {
         this._map = (undefined: any);
     }
 
+    _isOutOfMapMaxBounds(position: Position) {
+        const bounds = this._map.getMaxBounds();
+        const coordinates = position.coords;
+
+        return bounds && (
+            coordinates.longitude < bounds.getWest() ||
+            coordinates.longitude > bounds.getEast() ||
+            coordinates.latitude < bounds.getSouth() ||
+            coordinates.latitude > bounds.getNorth()
+        )
+    }
+
     _onSuccess(position: Position) {
+        if (this._isOutOfMapMaxBounds(position)) {
+            this._watchState = 'ACTIVE_ERROR';
+            this._geolocateButton.classList.remove('mapboxgl-ctrl-geolocate-active');
+            this._geolocateButton.classList.add('mapboxgl-ctrl-geolocate-active-error');
+
+            this.fire(new Event('outofmaxbounds', position));
+            this._finish();
+            
+            return;
+        }
+
         if (this.options.trackUserLocation) {
             // keep a record of the position so that if the state is BACKGROUND and the user
             // clicks the button, we can move to ACTIVE_LOCK immediately without waiting for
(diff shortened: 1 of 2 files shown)


andrewharvey:
This won't entirely fix #4872, since the underlying issue there is requesting a `flyTo` outside the `maxBounds` is currently undefined behaviour, so that would need to be resolved to close #4872.

If we decide in #4872 to require the API user to make their own check, then in my opinion this proposed solution is the way to go. We make that check first internally in GeolocateControl, and then only flyTo if it's valid, and then emit an error back to the API user so they can decide what to do (like show an alert), exactly how you've done this PR.

In lieu of a decision on #4872, I think it's good to implement this PR even if just for the time being (although based on my comment at https://github.com/mapbox/mapbox-gl-js/issues/4872#issuecomment-531558961, I do think this is probably the right approach).

src/ui/control/geolocate_control.js
```diff
@@ -137,7 +137,31 @@ class GeolocateControl extends Evented {
...
+            coordinates.longitude < bounds.getWest() ||
+            coordinates.longitude > bounds.getEast() ||
+            coordinates.latitude < bounds.getSouth() ||
+            coordinates.latitude > bounds.getNorth()
+        );
+    }
+
     _onSuccess(position: Position) {
+        if (this._isOutOfMapMaxBounds(position)) {
+            this._watchState = 'ACTIVE_ERROR';
```
andrewharvey: per what we do at
https://github.com/mapbox/mapbox-gl-js/blob/87cac3be79880fe6b6cf532baa023c52b112cf1f/src/ui/control/geolocate_control.js#L221-L245

we should only set the ACTIVE_ERROR state when the map is in active tracking, if you're in background tracking and then you go outside the maxBounds then we should leave the error icon as a background error and not active error.

MoradiDavijani: You're right. Is it fine to add another class for out of bounds poistion? Something like `mapboxgl-ctrl-geolocate-out-of-bounds`?

andrewharvey: Unless you're proposing a new style for the button, then I'd say stick with the current classes.

test/unit/ui/control/geolocate.test.js
```diff
@@ -48,6 +48,29 @@ test('GeolocateControl error event', (t) => {
     geolocation.sendError({code: 2, message: 'error message'});
 });
 
+test('GeolocateControl outofmaxbounds event', (t) => {
```
andrewharvey: could you also add a test case for the background error state?

src/ui/control/geolocate_control.js
```diff
@@ -137,7 +137,57 @@ class GeolocateControl extends Evented {
...
+
+        return bounds && (
+            coordinates.longitude < bounds.getWest() ||
+            coordinates.longitude > bounds.getEast() ||
+            coordinates.latitude < bounds.getSouth() ||
+            coordinates.latitude > bounds.getNorth()
+        );
+    }
+
+    _goToErrorState() {
```
andrewharvey: do you think `_setErrorState` would be clearer?

MoradiDavijani: Sure, thanks.

andrewharvey:
cc @ryanhamley I'm happy with this change, are we good to merge or should I wait for someone from Mapbox to have a chance to review?

ryanhamley:
Looks good to me! Thanks @MoradiDavijani and @andrewharvey

PR #8758, opened 2019-09-16
Increase threshold for icon-text-fit collision tests

jmalanen:


alexshalamov:
Looks like interpolation issue. Locally test passes, @jmalanen does the test pass when you run it locally without new "allowed" threshold?

image

jmalanen:
> Looks like interpolation issue. Locally test passes, @jmalanen does the test pass when you run it locally without new "allowed" threshold?
> 
> image

No, it does not pass locally when running in the same way as in CI: `xvfb-run --server-args="-screen 0 1024x768x24" ...`

Issue #8655, opened 2019-08-18
Consider making "className" dynamic


noxecane:
## Motivation
Issue #6299 made a case to add `className` to popup options, hence making it possible to style the entire popup. I'm in a position where that style changes dynamically every second or so, and would want to update the specific popup to reflect the new style.

## Suggestions
1. A method added to the `Popup` class, `toggleClass` that will give me the ability remove/add  a class dynamically since `className` is basically a proxy to `classList.add`
2. A method added to the `Popup` class, `setClassName` that will give me the ability where I update all the classes in one call

## Mockup
```js
const myPop = new Popup({ className: 'rounded offline' }).setText('This is my home')

// .....in some evented scenario for suggestion-1
listener.on('new-event', (x) => {
  x.toggleClass('is-open')
})

// .....same scenario for suggestion-2 with more context
listener.on('new-event', (x) => {
  x.setClassName(x.online ? 'rounded online' : 'rounded offline')
})
```

I would prefer the second one as it fits my context the most.
```

**Comment by ahk** [comment]
I think both options are fine though I think the `toggleClass` would see more general usage. I'd be inclined to accept both of these suggestions.

Would you have any interest in making a pull request for this? I'd be happy to help get it merged. If not I may be able to fit this in next week.

**Comment by Ashot-KR** [comment]
@ahk here you go #8759. Description in PR

@noxecane actually you can do it without any changes of `Popup` class, but in a little bit verbose way:
```js
const popup = new Popup()

// can access to `ClassList` of popup instance:
const { classList } = popup.getElement()

// use .add, .remove, .toggle of ClassList
classList.add('new-class')
classList.remove('new-class')
classList.toggle('toggle-class')
```

noxecane:
Wow thanks...just saw this....26 days ago 🙆🏽‍♂️

PR #8759, opened 2019-09-16
add Popup methods to manipulate container class names

Ashot-KR:
Added 3 methods to `Popup`:
`addClassName` – add className to container
`removeClassName` – remove className from container
`toggleClassName` – toggle className

All methods is just shortcut wrappers around `popup._container.classList` methods respectively.

Related to #8655


Ashot-KR:
I agree.
I reverted initial `className` setup to `forEach` instead of `classList.add` with multiple arguments (yes, i forgot to check browser compatibility before i did that :bowtie:)
`addClassName`, `removeClassName` and `toggleClassName` now accepts only single `string` argument

colleenmcginnis:
Thanks @mourner! Tagging in @danswick for docs review.

.gitignore
```diff
@@ -25,3 +25,4 @@ _site
 yarn-error.log
 yarn-debug.log
 npm-debug.log
+.idea
```
vakila: Is this addition intentional/necessary?

src/ui/popup.js
```diff
@@ -359,10 +400,10 @@ export default class Popup extends Evented {
             this._tip       = DOM.create('div', 'mapboxgl-popup-tip', this._container);
             this._container.appendChild(this._content);
             if (this.options.className) {
-                this.options.className.split(' ').forEach(name =>
-                    this._container.classList.add(name));
+                this.options.className.split(' ').forEach(name => {
+                    this._container.classList.add(name);
+                });
```
vakila: Is this change necessary, or perhaps just something left over from a previous change that modified this and was then reverted?

src/ui/popup.js
```diff
@@ -332,6 +332,47 @@ export default class Popup extends Evented {
         return this;
     }
 
+    /**
+     * Adds class to container element.
```
vakila:
```suggestion
     * Adds a CSS class to the popup container element.
```

src/ui/popup.js
```diff
@@ -332,6 +332,47 @@ export default class Popup extends Evented {
...
+     *
+     * @param {string} className non empty string
```
vakila:
```suggestion
     * @param {string} className Non-empty string with CSS class name to add to popup container 
```

src/ui/popup.js
```diff
@@ -332,6 +332,47 @@ export default class Popup extends Evented {
...
+     * @example
+     * let popup = new mapboxgl.Popup()
+     * popup.addClassName('some-class')
+     */
+    addClassName(className: string) {
+        this._container.classList.add(className);
+    }
+
+    /**
+     * Removes class from container element.
```
vakila:
```suggestion
     * Removes a class from the popup container element.
```

src/ui/popup.js
```diff
@@ -332,6 +332,47 @@ export default class Popup extends Evented {
...
+     *
+     * @param {string} className non empty string
```
vakila:
```suggestion
     * @param {string} className Non-empty string with CSS class name to remove from popup container 
```

src/ui/popup.js
```diff
@@ -332,6 +332,47 @@ export default class Popup extends Evented {
...
+     * @example
+     * let popup = new mapboxgl.Popup()
+     * popup.removeClassName('some-class')
+     */
+    removeClassName(className: string) {
+        this._container.classList.remove(className);
+    }
+
+    /**
+     * When only one argument is present toggle the class value.
```
vakila:
```suggestion
     * Add or remove the given class on the popup container, depending on whether the container currently has that class.
```

src/ui/popup.js
```diff
@@ -332,6 +332,47 @@ export default class Popup extends Evented {
...
+     *
+     * @param {string} className non empty string
```
vakila:
```suggestion
     * @param {string} className Non-empty string with CSS class name to add/remove
```

test/unit/ui/popup.test.js
```diff
@@ -495,7 +495,7 @@ test('Popup#remove is idempotent (#2395)', (t) => {
     t.end();
 });
 
-test('Popup adds classes from className option', (t) => {
+test('Popup adds classes from className option, methods for class manipulations works properly', (t) => {
```
vakila: Thanks for adding these tests!

Perhaps we should also have tests for edge cases, e.g. if someone passes an empty string (`''`) or a space-separated string (`'multiple class names'`) like they can pass to the `className` option? Would be great to capture the intended behavior for those cases.

.gitignore
Ashot-KR: `.idea` folder contains project settings for Jetbrains IDEs (Intellij/PhpStrom/WebStorm etc). These IDE pretty popular in webdev, so i thought having it in `.gitingore` could be usefull.
Anyway it's not necessary and could be removed.

src/ui/popup.js
Ashot-KR: In previous commit i rewrite foreEach loop to one line function `this._container.classList.add(...name)`(or something like that), but i forgot to check browser compatibility – IE11 does not support multiple arguments, so i revert it to old good `forEach` to keep IE11 working. I noticed it [in this comment](https://github.com/mapbox/mapbox-gl-js/pull/8759#issuecomment-539913416)

vakila: Thanks for the quick reply! Yeah I noticed the revert, I think just the brackets changed from the original (https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/popup.js#L367-L368) - not really a big deal but no point touching it if we don't need to.

Ashot-KR: Without brackets function returns value, which is pointless for `forEach`. I think it's better use arrow function with brackets here. But as you said it's not a big deal, so i can remove it if you think that we should keep original variant.

test/unit/ui/popup.test.js
Ashot-KR: My idea for these methods was to keep it closer as possible to native `classList` methods. Native methods throws `DOMException` in these cases(`''` or `'multiple class names'`). And at this point we have 2 options:
1. Just add `t.throws` test for these cases
2. Add checks to my methods to prevent exceptions (e.g. `if (className) { this._container.classList.add(className) }`)

Which one is preferable?

vakila: 1 is fine - the idea is to use the tests to capture/document that we've thought about these edge cases & what behavior we expect, it's fine if we expect errors.

src/ui/popup.js
vakila: Yes exactly, all this does is change the return value of the function, which isn't used anyway. I don't see this change as necessary/an improvement but you can leave it in if you want to.

vakila:
@Ashot-KR Thanks for getting back to us quickly! Left a few clarifications re the points I raised, but I wouldn't consider those real blockers. Could you resolve the merge conflicts & lint errors so we can get this ready to merge in? Thanks!

test/unit/ui/popup.test.js
Ashot-KR: tests added

Ashot-KR:
@vakila conflict resolved
also added some changes from conversations above

vakila:
Awesome, thanks a lot @Ashot-KR! Appreciate the extra tests :heart: 

I just noticed it looks like the `addClass` docs updates didn't actually make it into https://github.com/mapbox/mapbox-gl-js/pull/8759/commits/b00e6eb0c7221bb9194477eb443b507300c2d925 - not a huge deal but for consistency with the other methods I just pushed those changes to your fork branch.

So once the checks run I think this is good to go! :rocket: Thanks again for your work on this!

Ashot-KR:
> docs updates didn't actually make it into b00e6eb 

Ooops, something went wrong…
Thanks for fixing that

PR #8761, opened 2019-09-16
Add few levels of nesting to text-color-overrides render test

alexshalamov:


alexshalamov:
@asheemmamoowala correct, I forgot to backport one fix that @ansis made for text-color override PR.

Issue #8560, opened 2019-07-26
Allow array as value to text-radial-offset


tristen:
## Motivation

Like `text-offset` that accepts number or an array of two numbers that adjusts the `x` and `y` coordinates, it would be great if `text-radial-offset` allowed the same functionality. That would account for the extra distance top positioned labels need for characters with descenders (i.e `g, p, q`):

<img width="286" alt="Screen Shot 2019-07-26 at 10 09 31 AM" src="https://user-images.githubusercontent.com/61150/61958276-09bbd680-af8f-11e9-9907-ad79d8e37133.png">

## Design Alternatives

Not provide `top` as a list of possible positions to `text-variable-anchor` but that's a little limiting.

## Design

```js
{
  text-radial-offset: [0.5, 1] // x, y
}
```

asheemmamoowala:
@tristen Are you proposing that the array takes only 2 values -`[a, b]`, where `a` is the offset applied to all non-`top` anchor values, and `b` is applied only to `top` ?

This would be very confusing to document and explain. For example, when `text-variable-anchor: ["top"]` , the single value should be used, but changing it to `text-variable-anchor: ["top", right]` would require switching to an array `[ <right_offset>, <top_offset>]`.

What if instead auto justification took into account the full height of symbols so that it anchors to the bottom of glyhps before adding the `text-radial-offset` value?

cc @vakila @ansis

tristen:
> This would be very confusing to document and explain. For example, when `text-variable-anchor: ["top"]` , the single value should be used, but changing it to `text-variable-anchor: ["top", right]` would require switching to an array `[ <right_offset>, <top_offset>]`.

Good point.

> What if instead auto justification took into account the full height of symbols so that it anchors to the bottom of glyhps before adding the `text-radial-offset` value?

👍

mapcarta:
Yes this is needed. One way to do this without confusion is that the 'text-radial-offset' array must contain either exactly two or four elements:

- An array with two elements would specify the vertical (top and bottom) and horizontal (left and right) spacing, in that order.
- An array with four elements would specify the top, right, bottom and left offsets, in that order.

This way 'text-variable-anchor' could be changed without any trouble. 

> What if instead auto justification took into account the full height of symbols so that it anchors to the bottom of glyhps before adding the text-radial-offset value?

I agree this the latter would be a far more elegant and desirable solution if it is possible.

mapcarta:
[...]

PR #8781, opened 2019-09-20
Introduce `descender` and `ascender` for font metrics + fix several text rendering issues

zmiao:
This pr is covering:
- [x] Introduce `descender` and `ascender` for font metrics, which are important for aligning text baselines when mixed fonts are used.
This pr will only load `descender` and `ascender` for fonts loaded from server-side ( _**background: mapbox/node-fontnik#160 will introduce new metadata ascender and descender, it would help fix the issue mapbox/mapbox-gl-js#191. This pr only covers reading data from pbf and partially replacing hardcoded baseline for fixing mixed fonts dis-alignment**_ ). Support for loading metrics for local fonts is not implemented yet.
- [x] Fix vertical text shaping for characters that have very big part below baselines, such as `p`, `g`, `y`, etc.

|  Before |  After|
|---|---|
|before1|expected|

- [x] Fix vertical text shaping with ZWSP and punctuations.

|  Before |  After|
|---|---|
| actual |expected|

- [x] Fix vertical text rendering with `text-offset` property: https://github.com/mapbox/mapbox-gl-js/issues/10687
  - [ ] Add new render tests(point label and line label)
- [x] Fix collision box generation for vertical text
- [x] Fix text rendering when both `text-offset` and `text-rotate` are set

```
 "text-rotate": 45,
  "text-offset": [
      3,
      6
  ]
```

|  Before |  After|
|---|---|
| expected|expected|

- [x] Enable corresponding ignored render tests
- [x] Check render tests failure

Follow-up:
Introduce font metrics for local glyph generation (TinySDF). 
Fix issue mapbox/mapbox-gl-js#8560 

## Launch Checklist


zmiao:
Some rendering results after applying the `ascender` value as baseline for fixing the dis-alignment of mixed fonts:
 
|   After fix |  Before fix |
|---|---|
|  Screen Shot 2019-09-17 at 6 43 29 PM |  Screen Shot 2019-09-17 at 6 44 50 PM |  
|  Screen Shot 2019-09-17 at 6 51 54 PM | Screen Shot 2019-09-17 at 6 51 20 PM |

src/symbol/shaping.js
```diff
@@ -479,15 +496,20 @@ function shapeLines(shaping: Shaping,
...
+
             const positions = glyphMap[section.fontStack];
             const glyph = positions && positions[codePoint];
-
             if (!glyph) continue;
 
+            // In order to make different fonts aligned, they must share a general baseline that starts from the midline
+            // of each font face.  Baseline offset is the vertical distance from font face's baseline to its top most
+            // position, which is the half size of the sum (ascender + descender). Since glyph's position is counted
+            // from the top left corner, the negative shift is needed. So different fonts shares the same baseline but
```
chloekraw: nit: different fonts **share**

src/symbol/shaping.js
```diff
@@ -479,15 +496,20 @@ function shapeLines(shaping: Shaping,
...
+            // with different offset shift. If font's baseline is not applicable, fall back to use a default baseline
```
chloekraw: > So different fonts shares the same baseline but with different offset shift.

Why do we want different offsets for different fonts? 

Or did you mean different characters have different offsets?

zmiao:
@chloekraw for your concern about the third picture,

> Looking at the third render test, I don't see a difference in the before vs. after photo. What am I missing?

 my intention of this is to show that there is no obvious rendering differences for vertical layout texts. Since the fix is mainly for horizontal text dis-aligned problem. Right now all the vertical texts aligned with the middle line of the shaping box. 

> It looks like there's too little space under the g, though perhaps there's not much we can or should do about that.

> It also looks like there's not enough space under the p / too much space under the o. If characters are placed vertically, we may want to offset so that the space between characters begins after the descent line if there is a descender, rather than using the true baseline.

This could be anther text shaping issue to fix later in the future. `Descender` needs to be taken into account.

> I think the eye is looking for the space between where one character ends and the next one starts to be even, and the idea of a baseline is basically moot when there's only a single character in the line.

Baseline is necessary even if there is only one character, we need to use baseline for creating glyph, and also taking baseline into account for `text-variable-offset` calculation, which is related to issue  #8560 I plan to land pr after this pr is merged.

src/symbol/shaping.js
zmiao: Screen Shot 2019-09-26 at 7 44 50 PM

For your question:
> Why do we want different offsets for different fonts?

Taking the above picture as example, the blue box is a shaping box.  There are two font types. `ascender` and `descender` are font-level attributes, for each font `ascender` is the distance above the baseline, and `descender` is the negative distance below the baseline. As you can see, these two fonts have different baseline and different `ascender` and `descender`. But our goal is to make the two character aligned visually, we could choose different alignment, align from the bottom line or align form the middle line. Here we choose using the middle line, it means we need to put each character around the middle line of the shaping box, they share the midline as a virtual baseline. 

When processing `shapeLines` function,  for each character, we create a `positionedGlyph` with its top-left coordinates (x, y) value in the axis plane. For value of y, which is actually the same height as `ascender`. As we can see the middle line of each glyph is (`ascender` - `descender`) / 2. (`descender` is negative). The offset of `ascender` along Y axis is (`ascender` - `descender`) / 2, since it is against the positive direction, the offset is  -(`ascender` - `descender`) / 2. 

> Or did you mean different characters have different offsets?

Since `ascender` and `descender` are font level attributes, so the offset will be different for different fonts. 

But later when we create glyph, the actual top of each character will be taken into account.

chloekraw: Oh, I see! Thank you for the detailed explanation, @zmiao, this is really helpful.

chloekraw:
@zmiao gotcha, thanks! I misunderstood the scope of your pr. 

> Since the fix is mainly for horizontal text dis-aligned problem. 

Makes sense!

zmiao:
Hi @mourner, for your concern:  

>  - It's unfortunate that this PR would add 36 MB of test data, practically doubling the size of the repo. Can we reduce this in any way?

Thanks for providing this information. The *.pbf data I added is intended to test some characters I used in the test. I will remove some characters so that corresponding pbf file could be erased. 
 
> - If `ascender` and `descender` are font-level attributes, do we have to save those in each glyph? Can we refactor the code to avoid the redundancy?

This is a good suggestion! I will adapt the code. Thanks!

zmiao:
@chloekraw  one updates, actually this patch will also fix vertical text dis-alignment if there are multiple font scales in one line. Please check my updated test result picture.

src/render/glyph_atlas.js
```diff
@@ -20,22 +20,31 @@ export type GlyphPosition = {
...
+};
+
+export type GlyphPositions = { [string]: GlyphPositionData };
 
 export default class GlyphAtlas {
     image: AlphaImage;
     positions: GlyphPositions;
 
-    constructor(stacks: { [string]: { [number]: ?StyleGlyph } }) {
+    constructor(stacks: {[string]: {glyphs: {[number]: ?StyleGlyph}, ascender: number, descender: number}}) {
```
asheemmamoowala: `{[number]: ?StyleGlyph}, ascender: number, descender: number}`

This could be condensed into a new type given that it is used in multiple places in the codebase.

src/symbol/shaping.js
```diff
@@ -210,12 +215,12 @@ function shapeText(text: Formatted,
         right: translate[0],
         writingMode,
         lineCount: lines.length,
-        yOffset: -17 // the y offset *should* be part of the font metadata
+        yOffset: -17, // the y offset *should* be part of the font metadata
```
asheemmamoowala: Since `yOffset` is not provided from the font metadata per glyph, it could be removed from this structure and the value hard-coded separately for cases where baseline data is not available.

chloekraw:
> @chloekraw one updates, actually this patch will also fix vertical text dis-alignment if there are multiple font scales in one line. Please check my updated test result picture.

Thanks @zmiao, looks great. So to be clear, how did you fix this issue? Does this PR only fix vertical text misalignment if all text is CJK and uses the same font family?

src/symbol/quads.js
```diff
@@ -99,93 +99,100 @@ export function getGlyphQuads(anchor: Anchor,
...
     const positionedGlyphs = shaping.positionedGlyphs;
     const quads = [];
 
-    for (let k = 0; k < positionedGlyphs.length; k++) {
-        const positionedGlyph = positionedGlyphs[k];
-        const glyphPositions = positions[positionedGlyph.fontStack];
-        const glyph = glyphPositions && glyphPositions[positionedGlyph.glyph];
-        if (!glyph) continue;
+    if (shaping.lineCount === 0) return quads;
+    const lineHeight = (Math.abs(shaping.top) + Math.abs(shaping.bottom)) / shaping.lineCount;
```
ansis: I think this might not work correctly in the case that `shaping.top` and `shaping.bottom` have the same sign. I'm not sure when that might be but maybe this should be `(bottom - top) / lineCount`?

src/symbol/shaping.js
```diff
@@ -35,7 +35,7 @@ export type PositionedGlyph = {
 
 // A collection of positioned glyphs and some metadata
 export type Shaping = {
-    positionedGlyphs: Array<PositionedGlyph>,
+    positionedGlyphs: {[number]: Array<PositionedGlyph>},
```
ansis: Is it important for the key to be any number or can this be an array of arrays instead? It seems like using an array might simplify some things elsewhere like the ` for (const lineIndex of Object.keys(positionedGlyphs).map(Number)) {` in quads.js

zmiao:
@asheemmamoowala  

> One question I have is whether this will affect local glyph generation adversely and by how much?

Good question. It won't affect individual glyph generation, it still uses the hardcoded value.  But I am concerning that if it is possible for the following case:

- Only part of the glyphs for the fontstack could be generated locally, but remaining part of glyphs of the same fontstack needs to be requested from pbf data, then `ascender` and `descender` which are now fontstack level attributes will be messed up. 

So from this point of view, `ascender` and `descender` still need to be glyph specific attributes.

zmiao:
@chloekraw 

> Thanks @zmiao, looks great. So to be clear, how did you fix this issue? Does this PR only fix vertical text misalignment if all text is CJK and uses the same font family?

This fix is working for mixed fonts. But to be honest, the vertical text misalignment is not fully fixed yet, if there are multiple lines inside the vertical text, and fonts are mixed, the look of the vertical shaped text is still not good. 

My fix in this pr for the vertical dis-alignment is  just replacing the hardcoded `yOffset` with  the glyph offset when baseline is available.  But still, vertical text shaping needs to be taken out as a separate issue and further fix which taking glyph's `ascender` and `descender` value into consideration is needed.

asheemmamoowala:
> Good question. It won't affect individual glyph generation, it still uses the hardcoded value. But I am concerning that if it is possible for the following case:

@zmiao Since local glyphs don't have these extra font metrics (baseline, ascender and descender) will the alignment between server generated and locally generated glyphs be worse that it is now?

zmiao:
@asheemmamoowala 
> Since local glyphs don't have these extra font metrics (baseline, ascender and descender) will the alignment between server generated and locally generated glyphs be worse that it is now?

I talked with the team, so for the same font, all the glyphs shall be either local rasterized or fetched.

If it is local rasterized, for gl-native, the `ascender` and `descender` could be fetched from the platform dependent font API(Not implemented yet).  But for gl-js, it is not possible. Right now both gl-native and gl-js will fall back to use Shaping::yOffset as a baseline.  There will be a follow-up task in gl-native side to fetch `ascender` and `descender` from the font API for local rasterized glyphs. For gl-js, @tmpsantos wants to align with you at first and discuss about the possible solutions. 
 
If it is fetched from server, then the `ascender` and `descender` value will directly be assigned from metadata got from the sever. 

So all in all, it won't become worse, the worst case would just be using the hardcoded baseline, which is the same behavior before this fix.

zmiao:
@ansis , updates about your question:

> And a general question: It looks like a font range pbf shares one ascender and descender. Is this because a range never has multiple fonts in it?

yes, it is true. For the same fontstack, `ascender` and `descender` value will be global.

src/symbol/shaping.js
```diff
@@ -35,17 +35,20 @@ export type PositionedGlyph = {
 
 // A collection of positioned glyphs and some metadata
 export type Shaping = {
-    positionedGlyphs: Array<PositionedGlyph>,
+    positionedGlyphs: Array<Array<PositionedGlyph>>,
```
zmiao: I got a flow error that I don't understand, @ansis @asheemmamoowala , any hints about it?
```
Error -----src/symbol/shaping.js:38:35



`PositionedGlyph` [1] is incompatible with array type [2].

   src/symbol/shaping.js:38:35
   38|     positionedGlyphs: Array<Array<PositionedGlyph>>,
                                         ^^^^^^^^^^^^^^^ [1]

References:
   src/symbol/shaping.js:38:29
   38|     positionedGlyphs: Array<Array<PositionedGlyph>>,
                                   ^^^^^^^^^^^^^^^^^^^^^^ [2]
```

ansis: I'm not sure exactly what is going on but I think it isn't compatible with the typing of `isEmpty`. Adding `any` to `function isEmpty(array: any) {` seems to avoid this. But maybe adding real typing to `isEmpty` would be better

zmiao: Thank you very much!

mourner:
Closing this PR for now since it's not clear when we can expect server-side changes required to land it, and related tickets haven't been updated in ~3 months. Let's reopen when there are clear next actions on the GL JS side.

zmiao:
@ansis @asheemmamoowala could you please have a second round of review for this pr? 🙇‍♀️

mourner (approved):
👍

src/symbol/quads.js
```diff
@@ -228,11 +242,30 @@ export function getGlyphQuads(anchor: Anchor,
...
+    const lineCounts = shaping.positionedLines.length;
+    const lineHeight = shapingHeight / lineCounts;
+    const getMidlineOffset = function(shaping, lineHeight, previousOffset, lineIndex) {
+        const currentLineHeight = (lineHeight + shaping.positionedLines[lineIndex].lineOffset);
+        if (lineIndex === 0) {
+            return previousOffset + currentLineHeight / 2.0;
+        }
+        const aboveLineHeight = (lineHeight + shaping.positionedLines[lineIndex - 1].lineOffset);
+        return previousOffset + (currentLineHeight + aboveLineHeight) / 2.0;
+    };
```
mourner, line 264: Nit: functions that don't need to be a closure (nested within another function) should be moved to the top scope.

src/symbol/quads.js
```diff
@@ -265,62 +302,123 @@ export function getGlyphQuads(anchor: Anchor,
...
+
+                // Relative position after rotating
+                // tr ----- br
+                //   |     |
+                //   |     |
+                // tl ----- bl
+                // After rotation, glyph lies on the horizontal midline.
+                // Shift back to tl's original x coordinate before rotation by applying 'xOffsetCorrection'.
+                const offsetCorrection = new Point(-yShift + halfAdvance, 0);
+                tl._add(offsetCorrection);
```
mourner, line 372: Nit: simplify to avoid `Point` allocation:

```suggestion
                tl.x += -yShift + halfAdvance;
```

src/symbol/quads.js
```diff
@@ -265,62 +302,123 @@ export function getGlyphQuads(anchor: Anchor,
...
+                    tl.x += delta / 2;
+                } else {
+                    // Place the glyph bitmap right in the center of the 24x24 point boxes
+                    const delta = verticalAdvance - paddedHeight;
+                    tl.x += delta / 2;
+                }
+                // Calculate other three points
+                tr = tl.add(new Point(0, -paddedWidth));
+                bl = tl.add(new Point(paddedHeight, 0));
+                br = tl .add(new Point(paddedHeight, -paddedWidth));
```
mourner, line 404: Nit: let's avoid `Point` allocations where not needed — e.g. doing `tr = new Point(...)` directly would make it one instead of two.

src/util/verticalize_punctuation.js
```diff
@@ -88,14 +88,14 @@ export const verticalizedCharacterMap = {
     '｣': '﹂'
 };
 
-export default function verticalizePunctuation(input: string) {
+export default function verticalizePunctuation(input: string, skipChontextChecking: boolean) {
```
ChrisLoer, line 91: Nit: Is this supposed to be "context"?

src/symbol/shaping.js
```diff
@@ -611,22 +637,44 @@ function shapeLines(shaping: Shaping,
...
+                const glyphData = glyphMap[section.fontStack];
+                if (!glyphData) continue;
+                const glyph = glyphData.glyphs[codePoint];
+                if (!glyph) continue;
+
+                metrics = glyph.metrics;
+                verticalAdvance = codePoint !== 0x200b ? ONE_EM : 0;
+
+                // In order to make different fonts aligned, they must share a general baseline that aligns with every
+                // font's real baseline. Glyph's offset is counted from the top left corner, where is the ascender line
```
ChrisLoer, line 656: Nit: where is the ascender -> where the ascender

ChrisLoer (commented):
🎉  I didn't dig too deep into the logic b/c it looks like a straight port from gl-native, and we already have a pretty full test-suite to run it against. I had just a couple typo-nits. Two other questions:

- Have you run perf benchmarks with the changes? I don't see anything that looks problematic, but there could be some surprises.
- Although this code has been in gl-native for a while, we haven't really exposed customers much to it yet, right? So how are we thinking about what their reaction will be if their map suddenly changes appearance because we went from guessing the baseline to actually loading it correctly from the font? In some cases they may have done something like added an offset to approximate the right baseline, or who knows what else. Is the idea just "this is a small change, it makes things overall better, some small tweaking to your map may be necessary?"

zmiao:
@ChrisLoer thank you very much for your comment, for your concerns:

> Have you run perf benchmarks with the changes? I don't see anything that looks problematic, but there could be some surprises.

I run the benchmark tests, the result seems good:
Screen Shot 2021-06-07 at 7 37 11 PM
Screen Shot 2021-06-07 at 7 35 37 PM

> Although this code has been in gl-native for a while, we haven't really exposed customers much to it yet, right? So how are we thinking about what their reaction will be if their map suddenly changes appearance because we went from guessing the baseline to actually loading it correctly from the font? In some cases they may have done something like added an offset to approximate the right baseline, or who knows what else. Is the idea just "this is a small change, it makes things overall better, some small tweaking to your map may be necessary?"

- The biggest change is about introducing the descender and ascender metrics, however, currently both server and local sides' glyphs don't contain these two metrics.  So it won't bring any surprises for now. If both sever fonts and local fonts are used, as long as there is one font doesn't have descender and ascender metrics, we will still use the default metrics, thus the glyphs rendering behavior will be almost the same for horizontal text. 
I also talked with @asheemmamoowala , the follow-up tasks would be 1. Investigating introduce font metrics for local fonts. 2. Preparation/test work to ensure the correct behavior with sever fonts having ascender/descender, once we decided to land https://github.com/mapbox/node-fontnik/pull/160. (Currently I also have render tests with glyph pbfs that containing the ascender/descender metrics generated with branch https://github.com/mapbox/node-fontnik/pull/160)

- The other changes are mainly for fixing sevral text rendering issues(vertical text rendering, wrong position of text box with `text-offset`/`text-rotate` property), the texts will be rendered with a better look, so I assume the customers will accept them.

PR #8788, opened 2019-09-21
Fix 'cache is not defined' TypeErrors in cachePut/cacheGet

vakila:
Following on from https://github.com/mapbox/mapbox-gl-js/pull/8780, move `catch`es to end of the promise chain to avoid `TypeError: cache is not defined` in the subsequent `then`s. 

Launch checklist:
- [x] describe changes
- [x] add tests(s)

### DIFF
--- a/src/util/tile_request_cache.js
+++ b/src/util/tile_request_cache.js
@@ -81,7 +81,6 @@ export function cacheGet(request: Request, callback: (error: ?any, response: ?Re
             // manually strip URL instead of `ignoreSearch: true` because of a known
             // performance issue in Chrome https://github.com/mapbox/mapbox-gl-js/issues/8431
             cache.match(strippedURL)
-                .catch(callback)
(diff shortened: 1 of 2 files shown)


src/util/tile_request_cache.js
```diff
@@ -95,7 +94,9 @@ export function cacheGet(request: Request, callback: (error: ?any, response: ?Re
 
                     callback(null, response, fresh);
                 });
```
mourner: Should we move other occurrences of `catch` like the one after `match`?

vakila:
To really cover our bases here I think we should add some unit tests to make sure errors are being handled as expected - I'm going to take a stab at that now.

vakila:
Added some basic tests for `cacheGet`. `cachePut` still needs more tests but may need to be refactored to make testing easier - will follow up on that in a separate PR

PR #8798, opened 2019-09-24
Add release testing page

kkaefer:
This adds a release testing page that allows cycling through various test cases quickly without having to enter the URLs manually or checking out the docs repository.

Run `npm run start-release` to create all required files and start the server.

A few notes:
- This PR is against the release branch, but I'll merge to master once we've proven its usefulness in release testing.
- All it does currently is to navigate through the pages. You can hit reload in your browser; it'll preserve the current state in the location hash.
- It prints a QR code to the console that you can scan with your phone for easy access.
- Possible features in a next iteration are:
  - Add testing description and notes
  - Add ability to load the same example with a previous version from the CDN
  - Include screenshots so testers can compare
  - Send feedback to a server after user has clicked a Confirm/Reject button


test/release/index.js
```diff
@@ -0,0 +1,136 @@
...
+    jsProdMin.href = "/dist/mapbox-gl.js";
+    const css = document.createElement("a");
+    css.href = "/dist/mapbox-gl.css";
+
+    const titleElement = document.querySelector('#title');
+    const container = document.querySelector('#container');
+    const prevButton = document.querySelector('#prev');
+    const nextButton = document.querySelector('#next');
+
+    const pages = [
```
arindam1993: combining all of the test json metadata into one file, with key's being these names file would make it DRY'er .

ahk:
Thank you for this, it's great.

kkaefer:
This is ready for re-review.

build/generate-release-list.js
```diff
@@ -0,0 +1,21 @@
+const octokit = require('@octokit/rest')();
+const fs = require('fs');
+
+const list = {};
+
+octokit
+    .paginate(octokit.repos.listReleases.endpoint({
```
asheemmamoowala: Does this need to get _all_ the releases? The dropdown to pick previous versions is nice, but in my ad-hoc testing I found the some of the examples do not work on older releases because the style or runtime functionality are not supported. 
A simple alternative would be to use URL query parameters to specify the version.

build/generate-release-list.js
```diff
@@ -0,0 +1,21 @@
+const octokit = require('@octokit/rest')();
```
asheemmamoowala: How/when is this being executed? I don't see any references to it in the `package.json`. Does this tool belong elsewhere?

kkaefer: It's referenced in https://github.com/mapbox/gl-internal/wiki/Javascript-release-process#publish-release-list

build/generate-release-list.js
kkaefer: It doesn't _have_ to get all releases; we could e.g. limit to the last 20 releases or so. Parsing the version from the dropdown works, but I found it convenient when testing on mobile devices.

test/release/index.js
```diff
@@ -0,0 +1,254 @@
+/* eslint-env browser */
+/* eslint-disable prefer-arrow-callback,prefer-template */
+/* eslint no-loop-func: "off" */
+/* eslint camelcase: "off" */
+/* global mapboxgl */
+/* global mapboxglVersions */
+
+const pages = {
```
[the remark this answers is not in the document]
kkaefer: Added

Issue #6708, opened 2018-05-22
Clarify documentation for map.loaded(), map.isStyleLoaded()


anandthakker:
**mapbox-gl-js version**: 0.45.0

Refs https://github.com/mapbox/mapbox-gl-js/issues/6707

The documentation for `map.loaded()` and `map.isStyleLoaded()` do not make it clear what the difference is between them.

```js
    /**
     * Returns a Boolean indicating whether the map is fully loaded.
     *
     * Returns `false` if the style is not yet fully loaded,
     * or if there has been a change to the sources or style that
     * has not yet fully loaded.
     *
     * @returns {boolean} A Boolean indicating whether the map is fully loaded.
     */
    loaded()
```

```js
    /**
     * Returns a Boolean indicating whether the map's style is fully loaded.
     *
     * @returns {boolean} A Boolean indicating whether the style is fully loaded.
     */
    isStyleLoaded()
```

It should be clear from reading the docs:
 - What "fully loaded" means / what might cause it to be not true.
 - Under what conditions `map.loaded() !== map.isStyleLoaded()`

mike-marcacci:
I agree that some clarification would be helpful here. My initial assumption was that `isStyleLoaded()` would return false before the initial load and between a `styledataloading` event and subsequent `styledata` event, indicating that layers were not ready for queries. However, it appears to return false under a far wider range of conditions.

stevage:
IMHO the important thing to include here is the implications of the current state: can you add sources, can you change paint or layout properties, can you manipulate the state otherwise.

mourner:
Related: #5052, which stems from the same confusion over what `map.loaded()` means.

hyperknot:
Is there a reference document about a map's lifecycle events? I'd like to know how are 

- map.loaded()
- map.isStyleLoaded()
- map.on('load')
- map.on('idle')
- map.on('style.load')

depending on each other during the lifecycle of map.

Namely, is such a recursive setTimeout loop needed for anything still?
https://github.com/mapbox/mapbox-gl-js/issues/2268#issuecomment-401979967

ollydixon:
A quick example of how I do it, but an ```isLoaded``` call back would be great!

```
const goToLocation = (mapEvent: MapboxGl.Map) => {
        map = mapEvent;
        console.log('goToLocation')
        if (mapEvent.loaded()) {
            map.resize()
            map.flyTo({ center: [10.7522, 59.9139], animate: false, zoom: 13 })
[...]

Issue #6707, opened 2018-05-22
`loaded()` and `on('load')` event do not work as expected


gmaclennan:
**mapbox-gl-js version**: 0.45.0

### Question

There are certain methods that you cannot call while the map is still loading, such as `map.addLayer()`. You can use the `map.on('load')` event to know when it is safe to call these methods, however there is no corresponding method like `map.ready()` for knowing that it is safe to call `map.addLayer()`. `map.loaded()` can also return false during other method calls, such as `map.setLayoutProperty()`, and you can safely call `map.addLayer()` when `map.loaded()` returns `false`. Also if `map.loaded()` returns `false` then `map.on('load')` only works the first time, so you need to track the initial load state yourself.

Every time I have tried to make a complex map with mapbox-gl I have hit some kind of frustrating bug that was related to loading state and these events. They have changed gradually over versions but they still don't work as expected.

It would be great if `mapbox-gl` could just track its ready state internally, so that calling `map.addLayer()` just worked as expected straight away. Either that or we need methods that correspond to load events. I want to do something like this:

```js
if (map.loaded()) {
  map.addLayer(...)
} else {
  map.on('load', () => map.addLayer(...))
}
```

But this will only work the first time (`map.on('load')` only fires once), but sometimes `map.loaded()` will return false if the map is moving.

anandthakker:
Thanks for describing this pain point @gmaclennan.

You're right: `map.loaded()` answers _both_ whether the style has been initially loaded _and_ whether there have been source/layer changes that haven't yet been processed.   In general, it's safe to do things like `addLayer()` any time after the former is true, regardless of the latter.  

I think adding a `map.hasStyle()` method that answers only the first question (and is, I think, what `map.ready()` in the OP is after), makes sense.

anandthakker:
I think `style.load` is equivalent to the _first_ `styledata`: https://github.com/mapbox/mapbox-gl-js/blob/a5a8dfe5b90737dbb295eaca7f41c667ae4060a8/src/style/style.js#L243-L244

`styledata` would also be fired subsequently, after other changes to the style have been applied: e.g., loading the 'sprite' sheet (a separate resource), adding/removing sources, adding and removing images

gmaclennan:
Thanks. I am continuing to experiment with this, and hit another strange behavior: I was trying to get layers to fade in after the tiles have loaded for certain map views. I am doing this:

```js
if (map.areTilesLoaded()) doSomething()
[...]

PR #8807, opened 2019-09-25
[docs] Address user feedback for /api: setStyle, getStyle, isStyleLoaded

colleenmcginnis:
Closes https://github.com/mapbox/mapbox-gl-js-docs/issues/53

This PR is one in a series that will address user feedback on the contents of the API docs including adding inline examples to all sections of the API docs and auditing `Related links` for relevance.

cc @mapbox/docs

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1032,9 +1032,14 @@ class Map extends Camera {
     }
 
     /**
-     * Updates the map's Mapbox style object with a new value. If a style already is set and options.diff is true,
-     * this compares the style against the map's current state and performs only the changes necessary to make
-     * the map style match the desired state.
+     * Updates the map's Mapbox style object with a new value.
+     *
+     * If a style is already set when this is used and options.diff is set to true, the map renderer will attempt to compare the given style
+     * against the map's current state and perform only the changes necessary to make the map style match the desired state. Changes in sprites
+     * (images used for icons and patterns) and glyphs (fonts for label text) **cannot** be diffed. If the sprites or fonts used in the current
+     * style and the given style are different in any way, the map renderer will force a full update, removing the current style and building
+     * the given one from scratch.
+     *
      *
      * @param style A JSON object conforming to the schema described in the
      *   [Mapbox Style Specification](https://mapbox.com/mapbox-gl-style-spec/), or a URL to such JSON.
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -1032,9 +1032,14 @@ class Map extends Camera {
     }
 
     /**
-     * Updates the map's Mapbox style object with a new value. If a style already is set and options.diff is true,
-     * this compares the style against the map's current state and performs only the changes necessary to make
-     * the map style match the desired state.
+     * Updates the map's Mapbox style object with a new value.
+     *
+     * If a style is already set when this is used and options.diff is set to true, the map renderer will attempt to compare the given style
```
asheemmamoowala: It would be useful to add why it is valuable to enable to diff. If a style is not changing drastically, it enables a cross-faded change from the current style to the next and it is more performant - by reusing existing source and layer definitions and transitioning them to a future state - it saves on CPU cycles, and possibly network bandwidth.

src/ui/map.js
```diff
@@ -1118,6 +1127,10 @@ class Map extends Camera {
      * Returns the map's Mapbox style object, which can be used to recreate the map's style.
      *
      * @returns {Object} The map's style object.
+     *
+     * @example
+     * var styleJson = map.getStyle();
```
asheemmamoowala: It would be nice to add a comment with a minimal example of the data structure this returns

src/ui/map.js
```diff
@@ -1129,6 +1142,9 @@ class Map extends Camera {
      * Returns a Boolean indicating whether the map's style is fully loaded.
      *
      * @returns {boolean} A Boolean indicating whether the style is fully loaded.
+     *
+     * @example
```
asheemmamoowala: Any chance I can add to the scope fo this PR to include improvements to address https://github.com/mapbox/mapbox-gl-js/issues/6708 and https://github.com/mapbox/mapbox-gl-js/issues/6707 as well?

src/ui/map.js
colleenmcginnis: Is the `@returns {Object} The map's style object.` line above not sufficient? Maybe we should add more detail there instead?

asheemmamoowala: Thats sufficient !

arindam1993:
LGTM, and merging, thanks @colleenmcginnis !

colleenmcginnis:
Thanks @arindam1993!

Issue #8771, opened 2019-09-18
Memory keeps increasing with each zoom interaction on iPhone / Macbook Safari browser


alankaanil:
**mapbox-gl-js version**: v1.3.1

**browser**: Safari for iPhone / Macbook

I am using a single HTML page with [example](https://docs.mapbox.com/mapbox-gl-js/example/simple-map/) provided in Mapbox-gl-js page.

### Steps to Trigger Behavior

 1. Open the web page on Safari browser
 2. Progressively Zoom out to maximum level, then zoom in back to initial level.
 3. Memory keeps increasing each time Step 2 is repeated, but it does not go down.

### Link to Demonstration

https://codepen.io/anilalanka/pen/pozxPGx
Please add your access token for testing.
```
<!DOCTYPE html>
<html>
<head>
    <meta charset='utf-8' />
    <title>Display a map</title>
    <meta name='viewport' content='initial-scale=1,maximum-scale=1,user-scalable=no' />
    <script src='https://api.tiles.mapbox.com/mapbox-gl-js/v1.3.1/mapbox-gl.js'></script>
    <link href='https://api.tiles.mapbox.com/mapbox-gl-js/v1.3.1/mapbox-gl.css' rel='stylesheet' />
    <style>
        body { margin:0; padding:0; }
        #map { position:absolute; top:0; bottom:0; width:100%; }
    </style>
</head>
<body>

<div id='map'></div>
<script>
    mapboxgl.accessToken = 'your-access-token';
    var map = new mapboxgl.Map({
        container: 'map', // container id
        style: 'mapbox://styles/mapbox/streets-v11', // stylesheet location
        center: [-74.50, 40], // starting position [lng, lat]
        zoom: 9 // starting zoom
    });
</script>

</body>
</html>
```

### Actual Behavior

The following are the Memory profile details captured on Macbook Safari Web inspector, based on style URL used.

When memory consumed is large enough, 
- Macbook Safari browser throws a warning saying `This webpage is using significant memory. Closing it may improve the responsiveness of your Mac.`. 
- iPhone Safari app refreshes the page.

#### 1. Using Mapbox default style URL:
Memory increases upto **1.06 GB in 80 seconds**, as shown below
[...]

PR #8814, opened 2019-09-27
Reduce allocated GPU textures 

arindam1993:
Merge #8850 before merging this.
cc @kkaefer  @asheemmamoowala 

- We were leaking the vertex buffers for the tile masks, this PR deletes those buffers when the tile is unloaded
- we also hold onto an unecessarilly large number of gpu texture objects, even though we always re-render and reupload into them in `draw_hillshade.js#prepareHillshade(..)`
- This PR attempts to reduce memory usage by
 1. returning a texture to the texture pool in `map.painter._tileTextures` when a tile is returned to the cache instead of when the tile is evicted form the cache
2. since a dem tile fetched from the cache does not contain the prepared dem texture, schudle the gpu draw again, this is a tradeoff of gpu render-call with memory.

Results:
800x800 viewport, data gathered by running the modified hillsahde example, with the instrumented gl calls, courtesy of @kkaefer  🙏 

this branch:
`{buffer: 102, texture: 16, frameBuffer: 6}`

on master:
`{buffer: 1516, texture: 105, frameBuffer: 51}`


src/render/texture.js
```diff
@@ -111,6 +113,10 @@ class Texture {
         return this.size[0] === this.size[1] && (Math.log(this.size[0]) / Math.LN2) % 1 === 0;
     }
 
+    get hashKey(): string {
+        return `${this.size[0]}.${this.premultiply ? 1 : 0}.${this.useMipmap ? 1 : 0}`;
```
astojilj: It should be safe not to care on UNPACK_PREMULTIPLY_ALPHA_WEBGL values when reusing texture allocations - in Texture.update value is set before texImage2D / texSubImage2D. It affects upload to texture.

ezraripps:
tested for #8771 did not fix https://www.bountysource.com/issues/80631646-memory-keeps-increasing-with-each-zoom-interaction-on-iphone-macbook-safari-browser

arindam1993:
This PR is definitely helping GPU memory usage, here are some benchmark numbers:

|           | gpu_memory | texture_memory | in_memory_textures | texture_upload_calls |
|-----------|------------|----------------|--------------------|----------------------|
| `master`  | 615        | 439            | 263                | 1665                 |
| `850ae21` | 375        | 198            | 43                 | 1784                 |

While this is a massive improvement, we're still uploading textures unnecessarily. I think a good way to move forward will be to use `TileCache` to cache dem textures in the painter, and use the pool similar to the one currently in the painter to cache the offscreen fbo's since those need to be re-rendered every frame as long as there is an expression that sets a paint property on the layer.

debug/hillshade_benchmark.html
```diff
@@ -0,0 +1,178 @@
+<!DOCTYPE html>
```
asheemmamoowala: Benchmark files are typically put in the `bench` folder. Similar to #8645 this technique could be used to detect leaks in specific layers or styles - either by adding it to the existing benchmarking tooling or as a separate set of benchmarks. That work can follow in a separate PR.

debug/index.html
```diff
@@ -15,8 +15,38 @@
 <div id='map'></div>
 
 <script src='../dist/mapbox-gl-dev.js'></script>
-<script src='access_token_generated.js'></script>
+<script src='../debug/access_token_generated.js'></script>
```
asheemmamoowala: Are changes in this file meant to be committed?

src/source/tile_cache.js
```diff
@@ -12,14 +12,14 @@ import type Tile from './tile';
  */
 class TileCache {
     max: number;
-    data: {[key: number | string]: Array<{ value: Tile, timeout: ?TimeoutID}>};
+    data: {[key: number | string]: Array<{ value: any, timeout: ?TimeoutID}>};
```
asheemmamoowala: Why do these types need to be relaxed? If it is to allow usage of methods only available on sub-classes of `Tile`, maybe those should be added to `Tile` or a type check used to conditionally execute those methods.

arindam1993: Because this cache never invokes anything on the `Tile` type, the design already uses an `onRemove` callback to define what needs to be called on removal of an element. So we can use it as datastore to store any data that has a 1:1 OR 1:n relationship with a tile, and automatically guarantee an upper bound on the memory consumption of those resources.

I want to use it for storing and caching `Texture` references in `Painter`, instead of adding more properties to `Tile`.

arindam1993:
|           | gpu_memory | texture_memory | in_memory_textures | texture_upload_calls |
|-----------|------------|----------------|--------------------|----------------------|
| `master`  | 615        | 439            | 263                | 1665                 |
| `850ae21` | 497        | 257            | 70                 | 956                 |

src/render/draw_hillshade.js
```diff
@@ -42,7 +42,7 @@ function drawHillshade(painter: Painter, sourceCache: SourceCache, layer: Hillsh
 function renderHillshade(painter, tile, layer, depthMode, stencilMode, colorMode) {
     const context = painter.context;
     const gl = context.gl;
-    const fbo = tile.fbo;
```
astojilj: @arindam1993 
I might have misunderstood the purpose of cache here.
The patch is introducing the cache in order to simplify evicting (when max size limit change) procedure - when there is need to remove content from cache, it is easier to go through cache.
This comes for a price we have to pay in every render call - instead of `tile.fbo`, there is `painter.getFboForTile('dem', tile.tileID);`. Note that similar would need to be also port to native.

Could we keep the previous and for evicting enumerate tiles in DEM sourceCache and remove tiles?

arindam1993: The purpose of the cache was to manage allocations and deallocations of per-tile GPU textures and FBO's.
Currently it's handled by the source cache, which can have an unnecessarily large number of cached tiles, some of which have been offscreen for a long time, so the idea was to have something in the middle. I've changed the design a bit now, so its a bit more temporal as well.

arindam1993:
Changed the design around a bit, here's what I have now.
- The fbo's with the offscreen rendered hillshade textures are no longer cached with the tile. Its discarded as soon as the tile goes offscreen, and will be re-rendered as soon as it comes back on-screen.
These were consuming a lion-share of the GPU memory for the tiles, so I think the tradeoff is probably worth it, most of the time is spent uploading the DEM texture instead of the offscreen pass.
- The DEM GPU textures are managed using the LRU tile_cache in the painter, with some new deferred shrinking logic. The size of the cache is set every frame, based on the number of required textures, growing it is cheap so that happens synchronously, but shrinking it is spread out over multiple frames and that process starts only if the target cache size( updated every frame) falls below 70% of the current cache size.

arindam1993:
I'm closing this for now since the performance changes won't be quantifiable until we have 
https://github.com/mapbox/mapbox-gl-js/pull/8931 ready, will reopen this once thats done.

asheemmamoowala:
Blocked by #8931

karimnaaji:
@arindam1993 is this still blocked and ? I see that https://github.com/mapbox/mapbox-gl-js/pull/8931 was cherry picked in some other PRs, so maybe the instrumentation we have in place would be enough to test this again if it is still relevant.

arindam1993:
I don't think this is super relevant anymore, caching/pooling strategies for fbo's and textures should be handled by specific layers internally based on what best works for that layer type.

Closing, and 🤦 'ing at my past self.

PR #8818, opened 2019-09-27
fix opacity interpolation for composition expressions

### LINKED ISSUE #8817: fill-opacity not interpolated correctly outside of a tile's zoom range
**mapbox-gl-js version**: 1.4.0

**browser**: all

### Steps to Trigger Behavior

 1. Create a style with an opacity expression that uses linear interpolation with both zoom and a data property:
```
"fill-opacity": [
            "interpolate",
            ["linear"],
            ["zoom"],
            0.5,
            0,
            1,
            ["match", ["get", "fakeprop"], "nope", 1, 1]
        ]

```
 2. start at zoom 0
 3. zoom to zoom 2

### Link to Demonstration

I'm opening a PR with a patch and regression test soon. This issue is just to document the problem.

### Expected Behavior

Opacity is applied as intended.

### Actual Behavior

Opacity is interpolated to a value greater than 1 resulting in the incorrect color for the split second before new tiles are loaded.

ansis:
fix #8817

The interpolation factor could be outside of the range 0..1 which could result in rendering errors.

This fix clamps the value to the correct range and adds a regression test cover the fix.

## Launch Checklist

I'm not sure if this bug exists in mapbox-gl-native.


test/integration/render-tests/regressions/mapbox-gl-js#8817/style.json
```diff
@@ -0,0 +1,67 @@
+{
+  "version": 8,
+  "metadata": {
+    "test": {
+      "width": 64,
+      "height": 64,
+      "operations": [
+          ["pauseSource", "geojson"],
```
asheemmamoowala: Would be great to have a comment explaining why `pauseSource` is needed for this test.

chloekraw:
@ansis should we check whether we clamp opacity interpolation for composition expressions in gl-native? even if we can't reproduce this specific bug, is there any reason we might not want to clamp it if it's not already?

PR #8829, opened 2019-10-02
add gpu timing map events

ansis:
Originally implemented by Chris in https://github.com/mapbox/mapbox-gl-js/pull/5967. I've rebased it and exposed it with just the event listeners instead of a map option.

Listen to `gpu-timing-frame` to get the gpu time for the frame and listen to `gpu-timing-layer` to get the gpu time for all individual layers. It is not recommended to listen to both.

This is not exposed as a public api.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/gl/context.js
+++ b/src/gl/context.js
@@ -115,7 +115,7 @@ class Context {
             gl.getExtension('OES_texture_half_float_linear');
         }
 
-        this.extTimerQuery = gl.getExtension('EXT_disjoint_timer_query');;
+        this.extTimerQuery = gl.getExtension('EXT_disjoint_timer_query');
     }
 
     setDefault() {
--- a/src/gl/context.js
+++ b/src/gl/context.js
@@ -63,6 +63,7 @@ class Context {
     extTextureFilterAnisotropic: any;
     extTextureFilterAnisotropicMax: any;
     extTextureHalfFloat: any;
+    extTimerQuery: any;
 
     constructor(gl: WebGLRenderingContext) {
         this.gl = gl;
@@ -113,6 +114,8 @@ class Context {
         if (this.extTextureHalfFloat) {
             gl.getExtension('OES_texture_half_float_linear');
         }
+
+        this.extTimerQuery = gl.getExtension('EXT_disjoint_timer_query');;
     }
 
     setDefault() {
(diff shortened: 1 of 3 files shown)


src/ui/map.js
```diff
@@ -1987,6 +1996,33 @@ class Map extends Camera {
...
+            const renderCPUTime = browser.now() - frameStartTime;
+            extTimerQuery.endQueryEXT(extTimerQuery.TIME_ELAPSED_EXT, gpuTimer);
+            setTimeout(() => {
+                const renderGPUTime = extTimerQuery.getQueryObjectEXT(gpuTimer, extTimerQuery.QUERY_RESULT_EXT) / (1000 * 1000);
+                extTimerQuery.deleteQueryEXT(gpuTimer);
+                this.fire(new Event('gpu-timing-frame', {
+                    cpuTime: renderCPUTime,
+                    gpuTime: renderGPUTime
+                }));
+            }, 50); // Wait 50ms to give time for all GPU calls to finish before querying
```
mourner: It's not clear how the threshold here works — is 50ms an arbitrary value that means "long enough to be sure"? Is this because `endQueryExt` happens async too, without any way to track when it executes?

src/gl/context.js
```diff
@@ -113,6 +114,8 @@ class Context {
         if (this.extTextureHalfFloat) {
             gl.getExtension('OES_texture_half_float_linear');
         }
+
+        this.extTimerQuery = gl.getExtension('EXT_disjoint_timer_query');
```
astojilj: @ansis, do we have a matrix of browsers where this is supported? Related to https://www.chromium.org/Home/chromium-security/ssca

asheemmamoowala: @ansis What is the current behavior on non-Chrome browsers? Do the event handlers ever get called? Are there any errors thrown?

astojilj: Yes, it is re-enabled after Chrome 70 on Desktop only: https://chromium.googlesource.com/chromium/src.git/+/cc1732e0386125c0c3aa04690663bfb43e6d7f68

I didn't see that it's precision is reduced, as it is done for performance.now (https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/timing/performance.cc?sq=package:chromium&dr=CSs&g=0&l=893).

PR #8830, opened 2019-10-03
get/set cameraPosition with altitude

andrewharvey:
## Launch Checklist

resurrecting #6093 to explore the possibility of exposing getting and setting the camera real world elevation, only a WIP at the moment.

addresses #3552

adds internal transform methods for get/setCameraPosition from a lng, lat, **altitude** (in meters), pitch and bearing as an alternative to setting a zoom.

This is useful for ensuring setting the view to an extrusion or 3D feature is set at the right height (altitude, zoom) that the top is visible.

An alternative would be be a zoomFromAltitude method which would just return the equivalent zoom for an altitude which you could then use for the normal view methods.

 - [ ] post benchmark scores
 - [ ] ~~tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec or visual changes~~
 - [ ] ~~tagged `@mapbox/gl-native` if this PR includes shader changes or needs a native port~~

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -381,8 +381,7 @@ class Transform {
         const pitch = this._pitch;
         const altitude = Math.cos(pitch) * this.cameraToCenterDistance;
         const latOffset = Math.tan(pitch) * this.cameraToCenterDistance;
-        const latPosPointInPixels = this.centerPoint.add(new Point(0, latOffset));
-        const lngLat = this.pointLocation(latPosPointInPixels);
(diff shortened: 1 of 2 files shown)


asheemmamoowala:
cc @elifitch

skreborn:
@andrewharvey FYI, some of the calculations you make are already available in [mercator_coordinate.js](https://github.com/mapbox/mapbox-gl-js/blob/master/src/geo/mercator_coordinate.js#L9).

ryanhamley:
Closing this as stale. It also seems likely to conflict with some future camera changes.

Issue #3731, opened 2016-12-01
Support max pitch > 60


lucaswoj:
As we improve support for custom rendering, there are legitimate cases for pitching the map beyond the hardcoded limit of 60 degrees. With the proper safeguards to prevent loading an absurd # of tiles, we could make the max pitch of the map configurable.

mourner:
> With the proper safeguards to prevent loading an absurd # of tiles

Yes! We've discussed some ideas about this with @ansis recently. Basically we should change the tile loading algorithm so that it loads lower zoom tiles on farther side, depending on how small they appear when projected.

ibgreen:
| loads lower zoom tiles on farther side, depending on how small they appear when projected. 

That would be really neat. If not, just rendering blur / some sort of background color in the distance would work fine as well - at least it would unblock our use cases.

ibgreen:
Also, in deck.gl we are combining maps with general 3D geometry - so our main camera is a general projection matrix, which we currently have to constrain to fit with mapbox zoom and pitch limits. If we remove those constraints, this means that we could have pitch > 90 degrees!

It would certainly be OK not to render the base map in this case (we won't be underground). But we might still want to see extrusions (like buildings) from your 3D styles when looking up from street level.

Maybe this should be a separate issue?

ansis:
besides level-of-detail tile loading we'll also need to fix:
- label-placement (the tiled placement approach will completely fall apart)
- antialiasing

ibgreen:
> label-placement (the tiled placement approach will completely fall apart)

For our purposes it would be OK if the labels disappeared at high pitch levels. They probably wouldn't be very legible anyway?

Mariusun:
This would be really great! Any progress on this issue?

ansis:
@Mariusun not on the main part of the issue but we are making progress one some sub-problems like improved label placement

yuechenglei:
I am new try to use mapboxGL. It's wonderful to change the slope. But I also have the need to pitch the map more to observe the 3D
The label don't need to slope too much I think...
Hope to have progress on this issue.

damien-murphy:
+1 This would also be useful for AR/VR Experiences for FPV and also when turning on 3D Buildings to look down streets

1ec5:
Related: mapbox/mapbox-gl-native#9037.

areknawo:
Any progress by now?

areknawo:
[...]

PR #8834, opened 2019-10-04
min and max pitch options

andrewharvey:
## Launch Checklist

add get/set min/maxPitch map options. This provides some groundwork for #3731, but doesn't change the maximum pitch you can set (still 60).

### DIFF
--- a/src/geo/transform.js
+++ b/src/geo/transform.js
@@ -46,18 +46,23 @@ class Transform {
     _renderWorldCopies: boolean;
     _minZoom: number;
     _maxZoom: number;
+    _minPitch: number;
+    _maxPitch: number;
     _center: LngLat;
     _constraining: boolean;
     _posMatrixCache: {[number]: Float32Array};
     _alignedPosMatrixCache: {[number]: Float32Array};
 
-    constructor(minZoom: ?number, maxZoom: ?number, renderWorldCopies: boolean | void) {
+    constructor(minZoom: ?number, maxZoom: ?number, minPitch: ?number, maxPitch: ?number, renderWorldCopies: boolean | void) {
         this.tileSize = 512; // constant
         this.maxValidLatitude = 85.051129; // constant
 
         this._renderWorldCopies = renderWorldCopies === undefined ? true : renderWorldCopies;
         this._minZoom = minZoom || 0;
         this._maxZoom = maxZoom || 22;
+
+        this._minPitch = minPitch || 0;
+        this._maxPitch = maxPitch || 60;
 
         this.setMaxBounds();
 
@@ -74,7 +79,7 @@ class Transform {
     }
 
     clone(): Transform {
-        const clone = new Transform(this._minZoom, this._maxZoom, this._renderWorldCopies);
+        const clone = new Transform(this._minZoom, this._maxZoom, this._minPitch, this.maxPitch, this._renderWorldCopies);
         clone.tileSize = this.tileSize;
         clone.latRange = this.latRange;
         clone.width = this.width;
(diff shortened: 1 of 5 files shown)


astojilj:
I'm not sure if it is overlapping with your patch: there's ongoing work on implementing [this](https://github.com/mapbox/mapbox-gl-native/pull/15230) in gl-js.
/cc @arindam1993

andrewharvey:
> I'm not sure if it is overlapping with your patch: there's ongoing work on implementing this in gl-js.

Oh I didn't see that. This PR doesn't do any of the fancy LOD features, it just adds get/set min/max zoom.

@arindam1993 if your work supersedes this, no worries, or if you want to build upon this that's fine. This could still be a nice to have feature even without the LOD (although it would become even more valuable with LOD and a higher maxzoom available).

Should I go ahead and add tests or leave as is until superseded by your work?

astojilj:
@andrewharvey 
#3731 is about supporting pitch larger than 60.
As mentioned, there's ongoing work on implementing larger value than 60 and that value will likely to get limited by the rendering engine itself - I see no need in providing additional control to user to increase the limit over the value supported by engine.
Is the use case here about specifying the subrange (of the range supported mapbox-gl-js) by the application developer - so that the application would allow pitch in e.g. range 0-30 degrees?

andrewharvey:
> As mentioned, there's ongoing work on implementing larger value than 60 and that value will likely get limited by the rendering engine itself

Sure, if possible it would be nice to have an update on #3731 about this, or just having that ticket assigned to someone at Mapbox to show it's currently being worked on.

> I see no need in providing additional control to user to increase the limit over the value supported by engine.

I agree, this PR specifically only lets you configure a min/max pitch within the bounds of what's supported. So right now 0-60.

> Is the use case here about specifying the subrange (of the range supported mapbox-gl-js) by the application developer - so that the application would allow pitch in e.g. range 0-30 degrees?

Yes that's exactly what this implements.

src/geo/transform.js
```diff
@@ -46,19 +46,24 @@ class Transform {
...
+    constructor(minZoom: ?number, maxZoom: ?number, minPitch: ?number, maxPitch: ?number, renderWorldCopies: boolean | void) {
         this.tileSize = 512; // constant
         this.maxValidLatitude = 85.051129; // constant
 
         this._renderWorldCopies = renderWorldCopies === undefined ? true : renderWorldCopies;
         this._minZoom = minZoom || 0;
         this._maxZoom = maxZoom || 22;
 
+        this._minPitch = (minPitch === undefined || minPitch === null) ? 0 : minPitch;
+        this._maxPitch = (maxPitch === undefined || maxPitch === null) ? 60 : maxPitch;
```
andrewharvey: I can't simply use `maxPitch || 60` because it will not accept `maxPitch: 0`.
I need to check both undefined and null because flow is picky.

src/ui/map.js
```diff
@@ -99,6 +101,11 @@ type MapOptions = {
 
 const defaultMinZoom = 0;
 const defaultMaxZoom = 22;
+
+// the default values, but also the valid range
+const defaultMinPitch = 0;
+const defaultMaxPitch = 60;
```
andrewharvey: Instead of defining both the default and limits, since the defaults are going to be the limits, I've used this constant for both the default and the limits.

src/ui/map.js
```diff
@@ -326,10 +338,22 @@ class Map extends Camera {
         options = extend({}, defaultOptions, options);
 
         if (options.minZoom != null && options.maxZoom != null && options.minZoom > options.maxZoom) {
-            throw new Error(`maxZoom must be greater than minZoom`);
+            throw new Error(`maxZoom must be greater than or equal to minZoom`);
```
andrewharvey: I could split this out to a separate PR if needed as it is unrelated.

test/unit/ui/map.test.js
```diff
@@ -807,14 +807,103 @@ test('Map', (t) => {
     t.test('throw on maxZoom smaller than minZoom at init', (t) => {
         t.throws(() => {
             createMap(t, {minZoom:10, maxZoom:5});
-        }, new Error(`maxZoom must be greater than minZoom`));
+        }, new Error(`maxZoom must be greater than or equal to minZoom`));
```
andrewharvey: part of the unrelated change earlier

andrewharvey:
I've fixed the failing tests. This does introduce a breaking change to the private Transform API, but since it's private and I've updated all internal use, is that an issue? To make it non-breaking I could change the order to just take on the new parameters to the end to avoid the breaking change.

Issue #8034, opened 2019-03-14
Give Geolocation Control Marker Pitch Alignment Properties


JohnCLo:
Currently when the geolocation control marker is added to the map, the blue marker has no options to change color or set pitch such as "circle-pitch-alignment": "map". For aesthetic purposes, it may be useful for the marker to have more paint options that are exposed to a normal circle type layer.

Current:
<img width="399" alt="Screen Shot 2019-03-13 at 11 21 56 PM" src="https://user-images.githubusercontent.com/1709234/54331227-05e27580-45e7-11e9-901c-b73c93a79229.png">

andrewharvey:
The dot is implemented as a Marker (https://github.com/mapbox/mapbox-gl-js/blob/c60597408379e2bf196118b0466e017806d3b958/src/ui/control/geolocate_control.js#L284) so you can't use the normal pitch-alignment capabilities of using a Circle from the style spec, this would need to be implemented by watching the pitch change and updating the CSS of the Marker.

Originally the dot was done as a Circle, but due to issues with injecting this into the users style, we made it a Marker.

mourner:
Fixed by #10817

PR #8836, opened 2019-10-04
Add bearing to markers

### LINKED ISSUE #3937: feature request: adding {"rotation-alignment": "map"} option to mapboxgl.Marker()
Hello.
For drawing arrowhead markers, I believe that mapboxgl.Marker() should have {"rotation-alignment": "map"} option. See the following examples:

https://gist.github.com/kkdd/928c67088c2debae4204bef55c170106

apboxgl marker

dburnsii:
**Description:**
This PR adds a parameter to a Marker that enables a bearing to be set. This keeps the marker flat on the map surface, and allowing the marker to point in any direction. This replaces a previous workaround that involved creating an additional layer on top of an existing layer.

**Before/After:**
SpinningMarkerTop
ChangingPitch
Rotating
Panning

**Changes to API**
New option for marker: `bearing`

New functions for marker: `marker.getBearing()` & `marker.setBearing(bearing)`

closes #3937


ryanhamley:
Can you say more about the use case for this? I'm specifically curious why setting a marker bearing should force it to be flat on the surface of the map.

dburnsii:
So the use case is for situations where we want whatever image we use to lay flat on the map, and potentially want it to move around (i.e. a location marker, updating with GPS). The method until now has been to create another layer with the image and move it, but that method has been shown to be sluggish and is overkill in most cases. This keeps the marker as a div and simplifies that functionality.

andrewharvey:
Have you thought about keeping this consistent with [`symbol`](https://docs.mapbox.com/mapbox-gl-js/style-spec/#layers-symbol), having separate [`rotation-alignment`](https://docs.mapbox.com/mapbox-gl-js/style-spec/#layout-symbol-icon-rotation-alignment) and [`pitch-alignment`](https://docs.mapbox.com/mapbox-gl-js/style-spec/#layout-symbol-icon-pitch-alignment) properties?

The latter would unlock #8034.

ryanhamley:
> Have you thought about keeping this consistent with symbol, having separate rotation-alignment and pitch-alignment properties?

Yeah, this is what I was about to propose as well. I don't think the pitch of the marker needs to be tied to the bearing. I like the general idea here though.

andrewharvey:
Here's the original issue #3937 for `rotation-alignment` for `Marker.

dburnsii:
Writing this I did feel like the pitch and the bearing could be added separately, but with both available it provides the functionality I need. I will update this to be consistent with the functionality in `symbol`.

dburnsii:
A core piece of this is to have precise control of the direction the marker is pointing. It doesn't look like a `symbol` has the ability to actually define the bearing. What would be a good way to include this while still being compatible with the `symbol` functionality?

andrewharvey:
> A core piece of this is to have precise control of the direction the marker is pointing. It doesn't look like a symbol has the ability to actually define the bearing.

https://docs.mapbox.com/mapbox-gl-js/example/animate-point-along-route/ shows `icon-rotate`.

ryanhamley:
Yep, I think it's reasonable to implement `marker-rotate`, `marker-rotation-alignment` and `marker-pitch-alignment`

andrewharvey:
Mind you the specific combination of:

```
"icon-rotation-alignment": 'map',
"icon-pitch-alignment": 'viewport'
```

, I can't see the use for, but all the other combinations I can understand and I think would be great for Marker to support.

You can tweak the example at https://docs.mapbox.com/mapbox-gl-js/example/add-image/ to see the affects of changing those properties for Symbol.

dburnsii:
Made the updates, the options are now `rotate`, `rotationAlignment`, and `pitchAlignment` when creating a marker, and setters and getters modify these values after object creation.

src/ui/marker.js
```diff
@@ -31,6 +34,9 @@ type Options = {
  * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
  * @param {boolean} [options.draggable=false] A boolean indicating whether or not a marker is able to be dragged to a new position on the map.
+ * @param {number} [options.rotate=0]
```
andrewharvey: Could you include a description on what setting this does, and document which units this number is in.

src/ui/marker.js
```diff
@@ -484,4 +510,61 @@ export default class Marker extends Evented {
     isDraggable() {
         return this._draggable;
     }
+
+    /**
+     * Sets the `rotate` property for pointing the marker in a particular direction
+     * @param {number} [newRotate=0] Sets the direction the marker is pointing
+     * @returns {Marker} `this`
+     */
+    setRotate(newRotate: number) {
```
andrewharvey: for consistency with the rest of the codebase this can be simply `rotate` and instead of `newRotate`.

src/ui/marker.js
```diff
@@ -19,7 +19,10 @@ type Options = {
     offset?: PointLike,
     anchor?: Anchor,
     color?: string,
-    draggable?: boolean
+    draggable?: boolean,
+    rotate?: number,
```
andrewharvey: rotate is fine, but I wonder if `rotation` is clearer?

dburnsii: Agreed, the variable has been changed to `rotation` throughout

src/ui/marker.js
dburnsii: Changed all the getters and setters

src/ui/marker.js
dburnsii: Added a short description to each option

andrewharvey:
@dburnsii I tested this out and it's awesome! Supporting rotationAlignment and pitchAlignment makes this much more flexible, opening up further use cases, whist being consistent with `Symbol`.

Just a couple more things I think this PR would need to address:

- [ ] unit tests
- [ ] more documentation in the JSDoc documenting the `map` and `viewport` options. Although we're implementing these options the same as the symbol options from the Style Spec, they are independent so I think they should be documented on their own.

src/ui/marker.js
```diff
@@ -31,6 +34,9 @@ type Options = {
  * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
  * @param {boolean} [options.draggable=false] A boolean indicating whether or not a marker is able to be dragged to a new position on the map.
+ * @param {number} [options.rotatation=0] The rotation angle of the marker (in degrees), relative to it's respective {@link Marker#rotationAlignment} setting.
```
andrewharvey: suggestion: "(clockwise, in degrees)"

andrewharvey: nit: `its` - no apostrophe? (https://theoatmeal.com/comics/apostrophe)

src/ui/marker.js
```diff
@@ -31,6 +34,9 @@ type Options = {
...
+ * @param {string} [options.pitchAlignment='auto'] The pitch alignment for the marker, allowing it to stay flat on the map or to stay vertical relative to the viewport.
+ * @param {string} [options.rotationAlignment='auto'] The rotation alignment for the marker, allowing for rotations relative to either the viewport or the map.
```
andrewharvey: I think this needs to make it clear that the options are `map` and `viewport`, how about this:

    `map` aligns the `Marker`'s rotation relative to the map, `viewport` aligns the `Marker`'s rotation relative to viewport.

src/ui/marker.js
vulkd: `rotatation`> `rotation` ?

dburnsii: Both resolved

src/ui/marker.js
```diff
@@ -31,6 +34,9 @@ type Options = {
  * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
  * @param {boolean} [options.draggable=false] A boolean indicating whether or not a marker is able to be dragged to a new position on the map.
+ * @param {number} [options.rotatation=0] The rotation angle of the marker (in degrees), relative to it's respective {@link Marker#rotationAlignment} setting.
+ * @param {string} [options.pitchAlignment='auto'] The pitch alignment for the marker, allowing it to stay flat on the map or to stay vertical relative to the viewport.
```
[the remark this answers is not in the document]
dburnsii: That is correct, auto essentially maintains compatibility with the way Markers behaved before I added this functionality. Leaving it as the default shouldn't break any existing setups.

src/ui/marker.js
dburnsii: Used this for both `pitchAlignment` and `rotationAlignment`

src/ui/marker.js
dburnsii: Fixed

index.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
```
andrewharvey: Did you commit this by accident? The debug pages are in `debug/` we could probably add this functionality into the existing `debug/markers.html` page or if you don't want to you could leave it out for a future PR.

andrewharvey:
Thanks for the unit tests, apart from the index.html which should either be part of the debug pages or removed, this PR is great!

src/ui/marker.js
```diff
@@ -484,4 +510,61 @@ export default class Marker extends Evented {
     isDraggable() {
         return this._draggable;
     }
+
+    /**
+     * Sets the `rotation` property for pointing the marker in a particular direction
```
ryanhamley: I think the wording on this comment isn't quite right. If the marker is set relative to the viewport, then it's not really correct to say it's pointing in a particular "direction".  There's also no indication in the description of the method or the `rotation` parameter that `rotation` is in degrees or what positive and negative degrees would do (clockwise vs counter-clockwise).

src/ui/marker.js
```diff
@@ -31,6 +34,9 @@ type Options = {
  * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
  * @param {boolean} [options.draggable=false] A boolean indicating whether or not a marker is able to be dragged to a new position on the map.
+ * @param {number} [options.rotation=0] The rotation angle of the marker (clockwise, in degrees), relative to its respective {@link Marker#rotationAlignment} setting.
```
ryanhamley: As far as I can tell, nothing in this PR restricts rotation to the clockwise direction (which I think is good!). If you set `rotation = -45`, then the marker will rotate 45 degrees counterclockwise, right? So we should spell that out here.

src/ui/marker.js
```diff
@@ -484,4 +510,61 @@ export default class Marker extends Evented {
...
+     * @returns {Marker} `this`
+     */
+    setRotation(rotation: number) {
+        this._rotation = rotation || 0;
+        this._update();
+        return this;
+    }
+
+    /**
+     * Returns the current rotation angle
```
ryanhamley: `in degrees`

src/ui/marker.js
```diff
@@ -484,4 +510,61 @@ export default class Marker extends Evented {
...
+     * @returns {number}
+     */
+    getRotation() {
+        return this._rotation;
+    }
+
+    /**
+     * Sets the `rotateAlignment` property for pointing the marker in a particular direction
```
ryanhamley: reference to direction here as well. I'm not sure what the best way to say this is, but I think this isn't quite right. @andrewharvey thoughts?

src/ui/marker.js
```diff
@@ -484,4 +510,61 @@ export default class Marker extends Evented {
...
+        return this._rotationAlignment;
+    }
+
+    /**
+     * Sets the `pitchAlignment` property for aligning the pitch of a marker with the viewport or the map
+     * @param {string} [alignment='auto'] Sets the pitch alignment property
+     * @returns {Marker} `this`
+     */
+    setPitchAlignment(alignment: string) {
+        this._pitchAlignment = alignment || 'auto';
```
ryanhamley: This would also need to change to match `rotateAlignment` when `alignment` is `auto`

src/ui/marker.js
andrewharvey: I would just use "Sets the `rotationAlignment` of the marker." Further documentation of exactly what rotationAlignment is is further up in the options and I don't think needs to be duplicated here? At least this is what other options like draggable/setDraggable do. Open to other opinions though.

src/ui/marker.js
andrewharvey: Agreed this should be changed, but it goes back to if we should repeat the documentation for the property, or omit duplicating it and expect people to refer back to the `rotation` property documentation.

I think less duplication is better as you don't need to re-read the same documentation in multiple places and ensure it's consistent.

` * @param {number} [options.rotation=0] The rotation angle of the marker (clockwise, in degrees), relative to its respective {@link Marker#rotationAlignment} setting.`

src/ui/marker.js
andrewharvey: Maybe just enough to replace "clockwise, in degrees" with "a positive rotation angle is clockwise, units in degrees"?

src/ui/marker.js
ryanhamley: I like that approach and I totally missed that this needs to be `rotationAlignment`

src/ui/marker.js
ryanhamley: Maybe `The rotation angle of the marker in degrees, relative to its respective {@link Marker#rotationAlignment} setting. A positive value will rotate the marker clockwise.`

src/ui/marker.js
dburnsii: Updated to `alignment || this._rotationAlignment`

src/ui/marker.js
dburnsii: Used that wording for `setPitchAlignment` and `setRotationAlignment`, and fixed the `rotateAlignment` typo.

src/ui/marker.js
dburnsii: Fixed.

src/ui/marker.js
dburnsii: Updated the parameter description.

src/ui/marker.js
dburnsii: Updated to "Sets the `rotationAlignment` of the marker."

src/ui/marker.js
```diff
@@ -72,6 +81,9 @@ export default class Marker extends Evented {
         this._color = options && options.color || '#3FB1CE';
         this._draggable = options && options.draggable || false;
         this._state = 'inactive';
+        this._rotation = options && options.rotation || 0;
+        this._pitchAlignment = options && options.pitchAlignment || 'auto';
```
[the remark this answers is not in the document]
dburnsii: Set to `options && options.pitchAlignment || this._rotationAlignment`, putting it after `this._rotationAlignment` is set.

src/ui/marker.js
```diff
@@ -31,6 +34,9 @@ type Options = {
  * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
  * @param {boolean} [options.draggable=false] A boolean indicating whether or not a marker is able to be dragged to a new position on the map.
+ * @param {number} [options.rotation=0] The rotation angle of the marker (clockwise, in degrees), relative to its respective {@link Marker#rotationAlignment} setting.
+ * @param {string} [options.pitchAlignment='auto'] `map` aligns the `Marker`'s pitch relative to the map, `viewport` aligns the `Marker`'s pitch relative to the viewport.
+ * @param {string} [options.rotationAlignment='auto'] `map` aligns the `Marker`'s rotation relative to the map, `viewport` aligns the `Marker`'s rotation relative to the viewport.
```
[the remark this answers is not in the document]
dburnsii: I used your wording for `pitchAlignment`, but there are extra properties that make the `icon-rotation-alignment`description as easy to drop in. Changed that to: 

"`map` aligns the `Marker`'s rotation relative to the map, maintaining a bearing as the map rotates. `viewport` aligns the `Marker`'s rotation relative to the viewport, agnostic to map rotations. `auto` is equivalent to `viewport`."  

Thoughts?

src/ui/marker.js
```diff
@@ -72,6 +81,9 @@ export default class Marker extends Evented {
         this._color = options && options.color || '#3FB1CE';
         this._draggable = options && options.draggable || false;
         this._state = 'inactive';
+        this._rotation = options && options.rotation || 0;
+        this._rotationAlignment = options && options.rotationAlignment || 'auto';
+        this._pitchAlignment = options && options.pitchAlignment || this._rotationAlignment;
```
ryanhamley: This wouldn't match `pitchAlignment` to `rotationAlignment` if `pitchAlignment = 'auto'` and `rotationAlignment = 'map'`. This should be `options && options.pitchAlignment && options.pitchAlignment === 'auto' ? this._rotationAlignment : options.pitchAlignment`

src/ui/marker.js
```diff
@@ -484,4 +510,61 @@ export default class Marker extends Evented {
...
+        return this._rotationAlignment;
+    }
+
+    /**
+     * Sets the `pitchAlignment` property of the marker.
+     * @param {string} [alignment='auto'] Sets the `pitchAlignment` property of the marker.
+     * @returns {Marker} `this`
+     */
+    setPitchAlignment(alignment: string) {
+        this._pitchAlignment = alignment || this._rotationAlignment;
```
ryanhamley: I think this should also be `alignment === 'auto' ? this._rotationAlignment : alignment`.

src/ui/marker.js
```diff
@@ -484,4 +510,61 @@ export default class Marker extends Evented {
...
+     * Returns the current `rotationAlignment` property of the marker.
+     * @returns {string}
+     */
+    getRotationAlignment() {
+        return this._rotationAlignment;
+    }
+
+    /**
+     * Sets the `pitchAlignment` property of the marker.
+     * @param {string} [alignment='auto'] Sets the `pitchAlignment` property of the marker.
```
ryanhamley: This doesn't actually default to auto. I think we should  remove the default here and make a note that `if alignment is 'auto', it will automatically match rotationAlignment`.

test/unit/ui/marker.test.js
```diff
@@ -465,3 +465,69 @@ test('Marker with draggable:true does not error if removed on mousedown', (t) =>
     t.ok(map.fire('mouseup'));
     t.end();
 });
+
+test('Marker can set rotationAlignment and pitchAlignment', (t) => {
```
ryanhamley: We should probably add a test that ensures setting `rotationAlignment === 'map'` and `pitchAlignment === 'auto'` results in `pitchAlignment === 'map'`. Both through the options object and by calling `set{Rotation/Pitch}Alignment`

dburnsii: Added a couple tests. One is for the option on initialization, the other is for the getter/setters.

src/ui/marker.js
dburnsii: Changed the wording on that.

src/ui/marker.js
dburnsii: Did this:

`this._pitchAlignment = alignment && alignment !== 'auto' ? alignment : this._rotationAlignment;`

To also check if it's undefined.

src/ui/marker.js
dburnsii: Fixed.

PR #8845, opened 2019-10-09
ImageBitmap and OffscreenCanvas for DEM tiles

### LINKED ISSUE #7555: Use ImageBitmap API in Chrome and Firefox to decode images

arindam1993:
This partially closes #7555, and builds on-top of #7592 .

This PR attempts to move the decoding of the DEM image off the main thread and into the worker by using a combination of `ImageBitmap` and `OffscreenCanvas`. It is implemented as a progressive enhancement  with fallback onto our existing pipeline of `ArrayBuffer -> BlobURL -> Image -> Canvas.getImageData`. 
The main bottleneck in the existing pipeline is `getImageData` on the main thread, with the time  for that call varying anywhere between 3-30 ms per tile. This PR moves that source of jank off the main thread by transferring the `ImageBitmap` directly to the worker thread and  using `OffscreenCanvas.getImageData` to do decode the image off the main thread and transferring the Arraybuffer back to the main thread.

## Performance Improvements:
 - Time to first `idle` event, this measures the amount of time taken for all the dem tiles in the initial viewport to load.
     `master`: 1300ms
     `this branch`: 1000ms
 - FPS measured by a quick zoom out on a style thats only includes the hillshade layer
     `master`: 
Screen Shot 2019-12-02 at 5 21 30 PM
     `this branch`: 
Screen Shot 2019-12-02 at 5 28 40 PM

There seem to be fewer framerate dips overall.

## Browser Testing Checklist:
 - [x] Chrome OSX 
 - [x] Chrome Windows
 - [x] Chrome Android
 - [x] Firefox OSX (fallsback)
 - [x] Firefox Windows (fallsback)
 - [x] Safari OSX (fallsback)
 - [x] Safari OSX ( with OffscreenCanvas and ImageBitmap enabled in the experimental features)
 - [x] Safari iPadOS (fallsback)
 - [x] Safari iOS ( thanks @ryanhamley  ) (fallsback)
 - [x] Firefox Windows (fallsback)
 - [x] IE11 (fallsback)

This branch:

Master:

## Benchmarks
Wasn't able to run comparative benchmarks coz of #9078 , but I was able to replicate the performance improvement in a hillshade loadtime benchmark.
Screen Shot 2019-12-05 at 12 51 46 PM

Screen Shot 2019-12-05 at 12 58 48 PM

## Launch Checklist


src/util/web_worker_transfer.js
```diff
@@ -223,8 +223,12 @@ export function deserialize(input: Serialized): mixed {
         input instanceof String ||
         input instanceof Date ||
         input instanceof RegExp ||
+<<<<<<< HEAD
         input instanceof ArrayBuffer ||
         input instanceof ImageBitmap ||
+=======
+        isArrayBuffer(input) ||
+>>>>>>> 8e8a28025dca26e258ad6a26c064d07596f11527
```
kkaefer: merge failure

src/source/raster_dem_tile_worker_source.js
```diff
@@ -1,31 +1,52 @@
...
+        const dem = new DEMData(uid, imagePixels, encoding);
         this.loaded = this.loaded || {};
         this.loaded[uid] = dem;
         callback(null, dem);
     }
 
+    getImageData(imgBitmap: ImageBitmap): RGBAImage {
+        // Lazily initialize OffscreenCanvas
+        if (!this.offcreenCanvas || !this.offcreenCanvasContext) {
+            this.offcreenCanvas = new OffscreenCanvas(512, 512);
```
kkaefer: Why does 512&times;512 suffice?

src/source/raster_dem_tile_worker_source.js
```diff
@@ -1,31 +1,52 @@
...
+            this.offcreenCanvasContext = this.offcreenCanvas.getContext('2d');
+        }
+
+        this.offcreenCanvas.width = imgBitmap.width;
+        this.offcreenCanvas.height = imgBitmap.height;
+        this.offcreenCanvasContext.drawImage(imgBitmap, 0, 0, imgBitmap.width, imgBitmap.height);
```
kkaefer: When reusing the OffscreenCanvas for decoding images that are smaller, the right/bottom backfill border might contain leftovers from the previously decoded image.

src/source/raster_dem_tile_worker_source.js
```diff
@@ -1,31 +1,52 @@
...
+        const imagePixels = (ImageBitmap && rawImageData instanceof ImageBitmap) ? this.getImageData(rawImageData) : rawImageData;
+        const dem = new DEMData(uid, imagePixels, encoding);
         this.loaded = this.loaded || {};
         this.loaded[uid] = dem;
         callback(null, dem);
     }
 
+    getImageData(imgBitmap: ImageBitmap): RGBAImage {
+        // Lazily initialize OffscreenCanvas
+        if (!this.offcreenCanvas || !this.offcreenCanvasContext) {
```
kkaefer: typo in variable name

src/util/offscreen_canvas_supported.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import window from './window';
+
+let supportsOffscreenCanvas = null;
+
+export default function offscreenCanvasSupported(): boolean {
+    if (supportsOffscreenCanvas == null) {
```
kkaefer: nit: `===`

src/source/raster_dem_tile_worker_source.js
arindam1993: Oh good catch! Didn't think about this!

arindam1993: I added a clearRect call and re-ran the benchmarks, doesn't seem to have a performance impact.

src/source/raster_dem_tile_worker_source.js
arindam1993: 256x256 is probably better I had this initially because thats what i though our tile resolution was.

src/util/offscreen_canvas_supported.js
arindam1993: Intentional as a `null/undefined` check

src/util/ajax.js
```diff
@@ -257,14 +258,38 @@ function sameOrigin(url) {
...
+    (img: any).expires = expires;
+    img.src = data.byteLength ? URL.createObjectURL(blob) : transparentPngUrl;
+}
+
+function arrayBufferToImageBitmap(data: ArrayBuffer, callback: (err: ?Error, image: ?ImageBitmap) => void) {
+    const blob: Blob = new window.Blob([new Uint8Array(data)], {type: 'image/png'});
+    window.createImageBitmap(blob).then((imgBitmap) => {
+        callback(null, imgBitmap);
+    }).catch((e) => {
+        console.log(e);
```
ryanhamley: Did you mean to leave this in?

src/util/ajax.js
```diff
@@ -257,14 +258,38 @@ function sameOrigin(url) {
 
 const transparentPngUrl = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAAC0lEQVQYV2NgAAIAAAUAAarVyFEAAAAASUVORK5CYII=';
 
+function arrayBufferToImage(data: ArrayBuffer, callback: (err: ?Error, image: ?HTMLImageElement) => void, cacheControl: ?string, expires: ?string) {
+    const img: HTMLImageElement = new window.Image();
+    const URL = window.URL || window.webkitURL;
```
ryanhamley: Do we need the `|| window.webkitURL` here?

src/util/offscreen_canvas_supported.js
kkaefer: Then let's use

```suggestion
    if (!supportsOffscreenCanvas) {
```

src/source/raster_dem_tile_worker_source.js
```diff
@@ -1,31 +1,55 @@
...
         this.loaded = this.loaded || {};
         this.loaded[uid] = dem;
         callback(null, dem);
     }
 
+    getImageData(imgBitmap: ImageBitmap): RGBAImage {
+        // Lazily initialize OffscreenCanvas
+        if (!this.offscreenCanvas || !this.offscreenCanvasContext) {
+            // Dem tiles are typically 256x256
+            this.offscreenCanvas = new OffscreenCanvas(256, 256);
```
kkaefer:
```suggestion
            this.offscreenCanvas = new OffscreenCanvas(imgBitmap.width, imgBitmap.height);
```

They're resized below anyway.

src/util/offscreen_canvas_supported.js
arindam1993: But then that would run when `supportsOffscreenCanvas` is `false` as well.

kkaefer: In what situations would `supportsOffscreenCanvas` be `undefined` then? Let's add some flowtype to the variable to make it clearer what states it can be in.

arindam1993: It really wound't ever be `undefined` I was just sticking to the most easiest `maybe type` refinement implementation for flow( I guess i still forgot the annotation) https://flow.org/en/docs/types/maybe/#toc-refining-maybe-types (the 2nd example)

I added the `?boolean` annotation.

arindam1993:
YAY! 🎉

astojilj:
This is causing https://github.com/mapbox/mapbox-gl-js/issues/9100.

PR #8849, opened 2019-10-09
Patch release - v1.4.1

ryanhamley:


src/style/style_layer/circle_style_layer_properties.js
```diff
@@ -17,6 +17,8 @@ import type Color from '../../style-spec/util/color';
 
 import type Formatted from '../../style-spec/expression/types/formatted';
 
+import type ResolvedImage from '../../style-spec/expression/types/resolved_image';
```
mourner: Sorry for missing earlier PRs, just wanted to note most of the `ResolvedImage` imports in this PR are unused. We should probably catch things like this with ESLint.

PR #8850, opened 2019-10-09
Ensure tile mask geometry buffers are cleared when tile is unloaded

arindam1993:
Clear tile mask buffers when tiles are unloaded.
Cherrypick of the fix for the memory leak from #8814 . 

## Launch Checklist

 - [ ] ~include before/after visuals or gifs if this PR includes visual changes~
 - [ ] ~write tests for all new functionality~
 - [ ] ~post benchmark scores~
 - [ ] ~manually test the debug page~
 - [ ] ~tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec or visual changes~

### DIFF
--- a/src/render/draw_debug.js
+++ b/src/render/draw_debug.js
@@ -88,6 +88,10 @@ function drawDebugTile(painter, sourceCache, coord) {
     program.draw(context, gl.LINES, depthMode, stencilMode, colorMode, CullFaceMode.disabled,
         debugUniformValues(posMatrix, Color.black), id,
         debugTextBuffer, debugTextIndexBuffer, debugTextSegment);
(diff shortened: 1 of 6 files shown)


asheemmamoowala:
@arindam1993 What's the best way to verify these changes?

arindam1993:
> @arindam1993 What's the best way to verify these changes?

I used the instrumented gl calls similar to #8814, and did a manual check on if the object count so see if it flattened out. But did not include that in this PR. Should I move the modified test page over as well?

I think an extension to the `gl-stats` page with a longer running test that plots growth over time would be a more formal way to check for things like this.

PR #8856, opened 2019-10-11
hide internal mercatorScale function from api docs

andrewharvey:
## Launch Checklist

In #8524 we added a [mercatorScale](https://github.com/mapbox/mapbox-gl-js/blob/2dbbf634906dc1b02b48cb740dadb6de16348475/src/geo/mercator_coordinate.js#L43-L54) method, which was only intended to be used internally but it ended up making it into the API docs (also at the bottom of Events in the TOC).

Screenshot from 2019-10-11 10-14-42

Technically this is probably a breaking change, but to be honest I couldn't work out how to even use this method publicly so I don't think it's an issue to remove this from the public API.

I tested this change does indeed remove it from the API docs.

 - [ ] post benchmark scores

### DIFF
--- a/src/geo/mercator_coordinate.js
+++ b/src/geo/mercator_coordinate.js
@@ -48,6 +48,7 @@ export function altitudeFromMercatorZ(z: number, y: number) {
  *
  * @param {number} lat Latitude
  * @returns {number} scale factor
+ * @private
  */
 export function mercatorScale(lat: number) {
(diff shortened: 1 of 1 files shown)


andrewharvey:
the failing tests are unrelated to this change

mourner:
Wow, flaky tests failed in both render and unit tests — that's a jackpot!

PR #8864, opened 2019-10-11
Expose RTLTextPluginStatus

ryanhamley:
## Launch Checklist

fixes https://github.com/mapbox/mapbox-gl-js/issues/7869

    - this plugin exposes the RTLTextPluginStatus (which can be `unavailable`, `loading`, `loaded` or `error`) through a `mapboxgl.getRTLTextStatusPlugin` method

**CHANGELOG**
<changelog>Expose `getRTLTextPluginStatus` method</changelog>

### DIFF
--- a/src/source/rtl_text_plugin.js
+++ b/src/source/rtl_text_plugin.js
@@ -48,7 +48,6 @@ export const setRTLTextPlugin = function(url: string, callback: ErrorCallback) {
     }
     pluginStatus = status.loading;
     pluginRequested = true;
-
     pluginURL = browser.resolveURL(url);
     _completionCallback = (error?: Error) => {
         if (error) {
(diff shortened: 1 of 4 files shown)


ryanhamley:
@asheemmamoowala I removed `pluginRequested`

src/source/rtl_text_plugin.js
```diff
@@ -48,6 +59,7 @@ export const setRTLTextPlugin = function(url: string, callback: ErrorCallback) {
             foregroundLoadComplete = true;
         }
     };
+    pluginStatus = status.loaded;
```
arindam1993: I think this needs to be in the `_completionCallback`, as a replacement to `foregroundLoadComplete`. 

Also the `completionCallback` is invoked once per worker so we probably need to have counter that counts whether the state has been switched into `loaded` for all workers.

ryanhamley: I removed `foregroundLoadComplete` and moved the loaded status into the callback but it doesn't seem like we have to do anything else. The plugin doesn't throw an error.

Issue #8825, opened 2019-09-30
Load RTL Text plugin on demand


asheemmamoowala:
## Motivation
The RTL text plugin is quite large - 37kb (web assembly + js loader) and including it by default on a map increases the load time of the map, even when no RTL text is present. 

#6136 proposes emitting a console warning when RTL text is found on the map, but the plugin is not loaded. It would be better to go a step further and make it possible to load the plugin on-demand when such text is first encountered.

## Design Alternatives
There are two ways this could be done ,

1.  Emit an `rtlpluginmissing` event the first time RTL text is encountered and the plugin is not loaded or loading.
2.  Expose a `Map` option that can be passed into the container with the URL to the RTL Text plugin. The plugin would be automatically loaded by the map the first time RTL text is encountered.
3. Support right-to-left scripts without requiring plugin ([original ticket](https://github.com/mapbox/mapbox-gl-js/issues/4007))

## Design
Option 1 doesn't go far enough to simplify the API. All developers hooking up this event will likely be writing the same or similar code. It would be an incomplete solution the problem at hand.
Given the current size of the plugin, and the desire to decrease load times, Option 3 doesn't seem like it would address this need in the short term. Maybe if text placement and shaping was not done in JS, but took advantage of Browser APIs (`Canvas` or `OffscreenCanvas`). 

Option 2 provide a complete solution for developers interested in reducing initial map load time, and having RTL text show up correctly. Using a URL parameter allows providing a mapbox-hosted or self-hosted URL.

### Mock-Up

```
// Mapbox hosted plugin
var map = new Map ({... , rtlTextPlugin: "https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js"});

// OR self-hosted plugin
var map = new Map ({... , rtlTextPlugin: "https://my.dev.server/mapbox-gl-rtl-text.js"});
```

### Implementation
At present, there is not a way within GL-JS to detect if a string is arabic or hebrew, and requires r-t-l shaping and styling. 

If there is a known range of unicode characters that are supported for RTL, then Workers could message the main thread when such text is encountered. The main thread would load the plugin and re-load the source that triggered the issue.

cc @chloekraw @arindam1993 @kbauhausness

jingsam:
Would be better if just reference like this:

```html
[...]

Issue #8664, opened 2019-08-21
mapbox-gl-plot


JerckyLY:
**mapbox-gl-js version**:

### Question
Is there an article or demo about mapbox-plot?

mourner:
No idea what this was about.

JerckyLY:
It’s about plotting military operations on a map

JerckyLY:
> No idea what this was about.

It’s about plotting military operations on a map

JerckyLY:
> > No idea what this was about.
> 
> It’s about plotting military operations on a map
here:   https://ilocation.oschina.io/plot4ol3sample/

mourner:
Still no clue what this is about. The link above is an OpenLayers 3 map and has nothing to do with Mapbox.

PR #8865, opened 2019-10-12
Add option to load rtl-text-plugin lazily

arindam1993:
Addresses #8825.
~Adds an `rtlTextPluginURL` option to the map, and map will load the rtlTextPlugin the first time rtl text is encountered~

Adds a new `lazy` parameter to `setRTLTextPlugin(url: string, cb: Function, lazy: boolean)`. Its false by default to keep existing behavior, but when set to true, it will cause `mapboxgl` to lazily load the `rtl-text-plugin` the first time it encounters rtl-text, it also prevents the rtl-text from rendering until the plugin is downloaded and parsed after which it loads in.

This includes a few more enhancements,
- It will now make the network request for the plugin only once, and then distribute the downloaded data to all workers via a blobUrl
- the callback function is now invoked only once when the plugin is parsed and ready on `all` workers, instead of being repeatedly invoked for each worker.

Under the hood, theres a re-write of how the main-thread -> worker communication for the plugin workers. The main thread now has full ownership of the state of the plugin and pushes it to the workers. The workers are reactive and download and parse the plugin once all the data is available.

## Launch Checklist

 - [ ] ~include before/after visuals or gifs if this PR includes visual changes~
 - [ ] ~post benchmark scores~
 - [ ] ~tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec or visual changes~


arindam1993:
> What I am seeing when loading this URL: http://localhost:9966/debug/rtl-plugin-autoload.html#12.22/27.34633/45.00641, with throttled network or CPU performance (from Chrome dev tools), is that the Arabic text shows up left-to-right, and flashes before being re-rendered correctly.
> 
> Would it be possible to hold off that text from being rendered so that it fades in after the plugin is loaded?

Yes, the expected behaviour should be, all LTR text loads as expected, and RTL text loads appears after a slight delay once the plugin finishes downloading.

> Additionally, this API only works with a URL to a separately hosted rtl-text-plugin. This means that developers who package the plugin, or bundle GL-JS differently will not be able to load this plugin on-demand. (cc @kappaxbeta @patfrat)

I think this PR shouldn't try to rework the delivery process, but we can make the loading lazy and slightly more efficient.

> There's a fair bit of overlap between this PR and #8864. It may make sense to commit #8664 first and then rebase this on top of that so that it can take advantage of the `loading`/`requested` state. In my debugging, the plugin was being fetched twice, based on the Devtools network tab.
> cc @ansis

Yes, will wait for #8864 to be ready, the plugin being fetched multiple times is the current status-quo though, since each worker requests for the plugin separately. I'll change that in this PR so that the main thread downloads the plugin and then the workers load it via blob url.

src/data/bucket/symbol_bucket.js
```diff
@@ -413,7 +414,9 @@ class SymbolBucket implements Bucket {
                 if (formattedText.containsRTLText()) {
                     this.hasRTLText = true;
                 }
-                text = transformText(formattedText, layer, feature);
+                if (this.hasRTLText && globalRTLTextPlugin.isLoaded() || !this.hasRTLText) {
+                    text = transformText(formattedText, layer, feature);
```
ryanhamley: Should there be a warning emitted here so that customers don't just see a map with no labels and wonder what the heck is going on?

src/source/rtl_text_plugin.js
```diff
@@ -1,50 +1,73 @@
...
 import browser from '../util/browser';
+import window from '../util/window';
 
 const status = {
-    unavailable: 'unavailable',
-    loading: 'loading',
+    unavailable: 'unavailable', // Not loaded
+    available: 'available', // Host url specified, but we havent started loading yet
+    loading: 'loading', // request in-flight
+    downloaded: 'downloaded', //plugin loaded and cached on main-thread and pluginBlobURL for worker is generated
```
asheemmamoowala: What is the difference between `downloaded` and `loaded`, and why would we want to expose that distinction?

src/ui/map.js
```diff
@@ -278,6 +278,7 @@ class Map extends Camera {
     _mapId: number;
     _localIdeographFontFamily: string;
     _requestManager: RequestManager;
+    _rtlTextPluginURL: ?string;
```
asheemmamoowala: This is no longer needed right?

src/source/rtl_text_plugin.js
```diff
@@ -55,23 +78,73 @@ export const setRTLTextPlugin = function(url: string, callback: ErrorCallback) {
...
+    }
+    pluginStatus = status.loading;
+
+    if (pluginURL) {
+        getArrayBuffer({url: pluginURL}, (error, data) => {
+            if (error || !data) {
+                throw error;
+            } else {
+                const rtlBlob = new window.Blob([data], {type: 'application/javascript'});
+                const URL = window.URL || window.webkitURL;
```
ryanhamley: why do we need `webkitURL` here? i can't even really find much information on it but it seems to be deprecated

src/util/script_detection.js
```diff
@@ -306,6 +309,15 @@ export function charInSupportedScript(char: number, canRenderRTL: boolean) {
     return true;
 }
 
+export function stringContainsRTLText(chars: string): boolean {
+    for (const char of chars) {
+        if (charInRTLScript(char.charCodeAt(0))) {
```
ryanhamley: Is it possible that a string could contain RTL text if the first character isn't in one of the blocks in the `charInRTLScript` function? Do Hebrew and Arabic have anything like the mixture of Arabic numerals and Latin characters with Japanese characters?

ryanhamley: Re-reading this, I realize I just asked if Arabic mixes in Arabic numerals which....🤦‍♂but it does add to my question of whether or not this is a reliable detection method

arindam1993:
> I think overall this looks fine. It's a little tough to follow the code pathway but that's probably a natural consequence of having to pipe some of this information around the code base. My biggest concern is probably the various statuses (their naming and the states they represent).
> 
> Also, we need to make sure we have good documentation on the expected behavior of this plugin now.

I agree with this, I think I can change it to be a more uni-directional state-sync where the main-thread pushes the state of the plugin to the workers. The workers then just need to react to the correct state change and run `importScripts` when necessary.

I think that will let me get rid of the downloaded status as well.

src/data/bucket/symbol_bucket.js
arindam1993: I've changed the behavior now so that only when lazy loading, does it skip rendering the text.

src/source/rtl_text_plugin.js
arindam1993: Removed, I've changed the architecture as described in the main PR body.

debug/rtl-plugin-autoload.html
```diff
@@ -0,0 +1,32 @@
...
+    </style>
+</head>
+
+<body>
+<div id='map'></div>
+
+<script src='../dist/mapbox-gl-dev.js'></script>
+<script src='../debug/access_token_generated.js'></script>
+<script>
+mapboxgl.setRTLTextPlugin('https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js', null, true);
```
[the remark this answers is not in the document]
arindam1993: I've made the plugin loading interactive in the `rtl.html` page, so you can click a button and see the status live.

src/source/rtl_text_plugin.js
asheemmamoowala: This may have been added in the past when Safari/Webkit didn't support `window.URL`. But that doesn't seem to be an issue any longer based on [caniuse](https://caniuse.com/#search=createObjectURL)

Issue #8870, opened 2019-10-14
geolocation error handling not working


kilianso:
Geolocation error handling is not working. 

**mapbox-gl-js version**: 1.4.1

**browser**: Safari 13.0.2 on Mac OS Mojave

### Steps to Trigger Behavior

 1. Disable location service in system settings.
 2. Open example below and try to locate user by clicking the geolocation control.

### Link to Demonstration

[Locate User Example](https://docs.mapbox.com/mapbox-gl-js/example/locate-user/
)
### Expected Behavior
1. If user clicks on "do not allow", control should be visually updated to indicate that it is not possible to set the location.
2. If user has disabled overall location services in system settings of Mac OS, there needs to be a message or visual indicator that mapbox control will not work.

### Actual Behavior

1. If user clicks on "do not allow" – control appearance is not updated.
2. If user has disabled the overall location services in system settings of Mac OS – there is no notification, or any other indicator, that the mapbox control for geolocation is not working.

kkaefer:
@kilianso you can test the PR in https://github.com/mapbox/mapbox-gl-js/pull/8871

PR #8871, opened 2019-10-15
Show a disabled geolocate button if user or system denies geolocation

kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/8870

image

This PR adds a new icon to be shown when the user or the system denied permission to use the location. It's shown when geolocation is disabled outright in Firefox and Chrome, or when the user or the system explicitly denies location after being prompted.

There are a few caveats:
- It's not possible to detect whether the user has denied location to begin with when loading a page. Safari doesn't implement [navigator.permissions](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/permissions) detection. Once the user has click on the geolocate button, the state is set correctly though.
- When macOS location services are disabled _on a system-wide level_, Firefox never calls any of the callbacks, and the location indicator keeps spinning. This is a bug in Firefox and tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1529591
- When macOS location services are disabled, Chrome sneakily circumvents the OS setting and generates a location from different sources (e.g. IP address or Wifi networks; in my case, the location was approximate, but still in the same neighborhood).

Before this patch, the geolocate button was hidden when geolocation was disabled. It is now being shown, but in the new disabled state.

I'm a bit unsure how to test this correctly, since it depends on complex interaction between the OS, and the user clicking on various browser UIs.

/cc @andrewharvey


src/ui/control/geolocate_control.js
```diff
@@ -287,17 +288,17 @@ class GeolocateControl extends Evented {
...
-        }
         this._container.addEventListener('contextmenu', (e: MouseEvent) => e.preventDefault());
         this._geolocateButton = DOM.create('button',
             `${className}-icon ${className}-geolocate`,
             this._container);
         this._geolocateButton.type = 'button';
         this._geolocateButton.title = 'Find my location';
         this._geolocateButton.setAttribute('aria-label', 'Find my location');
+        if (supported === false) {
+            warnOnce('Geolocation support is not available, the GeolocateControl will be disabled.');
```
ryanhamley: Nit: I know this wasn't introduced here, but it bugs me. Can we either make this two sentences or make it `so the GeolocateControl...`?

ryanhamley:
>I'm a bit unsure how to test this correctly, since it depends on complex interaction between the OS, and the user clicking on various browser UIs.

Another argument for adding an end-to-end test suite

andrewharvey:
Could we update the title/aria-label when disabled to be something like "Location not available"?

andrewharvey:
This is good!

We should make sure to include a note in the changelog since the behaviour has changed. Anyone who prefers to hide the control when not available can still override the css `.mapboxgl-ctrl-geolocate:disabled { display: none }`.

kkaefer:
I tested in Safari in the following way:

```bash
brew install mkcert
mkcert -install
mkcert localhost

python -c "import BaseHTTPServer, SimpleHTTPServer, ssl; \
           handler = SimpleHTTPServer.SimpleHTTPRequestHandler; \
           httpd = BaseHTTPServer.HTTPServer(('localhost', 4443), handler); \
           httpd.socket = ssl.wrap_socket(httpd.socket, \
                                          certfile='./localhost.pem', \
                                          keyfile='./localhost-key.pem', \
                                          server_side=True); \
           httpd.serve_forever()"
```

Then accessed the server at `https://localhost:4443/debug/debug.html`

Issue #8842, opened 2019-10-08
rotate and pitch events are fired at the same time


cs09g:
When `pitch` and `rotate` events' added on the map, rotating the map triggers the `pitch` event fired.

**mapbox-gl-js version**:
1.4.0

**browser**:
Chrome 77.0.3865.90
### Steps to Trigger Behavior

 1. add `pitch` event on the map.
 2. add `rotate` event on the map.
 3. rotate the map.

### Link to Demonstration
https://jsfiddle.net/cs09g/9wb6unqt/

### Expected Behavior
`pitch` event should not be fired.

### Actual Behavior
`pitch` and `rotate` events are fired at the same time.

cs09g:
I'll try to fix this.

PR #8869, opened 2019-10-14
add condition for rotate and pitch events

cs09g:
This is a fix for #8842. It works as expected in a browser. But, unfortunately, tests couldn't make the map move so are failed.
Is this the right way? Any suggestions are welcome!

## Launch Checklist

 - [x] briefly describe the changes in this PR
 - [ ] write tests for all new functionality
 - [x] manually test the debug page

PR #8872, opened 2019-10-15
Fire 'rotate' and 'pitch' events only when their value changed

kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/8842, and supersedes https://github.com/mapbox/mapbox-gl-js/pull/8869 with a more concise implementation.

Also fixes a bug where the `pitch` event was still reading the old pitch value. Now, that value reflects the change that the user made.

### DIFF
--- a/test/unit/ui/handler/drag_rotate.test.js
+++ b/test/unit/ui/handler/drag_rotate.test.js
@@ -124,10 +124,37 @@ test('DragRotateHandler pitches in response to a right-click drag by default', (
     map.on('pitchend',   pitchend);
 
     simulate.mousedown(map.getCanvas(), {buttons: 2, button: 2});
-    simulate.mousemove(map.getCanvas(), {buttons: 2, clientX: 10, clientY: 10});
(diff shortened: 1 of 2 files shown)


test/unit/ui/handler/drag_rotate.test.js
```diff
@@ -151,7 +178,7 @@ test('DragRotateHandler pitches in response to a control-left-click drag', (t) =
     map.on('pitchend',   pitchend);
 
     simulate.mousedown(map.getCanvas(), {buttons: 1, button: 0, ctrlKey: true});
-    simulate.mousemove(map.getCanvas(), {buttons: 1,            ctrlKey: true, clientX: 10, clientY: 10});
+    simulate.mousemove(map.getCanvas(), {buttons: 1,            ctrlKey: true, clientX: 10, clientY: -10});
```
kkaefer: Moving downwards only rotates, and doesn't pitch. Change the value to move upwards so that the map gets pitched.

Issue #5784, opened 2017-12-01
Map navigation controls not visible in windows high contrast mode


PascalvanOol:
**mapbox-gl-js version**: latest mapbox sample page

### Steps to Trigger Behavior

1. Change your windows theme to high contrast #1
2. Open IE 11 browser
3. Open "Display map navigation controls" example:
https://www.mapbox.com/mapbox-gl-js/example/navigation/

### Expected Behavior
Navigation controls to show a icon in contrasting color.

### Actual Behavior
Navigation controls are shown all black.

jfirebaugh:
From my research, to fix this we need to avoid using `background-image`, which is ignored by IE11 in high-contrast modes. Instead, we should be using inline SVG with a rule such as:

```
@media screen and (-ms-high-contrast: active) {
  .button-icon {
    fill: ButtonText;
  }
}
```

https://css-tricks.com/accessible-svgs-high-contrast-mode/

kkaefer:
Fixed by #8874. Please see https://github.com/mapbox/mapbox-gl-js/pull/8874#issuecomment-548376940 for screenshots of how it looks in Windows High Contrast mode.

Issue #8767, opened 2019-09-17
Discussion: mapbox constrols


bravecow:
Hi. I have [some controls](https://github.com/bravecow/mapbox-gl-controls) for personal usage.
What I like about them:
1. [each control have separate directory with js + css + icons](https://github.com/bravecow/mapbox-gl-controls/tree/master/src/ruler)
mapbox now have 600-lines CSS
2. valid CSS / SCSS, mapbox CSS is not: https://github.com/mapbox/mapbox-gl-js/blob/master/src/css/mapbox-gl.css#L154
3. injected icons inside control button: https://github.com/bravecow/mapbox-gl-controls/blob/master/src/ruler/icon-ruler.js:
* no additional GET request for each resource
* easy for styling. we can style inline SVG with CSS.
* easy for sizing. for now customizing size looks like

As result:
* reduce JS bundle for mapbox
* removing `svg-load` plugin
* remove some GET requests for icons
* remove CSS prefixes, use [autoprefixer](https://github.com/postcss/autoprefixer).
* move controls from this repo to reduce package.json deps (if any)
* controls will not use any internal events or methods. it will help to understand real needs and provide better API for users (e.g. no public event for style load #8765 for now)
* move mapbox language, mapbox draw and other controls to this monorepo. they will have same bundling tools, codestyle, docs, etc. I think it will make future support easier
* braking change

asheemmamoowala:
@bravecow Thank you for starting this discussion, and sharing the gl-controls repo. 

We have recently had internal discussions about bundle size and plugin maintenance and are looking to incorporate some of that work into our roadmap for the rest of the year.

> reduce JS bundle for mapbox

Have you tried this? What sort of size reduction were you able to achieve by removing the controls. Only the Navigation, Geolocate, Fullscreen, and Scale controls would be removable.
> move mapbox language, mapbox draw and other controls to this monorepo. they will have same bundling tools, codestyle, docs, etc. I think it will make future support easier

👍 

> braking change

While it would be a breaking change to move the existing controls, moving the other plugins might be a good way to start this.

bravecow:
> Have you tried this? What sort of size reduction were you able to achieve by removing the controls. Only the Navigation, Geolocate, Fullscreen, and Scale controls would be removable.

`build-prod-min`:
`mapbox-gl.js`: 708KB
`mapbox-gl-without-controls.js`: 692KB - without controls
[...]

PR #8874, opened 2019-10-15
Add rules for High Contrast mode on Windows

kkaefer:
High Contrast White:

high-contrast-white-ie11

High Contrast Black:

high-contrast-black-ie11

We're inlining SVGs in our stylesheet, and this patch inlines almost identical SVGs several times. This is likely compressed away with gzipping the CSS file, but we could look into using a different way of rendering the icons, e.g. by embedding the SVG in the document and then using CSS to style individual paths rather than using `postcss` to materialize the recolored SVGs at build time.


kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/5784

andrewharvey:
> We're inlining SVGs in our stylesheet, and this patch inlines almost identical SVGs several times. This is likely compressed away with gzipping the CSS file, but we could look into using a different way of rendering the icons, e.g. by embedding the SVG in the document and then using CSS to style individual paths rather than using postcss to materialize the recolored SVGs at build time.

This is being discussed in #8767

kkaefer:
@andrewharvey Thanks for pointing out that issue, I wasn't aware of it.

kkaefer:
> Do we have any qualms about using this given the warning on MDN?

What MDN page is this from?

ryanhamley:
> > Do we have any qualms about using this given the warning on MDN?
> 
> What MDN page is this from?

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast

kkaefer:
Updated this PR with the optimized SVGs from https://github.com/mapbox/mapbox-gl-js/pull/8884. I also made a few changes to unify the way we use CSS:

- All clickable buttons are `<button>` elements. They get the `disabled` attribute set when they aren't active. Previously, some were disabled, some had the `mapboxgl-ctrl-icon-disabled` class set.
- All `<button>` elements now have a child node `<span class="mapboxgl-ctrl-icon" aria-hidden="true"></span>` that contains the icon as the background image. This is necessary because some icons are transformed and the clickable/hoverable area would change otherwise.
- The buttons are now 29 pixels wide instead of 30. I made this change because most icons have an odd width so that there's a pixel that is in the dead center of the icon.
- Removed a few `>` specifiers; they aren't necessary and make it harder to override the CSS

kkaefer:
IE11 regular:

regular

IE11 High Contrast Mode Black on White:

high-contrast-white

IE11 High Contrast Mode White on Black:

high-contrast-black

Issue #7975, opened 2019-03-01
mapbox.once event gives  Uncaught TypeError


sumitparkash:
**mapbox-gl-js version**
:0.53.0

I have added a layer in the mapbox having 3 points.

I want to show a popup on click of the point only once. For that, I have used the mabpox **once** event  as defined in documentation
https://docs.mapbox.com/mapbox-gl-js/api/#evented#once

However, when you click on the point it gives you error
**evented.js:121 Uncaught TypeError: s.call is not a function
    at r.dt.fire (evented.js:121)**

this is the code I used
`map.once('click', id, function (e) {
                    var coords = getMarkerLatLng(e, type);
                    new mapboxgl.Popup()
                            .setLngLat(coords)
                            .setHTML("Pooint Clicked")
                            .addTo(map);
                })`

If you change the **once** event with the **on**  event , then everything works fine.  .i.e this code works fine. 
`map.on('click', id, function (e) {
                    var coords = getMarkerLatLng(e, type);
                    new mapboxgl.Popup()
                            .setLngLat(coords)
                            .setHTML("Pooint Clicked")
                            .addTo(map);
                })`

But this function will be called every time. I need to call the function only  once when the user click on the point for the first time. 

I have created a  js fiddle to demonstrate the issue
https://jsfiddle.net/devsumitparkash/x5rk1n83/

Also wanted to know is there a way to send  extra data to the  mapbox click event like we have in jquery
`$('.leadtoscore').on('click', {event_type: 'shot'}, add_event);`

ansis:
It looks like `once` only supports two arguments:

`map.once('click', function(e) { ... });`

There is a special version of `on` implemented for `Map` that supports the third (middle) layer argument. @mourner should we support a three argument version of map.once?

@sumitparkash in the mean time you could use `map.off` within your callback function to remove the event listener. Or you could use the two argument version and manually call `queryRenderedFeatures` with your layer id within the callback.

mourner:
> @mourner should we support a three argument version of map.once?

I think we should, for consistency.

sumitparkash:
@ansis  thanks for your suggestion. Already using `map.off` .
But Still you be nice if `map.once()` behaves same as `map.on`

PR #8875, opened 2019-10-15
Add third parameter to once event handler

malwoodsantoro:
## Launch Checklist

Fixes https://github.com/mapbox/mapbox-gl-js/issues/7975 by allowing the use of a layer ID argument with the once event.

 - [ ] post benchmark scores

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -439,6 +439,55 @@ class Map extends Camera {
         return this._mapId;
     }
 
+
+    _createDelegatedListener(type: MapEvent, layerId: any, listener: any) {
+        if (type === 'mouseenter' || type === 'mouseover') {
+            let mousein = false;
+            const mousemove = (e) => {
+                const features = this.getLayer(layerId) ? this.queryRenderedFeatures(e.point, {layers: [layerId]}) : [];
+                if (!features.length) {
+                    mousein = false;
+                } else if (!mousein) {
+                    mousein = true;
+                    listener.call(this, new MapMouseEvent(type, this, e.originalEvent, {features}));
+                }
+            };
+            const mouseout = () => {
+                mousein = false;
+            };
+            return {layer: layerId, listener, delegates: {mousemove, mouseout}};
+        } else if (type === 'mouseleave' || type === 'mouseout') {
+            let mousein = false;
+            const mousemove = (e) => {
+                const features = this.getLayer(layerId) ? this.queryRenderedFeatures(e.point, {layers: [layerId]}) : [];
+                if (features.length) {
+                    mousein = true;
+                } else if (mousein) {
+                    mousein = false;
+                    listener.call(this, new MapMouseEvent(type, this, e.originalEvent));
+                }
+            };
+            const mouseout = (e) => {
+                if (mousein) {
+                    mousein = false;
+                    listener.call(this, new MapMouseEvent(type, this, e.originalEvent));
+                }
+            };
+            return {layer: layerId, listener, delegates: {mousemove, mouseout}};
+        } else {
+            const delegate = (e) => {
+                const features = this.getLayer(layerId) ? this.queryRenderedFeatures(e.point, {layers: [layerId]}) : [];
+                if (features.length) {
+                    // Here we need to mutate the original event, so that preventDefault works as expected.
+                    e.features = features;
+                    listener.call(this, e);
+                    delete e.features;
+                }
+            };
+            return {layer: layerId, listener, delegates: {[type]: delegate}};
+        }
+    };
+
     /**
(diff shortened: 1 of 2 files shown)


src/util/evented.js
```diff
@@ -89,7 +89,7 @@ export class Evented {
      * @param {Function} listener The function to be called when the event is fired the first time.
      * @returns {Object} `this`
      */
-    once(type: string, listener: Listener) {
+    once(type: *, listener: Listener) {
```
kkaefer: I think this should remain `string`. Any reason this had to change?

src/ui/map.js
```diff
@@ -877,6 +879,25 @@ class Map extends Camera {
...
+
+        if (listener === undefined) {
+            return super.once(type, layerId);
+        }
+
+        const delegatedListener = this._createDelegatedListener(type, layerId, listener);
+
+        this._delegatedListeners = this._delegatedListeners || {};
+        this._delegatedListeners[type] = this._delegatedListeners[type] || [];
+        this._delegatedListeners[type].push(delegatedListener);
```
kkaefer: This function uses `Evented#once` to add listeners, which means that the event listener will be added to `this._oneTimeListeners`. Once the event is fired, it will be removed from that list. However, it will _not_ be removed from this._delegatedListeners`, which means that handlers are going to accumulate in this array causing memory growth.

src/util/evented.js
malwoodsantoro: @kkaefer flow fails without `*` and `on` currently uses `*` as well (assuming for similar reasons).

src/ui/map.js
```diff
@@ -835,6 +837,51 @@ class Map extends Camera {
         return this;
     }
 
+    /**
+     * Adds a listener for events of a specified type.
```
ryanhamley: This should match the comment in `evented`

>Adds a listener that will be called only once to a specified event type.

src/ui/map.js
```diff
@@ -835,6 +837,51 @@ class Map extends Camera {
...
+     * @instance
+     * @param {string} type The event type to add a listen for.
+     * @param {Function} listener The function to be called when the event is fired.
+     *   The listener function is called with the data object passed to `fire`,
+     *   extended with `target` and `type` properties.
+     * @returns {Map} `this`
+     */
+
+    /**
+     * Adds a listener for events of a specified type occurring on features in a specified style layer.
```
ryanhamley: And this one should basically match the comment above but with the addition of `occurring on features in a specified style layer`

src/ui/map.js
```diff
@@ -835,6 +837,51 @@ class Map extends Camera {
...
     }
 
+    /**
+     * Adds a listener for events of a specified type.
+     *
+     * @method
+     * @name once
+     * @memberof Map
+     * @instance
+     * @param {string} type The event type to add a listen for.
```
ryanhamley: typo: `listener` not `listen`

vakila:
:wave: Hi @malwoodsantoro just wanted to check in on this - are we still planning to implement this? Anything we can do to help?

ryanhamley:
This just needs a second review @vakila

vakila:
@kkaefer I think @malwoodsantoro addressed your change requests, could you :eyes: and update your review? :pray:

src/util/evented.js
kkaefer: Ah, flow seems to have trouble matching a string to a string literal :/

src/ui/map.js
```diff
@@ -742,6 +742,54 @@ class Map extends Camera {
...
+                    mousein = false;
+                } else if (!mousein) {
+                    mousein = true;
+                    listener.call(this, new MapMouseEvent(type, this, e.originalEvent, {features}));
+                }
+            };
+            const mouseout = () => {
+                mousein = false;
+            };
+            return {layer: layerId, listener, delegates: {mousemove, mouseout}};
```
kkaefer: `_createDelegatedListener` returns `layer` and `listener`, but they're not used anywhere; instead the calling code accesses only the `delegates subobject. Let's remove these properties and return the delegates directly.

vakila: @kkaefer it looks like those properties are used in `removeDelegatedListener`, no?  https://github.com/mapbox/mapbox-gl-js/blob/ffe7d5d0488e13ef39629b3e46416a4a4e1edd0e/src/ui/map.js#L914

kkaefer: 🤦‍♂ you're right

vakila:
Awesome, let's merge this! :rocket: Thanks @malwoodsantoro @ryanhamley @kkaefer !

PR #8881, opened 2019-10-18
Report CSS size for PRs

kkaefer:
Adds size reporting for the CSS file, as well as gzipped size in check title.

It also adds the CSS file size to the bundle size calculation, so we should expect a jump by a few kilobytes.

Adds [cssnano](https://cssnano.co) for minifying CSS.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,7 +165,11 @@ jobs:
     steps:
       - attach_workspace:
           at: .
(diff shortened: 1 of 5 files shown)


kkaefer:
> Should we also produce a minified css file for the dist folder that can be included in releases?

This PR includes that as well.

ryanhamley:
Cool. I realized after I asked that including it in `postcss` probably did that but just wanted to be sure. Looks good 👍

PR #8883, opened 2019-10-21
add essential option to AnimationOptions to ignore prefers-reduced-motion for essential animations

### LINKED ISSUE #8743: A more granular control over handling of prefers-reduced-motion
## Motivation

There's quite a bit of confusion over GL JS handling of `prefers-reduced-motion` OS option (see #8288 and #8494 for background). Examples: #8716, #8648

## Design

From https://github.com/mapbox/mapbox-gl-js/pull/8494#issuecomment-529168645 by @andrewharvey:

> could be we expose some kind of option so the app developer can say if this transition is considered essential or not?

We could add `{essential: true}` to animation options (`false` by default) that makes it animate regardless of `prefers-reduced-motion`. That would allow developers to fully control the experience, marking animations as essential where necessary. cc @arindam1993

andrewharvey:
## Launch Checklist

Closes #8743 

When a user has set `prefers-reduced-motion: reduce` #8494 changed the default behaviour to never animate camera transitions, however user feedback and general views about how prefers-reduced-motion should behave at https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion indicates that a user setting this value usually means they only want to reduce non-essential animations, some animations which are considered essential to the app behaviour should still occur.

This PR introduces an `essential` option to `AnimationOptions` so developers can specify when animations should still occur regardless of prefers-reduced-motion setting. 

 - [ ] ~~tagged `@mapbox/studio` and/or `@mapbox/maps-design` if this PR includes style spec or visual changes~~
 - [ ] ~~tagged `@mapbox/gl-native` if this PR includes shader changes or needs a native port~~

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -658,8 +658,8 @@ class Camera extends Evented {
      * details not specified in `options`.
      *
      * Note: The transition will happen instantly if the user has enabled
-     * the `reduced motion` accesibility feature enabled in their operating system, unless options includes
-     * `essential: true`.
+     * the `reduced motion` accesibility feature enabled in their operating system,
+     * unless 'options' includes `essential: true`.
      *
      * @memberof Map#
      * @param options Options describing the destination and animation of the transition.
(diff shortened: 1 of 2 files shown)


andrewharvey:
I managed to fix the unit test, which I also validated fails before the change in this PR, then passes after the change in this PR.

PR #8884, opened 2019-10-21
Use inline SVGs and refresh CSS

kkaefer:
Supersedes https://github.com/mapbox/mapbox-gl-js/pull/8874, fixes https://github.com/mapbox/mapbox-gl-js/issues/5784. Implements https://github.com/mapbox/mapbox-gl-js/issues/8767.

We now insert SVG nodes directly into the DOM for displaying all icons and buttons instead of using data URLs in the CSS file. This enables us to restyle them more easily, e.g. for dark mode or high-contrast mode. Also reduces the size of some of the SVG icons significantly and changes the Mapbox logo to a version that is aligned better to the pixel grid on low-res devices.


kkaefer:
This changes the way we render icons significantly, so we need some extensive testing in all browsers. It's also a breaking change for users who are customizing the controls with CSS. While all of the classes are still there, the elements will now have `svg` child nodes with the actual icon. Users can either hide those, or instantiate the *Controls with custom SVGs.

kkaefer:
Before/After:

out

Attribution Before/After:

attribution

IE 11 Black Mode:
high-contrast-black

IE 11 White Mode:
high-contrast-white

mourner:
Heads up that this PR will require an update on GL Native side which currently uses the Node-specific shaders entry point (if this didn't change recently).

andrewharvey:
cc @bravecow

andrewharvey:
I would have thought keeping everything in a CSS file would have been better, that way it's easy for someone to either supply their own CSS to override certain classes to provide their own theme, or just drop in replace the provided CSS.

What are the changes to shaders? How does this relate to the themeing?

I have concerns also with the way SVG's can now be provided as arguments to the controls, some of these like the FullScreen and Geolocate controls are now expecting certain element IDs or Classes in the SVG, but none of that is documented in the public facing API docs. Currently (before this PR) you'd just go through the CSS and you'd seed which classes you need to style, but now without that, there's no documentation for what the custom SVG must look like.

bravecow:
SVG inline 👍 but for 4 icons you call `new window.DOMParser()....`

For the same reason for `mapbox-gl-controls` I created https://github.com/bravecow/rollup-inline-svg that returns SVG DOM Node using `window.DOMParser()`. (didn't find like this among all SVG rollup plugins...)

src/ui/control/logo_icon.svg
```diff
@@ -0,0 +1,33 @@
...
+    </defs>
+    <style>
+        .mapboxgl-ctrl-logo-outline {
```
bravecow: I think it is better to move styles to CSS (so they could be minified, autoprefixed, ...) or SVG attributes.

src/ui/control/fullscreen_icon.svg
```diff
@@ -0,0 +1,18 @@
+<svg viewBox="0 0 29 29" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <defs>
+        <path id="expand" d="M7.5 5C5.75 5 5 5.75 5 7.5V13h1l1.5-3 5.5 4 1-1-4-5.5L13 6V5H7.5z"/>
```
bravecow: If we inline SVG, we should care about unique `id`(as they are just plain nodes) attributes: https://stackoverflow.com/a/37000951/643514 (https://css-tricks.com/youre-inlining-svg-icons-deal-unique-titles-ids/)

I suggest to prefix them with `mapbox-` at least. (in case it is not already done by `svgo` or something).

src/css/mapbox-gl.css
```diff
@@ -85,8 +85,8 @@
 }
```
bravecow: Since `postcss-nested` is not valid syntax for CSS, I suggest to change extension to `.scss` to save code highlighting in code editors.

In future using `postcss-import` you can [simplify this CSS like this](https://github.com/bravecow/mapbox-gl-controls/blob/b1c1a51fff9c963cf4fcab83ac44788a90266b45/src/theme.scss#L1-L7).  😉

kkaefer:
Closing in favor of https://github.com/mapbox/mapbox-gl-js/pull/8874

PR #8905, opened 2019-10-25
Catch errors when createMap is called

tristen:
This commit adds a catch statement to any `createMap` calls in benchmarks so a meaningful error is provided when something goes wrong.

### DIFF
--- a/bench/benchmarks/remove_paint_state.js
+++ b/bench/benchmarks/remove_paint_state.js
@@ -53,7 +53,10 @@ class RemovePaintState extends Benchmark {
                     style
                 }).then(map => {
                     this.map = map;
-                });
+                })
+                    .catch(error => {
+                        console.error(error);
+                    });
             });
     }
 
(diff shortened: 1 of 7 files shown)


bench/benchmarks/remove_paint_state.js
```diff
@@ -53,7 +53,10 @@ class RemovePaintState extends Benchmark {
                     style
                 }).then(map => {
                     this.map = map;
-                });
+                })
+                    .catch(error => {
```
ryanhamley: Nit: the spacing is off here

tristen: `yarn run lint` told me to write it this way 🤷‍♂

ryanhamley: ok 🤷‍♂

tristen:
> Would it be better to throw an error and crash the tests if an error occurs or is just logging the error sufficient?

Ahhh this isn't really user-facing so I think logging the error is OK.

PR #8913, opened 2019-10-27
Fix animation performance

### LINKED ISSUE #8701: Animations of circle layer extremely sluggish in version 1.30
I am animating the appearance and disappearance of ~2000 circles over 10 seconds using a data driven circle-layer. In version 1.2.1, this was running fine at ~35fps, with the circles updating continuously. In version 1.3.0 the circles no longer update continuously - they update in batches only about once every 2 seconds, giving a very sluggish/jerky experience. (NB: If you just look at the fps, it's now reporting ~50fps, so by that metric it's faster than before, but that's because it's no longer doing the hard work of actually updating the circles. As soon as you look at the map as opposed to the fps meter, you see how bad the performance is now.)

**mapbox-gl-js version**: 1.3.0

**browser**: Chrome 76

### Steps to Trigger Behavior

 1. Add a data-driven circle layer.
 2. Animate the underlying data so that, say, 2000 circles appear randomly with a lifetime of 1 second each, over a 10 second interval.
 3. Watch how extremely sluggish it is compared to MapboxGL version 1.2.1.

### Expected Behavior
Animation at least as smooth as version 1.2.1.
### Actual Behavior
Animation dramatically less smooth than version 1.2.1.

msbarry:
Fix #8701 by processing messages immediately on the main thread and only queueing them up in workers. Updated `animate.html` to reproduce the issue since the performance issues started happening when geojson animation touches multiple tiles, so each `setData` request triggers multiple `reloadTile` messages and responses to them start piling up in the main thread. In chrome on Mac OSX, processed `setData` requests increased from <10 per second before this change to 30-40 after the change.

Before:
before
Funny thing is that reported updated per second goes up whenever I start a screen capture.

After:
after2

You can still see some renders where one tile shows new data and another tile shows old data. It would be nice to get rid of that entirely, but this at least brings performance back to what it was before 1.3.0.

 - [ ] write tests for all new functionality (not sure if one makes sense?)

### DIFF
--- a/src/util/actor.js
+++ b/src/util/actor.js
@@ -22,6 +22,7 @@ class Actor {
     target: any;
     parent: any;
     mapId: ?number;
+    isWorker: boolean;
     callbacks: { number: any };
     name: string;
     tasks: { number: any };
@@ -33,6 +34,7 @@ class Actor {
         this.target = target;
         this.parent = parent;
         this.mapId = mapId;
+        this.isWorker = typeof mapId !== 'number';
         this.callbacks = {};
         this.tasks = {};
         this.taskQueue = [];
@@ -107,7 +109,7 @@ class Actor {
                 cancel();
             }
         } else {
-            // Store the tasks that we need to process before actually processing them. This
+            // In workers, store the tasks that we need to process before actually processing them. This
             // is necessary because we want to keep receiving messages, and in particular,
             // <cancel> messages. Some tasks may take a while in the worker thread, so before
             // executing the next task in our queue, postMessage preempts this and <cancel>
@@ -115,7 +117,13 @@ class Actor {
             // process() flow to one at a time.
             this.tasks[id] = data;
             this.taskQueue.push(id);
-            this.invoker.trigger();
+            if (this.isWorker) {
+                this.invoker.trigger();
+            } else {
+                // In the main thread, process messages immediately so that other work does not slip in
+                // between getting partial data back from workers.
+                this.process();
+            }
         }
     }
 
--- a/src/util/actor.js
+++ b/src/util/actor.js
@@ -1,6 +1,6 @@
 // @flow
 
-import {bindAll} from './util';
+import {bindAll, isWorker} from './util';
 import {serialize, deserialize} from './web_worker_transfer';
 import ThrottledInvoker from './throttled_invoker';
 
@@ -22,7 +22,6 @@ class Actor {
     target: any;
     parent: any;
     mapId: ?number;
-    isWorker: boolean;
     callbacks: { number: any };
     name: string;
     tasks: { number: any };
@@ -34,7 +33,6 @@ class Actor {
         this.target = target;
         this.parent = parent;
         this.mapId = mapId;
-        this.isWorker = typeof mapId !== 'number';
         this.callbacks = {};
         this.tasks = {};
         this.taskQueue = [];
@@ -117,7 +115,7 @@ class Actor {
             // process() flow to one at a time.
             this.tasks[id] = data;
             this.taskQueue.push(id);
-            if (this.isWorker) {
+            if (isWorker()) {
                 this.invoker.trigger();
             } else {
                 // In the main thread, process messages immediately so that other work does not slip in
--- a/debug/animate.html
+++ b/debug/animate.html
@@ -43,8 +43,8 @@ function crossWithAngle(angle) {
     return {
         "type": "FeatureCollection",
         "features": [
-            line(angle, radius),
-            line(angle, radius + Math.PI / 2)
+            line(angle - Math.PI / 4, radius),
+            line(angle + Math.PI / 4, radius)
         ]
     };
 }
@@ -63,6 +63,16 @@ map.on('load', function () {
         "paint": {
             "line-width": 4,
             "line-color": "#007cbf"
+        }
+    });
+
+    map.addLayer({
+        "id": "dot",
+        "source": "lines",
+        "type": "circle",
+        "paint": {
+            "circle-radius": 10,
+            "circle-color": "#007cbf"
         }
     });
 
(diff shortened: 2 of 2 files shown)


msbarry:
I think it is possible to remove the rendering artifacts by batching requests sent while handling a single response to ensure all of the `reloadTile`, etc. messages resulting from a `setData` get handled in the same animation frame.

https://github.com/msbarry/mapbox-gl-js/commit/1f6cd7964225948b3de993f4f71b80dca53216bd

This adds some complexity and might not be desirable in all cases. Let me know if you if you think it's worth pursuing.

maybe

kkaefer:
Thank your for proposing this fix. Your theory seems very plausible, but before merging, I'd like to verify that it does indeed fix the problem. I can't spot a difference (even for full a fullscreen, 2560×1440 map) on my main computer. Same goes for a phone. On a slower/older computer, I can observe that frames are dropped, and I see the tearing across multiple tiles, but I can't observe a difference between 1.5.0 and this PR either :(

Could you please share the debug code that you used to print the update/second messages, as well as tell us what device you're using to observe the improvement?

msbarry:
For me the geojson updates need to touch many tiles for this to start happening, so I changed debug/animate.html to spin the 2 lines, on a bigger monitor you might need to make it bigger and possibly add lines at different angles to get it touching a lot of tiles on each update. I also changed debug/animate.html to log the number of geojson updates each second to the console. Can you reproduce any of the performance differences in examples people link to from #8701? 

Animation performance was never great, but it got a lot worse after 1.3.0 and from what I can see this brings it back to what it was beforehand so I suspect you're not seeing the performance drop the rest of us are with 1.3.0+.  We would need something more complex like https://github.com/msbarry/mapbox-gl-js/commit/87688f7147c42faa11d757534f03893d65864976 to improve on this and remove the tearing completely, or maybe have the `setData` call also reload tiles to reduce the number of round trips for geojson animation.

I'm running mac os 10.14.4 with chrome 77.0.3865.120 on a 2015 macbook pro.

msbarry:
Also FWIW I pinged #8701 and this does seem to be resolving the issue that people observed.

mourner:
@msbarry yes, this is great — we intend to merge the PR after doing some additional browser testing.

src/util/actor.js
```diff
@@ -107,15 +109,21 @@ class Actor {
...
+            // In workers, store the tasks that we need to process before actually processing them. This
             // is necessary because we want to keep receiving messages, and in particular,
             // <cancel> messages. Some tasks may take a while in the worker thread, so before
             // executing the next task in our queue, postMessage preempts this and <cancel>
             // messages can be processed. We're using a MessageChannel object to get throttle the
             // process() flow to one at a time.
             this.tasks[id] = data;
             this.taskQueue.push(id);
-            this.invoker.trigger();
+            if (this.isWorker) {
```
arindam1993: Thanks for your contribution @msbarry ! :)
Could you please change this to use the `isWorker()` function in `src/util.js`. Code looks good otherwise.

I'll post browser testing reports here.

arindam1993:
Ran some browser tests with @msbarry 's updated animate example, 

| browser                         | master | pr    |
|---------------------------------|--------|-------|
| chrome (osx)                    | 60     | 60    |
| chrome(android) (2016 pixel xl) | 15-20  | 30-50 |
| chrome(windows)                 | 30-60  | 50-60 |
| firefox(osx)                    | 55-60  | 55-60 |
| firefox(windows)                | 15-30  | 15-30 |
| ie 11                           | 0-1    | 2-5   |

The numbers in the column represent the updates per second from the example, looks like the performance difference is significant enough especially on mobile, and on windows. 

This should be good to land if you make the `isWorker` change!

src/util/actor.js
msbarry: Thanks! Just switched to `isWorker()`

msbarry:
> Ran some browser tests with @msbarry 's updated animate example,
> 
> browser	master	pr
> chrome (osx)	60	60
> chrome(android) (2016 pixel xl)	15-20	30-50
> chrome(windows)	30-60	50-60
> firefox(osx)	55-60	55-60
> firefox(windows)	15-30	15-30
> ie 11	0-1	2-5
> The numbers in the column represent the updates per second from the example, looks like the performance difference is significant enough especially on mobile, and on windows.
> 
> This should be good to land if you make the `isWorker` change!

Thanks for the tests!  So strange though, I wonder what could be causing the <5 updates per second I was seeing on chrome on osx without the fix.

arindam1993:
Hey @mourner , @kkaefer and were chatting about how this PR could bring back  the issue of requests being stuck in the pending state on IE11 when tile requests get cancelled, I just did some testing and it looks like it has 😬.

Sorry I forgot about this during my initial testing.

mourner:
@arindam1993 🤔 is there some down-the-middle approach we could take, or do we have to fully revert?

arindam1993:
I have an idea based on what this PR does, but flipping it around.
Use the async-queue on the main thread, but use requestAnimationFrame to flush the queue instead of `MessageChannel`/`setTimeout(..,0)`, and switch the worker side to be synchronous.

This should mean that we should get 1 `setData` call sent to the worker every frame when animating, and still have the benefit of queueing things up in the main thread so requests can get cancelled.

Lets not revert, I'll try out the above idea and see if it helps.

msbarry:
Thanks @arindam1993, would definitely be ideal not to revert so people blocked on this issue can upgrade to the next release. Can you ping me on the fix and I can test it out on my setup since it seems to be finicky to reproduce?

If that’s not sufficient, another possibility might be to change the message queue interface to send messages back and forth in batches, so any message sent on the same event loop get batched your end handled in same event loop on the other side, something like what I prototyped in https://github.com/msbarry/mapbox-gl-js/commit/87688f7147c42faa11d757534f03893d65864976. This also removes tearing, but adds some complexity.

msbarry:
I did some quick testing on my side with chrome on osx

| version | updates per second |
| --- | --- |
| this PR | 30-40 |
| synchronous worker, message queue in main thread | 5-10 |
| debounce and batch send with message queue on each side https://github.com/mapbox/mapbox-gl-js/commit/769feb6f2601f1b41acb4e4f5550cff4be57cbb2 | 20-25 and no tearing |

PR #8918, opened 2019-10-28
Add stalebot config for processing stale PR's

arindam1993:
This add's a bot that automatically closes PR's that have had no activity for > 30 days.

It also uses the `DO NOT MERGE` label as an explicit way of skipping the auto-close. We could rename that label to `backburner` if that makes more sense.

This also only looks at PR's right now, we can add similar automation for issues later.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,18 +1,19 @@
 # Configuration for probot-stale - https://github.com/probot/stale
+only: pulls
 
+# conditions by whcih to exempt closing
 exemptMilestones: true
-
-staleLabel: archived
-
 exemptLabels:
   - "DO NOT MERGE"
 
-pulls:
-  daysUntilStale: 30
-  daysUntilClose: -1
-  markComment: false
-  closeComment: >
-    Thanks for your contribution, but this PR has not had any activity for a month and has been automatically
-    detected as stale and is being closed.
-    Please feel free to reopen if you have any additional progress to share :), all automatically closed PR's can be searched
-    for using the "archived" tag.
+staleLabel: archived
+
+daysUntilStale: 30
+daysUntilClose: -1 # Close immediately
+
+markComment: false
+closeComment: >
+  Thanks for your contribution, but this PR has not had any activity for a month and has been automatically
+  detected as stale and is being closed.
+  Please feel free to reopen if you have any additional progress to share :), all automatically closed PR's can be searched
+  for using the "archived" tag.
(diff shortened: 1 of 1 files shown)


andrewharvey:
> This add's a bot that automatically closes PR's that have had no activity for > 30 days.

I'm strongly against this. When PR's which may not be a priority for the maintainers to review come in, they sometimes sit around for quite a while, I think that's okay I'm not expecting anyone else drop everything to review a PR, but at least the PR should remain open until it has either been accepted and merged or rejected.

arindam1993:
> > This add's a bot that automatically closes PR's that have had no activity for > 30 days.
> 
> I'm strongly against this. When PR's which may not be a priority for the maintainers to review come in, they sometimes sit around for quite a while, I think that's okay I'm not expecting anyone else drop everything to review a PR, but at least the PR should remain open until it has either been accepted and merged or rejected.

Thats fair, auto closing is too harsh. How does just marking PR's as stale with a label sound?

The use-case this is trying to address is PR's that are left open for too long have diverged so far from `master` that they are hard to get back upto speed, and we atleast need a way to categorize these PR's that have fallen through the cracks, and send the original contributor some kind of reminder.

andrewharvey:
> Thats fair, auto closing is too harsh. How does just marking PR's as stale with a label sound?
> 
> The use-case this is trying to address is PR's that are left open for too long have diverged so far from master that they are hard to get back upto speed, and we atleast need a way to categorize these PR's that have fallen through the cracks, and send the original contributor some kind of reminder.

Usually it's one of three situations:

- the PR is ready from the authors perspective, they are just waiting for a review/merge
- the PR is incomplete but the author needs feedback from maintainers before they move ahead with completing the PR
- the PR is still a work in progress

From the human side, auto closing can be very harmful for moral especially for case 1 or 2. Dealing with each PR on a case by case situation and working together with everyone involved on why it's stalled and starting with seeing what can be done to unstall it I think is better.

If it's the case where changes are requested and it's been a while, then I agree with you starting with a comment asking if they are still interested in working on it or not is a good first step.

vakila:
@arindam1993 Thanks for making this suggestion and @andrewharvey thanks for raising those very valid concerns - this was part of our ongoing exploration of ways to help the team stay on top of issues/PRs as they come in and prevent things from falling through the cracks as they have done in the past, all with the aim of delivering a better contributor experience. After some discussion we've decided that auto-closing/marking isn't the best solution right now, so we're closing this out. Thanks! :heart:

PR #8929, opened 2019-10-31
[wip] add showUserHeading option to GeolocateControl

### LINKED ISSUE #8034: Give Geolocation Control Marker Pitch Alignment Properties

### LINKED ISSUE #8329: Add heading to GeolocateControl
## Motivation
Navigating yourself from a map with only position is difficult, showing a direction indicator or compass heading makes it much easier. Most modern mobile devices have a compass built in. Google Maps supports it.

## Design
  map.addControl(new mapboxgl.GeolocateControl({
        positionOptions: {
                  enableHighAccuracy: true
              },
        fitBoundsOptions: {
          maxZoom:7,
        },
        trackUserLocation: true
        trackUserHeading: true <----
  }), 'bottom-left');

### Mock-Up
Example from Google Maps, see the direction arrow pointing out of the blue dot.
xbZ8a

andrewharvey:
## Launch Checklist

Add a `showUserHeading` option to the GeolocateControl to show the device heading as an arrow next to the user location dot.

Sensor access is disabled by default in Safari from iOS 12.2+, so for this to work you need to enable that setting in Safari Settings, once enabled it uses `deviceOrientationEvent.webkitCompassHeading`.

closes #8329
closes #8034

heading

 - [ ] post benchmark scores
- [ ] cross device testing
- [ ] investigate if it's better to use https://developer.mozilla.org/en-US/docs/Web/API/Coordinates/heading instead of `deviceorientation`.

### DIFF
--- a/debug/debug.html
+++ b/debug/debug.html
@@ -47,6 +47,7 @@ map.addControl(new mapboxgl.GeolocateControl({
     },
     trackUserLocation: true,
     showUserLocation: true,
+    showUserHeading: true,
     fitBoundsOptions: {
         maxZoom: 20
     }
--- a/src/css/mapbox-gl.css
+++ b/src/css/mapbox-gl.css
@@ -746,9 +746,10 @@ a.mapboxgl-ctrl-logo.mapboxgl-compact {
     height: 0;
     border-left: 7.5px solid transparent;
     border-bottom: 7.5px solid #1da1f2;
-    transform: translate(0px, -28px) skewY(-20deg);
+    transform: translate(0, -28px) skewY(-20deg);
     position: absolute;
 }
+
 .mapboxgl-user-location-show-heading .mapboxgl-user-location-heading::after {
     content: "";
     width: 0;
--- a/src/css/mapbox-gl.css
+++ b/src/css/mapbox-gl.css
@@ -735,6 +735,30 @@ a.mapboxgl-ctrl-logo.mapboxgl-compact {
     box-shadow: 0 0 3px rgba(0, 0, 0, 0.35);
 }
 
+.mapboxgl-user-location-show-heading .mapboxgl-user-location-heading {
+    width: 0;
+    height: 0;
+}
+
+.mapboxgl-user-location-show-heading .mapboxgl-user-location-heading::before {
+    content: "";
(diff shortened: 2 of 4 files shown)


ahk:
Thanks for this PR. I'm curious if you have knowledge/experience/thoughts about how many users choose to enable sensor access in mobile Safari. Is this different from the permissions dialog that asks if location data should be enabled on a per-site basis?

andrewharvey:
> Thanks for this PR. I'm curious if you have knowledge/experience/thoughts about how many users choose to enable sensor access in mobile Safari. Is this different from the permissions dialog that asks if location data should be enabled on a per-site basis?

No idea how many enable it, but probably very little have it enabled since it's off by default and most users would have no reason to enable it, that said an application could prompt you to go into settings and enable, and I think applications can still do that independently if they like. So I still think we should support it when it is enabled.

It's an all or nothing switch in Safari (in the app settings) it's different to the permissions dialog for location access.

I investigated the coords.heading field from the geolocation API https://developer.mozilla.org/en-US/docs/Web/API/Coordinates/heading and found that iOS Safari doesn't support it (tested on iOS 12.4.2). Chrome on Android does, but that already supports an absolute bearing from `deviceorientation`.

ahk:
> I investigated the coords.heading field from the geolocation API
Do you think we should also support this field as well? I have FireFox and FireFox Preview on my Android phone. As well it's possible for desktop/laptop FireFox to have a compass reading in some weird cases right?

ahankeosi:
Any movement on this?  Would really love this feature!

andrewharvey:
> Any movement on this? Would really love this feature!

Nope, feel free to pick up my work in progress if you feel it's helpful.

CptHolzschnauz:
Stick to one's guns ! Any movement to PR?

andrewharvey:
this was completed by @tsuz in #10817

PR #8949, opened 2019-11-06
Add some memory stats to gl-stats

ryanhamley:
## Launch Checklist

    - adds a few basic memory usage stats to our stats
        - `array_buffer_count` - shows how often we're explicitly allocating large chunks of memory
        - `total_objects_in_memory` - this can be tracked over a number of sequential runs to see if it continually grows which could indicate a memory leak (based on an idea from https://media-codings.com/articles/automatically-detect-memory-leaks-with-puppeteer)
        - `js_heap_size_used` - shows how much of the heap GL JS utilizes in raw numbers
        - `percent_js_heap_used` - shows the percent of the heap GL JS utilizes

Are these all useful? Am I missing other useful memory allocation and consumption metrics? Ideally, these metrics can be combined with https://github.com/mapbox/mapbox-gl-js/pull/8931 to allow us to observe memory allocation and consumption over time.

### DIFF
--- a/bench/gl-stats.js
+++ b/bench/gl-stats.js
@@ -22,6 +22,17 @@ function waitForConsole(page) {
     });
 }
 
+async function countInstancesInMemory(page, prototype) {
+    // Query all buffer instances into an array
+    const instances = await page.queryObjects(prototype);
+    // Count amount of buffer objects in heap
+    const count = await page.evaluate((buffers) => buffers.length, instances);
+    await instances.dispose();
+    await prototype.dispose();
+
+    return count;
+}
+
 (async () => {
     const browser = await puppeteer.launch({
         args: ['--no-sandbox', '--disable-setuid-sandbox']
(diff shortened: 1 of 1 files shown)


mourner:
@ryanhamley this looks great, certainly very useful! A few questions:

- Are the buffer instance counts for the main thread or overall including workers?
- Can we measure the sum of buffer sizes (not just counts)?
- How stable is `js_heap_size_used`, does it jump between runs? I remember an equivalent metric in Node being very volatile due to GC kicking in inconsistently.

ryanhamley:
>Are the buffer instance counts for the main thread or overall including workers?
Main thread. Puppeteer gives access to workers, but as far as I can tell, the worker threads do not expose the `queryObjects` method which would allow us to count buffers in memory. But we pass most buffers back to the main thread, correct? So counting on the main thread should give us a good idea.

>Can we measure the sum of buffer sizes (not just counts)?
I'm working on this. I think it should be possible but I haven't worked out exactly how to iterate over the elements in the JSHandle of ArrayBuffers in Puppeteer yet.

>How stable is js_heap_size_used, does it jump between runs? I remember an equivalent metric in Node being very volatile due to GC kicking in inconsistently.

It seems pretty stable. `array_buffer_count` is almost always the same and we almost always use ~60% of the JS heap, give or take a percent.

ryanhamley:
@arindam1993 and I have begun work on integrating his performance metric suite with puppeteer so that we can gather these memory metrics over a number of runs. you can see that work in https://github.com/mapbox/mapbox-gl-js/pull/8931

ryanhamley:
This is a prototype PR which has been integrated with other work on metrics so I'm closing it

PR #8953, opened 2019-11-06
allow arrow keys to rotate and pitch the map when focused on the NavigationControl compass button

### LINKED ISSUE #8459: arrow keys to change bearing when bearing button is active
To improve keyboard accessibility, when the bearing control button is active, I think the arrow keys should change bearing and pitch, similar to shift+arrow keys when the map is active.

andrewharvey:
## Launch Checklist

closes #8459

given you can rotate and pitch with the mouse on the compass button, the same actions should be achievable with the keyboard only.

there were no existing tests for the NavigationControl so I haven't added any, should I go ahead and add some unit tests for the NavigationControl or does this control not need unit tests?

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -5,6 +5,7 @@ import {extend, bindAll} from '../../util/util';
 import DragRotateHandler from '../handler/drag_rotate';
 
 import type Map from '../map';
+import {keyboardBearingStep, keyboardPitchStep} from '../constants';
 
 type Options = {
     showCompass?: boolean,
@@ -103,9 +104,6 @@ class NavigationControl {
             DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
             DOM.addEventListener(this._compass, 'touchstart', this._handler.onMouseDown, {passive: false});
             DOM.addEventListener(this._compass, 'keydown', (e) => {
-                const bearingStep = 15,
-                    pitchStep = 10;
-
                 let bearingDir = 0;
                 let pitchDir = 0;
 
@@ -129,8 +127,8 @@ class NavigationControl {
                     delayEndEvents: 500,
                     easing: easeOut,
 
-                    bearing: this._map.getBearing() + bearingDir * bearingStep,
-                    pitch: this._map.getPitch() + pitchDir * pitchStep
+                    bearing: this._map.getBearing() + bearingDir * keyboardBearingStep,
+                    pitch: this._map.getPitch() + pitchDir * keyboardPitchStep
                 }, {originalEvent: e});
             });
             this._handler.enable();
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -110,18 +110,18 @@ class NavigationControl {
                 let pitchDir = 0;
 
                 switch (e.code) {
-                    case 'ArrowLeft':
-                        bearingDir = -1;
-                        break;
-                    case 'ArrowRight':
-                        bearingDir = 1;
-                        break;
-                    case 'ArrowUp':
-                        pitchDir = -1;
-                        break;
-                    case 'ArrowDown':
-                        pitchDir = 1
-                        break;
+                case 'ArrowLeft':
+                    bearingDir = -1;
+                    break;
+                case 'ArrowRight':
+                    bearingDir = 1;
+                    break;
+                case 'ArrowUp':
+                    pitchDir = -1;
+                    break;
+                case 'ArrowDown':
+                    pitchDir = 1;
+                    break;
                 }
 
                 this._map.easeTo({
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -102,6 +102,37 @@ class NavigationControl {
             this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass});
             DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
             DOM.addEventListener(this._compass, 'touchstart', this._handler.onMouseDown, {passive: false});
+            DOM.addEventListener(this._compass, 'keydown', (e) => {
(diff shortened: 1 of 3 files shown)


src/ui/control/navigation_control.js
```diff
@@ -102,6 +102,37 @@ class NavigationControl {
             this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass});
             DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
             DOM.addEventListener(this._compass, 'touchstart', this._handler.onMouseDown, {passive: false});
+            DOM.addEventListener(this._compass, 'keydown', (e) => {
+                const bearingStep = 15,
```
andrewharvey: I've made the implementation here since the zoom in/out button clicks are performed here, and the KeyboardHandler seems focused on handling keyboard events on the map, not the NavigationControl. This does lead to these constants defined twice, is that okay or should they be defined once and them imported into both the KeyboardHandler and NavigationControl?

ahk:
> does this control not need unit tests?

Do you have an idea how you would test this? I'd be really appreciative if you drafted some simple ones to get us started. I don't have strong opinions on how to unit test this.

I personally like supporting accessibility features but as with many of our smaller features we're a bit challenged right now adequately retesting these things. Having some basic tests to verify regressions would help a lot with adding more small improvements like this to things that fall outside our main development path.

src/ui/control/navigation_control.js
ahk: Sorry, which constants are defined twice here?

We don't have a clean separation of focus and input handling, probably KeyboardHandler is a bit misnamed if we have multiple controls that need to handle the keyboard but KeyboardHandler just represents the focused map.

andrewharvey: bearingStep and ptichStep are also defined at https://github.com/mapbox/mapbox-gl-js/blob/237713463802baa397db21dce14d8987b8e36e66/src/ui/handler/keyboard.js#L8-L9

andrewharvey:
> Do you have an idea how you would test this? I'd be really appreciative if you drafted some simple ones to get us started. I don't have strong opinions on how to unit test this.

invoke the key event and then compare the map rotation and pitch to before the key event was invoked.

I'll do some more work to add in unit tests.

src/ui/control/navigation_control.js
ahk: Ah thank you. I think it's good to have them defined once and imported. I don't think we'd want to modify these values separately. Maybe it's a good time to start a `src/ui/constants.js`? I am having a hard time finding other examples in the project of a shared constants file but it seems reasonable to me.

andrewharvey:
So I was part way through implementing unit tests and just saw #9036 which if implemented would nullify this change. So let's hold off for now on this PR until #9036 is resolved.

andrewharvey:
It might be useful to still keep this functionality even if #9036 goes ahead since shift+arrow is hard to discover, whereas the arrow keys when focused on the compass might be more discoverable? I'm not sure.

vakila:
Hi @andrewharvey, first of all thanks for all your work on this - didn't mean to derail it with #9036 but I do think we need to examine whether this functionality is necessary. That said, as I mentioned in https://github.com/mapbox/mapbox-gl-js/issues/9036#issuecomment-560570631 I'm not sure that implementing #9036 means we shouldn't still consider merging this - I agree with your last comment:

> It might be useful to still keep this functionality even if #9036 goes ahead since shift+arrow is hard to discover, whereas the arrow keys when focused on the compass might be more discoverable? I'm not sure.

I'm not sure either - I can see both arguments: 
1. The keyboard controls should have parity with the touch controls, so if we remove the drag-rotate/-pitch compass control then we should not add keyboard-rotate/-pitch controls. (AFAICT clicking the buttons to zoom in/out or reset bearing/pitch is already matched by default behavior of enter/space key when buttons are focused, so we would have parity there without any additional changes.) The fact that we need to reduplicate some of the keyboard handler logic within the `NavigationControl` here could be an indication that we're trying to cram more functionality into this control than it really should have, given that there are existing mechanisms for manipulating the map without this control.
2. It might be generally good to give keyboard users more options for how they can control the map, including a modifier-free option by focusing the compass and then using arrows to change bearing/pitch. I'm not entirely sure why, but to me it makes more sense to manipulate the compass button that way (with keyboard) than with the current touch gesture. (But touch users should also have other options, e.g. a touch gesture to control pitch, which we are working on.)

@andrewharvey do you have any feedback/data points from users or other devs that the shift-arrow control is too hard to discover and that focusing the compass and then using arrows would be more user-friendly? That would make sense to me but just based on speculation as I don't really have any data to go off of - wondering if you have data points on that or how we could go about getting some. 

I could go either way on this - if we have reason to believe that this will provide the best experience for keyboard-only/keyboard-preferred users, I'm fine with adding this regardless of #9036. Either way, I would love to merge in the test(s) you added, they're sorely needed! 

cc @ahk @ryanhamley

andrewharvey:
> didn't mean to derail it with #9036 but I do think we need to examine whether this functionality is necessary. 

Oh not at all, I agree that it's important to re-consider if the current UX is worth retaining going forward.

> @andrewharvey do you have any feedback/data points from users or other devs that the shift-arrow control is too hard to discover and that focusing the compass and then using arrows would be more user-friendly? That would make sense to me but just based on speculation as I don't really have any data to go off of - wondering if you have data points on that or how we could go about getting some.

Not really, I'm not sure any particular method is better :man_shrugging: .

I think this PR is quite low priority so happy to leave it sit here while we see how #9036 pans out, and in case there is any other user feedback.

> Either way, I would love to merge in the test(s) you added, they're sorely needed!

The unit tests I did were just for the functionality added in this PR, so no harm done if you're working on some too. I'm not fussed, if you have some ready to ship that's great!

mourner:
Looks like this is stale — let's close, and possibly reopen when we're ready to work on landing this again.

PR #8956, opened 2019-11-08
Work around CacheStorage memory leak in Safari

kkaefer:
We are periodically removing items from the cache using the CacheStorage API. Unfortunately, Safari seems to leak significant amounts of memory when requesting a new Cache object, then calling `cache.keys()`. Instead of requesting a new Cache object for every operation, we are now reusing the same Cache object for all operations on the cache.

I ticketed this upstream at https://bugs.webkit.org/show_bug.cgi?id=203991

### DIFF
--- a/src/util/tile_request_cache.js
+++ b/src/util/tile_request_cache.js
@@ -17,6 +17,8 @@ export type ResponseOptions = {
     headers: window.Headers
 };
 
+const cacheRequest = window.caches ? window.caches.open(CACHE_NAME) : null;
+
 let responseConstructorSupportsReadableStream;
 function prepareBody(response: Response, callback) {
     if (responseConstructorSupportsReadableStream === undefined) {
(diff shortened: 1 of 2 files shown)


kkaefer:
Memory graph before:

cache_before_clipped

Memory graph after:

cache_after_clipped

(compare y-axes)

src/util/tile_request_cache.js
```diff
@@ -17,6 +17,8 @@ export type ResponseOptions = {
     headers: window.Headers
 };
 
+const cacheRequest = window.caches ? window.caches.open(CACHE_NAME) : null;
```
asheemmamoowala: @ansis @kkaefer Is there a chance that these access or method will throw an exception? I'm not seeing any documented cases in the MDN docs, and have tested in Sfari, Firefox and Chrome with `localhost`, `127.0.0.1`, and  over `http://`. 
This should probably be tested with Windows browsers and in incognito modes as well.

kkaefer: These functions generally shouldn't throw exceptions because they return promises. We use `Promise#catch` in some situations to cover the failure case.

kkaefer:
Tested in IE11, Edge, macOS Chrome/Firefox/Safari.

Issue #8819, opened 2019-09-28
Proposal: require more complete JSDoc documentation


danswick:
The plurality of the feedback the docs team has been receiving since launching @mapbox/docs ' [feedback module](https://github.com/mapbox/dr-ui/pull/135) has pertained to the Mapbox GL JS docs. Much of _that_ feedback has been about missing parameters, return values, or descriptions in our API documentation generated from JSDoc comments. I would like to propose adding some stricter tests for requiring JSDoc descriptions, parameters, and return values for all public methods. 

## Testing / linting

There's a handy and popular [ESLint plugin](https://github.com/gajus/eslint-plugin-jsdoc) for checking various JSDoc parameters. One benefit of an ESLint plugin is integration with text editors so GL JS contributors can get feedback while working in the codebase. 

## Establishing a baseline

In order to get a sense for the scope of work required to get Mapbox GL JS to a "complete" state where JSDoc is relatively complete and linting/tests pass, is set up a working branch here: https://github.com/mapbox/mapbox-gl-js/compare/eslint-jsdoc?expand=1 with the following options configured: 

```
"jsdoc/check-param-names": 1,
"jsdoc/require-jsdoc": 1,
"jsdoc/require-param": 1,
"jsdoc/require-param-description": 1,
"jsdoc/require-param-name": 1,
"jsdoc/require-returns": 1,
"jsdoc/require-returns-description": 1
```

In its current form, my test branch has about 400 JSDoc issues, many of which are for missing JSDoc comments for non-public methods. I could see a few potential paths forward: 

1. For the `missing JSDoc` warnings, add minimal JSDoc comments for every missing instance flagged by ESLint and mark all as `@private` (there is an option to ignore all methods marked as private). This option would allow us to be strict about requiring documentation everywhere without requiring a tremendous amount of effor up-front. While my personal opinion is that even private methods should be well-documented, I'm not sure if option is valuable without more detailed descriptions, which would be quite difficult for a causal contributor (🙋‍♂️) to add for many internal methods.
[...]

PR #8960, opened 2019-11-09
Eslint JSDoc fixes 

danswick:
Per https://github.com/mapbox/mapbox-gl-js/issues/8819, this is a companion PR to https://github.com/mapbox/mapbox-gl-js/pull/8961 to address the JSDoc issues surfaced by a proposed eslint configuration change. There are a number of open questions that would be great to nail down before moving further (added inline as I review the diff).


src/geo/transform.js
```diff
@@ -206,10 +206,10 @@ class Transform {
 
     /**
      * Return a zoom level that will cover all tiles the transform
-     * @param {Object} options
-     * @param {number} options.tileSize
-     * @param {boolean} options.roundZoom
-     * @returns {number} zoom level
+     * @param {Object} options options
```
danswick: There are a lot of `options` params. It might not be worthwhile to add descriptions to each, but I am not sure how to configure eslint to ignore them.

src/ui/handler/scroll_zoom.js
```diff
@@ -79,15 +79,15 @@ class ScrollZoomHandler {
 
     /**
      * Set the zoom rate of a trackpad
-     * @param {number} [zoomRate = 1/100]
+     * @param {number} [zoomRate=1/100] Zoom rate in some unit.
```
danswick: I'm not sure what this unit is (same for wheelRateZoom)

src/ui/map.js
```diff
@@ -563,6 +563,7 @@ class Map extends Camera {
     /**
      * Returns the map's geographical bounds. When the bearing or pitch is non-zero, the visible region is not
      * an axis-aligned rectangle, and the result is the smallest bounds that encompasses the visible region.
+     * @returns {LngLatBounds} The geographical bounds of the map as {@link LngLatBounds}.
```
danswick: There are **a lot** of JSDoc descriptions that are really just a description of the return value. It may make more sense to move these descriptions to the `@return` section and use the description for more general information.

src/ui/map.js
```diff
@@ -810,6 +812,7 @@ class Map extends Camera {
 
     /**
      * Returns true if the map is panning, zooming, rotating, or pitching due to a camera animation or user gesture.
+     * @returns {Boolean} True if the map is moving.
```
danswick: Another example of the general description stealing the `@return` value's thunder.

src/ui/map.js
```diff
@@ -1367,6 +1372,7 @@ class Map extends Camera {
         return this.style.getSource(id);
     }
 
+    // eslint-disable-next-line jsdoc/require-returns
```
danswick: I wasn't sure if it would be worthwhile to describe the return value for methods like these where a return value **exists** but it might not be useful to the user.

src/ui/marker.js
```diff
@@ -579,7 +580,7 @@ export default class Marker extends Evented {
 
     /**
      * Returns the current `rotationAlignment` property of the marker.
-     * @returns {string}
+     * @returns {string} The current rotational alignment of the marker.
```
danswick: Some more examples of awkward description/`@returns` conflicts.

mourner:
This is awesome! Nit: looks like you're using an old version of Yarn — can you upgrade to the latest so that the PR doesn't remove integrity fields in yarn.lock?

src/data/array_types.js
```diff
@@ -854,6 +854,7 @@ export class CollisionBoxArray extends StructArrayLayout6i1ul2ui2i24 {
     /**
      * Return the CollisionBoxStruct at the given location in the array.
      * @param {number} index The index of the element.
+     * @private
```
mourner: Is there a way to prevent redundantly specifying `@private` on every method if the parent class is marked `@private` anyway?

src/geo/transform.js
mourner: Let's see how this looks on the live docs — we might need to come up with a way to filter this out or maybe PR the plugin repo with an option to ignore.

src/data/array_types.js
asheemmamoowala: Looks like this is not supported. From the JSDoc [docs](https://jsdoc.app/tags-private.html):
> The @private tag is not inherited by child members. For example, if the @private tag is added to a namespace, members of the namespace can still appear in the generated output; because the namespace is private, the members' namepath will not include the namespace.

vakila:
:wave: Hi @danswick @mourner just checking in on this - are we still planning to move this ahead? Anything we can do to help?

mourner:
👋 @danswick any updates?

danswick:
@mourner and @vakila my apologies for letting this get so stale. I am planning to catch this branch up and get this PR into a reviewable state this week!

src/source/canvas_source.js
```diff
@@ -167,10 +167,6 @@ class CanvasSource extends ImageSource {
      * @method setCoordinates
      * @instance
      * @memberof CanvasSource
-     * @param {Array<Array<number>>} coordinates Four geographical coordinates,
```
danswick: I don't think we need to specify params for inherited methods.

src/source/image_source.js
```diff
@@ -137,7 +137,7 @@ class ImageSource extends Evented implements Source {
      * Updates the image URL and, optionally, the coordinates. To avoid having the image flash after changing,
      * set the `raster-fade-duration` paint property on the raster layer to 0.
      *
-     * @param {Object} options
+     * @param {Object} options [options]
```
danswick:
```suggestion
     * @param {Object} options Options object.
```

src/source/tile_cache.js
```diff
@@ -191,6 +191,8 @@ class TileCache {
     /**
      * Remove entries that do not pass a filter function. Used for removing
      * stale tiles from the cache.
+     *
+     * @param filterFn filter function to check.
```
danswick: I'm not really sure how to describe this param.

src/ui/map.js
```diff
@@ -847,6 +851,7 @@ class Map extends Camera {
 
     /**
      * Returns true if the map is rotating due to a camera animation or user gesture.
+     * @returns {Boolean} True if hte map is rotating.
```
danswick:
```suggestion
     * @returns {Boolean} True if the map is rotating.
```

danswick:
Ok, I think this PR is ready for a closer review. For the reviewer:

- The goal of this PR is to get us to a "clean" JSDoc lint so that the core GL JS team can start incorporating JSDoc checks into their normal workflow. 
- It would be great to get feedback on the specific documentation changes in the diff. What is factually incorrect? Unclear? Redundant? 
- I left a smattering of comments throughout when I came across an edge case or something that just seemed awkward. If you could pay particular attention to those comments, that would be very helpful. 

Please let me know if anything here is unclear along the way!

src/source/image_source.js
```diff
@@ -137,7 +137,7 @@ class ImageSource extends Evented implements Source {
      * Updates the image URL and, optionally, the coordinates. To avoid having the image flash after changing,
      * set the `raster-fade-duration` paint property on the raster layer to 0.
      *
-     * @param {Object} options
+     * @param {Object} options Options object.
```
arindam1993: We have the type annotations, should we use that instead of `Object`?

src/ui/handler/scroll_zoom.js
arindam1993: Theres a lot of magic functions and numbers this goes through, its hard to really describe a unit.
maybe something like
`The rate used to scale the mouse wheel deltaY to a zoom value`

src/source/canvas_source.js
karimnaaji: Is this information expressed anywhere else after we delete it?

src/source/tile_cache.js
karimnaaji: What about describing what the function signature would be? Worthwhile information in that case would be what it should return and in each case how it would impact the filtering.

It's taking as input a `Tile`, and the function `filterFn` controls whether the tile would be filtered _out_ during the filter process. What about:

```
@param filterFn filter function, takes as input a Tile, and controls whether the tile would be filtered out. If the function returns false, the tile would be filtered out, it won't otherwise.
```

Feel free to make it more succinct.

src/ui/handler/scroll_zoom.js
```diff
@@ -79,15 +79,15 @@ class ScrollZoomHandler {
...
      */
     setZoomRate(zoomRate: number) {
         this._defaultZoomRate = zoomRate;
     }
 
     /**
      * Set the zoom rate of a mouse wheel
-     * @param {number} [wheelZoomRate = 1/450]
+     * @param {number} [wheelZoomRate=1/450] Zoom rate in some unit.
```
karimnaaji: nit: It might be misleading to see that type of information, I'd rather leave it out than abstract, but that's personal preference.

src/ui/marker.js
```diff
@@ -418,7 +419,7 @@ export default class Marker extends Evented {
 
     /**
      * Get the marker's offset.
-     * @returns {Point}
+     * @returns {Point} The marker's screen coordinates in pixels.
```
karimnaaji: nit: Logical pixels.

src/ui/marker.js
```diff
@@ -598,7 +599,7 @@ export default class Marker extends Evented {
 
     /**
      * Returns the current `pitchAlignment` property of the marker.
-     * @returns {string}
+     * @returns {string} The current pitch alignment of the marker.
```
karimnaaji: nit: Might be worth mentioning degrees.

src/ui/handler/scroll_zoom.js
danswick: @karimnaaji oh yeah, I left this in intentionally because I have no idea what the unit is :) I was hoping folks from the GL JS team would help fill in the gaps!

src/geo/transform.js
```diff
@@ -206,10 +206,10 @@ class Transform {
...
+     * @param {number} options.tileSize Tile size, expressed in pixels.
```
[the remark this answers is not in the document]
danswick: @karimnaaji I'm don't think our end users will understand what we mean by "logical pixels." How about "screen pixels" instead?

src/source/image_source.js
danswick: I think that's a good long-term goal, but I think it's out of scope for this change. Or is there a technique you know of to link type annotations automatically?

src/geo/transform.js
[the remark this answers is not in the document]
karimnaaji: That would work I think, I've seen Mapkit make use of CSS pixels to denote the same, anything that suggests invariance over pixel resolution/retina displays is good.

src/ui/handler/scroll_zoom.js
karimnaaji: The way this is used is intended to be a scalar applied per frame tick to the delta value of a scroll wheel event, which can be in a few different units like DOM_DELTA_LINE, or DOM_DELTA_PIXELS.

Maybe we can speak of _how_ it influences the behavior and link to the different potential event units:
```
A scaling factor applied per frame tick to the delta of a WheelEvent in WheelEvent.deltaMode unit, higher is faster
```
and refer to https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode or other.

danswick:
Ah thanks for catching that, @mourner! I reverted earlier but apparently failed to update my local yarn installation correctly. It should be fixed now!

PR #8961, opened 2019-11-09
Add eslint config for JSDoc

danswick:
Per https://github.com/mapbox/mapbox-gl-js/issues/8819, this PR adds a number of JSDoc rules to this repo's eslint configuration. All of these rules are currently set to `warn` rather than `error`, so tests will continue to pass even if some JSDoc isn't in compliance with its corresponding eslint rules. 

These config changes include the following rules: 

- `check-param-names`: Ensures that parameter names in JSDoc match those in the function declaration.
- `require-param`: Requires that all function parameters are documented.
- `require-param-description`: Requires that @param tag has description value.
- `require-param-name`: Requires that all function parameters have name.
- `require-returns`: Requires returns are documented.
- `require-returns-description`: Requires that @returns tag has description value.

These config changes explicitly **exclude** the following rules, per discussion in #8819:

- `require-jsdoc`: There are a lot of internal methods that don't include JSDoc currently. It would be a tremendous effort to document all of them now.
- `settings.jsdoc.ignorePrivate`: We should ignore private methods that include incomplete JSDoc for now.

These changes are an initial step toward more complete JSDoc testing for this project. I felt it would be a bit easier to digest if I broke up the eslint rule changes and their corresponding JSDoc fixes into separate PRs to make reviewing easier and help keep this initiative moving. 

These are the steps I'm currently envisioning, but am hoping contributors will chime in with feedback and suggestions: 

1. Merge these eslint configuration changes and begin getting linting feedback on JSDoc right away.
2. Address all JSDoc linting issues with minimal fixes. See my draft PR here: https://github.com/mapbox/mapbox-gl-js/pull/8960. There are a number of open questions in this PR that would be great to discuss.
3. Build the full docs page and sweep for legibility and completeness. Incorporate as much of the [feedback](https://github.com/mapbox/mapbox-gl-js-docs/issues) we've been getting from the feedback module as possible. 
4. Once all eslint warnings are addressed, switch from `warn` to `fail` to ensure new contributions continue to maintain the documentation.
5. Finally, require JSDoc for all methods, including private methods (long term)

I would love @mapbox/gl-js's feedback on 1) the proposed eslint config changes and 2) the proposed process for getting to fully tested JSDoc. 

cc @mapbox/docs

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -75,12 +75,12 @@
     "no-multiple-empty-lines": [ "error", { 
         "max": 1
     }],
-    "jsdoc/check-param-names": 1,
-    "jsdoc/require-param": 1,
-    "jsdoc/require-param-description": 1,
-    "jsdoc/require-param-name": 1,
-    "jsdoc/require-returns": 1,
-    "jsdoc/require-returns-description": 1
+    "jsdoc/check-param-names": "warn",
+    "jsdoc/require-param": "warn",
+    "jsdoc/require-param-description": "warn",
+    "jsdoc/require-param-name": "warn",
+    "jsdoc/require-returns": "warn",
+    "jsdoc/require-returns-description": "warn"
   },
   "settings": {
     "jsdoc":{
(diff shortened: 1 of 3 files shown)


ryanhamley:
This looks really great @danswick! Thanks for taking this on. As @mourner noted in #8960, you need to update Yarn to a newer version to avoid all the `integrity` fields changing in `yarn.lock`.

I think the rules you've included here and the general approach you've outlined make sense. I'd like to get feedback from others on the team before approving it though.

vakila:
Seconded (thirded?) that this looks awesome! Big :+1:

PR #8966, opened 2019-11-11
make vertex and index buffers resizable

ansis:
This isn't used yet but is one small step towards fast updates to layout properties. (I'm pulling the out of a old, unfinished, stale branch).

## Launch Checklist

 - [n/a] post benchmark scores

### DIFF
--- a/src/gl/index_buffer.js
+++ b/src/gl/index_buffer.js
@@ -6,15 +6,19 @@ import type {TriangleIndexArray, LineIndexArray, LineStripIndexArray} from '../d
 import type Context from '../gl/context';
 
 class IndexBuffer {
+    length: number;
     context: Context;
     buffer: WebGLBuffer;
     dynamicDraw: boolean;
+    resizable: boolean;
 
-    constructor(context: Context, array: TriangleIndexArray | LineIndexArray | LineStripIndexArray, dynamicDraw?: boolean) {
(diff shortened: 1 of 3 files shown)


mourner:
Given that this is a pretty small change, should we include it when we actually use it? Or do you anticipate taking advantage of this relatively soon?

src/gl/index_buffer.js
```diff
@@ -35,12 +39,17 @@ class IndexBuffer {
...
         // The right VAO will get this buffer re-bound later in VertexArrayObject#bind
         // See https://github.com/mapbox/mapbox-gl-js/issues/5620
         this.context.unbindVAO();
         this.bind();
-        gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, 0, array.arrayBuffer);
+        if (this.length === array.length) {
+            assert(this.dynamicDraw || this.resizable);
+            gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, 0, array.arrayBuffer);
+        } else {
+            assert(this.resizable);
```
arindam1993: Is the resizable boolean really necessary, I assume it is so we can be more explicit about when we want to pay the performance cost of doing a full bufferData.
But could we also do that by adding a warning thats stripped away in production?

kkaefer: Is there any advantage to having a buffer non-resizable? Why aren't all buffers resizable by default? We don't necessarily need to resize them.

ansis:
Closing for now. I'll try to merge it as part of a larger pr.

Issue #8912, opened 2019-10-26
Is it possible to use THREE.LOD?


AmeliaWang93:
I try to use THREE.LOD, but it doesn't change after zooming.  
```html
<!DOCTYPE html>
<html>
<head>
    <meta charset='utf-8'/>
    <title>Add a 3D model</title>
    <meta name='viewport' content='initial-scale=1,maximum-scale=1,user-scalable=no'/>
    <script src='https://api.tiles.mapbox.com/mapbox-gl-js/v1.4.1/mapbox-gl.js'></script>
    <link href='https://api.tiles.mapbox.com/mapbox-gl-js/v1.4.1/mapbox-gl.css' rel='stylesheet'/>
    <style>
        body {
            margin: 0;
            padding: 0;
        }

        #map {
            position: absolute;
            top: 0;
            bottom: 0;
            width: 100%;
        }
    </style>
</head>
<body>

<script src='https://unpkg.com/three@0.106.2/build/three.min.js'></script>

<div id='map'></div>
<script>
    mapboxgl.accessToken = '<your token>';
    var map = window.map = new mapboxgl.Map({
        container: 'map',
        style: 'mapbox://styles/mapbox/light-v10',
        zoom: 18,
        center: [148.9819, -35.3981],
        pitch: 60,
        antialias: true // create the gl context with MSAA antialiasing, so custom layers are antialiased
    });

    // parameters to ensure the model is georeferenced correctly on the map
    var modelOrigin = [148.98190, -35.39847];
    var modelAltitude = 0;
    var modelRotate = [Math.PI / 2, 0, 0];

    var modelAsMercatorCoordinate = mapboxgl.MercatorCoordinate.fromLngLat(modelOrigin, modelAltitude);

    // transformation parameters to position, rotate and scale the 3D model onto the map
    var modelTransform = {
        translateX: modelAsMercatorCoordinate.x,
        translateY: modelAsMercatorCoordinate.y,
        translateZ: modelAsMercatorCoordinate.z,
        rotateX: modelRotate[0],
        rotateY: modelRotate[1],
        rotateZ: modelRotate[2],
        /* Since our 3D model is in real world meters, a scale transform needs to be
         * applied since the CustomLayerInterface expects units in MercatorCoordinates.
         */
        scale: modelAsMercatorCoordinate.meterInMercatorCoordinateUnits()
    };

    var THREE = window.THREE;

    // configuration of the custom layer for a 3D model per the CustomLayerInterface
[...]

Issue #5563, opened 2017-10-31
Further distances on pitched heatmaps "flicker" on every frame


lbud:
Pitching maps with heatmap layers cause "far" kernels to change on every frame (kind of flicker, in a blurry heatmap sort of way).

![heatmap-blur](https://user-images.githubusercontent.com/3170312/32201865-55979158-bd96-11e7-9cc9-6a03a2fdc6bc.gif)

Reproducible on any heatmap, including the [examples page](https://www.mapbox.com/mapbox-gl-js/example/heatmap-layer/).
cc @mourner

mourner:
Wow, not sure what's going on there....

lbud:
@mourner the effect sort of reminds me of something I ran into in fill-extrusion layers in gl-native, where sometimes the depth buffer resolution was too low (https://github.com/mapbox/mapbox-gl-native/pull/8431#issuecomment-296860060), causing jagged edges on the sides of extrusions; we fixed this by setting the near clip plan in the fill-extrusion layer projection matrix to 100 so as not to waste a lot of precision on shapes extremely close to the camera (https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/72278bfe94ce0d19b515bb92e9d0be88eb53aa89). Not sure if this is relevant to what's going on here, but 🤷‍♀️

wysisoft:
I don't think this is as much a "further distances" problem as much as a "small radius" problem. Radii above 5 seems to behave well. Radii 3 and under seem to flicker.

pathmapper:
Agree with @wysisoft, there is no need to pitch a heatmap to trigger the bug. This GIF is just the official example zoomed out so that `heatmap-radius = 2`.

![heatmap-flicker](https://user-images.githubusercontent.com/20856381/68531859-34038b00-0317-11ea-8ba0-43c81158a4b5.gif)

ryanhamley:
Thanks for pointing out what's going on here @pathmapper and @wysisoft

I can confirm that this is reproducible in our debug heatmap page as well, if you lower the `heatmap-radius` to <= 2. I'm not very familiar with the heatmap code though. @mourner does this new information point to a potential cause? could it be related to texture filtering a la https://github.com/mapbox/mapbox-gl-js/pull/8738?

mpulkki-mapbox:
[...]

PR #8975, opened 2019-11-13
LOD support for tile coverage

mpulkki-mapbox:
Visible tiles on the screen are currently found by projecting the viewport quad on the map and iterating through intersecting tiles with a scanline algorithm. It is an efficient solution when the map is viewed from relatively small pitch angles, but the algorithm does not scale well with the visible land coverage.

This pull request adds a support for streaming in tiles of different zoom levels based on their distance to map center. It is unnecessary to stream and render distant tiles in full detail if they cover only small portion of the screen. The immediate effect of this change is to remove the pitch angle limitation and allow arbitrary camera orientations (although camera behavior remains unchanged).

The new tileCoverage implementation traverses the tile quadtree depth-first and checks visibility of each tile against the camera frustum. The algorithm is configured to choose 6x6 tile area of target zoom level before switching to lower zoom tiles. The max allowed zoom level difference between adjacent tiles is 1. 

There might be a performance hit especially on higher zoom levels so new benchmark tests are required. 

#### TODO:
 - [x] new unit tests for Aabb and frustum primitives
 - [x] new unit tests for tileCoverage with high pitch angle
 - [x] Benchmark the performance difference between old and new implementation

#### Future improvements:
 - allow tile overlap to reduce the number of visible tiles even more

Related issues: https://github.com/mapbox/mapbox-gl-js/issues/3731, https://github.com/mapbox/mapbox-gl-native/issues/9037,
https://github.com/mapbox/mapbox-gl-native/issues/6908

LOD in action:
lod_high_pitch
lod_high_pitch_2


src/geo/transform.js
```diff
@@ -547,7 +615,7 @@ class Transform {
         // (the distance between[width/2, height/2] and [width/2 + 1, height/2])
         const halfFov = this._fov / 2;
         const groundAngle = Math.PI / 2 + this._pitch;
-        const topHalfSurfaceDistance = Math.sin(halfFov) * this.cameraToCenterDistance / Math.sin(Math.PI - groundAngle - halfFov);
+        const topHalfSurfaceDistance = Math.sin(halfFov) * this.cameraToCenterDistance / Math.sin(clamp(Math.PI - groundAngle - halfFov, 0.01, Math.PI - 0.01));
```
mpulkki-mapbox: Projection matrix generation was broken with high pitch values. `topHalfSurfaceDistance` might negative value causing negative far plane distance. This could be seen as either disappearing ground or as floating symbols.

andrewharvey:
I haven't had a chance to fully review or test, but have you thought about loading in higher zoom level tiles for tiles closer to the camera (which are stretched beyond their native resolution at low pitch)? From you screenshot it doesn't look like this happens.

mpulkki-mapbox:
> I haven't had a chance to fully review or test, but have you thought about loading in higher zoom level tiles for tiles closer to the camera (which are stretched beyond their native resolution at low pitch)? From you screenshot it doesn't look like this happens.

This is exactly what's happening: quality of distant tiles is reduced by loading lower zoom level versions of them. In other words, tiles near the camera are from a higher zoom level than the distant ones. The visible LOD curve on the screenshots is quite gentle and the maximum possible zoom level is still dictated by the map zoom,

andrewharvey:
> This is exactly what's happening: quality of distant tiles is reduced by loading lower zoom level versions of them. In other words, tiles near the camera are from a higher zoom level than the distant ones. The visible LOD curve in the screenshots is quite gentle and the maximum possible zoom level is still dictated by the map zoom,

I meant higher than the zoom level of the map. The distant tiles would come from lower zoom levels as I can see in your screenshot, but what about near tiles closer to the camera should be using tiles from higher zoom levels?

What is normally 512px pixels gets stetched out to much >512px near the screen at low pitches.

jingsam:
> > This is exactly what's happening: quality of distant tiles is reduced by loading lower zoom level versions of them. In other words, tiles near the camera are from a higher zoom level than the distant ones. The visible LOD curve in the screenshots is quite gentle and the maximum possible zoom level is still dictated by the map zoom,
> 
> I meant higher than the zoom level of the map. The distant tiles would come from lower zoom levels as I can see in your screenshot, but what about near tiles closer to the camera should be using tiles from higher zoom levels?
> 
> What is normally 512px pixels gets stetched out to much >512px near the screen at low pitches.

I suggest to not over engineer this feature. Users may observe mismatch features near screen. For example, we style road at zoom 10 with 2px, zoom 11 with 4px. If we mix zoom 10 and zoom 11 tiles in one map viewport, that may cause inconsistency.

@mpulkki-mapbox provide a good solution to reduce tile request when map is pitched a large angle. Shall we just simply frog the skyline to total avoid load tiles far away screen?

image

andrewharvey:
> I suggest to not over engineer this feature. Users may observe mismatch features near screen. For example, we style road at zoom 10 with 2px, zoom 11 with 4px. If we mix zoom 10 and zoom 11 tiles in one map viewport, that may cause inconsistency.

So I think even with this the style's zoom would be the same across all tiles in the viewport (the center tile zoom) just the vector tile would be from a range of zoom levels. @mpulkki-mapbox is that correct?

> @mpulkki-mapbox provide a good solution to reduce tile request when map is pitched a large angle. Shall we just simply frog the skyline to total avoid load tiles far away screen?

Absolutely I agree this is an awesome and long awaited feature. I just wanted to see if the similar case of near viewport tiles has been considered, not saying it has to be addressed in this PR.

mourner:
Super pumped about this finally getting in! A feature I waited for so long. 🙏 

> For example, we style road at zoom 10 with 2px, zoom 11 with 4px. If we mix zoom 10 and zoom 11 tiles in one map viewport, that may cause inconsistency.

This is something we haven't fully decided on but I'm personally leaning towards using a single zoom value for style evaluation across all visible tiles, regardless of what zoom they come from. So in the case above the line width would be the same, but there would be a chance of some roads cutting abruptly if they're not present on lower zooms. 

This seems OK to me because it's consistent with how under/overzooming works while tiles are loading — e.g. you have tiles on z10, you zoom in to z11 fast, and while you wait for z11 tiles to load, you see z10 tiles scaled and styled according to z11.

ivovandongen:
@jingsam 

>@mpulkki-mapbox provide a good solution to reduce tile request when map is pitched a large angle. Shall we just simply frog the skyline to total avoid load tiles far away screen?

I think this is could be done additionally to LOD tile cover. This suits some use cases very well, where in others you might rather have a "clean" horizon in the distance.

src/geo/transform.js
```diff
@@ -274,17 +274,85 @@ class Transform {
...
+                wrap,
+                fullyVisible: false
+            };
+        };
+
+        // Do a depth-first traversal to find visible tiles and proper levels of detail
+        const stack = [];
+        const result = [];
+        const maxZoom = z;
+        const zOffset = (options.reparseOverscaled ? actualZ : z) - z;
```
astojilj: This relates to @mourner 's comment here: https://github.com/mapbox/mapbox-gl-js/pull/8975#issuecomment-553787111

Using the constant overscaleID (so the same overscaleID and different canonical.z) for all tiles returned by coveringTiles() works but it produces a bit too many symbols in background (for vector tile cover). I think this would be explained better with screenshots showing both alternatives.

mpulkki-mapbox:
> So I think even with this the style's zoom would be the same across all tiles in the viewport (the center tile zoom) just the vector tile would be from a range of zoom levels. @mpulkki-mapbox is that correct?

Currently style zoom level is using LOD as well. I'm prototyping the use of constant zoom level across all tiles.

mpulkki-mapbox:
Some visible color tint between satellite images of different zoom.
lod_satellite_tint

mpulkki-mapbox:
I went through some of the styles and noticed that there are some aliasing issues with the heatmap when pitch is big enough
heatmap_aliasing

astojilj:
> I went through some of the styles and noticed that there are some aliasing issues with the heatmap when pitch is big enough

@mpulkki-mapbox https://github.com/mapbox/mapbox-gl-js/issues/5563#issue-269795931 ?

mpulkki-mapbox:
Blocked by #8979

mourner:
@mpulkki-mapbox any next steps I can help with to move this PR forward? Let's land it soon!

mpulkki-mapbox:
@mourner: This PR should be good to go if we keep the maximum pitch angle unchanged. I've had a bad habit of trying to solve other open issues like the symbol placement before landing this one :). I'd like to add few unit tests and run benchmarks first though.

mpulkki-mapbox:
I've marked this pull request as ready for review. The maximum pitch angle is left unchanged so this PR has no immediate impact on behavior. `coveringTiles` should provide similiar results as the original implementation when pitch is < 60. Open questions and problems presented earlier in this PR could be solved on later tickets as we prepare to fully support the bigger pitch angle.

There is a small performance regression in `LayerBackground` benchmark though. The test adds 50 background layers to the map resulting in 50 calls being made to `coveringTiles` which is the most taxing code in otherwise simple logic. The absolute increment in code execution time is quite small but it gets emphasized in this specific benchmark. It should be noted that the small performance hit becomes visible only if a style has dozens of background layers. There are also plenty of spots to optimize in the background layer rendering if it ends up being a bottleneck.

image

test/unit/util/tile_cover.test.js
```diff
@@ -1,95 +0,0 @@
...
-                    new MercatorCoordinate(0, 0.5)
-                ],
-                res = tileCover(z, coords, z);
-            t.deepEqual(res, [new OverscaledTileID(2, 0, 2, 0, 1)]);
-            t.end();
-        });
-
-        t.test('calculates tile coverage at w > 0', (t) => {
-            const z = 2,
-                coords = [
```
astojilj: Could we repurpose the same tests to test Transform.coveringTiles (seems like a square viewport and no pitch nor heading)? Or, probably better, have other tests, covering different value of wrap (< 0, > 0...), coverage across meridians?

src/geo/transform.js
```diff
@@ -276,17 +276,88 @@ class Transform {
...
+                stack.push(newRootTile(-i));
+                stack.push(newRootTile(i));
+            }
+        }
+
+        stack.push(newRootTile(0));
+
+        // Stream position will determine the "center of the streaming",
+        // ie. where the most detailed tiles are loaded.
+        const streamPos = [centerPoint.x, centerPoint.y, 0];
```
mourner: Nit: you could simply use `centerPoint` below instead of defining `streamPos` for the same thing, for clarity.

src/geo/transform.js
```diff
@@ -276,17 +276,88 @@ class Transform {
...
+            let fullyVisible = it.fullyVisible;
+
+            // Visibility of a tile is not required if any of its ancestor if fully inside the frustum
+            if (!fullyVisible) {
+                const intersectResult = it.aabb.intersects(cameraFrustum);
+
+                if (intersectResult === 'none')
+                    continue;
+
+                fullyVisible = intersectResult === 'contains';
```
mourner: Nit: it might be a little faster / more idiomatic to return numbers instead of strings for `intersects` — e.g. 0 for "none", 1 for "intersects" and 2 for "contains".

src/geo/transform.js
```diff
@@ -276,17 +276,88 @@ class Transform {
...
+            }
+
+            const distanceXY = it.aabb.distanceXY(streamPos);
+            const longestDim = Math.max(Math.abs(distanceXY[0]), Math.abs(distanceXY[1]));
```
mourner: Since we don't need distanceXY values separately, replacing `distanceXY` with `longestDim` so that it returns a single value without allocating a two-element array unnecessarily will be faster.

src/geo/transform.js
```diff
@@ -276,17 +276,88 @@ class Transform {
...
+                result.push({
+                    tileID: new OverscaledTileID(it.zoom === maxZoom ? overscaledZ : it.zoom, it.wrap, it.zoom, x, y),
+                    distanceSq: vec2.sqrLen([streamPos[0] - 0.5 - x, streamPos[1] - 0.5 - y])
+                });
+                continue;
+            }
+
+            for (let i = 0; i < 4; i++) {
+                const childX = (x << 1) + (i % 2);
+                const childY = (y << 1) + Math.floor(i / 2);
```
mourner: Nit for consistency:

```suggestion
                const childY = (y << 1) + (i >> 1);
```

test/unit/util/tile_cover.test.js
mpulkki-mapbox: This is a very good point. I've included new tests that should cover these cases.

test/unit/geo/transform.test.js
```diff
@@ -153,6 +153,103 @@ test('transform', (t) => {
...
+            new OverscaledTileID(8, 0, 8, 146, 74),
+            new OverscaledTileID(8, 0, 8, 146, 73)
+        ]);
+
+        transform.zoom = 2;
+        transform.pitch = 0;
+        transform.bearing = 0;
+        transform.resize(300, 300);
+        t.test('calculates tile coverage at w > 0', (t) => {
+            transform.center = {lng: 630.01, lat: 0.01};
```
astojilj: Nice. LGTM.

mpulkki-mapbox:
> good to go. Just curious, did `BackgroundLayer` bench performance improve after the recent updates?

I didn't see any notable difference on my test machine.

joedjc:
Is there any chance this will enable this...https://github.com/mapbox/mapbox-gl-js/issues/8912?

tangjuncheng1986:
How to use the pitch >60 in the branch?

src/geo/transform.js
```diff
@@ -276,17 +276,88 @@ class Transform {
...
-            this.pointCoordinate(new Point(this.width, 0)),
-            this.pointCoordinate(new Point(this.width, this.height)),
-            this.pointCoordinate(new Point(0, this.height))
-        ];
-        return tileCover(z, cornerCoords, options.reparseOverscaled ? actualZ : z, this._renderWorldCopies)
-            .sort((a, b) => centerPoint.dist(a.canonical) - centerPoint.dist(b.canonical));
+        const centerPoint = new Point(numTiles * centerCoord.x, numTiles * centerCoord.y);
+        const cameraFrustum = Frustum.fromInvProjectionMatrix(this.invProjMatrix, this.worldSize, z);
+
+        if (this.pitch <= 60.0)
```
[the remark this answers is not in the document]
andrewharvey: @astojilj @mpulkki-mapbox sorry I'm a bit late here, but any reason for not doing LOD for pitch <= 60? Even at pitch 60 especially when you have a bearing like 45 degrees and using 256x256 satellite tiles it feels like some of those tiles right at furthest point away probably could be LOD.

andrewharvey:
@tangjuncheng1986 see the issue tracking that over at #3731, per https://github.com/mapbox/mapbox-gl-js/issues/3731#issuecomment-575593404 looks like we might just need to change the maximum pitch limit.

PR #8979, opened 2019-11-15
Change the type of tile id key to string

mpulkki-mapbox:
### Description

I found a possible bug or misuse in the key calculation of OverscaledTileID. Canonical z is not taken into account when computing the key hash, but it is used in the equality comparison of the class. This will result in hash collisions and undefined behavior if the key is used as a hashmap key.

This bug appears when trying to load tiles from different zoom levels but using constant overscaledZ across all tiles. At one point the code might be requesting a tile (w, oz, z, x, y) `[2, 1, 0, 0, 0]` with key 513. When LOD kicks in this area would be split into 4 smaller tiles: `[2, 1, 1, 0, 0], [2, 1, 1, 1, 0], [2, 1, 1, 0, 1], [2, 1, 1, 1, 1]`. We now have two different tile ids `[2, 1, 0, 0, 0]` and `[2, 1, 1, 0, 0]` that both have the same key 513. This will break the tile cache resulting in wrong tile being rendered.

image

This bug/misuse is not present in gl-js master because all tiles have same zoom level. Gl-native should work as well because the CanonicalTileID is used properly with hashmaps.

A side note: source_cache unit tests are using wrong OverscaledTileIDs to check equality in multiple places. The following snippet from the unit tests and it passes fine.
```
                t.deepEqual(sourceCache.getRenderableIds(), [
                    new OverscaledTileID(16, 0, 16, 8192, 8192).key,
                    new OverscaledTileID(16, 0, 16, 8191, 8192).key,
                    new OverscaledTileID(16, 0, 16, 8192, 8191).key,
                    new OverscaledTileID(16, 0, 16, 8191, 8191).key
                ]);
```
Same test passes with following modifications to z values.
```
                t.deepEqual(sourceCache.getRenderableIds(), [
                    new OverscaledTileID(16, 0, 16, 8192, 8192).key,
                    new OverscaledTileID(16, 0, 15, 8191, 8192).key,
                    new OverscaledTileID(16, 0, 14, 8192, 8191).key,
                    new OverscaledTileID(16, 0, 14, 8191, 8191).key
                ]);
```
`sourceCache.getRenderableIds()` actually returns a list of ids that have canonical zoom level of 14 which is not captured by these tests.

### Possible fix
There are at least two possible solutions to fix this:
1. Fix the key generation `this.key = calculateKey(wrap, overscaledZ, z, x, y);`. *This is the approach used in this pull request*.
2. The key generation is not inherently broken, but its usage with hashmaps is. The key can be used to find the right hash bucket, but it doesn't work as an identifier without an equality operator. The fix would be to use OverscaledTileID as a hashmap key like it's done in gl-native. 

### TODO
 - [x] Unit tests have to be updated

### DIFF
--- a/src/source/source_cache.js
+++ b/src/source/source_cache.js
@@ -44,7 +44,7 @@ class SourceCache extends Evented {
     _source: Source;
     _sourceLoaded: boolean;
     _sourceErrored: boolean;
-    _tiles: {[any]: Tile};
+    _tiles: {[string]: Tile};
     _prevLng: number | void;
     _cache: TileCache;
     _timers: {[any]: TimeoutID};
@@ -52,9 +52,9 @@ class SourceCache extends Evented {
     _maxTileCacheSize: ?number;
     _paused: boolean;
     _shouldReloadOnResume: boolean;
-    _coveredTiles: {[any]: boolean};
+    _coveredTiles: {[string]: boolean};
     transform: Transform;
-    _isIdRenderable: (id: number, symbolLayer?: boolean) => boolean;
+    _isIdRenderable: (id: string, symbolLayer?: boolean) => boolean;
     used: boolean;
     _state: SourceFeatureState;
 
@@ -179,25 +179,25 @@ class SourceCache extends Evented {
     /**
      * Return all tile ids ordered with z-order, and cast to numbers
      */
-    getIds(): Array<number> {
-        return Object.keys(this._tiles).map(Number).sort(compareKeyZoom);
+    getIds(): Array<string> {
+        return (Object.values(this._tiles): any).map((tile: Tile) => tile.tileID).sort(compareTileId).map(id => id.key);
     }
 
-    getRenderableIds(symbolLayer?: boolean): Array<number> {
-        const ids = [];
+    getRenderableIds(symbolLayer?: boolean): Array<string> {
+        const renderables: Array<Tile> = [];
         for (const id in this._tiles) {
-            if (this._isIdRenderable(+id, symbolLayer)) ids.push(+id);
+            if (this._isIdRenderable(id, symbolLayer)) renderables.push(this._tiles[id]);
         }
         if (symbolLayer) {
-            return ids.sort((a_, b_) => {
-                const a = this._tiles[a_].tileID;
-                const b = this._tiles[b_].tileID;
+            return renderables.sort((a_: Tile, b_: Tile) => {
+                const a = a_.tileID;
+                const b = b_.tileID;
                 const rotatedA = (new Point(a.canonical.x, a.canonical.y))._rotate(this.transform.angle);
                 const rotatedB = (new Point(b.canonical.x, b.canonical.y))._rotate(this.transform.angle);
                 return a.overscaledZ - b.overscaledZ || rotatedB.y - rotatedA.y || rotatedB.x - rotatedA.x;
-            });
+            }).map(tile => tile.tileID.key);
         }
-        return ids.sort(compareKeyZoom);
+        return renderables.map(tile => tile.tileID).sort(compareTileId).map(id => id.key);
     }
 
     hasRenderableParent(tileID: OverscaledTileID) {
@@ -208,7 +208,7 @@ class SourceCache extends Evented {
         return false;
     }
 
-    _isIdRenderable(id: number, symbolLayer?: boolean) {
+    _isIdRenderable(id: string, symbolLayer?: boolean) {
         return this._tiles[id] && this._tiles[id].hasData() &&
             !this._coveredTiles[id] && (symbolLayer || !this._tiles[id].holdingForFade());
     }
@@ -226,7 +226,7 @@ class SourceCache extends Evented {
         }
     }
 
-    _reloadTile(id: string | number, state: TileState) {
+    _reloadTile(id: string, state: TileState) {
         const tile = this._tiles[id];
 
         // this potentially does not address all underlying
@@ -245,7 +245,7 @@ class SourceCache extends Evented {
         this._loadTile(tile, this._tileLoaded.bind(this, tile, id, state));
     }
 
-    _tileLoaded(tile: Tile, id: string | number, previousState: TileState, err: ?Error) {
+    _tileLoaded(tile: Tile, id: string, previousState: TileState, err: ?Error) {
         if (err) {
             tile.state = 'errored';
             if ((err: any).status !== 404) this._source.fire(new ErrorEvent(err, {tile}));
@@ -313,7 +313,7 @@ class SourceCache extends Evented {
     /**
      * Get a specific tile by id
      */
-    getTileByID(id: string | number): Tile {
+    getTileByID(id: string): Tile {
         return this._tiles[id];
     }
 
@@ -423,7 +423,7 @@ class SourceCache extends Evented {
         this._prevLng = lng;
 
         if (wrapDelta) {
-            const tiles = {};
+            const tiles: {[string]: Tile} = {};
             for (const key in this._tiles) {
                 const tile = this._tiles[key];
                 tile.tileID = tile.tileID.unwrapTo(tile.tileID.wrap + wrapDelta);
@@ -489,12 +489,12 @@ class SourceCache extends Evented {
         const retain = this._updateRetainedTiles(idealTileIDs, zoom);
 
         if (isRasterType(this._source.type)) {
-            const parentsForFading = {};
+            const parentsForFading: {[string]: OverscaledTileID} = {};
             const fadingTiles = {};
             const ids = Object.keys(retain);
             for (const id of ids) {
                 const tileID = retain[id];
-                assert(tileID.key === +id);
+                assert(tileID.key === id);
 
                 const tile = this._tiles[id];
                 if (!tile || tile.fadeEndTime && tile.fadeEndTime <= browser.now()) continue;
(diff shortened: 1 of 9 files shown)


src/source/tile_id.js
```diff
@@ -164,11 +164,11 @@ export class OverscaledTileID {
     }
 }
 
-function calculateKey(wrap: number, z: number, x: number, y: number) {
+function calculateKey(wrap: number, overscaledZ: number, z: number, x: number, y: number) {
     wrap *= 2;
     if (wrap < 0) wrap = wrap * -1 - 1;
     const dim = 1 << z;
-    return ((dim * dim * wrap + dim * y + x) * 32) + z;
+    return (((dim * dim * wrap + dim * y + x) * 32) + z) * 32 + overscaledZ;
```
ansis: I'm trying to figure out whether we have enough precision in javascript to safely represent this key. 2^53 - 1 is the maximum safe integer js numbers can hold. Each of the zoom components takes 5 bits which leaves 43 bits for the rest. Ignoring the world wrapping part would leave 21 bits for x and y.

It would be rare for this to be a problem but maybe we should switch to strings here to be safe. I'm not sure if there are perf implications.

Or at least `assert(z < 22)`.

mourner: > Or at least `assert(z < 22)`.

We can't — the default `maxZoom` is `22` but can be bumped further through `options` (e.g. to `24`).

We could try winning a few bytes by encoding the difference between `overscaledZ` and `z` rather than the full `overscaledZ`, which should have a smaller range of possible values.

src/source/tile_id.js
```diff
@@ -164,11 +164,26 @@ export class OverscaledTileID {
...
-    return ((dim * dim * wrap + dim * y + x) * 32) + z;
+    let id = `${dim * dim * wrap + dim * y + x}`;
+    // Always reserve two characters for both z values in order to
+    // keep the sorting behavior consistent
+    if (z < 10)
+        id += "0";
+    id += z;
+    if (overscaledZ < 10)
+        id += "0";
+    id += overscaledZ;
```
kkaefer: We could use non-base 10 number to string conversions, e.g. `overscaledZ.toString(36)`, same for the other numbers. Would be interesting to see if shorter strings can be interned better, or are stored directly inline in the JSValue node.

astojilj:
After rebasing this patch to ba2d04d0a40772d9d26ab6a60cef49c8b37cc2d8, existing benchmarks  don't show regression when running http://localhost:9966/bench/versions?compare=master.

There is a consistent performance improvement in LayerLine and LayerCircle benchmark and I find it is worth analyzing why:

Screen Shot 2019-12-03 at 22 52 24

Using the same approach as in [here](https://github.com/mapbox/mapbox-gl-js/pull/9050#issue-347369267), a large number of raster tiles rendering, with 6 x CPU throttle shows the same results for both cases, 18FPS.

astojilj:
> hooray emoji

Too early apparently. On different hardware, including iOS and desktops, there is still noticeable LayerRaster regression.

src/source/tile_id.js
```diff
@@ -164,11 +164,26 @@ export class OverscaledTileID {
     }
 }
 
-function calculateKey(wrap: number, z: number, x: number, y: number) {
+function calculateKey(wrap: number, overscaledZ: number, z: number, x: number, y: number): string {
```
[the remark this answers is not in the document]
mpulkki-mapbox: this is not the case anymore after the commit https://github.com/mapbox/mapbox-gl-js/commit/ba2d04d0a40772d9d26ab6a60cef49c8b37cc2d8

src/source/tile_id.js
```diff
@@ -164,11 +164,26 @@ export class OverscaledTileID {
...
+    if (z < 10)
+        id += "0";
+    id += z;
+    if (overscaledZ < 10)
+        id += "0";
+    id += overscaledZ;
+    return id;
+}
+
+export function tileIDKeyComparison(keyA: string, keyB: string): number {
```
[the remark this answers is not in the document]
mpulkki-mapbox: I'm not sure if we need this functionality at all. The only place where tiles are being sorted based on their zoom-level is `getIds()` and 'getRenderableIds()` in source_cache, but I've already modified that code to use tile ids instead of keys. Do we have any other requirements for the order of keys that I might have missed?

src/source/tile_id.js
mpulkki-mapbox: This is a very good point. I tried to keep functionality and logic in `calculateKey` as simple as possible because I noticed that `LayerRaster` benchmark is very sensitive to any changes in the key calculation function. These findings have resulted in some really good optimizations in `findLoadedParent` and `draw_raster` functions by @astojilj so they're not bottlenecks anymore. Non-base 10 encodings would definitely be better so I'll see what performance implications there might be.

mpulkki-mapbox:
@kkaefer thanks for pointing out that I should post benchmarks for performance comparisons! This is a very interesting subject that has lead us to very fascinating findings. As @astojilj already pointed out, there's no performance regression in most of the benchmarks nor in free form testing (like flying over SF manually). The biggest exception is `LayerRaster` benchmark that has been very sensitive to any changes in the `calculateKey` function. This is the consistent result I get when running the PR in chrome:

image

This lead us to findings that `findLoadedParent` function was being a bottleneck in `draw_raster`. After @astojilj's latest (merged) optimization patch https://github.com/mapbox/mapbox-gl-js/commit/ba2d04d0a40772d9d26ab6a60cef49c8b37cc2d8 the performance regression is much smaller already 

LayerRaster_chrome_optimizations

src/source/tile_id.js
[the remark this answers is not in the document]
kkaefer: Let's remove it if it's not used :)

mpulkki-mapbox:
Here's a one more comparison of how base-36 encoding might hurt the performance :). There is a clear hit especially in the minimum execution time.

String id **with** base-36 encoding:
LayerRaster_string_id_base36
String id **without** different base encoding:
LayerRaster_string_id

mpulkki-mapbox:
I wasn't satisfied with the LayerRaster performance so I did a quick round of optimization to the `draw_raster` function. The draw function always interpolates (opacity) between a tile and its closest loaded parent. This causes an extensive amount of calls made to `calculateKey` function in a worst case scenario where no parent tiles are loaded. For example: 500 visible tiles on a zoom level 16 might result in 500 * 15 * 2 = 15000 function calls made to `calculateKey` during `findLoadedParent`.

The optimized approach uses a lookup table to cache ancestor information during the tree traversal. This almost negates the performance regression.

LayerRaster_final_fix

src/source/source_cache.js
```diff
@@ -870,8 +926,10 @@ class SourceCache extends Evented {
 SourceCache.maxOverzooming = 10;
 SourceCache.maxUnderzooming = 3;
 
-function compareKeyZoom(a, b) {
-    return ((a % 32) - (b % 32)) || (b - a);
+function compareTileId(a: OverscaledTileID, b: OverscaledTileID): number {
+    const aWrap = Math.abs(a.wrap * 2) - +(a.wrap < 0);
+    const bWrap = Math.abs(b.wrap * 2) - +(b.wrap < 0);
```
mourner: Can you explain the `wrap` sorting logic? I can't quite get what's going on here...

src/source/source_cache.js
```diff
@@ -628,6 +647,43 @@ class SourceCache extends Evented {
         return retain;
     }
 
+    _updateLoadedParentTileCache() {
+        this._loadedParentTiles = {};
+
+        for (const tile of (Object.values(this._tiles): any)) {
```
mourner: Unfortunately `Object.values` is not supported in IE11. A `for in` loop will do though.

src/source/source_cache.js
```diff
@@ -628,6 +647,43 @@ class SourceCache extends Evented {
...
+                if (parentTile) {
+                    break;
+                }
+
+                currentId = parentId;
+            }
+
+            // Cache the result of this traversal to all newly visited tiles
+            for (let i = 0; i < path.length; i++) {
+                this._loadedParentTiles[path[i]] = parentTile;
```
mourner: Nit: `for (const key of path)`.

src/source/source_cache.js
mpulkki-mapbox: Different copies of the world are sorted based on their distance to the center where wrap is 0. This implementation converts wrap values to unsigned distances by reserving odd numbers for copies with negative wrap and even numbers for copies with positive wrap. The sorted order of wrap values would be: 0, -1, 1, -2, 2, -3, 3 and so on.

This is effectively the same as 
```
    wrap *= 2;
    if (wrap < 0) wrap = wrap * -1 - 1;
```

src/source/source_cache.js
mpulkki-mapbox: This is a good catch! Do we have any automatic ways of validating pull requests against different browsers?

mourner: @mpulkki-mapbox I explored options for linting things like this, but it's quite difficult — e.g. while you can catch things like `Object.values`, many other things IE11 doesn't support like `array.find` are impossible to catch because they're instance methods and linting plugins can't derive object types.

src/source/source_cache.js
mourner: Oh, so like zig-zag encoding. Gotcha. Maybe worth adding a comment since it's not obvious.

PR #8987, opened 2019-11-18
An option to use a feature property as ID for feature state

### LINKED ISSUE #6019: RFC: Feature identity for interactivity

mourner:
Closes #6019. An alternative proposal to #8448. Adds a source option to promote a specific property (either number or string) to `feature.id` for feature state, removing any limits on what type of values you can use (strings, floats, booleans, etc.):

```js
map.addSource('counties', {
    "type": "vector",
    "url": "mapbox://mapbox.82pkq93d",
    "promoteId": {"original": "COUNTY"}
});
...
map.setFeatureState({
    source: 'counties', 
    sourceLayer: 'original', 
    id: 'Marin County' // reference a feature by a county name for feature-state
}, {hover: true});
```

You can see how this works in the new `featurestate.html` debug page. Compared to #8448:

- You can use meaningful keys with feature state set/remove/get methods, making the API easier to use.
- Hides string to number hashing as an internal detail.

### TODO

- [x] Fix Flow errors
- [x] Extend to GeoJSON sources
- [x] Update unit tests 
- [x] Add spec docs

## Launch Checklist


mourner:
@asheemmamoowala good points!

> Your change breaks that by requiring the feature object to be modified so that id points to the `feature-state`-specific id instead.

Thinking more about it, should we consider overriding the feature object's `id`?  I can't think of a use case where it would be useful to keep the original feature `id` while opting to use a property for identification instead. Then we could rename the option to `promoteId` to make that override explicit.

It's also technically not a breaking change because the override only happens when you use that option, and if you do, passing the feature object to `setFeatureState` still works (even if the `id` is replaced). It will also be relatively easy to explain how this works in the docs.

In comparison, the option of having different code paths that treat `feature.id` differently depending on whether you also supply `properties` may be quite confusing for users.

Using an additional `featureStateID` property avoids the override, but also increases the API complexity, so is more prone to errors, and it will be more verbose.

asheemmamoowala:
> should we consider overriding the feature object's `id`? I can't think of a use case where it would be useful to keep the original feature id while opting to use a property for identification instead.

That works! This would prevent you from applying feature-state to an entire class of features - but that is not the intent of these APIs and can be done with data-driven styling already.

Developers will want to work with the `id` in it's original form. How would you propose `promoteId` would work in that case? Specifically - will the promoted property value or its hash be used/exposed in
- the `id` of features returned from `qRF`
- the input to `Map#(set/get/remove)FeatureState` methods

ryanbaumann:
cc/ @sansumbrella @jbranigan @cmtoomey

mourner:
> This would prevent you from applying feature-state to an entire class of features

Technically it doesn't prevent anything — feature `id` doesn't have to be unique. E.g. see the county example I committed in the PR — it highlights multiple counties with the same name.

> Developers will want to work with the `id` in it's original form. How would you propose `promoteId` would work in that case? Specifically - will the promoted property value or its hash be used/exposed in
> 
> * the `id` of features returned from `qRF`
> * the input to `Map#(set/get/remove)FeatureState` methods

`feature.id` of features returned from `qRF` will have the promoted value (e.g. a string property). `set/get/removeFeatureState` will work with `id` in the form of that promoted value. The main difference of this PR and #8448 is that here, the derived integer value is never exposed publicly — it's an internal implementation detail of `FeaturePositionMap` only used for hashing keys, similar to how v8 internally implements object lookup by key.

asheemmamoowala:
> see the county example I committed in the PR — it highlights multiple counties with the same name.

👍 

> feature.id of features returned from qRF will have the promoted value (e.g. a string property). set/get/removeFeatureState will work with id in the form of that promoted value.

👍

src/data/feature_index.js
```diff
@@ -146,12 +146,12 @@ class FeatureIndex {
...
                 params.layers,
                 styleLayers,
-                (feature: VectorTileFeature, styleLayer: StyleLayer) => {
+                (feature: VectorTileFeature, styleLayer: StyleLayer, id: ?string | number) => {
                     if (!featureGeometry) {
                         featureGeometry = loadGeometry(feature);
                     }
                     let featureState = {};
-                    if (feature.id) {
+                    if (id) {
```
ansis: Should this be `id !== undefined` to handle 0? (this would have been a pre-existing bug)

src/data/feature_index.js
```diff
@@ -247,6 +249,16 @@ class FeatureIndex {
 
         return false;
     }
+
+    getId(feature: VectorTileFeature, sourceLayerId: string): string | number | void {
+        let id = feature.id;
+        if (this.promoteId) {
+            const propName = typeof this.promoteId === 'string' ? this.promoteId : this.promoteId[sourceLayerId];
```
ansis: Is the intention to allow a single string that is used as the promoteId for all source-layers?

src/data/feature_position_map.js
```diff
@@ -79,6 +82,14 @@ export default class FeaturePositionMap {
     }
 }
 
+export function getNumericId(value: mixed) {
+    const numValue = +value;
```
ansis: I think this incorrectly handles strings that can be converted to javascript numbers but not javascript integers:
- integers javascript can't safely represent `'9007199254740993'`
- decimals `'1.2'`
- `'Infinity'`

It also looks like users of this function expect integers but this can return a float if given a float. Is that ok because it should never be given a float?

src/data/feature_index.js
mourner: Good catch, will fix!

src/data/feature_index.js
mourner: Yes — for better ergonomics if all source layers share the same unique property (which can often be the case), and also to better support GeoJSON sources (no need to set `{_geojsonTileLayer: '...'}`).

src/data/feature_position_map.js
mourner: Initially this was hashing all non-integer ids, but then I changed it because I thought there's no longer a need to have it limited to integer — those get saved in a Float64Array anyway, which supports decimals, values like `Infinity`, and values bigger than `MAX_SAFE_INTEGER`.

mourner: > It also looks like users of this function expect integers 

Oh, actually, in addition to the above, this function being exported is a leftover from earlier — how things get stored in the position map should be an internal detail so there won't be "users" of this. I'll remove the "export".

ansis: > those get saved in a Float64Array anyway, which supports decimals, values like Infinity, and values bigger than MAX_SAFE_INTEGER

Neither javascript numbers (which are 64 bit floats) or Float64Array can store integers > MAX_SAFE_INTEGER safely. Both `'9007199254740993'` and `'9007199254740992'` will look like the same id to this implementation. This might seem unlikely but we ran into this with iD editor because osm uses sequential 64 integers as ids.

Similar problems theoretically exist with decimals but seem less likely.

mourner: Fair enough — updated to hash as a string in case a numeric id is more than `MAX_SAFE_INTEGER`.

mourner:
cc @mapbox/gl-native — this would likely need to be included in the feature-state port work.

waissbluth:
has this made it to a release? thank you

mourner:
@waissbluth no, but it will get into the next release which is a few weeks away.

Eh2406:
Looks like the documentation did not get updated to have this change `Features are identified by their id attribute, which must be an integer or a string that can be cast to an integer.` https://docs.mapbox.com/mapbox-gl-js/api/?size=n_10_n#map#setfeaturestate

PR #8990, opened 2019-11-19
Fix IE 11 support

pakastin:
IE 11 doesn't support `Array.prototype.find`, this PR fixes that by using `forEach` instead of `find`

## Launch Checklist


pakastin:
Ah yeah, fixed now! There were some other linting issues too in other files, but I didn't touch them.

mourner:
@pakastin thanks! The other ones are just warnings, and they're temporary (see #8960)

Issue #8917, opened 2019-10-28
RFC: Stretchable icons


kkaefer:
![nine-part icons](https://user-images.githubusercontent.com/52399/67698534-cff1d600-f9aa-11e9-8e92-148332c11a09.png)

We intend to implement stretchable icons, aka [9-part icons](https://github.com/mapbox/mapbox-gl-js/issues/4143). Mapbox GL currently supports  `icon-size`, which linearly resizes the entire  icon in both dimensions. Stretchable icons allow defining certain areas of an icon that may be stretched, leaving the remaining areas of an icon at a fixed size. This is useful for icons that need to accomodate text of varying width, e.g. highway shields, or text frames.

This is how we imagine our implementation will look like:

### Sprite JSON format

The metadata file that describes a sprite image already contains information about where an image is within a sprite, as well as the pixel ratio. We're going to add three fields:

* `content`: An array of 4 numbers, which represent two coordinates relative to the top left of the _icon_ that span up a rectangle. Both coordinates must be within the dimensions of the icon, and the second coordinate must be greater than the first coordinate. The rect must have have a positive area. When sizing icons with `icon-text-fit`, the icon size will be adjusted so that the this content box fits exactly around the text. If one axis of the icon is not stretchable, the center of the content box will be aligned with the center of the text. Any `icon-text-fit-padding` values will be applied around the text, and the icon size will be adjusted appropriately. 
  Example: `"content": [ 2, 4, 16, 9 ]`

* `stretch-x`/`stretch-y`: An array of 2-element arrays that specify between which coordinates the icon may be stretched. Consequently, an image may have more than one stretch area, e.g. to allow for a central embellishment that should remain at fixed width. Stretch areas are specified independently for both axes, and either axis may not have any stretch areas at all, meaning that it should only be stretched one one axis. The 2-element arrays specify absolute distances in pixels to the top left corner. Stretch areas may not overlap and they must be ordered ascendingly.
  Example: `"stretch-x": [[ 2, 8 ]]`

  

### Icon Rendering

Existing, non-stretchable icons are rendered with a quad (i.e. two triangles) that transfers (and scales) the texture from the image atlas. For stretchable icons, we still copy the icon as is into the image atlas, but instead use multiple quads to render the icons. The quads copy the appropriate portion of the icon from the image atlas, either in a true-to-pixels way, or stretched, based on the metadata found in the Sprite JSON.

[...]

Issue #6856, opened 2018-06-23
text background


andrewharvey:
## Motivation

I'm thinking about using a rectangular or rounded background behind text, like this. As far as I can tell this would need to be supported in the style spec, as it's not possible at the moment.

![text-background](https://user-images.githubusercontent.com/117278/41810238-b4156f54-773e-11e8-9d17-21decf6677b6.png)

@mapbox/maps-design any thoughts? Do you think it's worth including this feature, or is it too specific to implement in the style spec?

mb12:
@andrewharvey Is this any different than the highway shield?  The shield image automatically resizes depending on the text over the shield image.

andrewharvey:
> @andrewharvey Is this any different than the highway shield? The shield image automatically resizes depending on the text over the shield image.

As I understand it, highway shields are just symbol icons set to allow overlap between the icon and text. They are fixed images, and don't resize or fit the text in any way. Yes you can use the `length` expression to use a different size shield based on the text length, but that only works for monospace text, and is only really suited for short `ref`s

mb12:
@andrewharvey My understanding is that if you use icon-text-fit, it should stretch the image. Its default value is none. By default it does not stretch the image. if you use icon-text-fit with "both" in a your regular shield style rule, it would stretch the image.

https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/reference/v8.json#L1081

andrewharvey:
Thank's for pointing that out @mb12! It isn't in Mapbox Studio so I missed it. I think that allows a crude workaround, but it still can't do

* corner rounding (they would get stretched)
* color as a paint property

stevage:
You could probably achieve some kind of workaround using Markers. Which sort of gets me thinking about some weird symbol/marker hybrid that is attached to data, but rendered using the DOM. Then you'd have enormous scope for creativity, at the cost of performance. Maybe using the forthcoming layer plugin architecture...

ChrisLoer:
Could you do color as a paint property by using icon-text-fit with an SDF icon?

I don't know a solution to corner rounding beyond "vector icons" (cc @jfirebaugh).

Tagging @mapbox/maps-design again since it doesn't look like the tag worked in the original comment.

stevage:
Ok, I tried out the hacky workaround. I'm actually pretty impressed how well it works!

https://bl.ocks.org/stevage/23d881a66e2bcca385d8cc074691b674

[...]

PR #8997, opened 2019-11-20
implement stretchable icons for icon-text-fit

ansis:
I'll update this description with examples and more details tomorrow.

This implements the multipart stretchable icons proposed in https://github.com/mapbox/mapbox-gl-js/issues/8917.

Screen Shot 2019-11-25 at 4 14 11 PM

## Launch Checklist

 - [x] post benchmark scores


src/data/bucket/symbol_attributes.js
```diff
@@ -4,8 +4,9 @@ import {createLayout} from '../../util/struct_array';
 
 export const symbolLayoutAttributes = createLayout([
     {name: 'a_pos_offset',  components: 4, type: 'Int16'},
-    {name: 'a_data',        components: 4, type: 'Uint16'}
-]);
+    {name: 'a_data',        components: 4, type: 'Uint16'},
+    {name: 'a_pixeloffset',        components: 4, type: 'Int16'}
```
kkaefer: Have you measured the effect of increasing the size by 8 bytes per vertex?

debug/index.js
```diff
@@ -0,0 +1,95 @@
+const border = 25;
```
kkaefer: Remove this and move a cleaned up version to a dedicated page?

src/render/image_atlas.js
```diff
@@ -21,10 +21,16 @@ export class ImagePosition {
...
+    stretchX: Array<Array<number>> | void;
+    content: Array<number> | void;
 
-    constructor(paddedRect: Rect, {pixelRatio, version}: StyleImage) {
+    constructor(paddedRect: Rect, {pixelRatio, version, stretchX, stretchY, content}: StyleImage) {
         this.paddedRect = paddedRect;
         this.pixelRatio = pixelRatio;
+        if (stretchX) this.stretchX = stretchX;
+        if (stretchY) this.stretchY = stretchY;
+        if (content) this.content = content;
```
kkaefer: Why is this guarded by `if`?

ansis:
> Have you measured the effect of increasing the size by 8 bytes per vertex?

`"tile_bufferData_vertex_bytes"` from bench/gl-stats.html. The benchmark renders a 500x500 mapbox streets map and zooms in by one zoom level.

| zoom | bytes before | bytes after | percent increase |
--- | --- | --- | --- |
| 11 | 1824012 | 1849196 | 1.4% |
| 12 | 2950196 | 3062868 | 3.8% |
| 13 | 1767812 | 1884932 | 6.6 % |
| 14 | 1751168 | 1836320 | 4.8% |

It looks like vertex data for mapbox streets is dominated by other layer types and only 6-24% of vertex data is for symbols.

I'm working on bechmarking rendering now.

ansis:
The existing paint benchmark focuses on benchmarking the cpu side of things so I manually ran some FPS tests (on macbook and a iphone se). There is no clear regression but I don't think the tests are bottlenecked by vertex size. 

Theoretically, if performance were 100% determined by buffer size then based on the buffer size changes this should at most hurt rendering by 6%. But since buffer size doesn't have anywhere near that big of an effect I think this should be negligible (if anything).

There is a whole bunch of vertex size stuff that could be optimized more but without solid, reliable benchmarks that show a benefit I'm not sure it's worth optimizing more right now.

debug/index.js
ansis: done in 69cf025

src/render/image_atlas.js
ansis: fixed in 8e14cc4

src/style/load_sprite.js
```diff
@@ -42,10 +42,10 @@ export default function(baseURL: string,
             const result = {};
 
             for (const id in json) {
-                const {width, height, x, y, sdf, pixelRatio} = json[id];
+                const {width, height, x, y, sdf, pixelRatio, stretchX, stretchY, content} = json[id];
```
kkaefer: I know we don't have validation for the existing properties either, but I think it'd be good to add validation so that we catch errors early before they go into placement.

ansis: added validation in 3e1038c6cf5c574d30a1c3fdf55abfedb9a70660

andrewharvey:
Just wanted to say @ansis this is an awesome feature, more power to designers now!

I must say at first I had a hard time understanding the API docs for what values to use in stretchX/Y and content, so I put together https://github.com/mapbox/mapbox-gl-js/issues/6856#issuecomment-559903043 to help understand.

Were you planning on adding your screenshot as a docs example?

ansis:
> I must say at first I had a hard time understanding the API docs for what values to use in stretchX/Y and content, so I put together #6856 (comment) to help understand.

> Were you planning on adding your screenshot as a docs example?

Yep, it's definitely hard to understand. I'll add a runable an example to the docs that will hopefully help

PR #9003, opened 2019-11-20
Disable using Transferables in Safari

arindam1993:
Related Issue #8771 

This PR attempts to detect whether its being run in Safari, or any WebKit based browser using a combination fo the `window.safari` namespace and userAgent parsing.
`window.safari` works on desktop, and userAgent strings from iOS devices include the device name(iPad/iPhone/iPod) in the string.

We can use and conditionally disable the use of transferable objects.


arindam1993:
Tested on my Macbook Pro and an iPad Pro, memory usage is peaking at around 1-1.3 gb and flattening out after. This is as reported by the Safari remote Web Inspector.
 I could not get it to crash after 10+mins of use.

asheemmamoowala:
@arindam1993 Can you take recordings of those hitches, and try to run our benchmarks on desktop Safari so we can see how this affects other areas of performance?

src/util/util.js
```diff
@@ -449,6 +449,27 @@ export function parseCacheControl(cacheControl: string): Object {
     return header;
 }
 
+let _isSafari = null;
+
+/**
+ * Retuns true if the when run WebKit derived browsers.
```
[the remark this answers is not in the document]
arindam1993: 🤦‍♂

src/util/util.js
```diff
@@ -452,11 +452,12 @@ export function parseCacheControl(cacheControl: string): Object {
 let _isSafari = null;
 
 /**
- * Retuns true if the when run WebKit derived browsers.
- * This is used to work-around a memory-leak in Safari caused by using web-worker transfers.
+ * Retuns true when run in WebKit derived browsers.
```
chloekraw: Returns

src/util/util.js
```diff
@@ -452,11 +452,12 @@ export function parseCacheControl(cacheControl: string): Object {
...
+ * This is used as a workaround for a memory leak in Safari caused by using Transferable objects to
+ * transfer data between WebWorker's and the main thread.
```
chloekraw: apostrophe not needed in web workers (and I think there should be a space)

arindam1993:
> This looks good to me. Do you think we should also add a debug/test page that lets us easily confirm what the value of `isSafari(...)` is any browser?

Added a `is-safari` debug page

PR #9005, opened 2019-11-21
[docs] clarify that text wrapping is disabled if text-max-width is 0

chloekraw:
It's a small change, but I think it's worth clarifying explicitly that setting `text-max-width: 0` disables text wrapping. A customer didn't realize this was an option from our docs, and I also felt there was enough ambiguity that I tested it before suggesting it.

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -1718,7 +1718,7 @@
       "default": 10,
       "minimum": 0,
       "units": "ems",
-      "doc": "The maximum line width for text wrapping.",
+      "doc": "The maximum line width for text wrapping. If `0`, text wrapping is disabled.",
       "requires": [
         "text-field"
       ],


ryanhamley:
This isn't true though. If `text-max-width: 0`, then the text wraps on every single line break. See https://github.com/mapbox/mapbox-gl-js/pull/8706#issuecomment-532355591

ansis:
@chloekraw I wasn't aware this had changed. It looks like using a large number for `text-max-width` is the right way to go.

chloekraw:
Hmm, well, we should definitely document that because that's nonintuitive `0` behavior. 

I'll take another look at what I was seeing tomorrow.

ryanhamley:
>that's nonintuitive 0 behavior.

I would argue that this behavior is exactly what you'd expect with `text-max-width: 0`. If the max width is 0, then you should be trying to break at every possible point because every part of the text is longer than the max width. Previously, `text-max-width: 0` meant that no line breaking occurred which effectively equated 0 and Infinity which are definitely not the same things. I don't see any good reason for a max width of 0 to disable line breaking.

chloekraw:
> I would argue that this behavior is exactly what you'd expect with text-max-width: 0. ... Previously, text-max-width: 0 meant that no line breaking occurred which effectively equated 0 and Infinity which are definitely not the same things.

🤷‍♂There are enough exceptions when it comes to behavior of `0` and `infinity` in math and programming that I wouldn't blindly assume anything about the behavior of `text-max-width: 0` . That's why I suggested adding documentation.

My goal from the beginning has been to prevent users from having to guess or test what happens when `text-max-width: 0` when they read that this input is possible in the docs. I'm not suggesting that we should make `0` disable line breaking as a feature.

> If the max width is 0, then you should be trying to break at every possible point because every part of the text is longer than the max width. 

Sure, even if you are correct, it would be totally reasonable for people to assume that line breaking happens after every character. Most users aren't familiar with the algorithm for line breaking in the renderer and how a "possible point" is defined. 

>  I don't see any good reason for a max width of 0 to disable line breaking.

What I meant by "nonintuitive" is that behavior that line breaks at every whitespace is nonintuitive as a feature. What is the use case for this?

Even if there is a use case for this, `text-max-width: 1` achieves this purpose in the majority of scenario as far as I could see in Studio.

Given current constraints in the style spec, I don't think there is a mathematically obvious way to represent a feature that disables line breaking. I think the best solution would be to create a special input for `text-max-width`. I thought `text-max-width: 0` made sense because of what I wrote above; if we feel that `text-max-width: 0` makes sense as a feature, then I would propose `text-max-width: -1`. 

But this isn't a priority, and I'm not suggesting that the feature happens, just that we clearly document behavior of the renderer :-)

PR #9015, opened 2019-11-22
Fix Click on Compass on some mobile (add clickTolerance to DragRotateHandler)

Yanonix:
This PR adds `clickTolerance` features to `DragRotateHandler` in order to fix click on compass

The "reset to north" of the compass button in `NavigationControl` was bypassed by the drag event, on some mobile version (like Samsung S8 on Chrome). So the drag rotate/pitch was applied and the user couldn't do a simple click to reset the north. Or at least, it needs to be a precise click/touch (with no moving)
This PR reuses the `clickTolerance` system from `DragPanHandler` to apply to the rotate one.
The clickTolerance value is set from the map options.
This is also applied on desktop version while using the drag on the right click or ctrl + left click.

One test is added
You can easily test it with `events.html`

### DIFF
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -103,7 +103,7 @@ class NavigationControl {
             }
             this._map.on('rotate', this._rotateCompassArrow);
             this._rotateCompassArrow();
-            this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass});
+            this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass, clickTolerance: map._clickTolerance});
             DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
             DOM.addEventListener(this._compass, 'touchstart', this._handler.onMouseDown, {passive: false});
             this._handler.enable();
--- a/src/ui/control/navigation_control.js
+++ b/src/ui/control/navigation_control.js
@@ -103,7 +103,8 @@ class NavigationControl {
             }
             this._map.on('rotate', this._rotateCompassArrow);
             this._rotateCompassArrow();
-            this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass, clickTolerance: map._clickTolerance});
+            // Temporary fix with clickTolerance (https://github.com/mapbox/mapbox-gl-js/pull/9015)
+            this._handler = new DragRotateHandler(map, {button: 'left', element: this._compass, clickTolerance: map.dragRotate._clickTolerance});
             DOM.addEventListener(this._compass, 'mousedown', this._handler.onMouseDown);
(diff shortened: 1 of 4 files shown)


ryanhamley:
I think this overall looks pretty good and makes sense to include. I don't know if we should make `_clickTolerance` a map option. Currently, it's passed to the `DragPanHandler` via `DragPanOptions` and I think it makes sense to expose it the same way with the `DragRotateHandler`. We also are likely going to significantly refactor the entire handler architecture soon but it makes sense to get this fix out now I think.

cc @vakila

Yanonix:
> That said, if you still want to add a `clickTolerance` option to `DragRotateHandler`, that sounds fine to me - but I agree with @ryanhamley that this shouldn't be an additional property on the `Map`. We already allow the user to [specify `clickTolerance` as a map option](https://github.com/mapbox/mapbox-gl-js/blob/1c0cc30db40d27ca38f130806760e7f99d7e9c3d/src/ui/map.js#L199), and when we [bind the gesture handlers](https://github.com/mapbox/mapbox-gl-js/blob/1c0cc30db40d27ca38f130806760e7f99d7e9c3d/src/ui/map.js#L421) we pass that option through to [each handler constructor](https://github.com/mapbox/mapbox-gl-js/blob/2020f3f9ced516086508925411bd098fbb4c3234/src/ui/bind_handlers.js#L31), so `options.clickTolerance` is already available to the `DragRotateHandler` constructor.
>
> So next steps to get this PR mergeable would be to revert the changes to `map.js` and `navigation_control.js`, and then it will be looking pretty good! Though again, it won't quite solve the root issue of the bug you experienced - but in #9036 I laid out some next steps I see towards solving that.

I do understand for the map's point of view, but in `NavigationControl`, it's a new instance of [`DragRotateHandler`](https://github.com/mapbox/mapbox-gl-js/pull/9015/files#diff-ba5267dcb7e59b5bd62c6a43d4150a06L106). It's not taken from the map so it can't get `clickTolerance`. I don't know what is the best way to give the parameter, here. A direct get from the `map.dragRotate` to read the `clickTolerance` ?

vakila:
Hi @Yanonix, 

> in NavigationControl, it's a new instance of DragRotateHandler

correct, but what I'm saying here and in https://github.com/mapbox/mapbox-gl-js/issues/9036 is that we should _remove_ that instance of `DragRotateHandler` created by `NavigationControl`.  I understand why you made `clickTolerance` a `Map` property in order to pass the value to the new `DragRotateHandler`, but I think we should take a different approach and remove the drag rotate compass functionality altogether.

So I would accept a patch that adds `clickTolerance` to the `DragRotateHandler`, if you think that would be useful generally, but I wouldn't accept adding a `Map` property just to pass a value to a handler instance we shouldn't be constructing in the first place and will soon remove. Does that make sense?

Yanonix:
I fully understand. I have changed the way to get the value without editing `map.js`.

PR #9016, opened 2019-11-22
update rendering after addImage and removeImage

### LINKED ISSUE #9004: map.removeImage and map.addImage do not update rendered map
I'm attempting to update an image by calling `map.removeImage` and then `map.addImage` in sequence. The reasoning behind this is that we want to be able to change icon sizes on the fly in Mapbox Studio, and `map.updateImage` fails if the update is of a different size than the original icon. I'm not aware of any way to update an image to one of a different size other than removing it and then adding a new image of the same ID.

If there's a workaround, some other way to update an image when its size changes, it'd be an enormous help to hear about it!

**mapbox-gl-js version**: v1.5.0

**browser**: Chrome 78, Firefox 70.0.1

### Steps to Trigger Behavior

 1. Add an image to the map via `map.addImage`.
 2. Remove that image from the map via `map.removeImage`.
 3. See no visible change in the map.
 4. Attempt to add an image via `map.addImage` with the same ID as the one just removed.
 5. See no visible change in the map.

### Link to Demonstration
https://codepen.io/elifitch/pen/3ea68096aec96ec9521cc26fef066569?editors=0110

**This demo has a few buttons.**
1. Removes the original rainbow image and replaces it with a pink one. Expect this to fail.
1. Remove the images from the map. Expect this to fail.
1. Add images of the same ID to the map. Expect this to fail.
1. Switch from the raindbow image to the pink image via `map.updateImage`. Expect this to succeed!

### Expected Behavior
I'd expect to see the map immediately update with the removed image. I'd expect to be able to update an image by removing it then adding another image of the same ID.

### Actual Behavior
Rendered map does not update when calling `map.removeImage`.

cc @mapbox/studio @mapbox/map-design-team

ansis:
fix #9004

A tile needs to be reparsed in order for it to use the new image. Reloading all tiles whenever an image is added would be expensive. For example:
- load a map
- add an image and use it in a geojson overlay

You don't want adding that image to make the basemap tiles be reloaded.

To get around that we track all the image ids used by each tile and only reload tiles that depend on images that changed. 

## Launch Checklist

 - [n/a] document any changes to public APIs (old behavior was not documented)
 - [ ] post benchmark scores

@elifitch


ansis:
Adding the following snippet to `debug/index.html` (and zooming out) shows a single icon that can be toggled by clicking on the map:

<details><summary>snippet</summary>
```js
let img;
let added = false;

function toggle() {
    if (added) {
        map.removeImage('cat');
    } else {
        map.addImage('cat', img);
    }
    added = !added;
}
map.on('style.load', () => {
    map.loadImage('https://upload.wikimedia.org/wikipedia/commons/thumb/6/60/Cat_silhouette.svg/400px-Cat_silhouette.svg.png', function(error, image) {
        if (error) throw error;
        img = image;
        map.on('click', toggle);
        toggle();
        map.addLayer({
            "id": "points",
            "type": "symbol",
            "source": {
            "type": "geojson",
            "data": {
            "type": "FeatureCollection",
            "features": [{
            "type": "Feature",
            "geometry": {
            "type": "Point",
            "coordinates": [-90, 65]
            }
            }]
            }
            },
            "layout": {
            "icon-image": "cat",
            "icon-size": 0.25
            }
        });
    });
});
```
</details>

elifitch:
@ansis Whenever I try to build and either include in a demo app, or just firing up the demo page with `yarn run start-debug`, I get this somewhat cryptic error. I'm following the instructions for Mac OS in Contributing.md. Am I missing something?

image

ansis:
@elifitch I'm not sure, I can't reproduce. Let's track this down through chat tomorrow

ansis:
> Have you tried this with patterns? Should we add a test for pattern as well?

Good idea! I added a pattern test in 95053f5

csimpi:
@ansis For me, this is still an issue.

PR #9017, opened 2019-11-22
Check whether context is lost before `gl.shaderSource(...)`.

### LINKED ISSUE #6312: Failed to execute 'shaderSource' on 'WebGLRenderingContext
We are seeing the following error in our client side error logging (eg real users are seeing this issue, but until recently I was unable to reproduce):

Failed to execute 'shaderSource' on 'WebGLRenderingContext': parameter 1 is not of type 'WebGLShader'.

I've just used testobject.com to test something unrelated, but managed to find a device that crashes with a "Rats WebGL hit a snag." error while trying to render our map.  Initially it loaded fine, but when we added a marker on the map, it crashed with that UX message and in our client side error logging the shaderSource error showed up.

Reloading the site now crashes webgl each time, regardless of adding custom markers or not. Even just rendering the map seems to do it.

The device in question is a:

Motorola Moto E2

Android version: 5.1
API level: 22
Resolution: 540 x 960 (hdpi)
Screen size: 4.5"
CPU: ARM | dual-core | 1200 MHz
RAM: 1024 MB
Internal storage: 8192 MB
Model number: Moto E2
ID: Motorola_Moto_E2_real_us
Location: United States

I suspect it's not supporting the correct webgl bits, and indeed this is dubious:

2018-03-10 15_05_36 - motorola moto e2 - android 5 1

from http://webglreport.com/?v=2 
(alrhough v1 *is* supported)

But if I try the mapbox example to test if mapbox is supported, that map renders fine... so not sure what we are doing that might be causing this?

### LINKED ISSUE #8986: Failed to execute 'shaderSource' on 'WebGLRenderingContext': parameter 1 is not of type 'WebGLShader'.
Version: 1.4.0

From Sentry:

`TypeError: Failed to execute 'shaderSource' on 'WebGLRenderingContext': parameter 1 is not of type 'WebGLShader'.`

Windows 7, Chrome Version: 78.0.3904

image

image

image

I also should note that we use mapbox map with `mapboxgl.supported()` method.

ansis:
fix #8986 fix #6312

When context is lost most gl calls are no-ops but this one can throw. When a program cannot be created `draw(...)` needs to fail silently just like all gl calls.

The thrown errors were unlikely to cause any problems but avoiding them keeps logs clearer.

I tested by triggering context loss just before program creation. I don't know how we can automate testing this.

## Launch Checklist

 - [?] write tests for all new functionality
 - [ ] post benchmark scores

### DIFF
--- a/src/render/program.js
+++ b/src/render/program.js
@@ -45,14 +45,20 @@ class Program<Us: UniformBindings> {
         const fragmentSource = defines.concat(prelude.fragmentSource, source.fragmentSource).join('\n');
         const vertexSource = defines.concat(prelude.vertexSource, source.vertexSource).join('\n');
         const fragmentShader = gl.createShader(gl.FRAGMENT_SHADER);
-        if (this.failedToCreate = gl.isContextLost()) return;
+        if (gl.isContextLost()) {
+            this.failedToCreate = true;
+            return;
+        }
         gl.shaderSource(fragmentShader, fragmentSource);
         gl.compileShader(fragmentShader);
         assert(gl.getShaderParameter(fragmentShader, gl.COMPILE_STATUS), (gl.getShaderInfoLog(fragmentShader): any));
         gl.attachShader(this.program, fragmentShader);
 
         const vertexShader = gl.createShader(gl.VERTEX_SHADER);
-        if (this.failedToCreate = gl.isContextLost()) return;
+        if (gl.isContextLost()) {
+            this.failedToCreate = true;
+            return;
+        }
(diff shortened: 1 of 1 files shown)


ansis:
@mourner yep, according to https://www.khronos.org/webgl/wiki/HandlingContextLost context can be lost between calls. The error is caused by passing a bad fragment shader to `shaderSource(...)` so we don't need to worry about context loss between the `isContextLoss(...)` and the `shaderSource(...)`.

Issue #8998, opened 2019-11-20
IE11 tile queue regression


mourner:
Capturing @arindam1993 https://github.com/mapbox/mapbox-gl-js/pull/8913#issuecomment-553622495 to make sure we address it before doing a release:

> this PR could bring back the issue of requests being stuck in the pending state on IE11 when tile requests get cancelled, I just did some testing and it looks like it has 😬.

arindam1993:
Did some testing to quantify the difference better:
I'm using the following debug page to trigger a large amount of tile load cancellations:
After the map `idle` event it rapidly zooms out and zooms back into the same spot.
```
<!DOCTYPE html>
<html>
<head>
    <title>Mapbox GL JS debug page</title>
    <meta charset='utf-8'>
    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
    <link rel='stylesheet' href='../dist/mapbox-gl.css' />
    <style>
        body { margin: 0; padding: 0; }
        html, body, #map { height: 100%; }
    </style>
</head>

<body>
<div id='map'></div>

<script src='../dist/mapbox-gl-dev.js'></script>
<script src='../debug/access_token_generated.js'></script>
<script>

var map = window.map = new mapboxgl.Map({
    container: 'map',
    zoom: 12.5,
    center: [-77.01866, 38.888],
    style: 'mapbox://styles/mapbox/streets-v10',
    hash: true
});

map.once('idle', function(){
    map.easeTo({ center: [-77.01866, 38.888], zoom: 1, duration: 1000});

    setTimeout(function(){
        map.easeTo({ center: [-77.01866, 38.888], zoom: 12.5, duration: 1000});
    }, 1200)
});

</script>
</body>
</html>
```

And here are the results:

| browser      | 1.5.0 | 1.5.1-beta |
|--------------|-------|------------|
| chrome (osx) | ~40   | ~70        |
| ie 11        | ~25   | ~120       |

The number indicates the number of requests sent out by looking at the network tab in the inspector.

PR #9022, opened 2019-11-25
Add budgets to Actor queue flushing

arindam1993:
This PR tries to achieve some middleground between the issues reported in #8998 and #8913.
It tries to do this with 2 main changes:
- The Actor on processes messages in batches, based on time or size budgets, whichever is reached first.
- On the main thread the `ThrottledInvoker` uses `requestAnimationFrame` instead of `MessageChannel` in order to reduce latency for animations. 

I've tried to adjust the batch sizes on the main-thread and worker side so we can have a balance between being able to pre-empt and cancel tasks, while maintaining enough throughput for smooth animation.

Testing with IE11, which needs pre-empting the most, the number of requests sent out for the tile-cancellation debug page went down to ~60 from the ~120 on `1.5.1-beta`. But its still higher than the ~25 on `1.5.0`.
Animation performance on `Chrome` is roughly the same for me.

@msbarry , would really appreciate if you could take a look at this as well!

## Launch Checklist

### DIFF
--- a/debug/tile-cancellation.html
+++ b/debug/tile-cancellation.html
@@ -26,12 +26,12 @@ var map = window.map = new mapboxgl.Map({
     hash: true
 });
 
-map.once('idle', function(){
-    map.easeTo({ center: [-77.01866, 38.888], zoom: 1, duration: 1000});
+map.once('idle', function() {
+    map.easeTo({center: [-77.01866, 38.888], zoom: 1, duration: 1000});
 
-    setTimeout(function(){
-        map.easeTo({ center: [-77.01866, 38.888], zoom: 12.5, duration: 1000});
-    }, 1200)
+    setTimeout(function() {
+        map.easeTo({center: [-77.01866, 38.888], zoom: 12.5, duration: 1000});
+    }, 1200);
 });
 
 </script>
--- a/debug/tile-cancellation.html
+++ b/debug/tile-cancellation.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Mapbox GL JS debug page</title>
+    <meta charset='utf-8'>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+    <link rel='stylesheet' href='../dist/mapbox-gl.css' />
+    <style>
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+
+<body>
+<div id='map'></div>
+
+<script src='../dist/mapbox-gl-dev.js'></script>
+<script src='../debug/access_token_generated.js'></script>
+<script>
+
+var map = window.map = new mapboxgl.Map({
+    container: 'map',
+    zoom: 12.5,
+    center: [-77.01866, 38.888],
+    style: 'mapbox://styles/mapbox/streets-v10',
+    hash: true
+});
+
+map.once('idle', function(){
+    map.easeTo({ center: [-77.01866, 38.888], zoom: 1, duration: 1000});
+
+    setTimeout(function(){
+        map.easeTo({ center: [-77.01866, 38.888], zoom: 12.5, duration: 1000});
+    }, 1200)
+});
+
+</script>
+</body>
+</html>
--- a/src/util/actor.js
+++ b/src/util/actor.js
@@ -143,8 +143,8 @@ class Actor {
             this._processQueueTop();
             taskCtr++;
         }
-        // We've reached our budget for this frame, defer processing to the rest for the next tick, this allows
-        // the deferred tasks to be preempted on slower browsers.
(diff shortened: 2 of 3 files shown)


msbarry:
Also, I wonder if it's necessary to have the task budget at all?  If we just have a time budget, then a lot of fast tasks can happen during animation but it will throttle heavy operations. If I change task budget to unlimited and both time budgets to 2, then I'm back up to 30-40 updates per second.

mourner:
> if we always set the `immediate` flag for all requests associated with a particular source (e.g. GeoJSON sources)

I thought about type-based categorization too, but GeoJSON sources aren't always used with light-data-fast-updates use case — e.g. the GeoJSON might be pretty heavy while never updating, making it similar in behavior to vector tiles.

arindam1993:
> Great improvement, but still a bummer to have a 2.5x regression for tiles stuck in a queue compared to the previous version.
> 
> One random idea that you already considered perhaps, but still thought I'd mention — can we differentiate the way we process the queue not based on arbitrary frame budgets, but based on a heuristic that somehow categorizes sources as either heavy / rarely updated (e.g. basemap vector tiles) or lightweight / frequently updated (e.g. animating GeoJSON)? That way we could have optimal performance in both cases if we can reliably categorize between the two. Perhaps by tracking how often the source is updated and switching dynamically?

Honestly, animation performance in IE11 is terrible by default anyways, so the best heuristic might be detecting the underlying platform, all modern browsers can cancel `fetch` in the worker anyways.
But dynamically turning up the task-budget when the map is `animating` i,e when a source is being updated frequently is also viable, idk if thats too much complexity vs the browser detection.

> Also, I wonder if it's necessary to have the task budget at all? If we just have a time budget, then a lot of fast tasks can happen during animation but it will throttle heavy operations. If I change task budget to unlimited and both time budgets to 2, then I'm back up to ~30 updates per second.

The task budget is for IE11 cancellation, in my testing no modern browser (`Chrome`/ `Firefox`) ever hit that limit, since its measuring time for a synchronous loop which is incredibly fast anyways. By removing the task budget modern browsers always end up processing everything in the queue every frame, which results in great animation performance, but since there is no deferral anymore the ability to pre-empt tasks is lost in IE11.



> How did you determine that both a time _and_ a task budget achieves the best compromise?

I started out with a time budget only, but I found out modern browsers never hit it ,but IE11 did  hit it rather inconsistently but it wasn't leaving enough tasks in the queue to allow for any decent amount of request cancellation. I then added the task budget and tuned it until I was able to get to a compromise between request cancellation and animation updates. The tiome budget mostly prevents some main thread hitching on IE11 afaict.


> The original issue fixed in #8633 was that tile requests sent to a worker couldn't be cancelled due to tasks being invoked in serial order. This is primarily a concern when loading vector tiles. The animation performance regression reported in #8701 doesn't use vector tile loading; instead it uses GeoJSON sources with repeated `setData`. The GeoJSON source has its own mechanism for deduplicating updates (self-sent coalesce messages), similar to how native does it. Over in the original PR, I wrote
> 
> > I opted for this model instead of a more generic priority queue because we don't have any use cases for a priority queue other than canceling tasks.
> 
> However, given that we actually do seem to have differing requirements for worker processing, a priority queue may actually make more sense. Alternatively, we could mark some tasks with an `immediate` flag which means they get executed immediately instead of put into the queue. This means that tasks may be executed out of order, but if we always set the `immediate` flag for all requests associated with a particular source (e.g. GeoJSON sources), this may be fine.

ansis:
I think there is a lot we could do to improve queuing, cancellation, prioritization but I think we should focus more narrowly on fixing the regression (if it exists) so that we can get the beta out.

From the numbers here it's not clear that something regressed in IE.

> Testing with IE11, which needs pre-empting the most, the number of requests sent out for the tile-cancellation debug page went down to ~60 from the ~120 on 1.5.1-beta. But its still higher than the ~25 on 1.5.0.

The only way you cancel a request before it starts is by delaying it. What these numbers suggest to me is that we are delaying requests less. This is a good thing? 

https://github.com/mapbox/mapbox-gl-js/pull/8633 tried to fix the problem that requests that were already cancelled were initiated anyways because the message that they weren't cancelled hadn't been opened yet. That was happening because workers do large tasks that delay for long chunks of time. My *guess* is that that is *not* what is happening now. My guess is that the delay in 1.5.0 was much longer than the time it took to open all the waiting messages.

Zooming at the wrong speed in any browser will trigger a lot of requests and cancellations.

If IE was striking a better balance in 1.5.0 it seems it was semi-random rather than something intentionally tuned to delay requests during rapid zooming.

If we need to fix the rapid zooming / cancellation problem now I think we should look into:
- some direct increasing delay for tile requests during rapid zooming (eg a setTimeout before the tile request gets sent to the worker)
- or limiting the number of concurrent requests: experimental branch here https://github.com/mapbox/mapbox-gl-js/tree/ie-alt-fix

arindam1993:
Closing since #9031 fixes this!

PR #9025, opened 2019-11-26
Type annotations

kkaefer:
Adding a few more type annotations. Had this stashed locally, just making sure it doesn't get lost.

### DIFF
--- a/src/style-spec/util/get_type.js
+++ b/src/style-spec/util/get_type.js
@@ -1,6 +1,6 @@
 // @flow
 
-export default function getType(val: any): string {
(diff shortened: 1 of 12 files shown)


src/style-spec/util/get_type.js
```diff
@@ -1,5 +1,6 @@
+// @flow
 
-export default function getType(val) {
+export default function getType(val: any): string {
```
mourner: Just wondering (didn't check), can you replace instances like this with `mixed`? Wherever the type either gets refined or doesn't matter for the resulting type.

PR #9026, opened 2019-11-26
fix urlRe: Forward slashes must be escaped.

fatelgit:
  fix typo


mourner:
Looks like `/` is not considered a special character in the JS regex engine (unlike e.g. PCRE), according to https://regex101.com/ and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions, so this PR doesn't have any effect — safe to   close.

PR #9028, opened 2019-11-26
allow rendering full world smaller than 512px

### LINKED ISSUE #3055: Fix bug preventing full map from rendering when viewport is < 512px
### Steps to Trigger Behavior
1. Create a map using, e.g., Mapbox Streets, in a div that is smaller than 512x512.
2. Zoom 'all the way out' (i.e. initialize the map to `center: [0, 0], zoom: 0`
### Expected Behavior

The whole world is shown.
### Actual Behavior

The whole world is not shown.

Repro: https://bl.ocks.org/anandthakker/raw/0e9afc1d09356f87449b5ffa3e49bd82/

ansis:
fix #3055

This allows full world maps to be rendered < 512px by allowing map camera zoom levels to be negative. This does *not* add negative zoom levels to sources or to styles.

When `zoom < 0`, it still loads the same covering tiles it would for zoom 0. It does not reparse tiles. It just pretends that the zoom is 0 when deciding which tiles to load.

Screen Shot 2019-11-26 at 3 03 43 PM

Breaking change?

Changing the map's default zoom level means that you can zoom out further than before. I'd say this isn't breaking because it's expanding the scope of allowed behavior rather changing behavior within the previously allowed range.

The limit is now `-2` which allows 128px maps. Should it go even lower?

try it out with:

```diff
diff --git a/debug/index.html b/debug/index.html
index 04198d03e..131a3d9e4 100644
--- a/debug/index.html
+++ b/debug/index.html
@@ -7,7 +7,7 @@
     <link rel='stylesheet' href='../dist/mapbox-gl.css' />
     <style>
         body { margin: 0; padding: 0; }
-        html, body, #map { height: 100%; }
+        html, body, #map { width: 200px; height: 200px; }
     </style>
 </head>
 
diff --git a/debug/satellite.html b/debug/satellite.html
index c142fa8ab..e5ed9bb8d 100644
--- a/debug/satellite.html
+++ b/debug/satellite.html
@@ -7,7 +7,7 @@
     <link rel='stylesheet' href='../dist/mapbox-gl.css' />
     <style>
         body { margin: 0; padding: 0; }
-        html, body, #map { height: 100%; }
+        html, body, #map { height: 128px; width: 128px; }
     </style>
 </head>
 ```

## Launch Checklist

 - [n/a] post benchmark scores

### CODE REVIEW

File: src/ui/map.js [hunk 1]
```diff
@@ -101,7 +101,7 @@ type MapOptions = {
     locale?: Object
 };
 
-const defaultMinZoom = 0;
+const defaultMinZoom = -2;
```
vakila: I think we should stick with 0 by default, and allow folks to explicitly pass in a negative `minZoom` if they really want to enable this.

test/unit/ui/handler/touch_zoom_rotate.test.js
```diff
@@ -40,8 +40,9 @@ test('TouchZoomRotateHandler fires zoomstart, zoom, and zoomend events at approp
 
     simulate.touchend(map.getCanvas(), {touches: []});
     map._renderTaskQueue.run();
-    t.equal(zoomstart.callCount, 1);
-    t.equal(zoom.callCount, 2);
+    // incremented because inertia starts a second zoom
```
vakila: :thinking: I'm not sure I understand what's going on here, why would this change affect this test? Seems like maybe something we need to fix in the `TouchZoomRotateHandler`?

vakila:
Thanks @ansis this is great, and so nice & simple too! Thanks also for adding the render test :pray: 

Left some minor comments and still need to test this out locally but generally it's looking great to me

test/unit/ui/handler/touch_zoom_rotate.test.js
ansis: Yeah, it took me a while to figure this out. A touch zoom can trigger two separate zoom events: one for while the fingers are down and one for the inertia zoom afterwards. Previously this wasn't showing up in the test because the map's minzoom was preventing the inertia zoom. The updated tests reflect the existing, actual behavior.

https://github.com/mapbox/mapbox-gl-js/blob/4e12ca76c45ef3923456f6245fb4565674fdf44f/src/ui/handler/touch_zoom_rotate.js#L237
https://github.com/mapbox/mapbox-gl-js/blob/4e12ca76c45ef3923456f6245fb4565674fdf44f/src/ui/handler/touch_zoom_rotate.js#L280

vakila: Oof - adding this to my long list of things to fix in the gesture handlers. OK for changing the test to reflect actual behavior, thanks for catching this

ryanhamley:
When I try this out, it seems like zooming and panning when z < 0 results in a lot of stuttering, jumpy interactions. Does anyone else see that behavior? It definitely doesn't seem to move smoothly.

src/ui/map.js
```diff
@@ -614,8 +614,8 @@ class Map extends Camera {
      * If the map's current zoom level is lower than the new minimum,
      * the map will zoom to the new minimum.
      *
-     * @param {number | null | undefined} minZoom The minimum zoom level to set (0-24).
-     *   If `null` or `undefined` is provided, the function removes the current minimum zoom (i.e. sets it to 0).
+     * @param {number | null | undefined} minZoom The minimum zoom level to set (-2 - 24).
+     *   If `null` or `undefined` is provided, the function removes the current minimum zoom (i.e. sets it to -2).
```
ryanhamley: I think this comment should be expanded to clarify that negative zoom levels equate to containers < 512px square. It's not possible to set a map to a negative zoom when the container is larger than a tile.

vakila:
@ansis I'm seeing the same lack-of-smoothness as @ryanhamley :point_up:  - I wonder if this is bc of the map itself or bc our handlers are struggling with the negative values? I'm not sure this is a dealbreaker though.

Perhaps more concerning, it looks like negative zoom values are still coming through in the source caching code as it requests tile IDs. I'm getting a lot of errors like this:
image

Guessing we need a `Math.max(0, ...)` or two somewhere in `source_cache.js` like maybe here: https://github.com/mapbox/mapbox-gl-js/blob/a1e1be22d5e44bd6550c64c55f9424f74e397204/src/source/source_cache.js#L324

vakila:
Also - just want to flag for anyone following along that given that we still force the map to fill the vertical space of the container, this won't satisfy the use case of showing the entire world in a small, portrait-oriented viewport, as described here https://github.com/mapbox/mapbox-gl-js/issues/3055#issuecomment-481941623 - e.g. for a container with `height: 200px; width: 100px` the map will look like this (zoom `-1.36`)
image

src/ui/map.js
ansis: Does fb015e8 look clear enough?

ansis:
@ryanhamley @vakila thanks for noticing that error. Yep, that line looks like it also needs a max but it looks like it can also be deduplicated with the other method: f2a3662219817e5825e2158857512d91b6f6402f

> Does this account for the issue Konstantin raised about certain Z values passed to shaders needing to be >= 0?

Yep, I'm fairly sure! Since the tiles are parsed with `zoom: 0` we never have to store lower values in the buffers.

vakila:
Thanks @ansis, the dedupe fix looks good! 

I noticed when zooming/panning around at negative zoom on a mobile device (in landscape) there is some weird jumping to z0 that I observe, some but not all of the time. I'm guessing it's related to this hardcoded 0 cutoff in `TouchZoomRotateHandler._onEnd()`, but I can't understand given this code how I'm _ever_ able to touch-zoom and get it to end up below z0 :thinking: In any case we need to change this to use the minzoom: https://github.com/mapbox/mapbox-gl-js/blob/a1e1be22d5e44bd6550c64c55f9424f74e397204/src/ui/handler/touch_zoom_rotate.js#L276-L278

I checked the rest of the handlers and I think this is the only one that hardcodes the minzoom like that, but if anyone wants to take another look please do

ansis:
> I noticed when zooming/panning around at negative zoom on a mobile device (in landscape) there is some weird jumping to z0 that I observe, some but not all of the time.
Thanks for testing this on mobile and catching this. fixed in 31f8918bbac535882babc0f2e0993354b8881fed

>  I'm guessing it's related to this hardcoded 0 cutoff in TouchZoomRotateHandler._onEnd(), but I can't understand given this code how I'm ever able to touch-zoom and get it to end up below z0 🤔 

I could only get it below zero in the middle of a touchzoom but it always reset to 0 at the end.

test/unit/ui/handler/touch_zoom_rotate.test.js
vakila: Opened https://github.com/mapbox/mapbox-gl-js/issues/9034

PR #9030, opened 2019-11-27
Update style specification's 'sdk-support' sections

alexshalamov:
It would be nice if style specification could define versioned dependencies between expressions and/or attributes, so that we could improve style compatibility checks.

For example:
In version 1.4.0 it is possible to use `image` expression in `icon-image` and `*-pattern` properties. Starting from 1.6.0, `image` expression could also be used as a section in `format` expression
```json
["format", "HelloWorld!", ["image", "globe_image"]]
```

This PR:
 - Updates `sdk-support` section for `text-writing-mode`
 - Adds `text-color` and `image` compatibility table entries for `format` expression

## Checklist

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2056,10 +2056,10 @@
       ],
       "sdk-support": {
         "basic functionality": {
-          "js": "next",
-          "android": "next",
-          "ios": "next",
-          "macos": "next"
+          "js": "1.3.0",
+          "android": "8.3.0",
+          "ios": "5.3.0",
+          "macos": "0.14.0"
         }
       },
       "expression": {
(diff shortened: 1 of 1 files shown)


alexshalamov:
@elifitch @samanpwbb @asheemmamoowala I'm trying to find a good name for a '[sdk-support'](https://github.com/mapbox/mapbox-gl-js/pull/9030/files#diff-a9f1cc7c35076f65c5b6c2fd1565cdbeR5181) section that would represent versioned dependencies between expressions and/or attributes.

Could you please take a look and propose a better name?

kkaefer:
This is so that we can introducing a top-level property e.g. `"required-features"` for styles that list required features, then alert the developer if the style they're using uses unknown features.

src/style-spec/reference/v8.json
```diff
@@ -2814,11 +2814,16 @@
...
       "image": {
-        "doc": "Returns an `image` type for use in `icon-image` and `*-pattern` entries. If set, the `image` argument will check that the requested image exists in the style and will return either the resolved image name or `null`, depending on whether or not the image is currently in the style. This validation process is synchronous and requires the image to have been added to the style before requesting it in the `image` argument.",
+        "doc": "Returns an `image` type for use in `icon-image`, `*-pattern` entries and as a section in `format` expression. If set, the `image` argument will check that the requested image exists in the style and will return either the resolved image name or `null`, depending on whether or not the image is currently in the style. This validation process is synchronous and requires the image to have been added to the style before requesting it in the `image` argument.",
         "group": "Types",
         "sdk-support": {
           "basic functionality": {
             "js": "1.4.0"
+          },
+          "supported in expression": {
+            "format": {
```
kkaefer: Why do we need nesting here? Could we simplify this so that it just reads `"image in format expression"`?

asheemmamoowala: @kkaefer Doing that would require a unique section in each property that can be nested into another in a non-trivial way (for eg: with `format`). Do you see an automated way of building this into a support verification matrix?

alexshalamov:
@asheemmamoowala Added two entries to `format` expression's sdk-support section. Please take a look.

PR #9031, opened 2019-11-27
re-add queuing for getResource calls

### LINKED ISSUE #8998: IE11 tile queue regression

ansis:
This is alternative to https://github.com/mapbox/mapbox-gl-js/pull/9022 . It tries to fix #8998 by  reverting the relevant behavior change in #8913.

This isn't an improvement. It just tries to avoid a regression.

#8913 made all worker tasks received by the main thread execute immediately. This pr re-adds queuing for `getResource` tasks.

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/src/util/actor.js
+++ b/src/util/actor.js
@@ -110,40 +110,42 @@ class Actor {
                 cancel();
             }
         } else {
-            // In workers, store the tasks that we need to process before actually processing them. This
-            // is necessary because we want to keep receiving messages, and in particular,
-            // <cancel> messages. Some tasks may take a while in the worker thread, so before
-            // executing the next task in our queue, postMessage preempts this and <cancel>
-            // messages can be processed. We're using a MessageChannel object to get throttle the
-            // process() flow to one at a time.
-            this.tasks[id] = data;
-            this.taskQueue.push(id);
-            if (isWorker()) {
+            if (isWorker() || data.type === 'getResource') {
+                // In workers, store the tasks that we need to process before actually processing them. This
+                // is necessary because we want to keep receiving messages, and in particular,
+                // <cancel> messages. Some tasks may take a while in the worker thread, so before
+                // executing the next task in our queue, postMessage preempts this and <cancel>
+                // messages can be processed. We're using a MessageChannel object to get throttle the
+                // process() flow to one at a time.
+                this.tasks[id] = data;
+                this.taskQueue.push(id);
                 this.invoker.trigger();
             } else {
                 // In the main thread, process messages immediately so that other work does not slip in
                 // between getting partial data back from workers.
-                this.process();
+                this.process(id, data);
             }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/util/actor.js
```diff
@@ -110,40 +110,42 @@ class Actor {
...
-            // In workers, store the tasks that we need to process before actually processing them. This
-            // is necessary because we want to keep receiving messages, and in particular,
-            // <cancel> messages. Some tasks may take a while in the worker thread, so before
-            // executing the next task in our queue, postMessage preempts this and <cancel>
-            // messages can be processed. We're using a MessageChannel object to get throttle the
-            // process() flow to one at a time.
-            this.tasks[id] = data;
-            this.taskQueue.push(id);
-            if (isWorker()) {
+            if (isWorker() || data.type === 'getResource') {
```
kkaefer: Instead of hardcoding the name of the message, can we find another way of marking tasks as immedidate vs. queueable, e.g. by checking for `!data.immediate` here, and adding `immediate as another argument to `Actor#send`?

src/util/actor.js
```diff
@@ -110,40 +110,42 @@ class Actor {
...
+            // Schedule another process call if we know there's more to process _before_ invoking the
+            // current task. This is necessary so that processing continues even if the current task
+            // doesn't execute successfully.
+            if (this.taskQueue.length) {
+                this.invoker.trigger();
+            }
+            if (!task) {
+                // If the task ID doesn't have associated task data anymore, it was canceled.
+                return;
+            }
```
kkaefer: Let's separate this into two functions

src/util/actor.js
ansis: done in d28ac5668f35d9354b9ae1fc7dfbb15b9296aec7

src/util/actor.js
ansis: done in d28ac5668f35d9354b9ae1fc7dfbb15b9296aec7

src/util/actor.js
```diff
@@ -53,7 +53,7 @@ class Actor {
      * @param targetMapId A particular mapId to which to send this message.
      * @private
      */
-    send(type: string, data: mixed, callback: ?Function, targetMapId: ?string): ?Cancelable {
+    send(type: string, data: mixed, callback: ?Function, targetMapId: ?string, mustQueue: ?string): ?Cancelable {
```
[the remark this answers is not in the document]
ansis: Thanks! done in 1028521eae756ae638a02fa8ec7423e0d8cfc43d

PR #9035, opened 2019-11-27
Add instrumentation for performance metrics tracking

arindam1993:
This PR cherrypicks the instrumentation code from #8931 , and adds a new `content.load` event which is basically the first `idle` event fired after the `load` event. 

This was added so we can measure the full load time, and account for the time for placement to finish and icons to load. I think we can also do this by listening for the `idle` event once after `load`, but this seems slightly less hackier.

The instrumentation gets stripped out in production builds so we don't take a bundle size hit. Infact this PR reduces the bundle size a tiny bit because of the refactoring of `performance.js` 🎉 

The rest of the changes from #8931 , including the bench suite and the driver code will be slowly cherrypicked over multiple PR's.

## Launch Checklist


ansis:
Overall this looks good. Let me know if any of these questions are outside the scope of this pr.

The frame measurement seems like it assumes the map is continuously being repainted. Would measuring the duration between successive frames be more accurate? It seems like the duration between the last frame before idling and the first frame should not be counted.

I think using the start-of-frame time value that the `requestAnimationFrame` callback is called with might lead to less noisy values.

Rendering more inexpensive frames would push up the `fps` value without improving performance. Would the absolute number of dropped frames be more appropriate for benchmarks? (I guess this might depend on the benchmark). Or maybe measuring dropped frames per second? I guess the worst 1% of frames (from your link https://cgvr.cs.ut.ee/wp/index.php/frame-rate-vs-frame-time/) helps with this.

> This was added so we can measure the full load time, and account for the time for placement to finish and icons to load.

This means the transition duration influences this load value. Reducing the transition duration would significantly reduce the measured time without a performance change. Is this ok?

>  I think we can also do this by listening for the idle event once after load, but this seems slightly less hackier.

I think your approach is fine but using the publicly documented events might be more stable over time.

Does the performance marking assume the page has only one map?

arindam1993:
> The frame measurement seems like it assumes the map is continuously being repainted. Would measuring the duration between successive frames be more accurate? It seems like the duration between the last frame before idling and the first frame should not be counted.
> 
> I think using the start-of-frame time value that the `requestAnimationFrame` callback is called with might lead to less noisy values.

Yes, the benchmark driver does set `map.repaint = true`. I was thinking capturing that kind of raw-raster time ( with frames that have no CPU work in them) would also be useful, and it also keeps things simpler from the instrumentation perspective.
I am tracking the fps via duration between frames, but with `performance.now()`, +1 to the idea of using the timestamp provided by `requestAnimationFrame` instead.

> Rendering more inexpensive frames would push up the `fps` value without improving performance. Would the absolute number of dropped frames be more appropriate for benchmarks? (I guess this might depend on the benchmark). Or maybe measuring dropped frames per second? I guess the worst 1% of frames (from your link https://cgvr.cs.ut.ee/wp/index.php/frame-rate-vs-frame-time/) helps with this.

Oh yea, I like the idea of tracking total dropped frames instead of 1% slowest frames, though a %age would probably better given the variable length of benchmark runs.

> > This was added so we can measure the full load time, and account for the time for placement to finish and icons to load.
> 
> This means the transition duration influences this load value. Reducing the transition duration would significantly reduce the measured time without a performance change. Is this ok?

I think so, the styles used for benchmarking should stay constant, and if we change something duration related internally its still a reduction in time thats visible to the end user.

> > I think we can also do this by listening for the idle event once after load, but this seems slightly less hackier.
> 
> I think your approach is fine but using the publicly documented events might be more stable over time.
 
Yea, I was thinking adding timing probes straight into the codebase instead of timing public facing events would be more reliable, that way we wouldn't capture the small overhead in `Evented`, plus the flexibility of not mixing the two concerns of an API event and a performance metric timing point seems useful o me.

I'm still not sure if need to expose this point in time as a new event though, its basically just sugar for `map.once('idle')`

> Does the performance marking assume the page has only one map?

Oh good point, yes it does, i think moving the metrics extraction to the `mapboxgl` namespace and restricting it to be available only in dev builds with the `Debug.extend...` would be a better approach.

src/util/performance.js
```diff
@@ -1,45 +1,74 @@
...
 
-wrapper.clearMeasures = (name: string) => {
-    if (performanceExists && performance && performance.clearMeasures)
-        return performance.clearMeasures(name);
-    else
-        return false;
+let lastFrameTime = null;
+let frameTimes = [];
+
+const minFramerateTarget = 30;
```
ansis: What are the tradeoffs between using 30 and 60 here?

arindam1993: Setting it at seemed to introduce a lot of noise due to a combination of  measurement error and being very rarely always locked at 60, so almost every frame ends up missing the target. 
The idea is to capture `really bad` frames in a metric, and half of our ideal seemed like a good start.

PR #9049, opened 2019-12-01
Fix map size in benchmarks

astojilj:
`display: none` on Chrome (Chrome 78.0.3904.108 (Official Build) (64-bit) on macos), results with container clientWidth and clientHeight equal to 0.
This [led to container size 400 x 300](https://github.com/mapbox/mapbox-gl-js/blob/3c36a67716d7c8ad8d69b15f7e5fd19b72e7a7ff/src/ui/map.js#L1922) instead of using specified size in e.g. bench/benchmark/layers.js (1024x768).

Edit:
400 x 300 might make more sense as it is faster. Maybe better lo leave it as it is but replace ignored 1024 x 768 by 400 x 300.

### DIFF
--- a/bench/lib/create_map.js
+++ b/bench/lib/create_map.js
@@ -6,9 +6,10 @@ export default function (options: any): Promise<Map> {
     return new Promise((resolve, reject) => {
         const container = document.createElement('div');
         container.style.width = `${options.width || 512}px`;
-        container.style.height = `${options.width || 512}px`;
+        container.style.height = `${options.height || 512}px`;
         container.style.margin = '0 auto';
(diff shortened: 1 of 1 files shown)


astojilj:
In the same time, I put breakpoint after idealTileIDs calculation in source_cache.js (6 tiles in tile cover) and in e.g. draw_lines.js run from LayerLine benchmarks (2 tiles used for rendering). If previous code, using 400x300 map always resulted with 1 tile being always used ... it might be more reliable than simply increasing the size by this fix: we might end up using different tile count in every benchmark run.... unless waiting for tiles to load.

PR #9050, opened 2019-12-01
optimize findLoadedParent

astojilj:
Performance optimization when rendering view with a large number of raster tiles.
Measurements on Chrome (Version 78.0.3904.108 with CPU 4x slowdown in performance tab) on MacMini i7 3.2 Ghz, 3840 x 1440 display.

`map.repaint = true` in [debug/satellite.html with pitch of 60 degrees](http://localhost:9966/debug/satellite.html#12.5/38.888/-77.01866/0/60)

FPS count, like on the screenshot below:
master: 19-20FPS
with this patch: 24.5-26 FPS

Screen Shot 2019-12-01 at 22 40 10

### DIFF
--- a/src/source/tile_id.js
+++ b/src/source/tile_id.js
@@ -95,6 +95,21 @@ export class OverscaledTileID {
         }
     }
 
+    /*
+     * calculateScaledKey is an optimization:
+     * when withWrap == true, implements the same as this.scaledTo(z).key,
+     * when withWrap == false, implements the same as this.scaledTo(z).wrapped().key.
+     */
+    calculateScaledKey(targetZ: number, withWrap: boolean) {
+        assert(targetZ <= this.overscaledZ);
+        const zDifference = this.canonical.z - targetZ;
+        if (targetZ > this.canonical.z) {
+            return calculateKey(this.wrap * +withWrap, targetZ, this.canonical.x, this.canonical.y);
+        } else {
+            return calculateKey(this.wrap * +withWrap, targetZ, this.canonical.x >> zDifference, this.canonical.y >> zDifference);
+        }
+    }
+
     isChildOf(parent: OverscaledTileID) {
         if (parent.wrap !== this.wrap) {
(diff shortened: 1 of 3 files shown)


src/source/tile_id.js
```diff
@@ -95,6 +95,16 @@ export class OverscaledTileID {
         }
     }
 
+    calculateScaledKey(targetZ: number, wrap: 0 | 1) {
```
kkaefer: I remember that `wrap` can also be `-1` in some situations, or `2` if we show tiles in an extreme wide map view.

astojilj: Added comment.
How does it sound https://github.com/mapbox/mapbox-gl-js/commit/46731324bec5dbe294a66ce11b687b1fe68de734#diff-e444e40548c379347d0aa642beffc811R99?
Renamed wrap to withWrap as it has boolean semantics here.

ansis: it should be fine to use use `wrap: boolean` and `true`/`false` for this

ansis:
Overall this looks good to me! So the savings are mostly from not needing to create all the tileid objects?

src/source/tile_id.js
astojilj: > it should be fine to use use `wrap: boolean` and `true`/`false` for this

Prefer to keep 1/0 as they are used in multiplication later. Through ternary or +withWrap would do the same. Documentation should cover it.

astojilj:
> Overall this looks good to me! So the savings are mostly from not needing to create all the tileid objects?

Yep.
We can further optimize this as it might now make sense to look for a parent this deep - currently, when e.g. rendering raster tile 22 we check for fade parents all the way from 21 to 0.

src/source/tile_id.js
astojilj: @ansis Thanks.
After offline discussion,  `withWrap: boolean`used and `+withWrap` in multiplication. Comment updated.
Updated patch pushed.

Issue #9051, opened 2019-12-02
Cannot re-add attribution control without attribution property in TileJSON


cs09g:
**mapbox-gl-js version**:
v1.6.1, v1.5.1, ... maybe more
**browser**:
latest chrome

### Steps to Trigger Behavior

 1. initialize a map with custom StyleJSON uses custom TileJSON which doesn't have `attribution` property. 
```javascript
var map = new mapboxgl.Map({
  style: "customStyle", // which means the file is not served from :mapbox
  attributionControl: false,
});
```
 2. add attribution control with `customAttribution`.
```javascript
var customAttribution = new mapboxgl.AttributionControl({
  customAttribution: "<a>abc</a>",
});
map.addControl(customAttribution);
```
 3. remove the attribution control.
```javascript
map.removeControl(customAttribution);
```
4. add the attribution control again.
```javascript
map.addControl(customAttribution);
```
### Expected Behavior
Attribution should be added again.

### Actual Behavior
Attribution wrapper is created but content's empty.

I suspect the attribution always empty from `source` so [here](https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/control/attribution_control.js#L164) is always `true`.

kkaefer:
Thanks for reporting this bug! Fix is in https://github.com/mapbox/mapbox-gl-js/pull/9052.

PR #9052, opened 2019-12-02
Reset internal attribution control cache when removing from map

kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/9051

When creating an attribution control, we're caching the HTML string in the AttributionControl object. That cache wasn't reset when the control was removed from the map. Upon re-adding, we incorrectly assumed that the HTML was already up-to-date because our internal cache said so. This patch fixes it by clearing the internal cache so that attribution shows up properly when re-adding the control.

h/t @cs09g


kkaefer:
`<changelog>Fix attribution control when repeatedly removing and adding it.</changelog>`

PR #9057, opened 2019-12-03
Correct mouse/touch position for elements scaled via CSS transform

kkaefer:
Inspired by https://github.com/mapbox/mapbox-gl-js/pull/9014, I looked at correcting mouse/touch interaction when the Map is transformed via the CSS `transform: scale()` property. It scales the screen pixel coordinates to logical map coordinates.

This is distinct from pixelRatio.

I tested in Safari, Chrome and Firefox on macOS.

### DIFF
--- a/src/css/mapbox-gl.css
+++ b/src/css/mapbox-gl.css
@@ -20,6 +20,11 @@
     background-color: salmon;
 }
 
+.mapboxgl-canvas-container {
+    width: 100%;
+    height: 100%;
+}
+
 .mapboxgl-canvas-container.mapboxgl-interactive,
 .mapboxgl-ctrl-group button.mapboxgl-ctrl-compass {
     cursor: -webkit-grab;
--- a/src/util/dom.js
+++ b/src/util/dom.js
@@ -104,22 +104,26 @@ DOM.suppressClick = function() {
 };
 
 DOM.mousePos = function (el: HTMLElement, e: MouseEvent | window.TouchEvent | Touch) {
-    const rect = el.getBoundingClientRect();
+    const rect = el.getBoundingClientRect(),
+        factorX = rect.width ? el.offsetWidth / rect.width : 1,
+        factorY = rect.height ? el.offsetHeight / rect.height : 1;
     const t = window.TouchEvent && (e instanceof window.TouchEvent) ? e.touches[0] : e;
     return new Point(
-        t.clientX - rect.left - el.clientLeft,
-        t.clientY - rect.top - el.clientTop
+        (t.clientX - rect.left) * factorX - el.clientLeft,
+        (t.clientY - rect.top) * factorY - el.clientTop
     );
 };
 
 DOM.touchPos = function (el: HTMLElement, e: TouchEvent) {
     const rect = el.getBoundingClientRect(),
+        factorX = rect.width ? el.offsetWidth / rect.width : 1,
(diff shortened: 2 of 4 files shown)


jingsam:
How css transform affects `el.clientLeft` ? Does it return the original size or tranformed size?

kkaefer:
Good catch, @jingsam. `clientLeft` does not have the transform applied, so I moved it to outside the multiplication. However, this only matters if a user applies a border or scrollbar to the `<div class="mapboxgl-canvas-container"></div>` that we use for tracking events, and I'm having a hard time seeing why anyone would do that.

kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/6079

kkaefer:
Previous attempt to solve this was discarded in https://github.com/mapbox/mapbox-gl-js/pull/8017. I tested this PR on iOS Safari and it works as expected.

kkaefer:
Unfortunately, [JSDom doesn't have a layout engine](https://github.com/jsdom/jsdom/issues/1322), so our unit tests fail due to division by zero.

jingsam:
@kkaefer great work. How is it affect performance?

kkaefer:
> Do you have an idea on how to fix the tests?

We could either mock the getters, but that kind of means we're only testing our mock values. I think a better approach would be to switch unit tests to run in a browser and e.g. use puppeteer on CI to run them.

asheemmamoowala:
Now that we have selenium tests (#9245), what are the next steps needed to land this PR?

kkaefer:
> what are the next steps needed to land this PR?

Write a selenium test that double clicks/taps, or sends mouse wheel events at a particular location and check that the resulting zoom correctly reflects the location when the container is zoomed via CSS.

karimnaaji:
@ryanhamley still good for release-d?

ryanhamley:
@kkaefer I tried to add a Selenium test and it works locally but fails on CI. Do you have any insight into what I might be doing wrong?

kkaefer:
@ryanhamley it could be related to pixel ratio. I think our Selenium tests on CI are running at 1&times;, while your tests locally are running at 2&times;.

sgolbabaei:
Is this good for release-e? @ryanhamley

ryanhamley:
@sgolbabaei I need to take a look at this tomorrow or Friday. The Selenium test I wrote fails on CI and I'm not sure how to get it to pass locally and on CI if the problem is the pixel ratio. Looks like it needs to be rebased against master too

ryanhamley:
After playing around with this some more, I'm not convinced that this fixes the issue. Or at least, I'm running into some confusing behavior. Compare [this demo](https://jsfiddle.net/fxi/gb5jqssu/) showing the bug and the behavior in [this forked version](https://jsfiddle.net/rsh412/59kLsegb/) which uses this branch. The click is correct, but zooming behaves differently. Similarly, in [this demo](https://jsbin.com/zilinadise/edit?html,output) which uses this branch, zooming appears to not work as expected. When I have basically the same set up in a local debug page though, things work as expected.

ryanhamley:
After consulting with the team, it looks like @karimnaaji is working on implementing dynamic scaling in a way that would make this change unnecessary. I'm going to close this PR in anticipation of that work being a more robust solution. If we find that the need for this still exists in the future, we can revisit it.

adamcbrewer:
@ryanhamley @karimnaaji is there any update with regards to dynamic scaling?

PR #9062, opened 2019-12-04
Raster optimization: how deep to lookup for parent for fade

astojilj:
Avoid sequential lookup from current zoom all the way to zoom level 0 as it affects performance.

Do the same approach as used in SourceCache.update when calculating retained raster tiles `const minCoveringZoom = Math.max(idealZ - SourceCache.maxOverzooming, sourceCache.getSource().minzoom);`.

If layer.paint.get('raster-fade-duration') is 0, no need to calculate parent since it is anyway ignored in getFadeValues.

The optimization is done based on measurements in https://github.com/mapbox/mapbox-gl-js/pull/8979#issuecomment-561582780

Before the patch:

Screen Shot 2019-12-04 at 11 59 28

After the patch:

Screen Shot 2019-12-04 at 11 57 47

### DIFF
--- a/src/render/draw_raster.js
+++ b/src/render/draw_raster.js
@@ -8,9 +8,9 @@ import StencilMode from '../gl/stencil_mode';
 import DepthMode from '../gl/depth_mode';
 import CullFaceMode from '../gl/cull_face_mode';
 import {rasterUniformValues} from './program/raster_program';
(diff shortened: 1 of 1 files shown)


astojilj:
Closing this temporarily while re-verifying the impact.

PR #9076, opened 2019-12-05
Edit convert functions to output linear interpolation

### LINKED ISSUE #6840: Expressions convert functions: output linear interpolation?
In the [functions to convert deprecated functions to expressions](https://github.com/mapbox/mapbox-gl-js/blob/a5a8dfe5b90737dbb295eaca7f41c667ae4060a8/src/style-spec/function/convert.js), we are converting functions with type `exponential` and base `1` to interpolation expressions with type `["exponential", 1]`. While this is the most exact conversion, would it be more useful to convert these to type `["linear"]`?

/cc @anandthakker @jfirebaugh @tristen @aparlato

Jo-IE:
This PR closes #6840. I edited the  `convertPropertyFunction` and the `convertZoomFunction` to output linear interpolations when `base` is 1. 

I also ran `yarn lint --fix` locally to fix linting errors. I am also getting some JSDocs warnings after running `yarn lint --fix.`

### DIFF
--- a/src/style-spec/function/convert.js
+++ b/src/style-spec/function/convert.js
@@ -1,27 +1,35 @@
 // @flow
 
-import assert from 'assert';
-import type {StylePropertySpecification} from '../style-spec';
+import assert from "assert";
+import type {StylePropertySpecification} from "../style-spec";
 
 export default convertFunction;
 
 function convertLiteral(value) {
-    return typeof value === 'object' ? ['literal', value] : value;
+    return typeof value === "object" ? ["literal", value] : value;
 }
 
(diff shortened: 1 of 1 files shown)


arindam1993:
Hey @Jo-IE 👋 , thanks a lot for your contribution! Much appreciated!

I think your logic is good overall, but could you please make your change without having run `yarn lint --fix`? The extra diffs are making the change hard to understand with the extra new-lines.

I'd recommend running `yarn run lint` only and fixing only the errors manually.

Jo-IE:
Thank you @arindam1993! I will make those adjustments

Issue #9082, opened 2019-12-06
getCenter is not correct after resize the container programatically


cs09g:
**mapbox-gl-js version**:
v1.5.0

### Question
When the container is resized by css programatically, the center of the map is visually changed. but when we call `getCenter`, it returns previous center of the map. It's the point of breaking consistency until we call `resize`. Cannot just get the center of the literally visible map? or Cannot watch the container size is changing to call `resize` automatically?

### Links to related documentation
https://jsfiddle.net/cs09g/q3x1ohun/17/

kkaefer:
I believe this is working as intended. Here's the map view before resizing:
![image](https://user-images.githubusercontent.com/52399/70314837-8e114680-1818-11ea-8acf-b61db810bd30.png)

Here's the map after resizing:

![image](https://user-images.githubusercontent.com/52399/70315018-fa8c4580-1818-11ea-9077-f23a9d1970ea.png)

The lat/lon of the map's center remains unchanged.

kkaefer:
Sorry, I can actually reproduce this issue in some circumstances.

andrewharvey:
It is documented at https://docs.mapbox.com/mapbox-gl-js/api/#map#resize that

> This method must be called after the map's container is resized programmatically or when the map is shown after being initially hidden with CSS.

cs09g:
`resize` method is can be called manually when the map's hidden with CSS but can't be called automatically by mapbox if the container's visible size is changed?

andrewharvey:
> resize method is can be called manually when the map's hidden with CSS but can't be called automatically by mapbox if the container's visible size is changed?

@cs09g My understanding is there is no reliable cross browser (for the GL JS support browsers at least) to detect when a container's size changes, so you need to manually call map.resize() when you change the container size. This is documented in the API.

PR #9083, opened 2019-12-06
Use ResizeObserver to trigger a canvas resize when the container size changes

kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/9082 in some browsers by using the new [ResizeObserver API](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to track when the map container's size is changing.

Previously, we'd only trigger a canvas resize when the window size was changing. However, when modifying the container size with CSS alone, a manual call to `Map#resize()` is necessary to ensure that the canvas is resized accordingly. If that doesn't happen, the canvas remains at its previous size, but it's either too small, or the protruding parts are hidden because the container has `overflow: hidden`. This results in a discrepancy between what a user sees and what our APIs, like `getCenter()`, report.

To resolve this issue, we need to trigger `Map#resize()` automatically when the container size changes. The `ResizeObserver` API is available in Chrome 64 and Firefox 69. Safari 13 supports it theoretically, but it's disabled by default and can only be enabled in the Developer menu.

Using this API fixes the isssue and eliminates the need to call Map#resize()` manually, but it doesn't work in all browsers supported by Mapbox GL. There's a polyfill available in https://github.com/juggle/resize-observer. However, it essentially uses polling via `requestAnimationFrame` for the first few frames, then adds a global `MutationObserver` (which is supported by all browsers in our list) in addition to a few other event handlers as a heuristic to capture all events that could lead to a dimension change.

I'm not sure what the best option is here:
- Do nothing and rely on users calling `resize()` manually
- Use `ResizeObserver` without polyfill
- Use `ResizeObserver` with polyfill (fairly large file size impact for a minor bug like this)
- Rework our accessor methods like `getCenter()` to perform a resize if needed (performance impact?)

/cc @mapbox/gl-js

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1959,6 +1959,10 @@ class Map extends Camera {
         const dimensions = this._containerDimensions();
         this._resizeCanvas(dimensions[0], dimensions[1]);
 
+        if (this._trackResize && window.ResizeObserver) {
+            new window.ResizeObserver(() => this.resize()).observe(container);
+        }
+
         const controlContainer = this._controlContainer = DOM.create('div', 'mapboxgl-control-container', container);
         const positions = this._controlPositions = {};
         ['top-left', 'top-right', 'bottom-left', 'bottom-right'].forEach((positionName) => {


andrewharvey:
I think it's fine to expect users to call resize manually when they change the container, since this is current practice and is documented at https://docs.mapbox.com/mapbox-gl-js/api/#map#resize. This seems like the simplest solution until the supported platforms catchup with ResizeObserver support. That said, I'm not against the other options.

kkaefer:
Yeah; I was thrown off by the fact that the size gets adjusted _sometimes_, e.g. when resizing the window. It should either be never or always.

ryanhamley:
using `ResizeObserver` without the polyfill seems like a good progressive enhancement to me

kkaefer:
My concern with this particular progressive enhancement is that it leads to users testing in Chrome and missing situations where they have to call `Map#resize()` because it works in Chrome without it.

kkaefer:
Not going to merge this because it introduces inconsistency across browsers and is more than just a progressive enhancement.

stanjdev:
I'm using React.
I put this into my handleClick function on a button that I am using to toggle the map visible and invisible. I placed it AFTER my setState boolean logic to trigger the map's visibility

    setTimeout(() => {
      window.dispatchEvent(new Event('resize'))
    }, 0)

andrewharvey:
Now we are in 2024 I think it's worth reopening this. ResizeObserver has been supported in Safari since 13.1 which was released in 2020 https://caniuse.com/resizeobserver

I still run into this issue often enough and can't see any reason not to implement ResizeObserver directly within the library.

cc @stepankuzmin @mourner

andrewharvey:
reopening

PR #9091, opened 2019-12-10
Fix lazy load of RTL text plugin on GeoJSON sources

### LINKED ISSUE #9075: RTL plugin lazy loading doesn't work on GeoJSON source
**mapbox-gl-js version**: 1.60

**browser**: Chrome latest stable

### Steps to Trigger Behavior
 1. Set RTL text lazy loading to true 
```
mapboxgl.setRTLTextPlugin(
  window.location.origin + '/static/js/@mapbox/mapbox-gl-rtl-text/mapbox-gl-rtl-text.min.js',
  err => {
    console.log(err)
  },
  true
)
```
 2. Load a map without RTL text
 3. Add text which contains RTL characters, for example "این یک متن فارسی است که در سمت راست قرار می‌گیرد"

### Expected Behavior
Network panel should load the RTL plugin + the matching PBF files

### Actual Behavior
RTL plugin does not get loaded + matching PBF files are not loaded either
Nothing is visible on the map in place of the text.

arindam1993:
This PR moves the responsibility of triggering the download of the `rtl-text-plugin` from `VectorTileSource` to `Tile`. This fixes #9075 

Here's a jsfiddle with a working demo. ( wait 5 sec for the async geojson to load)
https://jsfiddle.net/egyoam5t/6/

## Launch Checklist

 <changelog> Fix `rtl-text-plugin` deferred loading not triggering for labels loaded from GeoJSON sources.</changelog>


test/unit/data/symbol_bucket.test.js
```diff
@@ -98,3 +84,25 @@ test('SymbolBucket integer overflow', (t) => {
...
 });
+
+test('SymbolBucket detects rtl text', (t) => {
+    const rtlBucket = bucketSetup('مرحبا');
+    const ltrBucket = bucketSetup('hello');
+    const options = {iconDependencies: {}, glyphDependencies: {}};
+    rtlBucket.populate([{feature}], options);
+    ltrBucket.populate([{feature}], options);
+
+    t.ok(rtlBucket.hasRTLText);
```
ryanhamley: should we also test `t.ok(!rtlBucket.isEmpty());`? It seems like that's more of what we want to know, since `hasRTLText` is used to determine if the bucket is empty or not.

test/unit/source/tile.test.js
```diff
@@ -263,6 +264,27 @@ test('expiring tiles', (t) => {
     t.end();
 });
 
+test('rtl text detection', (t) => {
+    t.test('Tile#hasRTLText is true when a tile loads a symbol bucket with rtl text', (t) => {
+        const tile = new Tile(new OverscaledTileID(1, 0, 1, 1, 1));
+        const symbolBucket = createSymbolBucket('test', 'Test', 'مرحبا', new CollisionBoxArray());
+        symbolBucket.hasRTLText = true;
```
ryanhamley: Why do we have to manually set `symbolBucket.hasRTLText = true`? Shouldn't that happen automatically like in the symbol bucket tests above? It seems like this test is kind of trivial if we manually set this to `true` since in `tile.js`, it basically just is `this.hasRTLText = symbolBucket.hasRTLText`.

arindam1993: ahh, its because `hasRTLText` can only be determined after bucket.populate has been called.
I'm only using `createSymbolBucket` to create a stub of a symbol bucket here. I've added a comment to explain.

test/unit/data/symbol_bucket.test.js
arindam1993: Added a separate test in https://github.com/mapbox/mapbox-gl-js/pull/9091/commits/315742d49d9017da10073566a279664c11c61e63

test/unit/source/tile.test.js
ryanhamley: There's a couple typos in the comment

Issue #8214, opened 2019-05-03
Locate the user behavior with WebKit & Blink


gkiosov:
**mapbox-gl-js version**: 0.54.0

**browser**: Mac OS (10.14.4) Safari (12.1) & Chrome (73.0.3683.103)

We have Angular 7 SPA with the Mabpox map and we noticed the strange behavior of the standard Mapbox locate user button.

### Steps to Trigger Behavior

1. Load a page with the map.
1. Press locate user button. It works fine — we are flying to our position on the map.
1. Change page to another (any) in the current browser tab (change route).
1. Turn back to the page with the map in the current browser tab.
1. Press locate user button again. Here is the bug: the locate user button is spinning endlessly and we are not flying to our position on the map.
1. Change browser tab without reloading the page with the map.
1. Turn back to the tab with the page with the map and the locate user button stops spinning and we start flying to our position on the map.

The bug exists only on the WebKit (Safari) and Blink (Chrome) browsers.

### Link to Demonstration

I can’t reproduce this bug with jsbin.

mourner:
> I can’t reproduce this bug with jsbin.

I'm afraid we can't help with this unless there's a reproducible minimal live test case.

ryanhamley:
I can't replicate this either but it sounds like it's related to routing in your Angular app. Perhaps navigating away from the map is interfering with the geolocation attempt.

gkiosov:
> > I can’t reproduce this bug with jsbin.
> 
> I'm afraid we can't help with this unless there's a reproducible minimal live test case.

Here is the vanilla js example — https://kiosov.pro/tmp/mapbox/locateuser.html

### Steps to Trigger Behavior

1. Click «show map 1» link.
1. Click the locate user button on the map 1.
1. Click «show map 2» link.
1. Click the locate user button on the map 2. You will see the bug I wrote above.
1. Change the browser tab and then return back on the tab with the example.

ryanhamley:
I can replicate the issue with that example. Thanks for putting that together!

gkiosov:
> I can replicate the issue with that example.

Sweet. Is there any chance to know how fast it will be fixed? Kinda critical for us. Thank you anyway!

ryanhamley:
No, there's no timeline for a fix on our end. We'd welcome a PR if you want to try fixing it yourself.

gkiosov:
[...]

PR #9092, opened 2019-12-11
Immediately update map when more than one geolocation watch event is triggered

ryanhamley:
## Launch Checklist

    - fixes https://github.com/mapbox/mapbox-gl-js/issues/8214
    - in situations where more than one geolocation watch listener is set up (i.e. when `trigger()` is called with `trackUserLocation: true` so that `geolocation.watchPosition()` is used), the second listener would not return `onSuccess` until a more refined geolocation position became available, which resulted in potentially minutes of inactivity, creating the impression that the control was not working. this change forces geolocation to immediately return the most recent cached position so that the control always updates the map right away.
    - the above conditions seem to be possible when there are multiple maps on a screen or when a map component in a React/Angular/Vue app is not destroyed when navigating to another "page" before navigating back to the map
 - [x] `<changelog>Fix bug in GeolocateControl where multiple instances of the control on one page would, in some cases, result in the user location not being updated</changelog>`

### DIFF
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -437,17 +437,17 @@ class GeolocateControl extends Evented {
                 this._geolocateButton.setAttribute('aria-pressed', 'true');
 
                 numberOfWatches++;
-                let options;
+                let positionOptions;
                 if (numberOfWatches > 1) {
-                    options = {maximumAge:600000, timeout:0};
+                    positionOptions = {maximumAge:600000, timeout:0};
                     noTimeout = true;
                 } else {
-                    options = this.options.positionOptions;
+                    positionOptions = this.options.positionOptions;
                     noTimeout = false;
                 }
 
                 this._geolocationWatchID = window.navigator.geolocation.watchPosition(
-                    this._onSuccess, this._onError, options);
+                    this._onSuccess, this._onError, positionOptions);
             }
         } else {
             window.navigator.geolocation.getCurrentPosition(
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -279,6 +279,7 @@ class GeolocateControl extends Evented {
                 // this represents a forced error state
                 // this was triggered to force immediate geolocation when a watch is already present
                 // see https://github.com/mapbox/mapbox-gl-js/issues/8214
+                // and https://w3c.github.io/geolocation-api/#example-5-forcing-the-user-agent-to-return-a-fresh-cached-position
                 return;
             } else {
                 this._setErrorState();
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -53,6 +53,9 @@ function checkGeolocationSupport(callback) {
     }
 }
 
+let numberOfWatches = 0;
+let noTimeout = false;
+
 /**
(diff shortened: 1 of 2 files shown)


src/ui/control/geolocate_control.js
```diff
@@ -423,8 +436,18 @@ class GeolocateControl extends Evented {
                 this._geolocateButton.classList.add('mapboxgl-ctrl-geolocate-waiting');
                 this._geolocateButton.setAttribute('aria-pressed', 'true');
 
+                numberOfWatches++;
+                let options;
```
andrewharvey: Just a personal preference, but I'd name this positionOptions so it's a bit clearer it's a https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions object.

src/ui/control/geolocate_control.js
```diff
@@ -53,6 +53,9 @@ function checkGeolocationSupport(callback) {
     }
 }
 
+let numberOfWatches = 0;
+let noTimeout = false;
```
andrewharvey: What confused me when reviewing is I thought these would be scoped for each control, but it seems they are shared globally across the page (which is why they work). I think it's worth adding a comment to explain what these are for and what they represent, and mention something about tracking number of watches on the page, even when multiple maps and geolocate controls may be present.

andrewharvey:
Thanks for digging into this issue @ryanhamley and finding a solution!

I haven't investigated but do you know if a watchPosition registered externally to Mapbox could also lead to the same scenario this PR aims to address? I think if it does and this PR doesn't fix it, it's worth adding a note to the documentation to warn users that other geolocate.watchPosition's may interfere with this control.

andrewharvey:
>  <changelog>Immediately update map when more than one geolocation watch event is triggered</changelog>

Suggestion: Fix bug in GeolocateControl where multiple instances of the control on one page would, in some cases, result in the user location not being updated.

andrewharvey:
Do you think it's possible to add a unit test for this to ensure we don't regress?

src/ui/control/geolocate_control.js
ryanhamley: yeah they have to be scoped to the global control state so that we can track when multiple watches are being created. i will update with a comment

ryanhamley:
Thanks for the review @andrewharvey!

>do you know if a watchPosition registered externally to Mapbox could also lead to the same scenario this PR aims to address?

That's a good question. I assume that an external `watchPosition` would be something that this doesn't handle well and I'm really not sure if it's really possible to handle it well. As far as I know, there's no way to detect if a watch listener has already been set up. I will test it out and if there's no solution, I'll update the docs.

>Do you think it's possible to add a unit test for this to ensure we don't regress?

Yeah, I'll have to think about the best way to do this. I think it could be possible.

ryanhamley:
I tested having a generic `watchPosition` set on a page outside of the Mapbox context and it didn't seem to affect this solution in any way so I think we're ok there.

I played around with adding a unit test though and I'm not sure it's possible given the current setup. Our `createMap` function takes the test object `t` as an argument and then calls `t.stub`. If we try to do this more than once to create multiple maps, the test throws an error that `t.stub is not a function`. I'm not entirely sure why it does that but it seems that it might not be possible to create multiple maps in a single test.

PR #9097, opened 2019-12-11
Use husky to run eslint + flow as pre-commit hook

kkaefer:
Adds [husky](https://github.com/typicode/husky) and [lint-staged](https://github.com/okonet/lint-staged) to run pre-commit hooks. We're running `eslint` and Flowtype on the affected files, as well as eslint on HTML pages and stylelint on CSS files.

This is what a denied commit looks like:

image

And this is what a successfully checked commit looks like:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,10 +64,12 @@
     "flow-bin": "^0.100.0",
     "gl": "~4.3.3",
     "glob": "^7.1.4",
+    "husky": "^3.1.0",
     "is-builtin-module": "^3.0.0",
     "jsdom": "^13.0.0",
     "json-stringify-pretty-compact": "^2.0.0",
     "jsonwebtoken": "^8.3.0",
+    "lint-staged": "^9.5.0",
     "mock-geolocation": "^1.0.11",
     "node-notifier": "^5.4.3",
     "npm-run-all": "^4.1.5",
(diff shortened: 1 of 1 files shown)


kkaefer:
It takes a few seconds, in particular flow can take some time if it needs to reindex. You can bypass the pre-commit checks with `--no-verify`.

andrewharvey:
> Would this mean we can no longer do multiple WIP commits in a PR while working on a complex feature, having to perfect flow types in each commit?

I agree, I do this all the time, especially since we squash PRs when merging. Particularly for flow types which it's common for new contributors to have zero experience with. I'd still rather see PRs with flow issues than no PR.

> It takes a few seconds, in particular flow can take some time if it needs to reindex. You can bypass the pre-commit checks with --no-verify.

While it's great you can bypass, I worry that it's one extra roadblock for new contributors, if they don't know about --no-verify, I wouldn't want them being blocked from pushing a PR since a pushed PR is the only way we'd be able to provide help to get the flow issues fixed.

Generally I find the current way these run when you push commits is perfect, the contributor isn't blocked while in development, but there's still that final check which prevents issues creeping into the master codebase.

kkaefer:
You're all right, let's not do this.

PR #9103, opened 2019-12-12
Add check for ImageBitmap type for texture uploads

### LINKED ISSUE #9100: raster broken by OffscreenCanvas on chrome on MacMini
After 85517f0560b199cc5c2e09e50085aabb6de27b78, debug/satellite.html shows black tiles on macos 10.15.1 on Chrome. No error in console.

Works fine on Firefox and Safari on the same hardware. Render tests pass.
 
Mac Mini 2018 3.2 GHz 6-Core Intel Core i7 (Intel UHD Graphics 630 1536 MB)
Chrome Version 78.0.3904.108 (Official Build) (64-bit)

Expected (works fine when `offscreenCanvasSupported()` returns `false`):
Screen Shot 2019-12-12 at 20 05 02

Actual:

Screen Shot 2019-12-12 at 20 04 09

cc @arindam1993 @kkaefer

arindam1993:
Fixes #9100 .

There was a mistake in `Texture` which caused `ImageBitmap`'s to be sent to the wrong `gltexImage2D` call.

### DIFF
--- a/src/render/texture.js
+++ b/src/render/texture.js
@@ -1,7 +1,7 @@
 // @flow
 
 import window from '../util/window';
-const {HTMLImageElement, HTMLCanvasElement, HTMLVideoElement, ImageData} = window;
+const {HTMLImageElement, HTMLCanvasElement, HTMLVideoElement, ImageData, ImageBitmap} = window;
 
 import type Context from '../gl/context';
 import type {RGBAImage, AlphaImage} from '../util/image';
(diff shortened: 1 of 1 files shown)


arindam1993:
> Should we add a test of some type to ensure this behavior doesn't regress?

Flow is the best test for something like this, I'm really surprised that even without the `ImageBitmap` added to `TextureImage` union type, there were no complaints.

arindam1993:
Dug a bit more, and I think its a Flow bug, the `img` variable here https://github.com/mapbox/mapbox-gl-js/blob/fc8ff35a2a6cf10651a2c90c3c753bf785f39c04/src/source/raster_tile_source.js#L134
has the `HTMLImageElement | ImageBitmap` type, but Flow still allowed passing it to `Texture` even though `Texture` didnt accept `ImageBitmap` before this PR.

PR #9107, opened 2019-12-13
Edit convert functions to output linear interpolation

### LINKED ISSUE #6840: Expressions convert functions: output linear interpolation?

Jo-IE:
This PR closes #6840. I edited the convertPropertyFunction and the convertZoomFunction to output linear interpolations when base is 1.

I submitted #9076 earlier but I am resubmitting the PR here without running `yarn lint --fix`

### DIFF
--- a/src/style-spec/function/convert.js
+++ b/src/style-spec/function/convert.js
@@ -153,13 +153,11 @@ function convertPropertyFunction(parameters, propertySpec, stops) {
         ];
     } else if (type === 'exponential') {
         const base = parameters.base !== undefined ? parameters.base : 1;
-        const expression =
-
-                [
-                    getInterpolateOperator(parameters),
-                    base === 1 ? ["linear"] : ["exponential", base],
-                    ["number", get]
-                ];
+        const expression = [
+            getInterpolateOperator(parameters),
+            base === 1 ? ["linear"] : ["exponential", base],
+            ["number", get]
+        ];
 
         for (const stop of stops) {
             appendStopPair(expression, stop[0], stop[1], false);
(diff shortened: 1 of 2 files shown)


kkaefer:
Thanks, this looks much easier to read! Looking at our test builds, it looks like the unit test still expects the old-style exponential function. You can fix this by changing the unit test's expectation, then run just that test locally with `build/run-node test/unit/style-spec/migrate.test.js`.

For next time: you can also force-push to an existing branch (with `git push --force`); Github will then update the original issue and there's no need to open a new pull request.

Jo-IE:
thank you @kkaefer! I will do that and make another commit

src/style-spec/function/convert.js
```diff
@@ -153,7 +153,18 @@ function convertPropertyFunction(parameters, propertySpec, stops) {
...
+                [
+                    getInterpolateOperator(parameters),
+                    ["linear"],
+                    ["number", get]
+                ] :
+                [
+                    getInterpolateOperator(parameters),
+                    ["exponential", base],
+                    ["number", get]
+                ];
```
mourner: Small nit: we could make this a more compact by using the ternary only for the `["linear"]` and `["exponential", base]` so that there's less duplication.

src/style-spec/function/convert.js
```diff
@@ -177,7 +188,14 @@ function convertZoomFunction(parameters, propertySpec, stops, input = ['zoom'])
...
         const base = parameters.base !== undefined ? parameters.base : 1;
-        expression = [getInterpolateOperator(parameters), ['exponential', base], input];
+        expression =
+            base === 1 ?
+                [getInterpolateOperator(parameters), ["linear"], input] :
+                [
+                    getInterpolateOperator(parameters),
+                    ["exponential", base],
+                    input
+                ];
```
mourner: Here too.

mourner:
@Jo-IE also, no need to resubmit the PR — you can just add commits to this branch.

Jo-IE:
Thank you @mourner i will make those changes. 
I resubmitted the initial PR to keep it simple because running `yarn lint --fix `changed a bunch of other files locally for me.

PR #9120, opened 2019-12-17
Strip more keys from the style spec

mourner:
Strip more unused keys from the style spec JSON, reducing the build size a little (-1.21KB minified).


mourner:
Nevermind, I made some wrong assumptions here. Not worth exploring.

PR #9122, opened 2019-12-17
Fix CSP errors caused by using a blobURL to load the rtl-text-plugin

### LINKED ISSUE #9118: rtl-text-plugin fails to load if CSP for website doesn't allow script-src: blob:
**mapbox-gl-js version**: `1.6.0`

**browser**: All

### Steps to Trigger Behavior

 1. Set a page's CSP to not allow `script-src: blob:`

### Cause

Side effect of #8865 , which made a change so that the Main thread now requests the `rtl-text-plugin`, and generates a blob-url. It then sends the blobURL over to the worker for loading with `importScripts`. Prior to #8865  each worker would directly use the main url to load the rtl text plugin via `importScripts` directly, which would lead to multiple parallel network requests and slower load times.

 However now, if the parent page doesnt allow `script-src: blob:` the `importScripts` for the blobURL will fail and causes rtl text to not be rendered.

arindam1993:
This fixes #9118 , instead of turning the plugin into a blob, and sending a blobURL to the worker, this fetches once on the main thread, and once the fetch is done directly send the host url to the workers. This solves the CSP issue since `importScripts` is called on the `https` url directly, and doesn't cause the workers to race as before since downloading it once on the main thread warms up the browser cache with the data.

## Launch Checklist

`<changelog>Fixes CSP violation errors when using rtl-text-plugin</changelog>`

### DIFF
--- a/src/source/rtl_text_plugin.js
+++ b/src/source/rtl_text_plugin.js
@@ -3,7 +3,6 @@
 import {Event, Evented} from '../util/evented';
 import {getArrayBuffer} from '../util/ajax';
 import browser from '../util/browser';
-import window from '../util/window';
 import assert from 'assert';
 import {isWorker} from '../util/util';
 
@@ -17,8 +16,7 @@ const status = {
 
 export type PluginState = {
     pluginStatus: $Values<typeof status>;
-    pluginURL: ?string,
-    pluginBlobURL: ?string
+    pluginURL: ?string
 };
 
 type ErrorCallback = (error: ?Error) => void;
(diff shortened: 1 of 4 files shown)


arindam1993:
We were already downloading on the main thread, but then we used to turn the contents into a bloburl and send the blob url to the workers to prevent the workers from making independent network calls.

This PR skips the bloburl creation part, and the workers just fetch from the host url directly. But since the main thread only informs the workers to start loading the plugin once the plugin download has finished, the workers always fetch from the browser cache.

Issue #8999, opened 2019-11-20
queryRenderedFeatures Exception:  Cannot read property 'sub' of undefined


cdhuangyong:
**mapbox-gl-js version**:
1.5.0
**browser**:
chorme
### Steps to Trigger Behavior

 1.  Layer type: fill-extrusion
      source geometry type :  MultiPolygon
 2.  When I call the method “queryrenderedfeatures” to query features，Sometimes an error is reported, like the title

  I found the rule that when the map level is relatively large, there will be no problem, while when the map level is relatively small, there will be some problems，I guess the simplification algorithm takes the length of the point less than 4

The location of the exception error：
（see:  https://github.com/cdhuangyong/issues/blob/master/error20191120.png?raw=true）
  **projectedFace[3] Sometimes undefined**

cdhuangyong:
This problem will appear when my polygon data is simplified. I hope to solve this problem as soon as possible. Thank you

ansis:
It looks like the image you're sharing is in a private repository. Could you share it publicly? A https://jsfiddle.net/ example that we can run would help us diagnose the issue

cdhuangyong:
@ansis This is a demo I wrote. When the mouse moves into the face, it will throw an exception
[https://jsfiddle.net/cdhuangyong/hgopj9uw/](url)

PR #9138, opened 2019-12-20
fix extrusion querying for polygons with coincident points

ansis:
fix extrusion querying for polygons with coincident points and for polygons with less than four points (#8999).

## Launch Checklist

 - [x] `<changelog>Fix query fill-extrusions made from polygons with coincident points and polygons with less than four points.</changelog>`

### DIFF
--- a/src/style/style_layer/fill_extrusion_style_layer.js
+++ b/src/style/style_layer/fill_extrusion_style_layer.js
@@ -67,7 +67,7 @@ function dot(a, b) {
     return a.x * b.x + a.y * b.y;
 }
 
-function getIntersectionDistance(projectedQueryGeometry: Array<Point>, projectedFace: Array<Point>) {
+export function getIntersectionDistance(projectedQueryGeometry: Array<Point>, projectedFace: Array<Point>) {
 
     if (projectedQueryGeometry.length === 1) {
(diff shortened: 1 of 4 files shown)


src/style/style_layer/fill_extrusion_style_layer.js
```diff
@@ -77,9 +77,20 @@ function getIntersectionDistance(projectedQueryGeometry: Array<Point>, projected
...
+        const a = projectedFace[i++];
+        let b, c;
+        while (!b || a.equals(b)) {
+            b = projectedFace[i++];
+            if (!b) return Infinity;
+        }
+        while (!c || a.equals(c) || b.equals(c)) {
+            c = projectedFace[i++];
+            if (!c) return Infinity;
+        }
```
arindam1993: Looks good to me, just one question.
Ss numerical stability a concern with with floating point rounding errors with the `equals` checks? Would doing a distance check with a small epsilon be safer?

ansis: Hmm... good question. I think the minimum distance (when not equal) should be safely large because vector tile positions are stored as (fairly small) integers. Also, any precision issues would just result in sorting errors which wouldn't be the end of the world.

I think it makes sense to leave it as is for now.

Issue #9148, opened 2019-12-29
Disable map rotation doesn't work with url hash


jessecrouch:
When using [all of the settings that disable map rotation](https://docs.mapbox.com/mapbox-gl-js/example/disable-rotation/), the map is still able to rotate if the user changes the hash in the url (ex: https://localhost?10/30.261/-97.717/-97.7)

Expected: User should not be able to rotate the map by any means. Bearing should lock.

Ideally there would be a map creation setting such as bearingLock, a permanent version of resetNorth.

ryanhamley:
I have a PR up for this https://github.com/mapbox/mapbox-gl-js/pull/9156

PR #9156, opened 2020-01-02
Prevent changing bearing via URL hash when rotation is disabled

ryanhamley:
## Launch Checklist

Fixes https://github.com/mapbox/mapbox-gl-js/issues/9148

    - This PR prevents the user from changing the bearing of the map via the URL hash if both `touchZoomRotate` and `dragRotate` are disabled
 - [x] `<changelog>Prevent changing bearing via URL hash when rotation is disabled</changelog>`

### DIFF
--- a/src/ui/hash.js
+++ b/src/ui/hash.js
@@ -119,10 +119,11 @@ class Hash {
     _onHashChange() {
         const loc = this._getCurrentHash();
         if (loc.length >= 3 && !loc.some(v => isNaN(v))) {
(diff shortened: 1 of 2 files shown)


src/ui/hash.js
```diff
@@ -119,10 +119,11 @@ class Hash {
     _onHashChange() {
         const loc = this._getCurrentHash();
         if (loc.length >= 3 && !loc.some(v => isNaN(v))) {
+            const bearing = this._map.dragRotate.isEnabled() && this._map.touchZoomRotate.isEnabled() ? +loc[3] : this._map.getBearing() || 0;
```
mourner: Nit: `getBearing()` always returns a number, so there shouldn't be a need for `|| 0`.

PR #9157, opened 2020-01-04
Update symbol-avoid-edges doc copy to acknowledge the existence of global collision detection

danswick:
## Launch Checklist

Updates the `symbol-avoid-edges` copy to address cases where the rule is no longer needed in order to avoid clipping labels. Because not all Mapbox GL clients (or other clients that implement the style spec) support global collision detection, I think it makes sense to add this note to the end of the doc section. 

 - [ ] `<changelog>...</changelog>` 

cc @mapbox/map-design-team

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -1006,7 +1006,7 @@
     "symbol-avoid-edges": {
       "type": "boolean",
       "default": false,
-      "doc": "If true, the symbols will not cross tile edges to avoid mutual collisions. Recommended in layers that don't have enough padding in the vector tile to prevent collisions, or if it is a point symbol layer placed after a line symbol layer.",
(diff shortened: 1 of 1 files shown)


src/style-spec/reference/v8.json
```diff
@@ -1006,7 +1006,7 @@
     "symbol-avoid-edges": {
       "type": "boolean",
       "default": false,
-      "doc": "If true, the symbols will not cross tile edges to avoid mutual collisions. Recommended in layers that don't have enough padding in the vector tile to prevent collisions, or if it is a point symbol layer placed after a line symbol layer.",
+      "doc": "If true, the symbols will not cross tile edges to avoid mutual collisions. Recommended in layers that don't have enough padding in the vector tile to prevent collisions, or if it is a point symbol layer placed after a line symbol layer. When using a client that supports global collision detection, like Mapbox GL JS version 0.42.0 or greater, enabling this property is not needed to prevent clipped labels at tile boundaries.",
```
chloekraw: nit: Instead of “like” I would suggest “i.e.”

danswick: Thanks for the suggestion, @chloekraw, but we try to avoid latin abbreviations https://hey.mapbox.com/documentation/style-guide/grammar-and-mechanics/#abbreviations.

PR #9176, opened 2020-01-13
Update Code of Conduct (v2.0, https links)

mourner:


kkaefer:
What are the differences between 1.2 and 2.0?

mourner:
@kkaefer the biggest change in a new section about enforcement of the rules, and generally tighter wording. Take a look at the full texts in the respective links to compare.

vakila:
Thanks for the update @mourner! 

@kkaefer I was wondering this too! Seems they don't have a changelog but from https://github.com/ContributorCovenant/contributor_covenant/pull/709:

> Changes include:
> 
>   Language updates to emphasize community over project scope
>   Additional examples of positive behavior
>   Clarification around sexual attention and advances
>   Community Impact Guidelines


@mourner  while we're updating this, per https://www.contributor-covenant.org/version/2/0/code_of_conduct#enforcement I think we need to explicitly let people know where & how to report violations or other concerns/questions. Without that the CoC is basically useless. @asheemmamoowala do we have a team email account we could use for that? (Asking folks to tag the team in public won't work for this.)

Also - not a blocker, but - we should add Mapbox to the [adopters list](https://github.com/ContributorCovenant/contributor_covenant/blob/release/static/adopters.csv)! I can do that

vakila:
https://github.com/ContributorCovenant/contributor_covenant/pull/750

PR #9179, opened 2020-01-13
Few fixes and additional API's for performance metrics instrumentation

arindam1993:
This is a follow up of #9035, with a few fixes and some additional API methods that help improve the reliability of collected metrics.
Primarily this adds 3 new methods:
`map.style.disableFetchCancellation()`/`map.style.enableFetchCancellation()`: This lets the test-runner disable request cancellation for a `cache-prewarm` run, which is run prior to the actual measurement runs. This ensures that all possible tiles that are required for the run are precached during the actual measurement run.

`map.style.getNumInflightRequests()`: This is used by the test-runner to wait until all in-flight requests are finished for a `cache-prewarm` run, again ensuring maximum cache-warmth. 🔥 

## Launch Checklist


src/util/ajax.js
```diff
@@ -129,7 +132,9 @@ function makeFetchRequest(requestParameters: RequestParameters, callback: Respon
 
         const requestTime = Date.now();
 
+        _inflightRequestCount++;
         window.fetch(request).then(response => {
+            _inflightRequestCount = Math.max(_inflightRequestCount - 1, 0);
```
ansis: Is the `max` there to handle double-decrements in case both an error handler and .cancel() are called? Would doing something id-based be more reliable?

```
const reqId = nextId++;
activeRequests[reqId] = true;
delete activeRequests[reqId];
infLightRequests = Object.keys(activeRequests).length;
```

arindam1993: yes, that is correct! I thought about that too, but I was trying to keep the instrumentation as low runtime overhead as possible

arindam1993:
> Is this waiting on anything besides the render test?
Nope nothing else and I have still no idea why adding an extra function is breaking esm :(

ryanhamley:
Do we still need this @arindam1993 ?

mourner:
Looks like this is stale now — let's revisit later if it's still relevant.

Issue #8777, opened 2019-09-19
ScaleControl uses inconsistent earth radius


exe-dealer:
**mapbox-gl-js version**: 1.3.1

**browser**: *

Why scale control uses different earth radius than mercator projection code? Is this ok?

https://github.com/mapbox/mapbox-gl-js/blob/2020f3f9ced516086508925411bd098fbb4c3234/src/ui/control/scale_control.js#L130

https://github.com/mapbox/mapbox-gl-js/blob/2020f3f9ced516086508925411bd098fbb4c3234/src/geo/mercator_coordinate.js#L9

ryanhamley:
Hmm, that's a good question. I think the radius in `scale_control` is fine.

From [Wikipedia](https://en.wikipedia.org/wiki/Mercator_projection):
>for small scale (large region) applications the variation may be ignored, and mean values of 6,371 km and 40,030 km may be taken for the radius and circumference respectively

@andrewharvey was there a specific reason you chose 6378137 for the radius in `mercator_coordinate`? Maybe it would be a good thing to pull magic values like this into a file that could be included elsewhere so we can define them once and not have inconsistencies in the code.

andrewharvey:
The figure in `mercator_coordinate` came from the [EPSG:3857 definition](https://www.epsg-registry.org/report.htm?type=selection&entity=urn:ogc:def:crs:EPSG::3857&reportDetail=short&style=urn:uuid:report-style:default-with-code&style_name=OGP%20Default%20With%20Code&title=EPSG:3857).

mourner:
@andrewharvey given how the value is used in `MercatorCoordinate`, I think we should change it to mean radius value rather than the equator one from EPSG:3857. The latter also defines inverse flattening (298.257223563) which defines how this radius [varies across the globe](https://en.wikipedia.org/wiki/Flattening).

andrewharvey:
> @andrewharvey given how the value is used in MercatorCoordinate, I think we should change it to mean radius value rather than the equator one from EPSG:3857. The latter also defines inverse flattening (298.257223563) which defines how this radius varies across the globe.

@mourner Sorry I don't understand. What does "radius value" mean here?

At the equator (ie. along the semi-major axis) the inverse flattening won't matter or affect it right? And then we use the scale factor to account for larger latitudes being larger on the map.

mourner:
[...]

Issue #8614, opened 2019-08-09
Marker shaking easeTo / flyTo - Firefox


pathmapper:
**mapbox-gl-js version**: v1.2.0

**browser**: Firefox 68.0.1 (64bit)

### Steps to Trigger Behavior

 1. Add custom marker to map
 2. On `load` slowly `easeTo`
 3. Open in Firefox

### Link to Demonstration

https://jsbin.com/xubonoyaqu/edit?output

### Expected Behavior

Same behavior as in Chrome, almost no shaking of the marker:

![marker_jitter_chrome](https://user-images.githubusercontent.com/20856381/62797898-33bbdf80-badd-11e9-9706-fa2c98db3a1b.gif)

### Actual Behavior

The marker is shaking a lot:

![marker_jitter_firefox](https://user-images.githubusercontent.com/20856381/62797920-4504ec00-badd-11e9-9043-0a8fe7fd9898.gif)

ryanhamley:
Interesting. I actually can't reproduce this in Firefox 68.0.1 (64 bit). Your example looks the same in Chrome and Firefox on my machine. Out of curiosity, what OS are you on?

pathmapper:
> what OS are you on?

Linux - Debian 9 (stretch) 64-bit

andrewharvey:
I could replicate on Linux with Firefox.

pathmapper:
I could replicate also with

- Firefox 68.0.1 on Windows 10 Pro 64-bit
- Firefox for Android 68.0 on Android 8.1.0

peterqliu:
Since the map moves smoothly but the (DOM) marker is choppy, I suspect this has to do with CSS transform implementations on different platforms. `transform` is hardware accelerated for most.

vanhumbeecka:
After upgrading, the whole 'flyTo' mechanism is broken. There is no standard 'smooth' transition anymore. Change happens instantly without 'flyTo'
Tested in: 
* FireFox 68.0.2 (64-bit) on MacOS
* Brave Version 0.67.125 Chromium: 76.0.3809.100 (Official Build) (64-bit)

Not only in my apps is the transition gone, also on the [official example](https://docs.mapbox.com/mapbox-gl-js/example/flyto/) page of Mapbox it is not working>

ahk:
Just curious do you know if this is a regression?

[...]

PR #9181, opened 2020-01-14
add circle to GeolocateControl showing accuracy of position

Meekohi:
## Motivation
The default GeolocateControl shows the viewer's position as reported by `navigator.geolocation.watchPosition` immediately, but this often has a very high margin of error. The dot is often in the wrong location, which can be confusing or even dangerous depending on the application.

## Implementation
Adds a transparent circle to GeolocateControl showing the accuracy of the position, as reported by `navigator.geolocation.watchPosition`.

Screen Shot 2020-01-14 at 10 30 00 AM

image

## Launch Checklist

 - [ ] <changelog>Add a transparent circle to `GeolocateControl` that shows the reported location accuracy. Toggled by `showAccuracy` option which defaults to `true`.</changelog>


Meekohi:
@mapbox/map-design-team this adds a transparent blue circle around the standard GeolocateControl marker. Currently using `#1da1f233` (shown in png above).

src/css/mapbox-gl.css
```diff
@@ -761,6 +761,13 @@ a.mapboxgl-ctrl-logo.mapboxgl-compact {
     display: none;
 }
 
+.mapboxgl-accuracy-circle {
```
andrewharvey: We should include `user-location-dot` somewhere in this name so it's clear exactly what it relates too.

src/ui/control/geolocate_control.js
```diff
@@ -78,6 +80,7 @@ let noTimeout = false;
  * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
  * @param {Object} [options.fitBoundsOptions={maxZoom: 15}] A [`fitBounds`](#map#fitbounds) options object to use when the map is panned and zoomed to the user's location. The default is to use a `maxZoom` of 15 to limit how far the map will zoom in for very accurate locations.
  * @param {Object} [options.trackUserLocation=false] If `true` the Geolocate Control becomes a toggle button and when active the map will receive updates to the user's location as it changes.
+ * @param {Object} [options.showAccuracyRadius=true] Show a transparent circle approximating the position accuracy as reported by `navigator.geolocation.watchPosition`. Set to `false` to disable.
```
andrewharvey: How about "By default a transparent circle will be drawn around the user location indicating the accuracy of the user's location"? What do you think?

I don't think we need to mention watchPosition,

src/ui/control/geolocate_control.js
```diff
@@ -94,12 +97,15 @@ class GeolocateControl extends Evented {
...
     _container: HTMLElement;
     _dotElement: HTMLElement;
+    _circleElement: HTMLElement;
     _geolocateButton: HTMLButtonElement;
     _geolocationWatchID: number;
     _timeoutId: ?TimeoutID;
     _watchState: 'OFF' | 'ACTIVE_LOCK' | 'WAITING_ACTIVE' | 'ACTIVE_ERROR' | 'BACKGROUND' | 'BACKGROUND_ERROR';
     _lastKnownPosition: any;
     _userLocationDotMarker: Marker;
+    _accuracyCircleMarker: Marker;
```
andrewharvey: Given both markers will always have the same location, do you think it would be better to combine this with the _userLocationDotMarker to have a single Marker? Still having it optional and still using two separate classes for styling.

src/ui/control/geolocate_control.js
```diff
@@ -120,6 +127,7 @@ class GeolocateControl extends Evented {
         this._map = map;
         this._container = DOM.create('div', `mapboxgl-ctrl mapboxgl-ctrl-group`);
         checkGeolocationSupport(this._setupUI);
+        this._map.on('move', this._onMove);
```
andrewharvey: Why do we need to watch for map move? The real world radius in meters to pixel conversion depends on latitude of the user location, not the latitude of the map.

src/ui/control/geolocate_control.js
```diff
@@ -476,6 +498,28 @@ class GeolocateControl extends Evented {
 
 export default GeolocateControl;
 
+function getDistance(latlng1, latlng2) {
```
andrewharvey: I don't think this needs to be addressed within this PR, but eventually we should eliminate the duplication between this and https://github.com/mapbox/mapbox-gl-js/blob/74b15736d414023345abab69b91995d949f7d11e/src/ui/control/scale_control.js#L124-L137 and in turn hopefully address #8777.

src/ui/control/geolocate_control.js
andrewharvey: I see this as just an extension of the showUserLocation option, should we support showUserLocation=false but still showing this accuracy circle? I'm not sure...

I'm on the fence with the default. After all we introduced showUserLocation=true as the default which changed behaviour when people upgraded. While I can understand we should try to avoid to unexpected changes in behaviour for users upgrading, this change is mostly cosmetic and if you're already showing the user location dot would you ever not want to show this accuracy circle?

src/css/mapbox-gl.css
Meekohi: updated: `mapboxgl-user-location-accuracy-circle`

src/ui/control/geolocate_control.js
Meekohi: Yeah I think this is the core question -- are there use cases for disabling the circle entirely? I do think defaulting to it visible makes sense although it is a change. I added the flag anticipating someone might see the change and want to go back to the old behavior (especially if they have done something unusual like customized the dot icon or etc).

I do think there is some value in explaining "how is the size of this circle chosen" in the docs, but can leave it out if there's agreement it isn't needed.

We could name it something friendly: "circle-of-confusion" is common but a bit wordy... the radius shown specifically represents the 95% confidence level in the latitude/longitude.

Maybe in this context just `showAccuracy` is sufficient?

src/ui/control/geolocate_control.js
Meekohi: Personally I think the code will be more complicated trying to force both of these objects to use one Marker: that would introduce some container element that will need to be styled carefully to avoid breaking anything, some additional `if/else` around the flag, and I'd be worried that might introduce some other breaking change related to code I'm not familiar with, whereas I'm pretty confident the current design should be independent. (Obvious caveat I've only been looking at this codebase for 2 days so... let me know ;D)

src/ui/control/geolocate_control.js
Meekohi: Yup -- had a hard time finding documentation on these events. Updated.

src/ui/control/geolocate_control.js
Meekohi: Yeah agreed it's a copy/paste job -- good spot to refactor these into some helpers in the future.

src/ui/control/geolocate_control.js
Meekohi: Updated to
> By default a transparent circle will be drawn around the user location indicating the accuracy (95% confidence level) of the user's location. Set to `false` to disable.

for now.

Meekohi:
Been playing with this more on my iPhone and I think it really needs some CSS animation or something to smooth it out, it's a bit chaotic otherwise -- I'll look into adding this.

Meekohi:
* Checklist asks that some tags be added to this PR, someone with permissions will have to handle that.

* Also not quite sure how to handle the CHANGELOG -- should I add a changelog line to 1.6.1? Start a 1.6.2? Is that something someone else handles when/if the PR is merged? I can't tell in other PRs how this was handled, but see @ryanhamley in the history ;)

Meekohi:
Trying to run the benchmark suite obliterated my entire computer and forced a restart 🤣. It doesn't look to me like any benchmarks actually exercise the GeolocateControl, and adding one DOM element shouldn't have any significant impact to performance, so I'm going to take it off the checklist.

Meekohi:
I'd consider this ready for a formal review. Maybe @mourner can chime in with fresh eyes when they have a chance?

src/ui/control/geolocate_control.js
```diff
@@ -78,6 +80,7 @@ let noTimeout = false;
  * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
  * @param {Object} [options.fitBoundsOptions={maxZoom: 15}] A [`fitBounds`](#map#fitbounds) options object to use when the map is panned and zoomed to the user's location. The default is to use a `maxZoom` of 15 to limit how far the map will zoom in for very accurate locations.
  * @param {Object} [options.trackUserLocation=false] If `true` the Geolocate Control becomes a toggle button and when active the map will receive updates to the user's location as it changes.
+ * @param {Object} [options.showAccuracy=true] If trackUserLocation is true, by default a transparent circle will be drawn around the user location indicating the accuracy (95% confidence level) of the user's location. Set to `false` to disable.
```
andrewharvey: "If trackUserLocation is true" is that true though, trackUserLocation just means that it uses watchPosition to update the location continuously. If trackUserLocation is false it will just do a single location request each time you press the button, so even if trackUserLocation=false you may still want this accuracy circle.

src/ui/control/geolocate_control.js
```diff
@@ -326,9 +350,11 @@ class GeolocateControl extends Evented {
         // when showUserLocation is enabled, keep the Geolocate button disabled until the device location marker is setup on the map
         if (this.options.showUserLocation) {
             this._dotElement = DOM.create('div', 'mapboxgl-user-location-dot');
-
             this._userLocationDotMarker = new Marker(this._dotElement);
 
+            this._circleElement = DOM.create('div', 'mapboxgl-user-location-accuracy-circle');
+            this._accuracyCircleMarker = new Marker(this._circleElement);
```
andrewharvey: Could you set `pitchAlignment: 'map'` so that when the map is pitched the accuracy circle remains flat on the map?

src/ui/control/geolocate_control.js
Meekohi: Good catch this should have been `showUserLocation`

src/ui/control/geolocate_control.js
Meekohi: 😮

andrewharvey:
> Also not quite sure how to handle the CHANGELOG -- should I add a changelog line to 1.6.1? Start a 1.6.2? Is that something someone else handles when/if the PR is merged? I can't tell in other PRs how this was handled, but see @ryanhamley in the history ;)

In that checklist you can just add an entry:

* `<changelog>Your changelog entry here</changelog>`

This is the way to go over editing CHANGELOG directly because the merge conflicts would be a nightmare, so that get's updated when the release is done.

src/ui/control/geolocate_control.js
Meekohi: Well that is pretty cool!

image

andrewharvey:
I tested this out using Chrome DevTools Sensor option and cycling through different locations I notice that as the map smoothly zooms in the accuracy circle will smoothly grow in size as it should, but then near the end it starts to jump around. Is that something you noticed?

src/ui/control/geolocate_control.js
andrewharvey: Yeah I was trying to make the same change to the dot in https://github.com/mapbox/mapbox-gl-js/pull/8929/files#diff-8c498cf163efa60ad84199c531ab7c5aR368-R372 but if you want you could do the same to the dot here.

andrewharvey:
Taking a step back I'm seeing the limitations of using a Marker element for this, even just manually zooming in and out results in jitter and panning the map around near the edge the circle it moves relative to the ground, which it shouldn't.

Normally for large circle the drawing out the circle as a polygon results in a much better experience, however the user location dot was intentionally done as a Marker to avoid mutating the users Style so to be honest I'm not sure the best way forward.

Meekohi:
@andrewharvey huh interesting -- I'd been assuming any 'jumping' I saw was due to the actual reported accuracy changing (actually kind of amusing, moving the phone under my desk immediately decreases the accuracy by ~50% or so), but I can reproduce this with the sensors devtools (I didn't know about those!) and will investigate.

pathmapper:
> I'm seeing the limitations of using a Marker element for this, even just manually zooming in and out results in jitter

This could be related to https://github.com/mapbox/mapbox-gl-js/issues/8614

src/ui/control/geolocate_control.js
mourner: We'll likely forget about it later so let's eliminate duplication now — I think the best place to put it is `LngLat` `distanceTo(lngLat)` method. This would follow the same convention as Leaflet.

src/ui/control/geolocate_control.js
```diff
@@ -476,6 +502,28 @@ class GeolocateControl extends Evented {
...
+}
+
+function updateCircleRadius(map, circle, accuracy) {
+    const y = map._container.clientHeight / 2;
+    const metersPerPixel = getDistance(map.unproject([0, y]), map.unproject([1, y]));
+    const circleDiameter = Math.ceil(2.0 * accuracy / metersPerPixel);
+    circle.style.width = `${circleDiameter}px`;
+    circle.style.height = `${circleDiameter}px`;
+}
+
```
mourner: Nit: for consistency, since we're passing 3 arguments here including the map instance, I think this should live as an `_updateCircle()` method on the control class.

src/ui/control/geolocate_control.js
```diff
@@ -120,6 +127,7 @@ class GeolocateControl extends Evented {
         this._map = map;
         this._container = DOM.create('div', `mapboxgl-ctrl mapboxgl-ctrl-group`);
         checkGeolocationSupport(this._setupUI);
+        this._map.on('zoom', this._onZoom);
```
mourner: When the map is heavily pitched, the scale of the dot will change as you pan vertically, so it might not be enough to subscribe to `zoom` — should we do `move` instead?

src/ui/control/geolocate_control.js
Meekohi: Opened a separate PR for this: https://github.com/mapbox/mapbox-gl-js/pull/9202

src/ui/control/geolocate_control.js
```diff
@@ -78,6 +80,7 @@ let noTimeout = false;
  * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
  * @param {Object} [options.fitBoundsOptions={maxZoom: 15}] A [`fitBounds`](#map#fitbounds) options object to use when the map is panned and zoomed to the user's location. The default is to use a `maxZoom` of 15 to limit how far the map will zoom in for very accurate locations.
  * @param {Object} [options.trackUserLocation=false] If `true` the Geolocate Control becomes a toggle button and when active the map will receive updates to the user's location as it changes.
+ * @param {Object} [options.showAccuracy=true] If showUserLocation is true, by default a transparent circle will be drawn around the user location indicating the accuracy (95% confidence level) of the user's location. Set to `false` to disable.
```
andrewharvey: > If showUserLocation is true, by default a transparent circle will be drawn around the user location indicating the accuracy (95% confidence level) of the user's location. Set to `false` to disable.

I think that should read either "If showUserLocation is true a transparent circle..." or "By default a transparent circle...". I find the current wording a bit confusing.

I think the "If true" makes sense if the default was false, but the "By default..." makes sense if the default is true. That how trackUserLocation and showUserLocation options are worded and I think they are clearer.

andrewharvey: I think "showAccuracyCircle" is clearer than just "showAccuracy"

andrewharvey:
just jotting down some of the open questions raised so far which I think need to be addresses before this is ready

- [ ] option naming (showAccuracy, showAccuracyCircle, etc)
- [ ] default state (on or off)
- [ ] single Marker (for both circle and dot) or two Makers (one for circle one for dot)

src/ui/control/geolocate_control.js
andrewharvey: I'd probably prefer a single Marker since their location is always going to be the same, but I can also see how two Markers is easier to implement and in some ways makes the code easier to manage (though single Marker also makes the Marker update code cleaner).

That said, I'm happy with either option.

ryanhamley:
`showAccuracyArea`? I'm ok with `showAccuracyCircle` as well.

I've been swayed by your arguments that default state should be on.

I'm agnostic on one or two markers. That's an implementation detail that could change in the future so I don't think it's a make or break decision.

Meekohi:
Hey all, procedural question --
#9202 (refactoring distance calculations) is now merged into master. Typically, I would take this PR and `git merge upstream/master`... but I know you also prefer to rebase onto master and I'm worried of making a big mess. Should I rebase this work onto master *now* instead and clean up any merge difficulties? Is that going to make the diff totally unusable (usually my experience when rebasing)?

ryanhamley:
@Meekohi we do squash and merge when the PR is merged to master which automatically rebases if possible. It's probably ok to hold off merging/rebasing manually unless you need to. If you do, `git rebase master` is usually fine.

Meekohi:
Okay rebased, merged, and updated to use the new distance refactor.

1. Updated docs as suggested
2. `default = true` -- it's a change I don't recommend lightly but it's very common in other mapping tools and a useful feature that I think most people will want on by default. As an aside, I've been playing with this more while I'm in Colombia this week (3G and not usually on wifi) and it's really nice to have the immediate feedback that you should wait a bit for the GPS signal to narrow down before you make a decision based on the map.
3. Personally I feel best about 2 markers vs trying to merge it into one. I think there may be some unexpected CSS consequences of adding a container element that holds both.

andrewharvey:
@Meekohi looks like something went wrong with the rebase, the diff now has a lot of other changes, which makes it hard to see what this PR is actually changing, could this be cleaned up? If you're not sure, maybe best to take a backup of your repo first just in case.

Meekohi:
🙄This is why I thought it's very strange this repo recommends rebasing instead of merging and asked about it. Github shows "three dot diffs" not "two dot diffs" -- I'll just start a new PR.

andrewharvey:
I'm interested if there's anything you think could be improved at https://github.com/mapbox/mapbox-gl-js/blob/master/CONTRIBUTING.md#version-control-conventions.

It should just be a matter of (on your-branch) `git rebase master` and then `git push --force origin your-branch.

Meekohi:
Nah that is fine and makes sense, I wouldn't propose to decide for everyone what is best. My 2¢ personal opinion is that when using Github, merging is the best workflow since all the "compare" Github features are built with that assumption in place. Hacks like `git push --force` are ok for small one-off contributors, but a real mess for collaboration (e.g. two people working on a feature together). When you finally accept the commit you can still "Squash and Merge" which gets you a nice commit history.

PR #9188, opened 2020-01-15
Upgrade Flow

kkaefer:
Apart from a few smaller changes and regressions in Flow, the biggest change is that Flow now checks `delete`s as well: I changed a few definitions for class members to be optional so that `delete` can set them `undefined`. This means we'll have stricter checking for non-nullish in many places.

### DIFF
--- a/src/gl/index_buffer.js
+++ b/src/gl/index_buffer.js
@@ -23,10 +23,6 @@ class IndexBuffer {
 
         context.bindElementBuffer.set(this.buffer);
         gl.bufferData(gl.ELEMENT_ARRAY_BUFFER, array.arrayBuffer, this.dynamicDraw ? gl.DYNAMIC_DRAW : gl.STATIC_DRAW);
-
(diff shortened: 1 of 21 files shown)


src/gl/index_buffer.js
```diff
@@ -23,10 +23,6 @@ class IndexBuffer {
 
         context.bindElementBuffer.set(this.buffer);
         gl.bufferData(gl.ELEMENT_ARRAY_BUFFER, array.arrayBuffer, this.dynamicDraw ? gl.DYNAMIC_DRAW : gl.STATIC_DRAW);
-
-        if (!this.dynamicDraw) {
-            delete array.arrayBuffer;
-        }
```
kkaefer: It doesn't really make sense to remove the arrayBuffer from the `StructArray` because there are still views of the ArrayBuffer that we didn't remove here, so it couldn't be GCed anyway.

kkaefer:
Let's not do this now.

Issue #8373, opened 2019-06-20
Expose evaluated properties in query results instead of serialized layer definitions


mourner:
Capturing from #8195:

> digging into 5 year commit history around this particular feature (query results including serialized layer), I discovered that the original intent was to return evaluated layout/paint values, not original definitions — see #879. At some point during many refactorings, this changed without anyone noticing.

> We could apply a temporary fix to mitigate performance issues, but an ideal solution would be to track down where that change happened, and see if we can avoid full serialization by reverting back to outputting final values of properties evaluated at the time of the query. I'll dig into this further.

PR #9198, opened 2020-01-16
Return evaluated properties in query results

ryanhamley:
## Launch Checklist

    - closes https://github.com/mapbox/mapbox-gl-js/issues/8373
    - serialize the layer once and cache it to avoid serializing on every feature or tile
    - evaluate properties and return their values to in the query
 - [x] post benchmark scores

### DIFF
--- a/src/data/feature_index.js
+++ b/src/data/feature_index.js
@@ -194,7 +194,7 @@ class FeatureIndex {
             }
 
             const styleLayer = styleLayers[layerID];
-            console.log('styleLayer', styleLayer);
+
             if (!styleLayer) continue;
 
             let featureState = {};
@@ -202,7 +202,23 @@ class FeatureIndex {
                 // `feature-state` expression evaluation requires feature state to be available
                 featureState = sourceFeatureState.getState(styleLayer.sourceLayer || '_geojsonTileLayer', id);
             }
-            console.log('featureState', featureState, feature);
+
+            const serializedLayer = serializedLayers[layerID];
+
+            const evaluatedPaint = {};
+            for (const property in serializedLayer.paint) {
+                const prop = styleLayer.paint.get(property);
+                evaluatedPaint[property] = prop && prop.evaluate ? prop.evaluate(feature, featureState) : prop;
+            }
+
+            const evaluatedLayout = {};
+            for (const property in serializedLayer.layout) {
+                const prop = styleLayer.layout.get(property);
+                evaluatedLayout[property] = prop && prop.evaluate ? prop.evaluate(feature, featureState) : prop;
+            }
+
+            serializedLayer.paint = evaluatedPaint;
+            serializedLayer.layout = evaluatedLayout;
 
             const intersectionZ = !intersectionTest || intersectionTest(feature, styleLayer, featureState);
             if (!intersectionZ) {
--- a/src/data/feature_index.js
+++ b/src/data/feature_index.js
@@ -290,7 +290,7 @@ register(
 export default FeatureIndex;
 
 function evaluateProperties(serializedProperties, styleLayerProperties, feature, featureState, availableImages) {
-    return mapObject(serializedProperties, (property, key, serializedProperties) => {
+    return mapObject(serializedProperties, (property, key) => {
         const prop = styleLayerProperties instanceof PossiblyEvaluated ? styleLayerProperties.get(key) : null;
         return prop && prop.evaluate ? prop.evaluate(feature, featureState, availableImages) : prop;
     });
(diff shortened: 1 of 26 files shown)


ryanhamley:
For some reason, when no geometry is supplied to the query, this update is consistently faster than `master`, but when the geometry is supplied, it's consistently slower. I'm investigating that.
Screen Shot 2020-01-27 at 2 58 07 PM

Screen Shot 2020-01-27 at 3 03 22 PM

chloekraw:
cc @pozdnyakov @mapbox/gl-native

ryanhamley:
After exploring an alternate approach in #9282 and testing out a branch that broke out layer serialization on its own (#9289), I discovered that the bulk of the performance hit in the `QueryPoint` benchmark is from serializing the layers in `queryRenderedFeatures`. If we serialize each layer as it's added and cache the results, we can get the benchmark almost to where `master` is as shown below. A test that removed the call to `listImages` in `queryRenderedFeatures` eliminated the remaining performance hit so the next step is to cache the list of availableImages.

Screen Shot 2020-02-11 at 6 42 19 PM

I ran the full benchmark suite against #9289 and didn't find any significant performance degradation from doing so while `QueryBox` showed a significant improvement.

ryanhamley:
Now that I've cached `serializedLayers` and `availableImages`, the benchmarks are all back inline with `master` and `queryBox` shows a marked improvement, which was the original goal of the ticket.

Screen Shot 2020-02-12 at 12 32 29 PM
Screen Shot 2020-02-12 at 12 38 45 PM
Screen Shot 2020-02-12 at 12 32 37 PM
Screen Shot 2020-02-12 at 12 32 48 PM
Screen Shot 2020-02-12 at 12 32 59 PM

src/data/feature_index.js
```diff
@@ -269,6 +295,16 @@ register(
...
 
+function evaluateProperties(serializedProperties, styleLayerProperties, evaluationParameters) {
+    const evaluatedProperties = {};
+    for (const property in serializedProperties) {
+        const prop = styleLayerProperties instanceof PossiblyEvaluated ? styleLayerProperties.get(property) : null;
+        evaluatedProperties[property] = prop && prop.evaluate ? prop.evaluate(evaluationParameters.feature, evaluationParameters.featureState, evaluationParameters.availableImages) : prop;
+    }
+
+    return evaluatedProperties;
+}
```
mourner: Nit: we have a `mapObject` function in `util`, so you could simplify this code to just the callback function for mapping.

src/data/feature_index.js
```diff
@@ -194,16 +198,34 @@ class FeatureIndex {
...
+                featureState = sourceFeatureState.getState(styleLayer.sourceLayer || '_geojsonTileLayer', id);
+            }
+
+            const serializedLayer = serializedLayers[layerID];
+
+            const evaluationParameters = {
+                feature,
+                featureState,
+                availableImages
+            };
```
mourner: Nit: probably no need to wrap those in an object if you immediately unwrap them in the lines below — separate arguments are fine.

src/style/style.js
```diff
@@ -408,11 +415,11 @@ class Style extends Evented {
             this.sourceCaches[sourceId].used = false;
         }
 
-        const availableImages = this.imageManager.listImages();
+        // const availableImages = this.imageManager.listImages();
```
mourner: Nit: remove leftover

src/style-spec/expression/types/resolved_image.js
```diff
@@ -18,8 +18,9 @@ export default class ResolvedImage {
         return this.name;
     }
 
-    static fromString(name: string): ResolvedImage {
-        return new ResolvedImage({name, available: false});
+    static fromString(name: string, availableImages?: ?Array<string>): ResolvedImage {
+        const available = availableImages ? availableImages.indexOf(name) > -1 : false;
+        return new ResolvedImage({name, available});
```
mourner: Is this change (including `coercion` expression) related to the PR, or an independent one?

Also noting `availableImages.indexOf` can have performance implications — how often is this method called and in what cases?

mourner:
Also just realized this would benefit from some query tests that cover the new functionality.

ryanhamley:
The benchmarks look good still. `QueryBox` shows a major performance improvement and everything else shows little to no change.

You can see a manual test I performed here where `"line-width": ["-", 13, 4]` is output as `"line-width": 9` 
Screen Shot 2020-02-27 at 7 13 38 PM

Screen Shot 2020-02-27 at 7 12 55 PM

@mourner do you have any thoughts on the best way to test this? our query tests aren't really set up to evaluate expressions and the unit tests for `style.js` use a mock `queryRenderedFeatures`.

mourner:
@ryanhamley awesome! Please edit the PR to add the changelog item + label.

PR #9202, opened 2020-01-17
refactor distance calculations

Meekohi:
* Refactors distance calculations into the `LngLat` class.
* Normalizes `ScaleControl` to use the same distance approximation as elsewhere (#8777)

This will also clean up #9181 and avoid copy/paste distance calculations everywhere. @mourner 

## Launch Checklist

 - [X] no visual changes
 - [X] <changelog>Updates `ScaleControl` to use same distance approximation as `MercatorCoordinate`


Meekohi:
Ahhhh sweet lord, this update has shifted some render-test results by a pixel here and there... need some advice on how to proceed.

I do feel strongly that the earlier `6378137` number is not sensible to use here -- if the Earth is an American football, this is like approximating it spherically by taking the "long-end" and making a sphere with that diameter (instead of taking the "average diameter").

Meekohi:
I visually verified that every failing test was a "one pixel shift" sort of thing that would be expected from this changed, and updated the `expected.png` for each. Diffs were all like:

diff

andrewharvey:
> I do feel strongly that the earlier 6378137 number is not sensible to use here -- if the Earth is an American football, this is like approximating it spherically by taking the "long-end" and making a sphere with that diameter (instead of taking the "average diameter").

I may be completely wrong about this, but my reasoning on why I thought 6378137 is correct at least for MercatorCoordinate, per my comments at #8777 is that while the oblate spheroid's (or oblate ellipsoid's) radius get's smaller as you get closer to the poles, the Web Mercator projection (EPSG:3857) is completely based on the equator radius not the earth mean radius, therefore we must also use the equator radius here. What am I missing?

Meekohi:
Interesting! I did a bunch more reading and apparently this is some confusion caused by our web mapping overlords:

"Google / Microsoft (and supposedly Yahoo) started using a spherical version of Mercator. For the spherical model of the Earth, they chose to use the semimajor axis of the WGS84/GRS80 ellipsoid, 6378137.0 meters, for the radius. This means that the sphere is much larger than the actual size of the Earth. There are many ways to calculate a sphere that's closer in size and/or shape to the Earth, but the radius value will be closer to 6371000 km."

So I guess the question is the value of running with that mistake vs correcting it... I was looking around at other geospatial libraries I use and found that most use 6371000, but curiously `turf.js` uses "6371008.8". It seems so specific I wonder if someone there has more insight into how it was chosen, and have asked at: https://github.com/Turfjs/turf/issues/1816

Also this amazing snippet: "As a point of trivia: 900913 was originally rejected as a credible projection by European Petroleum Survey Group (EPSG) so Google (who invented it for the web) gave it their own number, which if you type it on an old fashioned calculator spells 'googlE' (substitute '9' for 'g' and squint a bit). Eventually EPSG relented and allowed 900913 into its geodetic repository but gave it the number 3857."

Meekohi:
lol well given that @mourner convinced turf.js to use 6371008.8 I am in favor of doing the same here unless there are any objections. I wasn't quite clear on how you came up with 6371008.8 though... if the major axis is 6,378,137m and semi-major axis is 6,356,752.314245m, wouldn't the implied mean radius be 6367444.65712m? Where did 6371008.8m come from?

pathmapper:
> Where did 6371008.8m come from?

In https://github.com/Turfjs/turf/issues/635#issuecomment-292011500 is a Wikipedia article linked:

https://en.wikipedia.org/wiki/Earth_radius#Mean_radius

src/geo/lng_lat.js
```diff
@@ -73,6 +73,40 @@ class LngLat {
...
 
+    /*
+    * Approximate radius of the earth.
+    * Uses the WGS-84 approximation. The radius at the equator is ~6378137 and at the poles is ~6356752. https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84
+    * A widely accepted "average radius" for the entire earth is 6371000 (leans slightly more accurate toward the equator where most people live)
+    *
+    * @returns {number} Average radius of the earth in meters.
+    */
+    static earthRadius(): number {
+        return 6371000;
```
Meekohi:
```suggestion
        return 6371008.8; // see https://en.wikipedia.org/wiki/Earth_radius#Mean_radius, or ftp://athena.fsv.cvut.cz/ZFG/grs80-Moritz.pdf p. 4
```

Meekohi:
Okay I have updated the radius to `6371008.8m` and I believe this is ready for review again.

I did have to update `integration/render-tests/extent/1024-symbol/actual.png`, and it might be good if someone with more knowledge about what that is actually trying to test could verify the change looks good to them. I assume what's happening is that a few symbols are falling off the edge of render area, and then a few other symbols are readjusting their position based on more free space being available.

Meekohi:
diff

src/geo/lng_lat.js
```diff
@@ -73,6 +73,40 @@ class LngLat {
...
+    /*
+    * Approximate radius of the earth.
+    * Uses the WGS-84 approximation. The radius at the equator is ~6378137 and at the poles is ~6356752. https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84
+    * 6371008.8 is one published "average radius" see https://en.wikipedia.org/wiki/Earth_radius#Mean_radius, or ftp://athena.fsv.cvut.cz/ZFG/grs80-Moritz.pdf p.4
+    *
+    * @returns {number} Average radius of the earth in meters.
+    */
+    static earthRadius(): number {
+        return 6371008.8;
+    }
```
mourner: Nit: I wonder if we need to expose this publicly as a static method? Maybe just do `export const earthRadius = 6371008.8;` and import it (alongside the default export) in `MercatorCoordinate`?

mourner:
> I may be completely wrong about this, but my reasoning on why I thought 6378137 is correct at least for MercatorCoordinate, per my comments at #8777 is that while the oblate spheroid's (or oblate ellipsoid's) radius get's smaller as you get closer to the poles, the Web Mercator projection (EPSG:3857) is completely based on the equator radius not the earth mean radius, therefore we must also use the equator radius here. What am I missing?

@andrewharvey yeah, this topic seems confusing, but my take is that despite the EPSG definition which outputs coordinates in "mercator meters", the radius is actually not involved in calculations in any way — you're projecting WGS coords (-180..180, -90..90) to screen coords directly. The radius gets into play in GL JS because of pitched view & 3D, since extrusion heights are defined in meters which we need to convert, and in that case preferring more accurate distances feels more correct, although not that much of a difference. For `distanceTo` though, we definitely should use the mean radius because this calculation is independent of the projection and we want more precise measurements. And if we do, consistency with other calculations will be a plus.

src/geo/lng_lat.js
Meekohi: I like it. Done.

Meekohi:
@ansis would love to get your eyes on this if you get some time (hopefully is quick). This PR is blocking https://github.com/mapbox/mapbox-gl-js/pull/9181

ansis:
Thanks for working through this!

The decision to use `6371008.8` sounds fine to me.

I also tried digging into the  `1024-symbol` test failure. It looks like these changes make the example load only one tile instead of two (the boundary of the test aligns with the boundary of the tile). This should be fine.

Could you rebase this on master and run the render tests again?
- a recently changed test is failing and probably just needs to be updated `feature-state/promote-id-fill-extrusion`
- the new `zoomed-fill/negative-zoom` is failing on my local machine. Is it working for you?

Meekohi:
@ansis rebased onto master. `zoomed-fill/negative-zoom` passes on my machine locally and circleci seems happy.

Meekohi:
Totally unrelated to this PR, but I find it amazing that there are papers published as recently at 2015, proposing new formula for computing the exact distance between two points on an ellipsoid (e.g. Tseung et al., 2015: https://pdfs.semanticscholar.org/3dbb/88daa5567ae857fee6657ec7851ba4a82568.pdf) -- seemingly still a somewhat open problem... I wonder if there are any applications that would benefit from having a `turf.js`-esque library but with more very accurate distances?

Meekohi:
pst, who should I ping to actually final-review and merge this? Sorry if I'm prodding it too much, but I want to get #9181 in motion before I run out of steam and get busy again.

Meekohi:
Very welcome! Added that diff in, will need a final 👍@ansis.

ansis:
thanks!

PR #9208, opened 2020-01-20
Support geographically-sized text symbols

### LINKED ISSUE #7163: interpolate expression: strange numeric issues
**mapbox-gl-js version**:
v0.48.0
**browser**:
Chrome 68.0.3440.106 64bit Windows 7
Firefox 61.0.2 (64-Bit) Windows 7

### Steps to Trigger Behavior
I want a text field on a map which height is given in geographical units (mercator "kilometers").
So the text field should cover the same area on the map independent of the zoom level.
This works using the interpolate expression except some very strange numeric scaling errors occuring.

Consider the following minimal example:

```
            map.addSource('text', { type: 'geojson', data: { type:'FeatureCollection', features: [ {type: 'Feature', geometry: { type: 'Point', coordinates: [0, 0] },
                    properties: {
                      fS: 300 / 78.206,
                      fS10: 300 * Math.pow(2, 10) / 78.206,
                      fS11: 300 * Math.pow(2, 11) / 78.206,
                      fS20: 300 * Math.pow(2, 20) / 78.206
                    }}],}});

            // non inlined versions -> should be identical but are not
            map.addLayer({id: 'text10',type: 'symbol',source: 'text', layout: {"text-allow-overlap": true,"text-field": "test","text-size":
              ['interpolate', ["exponential", 2], ['zoom'],0,['get', 'fS'], 10, ['get', 'fS10'],]}
            });
            map.addLayer({id: 'text11',type: 'symbol',source: 'text', layout: {"text-allow-overlap": true,"text-field": "test","text-size":
              ['interpolate', ["exponential", 2], ['zoom'],0,['get', 'fS'], 11, ['get', 'fS11'],]}
            });
            map.addLayer({id: 'text20',type: 'symbol',source: 'text', layout: {"text-allow-overlap": true,"text-field": "test","text-size":
              ['interpolate', ["exponential", 2], ['zoom'],0,['get', 'fS'], 20, ['get', 'fS20'],]}
            });

            // inlined versions -> as expected they are identical
            map.addLayer({id: 'texti10',type: 'symbol',source: 'text', layout: {"text-allow-overlap": true,"text-field": "test","text-size":
              ['interpolate', ["exponential", 2], ['zoom'],0,300 / 78.206, 10, 300 * Math.pow(2, 10) / 78.206,]}
            });
            map.addLayer({id: 'texti11',type: 'symbol',source: 'text', layout: {"text-allow-overlap": true,"text-field": "test","text-size":
              ['interpolate', ["exponential", 2], ['zoom'],0,300 / 78.206, 11, 300 * Math.pow(2, 11) / 78.206,]}
            });
            map.addLayer({id: 'texti20',type: 'symbol',source: 'text', layout: {"text-allow-overlap": true,"text-field": "test","text-size":
              ['interpolate', ["exponential", 2], ['zoom'],0,300 / 78.206, 20, 300 * Math.pow(2, 20) / 78.206,]}
            });
```
So the height of the text field should be approx. 300km at the equator.
The first three layers use the get expression to retrieve the calculated values from the source.
The second three layers inline the expressions in the interpolate expression.

### Link to Demonstration
(sorry, I don't know how to add mapbox-gl to jsbin without using and sharing my access key)

### Expected Behavior
All six text fields should have the same size and lineup independent of the zoom level.

### Actual Behavior
Only the three inlined text fields line up.
The other three text fields using the get expression severe serious scaling issues.

It is very interesting that the inlined version work as expected. Maybe mapbox is evaluating the non inlined version using a shader at float precision and the inlined version using javascripts double precision? However, at zoom level 0 fS evaluates to ~4 and at zoom level 11 to ~7800. Even using float precision to calculate the interpolation, I would not expect such large errors when exponentially scaling between the two values. Maybe the calculation itself has some numeric issues?

Thank you very much in advance!

### LINKED ISSUE #7563: Interpolating text-size and exponentially scaling by zoom level with expressions
## Motivation
Our map renders layers that were created in CAD software. Every tile source contains its own custom set of labels with a "font-size" property. For example, a simple property looks like this:

`
{ "label": "Map Label", "font-size": 72 }
`

**What we need to do is allow users to increase the map's text size as high as 200% or decrease it as low as 50%.** The problem is that we cannot achieve the desired outcome with expressions.

To re-create the problem, let's say we already have an outside operation that calculates how the font size should scale exponentially from zoom 5 to zoom 25 that gets defined to the property's 'font-size-min' and 'font-size-max', respectively:

```
  map.setLayoutProperty('points','text-size',[
    'interpolate',
    ['exponential', 2],
    ['zoom'],
    5, ['get','font-size-min'], // let's assume this == 18
    25, ['get','font-size-max'] // let's assume this == 300 (exceeds the 255 limit)
  ]);
```

The code above should behave similar to just hard-coding these values:

```
  map.setLayoutProperty('points','text-size',[
    'interpolate',
    ['exponential', 2],
    ['zoom'],
    5, 18,
    25, 300
  ]);
```
[See a JSFiddle example](https://jsfiddle.net/8m7Lzu21/1/)

When using the JSFiddle example, clicking "Without expression" increases the font size as expected (even beyond the 255 maximum). However, when you click the "With expression" button, the font size reverts back to a default text size (24) and we get this error in the console:
`points: Value for "text-size" is >= 256. Reduce your "text-size".`

Our attempt to get around this is to use the "min" expression like so:
```
 map.setLayoutProperty('points','text-size',[
    'interpolate',
    ['exponential', 2],
    ['zoom'],
    5, ['min', ['get','font-size-min'], 255], 
    25, ['min', ['get','font-size-max'], 255]
  ]);
```

The problem with "min" is that we lose the desired rate that the text scales based on the zoom level. We're ok with enforcing a maximum value of 255 if that's what it takes, but we don't want to change the rate of scale that we see when "min" isn't used.

## Potential solution

I see one of three ways this could be resolved:

1. Don't enforce a text-size maximum at all (this might have performance consequences).
1. When the expression exceeds the maximum value, fallback to the maximum of 255 rather than the default value of 24.
2. Somehow let "min" not affect the overall rate of change at which the text-size scales.

Please let me know if you need me to clarify anything. Thanks!

mike-unearth:
Fixes #7163
Fixes #7563

This PR adds support for text layers that scale with the map by removing the restriction on text-size for `source` and `composite` expressions.

The main problem is that text-size has been given the same precision as icon-size, even though the former is in units of pixels and the latter is in units of relative size. Integer precision is fine for pixel units, but terrible for icon-size units. To support icon-size, these values are scaled by `SIZE_PACK_FACTOR` (previously 256 but recently reduced to 128), packed in to a UInt16, and then divided by `SIZE_PACK_FACTOR` in the shader. This change skips that upscaling and downscaling when we're dealing with text-size instead of icon-size, significantly increasing the upper range of text-size without reducing the precision of icon-size.

Here are some JSFiddles to demonstrate:

Using [v1.6.1](https://jsfiddle.net/82c5zn0L/15/): Note that no text is rendered and a console warning says `Value for "text-size" is >= 255. Reduce your "text-size".`

Using [this fork](https://jsfiddle.net/Ln38bcw1/): Note that you as you zoom, the text scales with the map.

 - [ ] post benchmark scores
 - [ ] tagged `@mapbox/studio` and/or `@mapbox/map-design-team` if this PR includes style spec or visual changes
 - [X] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog:
`<changelog>Adds support for geographically-sized text</changelog>`

### DIFF
--- a/src/shaders/symbol_sdf.vertex.glsl
+++ b/src/shaders/symbol_sdf.vertex.glsl
@@ -58,9 +58,9 @@ void main() {
     float size;
 
     if (!u_is_size_zoom_constant && !u_is_size_feature_constant) {
-        size = mix(a_size_min, a_size[1], u_size_t) / 128.0;
+        size = mix(a_size_min, a_size[1], u_size_t) / (u_is_text ? 1.0 : 128.0);
     } else if (u_is_size_zoom_constant && !u_is_size_feature_constant) {
-        size = a_size_min / 128.0;
+        size = a_size_min / (u_is_text ? 1.0 : 128.0);
     } else {
         size = u_size;
     }
--- a/src/shaders/symbol_text_and_icon.vertex.glsl
+++ b/src/shaders/symbol_text_and_icon.vertex.glsl
@@ -58,9 +58,9 @@ void main() {
     float size;
 
     if (!u_is_size_zoom_constant && !u_is_size_feature_constant) {
-        size = mix(a_size_min, a_size[1], u_size_t) / 128.0;
+        size = mix(a_size_min, a_size[1], u_size_t);
     } else if (u_is_size_zoom_constant && !u_is_size_feature_constant) {
-        size = a_size_min / 128.0;
+        size = a_size_min;
     } else {
         size = u_size;
     }
--- a/src/symbol/symbol_layout.js
+++ b/src/symbol/symbol_layout.js
@@ -496,19 +496,13 @@ function addTextVertices(bucket: SymbolBucket,
 
     if (sizeData.kind === 'source') {
         textSizeData = [
-            SIZE_PACK_FACTOR * layer.layout.get('text-size').evaluate(feature, {})
+            layer.layout.get('text-size').evaluate(feature, {})
         ];
-        if (textSizeData[0] > MAX_PACKED_SIZE) {
-            warnOnce(`${bucket.layerIds[0]}: Value for "text-size" is >= ${MAX_GLYPH_ICON_SIZE}. Reduce your "text-size".`);
-        }
     } else if (sizeData.kind === 'composite') {
         textSizeData = [
-            SIZE_PACK_FACTOR * sizes.compositeTextSizes[0].evaluate(feature, {}, canonical),
-            SIZE_PACK_FACTOR * sizes.compositeTextSizes[1].evaluate(feature, {}, canonical)
+            sizes.compositeTextSizes[0].evaluate(feature, {}, canonical),
+            sizes.compositeTextSizes[1].evaluate(feature, {}, canonical)
         ];
-        if (textSizeData[0] > MAX_PACKED_SIZE || textSizeData[1] > MAX_PACKED_SIZE) {
-            warnOnce(`${bucket.layerIds[0]}: Value for "text-size" is >= ${MAX_GLYPH_ICON_SIZE}. Reduce your "text-size".`);
-        }
     }
 
     bucket.addSymbols(
--- a/src/symbol/symbol_size.js
+++ b/src/symbol/symbol_size.js
@@ -77,9 +77,9 @@ function evaluateSizeForFeature(sizeData: SizeData,
                                 {uSize, uSizeT}: { uSize: number, uSizeT: number },
                                 {lowerSize, upperSize}: { lowerSize: number, upperSize: number}) {
     if (sizeData.kind === 'source') {
-        return lowerSize / SIZE_PACK_FACTOR;
+        return lowerSize;
     } else if (sizeData.kind === 'composite') {
-        return interpolate(lowerSize / SIZE_PACK_FACTOR, upperSize / SIZE_PACK_FACTOR, uSizeT);
+        return interpolate(lowerSize, upperSize, uSizeT);
     }
     return uSize;
 }


asheemmamoowala:
Hello @mike-unearth! Thank you for contributing to this project - and for the jsfiddle showing the new behavior!

In the demo, starting at zoom 17, you can see that text starts to be displayed less precisely. This, I believe, is hitting fundamental limitations of the SDF approach for rendering symbols in this project.

Past zoom 20, the text disappears entirely. I'm not certain it is related to this change, but it is not an issue on `master` so there's probably an investigation needed here. 

**Zoom 17**
image

**Zoom 18**
image


cc @ansis @alexshalamov @arindam1993

mike-unearth:
Hi @asheemmamoowala - thanks for the review! Sorry for the delayed response, I've been out of the country. To your points:
- SDF gets ugly at large sizes: completely agree that this is an issue with text in Mapbox, but I'm not sure that's related to the scope of this PR. I know that my customers are willing to accept this, but they are not willing to use our app without geo-scaled text. 
- Text disappears at high zoom: this is because the default setting for `text-allow-overlap` is `false`. Adding `'text-allow-overlap': true` to the `layout` style fixes it.

mike-unearth:
^ Don't mind the force-pushes above, I was working to get all tests passing. Fixed issues with text along a line and with non-text SDF icons.

asheemmamoowala:
@mike-unearth Thanks for bearing with the slow turn around on this. This feature would work better after the introduction of 2+ channel SDFs, as described in https://github.com/mapbox/mapbox-gl-native/issues/137. I've opened https://github.com/mapbox/mapbox-gl-js/issues/9857 in the mean time to track this feature request.

PR #9210, opened 2020-01-21
Fix promoteId for line layers

### LINKED ISSUE #9196: Feature state doesn't change for line layers that use the source with `promoteId` defined
**mapbox-gl-js version**: builded from 74b15736d414023345abab69b91995d949f7d11e

**browser**: Google Chrome 79.0.3945.117

### Steps to Trigger Behavior

 1. Build `mapbox-gl.js` and `mapbox-gl.css` from 74b15736d414023345abab69b91995d949f7d11e;
 2. Add source with vector tiles;
 3. Provide `promoteId` as described in https://github.com/mapbox/mapbox-gl-js/pull/8987;
 4. Add fill layer (for polygons) which has color based on `hover` feature state;
 5. Add line layer (for polygons borders) which has color based on `hover` feature state.

### Link to Demonstration

https://codepen.io/Mint_BY/pen/dyPqVjx (I can't put builded bundle in there, so the code needs to be downloaded and run in local env using builded bundle). Example from https://github.com/mapbox/mapbox-gl-js/pull/8987 with slight modifications.

### Expected Behavior

On county hover both fill layer and line layer should change its color. This is the behavior I noticed when applied the same changes to hover example from docs (https://docs.mapbox.com/mapbox-gl-js/example/hover-styles/) except GeoJSON without `promoteId` is used there.

### Actual Behavior

Only county has its color changed on hover. Border color remains the same.

mourner:
Closes #9196. Existing render test only tested `promoteId` circle layers — we should probably add a set of tests for each other layer type before we can merge.

## Launch Checklist


mourner:
@karimnaaji thanks for the missing tests! Rebased & pulled them here from your branch.

Issue #9192, opened 2020-01-15
SKU tokens are not added to Mapbox tiles added with the "tiles" array property


ryanhamley:
**mapbox-gl-js version**:
All

**browser**:
N/A

### Steps to Trigger Behavior

 1. Add a source to the map by providing a Mapbox URL to the `tiles` array property
```
map.addLayer(
    {
        id: 'example',
        type: 'raster',
        source: {
            type: 'raster',
            tiles: [
                'https://api.mapbox.com/v4/<mymap>/{z}/{x}/{y}.png?access_token=<mytoken>'
            ],
        }
    }
);
```

### Link to Demonstration

N/A

### Expected Behavior
All Mapbox resources should have SKU tokens attached to them.

### Actual Behavior
Mapbox resources requested through the `tiles` property do not have a SKU token

I believe the problem is this code. We only canonicalize resources if `options.url` is present. We should also determine if the URLs provided to the `tiles` property are Mapbox URLs and, if so, add SKU tokens to them.

https://github.com/mapbox/mapbox-gl-js/blob/90dbfec4591f744a005657fb1b6c06c33d49e167/src/source/load_tilejson.js#L30-L32

PR #9217, opened 2020-01-22
Canonicalize Mapbox tile URLs from inline TileJSON as well

kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/9192 by canonicalizing tile URLs that are set via runtime styling or inline. It preserves access tokens, and will fall back to the map's access token if no tile source specific access token is set.

## Launch Checklist

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Tile requests to Mapbox sources that don't use Mapbox URLs are now billed correctly</changelog>`


ryanhamley:
This looks good to me. You mentioned that there were a lot of edge cases; can you elaborate on that a bit?

Issue #5474, opened 2017-10-17
Generate collision circles in foreground


ChrisLoer:
This is a follow-up to PR #5150.

We currently generate "collision circle" geometry in the background using tile coordinates (in `collision_feature.js`). After #5150, most collision detection logic is run in the foreground using viewport coordinates. At render time, we figure out how much space a label will take up, and then decide _which_ of the pre-generated collision circles to use based on how far the edges of the label are from the label anchor. This step is awkward and relies on an approximation to convert back from "viewport unit distance" to "tile unit distance".

It would be cleaner to generate the collision circle geometry on the fly at render time, using viewport coordinates. Ideally we could do the circle generation at the same time we were calling `placeFirstAndLastGlyph`, although there are some challenges to deal with (the "label plane" used for glyph placement is not necessarily the same as the viewport plane, so some extra projection might be necessary).

/cc @ansis @anandthakker

mourner:
This was implemented in #9219

Issue #3534, opened 2016-11-03
Symbol features are not returned by `queryRenderedFeatures`


lucaswoj:
GL JS `master` isn't returning any symbol features from `Map#queryRenderedFeatures`. Why isn't this causing any `test-suite` failures? The bug bisects to https://github.com/mapbox/mapbox-gl-js/pull/3478. 

cc @jfirebaugh @mourner @mapsam 

**mapbox-gl-js version**: `master` as of 46a73c4efcf119d287810e0f7ec305eb9d4102a6 / https://github.com/mapbox/mapbox-gl-js/pull/3478

### Steps to Trigger Behavior

call `Map#queryRenderedFeatures` at a point with a symbol feature

(tricky to provide a jsbin because this bug occurs only in `master`)

### Expected Behavior

the symbol feature is returned

### Actual Behavior

the symbol feature is not returned

jfirebaugh:
I can take a look later today. I believe all of the query tests are for very simple styles, usually with only one layer. One thing I would check is if this only shows up if there are other layers present with the symbol layer.

lucaswoj:
## Debugging Notes

* the bug was introduced in https://github.com/mapbox/mapbox-gl-js/pull/3478
* the bug does not manifest itself on a map with only symbol layers & background layers
* the bug does manifest itself on a map with a symbol layer and a line or fill layer
* the symbol in the StructArray returned by CollisionTile#queryRendered has an incorrect bucketIndex

## 💡 

As of #3478, `SymbolBucket` no longer inherits from `Bucket`.

Most of the constructor logic from `Bucket` was copied to `SymbolBucket` but the initialization of the `index` property was not https://github.com/mapbox/mapbox-gl-js/pull/3478/files#diff-621f09c210ab28c3dbf7bc1f0f03c6b2R116

## Next Steps

 - [ ] initialize `index` within the `SymbolBucket` constructor
 - [ ] add a regression test to `mapbox-gl-test-suite`
 - [ ] 🚢

PR #9219, opened 2020-01-22
Collision circle generation at runtime

mpulkki-mapbox:
Collision circle generation logic is currently performed in two different coordinate spaces, tile and label space, and the final decision of which circles to use is made by stitching these results together. First, a certain number of collision circle candidates are pre-generated for each label in tile space. Secondly, circles that should be used in collision check are determined at runtime. Some of the previous work can be found here mapbox/mapbox-gl-js#5474.

Limitations and problems with this approach becomes apparent quite fast. There seems to be quite a lot of approximation logic in the code that produce correct looking results but start to break down with certain input. The biggest issues we're facing if we were to remove the 60 degree pitch limit:

1. The number of generated collision circles per label are 300% of what would normally be the max size. This might not be enough as some labels might extend past the padding.
image
2. Projection from label space to tile space is a crude approximation that produces inaccurate results. This might result in too few collision circles being used.
3. Perspective correction is not taken into account when choosing which collision circles to use.
image

These problems are difficult to solve without modifying the underlying algorithm. It would be possible to increase the number of generated collision circle candidates and improve the tile projection approximation, but it would just move the problem forward and not fix it. This pull requests implements 100% runtime collision circle generation that has already been brought up mapbox/mapbox-gl-js#5474. Some of the benefits of this approach are:

- Very simple to implement. Most of the required code is already written and can be reused.
- You get what you see. Same code is used for rendering and collision checks. No more crude approximations.
- No need for persistent storage for pregenerated circles.
- Perspective correction is properly applied for glyphs.

image

The code has been simplified to place collision circles by interpolating between road vertices on the label plane. Interpolation itself is cheap and the label plane information is a by-product that has been already computed during previous steps.

I'm expecting to see reduced number of collision circles especially with distant labels. The old implementation has to use somewhat expensive vertex projection for each collision circle candidate to see if it should be used in the collision check. This does not scale well with distant labels where you might need to go through 20 candidates to only use 4 of them. This is unnecessary as the screen space coverage is already known.

I need to run some benchmarks to see if there's any difference in performance. Debug rendering of collision circle placement is a bit more complex now with only screen space coordinates available, but the performance hit should be small.


mpulkki-mapbox:
(update to the previous post)
## Overview
This pull request transforms collision circle generation into a runtime operation. Motivation behind the change is to have a more precise and flexible collision detection for line labels that sets no expectations for camera orientation. The new placement logic shares code with rendering to better reflect what's actually visible on the screen.

This has become a relevant issue now that we're ready to support lod-based tile streaming. The current implementation is not a scalable solution to be used with less constrained camera as some corner cases like inaccuracies in collision detection are already present with the 60 degree pitch limitation. Improving line label placement is a prerequisite for increasing the maximum pitch angle and for any 3d-related features we may end up doing in the future.

#### Stockholm (new)
gh_stockholm_new

#### Stockholm (master)
gh_stockholm_master

## What's new?
The core idea behind this change is quite simple and well defined, but lot of supporting code had to be updated as well.
 1. **Collision circles are no longer pre-generated during the symbol layout process**
    Collision circles were previously generated during `performSymbolLayout` in a worker thread as part of the bucket creation process. A certain number of circles were created for each collision feature in tile coordinates and then serialized together with collision boxes. Shapes used for both collision detection and rendering were later deserialized from this array.

    The new role of `performSymbolLayout` involves evaluating collision circle diameter (previously text height) per feature and setting a flag to notify if collision circle generation should be done at runtime. Pixel based property `Text-padding` is also applied during runtime.

 2. **Collision circle generation at runtime**
    Collision circle candidates used in collision check were previously determined by finding positions of the first and last glyph on the label plane and then approximating their distances from the tile anchor. These distances were then compared to circle offsets in order to find the most optimal circle coverage for the label. I found this part of the algorithm particularly difficult to follow as it seemed to use very different numbers for transformations compared to the rendering logic. 

    The new approach uses screen space projection of the label path to find the most optimal number of collision circles. The projected path is already available as the very same information is required to determine which pregenerated collision circles to activate. The usage of this data should add no extra costs as collision checks are performed in screen coordinates which happens to be identical to the non-pitched label space (with a small translation added for viewport padding). An extra transformation step has to be done though for labels in pitched label plane in order to find their screen coordinates.

    _An example comparison. Left: master, right: new_
    gh_example_master gh_example_new_default

    _Added flexibility_
    gh_example_new_freq0 gh_example_new_freq1 gh_example_new_pad

 3. **Ownership of the debug data**
    With a fixed number of collision circles per bucket it was easy to create static sized vertex and index buffers for the data. Static vertex properties were stored in their own vertex buffer while more dynamic properties like "isUsed" and "hasCollision" were uploaded to gpu in their own vb after the placement process was finished.

    These buffers have now been removed and dynamic properties of debug circles (center_x, center_y, radius, flags) are stored in the `Placement`-class until the placement operation has been finished. Debug circles are pushed together with projection matrices to their corresponding buckets in `updateBucketOpacities`.

 4. **Debug rendering**
    Debug rendering is a bit more difficult task now that circles are stored directly in screen space coordinates. It wouldn't be a problem if positions were updated every frame before rendering, but the symbol placement is a continuous process that can span over multiple frames invalidating previously computed positions if camera moves. Screen space coordinates are always bound to a specific camera position and orientation. The solution is to re-project circles from the placement screen space to the new one used in rendering.

    To render an arbitrary number of circles I ended up implementing a simple webgl 1.0 compatible shader instancing that requires no extra extensions. Circles are sent to the gpu in batches of 64 instances encoded in a single vec4 uniform array and vertex data is just a collection of incremental index values. Memory footprint of this approach is small as only 16 bytes per circle is needed to be uploaded to the gpu.
    ```javascript
    // javascript
    const vertex_data = [0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5,...];  // 2 values per vertex
    ```
    ```glsl
    // glsl
    uniform vec4 u_quads[64];
    attribute vec2 a_idx;
    ...
    int quad_idx = floor(a_idx.x / 4);
    int vertex_idx = a_idx.x % 4;
    ```

    Alternative instancing/rendering implementations:
      - WebGL instancing
        - Not a WebGL 1.0 baseline feature
      - Single vertex buffer, circular update
        - Idea would be to append vertex data to the end of the buffer and start from the beginning when the buffer is full. Downside of this approach is that `bufferSubData` is not guaranteed to be an async operation. Some kind of "no_overwrite" flag and explicit buffer orphaning would be required as well. 
      - Encoding instance data into a texture
        - Vertex texture fetch support is not required in WebGL 1.0

 5. **Perspective correction of line labels is fixed**
    Perspective correction formula is (according to [Text Rendering](https://github.com/mapbox/mapbox-gl-native/wiki/Text-Rendering))
    ```javascript
    const perspectiveRatio = 0.5 + 0.5 * (camera_to_center / camera_to_anchor);
    const pitchScaledFontSize = pitchWithMap
      ? fontSize / perspectiveRatio
      : fontSize * perspectiveRatio;
    ```
    `updateLineLabels` in `projection.js` was using inverse of the perspective correction algorithm which produced squeezed labels near bottom and top of the screen: 
    ```javascript
    const perspectiveRatio = 0.5 + 0.5 * (camera_to_anchor / camera_to_center);
    const pitchScaledFontSize = pitchWithMap
      ? fontSize * perspectiveRatio
      : fontSize / perspectiveRatio;
    ```
    This fix effectively resolves the issue reported in ticket https://github.com/mapbox/mapbox-gl-native/issues/15167.
    
## What's new for users?
This PR might slightly change the set of visible labels on screen at given time as more precise collision circle placement allows more tightly packed labels. An example from the render tests:
#### Chicago (new)
gh_chicago_actual
#### Chicago (master)
gh_chicago_expected

## Benchmarks
I could not find any noticeable and reproducible difference in performance between this change and the master. I ran JS benchmarks and did some manual testing as well. I tracked the number of collision circles and boxes added to the collision grid and the number of collision checks done per placement operation. Tests were done using a fast i9 laptop and a slower Amazon Fire HD tablet.

**`debug.html`**

<details>
<summary>Petrolina (Brazil)</summary>

resolution: 1272 x 1323
**16.31/-9.42056/-40.499172/-159.5/2**
|                    | new | master | diff |
|--------------------|-----|--------|------|
| circles            | 668 | 822    | -19% |
| boxes              | 82  | 77     | +6%  |
| intersection tests | 137 | 214    | -35% |

**16.31/-9.416197/-40.499192/98.9/60**
|                    | new | master | diff |
|--------------------|-----|--------|------|
| circles            | 600 | 714    | -16% |
| boxes              | 129 | 126    | +2%  |
| intersection tests | 416 | 503    | -18% |

**16.32/-9.420562/-40.49918/-49.9/85**
|                    | new  | master | diff |
|--------------------|------|--------|------|
| circles            | 399  | 525    | -24% |
| boxes              | 205  | 211    | -3%  |
| intersection tests | 1935 | 2147   | -10% |
</details>
<details>
<summary>San Francisco</summary>

resolution: 2560 x 1336
**15.36/37.736134/-122.442253/-93.3**
|                    | new | master | diff |
|--------------------|-----|--------|------|
| circles            | 819 | 937    | -5%  |
| boxes              | 170 | 166    | +2%  |
| intersection tests | 267 | 308    | -14% |

**15.36/37.736676/-122.444918/-38.9/58**
|                    | new  | master | diff |
|--------------------|------|--------|------|
| circles            | 1919 | 2361   | -19% |
| boxes              | 423  | 417    | +1%  |
| intersection tests | 3065 | 3426   | -11% |

**14.13/37.76903/-122.42653/-130.1/79**
|                    | new  | master | diff |
|--------------------|------|--------|------|
| circles            | 690  | 748    | -8%  |
| boxes              | 351  | 347    | +1%  |
| intersection tests | 3173 | 3176   | -0%  |
</details>
<details>
<summary>Stockholm</summary>

resolution: 1272 x 1323
**16.15/59.321101/18.072541/0/73**
|                    | new  | master | diff |
|--------------------|------|--------|------|
| circles            | 760  | 918    | -18% |
| boxes              | 697  | 673    | +3%  |
| intersection tests | 4102 | 4462   | -8%  |

**16.31/59.32457/18.071447/45.6/2**
|                    | new  | master | diff |
|--------------------|------|--------|------|
| circles            | 836  | 947    | -12% |
| boxes              | 537  | 595    | -9%  |
| intersection tests | 4787 | 4774   | +0%  |
</details>

**`debug/circles.html`**

<details>
<summary>Stockholm</summary>

resolution: 1272 x 1323
**16.31/59.324659/18.071288/88.8**
|                    | new | master | diff |
|--------------------|-----|--------|------|
| circles            | 396 | 427    | -8%  |
| boxes              | 209 | 207    | +1%  |
| intersection tests | 463 | 500    | -8%  |

**16.31/59.314411/18.078096/164.8/54**
|                    | new | master | diff |
|--------------------|-----|--------|------|
| circles            | 445 | 579    | -24% |
| boxes              | 463 | 447    | +3%  |
| intersection tests | 940 | 1080   | -13% |
</details>

Depending on the scenario, the number of circles were reduced by 5-24% and the number of intersection tests by 0-35% with the current configuration. The number of collision boxes could actually grow.

## Notes and open questions
 - Most of the pages `debug/*.html` were tested.
 - `Text-anchor` property is ignored in collision circle placement #9313 . This PR uses more conservative approach by placing collision circles to cover both "flipped" and "non-flipped" cases.
 - `Text-padding` for line labels is now being applied at runtime during collision checks. Previously the padding was applied to circles in the preprocessing step.
 - Render and query test baselines should be updated. All of the failures are caused by a slightly different layout due to more precise placement. This is of course something that should be confirmed.
 - Most of the code comments should be updated to reflect the new logic.
 - draw_collision_debug.js recreates vertex and index buffers for each frame which is not optimal.
 - There's plenty of room for optimizing the interpolation logic.

mpulkki-mapbox:
cc @mapbox/gl-core, @mapbox/gl-js, @mapbox/gl-native

astojilj:
Logging offline review discussions status:
- 14 failing render tests would get fixed by rebasing expectations. New behavior looks better since labels are tighter packed and don't overlap.
- for 4 failing query tests, explaining the state before/after would help the review,
- good to describe difference in text-anchor behavior, previous vs new (more conservative approach by placing collision circles to cover both "flipped" and "non-flipped" cases).

mpulkki-mapbox:
Thanks for the comments, @astojilj! There are few intended changes that I should open up more.

> * good to describe difference in text-anchor behavior, previous vs new (more conservative approach by placing collision circles to cover both "flipped" and "non-flipped" cases).

The first one is different handling of flipped vs. non-flipped line labels in circle generation. I noticed some weird behavior in circle placement when text-anchor was used with line labels. Some of the labels were properly covered by collision shapes but about half of them were totally off. After further inspection this was caused by different handling of "flipped" and "non-flipped" labels in rendering and collision detection logic. Collision circle placement is treating all labels as "non-flipped", ie. it doesn't care if labels would appear upside down on the screen. This behavior combined with non-centered text anchor might result in huge misalignment between labels and collision shapes. I reported this issue first here https://github.com/mapbox/mapbox-gl-js/issues/9313.

I ended up implementing "conservative offsets" for glyph placement where both possibilities are taken into account. Enough circles will be generated to cover both placement cases (as seen in the following image) which should fix problems with horizontal anchor for the cost of a larger area reserved for labels. It might not be the most optimal approach and it's not required for this PR. Perhaps I should cut it out and implement in another PR?

image

Another intended change is the usage of "Text-padding". Instead of including pixel padding in precomputed collision shapes it's now being added at runtime directly in screen coordinates. It should still conform to the specification "Size of the additional area around the text bounding box used for detecting symbol collisions."

mpulkki-mapbox:
One additional benefit of this change is the reduced memory usage due to circles not being stored anymore. I positioned a map using `streets-v10` to few different locations and printed out the number of collision boxes and circles added to the `collisionBoxArray` during `performSymbolLayout` operation.

```
console.log(bucket.layers[0].id + ": " + (bucket.collisionBoxArray.length - preArrayLen));
```

San Francisco `22/49.28194692/-123.1107474/123.5/60`

|                                 | New  | Master |
|---------------------------------|------|--------|
| tunnel-oneway-arrows-blue-minor | 0    | 0      |
| tunnel-oneway-arrows-blue-major | 0    | 0      |
| tunnel-oneway-arrows-white      | 0    | 0      |
| turning-features-outline        | 0    | 0      |
| road-oneway-arrows-blue-minor   | 514  | 514    |
| road-oneway-arrows-blue-major   | 280  | 280    |
| level-crossings                 | 0    | 0      |
| road-oneway-arrows-white        | 0    | 0      |
| turning-features                | 0    | 0      |
| bridge-oneway-arrows-blue-minor | 0    | 0      |
| bridge-oneway-arrows-blue-major | 50   | 50     |
| bridge-oneway-arrows-white      | 0    | 0      |
| housenum-label                  | 132  | 132    |
| poi-scalerank4-l15              | 70   | 70     |
| poi-scalerank4-l1               | 68   | 68     |
| poi-parks_scalerank4            | 0    | 0      |
| poi-scalerank3                  | 18   | 18     |
| poi-parks-scalerank3            | 2    | 2      |
| poi-scalerank2                  | 0    | 0      |
| poi-parks-scalerank2            | 0    | 0      |
| poi-parks-scalerank1            | 0    | 0      |
| poi-scalerank1                  | 0    | 0      |
| road-label-small                | 0    | 12731  |
| road-label-medium               | 0    | 2786   |
| road-label-large                | 0    | 18058  |
| road-shields-black              | 46   | 46     |
| road-shields-white              | 0    | 0      |
| rail-label                      | 2    | 2      |
| Total                           | 1182 | 34757  |

The number of collision shapes created for all features is down to 5% in this specific overscaled tile. In raw memory usage this is `(34757 - 1182) * 24 = 805800 bytes = 806 KB` less bytes used. It should also be noted that the previous implementation duplicates collision shapes into a separate array in foreground so the memory reduction could be double of this number. Vertex buffers used to store debug circles are also all gone. 


Helsinki `#14.59/60.16331/24.93807`

|       | New | Master |
|-------|-----|--------|
| Total | 164 | 1274   |

The number of collision shapes in Helsinki on zoom level 14 leads to reduction of 83%. This is equal to `(1274 - 164) * 24 = 26.6 KB`.

mpulkki-mapbox:
> * good to describe difference in text-anchor behavior, previous vs new (more conservative approach by placing collision circles to cover both "flipped" and "non-flipped" cases).

I ended up reverting this fix for now.

src/render/draw_collision_debug.js
```diff
@@ -41,9 +50,165 @@ function drawCollisionDebugGeometry(painter: Painter, sourceCache: SourceCache,
...
+    // Vertex data is just a collection of incremental index values pointing to the quads-array.
+    //
+    // If one quad uses 4 vertices then all required values can be deduced from the index value:
+    //   int quad_idx = int(vertex.idx / 4);
+    //   int corner_idx = int(vertex.idx % 4);
+    //
+    // OpenGL ES 2.0 spec defines that the maximum number of supported vertex uniform vectors (vec4) should be
+    // at least 128. Choosing a safe value 64 for the quad array should leave enough space for rest of the
+    // uniform variables.
+    const maxQuadsPerDrawCall = 64;
```
karimnaaji: Should we inject a define in the circleProgram prelude for that value https://github.com/mapbox/mapbox-gl-js/blob/master/src/render/program.js#L40? 

That would prevent mismatch with the variable defined in src/shaders/collision_circle.vertex.glsl.

src/symbol/collision_index.js
```diff
@@ -84,56 +86,34 @@ class CollisionIndex {
...
-                          collisionGroupPredicate?: any): { circles: Array<number>, offscreen: boolean } {
+                          collisionGroupPredicate?: any,
+                          circlePixelDiameter: number,
+                          textPixelPadding: number): { circles: Array<number>, offscreen: boolean, collisionDetected: boolean } {
         const placedCollisionCircles = [];
 
-        const projectedAnchor = this.projectAnchor(posMatrix, symbol.anchorX, symbol.anchorY);
+        const perspectiveRatio = this.projectAnchor(posMatrix, symbol.anchorX, symbol.anchorY).perspectiveRatio;
+        const labelPlaneFontSize = pitchWithMap ? fontSize / perspectiveRatio : fontSize * perspectiveRatio;
+        const labelPlaneFontScale = labelPlaneFontSize / 24;
```
karimnaaji: Is 24 supposed to be representing ONE_EM defined in src/symbol/one_em.js?

mpulkki-mapbox: Yes it is, thanks for pointing it out :)

mpulkki-mapbox:
All failures in both render tests and query tests are caused by changes in the symbol layout. There are either more visible labels than before or some of the previously invisible labels are now hiding lower priority labels underneath them. Three changes introduced in this PR are causing the failures: 
 1. Perspective correction is now properly applied to labels near bottom and top of the screen removing the "squeezed" effect
 2. More precise collider placement allows more labels to be visible
 3. Collision circle debug rendering has changed

I double checked the two failing regression tests and updating them should not have a conflict with original issues (https://github.com/mapbox/mapbox-gl-js/issues/3534 and https://github.com/mapbox/mapbox-gl-js/issues/5911).

(I also found out that the perspective correction issue has been addressed before in https://github.com/mapbox/mapbox-gl-js/pull/5913, but `projection.js` was not updated at that point.)

src/render/draw_collision_debug.js
mpulkki-mapbox: That's a great idea and something we should definitely do. `ProgramConfiguration` doesn't seem to be flexible enough for this purpose though as it's built around binding style and tile layer properties into uniforms and vertex attributes. Do you have some other approach in mind?

src/symbol/collision_index.js
```diff
@@ -143,99 +124,119 @@ class CollisionIndex {
...
+            const radius = circlePixelDiameter * 0.5 * perspectiveRatio + textPixelPadding;
+            const screenPlaneMin = new Point(-viewportPadding, -viewportPadding);
+            const screenPlaneMax = new Point(this.screenRightBoundary, this.screenBottomBoundary);
+            const interpolator = new PathInterpolator();
+
+            // Construct a projected path from projected line vertices. Anchor points are ignored and removed
+            const first = firstAndLastGlyph.first;
+            const last = firstAndLastGlyph.last;
+
+            let projectedPath = first.path.slice(1).reverse().concat(last.path.slice(1));
```
alexshalamov: @mourner would Array.push be faster than concat?

src/symbol/collision_index.js
```diff
@@ -143,99 +124,119 @@ class CollisionIndex {
...
+                        }
                     }
                 }
             }
         }
 
         return {
-            circles: (collisionDetected || !inGrid) ? [] : placedCollisionCircles,
-            offscreen: entirelyOffscreen
+            circles: ((!showCollisionCircles && collisionDetected) || !inGrid) ? [] : placedCollisionCircles,
```
alexshalamov: is there a need to construct an empty `[]` array?

src/symbol/symbol_layout.js
```diff
@@ -713,6 +714,27 @@ function addSymbol(bucket: SymbolBucket,
...
+        if (feature && feature.circleDiameter)
+            return Math.max(feature.circleDiameter, prevHeight);
+        return prevHeight;
+    };
+
+    collisionCircleDiameter = getCollisionCircleHeight(textCollisionFeature, collisionCircleDiameter);
+    collisionCircleDiameter = getCollisionCircleHeight(verticalTextCollisionFeature, collisionCircleDiameter);
+    collisionCircleDiameter = getCollisionCircleHeight(iconCollisionFeature, collisionCircleDiameter);
+    collisionCircleDiameter = getCollisionCircleHeight(verticalIconCollisionFeature, collisionCircleDiameter);
+    const useRuntimeCollisionCircles = (collisionCircleDiameter > -1) ? 1 : 0;
```
alexshalamov: true / false?

src/symbol/collision_index.js
mourner: Yeah — if this is on a hot path, I would rewrite this with a simple `for` loop to avoid the 3 array allocations.

src/symbol/symbol_layout.js
mpulkki-mapbox: type conversion from bool to int is required because the variable is serialized as a number

src/symbol/collision_index.js
mpulkki-mapbox: nice catch! I'll try to get rid of the extra array allocations

src/render/draw_collision_debug.js
karimnaaji: Thanks for checking, if that requires some extra work it's fine to leave as is for the time being. We can revisit and extend `ProgramConfiguration` for that type of generic shader config injection when it shows problems.

src/symbol/collision_index.js
mpulkki-mapbox: I think it could be replaced with null

src/render/uniform_binding.js
```diff
@@ -133,13 +133,20 @@ class UniformMatrix4f extends Uniform<Float32Array> {
     }
 }
 
+class Uniform4fv extends Uniform<Float32Array> {
+    set(v: Float32Array): void {
+        this.gl.uniform4fv(this.location, v);
```
karimnaaji: Only relates to debug code so not a blocker, but have you evaluated the difference vs instancing with uniforms against a dynamic streaming vertex buffer thrown away and resized every frame? Stream buffer should only reallocate sparingly, I'm asking since uniform4v can be fairly slow and takes a high percentage of the frame time.

Screen Shot 2020-03-18 at 12 12 47 PM

We should add a warning note here to prevent any usage of that in release builds for something else debug code path.

src/symbol/path_interpolator.js
```diff
@@ -0,0 +1,61 @@
...
+
+        while (distOfCurrentIdx < distToTarget && currentIndex < this._distances.length) {
+            distOfCurrentIdx = this._distances[++currentIndex];
+        }
+
+        // Interpolate between the two points of the segment
+        const idxOfPrevPoint = currentIndex - 1;
+        const distOfPrevIdx = this._distances[idxOfPrevPoint];
+        const segmentLength = distOfCurrentIdx - distOfPrevIdx;
+        const segmentT = segmentLength > 0 ? (distToTarget - distOfPrevIdx) / segmentLength : 0;
```
karimnaaji: Any chance of having `segmentLength == 0`?

src/symbol/collision_index.js
```diff
@@ -143,99 +124,123 @@ class CollisionIndex {
...
+            // Tolerate a slightly longer distance than one diameter between two adjacent circles
+            const circleDist = radius * 2.5;
+
+            // The path might need to be converted into screen space if a pitched map is used as the label space
+            if (labelToScreenMatrix) {
+                const screenSpacePath = projectedPath.map(p => projection.project(p, labelToScreenMatrix));
+
+                // Do not try to place collision circles if even of the points is behind the camera.
+                // This is a plausible scenario with big camera pitch angles
+                if (screenSpacePath.some(point => point.signedDistanceFromCamera <= 0))
```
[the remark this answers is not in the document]
mpulkki-mapbox: Partly yes. https://github.com/mapbox/mapbox-gl-js/issues/9334 addresses horizontal labels that are projected to the screen in a shader with very little cpu intervention. Line label placement already handles cases where vertex of a projected path is behind the camera by truncating segments that are intersecting with the front plane.

https://github.com/mapbox/mapbox-gl-js/blob/d585ec5ce3f0794599421f27357947c91a335ee0/src/symbol/projection.js#L386-L399

The additional distance check had to be added for map aligned line labels as they might be transformed to behind the camera.

src/render/program.js
```diff
@@ -92,7 +92,11 @@ class Program<Us: UniformBindings> {
         for (let i = 0; i < numUniforms; i++) {
             const uniform = gl.getActiveUniform(this.program, i);
             if (uniform) {
-                uniformLocations[uniform.name] = gl.getUniformLocation(this.program, uniform.name);
+                // Strip ending '[0]' from array uniforms
+                let uniformName = uniform.name;
+                if (uniformName.endsWith("[0]"))
```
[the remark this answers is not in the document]
mpulkki-mapbox: Having "[0]" suffix is at least valid opengl es 2.0 and this was observed behavior on tested browsers. I'll revisit the debug rendering logic and double check if this is a safe assumption.

src/render/uniform_binding.js
mpulkki-mapbox: This is a good catch. I don't think I have comparable performance numbers for different implementations. I'll have a look!

src/symbol/path_interpolator.js
mpulkki-mapbox: I found these cases only in the render test `debug/collision-lines-pitched`. There are multiple zero length line segments in the test data. I didn't investigate the issue any further.

src/render/draw_collision_debug.js
```diff
@@ -8,28 +8,37 @@ import type SymbolBucket from '../data/bucket/symbol_bucket';
 import DepthMode from '../gl/depth_mode';
 import StencilMode from '../gl/stencil_mode';
 import CullFaceMode from '../gl/cull_face_mode';
-import {collisionUniformValues} from './program/collision_program';
+import {collisionUniformValues, collisionCircleUniformValues} from './program/collision_program';
+
+import {StructArrayLayout2i4, StructArrayLayout3ui6, CollisionCircleLayoutArray} from '../data/array_types';
```
ansis: Same as previous comment, I'd add names for these as well, even if they are just `QuadVerticesArray`

src/symbol/placement.js
```diff
@@ -647,6 +709,15 @@ export class Placement {
             }
         }
 
+        if (showCollisionBoxes && bucket.bucketInstanceId in this.collisionCircleArrays) {
+            const circleArray = this.collisionCircleArrays[bucket.bucketInstanceId];
+
+            // Store viewport and inverse projection matrices per bucket
+            mat4.invert(circleArray.invProjMatrix, posMatrix);
+            circleArray.viewportMatrix = this.collisionIndex.getViewportMatrix();
+            circleArray.bucket = bucket;
```
ansis: As you asked, holding onto this reference here might not be great since it's possible for the bucket to no longer be used when `commit()` is called. While I don't think it should error (even if the bucket is not used the reference should be valid enough) it might be cleaner to do it elsewhere.

Would it be possible to make the changes to the buffer in either within this method, or later within `updateLayerOpacities(...)` when we update the bucket's buffers that are actually used for rendering?

src/data/bucket/symbol_bucket.js
```diff
@@ -3,15 +3,14 @@
...
-    collisionCircleLayout,
     dynamicLayoutAttributes
 } from './symbol_attributes';
 
 import {SymbolLayoutArray,
     SymbolDynamicLayoutArray,
     SymbolOpacityArray,
     CollisionBoxLayoutArray,
-    CollisionCircleLayoutArray,
+    StructArrayLayout4f16,
```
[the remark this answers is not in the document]
mpulkki-mapbox: Thanks for pointing this out! I'll add a type definition for this array

src/render/uniform_binding.js
karimnaaji: Thanks for checking this! I think the new approach looks a bit simpler now as well.

src/symbol/collision_index.js
```diff
@@ -84,56 +87,34 @@ class CollisionIndex {
...
                           showCollisionCircles: boolean,
                           pitchWithMap: boolean,
-                          collisionGroupPredicate?: any): { circles: Array<number>, offscreen: boolean } {
+                          collisionGroupPredicate?: any,
+                          circlePixelDiameter: number,
+                          textPixelPadding: number): { circles: Array<number>, offscreen: boolean, collisionDetected: boolean } {
         const placedCollisionCircles = [];
 
-        const projectedAnchor = this.projectAnchor(posMatrix, symbol.anchorX, symbol.anchorY);
+        const perspectiveRatio = this.projectAnchor(posMatrix, symbol.anchorX, symbol.anchorY).perspectiveRatio;
```
ansis: Both `projectAnchor(...)` and `projection.project(...)` (a couple lines later) project the point. Maybe `projectAnchor` can be removed completely? It looks like the perspectiveRatio math was already previously duplicated here, in draw_symbol.js and in projection.js. Maybe all three of those can be deduplicated as well?

src/symbol/collision_index.js
```diff
@@ -143,99 +124,123 @@ class CollisionIndex {
...
-                            continue;
-                        }
-                    }
+            let segments = [];
+
+            if (projectedPath.length > 0) {
+                // Quickly check if the path is fully inside or outside of the padded collision region.
+                // For overlapping paths we'll only create collision circles for the visible segments
+                const minPoint = projectedPath[0].clone();
+                const maxPoint = projectedPath[1].clone();
```
ansis: Is it impossible for `projectedPath[1]` to be undefined?

src/symbol/collision_index.js
```diff
@@ -143,99 +124,123 @@ class CollisionIndex {
...
+                    maxPoint.y = Math.max(maxPoint.y, projectedPath[i].y);
+                }
+
+                const region = computeRectClipRegion(minPoint, maxPoint, screenPlaneMin, screenPlaneMax);
+
+                // 0 == fully visible, -1 == partly visible, > 1 not visble at all
+                if (region === 0) {
+                    segments = [projectedPath];
+                } else if (region < 0) {
+                    segments = clipLine([projectedPath], screenPlaneMin.x, screenPlaneMin.x, screenPlaneMax.x, screenPlaneMax.y);
```
ansis: This isn't very important but I'm wondering if it's worth clipping here. I haven't profiled but I'm wondering if it would be just as cheap to generate circles for the parts of the lines that are offscreen and do bounds-checks on circles individually.

src/symbol/path_interpolator.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    lerp(t: number): ?Point {
+        if (!this.points.length) {
+            return null;
+        } else if (this.points.length === 1) {
+            return this.points[0];
+        }
+
+        t = clamp(t, 0, 1);
```
ansis: nit: could this be an assertion?

src/symbol/collision_index.js
mpulkki-mapbox: In most scenarios this would be true. Problems arise with overscaled tiles where labels can be near the front plane. A point near the front plane (distance < 0.1) can be projected really far away from the screen (due to the division by `w`). I've encountered cases where more than 1000 circles were generated because one vertex of the line was near the front plane.

src/symbol/collision_index.js
mpulkki-mapbox: It should be impossible. `placeGlyphAlongLine` constructs a path from the anchor point to the position of first/last symbol. Projected path is constructed from anchor-to-first and anchor-to-last segments. I added an extra assertion.

src/symbol/collision_index.js
mpulkki-mapbox: That's a great idea! I've added a function to projection.js for computing perspective ratio. Two projections is still required in collision_index though.

src/symbol/placement.js
mpulkki-mapbox: I've removed the bucket reference and moved the update logic to `updateLayerOpacities()`

ansis: Looks good!

src/symbol/collision_index.js
ansis: Ahh, right. I didn't notice they were being projected to different planes.

src/symbol/collision_index.js
ansis: Yep, that makes sense! Thanks

mourner:
@mpulkki-mapbox I haven't looked at the code in detail, just a wild guess: while the PR intends to considerably simplify collision logic, it at the same time adds a few kilobytes to the bundle; could this indicate that maybe there's some potentially unused code that was left in place?

mpulkki-mapbox:
> @mpulkki-mapbox I haven't looked at the code in detail, just a wild guess: while the PR intends to considerably simplify collision logic, it at the same time adds a few kilobytes to the bundle; could this indicate that maybe there's some potentially unused code that was left in place?

@mourner thanks for bringing this up! I did a small refactoring pass and got rid of few extra bytes. I also double checked that this change _should_ not leave any (significant part of the) code unused. One culprit for the increased size is the debug rendering logic that was previously shared by both box and circle rendering. 

@ansis: thanks for the review! I've updated all of the failing query tests.

PR #9220, opened 2020-01-22
Fix line distances breaking gradient across tile boundaries

### LINKED ISSUE #8969: line-gradient colours are inconsistent when zooming into a geojson line
I noticed as you zoom into a single geojson line with `line-gradient` styling that the style seems to 'reset'. 
I first noticed this in my application, but it is also present on the examples page on the Mapbox site with all browsers I've tried:
image
The above gif is just an example, but this behaviour occurs all along the line. e.g zoom into the bridge.

**mapbox-gl-js version**: v1.5.0

**browsers**: 
Chrome 73.0.3683.86 
Firefox 69.03
Edge 41.16299.1419.0

### Steps to Trigger Behavior

 1. Zoom into line > level 14

### Link to Demonstration
https://docs.mapbox.com/mapbox-gl-js/example/line-gradient/
https://output.jsbin.com/kesogobuvu (as above, but with buttons to zoom to problem areas)

### Expected Behavior
Line-gradient colours should be consistent across all zoom levels for a single line.

### Actual Behavior
Colour gradient gets reset at multiple points along line as you zoom in.

karimnaaji:
Closes #8969, ensures scaled line distance is computed at least once when calculating the entire line distance, otherwise we hit a corner case where the distance isn't updated. Specifically when the beginning of the line doesn't have a sharp corner.

## Launch Checklist

 - [x] tagged @mapbox/gl-native might need to be ported if #8303 was already ported to native
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix line distances breaking gradient across tile boundaries</changelog>`


src/data/bucket/line_bucket.js
```diff
@@ -537,6 +541,10 @@ class LineBucket implements Bucket {
             (this.clipStart + (this.clipEnd - this.clipStart) * this.distance / this.totalDistance)  * (MAX_LINE_DISTANCE - 1) :
             this.distance;
     }
+
+    updateDistance(prev: Point, next: Point) {
+        this.distance += prev.dist(next);
+    }
```
mourner: Nit: since we want to call `updateScaledDistance` after every `updateDistance`, we might as well call it inside `updateDistance` to DRY up the code a little.

Issue #8461, opened 2019-07-11
Can't access sprite file by relative URL


stevage:
**mapbox-gl-js version**: 1.1.0

I'm wanting to use a self-hosted sprite file, available at `/sprite@2x.json` and `/sprite@2x.png` on the domain where I'm hosting the site.

Putting this in my style file doesn't work:

```
            sprite: '/sprite'
```

> `Error: Unable to parse URL object`

These also fail: `sprite`, `http:///sprite`, `./sprite`.

Is there another way to achieve this, or are relative URLs just not supported for sprite files?

This is problematic because it makes deploying more work: I need code that detects whether running locally (`localhost:8080//...`) or on the deployed site, and now I also need to know the deployed site's URL.

stevage:
Well, I guess this workaround works:

```
sprite: `${window.location}sprite`
```

mourner:
We don't support relative URLs intentionally — see the discussion at #3636 and linked tickets.

stevage:
Thanks for the context. Interesting rationale. Makes sense that a style file needs to work in contexts other than web, I guess.

PR #9225, opened 2020-01-23
Allows local urls for sprites

webdeb:
## Launch Checklist

Local urls for sprites like "sprite: /assets/sprites" were not possible due to the implicit requirement of (protocol://). Slightly changed the "urlRe" to ignore missing protocol and the same for formatUrl. Additional tests were added, no breaking changes.

### DIFF
--- a/src/util/mapbox.js
+++ b/src/util/mapbox.js
@@ -197,7 +197,7 @@ function hasCacheDefeatingSku(url: string) {
     return url.indexOf('sku=') > 0 && isMapboxHTTPURL(url);
 }
 
-const urlRe = /^(\w+):\/\/([^/?]*)(\/[^?]+)?\??(.+)?/;
+const urlRe = /^(\w+)?:?\/?\/([^/?]*)(\/[^?]+)?\??(.+)?/;
 
 function parseUrl(url: string): UrlObject {
     const parts = url.match(urlRe);
@@ -214,7 +214,7 @@ function parseUrl(url: string): UrlObject {
 
 function formatUrl(obj: UrlObject): string {
     const params = obj.params.length ? `?${obj.params.join('&')}` : '';
-    return `${obj.protocol}://${obj.authority}${obj.path}${params}`;
+    return `${obj.protocol ? `${obj.protocol}://` : '/'}${obj.authority}${obj.path}${params}`;
 }
 
 export {isMapboxURL, isMapboxHTTPURL, hasCacheDefeatingSku};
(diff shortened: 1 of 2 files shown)


andrewharvey:
Related tickets #8461 and #3636

webdeb:
Thanks for the links @andrewharvey 

Seems to be an ongoing discussion since long time ago, so long, I don't remember the actual date. :) 
I don't mind if this will get rejected, my current approach was to use

```
`${window.location.origin}/assets/map-sprite`
``` 

But, this is somewhat ugly imo, especially with ssr, you have to built on top of that workaround, mock the window obj. 

Or this beauty here https://github.com/mapbox/mapbox-gl-js/issues/3636#issuecomment-282036893
It got 2 x 👍 (I am not sure, whats likeable about it, its frustrating...)

IMHO its natural to have custom sprites served from the current server, so please allow us to make our lives flexible, even if its dangerous and will break in other contexts (we know what we do, most of the time).

kkaefer:
> IMHO its natural to have custom sprites served from the current server

Styles are intended to be consumed by multiple different platforms, e.g. iOS applications, Android apps, and websites. Out of those, websites are the only one that have the concept of "current server" since they have an origin. Allowing relative URLs to styles means that styles won't work on those platforms anymore.

However, there is a much cleaner way to allow relative URLs for your website by providing a `transformRequest` callback function when initializing the Map object. E.g. adding this callback:

```js
transformRequest: (url, resourceType) => {
    if (/^local:\/\//.test(url)) {
        return { url: new URL(url.substr('local://'.length), location.href).href };
    }
}
```

allows you to specify relative sprite URLs like `"sprite": "local://./path/to/sprite"` in your style. Please note that the `./` is necessary to denote that it's a relative path. you can leave out the `.` to specify a domain-relative path. Note that `local://` is an arbitrary protocol; you can use whatever string you want there as long as it's alphanumeric (matches the regex `(\w+):\/\/`).

See the [documentation for `options.transformRequest`](https://docs.mapbox.com/mapbox-gl-js/api/#map) for more information.

Given that there's no reasonable cross-platform way of adding support for relative URLs, but a JS-specific solution, I'm going to close this PR. Thanks for your contribution!

webdeb:
@kkaefer thanks for your explanation.

But your suggestion, to use `local://./path` in combination with `transformRequest` (even that it looks better and more elegant, than other suggestions), still adds complexity and performance impact on **all requests** to solve this problem, where my PR aimed to eliminate it and avoid further discussions on that topic. Furthermore, by applying it, I do not see, any impact for people and apps, that run mapbox in their native server-free env.

Issue #9224, opened 2020-01-23
load additional Sprite


webdeb:
**mapbox-gl-js version**:  1.6.1

### Question
how can I load an additional sprite?
I saw, that I could load an additional image with the `addImage` method, however this method loads images only one-by-one. I would like to load a complete sprite how it is defined by the api.

As I understood from the code, I would have to reimplement the logic in https://github.com/mapbox/mapbox-gl-js/blob/b61d19b8ed84b37f729bba44503b8d69ab209c13/src/style/style.js#L253

Just wanted to ask, if you mind to expose it as a method of `style` so we could add multiple sprites after the initial loading?

```js
style.loadSprite(url)
```

Thank you

kkaefer:
See https://github.com/mapbox/mapbox-gl-js/pull/9226#issuecomment-578098987

PR #9226, opened 2020-01-23
Expose loadSprite as style.method

webdeb:
## Launch Checklist

This basically exposes loadSprite as a separate method of the Style class. I think it is also kind of refactoring, since the code is very specific and it makes sense to put the logic out of  _load anyway (even a private method). 
As a public method it allows the user to load its own sprites after the initial loading of the basic style, instead of loading each image one-by-one (#9224) 

Please give me some advises where to put the documentation, and tests, I did't wanted to keep me back with it yet, first need to know what you think about it in general. Thank you.

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -250,20 +250,7 @@ class Style extends Evented {
         }
 
         if (json.sprite) {
-            this._spriteRequest = loadSprite(json.sprite, this.map._requestManager, (err, images) => {
-                this._spriteRequest = null;
-                if (err) {
-                    this.fire(new ErrorEvent(err));
-                } else if (images) {
-                    for (const id in images) {
-                        this.imageManager.addImage(id, images[id]);
-                    }
-                }
-
-                this.imageManager.setLoaded(true);
-                this.dispatcher.broadcast('setImages', this.imageManager.listImages());
-                this.fire(new Event('data', {dataType: 'style'}));
-            });
+            this.loadSprite(json.sprite);
         } else {
             this.imageManager.setLoaded(true);
         }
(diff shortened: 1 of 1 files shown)


kkaefer:
Thanks for your contribution. This issue is a little more complex than just exposing this function, e.g. it doesn't cover cases where the same URL is added multiple times, or when requesting multiple sprites simultaneously. Additionally, it's unclear what happens when different sprites provide an image with the same name. There are also various notifications that don't get fired for subsequent sprite loads (`ImageManager#setLoaded`), which could leave partially parsed tiles in limbo.

We're planning to expand the way icons are supplied to support a bigger variety of use cases rather than just one sprite. However, due to the complexity of multiple sprite sheets or other image sources, we'll go through an RFC process to ensure that we've covered as many use cases as well as edge cases as possible before embarking on an implementation.

Given this explanation, I'm going to close this PR for now since we can't merge it at the moment, even with modifications. However, we're very interested in learning more about your use cases for having multiple sprites because it will help us design an updated sprite/image/icon system that covers your use case.

webdeb:
@kkaefer ok, that are good points that you first will go through the RFC process, I like that.
Are you ok, to change it to a private `_loadSprite` method right now, as already stated, it also makes the `_load` method a bit cleaner?

kkaefer:
That's a change that we'd probably accept. I assume your intention is to call this private method. Please refer to my explanation above for why that's not a good idea and will probably break in various edge cases.

webdeb:
> That's a change that we'd probably accept.
awesome 🎉 

> I assume your intention is to call this private method.

Yes, for sure. 😃

> Please refer to my explanation above for why that's not a good idea and will probably break in various edge cases

- [x] I understand the risks

webdeb:
@kkaefer can you reopen? I forced-pushed the changes into this branch, but I guess changes are not tracked by closed PRs

kkaefer:
Unfortunately, I can't reopen this PR since Github won't let me with the message "The wb-loadSprite-as-style-method branch was force-pushed or recreated.".

webdeb:
np https://github.com/mapbox/mapbox-gl-js/pull/9231

PR #9229, opened 2020-01-24
Hide glyphs behind the camera

mpulkki-mapbox:
Our symbol placement and rendering logic is currently ignoring depth values of projected anchor points and symbol vertices. This is fine as most of the projected vertices end up being in front of the camera. But due to the nature of perspective projection where x and y coordinates are divided by depth (w-component), vertices behind the camera have negative depth values and will have inverse coordinates on the screen. For example an anchor in bottom-left part of the clip-space might have a coordinate [-x, -y]. With negative depth value this will end up being [x, y] in homogenous NDC-space after the division by w. Typically GPU will clip these primitives after they've gone through a vertex shader.

This becomes a problem with > 60 pitch angles where screen boundary checks might not catch symbols directly behind the camera.

symbols_floating_above_ground

This pull request adds few extra depth checks for both line aligned labels and labels that are handled entirely in the shader.


mpulkki-mapbox:
> Can we add a render test that would capture the change?

Yes definitely! I'll have a look what kind of task it is to enable > 60 pitch angles for render tests

mpulkki-mapbox:
I noticed that it's not possible to override the hard-coded pitch limit from a render test. Should I postpone this PR until we have officially removed the 60 degree pitch limit? What's your opinion @mourner?

Issue #9240, opened 2020-01-28
"in" expression with array doesn't work


bravecow:
**mapbox-gl-js version**: 1.7.0

Doesn't work:

```
'circle-color': [
    'case',
   ['in', 1, [1, 2]], '#000',
   /* other */ '#ccc'
]
```

Error:
`evented.js:140 Error: layers.population.paint.circle-color[1][2][0]: Expression name must be a string, but found number instead. If you wanted a literal array, use ["literal", [...]]`

jsfiddle: https://jsfiddle.net/c31d2jgo/1/

But works with strings:

```
'circle-color': [
   'case',
   ['in', 1, '123'], '#000',
   /* other */ '#ccc'
]
```

Sorry if it is not a bug and I am missing something 🤷‍♂

bravecow:
I considered that `in` is a new replacement for `['match', value, array, true, false]`.

ryanhamley:
Ah, this is a confusing aspect of expressions generally, but not a bug. You can see in the error that it's referencing `[1][2][0]` which points us to the array literal that's supplied as the third argument to the `in` expression. When an expression is being parsed, any time an array literal is encountered, the parsing engine looks at the first element of the array (in this case, the number `1`) and tries to match it against the list of expressions so an array literal of just numbers or objects or whatever will fail because the parsing engine expects it to be another expression. The way around this is to use the `literal` operator to tell the parsing engine that the next array it encounters should not be parsed as an expression.

So `['in', 1, ['literal', [1, 2]]]` should work as expected. You don't normally have to do this when accessing properties (e.g. `['get', 'myArrayProp']`) because the parsing engine already knows that the returned value is not going to be another expression.

bravecow:
@ryanhamley Oh, please, consider to imporove description of this expression in docs:

https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#in

It is clearly said that haystack is array or string (nothing about `literal`). And it works without `literal` with string.

Thanks for clarification!

PR #9242, opened 2020-01-28
Clarify in documentation

ryanhamley:
## Launch Checklist

     - The `in` expression can be confusing because the docs say that the second argument can be an array. However, supplying an array literal throws an error unless the argument is wrapped in the `literal` expression. This PR updates the docs to explicitly call out this nuance.
     - Thanks to @bravecow for bringing this up in https://github.com/mapbox/mapbox-gl-js/issues/9240#issuecomment-579469818

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2589,7 +2589,7 @@
         }
       },
       "in": {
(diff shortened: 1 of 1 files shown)


asheemmamoowala:
Closed in favor of https://github.com/mapbox/mapbox-gl-js-docs/pull/212

Issue #4593, opened 2017-04-17
ReferenceError: self is not defined


jhjhjh96:
Mapbox gl does not work in my node js.

-error code-
C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:402
"use strict";module.exports=self;

ReferenceError: self is not defined
    at Object.__dirname.196 (C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:402:29)
    at s (C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:1:684)
    at C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:1:735
    at Object.__dirname.194../window (C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:398:25)
    at s (C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:1:684)
    at C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:1:735
    at Object.__dirname.65.../package.json (C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:139:26)
    at s (C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:1:684)
    at e (C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:1:855)
    at C:\Users\YJH\Documents\Visual Studio 2017\Projects\mapbox\mapbox\node_modules\mapbox-gl\dist\mapbox-gl.js:1:873

I don't know, what was wrong.
I read a mapbox document. Initial setting also installed such as 'webpack', 'browserify' ect...
Help me solve this problem.

[...]

PR #9243, opened 2020-01-29
Fix "ReferenceError: self is not defined" (#4593)

NameFILIP:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/4593

## Launch Checklist

 - [ ] post benchmark scores
 - [ ] tagged `@mapbox/studio` and/or `@mapbox/map-design-team` if this PR includes style spec or visual changes

### DIFF
--- a/src/util/browser/window.js
+++ b/src/util/browser/window.js
@@ -2,4 +2,6 @@
 /* eslint-env browser */
 import type {Window} from '../../types/window';
 
-export default (self: Window);
+const win: ?Window = typeof self !== 'undefined' ? self : undefined;
+
+export default win;


kkaefer:
The file [`src/util/browser/window.js`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/browser/window.js) is specifically intended for being used in a browser. There's a non-browser version in [`src/util/window.js`](https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/window.js), which is used by default. Instead of altering the browser-only version, modify your setup to avoid directly including the browser-specific version. Some bundlers read the [`browsers` key in our `package.json`](https://github.com/mapbox/mapbox-gl-js/blob/77cfc5c135fa23a432e065e286326cfe87b2400d/package.json#L110-L114) which specifies browser-specific file replacements. If you're building a bundle for use in Node.js, you should ignore these files.

NameFILIP:
In my case, I'm building a bundle to be used on the frontend, but the build process is happening in Node.

@kkaefer what is the downside of writing resilient code like this? It'll still work in the browser and in addition, will not crash the node environment (while being imported, not used). Look how many people are complaining about it: https://github.com/mapbox/mapbox-gl-js/issues/4593

Issue #9241, opened 2020-01-28
Move from Tap to Tape for unit tests


mourner:
Out current unit tests are powered by [Tap](https://github.com/tapjs/node-tap). This is problematic for multiple reasons:

- Tap is Node-only, which makes it hard to implement running the suite across browsers instead of just Node.
- We're currently stuck on Tap v12.4 because later versions break our setup and I haven't been able to track down the exact root cause. There were [~70 releases](https://github.com/tapjs/node-tap/releases) since then, with the latest being v14.10.6.
- The project philosophy generally makes it very hard to debug issues like the one above — Tap increasingly tries to cram every possible feature into it, currently having `typescript`, `esm`, `remove-flow-types` and `nyc` in its direct dependencies (while we use the last three separately and this likely leads to weird conflicts).
- It's a bloated project with a 2700-line package lock (~100MB in `node_modules`).

We should switch to [Tape](https://github.com/substack/tape). It's already used in our project for browser query tests, and extending its use across the project will make our test setup simpler, lighter and easier to manage. Tape advantages:

- Works in all browsers.
- Minimalistic. No bloat and unnecessary dependencies, easy to debug.
- More popular — 3x the GH stars and 7x "used by" repos.
- Much better tested. Compare [this CI matrix](https://travis-ci.org/substack/tape/builds/639026921) to the [Tap one](https://travis-ci.org/tapjs/node-tap/builds/636675053).

The biggest obstacle to the switch is that a lot of our unit tests depend on `beforeEach` / `afterEach` hooks, which are not supported by Tape. The latter is a [design choice](https://github.com/substack/tape/issues/59), which prefers explicit setup/teardown inline in each test for simplicity and clarity. I like this approach more, but it may take substantial effort to refactor our suite to it.

cc @arindam1993 @kkaefer

arindam1993:
Another blocker I found while attempting this port (for the browser part of using tape) was all the node API's we use for `fs` and `path` within our tests to load fixtures.

mourner:
We're using Vitest now

PR #9245, opened 2020-01-30
Add browser testing with Selenium/WebDriver

kkaefer:
This adds a new Circle CI build called `test-browser` that runs Selenium/WebDriver based tests directly on Circle CI. We're already running on CircleCI's Docker image that includes browsers, along with the required `geckodriver` and `chromedriver`. Therefore, we don't have to maintain additional Docker images for Selenium testing.

This PR adds a few example tests for things we could test in a browser and is supposed to serve as a frame of reference for future Selenium tests required by other PRs, e.g. https://github.com/mapbox/mapbox-gl-js/pull/9057.


test/util/test.js
```diff
@@ -25,6 +26,15 @@ type CreateTest = (typeof sinon) & {
...
 
+// Add function for floating point equality checking.
+tap.Test.prototype.almostEqual = function(found, wanted, epsilon, message) {
+    return this.ok(
+        almostEqual(found, wanted, epsilon),
+        message || "should be almost equal",
+        {found, wanted, compare: "~="}
+    );
+};
+
```
mourner: See also https://github.com/mapbox/mapbox-gl-js/blob/master/test/util/index.js#L33 which has the same purpose and is used throughout various unit tests. Needs to be de-duplicated. I prefer the explicit version rather than a prototype augment to make #9241 easier in the future.

test/browser/fixtures/index.html
```diff
@@ -0,0 +1,58 @@
...
...439,57.01672],[-158.46112,57.21692],[-157.72278,57.57013],[-157.55029,58.32831],[-157.04169,58.91888],[-158.1947,58.61591],[-158.51721,58.7879],[-159.05859,58.42432],[-159.71161,58.93152],[-159.9812,58.57269],[-160.35529,59.07111],[-161.35498,58.6709],[-161.96881,58.67169],[-162.05499,59.26691],[-161.87408,59.63373],[-162.51801,59.98969],[-163.8183,59.7981],[-164.66217,60.26752],[-165.34637,60.50751],[-165.35077,61.07391],[-166.1214,61.50012],[-165.73437,62.07507],[-164.91919,62.63312],[-164.5625,63.14648],[-163.7533,63.21948],[-163.0672,63.05951],[-162.2605,63.54187],[-161.53442,63.45587],[-160.77252,63.76611],[-160.95831,64.2229],[-161.51801,64.40289],[-160.77771,64.7887],[-161.39191,64.77728],[-162.453,64.55951],[-162.75781,64.33868],[-163.54639,64.55908],[-164.96082,64.44708],[-166.42529,64.68671],[-166.84497,65.08893],[-168.11047,65.6701],[-166.7052,66.08832],[-164.47467,66.57672],[-163.65253,66.57672],[-163.78851,66.07727],[-161.6778,66.11609],[-162.48969,66.73547],[-163.71973,67.11652],[-164.43091,67.61627],[-165.3902,68.04291],[-166.7644,68.35889],[-166.20471,68.88312],[-164.43079,68.91553],[-163.16858,69.37109],[-162.93048,69.85809],[-161.90887,70.33331],[-160.93481,70.447
```
mourner: Can we move the map setup here (or at least the map options object) from the fixture into the test itself so that it's easier to add new tests with different setups and everything relevant is in one place?

test/util/test.js
kkaefer: Ah, I didn't find this method.

test/browser/fixtures/index.html
kkaefer: Moved the JSON to a separate file and the initialization to `browser.js` with the option of overriding individual keys.

PR #9253, opened 2020-02-01
add circle to GeolocateControl showing accuracy of position

Meekohi:
Continuation of #9181 to get a clean rebase.

## Motivation
The default GeolocateControl shows the viewer's position as reported by `navigator.geolocation.watchPosition` immediately, but this often has a very high margin of error. The dot is often in the wrong location, which can be confusing or even dangerous depending on the application.

## Implementation
Adds a transparent circle to GeolocateControl showing the accuracy of the position, as reported by `navigator.geolocation.watchPosition`.

Screen Shot 2020-01-14 at 10 30 00 AM

image

## Launch Checklist

### DIFF
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -267,7 +267,7 @@ class GeolocateControl extends Evented {
             this._accuracyCircleMarker.setLngLat(center).addTo(this._map);
             this._userLocationDotMarker.setLngLat(center).addTo(this._map);
             this._accuracy = position.coords.accuracy;
-            if (this.options.trackUserLocation && this.options.showAccuracyCircle) {
+            if (this.options.showUserLocation && this.options.showAccuracyCircle) {
                 this._updateCircleRadius();
             }
         } else {
@@ -288,7 +288,7 @@ class GeolocateControl extends Evented {
     }
 
     _onZoom() {
-        if (this.options.trackUserLocation && this.options.showAccuracyCircle) {
+        if (this.options.showUserLocation && this.options.showAccuracyCircle) {
             this._updateCircleRadius();
         }
     }
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -80,7 +80,7 @@ let noTimeout = false;
  * @param {Object} [options.positionOptions={enableHighAccuracy: false, timeout: 6000}] A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
  * @param {Object} [options.fitBoundsOptions={maxZoom: 15}] A [`fitBounds`](#map#fitbounds) options object to use when the map is panned and zoomed to the user's location. The default is to use a `maxZoom` of 15 to limit how far the map will zoom in for very accurate locations.
  * @param {Object} [options.trackUserLocation=false] If `true` the Geolocate Control becomes a toggle button and when active the map will receive updates to the user's location as it changes.
- * @param {Object} [options.showAccuracyCircle=true] By default a transparent circle will be drawn around the user location indicating the accuracy (95% confidence level) of the user's location. Set to `false` to disable.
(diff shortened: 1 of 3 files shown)


src/ui/control/geolocate_control.js
```diff
@@ -251,9 +263,31 @@ class GeolocateControl extends Evented {
 
     _updateMarker(position: ?Position) {
         if (position) {
-            this._userLocationDotMarker.setLngLat([position.coords.longitude, position.coords.latitude]).addTo(this._map);
+            const center = new LngLat(position.coords.longitude, position.coords.latitude);
+            this._accuracyCircleMarker.setLngLat(center).addTo(this._map);
+            this._userLocationDotMarker.setLngLat(center).addTo(this._map);
+            this._accuracy = position.coords.accuracy;
+            this._onZoom();
```
andrewharvey: My personal preference would be to wrap all the checks you have in `_onZoom` inside `_updateCircleRadius`, and then just call `_updateCircleRadius` here and also `on('zoom')`.

Since you're calling _onZoom for something that isn't due to a zoom event, I think changing this would make the code slightly more readable.

andrewharvey: Though I guess that's an opinionated change too, since now you're calling `_updateCircleRadius` even though you might not be updating it (based on the check). I'm not sure what's best, but I just think it's best not to have onZoom triggered by something that isn't a zoom change.

Meekohi: Good point, agreed 👍

src/ui/control/geolocate_control.js
```diff
@@ -251,9 +263,33 @@ class GeolocateControl extends Evented {
...
+        const a = this._map.unproject([0, y]);
+        const b = this._map.unproject([1, y]);
+        const metersPerPixel = a.distanceTo(b);
+        const circleDiameter = Math.ceil(2.0 * this._accuracy / metersPerPixel);
+        this._circleElement.style.width = `${circleDiameter}px`;
+        this._circleElement.style.height = `${circleDiameter}px`;
+    }
+
+    _onZoom() {
+        if (this.options.trackUserLocation && this.options.showAccuracyCircle) {
```
andrewharvey: We don't need to check for `trackUserLocation` because you can still have an accuracy circle and user location dot shown with `trackUserLocation=false` and in that case we still need to update the radius on zoom.

src/ui/control/geolocate_control.js
```diff
@@ -251,9 +263,33 @@ class GeolocateControl extends Evented {
 
     _updateMarker(position: ?Position) {
         if (position) {
-            this._userLocationDotMarker.setLngLat([position.coords.longitude, position.coords.latitude]).addTo(this._map);
+            const center = new LngLat(position.coords.longitude, position.coords.latitude);
+            this._accuracyCircleMarker.setLngLat(center).addTo(this._map);
+            this._userLocationDotMarker.setLngLat(center).addTo(this._map);
+            this._accuracy = position.coords.accuracy;
+            if (this.options.trackUserLocation && this.options.showAccuracyCircle) {
```
andrewharvey: We don't need to check for `trackUserLocation` because you can still have an accuracy circle and user location dot shown with `trackUserLocation=false` and in that case we still need to update the radius after we have received the geolocation position and set `this._accuracy`.

andrewharvey: In fact it wouldn't hurt to try and add a unit test to catch this if possible.

Meekohi:
Not nitpicking at all! Updated and added a related test.

src/ui/control/geolocate_control.js
```diff
@@ -267,7 +267,7 @@ class GeolocateControl extends Evented {
             this._accuracyCircleMarker.setLngLat(center).addTo(this._map);
             this._userLocationDotMarker.setLngLat(center).addTo(this._map);
             this._accuracy = position.coords.accuracy;
-            if (this.options.trackUserLocation && this.options.showAccuracyCircle) {
+            if (this.options.showUserLocation && this.options.showAccuracyCircle) {
```
andrewharvey: One final point now is just to update the API doc to note that `showAccuracyCircle` only has effect when `showUserLocation` is set to true.

Meekohi: 👍 5c3713335508f31b3273731ab29a32cbc5354fdc

Issue #8752, opened 2019-09-11
Mapbox line-pattern distorts the image


alexcza:
I am applying **line-pattern** property to a Line but the result is distorted.

The image width is a factor of two (32 or 64px) and I am using a fixed width for the line (I am not trying to scale it with the zoom - which is much worse btw)

paint: {
         'line-pattern':fillPatternName,
         'line-width': 32,
       },

**Example:**
Here is the pattern: https://imgur.com/a/uJqkkTj
And this is the actual result: https://imgur.com/a/4sj1jgv

As you can see it will create an angle in the pattern and also the repeat is inconsistent as it will start repeating the pattern without finishing the previous pattern draw (overlaping).

asheemmamoowala:
@alexcza Can you please provide a minimal reproducing sample to help us diagnose the underlying issue.

asheemmamoowala:
Closing this ticket for lack on information. 

The issue can be reopened if you can provide a minimal, complete, verifiable demonstration of the issue. We recommend using https://jsbin.com. Please include information about the browser version and operating system used. This helps the maintainers and community understand and address the problem accurately.

jrdb:
Hello @asheemmamoowala,  could you please reopen?  I'm encountering this issue as well.  

https://jsbin.com/zegefucuri
OSX, Chrome Version 79.0.3945.88 (Official Build) (64-bit)

astojilj:
This is resolved for geoJSON source, PTAL at https://docs.mapbox.com/mapbox-gl-js/example/line-pattern and distortion-free dotted pattern line.

Issue #6296, opened 2018-03-07
Pattern scale reset


natslaughter:
**Current Issue**
In order for fill patterns to scale with tile sizes between zoom level integers, patterns are scaled up to 200%. If a map designer wants to use a fill pattern, they must be okay with a large range of pattern scaling for their map (_a_). If the map comes to rest in-between integer zoom levels, the patterns can be way off the intended map design (_b_). 

**Proposed Solution**
Reset the pattern scale to 100% when the map is at rest. “Rest” could be defined by 200ms of no zooming, but this should be tested. If a user is zooming in and out with a high frequency of stops, a constant resetting of pattern scale could be visually distracting. Currently, place-labels and other symbol layers are being re-rendered when the map is ‘at rest,’ so perhaps resetting the pattern scale would go unnoticed. I could see this enhancement being beneficial for line patterns as well. 

_a: scaling range of fill-patterns for one zoom level_
![proposal1-1](https://user-images.githubusercontent.com/16616543/37106884-2b2105ce-2233-11e8-86ab-1228f860e2ab.png)

_b: visual difference between scale polars_
![proposal1-2](https://user-images.githubusercontent.com/16616543/37106886-2e8a0d82-2233-11e8-9713-3c048a3bf556.png)

cc @ansis @kkaefer @eschow

nickidlugash:
Other suboptimal effects of the current implementation:
- Any scaling over 100% results in reduced resolution, which can be quite apparent at 199% scale on high-resolution screens. Switching over to vector icons should eventually fix this issue.
- While scaling up/down across an integer zoom level is continuous, the reset between zoom levels is abrupt between 199% and 100%, though there is a hard-coded cross-fade value (300ms?) to ease the transition.

Hopefully the proposed solution will ease these two issues a bit too, since more resets should mean fewer instances of scaling close to 199%.

> Reset the pattern scale to 100% when the map is at rest.

To clarify, would we want to do this in addition or instead of reseting at integer zoom levels? My instinct is that it should be in addition, since there may be situations (like during animations) where a map will be zooming across multiple zoom levels with no rest.

> I could see this enhancement being beneficial for line patterns as well.

And line dasharrays!

nickidlugash:
@ansis Any thoughts on how big of a lift it would be to test out this implementation?

tristen:
[...]

PR #9266, opened 2020-02-05
Fix line-pattern artifacts under overscaling and underscaling

karimnaaji:
This fixes overscaling/underscaling of line pattern as well as render differences between different physical pixel ratio sprites when used as line patterns. Refer https://github.com/mapbox/mapbox-gl-native/issues/15472

### Visual differences

The behavior is changed to match the following:

1. When the line pattern fits on a line smaller than its own resolution, we downscale while keeping the aspect ratio at integer zoom level
2. When the line pattern fits on a line larger than its own resolution, we upscale while keeping the aspect ratio at integer zoom level
3. When the line pattern has different pixel ratio attributes (1x, 2x, 3x...), the pattern screenspace resolution should remain stable (e.g. the physical resolution shouldn't impact the size at which the pattern is rendered but instead increase the quality of the visual while keeping its logical resolution)

The current behavior does not account for 1. and 3.

### Before

underscale-2x@2x
underscale-2x@2x-before
underscale-3x@2x
underscale-3x@2x-before
overscale-1x@2x
overscale-1x@2x-before
overscale-2x@2x
overscale-2x@2x-before

### After

underscale-2x@2x
underscale-2x@2x-after
underscale-3x@2x
underscale-3x@2x-after
overscale-1x@2x
overscale-1x@2x-after
overscale-2x@2x
overscale-2x@2x-after

### Shader differences

This will vary per-platform/shader compiler but gives an idea of the difference.
Try to move for more `mad` instructions instead of `div`. Per-instruction count differences in shader `line_pattern.fragment.glsl` :

 before: 6 `mad`, 12 `div`, 12 `add`, 9 `mul`
 after: 13 `mad`, 8 `div`, 6 `add`, 9 `mul`

This increases the number of `mad` instructions (multiply-add) for fewer cycles and decreases the number of divisions and additions.

## Launch Checklist

 - [x] tagged @mapbox/studio and @mapbox/map-design-team if this PR includes style spec or visual changes
 - [x] tagged @mapbox/gl-native there are shader changes that will require porting to native
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Improve scaling of patterns used in line-pattern on all device resolutions and pixel ratios</changelog>`


ansis:
Line width can change with zoom. Now that the rendered width of the pattern is scaled by the line width, the width of the pattern can change with zoom. This creates the marching ants effect (see the `show-marching-ants` branch and zoom in slowly).

Dashed lines had a similar problem and this was fixed but it's pretty hacky. We added an internal pseudo-paint-property called floorwidth:
https://github.com/mapbox/mapbox-gl-js/blob/2a0ae8eacfc42451bde4f02f53dc9f7860f405f6/src/shaders/line_sdf.vertex.glsl#L46
https://github.com/mapbox/mapbox-gl-js/blob/2a0ae8eacfc42451bde4f02f53dc9f7860f405f6/src/style/style_layer/line_style_layer.js#L75-L76

I think you should be able to copy the floorwidth pragmas from line_sdf.vertex.glsl and use that value instead. I think it might not play nicely with the crossfading but that's a mess anyways.


> 1. When the line pattern fits on a line smaller than its own resolution, we downscale while keeping the aspect ratio at integer zoom level

The new behavior seems good here but is this a breaking change? I'm not sure if the existing behavior is intentional but someone could be depending on it. I can kind of understand how some patterns may be clipped and still look intentional. I think we could probably get away with changing this and considering it a fix instead of a breaking change.

karimnaaji:
@ansis good catch! That makes sense to use floorwidth value instead of the dynamic width, and it does indeed fix the marching ants issue.

>The new behavior seems good here but is this a breaking change? I'm not sure if the existing behavior is intentional but someone could be depending on it. I can kind of understand how some patterns may be clipped and still look intentional. I think we could probably get away with changing this and considering it a fix instead of a breaking change.

Thanks for bringing that up because it's something that could be concerning.

Someone could depend on that in some particular situation, the most accidental use would be if it gave this result and the user worked around/with it. Do you know whether the stretching and/or clipping were considered when this was introduced?

ansis:
> the most accidental use would be if it gave this result and the user worked around/with it.

Yep, most likely.

> Do you know whether the stretching and/or clipping were considered when this was introduced?

I think I may have introduced this a long time ago but it probably wasn't carefully considered and it may have been before we had people actually using this. It was also never documented. I think I was also designing this for things that were more "textures" instead of "patterns", for example a rough paper surface or pencil shading where clipping is ok and scaling could be problematic. But I'm not sure I've really seen people using it like this in practice. And I have seen a lot of people surprised by the behavior.

I think we should make this change. I'm not sure whether it should be considered breaking. I think it's right on the fence. If we do consider it breaking we should probably wait until we have other breaking changes we need to make.

chloekraw:
@karimnaaji great work on this and thanks for the thorough PR description. I learned quite a bit just from reading it!

Considering whether this change is breaking is a good idea, and we should call it out at the top of the changelog as a softly breaking change with a link to the pr description. Maybe we’ll get feedback in the beta that makes us think twice about merging. 

In general, I think it’s pretty unlikely that a designer who has created a pattern wants that pattern to visually look different at any point. Designers tend to want things to look pixel perfect exactly in the way that they’ve designed it, positive and negative space included. Else, they would intentionally design the variant as a separate pattern use that at the zoom levels they would like that pattern to appear. I think it’s more likely that some customers don’t mind the current behavior as opposed to intentionally wanting it in their design. 

The feedback in these tickets also make me think that this PR would be a welcome change.

• https://github.com/mapbox/mapbox-gl-js/issues/8752
• https://github.com/mapbox/mapbox-gl-js/issues/6296

karimnaaji:
@ansis I've tested this on native and haven't noticed any issues on my manual testing https://github.com/mapbox/mapbox-gl-native/commit/0572a167a039d7526830643ddc0327d202f4608d. Let me know if there would be any other concerns you'd like me to address and otherwise we could call this a softly breaking change to prevent any surprise on the user's end.

karimnaaji:
To recap some of the discussions we have two options:
- Keep this change as is and communicate it as a bug fix with visual breaking change, requiring changes  to assets in the following case:
  - The asset is intended to be displayed as a _surface pattern_, such as in [this stylesheet](https://api.mapbox.com/styles/v1/saman/ciql4uao1000xbkm7knq5qf52.html?fresh=true&title=view&access_token=pk.eyJ1Ijoic2FtYW4iLCJhIjoiS1ptdnd0VSJ9.19qza-F_vXkgpnh80oZJww#16.58/44.984669/-93.263967/-9.5/45) and _any_ line has a line width smaller than the pattern it tries to fit to, resulting in automatic visual differences: [before](https://jsfiddle.net/kanaa/es3ndo4b/13/) vs [after](https://jsfiddle.net/kanaa/1m290bL5/2/)
   - Advantage: we fix [this type](https://github.com/mapbox/mapbox-gl-native/issues/15472) or [this type](https://api.mapbox.com/styles/v1/karimnaaji/ck6tjy68718pm1imt42skn9k9.html?fresh=true&title=view&access_token=pk.eyJ1Ijoia2FyaW1uYWFqaSIsImEiOiJjazVkNGZsbHUxc2g4M2ptbXp6aXphZDc4In0.ctuAToHAEOrKxqKzBDANIg#13.63/35.67814/139.76383) of clipping issue, deal better with different pixel ratio
   - Disavantage: we might cut off some valid use cases, introducing a requirement to revisit some assets in the use case previously described
- Introduce a boolean toggle in the stylespec for `line-pattern`, defaulting to the current behavior, otherwise to this new behavior.
  - Advantage: we support more use cases and we are backward compatible, not introducing potential regressions and carry all the advantages of the breaking change option
  - Disavantage: we introduce new stylespec information

karimnaaji:
@ansis, do you have any further thoughts? I have gathered some concerns from the map teams but the visual impact would be on a very few number of internal styles and, to their knowledge, none of external styles; where in such case we would communicate how to get the initial behavior back. But I'd like to discuss our options to move this forward, especially the second one as it would be the least disruptive for the user.

chloekraw:
>   - The asset is intended to be displayed as a _surface pattern_, such as in [this stylesheet](https://api.mapbox.com/styles/v1/saman/ciql4uao1000xbkm7knq5qf52.html?fresh=true&title=view&access_token=pk.eyJ1Ijoic2FtYW4iLCJhIjoiS1ptdnd0VSJ9.19qza-F_vXkgpnh80oZJww#16.58/44.984669/-93.263967/-9.5/45) and _any_ line has a line width smaller than the pattern it tries to fit to, resulting in automatic visual differences: [before](https://jsfiddle.net/kanaa/es3ndo4b/13/) vs [after](https://jsfiddle.net/kanaa/1m290bL5/2/)

I'm not sure if it's just me, but I can barely see the difference between the before and after jsfiddles when I zoom. 🤔 _Is_ scaling problematic if you add a rough paper surface or pencil shading texture to a line? If we have a visual example of that, that would be compelling to me. 

> Introduce a boolean toggle in the stylespec for line-pattern, defaulting to the current behavior, otherwise to this new behavior.

I have a slight preference for not introducing an option but I can get behind it. So the change would be making `line-pattern` accept an optional second argument where you specify whether the pattern is clipped, and by default it's `true`? 

Would it be better to make the developer specify `clip` or `scale`, so that we leave open the possibility of supporting some other option in the future?

karimnaaji:
>Is scaling problematic if you add a rough paper surface or pencil shading texture to a line? If we have a visual example of that, that would be compelling to me.

When using a texture for use as a surface, it _scales to fit_, it might be desired or undesired depending on the use case, here is a visual difference of that:

clip

>So the change would be making line-pattern accept an optional second argument where you specify whether the pattern is clipped, and by default it's true?
>Would it be better to make the developer specify clip or scale, so that we leave open the possibility of supporting some other option in the future?

Yes to both of these questions, I think we generally are missing control over how we would like textures to fit. Having a few texture mapping parameters to control how this works in both 2d and 3d is probably something we should consider, e.g:
2d:
pattern-mapping:

chloekraw:
@karimnaaji I have a suggestion for modifying this changelog entry from

> `<changelog>Fix line-pattern artifacts under overscaling and underscaling</changelog>`

to 

> `<changelog>Improve scaling of patterns used in `line-pattern` across resolutions and pixel ratios</changelog>`

karimnaaji:
@chloekraw updated!

PR #9267, opened 2020-02-06
Remove the simplex font and switch to using canvas for debug text

arindam1993:
This change removes the inline `simplex` font used for rendering the tile id when using `showTileBoundaries` this reduces the bundle size a bunch and also leads to much sharper text.

It also gives us the flexibility of adding more complex debug text with ease, since we can just draw whatever and wherever to the `tileOverlay` Canvas element.

Before:
Screen Shot 2020-02-11 at 3 44 11 PM
After:
Screen Shot 2020-02-11 at 3 43 34 PM

## Launch Checklist

 - [x] post benchmark scores
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Reduce bundle size and improve visual quality of showTileBoundaries debug text</changelog>`

cc @mapbox/studio @mapbox/gl-native


alexshalamov:
@arindam1993 would it be possible to create new shader? In gl-native, for some platforms we can try drawing debug text overlays using system APIs, for others, we can fallback to simplex font and use old shader.

arindam1993:
> @arindam1993 would it be possible to create new shader? In gl-native, for some platforms we can try drawing debug text overlays using system APIs, for others, we can fallback to simplex font and use old shader.

Could do that, but is that mainly for maintaining compatibility with native? That would also be possible by passing in a 1px alpha texture, which what I do in this PR as well

alexshalamov:
> Could do that, but is that mainly for maintaining compatibility with native?

As the shaders are shared, once we bump gl-js we would be forced to use new debug shader and it would be quite difficult to implement cross-platform 'canvas' in native.

It would be great if we would have two debug shaders, gl-js would use new shader that expects text to be drawn by canvas, meanwhile, gl-native would use old shader that uses simplex font.

src/render/draw_debug.js
```diff
@@ -36,6 +31,10 @@ function drawDebugTile(painter, sourceCache, coord: OverscaledTileID) {
     const colorMode = painter.colorModeForRenderPass();
     const id = '$debug';
 
+    context.activeTexture.set(gl.TEXTURE0);
+    // Bind the empty texture for drawing outlines
+    painter.emptyTexture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
+
     program.draw(context, gl.LINE_STRIP, depthMode, stencilMode, colorMode, CullFaceMode.disabled,
```
arindam1993: Unchanged call for drawing outlines I was talking about.

arindam1993:
> > Could do that, but is that mainly for maintaining compatibility with native?
> 
> As the shaders are shared, once we bump gl-js we would be forced to use new debug shader and it would be quite difficult to implement cross-platform 'canvas' in native.
> 
> It would be great if we would have two debug shaders, gl-js would use new shader that expects text to be drawn by canvas, meanwhile, gl-native would use old shader that uses simplex font.

Yes, but you wouldn't need to implement a cross platform canvas. The new shader is backwards-compatible and will behave the same as long as you attach a 1x1 transparent img to `u_overlay` and 1 to `u_overlay_scale`. This PR uses this behaviour to render the red outline lines with minimal changes.

alexshalamov:
> This PR uses this behaviour to render the red outline lines with minimal changes.

Yes, thanks. I see now https://github.com/mapbox/mapbox-gl-js/pull/9267/files#diff-8393648de39005607c39c441a8b144caR8

debug/2762.html
```diff
@@ -5,6 +5,7 @@
     <meta charset='utf-8'>
     <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
     <link rel='stylesheet' href='../dist/mapbox-gl.css' />
+    <link href="https://fonts.googleapis.com/css?family=Open+Sans:700&display=swap" rel="stylesheet">
```
ansis: Do we need to load these fonts for these pages? Would it be fine to just end up with whatever the system has already?

arindam1993:
@alexshalamov how should we track the change required on native? I'm happy to get my feet wet and work on the native change.

olso:
How is the bundle size reduced? https://bundlephobia.com/result?p=mapbox-gl@1.9.0-beta.1

It's getting larger! @arindam1993

PR #9279, opened 2020-02-10
Check if this._circleElement exists before trying to change it's style

pakastin:
Fixes an issue when this._circleElement doesn't exist before trying to update it.

### DIFF
--- a/src/ui/control/geolocate_control.js
+++ b/src/ui/control/geolocate_control.js
@@ -283,8 +283,10 @@ class GeolocateControl extends Evented {
         const b = this._map.unproject([1, y]);
         const metersPerPixel = a.distanceTo(b);
         const circleDiameter = Math.ceil(2.0 * this._accuracy / metersPerPixel);
-        this._circleElement.style.width = `${circleDiameter}px`;
-        this._circleElement.style.height = `${circleDiameter}px`;
+        if (this._circleElement) {
+          this._circleElement.style.width = `${circleDiameter}px`;
+          this._circleElement.style.height = `${circleDiameter}px`;
+        }
     }
 
     _onZoom() {


andrewharvey:
Could you describe what situation lead to this code path being called without `this._circleElement` being defined?

We also have an assert a few lines up so it should be one or the other, the assert or this check.

andrewharvey:
cc @Meekohi

Meekohi:
Yeah if there is some way for this to come up in practice we should address the root issue -- otherwise I was hoping the `assert` makes it clear that `this._circleElement` always exists when this is called.

pakastin:
Well, it just broke aviamaps.com/map immediately when page loaded so 🤷‍♂️

andrewharvey:
@pakastin When I try a minimal test case it's working as expected, are you doing anything special?

It would be helpful if you could work the error you're seeing down into a minimal test case, so we can work out what's causing this issue.

pakastin:
Ok I found the cause. I was calling fitBounds before map load event. That fixed the problem. I would make a check though, since `this._circleElement` is created conditionally: https://github.com/mapbox/mapbox-gl-js/blob/7a4eb62e62d30de0d71f3c02191e3c79e50c9489/src/ui/control/geolocate_control.js#L363-L372

andrewharvey:
Oh I see at line https://github.com/mapbox/mapbox-gl-js/blob/7a4eb62e62d30de0d71f3c02191e3c79e50c9489/src/ui/control/geolocate_control.js#L129 `checkGeolocationSupport` is async so the `zoom` event is being registered and may be triggered before setupUI is run.

@pakastin or @Meekohi do you want to fix the root cause here? Should just be a matter of putting `this._map.on('zoom', this._onZoom);` inside `setupUI`, and if possible a unit test which would have caught this issue, if it's possible.

Meekohi:
Yup I think I get the flow -- I'll try to tackle this tonight unless someone can get to it earlier.

andrewharvey:
Thanks for reporting this bug @pakastin and sumbitting the PR. I'll close this one in favour of #9288 which addresses the root cause.

PR #9284, opened 2020-02-11
Use non-breaking space

pakastin:
Recent commit adding space between distance and unit introduced an ugly line break on mobile. Non-breaking space fixes that.


pakastin:
85A824D5-539C-47A9-A460-4F0381DBFABF_1_201_a

pakastin:
Yeah, no worries!

PR #9285, opened 2020-02-11
Make sure scale text won't overflow

pakastin:
On certain situations scale text overflows:
824482D2-4F15-4259-8EFD-45F6016BC0D0_1_201_a

Using `overflow: hidden` fixes that.

### DIFF
--- a/src/css/mapbox-gl.css
+++ b/src/css/mapbox-gl.css
@@ -536,6 +536,7 @@ a.mapboxgl-ctrl-logo.mapboxgl-compact {
     padding: 0 5px;
     color: #333;
     box-sizing: border-box;
+    overflow: hidden;
 }
 
 .mapboxgl-popup {


gely:
Personally, I think it looks better without left and right borders, and this issue somewhat goes away. I always override in CSS.

Screen Shot 2020-02-11 at 2 18 00 PM

asheemmamoowala:
@pakastin Could you please provide a screenshot of the new behavior and any other implications of accepting this change.

pakastin:
Looks good!

ryanhamley:
@pakastin Can you reliably reproduce this error? If so, how? I'm having trouble figuring out how to make the text overflow like in your screenshot. If we can reliably reproduce it, then perhaps there's an underlying bug in calculating the length of the scale that we can fix instead of this more cosmetic solution

ryanhamley:
I'm going to close this PR for now since a couple of questions were raised that were never answered. I still think it would be best to understand why this is happening. Feel free to reopen if you can provide an example where this is consistently reproducible.

Issue #8065, opened 2019-03-21
Invalid type on LngLatBounds.extend


fc:
It appears that `LngLatLike` and `LngLatBoundsLike` should also be allowed since the convert methods used here accept those types:
https://github.com/mapbox/mapbox-gl-js/blob/master/src/geo/lng_lat_bounds.js#L85-L91

I believe it is just a matter of changing it in the comments and in the type definition:
`LngLat | LngLatBounds` to `LngLat | LngLatBounds | LngLatLike | LngLatBoundsLike`, or to just `LngLatLike | LngLatBoundsLike` - not sure off the top of my head if the `instanceof` call will need `LngLat` to be explicitly set as a type (refer to lines 74 and 78).

I haven't created a PR since I don't have the time to test right now....but should be enough info for someone to look into it.

**mapbox-gl-js version**:
Latest version

### Expected Behavior
No type errors

### Actual Behavior
Invalid types
Something like this will fail TypeScript type checks:
```
bounds.extend([number, number]);
```

fc:
`@types/mapbox-gl` will also need to be updated once a fix here has been done:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mapbox-gl/index.d.ts#L883

asheemmamoowala:
@fc Thanks for the detailed issue, want to submit a PR?

marcofugaro:
any updates on this? should `@types/mapbox-gl` be updated to `v1.0.0`?

asheemmamoowala:
> any updates on this? 

We would welcome a PR contribution for the issue in the OP. 

> should @types/mapbox-gl be updated to v1.0.0?

Could you please create a separate issue for this

sgolbabaei:
closed by https://github.com/mapbox/mapbox-gl-js/pull/9293

Issue #8968, opened 2019-11-12
Map.panBy and Point.convert API


blq:
`map.panBy` specifies that its first argument can be a `PointLike` object. Though if you pass in an `{x, y}` literal it fails in `camera.js` due to the ```Point.convert(offset).mult()``` call that assumes that the output from `Point.convert` is a real Point instance, which is not what `Point.convert` spec actually says, it will pass through Array and other objects unchanged. 

/Fredrik Blomqvist

mourner:
Looks like we want to bring this PR in to fix this: https://github.com/mapbox/point-geometry/pull/12

blq:
Not sure worth opening another issue? But I just noticed a similar API vs implementation discrepancy for `LngLatBounds.extend(obj)`. The API docs claims `LngLatLike` is handled but doesn't work.

ryanhamley:
Hi @blq this discrepancy was fixed in https://github.com/mapbox/mapbox-gl-js/pull/9293 and released earlier this week as part of 1.9.0

blq:
@ryanhamley Thanks, should've searched better! But.. I think LngLatBounds still has bugs. I'll post some examples in that thread.

PR #9293, opened 2020-02-12
Letting `LngLatBounds.extend` accepts LngLatLike and LngLatBoundsLike as well.

sgolbabaei:
This PR is to address issue #8065 

Previously in the code base, `extend` function in `LngLatBounds` class would only accept LngLat and LngLatBounds. This resulted into seeing a typescript types check failure. 

With this change we will be able to pass an object of types: `LngLat`, `LngLatBounds`, `LngLatLike` or `LngLatBoundsLike` to `LngLatBounds.extend` function.

Also, previously, while we were checking to see if the passed object is an array or not, we have missed a case which `LngLatBoundsLike` could have be define as an array of four numbers. Updated the code and also the test file to capture that case as well.

tested the change using `debug/bounds.html`. Will appreciate more test scenarios to see if the change is working or not.

## Launch Checklist

 </changelog>`

### DIFF
--- a/src/geo/lng_lat_bounds.js
+++ b/src/geo/lng_lat_bounds.js
@@ -66,7 +66,7 @@ class LngLatBounds {
      * @param {LngLat|LngLatBounds} obj object to extend to
      * @returns {LngLatBounds} `this`
      */
-    extend(obj: LngLat | LngLatBounds) {
+    extend(obj: LngLatLike | LngLatBoundsLike) {
         const sw = this._sw,
             ne = this._ne;
         let sw2, ne2;
@@ -83,7 +83,9 @@ class LngLatBounds {
 
         } else {
             if (Array.isArray(obj)) {
-                if (obj.every(Array.isArray)) {
+                // if every element in the Obj array is also an array or
+                // if the LngLatBoundLike is defiend as a array of 4 numbers
+                if (obj.length > 2 || obj.every(Array.isArray)) {
                     return this.extend(LngLatBounds.convert(obj));
                 } else {
                     return this.extend(LngLat.convert(obj));
--- a/src/geo/lng_lat_bounds.js
+++ b/src/geo/lng_lat_bounds.js
@@ -61,9 +61,9 @@ class LngLatBounds {
     }
 
     /**
-     * Extend the bounds to include a given LngLatLike or LngLatBoundsLike.
+     * Extend the bounds to include a given LngLat, LngLatLike, LngLatBounds or LngLatBoundsLike.
      *
-     * @param {LngLatLike|LngLatBoundsLike} obj object to extend to
+     * @param {LngLat|LngLatLike|LngLatBounds|LngLatBoundsLike} obj object to extend to
      * @returns {LngLatBounds} `this`
      */
     extend(obj: LngLatLike | LngLatBoundsLike) {
--- a/src/geo/lng_lat_bounds.js
+++ b/src/geo/lng_lat_bounds.js
@@ -83,9 +83,7 @@ class LngLatBounds {
 
         } else {
             if (Array.isArray(obj)) {
-                // obj is LngLatBounds object if every element in the Obj array is also an array or
-                // if the LngLatBoundLike is defiend as a array of 4 numbers
-                if (obj.length > 2 || obj.every(Array.isArray)) {
+                if (obj.length === 4 || obj.every(Array.isArray)) {
                     return this.extend(LngLatBounds.convert(obj));
                 } else {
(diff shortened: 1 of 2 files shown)


sgolbabaei:
Not sure how to update this file:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mapbox-gl/index.d.ts#L1082

Is it auto generated?

test/unit/geo/lng_lat_bounds.test.js
```diff
@@ -81,6 +82,14 @@ test('LngLatBounds', (t) => {
         t.equal(bounds1.getNorth(), 15);
         t.equal(bounds1.getEast(), 15);
 
+        const bounds4 = new LngLatBounds([-20, -20, 20, 20]);
```
asheemmamoowala: It would be good to complete test coverage with tests/test cases that
- extend by the various `LngLatLike` inputs 
- check invalid inputs - those that don't match `LngLatLike` or `LngLatBoundsLike`

src/geo/lng_lat_bounds.js
```diff
@@ -83,7 +83,9 @@ class LngLatBounds {
 
         } else {
             if (Array.isArray(obj)) {
-                if (obj.every(Array.isArray)) {
+                // obj is LngLatBounds object if every element in the Obj array is also an array or
+                // if the LngLatBoundLike is defiend as a array of 4 numbers
```
asheemmamoowala: We usually add code comments to document non-obvious choices and higher-level problem solving. This comment doesn't seem necessary.

src/geo/lng_lat_bounds.js
```diff
@@ -61,12 +61,12 @@ class LngLatBounds {
     }
 
     /**
-     * Extend the bounds to include a given LngLat or LngLatBounds.
+     * Extend the bounds to include a given LngLatLike or LngLatBoundsLike.
      *
-     * @param {LngLat|LngLatBounds} obj object to extend to
+     * @param {LngLatLike|LngLatBoundsLike} obj object to extend to
```
asheemmamoowala: Update the doc to specify the bounds can be extended to include a point or another bounds.

test/unit/geo/lng_lat_bounds.test.js
sgolbabaei: sounds good, will add more test cases.

src/geo/lng_lat_bounds.js
ryanhamley: This should also be `LngLat|LngLatLike|LngLatBounds|LngLatBoundsLike` right?

sgolbabaei: not sure if it needs to be `LngLat|LngLatLike|LngLatBounds|LngLatBoundsLike` or `LngLatLike|LngLatBoundsLike`.
I have updated the input parameters to be`extend(obj: LngLatLike | LngLatBoundsLike)`but  I think each of the `LngLatLike` and `LngLatBoundsLike` would cover `LngLat` and `LngLatBounds` based on their definition:

`export type LngLatBoundsLike = LngLatBounds | [LngLatLike, LngLatLike] | [number, number, number, number];` & 
`export type LngLatLike = LngLat | {lng: number, lat: number} | {lon: number, lat: number} | [number, number];`

blq:
Hi, found this issue via #8968 
I think this still has errors, see this:
```
// empty constructor reference + two extend. correct 
var b1 = new mapboxgl.LngLatBounds();
b1.extend([10, 20]);
b1.extend([20, 30]);
b1.toString() === "LngLatBounds(LngLat(10, 20), LngLat(20, 30))";

// LngLatLike -> Error exception. Expected: same as "b1"
var b2 = new mapboxgl.LngLatBounds();
b2.extend({ lng: 10, lat: 20 });
b2.extend({ lng: 20, lat: 30 });
b2.toString() => Throws "Uncaught TypeError: Cannot read property 'toString' of undefined".


// Initial same pos Array constructor + extend reference. correct.
var b3 = new mapboxgl.LngLatBounds([10, 20], [10, 20]);
b3.extend([20, 30]);
b3.toString() === "LngLatBounds(LngLat(10, 20), LngLat(20, 30))";

// LngLatLike invalid result. Expected: same as "b3"
var b4 = new mapboxgl.LngLatBounds({ lng: 10, lat: 20 }, { lng: 10, lat: 20 });
b4.extend({ lng: 20, lat: 30 });
b4.toString() === "LngLatBounds(LngLat(10, 20), LngLat(10, 20))";
```

ryanhamley:
cc @sgolbabaei See the comment above

PR #9305, opened 2020-02-18
Clearing the underlying source's tiles while removing a layer.

### LINKED ISSUE #8634: setFeatureState Throws Error
**mapbox-gl-js version**: v1.2.1

**browser**: Chrome Version 70.0.3538.77 (Official Build) (64-bit)

### Steps to Trigger Behavior

 1. Add vector tile source S
 2. Add 2 layers, layer A and layer B, both using the same source-layer from vector tile source S
 3. use setFeatureState on layer A
4. remove layer B
5. use setFeatureState on layer A

### Link to Demonstration

https://jsbin.com/yuruvibiqa/edit?html,js,output

I removed my mapboxgl.accessToken.
The error occurs after the setTimeout function removes a layer.

### Expected Behavior

setFeatureState works as intended.

### Actual Behavior

setFeatureState throws the following error:

Uncaught TypeError: Cannot read property 'queryRadius' of undefined
    at Ou.setFeatureState (tile.js:464)
    at Vu.coalesceChanges (source_state.js:154)
    at i.prepare (source_cache.js:170)
    at ao.render (painter.js:340)
    at r._render (map.js:1753)
    at map.js:1837

sgolbabaei:
<changelog>Bug</changelog>
This PR fixes #8634 

A summary of the problem:

if we remove a layer, the layer still persist in the existing main thread's tiles and buckets list and if we use SetFeatureState(), while it goes through all the buckets, it gets to invalid layers(deleted layers).

An alternative fix to this problem could be any time we are removing layers, we clear the the underlying source's tiles. Basically call `this._updatedSources[layer.source] = 'clear';` in
https://github.com/mapbox/mapbox-gl-js/blob/master/src/style/style.js#L736
which this could potentially be computationally expensive and would result in seeing a glitch which is not very pleasant.

ezgif com-video-to-gif

Noticed in the gif above, the layer is being removed on click on the map. You can see the glitch.

We can use `this._updatedSources[layer.source] = 'reload';` but that will not fix this issue as well since reload will use the old version of tiles until the workers generate the new version of tiles..

This change won't fix the root cause of this problem (having different versions of tiles in workers and main thread) but will fix this instance of it. Maybe in future we can work on refactoring which can make it easy to keep track of different versions of tiles in workers and main threads. 

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog></changelog>`

### DIFF
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -750,6 +750,7 @@ class Style extends Evented {
         this._layerOrderChanged = true;
         this._changed = true;
         this._removedLayers[id] = layer;
+        this._updatedSources[layer.source] = 'clear';
         delete this._layers[id];
         delete this._updatedLayers[id];
         delete this._updatedPaintProps[id];
--- a/src/style/style.js
+++ b/src/style/style.js
@@ -750,7 +750,6 @@ class Style extends Evented {
         this._layerOrderChanged = true;
         this._changed = true;
         this._removedLayers[id] = layer;
-        this._updatedSources[layer.source] = 'clear';
         delete this._layers[id];
         delete this._updatedLayers[id];
         delete this._updatedPaintProps[id];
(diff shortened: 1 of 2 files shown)


src/style/style.js
```diff
@@ -769,6 +769,17 @@ class Style extends Evented {
         return this._layers[id];
     }
 
+    /**
+     * Return a boolean specifying if the layer is valid (not deleted) with the given `id`.
+     *
+     * @param {string} id - id of the desired layer
+     * @returns {boolean} a boolean specifying if the given layer is valid
+     */
+    isValidLayer(id: string): boolean {
```
asheemmamoowala: nit: Can this function be renamed  to indicate that it is checking for the existence of a layer not its validity. Some ideas: `hasLayer` or `isCurrentLayer`

Issue #9327, opened 2020-02-23
Map displays labels on a diagnoal


Incubbus:
**mapbox-gl-js version**: 1.8.0

**browser**: Chrome, Edge, Firefox

### Steps to Trigger Behavior

 1. Create a map as in [the tutorial](https://docs.mapbox.com/mapbox-gl-js/overview/).

### Link to Demonstration

https://codepen.io/Incubbus/pen/BaNQGLW

### Expected Behavior

A map with correct label positioning.

### Actual Behavior

A map where labels are all on one diagonal.

![Image of buggy map](https://i.stack.imgur.com/jVTmN.jpg)
### Additional Info

When using 1.7.0 it works.

clshortfuse:
Getting this on a Windows 10 and Windows 7 PC. I'm guessing it's driver issue because they both share Intel HD drivers.

Rosoyuki:
I have the same problem. In all examples of the website. Chrome, Firefox, Opera.
PD: Internet explorer displays normal.

Incubbus:
As @clshortfuse mentioned, it **may** happen mostly on windows. Though an acquaintance told me he had this issue appear on his iPhone with an earlier version (1.7,x i think).
This is a little terrifying, considering people use this for their production-grade products, too).

pakastin:
Some of our clients are having the exact issue with Windows and Intel HD Graphics 4600 card:
![IMG_20200211_213322](https://user-images.githubusercontent.com/1475902/75145903-98016200-5702-11ea-87c8-899a36dae908.jpg)

..and another client as well:
![Photo 2020-02-24 12 07 43 em](https://user-images.githubusercontent.com/1475902/75145769-4ce74f00-5702-11ea-9c0a-54217664a622.png)

akx:
If 1.7.0 works fine, then if you have the faulting hardware at hand, you can help debug this with a [Git Bisect](https://www.metaltoad.com/blog/beginners-guide-git-bisect-process-elimination) process. (I only have Nvidia and/or Macs, so I can't help here.)

```
$ git bisect start
$ git bisect bad v1.8.0
$ git merge-base v1.8.0 v1.7.0
74b15736d414023345abab69b91995d949f7d11e
$ git bisect good 74b15736d414023345abab69b91995d949f7d11e
Bisecting: 23 revisions left to test after this (roughly 5 steps)
```
[...]

PR #9333, opened 2020-02-24
Bugfix for #9327

karimnaaji:
Critical bug fix for https://github.com/mapbox/mapbox-gl-js/issues/9327

Fix confirmed by two users (@matthias-burgfried, @Incubbus).

Once we get a repro we can reintroduce that feature appropriately.


karimnaaji:
Follow up for this PR in https://github.com/mapbox/mapbox-gl-js/issues/9334

Issue #9126, opened 2019-12-18
Memory usage doesn't come down when the map is removed/unmounted from the DOM


iamanvesh:
While #8771 fixed the memory issue while using the map for prolonged period of time, we found that the memory used by the map is not being garbage collected when it un-mounts from the DOM. Please find attached the screenshots of the memory usage (react app). To test this behavior in isolation I've created a simple app which has only two pages, one of which loads a map onto the DOM and the other just has some text. Please use a resource intensive style URL (which has at least 4-5 layers and >300-500 features) for testing this scenario. Our dataset has >100K features spread across the USA.

**Initial load of the app (Map has not been mounted yet):**

![image](https://user-images.githubusercontent.com/6526938/71105687-f9c5bd00-21e3-11ea-9762-ec830b8baa96.png)

**Peak memory usage of the map (after panning/zooming around multiple times):**

![image](https://user-images.githubusercontent.com/6526938/71105749-1661f500-21e4-11ea-8f41-41b469af5935.png)

**~15 seconds after the map unmounts:**

![image](https://user-images.githubusercontent.com/6526938/71105818-2aa5f200-21e4-11ea-84ea-fd6c70230a90.png)

**When the map unmounts after repeating these steps 4 times:**

![image](https://user-images.githubusercontent.com/6526938/71106100-b91a7380-21e4-11ea-8636-1e91dcd6267a.png)

**mapbox-gl-js version**: 1.6.0

**browser**: Predominantly on Safari but Chrome retains some memory too.

### Steps to Trigger Behavior
1. Run the app by building the production version of the app [steps in the readme].
2. Open the home page on the browser and start watching the memory (using Task manager in chrome or Activity Monitor / Safari Timeline). It should be ~50-60MB.
3. Now open the map page and zoom/pan around until you see the memory grow upto 500-600mb.
4. Go back to the empty page.
5. Repeat the steps 3 & 4 for some time and when you finally come back to the empty page after 4th or 5th time, the app retains a significant amount of memory (~500mb).

### Link to Demonstration

Without react: https://jsbin.com/zimigow/edit
With react: https://github.com/iamanvesh/mapbox-memory-leak-sample

### Expected Behavior
The memory usage should come down to what is observed on the initial page load.

### Actual Behavior
The memory usage stays at around ~500mb even after waiting for ~30-60 seconds.

vakila:
Hi @iamanvesh - thanks for using Mapbox & for the detailed report! 

[...]

PR #9337, opened 2020-02-24
Patch memory leaks caused by event handlers bound to document

arindam1993:
This potentially addresses https://github.com/mapbox/mapbox-gl-js/issues/9126. 

We temporarily bind handlers to `document` in our event handlers when certain interactions are in flight. If the map is removed during that time, it causes the entire `Map` object to be retained, since its in the scope of the event handlers.

`<changelog>Fixed a memory leak that occurred if the map was removed while certain user interactions were in progress.</changelog>`

### DIFF
--- a/src/ui/handler/box_zoom.js
+++ b/src/ui/handler/box_zoom.js
@@ -147,13 +147,20 @@ class BoxZoomHandler {
         }
     }
 
-    _finish() {
-        this._active = false;
-
+    _unbind() {
         window.document.removeEventListener('mousemove', this._onMouseMove, false);
         window.document.removeEventListener('keydown', this._onKeyDown, false);
         window.document.removeEventListener('mouseup', this._onMouseUp, false);
+    }
 
+    teardown() {
+        this._unbind();
+    }
+
+    _finish() {
+        this._active = false;
+
+        this._unbind();
         this._container.classList.remove('mapboxgl-crosshair');
 
         if (this._box) {
--- a/src/ui/handler/box_zoom.js
+++ b/src/ui/handler/box_zoom.js
@@ -153,7 +153,7 @@ class BoxZoomHandler {
         window.document.removeEventListener('mouseup', this._onMouseUp, false);
     }
 
-    teardown() {
+    remove() {
         this._unbind();
     }
 
--- a/src/ui/handler/box_zoom.js
+++ b/src/ui/handler/box_zoom.js
@@ -147,20 +147,16 @@ class BoxZoomHandler {
         }
     }
 
-    _unbind() {
+    remove() {
         window.document.removeEventListener('mousemove', this._onMouseMove, false);
         window.document.removeEventListener('keydown', this._onKeyDown, false);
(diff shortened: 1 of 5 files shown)


iamanvesh:
Hi @arindam1993,
I've tried using the build from this branch with the sample code that I provided in #9126 (jsbin without react) and I don't see any improvement in memory usage. It's still retaining a lot of memory in Safari.

arindam1993:
@iamanvesh I posted some results from some digging in #9126 , could we continue the conversation there?

src/ui/handler/box_zoom.js
```diff
@@ -147,13 +147,20 @@ class BoxZoomHandler {
...
-    _finish() {
-        this._active = false;
-
+    _unbind() {
         window.document.removeEventListener('mousemove', this._onMouseMove, false);
         window.document.removeEventListener('keydown', this._onKeyDown, false);
         window.document.removeEventListener('mouseup', this._onMouseUp, false);
+    }
+
+    teardown() {
```
kkaefer: In the rest of the codebase, we use `remove()` for this action. Let's keep up the consistency.

src/ui/handler/box_zoom.js
```diff
@@ -147,13 +147,20 @@ class BoxZoomHandler {
         }
     }
 
-    _finish() {
-        this._active = false;
-
+    _unbind() {
```
kkaefer: Let's guard against the race condition/double invocation that could occur when the user removes the map while a handler is active.

arindam1993: I don't think thats possible, infact this PR would that from happening.
Scenario one:
If a handler callback is executing, `map.remove()` can't be called, because well, JS is single threaded.
Scenario two:
And if `map.remove()` is called while a handler is still active i.e the user has the mouse held down,  the callback gets unbound, so it would never invoke once the user eventually say, does a mouseup. Previous to this it would actually invoke the mouseup.

kkaefer: > If a handler callback is executing, `map.remove()` can't be called, because well, JS is single threaded.
You can call `map.remove()` from the callback ;)

arindam1993: Yea.. but why would we ever do that?

kkaefer: It might not be intentional, e.g. an application could transition away from a map view (and calling `map.remove()`) while the user is still interacting with it.

arindam1993: But that would be Scenario 2, which this fixes? `map.remove()` would still not be synchronously called in our callback, because thats our code and not user code?
That transition away would run in a different run-loop tick, so any of the enteraction callbacks would have been fully executed.

mourner:
We should probably wait until #9365 lands for an easier time reconciling the code.

arindam1993:
Sticking the `DO NOT MERGE` label until #9365  is merged.

arindam1993:
Closing since #9508 addresses thise.

PR #9343, opened 2020-02-25
Release v1.8.1

kkaefer:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mapbox-gl",
   "description": "A WebGL interactive maps library",
-  "version": "1.8.1-dev",
+  "version": "1.8.2-dev",
```
mourner: We should bump this only in `master`, not in the release branch for the release tag, right?

kkaefer: No, the release-branch is on -dev, but there's an intermediary commit in this PR that sets it to `1.8.2`. I'll tag this commit as the release. I've been trialing this in this release to avoid intermediary builds made from this branch showing up as an actual version rather than a `-dev` version.

PR #9346, opened 2020-02-26
Remove deprecated GPU timing logging

mourner:
All the current browsers [no longer support](https://caniuse.com/#search=EXT_disjoint_timer_query) the `EXT_disjoint_timer_query` extension that powered the GPU timing event — this happened because of a [security exploit](https://www.vusec.net/projects/glitch/) that took advantage of it.

Since we can no longer use it, let's remove it.

```
<changelog>Removed `gpu-timing-frame` and `gpu-timing-layer` events because current browsers no longer support the instrumentation API that powered them.</changelog>
```

### DIFF
--- a/src/gl/context.js
+++ b/src/gl/context.js
@@ -63,7 +63,6 @@ class Context {
     extTextureFilterAnisotropic: any;
     extTextureFilterAnisotropicMax: any;
     extTextureHalfFloat: any;
-    extTimerQuery: any;
 
     constructor(gl: WebGLRenderingContext) {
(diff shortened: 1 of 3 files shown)


mourner:
The support tables are outdated — looks like it have been readded, at least in Chrome.

JannikGM:
For the record:
This is also available in Firefox, but needs `webgl.enable-privileged-extensions: true` in about:config.

If the extension is not enabled but the listeners are created, mapbox will actually cause errors. This feature deserves some sanity checks.

PR #9352, opened 2020-02-27
Introduce `within` expression

zmiao:
## Launch Checklist

This PR is porting from GL Native:
1.  `within` expression implementation for feature with Point geometry type and using it with paint property or filter expression: mapbox/mapbox-gl-native#16157.
2. Using `within` with layout property: mapbox/mapbox-gl-native#16194
3. Support Line within Polygon: mapbox/mapbox-gl-native#16220

within expression now supports Point/LineString Feature

    Add a new expression  `within`, which would enable checking whether a feature is inside a pre-defined geometry set/boundary or not. This `within` expression returns a boolean value: true  indicates that the feature being evaluated is inside the geometry set.

The returned value can be then consumed as input by another expression or used directly by a paint/layer property, or combine with filter expression.

```
"circle-color": ["case", ["within", {"type": "Polygon", 
   "coordinates": [[[0, 0], [5, 0], [5, 5], [0, 5], [0, 0]]]}],  
   [ "rgb", 153, 128, 250], ["rgb", 249, 138,175]]
```
```
"text-field": ["case", ["within", {"type": "Polygon", 
     "coordinates": [[[0, 0], [5, 0], [5, 5], [0, 5], [0, 0]]]}],  
      "inside", "outside"]
```
```
 "filter": ["within", {"type": "Polygon", 
     "coordinates": [[[0, 0], [5, 0], [5, 5], [0, 5], [0, 0]]]}]
```
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: 
`<changelog>
Add `within` expression for testing whether an evaluated feature lies within a given GeoJSON object: `"icon-opacity": ["case", ["within", "some-polygon"], 1,  0]`,  `"filter" : ["within", "some-polygon"]`
</changelog>`
 - [x] update js compatibility version in v8.json after pr get approved.

 @mapbox/studio  @mapbox/map-design-team


zmiao:
@mourner Thanks for the suggestion. I have updated my pr, now it will pre check if geometry is needed for filter or not

tristen:
👋 @zmiao this is exciting work! I'm curious about the decision to use `coordinates` rather than associating it with data requested from the source object. i.e 

```js
{
  'circle-color': [
    'case',
    [
      'within', {
        source: 'some-data',
        'source-layer': 'some-source-layer',
        filter: ['==', ['get', 'class'], 'points-that-should-be-red']
      }
    ],  
    "red",
    "blue"
  ]
}
```

cc/ @mapbox/map-design-team for visibility

test/expression.test.js
```diff
@@ -4,6 +4,57 @@ import {isFunction} from '../src/style-spec/function';
...
+        const geom = [];
+        const ring = lines[i];
+        for (let j = 0; j < ring.length; j++) {
+            convertPoint(ring[j], canonical, geom);
+        }
+        out.push(geom);
+    }
+}
+
+function getGeomtry(feature, geometry, canonical) {
```
1ec5: Typo: geometry.

src/style-spec/expression/definitions/within.js
```diff
@@ -0,0 +1,273 @@
...
+        this.polyBBox = [Infinity, Infinity, -Infinity, -Infinity];
+        calcBBox(this.polyBBox, this.geometries.coordinates, this.geometries.type);
+    }
+
+    static parse(args: $ReadOnlyArray<mixed>, context: ParsingContext) {
+        if (args.length !== 2)
+            return context.error(`'within' expression requires exactly one argument, but found ${args.length - 1} instead.`);
+        if (isValue(args[1])) {
+            const geojson = (args[1]: Object);
+            if (geojson.type === 'FeatureCollection') {
```
1ec5: Can the `within` expression’s sole argument be an expression that evaluates to a geometry, or does it have to be a literal geometry?

src/style-spec/expression/index.js
```diff
@@ -25,12 +25,15 @@ import type {Result} from '../util/result';
...
+import type Point from '@mapbox/point-geometry';
+import type {CanonicalTileID} from '../../source/tile_id';
 
 export type Feature = {
     +type: 1 | 2 | 3 | 'Unknown' | 'Point' | 'MultiPoint' | 'LineString' | 'MultiLineString' | 'Polygon' | 'MultiPolygon',
     +id?: any,
     +properties: {[_: string]: any},
-    +patterns?: {[_: string]: {"min": string, "mid": string, "max": string}}
+    +patterns?: {[_: string]: {"min": string, "mid": string, "max": string}},
+    +geometry?: Array<Array<Point>>
```
1ec5: This looks like the structure of a Polygon geometry, but wouldn’t a Point or LineString would have a different type? Or is that variation irrelevant to the type annotation here?

src/style-spec/expression/definitions/within.js
```diff
@@ -0,0 +1,273 @@
...
+            } else if (geojson.type === 'Feature') {
+                const type = geojson.geometry.type;
+                if (type === 'Polygon' || type === 'MultiPolygon') {
+                    return new Within(geojson, geojson.geometry);
+                }
+            } else if (geojson.type  === 'Polygon' || geojson.type === 'MultiPolygon') {
+                return new Within(geojson, geojson);
+            }
+        }
+        return context.error(`'within' expression requires valid geojson source that contains polygon geometry type.`);
```
ryanhamley: if this doesn't take an actual source object as an argument as @tristen suggested in another comment, then using the word `source` here may be confusing since it has a specific meaning in the context of GL JS

test/expression.test.js
ryanhamley: typo: `getGeomtry`

src/style-spec/expression/definitions/within.js
```diff
@@ -0,0 +1,273 @@
...
+        this.geometries = geometries;
+        this.polyBBox = [Infinity, Infinity, -Infinity, -Infinity];
+        calcBBox(this.polyBBox, this.geometries.coordinates, this.geometries.type);
+    }
+
+    static parse(args: $ReadOnlyArray<mixed>, context: ParsingContext) {
+        if (args.length !== 2)
+            return context.error(`'within' expression requires exactly one argument, but found ${args.length - 1} instead.`);
+        if (isValue(args[1])) {
+            const geojson = (args[1]: Object);
```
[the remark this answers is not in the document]
1ec5: A nested expression would be fine with NSExpression on iOS/macOS; not sure about Android. I asked about this too in https://github.com/mapbox/mapbox-gl-js/pull/9352#discussion_r385956784 because I wasn’t sure whether to specifically prohibit a nested expression and only allow a literal shape.

src/style-spec/expression/definitions/within.js
zmiao: We decided that the first version of `within` Expression will have inlined GeoJSON Object as arguments. Follow-up would be something like https://github.com/mapbox/mapbox-gl-native-team/issues/181. But it needs to be carefully designed before starting implementation.

src/style-spec/expression/index.js
zmiao: It keeps the same structure as Geometry that loaded from VectorTileFeature

```
loadGeometry(): Array<Array<Point>>;
```

src/style-spec/expression/definitions/within.js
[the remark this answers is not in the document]
zmiao: From my understanding, by applying the geometry with extra expression does the same thing. 
If we want to use `within` together with `get`, it needs to be format like 
```
[`within`, [`get`, `polygon`, `geojsonObj`]]
```
We any way need to pass the geojsonObj.  

Android SDK is trying to add binding for within expression. They are passing the argument as a map object, please check here https://github.com/mapbox/mapbox-gl-native-android/pull/198. 

I think one thing need to be updated is inside MGLConversion.h, currently `toGeoJSON()` is not implemented.

zmiao:
> 👋 @zmiao this is exciting work! I'm curious about the decision to use coordinates rather than associating it with data requested from the source object. i.e

> ```
> {
>   'circle-color': [
>     'case',
>     [
>       'within', {
>         source: 'some-tileset',
>         'source-layer': 'some-source-layer',
>         filter: ['==', ['get', 'class'], 'points-that-should-be-red']
>       }
>     ],  
>     'red',
>     'blue'
>   ]
> }
> ```

@tristen Thanks for bringing up this question. 

From the expression you wrote above, the filter will give all the features that have pre-defined `class `included, the result may change if geometry boundary cover across several tiles. The `within` algorithm assume the geometry boundary is constant.

By consuming `filter` expression inside, it will cause filtering all of the features inside one tile per within expression evaluation, it would be very inefficient while the features are pretty dense. 

When processing expression evaluation, the caller is each feature, however, `filter` expression assumes to evaluate each feature inside the tile, which means all the features needs to be passed to the evaluation context.  Right now there is no proper way to filter features at evaluation stage, and filter is targeting for each tile, but the geometry boundary may cover across several tiles.

It is a very good suggestion that taking advantage of using source data. And it is true that inlined GeoJSON boundary is not the best choice, making it sharable/fetchable would be more flexible, here is similar idea of fetching geometry from source data in this issue ticket: https://github.com/mapbox/mapbox-gl-native-team/issues/181.

src/style-spec/expression/definitions/within.js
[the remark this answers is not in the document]
zmiao:
```
[`within`, [`get`, `polygon`, `geojsonObj`]]
```
 or 
```
[`within`, [`literal`, `geojsonObj`]]
```
Another reason that it is better not to use a nested expression (like above) is, it will require evaluation of the inner expression for getting the geojsonObj every time when invoking `wtihin` evaluation, and it can not ensure if the geojsonObj provides valid geometry or not. If the geometry is invalid, it won't issue error when parsing the expression, only until evaluation stage, but that would be too late.

On the other hand, if the features are very dense or the geojsonObj contains pretty big amounts of points for the polygon, the process would be really inefficient. 

However, I do find something that `within` expression would break the style spec :
> 

> Expressions are represented as JSON arrays. The first element of an expression array is a string naming the expression operator, e.g. "*" or "case". Elements that follow (if any) are the arguments to the expression. Each argument is either a literal value (**a string, number, boolean, or null)**, or another expression array.

Style spec emphasized that valid argument could only be` a string, number, boolean, or null`,  but `within` requires  `object` argument.  @asheemmamoowala @mourner @mapbox/gl-js

src/style-spec/reference/v8.json
```diff
@@ -3454,6 +3454,15 @@
           }
         }
       },
+      "within": {
+        "doc": "Returns `true` if the feature being evaluated is inside the pre-defined geometry boundary, `false` otherwise. The arguments are required to be valid GeoJSON object which contains `Polygon` geometries. Currently only features with `Point` or `LineString` geometry types are supported.",
```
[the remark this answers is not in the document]
zmiao: @ryanhamley Thank you very much for the advice! I updated the phrasing a little bit, welcome for the feedback ^ ^

src/style-spec/expression/definitions/within.js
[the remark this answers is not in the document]
1ec5: > Style spec emphasized that valid argument could only be` a string, number, boolean, or null`,  but `within` requires  `object` argument.

I think it was only a matter of time before array would have to start holding objects anyways. We kind of skirted around the issue when implementing `collator` objects, because we limited their use to a specific argument of a specific set of expression operators.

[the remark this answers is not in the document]
1ec5: > Another reason that it is better not to use a nested expression (like above) is, it will require evaluation of the inner expression for getting the geojsonObj every time when invoking `wtihin` evaluation, and it can not ensure if the geojsonObj provides valid geometry or not. If the geometry is invalid, it won't issue error when parsing the expression, only until evaluation stage, but that would be too late.

I think we’re on the same page. I was only referring to supporting something like `["within", ["match", …, geoJSONObj]]`. I wasn’t referring to `literal` expressions per se; sorry for being unclear with that terminology.

src/style-spec/expression/definitions/within.js
```diff
@@ -0,0 +1,273 @@
...
+            }
+        }
+        return false;
+    }
+
+    eachChild() {}
+
+    possibleOutputs() {
+        return [true, false];
+    }
```
mourner: This will need a rebase after #9355

src/style-spec/feature_filter/index.js
```diff
@@ -82,10 +84,22 @@ function createFilter(filter: any): FeatureFilter {
...
 }
 
+function filterNeedGeometry(filter) {
+    if (filter === true || filter === false) {
+        return false;
+    }
+    if (!Array.isArray(filter) || filter.length === 0) {
+        return false;
+    }
+    return filter[0] === 'within';
```
mourner: Nit: this could be simplified to `return filter && filter[0] === 'within'` — no need for extensive checks.

src/data/bucket/circle_bucket.js
```diff
@@ -86,25 +87,32 @@ class CircleBucket<Layer: CircleStyleLayer | HeatmapStyleLayer> implements Bucke
...
-                    geometry,
-                    patterns: {},
-                    sortKey
-                };
-
-                bucketFeatures.push(bucketFeature);
-            }
+            const needGeometry = this.layers[0]._featureFilter.needGeometry;
+            const evaluationFeature = {type: feature.type,
+                id: ('id' in feature ? feature.id : null),
```
mourner: You need to use `id` rather than `feature.id` here for the GeoJSON `promoteFeatureID` feature to work.
```suggestion
                id,
```

src/data/bucket/circle_bucket.js
```diff
@@ -86,25 +87,32 @@ class CircleBucket<Layer: CircleStyleLayer | HeatmapStyleLayer> implements Bucke
...
+                properties: feature.properties,
+                geometry: needGeometry ? loadGeometry(feature) : []};
```
mourner: Nitpick, but this code feels quite repetitive — any way we could handle the branching with geometry loading in a cleaner way? Not necessarily in this PR — we could do it as a follow-up, but this is a good candidate for refactoring. Another thing that doesn't feel good is passing `canonical` parameter across the whole code. One way we could try cleaning this up is introduce a new `EvaluationFeature` class that wraps a vector tile feature, handles lazy loading of geometry, and contains `canonical` as state for later use.

src/style-spec/expression/definitions/within.js
```diff
@@ -0,0 +1,273 @@
...
+    return ((p1[1] > p[1]) !== (p2[1] > p[1])) && (p[0] < (p2[0] - p1[0]) * (p[1] - p1[1]) / (p2[1] - p1[1]) + p1[0]);
+}
+
+// ray casting algorithm for detecting if point is in polygon
+function pointWithinPolygon(point, rings) {
+    let inside = false;
+    for (let i = 0, len = rings.length; i < len; i++) {
+        const ring = rings[i];
+        for (let j = 0, len2 = ring.length; j < len2 - 1; j++) {
+            if (onBoundary(point, ring[j], ring[j + 1])) return false;
```
mourner: I'm not sure this check is necessary. As far as I know, the general convention for geometric predicates is to not accept objects on the boundary as being contained inside the polygon. See for example an answer to this question https://gis.stackexchange.com/questions/170264/python-point-in-polygon-boundary-and-vertex-check-ray-casting

src/style-spec/expression/definitions/within.js
```diff
@@ -0,0 +1,273 @@
...
+        // q1->p1 (x1, y1), q1->p2 (x2, y2), q1->q2 (x3, y3)
+        const x1 = p1[0] - q1[0];
+        const y1 = p1[1] - q1[1];
+        const x2 = p2[0] - q1[0];
+        const y2 = p2[1] - q1[1];
+        const x3 = q2[0] - q1[0];
+        const y3 = q2[1] - q1[1];
+        if ((x1 * y3 - x3 * y1) * (x2 * y3 - x3 * y2) < 0) return true;
+        return false;
+    };
```
mourner: All the anonymous functions here (`perp`, `twoSided`) should be cut out outside because JS will allocate the functions on every `lineIntersectLine` call, unlike C++ that usually optimizes lambdas away.

src/style-spec/expression/definitions/within.js
```diff
@@ -0,0 +1,273 @@
...
+    // If lines are intersecting with each other, the relative location should be:
+    // a and b lie in different sides of segment c->d
+    // c and d lie in different sides of segment a->b
+    if (twoSided(a, b, c, d) && twoSided(c, d, a, b)) return true;
+    return false;
+}
+
+function lineIntersectPolygon(p1, p2, polygon) {
+    for (let i = 0; i < polygon.length; ++i) {
+        const ring = polygon[i];
```
mourner: Nit: you can do `for (const ring of polygon)`.

src/data/bucket/circle_bucket.js
zmiao: I tried with this way, but it will cause the render test `regressions/mapbox-gl-js#4235` failure, as id here is `undefined` instead of `null`,  as `filter-has-id` only compare with `null` here
https://github.com/mapbox/mapbox-gl-js/blob/e5db688fe1a583aa002a5387b15218d6a9d3f949/src/style-spec/expression/definitions/index.js#L455-L458

src/style-spec/expression/definitions/within.js
zmiao: It is necessary, I tried to test the points on the boundary, without this check it would fail.
Referring from https://en.wikipedia.org/wiki/Point_in_polygon:
>  This method won't work if the point is on the edge of the polygon, or if the ray intersects with a corner of the polygon.

mourner: What I mean is that the point on the boundary shouldn't be considered as "contained within" the polygon according to conventional topology rules. See the articles linked in the answer above:
- https://lin-ear-th-inking.blogspot.com/2007/06/subtleties-of-ogc-covers-spatial.html
- https://shapely.readthedocs.io/en/latest/manual.html#binary-predicates

zmiao: If I understood correctly,  `within` shall return false if the point is on the boundary, so in the code, if a point is on the boundary, it immediately return `false`, it will ignore the following ray casting checks.

src/style-spec/expression/definitions/within.js
zmiao: Thanks 🙇‍♀

src/data/bucket/circle_bucket.js
zmiao: I will think about it and do it as a follow-up pr. Thank you for the suggestion!

src/style-spec/feature_filter/index.js
zmiao: I remove this checking function. But since the return type from `convertFilter(filter)` is mixed, I found it requires type checking before accessing the array member.

src/data/bucket/circle_bucket.js
mourner: Maybe we should change this expression to `!=` (which also compares against `undefined`) then?

zmiao: Agree! I was afraid to do so because I wander if there is some special logic behind it. I will change it!

chloekraw:
@zmiao thanks for the thoughtful changelog entry! I think that many of the implementation details are covered in documentation, and that a lighter changelog entry coupled with a syntax example can adequately convey what developers can do with the feature. My suggestion is to tighten up the changelog entry from

`<changelog> Introduce a new expression within. The within expression enables checking whether a feature is inside a pre-defined geometry set/boundary or not. This within expression returns a boolean value, true indicates that the feature being evaluated is inside the geometry set. The returned value can be then consumed as an input by another expression or used directly by a paint or layout property. </changelog>`

to 

`<changelog>Add `within` expression for testing whether an evaluated feature lies within a given GeoJSON object: `"icon-opacity": ["case", ["==", ["within", "some-polygon"], true], 1,
      ["==", ["within", "some-polygon"], false],  0]`</changelog>`

zmiao:
@chloekraw Thank you for the suggestion!

atyshka:
Thanks for adding this feature! Is there any existing way to reverse the arguments, i.e when applying a within to polygon features, test for a point within each of those polygons? It would require rearranging the variable order. Is it possible to do that in a hacky way (maybe with let/var) or would a pr be necessary to add a similar "contains" decision that is identical to within, but with order reversed? See #9891 for more info

strech345:
I'm missing a description of the style specification for `within`. Is it some where?

what does exactly _within_ means? Need a line be full included?
Are MultiPoints, MultiLines on the left and MultiPolygons on the right side supported?
will the expression also be extended to check also polygon within polygon?

Also would be nice to check if centroids of polygons are within some poylgon.

ryanhamley:
@strech345 the documentation is here https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#within

>Need a line be full included?

Yes. If any part of a line falls outside the boundary of the GeoJSON object, `within` returns `false`

>Are MultiPoints, MultiLines on the left and MultiPolygons on the right side supported?

The object supplied to the `within` expression (the "right" side) can be a Polygon, MultiPolygon, Feature, or FeatureCollection. The object checked to be within the boundary must be a Point or Line.

>will the expression also be extended to check also polygon within polygon?

There's no plans at this time for this

strech345:
@ryanhamley thanks

PR #9355, opened 2020-02-27
Simplify text-font static output validation

mourner:
Simplifies output validation for data-driven `text-font` property. We never actually use `possibleOutputs` of expressions except when validating if they're statically defined.

### DIFF
--- a/src/style-spec/expression/compound_expression.js
+++ b/src/style-spec/expression/compound_expression.js
@@ -39,8 +39,8 @@ class CompoundExpression implements Expression {
         this.args.forEach(fn);
     }
 
-    possibleOutputs() {
-        return [undefined];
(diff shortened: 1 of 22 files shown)


asheemmamoowala:
@mourner `possibleOutputs` seem to be used in GL-Native, in at least one place:

https://github.com/mapbox/mapbox-gl-native/blob/8be135231d9efe41a3b12037518d02b36104e8cf/src/mbgl/style/layers/symbol_layer_impl.cpp#L36-L40

cc @pozdnyakov @mapbox/gl-native

mourner:
@asheemmamoowala yes, and it's the same single place as here in GL JS for which tracking `possibleOutputs` as an array is an overkill.

PR #9358, opened 2020-02-28
Remove unused struct code

mourner:
- Removes array struct setters that were not used anywhere in the code — now the exceptions are explicitly listed in the generation code and they are just a couple. 
- Removes struct classes that were unused due to having explicit array getters instead.

Saves 453 bytes in the gzipped bundle.

### DIFF
--- a/src/util/struct_array.js.ejs
+++ b/src/util/struct_array.js.ejs
@@ -46,8 +46,12 @@ for (const {name, member, component} of components) {
     const componentAccess = `this._structArray.${member.view}[${index}]`;
 -%>
     get <%=name%>() { return <%=componentAccess%>; }
+<%
+if (name === 'crossTileID' || name === 'placedOrientation' || name === 'hidden') {
+-%>
     set <%=name%>(x: number) { <%=componentAccess%> = x; }
 <%
+}
 }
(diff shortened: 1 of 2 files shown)


mourner:
GitHub got stuck with checks for some reason but Circle CI build fully passed.

src/util/struct_array.js.ejs
```diff
@@ -46,9 +46,13 @@ for (const {name, member, component} of components) {
     const componentAccess = `this._structArray.${member.view}[${index}]`;
 -%>
     get <%=name%>() { return <%=componentAccess%>; }
+<%
+if (name === 'crossTileID' || name === 'placedOrientation' || name === 'hidden') {
```
kkaefer: Let's add a comment here why these are the only properties that need a setter.

src/util/struct_array.js.ejs
```diff
@@ -26,7 +26,7 @@ for (const member of members) {
     }
 }
 
-if (includeStructAccessors) {
+if (includeStructAccessors && StructArrayClass !== 'GlyphOffsetArray' && StructArrayClass !== 'SymbolLineVertexArray') {
```
kkaefer: Let's add a comment as to why they are excluded.

PR #9367, opened 2020-03-03
Start quickly with Streets v11

1ec5:
#8056 upgraded most of the codebase to Streets v11, but the “Quickstart” section of the “Overview” page still referred to Streets v9.

/cc @mapbox/docs


katydecorah:
We should delete publisher-production branch on this repo since it no longer points to the website. This change needs to be made in https://github.com/mapbox/mapbox-gl-js-docs instead.

Issue #8020, opened 2019-03-11
`fill-pattern` ignores image's `pixelRatio`


flawyte:
**mapbox-gl-js version**: 0.53.1
**browser**: Chrome 72.0.3626.121

### Steps to Trigger Behavior

 1. Load an image using `loadImage()`
 2. Declare it using `addImage()`, passing `{ pixelRatio: 4 }` as the [3rd parameter](https://docs.mapbox.com/mapbox-gl-js/api/#map#addimage)
 3. Use this image in a `fill-pattern` property

### Link to Demonstration

https://jsbin.com/vevosisuyo/edit?html,js,output

### Expected Behavior

Pattern should be "scaled down" `pixelRatio` times.

### Actual Behavior

`pixelRatio` is ignored and the pattern is never scaled when repeated.

> I don't know if it's expected behavior or not, but the documentation says nothing about this specific case intentionally not being supported so I thought it might be a bug.

ansis:
Yep, this looks like a bug! Thanks for reporting and supplying the example

julians:
Is there any news on this?

karimnaaji:
I think this might also affect `line-pattern`, I discovered this issue while working on https://github.com/mapbox/mapbox-gl-js/pull/9266 and trying to use a pattern added with `addImage` with different pixel ratios.

karimnaaji:
Hey @julians @flawyte! Some updates on this; I found the root cause and will polish and send it out for review.
In the meantime, if you want to try it out and ensure that it fixes your use cases you can see it in this branch https://github.com/mapbox/mapbox-gl-js/tree/fix-8020

karimnaaji:
Closing as fixed by #9372 , available in next release.

PR #9372, opened 2020-03-04
Fix addImage API option for pixel ratio 

karimnaaji:
**Before**

On a 1x screen display, after loading an image of resolution of `64x64px`, result of a call to `map.addImage('pattern', image, { pixelRatio: 1 });`:

before-1x-cropped

On a 2x screen display, after loading an image of resolution of `128x128px`, result of a call to `map.addImage('pattern', image, { pixelRatio: 2 });`:

before-2x-cropped

**After**
On a 1x screen display, after loading an image of resolution of `64x64px`, result of a call to `map.addImage('pattern', image, { pixelRatio: 1 });`:

after-1x-cropped

On a 2x screen display, after loading an image of resolution of `128x128px`, result of a call to `map.addImage('pattern', image, { pixelRatio: 2 });`:

after-2x-cropped

_So what's wrong with these cats?_

The current behavior is only correct under certain circumstances, for example, when adding a xx image on a xx display. As long as the pixel ratio matches the resolution display, the behavior is correct.

Though, when adding assets through the public API, the user is able to provide as an option a user defined pixel ratio for the asset; the pixel ratio for this asset might be different from the currently used screen ratio, refer the [public API reference documentation](https://docs.mapbox.com/mapbox-gl-js/api/#map#addimage). Refer to #8020.

Some references taken from the render tests that are added in this PR and show the behavior on the cases that fail in the master branch:

**Before**
2x on 2x:
2x-on-2x-add-image-master
3x on 2x:
3x-on-2x-add-image-master
4x on 2x:
4x-on-2x-add-image-master

**After**
2x on 2x:
2x-on-2x-add-image
3x on 2x:
3x-on-2x-add-image
4x on 2x:
4x-on-2x-add-image

This PR fixes the behavior by always preferring the asset pixel ratio over the device pixel ratio and not assume that they would be representing the same, this ensures that the rendering is consistent even when the user forgets to provide a matching asset. For example, after loading an image of resolution of `128x128px`, the result of a call to `map.addImage('pattern', image, { pixelRatio: 2 });` over different screens:

**Before**

1x screen:
before-1x-2-cropped

2x screen:
before-2x-2-cropped

**After**

1x screen:
after-2x-2-cropped

2x screen:
after-1x-2-cropped

Does not supersedes https://github.com/mapbox/mapbox-gl-js/pull/9266 but fixes a few of the issues that were brought up on the PR.

## Launch Checklist

 - [x] @mapbox/gl-native this would require a native port
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix addImage API option for pixel ratio by adjusting *-pattern images based on the asset pixel ratio instead of the device pixel ratio</changelog>`


asheemmamoowala:
cc @mapbox/map-designers @mapbox/studio

dereklieu:
> preferring the asset pixel ratio over the device pixel ratio

Hey @karimnaaji can you clarify the difference between asset vs device pixel ratio?

karimnaaji:
@dereklieu The device pixel ratio is the pixel ratio incoming from the physical screen that you are using, we are querying that value from the browser which has ways to retrieve that on the hardware internally.

The asset pixel ratio is the pixel ratio that the asset has been designed for @1x if it's intended to be displayed on a 1x screen display, @2x if it's intended to be displayed on a 2x screen display, etc..

The asset pixel ratio is automatically retrieved from the tag that's associated with it when using the @ syntax along with the file name. Though, when adding assets with the runtime API there is an [option to provide the image pixel ratio](https://docs.mapbox.com/mapbox-gl-js/api/#map#addimage) but this option parameter ends up being unused down the line as we are only using the device pixel ratio, that's the case that this PR aims to fix.

Let me know if that makes sense or require further clarification!

dereklieu:
Ah, thanks for clearing that up.

> but this option parameter ends up being unused down the line as we are only using the device pixel ratio, that's the case that this PR aims to fix.

^ This was helpful for me to understand what's happening here. 

In Studio we've had occasional problems with using the runtime `addImage` or `updateImage` API when a user has zoomed their browser, which throws off the reported pixel ratio. This fix sounds like it would help mitigate some of those issues.

karimnaaji:
>In Studio we've had occasional problems with using the runtime addImage or updateImage API when a user has zoomed their browser, which throws off the reported pixel ratio. This fix sounds like it would help mitigate some of those issues.

I don't want to set false expectations as this might be a different problem, would you have a reference issue so I can confirm?

dereklieu:
The best reference issue is https://github.com/mapbox/studio/issues/12223 which is not very good, but broadly captures the problem and a few of the steps we've taken to mitigate. 

I'm not counting on this as a fix-all, I am just glad that the pixel ratio option in the runtime api is more trustworthy, as I think it will help us debug this issue down the line.

src/data/bucket/pattern_attributes.js
```diff
@@ -4,5 +4,7 @@ import {createLayout} from '../../util/struct_array';
 export default createLayout([
     // [tl.x, tl.y, br.x, br.y]
     {name: 'a_pattern_from', components: 4, type: 'Uint16'},
-    {name: 'a_pattern_to', components: 4, type: 'Uint16'}
+    {name: 'a_pattern_to', components: 4, type: 'Uint16'},
+    {name: 'a_pixel_ratio_from', components: 1, type: 'Uint8'},
+    {name: 'a_pixel_ratio_to', components: 1, type: 'Uint8'},
```
ansis: Can these two new attributes be combined into a single attribute? each attribute is aligned to a 4-byte boundary so putting them into one will save 4 bytes per vertex

Or maybe even packed into the existing attributes? I think it's probably *not* worth it at this point but it might be worth keeping in mind if we ever need to optimize this further.

karimnaaji: That's a good point! I was thinking about this but assumed it was not the common case. In my testing we didn't hit this code path in the default styles but when using data driven styling in expressions such as `"line-pattern": ["case", ["==", ["get", "class"], "street"], "generic_icon", "generic_metro" ]`.

I'm ok to give that an afterthought and profile more to see if the frequency at which we hit that justify this optimization.

chloekraw:
@karimnaaji one question: `addImage` is used for more than just patterns, and it looks like this fix applies to `addImage` broadly. Is that true?

Also, a small suggestion to adjust this changelog entry from

> `<changelog>Fix addImage API option for pixel ratio by adjusting *-pattern images based on the asset pixel ratio instead of the device pixel ratio</changelog>`

to 

> `<changelog>Correctly adjust images added with `addImage.pixelRatio` by the asset pixel ratio, not the device pixel ratio</changelog>`

karimnaaji:
@chloekraw `icon-image` is already working and unaffected by this PR, it was the other possible use case from here: https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/map.js#L1464_L1468. Updating with your suggestion:
><changelog>Correctly adjust patterns added with addImage(id, image, pixelRatio) by the asset pixel ratio, not the device pixel ratio</changelog>

src/data/bucket/pattern_attributes.js
karimnaaji: Captured it for follow up in https://github.com/mapbox/mapbox-gl-js/issues/9407.

PR #9375, opened 2020-03-04
Don't reload the entire font range when a glyph is missing in a font

### LINKED ISSUE #8027: GlyphManager continually re-requests ranges with missing glyphs
Consider the `format` expression demo:

```
map.setLayoutProperty(layer, 'text-field', ['format',
  ['get', 'name_en'], { 'font-scale': 1.2 },
  '\n', {},
  ['get', 'name'], {
    'font-scale': 0.8,
    'text-font': ['literal', [ 'DIN Offc Pro Italic', 'Arial Unicode MS Regular' ]]
  }]);
```

`\n` isn't in `0-255.pbf` for this fontstack, which is fine, because it doesn't need to draw anything, and the actual line breaking is handled separately in the shaping. But whenever a tile gets reloaded, `getGlyphs` won't find the glyph here:

https://github.com/mapbox/mapbox-gl-js/blob/c189971a3a49df118047dba442c7287e4ddde141/src/render/glyph_manager.js#L59-L63

So it will re-request the entire range. This probably isn't a _huge_ deal because the data should be cached, and it's not super-expensive to reload the PBF... but it's wasteful and it looks ugly in the network monitor (it grabbed my attention because if you have the cache disabled it looks like a major source of data traffic in maps that reload tiles frequently).

I don't think gl-native has the same issue since its glyph manager checks for already-loaded _ranges_ instead of _glyphs_:

https://github.com/mapbox/mapbox-gl-native/blob/096dea16e31c6e0d91eba0cab1ab2efc5d6ae853/src/mbgl/text/glyph_manager.cpp#L45-L52

cc @ryanhamley @asheemmamoowala

oterral:
Fix #8027 

Don't reload the entire font range when a glyph is missing in a font.

## Launch Checklist

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Don't reload the entire font range when a glyph is missing in a font.
</changelog>`

### DIFF
--- a/test/unit/render/glyph_manager.test.js
+++ b/test/unit/render/glyph_manager.test.js
@@ -1,4 +1,4 @@
-import {test} from '../../util/test';
+import {test, only} from '../../util/test';
 import parseGlyphPBF from '../../../src/style/parse_glyph_pbf';
 import GlyphManager from '../../../src/render/glyph_manager';
 import fs from 'fs';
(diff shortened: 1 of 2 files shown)


test/unit/render/glyph_manager.test.js
```diff
@@ -28,6 +28,36 @@ test('GlyphManager requests 0-255 PBF', (t) => {
     });
 });
 
+test('GlyphManager doesn\'t request twice 0-255 PBF if a glyph is missing', (t) => {
```
sgolbabaei: Thanks a lot for your pull request. This change looks great. Just a small suggestion here, since this test and `GlyphManager requests 0-255 PBF` test are using a lot of similar code, I would suggest using helper functions in order to avoid repeating the same code.

oterral: Done

sgolbabaei:
Look good to me. Thanks for the changes.

PR #9377, opened 2020-03-04
Prevent creation of unused depth renderbuffer attachments

karimnaaji:
Prevent creation of unused GPU resources for depth renderbuffer attachments:
- painter: should save width * height * 2 bytes of depth renderbuffer memory 
- heatmap: should save (width * height) / 4 * 2 bytes of depth renderbuffer memory

`<changelog>Slightly improve GPU memory footprint</changelog>`

### DIFF
--- a/src/gl/framebuffer.js
+++ b/src/gl/framebuffer.js
@@ -1,5 +1,6 @@
 // @flow
 import {ColorAttachment, DepthAttachment} from './value';
+import assert from 'assert';
 
 import type Context from './context';
 
@@ -11,7 +12,7 @@ class Framebuffer {
     colorAttachment: ColorAttachment;
     depthAttachment: DepthAttachment;
 
-    constructor(context: Context, width: number, height: number) {
+    constructor(context: Context, width: number, height: number, hasDepth: boolean) {
         this.context = context;
         this.width = width;
         this.height = height;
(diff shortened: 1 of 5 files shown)


asheemmamoowala:
@karimnaaji These types of changes don't usually warrant being listed in the Changelog as there is not developer facing impact.

karimnaaji:
@asheemmamoowala should we compile a changelog policy? I personally think performance improvements should be added in any changelog as they show the effort to keep a healthy library, that might be reconsidered if it becomes cluttered or hide more important information.

src/gl/framebuffer.js
```diff
@@ -11,15 +12,19 @@ class Framebuffer {
...
         const gl = context.gl;
         const fbo = this.framebuffer = gl.createFramebuffer();
 
         this.colorAttachment = new ColorAttachment(context, fbo);
-        this.depthAttachment = new DepthAttachment(context, fbo);
+        if (hasDepth) {
+            this.depthAttachment = new DepthAttachment(context, fbo);
+        }
+        const status = gl.checkFramebufferStatus(gl.FRAMEBUFFER);
+        assert(status === gl.FRAMEBUFFER_COMPLETE);
```
mourner: Merging into one line that gets stripped so that we don't get framebuffer status wastefully in production builds.

```suggestion
        assert(gl.checkFramebufferStatus(gl.FRAMEBUFFER) === gl.FRAMEBUFFER_COMPLETE);
```

Issue #9368, opened 2020-03-03
Enable `symbol-sort-key` when text/icons on a layer have opposing overlap rules


chloekraw:
When `icon-allow-overlap: true` and `text-allow-overlap: false` or vice-versa for a symbol layer, there is no way to sort icons + text consistently with [`symbol-sort-key`](https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-symbol-symbol-sort-key).

Sort key determines draw order in the renderer. Features with lower sort keys are drawn and placed first. 
- When overlap is enabled, features with higher sort keys can be drawn and placed over previous features, so features with higher sort keys have higher priority. 
- When overlap is disabled, features with higher sort keys cannot be placed over previous features because of the way collision detection works. Therefore, during placement, features with _lower_ sort keys have higher priority.

This behavior has always existed in the renderer and it won't be straightforward to change.

--- 

<img width="784" alt="Screen Shot 2020-03-02 at 11 05 49 PM" src="https://user-images.githubusercontent.com/34117059/75751037-60824f00-5cda-11ea-8866-6260a207b9c8.png">

ryanbaumann:
As a developer, I want to create a `rank` value for features in a symbol layer to control where they are drawn and placed.  The challenging part is when I have to rank features differently and/or inversely when I toggle collision detection between symbol features on/off.

A simple solution for a developer would be to have an "always high values" or "always low values" are prioritized using `symbol-sort-key`, and not needing to change the ascending vs. descending order when I toggle collision detection on or off.

PR #9379, opened 2020-03-05
[docs] clarify differences in `symbol-sort-key` with and without overlap

chloekraw:
We've gotten some feedback that https://github.com/mapbox/mapbox-gl-js/issues/9368 is not very obvious from the existing `symbol-sort-key` docs, particularly if you are not that familiar with the renderer. This PR makes it more obvious that the behavior of the sort key is different with and without `*-allow-overlap` by making the two sentences more parallel.

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -1025,7 +1025,7 @@
     },
     "symbol-sort-key": {
       "type": "number",
-      "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above features with a lower sort key when they overlap. Features with a lower sort key will have priority over other features when doing placement.",
+      "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above other features when they overlap. Features with a lower sort key will have priority over other features during placement when they do not overlap.",
       "sdk-support": {
         "basic functionality": {
           "js": "0.53.0",
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -1025,7 +1025,7 @@
     },
     "symbol-sort-key": {
       "type": "number",
-      "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above other features when they overlap. Features with a lower sort key will have priority over other features during placement when they do not overlap.",
+      "doc": "Sorts features in ascending order based on this value. When `*-allow-overlap` is enabled for symbols, features with a higher sort key will appear above other features. When overlap is disabled, features with a lower sort key will have priority over other features during placement.",
       "sdk-support": {
         "basic functionality": {
           "js": "0.53.0",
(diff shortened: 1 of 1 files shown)


src/style-spec/reference/v8.json
```diff
@@ -1025,7 +1025,7 @@
     },
     "symbol-sort-key": {
       "type": "number",
-      "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above features with a lower sort key when they overlap. Features with a lower sort key will have priority over other features when doing placement.",
+      "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above other features when they overlap. Features with a lower sort key will have priority over other features during placement when they do not overlap.",
```
ryanbaumann: I recommend that we make the doc very clear with an example.

For example, when symbol collision detection is enabled, features with a lower symbol-sort-key will be placed before features with a higher symbol-sort key.  When symbol collision detection is disabled, features with a higher symbol-sort-key will be drawn on top of features witha  lower symbol-sort-key if they overlap.

chloekraw: Hm, your example is not an example as much as the only way which this distinction can be triggered. “when features overlap” is intended to mean “when you have them overlap with `*-allow-overlap”. That’s the main way to “enable/disable symbol collision detection.” I think leading with “when symbol collision detection is enabled/disabled...” is more confusing when we don’t have a property named `”allow-collision-detection”`, but I’ll add a commit that borrows some of this phrasing.

chloekraw:
@ansis, what do you think about this change? I think it increases clarity but I'm not sure if it keeps with precedent and is in the voice that we use for style spec docs.

chloekraw:
My alternative proposal is to only focus on what happens to features with higher sort keys:

> Sorts features in ascending order based on this value. Features are drawn and placed in sort key order. When `*-allow-overlap` is enabled, features with higher sort keys will appear above earlier-placed features. When `*-allow-overlap` is disabled, features with higher sort keys may be hidden by collision detection with earlier-placed features.

@ryanbaumann @mtirwin which one do you like better?

cc @adrianababakanian @malwoodsantoro @samfader

ryanbaumann:
@chloekraw How's this tweak as a compromise?

> Sorts features in ascending order based on this value. Features are drawn and placed **based on** sort key order. When **`text-allow-overlap` or `icon-allow-overlap`** are enabled, features with higher sort keys will **be drawn on top of features with lower sort order keys when they overlap**. When **`text-allow-overlap` or `icon-allow-overlap`** disabled, features with **lower** sort keys will be placed first, meaning they will **appear first when they overlap with a feature with a high sort key)**.

chloekraw:
@ryanbaumann thanks for the feedback! I'll push a commit with most of these suggestions, but your last sentence is a bit misleading.

Features with a lower sort key are **always** placed first. That's why they are underneath features with a higher sort key when overlap is enabled. 

`symbol-sort-key` does not change the behavior of the renderer when overlap is enabled or disabled. The renderer behaves in exactly the same way in both cases. It just so happens that the same behavior does not match what humans expect when it comes to prioritizing features in both cases.

PR #9385, opened 2020-03-06
Fix corner case with dash-array 0-length collapsing

karimnaaji:
Fix corner case on an exception discovered and reported by @tristen and add some more test coverage around it.

`<changelog>Prevent exception resulting from 'line-dash-array' of empty length</changelog>`


karimnaaji:
This fix does not apply for gl-native, a similar check was already added while porting.

PR #9387, opened 2020-03-06
[render tests] Adjust tile mode render tests

pozdnyakov:
- increased tolerance for `text-variable-anchor/left-top-right-buttom-offset-tile-map-mode`
- hide collision data for `tile-mode/streets-v11`


pozdnyakov:
> left-top-right-bottom-offset-tile-map-mode

will fix with the following pr

PR #9391, opened 2020-03-07
Add methods to mapboxgl namespace to allow preloading of workers.

arindam1993:
Some profiling revealed that WebWorker initialization by the browser accounts for a large portion of our map load time.

This PR adds some API methods to the `mapboxgl` namespace that lets app developers using our SDK work around this problem by allowing them manual control of the global WebWorker pool.
This should be especially beneficial for single-page-apps that don't have the GL map on their first-page, since they can preload workers in the background before the map is even needed.

## How much faster is it?

The test page I used to profile the difference repeatedly destroyed and re-created the map and measure load times.
```
var loadTimes = window.loadTimes = [];
var loadCtr = 0;
var map;
var loadMap = function() {
    if (map) {
        map.remove();
    }

    var start = performance.now();
    map = new mapboxgl.Map({
        container: 'map',
        zoom: 12.5,
        center: [-77.01866, 38.888],
        hash: true
    });

    map.on('load', function() {
        loadCtr++;
        loadTimes.push(performance.now() - start);

        if (loadCtr < 20) {
            setTimeout(loadMap, 100);
        }
    });
};
loadMap();
```

The results are pretty good outstanding, with subsequent map loads speeding up upto 2x when the Map instance does not destroy the workers with it.
Screen Shot 2020-03-06 at 5 18 37 PM

This is true when using the split `csp` bundle for web-workers as well.

## How does it work?

Developers can now use `mapboxgl.preloadWorkerPool()` to start the workers whenever they want. This method has no effect if the workers are already started. For maximum benefit developers can call this in `<head>` to start loading workers as soon as the page loads. 
By preloading workers, the onus of destroying them when no-longer needed is also on the developer, the `Map` instance will no longer teardown the workers when its being removed. This can be achieved by calling `mapboxgl.removePreloadedWorkerPool()` explicitly.

@mapbox/studio  Y'all might like this. Since the first page of studio is not a map.
## Launch Checklist

 - [x] post benchmark scores
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Add mapboxgl.preloadWorkerPool() and mapboxgl.removePreloadedWorkerPool() methods to allow developers to optimize load times for their maps.</changelog>`


src/util/global_worker_pool.js
```diff
@@ -15,3 +15,24 @@ export default function getGlobalWorkerPool () {
...
+
+export function preloadWorkerPool() {
+    const workerPool = getGlobalWorkerPool();
+    workerPool.acquire(PRELOAD_POOL_ID);
+}
+
+export function removePreloadedWorkerPool() {
+    const pool = globalWorkerPool;
+    if (!pool) {
+        console.warn('Attempted to remove preloaded worker pool before creating one, skipping. Use mapboxggl.preloadWorkerPool() to preload a global worker pool');
```
ryanhamley: typo: `mapboxggl`

src/index.js
```diff
@@ -46,6 +47,8 @@ const exported = {
     MercatorCoordinate,
     Evented,
     config,
+    preloadWorkerPool,
+    removePreloadedWorkerPool,
```
ryanhamley: We should add documentation for these methods. They'll be helpful for a lot of folks, but they could also be tricky if we don't give some guidance. For example, we'd probably recommend calling `map.remove()` when unmounting a React component, but you probably shouldn't call `mapboxgl.removePreloadedWorkerPool()` at that point in the lifecycle.

src/util/global_worker_pool.js
kkaefer: I don't think we should have a warning message here; this should just be a no-op.

src/util/global_worker_pool.js
```diff
@@ -15,3 +15,24 @@ export default function getGlobalWorkerPool () {
...
+        return;
+    }
+
+    // Remove the pool only if all maps that referenced the preloaded global worker pool have been removed.
+    if (pool.isPreloaded() && pool.numActive() === 1) {
```
kkaefer: Instead of relying on the user to remove worker pools, we could flag the pool for deletion (refcount style) and delete it once the last map instance disappears. It looks like the workers are automatically terminated if we release `PRELOAD_POOL_ID`.

src/index.js
kkaefer: I think we should reconsider this name. "worker pool" isn't the first them GL JS devs are going to look for when they want to speed up load time in their web app, and also leaks our internal architecture. Instead, I think we could name it `initialize()`, `prewarm()`, or something similar.

src/util/global_worker_pool.js
arindam1993: We do that already, but that means that a developer wont be able to cache workers between map tear down and re-initialize, which is where most of the performance gain is.
Eg. In studio go from style list -> a style  -> back to style list -> to a different style.

src/index.js
```diff
@@ -47,8 +47,32 @@ const exported = {
     MercatorCoordinate,
     Evented,
     config,
-    preloadWorkerPool,
-    removePreloadedWorkerPool,
+    /**
+     * Create's shared resources, primarily WebWorkers, for the Map ahead of time, this significantly improves load times in certain situations.
```
kkaefer: Let's remove the plural apostrophe

src/index.js
```diff
@@ -47,8 +47,32 @@ const exported = {
...
+     * `mapboxgl.workerUrl` and `mapboxgl.workerCount`, if being used, must be set before `prewarm()` is called in order to have an effect.
+     *
+     * By default, the lifecycle these resources is managed automatically, and they are lazily initialized when a Map is first created.
+     * By invoking `prewarm()`, these resources can be created ahead of time, and will not be cleared when Map is removed.
+     * This allows them to be re-used by new instances of Map that are created later. These resources can be manually cleared
+     * by calling `mapboxgl.clearPrewarmedResources()`, though this will typically not be necessary.
+     *
+     * This is primarily useful when using GL-JS maps in a single page app, wherein a user would naviate
+     * bewteen various views that can cause Map instances to constantly be created and destroyed.
```
kkaefer:
```suggestion
     * Initializes resources like WebWorkers that can be shared across maps to lower load
     * times in some situations. `mapboxgl.workerUrl` and `mapboxgl.workerCount`, if being
     * used, must be set before `prewarm()` is called to have an effect.
     *
     * By default, the lifecycle of these resources is managed automatically, and they are
     * lazily initialized when a Map is first created. By invoking `prewarm()`, these
     * resources will be created ahead of time, and will not be cleared when the last Map
     * is removed from the page. This allows them to be re-used by new Map instances that
     * are created later. They can be manually cleared by calling
     * `mapboxgl.clearPrewarmedResources()`. This is only necessary if your web page remains
     * active but stops using maps altogether.
     *
     * This is primarily useful when using GL-JS maps in a single page app, wherein a user
     * would navigate between various views that can cause Map instances to constantly be
     * created and destroyed.
```

src/index.js
```diff
@@ -47,8 +47,32 @@ const exported = {
...
+     *
+     * @function prewarm
+     * @example
+     * mapboxgl.prewarm()
+     */
+    prewarm,
+    /**
+     * Clears up resources that have been previously created by `mapboxgl.prewarm()`.
+     * This should typically not be necessary, and only be used if you expect the user of your app to not return to a Map view at any point in your application.
```
kkaefer:
```suggestion
     * Clears up resources that have previously been created by `mapboxgl.prewarm()`.
     * Note that this is typically not necessary. You should only call this function
     * if you expect the user of your app to not return to a Map view at any point
     * in your application.
```

PR #9409, opened 2020-03-11
Fix for IE11 not supporting startsWith

karimnaaji:
Replace `startsWith` with `substr`


arindam1993:
LGTM, thanks for the quick fixes @karimnaaji  🎉

Issue #9413, opened 2020-03-12
Add scale to default Marker options


adrianababakanian:
## Motivation

Some developers may want to use the default `Marker` `el` (a light-blue, droplet-shaped SVG) as-is in terms of visual appearance, but with a modified size. As documented in the API reference [here](https://docs.mapbox.com/mapbox-gl-js/api/#marker), there are currently no sizing-related `options` in the `Marker`'s `options` parameter.

## Design Alternatives

[This example](https://docs.mapbox.com/mapbox-gl-js/example/custom-marker-icons/) in the documentation shows how to add custom icons with markers by defining a DOM node for each GeoJSON feature (`var el = document.createElement('div')`), but it would be quite a bit of work to re-create the default marker just to modify its size. The fact that the marker is created with SVG components introduces an additional complication when left to the end-user, since scaling SVGs is not entirely intuitive ([this reference](https://css-tricks.com/scale-svg/) explains the concept well, in my opinion).

## Design

For consistency with how similar CSS terminology would affect the SVG marker added to the map, I recommend that we add `scale` to the `options` parameter rather than `width` or `height`. A `scale` is admittedly less flexible than being able to explicitly specify a `width` and `height`, but ultimately benefits the end user so that they don't have to strategize the dimensions to prevent distortion.

**Demonstrating how `scale` is more intuitive than `width` or `height` for SVGs:**

![scale-width-height](https://user-images.githubusercontent.com/15935667/76490682-058edd00-63e9-11ea-8b44-7e416845ecf5.png)

### Mock-Up

    /* Display a Marker at 2x the default size: width="27px", height="41px" */
    var marker = new mapboxgl.Marker({
    	scale: 2 // a number greater than or equal to zero
    })

### Concepts

As mentioned in the Design section, I think that `scale` is an appropriate terminology to be introduced to the `Marker`'s `options` parameter.

### Implementation

Minimal changes are required in the Mapbox GL ecosystem. `/src/ui/marker.js` is the most relevant file. A `_scale: number` class variable should be added, and the `svg` object representing the marker could be modified accordingly by:

    svg.setAttributeNS(null, 'transform', `scale(${this._scale})`);

stevage:
Just a comment that the terminology used in the `symbol` layer type is `icon-size`, which behaves exactly like a scale. You can't be consistent with the whole world but maybe it makes more sense to be internally consistent (within mapbox-gl) rather than externally consistent (with CSS)?

[...]

PR #9414, opened 2020-03-12
Add scale to Marker options

adrianababakanian:
## Overview
Closes https://github.com/mapbox/mapbox-gl-js/issues/9413.

Allows users to customize the size of the default `Marker` by scaling the relevant SVG.

## Launch Checklist

 - [ ] post benchmark scores
 - [x] ~tagged `@mapbox/studio` and/or `@mapbox/map-design-team` if this PR includes style spec or visual changes~
 - [x] ~tagged `@mapbox/gl-native` if this PR includes shader changes or needs a native port~
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog></changelog>`

**Visuals:**
Screen Shot 2020-03-11 at 10 20 06 PM

**Changes to API:**
Adds a new option for the default marker: `scale`.

**Changelog:**
Add `scale` option to markers, to support custom sizing when using the default `Marker` SVG element.

### DIFF
--- a/debug/markers.html
+++ b/debug/markers.html
@@ -45,8 +45,11 @@ const spinningMarkers = [];
         var g = Math.round(Math.random() * 255);
         var b = Math.round(Math.random() * 255);
 
+        var sc = Math.random(0, 25);
+
         var marker = new mapboxgl.Marker({
             color: `rgb(${r}, ${g}, ${b})`,
+            scale: sc,
             draggable: true,
             rotationAlignment,
             pitchAlignment
--- a/debug/markers.html
+++ b/debug/markers.html
@@ -45,7 +45,7 @@ const spinningMarkers = [];
         var g = Math.round(Math.random() * 255);
         var b = Math.round(Math.random() * 255);
 
-        var sc = Math.random(0, 25);
+        var sc = Math.random() * 2.5 + 0.5;
 
         var marker = new mapboxgl.Marker({
             color: `rgb(${r}, ${g}, ${b})`,
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -101,7 +101,7 @@ export default class Marker extends Evented {
             svg.setAttributeNS(null, 'display', 'block');
             svg.setAttributeNS(null, 'height', '41px');
             svg.setAttributeNS(null, 'width', '27px');
-            svg.setAttributeNS(null, 'viewBox', '0 0 27 41');
+            svg.setAttributeNS(null, 'viewBox', `0 0 27 41`);
 
             const markerLarge = DOM.createNS('http://www.w3.org/2000/svg', 'g');
             markerLarge.setAttributeNS(null, 'stroke', 'none');
@@ -184,7 +184,9 @@ export default class Marker extends Evented {
             page1.appendChild(circleContainer);
 
             svg.appendChild(page1);
-            svg.setAttributeNS(null, 'transform', `scale(${this._scale})`);
+
+            svg.setAttributeNS(null, 'width', `${27 * this._scale}px`);
+            svg.setAttributeNS(null, 'height', `${41 * this._scale}px`);
 
             this._element.appendChild(svg);
 
@@ -397,6 +399,12 @@ export default class Marker extends Evented {
         }
 
         this._pos = this._map.project(this._lngLat)._add(this._offset);
+
+        const svg = this._element.firstChild;
+        if (svg) {
+            svg.setAttributeNS(null, 'width', `${27 * this._scale}px`);
+            svg.setAttributeNS(null, 'height', `${41 * this._scale}px`);
+        }
 
         let rotation = "";
         if (this._rotationAlignment === "viewport" || this._rotationAlignment === "auto") {
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -34,7 +34,7 @@ type Options = {
  *   Options are `'center'`, `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`, `'top-right'`, `'bottom-left'`, and `'bottom-right'`.
  * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
- * @param {number} [options.scale=1] The scaled to use for the default marker if options.element is not provided. The default is 1.
+ * @param {number} [options.scale=1] The scale to use for the default marker if options.element is not provided. The default is 1.
  * @param {boolean} [options.draggable=false] A boolean indicating whether or not a marker is able to be dragged to a new position on the map.
  * @param {number} [options.rotation=0] The rotation angle of the marker in degrees, relative to its respective `rotationAlignment` setting. A positive value will rotate the marker clockwise.
  * @param {string} [options.pitchAlignment='auto'] `map` aligns the `Marker` to the plane of the map. `viewport` aligns the `Marker` to the plane of the viewport. `auto` automatically matches the value of `rotationAlignment`.
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -19,6 +19,7 @@ type Options = {
     offset?: PointLike,
     anchor?: Anchor,
     color?: string,
+    scale?: number,
     draggable?: boolean,
     rotation?: number,
     rotationAlignment?: string,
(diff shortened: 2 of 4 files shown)


src/ui/marker.js
```diff
@@ -33,6 +34,7 @@ type Options = {
  *   Options are `'center'`, `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`, `'top-right'`, `'bottom-left'`, and `'bottom-right'`.
  * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
+ * @param {number} [options.scale=1] The scale to use for the default marker if options.element is not provided. The default is 1.
```
andrewharvey: "The default is 1." Minor nit pick, that's okay, but for the most part in the rest of the docs the default value is omitted from the description and left to documentation.js to format the default from the JSDoc.

I know I put the default in the color option above, though since not everyone speaks hex codes it's probably justified in the color option.

andrewharvey:
This looks great :+1: from me.

src/ui/marker.js
```diff
@@ -180,6 +184,7 @@ export default class Marker extends Evented {
             page1.appendChild(circleContainer);
 
             svg.appendChild(page1);
+            svg.setAttributeNS(null, 'transform', `scale(${this._scale})`);
```
asheemmamoowala: `setAttributeNS` has unknown support in IE11, it would be good to verify if it works at the root SVG element or switch to using `setAttribute` if viable.

src/ui/marker.js
adrianababakanian: Great point, I'll remove that sentence and leave it up to documentation.js to format the default.

src/ui/marker.js
adrianababakanian: Makes sense @asheemmamoowala -- I went with `setAttributeNS` rather than `setAttribute` for consistency with the usage [here](https://github.com/mapbox/mapbox-gl-js/blob/b238b15f64e00540c29d0ec3d2065861aa90d1f9/src/ui/marker.js#L95-L100). Just to be sure, would you recommend testing with IE11 or switching to `setAttribute` before moving forward?

test/unit/ui/marker.test.js
```diff
@@ -27,6 +27,12 @@ test('Marker uses a default marker element with custom color', (t) => {
     t.end();
 });
 
+test('Marker uses a default marker element with custom scale', (t) => {
+    const marker = new Marker({scale: 0.8});
+    t.ok(marker.getElement().innerHTML.includes('0.8'));
```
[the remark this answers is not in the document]
adrianababakanian: Thanks @ryanhamley , I updated to test for correct scaled dimensions both when the `scale` is less than 1 and greater than 1.

adrianababakanian:
Thanks again for the review @ryanhamley @asheemmamoowala ! This is ready for another review. A couple of questions I would appreciate your opinion on:

* What are your thoughts on changing the `scale` terminology to `size`, as suggested in https://github.com/mapbox/mapbox-gl-js/issues/9413#issuecomment-602121472? This would be a bit more consistent with how `icon-size` is used in the style specification (i.e. "units in factor of the original icon size"), though at the same time I find `scale` to be a bit more intuitive.
* I added a `_defaultSVG` class variable to `Marker`. But, it seems like Flow doesn't include an `SVGElement` type out of the box, so I specified `HTMLElement` as the `_defaultSVG`'s type in order to be able to use `Element.setAttributeNS()` in `Marker#_update()`. This doesn't seem like the greatest decision -- do you have any suggestions for a better alternative?

src/ui/marker.js
```diff
@@ -548,6 +562,25 @@ export default class Marker extends Evented {
...
+    }
+
+    /**
+     * Returns the current `scale` property of the default marker.
+     * @returns {number} The current scale of the marker.
+     */
+    getScale() {
+        return this._scale;
+    }
+
```
ryanhamley, line 583: So....I know that I suggested these originally, but I misunderstood the PR and now I think it would probably be best to remove these methods. Sorry :(

My reasoning now is that this PR only applies to the default SVG marker. I don't think there's any way to universally resize HTML elements regardless of type, so if a user supplies their own marker element, these methods won't work. Given that, we should avoid adding methods to the API that only work some of the time. It might be best to rely only on the scale option.

debug/markers.html
```diff
@@ -45,8 +45,11 @@
         var g = Math.round(Math.random() * 255);
         var b = Math.round(Math.random() * 255);
 
+        var sc = Math.random(0, 25);
```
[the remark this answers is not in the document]
adrianababakanian: Thanks for catching this! Agreed, those bounds sounds more reasonable.

src/ui/marker.js
```diff
@@ -181,7 +186,11 @@ export default class Marker extends Evented {
 
             svg.appendChild(page1);
 
+            svg.setAttributeNS(null, 'width', `${27 * this._scale}px`);
+            svg.setAttributeNS(null, 'height', `${41 * this._scale}px`);
```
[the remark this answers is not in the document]
adrianababakanian: Makes sense -- I actually did add `defaultHeight` and `defaultWidth` variables in an earlier version and then went back to using the numbers directly, hence the residual backticks in 105 which I missed when changing back to quotes. I'll pull the default values back in to variables!

src/ui/marker.js
adrianababakanian: No worries, that reasoning sounds good to me! I do understand why a developer using these methods with a custom HTML element for their marker would have a strange experience. Even though calling `Marker#setScale` would have no effect if `this._defaultMarker` is `false`, providing methods which have an effect conditioned on the value of a parameter is not ideal.

andrewharvey: I can see both sides here, one use case for changing the scale is having a marker made bigger when you select it, however I can see how since this only applies to the default marker it would be confusing to have a setScale which does nothing.

To still support this make it larger when selected use case, I do lean on the side that we should support get/setScale, with a note in the JS doc that it only applies when using the default marker, and a console warning if used on a non-default marker?

adrianababakanian:
Thank you for the continued feedback @ryanhamley , it's much appreciated. This is ready for another round of eyes when you have the chance.

src/ui/marker.js
```diff
@@ -393,6 +402,11 @@ export default class Marker extends Evented {
 
         this._pos = this._map.project(this._lngLat)._add(this._offset);
 
+        if (this._defaultMarker) {
+            this._defaultSVG.setAttributeNS(null, 'width', `${27 * this._scale}px`);
+            this._defaultSVG.setAttributeNS(null, 'height', `${41 * this._scale}px`);
```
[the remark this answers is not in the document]
adrianababakanian: I believe that this addition to the `_update` method is not necessary without the getter + setter, so I've removed it in the most recent commit.

src/ui/marker.js
adrianababakanian: I do also see use cases where the getter + setter for `scale` would be useful, and I'd be open to adding the methods back. It is a bit limiting to only be able to specify marker sizing at creation. I do understand how these methods would be an "exception to the rule," however, since currently there are no other user interface components in the API with similarly "conditional" instance members.

If `Marker#getScale` and `Marker#setScale` were included, it would probably make sense to add `Marker#getColor` and `Marker#setColor` too, since `scale` and `color` are very similar. Though, that is outside the scope of this PR.

src/ui/marker.js
```diff
@@ -33,6 +34,7 @@ type Options = {
  *   Options are `'center'`, `'top'`, `'bottom'`, `'left'`, `'right'`, `'top-left'`, `'top-right'`, `'bottom-left'`, and `'bottom-right'`.
  * @param {PointLike} [options.offset] The offset in pixels as a {@link PointLike} object to apply relative to the element's center. Negatives indicate left and up.
  * @param {string} [options.color='#3FB1CE'] The color to use for the default marker if options.element is not provided. The default is light blue.
+ * @param {number} [options.scale=1] The scale to use for the default marker if options.element is not provided.
```
ryanhamley, line 37: I think it might be good to specify that the default scale corresponds to a height of 41px and a width of 27px. That would give developers the information up front so they don't have to dig into the code to figure out what scale factor they want or just blindly guess.

adrianababakanian: Great point. I'll add that to the description, and then go ahead and merge.

PR #9426, opened 2020-03-17
update pr template team tags

chloekraw:
per chat with @riastrad @brsbl, we want to make sure @mapbox/static-apis are aware of style spec changes coming down the pipeline that may affect their product.

I also validated all members of @mapbox/studio are members of @mapbox/map-design-team; there's no need to tag both.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,7 +8,7 @@
  - [ ] document any changes to public APIs
  - [ ] post benchmark scores
  - [ ] manually test the debug page
- - [ ] tagged `@mapbox/map-design-team` `@mapbox/gl-native` `@mapbox/static-apis` if this PR includes style spec API or visual changes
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,7 +8,7 @@
  - [ ] document any changes to public APIs
  - [ ] post benchmark scores
  - [ ] manually test the debug page
- - [ ] tagged `@mapbox/studio` and/or `@mapbox/map-design-team` if this PR includes style spec or visual changes
+ - [ ] tagged `@mapbox/map-design-team` `@mapbox/gl-native` `@mapbox/static-apis` if this PR includes style spec API or visual changes
```
mourner: `@mapbox/gl-native` is listed below so it shouldn't be included here, right?

chloekraw: sure, makes sense, thanks for the catch - it's unlikely this box would be checked and not the next one (i.e. style spec API or visual change that does not require a native port / shader change)

PR #9437, opened 2020-03-19
add an extra check for paint member of style layer

mannnick24:
We are getting a browser dependent, intermittent error from fill_style_layer.js:45
the query radius function expects paint memeber to be truthy and its not
so we have a TypeError: this.paint is undefined

I've been trying for a couple of days to reproduce this in a fiddle and I can't (yet)

### DIFF
--- a/src/source/tile.js
+++ b/src/source/tile.js
@@ -404,9 +404,9 @@ class Tile {
             if (!sourceLayer || !sourceLayerStates || Object.keys(sourceLayerStates).length === 0) continue;
 
             bucket.update(sourceLayerStates, sourceLayer, this.imageAtlas && this.imageAtlas.patternPositions || {});
-
-            if (painter && painter.style && painter.style.getLayer(id).paint) {
-                this.queryPadding = Math.max(this.queryPadding, painter.style.getLayer(id).queryRadius(bucket));
+            const layer = painter.style.getLayer(id);
+            if (painter && painter.style && layer.paint) {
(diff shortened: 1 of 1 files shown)


src/source/tile.js
```diff
@@ -404,7 +404,8 @@ class Tile {
             if (!sourceLayer || !sourceLayerStates || Object.keys(sourceLayerStates).length === 0) continue;
 
             bucket.update(sourceLayerStates, sourceLayer, this.imageAtlas && this.imageAtlas.patternPositions || {});
-            if (painter && painter.style) {
+
+            if (painter && painter.style && painter.style.getLayer(id).paint) {
                 this.queryPadding = Math.max(this.queryPadding, painter.style.getLayer(id).queryRadius(bucket));
```
mourner:
```suggestion
            const layer = painter.style.getLayer(id);
            if (painter && painter.style && layer.paint) {
                this.queryPadding = Math.max(this.queryPadding, layer.queryRadius(bucket));
```

fminuti:
Sorry, but what's the meaning of first use "painter.style" and THEN check it (inside the IF)?

mannnick24:
@fminuti quite right... checking painter && painter.style again is not needed, I'd be hesitant to change someone else's codebase though :-)
I'd probably get the layer = painter.style.getLayer(id) at the top and check its truthy instead of the has

mourner:
@fminuti good catch! My bad. Submitted #9472 to follow up.

Issue #9442, opened 2020-03-20
`within` expression returns incorrect results for polygons crossing the anti-meridian


ryanhamley:
**mapbox-gl-js version**: v1.9.0

**browser**: any

### Steps to Trigger Behavior

 1. create a polygon that stretches across the anti-meridian (international date line)
 2. create a point with a wrapped longitude coordinate inside the polygon (e.g. -185)
 3. set a property of the point to use the `within` expression
4. observe that the `within` expression returns `false` when it should return `true`

### Link to Demonstration
https://jsfiddle.net/rsh412/r0ycv14w/14/
<img width="598" alt="Screen Shot 2020-03-20 at 12 40 05 PM" src="https://user-images.githubusercontent.com/4523080/77200521-f6bebf00-6aa7-11ea-8466-123292e21b84.png">

### Expected Behavior
`within` recognizes wrapped coordinates and returns the proper boolean result for the feature

### Actual Behavior
`within` does not recognize wrapped coordinates

@zmiao is working on a fix in https://github.com/mapbox/mapbox-gl-js/pull/9440

PR #9444, opened 2020-03-20
v1.9.0

ryanhamley:


CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
...
 
 ### ✨ Features
-* Add `within` expression for testing whether an evaluated feature lies within a given GeoJSON object ([#9352](https://github.com/mapbox/mapbox-gl-js/pull/9352)). For example:<br>
+* Add `within` expression for testing whether an evaluated feature lies within a given GeoJSON object ([#9352](https://github.com/mapbox/mapbox-gl-js/pull/9352)). Note: We are aware of an edge case in which points with wrapped coordinates (e.g. longitude -185) are not evaluated properly. See ([#9442](https://github.com/mapbox/mapbox-gl-js/issues/9442)) for more information. An example of the `within` expression:<br>
```
chloekraw: Since we now have an additional note about this feature, should we convert formatting into two sub-bullets? Then I think we could also delete the `Note:`. 

- Add `within` expression for testing whether an evaluated feature lies within a given GeoJSON object (#9352). 
   - We are aware of an edge case in which points with wrapped coordinates (e.g. longitude -185) are not evaluated properly. See (#9442) for more information. 
   - An example of the within expression:
"icon-opacity": ["case", ["==", ["within", "some-polygon"], true], 1, ["==", ["within", "some-polygon"], false], 0]

ryanhamley: That sounds good to me

Issue #9443, opened 2020-03-20
Add support for slicing strings and arrays in expressions


lbutler:
## Motivation
Slicing of both strings and arrays is listed on the expression API master ticket  #6484.

I currently have data sets where I need to extract part of a string before the underlying data can be used. A simple example for me would be trying to style an attribute like "100mm". Ideally, I'd like to strip the unit from the string and be left with only the integer. In more complex cases, I also need to convert the integer based on the unit that is found.

As this is a significant limiter for my own use case, if the below is OK, I'd like to try to implement the new expression and create a pull request.

## Design Alternatives
At the moment I believe with expressions, there are no options to modify or change strings. Any modifications would need to be done to the base data before it is handed to the expression.

When search/regex is added it could be used as an alternative. However, that feature seems like it may be more complex to add at this time.

## Design
Javascript contains multiple methods to extract part of a string or an array; as the master ticket specifically says slice and the methods are the same for strings and arrays I believe this is what should be implemented.

If we were just looking at strings you could go with a more verbose `Left`, `Right` & `Mid` expressions similar to Excel, which may be more approachable for newer programmers. However, you can achieve all the same from the single `Slice` method and it isn't a large learning curve with plenty of examples online.

### Mock-Up

Slice Expression API:
```javascript
["slice", string | array, beginIndex: number]: string | array,
["slice", string | array, beginIndex: number, endIndex: number]: string | array,
```

String examples as per [MDN .slice() docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice)

```javascript
["slice", "The morning is upon us.", 1, 8]  // OUTPUT: he morn
["slice", "The morning is upon us.", 4, -2] // OUTPUT: morning is upon u
["slice", "The morning is upon us.", 12]    // OUTPUT: is upon us.
["slice", "The morning is upon us.", 30]    // OUTPUT: ""

["slice", "The morning is upon us.", -3]     // OUTPUT: 'us.'
["slice", "The morning is upon us.", -3, -1] // OUTPUT: 'us'
["slice", "The morning is upon us.", 0, -1]  // OUTPUT: 'The morning is upon us'
```

Array examples as per [MDN .slice() docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)

```javascript
[...]

Issue #6484, opened 2018-04-09
Master ticket tracking expression API completeness


anandthakker:
This ticket is intended to define what set of operators would constitute a "complete" expression API and track its implementation.  

 - Math https://github.com/mapbox/mapbox-gl-js/issues/5853
   - [ ] All the `Math.*` functions from Javascript
   - [ ] All the relevant functions from `NSExpression` 
   - [ ] `is-nan` https://github.com/mapbox/mapbox-gl-js/issues/5494
 - Geo
   - [ ] Haversine distance https://github.com/mapbox/mapbox-gl-js/issues/6062
 - Color
   - [ ] All the color-modifying functions from CSS4 https://github.com/mapbox/mapbox-gl-js/issues/5676
 - Arrays
   - [ ] Construction https://github.com/mapbox/mapbox-gl-js/issues/6155
   - [ ] Search https://github.com/mapbox/mapbox-gl-js/issues/6473
   - [ ] Slicing
 - Strings
   - [ ] Search https://github.com/mapbox/mapbox-gl-js/issues/4113 (but see also Regular Expressions)
   - [ ] Replace https://github.com/mapbox/mapbox-gl-js/issues/4100 (but see also Regular Expressions)
   - [ ] Slicing
 - [ ] Regular Expressions https://github.com/mapbox/mapbox-gl-js/pull/6228
 - Locale and language
   - [ ] Test for script support https://github.com/mapbox/mapbox-gl-js/issues/5807
   - [ ] Case and diacritic folding https://github.com/mapbox/mapbox-gl-js/issues/4136

1ec5:
Inspired by this issue, I attempted to compile a similar issue to track feature parity with NSPredicate and NSExpression: mapbox/mapbox-gl-native#11786.

gorshkov-leonid:
Hi, I have not found 
1. toHex(or toString(16)) 
1. substrig to save to variable
1. variable(set) / object(set) / array(s**p**lice)
1. how about supporting #FFFFFFFF(8 bytes) from [css4 ](http://dev.w3.org/csswg/css-color/#hex-notation)

stevage:
Oh, I didn't know about this ticket!

Me, I would love more string-handling functions:

- left
- right
- split
- titlecase (yes I know it's difficult to do in a locale-independent way)

lbutler:
> Oh, I didn't know about this ticket!
> 
> Me, I would love more string-handling functions:
> 
> * left
> * right
[...]

PR #9450, opened 2020-03-23
Add `slice` and `index-of` expressions

lbutler:
## Launch Checklist

The expression API master ticket #6484 lists slicing for both strings and arrays. As per #9443, @arindam1993 & @ryanhamley both suggested also including `index-of` to be able to search a string or array otherwise you would be limited to hardcoded indexes only.

This is my first PR into mapbox-gl-js so please let me know if I need to change/update anything to get this merged in, thanks!

    - implements `slice` to extract part of string or array
    - implements `index-of` to find the index of a value within a string or array
    - Fixed: #9443
    - `in` expression will now search for `null` within an array instead of always returning false
    - Fixed: #9465
 - [X] tagged `@mapbox/map-design-team` `@mapbox/static-apis` if this PR includes style spec API or visual changes
 - [X] tagged `@mapbox/gl-native` if this PR includes shader changes or needs a native port
 - [X] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Add `index-of` and `slice` expressions to search arrays and strings for the first occurrence of a specified value and return a section of the original array or string.</changelog>`

Edit Todo:
 - [X] Fix Flow type errors - apologies, first time with flow and didn't run the tests locally


src/style-spec/expression/definitions/in.js
```diff
@@ -55,37 +73,59 @@ class In implements Expression {
...
+            return new In(type, needle, haystack);
+        }
     }
 
     evaluate(ctx: EvaluationContext) {
         const needle = (this.needle.evaluate(ctx): any);
         const haystack = (this.haystack.evaluate(ctx): any);
 
-        if (needle == null || !haystack) return false;
+        if (!haystack && this.type.kind === 'boolean') return false;
```
lbutler: Previously `in` would allow the needle to be null but would always return false, this means that if you had an array with null values in it and you were searching for the nulls it would always return false.

```javascript
[1,2,3,null,5].indexOf(null) >= 0 === true
```

In addition, the check returns false if the haystack is null. This was raised in [the initial review](https://github.com/mapbox/mapbox-gl-js/pull/8876/files#r340824010) for the `in` expression.

> The difference is in whether or not we're asserting that the value is a string or array. If we do ['string', ['get', 'str']] and str is null, then we get a type error. But if we don't assert it's a string or array and just directly return null then the check on line 65 of in.js returns false because there's no value that can be checked

I've left it so that the `in` expression keeps this return path, but potentially we need to make `index-of` return -1 so it's consistent between the two expressions?

src/style-spec/expression/definitions/in.js
```diff
@@ -33,18 +41,28 @@ class In implements Expression {
...
     static parse(args: $ReadOnlyArray<mixed>, context: ParsingContext) {
-        if (args.length !== 3) {
+        const name: string = (args[0]: any);
+        assert(types[name], name);
+
+        if (name === 'in' && args.length !== 3) {
             return context.error(`Expected 2 arguments, but found ${args.length - 1} instead.`);
         }
 
+        if (name === 'index-of' && (args.length <= 2 ||  args.length >= 5))
```
arindam1993: I was actually suggesting an inheritance based refactor rather than a runtime check based one.

asheemmamoowala:
cc @mapbox/gl-native @mapbox/studio

src/style-spec/expression/definitions/in.js
lbutler: Ahh that sucks because that what I initially did but then swapped it to this method which is how it is in [`./coercion.js`](src/style-spec/expression/definitions/coercion.js).

There is also another method used in [`./comparison.js`](src/style-spec/expression/definitions/comparison.js) which uses a function to create the class.

I didn't want to add a third method of doing it so I went this way, probably should have gotten some clarification before I went and did these changes!

arindam1993: Ah, I was unaware of those patterns in those files. They seem like a pattern carried over from ES5, but I'm not 100% certain on this. @ryanhamley do you have any thoughts?

src/style-spec/expression/definitions/slice.js
```diff
@@ -0,0 +1,96 @@
...
+import type EvaluationContext from '../evaluation_context';
+import type {Type} from '../types';
+import type {Value} from '../values';
+
+function isSliceableType(type: Type) {
+    return type.kind === 'array' ||
+           type.kind === 'string' ||
+           type.kind === 'value';
+}
+function isAcceptableInputRuntimeValue(input: Array<Value> | string) {
```
arindam1993: minot nit: this is the same logic as `isSearchableRuntimeValue`, could you move it to a shared util file.

Sorry if this seems v minor, just trying to be extra strict on bundle size rn.

src/style-spec/expression/definitions/in.js
lbutler: No problems at all, happy to modify to whatever the mapbox team thinks fits best. I see at least three options:

1. As is, used in `coercion` - single class and check `arg[0]` to change function return (will admit, it is a bit messy..)
2. Factory method used in `comparision`
3. New pattern for expressions using a base class plus  inheritances classes for `index-of` and `in`

Potentially as a much larger refactor you could change the `Definition` type that goes into `CompoundExpression.register()` to accept generic return types, or overload with a return type and then you wouldn't need custom definitions for `index-of`, `slice` and `in` 

Either way, we can wait for @ryanhamley to provide some input and I can have enough crack at this PR tonight.

ryanhamley: I think `coercion` is a bit of a special case that's mostly used internally. Instead of combining `in` and `index-of` or having `index-of` inherit from `in`, I think it makes more sense to move common functions such as `isComparableType` to a single file. It may make sense to put them in `expression/types.js` since there are already some shared type checking functions in that file.

src/style-spec/expression/definitions/in.js
ryanhamley: We need to keep the type checking (e.g. `['string', ['get', 'str']] // -> null` should return a type error). But maybe we should allow the needle to be null. See https://github.com/mapbox/mapbox-gl-js/pull/7197#pullrequestreview-153154765

>since "in" is using the same equality semantics as "==" (i.e. JS ===), it should have the same constraints, so it can't be used with types where === doesn't make sense. Specifically the type of needle should be limited to these types. If the static type is not one of those five, it should be a parse error. Or, if the static type is value and the runtime type isn't one of the other four, it should be an evaluation error.

`null === null` returns `true` though so it seems like we should accept that value and this was an oversight. Technically, `undefined` should work too, but I'm not sure we want to do that. You would need to update the `isComparableRuntimeValue` function to include `null` and update the error on line 68 to mention that `null` is an acceptable value

src/style-spec/expression/definitions/in.js
lbutler: @ryanhamley & @arindam1993 thanks both for your input. I've refactored with the latest commits. I've moved the functions that check the types into `expression/types.js` and updated the three expressions to use these common functions.
I haven't created a base class that the other extends but if that's what we want to do then I'm happy to do another refactor.

src/style-spec/expression/definitions/in.js
lbutler: Just letting you know that this was updated in the latest commit and I added a test case as well

PR #9454, opened 2020-03-23
fix fill extrusion querying for colinear points

ansis:
This fixes the problem partially fixed by https://github.com/mapbox/mapbox-gl-js/pull/9138. The original fix handled only the case where the points where coincident. Colinear points are also problematic.

When checking for intersections with a fill-extrusion it has to calculate the distance to the plane the polygon exist on. If the first points of the geometry were colinear this calculation would fail.

It now:
- finds two distinct starting points a, b
- loops over all the remaining vertices looking for a value that is not colinear with a, b

Usually this will be the first vertex it encounters.

## Launch Checklist

 - [n/a] post benchmark scores
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix a bug where fill-extrusion features with colinear points were not return by map.queryRenderedFeatures(...)</changelog>`

@tmpsantos this probably needs to be ported to -native.


src/style/style_layer/fill_extrusion_style_layer.js
```diff
@@ -81,34 +81,40 @@ export function getIntersectionDistance(projectedQueryGeometry: Array<Point>, pr
...
 
-        const ab = b.sub(a);
-        const ac = c.sub(a);
-        const ap = p.sub(a);
+            const dotABAB = dot(ab, ab);
+            const dotABAC = dot(ab, ac);
+            const dotACAC = dot(ac, ac);
+            const dotAPAB = dot(ap, ab);
+            const dotAPAC = dot(ap, ac);
+            const denom = dotABAB * dotACAC - dotABAC * dotABAC;
```
arindam1993: can this be replaced by `ab.cross(ac)`? you could also use that to check for co-linerarity, right?

ansis: The objects are the `Point` class which doesn't have `.cross()` (I think that is in one of the gl-matrix classes). If it was already there this would be probably cleaner but I'm not sure it's worth adding at this point

> you could also use that to check for co-linerarity, right?

At first I checked for collinearity directly but because of precision stuff we'd need to pick a threshold for what we count as colinear. At first I hit a case where my colinear check said they were not colinear enough but the whole method still failed to produce a valid distance. I think checking the output should be more reliable because it avoids that gap

arindam1993: Fair enough, makes sense to me!

PR #9457, opened 2020-03-24
Convert recursive quicksort into iterative to avoid stack overflow

vallendm:
## Launch Checklist

We've recently switched to using the feature state property in our mapbox implementation. We found during one test - which includes rendering 30000+ points - a stack-overflow eminating from feature_position_map.js. This was in the sort function which is implemented as a custom quick sort using recursion. The issue is that there are certain circumstances where quicksort can have a worst-case scenario. The pivot position chosen can end up being the next sorted item in the list - whick leads to a recusion to sort the remaining N-1 items. If this happens continuously we can infact recurse through all items to be sorted which causes stack overflow issues. This change simply rewrites the sort to use a queue and an iterative approach. 

 - [ x ] briefly describe the changes in this PR
 - [ ] post benchmark scores
 - [ ] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix "Maximum call stack size exceeded" error when using feature-state.</changelog>`

### DIFF
--- a/src/data/feature_position_map.js
+++ b/src/data/feature_position_map.js
@@ -95,25 +95,42 @@ function getNumericId(value: mixed) {
 }
 
 // custom quicksort that sorts ids, indices and offsets together (by ids)
-function sort(ids, positions, left, right) {
-    if (left >= right) return;
+// implemented using iterative method rather than recursive
+function sort(ids, positions, _left, _right) {
 
-    const pivot = ids[(left + right) >> 1];
-    let i = left - 1;
-    let j = right + 1;
+    const stack = [];
 
-    while (true) {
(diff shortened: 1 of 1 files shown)


mourner:
Thanks a lot for the contribution @vallendm! I wasn't sure switching to manual queue was the best option, so I found an article about fixing this more idiomatically (and perhaps also faster / simpler) by [manual tail call optimization](https://www.geeksforgeeks.org/quicksort-tail-call-optimization-reducing-worst-case-space-log-n/) — can you try #9463 and see if it fixes your issue?

mourner:
Let's close this in favor of #9463 for now (but keep it in mind as an alternative).

PR #9463, opened 2020-03-25
Optimize tail calls in feature state sorting

mourner:
A simpler alternative to #9457.  Fixes a potential stack overflow when using feature state with very large input. cc @vallendm 

Explanation of the change: https://www.geeksforgeeks.org/quicksort-tail-call-optimization-reducing-worst-case-space-log-n/

## Launch Checklist

 - ~~write tests for all new functionality~~ already covered
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix a bug where using feature state on a large input could cause a stack overflow error</changelog>`

### DIFF
--- a/src/data/feature_position_map.js
+++ b/src/data/feature_position_map.js
@@ -95,25 +95,31 @@ function getNumericId(value: mixed) {
 }
 
 // custom quicksort that sorts ids, indices and offsets together (by ids)
+// uses Hoare partitioning & manual tail call optimization to avoid worst case scenarios
 function sort(ids, positions, left, right) {
-    if (left >= right) return;
+    while (left < right) {
+        const pivot = ids[(left + right) >> 1];
+        let i = left - 1;
+        let j = right + 1;
 
-    const pivot = ids[(left + right) >> 1];
-    let i = left - 1;
-    let j = right + 1;
+        while (true) {
+            do i++; while (ids[i] < pivot);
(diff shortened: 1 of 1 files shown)


arindam1993:
All browsers don't support TCO though

mourner:
@arindam1993 yes, otherwise this PR wouldn't be necessary :)

arindam1993:
🤦🏽‍♂️ I see it now, my bad.

karimnaaji:
@vallendm can you give this PR a try to see if it fixes the issue you initially brought up in your initial contribution? We'd like to keep this in track for the next release.

vallendm:
Appologizes for not replying to earlier comment - I was on vacation last week. I just pulled this PR and run my initial testcase against it and there are not stack-space issues. I also extracted the arguments to the sort which were causing the issue into another testfile and using the updated sort method that is fine too. I put some extra trace and the stack only gets to 13 deep. 

The only thing that I noticed is that there isn't a call to the sort function when zooming out to tiles at zoom level 0 so has anything else changed?

karimnaaji:
Ok that's good to hear! I don't believe anything else has changed in that part of the code recently, which versions were you comparing?

vallendm:
@karimnaaji - the previous version that I was looking at was v1.8.1.

PR #9466, opened 2020-03-25
Upgrade direct dependencies & do yarn upgrade

mourner:
- Upgrade direct dependencies, including updated `@mapbox/geojson-rewind`.
- Do `yarn upgrade` to upgrade subdependencies to latest patched versions, addressing security warnings.

### DIFF
--- a/build/test/build-tape.js
+++ b/build/test/build-tape.js
@@ -6,6 +6,7 @@ const fs = require('fs');
 module.exports = function() {
     return new Promise((resolve, reject) => {
         browserify(require.resolve('../../test/util/tape_config.js'), { standalone: 'tape' })
+            .transform("babelify", {presets: ["@babel/preset-env"], global: true})
             .bundle((err, buff) => {
                 if (err) { throw err; }
 
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "vt-pbf": "^3.1.1"
   },
   "devDependencies": {
+    "@babel/core": "^7.9.0",
     "@mapbox/flow-remove-types": "^1.3.0-await.upstream.2",
     "@mapbox/gazetteer": "^4.0.4",
     "@mapbox/mapbox-gl-rtl-text": "^0.2.1",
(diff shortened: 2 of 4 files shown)


ryanhamley:
The query tests are failing @mourner

mourner:
@ryanhamley right, noticed this after requesting a review... Although I'm pretty confused by the error. Something to do with `testem`, maybe @arindam1993 you have an idea of what's wrong?

mourner:
Digging deeper, it looks like `tap-parser` is no longer browserifiable without any transforms — will revert to an older version and file an issue there.

mourner:
Well this is super baffling! If you use Yarn to install `tap-parser` deps, it's not browserifiable, but if you use NPM (even after deleting `package-lock.json`), it works! Some weird stuff going on.

arindam1993:
Sigh! Looks like one of tap-parser' deps uses ES6 module syntax, we could add a transformer to that browserify script.

arindam1993:
I was able to get it to work by adding 
```
.transform("babelify", {presets: ["@babel/preset-env"], global: true})
```
and that fixed it.


I heard you like dependencies, so I used more dependencies to solve your dependency error. 😬

arindam1993:
Here's the change
https://github.com/mapbox/mapbox-gl-js/commit/b95e422579462c29b3e6a977b8b7e0104b1f6ad7

mourner:
@arindam1993 that's fixing the symptom, but I'm still baffled by why it doesn't work with Yarn but works with NPM! There must be something deeper here.

arindam1993:
Sigh, after some thought I think adding an ES6 transformer is ok for now.

The main problem is that `tape` has to be built separately with `browserify`, at the time when I first did `rollup` has issues bundling `tape` due to weird issues with `rollup-plugin-node-globals`, `rollup-plugin-node-builtins` and `rollup-plugin-commonjs`.

Things might have changed since then, but I think a better long term approach would be to submit an upstream patch to `tape`, with a refactor that reorganizes the platform dependencies into a few files that can be proxied to browser versions directly using the `browser` field in package.json, rather than having to rely on maintaining a list of node-globals and builtins we would need on our end. Browserify pretty much solves this by injecting everything right now.

PR #9474, opened 2020-03-27
Update per mdn webgl recommendations

karimnaaji:
- Delete shader handles after succesful program compilation
- Query render to half float texture instead of calling `checkFramebufferStatus`, which could involve pipeline flush + round-trip on first frame

### DIFF
--- a/src/render/draw_heatmap.js
+++ b/src/render/draw_heatmap.js
@@ -100,17 +100,10 @@ function bindFramebuffer(context, painter, layer) {
 function bindTextureToFramebuffer(context, painter, texture, fbo) {
     const gl = context.gl;
     // Use the higher precision half-float texture where available (producing much smoother looking heatmaps);
-    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, painter.width / 4, painter.height / 4, 0, gl.RGBA,
-        context.extTextureHalfFloat ? context.extTextureHalfFloat.HALF_FLOAT_OES : gl.UNSIGNED_BYTE, null);
-
+    // Otherwise, fall back to a low precision texture
+    const internalFormat = context.extRenderToTextureHalfFloat ? context.extTextureHalfFloat.HALF_FLOAT_OES : gl.UNSIGNED_BYTE;
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, painter.width / 4, painter.height / 4, 0, gl.RGBA, internalFormat, null);
     fbo.colorAttachment.set(texture);
-
-    // If using half-float texture as a render target is not supported, fall back to a low precision texture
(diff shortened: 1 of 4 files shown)


src/render/draw_heatmap.js
```diff
@@ -100,17 +100,10 @@ function bindFramebuffer(context, painter, layer) {
...
+    const internalFormat = context.extRenderToTextureHalfFloat ? context.extTextureHalfFloat.HALF_FLOAT_OES : gl.UNSIGNED_BYTE;
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, painter.width / 4, painter.height / 4, 0, gl.RGBA, internalFormat, null);
     fbo.colorAttachment.set(texture);
-
-    // If using half-float texture as a render target is not supported, fall back to a low precision texture
-    if (context.extTextureHalfFloat && gl.checkFramebufferStatus(gl.FRAMEBUFFER) !== gl.FRAMEBUFFER_COMPLETE) {
-        context.extTextureHalfFloat = null;
-        fbo.colorAttachment.setDirty();
-        bindTextureToFramebuffer(context, painter, texture, fbo);
-    }
```
mourner: The reason why it was implemented this way is because of this excerpt from [the `EXT_color_buffer_half_float` spec](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_color_buffer_half_float.txt):

> Although it is expected that all implementations supporting this extension support rendering to at least one HALF_FLOAT_OES format, applications must check framebuffer completeness to determine which formats are supported.

So basically it says that the extension being enabled doesn't guarantee that writing to a half-float framebuffer will succeed. I don't know how often this situation happens in practice (maybe never?), but we should be aware of the risk of this PR breaking heatmaps for some users (where it won't fall back to low precision properly).

karimnaaji: I think that's a recommendation for the desktop GL extension `GL_EXT_color_buffer_half_float` which gets exposed a bit differently for the webgl one.

For example, [here](https://github.com/chromium/chromium/blob/b9761b6796247f3436bd816b36e524ccc86ea09d/gpu/command_buffer/service/feature_info.cc#L1784) are the thorough tests that chromium goes through before exposing it to the WebGL API, and does the completeness check. I have not seen similar suggestion in [mdn](https://developer.mozilla.org/en-US/docs/Web/API/EXT_color_buffer_half_float) or [the webgl spec reference](https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/).

I tested this last friday on both a iphone 6 and mali-t760 (galaxy s3) which don't report supporting this extension, and fallback appropriately. I would be ok in leaving it as is, but with these concerns we could leave an assert probe and see whether it gets triggered until the beta completes?

karimnaaji:
I would like to add heatmap.html as part of the release testing, since its code path has a few things that would be nice to test out as part of our release process (offscreen rendering and these type of extensions).

Issue #9479, opened 2020-03-30
v1.9.0 add "fill-extrusion-pattern" webgl is not working properly


nuancheng:
The latest version adds 3D architectural texture, and the webgl is not working properly
V1.8.0 is normal

**mapbox-gl-js version**: v1.9.0

**browser**: Google Chrome  80.0.3987.149

### Steps to Trigger Behavior

 1. v1.9.0 "vector" layer add "fill-extrusion-pattern" img

### Link to Demonstration
`"fill-extrusion-pattern": {
                property: "height",
                type: "interval",
                stops: [[40, "catcat"]]
              },`

wrong：https://jsbin.com/yazutumoxa/edit?html,output
Expectant：https://jsbin.com/yazutumoxa/3/edit?html,output

nuancheng:
wrong url：https://jsbin.com/kiyarubuji/edit?html,output

mourner:
@karimnaaji might be related to the change in #9372?

karimnaaji:
I checked the sanity of #9372 to see if it introduced a regression. There are no regression introduced by this specific PR, though, while looking at that I found another bug, which is in the vertex layout of pattern attributes. 

Explicitly setting [this offset to 0](https://github.com/mapbox/mapbox-gl-js/blob/master/src/data/program_configuration.js#L326) assumes that all vertices within this layout start at the same byte in the vertex format. This is incorrect since [this layout](https://github.com/mapbox/mapbox-gl-js/blob/master/src/data/bucket/pattern_attributes.js) has 4 different components, each of which starts at byte offsets 0, 8, 16 and 17. This may have been silently broken for some time and adding new attributes in this layout did not take effect for DDS.

For this particular issue, @nuancheng , you need to use `addImage` with `{pixelRatio: 2}` option since  you are loading a `@2x` image manually, the difference is that in v1.8.0 we were using the device pixel ratio, which happened to match this 2x on a 2x screen. So this example will actually _accidentally_ works under a 2x screen in v1.8.0. But, on my 1x screen this image pixel ratio is downgraded to a 1x incorrectly in v1.8.0:

![Screen Shot 2020-03-30 at 4 28 46 PM](https://user-images.githubusercontent.com/7061573/77971501-afa0ad00-72a3-11ea-9448-b12746a2fae7.png)
 
[...]

PR #9482, opened 2020-03-30
Fix pattern attributes vertex layout

karimnaaji:
While testing out issue https://github.com/mapbox/mapbox-gl-js/issues/9479 I found a bug in the way we create the vertex buffers for `CrossFadedCompositeBinder`.

Explicitly setting [this offset to 0](https://github.com/mapbox/mapbox-gl-js/blob/master/src/data/program_configuration.js#L326) assumes that all vertices within this layout start at the same byte in the vertex format. This is incorrect since [this layout](https://github.com/mapbox/mapbox-gl-js/blob/master/src/data/bucket/pattern_attributes.js) has 4 different components, each of which starts at byte offsets 0, 8, 16 and 17, and we end up giving the wrong offsets to `gl.vertexAttribPointer`.

This may have been silently broken for some time and adding new attributes in this layout did not take effect for DDS so I added a render test around it.

## Launch Checklist

 - [x] `<changelog>Fix pixel ratio option for *-pattern properties referencing images added with addImage when used with data driven styling</changelog>`


src/data/program_configuration.js
```diff
@@ -312,20 +312,13 @@ class CrossFadedCompositeBinder implements AttributeBinder {
...
         this.type = type;
         this.useIntegerZoom = useIntegerZoom;
         this.zoom = zoom;
         this.layerId = layerId;
 
-        this.paintVertexAttributes = names.map((name) => ({
-            name: `a_${name}`,
-            type: 'Uint16',
-            components: 4,
-            offset: 0
```
karimnaaji: The issue lies here, instead of using `0`, we need to use `patternAttributes.members` which has the correct offsets and sizes.

src/data/program_configuration.js
```diff
@@ -371,8 +364,8 @@ class CrossFadedCompositeBinder implements AttributeBinder {
 
     upload(context: Context) {
         if (this.zoomInPaintVertexArray && this.zoomInPaintVertexArray.arrayBuffer && this.zoomOutPaintVertexArray && this.zoomOutPaintVertexArray.arrayBuffer) {
-            this.zoomInPaintVertexBuffer = context.createVertexBuffer(this.zoomInPaintVertexArray, this.paintVertexAttributes, this.expression.isStateDependent);
-            this.zoomOutPaintVertexBuffer = context.createVertexBuffer(this.zoomOutPaintVertexArray, this.paintVertexAttributes, this.expression.isStateDependent);
+            this.zoomInPaintVertexBuffer = context.createVertexBuffer(this.zoomInPaintVertexArray, patternAttributes.members, this.expression.isStateDependent);
+            this.zoomOutPaintVertexBuffer = context.createVertexBuffer(this.zoomOutPaintVertexArray, patternAttributes.members, this.expression.isStateDependent);
```
arindam1993: Isn't this hard-coding though? are patterns the only kind of corss faded composite attributes?

karimnaaji: Yes they are the only type of vertices at this time, we don't emplace any other type in this layout here so we can make it explicit https://github.com/mapbox/mapbox-gl-js/pull/9482/files#diff-0ecc9a72f8fbadd5fe4d9a5f4ab26691R350. We can leave a note to prevent programmer error if this gets extended.

karimnaaji: I think the note should be in `paintAttributeNames` [here](https://github.com/mapbox/mapbox-gl-js/blob/b238b15f64e00540c29d0ec3d2065861aa90d1f9/src/data/program_configuration.js#L625), as adding attributes might lead to issues with the other type of binders such as [here](https://github.com/mapbox/mapbox-gl-js/blob/b238b15f64e00540c29d0ec3d2065861aa90d1f9/src/data/program_configuration.js#L241) or [here](https://github.com/mapbox/mapbox-gl-js/blob/b238b15f64e00540c29d0ec3d2065861aa90d1f9/src/data/program_configuration.js#L170).

arindam1993: How about cross-referencing from a lookup table?

We could either update the lookup table by hand, or codegen it in `generate-struct-arrays.js`.

karimnaaji: Logging offline discussion: we decided on going the simple way for the time being to cut out extra logic.

arindam1993: 👍

PR #9487, opened 2020-04-01
Added ability to draw arcs

ws-hofshteyn:
The main purpose of these changes is making an arc from a circle.

There may be cases when needed to add a not fully completed circle. One of examples is a displaying lighthouses for marine map. There is an international specification which contains a fully description of visualization it on map (*12.8.6.1 Sector lights (see S-4 – B-475.1)*):

[https://iho.int/uploads/user/pubs/standards/s-4/S4_V4-8-0_Oct_2018_EN.pdf](S4_V4-8-0_Oct_2018_EN.pdf)

Mapbox-gl does not contains such functionality. The empty space on a circle could have different size. Here is a sample code that drawing a circle from 0 to 270 angle and a screenshot of how does it looks like with the current changes:

```
map.on('load', () => {
  map.addSource('points', {
    'type': 'geojson',
    'data': {
      'type': 'FeatureCollection',
      'features': [
        {
          'type': 'Feature',
          'geometry': {
            'type': 'Point',
            'coordinates': [0, 0]
          }
        }
      ]
    }
  });
  map.addLayer({
    'id': 'points',
    'type': 'circle',
    'source': 'points',
    'paint': {
      'circle-radius': 50,
      'circle-opacity': 0,
      'circle-stroke-width': 3,
      'circle-stroke-color': '#006400',
      'circle-stroke-opacity': 1,
      'circle-start-angle': 0,
      'circle-end-angle': 270
    },
  });
});
```

Arc example

## Launch Checklist


src/shaders/circle.fragment.glsl
```diff
@@ -25,6 +29,15 @@ void main() {
 
     float opacity_t = smoothstep(0.0, antialiased_blur, extrude_length - 1.0);
 
+    float ang = atan( v_data.x, v_data.y );
+    float ang0r = (180. - (ang_end + 180.) + 720.) / 180. * 3.14;
+    float ang1r = (180. - (ang_start + 180.) + 720.) / 180. * 3.14;
+    if (ang < 0.)
+        ang += 6.28;
+    if ( !((ang > ang0r && ang < ang1r) ||
```
karimnaaji, line 37: Branches in shaders can be expensive, this block can be simplified with the following:

```c
#define DEG_TO_RAD(v) (v * 0.01745329251994329576)
#define PI 3.14159265358979323846

float angle_r = atan(v_data.y, -v_data.x) + PI;
float angle_start_r = DEG_TO_RAD(ang_start);
float angle_end_r = DEG_TO_RAD(ang_end);
opacity_t = mix(opacity_t, 0.0, float(angle_r < angle_start_r || angle_r > angle_end_r));
```

If we could do the conversion deg -> rad outside of the shader that would be ideal too.

src/shaders/circle.fragment.glsl
```diff
@@ -2,6 +2,8 @@ varying vec3 v_data;
 
 #pragma mapbox: define highp vec4 color
 #pragma mapbox: define mediump float radius
+#pragma mapbox: define mediump float ang_start
```
karimnaaji, line 5: I would rename `ang_start` and `ang_end` to their full name `angle_start` and `angle_end` to be more explicit.

src/style-spec/reference/v8.json
```diff
@@ -4450,6 +4450,66 @@
       },
       "property-type": "data-driven"
     },
+    "circle-ang-start": {
+      "type": "number",
+      "default": 0,
+      "minimum": 0,
+      "transition": true,
+      "units": "angle",
+      "doc": "Angle value to start drawing a circle.",
```
karimnaaji, line 4459: Could you update the documentation to explain in which quadrant it starts, which direction it is going in and update the unit to degrees instead of angle?

andrewharvey: and ideally what it's relative to (map grid north instead of viewport)?

src/style-spec/reference/v8.json
```diff
@@ -4450,6 +4450,66 @@
...
+      "minimum": 0,
+      "transition": true,
+      "units": "angle",
+      "doc": "Angle value to finish drawing a circle.",
+      "sdk-support": {
+        "basic functionality": {
+          "js": "----",
+          "android": "----",
+          "ios": "----",
+          "macos": "----"
```
samanpwbb, line 4495: The convention for documenting sdk support is to exclude the key if there is no support. For gl-js, best to use "1.10.0" for now (we'll then update depending on release schedule).

src/style-spec/reference/v8.json
```diff
@@ -4450,6 +4450,66 @@
...
+        },
+        "data-driven styling": {
+          "js": "0.18.0",
+          "android": "5.0.0",
+          "ios": "3.5.0",
+          "macos": "0.4.0"
+        }
```
samanpwbb, line 4502: data-driven styling key needs to be updated to be accurate

ws-hofshteyn: Not fully clear to what value I should update. The minor versions?

samanpwbb: Remove all keys except for js, and then if on js, if the property supports evaluating data expressions (like `['get', 'foo']`), then set the js value to the next release version ("1.10.0").

src/style-spec/reference/v8.json
```diff
@@ -4450,25 +4450,19 @@
...
-    "circle-ang-start": {
+    "circle-start-angle": {
       "type": "number",
       "default": 0,
       "minimum": 0,
       "transition": true,
-      "units": "angle",
-      "doc": "Angle value to start drawing a circle.",
+      "units": "degrees",
+      "doc": "Angle value to start drawing a circle. The circle is represented as a coordinate system relative to the viewport and 0° - value of the beginning of the first quadrant. Then drawing goes counter clockwise.",
```
karimnaaji, line 4459: 👍

andrewharvey: I'm a bit confused by "relative to the viewport and 0°". For other rotation values, we have a rotation alignment property to specify either relative to the viewport or relative to the map eg. https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-symbol-icon-rotation-alignment. So I think for conistancy we should either also have an angle-alignment property which is map or viewport, or specify in the description if this is map or viewport.

Although I think most of the time you'd want a map alignment (eg. direction range of a lookout/viewpoint), you may also sometimes want a viewport alignment?

karimnaaji: That makes sense, based on the [initial document](https://iho.int/uploads/user/pubs/standards/s-4/S4_V4-8-0_Oct_2018_EN.pdf) it looks like the initial intention was to support it relative to the map, but having options would be more consistent. @ws-hofshteyn have you tested this on tilted views?

andrewharvey: > have you tested this on tilted views?

and rotated views

karimnaaji:
Thanks for adding the render tests @ws-hofshteyn !

We'll postpone this to the next release, since we have two items of concerns left:
- Potential reach of `GL_MAX_VERTEX_ATTRIBS`/`GL_MAX_VARYING_VECTORS`: This seems ok for WebGL as per http://webglstats.com/webgl/parameter/MAX_VERTEX_ATTRIBS. Before going forward we'll have to address going over these limits to prevent conflicting port with gl-native using gl es 2.0 specs; which has more constraining limits, refer: http://docs.gl/es2/glGet.
- Determine and clarify whether we need a relative viewport or map alignment property. This would be consistent with other specs already in place, such as https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-symbol-icon-rotation-alignment.

ansis:
@ws-hofshteyn thanks for taking a shot at this problem!

I'm imagining that there are multiple further capabilities that might be useful. For example, drawing multiple arcs for a single feature, combining arcs with circles, collision detection.

I'm wondering if it would be possible to provide these in a more generic/composable way instead of as additional style spec properties for circles. For example, would some way of generating images that can be used with symbol layers work? @kkaefer do you have any thoughts about whether this would tie into vector icons work?

karimnaaji:
Converting this PR to draft for the time being, we'll look at the approach when considering vector icons.

ryanhamley:
@karimnaaji @ansis what is the status of this PR? is this something we'd like to consider still?

shaarman:
This looks very interesting. Could it potentially be used to create pie charts (or perhaps donut charts), by stacking multiple circle layers together?

ws-hofshteyn:
@karimnaaji , Hi! Are there any news?

Nikcharn:
any news?

ryanhamley:
Hi @ws-hofshteyn I apologize for the lengthy delay. This implementation is well done and seems to solve this particular use case, but I don't think it's something we can maintain across all our products. Thanks for your interest and effort.

PR #9489, opened 2020-04-01
Allow client to retry RTLTextPlugin load after NetworkError

ryanhamley:
## Launch Checklist

    - When the RTLTextPlugin fails due to a `NetworkError`, it's currently impossible to try loading the plugin again. We should allow this to be a recoverable error by setting the `pluginStatus` to `error` so that `setRTLTextPlugin` can be called again
    - the error in `triggerPluginCompletionEvent` is a string in this case because it has to be serialized from the worker. this means we have to check for presence of the `NetworkError` string rather than check for a particular error code
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Correctly set RTL text plugin status if the plugin URL could not be loaded. 
This allows adding retry logic if there was a network error when loading the plugin.</changelog>`

### DIFF
--- a/debug/rtl.html
+++ b/debug/rtl.html
@@ -30,6 +30,7 @@
 <script src='/dist/mapbox-gl-dev.js'></script>
 <script src='/debug/access_token_generated.js'></script>
 <script>
+// badPluginURL allows us to simulate a NetworkError that we can recover from by retrying plugin loading
 var badPluginURL = 'http://localhost:9966/debug/2762.html';
 var goodPluginURL = 'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.0/mapbox-gl-rtl-text.js';
 var map = window.map = new mapboxgl.Map({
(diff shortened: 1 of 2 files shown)


debug/rtl.html
```diff
@@ -30,7 +30,8 @@
 <script src='/dist/mapbox-gl-dev.js'></script>
 <script src='/debug/access_token_generated.js'></script>
 <script>
-var pluginURL = 'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.0/mapbox-gl-rtl-text.js';
```
asheemmamoowala, line 33: nit: Add commentt o indicate what is happening and expected if you plan to commit this change

asheemmamoowala:
The changelog entry could be more clear about the underlying bug that is being fixed:
How about:
```
Correctly set RTL text plugin status if the plugin URL could not be loaded. 
This allows adding retry logic if there was a network error in loading the plugin.
```

Issue #9384, opened 2020-03-06
Shader compilation speedup


arindam1993:
@sgolbabaei  and I had a really good session yesterday and discovered some neat optimization opportunities.
## The Hypothesis:

Currently, we introduce some jank and take a hit on our load times due to slowness in compiling shaders, which is mostly caused by calling  `gl.getProgramParameter()` [to retrieve the number of uniforms and attributes in each program](https://github.com/mapbox/mapbox-gl-js/blob/fa3616825411009284b99daae7f27368fad774af/src/render/program.js#L79).
![Screen Shot 2020-02-27 at 4 02 53 PM](https://user-images.githubusercontent.com/1449257/76040866-65771680-5f05-11ea-806c-05ee740b2099.png)
The above screenshot shows it takes roughly 100 ms on a Macbook Pro to render the first frame.

Further investigation showed that querying for attributes is more time consuming than querying for uniforms, 
![image2](https://user-images.githubusercontent.com/1449257/76041008-e59d7c00-5f05-11ea-94ef-00d5e7a3b680.png)
Logically it seems like we can eliminate the call to `gl.getProgramParameter()` completely since our styling pipeline with `ProgramConfiguration` + `Binders` already tracks and computes the necessary attributes and uniforms necessary for runtime styling. We should be able to combine that, with the static data in the shader to extract this metadata ourselves, instead of querying gl.

## The Experiment:
For a quick prototype, we created a style with just one layer type, lines. The screenshot above shows timing logs for each `gl.getProgramParameter()` call. It’s roughly between 10 ~30ms and it seems to scale with the number of attributes. 

In `line.vertex.glsl` there are two fixed attributes `a_data` and `a_pos_normal`, by adding a tiny predefined metadata object we can quickly lookup the number of fixed attributes at runtime.
```
export const shaderMetaData = {
    line: 2,
    lineGradient: 2,
    linePattern: 2,
    lineSDF: 2,
    clippingMask: 1
}
```
We can likely generate this at compile time with some codegen that analyzes the shader text, rather than updating it by hand.

For calculating the variable attributes, i,e the ones that depend on runtime-styling parameters, we can go through all attribute Binders and sum up the number of vertex attributes. 

## The Results:
![image1](https://user-images.githubusercontent.com/1449257/76041174-39a86080-5f06-11ea-836e-ead8a285bd82.png)

The attribute Query time is down from 10 ~30ms to sub 1ms. :tada:

[...]

PR #9491, opened 2020-04-02
[release-caspian] Fix binder.setUniform error  symbol in draw loop

### LINKED ISSUE #9468: Exception while rendering: this.binders[l].setUniform is not a function
I upgraded from v1.7.0 to 1.8.1, but had to roll back because of elevated exceptions across all browsers that looked like:

```
this.binders[l].setUniform is not a function TypeError: this.binders[l].setUniform is not a function
   at SourceExpressionBinder.setUniforms (src/data/program_configuration.js:519)
   at Program.draw (src/render/program.js:135)
   at drawSymbolElements (src/render/draw_symbols.js:389)
   at drawLayerSymbols (src/render/draw_symbols.js:378)
   at drawSymbols (src/render/draw_symbols.js:74)
   at Painter.renderLayer (src/render/painter.js:495)
   at Painter.render (src/render/painter.js:453)
   at Map._render (src/ui/map.js:2195)
   at Map.triggerRepaint (src/ui/map.js:2317)
```

I can only see that these exceptions started getting thrown, but I do not know exactly what triggers them or if they result in any incorrect behavior. 

**mapbox-gl-js version**: 1.8.1

**browser**: Chrome (Mac, Windows, Linux, Android), Safari (iOS, Mac), Edge, IE11

**Steps to Trigger Behavior**: unknown

arindam1993:
Fixes #9468 .

This regression was caused as an unintended sideeffect of https://github.com/mapbox/mapbox-gl-js/commit/d85afadd994d93fa7bb42bb614de02217030ee94.

The Uniform caching in that change works as expected, except in the symbol layer wherein the `Program` instance is only initialized once for all tiles in the draw loop.

Since the tiles need a round trip from the worker when certain paint properties are updated, for a few frames there can be mix of old and new tiles on screen, prior to the optimization the uniform bindings would be re-evaluated every frame, so it would self-heal. But the new implementation requires that the instance of `ProgramConfiguration` passed to `Program#constructor` and `Program#draw` be the same.

## Launch Checklist

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fixes a bug  in which an exception would get thrown when updating symbol layer paint property using `setPaintProperty`</changelog>`

### DIFF
--- a/src/render/draw_symbol.js
+++ b/src/render/draw_symbol.js
@@ -240,8 +240,6 @@ function drawLayerSymbols(painter, sourceCache, layer, coords, isText, translate
 
     const depthMode = painter.depthModeForSublayer(0, DepthMode.ReadOnly);
 
-    let program;
-    let size;
     const variablePlacement = layer.layout.get('text-variable-anchor');
 
     const tileRenderState: Array<SymbolTileRenderState> = [];
@@ -259,10 +257,8 @@ function drawLayerSymbols(painter, sourceCache, layer, coords, isText, translate
         const sizeData = isText ? bucket.textSizeData : bucket.iconSizeData;
         const transformed = pitchWithMap || tr.pitch !== 0;
 
-        if (!program) {
-            program = painter.useProgram(getSymbolProgramName(isSDF, isText, bucket), programConfiguration);
-            size = symbolSize.evaluateSizeForZoom(sizeData, tr.zoom);
-        }
+        const program = painter.useProgram(getSymbolProgramName(isSDF, isText, bucket), programConfiguration);
+        const size = symbolSize.evaluateSizeForZoom(sizeData, tr.zoom);
 
         let texSize: [number, number];
         let texSizeIcon: [number, number] = [0, 0];


arindam1993:
Every other layer does it this way tho, I ran some profiles on my macbook and nothing was really showing up there. Also, the caching of `size` also seems logically incorrect because it could get initialized from a text bucket, but apply the value to icon sizes later all in the draw loop when it encounters an icon bucket.

> So maybe a gentler fix would be to simply check for binder.setUniform before calling it in ProgramConfiguration, which would theoretically fix the error for these brief frames without having to change the drawing code.

Wouldn't that fix also have potential sideffects wherein the shader has a uniform but then the binders have an attribute array?.

@sgolbabaei is actually working on a refactor of the relationship between `ProgramConfiguration`, `Program` and `Bucket` as a part of #9384. I think she's already got it to the point wherein `ProgramConfiguration` doesnt depend on layoutAttributes anymore, so. it can be created on the main thread now, without any tile data. That should make it much easier to have a simpler ( and maybe more performant ?)mapping between `Program`s and `ProgramConfiguration`s without all the loop-de-loop that there is right now.

Because right now, `ProgramConfigurations` are created in the worker in a `Bucket`, even though all  its state is largely dependent on the style layer and its props, and then its passed into the `Program` constructor, and then re-passed into the `Program#draw`, while `Program` is also caching metadata it received from `ProgramConfiguration` back when it was first created.

mourner:
@arindam1993 found the original optimization I recalled — #3499 — apparently at that point it was only setting state once on program change, but that got refactored out at some point or another, likely after introducing program state tracking. So the change should be good after all, and I'm really looking forward to @sgolbabaei's refactoring — currently the relation is indeed pretty confusing (I'm still struggling with fully understanding how it all works even after all these years).

PR #9496, opened 2020-04-02
Allow transforming non web mercator WMS sources

chrispahm:
The main purpose of this PR is to allow adding external WMS layers which are not projected in Web Mercator. 

Some publicly available datasets are only available in other, non web mercator projections (e.g. [ETRS89](https://en.wikipedia.org/wiki/European_Terrestrial_Reference_System_1989#cite_note-4) for many local datasets in Europe). To the best of my kownledge, adding them as a layer to a Maxpbox map either requires to transform the coordinates on the fly using a proxy server, or fiddling with the [`transformRequest`](https://docs.mapbox.com/mapbox-gl-js/api/#map) callback in the `map` options.

With the proposed feature in this PR, users would (only) need to define a `transformBBOX` callback in the source object, and add the `BBOX={bbox-transform}` query parameter to their tiles urls:

```js
map.addSource('wms-test', {
        "type": "raster",
        transformBBOX (bbox) {
            // convert bbox string into an array of (number) coordinates
            var array = bbox.split(',').map(function (coord) {
                return Number(coord);
            });
            // transform bbox from web mercator into WGS84 (or any other projection)
            // in this example, coordinates need to be switched from lng/lat (mapbox)
            // to lat/lng as well
            var toWGS84 = new proj4.Proj('EPSG:4326');
            var fromMapbox = new proj4.Proj('EPSG:3857');
            var leftUpper = proj4(fromMapbox, toWGS84, array.slice(0, 2)).reverse();
            var rightLower = proj4(fromMapbox, toWGS84, array.slice(2)).reverse();
            // return a bbox string
            return leftUpper.concat(rightLower).join(',');
        },
        "tiles": ['https://services.bgr.de/wms/boden/eusr5000/?BBOX={bbox-transform}&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image/png&TRANSPARENT=true&LAYERS=5&CRS=EPSG:4326&STYLES=&WIDTH=256&HEIGHT=256'
        ],
        "tileSize": 256
    });
```

The callback is called with an `EPSG:3857` formatted BBOX `string`, and should return a BBOX `string` as well. 
When the query parameter `BBOX={bbox-transform}` is present, but no `transformBBOX` method is defined, the BBOX will be projected in `EPSG:3857` (similar to using `BBOX={bbox-epsg-3857}`.

Here's a picture of the above example in action (displaying the different soil regions within Europe):
image

Thank you for this amazing library and for considering my PR!

## Launch Checklist

 - [ ]  tagged @mapbox/map-design-team @mapbox/static-apis if this PR includes style spec API or visual changes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -139,7 +139,7 @@ Here is a recommended way to get setup:
 1. Fork this project
 2. Clone your new fork, `git clone git@github.com:GithubUser/mapbox-gl-js.git`
 3. `cd mapbox-gl-js`
-4. Add the Mapbox repository as an upstream repository: `git add remote upstream git@github.com:mapbox/mapbox-gl-js.git`
+4. Add the Mapbox repository as an upstream repository: `git remote add upstream git@github.com:mapbox/mapbox-gl-js.git`
 5. Create a new branch `git checkout -b your-branch` for your contribution
 6. Write code, open a PR from your branch when you're ready
 7. If you need to rebase your fork's PR branch onto master to resolve conflicts: `git fetch upstream`, `git rebase upstream/master` and force push to Github `git push --force origin your-branch`
--- a/debug/wms_transform.html
+++ b/debug/wms_transform.html
@@ -24,9 +24,9 @@ var fromMapbox = new proj4.Proj('EPSG:3857');
 
 var map = window.map = new mapboxgl.Map({
     container: 'map',
-    zoom: 16,
-    center: [-74.0447, 40.6892],
-    style: 'mapbox://styles/mapbox/streets-v9',
+    zoom: 4,
+    center: [4.72, 47.91],
+    style: 'mapbox://styles/mapbox/light-v10',
     hash: true
 });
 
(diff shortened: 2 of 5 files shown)


ryanhamley:
Hi @chrispahm thanks for submitting this PR. I think there's a couple things weighing against us accepting it though. For one, quite a bit of GL JS is predicated on the sources being Mercator. This approach may make it easy for users to mess up their maps while being useful for a relatively small set of power users. We would like to expand beyond Mercator projections at some point, but it will require a comprehensive approach to get it right.

GL JS is also part of an ecosystem that includes many other libraries such as GL Native and Mapbox Studio. This PR partially changes the `raster` source which is defined by the [style spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#raster). The style spec doesn't have a great method for deprecating attributes or making breaking changes and is shared among the Mapbox ecosystem, so we're hesitant to add new attributes unless there's a very good reason to (this PR would also need to update the style-spec itself but that's a separate issue).

We appreciate the contribution but I think this is not something we can accept at this time. Thank you.

PR #9512, opened 2020-04-08
fix DragRotate when mouseup occurs outside window or iframe

### LINKED ISSUE #4622: map stuck in DragRotate when mouse up occurs outside an iframe
**mapbox-gl-js version**: 0.36.0

### Steps to Trigger Behavior

Similar to #3389, when a map appears in an iframe (eg. https://www.mapbox.com/mapbox-gl-js/examples/) and:

1. User starts a DragRotate (either via Ctrl+left mouse or right mouse)
2. Drags the cursor outside the iframe (with Ctrl or right mouse still down)
3. Then releases Ctrl or right mouse outside the iframe
4. Move the mouse back in the iframe and the map is stuck in drag rotate mode.

### Possible Solution
Similar to the other solution #4390 we can detect when the mouse leaves the iframe with

```js
window.addEventListener('mouseout', function (e) {
    if (e.toElement === null) {
      // mouse left iframe
   }
});
```

However we probably want to allow users to DragRotate even with the mouse leaving the iframe, and only prevent the limited case where the mouse up occurs outside, so I'm guessing we might want to check the first `mousemove` on the map after a window `mouseout` event if the Ctrl or right mouse is still down, and if not end the interaction.

ansis:
fix #4622

When a user:
- starts a mouse rotation in Chrome
- moves the mouse outside the window or iframe
- releases the button

no `mouseup` event is ever fired so the gesture never ends. This PR works around this by ending the gesture on the first `mousemove` that doesn't have this value.

From the issue:
> In some browsers, this can be fixed by avoiding calling e.preventDefault() on mousedown (see d3/d3-drag#9).

We don't call preventDefault() on mousedown and this still happens.

---

I tried adding a selenium test but it complained when I moved the mouse outside of the window.

Manually tested in:

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge
- [x] IE11

TODO:
- [x] fix all the unit tests that don't specify `buttons` for `mousemove` events it simulates

## Launch Checklist

 - [n/a] post benchmark scores
 - [n/a] tagged `@mapbox/map-design-team` `@mapbox/static-apis` if this PR includes style spec API or visual changes
 - [n/a] tagged `@mapbox/gl-native` if this PR includes shader changes or needs a native port
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix bug where map got stuck in a DragRotate interaction if it's mouseup occurred outside of the browser window or iframe.</changelog>`

 
cc @andrewharvey

### DIFF
--- a/src/ui/handler/mouse.js
+++ b/src/ui/handler/mouse.js
@@ -66,7 +66,7 @@ class MouseHandler {
             // the window or iframe:
             // https://github.com/mapbox/mapbox-gl-js/issues/4622
             //
-            // If the button is no longer pressed during this `mousemove` it have have
+            // If the button is no longer pressed during this `mousemove` it may have
             // been released outside of the window or iframe.
             this.reset();
             return;
--- a/src/ui/handler/mouse.js
+++ b/src/ui/handler/mouse.js
@@ -6,6 +6,17 @@ import type Point from '@mapbox/point-geometry';
 const LEFT_BUTTON = 0;
 const RIGHT_BUTTON = 2;
 
+// the values for each button in MouseEvent.buttons
+const BUTTONS_FLAGS = {
+    [LEFT_BUTTON]: 1,
+    [RIGHT_BUTTON]: 2
+};
+
+function buttonStillPressed(e: MouseEvent, button: number) {
+    const flag = BUTTONS_FLAGS[button];
+    return e.buttons === undefined || (e.buttons & flag) !== flag;
+}
+
 class MouseHandler {
 
     _enabled: boolean;
@@ -50,6 +61,17 @@ class MouseHandler {
         if (!lastPoint) return;
         e.preventDefault();
 
+        if (buttonStillPressed(e, this._eventButton)) {
+            // Some browsers don't fire a `mouseup` when the mouseup occurs outside
+            // the window or iframe:
+            // https://github.com/mapbox/mapbox-gl-js/issues/4622
+            //
+            // If the button is no longer pressed during this `mousemove` it have have
+            // been released outside of the window or iframe.
+            this.reset();
+            return;
+        }
+
         if (!this._moved && point.dist(lastPoint) < this._clickTolerance) return;
         this._moved = true;
         this._lastPoint = point;
@@ -59,6 +81,7 @@ class MouseHandler {
     }
 
     mouseupWindow(e: MouseEvent) {
+        if (!this._lastPoint) return;
         const eventButton = DOM.mouseButton(e);
         if (eventButton !== this._eventButton) return;
         if (this._moved) DOM.suppressClick();
--- a/test/unit/ui/handler/drag_pan.test.js
+++ b/test/unit/ui/handler/drag_pan.test.js
@@ -13,6 +13,9 @@ function createMap(t, clickTolerance, dragPan) {
     });
 }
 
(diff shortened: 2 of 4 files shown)


src/ui/handler/mouse.js
```diff
@@ -50,6 +61,17 @@ class MouseHandler {
         if (!lastPoint) return;
         e.preventDefault();
 
+        if (buttonStillPressed(e, this._eventButton)) {
+            // Some browsers don't fire a `mouseup` when the mouseup occurs outside
+            // the window or iframe:
+            // https://github.com/mapbox/mapbox-gl-js/issues/4622
+            //
+            // If the button is no longer pressed during this `mousemove` it have have
```
ryanhamley, line 69: typo: `it have have`

src/ui/handler/mouse.js
```diff
@@ -6,6 +6,17 @@ import type Point from '@mapbox/point-geometry';
...
 
+// the values for each button in MouseEvent.buttons
+const BUTTONS_FLAGS = {
+    [LEFT_BUTTON]: 1,
+    [RIGHT_BUTTON]: 2
+};
+
+function buttonStillPressed(e: MouseEvent, button: number) {
+    const flag = BUTTONS_FLAGS[button];
+    return e.buttons === undefined || (e.buttons & flag) !== flag;
```
ryanhamley, line 17: Is it possible to press the left button, move the mouse outside the window, release the button, then click the left button again and bring it back inside the window so that this isn't triggered? Or does releasing the button always make `e.buttons` undefined?

ansis: > Is it possible to press the left button, move the mouse outside the window, release the button, then click the left button again and bring it back inside the window so that this isn't triggered?
I think it might be possible but I wasn't able to trigger it. With both ctrl-click and right click that starts outside the browser it only starts firing mousemoves back in the window after mouseup. But I think if it did happen it should be ok. The gesture would continue where the user left off.

> Or does releasing the button always make e.buttons undefined?

`e.buttons` got added in safari in 11. To handle those browsers more gracefully it's saying "yes the button is pressed" in those cases even if we can't tell. If this bug were present in those browsers it would remain but everything else would still work.

ansis: And as you noted it's not present in android chrome at all. This would handle that gracefully if you had a mouse connected to Android.

ansis:
> How reliable is this approach? It looks like we've been using the DOM.mouseButton() method for a long time so I guess it's pretty solid, but MDN is pretty explicit about MouseEvent.button being unreliable for some mouse events. It can also be reconfigured by some users though I guess that's a probably a rare edge case. 

We've used `.button` for a long time and I think it's been reliable for mousedown and mouseup. Users configuring their buttons differently would change which button triggers this but that should be fine. It should be whatever is their "secondary button".

> MouseEvent.buttons is not supported on Android Chrome or Android Webview though.

In this case the check treats the button as always still being pressed, which wouldn't fix this issue but wouldn't break anything futher.

src/ui/handler/mouse.js
ansis: fixed in 685c28880

ansis:
@ryanhamley could you rereview this?

PR #9521, opened 2020-04-09
[release-caspian] Fix style-spec isolation for within expression

### LINKED ISSUE #9524: @mapbox/mapbox-gl-style-spec references unbundled modules => build-time error
When referencing modules that are not explicitly exported in /dist, as Studio does here and there to enable advanced features, I get a build-time error.

image

It looks like `within.js` is in fact referencing a data constant that is not bundled with the style spec package itself. https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/expression/definitions/within.js#L11

**mapbox-gl-js version**: 1.9.1

**browser**: N/A

### Steps to Trigger Behavior

 1. Import `@mapbox/mapbox-gl-style-spec`.
 2. Attempt to reference the within expression definition.

### Link to Demonstration

Creating an online example of this was difficult, but Mapbox Studio is a current one. Happy to work with y'all on replicating this behavior.

### Expected Behavior

Possible to import modules from the src directory of `@mapbox/mapbox-gl-style-spec`.

### Actual Behavior

Importing modules from the src directory of `@mapbox/mapbox-gl-style-spec` references files that are not bundled with the package.

---

Posting here and tagging as `release-blocker` per advice from @arindam1993 🙇

arindam1993:
Fixes #9524

This fix also adds a check in the style-spec build script, that causes it to error out when an external file is imported.

## Launch Checklist

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix files within the style-spec package importing files from the parent repository tree. </changelog>`

### DIFF
--- a/src/style-spec/expression/definitions/within.js
+++ b/src/style-spec/expression/definitions/within.js
@@ -7,8 +7,6 @@ import type {Expression} from '../expression';
 import type ParsingContext from '../parsing_context';
 import type EvaluationContext from '../evaluation_context';
 import type {GeoJSON, GeoJSONPolygon, GeoJSONMultiPolygon} from '@mapbox/geojson-types';
-import MercatorCoordinate from '../../../geo/mercator_coordinate';
-import EXTENT from '../../../data/extent';
 import Point from '@mapbox/point-geometry';
 import type {CanonicalTileID} from '../../../source/tile_id';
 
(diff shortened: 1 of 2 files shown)


src/style-spec/expression/definitions/within.js
```diff
@@ -32,7 +40,7 @@ function boxWithinBox(bbox1: BBox, bbox2: BBox) {
 }
 
 function getTileCoordinates(p, canonical: CanonicalTileID) {
-    const coord = MercatorCoordinate.fromLngLat({lng: p[0], lat: p[1]}, 0);
+    const coord = {x: mercatorXfromLng(p[0]), y: mercatorYfromLat(p[1])};
```
mourner, line 43: Nit: if we're not using `MercatorCoordinate.fromLngLat`, there's no need to wrap the vlaues in an {x, y} object since it's immediately discarded.

karimnaaji:
@arindam1993 can you also make a PR that targets `release-danube` branch?

elifitch:
If this could be published under a beta tag, I'd be happy to pull it into studio and confirm that we're good.

arindam1993:
> If this could be published under a beta tag, I'd be happy to pull it into studio and confirm that we're good.

I don't think we should spam a public registry with iterative builds like that. But a private playground npm package to allow for sharing of WIP build is a good idea.

PR #9525, opened 2020-04-09
Add render-tests for line-dasharray case

tmpsantos:
Trick to mimic an outline with a different color.


tmpsantos:
Follow-up https://github.com/mapbox/mapbox-gl-js/pull/9532

PR #9550, opened 2020-04-13
More explicit files property in package.json

karimnaaji:
To prevent potential upload of erroneous files when publishing to npm.
Refer https://docs.npmjs.com/files/package.json\#files.


.npmignore
```diff
@@ -0,0 +1 @@
+.DS_Store
```
karimnaaji, line 1: From https://docs.npmjs.com/files/package.json#files: 

>You can also provide a .npmignore file in the root of your package or in subdirectories, which will keep files from being included. At the root of your package it will not override the “files” field, but in subdirectories it will

As I saw one of these in our 1.9.1 upload, https://unpkg.com/browse/mapbox-gl@1.9.1/src/.

package.json
```diff
@@ -160,8 +160,9 @@
   },
   "files": [
     "build/",
-    "dist/",
-    "flow-typed/",
+    "dist/mapbox-gl*",
+    "dist/style-spec/",
```
karimnaaji, line 164: Restricting the scope of `/dist` to prevent `versions.json`, `versions.jsonp` and any other dev files from being included.

package.json
```diff
@@ -160,8 +160,9 @@
   },
   "files": [
     "build/",
-    "dist/",
-    "flow-typed/",
+    "dist/mapbox-gl*",
```
asheemmamoowala, line 163: Since there are source maps, are `mapbox-gl-unminified` and `mapbox-gl-dev` required? If not, maybe this can be made more explicit

karimnaaji:
@mourner , ok that makes sense I just saw your comment in the chat, I'll remove the last commit https://github.com/mapbox/mapbox-gl-js/pull/9550/commits/15d9e87da6d02b4b277d6a5a20c2e1107b4a5886 to leave these out and see whether the rule could be made more strict!

karimnaaji:
After giving this more thoughts I think it would be a bit of a risk to have very strict rules, we may inconveniently have the opposite effect of not packaging things we want when adding or removing files. 

I opted for a slightly stricter set of rules along with a helper to see what files are about to be added and deleted. This step should be thoroughly followed during the release process to track any suspicious changes.

```
$ yarn run diff-tarball
19 files are about to be added in the new tarball
+ src/ui/handler/click_zoom.js
+ src/ui/handler/shim/dblclick_zoom.js
+ build/diff-tarball.js
+ src/ui/handler/shim/drag_pan.js
+ src/ui/handler/shim/drag_rotate.js
+ src/style/format_section_override.js
+ src/ui/handler_inertia.js
+ src/ui/handler_manager.js
+ src/ui/handler/handler_util.js
+ src/style-spec/expression/definitions/index_of.js
+ src/ui/handler/map_event.js
+ src/ui/handler/mouse.js
+ src/symbol/path_interpolator.js
+ src/style-spec/expression/definitions/slice.js
+ src/ui/handler/tap_drag_zoom.js
+ src/ui/handler/tap_recognizer.js
+ src/ui/handler/tap_zoom.js
+ src/ui/handler/touch_pan.js
+ src/ui/handler/shim/touch_zoom_rotate.js
174 files are about to be deleted in the new tarball
- dist/.DS_Store
- dist/mapbox-gl-unminified-master.js
- dist/versions.json
- dist/versions.jsonp
- src/.DS_Store
...
```

PR #9556, opened 2020-04-14
[docsprint] CameraOptions - update pitch and bearing definitions, add code examples

katydecorah:

Updates the JSDoc for `#cameraoptions` to include an inline code snippet, links to related examples, and updates the definition for `pitch` (borrowed from https://docs.mapbox.com/help/glossary/camera/) and updates the definition of `bearing` to match.

image

Fixes https://github.com/mapbox/mapbox-gl-js-docs/issues/141

### DIFF
--- a/src/ui/camera.js
+++ b/src/ui/camera.js
@@ -33,11 +33,26 @@ import type {PaddingOptions} from '../geo/edge_insets';
  * @typedef {Object} CameraOptions
  * @property {LngLatLike} center The desired center.
  * @property {number} zoom The desired zoom level.
- * @property {number} bearing The desired bearing, in degrees. The bearing is the compass direction that
- * is "up"; for example, a bearing of 90° orients the map so that east is up.
- * @property {number} pitch The desired pitch, in degrees.
+ * @property {number} bearing The desired bearing in degrees. The bearing is the compass direction that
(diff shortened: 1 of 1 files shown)


src/ui/camera.js
```diff
@@ -33,11 +33,26 @@ import type {PaddingOptions} from '../geo/edge_insets';
  * @typedef {Object} CameraOptions
  * @property {LngLatLike} center The desired center.
  * @property {number} zoom The desired zoom level.
- * @property {number} bearing The desired bearing, in degrees. The bearing is the compass direction that
- * is "up"; for example, a bearing of 90° orients the map so that east is up.
- * @property {number} pitch The desired pitch, in degrees.
+ * @property {number} bearing The desired bearing in degrees. The bearing is the compass direction that
+ * is "up". For example, `bearing: 90` orients the map so that east is up.
+ * @property {number} pitch The desired pitch in degrees. The pitch is the angle towards the horizon
```
andrewharvey, line 38: "angle towards the horizon"

maybe it's just me, but that leaves me confused since an angle is measure between two things, so I always think of this as the angle between the camera and the map plane's normal vector. 

> `pitch: 0` results in a two-dimensional map,
￼> * as if your line of sight forms a perpendicular angle with the earth's surface.

what about adding "ie. looking straight down on the map", just a suggestion :shrug:

PR #9559, opened 2020-04-14
[docsprint] Add detail to docs on setFeatureState, removeFeatureState, getFeatureState

colleenmcginnis:

Adds detail and examples on `setFeatureState`, `removeFeatureState`, and `getFeatureState`.

screencapture-localhost-8080-mapbox-gl-js-api-2020-04-13-19_20_16

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1912,22 +1912,21 @@ class Map extends Camera {
      * Sets the `state` of a feature.
      * A feature's `state` is a set of user-defined key-value pairs that are assigned to a feature at runtime.
      * When using this method, the `state` object is merged with any existing key-value pairs in the feature's state.
-     * Features are identified by their `feature.id` attribute, which must be an integer or a string that can be cast to an integer.
+     * Features are identified by their `feature.id` attribute, which can be any number or string.
+     * 
+     * This method can only be used with sources that have a `feature.id` attribute. The `feature.id` attribute can be defined in three ways: 
+     * - For vector or GeoJSON sources, including an `id` attribute in the original data file.
+     * - For vector or GeoJSON sources, using the [`promoteId`](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#vector-promoteId) option at the time the source is defined.
+     * - For GeoJSON sources, using the [`generateId`](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson-generateId) option to auto-assign an `id` based on the feature's index in the source data. If you change feature data using `map.getSource('some id').setData(..)`, you may need to re-apply state taking into account updated `id` values.
+     *
+     * _Note: You can use the [`feature-state` expression](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#feature-state) to access the values in a feature's state object for the purposes of styling._
      *
      * @param {Object} feature Feature identifier. Feature objects returned from
      * {@link Map#queryRenderedFeatures} or event handlers can be used as feature identifiers.
-     * @param {string | number} feature.id Unique id of the feature. Must be an integer or a string that can be cast to an integer.
+     * @param {string | number} feature.id Unique id of the feature.
      * @param {string} feature.source The id of the vector or GeoJSON source for the feature.
      * @param {string} [feature.sourceLayer] (optional) *For vector tile sources, `sourceLayer` is required.*
      * @param {Object} state A set of key-value pairs. The values should be valid JSON types.
-     *
-     *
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -1909,23 +1909,43 @@ class Map extends Camera {
 
     // eslint-disable-next-line jsdoc/require-returns
     /**
-     * Sets the state of a feature. The `state` object is merged in with the existing state of the feature.
-     * Features are identified by their `id` attribute, which must be an integer or a string that can be
-     * cast to an integer.
+     * Sets the `state` of a feature.
+     * A feature's `state` is a set of user-defined key-value pairs that are assigned to a feature at runtime.
+     * When using this method, the `state` object is merged with any existing key-value pairs in the feature's state.
+     * Features are identified by their `feature.id` attribute, which must be an integer or a string that can be cast to an integer.
```
asheemmamoowala, line 1915: > which must be an integer or a string that can be cast to an integer.

When using `promoteId` in a source definition, this is no longer a requirement. In that case the `feature.id` can be any string or number.

src/ui/map.js
```diff
@@ -1909,23 +1909,43 @@ class Map extends Camera {
...
+     * @param {string} feature.source The id of the vector or GeoJSON source for the feature.
+     * @param {string} [feature.sourceLayer] (optional) *For vector tile sources, `sourceLayer` is required.*
      * @param {Object} state A set of key-value pairs. The values should be valid JSON types.
      *
-     * This method requires the `feature.id` attribute on data sets. For GeoJSON sources without
-     * feature ids, set the `generateId` option in the `GeoJSONSourceSpecification` to auto-assign them. This
-     * option assigns ids based on a feature's index in the source data. If you change feature data using
-     * `map.getSource('some id').setData(..)`, you may need to re-apply state taking into account updated `id` values.
+     *
+     * This method can only be used with sources that have a `feature.id` attribute.
```
asheemmamoowala, line 1925: Or if the `promoteId` attribute is provided in the source definition.

src/ui/map.js
```diff
@@ -1909,23 +1909,43 @@ class Map extends Camera {
...
+     *       sourceLayer: 'my-source-layer',
+     *       id: e.features[0].id,
+     *     }, {
+     *       hover: true
+     *     });
+     *   }
+     * });
+     *
      * @see [Create a hover effect](https://docs.mapbox.com/mapbox-gl-js/example/hover-styles/)
+     * @see [Create interactive hover effects with Mapbox GL JS](https://docs.mapbox.com/help/tutorials/create-interactive-hover-effects-with-mapbox-gl-js/)
```
asheemmamoowala, line 1948: Should the link text indicate that this is a tutorial? The link text for both see-also's is very similar

src/ui/map.js
```diff
@@ -1934,38 +1954,84 @@ class Map extends Camera {
...
-     *  required.*
+     * @param {string} target.source The id of the vector or GeoJSON source for the feature.
+     * @param {string} [target.sourceLayer] (optional) *For vector tile sources, `sourceLayer` is required.*
      * @param {string} key (optional) The key in the feature state to reset.
+     *
+     * @example
+     * // Reset the entire state object for all features
+     * // in the `bar` source
+     * map.removeFeatureState({
+     *   source: 'my-source'
```
asheemmamoowala, line 1974: comment and code don't match

src/ui/map.js
colleenmcginnis: @asheemmamoowala I'm trying to wrap my head around this. Would something like this be true? 

>This method can only be used with sources have a `feature.id` attribute. The `feature.id` attribute can be defined in three ways: 
>- For vector sources or GeoJSON sources, an `id` attribute in the original data file.
>- For vector sources, using the [`promoteId`](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#vector-promoteId) option at the time the source is defined.
>- For GeoJSON sources, using the [`generateId`](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson-generateId) option to auto-assign them. This option assigns the `id` based on a feature's index in the source data. If you change feature data using `map.getSource('some id').setData(..)`, you may need to re-apply state taking into account updated `id` values.

asheemmamoowala: That looks good. Just a small correction:

> For vector sources, using the `promoteId` option at the time the source is defined.

This works for both vector and geojson sources.

colleenmcginnis: Wow. 🤯 Will update!

PR #9560, opened 2020-04-14
[docsprint] map.on: remove extra on section, link to event types, add code snippet, add related

danswick:
ℹ️ This PR is part of a larger effort to improve generated API documentation. It targets the docsprint branch, which will serve as the major feature branch for this work.

This PR makes a number of changes to the `map.on()` sections. It also raises some questions (below) about the best way to proceed. 

1. Deletes the extra `map.on()` section for calls that do not supply a `layerId`. I think it's sufficient to note that the `layerId` is optional and to stick with one section here. Very few (if any) of our examples do not use a `layerId` anyway.
2. Adds markdown-styled links to each event type. Unfortunately, there's a conflict between the way we do heading hashes and the way JSDoc assigns namespaces for events, so we can't use the JSDoc-style `{@link}` tag. 

### Questions

* Do the event types require any further narrative or inline code snippets? 
* Many of the event types include a `{@see}` tag, but others do not. I would still like to review these and make sure they're all accounted for. 
* Should there be more / different code snippets?

### DIFF
--- a/src/ui/events.js
+++ b/src/ui/events.js
@@ -344,6 +344,8 @@ export type MapEvent =
      * @memberof Map
      * @instance
      * @property {MapMouseEvent} data
+     * @see [Center the map on a clicked symbol](https://www.mapbox.com/mapbox-gl-js/example/center-on-symbol/)
+     * @see [Display a popup on click](https://docs.mapbox.com/mapbox-gl-js/example/popup-on-click/)
      */
     | 'mouseenter'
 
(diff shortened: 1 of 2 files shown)


danswick:
My original attempt did not take into account the fact that some events are compatible with `layerId` and others are not. I decided to go with a compatibility table that lists all of the (many, many) available event types: 

image

adrianababakanian:
Hi @danswick , small question: wondering if there is a reason why you decided to leave the "Compatible with `layerId`" column empty for events which are not compatible, rather than explicitly writing "no"? I supposed it is implied given that events which are compatible say "yes" (and a lot of "no"s listed out like that might not look the most aesthetically pleasing) but while scrolling through the API reference I found the many empty cells to look a bit unintentional, especially once I had scrolled past any visible "yes"s.

Very wordy way of asking -- what would be your thoughts if I opened a PR adding "no" to each of these empty cells, for the sake of explicitness? There are definitely pros and cons to each approach so I wanted to get your feedback before going for it 😄

danswick:
@adrianababakanian we discussed this during the final review. I tried putting "no" in the column, but found it difficult to parse the difference between "yes" and "no" visually at a glance. Maybe the benefit of being explicit is worth the reduction in glanceability (probably a word).

PR #9574, opened 2020-04-16
Fix possible undefined exception on paint variable of

mannnick24:
I've added an extra check for fill_extrusion_style_layer paint member being truthy. there seems to be a race condition that can leave it undefined.
I've added a couple of tests for the functions in the style layer to check that they bail out gracefully when paint is not set.


asheemmamoowala:
Superseded by #9658

PR #9577, opened 2020-04-16
[docsprint] Add inline examples for Point & PointLike types

adrianababakanian:

Adds inline code example for the `Point` and `PointLike` types. 

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -2582,10 +2582,15 @@ function removeNode(node) {
  * `x` and `y` properties representing screen coordinates in pixels.
  *
  * @typedef {Object} Point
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -2582,10 +2582,15 @@ function removeNode(node) {
  * `x` and `y` properties representing screen coordinates in pixels.
  *
  * @typedef {Object} Point
+ * @example
+ * var point = Point(-77, 38);
```
andrewharvey, line 2586: I believe that needs to be `new Point(...` and then actually needs to be called as `new mapboxgl.Point(...

PR #9578, opened 2020-04-16
[docsprint] map.on additional suggestions

colleenmcginnis:
Ref #9560 

ℹ️ This PR is part of a larger effort to improve generated API documentation. It targets the `docsprint-ds-map-on` branch, which will merge into the major feature branch for this work.

@danswick I decided to commit my suggestions separately as to not mess up your branch if you disagree with my suggestions. 💁‍♀️  My suggestions include:

- Adding a second example when the `layerId` param is optional.
- Adding `**Note:**` when the `layerId` param is optional.
- Adding `**Important:**` when the `layerId` param is required.
- Attempting to clarify the difference between `mouseover` and `mousemove`, which are currently described using the same language. (We might need some clarification from @asheemmamoowala on this.)

cc @katydecorah

### DIFF
--- a/src/ui/events.js
+++ b/src/ui/events.js
@@ -344,6 +344,8 @@ export type MapEvent =
      * @memberof Map
      * @instance
      * @property {MapMouseEvent} data
+     * @see [Center the map on a clicked symbol](https://www.mapbox.com/mapbox-gl-js/example/center-on-symbol/)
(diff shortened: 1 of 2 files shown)


danswick:
excellent suggestions! i screwed up the order of events in The Great Merging, so I cherry picked your commit over to the `doscprint` branch: https://github.com/mapbox/mapbox-gl-js/pull/9607.

PR #9589, opened 2020-04-17
[docsprint] Clarify meaning of Map#isSourceLoaded

adrianababakanian:

In the docstring for `Map#isSourceLoaded`, clarifies what "source is loaded" means. Uses the same definition as the docstring for the `MapDataEvent` type in `src/ui/events.js`, i.e.:
```
* @property {boolean} [isSourceLoaded] True if the event has a `dataType` of `source` 
* and the source has no outstanding network requests.
```

Screen Shot 2020-04-17 at 1 32 22 PM

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1393,7 +1393,8 @@ class Map extends Camera {
     }
 
     /**
-     * Returns a Boolean indicating whether the source is loaded.
+     * Returns a Boolean indicating whether the source is loaded. True if the source with
+     * the given ID in the map's style has no outstanding network requests.
      *
      * @param {string} id The ID of the source to be checked.
      * @returns {boolean} A Boolean indicating whether the source is loaded.
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -1393,8 +1393,8 @@ class Map extends Camera {
     }
 
     /**
-     * Returns a Boolean indicating whether the source is loaded. True if the source with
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -1393,7 +1393,8 @@ class Map extends Camera {
     }
 
     /**
-     * Returns a Boolean indicating whether the source is loaded.
+     * Returns a Boolean indicating whether the source is loaded. True if the source with
```
andrewharvey, line 1396: should we format that True as `true`?

adrianababakanian: There appears to be a bit of inconsistency in the API reference -- sometimes boolean values are formatted as `true` and other times as "true". I prefer the `true` format as well, but went with "True" due to capitalization. 

I am open to changing it/hearing with others think!

ryanhamley: I think we should standardize on `true` as it's a value within Javascript. In this particular case, we can sidestep the capitalization worries by rewording slightly. How about something like "Returns `true` if the source is loaded, otherwise `false`"?

adrianababakanian: That makes sense @ryanhamley -- I made a small update. It's slightly wordier than your suggestion so as to still address the initial feedback motivating this PR (i.e. what does it mean for a source to be loaded). This means there are now two sentences starting with "Returns...", but the clarity might be worth the slight redundancy here.

PR #9590, opened 2020-04-17
Clean check in `prepublishOnly`

karimnaaji:
Revised release process after running through it again manually. Runs clean check as a first step in `prepublishOnly`. This runs before the diffing with previously published tarball.


package.json
```diff
@@ -127,7 +127,7 @@
     "build-prod-min": "rollup -c --environment BUILD:production,MINIFY:true",
     "build-csp": "rollup -c rollup.config.csp.js",
     "build-query-suite": "rollup -c test/integration/rollup.config.test.js",
-    "build-flow-types": "cp build/mapbox-gl.js.flow dist/mapbox-gl.js.flow && cp build/mapbox-gl.js.flow dist/mapbox-gl-dev.js.flow",
+    "build-flow-types": "mkdir -p dist && cp build/mapbox-gl.js.flow dist/mapbox-gl.js.flow && cp build/mapbox-gl.js.flow dist/mapbox-gl-dev.js.flow",
```
asheemmamoowala, line 130: The `build-flow-types` build step is also used in our CI build pipeline, but in a different order:

https://github.com/mapbox/mapbox-gl-js/blob/ced93c1624e37e618024f621a1744ff24b279c26/.circleci/config.yml#L152-L167

I suggest moving this into the `publish-clean-check` step. It belongs there as part of the `clean` up

package.json
```diff
@@ -141,6 +141,7 @@
     "start-bench": "run-p build-token watch-benchmarks start-server",
     "start-release": "run-s build-token build-prod-min build-css print-release-url start-server",
     "diff-tarball": "build/run-node build/diff-tarball && echo \"Please confirm the above is correct [y/n]? \"; read answer; if [ \"$answer\" = \"${answer#[Yy]}\" ]; then false; fi",
+    "publish-clean-check": "git clean -fdx && yarn install",
```
asheemmamoowala, line 144: nit: This step does not check anything. Consider renaming to a clearer name, like `prepare-for-publish`

package.json
karimnaaji: I prefer to leave it here so it's not order dependent like in our CI; the second part of this command depends on `dist` already existing since it copies into it. Our CI runs `yarn run build-style-spec` [which creates `dist/style-spec/`](https://github.com/mapbox/mapbox-gl-js/pull/9590/files/f8708f37d12bc5e1e0e2016a16843f7a2ad8a8d0#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R132) folder and `yarn run build-flow-types` runs after, so this folder happens to exist in this order only with a clean checkout.

package.json
karimnaaji: Updated in https://github.com/mapbox/mapbox-gl-js/pull/9590/commits/69b4df144c110dcef7cf1570f3140e0bd3f78d86.

PR #9592, opened 2020-04-17
[docsprint] Add inline snippet to marker#addTo

HeyStenson:
Adds inline code snippet for [marker#addTo](https://docs.mapbox.com/mapbox-gl-js/api/#marker#addto), and makes small edits to its description.

## briefly describe the changes in this PR

Screen Shot 2020-04-17 at 2 41 37 PM

cc @asheemmamoowala @danswick @katydecorah

### DIFF
--- a/src/ui/marker.js
+++ b/src/ui/marker.js
@@ -216,7 +216,7 @@ export default class Marker extends Evented {
     }
 
     /**
-     * Attaches a `Marker` to a `Map` object.
+     * Attaches the `Marker` to a `Map` object.
      * @param {Map} map The Mapbox GL JS map to add the marker to.
      * @returns {Marker} `this`
      * @example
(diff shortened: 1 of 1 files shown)


src/ui/marker.js
```diff
@@ -216,9 +216,13 @@ export default class Marker extends Evented {
     }
 
     /**
-     * Attaches the marker to a map
+     * Attaches a `Marker` to a `Map` object.
```
andrewharvey, line 219: just nit picking here, but addTo doesn't add any `Marker`, it only adds `this` `Marker`, so should that be:
_attaches **the** `Marker` to_...

maybe I'm overthinking it.

HeyStenson: > maybe I'm overthinking it.

@andrewharvey Not at all! That distinction makes it a lot clearer - I will update the PR. Thank you!

PR #9602, opened 2020-04-20
Add tests for `distance` expression

zmiao:
Gl-native implementation: https://github.com/mapbox/mapbox-gl-native/pull/16397


test/integration/render-tests/distance/paint-polygon/style.json
```diff
@@ -0,0 +1,1480 @@
+{
+  "version": 8,
+  "metadata": {
+    "test": {
+      "width": 500,
```
asheemmamoowala, line 5: @zmiao This is quite a large size for a render test. Would it be possible to make a smaller test image?

test/integration/expression-tests/distance/point-to-line-distance/test.json
```diff
@@ -0,0 +1,94 @@
...
+          }
+        }]
```
chloekraw, line 11: I think there's an extra set of close brackets here that is causing problems with this file

test/integration/render-tests/distance/paint-polygon/style.json
zmiao: The intention of this test case is to simulate a real world route and to see if it works. That's why the image size is very big, but I will try to decrease the zoom level and minimize the size.

zmiao:
@chloekraw Thank you very much for the thorough review and creating such a clear table! I agree that the tests are insufficient, in order to prove the equivalent evaluation result between the feature and input data, more test cases need to be added. I am working on it!

test/integration/render-tests/distance/paint-polygon/style.json
chloekraw: Thanks for the context @zmiao. If it's OK to have one larger size image, I do think it would be valuable to have at least one render test that accurately reflects the real world route line use case.

Issue #9605, opened 2020-04-20
map.setFilter() bug with 'within' expression within 'all' expression


dpdiliberto:
I created an example by going off [this example](https://bl.ocks.org/ryanbaumann/fd8698e7eebf39f1c20c8fdcd8c8e80e/4eaaedf1ea53cc7ba2e8ea4fcb3bad70416bb7cf) to test setting a filter based on 1) features within a certain polygon using the `within` expression, as well as 2) features that met certain data conditions using the `in` expression. 

[My example](https://jsbin.com/yoqunojelu/edit?html,output) works with the `within` expression or `in` expression individually, but not when they are both within an `all` expression. Here is my filter expression that does not work:

`map.setFilter('population', ['all', ['in', 'Hispanic', ['get', 'ethnicity']], ['within', feature]]);`

I was able to get this to work with the following workaround:

`map.setPaintProperty('population', 'circle-opacity' , ['to-number', ['all', ['within', feature], ['==', ['get', 'ethnicity'], 'Hispanic']]]);`

Is this a bug or expected behavior?

**mapbox-gl-js version**: 1.9.0

**browser**: Chrome Version 80.0.3987.163

### Steps to Trigger Behavior

 1. `map.setFilter('population', ['all', ['in', 'Hispanic', ['get', 'ethnicity']], ['within', feature]]);` does not work
 2.`map.setFilter('population', ['within', feature]);` and `map.setFilter('population', ['in', 'Hispanic', ['get', 'ethnicity']]);` work individually
 3. I was able to work around this with @arindam1993 's help with `map.setPaintProperty('population', 'circle-opacity' , ['to-number', ['all', ['within', feature], ['==', ['get', 'ethnicity'], 'Hispanic']]]);`

### Link to Demonstration

https://jsbin.com/yoqunojelu/edit?html,output

### Expected Behavior

Being able to filter for both conditions: 1) being within a polygon and 2) certain data conditions

### Actual Behavior

It is not possible to filter for both when using the `within` expression.

arindam1993:
Hi @zmiao , would you happen to know what can cause this off the top of your head?

zmiao:
@dpdiliberto @arindam1993 I am trying to debug this issue, but for both pages I got access errors for the isochone API: Failed to load resource: the server responded with a status of 401 (Unauthorized). Can you help me checking this?

zmiao:
[...]

PR #9611, opened 2020-04-21
Fix using `within` expression with complex `filter`

zmiao:
Fix: https://github.com/mapbox/mapbox-gl-js/issues/9605

## Launch Checklist

 - [n/a] post benchmark scores
 - [n/a] tagged `@mapbox/map-design-team` `@mapbox/static-apis` if this PR includes style spec API or visual changes
 - [n/a] tagged `@mapbox/gl-native` if this PR includes shader changes or needs a native port
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix bug when using a combined expression of `within` and other expressions as a filter.</changelog>`


src/style-spec/feature_filter/index.js
```diff
@@ -96,7 +94,19 @@ function compare(a, b) {
...
 
+function geometryNeeded(filter) {
+    if (!Array.isArray(filter)) return false;
+    let needed = false;
+    for (let index = 0; index < filter.length; index++) {
+        const val = filter[index];
+        const op = Array.isArray(val) ? val[0] : val;
+        needed = needed || (op === 'within');
+    }
+    return needed;
```
arindam1993, line 105: looks like this is a linear traversal of the first level of the filter, wouldn't that still break if the within expression is nested futher inside?

would DF/BF be better instead?

zmiao: Good point! Fixed, test is updated accordingly. Thank you!

arindam1993: ❤️ some clean recursion

PR #9618, opened 2020-04-22
Reverse tap-drag-zoom direction

### LINKED ISSUE #9612: decide on TapDragZoom direction before release
We need to figure out with dragging down zooms in our out for the one-finger tap-drag-zoom gesture.

> I just tested the new gestures and they feel great!
One thing that didn't feel right though was the TapDragZoom that I feel it should be inverted so that drag down zooms in instead of out. That's what users using the gesture on gmaps is used to and it's also how it works if you're using the two-finger scroll gesture on a touchpad. Did you discuss this internally?

Thanks for catching this so that we can be more conscious about this.

In Google maps dragging down zooms in.
In Apple maps dragging down zooms out.

In desktop browsers on a macbook scrolling down zooms in by default (but flipping scroll settings can change this). I'm not sure about other browsers.

cc @Neorth @bradjonesca

Neorth:
Closes #9612

### DIFF
--- a/src/ui/handler/tap_drag_zoom.js
+++ b/src/ui/handler/tap_drag_zoom.js
@@ -62,7 +62,7 @@ export default class TapDragZoomHandler {
             this._active = true;
 
             return {
-                zoomDelta: dist / -128
+                zoomDelta: dist / 128
             };
         }
     }


karimnaaji:
@ansis Can you give this a look? Does it align with the behavior you were going for?

Issue #9639, opened 2020-04-28
Rotation on Modern iPhone + Chrome


Equanimaster:
**mapbox-gl-js version**: 1.9.1

**browser**: iPhone - Google Chrome.

### Steps to Trigger Behavior

 1. Map container width: 100%, height: 100%, fill whole browser window.
 2. Open on iPhone on Chrome browser.
 3. Rotate multiple time phone to switch portrait to landscape mode.

### Expected Behavior
Mapbox should catch "orientationchange" event and trigger .resize();

### Actual Behavior
Mapbox is resizing container, but canvas remains small.
The size of canvas is the last visible area - new area.

![before_rotation](https://user-images.githubusercontent.com/42443808/80478208-dbbc5600-8955-11ea-9e53-698b0979eac2.jpg)

![after_rotation](https://user-images.githubusercontent.com/42443808/80478206-db23bf80-8955-11ea-97a9-a30cb948370e.jpg)

arindam1993:
Good point @Equanimaster !

We should bind `_onWindowResize` to the `orientationchange` event [here](https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/map.js#L438), so `trackResize` works for this use case.

The only caveat is, I'm not completely sure if the parent container has already resized when `orientationchange` gets fired.

Patch is welcome @Equanimaster !

Equanimaster:
I have hard-coded into my code this orientationchange  + resize() and its working like a charm.

kiraarghy:
Hey @arindam1993,

I've run into the same issues as @Equanimaster, I'm happy to pick this up and submit a PR given they haven't submitted anything yet.

However, looking at the mdn docs, the orientationchange event seems to be [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Window/orientationchange_event#Browser_compatibility), is that something the mapbox-gl-js team is happy with?

Thanks.

Equanimaster:
I was thinking about the same,
but I'm not sure that this is a good fix,
because double trigger of resize on other browsers,
could cause problems.
I've never contributed to a project like this on github,
if there's a guide I could try commit my idea.

ryanhamley:
Closed by https://github.com/mapbox/mapbox-gl-js/pull/9652

PR #9652, opened 2020-05-04
kiraarghy:rotation-fix-iphone

kiraarghy:
In response to issue #9639.

Bug:

1. I go to my map on iPhone safari/ chrome.
2. I move my iPhone to landscape.
3. I move my iPhone to portrait.
4. The map does not resize to fit the portrait layout.

Added to map.js a screen orientation change event listener, this will trigger the `_windowResize method` so that when used on devices, orientation changes will trigger a resize. 

Unfortunately, the `orientationchange` event listener seems to be [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Window/orientationchange_event#Browser_compatibility) so to ensure future compatibility I have carried out this fix with the [Screen.Orientation API](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/onchange) with a fallback to the deprecated `orientationchange` event for browsers which don't support the Screen.Orientation API.

I've user tested this locally on a simulated and using a friend's iPhone and this seems to work but additional user testing would be very welcome.

I'm having some trouble getting my head around how testing for this project works, could someone assist me in that?

Thanks!

## Launch Checklist

 - [ ] post benchmark scores
 - [ ] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix for iPhone orientation change not triggering map resize by adding event listener for orientation change.</changelog>`

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -436,9 +436,11 @@ class Map extends Camera {
         if (typeof window !== 'undefined') {
             window.addEventListener('online', this._onWindowOnline, false);
             window.addEventListener('resize', this._onWindowResize, false);
-            if (typeof window.screen.orientation !== 'undefined') {
+            if (typeof window.screen.orientation === 'object') {
+                // ensures browser supports latest ScreenOrientation API
                 window.screen.orientation.addEventListener('change', this._onWindowResize, false);
             } else {
+                // If not then use deprecated 'orientationchange' event
                 window.addEventListener('orientationchange', this._onWindowResize, false);
             }
         }
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -436,6 +436,11 @@ class Map extends Camera {
         if (typeof window !== 'undefined') {
             window.addEventListener('online', this._onWindowOnline, false);
             window.addEventListener('resize', this._onWindowResize, false);
+            if (typeof window.screen.orientation !== 'undefined') {
+                window.screen.orientation.addEventListener('change', this._onWindowResize, false);
+            } else {
+                window.addEventListener('orientationchange', this._onWindowResize, false);
+            }
         }
 
         this.handlers = new HandlerManager(this, options);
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -436,6 +436,13 @@ class Map extends Camera {
         if (typeof window !== 'undefined') {
             window.addEventListener('online', this._onWindowOnline, false);
             window.addEventListener('resize', this._onWindowResize, false);
+            if (typeof window.screen.orientation === 'object') {
+                // ensures browser supports latest ScreenOrientation API
+                window.screen.orientation.addEventListener('change', this._onWindowResize, false);
```
andrewharvey, line 441: wherever an event listener is added on map create, it needs to also be removed on map remove, eg https://github.com/mapbox/mapbox-gl-js/blob/42ccea03b45d147628f261bca62c27fdace5eaa1/src/ui/map.js#L2566-L2569

kiraarghy:
I was looking into the issue @Equanimaster pointed out about the possibility of more resizing events being dispatched but in my profiling, this seems to result in the same number of calls to `_onWindowResize` as occurs in master currently.

Master:
Screenshot 2020-05-06 at 16 00 22

My branch: 
Screenshot 2020-05-06 at 16 07 26

I think left to do is the benchmark scores and tests, @andrewharvey can you help at all with the testing?

kiraarghy:
Hey is there any chance someone could have a look at this PR and help out?

src/ui/map.js
```diff
@@ -436,6 +436,13 @@ class Map extends Camera {
...
+            } else {
+                // If not then use deprecated 'orientationchange' event
+                window.addEventListener('orientationchange', this._onWindowResize, false);
```
mourner, line 444: As far as I can see, `window` `orientationchange` event [is a part of a living standard](https://developer.mozilla.org/en-US/docs/Web/API/Window/orientationchange_event), it's not deprecated. So you can get rid of the branching and only use this one. Perhaps it was mistaken for [screen orientationchange](https://developer.mozilla.org/en-US/docs/Web/API/Screen/onorientationchange), which is indeed deprecated?

kiraarghy: Yep mistake on my part, have included this change in my last commit.

PR #9654, opened 2020-05-04
Fix instrumentation for memory metrics

ryanhamley:
## Launch Checklist

    - this pulls in a change from https://github.com/mapbox/mapbox-gl-js/pull/9179 that's needed to run memory metrics against any branch

### DIFF
--- a/src/ui/map.js
+++ b/src/ui/map.js
@@ -2525,7 +2525,7 @@ class Map extends Camera {
             this.fire(new Event('idle'));
         }
 
-        if (!this._fullyLoaded) {
+        if (this._loaded && !this._fullyLoaded && !somethingDirty) {
             this._fullyLoaded = true;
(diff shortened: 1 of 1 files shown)


src/ui/map.js
```diff
@@ -2522,13 +2522,14 @@ class Map extends Camera {
...
             this.triggerRepaint();
         } else if (!this.isMoving() && this.loaded()) {
-            if (!this._fullyLoaded) {
-                this._fullyLoaded = true;
-                PerformanceUtils.mark(PerformanceMarkers.fullLoad);
-            }
             this.fire(new Event('idle'));
         }
 
+        if (!this._fullyLoaded) {
```
arindam1993, line 2528: I think this check needs to be `this._loaded && !this._fullyLoaded && !somethingDirty`

ryanhamley: fixed

PR #9655, opened 2020-05-05
[docs] Add distance expression entry; update other expressions

chloekraw:
Screen Shot 2020-05-08 at 12 34 26 AM
Screen Shot 2020-05-08 at 12 34 46 AM
Screen Shot 2020-05-08 at 12 47 47 AM
Screen Shot 2020-05-08 at 12 47 35 AM
Screen Shot 2020-05-08 at 12 35 25 AM
Screen Shot 2020-05-08 at 3 23 16 PM
Screen Shot 2020-05-08 at 3 23 20 PM

### Changelog:
- `distance`: 
   - added entry and SDK support table
- `within`: 
   - re-organized entry to increase clarity and conciseness
- `format`: 
   - mentioned the `image` expression explicitly in the documentation entry
   - re-formatted with bullet points
   - added version support information for Android SDK
- `index-of`: 
   - reworded entry to match language of `in` expression
   - removed misleading usage of "second argument," which doesn't match how current docs for [`get`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#get) and [`has`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#has) use "second argument." Considering `["index-of", needle, haystack, fromIndex]`, it's the third argument that can be optionally provided to configure the start index of the search.
- `slice`: 
   - reworded entry to match language of `in` expression
   - moved group from `String` to `Lookup` because `slice` is closer to the `in` and `index-of` expressions in function than other expressions in the `String` group. No other expressions in the `String` group work on arrays.
   - removed misleading usage of "second argument." Considering `["slice", input, startIndex, endIndex]`, it's the third argument that can be optionally provided to configure the end index.
- `match`: 
   - minor nitpick to match other entries which use `-` instead of `*` for formatting bullet points
- `interpolate`: no changes made

cc @lbutler @zmiao @alexshalamov @1ec5

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -695,10 +695,16 @@
       "doc": "Sorts features in ascending order based on this value. Features with a higher sort key will appear above features with a lower sort key.",
       "sdk-support": {
         "basic functionality": {
-          "js": "1.2.0"
+          "js": "1.2.0",
+          "android": "9.2.0",
+          "ios": "5.9.0",
+          "macos": "0.16.0"
         },
         "data-driven styling": {
-          "js": "1.2.0"
+          "js": "1.2.0",
+          "android": "9.2.0",
+          "ios": "5.9.0",
+          "macos": "0.16.0"
         }
       },
       "expression": {
@@ -1065,17 +1071,17 @@
       "type": "enum",
       "values": {
         "auto": {
-          "doc": "If `symbol-sort-key` is set, sort based on that. Otherwise sort symbols by their y-position relative to the viewport."
+          "doc": "Sorts symbols by `symbol-sort-key` if set. Otherwise, sorts symbols by their y-position relative to the viewport if `icon-allow-overlap` or `text-allow-overlap` is set to `true` or `icon-ignore-placement` or `text-ignore-placement` is `false`. If none of these options are set, no sorting is applied; symbols are rendered in the same order as the source data."
         },
         "viewport-y": {
-          "doc": "Symbols will be sorted by their y-position relative to the viewport."
+          "doc": "Sorts symbols by their y-position relative to the viewport if `icon-allow-overlap` or `text-allow-overlap` is set to `true` or `icon-ignore-placement` or `text-ignore-placement` is `false`."
         },
         "source": {
-          "doc": "Symbols will be rendered in the same order as the source data with no sorting applied."
+          "doc": "Sorts symbols by `symbol-sort-key` if set. Otherwise, no sorting is applied; symbols are rendered in the same order as the source data."
         }
       },
       "default": "auto",
-      "doc": "Controls the order in which overlapping symbols in the same layer are rendered",
+      "doc": "Controls the order in which symbols in the same layer are rendered.",
       "sdk-support": {
         "basic functionality": {
           "js": "0.49.0",
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -1081,7 +1081,7 @@
         }
       },
       "default": "auto",
-      "doc": "Controls the order in which symbols in the same layer are rendered.",
+      "doc": "Determines whether overlapping symbols in the same layer are rendered in the order that they appear in the data source or by their y-position relative to the viewport. To control the order and prioritization of symbols otherwise, use `symbol-sort-key`.",
       "sdk-support": {
         "basic functionality": {
           "js": "0.49.0",
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2621,7 +2621,7 @@
         }
       },
       "index-of": {
-        "doc": "Returns the first index at which an item can be found in an array or a substring can be found in a string, or `-1` if the input is not found. If set, the second argument is used as the start index.",
+        "doc": "Returns the first position at which an item can be found in an array or a substring can be found in a string, or `-1` if the input cannot be found. Accepts an optional index from where to begin the search.",
         "group": "Lookup",
         "sdk-support": {
           "basic functionality": {
@@ -2630,7 +2630,7 @@
         }
       },
       "slice": {
-        "doc": "Returns an item from an array or a substring from a string defined by the first argument as the start index. If set, the second argument is used as the end index. The return value is inclusive of the start index but not of the end index.",
+        "doc": "Returns an item from an array or a substring from a string from a specified start index, or between a start index and an end index if set. The return value is inclusive of the start index but not of the end index.",
         "group": "Lookup",
         "sdk-support": {
           "basic functionality": {
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -2630,7 +2630,7 @@
         }
       },
       "slice": {
-        "doc": "Returns an item from an array or a substring from a string defined by the first argument as the start index and a second argument, if set, as the end index. The expression is inclusive of the start index, but not of the end index.",
+        "doc": "Returns an item from an array or a substring from a string defined by the first argument as the start index. If set, the second argument is used as the end index. The return value is inclusive of the start index but not of the end index.",
         "group": "Lookup",
         "sdk-support": {
           "basic functionality": {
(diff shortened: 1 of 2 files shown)


chloekraw:
@colleenmcginnis I tried to push my branch to our staging server but it didn't work. Not sure if it's possible for this part of our documentation?

src/style-spec/reference/v8.json
```diff
@@ -3345,6 +3355,17 @@
           }
         }
       },
+      "distance": {
+        "doc": "Returns the shortest distance in meters between the evaluated feature and the input geometry. The evaluated feature and input value can be a valid GeoJSON of type `Point`, `MultiPoint`, `LineString`, `MultiLineString`, `Polygon`, `MultiPolygon`, `Feature`, or `FeatureCollection`. Results at zoom level 12 and below may vary in precision due to precision loss in conversions between tile coordinates and `[Longitude, Latitude]` at lower zoom levels.",
```
chloekraw, line 3359: > Results at zoom level 12 and below may vary in precision due to precision loss in conversions between tile coordinates and `[Longitude, Latitude]` at lower zoom levels."

Ah, I could simplify this part to "[encoding geometries](https://docs.mapbox.com/vector-tiles/specification/#encoding-geometry)". Maybe something like:

Distance values returned at zoom level 12 and below may vary in precision due to loss in precision from encoding geometries.

colleenmcginnis:
>@colleenmcginnis I tried to push my branch to our staging server but it didn't work. Not sure if it's possible for this part of our documentation?

It's possible, but it's not something we do often. It would require updating the submodule in /mapbox-gl-js-docs to this commit and then pushing to `publisher-staging` from there. As far as I know, this repo is no longer connected to publisher at all.

If you were just pushing to staging to check the formatting, that's something I (or likely someone from the GL JS team) could do as a part of a PR review. In the future we could pair on something like this, but unfortunately this week is a busy one for our team so if this is an urgent change I'd recommend asking for a reviewer to check formatting locally instead.

src/style-spec/reference/v8.json
```diff
@@ -2816,7 +2825,7 @@
         }
       },
       "format": {
-        "doc": "Returns `formatted` text containing annotations for use in mixed-format `text-field` entries. For a `text-field` entries of a string type, following option object's properties are supported: If set, the `text-font` value overrides the font specified by the root layout properties. If set, the `font-scale` value specifies a scaling factor relative to the `text-size` specified in the root layout properties. If set, the `text-color` value overrides the color specified by the root paint properties for this layer.",
+        "doc": "Returns a `formatted` string for use in the `text-field` property. The input may contain a plain `string`, an `image` provided through the [`'image'`](#types-image) expression, or a combination  of the two. Supported styling options:\n- `\"text-font\"`: Overrides the font stack specified by the root layout property.\n- `\"text-color\"`: Overrides the color specified by the root paint property.\n- `\"font-scale\"`: Applies the specified a scaling factor relative to the `text-size` specified by the root layout property.",
```
asheemmamoowala, line 2828: Removing the `mixed-format` makes it harder to understand the purpose of this expression type and when to use it. It's also not clear from the description how to pair the sections with their override options.

```suggestion
        "doc": "Returns a `formatted` string for displaying mixed-format text in the `text-field` property. The input may contain a string literal or expression, or an  [`'image'`](#types-image) expression, or a combination  of the two. Strings can be followed with a style override object that supports the following properties:\n- `\"text-font\"`: Overrides the font stack specified by the root layout property.\n- `\"text-color\"`: Overrides the color specified by the root paint property.\n- `\"font-scale\"`: Applies the specified number as a scaling factor relative to the `text-size` specified by the root layout property.",
```

src/style-spec/reference/v8.json
asheemmamoowala, line 3359: > The evaluated feature and input value can be a valid GeoJSON of type

Does this imply that the evaluated feature cannot be part of a vector tile?

> `Point`, `MultiPoint`, `LineString`, `MultiLineString`, `Polygon`, `MultiPolygon`, `Feature`, or `FeatureCollection`

Include details on what distance is measured among some of the pairs and what happens withe `Multi*` and `FeatureCollection` types?

cc @zmiao

src/style-spec/reference/v8.json
```diff
@@ -3345,6 +3355,17 @@
           }
         }
       },
+      "distance": {
+        "doc": "Returns the shortest distance in meters between the evaluated feature and the input geometry. The evaluated feature and input value can be a valid GeoJSON of type `Point`, `MultiPoint`, `LineString`, `MultiLineString`, `Polygon`, `MultiPolygon`, `Feature`, or `FeatureCollection`. Results at zoom level 12 and below may vary in precision due to precision loss in conversions between tile coordinates and `[Longitude, Latitude]` at lower zoom levels.",
+        "group": "Math",
```
asheemmamoowala, line 3360: What do you think of adding a new `Geospatial` group for `within` and `distance`?

src/style-spec/reference/v8.json
chloekraw: Hmm, good point, it shouldn't imply that. I'm open to better suggestions! 

> Include details on what distance is measured among some of the pairs and what happens withe Multi* and FeatureCollection types?

Yeah, good idea.

src/style-spec/reference/v8.json
chloekraw: I think we can always do this later if we find ourselves adding more geospatial expressions. Generally I'm hesitant to create a new category for fewer than 3 members, and `within` fits in the Decision group pretty well.  I agree, `distance` is a bit more awkward in the Math group, but I think it works fine enough.

src/style-spec/reference/v8.json
chloekraw: @zmiao for the input data, is what I wrote correct about what’s supported?

Yeah, maybe omitting “evaluated” feature from the description could work. What do you think about Asheem’s suggestion to add more detail on which distance is measured and the behavior for Multi* types? Maybe you could provide the info here and we could evaluate whether it’s worth including

asheemmamoowala:
Noting here that the the unit test failure is from a missing implementation:
```
test/unit/style-spec/expression.test.js .............. 9/10
  v8.json includes all definitions from style-spec
  not ok should be equivalent
    --- wanted
    +++ found
       "coalesce"
       "collator"
       "concat"
       "cos"
    -  "distance"
       "downcase"
       "e"
       "feature-state"
       "floor"
    at:
      line: 14
      column: 7
      file: test/unit/style-spec/expression.test.js
      type: Test
```

src/style-spec/reference/v8.json
chloekraw: @zmiao @asheemmamoowala which do you think is better?

asheemmamoowala: > Distance values returned at zoom level 12 and below may vary in precision due to loss in precision from encoding geometries.

☝️

zmiao: > for the input data, is what I `wrote` correct about what’s supported?
Yes, it is correct.

> What do you think about Asheem’s suggestion to add more detail on which distance is measured for different combinations and the behavior for Multi* types? 

I think it is a good idea to explain it, I will provide the details soon.

chloekraw:
@mapbox/gl-js I added screenshots previewing all of the expressions doc updates in the first post of this ticket. This PR pairs with [this branch](https://github.com/mapbox/mapbox-gl-js-docs/compare/ck-update-expressions-metadata) in mapbox-gl-js-docs. 

@ryanhamley I wasn't sure about the "needle / haystack" terminology in the [`in` expression metadata](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#in). It seems much more specific than other terms that we use -- I think when values can be multiple types we previously used generic terms like `InputType` and `OutputType` to stay away from non-generalizable details that are specific enough that they can (and do) serve as the variable names in the implementation.

It's a tough trade-off between clarity and consistency. I tried the version below first and I do think it's too messy and unclear with the multiple `|` and `,` so I'm thinking about changing it to `["in", InputType (boolean, string, or number), InputType (array or string)]`. Thoughts?

Screen Shot 2020-05-08 at 12 39 16 AM

Screen Shot 2020-05-08 at 12 35 30 AM

ryanhamley:
@chloekraw I think if that's what we've standardized the rest of the docs around then it's ok to make it more generic. I definitely prefer the `["in", InputType (boolean, string, or number), InputType (array or string)]` formulation to be clearer about the grouping of types. I think the only concern is that it may become confusing about which argument is which. That's partially what the `needle` and `haystack` nomenclature represents, but we could think of other ways to be more explicit about this distinction. Or maybe the implicit recognition by type is sufficient. It's explicitly documented as finding a value in a string or array and only the second argument supports those types.

TLDR: I'm not married to the `needle/haystack` terminology. It was mostly just to clarify the order of arguments.

chloekraw:
@ryanhamley thanks! yeah, I think that makes sense, and I do appreciate that your version is clearer about the usage. I think we can also achieve that goal through creating a GL JS example. I was trying to think of a way to add a sentence to the expression definition that could clarify the first vs. second argument, but I came up blank. I think the first argument could be described as a e.g. "search query" but I really couldn't think of anything for the second argument. If you have a suggestion, let me know!

chloekraw:
@asheemmamoowala @ryanhamley @ansis @1ec5 @zmiao @alexshalamov thanks for all the feedback! I'm going to merge this and cherrypick to `release-danube` now, but happy to revisit any of the changes in the future.

PR #9658, opened 2020-05-05
set an interim paint in style_layer

mannnick24:
this is in place of pr https://github.com/mapbox/mapbox-gl-js/pull/9574

when a StyleLayer is created its paint member is undefined untill recalculate is called. there are render scenarios where paint is accessed between construction and recalculate being called first time

see https://github.com/mapbox/ibm-cognos-collab/issues/154

Instead of checking whether paint is undefined in a layer when it is used, set an interim possiblyEvaluated value if there are any paint properties.

There is a very simple test for this

### DIFF
--- a/src/style/style_layer.js
+++ b/src/style/style_layer.js
@@ -10,7 +10,7 @@ import {
     emitValidationErrors
 } from './validate_style';
 import {Evented} from '../util/evented';
-import {Layout, Transitionable, Transitioning, Properties, PossiblyEvaluatedPropertyValue} from './properties';
+import {Layout, Transitionable, Transitioning, Properties, PossiblyEvaluated, PossiblyEvaluatedPropertyValue} from './properties';
 import {supportsPropertyExpression} from '../style-spec/util/properties';
 
 import type {FeatureState} from '../style-spec/expression';
@@ -90,6 +90,7 @@ class StyleLayer extends Evented {
         }
 
         if (properties.paint) {
+
             this._transitionablePaint = new Transitionable(properties.paint);
 
             for (const property in layer.paint) {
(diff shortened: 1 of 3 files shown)


mannnick24:
@asheemmamoowala there is a good use case and a fiddle on https://github.com/mapbox/ibm-cognos-collab/issues/153 
this was closed when I added a check on the layer.paint member
this change seeks to prevent paint from ever being undefined, by setting a default value. 
I'm adding a test that is really simple

mannnick24:
@asheemmamoowala I'm working on the idea that the race is between style.js update - which calls layer.recalculate -

src/style/style_layer.js
```diff
@@ -101,6 +102,13 @@ class StyleLayer extends Evented {
 
             this._transitioningPaint = this._transitionablePaint.untransitioned();
         }
+
+        if (properties.paint) {
+            // init default paint to cover until recalculate is called
+            // see https://github.com/mapbox/ibm-cognos-collab/issues/154
```
mourner, line 108: Nit: this is not a public issue so it shoudn't be included.

src/style/style_layer.js
```diff
@@ -101,6 +102,13 @@ class StyleLayer extends Evented {
 
             this._transitioningPaint = this._transitionablePaint.untransitioned();
         }
+
+        if (properties.paint) {
```
mourner, line 106: Nit: you can remove this `if` because there's the same condition above.

karimnaaji:
Merged as https://github.com/mapbox/mapbox-gl-js/commit/ef0884e019c5e294b49ad83267e810575e54f53a, will be part of v1.10.1. Thanks @mannnick24 !

Issue #9630, opened 2020-04-25
Inconsistent behaviours of “coalesce” with “image”


frodrigo:
According to the documentation, #8684 and #8052. “coalesce” should return the first existing “image”.

With last v1.9.1. Same behaviours on Firefox and Chrome.
https://js.do/code/432516

The style have icon for `start_11` and `oneway`, but not for ` not_existing`. Nevertheless, the flowing code return no icon.

```
    "icon-image": [
      "coalesce",
      ["image", "not_existing"],
      ["image", "star_11"]
    ],
```

The same code in text return `not_existing`:
```
    "text-field": [
      "string",
      [
        "coalesce",
        ["image", "not_existing"],
        ["image", "star_11"]
      ]
    ]
```

Playing more with coalesce give some “random” results

```
    "icon-image": [
      "coalesce",
      ["image", "not_existing"]
    ],
    "text-field": [
      "string",
      [
        "coalesce",
        ["image", "not_existing"]
      ]
    ]
```
Give unexpected ` not_existing` (expected null)

```
    "icon-image": [
      "coalesce",
      ["image", "star_11"]
    ],
    "text-field": [
      "string",
      [
        "coalesce",
        ["image", "star_11"]
      ]
    ]
```
Give the `star_11`.

```
      "coalesce",
      ["image", "not_existing"],
      ["image", "star_11"],
```
Give unexpected ` not_existing` (expected ` star_11`)

```
      "coalesce",
      ["image", "star_11"],
      ["image", "not_existing"]
```
Give the `star_11`.

```
      "coalesce",
      ["image", "not_existing"],
      "star_11"
```
Give the `star_11`.

```
      "coalesce",
      ["image", "not_existing"],
      ["image", "star_11"],
      ["image", "oneway"]
```
Give unexpected ` not_existing` (expected ` star_11`)

```
      "coalesce",
      ["image", "not_existing"],
      ["image", "star_11"],
      "oneway"
```
Give unexpected ` oneway` (expected ` star_11`)

arindam1993:
Capturing internal convo,

@ryanhamley found the issue. It's a regression introduced with https://github.com/mapbox/mapbox-gl-js/pull/9352/files#diff-0856eb2171400e111b6a65dde5e0c789R76

since the `canonical` param was introduced in the middle of the function signature, it seems like certain places that call it place the data in the `canonical`  param instead of `availableImages`.

@zmiao would you please be able to look at this? 🙏

[...]

PR #9668, opened 2020-05-08
Fix image expression

ryanhamley:
## Launch Checklist

 
This PR addresses two separate issues with the `image` expression operator.
- The changes in `src/style` address https://github.com/mapbox/mapbox-gl-js/issues/9630. https://github.com/mapbox/mapbox-gl-js/pull/9352 introduced an optional parameter in front of `availableImages` in the arguments list which accidentally broke `Layout` and `Transitioning` expression evaluation. This PR addresses that by adding the optional argument for those expression evaluation pathways.
- Fixing ⬆️led us to discover a separate issue in which `image` expressions on layers embedded directly into a style (rather than added at runtime) were being evaluated before `availableImages` was populated in the source's workers. The @mapbox/map-design-team reported the same issue in https://github.com/mapbox/mapbox-core-style-components/issues/917 although it wasn't clear at first that these two issues revealed the same bug.

Before https://github.com/mapbox/mapbox-core-style-components/issues/917:
Screen Shot 2020-05-07 at 6 32 02 PM
After:
Screen Shot 2020-05-07 at 6 31 08 PM

Before https://github.com/mapbox/mapbox-gl-js/issues/9630:
_Note: The "not_existing" text label is a fallback to demonstrate the absence of the icon_
Screen Shot 2020-05-07 at 8 39 28 PM
After:
Screen Shot 2020-05-07 at 8 38 48 PM

 - [ ] post benchmark scores
 - [x] tagged `@mapbox/map-design-team` `@mapbox/static-apis` if this PR includes style spec API or visual changes

Tests and a Flow fix are WIP

### DIFF
--- a/src/source/worker.js
+++ b/src/source/worker.js
@@ -77,6 +77,12 @@ export default class Worker {
 
     setImages(mapId: string, images: Array<string>, callback: WorkerTileCallback) {
         this.availableImages[mapId] = images;
+        for (const workerSource in this.workerSources[mapId]) {
+            const ws = this.workerSources[mapId][workerSource];
+            for (const source in ws) {
+                ws[source].availableImages = images;
+            }
+        }
         callback();
     }
 
--- a/src/source/worker_source.js
+++ b/src/source/worker_source.js
@@ -71,6 +71,7 @@ export type WorkerDEMTileCallback = (err: ?Error, result: ?DEMData) => void;
  * @param layerIndex
  */
 export interface WorkerSource {
+    availableImages: Array<string>,
     // Disabled due to https://github.com/facebook/flow/issues/5208
     // constructor(actor: Actor, layerIndex: StyleLayerIndex): WorkerSource;
 
--- a/src/style/properties.js
+++ b/src/style/properties.js
@@ -317,10 +317,10 @@ export class Transitioning<Props: Object> {
         this._values = (Object.create(properties.defaultTransitioningPropertyValues): any);
     }
 
-    possiblyEvaluate(parameters: EvaluationParameters, availableImages?: Array<string>): PossiblyEvaluated<Props> {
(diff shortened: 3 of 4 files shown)


ryanhamley:
It's not clear to me why [this existing render test](https://github.com/mapbox/mapbox-gl-js/blob/master/test/integration/render-tests/icon-image/image-expression/style.json) would not have caught these bugs. The setup to for the test seems to be exactly what caused these issues but that test never failed.

mourner:
> * introduced an optional parameter in front of `availableImages` in the arguments list which accidentally broke `Layout` and `Transitioning` expression evaluation

One more reason to eliminate things like `(this: any)` from the codebase as much as possible — Flow could have caught this.

> It's not clear to me why [this existing render test](https://github.com/mapbox/mapbox-gl-js/blob/master/test/integration/render-tests/icon-image/image-expression/style.json) would not have caught these bugs. The setup to for the test seems to be exactly what caused these issues but that test never failed.

Not sure but one guess would be that Node web workers mock is too fast to process messages (`postMessage` uses `setImmediate`), while a real browser takes a few ticks for the communication. One more good reason to explore render tests in the browser.

ryanhamley:
>One more reason to eliminate things like (this: any) from the codebase as much as possible — Flow could have caught this.

Perhaps I'm misunderstanding Flow's documentation but https://flow.org/en/docs/types/functions/#toc-optional-parameters says that optional parameters can be missing. Is there any built-in way in Flow to avoid this problem if you have multiple optional parameters in a function signature?

ryanhamley:
My best guess on why the existing render test didn't fail is that Node was able to load and parse the local sprite sheet fast enough to cover up this bug. In production, it takes longer to load a sprite sheet over the network and parse potentially hundreds of sprites. A browser test could help with this but I'm not sure how we could make use of Selenium to determine if the correct image has been loaded.

PR #9679, opened 2020-05-12
map.isMoving() match move events fix #9647

### LINKED ISSUE #9647: `Map.isMoving()` incorrectly returns true inside `mouseup` event handler
**mapbox-gl-js version**:
1.10.0

**browser**:
Chrome Version 81.0.4044.129 (Official Build) (64-bit)

It looks like there was a possible regression introduced in `mapbox-gl@1.10.0` that affects the result of `map.isMoving()` inside the `mouseup` event handler.

### Steps to Trigger Behavior

 1. Add a layer to the map, then set a `mouseup` event handler on it.
 2. Click on one of the layer's feature and call `map.isMoving()` in the handler.

### Link to Demonstration
🐞`mapbox-gl@1.10.0` - https://codesandbox.io/s/magical-leavitt-6zhso
✅ `mapbox-gl@1.9.1` - https://codesandbox.io/s/lucid-hawking-ruywl

### Expected Behavior
Clicking on Maine should output `map.isMoving() === false`, ie:

```
map.on('mouseup', statesLayerId, () => {
  map.isMoving() // false
}
```

### Actual Behavior
Clicking on Maine outputs `map.isMoving() === true`

ansis:
Previously isMoving() would return true if any interaction handler was active. Handlers are sometimes active even if they haven't changed the map yet. This resulted in the isMoving() returning true when the map hasn't moved.

This pr makes isMoving() aligned with movestart/move/moveend events. isMoving() returns true only:
- during a movestart
- after a movestart but before a moveend

Since move events may be fired after several mouse events have been batched, the camera changes a mouseevent will *later* cause won't be reflected in isMoving() when that mouseevent is fired.

For example, if you pan the map (and pause to eliminate inertia) a `mouseup` event will have `isMoving() === true` even though that event will later cause a `moveend` without further camera changes. This behavior existed in 1.9.1 (https://codesandbox.io/s/lucid-hawking-ruywl).

## Launch Checklist

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix bug where map.isMoving() returned true while map was not moving.</changelog>`

### DIFF
--- a/test/unit/ui/handler/drag_rotate.test.js
+++ b/test/unit/ui/handler/drag_rotate.test.js
@@ -273,6 +273,7 @@ test('DragRotateHandler ensures that map.isMoving() returns true during drag', (
 
     simulate.mousedown(map.getCanvas(), {buttons: 2, button: 2});
     simulate.mousemove(map.getCanvas(), {buttons: 2, clientX: 10, clientY: 10});
+    map._renderTaskQueue.run();
     t.ok(map.isMoving());
 
     simulate.mouseup(map.getCanvas(),   {buttons: 0, button: 2});
--- a/test/unit/ui/handler/drag_rotate.test.js
+++ b/test/unit/ui/handler/drag_rotate.test.js
@@ -277,6 +277,7 @@ test('DragRotateHandler ensures that map.isMoving() returns true during drag', (
     t.ok(map.isMoving());
 
     simulate.mouseup(map.getCanvas(),   {buttons: 0, button: 2});
+    map._renderTaskQueue.run();
     t.ok(!map.isMoving());
 
     map.remove();
--- a/test/unit/ui/map/isMoving.test.js
+++ b/test/unit/ui/map/isMoving.test.js
@@ -39,11 +39,17 @@ test('Map#isMoving returns true during a camera zoom animation', (t) => {
 test('Map#isMoving returns true when drag panning', (t) => {
     const map = createMap(t);
 
+    map.on('movestart', () => {
+        t.equal(map.isMoving(), true);
+    });
     map.on('dragstart', () => {
         t.equal(map.isMoving(), true);
     });
 
     map.on('dragend', () => {
+        t.equal(map.isMoving(), false);
+    });
+    map.on('moveend', () => {
         t.equal(map.isMoving(), false);
         map.remove();
         t.end();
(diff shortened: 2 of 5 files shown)


test/unit/ui/handler/drag_rotate.test.js
```diff
@@ -273,6 +273,7 @@ test('DragRotateHandler ensures that map.isMoving() returns true during drag', (
 
     simulate.mousedown(map.getCanvas(), {buttons: 2, button: 2});
     simulate.mousemove(map.getCanvas(), {buttons: 2, clientX: 10, clientY: 10});
+    map._renderTaskQueue.run();
```
karimnaaji, line 276: This particular test is now failing from CI, is this call actually processing the events correctly?

ansis: I was missing another one of these calls after the next event.

In old releases `map.isMoving()` would return true immediately after a `mousemove` event that causes a rotation. Now `map.isMoving()` only returns true on the next frame when the move events have fired and the camera events have changed. Does this seem ok to you?

test/unit/ui/map/isMoving.test.js
```diff
@@ -39,12 +39,18 @@ test('Map#isMoving returns true during a camera zoom animation', (t) => {
 test('Map#isMoving returns true when drag panning', (t) => {
     const map = createMap(t);
 
+    map.on('movestart', () => {
```
karimnaaji, line 42: Do you know if we have enough coverage around whether those events are triggered under a similar context?

karimnaaji: This test case is focused around `isMoving`. But what I'm thinking is covering whether we call the lambda function at all, if we fail to call that lambda it would go unnoticed.

Maybe collecting how many times those lambdas are called and testing that at the end of this block could be useful.

test/unit/ui/handler/drag_rotate.test.js
karimnaaji: I would not think that having a one frame delay difference for this type of API as being critical. As long as we still cover the following requirements:
>Returns true if the map is panning, zooming, rotating, or pitching due to a camera animation or user gesture.

test/unit/ui/map/isMoving.test.js
ansis: I think in this case there is enough coverage elsewhere but we could change the test to avoid this problem.

PR #9683, opened 2020-05-13
fix pinching while touching markers

### LINKED ISSUE #9675: Pinch-to-zoom broken when first touch is on a marker on touch devices
Since upgrading to v1.10.0, when a user uses a touch device (iPhone) and pinches to zoom on a map but there is a marker on their first touch (e.g. a marker is in the way), zoom will not happen at all. Only if user pinches to zoom outside the marker, the map will zoom.

This is new behavior and not expected I believe. Especially on a map with lots of markers (like my site: nomadlist.com/map), pinch zooming becomes very difficult as the user has to position both fingers outside any marker. It's also not how any other map API or app does it (e.g. Gmaps).

Example from Mapbox itself, open on a touch device like iPhone, zoom out, and try pinch zooming while first touch is on a marker:
https://docs.mapbox.com/mapbox-gl-js/example/custom-marker-icons/

ansis:
fix #9675

When touches touch the markers instead of the canvas they have different targets. Gestures were using `touchEvent.targetTouches` as the list of relevant touches. This broke multi-finger gestures that touch markers.

The fix switches to `touchEvent.touches` and filters out touches who's target is not a child element of the map's canvas container.

## Launch Checklist

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix multi-finger gestures that touch markers.</changelog>`

### DIFF
--- a/src/ui/handler/handler_util.js
+++ b/src/ui/handler/handler_util.js
@@ -2,7 +2,7 @@
 
 import assert from 'assert';
 
-export function indexTouches(touches: TouchList, points: Array<Point>) {
+export function indexTouches(touches: Array<Touch>, points: Array<Point>) {
     assert(touches.length === points.length);
     const obj = {};
     for (let i = 0; i < touches.length; i++) {
--- a/src/ui/handler/tap_drag_zoom.js
+++ b/src/ui/handler/tap_drag_zoom.js
@@ -30,7 +30,7 @@ export default class TapDragZoomHandler {
         this._tap.reset();
     }
 
-    touchstart(e: TouchEvent, points: Array<Point>) {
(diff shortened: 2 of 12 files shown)


karimnaaji:
Do you think it would be useful to create a page with a high density of randomly placed markers, and add it as part of our release testing [here](https://github.com/mapbox/mapbox-gl-js/blob/master/test/release/index.js#L8_L66)? (making sure we would inevitably touch one while doing a gesture with a density similar to where the issue was shown https://nomadlist.com/map).

We already have markers in the list of pages but I have the feeling it would be harder to test that fix from there.

ansis:
> Do you think it would be useful to create a page with a high density of randomly placed markers, and add it as part of our release testing here? 

Yep. The existing marker page might work if you intentionally try to test for this

Issue #9682, opened 2020-05-12
Better handling of line-gradient for step expression interpolation


karimnaaji:
`line-gradient` with `step` should produce a harder transition at color boundaries than when using linear expression interpolators. This can be achieved by either using a larger texture size when rendering the gradient or changing the texture sampler interpolation type to NEAREST. We need to evaluate both approaches to fit nav requirements https://github.com/mapbox/mapbox-navigation-ios and https://mapbox/mapbox-navigation-android.

![Screen Shot 2020-05-12 at 4 49 29 PM](https://user-images.githubusercontent.com/7061573/81756998-3aafbe00-9472-11ea-82ba-cde1cec27ff5.png)

karimnaaji:
Fixed in https://github.com/mapbox/mapbox-gl-js/pull/9694

PR #9694, opened 2020-05-15
Revise step interpolation for lines, increase precision of line progress as needed

karimnaaji:
Refer https://github.com/mapbox/mapbox-gl-js/issues/9682.

A nav native use case requires hard transition between different color stops while using the `line-gradient` property. We currently do not provide the best result for their use case. This PR is an attempt to reduce the limitations from this use case by increasing the precision and switching to a more appropriate texture sampler for the [`step`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#step) property:

>Produces discrete, stepped results by evaluating a piecewise-constant function defined by pairs of input and output values ("stops")

**Changes**
- Evaluate worst scenario coverage of line string and increase texture size resolution accordingly, by making sure we are not going over hardware size limits.
- Change the sampling to nearest: to maximize the results under most circumstances, the sampling is changed from linear to nearest for step interpolant. Without taking into consideration the other precision limitations, this gives better results overall.
- To limitate memory issues on their use case, reuse the memory placement of the color ramp if it was already created (also done for heatmaps). This allows to update the expression of the color ramp gradient without involving reallocation for memory that's already there: https://github.com/mapbox/mapbox-gl-js/commit/13779a5547d117321fe3c2a6a0e2643921ee4f58
- Increase the precision of line progress when we pass some threshold known to lack precision 2c7ad4685a61d74a43da6c86ec349a7f713d16a1. The memory cost is not involved in the common case.
- Tile line gradient textures and render them between the line clips of the current tile, this means the texture is now per tile instead of per layer. This was the best approach to achieve maximum accuracy across zoom levels and long routes.

**Differences from the above changes with master**

actual
expected
diff

**Approach comparisons**

(1) Instead using the texture to encode the color ramp, encode the expression stop in the texture and use glsl functions to evaluate in the shader.
The advantage of this approach would be to reduce the texture memory being used under this use case. The disavantage is the dependency on the expression and the shader code that would be interoduced. Here is how it would look visually (using the [step glsl function](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/step.xhtml)): [step-glsl.mov.zip](https://github.com/mapbox/mapbox-gl-js/files/4631636/step-glsl.mov.zip)
The results are not different enough from (2) to warrant the complexity.
(2) Using nearest with a very large texture (the approach of this PR): [16k-texture-nearest.mov.zip](https://github.com/mapbox/mapbox-gl-js/files/4631637/16k-texture-nearest.mov.zip)
(3) Using a very large texture while keeping linear as a texture sampler: [16k-texture-linear.mov.zip](https://github.com/mapbox/mapbox-gl-js/files/4631638/16k-texture-linear.mov.zip)

Overall this seems like the best we can do for the time being, we should evaluate having a different code path for this type of use case, as well as finding a way to increase texture uv precision over very large route lines (by splitting the geometry use colors per vertex, where vertices are deduplicated when colors are changing, with the drawback of using more memory).

## Launch Checklist

 - [n/a] post benchmark scores
 - [x] tagged @mapbox/map-design-team for visual changes
 - [x] tagged @mapbox/gl-native for native port
 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Revise step interpolation for line-gradient to result in harder visual transition between stops and support for long routes</changelog>`
- [x] Fix issue with gradient textures invalidation (https://github.com/mapbox/mapbox-gl-js/pull/9694/commits/754be74a3e35bc99d854b5272d866365585cd1b1)


src/style/style_layer/line_style_layer.js
```diff
@@ -59,14 +61,19 @@ class LineStyleLayer extends StyleLayer {
 
     _handleSpecialPaintPropertyUpdate(name: string) {
         if (name === 'line-gradient') {
-            this._updateGradient();
+            const expression: ZoomConstantExpression<'source'> = ((this._transitionablePaint._values['line-gradient'].value.expression): any);
+            this.stepInterpolant = expression._styleExpression.expression instanceof Step;
```
asheemmamoowala, line 65: There are cases where the top-level expression is not directly a `step` expression, even though this expression does not support dds.

For example:
```
['let', 
'my_var', ['+', 1, 1],
['step', .....
```
While, this is not required, it is allowed by the expression system.

karimnaaji: That's a good point, I will test this part to better account for detecting the expression.

karimnaaji:
>The only other question I have is if this is stable at high zooms with very long routes, say cross-continental

We're still limited for cross-continental route, and it's relatively not stable at high zooms. For that type of route, at zoom 18, we're trying to represent too many pixels with the current architecture so it would start to break because of that (~36 million for cross continental route at z18/19). This goes unnoticed with gradients but is not forgiving for a precise interpolant like step. We may also suffer jumps because line progress is encoded in 15 bits.

To support the long route case, a better approach would be to build a different vertex layout in line buckets, have different colors as part of the vertex layout, and inject splits as we evaluate the expression while building the line.

example

where vertices v0..v3 are given a color attribute, v4..v7 another, and the split (@ v2/v4) happens as the step expression is evaluated. I think we should give the above a try after we mitigate the issue with this proposal, it's more involved and we should be careful to not degrade the most common case.

For mid-long routes, I have a test that accounts for a rather long distance (before we start hitting precision issues).
```
test/integration/render-tests/line-gradient/gradient-step-zoomed/style.json
```
The geometry used was the following:

Screen Shot 2020-05-20 at 4 04 17 PM

karimnaaji:
Something I just thought about for very long routes is to increase the precision of v_lineprogress and create a second vertex buffer specifically for it, where it wouldn't be encoded in 15bits but either f64/32. This has a memory cost, but by having it in a separate vertex buffer, we'd allow to hit that cost only on cases where incoming geometry to be represented is known to not be enough. Worth trying for this PR!

src/style/style_layer/line_style_layer.js
karimnaaji: I've tried a few different combinations of the above to write a test for it but couldn't the expression to accept it, but I could be missing something. With what I tried, `line-gradient` seemed to be fairly restrictive in what it can ingest as input ([`ColorRampProperty`](https://github.com/mapbox/mapbox-gl-js/blob/5edbaf1a8186dfa0aad8c955ee958c09492ec855/src/style/style_layer/line_style_layer_properties.js#L63)): `layers.line.paint.line-gradient: color expected, array found`. We might be safe here.

karimnaaji:
I gave a quick try at https://github.com/mapbox/mapbox-gl-js/pull/9694#issuecomment-631821635 and it does better handle long routes. I'll take care of this as follow up work so we can start porting this PR to gl-native first.

karimnaaji:
Follow up of this PR implemented in https://github.com/mapbox/mapbox-gl-js/commit/03262f1c682568fe9ca6bd6fe85694e20cff031a for support of long routes. I haven't merge it in here now as we can't introduce shader related changes if we'd like to have that ported in gl-native in a reasonable amount of time (dependency on other port before we can update the submodule there).

karimnaaji:
@ansis Some comparisons done on a cross continental route to facilitate the review:

**new z4**
new-z4
**old z10**
master-z4


**new z10**
new-z10
**old z10**
master-z10

Videos for full frames (degraded by gifs):
[videos.zip](https://github.com/mapbox/mapbox-gl-js/files/4691592/videos.zip)

<details>
<summary>Path used for the above</summary>

```diff
diff --git a/debug/line-gradient.html b/debug/line-gradient.html
index 89b443e35..fccd06a1d 100644
--- a/debug/line-gradient.html
+++ b/debug/line-gradient.html
@@ -31,7 +31,7 @@
             "type": "Feature",
             "properties": {},
             "geometry": {
-                "coordinates": [[-77.044211, 38.852924 ], [-77.045659, 38.860158 ], [-77.044232, 38.862326 ], [-77.040879, 38.865454 ], [-77.039936, 38.867698 ], [-77.040338, 38.86943 ], [-77.04264, 38.872528 ], [-77.03696, 38.878424 ], [-77.032309, 38.87937 ], [-77.030056, 38.880945 ], [-77.027645, 38.881779 ], [-77.026946, 38.882645 ], [-77.026942, 38.885502 ], [-77.028054, 38.887449 ], [-77.02806, 38.892088 ], [-77.03364, 38.892108 ], [-77.033643, 38.899926 ] ],
+                "coordinates": [[-120.044211, 38.852924 ], [-77.045659, 38.860158 ] ],
                 "type": "LineString"
             }
         }]
@@ -46,6 +46,7 @@
             data: gj
         });
 
+        const routeCompleted = 0.3;
         map.addLayer({
             type: 'line',
             source: 'line',
@@ -54,22 +55,27 @@
                 'line-color': 'red',
                 'line-width': 14,
                 'line-gradient': [
-                    'interpolate',
-                    ['linear'],
+                    'step',
                     ['line-progress'],
-                    0, "rgba(0, 0, 255, 0)",
-                    0.1, "royalblue",
-                    0.3, "cyan",
-                    0.5, "lime",
-                    0.7, "yellow",
-                    1, "red"
+                    "rgba(0, 0, 255, 0.1)",
+                    1.0, "yellow"
                 ]
             },
-            layout: {
-                'line-cap': 'round',
-                'line-join': 'round'
-            }
         });
+
+        setInterval(function () {}, 1000);
+        let offset = 0.001;
+        function animate(timestamp) {
+            map.setPaintProperty('line', 'line-gradient',  [
+                    'step',
+                    ['line-progress'],
+                    "rgba(0, 0, 255, 0.1)",
+                    Math.cos(offset) * 0.5 + 0.5, "yellow"
+                ]);
+            offset += 0.001;
+            requestAnimationFrame(animate);
+        }
+        animate(0);
     });
 
 </script>

```

</details>

asheemmamoowala:
@karimnaaji What is the max precision possible after these changes? I find that the step expression supports a precision of ~0.0001. The outcome is that on the cross-continental route line described in https://github.com/mapbox/mapbox-gl-js/pull/9694#issuecomment-634921847, the max step of the line gradient is still very large, meaning that one wouldn't be able to use this expression to update a car's position on a long route.

karimnaaji:
>I find that the step expression supports a precision of ~0.0001

That sounds about right. To clarify with some numbers what you're experiencing: at z18, from east to west (north america), that's ~8k tiles that we try to represent with 4 096 000 px at 512 px a tile, so we need changes w/ 2.44140625e-7 accuracy; which we can now represent in uv space with an f32 in https://github.com/mapbox/mapbox-gl-js/pull/9694/commits/2c7ad4685a61d74a43da6c86ec349a7f713d16a1, but depending on the hardware max texture size, we hit a bound between 0.00024414062 to 0.00006103515 for that specific case... We would need a texture of about 4096*1024, (2.38418579e-7 accuracy) to match the line progress, that's 4mb of data.

Two more ideas:
(1) Increase texture size even more and wrap texture at line ending: uses more texture space and wouldn't get us that far ahead in precision, a lot more data to update, and more memory involved (the 4mb mentioned above): I have performance concerns with that, we shouldn't do it.
(2) Tile those textures, that means we render them partially between the line clips, while adjusting the uvs based on those: one issue to overcome is that the clips are per line and multiple lines can be drawn in the same draw call. But this should get us to the tile unit accuracy.

karimnaaji:
(2) implemented in https://github.com/mapbox/mapbox-gl-js/pull/9694/commits/9d73d1a94c638e21679aaa5907c4ebf49937bc55, but needs a last optimization and cleanup pass.

karimnaaji:
@ansis We're done hunting for better precision, please have a look.

test/integration/render-tests/line-gradient/gradient-step-zoomed-cross-continental/style.json
```diff
@@ -0,0 +1,98 @@
...
+              "type": "LineString",
+              "coordinates": [
+                [
+                  -77.033643,
+                  38.899926
+                ],
+                [
+                  -120.033643,
+                  38.899926
+                ]
```
1ec5, line 33: This straight line is 3&nbsp;686&nbsp;kilometers long. Assuming the intention is to support gradient lines of up to the same distance in gl-native for use in the Mapbox Navigation SDK, it might be worth considering that the Mapbox Directions API can return a route [up to 10&nbsp;000&nbsp;kilometers long](https://docs.mapbox.com/api/navigation/#directions-api-restrictions-and-limits). I’m not sure how common that would be in practice, however. Another mitigating factor would be that the navigation SDK cuts the route line up by leg.

karimnaaji: Yes this test is for testing potential visual regression, not necessarily stress testing the issues we're trying to overcome. I wouldn't see an issue with having routes that long.

src/style/style_layer/line_style_layer.js
```diff
@@ -44,7 +44,7 @@ class LineStyleLayer extends StyleLayer {
     _unevaluatedLayout: Layout<LayoutProps>;
     layout: PossiblyEvaluated<LayoutProps>;
 
-    gradientTextureInvalidated: boolean;
+    gradientGeneration: number;
```
asheemmamoowala, line 47: nit: The `ImageAtlas` and `ImageManager` use `version` to represent multiple generations of the same resource. Should we be consistent with that terminology?

karimnaaji: Makes sense! Updated.

src/shaders/line_gradient.fragment.glsl
```diff
@@ -22,9 +29,15 @@ void main() {
     float blur2 = (blur + 1.0 / u_device_pixel_ratio) * v_gamma_scale;
     float alpha = clamp(min(dist - (v_width2.t - blur2), v_width2.s - dist) / blur2, 0.0, 1.0);
 
-    // For gradient lines, v_lineprogress is the ratio along the entire line,
-    // scaled to [0, 2^15), and the gradient ramp is stored in a texture.
-    vec4 color = texture2D(u_image, vec2(v_lineprogress, 0.5));
+    highp float texel_height = 1.0 / u_image_height;
+    highp float half_texel_height = 0.5 * texel_height;
+    highp vec2 uv = vec2(
```
karimnaaji, line 34: Note that this introduces a dependent texture read (prevents prefetch), most es 3.0 capable hardware won't take the hit but it's worth profiling.

src/data/bucket/line_bucket.js
```diff
@@ -517,11 +533,19 @@ class LineBucket implements Bucket {
...
             // the `w` component.
-            ((dir === 0 ? 0 : (dir < 0 ? -1 : 1)) + 1) | ((linesofar & 0x3F) << 2),
-            linesofar >> 6);
+            ((dir === 0 ? 0 : (dir < 0 ? -1 : 1)) + 1) | ((linesofarScaled & 0x3F) << 2),
+            linesofarScaled >> 6);
+
+        // Constructs a second vertex buffer with higher precision line progress
+        if (this.lineClips) {
+            this.layoutVertexArray2.emplaceBack(
+                this.scaledDistance - this.lineClips.start,
```
karimnaaji, line 545: By always realigning the distance with clip start and rendering the gradient between start and end while tiling it, we allow to have way larger distances represented, this is where the increase in precision comes from.

The last parameter of this new attribute layout is used to select which line the gradient falls through when multiple lines are represented in the same bucket geometry.

karimnaaji:
@ansis PTAL and let me know if anything would help or should be made more clear to facilitate the review. Note that the focus is on precision increase, we may want to address any potential other improvements in other patches as this already gets us a step forward.

src/util/util.js
```diff
@@ -209,6 +212,39 @@ export function uuid(): string {
     return b();
 }
 
+/**
+ * Return whether a given value is a power of two
+ * @private
+ */
+export function isPowerOfTwo(value: number): boolean {
+    return Math.log2(value) % 1 === 0;
```
mourner, line 220: `Math.log2` is [not available](https://caniuse.com/#feat=mdn-javascript_builtins_math_log2) in IE11, you have to use `Math.log(value) / Math.LN2`.

src/util/util.js
```diff
@@ -209,6 +212,39 @@ export function uuid(): string {
...
+}
+
+/**
+ * Return the next power of two, or the input value if already a power of two
+ * @private
+ */
+export function nextPowerOfTwo(value: number): number {
+    if (value <= 1) return 1;
+    return Math.pow(2, Math.ceil(Math.log2(value)));
```
mourner, line 229: Eliminate `Math.log2` too

src/util/color_ramp.js
```diff
@@ -1,28 +1,63 @@
...
+            renderPixel(0, j, progress);
+        }
+    } else {
+        for (let clip = 0, stride = 0; clip < height; ++clip, stride += width * 4) {
+            for (let i = 0, j = 0; i < width; i++, j += 4) {
+                // Remap progress between clips
+                const progress = i / (width - 1);
+                const evaluationProgress = mapValue(progress, 0.0, 1.0,
+                    params.clips[clip].start,
+                    params.clips[clip].end);
```
mourner, line 55: Looks like `mapValue` is only used in this one place, and with 0–1 range, so it's probably not necessary. How about rewriting as:

```js
const {start, end} = params.clips[clip];
const evaluationProgress = interpolate(start, end, progress);
```

karimnaaji: Yep good call, updated it here and in the shader.

src/util/util.js
karimnaaji: Updated

src/util/util.js
karimnaaji: Thanks, wasn't aware of that, updated!

src/data/bucket/line_attributes_ext.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import {createLayout} from '../../util/struct_array';
+
+const lineLayoutAttributesExt = createLayout([
+    {name: 'a_line_progress', components: 1, type: 'Float32'},
+    {name: 'a_line_clip', components: 1, type: 'Float32'},
+    {name: 'a_split_index', components: 1, type: 'Float32'},
+]);
```
ansis, line 8: This makes vertices three times bigger with 24 bytes per vertex instead of 8 bytes per vertex when the clip stats are present. Right now this would only affect geojson layers, right?

Do `a_line_progress` and `a_line_clip` need to be stored separately or can we store `a_line_progress / a_line_clip` as a single attribute?

karimnaaji: This extension is an optional 12 bytes per vertex which only affects geojson layers that request for line metrics, memory is the main reason to have them in a separate buffer. It's a total of 8 (LineLayoutArray) + 12 (LineExtLayoutArray) for the geojson + line metrics code path.

If we want to maximize accuracy I don't think we would be able to lower precision here, it would be quite destructive for what we're trying to improve. But there are other ways we can save on memory by using vertex divisor. Clips and split_index replicate across the geometry so it would be a good candidate for instancing optimization: https://developer.mozilla.org/en-US/docs/Web/API/ANGLE_instanced_arrays.

karimnaaji: As for using a single attribute I have no preference, we could directly bake it in a single if that makes more sense?
```
{name: 'a_line_metrics', components: 2, type: 'Float32'},
{name: 'a_split_index', components: 1, type: 'Float32'},
```

ansis: By baking into a single attribute I meant doing the division of `a_line_progress / a_line_clip` before adding it to the buffer rather than in the shader. Would this work?

karimnaaji: Oh right sorry I misread. I think that should work, we're in the range 0..1 and should stay within that range even by doing the division beforehand, so it should be safe to do so. I'll update for that.

karimnaaji: Added in https://github.com/mapbox/mapbox-gl-js/pull/9694/commits/2adb186f726e05fead7fb855b688de22ecaa49fe as well as some other improvements.

PR #9702, opened 2020-05-19
[doc] Document `volatile` source property

pozdnyakov:
The source `volatile` flag value can be defined in the style as following:

```
"example_source": {
   "volatile": true
}
```

For the sources with the `volatile` flag set, the tiles are not stored in the local storage. The `volatile` flag is applicable for all source types, which allow loading tiles from a remote server i.e. `vector`, `raster`, `raster-dem` source types.

### DIFF
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -184,7 +184,7 @@
     "volatile": {
       "type": "boolean",
       "default": false,
-      "doc": "Setting this flag to true indicates that the source tiles shall not be stored in the local cache." 
+      "doc": "A setting to determine whether a source's tiles are cached locally."
     },
     "*": {
       "type": "*",
@@ -259,7 +259,7 @@
     "volatile": {
       "type": "boolean",
       "default": false,
-      "doc": "Setting this flag to true indicates that the source tiles shall not be stored in the local cache." 
+      "doc": "A setting to determine whether a source's tiles are cached locally."
     },
     "*": {
       "type": "*",
@@ -334,7 +334,7 @@
     "volatile": {
       "type": "boolean",
       "default": false,
-      "doc": "Setting this flag to true indicates that the source tiles shall not be stored in the local cache." 
+      "doc": "A setting to determine whether a source's tiles are cached locally."
     },
     "*": {
       "type": "*",
--- a/src/style-spec/reference/v8.json
+++ b/src/style-spec/reference/v8.json
@@ -181,6 +181,11 @@
       "type": "promoteId",
       "doc": "A property to use as a feature id (for feature state). Either a property name, or an object of the form `{<sourceLayer>: <propertyName>}`. If specified as a string for a vector tile source, the same property is used across all its source layers."
     },
+    "volatile": {
+      "type": "boolean",
+      "default": false,
+      "doc": "Setting this flag to true indicates that the source tiles shall not be stored in the local cache." 
+    },
     "*": {
       "type": "*",
       "doc": "Other keys to configure the data source."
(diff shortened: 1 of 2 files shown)


src/style-spec/reference/v8.json
```diff
@@ -181,6 +181,11 @@
       "type": "promoteId",
       "doc": "A property to use as a feature id (for feature state). Either a property name, or an object of the form `{<sourceLayer>: <propertyName>}`. If specified as a string for a vector tile source, the same property is used across all its source layers."
     },
+    "volatile": {
+      "type": "boolean",
+      "default": false,
+      "doc": "Setting this flag to true indicates that the source tiles shall not be stored in the local cache."
```
chloekraw, line 187:
```suggestion
      "doc": "A setting to determine whether a source's tiles are cached locally." 
```

chloekraw: I think this language is a bit closer to how other source properties are documented.

src/style-spec/reference/v8.json
```diff
@@ -251,6 +256,11 @@
       "type": "string",
       "doc": "Contains an attribution to be displayed when the map is shown to a user."
     },
+    "volatile": {
+      "type": "boolean",
+      "default": false,
+      "doc": "Setting this flag to true indicates that the source tiles shall not be stored in the local cache."
```
chloekraw, line 262: @pozdnyakov we should update the documentation here too so that the language is the same everywhere

src/style-spec/reference/v8.json
```diff
@@ -321,6 +331,11 @@
       "default": "mapbox",
       "doc": "The encoding used by this source. Mapbox Terrain RGB is used by default"
     },
+    "volatile": {
+      "type": "boolean",
+      "default": false,
+      "doc": "Setting this flag to true indicates that the source tiles shall not be stored in the local cache."
```
chloekraw, line 337: @pozdnyakov we should update the documentation here too so that the language is the same everywhere

src/style-spec/reference/v8.json
pozdnyakov: Sure, I just applied your proposal (thanks for it!) to pull it locally :) with the latest patch this change is applied everywhere

pozdnyakov:
> Could you add an SDK Support table for the non-standard portions of each source type that are not part of the TileJSON specification

The `volatile` flag has not yet been officially released for any platform,  so not sure what to put to the table.. wdyt?

src/style-spec/reference/v8.json
chloekraw: oh gotcha, thanks!

chloekraw:
> The volatile flag has not yet been officially released for any platform, so not sure what to put to the table.. wdyt?

hmmmm, good question. Atm, I'm 👍 to leave it as "Not yet available" for all SDKs, merge this PR, and we will push an update to the docs in new PRs if we expose them in the next minor release of the iOS or Android SDK, when timing and planning for those become clearer.

pozdnyakov:
> Atm, I'm +1 to leave it as "Not yet available" for all SDKs

Done, with the latest patch

src/style-spec/reference/v8.json
```diff
@@ -181,6 +181,19 @@
       "type": "promoteId",
       "doc": "A property to use as a feature id (for feature state). Either a property name, or an object of the form `{<sourceLayer>: <propertyName>}`. If specified as a string for a vector tile source, the same property is used across all its source layers."
     },
+    "volatile": {
+      "type": "boolean",
+      "default": false,
+      "doc": "A setting to determine whether a source's tiles are cached locally.",
+      "sdk-support": {
+        "basic functionality": {
+          "js": "Not yet available",
```
chloekraw, line 190: Oh, doesn't the "not yet available" show up automatically if you exclude the platform from the `basic functionality: { }` section?

PR #9703, opened 2020-05-19
add a guard check to Attribute binder expression

mannnick24:
intermittent console error: http://api.mapbox.com/mapbox-gl-js/v1.10.1/mapbox-gl.js 28:193407 Uncaught TypeError: this.expression.evaluate is not a function

These are speculative changes intended to be the start of a discussion, as I'm not sure this gets to the heart of the matter

### DIFF
--- a/src/data/program_configuration.js
+++ b/src/data/program_configuration.js
@@ -26,6 +26,7 @@ import type {StructArray, StructArrayMember} from '../util/struct_array';
 import type VertexBuffer from '../gl/vertex_buffer';
 import type {ImagePosition} from '../render/image_atlas';
 import type {
+    Expression,
     Feature,
     FeatureState,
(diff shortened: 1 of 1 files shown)


mannnick24:
Hey @asheemmamoowala not been able to reproduce in a render test, except in a selenium test environment, I'll have one more try at a unit test.

mannnick24:
@asheemmamoowala I should be able to look at this shortly

ryanhamley:
@mannnick24 I'm going to close this as stale. Since we don't understand the cause of this bug, it may be better to start with opening an issue for more investigation, especially once you can isolate a reproducible example. If you can figure out a test that consistently shows this though, you can reopen this PR and we can reassess it then. Thanks.

PR #9708, opened 2020-05-20
You won't believe this ONE WEIRD TRICK!

arindam1993:
After having seen this https://youtu.be/ff4fgQxPaO0
the TLDW ( Too long did not watch ) for the video is:
JSON parsing is significantly faster than JS parsing, so by turning inline object literals into strings and wrapping  them with `JSON.parse` you prevent the JS parser from working extra hard.

While doing this I also discovered that we were importing the entirety of `package.json` into our bundle, just to read the current version, so I replaced it with a global `__SDK_VERSION` that gets replaced at build time. Yay! bundle size savings!

## Launch Checklist

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Reduce bundle size and inline style-spec as a JSON string to improve initial js parsing performance.</changelog>`


ryanhamley:
@arindam1993 is this something we still want to pursue or should we close it?

arindam1993:
closer

PR #9751, opened 2020-06-03
Improve rendering performance of symbols with symbol-sort-key

osvodef:
Pull request https://github.com/mapbox/mapbox-gl-js/pull/7678/ introduced sorting of symbol features across tiles. Unfortunately, this kind of sorting requires that symbols with distinct values of sort key are rendered in separate draw calls.

This can be a huge performance hit for a large number of labels, especially if `symbol-sort-key` is data driven (e. g. if we want to prioritize city labels by population). In such a case pretty much every label will end up in a separate draw call, which can result in hundreds of extra draw calls.

However, this sorting is only required if symbols of the current layer can overlap (i. e. when one of the `*-allow-overlap` or `*-ignore-placement` props is true). Otherwise, the rendering order is not important. This pull requests adds a check for this. If it detects that labels of the current layer cannot overlap, it puts all of them in one segment as before (and they get rendered in one draw call).

This only affects rendering order and should not affect prioritization during placement at all.

## Launch Checklist


src/render/draw_symbol.js
```diff
@@ -236,7 +236,12 @@ function drawLayerSymbols(painter, sourceCache, layer, coords, isText, translate
     // Unpitched point labels need to have their rotation applied after projection
     const rotateInShader = rotateWithMap && !pitchWithMap && !alongLine;
 
-    const sortFeaturesByKey = layer.layout.get('symbol-sort-key').constantOr(1) !== undefined;
+    const canOverlap = layer.layout.get('icon-ignore-placement') ||
```
osvodef, line 239: This should probably be refactored into a separate function, but I'm not sure about the best place for it

chloekraw:
cc @pozdnyakov

ryanhamley:
@mourner can you review this since Mikko is out of office for a bit?

src/render/draw_symbol.js
mourner: I think we should try add this to SymbolBucket's constructor to avoid the overhead of retrieving this value for every feature while it's guaranteed to be the same, and because there's already similar set up happening there (`this.sortFeaturesByKey`). In fact a very similar logic is already present there for `sortFeaturesByY`.

osvodef: Nice! I made `canOverlap` a field in `SymbolBucket`, and used it to calculate `sortFeaturesByY` as well.

src/render/draw_symbol.js
```diff
@@ -333,7 +334,8 @@ function drawLayerSymbols(painter, sourceCache, layer, coords, isText, translate
             hasHalo
         };
 
-        if (sortFeaturesByKey) {
+        if (hasSortKey && bucket.canOverlap) {
+            sortFeaturesByKey = true;
```
osvodef, line 338: I'm just not sure about this particular bit — we need `sortFeaturesByKey` after this loop, so I had to assign it like this.

src/data/bucket/symbol_bucket.js
```diff
@@ -609,6 +614,7 @@ class SymbolBucket implements Bucket {
                lineOffset: [number, number],
                alongLine: boolean,
                feature: SymbolFeature,
+               layer: SymbolStyleLayer,
```
mourner, line 617: `layer` is no longer used, so this is a leftover right?

src/render/draw_symbol.js
mourner: Yeah, this looks good.

src/symbol/symbol_layout.js
```diff
@@ -665,6 +667,7 @@ function addSymbol(bucket: SymbolBucket,
                 iconOffset,
                 iconAlongLine,
                 feature,
+                layer,
```
mourner, line 670: 3 `layer` leftovers here too,

src/data/bucket/symbol_bucket.js
osvodef: Oh, didn't catch that! Fixed 🙂

src/symbol/symbol_layout.js
osvodef: Fixed

mourner:
Thank you!

Issue #9755, opened 2020-06-03
Prevent draggable marker touch event propagation


MarcL:
**mapbox-gl-js version**: v1.9.0

### Question

I'm using Mapbox in a Facebook Messenger chatbot which renders the map inside an HTML webview. When dragging a draggable marker, the `touchmove` event also gets propagated to the parent webview. This then moves it up and down. I can't immediately see a way to get the event in the drag handlers so that I could call `event. stopPropagation()`.

Here's an example of the issue:

![mapbox-messenger](https://user-images.githubusercontent.com/1965510/83686993-0e0b4400-a5e3-11ea-99ad-f080d71de393.gif)

Is there an accepted method to achieve this?
Can I get access to the events in the drag handlers or is there a better way to do this?

Thanks!

### Links to related documentation

https://docs.mapbox.com/mapbox-gl-js/api/markers/#marker-events

mourner:
Try this:

```js
map.on('touchmove', (e) => e.originalEvent.stopPropagation());
```

MarcL:
Hi @mourner.

Thanks for the reply. That's perfect - thank you.
I didn't know about `originalEvent` so I was attempting to do `e.stopPropagation()`.
I also had to focus the map too as I have an input on the screen.

Thanks,
Marc

MarcL:
Hi @mourner! 👋 

Sorry to re-open this issue. I've just got back to working on the Messenger webview and your solution above only solves half of the problem.

It now avoids propagating the `touchmove` event in the map and that works great.
However, when you have a draggable marker, it moves the webview when you move that around as shown in the gif below:

![mapbox-webview-touch](https://user-images.githubusercontent.com/1965510/84602062-ab386900-ae7c-11ea-98ca-1bb1c52c61fb.gif)

I thought I might be able to stop event propagation on the `drag` event for a `Marker` but it doesn't appear to give me the `event.originalEvent` in the same way as the `Map` handler does.

Some example code I tried but it doesn't work as `originalEvent` is `undefined`:

```javascript
const marker = new mapboxgl.Marker({
        draggable
    })
    .setLngLat([longitude, latitude])
    .addTo(map);

// Complains that "event.originalEvent" is undefined
marker.on('drag', (event) => event.originalEvent.stopPropagation();
```

I'm sure it's something obvious again but is there a way to allow dynamic markers but avoid the event propagation to the parent webview again?

Thanks!

MarcL:
Hi @mourner.

[...]

PR #9757, opened 2020-06-03
fix double tap zoom on iOS 13

### LINKED ISSUE #9756: double tap to zoom in does not work on iOS 13
**mapbox-gl-js version**: 1.10.1

**browser**: iOS Safari 13

### Steps to Trigger Behavior

 1. open any map in iOS safari 13
 2. double tap

### Link to Demonstration

https://docs.mapbox.com/mapbox-gl-js/example/simple-map/

### Expected Behavior

Map zooms in.

### Actual Behavior

Map does not zoom in.

ansis:
fix #9756

iOS 13 does not fire the second touchstart/end events in a double tap if the touchstart listener is non-passive. This fix makes it passive.

Calling preventDefault() allows the second event to be fired but suppresses other events like `click`.

`touchmove` needs to remain non-passive so that it can be used to prevent touches from scrolling or scaling the page on some versions of iOS Safari.

---

This fix does NOT work if a parent of the canvas (including the window/document/body) has a non-passive touchstart listener that doesn't call `preventDefault()`. I don't know of a fix for this that doesn't suppress all click dom events as well.

---

## Launch Checklist

 - [x] add an entry inside this element for inclusion in the `mapbox-gl-js` changelog: `<changelog>Fix zooming with a double tap on iOS SafarI 13.</changelog>`

I have done initial testing on iOS safari v12 and v13 but these needs to be tested more thoroughly on other touch devices.

### DIFF
--- a/src/ui/handler_manager.js
+++ b/src/ui/handler_manager.js
@@ -128,12 +128,14 @@ class HandlerManager {
         const el = this._el;
 
         this._listeners = [
-            // Bind touchstart and touchmove with passive: false because, even though
-            // they only fire a map events and therefore could theoretically be
-            // passive, binding with passive: true causes iOS not to respect
-            // e.preventDefault() in _other_ handlers, even if they are non-passive
-            // (see https://bugs.webkit.org/show_bug.cgi?id=184251)
-            [el, 'touchstart', {passive: false}],
+            // This needs to be `passive: true` so that a double tap fires two
+            // pairs of touchstart/end events in iOS Safari 13. If this is set to
+            // `passive: false` then the second pair of events is only fired if
+            // preventDefault() is called on the first touchstart. Calling preventDefault()
+            // undesirably prevents click events.
+            [el, 'touchstart', {passive: true}],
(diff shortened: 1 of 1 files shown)


ansis:
Note: changing the listener to passive means that this won't prevent a click event from being fired anymore:

```
map.on('touchstart', e => e.originalEvent.preventDefault());
```

This might be considered a breaking change but:
- it was never explicitly supported
- it hasn't always been supported

ansis:
> I'd approve this but I think the fact that map.on('touchstart', e => e.originalEvent.preventDefault()); no longer works may deserve some discussion. It's not officially supported but preventing the original event seems to be a fairly common use case and we do point it out as a possible solution. Vlad did so on a ticket today #9755 (comment) albeit for touchmove. It also seems weird that this workaround would be available for every event except for touchstart.

The linked example uses `stopPropagation()` which will continue working the same way. The only thing preventDefault() on touchstart does in this case is prevent the relevant mouse events from firing. While someone may be relying on this behavior I think it's fairly unlikely.

I see no other way of working around this browser bug

PR #9764, opened 2020-06-07
Include in package only necessary files

TrySound:
mapbox-gl package is very big (https://packagephobia.now.sh/result?p=mapbox-gl).

In this diff I'm trying to reduce it a bit but excluding build files.

- .flowconfig and flow-typed in packages are not used by flow
- build scripts are useless without dev dependencies
- all files in dist are published

## Launch Checklist

 - [ ] post benchmark scores

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,11 +163,7 @@
     "codegen": "build/run-node build/generate-style-code.js && build/run-node build/generate-struct-arrays.js"
   },
   "files": [
-    "build/",
-    "dist/mapbox-gl*",
-    "dist/style-spec/",
-    "flow-typed/*.js",
-    "src/",
-    ".flowconfig"
+    "dist/",
+    "src/"
   ]
 }


karimnaaji:
>In this diff I'm trying to reduce it a bit but excluding build files.

Thanks for the PR! Library size is definitely something we keep track of and put regular effort into reducing. How much of a saving did you notice with these new filtering rules?

package.json
```diff
@@ -163,11 +163,7 @@
     "codegen": "build/run-node build/generate-style-code.js && build/run-node build/generate-struct-arrays.js"
   },
   "files": [
-    "build/",
```
karimnaaji, line 166: It's a bit of a risk for us to change the content of the package as users may have a hard dependency on these.

>build scripts are useless without dev dependencies

Agreed, though it's really hard to know what type of workflow some library users have beforehand, maybe the dev dependencies are added by them. 

>all files in dist are published

I think this is one would be a no-op and we preferred being explicit, changing this shouldn't have any impact and prevents accidental addition.

>.flowconfig and flow-typed in packages are not used by flow

I can't confirm this one. Do you know @arindam1993 ?

arindam1993: yes this is correct

arindam1993:
I think we should try and decrease the install size, in theory all we need to publish is `/dist` since we only offically support browsers and our prebundled files.
That approach is what would net meaningful install-size improvements, but over the years we've accumulated a few users both internally at mapbox and externally who have custom build systems that rely on us including source files.

While getting rid of the flow and `/build` files makes sense, the impact to install size is minimal 1-2 Megs for a 35meg install.

Please feel free to most an issue on why you'd like a smaller install size @TrySound , we can definitely think about doing the heavy part ( removing `/src` and sourcemaps )  if there's enough interest from the community.

Closing this because of the risk to some downstream consumers.

arindam1993:
Also @TrySound , your build system should only be pulling in the ~190kb `dist/mapboxgl.js` and nothing else, so the actual app bundle should be fairly small.

TrySound:
I'm not concerned about bundle size here, only about node_modules bloating. I will reach out you when I'll have better proposal.

Issue #9662, opened 2020-05-06
Opening a marker's pop-up should set focus to the pop-up's content


skorasaurus:
**mapbox-gl-js version**:
confirmed on 1.10.0

**browser**:
confirmed with firefox 75 on ubuntu and chrome on windows 

### Steps to Trigger Behavior

 1. [Open demonstration page](http://bl.ocks.org/skorasaurus/40d6f977d2db8038b99ed6aca5354f73)
 2. press tab several times so a marker a has focus
 3. press enter to open its pop-up
 4. Press tab again

### Link to Demonstration
Demonstration source
https://gist.github.com/skorasaurus/40d6f977d2db8038b99ed6aca5354f73 

Viewable demonstration
http://bl.ocks.org/skorasaurus/40d6f977d2db8038b99ed6aca5354f73

### Expected Behavior

the keyboard focus should change to the pop-up's contents of the marker 
whether through a designated keydown event (pressing enter on the keyboard while focus is on the marker) or the click event. 

In other words, a user who is only able to use a keyboard will tab through 
each tabbable element, reach a marker that they would like to access, press enter to open, and expects that during their next tab, the focus would be on the first 
tabbable element (the link, a heading, or a close button for the pop-up)
in that pop-up.  

### Actual Behavior

After a user opens the pop-up, and presses tab, the focus changes to the next 
item in the source. 

In the test case that loads a geojson array of 27 elements, if a user tabs after opening, the focus changes to the next point in the source data. As a result, if the user opens up a pop-up for a point that was 4th in the source data, they'll have to tab through the remaining 23 entries, plus any other elements on the page. 

This is a bug because users who are only able to use a keyboard to navigate have to go through significant more work in order to access the content of a pop-up (which may contain a link).

ryanhamley:
Thanks for the thorough report @skorasaurus! We welcome contributions, if you're interested in supplying a pull request to fix this issue. Our [contributor guidelines](https://github.com/mapbox/mapbox-gl-js/blob/master/CONTRIBUTING.md) can get you started.

avpeery:
Hi @ryanhamley it looks like this issue was addressed in the merged PR #9774 (https://github.com/mapbox/mapbox-gl-js/pull/9774). Should this issue be closed? Thanks!

ryanhamley:
Yep looks like an oversight. Thanks for checking @avpeery

PR #9774, opened 2020-06-10
9662 bug a11y focus popup content

watofundefined:
Hi folks - this is my first PR - so please shout if I missed some of your rules or anything!

## Launch Checklist

PR for bug: https://github.com/mapbox/mapbox-gl-js/issues/9662

### Commit b546e09

**Q1**: Should consumers be able to disable this auto-focus functionality?

**Q2** Should <kbd>Esc</kbd> close the opened popup? Currently it doesn't.

Otherwise it's nothing big - for better keyboard UX I changed the order in which elements in popup are added to DOM - this shouldn't change anything since the close button has `position: absolute;`.

I also wanted to add functionality that re-focuses the previously focused element when popup is closed.
But it was getting messy in one scenario:

- **user opens a popup** -> popup stores a reference to a previously focused element
- **user clicks on a different marker** -> new popup is opened and *after* that the old popup is removed - this would result in a situation where the new popup would store a reference to a focusable element from the old popup (already bad but not too bad)
- **user clicks on a different marker** -> same thing happens **but** since the new popup is added before the old one is removed we would first focus element in the new popup and then the popup which is on its way to DOM graveyard would jump in and it would try to re-focus the element from the very first popup which is long gone - so the third popup would lose focus and `document.activeElement` would be either `null` or `document.body`.

Both popup's removal (popup.js) and popup's appearing (marker.js) are driven by map's "click" event and I don't see an easy way how to ensure they happen in different order.

So I was thinking that the marker could pass a reference to its DOM element to the popup instance.
Then popup would focus it when it'll be closed - but only if popup was closed by the 'x' button - i.e. not by clicking on another marker.
It' an easy change and it would give users good keyboard UX. (I'm happy to add it to the PR if you think it's a good idea)

### Commit 068d10d  

I added a lot of details to the commit message but I don't know mapbox-gl-js that well so please think if there are some valid cases when map's container is scrolled (it's the `div` with id "mapid")

### DIFF
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -38,7 +38,7 @@ export type PopupOptions = {
 const focusQuerySelector = [
     "a[href]",
     "[tabindex]:not([tabindex='-1'])",
-    "contenteditable",
+    "[contenteditable]:not([contenteditable='false'])",
     "button:not([disabled])",
     "input:not([disabled])",
     "select:not([disabled])",
--- a/src/ui/popup.js
+++ b/src/ui/popup.js
@@ -128,6 +128,7 @@ export default class Popup extends Evented {
 
         this._map.on('remove', this.remove);
         this._update();
+        this._focusFirstElement();
 
         if (this._trackPointer) {
             this._map.on('mousemove', this._onMouseMove);
@@ -397,8 +398,11 @@ export default class Popup extends Evented {
      */
     setDOMContent(htmlNode: Node) {
         this._createContent();
+        // The close button should be the last tabbable element inside the popup for a good keyboard UX.
         this._content.appendChild(htmlNode);
+        this._createCloseButton();
         this._update();
+        this._focusFirstElement();
         return this;
     }
 
@@ -455,6 +459,9 @@ export default class Popup extends Evented {
         }
 
         this._content = DOM.create('div', 'mapboxgl-popup-content', this._container);
+    }
+
+    _createCloseButton() {
         if (this.options.closeButton) {
             this._closeButton = DOM.create('button', 'mapboxgl-popup-close-button', this._content);
             this._closeButton.type = 'button';
@@ -462,7 +469,6 @@ export default class Popup extends Evented {
             this._closeButton.innerHTML = '&#215;';
             this._closeButton.addEventListener('click', this._onClose);
         }
-
     }
 
     _onMouseUp(event: MapMouseEvent) {
@@ -477,7 +483,7 @@ export default class Popup extends Evented {
         this._update(event.point);
     }
 
-    _update(cursor: PointLike) {
+    _update(cursor: ?PointLike) {
         const hasPosition = this._lngLat || this._trackPointer;
 
         if (!this._map || !hasPosition || !this._content) { return; }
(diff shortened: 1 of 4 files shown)


src/ui/popup.js
```diff
@@ -542,6 +548,28 @@ export default class Popup extends Evented {
         applyAnchorClass(this._container, anchor, 'popup');
     }
 
+    _focusFirstElement() {
+        if (!this._container) return;
+
+        // This approach isn't covering all the quirks and cases but it should be good enough.
+        // If we would want to be really thorough we would need much more code, see e.g.:
+        // https://github.com/angular/components/blob/master/src/cdk/a11y/interactivity-checker/interactivity-checker.ts
```
ryanhamley, line 556: I think we probably don't need this comment in the code but it would be good to document this in the PR so that we can reference it in the future if we want to make this more robust.

src/ui/popup.js
```diff
@@ -477,7 +483,7 @@ export default class Popup extends Evented {
         this._update(event.point);
     }
 
-    _update(cursor: PointLike) {
+    _update(cursor: ?PointLike) {
```
ryanhamley, line 486: Did something change that made it necessary to make this a Maybe type?

src/ui/popup.js
```diff
@@ -542,6 +548,28 @@ export default class Popup extends Evented {
         applyAnchorClass(this._container, anchor, 'popup');
     }
 
+    _focusFirstElement() {
+        if (!this._container) return;
+
+        // This approach isn't covering all the quirks and cases but it should be good enough.
+        // If we would want to be really thorough we would need much more code, see e.g.:
+        // https://github.com/angular/components/blob/master/src/cdk/a11y/interactivity-checker/interactivity-checker.ts
+        const selectors = [
```
ryanhamley, line 557: There's no need to allocate this array every time this function is called. We could move it out of the function. You could also save it as a `querySelector` to avoid `join`ing the array every time.

watofundefined:
Hey @ryanhamley, no problem at all, thanks for feedback!

I've added a new popup option `focusAfterOpen`, defaulting to `true`.

As for the bug, I can replicate the behaviour you describe in both Firefox and Chrome.
It's a weird one though - it seems like it's just the **browser's default** `outline` style that's missing.
When I check `document.activeElement` in console, I can see that the element is focused correctly.
When I add `a:focus {background-color: red;}`, it's also applied correctly.

Here's what I tried (without luck):
- setting focus on `body` when popup's closed, no luck there.
- wrapping the initial `focus()` in `setTimeout` to see if it's a timing issue 

Any ideas?

Cheers!

src/ui/popup.js
watofundefined: I moved it to the top-level scope - leaving the `[item1, item2].join(', ')` for better readability.
Let me know if that's fine with you - I can change it to whatever you think works better with the codebase.
I did try one long string, but it was around 190 characters which is just too much, and two strings on two lines didn't look right to me.

test/unit/ui/popup.test.js
```diff
@@ -651,3 +651,97 @@ test('Popup closes on Map#remove', (t) => {
...
+
+test('Element with tabindex="-1" is not focused', (t) => {
+    const popup = new Popup({closeButton: true})
+        .setHTML('<span tabindex="-1" data-testid="abc">Test</span>')
+        .setLngLat([0, 0])
+        .addTo(createMap(t));
+
+    const focusableEl = popup._container.querySelector("[data-testid='abc']");
+
+    t.notEqual(window.document.activeElement, focusableEl);
```
tristen, line 717: Can this assert below that the closeButton is in focus?

test/unit/ui/popup.test.js
```diff
@@ -651,3 +651,97 @@ test('Popup closes on Map#remove', (t) => {
...
+    t.end();
+});
+
+test('If popup content contains a disabled input followed by a focusable element then the latter is focused', (t) => {
+    const popup = new Popup({closeButton: true})
+        .setHTML(`
+            <button disabled>No focus here</button>
+            <span tabindex="0" data-testid="abc">Test</span>
```
tristen, line 725: For variance, Mind changing this to a different focusable element? like a contenteditable region or a:href?

tristen (changes requested):
👋 @watofundefined apologies on the delay here. @ryanhamley asked me to give this a second set of eyes. 

Weirdly, I can't reproduce the bug you and @ryanhamley ran into. If `document.activeElement` is still in focus correctly and its the browsers default handling of `outline` I personally think it's fine as is.

This overall looks great! Just had some small requests in the tests. Let's get this merged!

tristen:
👋 @watofundefined just checking in. This looks close! are you able to make these remaining adjustments?

test/unit/ui/popup.test.js
watofundefined: It seems `contenteditable` is not supported by jsdom so I used `select`. But while playing around with `contenteditable` I've noticed that the selector is wrong - hence the update in `popup.js`. 

I tested it locally and it works as expected for these variants:
- `<div contenteditable>...</div>` (focuses)
- `<div contenteditable="true">...</div>` (focuses)
- `<div contenteditable="false">...</div>` (doesn't focus)

watofundefined:
Hey @tristen, sorry for the delay, it totally slipped my mind.
Have a look at the latest commit and let me know if you spot anything else!

tristen (approved):
@watofundefined awesome! I'll let @ryanhamley hit the merge but these changes look great to me

ryanhamley (approved):

